PR #423, opened 2015-01-02
fix uninitialized member variables

Thoronador:
This pull request fixes some uninitialized member variables in various classes by initializing them to NULL, 0 or similar "standard" values.

For reasons why this should be done, see http://cwe.mitre.org/data/definitions/457.html.
>In some languages such as C and C++, stack variables are not initialized by default. They generally contain junk data with the contents of stack memory before the function was invoked. An attacker can sometimes control or read these contents. In other languages or conditions, a variable that is not explicitly initialized can be given a default value that has security implications, depending on the logic of the program. The presence of an uninitialized variable can sometimes indicate a typographic error in the code.


apps/opencs/model/world/refidadapterimp.cpp
```diff
@@ -431,7 +438,15 @@ void CSMWorld::MiscRefIdAdapter::setData (const RefIdColumn *column, RefIdData&
         InventoryRefIdAdapter<ESM::Miscellaneous>::setData (column, data, index, value);
 }
 
-CSMWorld::NpcColumns::NpcColumns (const ActorColumns& actorColumns) : ActorColumns (actorColumns) {}
+CSMWorld::NpcColumns::NpcColumns (const ActorColumns& actorColumns)
+: ActorColumns (actorColumns),
+  mFlags(std::map<const RefIdColumn *, unsigned int>()),
```
scrawl: This line is kinda pointless, that is what the default map constructor would be doing anyway.

Thoronador: This line might seem sort of pointless, but I've got the habit of initializing as much variables as possible in the initializer list. That's basically why this line is there, and it doesn't do any harm (except taking a few seconds to type it), since the map's default constructor would do something similar.

If that line bugs you, I sure can remove it (or you can do that during the merge, whatever is more convenient for you).

scrawl: I have the opposite habit personally. No-op initializers add redundancy, so when I want to change or replace that member variable I have to go hunting down all the constructors to fix them. This isn't a very big deal, just found it odd.

Thoronador: The compiler should catch such cases, if anyone changes the type of the variable or removes it.

Anyway, I removed it in a new commit: 44e01d0eaaad9
Include it, if you want to get rid of that line. Leave it out, if you don't.

Thoronador:
The Travis CI build failed on Mac, because it exceeded a time limit. However, the Linux build on Travis CI is just fine. This problem seems to affect all of the more recent builds, starting with [build 1434](https://travis-ci.org/OpenMW/openmw/builds/45621383) on the master branch.

I'm not quite sure whether this problem is specific to Mac OS X or just a problem with Travis, but maybe someone should look into that or more future builds might "fail" when they actually seem to be alright.

scrawl:
> The Travis CI build failed on Mac, because it exceeded a time limit. However, the Linux build on Travis CI is just fine. This problem seems to affect all of the more recent builds, starting with build 1434 on the master branch.
> 
> I'm not quite sure whether this problem is specific to Mac OS X or just a problem with Travis, but maybe someone should look into that or more future builds might "fail" when they actually seem to be alright.

I saw that. Summoning @corristo ...

psi29a:
Same problems as before, trying to get qt bottle from sourcefordge. @corristo poke poke, can we upload the bottle to our downloads please? :)

I'll try reloading the build.

zinnschlag:
This looks a bit over-zealous to me.

1. You are initialising some variables that are initialised anyway (for example variables of type std::string).
2. The load* structs aren't meant to be initialised in the constructor (actually they shouldn't really have a constructor). These are the typical dumb structs. They are either initialised by the load function or by the blank function.

psi29a:
A kickin the head, and travis osx behaves. :)

Thoronador:
@psi29a Thanks for fixing the OS X issue. :)

@zinnschlag:
> You are initialising some variables that are initialised anyway (for example variables of type std::string).

As I have already said to scrawl above, I've got the habit of initializing as much variables as possible in the constructor.

> The load* structs aren't meant to be initialised in the constructor (actually they shouldn't really have a constructor). These are the typical dumb structs. They are either initialised by the load function or by the blank function.

Maybe that is just my opinion, but there shouldn't be any "dumb structs". If - by mistake - the programmer forgets to call the blank() function or something similar, then we might end up with struct member variables that do not have a defined value. This can lead to undefinied/weird behaviour which is hard to debug. Just trying to save us some trouble here.
By the way, the constructors with initializer lists do not create any overhead, because the variables are already constructed with their initialized values in place instead of creating them with some random memory garbage as values. So I don't see why there shouldn't be initialization; it doesn't hurt performance (if that is what you are worried about) and it ensures that those structs have explicitly defined intial values. That's a good thing to have.

zinnschlag:
> By the way, the constructors with initializer lists do not create any overhead, because the variables are already constructed with their initialized values in place instead of creating them with some random memory garbage as values.

There is a small overhead, because the variable still needs to be initialised twice. But I wouldn't use that as an argument, because it would be a really small overhead. Just nitpicking here.

I can kinda follow your argumentation about not having dumb structures and initialising everything in the constructor, but that is just not how our components/esm is build (which wasn't my decision btw) and I value consistency way more here. So unless you want to overhaul all of components/esm, I suggest you change that back. Piecemeal improvements will only lead to confusion.

Thoronador:
> I can kinda follow your argumentation about not having dumb structures and initialising everything in the constructor, but that is just not how our components/esm is build (which wasn't my decision btw) and I value consistency way more here. So unless you want to overhaul all of components/esm, I suggest you change that back. Piecemeal improvements will only lead to confusion.

Since I do not have the time to dig through several thousand lines of code just to keep the consistency, I've changed that back: 773669952b5baaba0cd8b1faae059a9f5a0b5314.

PR #424, opened 2015-01-02
Opening window and User Settings window cleanup

maqifrnswa:
Feature #1325

- Added a critical pop-up window to notify users of failed file loading (sound on systems that support it, such as Windows)
- Made error text red to highlight error
- User settings window is created at current mouse location


zinnschlag:
Not okay with the message box. Seems redundant. And annoying. Note that we generally avoid message boxes in OpenCS.
The rest looks okay though.

maqifrnswa:
message box removed

PR #425, opened 2015-01-02
Made templated Nif readers

EmperorArthur:
The code redundancy for the vector geters was just annoying.  Plus, this makes the option of a version dependent getter for nif file data a single function away should we decide to support multiple nif versions in the future.

### DIFF
--- a/components/nif/nifstream.hpp
+++ b/components/nif/nifstream.hpp
@@ -5,6 +5,8 @@
 
 #include <stdint.h>
 #include <stdexcept>
+#include <typeinfo>
+#include <string>
 
 #include <OgreDataStream.h>
 #include <OgreVector2.h>
@@ -86,7 +88,7 @@ public:
 
     //Templated functions to handle reads
     template <typename T>
-    T get(){throw std::runtime_error("Can not get this type of data from a NIF File!");}
+    T get(){throw std::runtime_error("Can not read a <"+std::string(typeid(T).name())+"> from a NIF File!  The get() function was called with the wrong template!");}
 
     ///Return a vector of whatever object is needed
     template <typename T>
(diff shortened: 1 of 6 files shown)


EmperorArthur:
Oh it works.  Unfortunately, specializing in the header will throw "multiple declaration" compiler errors.

You can test it yourself using niftest.  Just pass a nif or bsa file as an argument to it and it will spit out an error if it can't read it.  Another option is to pass your Morrowind data directory to components/nif/tests/test.sh.  That will scan the whole thing for broken files.

I'd also love it if generics could throw compiler errors, but couldn't find a way to do so.

kcat:
> Oh it works.

Have you tried with optimizations enabled? In debug mode, it won't inline functions so the calls to the specific versions will remain, but when calls get inlined it won't realize there are specializations declared in a separate translation unit.

> Unfortunately, specializing in the header will throw "multiple declaration" compiler errors.

Mark the specialized definitions as ```inline```. C++ allows multiple definitions of inline functions, and the linker will just pick one of them.

components/nif/nifstream.hpp
```diff
@@ -84,12 +84,23 @@ class NIFStream {
...
 
-    void getShorts(std::vector<short> &vec, size_t size);
-    void getFloats(std::vector<float> &vec, size_t size);
-    void getVector2s(std::vector<Ogre::Vector2> &vec, size_t size);
-    void getVector3s(std::vector<Ogre::Vector3> &vec, size_t size);
-    void getVector4s(std::vector<Ogre::Vector4> &vec, size_t size);
-    void getQuaternions(std::vector<Ogre::Quaternion> &quat, size_t size);
+    //Templated functions to handle reads
+    template <typename T>
+    T get(){throw std::runtime_error("Can not get this type of data from a NIF File!");}
```
Mingun: For facilitation of debugging it is better instead of `"this type"` to place in the exception message information on specific type, through `typeid(T).name()`.

EmperorArthur:
It's compiling and working even in release builds.  I even tried to break it by making the default template inline.  It still works without a problem.

kcat:
I see. Still, it makes sense to inline the specializations since they're just forwarding to another call.

scrawl:
This pull request is causing crashes for some testers, see https://bugs.openmw.org/issues/2257 and https://bugs.openmw.org/issues/2250

PR #426, opened 2015-01-03
Make Travis use Qt from our Homebrew tap

corristo:


corristo:
Need to modify bottles and reupload them.

corristo:
Needs more work.

PR #427, opened 2015-01-04
Bug fixes

scrawl:


zinnschlag:
That does not build for me:

/home/marc/OpenMW/openmw/apps/openmw/mwmechanics/actors.cpp:39:25: fatal error: summoning.hpp: No such file or directory
compilation terminated.

scrawl:
That wasn't supposed to be committed. I have no idea how it ended up here. Fixed.

PR #428, opened 2015-01-04
Remove unused "skipOnZero" opcode

scrawl:
This opcode is never generated by the compiler and thus causes a warning:

```
components/compiler/generator.cpp:153:10: warning: ‘void {anonymous}::opSkipOnZero(Compiler::Generator::CodeContainer&)’ defined but not used [-Wunused-function]
     void opSkipOnZero (Compiler::Generator::CodeContainer& code)
```

### DIFF
--- a/components/compiler/generator.cpp
+++ b/components/compiler/generator.cpp
@@ -150,10 +150,13 @@ namespace
         code.push_back (Compiler::Generator::segment0 (2, offset));
     }
 
+    /*
+    Currently unused
     void opSkipOnZero (Compiler::Generator::CodeContainer& code)
     {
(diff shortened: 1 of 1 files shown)


zinnschlag:
Its currently not used. But it may be in the future as new control structure or optimisations are added. Is having an unused function really something the compiler should warn about?

zinnschlag:
Actually, since this function resides inside the cpp file only, I would be okay with removing the opSkipOnZero function. But the actual VM-instruction should not be removed (OpSkipZero).

scrawl:
Very well then, I commented out the opSkipOnZero function and left the rest alone.

PR #431, opened 2015-01-06
Bug fixes

scrawl:


zinnschlag:
I do not agree with this statement. Neither the clarify over correctness part nor that grid size is clearer. But not worth getting into a discussion about. Will merge

PR #433, opened 2015-01-07
Savegame mime type

scrawl:
Use an .omwsave extension for savegame files. Register a mimetype for freedesktop.org-compliant systems. 
This allows us to load .omwsave files by clicking on them in a file manager. In a system install of OpenMW, this should work automatically.
For testing this with a local build, you have to do the following:
<pre>
cp files/openmw.desktop ~/.local/share/applications/
cp files/openmw-mimeinfo.xml ~/.local/share/mime/packages/
update-mime-database ~/.local/share/mime
update-desktop-database ~/.local/share/applications
</pre>
Then create a script called "openmw" in your $PATH, which changes to the build folder and runs the executable:
```
#!/bin/bash
cd <Build folder>
./openmw $*
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -386,6 +386,8 @@ configure_file(${OpenMW_SOURCE_DIR}/files/opencs/defaultfilters
 if (NOT WIN32 AND NOT APPLE)
     configure_file(${OpenMW_SOURCE_DIR}/files/openmw.desktop
         "${OpenMW_BINARY_DIR}/openmw.desktop")
+    configure_file(${OpenMW_SOURCE_DIR}/files/openmw-mimeinfo.xml
+        "${OpenMW_BINARY_DIR}/openmw-mimeinfo.xml")
     configure_file(${OpenMW_SOURCE_DIR}/files/opencs.desktop
         "${OpenMW_BINARY_DIR}/opencs.desktop")
 endif()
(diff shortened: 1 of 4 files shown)


zinnschlag:
That would only work if the content file list matches that of the saved game file, right? Sounds like asking for trouble.

scrawl:
If attempting to load a savegame from a different game file than the currently selected one, an appropriate error should be reported to the user.

kcat:
What about available mods? If the mod list is completely different, you can expect nothing but problems. However, a slightly different mod list is not necessarily a problem.

In Skyrim, you'll get a warning if you try to load a savegame that lists mods that no longer exist. Perhaps OpenMW should display a similar warning if the mod list is different, since adding mods could be an issue too (depending on what the mod does and what's already been done in the save).

scrawl:
> In Skyrim, you'll get a warning if you try to load a savegame that lists mods that no longer exist. Perhaps OpenMW should display a similar warning if the mod list is different, since adding mods could be an issue too (depending on what the mod does and what's already been done in the save).

I guess we could add such a warning message, but I don't see what it has to do with this pull request.

zinnschlag:
When trying to load a saved game from inside OpenMW you are only getting the matching saved game files to choose from. This kind of pre-selection does not happen when starting OpenMW via a saved game file. We would introduce a new failure case that could be very confusing for some users.

scrawl:
I don't see your point. How is this a new failure case? It already exists now; having the non-matching files not show up in the UI is just one way to handle it. And one that might cause a similar kind of confusion. "why are the files not even showing up?" is definitely more frustrating then having to read an error message that actually tells you what is going on.

zinnschlag:
Okay. Let me word that more precise: With matching files I meant matching .omwgame file (which is what we currently have implemented).

Getting an error message in case of starting with a saved game file that does not match the game file sounds like a reasonable solution. However that does not seem to be what is currently implemented. If an exception is thrown during load we get an error message on the console and OpenMW enters silently the main menu. If no exception is thrown, OpenMW just continues in a broken state. Neither option is good and while loading with no exception is not a very likely case, we shouldn't rely on it.

@kcat That is actually a good point. The user may want to load a saved game with the list of .omwaddon files that are different from the time of saving. We should at least provide a warning. However I agree with scrawl that this is a separate issue. Do we have it on the tracker? If not, would you file one?

scrawl:
> If an exception is thrown during load we get an error message on the console and OpenMW enters silently the main menu.

Actually, the error message is shown to the user in a message box: https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwstate/statemanagerimp.cpp#L262
I added that a while ago.

zinnschlag:
Okay, I did not see this. Still not entirely happy about relying on exceptions being thrown, but this looks acceptable. Will merge.

PR #434, opened 2015-01-09
Slaughter fish attacks when player only knee deep in water (Fixes #2076)

dteviot:
The logic to decide if a pure water creature can attack (becuse target is in deep enough water)
or a non-swimmer can attact swimming target is duplicated in two places.
\apps\openmw\mwmechanics\aicombat.cpp, lines 209 - 214
\apps\openmw\mwmechanics\actors.cpp lines 311-315
Should really be pulled into single location, but I'm not sure where.
There's also a potential bug in that it's checking if the TARGET is at wading depth, should really check if the water is deep enough for the ATTACKER.
But I suspect it's good enough.

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1978,7 +1978,8 @@ namespace MWWorld
 
     bool World::isSubmerged(const MWWorld::Ptr &object) const
     {
-        return isUnderwater(object, 1.85f);
+        const float neckDeep = 1.85f;
+        return isUnderwater(object, neckDeep);
     }
 
     bool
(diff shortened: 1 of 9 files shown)


apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1978,25 +1978,34 @@ namespace MWWorld
...
     bool World::isSubmerged(const MWWorld::Ptr &object) const
     {
-        const float *fpos = object.getRefData().getPosition().pos;
-        Ogre::Vector3 pos(fpos[0], fpos[1], fpos[2]);
-
-        const OEngine::Physic::PhysicActor *actor = mPhysEngine->getCharacter(object.getRefData().getHandle());
-        if(actor) pos.z += 1.85*actor->getHalfExtents().z;
-
-        return isUnderwater(object.getCell(), pos);
+        return isUnderwater(object, 1.85f);
```
Mingun: I think, that magical constants like that must be documented.

PR #437, opened 2015-01-10
Fixes for #810 and #811

dteviot:


apps/launcher/datafilespage.cpp
```diff
@@ -306,7 +299,7 @@ void Launcher::DataFilesPage::checkForDefaultProfile()
 bool Launcher::DataFilesPage::showDeleteMessageBox (const QString &text)
 {
     QMessageBox msgBox(this);
-    msgBox.setWindowTitle(tr("Delete Profile"));
+    msgBox.setWindowTitle(tr("Delete Content List"));
     msgBox.setIcon(QMessageBox::Warning);
     msgBox.setStandardButtons(QMessageBox::Cancel);
     msgBox.setText(tr("Are you sure you want to delete <b>%0</b>?").arg(text));
```
Mingun: Looks like as incorrect placeholder: first argument must have placeholder `%1`.

dteviot:
FYI. The '%0" issue noted by Mingun did not make the code fail.  Looks like Qt has a bug in that the number after each '%' does nothing.  But I've changed code anyway, in case Qt ever fix the issue.

PR #445, opened 2015-01-15
Physics fixes

scrawl:


apps/openmw/mwworld/scene.cpp
```diff
@@ -21,6 +21,18 @@
...
 namespace
 {
+
+    void addObject(const MWWorld::Ptr& ptr, MWWorld::PhysicsSystem& physics,
+                   MWRender::RenderingManager& rendering)
+    {
+        std::string model = Misc::ResourceHelpers::correctActorModelPath(ptr.getClass().getModel(ptr));
+        std::string id = ptr.getClass().getId(ptr);
+        if (id == "prisonmarker" || id == "divinemarker" || id == "templemarker" || id == "northmarker")
+            model = "";
```
Mingun: I think, the comment is necessary, why it becomes

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -2107,6 +2107,9 @@ namespace MWWorld
         Ogre::Vector3 playerPos(refdata.getPosition().pos);
 
         const OEngine::Physic::PhysicActor *physactor = mPhysEngine->getCharacter(refdata.getHandle());
+        if (!physactor)
+            return 0; // shouldn't happen
```
Mingun: May be also add `assert`?

apps/openmw/mwworld/scene.cpp
scrawl: Added comment.

PR #446, opened 2015-01-17
OpenCS shows plug-ins with load order issues.

dteviot:
1. FileDiaog in OpenCS now shows warning icon beside .esm./.esp files with load order problems.
2. omwlaucher -> replaced "stop" icon with "warning" icon for files with load order problems.

### DIFF
--- a/libs/openengine/ogre/renderer.hpp
+++ b/libs/openengine/ogre/renderer.hpp
@@ -6,6 +6,7 @@
  */
 
 #include <string>
+#include <stdint.h>
 
 #include <OgreTexture.h>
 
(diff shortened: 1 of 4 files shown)


libs/openengine/ogre/renderer.hpp
```diff
@@ -6,6 +6,7 @@
  */
 
 #include <string>
+#include <stdint.h>
```
Mingun: Better use `<cstdint>` for C++. This include for C.

zinnschlag:
Normally you would be right, but we are using stdint.h exclusively. There have been some compatibility problems on some platforms with the c-variant.

However with these changes OpenCS segfaults for me: https://gist.github.com/zinnschlag/3f8ddf5aa53791a48121

dteviot:
I believe I've fixed the segfault you experienced.
My apologies for the inconvenience.

PR #447, opened 2015-01-19
Essimporter

scrawl:
Largely incomplete, most state is still dropped, but this should be a solid base to work off of.
You can try it out on your save files, if there's one it doesn't load (i.e. exits with an error) I want to know about it.


psi29a:
Travis failed hard on Linux, have a look please. :)

scrawl:
> Travis failed hard on Linux, have a look please. :)

Fixed.

apps/essimporter/converter.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+            for (std::vector<CellRef>::const_iterator refIt = it->second.mRefs.begin(); refIt != it->second.mRefs.end(); ++refIt)
+            {
+                const CellRef& cellref = *refIt;
+                ESM::CellRef out;
+                out.blank();
+
+                if (cellref.mIndexedRefId.size() < 8)
+                {
+                    std::cerr << "CellRef with no index?" << std::endl;
```
Mingun: I think, you need output, what exactly CellRef have no index

scrawl: This message is no longer relevant as of now (code I haven't pushed yet, part of a large WIP commit), so I will not fix it.
In short, CellRefs with no index are actually very common.

PR #448, opened 2015-01-20
Fix broken windows builds due to README change

ace13:


scrawl:
Woops. I had no idea we were installing the readme.

PR #450, opened 2015-01-20
missing include in essimporter/main.cpp

maqifrnswa:
error: 'cout' is not a member of 'std'
error: 'cerr' is not a member of 'std'
https://launchpadlibrarian.net/195402684/buildlog_ubuntu-trusty-amd64.openmw_0.34.0%2Bgit20150120.442-0~ubuntu14.04.1_FAILEDTOBUILD.txt.gz

### DIFF
--- a/apps/essimporter/main.cpp
+++ b/apps/essimporter/main.cpp
@@ -1,3 +1,5 @@
+#include <iostream>
+
 #include <boost/program_options.hpp>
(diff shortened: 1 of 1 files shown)


scrawl:
Thank you! I wonder what's implicitely pulling in iostream on my machine. Maybe a different boost version.

PR #452, opened 2015-01-23
Fix build when using new boost with Ogre.

ace13:
This fixes a [Parse error at "BOOST_JOIN"] if you're using - for instance - Boost 1.57.0

### DIFF
--- a/apps/launcher/graphicspage.cpp
+++ b/apps/launcher/graphicspage.cpp
@@ -12,6 +12,9 @@
 
 #include <SDL.h>
 
+#include <OgreRoot.h>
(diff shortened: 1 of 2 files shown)


scrawl:
What if we forward declare the types instead and include in the .cpp only? That would fix the parse error too, right?

ace13:
Indeed it should, moc only runs on the .hpp's as far as I know.

scrawl:
Ok, then I would prefer that solution to keep the number of #ifs to a minimum.

ace13:
Something like this then?

PR #453, opened 2015-01-24
import script variables

scrawl:
This may be controversial, as it can not use the safeguards for changed variable names. The variable names are unfortunately not available in the save file.


scrawl:
Forgot to add a file again, fixup coming...

scrawl:
Ok, done.

PR #456, opened 2015-01-26
Dialog fix for line breaks in the middle of words.  (Bug #1049)

jordan-ayers:
Adjust the TypesetBookImpl::Typesetter to not assume a word break at the end of write() calls.
A word break is assumed if any of the other content insertion methods are used (section break, add/select content, etc).

I don't think this should have any effect on other text handling.  I did review some of the journal and a few books as a quick check for unintended side-effects, and saw no changes.

I added notes for a test case in the bug tracker ticket that can be reached quickly from a new game:
https://bugs.openmw.org/issues/1049


scrawl:
Works for me.

PR #464, opened 2015-02-03
Tweaks

Lazaroth:
Some tweaks to water, default sound level and mygui layouts. I've written more detailed info under each commit.


zinnschlag:
Agreed.

Lazaroth:
I thought that might be the case, which is why I just did it on the quick ones. I'll remove the bars and leave it at the buttons tweaks.

Lazaroth:
There. Just some pixels here and there on the buttons now, to make it line up better. Let me know if there's anything else.

scrawl:
Ok, looks good now.

PR #466, opened 2015-02-03
MIT License header added to all terrain component files.

jacmoe:
;)

### DIFF
--- a/components/terrain/buffercache.cpp
+++ b/components/terrain/buffercache.cpp
@@ -1,3 +1,24 @@
+/*
+ * Copyright (c) 2015 scrawl <scrawl@baseoftrash.de>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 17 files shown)


scrawl:
Was going to do this myself, but I'm still debating whether to use license headers or just a COPYING file in the relevant directory.
By the way, the license also applies to the terrain shader file found in files/materials/terrain.shader (even though the terrain component is technically usable without it, using a fixed function material).

jacmoe:
The reason why I added it to the individual files is that they are probably going to live next to other files, and/or modified. I issued a pull request just in case you were going to do the same. :)

jacmoe:
To clarify: I usually add license headers to all files because it makes it ubiquitously clear what license they bear and to whom they are attributed. Especially when they might end up the strangest of places. :)

scrawl:
https://github.com/OpenMW/openmw/commit/ee574e08ef48c2496ad522c9fa8aecd5b218dccc

jacmoe:
Good man! :)

2015-02-12 2:51 GMT+01:00 scrawl <notifications@github.com>:

PR #468, opened 2015-02-04
temp fix for Android

sandstranger:


scrawl:
Not acceptable. Instead we need to find out why the delete function doesn't work.

sandstranger:
This is madness , but openmw crashed if declared more than one bools types in refdata.hpp

scrawl:
Sounds like an alignment bug. Reordering the bools might work around it, i.e. declare the 4 bools after one another in the header.

sandstranger:
Please not  merge this commit  to the repository,I found more bugs with bools, which returns wrong values..

sandstranger:
Please merge this commit to the repository,  I will deal with graphics later . 
Now I need start refactor Java code to make it work without bugs

apps/openmw/mwmechanics/creaturestats.hpp
```diff
@@ -33,6 +33,23 @@ namespace MWMechanics
         MagicEffects mMagicEffects;
         Stat<int> mAiSettings[4];
         AiSequence mAiSequence;
+#ifdef ANDROID
```
Mingun: The order of fields is important? Why not simple reorganize them as you made for Android, for all platforms and delete `#ifdef`?

apps/openmw/mwworld/refdata.hpp
```diff
@@ -33,12 +33,17 @@ namespace MWWorld
             MWScript::Locals mLocals; // if we find the overhead of heaving a locals
                                       // object in the refdata of refs without a script,
                                       // we can make this a pointer later.
+#ifdef ANDROID
```
Mingun: Also as above

apps/openmw/mwmechanics/creaturestats.hpp
```diff
@@ -33,19 +33,22 @@ namespace MWMechanics
...
+        mAlarmed,
+        mAttacked,
+        mAttackingOrSpell,
+        mKnockdown,
+        mKnockdownOneFrame,
+        mKnockdownOverOneFrame,
+        mHitRecovery,
+        mBlock,
+        mRecalcMagicka;
         int mFriendlyHits;
```
zinnschlag: Can we please not do that? One separate declaration per variable (as we had before).

sandstranger:
It is done

zinnschlag:
Sorry, but that does not look good. These two pieces of code should be equivalent. Are we actually talking about a compiler bug here? I think further investigations are required regarding why the clean version of this piece of code causes problems on Android.

sandstranger:
Ok ,I will try to deal with it , perhaps the problem was in the flags gcc, which I used in the building

sandstranger:
game crashed here due to use gcc flags: -O3 ,-O2 or -Ofast

sandstranger:
Even without -O2, -O3, -Ofast game crashes after start loading game .
but if change mDeleted like this :
            bool mDeleted; // separate delete flag used for deletion by a content file
            bool mHasLocals;
            bool mEnabled;
            int mCount; // 0: deleted
game stop crashes

zinnschlag:
Closed or ready to merge?

sandstranger:
ready to merge I think

kcat:
Did you ever figure out what's wrong? As far as I can see, the fixes are still just reordering some fields. Since that makes no sense, hiding the problem with random changes feels like the wrong thing to do if we want it stable. Plus, these 'add ifdef'/'delete ifdefs'  commits make the log a bit messy since they're just reverting what the patchset just did. Would it be possible to do a rebase without those commits?

sandstranger:
Ok , than I will delete it . Problem with crashing game connected with mDeleted . How to fix it with another way I do not know yet

PR #469, opened 2015-02-05
Fix: Importer creates a single Content List (Fixes #2345)

dteviot:


apps/launcher/datafilespage.hpp
```diff
@@ -58,6 +58,10 @@ namespace Launcher
         void on_newProfileAction_triggered();
         void on_deleteProfileAction_triggered();
 
+    public:
+        /// Content List that is aways present
```
scrawl: typo: always

PR #473, opened 2015-02-07
Add training progress bar, implement jail screen, tweak wait dialog (Fixes #1714)

MiroslavR:


apps/openmw/mwgui/jailscreen.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "../mwworld/esmstore.hpp"
+#include "../mwworld/store.hpp"
+#include "../mwworld/class.hpp"
+
+#include "jailscreen.hpp"
+
+namespace MWGui
+{
+    JailScreen::JailScreen()
+        : WindowBase("openmw_loading_screen.layout"),
```
scrawl: We need a separate layout here, I think. This one would be slightly off-center, because the loading progress bar is at the bottom of a 300x300 screen.

MiroslavR:
Should be alright now.

scrawl:
The speed of the jail progress bar seems a little off. Went to jail with a 4900 bounty. The bar took 1 second to complete in MW, ~15 seconds in openmw.

MiroslavR:
No idea. The jail progress time in vanilla seems inconsistent. Can you try (in vanilla):
```
coc balmora
setpccrimelevel 4900
gotojail (this takes ~2 seconds for me)
setpccrimelevel 4900
gotojail (this takes longer, I could record 5 seconds up to 18 seconds)
```
Running on wine 1.7.35 CSMT @ 60 fps.

scrawl:
Right, I get a similar result. Going to jail takes longer if you were already near the jail.  Going to jail from a further away cell is faster. I think the latter is the more likely case, so maybe we should calibrate on that one.
It might just be FPS-dependent as well. The waiting/resting progress bar definitely is.

MiroslavR:
Still the results are completely inconsistent and seem to be FPS-dependent. Honestly, I don't know what more I can do with this PR. To find out what the devs were aiming for would likely require reverse-engineering the game, which is beyond my capabilities.

PR #474, opened 2015-02-07
Bugfix/781 - Change sun trajectory

jordan-ayers:
This moves the sun to where OMW #781 says it should be.

I did not try to verify the accuracy of the 15 degrees number, but it seemed close.

I also re-named an argument variable because 'is_moon' seemed very misleading.

### DIFF
--- a/apps/openmw/mwworld/weather.cpp
+++ b/apps/openmw/mwworld/weather.cpp
@@ -443,13 +443,13 @@ void WeatherManager::update(float duration, bool paused)
 
     int facing = (mHour > 13.f) ? 1 : -1;
 
-    bool sun_is_moon = mHour >= mNightStart || mHour <= mSunriseTime;
+    bool is_night = mHour >= mNightStart || mHour <= mSunriseTime;
 
     Vector3 final(
             (height - 1) * facing,
             (height - 1) * facing,
             height);
-    mRendering->setSunDirection(final, sun_is_moon);
+    mRendering->setSunDirection(final, is_night);
 
     /*
      * TODO: import separated fadeInStart/Finish, fadeOutStart/Finish
--- a/apps/openmw/mwworld/weather.cpp
+++ b/apps/openmw/mwworld/weather.cpp
@@ -429,30 +429,32 @@ void WeatherManager::update(float duration, bool paused)
     else
         mRendering->getSkyManager()->sunEnable();
 
-    // Update the sun direction.  Run it east to west at 15 degrees south from overhead.
-    // The sun's speed at day and night will differ, since mSunriseTime and mNightStart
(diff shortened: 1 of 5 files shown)


apps/openmw/mwworld/weather.cpp
```diff
@@ -427,29 +429,33 @@ void WeatherManager::update(float duration, bool paused)
...
-    // sun angle
-    float height;
-
-    //Day duration
-    float dayDuration = (mNightStart - 1) - mSunriseTime;
+    // Update the sun direction.  Run it east to west at 15 degrees south from overhead.
+    // The sun's speed at day and night will differ, since mSunriseTime and mNightStart
+    // mark when the sun is level with the horizon.
+    {
+        assert( mNightStart > mSunriseTime );
```
EmperorArthur: You might want to consider changing this to an if/throw block rather than an assert.

jordan-ayers: I put the assert in to highlight that the code isn't trying to handle weird edge cases (like midnight being during the day, or sunrise/sunset being set to the same times).  If these assumptions are wrong, it's probably a soft rendering error, mapping the sun to the wrong spot.  If a throw is used, then the visual bug becomes a fatal error, and that seemed excessive.  Shouldn't we let the users play, if safe to do so?

There were a few parts I didn't dig into as much, since they had no visual effect:
Why is a visible sun trajectory calculated for nighttime, but hidden by the 'is_night' flag?  Why wasn't the sun just turned off like the moons are?  I think there were some comments relevant to this in the renderer sky.cpp file, but I didn't study that part of the system well enough to propose any simplifications.  The sun's direction could have been set to take the sun under the world plane instead, but since it seems to be turned off at night, I didn't change that part of the code.

zinnschlag:
I am not entirely sure if that is still the case, but we used to leave asserts in for release builds. Anyway, we use asserts very sparsely. Generally only in situations where continuing is absolutely not possible. In all other cases we throw exceptions and then handle them in an appropriate place.
Here however the code location is in the middle of a larger update function with lots of independent stuff before and after it. I suggest to check the condition and if the check fails write the error to cerr and skip.

jordan-ayers:
OK.  Error messages over asserts for softer failures.  Let me take another look and see if I can cover the other cases without bloating it out too much as well.  (Though the moon handling below looks like it also assumes the 'normal' case, so it may still need the warning even if the Sun code is complete.)

jordan-ayers:
That should cover all sunrise / sunset time combinations, assuming the values are in the [0-24) range.

The moon calculations make some similar assumptions to the ones I just changed, but that should be covered when OMW Bug #672 is looked at.  I don't think we've got enough details to put in correct moon handling yet.

PR #475, opened 2015-02-09
ini importer sorts content files by date modified (Fixes #2346)

dteviot:
Note, this is a partial fix.
It should work correctly for any OS that uses UTF-8 strings for filenames in OS calls.
It won't work correctly for content files that use non-ASCII characters on Windows.
But I think 
(a) the number of files falling into that case will be small.
(b) most people are only going to run the importer once.
(c) The fix is better than what we had.
(d) If someone wants to fix the problem fully, this gives them a starting point.

### DIFF
--- a/apps/launcher/maindialog.cpp
+++ b/apps/launcher/maindialog.cpp
@@ -257,24 +257,8 @@ void Launcher::MainDialog::changePage(QListWidgetItem *current, QListWidgetItem 
         current = previous;
 
     int currentIndex = iconWidget->row(current);
-//    int previousIndex = iconWidget->row(previous);
-
     pagesWidget->setCurrentIndex(currentIndex);
-
-    //    DataFilesPage *previousPage = dynamic_cast<DataFilesPage *>(pagesWidget->widget(previousIndex));
-    //    DataFilesPage *currentPage = dynamic_cast<DataFilesPage *>(pagesWidget->widget(currentIndex));
-
-    //    //special call to update/save data files page list view when it's displayed/hidden.
-    //    if (previousPage)
-    //    {
-    //        if (previousPage->objectName() == "DataFilesPage")
-    //            previousPage->saveSettings();
-    //    }
-    //    else if (currentPage)
-    //    {
-    //        if (currentPage->objectName() == "DataFilesPage")
-    //            currentPage->loadSettings();
-    //    }
+    mSettingsPage->resetProgressBar();
 }
 
 bool Launcher::MainDialog::setupLauncherSettings()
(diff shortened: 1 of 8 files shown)


scrawl:
As far as I know, the utf8 path should *just work*, because we're configuring boost filesystem to use it:  https://github.com/OpenMW/openmw/blob/master/components/files/windowspath.cpp#L37

There is a build error, see the travis log (need to use > > instead of >> for the nested template)

scrawl:
Nevermind the previous comment. It seems the mwiinimporter isn't using the ConfigurationManager, so the utf8 path setup isn't called. You should be able to copy it though.

dteviot:
The relevant code to imbue the boost::filesystem is already present.
https://github.com/OpenMW/openmw/blob/master/apps/mwiniimporter/main.cpp#L57

scrawl:
Well, then the "fixme: may need to use codecvt_type" comment can be removed, right?

scrawl:
I tried this branch merged into master. Content file importer isn't working for me. This is what I have in Morrowind.ini:

<pre>
[Game Files]
GameFile0=Tribunal.esm
GameFile1=Bloodmoon.esm
GameFile2=Morrowind.esm
GameFile3=test.esp
</pre>

I start with this launcher config:
<pre>
[Profiles]
currentprofile=Default
Default/content=Morrowind.esm
Default/content=Tribunal.esm
Default/content=Bloodmoon.esm
</pre>

After importing into "test" I then have the following launcher config:
<pre>
[Profiles]
test/content=Tribunal.esm
test/content=Bloodmoon.esm
test/content=Morrowind.esm
test/content=Blank_ESM_2.0.esm
test/content=test.esp
currentprofile=test
Default/content=Morrowind.esm
Default/content=Tribunal.esm
Default/content=Bloodmoon.esm
</pre>

I have no idea where "Blank_ESM_2.0.esm" is coming from. I didn't have that in any previous profile nor in Morrowind.ini.

Also interesting, there is a timestamped, empty, non-selected profile in the launcher immediately after importing. It disappears when restarting the launcher.

Also I once had the problem of a profile I had already deleted through the launcher UI showing up again, but haven't found how to reproduce that yet.

zinnschlag:
Continuing or PM discussion here.

First, please don't do the fake progress bar. Use the bounce-version.

Second, I am getting 100% progress here at the end (not sure why that doesn't work for you).

Third, the popup window asking for the content file list name is not okay. As I said before we should avoid pop ups. The name does not actually need to come from the user. Create a suitable default name (after checking that there isn't already a matching content list) and make it the active content list. The user can always rename it, if he does not like the original name.

PR #477, opened 2015-02-09
Enable link-time optimization on MSVC

ace13:
Enabling link-time optimization globally will increase the build time of release builds by another couple of minutes, but since it results in free size as well as speed benefits I'm okay with that.

Only sour part is that this has completely negated all my work in speeding up the build times on Windows. :disappointed:


ace13:
This will fix #2349 by the way.

ace13:
I'll just close this PR since the branch was merged into master.

PR #478, opened 2015-02-10
Fix MSVC build due to missing M_PI define

ace13:
Apparently the header guard on cmath only adds the defines if _USE_MATH_DEFINES is defined when cmath is included for the first time.

So enabling that define and moving cmath up to be the first include, so that nothing else includes it without the define set.

### DIFF
--- a/apps/openmw/mwworld/weather.cpp
+++ b/apps/openmw/mwworld/weather.cpp
@@ -1,6 +1,7 @@
-#include "weather.hpp"
-
+#define _USE_MATH_DEFINES
 #include <cmath>
+
+#include "weather.hpp"
 
 #include <components/esm/weatherstate.hpp>
 


psi29a:
I gave osx a swift kick...

psi29a:
<pre>
==> Downloading https://downloads.sf.net/project/machomebrew/Bottles/qt-4.8.6.ma
curl: (22) The requested URL returned error: 404 Not Found
Error: Failed to download resource "qt"
Download failed: https://downloads.sf.net/project/machomebrew/Bottles/qt-4.8.6.mavericks.bottle.6.tar.gz
Warning: Bottle installation failed: building from source.
==> Downloading http://qtmirror.ics.com/pub/qtproject/official_releases/qt/4.8/4
######################################################################## 100.0%
==> Downloading https://raw.githubusercontent.com/DomT4/scripts/440e3cafde5bf6ec
######################################################################## 100.0%
</pre>
Would be nice to bottle this again...

ace13:
I like that it actually shows that it wasn't a build error that failed, rather the pre-build. (The blue checkmark instead of a red one)

I should take another look into appveyor for OpenMW I think, though I'd have to package up all the build deps in some way for it to build in a reasonable time.

corristo:
@psi29a I remember about that, just had no chance to fix this.

psi29a:
@ace13 we can ask @lgromanowski to host those deps if/when you are ready. :)

ace13:
It's not the hosting of them that's the problem, I could take care of that myself.

What's going to be fun is writing up the cmake commandline argument needed to set up a windows build environment from scratch. :grimacing:

corristo:
@psi29a FYI, I moved Qt binaries to our server: https://github.com/OpenMW/openmw/pull/480

psi29a:
^5 
Good work.

PR #481, opened 2015-02-12
Remove setting CMAKE_BUILD_TYPE to Debug

slothlife:
Setting was causing single-target configurations (ninja, make) to
incorrectly link vs debug runtimes on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,12 @@
 project(OpenMW)
 
+# If the user doesn't supply a CMAKE_BUILD_TYPE via command line, choose one for them.
+IF(NOT CMAKE_BUILD_TYPE)
+    SET(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
+        "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel."
+        FORCE)
+ENDIF()
+
 if (APPLE)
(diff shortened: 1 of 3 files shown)


scrawl:
Great find. I really don't think this was meant to be committed in the first place. Blame shows commit f6203cac2ab29712a.

kcat:
Though I would recommend setting it to RelWithDebInfo if it's not set, i.e.
```
IF(NOT CMAKE_BUILD_TYPE)
    SET(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
        "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel."
        FORCE)
ENDIF()
```
CMake has a habit of not setting a build type if one isn't explicitly set on the command line, so it defaults to having no optimization or debug flags, which is a pretty useless build.

slothlife:
Perhaps the top level CMakelists.txt is a good candidate for that kcat. I always explicitly set it via batch scripts when I build, so I haven't run into that.

slothlife:
Sorry, but are you waiting on me to make the change in this Pull Request?

zinnschlag:
kcat's remark is worth considering IMO.

psi29a:
@slothlife Please implement kcat solution, then we can merge this in.

slothlife:
Sorry for the delay. My free time wasn't as free as I had hoped.

PR #482, opened 2015-02-13
Implement OpenCS reference validation (Feature #831)

NeveHanter:


zinnschlag:
Mostly okay, but I am not so sure about the charge check. I am getting a bunch of warning for Morrowind.esm. Isn't -1 valid in some cases? Also, this is a union and to verify this field we would need to know which type is in use. I don't think we can do that without additional context, which may mean we can't verify the charge at all.

@scrawl Any additional input?

scrawl:
-1 means "full charge" and is perfectly valid.

NeveHanter:
I've changed code to check for reference type and then decide whether check charge's int or float value. If I understood correctly float one is durability and can't be negative, and int one is item charges, which can be -1 if at full charge, please correct me if I'm wrong.

apps/opencs/model/tools/referencecheck.cpp
```diff
@@ -0,0 +1,117 @@
...
+        messages.push_back(std::make_pair(id, " is an empty reference"));
+    } else {
+        // Check for non existing referenced object
+        if (mReferencables.searchId(cellRef.mRefID) == -1) {
+            messages.push_back(std::make_pair(id, " is referencing non existing object " + cellRef.mRefID));
+        } else {
+            // Check if reference charge is valid for it's proper referenced type
+            CSMWorld::RefIdData::LocalIndex localIndex = mDataSet.searchId(cellRef.mRefID);
+            if (localIndex.second == CSMWorld::UniversalId::Type_Armor ||
+                localIndex.second == CSMWorld::UniversalId::Type_Weapon) {
```
scrawl: ChargeFloat is used for lights. ChargeInt for everything else.

NeveHanter:
Fixed as in @scrawl notes

PR #483, opened 2015-02-14
Bugfix/1533 - Teleportation in Mournhold

jordan-ayers:
This is a brute-force search to support intervention / prison behavior while in Mournhold.

Any feedback on if there are serious performance issues with the cell-loading used would be appreciated.  I'm not familiar enough with the cell loading / unloading to know if any caching / special handling of the needed data is appropriate.  I'd like to find a basically acceptable approach before cleaning it up the rest of the way.

A side effect of the changes is that the stolen_goods container list isn't needed anymore.  I found that the prison markers contain teleport links to the cell containing the evidence chest to use.

(The OMW Bug mentions edge cases for over-world marker handling, but that hasn't been changed here.)

### DIFF
--- a/apps/openmw/mwworld/cellstore.hpp
+++ b/apps/openmw/mwworld/cellstore.hpp
@@ -187,7 +187,7 @@ namespace MWWorld
             }
 
             template <class T>
-            CellRefList<T>& getReadOnly() {
+            const CellRefList<T>& getReadOnly() {
                 throw std::runtime_error ("Read Only CellRefList access not available for type " + std::string(typeid(T).name()) );
             }
 
@@ -365,7 +365,7 @@ namespace MWWorld
     }
 
     template<>
-    inline CellRefList<ESM::Door>& CellStore::getReadOnly<ESM::Door>()
+    inline const CellRefList<ESM::Door>& CellStore::getReadOnly<ESM::Door>()
     {
         return mDoors;
     }
--- a/apps/openmw/mwworld/cellstore.hpp
+++ b/apps/openmw/mwworld/cellstore.hpp
@@ -3,6 +3,8 @@
 
 #include <algorithm>
 #include <stdexcept>
+#include <string>
+#include <typeinfo>
 #include <boost/shared_ptr.hpp>
 
 #include "livecellref.hpp"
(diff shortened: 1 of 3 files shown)


zinnschlag:
I'll leave this PR open a bit longer to see if we get any additional input regarding this issue.

scrawl:
Regarding performance, it would be nice if the search algorithm wouldn't trigger "mHasState" in CellStore to be set true, or alternatively reset it to false after it's done searching. That will at least prevent the cells from bloating the savegame.

jordan-ayers:
I'll check (probably tomorrow) to see if I can avoid affecting mHasState.  The search routine previously used for finding the stolen goods container references was probably not affecting this.

jordan-ayers:
OK.  It's not altering mHasState anymore (though it doesn't seem to change the size of my save files at all).

I added a 'getReadOnly()' variant for fetching the Door list.  There's no enforcement in the class that the caller won't alter the refs, but it seemed like the cleanest way to mark that the door search was not making changes.

I also switched to the built-in search routine for the marker lookup, since it preserves mHasState when nothing is found, and most of the code doesn't actually care which kind of world object the marker is.  If for some reason the prison marker's not a door, confiscate items will log a complaint, but not take any items.

jordan-ayers:
The cellstore->search change hits one of the cleanup items I was looking at.

I'm not aware of a better way to handle the door link searching at the moment, so I think it's ready to merge.

apps/openmw/mwworld/cellstore.hpp
```diff
@@ -184,6 +184,11 @@ namespace MWWorld
                 throw std::runtime_error ("Storage for this type not exist in cells");
             }
 
+            template <class T>
+            CellRefList<T>& getReadOnly() {
+                throw std::runtime_error ("Read Only access not available for this type");
```
Mingun: Please, instead `this type` in message use `typeid(T).name()` -- this is better for debugging.

apps/openmw/mwworld/cellstore.hpp
```diff
@@ -357,6 +362,12 @@ namespace MWWorld
         return mWeapons;
     }
 
+    template<>
+    inline CellRefList<ESM::Door>& CellStore::getReadOnly<ESM::Door>()
```
scrawl: Why not make this function return a const reference? Am I missing something here?

jordan-ayers: No, that makes more sense.  I adjusted the functions around the older code's door loop, and forgot to change that.

jordan-ayers:
Improvements to the error text and the const handling are now in place.

PR #485, opened 2015-02-15
CI: fix `make package` condition by adding missing spaces

corristo:


corristo:
`make package` now performed successfully.

PR #486, opened 2015-02-16
Fix various MSVC warnings

slothlife:


slothlife:
Is there any further action required on my part? All I've done here is basically added to the global list of ignored warnings (which appear to be already ignored on Linux), and added the ignored warning flags to a few of the internal static libraries which were missing it. I also added a missing virtual destructor.

scrawl:
Most likely this just fell through the cracks. @zinnschlag?

zinnschlag:
Sorry. Will merge.

PR #490, opened 2015-02-17
Telekinesis allows safe opening of traps (Fixes #1916)

dteviot:
When trap activated at beyond normal activation distance, assume telekinesis used and detonate trap at trapped object's location.
Also some minor code refactoring of spellcasting.
1. Corrected parameter passed to explodeSpell().
2. For loop now correctly does an early exit.
3. Removed duplicated tests.

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -269,6 +269,8 @@ namespace MWBase
             virtual MWWorld::Ptr  getFacedObject() = 0;
             ///< Return pointer to the object the player is looking at, if it is within activation range
 
+            virtual float getMaxActivationDistance() = 0;
+
             /// Returns a pointer to the object the provided object would hit (if within the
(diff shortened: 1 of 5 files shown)


apps/openmw/mwbase/world.hpp
```diff
@@ -269,6 +269,8 @@ namespace MWBase
             virtual MWWorld::Ptr  getFacedObject() = 0;
             ///< Return pointer to the object the player is looking at, if it is within activation range
 
+            virtual float getMaxActivationDistance() = 0;
```
scrawl: The implementation for this function doesn't seem to be included in this pull request. Forgot to add it?

dteviot:
> The implementation for this function doesn't seem to be included in this pull request. Forgot to add it?

No.  The implementation of the function is already present in the MWWorld::World class.
Although if you've got a better way to expose it via a MWBase::World pointer I'd like to know it.

And proof it works should be the Travis CI build passing.

PR #491, opened 2015-02-20
Silenced clang warning by converting int to string

naclander:
Clang was giving a warning, so I used `boost::lexical_cast` to explictly convert the values. I would have used `string::to_string()`, but we don't support C++11 yet.


zinnschlag:
Both DestCell and Faction are strings. I don't see why you would get a warning there. Good catch on the FactionRank though.

naclander:
Is there anything preventing this change from being pulled?

zinnschlag:
If you make a change to an existing pull request, please always add a comment. Otherwise it won't show up in my newsfeed.
Anyway, the PR is now queued up for merging. I will get to it in a bit.

PR #496, opened 2015-02-25
bugfix/2400 - Correction to teleportation changes.

jordan-ayers:
Related to OMW Bug #1533

I missed the check against the starting cell being an exterior when I implemented getClosestMarker for the teleportation / map changes.  This adds the check back in, and fixes the game crash on teleport / jail while in an exterior cell.  (Apparently my testing was just with coc and Mournhold interiors at the end of that change sequence.)

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -2832,6 +2832,10 @@ namespace MWWorld
 
     MWWorld::Ptr World::getClosestMarker( const MWWorld::Ptr &ptr, const std::string &id )
     {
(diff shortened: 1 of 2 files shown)


jordan-ayers:
Note:  This fixes the outside case.  Bug 2400 also reports problems with 'go to jail' in cells with no path to a jail.  That issue is not yet fixed in this patch.  I'll try to add a fix for that soon, but this patch corrects the majority of the symptoms from the problem.

jordan-ayers:
Updated to fix the 'toddtest' issue as well.

Jail will now behave is in vanilla Morrowind, with no confiscation or teleportation if a prison marker cannot be found from the current cell.

PR #502, opened 2015-02-28
list in launcher shows addon files with no dependencies (Fixes #2348)

dteviot:
Also, game files must end with ".esm" or ".omwgame".


components/contentselector/model/esmfile.cpp
```diff
@@ -62,6 +62,12 @@ QByteArray ContentSelectorModel::EsmFile::encodedData() const
     return encodedData;
 }
 
+bool ContentSelectorModel::EsmFile::isGameFile() const
+{ 
+    return (mGameFiles.size() == 0) &&
+        (mFileName.endsWith(QLatin1String(".esm")) || mFileName.endsWith(QLatin1String(".omwgame")));
```
scrawl: This check should be case insensitive.

PR #503, opened 2015-02-28
fixes shaders for glsles2

sandstranger:


sandstranger:
argh This fix bad , Sh_Glsles will never work

PR #504, opened 2015-02-28
Limit maximum attribute damage (Fixes #2367)

dteviot:
Maximum damage that an attribute can have = base + fortify.

### DIFF
--- a/apps/openmw/mwmechanics/activespells.cpp
+++ b/apps/openmw/mwmechanics/activespells.cpp
@@ -152,12 +152,7 @@ namespace MWMechanics
     void ActiveSpells::addSpell(const std::string &id, bool stack, std::vector<ActiveEffect> effects,
                                 const std::string &displayName, int casterActorId)
     {
-        bool exists = false;
-        for (TContainer::const_iterator it = begin(); it != end(); ++it)
-        {
-            if (id == it->first)
-                exists = true;
-        }
+        TContainer::iterator it(mSpells.find(id));
 
         ActiveSpellParams params;
         params.mTimeStamp = MWBase::Environment::get().getWorld()->getTimeStamp();
@@ -165,12 +160,42 @@ namespace MWMechanics
         params.mDisplayName = displayName;
         params.mCasterActorId = casterActorId;
 
-        if (!exists || stack)
-            mSpells.insert (std::make_pair(id, params));
+        if (it == end() || stack)
+        {
+            mSpells.insert(std::make_pair(id, params));
+        }
         else
-            mSpells.find(id)->second = params;
+        {
+            // addSpell() is called with effects for a range.
+            // but a spell may have effects with different ranges (e.g. Touch & Target)
+            // so, if we see new effects for same spell assume additional 
+            // spell effects and add to existing effects of spell
+            mergeEffects(params.mEffects, it->second.mEffects);
+            it->second = params;
+        }
 
         mSpellsChanged = true;
+    }
+
+    void ActiveSpells::mergeEffects(std::vector<ActiveEffect>& addTo, const std::vector<ActiveEffect>& from)
+    {
+        for (std::vector<ActiveEffect>::const_iterator effect(from.begin()); effect != from.end(); ++effect)
+        {
+            // if effect is not in addTo, add it
+            bool missing = true;
+            for (std::vector<ActiveEffect>::const_iterator iter(addTo.begin()); iter != addTo.end(); ++iter)
+            {
+                if (effect->mEffectId == iter->mEffectId)
+                {
+                    missing = false;
+                    break;
+                }
+            }
+            if (missing)
+            {
+                addTo.push_back(*effect);
+            }
+        }
     }
 
     void ActiveSpells::removeEffects(const std::string &id)
--- a/apps/openmw/mwmechanics/stat.hpp
+++ b/apps/openmw/mwmechanics/stat.hpp
@@ -235,26 +235,29 @@ namespace MWMechanics
     class AttributeValue
     {
         int mBase;
-        int mModifier;
+        int mFortified;
+        int mModifier; // net effect of Fortified, Drain & Absorb
         float mDamage; // needs to be float to allow continuous damage
 
     public:
-        AttributeValue() : mBase(0), mModifier(0), mDamage(0) {}
+        AttributeValue() : mBase(0), mFortified(0), mModifier(0), mDamage(0) {}
 
         int getModified() const { return std::max(0, mBase - (int) mDamage + mModifier); }
         int getBase() const { return mBase; }
         int getModifier() const {  return mModifier; }
 
         void setBase(int base) { mBase = std::max(0, base); }
-        void setModifier(int mod) { mModifier = mod; }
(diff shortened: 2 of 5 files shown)


apps/openmw/mwmechanics/activespells.cpp
```diff
@@ -165,14 +160,44 @@ namespace MWMechanics
...
+        }
         else
-            mSpells.find(id)->second = params;
+        {
+            // addSpell() is called with effects for a range.
+            // but a spell may have effects with different ranges (e.g. Touch & Target)
+            // so, if we see new effects for same spell assume additional 
+            // spell effects and add to existing effects of spell
+            mergeEffects(params.mEffects, it->second.mEffects);
+            it->second = params;
```
scrawl: This code is meant to fix https://bugs.openmw.org/issues/2285 , right?

apps/openmw/mwmechanics/stat.hpp
```diff
@@ -235,26 +235,29 @@ namespace MWMechanics
...
         int getModified() const { return std::max(0, mBase - (int) mDamage + mModifier); }
         int getBase() const { return mBase; }
         int getModifier() const {  return mModifier; }
 
         void setBase(int base) { mBase = std::max(0, base); }
-        void setModifier(int mod) { mModifier = mod; }
+        void setModifiers(int fortify, int drain, int absorb);
 
-        void damage(float damage) { mDamage += damage; }
+        void damage(float damage) { mDamage = std::min(mDamage + damage, (float)(mBase + mFortified)); }
```
scrawl: Why only take the fortifier into account? Why not just mDamage = std::min(mDamage + damage, getModified()); ?

I've tested using "drain intelligence on self" and "damage intelligence on self" in vanilla MW and the drain effect is taken into account there, meaning you can't damage your intelligence further if it would cause the modified value to go below zero.

apps/openmw/mwmechanics/activespells.cpp
```diff
@@ -165,14 +160,44 @@ namespace MWMechanics
...
 
+    void ActiveSpells::mergeEffects(std::vector<ActiveEffect>& addTo, const std::vector<ActiveEffect>& from)
+    {
+        for (std::vector<ActiveEffect>::const_iterator effect(from.begin()); effect != from.end(); ++effect)
+        {
+            // if effect is not in addTo, add it
+            bool missing = true;
+            for (std::vector<ActiveEffect>::const_iterator iter(addTo.begin()); iter != addTo.end(); ++iter)
+            {
+                if (effect->mEffectId == iter->mEffectId)
```
scrawl: This would be slightly incorrect when there's skills/attributes involved. Try effect->mEffectId == iter->mEffectId && effect->mArg == iter->mArg.

dteviot:
@scrawl
>Why only take the fortifier into account? Why not just
>
>mDamage += std::min(damage, std::max(0.f, getModified())); ?

The implementation of getModified() is this:
```C++
    int getModified() const { return std::max(0, mBase - (int) mDamage + mModifier); }
```
which you will note is already clamping the value to a minimum of 0.
So I don't need to worry that modified value will go negative.
As for applying damage before drain and absorb, that's becuase drain and absorb are "over time" effects which end when the spell ends while Damage is a instant, permanent effect.
Thus, someone who's going to encounter a "damage" situation could do the following exploit.
1. Cast drain on their own attributes.
2. When hit with damage spells, the level of damage will be limited by the drain in effect.
3. When drain wears off, player has advantage of the reduced damage.

> This would be slightly incorrect when there's skills/attributes involved. Try effect->mEffectId == iter->mEffectId && effect->mArg == iter->mArg.

Good point.  This I will do.

scrawl:
Damage isn't instant in vanilla, it applies over time. OpenMW used to apply it instantly, but that was fixed. But it seems like there are still some leftovers of the old implementation. Let me inquire about that...

The exploit you describe is possible, but I wouldn't worry about it. Most of the time users wouldn't know in advance that they're going to be hit with a damage spell. Either way, we don't get to fix such MW exploits, at least not before scriptable game mechanics are implemented.

scrawl:
Continuing discussion for instant/over-time damage effects on https://github.com/OpenMW/openmw/commit/b8d5a9486ac04270034126df753fe7e69d198b41.

dteviot:
> The exploit you describe is possible, but I wouldn't worry about it. Most of the time users wouldn't know in advance that they're going to be hit with a damage spell

It's not just an intentional exploit.  The various bonewalkers cast a mix of damage and drain on strength and endurance.
That said, from a user's game play perspective there's negligible difference between
```C++
mDamage = std::min(mDamage + damage, getModified())
```
and
```C++
mDamage = std::min(mDamage + damage, mBase + mFortifed))
```
and as the first results in simpler code, it's the better choice.

dteviot:
@scrawl
On reflection I think I was wrong, this one has a bug.
```C++
mDamage = std::min(mDamage + damage, getModified())
```

Bug is as follows:
1. Say user has Endurance of 30.
2. Endurance takes 30 points damage, so net Endurance is 0. i.e. Base = 30 and Damage = 30.
3. User now takes a drain Endurance of 10 points. (Net is still 0, but is made of Base = 30, Modifier = -10 and Damage = 20.)
4. When drain wears off, new Endurance will go to 10.  (Base of 30, Damage of 20.)

That said, probably isn't going to be something most users would notice.

scrawl:
I don't follow. Why would the damage get reduced to 20 in that case? The damage is applied once at the beginning and then stays that way, right?

dteviot:
@scrawl 
> I don't follow. Why would the damage get reduced to 20 in that case? The damage is applied once at the beginning and then stays that way, right?

You are correct, I am in error.
However, I believe there is still a bug with 
```C++
mDamage += std::min(damage, std::max(0, mBase - (int) mDamage + mModifier))
```
Try this scenario.
1. Say user has Base Endurance of 30, + fortified by 100 points from spell(s).
2. User takes 130 points of endurance damage. User now has net endurance of 0.
3. Fortify spell wears off. User now has net -70 endurance.  (This will show as 0 endurance on UI.)
4. User casts restore endurance.  Unless does more than 70 points, will appear to do nothing. (Original bug.)

Conceptually, my argument is, damage can be applied to fortify, but when the fortify wears off, the damage applying to it must also be removed.  Hence, we need to track the fortify value.

scrawl:
I just tried your scenario in MW as follows:
* Start with a character that has 30 base strength.
* Cast "Adrenaline Rush" (fortifies strength by 50 pts for one minute).
* Damage your Strength by 80 pts using a self-made spell.
* Wait for the Adrenaline Rush to wear off.
* Attempt to cast "restore strength" spell. Result: no restoration on the first try.

So,
<pre> mDamage += std::min(damage, getModified());</pre>
 seems to be close to what MW is doing.

Please note eventually we should remove the mDamage variable all together and apply damage to the mModified stat instead. As pointed out on bug 2094, in MW you can also restore attributes that have been drained. It does not keep track of draining and damage separately. We should do the same eventually, but that requires a careful rewrite of the magic effect system.
So don't loose too much sleep over this decision, when the code is going to be rewritten the end anyway.

PR #505, opened 2015-03-01
Fix OpenCS crashing since commit 9d6145 by showing gamefiles if the cont...

cc9cii:
...ent selector was created from OpenCS.  See [this forum post] (https://forum.openmw.org/viewtopic.php?f=20&t=2771&p=31321#p31299) for the bug report.

Please review since the fix is ugly - a boolean is passed around, polluting the interfaces. Hopefully there's a cleaner solution than this.

### DIFF
--- a/apps/opencs/view/doc/filedialog.cpp
+++ b/apps/opencs/view/doc/filedialog.cpp
@@ -24,7 +24,7 @@ CSVDoc::FileDialog::FileDialog(QWidget *parent) :
     resize(400, 400);
 
     setObjectName ("FileDialog");
-    mSelector = new ContentSelectorView::ContentSelector (ui.contentSelectorWidget);
+    mSelector = new ContentSelectorView::ContentSelector (ui.contentSelectorWidget, true);
     mAdjusterWidget = new AdjusterWidget (this);
 }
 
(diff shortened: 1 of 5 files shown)


zinnschlag:
This is confusing. Why are we showing game files in the content selector? That is not supposed to happen.
Can you elaborate a bit more on this change?

cc9cii:
The content selector for gamefile looks like this:
![content selector] (http://i872.photobucket.com/albums/ab289/cc9cii/select_gamefile.jpg).

Then one can select further files:
select Tribunal.

However the current master branch of OpenCS does not progress to the second screen, instead results in an empty vector and back() results in SEGV.

dteviot:
This _IS_ my mistake.

This line was included in the pull by mistake.
file: "components/contentselector/model/contentmodel.cpp" line 115, where I replaced
[code]
return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsUserCheckable;
[/code]
with 
[code]
    return Qt::NoItemFlags;
[/code]
It was part of something else I was trying that didn't work and I failed to clean it up properly.
I believe if you restore the line
    return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsUserCheckable;
the problem should be fixed.

cc9cii:
Ok, thanks for your note.  I've removed the stuff to make opencs different to the launcher.

PR #506, opened 2015-03-01
Fix Bug #2402.  SKIL records should not have NAME subrecord.

cc9cii:


zinnschlag:
Okay for now, but we really need to rewrite ESMWriter/Reader (post 1.0). The current situation is just a mess.

scrawl:
From a quick look at the code MagicEffect records would have the same problem,

apps/opencs/model/doc/savingstages.hpp
```diff
@@ -103,8 +103,20 @@ namespace CSMDoc
...
+            // If there are more idcollection records that don't use NAME then a more
+            // generic solution may be required.  The conversion code was simply
+            // copied from esmwriter. There's room for improving speed a little bit
+            // here if it turns out to be an issue.
+            uint32_t name = mCollection.getRecord (stage).mModified.sRecordId;
+            mState.getWriter().startRecord (name);
+            std::string type;
+            for (int i=0; i<4; ++i)
+                /// \todo make endianess agnostic
+                type += reinterpret_cast<const char *> (&name)[i];
```
scrawl: This can be done simpler like this:

<pre>
if (name != ESM::REC_SKILL)
</pre>

scrawl:
Script records are also a problem. They have the name built into the SCHD struct, so an extra NAME record would cause load failure there too.

cc9cii:
Thanks, I'll submit an update.

PR #512, opened 2015-03-03
if cell doesn't exist, PositionCell and PlaceItemCell warn std::err but ...

maqifrnswa:
...still execute, bug #2407


scrawl:
I think the warning should be logged to the console in addition to stderr, like so:
<pre>
runtime.getContext().report(msg);
</pre>

maqifrnswa:
got it, see:
https://github.com/OpenMW/openmw/pull/515

PR #513, opened 2015-03-03
scoped Input

Antifleischwolf:
Not really sure what to write about a 2 line change but here is the link to the issue tracker:
https://bugs.openmw.org/issues/2409

**Edit**: I meant it fixes this part: "If you later write it into the console or save menu, the game won't crash, but it will switch your weapons even though you've selected a text field."

### DIFF
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -666,11 +666,9 @@ namespace MWInput
             SDL_StopTextInput();
 
         bool consumed = SDL_IsTextInputActive();
-        if (kc != OIS::KC_UNASSIGNED)
-        {
-            bool guiFocus = MyGUI::InputManager::getInstance().injectKeyPress(MyGUI::KeyCode::Enum(kc), 0);
-            setPlayerControlsEnabled(!guiFocus);
-        }
+        bool guiFocus = MyGUI::InputManager::getInstance().injectKeyPress(MyGUI::KeyCode::Enum(kc), 0);
+        setPlayerControlsEnabled(!guiFocus);
+
         if (!mControlsDisabled && !consumed)
             mInputBinder->keyPressed (arg);
         mJoystickLastUsed = false;
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -665,11 +665,9 @@ namespace MWInput
                 && MWBase::Environment::get().getWindowManager()->getMode() == MWGui::GM_Console)
             SDL_StopTextInput();
 
-        bool consumed = false;
+        bool consumed = SDL_IsTextInputActive();
         if (kc != OIS::KC_UNASSIGNED)
         {
-            consumed = SDL_IsTextInputActive() &&
-                    ( !(SDLK_SCANCODE_MASK & arg.keysym.sym) && std::isprint(arg.keysym.sym)); // Little trick to check if key is printable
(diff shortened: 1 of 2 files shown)


scrawl:
Expanding the key map is good, but we should fix the root cause too.

Does this patch also fix the issue (without applying the key map change)?
<pre>
diff --git a/apps/openmw/mwinput/inputmanagerimp.cpp b/apps/openmw/mwinput/inputmanagerimp.cpp
index 284919f..842959c 100644
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -665,14 +665,11 @@ namespace MWInput
                 && MWBase::Environment::get().getWindowManager()->getMode() == MWGui::GM_Console)
             SDL_StopTextInput();
 
-        bool consumed = false;
-        if (kc != OIS::KC_UNASSIGNED)
-        {
-            consumed = SDL_IsTextInputActive() &&
-                    ( !(SDLK_SCANCODE_MASK & arg.keysym.sym) && std::isprint(arg.keysym.sym)); // Little trick to check if key is printable
-            bool guiFocus = MyGUI::InputManager::getInstance().injectKeyPress(MyGUI::KeyCode::Enum(kc), 0);
-            setPlayerControlsEnabled(!guiFocus);
-        }
+        bool consumed = SDL_IsTextInputActive() &&
+                ( !(SDLK_SCANCODE_MASK & arg.keysym.sym) && std::isprint(arg.keysym.sym)); // Little trick to check if key is printable
+        bool guiFocus = MyGUI::InputManager::getInstance().injectKeyPress(MyGUI::KeyCode::Enum(kc), 0);
+        setPlayerControlsEnabled(!guiFocus);
+
         if (!mControlsDisabled && !consumed)
             mInputBinder->keyPressed (arg);
         mJoystickLastUsed = false;
</pre>

Antifleischwolf:
Your patch fixed it for every key that isn't printable.
So I just removed the test whether it is printable or not.
I also left the check for an unassigned key because I'm not 100% certain about what MyGui would do.

scrawl:
There is no check for KC_UNASSIGNED in keyReleased either, so removing the check all together shouldn't cause any problems. Also we will still need the "setPlayerControlsEnabled" call for unknown MyGUI key codes.

scrawl:
> So I just removed the test whether it is printable or not.

That we can not do.
The printable check is there so that you can continue using non-printable keybindings while a textfield is being edited.
Example: Taking a screenshot (F12) while the console is open should be possible, but after removing the printable check it is not.

Antifleischwolf:
You can leave the editbox with escape and then take a screenshot.
That would leave the possibility to use non-printable key bindings while they wouldn't interfere while typing.

scrawl:
You can, but the user would expect that the screenshot function still works whether or not they're focusing an editbox. It certainly does on my desktop.
There are other examples, e.g. the GameMenu binding (Esc key) when a dialog is up should immediately close that dialog. After this patch, it no longer does that when the dialog contains an editbox, you have to press Esc *twice*.
You don't need non-printable keys when you're typing something. So where is the interference?

Antifleischwolf:
What about shift, home, end, arrow keys...
More or less every non-printable key that gets used to navigate within a text input.

scrawl:
You have a point there. Maybe a proper fix would be to define a scope for each binding. Currently we have player control bindings (i.e. only active in Game mode, when no GUI window is opened), and other bindings. We could expand this to global bindings (e.g. screenshot, game menu) and local bindings. Local bindings would be disabled when a text input field is active, global bindings would be always enabled. The user would then be expected to bind global bindings to keys that do not interfere with text input.

jeffreyhaines:
@Antifleischwolf how is this issue going? Do you intend to resolve this issue? If not could you please close this pull-request. Thank you

PR #516, opened 2015-03-04
correction from Scrawl.

dteviot:
Now correctly handles skills/attributes.


scrawl:
I don't agree with the signal change. AFAIK this would mean activation with the keyboard no longer works.

dteviot:
@scrawl
> I don't agree with the signal change. AFAIK this would mean activation with the keyboard (enter/space) no longer works.

Fair enough. However, I still believe that as it stands the purpose of the code is not clear.  I only figured out it's purpose (and that activate is a side effect of double click) by accident.
Can I assume the only change you'd agree with to improve clarity would be adding a comment to slotAddonTableItemActivated() that it's called when an item is double clicked or activated via keyboard, and it's purpose is to toggle the check state?

scrawl:
Yes, adding comments is fine of course.

PR #519, opened 2015-03-06
oengine duplication fix

scrawl:
The OENGINE_BULLET files were getting compiled twice (once for opencs and once for openmw). I've moved them to a static library to fix this.


psi29a:
This has cut the amount of spam I get from clang by half, so thanks! :D

PR #520, opened 2015-03-06
Fix for MSVC Warnings C4099 & C4800

dteviot:


apps/esmtool/record.cpp
```diff
@@ -15,8 +15,8 @@ void printAIPackage(ESM::AIPackage p)
         std::cout << "    Distance: " << p.mWander.mDistance << std::endl;
         std::cout << "    Duration: " << p.mWander.mDuration << std::endl;
         std::cout << "    Time of Day: " << (int)p.mWander.mTimeOfDay << std::endl;
-        if (p.mWander.mShouldRepeat != 1)
-            std::cout <<  "    Should repeat: " << (bool)p.mWander.mShouldRepeat << std::endl;
+        if (!p.mWander.mShouldRepeat)
+            std::cout <<  "    Should repeat: " << p.mWander.mShouldRepeat << std::endl;
```
scrawl: The cast to bool is necessary, because the variable is declared as unsigned char and would format a char in the output.

apps/openmw/mwgui/windowmanagerimp.cpp
```diff
@@ -566,10 +566,10 @@ namespace MWGui
 
                     // Show the windows we want
                     mMap            ->setVisible(eff & GW_Map);
```
scrawl: Funny that this line isn't causing a warning. But can we change it for consistency anyway? Looks a bit strange like this.

apps/esmtool/record.cpp
scrawl: Oh, I see you changed the declaration too. But I suggest to change it back. None of the other esm structs are using bools. We need to be explicit about the size of the struct, so unsigned char is better style, I think.

PR #521, opened 2015-03-07
Editor - Bug #2429: The cloned record does not have the original record's data.

cc9cii:
Should fix bug #2429.

~~Actually, maybe it should be "Added", closing until a new solution is pushed.~~

### DIFF
--- a/apps/opencs/model/world/record.hpp
+++ b/apps/opencs/model/world/record.hpp
@@ -40,9 +40,9 @@ namespace CSMWorld
         ESXRecordT mBase;
         ESXRecordT mModified;
 
-        Record() = default;
-        Record(const Record&) = default;
-        Record& operator= (const Record&) = default;
+        Record();
+        Record(const Record& record);
+        Record& operator= (const Record& record);
 
         Record(State state,
                 const ESXRecordT *base = 0, const ESXRecordT *modified = 0);
@@ -68,6 +68,31 @@ namespace CSMWorld
         void merge();
         ///< Merge modified into base.
     };
+
+    template <typename ESXRecordT>
+    Record<ESXRecordT>::Record()
+    : mBase(), mModified()
+    { }
+
+    template <typename ESXRecordT>
+    Record<ESXRecordT>::Record(const Record& record)
+    : mBase(record.mBase), mModified(record.mModified)
+    {
+        mState = record.mState;
+    }
+
+    template <typename ESXRecordT>
+    Record<ESXRecordT>& Record<ESXRecordT>::operator= (const Record& record)
+    {
+        if(this != &record)
+        {
+            mBase = record.mBase;
+            mModified = record.mModified;
+            mState = record.mState;
+        }
+
+        return *this;
+    }
 
     template <typename ESXRecordT>
     Record<ESXRecordT>::Record(State state, const ESXRecordT *base, const ESXRecordT *modified)
--- a/apps/opencs/model/world/record.hpp
+++ b/apps/opencs/model/world/record.hpp
@@ -63,7 +63,7 @@ namespace CSMWorld
     };
 
     template <typename ESXRecordT>
-    Record<ESXRecordT>::Record(State state, const ESXRecordT *base = 0, const ESXRecordT *modified = 0)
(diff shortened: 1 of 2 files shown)


cc9cii:
About to push a new solution.

cc9cii:
@zinnschlag, I don't think this is correct, either.  It seems wrong to be copying stuff in a clone() command.  Maybe State_ModifiedOnly should mean the data is read from mBase rather than mModified?

zinnschlag:
I don't follow what this PR is about.

> Maybe State_ModifiedOnly should mean the data is read from mBase rather than mModified? 

No. ModifiedOnly means that the record does not exist in base. Only in modified.

And cloning records should not clone base data. That would break revert. Base is immutable after loading. You are not allowed to change it at all.

cc9cii:
> I don't follow what this PR is about

When a referenceable record is cloned, only mBase is copied then mState is set as State_ModifiedOnly.  That means later when the newly cloned record is inserted to the table, mModified (uninitialised at this point) is returned from the Record<ESXRecordT>::get() method.

Two possible solutions I can see are either:

(a) copy mBase to mModified during clone() which is what I've done, or
(b) have Record<ESXRecordT>::get() return mBase when mState == State_ModifiedOnly

Maybe you can see another way?

zinnschlag:
Option (a) is actually correct. Guess I didn't get it because the back and forth stuff is a bit confusing. As mentioned before (b) would break pretty much everything. It is really, really wrong.

The implementation of (a) in this PR is a bit unfortunate and also not entirely correct (you don't set mState correctly). Maybe it is time to add a new constructor to CSMWorld::Record? Something like:

RecordBase (State state, const ESXRecordT *base = 0, const ESXRRecordT *modified = 0)

with mBase and mModified only set if the respective argument is not 0
?

You can then call this constructor with (ModifyOnly, 0, &(this->get())). This would allow writing the clone function in a single instruction that does not risk having a memory leak (in case of an exception).

cc9cii:
Ok, will do. A question, though.  Commands use clone() methods for undo/redo.  Wouldn't they want both mBase and mModified (as well as mState) preserved?  i.e. do we also need another clone() method?

cc9cii:
Seems to make every record for a new addon to be "Added", which doesn't seem correct.  Are they meant to be "Base"?

cc9cii:
Fixed.

zinnschlag:
>  A question, though. Commands use clone() methods for undo/redo. Wouldn't they want both mBase and mModified (as well as mState) preserved? i.e. do we also need another clone() method?

Well, shit! You are right. When I was looking at the clone function I was thinking of the clone command. But these are not the same. Sorry, got confused there.

It seems that the original Record clone function was correct. And when we look at the clone command, it actually does not make use of the clone function, but uses Collection<ESXRecordT, IdAccessorT>::cloneRecord, which seems to do the right thing too (what you proposed as (a)).

Actually, it looks like pretty much everything has been working as it is supposed to all along. So that brings us back to my original question. What problem do you want to address with this PR?

cc9cii:
> What problem do you want to address with this PR?

The problem: Bug #2429. The cloned record does not have the original record's data.

zinnschlag:
Okay. Now I get it. The problem was specific to RefIdCollection. Apparently the RefIdCollection::cloneRecord function is broken.

We are almost there now. Your modifiedCopy function still has a potential memory leak. Either go with the new constructor we had earlier or just use a std::auto_ptr. Also, remove the record->mBase = ESXRecordT() line. It is unnecessary and does not what you think what it is doing anyway. And while we are at it, you can construct the new Record object with the default constructor instead of the copy constructor. Anything that is copied is overwritten anyway.

cc9cii:
Yeah, I've been sloppy :-(  Sorry about that.

zinnschlag:
No problem. Shit happens.

One more round, unfortunately. We don't C++11 yet. Therefore the = default statement can not be used.

cc9cii:
Done.  Not sure if I can safely let the compiler to generate copy constructor and assignment operator?

zinnschlag:
You need the default constructor, because if there are any other constructors the compiler will not create one automatically. Copy constructor and assignment operator are not affected by this. The defaults will do fine.

I am going to quickly clean that up myself and then merge.

PR #524, opened 2015-03-09
fix compiler detection and adding build flags in CMakeLists.txt

maqifrnswa:
compiler flags are not being set on linux (ubuntu) for two reasons:

1) CMAKE_COMPILER_IS_GNUCC is not being set, so it never enters the section of the script that sets the flags. I can't tell exactly why (ubuntu/debian using a wrapper c++ cc over g++ gcc? maybe since the project is not defined as a CXX or CC project it doesn't set the CMAKE_COMPILER_IS_GNU* variables?). However, it works when I check  CMAKE_CXX_COMPILER_ID instead of CMAKE_COMPILER_IS_GNUCC. Also gives us the ability to set separate flags for clang and gnu compilers, which I use for -Wno-unused-but-set-parameter which doesn't exist in clang.

2) COMPILE_FLAGS is not a GLOBAL property:
http://www.cmake.org/cmake/help/v3.2/manual/cmake-properties.7.html#properties-of-global-scope
It is only target/directory level properties. I use CMAKE_CXX_FLAGS instead, and it works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,22 +349,22 @@ if (NOT WIN32 AND NOT APPLE)
         "${OpenMW_BINARY_DIR}/openmw-cs.desktop")
 endif()
 
-# Compiler settings
-if (CMAKE_COMPILER_IS_GNUCC)
-    set_property(GLOBAL APPEND_STRING PROPERTY COMPILE_FLAGS "-Wall -Wextra -Wno-unused-parameter -Wno-reorder -std=c++98 -pedantic -Wno-long-long")
+# CXX Compiler settings
+if (CMAKE_CXX_COMPILER_ID STREQUAL GNU OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wno-unused-parameter -Wno-reorder -std=c++98 -pedantic -Wno-long-long")
 
     execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion
                 OUTPUT_VARIABLE GCC_VERSION)
-    if ("${GCC_VERSION}" VERSION_GREATER 4.6 OR "${GCC_VERSION}" VERSION_EQUAL 4.6)
-        set_property(GLOBAL APPEND_STRING PROPERTY COMPILE_FLAGS "-Wno-unused-but-set-parameter")
-    endif("${GCC_VERSION}" VERSION_GREATER 4.6 OR "${GCC_VERSION}" VERSION_EQUAL 4.6)
+    if (CMAKE_CXX_COMPILER_ID STREQUAL GNU AND "${GCC_VERSION}" VERSION_GREATER 4.6 OR "${GCC_VERSION}" VERSION_EQUAL 4.6)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-but-set-parameter")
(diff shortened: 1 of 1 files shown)


scrawl:
CMAKE_CXX_FLAGS we were using previously, but for me that only worked when CMAKE_BUILD_TYPE is set to "None". It did not work for "RelWithDebInfo".

COMPILE_FLAGS does work for me (I checked using VERBOSE=1 make and can see the flags are passed). Different cmake versions maybe?

<pre>
scrawl-laptop:~$ cmake --version
cmake version 2.8.12.2
</pre>

maqifrnswa:
hm same version
~/openmw/openmw/build$ cmake --version
cmake version 2.8.12.2

on the proposed branch:

    $ cmake -DCMAKE_BUILD_TYPE="Release" ..
    $ VERBOSE=1 make
    cd /home/showard/openmw/openmw/build/libs/openengine && /usr/bin/c++   -DOGRE_PLUGIN_DEBUG_SUFFIX=\"_d\" -DOGRE_PLUGIN_DIR=\"/usr/lib/i386-linux-gnu/OGRE-1.9.0\" -Wall -Wextra -Wno-unused-parameter -Wno-reorder -std=c++98 -pedantic -Wno-long-long -Wno-unused-but-set-parameter ...

So it works

on master:

    $ cmake -DCMAKE_BUILD_TYPE="Release" ..
    $ VERBOSE=1 make
    $ cd /home/showard/openmw/openmw/build/libs/openengine && /usr/bin/c++   -DOGRE_PLUGIN_DEBUG_SUFFIX=\"_d\" -DOGRE_PLUGIN_DIR=\"/usr/lib/i386-linux-gnu/OGRE-1.9.0\" -O3 -DNDEBUG -I/home/showard/openmw/openmw/. ...

no flags were added from the part of CMakeLists.txt we were talking about.

I also used the -DCMAKE_EXPORT_COMPILE_COMMANDS=ON option to see the compile_commands.json and confirm the flags were not there on master and were there on the proposed branch.

Also, the fact that the travis build failed (due to log length limit being reached from so many -Wall warnings that were not showing up before) I think shows that the flags weren't being passed before this pull request, at least on travis perhaps?

maqifrnswa:
regarding the travis build failure: now that it knows about the warning flags, we should tell it what are system libraries so it doesn't report warnings for those, e.g.
http://www.cmake.org/cmake/help/v3.0/command/include_directories.html

    include_directories("."     ${LIBS_DIR}
    SYSTEM
    ${OGRE_INCLUDE_DIR} ${OGRE_INCLUDE_DIR}/Ogre ${OGRE_INCLUDE_DIR}/OGRE ${OGRE_INCLUDE_DIRS} ${OGRE_PLUGIN_INCLUDE_DIRS}
    ${OGRE_INCLUDE_DIR}/Overlay ${OGRE_Overlay_INCLUDE_DIR}
    ${SDL2_INCLUDE_DIR}
    ${Boost_INCLUDE_DIR}
    ${PLATFORM_INCLUDE_DIR}
    ${MYGUI_INCLUDE_DIRS}
    ${MYGUI_PLATFORM_INCLUDE_DIRS}
    ${OPENAL_INCLUDE_DIR}
    ${BULLET_INCLUDE_DIRS}
    )

scrawl:
Ok, you are right, setting the flags works on all build types. The uninitialized warnings are new. But it looks like a problem in our code and not Ogre headers. And I think we want to leave warnings on system headers enabled in general.

maqifrnswa:
1) the new commit (d7baf923536a01f1f77713c4869480cfb8403ac6) doesn't hide warnings (but it does help with using include-what-you-use from clang), so it's an optional commit for this bug.

2) The warnings from travis on linux come from pull #526.

3) The warnings from travis on macosx comes from the fact that -Woverloaded-virtual is part of -Wall on clang (which the travis macosx uses) and is not part of -Wall in gcc (which is what the linux build uses).

PR #526, opened 2015-03-09
Properly initialize KeyT in nifkey.hpp

EmperorArthur:
Travis logs from pull Request #524 showed a warning that KeyT was not being properly initialized.
This pull fixes that.

### DIFF
--- a/components/nif/nifkey.hpp
+++ b/components/nif/nifkey.hpp
@@ -18,6 +18,15 @@ struct KeyT {
     float mTension;    // Only for TBC interpolation
     float mBias;       // Only for TBC interpolation
     float mContinuity; // Only for TBC interpolation
+    KeyT() :
+        mValue(T()),
+        mForwardValue(T()),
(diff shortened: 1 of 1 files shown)


components/nif/nifkey.hpp
```diff
@@ -18,6 +18,15 @@ struct KeyT {
     float mTension;    // Only for TBC interpolation
     float mBias;       // Only for TBC interpolation
     float mContinuity; // Only for TBC interpolation
+    KeyT() :
+        mValue(T()),
```
scrawl: This will still not work for Ogre::Vector3, because its default constructor does not initialize anything.

EmperorArthur: Of course not.  That would be too easy.

scrawl:
FWIW, the OSG branch should have this warning fixed, because OSG's vectors are initialized by default and the (currently unused) TBC values were commented out.

PR #527, opened 2015-03-11
fix -Wnewline-eof warnings

maqifrnswa:
This is one of the default warnings on clang

C-standard says you need a new line at the end of a file
c++ before c++11 says you need a new line at the end of a file
c++11 and later says it doesn't matter.

So it can be ignored or merge this branch. If ignored, we can do -Wno-newline-eof on clang


scrawl:
I approve of this change. Actually my editor auto-adds the newline at the end of the file whenever I save, so we often end up with unnecessary changes. 

Most of these files seem to be coming from @zinnschlag. Perhaps you could configure your editor to also add a newline to end of files?

zinnschlag:
I just checked. My current editor does automatically add a newline. 

However I went though about half a dozen editors during my participation in OpenMW. I guess one of them must have been badly configured.

Anyway, not a current problem.

PR #530, opened 2015-03-12
Add saving land and land textures.  Should resolve Bug #2447.

cc9cii:
Only tested by examining the saved file with a binary editor (for LAND and LTEX).

### DIFF
--- a/apps/opencs/model/doc/saving.cpp
+++ b/apps/opencs/model/doc/saving.cpp
@@ -93,6 +93,10 @@ CSMDoc::Saving::Saving (Document& document, const boost::filesystem::path& proje
 
     appendStage (new WritePathgridCollectionStage (mDocument, mState));
 
+    appendStage (new WriteLandCollectionStage (mDocument, mState));
+
+    appendStage (new WriteLandTextureCollectionStage (mDocument, mState));
+
     // close file and clean up
     appendStage (new CloseSaveStage (mState));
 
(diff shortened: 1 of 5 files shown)


psi29a:
*rubs hands together greedily*

psi29a:
Terrain is still not being loaded by OpenMW. 

Tested by opening the ESM, saving it out as omwaddon then loading OpenMW. The result is the same, see only the water plain and falling through.

cc9cii:
Ok, will have a look.

zinnschlag:
IIRC some of the terrain data is loaded in asynchronously on demand. This would of course be a problem for saving, because the data you want to save might not be there.

Forcing a full load for everything in the modified section might be the solution here, because unless the user only wants to look at the content file without saving it, that needs to be loaded anyway. In this case it should be made sure that the data is completely loaded during the load process (and not just have the asynchronous loading triggered); maybe even.bypassing the asynchronous aspect of the loading process.

Another option might be to load the missing data when saving. This would save loading time in the case of the user only wanting to look at a content file without saving it. However I suspect that this approach would be a lot more complicated.

cc9cii:
@psi29a please try this one.

@zinnschlag you fixed the data loading issue last time yourself.

zinnschlag:
Really? can't remember ever working on that. Getting old, I guess.

psi29a:
Well done!  File size jumped from 1MiB to 4MiB which is more inline with the original ESM file. In addition to this, I can now see terrain and do everything normally.  Great job and thank you!

@zinnschlag no worries, you've been at this awhile. ;)

PR #531, opened 2015-03-13
Silencing more MSVC 2013 warnings.

dteviot:


dteviot:
I'm not sure that the Travis build failure is my doing.
The error in the log looks like this:
> The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).
> The build has been terminated.
But I can't see how all the warnings are my doing.

psi29a:
It is the OSX build that is failing... 
https://travis-ci.org/OpenMW/openmw/jobs/54216889

indeed, the warnings coming out of this thing is killing the job. This is Clang giving all the warnings, while Linux is using gcc... and giving less warnings.

dteviot:
This is a total mess.  I'll wait until cc9cii's PR that gets the Mac building again is merged into trunk and resubmit.

PR #532, opened 2015-03-13
Fix some compiler warnings.

cc9cii:


apps/opencs/view/world/dragrecordtable.cpp
```diff
@@ -3,7 +3,7 @@
 #include "../../model/world/tablemimedata.hpp"
 #include "dragrecordtable.hpp"
 
-void CSVWorld::DragRecordTable::startDrag (const CSVWorld::DragRecordTable& table)
+void CSVWorld::DragRecordTable::startDragToTable (const CSVWorld::DragRecordTable& table)
```
zinnschlag: Isn't that a drag from the table instead?

cc9cii: You are probably correct.  Didn't really think too hard on this one.  I'll submit a patch.

PR #537, opened 2015-03-15
Fix comparison of integers of different signs

naclander:
Clang reported comparison of unsigned long with long. This cast should fix it.

### DIFF
--- a/components/bsa/bsa_file.cpp
+++ b/components/bsa/bsa_file.cpp
@@ -111,7 +111,7 @@ void BSAFile::readHeader()
     // Each file must take up at least 21 bytes of data in the bsa. So
     // if files*21 overflows the file size then we are guaranteed that
     // the archive is corrupt.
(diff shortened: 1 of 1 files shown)


components/bsa/bsa_file.cpp
```diff
@@ -111,7 +111,7 @@ void BSAFile::readHeader()
     // Each file must take up at least 21 bytes of data in the bsa. So
     // if files*21 overflows the file size then we are guaranteed that
     // the archive is corrupt.
-    if((filenum*21 > fsize -12) || (dirsize+8*filenum > fsize -12) )
+    if((filenum*21 > unsigned(fsize -12)) || (dirsize+8*filenum > unsigned(fsize -12)) )
```
Mingun: I think, instead of cast it is better just place `-12` to other side of `>` with inverted sign.

PR #539, opened 2015-03-16
OpenCS FileDialog crash fix

scrawl:
The file dialog would crash when no game file is selected and an addon file with no dependency is checked, then unchecked.


psi29a:
Nice find, I ran into this over the weekend. :)

PR #541, opened 2015-03-16
fix memory leaks in emulating command line

sandstranger:


apps/openmw/android_commandLine.cpp
```diff
@@ -1,18 +1,20 @@
...
 
 JNIEXPORT void JNICALL Java_ui_activity_GameActivity_commandLine(JNIEnv *env,
 		jobject obj, jint argc, jobjectArray stringArray) {
 	jboolean iscopy;
 	argcData = (int) argc;
-	argvData[0]="openmw";
-	for (int i = 0; i < argcData; i++) {
-		jstring string = (jstring) (*env).GetObjectArrayElement(stringArray, i);
-		argvData[i+1] = (env)->GetStringUTFChars(string, &iscopy);
+	argvData= new const char * [argcData+1];
```
Mingun: Hmm... You allocate memory with `new`, but release it with `free`. This is bad practice.

sandstranger:
I see ,than I will use use delete

PR #542, opened 2015-03-16
release jni memory

sandstranger:


apps/openmw/android_commandLine.cpp
```diff
@@ -1,20 +1,27 @@
...
-	argvData= new const char * [argcData+1];
+	argvData = new const char *[argcData + 1];
 	argvData[0] = "openmw";
-	for (int i = 1; i < argcData+1; i++) {
-		jstring string = (jstring) (*env).GetObjectArrayElement(stringArray, i-1);
+	for (int i = 1; i < argcData + 1; i++) {
+		jstring string = (jstring) (env)->GetObjectArrayElement(stringArray,
+				i - 1);
 		argvData[i] = (env)->GetStringUTFChars(string, &iscopy);
+		(env)->DeleteLocalRef(string);
```
Mingun: It is safe delete local ref to string, if `iscopy` is `false`? And for `stringArray`?

sandstranger:
"Calling DeleteLocalRef is very similar to setting a Java local variable to null. Both can make an object eligible for garbage collection a little earlier; but both are ultimately unnecessary as the JVM will do it automatically when the method returns. You'd use DeleteLocalRef if you had a long-running native method and wanted to clean up your mess as you go -- the same reason given for (occasionally) setting local vars to null. "
 Taken from here http://www.coderanch.com/t/274355/java/java/call-DeleteLocalRef . So I suppose it is safe .

PR #543, opened 2015-03-17
Travis clang analyzer

maqifrnswa:
adds clang 3.6 and clang static analyzer testing as optional builds on travis. Can also be used to test c++11 building if needed since clang 3.6 can do -std=c++11.


maqifrnswa:
ready for merge. code coverage with clang doesn't work due to travis ci using precise toolchains, whenever they upgrade you can get rid of the check that disables it on clang, if desired.

PR #550, opened 2015-04-01
Silence "potentially evaluated expression" warnings

maqifrnswa:
boost::shared_ptr is not a real pointer, and the *operator of boost::shared_ptr is evaluated, so -Wpotentially-evaluated-expression catches them when used in typeid(*boost::shared_ptr), even though they are safe. I pretty much copied what LibreOffice did for the same problem, just turn off the warnings around functions we know are safe. This will silence a lot of noise in the clang static-analyzer builds. The warnings are true warnings, but we know them to be safe as long as we don't start overloading * in the derived classes of boost::shared_ptr used in openmw.

also a commit fixing a newline-eof warning


zinnschlag:
Don't agree with this. Makes the code harder to read and also harder to write. With this solution any new code that uses this kind of construct would be required to reimplement this pre-processor spew, which is an unnecessary burden on the coder.
IMO having adjustments to warning levels around individual bits of code is generally a bad idea (not to mentioned that this also introduces more platform specific code).
This is an example of an overreaching warning. Best just disable it and be done with it once and for all.

maqifrnswa:
ok - just disabled it for clang >= 3.6 on linux

PR #551, opened 2015-04-03
Provide incremental update of SpellWindow (Fixes #2411)

dteviot:
When SpellWindow is visible, every 0.5 seconds update the cost/changes for spells/enchanted items shown.
Also, check to see if more substantial update of the window is required.


scrawl:
Can you please use a plain struct instead of boost::tuple? Other than that, looks ok. Tested the impact on a character with a few hundred spells, the incrementalUpdate() takes 1-2ms. That should be fine for now given it's run only 2 times a second. Maybe post 1.0 we can experiment with an event-based system.

dteviot:
For unknown reason, fix requested by Scrawl did not get added.  So closing and trying again.

dteviot:
Seems to be a problem at github, can't create new PR either.

PR #553, opened 2015-04-09
Feature #1219

cc9cii:
Working:
* Actor - inventory, spells, destinations, AI packages
* Container - contents
* Pathgrid - points, edges
* Region - sounds
* Faction - reaction map
* Race/Birthsign - spells
* Enchantment/Spell - effects
* Potion - effects
* Armor/Clothes - part refs
* Items/creatures - levelled lists
* dialogue subview display in non-table format - levelled creatures/items
* incorporated enum delegates
* TopicInfos result script shown in dialogue subview

Known issue:
* AiWander idle is shown as a single integer


zinnschlag:
That is a lot of stuff. From a first cursory review it looks mostly okay. I will add a few inline comments in a bit.

I noticed that you only deal seem to deal with table-like substructures. There are probably other cases (some of the script fields maybe?).

For the tables, is that meant to be the exhaustive list or just the ones that you have gotten working at this point; because it seems this list isn't complete. At least it is missing AI packages.

Haven't got around to run tests myself yet. Does revert work on sub-tables? From looking at the code it would guess that might be a problem.

apps/opencs/model/world/columnbase.hpp
```diff
@@ -121,14 +131,32 @@ namespace CSMWorld
...
         virtual std::string getTitle() const;
 
-        virtual int getId() const;
+        virtual int getId() const; // FIXME: why have an accessor for a public member?
+    };
+
+    class NestableColumn : public ColumnBase
+    {
+        std::vector<NestableColumn *> mNestedColumns;
+        bool mHasChildren;
```
zinnschlag: Is mHasChildren needed? Seems to me you might be able to just check if mNestedColumns.empty().

apps/opencs/model/world/columnimp.hpp
```diff
@@ -2265,6 +2264,148 @@ namespace CSMWorld
...
+    template<typename ESXRecordT>
+    struct PathgridIndexColumn : public Column<ESXRecordT>
+    {
+        PathgridIndexColumn()
+        : Column<ESXRecordT> (Columns::ColumnId_PathgridIndex, ColumnBase::Display_Integer)
+        {}
+
+        virtual QVariant get (const Record<ESXRecordT>& record) const
+        {
+            return QVariant(); // FIXME
```
zinnschlag: What is this about?

apps/opencs/model/world/columns.cpp
```diff
@@ -256,6 +287,11 @@ int CSMWorld::Columns::getId (const std::string& name)
 
 namespace
 {
+    static const char *sSkills[] =
```
zinnschlag: Looks like something is missing here. But also possible duplication. We must already have the list of skills elsewhere (as a separate record table IIRC).

apps/opencs/model/world/commands.cpp
```diff
@@ -171,3 +172,71 @@ void CSMWorld::CloneCommand::undo()
...
+                                                    int parentColumn,
+                                                    QUndoCommand* parent) :
+    mId(id),
+    mModel(model),
+    mParentColumn(parentColumn),
+    QUndoCommand(parent),
+    mNestedRow(nestedRow),
+    NestedTableStoring(model, id, parentColumn)
+{
+    setText (("Delete nested row in " + mId).c_str());
```
zinnschlag: Not sure about the wording. This is text the user will see and probably be confused about "nested row".

apps/opencs/model/world/refidcollection.hpp
```diff
@@ -44,7 +46,7 @@ namespace CSMWorld
 
         private:
 
-            const RefIdAdapter& findAdaptor (UniversalId::Type) const;
+            /*const*/ RefIdAdapter& findAdapter (UniversalId::Type) const;
```
zinnschlag: Problematic. The const is there for a reason; namely that adapters don't hold document state and therefore should not be modified.

apps/opencs/model/world/collection.hpp
```diff
@@ -120,6 +120,8 @@ namespace CSMWorld
 
             virtual const Record<ESXRecordT>& getRecord (int index) const;
 
+            virtual Record<ESXRecordT>& getRecord (int index);
```
zinnschlag: Do we really need this one? That would give us two methods to modify the record (non-const getRecord and setRecord).

cc9cii:
> I noticed that you only deal seem to deal with table-like substructures. There are probably other cases (some of the script fields maybe?).

> For the tables, is that meant to be the exhaustive list or just the ones that you have gotten working at this point; because it seems this list isn't complete. At least it is missing AI packages.

To be honest I have no idea what is missing.  My main motivation was to get pathgrid tables working.  If there are other things I need to look into, just let me know.

> Does revert work on sub-tables?

Only tested a few scenarios but seems ok.

I will comment separately to the in-line comments.

zinnschlag:
> Known issue: Editing "ID" within nested tables in dialogue subview does not work

Not sure if that is related. I recently run into a lot of trouble with Display_ enum mix-ups, that affected (among others) the ability to modify data. The fixes are in my replace branch. Maybe merging that into your branch is worth a try?

apps/opencs/model/world/collection.hpp
cc9cii: I needed a way to get a reference to a record to modify (i.e. non-const version), passed as a parameter, but couldn't think of a way to get it working with setRecord.  I'll have another think about it. 

It's being used here:

    template<typename ESXRecordT, typename IdAccessorT>
    void SubCellCollection<ESXRecordT, IdAccessorT>::addNestedRow(int row, int column, int position)
    {
        getAdapter(Collection<ESXRecordT, IdAccessorT>::getColumn(column))->addNestedRow(
                Collection<ESXRecordT, IdAccessorT>::getRecord(row), position);
    }

apps/opencs/model/world/columnimp.hpp
cc9cii: Opps, should be a quick fix

apps/opencs/model/world/columns.cpp
cc9cii: Not entirely sure what sirherrbatka had in mind here.

apps/opencs/model/world/refidcollection.hpp
cc9cii: That is from one of sirherrbatka's commits.  I guess one of the adapters is modifying the document.  I'll see if there is a way around it.

apps/opencs/model/world/columns.cpp
zinnschlag: Probably best to remove it then.

apps/opencs/model/world/columnbase.hpp
cc9cii: You are correct. Unnecessary.  But I just like having members if they allow avoiding lots of function calls.

apps/opencs/model/world/commands.cpp
cc9cii: Ok.

apps/opencs/model/world/refidcollection.hpp
zinnschlag: Adapters can modify the document even when they are const.

cc9cii: Right you are.  Fixed.

apps/opencs/model/world/columns.cpp
cc9cii: Done

apps/opencs/model/world/columnimp.hpp
cc9cii: Actually, they are intended to return invalid values. Since these columns have sub-rows, the methods in idadapterimp.hpp are used instead.

apps/opencs/model/world/columnbase.hpp
cc9cii: Replaced.

apps/opencs/model/world/collection.hpp
cc9cii: Workaround is not co pretty.  Both getRecord() and setRecord() provide access to const references only, so this messing around with assign() is needed.  It would be useful to have something like modifyRecord() with a non-const reference.

    template<typename ESXRecordT, typename IdAccessorT>
    void SubCellCollection<ESXRecordT, IdAccessorT>::addNestedRow(int row, int column, int position)
    {
        Record<ESXRecordT> record;
        record.assign(Collection<ESXRecordT, IdAccessorT>::getRecord(row));

        getAdapter(Collection<ESXRecordT, IdAccessorT>::getColumn(column)).addNestedRow(record, position);

        Collection<ESXRecordT, IdAccessorT>::setRecord(row, record);
    }

cc9cii:
The ID column editing turned out to be indexing issue.  Not quite 100% with non-editable columns, however.

cc9cii:
Done.  Please review.

zinnschlag:
1. The tables in the dialogue subview are too small (vertically) to be useful.

2. The layout of the dialogue sub views is odd. We seem to have all the regular (old) fields in a section with an additional scrollbar (which shouldn't be there to begin with) and then all the new fields outside of it. This separation does not make sense. 

3. Segfault on shutdown. Steps to reproduce:
    1. Edit Record for NPC referenceable
    2. Scroll down to destinations
    3. Add row

4. Info record dialogue subviews are still missing the script field.

5. Does not compile warning-free:

> In file included from /home/marc/OpenMW/openmw/apps/opencs/model/world/refidcollection.cpp:9:0:
> /home/marc/OpenMW/openmw/apps/opencs/model/world/refidadapterimp.hpp: In member function ‘QVariant CSMWorld::BodyPartRefIdAdapter<ESXRecordT>::getNestedData(const CSMWorld::RefIdColumn*, const CSMWorld::RefIdData&, int, int, int) const [with ESXRecordT = ESM::Clothing]’:
> /home/marc/OpenMW/openmw/apps/opencs/model/world/refidcollection.cpp:857:1:   instantiated from here
> /home/marc/OpenMW/openmw/apps/opencs/model/world/refidadapterimp.hpp:1734:21: warning: comparison is always true due to limited range of data type [-Wtype-limits]
> /home/marc/OpenMW/openmw/apps/opencs/model/world/refidadapterimp.hpp: In member function ‘QVariant CSMWorld::BodyPartRefIdAdapter<ESXRecordT>::getNestedData(const CSMWorld::RefIdColumn*, const CSMWorld::RefIdData&, int, int, int) const [with ESXRecordT = ESM::Armor]’:
> /home/marc/OpenMW/openmw/apps/opencs/model/world/refidcollection.cpp:857:1:   instantiated from here
> /home/marc/OpenMW/openmw/apps/opencs/model/world/refidadapterimp.hpp:1734:21: warning: comparison is always true due to limited range of data type [-Wtype-limits]

cc9cii:
Thanks for the review.

Points 1 and 2 are UI related and will always be subjective.  I can go back to the previous way of having a single scrollbar (I found that having to scroll down all the time was getting irritating after a while).

I can't reproduce the segfault at exit on my system.  Any chance of a stack trace please?

Point 4: are you talking about ESM::DialInfo::mResultScript?  This goes to the "Topic Infos" table, yes?

I'll fix the warnings.

cc9cii:
That warning appears to be gcc pedantry (see http://stackoverflow.com/questions/10434640/gcc-comparison-is-always-true-due-to-limited-range-of-data-type-in-template) but I'll still see if there is a workaround.

cc9cii:
DialInfo::mResultScript

zinnschlag:
Another crash: 

> openmw-cs: /home/marc/OpenMW/openmw/apps/opencs/model/world/nestedtableproxymodel.cpp:62: virtual int CSMWorld::NestedTableProxyModel::rowCount(const QModelIndex&) const: Assertion `!index.isValid()' failed.

Happened after clicking on the Edit Record menu item of a NPC in the referenceable table.

Here is the callstack: https://gist.github.com/zinnschlag/b7e60b358ca17a783ce5

I would give you the callstack for the other segfault too, but I can't even get to it now.

cc9cii:
Sorry, it was a copy/paste from an older branch that caused it.

zinnschlag:
Okay. No problem. Shit happens.

Here is the other callstack: https://gist.github.com/zinnschlag/1f12a1a56b97de847620

cc9cii:
It's a stab in the dark, but would you mind trying this one please.

cc9cii:
I think the issue with undo() is existing in master branch already.  undo() is using a pointer to a table model that is already deleted when a subview (or dialoguesubview) is closed.

I think one solution might be for the document to maintain the model pointers.  Another might be to ask for accept/reject when a user tries to close a subview.

I'll have a go with the first option.

zinnschlag:
Your fix seems to work. Will merge.

> I think the issue with undo() is existing in master branch already. undo() is using a pointer to a table model that is already deleted when a subview (or dialoguesubview) is closed. Not sure why the pointer is invalid at this point, since we get it from mData ??

This is a separate issue and a seriously weird one. The models do not get deleted until the document is getting deleted. The logic here is correct. There must be a bug somewhere. Mind filing an issue on the tracker?

zinnschlag:
Just a wild guess. Maybe the commands operate on the proxy models instead of the actual models.  That would be wrong and result in the crash we are seeing because the proxy models will get deleted when the respective table view gets killed.

cc9cii:
Your guess is correct.  When CSVWorld::Table calls setModel() it passes the pointer of the proxy model.  Qt then uses this when calling undo() triggered from signals.

apps/opencs/model/world/refidadapterimp.hpp
```diff
@@ -808,6 +840,1248 @@ namespace CSMWorld
...
+                        content.mWander.mDistance = static_cast<short>(value.toInt());
+                    else
+                        return; // return without saving
+                case 2:
+                    if (content.mType == ESM::AI_Wander ||
+                            content.mType == ESM::AI_Follow || content.mType == ESM::AI_Escort)
+                        content.mWander.mDuration = static_cast<short>(value.toInt());
+                    else
+                        return; // return without saving
+                case 3:
```
elsid, line 1549: It's unclear whether this fallthrough is intentional.

PR #556, opened 2015-04-24
Undo delete - referenceables table

cc9cii:
Please review before merging. There ~~are two issues~~ is still an issue with this:

* ~~Had to bloat IdTree interface by adding a seemingly unrelated searchId() method to extract UniversalId associated with a record  EDIT: maybe CSVWorld::Table::getUniversalId() can be used instead?~~
* Undo() appends the previously deleted record at the end of the container vector.  On the screen the record is placed somewhere else.  To the end user it appears as if undo has failed. (but by re-ordering the columns one can see that undo was successful)

Not sure what to do about ~~either of the issues~~ this issue.

### DIFF
--- a/apps/opencs/model/world/commanddispatcher.cpp
+++ b/apps/opencs/model/world/commanddispatcher.cpp
@@ -8,6 +8,7 @@
 #include "../doc/document.hpp"
 
 #include "idtable.hpp"
+#include "idtree.hpp"
 #include "record.hpp"
 #include "commands.hpp"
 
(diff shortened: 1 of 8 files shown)


zinnschlag:
Will review in a bit.

About the ordering issue: That is not specific to sub-tables, right? You are bringing up a general usability issue. And it probably also affects redo and the new record function. That should be added to the bug tracker.

Possible solution: When records are added to a table that currently has views the views jump to these records (or the first of them anyway). We could also automatically select the respective records. This behavior (both the jumping and the selection) should be controllable through user settings.

cc9cii:
Ok, will submit a bug report.

PR #557, opened 2015-04-25
Moved References

cc9cii:
Feature #2262.


cc9cii:
Morrowind.esm original

Move the rock via 3D editing. Save as Morrowind.omwgame then load the new save.

Morrowind.omwgame moved rock

cc9cii:
Just noticed that I introduced a bug in cellref.cpp.

cc9cii:
Hmmm.... I think this isn't quite working.  I don't see a MVRF record in Morrowind.omwgame.

cc9cii:
@zinnschlag, should the editing views (references subtable, dialogue as well as 3d editing) automatically update the record's cell data when the ref's position changes?  At the moment they simply update the x/y/z values.

cc9cii:
Closing till I make some more progress

cc9cii:
Made some changes, please review, especially the saving part.

zinnschlag:
Sorry, should have been more specific about that. Loading a move that was made with vanilla TES-CS was always working AFAIK. Seems a lot of things have changed while I was away. Will have a look.

cc9cii:
Saving and re-loading also works.  This is Tribunal.omwgame with a rock moved:

hex editor

cc9cii:
> Loading a move that was made with vanilla TES-CS was always working AFAIK

Yes but the loaded references were placed with the wrong cell (see the changes in data.cpp)

zinnschlag:
Found a problem. Steps to reproduce:
- Open the Balmora cells in a scene (-3,-2 to -4, -3)
- Move the Silt Strider to the other side of the scene.
- Save

Console output:

std::bad_alloc
Unclosed record remaining

cc9cii:
I don't get that.  Any chance of backtrace or log files please?  I won't be able to get to the Ubuntu VM till tomorrow.

balmora

cc9cii:
Should be fixed.

zinnschlag:
Better. No more crash. But instead of moving the Silt Strider, it is copied now (the original is still in place after a reload in OpenMW-CS).

cc9cii:
There are two problems.

First, Data::mRefLoadCache was getting cleared when content file changes.  So the ref loaded by content file 0 (Morrowind.esm) was missed by the moved ref loaded by content file 1 (Tribunal.omwgame).

Second, the map search in RefCollection::load() uses mRefNum as the key to find the ref.  But because the content file number is different, a cached entry is missed and a new ref is loaded (and now we have two, the original as well as the moved one).

I hacked the files to get things working, but rather ugly.  Also, I'm not 100% sure if the moved ref should have the content file number 1 (probably).

cc9cii:
The content file numbers appear to be our local construct, so ignoring it when looking through the cache.

~[fixed](http://i872.photobucket.com/albums/ab289/cc9cii/MVRF-Balmora-fixed.jpg)

zinnschlag:
Okay. That seems to work. I think this is ready to be merged into master and also ready for the 0.36.0 release. I want more testing anyway and releasing it now seems like the right way to get that. Agreed?

cc9cii:
Yes, agreed.

zinnschlag:
Okay. Will merge. Thanks.

components/esm/loadcell.cpp
```diff
@@ -176,10 +176,20 @@ bool Cell::getNextRef(ESMReader &esm, CellRef &ref, bool& deleted)
 
     // NOTE: We should not need this check. It is a safety check until we have checked
     // more plugins, and how they treat these moved references.
-    if (esm.isNextSub("MVRF")) {
-        // skip rest of cell record (moved references), they are handled elsewhere
-        esm.skipRecord(); // skip MVRF, CNDT
-        return false;
+    if (esm.isNextSub("MVRF"))
+    {
+        if (ignoreMoves)
```
scrawl: Isn't this naming backwards? <i>ignoreMoves</i> would suggest when the parameter is true, the MVRF record is skipped, but the code does the opposite.

components/esm/cellref.cpp
```diff
@@ -4,13 +4,34 @@
...
+void ESM::RefNum::load (ESMReader& esm, bool wide)
 {
-    loadId(esm, wideRefNum);
+    if (wide)
+        esm.getHNT (*this, "FRMR", 8);
+    else
+        esm.getHNT (mIndex, "FRMR");
+}
+
+void ESM::RefNum::save (ESMWriter &esm, bool wide, const std::string& tag) const
```
scrawl: This is confusing. Why does the save method need a tag argument, but the load method does not? It would make sense to use the same data structure for loading and saving.

components/esm/cellref.cpp
```diff
@@ -19,10 +40,8 @@ void ESM::CellRef::loadId(ESMReader &esm, bool wideRefNum)
     if (esm.isNextSub ("NAM0"))
         esm.skipHSub();
 
-    if (wideRefNum)
-        esm.getHNT (mRefNum, "FRMR", 8);
-    else
-        esm.getHNT (mRefNum.mIndex, "FRMR");
+    if (!ignoreRefNum)
```
scrawl: I can't find any uses of the ignoreRefNum parameter. Is that a leftover of earlier code?

cc9cii: Might be best for Zini to comment.  I basically fixed a few bugs without thinking too much about what was already present.

components/esm/cellref.cpp
zinnschlag: During the save procedure is RefNum is saved with a MVRF tag in some cases (see savingstages.cpp). I am actually not quite sure why the same isn't true for the load part.

components/esm/cellref.cpp
zinnschlag: Yeah, leftover. Will fix.

PR #560, opened 2015-04-29
Handle plugins that has 0x00 for levelled list types,

cc9cii:
for example Ravenloft v5.02d, to use 0x01.

Unfortunately the fix spams the console with error messages.

### DIFF
--- a/apps/opencs/model/world/columns.cpp
+++ b/apps/opencs/model/world/columns.cpp
@@ -222,7 +222,6 @@ namespace CSMWorld
             { ColumnId_HitSound, "Hit Sound" },
             { ColumnId_AreaSound, "Area Sound" },
             { ColumnId_BoltSound, "Bolt Sound" },
-            { ColumnId_OriginalCell, "Original Cell" },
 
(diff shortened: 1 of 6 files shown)


scrawl:
Those are flags and not a type enum. All bitmask combinations (including 0x00) are perfectly valid. The original code is wrong.
Also, could you make it use the enum please instead of 0x01/0x02 literals.

cc9cii:
Ok.  Where can I find information about what the bitmask flags are and what they mean?

cc9cii:
Vanilla CS uses checkboxes.  We don't have a way of doing that for bitmasks in the dialogue subview at the moment (for nested columns).  Also, vanilla has "blocked" which we don't seem to have.

zinnschlag:
For bitmasks each bit should form a column of its own (boolean value). And we don't use blocked. We can ignore it completely.

cc9cii:
All done, except I seem to have found an issue with undo/redo.  Investigating...

cc9cii:
Undo labels fixed as well.

PR #564, opened 2015-05-08
NPCs react to theft when game paused (Fixes #2521)

dteviot:
BugFix: itemTaken() needs the item's original cell to check for ownership.


dteviot:
Scrawl fixed this before me.

PR #565, opened 2015-05-09
Add TopicInfos special conditions table.

cc9cii:
Should resolve Bug #2524.

Does not validate variable/function names. Does not auto detect which values should be float.


cc9cii:
screenshot

zinnschlag:
Looks mostly good. I am not too happy about the way mapping between functions and function numbers is handle. Basically you are adding the mapping twice (once in each direction) and produce a lot of code duplication with it. Its not essential but having an implementation that does not list each number and each function name twice would be preferable.

cc9cii:
Ok, will do.  Since these were known in advance and won't change (for Morrowind, anyway) I didn't see much issue, but I guess duplication is still duplication.

PR #566, opened 2015-05-10
temporarily remove the test for bullet version

zinnschlag:
Just asking travis for a quick test run of this change.


scrawl:
That will get it building, but the movement physics will be incorrect when built with bullet 2.82 or earlier. If you want to revert the requirement, then best revert the two associated commits too (I am fine with that, btw).

zinnschlag:
Hm, didn't see any problems. But I only did minimal tests. Okay, will revert.

PR #568, opened 2015-05-13
Fix for Bug #2543.  Spells magic effect id typo.

cc9cii:
Added another fix (thanks scrawl)


psi29a:
Is this for 0.36 and/or master?

cc9cii:
That would be up to Zini I guess.  I was thinking that master would be ok.

PR #570, opened 2015-05-15
Modifying an object instance (e.g. 3D edit) triggers the instances

cc9cii:
table to scroll to the corresponding record.  Should resolve Feature #2554.


zinnschlag:
Limiting this feature to instances is problematic. It will still catch other cases besides modifications in the 3D-view (like modifications in another view of the same table), which would cause an inconsistency between the instance table and all other tables.

cc9cii:
That's fair enough.  Any idea on how we can provide the functionality without causing inconsistencies?

cc9cii:
How about per-subview (e.g. checkbox) setting to auto update after a modify?  That way only the user has knowledge and control over how the view updates.  Also, it means that this can be applied to all tables.

zinnschlag:
Difficult. With the bottom bar (status and creation) and the top bar (record filter and soon column filter) I don't see how we can add another checkbox in there without cluttering up things and taking away even more valuable screenspace.
Hm, a context menu item maybe?

zinnschlag:
Or maybe add a small button (re-using the user settings icon ?) to the top panel that temporarily pops up a configuration panel over the filter panel in the same way as the add/clone panel pops up over the status bar. This configuration panel would then contain both the column filter selected widget, the checkbox you proposed and everything else that we may come up with later.

cc9cii:
Ok, I'll think about this.

cc9cii:
Done, please review.

cc9cii:
temporarily closing while investigating something odd

cc9cii:
Figured it out, reopening.

cc9cii:
Done, ready for merge (after a review, that is)

cc9cii:
Any feedback on this PR please?

zinnschlag:
Looks mostly okay (except for the point I commented on inline).

There is one issue with the GUI:

Clicking on the config icon should replace the bar, not just add the checkbox. Otherwise we would have no room for the column filter UI.

cc9cii:
Closing.  Kinda useless since osg will be merged soon.

PR #573, opened 2015-05-17
Add an optional horizontal scrollbar to the main view window.

cc9cii:
Should resolve Feature #2549.

- TableSubviews ~~and DialogueSubviews~~ now provide size hints
- Option to stop the growth of the window at the screen boundary for multi-monitor setup with different resolution.
- Three options in user preferences:
  Grow Only: No change to current behaviour, except the use of size hints
  Scrollbar Only: Simple addition of a scrollbar, the view window does not grow
  Grow then Scroll: Window grows as per current behaviour. The scrollbar appears once it cannot grow any further.

### DIFF
--- a/apps/opencs/view/doc/view.cpp
+++ b/apps/opencs/view/doc/view.cpp
@@ -563,12 +563,7 @@ void CSVDoc::View::addSubView (const CSMWorld::UniversalId& id, const std::strin
     //
     CSMSettings::UserSettings &settings = CSMSettings::UserSettings::instance();
     QString mainwinScroll = settings.settingValue ("window/mainwindow-scrollbar");
-    if (mainwinScroll.isEmpty() || mainwinScroll == "Scrollbar Only")
-        mScrollbarOnly = true;
-    else if(mainwinScroll == "Grow Only")
-        mScrollbarOnly = false;
-    else
-        mScrollbarOnly = false;
+    mScrollbarOnly = mainwinScroll.isEmpty() || mainwinScroll == "Scrollbar Only";
 
     QDesktopWidget *dw = QApplication::desktop();
     QRect rect;
--- a/apps/opencs/view/doc/view.cpp
+++ b/apps/opencs/view/doc/view.cpp
@@ -9,6 +9,10 @@
 #include <QDockWidget>
 #include <QtGui/QApplication>
 #include <QDesktopWidget>
+#include <QScrollArea>
+#include <QHBoxLayout>
+#include <QDesktopWidget>
+#include <QScrollBar>
 
 #include "../../model/doc/document.hpp"
 #include "../../model/settings/usersettings.hpp"
@@ -16,6 +20,7 @@
 #include "../../model/world/idtable.hpp"
 
 #include "../world/subviews.hpp"
+#include "../world/tablesubview.hpp"
 
 #include "../tools/subviews.hpp"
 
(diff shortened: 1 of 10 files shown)


apps/opencs/view/doc/view.cpp
```diff
@@ -527,6 +550,59 @@ void CSVDoc::View::addSubView (const CSMWorld::UniversalId& id, const std::strin
...
+    // - Move the mainwindow position sideways if necessary to fit within the screen.
+    // - Any more additions increases the size of the mSubViewWindow (horizontal scrollbar
+    //   should become visible)
+    // - Move the scroll bar to the newly added subview
+    //
+    CSMSettings::UserSettings &settings = CSMSettings::UserSettings::instance();
+    QString mainwinScroll = settings.settingValue ("window/mainwindow-scrollbar");
+    if (mainwinScroll.isEmpty() || mainwinScroll == "Scrollbar Only")
+        mScrollbarOnly = true;
+    else if(mainwinScroll == "Grow Only")
```
zinnschlag: This if statement looks redundant.

apps/opencs/view/world/dialoguesubview.cpp
```diff
@@ -581,7 +583,15 @@ CSVWorld::DialogueSubView::DialogueSubView (const CSMWorld::UniversalId& id, CSM
...
-    QWidget *mainWidget = new QWidget(this);
+    //QWidget *mainWidget = new QWidget(this);
+    CSVDoc::SizeHintWidget *mainWidget = new CSVDoc::SizeHintWidget;
+
+    const QRect rect = QApplication::desktop()->screenGeometry(this);
+    int frameHeight = 40; // set a reasonable default
+    QWidget *topLevel = QApplication::topLevelAt(pos());
+    if (topLevel)
+        frameHeight = topLevel->frameGeometry().height() - topLevel->height();
+    mainWidget->setSizeHint(QSize(400, rect.height()-frameHeight)); // FIXME: 400
```
zinnschlag: What is the issue with this magic number?

zinnschlag:
I suggest to move the SizeHintWidget class to a separate hpp/cpp file pair, since it isn't just some helper class for SubView.

And the the user settings need tooltips that explain what exactly they are doing, because that isn't entirely obvious.

We discussed before the possibility of having this feature both for horizontal and vertical growth. Wouldn't that make sense?

apps/opencs/view/doc/view.cpp
cc9cii: Yes.  That section was doing something earlier, but not needed any longer.  Will fix.

apps/opencs/view/world/dialoguesubview.cpp
cc9cii: There is no easy way to pre-calculate the required width, so I'll just remove the size hint from dialogue subview.

cc9cii:
> move the SizeHintWidget class to a separate hpp/cpp file pair

Ok, will move the class out.

> user settings need tooltips

Wasn't sure how this was done for radio buttons...

> this feature both for horizontal and vertical growth

The vertical scroll bar also appears already - once you stack enough widgets and shrink the height of the window.  However you'd normally not see it because Qt adds the dockwidgets to the right and we don't set minimum heights on widgets.  To make vertical use effectively I suspect we have to overwrite Qt's dock management implementation.

PR #574, opened 2015-05-17
Silence -Wreorder warnings, part 2

rohit-n:


scrawl:
Thanks for taking the time to do this, I'm all for warning-less code.

PR #576, opened 2015-05-19
More dialogue only editable fields

cc9cii:
Should resolve Bug #2546 (Race), Bug #2547 (NPC) and Bug #2551 (Cell).

~~Not 100% confident about water level handling, although I did my best to interpret correctly.~~


apps/opencs/model/world/columns.cpp
```diff
@@ -62,7 +62,7 @@ namespace CSMWorld
             { ColumnId_StarterSpell, "Starter Spell" },
             { ColumnId_AlwaysSucceeds, "Always Succeeds" },
             { ColumnId_SleepForbidden, "Sleep Forbidden" },
-            { ColumnId_InteriorWater, "Interior Water" },
+            { ColumnId_Water, "Has Water" },
```
zinnschlag: Don't think that this is a good idea, because this really is only for interiors. Might confuse people when it shows up in an exterior cell.

cc9cii:
Ok, I was confused because there are two flags for interior water - one as a bit flag and another boolean.  If you're going to revert back then the other one needs to be renamed as well (or there'll be two with the same column id)

zinnschlag:
The mWaterInt bool is set if the cell record contains a water level sub-record. The name is kinda bad.

Reviewing again now.

zinnschlag:
Another problem. Column names must be unique. You have at least one name used twice.

cc9cii:
Which one?

apps/opencs/model/world/refidadapterimp.cpp
```diff
@@ -538,6 +552,340 @@ void CSMWorld::NpcRefIdAdapter::setData (const RefIdColumn *column, RefIdData& d
...
+void CSMWorld::NpcSkillsRefIdAdapter::setNestedData (const RefIdColumn *column,
+        RefIdData& data, int row, const QVariant& value, int subRowIndex, int subColIndex) const
+{
+    Record<ESM::NPC>& record =
+        static_cast<Record<ESM::NPC>&> (data.getRecord (RefIdData::LocalIndex (row, UniversalId::Type_Npc)));
+    ESM::NPC npc = record.get();
+    ESM::NPC::NPDTstruct52& npcStruct = npc.mNpdt52;
+
+    if (subRowIndex < 0 || subRowIndex >= ESM::Skill::Length)
+        std::cout << "setNestedDatat index" << std::endl;
```
cc9cii: Oops...

cc9cii:
Any hints on which column names are repeated?  I checked columns.hpp several times and I can't see any duplications.

zinnschlag:
There is at least Skills and Rank. But I also found ID in the code before these commits. Wouldn't hurt to systematically check for duplicates.

cc9cii:
I don't understand.  ColumnId_Skill doesn't even appear to be used anywhere in the entire code base.  Similarly, ColumnId_Rank is used only once (struct RankColumn).  What am I missing here?

cc9cii:
Done.

PR #577, opened 2015-05-19
Reset an activated reference after activation's execution (Fix #1738)

smbas:


scrawl:
Works for me, and shouldn't cause any regressions with the previous activate bugs (#1166, #1346).

psi29a:
Nice job! :)

PR #578, opened 2015-05-20
Wording and debugging of latex code of openCS manual

Cramal:
Mainly correcting the latex code so it compile and create the pdf

Some wording and style correction also


zinnschlag:
Line ending problems and the merge conflict unfortunately means that this is a no go. Sorry.

PR #580, opened 2015-05-21
Check for the order of ESM::DialInfo::mId before inserting a new reco…

cc9cii:
…rd.  Should resolve Bug #2569.

The solution is to make the index order match the map storage order.  But that means for some records the index of the next topic is smaller than the index of the previous topic.  I don't know if this will cause any issues.

### DIFF
--- a/apps/opencs/model/world/infocollection.cpp
+++ b/apps/opencs/model/world/infocollection.cpp
@@ -28,7 +28,7 @@ void CSMWorld::InfoCollection::load (const Info& record, bool base)
         {
             index = getInfoIndex (record2.get().mPrev, topic);
 
-            if (index!=-1)
+            if (index!=-1 && getRecord(index).get().mId < record2.get().mId )
                 ++index;
         }
 


zinnschlag:
>  But that means for some records the index of the next topic is smaller than the index of the previous topic. I don't know if this will cause any issues.

Yes. That would break everything (everything info record related anyway).

cc9cii:
Then how do we deal with situations where the input is incorrect?  (as it appears to be the case with Container Compainion.esp)

cc9cii:
I still think getTopicRange() is broken.

  std::map<std::string, int>::const_iterator iter = getIdMap().lower_bound (topic2);

does *not* return the first iterator that has the lowest index of the same topic.  Instead, it returns the iterator that has the lowest value of mId for the same topic.

zinnschlag:
> I still think getTopicRange() is broken.

> std::map::const_iterator iter = getIdMap().lower_bound (topic2);

You are correct. gettopicRange is broken.

> If infocollection relies on that condition being invariant (i.e. mId order is the same as index order), then infocollection does not work the way it was intended.

That is not the case. 

mIndex is sorted by the mId field (for quick lookups)
mRecords is sorted according to topics and the prev/next fields.

PR #582, opened 2015-05-22
Add record state tracking to the ModifyCommand (Fix #2555)

smbas:
Now the ModifyCommand also restores the record state when undoing.

I think it needs some review from more experienced developers and, perhaps, some more testing (On my machine all goes well (Win7 x64, OpenMW x86 build)). I don't know is the dynamic_cast the best solution for this, but I don't see another solution that doesn't require rewriting the current structure of the ModifyCommand.

Since the ModifyCommand receives general QAbstractItemModel I used additional variable to ensure that the model is really IdTable's successor and has a record state column. Is it redundant or not?

### DIFF
--- a/apps/opencs/model/world/commands.cpp
+++ b/apps/opencs/model/world/commands.cpp
@@ -14,7 +14,7 @@
 
 CSMWorld::ModifyCommand::ModifyCommand (QAbstractItemModel& model, const QModelIndex& index,
                                         const QVariant& new_, QUndoCommand* parent)
-: QUndoCommand (parent), mModel (&model), mIndex (index), mNew (new_)
+: QUndoCommand (parent), mModel (&model), mIndex (index), mNew (new_), mHasRecordState(false)
 {
     if (QAbstractProxyModel *proxy = dynamic_cast<QAbstractProxyModel *> (&model))
     {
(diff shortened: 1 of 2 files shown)


cc9cii:
That's great.  Maybe do something similar with delete command as well?

zinnschlag:
The modify code looks good to me. Why would we need adjustments to the delete command? I am not aware of any problems there.

cc9cii:
I'm not aware of any specific problems, either.  Was just thinking out loud, you know, if modify had problems perhaps check  delete as well, etc, etc.

smbas:
The modify (undo) doesn't work after deletion (undo), because undo of the DeleteCommand creates new record with new index, but ModifyCommand stores the old index of the record. This can be fixed by adding id of the modified record to the ModifyCommand. But it will add id search to this command on the undo operation. Some optimization to this may be perform search only if ids don't match.

zinnschlag:
Crap. Just fixing the ModifyCommand won't help, because there are more commands that might be affected. Maybe we need to have another look at all the command classes. Maybe store the ID and the column instead of row and column (the index)? I suspect that would cause problems for the InfoCollection though, because there the row number actually matters.

Anyway, I am going to merge this for now.

PR #593, opened 2015-05-27
Resolve namespace clash issue in osx.

cc9cii:
User preference settings stuff removed until a thread safe method is sorted out.


apps/opencs/model/tools/pathgridcheck.cpp
```diff
@@ -29,6 +21,12 @@ int CSMTools::PathgridCheckStage::setup()
 
 void CSMTools::PathgridCheckStage::perform (int stage, CSMDoc::Messages& messages)
 {
+    // NOTE: This is horribly inefficient but in order to use signals the entire Stage class
+    // hierarchy needs to be braught under Qt which seems like an overkill for a small
+    // performance gain during verify operations
+    CSMSettings::UserSettings &userSettings = CSMSettings::UserSettings::instance();
```
zinnschlag: This is problematic, because the verifier runs in a different thread. We can't access the user settings from here. Also, this would allow changing configuration midways through a verifier run.

cc9cii: Ok - so not possible to use settings at all?  I tried putting the user setting in the constructor but means a restart before a new setting will take hold.

cc9cii:
No longer checking at each stage.

zinnschlag:
Yeah, we currently don't have a way to use user settings in operations. That is certainly something that needs to be fixed. I am just not sure how. With all the weirdness of Qt multi-threading I am already happy that we have it working at all.

Generally I see two possible approaches, but I am not sure if either of them is feasible:

1. Update the operations via a signal. Then extend the interface of stage so the update can be passed on to the contained stage objects. This approach would be completely viable, if it weren't for the fact that we also need to avoid changing settings in the middle of an operation.

2. Protect the user settings via a mutex.

cc9cii:
Ok, will have another try.

cc9cii:
Done

zinnschlag:
Same problem. Signal handler is accessed from two different threads. Besides, that is a specialized solution for the pathgrid setting. We need a general solution, because eventually we will have user settings in other operations/stages.

cc9cii:
> Same problem. Signal handler is accessed from two different threads.

I don't see any thread safety issue here.  The object is already fully constructed in Tools::getVerifier() before any thread stuff begins, then afterwards only updated via signals.

> We need a general solution

SignalHandler can be passed to any object constrctor in a similar way to pathgrid check ;-)

But seriously, I'm not sure how we can allow user preference setting access without some complicated code.

zinnschlag:
> I don't see any thread safety issue here. The object is already fully constructed in Tools::getVerifier() before any thread stuff begins, then afterwards only updated via signals.

I'll comment the relevant pieces of code inline.

apps/opencs/model/tools/signalhandler.cpp
```diff
@@ -0,0 +1,23 @@
...
+    connect (&CSMSettings::UserSettings::instance(),
+             SIGNAL (userSettingUpdated(const QString &, const QStringList &)),
+             this,
+             SLOT (updateUserSetting (const QString &, const QStringList &)));
+}
+
+void CSMTools::SignalHandler::updateUserSetting (const QString &name, const QStringList &list)
+{
+    if (name=="verifier/pathgrid-extra-check")
+        mExtraCheck = list.at(0) == "true";
```
zinnschlag: Here you access mExtraCheck from the main thread.

apps/opencs/model/tools/pathgridcheck.cpp
```diff
@@ -8,22 +8,24 @@
...
 {}
 
+CSMTools::PathgridCheckStage::~PathgridCheckStage ()
+{
+    delete mSigHandler;
+}
+
 int CSMTools::PathgridCheckStage::setup()
 {
+    mExtraCheck = mSigHandler->extraCheck();
```
zinnschlag: Here you access the extraCheck in mSigHandler from the operations-thread.

apps/opencs/model/tools/signalhandler.cpp
cc9cii: Doesn't Qt's signals have event queues to call that method in the operations-thread?  If not, I guess i misunderstood how they work.

cc9cii:
One more try pushed.

cc9cii:
I think all that extra signal/slot stuff are unnecessary.  From Qt's documentation:

> Auto Connection (default) If the signal is emitted in the thread which the receiving object has affinity then the behavior is the same as the Direct Connection. Otherwise, the behavior is the same as the Queued Connection.
> Queued Connection The slot is invoked when control returns to the event loop of the receiver's thread. The slot is executed in the receiver's thread.

In other words updateUserSetting(), which uses the default connection method, would have used Queued Connection and invoked *after* the event loop comes to the operations thread.

zinnschlag:
Using a different connection type does not help at all, if both the sender and the receiver run in the main thread. Qt multi-threading is highly unintuitive.

cc9cii:
Don't follow you.  Are you saying that Qt's connections don't work as per their documentation?

zinnschlag:
Having a queued connection between two objects in the same thread does not help if you access one of the objects afterwards from the 2nd thread. Your signal handler (which is setting the flag in question) is running in the main thread, while the operation (which is reading the flag) is running in the operations thread.

cc9cii:
> Your signal handler (which is setting the flag in question) is running in the main thread, while the operation (which is reading the flag) is running in the operations thread.

Not after  commit 01eba7b, essentially waits in a loop until the conditions are right.

zinnschlag:
That commit doesn't do anything. The current thread in the updateExtraCheck will always be the same as the thread of the SignalHandler object. Therefore the body of the if construct will never be executed.

cc9cii:
I guess there is no way of convincing you. (or myself I guess, since I could be wrong)  I will revert all except the part that is breaking osx builds as a separate PR.

cc9cii:
Ok, done.  Only the OSX fix remains.

cc9cii:
I think I understand what you're getting at.  You're saying that signals/slots might be thread safe but the read/write of the data (payload) isn't - yes?  In this case a read/write of a flag should be atomic and mutex guards should not be necessary, no?

zinnschlag:
> I think I understand what you're getting at. You're saying that signals/slots might be thread safe but the read/write of the data (payload) isn't - yes?

The signal/slot mechanism can safely be used between threads (there is no might, if it is set up properly). Normal variable access can not be safely used between threads. Currently the signal-slot connection to your signal handler is pointless, because the signal handler runs in the main thread, and the actual inter-thread communication happens exclusively via the other method (normal variable access).

>  In this case a read/write of a flag should be atomic and mutex guards should not be necessary, no?

A mutex guarded or an atomic variable would be an option. However it would be also inconstant with what we are using right now everywhere else to communicate between threads (which is the signal-slot system).

I had another look at the code. And it seems that the implementation is actually really easy. The Operation class does run in the operation thread. So why not use that? Configure it with the current settings value before calling OperationHolder::setOperation (the object has not been moved to the operations thread at this point and you can safely access it from the main thread.

After calling setOperation connect it to the regular user settings update mechanism. Actually, the implementation of that could be moved into the setOperation function, if you make the selection of relevant user settings configurable in Operation.

At this point Operation has the current settings. So all you need to do is to pass along these settings when setting up the stages.

cc9cii:
Sound ok, but can we implement that as a separate item?  As it is all OSX builds (as well as travis) is breaking due to the namespace issue (fixed in this PR)

PR #596, opened 2015-05-29
Allow users to select syntax highlighting colours.

cc9cii:
Should resolve Feature #2507.


zinnschlag:
We need a better colour selector. Having no selectable presets and not being able to see the selected colour is a substantial usability issue.

cc9cii:
> usability issue

Agreed.  But it is better than having no way to change the colours at all.  I'll mark the feature partially complete.

> We need a better colour selector

Also agreed, but I'm not keen on working on the user settings ui.

zinnschlag:
Okay.

PR #600, opened 2015-05-31
Feature #880: ID auto-complete

smbas:
**List of columns with the ID completion:**

Regions:
* Sleep Encounter (Objects)
* Sound Name (Sounds) **In Dialog SubView**

Cells:
* Region (Regions)
   
Objects:
* Model (Meshes)
* Script (Scripts)
* Icon (Icons)
* Enchantment (Enchantments)
* Original Creature (Objects)
* Open Sound (Sounds)
* Close Sound (Sounds)
* Race (Races)
* Class (Classes)
* Faction (Factions)
* Hair (BodyParts)
* Head (BodyParts)
* Male Part (BodyParts) **In Dialog SubView**
* Female Part (BodyParts) **In Dialog SubView**
* Item ID (Objects) **In Dialog SubView**
* Spell Id (Spells) **In Dialog SubView**
* Dest Cell (Cells) **In Dialog SubView**
* Levelled Item (Objects) **In Dialog SubView**

Instances:
* Object ID (Objects)
* Owner (Objects)
* Soul (Objects)
* Faction (Factions)
* Teleport Cell (Cells)
* Key (Objects)
* Trap (Spells)
* Owner Global (Globals)

Magic Effects:
* Icon (Icons)
* Particle (Textures)
* Casting Object (Objects)
* Hit Object (Objects)
* Area Object (Objects)
* Bolt Object (Objects)
* Casting Sound (Sounds)
* Hit Sound (Sounds)
* Area Sound (Sounds)
* Bolt Sound (Sounds)

Classes:
* Major Skill 1 (Skills)
* Major Skill 2 (Skills)
* Major Skill 3 (Skills)
* Major Skill 4 (Skills)
* Major Skill 5 (Skills)
* Minor Skill 1 (Skills)
* Minor Skill 2 (Skills)
* Minor Skill 3 (Skills)
* Minor Skill 4 (Skills)
* Minor Skill 5 (Skills)

Factions:
* Skill 1 (Skills)
* Skill 2 (Skills)
* Skill 3 (Skills)
* Skill 4 (Skills)
* Skill 5 (Skills)
* Skill 6 (Skills)
* Skill 7 (Skills)
* Faction (Factions) **In Dialog SubView**

Races:
* Spell Id (Spells) **In Dialog SubView**
   
Birthsigns:
* Texture (Textures)
* Spell Id (Spells) **In Dialog SubView**

Topic Infos:
* Actor (Objects)
* Race (Races)
* Class (Classes)
* Faction (Factions)
* Cell (Cells)
* PC Faction (Factions)
* Sound File (Sound Files)

Body Parts:
* Model (Meshes)
* Race (Races)

Sounds:
* Sound File (Sound Files)

Sound Generators:
* Creature (Objects)
* Sound (Sounds)


zinnschlag:
Will have to spend a bit of time on reviewing this. Won't happen today, but I should get around to it tomorrow.

zinnschlag:
Two larger issues and a couple of minor ones. I list the larger issues here and comment on the others inline:

1. There seems a to be a lot of data duplication. In idcompletionmanager.cpp and viewmanager.cpp
you have listings of display types. From a maintenance point of view that is at least unfortunate. I haven't looked at the details yet, but it seems we should already have this data stored somewhere. We should try to re-use it, so adding new columns does not become more error-prone by spreading out column data all over the place.

2. As far as I can tell the feature does only work for dialogue sub views. We also need it in the tables.

apps/opencs/model/world/idcompletionmanager.hpp
```diff
@@ -0,0 +1,38 @@
...
+            // Don't allow copying
+            IdCompletionManager(const IdCompletionManager &);
+            IdCompletionManager &operator = (const IdCompletionManager &);
+
+            void generateCompleters(CSMWorld::Data &data);
+
+        public:
+            IdCompletionManager(CSMWorld::Data &data);
+
+            bool isCompleterExistFor(CSMWorld::Columns::ColumnId id) const;
```
zinnschlag: hasCompleter or hasCompleterFor maybe?

isCompleterExistFor is rather lengthy and obscure.

apps/opencs/view/doc/viewmanager.cpp
```diff
@@ -60,6 +61,40 @@ CSVDoc::ViewManager::ViewManager (CSMDoc::DocumentManager& documentManager)
...
+        CSMWorld::ColumnBase::Display_Class,
+        CSMWorld::ColumnBase::Display_Creature,
+        CSMWorld::ColumnBase::Display_Faction,
+        CSMWorld::ColumnBase::Display_String,
+        CSMWorld::ColumnBase::Display_GlobalVariable,
+        CSMWorld::ColumnBase::Display_Icon,
+        CSMWorld::ColumnBase::Display_Mesh,
+        CSMWorld::ColumnBase::Display_Miscellaneous,
+        CSMWorld::ColumnBase::Display_Music,
+        CSMWorld::ColumnBase::Display_None, // Inplace editing (Table SubView) creates QLineEdit using Display_None
```
zinnschlag: That looks wrong. There shouldn't be an auto completer for type Display_None.

apps/opencs/view/doc/viewmanager.cpp
```diff
@@ -60,6 +61,40 @@ CSVDoc::ViewManager::ViewManager (CSMDoc::DocumentManager& documentManager)
...
         new CSVWorld::IdTypeDelegateFactory());
 
+    // Columns with QLineEdit editor
+    static const CSMWorld::ColumnBase::Display sIdCompletionColumns[] =
+    {
+        CSMWorld::ColumnBase::Display_Cell,
+        CSMWorld::ColumnBase::Display_Class,
+        CSMWorld::ColumnBase::Display_Creature,
+        CSMWorld::ColumnBase::Display_Faction,
+        CSMWorld::ColumnBase::Display_String,
```
zinnschlag: Display_String indicates that it is a string that does not represent any type of ID. There should be no way of auto-complete that.

apps/opencs/view/doc/viewmanager.cpp
```diff
@@ -60,6 +61,40 @@ CSVDoc::ViewManager::ViewManager (CSMDoc::DocumentManager& documentManager)
...
+        CSMWorld::ColumnBase::Display_Npc,
+        CSMWorld::ColumnBase::Display_Race,
+        CSMWorld::ColumnBase::Display_Referenceable,
+        CSMWorld::ColumnBase::Display_Region,
+        CSMWorld::ColumnBase::Display_Script,
+        CSMWorld::ColumnBase::Display_Skill,
+        CSMWorld::ColumnBase::Display_Sound,
+        CSMWorld::ColumnBase::Display_SoundRes,
+        CSMWorld::ColumnBase::Display_Static,
+        CSMWorld::ColumnBase::Display_String,
```
zinnschlag: Duplicate.

apps/opencs/view/doc/viewmanager.cpp
smbas: https://github.com/OpenMW/openmw/blob/master/apps/opencs/model/world/columnimp.hpp#L736-L739
https://github.com/OpenMW/openmw/blob/master/apps/opencs/model/world/columnimp.hpp#L1269-L1273
https://github.com/OpenMW/openmw/blob/master/apps/opencs/model/world/columnimp.hpp#L1830-L1834

Do this columns contain ID information or not?

apps/opencs/view/doc/viewmanager.cpp
smbas: Because of this: https://github.com/OpenMW/openmw/blob/master/apps/opencs/view/world/util.cpp#L146-L150

Standard Qt view can create editors only using this method. Dialog Subview calls createEditor() method explicitly. Tables make it implicitly through the standard interface.

apps/opencs/view/doc/viewmanager.cpp
smbas: Similar list presents in the codebase: https://github.com/OpenMW/openmw/blob/master/apps/opencs/model/world/columnbase.cpp#L26-L97
Need public method to get it.

But the problem with None display will remain. See above.

apps/opencs/model/world/idcompletionmanager.hpp
smbas: No problem, I'll rename it to hasCompleterFor.

smbas:
Tables also have ID auto-complete. Display_None is needed for this.

apps/opencs/view/doc/viewmanager.cpp
zinnschlag: Yes, but they are not supposed to be of type Display_String. Apparently we never updated the display types for these columns.

apps/opencs/view/doc/viewmanager.cpp
zinnschlag: Yeah, but Display_None is meant as a special flag (see the // Do not use comment in the definition). Maybe it would make sense to add a new Display enum for this use case?

apps/opencs/view/doc/viewmanager.cpp
zinnschlag: If we can use this list, that would be the way to go. Maybe add a listIdTypes function similar to the one used in UniversalId?

zinnschlag:
I didn't get any auto-complete in tables when I tested your branch. Will have another go at it once the other issues have been addressed. Don't expect a different result though.

apps/opencs/view/doc/viewmanager.cpp
smbas: We can extract proper Display type using index.data() with Display_Role. To do this we also need to add Display role in the data() methods of IdTable and IdTree.

apps/opencs/view/doc/viewmanager.cpp
smbas: There is one small problem. This list overlaps with EnumDelegate display types. So creating factories for the IdCompletionDelegate must be located before EnumDelegate factories.

smbas:
With this commits the tables are now have proper editors, but a checkbox in the table view is ugly:
untitled

If you'll still have not working ID auto-complete in the tables, check it also in nested tables (Inventory, Spells) in the Objects dialog subview. Also I'll need to know what's your OS.

zinnschlag:
Looks good.

Regarding booelan values: No point in touching that with the auto-complete feature. It worked fine before, so just exclude boolean columns?

Regarding auto-complete in tables: Looks like my tests were too limited. It works in some (or even most cases). So far the only case I found were it does not work is the region column in the cell table. Can you confirm that?

OS is Mint 13, but I don't think that matters here.

smbas:
I have auto-complete for the Region column.
region

Checkboxes have nothing to do with the auto-completion. Earlier, table editors were created using the default delegate. Now they are created using the createEditor() method from CommandDelegate (as in Dialog subviews). 
I can restore the previous behavior, but tables will not have spinboxes for numeric columns (there are other editors, like PlainTextEdit, but I think that spinboxes are more useful in table subviews than others).

zinnschlag:
> I have auto-complete for the Region column.

Very odd. I haven't found any other problem column yet. If it is really just this one column for a subset of user that shouldn't stop us from merging.

> Checkboxes have nothing to do with the auto-completion. Earlier, table editors were created using the default delegate. Now they are created using the createEditor() method from CommandDelegate (as in Dialog subviews). 
> I can restore the previous behavior

That is what I meant. This should be addressed on a per display type basis. For boolean the previous version was clearly better.

smbas:
I'll try to find a solution for this problem. If I'll have no success in this, I restore the previous behavior.

smbas:
This is a hack mostly, but it works.

And about the absence of the auto-completion for Region column: Are there any errors/messages in the console when you enter the text?

zinnschlag:
Funny. The region column just auto-completed for me (I don't think there were any error messages before). I hope this isn't some kind of intermittent bug.
Looks like there isn't anything more we can do about it right now. I am going to merge the PR. This will gives us more eyes on the feature and hopefully highlight if there are any remaining problems.

PR #601, opened 2015-05-31
Scroll to the top of the topic list when dialog is started. Should fix the bug #2504

smbas:


scrawl:
That would fix only the dialogue topic list, but not other cases (e.g. list of available enchantment effects). The fix should be built into MWList by default.

smbas:
I don't see such solution at the moment. Perhaps, WMList might reset the scrollbar, when it is hidden (list). But how it can know that?! I'll look into this issue more deeply.

smbas:
Scrollbars save their states not only in WMList, but also in other widgets that have scrollbar(s). This happens with windows that aren't recreated each time they are opened (e.g. Dialogue Window, Trade Window, Settings Window etc.).

**Vanilla behavior:**
In vanilla MW scrollbars are always reset when a player closes a window through the buttons "Cancel", "Goodbye" etc. In the menu mode scrollbars stay in their previous positions.

MWList in this situation works as needed. When a window is hidden, it doesn't reset to the initial position. I think we need manually add such reset code to the appropriate windows, when a player press "close" buttons. What do you think about it?

scrawl:
I seem to not have read the bug description correctly. We have at least two separate problems:
- The first time an MWList shows, it starts scrolled to the very bottom. It should start at the top.
-  When opening dialogue, we should scroll to the top, regardless of the previous scroll offset from a previously opened dialog window. For this your current scrollToTop() solution should be fine. I will add some inline comments on the code.

With the second problem fixed, the first problem would no longer be an issue in the dialog window case, but still a problem for other uses of MWList. Might as well file a separate bug (was on my todo list for a while).

apps/openmw/mwgui/dialogue.cpp
```diff
@@ -455,6 +456,7 @@ namespace MWGui
             mKeywordSearch.seed(Misc::StringUtils::lowerCase(*it), intptr_t(t));
         }
         mTopicsList->adjustSize();
+        mTopicsList->scrollToFirstItem();
```
scrawl: We can't scrollToFirstItem here. We get an updateTopics() call every time the player selects a topic. That means the list would incorrectly scroll to the top every time a topic is clicked.

apps/openmw/mwgui/dialogue.cpp
```diff
@@ -364,6 +364,7 @@ namespace MWGui
         bool sameActor = (mPtr == actor);
         mPtr = actor;
         mTopicsList->setEnabled(true);
+        mTopicsList->scrollToFirstItem();
```
scrawl: Best not scrollToFirstItem in the (!resetHistory && sameActor) case, since no new window was opened.

smbas: I did it as in vanilla MW.

scrawl: You can test this case by starting a dialogue, opening the console, click on the actor you're talking with and enter "forceGreeting".
That does not reset the topic scrollbar in vanilla MW.

smbas: As I have written above, in vanilla MW scrollbars are reset when the player press the close button (Ok,  Cancel, Close etc.).

smbas: I have looked at the scripting code for dialogues and understood you.

smbas:
Now it must work.

smbas:
Scrollbars in other windows may be reset in a similar way (in the exit() function). Of course, if you'll accept such solution.

scrawl:
Works for me. Feel free to apply the same to other MWList uses. Generally, a reset on start...() would be more logical than on exit(), but both is OK.

Good catch on the size_t bug.

smbas:
Your test case with forceGreeting in the console won't work, if reset will be in the start function.

scrawl:
It would work, if we insert it in the right place:
<pre>
    void DialogueWindow::startDialogue(MWWorld::Ptr actor, std::string npcName, bool resetHistory)
    ...
        if (resetHistory || !sameActor)
        {
            for (std::vector<DialogueText*>::iterator it = mHistoryContents.begin(); it != mHistoryContents.end(); ++it)
                delete (*it);
            mHistoryContents.clear();

            mTopicsList->scrollToTop();
        }
</pre>

smbas:
I already understand this.

smbas:
But it can be not appropriate for the other windows, maybe.

smbas:
I'll make appropriate changes for the other windows with scrollbars, and create a new pull request.

smbas:
The negation of the viewPosition is needed for checking this condition: https://github.com/OpenMW/openmw/blob/6cb221f8d09cce21ccdba5cae6e3fac0f0e49b1b/components/widgets/list.cpp#L103-L104.

PR #605, opened 2015-06-05
Reset Console editbox before command execution

smbas:
Should fix the bug #2593.

It seems that the accept event for Console command line was called multiple times during command execution, when GUI was disabled.

Test it!


scrawl:
I believe the cause for the issue is key repeating. During the loading screen, we get MyGUI frame events, but no input manager updates. MyGUI believes the Enter key to be pressed down for the duration of the loading screen, and then starts generating repeat key events, if the loading screen is sufficiently long.

Your workaround looks good, but we should also add a comment explaining why that particular order is important.

smbas:
I have forgot to add "... for the same command" at the end of the comment.

PR #606, opened 2015-06-06
AiTravel logic merged into AiPackage.

dteviot:
Eliminates duplicated code.


zinnschlag:
@scrawl It seems this one has fallen through the cracks. What is your opinion?

scrawl:
Didn't notice there was a new commit. I think this could use more discussion. Will reply on the [forum](https://forum.openmw.org/viewtopic.php?f=6&t=2916&p=33073#p33073).

PR #607, opened 2015-06-06
Keep comments

cc9cii:
Allow comments (lines starting with # character) and blank lines in openmw.cfg. Feature #2535.

- controlled via a checkbox in launcher settings
- spaces in front of '#' are allowed & ignored
- more of a workaround than a full featured solution, but probably good enough

### DIFF
--- a/components/config/gamesettings.cpp
+++ b/components/config/gamesettings.cpp
@@ -265,7 +265,10 @@ bool Config::GameSettings::writeFileWithComments(QFile &file)
         // Below is based on readFile() code, if that changes corresponding change may be
         // required (for example duplicates may be inserted if the rules don't match)
         if ((*iter).isEmpty() || (*iter).contains(QRegExp("^\\s*#")))
+        {
             stream << *iter << "\n";
+            continue;
+        }
 
         if (settingRegex.indexIn(*iter) == -1 || settingRegex.captureCount() < 2)
             continue;
--- a/components/config/gamesettings.cpp
+++ b/components/config/gamesettings.cpp
@@ -180,7 +180,7 @@ bool Config::GameSettings::writeFile(QTextStream &stream)
 //
 // - If a line in file exists with matching key and first part of value (before ',',
 //   '\n', etc) also matches, then replace the line with that of mUserSettings.
-// - else remove line (maybe replace the line with '#' in front instead?)
+// - else remove line (TODO: maybe replace the line with '#' in front instead?)
 //
 // - If there is no corresponding line in file, add at the end
 //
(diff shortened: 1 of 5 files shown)


scrawl:
Isn't the option redundant? Why would you *not* want to keep comments?

cc9cii:
I agree with you, but I never know what the end user wants (i.e. maybe some prefer the current behaviour)

scrawl:
> I agree with you, but I never know what the end user wants (i.e. maybe some prefer the current behaviour)

By that logic, we would need an option for every single change that makes it into the OpenMW codebase. Madness.

Every option we introduce comes at a maintenance cost, not to mention the additional complexity that the user is faced with. If we can't think of any reasons why a user might not like the change, then the option shouldn't exist.

In this particular case, keeping comments has no effect on OpenMW, it just means the config reader won't discard information that the user has entered. That is universally a good thing.

cc9cii:
Fair enough and like I said, I agree with you already.  I'm just being overly cautious i guess.  I can remove the option and if someone really wants it it can be re-added.

components/config/gamesettings.cpp
```diff
@@ -173,6 +174,138 @@ bool Config::GameSettings::writeFile(QTextStream &stream)
...
+                stream << key << "=\"" << value << "\"\n";
+                continue;
+            }
+        }
+        stream << key << "=" << value << "\n";
+    }
+
+    if (!userSettingsCopy.empty())
+    {
+        stream << "# new entries" << "\n";
```
scrawl: What is the point of this comment?

I noticed that my openmw.cfg now contains
<pre>
# new entries
# new entries

</pre>

at the end, and I fear that more of these will accumulate every time that I use the launcher, or on some other event. Can we remove it please?

cc9cii:
Ok, will do.

scrawl:
I just noticed this breaks the ordering of entries. Try starting the launcher, reordering a content file, then re-start. The change is not saved.

cc9cii:
Ok, will have a look.

cc9cii:
Well, ok, this is going to be more complicated than I had wished.

A simple solution is to place all 'content=' items at the end with the new order.  But that will mean the comments associated with those content items will be placed elsewhere.

A more complicated (but still not foolproof) solution might be to have the comments above an existing 'content=' entry to be kept together, with 'above' meaning comments up to an empty line. i.e. all three lines are kept as a block:
```
# this is a comment that stays with Morrowind.esm
#
# this is another comment that stays with Morrowind.esm
content=Morrowind.esm
```
As opposed to:
```
# this is a general comment that does not move

# this is a comment that stays with Morrowind.esm
content=Morrowind.esm
```
All of this might end up being quite complicated.

scrawl:
content= isn't the only setting where the order has a meaning. We also have
<pre>
data=
fallback-archive=
</pre>
, and who knows if more settings of this type will be introduced in the future.

> All of this might end up being quite complicated.

I can't think of a simple solution either. Perhaps we should revert the whole thing then for the time being.

kcat:
What about something like:
```
# Comments below this line will be lost!
fallback-archive=...
data=...
content=...
```
Then stop saving comments when you read that comment line.

scrawl:
Right, but then the commenting feature isn't so useful anymore. The original request was for being able to comment the purpose of a data directory, and being able to # comment out data lines so they can be re-enabled at a later time.

Will revert for now while we are sorting out a correct solution.

cc9cii:
I think I might have something.  Will push a PR in a few hours.

PR #608, opened 2015-06-07
Add handling DELE subrecords in Script records.

cc9cii:
Should resolve Bug #2211.  (Also related to Bug #2413)

- load() is split into two in order to handle DELE subrecords

### DIFF
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -23,14 +23,14 @@ opencs_units (model/world
 
 
 opencs_units_noqt (model/world
-    universalid record commands columnbase scriptcontext cell refidcollection
+    universalid record commands columnbase scriptcontext cell refidcollection idcollection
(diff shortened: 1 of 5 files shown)


scrawl:
We need a more generic solution. First off, your solution fixes the editor, but not the game. (There *was* a fix for the game before, but it broke with commit 49d912e5b6b1d728bf10c92d0b803dfda7596b89).

Second, I am pretty sure that applies to any record, not just script records. Records might include a DELE tag at any specific point, which is currently not handled correctly. Start of the record, end of the record, in between, you name it. There are various test cases scattered around the bugtracker, although I can't remember where I found the "in between" case.

https://bugs.openmw.org/issues/2413
https://bugs.openmw.org/issues/1414
https://bugs.openmw.org/issues/2211

I think the only correct fix is to have each load method handle deletion:
<pre>
/// @return false if the record has been deleted
bool load(ESMReader& reader);
</pre>

zinnschlag:
Agreed. Moving the deletion handling during loading to the esm structs is the right approach here. Maybe we should also consider the save case?

PR #611, opened 2015-06-11
Fixed "running in circles" bug caused when recalc path and closest way point is the one NPC has just passed.

dteviot:


apps/openmw/mwmechanics/aipackage.hpp
```diff
@@ -4,6 +4,8 @@
 #include "pathfinding.hpp"
 #include <components/esm/defs.hpp>
 
+#include "../mwworld/cellstore.hpp"
```
terrorfisch: I think a forward declaration would be sufficient here.

scrawl:
Can we fix this for all AiPackages at once please? 

Is there any reason now to build a "non-synced" path? If the answer is <i>no</i>, then we should merge buildSyncedPath into buildPath so that all AiPackages profit from the change.

dteviot:
See https://forum.openmw.org/viewtopic.php?f=6&t=2916&p=33121 for discussion

scrawl:
@zinnschlag good for merge, I think.

scrawl:
Noticed a problem introduced by this commit. In "Balmora, South Wall Cornerclub" Phane Rielle always faces the wall, instead of towards the room.

dteviot:
@scrawl,
I'll look into it.  After I confirm Windows builds are working again later tonight.

dteviot:
@scrawl
I do not believe there is a bug.
"phane rielle" has an AI of wander 64 with a speed of 46.
Looking at the Path Grid, this means there are two points he moves between.  One directly behind the bar, the other beside the wall behind the first waypoint.
When phane walks to the wall waypoint, when he stops he faces the wall.  Then when he walks back to the bar, when he stops he faces the bar.
He does seem to spend a lot of time at the wall waypoint, and it can take a while before he moves.

dteviot:
@scrawl,
On further thought, this may be a bug.
Introduced by https://github.com/dteviot/openmw/commit/be6ee927b906313cc7a19ed4298da725c22aebbd
Basically, if AiWander is not 0, then distance is expanded until NPC has two Path Nodes to wander between.
I suspect a wander distance of 64 is so small that Phane does not wander in vanilla.
Could be fixed by only doing the distance expansion if wander is greater than some tolerance value. (e.g. 128 or 256).
Note to self, should analyse the Morrowind data files, and see what the wander distances are.  May be other characters with same problem. (Should allow me to figure out what the best threshold value is.) Will do this weekend if I remember.

Incidentally, I suspect be6ee92 is also responsible for the "Disallow AiWander fast-forward for water creatures" bug that you've checked in a fix for.  As it stands, be6ee92 will probably make a water creature try to swim onto land, as again, most way-points are on land.  Suggested fix, getAllowedNodes() should probably return nothing for water creatures.
Instead, water creatures probably should get their own pathing.  If I remember, I'll do a little studying of Vanilla this weekend.

dteviot:
Further discussion at https://forum.openmw.org/viewtopic.php?f=6&t=2916&p=33235

PR #614, opened 2015-06-12
AiCombat: fixes and reworking

mrcheko:
Fixed some combat behavoiur bags
Added some combat features
Finished some Todo's
6f0c8bf - attempt to fix http://bugs.openmw.org/issues/1366, but seems like it can't be fixed on mechanics level (rather on physics one) - after this commit some things work: guards rather often can hit mudcrabs, rats etc. (guards with longspears are absolutely OK here). Cliff racers still don't hit:( Note: I tried here different configurations of hit distance tweaking, changing formulas in World::getHitContact and PhysicsSystem::getHitContact but to no avail - some creatures like cliff racer, hunger, ogrim still missing almost every attack.


scrawl:
This is going to cause a lot of conflicts with the OSG branch. I don't know which branch is going to get merged first, but either you or I will have a fun merging session ahead.

apps/openmw/mwmechanics/aicombataction.cpp
```diff
@@ -432,6 +465,8 @@ namespace MWMechanics
     {
         if (actor.getClass().hasInventoryStore(actor))
         {
+            mIsNpc = true;
```
scrawl: The name of this variable is misleading. Creatures can have an InventoryStore.

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -326,95 +284,66 @@ namespace MWMechanics
...
-                weapRange = fHandToHandReach;
-            }
-            else if (weaptype != WeapType_PickProbe && weaptype != WeapType_Spell && weaptype != WeapType_None)
+            if ((weapon == NULL && storage.mLastWeapon != "handToHand") 
+                || (weapon != NULL && weapon->mModel != storage.mLastWeapon))
             {
-                // All other WeapTypes are actually weapons, so get<ESM::Weapon> is safe.
-                weapon = weaponSlot->get<ESM::Weapon>()->mBase;
-                weapRange = weapon->mData.mReach;
+                storage.updateWeaponAttackDurations(actor, weapType, weapon == NULL ? 1.f : weapon->mData.mSpeed, ESM::RT_Self);
```
scrawl: I actually had a different idea for dealing with the attack durations problem. What if we pass the character controller to AiPackage::execute, ask the CharacterController about the current completion % of the weapon animation, then stop the attack input when arriving at the desired %?

That would get rid of all this duplicated logic, and make the AI code much simpler.

scrawl: The basic idea is to bring AI and character controller closer together, so instead of inspecting the world state and making second guesses about what's going on, the AI can inspect itself.

mrcheko: Well, I don't  see painless way how to pass CharacterController in AiCombat start triggers even in MWMechanics::Actors but there are also some triggers in MechanicsManager. As I can think of it now we will anyway have to inspect world state to get that. And then in all startCombat calls we'll have to firstly get somewhere CharacterController, so it also will give some pain.

> What if we... ask the CharacterController about the current completion % of the weapon animation...

As I understand this will require some calculations in animation handler class (?), and you going to ask it every frame. And in my solution at first animation should be found in World then some it's states, but it happens rather rarely (only on actor weapon changing) so you don't have to query that each frame.

mrcheko:
It it's going to cause a lot of conflicts I think I will delay these commits then. Though I don't get why.
OSG will not be merged very soon, right?

apps/openmw/mwmechanics/aicombat.cpp
scrawl: We don't need the CharacterController in startCombat. startCombat just contructs the AiPackage, but doesn't execute it.

Btw, now I remember there were some other motivations for having access to the CharacterController.

To balance the combat AI with respect to vanilla MW ( [#1876](https://bugs.openmw.org/issues/1876) ), we need to insert a random delay between the end of the last attack and the start of the next one. As of now the AI doesn't even have sufficient knowledge about the attack process to accurately time that delay. We don't know how long the "follow" part of the attack animation is. The current timing is mostly based on guesswork.

Neither does the AI have knowledge of animation states e.g. knockdown, blocking, ... that would affect the attack timings. These are only known to the CharacterController.

> As I understand this will require some calculations in animation handler class (?), and you going to ask it every frame. 

The animation handler already does that calculation every frame anyway, since it needs to know when to start the next animation. See here: https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/character.cpp#L1117

scrawl:
> Though I don't get why there will be conflicts.

Had to make more changes than I'd liked to get rid of the Ogre vectors and math.

apps/openmw/mwmechanics/aicombat.cpp
mrcheko: > ...startCombat just contructs the AiPackage, but doesn't execute it

Anyway there will be rather unpleasant extension of AiSequence::execute, where you'll need to check if package is AiCombat and perform search for that CharacterController on every execution, so specifically for this maybe better would be to queue world state to find CharacterController in Aicombat constructor.
For bug (**#1876**) we can add whole animation duration, but also we can assume the follow part is not that significant in terms of time and just naively increase a period between them.
I agree that in terms of flexibility your approach is better, but those knockouts and blocks affect actual combat minimally.
You said previously
>That would get rid of all this duplicated logic, and make the AI code much simpler.

We anyway will need to write a somewhat similar to **updateWeaponAttackDurations** function which should decide if weapon group's animation is currently playing and if yes: check if it is min-to-max attack (and check complete % here) or pre-/post- min-to-max attack; if no - than increment timer till  next attack start. Also handle here if we are unconsious or blocking etc.

So IMO the only direct reason to move to CharacterController as of now - is adding random delay between actual end of attack animation and start of a new one. And there is more simple solution (but no so exact).

scrawl: > Anyway there will be rather unpleasant extension of AiSequence::execute, where you'll need to check if package is AiCombat and perform search for that CharacterController on every execution, so specifically for this maybe better would be to queue world state to find CharacterController in Aicombat constructor.

You can't assign a CharacterController in the AiCombat constructor. AiCombat just holds the package information, but can't store any information about the active cells. The AiCombat package persist even when the actor leaves the active cells (at least for the moment, which I plan on reworking for #2172).

As for passing the characterController to execute, that's simple:

<pre>
void AiSequence::execute (const MWWorld::Ptr& actor, AiState& state, CharacterController* character, float duration)
...
            if (package->execute (actor,state,character,duration))
</pre>

Actors::update:
<pre>
                            if (isConscious(iter->first))
                                stats.getAiSequence().execute(iter->first,iter->second->getAiState(), iter->second->getCharacterController(), duration);
</pre>

From what I see the entire getMinMaxAttackDurations function is then not necessary. That's a whole 70 lines of duplicated logic removed. Sizable improvement, IMO.

mrcheko: >if (package->execute (actor,state,character,duration))

It's not so graceful IMO since only AiCombat needs CharacterController (I doubt it will be needed in other packages), but it's a minimal thing. Also CharacterController will require some inner changes to explicitely handle "release" animation text key for spellcasting and "hit" for creatures attack animations.
But maybe indeed code will become more clear, so I'll try to implement your approach.

scrawl: > Also CharacterController will require some inner changes to explicitely handle "release" animation text key for spellcasting and "hit" for creatures attack animations.

In the [osg branch](https://github.com/scrawl/openmw/blob/osg/apps/openmw/mwmechanics/character.cpp#L724) I've already moved all text key handling into the characterController, because it seemed the logical thing to do.

mrcheko: I thought about that, too. As I see you haven't started to use handleTextKey from characterController yet but only moved it there?

scrawl: It *is* used, via a listener interface (Animation:.setTextKeyListener).

mrcheko: Thanks, I got it. So taking into account how different is now ogre omw and osg one how am I supposed to commit changes which are located just on "the edge of differences"?:) Or should I wait for some merge with osg branch first?

scrawl: Good question. I'd say either we wait, or you base your changes on the osg branch.

mrcheko: Maybe it will be better to merge current commits in and after merge with OSG I'll rework code concerning our characterController discussion? It then will require resolving conflicts but as I understand - just to switch from Ogre::vector3 to OSG vector.

scrawl: I made some progress here: https://github.com/OpenMW/openmw/compare/master...scrawl:aicombat 
Very happy with the results so far. Side note: a combat AI with no delays between attacks is absolutely vicious. I need to add the fCombatDelayCreature/Npc next.

mrcheko: Sorry, I missed last changes to aicombat. Does your system already work for all types of weapons and spells?

scrawl: Yes, this works for all weapons and spells.

scrawl:
Can you submit a separate pull request for https://github.com/OpenMW/openmw/commit/6f0c8bf33a168884bbd19a9c1d3358c2e50a31b4 ?

mrcheko:
Yes, I will submit it shortly

mrcheko:
Hm, I think I'll better setup OSG normally (still didn't do that) and re-check that commit after. Also what about other changes from this request? There are other fixes and code reworkings which should be adopted (even taking into account future changes in your combat branch).

scrawl:
Sure, you can resubmit the rest as well. Though perhaps it's best to submit https://github.com/OpenMW/openmw/commit/6f0c8bf33a168884bbd19a9c1d3358c2e50a31b4 separately as it's a rather uncontroversial fix - the other commits are more complicated and I haven't fully reviewed them yet. With separate PRs we can merge the target height fix sooner, which would be good as the issue is kinda urgent.

mrcheko:
I tried 6f0c8bf alone - it gives some good effect (for guards to kill rats & mudcrabs eventually) but it's rather small actually. I believe the problem is in the PhysicsSystem::getHitContact function, exactly - with btConeShape detection shape. And I can't resolve these things on mechanics level :( I think we need to ask for help from our physics programmers.

dteviot:
https://github.com/mrcheko/openmw/commit/cec6201b12fe033a7ebd6a11eed64460dd3c2c46 is no longer needed.  I independently discovered this and a fix has been checked into trunk. -> https://github.com/OpenMW/openmw/commit/4256e151b131453ca045fae270565e6a71e656b9

mrcheko:
A lot has changed since then.

PR #616, opened 2015-06-13
SoundGens verifier (Feature #1999)

smbas:
Checks:
* Is the sound exist
* Is the creature exist (if it's specified)
* Is the creature of the type Type_Creature (if it's specified)

Example:

soundgencheck

### DIFF
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -41,7 +41,7 @@ opencs_units (model/tools
 opencs_units_noqt (model/tools
     mandatoryid skillcheck classcheck factioncheck racecheck soundcheck regioncheck
     birthsigncheck spellcheck referencecheck referenceablecheck scriptcheck bodypartcheck
-    startscriptcheck search searchoperation searchstage pathgridcheck
+    startscriptcheck search searchoperation searchstage pathgridcheck soundgencheck
     )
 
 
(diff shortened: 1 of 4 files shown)


smbas:
Add check:
* Is the sound specified

scrawl:
SoundGen.mCreature doesn't need to be a valid creature record. It can also be a string specified in a creature record's mOriginal field.

For example:

SoundGen.mCreature = "slaughterfish"

Creature: slaughterfish_big
mOriginal = "slaughterfish"

Creature: slaughterfish_small
mOriginal = "slaughterfish"

See the implementation in [OpenMW](https://github.com/scrawl/openmw/blob/master/apps/openmw/mwclass/creature.cpp#L674).

smbas:
So, mOriginal may contain an ordinary string that isn't a creature ID. Right?

smbas:
Perhaps, we need also a check for a original creature. For example, if SoundGen.mCreature = "slaughterfish_big", the checker informs that this creature has the original "slaughterfish" and it must be specified instead.

zinnschlag:
There is a slaughterfish ID in my Morrowind.esm file. If the original field specifies an original creature, this implies that there is an original creature ID. Do we actually know of any use cases for a non-existent original creature or if that works in vanilla? If not I suggest we go for the strict version, because otherwise that would complicate things quite a bit on the OpenMW-CS side (we currently have no support for a string field that may or may not be an ID).

Also, when looking at the OpenMW code it seems the current implementation would not allow to specify a separate sound generator for a derived creature. Again, I don't know if vanilla does the same, but it sounds a bit unintuitive and even if that is the case allowing such a specialisation would be a natural expansion that should not break anything.

smbas:
The checker doesn't find any non-creature ID, when checks the Morrowind.esm.

scrawl:
Ok, then let's keep it simple for now. If the verifier doesn't find any false positives in official ESM's, then I'm ok with merging this PR as-is.

PR #624, opened 2015-06-16
Enum- and DataDisplayDelegate fixes

smbas:
Changes:
* Text in Modified and Record Type columns is always within the table cell

**Before:**
before
**Now:**
now

* Proper resizing of this columns (the content size + the size of the combo box arrow).

I don't know are there any problems with this on Linux or OSX. So, test it!


zinnschlag:
Works on Linux. Will merge.

PR #625, opened 2015-06-16
Color picking (Feature #872)

smbas:
color_picking

The color picker is a pop-up window. The rectangle's color on the button is changed simultaneously with the change of color in the Color Picker. Changes are applied when the pop-up is closed (a dialog subview) or when the selected cell loose the focus (a table subview).

**Need testing on Linux and OSX, particularly: the pop-up layout and position.**

**Potential issues:**
No ability to cancel your choice in a dialog subview (except for Undo). In a table you can close the editor using Escape key and the changes won't be applied (first press closes the pop-up, second - the editor).


zinnschlag:
Mostly good. Tested on Linux Mint (Mate Desktop). Position and layout is fine.

Regarding the closing feature: Maybe add some kind of close icon or button to the popup window?

I noticed one problem (only happening in tables): It required actually 3 clicks to get the popup window to appear. The first 2 highlight the cell for editing and the 3rd produces the window.

smbas:
What behavior do you want for the close button: accept the color or cancel it? Perhaps, it'll be better to add Ok (or Select) button to accept the color. In this case, the editor color (rect) will be changed only after Ok button is pressed (not simultaneously with the color picker).

Regarding the problem you've mentioned: Do you want the pop-up will appear immediately after double click (editor creation)?

zinnschlag:
> What behavior do you want for the close button: accept the color or cancel it? Perhaps, it'll be better to add Ok (or Select) button to accept the color. In this case, the editor color (rect) will be changed only after Ok button is pressed (not simultaneously with the color picker).

The OK variant is against the general design philosophy of OpenMW-CS (which is borrowed from GNOME) about input taking effect without an okay-step.

I suggest to only add a close function that will accept any changes already made.

> Regarding the problem you've mentioned: Do you want the pop-up will appear immediately after double click (editor creation)?

Yes. That's how everything else works. We should try to stay consistent here.

smbas:
> The OK variant is against the general design philosophy of OpenMW-CS (which is borrowed from GNOME) about input taking effect without an okay-step.

> I suggest to only add a close function that will accept any changes already made.

In this case, I think the current behavior (accept on close) is good, and additional close button will be confusing.

And I'll try to implement the needed behavior for the pop-up in tables (show immediately after the editor creation).

smbas:
The pop-up is now shown after a double click on the table cell. Check it on Linux.

zinnschlag:
Works for me.

PR #628, opened 2015-06-19
Creation bar hints (Feature #1591)

smbas:
I've added an ID completion to creators for the following tables:
* Instances (Cell ID)
* Topic Infos (Topic ID)
* Journal Infos (Journal ID)

**Example:**
creation_bar_completion

Also this PR contains a fix for the creation bar of Dialogue subview in Info tables. The short description of the relevant bug: when you enter the correct Topic/Journal ID in the creation bar, Create button doesn't become available.

**P.S.:** I hope I have correctly understood the task.


zinnschlag:
This is not related to 1591 at all, which was meant for the respective fields to act as a drop target for drag & drop operations (e.g. drop a cell from the cell table into the cell field of the instance creation bar). But it is a worthwhile improvement nevertheless. Will merge.

PR #632, opened 2015-06-24
OpenCS: improvements for drag'n'drop

smbas:
What was done:
* Rework DropLineEdit to accept drops of a specific type (Display)
* Drag events are accepted only for proper cells in tables views (a plus is appeared on right columns)
* Nested tables accept drops

\+ minor fixes.

Check if there are any issues.


zinnschlag:
Looks okay. I will make one minor adjustment and then merge.

PR #633, opened 2015-06-25
Drop targets in creation bars

smbas:
The creation bars for Instances, Topic Infos and Journal Infos accept drops.


zinnschlag:
One problem: Dragging an ID with non-ASCII characters to one of the bars produces garbage characters. I guess the conversion procedure between QString und std::string is missing somewhere? Maybe somewhere within the recent drag and drop changes? At least I don't remember it happening before, but I could be wrong about that.

zinnschlag:
Additional info: The problem does not show up in drags that have other targets than the creation bars (tested script editor and filter bar).

smbas:
You were right about the conversion. The previous code (before drag'n'drop changes) also contained this bug.

PR #635, opened 2015-06-25
Fixes for Info tables (mostly for their sorting)

smbas:
What was done:
* Fix the incorrect position of a newly added row after the sorting
* Update the first row cache only after rows' insertion/deletion
* Ignore the letter case of topics in rows' reordering (Bug #2714)

**One issue is not resolved (for Info tables sorting)**: addition of a new row ignores the sort order, so the row is placed at the incorrect position (the model must be manually re-sorted to fix this). One obvious solution is to do the re-sorting automatically after a new row is inserted. But it breaks the "Jump to Added Record" behavior. Any suggestions for how to fix this?

### DIFF
--- a/apps/opencs/model/world/infocollection.cpp
+++ b/apps/opencs/model/world/infocollection.cpp
@@ -97,7 +97,8 @@ bool CSMWorld::InfoCollection::reorderRows (int baseIndex, const std::vector<int
         return false;
 
     // Check that topics match
-    if (getRecord (baseIndex).get().mTopicId!=getRecord (lastIndex).get().mTopicId)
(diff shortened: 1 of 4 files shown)


zinnschlag:
Difficult. Can we first re-sort and then jump to the added record? Not sure if that would be feasible.

That is a serious problem. Of course we have the same behaviour with all the other tables (a general fix would certainly be welcome), but for the info tables it actually breaks the visualisation of the data structure.
If we can't find an immediate solution the issue should be added to the tracker.
I will go ahead and merge anyway (doesn't break anything that wasn't broken before).

scrawl:
Please don't use QString::toStdString(). That won't give you an utf-8 string. Instead use
<pre>
toUtf8().constData()
</pre>

PR #637, opened 2015-06-28
Expand AiWander by 50% (Fixes #2726)

dteviot:
When AiWander distance has < 2 waypoints, expand distance by no more than 50%.

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -708,12 +708,21 @@ namespace MWMechanics
 
             // mAllowedNodes for this actor with pathgrid point indexes based on mDistance
             // NOTE: mPoints and mAllowedNodes are in local co-ordinates
-            for(unsigned int counter = 0; counter < pathgrid->mPoints.size(); counter++)
+            pointsWithinDistance(npcPos, pathgrid, mDistance);
+
+            // In vanilla Morrowind, sometimes distance is too small to include at least two points. (Bug #1317)
+            // in which case, see if we can get two points by increasing wander distance by 50%.
+            // Special cases we need to deal with. (Bug #2726)
+            // 1. If no waypoint nearby, don't increase search range.
+            // 2. As waypoints are usually >= 256 units appart, When wander distance is < 256, 
+            //    then want NPC to move a few steps around initial position rather than follow waypoints.
+            //    So don't expand seach range
+            const int minExpandWanderDistance = 256;
+            if ((mAllowedNodes.size() == 1) && (minExpandWanderDistance <= mDistance))
             {
-                Ogre::Vector3 nodePos(PathFinder::MakeOgreVector3(pathgrid->mPoints[counter]));
-                if(npcPos.squaredDistance(nodePos) <= mDistance * mDistance)
-                    mAllowedNodes.push_back(pathgrid->mPoints[counter]);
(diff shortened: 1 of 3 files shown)


scrawl:
Yeah, I'm not fond of the expanding idea. The correct fix is to support wandering with no pathgrid nodes. This PR is much better than the current behaviour though (not sure how that slipped past review). So let's merge it.

dteviot:
@cc9ii, @scrawl 
> FWIW, I don't believe vanilla works this way at all. i.e. AIWander does not require pathgrid points.

I'm aware of this.  I'm working on what to do when there are no points, but it's more tricky.  As scrawl says, this is at least an improvement over the existing behaviour.

TorbenCarrington:
This is not correct. Expansion may feel nice for now but it is entirely a
band-aid solution and I don't think merging would be a good idea.
Wandering with no pathgrid nodes was an issue involved in the pathgrid
format itself if I remember correctly, which is why I didn't implement it.
The true pathgrid also creates nodes for the initial position of each NPC
on the map and also allows for stopping in between path nods. Meaning
everywhere there is a line between nodes, the NPC can go there as a final
destination with AIWander. As well as anywhere (even places not linked to
with path nodes) that an NPC was placed on the ground when initially
loading the cell, is also a usable destination for building a path.

I am heavily against merging this, but it is not up to me.

TorbenCarrington:
@cc9cii 
Actually, creatures do rely on pathgrids in vanilla if I remember correctly. Morrowind generates a blanket pathgrid for exterior cells that do not have one, this pathgrid is very basic and just has holes for static objects and water (being the reason NPCs and creatures in the wild walk around like idiots a lot of the time).

If we correct the pathgrid generation in OpenMW and adjust AIWander to work with it, the current implementation would be vanilla accurate.

dteviot:
https://forum.openmw.org/viewtopic.php?f=6&t=2970
Created thread to continue discussion, so we can have a permanent, more easily findable record of my screw-up.

zinnschlag:
@scrawl What is your opinion about TorbenCarrington's comments. Should we go ahead with the workaround anyway or wait for the discussion to continue?

scrawl:
> Expansion may feel nice for now but it is entirely a
band-aid solution and I don't think merging would be a good idea.

This PR did not introduce the range expansion problem. It's already in master now. Therefore I stick to my previous statement:

> This PR is much better than the current behaviour though

cc9cii:
> Morrowind generates a blanket pathgrid for exterior cells that do not have one, this pathgrid is very basic and just has holes for static objects and water 

@TorbenCarrington do you have any more information about this?  I don't believe vanilla does this.

zinnschlag:
Right. This does not merge cleanly though.

TorbenCarrington:
I don't :/ It has been a year since I worked on the AI. Perhaps Gus would
know more.

dteviot:
@scrawl
Conflict is Scrawls OSG work.  (Big surprise.)
I'd suggest I just cancel this PR and do a new one that just rips out the code I put in for #1317. This will fix #2726 and reintroduce #1317. (IMHO, 2716 is a more serious bug than 1317)  I can then supply a new fix for #1317, as proposed in thread https://forum.openmw.org/viewtopic.php?f=6&t=2970.
Question: Does the proposal on the thread sound like the correct way to fix #1317?

scrawl:
#648

PR #638, opened 2015-06-28
OpenSceneGraph port

scrawl:
See https://bugs.openmw.org/projects/openmw/roadmap#openscenegraph and https://openmw.org/2015/openscenegraph-port-playable for the list of changes.

Since that announcement, I've fixed the worst bugs and got some more performance improvements in.

I think it's about time to merge to the mainline, as there are more and more merge conflicts popping up on seemingly unrelated code, e.g. https://github.com/OpenMW/openmw/pull/614.

Known issues:
- Terrain could perform worse on low-end graphics cards
- OpenCS cell markers and object selection not ported yet

Build environment changes:
https://wiki.openmw.org/index.php?title=Development_Environment_Setup_OSG


scrawl:
The line delta as per github is incorrect. It doesn't seem to account for added and removed files. The correct one is:

<pre>
git diff upstream/master..osg --shortstat
 713 files changed, 25430 insertions(+), 48078 deletions(-)
</pre>

zinnschlag:
Lol, we are breaking github with mega pull requests.

Anyway, I am going to put this PR on hold for 2 days. Currently I am still on Mint 13, which means I would have to build OSG myself. Which is of course not a problem, but since my new computer (Mint 17.1 with OSG available out of the box, I think) is scheduled to arrive tomorrow afternoon, this would be a waste of time.
Sorry for the delay.

btw. the first link in the original comment does not work for me.

psi29a:
You can always do a rebase -i to squash

zinnschlag:
Around 200-240 fps in Balmora with a few places dropping down to 180. Not bad.

I found a couple of issues, but none are bad enough to be a blocker for the merge:
- When firing up OpenMW the whole screen gets really bright (gamma, I presume). Even the OpenMW window itself is way too bright.
- In OpenMW-CS a few signal-slot connections are broken (triggers error messages at runtime)
- The direction of the scrollwheel in OpenMW-CS scene subviews is inverted.
- When closing OpenMW-CS I get a segfault (haven't gotten around yet to see if it is repeatable).

Will merge in a bit.

zinnschlag:
btw. what about the openscenegraph roadmap on the tracker? Should we merge it into openmw-0.37 or would you like to keep it separate for now?

zinnschlag:
The segfault is reproducable (close OpenMW-CS with a scene subview open). Here is the backtrace: https://gist.github.com/zinnschlag/9eeb6a419423edad3a90

scrawl:
I knew about this segfault. The gist of it is that we need to explicitely close scene views before the editor closes. Would that be doable?
Alternatively I could wrap the resource managers into a shared_ptr, but that's a bit more work.

scrawl:
> When firing up OpenMW the whole screen gets really bright (gamma, I presume). Even the OpenMW window itself is way too bright.

Do you have a gamma = line in your settings.cfg? AFAIK setting the gamma on linux is currently broken because of an SDL bug. Perhaps it is miraculously working on your machine.

> The direction of the scrollwheel in OpenMW-CS scene subviews is inverted.

The convention is adopted from flight simulators. I don't see an easy way to change this. I would expect that we will eventually subclass from the TrackballManipulator to add custom functionality anyway.

scrawl:
>btw. what about the openscenegraph roadmap on the tracker? Should we merge it into openmw-0.37

Done.

zinnschlag:
> I knew about this segfault. The gist of it is that we need to explicitely close scene views before the editor closes. Would that be doable?
> Alternatively I could wrap the resource managers into a shared_ptr, but that's a bit more work.

IIRC the subviews are closed via deleteLater (necessary to avoid some other issues). We can't really get any more control over that without rewriting a lot of code. I think your second option would actually be less work.

zinnschlag:
> Do you have a gamma = line in your settings.cfg? AFAIK setting the gamma on linux is currently broken because of an SDL bug. Perhaps it is miraculously working on your machine.

I have 

> gamma = 1.00

Removing it does not change anything.

zinnschlag:
> The convention is adopted from flight simulators. I don't see an easy way to change this. I would expect that we will eventually subclass from the TrackballManipulator to add custom functionality anyway.

The way it is currently working is highly counter-intuitive. That needs fixing before 1.0 one way or the other.

scrawl:
> I have
> 
> gamma = 1.00
> Removing it does not change anything.

Any difference if you comment out the line https://github.com/OpenMW/openmw/blob/master/components/sdlutil/sdlvideowrapper.cpp#L67 ?

zinnschlag:
Yes, that fixed it.

scrawl:
Well, it isn't supposed to go into that code path unless you are using non-default gamma or contrast settings. Do you have a 
<pre>
contrast =
</pre>
line in the settings file?

If you set both settings to 1 (which is the default), the screen settings shouldn't be touched.

kcat:
> If you set both settings to 1 (which is the default), the screen settings shouldn't be touched.

Even if the system doesn't have 1 as the default?

scrawl:
> Even if the system doesn't have 1 as the default?

Yes. The intention is to leave settings alone unless the user explicitely requests a change.

<pre>
        if (gamma == mGamma && contrast == mContrast)
            return;
</pre>

kcat:
That seems pretty kludgy, as there could be cases where the user sets 1 while the system normally has it at something else, and they want it to change in game. Having to manually go into the options every time to move it away from one and back would be pretty annoying.

zinnschlag:
Found the problem. I had gamma in two sections (General and Video). Probably a leftover from an earlier version? The one in Video was 2.2. Once I set it to 1.0 everything worked as expected.

Still, shouldn't we disable that for Linux (at least when in windowed mode)? OpenMW really shouldn't affect anything beyond its own window.

PR #640, opened 2015-06-29
Some bugfixes

smbas:
A drag event must be ignored whenever it isn't accepted to prevent drops on non-editable table fields.


scrawl:
Could you do a rebase please to remove the extraneous merge commits?

smbas:
Fix for the bug #2734. Since my last commit (for drag'n'drop) haven't been merged, I have to add this fix here.

PR #641, opened 2015-06-30
Save file versioning

scrawl:
The first commit introduces the version field for savegames. The second commit makes use of this field by going down a more efficient code path for non-legacy files. Improves save file loading speed by 50% in my tests.

### DIFF
--- a/apps/essimporter/importer.cpp
+++ b/apps/essimporter/importer.cpp
@@ -327,7 +327,7 @@ namespace ESSImport
 
         ESM::ESMWriter writer;
 
-        writer.setFormat (ESM::Header::CurrentFormat);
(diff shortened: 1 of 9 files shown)


zinnschlag:
Sounds reasonable. But I don't like the use of the version field. We decided to not use it in content files (because the weirdness of the version field in the original format). If we make use of it now in the saved game file, we will end up with two version identifiers (version and format). I suggest to use the format field instead (not supported yet in writer I believe, but it is implemented in reader).

zinnschlag:
Also, we need to a test when loading saved game files that rejects files cleanly that have a format value larger then the last one we know.

scrawl:
Done and done.

We already had a check in place to hide savegames of too new format from the UI. Now they will show, but you get an appropriate error on load.

PR #642, opened 2015-06-30
Android build fixes

sandstranger:


scrawl:
Is this actually working yet? I was under the impression android needs shaders to work. I don't see the point merging this in if the port isn't working yet.

sandstranger:
game works, but shows only a gray screen, because there is no shaders .

scrawl:
Regarding the cursor creation issue: the only reason we are using a new graphics context to convert the cursors, is needing to unpack the cursor from compressed DDS format to a bitmap that SDL can use. I believe you mentioned on the forum that Android can't use compressed DDS textures and you were converting them all anyway. That makes the cursor manager in its current form unnecessary for Android.

sandstranger:
About compressed textures .
To unpack textures to memory,  need to cut out code   in osg about texture compression and always use
const osg :: Image & image
GLint internalFormat = image.getInternalTextureFormat ();

sandstranger:
I can not test it yet , because my videocard not support dxt compression ,and when I try to load rgba textures openmw crashes

Svetomech:
>because my videocard not support dxt compression

At all? Because I was sure my didn't too but [GLtools](https://play.google.com/store/apps/details?id=com.n0n3m4.gltools#) proved me wrong.

sandstranger:
http://www.arm.com/products/multimedia/mali-performance-efficient-graphics/mali-t760.php
mali t760 support only etc1, etc2,ASTC texture compression

sandstranger:
I have no idea will it work or not .I take a long break at this development, because I'm tired again

PR #643, opened 2015-07-01
Fix build error and a pair of warnings

ace13:
The error was about casting between osg::Callback* and osg::NodeCallback*
The warnings are both about virtual classes with non-virtual destructors

### DIFF
--- a/apps/openmw/mwrender/animation.hpp
+++ b/apps/openmw/mwrender/animation.hpp
@@ -93,7 +93,6 @@ protected:
         boost::shared_ptr<float> mTimePtr;
 
     public:
-        virtual ~AnimationTime() { }
 
         void setTimePtr(boost::shared_ptr<float> time)
         { mTimePtr = time; }
--- a/apps/openmw/mwrender/animation.hpp
+++ b/apps/openmw/mwrender/animation.hpp
@@ -93,6 +93,7 @@ protected:
         boost::shared_ptr<float> mTimePtr;
 
     public:
+        virtual ~AnimationTime() { }
 
         void setTimePtr(boost::shared_ptr<float> time)
         { mTimePtr = time; }
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -721,12 +721,12 @@ public:
                 if (stateset->getAttribute(osg::StateAttribute::MATERIAL))
                 {
                     SceneUtil::CompositeStateSetUpdater* composite = NULL;
-                    osg::NodeCallback* callback = node.getUpdateCallback();
+                    osg::NodeCallback* callback = dynamic_cast<osg::NodeCallback*>(node.getUpdateCallback());
                     while (callback)
                     {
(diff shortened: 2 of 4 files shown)


scrawl:
Are you using osg-svn? We use 3.2.0.

The virtual destructor addition is ok.

ace13:
Apparently I'm using OSG 3.3.8.
I guess I could sit down and recompile new windows dependencies against OSG 3.2.2 if necessary, would rather not have to do that though since I'm away from my desktop for a week or two.

Does the change break anything on OSG 3.2.0?

scrawl:
> Does the change break anything on OSG 3.2.0?

No, but I would rather avoid the dynamic_casts.

I have fixed this error before. See sceneutil/controller.cpp:

<pre>
#if OSG_MIN_VERSION_REQUIRED(3,3,3)
        osg::Callback* callback = node.getUpdateCallback();
#else
        osg::NodeCallback* callback = node.getUpdateCallback();
#endif
</pre>

> I guess I could sit down and recompile new windows dependencies against OSG 3.2.2 if necessary, would rather not have to do that though since I'm away from my desktop for a week or two.

If (when) you do that, just out of curiosity, are your current OSG builds using [single or double precision](https://wiki.openmw.org/index.php?title=Development_Environment_Setup#OpenSceneGraph)?

> The virtual destructor addition is ok.

Looking at this again, shouldn't the virtual destructor be added to SceneUtil::ControllerSource?

ace13:
My current builds might have double-precision, but it's hard to say since I've rebuilt another version of OSG aftwards to try and fix XP compatability.

> Looking at this again, shouldn't the virtual destructor be added to SceneUtil::ControllerSource?

Really any class with a virtual function should have a virtual destructor, since the default destructor is non-virtual.
But those two were the only ones that gave me compiler warnings for it.

scrawl:
> Really any class with a virtual function should have a virtual destructor, since the default destructor is non-virtual.
> But those two were the only ones that gave me compiler warnings for it.

The two classes you've fixed are both derived from SceneUtil::ControllerSource. Wouldn't adding the virtual destructor to the base class also fix the warning?

ace13:
Both of those classes contain member variables with non-trivial destructors; `boost::shared_ptr<float>` for AnimationTime and `std::string` for WeaponAnimationTime.
So if any of them gets subclassed then those two variables will not be destroyed properly, unless there's a virtual destructor.

Adding a virtual destructor to SceneUtil::ControllerSource would be a good idea, but it wouldn't solve the warning.

kcat:
> Adding a virtual destructor to SceneUtil::ControllerSource would be a good idea, but it wouldn't solve the warning.

When a class has a virtual destructor, all child classes derived from it will implicitly have a virtual destructor.

ace13:
> When a class has a virtual destructor, all child classes derived from it will implicitly have a virtual destructor.

Interesting, I did not actually know that.
I've always been taught to always put virtual destructors in all classes with virtual functions.

ace13:
Well, finally had the time to sit down and fix this properly.

PR #646, opened 2015-07-03
Edit 'ID' in the dialogue sub-views (Feature #2565)

smbas:
LineEdits and Labels (that contain IDs) have an Edit 'ID' action in their context menus.
context_menu_label
context_menu_lineedit

**And a question:** What about nested tables?


zinnschlag:
Yes, we also need this feature in nested tables.

smbas:
context_menu_nested_tables

PR #647, opened 2015-07-03
Configuration of Extended commands (Feature #1463)

smbas:
extended_panel


zinnschlag:
Looks mostly okay. Issues:

- The wording in the user settings dialogue is a bit strange. And it needs a tooltip. Maybe "Ask user for details when performing an extended delete/revert" ? And for the tooltip "Delete and revert have an extended form that also affects associated records. If this option is enabled OpenMW-CS will ask the user which of these should be affected. Otherwise all will be affected without additional user input."
This is just a suggestion. Feel free to refine it or write something else.

- The panel should be closeable via escape.

- When deleting the respective record while the panel is open, the panel should close automatically.

- The layout could use a bit of refinement. Too much vertical border. I think the checkboxes should be aligned to the left. And the checkboxes also do not seem to be vertically aligned with the buttons.

smbas:
About auto-closing: What to do when several records are selected? Must the panel be closed after deletion of ALL records? But in this case, what to do when one record is deleted: remove it from a command dispatcher (the associated records won't be deleted)?

smbas:
Also about the user setting:
**Name:** 
Manually specify affected record types for an extended delete/revert.
**Tooltip:** 
Delete and revert commands have an extended form that also affects associated records.
If this option is enabled, types of affected records are selected manually before a command execution. Otherwise, all associated records are deleted/reverted immediately.

zinnschlag:
> About auto-closing: What to do when several records are selected? Must the panel be closed after deletion of ALL records? But in this case, what to do when one record is deleted: remove it from a command dispatcher (the associated records won't be deleted)?

I suggest to only close the panel when no applicable records are left. Note that this is not the same as no records being left.

Rewording suggestion is fine.

smbas:
extended_panel_2

scrawl:
@zinnschlag, any comments? This PR has been sitting here for quite a while.

zinnschlag:
Must have fallen through the cracks. The github news feed has more holes than a piece of Swiss cheese. Really annoying.

Anyway, looks good. Will merge.

PR #648, opened 2015-07-04
Removed fix for #1317 (Fixes #2726)

dteviot:
Fix for #1317 created bigger problems than it fixed. So removing.
Also, some minor code improvements.


zinnschlag:
Shall we reopen #1317 then?

PR #653, opened 2015-07-07
AiWander uses points between PathGrid points (Fixes #1317)

dteviot:
When there is only on PathGrid point within a NPC's wander distance, expand possible wander destinations by using positions between PathGrid points.
Also, assorted code re-factoring.

### DIFF
--- a/apps/openmw/main.cpp
+++ b/apps/openmw/main.cpp
@@ -14,7 +14,9 @@
 
 #if defined(_WIN32)
 // For OutputDebugString
+#ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
+#endif
 #include <windows.h>
 // makes __argc and __argv available on windows
 #include <cstdlib>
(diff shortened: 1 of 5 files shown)


dteviot:
> Any particular reason that tolerance parameter was removed? I believe the original intent was to prevent actors crowding on pathgrid points

When I did a blame, history showed the value of 64 was introduced by me.  Something that leaked in when I was trying to fix the "Spinning Netches" problem. (Which occurred because the flying netches could not get close enough to the nav points.)  I reduced the value to same one used everywhere else because if tolerance is 64, then NPCs with a wander of 64 (or less) don't move.

scrawl:
> Proposed fix:
> 1. Put that tolerance back,
> 2. Set minimum wander distance to 128 units.

1. Sure. 2., what is that supposed to achieve?

dteviot:
>  2., what is that supposed to achieve?

By minimum wander distance, I'm referring to the threshold set in line 743 of AddPointBetweenPathGridPoints() in the PR. https://github.com/dteviot/openmw/commit/1239667cb415835d2d4cb6f3aba941c92e4fecaa.
(Obviously, I need to rename this variable from "threshold" to "MinimumWanderDistance")

If the tolerance on mPathFinder.checkPathCompleted() is set to 64, and wander distance for an actor is 64 or less the actor won't wander, because when mPathFinder.checkPathCompleted() is called next frame, actor decides it's trip has completed, before actor has moved.

PR #656, opened 2015-07-09
Update boost in travis-ci to fix build failure

maqifrnswa:


psi29a:
Nice work! :)

scrawl:
Is boost::thread the only lib that's throwing errors? If so, we could move to std::thread as Chris suggested.

maqifrnswa:
I can't tell since it fails so early. Maybe things with shared_ptr might throw errors also - so you can use the new std::shared_ptr. I'm all for using std instead of boost

zinnschlag:
I already switched the use of boost shared_ptr to the standard ones in c++11. If anyone wants to port over thread too, be my guest.

psi29a:
Nice, so we'll have less boost dependency?  (yay for packaging perspective)

zinnschlag:
I think boost::thread is using boost::shared_ptr internally, so we are actually not losing any dependencies yet. But as I wrote, it is a possibility, if anyone wants to put the time in.

scrawl:
> I can't tell since it fails so early. Maybe things with shared_ptr might throw errors also - so you can use the new std::shared_ptr. I'm all for using std instead of boost

boost::filesystem is throwing similar errors, so looks like we'll still need the newer boost.
<pre>
/usr/include/boost/type_traits/is_convertible.hpp:135:76: error: call to
      implicitly-deleted copy constructor of
      'boost::filesystem3::directory_iterator'
  ...const value = sizeof( boost::detail::checker<To>::_m_check(_m_from, 0) )
                                                                ^~~~~~~
/usr/include/boost/type_traits/is_convertible.hpp:298:33: note: in instantiation
      of template class
      'boost::detail::is_convertible_basic_impl<boost::filesystem3::directory_iterator
      &, boost::filesystem3::directory_iterator>' requested here
               ::boost::detail::is_convertible_basic_impl<ref_type,To>::value,
                                ^
/usr/include/boost/config/suffix.hpp:422:72: note: expanded from macro
      'BOOST_STATIC_CONSTANT'
#     define BOOST_STATIC_CONSTANT(type, assignment) static const type assignment
                                                                       ^
/usr/include/boost/type_traits/is_convertible.hpp:371:13: note: in instantiation
      of template class
      'boost::detail::is_convertible_impl<boost::filesystem3::directory_iterator,
      boost::filesystem3::directory_iterator>' requested here
   : public is_convertible_impl_dispatch_base<From, To>::type
            ^
/usr/include/boost/type_traits/is_convertible.hpp:418:71: note: in instantiation
      of template class
      'boost::detail::is_convertible_impl_dispatch<boost::filesystem3::directory_iterator,
      boost::filesystem3::directory_iterator>' requested here
BOOST_TT_AUX_BOOL_TRAIT_DEF2(is_convertible,From,To,(::boost::detail::is...
                                                                      ^
</pre>

scrawl:
What is the minimum version of boost that doesn't have the errors? We'll need this for our new build instructions.

maqifrnswa:
boost 1.48.0.3 and later apparently fixed this bug (http://stackoverflow.com/questions/11302758/error-while-copy-constructing-boostshared-ptr-using-c11)
> I checked, and apparently it is fixed as of version 1.48.0. But it doesn't say which revision, so I must assume that it is fixed from version 1.48.0.3 (since you tested with 1.48.0.2 and it failed, and the 1.48.0.3 is the last minor revision within 1.48.0). –  Mikael Persson Jul 3 '12 at 0:39

Boost 1.51 is the first boost to be compiled and tested against c++11 (versus c++0x).

I don't know what the minimum boost is with no errors, we could just set up a ppa and run through all of them. My guess is 1.49 would be a good starting point.

PR #658, opened 2015-07-10
Slow down simulated mouse with right trigger  in menus.

rohit-n:
EDIT: I'm reducing the scope of this PR for something more tenable. See posts below.

### DIFF
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -370,6 +370,9 @@ namespace MWInput
                 float zAxis = mInputBinder->getChannel(A_LookUpDown)->getValue()*2.0f-1.0f;
                 const MyGUI::IntSize& viewSize = MyGUI::RenderManager::getInstance().getViewSize();
 
+                xAxis *= (1.5f - mInputBinder->getChannel(A_Use)->getValue());
+                yAxis *= (1.5f - mInputBinder->getChannel(A_Use)->getValue());
+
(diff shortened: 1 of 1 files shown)


scrawl:
I'm not too fond of this implementation. From what I see you would need to change every menu in the game to add support.

My suggestion is to handle everything in WindowManager, walk over the currently visible MyGUI widgets (you don't need to talk to the individual menu classes to get those), filter out widgets that can't be interacted with, and then cycle through the remaining widgets with gamepad buttons.

rohit-n:
Thanks for the feedback.

I'm looking at the code now - I'm guessing `MyGUI::VectorWidgetPtr mListWindowRoot` is what I need to look at when iterating through widgets?

scrawl:
You can use:
<pre>
MyGUI::Gui::getInstance().getEnumerator()
</pre>

LavenderMoon:
How does the original Xbox version of Morrowind handle the inventory? I haven't played it in forever.

rohit-n:
@LavenderMoon I never played the Xbox version, but assume you just used the d-pad to move through items.

I haven't had much time to work on this solution, so I have an alternate proposal that sounds more elegant: if the player is in menu using the controller, what about using a button (maybe one of the triggers) to slow down the movement of the simulated mouse? This would make it much easier to make fine selections, which is especially important in the dialog and inventory menus. This would make controller gameplay much more enjoyable, in my opinion.

scrawl:
Sounds like a stopgap solution, but I'm fine with it.

PR #664, opened 2015-07-14
Windows build fix

Koncord:
gcc.exe (Rev5, Built by MSYS2 project) 4.9.2

### DIFF
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -123,11 +123,11 @@ target_link_libraries(openmw
     ${OPENAL_LIBRARY}
     ${SOUND_INPUT_LIBRARY}
     ${BULLET_LIBRARIES}
-    ${MYGUI_LIBRARIES}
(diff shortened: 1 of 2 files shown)


scrawl:
I feel this fix is a bit too subtle and inconsistent with what we're doing so far. The order of linking ideally shouldn't make a difference. Could you add the missing library to target_link_libraries in CMakeLists for [components](https://github.com/OpenMW/openmw/blob/master/components/CMakeLists.txt#L182) instead?

Koncord:
You mean moving ${MYGUI_LIBRARIES} from openmw to components?

scrawl:
Don't move it there, add it. Keep the main CMakeLists as it currently is in master, then add a `${MYGUI_LIBRARIES}` to components/CMakeLists.txt.

The idea is to explicitely link to everything we are using, even if that library is implicitely pulled in from somewhere else. This will reduce the chance of breakage in the future.

PR #666, opened 2015-07-15
Animation refactoring & fixes

scrawl:
For review by @kcat 

* Rename Animation::Group to Animation::BlendMask. The term group was already used elsewhere.
* Extended the animation state priority system, now supports separate priority values for each bone group.
* Refactor weapon animations to use the new priority system in place of the changeGroups hack.
* Allow attacks during block animations, continue playing the block animation on the LeftArm.


scrawl:
`02:34 <KittyCat> nothing immediately jumps out at me, though I'm not as familiar with the code these days`

PR #667, opened 2015-07-17
Add support for linking OpenSceneGraph statically and fixes for MyGUI and QT static linking

slothlife:


components/resource/texturemanager.cpp
```diff
@@ -1,11 +1,24 @@
...
+#ifdef OSG_LIBRARY_STATIC
+// This list of plugins should match with the list in the top-level CMakelists.txt.
+USE_OSGPLUGIN(png)
+USE_OSGPLUGIN(tga)
+USE_OSGPLUGIN(dds)
+USE_OSGPLUGIN(jpeg)
+
+// Sets the default windowing system interface according to the OS.
+// Necessary for OpenSceneGraph to do some things, like decompression.
+USE_GRAPHICSWINDOW()
```
scrawl: I would move this to the place where WindowingSystemInterface is actually used, i.e. in components/sdlutil/sdlcursormanager.cpp.

slothlife: OK, I was having trouble finding a place to put this that made sense.

components/sdlutil/sdlcursormanager.cpp
```diff
@@ -10,9 +10,17 @@
 #include <osg/Geometry>
 #include <osg/Texture2D>
 #include <osg/TexMat>
+#include <osgDB/Registry>
```
scrawl: One more nitpick, this include can be removed.

PR #668, opened 2015-07-17
Change color of crosshair to red when object is owned

kunesj:
This pull request makes it easy to tell if taking object will be crime or not. When pointing on owned object the crosshair turns red.

You need to add "show owned = true" under [Game] is settings.cfg for this to be enabled.

This is the first time I'm messing with OpenMW code, so I hope that it doesn't break anything...

### DIFF
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -6,8 +6,6 @@
 #include <list>
 #include <stdint.h>
 
-#include "../mwworld/ptr.hpp"
-
 namespace osg
 {
     class Vec3f;
@@ -25,6 +23,7 @@ namespace MWWorld
 {
     class Ptr;
     class CellStore;
+    class CellRef;
 }
 
 namespace Loading
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -6,6 +6,8 @@
 #include <list>
 #include <stdint.h>
 
+#include "../mwworld/ptr.hpp"
+
 namespace osg
 {
     class Vec3f;
@@ -211,6 +213,8 @@ namespace MWBase
 
             /// Has the player stolen this item from the given owner?
             virtual bool isItemStolenFrom(const std::string& itemid, const std::string& ownerid) = 0;
+            
+            virtual bool isAllowedToUse (const MWWorld::Ptr& ptr, const MWWorld::CellRef& cellref, MWWorld::Ptr& victim) = 0;
     };
 }
 
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -219,7 +219,6 @@ namespace MWBase
             virtual void unsetSelectedWeapon() = 0;
 
             virtual void showCrosshair(bool show) = 0;
-            virtual void setCrosshairOwned(bool owned) = 0;
             virtual bool getSubtitlesEnabled() = 0;
             virtual bool toggleGui() = 0;
 
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -219,6 +219,7 @@ namespace MWBase
             virtual void unsetSelectedWeapon() = 0;
 
             virtual void showCrosshair(bool show) = 0;
+            virtual void setCrosshairOwned(bool owned) = 0;
             virtual bool getSubtitlesEnabled() = 0;
             virtual bool toggleGui() = 0;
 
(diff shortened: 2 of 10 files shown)


scrawl:
Is the crosshair really the best place for this feature? Problem is, there is no crosshair in third person mode.

apps/openmw/mwbase/windowmanager.hpp
```diff
@@ -219,6 +219,7 @@ namespace MWBase
             virtual void unsetSelectedWeapon() = 0;
 
             virtual void showCrosshair(bool show) = 0;
+            virtual void setCrosshairOwned(bool owned) = 0;
```
scrawl: We could do this without interface additions to `WindowManager`. For example, update the stolen state in the WindowManager::setFocusObject function.

apps/openmw/mwgui/tooltips.cpp
```diff
@@ -342,10 +346,35 @@ namespace MWGui
...
+        {
+            MWBase::WindowManager *wm = MWBase::Environment::get().getWindowManager();
+            
+            if(!mFocusObject.isEmpty())
+            {
+                MWWorld::CellRef& cellref = mFocusObject.getCellRef();
+                MWWorld::Ptr ptr = MWBase::Environment::get().getWorld()->getPlayerPtr();
+                MWWorld::Ptr victim;
+                
+                MWMechanics::MechanicsManager* mm = new MWMechanics::MechanicsManager;
```
scrawl: You're not supposed to create new instances of MechanicsManager. Instead get the singleton instance.
<pre>
MWBase::MechanicsManager* mm = MWBase::Environment::get().getMechanicsManager();
</pre>

apps/openmw/mwgui/tooltips.cpp
```diff
@@ -17,6 +17,7 @@
 #include "../mwworld/class.hpp"
 #include "../mwworld/esmstore.hpp"
 #include "../mwmechanics/spellcasting.hpp"
+#include "../mwmechanics/mechanicsmanagerimp.hpp"
```
scrawl: You should include `../mwbase/mechanicsmanager.hpp` instead (the interface, not the implementation).

apps/openmw/mwgui/tooltips.cpp
kunesj: Are you sure? That's giving me:
```
/home/jirka642/Programming/Git/kunesj-openmw/apps/openmw/mwgui/tooltips.cpp: In member function ‘void MWGui::ToolTips::checkOwned()’:
/home/jirka642/Programming/Git/kunesj-openmw/apps/openmw/mwgui/tooltips.cpp:368:36: error: ‘class MWBase::MechanicsManager’ has no member named ‘isAllowedToUse’
                 bool allowed = mm->isAllowedToUse(ptr, cellref, victim); // 0 - owned, 1 - not owned
```

scrawl: In that case the `isAllowedToUse` function would need to be added to the public interface in MWBase::MechanicsManager.

kunesj:
OK, fixed both mechanicsManager and windowManager stuff.

>Problem is, there is no crosshair in third person mode.
I didn't think about that, since I never play in third person mode...

apps/openmw/mwbase/mechanicsmanager.hpp
```diff
@@ -6,6 +6,8 @@
 #include <list>
 #include <stdint.h>
 
+#include "../mwworld/ptr.hpp"
```
scrawl: This include is not necessary. There is already a forward declaration for MWWorld::Ptr in mechanicsmanager.hpp.

apps/openmw/mwgui/tooltips.cpp
```diff
@@ -342,10 +343,31 @@ namespace MWGui
...
 
         return tooltipSize;
     }
+    
+    bool ToolTips::checkOwned()
+    {
+        // true=owned, false=notOwned
+        if(!mFocusObject.isEmpty())
+        {
+            MWWorld::CellRef& cellref = mFocusObject.getCellRef();
```
scrawl: You can add `const` here since cellref won't be modified.

kunesj:
What about now?

scrawl:
Good for me. Just waiting for a second review by Zini and then we can merge.

kcat:
I believe the way MCP (or MGE?) does this is by adding "Owned" to the hovering tooltip. I think that may be clearer and more foolproof than tinting the crosshair red (what if the user has the crosshair disabled? or if they've modded in a redish crosshair?).

scrawl:
> I think that may be clearer and more foolproof than tinting the crosshair red (what if the user has the crosshair disabled? or if they've modded in a redish crosshair?).

I really like the color distinction on the crosshair though. Maybe we could change both crosshair and tooltip? Anyway, it's good enough to merge now and can be expanded later.

Lazaroth:
An option is to change "show owned" in the cfg from a boolean to an int, and have:
0: nothing changed
1: show text "Owned"
2: tint crosshair
3: both

zinnschlag:
Crosshair is good enough for now. Will merge.

PR #674, opened 2015-07-18
Change color of both tooltip and crosshair when pointing on owned object

kunesj:
Related to #668 

Added some stuff that should have been in previous pull request...

1) Not just crosshair but tooltip changes colour now too

2) Changed settings to enable changing only crosshair or tooltip colour

```
[Game]
# Change crosshair/toolTip color when pointing on owned object
#0: nothing changed
#1: tint toolTip
#2: tint crosshair
#3: both
show owned = 0
```

3) Added colour settings to settings.cfg

```
[GUI]
# colour definitions (red green blue alpha)
color background owned = 0.15 0 0 1
color crosshair owned = 1 0.15 0.15 1
```

4) It's now possible to use values from settings.cfg in mygui xml files. ([thread](https://forum.openmw.org/viewtopic.php?f=6&t=2999))
Example:
```
<Property key="Colour" value="#{setting=GUI,color background owned}"/> 
```
will give you value (string) of "color background owned" under "[GUI]" section in settings.cfg

5) screenshot

screenshot from 2015-07-18 22 15 03

### DIFF
--- a/apps/openmw/mwgui/windowmanagerimp.cpp
+++ b/apps/openmw/mwgui/windowmanagerimp.cpp
@@ -1095,11 +1095,22 @@ namespace MWGui
     void WindowManager::onRetrieveTag(const MyGUI::UString& _tag, MyGUI::UString& _result)
     {
         std::string tag(_tag);
+        
+        std::string MyGuiPrefix = "setting=";
+        size_t MyGuiPrefixLength = MyGuiPrefix.length();
 
         std::string tokenToFind = "sCell=";
         size_t tokenLength = tokenToFind.length();
-
-        if (tag.compare(0, tokenLength, tokenToFind) == 0)
+        
+        if(tag.compare(0, MyGuiPrefixLength, MyGuiPrefix) == 0)
+        {
+            tag = tag.substr(MyGuiPrefixLength, tag.length());
+            std::string settingSection = tag.substr(0, tag.find(","));
+            std::string settingTag = tag.substr(tag.find(",")+1, tag.length());
+            
+            _result = Settings::Manager::getString(settingTag, settingSection);            
+        }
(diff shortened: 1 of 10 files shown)


kunesj:
I am using monitor with very bad display, so if the current default colours look weird on normal monitor, tell me what would look better...

apps/openmw/mwgui/windowmanagerimp.cpp
```diff
@@ -1095,11 +1095,22 @@ namespace MWGui
...
+        
+        std::string MyGuiPrefix = "setting=";
+        size_t MyGuiPrefixLength = MyGuiPrefix.length();
 
         std::string tokenToFind = "sCell=";
         size_t tokenLength = tokenToFind.length();
-
-        if (tag.compare(0, tokenLength, tokenToFind) == 0)
+        
+        if(tag.compare(0, MyGuiPrefixLength, MyGuiPrefix) == 0)
```
scrawl: Please add a description of this new syntax to the function comment in windowmanagerimp.hpp.

Lazaroth:
I tried out #668 earlier today and noticed a potential issue. It's not that big of a problem, but could lead to a misunderstanding.
http://i.imgur.com/yYMlYXv.jpg
In the above image I'm looking at the counter. There's nothing there to steal, but it is still marked as red. While that isn't that big a problem, to the left in the image is a "doorway" that you can open. It is also marked as red, and the player might be afraid to open it since it shows as red despite being a thing that you are allowed to interact with.

Do other people see this a problem and if so, can it be fixed with a simple check to see if the thing you look at is e.g. movable or a container?

zinnschlag:
That is an interesting point. But I am going to merge this anyway. Any further improvements can be handled in a new PR.

PR #676, opened 2015-07-19
Refactoring ai wander

dteviot:


apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -221,12 +221,12 @@ namespace MWMechanics
 
         if(movement.mRotation[2] != 0)
         {
-            if(zTurn(actor, osg::DegreesToRadians(movement.mRotation[2]))) movement.mRotation[2] = 0;
+            if(zTurn(actor, movement.mRotation[2])) movement.mRotation[2] = 0;
```
scrawl: Kinda inconsistent. You first assign a value in radians to movement.mRotation, which is then overwritten by zTurn with a value in degrees, or set to 0 if zTurn wasn't called. The CharacterController then goes on to use the target rotation value in degrees. 
The clean thing to do would be to use either degrees or radians within the same data structure all along.

scrawl: To be clear, I mean the MWMechanics::Movement struct. We should use it to store rotations in Radians <i>or</i> Degrees, but not both.

dteviot:
> which is then overwritten by zTurn with a value in degrees, or set to 0 if zTurn wasn't called

I'm confused.  When I look at smoothTurn() in  https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/steering.cpp, which is want zTurn calls, it looks to me like it's storing a value in radians, not degrees.  What am I missing?

> To be clear, I mean the MWMechanics::Movement struct. We should use it to store rotations in Radians or Degrees, but not both.

That's what I I thought I was doing.  

Aside, I think we should be working exclusively in radians or degrees.  And as computers use radians natively, that would be the preferred unit.  Although we might have to convert from degrees in a few places to be compatible with Vanilla data.  Of course, converting everything to radians would take a while.

scrawl:
Hmm... you're right. Looks like I misremembered what the CharacterController is doing. It reads the target rotation in radians, then converts it to degrees for World::rotateObject. Carry on then.

apps/openmw/mwmechanics/aiwander.hpp
```diff
@@ -63,6 +63,13 @@ namespace MWMechanics
                 Greet_InProgress,
                 Greet_Done
             };
+
+            enum WanderState {
+                Wander_ChooseAction,
+                Wander_IdleNow,
+                Wander_MoveNow,
+                Wander_Walking,
```
scrawl: apps/openmw/mwscript/../mwmechanics/aiwander.hpp:71:31: warning: comma at end of enumerator list [-Wpedantic]
                 Wander_Walking,
                               ^

scrawl:
Looks good to me, aside from the enumerator comma warning.

PR #677, opened 2015-07-19
Fix definition conflict

Koncord:
OpenSceneGraph\include\osgAnimation\MorphGeometry:32:13: error: expected identifier before numeric constant RELATIVE.
RELATIVE defined in wingdi.h.


scrawl:
What includes wingdi.h? Our source tree doesn't, and neither does OSG.

Koncord:
GL.h in mingw include windows.h and windows.h include wingdi.h.

PR #678, opened 2015-07-19
DELE sub-record handling

smbas:
It was not so easy. A lot of code was reworked.

**Most substantial changes:**
* Move DELE handling to the ESM records (mIsDeleted flag). The records that can't be deleted: Game Settings, Skills, Magic Effects.
* Move NAME handling to the ESM records (in some records DELE is appeared before NAME).
* Most records search DELE at every sub-record's position (for example, SNDG record in a deleted form contains only NAME and then DELE in the version 1.20, but in 1.30 it contains all sub-records and DELE at the end).
* Implement saving of deleted records in OpenCS.

All seems to work, but it's needed more testing by someone with a lot of mods.


components/esm/cellref.cpp
```diff
@@ -43,60 +53,90 @@ void ESM::CellRef::loadId (ESMReader& esm, bool wideRefNum)
...
+            case ESM::FourCC<'N','A','M','0'>::value:
+                esm.skipHSub();
+                break;
+            case ESM::FourCC<'D','E','L','E'>::value:
+                esm.skipHSub();
+                mIsDeleted = true;
+                break;
+            default:
+                esm.cacheSubName();
+                isLoaded = true;
```
scrawl: What happens here when a completely unknown subrecord is encountered? Looks like while(getNextCellRef(...)) would enter an infinite loop.

scrawl:
Ouch, that's big. We should have split the task into smaller sub-tasks first.

Not sure if storing deleted flag and ID in the record itself is a good idea. That's duplicating information we already have in the Store. My suggestion would have been:

<pre>
void load(std::string& id, bool& deleted, ESMReader&);
</pre>

or

<pre>
bool load(std::string& id, ESMReader&);
</pre>

or

<pre>
class ESMReader
{
...
    std::string mLastReadID;
}
</pre>

@zinnschlag opinions?

components/esm/cellref.cpp
smbas: Yes, you are right. But this would happen with the old version of this code too.
I think that RefNum::load() (and CellRef::loadId, of course) must return a boolean value. If it's true we can loadData().

scrawl:
> Not sure if storing deleted flag and ID in the record itself is a good idea. That's duplicating information we already have in the Store.

I guess it's not worth doing another big rework. IMO we can merge it like this (once I'm done reviewing the rest of the code).

apps/openmw/mwworld/esmstore.cpp
```diff
@@ -96,33 +96,21 @@ void ESMStore::load(ESM::ESMReader &esm, Loading::Listener* listener)
...
-              continue;
-            }
-            it->second->load(esm, id);
-
-            // DELE can also occur after the usual subrecords
-            if (esm.isNextSub("DELE")) {
-              esm.skipRecord();
-              it->second->eraseStatic(id);
-              continue;
+            RecordId id = it->second->load(esm);
```
scrawl: I think it's easier to handle deletion (i.e. call eraseStatic) in the load method. Then the return value becomes just a std::string instead of the RecordId struct.

scrawl: Or maybe not. Not sure how to prevent deleted records from being inserted to mIds with that approach.

zinnschlag:
@scrawl I agree. Storing the delete flag in the record is not the right way to do it. I am not sure about what you mean with the ID. We have been storing the ID in the records for ages and that is actually required in many places. Or were you referring to something more specific? (haven't looked at all the code changes yet)

smbas:
@zinnschlag Should I remove the deleted flag and change the load and save methods the following way (?):

void load(ESMReader &esm, bool &isDeleted);
void save(ESMWriter &esm, bool isDeleted = false) const;

zinnschlag:
Yeah, that looks reasonable.

smbas:
The task is done.

Also I'm not very happy with the current code for saving of ESM records. The position of the DELE sub-record is very inconsistent (https://forum.openmw.org/viewtopic.php?f=6&t=2978). We can try to unify it since the load methods of ESM records don't rely on a sub-record order (except for Cell and Land records that allow the DELE sub-record only at the start positions). Or do we need to reflect the structure of the original format here?

zinnschlag:
Unifying the save code sounds reasonable.

zinnschlag:
Error on load in OpenMW-CS: Unknown subrecord. Record: REGN Subrecord DELE Offset: 0x183

Steps to reproduce:
- Create a new addon for Morrowind.esm
- Delete the first region
- Save
- Close
- Load Morrowind and addon into OpenMW-CS

smbas:
Save format changes:
* Most records: NAME-DELE (end)
* Cells: NAME-DATA-DELE (list of references) (end)
* Infos: INAM-PNAM-NNAM-DELE (end)
* Lands: INTV-DATA-DELE (end)
* LandTextures: NAME-INTV-DATA-DELE (end)
* Pathgrids: NAME-DATA-DELE (end)
* Scripts: SCHD-DELE (end)

zinnschlag:
Looks okay to me. @scrawl: ?

components/esm/loaddial.hpp
```diff
@@ -31,7 +31,7 @@ struct Dialogue
         Greeting = 2,
         Persuasion = 3,
         Journal = 4,
-        Deleted = -1
+        Unknown = -1 // Used for deleted dialogues
```
scrawl: The naming was changed to "Unknown", and then a comment added that previous naming "Deleted" was actually correct. Why?

scrawl:
The potential fallout from this PR is huge. I want to do a lot more testing (and maybe some profiling) before merging this.

Speaking of testing, we don't even have ESM tests yet... we should have nice and simple test ESPs for testing things like: merging dialogue infos in the correct order, overwriting records, deleting records, saving a file just like it was read in, etc, etc.

apps/openmw/mwworld/store.cpp
```diff
@@ -179,16 +185,19 @@ namespace MWWorld
...
-        std::pair<typename Static::iterator, bool> inserted = mStatic.insert(std::make_pair(idLower, T()));
+        record.load(esm, isDeleted);
+        Misc::StringUtils::toLower(record.mId);
+
+        std::pair<typename Static::iterator, bool> inserted = mStatic.insert(std::make_pair(record.mId, record));
         if (inserted.second)
             mShared.push_back(&inserted.first->second);
 
-        inserted.first->second.mId = idLower;
-        inserted.first->second.load(esm);
```
scrawl: By removing these two lines the overwriting of existing records was broken.

apps/openmw/mwworld/store.cpp
```diff
@@ -1031,16 +1064,20 @@ namespace MWWorld
     //=========================================================================
 
     template <>
```
scrawl: Now that name handling is in the record classes, we should be able to get rid of the Store specializations for ESM::Script and ESM::StartScript.

apps/openmw/mwworld/store.cpp
scrawl: That raises a problem... to merge records in this way, we need the ID of the record, <i>before</i> we load the rest of the record.

Question: do we need merging? It seems like the superior approach, but right now the ESM records save all fields anyway (even non-modified fields). But we might want to expand on this in the future. @zinnschlag: your opinion please?

components/esm/loaddial.hpp
smbas: I've changed it, because of a deleted flag that was added to the record.

scrawl:
Question: what happens in MW if content file A deletes a record, then content file B (loaded afterwards) makes a modification to it? Does the record stay deleted?

kcat:
Can't say for sure about vanilla Morrowind, but I know with Oblivion that causes a crash. When a record gets deleted the engine no longer recognizes it, so trying to do anything to it would try to edit a non-existent object. It led to a tool to mass "undelete and disable" references, because disabled references would still be modifiable thus preventing such crashes, while still being ignored in-game (of course, if something subsequently tries to enable such a reference, it will show up again when it probably shouldn't).

smbas:
Just tested a few such cases in MW: records don't stay deleted. All modifications are applied.

apps/openmw/mwworld/store.cpp
zinnschlag: You mean like only modify a subrecord in an addon and leave the rest of the original record unchanged? Sounds useful, but also very complicated. Definitely not on my list for the near or medium future. OpenMW 2.0-ish? I don't think we need to take that into consideration now.

smbas: To be able to merge records we can split the load method into two methods: loadId() and loadData(). But in this case, both methods must have a deleted flag as a parameter. To know whether a record is deleted we must checked them both. 

What's your opinions about this, @zinnschlag, @scrawl?

scrawl: > I don't think we need to take that into consideration now.

Right. @smbas: just fix the code to restore the overwriting of records (no merging needed), then we're good.

scrawl:
I'd like to put this PR on hold until I've created the aforementioned unit tests.

artemutin:
I have tested this branch with my favorite BIG mode, and could report about some issues - referencing a deleted ref (while object ID still remains undeleted) in script by ->disable or ->unlock cause engine to throw an error "Unknown id: someID" and stop script execution. It happens with NPCs and Activators as well. In vanilla MW nothing happens and scripts works well.
Also this mode deletes a spell, but leaves it in Spells list of NPC. As a result i get a warning "Ignoring non-existing spell 'blabla' on NPC 'npcID' "
Also i got some other type of errors, may be it's not related to this branch. I got this on loading Seyda Neen Cell -- error during rendering 'erene llenim': vector::_M_range_check:__n(which is 0)>=this->size()(which is 0)

scrawl:
@artemutin: these problems seem unrelated to the branch in question. Do you get the same problems on the master branch?

> Another issue -- spell was deleted from game, but was leaved in Spells list of NPC. As a result i got a warning "Ignoring non-existing spell 'blabla' on NPC 'npcID' "

Working as expected. Using a deleted object <i>should</i> result in a warning.

artemutin:
1) No, I have not. Master branch works well, but without this mod, obviously.
2) Yep, it should. I understand now, that this is problem with inaccurate scripts and object reference deletion in general in this mod, rather than bug in OpenMW. May be it shouldn't break script execution with exception and just issue a warning, like missing spell reference did. And this kind of errors are candidates to be checked with verifier.

scrawl:
Sorry for the delay in accepting the PR. Didn't want it in 0.37 because it's a big and scary change. I just solved those pesky merge conflicts and will write the unit tests tomorrow. Then we can merge if everything is well.

psi29a:
+1 for unit tests! :)

scrawl:
Some profiling data: 

esm_rewrite:
<pre>
content files: 191.466 ms
</pre>

master:
<pre>
content files: 174.379 ms
</pre>

margin of error is ~4 ms.

```
diff --git a/apps/openmw/mwgui/loadingscreen.cpp b/apps/openmw/mwgui/loadingscreen.cpp
index 321d5e6..b0425bb 100644
--- a/apps/openmw/mwgui/loadingscreen.cpp
+++ b/apps/openmw/mwgui/loadingscreen.cpp
@@ -237,6 +237,7 @@ namespace MWGui
 
     bool LoadingScreen::needToDrawLoadingScreen()
     {
+        return false;
         if ( mTimer.time_m() <= mLastRenderTime + (1.0/mTargetFrameRate) * 1000.0)
             return false;
 
diff --git a/apps/openmw/mwworld/worldimp.cpp b/apps/openmw/mwworld/worldimp.cpp
index 7e78bef..25eb0ed 100644
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -8,6 +8,8 @@
 #include <tr1/unordered_map>
 #endif
 
+#include <osg/Timer>
+
 #include <osg/Group>
 #include <osg/ComputeBoundsVisitor>
 #include <osg/PositionAttitudeTransform>
@@ -179,8 +181,11 @@ namespace MWWorld
         gameContentLoader.addLoader(".omwaddon", &esmLoader);
         gameContentLoader.addLoader(".project", &esmLoader);
 
-        loadContentFiles(fileCollections, contentFiles, gameContentLoader);
+        // esm_rewrite: content files: 191.466
 
+        osg::Timer timer;
+        loadContentFiles(fileCollections, contentFiles, gameContentLoader);
+std::cout << "content files: " << timer.time_m() << std::endl;
         listener->loadingOff();
 
         // insert records that may not be present in all versions of MW
```

scrawl:
With Info loading commented out, the difference is much smaller:

esm_rewrite:
content files: 133.964 ms

master:
content files: 128.103 ms

```
diff --git a/apps/openmw/mwworld/esmstore.cpp b/apps/openmw/mwworld/esmstore.cpp
index 50324f3..e77a591 100644
--- a/apps/openmw/mwworld/esmstore.cpp
+++ b/apps/openmw/mwworld/esmstore.cpp
@@ -71,6 +71,8 @@ void ESMStore::load(ESM::ESMReader &esm, Loading::Listener* listener)
 
         if (it == mStores.end()) {
             if (n.val == ESM::REC_INFO) {
+                esm.skipRecord();
+                /*
                 if (dialogue)
                 {
                     dialogue->readInfo(esm, esm.getIndex() != 0);
@@ -80,6 +82,7 @@ void ESMStore::load(ESM::ESMReader &esm, Loading::Listener* listener)
                     std::cerr << "error: info record without dialog" << std::endl;
                     esm.skipRecord();
                 }
+                */
             } else if (n.val == ESM::REC_MGEF) {
                 mMagicEffects.load (esm);
             } else if (n.val == ESM::REC_SKIL) {
```

Which is strange, because nothing really changed in the info loading code as far as I could tell.

scrawl:
Added the first test, all good so far.
<pre>$ diff test_dialogue_merging_master.txt test_dialogue_merging.txt
$</pre>

scrawl:
#805

PR #679, opened 2015-07-20
Fix additem/removeitem (bug #2788)

Koncord:
Behavior as the original game.

### DIFF
--- a/apps/openmw/mwscript/containerextensions.cpp
+++ b/apps/openmw/mwscript/containerextensions.cpp
@@ -56,6 +56,12 @@ namespace MWScript
                     if (count == 0)
                         return;
 
+                    if(Misc::StringUtils::ciEqual(item, "gold_005")
+                            || Misc::StringUtils::ciEqual(item, "gold_010")
+                            || Misc::StringUtils::ciEqual(item, "gold_025")
+                            || Misc::StringUtils::ciEqual(item, "gold_100"))
+                        item = "gold_001";
+
                     MWWorld::Ptr itemPtr = *ptr.getClass().getContainerStore (ptr).add (item, count, ptr);
 
                     // Spawn a messagebox (only for items added to player's inventory and if player is talking to someone)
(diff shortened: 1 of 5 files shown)


apps/openmw/mwworld/containerstore.cpp
```diff
@@ -298,26 +300,20 @@ MWWorld::ContainerStoreIterator MWWorld::ContainerStore::addImp (const Ptr& ptr,
 
     // gold needs special handling: when it is inserted into a container, the base object automatically becomes Gold_001
     // this ensures that gold piles of different sizes stack with each other (also, several scripts rely on Gold_001 for detecting player gold)
-    if (Misc::StringUtils::ciEqual(ptr.getCellRef().getRefId(), "gold_001")
-        || Misc::StringUtils::ciEqual(ptr.getCellRef().getRefId(), "gold_005")
-        || Misc::StringUtils::ciEqual(ptr.getCellRef().getRefId(), "gold_010")
-        || Misc::StringUtils::ciEqual(ptr.getCellRef().getRefId(), "gold_025")
-        || Misc::StringUtils::ciEqual(ptr.getCellRef().getRefId(), "gold_100"))
+    if(MWClass::Miscellaneous().isGold(ptr))
```
scrawl: You should use ptr's class `ptr.getClass()` instead of constructing a new one.

scrawl:
Are you sure this won't break non-scripted gold pickup? If I activate an item of `gold_100` that's lying around in the gameworld, that's supposed to add 100 gold pieces to the inventory.

Koncord:
Whoops, I forgot non-scripted pick-up.

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -13,6 +13,8 @@
 #include "../mwmechanics/creaturestats.hpp"
 #include "../mwmechanics/levelledlist.hpp"
 
+#include "../mwclass/misc.hpp"
```
scrawl: Include can be removed.

apps/openmw/mwscript/containerextensions.cpp
```diff
@@ -119,6 +131,12 @@ namespace MWScript
                     if (count == 0)
                         return;
 
+                    if(Misc::StringUtils::ciEqual(item, "gold_005")
+                        || Misc::StringUtils::ciEqual(item, "gold_010")
+                        || Misc::StringUtils::ciEqual(item, "gold_025")
+                        || Misc::StringUtils::ciEqual(item, "gold_100"))
+                    item = "gold_001";
```
scrawl: This line looks confusing because of the missing indent.

PR #681, opened 2015-07-20
Rebase2 android build fix

sandstranger:
finally figured out how to create branches in git.....
In previous commit I broke the game with my delusions completely . It is even not started.
Conclusion: gles2shadergenvisitor not working ,need custom shaders
Gles texture сompression will not work .
Let textures will be decompressed in memory .The easiest way to do this, use gltools
For this openmw should be build for armv7 arch.
Anyway armv8 gives nothing in  performance
Armv8 may allow the application to use more memory, but other than that nothing more


scrawl:
It's nice that you're sharing your progress, but I'm not going to merge until something usable comes out of this.

sandstranger:
Ok

sandstranger:
it broken

PR #682, opened 2015-07-22
Increase the max view distance to something more reasonable for a high end machine.

EmperorArthur:
Exterior grid size must be at least 6 for the new maximum to work correctly.

Let the discussion begin.  While I'm not sure super far should be the default, this certainly improves the game experience.

Note:
In mwworld/scene.cpp: 330
halfGridSize is an int, and uses integer division on "exterior grid size"
This means odd numbers (like the original '3') should not be used in exterior grid size.

Ideally this should be another setting in the cfg file, but with a quick look I couldn't figure out how to get that to work correctly.

### DIFF
--- a/files/mygui/openmw_settings_window.layout
+++ b/files/mygui/openmw_settings_window.layout
@@ -345,7 +345,7 @@
                             <UserString key="SettingName" value="viewing distance"/>
                             <UserString key="SettingValueType" value="Float"/>
                             <UserString key="SettingMin" value="2000"/>
-                            <UserString key="SettingMax" value="6666"/>
+                            <UserString key="SettingMax" value="30000"/>
                         </Widget>
                         <Widget type="TextBox" skin="SandText" position="4 178 332 18" align="Left Top">
                             <Property key="Caption" value="#{sNear}"/>


zinnschlag:
A clear no to the increased default grid size. Even ignoring the performance impact, there is still the fact that it might break things. Some scripts in some content files may depend on the 3x3 grid.

zinnschlag:
To elaborate on that further:

Maybe we need two grid sizes: A rendering grid size and an everything else grid size (physics, scripting). But then, do we actually need the rendering grid size? We could calculate it automatically from the viewing distance (with 3x3 as a minimum).
I would consider the other grid size a part of the content data (new GMST eventually), because this parameter can depend on the content. Some games may not run properly with more than 3x3 and some future games may require more than 3x3 (for example for extensive NPC schedules or other NPC activities).

kcat:
AI and script processing should very much be separated from terrain and (some) object rendering. Before the switch to OSG, we had a far terrain option which massively increased the view distance (pretty much being able to show the whole game world), but only the nearest 3x3 cells would be high detail with objects and actors in it. There were only two downsides: fog was broken since it was tied to the view distance (so with the view distance way out, the fog would never be visible), and the vanilla game uses models like rocks and lava to detail the terrain which wouldn't be visible when outside the 3x3 grid.

Bringing back the far terrain would be the first step, I think. After that, fixing the fog, and then improving the distant LOD capabilities (better far terrain, certain objects remaining visible in the distant LOD), which would be separate from the actor/AI cell grid size.

EmperorArthur:
Unfortunately, things like distant land don't help with places like Vivec, though LOD should work.

I can understand not wanting to change the "exterior grid size", but I do feel it would be helpful to have the distance slider go much further than the previous max of 6666.  Even if that's the only change not seeing fog up close definitely makes the game feel that much more open.

As a related note I've opened Bug #2793, since openmw isn't currently loading that 3x3 grid like it should.

scrawl:
Both changes are a no-go.
* Default settings shouldn't be chosen for high-end machines. The game should run fine on first launch even with mediocre hardware.
* Processing scripts for more cells could break the game in some way. Seems unlikely, but with the level of shoddiness we have seen in MW scripting in the past it wouldn't entirely surprise me.
* The maximum of the view distance slider was specifically chosen for pop-in to be a non-issue with the default 3x3 grid. If we increase the maximum, there is no obvious way anymore to get to that particular sweet spot.

PR #683, opened 2015-07-23
Improve water walk

Koncord:


Koncord:
There is a chance that Vanilla Morrowind does not use actor height.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -763,6 +763,35 @@ namespace MWMechanics
...
+                const ESM::Position casterPosition = mCaster.getRefData().getPosition();
+                const float casterHeight = mCaster.getCellRef().getScale();
+
+                if(MWBase::Environment::get().getWorld()->castRay(
+                        casterPosition.pos[0], casterPosition.pos[1], casterPosition.pos[2],
+                        casterPosition.pos[0], casterPosition.pos[1], cell->getWaterLevel() + casterHeight ))
+                {
+                    // not enough space to place caster
+                    MWBase::Environment::get().getWindowManager()->messageBox ("#{sMagicInvalidEffect}");
+                    return 0;
```
scrawl: Should not return here. The spell being cast might have additional effects other than water walking, which we still need to apply.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -763,6 +763,35 @@ namespace MWMechanics
...
+        if(mCaster.getClass().isActor()
+                && MWBase::Environment::get().getWorld()->getStore().get<ESM::MagicEffect>()
+                .find(ESM::MagicEffect::WaterWalking))
+        {
+            const MWWorld::CellStore *cell = mCaster.getCell();
+            
+            if(cell->getCell()->hasWater())
+            {
+                const ESM::Position casterPosition = mCaster.getRefData().getPosition();
+                const float casterHeight = mCaster.getCellRef().getScale();
```
scrawl: getScale() isn't the actor's height, it's a scaling factor applied to the model (which you'll find is usually 1).

To get the actor height, you would need to get it from the PhysicsSystem::getHalfExtents, then multiply by getScale().

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -763,6 +763,35 @@ namespace MWMechanics
...
+
+                if(MWBase::Environment::get().getWorld()->castRay(
+                        casterPosition.pos[0], casterPosition.pos[1], casterPosition.pos[2],
+                        casterPosition.pos[0], casterPosition.pos[1], cell->getWaterLevel() + casterHeight ))
```
scrawl: We shouldn't do this raycast if the player is already above water, right?

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -763,6 +763,35 @@ namespace MWMechanics
 
         int school = 0;
 
+        if(mCaster.getClass().isActor()
+                && MWBase::Environment::get().getWorld()->getStore().get<ESM::MagicEffect>()
+                .find(ESM::MagicEffect::WaterWalking))
+        {
+            const MWWorld::CellStore *cell = mCaster.getCell();
```
scrawl: You are checking caster stats, but the caster might not be the recipient of the water walking effect (e.g. with on target spells).

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -763,6 +763,35 @@ namespace MWMechanics
...
+                    // not enough space to place caster
+                    MWBase::Environment::get().getWindowManager()->messageBox ("#{sMagicInvalidEffect}");
+                    return 0;
+                }
+                else if(MWBase::Environment::get().getWorld()->isUnderwater(cell, casterPosition.asVec3()))
+                {
+                    // move the caster to surface
+                    ESM::Position newCasterPosition = casterPosition;
+                    newCasterPosition.pos[2] = cell->getWaterLevel();
+                    mCaster.getRefData().setPosition(newCasterPosition);
```
scrawl: Is this even working? To correctly move an actor, you need to use World::moveObject.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -763,6 +763,35 @@ namespace MWMechanics
 
         int school = 0;
 
+        if(mCaster.getClass().isActor()
```
scrawl: The castSpell function isn't the correct place to do this check. There are other ways of casting effects (enchantments, potions, etc.) that don't go through castSpell.

Either you do it in somewhere in CastSpell::inflict, or directly in PhysicsSystem where Water Walking is actually applied (preferable IMO, since you can easily get actor's physical height from there).

Observation in MW: The underwater check isn't done when the spell is cast, but rather every frame. You can test this by disabling collision (TCL), casting water walking, going underwater, then do another TCL (player is moved above water). The implementation in physics system would support this use case.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -763,6 +763,35 @@ namespace MWMechanics
...
+                && MWBase::Environment::get().getWorld()->getStore().get<ESM::MagicEffect>()
+                .find(ESM::MagicEffect::WaterWalking))
```
scrawl: I'm not sure what you are trying to do here. The WaterWalking effect will always exist in the Store (with valid game files, anyway), so there's no point in checking it.

Koncord:
Need review.

scrawl:
That's better. I'd still prefer moving the code to PhysicsSystem, that way you can easily address both FIXME comments.

scrawl:
More testing in MW:

* If the player is deep in the water (about 350 units), attempting to cast water walk gives "You cannot cast this effect at the moment". The effect is not added to active spell icons.
* If the player is close to the surface, but an object is blocking the surface, attempting to cast water walk does not give an error, but does not move the player to the surface either. The player will be moved to the surface as soon as the blocking object disappears or is circumvented.
* The surface can not be blocked by other NPCs or creatures, only objects can.

Koncord:
I don't know how to delete effects in PhysicsSystem =(
```
iter->first.getClass().getCreatureStats(iter->first).remove(ESM::MagicEffect::WaterWalking);
```
It does nothing, because it has a constant modifier.

Koncord:
> If the player is close to the surface, but an object is blocking the surface, attempting to cast water walk does not give an error, but does not move the player to the surface either. The player will be moved to the surface as soon as the blocking object disappears or is circumvented.
I cannot repeat it.

PR #688, opened 2015-07-24
#2725 and #2730

Dieho:


scrawl:
Please configure your editor to use 4 spaces, not tabs.

apps/opencs/view/world/dialoguesubview.hpp
```diff
@@ -237,11 +237,13 @@ namespace CSVWorld
 
             bool isLocked() const;
 
+			int GetColumnIndex(CSMWorld::Columns::ColumnId column);
```
scrawl: Naming should be ```getColumnIndex``` to be consistent with the existing methods in this file.

zinnschlag:
Also, don't add a getColumnIndex function to the Dialogue class. That is redundant and very inconsistent with what we are doing everywhere else.

smbas:
The 0 column in tables is the ID column (i.e. ColumnId_Id).

apps/opencs/view/world/dialoguesubview.hpp
```diff
@@ -236,12 +236,12 @@ namespace CSVWorld
...
 
             bool isLocked() const;
-
+        
         public:
 
             SimpleDialogueSubView (const CSMWorld::UniversalId& id, CSMDoc::Document& document);
 
-            virtual void setEditLock (bool locked);
+	    virtual void setEditLock (bool locked);
```
scrawl: There's still tabs here.

PR #689, opened 2015-07-24
Task #2793. Changed a settings variable responsible for number of loaded exterior cells.

pkubik:
Changed a settings variable responsible for number of loaded exterior cells.
https://bugs.openmw.org/issues/2793

Tested for few values. Maybe I should add some greater-than-0 check?

### DIFF
--- a/apps/openmw/mwworld/scene.cpp
+++ b/apps/openmw/mwworld/scene.cpp
@@ -327,7 +327,7 @@ namespace MWWorld
         std::string loadingExteriorText = "#{sLoadingMessage3}";
         loadingListener->setLabel(loadingExteriorText);
 
-        const int halfGridSize = Settings::Manager::getInt("exterior grid size", "Cells")/2;
+        const int halfGridSize = Settings::Manager::getInt("exterior cell load distance", "Cells");
 
         CellStoreCollection::iterator active = mActiveCells.begin();
         while (active!=mActiveCells.end())
--- a/files/settings-default.cfg
+++ b/files/settings-default.cfg
@@ -101,7 +101,7 @@ local map widget size = 512
 local map hud widget size = 256
 
 [Cells]
-exterior grid size = 3
+exterior cell load distance = 1
 
 [Camera]
 near clip = 5


scrawl:
> Maybe I should add some greater-than-0 check?

0 is a valid value for this setting (though probably not one you would use during normal play).

pkubik:
> 0 is a valid value for this setting (though probably not one you would use during normal play).

Ok, even lesser than 0 values aren't causing some unpredictable crashes. I think it should be obvious that something might go wrong when any distance is lesser than 0.

EmperorArthur:
Negative values do not make sense for this setting.
You could add getUInt and setUInt functions to components/settings/settings.[ch]pp
And just change halfGridSize to an unsigned int.

zini:  You ok with that?

Actually, I prefer to make these types of getters and setters templated, but that caused issues when I tried it for the nif files.

zinnschlag:
No to the unsigned. All that would do is causing warnings about mixing signed and unsigned. I stand to my opinion that you should generally avoid using unsigned ints in arithmetic context as much as possible.

Values < 0 aren't valid and that we don't get any problems with them is just an accident. While it is not strictly necessary to check settings for validity, it doesn't hurt either. Two options here:

- either silently replace values <0 with 0 when getting the settings value
- on startup (probably somewhere in the world constructor) check for negative values and if present, fix them and report it via std::cerr.

pkubik:
> on startup (probably somewhere in the world constructor) check for negative values and if present, fix them and report it via std::cerr

I thought that we are using setting manager on every function call instead of some const field (set in the constructor) because we assume that settings might be changed during gameplay. Am I correct?
Ideally we would need to make some kind of checking function called on startup and every settings modification. 
I'm not sure it's really necessary however. Too much fuss. Note that almost every settings value should has some kind of constraint. I opt for silently replacing with 0.

zinnschlag:
The grid size can't be modified during a session currently. It requires a restart. Once we have the required UI element in the settings window (post-1.0), we will just limit it at 0. Therefore checking and adjusting once during startup would be sufficient.
But I am okay with silently adjusting the value to 0.

pkubik:
I don't see any checks for other settings. How about accepting the change as it is, and think about some more general solution later?

zinnschlag:
Agreed. Will merge.

PR #691, opened 2015-07-26
Refactoring ai wander

dteviot:


apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -355,6 +261,150 @@ namespace MWMechanics
...
+        // Check if idle animation finished
+        GreetingState& greetingState = storage.mSaidGreeting;
+        if (!checkIdle(actor, storage.mIdleAnimation) && (greetingState == Greet_Done || greetingState == Greet_None))
+        {
+            storage.mState = Wander_ChooseAction;
+        }
+    }
+
+    void AiWander::onWalkingStatePerFrameActions(const MWWorld::Ptr& actor, 
+        float duration, AiWanderStorage& storage, ESM::Position& pos)
```
EmperorArthur: Interesting.
Have you considered a renaming this to something like "onMovePerFrameActions(...)".
Then you could factor out the movement part of aicombat and aifollow, and not have walking in the name.

Correct me if I'm wrong, but that seems to be the direction you're looking to go.  That is, factoring everything out so the ai system is a neat state machine within each of the packages.

Heck, given that aifollow toggles between walking and running, you could probably merge all the ai "onMovePerFrameActions(...)" into one shared function, and take advantage of the aicombat code when moving quickly.

dteviot:
Yes, that is my ultimate goal.  But to get there I have to fully understand the AiPackages, or at least, the movement part of them, and the refactoring is the first step.

dteviot:
On further reflection, I'd defer the renaming until I've managed to pull all the movement related functions in AiCombat together.  As we'd probably want to use that logic (which I believe is the most complete) for all movement cases.

PR #694, opened 2015-07-27
Nif tester

EmperorArthur:
This is a tool, and a crude linux script to allow users to see if their nif files will work with openmw.

As someone who has had a texture replacer remove all my lockpicks from the game, I consider this to be a rather important addition.  Especially since I didn't notice until checking the games error logs.

### DIFF
--- a/apps/niftest/niftest.cpp
+++ b/apps/niftest/niftest.cpp
@@ -1,0 +1,87 @@
+///Program to test .nif files both on the FileSystem and in BSA archives.
+
+#include <iostream>
+#include <fstream>
+
+#include <components/nif/niffile.hpp>
+#include <components/files/constrainedfilestream.hpp>
+#include <components/vfs/manager.hpp>
+#include <components/vfs/bsaarchive.hpp>
+
+
+///See if the file has the named extension
+bool hasExtension(std::string filename, std::string  extensionToFind)
+{
+    std::string extension = filename.substr(filename.find_last_of(".")+1);
+
+    //Convert strings to lower case for comparison
+    std::transform(extension.begin(), extension.end(), extension.begin(), ::tolower);
+    std::transform(extensionToFind.begin(), extensionToFind.end(), extensionToFind.begin(), ::tolower);
+
+    if(extension == extensionToFind)
+        return true;
+    else
+        return false;
+}
+
+///See if the file has the "nif" extension.
+bool isNIF(std::string filename)
+{
+    return hasExtension(filename,"nif");
+}
+///See if the file has the "bsa" extension.
+bool isBSA(std::string filename)
+{
+    return hasExtension(filename,"bsa");
+}
+
+///Check all the nif files in the given BSA archive
+void readBSA(std::string filename)
+{
+    VFS::Manager myManager(false);
+    myManager.addArchive(new VFS::BsaArchive(filename));
+    myManager.buildIndex();
+
+    std::map<std::string, VFS::File*> files=myManager.getIndex();
+    for(std::map<std::string, VFS::File*>::const_iterator it=files.begin(); it!=files.end(); ++it)
+    {
+      std::string name = it->first;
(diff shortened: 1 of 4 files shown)


scrawl:
> Especially since I didn't notice until checking the games error logs.

Perhaps we should be using an error mesh (the exclamation point one) for meshes that don't exist or failed to load, like the vanilla engine does.

scrawl:
That only catches errors in the NIF reader, but we could also have errors in the NIF loader. Perhaps the tool should be renamed to nifreadertest.

I don't agree with enabling the build by default. Tests are something that most users won't need.  Plus, we already have 8 executables that are built by default, which is a big enough number as it is.

EmperorArthur:
True, it doesn't touch the loader, but I'd like to leave the name as is.  That way if/when I get a chance to have it use the loader it won't have to go through another rename.

apps/niftest/niftest.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return hasExtension(filename,"bsa");
+}
+
+///Check all the nif files in the given BSA archive
+void readBSA(std::string filename)
+{
+    VFS::Manager myManager(false);
+    //This does not work, since manager tries to delete it on cleanup
+        //VFS::BsaArchive myBsa(filename);
+        //myManager.addArchive(&myBsa);
```
scrawl: What's the point of keeping non-working code as a comment?

scrawl:
> True, it doesn't touch the loader, but I'd like to leave the name as is. That way if/when I get a chance to have it use the loader it won't have to go through another rename.

Sure, makes sense.

apps/niftest/niftest.cpp
EmperorArthur: Sorry, habit.
It's a note to self that "take's ownership" means I shouldn't feed the function local variables.
Would you like me to rebase and remove it?

scrawl: > Would you like me to rebase and remove it?

Yes please.

EmperorArthur:
Rebased

dreamer-dead:
Hi!

@EmperorArthur, @scrawl, sorry to bother you, as I am not an active participant of the project.
I have few remarks and questions if you don't mind.
Wiki page about CC (https://wiki.openmw.org/index.php?title=Code_Formatting_Conventions) isn't too big to give me all answers =(

- What about include headers sorting inside each group?
- What do you think about some sort of CC style checker script?
For example, it could run as git precommit hook to check new code.
In this particular pull-request I see some lines with wrong indents, braces and spaces.
- What about commented lines? Do they allowed by the CC?

And my remarks about the code of this PR.
- Shouldn't we pass std::string by const-ref as "const std::string&"?
- Lower string conversion by std::transform + ::tolower is a bad idea in general, it can
  failed on multy-bytes strings. And paths on Linux is an UTF-8 strings, IIRC. http://stackoverflow.com/questions/313970/how-to-convert-stdstring-to-lower-case
  Is it possible to use boost::filesystem::path instead?

- What about return a nonzero exit code from main in case of error?
  It can be useful in scripts.

- It would be nice to have a verbose mode for this utility.
  Like `niftest --verbose [...args]`
  It allows us to uncomment lines like
  std::cout << "Decoding: " << name << std::endl;
  and get a more readable output if needed.

- If utility output will be parsed by scripts, then we could have a special output mode.

EmperorArthur:
All very good points.
The answer to most of those why dont you... is pretty simple.  This is a perfect example of me coding something to "scratch my own itch".

For instance, the original version of this was created back before the osg port, buried in components/nif/tests/  back then I too lazy to figure out how to do directory recursion in c++ so I wrote a script instead.

Another thing is that I don't want to get too into detail with an initial pull request for a new idea.  My track record with openmw isn't exactly great.  The smaller the changes, the less I'll have to redo later.  It also means the earlier I can find out if my idea is acceptable to the project.

PR #696, opened 2015-07-28
Be more descriptive when the NIF loader does not handle something.

EmperorArthur:
As per Feature #2801

So others don't have to browse the source code just to figure out what part of a texture is not being used by openmw.

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -903,7 +903,7 @@ namespace NifOsg
                 int uvSet = *it;
                 if (uvSet >= (int)data->uvlist.size())
                 {
-                    std::cerr << "Warning: using an undefined UV set " << uvSet << " on TriShape " << triShape->name << " in " << mFilename << std::endl;
+                    std::cerr << "Warning: using an undefined UV set " << uvSet << " on TriShape \"" << triShape->name << "\" in " << mFilename << std::endl;
                     continue;
                 }
 
@@ -1264,13 +1264,34 @@ namespace NifOsg
                 {
                     if (texprop->textures[i].inUse)
                     {
-                        if (i != Nif::NiTexturingProperty::BaseTexture
-                                && i != Nif::NiTexturingProperty::GlowTexture
-                                && i != Nif::NiTexturingProperty::DarkTexture
-                                && i != Nif::NiTexturingProperty::DetailTexture)
+                        switch(i)
                         {
-                            std::cerr << "Warning: unhandled texture stage " << i << " in " << mFilename << std::endl;
-                            continue;
+                            //These are handled later on
+                            case Nif::NiTexturingProperty::BaseTexture:
+                            case Nif::NiTexturingProperty::GlowTexture:
+                            case Nif::NiTexturingProperty::DarkTexture:
+                            case Nif::NiTexturingProperty::DetailTexture:
+                                break;
+                            case Nif::NiTexturingProperty::GlossTexture:
+                            {
+                                std::cerr << "NiTexturingProperty::GlossTexture in " << mFilename << " not currently used." << std::endl;
+                                continue;
+                            }
+                            case Nif::NiTexturingProperty::BumpTexture:
+                            {
+                                std::cerr << "NiTexturingProperty::BumpTexture in " << mFilename << " not currently used." << std::endl;
+                                continue;
+                            }
+                            case Nif::NiTexturingProperty::DecalTexture:
+                            {
+                                std::cerr << "NiTexturingProperty::DecalTexture in " << mFilename << " not currently used." << std::endl;
+                                continue;
+                            }
+                            default:
+                            {
+                                std::cerr << "Warning: unhandled texture stage " << i << " in " << mFilename << std::endl;
+                                continue;
+                            }
                         }
 
                         const Nif::NiTexturingProperty::Texture& tex = texprop->textures[i];


components/nifosg/nifloader.cpp
```diff
@@ -1311,13 +1306,13 @@ namespace NifOsg
 
                             stateset->setTextureAttributeAndModes(texUnit, texEnv, osg::StateAttribute::ON);
```
scrawl: This change introduces fall-throughs that are incorrect.

components/nifosg/nifloader.cpp
```diff
@@ -1342,9 +1358,18 @@ namespace NifOsg
             }
             // unused by mw
             case Nif::RC_NiShadeProperty:
+            {
+                std::cerr << "RC_NiShadeProperty in " << mFilename  << " not currently used." << std::endl;
```
scrawl: The presence of these 3 properties in a file is not an error. As the comment above says, MW does not use those, so we shouldn't either. I specifically removed those warnings because people were complaining about the log spam with vanilla assets.

EmperorArthur:
Fixing now.

EmperorArthur:
That should do it.
Looking at the nif format I see that there's no reason to implement NiShadeProperty or NiDitherProperty, but the idea of objects having a fog emitter as from NiFogProperty seems useful.

components/nifosg/nifloader.cpp
```diff
@@ -1332,6 +1329,28 @@ namespace NifOsg
...
+                        }
+                        case Nif::NiTexturingProperty::GlossTexture:
+                        {
+                            std::cerr << "NiTexturingProperty::GlossTexture in " << mFilename << " not currently used." << std::endl;
+                            continue;
+                        }
+                        case Nif::NiTexturingProperty::BumpTexture:
+                        {
+                            std::cerr << "NiTexturingProperty::BumpTexture in " << mFilename << " not currently used." << std::endl;
+                            continue;
```
scrawl: You want to continue; before handling the texture. As it is now, e.g. a bump map would show up as a color texture, when it shouldn't show up at all so long as we can't handle it.

scrawl:
> the idea of objects having a fog emitter as from NiFogProperty seems useful.

It wouldn't really be a fog "emitter", just an override for the scene's fog settings. The problem is, you can't set the override in relation to the existing (weather-dependent) fog settings, so whatever you set would feel completely out of place.

Either way, this is a kinda pointless comment. We will use the NIF format as Morrowind used it. Any further enhancements should be reserved for the native asset format we will offer post-1.0.

components/nifosg/nifloader.cpp
```diff
@@ -1340,13 +1359,15 @@ namespace NifOsg
...
-            case Nif::RC_NiShadeProperty:
-            case Nif::RC_NiDitherProperty:
-            case Nif::RC_NiFogProperty:
-            {
+            // These three are unused by mw, and reporting them causes log spam from the original assets
+            case Nif::RC_NiShadeProperty:   //Flag to enable smooth phong shading (we already do this)
+            case Nif::RC_NiDitherProperty:  //Flag to enable dithering
                 break;
-            }
+            case Nif::RC_NiFogProperty:     //Add a fog emitter to the object
```
scrawl: As per the last comment, not really a fog emitter.

I must say that I don't see the point in commenting unused properties. Since the properties were never utilized, it is impossible to say what they were really supposed to do. Best to just ignore them and move on. So remove the comments please.

EmperorArthur:
Well, I broke it with that last rebase.  Give me a bit to fix it.

components/nifosg/nifloader.cpp
```diff
@@ -1289,50 +1280,81 @@ namespace NifOsg
...
+                        case Nif::NiTexturingProperty::GlowTexture:
+                        {
+                            osg::TexEnvCombine* texEnvCombine = new osg::TexEnvCombine;
+                            texEnvCombine->setCombine_Alpha(osg::TexEnvCombine::REPLACE);
+                            texEnvCombine->setSource0_Alpha(osg::TexEnvCombine::PREVIOUS);
+                            texEnvCombine->setCombine_RGB(osg::TexEnvCombine::ADD);
+                            texEnvCombine->setSource0_RGB(osg::TexEnvCombine::PREVIOUS);
+                            texEnvCombine->setSource1_RGB(osg::TexEnvCombine::TEXTURE);
+
+                            texEnv = dynamic_cast<osg::TexEnv*>(texEnvCombine);
```
scrawl: This cast will always fail, because TexEnvCombine is not derived from TexEnv.

I really don't like where this iteration process is going. Can we please just focus on the one thing we want to fix, and leave the rest of the code alone? I suggest to start from scratch, then replace the if (i != Nif::NiTexturingProperty::BaseTexture...) statement with a switch statement, that logs a different error message for each texturing type. That is the only code you need to change.

components/nifosg/nifloader.cpp
```diff
@@ -1289,50 +1280,81 @@ namespace NifOsg
...
-                            osg::TexEnvCombine* texEnv = new osg::TexEnvCombine;
-                            texEnv->setCombine_Alpha(osg::TexEnvCombine::REPLACE);
-                            texEnv->setSource0_Alpha(osg::TexEnvCombine::PREVIOUS);
-                            texEnv->setCombine_RGB(osg::TexEnvCombine::ADD);
-                            texEnv->setSource0_RGB(osg::TexEnvCombine::PREVIOUS);
-                            texEnv->setSource1_RGB(osg::TexEnvCombine::TEXTURE);
-
-                            stateset->setTextureAttributeAndModes(texUnit, texEnv, osg::StateAttribute::ON);
+                        case Nif::NiTexturingProperty::BaseTexture:
+                            texEnv = new osg::TexEnv;
```
scrawl: Just pointing out for completion sake that this line was redundant and would have caused additional API overhead as well.

EmperorArthur:
Here you go.  Minimal changes.

scrawl:
Thanks.

PR #697, opened 2015-07-28
Possible fix for Running in Circles when overshoot waypoint.

dteviot:


scrawl:
I think that I already explained my (negative) opinion on this fix in a forum post. https://forum.openmw.org/viewtopic.php?f=6&t=2916&start=10#p33083

My suggestion is to call the navigational AI (at least the turning and arrival detection) every frame. That will likely fix most real-world cases of the problem. If that's not enough, then we should improve the turning AI - i.e. take the rotational speed into account: if necessary, switch to walking or stop completely before attempting a particularly sharp turning manoeuvre.

dteviot:
@scrawl
> My suggestion is to call the navigational AI (at least the turning and arrival detection) every frame. That will likely fix most real-world cases of the problem.

It reduces the problem, but doesn't completely solve it.  Probably because I'm only managing to run at 4 FPS. (See https://forum.openmw.org/viewtopic.php?f=6&t=2916&start=30.  AiTravel updates every frame.  AiCombat does not.  (I suspect at higher FPS the problem will disappear, so you're probably correct when you say "fix most" .)

I'm happy to modify AiCombat to do the path checking every frame.  But, to do this I'd want to do some refactoring of the AiCombat code, to separate the movement logic from the rest (similar to what I've attempted for AiWander)  which raises the issue.
* There's a pending PR for AiComabt https://github.com/OpenMW/openmw/pull/614.  Refactoring AiCombat would probably make that check-in very difficult to do.  I'm not sure if I should proceed before this is merged.

scrawl:
> I'm only managing to run at 4 FPS

Is that a debug or release build? What graphics card are you using?

> There's a pending PR for AiComabt #614. Refactoring AiCombat would probably make that check-in very difficult to do. I'm not sure if I should proceed before this is merged.

Agreed, you guys should coordinate on this to prevent unnecessary work.

dteviot:
@scrawl
>Is that a debug or release build? 
Windows, 32bit, Debug
>What graphics card are you using?
GTX 460

scrawl:
> Windows, 32bit, Debug

Ok, that explains it. Make sure to use a Release or RelWithDebInfo build if you want good performance. OSG Debug builds are really slow on Linux. I can only imagine it must be worse on Windows.

PR #698, opened 2015-07-29
Don't try to play bad idle animations (Prevents Log spam)

EmperorArthur:
This is mostly propagating the error up the stack so the game can do something about it.

Working on avoiding log spam from calling an animation that doesn't exist every frame.

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -1,6 +1,7 @@
 #include "aiwander.hpp"
 
 #include <cfloat>
+#include <iostream>
 
 #include <components/misc/rng.hpp>
 
@@ -621,12 +622,17 @@ namespace MWMechanics
         actor.getClass().getMovementSettings(actor).mPosition[1] = 0;
     }
 
-    void AiWander::playIdle(const MWWorld::Ptr& actor, unsigned short idleSelect)
+    bool AiWander::playIdle(const MWWorld::Ptr& actor, unsigned short idleSelect)
     {
         if ((GroupIndex_MinIdle <= idleSelect) && (idleSelect <= GroupIndex_MaxIdle))
         {
             const std::string& groupName = sIdleSelectToGroupName[idleSelect - GroupIndex_MinIdle];
-            MWBase::Environment::get().getMechanicsManager()->playAnimationGroup(actor, groupName, 0, 1);
+            return MWBase::Environment::get().getMechanicsManager()->playAnimationGroup(actor, groupName, 0, 1);
+        }
+        else
+        {
+            std::cerr<< "Attempted to play out of range idle animation \""<<idleSelect<<"\" for " << actor.getCellRef().getRefId() << std::endl;
+            return false;
         }
     }
 
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -392,17 +392,23 @@ namespace MWMechanics
         short unsigned& idleAnimation = storage.mIdleAnimation;
         idleAnimation = getRandomIdle();
 
+        //If we should be moving
         if (!idleAnimation && mDistance)
         {
             storage.mState = Wander_MoveNow;
         }
         else
         {
-            // Play idle animation and recreate vanilla (broken?) behavior of resetting start time of AIWander:
+            //Recreate vanilla (broken?) behavior of resetting start time of AIWander:
             MWWorld::TimeStamp currentTime = MWBase::Environment::get().getWorld()->getTimeStamp();
             mStartTime = currentTime;
-            playIdle(actor, idleAnimation);
             storage.mState = Wander_IdleNow;
+        }
+
+        //If we aren't going to just stand
+        if(idleAnimation)
+        {
+            playIdle(actor, idleAnimation);
         }
     }
 
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -67,7 +67,7 @@ namespace MWMechanics
         AiWander::WanderState mState;
         
         unsigned short mIdleAnimation;
-        std::vector<unsigned short> mBadIdles; //Idle animations that when called cause errors
+        std::vector<unsigned short> mBadIdles; // Idle animations that when called cause errors
 
         PathFinder mPathFinder;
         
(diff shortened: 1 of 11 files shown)


EmperorArthur:
Note that the issue is idle animations not existing.  Even when I do find an acceptable solution, I'll have to fix it if pull #691 is merged.

scrawl:
Well, this change on its own is kinda pointless. I'll wait for further developments before considering it.

EmperorArthur:
Here's the actual fix.  It's not the most elegant solution, but it still alerts the user to the problem, but doesn't fill the log with thousands of useless lines.

scrawl:
Do you have an example where your code goes into effect? I know I've seen this kind of log spam in the past, but can't seem to make it happen right now.

EmperorArthur:
I do, but I can't tell you which of my mods is causing it.  It's probably "Morrowind Rebirth".

At Fort Moonmoth:

With the patch:
Animation idle2 not found for mr_practice_dummy
Unable to play idle animation 2 for mr_practice_dummy
Animation idle3 not found for mr_practice_dummy
Unable to play idle animation 3 for mr_practice_dummy
Animation idle4 not found for mr_practice_dummy
Unable to play idle animation 4 for mr_practice_dummy

Without I get many lines that look like this every second:
Animation idle2 not found for mr_practice_dummy
Animation idle3 not found for mr_practice_dummy
Animation idle4 not found for mr_practice_dummy
...

EmperorArthur:
Yeah, that makes sense, but would have been a bit too much tabbing for my taste.
Instead I've taken advantage of your refactor to just return from the function where appropriate.

scrawl:
I like the idea, but would like to see a few things changed.

* We are printing the same error twice.
* Maybe we should check if the animation exists *before* attempting to play it.
* Unnecessary comments. Don't comment what each statement does, only comment why we are doing it.
* Please use ```// comment```, not ```//comment```, to stay consistent with the existing code.

scrawl:
Note that AiWander has direct access to the CharacterController (which is a recent change), so you can talk directly to that instead of implementing your changes in MechanicsManager.

EmperorArthur:
1.  We are printing the same error twice:
Not really happy about that, but are you ok with just relying on "Animation idle4 not found for mr_practice_dummy"?
2.  Maybe we should check if the animation exists before attempting to play it:
I don't know enough about the system to do this right now.  As is, propagating up the stack just means the check is made at a different point.
3.  Unnecessary comments. Don't comment what each statement does, only comment why we are doing it:
I'm not the smartest cookie, so I like to document what something means.  I suppose some of these, like 'idleAnimation' is '0' do nothing should be explained either in AiWanderStorage or getRandomIdle.
4:  Please use // comment, not //comment, to stay consistent with the existing code:
Will fix
5:  Note that AiWander has direct access to the CharacterController (which is a recent change), so you can talk directly to that instead of implementing your changes in MechanicsManager:
Don't know enough about it to feel confident doing this right now.

EmperorArthur:
Need feedback on that "more elegant solution"  I'm pretty sure that will permanently affect the save file.

Easily reverted if you think it's too intrusive.

My research has revealed a glitch in the AIWander implementation.  If an actor does not have an idle animation and can't move, then we will try again every other frame.
I'm looking for a fix now.

EmperorArthur:
Ok, this should be good now.

apps/openmw/mwmechanics/aiwander.hpp
```diff
@@ -151,6 +154,9 @@ namespace MWMechanics
             static const std::string sIdleSelectToGroupName[GroupIndex_MaxIdle - GroupIndex_MinIdle + 1];
 
             static int OffsetToPreventOvercrowding();
+
+            //Idle animations that when called cause errors
+            std::vector<unsigned short> bad_idles;
```
scrawl: Two problems:
* Use the mBadIdles notation for member variables.
* The AiPackage should not contain run-time state, just basic information about the package. This change as is would increase the memory usage for every actor in the game (even in inactive cells). Runtime state should be added to AiWanderStorage.

EmperorArthur:
One more time.

scrawl:
> Don't warn twice when unable to play an idle animation

Isn't that still a problem? From what I see we would get

```
+            std::cerr<< "Error in Actors::playAnimationGroup:  Unable to find " << ptr.getTypeName() << std::endl;
```
and

```
        std::cerr<< "Animation "<<groupname<<" not found for " << mPtr.getCellRef().getRefId() << std::endl;
```

scrawl:
Nevermind, that's a different error (actor doesn't exist).

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -67,6 +67,7 @@ namespace MWMechanics
         AiWander::WanderState mState;
         
         unsigned short mIdleAnimation;
+        std::vector<unsigned short> mBadIdles; //Idle animations that when called cause errors
```
scrawl: Comment formatting

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -391,18 +393,30 @@ namespace MWMechanics
         short unsigned& idleAnimation = storage.mIdleAnimation;
         idleAnimation = getRandomIdle();
 
+        // If we should be moving
```
scrawl: This comment should be removed

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -391,18 +393,30 @@ namespace MWMechanics
...
         if (!idleAnimation && mDistance)
         {
             storage.mState = Wander_MoveNow;
+            return;
         }
-        else
+        // If we aren't going to just stand
```
scrawl: and this one

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -391,18 +393,30 @@ namespace MWMechanics
...
+        if(idleAnimation)
         {
-            // Play idle animation and recreate vanilla (broken?) behavior of resetting start time of AIWander:
-            MWWorld::TimeStamp currentTime = MWBase::Environment::get().getWorld()->getTimeStamp();
-            mStartTime = currentTime;
-            playIdle(actor, idleAnimation);
-            storage.mState = Wander_IdleNow;
+            // If the idle animation actually exists
```
scrawl: and this one

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -391,18 +393,30 @@ namespace MWMechanics
...
+                    storage.mBadIdles.push_back(idleAnimation);
+                    storage.mState = Wander_ChooseAction;
+                    return;
+                }
+            }
         }
+        // Recreate vanilla (broken?) behavior of resetting start time of AIWander:
+        mStartTime = MWBase::Environment::get().getWorld()->getTimeStamp();
+        storage.mState = Wander_IdleNow;
+        return;
```
scrawl: Why a return on the last line of a function body? Looks very odd.

EmperorArthur: Yep, but I wanted consistency with all the other "storage.mState = ..." commands.
My natural development style assumes that I'm an idiot.

PR #699, opened 2015-07-30
Implement accurate moon settings (fixes #672)

slothlife:


scrawl:
That's awesome.

One more issue that was on my list: the rotation of the moon texture is off by about 45 degrees . Or rather, it does not match the trajectory. You can have a shot at this problem if you want, otherwise I will do after the merge.

slothlife:
Hmm, yeah, after looking at Morrowind, I think you're right. It's strange that the rotation isn't really reflected in the Morrowind scene graph. It's possible that it is rolled by the axis offset (so 35 degrees for Masser).

I tried a quick fix in Moon::setState(), but that rotated the moon out of its initial position. I suspect my weak understanding of 3D math is hurting me here. Let's just go ahead and merge it.

apps/openmw/mwworld/weather.cpp
```diff
@@ -37,6 +37,170 @@ namespace
     }
 }
 
+MoonModel::MoonModel(const std::string& name, MWWorld::Fallback& fallback)
```
scrawl: ```MWWorld::Fallback&``` could be ```const```.

slothlife:
Rather than trying to reuse Moon::setDirection(), I wrote some code that correctly adjusts the attitude of the moon. I believe this fixes the problem with the rotation of the moon texture relative to its trajectory.

Given that this pull request is already marked as Reviewed, I will not make the commit here unless whomever merges this request would like me to do so.

scrawl:
That's the solution I was going for, but it wasn't quite working yet.

Removed the reviewed label, so you can add another commit.

PR #701, opened 2015-07-31
Streamline AiWander::getRandomIdle()

EmperorArthur:
We were finding a GMST potentially hundreds of times each frame.
This should provide a decent speed boost in crowded areas.

I think I can do even better in the future, but that's a separate topic.
See https://forum.openmw.org/viewtopic.php?f=6&t=3019

Note:  This may cause merge conflicts with my other PR that touches AI wander.  I'll fix one of them after the other is merged.

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.hpp
+++ b/apps/openmw/mwmechanics/aiwander.hpp
@@ -151,6 +151,8 @@ namespace MWMechanics
             static const std::string sIdleSelectToGroupName[GroupIndex_MaxIdle - GroupIndex_MinIdle + 1];
 
             static int OffsetToPreventOvercrowding();
+
+            float fIdleChanceMultiplier;
     };
     
     
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -113,6 +113,8 @@ namespace MWMechanics
 
         mPopulateAvailableNodes = true;
 
+        fIdleChanceMultiplier = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fIdleChanceMultiplier")->getFloat();
+
     }
 
     AiPackage * MWMechanics::AiWander::clone() const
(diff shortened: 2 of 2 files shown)


EmperorArthur:
Example:

If you have a creature that can't doesn't move or play idle animations getRandomIdle() is called every other frame.  The original implementation looked up the GMST 8 times for every call.  Multiply this by several actors and you have an issue.

Morrowind Rebirth has training dummies as NPCs that trigger this exact issue.

apps/openmw/mwmechanics/aiwander.hpp
```diff
@@ -151,6 +151,8 @@ namespace MWMechanics
             static const std::string sIdleSelectToGroupName[GroupIndex_MaxIdle - GroupIndex_MinIdle + 1];
 
             static int OffsetToPreventOvercrowding();
+
+            float fIdleChanceMultiplier;
```
scrawl: Same issue as in #698 - avoid adding fields to AiPackage that don't belong in the package description.

EmperorArthur: Not quite the same.  This is a global, not part of an individual actor's state information.  Every actor will have the exact same value.
If I could figure out how it properly initialize it at runtime, it would be a static.

terrorfisch: Is there a reason to make it a member variable and not use the old implementation with a static variable in the function?

EmperorArthur:
Yep, find("fIdleChanceMultiplier") is, according to stack overflow, at least Log(n) complexity for the number of items in the std::Map.

Basically the old/current time of a function that's called multiple times per frame depends on the number of globals.  Worse, that implementation was preforming that lookup 8 times in a row.

It's a speed/memory trade off from the optimization point of view.  For the cost of an extra float we remove 5 function calls, one of which is O(Log(n)).

Actually, given that it was inside the for loop that's really 20 calls, including the costly find(...).

scrawl:
Now that I look at it again, the old implementation *was* already a static variable. That means we only find the GMST once on the first call of the function, on the next call it'll use the cached value.

<pre>
-            static float fIdleChanceMultiplier = MWBase::Environment::get().getWorld()->getStore()
-                .get<ESM::GameSetting>().find("fIdleChanceMultiplier")->getFloat();
</pre>

Which makes this PR pointless all together.

EmperorArthur:
Hmm, it appears github stops updating pull requests once they're closed.  You know, I was doing more than moving fIdleChanceMultiplier.

Oh well, who really cares about 16 casts and 8 floating point divisions each frame.  Or readability.

EmperorArthur:
Please re-open the pull request, and take a look at the other optimizations that were made.  They might be minor but they at least improve readability.

Github stops updating pull requests once they're closed so the rebase I just did doesn't show up here.

scrawl:
> Oh well, who really cares about 16 casts and 8 floating point divisions each frame. Or readability.

Precisely. If you have not found a bottleneck in profiling, then you are wasting your time. I do not consider micro-cleanup or micro-optimizations to be particularly effective or worthwhile.

zinnschlag:
Agreed.

PR #702, opened 2015-08-02
Reduced Running In Circles with AiCombat.

dteviot:


apps/openmw/mwmechanics/pathfinding.cpp
```diff
@@ -291,29 +296,28 @@ namespace MWMechanics
...
     }
 
+    bool PathFinder::isBetweenWaypoints(const ESM::Pathgrid::Point& pointToCheck,
+        const ESM::Pathgrid::Point& firstWayPoint,
+        const ESM::Pathgrid::Point& secondWayPoint)
+    {
+        float distBetweenPoints = sqrDistanceIgnoreZ(firstWayPoint, secondWayPoint);
+        float distTo2ndPoint = sqrDistanceIgnoreZ(pointToCheck, secondWayPoint);
+        return (distTo2ndPoint < distBetweenPoints);
+    }
```
scrawl: Seems problematic. I can easily imagine a path where the "between waypoint" check fails. With the path in the attached diagram, the new code would send the actor running against the wall.
path

dteviot:
I believe this problem is more theoretical than actual.
For this to occur:
1. You'd need a pathgrid and geometry like this.
2. You'd need to get an actor into that position.
3. You'd need the actor to generate a path when at that position.

I'd also note, 
1. We already have a similar problem of an actor getting stuck behind a wall (or other geometry) when the nearest pathgrid point is on the other side of the geometry.
2. The fundamental error is that when creating a path, if the start (or end) location is not on a pathgrid point or on the edge between points, we don't check that the point is reachable.  This is what I was planning on working on next weekend.

dteviot:
@scrawl,
I take it you still don't accept isBetweenWaypoints().
Would you accept it if it included a Line of Sight (LOS) check that the 2nd waypoint can be reached?
If so, I'll resubmit this PR with 248c157 removed, and submit 248c157 with a LOS check later.

scrawl:
> I take it you still don't accept isBetweenWaypoints().

I'm just a little reluctant. The issue I brought up may not exist in practice, but is there anything wrong with the previous solution?

> Would you accept it if it included a Line of Sight (LOS) check that the 2nd waypoint can be reached?

Not keen on adding more LOS checks. We are trying to get away from those, right? When I investigated AI a few months ago, the "successful LOS = run straight to target" logic was the main problem causing it to run against stair walls, fall off cliffs, etc. The pathgrid should be honored as much as possible.

dteviot:
> I'm just a little reluctant. The issue I brought up may not exist in practice, but is there anything wrong with the previous solution?

It doesn't work when AiCombat switches between LOS and following the path grid. The circling behaviour is shown in latest images posted to https://forum.openmw.org/viewtopic.php?f=6&t=2916

It also (sort of) doesn't work when you start a path from fresh and the closest point is behind you. And there's a similar behaviour when the final end point on a path is between the last two pathgrid points.
I suspect this weird circling is why the "shortcutting" was done on AiFollow, if you don't do shortcutting,  For short trips, the Actor goes on a really weird looking path.

>The pathgrid should be honored as much as possible.
I agree.  I suspect the problem with shortcutting is that it it being used for paths that are too long.
I think the LOS should only be done for checking that the "entry" and "exit" points onto the pathgrid can be reached.  (Incidentally, one problem I noticed with shortcutting is that it doesn't allow for the width of the actor.  In the sample pathgrid image you gave above, I suspect the Actor travelling to the top waypoint would wind up brushing against the wall and getting stuck.)

dteviot:
@scrawl 
I note the following existing code in AiCombat
https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/aicombat.cpp#L535-545
(Incidentally, how do I get the actual code snippet to show here?)
Conceptually, it's very similar to the "Test if 2nd node is better choice than first", except it's applied to the end node.
Anyway, my thought is this code should be moved into buildPath() in https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/pathfinding.cpp, so that all AI Packages can benefit from it.  (I'd also suggest that similar logic be applied to the first waypoint node.  (i.e. If final destination is closer to start than first waypoint of path grid, just head straight for the final destination.)

PR #705, opened 2015-08-03
Magic effect verifier

smbas:
The verifier performs the following checks:
* Is the base cost negative?
* Is the icon specified? Does the icon exist? **required string field**
* Is the particle specified? Does the particle (texture) exist?  **required string field**
* Is the description not empty? **required string field**
* Does the object (hit/area/bolt/casting) exist and has a proper type?
* Does the sound (hit/area/bolt/casting) exist?

The verifier checks all magic effects, even some of them aren't presented in a content file (such as #138SummonWolf in Morrowind.esm). This leads to unrelated errors. Ignore such effects would be the right solution, but there is no way to distinguish them.

Also I don't know what to do with an empty School field. Is it correct to have a magic effect with no school specified?

### DIFF
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -41,7 +41,7 @@ opencs_units (model/tools
 opencs_units_noqt (model/tools
     mandatoryid skillcheck classcheck factioncheck racecheck soundcheck regioncheck
     birthsigncheck spellcheck referencecheck referenceablecheck scriptcheck bodypartcheck
-    startscriptcheck search searchoperation searchstage pathgridcheck soundgencheck
+    startscriptcheck search searchoperation searchstage pathgridcheck soundgencheck magiceffectcheck
     )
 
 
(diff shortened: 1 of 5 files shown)


cc9cii:
I suspect that having no school will lead to unexpected NPC autocalc results.

zinnschlag:
> The verifier checks all magic effects, even some of them aren't presented in a content file (such as #138SummonWolf in Morrowind.esm). This leads to unrelated errors. Ignore such effects would be the right solution, but there is no way to distinguish them.

The obvious solution would be to modify the injected records (at the time of injection) in a way that they don't trigger errors. However I am not sure how to do that in case of icons and textures.

> I suspect that having no school will lead to unexpected NPC autocalc results.

I don't think you can have no school, since the school is specified via an index instead of an ID (for now). An invalid school index would make the record invalid.

smbas:
Currently the verifier doesn't allow empty Particle field, but in TES-CS it can be empty. Should I remove the check for an empty Particle?

zinnschlag:
I am not sure. Is OpenMW taking a effect record without particles?

scrawl:
Particle is a texture filename used to override the default texture for the hand casting effect. If it's empty, the original texture will be used. So I would say that's valid. https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/character.cpp#L1194

zinnschlag:
Okay. Then we are good?

PR #707, opened 2015-08-04
Add appdata file for openmw and install it

nobrakal:
Hi,

For the need of Fedora (https://fedoraproject.org/wiki/Packaging:AppData), I create an appdata file, (a sub part of the AppStream project, you can find more informations here: http://www.freedesktop.org/software/appstream/docs/chap-AppStream-About.html#about-whatis) . 
The purpose of this file is to give more informations to any software center on OpenMW.

I add myself as updatecontact (contact for the evolution of the appdata spec), because I don't know if openmw have a mailing list.
I also use old screenshot beceause I can't find more recent (and none for openmw-cs).

I also add basic rules to install it on CMakeLists :)

### DIFF
--- a/files/openmw.appdata.xml
+++ b/files/openmw.appdata.xml
@@ -8,13 +8,13 @@
  <summary>Unofficial open source engine re-implementation of the game Morrowind</summary>
  <description>
   <p>
-  OpenMW is a recreation of the engine for the popular role-playing game Morrowind by Bethesda Softworks. You need to own and install the original game for OpenMW to work.
+  OpenMW is a new engine for 2002's Game of the Year, The Elder Scrolls 3: Morrowind.
   </p>
   <p>
-  The main quests in Morrowind, Tribunal and Bloodmoon are all completable. Some issues with side quests are to be expected (but rare).
+  It aims to be a fully playable (and improved!), open source implementation of the game's engine and functionality (including mods). 
   </p>
   <p>
-  Pre-existing modifications created for the original Morrowind engine can be hit-and-miss, but a mod created for Morrowind may not necessarily run in OpenMW.
+  You will still need the original game data to play OpenMW.
   </p>
  </description>
 
--- a/files/openmw.appdata.xml
+++ b/files/openmw.appdata.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2015 Alexandre Moine <nobrakal@gmail.com> -->
+<application>
+ <id type="desktop">openmw.desktop</id>
+ <metadata_license>CC0-1.0</metadata_license>
+ <project_license> GPL-3.0 and MIT and zlib</project_license>
+ <name>OpenMW</name>
+ <summary>Unofficial open source engine re-implementation of the game Morrowind</summary>
+ <description>
+  <p>
+  OpenMW is a recreation of the engine for the popular role-playing game Morrowind by Bethesda Softworks. You need to own and install the original game for OpenMW to work.
+  </p>
+  <p>
+  The main quests in Morrowind, Tribunal and Bloodmoon are all completable. Some issues with side quests are to be expected (but rare).
+  </p>
+  <p>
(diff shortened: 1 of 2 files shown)


files/openmw.appdata.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2015 Alexandre Moine <nobrakal@gmail.com> -->
+<application>
+ <id type="desktop">openmw.desktop</id>
+ <metadata_license>CC0-1.0</metadata_license>
+ <project_license> GPL-3.0 and MIT and zlib</project_license>
```
scrawl: Which part of OpenMW is licensed under zlib?

files/openmw.appdata.xml
```diff
@@ -0,0 +1,39 @@
...
+ <summary>Unofficial open source engine re-implementation of the game Morrowind</summary>
+ <description>
+  <p>
+  OpenMW is a recreation of the engine for the popular role-playing game Morrowind by Bethesda Softworks. You need to own and install the original game for OpenMW to work.
+  </p>
+  <p>
+  The main quests in Morrowind, Tribunal and Bloodmoon are all completable. Some issues with side quests are to be expected (but rare).
+  </p>
+  <p>
+  Pre-existing modifications created for the original Morrowind engine can be hit-and-miss, but a mod created for Morrowind may not necessarily run in OpenMW.
```
scrawl: Not too happy about the duplication of readme.txt. If we really need a long description, please write one that isn't going to get outdated.

files/openmw.appdata.xml
psi29a: Files: extern/oics/tiny*
Copyright: 2000-2006, Lee Thomason (www.grinninglizard.com)
License: Zlib

Files: libs/openengine/bullet/*
Copyright: Copyright (c) 2012 Advanced Micro Devices, Inc.  http://bulletphysics.org
License: Zlib

files/openmw.appdata.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2015 Alexandre Moine <nobrakal@gmail.com> -->
+<application>
+ <id type="desktop">openmw.desktop</id>
+ <metadata_license>CC0-1.0</metadata_license>
```
scrawl: You are saying the file is licensed under CC0 public domain, but still have a copyright header at the top. The idea of the public domain is to waive copyrights, so if you are going to use that, there should be no copyright statement.

files/openmw.appdata.xml
scrawl: libs/ doesn't exist anymore. You are right about tinyxml though.

scrawl: Btw, we do have a cmake option to use the system library of tinyxml.

files/openmw.appdata.xml
kcat: I think that depends on jurisdiction. Some places don't allow you to specifically place something in the public domain, and instead you have to just use a very liberal license (which I presume what CC0 is; a copyright license that imposes as few restrictions as possible). In such a case, it still needs proper copyright attribution otherwise it's assumed All Rights Reserved.

files/openmw.appdata.xml
psi29a: 1) That will be updated in 0.37 release, but still holds true for 0.36.1 release.
2) We use the system library of tinyxml, but the files still exist in our sources. We have to appease the arm-chair lawyers and cover that too.

files/openmw.appdata.xml
nobrakal: You are totally right, I will update it soon (but english is not my native language, I apology in advance for english mistakes)

files/openmw.appdata.xml
nobrakal: For example, in France, my country, I am obliged to name myself, because a totally free license (like CC-0) is not legal. But the license still CC-0, even with my name :)

nobrakal:
Description rewrited :)
'appstream-util validate-relax' (the tool used to check .appdata.xml) require at least two <p>, this is why I have to make a "long" description.

PR #708, opened 2015-08-05
Fix several sky rendering bugs, maybe also #639

slothlife:
Added code to hide the moons, sun, and stars for certain weather
effects. Lightly refactored CelestialBody and derived classes. Fixed
moons switching phase at 24:00.

### DIFF
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -321,12 +321,12 @@ private:
     int mMeshType;
 };
 
-// A base class for the sun and moons. Must be stored in an osg::ref_ptr due to being
-// derived from osg::Referenced.
+/// A base class for the sun and moons.
+/// \note Must be stored in an osg::ref_ptr due to being derived from osg::Referenced.
 class CelestialBody : public SceneUtil::StateSetUpdater
 {
 public:
-    CelestialBody(osg::Group* parentNode, float scaleFactor = 1.f, int numUvSets=1)
+    CelestialBody(osg::Group* parentNode, float scaleFactor, int numUvSets)
     {
         mGeode = new osg::Geode;
         osg::ref_ptr<osg::Geometry> geom = createTexturedQuad(numUvSets);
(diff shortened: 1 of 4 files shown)


psi29a:
I noticed this too while working on the template... how can I see stars/moons/sun while it is storming/raining? :)  Thanks!

apps/openmw/mwrender/sky.cpp
```diff
@@ -336,42 +336,85 @@ class CelestialBody
...
+
+    virtual void setDefaults(osg::StateSet* stateset)
+    {
+        osg::ref_ptr<osg::TexEnvCombine> texEnv(new osg::TexEnvCombine);
+        texEnv->setCombine_Alpha(osg::TexEnvCombine::MODULATE);
+        texEnv->setSource0_Alpha(osg::TexEnvCombine::PREVIOUS);
+        texEnv->setSource1_Alpha(osg::TexEnvCombine::CONSTANT);
+        texEnv->setCombine_RGB(osg::TexEnvCombine::REPLACE);
+        texEnv->setSource0_RGB(osg::TexEnvCombine::PREVIOUS);
+        texEnv->setConstantColor(osg::Vec4f(1.0f, 1.0f, 1.0f, 1.0f));
```
scrawl: It's easier (and more efficient) to achieve transparency with the Material's diffuse color. You can find some examples elsewhere in sky.cpp.

apps/openmw/mwrender/sky.cpp
```diff
@@ -322,11 +321,12 @@ class ModVertexAlphaVisitor : public osg::NodeVisitor
     int mMeshType;
 };
 
-class CelestialBody
+// A base class for the sun and moons. Must be stored in an osg::ref_ptr due to being
+// derived from osg::Referenced.
```
EmperorArthur: Have you considered changing these to Doxygen style comments?
Like this (note the extra '/'):
/// A base class for the sun and moons.
/// \note Must be stored in an osg::ref_ptr due to being derived from osg::Referenced.

apps/openmw/mwrender/sky.cpp
slothlife: I'm still pretty weak when it comes to general graphics programming.

I've been struggling with trying to get this working with a diffuse alpha. I've tried using AlphaFader and by creating a material and setting the diffuse color myself and still can't get it to work, despite the fact that I can confirm that the material property is set correctly.

I have however removed the pointless dummy texture and done the modulation on the original texture and confirmed that works as expected.

Should I just push a commit with the broken code that uses the diffuse color, so maybe you can see what silly mistake I'm making? Or should I just commit the code using texture combiners (admittedly in a better way than the original code)?

apps/openmw/mwrender/sky.cpp
slothlife: Good point, I hadn't seen it used very much in the project, but the wiki does ask that they be used.

apps/openmw/mwrender/sky.cpp
scrawl: > Should I just push a commit with the broken code that uses the diffuse color, so maybe you can see what silly mistake I'm making?

Sounds good.

scrawl: I see the problem. You are using createAlphaTrackingUnlitMaterial(), that uses the vertex colors to override diffuse color, so the diffuse color you are setting has no effect. Use createUnlitMaterial() instead.

scrawl: I guess that method naming could be improved. createVertexColorTrackingUnlitMaterial()?

slothlife: Thanks for finding that! That fixed it.

I think I'm just confused by how glColorMaterial works. I thought that using that would permit the diffuse color to be set via glColor for the entire material. I was under the impression that with GL_LIGHTING enabled, that it would ignore the vertex colors, and instead use the material colors with the texture colors in the lighting equation to shade it properly.

I guess I need to read that documentation again!

scrawl:
I kept the Updater classes separate because an object should not reference itself. I believe your refactor introduces a memory leak in that regard.

apps/openmw/mwworld/weather.cpp
```diff
@@ -289,44 +294,54 @@ WeatherManager::WeatherManager(MWRender::RenderingManager* rendering, MWWorld::F
 
     //Weather
     Weather clear;
+    clear.mObstructsCelestialBodies = false;
```
scrawl: I did some experiments in vanilla MW to see if this is really hardcoded. I believe it's the
<pre>
Glare View=1
</pre>
weather setting.

slothlife: Regarding the memory leak: you're right. I'll reintroduce a separate object for that purpose. I wish it wasn't necessary, since the code is a lot smaller without it.

Regarding Glare View: I'll experiment tonight with different values. The name would leave me to believe it only affected the glare from the sun, but after playing with the various moon settings I would not be surprised if it did more than what says.

scrawl: On a related note, I wonder what "Clouds Maximum Percent" is supposed to do. It doesn't seem to affect the cloud transparency as it does in OpenMW. Only thing I noticed is that when set to "0", the moon texture becomes garbage.

slothlife: I think you're right on Glare View. I'll switch to using that value instead.

slothlife: I tried modifying Clouds Maximum Percent and really saw no difference. On Windows at least, the moon texture didn't appear to be impacted by setting the value to 0. Values in Morrowind.ini vary from 1.0, to 0.66.

slothlife: I can't resist a good mystery.

I've been experimenting with more values, and I THINK what Clouds Maximum Percent affects is the maximum amount of transparency that the clouds can start at. That means that you ordinarily won't see this happen, since the clouds are normally opaque, BUT you can see it quite easily during a weather transition. Try changing Rain's max % to 0.0 and then change the weather from something like Clear to Rain in MW: you should see the rain clouds immediately pop into the sky, fully opaque. It's kind of funny since the sky color hasn't finished transitioning so there's a dark grey set of clouds with a cheerful blue horizon still there.

slothlife: After trying more values, I think it actually may describe how quickly to transition the clouds to the new texture. If you set small values, you'll see that the clouds transition quickly, but not immediately, to the new texture. It doesn't appear to have an impact when transitioning from a 0.0 -> 1.0, so I think it only matters what the target weather pattern is, not what it was.

Rather than continue to clutter the request with more research findings, I'd prefer to take this to the forums and eventually submit a pull request later, once it is understood a little better.

scrawl:
This should fix bug 639 indeed. Do you have an account on the bugtracker so I can assign it to you?

slothlife:
Yes, Chris Boyce / slothlife.

PR #711, opened 2015-08-08
Update OS X packaging to work with OSG

corristo:
TODO:
- [x] Add attribution & licence to borrowed cmake modules


zinnschlag:
Merge now or wait for the to do part? Ideally license/attributing statement should be added at the same time as the borrowed code.

corristo:
@zinnschlag I'll add an attribution before merge.

corristo:
@zinnschlag @scrawl added the attribution, is it okay?

scrawl:
Looks ok to me.

PR #712, opened 2015-08-09
Refactoring of AiCombat

dteviot:
Mostly commit 1d4be08f6e4c2dbd89cc0c3408a8231ee4497277 from mrcheko's PR #614

### DIFF
--- a/apps/openmw/mwmechanics/aipackage.cpp
+++ b/apps/openmw/mwmechanics/aipackage.cpp
@@ -133,7 +133,7 @@ bool MWMechanics::AiPackage::doesPathNeedRecalc(ESM::Pathgrid::Point dest, const
 
 bool MWMechanics::AiPackage::isTargetMagicallyHidden(const MWWorld::Ptr& target)
 {
-    MagicEffects& magicEffects(target.getClass().getCreatureStats(target).getMagicEffects());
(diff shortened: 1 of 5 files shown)


EmperorArthur:
Neat.

Shouldn't there be some sort of a check in aifollow and, maybe, aiescort for "isTargetMagicallyHidden(...)"?

What's the ai action in Vanilla MW when the person they're following goes invisible?

apps/openmw/mwmechanics/aipackage.cpp
```diff
@@ -129,3 +130,10 @@ bool MWMechanics::AiPackage::doesPathNeedRecalc(ESM::Pathgrid::Point dest, const
 {
     return distance(mPrevDest, dest) > 10;
 }
+
+bool MWMechanics::AiPackage::isTargetMagicallyHidden(const MWWorld::Ptr& target)
+{
+    MagicEffects& magicEffects(target.getClass().getCreatureStats(target).getMagicEffects());
```
scrawl: Should be ```const MagicEffects&```.

dteviot:
> What's the ai action in Vanilla MW when the person they're following goes invisible?

In Vanilla, a AiFollow has no trouble following the player when the player is invisible.

@scrawl
> Should be const MagicEffects&.
Done.  Thanks.

PR #713, opened 2015-08-09
Starting work improve pathfinding

dteviot:
@scrawl

I suspect lines 224-228 are incorrect.
When pathgrid returns a single node, it's being discarded and actor walks directly to destination.
Needs to build a path using the pathgrid node and the destination.

I also suspect that lines 232-245 may not be needed, 
aStarSearch() should not return an empty set as we're supplying two different points for a path.
Only way I can see aStarSearch() could return an empty path is if the path data we've been given is wrong. I have yet to check to see if aStarSearch() handles that case.

### DIFF
--- a/apps/openmw/mwmechanics/pathfinding.cpp
+++ b/apps/openmw/mwmechanics/pathfinding.cpp
@@ -51,8 +51,7 @@ namespace
                                                   const MWWorld::CellStore *cell,
                                                   const osg::Vec3f pos, int start)
     {
-        if(!grid || grid->mPoints.empty())
-            return std::pair<int, bool> (-1, false);
+        assert(grid && !grid->mPoints.empty());
 
         float distanceBetween = distanceSquared(grid->mPoints[0], pos);
         int closestIndex = 0;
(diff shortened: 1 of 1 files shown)


scrawl:
I'm having trouble following your line numbers. Are those for the previous code or for your modified code?

dteviot:
Those are the line numbers AFTER the patch is applied.  (Note, they're existing code that I haven't removed/fixed yet.)
In the pre-PR code they are:
https://github.com/OpenMW/openmw/blob/3046795a9c1197ed7fc9766fe78de84e37da1672/apps/openmw/mwmechanics/pathfinding.cpp#L228-232
and 
https://github.com/OpenMW/openmw/blob/3046795a9c1197ed7fc9766fe78de84e37da1672/apps/openmw/mwmechanics/pathfinding.cpp#L236-249
respectively.

Incidentally, how do you get the actual code lines to show in this comment?

apps/openmw/mwmechanics/pathfinding.cpp
```diff
@@ -51,8 +51,7 @@ namespace
                                                   const MWWorld::CellStore *cell,
                                                   const osg::Vec3f pos, int start)
     {
-        if(!grid || grid->mPoints.empty())
-            return std::pair<int, bool> (-1, false);
+        assert(grid && !grid->mPoints.empty());
```
scrawl: This change breaks consistency with getClosestPoint. Either make this change in both functions or change none of them.

scrawl:
> When pathgrid returns a single node, it's being discarded and actor walks directly to destination.
Needs to build a path using the pathgrid node and the destination.

> I also suspect that lines 232-245 may not be needed, 
> aStarSearch() should not return an empty set as we're supplying two different points for a path.

Makes sense.

PR #714, opened 2015-08-11
OS X: enable high resolution mode for OpenMW & OpenCS

corristo:
We don't support retina completely yet, but it's still better than
blurry mess, and Qt does pretty good job for OpenCS already.


corristo:
Please don't merge yet, I just read about recently UI scaling support, I'd try to use scale = 2 as default on retina screens.

corristo:
@zinnschlag let's merge this now, further changes require more investigation.

PR #715, opened 2015-08-13
Utility functions for getting magical effect magnitude.

dteviot:


scrawl:
Not sure about this. I don't see the extra typing as a problem. Ease of refactoring could be an argument, but then again Class might not be the right place for the function. Also, the function names are missing the get prefix. I'm curious about Zini's opinion.

dteviot:
> I don't see the extra typing as a problem

I'm not worried about the extra typing, I'm trying to make the code more readable.
I was initially trying (hoping) to convert something like this
```C++
player.getClass().getCreatureStats(player).getMagicEffects().get(ESM::MagicEffect::Paralyze).getMagnitude() > 0
```
into 
```C++
player.magicEffects[ESM::MagicEffect::Paralyze)] > 0
```
unfortunately, this is not easily doable.  The best I could easily do was this:
```C++
player.getClass().magicEffectMagnitude(player, ESM::MagicEffect::Paralyze) > 0
```
which I think is still an improvement.  And I think will make heading to the ultimate goal easier.

> Class might not be the right place for the function.

Agreed, its not ideal.  But at the current time, it's what we've got.

Aside, I suspect I could add a float conversion operator to the MagicEffect class.  That would eliminate the need for the getMagnitude() function.

dteviot:
Actually, my prior statement wasn't entirely correct.
What actually started this was the observation that in our code to EffectParam::getMagnitude() were always preceded by get(<EffectKey>).
i.e. Code always went 
```C++
get(<EffectKey>).getMagnitude()
```

So, it seemed to make sense to replace them these with the "MagiceEffects::magnitude(<EffectKey>)" call.
The rest of the changes snowballed from there, applying the same concept. 

Also, getMagnitude wasn't always preceeded by get(<EffectKey>), there was 1 place where it was not.

zinnschlag:
Not in favour of this change. The missing get criticism is valid, no matter if it was used consistently before or not. But we really should try to keep the Class interface minimalistic. There is already enough bloat without convenience functions.

I agree that the code could be more readable. But that might be more a case of restructuring it than re-doing the interface. For example in creature.cpp the code is accessing the levitate magnitude twice. This is slightly inefficient but more importantly it is redundant. Maybe restructure it in a way that the value is stored in a local variable?

Random early morning half-cooked thought: Can operator overloading be used to make the code more readable? The MagicEffect::get for example would look more natural as operator[].
Ptr does not have much pointer syntax and getClass is kinda tacked on. Maybe it would make sense to add operators * and -> for accessing the class.

dteviot:
> Maybe restructure it in a way that the value is stored in a local variable?

I've done that (moving common sub-expressions) in a number of places in this PR. e.g. Combat.cpp.  I missed the levitate instances though.

> The MagicEffect::get for example would look more natural as operator[].

There are only two places in OpenMW code where getMagnitude() is not called after calling MagicEffects::get(const EffectKey& key).
1. Inside MagicEffects::setModifiers() and
2. CreatureStats::modifyMagicEffects()

Thus, my initial change of adding MagicEffects::magnitude(effectKey), as I believe
```c++
magnitude(<effectKey>) 
```
is more readable than.
```c++
get(<effectKey>).getMagnitude()
```

I also believe it's more readable than
```c++
[<effectKey>].getMagnitude()
```

Given that, I believe that replacing MagicEffects::get() with MagicEffects::operator[] is not worth the effort.

I will agree that 
```c++
getMagnitude(<effectKey>) 
```
is more readable than magnitude(effectKey).
In your opinion, would getMagnitudeOf(effectKey) be even better?

Extending the above logic, expressions like this

```c++
if (player.getClass().getCreatureStats(player).getMagicEffects().get(ESM::MagicEffect::Paralyze).getMagnitude() > 0
```

became

```c++
if (player.getClass().getCreatureStats(player).magicEffectMagnitude(ESM::MagicEffect::Paralyze) > 0
```

by adding 
CreatureStats::magicEffectMagnitue(const EffectKey& key).

Aside: in the case of ESM::MagicEffect::Paralyze, I think an even more readable solution is adding 
CreatureStats::isParalyzed().

I then extended this principle further (too far) when I added Class::getMagicErects() and Class::magicEffectMagnitude().


> Ptr does not have much pointer syntax and getClass is kinda tacked on. Maybe it would make sense to add operators * and -> for accessing the class.

I think the fundamental problem is that the Class class seems to be a "dumping group" for assorted functions.
My thoughts on the correct fix is to adjust the class hierarchy, moving the functions in class into the correct derived classes.
e.g. canFly() is only relevant for an NPC or Creature.  (I'm suspecting there's a missing class in the hierarchy that should have behaviours common to NPC and creature. "Character" perhaps?)
Of course doing this would require we use correctly typed pointers, rather than passing World::ptr objects around.
And that would probably require considerable code restructuring (and possibly some pointer casting.)
I will point out, I haven't looked into this issue in depth, so I'm probably missing something obvious why this isn't a viable approach.
I'd be willing to look into this if you thought this was a viable approach, but at current time, I'd really like to get the pathfinding issues sorted.

Side thought:
```c++
if (player.getCreatureStats().isParalyzed()
```

doesn't look too bad, but that would require pushing at least some of the functions that are in Class into World::Ptr.

EmperorArthur:
World::ptr is an abomination.

While there's no easy way to fix it all at once, we could start using more composition, and calling things directly.  For instance, here's the original function:
>Creature::getEncumbrance (const MWWorld::Ptr& ptr) const
>...
>float weight = getContainerStore (ptr).getWeight();	 
>const MWMechanics::CreatureStats& stats = getCreatureStats (ptr);

Every creature has an inventory, and stats, so we could move things around, and do something like:
>Creature::getEncumbrance (const MWWorld::Ptr& ptr) const
>...
>float weight = getCreature(ptr).containterStore.getWeight();	 
>const MWMechanics::CreatureStats& stats = getCreature(ptr).stats;

Then you're just a hop skip and jump away from factoring World::Ptr out of that function.

zinnschlag:
I do not see a fundamental problem here. Neither with Class nor with Ptr. They are both adequate solutions for the world/object model we are currently using (of course the object model itself is a different issue and we may consider reviewing it OpenMW 2.0-ish, which may factor both Class and Ptr out of existence).

Yes, the code is verbose. But that is not a bad thing. It only becomes an issue when the verbosity comes from repeating the same piece of code within a function (the levitate example).

A polymorphic or templated Ptr is a no go, because the main purpose of Ptr is to allow uniform access to objects without consideration of the type. Any change here would result in either destroying this property or at least losing the ability to pass Ptr by value.

I find the long chains of function calls actually a bit clearer than the version with the convenience functions, because they state more explicitly what is happening. But if you absolutely must, how about a collection of free utility functions? This is C++ after all and not scrawny Java. We don't have to use OOP all the time.
```C++
float getCreatureMagicEffectMagnitude (MWWorld::Ptr actor, const EffectKey& key)
{
  return actor.getClass().getCreatureStats (actor),getMagicEffects().get (key).getMagnitude();
}
```

jeffreyhaines:
You could always do it this way to shorten your code. *Excuse my lack of type knowledge...

```C++
MagicEffects magicEffects = player.getClass().getCreatureStats(player).getMagicEffects();
if (magicEffects.get(ESM::MagicEffect::Paralyze).getMagnitude() > 0 )
{
    // Code
}
```

For me I appreciate enums and this ESM::MagicEffect::Paralyze functions like such to me.

PR #718, opened 2015-08-16
Added isParalyzed() and isPlayer() functions.

dteviot:


zinnschlag:
Can we move the isPlayer function to a different place? MWBase is meant to be just interfaces and access to sub systems. MMMechanics sounds more like it (assuming you want to add more actor related convenience functions, otherwise it would be MWWorld). Preferably in a new cpp file (actorutil.hpp/cpp maybe?).

I am also not so keen on the function itself. Seems to be more narrowly defined than it needs to be. How about a getPlayer function instead? ptr==getPlayer() is no less clear than isPlayer (ptr) and getPlayer() can be used in other context too.

dteviot:
> Can we move the isPlayer function to a different place? 

Will do.

> How about a getPlayer function instead?

I was going to ask you about also adding a getPlayer() function.  I wanted to get your feedback on the isPlayer() first, before I attempted it.

> ptr==getPlayer() is no less clear than isPlayer (ptr) 

You're obviously smarter than me, as I find isPlayer(ptr) more clear than ptr == getPlayer()
With ptr == getPlayer(), I have to parse it in multiple stages when I see it.
Goes something like.
1. I have a pointer, I'm testing it for equality.
2. The something is a function()
3. Function is returning the player
4. I am comparing a pointer for equality to player.
5. conclusion  => I am testing if ptr is the player.

When I see "isPlayer(ptr)" it's much closer to the intent.
Ideally I'd like the code to go 
if (ptr.isPlayer()) 

Which is exactly what I would say in normal English.  But I'll take what I can get.

My personal preference is to add both isPlayer() and getPlayer() functions.  Is that acceptable?

While I have you.  I've noticed multiple places where the test for isWerewolf goes something like.

ptr.getClass().IsNpc()  && ptr.getClass().getNpcStas().isWerewolf().

I'm thinking this can be improved by adding a virtual isWerewolf() function to CreatureStats.  For CreatureStats, it would return false.  And obviously it's overridden in NpcStats.  
(Alternately, move the mIsWerewolf boolean and isWerewolf() function to CreatureStats and set mIsWerewolf to false in the constructor.  But, I think the virtual function approach better reflects the intent.) Any thoughts?

zinnschlag:
> When I see "isPlayer(ptr)" it's much closer to the intent.
> Ideally I'd like the code to go
> if (ptr.isPlayer()) 

I do not agree with this position. Okay, that may apply particularly for someone with a stronger background in mathematics, but a=b is much clearer to me than isOfTypeB (a). The former explicitly states what is happening with only the language primitives, while the later introduces a name without providing any useful additional abstraction.
Yes, isPlayer is closer to English, but when it comes to clarity a mathematical notation will beat English every time.

> My personal preference is to add both isPlayer() and getPlayer() functions. Is that acceptable?

Not really. With an isPlayer method and a getPlayer method we would have two methods on the same abstraction level for checking if an object is the player. While these utility functions don't bloat any classes interface, we should still try to keep them minimal. I really do not see the pioint of isPlayer.


> ptr.getClass().IsNpc() && ptr.getClass().getNpcStas().isWerewolf().

> I'm thinking this can be improved by adding a virtual isWerewolf() function to CreatureStats. For CreatureStats, it would return false. And obviously it's overridden in NpcStats.

Or you could add just another utility global utility function IsWerewolf that is implemented via the expression quoted above. CreatureStats isn't actually a polymorphic class and making it one just for the sake of isWerewolf is overkill.

btw. I am closing this PR. Move any further discussion to the new PR please.

PR #722, opened 2015-08-16
Bug #2281 NPCs don't scream when they die

jeffreyhaines:
Added dying grunt for actor deaths

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -18,6 +18,7 @@
 #include "../mwbase/world.hpp"
 #include "../mwbase/environment.hpp"
 #include "../mwbase/windowmanager.hpp"
+#include "../mwbase/dialoguemanager.hpp"
 #include "../mwbase/soundmanager.hpp"
 #include "../mwbase/mechanicsmanager.hpp"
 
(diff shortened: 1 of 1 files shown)


scrawl:
I don't think this is the right way to do it. As I wrote in the issue description:

> There is a "scream" soundgen tag in the animation

Which, when encountered, presumably should trigger the "hit" voice dialogue entry.

jeffreyhaines:
So say we get the soundgen tag "scream" and we return with "hit" string.
When we go to the play the sound, it is safe to ask, if ("hit") say hit?

Should I use another string to return? Safer than "hit"?

scrawl:
I presume you're talking about the Npc::getSoundIdFromSndGen function.

"hit" is not a sound ID, it's a dialogue topic ID.

jeffreyhaines:
Right so how do we get from Npc::getSoundIdFromSndGen function with "scream" to play the hit sound?

jeffreyhaines:
Don't we want to process the tag, and return "THIS_IS_THE_I_WAS_KILLED_TAG"
so that when we run the soundgen code we call the dialogmgr.say(ptr, "hit")...

scrawl:
> Don't we want to process the tag, and return "THIS_IS_THE_I_WAS_KILLED_TAG"
so that when we run the soundgen code we call the dialogmgr.say(ptr, "hit")...

That's what I would call an awful hack. Better do some refactoring. For example, we could have the function handle playing the sound, instead of returning the ID to play.

<pre>
void Class::handleSndGen(const MWWorld::Ptr&, const std::string&)
</pre>

jeffreyhaines:
Ok. Let me see what I can do.

PR #724, opened 2015-08-16
Make SetDelete a no-op for items in containers (Fixes #2864)

scrawl:
For vanilla MW compatibility, SetDelete should be a no-op for items in containers. You can use RemoveItem instead.

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -256,6 +256,7 @@ namespace MWBase
             virtual void fixPosition (const MWWorld::Ptr& actor) = 0;
             ///< Attempt to fix position so that the Ptr is no longer inside collision geometry.
 
+            /// @note No-op for items in containers. Use ContainerStore::removeItem instead.
             virtual void deleteObject (const MWWorld::Ptr& ptr) = 0;
             virtual void undeleteObject (const MWWorld::Ptr& ptr) = 0;
 
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1050,7 +1050,7 @@ namespace MWWorld
 
     void World::deleteObject (const Ptr& ptr)
     {
-        if (!ptr.getRefData().isDeleted())
+        if (!ptr.getRefData().isDeleted() && ptr.getContainerStore() == NULL)
         {
             ptr.getRefData().setCount(0);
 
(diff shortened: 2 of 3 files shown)


zinnschlag:
RemoveItem is actually not a proper replacement for SetDelete, because RemoveItem works via ID and SetDelete works on the current object. With SetDelete an object can delete itself from any container, which is not possible with RemoveItem. SetDelete also works properly when there are multiple stacks with the same ID. Removeitem doesn't.

Also, a bit of googling indicates that in vanilla SetDelete on an container item actually does not work properly (which is consistent with what I remember from my own scripting experience). The plugin in the bug report may just be lucky that it doesn't get a crash.

How about this: We go with the no-op variant for now, but write a warning message to the console (indicating that the script is defective). Then at least the problem will not pass silently any more; kinda like flagging a language feature as depreciated to allow it to be removed/changed for later versions of the language.

scrawl:
With the script in question (and presumably other scripts), adding a warning message to SetDelete we would get that message spammed every frame. That isn't very nice.

> (indicating that the script is defective)

Using a no-op does not make a script defective, at least not by my definition of defective.

scrawl:
I also noticed that, in its current implementation, World::deleteObject does not correctly delete container items. For example, it does not flag the container as modified, adjust the equipment slots, or the selected enchanted item slot. See the comment in RefData:

<pre>
            /// \warning Do not call setCount() to add or remove objects from a
            /// container or an actor's inventory. Call ContainerStore::add() or
            /// ContainerStore::remove() instead.
</pre>

So making that a no-op actually makes sense from an implementation standpoint too.

zinnschlag:
> Using a no-op does not make a script defective, at least not by my definition of defective.

I meant for vanilla MW. Since you shouldn't call setDelete on an inventory item in vanilla the script would have to be flagged as broken and we are free to do what we want in this situation. As I wrote before I am okay with going for no-op for now (especially if the current implementation is broken), but it is an unintuitive solution and kills a very useful piece of script functionality that I was planning to put to use in Redemption after we transitioned to OpenMW 1.0. :(

We will have to add that back after 1.0 in some way. I guess we could add a new script instruction that covers both the old use cases for setDelete and inventory delete. Seems redundant though. I would have preferred to use SetDelete, which is a problem if we now proclaim that its okay (by not warning about) to call setDelete on inventory items with the no-op implementation.

Anyway, I will merge. The topic is not worth having a prolonged discussion about it.

PR #726, opened 2015-08-17
Added isParalyzed() and isPlayer() functions

dteviot:
Resubmit of #718, with isPlayer() moved to MWMechanics namespace.

### DIFF
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -78,7 +78,7 @@ add_openmw_dir (mwclass
     )
 
 add_openmw_dir (mwmechanics
-    mechanicsmanagerimp stat creaturestats magiceffects movement
+    mechanicsmanagerimp stat creaturestats magiceffects movement actorutil
     drawstate spells activespells npcstats aipackage aisequence aipursue alchemy aiwander aitravel aifollow aiavoiddoor
     aiescort aiactivate aicombat repair enchanting pathfinding pathgrid security spellsuccess spellcasting
     disease pickpocket levelledlist combat steering obstacle autocalcspell difficultyscaling aicombataction actor summoning
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -704,8 +704,8 @@ void OMW::Engine::activate()
         return;
 
     MWWorld::Ptr player = MWBase::Environment::get().getWorld()->getPlayerPtr();
-    if (player.getClass().getCreatureStats(player).getMagicEffects().get(ESM::MagicEffect::Paralyze).getMagnitude() > 0
-            || player.getClass().getCreatureStats(player).getKnockedDown())
+    const MWMechanics::NpcStats &playerStats = player.getClass().getNpcStats(player);
+    if (playerStats.isParalyzed() || playerStats.getKnockedDown())
         return;
 
     MWWorld::Ptr ptr = MWBase::Environment::get().getWorld()->getFacedObject();
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -605,11 +605,9 @@ namespace MWClass
     {
         float weight = getContainerStore (ptr).getWeight();
 
-        const MWMechanics::CreatureStats& stats = getCreatureStats (ptr);
-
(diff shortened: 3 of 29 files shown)


dteviot:
> CreatureStats isn't actually a polymorphic class 

I'm confused. NpcStats already derives from CreatureStats, doesn't that make CreatureStats polymorphic?

>Not really. With an isPlayer method and a getPlayer method we would have two methods on the same abstraction level for checking if an object is the player.

With respect, I disagree.  getPlayer() is not just for checking if an object is the player.  Or it won't be once I implement it.  It's a function that returns a player object. So we have two functions with different purposes:
.
getPlayer() returns an object that is a player.
isPlayer(actor) checks if "actor" is a player.

using getPlayer() to implement isPlayer() is an implementation detail, that exploits the (currently valid) assumptions that 
1. There is only one player object.
2. Obtaining the single player object is efficient.

I'll also point out that until recently (I think it was Scrawl's OSG port) that I've seen the isPlayer() logic expressed as something like 
actor->getClass().mName == "player"

zinnschlag:
> I'm confused. NpcStats already derives from CreatureStats, doesn't that make CreatureStats polymorphic?

No virtual functions. Deriving NpcStats vom CreatureStats is actually no more than a hack. It does not serve any purpose other than a small bit of convenience for the code using the stats classes.


Regarding getPlayer. The assumptions you are listing are valid not only currently. It is reasonable to assume that this will never change. If that were not the case, then of course you would be right.

A potential multi-player implementation could be an issue. But I don't think we will end up with something that would require multiple player objects. On the client side there will only ever be one player. And a quick sample I took of code that might end up on the server side finds only uses of ==getPlayer that won't survive the de-hardcoding phase (which will definitely happen before any kind of multi-player implementation).

dteviot:
Which of the following is easier to understand?
```C++
if( mPtr == MWMechanics::getPlayer() && !isAnimPlaying(mCurrentDeath) )
{
  …
}
```

```C++
if( MWMechanics::isThePlayer(mPtr) && !isAnimPlaying(mCurrentDeath) )
{
  …
}
```

Is "isThePlayer()" more clear than "isPlayer()"?

zinnschlag:
The first one. Clearly. Doesn't even require reading. The == is obvious at a glance.

> Is "isThePlayer()" more clear than "isPlayer()"?

Nope. Still a function name that needs to be read and parsed.

dteviot:
> The first one. Clearly. Doesn't even require reading. The == is obvious at a glance.

You're obviously a much better programmer than I.,You can read and parse the first one at a glance. I can not.

I have to look at it, then break it into pieces. I see the && and == operators and have to spend a second recalling the precedence rules to realize I have to evaluate the == first.  I then need to extract out the bit up to the && to get the "mPtr == Mechanics::getPlayer()" piece.
I then need to read and parse the getPlayer() function evaluate it and apply the equality operator.  I then need to realize that mPtr is a pointer, so getPlayer() is probably also a pointer,  I'm therefore testing two pointers for equivelence.  i.e. Are these pointers the same?  ==> I testing is mPtr the Player.

Aside: Obviously if there were brackets around (mPtr == Mechanics::getPlayer()) then I don't need to figure out the precedence, but then I need to do a "find the matching bracket", so there's not a lot of gain.

I believe that that most programmers are at my level of skill, not yours. Therefore writing it in the way that is easiest for me to read would also be the way that is easiest for them to read.  (OK, I've probably got far to big an ego.)  However, if you wish the code written in a way that easier for you to read, well, you're the one with ultimate say.

zinnschlag:
> I believe that that most programmers are at my level of skill, not yours.

That's not a question of how good you are as a programmer. I am pretty sure that I would have preferred the version with the == when I was just starting to write code in earnest (BASIC and I was fairly terrible at that time like any other beginner). It may just be an issue of different people's mind's working differently. As such I don't necessarily agree with a statement that your position is more common than mine.

Anyway, we are getting into nitpicking territory here. I wouldn't have pushed the issue so much only for the readability aspect. But there is also the issue of keeping the set of helper functions minimal and doing it my way saves us one function.

So yes, please go with the == option. The rest of the PR looks good. Once this change is in, we can have a merge.

dteviot:
Resubmitted as #735

PR #727, opened 2015-08-18
Bug #2281 NPCs don't scream when they die

jeffreyhaines:


jeffreyhaines:
Is this good? I see I can get the PlayType from the code within. For example we know left right and land. Then I can put back the parsing code in Character.cpp and pass with soungen, volume and pitch.

scrawl:
Seems like we are on the wrong track. Inspecting the death animations in base_anim.nif closer, some of them do include "moan" "roar" or "scream" keys, but not all do. Also, if you kill an NPC in vanilla MW with the console, they immediately scream even before you close the console, not during animation playback.

Furthermore, it doesn't seem like say(ptr, "hit"); is the correct scream either. Test with Ranis Athrys in Balmora, Guild of Mages. The screams generally sound different in MW, also, in OpenMW you'll get voice lines like "Stupid!" as well that are not screams.

More research needed, I guess.

scrawl:
> Furthermore, it doesn't seem like say(ptr, "hit"); is the correct scream either. Test with Ranis Athrys in Balmora, Guild of Mages. The screams generally sound different in MW, also, in OpenMW you'll get voice lines like "Stupid!" as well that are not screams.

Ignore that. My vanilla install apparently has german voice files, my OpenMW install english voices. That's why it seemed different to me.

scrawl:
With that said, looks like your first PR was correct all along. Sorry for the waste of time.

jeffreyhaines:
I'm doing some more refactoring. I think its better to have the handleSoundGen with passed soundgen string, float volume and pitch.

PR #728, opened 2015-08-18
Remove empty line at the beginning of files

scrawl:
git ls-files -z | xargs -0 sed -i '1{/^$/d}'


jeffreyhaines:
I was thinking about doing this. good catch

zinnschlag:
I actually prefer having a blank line at the beginning of a source file (which is probably why there are so many). Habit, I guess. Anyway, not getting into a discussion about that. Will merge.

PR #732, opened 2015-08-19
Time played displayed in save/load menus (Feature #2528)

jeffreyhaines:


scrawl:
The problem is, we don't have a localization string for "Time played: ". The text will look out of place next to the other labels in the save menu that are all localised to the user's MW install language. I suggest not showing Time Played unless the user specifically enables it via a new setting in settings.cfg. @zinnschlag, any other ideas?

jeffreyhaines:
How do we create localization for it? This is something the community will have to create. 
I should have it enabled in the settings.cfg. There was a user on the IRC channel who said he didn't want to see it out of his own shame and love for the game.

I'd also like to have the ability to change the 24 hour clock if we can. I can't read 24 hour clocks well. But thats for another day.

scrawl:
> How do we create localization for it?

If I correctly recall @zinnschlag's position: We can't do that yet. The current localisation system we inherited from Morrowind is inadequate and will be replaced post-1.0 by a new system anyway.

jeffreyhaines:
Ah well then. Maybe I can use some universal characters?

I'm looking into how to do settings now

jeffreyhaines:
I honestly don't think its a big deal that its in English.. the whole settings file is in English so some basic understanding is needed.

Either way I added a setting to turn time played on. Although I think its clutch to have it on as default...

But... Pushback hehe.

apps/openmw/mwgui/savegamedialog.cpp
```diff
@@ -361,6 +378,11 @@ namespace MWGui
             << MWBase::Environment::get().getWorld()->getMonthName(mCurrentSlot->mProfile.mInGameTime.mMonth)
             <<  " " << hour << " " << (pm ? "#{sSaveMenuHelp05}" : "#{sSaveMenuHelp04}");
 
+        if (Settings::Manager::getBool("timeplayed","Saves"))
+        {
+            text << "\n" << "Time played: " << formatTimeplayed((int)floor(mCurrentSlot->mProfile.mTimePlayed));
```
scrawl: Using ```int``` in one place and ```long int``` in another place is inconsistent. Also, why not just use a ```double```, mTimePlayed is a ```double``` after all. Why the need to cast anything?

jeffreyhaines: It makes it easier to do the math... Cleaner code to pass the format function a long int than a double... Because then you'll get into using fmods and floors anyway

jeffreyhaines: Maybe we should change mTimePlayed to be in long int and not double since we don't need milliseconds.

jeffreyhaines:
@zinnschlag Should we enable this by default even though we don't have localization yet? It's a pretty cool feature that our users should see. I'm thinking inexperienced players will be unaware of it.

zinnschlag:
Sorry for the long wait. Yes, we can't do additional localisation yet. At least not cleanly.

IIRC we already have some unlocalised strings in the settings, so the damage is already done and we can as well go ahead with having one more string. The cleanup we will have to do once we hit 1.0 won't be pretty.

Anyway, will merge.

PR #734, opened 2015-08-19
Feedback needed: Quitting from Mac OS X menu bar

jeffreyhaines:
I'd like some feedback about this kind of feature addition. 

This seems a bit hacky. I think the launcher source isn't very supportive of this kind of thing. 

Also I still can't quit or force quit from the launcher from the Mac OS X doc.

Should the launcher be refactored?

### DIFF
--- a/apps/launcher/maindialog.cpp
+++ b/apps/launcher/maindialog.cpp
@@ -5,6 +5,7 @@
 #include <QLabel>
 #include <QDate>
 #include <QTime>
+#include <QMenuBar>
 #include <QMessageBox>
 #include <QPushButton>
 #include <QFontDatabase>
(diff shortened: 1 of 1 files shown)


scrawl:
This will look weird and waste space on platforms that use an in-window menu bar.

scrawl:
My inclination is to reject this. We already have two ways of quitting the launcher (Close button, window X button). We don't need to add more that are platform dependent. Adding a whole menu for the purpose of one button is overkill.

corristo:
What's wrong with exiting with Cmd-Q or Launcher -> Quit? I don't see any point for additional menus here.

jeffreyhaines:
@Scrawl Its not overkill its just supporting the excellent features set of the OS. As a user, I always use command-q to quit my programs.

@corristo I was going to work on it O.o??

corristo:
@jeffreyhaines sure, that would be cool. Here's a pointer:
https://www.thomaskeller.biz/blog/2010/05/02/quitting-a-qt-application-from-the-mac-os-x-dock/

Basically you'll have to subclass QApplication and override one of its methods.

PR #736, opened 2015-08-20
Correctly handle disjoint pathgrid (Fixes #2871)

dteviot:
Bugfix:
When
1. Cell has multiple subgrids  (i.e. path grid is disjoint)
2. Distance between destination and pathgrid point 0 is less than distance to points of subgrid closest to start point
Then getClosestReachablePoint() returns pathgrid point 0 as the end point.
This is invalid, this end point cannot be reached from the start point.

### DIFF
--- a/apps/openmw/mwmechanics/pathfinding.cpp
+++ b/apps/openmw/mwmechanics/pathfinding.cpp
@@ -53,7 +53,8 @@ namespace
     {
         assert(grid && !grid->mPoints.empty());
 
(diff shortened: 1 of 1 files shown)


cc9cii:
I left some comments in your repo.  Your earlier commit broke the way code was intended to work which requires further changes such as this one.  If you're going to maintain this section of the codebase you should go ahead and re-write it the way you prefer.  As it currently is, there are mixed paradigms which does not sit well (e.g. I wrote it for the caller to ensure validity before & after, but you seem to prefer the called function to do that with asserts, etc)

PR #737, opened 2015-08-20
[MAC] Launcher now quits with doc and cmd+q (Bug #2862)

jeffreyhaines:


jeffreyhaines:
Whoops, how do I prevent that merge from showing up?

scrawl:
That segfaults for me.

<pre>
Program received signal SIGSEGV, Segmentation fault.
__GI___pthread_mutex_lock (mutex=0x7fffffffd2f0) at ../nptl/pthread_mutex_lock.c:66
66	../nptl/pthread_mutex_lock.c: No such file or directory.
(gdb) bt
#0  __GI___pthread_mutex_lock (mutex=0x7fffffffd2f0) at ../nptl/pthread_mutex_lock.c:66
#1  0x00007ffff48070da in XrmQGetResource () from /usr/lib/x86_64-linux-gnu/libX11.so.6
#2  0x00007ffff47e4a42 in XGetDefault () from /usr/lib/x86_64-linux-gnu/libX11.so.6
#3  0x00007ffff43ab0d0 in _XcursorGetDisplayInfo () from /usr/lib/x86_64-linux-gnu/libXcursor.so.1
#4  0x00007ffff43ab1a9 in XcursorGetDefaultSize () from /usr/lib/x86_64-linux-gnu/libXcursor.so.1
#5  0x00007ffff43ad435 in XcursorLibraryLoadCursor () from /usr/lib/x86_64-linux-gnu/libXcursor.so.1
#6  0x00007ffff706ed5c in ?? () from /usr/lib/x86_64-linux-gnu/libQtGui.so.4
#7  0x00007ffff706f485 in QCursor::handle() const () from /usr/lib/x86_64-linux-gnu/libQtGui.so.4
#8  0x00007ffff707b364 in ?? () from /usr/lib/x86_64-linux-gnu/libQtGui.so.4
#9  0x00007ffff7080398 in QWidgetPrivate::create_sys(unsigned long, bool, bool) () from /usr/lib/x86_64-linux-gnu/libQtGui.so.4
#10 0x00007ffff703231d in QWidget::create(unsigned long, bool, bool) () from /usr/lib/x86_64-linux-gnu/libQtGui.so.4
#11 0x00007ffff703a86a in QWidget::setVisible(bool) () from /usr/lib/x86_64-linux-gnu/libQtGui.so.4
#12 0x0000000000421270 in show (this=0x7fffffffda60) at /usr/include/qt4/QtGui/qwidget.h:497
#13 main (argc=1, argv=<optimized out>) at /home/scrawl/Dev/openmw/apps/launcher/main.cpp:55
</pre>

The reason for the signal(..) call was to prevent SDL from eating up interrupt (i.e. Ctrl+C events), see bug [2328](https://bugs.openmw.org/issues/2328) .  Removing the signal call will likely make that issue reappear.

jeffreyhaines:
@scrawl let me know if that fixes things for you

scrawl:
That doesn't build.

<pre>
/home/scrawl/Dev/openmw/apps/launcher/graphicspage.cpp:178:27: error: ‘signal’ was not declared in this scope
     signal(SIGINT, SIG_DFL);// We don't want to use the SDL event loop in the launcher,
</pre>

I don't think adding back signal() will fix the above segfault. My guess is QT and SDL are interfering in some way so we need to initialize SDL before QT is initialized. I vaguely recall that kind of issue happening in the past.

jeffreyhaines:
Try it now

jeffreyhaines:
@scrawl how are you running the launcher in terminal? I can't get it to have a console run

scrawl:
Still doesn't build. I think you added the include to the wrong file. Is this building OK on your end or are you just pushing untested stuff?
<pre>
/home/scrawl/Dev/openmw/apps/launcher/graphicspage.cpp:178:27: error: ‘signal’ was not declared in this scope
     signal(SIGINT, SIG_DFL);// We don't want to use the SDL event loop in the launcher,
                           ^
</pre>

> how are you running the launcher in terminal? I can't get it to have a console run

You open a terminal, navigate to the directory containing the launcher executable, then run it with ./openmw-launcher.

jeffreyhaines:
Oh shucks haha - sorry about that.. ran make clean

jeffreyhaines:
Ok - everything builds correctly. I can't get ./openmw-launcher to open the launcher.. Must be a linux thing? Is that the OS you're using?

jeffreyhaines:
Correction, explore the app package jeff...

scrawl:
Yes, that builds. The crash from my 2nd comment is still present however.

If I'm reading the situation correctly, we have a dilemma:
* Initializing SDL after QT produces a crash on Linux
* Initializing QT after SDL breaks dock-exit on OS X (is that correct?)

How about this: we boot up SDL once to retrieve list of resolutions, then exit SDL, then start the QApplication. Would that work for both cases?

>  I can't get ./openmw-launcher to open the launcher.. Must be a linux thing? 

From what I know OS X should work the same way since it is also based on UNIX, but then again I've never really used OS X. Can't help you with that.

jeffreyhaines:
>How about this: we boot up SDL once to retrieve list of resolutions, then exit SDL, then start the >QApplication. Would that work for both cases?

That would not work, because when we switch screens we need to get the resolutions from them. Everything for graphics setting should be located in the graphics page. If we return to putting SDL in the beginning of main, we should just toss is out and find another solution, because it's not good enough.

About the crash. I am not seeing it. I've ran ./openmw-lauch and control-c. Can you please make clean or post some more information about this crash you are seeing.

scrawl:
> That would not work, because when we switch screens we need to get the resolutions from them. 

We could get the resolutions for all screens at startup.

> About the crash. I am not seeing it. I've ran ./openmw-lauch and control-c.

Likely a linux-specific crash then.

> Can you please make clean or post some more information about this crash you are seeing.

I've posted the backtrace in comment 1.

jeffreyhaines:
What version of linux are you using?

scrawl:
Ubuntu 14.04.

jeffreyhaines:
What graphics card are you using and what driver?

jeffreyhaines:
Are you using the Genome Terminal? This is looking like a linux bug.

corristo:
>I can't get ./openmw-launcher to open the launcher.. Must be a linux thing?

You can use `OpenMW.app/Contents/MacOS/openmw-launcher` to launch binary directly and read all output. To bypass launcher use ``OpenMW.app/Contents/MacOS/openmw`.

jeffreyhaines:
@corristo Thanks! I did figure it out if you read the above :)

corristo:
@jeffreyhaines yeah, sorry, I'm kinda slow today.

jeffreyhaines:
@scrawl Let me know what that does for you

scrawl:
Same crash.

> What graphics card are you using and what driver?

That is hardly relevant here. QT does not use the graphics card. You should see the crash on any Linux system using QT4 and X11.

MiroslavR:
I also get exactly the same crash scrawl is experiencing (on Linux).

jeffreyhaines:
@scrawl I'm taking a look at the OSG 3.2 documentation. It looks like we might be able to drop SDL in favor of osg to query the screen settings. I'm not 100% sure yet. 

Quick question about the OSG dependency: Which one are we supporting, which one are you using? I'm using 3.2.2. Perhaps this related to plugins @corristo needs to set up and the code you have written (for save game screen shots that is)?

Stay tuned!

jeffreyhaines:
@scrawl What are your ideas for multi monitor support? Is that a post 1.0 feature or is that something that would be easier done pre-1.0?

scrawl:
> It looks like we might be able to drop SDL in favor of osg to query the screen settings.

Yes, we could theoretically do that. But we use SDL to create the window, so it makes sense to also use SDL to query available settings. Imagine SDL and OSG used a slightly different method for retrieval (different refresh rates, bit depths, AA modes, etc), resulting in mismatched settings. Or they're not recognizing the same screens. Not sure if that happens in practice, I'd just prefer to stay consistent in our usage.

What about my suggestion of booting up SDL to retrieve settings, then shut it down again before starting QT?

scrawl:
We have already multi monitor support in terms of specifying a screen to launch on. Still need the ability to launch a window across all screens. That is an easy patch so I would accept it pre-1.0 if someone wrote it.

jeffreyhaines:
>What about my suggestion of booting up SDL to retrieve settings, then shut it down again before >starting QT?

It's not a good approach. Passing a variable from main to a window to then be passed to a page isn't correct. 

I think we should remove SDL from the launcher. I understand your concerns and will investigate.

Good design first :) We should only need to visit the launcher for quick additions, and swapping of components, i.e. the way we get graphics settings is a component. That is why we shouldn't build the retrieval of graphics setting into our design.

jeffreyhaines:
>That is an easy patch so I would accept it pre-1.0 if someone wrote it.

@scrawl I'll look into this. Maybe this is related to what I'm doing in the launcher. Do we even need SDL still, might be a question to ask.

scrawl:
> It looks like we might be able to drop SDL in favor of osg to query the screen settings.

> Yes, we could theoretically do that. But we use SDL to create the window, so it makes sense to also use SDL to query available settings. Imagine SDL and OSG used a slightly different method for retrieval (different refresh rates, bit depths, AA modes, etc), resulting in a different resolution list. Or they're not recognizing the same screens. Not sure if that happens in practice, I'd just prefer to stay consistent in our usage.

Additionally, SDL is more fully featured with regards to available display servers than OSG is (Wayland and Mir support on Linux). Relying on OSG for the windowing system tasks would restrict the ability to run OpenMW on these display systems.

jeffreyhaines:
>Additionally, SDL is more fully featured with regards to available display servers than OSG is (Wayland >and Mir support on Linux). Relying on OSG for the windowing system tasks would restrict the ability to ?>run OpenMW on these display systems.

I'm not sure thats entirely accurate. Let me do some research and get back to this.

jeffreyhaines:
@MiroslavR and @scrawl : let me know is the addition of  SDL_INIT_NOPARACHUTE worked for you. Did you build and test my latest commit?

MiroslavR:
> let me know is the addition of SDL_INIT_NOPARACHUTE worked for you

No, it hasn't fixed the crash.

> Did you build and test my latest commit?

Yes.

jeffreyhaines:
Hey @scrawl or @zinnschlag can we close this (not merge) and I'll start over. Ghost was me before I deleted my old account.

nikolaykasyanov:
@jeffreyhaines here you go!

PR #738, opened 2015-08-20
removed incorrect optimization.

dteviot:
Now it fixes #2871

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -806,7 +806,7 @@ namespace MWMechanics
 
     void AiWander::SetCurrentNodeToClosestAllowedNode(osg::Vec3f npcPos)
     {
-        float distanceToClosestNode = FLT_MAX;
+        float distanceToClosestNode = std::numeric_limits<float>::max();
         unsigned int index = 0;
         for (unsigned int counterThree = 0; counterThree < mAllowedNodes.size(); counterThree++)
         {
--- a/apps/openmw/mwmechanics/autocalcspell.cpp
+++ b/apps/openmw/mwmechanics/autocalcspell.cpp
@@ -181,7 +181,7 @@ namespace MWMechanics
 
     void calcWeakestSchool (const ESM::Spell* spell, const int* actorSkills, int& effectiveSchool, float& skillTerm)
     {
-        float minChance = FLT_MAX;
+        float minChance = std::numeric_limits<float>::max();
 
         const ESM::EffectList& effects = spell->mEffects;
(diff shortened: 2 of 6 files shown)


zinnschlag:
Could we please use the modern form (numeric_limits) instead of the macro crap (FLT_MAX) ?

scrawl:
I don't feel too strongly about that. The macro is there anyway, it isn't in our control. Not using it doesn't make the macro any less harmful. Whatever works.

scrawl:
Travis build failing.

cc9cii:
Maybe I'm missing something - what does this PR fix?

dteviot:
It's correctly fixing #2871.
Without this change, when getClosestReachablePoint() initializes, it sets closestReachableIndex to node 0 of the path grid. And closestDistanceReachable to the distance to this node.
Note that node 0 is not guaranteed to be a member of the same subgrid as the start node (when the path grid is composed of multiple disconnected sub grids.)
So, unless there's a node in the subgrid that start node belongs to that is closer to the endpoint than node 0, node 0 will be returned as the end point.  Which violates the post-condition that start and end must be members of the same sub grid.  I.e. There is a path between the nodes.
Being members of the same sub grid is required for aStarSearch to find a path between them.

cc9cii:
Right.  So the title is a little misleading then.  i.e. not really anything to do with optimisation

This change is needed because the change https://github.com/OpenMW/openmw/commit/5aeabe22f0b58b2b69ffb9e67b558e54ff236caa removed the checking of the return value which would be false if closestReachableIndex != closestIndex (would have caught the edge cases you describe).

If the code is easier to read and therefore easier to maintain then I have no issue with this change.

dteviot:
@cc9cii 

> Right. So the title is a little misleading then. i.e. not really anything to do with optimisation

That depends on how you look at it.
The code was trying to initialize variables with the values for node[0].  
I believe this was done to avoid having to run the loop for the zeroth node (as the loop counter starts at the node[1].)  i.e. This was done as an optimization.

I would add that there were in fact 2 bugs in getClosestReachablePoint().  The incorrect loop initialization, and that it did not track both the distance to the closest node, and distance to the closest node that was connected to the start node.

> This change is needed because the change 5aeabe2 removed the checking of the return value which would be false if closestReachableIndex != closestIndex (would have caught the edge cases you describe).

No.  This change is needed because without it, getClosestReachablePoint() can return an end point that is not connected to the start node.
The test closestReachableIndex != closestIndex is used to determine if the final destination passed into buildPath() can be reached from the endNode returned by getClosestReachablePoint().

What caused the crash in #2871 was my adding an assertion for the post-condition that aStarSearch() returns an non-empty list.

PR #739, opened 2015-08-20
Enable by default time played tracker (Feature #2528)

jeffreyhaines:


scrawl:
I disagree with this as explained in the last PR. An unlocalised string in an otherwise fully localised dialog looks out of place.

Even ignoring the localisation issue, time played is probably not something that everyone wants to see as it kinda takes you out of the game by breaking the fourth wall. I don't know of any games that show this by default.

Regarding inexperienced users not knowing of this hidden option, just document it on the wiki [Config file tweaks](https://wiki.openmw.org/index.php?title=Config_file_tweaks) and we should be set.

jeffreyhaines:
I disagree. Many companies use English text thats not always localized in the product. 

@zinnschlag you didn't have a problem with this right?

I agree not everyone wants to see time played, it's still a cool thing to have turned on. 

Also Skyrim has this enabled by default.

LavenderMoon:
As does Fallout: New Vegas.

zinnschlag:
And Gothic.

We are displaying the time and date of a saved game anyway, so displaying the time played doesn't change anything.

The lack of localisation is of course a problem. But as mentioned before, we already have that problem anyway (FOV setting ?), so adding one more isn't making things much worse. We can fix the problem after 1.0.
I still would prefer if people would hold back on adding features that require new localised strings.

scrawl:
> We are displaying the time and date of a saved game anyway

That is displayed via strftime(), which handles localisation correctly. There is no locale time format for "time played" type durations, however (another inconsistency).

> The lack of localisation is of course a problem. But as mentioned before, we already have that problem anyway (FOV setting ?), so adding one more isn't making things much worse. 

Yes, there are some unlocalised strings in options-menu. These were kinda unavoidable. The vanilla options menu is way too limiting to follow 1:1. However, the options is something that you visit once or twice and then forget about, so I don't consider that a big deal. The save dialog on the other hand you will be constantly in contact with during gameplay, so a break in consistency is a big deal.

This is coming from me who used to use a Linux system set to german. I found incomplete translations so off-putting and irritating that I switched my whole system to english at one point. Haven't looked back since.

jeffreyhaines:
I don't know what the choice should be. I do know this is a pretty cool feature. I think it should be enabled by default. If it gets annoying to someone they can turn it off. I usually quick save so I don't see the save menu too much.

We could create a poll and vote for default on or off in the forums.

zinnschlag:
> This is coming from me who used to use a Linux system set to german. I found incomplete translations so off-putting and irritating that I switched my whole system to english at one point. Haven't looked back since.

Same here.


Anyway, I don't see a substantial issue. Its only a default setting for a very tiny piece of text that the user only sees occasionally and that we will address properly in OpenMW 1.1. Then again I wished people would wait with implementing features that we can't support cleanly yet. I guess I don't have a clear opinion on this issue.

jeffreyhaines:
If this still works why don't we merge it in. My old account btw...

Allofich:
The setting still exists so it should work, but this PR just changes the default setting for an option, and I don't see why we should assume people would want it activated. There are other options in the settings file that we could change the default for, too, but these are personal preferences. I think we could have people disagreeing and maybe feeling frustrated if we start advocating certain options as the ones to be on by default.

I do think, though, that the available settings are not very obvious. I guess they're on the wiki but I wasn't really aware of them until recently. Maybe they could be made available through the OpenMW launcher?

jeffreyhaines:
Yeah that definitely could be in the launcher at some point. I do have ideas/intentions to create a new editor which would be pretty cool :D 

At this point I want to clean the bug tracker up before I start anything. This pull-request is dated and should be closed as such.

Allofich:
Yes, I agree with you wanting to clean up old dated bug issues and pull requests, so we are in agreement there :)

jeffreyhaines:
Mhm https://github.com/OpenMW/openmw/pull/1063 ;)

scrawl:
The best place for discussion about setting defaults would be on the forum. Though I think we've already wasted enough time discussing this particular one. I'll close this PR for now since it's not really a code change; if you still want to see this changed open a discussion on the forum first.

jeffreyhaines:
No closing is what I requested and expected. Thanks

PR #745, opened 2015-08-24
Pulled duplicated functions into common base class

dteviot:
Factoring out common code from Npc and Creature into base class.
This is just starting with the functions that are the same.
I thought I'd start with them and see what Scrawl & Zini think before I put in any more work.

There's obviously more that can be done.  e.g. getSpeed() , onHit(). etc.

### DIFF
--- a/apps/openmw/mwclass/mobile.cpp
+++ b/apps/openmw/mwclass/mobile.cpp
@@ -1,0 +1,87 @@
+#include "mobile.hpp"
+
+#include <components/esm/loadmgef.hpp>
+
+#include "../mwbase/environment.hpp"
+#include "../mwbase/world.hpp"
+#include "../mwbase/mechanicsmanager.hpp"
+#include "../mwbase/soundmanager.hpp"
+
+#include "../mwmechanics/creaturestats.hpp"
+#include "../mwmechanics/movement.hpp"
+#include "../mwmechanics/magiceffects.hpp"
+
+#include "../mwphysics/physicssystem.hpp"
+
+#include "../mwworld/inventorystore.hpp"
+
+namespace MWClass
+{
+    Mobile::Mobile() {}
+
+    Mobile::~Mobile() {}
+
+    void Mobile::adjustPosition(const MWWorld::Ptr& ptr, bool force) const
+    {
+        MWBase::Environment::get().getWorld()->adjustPosition(ptr, force);
+    }
+
+    void Mobile::insertObject(const MWWorld::Ptr& ptr, const std::string& model, MWPhysics::PhysicsSystem& physics) const
+    {
+        if (!model.empty())
+        {
+            physics.addActor(ptr, model);
+            if (getCreatureStats(ptr).isDead())
+                MWBase::Environment::get().getWorld()->enableActorCollision(ptr, false);
+        }
+        MWBase::Environment::get().getMechanicsManager()->add(ptr);
+    }
+
+    void Mobile::block(const MWWorld::Ptr &ptr) const
+    {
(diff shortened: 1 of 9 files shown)


apps/openmw/mwclass/mobile.hpp
```diff
@@ -0,0 +1,47 @@
...
+{
+    /// \brief Class holding functionality common to Creature and NPC
+    class Mobile : public MWWorld::Class
+    {
+    protected:
+
+        Mobile();
+
+    public:
+        ~Mobile();
```
jeaye: For future-proofing, this dtor should be explicitly marked `virtual`.

zinnschlag: The destructor is virtual, because the Class destructor is virtual. But yes, for the sake of clarity and consistency the destructor should be marked as virtual. I guess that's what you meant?

jeaye: On Mon, Aug 24, 2015 at 06:27:32AM -0700, Marc Zinnschlag wrote:
>    The destructor is virtual, because the Class destructor is virtual. But
>    yes, for the sake of clarity and consistency the destructor should be
>    marked as virtual. I guess that's what you meant?

Of course it's currently virtual, due to the base class' dtor being virtual. By "future-proofing" I mean marking it virtual so that, if the base class is ever changed/removed, we don't end up with broken dtors.

Trying to save our future selves.

zinnschlag:
Basically okay with this change. One issue: What you call Mobile we call Actor everywhere else in the code. Let's be consistent here.

If you can factor out more that's fine; as long as the code in Mobile does not need to distinguish between actor types.

apps/openmw/mwclass/mobile.cpp
```diff
@@ -0,0 +1,87 @@
...
+            break;
+        case ESM::Skill::MediumArmor:
+            sndMgr->playSound3D(ptr, "Medium Armor Hit", 1.0f, 1.0f);
+            break;
+        case ESM::Skill::HeavyArmor:
+            sndMgr->playSound3D(ptr, "Heavy Armor Hit", 1.0f, 1.0f);
+            break;
+        default:
+            return;
+        }
```
zinnschlag: Can we keep the standard indenting style please? I know we aren't super consistent with switch-case, but that doesn't mean we need to make it worse when refactoring.

apps/openmw/mwclass/mobile.hpp
zinnschlag: Not going to happen since the entire point of the Class hierarchy is to access the sub-classes via the Class interface. But we are getting into nitpicking area here.

jeaye: >    Not going to happen since the entire point of the Class hierarchy is to
>    access the sub-classes via the Class interface. But we are getting into
>    nitpicking area here.

Yeah, whitespace corrections are certainly more important than future-proof dtors. :P

dteviot:
>  But yes, for the sake of clarity and consistency the destructor should be marked as virtual.

Done.

> What you call Mobile we call Actor everywhere else in the code. Let's be consistent here.

My bad.  I thought I'd seen actor used elsewhere to also refer to doors, containers, and sound effects.  I think I was wrong.  Class renamed.

> as long as the code in Mobile does not need to distinguish between actor types

That's my plan. (Code in Actor will not distinguish between Creature and Npc)

> Can we keep the standard indenting style please?

Restored original formatting.

zinnschlag:
> My bad. I thought I'd seen actor used elsewhere to also refer to doors, containers, and sound effects.

This is OpenMW, not Unreal Engine.

Merging ...

PR #747, opened 2015-08-27
Refactor weather transitions to act more like MW

slothlife:
Fixed several issues:
* Waiting/jail time/training all now properly skip remaining transitions
* ChangeWeather no longer permanently sets the region's weather
* ChangeWeather being called during a transition now correctly queues up
another transition
* Corrected transition delta and factor calculations
* ModRegion settings are now saved

These changes unfortunately break the current saved game format. I apologize in advance for the ugly diff, it may be easier to do a side by side comparison. Feedback welcome because this is the first truly large commit to the project I've made.

### DIFF
--- a/apps/openmw/mwworld/weather.cpp
+++ b/apps/openmw/mwworld/weather.cpp
@@ -736,7 +736,7 @@ bool WeatherManager::readRecord(ESM::ESMReader& reader, uint32_t type)
             mWeatherUpdateTime = state.mWeatherUpdateTime;
             mTransitionFactor = state.mTransitionFactor;
             mCurrentWeather = state.mCurrentWeather;
-            mNextWeather = state.mCurrentWeather;
+            mNextWeather = state.mNextWeather;
             mQueuedWeather = state.mQueuedWeather;
 
             mRegions.clear();
@@ -884,6 +884,8 @@ inline void WeatherManager::updateWeatherTransitions(const float elapsedRealSeco
             mCurrentWeather = mNextWeather;
         }
 
+        mNextWeather = invalidWeatherID;
+        mQueuedWeather = invalidWeatherID;
         mFastForward = false;
     }
 }
--- a/apps/openmw/mwworld/weather.cpp
+++ b/apps/openmw/mwworld/weather.cpp
@@ -670,19 +670,12 @@ osg::Vec3f WeatherManager::getStormDirection() const
     return mStormDirection;
 }
 
-void WeatherManager::advanceTime(double hours)
+void WeatherManager::advanceTime(double hours, bool incremental)
 {
-    // This is called when the player sleeps/waits, serves jail time, travels, or trains.
-    // In Morrowind, when any of those events occur, all weather transitions are immediately applied, regardless of
-    // whatever transition time might have been remaining.
+    // In Morrowind, when the player sleeps/waits, serves jail time, travels, or trains, all weather transitions are
+    // immediately applied, regardless of whatever transition time might have been remaining.
     mTimePassed += hours;
-    mFastForward = true;
-}
-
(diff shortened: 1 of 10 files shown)


scrawl:
To handle the save game break, I suggest:
* Increment the ESM::SavedGame::sCurrentFormat
* Detect older formats of weather state and discard them, e.g.
<pre>
if (reader.getFormat() < ESM::SavedGame::sCurrentFormat)
   reader.skipRecord();
else
{
   ESM::WeatherState state;
   reader.load(esm);
   mWeatherManager->readRecord(state);
}
</pre>

slothlife:
Thanks, I was worried about breaking older saves. I tested with an older save and it appears to discard them correctly.

slothlife:
Added another commit to make loading of older weather state records more robust. This way, if someone updates the save game format (version 3+), it won't automatically break importing of older compatible weather records (version 2).

apps/openmw/mwworld/weather.cpp
```diff
@@ -101,6 +104,95 @@ float Weather::cloudBlendFactor(float transitionRatio) const
...
+
+    return mWeather;
+}
+
+void RegionWeather::chooseNewWeather()
+{
+    // All probabilities must add to 100 (responsibility of the user).
+    // If chances A and B has values 30 and 70 then by generating 100 numbers 1..100, 30% will be lesser or equal 30
+    // and 70% will be greater than 30 (in theory).
+    int chance = Misc::Rng::rollDice(100) + 1; // 1..100
```
Mingun: Why not just sum all `mChances` and then use this sum instead of `100`? This sum may be cached for perfomance.

slothlife: I didn't modify that code much, it was lifted from the original algorithm as is.

Technically, the user is required to provide values such that the sum is 100, otherwise it has erratic behavior according to various sources (MWSFD, UESP). I didn't investigate the original's erratic behavior myself, so I can't say if your suggestion is correct.

slothlife:
Working remotely, it seems my SSH client didn't pick up the file modification I'd made and then committed the previous broken change.

scrawl:
Not so sure if it's correct to store original weather chances in the save file. That means if I create a mod that changes region weather chances, I'd need to start a new game for the changes to take effect, right?

From what I could tell the MW .ess saves don't store this info.

slothlife:
I found that it does seem to save them. As an experiment, I did ModRegion "Bitter Coast Region" 0 0 0 0 0 0 0 0 100 0 (which starts up snow fall). I then traveled to a different region ("Ascadian Isles Region"), saved the game, and exited. I then started the game and loaded my save game, waited 24 hours, and then traveled back and found that it was still snowing. Waiting 24 hours there had no effect as well.

scrawl:
Right, see my edit on last comment. I believe it's stored in the REGN records. Now that the weather manager supports it, we should be able to ESS-import the region weather chances as well.

slothlife:
Is that something you'd want in this PR, or in a future one? I'm planning more work on the weather manager in the future anyway.

scrawl:
> Is that something you'd want in this PR, or in a future one?

I don't have a clear preference. Whenever you feel like it.

Please rebase the build fix commit. I prefer to have all commits buildable so we don't need to skip anything when running a bisect.

Other than that, PR looks good to merge.

scrawl:
About World::advanceTime / World::advanceTimeByFrame: maybe we could add a ```bool fastForward = false``` parameter to advanceTime, to get rid of the code duplication?

slothlife:
Sure, I'll have to do that a little later, I'm not at the computer right
now.

slothlife:
Rebased the last commit.

slothlife:
By the way, I have in all of the changes that I planned to do in this PR.

PR #748, opened 2015-08-27
extracted isFlagBitSet()

dteviot:


zinnschlag:
There is little point in having a private non-virtual function in a class derived from Class, since these classes don't have any state. Maybe have it as a free function in an anonymous namespace instead (in the cpp file)?

dteviot:
>  Maybe have it as a free function in an anonymous namespace instead (in the cpp file)?

Done.

PR #753, opened 2015-09-09
Improve thunderstorm support.

slothlife:
Reversed settings for thunderstorms. Added thunder support to all
weather types. Implemented a simple lightning flash effect similar to
MW.

### DIFF
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -1181,20 +1181,6 @@ void SkyManager::setSecundaState(const MoonState& state)
     mSecunda->setState(state);
 }
 
-void SkyManager::setLightningStrength(const float factor)
-{
-    if (!mCreated) return;
-    /*
-    if (factor > 0.f)
-    {
-        mLightning->setDiffuseColour (ColourValue(2*factor, 2*factor, 2*factor));
-        mLightning->setVisible(true);
-    }
-    else
-        mLightning->setVisible(false);
-        */
-}
-
 void SkyManager::setDate(int day, int month)
 {
     mDay = day;
--- a/apps/openmw/mwrender/sky.hpp
+++ b/apps/openmw/mwrender/sky.hpp
@@ -140,8 +140,6 @@ namespace MWRender
         void setMasserState(const MoonState& state);
         void setSecundaState(const MoonState& state);
 
-        void setLightningStrength(const float factor);
-
         void setGlare(const float glare);
         void setGlareEnabled(bool enabled);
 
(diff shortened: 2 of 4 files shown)


scrawl:
@slothlife are you working on anything weather related? I am working on the sun glare right now. Just wanted to give you a heads up so we can avoid conflicts. My changes are mostly limited to the Sun class.

slothlife:
There shouldn't be any conflicts. I was trying to do the same thing but I
am running into lots of problems I was having trouble figuring out. I'll be
out of town for a day or 2 and won't have time to finish looking into it.

It's worth noting that QueryGeometry has a way to retrieve the number of
pixels that passed the depth test, but I was running into strange problems
where I believe the CameraRelativeTransform was leading to strange results
when OcclusionQueryNode was calculating the distance to the eye point.

scrawl:
Yeah, I have found the getNumPixels method. I'm not sure why I didn't find it or consider it usable last time I looked at this, but it works fine. I haven't had any issues with the queries so far. Where did you attach your OcclusionQueryNode?

Here's what I have so far: https://github.com/scrawl/openmw/tree/sky

To test the query results, I used this:

<pre>
        mGeode->addCullCallback(new PrintCallback(mOcclusionQueryVisiblePixels, mOcclusionQueryTotalPixels));

...
	
    struct PrintCallback : public osg::NodeCallback
    {
    public:
        PrintCallback(osg::OcclusionQueryNode* oqn1, osg::OcclusionQueryNode* oqn2)
            : mOqn1(oqn1), mOqn2(oqn2)
        {

        }

        virtual void operator()(osg::Node* node, osg::NodeVisitor* nv)
        {
            osgUtil::CullVisitor* cv = static_cast<osgUtil::CullVisitor*>(nv);

            std::cout << "visible: " << mOqn1->getQueryGeometry()->getNumPixels(cv->getCurrentCamera()) << std::endl;
            std::cout << "total: " << mOqn2->getQueryGeometry()->getNumPixels(cv->getCurrentCamera()) << std::endl;

            traverse(node, nv);
        }

        osg::OcclusionQueryNode* mOqn1;
        osg::OcclusionQueryNode* mOqn2;
    };

</pre>

I think it's going to turn out very nicely, in fact with the OcclusionQueryNode and a callback it's very easy to do multi camera support, something that would have been near impossible with the old design. 

I have also studied the vanilla sky a bit more and found that the "glare" effect you get that fades depending on how directly you are looking at the sun, is just a full screen additive quad, using those ini settings:
<pre>
        /*
        Sun Glare Fader Max=0.5
        Sun Glare Fader Angle Max=30.0
        Sun Glare Fader Color=222,095,039
        */
</pre>

The flash texture just scales from 0 to 1 depending on the current occlusion test, no fading (vanilla used a raytest for occlusions, so there are no partial occlusions - but we are changing that).

I should have it finished by tomorrow or the day after. I want to get the multi camera support right, not because we currently need it, but it feels wrong to pass a specific Camera* to the Sky class.

slothlife:
I had put the OCN between mTransform and mGeode, but I'm sure i just hooked
it up wrong.

It seems you're having much better results. I hadn't gotten far enough to
begin determining what any previous settings did.

PR #754, opened 2015-09-10
Some minor refactoring

dteviot:


zinnschlag:
Not really convinced by the efficiency argument. Sounds like premature optimisation. But okay. Will merge.

PR #756, opened 2015-09-13
my patches

zelurker:
Well as I said on the forum you wouldn't want to merge all these patches, especially the one on magic experience, but you are the one who wanted a pull request, so here is one...

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -1152,9 +1152,9 @@ namespace MWPhysics
 
                 bool waterCollision = false;
                 if (effects.get(ESM::MagicEffect::WaterWalking).getMagnitude()
-                        && cell->getCell()->hasWater()
-                        && !world->isUnderwater(iter->first.getCell(),
-                                               osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
+                        && cell->getCell()->hasWater())
+//                        && !world->isUnderwater(iter->first.getCell(),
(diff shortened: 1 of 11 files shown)


apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -1152,9 +1152,9 @@ namespace MWPhysics
 
                 bool waterCollision = false;
                 if (effects.get(ESM::MagicEffect::WaterWalking).getMagnitude()
-                        && cell->getCell()->hasWater()
-                        && !world->isUnderwater(iter->first.getCell(),
-                                               osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
+                        && cell->getCell()->hasWater())
+//                        && !world->isUnderwater(iter->first.getCell(),
+//                                               osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
```
psi29a: It is better to remove code then to comment it out, if we need it, we can always go back in time with git to retrieve this.

psi29a:
@scrawl can you also eyeball this please :)

scrawl:
Please create a separate branch and pull request for each topic. I have a few comments, but I will wait for separate PRs before posting them.

apps/openmw/mwphysics/physicssystem.cpp
dteviot: The waterCollision variable is never used. So lines 1153 to 1158 can be removed.

dteviot: I'm wrong.  Variable is used.
For some reason, when I did a search for waterCollision, only the two instances were shown.
D'oh.

PR #758, opened 2015-09-14
Add CONTRIBUTING.md

scrawl:
This pull request will also serve as an example implementation of the proposed guidelines:

> * Summary of the changes made

Add a CONTRIBUTING.md file containing guidelines for pull requests.

> * Reasoning / motivation behind the change

The contents of CONTRIBUTING.md will show up in github's pull request UI. We should use it to display some guidelines on what we are looking for in a good PR, so that we will hopefully see an overall improvement in the quality of PRs.

> * What testing you have carried out to verify the change

Viewed the file on github to make sure the markdown syntax is correct, cross-checked with the rules on our [wiki](https://wiki.openmw.org/index.php?title=Developer_Reference).

> * If applicable, a link back to the bug report or forum discussion that prompted the change

None, the idea just crossed my mind.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,16 @@
+Description
+===========
+
+Your pull request description should include:
+
+* Summary of the changes made
+* Reasoning / motivation behind the change
+* What testing you have carried out to verify the change
+* If applicable, a link back to the bug report or forum discussion that prompted the change
+
+Other notes
+===========
+
+* Separate your work into multiple pull requests whenever possible. As a rule of thumb, each feature and each bugfix should go into a separate PR, unless they are closely related or dependent upon each other. Small pull requests are easier to review, and are less likely to require further changes before we can merge them. A "mega" pull request with lots of unrelated commits in it is likely to get held up in review for a long time.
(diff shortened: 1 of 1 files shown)


zinnschlag:
Interesting. I didn't know about the CONTRIBUTING.md feature. Generally okay with it. But I think we should move the "If applicable" part out of the link paragraph and into the "Your pull request description should include" line.

A few examples:
- In case of fixing a confirmed bug (with a link to the tracker given), we clearly do not need a statement about reasoning / motivation.
- Source formatting changes, improved comments and some other modifications that do not actually modify the non-comment part of the code do not require testing.
- If the PR is just about fixing a bug or even implementing a narrowly defined feature (again with a link to the tracker given), there is no need for a summary, unless there is something specific to tell about the fix that is not already given in the issue description.

And maybe move the tracker link item to the top of the list? IMHO its the most important one.

scrawl:
> But I think we should move the "If applicable" part out of the link paragraph and into the "Your pull request description should include" line.
> And maybe move the tracker link item to the top of the list? IMHO its the most important one.

I figured to not overstate the "if applicable" part, since I'd much rather have too much information than too little. Agreed with your examples though. Anyway, done and done.

PR #759, opened 2015-09-15
Mannequins

zelurker:
2 kind of mannequins handled, the classic one with a script attached containing skipanim, tested here in yorick's tower.
And a less classic one with no script, tested here in "a good place to stay"

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -697,18 +697,9 @@ CharacterController::CharacterController(const MWWorld::Ptr &ptr, MWRender::Anim
             mAnimation->showCarriedLeft(updateCarriedLeftVisible(mWeaponType));
         }
 
-        if(!cls.getCreatureStats(mPtr).isDead() ||
-		(cls.isNpc() &&
-		 mPtr.get<ESM::NPC>()->mBase->mFlags & 8)) // persistent !
-	    /* From http://wiki.theassimilationlab.com/mmw/Morrowind_Script_Library#Mannequins
-	     * mannequins are npc with the persist bit, they don't play the
-	     * death animation and just stay here */
-            mIdleState = CharState_Idle;
-        else
-        {
-            int deathindex = mPtr.getClass().getCreatureStats(mPtr).getDeathAnimation();
-            playDeath(1.0f, CharacterState(CharState_Death1 + deathindex));
-        }
+	// Don't test if the character is dead here to give a chance to an
+	// eventual attached script to stop the death animation for mannequins
+	mIdleState = CharState_Idle;
     }
     else
     {
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -697,7 +697,12 @@ CharacterController::CharacterController(const MWWorld::Ptr &ptr, MWRender::Anim
             mAnimation->showCarriedLeft(updateCarriedLeftVisible(mWeaponType));
         }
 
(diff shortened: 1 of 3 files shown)


kcat:
FWIW, it would be better to title patches and pull requests based on what the it does to the code, rather than specific usecase behavior. "Mannequins" doesn't say anything about what the patchset actually does to the engine.

zinnschlag:
I am actually okay with it. Its enough it find it and we do have a description in the pull request.

kcat:
Sure, I don't mean it to be a blocker for merging if the changes are good, I just mean for future reference it might be preferable to give more meaningful commit messages.

zelurker:
Yeah well this was also discussed quite a lot on the forum, it's right that
the commit messages assume everybody knows it's about mannequins lying on
their backs, maybe I should have reminded that... better luck next time... !


2015-09-15 8:09 GMT+02:00 kcat <notifications@github.com>:

apps/openmw/mwmechanics/character.cpp
```diff
@@ -697,7 +697,12 @@ CharacterController::CharacterController(const MWWorld::Ptr &ptr, MWRender::Anim
             mAnimation->showCarriedLeft(updateCarriedLeftVisible(mWeaponType));
         }
 
-        if(!cls.getCreatureStats(mPtr).isDead())
+        if(!cls.getCreatureStats(mPtr).isDead() ||
+		(cls.isNpc() &&
+		 mPtr.get<ESM::NPC>()->mBase->mFlags & 8)) // persistent !
```
scrawl: If you have found a new flag, the flag should be added to the enum in ESM::Npc, to esmtool and OpenCS. I kinda doubt this is correct, though. We already have ```bool mPersistent``` in ESM::Npc set based on the record flags. And what does persist have to do with mannequins?

> From http://wiki.theassimilationlab.com/mmw/Morrowind_Script_Library#Mannequins
> mannequins are npc with the persist bit, they don't play the
> death animation and just stay here

I would guess they don't play death animations because of the skipAnim call, the Persist bit should have nothing to do with it.

zelurker: Yeah I didn't interpret the description there correctly, the only thing
persist does is that it just prevents them from disappearing after a while.
And this means I was too fast and too optimistic about that, this bit is
probably unrelated.
Which means that I will have to look again how these mannequins behave (the
1st kind, the one from "a good place to stay").
Ok, I'll post again later about that, but I won't do it just now...

2015-09-15 14:26 GMT+02:00 scrawl <notifications@github.com>:

zelurker: Done, so finally no more bit test. The problem was that the character was
killed in CharacterController::CharacterController even before any attached
script has a chance to stop the animation.
So I just removed completely the test isDead from here, it will be killed
later anyway.
Checked on both mods, everything works. And went to a cell with a dead
body, it stayed dead.
This time I think everything is ok (finally !)


2015-09-15 18:59 GMT+02:00 Emmanuel Anne <emmanuel.anne@gmail.com>:

scrawl:
Here's my attempt, could you test if it works for the mannequins?

https://github.com/scrawl/openmw/commit/46c06ec06473f7912b5b189aa28c1e0529d2fe26

scrawl:
Relevant bug - https://bugs.openmw.org/issues/2513

zelurker:
perfect, the deads remain dead and all the mannequins I could test work,
you can use this instead of this branch, it's shorter and more efficient !


2015-09-17 4:35 GMT+02:00 scrawl <notifications@github.com>:

scrawl:
#767

PR #760, opened 2015-09-15
fix setangle

zelurker:
Apparently setangle uses a different rotation order as the default one used in game, discussed extensively in the forum.
Tested mainly in "book rotate"

### DIFF
--- a/apps/openmw/mwworld/scene.cpp
+++ b/apps/openmw/mwworld/scene.cpp
@@ -67,7 +67,7 @@ namespace
             osg::Quat rot(z, osg::Vec3(0,0,-1));
             if (!ptr.getClass().isActor()) {
 		if (prio)
-		    rot = osg::Quat(y, osg::Vec3(0,-1,0)) * osg::Quat(x, osg::Vec3(-1,0,0)) * rot;
+		    rot = osg::Quat(y, osg::Vec3(-1,0,0)) * osg::Quat(x, osg::Vec3(0,-1,0)) * rot;
 		else
 		    rot = rot * osg::Quat(y, osg::Vec3(0,-1,0)) * osg::Quat(x, osg::Vec3(-1,0,0));
 	    }
--- a/apps/openmw/mwworld/scene.cpp
+++ b/apps/openmw/mwworld/scene.cpp
@@ -67,7 +67,7 @@ namespace
             osg::Quat rot(z, osg::Vec3(0,0,-1));
             if (!ptr.getClass().isActor()) {
 		if (prio)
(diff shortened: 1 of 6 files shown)


scrawl:
That is the same conclusion that I came to on bug https://bugs.openmw.org/issues/2062. Rotate script function is also affected. If implemented correctly, this PR should fix bug 2062 as well.

scrawl:
I need to take a long look at this problem. Our rotation system is overcomplicated. I believe the notion of local and world rotations was just a bandaid for the rotation order problem, that we shouldn't need anymore now.

zelurker:
lol, did I mention I hadn't tested the y axis of rotation ? ;-)
That's because I hadn't found any script using it, and I was too lazy to
test it myself (it's not that easy, it requires launching also vanilla
morrowind to compare).
Anyway, of course I had picked the wrong order. So the order is correct
now, it fixes this bug too by the way.
Use coc "old mournhold: forgotten sewer" for testing and then rotate the
crank along the tube, this time it rotates right (with the commit I just
made !).

Sorry about that, I didn't know !


2015-09-15 14:36 GMT+02:00 scrawl <notifications@github.com>:

apps/openmw/mwworld/scene.cpp
```diff
@@ -43,22 +43,34 @@ namespace
     }
 
     void updateObjectLocalRotation (const MWWorld::Ptr& ptr, MWPhysics::PhysicsSystem& physics,
-                                    MWRender::RenderingManager& rendering)
+                                    MWRender::RenderingManager& rendering,
+				    int prio = 0)
```
Mingun: `prio` has `bool` semantic, so, I think, it  must be of type `bool`. Also, name `prio` is cryptic. I think its name must reflect multiplication order.

zelurker: it's int because I wanted to leave room for an eventual 3rd order even if
it's extremely unlikely.
We can have "rotation_order" instead if you prefer, prio was for axis
priority in the rotations of course.

2015-09-15 21:20 GMT+02:00 Mingun <notifications@github.com>:

scrawl:
> Our rotation system is overcomplicated. I believe the notion of local and world rotations was just a bandaid for the rotation order problem, that we shouldn't need anymore now.

Should be fully addressed by PR #802. 

I didn't notice your "setangle/getangle do not wrap angles but other rotations do" commit until now, but I independently found this problem, it was also fixed. Please check if that works for you.

zelurker:
it seems ok for me, tested with book rotate, the fishing boat in fishing
academy, and the submarine there too.
These were the 3 main problems they seem to work, and it's good to have
removed this local rotation.
You can finally close this !

2015-11-12 2:00 GMT+01:00 scrawl <notifications@github.com>:

psi29a:
Error: Tab characters found!

PR #761, opened 2015-09-15
Walkonwater

zelurker:
Just to push the actor out of water when he is casting this spell, as in the vanilla game. Tested ingame, in different places filled with water !

### DIFF
--- a/apps/openmw/mwphysics/actor.cpp
+++ b/apps/openmw/mwphysics/actor.cpp
@@ -150,6 +150,16 @@ void Actor::setCanWaterWalk(bool waterWalk)
     {
         mCanWaterWalk = waterWalk;
         updateCollisionMask();
+	if (waterWalk) {
+	    ESM::Position pos = mPtr.getRefData().getPosition();
+	    if (pos.pos[2] < -100) {
+		/* I didn't look for the exact limit but watercolision will
+		 * push the player out of water only if there is not too much
+		 * water, for -100 and beyond position must be adjusted ! */
+		pos.pos[2] = -50;
+		mPtr.getRefData().setPosition(pos);
+	    }
+	}
     }
 }
 
--- a/apps/openmw/mwphysics/actor.cpp
+++ b/apps/openmw/mwphysics/actor.cpp
@@ -150,16 +150,6 @@ void Actor::setCanWaterWalk(bool waterWalk)
     {
         mCanWaterWalk = waterWalk;
         updateCollisionMask();
-	if (waterWalk) {
-	    ESM::Position pos = mPtr.getRefData().getPosition();
-	    if (pos.pos[2] < -100) {
-		/* I didn't look for the exact limit but watercolision will
-		 * push the player out of water only if there is not too much
-		 * water, for -100 and beyond position must be adjusted ! */
-		pos.pos[2] = -50;
-		mPtr.getRefData().setPosition(pos);
-	    }
-	}
     }
 }
 
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -396,8 +396,8 @@ namespace MWPhysics
             }
 
             bool isOnGround = false;
-            if (!(inertia.z() > 0.f) && !(newPosition.z() < swimlevel))
-            {
+            if (!(inertia.z() > 0.f) /* && !(newPosition.z() < swimlevel) */)
+            { // I don't want any limit for the walk on water effect !
                 osg::Vec3f from = newPosition;
                 osg::Vec3f to = newPosition - (physicActor->getOnGround() ?
                              osg::Vec3f(0,0,sStepSizeDown+2.f) : osg::Vec3f(0,0,2.f));
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -1152,9 +1152,9 @@ namespace MWPhysics
 
                 bool waterCollision = false;
                 if (effects.get(ESM::MagicEffect::WaterWalking).getMagnitude()
-                        && cell->getCell()->hasWater()
-                        && !world->isUnderwater(iter->first.getCell(),
-                                               osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
+                        && cell->getCell()->hasWater())
+//                        && !world->isUnderwater(iter->first.getCell(),
+//                                               osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
                     waterCollision = true;
 
                 ActorMap::iterator foundActor = mActors.find(iter->first);


scrawl:
This is for bug https://bugs.openmw.org/issues/1138, you can find a previous fix attempt at PR #683.

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -1152,9 +1152,9 @@ namespace MWPhysics
 
                 bool waterCollision = false;
                 if (effects.get(ESM::MagicEffect::WaterWalking).getMagnitude()
-                        && cell->getCell()->hasWater()
-                        && !world->isUnderwater(iter->first.getCell(),
-                                               osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
+                        && cell->getCell()->hasWater())
+//                        && !world->isUnderwater(iter->first.getCell(),
+//                                               osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
```
scrawl: Code that is no longer needed should be removed instead of commented out.

apps/openmw/mwphysics/actor.cpp
```diff
@@ -150,6 +150,16 @@ void Actor::setCanWaterWalk(bool waterWalk)
     {
         mCanWaterWalk = waterWalk;
         updateCollisionMask();
+	if (waterWalk) {
+	    ESM::Position pos = mPtr.getRefData().getPosition();
+	    if (pos.pos[2] < -100) {
```
scrawl: Don't use the number -100. That is wrong in several ways.

One, you can not assume the water to be at z=0. You need to take the cells' water level into account.

Second, you are assuming the player/actor is 100 units high. Use the actual height of the actor instead.

apps/openmw/mwphysics/actor.cpp
```diff
@@ -150,6 +150,16 @@ void Actor::setCanWaterWalk(bool waterWalk)
...
+		/* I didn't look for the exact limit but watercolision will
+		 * push the player out of water only if there is not too much
+		 * water, for -100 and beyond position must be adjusted ! */
+		pos.pos[2] = -50;
```
scrawl: That the player is being pushed out of water when slightly below it is just a coincidence. A peculiarity of the movement controller that might stop working or intentionally go away at some point in the future. Do not rely on it.

For the number -50, the same comment as for -100 applies.

apps/openmw/mwphysics/actor.cpp
```diff
@@ -150,6 +150,16 @@ void Actor::setCanWaterWalk(bool waterWalk)
...
+		mPtr.getRefData().setPosition(pos);
+	    }
+	}
```
scrawl: This code is also in the wrong place. A function called setCanWaterWalk(bool) should not be expected to have such side effects.

apps/openmw/mwphysics/actor.cpp
```diff
@@ -150,6 +150,16 @@ void Actor::setCanWaterWalk(bool waterWalk)
...
         mCanWaterWalk = waterWalk;
         updateCollisionMask();
+	if (waterWalk) {
+	    ESM::Position pos = mPtr.getRefData().getPosition();
+	    if (pos.pos[2] < -100) {
+		/* I didn't look for the exact limit but watercolision will
+		 * push the player out of water only if there is not too much
+		 * water, for -100 and beyond position must be adjusted ! */
+		pos.pos[2] = -50;
+		mPtr.getRefData().setPosition(pos);
```
scrawl: Before moving the actor, we should check if there's an obstacle in the way, or you could get stuck.

zelurker: Actually even in the vanilla game you could get stuck if using walk on
water in places where there is no room to walk on water !
For the "wrong place", well it seems ok to me since it always happens when
casting "walk on water" in vanilla, but I am opened to suggestions...


2015-09-15 14:13 GMT+02:00 scrawl <notifications@github.com>:

Koncord:
https://bugs.openmw.org/attachments/1489/potential%20fix.diff
This is works fine for me, except effect of removal. I hope this will help you!

zelurker:
There is no maxDepthForCast in the original game. Plus as it is here it
would really not be convenient, I tried this at a place where I saw a few
pearls which I could try to take and I couldn't cast the spell anymore once
at their level.
So for now I prefer my solution, but maybe the idea of a max depth to cast
is good, but for very big depths in this case...
Except that it works.

2015-09-17 3:34 GMT+02:00 Stanislav Zhukov <notifications@github.com>:

zelurker:
I found the perfect place to test this kind of patch :
coc "urshilaku, fragile burial"
there is an underwater passage there, as described there :
http://www.uesp.net/wiki/Morrowind:Urshilaku_Burial_Caverns
with a "worm lord" inside. Well if you cast walk on water wen you are still
under water but you see the surface above you just after passing the door :
 - with my version, you finish outside the world, passed through the
ceiling apparently, and the worst part is that you can't go back without
using the console tcl command !
 - with yours, it tries to tell that you can't cast that here, but as you
said in comments, it doesn't work. So you just go forward a little bit
after that, and bam ! same effect, through the ceiling and stuck there.

Perfect place to test, but I am not sure vanilla handles this place well
too, I didn't try to run vanilla to test this though...


2015-09-17 9:22 GMT+02:00 Emmanuel Anne <emmanuel.anne@gmail.com>:

zelurker:
And I found the solution while investigating the collisions, there was just
a test for fSwimLevel, if you remove this test the player is slowly pushed
to the surface while testing collisions, so it works everywhere and there
is no risk to be pushed out of the cell like before. It's even better than
vanilla because in vanilla you suddenly appear on top of water, here you
see the movement.


2015-09-18 19:11 GMT+02:00 Emmanuel Anne <emmanuel.anne@gmail.com>:

apps/openmw/mwphysics/physicssystem.cpp
psi29a: Please do what scrawl suggests, we can always go back in git history to get it if we needed it again.

zelurker: You know that it would take you only a few seconds to remove these lines by
yourselves ?
For me, each time I go back to this branch it means recompiling a lot of
stuff when I go back to my normal branch, it's just removing instead of
commenting out, a pure loss of time for me, you are welcome to do it
anytime you want !
Anyway, I guess I'll do it later, just to be sure it's ok now !

2015-10-01 21:59 GMT+02:00 Bret Curtis <notifications@github.com>:

psi29a: We appreciate it that you want to help, but with a project of this size, it makes our job easier when everyone follows the same coding guidelines.

I bring this up because you ignored scrawl and mine's advice twice now, 17 days ago and again yesterday for now comments.

We would rather bounce the patch back to your with critique to make it better and to ease you into our workflow.

As for wasting your time, I'm sorry about that. We had no idea that your time was worth more than our time.

zelurker: Don't forget too that it's very hard to arrive into a new project, things
that you would be able to do in a few minutes because you know the code
take a few days to new comers... It just helps when you show instead of
just saying, and after that people try to do their best next time.
Precious is probably not the right word to talk about my time, but I still
see it as a waste of time anyway !
Anyway I made a new branch, walk_on_water, because with this you can have a
very very tiny patch just removing a few lines to make all this to work
flawlessly.

2015-10-02 9:24 GMT+02:00 Bret Curtis <notifications@github.com>:

psi29a: You're not alone!  We're very welcoming to do new developers and we try to help guide them, like we are doing with you. Getting it right the first time pays off in the long run, and you're doing well otherwise! Great job!

Feel free to close this PR since you've opened a new one.

PR #762, opened 2015-09-15
fix yorick's ring script

zelurker:
This is a script attached to a ring which executes when equipped and displays a messagebox to ask what to do. With openmw it stops with the throw because apparently the previous messagebox is still active. This is a simple workaround.

### DIFF
--- a/apps/openmw/mwgui/messagebox.cpp
+++ b/apps/openmw/mwgui/messagebox.cpp
@@ -117,8 +117,12 @@ namespace MWGui
 
     bool MessageBoxManager::createInteractiveMessageBox (const std::string& message, const std::vector<std::string>& buttons)
     {
+
         if(mInterMessageBoxe != NULL) {
-            throw std::runtime_error("There is a message box already");
+	    // Yorick's ring comes here, I guess it's because the script is
+	    // executed more than once, but anyway just deleting the old
+	    // box allows it to run
+	    delete mInterMessageBoxe;
         }
 
         mInterMessageBoxe = new InteractiveMessageBox(*this, message, buttons);


zinnschlag:
Okay. As mentioned in the other post. We need to figure out what MW does here. Is OpenMW incorrectly running the script multiple times or is OpenMW incorrectly throwing an exception.

scrawl:
I just tested what happens in MW.
<pre>
messagebox, "a", "b", "c"
messagebox, "a", "d", "e"
</pre>

The second messagebox just overwrites the previous one.

So the change in this PR is fine. However I would like to see one more change before merging: we should print a warning message when overwriting a messagebox, because that usually indicates problems in the scripts. For example, say mod A made a messagebox, while at the same time mod B made another messagebox, unaware of mod A. One of those mods will then get the wrong result from getButtonPressed. This kind of conflict is something we can't fix until the scripting language is extended post-1.0, but we can at least emit a warning when it happens.

zelurker:
If you are in the subject of messageboxes, I also stumbled about 2
commented lines, and they need to be uncommented to allow messageboxes of
the form :
messagebox "message", "Equip"
Equip is seen as a keyword, not an argument and it produces an error. I
didn't open a pull request for that since the fix was already in the
source, but just commented. I uncomented it and tested it for a while and
found no issue so far. It's this :

diff --git a/components/compiler/scanner.cpp
b/components/compiler/scanner.cpp
index 3c5bb77..e93d66e 100644
--- a/components/compiler/scanner.cpp
+++ b/components/compiler/scanner.cpp
@@ -294,8 +294,8 @@ namespace Compiler
             name = name.substr (1, name.size()-2);
 // allow keywords enclosed in ""
 /// \todo optionally disable
-//            cont = parser.parseName (name, loc, *this);
-//            return true;
+            cont = parser.parseName (name, loc, *this);
+            return true;
         }

         int i = 0;



2015-11-12 18:25 GMT+01:00 scrawl <notifications@github.com>:

scrawl:
We knew about this, it's a side effect of MW allowing instructions to be enclosed in "" and still work, like
<pre>
player->"getlevel"
</pre>

See https://bugs.openmw.org/issues/2794

@Zini: Maybe we should revert the change? It seems like
<pre>
messagebox "asdf", "Equip"
</pre>
is a far more likely use case than some mod using instructions in quotes for some reason (which you can simply workaround by removing the quotes, but you can't workaround the former issue).

scrawl:
Bug [2991](https://bugs.openmw.org/issues/2991) "can't use keywords in strings for MessageBox"

zinnschlag:
I don't have a strong opinion about this, because both options suck. But I am slightly in favour of keeping the current implementation, because:

1. There are actually not that many script instructions that are regular English nouns or verbs that could be reasonable used as a single word string. Almost all of them at least two words like OnActivate or PlaceAtPc. I don't see one of the problem cases being a lot more likely than the other.
2. There are easy workarounds. Just append a space to the end of the word or simply rephrase the text.

But maybe there is a way to get both features working. Need to ponder about it.

zelurker:
2: yeah that's what I did for the mod which had the problem, but it was for
1 script, and there were 10+ maybe 20 of them to edit, so after that I
uncommented the fix in openmw, and finally I am not using the mod anyway !

But yes there is a better way to fix this, interpret based on context, but
it's easier said than done considering the way everything works...

2015-11-13 10:51 GMT+01:00 Marc Zinnschlag <notifications@github.com>:

scrawl:
https://github.com/OpenMW/openmw/commit/db71634a2d66dd6a41a706410463364e90b91c9b

PR #763, opened 2015-09-16
Duration in tooltips

zelurker:
I know it's not in vanilla morrowind, it's in skyrim actually for spell effects, and you easily get used to it, so since it's easy to add to morrowind, here it is.
And I also added the duration for lights, because I had forgotten they just disappear after some time !
Now at least you'll know why they disappear... ! (I didn't know their default duration is only 120s, quite ridiculous !)

### DIFF
--- a/apps/openmw/mwclass/light.cpp
+++ b/apps/openmw/mwclass/light.cpp
@@ -169,7 +169,7 @@ namespace MWClass
 
         std::string text;
 
-	text += "\nDuration: " + MWGui::ToolTips::toString(ptr.getClass().getRemainingUsageTime(ptr));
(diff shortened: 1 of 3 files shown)


scrawl:
Don't use tabs please (I just noticed this applies to your other PR's too). We use 4 spaces to indent.

scrawl:
Interesting idea. I'm fine with this, but since it isn't vanilla I suggest to introduce a new setting (that is disabled by default) to control the display of durations. Maybe 
<pre>
[Game]
# Show the remaining duration of magic effects and lights
show effect duration = false
</pre>

zelurker:
Ok for the setting, done, and ok for expandtab !

2015-09-16 19:54 GMT+02:00 scrawl <notifications@github.com>:

PR #764, opened 2015-09-16
Tab extermination

scrawl:
Removes tab characters for all source files in apps/ and components/. Adds a step to Travis-CI builds that checks for tab characters in source files, and fails the build if one is found.

The intent here is to notice mixed indentation *before* we merge a pull request. We should never merge a PR with tabs in it, because:
- The code won't line up depending on your editor's tab width.
- Removing tabs post-merge can create merge conflicts, and obscures the results of tools like git blame or git log with irrelevant changes.

Spotting & fixing tabs manually is annoying (and evidently, every once in a while something will slip through). So why not let travis handle it?

Testing performed:
- Added tab character, build fails https://travis-ci.org/scrawl/openmw/jobs/80709680


zinnschlag:
Die, tabs, Die!

PR #770, opened 2015-09-21
scale automatically documents to read (books, scrolls and journal)

zelurker:
Quite a bit to explain this time.
A few days ago I found the topic about "journal size" on the forum there : https://forum.openmw.org/viewtopic.php?f=2&t=3093
so as mentioned in the topic, I tried gui "scaling factor = 2" on a 1680x1050 screen to test. It's excellent for books, but it looks bad for most gui screens (even the title screen) and it's horrible for the character status screen.
So I tried to make a patch for this.
My first idea was to try to use mainWidget->setScale(width*2,height*2) in Layout when the layout is one of these (book, scroll or journal). But it doesn't work, the place is reserved on screen, but apparently the sub widgets are not resized and the font doesn't change.
So option 2 : adapt gui scaling factor to work only for these cases. It was longer than what I thought, but in the end it works well. It adapts to the resolution used (I tested with a window that I manually reiszed to many resolutions) and to the current gui scaling factor. It works for books, scrolls and journal.

Morrowind does this automatically, it resizes everything to read to fullscreen, so I don't think we need a setting to enable/disable this.

### DIFF
--- a/apps/openmw/mwbase/inputmanager.hpp
+++ b/apps/openmw/mwbase/inputmanager.hpp
@@ -56,6 +56,7 @@ namespace MWBase
             /// Returns if the last used input device was a joystick or a keyboard
             /// @return true if joystick, false otherwise
             virtual bool joystickLastUsed() = 0;
+            virtual void setScale(float uiscale) = 0;
     };
 }
 
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -272,7 +272,7 @@ namespace MWBase
 
             virtual void enableRest() = 0;
             virtual bool getRestEnabled() = 0;
-            virtual bool getJournalAllowed() = 0; 
+            virtual bool getJournalAllowed() = 0;
 
             virtual bool getPlayerSleeping() = 0;
             virtual void wakeUpPlayer() = 0;
@@ -359,6 +359,9 @@ namespace MWBase
             virtual void requestMap(std::set<MWWorld::CellStore*> cells) = 0;
             virtual void removeCell(MWWorld::CellStore* cell) = 0;
             virtual void writeFog(MWWorld::CellStore* cell) = 0;
+            virtual void setScale(float scalingFactor) = 0;
+            virtual float getBookScale() = 0;
+            virtual float getScrollScale() = 0;
     };
 }
 
--- a/apps/openmw/mwgui/journalwindow.cpp
+++ b/apps/openmw/mwgui/journalwindow.cpp
@@ -15,6 +15,7 @@
 #include <components/misc/stringops.hpp>
 #include <components/widgets/imagebutton.hpp>
 #include <components/widgets/list.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/soundmanager.hpp"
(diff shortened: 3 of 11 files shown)


scrawl:
I don't like this approach. The patch also resizes the HUD, message boxes and anything else that happens to be on screen while the journal/book is open.

Unfortunately I can't give you an alternative suggestion. Scaling isn't very well supported by MyGUI.

zelurker:
I must say I didn't love it too, but that's the best thing I could find,
and it actually works very well, it's a definite improvement over what we
had before !


2015-09-21 15:17 GMT+02:00 scrawl <notifications@github.com>:

zelurker:
Hey it would be nice that you accept it anyway : it's in 1 commit, so you
can easily revert it if one day you find the super solution better than
everything. But if you do it will probably take quite sometime and
meanwhile this is much better than a tiny document to read in the middle of
a very big screen. I agree it's not ideal, but it's better than nothing and
usually people improve code over time...


2015-09-21 21:46 GMT+02:00 Emmanuel Anne <emmanuel.anne@gmail.com>:

scrawl:
> meanwhile this is much better than a tiny document to read in the middle of
a very big screen.

The text is the same size as all other text in the game (dialog window, etc). I don't see a substantial usability issue there.

As for current state of this PR, the HUD becoming larger when a book opens is clearly a bug and can be irritating. I wouldn't want to merge until that's fixed.

zelurker:
irritating ??? Err, did you try "scaling factor = 2" to see how
"irritating" it is ?
And if you find a tiny book in the center of a screen has no readability
issue, we don't share the same view, clearly.

Well in this case I guess I'll have a real fork because I don't plan to
investigate about your "irritation" any time soon.


2015-09-22 18:57 GMT+02:00 scrawl <notifications@github.com>:

zinnschlag:
Having the scale of the UI changing while reading a book is clearly not acceptable. But the small books are an issue. With no good way to implement the correct behaviour around the corner, maybe we should think a bit out of the box.

This is just an idea and I do not know how feasible its implementation would be. But how about adding an option (cfg file only for now) that enables large books (basically full screen) and hides the UI while the book is open?

psi29a:
@zelurker while your contributions are appreciated, you need to be able to take rejection as not being personal.

Scrawl and Zini have a point, until the problem is corrected in MyGUI, there is really nothing further to be done.

Zini's suggestion is a possible work-around until MyGUI accepts proper scaling, unless you are willing to also send a PR to MyGUI to fix this issue correctly.

scrawl:
Idea:

Create a new type of Layer that supports scaling. Layers can be registered at runtime in MyGUI, just like other Factory items, so no changes to the upstream MyGUI code would be required.

This particular Layer implementation would talk to our RenderManager and enlarge the glViewport before rendering, then restore to its original values when done.

Widgets that require scaling (i.e. journal and books) would then be created on that particular layer.

scrawl:
We would need to override at least the following ILayer methods:

<pre>
		virtual ILayerItem* getLayerItemByPoint(int _left, int _top) const = 0;
</pre>
Needs to translate screen coordinates to the original (non-scaled) coordinates. We need this to get mouse input to work correctly.

<pre>
		virtual void renderToTarget(IRenderTarget* _target, bool _update) = 0;
</pre>
Needs to modify the viewport before and after rendering.

scrawl:
I've started to implement the scaling layer idea, it works very well. I will clean it up and push tomorrow.
layer

zelurker:
Yeah well I had something working on my side for more than 1 month now
while you were improving this.
But congratulations anyway ! :)

2015-11-08 3:53 GMT+01:00 scrawl <notifications@github.com>:

scrawl:
commit 516f2765a15c7c80fcc8784f953c1d5e2842bbd6

PR #771, opened 2015-09-21
Sky

scrawl:
cc @slothlife 

Implement sun flare effects, with several improvements over the old (Ogre) implementation:
- More accurate occlusion query, using a circular shape that matches the sun rather than a square.
- Implement the full-screen glare effect that we were missing before (makes the screen look "warmer" when you look directly at the sun during peak hours).
- Multiple camera support. Not used at the moment, but who knows when we'll need it. Using OSG callbacks made this so easy to support that I saw no reason to tie the code to a specific camera.

Miscellaneous sky changes:
- Implement Sun Disc Sunset Color
- Fade the sun in/out during sunrise/sunset


psi29a:
This is great! :)

Multiple camera support: I see this as an interesting feature if/when someone wants to make portals/rifts mod, or even later in multi-player support for spectators/ride-along.

PR #777, opened 2015-09-29
Fishing academy

zelurker:
Here are some small commits to fix problems with scripts in the fishing academy mod.
The mod still needs to be edited for syntax errors and also I couldn't find how the movement detection is reset to 0 in the vanilla game (a bgear variable needs to be set to 0 when the keys are released but I don't understand how it's done - just adding a single line fixes it in openmw).
With these changes almost everything works, the boats appear and can be moved, the collisions don't work though because apparently the setpos command doesn't check for collisions in openmw (as far as I can tell).

### DIFF
--- a/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
+++ b/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
@@ -710,14 +710,23 @@ namespace MWDialogue
         if (map != mChangedFactionReaction.end() && map->second.find(fact2) != map->second.end())
             return map->second.at(fact2);
 
-        const ESM::Faction* faction = MWBase::Environment::get().getWorld()->getStore().get<ESM::Faction>().find(fact1);
+        try {
+            const ESM::Faction* faction = MWBase::Environment::get().getWorld()->getStore().get<ESM::Faction>().find(fact1);
 
-        std::map<std::string, int>::const_iterator it = faction->mReactions.begin();
-        for (; it != faction->mReactions.end(); ++it)
-        {
-            if (Misc::StringUtils::ciEqual(it->first, fact2))
+            std::map<std::string, int>::const_iterator it = faction->mReactions.begin();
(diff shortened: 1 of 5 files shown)


scrawl:
@zinnschlag:

> This looks correct, but I don't like it. Too verbose. Not sure about a better solution though. Maybe changing updatePtr to take two arguments (from, to) and only change the ptr to to, if it was from originally? 

I was just about to suggest the same change then saw your comment. :)

@zelurker: Can you open a new PR for each commit in this PR please.

zelurker:
After some thinking yeah changing updatePtr is a better idea, it makes some
simpler code and it makes sure all the calls to updatePtr are fixed, so I
did it here, with a revert first on the previous commit and then the new
version modifying updatePtr.
Tested on fishing academy, it works.


2015-09-30 2:06 GMT+02:00 scrawl <notifications@github.com>:

scrawl:
I totally forgot about this PR. The commit "improves InterpreterContext::updatePtr" looks good. However I noticed a bug in positionCell that was already present before your commit:
<pre>
ptr = MWWorld::Ptr(ptr.getBase(), store);
</pre>
This is not the proper way to get the updated Ptr, it will only work for the player which isn't owned by any cell. For other objects, moving between cells makes the object owned by that cell and thus the getBase() pointer will change. 

It's fixed in master, which is going to conflict with this PR. Can you please pull the latest, then reapply your updatePtr patch and make a new PR.

zelurker:
Better late than never, but it's extreme here.
Anyway just made a new commit request, typo error in the branch name but
you should find it.
It's tested, it fixes the disappearing boat again.


2015-11-12 14:36 GMT+01:00 scrawl <notifications@github.com>:

scrawl:
Seems like the changes here are already applied or no longer needed.

Except for the commit "faction not found is not fatal like in vanilla ", but I don't think this one should be applied as is. We don't want the error to be silenced. If someone can come up with a better fix for this open a new PR.

PR #779, opened 2015-10-02
walk on water adjustments

zelurker:
the actor is always pushed out of water when casting the spell
This replaces the old walkonwater branch, way shorter !


scrawl:
See my comment on #761:

> That the player is being pushed out of water when slightly below it is just a coincidence. A peculiarity of the movement controller that might stop working or intentionally go away at some point in the future. Do not rely on it.

Allofich:
@scrawl 
I think this PR can be closed now after the recent water walking updates.

PR #781, opened 2015-10-04
Fix to https://bugs.openmw.org/issues/2914

artemutin:
Quick fix to this bug: "Magicka not recalculated on character generation" -  https://bugs.openmw.org/issues/2914. I have expanded already existing NpcStats.updateHealth method to update all 3 dynamic stats.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -295,7 +295,7 @@ namespace MWMechanics
             creatureStats.getSpells().add(*it);
 
         // forced update and current value adjustments
-        //mActors.updateActor (ptr, 0);
+        mActors.updateActor (ptr, 0);
 
         for (int i=0; i<3; ++i)
         {


scrawl:
This is duplicating code we already have elsewhere.

artemutin:
OK.Then I have some options to offer. If we don't want code duplication and fully rely on current behavior, the simplest way i see to fix bug is to uncomment this line in [buildPlayer()](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp#L298). But it belongs to the large commit of yours, and it was commented for some purpose, i guess?
Or, because  all seems to worked normal in version 0.36.1, should I try to bisect to find bad commit?
Also, i don't like current inconsistence - we call to [NpcStats::updateHealth](https://github.com/OpenMW/openmw/blob/9918b573419c62af85fe8ac7d750844409fc251b/apps/openmw/mwmechanics/npcstats.cpp#L305) to update health, next magicka recalculation is broken (or reling on [Actors::updateActor](https://github.com/OpenMW/openmw/blob/56163dcb2a33fdf67bd78dc63a6134c4f408505a/apps/openmw/mwmechanics/actors.cpp#L221) and Actors::calculateDynamicStats) and finally, stamina recalculates in [CreatureStats::setAttribute](https://github.com/OpenMW/openmw/blob/e76401d5eaf16fa752d94019449d81f788bbaf8c/apps/openmw/mwmechanics/creaturestats.cpp#L133) . 
So, what do you think?

scrawl:
> the simplest way i see to fix bug is to uncomment this line in buildPlayer(). But it belongs to the large commit of yours, and it was commented for some purpose, i guess?

That line was left commented by accident, it looks like. So if we re-enable it, everything is fine?

artemutin:
Yep. This one line fix this bug. But all this stuff, as I mentioned above, is a kind of a mess. 
Heh, while testing all these find yet another confirmation: If you after leaving a review menu on character creation, stand still and open stat menu, value of fatigue will be 200/200, no matter what correct value was calculated on character creation. Then move a bit, and value become correct.

artemutin:
Oh, and I just have fucked up with a push, should study git more ><. Here is one line to fix it.

PR #785, opened 2015-10-07
Search results edit redirection for scripts  (Feature #2946)

artemutin:
Search results in scripts files redirects to exact line and column, the same behaviour as a Verifier has.
https://bugs.openmw.org/issues/2946

### DIFF
--- a/apps/opencs/view/world/scriptsubview.cpp
+++ b/apps/opencs/view/world/scriptsubview.cpp
@@ -198,26 +198,38 @@ void CSVWorld::ScriptSubView::useHint (const std::string& hint)
     if (hint.empty())
         return;
 
-    if (hint[0]=='l')
-    {
-        std::istringstream stream (hint.c_str()+1);
-
-        char ignore;
-        int line;
-        int column;
-
-        if (stream >> ignore >> line >> column)
+    size_t line = 0, column = 0;
+    std::istringstream stream (hint.c_str()+2);
+    switch(hint[0]){
+        case 'R':
+        case 'r':
         {
-            QTextCursor cursor = mEditor->textCursor();
+            QModelIndex index = mModel->getModelIndex (getUniversalId().getId(), mColumn);
+            QString source = mModel->data (index).toString();
+            size_t pos;
+            stream >> pos >> pos;
 
-            cursor.movePosition (QTextCursor::Start);
-            if (cursor.movePosition (QTextCursor::Down, QTextCursor::MoveAnchor, line))
-                cursor.movePosition (QTextCursor::Right, QTextCursor::MoveAnchor, column);
-
-            mEditor->setFocus();
-            mEditor->setTextCursor (cursor);
+            for (size_t i = 0; i <= pos; ++i){
+                if (source[(unsigned) i] == '\n'){
+                    ++line;
+                    column = i;
+                }
+            }
+            column = pos - column - (line > 0 ? 1 : 0);
+            break;
         }
+        case 'l':
+            stream >> line >> column;
     }
+
+    QTextCursor cursor = mEditor->textCursor();
+
+    cursor.movePosition (QTextCursor::Start);
+    if (cursor.movePosition (QTextCursor::Down, QTextCursor::MoveAnchor, line))
+        cursor.movePosition (QTextCursor::Right, QTextCursor::MoveAnchor, column);
+
+    mEditor->setFocus();
+    mEditor->setTextCursor (cursor);
 }
 
 void CSVWorld::ScriptSubView::textChanged()
--- a/apps/opencs/view/world/scriptsubview.cpp
+++ b/apps/opencs/view/world/scriptsubview.cpp
@@ -198,28 +198,31 @@ void CSVWorld::ScriptSubView::useHint (const std::string& hint)
     if (hint.empty())
         return;
 
-    size_t line = 0, column = 0;
-    std::istringstream stream (hint.c_str()+2);
+    unsigned line = 0, column = 0;
+    char c;
+    std::istringstream stream (hint.c_str()+1);
     switch(hint[0]){
         case 'R':
         case 'r':
         {
             QModelIndex index = mModel->getModelIndex (getUniversalId().getId(), mColumn);
             QString source = mModel->data (index).toString();
-            size_t pos;
-            stream >> pos >> pos;
+            unsigned pos, dummy;
+            if (!(stream >> c >> dummy >> pos) )
+                return;
 
-            for (size_t i = 0; i <= pos; ++i){
-                if (source[(unsigned) i] == '\n'){
+            for (unsigned i = 0; i <= pos; ++i){
+                if (source[i] == '\n'){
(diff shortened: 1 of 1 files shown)


zinnschlag:
That is a little more complex than it needs to be. I don't like the extension of the hint format with redundant data. Even with the default R and r hints all the information you need to calculate line and column are available. I suggest you do these calculations in the ScriptSubView instead.

btw. QString has operator[]. Therefore there is no need to have an index count and an iterator when running over it while counting.

artemutin:
I tried to remove all calculation to subview, and got into trouble - it looks like, that ScriptEditor deletes some characters from string. Text in the searcher most of times is longer, than text of the same script in the this->mEditor->toPlainText(). Therefore, transferred linear position in R: hint, becomes useless in determination of exact position of searched text. It confuses me a lot, because it should also makes unprecise my calculations inside searcher, but it is not. 
May be i miss something obvious, i'll look more precisely into it tomorrow.

artemutin:
I had a close look at strings and found, that QPlainTextEdit eats a line endings. \r\n in setPlainText() becomes  \n in output of toPlainText(). This behavior fits with that fact, that cursor postion, calculated in searcher stays valid in scriptsubview.
Here is thread about it - https://forum.qt.io/topic/42464/qplaintextedit-eats-line-endings/5
Looks like that this behavior is platform-dependent, so how should i deal with that? Any advices welcome.

zinnschlag:
You can grab the original text from the model. Example in ScriptSubView::dataChanged.

artemutin:
Thanks for the help, it works now. 
Btw, is it okay, that branch has all these previous commits with adding and deleting? May be I should delete it or merge in some way?

psi29a:
You can do a rebase squash, which will compress everything into one commit.
```
rebase -i master
```

This lets you pick which commits to squash, usually the ones after the first commit... feel free to squash all or just the ones you don't want showing up.

Then it will ask you about the commit messages. 

When finish, do a git push --force to overwrite your branch (and your PR here).

apps/opencs/view/world/scriptsubview.cpp
```diff
@@ -198,26 +198,38 @@ void CSVWorld::ScriptSubView::useHint (const std::string& hint)
...
-            cursor.movePosition (QTextCursor::Start);
-            if (cursor.movePosition (QTextCursor::Down, QTextCursor::MoveAnchor, line))
-                cursor.movePosition (QTextCursor::Right, QTextCursor::MoveAnchor, column);
-
-            mEditor->setFocus();
-            mEditor->setTextCursor (cursor);
+            QModelIndex index = mModel->getModelIndex (getUniversalId().getId(), mColumn);
+            QString source = mModel->data (index).toString();
+            size_t pos;
+            stream >> pos >> pos;
```
zinnschlag: This looks weird. Not entirely sure if it is even legal.

apps/opencs/view/world/scriptsubview.cpp
```diff
@@ -198,26 +198,38 @@ void CSVWorld::ScriptSubView::useHint (const std::string& hint)
...
-    {
-        std::istringstream stream (hint.c_str()+1);
-
-        char ignore;
-        int line;
-        int column;
-
-        if (stream >> ignore >> line >> column)
+    size_t line = 0, column = 0;
+    std::istringstream stream (hint.c_str()+2);
```
zinnschlag: The error checking is a bit weak here. If hint isn't empty hint.c_str()+1 is guaranteed to be valid. The same can not be said of +2 in a case of an ill-formed hint. Please use the method utilised in the original code.

apps/opencs/view/world/scriptsubview.cpp
```diff
@@ -198,26 +198,38 @@ void CSVWorld::ScriptSubView::useHint (const std::string& hint)
...
-
-            mEditor->setFocus();
-            mEditor->setTextCursor (cursor);
+            QModelIndex index = mModel->getModelIndex (getUniversalId().getId(), mColumn);
+            QString source = mModel->data (index).toString();
+            size_t pos;
+            stream >> pos >> pos;
+
+            for (size_t i = 0; i <= pos; ++i){
+                if (source[(unsigned) i] == '\n'){
```
zinnschlag: What's the point of this cast? Looks like a no-op to me. At most it narrows the type down, which shouldn't do anything under realistic conditions.

apps/opencs/view/world/scriptsubview.cpp
```diff
@@ -198,26 +198,38 @@ void CSVWorld::ScriptSubView::useHint (const std::string& hint)
...
+                    ++line;
+                    column = i;
+                }
+            }
+            column = pos - column - (line > 0 ? 1 : 0);
+            break;
         }
+        case 'l':
+            stream >> line >> column;
```
zinnschlag: Again, error checking. If the hint is ill-formed, depending on the version of the C++ standard the code is compiled with, the value of line and column may be undefined.

apps/opencs/view/world/scriptsubview.cpp
artemutin: If i dont make this cast, compilation on my machine fails with this error message:
error: ambiguous overload for 'operator[]' (operand types are 'QString' and 'size_t {aka long unsigned int}')
It cant choose between [int] and [uint] methods. I can use "unsigned i = 0" though, if this cast is problem.

apps/opencs/view/world/scriptsubview.cpp
artemutin: Okay, i just thought that hint is internal thing, and we could rely on his consistency with format [char]: some info.
But I can't understand how this check works: "if (stream >> ignore >> line >> column)". >> operator returns istream&, that somehow casts to bool, but according to [this](http://www.cplusplus.com/reference/sstream/istringstream/?kw=istringstream), bool operator was introduced in c++11.

apps/opencs/view/world/scriptsubview.cpp
artemutin: Why not? >> has left-to-right associativity. Of course i can use a dummy variable, just thought it's not necessary.

zinnschlag:
I will merge that now. Good enough. But for future reference try to declare variables as local as possible (c) and for Qt-related index problems you can just use int. There is nothing to be gained by using an unsigned type.

apps/opencs/view/world/scriptsubview.cpp
zinnschlag: Even internally we should shield ourself against undefined behaviour in case of errors where we can. Defensive coding and such. But once we get past 1.0 and have editor plugins the hint system will be exposed to plugins. Having a broken plugin make OpenMW-CS crash with a segfault would be bad.

> But I can't understand how this check works: "if (stream >> ignore >> line >> column)". >> operator returns istream&, that somehow casts to bool, but according to this, bool operator was introduced in c++11.

Nope. That worked since c++98 at least. Probably even pre-standard C++. Implicit cast to const void * IIRC.

PR #788, opened 2015-10-11
Fix for (Bug #2956)

artemutin:
Here it is - [bug](https://bugs.openmw.org/issues/2956).
What is done: 
* Preventing  update of filter model on every typed char in description field.
* Preventing reapply of empty filter on filter model update.

### DIFF
--- a/apps/opencs/view/filter/editwidget.cpp
+++ b/apps/opencs/view/filter/editwidget.cpp
@@ -7,7 +7,7 @@
 #include "../../model/world/data.hpp"
 
 CSVFilter::EditWidget::EditWidget (CSMWorld::Data& data, QWidget *parent)
-: QLineEdit (parent), mParser (data)
+: QLineEdit (parent), mParser (data), mIsEmpty(true)
 {
     mPalette = palette();
     connect (this, SIGNAL (textChanged (const QString&)), this, SLOT (textChanged (const QString&)));
@@ -27,6 +27,16 @@ CSVFilter::EditWidget::EditWidget (CSMWorld::Data& data, QWidget *parent)
 
 void CSVFilter::EditWidget::textChanged (const QString& text)
 {
+    //no need to parse and apply filter if it was empty and now is empty too.
+    //e.g. - we modifiing content of filter with already opened some other (big) tables.
+    if (text.length() == 0){
+        if (mIsEmpty)
+            return;
+        else
+            mIsEmpty = true;
+    }else
+        mIsEmpty = false;
+
     if (mParser.parse (text.toUtf8().constData()))
     {
         setPalette (mPalette);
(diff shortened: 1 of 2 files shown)


zinnschlag:
2nd commit is fine. I do not agree with the first one. Reasons:

- Updates for long text fields are performed now only when the focus leaves the widget. That is generally not desirable.
- Changes to the description field of a filter should not trigger re-application of filters at all.

artemutin:
I am agree with second reason, i have thought about it and find, that dataChanged signal has a way of reporting about different kind of changes, by providing a vector of data roles, which was modified. If this is option, i could dig more into it, because quick look to code didn't clarify much to me.

zinnschlag:
Sorry for the late reply (busy day yesterday). I haven't looked at the code for a while, but you should be able to intercept undesired updates on the slot that receives the update information. These updates come with row and column index. If you filter out updated that only affect the description column you should be fine.

artemutin:
Thank you for the help. I should look more carefully to existed code, i missed the actual meaning of Indexes passed in dataChanged slot.
So, i found two reasons for filters to slow down on large tables. 
1) The one call IdTable.setData emits two signal, with data column, that was modified, and state column. The filterDataChanged receives them both, and triggers filter reapplication two times. I fix this with new commit.
2) While debug, i found, that filterDataChanged actually receives aforementioned pair of signals two times. It is strange, and i'll look more into it tomorrow.
I've also got some newbie's questions, if I may ask.
1) Is checks, that I made in last commit enough for code to be safe in a way of "defensive coding and such"? For now it checks columns in every call to a signal, may be it is safe to take it up to constructor of widget? Also, if actually one of columns in table will be missed, findColumnIndex will throw an exception. I pass it "as is", but i can use a try-catch to modify it and give more precise description, what is a preferred way?
2) In order to investigate the second reason of slowdown, i need to debug and find unknown emitter of the signal. What is the best way to do it?

zinnschlag:
>  Is checks, that I made in last commit enough for code to be safe in a way of "defensive coding and such"?

Not really. The problem is here:

> topLeft.column() == bottomRight.column()

We should not rely on an update only affect one column. That can implode really badly.

> For now it checks columns in every call to a signal, may be it is safe to take it up to constructor of widget?

You mean storing the column index in a member variable that is initialised in the constructor? Sure, we already do that in several places.

> Also, if actually one of columns in table will be missed, findColumnIndex will throw an exception. I pass it "as is", but i can use a try-catch to modify it and give more precise description, what is a preferred way?

This is a bit tricky, because Qt. The documentation of Qt's exception guarantees is pretty weak. However generally throwing an exception as reactions to user input is okay. Most of the code is build to handle that properly.
In this case I wouldn't worry about it. A missing column would be a pretty big bug and OpenMW-CS is probably handling it gracefully enough. Not worth worrying about.

zinnschlag:
The idea here is that topLeft and bottomRight describe a rectangle within the table. That's how these update signals are designed. Any model subclass that does not follow this principle would have to be considered broken.

artemutin:
I got it. I can't rely on signal to send update by one column at time because it's not a part of description in docs or in function's argument names and signature, but from the same point of view, column indexes must be correct or at least it is not my problem.

PR #789, opened 2015-10-13
Cell Arrows

zinnschlag:
https://bugs.openmw.org/issues/1280

Work in progress ...

### DIFF
--- a/apps/opencs/view/render/cellarrow.cpp
+++ b/apps/opencs/view/render/cellarrow.cpp
@@ -118,8 +118,9 @@ void CSVRender::CellArrow::buildShape()
     for (int i=0; i<6; ++i)
         colours->push_back (osg::Vec4f (0.8f, (i==2 || i==5) ? 0.6f : 0.4f, 0.0f, 1.0f));
 
-    geometry->setColorArray (colours);
-    geometry->setColorBinding (osg::Geometry::BIND_PER_VERTEX);
+    geometry->setColorArray (colours, osg::Array::BIND_PER_VERTEX);
+
+    geometry->getOrCreateStateSet()->setMode (GL_LIGHTING, osg::StateAttribute::OFF);
 
     osg::ref_ptr<osg::Geode> geode (new osg::Geode);
     geode->addDrawable (geometry);
--- a/apps/opencs/view/render/cellarrow.cpp
+++ b/apps/opencs/view/render/cellarrow.cpp
@@ -116,7 +116,7 @@ void CSVRender::CellArrow::buildShape()
     for (int i=0; i<6; ++i)
         colours->push_back (osg::Vec4f (1.0f, 0.0f, 0.0f, 1.0f));
     for (int i=0; i<6; ++i)
-        colours->push_back (osg::Vec4f (1.0f, 0.0f, 0.4f, 1.0f));
+        colours->push_back (osg::Vec4f (0.8f, (i==2 || i==5) ? 0.6f : 0.4f, 0.0f, 1.0f));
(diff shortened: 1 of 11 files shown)


zinnschlag:
Interactive functionality still missing at this point. Need to work out button mapping (not easy; bloody mouse and keyboard do not have enough buttons for all functions)

@scrawl First attempt at properly working with OSG rendering. Could you look over it (CellArrow class) to see if I missed something?

apps/opencs/view/render/cellarrow.cpp
```diff
@@ -0,0 +1,148 @@
...
+    side1->push_back (5+6);
+    side1->push_back (3+6);
+    geometry->addPrimitiveSet (side1);
+
+    osg::DrawElementsUShort *side2 = new osg::DrawElementsUShort (osg::PrimitiveSet::QUADS, 0);
+    side2->push_back (4+6);
+    side2->push_back (5+6);
+    side2->push_back (2+6);
+    side2->push_back (1+6);
+    geometry->addPrimitiveSet (side2);
```
scrawl: You could put those faces in the same DrawElementsUShort (i.e. the same draw call) for more efficient rendering. There's no need to start a new draw call for every face.

<pre>
    osg::DrawElementsUShort *quads = new osg::DrawElementsUShort (osg::PrimitiveSet::QUADS, 0);
// back
+    quads->push_back (3+6);
+    quads->push_back (4+6);
+    quads->push_back (1+6);
+    quads->push_back (0+6);

// side1
+    quads->push_back (0+6);
+    quads->push_back (2+6);
+    quads->push_back (5+6);
+    quads->push_back (3+6);

// side2
+    quads->push_back (4+6);
+    quads->push_back (5+6);
+    quads->push_back (2+6);
+    quads->push_back (1+6);
+    geometry->addPrimitiveSet (quads);
</pre>

You could also put top and bottom in the same array, but would need to decide on one index type (triangles, quads, triangle strip, ...)

zinnschlag:
Fixed. Thanks.

scrawl:
Just tried it out. The lighting on the arrows seems to be glitched, most likely because there are no normals.

If you don't want to add normals, you can disable lighting like so:

<pre>
    osg::StateSet* stateset = geometry->getOrCreateStateSet();
    stateset->setMode(GL_LIGHTING, osg::StateAttribute::OFF);
</pre>

apps/opencs/view/render/cellarrow.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    osg::Vec4Array *colours = new osg::Vec4Array;
+
+    for (int i=0; i<6; ++i)
+        colours->push_back (osg::Vec4f (1.0f, 0.0f, 0.0f, 1.0f));
+    for (int i=0; i<6; ++i)
+        colours->push_back (osg::Vec4f (0.8f, (i==2 || i==5) ? 0.6f : 0.4f, 0.0f, 1.0f));
+
+    geometry->setColorArray (colours);
+    geometry->setColorBinding (osg::Geometry::BIND_PER_VERTEX);
```
scrawl: You can merge these two calls into one (and avoid using setColorBinding, which is a deprecated function).

<pre>
geometry->setColorArray (colours, osg::Geometry::BIND_PER_VERTEX);
</pre>

zinnschlag: That doesn't build for me. 

geometry->setColorArray (colours, osg::Array::BIND_PER_VERTEX);

does though. Fixing ...

PR #795, opened 2015-11-03
Fix for bug #2952: Enchantment Merchant Items reshuffled EVERY time 'barter' is clicked

MatthewRock:
This should fix the bug #2952. Unfortunately, it's my first time when I'm sending pull request, meaning that I don't know what is causing the collision and how can I resolve it - therefore, I would be grateful for aid in sending the pull request.

As for the commit, it's fix for the bug, plus some documentation, so when someone is stuck there next time, it should be easier to jump in and fix the problem. Personally I don't think that the solution is the best one, but I couldn't find anything better.


jeaye:
@MatthewRock There are other changes being pulled in here which are unrelated to your actual change, which was only a single commit. You shoud do something like:

```bash
$ git remote add upstream https://github.com/OpenMW/openmw.git
$ git fetch upstream
$ git checkout upstream/master
$ git checkout -b fix-2952
$ git cherry-pick 21f3657c0e6611818856a5d80d0e7958a39d4047
$ git push origin fix-2952
```

Then make a new PR from the `fix-2952` branch into OpenMW master. Fixes should be on their own branch, as I've shown, and you should branch from OpenMW's latest master. Your new PR should be just a single commit and, hopefully, the conflicts will be gone. If not, the conflicts should occur when you `git cherry-pick` and you can resolve them locally.

MatthewRock:
I agree. It was the first code I got to work with, and it took me some time to understand what was really going on(also, documentation is unfortunately a bit lacking).

When I fix these issues, should I edit this pull request, or delete this one and create new?

jeaye:
@MatthewRock First fix the PR so that it's bringing in just the commit you want; follow the steps I gave above. Then fix up the code.

MatthewRock:
Roger. Thanks for help!

PR #796, opened 2015-11-03
Fix for bug #2952

MatthewRock:
Enchantment Merchant Items reshuffled EVERY time 'barter' is clicked.

I hope everything is okay now.

### DIFF
--- a/apps/openmw/mwmechanics/levelledlist.hpp
+++ b/apps/openmw/mwmechanics/levelledlist.hpp
@@ -1,9 +1,12 @@
 #ifndef OPENMW_MECHANICS_LEVELLEDLIST_H
 #define OPENMW_MECHANICS_LEVELLEDLIST_H
 
-#include <components/misc/rng.hpp>
-
+#include <algorithm>
+#include <string>
+#include <vector>
 #include <iostream>
+
+#include <components/misc/rng.hpp>
 
 #include "../mwworld/ptr.hpp"
 #include "../mwworld/esmstore.hpp"
@@ -16,7 +19,6 @@
 
 namespace MWMechanics
 {
-
     /// @return ID of resulting item, or empty if none
     inline std::string getLevelledItem (const ESM::LevelledListBase* levItem, bool creature, unsigned char failChance=0)
     {
@@ -80,7 +82,5 @@ namespace MWMechanics
                 return getLevelledItem(ref.getPtr().get<ESM::CreatureLevList>()->mBase, true, failChance);
         }
     }
-
 }
-
 #endif
--- a/apps/openmw/mwworld/containerstore.cpp
+++ b/apps/openmw/mwworld/containerstore.cpp
@@ -27,9 +27,9 @@ namespace
         float sum = 0;
 
         for (typename MWWorld::CellRefList<T>::List::const_iterator iter (
-            cellRefList.mList.begin());
-            iter!=cellRefList.mList.end();
-            ++iter)
+                                                                          cellRefList.mList.begin());
+             iter!=cellRefList.mList.end();
+             ++iter)
         {
             if (iter->mData.getCount()>0)
                 sum += iter->mData.getCount()*iter->mBase->mData.mWeight;
@@ -40,7 +40,7 @@ namespace
 
     template<typename T>
     MWWorld::Ptr searchId (MWWorld::CellRefList<T>& list, const std::string& id,
-        MWWorld::ContainerStore *store)
+                           MWWorld::ContainerStore *store)
     {
         std::string id2 = Misc::StringUtils::lowerCase (id);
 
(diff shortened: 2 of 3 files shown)


apps/openmw/mwworld/containerstore.cpp
```diff
@@ -461,27 +461,53 @@ void MWWorld::ContainerStore::addInitialItem (const std::string& id, const std::
...
+    //See if any of generated items was sold, take note for further restocking:
     for (std::map<std::string, int>::iterator it = mLevelledItemMap.begin(); it != mLevelledItemMap.end(); ++it)
     {
-        if (count(it->first) >= it->second)
+        int itemCount = count(it->first);
+        //If it was sold
+        if (itemCount < it->second)
+        {
+            //Remove it from the shop
             remove(it->first, it->second, ptr);
```
scrawl: I think this line is wrong. The sold item is already gone. No need to remove it.

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -461,27 +461,53 @@ void MWWorld::ContainerStore::addInitialItem (const std::string& id, const std::
...
+            std::string group = levelledItemGroup(it->first, items);
+            //And add information to the allowedForReplace map
+            std::map<std::string, int>::iterator groupInMap = allowedForReplace.find(group);
+            if( groupInMap == allowedForReplace.end())
+                groupInMap->second = 0;
+            groupInMap->second += std::abs(it->second - itemCount);
+            //Also, remove the record from mLevelledItemMap, if it was the last item in stock.
+            if(!itemCount)
```
scrawl: We also need to adjust mLevelledItemMap if that wasn't the last item in stock. Set the map value to the actual count of items i.e. count(it->first);.

apps/openmw/mwmechanics/levelledlist.hpp
```diff
@@ -80,7 +82,50 @@ namespace MWMechanics
...
+            continue;
+        // If the item belongs to levelled list, return it
+        if (MWBase::Environment::get().getWorld()->getStore().get<ESM::ItemLevList>().search(it->mItem.toString()))
+        {
+            std::string group = Misc::StringUtils::lowerCase(it->mItem.toString());
+            if(itemInLevelledGroup(item, group))
+                return group;
+        }
+    }
+    return std::string();
```
scrawl: These functions belong to the restocking logic and aren't reused anywhere else so I would rather them placed in containerstore.cpp.

apps/openmw/mwmechanics/levelledlist.hpp
```diff
@@ -80,7 +82,50 @@ namespace MWMechanics
                 return getLevelledItem(ref.getPtr().get<ESM::CreatureLevList>()->mBase, true, failChance);
         }
     }
+}
 
+//Returns list of items in group. You get them by returned value->mId.
+inline const std::vector<ESM::LevelledListBase::LevelItem>& itemsInGroup(const std::string& group)
+{
+    MWWorld::ManualRef ref (MWBase::Environment::get().getWorld()->getStore(), group);
+    const ESM::ItemLevList* levItem = ref.getPtr().get<ESM::ItemLevList>()->mBase;
```
scrawl: You can get to the ESM::ItemLevList without constructing an item based on it.
```
const ESM::ItemLevList* levItem = MWBase::Environment::get().getWorld()->getStore().get<ESM::ItemLevList>().find(group);
```

And that makes the itemsInGroup function so short that I would remove it all together, just place the code inline.

apps/openmw/mwmechanics/levelledlist.hpp
```diff
@@ -80,7 +82,50 @@ namespace MWMechanics
...
+inline bool itemInLevelledGroup(const std::string& name, const std::string& group)
+{
+    //Get all items from the group
+    const std::vector<ESM::LevelledListBase::LevelItem>& elements = itemsInGroup(group);
+    //And see if item we're looking for is there.
+    for(std::vector<ESM::LevelledListBase::LevelItem>::const_iterator it = elements.begin(); it != elements.end(); ++it)
+    {
+        if(it->mId == name)
+            return true;
+    }
```
scrawl: This function doesn't account for the possibility of nested levelled lists. You can have a levelled list that contains an item that is another levelled list.

scrawl:
I think the cleanest fix would be to adjust mLevelledItemMap to include the ID of the levelled list that spawned the item. Then we don't have to do all this probing to find out where the item could have came from. Note however that a change to the map would break the saved game format. But we can handle the break gracefully - for old savegames that don't include the new info, items will restock with the "incorrect" behaviour once, then after the first respawn the levelled list source info is there so all restocks after that will work as intended.

apps/openmw/mwmechanics/levelledlist.hpp
MatthewRock: Can I rely on checking if item is levelled list by checking first two characters ("l_")?. Also, should I change naming from levelledGroup to levelledList?

MatthewRock:
Sorry for the mess, but my emacs indented the code... Do we have some style guide for this type of indentation?(switch/case | indentation of multiline multiple arguments of function)? The changed one looks better to me.

As for the fix, I didn't want to break anything, so I decided to create a new solution, using the resources I can get. Right now it looks pretty decent to me, but if you want, I can take your path. Do you mean replacing std::string with std::pair<std::string, std::string> in mLevelledItemMap?

scrawl:
> Sorry for the mess, but my emacs indented the code... Do we have some style guide for this type of indentation?(switch/case | indentation of multiline multiple arguments of function)?

Not really. Just use whatever the surrounding code you are working in is using. And please don't reformat existing code, that makes it harder to find the actual changes in the diff view.

When adjusting the code based on review comments please make new commits instead of rebasing, otherwise I can't tell what has changed.

> As for the fix, I didn't want to break anything, so I decided to create a new solution, using the resources I can get. Right now it looks pretty decent to me, but if you want, I can take your path. 

I'm sure the current fix will work, but it's a little more complicated than necessary, and fails to handle some edge cases (like the nested levelled lists, or a levelled list being changed by a mod in between game sessions, or not knowing if an item came from list A or list B if both lists contain the item). So I would prefer the alternate approach of changing mLevelledItemMap.

> Do you mean replacing std::string with std::pair in mLevelledItemMap?

Yeah, that would be one way of doing it. Or a map<std::string, std::map<std::string, int> >.

MatthewRock:
Yeah, I am really sorry for the reformatting thing, it happened accidentally. Alright then, I will go on and replace current fix with the one you're suggesting.

scrawl:
#850

PR #798, opened 2015-11-07
gcd fix : current stats modified by a ratio instead of a diff

zelurker:
This one took me quite some time. Contrary to what the mod status page says, gcd doesn't work well at all with openmw, current release at all. It shows a lot when using the mentor ring found at the beginning of the game :
 - start a new game with gcd
 - when gcd is initialized (you get the message), add the mentor ring to the inventory : player->additem ring_mentor_unique 1
 - then cast a spell so that mana points are not maximum anymore
 - then each time to put the ring on and then remove it you'll loose some spell points. If done enough times spell points can fall to 0 rather quickly.

After some digging in the scripts of gcd, it's because when the intelligence is modified, it expects that the ratio between current magicka and maximum magicka stays the same, which means that you can't do currentMagicka + diff as it was done until now.
So this patch just does that, it computes the ratio before changing current and modified states and then applies the ratio to get the new current value instead of adding diff like before.
Now gcd works !

### DIFF
--- a/apps/openmw/mwmechanics/stat.cpp
+++ b/apps/openmw/mwmechanics/stat.cpp
@@ -141,8 +141,13 @@ namespace MWMechanics
     template<typename T>
     void DynamicStat<T>::modify (const T& diff, bool allowCurrentDecreaseBelowZero)
     {
-        mStatic.modify (diff);
-        setCurrent (getCurrent()+diff, allowCurrentDecreaseBelowZero);
+	double ratio;
+	if (mStatic.getModified()>0)
+	    ratio = getCurrent()/mStatic.getModified();
+	else
+	    ratio = 0;
+	mStatic.modify (diff);
+	setCurrent (mStatic.getModified()*ratio, allowCurrentDecreaseBelowZero);
     }
     template<typename T>
     void DynamicStat<T>::setCurrent (const T& value, bool allowDecreaseBelowZero)


scrawl:
Yes, I knew we were missing ratio adjustment, see https://wiki.openmw.org/index.php?title=Research:Stats_and_Levelling#Actor_dynamic_stats .

According to the wiki this applies to magicka and fatigue. We need to test if it applies to health as well.

scrawl:
This PR should fail the travis build because tabs were used for indentations instead of spaces. Let's see if the script works...

zelurker:
For the health :
loading game in morrowind, health 47/47
player->modcurrenthealth -23
-> 24/47
player->modendurance 10
-> 29/57
confirmed !

Sorry for the tabs, I re-enabled the tabs here and forgot about this.

2015-11-08 0:54 GMT+01:00 scrawl <notifications@github.com>:

scrawl:
There are other methods we need to adjust as well.
<pre>
void DynamicStat<T>::setModifier (const T& modifier, bool allowCurrentDecreaseBelowZero)
void DynamicStat<T>::setModified (T value, const T& min, const T& max)
void DynamicStat<T>::setBase (const T& value)
</pre>

zelurker:
Actually my test was a little fast this morning, I forgot gcd was enabled !
Without gcd you can't see the effect on the health of a change to the
endurance, it probably only happens on level change, which makes it much
harder to see.
Anyway that's the expected behavior from gcd, and the guy who wrote this
spent a huge amount of time testing it, so it's probably the right thing to
do !
Not sure why you would want to change the other methods, in my tests
everything is ok now, this pull request was mainly to let you know about
this, it's done !


2015-11-08 13:31 GMT+01:00 scrawl <notifications@github.com>:

scrawl:
> Not sure why you would want to change the other methods, in my tests
everything is ok now, this pull request was mainly to let you know about
this, it's done !

Because not all code setting modifiers is using the setModified(...) method that you changed, there are different overrides provided for convenience that are supposed to do the same thing, only now they don't do the same thing any more because you applied the change to one but not the others. For example, applying a fortify magicka effect would now still use the old behaviour whereas magic gained by intelligence boosts uses the new ratio behaviour. This is inconsistent and can only lead to further bugs and surprises.

scrawl:
> Without gcd you can't see the effect on the health of a change to the
endurance, it probably only happens on level change, which makes it much
harder to see.

Yeah, I guess levelling up is the only way to test.

scrawl:
I tested behaviour with health in the vanilla MW executable. Unlike magicka, the current/max ratio is not preserved. If you get X health added by a levelup, that value is added to the current health as well as to the maximum health.
<pre>
; health 31/62
enablelevelupmenu
; health 34/65
</pre>

PR #800, opened 2015-11-11
add support for ffmpeg29 thanks to Andreas Cadhalpun

psi29a:
It would be nice if this could go into 0.37

This should close:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=803848

This should also close this:
https://bugs.openmw.org/issues/3003


zinnschlag:
This does not build for me: https://gist.github.com/zinnschlag/e5cbdef5f95ac71975c2

And Travis does not seem to like it either. Looks like the change broke compatibility with older versions of ffmpeg.

psi29a:
That's a pity, looks like we'll have to do a bit of #ifdef on ffmpeg versions.  Because when ffmpeg29 transition happens in Debian (soon), then OpenMW will no longer compile.

psi29a:
To many collisions between libav and ffmpeg it would seem. This really needs some #ifdefs, unless we can figure out a more subtle way of handling the libav stuff.

scrawl:
The plan was to drop libav support once the real FFmpeg is in the debian repos again. Maybe it's time to do that now.

psi29a:
@scrawl we need a solution for Travis, as it is still 12.04 with libav. I tried to backport ffmpeg, but it collides with libav.

Another solution is to use ffmpeg static libraries:
http://johnvansickle.com/ffmpeg/

scrawl:
Just to be clear, FFmpeg 2.9 is not released yet and we don't know when it will be. So I don't consider this PR critical for the 0.37 release.

psi29a:
That's fine, we can keep this for 0.37 release in downstream (only effects Debian/Ubuntu to shutup ffmpeg2.8 warnings).

scrawl:
Please do not make downstream patches unless absolutely necessary. A bunch of compile warnings are no compelling reason. 

The problem with downstream patches is that when we receive bug reports, we need to consider where the user got their package from and what downstream patches the respective packager may or may not have applied that might cause additional bugs not present in the main tree.

psi29a:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=803848

"Dear Maintainer,

your package fails to build with the upcoming ffmpeg 2.9.
This bug will become release-critical at some point when the
ffmpeg2.9 transition gets closer."

psi29a:
Just pushed ffmpeg2.8 for precise up to our ppa. Can you test this out Zini? Which Distro/version/release are you running?

zinnschlag:
The ppa is still ppa:openmw/openmw ?
Did an apt-get update, but don't see anything.

I am on Linux Mint 17.1.

zinnschlag:
Hm, my software sources say openmw/openmw/ubuntu trusty main. And indeed it seems Mint 17.1 is based on 14.04.

psi29a:
Roget that, I'll get ffmpeg to Trusty asap! :)

psi29a:
@zinnschlag FFMpeg 2.8 is now live ppa:openmw/openmw for trusty as well.

zinnschlag:
Actually the mint update manager handled that out of the box. No need to manually apt-get anything. The branch builds and runs for me now.

So, what is the verdict? Shall we go for it? 0.37.0 or 0.38.0? @scrawl @kcat

psi29a:
The argument is that the only reason libav exists is because Debian/Ubuntu supported it. That support is now gone and there is no other distro that use libav by default.

kcat:
It should definitely be done at some point. Whether or not for 0.37 I don't know, since technically you already started the RC phase, but scrawl also just recently made some sizeable changes when fixing a memory leak, which could probably do with a little extra testing. If psi29a thinks it should be for 0.37, I won't object.

scrawl:
I would argue for 0.38, because:
- it's too complex for a last-minute change and needs extensive testing.
- if we change now, release packagers need to upgrade their FFmpeg builds which is going to delay the release even more.
- FFmpeg 2.9 is not released nor in RC yet.

psi29a:
To address the points:
- Not as complex as your recent changes. It actually fixes the warnings produced in current builds now. I've play tested this with Morrowind and AoA without problems.
- The only packagers affected by this change are Debian/Ubuntu, as everyone else uses FFMpeg anyway. No one needs to upgrade their existing FFMpeg packages, but this does fix their warnings.
- This is correct, but I'm all for dropping libav... the sooner the better.

scrawl:
> The only packagers affected by this change are Debian/Ubuntu, as everyone else uses FFMpeg anyway. No one needs to upgrade their existing FFMpeg packages, but this does fix their warnings.

Good point.

> I'm all for dropping libav... the sooner the better.

Me too. This PR hasn't actually dropped the support yet though. Do a search for "avresample" and nuke those, mostly the libavwrapper.

Well, in that case I won't object to this going in 0.37. I still need to review and test it though.

psi29a:
@ace13 can you validate that you only use ffmpeg, and which version?

@corristo same to you, which ffmpeg?

corristo:
@psi29a here's what I'm using now: https://github.com/OpenMW/homebrew-openmw/blob/master/openmw-ffmpeg.rb#L5

scrawl:
avcodec_default_get_buffer2 was added in commit https://github.com/FFmpeg/FFmpeg/commit/759001c534287a96dc96d1e274665feb7059145d . It looks like the minimum required version would be FFmpeg 2.0. 1.2.4 won't work.

psi29a:
rebased with a more thorough purging of libav

ace13:
I'm using ffmpeg, version 2.5.2

psi29a:
@corristo how difficult/work would it be to bump the ffmpeg up?

corristo:
@psi29a not so hard, I guess. Let's go for it.

scrawl:
I tested this branch with the ffmpeg from OpenMW PPA on Ubuntu 14.04, works great. Valgrind didn't report any memory leaks.

psi29a:
Yay!? :D

zinnschlag:
Go!

PR #803, opened 2015-11-12
Play_NoUnderwater sound flag

scrawl:
@kcat 

Add the Play_NoUnderwater flag for sounds that should not be heard when the listener is underwater. Used for rain, ashstorm and blizzard sounds.
Change PlayMode enum argument to int so we don't have to create a new enum value for each possible combination of the bitflags.

### DIFF
--- a/apps/openmw/mwbase/soundmanager.hpp
+++ b/apps/openmw/mwbase/soundmanager.hpp
@@ -35,14 +35,12 @@ namespace MWBase
                 Play_NoTrack = 1<<2, /* (3D only) Play the sound at the given object's position
                                       * but do not keep it updated (the sound will not move with
                                       * the object and will not stop when the object is deleted. */
-                Play_RemoveAtDistance = 1<<3, /* (3D only) If the listener gets further than 2000 units away
+                Play_RemoveAtDistance = 1<<3 /* (3D only) If the listener gets further than 2000 units away
                                                 from the sound source, the sound is removed.
                                                 This is weird stuff but apparently how vanilla works for sounds
                                                 played by the PlayLoopSound family of script functions. Perhaps we
                                                 can make this cut off a more subtle fade later, but have to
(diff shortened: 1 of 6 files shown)


kcat:
Should those sounds really not be heard when underwater? Seems to me that would create a jarring transition when moving above and below the water line, since all other above-water sounds can still be heard.

scrawl:
It's what I observed in the original game and it seems more realistic to me that way. The ashstorm and blizzard sounds are mostly made up of wind rushing past your ears which isn't going to happen underwater.

The jarring transition is there anyway since we are changing from the underwater pitch adjustment effect to clear sounds.

Regarding rain, it looks like I was wrong, the original game does still play rain sounds underwater. I guess this solves the mystery why the "Rain Loop Sound ID" and "Ambient Loop Sound ID" in the INI files are separate! Will fix in a bit.

scrawl:
On a second (more careful) look, even ashstorm sound is still played underwater. It's just very difficult to hear because the underwater effect they are using almost completely drowns out the high frequencies.

Looking at Morrowind.ini, we have:
<pre>
NearWaterUnderwaterFreq=0.3
NearWaterUnderwaterVolume=0.9
</pre>
This could be what we have as the gain and pitch multipliers for the underwater effect (currently 0.9 and 0.7). I'll do some more testing with that.

scrawl:
More observations:
NearWaterUnderwaterFreq and NearWaterUnderwaterVolume are indeed pitch and volume multipliers like the ones we are already using. Using the values from the INI makes the weather sounds much more subtle when underwater. However: other sounds do not sound very good with this effect. Note that in vanilla, only sounds coming from above water (e.g. weather sounds) have the effect applied. Sounds underwater (e.g. hitting a fish) play like normal.

I think I'll go for this approach. While it's not physically accurate, neither is what we're doing now. Once we get a proper underwater post-effect that sounds better, we could apply it to all sounds. But that is of course a post-1.0 feature.

PR #805, opened 2015-11-13
esm_rewrite

scrawl:
This is PR #678 with additions:
- Solved merge conflicts
- Added some unit tests for MWWorld::Store, all seems to work well comparing the output in this branch against master branch


scrawl:
Bugs we can close:

[Handle deleted cells](https://bugs.openmw.org/issues/720)
[handle DELE subrecord at the end of a record](https://bugs.openmw.org/issues/2211)
[ESM error Unknown subrecord in Grandmaster of Hlaalu](https://bugs.openmw.org/issues/2413)
["Expected subrecord" bugs showing up.](https://bugs.openmw.org/issues/2988)
[ESM Error: Unknown subrecord if Deus Ex Machina mod is loaded](https://bugs.openmw.org/issues/2961)

zinnschlag:
Since there still are some issues, do you want to wait with the merge or merge now and have them fixed later?

scrawl:
What issues do you mean? Deleting cells? I think we should merge now.

scrawl:
@zinnschlag I've just tried deleting a cell in the original CS. The cell is marked as '*D' in the CS, but can still be viewed. And in-game the cell is still rendered as normal. I suppose we can't implement cell deletion in OpenMW as originally planned since that might break compatibility with some mods.

To be honest I'm not sure what the purpose of deleting a cell is in the first place.

Should we reject [bug 720](https://bugs.openmw.org/issues/720) then?

zinnschlag:
The purpose of deleting a cell is to remove the cell from the game. A plugin could modify the world by removing parts of it, e.g. eradicate a part of the landmass.
Also for team based development where content is developed as plugins and then regularly integrated into a new master file the ability to delete cells is obviously useful.

I don't agree with rejecting bug 720. At most we could move it to post-1.0. i.e. content files saved in a format later then the 1.0 one will delete cells properly.

btw. Did you test both on interior and exterior cells? With all the flakiness of original MW I would not be surprised if there is a difference in behaviour.

scrawl:
Yes, I've tried an exterior and an interior cell. Neither works.

PR #807, opened 2015-11-14
Add FFMPEG to include path for OpenMW

ace13:
I'm a bit confused; `mwsound/ffmpeg_decoder.hpp/cpp` requires FFMPEG headers to compile, how did this work in the first place?

### DIFF
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -113,7 +113,10 @@ endif ()
 
 # Sound stuff - here so CMake doesn't stupidly recompile EVERYTHING
 # when we change the backend.
-include_directories(${SOUND_INPUT_INCLUDES})
+include_directories(
+    ${SOUND_INPUT_INCLUDES}
+    ${FFMPEG_INCLUDE_DIRS}
+)
 
 target_link_libraries(openmw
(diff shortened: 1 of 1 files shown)


scrawl:
For me, FFMPEG_INCLUDE_DIRS is simply /usr/include/, which is already in the include path by default I guess.

It looks like SOUND_INPUT_INCLUDES is totally unused right now. How about setting SOUND_INPUT_INCLUDES to FFMPEG_INCLUDE_DIRS [here](https://github.com/OpenMW/openmw/blob/master/CMakeLists.txt#L118), akin to SOUND_INPUT_LIBRARY.

ace13:
I was wondering more how it ever worked on Windows, since there's no such
thing as a common include dir there.
More to the point; wouldn't it make more sense to remove the SOUND_INPUT_*
variables? FFMPEG is pretty baked in by now afterall.

scrawl:
> More to the point; wouldn't it make more sense to remove the SOUND_INPUT_*
variables? FFMPEG is pretty baked in by now afterall.

I'm fine with that too.

ace13:
Maybe put that down as a separate PR though? I would like to have this in
sooner rather than later, so the builds will work

PR #808, opened 2015-11-16
OS X: packaging & Qt5 fixes, disable glTextureStorage2d, drop no longer needed 3rd-party cmake code

corristo:


corristo:
@scrawl fair enough. I'll address it after `openmw-37` is merged into master.

PR #809, opened 2015-11-17
.osg mesh format support

scrawl:
Implemented basic support for the .osg mesh formats. Does not yet support all features that NIF's support, e.g. skeletal animations.

Documentation at: https://wiki.openmw.org/index.php?title=Native_Mesh_Format

### DIFF
--- a/apps/opencs/model/world/resourcesmanager.cpp
+++ b/apps/opencs/model/world/resourcesmanager.cpp
@@ -19,7 +19,9 @@ void CSMWorld::ResourcesManager::setVFS(const VFS::Manager *vfs)
     mVFS = vfs;
     mResources.clear();
 
-    static const char * const sMeshTypes[] = { "nif", 0 };
+    // maybe we could go over the osgDB::Registry to list all supported node formats
+
+    static const char * const sMeshTypes[] = { "nif", "osg", "osgt", "osgb", "osgx", "osg2", 0 };
 
     addResources (Resources (vfs, "meshes", UniversalId::Type_Mesh, sMeshTypes));
     addResources (Resources (vfs, "icons", UniversalId::Type_Icon));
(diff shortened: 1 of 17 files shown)


zinnschlag:
Cool! I vaguely remember that the NIF loader had a significant impact on load times. That was way back though. Not sure if that still has been the case in recent times (both before and after the OSG switch). Do you have any data on possible performance benefits from using the native format instead of NIF?

scrawl:
The Ogre nif loader was slow because it had to be invoked once per instance of an object. Now, we are invoking the nif loader once per object and then instances are made by cloning the loaded scene graph.

I would imagine the osgb format should be faster to load than nif files, but I don't have any data on this. Which reminds me, we'll need a tool to convert NIF files to osg files. Once all features from NIF's are supported, the user could also convert his whole MW installation to osgb files for a (hopefully) faster loading time.

Ultimately though, both loaders can be invoked from background threads (which we are not doing yet), so optimizations might be pointless anyway.

psi29a:
Would we be able to allow hooks/plugins for the new tool?

I was thinking for example the ability to convert s3tc textures to s2tc, astc, uncompressed or whatever that may be embedded in NIF and OSG files.

kcat:
Isn't s3tc the same as s2tc, just that the codec ignores the colors and uses "junk"? If so, there shouldn't be any need to convert such files (and I can't see such a result actually being useful, if the colors are completely messed up). Converting s3tc to some other format like astc or uncompressed, would also require having a proper decoder, which would run afoul of patents.

psi29a:
For people that want to use OpenMW on platforms that do not support s3tc.

The patent expires in about 2 years: http://www.patentcountdown.org/  (unless someone renews it)

The question remains about hooks, which could be used for something like this.

PR #817, opened 2015-11-23
Added comments (and commentary) to the settings-default.cfg file.

cfcohen:
These changes started as comments to myself while experimenting with less documented configuration settings.  My goals were to:

* document the current settings factually (types, defaults, interactions)
* identify temporarily disabled or deprecated settings (e.g. OSG upgrade)
* provide guidance on recommended values and ranges
* explain the purpose of settings that might not be clear to novice users

If desired I can add this same commentary to the wiki instead of or in addition to this file. If this commit is considered "too heavy" for future maintenance,  please feel free to reject it.

### DIFF
--- a/files/settings-default.cfg
+++ b/files/settings-default.cfg
@@ -28,92 +28,6 @@ viewing distance = 6666.0
 # dramatically affect performance, see documentation for details.
 exterior cell load distance = 1
 
-[GUI]
-
-# Color for tool tips and crosshair when owned by an NPC (R G B A).
-color background owned = 0.15 0.0 0.0 1.0
-color crosshair owned = 1.0 0.15 0.15 1.0
-
-# Red flash visually showing player damage.
-hit fader = true
-
-# Transparency of GUI windows (0.0 to 1.0, transparent to opaque).
-menu transparency = 0.84
-
-# Scales GUI window and widget size. (<1 is smaller, >1 is larger).
-scaling factor = 1.0
-
-# Stretch menus, load screens, etc. to the window aspect ratio.
-stretch menu background = false
-
-# Subtitles for NPC spoken dialog and some sound effects.
-subtitles = false
-
-# Time until tool tip appears when hovering over an object (0.0 is
-# instantly, 1.0 is the maximum delay of about 1.5 seconds).
-tooltip delay = 0.0
-
-# Werewolf overlay border around screen or window.
-werewolf overlay = true
-
-[Game]
-
-# Always use the best mode of attack: e.g. chop, slash or thrust.
-best attack = false
-
-# Difficulty.  Expressed as damage dealt and received. (-100 to 100).
-difficulty = 0
-
-# Show duration of magic effect and lights in the spells window.
-show effect duration = false
-
-# Color crosshair and tool tip when object is owned by an NPC. (O is
-# no color, 1 is tool tip only, 2 is crosshair only, and 3 is both).
-show owned = 0
-
-[General]
-
-# Anisotropy reduces distortion in textures at low angles (0 to 16).
-anisotropy = 4
-
-# Camera field of view in degrees (30.0 to 110.0).
-field of view = 55.0
-
-# File format for screenshots.  (jpg, png, tga, and possibly more).
-screenshot format = png
-
-# Isotropic texture filtering.  (bilinear or trilinear).
-texture filtering = trilinear
-
-[HUD]
-
-# Displays the crosshair or reticle when not in GUI mode.
-crosshair = true
-
-[Input]
-
-# Zoom in and out from player in third person view with mouse wheel.
-allow third person zoom = false
-
-# Player is running by default.
-always run = false
-
-# Camera sensitivity when not in GUI mode. (0.1 to 5.0).
-camera sensitivity = 1.0
-
-# Vertical camera sensitivity multiplier when not in GUI mode.
-# Because it's a multiplier values should be near one (0.5 to 1.5).
-camera y multiplier = 1.0
-
-# Capture control of the cursor prevent movement outside the window.
-grab cursor = true
-
-# Invert the vertical axis while not in GUI mode.
-invert y axis = false
-
-# Key controlling sneak toggles setting instead of being held down.
-toggle sneak = false
-
 [Map]
 
 # Size of each exterior cell in pixels in the world map. (12 to 24).
@@ -132,56 +46,102 @@ local map resolution = 256
 # details which may affect cell load performance. (256 to 1024).
 local map widget size = 512
 
-[Objects]
+[GUI]
 
-# Enable shaders for objects other than water. Unused.
-shaders = true
+# Scales GUI window and widget size. (<1 is smaller, >1 is larger).
+scaling factor = 1.0
+
+# Transparency of GUI windows (0.0 to 1.0, transparent to opaque).
+menu transparency = 0.84
+
+# Time until tool tip appears when hovering over an object (0.0 is
+# instantly, 1.0 is the maximum delay of about 1.5 seconds).
+tooltip delay = 0.0
+
+# Stretch menus, load screens, etc. to the window aspect ratio.
+stretch menu background = false
+
+# Subtitles for NPC spoken dialog and some sound effects.
+subtitles = false
+
+# Red flash visually showing player damage.
+hit fader = true
+
+# Werewolf overlay border around screen or window.
+werewolf overlay = true
+
+# Color for tool tips and crosshair when owned by an NPC (R G B A).
+color background owned = 0.15 0.0 0.0 1.0
+color crosshair owned = 1.0 0.15 0.15 1.0
+
+[HUD]
+
+# Displays the crosshair or reticle when not in GUI mode.
+crosshair = true
+
+[Game]
+
+# Color crosshair and tool tip when object is owned by an NPC. (O is
+# no color, 1 is tool tip only, 2 is crosshair only, and 3 is both).
+show owned = 0
+
+# Always use the best mode of attack: e.g. chop, slash or thrust.
+best attack = false
+
+# Difficulty.  Expressed as damage dealt and received. (-100 to 100).
+difficulty = 0
+
+# Show duration of magic effect and lights in the spells window.
+show effect duration = false
+
+[General]
+
+# Anisotropy reduces distortion in textures at low angles (0 to 16).
+anisotropy = 4
+
+# Camera field of view in degrees (30.0 to 110.0).
+field of view = 55.0
+
+# File format for screenshots.  (jpg, png, tga, and possibly more).
+screenshot format = png
+
+# Isotropic texture filtering.  (bilinear or trilinear).
+texture filtering = trilinear
+
+[Input]
+
+# Capture control of the cursor prevent movement outside the window.
+grab cursor = true
+
+# Key controlling sneak toggles setting instead of being held down.
+toggle sneak = false
+
+# Player is running by default.
+always run = false
+
+# Zoom in and out from player in third person view with mouse wheel.
+allow third person zoom = false
+
+# Camera sensitivity when not in GUI mode. (0.1 to 5.0).
+camera sensitivity = 1.0
+
+# Vertical camera sensitivity multiplier when not in GUI mode.
+# Because it's a multiplier values should be near one (0.5 to 1.5).
+camera y multiplier = 1.0
+
+# Invert the vertical axis while not in GUI mode.
+invert y axis = false
 
 [Saves]
-
-# Automatically save the game whenever the player rests.
-autosave = true
 
 # Name of last character played, and default for loading save files.
 character =
 
+# Automatically save the game whenever the player rests.
+autosave = true
+
 # Display the time played on each save file in the load menu.
 timeplayed = false
-
-[Shadows]
-
-# Actors cast shadows.  Unused.
-actor shadows = true
-
-# Debugging of shadows.  Unused.
-debug = false
-
-# Enable shadows. Other shadow settings disabled if false. Unused.
-enabled = false
-
-# Fraction of distance after which shadow starts to fade out.  Unused.
-fade start = 0.8
-
-# Miscellaneous object cast shadows.
-misc shadows = true
-
-# Distance for shadows if not split.  Smaller is poorer. Unused.
-shadow distance = 1300
-
-# Split shadow maps, allowing for a larger shadow distance.  Unused.
-split = false
-
-# Distance for shadows if split.  Unused.
-split shadow distance = 14000
-
-# Static objects cast shadows.  Unused.
-statics shadows = true
-
-# Terrain cast shadows.  Unused.
-terrain shadows = true
-
-# Size of the shadow textures in pixels.  Unused.  (256 to 2048).
-texture size = 1024
 
 [Sound]
 
@@ -190,11 +150,11 @@ device =
 
 # Volumes are 0.0 for silent and 1.0 for the maximum volume.
 
-# Footsteps volume.
-footsteps volume = 0.2
-
 # Master volume.  Controls all other volumes.
 master volume = 1.0
+
+# Footsteps volume.
+footsteps volume = 0.2
 
 # Music tracks volume.
 music volume = 0.5
@@ -205,34 +165,14 @@ sfx volume = 1.0
 # Voice dialog volume.
 voice volume = 0.8
 
-[Terrain]
-
-# Distant land is rendered?  Unused.
-distant land = false
-
-# Use shaders for terrain?  Unused.
-shader = true
-
 [Video]
-
-# Anti-aliasing reduces texture distortion.  (0, 2, 4, 8, 16).
-antialiasing = 0
-
-# Game video contrast.  (0.0 to 1.0).  No effect in Linux.
-contrast = 1.0
-
-# Maximum frames per second.  (1.0 to 200.0).
-framerate limit = 0.0
-
-# OpenMW takes complete control of the screen.
-fullscreen = false
 
 # Resolution of the Open window or screen.  (600 to 2560).
 resolution x = 800
 resolution y = 600
 
-# An operating system border is drawn around the OpenMW window.
-window border = true
+# OpenMW takes complete control of the screen.
+fullscreen = false
 
 # Determines which screen OpenMW is on.  (0 or 1).
 screen = 0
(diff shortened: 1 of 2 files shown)


files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
 
-spells x = 0.625
-spells y = 0.5725
-spells w = 0.375
-spells h = 0.4275
+# Unused?
+companion h = 0.375
+companion w = 0.75
+companion x = 0.25
+companion y = 0
```
scrawl: Used by the companion UI, e.g. [Calvus Horatius](http://www.uesp.net/wiki/Tribunal:Calvus_Horatius) in the Tribunal expansion.

EmperorArthur:
This looks great.  As good as it is to have things in the wiki, It's always nice when the code itself is documented, and people don't have to open up a browser and search for things.  Especially in the, hopefully unlikely, event that the website is down.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
-inventory x = 0
-inventory y = 0.4275
-inventory w = 0.6225
-inventory h = 0.5725
+# Minimize the OpenMW window if it loses cursor focus.  This setting
+# has no effect if the "fullscreen" setting is false.  This setting is
+# primarily useful for single screen configurations, so that the
+# OpenMW screen in full screen mode can be minimized when the
+# operating system regains control of the mouse and keyboard.  On
+# multiple screen configurations, disabling this option make make it
```
scrawl: typo 'make make'

scrawl:
This is great documentation. I am a bit torn if it should be on the wiki or in the settings file.

Wiki:
* Easier on the eyes to read due to non-monospaced font.
* Can add different instructions for different versions of OpenMW. The settings-default.cfg in the repo on the other hand shouldn't contain old settings that were renamed or are no longer in use.
* Can add special formatting, navigation side bar, tables, sortable columns, etc.

Repo:
* It's less likely that we forget to update the documentation after a code change.

Opinions?

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+near clip = 5.0
+
+# This boolean setting determines whether objects that render to one
+# pixel or smaller will be culled.  It generally improves performance
+# to enable this feature.
+small feature culling = true
 
-vsync = false
+# Set the maximum visible distance.  Larger values significantly
+# improve rendering in exterior spaces, but also increase the amount
```
scrawl: amount of

kcat:
Why not both? The wiki could be for public releases, while the code/repo is (generally) always up to date with git head.

scrawl:
Having both would be duplicating information thus more work to maintain.

cfcohen:
I think that the ultimate authoritative source is the settings file itself, and that's why I put my comments there.   I really wanted to put them in my user settings.cfg, but the config file rewrite process discards comments in that file. :-(   I think I'll open a feature request to preserve comments during rewrite if no one objects.

As for the wiki, I volunteered in part for the reasons that Scrawl suggested, and in part because I think users will be more likely to stumble across it there.  The only real downside to keeping in in both that I can see if the additional maintenance effort required.  But I can at least get both nicely up-to-date once...  I'm on Thanksgiving vacation this week and decided to try and contribute back some effort.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# than you should be able to.  This can be observed in game by looking
+# at distant objects and rotating the camera so the object are near
+# the edge of the screen.  As a result, the "viewing distance" setting
+# should further be reduced by a factor that depends on the "field of
+# view" setting.  In the default configuration this reduction is 7%.
+# Using this factor, approximate values recommended for other
+# "exterior cell load distance" settings are: 14285 for 2 cells, 21903
+# for 3 cells, 29522 for 4 cells, and 35924 for 5 cells.
+# 
+# Reductions of up 25% or more can be required to completely eliminate
```
scrawl: up to

cfcohen:
Additional reasons to update the wiki:

* Hyperlinks to relevant wikipedia articles on graphics technologies (which I kind of wanted to include in the config file, but didn't fell like it was appropriate)
* Level headings can be used to group options by section easing navigation
* Links to code, tickets, etc.

kcat:
I'd still like them in the file itself, personally. That way I can be sure that the documentation in the file is in accordance with the version, since the wiki may be newer or older than the code I (or someone else) has.

cfcohen:
Kcat, if the decision is that only one place should be updated due to maintenance concerns, I agree that it should be the config file itself.  I'm arguing that it's worth trying to do both. :-)   But I'll probably be distracted again (and not contributing regularly) once they've get out of sync, so that's not really my call.

scrawl:
How about a hybrid approach - the settings file has a basic explanation of what a setting affects, the wiki has detailed use scenarios, recommended values, explanations of what graphics techniques are, and all that stuff.

For a coder / contributor,
<pre>
# Maximum framerate in frames per second, 0 = unlimited
framerate limit = 0
</pre>
is most likely enough to know.

This version on the other hand would be aimed more at the average user:
<pre>
+# This floating point setting determines the maximum frame rate in
+# frames per second.  If this setting is 0.0, the frame rate is
+# unlimited.  There are several reasons to consider capping your frame
+# rate, especially if you're already experiencing a relatively high
+# frame rate (greater than 60 frames per second).  Lower frame rates
+# will consume less power and generate less heat and noise.  Frame
+# rates above 60 frames per second rarely produce perceptible
+# improvements in visual quality.  Capping the frame rate may in some
+# situations reduce the perception of choppiness (highly variable
+# frame rates during game play) by lowering the peak frame rates.
+# This setting interacts with the "vsync" setting in the Video section
+# in the sense that enabling vertical sync limits the frame rate to
+# the refresh rate of your monitor (often 60 frames per second).
+framerate limit = 0.0
</pre>

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+#   This path may vary depending on your installation hard drive, your
+#   Windows username, and your default language.
+#
+# Additionally, the user settings file is often written to disk when
+# exiting OpenMW, so comments and changes to that file may also be
+# discarded after running OpenMW.  While most changes to the file will
+# reflect setting changes made in game, some settings can have a wider
+# range of values in the settings file than the GUI settings widgets
+# allow.  You may want to exercise some caution and backup this file
+# when editing it by hand.
```
scrawl: Agree with the caution part. Maybe we can add another sentence: "Certain values may cause the game to malfunction or crash".

cfcohen:
The hybrid approach seems fine to me.  The wiki format will be better suited to educating casual users which was one of my goals.   I figured if I couldn't figure it out without reading the source code then no one else was likely to either.   I was a little concerned that this commit would be "too wordy" for the config file.

I propose that I update the wiki and then submit a new pull request for a lighter weight update to the settings.cfg file tomorrow.   The settings file should definitely reference the wiki then, and every setting should have at least a one or two line description in th settings.cfg file hitting the most important point.

I see you've found a couple of typos and explained the "companion" interface that I couldn't find.  I haven't actually played through Tribunal yet! Are there any other glaring inaccuracies?  Obviously I was guessing in a couple of places.

Finally, I discovered some "bugs" while experimenting with these values.   In several places the code does not validate that the settings are properly bounded, but instead assumes that the settings file was generated by the Options GUI.  I was going to try and fix a few of these myself, and then open bugs for the rest.   Guidance on this is welcomed.

scrawl:
> I propose that I update the wiki and then submit a new pull request for a lighter weight update to the settings.cfg file tomorrow. The settings file should definitely reference the wiki then, and every setting should have at least a one or two line description in th settings.cfg file hitting the most important point.

Sounds great.

> In several places the code does not validate that the settings are properly bounded, but instead assumes that the settings file was generated by the Options GUI. I was going to try and fix a few of these myself, and then open bugs for the rest. Guidance on this is welcomed.

I would evaluate this on a case by case basis. We can bound settings if incorrect values there would cause straight crashes or malfunctions. But at the same time we should avoid arbitrary limitations on how high (or low) a setting can go.

scrawl:
In addition we may want to log warning messages if an out of bounds setting has been read.

files/settings-default.cfg
cfcohen: I'd argue that crashing due to config values shouldn't be possible.   Exactly how to determine what's a "valid" value is still an open question.  At one point I considered the limits in the Options GUI to be authoritative, but there are several settings where that's seriously counter-productive (e.g. viewing distance).  In the end, I think values that crash or produce ridiculous results should be disallowed.  I'm still investigating what those values are for some settings.   See my recent comment on "bugs".

cfcohen:
I agree on limitations and warnings and will attempt some reasonable proposals in that vein.  If I propose code to limit a value I will provide a justification for why it could never be a "useful" value.  Then we can debate whether I'm actually correct or not for that particular setting and limit.

files/settings-default.cfg
scrawl: > In the end, I think values that crash or produce ridiculous results should be disallowed.

Some settings are so complicated with respect to their interactions that it's hard to verify if a value will or will not crash. In particular with cell load distances, resolutions, render texture sizes, and so on, which depend upon the user's system ram, graphics card RAM, loaded mods, etc.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
 
+# Enable or disable subtitles for NPC spoken dialog (and some sound
+# effects).  Subtitles will appear in a tool tip box in the lower
+# center of the screen.  The setting can be toggled in game with the
+# "Subtitles" button in the Prefs panel of Options menu.
 subtitles = false
 
-hit fader = true
+# Set the delay between when you begin hovering over an item and when
+# it's tooltip appears.  This setting is a floating point value
```
scrawl: it's -> its

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# between 0.0, which displays the tool tip instantly and 1.0 which
+# results in the maximum delay (approximately 1.5 seconds).  This
+# setting does not affect the tooltip delay for object under the
```
scrawl: objects

cfcohen:
Where should I ask questions about existing settings?   In a forum thread?  On the wiki?  In this commit I danced around a couple of issues that might be worthy of a discussion.  For example:

* Is there a reason that the map load radius is hard-coded at 3x3 versus using viewing distance?
* Is the difference in grab cursor behavior after pressing escape intentional?   It's useful enough that I'm not sure I'd want to change it, but it's a little inconsistent.
* Is there a way to re-enable shadows for alpha testers?
* Are the options marked unused really unused (currently)?
* I think there's a bug in the way that always run is updated in the config file.

And so on...

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# crosshair in the "look mode", only widgets in the GUI windows.  This
+# setting can be adjusted in game with the "Menu Help Delay" slider in
+# the Prefs panel of the Options menu.
+tooltip delay = 0.0
+
+# Enable or disable the werewolf overlay.  Unable to evaluate fully
+# due to issues with becoming a werewolf.
```
scrawl: You can do ```player->becomewerewolf``` in the console to become a werewolf, but generally this feature won't work in  the Bloodmoon expansion is not installed.

The setting controls a full screen texture effect that is displayed only in first person mode for werewolf players.

cfcohen:
I had multiple problems with the werewolf overlay.   I swear I have a pretty standard setup, but it couldn't find the overlay in the BSA.   I had to extract it by hand and put it in the texture directory.   On top of that I was stuck in the middle of the floor and couldn't move while in werewolf form.   When the overlay is missing, the default emergency texture (which is a non-transparent full screen texture) isn't particularly helpful. :-) ... and so on...

It was going to be a whole separate issue to report on that once I figured out what was going on.

scrawl:
Sounds like the BSA file hasn't been added to openmw.cfg. You need a
<pre>
fallback-archive=Bloodmoon.bsa
</pre>

(and the same for Tribunal, if that's also installed)

cfcohen:
You nailed that one.  I had a fallback-archive line for Tribunal, but not for Bloodmoon.   That probably explains some other problems I've been having as well.   I recently realized that I hadn't run the "Wizard" properly in a long time, since I've had my game-data directory since OpenMW 0.2x or earlier.

Should we move this conversation to the forums?

scrawl:
Let's move questions about settings to the forums.

> Is the difference in grab cursor behavior after pressing escape intentional? 

Yes. The reasoning is that you'll usually pause the game with Escape when you want to go back to the desktop to do something, so it's desirable to unlock the cursor.

> Is there a reason that the map load radius is hard-coded at 3x3 versus using viewing distance?

Using cell load radius for the map load radius might be inadvisable, because rendering a lot of maps further increases loading delays when switching cells. I think we should make the map loading radius a separate setting.

> Is there a way to re-enable shadows for alpha testers?

Speaking of when shadows will be reimplemented with OSG? Haven't thought about that in much detail yet. The new implementation might be totally different from the old shadows. But generally we'll want some way for semi-transparent meshes to cast shadows, yes.

> Are the options marked unused really unused (currently)?

I haven't even read the whole file yet (it's long) but I will keep reading and tell you if something is incorrectly marked.

> I think there's a bug in the way that always run is updated in the config file.

In what way? Code looks ok to me.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
 werewolf overlay = true
 
-stretch menu background = false
+[Game]
 
-# colour definitions (red green blue alpha)
-color background owned = 0.15 0 0 1
-color crosshair owned = 1 0.15 0.15 1
+# If this boolean setting is true, the character will always use the
```
scrawl: write "player character" just to avoid confusion.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# to be held while sneaking.  Players that spend significant time
+# sneaking may find the character easier to control with this option
+# enabled.
+toggle sneak = false
+
+# This setting continues to be loaded and saved, but has no known
+# effect.  Presumably it and a related but also removed option named
+# "ui y sensitivity" used to control mouse sensitivity while in GUI
+# mode.  The default value is 1.0.
+ui sensitivity = 1.0
```
scrawl: This setting is no longer used because we're using hardware cursor movement. I must have forgotten to remove it when hardware cursors were implemented.

With hardware cursors, the only way to control the cursor's sensitivity is via the operating system's mouse settings. Having the game access and change the operating system settings would be a) complicated to implement and b) potentially inadvisable since we'd be left with the application specified settings if the application crashed.

Removed in master.

EmperorArthur:
If you have information in the wiki that's not in the file the link to the wiki needs to be extremely obvious.

Also, if yo do smaller comments, I'd really appreciate it if each had the minimum and maximum values in there.

cfcohen:
This conversation has been moved to: http://forum.openmw.org/viewtopic.php?f=6&t=3195

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# mode map display. The value is the width in pixels of each cell in
+# the map, so larger values result in larger more detailed world maps,
+# while smaller values result in smaller less detailed world maps.
+# However, the native resolution of the map source material appears to
+# be 9 pixels per unexplored cell and approximately 18 pixels per
+# explored cell, so values larger than 36 don't produce much
+# additional detail.  Similarly, the size of place markers is
+# currently fixed at 12 pixels, so values smaller than this result in
+# overlapping place markers.  Values from 12 to 36 are recommended.
+# For reference, Vvardenfell is approximately 41x36 cells.
```
scrawl: This setting needs to be marked with a warning that it affects save files. The currently explored global map overlay is stored in the save file as an image. When you increase the map size, the overlay of pre-existing save games will be up-scaled on load (thus become blurry). Note that when you visit a cell again the overlay for that cell gets re-generated, so you can "fix" the problem by revisiting all cells you've already visited.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# This integer setting controls the resolution of the GUI mode local
+# map widget.  Larger values generally increase the visible detail in
+# map.  If this setting is half the "local map widget size" or
+# smaller, the map will generally be be fairly blurry.  Setting the
+# both options to the same value results in a map with good detail.
+# Values that exceed the "local map widget size" setting by more than
+# a factor of two are unlikely to provide much of an improvement in
+# detail since they're subsequently scaled back to the approximately
+# the map widget size before display.  The video resolution setting
+# interacts with this setting in that regard.
```
scrawl: Please note that larger map resolutions result in longer loading times, because the map is rendered on-demand when you enter a new cell. Also note that graphics cards have resolution limits, so the game may crash for abnormally large sizes or run out of VRAM.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
-[Cells]
-exterior cell load distance = 1
+[Objects]
 
-[Camera]
-near clip = 5
+# This boolean setting currently has no known impact, but is
+# presumably intended to enable shaders for objects other than water.
+# Whenever the setting file is written by the game, this option is
+# currently reset to false.
```
scrawl: > This boolean setting currently has no known impact, but is presumably intended to enable shaders for objects other than water.

Correct.

> Whenever the setting file is written by the game, this option is currently reset to false.

Can not confirm that.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
-distant land = false
+# This boolean setting determines whether the game will be
+# automatically saved when the character rests.  This setting can be
+# toggled in game with the "Auto-Save when Rest" button in the Prefs
+# panel of the Options menu.
+autosave = true
 
-shader = true
+# This boolean setting determines whether the amount of the time the
+# player has spent playing will displayed for each saved game in the
```
scrawl: will be

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# detailed shadows and a better visual effect.
+texture size = 1024
 
 [Sound]
-# Device name. Blank means default
+
+# This string setting determines which audio device to use. A blank or
+# missing setting means to use the default device, which should
+# usually be sufficient, but if you need to explicitly specify a
+# device name try doing so here.
```
scrawl: The device names you can use will be listed in the openmw.log file.

cfcohen:
Good points.  You alluded to the first issue somewhere else (in the existing wiki page maybe?), but I didn't really understand until now.   I'll try to document that more fully.  I had noticed the longer load times for the map resolution, but it didn't make the cut for the first draft of the settings.cfg file.  I'll cover that more fully in the Wiki.   I'll followup with some more questions about these tomorrow, but for now the wife is requesting that I go have a real-world life...

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
 
-inventory barter x = 0
-inventory barter y = 0.4275
-inventory barter w = 0.6225
-inventory barter h = 0.5725
+# This floating point setting controls the gamma correction for all
+# video in the game.  This setting does not currently work under
+# Linux, and the in-game setting in the Options menu has been
+# disabled.
+gamma = 1.00
```
scrawl: Would it make sense to put this next to the 'contrast' option as it used to? These two options are kind of related.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+refraction = false
 
-companion x = 0.25
-companion y = 0
-companion w = 0.75
-companion h = 0.375
+# Refracted texture size.  In the Video panel of the options menu, the
+# choices are Low (512), Medium (1024) and High (2048).  This setting
+# determines the resolution of the textures used for rendering objects
+# on the other wide of the plane of water (which have a wavy
```
scrawl: other side I presume.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# appearance caused the by the refraction).  Higher values produces
+# better visuals and result in a marginally lower framerate depending
+# on your graphics hardware.  The "refraction" setting in the "Water"
+# section must be enabled for this setting to have any effect.
```
scrawl: 'rtt size' also controls the size of the reflection texture, so "The "refraction" setting in the "Water"
+# section must be enabled for this setting to have any effect." is not correct.

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+# The character inventory window while bartering.  It displays goods
+# owned by the character while bartering.  Activated by clicking on the
+# "Barter" choice in the dialog window for an NPC.
+inventory barter h = 0.5725
+inventory barter w = 0.6225
+inventory barter x = 0
+inventory barter y = 0.4275
+
+# Unused?
+inventory companion h = 0.5725
```
scrawl: See "companion"

files/settings-default.cfg
```diff
@@ -1,244 +1,709 @@
...
+spells x = 0.625
+spells y = 0.5725
+
+# The stats window, displaying level, race, class, skills and stats.
+# Activated by clicking on any of the three bars in the lower left
+# corner of the HUD.
+stats h = 0.4275
+stats w = 0.375
+stats x = 0
+stats y = 0
```
scrawl: I made it all the way down! Woooooohooo!

cfcohen:
I've added comments on the forum thread about the recent changes, and why I made them.   I'm still planning on updating the Wiki, but I'd like to keep the settings in a consistent order, and obviously this commit relates to that.

scrawl:
Could you please separate the settings-default.cfg changes and the config writer changes into separate PRs. I would like to merge the new documentation right away but haven't fully reviewed the config writer changes yet (general idea seems good to me though).

cfcohen:
Sure, but I'm not sure I know how to do that. :-(  Can you coach me a bit?

cfcohen:
Nope.  That didn't do what you wanted. :-(  I'm guessing that I need to revert to the master head and make a new branch and then commit the file on that branch... or something like that.  Sorry, I'm not up to speed on branching in git. :-(

scrawl:
Check out commit d7e27fa9f443bcc1aa6394314456ad457ef7cf70 "New brief comments version of settings.cfg." and create a new branch from it.
<pre>
git checkout d7e27fa9f443bcc1aa6394314456ad457ef7cf70
git checkout -b settings_documentation
git push origin settings_documentation
</pre>

Next, create a new branch containing the config changes, based on the latest master (commit da856eed9512aa202f93b955552f6f7d8ca7ecd2).
<pre>
git fetch https://github.com/OpenMW/openmw.git 
git checkout da856eed9512aa202f93b955552f6f7d8ca7ecd2
git checkout -b config_writer_changes

# cherry pick the config writer commits from the master branch to our new branch
git cherry-pick 8e0497dcb0b1561db02b771c8ebe4a22098d208c
git cherry-pick 9ec5bd55dd834be9d1346e7f59b4dde166a14ae8

git push origin config_writer_changes
</pre>

Now you can open PRs with github UI for the new branches.

cfcohen:
Ok. I think I can do that.  Trying now.  Good learning experience for me.

scrawl:
Unfortunately commit https://github.com/cfcohen/openmw/commit/8e0497dcb0b1561db02b771c8ebe4a22098d208c touches both config writer code and the settings file. That is a bit of a mess.You'll need to manually apply those settings file changes to a new commit in the settings_documentation branch, then as for the config writer branch just apply the changes manually (ignoring the settings file change) and create a commit instead of cherry-picking the existing commits. That should fix the issue.

cfcohen:
I think I understand that as well.

cfcohen:
Should I close this pull request now?

scrawl:
#819, #820

PR #818, opened 2015-11-23
opengl es

sandstranger:
This PR not for merge. It is for storage.
PS : And I stuck on opengl es lighting in water fragment shader.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,19 @@ include(OpenMWMacros)
 
 if (ANDROID)
     set(CMAKE_FIND_ROOT_PATH ${OPENMW_DEPENDENCIES_DIR} "${CMAKE_FIND_ROOT_PATH}")
+    set(OSG_PLUGINS_DIR ${OSG_PLUGINS_DIR})
+    add_definitions (-DOSG_PLUGINS_DIR)
+    set(OPENGLES TRUE CACHE BOOL "enable opengl es support for android" FORCE)
 endif (ANDROID)
+
+option(OPENGLES "enable opengl es support" FALSE )
+
+if (OPENGLES)
+    INCLUDE(cmake/FindOpenGLES.cmake)
+    INCLUDE(cmake/FindOpenGLES2.cmake)
+    add_definitions (-DOPENGLES)
+    INCLUDE_DIRECTORIES(${OPENGLES_INCLUDE_DIR})
+endif (OPENGLES)
 
 # doxygen main page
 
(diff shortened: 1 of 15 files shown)


EmperorArthur:
Quick question.  Have you tried using the Opengl es code on Windows, Mac, Linux?  Does it compile and run on these platforms?  If it does, we could try modifying the shaders so there's only one code path.

psi29a:
Nice work here! :)

sandstranger:
I have not tried to run it on linux yet , because now I compile OSG  and sd2 with gles2  for Linux

psi29a:
Debian is a Linux distribution, just a fyi.

Android is also Linux, per definition. I think what @EmperorArthur meant is does the gles2 version run on other platforms aside from Android.

Normally yes, it should work provided the drivers of that platform support OpenGL ES2.

sandstranger:
in theory it should work on linux , but I want to check it

psi29a:
Linux is a kernel, Android uses Linux. Android is just another distribution, like Debian, Ubuntu or Redhat as they too use Linux.

scrawl:
> This PR not for merge

scrawl:
I didn't comment on this PR because you wrote it's not merge-ready, but now that it was accidentally merged I should explain why I reverted it.

```
#ifdef OPENGLES
+#include <GLES/gl.h>
+#endif
```
Shouldn't OSG handle the including of GL headers for us? That's how it works for desktop GL, and since you only modified a couple of files not everything using GL_ enums I have to assume the includes generally work.

Duplication of water shader: no way. But we can add conditional compiling (i.e. #if OPENGL_ES checks in the shader, then insert the correct define in water.cpp based on what OSG has been compiled with).

> PS : And I stuck on opengl es lighting in water fragment shader.

Where did you get osg_LightSourceParameters? I can't find a reference to that in the OSG source code.
Generally we'll need to apply light states in a Uniform since OpenGL ES2 does not support glLight. It's a bit too early to think about all this since OpenMW doesn't have a full shader pipeline yet.

sandstranger:
I get osg_Light here https://github.com/gwaldron/osgearth/blob/master/src/osgEarth/ShaderUtils.cpp

sandstranger:
And How I need use osg gl headers ? Without gles header I have got ton of errors in building
/home/sandstranger/Android/openmw/components/myguiplatform/myguirendermanager.cpp: In member function 'virtual void osgMyGUI::Drawable::drawImplementation(osg::RenderInfo&) const':
/home/sandstranger/Android/openmw/components/myguiplatform/myguirendermanager.cpp:117:29: error: 'GL_VERTEX_ARRAY' was not declared in this scope
         glEnableClientState(GL_VERTEX_ARRAY);
                             ^
/home/sandstranger/Android/openmw/components/myguiplatform/myguirendermanager.cpp:117:44: error: 'glEnableClientState' was not declared in this scope
         glEnableClientState(GL_VERTEX_ARRAY);

scrawl:
> Update : Osg not included GLES/gl.h header in osg/GL , because osg compiled for gles2.
There is no methods like glEnableClientState in Android ndk gles2 header .
It is available only on gles header . And I believe, it is impossible to use gles and gles2 together , because it is different API . I will try to build openmw with opengl es 1.1 render , instead using 2.0 , but I suppose it will not work . But if openmw and osg will be successfully build with gles1 , graphics will be rendered without shaders, as I understand

That sounds correct to me.

sandstranger:
It is building and starting with gles1 , but than crashes with
 2-02 09:37:49.373 3751-4067/com.libopenmw.openmw D/MALI: gles_state_set_error_internal:57: GLES error info:<pname> is not an accepted value .
I will look at this error later , but I suppose this error connected with SDL2

sandstranger:
it seems it will not work on GLES1 . I have got ton erros in log about rendering .

psi29a:
That's great news! :)   Thank you for your work!

sandstranger:
But I still have got issue with textures .I suppose I unpack all textures wrong

zinnschlag:
Sorry for the accidental merge. The don't merge part escaped me. Bit of a mixup. Would be nice to have a label for that. Anyway, do we want to remove these commits or can they stay?

psi29a:
@zinnschlag we can create a label for that. ;)    "DO NOT MERGE" or something in a big ugly colour.

zinnschlag:
I think we can do without the yelling. But yeah, this looks useful; in particular since it also shows up in the PR list.

PR #819, opened 2015-11-24
Settings documentation

cfcohen:
So this is just the changes to the documentation.  Including the reordering and the silly commit. The code changes are not strictly required for the reordering, but without the code changes, the files will just get sorted again anyway.  Hopefully this is what you wanted.  I think I know how to fix the code branch as well.  Should I "close" the pull request on master?

### DIFF
--- a/files/settings-default.cfg
+++ b/files/settings-default.cfg
@@ -8,7 +8,7 @@
 #
 # The version of this file that actually controls the behavior of
 # OpenMW is user specific, and it's location can be found in the
-# documentation above.  This file is probably NOT that file.
+# documentation above.  This file is probably NOT that file...
 
 [Camera]
 
@@ -28,92 +28,6 @@ viewing distance = 6666.0
 # dramatically affect performance, see documentation for details.
 exterior cell load distance = 1
 
-[GUI]
-
-# Color for tool tips and crosshair when owned by an NPC (R G B A).
-color background owned = 0.15 0.0 0.0 1.0
-color crosshair owned = 1.0 0.15 0.15 1.0
-
-# Red flash visually showing player damage.
-hit fader = true
-
-# Transparency of GUI windows (0.0 to 1.0, transparent to opaque).
-menu transparency = 0.84
-
-# Scales GUI window and widget size. (<1 is smaller, >1 is larger).
-scaling factor = 1.0
-
-# Stretch menus, load screens, etc. to the window aspect ratio.
-stretch menu background = false
-
-# Subtitles for NPC spoken dialog and some sound effects.
-subtitles = false
-
-# Time until tool tip appears when hovering over an object (0.0 is
-# instantly, 1.0 is the maximum delay of about 1.5 seconds).
-tooltip delay = 0.0
-
-# Werewolf overlay border around screen or window.
-werewolf overlay = true
-
-[Game]
-
-# Always use the best mode of attack: e.g. chop, slash or thrust.
-best attack = false
-
-# Difficulty.  Expressed as damage dealt and received. (-100 to 100).
-difficulty = 0
-
-# Show duration of magic effect and lights in the spells window.
-show effect duration = false
-
-# Color crosshair and tool tip when object is owned by an NPC. (O is
-# no color, 1 is tool tip only, 2 is crosshair only, and 3 is both).
-show owned = 0
-
-[General]
-
-# Anisotropy reduces distortion in textures at low angles (0 to 16).
-anisotropy = 4
-
-# Camera field of view in degrees (30.0 to 110.0).
-field of view = 55.0
-
-# File format for screenshots.  (jpg, png, tga, and possibly more).
-screenshot format = png
-
-# Isotropic texture filtering.  (bilinear or trilinear).
-texture filtering = trilinear
-
-[HUD]
-
-# Displays the crosshair or reticle when not in GUI mode.
-crosshair = true
-
-[Input]
-
-# Zoom in and out from player in third person view with mouse wheel.
-allow third person zoom = false
-
-# Player is running by default.
-always run = false
-
-# Camera sensitivity when not in GUI mode. (0.1 to 5.0).
-camera sensitivity = 1.0
-
-# Vertical camera sensitivity multiplier when not in GUI mode.
-# Because it's a multiplier values should be near one (0.5 to 1.5).
-camera y multiplier = 1.0
-
-# Capture control of the cursor prevent movement outside the window.
-grab cursor = true
-
-# Invert the vertical axis while not in GUI mode.
-invert y axis = false
-
-# Key controlling sneak toggles setting instead of being held down.
-toggle sneak = false
-
 [Map]
 
 # Size of each exterior cell in pixels in the world map. (12 to 24).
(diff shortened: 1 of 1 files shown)


scrawl:
From the comments, It's not immediately clear which values are just recommendations and which are hard limits. Maybe write  ``` (e.g. x to y)``` for recommended values and ```(x to y)``` for 'real' limits.

> +# Determines which screen OpenMW is on.  (0 or 1).

What about triple and quadruple monitor setups?

Some settings are way too system specific to give any recommendations for (e.g. resolutions).

scrawl:
This PR removes the warning at the top of the file
> -# WARNING: Editing this file might have no effect, as these
-# settings are overwritten by your user settings file.

and replaces it with 

> +# The version of this file that actually controls the behavior of
+# OpenMW is user specific, and it's location can be found in the
+# documentation above.  This file is probably NOT that file...

further down the file. I don't think this new warning is clear enough. Also, the warning should be at the top of the file, because that's the first thing you read.

cfcohen:
I've made the requested changes.  There was a request to provide "ranges of values" for settings, and while hard limits are more factually correct, in many cases they'll be less useful in my opinion.  The ranges specified in the brief description are intended to be subjective assessments of the most common values.   I commented on this more extensively in a forum post.   But as I said there, I'm not attached to any particular values.

On the warning issue, I had intentionally changed the wording because previously those comments were only in the settings-default.cfg file, and that file almost never has any influence as far as I can tell, so the wording was almost too weak there.  On the other hand, I believe that the old wording is far too strong if this text appears at the top of the correct file (as it does in my configuration right now).  Additionally, as we discuss the settings.cfg code change, I suspect that either this comment or something similar is very likely to end up in the correct file, in order to avoid maintaining yet more text.

scrawl:
Still not happy with this type of wording:

<pre>
+# Zoom level in pixels for HUD map widget.  64 is one cell, 128 is 1/4
+# cell, 256 is 1/8 cell.  See documentation for details. (64 to 256).
</pre>
It suggests that only values from 64 to 256 are legal. Why not
<pre>
+# Zoom level in pixels for HUD map widget.  64 is one cell, 128 is 1/4
+# cell, 256 is 1/8 cell.  See documentation for details. (e.g. 64 to 256).
</pre>

The same applies to almost all the other ranges suggested in this file. They're just suggestions but the wording sounds like the ranges are hard limits.

<pre>
+# Size of local map in GUI window in pixels.  See documentation for
+# details which may affect cell load performance. (256 to 1024).
+local map widget size = 512
</pre>
This is incorrect, local map widget size does not affect performance. Only the resolution does.

> +# Maximum frames per second.  (1.0 to 200.0).
> +framerate limit = 0.0

Suggesting a range here is counterproductive because the FPS you get depends too much on your system. And why start at 1? Who wants to play with 1 FPS? Why not 0.0000001? Why not 5? Why not 30? Who is this suggestion targeted at? I think you get the idea that I don't like arbitrary values. Every user knows what a framerate is and what sort of framerate they're getting. Furthermore, this new comment no longer mentions the most important aspect to know that "0" means unlimited framerate.

> +# Video gamma setting. (0.0 to 1.0).  No effect in Linux.
> +gamma = 1.0

Definitely not (0.0 to 1.0). Gamma is an exponent that makes colors brighter if > 1.0 and darker if < 1.0. The range suggested by the settings window is: 

```
    <UserString key="SettingMin" value="0.1"/>
    <UserString key="SettingMax" value="3.0"/>
```
> On the other hand, I believe that the old wording is far too strong if this text appears at the top of the correct file (as it does in my configuration right now). 

Well, it shouldn't. How did it get there? You pasted the whole file manually? That's kind of your own fault then ;)

I guess we can clarify
<pre>
Editing this file
</pre>
to
<pre>
Editing 'settings-default.cfg'
</pre>

cfcohen:
Scrawl wrote:

> The same applies to almost all the other ranges suggested in this file. They're just suggestions but the wording sounds like the ranges are hard limits.

Exactly.  That was a conscious choice as I mentioned before. ;-) If you're saying you want them to be absolute limits, I'll do that.   If you're just asking me to change specific recommendations, I'll do that as well.  It's not a big deal, and you get to make the decision.  But if you want me to set the limits, we'll need to establish how I decide what the limits are.   It's clearly not what the GUI sliders use.  I suppose I could use what crashes and what doesn't. :-)  To be honest they all seem subjective at some level to me, so I just embraced that and went with it.  For example, is setting "local map hud widget size" to a value smaller than 21 an absolute limit just because there's no map data to display?  I've not found much if any code that enforces any limits.  It's possible that you have a clearer idea what you want because you have a clearer idea of what supposed to happen.  I don't -- that's why I was investigating.

Thanks for the catch on the gamma setting.  That was one of the later settings I investigated, and when it didn't work at all for me, I made a comment to that effect and moved on.  I'll improve that, and I'll review the XML for the ranges.  I'd been using an experimental approach. :-(

The wording change to include the filename 'settings-default.cfg' would certainly make me happier.  I'll include that in the next commit for you to review.

> Well, it shouldn't. How did it get there? You pasted the whole file manually? That's kind of your own fault then ;)

Actually, what I did was from my forum post "Stupid configuration tricks with the new code..." where I wrote:

> Start OpenMW, and replace the user settings.cfg with the settings-default.cfg while running. On exit, your settings should be saved, but in the new order, and with the default comments.

This is the easiest way that I currently know of to go from a completely comment free configuration file to one with some explanation of what each setting does without losing all of your existing settings.  I perceive that a number of people might find that useful.

Cory

files/settings-default.cfg
```diff
@@ -1,242 +1,342 @@
-# WARNING: Editing this file might have no effect, as these
-# settings are overwritten by your user settings file.
+# WARNING: If this file is named settings-default.cfg, then editing
+# this file might have no effect, as these settings may be overwritten
+# by your user settings.cfg file (see docmentation for its location).
```
scrawl: typo (docmentation)

files/settings-default.cfg
```diff
@@ -1,242 +1,342 @@
...
+
+# Sound effects volume.
+sfx volume = 1.0
+
+# Voice dialog volume.
+voice volume = 0.8
+
+[Video]
+
+# Resolution of the Open window or screen.
```
scrawl: "OpenMW window or screen", I presume.

scrawl:
> Exactly. That was a conscious choice as I mentioned before. ;-) If you're saying you want them to be absolute limits, I'll do that.

That's not what I mean. I'm fine with having "recommended" ranges in the settings file, but it should be obvious that it's just a recommendation and not the hard limit.

For example, this here sounds like the lowest possible value for "near clipping plane" to go is 0.01, but all values > 0 are perfectly valid.

> +# Near clipping plane (0.01 to 18.0).

I suggest a different wording:

> +# Near clipping plane (e.g. 0.01 to 18.0).

Or even better

> +# Near clipping plane (> 0, e.g. 0.01 to 18.0).

The same concern applies to all the other (x to y) recommendations in the file.

This issue was not addressed yet:

> > +# Size of local map in GUI window in pixels.  See documentation for
> > +# details which may affect cell load performance. (256 to 1024).
> > +local map widget size = 512

> This is incorrect, local map widget size does not affect performance. Only the resolution does.

cfcohen:
I understand now.  This commit should be satisfactory (I hope).

> This issue was not addressed yet:

Oops.  I corrected it on the wiki and thought that I had corrected it in the settings-default.cfg as well.   Obviously I had not.  Good catch.

scrawl:
Thank you very much for spending time on this. I really appreciate having more documentation because I suck at writing it myself. It's not an exciting thing to do but needs to be done none the less.

psi29a:
Great work Cory!

cfcohen:
psi29a: Thanks.

scrawl:  As you can probably tell, I'm not prepared to be making major commits to the OpenSceneGraph code. :-)   But I have been trying to contribute something back to the project in the form of documentation, bug reports, minor code contributions, etc.  I'm very interested in using the OpenMW code to build open source games in the Morrowind genre.   Unfortunately, it's hard for me to stay focused for very long, so I have to pick small tasks that I can complete before I get distracted.  Hopefully I don't cost more than I contribute. :)

PR #820, opened 2015-11-24
Config writer changes

cfcohen:
Ok. So I think I got this one correct as well.  This is supposed to just have the code changes for the settings read code.  Once you've accepted these two pull requests, I can remove the branches, sort out any problems on my master, and go back to having a single branch, correct?  I'll try to be more careful about anticipating your desire to split things in the future. :-)


components/settings/settings.cpp
```diff
@@ -117,6 +120,230 @@ class SettingsFileParser
...
+                missed = true;
+                changed = true;
+            }
+            // Ensure that there's a newline before the next section if we added settings.
+            if (missed) ostream << std::endl;
+        }
+
+        // This logic triggers when the input file was effectively empty.  It could be extended
+        // to doing something more intelligent instead, like make a copy of the default settings
+        // file (complete with comments) before continuing.  Other code prevents OpenMW from
```
scrawl: IMO we shouldn't copy the default settings file into the user settings file. The idea when I wrote the settings system was that the user settings file only contains changed settings, so that:
1. If the "upstream" default of a setting changes, users are automatically upgraded to the new default if they never touched that particular setting.
2. It's easier for the user to tell which options they've changed. The file is smaller giving more convenient access to the settings that the user is interested in.

components/settings/settings.cpp
```diff
@@ -117,6 +120,230 @@ class SettingsFileParser
...
+        // executing to this point with a missing config file.
+        if (!existing) {
+            ostream << "# This file was automatically generated." << std::endl << std::endl;
```
scrawl: Maybe add a link to the documentation for settings-default.cfg or link to the wiki documentation here.

components/settings/settings.cpp
cfcohen: I actually agree with you, but that's not how it currently works (at all as far as I can tell).  I defined one setting (near clip) in my user settings.cfg, and ran a build of openmw-launcher without my changes.   After loading a save file and immediately exiting, my user settings.cfg contained 121 settings in 16 sections. :-o  

I assumed that was on purpose.  ;-)  If the goal is to have the file only contain settings that have actually changed, this code can move us towards that behavior.   Also, following your recent removal of "ui sensitivity", I received assertion style errors refusing to run OpenMW because my settings.cfg did not contain the required value.  That provided more evidence to me that the file was intended to be complete.

To play devil's advocate for a moment, it does not seem like very many of these settings would be the same for every one on a given machine, or that there's a reasonable default to "build on" since many of the settings are a subjective matter of personal preference.  In some ways a system that simply helps the user manage a full copy of their personal configuration as the game engine evolves might be a perfectly reasonably approach.  That's the approach I was designing for, but not because I preferred it -- because that's how it seems to work now. 

Should I understand your comments here to mean that the goal is to reverse that behavior? If so I'm fine with that...

scrawl: > ran a build of openmw-launcher without my changes. After loading a save file and immediately exiting, my user settings.cfg contained 121 settings in 16 sections. :-o

Was not aware of that. Definitely not intended. One more reason to nuke the launcher settings code and make it use the existing settings component.

Try removing your settings file, then running openmw directly, not through the launcher. That should yield the intended behaviour. 

> Also, following your recent removal of "ui sensitivity", I received assertion style errors refusing to run OpenMW because my settings.cfg did not contain the required value.

If the settings.cfg does not contain a required value, we fall back to settings-default.cfg. In this case, I have to assume you pulled the latest code, but didn't actually compile it - so the setting was removed, but the code accessing the setting was still there. In such cases throwing a runtime error is the correct thing to do.

> That's the approach I was designing for, but not because I preferred it -- because that's how it seems to work now.
> 
> Should I understand your comments here to mean that the goal is to reverse that behavior?

Yes.

As far as I could tell, your branch <i>does</i> behave correct w.r.t. not copying the complete default settings into user settings (barring the launcher bug). I just mentioned the issue because the comment on this line suggested that as a future possibility:

> It could be extended
+        // to doing something more intelligent instead, like make a copy of the default settings
+        // file (complete with comments) before continuing

cfcohen: > Was not aware of that. Definitely not intended. One more reason to nuke the launcher settings code and make it use the existing settings component.

Agreed.  Does that mean I'm on the right track to write a wrapper for the working settings code to make it use QStrings and QMaps for the Launcher?   If so I'll start working on that.  My plan was to convert the one data structure to the other and maintain the existing API with the exception of passing an QIOstream.   I intend to pass file name instead and wrap the call in a try block to generate the nice Qt warning dialog.

> I have to assume you pulled the latest code, but didn't actually compile it.

That's likely.  But the settings-default.cfg would have been the old one as well, correct?  Perhaps there was another factor in play as well (like using or not using the launcher).

> As far as I could tell, your branch does behave correct w.r.t. not copying the complete default settings into user settings (barring the launcher bug). I just mentioned the issue because the comment on this line suggested that as a future possibility:

Yes, I thought that only updating existing options was the right approach as well.  The comment  was in part because I thought that's how you wanted it to work.   I'm clear on what the desired behavior is now.   I still have one question though:

If a user changes a setting to a non-default value, and then changes it back to a default one, do we want to remove it from the user configuration?   Is there a data structure in the code that says what the default value is?  If so, I don't know where it is.  I supposed I could read the default-settings.cfg and make a different Settings map to compare to...  My current code uses the approach that whatever is in the user configuration is what should be written back out to the user configuration.  Conceptually it is possible to detect that the setting has reverted to the default value, and then remove both the setting and the comments since the last setting from the user configuration.   That would be very clear and consistent with the overall approach in my opinion.  Is this what you want, or is the plan that once user configured, always user configured?

scrawl: > If a user changes a setting to a non-default value, and then changes it back to a default one, do we want to remove it from the user configuration?

I wouldn't do that, no.

> That's likely. But the settings-default.cfg would have been the old one as well, correct?

The settings-default.cfg is copied into the build folder by running cmake. Code is recompiled with make. So running one but not the other can result in mismatched executable / settings ('make' automatically runs cmake though).

scrawl: > Does that mean I'm on the right track to write a wrapper for the working settings code to make it use QStrings and QMaps for the Launcher?

Sounds good.

cfcohen:
This version corrects some of the issues, but still isn't finished.  I ended up eliminating GraphicsSettings entirely because it was much cleaner that way.

I'm seeking your feedback and guidance on how to address the remaining issues.  In particular, the launcher either saves all of the graphics settings or none of them, based on whether GraphicsPage::saveSettings() was called.   That's not consistent with your guidance to only save modified settings, but I'm unclear on how you want me to tell what's been modified.  

I could set a boolean for each widget when the widget changes, and then test the boolean to decide whether to call mEngineSettings.setXXX().  I think that's unappealing because it would include situations where the user changed their mind and reverted to the default before "saving".

I believe that the correct approach is to test the new value against the value loaded from the settings file, which would require keeping a set of settings from the initial load.   I also think there should be a system for removing default values from the settings.cfg, but you seem to disagree.  That's fine, but it means that most users will eventually degenerate into all settings being in the settings.cfg.   If we do this, maybe we should just accept that all graphics settings will be written by the launcher and move on.  I feel like this is inconsistent with the idea that the settings.cfg file only contains the modified values, and it's probably part of how we degenerated into a system that always wrote everything.

Also, there's currently no mechanism for installing any comments in the settings.cfg file in any official way so it kind of defeats the purpose of preserving the comments when re-writing settings.cfg.  This is not my preferred solution, but it's at least acceptable since a user can install their own settings.cfg and get comment preserving behavior.  Are you opposed to adding comments to the settings.cfg file by default?  If not would any approach that involved reading the settings-default.cfg to get the comments above each setting, and copy them into the settings.cfg when adding a new option for the first time be acceptable?   If this was where we also obtained data about the "default" setting, does that change your opinion on removing default settings?  Is it clear what I proposing?  If not maybe I should describe the system more completely (or even write it) before you decide.

scrawl:
> I'm seeking your feedback and guidance on how to address the remaining issues. In particular, the launcher either saves all of the graphics settings or none of them, based on whether GraphicsPage::saveSettings() was called. That's not consistent with your guidance to only save modified settings, but I'm unclear on how you want me to tell what's been modified.

We could connect the modification events on individual widgets to the respective Settings::set... call rather than setting <i>everything</i> on exit.

At least now only the launcher-configurable video settings always get written. That's much better than the previous behaviour, good enough for now. I would rather get this PR merged sooner and split further improvements to new PRs.

> That's fine, but it means that most users will eventually degenerate into all settings being in the settings.cfg.

Only if the user has touched all settings. I expect the average user will not fiddle with semi-internal engine settings like the map sizes. If you find your settings file gets too messy, you can manually remove entries if you want.

> Are you opposed to adding comments to the settings.cfg file by default? If not would any approach that involved reading the settings-default.cfg to get the comments above each setting, and copy them into the settings.cfg when adding a new option for the first time be acceptable?

I'm ok with that for a future PR.

> does that change your opinion on removing default settings?

No. A setting explicitely set by the user should be kept, no matter if it's the default.

scrawl:
This is the settings file after a quick test run:

<pre>
# This file was automatically generated.


[Video]

antialiasing = 0

fullscreen = false

resolution x = 1024

resolution y = 768

screen = 0

vsync = false

window border = true

# Automatically added category.
[Saves]

character = ashstorm
</pre>

IMO the Automatically added / generated notices are too verbose.

If we want to put a comment on top of newly generated files, how about something like this:

> This is the OpenMW user settings file. The user settings file only contains explicitely changed settings. If you would like to revert a setting to its default, simply remove it from this file. For available settings, see the file 'settings-default.cfg' or the documentation at https://wiki.openmw.org/index.php?title=Settings .

cfcohen:
> We could connect the modification events on individual widgets to the respective Settings::set... call rather than setting everything on exit.

Ok.  I'll commit that next (perhaps later this evening).

> I expect the average user will not fiddle with semi-internal engine settings like the map sizes.

Good point.   I hadn't thought of that.  At worst it will degenerate to the settings that are configurable in the Options menu, which isn't too bad/many.

> IMO the Automatically added / generated notices are too verbose.

Agreed.   Part of my goal there was to serve as a placeholder for some better comment generation approach, but I'll remove that from this pull request.

I like the proposed comment.  I presume that we'd just embed that text in the source code to reduce complexity, and that would be the only comment by default in this pull request.

And the more I think about it, the current approach is simple and clean for novice users, upgrades smoothly as users tinker more in the Options menu, and if they really want to edit their settings.cfg like I did with comment in it, then maybe some advice on the wiki about how to copy settings-default.cfg to settings.cfg while running will suffice as a way to install comments.

cfcohen:
This commit should ready to accept now (or at least close).  I tried to implement detecting the changed settings using the connection of slots to the widgets, and that was a total failure. :-(  The fullscreen, screen number, standard resolution and custom resolution widgets are connected in a fairly complex way, and its practically impossible to keep both the GUI and the settings file updates in sync without generating unnecessary updates.

The revised approach was to simply make a copy of mEngineSettings into mInitialSettings and compare each before writing.  It's perhaps slightly less efficient, but there's less code this way, and it's _much_ clearer how the code works, which is what really matters.   After messing with the former approach for more than an hour, this one worked on the first try in only a few minutes.

I added a comment like you suggested, and just embraced that there would be no other comments in the file for novice users (which eliminated some fussy code involving newlines).  I am now very comfortable with this approach as a final solution.  Expert users can install their own file with comments.

Among the nice features resulting from this change is that users can run openmw-launcher, and openmw and still have no settings.cfg file afterwards, which is what I think you intended all along.

My proposal for a new/separate pull request is to resolve the only other remaining code cleanup issue that I'm currently aware of in settings management.   The OpenMW Launcher still uses the same slightly broken config/settingsbase.cpp code for the launcher settings as the Graphics tab did.   It would not be a huge change to add the "multi-value" support from the launcher settings to the Settings::Manager class, and then eliminate the former component entirely.   

The plan would be to: Make settings values an std::vector of strings instead of a single string.  Modify each of getInt, getBool, getString, getFloat to confirm that there's only one value, generate a warning if there's not, and return the first value.  Then add a new getStringMulti that returns a set of strings (this is the one API that the launcher settings needs).  In the future getBoolMulti, getIntMulti, etc. could be implemented if needed.   The setX APIs woudl replace values for backwards compatibility and there could be addX and setXMulti APIs as well if needed.

The only down side would be the slight performance hit of storing all settings in a vector, but that might end up being needed someday, and again I think we're primarily talking about code cleanliness and consistency as the goals here, not performance.   Please let me know if you'd like to see a new pull request for this.

scrawl:
Why the need to copy to mInitialSettings? The initial settings are still in mEngineSettings immediately before we save. Something like this should work:

<pre>
    bool cVSync = vSyncCheckBox->checkState();
    if (iVSync != mEngineSettings.getBool("vsync", "Video")
        mEngineSettings.setBool("vsync", "Video", cVSync);
</pre>

> My proposal for a new/separate pull request is to resolve the only other remaining code cleanup issue that I'm currently aware of in settings management. The OpenMW Launcher still uses the same slightly broken config/settingsbase.cpp code for the launcher settings as the Graphics tab did. It would not be a huge change to add the "multi-value" support from the launcher settings to the Settings::Manager class, and then eliminate the former component entirely.

> The plan would be to: Make settings values an std::vector of strings instead of a single string. Modify each of getInt, getBool, getString, getFloat to confirm that there's only one value, generate a warning if there's not, and return the first value. Then add a new getStringMulti that returns a set of strings (this is the one API that the launcher settings needs). In the future getBoolMulti, getIntMulti, etc. could be implemented if needed. The setX APIs woudl replace values for backwards compatibility and there could be addX and setXMulti APIs as well if needed.

openmw.cfg and launcher.cfg are a totally different beast. The notion of default settings does not exist.  We don't just need multi-values, they need to be returned in the correct order and need to be merged in the correct order when multiple files are read. We need to remove values (e.g. a content file) from the original file if the settings to write don't include that value any more. And someone requested to keep commented out multi-values in the file <i>in</i> the correct order, which resulted in some really complex writing code. These are totally different requirements than the graphics settings have. Let's keep the code separate. I appreciate clean-up, but usually there has to be some higher purpose or feature in mind for the clean up to be worth it.

cfcohen:
> Why the need to copy to mInitialSettings?

Uh.  I have no idea.   I'll make that change now. :-)  I think it's because I was still in the mind set of having slots update the settings object as the widgets changed, and I hadn't fully switched mentally to the new approach.  Good catch.

> The notion of default settings does not exist. 

Loading config file: launcher.cfg 
Loading config file: /home/cory/.config/openmw/launcher.cfg 
Loading config file: /home/cory/.config/openmw/openmw.cfg 
Loading config file: /usr/local/etc/openmw/openmw.cfg 
Loading config file: openmw.cfg 
Loading config file: /home/cory/.config/openmw/openmw.cfg

Then perhaps we're loading some things that we shouldn't be? ;-)

> And someone requested to keep commented out multi-values in the file in the correct order, which resulted in some really complex writing code.  These are totally different requirements than the graphics settings have.

I saw this code for openmw.cfg, and agree completely in that case and would not even propose changing that code for exactly that reason.  However, my understanding is that this is _not_ true of components/config/settingsbase.hpp and components/config/launchersettings.?pp, which appears to simply be another windows INI file, but with multi-values.   In fact, until this change, GraphicsSettings and LauncherSettings were closely related (both sharing SettingsBase), and now they're not.  We have three systems for handling config values.  Config::GameSettings was always the "odd man out".

I'm not arguing, because I'd be just as happy not to do it, but I wanted to make sure that you were operating on complete information when making the decision.  There seems to be several components related to configuration file loading that have gotten stale, and aren't functioning as intended. ;-)

scrawl:
I will do a final test run and then merge if everything's ok.

> > The notion of default settings does not exist.

> Loading config file: launcher.cfg 
> Loading config file: /home/cory/.config/openmw/launcher.cfg 
> Loading config file: /home/cory/.config/openmw/openmw.cfg 
> Loading config file: /usr/local/etc/openmw/openmw.cfg 
> Loading config file: openmw.cfg 
> Loading config file: /home/cory/.config/openmw/openmw.cfg

>  Then perhaps we're loading some things that we shouldn't be? ;-)

We have a hierarchy of config files, but they do not function as default settings, at least I wouldn't call it that. The multi-values in all the files get merged together, rather than only the value in the most derived file taking effect. Configuring the engine resources / data files is an entirely different use case from simple graphics settings. We shouldn't try to shoehorn one system into another. It would become a spaghetti monster prone to bugs and hard to extend. The code is complicated enough as-is.

> There seems to be several components related to configuration file loading that have gotten stale, and aren't functioning as intended. ;-)

What is not functioning as intended?

cfcohen:
Sorry for the commit flailing.  I should have read your example more carefully before committing.  It's cleaner the way you did it.  I'm not used to my mistakes being so visible. :-)

cfcohen:
> What is not functioning as intended?

I can't say since I don't know how it's supposed to work.   Experimentally, we appear to load multiple launcher.cfg files, with no merging of mutli values, but merging of config _files_ and then we write out the entire state to the users launcher.cfg every time, effectively disabling the earlier file  for future loads.   If that's what it's supposed to do, then I guess it's fine.

scrawl:
There is only one launcher.cfg as far as I can tell. Not sure why the code tries to load one that isn't there.

cfcohen:
Maybe the fix then is to simply remove the logic that attempts to load more than one launcher.cfg.   Then it would make sense to me even if it mostly duplicated code from elsewhere (e.g. two different Windows INI parsing systems).  I think I'm losing steam (or maybe more correctly running out of vacation time) anyway.   I just wanted you to be aware...

PR #821, opened 2015-11-25
Sound manager refactoring

kcat:
This pull request attempts to fix up some things I haven't been happy about with the audio code. Including:

* Voice files (```say``` commands) are now streamed instead of buffered. There's not much sense in buffering them since they tend to be long and not often spoken.
* Buffer management is handled by the sound manager, rather than the output implementation. That is to say, the sound manager handles tracking the buffer cache size, and (un)loads buffers on-demand.
* ESM data is preloaded into a list of Sound_Buffer objects (which has the resource name, volume, and min/max distance fixups preapplied, and stores an opaque handle for the output implementation's buffer). This allows for referencing individual sound records via index, reducing the number of string lookups, copying, and comparisons. In the future, it would enable other subsystems to do a one-time string lookup for a sound buffer index (or list of indices if using multiple sounds at random) and simply use those from then on.
* The streaming code has been cleaned up a lot.
* Active sounds are now keyed on the ```MWWorld::Ptr```, which provides more efficient lookups.
* OpenAL now batches updates on implementations that can do it, ensuring that all position and volume changes occur at the same time (implementations that can't do it behave as they always have).
* Rename ```playManualSound3D``` to ```playSound3D```. This is C++, we can have two functions that have the same name and differ by parameters.
* Get rid of ```Play_NoTrack```. It just does the same thing as ```playManualSound3D```, except getting the initial position from a ```Ptr```. Now it's just passed the position from the ```Ptr```.


scrawl:
> Voice files (say commands) are now streamed instead of buffered. There's not much sense in buffering them since they tend to be long and not often spoken.

This would be great but unfortunately NPC speech isn't fully streamed yet. The lip data is still created by decoding the full file at once. Would it be possible to generate the lip data in the streaming thread? I imagine we could "build" the sound buffer as it's decoded, then create or update the loudness vector up to that particular point when a new buffer chunk comes in.

> ESM data is preloaded into a list of Sound_Buffer objects (which has the resource name, volume, and min/max distance fixups preapplied, and stores an opaque handle for the output implementation's buffer). This allows for referencing individual sound records via index, reducing the number of string lookups, copying, and comparisons. In the future, it would enable other subsystems to do a one-time string lookup for a sound buffer index (or list of indices if using multiple sounds at random) and simply use those from then on.

I see a few problems with this.

I don't imagine many subsystems will start using the index lookup feature, since strings are easier to work with and easier to debug. Most subsystems aren't going to change much more anyway since we're just reimplementing Morrowind, and most post-1.0 features will be offered via scripting support. New sound scripting capabilities may or may not benefit from indexing of sound definitions.

There seems to be a performance overhead from the preloading of the Sound_Buffer list (about 3-4ms for me). It would be better to load on demand, so the loading of each sound is completely negligible and the overall loading is more evenly distributed over the course of a play session.

A static list doesn't allow for modifying or inserting sound definitions at runtime (not used at the moment, but I don't see a reason not to allow it given that other record types like levelled lists, and craftable item types already have that capability). 

If we want to cut down on the number of string comparisons for ESM records, maybe it would make more sense to do this on the ESM store level (e.g. support hash keys). That's probably not worth the effort though.

kcat:
> This would be great but unfortunately NPC speech isn't fully streamed yet. The lip data is still created by decoding the full file at once. Would it be possible to generate the lip data in the streaming thread? I imagine we could "build" the sound buffer as it's decoded, then create or update the loudness vector up to that particular point when a new buffer chunk comes in.

The lip data is created the first time a voice file is played, then kept indefinitely for the runtime of the game. It'd be tricky to load it when streaming because the streaming update size may not be a good multiple for the lip "fps", causing the loudness samples to use a differing number of audio samples. It'd be kind of messy too, as the streaming sound would have to know when to call back to the sound manager, and do so asynchronously. And thinking ahead we may want the lip data to come in as premade files rather than by analyzing the audio data (e.g. to support shapes and blendings, not unlike Oblivion and Skyrim).

If there was a way to know ahead of time that a voice will be played, it could decode and analyze the audio in the background and have it ready by the time the voice is eventually played. This would work for loading lip data from files, too. The trick is to know ahead of time when a given voice file is going to be used.

An alternative is to analayze the audio and create the loudness data in the background when its called to play, and the sound manager will wait to play the actual voice until that's done. It depends on how much delay you're willing to allow.

> I don't imagine many subsystems will start using the index lookup feature, since strings are easier to work with and easier to debug. Most subsystems aren't going to change much more anyway since we're just reimplementing Morrowind, and most post-1.0 features will be offered via scripting support. New sound scripting capabilities may or may not benefit from indexing of sound definitions.

Even within the sound manager itself, being able to reference and track sounds with an integer is far better than looking up and copying strings. Other subsystems would just be a bonus. Aside from the benefits it's already employing, there are future benefits to be had too. In particular, I'm thinking of cases where a randomized sound is played. Currently that's accomplished by looking up all sound records that start with a given string (using case-insensitive string compares), and picking one to give to the sound manager.

Given the new setup, which has a case-smashed sorted name list, it would be dead-simple to be able to tell the sound manager to play a randomized sound that starts with a given string. Since it's using a lower_bound binary search lookup anyway, it can also do an upper_bound binary search using a constrained comparison, then just roll for something between those two. But that relies on all sound buffers having been prepared in advance (if they get prepared as-needed, it can't ever assume all records that start with a given string have been prepared, and so has to constantly search them).

> There seems to be a performance overhead from the preloading of the Sound_Buffer list (about 3-4ms for me). It would be better to load on demand, so the loading of each sound is completely negligible and the overall loading is more evenly distributed over the course of a play session.

3 to 4ms on initialization? That's negligible, and better than doing continual resizing of the buffer list every time a new sound is encountered. Not preloading the buffer list also has some performance overhead: either the list is unsorted, making string lookups slower as it has to do linear searches instead of the current binary searches, or the list uses sorted insertions, which invalidates indices and requires more string copies and lookups.

> A static list doesn't allow for modifying or inserting sound definitions at runtime (not used at the moment, but I don't see a reason not to allow it given that other record types like levelled lists, and craftable item types already have that capability).

I don't see how creating new sound record definitions at runtime could work in any meaningful manner. The values you'd specify for the record depend on the contents of the resource it describes (i.e. the volume and minrange fields depend on the overall volume of the recording in the named sound file), and if you know that, you may as well predefine a record entry.

But even still.. it is theoretically possible to add new sound records, at least as long as only the sound manager deals with the indices. When a new sound record is created, the sound manager would be alerted. It can do a sorted insert for the new record, then go through its lists and fix all displaced indices (add one to all indices at or above the insertion point).

scrawl:
I'm still not convinced that the practical benefits of the integer indices warrant the added complexity, but I'll shut up about it now. If Zini has no objections we can merge.

> The lip data is created the first time a voice file is played, then kept indefinitely for the runtime of the game. It'd be tricky to load it when streaming because the streaming update size may not be a good multiple for the lip "fps", causing the loudness samples to use a differing number of audio samples.

> It'd be kind of messy too, as the streaming sound would have to know when to call back to the sound manager, and do so asynchronously.

A bit tricky, but doable. I'd like to try it at some point. The benefit would be getting rid of the lag that I'm seeing when lots of NPC's start talking at the same time (try aggro-ing the mages guild).

> An alternative is to analyze the audio and create the loudness data in the background when its called to play, and the sound manager will wait to play the actual voice until that's done. It depends on how much delay you're willing to allow.

This would be my second best choice if the streaming idea fails.

> If there was a way to know ahead of time that a voice will be played

Nope, no way. The dialogue system is to complicated to predict it in advance.

kcat:
If it's more desirable, another possibility might be to store the ```Sound_Buffer```` objects in a ```std::deque``` so pointers to individual entries remain constant (so long as new entries are pushed on to the end), and then referencing ```Sound_Buffer*```s instead of indices and having a ```str::string,Sound_Buffer*``` map for string lookup. This wouldn't be able to efficiently look up a randomized sound, but it would be better than it was.

zinnschlag:
Actually, I also have problems with the indexing; mostly for the same reasons scrawl mentioned.

The deque idea looks a bit more reasonable. The randomised sound feature isn't too important IMO. We are probably moving away from that approach post 1.0.

I do not agree with the argument about mutable records that scrawl made. There are plenty of record types that are immutable and I would expect most to stay that way. Especially, I can not think of a situation where we would need to add or modify sounds at runtime.


Apart from this point I am okay with merging this PR. Having the buffer size configurable would be nice, but I don't mind merging without it.

kcat:
Switched to the deque+map, and added config options for the cache size.

PR #827, opened 2015-11-28
Changed relative includes to library header format

EmperorArthur:
I was playing around with http://www.flourish.org/cinclude2dot/cinclude2dot and found these.  They were the only relative include paths, besides to_utf8/tests, in all of components.

### DIFF
--- a/components/bsa/bsa_file.cpp
+++ b/components/bsa/bsa_file.cpp
@@ -28,8 +28,6 @@
 #include <boost/filesystem/path.hpp>
 #include <boost/filesystem/fstream.hpp>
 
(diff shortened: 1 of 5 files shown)


EmperorArthur:
Found an actual error.  Apparently the compilers are forgiving about quotes vs angle brackets.

EmperorArthur:
Not sure if this one's an error or not.
In "components/contentselector/view/contentselector.hpp"
```
#include "ui_contentselector.h"
```
That file doesn't exist.

scrawl:
ui_*.h files are generated during the build process from QT .ui files. I guess you'll need to point your tool to consider the build folder as include path .

PR #828, opened 2015-11-28
work around X server freeze in crashcatcher

scrawl:
This change appears to fix a random X server lock-up that I get about 1 in 10 times when a crash is caught. I'm presuming it's a bug in the X server since faulty applications shouldn't be able to crash or freeze the X server under any circumstances. I am unsure why this particular change works around the issue but I'm not about to go digging in the X server code to find out. That thing is big and scary...

### DIFF
--- a/apps/openmw/crashcatcher.cpp
+++ b/apps/openmw/crashcatcher.cpp
@@ -382,16 +382,17 @@ static void crash_handler(const char *logfile)
     fflush(stdout);
 
     if(crash_info.pid > 0)
-    {
         gdb_info(crash_info.pid);
-        kill(crash_info.pid, SIGKILL);
-    }
 
     if(logfile)
     {
         std::string message = "OpenMW has encountered a fatal error.\nCrash log saved to '" + std::string(logfile) + "'.\n Please report this to https://bugs.openmw.org !";
         SDL_ShowSimpleMessageBox(0, "Fatal Error", message.c_str(), NULL);
     }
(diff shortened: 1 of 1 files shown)


kcat:
I'm not sure you want to do that, because it will cause the crashed process to stay up while waiting for the user to handle the the message box. If the mouse is locked, or the game is in fullscreen, getting to the message box could be a problem.

scrawl:
The message box appears on top of a full screen window for me just fine. 

Mouse locking could be an issue. But you can always close the message box with the enter key.

scrawl:
Changed to use a usleep() as per kcat's suggestion, which works around the bug as well, while preserving the previous behaviour w.r.t. closing the crashed window.

psi29a:
Will this code be run in windows?  You'll have to create your own usleep function. 

If we use a C++11 compiler then you can use this:
```
#include <chrono>
#include <thread>

std::this_thread::sleep_for(std::chrono::microseconds(usec));
```

If we don't have C++11, but have boost, then you can do this instead:

```
#include <boost/thread/thread.hpp>
#include <boost/date_time/posix_time/posix_time.hpp>

boost::this_thread::sleep(boost::posix_time::microseconds(usec));
```

scrawl:
> Will this code be run in windows?

No. crashcatcher.cpp is only used on unix systems.

psi29a:
Roger that! ;)

PR #829, opened 2015-11-29
Fix Bug #2467

EmperorArthur:
Use fCorpseClearDelay instead of iMonthsToRespawn.

Thanks to musicgusto for finding where in the code to change, and the correct GMST to use

### DIFF
--- a/apps/openmw/mwworld/cellstore.cpp
+++ b/apps/openmw/mwworld/cellstore.cpp
@@ -738,6 +738,11 @@ namespace MWWorld
                     Ptr ptr (&*it, this);
                     ptr.getClass().respawn(ptr);
                 }
+            }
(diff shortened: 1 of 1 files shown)


EmperorArthur:
Crap.  Don't merge.  Forgot to do something.

EmperorArthur:
I'm rebasing now.

EmperorArthur:
This is better, but the issue is mLastRespawn.
With this code, containers will never respawn.

EmperorArthur:
The issue is https://bugs.openmw.org/issues/2369

I propose that we do the timing check here:
https://github.com/OpenMW/openmw/blob/de98ee006228b5b2cd46102d544665463db044b6/apps/openmw/mwclass/npc.cpp#L1205

Add a last loaded timestamp or something to NPC custom data.

I don't want to touch anything that would affect the save format, so feel free to tackle this issue.

scrawl:
#908

PR #832, opened 2015-12-01
Quick fix for building on MSVC

ace13:
`components\nifosg\controller.hpp(55): C2899: typename cannot be used outside a template declaration`
Removing the second typename solves it, probably because that's technically in definition and not declaration.

### DIFF
--- a/components/nifosg/controller.hpp
+++ b/components/nifosg/controller.hpp
@@ -52,7 +52,7 @@ namespace NifOsg
         {
         }
 
-        ValueInterpolator(boost::shared_ptr<const MapT> keys, typename MapT::ValueType defaultVal = typename MapT::ValueType())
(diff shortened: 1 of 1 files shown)


zinnschlag:
I'll merge that for now, but we should try to avoid this kind of workarounds.

ace13:
I would love to avoid platform/compiler specific defines like this, but when MSVC does things its own way then it's hard to do.

I did read through some of the documents regarding `typename`, and it's just vague enough to be debatable which of the compilers do the right thing in this scenario. But sticking out far enough to require a workaround is probably not the right thing to do, regardless of if you're right or not.

scrawl:
Would a typedef help?

<pre>
        typedef typename MapT::ValueType VT;

        ValueInterpolator(boost::shared_ptr<const MapT> keys, VT defaultVal = VT())
</pre>

ace13:
Just had time to check, and yeah, it seems to work with a typedef as well. Would've thought that would have the same problem.

PR #833, opened 2015-12-02
Reuse DATADIR in data paths

AMDmi3:


scrawl:
This patch uses DATADIR on line 91 before it's defined on line 370.

psi29a:
So we either need to move our defines (SET) near the top, or push the data files file locations down after the defines.

scrawl:
https://github.com/OpenMW/openmw/commit/7d6fa1b65a828f700170fbaa4f25af090f4cd3bc

PR #835, opened 2015-12-02
enable opengl es1

sandstranger:
In this PR set on GLES1 . It shows absolutely all objects, including water, sky , terrain. OSG must be build with GLES1 profile . SDL2 must be build wit gles1 .
PS : It works of course , only if set water shader to off in settings, because gles1 not support shaders , and textures of course must be unpacked , if videocard not support dxt compression.
It shows 60 fps on external locations in full-hd resolution. LOL . I use scrawl osg fork.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,14 +168,23 @@ endif (ANDROID)
 
 option(OPENGL_ES "enable opengl es support" FALSE )
 
+option(USE_QT "Use Qt in building" TRUE )
+
+if (NOT BUILD_LAUNCHER AND NOT BUILD_OPENCS)
+    set(USE_QT FALSE CACHE BOOL "disable Qt" FORCE )
+else()
+    set(USE_QT TRUE CACHE BOOL "enable Qt" FORCE )
+endif()
+
+add_definitions (-DUSE_QT)
+ 
 if (OPENGL_ES)
-    INCLUDE(cmake/FindOpenGLES.cmake)
+    find_package(OpenGLES)
     add_definitions (-DOPENGL_ES)
-    INCLUDE_DIRECTORIES(${OPENGLES_INCLUDE_DIR})
 endif (OPENGLES)
 
 # Dependencies
-if (NOT ANDROID)
+if (USE_QT)
     set(DESIRED_QT_VERSION 4 CACHE STRING "The QT version OpenMW should use (4 or 5)")
     message(STATUS "Using Qt${DESIRED_QT_VERSION}")
 
@@ -220,7 +229,7 @@ IF(BOOST_STATIC)
     set(Boost_USE_STATIC_LIBS   ON)
 endif()
 
-if (NOT ANDROID)
+if (USE_QT)
     find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgQt osgUtil osgFX)
 else()
     find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgUtil osgFX)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,10 +45,6 @@ endif(EXISTS ${PROJECT_SOURCE_DIR}/.git)
 # Macros
 include(OpenMWMacros)
 
-if (ANDROID)
-    set(CMAKE_FIND_ROOT_PATH ${OPENMW_DEPENDENCIES_DIR} "${CMAKE_FIND_ROOT_PATH}")
-endif (ANDROID)
-
 # doxygen main page
 
 configure_file ("${OpenMW_SOURCE_DIR}/docs/mainpage.hpp.cmake" "${OpenMW_BINARY_DIR}/docs/mainpage.hpp")
@@ -161,28 +157,24 @@ endif()
 
 if (ANDROID)
     set(CMAKE_FIND_ROOT_PATH ${OPENMW_DEPENDENCIES_DIR} "${CMAKE_FIND_ROOT_PATH}")
-    set(OSG_PLUGINS_DIR ${OSG_PLUGINS_DIR})
-    add_definitions (-DOSG_PLUGINS_DIR)
+    add_definitions (-DOSG_PLUGINS_DIR=${OSG_PLUGINS_DIR})
     set(OPENGL_ES TRUE CACHE BOOL "enable opengl es support for android" FORCE)
 endif (ANDROID)
 
 option(OPENGL_ES "enable opengl es support" FALSE )
 
-option(USE_QT "Use Qt in building" TRUE )
+if (OPENGL_ES)
+    add_definitions(-DOPENGL_ES)
+endif(OPENGL_ES)
 
 if (NOT BUILD_LAUNCHER AND NOT BUILD_OPENCS AND NOT BUILD_WIZARD)
-    set(USE_QT FALSE CACHE BOOL "disable Qt" FORCE )
+   set(USE_QT FALSE)
 else()
-    set(USE_QT TRUE CACHE BOOL "enable Qt" FORCE )
+   set(USE_QT TRUE)
 endif()
 
-add_definitions (-DUSE_QT)
+add_definitions(-DUSE_QT)
  
-if (OPENGL_ES)
-    find_package(OpenGLES)
-    add_definitions (-DOPENGL_ES)
-endif (OPENGLES)
-
 # Dependencies
 if (USE_QT)
     set(DESIRED_QT_VERSION 4 CACHE STRING "The QT version OpenMW should use (4 or 5)")
@@ -230,10 +222,11 @@ IF(BOOST_STATIC)
 endif()
 
 if (USE_QT)
-    find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgQt osgUtil osgFX)
-else()
-    find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgUtil osgFX)
+    set (OSG_QT osgQt)
 endif()
+
+find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle ${OSG_QT} osgUtil osgFX)
+
 include_directories(${OPENSCENEGRAPH_INCLUDE_DIRS})
 
 if(OSG_STATIC)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,6 @@ endif()
 
 if (ANDROID)
     set(CMAKE_FIND_ROOT_PATH ${OPENMW_DEPENDENCIES_DIR} "${CMAKE_FIND_ROOT_PATH}")
-    add_definitions (-DOSG_PLUGINS_DIR=${OSG_PLUGINS_DIR})
     set(OPENGL_ES TRUE CACHE BOOL "enable opengl es support for android" FORCE)
 endif (ANDROID)
 
@@ -172,8 +171,6 @@ if (NOT BUILD_LAUNCHER AND NOT BUILD_OPENCS AND NOT BUILD_WIZARD)
 else()
    set(USE_QT TRUE)
 endif()
-
-add_definitions(-DUSE_QT)
  
 # Dependencies
 if (USE_QT)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,7 +170,7 @@ option(OPENGL_ES "enable opengl es support" FALSE )
 
 option(USE_QT "Use Qt in building" TRUE )
 
-if (NOT BUILD_LAUNCHER AND NOT BUILD_OPENCS)
+if (NOT BUILD_LAUNCHER AND NOT BUILD_OPENCS AND NOT BUILD_WIZARD)
     set(USE_QT FALSE CACHE BOOL "disable Qt" FORCE )
 else()
     set(USE_QT TRUE CACHE BOOL "enable Qt" FORCE )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,20 +159,36 @@ if (WIN32)
     add_definitions(-DNOMINMAX -DWIN32_LEAN_AND_MEAN)
 endif()
 
+if (ANDROID)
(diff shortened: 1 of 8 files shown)


scrawl:
Very nice. I'll leave some comments on the individual files.

components/sdlutil/sdlcursormanager.cpp
```diff
@@ -217,6 +217,13 @@ namespace SDLUtil
 
     void SDLCursorManager::_createCursorFromResource(const std::string& name, int rotDegrees, osg::Image* image, Uint8 size_x, Uint8 size_y, Uint8 hotspot_x, Uint8 hotspot_y)
     {
+         #ifdef ANDROID
+             return;
+         #endif
+
+        if (mCursorMap.find(name) != mCursorMap.end())
+             return;
```
scrawl: Why was this line duplicated?

components/CMakeLists.txt
```diff
@@ -172,38 +176,62 @@ include_directories(${BULLET_INCLUDE_DIRS} ${CMAKE_CURRENT_BINARY_DIR})
...
+        ${OSGVIEWER_LIBRARIES}
+        ${OSGGA_LIBRARIES}
+        ${OSGFX_LIBRARIES}
+        ${OSGANIMATION_LIBRARIES}
+        ${BULLET_LIBRARIES}
+        ${SDL2_LIBRARY}
+        # For MyGUI platform
+        ${OPENGL_gl_LIBRARY}
+        ${MYGUI_LIBRARIES}
+        )
```
scrawl: Use an intermediate variable here to avoid duplicating the whole target_link_libraries. E.g.

<pre>
if (ANDROID)
   set(GL_LIB ${OPENGLES_gl_LIBRARY})
else()
   set(GL_LIB ${OpenGL_gl_LIBRARY})
endif()

target_link_libraries(
....
${GL_LIB}
...
)
</pre>

apps/openmw/CMakeLists.txt
```diff
@@ -140,18 +140,50 @@ target_link_libraries(openmw
...
+	png
     )
 endif (ANDROID)
 
+
+if (OPENGL_ES)
+    target_link_libraries(openmw
+        ${OPENGLES_gl_LIBRARY}
+    )
+endif (OPENGL_ES)
```
scrawl: This line shouldn't be needed if you add the OPENGLES_gl_LIBRARY to components/CMakeLists.txt as suggested in my last comment.

apps/openmw/CMakeLists.txt
```diff
@@ -140,18 +140,50 @@ target_link_libraries(openmw
...
-        cpufeatures
         BulletCollision
         LinearMath
+        z
+        osg
+        osgDB
+        osgAnimation
+        osgText
+        osgUtil
+        osgShadow
```
scrawl: Better to use e.g. ${OSGSHADOW_LIBRARIES} instead of plain 'osgShadow'.

Also, we don't use osgShadow yet.

Also, osg, osgUtil, osgText, osgDB and osgAnimation are included in ${OPENSCENEGRAPH_LIBRARIES} so shouldn't need to specify them separately.

CMakeLists.txt
```diff
@@ -204,7 +220,11 @@ IF(BOOST_STATIC)
     set(Boost_USE_STATIC_LIBS   ON)
 endif()
 
-find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgQt osgUtil osgFX)
+if (NOT ANDROID)
+    find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgQt osgUtil osgFX)
```
scrawl: <pre>
set (OSG_COMPONENTS osgDB osgViewer osgText osgGA osgAnimation osgParticle osgQt osgUtil osgFX)
if (NOT ANDROID)
   set (OSG_COMPONENTS ${OSG_COMPONENTS} osgQt)
endif()

find_package(OpenSceneGraph 3.2.0 REQUIRED ${OSG_COMPONENTS})
</pre>

scrawl: For the QT stuff, it might be better to add a new variable USE_QT that controls use of osgQt, the finding of QT libraries and the compilation of QT code in components/. USE_QT would be set to false if BUILD_LAUNCHER and BUILD_OPENCS are both disabled.

CMakeLists.txt
```diff
@@ -159,20 +159,36 @@ if (WIN32)
...
 
-if (DESIRED_QT_VERSION MATCHES 4)
-    find_package(Qt4 REQUIRED COMPONENTS QtCore QtGui QtNetwork QtOpenGL)
-else()
-    find_package(Qt5Widgets REQUIRED)
-    find_package(Qt5Core REQUIRED)
-    find_package(Qt5Network REQUIRED)
-    find_package(Qt5OpenGL REQUIRED)
+if (OPENGL_ES)
+    INCLUDE(cmake/FindOpenGLES.cmake)
```
scrawl: Normally you would do find_package(OpenGLES) instead of including the cmake file. With find_package, you'll get error reporting if OpenGL ES is not found.

sandstranger:
I have got now issues with fonts and textures
http://imgur.com/rWyAm3o
I will try to solve it later.

psi29a:
Great work! :)

CMakeLists.txt
```diff
@@ -159,20 +159,45 @@ if (WIN32)
...
-    find_package(Qt5OpenGL REQUIRED)
+    set(USE_QT TRUE CACHE BOOL "enable Qt" FORCE )
+endif()
+
+add_definitions (-DUSE_QT)
+ 
+if (OPENGL_ES)
+    find_package(OpenGLES)
+    add_definitions (-DOPENGL_ES)
+endif (OPENGLES)
```
scrawl: Should be endif (OPENGL_ES). I get a warning:

<pre>
CMake Warning (dev) in CMakeLists.txt:
  A logical block opening on the line

    /home/scrawl/Dev/openmw/CMakeLists.txt:181 (if)

  closes on the line

    /home/scrawl/Dev/openmw/CMakeLists.txt:184 (endif)

  with mis-matching arguments.
This warning is for project developers.  Use -Wno-dev to suppress it.
</pre>

CMakeLists.txt
```diff
@@ -159,20 +159,45 @@ if (WIN32)
...
+    set(USE_QT FALSE CACHE BOOL "disable Qt" FORCE )
 else()
-    find_package(Qt5Widgets REQUIRED)
-    find_package(Qt5Core REQUIRED)
-    find_package(Qt5Network REQUIRED)
-    find_package(Qt5OpenGL REQUIRED)
+    set(USE_QT TRUE CACHE BOOL "enable Qt" FORCE )
+endif()
+
+add_definitions (-DUSE_QT)
```
scrawl: This definition is not used.

CMakeLists.txt
```diff
@@ -159,20 +159,45 @@ if (WIN32)
...
+    set(OPENGL_ES TRUE CACHE BOOL "enable opengl es support for android" FORCE)
+endif (ANDROID)
 
-set(DESIRED_QT_VERSION 4 CACHE STRING "The QT version OpenMW should use (4 or 5)")
-message(STATUS "Using Qt${DESIRED_QT_VERSION}")
+option(OPENGL_ES "enable opengl es support" FALSE )
 
-if (DESIRED_QT_VERSION MATCHES 4)
-    find_package(Qt4 REQUIRED COMPONENTS QtCore QtGui QtNetwork QtOpenGL)
+option(USE_QT "Use Qt in building" TRUE )
```
scrawl: USE_QT should be a plain variable, not an option. It makes no sense to show it as option in cmake-gui since whatever you set there is overwritten by the cmake script anyway.

The whole block can be replaced by:

<pre>
if (NOT BUILD_LAUNCHER AND NOT BUILD_OPENCS AND NOT BUILD_WIZARD)
   set(USE_QT FALSE)
else()
   set(USE_QT TRUE)
endif()
</pre>

CMakeLists.txt
```diff
@@ -159,20 +159,45 @@ if (WIN32)
     add_definitions(-DNOMINMAX -DWIN32_LEAN_AND_MEAN)
 endif()
 
-# Dependencies
+if (ANDROID)
+    set(CMAKE_FIND_ROOT_PATH ${OPENMW_DEPENDENCIES_DIR} "${CMAKE_FIND_ROOT_PATH}")
+    set(OSG_PLUGINS_DIR ${OSG_PLUGINS_DIR})
+    add_definitions (-DOSG_PLUGINS_DIR)
```
scrawl: This definition seems to be unused. Maybe you are confusing definitions with variables? add_definitions affects the compiling of code. set(VAR VALUE) affects the cmake script.

Also, if you wanted to pass the plugin directory to the code, you'd need to add a value to the definition as well e.g.
<pre>
add_definitions (-DOSG_PLUGINS_DIR=${OSG_PLUGINS_DIR})
</pre>

CMakeLists.txt
```diff
@@ -159,20 +159,45 @@ if (WIN32)
     add_definitions(-DNOMINMAX -DWIN32_LEAN_AND_MEAN)
 endif()
 
-# Dependencies
+if (ANDROID)
+    set(CMAKE_FIND_ROOT_PATH ${OPENMW_DEPENDENCIES_DIR} "${CMAKE_FIND_ROOT_PATH}")
```
scrawl: Would it be possible to change the CMAKE_FIND_ROOT_PATH in your cmake call? e.g. instead of
<pre>
cmake -DOPENMW_DEPENDENCIES_DIR=path/to/deps
</pre>
you would call
<pre>
cmake -DCMAKE_FIND_ROOT_PATH=path/to/deps
</pre>

This way we avoid cluttering our cmake script with more platform specific code.

Also by the way, the CMAKE_FIND_ROOT_PATH is set twice, once on line 163 and once on line 49.

CMakeLists.txt
```diff
@@ -159,20 +159,45 @@ if (WIN32)
...
+    set(OSG_PLUGINS_DIR ${OSG_PLUGINS_DIR})
```
scrawl: What is the purpose of this line? Setting a variable to the same value it's had before is a no-op.

CMakeLists.txt
```diff
@@ -204,7 +229,11 @@ IF(BOOST_STATIC)
     set(Boost_USE_STATIC_LIBS   ON)
 endif()
 
-find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgQt osgUtil osgFX)
+if (USE_QT)
+    find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgQt osgUtil osgFX)
+else()
+    find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle osgUtil osgFX)
+endif()
```
scrawl: Use an intermediate variable here to avoid the duplication of library lists. Suggestion in comment https://github.com/OpenMW/openmw/pull/835#discussion_r46482768

CMakeLists.txt
```diff
@@ -159,20 +159,45 @@ if (WIN32)
...
-    find_package(Qt5Core REQUIRED)
-    find_package(Qt5Network REQUIRED)
-    find_package(Qt5OpenGL REQUIRED)
+    set(USE_QT TRUE CACHE BOOL "enable Qt" FORCE )
+endif()
+
+add_definitions (-DUSE_QT)
+ 
+if (OPENGL_ES)
+    find_package(OpenGLES)
```
scrawl: You want to add a REQUIRED so the cmake script will complain when the OpenGL ES library is not found.
<pre>
find_package(OpenGLES REQUIRED)
</pre>

Also, it might be better to move the find_package call to the file where the ```${OPENGLES_gl_LIBRARY}``` is actually used, i.e. in components/CMakeLists.txt next to ```find_package(OpenGL REQUIRED)```.

sandstranger:
I tried to use cmake -DOPENMW_DEPENDENCIES_DIR=path/to/deps
but it found nothing in deps folder. Maybe it is issue in android cmake toolchain

scrawl:
> add_definitions (-DOSG_PLUGINS_DIR=${OSG_PLUGINS_DIR})

> add_definitions(-DUSE_QT)

These definitions seem to be unused.

sandstranger:
add_definitions (-DOSG_PLUGINS_DIR=${OSG_PLUGINS_DIR}) used for linking osg_plugins in 
[code]
 if (ANDROID)
+    set (OSG_PLUGINS
+        -Wl,--whole-archive
+        ${OSG_PLUGINS_DIR}/libosgdb_dds.a
+        ${OSG_PLUGINS_DIR}/libosgdb_bmp.a
+        ${OSG_PLUGINS_DIR}/libosgdb_tga.a
+        ${OSG_PLUGINS_DIR}/libosgdb_gif.a
+        ${OSG_PLUGINS_DIR}/libosgdb_jpeg.a
+        ${OSG_PLUGINS_DIR}/libosgdb_png.a
+        -Wl,--no-whole-archive
+    )
[/code]

scrawl:
> add_definitions (-DOSG_PLUGINS_DIR=${OSG_PLUGINS_DIR}) used for linking osg_plugins here
> if (ANDROID)
> set (OSG_PLUGINS
> -Wl,--whole-archive
> ${OSG_PLUGINS_DIR}/libosgdb_dds.a
> ${OSG_PLUGINS_DIR}/libosgdb_bmp.a
> ${OSG_PLUGINS_DIR}/libosgdb_tga.a
> ${OSG_PLUGINS_DIR}/libosgdb_gif.a
> ${OSG_PLUGINS_DIR}/libosgdb_jpeg.a
> ${OSG_PLUGINS_DIR}/libosgdb_png.a
> -Wl,--no-whole-archive
> )

You are confusing cmake variables with definitions. Definitions are passed on to the compiler flags. Variables are available within the cmake script.

Here is an example use of a definition:
<pre>
# in the cmake script
add_definitions(-DFOOBAR)

// in the .cpp file
#if FOOBAR
// do something
#endif
</pre>

Here is an example use of a cmake variable:
<pre>
# in the cmake script
set(VAR "foobar")
if (VAR STREQUAL "foobar")
   message(STATUS "var is foobar")
endif()
</pre>

sandstranger:
Ok , I understand it now

sandstranger:
As I thought it was problem with loading texture formats in gles1 , I convert one texture to RGB565 and is it rendered now in game

sandstranger:
It seems it is issue in osg osg dds loader for gles 1. I convert all textures to tga and it loaded without any problems . In dds container only RGB565 format rendered

PR #836, opened 2015-12-03
More audio cleanups and crash fix

kcat:
* Separates Sound objects from the output implementation.
* Fixes a crash during chargen, caused by some say() calls not associated with a ```Ptr```.


zinnschlag:
This will fix: https://bugs.openmw.org/issues/3053

PR #839, opened 2015-12-04
Fix a build issue

ace13:
```
C2536: 'SceneUtil::LightSource::SceneUtil::LightSource::mLight' : cannot specify explicit initializer for arrays
```

### DIFF
--- a/components/sceneutil/lightmanager.cpp
+++ b/components/sceneutil/lightmanager.cpp
@@ -260,10 +260,10 @@ namespace SceneUtil
 
     LightSource::LightSource(const LightSource &copy, const osg::CopyOp &copyop)
         : osg::Node(copy, copyop)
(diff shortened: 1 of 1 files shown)


scrawl:
<pre>error: ‘copy_n’ is not a member of ‘std’</pre>
Missing ```algorithm``` include?

scrawl:
copy_n appears to be a C++11 function.

ace13:
I'm apparently far too used to C++11.

scrawl:
std::begin is also C++11.

I've pushed different fix to master, that should work.

ace13:
Thanks, sorry for being too new :grin:

ace13:
I really need to finish the final remaining AppVeyor issues, so we can have windows CI support for pull requests.
The 1h+ build time for those is not that nice though...

PR #840, opened 2015-12-04
Object movement between cells

scrawl:
https://bugs.openmw.org/issues/1176

~~Still need to add save game handling, restore some game functionality~~ and do lots more testing. Before I continue I want to ask if I'm on the right track with this. I was inclined not to build merging into the CellRefList because that's also used in ContainerStore. I don't feel that handling merging separately for each reference type is a good idea. We want the system to be easy to maintain and extend after all. Basically the type-specific list access had to be removed, and that's why there's still game code stubbed out that I need to port over to the new forEach() mechanism. Thoughts?

### DIFF
--- a/apps/openmw/mwworld/cellstore.cpp
+++ b/apps/openmw/mwworld/cellstore.cpp
@@ -179,8 +179,54 @@ namespace MWWorld
         return (ref.mRef.mRefnum == pRefnum);
     }
 
+    void CellStore::moveFrom(const Ptr &object, CellStore *from)
+    {
+        MovedRefTracker::iterator found = mMovedToAnotherCell.find(object.getBase());
+        if (found != mMovedToAnotherCell.end())
+        {
+            // A cell we had previously moved an object to is returning it to us.
+            assert (found->second == from);
+            mMovedToAnotherCell.erase(found);
+        }
+        else
+        {
+            mMovedHere.insert(std::make_pair(object.getBase(), from));
+        }
+    }
+
+    void CellStore::moveTo(const Ptr &object, CellStore *cellToMoveTo)
+    {
+        MovedRefTracker::iterator found = mMovedHere.find(object.getBase());
+        if (found != mMovedHere.end())
+        {
+            // Special case - object didn't originate in this cell
+            // Move it back to its original cell first
+            CellStore* originalCell = found->second;
+            assert (originalCell != this);
+            originalCell->moveFrom(object, this);
+
+            mMovedHere.erase(found);
+
+            // Now that object is back to its rightful owner, we can move it
+            originalCell->moveTo(object, cellToMoveTo);
+
+            updateMergedRefs();
+            return;
+        }
+
+        cellToMoveTo->moveFrom(object, this);
+        mMovedToAnotherCell.insert(std::make_pair(object.getBase(), cellToMoveTo));
+
+        updateMergedRefs();
+    }
+
+    void CellStore::updateMergedRefs()
+    {
+
+    }
+
     CellStore::CellStore (const ESM::Cell *cell)
-      : mCell (cell), mState (State_Unloaded), mHasState (false), mLastRespawn(0,0)
+        : mCell (cell), mState (State_Unloaded), mHasState (false), mLastRespawn(0,0)
     {
         mWaterLevel = cell->mWater;
     }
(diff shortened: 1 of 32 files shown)


scrawl:
Also:
* For objects created at runtime (that don't have a RefNum) we don't need the movement tracker. Just delete old object and create new object in the new cell. (Not implemented yet).

scrawl:
Should we rename cell functors (Functor) to Visitor? It is pretty much an implementation of the visitor pattern.

zinnschlag:
Hm, surprising! That is not how I would have done it. It seems you are not actually moving the object from the core cell data structure of the original cell to the core cell data structure of the destination cell anymore. Instead you keep the LiveCellRef where it is and add another data structure on top of it to keep track of movement.
My original thoughts on this problem were to keep the old copy and delete approach and just add a tracker for it.
But I suppose your approach works too. And it actually has advantages, because the original LiveCellRef is maintained (not invalidating pointers to it). The question then is, would it be worthwhile to do the same for instances created at runtime?

Yes, about the renaming to Visitor. This is clearly a visitor pattern.

apps/openmw/mwworld/cellstore.cpp
```diff
@@ -179,8 +182,110 @@ namespace MWWorld
...
+            // A cell we had previously moved an object to is returning it to us.
+            assert (found->second == from);
+            mMovedToAnotherCell.erase(found);
+        }
+        else
+        {
+            mMovedHere.insert(std::make_pair(object.getBase(), from));
+        }
+
+        if (mState == State_Loaded)
```
zinnschlag: This has potentially poor performance when moving many actors between cells. The whole point of the preload state is to not load cells that we actually don't need to touch. But that is clearly not the case here. I suggest to just load the cell first when in preload state instead of the else branch.

scrawl: That's what I wanted to do at first, but we don't have the ESMReader to load from.

scrawl:
The problem with "copy and delete" is that "delete" does not mean inaccessible. Deleted objects can still be accessed by a script (e.g. to un-delete them, https://github.com/OpenMW/openmw/pull/392).  Or did you mean actually delete rather than marking as count 0? That seems dangerous.

> The question then is, would it be worthwhile to do the same for instances created at runtime?

Instances created at runtime don't have a refnum so can't be uniquely identified in a save game. When saving/loading we would have to copy&delete for these instances anyway. So why not do it right away?

apps/openmw/mwworld/cellstore.cpp
zinnschlag: Would it be worth to give the CellStore a reference to the World::ESMStore and the esm reader vector as a member variables? The size increase of the CellStore class should be minimal and it would simplify the interface a bit.

zinnschlag:
> The problem with "copy and delete" is that "delete" does not mean inaccessible. Deleted objects (if they have a refnum) can still be accessed by a script (e.g. to un-delete them, #392). Or did you mean actually delete rather than marking as count 0? That seems dangerous.

Nope. That would be dangerous indeed. Your solution is indeed better.

> Instances created at runtime don't have a refnum so can't be uniquely identified in a save game. When saving/loading we would have to copy&delete for these instances anyway. So why not do it right away?

For the reason you have given above.

scrawl:
> For the reason you have given above.

The script access for deleted objects does not affect objects with no refnum. These become inaccessible forever when they're deleted.

apps/openmw/mwworld/cellstore.cpp
scrawl: Sure, will do.

scrawl:
Ready for review.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -721,11 +720,11 @@ namespace MWWorld
         const Scene::CellStoreCollection& collection = mWorldScene->getActiveCells();
         for (Scene::CellStoreCollection::const_iterator cellIt = collection.begin(); cellIt != collection.end(); ++cellIt)
         {
-            FindContainerFunctor functor(ptr);
-            (*cellIt)->forEachContainer(functor);
+            FindContainerVisitor visitor(ptr);
+            //(*cellIt)->forEachContainer(visitor);
```
MiroslavR: This visitor isn't fed anything right now.

scrawl: Fixed, thanks!

scrawl:
@zinnschlag had a chance to look at this yet?

zinnschlag:
Sorry for the delay. I missed your request for another review. Anyway, looks okay to me.

PR #841, opened 2015-12-05
Write save games to a memory stream first

scrawl:
Two motivations for doing this:

- Previously, if the user chooses to overwrite an existing save file, and we have an exception thrown during the save process, the existing file would be lost. This is no longer an issue.
- Many small writes to a file are slow. Very slow. Writing to memory first then writing the completed file to disk appears to be ~500% faster.

### DIFF
--- a/apps/openmw/mwstate/statemanagerimp.cpp
+++ b/apps/openmw/mwstate/statemanagerimp.cpp
@@ -207,7 +207,9 @@ void MWState::StateManager::saveGame (const std::string& description, const Slot
         else
             slot = getCurrentCharacter()->updateSlot (slot, profile);
 
-        boost::filesystem::ofstream stream (slot->mPath, std::ios::binary);
+        // Write to a memory stream first. If there is an exception during the save process, we don't want to trash the
+        // existing save file we are overwriting.
+        std::stringstream stream;
 
         ESM::ESMWriter writer;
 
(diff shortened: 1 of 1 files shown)


scrawl:
We can probably close https://bugs.openmw.org/issues/2495 now. I have a save file with half the map explored and 400 journal pages, took 0.3s to save.

The save files are still bigger compared to vanilla MW but that's a separate issue.

scrawl:
I wonder if similar improvement could be applied to loading (i.e. load whole file into memory first, then parse it).

scrawl:
https://stackoverflow.com/questions/12997131/stdfstream-buffering-vs-manual-buffering-why-10x-gain-with-manual-buffering

scrawl:
> I wonder if similar improvement could be applied to loading (i.e. load whole file into memory first, then parse it).

Slightly better, but not 5x as with saving.

<pre>
diff --git a/apps/openmw/mwstate/statemanagerimp.cpp b/apps/openmw/mwstate/statemanagerimp.cpp
index b76d9ea..b2830de 100644
--- a/apps/openmw/mwstate/statemanagerimp.cpp
+++ b/apps/openmw/mwstate/statemanagerimp.cpp
@@ -345,8 +345,16 @@ void MWState::StateManager::loadGame (const Character *character, const std::str
     {
         cleanup();
 
+        osg::Timer timer;
+        boost::filesystem::ifstream filestream;
+        filestream.open(filepath, std::ios::binary);
+
+        std::stringstream* buffer = new std::stringstream;
+        Files::IStreamPtr bufferPtr (buffer);
+        (*buffer) << filestream.rdbuf();
+
         ESM::ESMReader reader;
-        reader.open (filepath);
+        reader.open (bufferPtr, filepath);
 
         if (reader.getFormat() > ESM::SavedGame::sCurrentFormat)
             throw std::runtime_error("This save file was created using a newer version of OpenMW and is thus not supported. Please upgrade to the newest OpenMW version to load this file.");
@@ -486,6 +494,8 @@ void MWState::StateManager::loadGame (const Character *character, const std::str
 
         // Do not trigger erroneous cellChanged events
         MWBase::Environment::get().getWorld()->markCellAsUnchanged();
+
+        std::cout << "loaded in " << timer.time_m() << std::endl;
     }
     {
</pre>

With patch:
> loaded in 2158.04

Without patch:
> loaded in 3726.11

No change at all if the file to load is already cached in memory by the OS.

zinnschlag:
The correct solution for issue one would be to save to a temporary file then delete the original save file and rename the temporary file. That's what we do in OpenMW-CS. Still the performance increase makes it worthwhile. Will merge.

PR #842, opened 2015-12-05
use tga textures instead dds

sandstranger:
Itn this PR I add tga texture support , because there is issues in osg dds loader in gles1 render . Unpacked dds textures not shows . So now I convert all textures to tga and store it with .dds name .

### DIFF
--- a/components/resource/texturemanager.cpp
+++ b/components/resource/texturemanager.cpp
@@ -164,21 +164,15 @@ namespace Resource
             }
 
             osg::ref_ptr<osgDB::Options> opts (new osgDB::Options);
+#ifndef USE_TGA 
             opts->setOptionString("dds_dxt1_detect_rgba"); // tx_creature_werewolf.dds isn't loading in the correct format without this option
+#endif
             size_t extPos = normalized.find_last_of('.');
             std::string ext;
             if (extPos != std::string::npos && extPos+1 < normalized.size())
                 ext = normalized.substr(extPos+1);
 
-#ifdef USE_TGA
-            osgDB::ReaderWriter* reader =NULL;
-            if (ends_with(ext,"dds"))
-              reader = osgDB::Registry::instance()->getReaderWriterForExtension("tga");
-            else
-              reader = osgDB::Registry::instance()->getReaderWriterForExtension(ext);
-#else
             osgDB::ReaderWriter* reader =osgDB::Registry::instance()->getReaderWriterForExtension(ext);
-#endif 
 
             if (!reader)
             {
@@ -204,14 +198,6 @@ namespace Resource
         }
     }
 
-#ifdef USE_TGA
-    inline bool TextureManager::ends_with(std::string const & value, std::string const & ending)
-    {
-        if (ending.size() > value.size()) return false;
-        return std::equal(ending.rbegin(), ending.rend(), value.rbegin());
-    }
-#endif
-
     osg::ref_ptr<osg::Texture2D> TextureManager::getTexture2D(const std::string &filename, osg::Texture::WrapMode wrapS, osg::Texture::WrapMode wrapT)
     {
         std::string normalized = filename;
@@ -236,20 +222,16 @@ namespace Resource
             }
 
             osg::ref_ptr<osgDB::Options> opts (new osgDB::Options);
+#ifndef USE_TGA
             opts->setOptionString("dds_dxt1_detect_rgba"); // tx_creature_werewolf.dds isn't loading in the correct format without this option
-            size_t extPos = normalized.find_last_of('.');
+#endif 
+           size_t extPos = normalized.find_last_of('.');
             std::string ext;
             if (extPos != std::string::npos && extPos+1 < normalized.size())
                 ext = normalized.substr(extPos+1);
-#ifdef USE_TGA
-            osgDB::ReaderWriter* reader =NULL;
-            if (ends_with(ext,"dds"))
-              reader = osgDB::Registry::instance()->getReaderWriterForExtension("tga");
-            else
-              reader = osgDB::Registry::instance()->getReaderWriterForExtension(ext);
-#else
+
             osgDB::ReaderWriter* reader =osgDB::Registry::instance()->getReaderWriterForExtension(ext);
-#endif
+
             if (!reader)
             {
                 std::cerr << "Error loading " << filename << ": no readerwriter for '" << ext << "' found" << std::endl;
--- a/components/resource/texturemanager.cpp
+++ b/components/resource/texturemanager.cpp
@@ -169,7 +169,17 @@ namespace Resource
             std::string ext;
             if (extPos != std::string::npos && extPos+1 < normalized.size())
                 ext = normalized.substr(extPos+1);
-            osgDB::ReaderWriter* reader = osgDB::Registry::instance()->getReaderWriterForExtension(ext);
+
+#ifdef USE_TGA
+            osgDB::ReaderWriter* reader =NULL;
+            if (ends_with(ext,"dds"))
+              reader = osgDB::Registry::instance()->getReaderWriterForExtension("tga");
+            else
+              reader = osgDB::Registry::instance()->getReaderWriterForExtension(ext);
+#else
+            osgDB::ReaderWriter* reader =osgDB::Registry::instance()->getReaderWriterForExtension(ext);
+#endif 
+
             if (!reader)
             {
                 std::cerr << "Error loading " << filename << ": no readerwriter for '" << ext << "' found" << std::endl;
@@ -193,6 +203,14 @@ namespace Resource
             return image;
         }
     }
+
+#ifdef USE_TGA
+    inline bool TextureManager::ends_with(std::string const & value, std::string const & ending)
+    {
+        if (ending.size() > value.size()) return false;
+        return std::equal(ending.rbegin(), ending.rend(), value.rbegin());
+    }
+#endif
 
     osg::ref_ptr<osg::Texture2D> TextureManager::getTexture2D(const std::string &filename, osg::Texture::WrapMode wrapS, osg::Texture::WrapMode wrapT)
     {
@@ -223,7 +241,15 @@ namespace Resource
             std::string ext;
             if (extPos != std::string::npos && extPos+1 < normalized.size())
                 ext = normalized.substr(extPos+1);
-            osgDB::ReaderWriter* reader = osgDB::Registry::instance()->getReaderWriterForExtension(ext);
+#ifdef USE_TGA
+            osgDB::ReaderWriter* reader =NULL;
+            if (ends_with(ext,"dds"))
+              reader = osgDB::Registry::instance()->getReaderWriterForExtension("tga");
+            else
+              reader = osgDB::Registry::instance()->getReaderWriterForExtension(ext);
+#else
+            osgDB::ReaderWriter* reader =osgDB::Registry::instance()->getReaderWriterForExtension(ext);
+#endif
             if (!reader)
             {
                 std::cerr << "Error loading " << filename << ": no readerwriter for '" << ext << "' found" << std::endl;
@@ -246,11 +272,14 @@ namespace Resource
             // We need to flip images, because the Morrowind texture coordinates use the DirectX convention (top-left image origin),
             // but OpenGL uses bottom left as the image origin.
             // For some reason this doesn't concern DDS textures, which are already flipped when loaded.
-            if (ext != "dds")
+#ifdef USE_TGA 
+           image->flipVertical();
+#else
+           if (ext != "dds")
             {
                 image->flipVertical();
             }
-
+#endif
             osg::ref_ptr<osg::Texture2D> texture(new osg::Texture2D);
             texture->setImage(image);
             texture->setWrap(osg::Texture::WRAP_S, wrapS);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,6 +162,12 @@ endif (ANDROID)
 
 option(OPENGL_ES "enable opengl es support" FALSE )
 
+option(USE_TGA "enable tga texture support instead dds" FALSE )
+
+if (USE_TGA)
+    add_definitions(-DUSE_TGA)
+endif(USE_TGA)
+
 if (OPENGL_ES)
     add_definitions(-DOPENGL_ES)
 endif(OPENGL_ES)
--- a/components/misc/resourcehelpers.cpp
+++ b/components/misc/resourcehelpers.cpp
@@ -34,7 +34,11 @@ bool Misc::ResourceHelpers::changeExtensionToDds(std::string &path)
     std::string::size_type pos = path.rfind('.');
     if(pos != std::string::npos && path.compare(pos, path.length() - pos, ".dds") != 0)
     {
+#ifdef USE_TGA
+        path.replace(pos, path.length(), ".tga");
+#else
         path.replace(pos, path.length(), ".dds");
+#endif
         return true;
     }
     return false;
--- a/components/resource/texturemanager.hpp
+++ b/components/resource/texturemanager.hpp
@@ -55,9 +55,6 @@ namespace Resource
 
         osg::ref_ptr<osg::Texture2D> mWarningTexture;
 
-#ifdef USE_TGA
-        inline bool ends_with(std::string const & value, std::string const & ending);
-#endif
         bool mUnRefImageDataAfterApply;
 
         TextureManager(const TextureManager&);
--- a/components/resource/texturemanager.hpp
+++ b/components/resource/texturemanager.hpp
@@ -55,6 +55,9 @@ namespace Resource
 
         osg::ref_ptr<osg::Texture2D> mWarningTexture;
 
+#ifdef USE_TGA
+        inline bool ends_with(std::string const & value, std::string const & ending);
+#endif
         bool mUnRefImageDataAfterApply;
 
         TextureManager(const TextureManager&);


scrawl:
The correct place to handle this is in ResourceHelpers::correctResourcePath. https://github.com/OpenMW/openmw/blob/master/components/misc/resourcehelpers.cpp#L43

Actually the file references should be 'tga' to begin with, but we correct them to dds because the MW developers apparently changed format last minute.

components/resource/texturemanager.cpp
```diff
@@ -194,6 +204,14 @@ namespace Resource
         }
     }
 
+#ifdef USE_TGA
+    inline bool TextureManager::ends_with(std::string const & value, std::string const & ending)
```
jeaye: This should be a const member function

components/resource/texturemanager.cpp
```diff
@@ -169,7 +169,17 @@ namespace Resource
...
             if (extPos != std::string::npos && extPos+1 < normalized.size())
                 ext = normalized.substr(extPos+1);
-            osgDB::ReaderWriter* reader = osgDB::Registry::instance()->getReaderWriterForExtension(ext);
+
+#ifdef USE_TGA
+            osgDB::ReaderWriter* reader =NULL;
+            if (ends_with(ext,"dds"))
+              reader = osgDB::Registry::instance()->getReaderWriterForExtension("tga");
+            else
+              reader = osgDB::Registry::instance()->getReaderWriterForExtension(ext);
```
jeaye: Consider
```c++
osgDB::ReaderWriter * const reader
(osgDB::Registry::instance()->getReaderWriterForExtension(ends_with(ext,"dds") ? "tga" : ext));
```
or similar, in both places in this file. Just looking to minimize duplication and keep things const.

components/resource/texturemanager.cpp
jeaye: Better yet, this shouldn't be a member function at all; it should be a static function in this cpp file. It doesn't rely on any members and it's not used outside of this file.

sandstranger:
I suppose it will be better to use it like this . With this changes  I can use tga textures with ".tga" extensions . I suppose it will find at first dds texture in bsa archive and then I change path to my custom tga texture

scrawl:
The changes to texturemanager.cpp should not be needed.

The change to resourcehelpers.cpp looks strange (why replace a file extension with '.tga' if it's already .tga? just don't do replacing to begin with).

CMake option could use a better name. How about

option(DONT_CORRECT_TEXTURE_FILE_EXTENSION "Don't attempt to correct texture file references to .dds extension (breaks vanilla Morrowind game)" FALSE )

Actually, maybe we don't need a new build option. We could change the ResourceHelpers to only do the .dds change if the referenced .tga file doesn't exist. The only reason we didn't implement it like that to begin with was poor performance of the resource system when we were still using Ogre, but that's probably no longer an issue. I will do some profiling later today to see where we're at.

sandstranger:
what about this line in water.cpp . 
  texname << "textures/water/" << texture << std::setw(2) << std::setfill('0') << i << ".dds";
Dds extension use hetre directy .

sandstranger:
What about gui and water textures ? For example here
  texname << "textures/water/" << texture << std::setw(2) << std::setfill('0') << i << ".dds";
        textures.push_back(mResourceSystem->getTextureManager()->getTexture2D(texname.str(), osg::Texture::REPEAT, osg::Texture::REPEAT));
Dds extension used here directly

sandstranger:
When I try look to sun game crashes

scrawl:
Please try out current master with the following patch applied:

<pre>
diff --git a/components/misc/resourcehelpers.cpp b/components/misc/resourcehelpers.cpp
index 0c26357..0ab1f22 100644
--- a/components/misc/resourcehelpers.cpp
+++ b/components/misc/resourcehelpers.cpp
@@ -63,15 +63,12 @@ std::string Misc::ResourceHelpers::correctResourcePath(const std::string &topLev
 
     std::string origExt = correctedPath;
 
-    // since we know all (GOTY edition or less) textures end
-    // in .dds, we change the extension
+    if (vfs->exists(origExt))
+        return origExt;
+
     bool changedToDds = changeExtensionToDds(correctedPath);
-    if (vfs->exists(correctedPath))
+    if (changedToDds && vfs->exists(correctedPath))
         return correctedPath;
-    // if it turns out that the above wasn't true in all cases (not for vanilla, but maybe mods)
-    // verify, and revert if false (this call succeeds quickly, but fails slowly)
-    if (changedToDds && vfs->exists(origExt))
-        return origExt;
 
     // fall back to a resource in the top level directory if it exists
     std::string fallback = topLevelDirectory + "\\" + getBasename(correctedPath);
</pre>

scrawl:
> When I try look to sun game crashes

Does OpenGL ES not support occlusion queries? It still shouldn't crash though, of course.

sandstranger:
fonts now loaded . It were issues with Android search Paths . Game trying to find fonts in path "fonts" , and fonts were in folder with name "FONTS"

kcat:
> What about gui and water textures ? For example here
texname << "textures/water/" << texture << std::setw(2) << std::setfill('0') << i << ".dds";
textures.push_back(mResourceSystem->getTextureManager()->getTexture2D(texname.str(), osg::Texture::REPEAT, osg::Texture::REPEAT));
Dds extension used here directly without using resourcehelper , so tga textures will never be loaded

We could probably do with improving the resource handling code in a slightly more general manner. It seems like the vanilla engine will try to use the original texture as named, and then if it can't find or load it, it will look for alternatives with different extensions. Or perhaps it just straight-up ignores the extension, and just tries all files with the same base name until one works. Audio seems to do something similar, where even though the game data specifies .wav files that don't exist, it will load a different file with the same base name and an .mp3 extension automatically (at least for voices).

scrawl:
> We could probably do with improving the resource handling code in a slightly more general manner. It seems like the vanilla engine will try to use the original texture as named, and then if it can't find or load it, it will look for alternatives with different extensions. Or perhaps it just straight-up ignores the extension, and just tries all files with the same base name until one works. Audio seems to do something similar, where even though the game data specifies .wav files that don't exist, it will load a different file with the same base name and an .mp3 extension automatically (at least for voices).

My idea was to phase out the "ignore file extensions" quirk post-1.0  (along with other questionable logic, like automatically appending textures/ if it's missing, or falling back to the root path even if a subdirectory was specified). That's why we have external ResourceHelpers class not within the resource manager itself. The native OSG format for instance isn't using the ResourceHelpers.

kcat:
> My idea was to phase out the "ignore file extensions" quirk post-1.0 (along with other questionable logic, like automatically appending textures/ if it's missing, or falling back to the root path even if a subdirectory was specified).

We can't get rid of it completely since the vanilla game requires it. At most, we could just restrict it to NIFs, but there would be some use to keeping it. For instance, if a texture replacer wants to use a different format for some reason (like ktx or tga), it would also have to edit every model or land texture definition that specifies the original file. That would be rather bloaty, and in some cases impossible (if the original model isn't allowed to be redistributed or modified by third parties).

With audio it's even worse, because while dds is ubiquitous and can hold most relevant pixel formats, audio files use different extensions for different compression formats, so if you want to use Vorbis, FLAC, Speex, or Opus in place of existing wav or mp3 files, you have to modify everything that references the original files. Similarly for videos, instead of being able to simply make a webm or mp4 or whatever, you also have to change the references to the original .bik files. Or models, instead of just providing a replacement .osg, you have to modify the game data to change all references to the original NIF. Not only is this messy, but it also creates compatibility issues if you have different mods modifying sets of resources with some overlap.

The way I imagine it working is the VFS would search/enumerate files using the archive priority. So if ArchiveA is listed first and presents file.ext1 and file.ext2, and ArchiveB is listed later and provides file.ext3 and file.ext4, a search for "file.*" would return file.ext3, file.ext4, file.ext1, and file.ext2 in that order. Reverse the archive order, and you get file.ext1, file.ext2, file.ext3, and file.ext4 instead. This way, when searching for a texture, sound, video, or model, it can enumerate all files ignoring the original extension, and pick the first one that has a usable extension.

scrawl:
The reason I don't like ignoring filenames is that it introduces ambiguity - i.e. what if an archive contains both texture.ext and texture.ext2? Which one is it supposed to use? Though to be honest, we already have ambiguities as it is because of the case-insensitive file handling. We might want to log a warning to the output if an archive contains the "same" filename twice.

I am considering your idea. Seems cleaner than what we have now.

Should we do the same for the osgt, osgb and osgx formats? Should we allow replacing of a NIF file with an OSG file? Seems like a useful capability.

How do we know what extensions are usable and what they represent? We have external libraries (FFmpeg and OSG) handling the formats, and what formats they support depends on the build options. There might be a way to retrieve supported formats at runtime, though perhaps we should limit to sensible formats (i.e. it wouldn't make sense to have a video file replace an audio file, even though video technically does contain audio).

scrawl:
> This way, when searching for a texture, sound, video, or model, it can enumerate all files ignoring the original extension, and pick the first one that has a usable extension.

Would be nice to have a third opinion from @zinnschlag before someone starts working on this.

kcat:
> The reason I don't like ignoring file extensions is that it introduces ambiguity - i.e. what if an archive contains both texture.ext and texture.ext2? Which one is it supposed to use? Though to be honest, we already have ambiguities as it is because of the case-insensitive file handling. We might want to log a warning to the output if an archive contains the "same" filename twice.

Yeah, we already have ambiguity . Logging a warning seems sensible in both cases (same filename with different cases, or same base file with a different extension, in the same archive), since I can't see a need to do that. You could either alter the base name, or put it in a different directory, if it's supposed to be a different resource.

> Should we do the same for the osgt, osgb and osgx formats? Should we allow replacing of a NIF file with an OSG file? Seems like a useful capability.

That would be good, yes.

> How do we know what extensions are usable and what they represent? We have external libraries (FFmpeg and OSG) handling the formats, and what formats they support depends on the build options. There might be a way to retrieve supported formats at runtime, though perhaps we should limit to sensible formats (i.e. it wouldn't make sense to have a video file replace an audio file, even though video technically does contain audio).

It might be good enough to not worry about what extensions are usable. As it is, texture files are all in textures/, audio files are all in music/ or sounds/, videos are all in videos/, and models are all in meshes/.  So if a model is referencing a texture, it's a safe assumption that any other file in the same directory would also be a texture. If it's an unsupported format (e.g. OSG fails to open it), it can either be treated as a resource failure immediately, or it can try the next file in the list until it either finds one that works.

File extensions aren't really a good indicator anyway. Container formats like dds, wav, mp4, webm, etc, can be "supported", even though they can have unsupported payload formats (a wav file can have supported PCM data or unsupported mp3 data, a dds file can have supported RGB8 data or unsupported DXT data, etc).

kcat:
> Similarly for videos, instead of being able to simply make a webm or mp4 or whatever, you also have to change the references to the original .bik files. Or models, instead of just providing a replacement .osg, you have to modify the game data to change all references to the original NIF.

And expanding on this, since the references to the files aren't self-contained (videos are referenced within scripts, models are referenced in objects, etc), different mods may have modified these records in different ways. So to truly replace a resource using a different format, you'd have to make a different patch for every mod that alters these records. That's quite a bit of work for what's supposed to just be a simple resource replacer. Comparatively, using the file in the archive with the highest priority, while ignoring the file extension, makes this much simpler, and it shouldn't break any existing content (and it actually fixes some vanilla content. getting rid of specific hacks like tga->dds and wav->mp3).

cfcohen:
> Should we do the same for the osgt, osgb and osgx formats? Should we allow replacing of a NIF file with an OSG file? Seems like a useful capability.

I also support this approach.  I have modified my local copy to do this (in a hackish way), my recent discussion (pull) request was intended to support exactly this kind of scenario, and I believe that allowing the local configuration to have some say in the file formats used is a good thing in general.

zinnschlag:
We can get rid of ignoring file extensions for newer content and I think that is worth doing for all the reasons already listed. Adding even more extension ignoring seems like a move in the wrong direction.

After 1.0 we will push for moving from loose file collections to archives (one per content file by default). We can then add a flag that disables the ignoring mechanism for this archive (and make this flag set by default).

I don't think that the argument about being able to switch file formats without changing lots of stuff referencing resources files is valid, except for textures used in meshes. We have a global search and replace in OpenMW-CS after all.

As for textures that is an argument. A pure texture replacer would have the option to unset said flag.
As an alternative maybe an optional texture mapping could be introduced that allows replacing textures referenced in a mesh with a different texture? I can see several use cases for this feature, beyond basic texture replacers, especially if the mapping can be manipulated at runtime.

kcat:
> We can get rid of ignoring file extensions for newer content and I think that is worth doing for all the reasons already listed.

The only real issue is a potential ambiguity if a single archive contains files in the same directory with the same base name, but different extensions. Given the directory structure of the game, I can't see a reason for mods to do this (nor can I think of existing mods that do it), rather than using a different base name or different sub-directory. And for cases where it does appear, we can log or display a warning, same as if an archive contains the same full filename in the same directory, just with different capitalization. In either case, which file you get is effectively undefined.

> After 1.0 we will push for moving from loose file collections to archives (one per content file by default). We can then add a flag that disables the ignoring mechanism for this archive (and make this flag set by default).

This seems odd to me. The reason for the extension switching, as it is now, is because of improperly made content (whether it's ESMs referencing wavs that are really mp3, or models referencing tgas that are really dds). If the content is made correctly, there's no need to replace tga with dds, or wav with mp3, and it only serves to make the issue even more complicated since it makes the hack even more hacky. It's only when you get to deliberate replacers that ignoring extensions becomes a thing, and a generalized way of handling that would be nice.

> I don't think that the argument about being able to switch file formats without changing lots of stuff referencing resources files is valid, except for textures used in meshes. We have a global search and replace in OpenMW-CS after all.

That still seems unnecessarily complicated to me. If you simply have some .osg models that you want to replace some .nifs (or some oggs to replace mp3s), you would have to also include an esp/omwaddon that does nothing but rename the files in all the related records. When these records may also be modified by various mods, you'll end up with a bunch of patches just to ensure you have the records in sync with the various combinations of mods (and keep them up to date if some mods are actively updated).

Like for example, you want to replace sounds/file1.mp3, sounds/file2.mp3, and meshes/file3.nif with sounds/file1.ogg, sounds/file2.ogg, and meshes/file3.osg. So, this would require an addon that changes the records/scripts that reference those files, just replacing the extensions.  Now, Mod1 edits the records/scripts associated with sounds/file1.mp3 and sounds/file2.mp3, along with doing some other stuff. Mod2 edits the records/scripts associated with sounds/file2.mp3 and meshes/file3.nif, along with doing some other stuff. Mod3 exists as a patch for Mod1 and Mod2 to work correctly together. If we require an addon just to change the file extension, there would now need to be 4 addons in total... one for the base game, one for users that have Mod1 only, one for users that have Mod2 only, one for users that have Mod1, Mod2, and Mod3.

Next, imagine Mod1 gets updated, and makes a change to one of the records/scripts that references sounds/file1.mp3. Now you have to update two of your addons, the one for Mod1 only and the one for Mod1+Mod2+Mod3, and users will have to find the right version of your addons for the given versions  they have of the other mods. And this is just two mods and a compatibility patch for 3 resource replacements -- imagine dozens of mods with dozens of resource replacements.

All that, vs simply supplying sounds/file1.ogg, sounds/file2.ogg, and meshes/file3.osg in an archive with a higher priority than those other mods (that the user may or may not have), and the engine automatically picks them up as replacement sounds and a mesh since they're in a higher priority archive.

As well, for scripts specifically (with the PlayBink and Say functions), you may not even be allowed to distribute those patched addons if the author doesn't allow distributing edited versions of their code.

zinnschlag:
I see your point, but still ... Implicitly changing the extensions is messy and confusing. I am less worried about the ambiguity. The bigger problem here IMO is that the current system is confusing and unintuitive for mod developers. We should try to move to something cleaner one way or the other.

Okay, new ideas. Even two of them (mutually exclusive):

1. Extend the resource mapping feature I proposed above to all types of resources. A modder creating a resource replacer would still have to list all changed resources in a content file. But that can be largely automated and in the future a content file would be needed anyway to activate the archive.

2. Do away with extensions all together when referencing resources.
    - In the resources tables we could hide the extensions (sounds awfully like the crap Windows is doing with its files, but here it would actually make sense).
    - For content files saved in newer formats we could strip away the extensions from all references to resources; except in scripts maybe and there we could print a warning.
    - When creating new references the editor would not add the extensions (for example when adding a texture from a resources table to a land texture table or when adding a mesh to a referenceable).
    - An archive with multiple files of the same category (e.g. texture) with the same base name would be seen as defective by OpenMW-CS (still usable, but with a warning in the load error log).
    - Since post-1.0 OpenMW-CS will offer tools for managing resources archives, we can even reject attempts to add such files to an archive.

That would leave us only with texture references in mesh files where we can't strip out the extensions without breaking existing 3rd party tools. Acceptable, IMO.

sandstranger:
It seems Opengl Es1 not support Occlusion Query , because game crashes here
/home/sandstranger/Android/osg-master/src/osg/OcclusionQueryNode
pc 00000000014a4d9c  /data/app/com.libopenmw.openmw-2/lib/arm64/libopenmw.so (osg::QueryGeometry::drawImplementation(osg::RenderInfo&) const+440)

kcat:
I had a response all written out, and then lost it when my browser crashed.  Feh.

> The bigger problem here IMO is that the current system is confusing and unintuitive for mod developers. We should try to move to something cleaner one way or the other.

I agree. I just think that treating the file extension for what it is (a hint as to the internal format of the file), and subsequently ignoring it when loading a resource (since we already know everything in sound/ is a sound, everything in textures/ is a texture, etc, so the actual format is unimportant), is clean. If the game has a sound file sound/file1.wav, and I want to replace it with sound/file1.ogg, the engine should know that I mean this as a replacement... they're both sounds with the same base filename in the same location, just different formats. It's intuitive and easy -- to replace a sound, put another sound in the same place with the same base name, and different formats are okay (and encouraged in some instances, to avoid old proprietary formats).

> Do away with extensions all together when referencing resources.

This is essentially it. Strip off the resource's file extension, and look through a priority-sorted list for the first file matching the base name and location. If the extension doesn't exist on the specified resource, it's used as-is for the search, and if it does, it's removed before searching. Same behavior, removes the need for the current hacks we have, and makes resource replacement clear and easy since it removes format from the question of if a replacer will work as-is (as much as nif to osg can be an issue, imagine the base nif being replaced by an osg in a large mod, while a prefered replacer has a nif -- in this case it has the same extension as the original and would've worked on its own, but another mod has changed the "expected" format and prevents it from working without something more; this gets pretty confusing pretty easily).

cfcohen:
> Do away with extensions all together when referencing resources.

This is my preference for what it's worth.

sandstranger:
There are many issues on gles 1 render . Moons shows at day , fog works not as expected and many others things . I will try now to build openmw with gles1 render on desktop Linux

sandstranger:
With opengl on Android unpacked dds textures works without any errors

PR #843, opened 2015-12-06
Add HRTF options and fix an incorrect SoundID

kcat:
The HRTF options require the ALC_SOFT_HRTF extension to work (available in OpenAL Soft 1.17.0). I had intended to also add a Sound tab to the launcher where you could select a sound device and HRTF (since they can't be in the in-game options, due to lack of text strings), but I still need a bit of time for figuring out how to make that work and what options to have for it.

The SoundID fix is for https://bugs.openmw.org/issues/3060

### DIFF
--- a/files/settings-default.cfg
+++ b/files/settings-default.cfg
@@ -174,6 +174,13 @@ buffer cache min = 14
 # to this much memory until old buffers get purged.
 buffer cache max = 16
 
+# Specifies whether to enable HRTF processing. Valid values are: -1 = auto,
+# 0 = off, 1 = on.
+hrtf enable = -1
+
+# Specifies which HRTF to use when HRTF is used. Blank means use the default.
+hrtf =
+
 [Video]
 
 # Resolution of the OpenMW window or screen.
--- a/apps/openmw/mwclass/npc.cpp
+++ b/apps/openmw/mwclass/npc.cpp
@@ -1103,7 +1103,7 @@ namespace MWClass
             if(world->isUnderwater(ptr.getCell(), pos) || world->isWalkingOnWater(ptr))
                 return "DefaultLandWater";
             if(world->isOnGround(ptr))
-                return "Body Fall Medium";
+                return "DefaultLand";
             return "";
         }
         if(name == "swimleft")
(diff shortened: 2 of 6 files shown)


psi29a:
That means we'll be forced to use OpenAL-Soft 1.7 that was literally _just_ released? ;)

kcat:
OpenMW won't require 1.17, no. It's just if that if the extension isn't there, the options are effectively ignored.

psi29a:
Super!

files/settings-default.cfg
```diff
@@ -174,6 +174,13 @@ buffer cache min = 14
 # to this much memory until old buffers get purged.
 buffer cache max = 16
 
+# Specifies whether to enable HRTF processing. Valid values are: -1 = auto,
+# 0 = off, 1 = on.
+hrtf enable = -1
```
scrawl: Out of curiosity, how well does 'auto' work? You probably don't want to use HRTF for speakers, so how does OpenAL know if a device is headphone or speaker?

kcat: > so how does OpenAL know if a device is headphone or speaker?

The audio system can indicate what it's configured as. At least for DSound, WASAPI, and PulseAudio, it can to some degree. OpenAL Soft also has its own config option for dictating if a device should be treated as headphones or speakers, so it will Just Work if the user has it preconfigured. Though it ultimately errs on the side of caution... if it doesn't know it's headphones, it treats it as speakers and won't use HRTF unless explicitly requested to do so.

cfcohen:
I've updated the setting page on the wiki at:

https://wiki.openmw.org/index.php?title=Sound_Settings_Section

Based on this conversation and my ability to fill in the details with Wikipedia and a some Google.  If someone could explain what values are valid for the "htrf" and/or where they're defined at, that would probably be helpful.

scrawl:
Available settings for 'hrtf' are listed in the game log. For example I get:
<pre>
Enumerated HRTF names:
  default-44100.mhr
  default-48000.mhr
</pre>
You might get different listing on a different system / different version of OpenAL.

PR #844, opened 2015-12-06
First draft of serialization routines.

cfcohen:
This isn't really a pull request, but rather a request for review and commentary (if you're willing).   My real goal here was to learn some more about OpenSceneGraph, the OSGB/OSGT file format, OpenMW, NIF, etc.  That has certainly happened. :-) As usual, I've reached the point where I'm in over my head, and could use your guidance regardless of whether you ever intend to accept this pull request.

The general idea is that there's a new command "nifconv" that initializes the engine just far enough to load meshes, and then invokes the OutputStream writer to serialize the object.   There are two new shared libraries that are loaded dynamically by OSG to both read and write the extended object types.   Long term, I'd like to be able to run completely off OSGB assets as a proof of concept.  I've successfully written most of the NIFs in standard distribution, but I don't claim that they're all working properly (just the writing completes without crashing).

There are known problems with this commit, including a memory corruption problem when serializing to OSGT format that doesn't appear to happen when serializing to OSGB format.   I continue to find the OSG serialization code fragile, and confusing. :-(  I'll comment in more detail on individual parts of the code where I have questions or specific requests for changes (some changes are just minor bug fixes in the current code).


apps/nifconv/nifconv.cpp
```diff
@@ -0,0 +1,155 @@
...
+    // engine.setResourceDir(variables["resources"].as<std::string>());
+        
+    std::auto_ptr<VFS::Manager> mVFS;
+    mVFS.reset(new VFS::Manager(mFSStrict));
+    VFS::registerArchives(mVFS.get(), mFileCollections, mArchives, true);
+    mResourceSystem.reset(new Resource::ResourceSystem(mVFS.get()));
+
+    Resource::SceneManager* sceneManager = mResourceSystem->getSceneManager();
+
+#ifdef ACTIVATE
```
cfcohen: The consensus on the Internet seems to be that OSG trick for using global static object constructors to initialize an auto-magical plugin architecture is a BadIdea(tm).  For a long time I couldn't figure out how to get this to work at all because global objects are pulled in from static libraries _only_ if they're referenced in the main program on Linux.   Apparently the default behavior is different under Windows (first warning sign in my opinion).  To make matters worse, there was literally no way to reference the object from main in the standard OSG code, since the object had nothing but a constructor and destructor.   I ended up extending OSG with a null activate() method so that I could cause the object to be linked into the main program statically.  This code is no longer used since I figure out how to make the shared libraries properly, but I'm still uncertain if this is a "wise" architecture.

components/CMakeLists.txt
```diff
@@ -236,6 +236,23 @@ if (UNIX AND NOT APPLE)
...
 endif()
 
+add_library(osgdb_serializers_nifosg SHARED
+    serializers/nifloader serializers/controller
+    serializers/particle serializers/userdata
+    )
+
+target_link_libraries(osgdb_serializers_nifosg components)
+
+add_library(osgdb_serializers_sceneutil SHARED
```
cfcohen: It appears that OSG automatically loads shared libraries based on the namespace of the objects in the tree.  Since some are in NifOsg and others are SceneUtil, we ended up with two shared libraries (icky!).   Perhaps the solution is to move everything into one name space, or even an "OpenMW" namespace so that the serializer would load allo of our serializers at once?  I dunno.   Your guidance would be appreciated.

components/nif/nifkey.hpp
```diff
@@ -50,6 +50,12 @@ struct KeyMapT {
 
     KeyMapT() : mInterpolationType(sLinearInterpolation) {}
 
+    // For serializer.
```
cfcohen: The built in serializer sometimes _requires_ getX and setX methods to exist with very specific declarations.  To violate this convention requires hand-coding a lot of extra stuff. :-(  For this version, the primary goal was just to get it working.   Is this a problem?

components/nifosg/controller.cpp
```diff
@@ -85,7 +85,7 @@ KeyframeController::KeyframeController(const KeyframeController &copy, const osg
 }
 
 KeyframeController::KeyframeController(const Nif::NiKeyframeData *data)
-    : mRotations(data->mRotations)
+    : mRotations(data->mRotations, osg::Quat())
```
cfcohen: I think this is just a bug in the current code.

components/nifosg/controller.hpp
```diff
@@ -48,13 +48,14 @@ namespace NifOsg
     {
     public:
         typedef typename MapT::ValueType ValueT;
+        typedef typename MapT::MapType InnerMapType;
```
cfcohen: This might be laziness on my part, but it was easier.

components/nifosg/controller.hpp
```diff
@@ -48,13 +48,14 @@ namespace NifOsg
...
 
         ValueInterpolator()
             : mDefaultVal(ValueT())
         {
         }
 
-        ValueInterpolator(boost::shared_ptr<const MapT> keys, ValueT defaultVal = ValueT())
```
cfcohen: There's an architectural problem with the serializer requiring me to remove const here. :-(   There might be a way to work around it, but it's not immediately obvious to me.

components/nifosg/nifloader.cpp
```diff
@@ -84,82 +84,6 @@ namespace
             collectDrawableProperties(nifNode->parent, out);
     }
 
-    class FrameSwitch : public osg::Group
```
cfcohen: FrameSwitch and BillboardCallback were just moved into the header so that I could reference them from serializers via #include.  It's possible that these nodes don't need to be exported at all, but I haven't figured out how to mark entire nodes as not needing to be exported, so they generate warnings if we don't implement an empty serializer for them.

components/nifosg/particle.hpp
```diff
@@ -30,11 +30,13 @@ namespace NifOsg
         ParticleSystem();
         ParticleSystem(const ParticleSystem& copy, const osg::CopyOp& copyop);
 
-        META_Object(NifOsg, NifOsg::ParticleSystem)
```
cfcohen: Bug?

components/nifosg/particle.hpp
```diff
@@ -194,13 +232,19 @@ namespace NifOsg
         Emitter();
         Emitter(const Emitter& copy, const osg::CopyOp& copyop);
 
-        META_Object(NifOsg, NifOsg::Emitter)
```
cfcohen: Bug?

components/sceneutil/skeleton.hpp
```diff
@@ -39,7 +39,7 @@ namespace SceneUtil
         Skeleton();
         Skeleton(const Skeleton& copy, const osg::CopyOp& copyop);
 
-        META_Node(NifOsg, Skeleton)
```
cfcohen: Bug?

components/serializers/controller.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include <osgDB/InputStream>
+#include <osgDB/OutputStream>
+#include <osgDB/Serializer>
+#include "fixes.hpp"
+
+#include <components/nifosg/nifloader.hpp>
+
+#include "keymap.hpp"
+
+SERIALIZER_KEYMAPT(XRotations, NifOsg::FloatInterpolator::InnerMapType, float, NifOsg::KeyframeController)
```
cfcohen: I really don't care for the macro programming approach here, but that's how OSG does it, and I wasn't inclined to rock the boat too much here.   I think if I were rewriting the serializer from scratch, I would have tried to do it all with templates, but maybe there's some non-obvious problem why that doesn't work.   The author certainly seems to have known how to use templates.

components/serializers/fixes.hpp
```diff
@@ -0,0 +1,8 @@
+#undef REGISTER_OBJECT_WRAPPER
```
cfcohen: This macro had several problems in my opinion.  The biggest is that the MyClass typedef can't be undone, literally preventing there from being more than one serializer per file. :-(  Apparently in OSG they worked around this by having one serializer source file per class. :-(

components/serializers/keymap.hpp
```diff
@@ -0,0 +1,59 @@
+
```
cfcohen: I think this file best demonstrates everything I don't like about OSG serialization, but what can I do?...

components/serializers/skeleton.cpp
```diff
@@ -0,0 +1,26 @@
...
+#define MyClass SceneUtil::Skeleton
+REGISTER_OBJECT_WRAPPER(SceneUtil_Skeleton_Serializer,
+                        new SceneUtil::Skeleton,
+                        SceneUtil::Skeleton,
+                        "osg::Object osg::Group SceneUtil::Skeleton")
+{
+#ifdef SERIALIZER_DEBUG
+    std::cout << "Setting up SceneUtil::Skeleton serializer..." << std::endl;
+#endif
+    // INCOMPLETE
```
cfcohen: Amazingly, because this is a subclass of osg::Group, most of the bones are already exported.  I decided that it wasn't worth going further here until I'd gotten some comments on the overall approach.

cfcohen:
Serious bummer about the Travis build failing. :-(  This builds with GCC 5.2, your OpenSceneGraph fork, and the current master of OpenMW...  I pinky swear...  Since it'll be a pain to recreate the Travis environment more closely to fix this right now, I'll just wait for you to say something before trying to troubleshoot it... :-(

apps/nifconv/nifconv.cpp
scrawl: We need a shared library anyway to make osgconv load our extensions.

components/CMakeLists.txt
scrawl: The serializer namespace has nothing to do with C++ namespaces, correct? In that case an OpenMW serializer namespace would be fine.

components/nifosg/controller.cpp
scrawl: Where is the bug? The second argument is optional and will be value-initialized to osg::Quat() if it's missing. Specifying the argument explicitely doesn't hurt of course.

components/nif/nifkey.hpp
scrawl: No problem with extra set/get functions if the members were public anyway.

components/nifosg/particle.hpp
scrawl: Correct. We can cherry-pick these to master if that makes it easier for you to maintain the patches.

scrawl:
A few things need to happen before we can consider this for inclusion:

- It needs to work with OSG 3.2. Or wait until we drop OSG 3.2 support.
- FrameSwitch needs to be phased out. It is an implementation detail that should not be written to files. The way it works is FrameSwitch has two children, the geometry for use in even frames and a clone of that geometry to use in odd frames. We shouldn't write both to the file. I was actually considering to implement the frame switch within the actual RigGeometry and MorphGeometry because the node switch is a bit unintuitive to use, and caused a regression just the other day.
- Give a bit more time for our OSG extensions to stabilize. We don't want the format in master if it's going to break the next day. I still plan on writing a replacement for osgParticle, for example.

components/serializers/skeleton.cpp
scrawl: Actually, nothing in Skeleton needs serializing. These are mostly cache variables. We could possibly serialize "mActive" but even that is only used at runtime, no point in a model containing a skeleton that's inactive right off the bat.

apps/nifconv/nifconv.cpp
cfcohen: Agreed, and that's working -- at least enough that osgconv and osgviewer will load the files with the custom extensions with out complaining.   They don't always display correctly, and I assume that's because we've not pulled everything need to act on the objects into the OSG render code?

scrawl: Certain things like animations won't start to play until the engine tells what to play, but the rendering should generally work out of the box.

components/CMakeLists.txt
cfcohen: I assume so, but I have not been able to figure this out yet.  Perhaps there's a "feature" in the OSG serializer code such that everything only compiles and runs correctly when the namespaces all match correctly.  This wouldn't be shocking given th number of macros involved, and the duplication of very similar values.  I'll try to play some tricks to get us into one openmw dynamic serializer library without changing any existing code.

components/nif/nifkey.hpp
cfcohen: In several cases I moved the private keyword.   Long term I presume that the preferred solution will be const getters for members that used to be private?  Also there's the issue of whether to create getters and setters in places where we're not 100% required to have them for consistency.

components/nifosg/controller.cpp
cfcohen: I had a compile error in GCC 5.2 that pointed me at this line for some reason.  Maybe I misunderstood.

components/nifosg/controller.hpp
cfcohen: No comment on this?  How serious is this problem?

components/nifosg/particle.hpp
cfcohen: Yes please take these if you're willing.   I've tried to change as little as possible in base OpenMW as possible because I wasn't sure if you'd even want these changes.

components/serializers/skeleton.cpp
cfcohen: This is very surprising to me.   There are custom skeletons in selected NIF files, and someone has to express the relationships between the bones.  Isn't this where that would happen?

scrawl: The bones are child MatrixTransform nodes of the skeleton. The osg::Group serializer already exports child nodes.

components/nifosg/controller.hpp
scrawl: The const is there because in certain cases the same keys might be shared between several interpolators/controllers, so we don't want to modify the keys.

Not a big deal removing the const, we can declare the same intent by adding a comment.

cfcohen:
> It needs to work with OSG 3.2. Or wait until we drop OSG 3.2 support.
> Give a bit more time for our OSG extensions to stabilize.

It sounds like patience is an appropriate approach here.  I'm in no hurry and mostly just wanted to discuss the changes.  Let assume that nothing is being accepted for a while.

> FrameSwitch needs to be phased out.

You had said this before, and I knew it, but I don't know enough to make a change like that in the code that actually affects OpenMW.    There are a number of other custom osg::Objects that don't appear to need serialization or might be serialized more efficiently in some other way.  For example, the BillboardCallback is empty, and so we could serialize a bool anywhere else that's appropriate, and then construct the callback dynamically when the bool is read.  I just don't know enough about the rendering engine to know what the right thing to do here is.  I'd appreciate any specific comments or criticisms in this regard.

Thanks for reviewing this.   Since you didn't reject this completely, I assume that it's roughly the right approach?

components/serializers/skeleton.cpp
cfcohen: Ah.  Understood then.

apps/nifconv/nifconv.cpp
cfcohen: Yes, it's the animations that I'm mostly referring to although I've found a few models that appeared to empty using this code. :-(  Things like billboarding don't seem to work, etc.

scrawl:
> For example, the BillboardCallback is empty, and so we could serialize a bool anywhere else that's appropriate, and then construct the callback dynamically when the bool is read

Where would you put that bool? The BillboardCallback is attached to an osg::MatrixTransform. The osg::MatrixTransform already has a serializer, so we can't change it.

A bool would be more efficient for a single callback, but less flexible w.r.t. adding any number of different callbacks.

> I assume that it's roughly the right approach?

Sure.

components/nifosg/particle.hpp
cfcohen: Actually it looks like the META_Object() macro determines which namespace is used to load the dynamic serializer library, so these might all need to change to OpenMW.  Let me continue to investigate before changing anything.

cfcohen:
OSG serialization determines which dynamic library to load based on the first parameter to the META_Object macro.  It is only used as a string in returning the "library name", and purely by convention it matches the namespace in OSG.  The second parameter in this macro is a name, that must be both a legitimate type and a string.   

These two values must match the values in the values in the third and fourth parameters of REGISTER_OBJECT_WRAPPER.  These are the "class name" and the "associated classes".  

Thus I think this is the right way to set the library and class names.

cfcohen:
I can provide commit b91201c915928d98effa1d20ae6ab2e67ba7daa0 separately if you'll accept it.   I'm also open to some more feedback.  My current plan is to just continue testing and fixing bugs.

scrawl:
Separate PR for b91201c9 would be great.

398bb9f doesn't look useful on its own.

scrawl:
On topic of serializers, I was wondering if it would be feasible to use the OSG serializers for implementing the ShowSceneGraph window. My concern is that printing the whole scene at once would be too slow (especially if vertices are included),  i.e. it would take several minutes to get the window to appear. Essentially I'm looking for a way to serialize each node individually without serializing its children, so that only nodes the user has currently expanded in the tree view are serialized.

Would it be possible to "overwrite" the default OSG serializers? In that case we might be able to overwrite the osg::Group serializer so that it can respect a new option in osgDB::Options controlling the writing of children.

Looking at the osgDB::ObjectWrapperManager, something like this should be possible, but maybe I'm overlooking an issue. Since you have spent some time with the serialization code already, I figured to ask what you think of the idea.

cfcohen:
I'm figuring it out as I go, and I started with OpenSceneGraph last week, so you're asking the wrong person. ;-) I'm not 100% certain I understand what you're proposing, but I've got the general idea I think.

My understanding is that the serializers for OpenSceneGraph are all static methods that are linked into the shared library.  The serializers are not part of the classes themselves.  As a consequence, it seems pretty clear to me that you could overwrite a few functions in that library without changing anything else, and not cause any serious maintenance/interoperability drama.  The relevant source is:

 OpenSceneGraph/src/osgWrappers/serializers/osg/Group

It even looks like there's a special case in osg::Group that invokes an ADD_METHOD_OBJECT that takes an osg::Parameters object that comes from somewhere.   That feature might support something like you're looking for without too much of a change.

I'll have another commit ready tomorrow probably.  I've been debugging various problems and more of the serializers are complete and correct now (or at least they're serializing the fields in objects).  I'm mostly just reading the existing source and systematically serializing everything that I find.  It's a good way to learn about each class if nothing else, but some of the objects clearly should be serialized (Frameswitch, At least one of the StateSetUpdater states, etc.)   I also think I may be duplicating data by not exporting references correctly (or maybe the NIFs did that?).

I'm also going to need some guidance on how to test whether the serialized results are really working however.   I'm thinking that a future step will be to load OSGB files instead of NIFs and see what breaks (or whether anything at all works).

cfcohen:
By the way, I've just started working on serializing SceneUtil::RigGeometry, and I'm curious why none of that uses osgAnimation::Bone, osgAnimation::Skeleton, or osgAnimation::RigGeometry.   Was there something inefficient or wrong with the OSG infrastructure?  Or is it just a case of not having converted fully to OSG yet?

scrawl:
> By the way, I've just started working on serializing SceneUtil::RigGeometry, and I'm curious why none of that uses osgAnimation::Bone, osgAnimation::Skeleton, or osgAnimation::RigGeometry. Was there something inefficient or wrong with the OSG infrastructure? Or is it just a case of not having converted fully to OSG yet?

The problem with osgAnimation::RigGeometry (and most other existing skinning systems) is that they do not support the data in NIF files properly. In osgAnimation, each bone has a bind matrix. But in NIF files, each RigGeometry has a set of bind matrices for the bones it's using. Skeletons can have multiple RigGeometries, so effectively each bone can also have multiple bind matrices (one for each RigGeometry that is using it).

cfcohen:
Now that RigGeometry is partially implemented, I've got work to do on StaticBoundingBoxCallback, GeomMorpherController, and UpdateMorphGeometry, but the commit was growing too large already.  There's a little drama involving the change to osg::Callback that has apparently occurred since OSG 3.2.  Hopefully this will build correctly in Travis.

As always, I'm interested in any feedback you can provide.  Have you been able to run any of this successfully?  Am I on a path to eventual acceptance here?  Because if not, the commit is beginning to touch a lot of stock OpenMW files. :-(   Mostly for little things like const keywords, getters and setters, but also the larger changes of moving declarations from the cpp files to the hpp files.  If you're inclined to eventually accept the request, I can just keep doing what I'm doing.  If not I'd like to discuss which parts could be split into a separate pull request.

For much of what I'm serializing, I don't fully understand how it's actually rendered, so I'm guessing about what needs to be serialized and what doesn't.   I'm not worried about serializing a few things needlessly (which can easily be removed) or missing a couple (which can be added), but if I'm significantly blowing it, I'd appreciate knowing sooner rather than later.  ;-) I'm also assuming that once things are mostly serialized you'll know what to do to check whether it's _really_ working.  I hope it's obvious, my intention here is to document the native OSG representation of the OpenMW extensions so that the Blender exporter can produce files with these extensions, allowing a full range of content to be created without NIFs.  There has been several discussions about this on the forums, and I assume that's still a desired feature.

I'll be unable to do any work on this from Dec 18-30 (Christmas vacation with family away from real computers).

cfcohen:
I continue to be unhappy with the places where I've moved members from private to public (required so that they can be access by the static serializers).  My preferred alternative is to define const getters and setters so that the exposed member is at least const when it's returned to the caller.  I've done this in a few places where I had to in order to use the add serialzier macros.  I don't see much that can be done about the need for for a setter.  Perhaps a friend class?  That seems like overkill to me.

This is a code style religious issue, and right now I'm doing the thing that's easiest to merge (moving to public).  I'd like guidance on what you really want done here, or whether I should continue to keep the merge as small as possible.

scrawl:
MorphGeometry will be awkward to handle because it's a bit of a mishmash - osgAnimation with some extensions hacked on top. We could fork to SceneUtil::MorphGeometry if we need more flexibility in how it's set up and serialized.

> For much of what I'm serializing, I don't fully understand how it's actually rendered, so I'm guessing about what needs to be serialized and what doesn't

A simple rule to follow would be: anything set from outside of the class definitely needs serializing. Things that are only set internally, or reset every frame anyway, don't need serializing (cache variables).

Some classes really need a rework first to be more serializer-friendly. For example, currently RigGeometry exports its own geometry (which is a clone of the source geometry), and then exports the source geometry too. We are basically writing the exact same geometry twice. Then there's the issue of exporting RigGeometry's vertex/normal arrays, which really shouldn't be exported at all since they're repopulated every frame anyway.

I'm really not happy with the serializer exporting implementation details, like the default update callbacks. But I can't think of an easy way around that :(.  If we remove the exported callbacks from the .osg file, does it still work properly? The callbacks should be set by default when the object is constructed, but maybe the loader overwrites those.

I'm starting to think for certain classes we might want to ignore certain inherited class serializers (e.g. don't invoke the Node serializer, so we don't write callbacks). Looking at way the serializer sets the class hierarchy in a string, I'm thinking it should be possible to just leave out one class? We might then have to serialize certain fields manually that used to be serialized by the parent class.

Addition of get/setters is fine. I've only skimmed over the changes, but what I've seen looks totally reasonable. public members are ok as temporary solution if that makes the merges easier.

>  Am I on a path to eventual acceptance here?

Yes, it's just we still have a long way to go.

cfcohen:
> Some classes really need a rework first to be more serializer-friendly.

I agree. But that's almost certainly too big for me.  My plan was to get as far as a I could without those kind of changes, and then see where the code was at.  For example, I'm curious if I can mostly play OpenMW with OSGB objects, which I haven't even tried yet.

> We are basically writing the exact same geometry twice.

I've learned to serializer osg::ref_ptr objects so that they just reference the other data.   That might be a workaround for now.  I've also just not worried about size/performance at all right now.

> I'm really not happy with the serializer exporting implementation details, like the default update callbacks.

I think the solution here is to create the appropriate callbacks in the constructors for the parent classes when possible.  For example, I just implemented the serializer for UpdateMorphGeometry, which has no members except a cached LastFrameNumber.  I think the alternative is to override osgAnimation::MorphGeometry as SceneUtil::MorphGeometry so that it automatically sets a CullCallback of UpdateMorphGeometry on construction.   Then we figure out how to suppress the warnings for not having a serializer for UpdateMorphGeometry, and it's gone completely from the serializer.   That's assuming that all of osgAnimation::MorphGeometry objects use that callback, which I'm a little uncertain about.  Also sadly, I've not found an OSG standard way to _NOT_ to serialize a class at all yet.   I went to the right place in OSG, and it's just not there. :-(  But that at least would be a clean addition to OSG, and almost certain to not have undesired side effects.

> Looking at way the serializer sets the class hierarchy in a string, I'm thinking it should be possible to just leave out one class?

I've got some core dumps for adding incorrect values (which aren't always easy to determine).   I've been reluctant to leave out values for fear that it would cause unexpected problems in OSG.   In particular, the ReadSerializer is very bad about expecting what it expects and not being terribly forgiving.

> Yes, it's just we still have a long way to go.

Agreed, and that's good enough for me right now.  I just wanted to make sure you weren't thinking "What the hell is he doing?" :-)

components/serializers/planarcollider.cpp
```diff
@@ -0,0 +1,17 @@
...
+REGISTER_OBJECT_WRAPPER2(NifOsg_PlanarCollider_Serializer,
+                         new NifOsg::PlanarCollider,
+                         NifOsg::PlanarCollider,
+                         "OpenMW::PlanarCollider",
+                         "osg::Object osgParticle::Operator OpenMW::PlanarCollider")
+{
+    SETUPMSG("OpenMW::PlanarCollider");
+    ADD_FLOAT_SERIALIZER(BounceFactor, 0.0f);
+    ADD_PLANE_SERIALIZER(Plane, osg::Plane());
+    ADD_PLANE_SERIALIZER(PlaneInParticleSpace, osg::Plane());
```
scrawl: Don't serialize PlaneInParticleSpace, it's set internally every frame.

components/serializers/skeleton.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include "serializer.hpp"
+
+REGISTER_OBJECT_WRAPPER2(SceneUtil_Skeleton_Serializer,
+                         new SceneUtil::Skeleton,
+                         SceneUtil::Skeleton,
+                         "OpenMW::Skeleton",
+                         "osg::Object osg::Group OpenMW::Skeleton")
+{
+    SETUPMSG("OpenMW::Skeleton");
+    // INCOMPLETE
```
scrawl: This is actually complete, nothing in Skeleton needs serializing except for children which are already handled by osg::Group serializer.

components/serializers/statesetupdater.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include "serializer.hpp"
+
+REGISTER_OBJECT_WRAPPER2(NifOsg_StateSetUpdater_Serializer,
+                         new SceneUtil::StateSetUpdater,
+                         SceneUtil::StateSetUpdater,
+                         "OpenMW::StateSetUpdater",
+                         "osg::Object osg::NodeCallback OpenMW::StateSetUpdater")
+{
+    SETUPMSG("OpenMW::StateSetUpdater");
+    // No serialization for: osg::ref_ptr<osg::StateSet> mStateSets[2];  Transient?
```
scrawl: > Transient?

Correct.

cfcohen:
I've been testing, and have some progress to report.  I have a hacked up version of OpenMW that will load and run with _only_ the following NIFs (no BSA archives, all converted OSGB files):

meshes/xbase_anim.1st.nif
meshes/xbase_anim_female.nif
meshes/xbase_animkna.nif
meshes/xbase_animkna.1st.nif
meshes/xbase_anim.nif

OpenMW will start successfully, but the player is unable to move without the base animations installed, which makes testing the other assets difficult.   What works:

Pretty much all statics (duh!)
Candle flame animations!
Doors open and close!
GUI, containers, dialogs
Cell loading, lots of interiors and exteriors near Balmora 
Mouths move (but eyes don't appear to blink)
Magically enchanted clothes shimmer (shader?)
NPC Animations (uses xbase_anim, correct?)

What doesn't (not terribly shocking):

Monster animations
Pretty much all spell effects (particles)
Probably lots more I haven't tested yet

I realize that most of the important/difficult stuff is still broken, but I'm also confident that I'm reading a lot of the custom data back into the game because I've enabled debugging in the serializers, and can see it being read.  There are no serialization errors, and all the serializers are enabled (I had to link them statically).

It appears that the remaining problems don't cause crashes (hooray!).  And there's not usually a ton of spew to the console with the exception of messages about missing banner idle animations.  There's a bunch of banners in Balmora, and they all cycle every second or two. :-(  Investigating that led to one of my primary questions for you:

The missing animations (really TextKeyMap entries) are for Idle1, Idle2, and Idle3.  Those are apparently stored in the KF file, and not the NIF file.  I hadn't even thought about the KF files :-(.  What would have to happen to get those entries serialized into the OSGB file?  I assume that the problem is that I didn't load them when I serialized the NIF, and I didn't know to load them with the OSGB either (probably signaled in the NIF file in nifloader?)

As for the other problems (especially particle effects) the surprising part is that I'm getting no errors at all.  I get the first person animations, and the game effect of the spell, but no particles. Any suggestions on where to instrument to confirm that I'm even trying to draw particles?  Is there any place that I can dump some data to confirm that I'm getting close, and not just loading a bunch of statics?  This is where I'm completely out of my league, and will need some more hints.

scrawl:
> meshes/xbase_anim.nif
> 
> OpenMW will start successfully, but the player is unable to move without the base animations installed

The ```xbase_*.nif``` don't contain animations (those are in the kf file). The animations are in separate files because they're merged at runtime in some cases (e.g. argonians get animations from xargonian_swimkna.kf in addition to xbase_anim.kf, females additionally get xbase_anim_female.kf, etc). 

We'll need to convert the kf files to .osg files (with some added filename pattern so we know that it's keyframes not a model, e.g. xbase_anim_kf.osg). The loaded equivalent of a kf file is the NifOsg::KeyframeHolder. I just pushed a change to inherit KeyframeHolder from osg::Object instead of osg::Referenced (part of another feature I was working on). Maybe that makes it easier to serialize?

Then we'll need to add code in Animation::addAnimSource and KeyframeManager::get to actually load the osg keyframe files.

> As for the other problems (especially particle effects) the surprising part is that I'm getting no errors at all. I get the first person animations, and the game effect of the spell, but no particles.

We appear to be missing serialization for Controller, ControllerFunction and ControllerSource. The ControllerFunction is what tells the game how "long" an effect lasts, so if it's missing the effect will probably not play at all (see the FindMaxControllerLengthVisitor).

cfcohen:
> We'll need to convert the kf files to .osg files (with some added filename pattern so we know that it's keyframes not a model, e.g. xbase_anim_kf.osg).

I was just starting to look into this as you posted.  Thanks for the assistance.  So this means that we can't merge the kf into the osgb at conversion time, because we wouldn't be able to "merge" results at runtime, correct?  I already found the code to load xMODEL.nif and converted it to load xMODEL.osgb instead.  Is there similar code for the kf files? (I'm about to go look for that code now).

> The loaded equivalent of a kf file is the NifOsg::KeyframeHolder.

I'll pull that change and start working on it.

> I just pushed a change to inherit KeyframeHolder from osg::Object instead of osg::Referenced (part of another feature I was working on). Maybe that makes it easier to serialize?  ... We appear to be missing serialization for Controller, ControllerFunction and ControllerSource. 

Actually, I've not serialized those classes because as far as I know it's not possible to serialize objects that don't inherit from osg::Object.  There are hard-coded dynamic casts in the serializer code.  I've been meaning to ask you where those were intended to be serialized or not, and whether converting them to Objects was acceptable.   I don't want to break other things, and would be grateful if you'd think through that part.   Alternatively, I can just add osg::Object as a parent, and call the appropriate META_Object macro if that's sufficient.

scrawl:
> as far as I know it's not possible to serialize objects that don't inherit from osg::Object

Ah, that's a shame. The osg::Object adds some unneeded baggage (_name, _userDataContainer, and _dataVariance). As well, it increases size of the class hierarchy which could make the ```dynamic_cast<Controller*>``` slower. But if that's the only way to serialize then I guess we have no choice.

> I've been meaning to ask you where those were intended to be serialized or not, and whether converting them to Objects was acceptable. I don't want to break other things, and would be grateful if you'd think through that part. Alternatively, I can just add osg::Object as a parent, and call the appropriate META_Object macro if that's sufficient.

Not sure if I get this correctly. By "add osg::Object as parent", you mean inheriting Controller from Object? Then what is meant by "converting them to Objects"? Using a proxy osg::Object that <i>contains</i> the Controller? It seems cleaner to me to inherit Controller from Object.

With Controller an osg::Object, does the multi inheritance still work correctly? I.e. can we have a class inherited from both Controller and StateSetUpdater (which are both osg::Object's, respectively)? Perhaps the Controller <- Object inheritance needs to be virtual.

scrawl:
> I was just starting to look into this as you posted. Thanks for the assistance. So this means that we can't merge the kf into the osgb at conversion time, because we wouldn't be able to "merge" results at runtime, correct?

Correct.

> Is there similar code for the kf files? (I'm about to go look for that code now).

Animation::addAnimSource and KeyframeManager::get

cfcohen:
> Ah, that's a shame. The osg::Object adds some unneeded baggage...

In particular, the libraryName() and className() methods defined by the META_Object macro appear to be used in the serialization code to determine which dynamic libraries to load, which serializers to invoke, etc.  It could still be possible to implement part of osg::Object and somehow hack it up, but given how fragile the serialization has been in general, I'm not inclined to try.

> It seems cleaner to me to inherit Controller from Object.

You're correct -- I hadn't thought through it in much detail yet.  I didn't mean to imply anything more complicated.

> With Controller an osg::Object, does the multi inheritance still work correctly?

I don't know.  The only thing that I've figured out in that regard is that I need to define OBJECT_CAST to dynamic_cast for any classes that are virtual, and that OSG doesn't currently seem to care if there are other virtual classes (e.g. SceneUtil::Controller) so long as osg::Object is a valid dynamic cast.  To my knowledge, we've never tested the multiple inheritance case.

cfcohen:
This batch of pushed commits stretches back to before some of our recent conversation (to the last failed travis build).  

The hackish changes to OpenMW to enable OSGB loading require one other change that's not committed because I'm struggling with cmake issues.  The serializer objects need to be linked statically into the openmw executable, because it doesn't load the serializers dynamically like osgconv and osgviewer do... I'm not sure why, and I didn't want to aggravate my travis problems further by changing the CMakeLists.txt again right now.   OpenMW will crash if you don't relink it with the extra objects.

On the topic of CMake, the problem with the serializers being in the components directory is completely inexplicable to me.  I think it may have something to do with the custom add_component_dir macros.  It literally just leaves nifloader.cpp.o out of the link command for no reason that I can see. :-(

kcat:
> We'll need to convert the kf files to .osg files (with some added filename pattern so we know that it's keyframes not a model, e.g. xbase_anim_kf.osg). The loaded equivalent of a kf file is the NifOsg::KeyframeHolder.

I wouldnt necessarily worry about this just yet (unless it's the last thing to do?). The way Morrowind handles animations is really bad, and I'd prefer to look into other ways to add in anaimations from native formats.

cfcohen:
> The way Morrowind handles animations is really bad, and I'd prefer to look into other ways to add in animations from native formats.

I agree strongly. :-)  But I've also assumed that step one is understanding how the existing code works, since it actually works. ;-)  I keep expecting to find some insight into why the NIF animations are so convoluted, but that hasn't happened yet.  My question to Scrawl earlier about why the NIF implementation didn't use the stock osg::RigGeometry was unsatisfying, and I still don't really understand.  But I'd like to support better, simpler, more standard animations in the OSGB format as well.

scrawl:
> The way Morrowind handles animations is really bad, and I'd prefer to look into other ways to add in anaimations from native formats.

We want the OSG format to be interoperable with NIF's, so you can export default game assets to OSG and then edit them. For that to happen we need to support NIF animations.

I agree the NIF animation system is a bit complex and convoluted. Actually we could use osgAnimation for new files (not converted from NIFs), which are [partially?] supported in the blender exporter already. Unfortunately, osgAnimation by default is very slow because it a) sets meshes to dynamic data variance, which trashes OSG's default threading mechanism, and b) animates vertices of meshes that are not in the viewing frustum. We probably want to implement something smarter than that before using osgAnimation as the default format for new files. Also, we'd still need to figure out how to have the game recognize osgAnimations (the serializer loads the animation nodes, but there's nothing in place yet to tell the game that the object contains animations and when to play these animations).

scrawl:
Having two different animation systems could be problematic as far as mods are concerned, with the way that equipment parts interact with the main skeleton. E.g. if the main character skeleton is a NIF, you wouldn't be able to attach a robe to it that uses osgAnimation, since osgAnimation bones are not compatible with NIF bones. For this to work out <i>all</i> the character assets would need to be built on osgAnimation.

scrawl:
>  My question to Scrawl earlier about why the NIF implementation didn't use the stock osg::RigGeometry was unsatisfying, and I still don't really understand.

Hah, I didn't fully understand that when I started working on animations either :) It all sort of falls into place when you have code in front of you that doesn't work, and you slowly realize <i>why</i> it doesn't work.

scrawl:
>  My question to Scrawl earlier about why the NIF implementation didn't use the stock osg::RigGeometry was unsatisfying, and I still don't really understand.

Sorry, I don't know how to explain this better. Reading up on skeletal animations might help if you didn't know what a bind matrix is.

Here are two posts from the OSG port thread, right before I ditched osgAnimation.
https://forum.openmw.org/viewtopic.php?f=6&t=2775&start=10#p31811
https://forum.openmw.org/viewtopic.php?f=6&t=2775&start=20#p31824

scrawl:
Well, here's a diagram if it helps.

How osgAnimation works:

<pre>
Skeleton
-Bone A
--Bind matrix for Bone A
--Bone B
--- Bind matrix for Bone B
--- RigGeometry 1
--- RigGeometry 2
</pre>

How NIF animations work:
<pre>
Skeleton
-Bone A
--Bone B
---RigGeometry 1
----Bind matrix 1 for Bone A
----Bind matrix 1 for Bone B
---RigGeometry 2
----Bind matrix 2 for Bone A
----Bind matrix 2 for Bone B
</pre>

cfcohen:
> It all sort of falls into place when you have code in front of you that doesn't work, and you slowly realize why it doesn't work.

That's effectively what I'm doing now.  ;-)  Thanks for trying to explain it more, and I'll read those posts and review the code more carefully.  I mostly meant that the response was unsatisfying because I have more to learn, not because you didn't explain it well.   And as for compatibility, what I'm really looking for is a better (possibly incompatible) animation system for new games developed in the OpenMW engine, but that's father in the future.

We've developed another problem. :-(   I don't think it's a major one, but it's mildly annoying, and I don't know what the best solution is.   Basically, because NifOsg::KeyframeHolder is an osg::Object and not an osg::Node, when I run osgconv on the serialized version of a KF file, it produces the error "Error no data loaded."  I've looked at the osgconv code, and it seems to have expected some renderable geometry (which it didn't find) so we'll hack something up if we want osgconv to convert the KeyframeHolders from osgb to osgt and vice-versa.  We could wrap the KeyframeHolder in an osg::Group or something, or produce our own version of osgconv that isn't so strict, or maybe something else.

I've got xxx.kf into xxx_kf.osgb serialization working, and I will be extending the KeyframeManager to try and load those files instead later this evening.   I hope to push again this evening, but I'm stepping away for a while.

scrawl:
>  We could wrap the KeyframeHolder in an osg::Group

Sounds like a reasonable workaround. You can store any Object in a Node's UserDataContainer. Similar to how NIF objects have their TextKeyMapHolder stored in the root node's UserDataContainer.

cfcohen:
Animations are mostly working! :-)  OpenMW runs and is largely correct with no NIFs or KFs except xbase_anim.nif and xbase_anim.kf.   What's working:

Banners flap in the breeze in Balmora with no NIF/KF support!
First person animations are generally working.
Most monster animations are mostly working (see caveats below).

It looks like the primary remaining issues are:

No player movement without the two files above (or maybe just one?).  Argonian NPCs are animated, but unable to actually move (they run in place.  This happened when I removed the xbase_animkna.[nif|kf] files, and I assume it demonstrates the same problem the player has when xbase_anim.[nif|kf] is removed.  As for monster animations, they idle normally and, can also attack, run, cast spells, etc.  But very frequently they just disappear.   It looks like this is an object placement interaction with the animation, because sometimes I can see them fall through the ground.  The most obvious case is the cliff racer, which functions almost normally, but comes down until it's in the ground to to fight.  I presume that because it can fly, it doesn't fall through the ground and pass into the void like most other monsters.

Your guidance on addAnimSource in particular was _exactly_ what I needed.  It may have taken me a while to figure that out without your help, so thanks!  Particle effects for the player are still not working (although I did see a frost atronach start to cast a spell before he disappeared).  Any suggestions on where to look for this problem?

I've not implemented the UserDataContainer workaround for the KeyframeHolder, but I'll try that tomorrow.  The build just failed on OSG 3.2, but it works on OSG 3.3.   I may look into that tonight. 

You haven't said explicitly whether you've tested this at all yet.  If you're ready to here's what I did:
Build.  Install libOpenMW.so and osgdb_serializers_openmw.so into osgPlugins-3.5.1.  Don't forget to remove the leading lib on the serializer library.  Then run "nifconv --all" which should read your current configuration and export all NIFs and KFs into ./meshes/.   Then I ran bsatool to extract all the archives, removed the BSAs, NIFs and KFs, and copied the OSGB files into the meshes directory.  I removed the BSAs from my openmw.cfg fallbacks.  Finally, I relinked openmw with serializers/*.o, and ran it.   The hacked up code in OpenMW should use a NIF or a KF if it exists, and load the OSGB version if it does not.

Please offer any suggestions you have on how to proceed or what to do next...

scrawl:
> As for monster animations, they idle normally and, can also attack, run, cast spells, etc. But very frequently they just disappear. It looks like this is an object placement interaction with the animation, because sometimes I can see them fall through the ground. The most obvious case is the cliff racer, which functions almost normally, but comes down until it's in the ground to to fight. I presume that because it can fly, it doesn't fall through the ground and pass into the void like most other monsters.

Ah, missing collision shapes. That's another can of worms we haven't opened yet. NIF files contain collision shapes, the NifOsg loader doesn't handle these. The NifBullet loader does.
For static objects, the absence of collision shapes isn't a huge problem (the loader currently auto-generates the collision shape from the rendering data when loading a .osg file). But actors really need a dedicated collision shape (determined from the Bounding Box node specified in the NIF file) for their movement to work properly. Some objects also contain animated collision shapes (e.g. the doors in Sotha Sil, Dome of Serlyn), where a KeyframeController moves the shape around.

I haven't really thought about handling collision shapes yet. I would be inclined to separate the rendering data from the collision data, e.g. by using the BulletWorldImporter's serialization format to specify shapes, but we'll also need to add some sort of meta-data specfying which shape(s) to use and whether or not they should be animated. This metadata could be in the .osg file or in another file.

scrawl:
> No player movement without the two files above (or maybe just one?). Argonian NPCs are animated, but unable to actually move (they run in place. This happened when I removed the xbase_animkna.[nif|kf] files, and I assume it demonstrates the same problem the player has when xbase_anim.[nif|kf] is removed. 

Appears to be another collision issue (the xbase_anim.nif contains the bounding box for the actor collision shape, so with it removed the MWPhysics::Actor can't be constructed correctly).

> You haven't said explicitly whether you've tested this at all yet. If you're ready to here's what I did:
Build. Install libOpenMW.so and osgdb_serializers_openmw.so into osgPlugins-3.5.1. Don't forget to remove the leading lib on the serializer library. Then run "nifconv --all" which should read your current configuration and export all NIFs and KFs into ./meshes/. Then I ran bsatool to extract all the archives, removed the BSAs, NIFs and KFs, and copied the OSGB files into the meshes directory. I removed the BSAs from my openmw.cfg fallbacks. Finally, I relinked openmw with serializers/*.o, and ran it. The hacked up code in OpenMW should use a NIF or a KF if it exists, and load the OSGB version if it does not.

Yes, that worked (apart from not being able to move). 
I've noticed entering "Sotha Sil, Outer Flooded Halls" results in a crash, though most other cells I've tested work fine.

scrawl:
> I haven't really thought about handling collision shapes yet. I would be inclined to separate the rendering data from the collision data, e.g. by using the BulletWorldImporter's serialization format to specify shapes, but we'll also need to add some sort of meta-data specfying which shape(s) to use and whether or not they should be animated. This metadata could be in the .osg file or in another file.

The cool thing about using BulletWorldImporter would be that it's already more powerful than NIF's. NIF's only support triangle shapes, which are not the most optimal shape for a physics engine (slow to process, and can cause glitches if the triangles are too big). BulletWorldImporter supports all the various primitive shapes that bullet supports (e.g. boxes, cylinders, cones, spheres, multi-spheres, plus you can do any combination of any number of primitives in a btCompoundShape). AFAIK there's support in Blender for editing collision shapes and exporting them to the .bullet format.

scrawl:
> The cool thing about using BulletWorldImporter would be that it's already more powerful than NIF's.

Plus it supports assiging values like mass and center of mass for a btRigidBody, which will become relevant when we implement support for non-static rigid bodies (a.k.a. flying cabbages).

scrawl:
> Particle effects for the player are still not working (although I did see a frost atronach start to cast a spell before he disappeared). Any suggestions on where to look for this problem?

See:

> We appear to be missing serialization for Controller, ControllerFunction and ControllerSource.

cfcohen:
Sorry for the slow ack, I've been working (not on OpenMW, but to pay the mortgage) most of the day. :-(

> Ah, missing collision shapes.

That's sufficient to point me in the right direction for a few days.  I'll learn some more about Bullet over the holidays, and see about exporting/importing a model.bullet file using BulletWorldImporter, from whatever NifBullet currently does.  Sounds like a great way to dip my toe into the physics code.   I assume we can decide from there if there's any merit to trying to incorporate that data blob into the OSG file.

>... so with it removed the MWPhysics::Actor can't be constructed correctly...

Because this is special case stuff anyway, it seems like maybe we should do something by default that ensure that there's _always_ a valid MWPhysics::Actor that can move.  Or is this way more complicated than I realize?  I haven't looked yet, but I will unless you wave me off this idea.

> We appear to be missing serialization for Controller, ControllerFunction and ControllerSource.

Thanks for the reminder.  I had obviously forgotten about that conversion, but now that I've been reminded I think I know how to proceed, and I will probably tackle that next.

> Yes, that worked (apart from not being able to move). 

I'm glad that you were able to get it to work as well.  That suggests that it's not a fluke that it seems to work for me. 

> Try this (should work for 3.2 and trunk)

Will do so soon.

> I've noticed entering "Sotha Sil, Outer Flooded Halls" results in a crash...

I'll investigate this as well (at some point).

Thanks again for the feedback and guidance.

cfcohen:
> Try this (should work for 3.2 and trunk):
>    osg::ref_ptr<NifOsg::KeyframeController> ctrl;
>    is >> ctrl.get();

This failed for 3.3 at least, with:

error: no match for ‘operator>>’ (operand types are ‘osgDB::InputStream’ and ‘NifOsg::KeyframeController*’)

A variant with "is >> *(ctrl.get());" fails similarly.   I'll reconfigure for the OSG 3.2 version at some point, and flail until I figure it out.  It appears that there was a problem that lead to the 3.3 introduction of readObjectOfType<T>.   There's a nearly identical case in riggeometry.cpp:110 that works, presumably because osg::Geometry has something that NifOsg::KeyframeController lacks.  I'll try to figure out what.

cfcohen:
I have SceneUtil::Controller partially serialized and I'm able to write NifOsg::ControllerFunction members (Frequency, Phase, StartTime, StopTime, and ExtrapolationMode) out to an OSGB file as part of a Controller.   It's a bit hacked up since I dynamic cast the SceneUtil::ControllerFunction into NifOsg::ControllerFunction in the SceneUtil::Controller serialization code.  This is wrong because it makes presumptions about the subclassed type of mFunction pointer, but it works in practice because the NifOsg variant is the only implementation (that I know of) right now.  I'm not convinced that the OSG serialization library knows how to dynamically invoke the correct serializer, especially when the local code doesn't know the true type of the pointer, but I haven't tried very hard either.  Of course there's no super easy way for me to know the true subclass either.

At the very least, the boost::shared_ptr<> for mSource and mFunction will need to be replaced with osg::ref_ptr<>, and mSource and mFunction will need to inherit from osg::Object as well.  And thus so will all of their subclasses (implicitly).  If OSG doesn't do the right thing automatically, we may have to add some virtual methods on the classes to return their dynamic type (probably the fake OpenMW::XXX one) to assist OSG in deciding how to serialize.  Because of the namespace games we've played, I think there's very little chance that OSG can just figure it out.   This is starting to become a decent sized cascading change.

Alternatively, we could assign each subclass of SceneUtil::Controller responsibility for serializing their mSource and mFunction members, and then SceneUtil::Controller would no longer need to inherit from osg::Object, eliminating the overhead you mentioned earlier.  This would require that SceneUtil::Controller be pure virtual, and that all subclasses have their own serializers, but that doesn't seem like a huge problem to me.  Additionally, each of those subtypes would need to know what type of mFunction they used, and not allow arbitrary subclasses, or we end up back in the first scenario.  In other words, the NifOsg controllers would always serialize a NifOsg::ControllerFunction.   I don't know if this is a problem currently or for intended future capability.  If it is, we could use a shared piece of code a switch on the type to hack around this.

Finally, the ControllerSource objects all appear to be completely dynamic and differentiated entirely by their getValue() implementations.  If this is true, they would probably be best serialized as simply the name of the type of ControllerSource to use.  Then the serialization reader would have a little switch statement that would read the string, and invoke the appropriate default constructor.  This would free SceneUtil::ControllerSource and all of it's children from being osg::Objects in addition to reducing the serializer complexity.  I'm currently inclined towards this approach.

Obviously, there's pros and cons to both approaches, and wanted to let you weigh in which you preferred before I went any further.  Plus it's 1am, and I have to work tomorrow. ;-)  Please let me know which approach you'd prefer (or a combination of their features), and I'll try to make that work.

scrawl:
> At the very least, the boost::shared_ptr<> for mSource and mFunction will need to be replaced with osg::ref_ptr<>, and mSource and mFunction will need to inherit from osg::Object as well. And thus so will all of their subclasses (implicitly).

That seems ok to me. Let's not try to be extra smart. Being consistent with the other serializers is more important.

> Finally, the ControllerSource objects all appear to be completely dynamic and differentiated entirely by their getValue() implementations. If this is true, they would probably be best serialized as simply the name of the type of ControllerSource to use.

Isn't that essentially what the serializer does when you tell it to serialize an empty object of type X?

cfcohen:
> That seems ok to me. Let's not try to be extra smart.

Ok.  I'll try that.  I also just realized how OSG might be smarter than I gave it credit for.

> Isn't that essentially what the serializer does when you tell it to serialize an empty object of type X?

Yes, but with a bunch of code to oermit OSG to do so.  Also there's one important detail of how that happens that I still don't understand.  I'll try to figure that out this evening.   Thanks, that guidance was sufficient to proceed.

cfcohen:
Overall, these commits appear to have resulted in working particle effects! :-)  There are some problems with placement (like fireballs appearing over the target rather than on the target) but I suspect this is caused by the bounding box problems we've discussed already.

Obviously the scope of the changes grew quite a bit because of the osg::ref_ptr conversion, but in the end it all worked out, and I think this is the logical conclusion of what you recommended.  I'm also pleased with it, and I got the object serializer to correctly detect subclasses using the macro. In addition to the several issues commented on specific lines of the commit, there are a few previously undiscussed issues as well.   

First, there are several serialization warnings about unsupported NiPathController objects in the osg tree.  NiPathController is in the Nif namespace, not the NifgOsg namespace and does not inherit from osg::Object, so it can't be serialized.   I assume that the solution is to create a similar class in NifOsg, and construct it from the Nif namespace object like many of the other classes in NifOsg.   If you agree, I'll do that.

Second, there are several subclasses of SceneUtil::ControllerSource that have not been serialized.  Specifically, EffectAnimationTime, AnimationTime, NullAnimationTime, WeaponAnimationTime and HeadAnimationTime.   These are all defined in openmw/mwrender and thus create some questions about class reuse.  Are they all transient?  Do they need to be serialized?  They don't appear in NIFs, so what creates them?

Sorry about the failing Travis builds.  Downgrading to OSG 3.2 and resolving those issues is the next item on my todo list.

scrawl:
> First, there are several serialization warnings about unsupported NiPathController objects in the osg tree

This warning is from the NIF loader, not from the OSG serializer. NiPathController are not created in the osg tree because we don't yet know what they do. Nothing to do on the serialization front.

> Second, there are several subclasses of SceneUtil::ControllerSource that have not been serialized. Specifically, EffectAnimationTime, AnimationTime, NullAnimationTime, WeaponAnimationTime and HeadAnimationTime.

OSG extensions defined in apps/ do not need serializing. We only need to worry about the model loading that happens in components/.

cfcohen:
More extensive testing of OSG 3.2 has revealed a problem.  Specifically the commit: https://github.com/openscenegraph/osg/commit/d7cf8d9395b29de38f1545c94ff3039aebe6bbde not only added serialization capability for osg::ComputeBoundingBoxCallback, it modified osg::RigGeometry to actually use the callback.  Without this change (which may be difficult to back port) it looks to me like OSG 3.2 is just broken. :-(

cfcohen:
Also, I should mention that the serialization for OSG 3.2 and OSG 3.5 are substantially different.  This prevents the models exported under one version from being used in the other version.  I had thought this was caused by OSG differences, but I see that the primary cause is the #if(OSG<3.3.3) at nifloader.cpp:853 and related #ifdefs.   This detail being exposed through serialization makes the files less portable than is probably desired.  Can you comment on why these conditionals are present for OSG 3.2?   Can we serialize only the osg::Geometry inside the Geode?  Is this another reason to abandon support for OSG 3.2?

scrawl:
OSG 3.4 added the ability to place osg::Drawable directly in the scene graph (as an osg::Node), whereas in 3.2 and earlier you had to use an osg::Geode node to which the drawables are attached. OSG 3.4 retains backwards compatibility in that you can still use osg::Geode's, but doing so is slightly slower than directly attaching the drawables to the scene graph (not by much, but measurably so). That is why we don't use the Geode's when compiled with 3.4.

cfcohen:
Should I try to only serialize the geometry (not the Geode) from an OSG 3.2 export, and then when loading it back in again, wrap what's loaded in an osg::Geode if we're running under OSG 3.2, or is that not worthwhile?  If we're committed to OSG 3.2 for very long, it would allow the same file to load under either version.  If we're not, it seems like a lot of effort for nothing.

scrawl:
> and then when loading it back in again, wrap what's loaded in an osg::Geode if we're running under OSG 3.2

Sounds complicated. I think you would have to override the default osg::Group serializer, and I'm pretty sure that osgconv would stop working.

How about we add a new option to the NifLoader: ```bool mBackwardsCompatible```. When this option is set, the loader will wrap Geometry in a Geode even when compiled against OSG 3.4. This new option would be 'off' by default, but set to 'on' in nifconv. The exported file will then work with both 3.2 and 3.4. As I said, using Geodes in 3.4 isn't 100% optimal from a performance standpoint, but we can worry about that later. Get it working first, then optimize.

cfcohen:
> Sounds complicated. I think you would have to override the default osg::Group serializer, and I'm pretty sure that osgconv would stop working.

Maybe I misunderstand.  For export in OSG 3.2, I propose roughly:

writer->writeObject(geode->getDrawable(0));    

For import in OSG 3.2:

drawable = reader->readNode(file);
geode = new Geode();
geode->addChild(drawable);
return geode;

Is the problem that with out an enclosing Geode that that the 3.2 osgconv would consider the serialization invalid?  If so, I can test that.  The API allows the serialization of individual osg::Objects, so it's not clear to me that this wouldn't work.  

> How about we add a new option to the NifLoader: bool mBackwardsCompatible. 

This could work as well, although I'd prefer not to have the extra Geode obviously.  Not so much because of performance, but for clarity and simplicity.  Neither of these approaches address the bug in RigGeometry however.

In my most recent local commits, I'm able to do several new things, including creating a "default" actor bounding box.  I set it to a btBoxShape, and even without xbase_anim.nif, the player is able to move, which is an improvement over being immobile.  I'm also able to load OSGB geometry (in the code labeled "TODO support .bullet shape files") and use NodeToShapeVisitor to construct a BulletShape, although I'm not sure to what effect, since animated creatures still fall through the ground when moving.  The default translation appears to have correct some problems with positioning however.

Finally, I've modified nifconv to process all NIFs with BulletShapeManager::createInstance() and I've serialized them using btDefaultSerializer.  With Tamriel Rebuilt and Better Bodies, there are 9285 non-compound shapes and 493 compound shapes.  There are 453 with mAnimatedShapes.size() > 0 and 842 with no mCollisionShape.  I figured the next step was to read them back in again as .bullet files instead of NIF files and see if it fixes creatures.

I think I'm beginning to understand what you meant when you said:

> NIF's only support triangle shapes, which are not the most optimal shape for a physics engine (slow to process, and can cause glitches if the triangles are too big).

It appears that the bullet collision shapes are currently just copies of the geometry.  I was under the impression that the NIFs had separate reduced polygon count geometry for the collision boxes.  And also you wrote:

> ... but we'll also need to add some sort of meta-data specfying which shape(s) to use and whether or not they should be animated. This metadata could be in the .osg file or in another file.

Which I'm still confused by.  If you're able to decide how this should work over the holiday, I'd be very interested in hearing your conclusions.  I'll push one more time before the Christmas holiday tomorrow, and then there won't be any progress for more than a week. :-(  Hopefully answering my questions is less work than just writing it.

scrawl:
> Maybe I misunderstand. For export in OSG 3.2, I propose roughly:
> 
> writer->writeObject(geode->getDrawable(0));
> 
> For import in OSG 3.2:
> 
> drawable = reader->readNode(file);
> geode = new Geode();
> geode->addChild(drawable);
> return geode;

The problem isn't so much in this logic, but where would you inject it? You could put it in the RigGeometry serializer since we already have a custom serializer for that, but what about plain osg::Geometry?

> Is the problem that with out an enclosing Geode that that the 3.2 osgconv would consider the serialization invalid?

Correct.

> I was under the impression that the NIFs had separate reduced polygon count geometry for the collision boxes

Correct.

cfcohen:
There been no action on this pull request for a long time because when I tested the code after the Christmas holidays, I was very disappointed with the performance of the OSG serializer.  You're welcome to do whatever you like with this code, but it doesn't seem fruitful to me to continue down this path.  As I suggested in the initial message of this pull request, this was intended to be a learning experience for me, and it was very successful.  I'm grateful for your feedback and assistance.  I've learned a lot about OpenMW in the process and will try again to contribute in the future.   I'm going to close the pull request, but if you would prefer some other action please let me know.

scrawl:
I presume by performance you don't mean speed, but that it was a pain to work with design/functionality wise?

The only alternative would be starting from scratch, so I think that the osg serializer is still our best bet. I plan to pick up this code eventually when our OSG extensions have matured a bit more.

Maybe we should keep the PR open, just so it's visible as a starting point if/when someone wants to continue down this path.

cfcohen:
Actually, I did mean speed.  It's approximately an order of magnitude slower than the NIF loader. :-(  One of my original goals was to see if I could load more cells with acceptable performance without having to multi-thread the resource loader (which currently seems to big for me and is on your list anyway).  I got distracted with learning and just getting it to work. When I returned to the performance issue it turned out to be a total fail. :-(

Although I did find the OSG serializer awkward to work with, overly complicated and reliant on a couple of design decisions that I think are in very poor form (the library-based static object constructors is at the top of the list).  While it's a great general purpose serializer, I'm now convinced that it's not really appropriate for use in a performance sensitive context.   I've been thinking about an OSG protobuf based serializer, but there's nothing to show.

As an aside, I've found more than 2000 good animations at mixamo.com in BVH format that are free (royalty free use license), and I have been working on curve fitting algorithms to convert them to keyframe/channel format.  Working on the OSG resource loader helped me realize the extent to which all of the complicated bits are in the animation code, so that was very helpful.

Since you suggested re-opening the pull request, I've done so.   I'd assumed that you would want to "clean it up" if it wasn't receiving active development.  I may return to this request at some point as well just to learn more about the physics code, bounding boxes etc.

scrawl:
I remember doing a quick test loading a static model, the osg version was slightly slower but almost negligibly so. There might be a more of a difference when it comes to loading particles, animations, controllers, etc. though.

Just to be sure, you profiled with a release build, right?

cfcohen:
It's been a while now since I conducted the test, but I'm fairly sure that it was a release build (both OSG and OpenMW).   I did not investigate which parts of the serializer were the problem, but I probably should have.   My guess would be that it's the animations that cause the problems.  If it turns out to just be one part of the serializer that performs poorly, this code might be salvageable.

scrawl:
I'm coming back to the serialization topic because I'm adding a scene graph inspector tool for debugging purposes (showscenegraph in the vanilla engine).

For reference, here's how one could create a serializer without using the osg #defines or the static object hack:

```
template <class Cls>
static osg::Object* createInstanceFunc() { return new Cls; }

void registerSerializer()
{
    osg::ref_ptr<osgDB::ObjectWrapper> wrapper = new osgDB::ObjectWrapper(createInstanceFunc<SceneUtil::PositionAttitudeTransform>, "SceneUtil::PositionAttitudeTransform", "osg::Object osg::Node osg::Group osg::Transform SceneUtil::PositionAttitudeTransform" );
    wrapper->addSerializer( new osgDB::PropByRefSerializer< SceneUtil::PositionAttitudeTransform, osg::Vec3f >(
        "position", osg::Vec3f(), &SceneUtil::PositionAttitudeTransform::getPosition, &SceneUtil::PositionAttitudeTransform::setPosition), osgDB::BaseSerializer::RW_VEC3F );

    osgDB::Registry::instance()->getObjectWrapperManager()->addWrapper(wrapper);
}
```
While there's some redundancy in these statements, at least its easier to see what the code is actually doing.

To make things a little cleaner, we could use a sub-class:
```
template <class Cls>
static osg::Object* createInstanceFunc() { return new Cls; }

class PositionAttitudeTransformSerializer : public osgDB::ObjectWrapper
{
public:
    PositionAttitudeTransformSerializer()
    {
        addSerializer( new osgDB::PropByRefSerializer< SceneUtil::PositionAttitudeTransform, osg::Vec3f >(
            "position", osg::Vec3f(), &SceneUtil::PositionAttitudeTransform::getPosition, &SceneUtil::PositionAttitudeTransform::setPosition), osgDB::BaseSerializer::RW_VEC3F );
    }
};

void registerSerializer()
{
    osgDB::Registry::instance()->getObjectWrapperManager()->addWrapper(new PositionAttitudeTransformSerializer);
}
```

psi29a:
@cfcohen you still around? I would like to discuss this PR with you. :)

akortunov:
Do we really need a such kind of serialization? IMO, "showscenegraph" is enough for debugging (but probably we need a some kind of *.osgt-viewer).

akortunov:
Closing this PR since it is incompatible with our current codebase, author is inactive for three years and there is no real discussion since 2017.

PR #847, opened 2015-12-06
disable occlusion query for gles1

sandstranger:
gles1 not support occlusion query , only gles 3.0 support it . 
If build game with occlusion query for gles1, game will crash on external locations


scrawl:
The crash really needs to be fixed in OSG though. Did you test my patch at https://github.com/scrawl/osg/tree/occlusionquery_fix ?

sandstranger:
Just now tested it . This patch for osg works

scrawl:
Ok, I will merge to scrawl/osg master branch and propose it to upstream OSG.

PR #849, opened 2015-12-07
Field of view

scrawl:
* Handle Werewolf FOV setting (note: need to re-run mwiniimporter for this to work, since the setting wasn't previously imported)
* Control the FOV for first person meshes separately from the gameworld FOV (https://bugs.openmw.org/issues/3051, https://bugs.openmw.org/issues/858)


cfcohen:
I updated the wiki for these options to the best of my ability.   Please free free to correct any mistakes I may have made.

PR #850, opened 2015-12-08
Fix for bug #2952: Enchantment Merchant Items reshuffled EVERY time 'barter' is clicked

MatthewRock:
Sorry that it took so long, but I'm busy lately.

[Bugtracker link](http://bugs.openmw.org/issues/2952)

Basically I did what you guys asked - now we remember the parent group. We save it in the save file, and later recover it. If there's no group, we add empty group, and later remove the item(so the bug happens only once for older saves).

As for algorithm, it's pretty easy, I guess. We keep list of items that were **not** sold, and later we restock with basicRestockValue - notSoldItems. We keep this negative, because levelled group items have negative count. I tested saving/loading as well as restocking(on Galbedir) and it seems to work for me.

### DIFF
--- a/apps/openmw/mwworld/containerstore.hpp
+++ b/apps/openmw/mwworld/containerstore.hpp
@@ -3,6 +3,7 @@
 
 #include <iterator>
 #include <map>
+#include <utility>
 
 #include <components/esm/loadalch.hpp>
 #include <components/esm/loadappa.hpp>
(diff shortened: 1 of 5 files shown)


apps/openmw/mwworld/containerstore.hpp
```diff
@@ -68,9 +69,9 @@ namespace MWWorld
             MWWorld::CellRefList<ESM::Repair>            repairs;
             MWWorld::CellRefList<ESM::Weapon>            weapons;
 
-            std::map<std::string, int> mLevelledItemMap;
-            ///< Stores result of levelled item spawns. <refId, count>
-            /// This is used to remove the spawned item(s) if the levelled item is restocked.
+            std::map<std::string, std::pair<int, std::string> > mLevelledItemMap;
```
scrawl: This data structure does not correctly handle the case where the same item ID is selected by multiple levelled lists.

E.g.

levi_1 can contain item_a, item_b. RNG selects item_a
levi_2 can contain item_a, item_c. RNG selects item_a

With this map we can only store one link i.e. to levi_1 or levi_2, but we need to store both.

MatthewRock:
I guess it's fixed now.

PR #853, opened 2015-12-10
fix building with Opengl ES1 on linux

sandstranger:
it is building now  and working with shared osg . sdl2 must be build only with  enabled gles1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,6 +165,7 @@ option(OPENGL_ES "enable opengl es support" FALSE )
 
 if (OPENGL_ES)
     add_definitions(-DOPENGL_ES)
+    SET(OPENGLES_egl_LIBRARY CACHE STRING "Set the OpenGL ES egl library.")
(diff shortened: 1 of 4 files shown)


scrawl:
We shouldn't need to link against EGL for a non-static build.

sandstranger:
Ok , Than it will be link only to Android , because all libraries for this platform building statically

pcercuei:
It fails at runtime here, if OSG is compiled for GLES1 and with OSG_WINDOWING_SYSTEM=None. I get "Failed to create pbuffer, failing back to normal graphics window.". It comes from components/sdlutil/sdlcursormanager.cpp, apparently the cursor manager doesn't use the SDL2 context to render?

scrawl:
The cursor manager uses OSG windowing system. SDL doesn't support pbuffers as far as I know.

pcercuei:
I sent a separate PR to work around the issue.

sandstranger:
I suppose  this issue is resolved now

PR #854, opened 2015-12-11
Play NPC voices from the head position

kcat:
Fixes http://bugs.openmw.org/issues/2198

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -14,6 +14,7 @@
 namespace osg
 {
     class Vec3f;
+    class Matrixf;
     class Quat;
     class Image;
 }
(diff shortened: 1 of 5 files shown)


scrawl:
Looks good, but why the change to return the full matrix when we only need position? I'm an advocate of coding things as they're needed, no need to complicate the code with features that aren't in use yet (and maybe never will be).

I was actually considering to write an optimized getWorldPos() that only accumulates position. Should be faster than accumulating the matrices (getWorldMatrices() appears to take ~30us each :/).

kcat:
> Looks good, but why the change to return the full matrix when we only need position?

I intended to play around with the idea of adding directionality to the sound to make it a touch duller when the NPC is looking away while talking. Since it would be an enhancement it should probably wait for now, but to even think about doing that, I would also use the rotation.

I suppose theoretically I could just use the actor's rotation, but that wouldn't account for rotation caused by animations or head tracking.

> I was actually considering to write an optimized getWorldPos() that only accumulates position. Should be faster than accumulating the matrices (getWorldMatrices() on the head node appears to take ~30us each :/).

Might also be faster to make a function that returns just the accumulated head matrix, instead of allocating and filling a vector of all transforms up to the given node.

PR #857, opened 2015-12-12
disable end enable moons on gles1

sandstranger:
I found only this way to enable and disable moons on gles1 . It works very bad of course , because moons instantly disappears from sky


scrawl:
We need to find out why the moons are not rendering correctly, not disable them completely.

sandstranger:
Ok

PR #858, opened 2015-12-13
Separate and expand texture filtering options

kcat:
This separates the options for texture filtering into the base filter (nearest, linear) and the mipmap filter (none (i.e. disabled), nearest, linear). As it was, the bilinear/trilinear option values were rather obtuse (bilinear still contained mipmapping, with no way to disable it), and there was no longer a way to set nearest filtering like there had been in the past.

I'm not too sure about the new layout of the video detail page. In particular, the Anisotropy option looks a bit weird. But it's all functional, from what I can tell.

Done based on this request: https://forum.openmw.org/viewtopic.php?f=3&t=3239

### DIFF
--- a/apps/openmw/mwgui/settingswindow.hpp
+++ b/apps/openmw/mwgui/settingswindow.hpp
@@ -34,7 +34,6 @@ namespace MWGui
             MyGUI::ScrollBar* mDifficultySlider;
             MyGUI::ScrollBar* mAnisotropySlider;
             MyGUI::ComboBox* mTextureFilteringButton;
-            MyGUI::ComboBox* mTextureMipmappingButton;
             MyGUI::TextBox* mAnisotropyLabel;
             MyGUI::Widget* mAnisotropyBox;
             MyGUI::Button* mShadersButton;
--- a/apps/openmw/mwgui/settingswindow.hpp
+++ b/apps/openmw/mwgui/settingswindow.hpp
@@ -53,7 +53,6 @@ namespace MWGui
             void onTabChanged(MyGUI::TabControl* _sender, size_t index);
             void onOkButtonClicked(MyGUI::Widget* _sender);
             void onTextureFilteringChanged(MyGUI::ComboBox* _sender, size_t pos);
-            void onTextureMipmappingChanged(MyGUI::ComboBox* _sender, size_t pos);
             void onSliderChangePosition(MyGUI::ScrollBar* scroller, size_t pos);
             void onButtonToggled(MyGUI::Widget* _sender);
             void onResolutionSelected(MyGUI::ListBox* _sender, size_t index);
--- a/apps/openmw/mwgui/settingswindow.hpp
+++ b/apps/openmw/mwgui/settingswindow.hpp
@@ -34,6 +34,7 @@ namespace MWGui
             MyGUI::ScrollBar* mDifficultySlider;
             MyGUI::ScrollBar* mAnisotropySlider;
             MyGUI::ComboBox* mTextureFilteringButton;
+            MyGUI::ComboBox* mTextureMipmappingButton;
             MyGUI::TextBox* mAnisotropyLabel;
             MyGUI::Widget* mAnisotropyBox;
             MyGUI::Button* mShadersButton;
@@ -53,6 +54,7 @@ namespace MWGui
             void onTabChanged(MyGUI::TabControl* _sender, size_t index);
             void onOkButtonClicked(MyGUI::Widget* _sender);
             void onTextureFilteringChanged(MyGUI::ComboBox* _sender, size_t pos);
+            void onTextureMipmappingChanged(MyGUI::ComboBox* _sender, size_t pos);
             void onSliderChangePosition(MyGUI::ScrollBar* scroller, size_t pos);
             void onButtonToggled(MyGUI::Widget* _sender);
             void onResolutionSelected(MyGUI::ListBox* _sender, size_t index);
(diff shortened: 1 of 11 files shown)


scrawl:
I'm totally fine with having these options in the settings file, but maybe we should keep the menu as it is? Most players won't know what mipmapping is, and some players won't know whether 'Nearest' or 'Linear' is the higher quality option. I think Bilinear and Trilinear filtering are more commonly used terms, and it's easier to tell what the higher quality option is (3 > 2). mipmapping = none looks terribly grainy so it shouldn't be in the menu to begin with. We want to limit to sensible choices to avoid confusion (e.g. in the past we've had players going nuts on the FoV slider without really knowing what it does, then wondering why their view is all distorted - in response we changed the range of the FoV slider to be more sensible).

There appears to be a bug too: If I set mipmapping to 'None' via ingame UI, I can't change it back anymore without restarting the game.

kcat:
I'm okay with keeping the in-game options simplified, though I'm not quite sure how to do it. I could tie the "Texture filtering" option to be the mipmapping option, with Bilinear->"nearest" and Trilinear->"linear" (that would basically be the original behavior). But what should I do when "none" is set in the config? Wouldn't that also risk getting overwritten even if the user didn't change it? Or if the user wanted to change it back and forth?

> If I set mipmapping to 'None' via ingame UI, I can't change it back anymore without restarting the game.

Hmm, I think what's happening is setting non-mipmapped filtering causes OSG to drop or not load mipmaps, and when it's reenabled, the mipmaps don't get reloaded because the original image in system mem was unref'd.

scrawl:
> Hmm, I think what's happening is setting non-mipmapped filtering causes OSG to drop or not load mipmaps, and when it's reenabled, the mipmaps don't get reloaded because the original image in system mem was unref'd.

I think it's a case sensitivity issue. The settings UI sets the setting to 'None', 'Linear' or 'Nearest', while the updateTextureFiltering() backend expects 'linear', 'nearest', or 'none'. That brings up a point we might want warning messages for unrecognized values.

> But what should I do when "none" is set in the config? Wouldn't that also risk getting overwritten even if the user didn't change it?

This would be a general issue with having settings in the config file that are outside the range of in-game sliders/combo boxes. AFAIK we are handling this correctly (the setting isn't set unless the user touches the combobox).

kcat:
> I think it's a case sensitivity issue. The settings UI sets the setting to 'None', 'Linear' or 'Nearest', while the updateTextureFiltering() backend expects 'linear', 'nearest', or 'none'.

The ```onTexture*Changed``` functions make the value lower-case before changing it in the settings manager. I'll add some warnings when an unrecognized value is encountered though, that's a good idea.

kcat:
Updated to make the in-game options like they used to be. I'm not seeing an issue with string cases, making me think reenabling mipmaps not immediately enabling mipmaps is how mipmaps are handled by OSG (and requires actually reloading the textures).

scrawl:
Now this PR sort of breaks compatibility with old settings files (texture filtering = "trilinear" or "bilinear" used to be a valid option, but now it's unrecognized and throws log warnings).

Possible fixes:
- when encountering "trilinear" / "bilinear" settings, convert to new settings format automatically (i.e. set filtering and mipmapping settings appropriately).
- or: rename "texture filtering" to something else. The downside would be previous users settings are lost.

kcat:
I think renaming ```texture filtering``` would probably be the better option. That way it doesn't alter the user's setting automatically, it's less prone to confusion (what if ```texture filtering = trlinear``` and ```texture mipmapping = nearest``` are set?), and it still allows users to go back and forth between old and new versions.

scrawl:
Looks like there is another bug (already present before your PR). updateTextureFiltering() isn't called at the start of the game, so what you set in the settings file has no effect until you touch the ingame filtering combobox.

How about configuring ```min``` and ```mag``` separately? For me, min=nearest doesn't seem very pleasant to look at, just makes objects in the distance look grainy. mag=nearest on the other hand could be more appealing (just pixelated, not grainy).

If we're going to change up the format, we might as well mimic the OpenGL syntax (i.e. min_filter and mag_filter, with 6 possible values each, minus the mipmap values not being valid for mag_filter).

scrawl:
> updateTextureFiltering() isn't called at the start of the game, so what you set in the settings file has no effect until you touch the ingame filtering combobox.

Hm, actually we have another copy of the filtering code in Engine::prepareEngine. I guess these should be combined in some way.

kcat:
> If we're going to change up the format, we might as well mimic the OpenGL syntax (i.e. min_filter and mag_filter, with 6 possible values each, minus the mipmap values not being valid for mag_filter).

I'm fine with having separate min and mag filter settings, but I was never a fan of the min filter controlling both the minification and mipmapping filters. Combining the min filter with the mipmap means you get a total of 3x2 values, rather than 3+2 with them separate. Adding a new value (which we could do with shaders, adding new filtering methods) would increase it by a multiple, instead of just by 1. It also makes it more annoying to manipulate the minification filter separately from the mipmaping method, since you have to consider the other to know what to set.

scrawl:
Fine, then let's have ```texture min filter```, ```texture mag filter``` and ```texture mipmapping```. That takes care of the settings file incompatibility as well.

Regarding duplication of the settings parsing code, we could add a new method TextureManager::setFilterSettings (const std::string&, const std::string&, const std::string&, int anisotropy) that would be used by Engine::prepareEngine and RenderingManager::updateTextureFiltering.

kcat:
Separated the min and mag filter options, and combined the duplicated code.

apps/openmw/mwgui/settingswindow.hpp
```diff
@@ -53,6 +53,7 @@ namespace MWGui
             void onTabChanged(MyGUI::TabControl* _sender, size_t index);
             void onOkButtonClicked(MyGUI::Widget* _sender);
             void onTextureFilteringChanged(MyGUI::ComboBox* _sender, size_t pos);
+            void onTextureMipmappingChanged(MyGUI::ComboBox* _sender, size_t pos);
```
scrawl: Leftover declaration

kcat:
Fixed

PR #861, opened 2015-12-15
Fix crash when running out of sound sources

scrawl:
@kcat 

We were inserting a new entry into mActiveSaySounds even if the playVoice call failed. Other code was assuming the mActiveSaySounds only contains valid Sound pointers, resulting in a crash.

The crash can be reproduced by commenting   ```// if(mFreeSources.empty())``` in OpenAL_Output::streamSound3D, then talk to an NPC.


kcat:
I suppose this makes sense. Was hoping to avoid the double lookup, but oh well. NULL checks wouldn't be any better.

PR #862, opened 2015-12-17
ConstPtr

scrawl:
Most of the existing code is not utilizing the ConstPtr yet. As always, before I do mass changes I would like to get some feedback on the overall approach.

### DIFF
--- a/apps/openmw/mwworld/cellstore.hpp
+++ b/apps/openmw/mwworld/cellstore.hpp
@@ -217,7 +217,7 @@ namespace MWWorld
             ConstPtr searchConst (const std::string& id) const;
             ///< Will return an empty Ptr if cell is not loaded. Does not check references in
             /// containers.
-            /// @note Does not trigger CellStore hasState flag. Do not modify the returned Ptr!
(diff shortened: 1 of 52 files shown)


zinnschlag:
Constness, yay!

That is exactly what I had in mind.

apps/openmw/mwworld/cellstore.hpp
```diff
@@ -214,7 +214,7 @@ namespace MWWorld
             /// containers.
             /// @note Triggers CellStore hasState flag.
 
-            Ptr searchConst (const std::string& id) const;
+            ConstPtr searchConst (const std::string& id) const;
             ///< Will return an empty Ptr if cell is not loaded. Does not check references in
             /// containers.
             /// @note Does not trigger CellStore hasState flag. Do not modify the returned Ptr!
```
zinnschlag: Do we still need that note? It's const so it should be obvious (and compiler enforced) that you should not change the reference.

psi29a:
Yay for const!  Time to pollute the code-base! ;)

PR #863, opened 2015-12-20
Bug Fix 3020: Added checks to verifier for container inventories

Aesylwinn:
Added a bit to the containerCheck method so that it also checks the contained items and generates the appropriate error message. I'm not sure what items are allowed in an inventory, so that needs verification. Also, this is my first pull request, so let me know if I did something wrong with git.

### DIFF
--- a/apps/opencs/model/tools/referenceablecheck.cpp
+++ b/apps/opencs/model/tools/referenceablecheck.cpp
@@ -399,39 +399,7 @@ void CSMTools::ReferenceableCheckStage::containerCheck(
         messages.push_back (std::make_pair (id, container.mId + " has an empty name"));
     
     //checking contained items
-    const std::vector<ESM::ContItem>& itemList = container.mInventory.mList;
-    for (unsigned i = 0; i < itemList.size(); ++i)
-    {
-        std::string itemName = itemList[i].mItem.toString();
-        CSMWorld::RefIdData::LocalIndex localIndex = mReferencables.searchId(itemName);
-
-        if (localIndex.first == -1)
-            messages.push_back (std::make_pair (id,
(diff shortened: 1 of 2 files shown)


scrawl:
This checks the items in Containers. We need the same check for items in Creatures and Npcs.

The wording of the error messages seems a bit unclear. How about:

> "contains unreferenced item " + itemName

"contains non-existing item (" + itemName + ")".

> contains illegal item " + itemName));

"contains item of invalid type (" + itemName + ")".

apps/opencs/model/tools/referenceablecheck.cpp
```diff
@@ -397,6 +397,41 @@ void CSMTools::ReferenceableCheckStage::containerCheck(
     //checking for name
     if (container.mName.empty())
         messages.push_back (std::make_pair (id, container.mId + " has an empty name"));
+    
+    //checking contained items
+    const std::vector<ESM::ContItem>& itemList = container.mInventory.mList;
+    for (unsigned i = 0; i < itemList.size(); ++i)
```
scrawl: Had to look up if 'unsigned i' is actually legal C++. Apparently it is. Can we change to 'unsigned int i' anyway just for clarity's sake?

Aesylwinn: I've looked at the npc and creature data classes, and it looks like they implement the inventory the same way, so I'll make a separate function for comparing inventories. 
I'll go ahead and change the error messages. 
I also didn't realize 'unsigned' was legal until looking it up after seeing it later in the same file. I have no problem with changing it.

kcat: ```size_t``` might actually be better, since that's essentially what the ```size()``` method returns.

PR #864, opened 2015-12-21
Allow alternate audio/video format resources

kcat:
A bit of a potentially-controversial one, this is. This effectively ignores file extensions when looking to open audio and video files (it does /not/ alter texture or model loading behavior). So if the game tries to play `music/foo.mp3` or `video/mw_intro.bik`, it may instead play `music/foo.ogg` or `video/mw_intro.webm` if such files exist in a higher-priority archive.

This also means we can get rid of the wav->mp3 special case, since trying to play `sound/vo/a/f/atk_af001.wav` (as indicated by the game data) will automatically pick up the existing `sound/vo/a/f/atk_af001.mp3` resource instead.


psi29a:
I like this, how does it handle if there are two files with the same extension... a mod for example?

Ragora:
Seems like it's just whichever one has the higher priority VFS mount, so I would figure that mods get a higher priority mount over base content. Though it would be an interesting thing to handle when two or more mods try to declare their own version of the same file, though at that point it might suffice to say that the mods can't be used together as it would be a fairly complicated thing to work out.

kcat:
What Ragora said. The order that the ```fallback-archive``` and ```data``` lines appear in openmw.cfg defines the priority for data paths/archives (similar to ```content``` lines for esm/esp/omwaddon priority). This is preexisting behavior.

What this PR does is, essentially, change the way it looks for and opens audio (music/ and sound/) and video files. Instead of trying to match the full name, it will try to match the base name (resource path + name without extension), still obeying archive priority, and use what it finds that way. So for example, if the game tries to play ```music/explore/mx_explore_1.mp3```, it'll look for a resource with a base name of ```music/explore/mx_explore_1```. If ModA (listed first) contains ```music/explore/mx_explore_1.mp3``` and ModB (listed last) contains ```music/explore/mx_explore_1.ogg```, then the ogg file gets used since it has a matching base name and is in the higher priority archive.

terrorfisch:
Are there cases where you don't whant this behaviour of the vfs for audio/video files? If not I would suggest to simplify the `Manager` class interface to having only the get function so the rest of the code can be unaware of the extension ignoring bahaviour.

kcat:
> Are there cases where you don't whant this behaviour of the vfs for audio/video files?

I can't think of a case in openmw itself to not want this for audio/video files. You don't want it for certain other file types, though (e.g. meshes, which have some instances of meshes/xfoo.nif and meshes/xfoo.kf, which you need to differentiate).

kcat:
Oops, accidentally hit Comment too soon. But in addition, openmw-cs might need reconsideration or fixing too... it still seems to use the same openmw.cfg as the game, so it pulls in *all* your active mod archives/directories you use while playing, instead of remaining "clean" for mod editing. So external resource validation and resource viewing will use whatever your mods have, rather than what you're actually editing with.

kcat:
Updated to clean away a couple now-unnecessary methods, moving the file selection logic into the sound manager class, and making it so the decoder opens exactly the file it's given (in case something in the future wants to open a specific audio file without this behavior).

kcat:
Any comments? This would allow to start easily making sound/video replacer mods using open formats.

zinnschlag:
That's an old pull request. Didn't we have a discussion about this topic before somewhere else? I can see the point of this PR. I would prefer an approach that doesn't make our file handling even more fuzzy, but anything I can think of would require a lot more work and would probably be a post-1.0 improvement.

@scrawl Your opinion?

About the findFirstOf function: This basically just discards the extension when looking for a file, which works in our specific situation, because files are sorted into different subdirectories by type/purpose. Still, it looks a bit risky. What if another file makes its way into there? A txt file with some documentation would be an obvious case. But there might be others. MW modding has never been one of the cleanest activities. And if the name of such a file is matching by accident, we have a problem.

scrawl:
Previous discussion on https://github.com/OpenMW/openmw/pull/842#issuecomment-162241674.

I feel conflicted about this PR. On the one hand it is useful, on the other hand it introduces new potential issues. Perhaps we could call it an "experimental" feature that the user has to enable in the settings.cfg first? 

> Still, it looks a bit risky. What if another file makes its way into there? A txt file with some documentation would be an obvious case. But there might be others. MW modding has never been one of the cleanest activities. And if the name of such a file is matching by accident, we have a problem.

Good point.

zinnschlag:
> Perhaps we could call it an "experimental" feature that the user has to enable in the settings.cfg first?

Agreed. But I would put it into openmw.cfg since it is also relevant for OpenMW-CS.

kcat:
> About the findFirstOf function: This basically just discards the extension when looking for a file, which works in our specific situation, because files are sorted into different subdirectories by type/purpose. Still, it looks a bit risky. What if another file makes its way into there? A txt file with some documentation would be an obvious case.

I'm not aware of any cases where this happens. It's customary for documentation files to go into a ```docs/``` sub-directory or the root directory, and I don't think it would be a bad idea to start actively encouraging putting files in their appropriate place (the ```sound/``` or ```video/``` directories being no place for non-sound or non-video files).

If it turns out to really be a problem, though, we could just alter the findFirstOf function later to accept only a given list of known extensions. This would be less than ideal though, since as mentioned in the other PR, we don't necessarily know which formats ffmpeg and such support, and relying on the file extension to determine the exact format isn't foolproof, so I'd rather only do it if absolutely necessary (i.e. if there's too many real-world cases where this breaks and can't be fixed by modders properly repackaging their stuff).

> Perhaps we could call it an "experimental" feature that the user has to enable in the settings.cfg first?

This also works around bugs in the vanilla game data, so disabling it would break many instances of where the game specifies wav but provides mp3 (short of reintroducing the explicit wav->mp3 workaround, which this is supposed to generalize). But even aside from that, I think it would be better to go all in and just have the feature so we can more quickly find out if it's really a problem, rather than wait and hope the right person with the right mod(s) happens to enable it and report a problem. If we make it an experimental feature that users have to enable, how can we tell if there's been enough testing to know if it's safe? And with this disabled, we'd have a harder time catching modders early if they do this.

zinnschlag:
> I'm not aware of any cases where this happens. It's customary for documentation files to go into a docs/ sub-directory or the root directory, and I don't think it would be a bad idea to start actively encouraging putting files in their appropriate place (the sound/ or video/ directories being no place for non-sound or non-video files).

It happens. Can't quote a situation where it happened for documentation files. But with other files it does happen. I just looked through my data directory. Seems there are a couple of .max files in the textures directory. Clearly an oversight of a modder, but as long as people have to manually fiddle around with files it will happen and that should not stop OpenMW from working correctly.
I agree about enforcing the file structure. But that is a post-1.0 feature and we will still have to deal with a lot of old mods that were made before enforced file structures.

kcat:
> Seems there are a couple of .max files in the textures directory.

Which mod is that? Perhaps we could try to contact the modder to let them know this could be a problem with OpenMW in the future.

Perhaps rather than only allowing specific extensions, we instead blacklist extensions that we discover are a problem. Since .max appears to be a 3DS Max scene/model file, there'd be no harm in ignoring it when looking up textures. Hopefully too blacklists will be shorter than whitelists.

zinnschlag:
> Which mod is that? Perhaps we could try to contact the modder to let them know this could be a problem with OpenMW in the future.

I don't know. This is the data directory I carry around with me since 2006 or so. No way to track what came from where.

zinnschlag:
We should try to come to a conclusion for this one. But I don't have any further input on it at this point.

@scrawl Anything else to add?

zinnschlag:
Actually, I do have something to add. I don't think using a config flag to control this feature will work. As soon as we are putting out an OpenMW build with the feature, these whacky modders will start to make use of it, which means they will produce content files that depend on this feature and if we ever going to remove or substantially alter this feature we would break these mods. That would be bad.

If we ever find a better way to handle this problem we still can retire it for newer version of the content file format without breaking old content.

scrawl:
> Perhaps rather than only allowing specific extensions, we instead blacklist extensions that we discover are a problem. Since .max appears to be a 3DS Max scene/model file, there'd be no harm in ignoring it when looking up textures. Hopefully too blacklists will be shorter than whitelists.

I think I'd rather have a whitelist than a blacklist. A blacklist seems like something you'd have to 
constantly update as more broken mods are found, whereas a whitelist of valid formats once complete should cover everything. An incomplete blacklist would have the potential to break previously working mods, whereas an incomplete whitelist would mean the extension wrangling feature can't be used yet for certain formats (which is not a dealbreaker, and there are workarounds for that).

Assuming that we expand this feature to meshes, it could be a problem that we have .nif and .kf files of the same basename in the meshes directory.

zinnschlag:
Agreed. A whitelist may have its downside (I guess there is no way to query media libraries for a list of supported extensions?), but a blacklist just won't do the job.

psi29a:
Not sure if it is useful...

ffmpeg -codecs
ffmpeg -formats

That would gives us a list of supported files on that particular instance of ffmpeg.

scrawl:
Well yes, there might be a way to query supported formats at runtime, but do we really want that? Let's say someone creates a mod using format A and then a user of that mod has not built FFmpeg with support for that format. If we don't consider unsupported files, then the mod will simply not work with no error message whatsoever. If we attempt to open the unsupported format, we should get a nice error message that will point the user in the right direction to fix their FFmpeg installation.

zinnschlag:
Makes sense. A hardcoded whitelist gives better error handling capabilities.

kcat:
If a hardcoded whitelist is what you want to use, then so be it. But I'd like to make one final case for a blacklist.

A blacklist means we don't have to specifically acknowledge a format for it to be usable. It also means we don't have to specifically acknowledge esoteric extensions to more common formats (oga, m4a, etc). Users and modders can just use what they want, without us having to guess what formats someone somewhere might want to use sometime. These are the two main things that bug me most about a whitelist, because all too often games (particularly commercial ones) restrict themselves to using specific formats, like mp3 or xwm, even though they're using backends or middleware that I know can just as easily handle better formats, like ogg or flac, if they'd just accept it. It's like you need permission from the developers to use specific formats.

Further, a whitelist will inevitably grow to be really big as more usable formats are discovered or made, while a blacklist will remain nearly empty as long as modders are cooperative. And even in the case where a blacklist fails due to a previously unknown problem mod, the user has the ability to fix it by deleting the errant file while we either try to get the modder to fix their packaging or add it to the blacklist... while conversely, a user or modder that wants to use a format we didn't previously recognize will have to wait until we acknowledge it and then require everyone using their mod to update the engine.

jeffreyhaines:
@kcat I think a whitelist, from a testing point of view

1) acknowledges our intent to support the format.
2) acknowledges that we've tested the format and it does work.

Both of which save us time and respect by preventing users from claiming foul if we don't support their favorite format. Given just a blacklist, I could create my own format that is completely invalid. I may also be harmful.

Also,

@kcat do you have an active bug ticket for this issue? One needs to be created before we proceed to merge and/or close this pull request. Please post a link to that new ticket here, thanks

kcat:
@jeffreyhaines but from a usability point of view, it effectively locks modders into using specific "approved" formats, as nothing but the explicitly listed extensions will work. Few modders are going to request format support aside from the obvious big ones (e.g. mp4, ogg, webm), because to do so would require them to wait for the next release for their mod to work, and additionally require users to use the bleeding-edge release to use the mod. It's usually a big new feature that gets people to do things like that (e.g. MGE XE's PBR renderer), not simply recognizing an extension that ffmpeg has always supported but simply had blocked.

There's the other issue too that file extensions aren't directly correlated to specific formats. An .ogg file, for example, can represent a multitude of formats, from vorbis, to flac, to opus, to theora, and others. Similar for mp4, which can employ a number of mpeg video/audio formats and AAC. Or wav, which can have raw pcm, various encoding formats, and even mp3. More often than not, an extension check mostly just tells you about the container, not the audio/video format itself.

Which leads on to the next point, that ffmpeg doesn't actually use the file extension to identify the format. A whitelist is easy to work around in allowing "non-approved" formats through by renaming the file to have a whitelisted extension, then ffmpeg will open it if it recognizes the file contents. As such, the intent of the whitelist becomes muddy... did we not whitelist an esoteric extension because we didn't know of it, or for an actual reason? A file not working with a blacklist is obvious, but a file not working with a whitelist is a question. Which then goes back around to pushing people toward using an "approved formats list", which is only a small subset of what ffmpeg can actually handle, rather than letting people use what they want that ffmpeg can handle.

kcat:
Regardless, though, this PR is so out of date I'd have to rewrite it anyway. I may look into it again at some point, but not right now.

PR #869, opened 2016-01-02
Use correct direction multipliers during awareness check

neuralroberts:
A sneaking player is more likely to be noticed when behind an NPC, than when in front. This is the opposite of vanilla behavior. It happens in both 0.37 and in the master branch.


neuralroberts:
Submitted as [bug #3122](https://bugs.openmw.org/issues/3122)

scrawl:
Thanks!

PR #871, opened 2016-01-07
Improve 'Drop' command behavior (Fixes #1544)

jordan-ayers:
Item drops will now attempt to re-stack equipped items with other items before dropping, and equipped items are only dropped if there aren't enough unequipped items to satisfy the command.

I tried using a ref for the inventoryStore, but that seemed to require either multiple fetches of the store or an uninitialized ref with a guard flag, which seemed worse than the pointer-based solution.

### DIFF
--- a/apps/openmw/mwscript/miscextensions.cpp
+++ b/apps/openmw/mwscript/miscextensions.cpp
@@ -21,6 +21,7 @@
 #include "../mwworld/class.hpp"
 #include "../mwworld/player.hpp"
 #include "../mwworld/containerstore.hpp"
+#include "../mwworld/inventorystore.hpp"
 #include "../mwworld/esmstore.hpp"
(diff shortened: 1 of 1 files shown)


jordan-ayers:
After thinking about this one some more, I'm pretty sure it's is incomplete / wrong.  I had been testing with armor, and assumed that the quantity for an equipped item in a slot would always be 1; but that's not required for projectiles.

I'll try to get that fixed in the next couple of days.

psi29a:
To prevent an accidental merge, can cancel/retract your merge request and re-open it when you're ready to merge again. :)

zinnschlag:
Or use the appropriate label.

jordan-ayers:
I saw the labels, but hadn't learned how to use them yet.  I'll bring this back after the projectiles are fixed.

PR #872, opened 2016-01-11
Lights in OpenMW-CS

scrawl:
Fixes https://bugs.openmw.org/issues/2484


zinnschlag:
Wow, that was some chaotic commenting. Sorry for that.

zinnschlag:
Should we go ahead with the merge and you can adjust the enums and add the documentation later? Lot's of stuff in this PR and if we keep this sitting around much longer we are probably going to run into merge conflicts (I have some pieces of code in the making that may overlap with it).

scrawl:
Should be all done.

PR #873, opened 2016-01-16
Remove unused code

rcutmore:
This removes some commented out code I noticed while looking around the code base. Feel free to close if you prefer pull requests with more substantive changes.


scrawl:
As a further enhancement, the setup() function can be private.

rcutmore:
@scrawl Thanks, I will open a separate pull request for that suggestion.

PR #875, opened 2016-01-16
Properly find MyGUI in /usr/local/lib

hristoast:
Properly find MyGUI when it's been installed in /usr/local versus a system-provided location.

### DIFF
--- a/cmake/FindMyGUI.cmake
+++ b/cmake/FindMyGUI.cmake
@@ -101,7 +101,7 @@ ELSE (WIN32) #Unix
             SET(MYGUI_LIBRARIES ${MYGUI_LIBRARIES} CACHE STRING "")
         ELSE (MYGUI_INCLUDE_DIRS)
             FIND_PATH(MYGUI_INCLUDE_DIRS MyGUI.h PATHS /usr/local/include /usr/include PATH_SUFFIXES MyGUI MYGUI)
-            FIND_LIBRARY(MYGUI_LIBRARIES mygui PATHS /usr/lib /usr/local/lib)
+            FIND_LIBRARY(MYGUI_LIBRARIES MyGUIEngine PATHS /usr/local/lib /usr/lib)
             SET(MYGUI_LIB_DIR ${MYGUI_LIBRARIES})
             STRING(REGEX REPLACE "(.*)/.*" "\\1" MYGUI_LIB_DIR "${MYGUI_LIB_DIR}")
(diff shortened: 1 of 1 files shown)


scrawl:
Pretty sure I've installed MyGUI to /usr/local before and it worked without issue - can you tell us some details about your setup?

The fact that you're taking the ELSE (MYGUI_INCLUDE_DIRS) path suggests something is awry with the pkgconfig file.

kcat:
I have MyGUI in /usr/local and it works fine. Something odd about this piece of code though is that it's looking for the ```mygui``` lib, but there's no libmygui.so or anything. It should be ```MyGUIEngine``` I think (you don't want to include the lib and .so portion, since different OSs may dress libs differently, e.g. OSX uses .dylib, which cmake handles automatically).

hristoast:
> Pretty sure I've installed MyGUI to /usr/local before and it worked without issue - can you tell us some details about your setup?

Yeah sure - [my distro](https://voidlinux.eu/) doesn't provide MyGUI or Unshield packages (or OpenMW) so I've compiled these from source and they live under `/usr/local`

> The fact that you're taking the ELSE (MYGUI_INCLUDE_DIRS) path suggests something is awry with the pkgconfig file.

[Here's my MYGUI.pc file](http://hastebin.com/avuhifemij.avrasm) - though I'm not sure what it *should* look like.

Thanks!

hristoast:
> It should be MyGUIEngine I think (you don't want to include the lib and .so portion, since different OSs may dress libs differently, e.g. OSX uses .dylib, which cmake handles automatically).

Ah okay, thanks - I can amend my commit, if need be.

scrawl:
Check the list of search paths that you have in pkg-config. If it's not configured by default to look in /usr/local/lib/pkgconfig, you may need to add that path manually. Mine says:

<pre>
$ pkg-config --variable pc_path pkg-config
/usr/local/lib/x86_64-linux-gnu/pkgconfig:/usr/local/lib/pkgconfig:/usr/local/share/pkgconfig:/usr/lib/x86_64-linux-gnu/pkgconfig:/usr/lib/pkgconfig:/usr/share/pkgconfig
</pre>
> Ah okay, thanks - I can amend my commit, if need be.

Sure do that and we can merge.

hristoast:
```
➜ pkg-config --variable pc_path pkg-config
/usr/lib/pkgconfig:/usr/share/pkgconfig
```

Sure enough, I am missing `/usr/local/lib/pkgconfig`! So this is indeed a separate issue.

PR #876, opened 2016-01-17
Feature 652 old [Do not merge yet] [Incomplete]

Aesylwinn:
Adds gmst verification of type and values. I also moved some of the data stored within the file so that it was not duplicated. I wasn't really happy with my implementation for making the data shared, but I can't think of a better way to do this. I applied fix 3132 to the data as well. The limits are just rough estimates that I made after taking a very quick look at the source code of openmw. I figured that they can be easily changed if needed.


apps/opencs/model/world/defaultgmsts.hpp
```diff
@@ -0,0 +1,54 @@
+#ifndef CSM_WORLD_DEFAULTGMSTS_H
+#define CSM_WORLD_DEFAULTGMSTS_H
+
+#include <cstddef>
+
+namespace CSMWorld {
+    class DefaultGMSTs {
```
zinnschlag: Having a class with only static member functions is a bit pointless. Normally you would use a namespace in this case.

apps/opencs/model/tools/gmstcheck.cpp
```diff
@@ -0,0 +1,96 @@
...
+    else if (gmst.mId[0] == 'f')
+    {
+        size_t floatCount = CSMWorld::DefaultGMSTs::getFloatCount();
+        for (size_t i = 0; i < floatCount; ++i)
+        {
+            if (gmst.mId == CSMWorld::DefaultGMSTs::getFloatName(i))
+            {
+                if (gmst.mValue.getType() != ESM::VT_Float)
+                    messages.add(id, gmst.mId + " has an invalid type; it should be a float", 
+                                 "Change the GMST type to a float", CSMDoc::Message::Severity_Error);
```
zinnschlag: I don't think we are generally giving suggestion in error messages how to fix the error. Let's try to be consistent here. Also, the wording could be improved. The type isn't really invalid. It is just incorrect.

apps/opencs/model/tools/gmstcheck.cpp
```diff
@@ -0,0 +1,96 @@
...
+    CSMWorld::UniversalId id (CSMWorld::UniversalId::Type_Gmst, gmst.mId);
+    
+    // Test for empty string
+    if (gmst.mValue.getType() == ESM::VT_String && gmst.mValue.getString().empty())
+        messages.add(id, gmst.mId + " is an empty string", "", CSMDoc::Message::Severity_Warning);
+    
+    // Checking type and limits
+    // optimization - compare it to lists based on naming convention (f-float,i-int,s-string)
+    if (gmst.mId.empty())
+        messages.add(id, "Game Setting has no name", "", CSMDoc::Message::Severity_SeriousError);
```
zinnschlag: If a record had no ID would couldn't load/access it properly. There is no point in checking for that.

apps/opencs/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ opencs_hdrs_noqt (model/doc
 
 
 opencs_units (model/world
-    idtable idtableproxymodel regionmap data commanddispatcher idtablebase resourcetable nestedtableproxymodel idtree infotableproxymodel
+    idtable idtableproxymodel regionmap data commanddispatcher idtablebase resourcetable nestedtableproxymodel idtree infotableproxymodel defaultgmsts
```
zinnschlag: I don't see anything Qt-specific in defaultgmsts. It should go into the _noqt section instead to silence the moc warnings.

apps/opencs/model/tools/gmstcheck.cpp
Aesylwinn: To be honest, I initially thought the hint was used as a tooltip. From what I can see, its only referenced in a search and replace algorithm (so not really used at all). 
I'll fix the wording.

apps/opencs/model/tools/gmstcheck.cpp
Aesylwinn: It was a sanity check that I added because I was unsure if this might be possible in the future. I'll remove it.

apps/opencs/model/tools/gmstcheck.cpp
Aesylwinn: I misunderstood what you were saying at first. I'll change that.

Aesylwinn:
I had an issue with git. I'm going to try and revert that last commit.

Aesylwinn:
I'm not really sure what happened to my git repository or how to fix it. I'm probably just going to make a new branch and pull request with my changes.

psi29a:
You created a new pull request, does this mean that this one can be closed?

zinnschlag:
Yes, it does. Closing ...

PR #878, opened 2016-01-18
feature 652: gmst verifier

Aesylwinn:
Here's the consolidated changes. I apologize for putting this in a new branch, but I really messed up with git. The previous pull request can still be viewed, but I believe I addressed all of the issues.

In order to get the arrays in the header to compile I had to mark them as extern. I also added constants for the sizes. Let me know if I did this right.


zinnschlag:
No worries about the new branch. I closed the other PR. Looks like we are almost there. Still a few small issues left. I will comment on them inline.

apps/opencs/model/tools/gmstcheck.cpp
```diff
@@ -0,0 +1,91 @@
...
+    // optimization - compare it to lists based on naming convention (f-float,i-int,s-string)
+    if (gmst.mId[0] == 'f')
+    {
+        for (size_t i = 0; i < CSMWorld::DefaultGMSTs::FloatCount; ++i)
+        {
+            if (gmst.mId == CSMWorld::DefaultGMSTs::Floats[i])
+            {
+                if (gmst.mValue.getType() != ESM::VT_Float)
+                    messages.add(id, "The type of " + gmst.mId + " is incorrect", 
+                                 "Change the GMST type to a float", CSMDoc::Message::Severity_Error);
```
zinnschlag: Still the issue with the advice on how to fix it. We don't do that. I suggest to completely remove the hint. We don't need it. The error message should specify the required type though, e.g. The type of xxx is incorrect. It should be a float. Or we could be even more explicit. "Expected float type for xxx but found int type".

apps/opencs/model/tools/gmstcheck.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <components/esm/loadgmst.hpp>
+
+#include "../world/idcollection.hpp"
+
+#include "../doc/stage.hpp"
+
+namespace CSMTools
+{
+    /// \brief VerifyStage: make sure that GMSTs are alright
+    class GMSTCheckStage : public CSMDoc::Stage
```
zinnschlag: Actually that would be GmstCheckStage according to our naming conventions.

psi29a:
Awesome work! :)

PR #880, opened 2016-01-19
use trusty

psi29a:
Let's see what Travis says.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ os:
  - linux
 # - osx
 language: cpp
+sudo: required
+dist: trusty
 branches:
   only:
     - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ before_script:
  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then ./CI/before_script.osx.sh; fi
 script:
  - cd ./build
- - if [ "$COVERITY_SCAN_BRANCH" != 1 ]; then ${ANALYZE}make -j2; fi
+ - if [ "$COVERITY_SCAN_BRANCH" != 1 ]; then ${ANALYZE}make -j3; fi
  - if [ "$COVERITY_SCAN_BRANCH" != 1 ] && [ "${TRAVIS_OS_NAME}" = "osx" ]; then make package; fi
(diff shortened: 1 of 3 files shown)


psi29a:
18 minutes to compile and run unit tests! :)

psi29a:
This will bump our travis builds to using Trusty (14.04). This has the implication that 0.39 will be fully Trusty tested during the CI process. Faster builds and opens the door for additional tools and better C++11 support. We can also stop with having to support Precise and keeping up with all those extra packages.

scrawl:
> Faster builds

https://travis-ci.org/OpenMW/openmw/builds/103302888 vs https://travis-ci.org/OpenMW/openmw/builds/103307806

Doesn't seem faster to me.

Btw, we could drop this workaround:
<pre>
sudo apt-get install -qq cmake-data #workaround for broken osgqt cmake script in ubuntu 12.04
</pre>

psi29a:
Comparison between Precise and Trusty VMs:
Boot Time	50-145s,		20-52s
Memory		~3 GB,		7.5 GB
Cores		~2 bursted,	2

So, in theory the builds should be faster. Obviously having an additional CPU core might make more of an impact.

psi29a:
I've managed to cut down the time from 21 minutes to 17 minutes. That is a small win right? ;)

scrawl:
> Comparison between Precise and Trusty VMs:
> Boot Time 50-145s, 20-52s
> Memory ~3 GB, 7.5 GB
> Cores ~2 bursted, 2
> 
> So, in theory the builds should be faster. Obviously having an additional CPU core might make more of an impact.

Are you sure about that?

https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future says that on travis-ci.<i>org</i>, precise and trusty use the same hardware. I guess there is a difference between travis-ci.org and travis-ci.com where the one is for FOSS and the other for companies.

We were using 3 jobs previously, but had occasional out of memory errors that aborted the build, which is why I changed to 2 jobs.

scrawl:
Try adding a `free -m` to the build so we can see the actual system specs :)

psi29a:
The numbers I cited are from their own blog, but we can verify. :)

psi29a:
total       used       free     shared    buffers     cached
Mem:          7480       1576       5904        149         75       1184
-/+ buffers/cache:        316       7164
Swap:            0          0          0

Looks like 7 GiB to me. :)

psi29a:
Thank you!

PR #881, opened 2016-01-24
Improve start script creation (Fixes #3024)

rcutmore:
This should fix some of the usability issues mentioned in [issue #3024](https://bugs.openmw.org/issues/3024) by making the input a drop target and adding auto-completion. Let me know if you'd like to see any changes to this.

### DIFF
--- a/apps/opencs/view/world/startscriptcreator.cpp
+++ b/apps/opencs/view/world/startscriptcreator.cpp
@@ -75,27 +75,15 @@ std::string CSVWorld::StartScriptCreator::getErrors() const
     std::string errors;
     if (scriptId.empty())
     {
-        if (!errors.empty())
-        {
-            errors += "<br>";
-        }
-        errors += "No Script ID entered";
+        errors = "No Script ID entered";
     }
     else if (getData().getScripts().searchId(scriptId) == -1)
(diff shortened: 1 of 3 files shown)


apps/opencs/view/world/startscriptcreator.cpp
```diff
@@ -1,20 +1,130 @@
...
-        return errors;
-    else if (getData().getScripts().searchId(getId()) == -1)
-        errors = "Script ID not found";
-    else if (getData().getStartScripts().searchId(getId()) > -1 )
-        errors = "Script with this ID already registered as Start Script";
+    // Check user input for any errors.
+    std::string errors;
+    if (scriptId.empty())
+    {
+        if (!errors.empty())
```
EmperorArthur: You define errors on Line 75, but never initialize it before checking it here.
Looking further down the code I can see the basic idea of only adding "<br>" when needed, but I'm not sure if this is the best way to go about it.

rcutmore: Thanks for catching this, the way I had it set up didn't make sense. I updated this to use a simple assignment for each error check since only one error will ever be returned.

PR #882, opened 2016-01-25
Some fixes for building on FreeBSD

t6:
Hi,

I'm the maintainer of the OpenMW FreeBSD port. There are some patches that need to be applied to build OpenMW on FreeBSD and I'd like to reduce the diff a little to make future updates of the port easier.

Thanks!

### DIFF
--- a/apps/openmw/mwworld/projectilemanager.hpp
+++ b/apps/openmw/mwworld/projectilemanager.hpp
@@ -3,6 +3,7 @@
 
 #include <string>
 
+#include <osg/observer_ptr>
 #include <osg/ref_ptr>
 #include <osg/PositionAttitudeTransform>
 
--- a/apps/essimporter/converter.hpp
+++ b/apps/essimporter/converter.hpp
@@ -121,7 +121,7 @@ public:
         {
             mContext->mPlayer.mObject.mCreatureStats.mLevel = npc.mNpdt52.mLevel;
             mContext->mPlayerBase = npc;
-            std::map<const int, float> empty;
+            std::map<int, float> empty;
             // FIXME: player start spells and birthsign spells aren't listed here,
             // need to fix openmw to account for this
             for (std::vector<std::string>::const_iterator it = npc.mSpells.mList.begin(); it != npc.mSpells.mList.end(); ++it)
--- a/apps/essimporter/convertinventory.cpp
+++ b/apps/essimporter/convertinventory.cpp
@@ -1,6 +1,7 @@
 #include "convertinventory.hpp"
 
 #include <components/misc/stringops.hpp>
+#include <cstdlib>
 
 namespace ESSImport
 {
--- a/apps/openmw/crashcatcher.cpp
+++ b/apps/openmw/crashcatcher.cpp
@@ -24,7 +24,7 @@
 #ifndef PR_SET_PTRACER
 #define PR_SET_PTRACER 0x59616d61
 #endif
-#elif defined (__APPLE__)
+#elif defined (__APPLE__) || defined (__FreeBSD__)
 #include <signal.h>
 #endif
 
--- a/apps/openmw/mwmechanics/spells.cpp
+++ b/apps/openmw/mwmechanics/spells.cpp
@@ -44,7 +44,7 @@ namespace MWMechanics
     {
         if (mSpells.find (spell)==mSpells.end())
         {
-            std::map<const int, float> random;
+            std::map<int, float> random;
 
             // Determine the random magnitudes (unless this is a castable spell, in which case
             // they will be determined when the spell is cast)
--- a/apps/openmw/mwmechanics/spells.hpp
+++ b/apps/openmw/mwmechanics/spells.hpp
@@ -33,7 +33,7 @@ namespace MWMechanics
 
             typedef const ESM::Spell* SpellKey;
 
-            typedef std::map<SpellKey, std::map<const int, float> > TContainer; // ID, <effect index, normalised random magnitude>
+            typedef std::map<SpellKey, std::map<int, float> > TContainer; // ID, <effect index, normalised random magnitude>
             typedef TContainer::const_iterator TIterator;
 
             struct CorprusStats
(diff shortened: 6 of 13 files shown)


apps/openmw/mwworld/projectilemanager.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include <string>
 
+#include <osg/observer_ptr>
```
scrawl: Why do we need to include observer_ptr when it's not being used in this file?

t6: We don't. This could be moved to ` apps/openmw/mwworld/projectilemanager.cpp`.

Here is the error I get when it is missing from either file:
```
In file included from ../apps/openmw/mwworld/projectilemanager.cpp:1:
In file included from ../apps/openmw/mwworld/projectilemanager.hpp:4:
In file included from /usr/include/c++/v1/string:439:
In file included from /usr/include/c++/v1/algorithm:627:
/usr/include/c++/v1/utility:253:9: error: implicit instantiation of undefined template 'osg::observer_ptr<osg::Camera>'
    _T1 first;
        ^
/usr/include/c++/v1/map:629:16: note: in instantiation of template class 'std::__1::pair<const osg::observer_ptr<osg::Camera>, std::__1::vector<SceneUtil::LightManager::LightSourceViewBound, std::__1::allocator<SceneUtil::LightManager::LightSourceViewBound> > >' requested here
    value_type __cc;
               ^
/usr/include/c++/v1/__tree:1440:14: note: in instantiation of template class 'std::__1::__value_type<osg::observer_ptr<osg::Camera>, std::__1::vector<SceneUtil::LightManager::LightSourceViewBound, std::__1::allocator<SceneUtil::LightManager::LightSourceViewBound> > >' requested here
    if (__nd != nullptr)
             ^
/usr/include/c++/v1/__tree:1433:5: note: in instantiation of member function 'std::__1::__tree<std::__1::__value_type<osg::observer_ptr<osg::Camera>, std::__1::vector<SceneUtil::LightManager::LightSourceViewBound, std::__1::allocator<SceneUtil::LightManager::LightSourceViewBound> > >, std::__1::__map_value_compare<osg::observer_ptr<osg::Camera>, std::__1::__value_type<osg::observer_ptr<osg::Camera>, std::__1::vector<SceneUtil::LightManager::LightSourceViewBound, std::__1::allocator<SceneUtil::LightManager::LightSourceViewBound> > >, std::__1::less<osg::observer_ptr<osg::Camera> >, true>, std::__1::allocator<std::__1::__value_type<osg::observer_ptr<osg::Camera>, std::__1::vector<SceneUtil::LightManager::LightSourceViewBound, std::__1::allocator<SceneUtil::LightManager::LightSourceViewBound> > > > >::destroy' requested here
    destroy(__root());
    ^
/usr/include/c++/v1/__tree:1119:78: note: in instantiation of member function 'std::__1::__tree<std::__1::__value_type<osg::observer_ptr<osg::Camera>, std::__1::vector<SceneUtil::LightManager::LightSourceViewBound, std::__1::allocator<SceneUtil::LightManager::LightSourceViewBound> > >, std::__1::__map_value_compare<osg::observer_ptr<osg::Camera>, std::__1::__value_type<osg::observer_ptr<osg::Camera>, std::__1::vector<SceneUtil::LightManager::LightSourceViewBound, std::__1::allocator<SceneUtil::LightManager::LightSourceViewBound> > >, std::__1::less<osg::observer_ptr<osg::Camera> >, true>, std::__1::allocator<std::__1::__value_type<osg::observer_ptr<osg::Camera>, std::__1::vector<SceneUtil::LightManager::LightSourceViewBound, std::__1::allocator<SceneUtil::LightManager::LightSourceViewBound> > > > >::~__tree' requested here
    template <class, class, class, class> friend class _LIBCPP_TYPE_VIS_ONLY map;
                                                                             ^
/usr/local/include/osg/ref_ptr:21:28: note: template is declared here
template<typename T> class observer_ptr;
                           ^
```

psi29a: I also wondered about why you needed to remove const from a few places.

t6: If I don't I get errors like:
```
In file included from ../apps/essimporter/importer.cpp:32:
../apps/essimporter/converter.hpp:128:79: error: no viable overloaded '='
                mContext->mPlayer.mObject.mCreatureStats.mSpells.mSpells[*it] = empty;
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~
/usr/include/c++/v1/map:889:10: note: candidate function not viable: no known conversion from 'map<int, [...], less<int>, [...]>' to 'const map<const int, [...], less<const int>, [...]>' for 1st argument
    map& operator=(const map& __m)
         ^
```

psi29a: What version of GCC or Clang are you using?

I'm able to use both without a problem with Debian userland.

t6: FreeBSD 10.2 comes with Clang 3.4.1 and LLVM's libc++.

psi29a: I'm using clang version 3.6.2-1, but I don't know if that will make a difference.

psi29a: In OSX, I'm using clang 3.2 without a problem. :s

t6: My guess is that the culprit here is FreeBSD's libc++ rather than Clang. When I created the port I tried playing around with different Clang and GCC versions but that lead to a lot of other pain (not related to OpenMW code).

Replacing `const int` with `int` was the most straightforward non-hacky solution to those compiler errors I could find.

scrawl: > /usr/include/c++/v1/utility:253:9: error: implicit instantiation of undefined template 'osg::observer_ptr<osg::Camera>'

`osg::observer_ptr<osg::Camera>` is used in components/sceneutil/lightmanager.hpp, so the include should be added there.

https://github.com/OpenMW/openmw/commit/aec8c3846103da0d65ca772d859c501f2d6d9c57

PR #883, opened 2016-01-26
Unlinking qt too

Dreysman:
There is another version of qt in homebrew, so if boost is unlinked, then qt should also be unlinked. I don't know everything about this project, so there might be a reason for why qt is not also unlinked?

### DIFF
--- a/CI/before_install.osx.sh
+++ b/CI/before_install.osx.sh
@@ -6,4 +6,5 @@ export CC=clang
 brew tap openmw/openmw
 brew update
 brew unlink boost
+brew unlink qt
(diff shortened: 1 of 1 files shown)


scrawl:
> There is another version of qt in homebrew, so if boost is unlinked, then qt should also be unlinked. I don't know everything about this project, so there might be a reason for why qt is not also unlinked?

Question for @corristo

corristo:
That's a good question. I suspect that only boost was preinstalled on Travis OS X images, that's why I unlinked it. Thanks for noticing the inconsistency.

Anyway, homebrew-based dependency workflow is deprecated & CI for OS X is temporary disabled, so there's no point in such changes.

PR #885, opened 2016-01-27
Feature 929: Info verifier

Aesylwinn:
This is the first part, the journal info verifier. I still have to work on the dialogue info verifier.
Zini mentioned allowing 0 or 1 when checking for names, but I have no clue what he was referring to.


zinnschlag:
You should only check if there are multiple names for a quest/journal topic. Having one is obviously okay. But having zero is also okay, because this feature was introduced only in the expansions and the base game quests do not have a name.

zinnschlag:
If you are planning to do separate verifiers for journal and regular topics, it might be a good idea to write a base class that factors out common code.

Aesylwinn:
Okay, I understand what you meant now. From what I see, the "Finished" infos are also this way, so should I not test for this or should I only report it if the the journal or its infos were modified.

zinnschlag:
I am not quite sure about Finished, but from your findings it is probably in the same camp as the quest name. No point reporting on it then.

Aesylwinn:
First, a couple of oddities I found in the base/expansion data:
A cell field assigned to "Wilderness"
A faction field assigned to "FFFF"
Several sound files are referenced that appear to be missing

I'm pretty sure they are just errors, but I figured I would mention them in case I'm wrong.

For topic info script verification, I was thinking I should directly add on to the script verifier. Would this be preferred, or should I find some way to make the relevant code shared?

On a side note, the code for interpreting the info conditions is in nestedcoladapterimp.cpp so I'll probably be messing with that file and moving the relevant code elsewhere.

zinnschlag:
> A cell field assigned to "Wilderness"

IIRC that's the name given to a cell not in any region. Not an error in my opinion.

Your other findings sound like errors and should be reported.

> For topic info script verification, I was thinking I should directly add on to the script verifier. Would this be preferred, or should I find some way to make the relevant code shared?

I suggest to skip it. If a dialogue script is valid depends on which NPC it is running on, which in turn depends of the current game state. Therefore we can not reliably check the correctness of dialogue scripts statically.

zinnschlag:
btw. Wilderness isn't hardcoded. You should instead check against the content of the GMST sDefaultCellName.

Aesylwinn:
Forgot to comment when I posted this. Anyway, for the info conditions I could potentially validate the values to some extent. I avoided it because it might get messy dealing with all the operators: >,>=,<,<=,=,!=.

zinnschlag:
We want to check at least some of these fields. Some we can't do (like local variables). But with the rest we should deal.

Aesylwinn:
My knowledge is severely lacking when  it comes to checking the condition functions. Many of them appeared to operate as a sort of boolean or a positive value, so I split them into those cases. It may be wrong.

For journals, I was wondering if it was possible to set a journal to an index even if no info exists for it. For example, there is a Journal condition requirement  "> 50" but that journal only has two infos set at 0.

Aesylwinn:
I'm not sure if I misunderstand the journal condition requirements, or if there are just many errors for them in the base/expansions. Does there need to be a journal info with a quest index that meets the condition?

zinnschlag:
You can always look up the source (apps/openmw/mwdialogue) if you are unsure about what a field does. And yes, there are that many errors in the vanilla esms (we need to find a way to deal with that eventually, but that is beyond this task). I don't understand your question regarding journal infos.

scrawl:
> For journals, I was wondering if it was possible to set a journal to an index even if no info exists for it. For example, there is a Journal condition requirement "> 50" but that journal only has two infos set at 0.

Yes, you can, and the base game actually relies on that in order to work.

https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwdialogue/quest.cpp#L58

Aesylwinn:
> You can always look up the source (apps/openmw/mwdialogue)
I did look there, but it wasn't closely enough. Looks like what I need is in the selectwrapper class. It's a little frustrating that the data is represented differently  :/. I'll go ahead and implement the changes when I get time.

> Yes, you can, and the base game actually relies on that in order to work.
Thanks, that was really puzzling me. I'll just drop the check for that then.

Aesylwinn:
Well, I was making quite a few changes to the way I was testing the info conditions, and I realized that I should probably implement some of the changes to the editor itself. Since that goes beyond the scope of the topic verifier feature, I'm going to be closing this pull request.

zinnschlag:
> Well, I was making quite a few changes to the way I was testing the info conditions, and I realized that I should probably implement some of the changes to the editor itself

Could you elaborate on that?

Also, we haven't merged the PR yet. I was under the impression that the verifier is not complete but in a working state, What is the reason for closing?

Aesylwinn:
>Could you elaborate on that?

Sorry, I should have explained more. I had previously mentioned that I might make move some stuff out of nestedcoladapterimp.cpp. After looking at it, I had decided against it because I noticed a few issues with the implementation, and didn't want it to get mixed in with this. Since this project is the first time I've used the Qt library, I'm unsure of how much I might need to change.

Some problems I noticed: Info Records cannot be reverted (when modified or deleted), Info Conditions share a field for function and variable name and there is no auto complete for either, all created info conditions can only be assigned integer values, also there are all the fixme comments

My plan is to create a wrapper for info conditions. The somewhat complicated method of decoding would take place in one area. Also, shared enums and string conversions for the rules would be available to both. If I put the valid ranges for functions there, I can use that for validation, and for assigning default values. Additionally, I need to know whether a function will be a boolean, integer, or numeric comparison. That might be useful for knowing when the user can enter integers or floats.

>What is the reason for closing?

It is in a mostly working state, but it would be affected by any changes I made above. When I said it went beyond the scope of the info verifier, I meant that what I wanted to do didn't fit within this specific feature. I've been pretty bad about isolating my commits so far :( . My plan is to make the wrapper and some revisions to the adapter. After than I can resurrect this pull request. I could also make a new pull request if that would be preferred.

zinnschlag:
It will be interesting to see what you come up with. But yes, properly splitting commits into feature branches would make things a lot easier for all of us. I don't really care how you organise it in regards to PRs.

PR #890, opened 2016-02-03
Refactor local script iteration (Bug #2806)

scrawl:
This should be much safer. Don't use recursion. Don't fail if mIgnore happens to be in the list twice. Don't rely on preconditions / assertions.

Still not sure about the exact cause of bug 2806, but I presume it had to do with a Ptr being in the list twice for some reason. I've added a warning message so we can catch that the next time it happens.

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -69,11 +69,9 @@ void OMW::Engine::executeLocalScripts()
     MWWorld::LocalScripts& localScripts = mEnvironment.getWorld()->getLocalScripts();
 
     localScripts.startIteration();
-
-    while (!localScripts.isFinished())
+    std::pair<std::string, MWWorld::Ptr> script;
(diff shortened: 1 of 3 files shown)


scrawl:
> Still not sure about the exact cause of bug 2806, but I presume it had to do with a Ptr being in the list twice for some reason.

Looks like that is it -I'm now getting the warning every time I boot into the game with --skip-menu at cell (0,0)
<pre>
warning, tried to add local script twice for kagouti_diseased
warning, tried to add local script twice for kagouti_diseased
</pre>

scrawl:
Levelled creatures are having their local script added twice (once as part of safePlaceObject and once as part of the cell's InsertVisitor). Looks like we'll have to change the way this spawning is done.

scrawl:
c9a67ab42357bdb28d63e243946dc734a9030610

PR #891, opened 2016-02-05
Add OpenMW CS manual to the repository.

HiPhish:
The manual depends on Sphinx to build. A makefile and a Windows batch
script are included for building, both were automatically generated by
Sphinx. I have left all settings at their default and I have tested that
HTML and PDF output build properly.


psi29a:
I don't see a Makefile, just the make.bat file. Did you forget to add this and other files?

Can you get rid of the 'cs-manual' directory, it isn't necessary. We might (and probably will) use sphinx for more than just your documentation. So a little bit of reordering is necessary here.

What I mean is that make.bat, Makefile and other bits should exist in /docs not /docs/cs-manual so shift everything back by one directory. How it looks when rendered is not dictated by directory structure but by layout of the index.rst and cascading rst files. We'll then build off from that.

HiPhish:
I see now why the makefile is missing: there is an entry `Makefile` in the project-wide `.gitignore`. Should I add an exception?
```
!docs/Makefile
```

zinnschlag:
Sure. Add the exception.

But I noticed that you placed your stuff directly into the docs directory. Please don't do that. There is already other stuff in there and things will get messy if we mix them up. Use a subdirectory as discussed on the forum.

HiPhish:
Oh come on, that's what I had done first and then @psi29a told me otherwise. I moved it back now.

psi29a:
@HiPhish my apologies, @zinnschlag overules me. :(

I would rename cs-manual to sphinx then, because I expect that we'll be using sphinx for more than just what you've done.

/docs/sphinx/ 

You OK with that Zini?

zinnschlag:
Sorry for the confusion. I think using the purpose of the directory instead of the name of the tool used in this directory is a better choice. If we have other stuff we want to add that is also using Sphinx we can easily make a new subdirectory alongside cs-manual.

Will merge.

psi29a:
We can do that yes, but if we add additional sub-directories under /docs then we have multiple instances of sphinx. If we ever publish to readthedocs.com, you'll have to pick one as it won't publish multiple instances of sphinx in a repo.

It is usually best-practice to have one instance that can handle multiple sets of documentation. Usually this the whole of /docs, but since you insist on not colliding with existing documentation such as licenses, then it is a good idea to have a sub-directory under docs dedicated to sphinx generated documentation.

zinnschlag:
I see your point, but I still think it is not a good idea to name subdirectories after the tool used instead of the purpose. We don't even know yet if we are going to use Sphinx for other things. Maybe leave it as it is for now and if new requirements come up, rename it?

HiPhish:
Quick question: should I keep writing the manual this separate branch `cs-manual` or should I write directly on master? I was thinking about keeping working on the branch and every time a PR gets merged I `fetch upstream master` and then `rebase upstream/master` my local branch.

zinnschlag:
We prefer people to work in feature branches, though it does not matter much in this case.

HiPhish:
@zinnschlag OK, but how do I proceed from now? In Particular, should I delete this branch, `checkout master`, `fetch upstream/master`, `checkout -b next-chater`? I tried that, but when I pulled from upstream I got an extra commit (merge branch upstream master or something like that) in addition to your commit.

psi29a:
Normal work-flow is to have you feature branch as Zini said. Then make a merge request that will go into master. 

If you want to keep working on your branch after it was merged then:

git checkout master
git pull upstream master
git checkout -b next-chapter
git rebase master

or you can create a totally new branch

git checkout master
git pull upstream master
git checkout -b chapter-2

HiPhish:
Thanks, I just figured it out the moment you posted your message: I had the default git configuration set to not do fast-forward merges, that's why the extra commit was always added in after the merge.

What does the `git rebase master` do? If I checkout a new branch (because of `-b` flag) from `master` shouldn't it already have `master` as its parent?

psi29a:
You're right, but rebase is in case you are going to reuse the same branch. A rebase will get your "old" branch up to date with master.

HiPhish:
OK, that makes sense.

PR #892, opened 2016-02-10
Render cell markers (Feature #3043)

rcutmore:
This adds rendering of cell markers. The markers just display the cell's ID (i.e. "#-1 -1") and use the cell's center as a pivot point. They can be turned on/off using the existing toggle button as well. Let me know if there is missing functionality that needs to be added or if I'm off-track at all.

I was able to find some screenshots of various work in progress versions of the previous markers, but I wasn't sure what the final version looked like. Below are some screenshots for this PR:

cell-markers-2
cell-markers

### DIFF
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -35,7 +35,7 @@ opencs_hdrs_noqt (model/world
 
 
 opencs_units (model/tools
-    tools reportmodel mergeoperation 
+    tools reportmodel mergeoperation
     )
 
 opencs_units_noqt (model/tools
@@ -90,7 +90,7 @@ opencs_units (view/render
 
 opencs_units_noqt (view/render
     lighting lightingday lightingnight
-    lightingbright object cell terrainstorage tagbase cellarrow
+    lightingbright object cell terrainstorage tagbase cellarrow cellmarker
     )
 
 opencs_hdrs_noqt (view/render
@@ -192,6 +192,7 @@ endif(APPLE)
 target_link_libraries(openmw-cs
     ${OSG_LIBRARIES}
     ${OPENTHREADS_LIBRARIES}
+    ${OSGTEXT_LIBRARIES}
     ${OSGUTIL_LIBRARIES}
     ${OSGVIEWER_LIBRARIES}
     ${OSGGA_LIBRARIES}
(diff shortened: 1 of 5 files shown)


scrawl:
Looks good :+1:

zinnschlag:
That needs some testing. I'll get back to you later today.

zinnschlag:
Looks mostly good. Two issues:

1. The # in front of the labels needs to be removed. We use it only as a marker for the cell ID. Outside of that it does not have any use.
2. The labels need some kind of background (coloured rectangle or something). Remember, the main function of the cell marker is to manipulate cells from within the 3D scene view. Having just a piece of text makes it very hard to click on. I suggest to use different colours for the background, depending on if the cell exists or not. Alternatively a different shape might work too.

This brings us to the interactive functionality of the cell markers. This wasn't listed in the original issue description, because we didn't have the mouse picking code back then. We can file another issue for this sub-feature (and you move on to another task) or you can just continue with implementing it.
The basic idea here is that clicking on the marker with the primary/secondary edit button will open a context menu with the following items (as applicable to the state of the cell):

* Remove from Scene View
* Select All (this affects the scene elements in the cell that are matching the current edit mode; right now only instance editing is implemented)
* Clear Selection (as above)
* Create Cell
* Delete Cell
* Revert Cell

rcutmore:
Okay, I'll work on updates to the marker itself, follow up for some
feedback, and go from there.

rcutmore:
@zinnschlag Is this closer to what you had in mind? I added a black rectangle around the markers for cells that exist and a red marker around those that do not. Let me know if you have a preference for different colour choices.

Also, right now I'm using the *mDeleted* property of Cell to determine whether or not it exists. Is this correct or is it more involved than that?

cell-markers

zinnschlag:
Looks good. Maybe add a , between the x and y value?

> Also, right now I'm using the mDeleted property of Cell to determine whether or not it exists. Is this correct or is it more involved than that?

You get the deleted state via the isDeleted function of the record (or via the modification state column). mDelete might not always be up to date.

rcutmore:
@zinnschlag When you have a chance let me know how the latest commit looks. Before creating the marker the cell's record is retrieved and its *isDeleted* function is used to determine the cell's existence. I also made a few changes to the marker's appearance.

If everything looks good I think it might be best to merge these changes and I'll look at the interactive functionality in the near future. I have some personal matters that need attention so I'm not sure what my availability will be over the next couple weeks or so.

scrawl:
Should these markers show in interior cells?
marker

zinnschlag:
Interiors have only a single cell. I don't see much use for a marker there.

rcutmore:
I'll submit a PR to not show markers for interior cells.

PR #893, opened 2016-02-10
Cell pre-loading

scrawl:
- Refactored resource management classes to make them thread-safe.
- Unload resources when they have not been used for a while.
- Introduced CellPreloader that uses a background thread to preemptively load resources found in close-by cells. 
- Introduced UnrefQueue to move most of the cell un-loading work to the background thread.

New settings (probably don't need to touch these unless you have a low end machine):
<pre>
# Preload cells in a background thread. All settings starting with 'preload' have no effect unless this is enabled.
preload enabled = true

# Preload adjacent cells when moving close to an exterior cell border.
preload exterior grid = true

# Preload possible fast travel destinations.
preload fast travel = true

# Preload the locations that doors lead to.
preload doors = true

# Preloading distance threshold
preload distance = 1000

# How long to keep preloaded cells and cached models/textures/collision shapes in cache
# after they're no longer referenced/required (in seconds)
cache expiry delay = 300
</pre>


psi29a:
That's a first... works great on appveyor but not on travis!

psi29a:
error: invalid use of incomplete type ‘class osg::UserDataContainer’

zinnschlag:
Very nice. Will merge.

Do we have an issue on the tracker for this feature?

zinnschlag:
Forget the question. Found it: https://bugs.openmw.org/issues/2482#change-14681

sandstranger:
It is very often crashes on Android with
#01 pc 0000000000b8ace0  /data/app/com.libopenmw.openmw-4/lib/arm64/libopenmw.so (SceneUtil::WorkThread::run()+44)
02-13 16:20:33.202 22153-22153/? I/AEE/AED:     #02 pc 00000000008775dc  /data/app/com.libopenmw.openmw-4/lib/arm64/libopenmw.so (OpenThreads::ThreadPrivateActions::StartThread(void*)+164)

sandstranger:
engine still crashes even with
preload enabled = false

psi29a:
@sandstranger what version of OSG are you using? You'll need to use 3.4 to fix bug in uncompressed dds textures. Maybe that is the rason?

sandstranger:
I am using osg 3.4.0 version , but I use png textures instead dds, because osg not loading uncompressed dds textures correctly

psi29a:
That was fixed in 3.4

sandstranger:
Just checked it. All uncompressed  dds textures still loaded with white color

psi29a:
Then this is another bug, unrelated to crash bug. :/

sandstranger:
This bug works only on Opengl es1

PR #894, opened 2016-02-10
hide touch camera on Android

sandstranger:
With this PR , touch camera hiding on Android automatically

### DIFF
--- a/apps/openmw/mwgui/windowmanagerimp.cpp
+++ b/apps/openmw/mwgui/windowmanagerimp.cpp
@@ -110,6 +110,10 @@
 #include "controllers.hpp"
 #include "jailscreen.hpp"
 
+#ifdef ANDROID
+#include "components/android/androidtouchcamerahelper.hpp"
+#endif
+
 namespace MWGui
 {
 
@@ -1118,6 +1122,9 @@ namespace MWGui
     void WindowManager::setCursorVisible(bool visible)
     {
         mCursorVisible = visible;
+#ifdef ANDROID
+        AndroidTouchCameraHelper::hideTouchCamera(visible);
+#endif
     }
 
     void WindowManager::onRetrieveTag(const MyGUI::UString& _tag, MyGUI::UString& _result)
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -28,6 +28,12 @@ endif()
 
 # source files
 
+if (ANDROID)
+    add_component_dir (android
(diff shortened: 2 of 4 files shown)


scrawl:
Can we handle this outside of the windowmanager code? E.g. by querying ```windowManager->getCursorVisible()``` every frame and set the touch state accordingly. If we can do that, can we move the code into the android-port repository just like the rest of the Java code?

sandstranger:
Yes it is possible .we can move all  this code in t android-port repository and build it as a separate library ,because android studio can work with c ++ code . I'll do it then

sandstranger:
I use my c++ class as
class ControlsHider : public MWGui::WindowManager {
public:
    ControlsHider();
    void* runBackgroundTask(void* arg);
    bool getCursorVisible() {
        MWGui::WindowManager::getCursorVisible();
    }
    void hideControls(bool needHideControls);
    void backgroundTask();

};:
but bool getCursorVisible() always return true even in the game . What am I doing wrong ?

Ragora:
@sandstranger
Looks like you're not actually returning the result of the window manager get cursor visible.

sandstranger:
even with
bool getCursorVisible() {
        return MWGui::WindowManager::getCursorVisible();
    }
it always return true . wtf

Ragora:
That's odd, I haven't poked OpenMW in ages. I would say it's either not doing quite what you think or its bugged somewhere. Would have to wait on @scrawl or some such then. I saw the notification and the issue initially seemed simple enough (I make mistakes like that all the time personally).

sandstranger:
maybe I get an incorrect reference to the class, because in buiding of this class I include already prebuilt openmw library with MWGui::WindowManager class

scrawl:
You'd need to get the existing instance of the game's WindowManager:
<pre>
MWBase::Environment::get().getWindowManager()
</pre>

sandstranger:
It is crashes , when I trying to get instance

scrawl:
When are you trying to get it? You would need to wait until the Engine::prepareEngine() has completed.

sandstranger:
Solved : It is crashes , because using java threads for getting cursor thread was very bad idea, because jni method must be called every frame on main java thread.
I will try to use c++ pthreads for it

sandstranger:
Ok , with this commit everything working fine 
https://github.com/OpenMW/android-port/commit/55bf00060326d74d04556ba67ec5e44435e9bf8a

PR #895, opened 2016-02-12
Remove support for OSG 3.2

scrawl:
Since commit e8662bea3133ba9dbb09b86c3abb1af39425e90d, we're using OSG functionality that contains an unfixed crash bug in version 3.2. The bug is fixed in version 3.4 (OSG commit 6351e5020371b0b72b300088a5c6772f58379b84)

The crash would happen with a certain type of DDS texture (uncompressed with builtin mipmaps). The vanilla Morrowind game does not include such textures, but openmw-template does and certain mods might contain these also.

See http://forum.openmw.org/viewtopic.php?f=2&t=3351&start=30#p37410

I suspect that http://forum.openmw.org/viewtopic.php?f=2&t=3365&p=37475#p37475 and https://bugs.openmw.org/issues/3204#change-14700 are also the same issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,7 +233,7 @@ if (USE_QT)
     set (OSG_QT osgQt)
 endif()
 
-find_package(OpenSceneGraph 3.2.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle ${OSG_QT} osgUtil osgFX)
+find_package(OpenSceneGraph 3.4.0 REQUIRED osgDB osgViewer osgText osgGA osgAnimation osgParticle ${OSG_QT} osgUtil osgFX)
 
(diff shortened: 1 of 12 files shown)


psi29a:
Boooo! ;)

psi29a:
We can forget Debian releases and Ubuntu releases for now.

zinnschlag:
Well, that is a problem. Can we maybe add OSG 3.4 to the OpenMW repository?

psi29a:
I'm working on it, it compiles on launchpad but complains about an invalid pointer in a osg plugin. End result is that it won't upload and considers the build a failure.

https://forum.openmw.org/viewtopic.php?f=2&t=3351&p=37508

I don't want to disable this plugin since this is the basis for what is going into Debian.. so it needs to be addressed eventually. If there is a patch for it, I can add it to get us through.

PR #896, opened 2016-02-12
Write chapters about tables and record types.

HiPhish:
The "Tables", "Record Types" and "Record Filters" chapters have been
adapted from the already existing manual.

### DIFF
--- a/docs/cs-manual/source/record-filters.rst
+++ b/docs/cs-manual/source/record-filters.rst
@@ -1,0 +1,293 @@
+Record Filters
+##############
+
+Filters are a key element of the OpenMW CS user interface, they allow rapid and
+easy access to records presented in all tables. In order to use this
+application effectively you need to familiarise yourself with all the concepts
+and instructions explained in this chapter. The filter system is somewhat
+unusual at first glance, but once you understand the basics it will be fairly
+intuitive and easy to use
+
+Filters are a key element to using the OpenMW CS efficiently by allowing you to
+narrow down the table entries very quickly and find what you are looking for.
+The filter system might appear unusual at first, you don't just type in a word
+and get all instances where it occurs, instead filters are first-class objects
+in the CS with their own table. This allows you to define very specific filters
+for your project and store them on disc to use in the next session. The CS
+allows you fine-grained control, you can choose whether to make a filter
+persistent between session, only for one session or use a one-off filter by
+typing it directly into the filter field.
+
+
+
+Terms used
+**********
+
+Filter
+   A Filter is generally speaking a tool able to filter the elements of a
+   table, that is select some elements while discarding others, according to
+   some criteria. These criteria are written using their own syntax.
+
+Criterion
+   A criterion describes some condition a record needs to satisfy in order to
+   be selected. They are written using a special syntax which is explained
+   below. We can logically combine multiple criteria in a filter for finer
+   control.
+
+Expression
+   Expressions are how we perform filtering. They look like functions in a
+   programming language: they have a name and accept a number of arguments.
+   The expression evaluates to either ``true`` or ``false`` for every record in
+   the table. The arguments are expressions themselves.
+
+Arity
+   The arity of an expression tells us how many arguments it takes. Expressions
+   taking no arguments are called *nullary*, those taking one argument are
+   known as *unary* expressions and those taking two arguments are called
+   *binary*.
+
+
+
+Interface
+*********
+
+Above each table there is a text field which is used to enter a filter: either
+one predefined by the OpenMW CS developers or one  made by you. Another
+important element is the filter table found under *View* → *Filters*. You
+should see the default filters made by the OpenMW team in the table. The table
+has the columns *Filter*, *Description* and *Modified*.
+
+ID
+   A unique name used to refer to this filter. Note that every ID has a
+   scope prefix, we will explain these soon.
+
+Modified
+   This is the same as for all the other records, it tells us whether the
+   filter is *added* or *removed*. Filters are specific to a project instead of
+   a content file, they have no effect on the game itself.
+
+Filter
+   The actual contents of the filter are given here using the filter syntax.
+   Change the expressions to modify what the filter returns.
+
+Description
+   A textual description of what the filter does.
+
+
+
+Using predefined filters
+************************
+
+To use a filter you have to type its ID into the filter field above a table.
+
+For instance, try to opening the objects table (under the world menu) and type
+into the filters field ``project::weapons``. As soon as you complete the text
+the table will show only the weapons. The string ``project::weapons`` is the ID
+of one of the predefined filters. This means that in order to use the filter
+inside the table you type its name inside the filter field.
+
+Filter IDs follow these general conventions:
+
+- IDs of filters for a specific record type contain usually the name of a
+  specific group. For instance the ``project::weapons`` filter contains the
+  term ``weapons``. Plural form is always used.
+
+- When filtering a specific subgroup the ID is prefixed with the name of the
+  more general filter. For instance ``project::weaponssilver`` will filter only
+  silver weapons and ``project::weaponsmagical`` will filter only magical
+  weapons.
+
+- There are few exceptions from the above rule. For instance there are
+  ``project::added``, ``project::removed``, ``project::modified`` and
+  ``project::base``. You might except something more like
+  ``project::statusadded`` but in this case requiring these extra characters
+  would not improve readability.
+
+We strongly recommend you take a look at the filters table right now to see
+what you can filter with the defaults. Try using the default filters first
+before writing you own.
+
+
+
+Writing your own filters
+************************
+
+As mentioned before, filters are just another type of record in the OpenMW CS.
+To create a new filter you will have to add a new record to the *Filters* table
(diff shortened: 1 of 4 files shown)


HiPhish:
This means my manual is now content-wise up to par with the previous one. I can now focus on writing new content. I'll take a while off to see how to approach it first before writing anything down, so please don't think I have abandoned the manual. You can still remind me of it if you think I'm taking too long.

HiPhish:
I forgot to mention, I have completely stripped out the part about regular expressions from the *filters* chapter. That's outside the scope of this manual and just adds bloat.

docs/cs-manual/source/record-filters.rst
```diff
@@ -0,0 +1,277 @@
...
+elements is the filter table found under *View* → *Filters*. You should see the
+default filters made by the OpenMW team in the table. The table has the columns
+*Filter*, *Description* and *Modified*.
+
+ID
+   A unique name used to refer to this filter. Note that every ID has a
+   namespace prefix, we will explain these soon.
+
+Modified
+   This is the same as for records, except that it applies to filters. It tells
```
zinnschlag: Shouldn't we call that "other records" instead? Filters are after all records like any other.

docs/cs-manual/source/record-filters.rst
```diff
@@ -0,0 +1,277 @@
...
+   us whether the filter is base (*untouched*), *modified* or *removed*.
+   Filters are specific to a project instead of a content file, they have no
```
zinnschlag: Actually, filter can not have base state and neither can they have modified or moved. We may allow at some point the inclusion of filters in the content files. But for now the only possible modification state is added.

docs/cs-manual/source/record-filters.rst
```diff
@@ -0,0 +1,277 @@
...
+==========
+
+Every default filter has the prefix ``project``. This is a *namespace*, a term
+borrowed from the C++ programming language along with the syntax. In the case
+of the OpenMW CS a namespace always means the scope of said object [#]_. The
+meaning of namespaces is as follows:
+
+``project::``
+   Indicates that the filter is to be used throughout the project in multiple
+   sessions. You can restart the CS and the filter will still be there.
```
zinnschlag: Maybe mention here that records in the project namespace are not saved with in the content file? That is the meaning of the namespace after all. Records that are stored in an internal project file (for the content file) that is specific to the machine/user account OpenMW-CS is run on.

docs/cs-manual/source/record-filters.rst
```diff
@@ -0,0 +1,277 @@
...
+
+``session::``
+   Indicates that the filter is not stored between multiple sessions and once
+   you quit the OpenMW CS application the filter will be gone. Until then it
+   can be found inside the filters table.
+
+In addition the these two scopes there is a third one called *one-shot*.
```
zinnschlag: Note that one-shot is a term I coined on the fly. If you can think of something better, feel free to use it.

zinnschlag: Also, technically one-shot isn't a scope, because one-shot filters aren't actually records. They are just a piece of text.

docs/cs-manual/source/record-filters.rst
```diff
@@ -0,0 +1,277 @@
...
+argument is the table column we want to test and the second argument is the
+value we want to test it against.
+
+Strings are sequences of characters and are case-insensitive. If a string
+contains spaces it must be quoted, otherwise the quotes are optional and
+ignored.
+
+
+Atomic Expressions
+------------------
```
zinnschlag: Calling that atomic is weird. Maybe use "Constant Expressions" instead?

docs/cs-manual/source/record-filters.rst
```diff
@@ -0,0 +1,277 @@
...
+since it still lacks expressions. In order to add your queries you have to edit
+the filter record.
+
+
+Replacing the default filters set
+=================================
+
+OpenMW CS allows you to substitute the default filters set provided by the
+developers your own filters. In order to do so you have to create a new
+project, add the desired filters, remove the undesired ones and save. Rename
```
zinnschlag: This is a bit misleading. You don't create a project. Any content file that you edit is implicitly a project.

docs/cs-manual/source/record-filters.rst
```diff
@@ -0,0 +1,277 @@
...
+the file to ``defaultfilters`` (remove the ``.omwaddon.project`` file
+extension) and place it inside your configuration directory.
```
zinnschlag: Also this mixes things up. You can replace the project filters as described before inside OpenMW-CS. Replacing the defaultfilters file is what would actually change the defaults for all content files that have not been edited on this computer/user-account before. Since this manual is only about the basics, maybe that should be taken out since it is an advanced task?

docs/cs-manual/source/record-types.rst
```diff
@@ -0,0 +1,64 @@
...
+*Objects* table contains information about each of these template objects, eg.
+its value and weight in the case of items, or an aggression level in the case
+of NPCs.
+
+Let’s go through all Record Types and discuss what you can tell OpenMW CS about
+each of them.
+
+Activator
+   When the player enters the cell containing this object a script is started.
+   Often it also has a Script attached to it, though it is not mandatory. These
```
zinnschlag: This is a bit confusing. Makes it sounds as if there were two scripts involved. What actually happens is that the script attached to the activator (not mandatory indeed) will run once per frame as long as the cell the activator is in is active.

docs/cs-manual/source/tables.rst
```diff
@@ -0,0 +1,177 @@
...
+
+
+Used Terms
+**********
+
+Record
+   An entry in OpenMW CS representing an item, location, sound, NPC or anything
+   else.
+
+Reference, Referenceable
```
zinnschlag: This is the old terminology that we don't use anymore. References are called instances now and referencables are called objects.

docs/cs-manual/source/tables.rst
```diff
@@ -0,0 +1,177 @@
...
+   When an item is placed in the world, it does not create a new record each
+   time. For example, the game world might contain a lot of exquisite belts on
```
zinnschlag: Technically it does create a new sub-record of type instance, which is actually handled just like a regular record in OpenMW-CS. We don't necessarily have to confuse the user with the details, but a less misleading wording would be helpful.

docs/cs-manual/source/tables.rst
```diff
@@ -0,0 +1,177 @@
...
+   console and writint: ``player- >addItem Caius_pants``. Either way, in both
+   Morrowind and OpenMW CS, the ID is the primary way to identify all these
+   different parts of the game.
+
+Modified
+   This column shows what has happened (if something has happened) to this
+   record. There are four possible states in which it can exist.
+
+Base
+   This record is part of the base game and is in its original state. Usually
```
zinnschlag: This is the wrong definition of base. Base is the collection of all the content files loaded except the one you are currently editing.

docs/cs-manual/source/tables.rst
```diff
@@ -0,0 +1,177 @@
...
+Name
+   This is how the game will show your location in-game.
+
+MapColour
+   This is a six-digit hexadecimal representation of the colour used to
+   identify the region on the map available in *World* → *Region Map*. If you
+   do not have an application with a colour picker, you can use your favourite
+   search engine to find a colour picker online.
+
+.. TODO  What does the last sentence above mean?
```
zinnschlag: Nothing. It is redundant now.

docs/cs-manual/source/tables.rst
```diff
@@ -0,0 +1,177 @@
...
+   Should water be rendered in this interior cell? The game world consists of
+   an endless ocean at height 0, then the landscape is added. If part of the
+   landscape goes below height 0, the player will see water. (See
+   illustration.)
+
+   .. TODO  Add illustration mentioned above
+
+   Setting the cell’s Interior Water to true tells the game that this cell is
+   both an interior cell (inside a cave, for example, rather than in the open
+   air) but that there still needs to be water at height 0. This is useful for
```
zinnschlag: Actually, the flag only indicates that there is water if it is an interior cell. It does not make the cell interior and for exterior cells it is useless.

zinnschlag:
Besides the points I commented on inline, the section about filters should also mention drag and drop.

HiPhish:
@zinnschlag I copy-pasted the text from Sirherbatka and only corrected the most blatant mistakes, my priority was getting what we already had to work. I'll go over your points soon.

zinnschlag:
Okay. Its not really surprising that the original text doesn't do it anymore since it is really outdated. But it seems a few things slipped through the original revision. Sorry, my mistake probably.

HiPhish:
@zinnschlag  I have updated my commit with your suggestions.

scrawl:
@zinnschlag what happened to this one?

zinnschlag:
Wow, really old documentation PR that somehow got lost in the noise. Looks still good to me. @psi29a Any objections to go ahead with the merge?

psi29a:
merge conflict... 
docs/source/openmw-cs/index.rst

PR #897, opened 2016-02-12
Crashcatcher: limit backtrace to a sensible number of stack frames

scrawl:
When a stack overflow occurs, trying to print the whole stack would cause the process to hang indefinitely.


kcat:
I'm surprised gdb doesn't automatically detect bad/looping call stacks and cut output. You'll want to keep the ```full``` specifier though, since that makes it try to print local variables.

scrawl:
> You'll want to keep the full specifier though, since that makes it try to print local variables.

Hm, I thought the 'full' just stands for 'all frames', but that is not the case. Updated to keep the ```full``` keyword.

PR #898, opened 2016-02-13
Physics

scrawl:
- Decouples physics framerate from the rendering framerate to always run at 60 fps. This makes the simulation behave more predictably and should reduce the risk of "phasing" through collision objects that could previously happen with low frame rates. See various reported bugs at https://bugs.openmw.org/issues/1587
- Perform interpolation in between physics frames. This reduces the 'shaking' effects that can occur due to the camera rotation updating at a different rate than the (physics controlled) camera position. ( https://bugs.openmw.org/issues/2737 )

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -270,7 +270,7 @@ namespace MWBase
             virtual MWWorld::Ptr moveObject (const MWWorld::Ptr& ptr, float x, float y, float z) = 0;
             ///< @return an updated Ptr in case the Ptr's cell changes
 
-            virtual MWWorld::Ptr moveObject(const MWWorld::Ptr &ptr, MWWorld::CellStore* newCell, float x, float y, float z) = 0;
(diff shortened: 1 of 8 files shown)


sandstranger:
With this commit , game runs with 3 fps on Android

scrawl:
That will happen when the CPU can't keep up with processing 60 physics frames per second as requested, it will fall behind and queue up more and more frames.

You can lower the ```const int maxAllowedSteps = 20;``` variable to reduce this effect, but the physics will run in slow motion either way (i.e. as if time was slowed down).

Aesylwinn:
Another solution would be to change the time step to something larger, such as 1/30 of a second. This would cause more collision penetration to occur, but also reduce the workload and reduce the chance that a physics update would be dropped. The `maxAllowedSteps` should probably be lowered to a more reasonable number. I remember reading a [bullet wiki](http://bulletphysics.org/mediawiki-1.5.8/index.php/Stepping_the_World) page several years ago that suggested 7.

sandstranger:
Yes, setting physicsDt to 1.f/30.0 value resolve issue with 3 fps
Maybe it will be better set physicsDt to 1.f/30.0  for Android ,because arm processors - shit

PR #899, opened 2016-02-14
Raise the required bullet version to 2.83

scrawl:
psi29a is going to like this one. More stuff to backport to our PPA.

Bullet 2.82 appears to have a bug that causes the player to be able to phase through certain objects (bug #1587).

### DIFF
--- a/CI/before_install.linux.sh
+++ b/CI/before_install.linux.sh
@@ -11,11 +11,11 @@ sudo apt-get update -qq
 sudo apt-get install -qq libgtest-dev google-mock
 sudo apt-get install -qq libboost-filesystem-dev libboost-program-options-dev libboost-system-dev libboost-thread-dev
 sudo apt-get install -qq libavcodec-dev libavformat-dev libavutil-dev libswscale-dev libswresample-dev
-sudo apt-get install -qq libbullet-dev libopenscenegraph-dev libmygui-dev libsdl2-dev libunshield-dev libtinyxml-dev libopenal-dev qtbase5-dev
+sudo apt-get install -qq libbullet-dev libopenscenegraph-dev libmygui-dev libsdl2-dev libunshield-dev libtinyxml-dev libopenal-dev libqt4-dev
 if [ "${ANALYZE}" ]; then sudo apt-get install -qq clang-3.6; fi
 sudo mkdir /usr/src/gtest/build
 cd /usr/src/gtest/build
-sudo cmake .. -DBUILD_SHARED_LIBS=1 -DDESIRED_QT_VERSION=5
+sudo cmake .. -DBUILD_SHARED_LIBS=1
 sudo make -j4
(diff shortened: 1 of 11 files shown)


zinnschlag:
Not overly happy either. Will wait with the merge until we get the all clear from @psi29a

psi29a:
Die in a fire... :'(

The good news is that 2.83 is in stretch/sid and should also be in Ubuntu Xenial, so it isn't as bad as OSG.

psi29a:
Good news: https://launchpad.net/~openmw/+archive/ubuntu/openmw/+packages
OSG 3.4 is available for all i386 and x86_64 across our 3 currently supported Ubuntu Releases.
Bullet 2.83 had to be backported to Trusty, but is also up.

@scrawl Please do an empty commit to trigger a new build to verify that Travis is happy.

scrawl:
Bullet is ok, but there was this:

<pre>
In file included from /usr/include/osgQt/GraphicsWindowQt:21:0,

                 from /home/travis/build/OpenMW/openmw/apps/opencs/view/render/scenewidget.cpp:9:

/usr/include/osgQt/Version:8:2: error: #error "Qt version mismatch detected! Make sure to compile applications using osgQt with the same major Qt version that osgQt has been compiled against."

 #error "Qt version mismatch detected! Make sure to compile applications using osgQt with the same major Qt version that osgQt has been compiled against."
</pre>

psi29a:
For Qt, I'm using what is default. I'm guessing that for Debian, the default was Qt5. We should probably go that way as well with Launcher and CS.

scrawl:
Last I checked, Qt5 had ugly font rendering on Linux and there were some other regressions. osgQt can't do threaded rendering when using Qt5.

The OpenMW/osg fork uses Qt4 by default.

scrawl:
Ultimately I plan on doing a manual qt integration in favor of osgQt, so that this dependency hell goes away. Not sure when I'll get around to that though.

zinnschlag:
I consider Qt4 our main platform for now (its certainly what I am using). As mentioned by scrawl there are still issues with Qt5 and I don't see the benefits of a move to Qt5. It doesn't offer a lot of improvements for us and seems overall not quite mature enough.

psi29a:
:'(  It will take awhile to get these packages rebuilt then, need to find the right options.

scrawl:
Not needed anymore, I've added osgQt to our repository in commit 647a5e091f4298ed390e511d7ce9fbfa64e0aa95 , this means it is built independently of OSG so it doesn't matter what version of Qt OSG was built with.

psi29a:
Thank you @scrawl

corristo:
@scrawl would Qt5 compatibility be kept?

scrawl:
Yes, when building OpenMW we still have the choice of Qt4 or Qt5.

PR #900, opened 2016-02-16
Remove optimization and debug flags on Travis-CI to speed up the build

scrawl:
From my local tests, this should be almost 2x faster. Let's see how it does on Travis-CI.


scrawl:
https://travis-ci.org/OpenMW/openmw/jobs/109657484
https://travis-ci.org/OpenMW/openmw/jobs/109606970

Down from 19 minutes to 12 minutes, I call that a win.

PR #901, opened 2016-02-16
[Not for merging] Empirical testing for AppVeyor build performance

ace13:


ace13:
*Release* build timed out after an hour, definitely slower than *Debug*.
*RelWithDebInfo* next.

ace13:
*RelWithDebInfo* also times out one hour into compiling, also definitely slower than *Debug*.
Time to finish the testing off with *MinSizeRel*

scrawl:
What about *None*, is that not available on VS?

ace13:
Build type *None* doesn't exist on VS, no. If you try to force something like that you just end up with *Debug*.

ace13:
From this testing I can conclude that the only build type fast enough to be able to complete in time, is *Debug*.

PR #902, opened 2016-02-17
Topic info editor improvement and verifier

Aesylwinn:
I still need to go over the acceptable values for the verifier and the autofill feature for conditions in the editor. Also, I need to double check that my wrapper is reading and writing the data correctly.

Other than what I mentioned above, the result is pretty much what I was thinking of. There are some major changes to how the editor works. Let me know what you think.

### DIFF
--- a/apps/opencs/model/world/infoselectwrapper.cpp
+++ b/apps/opencs/model/world/infoselectwrapper.cpp
@@ -1,0 +1,855 @@
+#include "infoselectwrapper.hpp"
+
+#include <limits>
+#include <sstream>
+
+const size_t CSMWorld::ConstInfoSelectWrapper::RuleMinSize = 5;
+
+const size_t CSMWorld::ConstInfoSelectWrapper::FunctionPrefixOffset = 1;
+const size_t CSMWorld::ConstInfoSelectWrapper::FunctionIndexOffset = 2;
+const size_t CSMWorld::ConstInfoSelectWrapper::RelationIndexOffset = 4;
+const size_t CSMWorld::ConstInfoSelectWrapper::VarNameOffset = 5;
+
+const char* CSMWorld::ConstInfoSelectWrapper::FunctionEnumStrings[] =
+{
+    "Rank Low",
+    "Rank High",
+    "Rank Requirement",
+    "Reputation",
+    "Health Percent",
+    "PC Reputation",
+    "PC Level",
+    "PC Health Percent",
+    "PC Magicka",
+    "PC Fatigue",
+    "PC Strength",
+    "PC Block",
+    "PC Armorer",
+    "PC Medium Armor",
+    "PC Heavy Armor",
+    "PC Blunt Weapon",
+    "PC Long Blade",
+    "PC Axe",
+    "PC Spear",
+    "PC Athletics",
+    "PC Enchant",
+    "PC Detruction",
+    "PC Alteration",
+    "PC Illusion",
+    "PC Conjuration",
+    "PC Mysticism",
+    "PC Restoration",
+    "PC Alchemy",
+    "PC Unarmored",
+    "PC Security",
+    "PC Sneak",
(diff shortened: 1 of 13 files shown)


Aesylwinn:
Good to know. I'll go ahead and change it for now. Would that be related to feature 890?

apps/opencs/model/world/infoselectwrapper.cpp
```diff
@@ -0,0 +1,893 @@
...
+        // Integer
+        case Function_Journal:
+        case Function_Item:
+        case Function_Dead:
+        case Function_RankLow:
+        case Function_RankHigh:
+        case Function_RankRequirement:
+        case Function_Reputation:
+        case Function_PcReputation:
+        case Function_PcLevel:
```
Aesylwinn: Should I change the valid range for player/actor level to be greater or equal to 1?

apps/opencs/model/world/infoselectwrapper.cpp
```diff
@@ -0,0 +1,893 @@
...
+        case Function_PcStrength:
+        case Function_PcBlock:
+        case Function_PcArmorer:
+        case Function_PcMediumArmor:
+        case Function_PcHeavyArmor:
+        case Function_PcBluntWeapon:
```
Aesylwinn: For player skills, should I change the valid range to be 5 or greater?

apps/opencs/model/world/infoselectwrapper.cpp
```diff
@@ -0,0 +1,893 @@
...
+        case Function_RankLow:
+        case Function_RankHigh:
+        case Function_Reputation:
+        case Function_PcReputation:
+        case Function_Journal:
+            return std::pair<int, int>(IntMin, IntMax);
+
+        case Function_Item:
+        case Function_Dead:
+        case Function_PcLevel:
```
Aesylwinn: For player/actor level, should I change the valid range to be 1 or greater?

apps/opencs/model/world/infoselectwrapper.cpp
```diff
@@ -0,0 +1,893 @@
...
+        case Function_PcStrength:
+        case Function_PcBlock:
+        case Function_PcArmorer:
+        case Function_PcMediumArmor:
+        case Function_PcHeavyArmor:
+        case Function_PcBluntWeapon:
+        case Function_PcLongBlade:
```
Aesylwinn: For skills, should I change the valid range to be 5 or greater?

Aesylwinn:
Other than my line comments, I consider the topic info verifier to be complete.

zinnschlag:
I suggest to ask these questions on the forum, because I am not quite sure which values would work here.

Aesylwinn:
It's a bit of a hack, but I got the auto completion working. Perhaps I should add another class that inherits from CommandDelegate to accomplish this?

Aesylwinn:
I re-added the journal verifier. I changed the name of the class and some of the error messages.

zinnschlag:
Works nicely. I would consider this feature complete. Is there anything you want to add or should we go ahead with the merge?

Aesylwinn:
You can go ahead with the merge.

zinnschlag:
Merged. Seems we finally have most of the verifiers out of the way. Let me know if you need help finding a new task.

psi29a:
yay! \o/

Thank you for this.

Aesylwinn:
:)
@zinnschlag I'm thinking of doing one of the rendering tasks if there's not too many people working on those. If you have a suggestion for a good one to start with, I'd appreciate it. Also, if you think a different area needs some attention, I can work on that instead.

zinnschlag:
The cell marker task is done (minus the popup menu). Everything else is still available. There is of course also the camera task, which is rather important, but also big and a bit tricky.

PR #910, opened 2016-03-02
feature-1279: render cell border markers

Aesylwinn:
Instead of marking corners I chose to draw an outline of the cell edges. One issue with this method is that the rocks jutting out of the terrain can hide it. Potentially I could make it draw on top of everything, but I don't know how to change the draw order for osg.


apps/opencs/view/render/cellborder.cpp
```diff
@@ -0,0 +1,97 @@
...
+        vertices->push_back(osg::Vec3f(scaleToWorld(x), scaleToWorld(y), landData->mHeights[landIndex(x, y)]));
+
+    geometry->setVertexArray(vertices);
+
+    // Color
+    osg::ref_ptr<osg::Vec4Array> colors = new osg::Vec4Array();
+    colors->push_back(osg::Vec4f(0.f, 0.5f, 0.f, 1.f));
+
+    geometry->setColorArray(colors);
+    geometry->setColorBinding(osg::Geometry::BIND_PER_PRIMITIVE_SET);
```
scrawl: setColorBinding is a deprecated function, instead use
<pre>
geometry->setColorArray(colors, osg::Array::BIND_PER_PRIMITIVE_SET); 
</pre>

Aesylwinn:
I could also add the vertical lines at the corners.

Aesylwinn:
I'm not an artist, so the icon is nothing fancy. It's just less of a placeholder. 
I looked into making the grid draw over everything, and the only method I found involved making another camera which doesn't seem worthwhile for this feature.

scrawl:
>  I looked into making the grid draw over everything, and the only method I found involved making another camera which doesn't seem worthwhile for this feature.

FWIW, you could do that by disabling the depth test and changing the StateSet's RenderBin to a high number.

zinnschlag:
That looks really good. Even the icon is okay. I don't think any further enhancements are required. Will merge.

Aesylwinn:
> FWIW, you could do that by disabling the depth test and changing the StateSet's RenderBin to a high number.
I'll keep that in mind for future use.

PR #911, opened 2016-03-03
(Sub)view refactoring without fix

Pi03k:
https://bugs.openmw.org/issues/2570
While analyzing code I've found some duplicate code, typos, zeros used to initialize pointers. I briefly tested it.


zinnschlag:
For future reference, we do not prefer NULL over 0. Some developers use NULL, but most of the code was written using 0 instead. Eventually that will all be replaced by nullptr. But for now changing existing code from 0 to NULL serves little purpose and only introduces additional potential for merge conflicts.

scrawl:
I like to use NULL for pointers and 0 for everything else. This should make the transition to nullptr a little bit easier. Agreed with the merge conflicts though. With minor style changes, just limit the changes to code that you were already changing anyway.

PR #913, opened 2016-03-06
Fix incorrect handling of baseDisposition >= 100 in testDisposition()

ae-g-i-s:
`MechanicsManager::getDerivedDisposition()` currently bounds its result to 0..100 (after applying faction modifications etc to the actor's `baseDisposition`).
`MWDialogue::Filter::testDisposition()`, which is used to filter candidate topics by disposition checks, uses this bounded value and applies the current `mTemporaryDispositionChange` of this dialogue (retrieved via `MWBase::Environment::get().getDialogueManager()->getTemporaryDispositionChange()`) to it.

If `baseDisposition` is >= 100, `MechanicsManager::getDerivedDisposition()` will usually return 100. If there is a topic that tests for a disposition, e.g. 70, and `mTemporaryDispositionChange` is < -30 (through failed Persuasion, for instance), the test will always fail, no matter how high `baseDisposition` is. See http://bugs.openmw.org/issues/3233 for a reproducer that calls `ModDisposition 20` every time a certain topic is clicked, enabling the user to increase their `baseDisposition` indefinitely.

This PR adds an optional parameter `bounded` to `MechanicsManager::getDerivedDisposition()`, set to `true` by default. If it is set to `false` (which is only the case in `MWDialogue::Filter::testDisposition()` as of now), it no longer bounds its return value to 0..100, but 0..`std::numeric_limits<int>::max()`.

I tested it for about two hours during normal play with a little regression test that I hope is correct; no errors were reported.
> int actorDisposition = MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mActor, false) + MWBase::Environment::get().getDialogueManager()->getTemporaryDispositionChange();
>
> // XXX Remove this - regression test
> int derivedDispositionOld = MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mActor, true);
> int actorDispositionOld = derivedDispositionOld + MWBase::Environment::get().getDialogueManager()->getTemporaryDispositionChange();
> if( (actorDisposition != actorDispositionOld) && (derivedDispositionOld != 100))
> {
>     std::cerr << "Possible REGRESSION: New disposition " << actorDisposition << " != " << actorDispositionOld << " (old) and derivedDispositionOld = " << derivedDispositionOld << std::endl;
> }
> // XXX End regression test

### DIFF
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -97,7 +97,7 @@ namespace MWBase
             virtual int getBarterOffer(const MWWorld::Ptr& ptr,int basePrice, bool buying) = 0;
             ///< This is used by every service to determine the price of objects given the trading skills of the player and NPC.
 
-            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr) = 0;
+            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr, bool bounded = true) = 0;
             ///< Calculate the diposition of an NPC toward the player.
 
             virtual int countDeaths (const std::string& id) const = 0;
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -97,7 +97,7 @@ namespace MWBase
             virtual int getBarterOffer(const MWWorld::Ptr& ptr,int basePrice, bool buying) = 0;
             ///< This is used by every service to determine the price of objects given the trading skills of the player and NPC.
 
-            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr) = 0;
+            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr, bool addTemporaryDispositionChange = false) = 0;
             ///< Calculate the diposition of an NPC toward the player.
 
             virtual int countDeaths (const std::string& id) const = 0;
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -97,7 +97,7 @@ namespace MWBase
             virtual int getBarterOffer(const MWWorld::Ptr& ptr,int basePrice, bool buying) = 0;
             ///< This is used by every service to determine the price of objects given the trading skills of the player and NPC.
 
-            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr, bool bounded = true) = 0;
+            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr) = 0;
             ///< Calculate the diposition of an NPC toward the player.
 
             virtual int countDeaths (const std::string& id) const = 0;
--- a/apps/openmw/mwgui/dialogue.cpp
+++ b/apps/openmw/mwgui/dialogue.cpp
@@ -640,8 +640,7 @@ namespace MWGui
         if(mMainWidget->getVisible() && mEnabled && mPtr.getTypeName() == typeid(ESM::NPC).name())
         {
             int disp = std::max(0, std::min(100,
-                MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mPtr)
-                    + MWBase::Environment::get().getDialogueManager()->getTemporaryDispositionChange()));
+                MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mPtr, true)));
             mDispositionBar->setProgressRange(100);
             mDispositionBar->setProgressPosition(disp);
             mDispositionText->setCaption(MyGUI::utility::toString(disp)+std::string("/100"));
--- a/apps/openmw/mwgui/tradewindow.cpp
+++ b/apps/openmw/mwgui/tradewindow.cpp
@@ -346,8 +346,7 @@ namespace MWGui
             else
                 d = int(100 * (b - a) / a);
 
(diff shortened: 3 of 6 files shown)


scrawl:
I just created a test plugin with a dialogue line that triggers at a disposition of 110, loaded it up in the vanilla engine, then tried to get this dialogue to execute by various means of raising disposition. I couldn't get that to work, which makes me think that the vanilla engine does clamp the final disposition value.

How about doing the clamping <i>after</i> the temporaryDispositionChange is added? Something like this:

<pre>
int MechanicsManager::getDerivedDisposition(const MWWorld::Ptr& ptr, bool addTemporaryDispositionChange)
....

if (addTemporaryDispositionChange)
   x +=  dialogueManager->getTemporaryDispositionChange();
int effective_disposition = std::max(0,std::min(int(x),100));//, normally clamped to [0..100] when used
return effective_disposition;
</pre>

As an added bonus, this solution would simplify the API (just one getDerivedDisposition call instead of getDerivedDisposition+getTemporaryDispositionChange), and should be less error prone.

ae-g-i-s:
Yeah, that sounds sane. I'll code it up this evening. I'll also have to refactor a bit of the code in MechanicsManager as another function is doing exactly that same thing (`getDerivedDisposition() + getTemporaryDispositionChange()`) already.

ae-g-i-s:
Done, adjusted the following callers:
* `Filter::testDisposition()`
* `MechanicsManager::getBarterOffer()`
* `DialogueWindow::onFrame()`
* `TradeWindow::onOfferButtonClicked()`

Only tested it briefly, but the diff says that it should be correct.

ae-g-i-s:
Huh. Travis says:
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

I don't think this is related to my changes, could you restart the build?

apps/openmw/mwbase/mechanicsmanager.hpp
```diff
@@ -97,7 +97,7 @@ namespace MWBase
             virtual int getBarterOffer(const MWWorld::Ptr& ptr,int basePrice, bool buying) = 0;
             ///< This is used by every service to determine the price of objects given the trading skills of the player and NPC.
 
-            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr) = 0;
+            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr, bool addTemporaryDispositionChange = false) = 0;
```
scrawl: I  think it would be better for the parameter default to be ```true```, since that's what most callers use. The only exception where we need ```false``` is when the temporary disposition change is recalculated in DialogueManager::persuade.

apps/openmw/mwgui/dialogue.cpp
```diff
@@ -640,8 +640,7 @@ namespace MWGui
         if(mMainWidget->getVisible() && mEnabled && mPtr.getTypeName() == typeid(ESM::NPC).name())
         {
             int disp = std::max(0, std::min(100,
-                MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mPtr)
-                    + MWBase::Environment::get().getDialogueManager()->getTemporaryDispositionChange()));
+                MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mPtr, true)));
```
scrawl: The min/max here is no longer necessary, ```getDerivedDisposition``` does it already.

apps/openmw/mwgui/tradewindow.cpp
```diff
@@ -346,8 +346,7 @@ namespace MWGui
             else
                 d = int(100 * (b - a) / a);
 
-            int clampedDisposition = std::max(0, std::min(MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mPtr)
-                + MWBase::Environment::get().getDialogueManager()->getTemporaryDispositionChange(),100));
+            int clampedDisposition = std::max(0, std::min(MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mPtr, true),100));
```
scrawl: Same here...

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -667,10 +670,9 @@ namespace MWMechanics
         MWWorld::Ptr playerPtr = getPlayer();
         const MWMechanics::NpcStats &playerStats = playerPtr.getClass().getNpcStats(playerPtr);
 
-        // I suppose the temporary disposition change _has_ to be considered here,
+        // I suppose the temporary disposition change (second param to getDerivedDisposition()) _has_ to be considered here,
         // otherwise one would get different prices when exiting and re-entering the dialogue window...
-        int clampedDisposition = std::max(0, std::min(getDerivedDisposition(ptr)
-            + MWBase::Environment::get().getDialogueManager()->getTemporaryDispositionChange(),100));
+        int clampedDisposition = std::max(0, std::min(getDerivedDisposition(ptr, true),100));
```
scrawl: ... and here.

scrawl:
I clicked on restart build, still failed. Seems like the server is having a hiccup.

psi29a:
I just slapped Travis again, lets see what happened. ;)

scrawl:
I made some minor cleanups and merged it. How would you like to be listed in AUTHORS.md - fullname, nickname or both?

ae-g-i-s:
Sorry, dropped the ball on this one, I had your comments in a local change, but didn't commit it yet bc of work stress...
Just aegis would be perfect. I'll see if there are any other gameplay bugs I can tackle.

PR #916, opened 2016-03-14
Editor camera

Aesylwinn:
I made a new pull request to get rid of the merge conflicts.

Some known issues:
1. The camera needs to be intelligently placed near loaded cells on scene startup
2. The OrbitCameraController lookAtCenter is garbage, I need to figure out the math
3. Input and tuning of some values needs to be done

### DIFF
--- a/apps/opencs/view/render/scenewidget.cpp
+++ b/apps/opencs/view/render/scenewidget.cpp
@@ -12,7 +12,6 @@
 #include <osgViewer/ViewerEventHandlers>
 #include <osg/LightModel>
 #include <osg/BoundingBox>
-#include <osg/ComputeBoundsVisitor>
 
 #include <osgGA/TrackballManipulator>
 #include <osgGA/FirstPersonManipulator>
@@ -315,7 +314,7 @@ void SceneWidget::update(double dt)
     }
     else
     {
-        osg::ComputeBoundsVisitor boundsVisitor;
+        CameraComputeBoundsVisitor boundsVisitor(Mask_Reference | Mask_Terrain);
         osg::BoundingBox &boundingBox(boundsVisitor.getBoundingBox());
 
         mRootNode->accept(boundsVisitor);
--- a/apps/opencs/view/render/scenewidget.cpp
+++ b/apps/opencs/view/render/scenewidget.cpp
@@ -328,7 +328,45 @@ void SceneWidget::update(double dt)
 
 void SceneWidget::settingChanged (const CSMPrefs::Setting *setting)
 {
-    storeMappingSetting(setting);
+    if (*setting=="3D Scene Input/p-navi-sensitivity")
+    {
+        mFreeCamControl->setCameraSensitivity(setting->toDouble());
+        mOrbitCamControl->setCameraSensitivity(setting->toDouble());
+    }
+    else if (*setting=="3D Scene Input/s-navi-sensitivity")
+    {
+        mFreeCamControl->setSecondaryMovementMultiplier(setting->toDouble());
+        mOrbitCamControl->setSecondaryMovementMultiplier(setting->toDouble());
+    }
+    else if (*setting=="3D Scene Input/navi-wheel-factor")
+    {
+        mFreeCamControl->setWheelMovementMultiplier(setting->toDouble());
+        mOrbitCamControl->setWheelMovementMultiplier(setting->toDouble());
+    }
+    else if (*setting=="3D Scene Input/navi-free-lin-speed")
+    {
+        mFreeCamControl->setLinearSpeed(setting->toDouble());
+    }
+    else if (*setting=="3D Scene Input/navi-free-rot-speed")
+    {
+        mFreeCamControl->setRotationalSpeed(setting->toDouble());
+    }
+    else if (*setting=="3D Scene Input/navi-free-speed-mult")
+    {
+        mFreeCamControl->setSpeedMultiplier(setting->toDouble());
+    }
+    else if (*setting=="3D Scene Input/navi-orbit-rot-speed")
+    {
+        mOrbitCamControl->setOrbitSpeed(setting->toDouble());
+    }
+    else if (*setting=="3D Scene Input/navi-orbit-speed-mult")
+    {
+        mOrbitCamControl->setOrbitSpeedMultiplier(setting->toDouble());
+    }
+    else
+    {
+        storeMappingSetting(setting);
+    }
 }
 
 void SceneWidget::selectNavigationMode (const std::string& mode)
--- a/apps/opencs/view/render/scenewidget.cpp
+++ b/apps/opencs/view/render/scenewidget.cpp
@@ -353,12 +353,6 @@ void SceneWidget::selectNavigationMode (const std::string& mode)
         mCurrentCamControl = mOrbitCamControl.get();
         mCurrentCamControl->setCamera(getCamera());
     }
-    else if (mode=="trackball")
-    {
-        mCurrentCamControl->setCamera(NULL);
-        mCurrentCamControl = mOrbitCamControl.get();
-        mCurrentCamControl->setCamera(getCamera());
-    }
 }
 
 bool SceneWidget::storeMappingSetting (const CSMPrefs::Setting *setting)
--- a/apps/opencs/view/render/scenewidget.cpp
+++ b/apps/opencs/view/render/scenewidget.cpp
@@ -164,7 +164,7 @@ SceneWidget::SceneWidget(boost::shared_ptr<Resource::ResourceSystem> resourceSys
     , mPrevMouseX(0)
     , mPrevMouseY(0)
     , mFreeCamControl(new FreeCameraController())
-    , mOrbitCamControl(new OrbitCameraController())
+    , mOrbitCamControl(new OrbitCameraController(mRootNode))
     , mCurrentCamControl(mFreeCamControl.get())
     , mCamPositionSet(false)
 {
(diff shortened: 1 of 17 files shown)


Aesylwinn:
I did drop the old indentation fix and replace it.

zinnschlag:
Sorry, won't have the time to test this today (and probably not tomorrow either). But I'll make sure to give you my feedback still this week.

zinnschlag:
Change of plans. Was able to give it a try after all. Here are my observations:

1. The turn rate in 1st person and free is very fast. Maybe we need to scale that down.
2. Switching from free into 1st person did not move the camera immediately into an orientation that is valid under 1st person.
3. The default orientation for 1st person could be better (when opening a scene in that mode)
4. Orbit is actually hard to test without a reasonable centre point. This is a part of the feature that was missing from the task description, I believe. We need a menu item on the context menu of the active orbit mode button, that lets the player place the centre at the centre of the current selection. Of course we also need to set a reasonable default centre point when opening a scene in orbit mode or switching to orbit mode.

Aesylwinn:
I put in the trackball mode mostly because of the preview widget which added the osg equivalent. Would it be alright to make it use the orbit camera?

2. I originally did that but ran into data corruption for the camera. I'll take a shot at it though.
4. The orbit does go around a point correctly (based on testing without the rotation adjustment), but the rotations are screwed up so that they don't face the center and suffer from gimbal lock. I think I might put in a temporary object to let a person know where the center is located.

I'll see what I can do, but it will be a day or two before I can work on the changes.

zinnschlag:
Orbit is the correct mode for Preview. And no hurry.

Aesylwinn:
Not sure if its a bug, but manipulating the camera view matrix before the first frame() makes osg enter an infinite loop in the cull visitor. Regardless, it's a simple modification to accomodate this behavior.

scrawl:
I don't see this issue (at least not in OpenMW). Can you post a small sample that creates the infinite loop?

<pre>
diff --git a/apps/openmw/engine.cpp b/apps/openmw/engine.cpp
index 62457ca..aa95f87 100644
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -615,6 +615,8 @@ void OMW::Engine::go()
 
     mViewer = new osgViewer::Viewer;
 
+    mViewer->getCamera()->setViewMatrix(osg::Matrix());
+
     osg::ref_ptr<osgViewer::StatsHandler> statshandler = new osgViewer::StatsHandler;
     statshandler->setKeyEventTogglesOnScreenStats(osgGA::GUIEventAdapter::KEY_F3);
 
</pre>

Aesylwinn:
OpenMW does not seem to suffer from this issue for whatever reason. I can reproduce it on my machine by adding the following to my code:
```
--- a/apps/opencs/view/render/cameracontroller.cpp
+++ b/apps/opencs/view/render/cameracontroller.cpp
@@ -98,6 +98,7 @@ namespace CSVRender
     {
         mLockUpright = true;
         mUp = up;
+        stabilize();
     }
 
     void FreeCameraController::unfixUpAxis()
```
Also, I retested it, and found it is not really an infinite loop. It just takes 5 or so seconds on my machine for the input to start being processed. Apologies, I was a bit impatient during my earlier testing. I'll note that I don't run into issues if I am not using the setViewAsMatrixLookAtFunction or manually setting the position/rotation with setTrans() and setRotate().

Aesylwinn:
There is a slight issue with the initial camera positioning, I'm looking into it.

Aesylwinn:
I should probably use a weighted average for selecting the camera position.

Aesylwinn:
When people are using the mouse to control the camera, should it look like they are dragging the terrain or should it look like they are moving the camera?

zinnschlag:
> When people are using the mouse to control the camera, should it look like they are dragging the terrain or should it look like they are moving the camera?

Camera, I would say.

> Also, should the context menu for orbit selection be placed in the worldspace tool mode overlay?

On the camera mode button. The button class has build in support for context menus.

Aesylwinn:
I noticed that some assets appear to have nothing rendered when previewed. For example, the base_animKnA.nif. I'm not sure if it's intentional, but I verified it wasn't caused by my changes.

zinnschlag:
Probably related to this: https://bugs.openmw.org/issues/1221

scrawl:
base_anim*.nif are not supposed to render anything on their own. They contain the basic skeleton that NPC body parts and equipment will be attached to.

Aesylwinn:
I'm working on the context menu, I believe that's all that needs to be added at this point.

Aesylwinn:
Unless there are any critiques, or any more context button options to be added, I think this feature is complete.

apps/opencs/view/render/scenewidget.cpp
```diff
@@ -314,8 +315,9 @@ void SceneWidget::update(double dt)
     }
     else
     {
-        CameraComputeBoundsVisitor boundsVisitor(Mask_Reference | Mask_Terrain);
+        osg::ComputeBoundsVisitor boundsVisitor;
         osg::BoundingBox &boundingBox(boundsVisitor.getBoundingBox());
+        boundsVisitor.setNodeMaskOverride(Mask_Reference | Mask_Terrain);
```
scrawl: I think you want the NodeVisitor::setTraversalMask() function.

apps/opencs/view/render/scenewidget.cpp
```diff
@@ -11,6 +11,10 @@
 #include <osgViewer/CompositeViewer>
 #include <osgViewer/ViewerEventHandlers>
 #include <osg/LightModel>
+#include <osg/BoundingBox>
+
+#include <osgGA/TrackballManipulator>
+#include <osgGA/FirstPersonManipulator>
```
scrawl: These two includes can be removed.

scrawl:
There's some odd behaviour when you attempt to pitch the camera by more than 90 degrees (starts rotating like crazy). We should probably put in a limit so that it just stops there.

scrawl:
I find Orbit a little awkward to use. Is there any way we could keep it upright (no roll) like the OSG OrbitManipulator? Also, the direction of mouse movements seems to be the inverse of what I expected.

Could be just me, so a 2nd opinion would be nice.

Aesylwinn:
> Is there any way we could keep it upright (no roll) like the OSG OrbitManipulator? 
The orbit camera is also used for previewing objects. Is it awkward when using it for that? If so, then perhaps another camera mode should be added.

scrawl:
First Person and Free Camera works for me. Maybe the default camera sensitivity could be increased a bit. I have to move the mouse all the way to the edge of the screen just to rotate by 180°.

In the future we may want to capture mouse input while rotating in 1st person mode. Not sure how technically feasible that is.

scrawl:
> In the future we may want to capture mouse input while rotating in 1st person mode. Not sure how technically feasible that is.

Or during any sort of drag operation, really.

Aesylwinn:
>In the future we may want to capture mouse input while rotating in 1st person mode. Not sure how technically feasible that is.

I'm not quite sure what you mean. Camera input is at the bottom of the input chain. The camera receives its input as long as none of the classes inheriting from the SceneWidget class override the function and don't pass it down. Are you saying that it should be possible to move the camera and move an object at the same time?

scrawl:
What I mean is to restrict the mouse cursor to the scene widget, so that you can keep going in one direction without hitting the edge of the screen.

Aesylwinn:
Oh, I see.

Aesylwinn:
Perhaps something like this: http://www.qtcentre.org/threads/53576-Mouse-lock-inside-window-widget
Though there is the possibility that events would be missed: https://forum.qt.io/topic/24842/mouse-lock-inside-window-widget/4

zinnschlag:
Regarding the orbit camera: It is intentional that the camera is not held upright. If it were, having the camera exactly above or below the centre would be awkward. We could however introduce a new function (camera mode button context menu and later keyboard shortcut) that moves the camera into an upright orientation.

Regarding capturing the mouse cursor: This is problematic. We will also need the ability to drag instances out of the widget (into other widgets). Of course that does not affect the camera movement, but still ...
I suggest we put that on hold until after 1.0, when we have a better idea about what is required here.

scrawl:
Sorry, that wasn't clear. With 'upright', I mean the roll axis. 

orbit_1

orbit_2

zinnschlag:
Right. Not sure how you would want to limit that. But you can always turn around by using the Q and E shortcuts.

Finally got around to do some testing myself. Here are my findings:
- movement in free and 1st person mode seems to be a bit too fast. But one size does not fit all and maybe we should just make that a user setting.
- in orbiting mode the directions feel a bit odd (inverted). But again, one size does not fit all. Maybe another user setting?

Everything else seems to work nicely. Well done.

Aesylwinn:
The numbers used for input scaling are all adjustable in the 3D input user settings. Some of them can be set to negative values which is basically inversion, though perhaps it is not obvious enough for the user.

Aesylwinn:
I think I'll make buttons that invert the mouse scaling and have different scalars for the different camera modes.

Aesylwinn:
I can get the behavior I think scrawl is describing by locking the orbit camera up axis. However, I don't think it really fits with the way the orbit camera is described. I would rather add a "terrain camera" mode that behaves differently control-wise.

Hack: https://github.com/Aesylwinn/openmw/commit/b74ca7bd9cf205256c0b347f0b3add637d7b42eb
The alternate mode is enable/disabled using the orbit camera context menu.

Aesylwinn:
I think everything else has been addressed.

zinnschlag:
Locking the camera would take away the ability to rotate it on purpose. That's not really a solution. We can talk about additional camera modes after 1.0.
Running a final test now. If no further problems show up, I will merge.

PR #917, opened 2016-03-17
Fix for bug 3248

Pi03k:
This is RAII version of this solution available since Qt5.3. Should I use it?
http://doc.qt.io/qt-5/qsignalblocker.html


zinnschlag:
Since our main Qt platform is still Qt 4 that is not an option.

PR #918, opened 2016-03-24
RefNum Assignment

zinnschlag:
This PR moved the assignment of RefNums from the instance creation code to the save code.

* Simplifies the code for creating new instances and avoids code duplication
* Avoids creating unneeded move sub records for newly created instances (when moving new instance from one cell to another within the same editing session in which the instance was created)

Possible downside: If an addon is used for playing between two editor saves, the ref nums may not stay consistent (e.g. if some of the new instances are deleted between saves). This can cause problems when saving a game in OpenMW when playing with the addon after the first editor save and then reloading the game with the addon after the 2nd editor save.
IMO this case is insignificant enough that we do not need to consider it, especially since fixing this would be non-trivial.

I think the ref num logic is correct, but considering how fiddly the whole instance business is, I would like a 2nd opinion before this goes into master and therefore into the imminent 0.39.0 release.

### DIFF
--- a/apps/opencs/model/doc/savingstages.cpp
+++ b/apps/opencs/model/doc/savingstages.cpp
@@ -265,12 +265,28 @@ void CSMDoc::WriteCellCollectionStage::perform (int stage, Messages& messages)
     std::map<std::string, std::deque<int> >::const_iterator references =
         mState.getSubRecords().find (Misc::StringUtils::lowerCase (cell.get().mId));
 
-    if (cell.isModified() || 
+    if (cell.isModified() ||
         cell.mState == CSMWorld::RecordBase::State_Deleted ||
         references!=mState.getSubRecords().end())
     {
         CSMWorld::Cell cellRecord = cell.get();
         bool interior = cellRecord.mId.substr (0, 1)!="#";
+
+        // count new references and adjust RefNumCount accordingsly
+        int newRefNum = cellRecord.mRefNumCounter;
+
+        if (references!=mState.getSubRecords().end())
+        {
+            for (std::deque<int>::const_iterator iter (references->second.begin());
+                iter!=references->second.end(); ++iter)
+            {
+                const CSMWorld::Record<CSMWorld::CellRef>& ref =
+                    mDocument.getData().getReferences().getRecord (*iter);
+
+                if (ref.get().mNew)
+                    ++cellRecord.mRefNumCounter;
+            }
+        }
 
         // write cell data
         writer.startRecord (cellRecord.sRecordId);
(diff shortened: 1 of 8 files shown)


scrawl:
> This PR moved the assignment of RefNums from the instance creation code to the save code.

> Possible downside: If an addon is used for playing between two editor saves, the ref nums may not stay consistent (e.g. if some of the new instances are deleted between saves). This can cause problems when saving a game in OpenMW when playing with the addon after the first editor save and then reloading the game with the addon after the 2nd editor save.

So from what I understand, the previous logic is better from a functionality perspective, but the new one better from a code maintenance perspective? You'll have to decide yourself if that tradeoff is worth it. Perhaps there's another way to simplify/deduplicate the code?

> Avoids creating unneeded move sub records for newly created instances (when moving new instance from one cell to another within the same editing session in which the instance was created)

This could be fixed in a different way - the creation of MVRF tags for cell references in State_ModifiedOnly ("Added") should never be allowed to happen to begin with, right?

zinnschlag:
> This could be fixed in a different way - the creation of MVRF tags for cell references in State_ModifiedOnly ("Added") should never be allowed to happen to begin with, right?

As I understand it that currently can happen and it kinda makes sense. Assume you have an addon file. The file gets published and people create saves with it.
Then the author of the addon makes some changes, including moving an instance created in the addon to another cell.
If we do not have a MVRF here, that will cause problems with the save.

zinnschlag:
However the distinction between having MVRF over multiple editing sessions and not having it within the original editing session seems quite arbitrary and will probably cause more confusion that it will help. So the best approach would be to go with what we have here and modify the code so that added instances never have MVRF tags.

scrawl:
> As I understand it that currently can happen and it kinda makes sense. Assume you have an addon file. The file gets published and people create saves with it.
Then the author of the addon makes some changes, including moving an instance created in the addon to another cell.

I don't remember if the moved ref implementation in OpenMW was written to support this case. Have you tested it?

One the one hand that does sound somewhat useful. On the other hand, the file you are editing will get clogged up with lots of possibly unnecessary tags. If we do save these, there should also be a tool to clean them (i.e. <i>really</i> apply movements of objects within the same plugin, then remove the tags).

zinnschlag:
> I don't remember if the moved ref implementation in OpenMW was written to support this case. Have you tested it?

I don't remember, actually. AFAIK this feature has been in the editor for ages.

> One the one hand that does sound somewhat useful. On the other hand, the file you are editing will get clogged up with lots of possibly unnecessary tags. If we do save these, there should also be a tool to clean them (i.e. really apply movements of objects within the same plugin, then remove the tags). 

You got a point here. But since we are kinda trying to finish things up and get OpenMW-CS ready for 1.0 I think we should shelf this feature for post 1.0 development.

PR #921, opened 2016-04-01
No murder bounty when a player follower commits murder (Bug #2852)

NeveHanter:
This commit fixes [https://bugs.openmw.org/issues/2852](url). I've tested that change and now it properly raises bounty when summon (Skeleton) or the follower (High-Pockets) kills an NPC.


apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1366,11 +1396,12 @@ namespace MWMechanics
...
         const MWMechanics::NpcStats& victimStats = victim.getClass().getNpcStats(victim);
 
-        // Simple check for who attacked first: if the player attacked first, a crimeId should be set
+        // Simple check for who attacked first:
+        // if the player or his follower/summon attacked first, a crimeId should be set.
         // Doesn't handle possible edge case where no one reported the assault, but in such a case,
         // for bystanders it is not possible to tell who attacked first, anyway.
         if (victimStats.getCrimeId() != -1)
-            commitCrime(attacker, victim, MWBase::MechanicsManager::OT_Murder);
+            commitCrime(attackerIsPlayerFollowerOrSummon ? player : attacker, victim, MWBase::MechanicsManager::OT_Murder);
```
scrawl: This doesn't seem to be in line with the original game. Try this:
<pre>
npc1->aifollow, player, 0,0,0,0
npc1->startcombat, npc2
</pre>
No bounty is added.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1355,9 +1355,39 @@ namespace MWMechanics
...
+            // Check if attacker is one of the player followers
+            std::list<MWWorld::Ptr> followers = getActorsSidingWith(player);
+            if (std::find(followers.begin(), followers.end(), attacker) != followers.end())
+                attackerIsPlayerFollowerOrSummon = true;
+
+            // If not, check if attacker is one of the player summoned creatures
+            if (!attackerIsPlayerFollowerOrSummon) {
+                try {
+                    CreatureStats &playerStats = player.getClass().getCreatureStats(player);
+                    std::map<CreatureStats::SummonKey, int> summonedCreatures = playerStats.getSummonedCreatureMap();
```
scrawl: The map could be a const reference.
```
const std::map<CreatureStats::SummonKey, int>& summonedCreatures = playerStats.getSummonedCreatureMap();
...
typedef std::map<CreatureStats::SummonKey, int>::const_iterator it_type;
```

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1355,9 +1355,39 @@ namespace MWMechanics
...
+        MWWorld::Ptr player = getPlayer();
+
+        if (attacker != player) {
+            // Check if attacker is one of the player followers
+            std::list<MWWorld::Ptr> followers = getActorsSidingWith(player);
+            if (std::find(followers.begin(), followers.end(), attacker) != followers.end())
+                attackerIsPlayerFollowerOrSummon = true;
+
+            // If not, check if attacker is one of the player summoned creatures
+            if (!attackerIsPlayerFollowerOrSummon) {
```
scrawl: This whole if statement is actually redundant, because summoned creatures have their aifollow package added automatically.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1355,9 +1355,39 @@ namespace MWMechanics
...
+                    std::map<CreatureStats::SummonKey, int> summonedCreatures = playerStats.getSummonedCreatureMap();
+
+                    typedef std::map<CreatureStats::SummonKey, int>::iterator it_type;
+                    for (it_type it = summonedCreatures.begin(); it != summonedCreatures.end(); ++it) {
+                        if (attacker.getClass().getCreatureStats(attacker).matchesActorId(it->second)) {
+                            attackerIsPlayerFollowerOrSummon = true;
+                            break;
+                        }
+                    }
+                } catch (std::runtime_error e) {}
```
scrawl: Why are we catching exceptions here, what particular exception are you encountering? Exceptions should normally be used for error reporting only, not for "expected" paths.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
NeveHanter: Yes, You're right, my bad, only summons increases bounty.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
NeveHanter: I wanted to catch std::runtime_exception which will be thrown if getClass returns instance which class doesn't override getCreatureStats method. But if that's impossible to happen here it should be unwrapped of try/except.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
NeveHanter: If I remove follower code, it should stay here because followers shouldn't raise bounty.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
NeveHanter: I will change that one, thanks :+1:

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
scrawl: I'm not talking about a difference between summons/normal followers, but the fact that *any* follower attacking first doesn't produce a bounty for the player.

I suppose that this is a) intended to work this way (kinda makes sense that the player isn't responsible for the follower's actions, if they acted on their own accord) or b) no one really considered this case since followers usually don't attack a peaceful NPC on their own, except maybe for scripted events.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
scrawl: ```attacker``` is expected to be an actor which has CreatureStats. If that assumption is violated for some reason, we certainly don't want to ignore the exception, we want to log it and abort whatever caused it. But since we already have plenty of exception catching higher up the stack, there is no need to bother with it here. I suggest to remove the try-catch all together.

This is all just for future reference, since I've already suggested dropping the if statement entirely.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
NeveHanter: I've tested `startcombat, npc_id` on the summoned skeleton and high pockets reference and after it killed the NPC it got me 1000 bounty with vanilla. Also I've checked if after I attacked an NPC, and then summon or high pockets killed it, i also got 1000 bounty. The same after I've summoned summon after attacking the NPC, it got me 1000 bounty.

scrawl:
Just checked again, you are right there is a difference in the handling of summons and normal followers.

But, the current state of this PR no longer fixes the original bug report.

The expected behaviour with aifollow-ing actors is:

- Player attacks first, follower kills: murder bounty should be added (this is what the bug report was about).
- Follower attacks first and kills without the player getting involved: no murder bounty should be added.

NeveHanter:
Hello again @scrawl,

I want to complete this PR at last.

I've checked vanilla behaviour again with following actors:
- summon,
- aifollow'ed NPC,
- aiescort'ed NPC,
- recruited High-Pockets

in the following scenarios:

- by issuing startcombat command on different NPC in crowded room,
- by me attacking NPC

In every of the eight scenarios, I've got 1000 bounty for "siding with me" actor killing other NPC, but in startcombat scenario no-one of the other NPCs in the room attacked me afterwards getting the bounty (of course guards wanted to throw me in jail anyway)

And this conflicts with your last comment: 

> Follower attacks first and kills without the player getting involved: no murder bounty should be added.

Additionally I've checked what happens if aifollow'ing/aiescort'ing actor summon kills an NPC, and when I've attacked that NPC and "follower" summon killed him, I've got the bounty, but when I've used startcombat on my follower and his summon killed the NPC, I've got no bounty at all.

NeveHanter:
@scrawl I've probably made every needed corrections and prepared code for the review.

NeveHanter:
Right, my bad as I've had problem in my implementation before and thought it was because summoned creature package wasn't there, everything is ok after removal :+1:

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -374,9 +374,13 @@ namespace MWMechanics
             if (LOS)
             {
                 MWBase::Environment::get().getMechanicsManager()->startCombat(actor1, actor2);
-                if (!againstPlayer) // start combat between each other
+                if (!againstPlayer) // start combat between each other if engaging actor is not one of the player followers
```
scrawl: Can you explain why this change is necessary?

NeveHanter: This is done to prevent not-committing the crime when startcombat is issued i.e. when I issue "startcombat elone" on arrille, he summons the skeleton and elone starts fighting it without skeleton commiting the crime

NeveHanter: More specifically this is done because `actorAttacked()` checks for `!seq.isInCombat(attacker)` before calling `commitCrime()`.

Which will of course fail when startcombat is issued in here also for actor2.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1094,6 +1082,7 @@ namespace MWMechanics
         }
 
         // Make surrounding actors within alarm distance respond to the crime
+        // TODO: Vanilla doesn't make neighbors attack player if one of his follower/summon attacks first (probably vanilla bug)
```
Allofich: This seems to be because in vanilla followers only give you a crime for a kill, not for attacking. The same thing will happen if you kill an NPC instantly. You will only get the murder bounty and nearby NPCs won't attack you. I guess the code to have them respond is tied to the "on attack" code.

jeffreyhaines:
@NeveHanter are you working on this issue? https://bugs.openmw.org/issues/2852

If so please mark yourself as the assignee. We wouldn't want others doubling up on this task.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
NeveHanter: So now, there's question, should we follow vanilla or not regarding this?

NeveHanter:
@scrawl Any more comments on this or above comment thread about `engageCombat` changes?

Allofich:
I think maybe he's waiting to merge things until the new release is out. Is that right, scrawl? Or he's just busy with other things.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
Allofich: I don't think it matters that much, because normally your follower won't start combat with a non-aggressive actor anyway. They just respond to ones that are aggressive to you or them. So this would only matter if there is a script using "startcombat" on the follower, or if the player is using console commands. I suppose the best thing would be to follow vanilla, just in case there is a mod that uses startcombat to make the player follower attack and expects the vanilla behavior, but it seems like a very rare case to me.

Allofich:
By the way, I just noticed that in vanilla, the player is given crime bounties for murders that happen during fighting even from the opponents hitting each other with area-effect spells. You can test with Arrille because he will use some spells that can hit other NPCs. It even happen if you select a non-aggressive NPC A, give it "startcombat" on NPC B who is fighting you due to a crime, and B is killed. I guess NPCs responding to a crime are labeled crime victims, and the player gets a murder bounty no matter how such labeled NPCs die. Right now I don't know if we should follow that implementation.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
jeffreyhaines: Yeah that sounds like the right vanilla behavior. When getting boots of blinding speed, the woman I was helping would just zip off to kill enemies which was annoying to say the least...

Allofich:
> I guess NPCs responding to a crime are labeled to be possible murder victims, and the player gets a murder bounty no matter how such labeled NPCs die.

Any opinions? I suppose I support doing it the same as vanilla, which I think is a good policy for laying a base that could potentially have alternate implementations added on top of it but will get us closer to 1.0.  In this case, you could imagine that the law of Morrowind blames the initiator of an assault for all deaths that result from it, even if they didn't directly do it.

apps/openmw/mwmechanics/actors.cpp
scrawl: > start combat between each other if engaging actor is not one of the player followers

That's not how vanilla works, AFAIK. Can you think of a different solution to the problem?

scrawl: Please check for any conflicts/overlaps with the just merged PR #1174.

scrawl:
> Any opinions? I suppose I support doing it the same as vanilla, which I think is a good policy for laying a base that could potentially have alternate implementations added on top of it but will get us closer to 1.0. In this case, you could imagine that the law of Morrowind blames the initiator of an assault for all deaths that result from it, even if they didn't directly do it.

Makes sense to me. In some cases (i.e. multiple damage effects in the same frame) you couldn't really tell who made the kill anyway.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -420,6 +420,15 @@ namespace MWMechanics
 
             if (LOS)
```
akortunov: You should add "{" after this line or remove "}" from 431 line to compile OpenMW properly.

scrawl:
Current blocker is https://github.com/OpenMW/openmw/pull/921#pullrequestreview-13450952 , I think.

Capostrophic:
This needs some massive cleanup, the actual changes here that are necessary to resolve the issue are somewhat small.

akortunov:
This PR can be closed now.

psi29a:
Closing on advice from @akortunov

PR #924, opened 2016-04-07
Update script editor tab width (Bug #3286)

rcutmore:
This change sets the tab width for the script editor to four spaces based on the monospace font. I attempted to change the tab width whenever the monospace font user setting changed but subsequent calls to `setTabStopWidth()` seemed to have no effect. I'll be away for the next week or so but can play with it more when I'm back if we want this to update with the setting.

Issue on bug tracker: [OpenMW-CS Script editor tab width](http://bugs.openmw.org/issues/3286)

### DIFF
--- a/apps/opencs/model/prefs/state.cpp
+++ b/apps/opencs/model/prefs/state.cpp
@@ -136,6 +136,9 @@ void CSMPrefs::State::declare()
     declareBool ("wrap-lines", "Wrap Lines", false).
         setTooltip ("Wrap lines longer than width of script editor.");
     declareBool ("mono-font", "Use monospace font", true);
+    declareInt ("tab-width", "Tab Width", 4).
+        setTooltip ("Number of characters for tab width").
+        setRange (1, 10);
     EnumValue warningsNormal ("Normal", "Report warnings as warning");
     declareEnum ("warnings", "Warning Mode", warningsNormal).
         addValue ("Ignore", "Do not report warning").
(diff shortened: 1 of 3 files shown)


zinnschlag:
Having the tab width customisable would be useful.

rcutmore:
Agreed, I will play with this more next week when I get back to see if I'm
overlooking something. Another possible option would be to insert four
spaces when the tab key is pressed. Is it preferable to have an actual tab
character instead of replacing with spaces for the script editor?

MiroslavR:
I think it would be great to have both tab and space indentation and allow the user to choose between them as well as their length, like most IDEs offer.

zinnschlag:
Mixing tabs and spaces would be highly problematic here. It is highly problematic anywhere but especially here. When you load a list of content files you will have scripts from different authors in your editor. If some of these use spaces and some use tabs that will get very messy.  And it would make the non-monospace mode effectively unusable.

Sorry, but we need to stay tab only.

Svetomech:
So, the space people are out of the way?

rcutmore:
@zinnschlag Do you think the tab width setting should be restricted to a certain range? We can leave it unrestricted but if the user selects a tab width of 0 (which wouldn't really make sense to begin with) then Qt ends up using a tab width equivalent to 11 characters. Maybe something like 1 to 10 characters?

zinnschlag:
1 to 10 sound reasonable.

rcutmore:
There's now a user setting for controlling the number of characters to use for tab width. The tab width also uses the appropriate font.

I removed the unused *setMonoFont* function. When I initially looked at this I didn't realize it wasn't in use for some reason, which is what caused my confusion with the tab width not changing after updating the font.

PR #925, opened 2016-04-09
Adjust trade window to allow 0 gold bartering

bentsherman:
These edits to the TradeWindow class should allow the player to decrease "Total Cost/Sold" to 0. Basically, each offer is now represented by a magnitude and a boolean indicating buy or sell.

Also, whenever an item is transferred, the player's offer is now reset to the merchant's offer instead of offset, which is consistent with the original game.

This change is minor and probably not critical to the current roadmap, but for that reason I figure it's also a good way for me to test the waters and get into the right workflow with you guys.

I tested these changes in game and encountered no problems.

### DIFF
--- a/apps/openmw/mwgui/tradewindow.cpp
+++ b/apps/openmw/mwgui/tradewindow.cpp
@@ -57,8 +57,10 @@ namespace MWGui
         , mSortModel(NULL)
         , mTradeModel(NULL)
         , mItemToSell(-1)
-        , mCurrentBalance(0)
-        , mCurrentMerchantOffer(0)
+        , mIsPlayerSelling(false)
+        , mIsMerchantBuying(false)
+        , mPlayerOffer(0)
+        , mMerchantOffer(0)
     {
         getWidget(mFilterAll, "AllButton");
         getWidget(mFilterWeapon, "WeaponButton");
@@ -97,7 +99,6 @@ namespace MWGui
         mDecreaseButton->eventMouseButtonReleased += MyGUI::newDelegate(this, &TradeWindow::onBalanceButtonReleased);
 
         mTotalBalance->eventValueChanged += MyGUI::newDelegate(this, &TradeWindow::onBalanceValueChanged);
-        mTotalBalance->setMinValue(INT_MIN+1); // disallow INT_MIN since abs(INT_MIN) is undefined
 
         setCoord(400, 0, 400, 300);
     }
@@ -120,8 +121,10 @@ namespace MWGui
     {
         mPtr = actor;
 
-        mCurrentBalance = 0;
-        mCurrentMerchantOffer = 0;
+        mIsPlayerSelling = false;
+        mIsMerchantBuying = false;
+        mPlayerOffer = 0;
+        mMerchantOffer = 0;
 
         restock();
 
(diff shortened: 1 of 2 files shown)


scrawl:
Please keep unneeded changes to a minimum, especially when it comes to game mechanics code. Things like renaming variables and formatting code only introduce noise, make the changes harder to review and increases the potential for regressions. The variable names need to stay as they are referred to on the [Research](https://wiki.openmw.org/index.php?title=Research) page.

apps/openmw/mwgui/tradewindow.cpp
```diff
@@ -361,15 +355,15 @@ namespace MWGui
...
+                    + gmst.find("fBargainOfferBase")->getFloat()
+                    + std::abs(int(npcTerm - pcTerm));
 
+            // notify and decrease merchant disposition on haggle failure
+            // note: haggle auto-fails if player tries to buy items and receive gold (see above)
             int roll = Misc::Rng::rollDice(100) + 1;
-            if(roll > x || (mCurrentMerchantOffer < 0) != (mCurrentBalance < 0)) //trade refused
-            {
+
+            if ( x < roll || (mIsPlayerSelling && !mIsMerchantBuying) ) {
```
scrawl: Note that ```x < roll``` is not the same as ```roll > x```. The [research page](https://wiki.openmw.org/index.php?title=Research:Trading_and_Services#Bartering) says ```roll <= x```, so it looks like the original code was wrong.

apps/openmw/mwgui/tradewindow.hpp
```diff
@@ -65,8 +65,10 @@ namespace MWGui
 
             int mItemToSell;
 
-            int mCurrentBalance;
-            int mCurrentMerchantOffer;
+            bool mIsPlayerSelling;
+            bool mIsMerchantBuying;
```
scrawl: Can you elaborate why we need these flags? I think the old way of using the sign to indicate buying/selling works fine. And if the price is 0, then buying or selling makes no difference anyway.

apps/openmw/mwgui/tradewindow.cpp
bentsherman: The original code was correct. The formula on the research page says the trade is accepted if `roll <= x`, so this line runs the barter failure code if `x < roll` or `roll > x`, which are equivalent.

apps/openmw/mwgui/tradewindow.hpp
bentsherman: The thing is that with the current code you can't decrease the offer to 0, and if it happens to be an even trade of items then you can't increase the offer. It's a minor difference but the original game does let you haggle to 0.

scrawl: Yes, I know - it says so in the pull request description. I'm just wondering if there's a less invasive way of implementing this. For example, by only changing these two lines:

<pre>
diff --git a/apps/openmw/mwgui/tradewindow.cpp b/apps/openmw/mwgui/tradewindow.cpp
index c084ec2..791f8df 100644
--- a/apps/openmw/mwgui/tradewindow.cpp
+++ b/apps/openmw/mwgui/tradewindow.cpp
@@ -477,14 +477,14 @@ namespace MWGui
         if (mCurrentBalance == INT_MAX || mCurrentBalance == INT_MIN+1)
             return;
         if(mCurrentBalance<=-1) mCurrentBalance -= 1;
-        if(mCurrentBalance>=1) mCurrentBalance += 1;
+        else mCurrentBalance += 1;
         updateLabels();
     }
 
     void TradeWindow::onDecreaseButtonTriggered()
     {
-        if(mCurrentBalance<-1) mCurrentBalance += 1;
-        if(mCurrentBalance>1) mCurrentBalance -= 1;
+        if(mCurrentBalance<=-1) mCurrentBalance += 1;
+        else mCurrentBalance -= 1;
         updateLabels();
     }
 
</pre>

And FWIW, setting the offer to 0 was already possible by using the text entry box.

bentsherman:
I take your point unneeded changes, I tried to keep them to a minimum but I also wanted to have some style consistency in the neighborhood of my edits. Also did not know about the code fragments on the research page so I'll keep that in mind.

I was planning to handle the TODO about moving the mechanics code next, so I can make sure the variables are consistent with the research page when the mechanics code is moved?

apps/openmw/mwgui/tradewindow.cpp
scrawl: You are right. Not sure what I was thinking. The negation had me all confused.

apps/openmw/mwgui/tradewindow.hpp
bentsherman: With one other change, I would agree with you:

```
diff --git a/apps/openmw/mwgui/tradewindow.cpp b/apps/openmw/mwgui/tradewindow.cpp
index c084ec2..ed81677 100644
--- a/apps/openmw/mwgui/tradewindow.cpp
+++ b/apps/openmw/mwgui/tradewindow.cpp
         if (mCurrentBalance == INT_MAX || mCurrentBalance == INT_MIN+1)
             return;
-        if(mCurrentBalance<=-1) mCurrentBalance -= 1;
-        if(mCurrentBalance>=1) mCurrentBalance += 1;
+        if (mCurrentBalance < 0) mCurrentBalance -= 1;
+        else mCurrentBalance += 1;
         updateLabels();
     }
 
     void TradeWindow::onDecreaseButtonTriggered()
     {
-        if(mCurrentBalance<-1) mCurrentBalance += 1;
-        if(mCurrentBalance>1) mCurrentBalance -= 1;
+        if (mCurrentBalance < 0) mCurrentBalance += 1;
+        else mCurrentBalance -= 1;
         updateLabels();
     }
 
@@ -495,13 +495,13 @@ namespace MWGui
 
-        if (mCurrentBalance > 0)
+        if (mCurrentBalance < 0)
         {
-            mTotalBalanceLabel->setCaptionWithReplacing("#{sTotalSold}");
+            mTotalBalanceLabel->setCaptionWithReplacing("#{sTotalCost}");
         }
         else
         {
-            mTotalBalanceLabel->setCaptionWithReplacing("#{sTotalCost}");
+            mTotalBalanceLabel->setCaptionWithReplacing("#{sTotalSold}");
         }
```

Basically, set the UI to show "Total Sold" instead of "Total Cost" when the offer is 0, otherwise the increase button acts unexpectedly. You can make this change however which way, I did it this way to use `mCurrentBalance < 0` across the board.

scrawl: Makes sense. Can you open a new pull request with just these changes? The commit credit should go to you since you originally found the issue. :) Let us know how you want to be listed in AUTHORS.md (nickname, fullname or both)?

PR #927, opened 2016-04-11
When waiting, NPCs can go where they wouldn't go normally (Bug #2981)

Salgat:
Fixes https://bugs.openmw.org/issues/2981. When NPCs wander during a rest, this makes sure that they only wander along their current pathgrid and not a nearby one that may be on the other side of a wall (that would require a door to get to).

I had to reuse GetClosestPoint(...) (that was originally only in pathfinding.cpp), so I made the method public static to avoid code duplication.

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -39,33 +39,6 @@ namespace MWMechanics
     // distance must be long enough that NPC will need to move to get there.
     static const int MINIMUM_WANDER_DISTANCE = DESTINATION_TOLERANCE * 2;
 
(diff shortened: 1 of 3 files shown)


psi29a:
gave travis a kick, should be green now. :)

scrawl:
lgtm, thanks! :+1: 

Please create an account on the bugtracker so that we can assign the issue to you. If you've already created one, then I think Zini needs to give you developer permissions before your name will show up in the Assignee field.

Salgat:
@zinnschlag My username is Salgat http://bugs.openmw.org/users/784 if you can add me Zini. If this is the wrong place to message, please let me know. Thanks!

zinnschlag:
Done.

PR #928, opened 2016-04-13
AiWander - Creatures in cells without pathgrids do not wander (Bug #1320)

Salgat:
Fixes https://bugs.openmw.org/issues/1320. Normally creatures should wander around their spawn point when not in combat. In vanilla Morrowind, creatures can simply walk through the ground/walls and through other creatures while wandering, while OpenMW prevents this due to physics collisions. Extra logic was added to handle when a creature is trying to walk against a wall or another creature so they don't look like they are hopelessly walking in place. Additionally, the AiWander class only does navigation through the pathgrid, so I had to add an alternative "manual mode" so that the wander logic knows it's not trying to follow a pathgrid.

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -16,7 +16,6 @@
 #include "../mwworld/class.hpp"
 #include "../mwworld/esmstore.hpp"
 #include "../mwworld/cellstore.hpp"
-#include "../mwworld/customdata.hpp"
 
 #include "creaturestats.hpp"
 #include "steering.hpp"
@@ -253,14 +252,16 @@ namespace MWMechanics
             // Typically want to idle for a short time before the next wander
             if (Misc::Rng::rollDice(100) >= 96) {
                 wanderNearStart(actor, storage, mDistance);
+            } else {
+                storage.setState(Wander_IdleNow);
             }
         } else if (mAllowedNodes.empty() && !storage.mIsWanderingManually) {
             storage.mCanWanderAlongPathGrid = false;
         }
 
         // If Wandering manually and hit an obstacle, stop
-        if (storage.mIsWanderingManually && mObstacleCheck.check(actor, duration*10.0f, 2.5f)) {
-            stopWalking(actor, storage);
+        if (storage.mIsWanderingManually && mObstacleCheck.check(actor, duration, 2.0f)) {
+            completeManualWalking(actor, storage);
         }
 
         // Don't try to move if you are in a new cell (ie: positioncell command called) but still play idles.
@@ -293,6 +294,8 @@ namespace MWMechanics
                     setPathToAnAllowedNode(actor, storage, pos);
                 }
             } 
+        } else if (storage.mIsWanderingManually && storage.mPathFinder.checkPathCompleted(pos.pos[0], pos.pos[1], DESTINATION_TOLERANCE)) {
+            completeManualWalking(actor, storage);
         }
 
         return false; // AiWander package not yet completed
@@ -365,7 +368,7 @@ namespace MWMechanics
 
             // Check if land creature will walk onto water or if water creature will swim onto land
             if ((!isWaterCreature && !destinationIsAtWater(actor, destination)) ||
-                (isWaterCreature && destinationThroughGround(currentPositionVec3f, destination))) {
+                (isWaterCreature && !destinationThroughGround(currentPositionVec3f, destination))) {
                 storage.mPathFinder.buildSyncedPath(currentPosition, destinationPosition, actor.getCell(), true);
                 storage.mPathFinder.addPointToPath(destinationPosition);
                 storage.setState(Wander_Walking, true);
@@ -390,6 +393,12 @@ namespace MWMechanics
     bool AiWander::destinationThroughGround(const osg::Vec3f& startPoint, const osg::Vec3f& destination) {
         return MWBase::Environment::get().getWorld()->castRay(startPoint.x(), startPoint.y(), startPoint.z(),
                                                               destination.x(), destination.y(), destination.z());
+    }
+
+    void AiWander::completeManualWalking(const MWWorld::Ptr &actor, AiWanderStorage &storage) {
+        stopWalking(actor, storage);
+        mObstacleCheck.clear();
+        storage.setState(Wander_IdleNow);
     }
 
     void AiWander::doPerFrameActionsForState(const MWWorld::Ptr& actor, float duration, AiWanderStorage& storage, ESM::Position& pos)
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -67,6 +67,11 @@ namespace MWMechanics
 
         // AiWander states
         AiWander::WanderState mState;
+
+        // Wandering near spawn logic
+        bool mIsWanderingManually;
+        ESM::Pathgrid::Point mPreviousWanderingNearSpawnLocation;
+        int mStuckTimer;
         
         unsigned short mIdleAnimation;
         std::vector<unsigned short> mBadIdles; // Idle animations that when called cause errors
@@ -81,9 +86,16 @@ namespace MWMechanics
             mGreetingTimer(0),
             mCell(NULL),
             mState(AiWander::Wander_ChooseAction),
+            mIsWanderingManually(false),
+            mStuckTimer(0),
             mIdleAnimation(0),
             mBadIdles()
             {};
+
+        void setState(const AiWander::WanderState wanderState, const bool isManualWander = false) {
+            mState = wanderState;
+            mIsWanderingManually = isManualWander;
+        }
     };
     
     AiWander::AiWander(int distance, int duration, int timeOfDay, const std::vector<unsigned char>& idle, bool repeat):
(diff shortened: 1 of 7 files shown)


scrawl:
We already have some code for the stuck detection in obstaclecheck.cpp. Ideally this code should be generalised so that we can reuse it here. Not a blocker for the merge though.

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -227,8 +239,20 @@ namespace MWMechanics
         }
 
         // Actor becomes stationary - see above URL's for previous research
-        if(mAllowedNodes.empty())
+        // If not an NPC and no pathgrid is available, randomly idle or wander around near spawn point
+        if(mAllowedNodes.empty() && !actor.getBase()->mClass->isNpc() && !storage.mIsWanderingManually) {
```
scrawl: From a quick test in the original engine, it seems that NPCs are also affected by this behaviour.

(FWIW, there would have been a shorter way to write the statement: actor.getClass().isNpc() )

Salgat: @scrawl I added code to handle NPCs also. Unfortunately, I don't know of any test cases for NPCs that are scripted to wander around an area without pathgrids (I removed pathgrids from Seyda Neen but even in Vanilla all the units just stand around); do you know of a good test case for that?

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -303,8 +334,55 @@ namespace MWMechanics
...
+     * Creatures simply wander a certain distance from their starting location, while NPCs wander a scripted
+     * distance (mDistance) from the position where they started the wander package.
+     * http://www.uesp.net/wiki/Tes3Mod:AIWander
+     */
+    void AiWander::wanderNearStart(const MWWorld::Ptr &actor, AiWanderStorage &storage, int wanderDistance, bool isNpc) {
+        const ESM::Pathgrid::Point currentPosition = actor.getRefData().getPosition().pos;
+
+        // Determine a random location within radius of original position
+        const float pi = 3.14159265359f;
+        const float maxWanderDistance = isNpc ? wanderDistance : MINIMUM_WANDER_DISTANCE * 14.0f;
```
scrawl: Are you sure that this is correct? From my testing, creatures also respect the wander distance given in their AI package. For example, giving a very small distance: 
<pre>
aiwander, 20, 0, 0
</pre>
a scrib would become stationary.

Given a very high distance:
<pre>
aiwander, 10000, 0, 0
</pre>
it would keep walking in one direction until it hit something.

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -303,8 +334,55 @@ namespace MWMechanics
...
+        const float wanderRadius = Misc::Rng::rollClosedProbability() * maxWanderDistance;
+        const float randomDirection = Misc::Rng::rollClosedProbability() * 2.0f * pi;
+        const float destinationX = mInitialActorPosition.x() + wanderRadius * std::cos(randomDirection);
+        const float destinationY = mInitialActorPosition.y() + wanderRadius * std::sin(randomDirection);
```
scrawl: I don't know how sophisticated the original engine's code is in this respect, but I don't think the above code gives a uniform distribution on a circle as intended. A higher radius ought to be more likely than a small radius, since it covers more area. As usual, SO has the correct answer: http://stackoverflow.com/questions/5837572/generate-a-random-point-within-a-circle-uniformly

Aesylwinn: @scrawl Should the distribution be uniform? I think it kind of makes sense that the npc or creature would tend to stick around a central point (like a home). The only exception I can think of would be if it were patrolling an area.

scrawl: Good question. Well, it's not that important anyway.

apps/openmw/mwmechanics/aiwander.cpp
Salgat: Currently I'm trying to find out why mDistance is always 0 and where the AIWander Range is getting set (I'm assuming mDistance, but either I'm wrong or there is a bug I need to fix).

scrawl:
Just spotted a swimming mudcrab. We need to add checks for land creatures not to wander into the water, and water creatures not attempt to swim out of it.

Salgat:
Alright wandering is now functioning for all units (I was having some issues with some never wandering). Now need to focus on tweaking collision detection and preventing creatures from leaving/entering water.

Salgat:
@scrawl If you can, please give it another go; it should be working (including mudcrabs not going for a dip).

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -303,11 +336,62 @@ namespace MWMechanics
...
+
+            // Check if land creature will walk onto water or if water creature will swim onto land
+            if ((!isWaterCreature && !destinationIsAtWater(actor, destination)) ||
+                (isWaterCreature && destinationThroughGround(currentPositionVec3f, destination))) {
+                storage.mPathFinder.buildSyncedPath(currentPosition, destinationPosition, actor.getCell(), true);
+                storage.mPathFinder.addPointToPath(destinationPosition);
+                storage.setState(Wander_Walking, true);
+                return;
+            }
+        } while (attempts--);
```
scrawl: There is a possible underflow here which should be avoided, even if the variable is no longer tested after the loop.

Easy fix:
<pre>
} while (--attempts);
</pre>

Salgat: You're right, since who knows if code later on may be added to use that variable in some other way. Added a commit following your advice, thanks!

scrawl:
I've noticed that creatures sometimes stop playing their idle animations.

Go here:
<pre>
player->position, 55789, -89544, 1161 , 0
</pre>
and watch the netch creatures for a minute or so.

That issue doesn't seem to be present in master.

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -16,6 +16,7 @@
 #include "../mwworld/class.hpp"
 #include "../mwworld/esmstore.hpp"
 #include "../mwworld/cellstore.hpp"
+#include "../mwworld/customdata.hpp"
```
scrawl: I think this include can be removed.

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -206,29 +217,51 @@ namespace MWMechanics
...
+            // Typically want to idle for a short time before the next wander
+            if (Misc::Rng::rollDice(100) >= 96) {
+                wanderNearStart(actor, storage, mDistance);
+            }
+        } else if (mAllowedNodes.empty() && !storage.mIsWanderingManually) {
+            storage.mCanWanderAlongPathGrid = false;
+        }
+
+        // If Wandering manually and hit an obstacle, stop
+        if (storage.mIsWanderingManually && mObstacleCheck.check(actor, duration*10.0f, 2.5f)) {
```
scrawl: Why was it needed to change the distance parameter? What makes this invocation of the obstacle check any different from others?

I'm asking because we've had problems in the past with some creatures erroneously triggering the stuck detector. The tricky part is that movement velocity isn't constant, it's controlled by the animation and can vary greatly over the course of the animation cycle. We'll need to improve this code eventually. Maybe check back with the physics engine to see if an actual collision has occured. For now it seems like a bad idea to have some invocations use a different threshold than others (could result in more issues).

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -303,11 +336,62 @@ namespace MWMechanics
...
+            const float randomDirection = Misc::Rng::rollClosedProbability() * 2.0f * pi;
+            const float destinationX = mInitialActorPosition.x() + wanderRadius * std::cos(randomDirection);
+            const float destinationY = mInitialActorPosition.y() + wanderRadius * std::sin(randomDirection);
+            const float destinationZ = mInitialActorPosition.z();
+            destinationPosition = ESM::Pathgrid::Point(destinationX, destinationY, destinationZ);
+            destination = osg::Vec3f(destinationX, destinationY, destinationZ);
+
+            // Check if land creature will walk onto water or if water creature will swim onto land
+            if ((!isWaterCreature && !destinationIsAtWater(actor, destination)) ||
+                (isWaterCreature && destinationThroughGround(currentPositionVec3f, destination))) {
```
scrawl: Noticed that slaughterfish would only wander if next to an obstacle. This test needs be reversed. ```!destinationThroughGround```

apps/openmw/mwmechanics/aiwander.cpp
Salgat: The distance was increased to handle it slowly sliding against an obstacle just enough to move, but jittering pretty heavily in the process which looked really bad. Normally for AI_FOLLOW or others, this isn't really a problem since it still is technically making progress towards its destination.

Salgat:
@scrawl Thanks for catching those two issues, I believe this is good now.

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -303,11 +339,68 @@ namespace MWMechanics
...
+    }
+
+    /*
+     * Returns true if the position provided is above water.
+     */
+    bool AiWander::destinationIsAtWater(const MWWorld::Ptr &actor, const osg::Vec3f& destination) {
+        float heightToGroundOrWater = destination.z() - MWBase::Environment::get().getWorld()->getDistToNearestRayHit(destination, osg::Vec3f(0,0,-1), 1000.0, true);
+        osg::Vec3f positionBelowSurface = destination;
+        positionBelowSurface[2] = positionBelowSurface[2] - heightToGroundOrWater - DESTINATION_TOLERANCE;
+        return MWBase::Environment::get().getWorld()->isUnderwater(actor.getCell(), positionBelowSurface);
```
scrawl: I'm seeing mudcrabs in water again. This function doesn't seem right. Line 3 equals to
<pre>
positionBelowSurface[2] = destination.z() - destination.z() + getDistToNearestRayHit(...) - DESTINATION_TOLERANCE
</pre>

I think that it should be

<pre>
positionBelowSurface[2] = destination.z() - getDistToNearestRayHit(...) - DESTINATION_TOLERANCE
</pre>

Also I think that DESTINATION_TOLERANCE shouldn't be used here (or at least reduced), or we won't see creatures wandering along the shore.

Salgat: @scrawl l Great catch! I spent about 10 minutes watching different mudcrabs wandering along the coast and I don't see this issue anymore. You're right, only a small offset is needed to make sure it's underwater. Thanks so far for your help.

PR #929, opened 2016-04-19
Land record table

Pi03k:
https://bugs.openmw.org/issues/936
land_table_ft_936

### DIFF
--- a/components/esm/loadland.hpp
+++ b/components/esm/loadland.hpp
@@ -110,7 +110,7 @@ struct Land
         static void transposeTextureData(const uint16_t *in, uint16_t *out);
     };
 
-    void load(ESMReader &esm, bool &isDeleted);
+    virtual void load(ESMReader &esm, bool &isDeleted);
     void save(ESMWriter &esm, bool isDeleted = false) const;
 
     void blank() {}
--- a/components/esm/loadland.hpp
+++ b/components/esm/loadland.hpp
@@ -110,7 +110,7 @@ struct Land
         static void transposeTextureData(const uint16_t *in, uint16_t *out);
     };
 
-    virtual void load(ESMReader &esm, bool &isDeleted);
+    void load(ESMReader &esm, bool &isDeleted);
     void save(ESMWriter &esm, bool isDeleted = false) const;
 
     void blank() {}
--- a/apps/opencs/model/doc/saving.cpp
+++ b/apps/opencs/model/doc/saving.cpp
@@ -97,7 +97,8 @@ CSMDoc::Saving::Saving (Document& document, const boost::filesystem::path& proje
     appendStage (new WriteLandTextureCollectionStage (mDocument, mState));
 
     // references Land Textures
-    appendStage (new WriteLandCollectionStage (mDocument, mState));
+    appendStage(new WriteCollectionStage<CSMWorld::IdCollection<CSMWorld::Land> >(
+        mDocument.getData().getLand(), mState));
 
     // close file and clean up
     appendStage (new CloseSaveStage (mState));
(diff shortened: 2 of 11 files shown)


components/esm/loadland.hpp
```diff
@@ -110,7 +110,7 @@ struct Land
         static void transposeTextureData(const uint16_t *in, uint16_t *out);
     };
 
-    void load(ESMReader &esm, bool &isDeleted);
+    virtual void load(ESMReader &esm, bool &isDeleted);
```
scrawl: Why virtual?

Pi03k: apps/opencs/model/world/land.hpp line 18 has this exact method in derived class. Without virtual this method would be shadowed. Adding virtual will allow for dynamic dispatch / polymorphism / method overriding. Right?

scrawl: Yes, but OpenCS <i>knows</i> that it is dealing with the derived class, so the dynamic dispatch should not be necessary. And as for the game itself, there is no reason for the function to be virtual since it's not being overridden there.

Pi03k: ok. will remove that.

Pi03k: done

zinnschlag:
Why the changes to the saving code? Looks random.

Pi03k:
Small refactoring. I've removed CSMDoc::WriteLandCollectionStage and replaced in with instantiation of template WriteCollectionStage (one class less - code easier to understand/maintain). WriteLandCollectionStage received reference to document class where it really needs only land collection.

zinnschlag:
It seems the perform function for the Land records is functionally identical to the general perform function of the WriteCollectionStage template (except for the scope check which isn't an issue). Not sure how that happened. But if Land saving is refactored we can as well throw out the specialised function entirely.

Pi03k:
Should I remove specialised function and leave default template impl?

zinnschlag:
Looks like that is the way to go. I suggest to do a quick test run with it afterwards (load, save, test in OpenMW). The save code is crucial for OpenMW-CS and any modification to it (no matter how trivial) should be verified through a test.

Pi03k:
It looks that vanilla (master) doesn't save Land structure (with merged initial support by @cc9cii ). With my changes situation is the same. I need to debug and fix that. Save doesn't communicate user about any difficulties. Something weird is going on with saving. I'll describe it later.

zinnschlag:
Just tested with master. Loading and saving of land records works fine here.

Pi03k:
How did you test that if there is no land table? Could you please describe your testing procedure?

zinnschlag:
- Load Morrowind.esm into OpenMW-CS
- Save as Morrowind.omwgame
- Load Morrowind.omwgame into OpenMW-CS -> terrain renders correctly
- Load Morrowind.omwgame into OpenMW -> terrain renders correctly

Pi03k:
I've run game with modified omwgame and it didn't crash. Doest it mean that terrain renders correctly?

zinnschlag:
> Could you please elaborate more on 'renders correctly'?

Looks like normal terrain. I haven't checked if it matches exactly the original but I didn't notice any differences.

> Is step consisting of conversion from esm to omwgame necessary each time?

OpenMW-CS can not save in esm format.

Pi03k:
I've posted my testing of master openmw-cs on forum
https://forum.openmw.org/viewtopic.php?f=7&t=3453

scrawl:
For reference, the last comment in said thread:
> 
You added a land record with an invalid name. That needs to be handled in the LAND table creator bar. The only valid IDs are of the format #x y
Instead of entering an ID string it might be better to have two integer fields for input.

Anyway, the LAND code wasn't written for that. I guess you may have trashed the data structure. In fact the esm/omwgame file format can not store LAND records with such an ID. No idea what kind of junk ended up being written to the file.


So the PR is just missing ID validation? Anything else?

akortunov:
Feature 936 is implemented. I think we can close this PR now.

PR #932, opened 2016-04-26
Added all generated artifacts from the build process to the .gitignor…

alexanderkjall:
…e file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ Doxygen
 .project
 .settings
 .directory
+.idea
 ## qt-creator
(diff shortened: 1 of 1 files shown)


scrawl:
We recommend an out-of-source build (i.e. ```mkdir build ; cd build ; cmake .. ```) so that all the clutter stays in the build folder. In-source builds are still supported though, so might as well merge the PR.

alexanderkjall:
It even said so on the wiki, sorry for not reading stuff through before starting to send pull requests :)

PR #934, opened 2016-05-02
Editor: Prevent duplicate pathgrids (Bug #3342)

rcutmore:
This updates *CSMWorld::Data::hasId* to include pathgrids, preventing duplicates from being entered with the editor. Related issue on bug tracker: [Editor: You can create pathgrids with the same ID](http://bugs.openmw.org/issues/3342).

Unrelated to this bug, I noticed pathgrids aren't being included in [CSMWorld::Data::getIds](https://github.com/OpenMW/openmw/blob/24ae9f4ac1cdcd74e702ec8f09ccc45a8542b484/apps/opencs/model/world/data.cpp#L1162). Maybe that method should be updated as well? I haven't had a chance to look into it much yet but it's used in [CSMWorld::ScriptContext::isId](https://github.com/OpenMW/openmw/blob/24ae9f4ac1cdcd74e702ec8f09ccc45a8542b484/apps/opencs/model/world/scriptcontext.cpp#L94) and [CSMWorld::RefIdCollection::getIds](https://github.com/OpenMW/openmw/blob/24ae9f4ac1cdcd74e702ec8f09ccc45a8542b484/apps/opencs/model/world/refidcollection.cpp#L850)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -102,9 +102,9 @@ Programmers
     Pieter van der Kloet (pvdk)
     pkubik
     Radu-Marius Popovici (rpopovici)
-    rcutmore
     rdimesio
     riothamus
+    Rob Cutmore (rcutmore)
     Robert MacGregor (Ragora)
     Rohit Nirmal
     Roman Melnik (Kromgart)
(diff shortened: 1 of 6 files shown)


zinnschlag:
That is not going to work. Pathgrid IDs are not unique. They are shared with other tables (Cells and Land for example).

rcutmore:
Got it, my mistake. Would it work to create a PathgridCreator subclass of
GenericCreator to override the getErrors method? From there we could check
only existing pathgrid IDs for a duplicate.

zinnschlag:
Yes, that is the right approach.

rcutmore:
I updated this to include a separate PathgridCreator class for detecting duplicate IDs. When you get a chance let me know if I'm missing anything.

PR #936, opened 2016-05-05
fix crash on tegra 3

sandstranger:
this PR fix crash on tegra 3 . It is supports only 16 bit depth buffer


sandstranger:
this pr is bad . I just check it , and game freeze on start . I suppose , I must check device max depth buffer value

PR #937, opened 2016-05-06
Editor: Fix issue adding start script (Bug #3331)

rcutmore:
This fixes the issue adding start scripts mentioned here: [OpenMW-CS, Start Scripts table: Adding a script doesn't refresh the list of Start Scripts and allows to add a single script multiple times](http://bugs.openmw.org/issues/3331).

Previously records were added in two steps by adding a blank record and then setting the data in a later step. During that process there was a check to see if the column was editable before setting the data. This just adds the column check to the current process.

### DIFF
--- a/apps/opencs/model/world/idtable.cpp
+++ b/apps/opencs/model/world/idtable.cpp
@@ -159,7 +159,11 @@ void CSMWorld::IdTable::addRecordWithData (const std::string& id,
     mIdCollection->appendBlankRecord (id, type);
 
     for (std::map<int, QVariant>::const_iterator iter (data.begin()); iter!=data.end(); ++iter)
-        mIdCollection->setData (index, iter->first, iter->second);
+    {
+        int columnIndex = getModelIndex(id, iter->first).column();
+        if (mIdCollection->getColumn(columnIndex).isEditable())
+            mIdCollection->setData(index, iter->first, iter->second);
+    }
 
     endInsertRows();
 }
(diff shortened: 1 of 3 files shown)


zinnschlag:
I don't understand this change or why it would fix anything. Dropping data looks really wrong. If it can not be applied why is it there in the first place?

zinnschlag:
Okay, I see now. It avoids triggering an exception that then stops the rest of the function to execute properly. The problem is that this exception should never be triggered in the first place. Either the way the data is added is wrong or the column class is incorrect.

rcutmore:
So the issue appears to be caused by [CSVWorld::StartScriptCreator::configureCreateCommand](https://github.com/OpenMW/openmw/blob/master/apps/opencs/view/world/startscriptcreator.cpp#L27). I think this overridden method needs to be removed and just rely on the base implementation, which is blank. Do you agree? If so, I will update this PR with that change.

This is originally how the *StartScriptCreator* class was set up but I misunderstood the purpose of the *configureCreateCommand* method when making changes for #881.

Before [this commit](https://github.com/OpenMW/openmw/commit/7d8bd56e11e3eaee1cc6a7685ed61931ed4ea8bb), data for new records was added using [CSMWorld::IdTable::setData](https://github.com/OpenMW/openmw/blob/master/apps/opencs/model/world/idtable.cpp#L74), which also contains the check for whether or not a column is editable and masked this particular issue. Is the column check in *setData* problematic for the same reasons as well?

zinnschlag:
Yes, StartScriptCreator is the problem. The ID column is already set when creating the record. It can not be modified afterwards. Sorry for not catching that.

setData looks okay at least in principle, since the problem isn't just silently ignored here. The issue is that trying to change a non-editable column is an error in the code. And the change you made in this PR is dropping it in complete silence.
That the command was only executed partially after the exception was thrown is actually a bit of a problem in itself. We may need to improve the execption-safety of the command system at some point. But it is not important for now.

rcutmore:
That makes sense, thanks. I've pushed the required changes. Let me know if you'd like me to squash these commits before merging.

PR #941, opened 2016-05-11
Permanent barter disposition

worstperson:
This adds a setting to allow temporary disposition changes to apply to an npc's permanent disposition similar to the option in MCP. The goal is to help remove the aspect of bartering where the user will always choose an improbably good deal and offer until it's accepted for maximum profit and experience. This is not a fix-all and works best balancing mods(like mods that raise merchant speechcraft skill, make speechcraft harder to raise, set iBarterSuccessDisposition to 0, etc).

### DIFF
--- a/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
+++ b/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
@@ -21,6 +21,8 @@
 #include <components/interpreter/interpreter.hpp>
 #include <components/interpreter/defines.hpp>
 
+#include <components/settings/settings.hpp>
+
 #include "../mwbase/environment.hpp"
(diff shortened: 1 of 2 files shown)


scrawl:
@zinnschlag is this ok? Seems like this should rather belong in the game files (new GMST?) post 1.0, but for now a settings file option is the best we can do.

jeffreyhaines:
@worstperson do you have an active bug ticket for this issue? One needs to be created before we proceed to merge and/or close this pull request. Please post a link to that new ticket here, thanks

zinnschlag:
Post 1.0 change. Do we really need this now as a setting? 1.0 isn't that far way.

worstperson:
Did not mean to leave this open for so long, closing...

PR #940, opened 2016-05-09
Find ffmpeg

kpp:
`FindFFmpeg.cmake` set FFMPEG_* variables, which is incorrect acording to CMake docs. I fixed that. As a byproduct I fixed handling of components, e.g.
```
find_package(FFmpeg REQUIRED COMPONENTS AVUTIL) # only avutil will be searched + it fail if there is no avutil
```

Output from normal build:
```
-- Found FFmpeg_AVCODEC 56.60.100

-- Found FFmpeg_AVFORMAT 56.40.101

-- Found FFmpeg_AVUTIL 54.31.100

-- Found FFmpeg_SWSCALE 3.1.101

-- Found FFmpeg_SWRESAMPLE 1.2.101

-- Found FFmpeg: /tmp/openmw-deps/openmw-deps/lib/libavcodec.a;/tmp/openmw-deps/openmw-deps/lib/libavformat.a;/tmp/openmw-deps/openmw-deps/lib/libavutil.a;/tmp/openmw-deps/openmw-deps/lib/libswscale.a;/tmp/openmw-deps/openmw-deps/lib/libswresample.a  found components:  AVCODEC AVFORMAT AVUTIL SWSCALE SWRESAMPLE 
```

PR #942, opened 2016-05-11
Find freetype

kpp:
FindFreetype.cmake sets FREETYPE_* variables, which is incorrect according to CMake docs. I fixed that.
Tested for freetype versions 2.5.2, 2.4.8


scrawl:
I can't test this (nor can our CI) because OpenMW doesn't directly use Freetype. MyGUI does. The find freetype module is just there for use with static MyGUI builds (which I have never done before). Anyone using a static MyGUI build can verify that it works?

kpp:
Place unconditional `find_package(Freetype)` inside root CMakeLists.txt or FindMyGUI.cmake.

scrawl:
Looks like FindMyGUI.cmake still uses ```FREETYPE_LIBRARIES```, not ```Freetype_LIBRARIES```.

kpp:
Right. I wanted to create PR for MyGUI right after this one is accepted. I ll fix it!

kpp:
@scrawl done

scrawl:
The change of ```FREETYPE_HOME``` -> ```FREETYPE_SDK``` will break the build instructions on the wiki and thus everyone's workflow. Maybe we should add backwards compatibility for using ```FREETYPE_HOME```?

scrawl:
Forget that, freetype is not used in non-static builds, so it's not even in the build instructions... but what about the FFMPEG_HOME / FFMPEG_SDK from last PR?

kpp:
> but what about the FFMPEG_HOME / FFMPEG_SDK from last PR?

Frankly speaking your FindFREETYPE module was never tested in CI. Actually you set FFMPEG_SDK in CI and searched for FFMPEG_HOME. It failed to find libraries but due to explicit manually set `DAVCODEC_LIBRARIES="$FFMPEG_SDK/lib/avcodec.lib"...` builds passed.

https://github.com/OpenMW/openmw/pull/940/files#diff-5732a79d0837L408
https://github.com/OpenMW/openmw/pull/940/files#diff-9df58364beef9L103

> will break the build instructions on the wiki and thus windows people's workflow. Maybe we should add backwards compatibility for using FREETYPE_HOME?

I was confused with FFMPEG_HOME / FFMPEG_SDK and decided to unify env paths. Now I see I should have followed wiki instructions. The new PR is coming.

PR #944, opened 2016-05-12
Edit Pathgrid

Aesylwinn:
Here's the preliminary work for pathgrid editing. You can select multiple nodes and move them. Once I figure out how to add and remove rows to nested columns, the rest should be simple to implement.

### DIFF
--- a/apps/opencs/model/world/commands.cpp
+++ b/apps/opencs/model/world/commands.cpp
@@ -8,6 +8,7 @@
 #include <QAbstractItemModel>
 #include <QAbstractProxyModel>
 
+#include "cellcoordinates.hpp"
 #include "idcollection.hpp"
 #include "idtable.hpp"
 #include "idtree.hpp"
@@ -251,17 +252,11 @@ void CSMWorld::CreatePathgridCommand::redo()
     record.get().blank();
     record.get().mCell = mId;
 
-    if (!mId.empty() && mId[0]=='#')
+    std::pair<CellCoordinates, bool> coords = CellCoordinates::fromId(mId);
+    if (coords.second)
     {
-        int x, y;
-        char ignore;
-
-        std::istringstream stream (mId);
-        if (stream >> ignore >> x >> y)
-        {
-            record.get().mData.mX = x;
-            record.get().mData.mY = y;
-        }
+        record.get().mData.mX = coords.first.getX();
+        record.get().mData.mY = coords.first.getY();
     }
 
     mModel.setRecord(mId, record, mType);
--- a/apps/opencs/model/world/commands.cpp
+++ b/apps/opencs/model/world/commands.cpp
@@ -301,40 +301,6 @@ void CSMWorld::UpdateCellCommand::undo()
     mModel.setData (mIndex, mOld);
 }
 
-CSMWorld::ModifyNestedCommand::ModifyNestedCommand (IdTree& model, const std::string& id, int nestedRow,
-    int nestedColumn, int parentColumn, const QVariant& new_, QUndoCommand* parent)
-    : QUndoCommand(parent)
-    , NestedTableStoring(model, id, parentColumn)
-    , mModel(model)
-    , mId(id)
-    , mNestedRow(nestedRow)
-    , mNestedColumn(nestedColumn)
-    , mParentColumn(parentColumn)
-    , mNew(new_)
-{
-    std::string title = model.headerData(parentColumn, Qt::Horizontal, Qt::DisplayRole).toString().toUtf8().constData();
-    setText (("Modify " + title + " sub-table of " + mId).c_str());
-
-    QModelIndex parentIndex = mModel.getModelIndex(mId, mParentColumn);
-    mModifyParentCommand = new ModifyCommand(mModel, parentIndex, parentIndex.data(Qt::EditRole), this);
-}
-
-void CSMWorld::ModifyNestedCommand::redo()
-{
-    QModelIndex parentIndex = mModel.getModelIndex(mId, mParentColumn);
-    QModelIndex nestedIndex = mModel.index(mNestedRow, mNestedColumn, parentIndex);
-    mModel.setData(nestedIndex, mNew);
-    mModifyParentCommand->redo();
-}
-
-
-void CSMWorld::ModifyNestedCommand::undo()
-{
-    QModelIndex parentIndex = mModel.getModelIndex(mId, mParentColumn);
-    mModel.setNestedTable(parentIndex, getOld());
-    mModifyParentCommand->undo();
-}
-
 
 CSMWorld::DeleteNestedCommand::DeleteNestedCommand (IdTree& model,
                                                     const std::string& id,
--- a/apps/opencs/model/world/commands.cpp
+++ b/apps/opencs/model/world/commands.cpp
@@ -17,7 +17,13 @@
 
 CSMWorld::ModifyCommand::ModifyCommand (QAbstractItemModel& model, const QModelIndex& index,
                                         const QVariant& new_, QUndoCommand* parent)
-    : QUndoCommand (parent), mModel (&model), mIndex (index), mNew (new_), mHasRecordState(false), mOldRecordState(CSMWorld::RecordBase::State_BaseOnly)
+    : QUndoCommand (parent)
+    , mModel (&model)
+    , mIndex (index)
+    , mNew (new_)
+    , mHasRecordState(false)
+    , mOldRecordState(CSMWorld::RecordBase::State_BaseOnly)
+    , mModifyNestedCommand(0)
 {
     if (QAbstractProxyModel *proxy = dynamic_cast<QAbstractProxyModel *> (&model))
     {
@@ -28,40 +34,57 @@ CSMWorld::ModifyCommand::ModifyCommand (QAbstractItemModel& model, const QModelI
 
     if (mIndex.parent().isValid())
     {
-        setText ("Modify " + dynamic_cast<CSMWorld::IdTree*>(mModel)->nestedHeaderData (
-                    mIndex.parent().column(), mIndex.column(), Qt::Horizontal, Qt::DisplayRole).toString());
+        IdTree& tree = static_cast<CSMWorld::IdTree&>(*mModel);
+
+        mModifyNestedCommand = new ModifyNestedCommand(tree, mIndex, new_, this);
+        setText(mModifyNestedCommand->text());
     }
     else
     {
         setText ("Modify " + mModel->headerData (mIndex.column(), Qt::Horizontal, Qt::DisplayRole).toString());
-    }
 
-    // Remember record state before the modification
-    if (CSMWorld::IdTable *table = dynamic_cast<IdTable *>(mModel))
-    {
-        mHasRecordState = true;
-        int stateColumnIndex = table->findColumnIndex(Columns::ColumnId_Modification);
-
-        int rowIndex = mIndex.row();
-        if (mIndex.parent().isValid())
+        // Remember record state before the modification
+        if (CSMWorld::IdTable *table = dynamic_cast<IdTable *>(mModel))
         {
-            rowIndex = mIndex.parent().row();
-        }
+            mHasRecordState = true;
+            int stateColumnIndex = table->findColumnIndex(Columns::ColumnId_Modification);
 
-        mRecordStateIndex = table->index(rowIndex, stateColumnIndex);
-        mOldRecordState = static_cast<CSMWorld::RecordBase::State>(table->data(mRecordStateIndex).toInt());
+            int rowIndex = mIndex.row();
+            if (mIndex.parent().isValid())
+            {
+                rowIndex = mIndex.parent().row();
+            }
+
+            mRecordStateIndex = table->index(rowIndex, stateColumnIndex);
+            mOldRecordState = static_cast<CSMWorld::RecordBase::State>(table->data(mRecordStateIndex).toInt());
+        }
     }
 }
 
 void CSMWorld::ModifyCommand::redo()
 {
-    mOld = mModel->data (mIndex, Qt::EditRole);
-    mModel->setData (mIndex, mNew);
+    if (mModifyNestedCommand)
+    {
+        mModifyNestedCommand->redo();
+    }
+    else
+    {
+        mOld = mModel->data (mIndex, Qt::EditRole);
+        mModel->setData (mIndex, mNew);
+    }
 }
 
 void CSMWorld::ModifyCommand::undo()
 {
-    mModel->setData (mIndex, mOld);
+    if (mModifyNestedCommand)
+    {
+        mModifyNestedCommand->undo();
+    }
+    else
+    {
+        mModel->setData (mIndex, mOld);
+    }
+
     if (mHasRecordState)
     {
         mModel->setData(mRecordStateIndex, mOldRecordState);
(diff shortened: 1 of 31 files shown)


apps/opencs/view/render/worldspacewidget.hpp
```diff
@@ -35,6 +36,14 @@ namespace CSVRender
     class CellArrow;
     class EditMode;
 
+    struct WorldspaceHitResult
+    {
+        bool hit;
+        osg::ref_ptr<TagBase> tag;
+        unsigned int i0, i1, i2;
+        osg::Vec3d worldPos;
+    };
```
zinnschlag: Why do we need this struct? The tag based approach seems to work fine. Also, i0, i1 and i2 are a bit cryptic as variable names go.

Aesylwinn: The tag based system is good, but does not pass enough information. Since I implemented the rendering as a single object, I need triangle data (the poorly named i0,i1,i2 variables) to properly select nodes. I also added the world position since it will probably be convenient and has already been found. This will be needed to add a pathgrid node at a terrain position.
The alternatives I can think of are to pass that data in the tag itself (seems to not fit a tags purpose though) or slightly restructure the worldspace widget so that it passes the mouse coordinates to the edit mode which can then decide how to use it.

Aesylwinn:
After thinking about it, it makes sense to me to only pass the mouse position to an EditMode. It provides the most flexibility in my opinion. The EditMode could then use its pointer to the world space widget to call some generalized functions as needed. Any thoughts? 

Also, I wanted to mention that something like what I currently have or what I'm proposing is likely to be needed for the terrain editor.

Aesylwinn:
My proposed change is probably also needed to implement drag selection.

zinnschlag:
If you pass only the coordinates you will have to run the trace twice in most cases. That is wasteful. Let's stay with the version you have now.

Aesylwinn:
Sorry, I don't quite understand. Why would there be multiple traces?

apps/opencs/view/render/pagedworldspacewidget.cpp
```diff
@@ -142,14 +142,15 @@ void CSVRender::PagedWorldspaceWidget::addEditModeSelectorButtons (
         "terrain-move");
 }
 
-void CSVRender::PagedWorldspaceWidget::handleMouseClick (osg::ref_ptr<TagBase> tag, const std::string& button, bool shift)
+void CSVRender::PagedWorldspaceWidget::handleMouseClick (const WorldspaceHitResult& hit, const std::string& button,
```
Aesylwinn: I see what you mean. With what I wanted to do, this would have to call the mousePick function and may lead to the there being 2 traces. However, the result could be cached if performance is the concern.

Aesylwinn:
I'll just implement it like you said. It will probably be easier to tell what needs to be done afterwards.

Aesylwinn:
It's mostly functional. I believe only the addition of pathgrid records remains unimplemented. Potentially I could add a drag indicator that shows a line from a node to the mouse position, but that would require raytracing in the drag event. Let me know what you think.

zinnschlag:
A drag indicator would make sense. If you can find a way to implement it without making a mess of the whole input/worldspace framework, go for it.

Aesylwinn:
I just reread the forum post and realized I hadn't shared the selection functionality, I'll go ahead and do that too.

apps/opencs/view/render/pathgrid.cpp
```diff
@@ -0,0 +1,547 @@
...
+            mSelectedGeode->addDrawable(mSelectedGeometry);
+        }
+    }
+
+    void Pathgrid::removePathgridGeometry()
+    {
+        if (mPathgridGeometry)
+        {
+            mPathgridGeode->removeDrawable(mPathgridGeometry);
+            mPathgridGeometry.release();
```
scrawl: Don't use release() to free a ref_ptr. That unlinks the contained object without actually deleting it, resulting in a memory leak.

apps/opencs/view/render/pathgrid.cpp
```diff
@@ -0,0 +1,547 @@
...
+
+namespace CSVRender
+{
+    class PathgridNodeCallback : public osg::NodeCallback
+    {
+        public:
+
+            virtual void operator()(osg::Node* node, osg::NodeVisitor* nv)
+            {
+                PathgridTag* tag = dynamic_cast<PathgridTag*>(node->getUserData());
```
scrawl: It makes no sense to dynamic_cast and then not checking the result of the cast.

One of the below would be fine:

<pre>
PathgridTag* tag = static_cast<PathgridTag*>(node->getUserData());
tag->getPathgrid()->update();
</pre>

<pre>
PathgridTag& tag = dynamic_cast<PathgridTag&>(*node->getUserData());
tag.getPathgrid()->update();
</pre>

<pre>
PathgridTag* tag = dynamic_cast<PathgridTag*>(node->getUserData());
if (tag)
   tag->getPathgrid()->update();
else
   std::cerr << "Warning, unable to run PathgridNodeCallback" << std::endl;
</pre>

apps/opencs/view/render/pathgrid.cpp
```diff
@@ -0,0 +1,547 @@
...
+    {
+        return mPathgrid;
+    }
+
+    QString PathgridTag::getToolTip(bool hideBasics) const
+    {
+        QString text("Pathgrid: ");
+        text += mPathgrid->getId().c_str();
+
+        if (!hideBasics)
```
scrawl: Better remove this, I think that empty if statements might trigger warnings on some static analysis tools.

scrawl:
Slightly off-topic, just noticed a bug with the new camera system: hold down a movement key, then move the mouse out of the scene window, release movement key, the camera keeps moving still. Not sure if that was already an issue before this PR. It's a random issue though, I couldn't replicate it anymore on the 2nd try. :/

Aesylwinn:
Thanks for the input.

Aesylwinn:
There are several issues with input at the moment. At least for me (with qt5), the mouse drag event is only fired when one of the mouse buttons is pressed which means tooltips only show up if you drag while holding the mouse button down. I'm not sure how to fix that, I assume its an issue with something else stealing the input. 

Additionally, for your issue, if the widget loses focus before one of the key release signals occur, then the camera continues moving. The fix is to probably listen for some sort of qt focus event and reset the camera input.

scrawl:
> Additionally, for your issue, if the widget loses focus before one of the key release signals occur, then the camera continues moving.

Yeah, but it seems to work most of the time. The above only happened to me once. I am using Qt4.

Aesylwinn:
For me, I can always reproduce the issue by holding down one of the movement keys, then using the mouse to click on something in a different widget/window.

scrawl:
Right, that does it.

Aesylwinn:
As far as the context menu goes, I was following the pattern used in the instance editing mode (which I repeated for the orbit mode as well). I don't mind changing it assuming Zini's fine with that. One thing to note is that we have run out of regular mouse buttons.

Aesylwinn:
I guess a key binding wouldn't be too awkward in this case.

Aesylwinn:
I've run into an issue when it comes to creating a pathgrid. I am currently using the CreateCommand class to create the record. However, there is data that cannot be set using the CreateCommand since columns are used to access everything. I can't think of a  way to get around it without breaking the abstraction. I imagine this same issue will pop up when it becomes possible to add exterior pathgrids.

Aesylwinn:
Should columns be added for the pathgrid cell position? It might be troublesome if they are editable by the user.

Aesylwinn:
So anyway, the last commit is how I would change the EditMode class. If you don't like it, I'll just revert it.

zinnschlag:
Regarding the menu issue: We simply don't have enough mouse buttons to consistently provide a context menu in the 3D scene. This menu would also become very cluttered if we move all the button menus there. I think the current approach is the better choice.
Ultimately the primary purpose of the button menus is to let the user explore the available functions. I would expect that commonly used functions are invoked mostly via keyboard shortcuts (which we still have to add; once we have a system for configurable shortcuts).

scrawl:
If we stay with current context menu approach, how about these two enhancements:

- Add the context menu to the window's Menu bar to give it some more visibility
- Make the Menu key work while in the scene view (and make the key configurable, e.g. Spacebar should be convenient)

zinnschlag:
The main menu idea is problematic because it would require us to introduce a concept of an active subview (per window), since the main menu is per window and not per subview. I would like to avoid that.

And merging all the button menus (including the yet to be implemented ones) into a single context menu will just be confusing.

Aesylwinn:
Just an idea, but what about a context menu with sub menus?

Aesylwinn:
And spacebar opens it at the current mouse location.

Aesylwinn:
Also, I would prefer to implement any changes to that functionality in another pull request.

zinnschlag:
Sorry, but I do not think this is a good idea. The number of button menu items will be quite large when we are done. I suspect that the resulting context menu would end up very bulky and unpleasant to use (even with sub menus).
How about we finish up the user interface for the scene subview first and then (short time before the OpenMW-CS 1.0 release) take another look at the situation? This way we will now exactly what we are dealing with.

Aesylwinn:
I'm fine with that. Do you have any opinion on the pathgrid creation issue?

Aesylwinn:
Not sure if it's the best way to add pathgrids, let me know what you think. Anyway, I discovered that the code in the nested column adapter for adding/removing edges is broken. I'm going to take a break from this for a couple days, so I'll be fixing that later.

apps/opencs/model/world/commands.cpp
```diff
@@ -235,6 +237,35 @@ void CSMWorld::CloneCommand::undo()
...
+void CSMWorld::CreatePathgridCommand::redo()
+{
+    CreateCommand::redo();
+
+    Record<Pathgrid> record = static_cast<const Record<Pathgrid>& >(mModel.getRecord(mId));
+    record.get().blank();
+    record.get().mCell = mId;
+
+    if (!mId.empty() && mId[0]=='#')
+    {
```
zinnschlag: We have a class for all that cell ID parsing: CSMWorld::CellCoordinates.

zinnschlag:
That does not seem to work properly. I opened a random exterior cell. The camera angle happened to be low (looking more or less along the terrain). Then I clicked with the primary edit button on the ground. The nodes appeared free floating in the air.

Aesylwinn:
I don't suppose it was a cell with water? I've noticed a similar effect with those cells. Not sure if it's an invisible plane or something else.

apps/opencs/model/world/commands.cpp
Aesylwinn: Proof I need to take a break :)

zinnschlag:
No water. Just an ordinary land cell.

apps/opencs/model/world/commands.cpp
zinnschlag: Not really. I came up with this class relatively late. There are still plenty of examples of the manual ID decoding in the code. You probably took one of these a template for your code.

Aesylwinn:
Well, if you were clicking on empty space it would place a node in front of you. If there was any sort of geometry in front of you that is interactable, such as a tree which has some transparent geometry, it would place it there. Sorry, I'm having trouble understanding/reproducing this on my end. :/

apps/opencs/view/render/pathgridmode.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    void PathgridMode::primaryEditPressed(const WorldspaceHitResult& hitResult)
+    {
+        // Get pathgrid cell
+        Cell* cell = getWorldspaceWidget().getCell (hitResult.worldPos);
+        if (cell)
+        {
+            // Add node
+            QUndoStack& undoStack = getWorldspaceWidget().getDocument().getUndoStack();
+            QString description = "Connect node to selected nodes";
```
scrawl: Description is incorrect. Should be "Add node".

Aesylwinn:
Still need to fix the editor's table undo for nested data.

Aesylwinn:
What remains to be done:
1. Fix whatever bug Zini found (some more data like a specific cell might help, if you can find it)
2. In depth testing of save data
3. I need to do a final review of my code

Aesylwinn:
I think the bug Zini found may be related to cell borders.

apps/opencs/view/render/worldspacewidget.cpp
```diff
@@ -320,49 +318,72 @@ void CSVRender::WorldspaceWidget::addEditModeSelectorButtons (CSVWidget::SceneTo
...
 
-    for (osgUtil::LineSegmentIntersector::Intersections::iterator iter = intersector->getIntersections().begin();
-         iter!=intersector->getIntersections().end(); ++iter)
+    // Get relevant data
+    for (osgUtil::LineSegmentIntersector::Intersections::iterator it = intersector->getIntersections().begin();
+         it != intersector->getIntersections().end(); ++it)
     {
+        osgUtil::LineSegmentIntersector::Intersection intersection = *it;
+
         // reject back-facing polygons
```
Aesylwinn: Actually, it seems that occasionally (very rare for me), clicks on the terrain sometimes don't register. I suspect it is caused by the "reject of back-facing polygons" bit. Either that or osg isn't properly raytracing.

Aesylwinn:
Turns out it was caused by the rejection of back-facing triangles.
`Ignored normal is -0.19178 0.980852 -0.0338955`

I would guess that the rejection needs to take the perspective matrix into account.

Aesylwinn:
Well, I took a shot at refactoring the undo mechanism, but the required changes just keep on snowballing. I'm not interested in continuing to pursue that solution. One option I can see:

Move all of the extra steps (the reordering, point connection count) to the save code. Automatic edge removal would have to be removed from the nested column adapter implementation, or the undo mechanism would continue to be broken for pathgrids. There is the possibility that other classes will continue to be broken, but those were pre-existing and can be fixed at another time.

Your thoughts?

zinnschlag:
Agreed. That is consistent with what we are doing in other places.

scrawl:
Wouldn't that cause verifier errors to be thrown if you run verify on a non-saved file?

Aesylwinn:
Good thought, I'll check.

Aesylwinn:
Yes, the verifier will complain about the connection count being inaccurate. Do we need that check? Seems pointless if the user isn't going to be manually editing that value. If the error already exists, then the file is invalid and can't be loaded anyway.

zinnschlag:
Agreed.

psi29a:
Yay! \o/

zinnschlag:
The undo limit is per undo steps not per memory. Will give it a final test drive tomorrow.

zinnschlag:
There are a couple more issues:
* The node selection marker looks almost exactly as the edges (which kinda suggests to the user that the edges are always selected). This needs to be visually differentiated more
* The hint in tooltip about the context menu is misleading, since the edit mode button does not have a context menu. I don't think there is a point in mentioning the context menu anyway beyond what is already done in the tooltips of buttons that do have a context menu.
* Not supporting the "context sensitive selection" option feels really strange. It should be possible to support it if we only trigger it at the beginning of drags.

Aesylwinn:
I'm not sure how the context-select should work in this case, so you'll have to explain where it feels strange if my changes aren't sufficient.

Aesylwinn:
The new color scheme isn't really consistent with the instance edit mode.

zinnschlag:
I'll review your latest set of changes tomorrow. Sorry for the delay. Really busy month.

Aesylwinn:
That's alright :)

zinnschlag:
Looks good. One final point: When you drag the waypoint, the waypoint stays behind and only the selection marker moves. Not a problem in itself, but it is also inconsistent with the way object moving is done.

Aesylwinn:
I wanted to avoid reconstructing the pathgrid geometry every frame. However, I could test it if you want. Potentially, any impact could be reduced by only updating the geometry every couple frames.

Aesylwinn:
Of course, it depends on how many points/edges the pathgrid has. I would expect the algorithm used for the generation to be fairly linear in terms of performance.

zinnschlag:
Yes, please try it. Consistency between edit modes is actually pretty important.

Aesylwinn:
On my computer, there is approximately a 1-2 fps hit after using this with moderately sized pathgrids. I have an i7 though, so it might be worse for others.

zinnschlag:
Also a high end i7 here. Won't be able to provide additional data. I am not too worried, but maybe we should someone with a lesser CPU give it a try. Ask on the forum maybe?

Aesylwinn:
Well, I obtained that fps data using a hacked together frame counter in the CompositeViewer update function. If someone else is going to run it, I should probably get the osg statistics panel to display again since that was lost when I reworked the input system. :/ Any key preference? It was bound to 's', but that obviously conflicts with the camera system.

Aesylwinn:
I'll test this on my laptop. It has a mobile i5 processor (2 cores + hyperthreading), so its probably somewhere in the middle.

Aesylwinn:
Well, it was basically the same fps hit. I don't think this requires more testing.

scrawl:
> Well, I obtained that fps data using a hacked together frame counter in the CompositeViewer update function. If someone else is going to run it, I should probably get the osg statistics panel to display again since that was lost when I reworked the input system. :/ Any key preference? It was bound to 's', but that obviously conflicts with the camera system.

How about 'F3' for consistency with OpenMW?

Aesylwinn:
That would make sense :)

zinnschlag:
Great. Will give it another really final test tomorrow and then merge.

PR #947, opened 2016-05-28
AI package fixes

Allofich:
Fixes for bug 3349.
https://bugs.openmw.org/issues/3349

Summary of the changes made

- Noncombat AI packages will be cycled instead of only run through once. The only exception is AIWander if passed with a "no repeat" flag through script or console.
- AIWander and AIFollow packages will end when their specified duration is up, instead of continuing forever.

Reasoning / motivation behind the change

Recreates AI cycling behavior from the original game. 

What testing you have carried out to verify the change

Tested in various situations in game. Patrolling NPCs at Dren Plantation, guard who escorts player in opening, AI commands given through console, and testaipackage.esp (in linked bug report.)

### DIFF
--- a/apps/openmw/mwmechanics/aiescort.cpp
+++ b/apps/openmw/mwmechanics/aiescort.cpp
@@ -72,7 +72,11 @@ namespace MWMechanics
         {
             mRemainingDuration -= duration;
             if (duration <= 0)
+            {
+                // Reset mStarted to false so that package can be repeated again
+                mStarted = false;
                 return true;
+            }
         }
 
         if (!mCellId.empty() && mCellId != actor.getCell()->getCell()->getCellId().mWorldspace)
@@ -100,7 +104,11 @@ namespace MWMechanics
             point.mConnectionNum = 0;
             point.mUnknown = 0;
             if(pathTo(actor,point,duration)) //Returns true on path complete
+            {
+                // Reset mStarted to false so that package can be repeated again
+                mStarted = false;
                 return true;
+            }
             mMaxDist = 450;
         }
         else
--- a/apps/openmw/mwmechanics/aipackage.cpp
+++ b/apps/openmw/mwmechanics/aipackage.cpp
@@ -45,7 +45,12 @@ bool MWMechanics::AiPackage::shouldCancelPreviousAi() const
     return true;
 }
 
-MWMechanics::AiPackage::AiPackage() : mTimer(0.26f) { //mTimer starts at .26 to force initial pathbuild
+bool MWMechanics::AiPackage::getRepeat() const
+{
+    return true;
+}
+
+MWMechanics::AiPackage::AiPackage() : mTimer(0.26f), mStarted(false) { //mTimer starts at .26 to force initial pathbuild
 
 }
 
@@ -74,7 +79,13 @@ bool MWMechanics::AiPackage::pathTo(const MWWorld::Ptr& actor, ESM::Pathgrid::Po
     if(mTimer > 0.25)
     {
         const ESM::Cell *cell = actor.getCell()->getCell();
-        if (doesPathNeedRecalc(dest, cell)) { //Only rebuild path if it's moved
+        // If repeating an AI package (mStarted has been set to false again), build a new path if needed so package doesn't immediately end
+        if (!mStarted && distance(pos.pos, dest) > 10) {
+            mStarted = true;
+            mPathFinder.buildSyncedPath(pos.pos, dest, actor.getCell(), true); //Rebuild path, in case the target has moved
+            mPrevDest = dest;
+        }
+        else if (doesPathNeedRecalc(dest, cell)) { //Only rebuild path if it's moved
             mPathFinder.buildSyncedPath(pos.pos, dest, actor.getCell(), true); //Rebuild path, in case the target has moved
             mPrevDest = dest;
         }
@@ -94,7 +105,11 @@ bool MWMechanics::AiPackage::pathTo(const MWWorld::Ptr& actor, ESM::Pathgrid::Po
     /// Checks if you aren't moving; attempts to unstick you
     //************************
     if(mPathFinder.checkPathCompleted(pos.pos[0],pos.pos[1])) //Path finished?
+    {
+        // Reset mTimer so that path will be built right away when a package is repeated
+        mTimer = 0.26f;
         return true;
+    }
     else
     {
         evadeObstacles(actor, duration, pos);
--- a/apps/openmw/mwmechanics/aipackage.hpp
+++ b/apps/openmw/mwmechanics/aipackage.hpp
@@ -87,6 +87,10 @@ namespace MWMechanics
             /// Upon adding this Ai package, should the Ai Sequence attempt to cancel previous Ai packages (default true)?
             virtual bool shouldCancelPreviousAi() const;
 
+            /// Return true if this package should repeat. Can only be false for AIWander, if AIWander is assigned
+            /// assigned through the console or script.
+            virtual bool getRepeat() const;
+
             bool isTargetMagicallyHidden(const MWWorld::Ptr& target);
 
         protected:
@@ -103,6 +107,9 @@ namespace MWMechanics
             ObstacleCheck mObstacleCheck;
 
             float mTimer;
+
+            // Set to true once package starts actually being executed
+            bool mStarted;
 
             ESM::Pathgrid::Point mPrevDest;
 
--- a/apps/openmw/mwmechanics/aisequence.cpp
+++ b/apps/openmw/mwmechanics/aisequence.cpp
@@ -230,6 +230,11 @@ void AiSequence::execute (const MWWorld::Ptr& actor, CharacterController& charac
 
         if (package->execute (actor,characterController,state,duration))
         {
+            // Put repeating noncombat AI packages on the end of the stack so they can be used again
+            if (isActualAiPackage(packageTypeId) && package->getRepeat())
+            {
+                mPackages.push_back(package->clone());
+            }
             // To account for the rare case where AiPackage::execute() queued another AI package
             // (e.g. AiPursue executing a dialogue script that uses startCombat)
             std::list<MWMechanics::AiPackage*>::iterator toRemove =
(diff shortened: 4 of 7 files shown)


scrawl:
A few observations (please correct me if any of this does not match your testing).

* When an AI package is given through scripts or console, the previous AI package is removed. So the "cycling" implemented here can really only be tested with AI packages preset through the CS.
* AI packages with a "duration" parameter, when invoked via scripts or console, will repeat when that duration is over, effectively making the duration useless. This may be what prompted the "Reset start time of AIWander" logic in the previous version.

Lastly, I've noticed an issue with AiTravel if given through scripts/console: when the NPC arrives at the destination, the AI package does end (i.e. no longer repeat) in the original game (getCurrentAiPackage returns -1). In OpenMW the package is repeated every frame, so getCurrentAiPackage returns 1. The NPC will just stand about in either case, but the difference in return value of getCurrentAiPackage is a problem.

scrawl:
https://bugs.openmw.org/issues/3349#note-2

> The last argument (the repeat parameter) is an uninitialized char and always evaluates to true, so every NPC will have the repeat flag set.

In what way is it uninitialised? It's part of the ESM::AiWander struct, which gets read [here](https://github.com/OpenMW/openmw/blob/master/components/esm/aipackage.cpp#L21). Unless I'm mistaken, the size of this struct is 14 bytes, so it should get initialized correctly.

Allofich:
> AI packages with a "duration" parameter, when invoked via scripts or console, will repeat when that duration is over, effectively making the duration useless. This may be what prompted the "Reset start time of AIWander" logic in the previous version.

AIWander is working the same as the original game as far as I know. In both,

AIWander (area) (duration) (timeofday) will end (return -1) after duration ends.
AIWander (area) (duration) (timeofday) (idling values) will loop endlessly, ignoring duration
AIWander (area) (duration) (timeofday) (idling values) (repeats) will loop endlessly or end with duration.

I believe this is how it works both before and after my changes.

For AIFollow and AIEscort, when given through the console in the original game they will last forever and ignore duration, so the behavior is the same right now, but I guess it's probably better to respect the duration. That would be more intuitive and give more options for scripting.

You're right about AITravel returning -1 when it finishes in the original game. I didn't think much about testing a "one-shot" AI like that because I was thinking about AI packages repeating. That will need to be fixed I guess.

About the "unitialized char" comment, just ignore that. I made that comment before becoming more familiar with the code.

Allofich:
Made a few updates to the above post.

scrawl:
> About the "unitialized char" comment, I said that because when I used a cout statement to check what was in the variable, I was getting a garbage character. I ended up not doing anything about this because it didn't seem to be hurting anything.

When you cout a 'char', it will be interpreted as a character, not a number. The expected value for the repeat would be a number (0 or 1). 0 is not a valid ASCII character, so that's where the garbage is coming from. You just need to cast the char to an integer before printing it.

<pre>
std::cout << (int)data.mShouldRepeat << std::endl;
</pre>

I make this mistake myself all the time, it's a really annoying "feature" of the language for there not to be a distinction between bytes/small numbers and characters ;)

Allofich:
Ah OK, well, no need to worry about it I guess.

Allofich:
Just checked in original game, and a single Travel package on a character as placed in the editor will continue to return 1 after they arrive, while it will be -1 for an AITravel sent through the console. I'm sure you could probably fix this immediately, but I'll go ahead and see if I can fix it. Even this willl  probably take some time for me, though :).

Allofich:
BTW, if I submit a change to the pull request, will it have to run through that whole travis and appveyor build again?

scrawl:
It's not a trivial fix because of the interaction with save games - i.e. upon saving/loading we need to remember if a travel package was given through the CS or ingame.

I've noticed that the ESM structs for AiTravel, AiFollow, AiEscort and AiActivate have an 'int mUnk;' (unknown) variable in them. Maybe this is where the original engine would store the repeat flag in save games?

> BTW, if I submit a change to the pull request, will it have to run through that whole travis and appveyor build again?

Yes. Don't worry about wasting computation time, we're not paying for it :P

Allofich:
Yes, about save games and loading, that was one thing I was concerned about when I was working on this issue. I don't know how that's done (basically my knowledge currently is limited to what's going on in AIPackage, AISequence and AIWander, AIEscort, etc.)

I added a bool variable to check whether a package had started being executed yet and wondered that this might not be saved and could cause trouble with a loaded game. In fact while testing saving and loading I encountered an issue, which I admit I was sort of relieved was there before I had made any changes. I just reported it on the bugtracker.

Allofich:
Already fixed the bug on the tracker, I see. :)

scrawl:
> I added a bool variable to check whether a package had started being executed yet and wondered that this might not be saved and could cause trouble with a loaded game.

I think that should not be an issue - the whole pathfinding state is not stored in save games. When you load the game it's all reset and the fact that "mStarted" is also reset should not be a problem (for pathfinding purposes anyway).

Actually the one thing that might be a problem is this:
<pre>
        if (!mStarted)
        {
            // Set mStartTime once this package starts being executed
            mStartTime = MWBase::Environment::get().getWorld()->getTimeStamp();
            mStarted = true;
        }
</pre>
The start time will reset when you save/load the game.

scrawl:
> The start time will reset when you save/load the game.
 Show all checks

Looks like that was already an issue before this PR actually, because the ```AiWander (const ESM::AiSequence::AiWander* wander)``` calls ```init```.

Allofich:
About your fix to Bug #3409 (Only last AI package in stack is recalled on loading saved game).

When I just tried applying that to my local build with this PR applied to it, I still had a problem with Hrargal the Crow. I did coc to Dren Plantation, saved immediately and loaded, and he was standing still at his first AITravel destination (the one he was heading to when I saved). I haven't tested yet on a build without my changes applied, but does it work fine there?

scrawl:
Works for me on master. I saved when he started the first AI package, loaded and then watched him run through all the destinations.

Allofich:
OK, probably a problem with my changes, then.

scrawl:
It works for me also with master merged into your branch.

Not sure how you applied the change, but if you did manually, then perhaps you removed the "// remove previous packages if required" from AiSequence::stack() instead of AiSequence::fill()? It's best to use git pull / git merge in either case, to prevent such mix ups.

Allofich:
Yes, I did it manually. I still need to get accustomed to github, which I just started using a few hours ago. (I thought it would be easier to use than it is.) Anyway, I'm tired and unable to think well right now so I need to sleep.

Do you what you want with this pull request, I suppose. I'll continue to work on it later but I won't be offended if you reject it or change it or whatever you like. I especially need to get used to working with github, which will be a bit of an obstacle to overcome before I can do anything else I think.

Allofich:
Also the OpenMW "branching policy" says to use branches off the master, and I believe I sent this in directly against the master, so I might not have done the pull request as is liked. I was having trouble figuring out how to send it in to a branch off the master, though. I can create branches on my own fork, but can't seem to on the openmw master. Is creating it on my own fork and submitting from there the right way to do it?

Allofich:
Just checked and yes, I took out the code in the wrong spot, haha. Too tired and making careless mistakes. Anyway, thanks for your help and answering my questions, etc. I'll keep working on this and other issues when I can.

scrawl:
I'm ok with merging the code as-is. The minor issues we discussed can be fixed in another PR. However, we'll need a commit message better than "initial commit" ;) You can use ```git commit --amend``` and then ```git push --force``` to change the commit message.

> Also the OpenMW "branching policy" says to use branches off the master, and I believe I sent this in directly against the master, so I might not have done the pull request as is liked. I was having trouble figuring out how to send it in to a branch off the master, though. I can create branches on my own fork, but can't seem to on the openmw master. Is creating it on my own fork and submitting from there the right way to do it?

You create a branch on your local fork and then submit that branch against the openmw master branch.

Don't worry about the branching policy for now, that only matters if you're working on more than one feature at the same time.

As for this PR, it doesn't really matter if I pull it in from Allofich:master or Allofich:aipackagefixes .

Allofich:
OK thanks. How can I change the aisequence.cpp file to include your change? Sorry, I really don't know how to do anything on Github yet.

scrawl:
> OK thanks. How can I change the aisequence.cpp file to include your change? 

Technically, you don't <i>have</i> to do that, if I merge the PR now then it all gets merged together automatically.

Anyway, if you still want to:
Add the official repository to your list of remotes:
<pre>
git remote add upstream https://github.com/OpenMW/openmw.git
</pre>

Then to download new changes you do this (note the changes are not applied to your own fork yet)
<pre>
git fetch upstream
</pre>

To apply the changes, you can either merge or rebase. Merge adds the new changes on top of your changes. Rebase will rewrite your fork so that your changes are based on the new code (i.e. to an outsider, it would seem that you were working on top of the new source all along).  I usually prefer rebase because it is cleaner (no merge commits). If you decide to use merge then you can no longer change the commit message in the way I described above, since your changes will no longer be the latest commit. So if you want to use merge, then change the commit message first.

Merge:
<pre>
git merge upstream/master
</pre>

Rebase:
<pre>
git rebase upstream/master
</pre>

Finally, use ```git push``` to publish (or ```git push --force``` if you've rewritten anything that was previously pushed, e.g. by changing commit messages or rebase) .

Allofich:
Thanks for the instructions. Despite what I said before, I'm still up. I was stuck for a while in the text editor while editing the commit message, lol. Unfortunately I have almost no experience with Unix.

Anyway, after stumbling my way through it I think it's all OK now. But wait, is that "merge branch master" commit message going to be stuck on? If it does, can you remove it or do I have to do something about it? Sorry for all this, I've really got to just spend a day or so familiarizing myself with github.

Allofich:
Oh and I really am going to sleep now, so I won't be able to respond to anything for a while.

Allofich:
Finally figured out how to correctly edit my pull request I think. Please let me know if I still need to change anything.

Allofich:
Appveyor build seems to be stuck.

Allofich:
AppVeyor build failed. There was a message about it failing to download OpenAL at the top which might be related.

scrawl:
Appveyor is still new here and sometimes fails for valid PRs (build timeout, issues with dependencies, etc).

Allofich:
I've just updated my code locally so that AITravel will return correctly, but as you said, it doesn't work after loading a save (All AITravel packages will become non-repeating).

Allofich:
OK, seems to be working fine on save/load now.

PR #948, opened 2016-05-29
Fixing font scalability: Player stats window

gblues:
I intend to update the rest of the UI in a similar manner, but I wanted to get some feedback on the approach I used before getting too carried away. If the other windows can use the same methods I implemented here, I plan to refactor them into a utility class.

Note that this requires changes to the openmw_stats_window.layout file to function properly. Old versions did not provide names for certain widgets that needed to be touched.

I have tested this in the following ways:

* Using Pelagiad TTF with size=18
* Using Pelagiad TTF with size=36
* Using Pelagiad TTF with size=72
* using Vanilla openmw_font.xml and Vanilla fonts

For screenshots, see my thread on the forums: https://forum.openmw.org/viewtopic.php?f=6&t=3580

### DIFF
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -42,7 +42,7 @@ add_openmw_dir (mwgui
     itemmodel containeritemmodel inventoryitemmodel sortfilteritemmodel itemview
     tradeitemmodel companionitemmodel pickpocketitemmodel controllers savegamedialog
     recharge mode videowidget backgroundimage itemwidget screenfader debugwindow spellmodel spellview
-    draganddrop timeadvancer jailscreen
+    draganddrop timeadvancer jailscreen textsizeutil
     )
 
 add_openmw_dir (mwdialogue
--- a/apps/openmw/mwgui/review.cpp
+++ b/apps/openmw/mwgui/review.cpp
@@ -12,6 +12,7 @@
 #include "../mwworld/esmstore.hpp"
 
 #include "tooltips.hpp"
+#include "textsizeutil.hpp"
 
 namespace
 {
@@ -25,9 +26,6 @@ namespace
 
 namespace MWGui
 {
-
-    const int ReviewDialog::sLineHeight = 18;
-
     ReviewDialog::ReviewDialog()
         : WindowModal("openmw_chargen_review.layout")
     {
@@ -232,8 +230,8 @@ namespace MWGui
         groupWidget->setCaption(label);
         mSkillWidgets.push_back(groupWidget);
 
-        coord1.top += sLineHeight;
-        coord2.top += sLineHeight;
+        coord1.top += TextSizeUtil::getLineHeight();
+        coord2.top += TextSizeUtil::getLineHeight();
     }
 
     MyGUI::TextBox* ReviewDialog::addValueItem(const std::string& text, const std::string &value, const std::string& state, MyGUI::IntCoord &coord1, MyGUI::IntCoord &coord2)
@@ -253,8 +251,8 @@ namespace MWGui
         mSkillWidgets.push_back(skillNameWidget);
         mSkillWidgets.push_back(skillValueWidget);
 
-        coord1.top += sLineHeight;
-        coord2.top += sLineHeight;
+        coord1.top += TextSizeUtil::getLineHeight();
+        coord2.top += TextSizeUtil::getLineHeight();
 
         return skillValueWidget;
     }
@@ -269,8 +267,8 @@ namespace MWGui
 
         mSkillWidgets.push_back(skillNameWidget);
 
-        coord1.top += sLineHeight;
-        coord2.top += sLineHeight;
+        coord1.top += TextSizeUtil::getLineHeight();
+        coord2.top += TextSizeUtil::getLineHeight();
     }
 
     void ReviewDialog::addSkills(const SkillList &skills, const std::string &titleId, const std::string &titleDefault, MyGUI::IntCoord &coord1, MyGUI::IntCoord &coord2)
--- a/apps/openmw/mwgui/review.hpp
+++ b/apps/openmw/mwgui/review.hpp
@@ -77,8 +77,6 @@ namespace MWGui
         void addItem(const std::string& text, MyGUI::IntCoord &coord1, MyGUI::IntCoord &coord2);
         void updateSkillArea();
 
-        static const int sLineHeight;
-
         MyGUI::TextBox *mNameWidget, *mRaceWidget, *mClassWidget, *mBirthSignWidget;
         MyGUI::ScrollView* mSkillView;
 
--- a/apps/openmw/mwgui/spellbuyingwindow.cpp
+++ b/apps/openmw/mwgui/spellbuyingwindow.cpp
@@ -16,11 +16,10 @@
 
 #include "../mwmechanics/creaturestats.hpp"
 #include "../mwmechanics/actorutil.hpp"
+#include "textsizeutil.hpp"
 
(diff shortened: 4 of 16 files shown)


Allofich:
Hi. My pull request failed on the AppVeyor build, too, even though it passed yesterday. I noticed it failed to download OpenAL. I don't know if that's relevant or not.

scrawl:
IMO this is a low priority feature - yes, it's nice to have a configurable font size, but if you are increasing that for the purpose of legibility, then why shouldn't the size of icons, maps etc. also increase? To do that we already have the UI scaling option. There is currently a bug that the UI scaling doesn't take advantage of higher font resolutions, but we could fix that by decoupling a font's resolution from its default size.

As for the proposed changes, it would be a lot of work to update everything in the way that you did. I was hoping for a more generic solution. Widgets should be created in the correct size right away rather than resizing after the fact. One way to achieve this would be to allow skin/layout XML files to retrieve the font size. We can add a tag "#{fontsize}" that gets replaced on load by the default font size. Similar in the way that font colors are retrieved from the INI file:
```
            <State name="disabled" colour="#{fontcolour=disabled}" shift="0"/>
```

So, instead of
```
            <Widget type="MWSkill" skin="MW_StatNameButton" position="158 158 154 18" name="MagicSkill6" align="Left Top"/>
```

we would have
```
            <Widget type="MWSkill" skin="MW_StatNameButton" position="158 158 154 #{fontsize}" name="MagicSkill6" align="Left Top"/>
```

In some cases you may want to do some math e.g.
```
            <Widget type="MWSkill" skin="MW_StatNameButton" position="158 158 154 #{fontsize+2}" name="MagicSkill6" align="Left Top"/>
```

Then you would need to change ```WindowManager::onRetrieveTag``` to support the new fontsize tag.

That said, it is still a huge amount of work and once you start editing the lower level skins, there is no turning back because all the rest of the UI will be broken until you update it. I think there are many windows still using fixed size layouts that need to be updated to the auto-layouting system (i.e. HBox, VBox, AutoSizedTextBox). And speaking of that, the auto-layouting system has some unexpected quirks that make it a PITA to use. I plan to overhaul that at some point.

gblues:
I'm not sure that is true. My understanding was that the height returned by getDefaultHeight is calculated from font lyrics, and therefore different from one typeface to the next--hence why two faces at the same point size can have dramatically different line heights. 

The upshot is that keying the font size to an ini setting does not eliminate the need to do runtime widget sizing. 

I need to test that this is actually the case. I will do so this afternoon.

Aesylwinn:
I don't know if you noticed, but the Travis build failed because of the usage of tabs instead of spaces. I would imagine it can be fixed with an editor setting or some git trickery.

scrawl:
> I'm not sure that is true. My understanding was that the height returned by getDefaultHeight is calculated from font lyrics, and therefore different from one typeface to the next--hence why two faces at the same point size can have dramatically different line heights.

A vector typeface has no default size. You can render it any size you want. The getDefaultHeight is the value declared in the XML file as ```<Property key="Size" value="17"/>```. This value is also the resolution in texels used for each glyph when MyGUI converts the vector definition into a texture.

gblues:
> A vector typeface has no default size. You can render it any size you want. The getDefaultHeight is the value declared in the XML file as <Property key="Size" value="17"/>. This value is also the resolution in texels used for each glyph when MyGUI converts the vector definition into a texture.

This is not correct. Well, it's kind of correct. It's true you can render a vector typeface at whatever size you want, but the size specified in the XML does not match or even correlate with the result of `ResourceTrueTypeFont::getDefaultHeight()`. This is because `ResourceTrueTypeFont::getDefaultHeight()` tells you how much vertical space you need to be able to render any glyph in the typeface at the size specified in the Size property, which means it's dependent on the height of the highest ascender and the depth of the lowest descender. These are unique from one font to the next.

Here is the testing I did to validate this conclusion:

I have a size of 36 specified in my XML:

    <Resource type="ResourceTrueTypeFont" name="Magic Cards">
        <Property key="Source" value="Pelagiad.ttf"/>
        <Property key="Size" value="36" />

In the statswindow constructor, I added logging to print the return value of `ResourceTrueTypeFont::getDefaultHeight()`. For Pelagiad,  I get 25.

I grabbed a copy of Palatino Linotype (pala.ttf), and edited the XML to use it instead, leaving the Size property the same:

    <Resource type="ResourceTrueTypeFont" name="Magic Cards">
        <Property key="Source" value="pala.ttf"/>
        <Property key="Size" value="36" />

The stats window renders correctly in both cases. :)

Having the text size be pulled from the settings.cfg was a good idea, but it didn't work out-of-the-box. Tags (of `#{foo}` variety, not to be confused with XML tags) only call `WindowManager::onRetrieveTag()` when they're in taggable fields, like a Value field or something that contains user-facing text. On the plus side, I refactored `WindowManager::onRetrieveTag()` while I was in there looking at it.

gblues:
This weekend I tackled getting the window caption to look right. I tried to find other avenues, and came up empty.

My first approach was to try to modify the skin in-memory before the windows were actually instantiated. I couldn't get this to work--I could get the skin, and get the list of ChildSkinInfo objects from it, but the changes I made to the ChildSkinInfo objects didn't save:

* If I tested immediately after assignment, it worked.
* If I iterated through the exact same vector list, my changes were there
* However if I re-requested the ChildSkinInfo list from the skin, my changes disappeared

I feel like I'm missing something obvious. I'm guessing maybe because the MyGUI class is returning a const reference, what it actually returns is a copy of the objects? But wouldn't that give a compiler error?

Aesylwinn:
Well, just an idea, but maybe you are making/using a copy of the list?

Aesylwinn:
MyGUI itself is not returning a [copy](https://github.com/MyGUI/mygui/blob/master/MyGUIEngine/src/MyGUI_ResourceSkin.cpp#L290).

gblues:
Must have been, but not intentionally. I'm on mobile right now, but I'll paste a snippet this afternoon and hopefully my mistake will be obvious. 😄

gblues:
So here's a snippet. I put this in a do { } block immediately after the game loads core.xml in windowmanagerimp.cpp:

    int lineHeight = TextSizeUtil::getLineHeight(); // returns actual height of loaded font; use any value > 16
    MyGUI::SkinManager& mgr = MyGUI::SkinManager::getInstance();
    MyGUI::ResourceSkin *skin = mgr.getSkin("MW_Window_Pinnable");
    MyGUI::VectorChildSkinInfo children = skin->getChild();
    for(MyGUI::VectorChildSkinInfo::iterator it = children.begin(); it != children.end(); ++it)
    {
        switch((*it).coord.height - 20)
        {
        case 4:
        case 8:
        case 18:
        (*it).coord.height += (-20) + lineHeight + 4;
        }
    }
    children = skin->getChild(); // if you comment out this line, the loop below will show the changes made in the loop above
    for (MyGUI::VectorChildSkinInfo::iterator it = children.begin(); it != children.end(); ++it)
    {
        std::cout << (*it).name << " { left: " << (*it).coord.left << " top: " << (*it).coord.top << " w: " << (*it).coord.width << " h: " << (*it).coord.height << " }" << std::endl;
    }

Aesylwinn:
`MyGUI::VectorChildSkinInfo children = skin->getChild();`
This is where you are making the copy. Should be VectorChildSkinInfo**&**. Anyway, I think you'll run into another problem since skin->getChild() returns a **const** reference.

gblues:
I appreciate the help and feedback from everyone on this thread. I got what I needed out of it, and I'm going to try a different approach to more cleanly work around MyGUI's limitations.

PR #949, opened 2016-05-29
Fix return values for AITravel

Allofich:
Bug report
https://bugs.openmw.org/issues/3411

Summary
Changes Travel, Escort, Follow and Activate packages to not repeat if they are issued in-game (console or script.)

Reasons
To fix return values for in-game issued AI packages, matching original game behavior..

Testing
Tested editor-placed and console-given AI packages in-game.

### DIFF
--- a/apps/openmw/mwmechanics/aisequence.cpp
+++ b/apps/openmw/mwmechanics/aisequence.cpp
@@ -29,13 +29,14 @@ void AiSequence::copy (const AiSequence& sequence)
         mPackages.push_back ((*iter)->clone());
 }
 
-AiSequence::AiSequence() : mDone (false), mLastAiPackage(-1) {}
+AiSequence::AiSequence() : mDone (false), mRepeat(false), mLastAiPackage(-1) {}
 
 AiSequence::AiSequence (const AiSequence& sequence)
 {
     copy (sequence);
     mDone = sequence.mDone;
     mLastAiPackage = sequence.mLastAiPackage;
+    mRepeat = sequence.mRepeat;
 }
 
 AiSequence& AiSequence::operator= (const AiSequence& sequence)
@@ -231,7 +232,7 @@ void AiSequence::execute (const MWWorld::Ptr& actor, CharacterController& charac
         if (package->execute (actor,characterController,state,duration))
         {
             // Put repeating noncombat AI packages on the end of the stack so they can be used again
-            if (isActualAiPackage(packageTypeId) && package->getRepeat())
+            if (isActualAiPackage(packageTypeId) && (mRepeat || package->getRepeat()))
             {
                 mPackages.push_back(package->clone());
             }
@@ -292,6 +293,7 @@ void AiSequence::stack (const AiPackage& package, const MWWorld::Ptr& actor)
             else
                 ++it;
         }
+        mRepeat=false;
     }
 
     // insert new package in correct place depending on priority
@@ -317,6 +319,10 @@ AiPackage* MWMechanics::AiSequence::getActivePackage()
 
 void AiSequence::fill(const ESM::AIPackageList &list)
 {
+    // If there is more than one package in the list, enable repeating
+    if (!list.mList.empty() && list.mList.begin() != (list.mList.end()-1))
+        mRepeat = true;
+
     for (std::vector<ESM::AIPackage>::const_iterator it = list.mList.begin(); it != list.mList.end(); ++it)
     {
         MWMechanics::AiPackage* package;
@@ -366,6 +372,19 @@ void AiSequence::readState(const ESM::AiSequence::AiSequence &sequence)
     if (!sequence.mPackages.empty())
         clear();
 
+    // If there is more than one non-combat, non-pursue package in the list, enable repeating.
+    int count = 0;
+    for (std::vector<ESM::AiSequence::AiPackageContainer>::const_iterator it = sequence.mPackages.begin();
+         it != sequence.mPackages.end(); ++it)
+    {    
+        if (isActualAiPackage(it->mType))
+            count++;
+    }
+
+    if (count > 1)
+        mRepeat = true;
+
+    // Load packages
     for (std::vector<ESM::AiSequence::AiPackageContainer>::const_iterator it = sequence.mPackages.begin();
          it != sequence.mPackages.end(); ++it)
     {
--- a/apps/openmw/mwmechanics/aipackage.cpp
+++ b/apps/openmw/mwmechanics/aipackage.cpp
@@ -47,7 +47,7 @@ bool MWMechanics::AiPackage::shouldCancelPreviousAi() const
 
 bool MWMechanics::AiPackage::getRepeat() const
 {
-    return true;
+    return false;
 }
 
 MWMechanics::AiPackage::AiPackage() : mTimer(0.26f), mStarted(false) { //mTimer starts at .26 to force initial pathbuild
--- a/apps/openmw/mwmechanics/aipackage.hpp
+++ b/apps/openmw/mwmechanics/aipackage.hpp
@@ -87,8 +87,7 @@ namespace MWMechanics
             /// Upon adding this Ai package, should the Ai Sequence attempt to cancel previous Ai packages (default true)?
             virtual bool shouldCancelPreviousAi() const;
 
-            /// Return true if this package should repeat. Can only be false for AIWander, if AIWander is assigned
-            /// assigned through the console or script.
+            /// Return true if this package should repeat. Currently only used for Wander packages.
             virtual bool getRepeat() const;
 
             bool isTargetMagicallyHidden(const MWWorld::Ptr& target);
--- a/apps/openmw/mwmechanics/aisequence.hpp
+++ b/apps/openmw/mwmechanics/aisequence.hpp
@@ -40,6 +40,9 @@ namespace MWMechanics
             ///Finished with top AIPackage, set for one frame
             bool mDone;
 
+            ///Does this AI sequence repeat (repeating of Wander packages handled separately)
+            bool mRepeat;
+
             ///Copy AiSequence
             void copy (const AiSequence& sequence);
 


Allofich:
One note about this. I did the fix by having the script/console based command call a different constructor (with an added repeat parameter), while the editor will continue to use the constructor it did before. I don't know if stylistically this is OK or not.

scrawl:
This change as is would break the loading of older savegames.

I just did some digging with ```openmw-essimporter --compare``` to see how the original engine stores the repeat flag. To my surprise, there was no repeat flag. Instead, it appears that editor-placed AI packages are not stored in the save game whatsoever. Only once you modify the AI package list ingame, then it will appear in the save file.

ess_aipackage

I guess this could be an elegant way to do it, because:
* Issueing AI package ingame immediately removes all editor-placed packages, so there is no reason to store them.
* The save files are not bloated with unnecessary state.

If we adopt this solution, then as a nice side effect, ess-imported savegames should function correctly with respect to the AI packages.

On the other hand, it raises a question about some AI state that technically <i>does</i> need to be saved - e.g. the remaining duration of the package, the return position for AiWander. Does that mean the original engine has a bug causing that state to be reset upon saving/loading the game? Or is it stored somewhere else separate from the ai package? Have not tested it yet.

Allofich:
I just tested with an editor-placed wander package in the original game, and it seem that saving and loading does have such a bug.The one-hour duration of the package didn't end until one hour after the time of the save.

Allofich:
Looks like the return position isn't saved either. I led an NPC away from his wander zone in combat mode and stopped combat. He started heading back but I saved the game and loaded. After loading the game he was wandering in the new area that I had led him to.

Allofich:
Without adding any information to save files, since the only data that I know of that is available after loading is the x, y and z of the travel package, I can't think of any way to distinguish it from an editor placed package other than comparing it against the travel packages on that npc placed through the editor. If the x, y and z matched, you could assume the package from the save game was placed through the editor. Unfortunately, in the case that the NPC got a console or script AITravel to the exact same destination, it would be incorrectly assumed to be editor-placed. I just tested the original game, and this does not happen there. The package correctly terminates as a ingame-issued package. I'll keep thinking but right now I don't know a good solution.

Allofich:
I thought about this issue earlier today. I realized there is another problem with the current implementation (after my commit). Since I made all AI packages except wander to cycle, AIActivate when issued through the console or script will endlessly activate the target.

I also realized that it may not be desirable to replicate the original game's behavior of endlessly cycling a single placed AI package on an actor, except for Wander. It is not so obvious that a Target or Follow or Escort is actually still running when the actor has reached their destination and is just standing still there. It is also inconsistent to have the package continue to run for an editor placed package but not for a console/script placed package. Only people who are familiar with the quirks of the engine will understand the different behavior.

I want to change the implementation so that Travel, Escort, Follow and Activate will only repeat if they are not the only AI package in the stack. Otherwise they will only run once. From my understanding, this should cleanly handle all cases. AIWander will retain its current implementation (repeats in all cases unless issued in a certain way through the console/script.) Actually, the AIWander implementation is also rather strange and non-obvious (aiwander 512 1 0 will not repeat but aiwander 512 1 0 0 will), but it matches the original game and I don't want to touch it right now.

What do you think? I will go ahead and try to implement as I said unless you want me to stop. Also, if you don't want to wait for me (it may take me some time) or have your own implementation you want to do, I won't mind of course.

Allofich:
New implementation uploaded. I also changed the pull request comment at the top of the page.

scrawl:
Interesting idea. Seems like a good compromise for the time being.

Some minor issues:

I don't understand what the "repeatCheck" flag is supposed to be good for - isn't it redundant?

The "mRepeat" variable should be initialized in the constructor.

Instead of special-casing AiWander based on type id, you could override the ```virtual getRepeat() const``` function in AiWander.

Btw, I just noticed another bug: the vanilla engine treats aifollow/aiescort durations in game hours, but in OpenMW the duration is assumed to be realtime seconds.

Allofich:
I'm rewriting it, but first, just to answer what I meant with way of writing it:

> I don't understand what the "repeatCheck" flag is supposed to be good for - isn't it redundant?

The idea was to save performance by setting this flag after checking a package, so it wouldn't be checked over and over every time the "execute" function ran. But I think it's probably unnecessarily bulking up the code, so I'm rewriting without it.

> The "mRepeat" variable should be initialized in the constructor.

Will do so in the new version.

> Instead of special-casing AiWander based on type id, you could override the virtual getRepeat() const function in AiWander.

Hopefully neither will be necessary, if I can rewrite the way I intend to.

Allofich:
New version uploaded. It should be a cleaner implementation, I think. Note, though, that this is just to fix the return flags. It does not make single AI editor-placed packages only run once as I was saying above. That means that this fix does not address the problem of AI behavior after loading a saved game. Currently, it makes all AI packages except wander non-repeating after loading a game.

First I want to be sure that this implementation is OK. Then, I will do a separate PR to make single editor-placed packages run once. I want to be more organized with my work, and next time I will try to use a topic branch instead of my master branch.

One question -- Will the commit message from the older version "Change AI stacks of only one package to not repeat" still end up showing on the OpenMW page? Do you want me to squash it into the latest commit? (I'm still a little shaky on how to do that.)

Allofich:
So frustrating. I can't seem to merge everything into one commit message even when following directions online. Will this be a problem?

Allofich:
OK, it should be all right now. Unfortunately I've restarted the AppVeyor build several times. Luckily no one else seems to be using it right now, but I'll try to do things more smoothly from now on.

apps/openmw/mwmechanics/aiescort.cpp
```diff
@@ -23,10 +23,11 @@
 
 namespace MWMechanics
 {
-    AiEscort::AiEscort(const std::string &actorId, int duration, float x, float y, float z)
+    AiEscort::AiEscort(const std::string &actorId, int duration, float x, float y, float z, bool repeat)
     : mActorId(actorId), mX(x), mY(y), mZ(z), mRemainingDuration(static_cast<float>(duration))
     , mCellX(std::numeric_limits<int>::max())
     , mCellY(std::numeric_limits<int>::max())
+    , mRepeat(repeat)
```
Aesylwinn: Instead of adding an 'mRepeat' variable to each subclass, you should probably put it in the AiPackage class and initialize it using the AiPackage constructor. Then you wouldn't need to duplicate the getRepeat function in each subclass.

Allofich: Thank you for your comment. That was the way I was originally doing it, but in order for the calls to the various AI constructors from aiextensions.cpp (for running an AI command from the console or script) to set the variable to false, it seemed to me this way the only way to do it. Am I overlooking something?

Allofich: I think I understand what you mean now. I'll take a look at fixing it that way.

Aesylwinn: You are setting it to false using the subclass contructors, right? You can pass that value to the base class constructor.

apps/openmw/mwmechanics/aifollow.cpp
```diff
@@ -27,9 +27,9 @@ struct AiFollowStorage : AiTemporaryBase
 
 int AiFollow::mFollowIndexCounter = 0;
 
-AiFollow::AiFollow(const std::string &actorId,float duration, float x, float y, float z)
+AiFollow::AiFollow(const std::string &actorId,float duration, float x, float y, float z, bool repeat)
 : mAlwaysFollow(false), mCommanded(false), mRemainingDuration(duration), mX(x), mY(y), mZ(z)
-, mActorRefId(actorId), mActorId(-1), mCellId(""), mActive(false), mFollowIndex(mFollowIndexCounter++)
+, mActorRefId(actorId), mActorId(-1), mCellId(""), mActive(false), mFollowIndex(mFollowIndexCounter++), mRepeat(repeat)
 {
```
Aesylwinn: For example, this constructor would become:
`AiFollow::AiFollow(const std::string &actorId,float duration, float x, float y, float z, bool repeat)
: AiPackage(repeat), mAlwaysFollow(false), mCommanded(false), mRemainingDuration(duration), mX(x), mY(y), mZ(z)
, mActorRefId(actorId), mActorId(-1), mCellId(""), mActive(false), mFollowIndex(mFollowIndexCounter++)`
You would then assign mRepeat in the AiPackage constructor

Allofich: Thanks, I didn't know you could do that. I'm already in the middle of rewriting it a slightly different way, but if it's problematic I'll try the way you suggested next.

Allofich: And regarding whether it's "problematic" or not, I'll of course yield to the judgment of you or scrawl or any of the other veterans.

Allofich: It all seems to be working fine, and with less clutter in the code now. Thanks.

Allofich:
Updated again, thanks to advice from Aesylwinn, with slimmer code, more generalized code. This still doesn't address repeat behavior after a load, which is currently unpredictable. If the implementation so far looks OK, I can work on that next, either on this PR or on a new one.

scrawl:
Looks generally ok, but I don't think we should merge this before the saving issue is addressed.

Allofich:
Sure, I'm working on it.

scrawl:
Just an idea. We don't need to save the repeat state for each package, do we? Either all packages in the sequence repeat, or none do. The only exception would be AiWander, but that already has the saving part implemented. You could add something like ```bool AiSequence::mRepeat``` and not have to change much else.

Allofich:
Probably can do that. Taking a look now...

Allofich:
I _think_ I can do it that way. Working on the implementation now...

Allofich:
Implemented and seems to be working fine. Thank you for your insight scrawl.

Allofich:
Looks I spoke too soon. Needs a bit more work.

Allofich:
Looks to all be working now. Will upload.

Allofich:
Cleaned up a couple comments.

Allofich:
Corrected an initialization order warning from the Travis build.

apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -231,7 +232,7 @@ void AiSequence::execute (const MWWorld::Ptr& actor, CharacterController& charac
         if (package->execute (actor,characterController,state,duration))
         {
             // Put repeating noncombat AI packages on the end of the stack so they can be used again
-            if (isActualAiPackage(packageTypeId) && package->getRepeat())
+            if (isActualAiPackage(packageTypeId) && (mRepeat || ((packageTypeId == 0) && package->getRepeat())))
```
scrawl: ```(packageTypeId == 0)``` seems redundant.

apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -317,6 +319,9 @@ AiPackage* MWMechanics::AiSequence::getActivePackage()
 
 void AiSequence::fill(const ESM::AIPackageList &list)
 {
+    // If there is more than one package in the list, enable repeating
+    if (list.mList.begin() != (list.mList.end() - 1))
```
scrawl: ```end()-1``` is undefined behaviour if the list is empty.

apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -366,6 +371,25 @@ void AiSequence::readState(const ESM::AiSequence::AiSequence &sequence)
     if (!sequence.mPackages.empty())
         clear();
 
+    // If there is more than one non-combat, non-pursue package in the list, enable repeating. Before the check, iterate through combat and pursue.
+    for (std::vector<ESM::AiSequence::AiPackageContainer>::const_iterator it = sequence.mPackages.begin();
+         it != sequence.mPackages.end(); ++it)
+        switch (it->mType)
```
scrawl: use ```isActualAiPackage()```

Allofich:
BTW, this latest implementation works correctly after loading a saved game as well.

apps/openmw/mwmechanics/aisequence.cpp
Allofich: Yes, you're correct... will fix it.

apps/openmw/mwmechanics/aisequence.cpp
Allofich: This was needed because aipackage.cpp was setting it to true by default. Fixing this.

Allofich:
Re-uploaded again.

apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -366,6 +372,18 @@ void AiSequence::readState(const ESM::AiSequence::AiSequence &sequence)
     if (!sequence.mPackages.empty())
         clear();
 
+    // If there is more than one non-combat, non-pursue package in the list, enable repeating.
+    for (std::vector<ESM::AiSequence::AiPackageContainer>::const_iterator it = sequence.mPackages.begin();
+         it != sequence.mPackages.end(); ++it)
```
scrawl: This doesn't quite do what the comment says. Imagine a sequence of 1. Follow, 2. Combat. The code would set mRepeat to true, but it shouldn't.

Allofich: I thought that combat would always come first?

Allofich: Would adding up a count every time isActualAiPackage returns true and then checking if it is greater than 1 be an acceptable solution?

scrawl: > I thought that combat would always come first?

Hm, true, so the above scenario never happens. But a random person reading just this particular function with no context might think that there is a bug.

> Would adding up a count every time isActualAiPackage returns true and then checking if it is greater than 1 be an acceptable solution?

Sounds better.

Allofich: OK, will do so. Thanks again for helping. I'm sure you could do this so much faster yourself.

Allofich: Looks like something's gone amiss after the last few changes. After loading a saved game AI travels aren't always repeating. Probably related to changing the AIPackage's getrepeat to default to off.

Allofich:
Re-uploaded yet again.

apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -366,6 +372,20 @@ void AiSequence::readState(const ESM::AiSequence::AiSequence &sequence)
     if (!sequence.mPackages.empty())
         clear();
 
+    // If there is more than one non-combat, non-pursue package in the list, enable repeating.
+    int count = 0;
+    for (std::vector<ESM::AiSequence::AiPackageContainer>::const_iterator it = sequence.mPackages.begin();
+         it != sequence.mPackages.end(); ++it)
+    {    
+        if (isActualAiPackage(it->mType) && (it != (sequence.mPackages.end())))
```
scrawl: ```(it != (sequence.mPackages.end())``` can be removed.

Allofich: Fixed.

Allofich:
Thanks for all your help, scrawl. Sorry to take your time.

scrawl:
Don't worry about that, your contributions are much appreciated. Sure, maybe I could have done the coding portion faster, but things like research/testing/coming up with ideas are usually a lot more work than the coding anyway ;)

Allofich:
Thanks. I felt bad that I might be taking you away from other things to correct my mistakes or not allowing you to work on the AI as you like because of this pull request. I think I am learning a lot, though. I'll also submit bugs, don't worry :). I actually have a few I could submit, but it seems like there are already plenty of things to work on as it is, so I've been keeping them aside.

PR #950, opened 2016-05-29
Authors: merge kpp and humbug

kpp:
kpp and humbug is the same person =)


scrawl:
Are you and your past you really the same person still?

Just kidding... oops :P

Svetomech:
The ship of Theseus problem, huh? Didn't know you were into philosophy, @scrawl.

PR #952, opened 2016-05-30
Landing audio candidate fix

elInfidel:
Bug #2256

Slightly modified the starting point of the landing animation preventing the user from cutting it off with a consecutive jump. After this change I'm no longer able to time consecutive jumps which cut off the audio.

To my knowledge there's a small gap of time between the landing animation starting and the landing audio actually being played. By setting the starting point of the animation further in I was able to eliminate the window with which the player could jump again after landing and cut off the audio.

I'm new to the project and unsure if this would be considered a fix. Admittedly it seems like putting a band-aid on the issue instead of fixing the root cause but I'm assuming the animation data is drawn from the original game files and we don't want to tamper with that instead.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -356,7 +356,7 @@ void CharacterController::refreshCurrentAnims(CharacterState idle, CharacterStat
             mCurrentJump.clear();
             if (mAnimation->hasAnimation("jump"))
                 mAnimation->play(jumpAnimName, Priority_Jump, jumpmask, true,
-                             1.0f, "loop stop", "stop", 0.0f, 0);
+                             1.0f, "loop stop", "stop", 0.3f, 0);
         }
     }
 


scrawl:
I don't think this is a good solution because a part of the landing animation will be skipped, so it won't look as intended. The sudden jump from one frame to another would cause a discontinuity in the player animation. Also, we should try to avoid arbitrary values (like the 0.3 seconds here), that value may not be appropriate for modified animation files.

I just checked how the original engine does it. Jumping while the landing animation is still playing does not interrupt the landing animation. I.e. the landing animation keeps playing while the player is on upward trajectory, then the jump animation starts once the landing animation is finished. It looks slightly strange but gets the job done in working around the sound problem.

Allofich:
I noticed a small difference in the timing of the landing sound between OpenMW and the original engine that might have some relation to this issue.

To reproduce:
1. Go into third person and move the camera so that you have a good view of your character's feet.
2. Go to somewhere with a flat surface. I used a bridge in Seyda Neen. Do a forward jump, and keep moving forward so that your character immediately changes to the walking animation upon landing.

When I do this in the original engine, the landing sound seems to play simultaneously with the landing, but in OpenMW there is a small but noticeable delay, with the sound playing after the walking animation has already started. If it's hard to notice, try opening the console when landing. You should be able to time it to open after the walking animation has started, but before the landing sound has played.

PR #953, opened 2016-06-04
Move mechanics code from TradeWindow to mwmechanics

bentsherman:
This commit takes care of a TODO that I saw in mwgui/tradewindow.cpp. It shouldn't change any game behavior.


MiroslavR:
What's the point of instantiating an empty class? The haggle function may as well be static.

bentsherman:
Well I noticed a lot of these mechanics classes have state (for example, the `Repair` class stores the repair tool), so it might make sense in the future to add some members to this `Trading` class such as the merchant reference. There are also other pieces of mechanics code peppered throughout `TradeWindow` that could be moved into `Trading`, so I'm betting on this class having state in the long run.

zinnschlag:
Sounds reasonable to me. @scrawl Okay with that?

PR #954, opened 2016-06-07
Improve duration calculations in Wander AI

Allofich:
Fixes: https://bugs.openmw.org/issues/3412

Summary of the changes made:
Wander AI now runs on a counter, matching observed behavior in the original game more closely. The checks for the end of duration are now unrelated to the current timestamp.

The remaining duration on the counter is saved and retrieved on loading, which fixes bug#3412. This is an enhancement over the original game.

This implementation introduces a new float mRemainingDuration and repurposes mStartTime as a way to store the remaining duration.

Existing save games will have a wrong duration for one run through all the Wander packages in the save file (since mStartTime did not mean what it does now) but the duration should be corrected when the packages finish or for any that were not yet in the save game.

Tested with NPCs, both editor-placed packages and console-issued AIWander commands.

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -97,8 +97,8 @@ namespace MWMechanics
     };
     
     AiWander::AiWander(int distance, int duration, int timeOfDay, const std::vector<unsigned char>& idle, bool repeat):
-        mDistance(distance), mDuration(duration), mTimeOfDay(timeOfDay), mIdle(idle), mRepeat(repeat)
-      , mStoredInitialActorPosition(false)
+        mDistance(distance), mDuration(duration), mRemainingDuration(duration), mTimeOfDay(timeOfDay), mIdle(idle),
+        mRepeat(repeat), mStoredInitialActorPosition(false)
     {
         mIdle.resize(8, 0);
         init();
@@ -122,6 +122,11 @@ namespace MWMechanics
             mDuration = 0;
         if(mDuration == 0)
             mTimeOfDay = 0;
+        if(mDuration > 24)
+        {
+            mDuration = 24;
+            mRemainingDuration = (mDuration);
+        }
 
         mPopulateAvailableNodes = true;
 
@@ -200,12 +205,7 @@ namespace MWMechanics
             mPopulateAvailableNodes = true;
         }
 
-        if (!mStarted)
-        {
-            // Set mStartTime once this package starts being executed
-            mStartTime = MWBase::Environment::get().getWorld()->getTimeStamp();
-            mStarted = true;
-        }
+        mRemainingDuration -= ((duration*MWBase::Environment::get().getWorld()->getTimeScaleFactor()) / 3600);
 
         cStats.setDrawState(DrawState_Nothing);
         cStats.setMovementFlag(CreatureStats::Flag_Run, false);
@@ -235,8 +235,9 @@ namespace MWMechanics
 
         if (isPackageCompleted(actor, storage))
         {
-            // Reset mStarted so that package will get a new mStarttime when it repeats
-            mStarted = false;
+            // Reset package so it can be used again
+            mRemainingDuration=mDuration;
+            init();
             return true;
         }
 
@@ -317,10 +318,8 @@ namespace MWMechanics
     {
         if (mDuration)
         {
-            // End package if duration is complete or mid-night hits:
-            MWWorld::TimeStamp currentTime = MWBase::Environment::get().getWorld()->getTimeStamp();
-            if ((currentTime.getHour() >= mStartTime.getHour() + mDuration) ||
-                (int(currentTime.getHour()) == 0 && currentTime.getDay() != mStartTime.getDay()))
+            // End package if duration is complete
+            if (mRemainingDuration <= 0)
             {
                     stopWalking(actor, storage);
                     return true;
@@ -784,6 +783,8 @@ namespace MWMechanics
 
     void AiWander::fastForward(const MWWorld::Ptr& actor, AiState &state)
     {
+        // Update duration counter
+        mRemainingDuration--;
         if (mDistance == 0)
             return;
 
@@ -923,11 +924,17 @@ namespace MWMechanics
 
     void AiWander::writeState(ESM::AiSequence::AiSequence &sequence) const
     {
+        float remainingDuration;
+        if (mRemainingDuration > 0 && mRemainingDuration < 24)
+            remainingDuration = mRemainingDuration;
+        else
+            remainingDuration = mDuration;
+
         std::auto_ptr<ESM::AiSequence::AiWander> wander(new ESM::AiSequence::AiWander());
         wander->mData.mDistance = mDistance;
         wander->mData.mDuration = mDuration;
         wander->mData.mTimeOfDay = mTimeOfDay;
-        wander->mStartTime = mStartTime.toEsm();
+        wander->mDurationData.mRemainingDuration = remainingDuration;
         assert (mIdle.size() == 8);
         for (int i=0; i<8; ++i)
             wander->mData.mIdle[i] = mIdle[i];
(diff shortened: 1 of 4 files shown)


scrawl:
Why do we need to store member variables for duration, start time, and end time? Can we not derive the 3rd one from the other two?

Since we're on topic of wander durations, do you know what mTimeOfDay is supposed to do? It's currently unused, despite being passed in as parameter to aiwander script instructions.

Allofich:
Yes, I think you're right we don't need the end time variable. I got caught up in making everything work, and didn't stop to think whether that variable was truly necessary.

 I believe TimeofDay is supposed to be the hour that the package will start. So you could specify a package to start at 3 p.m. or something.

Allofich:
Oh, I remember now. One reason I was storing the end time and end day was so there wouldn't be the same calculations done every time isPackageCompleted is run.

Honestly I don't know what kind of performance impact if any that has, or whether it is preferable to saving memory by not storing the values as member variables.

scrawl:
This is a trivial calculation that is run at most a few dozen times per second, so in this case I prefer simplicity over speed. In general, when an optimization makes the code more complex, then chances are it is doing more harm than good. If what you are working on is not a performance bottleneck (or you don't know if it is), then just don't optimize.

Allofich:
OK, sure. I'm rewriting it without the variables. Seem to have run into a small bump with it erroneously ending a package, but I'll upload again when it tests correctly.

Allofich:
Ah, I see what's happening now. I was testing giving an NPC a 48-hour AIWander through the console and wondering why it was finishing early on only the first run, but it's because with a package that long the initial mStartTime of 0 is considered within range and "valid" and gets used.

Allofich:
I'm not sure what to do about that. I could just say that a start time of 0 is not valid and get a new time. Normally this would be the case but I kind of wanted this to work even with a "time travel" mod or something :)

Allofich:
Unless a clever idea comes up I'm just going to make a starttime of 0 be invalid, since it would be extremely rare to have this even if there was a mod that had the player go back to crossing from day -1 to 0, if that's even possible.

scrawl:
We don't have a way to denote an invalid/uninitialized time stamp. I suggest to set the start time from the constructor so it's always valid.

Allofich:
I uploaded a new implementation before noticing your latest message.
Unless I'm overlooking something, the problem with setting from the constructor is that when the wander package is not the first in the stack, it's start time needs to not be set until it is actually executed, or it will end up not running its full duration.

Allofich:
The variable names in the declaration of isPackageTimeValid are reversed. It should be
            bool isPackageTimeValid(float endHour, int endDay);
not
            bool isPackageTimeValid(float endDay, int endHour);

I'm going to wait on uploading the fix for now though, so any other fixes can be made. I don't want to choke up the auto-builds with a ton of submissions again.

Allofich:
I'm thinking about how I can set the timestamp from the constructor, btw.

Allofich:
All I can think of, is to always get a new timestamp on running execute(), and to introduce a new boolean specifically to set when loading a game, that will cause the timestamp to not be reset then. In that case, I don't think it's necessary to set in the constructor, though.

Allofich:
Actually I think I thought of a way to do it without adding a new boolean.

scrawl:
> Unless I'm overlooking something, the problem with setting from the constructor is that when the wander package is not the first in the stack, it's start time needs to not be set until it is actually executed, or it will end up not running its full duration.

If you want it to work like that, the most logical way to achieve it would be to store a remaining duration that's decremented on each frame. Unfortunately, we're kind of stuck with the savegame format that we already have.

Does it actually work that way in the original engine though? What happens when you give wander package with duration, startCombat, advance time, then end combat? Does that count as duration being used up or not? And what happens if you give a wander package to an NPC that's not in an active cell, advance time, then go to that cell? Have we tested any of that?

Allofich:
> What happens when you give wander package with duration, startCombat, advance time, then end combat? Does that count as duration being used up or not?

Just testing it now, and looks like it doesn't count. The combat was automatically ended after I've advanced time and returned (using coc) but the NPC doesn't properly get out of its combat stance.

> What happens if you give a wander package to an NPC that's not in an active cell, advance time, then go to that cell?

The package didn't advance through its duration until I used coc into the cell with the NPC.

If we want to emulate the original engine exactly, looks running down a counter that is only active while the cell is loaded would be the way to go.

Allofich:
I did upload a new version of the implementation I was talking about here to
https://github.com/Allofich/openmw/tree/AIWanderDuration2
but maybe you'd rather focus instead on getting things to work like the original game?

Allofich:
I'm also going to test advancing time after walking to an adjacent cell, not using coc.

scrawl:
> If we want to emulate the original engine exactly, looks like running down a counter that is only active while the cell is loaded would be the way to go.

Yeah, I think a remainingDuration counter would be best.

The only problem with that is the compatibility with existing save files. How about converting the counter to a startTime on saving, then converting back to remainingDuration on load?

Allofich:
I just tested walking between an interior and an exterior, and the NPC I gave an aiwander to in the interior did not run down their counter while I was advancing time outside.

> The only problem with that is the compatibility with existing save files. How about converting the counter to a startTime on saving, then converting back to remainingDuration on load?

StartTime is a timestamp, but remainingDuration would be a float or something, right? How would we convert between that? (it might be obvious to you...)

scrawl:
> StartTime is a timestamp, but remainingDuration would be a float or something, right? How would we convert between that? (it might be obvious to you...)

By that I mean a start time specifically chosen so that startTime + duration = currentTime + remainingDuration (if time was proceeding normally, that is). On loading we can then deduce the remainingDuration as startTime + duration - currentTime. This should be a lossless conversion because the currentTime does not change between saving and loading.

Allofich:
OK, well, if I'm going to do it, it will have to be tomorrow or later. I suppose maybe I was getting a bit carried away and not thinking about emulating the original game. This isn't Oblivion's radiant AI after all. ;)

Allofich:
Oh, do you want the behavior of ending wander packages between midnight and 1 a.m. to stay? Personally that seems unnecessary to me...

scrawl:
> Oh, do you want the behavior of ending wander packages between midnight and 1 a.m. to stay? Personally that seems unnecessary to me...

Remove it if you want, fine with me.

Allofich:
New implementation is up at
https://github.com/Allofich/openmw/commit/7414d35649bbd11326adac14467b3820e2b7cdf8.

I updated the description at the top of this pull request with information about the implementation.

Allofich:
Some observations about original game behavior:
1. Console-given AIWander commands will not just end between 12:00 a.m. and 1 p.m., but will end at any point after 12:00 a.m. is crossed (2 a.m., 3 a.m., etc.). The recreation of this in OpenMW was not quite correct. My speculation is that this might have been put in as a way to ensure that no package could run over 24 hours (which is the limit for a Wander package placed in the Construction Set). Construction Set-placed Wander packages, on the other hand, will end at exactly 12:00 a.m. if the actor is in the active cell. If the player leaves the cell (makes it non-active) and comes back after 12:00 a.m. after resting, the package seems to become buggy and never end at all.

I chose not to implement any special midnight-based behavior, but I did limit packages to 24 hours.

2. The wander duration "counter" does proceed during combat. This does not happen in OpenMW, with or without my changes. I guess the increment of the duration counter would need to be changed to a function that is called from aisequence or somewhere that is managing the AI stack, instead of running in execute as it does now.

3. Packages do not return "completed" if the duration is passed through fast-forwarding. In OpenMW, they do return as completed. I don't know which is more desirable.

Allofich:
> On loading we can then deduce the remainingDuration as startTime + duration - currentTime. This should be a lossless conversion because the currentTime does not change between saving and loading.

I didn't do it this way. When I checked the currenttime in the loading function
AiWander::AiWander (const ESM::AiSequence::AiWander* wander)
I would always get 9. Maybe that wasn't the place to check for it. Instead I stored the remaining duration in startTime's "hour" variable. The "day" variable is now unused.

scrawl:
I do not think the two issues you listed are particularly important, functionally this implementation seems fine to merge with me.

> I didn't do it this way. When I checked the currenttime in the loading function
AiWander::AiWander (const ESM::AiSequence::AiWander* wander)
I would always get 9. Maybe that wasn't the place to check for it. Instead I stored the remaining duration in startTime's "hour" variable. The "day" variable is now unused.

Ah, that's probably because the game time isn't set up yet when the AI packages are loaded (i.e. the global containing the datetime is somewhere further along in the save file).

I suggest that the ESM::AiWander struct should be updated to reflect this new usage (i.e. replace the ESM::TimeStamp with two floats, one titled "remainingDuration" and one titled "unused"). You can add a comment next to it saying that it used to be a timestamp.

I see "3600" quite a lot of times in the code. It might be more convenient to store the duration in hours.

Allofich:
> replace the ESM::TimeStamp with two floats, one titled "remainingDuration" and one titled "unused"

I don't quite understand. If I do that then the load process needs to be changed, too, and existing save games will no longer be usable, right? Or is that acceptable, or should there be some check for both the timestamp and the two floats? (not sure if that's possible) Maybe I'm misunderstanding something.

scrawl:
This is the definition of timestamp:

<pre>
struct TimeStamp
{
    float mHour;
    int mDay;
};
</pre>
So what is stored in the file is effectively just a float and an int.
You can replace TimeStamp with a float and int, existing saves will still load just fine.

Allofich:
Finally seem to have gotten it working. ;) Let me just give it one last test and clean up and I'll upload it.

Allofich:
Oops, uploaded an old version. Just a moment.

scrawl:
@zinnschlag Could you please add "R. D." on the bug tracker to the developers group, I am unable to assign the issue to him.

MiroslavR:
@zinnschlag Ping?

PR #957, opened 2016-06-09
Added 1.5 bonus for ranged weapons when weapons are rated

svaante:
#3405

Tested Melar baram in vanilla and he indeed uses a range spell which is none damaging and in openmw he is using firefist which is a touch spell and his only damaging spell, to get the behavior as in vanilla the rating of damage vs none damage spells needs rework. 
Did some more research on vanilla for the target weapon bug and distance does not matter when the ai is choosing a weapon. Did get Mulvisie Othril to attack with the blade only if I bumped up the shortblade skill to 79 or equipped it with a glass dagger. Which if vanilla rates weapons as expected damage it should corresponds to roughly a flat bonus of + 1.5 rating if ranged.
If the bonus was applied the ai attacked when shortblade was raised to 77 instead or equipped it with a glass dagger.
The solution does not mirror the exact behavior of vanilla, a more precise bonus or another formula can be applied to fix Mulvisie Othril for different weapons and skills etc but the testing process for vanilla that is known to me is way to slow. That is summoning the npc and giving out stats and weapons.

### DIFF
--- a/apps/openmw/mwmechanics/aicombataction.cpp
+++ b/apps/openmw/mwmechanics/aicombataction.cpp
@@ -109,6 +109,10 @@ namespace MWMechanics
             return 0.f;
 
         float rating=0.f;
+        float bonus=0.f;
+
+        if (weapon->mData.mType >= ESM::Weapon::MarksmanBow && weapon->mData.mType <= ESM::Weapon::MarksmanThrown)
+            bonus+=1.5f;
 
         if (weapon->mData.mType >= ESM::Weapon::MarksmanBow)
         {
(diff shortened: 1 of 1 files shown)


Ghostbird:
Is there a more generic way to test for ranged weapons? Currently the approach here is fine, but if we want to add new types of ranged weapons post 1.0 this is going to behave like a magic weapon type.
I know melee weapons have a range rating in vanilla Morrowind, do ranged weapons have something similar with a sensible value? If so, we could simply make the bonus for range depend on the weapon range.
This would work nicely with any future weapon types. Additionally, this may make NPCs intelligently decide to switch over to melee earlier when they're wielding a halberd versus when they're wielding a dagger.
As an additional bonus, I know there exist some wands mods for Morrowind. These work by adding a small stick that is a blunt weapon with a HUGE range, no real damage, and a Cast When Strikes enchantment with a spell of cast type Target. My proposed implementation may make NPCs capable of using these items intelligently (provided their blunt weapon skill is high enough)

scrawl:
I will merge it for now, we can enhance later. Might be worth checking if a bonus should be applied to ranged spells as well?

@svaante please let us know your username on the bugtracker, so that @zinnschlag can add you to the developers group and then we can assign bug 3405 to you.

svaante:
Yeah hi its "svaante".

PR #958, opened 2016-06-09
Add ability to remove multiple rows in subrecords.

Aesylwinn:
This commit implements [feature#2974](https://bugs.openmw.org/issues/2974), making the deletion of multiple subrecords possible.

On a side note, I noticed a particularly nasty hack for checking if subrecords are fixed tables which currently only works for the classes implemented in refidadapterimp.cpp. It is supposed to disable the creation of the add/remove rows context menu actions for fixed tables.

### DIFF
--- a/apps/opencs/view/world/nestedtable.cpp
+++ b/apps/opencs/view/world/nestedtable.cpp
@@ -9,6 +9,7 @@
 #include "../../model/world/universalid.hpp"
 #include "../../model/world/commands.hpp"
 #include "../../model/world/commanddispatcher.hpp"
+#include "../../model/world/commandmacro.hpp"
 
 #include "tableeditidaction.hpp"
 #include "util.hpp"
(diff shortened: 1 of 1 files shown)


Aesylwinn:
@zinnschlag Have you gotten a chance to review this? I'm not entirely sure what was meant by "subrecords" in the feature request, but I assumed it meant the nested tables since everything else seems to have the requested functionality implemented. This commit simply adds the ability to remove multiple rows from a nested table.

As far as the side note goes, I only mentioned it since it obfuscates the code and probably needs to be redesigned at some point. I'm not sure if it should be added as a task to the bug tracker or not though.

zinnschlag:
Opps. Completely missed this one. Sorry. In the future please don't hesitate to poke me when I haven't reacted to a PR within 3-4 days. Reviewing now ...

zinnschlag:
> As far as the side note goes, I only mentioned it since it obfuscates the code and probably needs to be redesigned at some point. I'm not sure if it should be added as a task to the bug tracker or not though.

Sure. We can always move it to a post 1.0 milestone, if we decide it is not so important.

Aesylwinn:
Okay, I'll add that at some point later today.

PR #959, opened 2016-06-10
Slowfall now stops momentum when jumping

svaante:
Bug #3071

First off i don't want to be a nuisance with old or meaningless bugs just want to get familiarized with the code base.
Now slowfall mirrors the momentum behavior in vanilla don't know if it makes more or less sense lore wise.


scrawl:
Thanks! We appreciate any bug fix, no matter how small or insignificant it may be.

PR #961, opened 2016-06-11
Accurately pick the cell ref.

Aesylwinn:
This fixes [bug#3427](https://bugs.openmw.org/issues/3427), though I'm not 100% sure if this is the right fix. I think that [RefNum.mContentFile](https://github.com/OpenMW/openmw/blob/master/components/esm/cellref.hpp#L17) refers to the content file it was loaded from, and not the content file it overrides. Anyone know off the top of their head?

### DIFF
--- a/apps/openmw/mwworld/cellstore.cpp
+++ b/apps/openmw/mwworld/cellstore.cpp
@@ -178,8 +178,12 @@ namespace MWWorld
 
         if (const X *ptr = store.search (ref.mRefID))
         {
-            typename std::list<LiveRef>::iterator iter =
-                std::find(mList.begin(), mList.end(), ref.mRefNum);
+            typename std::list<LiveRef>::iterator iter;
+            for (iter = mList.begin(); iter != mList.end(); ++iter)
+            {
+                if (iter->mRef.getRefNum().mIndex == ref.mRefNum.mIndex)
+                    break;
+            }
 
             LiveRef liveCellRef (ref, ptr);
 


scrawl:
I think you should be looking at the editor code instead, this portion of the game code has been working fine for years.

RefNum.mContentFile refers to the content file that is being overwritten. See [here](https://github.com/OpenMW/openmw/blob/master/components/esm/loadcell.cpp#L19) the code that sets up the RefNum. The editor would need to do the reverse of this function on saving. Not sure if that is already being done.

Aesylwinn:
I was looking at the game code mainly so I could figure out what was going wrong/how it was supposed to work. The save code itself is fine. Using bless and gdb, I was able to determine that the problem is that mContentFile is inaccurate when it is being saved.

Aesylwinn:
When there is a chain of omwaddons that modify the same cell reference, it appears that openmw requires that mContentFile needs to be the content file that first added it.

Aesylwinn:
This matches what the TES CS does.

Aesylwinn:
Here's what I've gleaned from I've looked at:
1. When mContentFile is 0, this means that it is added by the current plugin.
2. An mContentFile equal to 1 refers to the 1st master file, 2 for the 2nd master file, etc.
3. From what I can tell, most Refs that are loaded in the CS have an mContentFile that is 1 less than it should be
4. mContentFile is [ignored](https://github.com/OpenMW/openmw/blob/master/apps/opencs/model/world/refcollection.cpp#L75) in the CS load code for some reason when it decides which Refs are the same

PR #962, opened 2016-06-11
Duration fixes for Follow and Escort

Allofich:
Fixes for bug https://bugs.openmw.org/issues/3432

Summary of the changes made:
- Follow and Escort changed to run on hours, not seconds
- End of remaining duration check fixed (was checking against wrong value)
- Fast-forwarding (resting) now affects remaining duration
- Remaining duration reset after package ends so the package can repeat
- Removed code that sets Escort to infinite duration if it has a destination specified (it was commented that this is in the Construction Set help file, but this behavior does not match what actually happens in the original engine and would be inconsistent with follow packages)

Note that in order to store the starting duration of a follow package, I needed an extra variable. I used the mAlwaysFollow boolean's spot in the save file for this. Saves are compatible as far as I could see. I tested with saving/loading with NPCs, positive and negative mRemainingDurations, and summons (which use the mAlwaysFollow boolean).

There was no variable I could use like this for Escort, so the escort package will just use the mRemainingDuration variable to judge whether or not the package had a duration. If the mRemainingDuration variable is anything but 0 the duration will be set to 1. This only affects repeating editor-placed escort packages.

Really what should happen, I think, is that editor-placed AI packages should read any unchanging data, such as mDuration (not mRemainingDuration) from the .esm/.esp or equivalent file. This seems to be what happens in the original game (the original game doesn't save remaining duration, either) and would fix another problem: currently if you introduce a mod that changes the AI packages of an NPC that has already been saved in the game file, the AI will not change in OpenMW because the AI packages are loaded from the save file. In the original game the new AI packages from the .esp will be loaded.

### DIFF
--- a/apps/openmw/mwmechanics/aiescort.cpp
+++ b/apps/openmw/mwmechanics/aiescort.cpp
@@ -24,29 +24,19 @@
 namespace MWMechanics
 {
     AiEscort::AiEscort(const std::string &actorId, int duration, float x, float y, float z)
-    : mActorId(actorId), mX(x), mY(y), mZ(z), mRemainingDuration(static_cast<float>(duration))
+    : mActorId(actorId), mX(x), mY(y), mZ(z), mDuration(duration), mRemainingDuration(static_cast<float>(duration))
     , mCellX(std::numeric_limits<int>::max())
     , mCellY(std::numeric_limits<int>::max())
     {
         mMaxDist = 450;
-
-        // The CS Help File states that if a duration is given, the AI package will run for that long
-        // BUT if a location is givin, it "trumps" the duration so it will simply escort to that location.
-        if(mX != 0 || mY != 0 || mZ != 0)
-            mRemainingDuration = 0;
     }
 
     AiEscort::AiEscort(const std::string &actorId, const std::string &cellId,int duration, float x, float y, float z)
-    : mActorId(actorId), mCellId(cellId), mX(x), mY(y), mZ(z), mRemainingDuration(static_cast<float>(duration))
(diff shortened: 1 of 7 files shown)


Allofich:
if (follow->mDuration>0)
        mAlwaysFollow = true;
    else
       mAlwaysFollow = false;

This needs to be changed. Since there are probably other things that need fixing I won't re-upload anything yet.

Allofich:
Maybe fooling with the mAlwaysFollow boolean is not such a good idea, since it is used very often (summoned creatures). For a temporary solution I could just do as I did for AIEscort and assign a value of 1 to mDuration when loading a saved game. This would only be a problem for editor-placed Follow packages of limited duration that are in a repeating stack, which I expect are rare.

Allofich:
See
https://github.com/Allofich/openmw/commit/e442afbc259d0476a41742d39a6602f4e60dce09
for the version that does not make changes to mAlwaysFollow.

scrawl:
I wanted to merge https://github.com/Allofich/openmw/commit/e442afbc259d0476a41742d39a6602f4e60dce09 ,  but I can't do so because the commit isn't referenced from any branch, I don't get even this commit when I clone your repository. Can you put it in a branch?

Allofich:
I've put it on the branch used by this pull request.

scrawl:
Error: Tab characters found!
apps/openmw/mwmechanics/aifollow.cpp:56: if (mRemainingDuration != 0)

Allofich:
Sorry about that. Fixed.

Allofich:
Turned off automatic insertion of tabs in Microsoft Visual Studio 2015, which had been causing this problem for me.

psi29a:
Great work! :)

PR #964, opened 2016-06-12
crash fixes for Android platform

sandstranger:
In this PR  fixed crash with loading textures and disabled vbo for android

### DIFF
--- a/components/myguiplatform/myguirendermanager.cpp
+++ b/components/myguiplatform/myguirendermanager.cpp
@@ -126,10 +126,13 @@ public:
             }
 
             osg::Texture2D* texture = batch.mTexture;
-            if(texture)
+            if(texture && texture->getTextureHeight() > 0 && texture->getTextureWidth() > 0)
                 state->applyTextureAttribute(0, texture);
 
-            osg::GLBufferObject* bufferobject = state->isVertexBufferObjectSupported() ? vbo->getOrCreateGLBufferObject(state->getContextID()) : 0;
+            osg::GLBufferObject* bufferobject = 0;
+#if !(defined(ANDROID))
+            bufferobject = state->isVertexBufferObjectSupported() ? vbo->getOrCreateGLBufferObject(state->getContextID()) : 0;
+#endif
             if (bufferobject)
             {
                 state->bindVertexBufferObject(bufferobject);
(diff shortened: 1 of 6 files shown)


sandstranger:
Arm processrors is shit , so it need to increase time step value

components/myguiplatform/myguirendermanager.cpp
```diff
@@ -126,10 +126,13 @@ class Drawable : public osg::Drawable {
             }
 
             osg::Texture2D* texture = batch.mTexture;
-            if(texture)
+            if(texture && texture->getTextureHeight() > 0 && texture->getTextureWidth() > 0)
```
scrawl: What is this check for? The texture size is set on construction ([here](https://github.com/sandstranger/openmw/blob/dcbb581df134a41c7dd827b94d282fb7b19d7797/components/myguiplatform/myguitexture.cpp#L93) and [here](https://github.com/sandstranger/openmw/blob/dcbb581df134a41c7dd827b94d282fb7b19d7797/components/myguiplatform/myguitexture.cpp#L62)), I can't imagine why it would be 0.

sandstranger:
I am using glshim - opengl 1.5 emulator , I believe that this is the problem , but I'll try to find out what is the reason that the size of the texture can turn out to be 0

sandstranger:
And game still crashes , I found new crash .
6-14 08:51:32.327 2923-3067/com.libopenmw.openmw E/MALI: gles_state_set_error_internal:56: GLES ctx: 0x7f9ab1e008, error code:0x500
06-14 08:52:58.105 2923-2962/com.libopenmw.openmw A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x10 in tid 2962 (mali-renderer)
This is all crash log

sandstranger:
Most android videocards support only 16 bit depth buffer size .

sandstranger:
Found new type of crash :
06-26 17:58:39.464 13068-13068/? I/AEE/AED:     #00 pc 00000000000a10e8  /data/app/com.libopenmw.openmw-1/lib/arm64/libopenmw.so
06-26 17:58:39.464 13068-13068/? I/AEE/AED:     #01 pc 0000000000d39080  /data/app/com.libopenmw.openmw-1/lib/arm64/libopenmw.so (SceneUtil::WorkThread::run()+136)
06-26 17:58:39.464 13068-13068/? I/AEE/AED:     #02 pc 0000000000a1b27c  /data/app/com.libopenmw.openmw-1/lib/arm64/libopenmw.so (OpenThreads::ThreadPrivateActions::StartThread(void*)+164)
06-26 17:58:39.464 13068-13068/? I/AEE/AED:     #03 pc 000000000001f38c  /system/lib64/libc.so (__pthread_start(void*)+52)
06-26 17:58:39.464 13068-13068/? I/AEE/AED:     #04 pc 000000000001b604  /system/lib64/libc.so (__start_thread+16)

PR #967, opened 2016-06-15
AI package cleanups

Allofich:
Some cleanups to AI package files.

AIEscort:
Removed TODO comment about testing vanilla behavior with destination of 0, 0, 0.
In testing I found that while in the original engine an editor-placed package works OK, using
AIEscort or AIEscortCell through the console with a destination of 0,0,0 causes the NPC to not move. Changing it to e.g. 1,0,0 will make it work. I suppose this was the reason for this TODO, but I don't see it as something that is desirable for us to emulate. Why would we want to prevent AIEscort or AIEscortCell from working with destination 0,0,0? Currently, OpenMW works with these destinations.

AIWander:
I moved variables that were commented with FIXME into AIWander's AIWanderStorage. I've tested and everything seems to be working OK, but as it involved code I was unfamiliar with, it's possible I could have messed something up somewhere. In particular a review of the changes I made to the fastforward function would be appreciated.

I also removed setting the bool mTrimCurrentNode to false in init(), as it is set so in the AIWanderStorage constructor, which I think should be enough.

### DIFF
--- a/apps/openmw/mwmechanics/aiescort.cpp
+++ b/apps/openmw/mwmechanics/aiescort.cpp
@@ -16,7 +16,6 @@
 #include "movement.hpp"
 
 /*
-    TODO: Test vanilla behavior on passing x0, y0, and z0 with duration of anything including 0.
     TODO: Different behavior for AIEscort a d x y z and AIEscortCell a c d x y z.
     TODO: Take account for actors being in different cells.
 */
@@ -109,7 +108,7 @@ namespace MWMechanics
         }
         else
         {
-            // Stop moving if the player is to far away
+            // Stop moving if the player is too far away
             MWBase::Environment::get().getMechanicsManager()->playAnimationGroup(actor, "idle3", 0, 1);
             actor.getClass().getMovementSettings(actor).mPosition[1] = 0;
             mMaxDist = 250;
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -75,6 +75,21 @@ namespace MWMechanics
         std::vector<unsigned short> mBadIdles; // Idle animations that when called cause errors
 
         PathFinder mPathFinder;
+
+        // do we need to calculate allowed nodes based on mDistance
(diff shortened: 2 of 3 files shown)


scrawl:
I'd say that ```bool mPopulateAvailableNodes;``` also has to be moved into AiWanderStorage, since it's referring to the initialization of the mAllowedNodes list which is now in AiWanderStorage.

Allofich:
Additional changes made:
bool mPopulateAvailableNodes, mObstacleCheck, mDoorCheckDuration and mStuckCount moved into AiWanderStorage. All variables in aiwander.hpp commented as needing to be moved to AiWanderStorage are moved now.

Initializing lines in init() for these variables removed.

24-hour limit on duration that I inserted in init() in a previous commit (related to an earlier attempt at working with TimeStamps) removed, since it's not necessary now and seems like an arbitrary limitation.

Bounds checking on timeOfDay in init(), 
if(mDuration == 0)
            mTimeOfDay = 0;
removed since timeOfDay is unimplemented.

Allofich:
If scrawl or anyone else is interested, here is a WIP version that goes further and makes the following changes:
- AIActivation.cpp rewritten to use MWMechanics namespace
- Attempted to address FIXME variables PathFinder mPathFinder and ObstacleCheck mObstacleCheck by putting them in struct AiTemporaryBase in aistate.hpp. This involved modifying virtually all of the AI files, and I only kind of knew what I was doing, so I'm not very confident about this change.

https://github.com/Allofich/openmw/commit/f0cdfa3a317a5c405bec1f5a4e474f6a641508b8

Unfortunately, this version has a problem that I don't know how to fix at the moment. An NPC cycling between two Activate packages will, after finishing the second activate package, get stuck attempting to walk into the location of the target of the second Activate package. If that target object is disabled so they can walk to the location where it was, they will then proceed to cycle back to the first target properly, but this problem doesn't happen without this change. I assume the NPC is trying to complete the path they created get to the second target object, but I don't know exactly what is happening or why the problem doesn't happen with the first activate package's target.

Allofich:
Made the WIP version into a second commit rebased on the first.

PR #969, opened 2016-06-17
Fix Microsoft Visual Studio 2015 warnings

Allofich:
Fixes the following warnings from Microsoft Visual Studio 2015

warning C4464: relative include path contains '..'

\apps\openmw\mwmechanics\aiactivate.cpp(45): warning C4456: declaration of 'target' hides previous local declaration
\apps\openmw\mwmechanics\aiactivate.cpp(30): note: see declaration of 'target'

\apps\openmw\mwmechanics\aisequence.cpp(95): warning C4457: declaration of 'package' hides function parameter
\apps\openmw\mwmechanics\aisequence.cpp(88): note: see declaration of 'package'

Also includes organizing of include statements.

### DIFF
--- a/apps/openmw/mwmechanics/aiactivate.cpp
+++ b/apps/openmw/mwmechanics/aiactivate.cpp
@@ -2,13 +2,12 @@
 
 #include <components/esm/aisequence.hpp>
 
-#include "../mwbase/world.hpp"
-#include "../mwbase/environment.hpp"
+#include "apps/openmw/mwbase/world.hpp"
+#include "apps/openmw/mwbase/environment.hpp"
 
-#include "../mwmechanics/creaturestats.hpp"
+#include "apps/openmw/mwworld/class.hpp"
 
-#include "../mwworld/class.hpp"
-
+#include "creaturestats.hpp"
 #include "steering.hpp"
 #include "movement.hpp"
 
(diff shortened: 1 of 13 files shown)


Allofich:
If this PR is OK, I can provide the same kind of include fixes for other files.

Aesylwinn:
Correct me if I'm wrong, but I believe that many files follow the same "../" pattern in multiple areas. To change all of the files would lead to potential merge conflicts.

Allofich:
After re-thinking this, I'm leaning toward abandoning this kind of change. I don't know if merge conflicts would happen, but there is another problem: it is probably not good to go by a "standard" (Microsoft Visual Studio's compiler) that not everyone has easy access to. In the case of travis and app-veyor warnings, everyone submitting here can see them and correct their own code, but other people may unknowingly add code that causes warnings in Microsoft Visual Studio, and I or another Visual Studio user can't go around and and fix (not even necessarily a "fix") other people's code afterwards.

Unrelated to Visual Studio warnings, I noticed that include statements could use a general cleanup, though, which I think I will open a PR for.

zinnschlag:
Agreed. This is perfectly fine C++ code. Changing it to suppress the warnings is not an improvement. The solution should be to silence these overly aggressive warnings instead.

psi29a:
You have to be in a file, for whatever reason, then I think it is reasonable to clean it up as well.

Allofich:
OK, will close the PR.

PR #971, opened 2016-06-18
AI package cleanups

Allofich:
Formatting cleanups to AI packages I am looking at contributing to.

This may look like a lot, but it's mostly just indenting fixes to match the conventions laid out here
https://wiki.openmw.org/index.php?title=Code_Formatting_Conventions

In addition to indenting fixes, aipackage.cpp was given a namespace block, and variables were grouped by type for aifollow and aiwander, and the constructor initializations reordered to match.

I was just recently advised against needlessly cleaning the code due to potential merge conflicts, but there seems to be a lull in activity at the moment and my thoughts were that it would be good to get these out of the way now before attempting any functional changes.


apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -674,126 +671,123 @@ namespace MWMechanics
...
+
+        if (weapon == NULL)
         {
-            movement.mPosition[1] = 1;
-            attackType = ESM::Weapon::AT_Thrust;
+            //hand-to-hand deal equal damage for each type
+            float roll = Misc::Rng::rollClosedProbability();
+            if(roll <= 0.333f)  //side punch
+            {
+                movement.mPosition[0] = Misc::Rng::rollClosedProbability() ? 1.0f : -1.0f;
```
terrorfisch: This line looks strange to me. Shouldn't it be something like?
```
movement.mPosition[0] = Misc::Rng::rollDice(2) ? 1.0f : -1.0f;
```

terrorfisch: Indeed the behaviour was changed in this commit: 3f28634d1f617691c672e41a3ee950e6daec8c77

scrawl:
Sorry, I do not think this is worthwhile. In addition to the risk of merge conflicts (which admittedly is low due to lack of activity in this area), doing these mass formatting changes also reduces the effectiveness of ```git blame```.

If you are changing a piece of code anyway, then you can clean up as you go, otherwise I would leave it alone.

Allofich:
No problem, scrawl. I appreciate your honesty. I'll go ahead and close this, then.

apps/openmw/mwmechanics/aicombat.cpp
scrawl: Good catch. Would you like to open a PR?

terrorfisch: Done

PR #972, opened 2016-06-20
Make NPCs use both hands in weaponless combat.

terrorfisch:
Bug was introduced during RNG encapsulation: 3f28634d1f617691c672e41a3ee950e6daec8c77

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -688,7 +688,7 @@ ESM::Weapon::AttackType chooseBestAttack(const ESM::Weapon* weapon, MWMechanics:
         float roll = Misc::Rng::rollClosedProbability();
         if(roll <= 0.333f)  //side punch
         {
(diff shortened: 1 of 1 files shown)


Allofich:
This isn't a criticism, just wondering. It seems like NPCs already use both hands in weaponless combat (when punching) before this fix, or at least they play the animation such. Can you explain what the difference is?

terrorfisch:
Isn't the direction of the slash dependent on the movement direction? However, I don't get the choosing of the attack move as it appears the return value of `chooseBestAttack()` is dropped.

Allofich:
In the original game, it's a bit difficult to tell because the NPC shifts around slightly while punching (although not as much as in OpenMW), but I think the punch for an NPC is not dependent on movement direction.

I looked at aicombat.cpp and as you say it seems like the return value of chooseBestAttack() isn't actually used anywhere. I tested with replacing the function to simply return the same attack type every time, but in game a variety of attacks are still used, so it seems the return value here is not actually used to choose the attack.

PR #973, opened 2016-06-21
Fixed drained stat bar appearance (#2680)

ckendrick:
Fixes [Bug #2680](http://bugs.openmw.org/issues/2680), shows the current value / base instead of the current value / drained value in both the stats window and on the HUD.


scrawl:
I don't think this is a GUI issue. The "drain health/magicka/fatigue" mechanic is implemented differently compared to the original game. The <i>current</i> value should be drained, not the <i>modified</i> one.

ckendrick:
Sorry about that, you are right. The way I have it, it wouldn't show the stats correctly with a modified health. Is it worth looking into fixing the implementation of the drain h/m/s mechanic?

jeffreyhaines:
@ckendrick are you still looking to complete this issue? If not could you close this pull request. Thank you

PR #974, opened 2016-06-21
Update disposition check between persuasion attempts

Allofich:
Fix for https://bugs.openmw.org/issues/3445, by updating the disposition check between persuasion attempts.

This seems to match the behavior I observe in the original engine.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.hpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.hpp
@@ -92,7 +92,7 @@ namespace MWMechanics
             virtual int getBarterOffer(const MWWorld::Ptr& ptr,int basePrice, bool buying);
             ///< This is used by every service to determine the price of objects given the trading skills of the player and NPC.
 
-            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr, bool addTemporaryDispositionChange = false);
+            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr, bool addTemporaryDispositionChange = true);
             ///< Calculate the diposition of an NPC toward the player.
 
             virtual int countDeaths (const std::string& id) const;


Allofich:
Really the title should have been "update disposition check between persuasion attempts" I guess.

scrawl:
Oops, looks like I introduced this bug when I refactored someone's PR.

getDerivedDisposition(ptr) is supposed to add temporary disposition by default. In mechanicsmanager.hpp:
<pre>
            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr, bool addTemporaryDispositionChange = true) = 0;
</pre>

However in mechanicsmanagerimp.hpp, the implementation for this function has a different default (classic mistake of changing code in one place while forgetting to update other places).
<pre>
            virtual int getDerivedDisposition(const MWWorld::Ptr& ptr, bool addTemporaryDispositionChange = false);
</pre>

I suggest that the declaration in mechanicsmanagerimp.hpp should be changed to default to addTemporaryDispositionChange=true. That should fix this issue, as well as the getDerivedDisposition() usage in MechanicsManager::getBarterOffer() that appears to have the same bug.

Allofich:
OK, well you can just make the change yourself if you like, unless you want me to get the "credit" for it.

Allofich:
I guess I'll go ahead and make the change to this PR, since you suggested I do so.

Allofich:
I'll test now just to make sure this fixes the issue.

Allofich:
Seems to work fine!

scrawl:
Commit message should say "dispositon" not "duration".

Allofich:
Oops. Will fix now.

Allofich:
Habit from working on the duration changes earlier.

Allofich:
One minor favor I would like to ask. If it is possible for you to delete messages on the bugtracker on the OpenMW site, can you delete my message on Bug #3445 about how persuasion check works? I realize now it is a bit of a spoiler and cheat to know that disposition 50 has the highest likelihood of success for persuasion checks. If I was just playing Morrowind I would not want to know that. If you cannot delete it or the policy is to not delete messages, that's fine.

Allofich:
Oh well nevermind I guess. The bugtracker is full of minor spoilers. I worry about these kind of minor things too much.

PR #975, opened 2016-06-24
Improve the MSVC prebuild script to allow compiled binaries to be run directly with no further modification

AnyOldName3:
Before this change, having run the MSVC prebuild script, and built any apps, resource and configuration files would be present in the Build_64 folder, but not the Debug or Release folder, preventing the compiled binaries actually being run.

This change copies these files to the Debug or Release folder (depending on build configuration mode) where they can be accessed by the binaries, allowing them to be run without any manual relocation of files.

This was discussed somewhat with Ace here: https://forum.openmw.org/viewtopic.php?f=8&t=3264&start=10#p36370

Having come across the same problem again, I decided to attempt to resolve it permanently: https://forum.openmw.org/viewtopic.php?f=22&t=3629#p40446

This has been tested by re-running the prebuild script in both debug and release mode, and observing that the correct files (as defined by Ace) appeared in the intended locations, and also that the compiled binaries now ran without error. This *should* be sufficient due to the small scope of changes.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -653,6 +653,19 @@ if [ -z $CI ]; then
 		cp "$DLL" $CONFIGURATION/osgPlugins-3.3.8
 	done
 	echo
+	
+	echo "Copying Runtime Resources/Config Files"
+	
+	echo "  gamecontrollerdb.txt"
+	cp $CONFIGURATION/../gamecontrollerdb.txt $CONFIGURATION/gamecontrollerdb.txt
+	echo "  openmw.cfg"
+	cp $CONFIGURATION/../openmw.cfg.install $CONFIGURATION/openmw.cfg
+	echo "  openmw-cs.cfg"
+	cp $CONFIGURATION/../openmw-cs.cfg $CONFIGURATION/openmw-cs.cfg
+	echo "  settings-default.cfg"
+	cp $CONFIGURATION/../settings-default.cfg $CONFIGURATION/settings-default.cfg
+	echo "  resources/"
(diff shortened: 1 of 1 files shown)


scrawl:
Is there a way to do the copying in the CmakeLists? I'm trying to avoid the list of resource files being duplicated in several different places.

Or perhaps there is a way to avoid the Debug/Release folders from being created in the first place?

AnyOldName3:
Unfortunately, I have no idea how to write CMake scripts. I might be able to investigate, though.

If the debug and release folders aren't used at all, then it would make having compiled binaries for both configurations difficult, as they have the same file names, so would overwrite each other. As well as this, the build script would need to be re run whenever configurations were switched between, as different versions of the runtime DLLs would be required.

As an alternative, it might be possible to parse the script containing the list of requirements in the prebuild script. This would fix the duplication issue, but would get complicated and potentially messy.

The ideal solution would be to make the INSTALL subproject of the VS solution actually work, but that's definitely outside the scope of things I'm confident doing.

scrawl:
> If the debug and release folders aren't used at all, then it would make having compiled binaries for both configurations difficult, as they have the same file names, so would overwrite each other. As well as this, the build script would need to be re run whenever configurations were switched between, as different versions of the runtime DLLs would be required.

If you want separate builds, just make a new build folder. Or rebuild when you switch, only takes a few minutes.

> The ideal solution would be to make the INSTALL subproject of the VS solution actually work

Ideally you wouldn't need to ```install``` to run the compiled binaries.

And it seems that ```install``` on windows is already used for deploying (runs CPack, etc.).

PR #976, opened 2016-06-24
Fix magicka and fatigue calculations from fortifying attributes

Allofich:
Fix for https://bugs.openmw.org/issues/2604.

This addresses the two issues mentioned in this bug, making magicka retain it's ratio between current magicka and maximum magicka before and after a fortify intelligence effect, and allowing magicka to go into the negatives after a fortify magicka spell has ended.

In game testing of the effects of magicka fortify and drain/damage spells showed that everything seemed to be working right.

This doesn't address the problem of the "fortify magicka" effect working differently from the original game. In the original game fortify magicka does not boost the maximum magicka but can take you over if the fortify is enough to take you over your pre-fortify maximum. Fortify health and fortify endurance are also like this.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -425,7 +425,9 @@ namespace MWMechanics
 
         DynamicStat<float> magicka = creatureStats.getMagicka();
         float diff = (static_cast<int>(magickaFactor*intelligence)) - magicka.getBase();
-        magicka.modify(diff);
+        float currentToBaseRatio = (magicka.getCurrent() / magicka.getBase());
+        magicka.setModified(magicka.getModified() + diff, 0);
+        magicka.setCurrent(magicka.getBase() * currentToBaseRatio);
         creatureStats.setMagicka(magicka);
     }
 
@@ -553,8 +555,9 @@ namespace MWMechanics
             DynamicStat<float> stat = creatureStats.getDynamic(i);
             stat.setModifier(effects.get(ESM::MagicEffect::FortifyHealth+i).getMagnitude() -
                              effects.get(ESM::MagicEffect::DrainHealth+i).getMagnitude(),
+                             // Magicka can be decreased below zero due to a fortify effect wearing off
                              // Fatigue can be decreased below zero meaning the actor will be knocked out
-                             i == 2);
+                             i == 1 || i == 2);
 
             creatureStats.setDynamic(i, stat);
         }
--- a/apps/openmw/mwmechanics/creaturestats.cpp
+++ b/apps/openmw/mwmechanics/creaturestats.cpp
@@ -157,7 +157,9 @@ namespace MWMechanics
                 int endurance    = getAttribute(ESM::Attribute::Endurance).getModified();
                 DynamicStat<float> fatigue = getFatigue();
                 float diff = (strength+willpower+agility+endurance) - fatigue.getBase();
-                fatigue.modify(diff);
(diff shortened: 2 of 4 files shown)


apps/openmw/mwmechanics/actors.cpp
```diff
@@ -553,8 +555,9 @@ namespace MWMechanics
             DynamicStat<float> stat = creatureStats.getDynamic(i);
             stat.setModifier(effects.get(ESM::MagicEffect::FortifyHealth+i).getMagnitude() -
                              effects.get(ESM::MagicEffect::DrainHealth+i).getMagnitude(),
+                             // Magicka can be decreased below zero due to a fortify effect wearing off
                              // Fatigue can be decreased below zero meaning the actor will be knocked out
-                             i == 2);
+                             i == 1 || 2);
```
Mingun: This condition always `true`. I think, you m,ean `i == 1 || i == 2`?

Allofich: Ah yes, thanks.

Allofich:
Just did a test and, as I expected, looks like endurance getting boosted by a fortify effect has the same problem.

Allofich:
Updated to fix endurance as well. I noticed there are problems with the display of the endurance value, though. If you equip a strength fortifying enchanted item at full fatigue, the fatigue display in the stat window won't update. Also, if you have full fatigue and load a game, the loaded game will show the fatigue from the pre-load save file.

scrawl:
Did you see PR #798 ? The situation is a bit messy because we have 3 similar copies of this function, your proposed change only addresses one of these. This could lead to further bugs in the future, as you would expect that ```setModified(getModified()+x)``` is the same as ```modify(x)```.

IMO what needs to be done is:
* ensure that logic in ```modify```, ```setModified``` and ```setModifier``` is the same (e.g. modify() would just be a shortcut to calling setModified(getModified()+diff). or we could just remove two of those functions and have the caller deal with adding a number.
* test all the use cases of ```modify```, ```setModified``` and ```setModifier``` for ratio preserving behaviour
* if needed, add a ```preserveRatio``` argument that can be true or false.

Allofich:
I didn't know about that PR. Well, I'll take another look at this in a while then, see if I can address the things you mention. Then we could close that other PR too I guess.

Allofich:
I've examined the three functions, and here's what I think.

- `setModifier` and `setModified` do not need to be modified for this commit because they do not set the current value of a stat. `setModifier` is used for tallying up modifiers before applying them outside the function (without preserving ratio) and `setModified` is used for commands like "setmagicka" or "modmagicka" in the console or script. These don't need to preserve the ratio, either.

- `modify` is only used by the fatigue and magicka recalculations, which need to have the ratio preserved. It could either have its name changed to describe that it preserves the ratio, have a preserveRatio flag added, or, since it is only used in two spots, be removed.

I've put up two implementations, one adding the flag so the old behavior of making the change ignore the ratio is still possible, and one removing the function and handling the ratio preservation where the calls from the magicka and fatigue recalculations are done using `setModified`.

The implementation that adds the flag is here. https://github.com/Allofich/openmw/commit/fec73e53ff489684074fd5793c85581ce1ccecbd

This changes `modify` in `DynamicStat` but it calls another `modify` used within `Stat` that is unrelated to preserving ratio (there is no `mCurrent` in `Stat`).

To simplify things and slim down the code, my suggestion is removing both `modify` functions so I'm uploading that implementation to this PR.

PR #978, opened 2016-06-25
Fix update of fatigue display

Allofich:
Fixes https://bugs.openmw.org/issues/3449

The fatigue display was updated before void MechanicsManager::update in mechanicsmanagerimp.cpp could update the stat window. Now fatigue's update works in the same way as magicka's update.

I made two implementations. In this one, which is the one I think is probably more desirable, magicka and fatigue calculating are separated into two functions.

In the other, which does have the advantage of fewer lines of code, magicka and fatigue are always recalculated together. It is available at https://github.com/Allofich/openmw/tree/fatigue2.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -310,7 +310,7 @@ namespace MWMechanics
     }
 
     MechanicsManager::MechanicsManager()
-    : mWatchedStatsEmpty (true), mUpdatePlayer (true), mClassSelected (false),
+    : mWatchedTimeToStartDrowning(0), mWatchedStatsEmpty (true), mUpdatePlayer (true), mClassSelected (false),
       mRaceSelected (false), mAI(true)
     {
         //buildPlayer no longer here, needs to be done explicitely after all subsystems are up and running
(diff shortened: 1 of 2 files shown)


scrawl:
Yes, the fatigue/magicka updates needs to be separate. AFAIK with creatures, there is a magicka value set in the game files, but if their attributes are changed at runtime then the autocalculated magicka is used. If you autocalculate magicka even though intelligence didn't change, then these creatures would not be using the same value of magicka as in the original game.

I don't understand why delaying the fatigue update would fix the display bug. Seems like a workaround to some other issue that we haven't identified yet?

scrawl:
> fatigue/magicka updates needs to be separate

See https://bugs.openmw.org/issues/2155 for more context on this.

Allofich:
I don't know exactly, but in `void MechanicsManager::update` is the following
           ` if(stats.getFatigue() != mWatchedStats.getFatigue() || mWatchedStatsEmpty)
            {
                static const std::string fbar("FBar");
                mWatchedStats.setFatigue(stats.getFatigue());
                winMgr->setValue(fbar, stats.getFatigue());
            }`

When I checked what values were being given, `stats.getFatigue()` and `mWatchedStats.getFatigue()` were always being the same in the condition where I fortified strength when I had max fatigue. If you have less than maximum fatigue, the update will work, because `mWatchedStats.getFatigue()` will be less than `stats.getFatigue()` as the fatigue is regenerating. Since it was already at maximum, though, that change wasn't occurring, and somehow updating fatigue in creaturestats.cpp must have been setting `mWatchedStats.getFatigue()` and `stats.getFatigue()` to their new, equal values before `update` had a chance to see it.

Allofich:
OK, thanks for the info that magicka and fatigue need to be separate.

scrawl:
I don't see how mWatchedStats could be modified from outside the mechanics manager. It is not being exposed or shared or anything like that.

Allofich:
I don't know. In my test case I have a maximum of 170 fatigue, being brought up to 195 by a strength enchantment. When I checked the mWatched and stats values, fatigue was being updated well before a magicka fortify and to its new value of 195, and by the time it got around to the update function stats and mWatched were both 195 and so the update wouldn't run.

Allofich:
I was checking by using a cout statement, and if I am not mistaken there were many couts from the fatigue updating function reading 195 before the line from update for the magicka change came around.

Allofich:
Just to be sure, do you have the bug I reported when you run on your computer?

scrawl:
> Just to be sure, do you have the bug I reported when you run on your computer?

Yes.

I see why this is happening. There is fatigue recalculation happening inside the mWatchedStats when strength is updated:
<pre>
mWatchedStats.setAttribute(i, stats.getAttribute(i));
</pre>

scrawl:
The change you propose does fix the bug because the update is now explicit (i.e. calculateFatigue won't be called for the mWatchedStats).

But there is a problem here:
<pre>
+        state.mRecalcDynamicStats = (mRecalcMagicka || mRecalcFatigue);
</pre>
If you save in the frame that fatigue was about to be updated, then both magicka and fatigue will be updated. We don't want that (bug https://bugs.openmw.org/issues/2155 )

Allofich:
Ah, right. Well, before this commit mRecalcDynamicStats was just functioning as mRecalcMagicka, despite the name. So, I can just return it to that.

Allofich:
There are a couple other places, too, that need to be changed so that fatigue and magicka aren't both recalculated.

Allofich:
Any other changes you want at the moment?

scrawl:
Maybe it would be better to use simple arrays instead of an ```mWatchedStats```.

<pre>
AttributeValue mWatchedAttributes[8];
....
</pre>

That way we can be sure that updating the watched value has no unintended side effects.

scrawl:
> That way we can be sure that updating the watched value has no unintended side effects.

CreatureStats::setDynamic() has side effects as well, when the health goes to zero:
<pre>

        if (index==0 && mDynamic[index].getCurrent()<1)
        {
            if (!mDead)
                mTimeOfDeath = MWBase::Environment::get().getWorld()->getTimeStamp();

            mDead = true;

            mDynamic[index].setModifier(0);
            mDynamic[index].setCurrent(0);

            if (MWBase::Environment::get().getWorld()->getGodModeState())
                MWBase::Environment::get().getMechanicsManager()->keepPlayerAlive();
        }
</pre>

Allofich:
OK, well, if I am going to do it I will have to look at it with a fresh mind later. I need some rest now. I'll be away from the computer for a while so take your time if you want to consider about what advice to give.

Allofich:
I changed mWatchedStats to an array for the tracking of attributes in  MechanicsManager::update(). I guess it was a pretty simple thing to do and I didn't actually have to rest first :).

Is there a problem with CreatureStats::setDynamic()? I see that it modifies the values, but in testing I didn't notice any problems with the display update of the health after setting health to 0, with or without god mode.

Also the behavior I had added that changed the recalcfatigue bool together with the recalcmagicka bool is gone now.

scrawl:
> Is there a problem with CreatureStats::setDynamic()? I see that it modifies the values, but in testing I didn't notice any problems with the display update of the health after setting health to 0, with or without god mode.

I think it will call World::keepPlayerAlive() <i>twice</i> when the player dies in god mode, which is not really a problem, just a bad practice I guess.

> I changed mWatchedStats to an array for the tracking of attributes in MechanicsManager::update().

For consistency, perhaps we should also store skills, health/magicka/fattigue and timeToStartDrowning as separate values, then the mWatchedStats can be removed.

The changes to CreatureStats might as well be reverted because they're no longer needed, right?

Allofich:
> I think it will call World::keepPlayerAlive() twice when the player dies in god mode, which is not really a problem, just a bad practice I guess.

You were right. This can be avoided by setting current to 1, not just modified to 1, during the resurrection process for god mode.

> For consistency, perhaps we should also store skills, health/magicka/fattigue and timeToStartDrowning as separate values, then the mWatchedStats can be removed.

Done. Also skills were a part of mWatchedStats.

> The changes to CreatureStats might as well be reverted because they're no longer needed, right?

That's right, just changing the attributes to use an array fixed the fatigue display, so I reverted the CreatureStats changes.

Allofich:
> This can be avoided by setting current to 1
Actually it can seemingly be avoided without problem by just removing a bit of roundabout code.

PR #979, opened 2016-06-26
Show correct class image in level-up dialog

MiroslavR:
Bug report: http://bugs.openmw.org/issues/3443

OpenMW's current behavior of choosing the player's class image in the level-up dialog is wrong.

What image is displayed actually depends on the number of skill increases for each of the three skill specializations (combat, magic & stealth). Unlike the increases that affect attribute multipliers, these numbers do not reset when the player levels up.

I have also updated essimporter to pull these numbers from vanilla savegames.

### DIFF
--- a/apps/openmw/mwgui/levelupdialog.cpp
+++ b/apps/openmw/mwgui/levelupdialog.cpp
@@ -340,7 +340,7 @@ namespace MWGui
                     ret = "pilgrim";
                 break;
             case 3:
-                if (magicFraction == 3 && stealthFraction == 3)
+                if (magicFraction == 3)
                     ret = "bard";
                 break;
             default:
--- a/apps/openmw/mwgui/levelupdialog.cpp
+++ b/apps/openmw/mwgui/levelupdialog.cpp
@@ -156,7 +156,9 @@ namespace MWGui
                 cls = &*it;
         }
 
-        setClassImage(mClassImage, cls->mId);
+        setClassImage(mClassImage, getLevelupClassImage(pcStats.getSkillIncreasesForSpecialization(0),
+                                                        pcStats.getSkillIncreasesForSpecialization(1),
+                                                        pcStats.getSkillIncreasesForSpecialization(2)));
 
         int level = creatureStats.getLevel ()+1;
         mLevelText->setCaptionWithReplacing("#{sLevelUpMenu1} " + MyGUI::utility::toString(level));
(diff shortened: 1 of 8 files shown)


components/esm/npcstats.cpp
```diff
@@ -95,6 +95,8 @@ void ESM::NpcStats::load (ESMReader &esm)
 
     esm.getHNT (mSkillIncrease, "INCR");
```
scrawl: mSpecIncreases needs to be initialized here. Initializing in blank() is not enough (blank() is only used for ESS-importing at the moment).

apps/openmw/mwgui/levelupdialog.cpp
```diff
@@ -248,4 +250,103 @@ namespace MWGui
...
+                    ret = "acrobat";
+                break;
+            case 5:
+                if (magicIncreases == 3)
+                    ret = "monk";
+                else
+                    ret = "pilgrim";
+                break;
+            case 3:
+                if (magicFraction == 3 && stealthFraction == 3)
```
Allofich: I think this `stealthFraction` is supposed to be `combatFraction`, right?

Allofich: In the bard calculation, I think `stealthFraction` is supposed to be `combatFraction,` right?

apps/openmw/mwgui/levelupdialog.cpp
```diff
@@ -248,4 +250,103 @@ namespace MWGui
...
+                    ret = "bard";
+                break;
+            default:
+                break;
+        }
+
+        return ret;
+    }
 }
```
Allofich: In the bard calculation, I think `stealthFraction` is supposed to be `combatFraction.`

(Sorry if you received multiple emails from GitHub, this is my first time using the diff inline commenting function and it took me a few tries to get it how I wanted it.)

Allofich:
Were you able to see or infer what the intended calculation was for the nightblade? Since it looks like the only class that was left out, if it's salvageable I think it would be a nice bug fix to put in OpenMW.

apps/openmw/mwgui/levelupdialog.cpp
MiroslavR: While what you said could have been the intent, the original code does use `stealthFraction` here. Of course, `stealthFraction == 3` is redundant then.

MiroslavR:
> Were you able to see or infer what the intended calculation was for the nightblade? Since it looks like the only class that was left out, if it's salvageable I think it would be a nice bug fix to put in OpenMW.

It's another of those redundant cases...

    switch (magicFraction) {
        case 4:
            if (magicFraction == 4)
                ret = "spellsword";
            else
                ret = "nightblade";
            break;
    }

It's definitely not the 'default' case, nor are there any fall-throughs that I can see. I have no idea why the compiler didn't optimize it out, but the code does appear to be unreachable. As for the original intent, we can only guess.

MiroslavR:
Anyway, I have addressed the code review comments.

apps/openmw/mwgui/levelupdialog.cpp
Allofich: Well, OK, but it did seem from when I tested the original game that bard would correctly show up when combat, magic and stealth were all equal, so it seems odd that combatfraction would be ignored. I trust that you have a better grasp of what exactly the original game is doing, though.

Allofich: After looking at your code more closely, yeah, I suppose combatFraction will have to be 3 here anyway, since stealthFraction is 3 and MagicFraction is 3, and the case of combatFraction being 4 is already handled in the combat switch.

Allofich: I wonder if those spots that use "increases" instead of "fraction" are mistakes in the original code? It looks like they wouldn't conflict with any of the other fraction conditions if they were fraction, and if increases are used, it seems like those class images would be quite rare to get.

Allofich: Another thing, which I noted in the bug report (I'm using the name "R.D." there) is that only skills that are advanced through actual use are counted, not skills gained through training or books. I'm not sure why the developers might have not wanted to allow training or books skills to have an effect, as to me they seem as much like a true part of the character as skills gained through actual use. That also might be an oversight.

Well, I don't know what the official OpenMW policy is on implementation is in situations like this, maybe to just stick to the original behavior. My own preference is to go for what looks most likely to have been the original intent, although that introduces the problem of interpretation. My preference would be, while documenting the original behavior in the source code or somewhere:
- use fractions in place of the increases
- count advances made through books and training
- implement nightblade (something like if stealthfraction = 3 seems like a likely candidate)
- reset the count on leveling

However I think you've already done great work and I don't mean to make any kind of demands of you. As you say this has little impact and it's not a big deal if you just want to be done with it.

Allofich: If the count was reset, though, I guess that introduces complications with importing from .ess files, though.

Allofich: > (something like if stealthfraction = 3 seems like a likely candidate)

Actually I guess if something were to be done here, a more consistent condition would be to use a combatfraction condition on spellsword, and then fall-through to nightblade, as It looks like only one type of fraction is used in each switch.

Looking more closely the conditions are certainly strange, though... "Healer" has a higher combatfraction (if it was changed to fraction) than "sorcerer," for example.

MiroslavR: There's a lot of maybes and making stuff up, which may go against our policy. I'd like to get the green light from @scrawl or @zinnschlag first before pursuing this.

Allofich: Sure, it was just a suggestion. I thought I would put it up for discussion while we're looking at the issue.

It seems like in other cases where the original game has an implementation that has been questioned, like that of pickpocketing or how bartering limits purchase and selling price to the minimum of the two, the original game's implementation has been put in and they are marked for "post 1.0" on the tracker, so maybe that's the way to go here as well. In that case, I suppose we might want to make this implementation ignore skill changes from book and training increases like vanilla. Maybe it does so, I can't tell from looking at the diff alone.

Allofich: I do think those redundant conditions in bard and spellsword may offer clues to what the original author(s) intended, so if we don't have them in the OpenMW source, I think it would be nice to have the original code you reverse-engineered documented, perhaps on the research pages of the OpenMW Wiki.

scrawl: > There's a lot of maybes and making stuff up, which may go against our policy.

I wouldn't reject an improved version, but I don't think it's very important either.

Did you see my comment here? https://github.com/OpenMW/openmw/pull/979/files/0db3af1b894adf1f2188ccae5c6f17950b5e74d1#r68496786

MiroslavR: > Did you see my comment here?

Yep, I have already committed the fix.

scrawl: > Yep, I have already committed the fix.
 
I see it now, looks like github was showing me an earlier version of this PR for some reason.

Will test and merge if there are no issues.

PR #980, opened 2016-06-27
Voice fix

scrawl:
For review by @kcat

Rewrites the lip syncing code with the intent of fixing bug [#3453](https://bugs.openmw.org/issues/3453).

I suspect that the FFmpeg_Decoder::rewind() is causing the bug. The implementation might be incomplete or incorrect, or there could be a bug in ffmpeg after all. I'm not too keen on investigating, so I decided to rewrite the code so as to no longer require rewind().

Instead of getting the loudness data for the whole file in advance, we now get it piece by piece as the sound is streamed. Aside from performing better and simplifying the sound manager code, this cuts out the first decoding pass so that we no longer need to rewind() the stream after it.

### DIFF
--- a/apps/openmw/mwsound/ffmpeg_decoder.cpp
+++ b/apps/openmw/mwsound/ffmpeg_decoder.cpp
@@ -389,16 +389,6 @@ void FFmpeg_Decoder::readAll(std::vector<char> &output)
     }
 }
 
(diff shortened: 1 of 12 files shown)


scrawl:
@kcat any objections to merging this?

kcat:
Looks okay to me from what I can see.

PR #982, opened 2016-06-29
Fix incorrect purging of constant-effect invisibility

MiroslavR:
Bug report: https://bugs.openmw.org/issues/3448

This PR fixes bugs that occur after breaking constant-effect invisibility:
 - effect icon not disappearing
 - effect reapplied after saving and reloading the game
 - effect reapplied after (un)equpping an item other than the one that gave the player invisibility

I'm not really sure whether https://github.com/OpenMW/openmw/blob/528de956da1508a51e0e9ec8af8dda9b9b3937f3/apps/openmw/mwmechanics/actors.cpp#L518 is affected by this change, though.

### DIFF
--- a/apps/openmw/mwworld/inventorystore.cpp
+++ b/apps/openmw/mwworld/inventorystore.cpp
@@ -722,7 +722,11 @@ void MWWorld::InventoryStore::rechargeItems(float duration)
 
 void MWWorld::InventoryStore::purgeEffect(short effectId)
 {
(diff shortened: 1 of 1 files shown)


scrawl:
> I'm not really sure whether https://github.com/OpenMW/openmw/blob/528de956da1508a51e0e9ec8af8dda9b9b3937f3/apps/openmw/mwmechanics/actors.cpp#L518 is affected by this change, though.

I don't see a problem with this line.

PR #983, opened 2016-06-30
Don't allow telekinesis on actors or teleport doors

Allofich:
Fix for Bug #2721.

The telekinesis check now does not allow using the extended range on actors or doors that lead to another cell.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -738,7 +738,9 @@ namespace MWRender
 
         mViewer->getCamera()->accept(*createIntersectionVisitor(intersector, ignorePlayer, ignoreActors));
 
-        return getIntersectionResult(intersector);
+        RayResult result = getIntersectionResult(intersector);
+        result.mDistanceToFirstIntersection = maxDistance * intersector->getFirstIntersection().ratio;
+        return result;
     }
 
     void RenderingManager::updatePtr(const MWWorld::Ptr &old, const MWWorld::Ptr &updated)
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -668,6 +668,7 @@ namespace MWRender
     {
         RenderingManager::RayResult result;
         result.mHit = false;
+        result.mRatio = 0;
         if (intersector->containsIntersections())
         {
             result.mHit = true;
@@ -675,6 +676,7 @@ namespace MWRender
 
             result.mHitPointWorld = intersection.getWorldIntersectPoint();
             result.mHitNormalWorld = intersection.getWorldIntersectNormal();
+            result.mRatio = intersection.ratio;
 
             PtrHolder* ptrHolder = NULL;
             for (osg::NodePath::const_iterator it = intersection.nodePath.begin(); it != intersection.nodePath.end(); ++it)
@@ -738,9 +740,7 @@ namespace MWRender
 
         mViewer->getCamera()->accept(*createIntersectionVisitor(intersector, ignorePlayer, ignoreActors));
 
-        RayResult result = getIntersectionResult(intersector);
-        result.mDistanceToFirstIntersection = maxDistance * intersector->getFirstIntersection().ratio;
-        return result;
+        return getIntersectionResult(intersector);
     }
 
     void RenderingManager::updatePtr(const MWWorld::Ptr &old, const MWWorld::Ptr &updated)
--- a/apps/openmw/mwworld/actiontrap.cpp
+++ b/apps/openmw/mwworld/actiontrap.cpp
@@ -7,20 +7,20 @@
 namespace MWWorld
 {
 
-    void ActionTrap::executeImp(const Ptr &actor)
+    void ActionTrap::executeImp(const Ptr &actor, float distance)
     {
         osg::Vec3f actorPosition(actor.getRefData().getPosition().asVec3());
         osg::Vec3f trapPosition(mTrapSource.getRefData().getPosition().asVec3());
         float activationDistance = MWBase::Environment::get().getWorld()->getMaxActivationDistance();
 
         // GUI calcs if object in activation distance include object and player geometry
-        const float fudgeFactor = 1.25f;
+        //const float fudgeFactor = 1.25f;
 
         // Hack: if actor is beyond activation range, then assume actor is using telekinesis
         // to open door/container.
         // Note, can't just detonate the trap at the trapped object's location and use the blast
         // radius, because for most trap spells this is 1 foot, much less than the activation distance.
-        if ((trapPosition - actorPosition).length() < (activationDistance * fudgeFactor))
+        if (distance < activationDistance)
         {
             // assume actor touched trap
             MWMechanics::CastSpell cast(mTrapSource, actor);
--- a/apps/openmw/mwworld/actiontrap.cpp
+++ b/apps/openmw/mwworld/actiontrap.cpp
@@ -7,36 +7,28 @@
 
 namespace MWWorld
 {
-    // actor activated object without telekinesis, so trap will hit
-    void ActionTrap::executeImp(const Ptr &actor) 
-    {
-        osg::Vec3f actorPosition(actor.getRefData().getPosition().asVec3());
-    
-        MWMechanics::CastSpell cast(mTrapSource, actor);
-        cast.mHitPosition = actorPosition;
-        cast.cast(mSpellId);
-
-        mTrapSource.getCellRef().setTrap("");
-    }
-
-    // actor activated object with telekinesis, so trap may or may not hit
     void ActionTrap::executeImp(const Ptr &actor, float distance)
     {
+        osg::Vec3f actorPosition(actor.getRefData().getPosition().asVec3());
         osg::Vec3f trapPosition(mTrapSource.getRefData().getPosition().asVec3());
-        float activationDistance = MWBase::Environment::get().getWorld()->getMaxActivationDistance();
+        float trapRange = MWBase::Environment::get().getWorld()->getMaxActivationDistance();
 
         // Note: can't just detonate the trap at the trapped object's location and use the blast
         // radius, because for most trap spells this is 1 foot, much less than the activation distance.
         // Using activation distance as the trap range.
 
-        if (distance < activationDistance) // actor activated object within range of trap
-            executeImp(actor);
-        else // actor activated object outside range of trap
+        if (distance > trapRange) // actor activated object outside range of trap
         {
             MWMechanics::CastSpell cast(mTrapSource, mTrapSource);
             cast.mHitPosition = trapPosition;
+            cast.cast(mSpellId);    
+        }
+        else // actor activated object within range of trap
+        {
+            MWMechanics::CastSpell cast(mTrapSource, actor);
+            cast.mHitPosition = actorPosition;
             cast.cast(mSpellId);
-            mTrapSource.getCellRef().setTrap("");
         }   
+        mTrapSource.getCellRef().setTrap("");
     }
 }
--- a/apps/openmw/mwworld/actiontrap.cpp
+++ b/apps/openmw/mwworld/actiontrap.cpp
@@ -1,6 +1,7 @@
 #include "actiontrap.hpp"
 
 #include "../mwmechanics/spellcasting.hpp"
+#include "../mwmechanics/actorutil.hpp"
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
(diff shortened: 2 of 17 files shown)


Allofich:
Looks like the original game doesn't allow using telekinesis on activators, either. I don't really like that, as if you are dexterous enough with telekinesis to open a chest and powerful enough to pick up a heavy piece of armor, it seems you should be able to, for example, turn a crank. I'm inclined to leave the use of activators in unless it causes a problem. 

We could also allow telekinesis for remote pickpocketing, as the bug reporter mentioned.

scrawl:
I would prefer a solution that doesn't call ```getFacedObject``` twice. It's a raycast that takes about 0.1-0.3ms depending on the scene.

If we extend ```getFacedObject``` to return the distance to the object hit, we could do it like this with a single call:
<pre>
object, distance = getFacedObject(getMaxActivationDistance() + telekinesisRangeBonus)
if object does not allow telekinesis and distance > getMaxActivationDistance():
   object = None
</pre>

Allofich:
OK, I'll take a look.

Do you have an opinion on allowing the use of activators?

scrawl:
> Looks like the original game doesn't allow using telekinesis on activators, either. I don't really like that, as if you are dexterous enough with telekinesis to open a chest and powerful enough to pick up a heavy piece of armor, it seems you should be able to, for example, turn a crank. I'm inclined to leave the use of activators in unless it causes a problem.

I don't have a strong opinion about this, as long as no one complains about their mods being broken by this change then we can leave it in I guess.

Speaking of activation distances, that reminds me of that bug with telekinesis on trapped containers. Having the distance to the faced object would allow for a proper fix there. We can pass the distance to the ```Action```, so that the ```ActionTrap``` can accurately decide whether or not the player is in telekinesis range.  Currently the check is not accurate (it's a plain distance check to the player, which doesn't take into account the geometry of the container, the camera position/orientation, etc.).

Allofich:
I made progress on this, but I can't figure out how to get the distance of the raycast to facedObject. I only know how to get a pointer to the facedObject and do a distance check between it and the player, like ActionTrap currently does. Is there a simple way to get the distance from castCameraToViewportRay? Also if that function casts from the "camera" it seems like you would get different results for first person and third person, but the current OpenMW implementation seems not have a problem with this, so I guess it's being accounted for somewhere?

Regarding telekinesis targets:

- I realized that there are some problems with allowing telekinesis on activators. It allows you to use beds from afar, and there are some quests in the game where you activate a shrine to read an inscription on it which doesn't really make sense with telekinesis. Other problems may come up as well so I decided for now to not allow activators.

- I decided for now to continue allowing pickpocketing of NPCs.

- I found out the original game doesn't allow telekinesis on book items or on lights (candles, etc.), either, but I decided to continue allowing these.

I have the telekinesis target conditions  done, it's just the distance calculation of the ray cast which I can't figure out. I took a guess at trying to use getEyePos() and the ray result's mHitPointWorld but getEyePos() seems to not be used anywhere and also seemed to make the game crash.

Allofich:
I'm not sure if it's the best method or not but I seem to have the distance through using 
`mRendering->getCameraPosition()` and `mHitPointWorld`.

scrawl:
> I have the telekinesis target conditions done, it's just the distance calculation of the ray cast which I can't figure out. I took a guess at trying to use getEyePos() and the ray result's mHitPointWorld but getEyePos() seems to not be used anywhere and also seemed to make the game crash.

getEyePos() was a leftover of some code that's no longer used. I've removed it.

> I'm not sure if it's the best method or not but I seem to have the distance through using 
mRendering->getCameraPosition() and mHitPointWorld from the ray result.

I guess that works, but a more elegant way might be to use the Intersection's ```ratio```. As far as I know, the ```ratio``` is the fraction of the maximum distance. I.e. a ratio of 1 would tell you the maximum distance was reached. So the distance should be maxDistance * intersection.ratio.

Allofich:
Here's a new implementation.

I decided to not allow telekinesis for pickpocketing after all. One reason is that I remembered that pickpocketing includes a check to look at the victim's inventory, which I assume is supposed to be like peeking into their bag or something, but I don't think of telekinesis as giving remote sight, so it seems to me you should not be able to peek into the inventory.

Another reason is that it keeps the game balance closer to the original, and prevents a potential situation of overpowered remote pickpocketing.

This means telekinesis should function the same with these changes as in the original engine, except that books and lights can also be used with it.

Traps are now using raycast-derived distance to check whether they hit the player. I'm using the maxactivation distance (192) as the trap range. In the original engine the trap range is about 170, although you are always hit if you activate without telekinesis on. That means there is a small margin between about 170 and 192 where whether you are hit by the trap or not depends only on whether telekinesis is active. I suppose we could recreate this behavior but I'm not sure that it's really desirable.

After changing the trap behavior, a test NPC using an Activate package would no longer be hit by a trap. They were activating at distance 200, which was seemingly just an arbitrary close number, so I changed it to use maxactivation distance so they will be hit by traps.

Allofich:
As a condition for using the raycast distance from `getFacedObject()` in the trap calculations, I put in a check that the object being activated by the player is actually the object that was returned by  `getFacedObject()` because I thought it was possible to use the console or scripts to activate any object, but when I tested in OpenMW I can't actually seem to do this, so this check might be unnecessary. If I removed it I could also remove the member variable mFacedObject that I added for it.

Allofich:
Updated by removing the check I mentioned in the comment just above, and otherwise simplifying the code.

I also discovered a minor issue. In the original engine, the range of traps is not the same as iMaxActivateDist (192), but varies by the object. From testing I got distances of around 140 to 170. This offers a small margin between  iMaxActivateDist and the trap range where the player can use telekinesis safely even though they are within normal activation range. This seems to be the only way to use telekinesis to safely open a trapped door that teleports to another cell, since if you are out of normal activation range you cannot use telekinesis on these doors, so there is a minor effect on gameplay. I decided to approximate the original engine's distance by taking an average of what I had observed until a better solution comes up.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1024,9 +1025,17 @@ namespace MWWorld
 
             float activationDistance = getMaxActivationDistance() + telekinesisRangeBonus;
 
-            facedObject = getFacedObject(activationDistance);
+            facedObject = getFacedObject(activationDistance, distanceToObject, true);
+
+            // Not allowing telekinesis on actors, on doors that teleport to other cells, or on activators
+            // Original engine doesn't allow telekinesis on books or lights, either
+            if (!facedObject.isEmpty() && (facedObject.getClass().isActor()
+                || facedObject.getCellRef().getTeleport() || facedObject.getClass().getTypeName() == "struct ESM::Activator")
```
MiroslavR: I think it would be better (and consistent with the rest of the code) to use `typeid(ESM::Activator).name()` here. That way if the struct name ever changes, we will get an error at compile time.

scrawl: Actually ```facedObject.getClass().getTypeName() == "struct ESM::Activator"``` would not work in the first place, the types are compiler-dependent. On Linux/GCC typeid names look like this (for an ESM::NPC for example):
<pre>
N3ESM3NPCE
</pre>

Allofich: OK, fixed it.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -738,7 +738,9 @@ namespace MWRender
 
         mViewer->getCamera()->accept(*createIntersectionVisitor(intersector, ignorePlayer, ignoreActors));
 
-        return getIntersectionResult(intersector);
+        RayResult result = getIntersectionResult(intersector);
+        result.mDistanceToFirstIntersection = maxDistance * intersector->getFirstIntersection().ratio;
```
scrawl: mDistanceToFirstIntersection is uninitialized when using the castRay() method. Even if it's not used right now, never a good idea to have uninitialized values flying around.

```getIntersectionResult``` returns the ```RayResult```, it should be initialized there. We can change it to ```result.mRatio``` if that makes it easier.

File: apps/openmw/mwworld/worldimp.cpp [hunk 3]
```diff
@@ -1024,9 +1025,17 @@ namespace MWWorld
 
             float activationDistance = getMaxActivationDistance() + telekinesisRangeBonus;
 
-            facedObject = getFacedObject(activationDistance);
+            facedObject = getFacedObject(activationDistance, distanceToObject, true);
+
+            // Not allowing telekinesis on actors, on doors that teleport to other cells, or on activators
+            // Original engine doesn't allow telekinesis on books or lights, either
+            if (!facedObject.isEmpty() && (facedObject.getClass().isActor()
+                || facedObject.getCellRef().getTeleport() || facedObject.getClass().getTypeName() == typeid(ESM::Activator).name())
```
scrawl: Perhaps it would be cleanest to move this logic to a new method ```allowTelekinesis``` in MWWorld::Class, implemented in classes like NPC, Activator and Door. That would allow us to get rid of the type check.

apps/openmw/mwworld/actiontrap.cpp
```diff
@@ -1,40 +1,45 @@
...
         {
-            // assume telekinesis used
             MWMechanics::CastSpell cast(mTrapSource, mTrapSource);
             cast.mHitPosition = trapPosition;
             cast.cast(mSpellId);
-        }
-        mTrapSource.getCellRef().setTrap("");
+            mTrapSource.getCellRef().setTrap("");
+        }   
     }
```
scrawl: Looks like there is some code duplication here. I would prefer a single method for both cases. You could treat a distance of -1 (or generally < 0) as "distance unavailable". Then we could have a single method executeImp(const MWWorld::Ptr& actor, float distance=-1).

As for whether or not telekinesis was used, you can check that inside the ```ActionTrap``` by observing the actor's magic effects. There is no need to call different methods of the ```Action``` depending on telekinesis or no telekinesis as the base ```Action``` itself has nothing to do with that. And it would be quite a mess if we added more states at some point (now you have 2 execute methods for telekinesis and no telekinesis, add another boolean state and you have 4, then 8, and so on).

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1011,10 +1011,11 @@ namespace MWWorld
     MWWorld::Ptr World::getFacedObject()
     {
         MWWorld::Ptr facedObject;
+        float distanceToObject;
```
scrawl: Is this extra variable needed? Why not write the distance directly to ```mDistanceToFacedObject```?

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1024,9 +1025,17 @@ namespace MWWorld
...
+
+            // Not allowing telekinesis on actors, on doors that teleport to other cells, or on activators
+            // Original engine doesn't allow telekinesis on books or lights, either
+            if (!facedObject.isEmpty() && (facedObject.getClass().isActor()
+                || facedObject.getCellRef().getTeleport() || facedObject.getClass().getTypeName() == typeid(ESM::Activator).name())
+                && (distanceToObject > getMaxActivationDistance()))
+                return 0;
         }
 
+        mDistanceToFacedObject = distanceToObject;
```
scrawl: If no object was hit, an uninitialized value is assigned to mDistanceToFacedObject.

apps/openmw/mwrender/renderingmanager.cpp
Allofich: Fixed.

apps/openmw/mwworld/worldimp.cpp
Allofich: Fixed.

Allofich:
New version uploaded.

I decided for now at least to make the distance of traps the same as activation range after all. In order to allow the player to use telekinesis to unlock and disarm trapped teleport doors, I needed the distance to the facedobject in the `door` class when deciding whether to let the player teleport. I decided to change my previous implementation and just use a "get" method to read `mDistanceToFacedObject` as needed for `door` and `actiontrap,` instead of passing it down through `action`. This let me revert the changes to the `execute()` and `executeimp()` methods, which I was struggling to keep from being messy.

It's a lot of commits and includes them going back and forth on each other, but I assume that's OK, right?

Allofich:
Making a minor cleanup.

MiroslavR:
This seems to have broken **third-person** interaction with certain objects. Try interacting with NPCs or doors in Seyda Neen.

scrawl:
> This seems to have broken third-person interaction with certain objects. Try interacting with NPCs or doors in Seyda Neen.

[Fixed](https://github.com/OpenMW/openmw/commit/67d3571faca5b19c4b09195290774b90b2c08b05).

PR #985, opened 2016-07-02
Show remnant corprus effects in magic window

MiroslavR:
Display magic effects that remain after curing corprus to the user. Vanilla doesn't show these effects, however I think this is a useful improvement.

### DIFF
--- a/apps/openmw/mwmechanics/spells.cpp
+++ b/apps/openmw/mwmechanics/spells.cpp
@@ -249,6 +249,10 @@ namespace MWMechanics
 
     void Spells::visitEffectSources(EffectSourceVisitor &visitor) const
     {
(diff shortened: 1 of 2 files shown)


kcat:
That extra ```std::map``` is going to cause quite a few micro-allocations as it goes over the magic effects, since it needs to allocate a new pair for each one being added (which will end up getting deallocated at the end of the function).

MiroslavR:
> That extra std::map is going to cause quite a few micro-allocations as it goes over the magic effects, since it needs to allocate a new pair for each one being added (which will end up getting deallocated at the end of the function).

That's a valid point. This should be much less of a problem now that magic effects are only rebuilt when necessary instead of every frame.

PR #987, opened 2016-07-06
Android crash fixes

sandstranger:
In this PR was fixed engine crash on old tegra and powerv devices .
Also it is impossible in this case not using jni,because for example I am using glshim , so glGetString(GL_EXTENSIONS) always returns fake opengl extensions
Java part of this PR
https://github.com/OpenMW/android-port/commit/36f96195bef01f269c564d08cbdd2cf7bb651af4


scrawl:
As far as I know, a depth buffer size of 16 or lower results in severe depth fighting artifacts for Morrowind. That's why we added the ``` SDL_GL_SetAttribute(SDL_GL_DEPTH_SIZE, 24);``` in the first place. If a device doesn't support that, then I guess the device should be declared unsupported for running OpenMW.

sandstranger:
Ok , than I will add this info to the Android wiki

PR #990, opened 2016-07-08
Key configuration and shortcuts

Aesylwinn:
I'm making this pull request now so that I can get some feedback on the general approach I'm taking. There are probably some unused headers, and I commented out a couple lines of code. Please ignore this for now, I'll polish it up later. :)

Key configuration can be modified in the settings menu (text editting at the moment, I'm planning on some form of key detection being used later).
Some example configurations that should work:
Key_A
Key_A,Key_F
ControlModifier+Key_A
Mouse5
Key_A,Mouse2

When I first looked into this, QShortcut looked like the place to start. However, I soon discovered that it does not support mouse buttons at all. At that point, I decided to try and implement my own solution. I took advantage of how the [Qt Mouse enum](url=http://doc.qt.io/qt-4.8/qt.html#MouseButton-enum) conflicts with the [Qt Key enum](url=http://doc.qt.io/qt-4.8/qt.html#Key-enum) so that I could keep using QKeySequence. From what I can tell, my implementation works (I do need to make an input manager at the widget level for activating potentially ambiguous key sequences).

Some benefits:
1. Mouse buttons can be used (1-5 are supported by qt) for anything
2. Multiple keys/buttons can be used for anything
3. It provides an alternative to using event filters to pass the input to any widget that may have shortcuts (a problem likely to arise with the scene widget and toolbar implementation)

Some downsides
1. Added complexity, classes must listen for events instead of keys
2. Requires the widget to pass input events to the shortcuts (eventually an input manager).
3. I haven't fully worked out how mouse tracking will happen, so that may cause some issues.


Aesylwinn:
I should mention that currently only the 1st/free camera is using it.

Aesylwinn:
I didn't realize the erase function didn't return an iterator until c++11. :/ On that subject, gcc is printing deprecation warnings about the use of std::auto_ptr now.

zinnschlag:
> 1. Mouse buttons can be used (1-5 are supported by qt) for anything.

Neat. I thought that would be pretty hard to do. Maybe I should finally invest into a 5 button mouse.

I like the overall approach. Unifying keyboard and mouse input seems reasonable.

Aesylwinn:
Cool. I'll continue with this approach then.

Aesylwinn:
@zinnschlag I'm going to be making quite a few changes to the scene and worldspace widgets to make them use my shortcuts. Would it be alright if I cleaned up some of the code/formatting there as well? It's starting to look a bit haggard with all of the changes it has undergone.

zinnschlag:
If you have to change the code anyway, then you can as well clean it up, including the code in the immediate vicinity.

Aesylwinn:
Zini, I changed the cell marker interaction so that it uses "primary/secondary select" instead of "shift + primary/secondary edit" for some of the functionality. Is that fine?

zinnschlag:
I am not sure what you mean. The shift key means accelerated movement. In what context was it used with the cell markers? There shouldn't have been any shift involved.

zinnschlag:
Or did you mean the cell arrows? These use shift IIRC, because we needed an additional modifier key.

Aesylwinn:
I was referring to the cell arrows, sorry. The arrows have 4 different actions you can take with them. I'm not sure of why they use the shift modifier instead of using the selection buttons.

apps/opencs/view/render/pagedworldspacewidget.cpp
```diff
@@ -142,75 +142,71 @@ void CSVRender::PagedWorldspaceWidget::addEditModeSelectorButtons (
...
 
-                bool modified = false;
+            bool modified = false;
 
-                if (shift)
-                {
-                    if (button=="p-edit")
-                        addCellSelection (x, y);
-                    else
-                        moveCellSelection (x, y);
```
Aesylwinn: Here's the relevant bit.

Aesylwinn:
Let me know if you want me to change it back to using shift, the last commit would allow it.

zinnschlag:
I see. What is the reason for this change, i.e. why is it an improvement over what we have before? IMO both variants are equally inconsistent with the rest of the input scheme. That is unavoidable unfortunately.

Aesylwinn:
The only improvement is that it uses a common input binding.

Aesylwinn:
It's already using the "edit" buttons to load in new cells.

Aesylwinn:
@zinnschlag It looks like [SpellSword14](https://forum.openmw.org/viewtopic.php?f=22&t=3675) is interested in working on the editor and is looking for some starting tasks.

zinnschlag:
Actually, I do not have a strong opinion regarding the cell arrow input scheme. Use whatever you think works best.

Aesylwinn:
I haven't fully tested it yet, but I don't believe there should be any more design changes for shortcut handling. Using a shortcut can be as simple as:
`CSMPrefs::Shortcut* saveShortcut = new CSMPrefs::Shortcut("document-save", this);
connect (saveShortcut, SIGNAL(activated()), this, SLOT(save()));`
Getting the widget child-parent relationships to properly update was the major slowdown. :/

The next steps are to add/replace shortcuts everywhere and make a clean interface for viewing/editing the shortcut keybindings.

Aesylwinn:
I was wondering why it seemed like I was always seeing a signed/unsigned warning when getting QSequence::count(). It turns out that this returns an unsigned int in Qt4 and a signed int in Qt5.

Aesylwinn:
I'm not sure of how I should be changing the toolbar descriptions to mention the specific key bindings. The current first person description:

> First Person
> Mouse-Look while holding the left button
> WASD movement keys
> Mouse wheel moves the camera forward/backward
> Strafing (also vertically) by holding the left mouse button and control
> Camera is held upright
> Hold shift to speed up movement

I could do a direct replacement for which keys are used, but that may get clunky with more complex key bindings. A mock-up (the actual text for keys may be different):

> First Person
> Mouse-Look while holding Left Mouse Button
> Movement keys: W(up), A(left), S(down), D(right)
> Mouse wheel moves the camera forward/backward
> Strafing (also vertically) by holding Control + Left Mouse Button
> Camera is held upright
> Hold Shift to speed up movement

I could also just tell the user to view the preferences page for the controls.

> First Person
> The camera is held upright and operates similar to an in game first person camera.
> See the settings page for key-bindings.

Any thoughts or opinions?

Aesylwinn:
One downside to listing the keys is that there is currently no way to modify the tooltip text if they change.

zinnschlag:
We want to have the keys explicitly listed in the tooltips, because that will be the main way how the user learns about them.
I am not sure about the implementation path. Maybe the tooltip text can have placeholders (e.g. the name of the key user setting) in a special escape sequence. The text would then have to be processed before showing it. Not sure if Qt offers any kind of hook that would allow for that. If everything else fails you can always modify the toolbar classes.

Aesylwinn:
I think the escape sequence is a good idea. There is a ToolTip event, so I should be able to listen for that event and process the string before it displays.

Aesylwinn:
Well, it wasn't as simple as I though it might be. The ModeButton class regenerates its tool tip whenever settings are modified. The SceneToolMode regenerates its tool tip every time it gets shown (but it doesn't have to parse the key bindings). I think it should be efficient enough for the moment.

Aesylwinn:
I've been thinking about how shortcut text should appear to the user. I think it should be short but clear. For mouse buttons, I was thinking it would be fine to show LMB, RMB, MMB, Mouse4, and Mouse5. For Keys, the actual name of the key should suffice, though modifiers should use the standard shorthand.

Shift+LMB
Ctl+A
Mouse5+LMB
Mouse4+S

Do you think shortening the mouse names will be an issue? 

Also, while I'm thinking of it, should everything in the top menu bar (save, undo, etc. menus) be assignable as a shortcut? I could leave some of them blank to start with, but still allow them to be configured.

zinnschlag:
Mouse button names like LMB are well established in other software by now. I would expect most users to be familiar with them. So no problem here.

Ideally we want every single function to be accessible via a keyboard shortcut. But leaving some shortcuts blank in the default settings is fine.

Aesylwinn:
I tested to make sure it still runs on Qt4. It does.

I need to add some shortcuts for various context menus, and I want to make expandable sections in the key bindings page. Hopefully I can finish that up pretty quickly. Then I need to take a couple days break before testing and reviewing my code.

Out of curiosity, why is there a "close" and an "exit" option in the file menu? They appear to do the same thing.

zinnschlag:
Close closes the document. Exit exits the application.

Aesylwinn:
I apologize if this is difficult to read through. My thoughts are pretty fragmented at the moment.

> Close closes the document. Exit exits the application.
I see. Perhaps "Close" should instead be called "Close Document" to prevent confusion.

I made the custom settings page for key bindings. I'm pretty happy with it. However, it has one slight problem: the QScrollArea isn't shrinking properly. It doesn't affect usability, but it would be nice to fix that. Any idea on how to accomplish that?

I'm thinking I should remove Qt's window navigation using the alt key since that conflicts and causes issues if you try to configure the alt modifier. The menu system appears to work outside of Qt’s event system, so I don’t think there is a good way for me to work around it.

I had mentioned using Qt's Standard Key for some defaults, however I think that may be a bad idea since it could potentially cause conflicts in the default configuration. I think I’ll post something on the forums to see if more defaults are wanted.

Aesylwinn:
Just a heads up, I’ll have family visiting next week which means I’ll probably have very little time to work on this feature. As I mentioned previously, I’ll be taking a small break for a day or two, so if this pull request is going to be merged relatively soon, I’ll have to make any requested changes over the weekend. I don’t know how that meshes with your schedule, so maybe it’s not possible. Anyway, almost all of the requested features were implemented. The only exception I’m aware of is the conflict detection. If I don’t have too many changes to make, I might be able to add that during the weekend, but you did mention in a forum post that it was optional.

zinnschlag:
I would rather finish the task up this weekend without the conflict detection than to delay it. There are indeed problems with the schedule (someone else going away for a while).

Anyway, just gave it a test drive. Nicely done.

I don't know what you mean regarding the scroll area. I do not see any problems here.

I am okay with some renaming of menu items (close). I just noticed a few other issues. "Load error log" is a bit confusing. Maybe it should be "Open load error log". Same for "Run Log". And "Show Status Bar" is really a toggle. Maybe "Toggle status bar" instead.
I also noticed that we have some inconsistency in capitalisation. Not quite sure what the right option would be, but inconsistency certainly isn't.

btw. there is a mistake in the label of Show Texture Assets. And "Show" is not the greatest choice of words for opening a subview. We could make it "Open" instead, but I think the wording in the menu (just drop the show/open entirely) is okay for the user settings too.

Aesylwinn:
>  I don't know what you mean regarding the scroll area. I do not see any problems here.

When I switch the key binding "subcategory", the scroll area stays the same size instead of shrinking to fit its contents. Maybe it's a platform dependent thing.

> We could make it "Open" instead, but I think the wording in the menu (just drop the show/open entirely) is okay for the user settings too.

Maybe "Open Texture Assets View" would be a good pattern? In the menu, you can get some context depending on what menu you have open. In the key bindings page, you don't have that context.

It's kind of funny how one of the current descriptions is "Show Body Parts".

zinnschlag:
> When I switch the key binding "subcategory", the scroll area stays the same size instead of shrinking to fit its contents. Maybe it's a platform dependent thing.

Okay. I see what you mean. Not a big problem. Definitely not worth delaying the next release over.

> Maybe "Open Texture Assets View" would be a good pattern?

Our OpenMW-CS UI terminology is still a bit shaky. Normally we call that a Subview. But that might end being just a term used internally. Maybe replace the word view with list where applicable? That is even more clear and bypasses all terminology problems.

Aesylwinn:
Zini, a couple of people on the forum are mentioning that they would like the default scene edit/select to match Blender's configuration. Do you have an opinion on that?

Otherwise, I think this is ready for review.

zinnschlag:
> Zini, a couple of people on the forum are mentioning that they would like the default scene edit/select to match Blender's configuration. Do you have an opinion on that?

As long as it fits into our systems of camera modes, I don't mind.

Aesylwinn:
Since I changed some of the defaults, you may want to clear out the "Key Bindings" section of your openmw-cs.config file.

zinnschlag:
Problem: The fast mode shortcut is colliding with other shortcuts. Shift is used for fast movement, but shift is also used for secondary mouse functions. Furthermore the fast shortcut should be global, not camera specific. We may at one point or another use it in editing functions. It can not be used anywhere else in the 3D scene.

Aesylwinn:
Whoops, I generally look before I move, so my quick test didn't catch that. 

When you say global, do you mean global to the scene, or global to the application?

Aesylwinn:
I'll have to make some changes to allow shared modifiers. I'll place the option in the scene page, I guess.

zinnschlag:
> When you say global, do you mean global to the scene, or global to the application?

There should be only a single key for fast movement for all editing functions in the 3D scene.

Aesylwinn:
I changed the way the settings are stored, so be sure to clear out the "Key Binding" section.

Aesylwinn:
The conflict detection is definitely not happening this weekend. My family has arrived. Let me know if you have any other problems.

PR #991, opened 2016-07-09
Implement Face instruction (Feature #1424)

MiroslavR:
Feature: https://bugs.openmw.org/issues/1424


scrawl:
Bug: when ```face``` is invoked on idling actors, they keep playing their idle animation while turning instead of the turning animation (easier to see if you reduce MAX_VEL_ANGULAR_RADIANS to a lower number).

MiroslavR:
> Bug: when face is invoked on idling actors, they keep playing their idle animation while turning instead of the turning animation (easier to see if you reduce MAX_VEL_ANGULAR_RADIANS to a lower number).

This seems to be a general bug present only in first person. Should be fixed.

PR #993, opened 2016-07-09
Pathfinding unification

mrcheko:
These commits make some pathfinding techniques from aicombat package the same for all other ai packages: this gives improved shortcut check (with up-down ray) as well as some other lesser improvements. All aicombat pathfinding code was moved and adapted to aipackage which simplified combat code a lot, some logic was moved to ai combat action.
aipackage::pathTo function was redesigned to fully guide actor to it's target location and should be called each frame.

These commits don't address open pathfinding issues. Specific packages features were retained (AiWander doesn't use shortcuts etc).


apps/openmw/mwmechanics/aiactivate.cpp
```diff
@@ -23,33 +23,29 @@ namespace MWMechanics
...
 
-        //Set the target desition from the actor
+        //Set the target destination for the actor
         ESM::Pathgrid::Point dest = target.getRefData().getPosition().pos;
 
-        if(distance(dest, pos.pos[0], pos.pos[1], pos.pos[2]) < 200) { //Stop when you get close
-            actor.getClass().getMovementSettings(actor).mPosition[1] = 0;
+        if (pathTo(actor, dest, duration, 200)) //Go to the destination
+        {
+            // activate when reached
```
Allofich: Just a minor thing, but I recently changed the `200` here to `MWBase::Environment::get().getWorld()->getMaxActivationDistance())` in my PR on telekinesis, but I see you are still using 200. It's not a big deal at the moment because AI is set to always trigger a trap if they activate a trapped object anyway, but this was briefly an issue for me when I was working on the code and AI was activating objects outside of the max activation distance (192) and therefore not being hit by traps. It seems to me we might as well use maxactivationdistance or less here anyway, since I don't see why AI actors should be able to activate objects from farther away than the player can.

Allofich: Actually, from some quick testing in-game, it looks like an NPC activates an object from closer than the player's maximum activation distance even if 200 is used here. I assume this may be related to the positioning of the camera, from which the player's activation range is calculated. Anyway, since it's not an issue for activating traps at the moment, so you can ignore my above comment.

mrcheko: Thx for the note. This was exactly the reason of the merge conflict resolved in f57858b.

scrawl:
In resolving conflicts it looks like you have not incorporated changes from commit 8fcc1911d98dbd4d041418591592cad2c7998ca5 .

mrcheko:
This should fix it, I have also made touch-spells use fCombatDistance. Everything else regarding 8fcc191 went in https://github.com/OpenMW/openmw/pull/993/files#diff-cf6fedf51bb2e00a0470e343b1f5e096R486

Allofich:
I'm seeing a couple instances of odd behavior when an NPC pathfinds along the pathgrid with a travel or escort package.

1. Every once in a while the NPC will walk a bit off from a node then retreat back to that node before continuing on. In one case the NPC even went back and forth like this two or three times.
2. After finishing the second of an editor-placed set of two packages (either 2 travels or 2 escorts) the NPC is reporting that the AI package is done several times in a row instead of just once and shaking in place for a moment.

Allofich:
To reproduce the above problems:

For 1
1. start OpenMW and in the console type "coc ToddTest"
2. get rid of the enemies, disable the door in front of the ordinators and two of the three ordinators
3. type "tpg" so you can see the pathgrid
4. select the remaining ordinator and give him "aitravel 550 1350 -530". This travel package should go OK.
5. next type "aitravel 2940 380 -600" and observe that towards the end he will go back and forth to nodes several times before arriving at the destination.

For 2
1. Create an NPC in the construction set with multiple travel or escort packages.
2. Give the NPC a script with
Begin scriptname
if (GetAIPackageDone)
	MessageBox, "Finished an AI Package."
endif
End scriptname
3. Run in game and observe that at the end of the final package the "Finished an AI Package." message gets spammed. This won't happen in the original game or in the current git master.

mrcheko:
Thx for the detailed description. I will check this later.

mrcheko:
1. There were some redundant checks before path shortcut. Observed behaviour actually was caused  by the destination point above the ground. Anyway those checks were inaccurate at least.
2. This was caused by my wrong assumption in https://github.com/OpenMW/openmw/commit/bce66c629ab54c47b151dc951657c713a157e63b#commitcomment-18188061: ai packages don't have copy constructor implementations so package's pathfinding state is not reset in `clone()`. And I had to reapply code that was trimmed in c57d177f1c804104a61b229b414f41f09d5676df (but in different manner :))

Allofich:
There was also a problem, BTW, of AI activate packages not rebuilding a path. Since the activate package ends when the actor gets within range and uses the object, and not when they reach the last pathgrid point, the final point in the path (located at the object to be activated) would remain even after the package ended. When the package came up again on a repeat, since it still had this leftover point it would just use it instead of rebuilding the path, which means it would walk in a straight line without pathfinding on the grid. This should hopefully be fixed now that you are clearing the path when a package finishes.

Allofich:
I tested out the new code. Looks like the problems I mentioned, including the one about activate packages, are fixed, nice work! I did notice one minor issue of the NPCs returning back the way they came to a pathgrid point after bumping into an obstacle while they were, I assume, attempting to take a shortcut between the points. This is again while using the ToddTest cell. Ideally I think we would want that to be fine-tuned to walking around the obstacle instead of reversing back to the pathgrid, maybe as part of overall improvements to obstacle avoidance, but I think that could be future work instead of this PR.

mrcheko:
Yes, pathfinding still not 100% smooth (in current master also) and has more problems to be solved.
But the purpose of this pull request was to move aicombat specific pathfinding enhancements to other ai packages.

And about the situtation you described: obstacle avoiding system is far from perfect at the moment and I believe it can be greatly improved. But anyway IMO you should consider using it only if the pathgrid is not available cause it [pathgrid] gives more reliable info for AI. Pathgrid is the only hand-made data we have for AI pathfinding after all. But this all is a bit offtopic here.

Aesylwinn:
@scrawl It looks like this may be getting missed. I believe mrchecko considers his goal achieved.

scrawl:
Thanks for the reminder, I forgot. I will test it this evening.

scrawl:
This PR breaks Guar attacks.

scrawl:
Just taking a guess here, maybe the changes in 345335309193b865a3c01b27a1e42b219a099dc3 were not merged correctly.

mrcheko:
@scrawl yes, you were right. Fixed it

scrawl:
Vertical aiming does not work anymore (try Guard vs. Rat or Cliffracer vs. Player)

mrcheko:
fixed

scrawl:
Not sure if related to the PR, but I noticed that Netch sometimes have issues aiming their spell. Shouldn't vertical aiming apply instantly, instead of using the smoothTurn function? The point of smoothTurn is to make the animation smooth, as turning on the Z axis actually rotates the creature physically, but that isn't needed for the X axis (which only affects aiming, not the creature itself).

mrcheko:
Are you talking about netch aiming while player is moving? `AimDirToMovingTarget`  isn't 100% accurate, also it still does not use `fTargetSpellMaxSpeed` GMST. I have a plan to repair that but after some work on pathfinding.

> but that isn't needed for the X axis

I actually think if we rotate smoothly by Z it's also logically that we do the same by X. Either way rotations rather fast so I don't see which problems it can cause.

scrawl:
> I actually think if we rotate smoothly by Z it's also logically that we do the same by X. Either way rotations rather fast so I don't see which problems it can cause.

Well, the netch_bull's spell casting animation apparently casts on the first frame. So the aiming has no time to adjust at all.

mrcheko:
No, aiming happens while target is in attack range not when an attacking animation starts.

scrawl:
I made a video: https://www.youtube.com/watch?v=zUPOY1jPOq8

Problem 1: creature doesn't start aiming until within attack range. Once within attack range, the attack fires immediately, giving no time to aim.

Problem 2: if the creature was looking away from the player before being aggro'd, projectile fires in completely wrong direction.

I suspect that problem 1) is partly caused by the fact that movement and aiming use the same vector. When the "move towards player" and "aim towards player" vectors are not the same, a conflict occurs. The vectors might not be the same because moving uses the creature position as a basis, but aiming uses the center of collision height.

How about not starting the attack until the desired aim vector is reached?

mrcheko:
thx for a video.
> How about not starting the attack until the desired aim vector is reached?

This is definitely the way to go. But also this is definitely not related to the PR :) I think this should be tracked on bugtracker for now. There are some other broken things in aiCombat so there will be another PR after a while.

scrawl:
Please take a look at [this](https://bugs.openmw.org/issues/3495#change-16064) regression. Looks like the spellcasting AI is generating a NaN target vector or something like that. No problem before this PR.

 <pre>
coc Addamasartus
placeatpc, itermerel, 1,0,0
itermerel->aifollow, player, 0,0,0,0
</pre>

scrawl:
clang warning:
<pre>
In file included from /home/travis/build/OpenMW/openmw/apps/openmw/mwscript/aiextensions.cpp:20:
/home/travis/build/OpenMW/openmw/apps/openmw/mwscript/../mwmechanics/aitravel.hpp:38:26: warning: 
      'MWMechanics::AiTravel::doesPathNeedRecalc' hides overloaded virtual
      function [-Woverloaded-virtual]
            virtual bool doesPathNeedRecalc(ESM::Pathgrid::Point dest, c...
                         ^
/home/travis/build/OpenMW/openmw/apps/openmw/mwscript/../mwmechanics/aipackage.hpp:114:26: note: 
      hidden overloaded virtual function
      'MWMechanics::AiPackage::doesPathNeedRecalc' declared here: different
      number of parameters (1 vs 2)
            virtual bool doesPathNeedRecalc(const ESM::Pathgrid::Point& ...
</pre>

scrawl:
AIWander regression:
- wait for NPC to start wandering
- enter building
- exit building
- NPC is now stuck in place

scrawl:
I am reverting this PR for now until the regressions are worked out.

mrcheko:
OK, correct me if I'm wrong: I'm going to merge master in this branch, revert your revert (3732979), continue to work and then open new PR?

scrawl:
Either that, or rebase onto the revert commit.

PR #995, opened 2016-07-10
Trap and lock fixes

Allofich:
Some fixes to make OpenMW behavior the same as original engine behavior.

1. The trap activation sound was playing on the actor, not the trap. Fixed this and removed resulting unused parameter in `actiontrap.`
2. Doors and containers that are disarmed with a key now play the disarm trap sound.
3. Lockpick and probe sounds are now 3d and play on the picked/probed object.
4. Keys can be used to disarm trapped objects of lock level 0 now. This produces a message "-- key used to open lock" which is a little odd maybe but the same message as the original game. I at first was doing an implementation that disarmed the trap without any message, but I thought that for gameplay purposes it is good to know what key was used, so I copied the original engine behavior.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1305,7 +1305,9 @@ bool CharacterController::updateWeaponState()
                 if(!resultMessage.empty())
                     MWBase::Environment::get().getWindowManager()->messageBox(resultMessage);
                 if(!resultSound.empty())
-                    MWBase::Environment::get().getSoundManager()->playSound(resultSound, 1.0f, 1.0f);
+                    MWBase::Environment::get().getSoundManager()->playSound3D(target,
+                    resultSound, 1.0f, 1.0f, MWBase::SoundManager::Play_TypeSfx,
+                    MWBase::SoundManager::Play_Normal);
             }
             else if (ammunition)
             {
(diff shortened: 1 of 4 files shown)


apps/openmw/mwmechanics/character.cpp
```diff
@@ -1305,7 +1305,9 @@ bool CharacterController::updateWeaponState()
                 if(!resultMessage.empty())
                     MWBase::Environment::get().getWindowManager()->messageBox(resultMessage);
                 if(!resultSound.empty())
-                    MWBase::Environment::get().getSoundManager()->playSound(resultSound, 1.0f, 1.0f);
+                    MWBase::Environment::get().getSoundManager()->playSound3D(target,
+                    resultSound, 1.0f, 1.0f, MWBase::SoundManager::Play_TypeSfx,
+                    MWBase::SoundManager::Play_Normal);
```
scrawl: Note, sounds coming from the player are played in [2D by default](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwsound/soundmanagerimp.cpp#L554), even when you call playSound3D. They can't be played in 3D because we don't supply an accurate position of the sound source (which makes all the more difference when it's so close to the listener/camera).

So in practice, I don't think this change makes a difference, but there's no harm either.

PR #996, opened 2016-07-10
Fix third-person activation range

Allofich:
Fix a mistake in my previous PR on telekinesis that caused third-person activation range to be miscalculated. The camera distance of third person was being erroneously added to the distance to the faced object.


Allofich:
Thanks to MiroslavR for pointing out the problem.

Allofich:
Looks like I was too slow! scrawl fixed it already.

PR #998, opened 2016-07-11
[Incomplete (ish)] Make OpenMW discount '#' characters that do not start a line as comment delimiters

AnyOldName3:
Before this change, OpenMW-Launcher would only consider text in openmw.cfg to be a comment if its line started with a '#', whereas the Boost library parsing this file would regard hash characters mid-line as the start of a comment. This caused https://bugs.openmw.org/issues/3045 when content files included a hash in their filename, despite the launcher implying that everything was fine.

A fix for this has been implemented by:
* Wrapping the input stream used to read the file with a filter which escapes '#' characters that do not start a line.
* Wrapping the objects which boost::program_options store settings to once they have been read so that they automatically undo the escaping process and receive input as intended.

Unfortunately, C++ isn't automatically doing something which (I think) Java would, so this doesn't compile, although I don't think it's too far off. There are also a couple of pieces of code which may need names changed or to be moved to a new file.

### DIFF
--- a/components/files/configurationmanager.hpp
+++ b/components/files/configurationmanager.hpp
@@ -84,8 +84,8 @@ struct escape_hash_filter : public boost::iostreams::input_filter
 		std::queue<int> mNext;
 		int mPrevious;
 
-		bool seenNonWhitespace = false;
-		bool finishLine = false;
+		bool mSeenNonWhitespace;
+		bool mFinishLine;
 };
 
 /**
--- a/components/files/configurationmanager.hpp
+++ b/components/files/configurationmanager.hpp
@@ -93,6 +93,8 @@ struct escape_hash_filter : public boost::iostreams::input_filter
  */
 class EscapeHashString : public std::string
 {
+	private:
+		std::string mData;
 	public:
 		static std::string processString(const std::string & str);
 		static std::vector<std::string> toStdStringVector(const std::vector<EscapeHashString> & vec);
--- a/components/files/configurationmanager.hpp
+++ b/components/files/configurationmanager.hpp
@@ -107,6 +107,7 @@ class EscapeHashString : public std::string
 		EscapeHashString(InputIterator first, InputIterator last);
 
 		std::string toStdString() const;
+		std::string * toStdStringPtr() const;
 };
 
 } /* namespace Cfg */
--- a/components/files/configurationmanager.hpp
+++ b/components/files/configurationmanager.hpp
@@ -95,6 +95,7 @@ class EscapeHashString : public std::string
 {
 	public:
 		static std::string processString(const std::string & str);
+		static std::vector<std::string> toStdStringVector(const std::vector<EscapeHashString> & vec);
 
 		EscapeHashString();
 		EscapeHashString(const std::string & str);
@@ -104,6 +105,8 @@ class EscapeHashString : public std::string
 		EscapeHashString(size_t n, char c);
 		template <class InputIterator>
 		EscapeHashString(InputIterator first, InputIterator last);
+
+		std::string toStdString() const;
 };
 
 } /* namespace Cfg */
--- a/components/files/configurationmanager.hpp
+++ b/components/files/configurationmanager.hpp
@@ -91,13 +91,12 @@ struct escape_hash_filter : public boost::iostreams::input_filter
 /**
  * \class EscapeHashString
  */
-class EscapeHashString : public std::string
+class EscapeHashString
 {
     private:
         std::string mData;
     public:
         static std::string processString(const std::string & str);
-        static std::vector<std::string> toStdStringVector(const std::vector<EscapeHashString> & vec);
 
         EscapeHashString();
         EscapeHashString(const std::string & str);
@@ -109,7 +108,20 @@ class EscapeHashString : public std::string
         EscapeHashString(InputIterator first, InputIterator last);
 
         std::string toStdString() const;
-        std::string * toStdStringPtr() const;
+
+        friend std::ostream & operator<< (std::ostream & os, const EscapeHashString & eHS);
+};
+
+std::istream & operator>> (std::istream & is, EscapeHashString & eHS);
+
+struct EscapeStringVector
+{
+    std::vector<Files::EscapeHashString> mVector;
+
+    EscapeStringVector();
+    virtual ~EscapeStringVector();
+
+    std::vector<std::string> toStdStringVector() const;
 };
 
 } /* namespace Cfg */
--- a/components/files/configurationmanager.hpp
+++ b/components/files/configurationmanager.hpp
@@ -2,8 +2,10 @@
 #define COMPONENTS_FILES_CONFIGURATIONMANAGER_HPP
 
 #include <map>
+#include <queue>
 
 #include <boost/program_options.hpp>
+#include <boost/iostreams/filtering_stream.hpp>
 
 #include <components/files/fixedpath.hpp>
 #include <components/files/collections.hpp>
@@ -61,6 +63,47 @@ struct ConfigurationManager
         TokensMappingContainer mTokensMapping;
 
         bool mSilent;
+};
+
+
+/**
+ * \struct escape_hash_filter
+ */
+struct escape_hash_filter : public boost::iostreams::input_filter
+{
+	static const int sEscape = '@';
+	static const int sHashIdentifier = 'h';
+	static const int sEscapeIdentifier = 'a';
+
+	escape_hash_filter();
+	virtual ~escape_hash_filter();
+
+	template <typename Source> int get(Source & src);
+
+	private:
+		std::queue<int> mNext;
+		int mPrevious;
+
+		bool seenNonWhitespace = false;
+		bool finishLine = false;
+};
+
+/**
+ * \class EscapeHashString
+ */
+class EscapeHashString : public std::string
+{
+	public:
+		static std::string processString(const std::string & str);
+
+		EscapeHashString();
+		EscapeHashString(const std::string & str);
+		EscapeHashString(const std::string & str, size_t pos, size_t len = std::string::npos);
+		EscapeHashString(const char * s);
+		EscapeHashString(const char * s, size_t n);
+		EscapeHashString(size_t n, char c);
+		template <class InputIterator>
+		EscapeHashString(InputIterator first, InputIterator last);
 };
 
 } /* namespace Cfg */
--- a/components/files/configurationmanager.hpp
+++ b/components/files/configurationmanager.hpp
@@ -71,21 +71,21 @@ struct ConfigurationManager
  */
 struct escape_hash_filter : public boost::iostreams::input_filter
 {
-	static const int sEscape;
-	static const int sHashIdentifier;
-	static const int sEscapeIdentifier;
+    static const int sEscape;
+    static const int sHashIdentifier;
+    static const int sEscapeIdentifier;
 
-	escape_hash_filter();
-	virtual ~escape_hash_filter();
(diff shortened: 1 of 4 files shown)


Ragora:
It might be better to just include anything after the equal sign to the next line to be a part of the literal value to be assigned. The config files don't really have any proper delimiters to make any better decisions on that and the comments look like they pretty much only appear on their own lines anyway.

zinnschlag:
I don't think that is the right approach. As discussed on the forum, the cleanest way to handle this problem would be to introduce an escape sequence (like %20 for space in an URL). This will not only handle # but also all other special characters that may cause us problems. And it will allow the cfg files to continue using # comments on the same line as other text.

Ragoras idea should also work, but you would have to investigate how to hook your custom parser into boost program options.

AnyOldName3:
That's more or less what it *is* doing with this change, only instead of '%' being the escape character, I've used '@'. It can be used to escape other characters - it's already escaping '@', too. The only change needed to make it completely general would be to make it accept a hexadecimal version of a unicode or ascii character.

If you're instead meaning that the escape sequence should be entered into the config file rather than being escaped automatically when it's read in, that can be done, too - I'd just need to remove the filter on the input stream and put it on the launcher's output stream. It just might make things slightly less obvious to the user if they were entering values manually, and the launcher would need to be modified so it considered things to be comments that started part way through a line.

There's not a whole lot of modification that can be applied to Boost's stuff other than filtering the input and making it store to variables with custom constructors which will remove the filter. It seems to be designed mostly as a black box which can't be modified without being totally replaced.

AnyOldName3:
It should build and run now, so I can get started on making it work like I was told to instead of how I invented.

I've learned that there's a little more freedom to meddle with boost::program_options than I thought, but not enough to easily allow Ragora's idea to be implemented.

I've also uncovered another bug that was hidden by this one - the GUI interprets a hash as an instruction to change the text colour, so that'll need to be fixed. In the attached screenshot, I'm loading RealSignposts#Test.esp:

image

scrawl:
> I've also uncovered another bug that was hidden by this one - the GUI interprets a hash as an instruction to change the text colour, so that'll need to be fixed. In the attached screenshot, I'm loading RealSignposts#Test.esp:

```MyGUI::Textiterator::toTagsString(text)``` should do the trick.

**Comment by AnyOldName3** [comment]
Travis is warning that it can't find a header which definitely does exist. Do I need to do something more than just import it?

Specifically, the issue is:

```
In file included from /home/travis/build/OpenMW/openmw/apps/openmw/main.cpp:6:0:
/home/travis/build/OpenMW/openmw/./components/fallback/validate.hpp:6:53: fatal error: components\files\configurationmanager.hpp: No such file or directory
 #include <components\files\configurationmanager.hpp>
```

**Comment by kcat** [comment]
Use forward slashes ```/```, backlashes ```\``` only work on Windows.

**Comment by AnyOldName3** [comment]
Huh. I could have sworn I copied and pasted that include instead of typing it out.

File: components/files/configurationmanager.hpp [hunk 1]
```diff
@@ -63,6 +65,53 @@ struct ConfigurationManager
...
+        EscapeHashString(const std::string & str);
+        EscapeHashString(const std::string & str, size_t pos, size_t len = std::string::npos);
+        EscapeHashString(const char * s);
+        EscapeHashString(const char * s, size_t n);
+        EscapeHashString(size_t n, char c);
+        template <class InputIterator>
+        EscapeHashString(InputIterator first, InputIterator last);
+
+        std::string toStdString() const;
+        std::string * toStdStringPtr() const;
```
zinnschlag: This function looks evil. Is there a reason why we have it?

components/files/configurationmanager.hpp
```diff
@@ -63,6 +65,53 @@ struct ConfigurationManager
...
+        int mPrevious;
+
+        bool mSeenNonWhitespace;
+        bool mFinishLine;
+};
+
+/**
+ * \class EscapeHashString
+ */
+class EscapeHashString : public std::string
```
zinnschlag: Inheriting from STL containers is generally a very bad idea.

AnyOldName3: I eventually found that out when something didn't work. My background is primarily in Java, where this would have been fine. I refactored it to not need to inherit from std::string, but apparently forgot to actually remove the inheritance itself. That'll be fixed in a moment.

components/files/configurationmanager.hpp
AnyOldName3: It can probably go - I was having an issue with namespaces, but it looked like it was actually with types, so I was double checking something, and then didn't clean it up. This should be resolved in a moment.

AnyOldName3: In fact, I'd cleaned it out of where it was called, so it only needed removing in one place.

components/files/configurationmanager.hpp
AnyOldName3: Apparently I was still relying on a lexical cast, but that should be resolved pretty quickly.

AnyOldName3: This is resolved.

components/files/configurationmanager.hpp
AnyOldName3: This is resolved.

AnyOldName3:
If the current system I've implemented isn't what's wanted, a decision needs to be made as what's actually wanted. I see two main options:

1. Zini's idea - we have an escape character, and interpret a number (potentially in hexadecimal) after it as an ASCII/Unicode code point, much like in a URL. ("%20" -> ' ', "%23" -> '#')

2. Another idea - we have an escape character, and have the character we want escaping directly after it, like how in most programming languages one can put a backslash before a quote character or another backslash. ("% " -> ' ', "%#" -> '#')

I personally think option 2 is a better idea for the following reasons:

* It's more readable if the character we actually want is right there in the config file.
* The hard work is already done as the filter I wrote can be easily modified.
* For characters which aren't E-ASCII (but are in e.g. UTF-16), we need more than two hex digits for option 1, so need to deal with that somehow. This can be worked around for the internal representation option 2 would have to use, though.

kcat:
I kinda prefer Zini's idea, where you do ```%<hex>``` to specify the ASCII code for characters that otherwise would confuse the parser. It's the same thing that URLs do when they contain similar "confusing" path or filename characters (like ```#``` which normally indicates a page anchor, but could be a filename or directory character).

Though I'm not quite sure how that would work with unicode chars. Perhaps something like UTF-8, where multiple 8-bit values are specified in a row with the appropriate bit pattern.

AnyOldName3:
URLs put a % before every two hex digits, then parse the resulting hex string as a UTF-8 character stream.  That means "François"  goes to "Fran%C3%A7ois", as 0xC3A7 is the UTF-8 code point for 'ç'. If for whatever reason we decided to encode every character, we get "%46%72%61%6E%C3%A7%6F%69%73", which then gets parsed as whatever's represented by 0x4672616EC3A76F6973. It can split it up because it was designed to have patterns, as outlined here: https://en.wikipedia.org/wiki/UTF-8#Description

This seems like the kind of thing that there'll be a boost library for, so I'm not hugely concerned about it right now. However, if someone was to come along and put "%46%72%61%6E%C3%A7%6F%69%73" in their config file (or some other more likely string, such as a couple of emoji, something in Chinese, or some ancient runes - all things that could be conceivably required, but would potentially trip something up somewhere along the line), it would be a nuisance to read, and even more of a nuisance to type.

tl;dr: Internally, this is a sensible representation to use, as it can hold everything and is well-defined, but it might not be the best thing to expose to the user.

Everything's somewhat complicated by std::string not technically supporting UTF-8, but Boost and Qt both converting to something with full unicode coverage, so certain things might have surprising results.

kcat:
> URLs put a % before every two hex digits, then parse the resulting hex string as a UTF-8 character stream. That means "François" goes to "Fran%C3%A7ois", as 0xC3A7 is the UTF-8 code point for 'ç'.

Yeah, since two hex digits is one byte, each % declares a one byte value, and characters above 0x7F in UTF-8 use multiple bytes which get concatenated for the unicode value. This method also makes it easier to handle non-ASCII-7 chars even for editors that don't understand UTF-8 (i.e. if the user uses a path with ```ç```, and it saves in some ASCII codepage instead of UTF-8 format, OpenMW could have trouble recognizing the character even with a ```%ç``` mechanism; but ```%C3%A7``` would work as expected regardless).

Aesylwinn:
I think it would be silly to require the user to write any non-ascii character as a hex code. Imagine if you use a writing system not based on the phoenician alphabet...

If you are analyzing the text as ascii characters, then you should be perfectly fine as long as it is encoded in ascii or utf8. If it's encoded in utf16 or utf32, then you need to convert it. Otherwise you may be corrupting the data you pass on depending on the endianness of the data.

kcat:
Less about the user having to write in hex codes, and more about the characters being preserved properly even if the user edits and saves the config file in a text editor that messes up UTF-8 chars.

zinnschlag:
> even if the user edits and saves the config file in a text editor that messes up UTF-8 chars.

That is really the user's problem, isn't it? Normally you would not have to edit the cfg files manually. This is reserved for more experienced users that want more direct control (or hate GUIs). One can assume that such a user knows what he is doing and uses a text editor that is not broken. Do these even still exist? I don't remember encountering any text editor that had problems with UTF-8 for at least half a decade.


I will give the code a more thorough review later this week (rather busy at the moment). I still prefer my original proposal. But what we have here is probably serviceable and the advantages of alternative implementations are marginal enough that it would be a waste of time to do for a different implementation now.

kcat:
> That is really the user's problem, isn't it? Normally you would not have to edit the cfg files manually.

AFAIK, there's currently no way to add data directories without manually editing the cfg file, which is the same cfg file that would pose the most problem. As for whether there are still text editors that have a problem with UTF-8, I wouldn't put it passed Windows, which has historically not liked UTF-8 (instead using 8-bit ASCII codepages, or UTF-16).

AnyOldName3:
Even regular Notepad works fine with UTF-8 in Windows. (I just tested.) It doesn't cope well with the unix-style line endings, though.

zinnschlag:
Sorry for the delay again. Not going to get around to review it today. Either tomorrow or the day after.

zinnschlag:
Finally done with the review. I have two comments:
- You are doing quite a bit of code reformatting on lines that you otherwise don't touch. We generally discourage that.
- We don't like cryptic abbreviations like eHS.

None of these issues are severe enough to justify further delays. I will do a test run now and if nothing else comes up I will merge your code.

psi29a:
yay! Great work. :)

AnyOldName3:
I noticed the reformatting. Visual Studio did that by itself, and wasn't showing me an easy way to put it back. If anyone happens to know a setting that can be changed, that would be appreciated.

As for the abbreviations, I'll keep that in mind in the future.

Thanks. :)

Aesylwinn:
http://stackoverflow.com/questions/18298072/how-do-i-turn-off-auto-formatting-in-vs2013-for-c

AnyOldName3:
Thanks, but that's not a perfect solution for me. The problem with turning it off completely is that I quite like the auto formatting when it's used on lines I've changed. I'll definitely investigate this before my next PR, though.

Aesylwinn:
> I quite like the auto formatting when it's used on lines I've changed.

To each their own. :)

AnyOldName3:
I just had a thought, and after some testing, I found an oversight. I should have a follow-up PR ready momentarily.

AnyOldName3:
As usual, the fix is slightly more complex than I thought, so won't be quite ready tonight.

PR #999, opened 2016-07-11
Trim unneeded code

Allofich:
Trim out unneeded variable mStarted and some redundant code.

When escort packages repeat and use aipackage.cpp's `pathTo()` repeat, they need something to get them past the very first call, because their `mPrevDest` is the same as `dest` and their path is empty. Checking for the empty path and using this to start a path recalculation seems to handle this situation. This should also avoid the theoretically possible? situation of `doesPathNeedRecalc()` returning false for a destination of 0,0,0, which is what `mPrevDest` is initialized to.

### DIFF
--- a/apps/openmw/mwmechanics/aiescort.cpp
+++ b/apps/openmw/mwmechanics/aiescort.cpp
@@ -68,7 +68,6 @@ namespace MWMechanics
             if (mRemainingDuration <= 0)
             {
                 mRemainingDuration = mDuration;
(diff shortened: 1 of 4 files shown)


Allofich:
It's not clear to me why, but while aiescort has a problem without the added check
 `mPathFinder.getPath().empty()` in `doesPathNeedRecalc()` due, it appears, to its path being empty on a repeated package, aiactivate seems to have a non-empty path on repeats and works fine even without the added check...

PR #1001, opened 2016-07-16
Make non-actors play spell cast visuals and sounds

Allofich:
These changes make trapped objects, or any non-actor object given a "cast" command through the console or script, play their casting sound effect and visual casting effect like the original game  (except for the glow).

There are still problems that this PR does not address.

- The visual effect is located differently from the original game. From observation, I think the original game's effect may be created at the same point used for an object's "position," while in OpenMW it is created higher up than that.

- There is no glow effect like there is in the original game. I started looking at trying to adapt the "getEnchantmentColor" and "addGlow" functions in animation.cpp for this, but it's uncertain ground for me, and I'm not sure how to proceed there.

In the original game statics seem to only get the glow when they cast a spell, not the other visual effect, so I disabled the non-glow effect for them.
 
- Maybe not strictly related, but the "cast" command should make an NPC play their spellcasting animation and aim at the target, which currently it doesn't.

### DIFF
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -1126,12 +1126,13 @@ namespace MWRender
     };
 
     void Animation::addSpellCastGlow(osg::Vec4f glowColor){
-        addGlow(mObjectRoot, glowColor, true);
+        if (mSpellGlowUpdater == NULL) // only start a new spell glow if there isn't one already 
+            addGlow(mObjectRoot, glowColor, true);
         MWBase::Environment::get().getMechanicsManager()->add(mPtr);
     }
 
     void Animation::updateSpellGlow(float duration){
-        if (mGlowUpdater != NULL && (mGlowUpdater->getHasDuration()))
+        if (mSpellGlowUpdater != NULL)
             mSpellGlowDuration += duration;
         if (mSpellGlowDuration >= 1.5f) // length of spell glow effect was measured from original game as around 1.5 seconds
             removeSpellGlow();
@@ -1145,10 +1146,10 @@ namespace MWRender
         else
             writableStateSet = osg::clone(mObjectRoot->getStateSet(), osg::CopyOp::SHALLOW_COPY);
 
-        std::vector<osg::ref_ptr<osg::Texture2D> > Textures = mGlowUpdater->getTextures();
-        int TexUnit = mGlowUpdater->getTexUnit();
-        mObjectRoot->removeUpdateCallback(mGlowUpdater);
-        mGlowUpdater = NULL;
+        std::vector<osg::ref_ptr<osg::Texture2D> > Textures = mSpellGlowUpdater->getTextures();
+        int TexUnit = mSpellGlowUpdater->getTexUnit();
+        mObjectRoot->removeUpdateCallback(mSpellGlowUpdater);
+        mSpellGlowUpdater = NULL;
 
         for (size_t index = 0; index < Textures.size(); index++)
         {
@@ -1157,7 +1158,10 @@ namespace MWRender
         }
 
         mObjectRoot->setStateSet(writableStateSet);
-        writableStateSet->removeUniform(mUniform);
+        if (writableStateSet->getUniform("envMapColor2")) // if we added a second uniform, remove that one instead of the original
+            writableStateSet->removeUniform("envMapColor2");
+        else
+            writableStateSet->removeUniform("envMapColor");
         mResourceSystem->getSceneManager()->recreateShaders(mObjectRoot);
         mSpellGlowDuration = 0.f;
     }
@@ -1186,9 +1190,13 @@ namespace MWRender
         FindLowestUnusedTexUnitVisitor findLowestUnusedTexUnitVisitor;
         node->accept(findLowestUnusedTexUnitVisitor);
         int texUnit = findLowestUnusedTexUnitVisitor.mLowestUnusedTexUnit;
-        mGlowUpdater = new GlowUpdater(texUnit, glowColor, textures, hasDuration);
-        //osg::ref_ptr<GlowUpdater> glowupdater (new GlowUpdater(texUnit, glowColor, textures));
-        node->addUpdateCallback(mGlowUpdater);
+
+        osg::ref_ptr<GlowUpdater> glowUpdater = new GlowUpdater(texUnit, glowColor, textures, hasDuration);
+
+        if (hasDuration) // store the glowUpdater for later removal and checking if this is a spell glow
+            mSpellGlowUpdater = glowUpdater;
+
+        node->addUpdateCallback(glowUpdater);
 
         // set a texture now so that the ShaderVisitor can find it
         osg::ref_ptr<osg::StateSet> writableStateSet = NULL;
@@ -1200,8 +1208,11 @@ namespace MWRender
             node->setStateSet(writableStateSet);
         }
         writableStateSet->setTextureAttributeAndModes(texUnit, textures.front(), osg::StateAttribute::ON);
-        mUniform = new osg::Uniform("envMapColor", glowColor);
-        writableStateSet->addUniform(mUniform);
+        
+        if (hasDuration && writableStateSet->getUniform("envMapColor")) // if we are adding a spell glow to something with an enchantment glow
+            writableStateSet->addUniform(new osg::Uniform("envMapColor2", glowColor));
+        else
+            writableStateSet->addUniform(new osg::Uniform("envMapColor", glowColor));
 
         mResourceSystem->getSceneManager()->recreateShaders(node);
     }
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -268,26 +268,29 @@ namespace MWRender
 
         virtual void setDefaults(osg::StateSet *stateset)
         {
-            stateset->setTextureMode(mTexUnit, GL_TEXTURE_2D, osg::StateAttribute::ON);
-
-            osg::TexGen* texGen = new osg::TexGen;
-            texGen->setMode(osg::TexGen::SPHERE_MAP);
-
-            stateset->setTextureAttributeAndModes(mTexUnit, texGen, osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
-
-            osg::TexEnvCombine* texEnv = new osg::TexEnvCombine;
-            texEnv->setSource0_RGB(osg::TexEnvCombine::CONSTANT);
-            texEnv->setConstantColor(mColor);
-            texEnv->setCombine_RGB(osg::TexEnvCombine::INTERPOLATE);
-            texEnv->setSource2_RGB(osg::TexEnvCombine::TEXTURE);
-            texEnv->setOperand2_RGB(osg::TexEnvCombine::SRC_COLOR);
-
-            stateset->setTextureAttributeAndModes(mTexUnit, texEnv, osg::StateAttribute::ON);
-
-            // Reduce the texture list back down by one when a temporary glow finishes
-            // to allow FindLowestUnusedTexUnitVisitor to choose the same texunit again.
             if (mDone)
-                stateset->getTextureAttributeList().resize(stateset->getTextureAttributeList().size() - 1);
+            {
+                stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXTURE);
+                stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXGEN);
+            }
+            else
+            {
+                stateset->setTextureMode(mTexUnit, GL_TEXTURE_2D, osg::StateAttribute::ON);
+                osg::TexGen* texGen = new osg::TexGen;
+                texGen->setMode(osg::TexGen::SPHERE_MAP);
+
+                stateset->setTextureAttributeAndModes(mTexUnit, texGen, osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
+
+                osg::TexEnvCombine* texEnv = new osg::TexEnvCombine;
+                texEnv->setSource0_RGB(osg::TexEnvCombine::CONSTANT);
+                texEnv->setConstantColor(mColor);
+                texEnv->setCombine_RGB(osg::TexEnvCombine::INTERPOLATE);
+                texEnv->setSource2_RGB(osg::TexEnvCombine::TEXTURE);
+                texEnv->setOperand2_RGB(osg::TexEnvCombine::SRC_COLOR);
+
+                stateset->setTextureAttributeAndModes(mTexUnit, texEnv, osg::StateAttribute::ON);
+                stateset->addUniform(new osg::Uniform("envMapColor", mColor));
+            }
         }
 
         virtual void apply(osg::StateSet *stateset, osg::NodeVisitor *nv)
@@ -295,7 +298,6 @@ namespace MWRender
             if (mColorChanged){
                 this->reset();
                 setDefaults(stateset);
-                mResourceSystem->getSceneManager()->recreateShaders(mNode);
                 mColorChanged = false;
             }
             if (mDone)
@@ -313,10 +315,10 @@ namespace MWRender
             {
                 if (mOriginalDuration >= 0) // if this glowupdater was a temporary glow since its creation
                 {
-                    for (size_t i = 0; i < mTextures.size(); i++)
-                        stateset->removeTextureAttribute(mTexUnit, mTextures[i]);
+                    stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXTURE);
                     this->reset();
                     mDone = true;
+                    mResourceSystem->getSceneManager()->recreateShaders(mNode);
                 }
                 if (mOriginalDuration < 0) // if this glowupdater was originally a permanent glow
                 {
@@ -325,9 +327,7 @@ namespace MWRender
                     mColor = mOriginalColor;
                     this->reset();
                     setDefaults(stateset);
-                    stateset->addUniform(new osg::Uniform("envMapColor", mColor));
                 }
-                mResourceSystem->getSceneManager()->recreateShaders(mNode);
             }
         }
 
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -90,7 +90,7 @@ namespace
     {
     public:
         GlowUpdater(int texUnit, osg::Vec4f color, const std::vector<osg::ref_ptr<osg::Texture2D> >& textures,
-            osg::ref_ptr<osg::Node> node, float maxduration, Resource::ResourceSystem* resourcesystem, osg::Uniform* uniform)
+            osg::ref_ptr<osg::Node> node, float maxduration, Resource::ResourceSystem* resourcesystem)
             : mTexUnit(texUnit)
             , mColor(color)
             , mTextures(textures)
@@ -99,7 +99,6 @@ namespace
             , mStartingTime(0)
             , mResourceSystem(resourcesystem)
             , mDone(false)
-            , mUniform(uniform)
             , mWatchedSpellGlow(NULL)
         {
         }
@@ -139,8 +138,8 @@ namespace
             // the permanent glow to display its glow texture cycling properly.
             if (mWatchedSpellGlow != NULL && mWatchedSpellGlow->isDone())
             {
-               mNode->removeUpdateCallback(mWatchedSpellGlow);
-               mWatchedSpellGlow = NULL;
+                mNode->removeUpdateCallback(mWatchedSpellGlow);
+                mWatchedSpellGlow = NULL;
             }
             
             // Set the starting time to measure glow duration from if this is a temporary glow
@@ -160,15 +159,11 @@ namespace
                     writableStateSet = osg::clone(mNode->getStateSet(), osg::CopyOp::SHALLOW_COPY);
                 
                 for (size_t i = 0; i < mTextures.size(); i++)
-                {
                     writableStateSet->removeTextureAttribute(mTexUnit, mTextures[i]);
-                }
 
-                writableStateSet->removeUniform(mUniform); // remove the uniform given to the temporary glow in addglow()
                 mNode->setStateSet(writableStateSet);
-                
                 this->reset(); // without this a texture from the glow will continue to show on the object
-                mResourceSystem->getSceneManager()->recreateShaders(mNode);             
+                mResourceSystem->getSceneManager()->recreateShaders(mNode);
                 mDone = true;
             }
         }
@@ -212,7 +207,6 @@ namespace
         float mStartingTime;
         Resource::ResourceSystem* mResourceSystem;
         bool mDone;
-        osg::Uniform* mUniform;
         osg::ref_ptr<GlowUpdater> mWatchedSpellGlow;
     };
 
@@ -1246,10 +1240,7 @@ namespace MWRender
         node->accept(findLowestUnusedTexUnitVisitor);
         texUnit = findLowestUnusedTexUnitVisitor.mLowestUnusedTexUnit;
     
-        osg::Uniform* uniform = new osg::Uniform("envMapColor", glowColor);
-
-
-        osg::ref_ptr<GlowUpdater> glowUpdater = new GlowUpdater(texUnit, glowColor, textures, node, glowDuration, mResourceSystem, uniform);
+        osg::ref_ptr<GlowUpdater> glowUpdater = new GlowUpdater(texUnit, glowColor, textures, node, glowDuration, mResourceSystem);
 
         node->addUpdateCallback(glowUpdater);
         if (node->getUpdateCallback() &&
@@ -1267,7 +1258,7 @@ namespace MWRender
             node->setStateSet(writableStateSet);
         }
         writableStateSet->setTextureAttributeAndModes(texUnit, textures.front(), osg::StateAttribute::ON);
-        writableStateSet->addUniform(uniform);
+        writableStateSet->addUniform(new osg::Uniform("envMapColor", glowColor));
 
         mResourceSystem->getSceneManager()->recreateShaders(node);
     }
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -38,6 +38,7 @@
 #include "../mwworld/esmstore.hpp"
 #include "../mwworld/class.hpp"
 #include "../mwworld/cellstore.hpp"
+#include "../mwbase/mechanicsmanager.hpp"
 
 #include "../mwmechanics/character.hpp" // FIXME: for MWMechanics::Priority
 
@@ -89,10 +90,11 @@ namespace
     class GlowUpdater : public SceneUtil::StateSetUpdater
     {
     public:
-        GlowUpdater(int texUnit, osg::Vec4f color, const std::vector<osg::ref_ptr<osg::Texture2D> >& textures)
+        GlowUpdater(int texUnit, osg::Vec4f color, const std::vector<osg::ref_ptr<osg::Texture2D> >& textures, bool hasDuration)
             : mTexUnit(texUnit)
             , mColor(color)
             , mTextures(textures)
+            , mHasDuration(hasDuration)
         {
         }
 
@@ -122,10 +124,26 @@ namespace
             stateset->setTextureAttribute(mTexUnit, mTextures[index], osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
         }
 
+        bool const getHasDuration()
+        {
+            return mHasDuration;
+        }
+
+        std::vector<osg::ref_ptr<osg::Texture2D> > const getTextures()
+        {
+            return mTextures;
+        }
+
+        int const getTexUnit()
+        {
+            return mTexUnit;
+        }
+
     private:
         int mTexUnit;
         osg::Vec4f mColor;
         std::vector<osg::ref_ptr<osg::Texture2D> > mTextures;
+        bool mHasDuration;
     };
 
     class NodeMapVisitor : public osg::NodeVisitor
@@ -340,6 +358,7 @@ namespace MWRender
         , mHeadYawRadians(0.f)
         , mHeadPitchRadians(0.f)
         , mAlpha(1.f)
+        , mSpellGlowDuration(0.f)
     {
         for(size_t i = 0;i < sNumBlendMasks;i++)
             mAnimationTimePtr[i].reset(new AnimationTime);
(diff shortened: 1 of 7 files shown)


Allofich:
The main change is that I moved the code for creating the visual effect and the sound out from the character controller code in character.cpp and into the spell code in spellcasting.cpp.

scrawl:
> There is no glow effect like there is in the original game. I started looking at trying to adapt the "getEnchantmentColor" and "addGlow" functions in animation.cpp for this, but it's uncertain ground for me, and I'm not sure how to proceed there.

```addGlow(mObjectRoot);``` should work, you'll just need to add code for removing the glow later, and some way to prevent interference with the ```setAlpha``` / ```setRenderBin``` methods that also change the mObjectRoot's StateSet.

Note that once a ```StateSet``` has been submitted for rendering, you can no longer modify it (thread safety issues). The common way to deal with this which you'll see throughout the code is cloning the existing StateSet and then assigning that new StateSet to the node.

**Comment by scrawl** [comment]
> In the original game statics seem to only get the glow when they cast a spell, not the other visual effect, so I disabled the non-glow effect for them.

Some statics do, e.g. ex_de_dock_steps_01 near the Seyda Neen lighthouse. I think it might be related to the size of the object. Objects that are too big do not cast the effect?

**Comment by Allofich** [comment]
Updated. This now adds the spell glow to non-actors when they cast spells. I made the spell glow last 1.5 seconds, as this seems to be about how long it takes based on recording in-game footage and then checking it in a video editor. It seemed to vary a little but generally be about 1.5 seconds.

I only have very little understanding of openscenegraph or opengl. I don't, for example, know what a "uniform" is. When removing the spell glow I basically did a reverse of the things that had been done in addglow(), including removing the uniform, but I don't know whether or not this is all necessary. If it's not the code could be shortened.

This glow works for items with an enchanted glow, which I discovered does not work in the original engine (after the spell glow ends, the enchantment glow is also removed. It reappears if you, for example, take the item and then put it back in the game world.)

I'm sure there will probably be things that need changing, but if things look good I think this should also be used for the telekinesis glow when using items with telekinesis.

I would like to leave the issues of the swirling visual effect and the behavior when making an actor use the "cast" command beyond the scope of this PR. I disabled the visual effect that I had initially enabled, so this is just the casting sound effect and the glow.

**Comment by Allofich** [comment]
Made some fixes, but to be honest I don't have a super clear understanding of it all.

The travis build put up a bunch of warnings about using the anonymous namespace in a header file. This was because I want to use the glowupdater class in a member variable and needed to declare a variable in the header. I moved the glowupdater class into the mwrender scope to hopefully quiet the warnings, but I realize there may have been a reason for having glowupdater in the anonymous namespace, in which case things will need to be rewritten again I guess.

I also noticed an issue where the spell glow would only work for three tries in a row. I discovered that each iteration was using one "texUnit" higher than the last, even though I didn't think this should happen. I fixed the issue by using clear() on the texture attribute list when removing the spell glow, but I don't know why this was necessary. Apparently doing removeTextureAttribute() on all the textures isn't enough, or maybe I was somehow missing one on each iteration?

Currently I'm not aware of any outstanding problems. I'll try to fix any warnings that the builds give.

**Comment by Allofich** [comment]
Fixed a couple warnings about some ineffective "const" qualifiers.

**Comment by Allofich** [comment]
Looks like the clear() on the texture attributes list is all that's needed, so trimmed some code.

**Comment by Allofich** [comment]
> some way to prevent interference with the setAlpha / setRenderBin methods that also change the mObjectRoot's StateSet.

Is there a problem? I don't see how interference with those methods can occur.

**Comment by Allofich** [comment]
Added telekinesis glow when using telekinesis on a door. From some testing what I saw was:

1. There only seems to be a telekinesis glow when activating animated doors.
2. There seems to be a somewhat irregular delay between the application of the glow if you repeatedly activate a door with telekinesis. I'm not sure whether it's worth trying to replicate this.
3. The glow only seems to last as long as the door's animation, even if the door is locked and doesn't move when you activate it.

I think 3 is worth implementing, but I'm not sure how to get the time of an animation. And if we deal with activating doors that have been halted partway through their animation (that can happen, right?) it might be kind of tricky...I think having a telekinesis glow is better than none, even if the timing isn't perfect, so I put timing it to the animation as a TODO for now.

**Comment by scrawl** [comment]
Uniforms are variables set by the application that can be used in the shader program (e.g. [here](https://github.com/OpenMW/openmw/blob/master/files/shaders/objects_fragment.glsl#L134) ).  Using them as a generic storage for non-shader purposes seems like a bad idea.

> +        MWBase::Environment::get().getMechanicsManager()->add(mPtr);

I'm afraid this change might break compatibility with the original game. Some classes are not added to the mechanics manager on purpose (e.g. statics), as they should not automatically play idle animations.

I assume that you've added this to be able to call updateSpellGlow() from runAnimation(). Another way to run updates would be to do it within the UpdateCallback (i.e. the SpellGlowUpdater). You can check the NodeVisitor's simulationTime to see how much time has passed. I haven't tested it but in theory an UpdateCallback should even be able to remove itself when it's no longer needed (i.e. call node->removeUpdateCallback(this) from within apply().

> I also noticed an issue where the spell glow would only work for three tries in a row. I discovered that each iteration was using one "texUnit" higher than the last, even though I didn't think this should happen. I fixed the issue by using clear() on the texture attribute list when removing the spell glow, but I don't know why this was necessary.

I still have this problem, are you sure that you've pushed all changes?

> Added telekinesis glow when using telekinesis on a door.

Isn't there also a glow effect when opening locks with a spell?

**Comment by Allofich** [comment]
> Uniforms are variables set by the application that can be used in the shader program (e.g. here ). 

I knew that they had something to do with shaders, because I did look them up. I assume I need to get rid of the uniform added in addglow() in order to get back to the state before running addglow().

> Using them as a generic storage for non-shader purposes seems like a bad idea.
Are you referring to me storing the uniform of something with an enchanted glow to add it back after applying the spell glow? I'll try to avoid doing that.

> I assume that you've added this to be able to call updateSpellGlow() from runAnimation(). Another way to run updates would be to do it within the UpdateCallback (i.e. the SpellGlowUpdater).

Yes, that's why I used add(). I actually tried writing this up earlier from the callback, but I didn't know whether it would work or be safe or correct to remove the callback from within itself. I'll try and see if I can do it that way.

> I still have this problem, are you sure that you've pushed all changes?

It looks like some changes didn't get pushed for some reason. Anyway I'm going to do some rewriting so I'll push it all then.

> Isn't there also a glow effect when opening locks with a spell?

I'm pretty sure there is. I forgot about that. I'll look at adding that in.

**Comment by Allofich** [comment]
Unfortunately, I'm getting a crash when I try to call node->removeUpdateCallback(this) from within apply().

**Comment by Allofich** [comment]
Maybe it would be possible to check whether the object that is going to get the glow is in the mechanicsmanager or not, and if it is do nothing, but if it isn't add it as a special case that only runs the spell glow updater? I'm going to try that.

**Comment by scrawl** [comment]
> Unfortunately, I'm getting a crash when I try to call node->removeUpdateCallback(this) from within apply().

Ah, that's too bad - probably because the object is deleting itself and then continues to run its traverse method with a deleted class instance.

I still think that the UpdateCallback is the way to go in the spirit of making the code self-contained. What about just leaving the SpellGlowUpdater there when the effect is gone. The performance overhead of an updater that doesn't do anything shouldn't be too bad, and I don't think we'll have too many of these created in a typical play session anyway.

**Comment by Allofich** [comment]
Yeah, that sounds like it might be the best way, and I agree that it should keep the changes from being as spread out. If I understand correctly the updatecallbacks seem to be removed when the cell is changed anyway.

Since you seem to be around, maybe I can ask you also about the uniform, as I still don't really understand it very well. Do I need to remove it when the glow ends? And, in the case of making an enchanted item cast a spell (a minor case but still one I would like to support), there will already be an "envMapColor" uniform, so do we need to somehow store that uniform, or at least it's glow color, and then reapply it when the temporary glow ends? Or just put a second uniform on top and then remove that one put on a new one (can there be two at once? Do they both need to be named "envMapColor"?) at the end? Also, is there a way to test in-game the effects from changing the uniform? Sorry for all the questions, but it would help me know what I am doing if you can answer any.

**Comment by Allofich** [comment]
Another question, sorry, but I'd appreciate it. Should I call recreateShaders again, which is called in addglow(), after removing the glow?

**Comment by scrawl** [comment]
> Do I need to remove it when the glow ends?

Not necessarily, but there is not much of a point to keep it I think?

> And, in the case of making an enchanted item cast a spell (a minor case but still one I would like to support), there will already be an "envMapColor" uniform, so do we need to somehow store that uniform, or at least it's glow color, and then reapply it when the temporary glow ends?

Yes, that's how you could do it.

> Or just put a second uniform (can there be two at once? Do they both need to be named "envMapColor"?) and then remove that one at the end?

You can add as many uniforms as you want (provided they have different names) but as I said before I don't think its good practice to add uniforms that the shader isn't using.

> Also, is there a way to test in-game the effects from changing the uniform?

The uniform only has an effect when shaders are used. To test, add to the settings.cfg:
<pre>
[Shaders]
force shaders = true
</pre>

> Should I call recreateShaders again, which is called in addglow(), after removing the glow?

Yes. The shader code generated depends on the textures that are used. When a texture is removed the shaders need to be recreated.

**Comment by Allofich** [comment]
Thanks. And, some good news: I was able to call removeUpdateCallback(this) from within apply() without crashing by making it the final line of the apply() method. Without doing this I couldn't seem to keep the object from continuing to flicker with the glow texture even after I'd tried to remove it.

Right now I seem to be having the problem with only three casts in a row possible before the object stops glowing again, though. Hopefully I can figure that out and get a new version up before too long. (Maybe not today)

**Comment by Allofich** [comment]
I seem to have fixed the problem with only three casts in a row. I've also made progress trimming the amount of changes to the code. Unless a new problem comes up I hopefully will have an improved version up within the next few days or so.

**Comment by scrawl** [comment]
> I was able to call removeUpdateCallback(this) from within apply() without crashing by making it the final line of the apply() method.

Even if that happens to not crash, it is not correct code. SpellGlowUpdater is derived from StateSetUpdater. After calling ```apply()```, it calls ```traverse(node, nv);``` to process nested callbacks and update child nodes.  The implementation of traverse(..) is:
<pre>
bool Callback::traverse(Object* object, Object* data) {
    if (_nestedCallback.valid()) return _nestedCallback->run(object, data);
    ...
</pre>
 The ```_nestedCallback``` is accessing a member variable which is no longer valid after ```this``` is deleted.

Technically, it *is* possible for an object to delete itself, but we have to be very careful not to access the object after deleting it. Not a good coding practice, I suppose. https://stackoverflow.com/questions/6952083/deleting-self-object-inside-class

So let's not have the object remove/delete itself, that was a stupid idea of mine.

**Comment by scrawl** [comment]
> Without somehow removing the updatecallback I couldn't seem to keep the object from continuing to flicker with the glow texture even after I'd tried to remove it, so this helps.

The flickering is because there are two copies of the StateSet taking turns every other frame. This is done because we can't modify a StateSet while it's being rendered. Try using the reset() / setDefaults() methods.

**Comment by Allofich** [comment]
I've updated it again. Functionally, the only problem I'm aware of, is that this doesn't work correctly when shaders are enabled. I don't know if this is a problem right now or not. Are shaders considered supported? Unfortunately I don't really have any idea what the problem is. The glow on an object with no glow yet will only glow once, and no glow will be added for an object with an enchantment glow.

Without shaders, I'm not aware of there functionally being any problems. The "Open" spell now creates a glow as well.

As far as the implementation, in order to get the permanent enchantment glow of an enchanted item to continue displaying correctly after making the item cast a spell, I had to make the enchanted glow's UpdateCallback remove the temporary glow's UpdateCallback. Otherwise the enchantment glow would stay stuck at a single texture, instead of cycling through all 32.

There was a problem of the texUnit climbing by one every time I gave a temporary spell glow to an object without an enchantment glow. After climbing the texUnit grew three times, the glow would stop appearing.I found that TextureAttributeList().size() for the stateset was growing by one during the glowUpdater's setDefaults() method, and when invoking addGlow(), the FindLowestUnusedTexUnitVisitor nodevisitor would pick a texUnit one higher to match the one higher size of TextureAttributeList. I wasn't sure what the best way to fix this was, but I decided to fix it by undoing the growth by 1 of TextureAttributeList in setDefaults() when a temporary glowUpdater ends and setDefaults() gets called again.

**Comment by Allofich** [comment]
The other way I thought of to fix the TexUnit issue is by getting the TexUnit from a spell glow updater if  one has previously been used on the object, and forcing it to be reused instead of calling FindLowestUnusedTexUnitVisitor. I'm thinking maybe now that is a preferable solution.

File: apps/openmw/mwrender/animation.cpp [hunk 1]
```diff
@@ -1106,7 +1183,25 @@ namespace MWRender
...
+
+        osg::Vec4f glowColor(1,1,1,1);
+        glowColor.x() = effect->mData.mRed / 255.f;
+        glowColor.y() = effect->mData.mGreen / 255.f;
+        glowColor.z() = effect->mData.mBlue / 255.f;
+
+        if (!mObjectRoot->getUpdateCallback()) // If there is no glow on object
+            addGlow(mObjectRoot, glowColor, 1.5); // Glow length measured from in-game as about 1.5 seconds
+
+        else if (dynamic_cast <GlowUpdater*>(mObjectRoot->getUpdateCallback())->isDone() == true) // If there was a temporary glow on object and it finished
```
scrawl: If you use a dynamic_cast you need to check the result of the cast.

```
A* a = dynamic_cast<A*>(b);
if (a)
    // do stuff
```

or

```
A& a = dynamic_cast<A&>(*b); // throws an exception if the cast fails
```

If you are sure that the cast will not fail you should use a static_cast.

In this case I would prefer storing a pointer to the SpellGlowUpdater in ```this``` to eliminate the need for casting.

apps/openmw/mwrender/animation.cpp
```diff
@@ -113,19 +120,89 @@ namespace
...
             int index = (int)(time*16) % mTextures.size();
             stateset->setTextureAttribute(mTexUnit, mTextures[index], osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
+
+            if ((mMaxDuration >= 0) && (time - mStartingTime > mMaxDuration)) // If this is a temporary glow and it has finished its duration
+            {
+                osg::ref_ptr<osg::StateSet> writableStateSet = NULL;
+                if (!mNode->getStateSet())
+                    writableStateSet = mNode->getOrCreateStateSet();
+                else
+                    writableStateSet = osg::clone(mNode->getStateSet(), osg::CopyOp::SHALLOW_COPY);
```
scrawl: No need for a copy here, the StateSet passed to apply() is safe to modify.

apps/openmw/mwrender/animation.cpp
```diff
@@ -1127,11 +1222,26 @@ namespace MWRender
             textures.push_back(tex);
         }
 
+        int texUnit;
```
scrawl: Why this declaration here?

scrawl:
> Are shaders considered supported? 

Yes. For normal-mapped objects, shaders are used by default. Even for other objects, users may wish to enable shaders for more realistic lighting (per-pixel, unclamped). Shaders may be used to implement shadows in the future. Also, shaders will be needed to port to platforms like GLES2 or GL3 core profile that don't support fixed-function rendering.

apps/openmw/mwrender/animation.cpp
```diff
@@ -113,19 +120,89 @@ namespace
...
+
+            // If there is a permanent enchantment glow already on this object, give the
+            // permanent glow a pointer to a temporary glow added in a nested update callback
+            // The permanent glow will remove the temporary glow when it finishes, allowing
+            // the permanent glow to display its glow texture cycling properly.
+            if (mWatchedSpellGlow != NULL && mWatchedSpellGlow->isDone())
+            {
+                mNode->removeUpdateCallback(mWatchedSpellGlow);
+                mWatchedSpellGlow = NULL;
+            }
```
scrawl: That is an ugly workaround. If you want to use multiple updaters then the proper way to do it would be using the ```CompositeStateSetUpdater```.
```
osg::ref_ptr<CompositeStateSetUpdater> composite (new CompositeStateSetUpdater);
composite->addController( spellglowupdater1 );
composite->addController( spellglowupdater2 );
mObjectRoot->addUpdateCallback(composite);
```

That way both updaters are not messing with the other's StateSet.

However... there is also a problem with the feature you are trying to implement. You want to have a second glow texture applied on top of the first one. Correct? Problem is that the shader isn't set up to handle that. It'll work in fixed function mode, but the shader would only show one of the glow textures. Extending the shader code to support an arbitrary number of glow textures would be pretty complicated.

Easiest thing would be not to allow spell glow on things that already glow but I guess you're too keen on that feature ;)

I suggest to just temporarily override the color of the first glow updater for the duration of the spell glow. Should look close enough.

scrawl:
> Functionally, the only problem I'm aware of, is that this doesn't work correctly when shaders are enabled. I don't know if this is a problem right now or not. Are shaders considered supported? Unfortunately I don't really have any idea what the problem is. 

> The glow on an object with no glow yet will only glow once

This was caused by a bug also affecting master. I could swear that I've fixed the same bug before but maybe that was in another section of the codebase. Fixed now.

> and no glow will be added for an object with an enchantment glow.

As I've explained the shader isn't made to handle more than one environment map.

apps/openmw/mwrender/animation.cpp
Allofich: OK, I'm storing the pointer again, as I did in an earlier implementation in this PR. For this, though, as I mentioned above I moved glowUpdater into the mwrender namespace so that I can avoid a slew of warnings on the travis build about using the anonymous namespace in the header file, where I declare the glowupdater that I store the pointer in (I'll be watching the travis builds just to make sure).

apps/openmw/mwrender/animation.cpp
Allofich: OK, removed it.

apps/openmw/mwrender/animation.cpp
Allofich: Sorry, probably left over from some code I wrote and then removed. Fixed.

apps/openmw/mwrender/animation.cpp
Allofich: In the new implementation, there shouldn't eve be more than one glowupdater applied at a time. In the case where there is already a glowupdater because of the object being an enchanted item, I temporarily change the color of the glow, as you suggest.

Allofich:
Thanks, the problem of the glow only happening once is fixed.

Unfortunately, there is still a minor issue when shaders are enabled. The spell glow won't appear on object that already has an enchanted glow, even though I'm not putting on a second glowupdater. I tried using recreateShaders(node) after the color is changed, but that didn't fix it. Does this have anything to do with the shader limitations you mentioned?

I added in the glow effect for lock spells, by the way, which I had forgotten.

There is one minor thing I considered implemented but dropped. I noticed that in the original game, when you use telekinesis to activate a trapped door, occasionally the door will play its trap spell glow instead of the telekinesis glow. It's pretty rare. I got it on 1 out of 10 tries when I did a test. I started to implement it but considering how rarely it occurs I thought maybe it is a bug rather than intended, so I just left the behavior for OpenMW as always playing the telekinesis effect.

scrawl:
You'll need to update the uniform after the color is changed.

apps/openmw/mwrender/animation.cpp
```diff
@@ -366,6 +247,124 @@ namespace
...
+            if (mDone)
+                stateset->getTextureAttributeList().resize(stateset->getTextureAttributeList().size() - 1);
+        }
+
+        virtual void apply(osg::StateSet *stateset, osg::NodeVisitor *nv)
+        {
+            if (mColorChanged){
+                this->reset();
+                setDefaults(stateset);
+                mResourceSystem->getSceneManager()->recreateShaders(mNode);
```
scrawl: recreateShaders() is not needed here. The change of color should be applied by updating the uniform.

apps/openmw/mwrender/animation.cpp
```diff
@@ -366,6 +247,124 @@ namespace
...
+                    mDone = true;
+                }
+                if (mOriginalDuration < 0) // if this glowupdater was originally a permanent glow
+                {
+                    mDuration = mOriginalDuration;
+                    mStartingTime = 0;
+                    mColor = mOriginalColor;
+                    this->reset();
+                    setDefaults(stateset);
+                    stateset->addUniform(new osg::Uniform("envMapColor", mColor));
```
scrawl: Try moving this line into setDefaults().

apps/openmw/mwrender/animation.cpp
```diff
@@ -366,6 +247,124 @@ namespace
...
+                }
+                mResourceSystem->getSceneManager()->recreateShaders(mNode);
```
scrawl: Move into the ```if (mOriginalDuration >= 0) { ... }``` statement.

apps/openmw/mwrender/animation.cpp
```diff
@@ -366,6 +247,124 @@ namespace
...
+            float time = nv->getFrameStamp()->getSimulationTime();
+            int index = (int)(time*16) % mTextures.size();
+            stateset->setTextureAttribute(mTexUnit, mTextures[index], osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
+
+            if ((mDuration >= 0) && (time - mStartingTime > mDuration)) // If this is a temporary glow and it has finished its duration
+            {
+                if (mOriginalDuration >= 0) // if this glowupdater was a temporary glow since its creation
+                {
+                    for (size_t i = 0; i < mTextures.size(); i++)
+                        stateset->removeTextureAttribute(mTexUnit, mTextures[i]);
```
scrawl: Easier:
<pre>
stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXTURE);
</pre>

apps/openmw/mwrender/animation.cpp
```diff
@@ -366,6 +247,124 @@ namespace
...
+            texEnv->setCombine_RGB(osg::TexEnvCombine::INTERPOLATE);
+            texEnv->setSource2_RGB(osg::TexEnvCombine::TEXTURE);
+            texEnv->setOperand2_RGB(osg::TexEnvCombine::SRC_COLOR);
+
+            stateset->setTextureAttributeAndModes(mTexUnit, texEnv, osg::StateAttribute::ON);
+
+            // Reduce the texture list back down by one when a temporary glow finishes
+            // to allow FindLowestUnusedTexUnitVisitor to choose the same texunit again.
+            if (mDone)
+                stateset->getTextureAttributeList().resize(stateset->getTextureAttributeList().size() - 1);
```
scrawl: if ```mDone == true```, then the all the ```setTexture...``` calls in this function are redundant. Try:

<pre>
if (mDone)
{
   stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXTURE);
   stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXGEN);
}
else
{
   stateset->setTextureMode(mTexUnit, GL_TEXTURE_2D, osg::StateAttribute::ON);
   // etc ........
}
</pre>

Allofich:
I've added all the changes. I didn't realize you were still making notes and the "Cleanups" commit doesn't include the stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXTURE); change yet, although I have it locally. Shall I go ahead and upload that?

With shaders on, the enchanted item is still not glowing correctly. Before it would just continue with its original glow, now it loses any glow whatsoever. Did I make a mistake in applying the fix you just suggested? I put adduniform() at the end of resetdefaults().

scrawl:
Done with notes for now. Not sure why there is a bug, I'll check it out.

Allofich:
Implemented all the notes. I just did some very quick testing but everything (except for the glow bug) seems to be working fine.

I envy you for your ability to see and fix things so quickly. I suppose that comes from experience.

scrawl:
Why do containers glow when I hit them with a fireball?

Allofich:
That's news to me. Let me see if it happens for me, too.

Allofich:
Just a guess, but probably it is from the changes to let open and lock work

Allofich:
Those are in spellcasting.cpp.

Allofich:
Ok, it's fixed.

Allofich:
Well, I'd wanted to avoid code duplication, but I guess it's probably better practice to move the initialization code I added in spellcasting.cpp so that it's only run when it's going to be used. Reuploading the last fix to do this.

scrawl:
This seems to fix all the issues. Some minor cosmetic changes in there too. The main problem was the initial setup of the stateset at the end of addGlow() that should not be done if we already had an existing updater.

```
diff --git a/apps/openmw/mwrender/animation.cpp b/apps/openmw/mwrender/animation.cpp
index bfd5c36..e2585c1 100644
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -1182,8 +1182,8 @@ namespace MWRender
         int mLowestUnusedTexUnit;
     };
 
-    void Animation::addSpellCastGlow(const ESM::MagicEffect *effect){
-
+    void Animation::addSpellCastGlow(const ESM::MagicEffect *effect)
+    {
         osg::Vec4f glowColor(1,1,1,1);
         glowColor.x() = effect->mData.mRed / 255.f;
         glowColor.y() = effect->mData.mGreen / 255.f;
@@ -1198,27 +1198,8 @@ namespace MWRender
 
     void Animation::addGlow(osg::ref_ptr<osg::Node> node, osg::Vec4f glowColor, float glowDuration)
     {
-        std::vector<osg::ref_ptr<osg::Texture2D> > textures;
-        for (int i=0; i<32; ++i)
-        {
-            std::stringstream stream;
-            stream << "textures/magicitem/caust";
-            stream << std::setw(2);
-            stream << std::setfill('0');
-            stream << i;
-            stream << ".dds";
-
-            osg::ref_ptr<osg::Image> image = mResourceSystem->getImageManager()->getImage(stream.str());
-            osg::ref_ptr<osg::Texture2D> tex (new osg::Texture2D(image));
-            tex->setName("envMap");
-            tex->setWrap(osg::Texture::WRAP_S, osg::Texture2D::REPEAT);
-            tex->setWrap(osg::Texture::WRAP_T, osg::Texture2D::REPEAT);
-            mResourceSystem->getSceneManager()->applyFilterSettings(tex);
-            textures.push_back(tex);
-        }     
-
         if (mGlowUpdater && mGlowUpdater->isDone())
-                node->removeUpdateCallback(mGlowUpdater);
+            node->removeUpdateCallback(mGlowUpdater);
 
         FindLowestUnusedTexUnitVisitor findLowestUnusedTexUnitVisitor;
         node->accept(findLowestUnusedTexUnitVisitor);
@@ -1230,25 +1211,43 @@ namespace MWRender
             mGlowUpdater->setDuration(glowDuration);
         }
         else
-        {    
+        {
+            std::vector<osg::ref_ptr<osg::Texture2D> > textures;
+            for (int i=0; i<32; ++i)
+            {
+                std::stringstream stream;
+                stream << "textures/magicitem/caust";
+                stream << std::setw(2);
+                stream << std::setfill('0');
+                stream << i;
+                stream << ".dds";
+
+                osg::ref_ptr<osg::Image> image = mResourceSystem->getImageManager()->getImage(stream.str());
+                osg::ref_ptr<osg::Texture2D> tex (new osg::Texture2D(image));
+                tex->setName("envMap");
+                tex->setWrap(osg::Texture::WRAP_S, osg::Texture2D::REPEAT);
+                tex->setWrap(osg::Texture::WRAP_T, osg::Texture2D::REPEAT);
+                mResourceSystem->getSceneManager()->applyFilterSettings(tex);
+                textures.push_back(tex);
+            }
+
             osg::ref_ptr<GlowUpdater> glowUpdater = new GlowUpdater(texUnit, glowColor, textures, node, glowDuration, mResourceSystem);
             mGlowUpdater = glowUpdater;
             node->addUpdateCallback(glowUpdater);
-        }
 
-        // set a texture now so that the ShaderVisitor can find it
-        osg::ref_ptr<osg::StateSet> writableStateSet = NULL;
-        if (!node->getStateSet())
-            writableStateSet = node->getOrCreateStateSet();
-        else
-        {
-            writableStateSet = osg::clone(node->getStateSet(), osg::CopyOp::SHALLOW_COPY);
-            node->setStateSet(writableStateSet);
+            // set a texture now so that the ShaderVisitor can find it
+            osg::ref_ptr<osg::StateSet> writableStateSet = NULL;
+            if (!node->getStateSet())
+                writableStateSet = node->getOrCreateStateSet();
+            else
+            {
+                writableStateSet = osg::clone(node->getStateSet(), osg::CopyOp::SHALLOW_COPY);
+                node->setStateSet(writableStateSet);
+            }
+            writableStateSet->setTextureAttributeAndModes(texUnit, textures.front(), osg::StateAttribute::ON);
+            writableStateSet->addUniform(new osg::Uniform("envMapColor", glowColor));
+            mResourceSystem->getSceneManager()->recreateShaders(node);
         }
-        writableStateSet->setTextureAttributeAndModes(texUnit, textures.front(), osg::StateAttribute::ON);
-        writableStateSet->addUniform(new osg::Uniform("envMapColor", glowColor));
-
-        mResourceSystem->getSceneManager()->recreateShaders(node);
     }
 
     // TODO: Should not be here
```

Allofich:
Are you going to apply that on top of this PR, or do you want me to do it? I don't know how to do it except manually reading it and typing it in.

Allofich:
Well I'll go ahead and start. Just say if you want me to stop.

scrawl:
Just apply it manually. There might be merge conflicts anyway since you made some more cleanups after I did my changes.

Allofich:
OK.

Allofich:
Finished typing it in. Giving it a test now.

Allofich:
Seems to work fine! I'll upload now.

scrawl:
I am sitll seeing that problem with the texUnit number climbing by one each time. Looks like removeTextureAttribute() doesn't reduce the size of the _textureAttributeList.

Allofich:
Yes, I see the problem happening, too. Well, I guess you know how I handled that problem, by using resize() on the list to reduce it by one when the glowupdater finished, or by getting the texUnit from glowupdater stored in mGlowUpdater and reusing it the next time addGlow() is called.

Allofich:
There are no merge conflicts so far, btw.

scrawl:
Here is a method I wrote to properly remove the texture and then trim the _textureAttributeList of empty entries. When testing it I realized that we also forgot to remove the TEXENV attribute.

```
        void removeTexture(osg::StateSet* stateset)
        {
            stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXTURE);
            stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXGEN);
            stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXENV);
            stateset->removeTextureMode(mTexUnit, GL_TEXTURE_2D);

            osg::StateSet::TextureAttributeList& list = stateset->getTextureAttributeList();
            while (list.size() && list.rbegin()->empty())
                list.pop_back();
        }
```

Allofich:
So, call that in setDefaults() in if (mDone) instead of

stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXTURE);
stateset->removeTextureAttribute(mTexUnit, osg::StateAttribute::TEXGEN);

right?

scrawl:
The uniform can be removed, too.

Allofich:
Just making sure. You mean add
stateset->removeUniform("envMapColor");
to the removeTexture() method, right?

scrawl:
Correct.

Allofich:
Ok, done. Will give it a quick test on my machine now.

Allofich:
Seems to be working fine. Tried it both with shaders enabled and without.

scrawl:
I believe there is a memory leak / circular reference. The Node holds the GlowUpdater and the GlowUpdater holds a ref_ptr to that Node. 

Suggested fix:
1. Use a raw pointer
or 2. Change the StateSetUpdater interface so that the node is passed to apply()

I think I'd prefer 1. to reduce the amount of changes to the rest of the codebase.

Allofich:
Thanks for all your hard work in fixing this up. I'll keep an eye on the automatic builds as they catch up and try to fix any warnings or errors.

Allofich:
Can you confirm that this is the fix you wanted?

scrawl:
Looks good. Will give it a final test and then merge.

scrawl:
Or not. There is a problem with the use of addGlow() in NpcAnimation. The new addGlow() implementation doesn't work properly for equipment parts. The function may be called multiple times with different nodes.

Allofich:
That's for bounded equipment, right?

Allofich:
Bounded equipment has no glow.

Allofich:
Just with some very preliminary testing, it seems like changing line 1218 from

if (mGlowUpdater && mGlowUpdater->isPermanentGlowUpdater())
to
if ((glowDuration >= 0) && mGlowUpdater && mGlowUpdater->isPermanentGlowUpdater())

seems to fix summoning bound equipment...

scrawl:
I would factor out the  ```else { ... }``` portion of the addGlow into a new method, then call only that new method from NpcAnimation.

Allofich:
OK. Give me a moment.

Allofich:
I'm making the else portion alone be addGlow(), and merging the parts before that into addSpellCastGlow()

Allofich:
I did this in a little bit of a rush, but everything seemed fine in testing.

scrawl:
Can we close https://bugs.openmw.org/issues/1645 ?

Allofich:
We have the cast sound and the glow, but we still need the effect (sparkles in the case of that feature request). So I'd say that the request is half or more completed, but it can't be closed quite yet.

PR #1004, opened 2016-07-20
Allow data paths to also use hash characters (follow up from #998)

AnyOldName3:
I realised I missed something in https://github.com/OpenMW/openmw/pull/998 - data paths weren't being 'unescaped' like all other variables, leading to even more confusing behaviour than before it was merged in this (hypothetical) situation.

Unfortunately, I'm having issues getting this to build which I don't really understand, which will be explained further in a line comment below.

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -93,7 +93,7 @@ IF(NOT WIN32 AND NOT APPLE)
     add_definitions(-DGLOBAL_CONFIG_PATH="${GLOBAL_CONFIG_PATH}")
 ENDIF()
 add_component_dir (files
-    linuxpath androidpath windowspath macospath fixedpath multidircollection collections configurationmanager
+    linuxpath androidpath windowspath macospath fixedpath multidircollection collections configurationmanager escape
     lowlevelfile constrainedfilestream memorystream
     )
 
--- a/components/fallback/validate.hpp
+++ b/components/fallback/validate.hpp
@@ -70,21 +70,31 @@ namespace Files {
             eSV->mVector.push_back(EscapeHashString(*it));
     }
 
-    struct EscapePathContainer {
-        PathContainer mContainer;
+    struct EscapePath {
+        boost::filesystem::path mPath;
+
+        static PathContainer toPathContainer(const std::vector<EscapePath> & escapePathContainer);
     };
 
-    std::istream & operator>> (std::istream & istream, EscapePathContainer & escapePathContainer)
+    typedef std::vector<EscapePath> EscapePathContainer;
+
+    PathContainer EscapePath::toPathContainer(const EscapePathContainer & escapePathContainer)
+    {
+        PathContainer temp;
+        for (EscapePathContainer::const_iterator it = escapePathContainer.begin(); it != escapePathContainer.end(); ++it)
+            temp.push_back(it->mPath);
+        return temp;
+    }
+
+    std::istream & operator>> (std::istream & istream, EscapePath & escapePath)
     {
         std::cout << "The new dodgy operator>> is being used" << std::endl;
 
         boost::iostreams::filtering_istream filteredStream;
-        filteredStream.push(unescape_hash_filter());
+        //filteredStream.push(unescape_hash_filter());
         filteredStream.push(istream);
 
-        boost::filesystem::path path;
-        filteredStream >> path;
-        escapePathContainer.mContainer.push_back(path);
+        filteredStream >> escapePath.mPath;
 
         return istream;
     }
--- a/components/fallback/validate.hpp
+++ b/components/fallback/validate.hpp
@@ -2,6 +2,8 @@
 #define OPENMW_COMPONENTS_FALLBACK_VALIDATE_H
 
 #include <boost/program_options.hpp>
+#include <boost/filesystem/path.hpp>
+#include <boost/iostreams/filtering_stream.hpp>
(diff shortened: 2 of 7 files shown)


components/fallback/validate.hpp
```diff
@@ -66,6 +69,33 @@ namespace Files {
...
+        PathContainer temp;
+        for (EscapePathContainer::const_iterator it = escapePathContainer.begin(); it != escapePathContainer.end(); ++it)
+            temp.push_back(it->mPath);
+        return temp;
+    }
+
+    std::istream & operator>> (std::istream & istream, EscapePath & escapePath)
+    {
+        boost::iostreams::filtering_istream filteredStream;
+        filteredStream.push(unescape_hash_filter());
```
AnyOldName3: This is where we add a filter to restore the contents of the input stream to its former state (as it's now made it through boost's comment ignoring system safely). However, this causes an "unresolved external symbol" error. Using an ```escape_hash_filter``` instead works fine, even though both filters are (as far as I understand) declared and defined in the same place.

AnyOldName3:
Further testing has revealed that ```unescape_hash_filter``` can be used in configurationmanager.cpp without issue, which makes this seem even weirder to me.

components/files/configurationmanager.cpp
```diff
@@ -145,6 +145,9 @@ bool ConfigurationManager::loadConfig(const boost::filesystem::path& path,
         boost::iostreams::filtering_istream configFileStream;
         configFileStream.push(escape_hash_filter());
         configFileStream.push(configFileStreamUnfiltered);
+
+        boost::iostreams::filtering_istream dummyStreamThatShouldMakeOtherCodeCompile;
+        dummyStreamThatShouldMakeOtherCodeCompile.push(unescape_hash_filter());
```
AnyOldName3: Adding this here as a stab in the dark made it work somehow.

AnyOldName3:
I've 'fixed' it by making boost use the filter somewhere else where it was even more in scope, and then it magically compiled when I used it where I wanted to all along. I find this strange and it worries me. It is also very probably a nausea-inducing hack.

components/files/configurationmanager.hpp
```diff
@@ -88,6 +88,17 @@ struct escape_hash_filter : public boost::iostreams::input_filter
         bool mFinishLine;
 };
 
+struct unescape_hash_filter : public boost::iostreams::input_filter
+{
+    unescape_hash_filter();
+    virtual ~unescape_hash_filter();
+    
+    template <typename Source> int get(Source & src);
```
Aesylwinn: I would guess that your template definition is causing some issues. Maybe try moving your implementation to the header file.

AnyOldName3:
Moving the implementation to the header, as @Aesylwinn suggested, and then changing things until it was happy with that, has allowed the whole thing to build without the hack. I think this is ready to merge now.

AnyOldName3:
Okay, so the CI tests show that moving stuff to the header only seems to be viable on Windows, not Linux or OSX. Great...

AnyOldName3:
That last commit seems to have fixed the issue. *Now* this should be ready to merge.

components/files/configurationmanager.cpp
```diff
@@ -164,6 +164,10 @@ bool ConfigurationManager::loadConfig(const boost::filesystem::path& path,
     return false;
 }
 
+const int escape_hash_filter::sEscape = '@';
+const int escape_hash_filter::sEscapeIdentifier = 'a';
+const int escape_hash_filter::sHashIdentifier = 'h';
```
terrorfisch: This is the fix for the build error. Have a look at the accepted answer here:

http://stackoverflow.com/questions/5391973/undefined-reference-to-static-const-int

zinnschlag:
Opps. There was a problem in the previous commit that I didn't catch at first and this one seems to make it worse.
In validate.hpp you define several non-template functions outside of class definitions. This will break the build if validate.hpp is ever included in more than one cpp file. You could flag them as inline, but the proper way to do it, is to move the implementation to a cpp file.

AnyOldName3:
I'm camping right now, so can't resolve this, but there are a few other things I feel need tweaking - for a start, a load of stuff in fallback/validate.xpp should probably be in a new file files/validate.xpp and the stuff I put in files/configurationmanager.xpp should be in a new files/escape.xpp.

AnyOldName3:
The last commit *should* have resolved the issues @zinnschlag brought up, but Visual Studio decided to be weird and so although this compiled on my machine, the CI systems really don't seem to like it. Can someone who knows how I'm supposed to add files explain what I'm supposed to do?

components/CMakeLists.txt
```diff
@@ -93,7 +93,7 @@ IF(NOT WIN32 AND NOT APPLE)
     add_definitions(-DGLOBAL_CONFIG_PATH="${GLOBAL_CONFIG_PATH}")
 ENDIF()
 add_component_dir (files
-    linuxpath androidpath windowspath macospath fixedpath multidircollection collections configurationmanager
+    linuxpath androidpath windowspath macospath fixedpath multidircollection collections configurationmanager escape
```
AnyOldName3: On my machine, this seemed to be enough to make everything find the new file once CMake was re-run, but the CI systems can't.

AnyOldName3:
Latest commit has genuinely fixed everything, so this is probably ready for review again.

PR #1005, opened 2016-07-22
Various reference handling fixes

MiroslavR:
Fixes bug: https://bugs.openmw.org/issues/3471

Most of the remnant references mentioned in the bug report are moved references that another plugin is deleting - as suspected. The rest are regular references that had their refID changed to an object of a different category (e.g. from a static to an activator), and then deleted afterwards by another plugin. Vanilla MW seems to support refID changes, however the TES CS does not seem to provide a way to achieve that.


MiroslavR:
> The refId replacement support should also fix https://bugs.openmw.org/issues/3085 , correct?

Correct. Didn't know about this bug report.

PR #1007, opened 2016-07-30
Save scripted animation state

MiroslavR:
This stores scripted animation state in saved games as well as allows it to persist after the animated ref's cell is unloaded.

Fixes:
- https://bugs.openmw.org/issues/1931
- https://bugs.openmw.org/issues/2150
- https://bugs.openmw.org/issues/3393

### DIFF
--- a/apps/essimporter/convertacdt.cpp
+++ b/apps/essimporter/convertacdt.cpp
@@ -1,3 +1,9 @@
+#include <string>
+#include <iostream>
+#include <limits>
+
+#include <components/misc/stringops.hpp>
+
 #include "convertacdt.hpp"
 
 namespace ESSImport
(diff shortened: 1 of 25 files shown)


zinnschlag:
Interesting I expected that to be state that is restricted to the current scene and not require persistence over cell unload/load. Apparently not.

Now the question is, does this also cover https://bugs.openmw.org/issues/1225 ? I am not aware of any other graphical state that needs saving

@scrawl Can you confirm there is no scene-only state in the AI either that still needs saving?

MiroslavR:
> Now the question is, does this also cover https://bugs.openmw.org/issues/1225 ? I am not aware of any other graphical state that needs saving

This PR only covers scripted animations (played via _PlayGroup_, _LoopGroup_ instructions). Looks like there is still magic VFX that needs to be saved.

MiroslavR:
There is still some work I want to do regarding this PR. Please do not merge yet.

MiroslavR:
This should be ready for review now. The previous implementation was a buggy mess. I have also thrown in partial support for importing this state from vanilla savegames.

Allofich:
I still find that Akulakhan is put back into its pre-destroyed state if I load a saved game, even one made with the latest master, after or during the death animation. Is that supposed to have been fixed by this PR? The bridge also comes back, although that also happens in the original game.

Allofich:
Here are two saves made using a recent master from a few days ago. anim1 was taken during the death animation, and anim2 was taken after it finished.
[anim1.zip](https://github.com/OpenMW/openmw/files/474358/anim1.zip)
[anim2.zip](https://github.com/OpenMW/openmw/files/474359/anim2.zip)

MiroslavR:
Thanks, should be fixed by #1060.

PR #1008, opened 2016-08-01
Fixes Bug 3451 - Invalid Enchantment IDs

Internecine:
https://bugs.openmw.org/issues/3451

Added exception handlers when trying to retreive enchantment effects of items.


scrawl:
I prefer
<pre>
a = search(id);
if (a)
   // do stuff
</pre>

instead of
<pre>
try
   a = find(id)
   // do stuff
catch (...)
   return
</pre>

Internecine:
Agreed, I have replaced Store::find with Store::search.

PR #1009, opened 2016-08-02
OpenMW-CS: Camera resets every time cell->view is pressed

lohikaarme:
Camera didn't reset when right clicking cell and pressing view if world renderer was open.

PS: Shouldn't "const int cellSize = 8192" be some where it can be accessed easily rather than be local variable.

### DIFF
--- a/apps/opencs/model/world/cellcoordinates.hpp
+++ b/apps/opencs/model/world/cellcoordinates.hpp
@@ -37,10 +37,9 @@ namespace CSMWorld
             ///
             /// \note The worldspace part of \a id is ignored
             static std::pair<CellCoordinates, bool> fromId (const std::string& id);
-
+            
             /// \return cell coordinates such that given world coordinates are in it.
             static std::pair<int, int> coordinatesToCellIndex (float x, float y);
-            
     };
 
     bool operator== (const CellCoordinates& left, const CellCoordinates& right);
--- a/apps/opencs/view/render/pagedworldspacewidget.hpp
+++ b/apps/opencs/view/render/pagedworldspacewidget.hpp
@@ -84,7 +84,7 @@ namespace CSVRender
             /// hint system.
 
             virtual ~PagedWorldspaceWidget();
-
+            
             /// Decodes the the hint string to set of cell that are rendered.
             void useViewHint (const std::string& hint);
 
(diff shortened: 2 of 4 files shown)


zinnschlag:
It seems that this won't build and won't merge cleanly either.

Apparent from that could you elaborate which problem you are trying to fix? Its not clear to me from the issue description.

Regarding cellSize: Unfortunately we have these constants all over the place. The proper way would be to have a single constant in ESM::Cell. If you want to go through the code and collect all the constants in one place, feel free to send another pull request (after this one has been sorted out).

lohikaarme:
I don't think there is open issue for it. Here is video showing what I am talking about:
https://www.youtube.com/watch?v=xWouSOxKROY

When _mCamPositionSet_ isn't set to false the view will not move the camera to newly open cell if there is already other cell open. After _mCamPositionSet_ is set to false then rendering update will move the camera near that cell. 

Is there any build log from AppVeyor?

MiroslavR:
> Is there any build log from AppVeyor?

`Failure: AppVeyor was unable to build non-mergeable pull request`

You need to resolve the conflicts.

Aesylwinn:
> Regarding cellSize: Unfortunately we have these constants all over the place. The proper way would be to have a single constant in ESM::Cell. If you want to go through the code and collect all the constants in one place, feel free to send another pull request (after this one has been sorted out).

The constant is already defined in [components/esm/loadland.hpp](https://github.com/OpenMW/openmw/blob/master/components/esm/loadland.hpp#L56)

zinnschlag:
> The constant is already defined in components/esm/loadland.hpp

Not really the correct place for it.

zinnschlag:
> I don't think there is open issue for it. Here is video showing what I am talking about:
https://www.youtube.com/watch?v=xWouSOxKROY

I see. You have the editor configured to re-use existing 3D scene subviews. That explains it.

lohikaarme:
It would make more sense be attribute of cell rather than it's loader.

lohikaarme:
> I see. You have the editor configured to re-use existing 3D scene subviews. That explains it.

Configured by how? Macro or option settings inside of the editor?

zinnschlag:
There is a user setting for that. I think under the Windows tab.

scrawl:
No offense but that is the messiest commit log I have ever seen. Please do a rebase next time.

lohikaarme:
Yeah sorry about that :(

PR #1010, opened 2016-08-05
Render water in editor.

Aesylwinn:
This implements [feature#1596.](https://bugs.openmw.org/issues/1596)

I wasn't sure what the best way to go about keeping up to date with the cell data. It seemed silly to add yet another slot to the WorldSpaceWidget class that would have to be inherited by both the Paged/Unpaged child classes and passed on to the correct cell. I can change it to do that if desired.

@scrawl 
I was having issues with the lighting flickering when I moved or looked around in certain spots, so I had to disable it. Any idea what could be causing that? Also, should I be using higher numbers for the rendering bins, or will 1000 and 2000 be fine?

### DIFF
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -87,6 +87,7 @@ opencs_units (view/render
     scenewidget worldspacewidget pagedworldspacewidget unpagedworldspacewidget
     previewwidget editmode instancemode instanceselectionmode instancemovemode
     orbitcameramode pathgridmode selectionmode pathgridselectionmode cameracontroller
+    cellwater
     )
 
 opencs_units_noqt (view/render
--- a/apps/opencs/view/render/cell.cpp
+++ b/apps/opencs/view/render/cell.cpp
@@ -16,6 +16,7 @@
 #include "../../model/world/refcollection.hpp"
 #include "../../model/world/cellcoordinates.hpp"
 
+#include "cellwater.hpp"
 #include "mask.hpp"
 #include "pathgrid.hpp"
 #include "terrainstorage.hpp"
@@ -111,6 +112,7 @@ CSVRender::Cell::Cell (CSMWorld::Data& data, osg::Group* rootNode, const std::st
         }
 
         mPathgrid.reset(new Pathgrid(mData, mCellNode, mId, mCoordinates));
+        mCellWater.reset(new CellWater(mData, mCellNode, mId, mCoordinates));
     }
 }
 
--- a/apps/opencs/view/render/cell.hpp
+++ b/apps/opencs/view/render/cell.hpp
@@ -35,6 +35,7 @@ namespace CSMWorld
 
 namespace CSVRender
 {
+    class CellWater;
     class Pathgrid;
     class TagBase;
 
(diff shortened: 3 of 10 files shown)


Aesylwinn:
I chose a grayish blue color for the water that I thought closely matched OpenMW's non-shader water. Maybe I should be using a texture?

Aesylwinn:
I think osg may be using vertex lighting.

scrawl:
Looks like you are not using the cell's water height yet. Btw, the water height for exterior cells must be set to -1, ignoring the Cell's water height value. When this is corrected you shouldn't need the polygon offset.

Point lights aren't going to work correctly with 1 big mesh as each mesh can be lit by a maximum of 8 lights. I believe the vanilla engine solves this with a quad-tree paging solution that splits the water into smaller meshes. This isn't implemented in OpenMW yet. For now just disable the point-lights:
<pre>
// keeps light 0 (sunlight) enabled
for (int i=1; i<8; ++i)
    stateset->setMode(GL_LIGHT0+i, osg::StateAttribute::OFF);
</pre>

For the sunlight to work correctly, your mesh is missing normals.

The mesh will need to be tesselated for vertex fog (and vertex lighting) to work correctly.

I suggest moving some code from OpenMW into a component that you can re-use (building the mesh, material, texture). In particular the [createSimpleWaterStateSet](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwrender/water.cpp#L528) and [createWaterGeometry](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwrender/water.cpp#L46).

For the renderbin number, OpenMW is using TRANSPARENT_BIN-1 (i.e. the water is rendered before all other transparent meshes). This isn't a perfect solution, but it looks better than TRANSPARENT_BIN+1 and better than TRANSPARENT_BIN (too much depth fighting going on). See bug [#2633](https://bugs.openmw.org/issues/2633). A solution is pending that is more flexible than a fixed renderbin number.

Aesylwinn:
Thanks, that was quite helpful. I can't believe I was forgetting normals. I'll go ahead and move the relevant bits to the component library.

Aesylwinn:
I'm going to need to figure out how resources work in the CS if I'm going to use the water texture.

scrawl:
The CSVRender::Cell can get you the ResourceSystem:
<pre>
ResourceSystem* resourceSystem = mData.getResourceSystem().get();

...

ImageManager* imageMgr = resourceSystem->getImageManager();
osg::ref_ptr<osg::Texture2D> waterTexture (imageMgr->getImage("water.png"));
</pre>

Aesylwinn:
Awesome. I'll get on that.

I don't think I need to ignore the point lights anymore. The the ambient light is getting set to full blast and some testing would suggest that the point lights no longer have any effect on the water.

scrawl:
> I don't think I need to ignore the point lights anymore. The the ambient light is getting set to full blast and some testing would suggest that the point lights no longer have any effect on the water.

That might not be the case when lighting is set to Night mode and the scene ambient is much lower.

Actually, I just remembered that point lights are not even enabled by default. Only by adding a [LightListCallback](https://github.com/OpenMW/openmw/blob/master/components/sceneutil/lightmanager.hpp#L149) to the object (or a parent node of the object) the point lighting is enabled. So there is nothing to be done here (no need to explicitely ignore point lights).

scrawl:
Are we done here?

Aesylwinn:
I did wonder what @zinnschlag would have to say about the way I was keeping the data up to date. I think in the future a couple of things may need to be reworked so that the WorldspaceWidget class is only responsible for adding/removing cells and the cells are hooked up so that they will keep their components up to date.

For now though, I believe it's done.

zinnschlag:
Will give it a review either today or tomorrow.

PR #1012, opened 2016-08-08
Fix pointer being deleted twice.

Aesylwinn:
This fixes [bug#3483.](https://bugs.openmw.org/issues/3483) It was caused by both Qt's memory management and an std::auto_ptr keeping track of the same pointer. I missed it after one of my redesigns.


scrawl:
@zinnschlag this should be cherry picked to 0.40 branch, right?

zinnschlag:
Done

PR #1015, opened 2016-08-10
Don't allow same spell sound to stack on a single object (don't merge)

Allofich:
Fix for https://bugs.openmw.org/issues/3490.

Details of how I tested and came to the conclusion of what to do can be seen by reading my comments there.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -613,8 +613,6 @@ namespace MWMechanics
                     }
                     target.getClass().unlock(target);
                 }
-                else
-                    MWBase::Environment::get().getSoundManager()->playSound3D(target, "Open Lock Fail", 1.f, 1.f);
                 return true;
             }
         }
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -737,6 +737,8 @@ namespace MWMechanics
                     "alteration", "conjuration", "destruction", "illusion", "mysticism", "restoration"
                 };
                 MWBase::SoundManager *sndMgr = MWBase::Environment::get().getSoundManager();
+                if (sndMgr->getSoundPlaying(mCaster, "Spell Failure " + schools[school]))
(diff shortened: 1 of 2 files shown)


Allofich:
I think this covers all the cases now. We could also check for the cast sound for consistency, but I'm not aware that it's possible to play the cast sound repeatably.

Allofich:
I took out the locked sound when open fails because it's not there in the original game (and because it seems a little strange to have a lock and key type sound when you are casting magic), but it looks like there is the lock open success sound in the original game. Should we keep the lock failed sound in OpenMW? I guess I would still prefer to remove it, just to mimic the original game.

Allofich:
Well... I'm now a little unsure about whether to go through with this PR.

A big part of the motivation for me to do so was that sounds overlapping on each other sounded quite bad to me (quite loud, almost with the sound popping). However, after changing to the OpenAL referenced here
https://bugs.openmw.org/issues/3381
it sounds significantly better. I'm no longer sure whether it's desirable to change the sound behavior. But maybe there are other benefits? @scrawl or anyone, any thoughts?

Allofich:
A good place to test (minor spoiler) is "ald daedroth." You should see a noisy fight with the destruction spell sound effect going on in front of you. I'm actually thinking now that it sounds better without this PR, as it makes the effect sound keep starting over every time it hits, same as the original game/

scrawl:
Not sure. I guess that is a more general issue not limited to spellcasting sounds.

Unless there are some problems with excessive clipping/static, then I think playing sound multiple times is useful - conveys more information.

This could be user configurable - "stack sounds" option? Just not sure what the best default would be.

Allofich:
> I guess that is a more general issue not limited to spellcasting sounds.

That does seem like it would be the case. I don't know what, other than spells/potions/enchantments, would repeatedly play a sound on an actor, though. Maybe some elaborate scripting?

This PR makes the game sound more like the original engine, but as long as the newer OpenAL version is used, I think that allowing overlapping sounds like in the OpenMW master gives a more natural-sounding experience, so I'm going to close this PR. Anyway it's an easy change if we want to make this a configurable option. Oh and I'm not so sure about removing the "open lock failed" sound of the Open spell, either, as leaving it in might make more sense, so I'm not planning now on re-submitting that.

PR #1016, opened 2016-08-10
Editor rotate/scale instances

Aesylwinn:
It seems a lot of the ground work was done when the "move instances" feature was completed. The correct Euler to quaternion conversion was a pain to find though. ;)

I tried to use blender as a base for how the controls work. Let me know how you like them.

### DIFF
--- a/apps/opencs/model/prefs/state.cpp
+++ b/apps/opencs/model/prefs/state.cpp
@@ -188,6 +188,7 @@ void CSMPrefs::State::declare()
             "Shift-acceleration factor during drag operations", 4.0).
         setTooltip ("Acceleration factor during drag operations while holding down shift").
         setRange (0.001, 100.0);
+    declareDouble ("rotate-factor", "Free rotation factor", 0.007).setPrecision(4).setRange(0.0001, 0.1);
 
     declareCategory ("Tooltips");
     declareBool ("scene", "Show Tooltips in 3D scenes", true);
--- a/apps/opencs/view/render/instancemode.cpp
+++ b/apps/opencs/view/render/instancemode.cpp
@@ -215,23 +215,26 @@ bool CSVRender::InstanceMode::primaryEditStartDrag (const QPoint& pos)
     if (mDragMode!=DragMode_None || mLocked)
         return false;
 
-
     WorldspaceHitResult hit = getWorldspaceWidget().mousePick (pos, getWorldspaceWidget().getInteractionMask());
-    if (hit.tag && CSMPrefs::get()["3D Scene Input"]["context-select"].isTrue())
-    {
-        getWorldspaceWidget().clearSelection (Mask_Reference);
-        if (CSVRender::ObjectTag *objectTag = dynamic_cast<CSVRender::ObjectTag *> (hit.tag.get()))
-        {
-            CSVRender::Object* object = objectTag->mObject;
-            object->setSelected (true);
-        }
-    }
 
-    std::vector<osg::ref_ptr<TagBase> > selection =
-        getWorldspaceWidget().getSelection (Mask_Reference);
-
+    std::vector<osg::ref_ptr<TagBase> > selection = getWorldspaceWidget().getSelection (Mask_Reference);
     if (selection.empty())
-        return false;
+    {
+        // Only change selection at the start of drag if no object is already selected
+        if (hit.tag && CSMPrefs::get()["3D Scene Input"]["context-select"].isTrue())
+        {
+            getWorldspaceWidget().clearSelection (Mask_Reference);
+            if (CSVRender::ObjectTag *objectTag = dynamic_cast<CSVRender::ObjectTag *> (hit.tag.get()))
+            {
+                CSVRender::Object* object = objectTag->mObject;
+                object->setSelected (true);
+            }
+        }
+
+        selection = getWorldspaceWidget().getSelection (Mask_Reference);
+        if (selection.empty())
+            return false;
+    }
 
     for (std::vector<osg::ref_ptr<TagBase> >::iterator iter (selection.begin());
         iter!=selection.end(); ++iter)
--- a/apps/opencs/view/render/instancemode.cpp
+++ b/apps/opencs/view/render/instancemode.cpp
@@ -27,6 +27,38 @@ int CSVRender::InstanceMode::getSubModeFromId (const std::string& id) const
     return id=="move" ? 0 : (id=="rotate" ? 1 : 2);
 }
 
+osg::Vec3f CSVRender::InstanceMode::quatToEuler(const osg::Quat& rot) const
+{
+    const float Pi = 3.14159265f;
+
+    float x, y, z;
+    float test = 2 * (rot.w() * rot.y() + rot.x() * rot.z());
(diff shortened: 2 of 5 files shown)


Aesylwinn:
The links: [feature#3166](https://bugs.openmw.org/issues/3166) and [feature#3167](https://bugs.openmw.org/issues/3167)

zinnschlag:
A couple of problems:
- When rotating and moving the mouse very far away from the rotated object, the selection sometimes jumps to a different object
- When rotating freely, it seems we can only rotate around a single axis.
- When scaling "freely" (not using the arrows) sometimes moving the mouse towards the object seems to scale the object up instead of down.

Aesylwinn:
> When rotating and moving the mouse very far away from the rotated object, the selection sometimes jumps to a different object

I've never experienced that myself and am unable to reproduce it. Can you provide more details?

> When rotating freely, it seems we can only rotate around a single axis.

I think that's not a problem? That's how I designed it. The object will rotate around the camera's backward axis when rotating freely. The system of control I'm using will only work with one axis and allows you to vary the rotation speed and precision based on how far the cursor is from the center of the screen. For what it's worth, that is how blender does it as well.

> When scaling "freely" (not using the arrows) sometimes moving the mouse towards the object seems to scale the object up instead of down.

The system scales towards the center of the screen, not the "center" of the object. I could implement that, but how should I handle cases where the object is off-screen, or multiple objects are selected? I could add an indicator for the rotate/scale modes that would make it more obvious that the controls work around the center of the screen.

Aesylwinn:
I looked through my code. There is nothing that should be modifying the selection. Maybe a corrupt build or a preexisting issue?

zinnschlag:
> I've never experienced that myself and am unable to reproduce it. Can you provide more details?

Happens more or less consistently in my test setup. I am not sure what factors affects this problem, so I'll give you an info dump: Relatively scene sub-view window. Context sensitive selection mode is enabled. Freely rotating an object with the mouse, then move the mouse close to the edge of the window (but not across). The selection will jump to a new object under the mouse which is then rotated. Interestingly when you move the mouse back to selection usually also jumps back to the original object.

> I looked through my code. There is nothing that should be modifying the selection. Maybe a corrupt build or a preexisting issue?

A pre-existing issue is entirely possible, Maybe a another feature interacting with this one?

> I think that's not a problem? That's how I designed it. The object will rotate around the camera's backward axis when rotating freely. The system of control I'm using will only work with one axis and allows you to vary the rotation speed and precision based on how far the cursor is from the center of the screen. For what it's worth, that is how blender does it as well.

Very unexpected behaviour in my opinion. But that is just my opinion. Maybe query the forum about this topic?

> The system scales towards the center of the screen, not the "center" of the object. I could implement that, but how should I handle cases where the object is off-screen, or multiple objects are selected? I could add an indicator for the rotate/scale modes that would make it more obvious that the controls work around the center of the screen.

Centre of the screen is really unintuitive. I don't think this will do. How about this:
Calculate the centre of the selection. Then calculate the distance of the mouse pointer from this centre. Increasing the distance scales up, decreasing it scales down.

Aesylwinn:
> Happens more or less consistently in my test setup. I am not sure what factors affects this problem, so I'll give you an info dump: Relatively scene sub-view window. Context sensitive selection mode is enabled. Freely rotating an object with the mouse, then move the mouse close to the edge of the window (but not across). The selection will jump to a new object under the mouse which is then rotated. Interestingly when you move the mouse back to selection usually also jumps back to the original object.

Does it only happen when you start a drag over another object, or will it also happen during a drag? I noticed that the context sensitive selection mode seems to have some issues when starting a drag.

> Centre of the screen is really unintuitive. I don't think this will do. How about this:
Calculate the centre of the selection. Then calculate the distance of the mouse pointer from this centre. Increasing the distance scales up, decreasing it scales down.

An interesting idea. I should probably cap the center to the edge of the screen if it is offscreen. I'll give that a shot.

I [posted](https://forum.openmw.org/viewtopic.php?f=7&t=3709) about the rotation on the forums.

Aesylwinn:
Perhaps there should only be one marker at the center of the selection instead of a marker attached to each selected object?

I decided not to cap the center to the edge of the screen.

zinnschlag:
> Does it only happen when you start a drag over another object, or will it also happen during a drag? I noticed that the context sensitive selection mode seems to have some issues when starting a drag.

Not sure if I understand you correctly. I start a drag operation to rotate an object. During the drag operation the mouse pointer moves over another object. Sometimes this object starts to be the one rotated.

> Perhaps there should only be one marker at the center of the selection instead of a marker attached to each selected object?

That is an interesting idea, but I am sceptical. In a busy scene it would be easy to miss the marker unless you make it so big that it gets in the way.

Aesylwinn:
>That is an interesting idea, but I am sceptical. In a busy scene it would be easy to miss the marker unless you make it so big that it gets in the way.

True. The marker could be made to always render on top, but I'm not sure that raytracing would work correctly with that.

Aesylwinn:
Well, I'm not sure how to handle this. I've enabled the context sensitive selection mode and tried rotating at the edges of the rendered view on Windows (Qt5 only) and Arch Linux (Qt4 and Qt5). However, I've been unable to reproduce another object being selected during a free rotate. I'm pretty sure the only way that could be happening is if false mouse button press/release events are being generated. Maybe I'm misunderstanding something?

I think I'll change the current behavior of selecting the object under the cursor when beginning a drag. Maybe that is the problem being referenced.

zinnschlag:
Can't reproduce the problem anymore.

Aesylwinn:
That's good. Can you verify that the context sensitive select functionality is working correctly? It will now only select an object at the beginning of a drag if no other object is currently selected. I believe that is the expected behavior, and it also conveniently solves the problem of false mouse events being fired.

Aesylwinn:
Well, I tried something different for the free rotation controls. Potentially one of the current axes of rotation could be changed to roll the object. That might make it easier to control.

If this is the direction we want to head towards, do I also need to change the controls for rotating around a marker axis (non-free rotate)?

zinnschlag:
Rotation controls looks good and the context sensitive mode works too. However the original selection change problem is back. I am beginning to suspect that I have a faulty mouse. If I keep a mouse button pressed, but because of some mechanical failure the button down signal would be interrupted for a short moment, that would result in the bug I am seeing, right?
I will get a new mouse tomorrow and then report back.

Aesylwinn:
> If I keep a mouse button pressed, but because of some mechanical failure the button down signal would be interrupted for a short moment, that would result in the bug I am seeing, right?

Yes, it is possible. If it fails two times in quick succession and gets interpreted as a click instead of a drag, the problem would come up again. 

Just a thought, but you could temporarily bind your primary edit button to the keyboard and see if you get the same problem.

zinnschlag:
I am not able to reproduce the problem with the new mouse. And now I am feeling rather silly. Sorry, for wasting your time.

This PR looks good to me. Is there anything left or can we go ahead with the merge.

Aesylwinn:
A context sensitive selection bug was found, so don't worry about it. This PR can be merged.

PR #1019, opened 2016-08-13
Don't allow re-casting of bound equipment spells

Allofich:
For Feature #3489 https://bugs.openmw.org/issues/3489.

First, here's a description of how I found that bound equipment spells work in the original engine.

1. If you cast a bound equipment spell, then re-cast it while it is still going, you get a "You cannot re-cast this spell while it is still in effect." message. Magicka is taken, and you can still advance your conjuration with it.
2. If you cast a spell with two bound equipment effects of different duration, the "you cannot re-cast" message will still appear as long as the longer effect is still going.
3. You can cast different spells that summon the same type of equipment at the same time. You don't get the "you cannot re-cast" message.
4. If the spell has a bound equipment ability and another ability that affects the player, you can re-cast it but this causes the bound equipment to disappear, while the other ability is applied normally. You cannot recreate the equipment with this spell until the other ability has worn off.
5. If the spell has a bound equipment and another ability that does not affect the player (I tested a ranged attack), you can re-cast and the bound equipment becomes permanently attached to the player, even after resting, and you obtain a copy of the equipment in your inventory on every cast.

5 is clearly buggy behavior, and the equipment disappearing in 4 is probably unintended, too. 2 and 3 are maybe arguable. The implementation in this PR is the same as the original game, except that bound equipment is not removed from the player like in 4, and neither do the bugs in 5 happen. This retains the behavior of 4 that the bound equipment cannot be recreated until the other effect on the player ends, which may be arguable, but I was having trouble dealing with individual effects rather than spells and that behavior in the original engine may have been intended, anyway.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -356,6 +356,14 @@ namespace MWMechanics
 
         bool castByPlayer = (!caster.isEmpty() && caster == getPlayer());
 
+        ActiveSpells targetSpells;
+        if (target.getClass().isActor())
+            targetSpells = target.getClass().getCreatureStats(target).getActiveSpells();
+
+        bool canCastAnEffect = false;    // For bound equipment.If this remains false
+                                         // throughout the iteration of this spell's 
+                                         // effects, we display a "can't re-cast" message
+
         for (std::vector<ESM::ENAMstruct>::const_iterator effectIt (effects.mList.begin());
             effectIt!=effects.mList.end(); ++effectIt)
         {
@@ -365,6 +373,16 @@ namespace MWMechanics
             const ESM::MagicEffect *magicEffect =
                 MWBase::Environment::get().getWorld()->getStore().get<ESM::MagicEffect>().find (
                 effectIt->mEffectID);
+
+            // Re-casting a bound equipment effect has no effect if the spell is still active
+            if (magicEffect->mData.mFlags & ESM::MagicEffect::NonRecastable && targetSpells.isSpellActive(mId))
+            {
+                if (effectIt == (effects.mList.end() - 1) && !canCastAnEffect && castByPlayer)
+                    MWBase::Environment::get().getWindowManager()->messageBox("#{sMagicCannotRecast}");
+                continue;
+            }
+            else
+                canCastAnEffect = true;
 
             if (!checkEffectTarget(effectIt->mEffectID, target, castByPlayer))
                 continue;


scrawl:
I think there's a flag in the ESM::MagicEffect for non-recastable effects.

Allofich:
There is. I'll see if I can use that instead, then.

Allofich:
It's working with using that flag in place of the isBoundEffect() method I added, so I removed that method. I'll wait to upload again in case you have other changes you want.

scrawl:
Just out of curiosity, now that the flag is used, what other spells are affected by this? Or are Bound spells the only case?

scrawl:
Have you tested what happens for Permanent effects (e.g. a ring enchanted with permanent "summon bound equipment"?)

Allofich:
Before I did the bug report, I did a quick test of the different magic effects, and I couldn't find anything but bound spells that showed the message. I didn't test every single spell, so it's possible I may have missed something, but I think I tested every lasting effect spell on the player.

I haven't tried permanent effects. I hadn't even thought of that, a permanent bound item. I'll give it a try, and also a regular bound item enchantment, which I forgot to test.

Allofich:
Oh, I haven't tested any spells from Tribunal or Bloodmoon, because believe it or not I haven't played those and don't want to spoil the content for myself. Do they add any kind of new magic effects that might be affected? I assumed that they don't do anything that deep level, just add new combinations of existing effects.

scrawl:
I have just done a esmtool dump of the 3 ESM's. Apparently, the Bound equipment effects are the only effects with the NonRecastable flag.

Allofich:
Thanks. Looks like the behavior isn't quite right for enchanted items, though. I haven't finished testing but this PR doesn't allow the enchantment to be used again after it's been done once. In the original engine you can, although it unequips or re-equips the item to and from your inventory. I'm about to test whether it affects the remaining duration of the spell or not.

Allofich:
Looks like it does reset the spell duration when you re-use the enchantment, so that needs to be implemented.

Allofich:
Permanent enchantment seems to work fine. The original engine does cause an extra pieces of equipment to be created in your inventory when you layer bound spells and enchantments on each other, which OpenMW doesn't do, but I don't think that really matters because the extra bound equipment in the inventory is just unusable clutter, right?

Allofich:
I just tested a scroll with a bound enchantment. You get the message as with spells, and the scroll is used up. We have that same behavior with this PR.

For enchanted items, though, is there an easy way to check whether what is being used in inflict() in spellcasting.cpp is an enchantment rather than a spell? I'm looking around but I see no obvious way yet.

Allofich:
Ahh, silly of me. Looks like the PR already works with enchantments. I didn't notice because the restoration sparkles from the fortify effect didn't play.

Allofich:
Since everything seems to be looking good so far, I'll go ahead and upload the version using the nonrecastable flag to the PR.

Allofich:
I guess there is one minor use of having the a new piece of bound equipment be added when you duplicate a spell. Since the new piece gets equipped, you get the sparkles from the fortify effect again, which makes it a little more visually noticeable that you have refreshed your bound equipment spell. I haven't experimented with that though and there might be downsides to adding the extra pieces of equipment.

PR #1021, opened 2016-08-14
Corrections to tooltip displays of zero-weight items

Allofich:
For https://bugs.openmw.org/issues/3497.

I started on this because I noticed that bound equipment in OpenMW shoes as "heavy" in the inventory tooltip. I found that in the original game, zero-weight armor doesn't show it's weight in the tooltip at all. There is no "unarmored" string in morrowind.esm that would be used for this.

After fixing this, I noticed that all categories of item except for potions don't display zero weight in the tooltip in the original engine. For consistency I implemented the same behavior for OpenMW.

I found that there was already code in OpenMW for lights to not display weight or value if weight is 0, which matches the original engine. Under the category of misc items, keys and gold were set to not display their weight or value ever. I can imagine a mod with keys and gold having weight, and I also found that the original engine uses the same rule for as for lights for misc items, too, that the display of weight and gold depends on whether the weight is 0, not whether objects are specifically gold or keys. I implemented this more general behavior. This makes the tooltip display for "lucky_coin," which was showing its 0 weight in OpenMW, the same as in the original game.

### DIFF
--- a/apps/openmw/mwclass/light.cpp
+++ b/apps/openmw/mwclass/light.cpp
@@ -159,11 +159,9 @@ namespace MWClass
 
         if (Settings::Manager::getBool("show effect duration","Game"))
             text += "\n#{sDuration}: " + MWGui::ToolTips::toString(ptr.getClass().getRemainingUsageTime(ptr));
-        if (ref->mBase->mData.mWeight != 0)
-        {
-            text += "\n#{sWeight}: " + MWGui::ToolTips::toString(ref->mBase->mData.mWeight);
-            text += MWGui::ToolTips::getValueString(ref->mBase->mData.mValue, "#{sValue}");
-        }
+
+        text += MWGui::ToolTips::getWeightString(ref->mBase->mData.mWeight, "#{sWeight}");
+        text += MWGui::ToolTips::getValueString(ref->mBase->mData.mValue, "#{sValue}");
 
         if (MWBase::Environment::get().getWindowManager()->getFullHelp()) {
             text += MWGui::ToolTips::getCellRefString(ptr.getCellRef());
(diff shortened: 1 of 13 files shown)


Allofich:
Minor change.

I noticed that in the original engine the rule of "don't display either weight or value if weight is zero" seems to apply to all items, or at least to lights, misc items, weapons and armor.

Right now I don't see this as being particularly useful other than hiding the value of keys, which were already being hidden in OpenMW, so I changed my mind about applying this rule generally across misc items. Instead, weight will show if it is non-zero, and value will show if it is non-zero, except for gold and keys. The display for "lucky_coin" still matches the original engine this way.

I took out the "don't display either weight or value if weight is zero" rule for lights, as I don't see why it should only apply there. Lights will show their weight or value if either is non-zero, the same as other items.

apps/openmw/mwclass/light.cpp
```diff
@@ -159,11 +159,9 @@ namespace MWClass
 
         if (Settings::Manager::getBool("show effect duration","Game"))
             text += "\n#{sDuration}: " + MWGui::ToolTips::toString(ptr.getClass().getRemainingUsageTime(ptr));
-        if (ref->mBase->mData.mWeight != 0)
-        {
-            text += "\n#{sWeight}: " + MWGui::ToolTips::toString(ref->mBase->mData.mWeight);
-            text += MWGui::ToolTips::getValueString(ref->mBase->mData.mValue, "#{sValue}");
-        }
+
+        text += MWGui::ToolTips::getValueString(ref->mBase->mData.mWeight, "#{sWeight}");
```
MiroslavR: I think you meant to use _getWeightString_ here (despite being equal to _getValueString_ in terms of functionality).

Allofich: Thanks.

PR #1025, opened 2016-08-15
Use loopfallback for idle animation groups

Allofich:
For https://bugs.openmw.org/issues/3499. This is the same as the second minor issue mentioned at the bottom of https://bugs.openmw.org/issues/3385 (the description I gave there of the issue isn't completely accurate.)

This doesn't completely fix 3499. For some reason, idle3 of the heart actor does not loop correctly like the others.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -2023,9 +2023,10 @@ bool CharacterController::playGroup(const std::string &groupname, int mode, int 
             mCurrentIdle.clear();
 
             mIdleState = CharState_SpecialIdle;
+            bool loopfallback = (entry.mGroup.compare(0,4,"idle") == 0);
             mAnimation->play(groupname, Priority_Default,
                              MWRender::Animation::BlendMask_All, false, 1.0f,
-                             ((mode==2) ? "loop start" : "start"), "stop", 0.0f, count-1);
+                             ((mode==2) ? "loop start" : "start"), "stop", 0.0f, count-1, loopfallback);
         }
         else if(mode == 0)
         {
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1542,9 +1542,10 @@ void CharacterController::update(float duration)
                 mAnimation->disable(mAnimQueue.front().mGroup);
                 mAnimQueue.pop_front();
 
+                bool loopfallback = (mAnimQueue.front().mGroup.compare(0,4,"idle") == 0);
                 mAnimation->play(mAnimQueue.front().mGroup, Priority_Default,
                                  MWRender::Animation::BlendMask_All, false,
-                                 1.0f, "start", "stop", 0.0f, mAnimQueue.front().mLoopCount);
+                                 1.0f, "start", "stop", 0.0f, mAnimQueue.front().mLoopCount, loopfallback);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Allofich:
It's not specifically idle3. Using "playgroup idle2" on the heart twice in a row, or "playgroup idle2" and then "playgroup idle4" will also make it not loop.

Allofich:
I see the problem now.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -1542,9 +1542,10 @@ void CharacterController::update(float duration)
                 mAnimation->disable(mAnimQueue.front().mGroup);
                 mAnimQueue.pop_front();
 
+                bool loopfallback = (mAnimQueue.front().mGroup.compare(0,4,"idle") == 0);
```
scrawl: Are you sure we don't need a case-insensitive comparison?

Allofich: I had wondered about that, but it seemed that it was already being converted to lower-case. When I typed "playgroup Idle2" and checked against "Idle," I got a false.

Allofich:
Something that also works more like the original with this PR is loopgroup.

Select an NPC, type "aiwander 0 0 0 0" to make them not move (not completely necessary in the original engine, but in OpenMW movement will interrupt the animation) and then type "loopgroup idle3 3." They will look behind themselves multiple times (4 because of my earlier PR), while they would only do so once before this PR.

I've noticed, though, a tiny thing. Looks like in the original engine actors run the animation of loopgroup idle3 x, "x" times, or at least once, so loopgroup idle3 0 and loopgroup idle3 1 are the same. This is how loopgroup behavior was before my previous PR. So looks like non-actors do loopgroup once more than actors do.

Allofich:
I have the behavior emulated on another branch.
https://github.com/Allofich/openmw/tree/anim2

I guess we should copy the original behavior for compatibility? I can put it in this PR or make a separate one. It undoes the previous PR on loopgroup, btw.

scrawl:
I'm testing out the savegame from bug 3499, hit the heart until it dies, can't spot a difference with your PR compared to master. What am I looking for?

Allofich:
You have to wait a while after hitting for the animation to finish and the loop to begin. When it gets that far, the master should simply halt the animation.

scrawl:
Ah I see it, the shield effect stops spinning after a while.

Allofich:
Right.

Allofich:
By the way, the "explosion effects" when you hit the heart are muted compared to the original game. Any idea off the top of your head what may cause that? Maybe a lighting issue, as the whole area surrounding the heart is more reddish in the original engine than it is in OpenMW? It's probably beyond my ability to work on, though.

scrawl:
There are a multitude of issues with particles not rendering exactly as in the original game. It's mostly a lighting issue, I think. I never quite figured out how the original engine lights particles.

Allofich:
OK, well, I guess it probably wouldn't be useful to make a bug report about that, so I will just leave that  alone for now.

Allofich:
One final thing about that, you can load the heart creature into the TES Construction Set and look at it's animations, and the explosion effects are more opaque there as well, so it may not only be lighting. Just in case you don't know, it's  a bit tricky to use. Click View->Preview Window to open it, then click "heart_akulakhan" in the Object Window's creature tab to load it in the preview window, then open up the creature properties and select the animations to watch them. Anyway, that's all from me on that subject for the moment.

MiroslavR:
You missed one _play()_ call at _unpersistAnimationState()_.

Allofich:
Thank you. Without that play(), this fix fails to work when a saved game that is in the middle of an animation is loaded. I'm adding the fix to my current PR, 1029.

PR #1028, opened 2016-08-15
Find osg_plugins

kpp:
Rewrites find_package for OpenSceneGraph plugins using LibFindMacros and osg_find_library


scrawl:
You've tested this right? I am not using static builds myself.

kpp:
Well, I set `OSG_STATIC=On` and all plugins were found. But I don't use static build either...

PR #1029, opened 2016-08-18
Improvements for playgroup and loopgroup

Allofich:
This fixes https://bugs.openmw.org/issues/3385.

Currently, in OpenMW banners never loop their animation in stormy weather. They are animated by the script OutsideBanner, which during stormy weather is constantly (every frame?) calling either loopgroup idle3 5 or playgroup idle. Everytime one of these is called, currently OpenMW sets the currently playing animation's loop count to 0, so a loop never occurs. This PR fixes this.

The following behavior is not in OpenMW master but was observed in the original engine and emulated in this PR:

-- When the same looped animation (an animation with a "Loop Start" key) is called as is already playing, all animations except for the already playing one are removed from the queue, even the newly called animation.

Ex.
`loopgroup idle3 5` (5 loops)
After playing through one loop and having 4 left
`playgroup idle` (still on loop 4, "idle" put in animation queue)
`loopgroup idle3 2` (still on loop 4. "idle" is removed from the animation queue, and no new "idle3" is added to the queue. The original loopcount of 4 continues to be used, not 2)

-- When an animation differing from the currently playing one is put in the queue, the currently playing animation's loops remain salvageable by running the animation again as long as the Loop Stop key has not been passed. Once it has been passed, trying to run the animation again will instead queue a new one.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1533,8 +1533,8 @@ void CharacterController::update(float duration)
 
     updateMagicEffects();
 
-    if(mAnimQueue.size() > 1 && (mAnimation->getLoopingEnabled(mAnimQueue.front().mGroup) == true))
-        mAnimation->setLoopingEnabled(mAnimQueue.front().mGroup, false);
+    if(!mAnimQueue.empty())
+        mAnimation->setLoopingEnabled(mAnimQueue.front().mGroup, mAnimQueue.size() <= 1);
 
     if(!cls.isActor())
     {
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1539,6 +1539,9 @@ void CharacterController::updateAnimQueue()
                              1.0f, "start", "stop", 0.0f, mAnimQueue.front().mLoopCount, loopfallback);
         }
     }
+
+    if(!mAnimQueue.empty())
+        mAnimation->setLoopingEnabled(mAnimQueue.front().mGroup, mAnimQueue.size() <= 1);
 }
 
 void CharacterController::update(float duration)
@@ -1549,9 +1552,6 @@ void CharacterController::update(float duration)
     float speed = 0.f;
 
     updateMagicEffects();
-
-    if(!mAnimQueue.empty())
-        mAnimation->setLoopingEnabled(mAnimQueue.front().mGroup, mAnimQueue.size() <= 1);
 
     if(!cls.isActor())
         updateAnimQueue();
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -2009,6 +2009,9 @@ bool CharacterController::playGroup(const std::string &groupname, int mode, int 
     }
     else
     {
+        if (!mAnimQueue.empty() && mAnimQueue.front().mGroup == groupname && isAnimPlaying(mAnimQueue.front().mGroup))
+            return true;
+
         count = std::max(count, 1);
 
         AnimationQueueEntry entry;
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1533,6 +1533,9 @@ void CharacterController::update(float duration)
 
     updateMagicEffects();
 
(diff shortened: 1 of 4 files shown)


Allofich:
I thought of a potentially better implementation, so please wait for now.

Allofich:
New version will work when saving and loading mid-animation.

Allofich:
Don't merge, there's a problem causing NPCs lips to not move. I'll look into it later.

Allofich:
Fixed the lips problem. Virtual functions weren't lined up with their parameters.

Aesylwinn:
Wow, that's the first time I've seen a goto statement used in c++ code. Looking at it, I'm pretty sure it could be removed by using a boolean. I'm not sure what scrawl thinks about that though.

Allofich:
Well, the goto statement was already there, so I think I'll leave it alone unless scrawl wants it changed.

If you don't mind, can I ask your opinion? (just ignore if you don't want to get involved)

I added the function parameter `size_t animqueuesize` to virtual function `runAnimation(),` but I don't use `size_t animqueuesize` for anything in the subclasses `NPCAnimation` and `CreatureWeaponAnimation,` which override with their own definitions for `runAnimation()` and so need the same parameter.

I read that a standard way to deal with unused parameters is to not give them a name in the definition, or to use a commented name, so I've done that, but I'm still not completely happy with it. I don't think I've seen anywhere else in the code where there is a function with an unused parameter (I may have just overlooked it.) Any ideas how I can avoid this, or do you think it is acceptable as it is?

I initially was going to try to use get() methods from within `runAnimation()` to get the character controller of the object playing the animation and read the queue that way, but I couldn't figure out how to do this, nor did I know if it would be correct to do so (do non-actors have character controllers that can be gotten this way?). Simply passing an argument seemed to be the simplest and most problem-proof way.

The purpose of the parameter is so that I can look at the animation queue's size from within `runAnimation().` This lets me delay deciding whether to set an object's animation loop count to 0 until the final frame of the animation loop is processed, which is the primary change of this PR.

Aesylwinn:
I would just leave "animqueuesize" as the parameter. Potentially, in the future someone may need that parameter, and leaving it named doesn't hurt anything. 

Unless you have access to the animation queue from the animation class, then passing it as a parameter is best.

Allofich:
OK, well I just put it back as it was. Thanks.

scrawl:
The Animation doesn't even know that a queue exists, so it doesn't make sense to pass the animation queue size to runAnimation. The animation queue is the CharacterController's business. 

What you are doing is disabling the looping feature as long as another animation is queued. Correct? I propose adding a new method Animation::setLoopingEnabled() to control this. You can then call that method from CharacterController::update.

scrawl:
Regarding the goto, feel free to refactor it.

Allofich:
New version up.

Allofich:
Changed a poorly worded comment. The travis and appveyor builds already passed without problem.

scrawl:
I made some minor changes. Do you agree with them? https://github.com/Allofich/openmw/compare/anim...scrawl:animloops

Allofich:
I'm away from my home computer and can't test or spend a long time looking at the code, but it all looks like a good consolidation and cleaning up to me.

It may not be necessary, but if you want to test, these are the things I would test when making changes.

1. Do objects with the "OutsideBanner" script animate correctly?
2. "Togglescripts," select a banner with console, give it "loopgroup idle3 2," close console and wait a split-second, select again and   give it "loopgroup idle2 1," close console, select quickly again (before loop finishes) and give "loopgroup idle3 5." Banner should continue as if you had only ever given it "loopgroup idle3 2."
3. Does test 2 work if you save and load in between console commands?
4. Does a creature also work? (I tested with "heart_akulakhan," which you can give idle through idle4 to.
5. Are there no problems with regular walking, talking, weapon swinging animations, etc.

Like I said, maybe not necessary to test all of that, it is just what I did, and what I would test if I were at my computer.

Allofich:
I was just able to test, and all the situations I described above work correctly. I'll merge it in to this PR.

Allofich:
I also added looping idle animations for the play() in unPersistAnimationState(), which MiroslavR pointed out. Without this, the previous fix for looping idles wouldn't work if an idle was loaded min-animation from a save. To test, hit "heart_akulakhan" and save while its in the middle of the shield rotating animation, then load that game and wait for it to stop spinning after a few seconds.

akortunov:
This PR breaks some other animations (in Animated Morrowind mod, for example), see [#3804](https://bugs.openmw.org/issues/3804).

PR #1032, opened 2016-08-21
Add virtual destructors

Allofich:
Addresses the following warnings from MSVC/AppVeyor.

\components\compiler\lineparser.hpp(96): warning C4265: 'Compiler::GetArgumentsFromMessageFormat': class has virtual functions, but destructor is not virtual
           instances of this class may not be destructed correctly

\components\interpreter\miscopcodes.hpp(91): warning C4265: 'Interpreter::RuntimeMessageFormatter': class has virtual functions, but destructor is not virtual
           instances of this class may not be destructed correctly

### DIFF
--- a/components/compiler/lineparser.cpp
+++ b/components/compiler/lineparser.cpp
@@ -59,6 +59,8 @@ namespace Compiler
        mExprParser (errorHandler, context, locals, literals), mAllowExpression (allowExpression)
     {}
 
+    LineParser::~LineParser() {}
+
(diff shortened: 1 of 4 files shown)


zinnschlag:
The change to the LineParser was incorrect (LineParser is derived from Parser, which has a virtual destructor).

Allofich:
Yeah, looks like both GetArgumentsFromMessageFormat and RuntimeMessageFormatter are derived from Misc::MessageFormatParser, so that would be the only destructor that was needed. Sorry, I should have been more careful.

PR #1035, opened 2016-08-25
Fix for weapon idles

Allofich:
Fix for https://bugs.openmw.org/issues/2288.

Observed original engine behavior is that for the one-handed first-person animation "Idle1h," it plays until the Stop Loop key from 2 to 5 times (loops 1 to 4 times), then plays through to the Stop key. The number of loops seems to be random within this range.

With default animations this is only relevant for Idle1h as far as I know. I did a quick search but didn't find any suitable mod animations for testing, but I manually switched the Idle1h and Idle2c (two-handed swords, axes, etc.) first-person animations keys with a hex editor, and the original engine then did the same behavior for Idle2c, so I suspect this behavior is actually used for a number of idles. For now, I believe this PR applies it for all "armed with object, spell, hand-to-hand" idles.

Changes in this PR:
1. Implemented the above behavior.
2. After implementing 1, I changed refreshIdleAnims() to update if the current idle animation stops due to finishing its loops. Otherwise the animation stops playing at the end of the loops.
3. I noticed that the point in playback of the idle animation was persisting even after moving, jumping, swimming or being hit, which caused a discontinuous look as it jumped back to the mid-animation frame after one of those actions finished. I made the idle always reset back to the first frame if it is interrupted by one of those actions. This matches original engine behavior and resolves the discontinuity  problem.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -500,7 +500,10 @@ void CharacterController::refreshIdleAnims(const WeaponInfo* weap, CharacterStat
                 idle += weap->shortgroup;
                 if(!mAnimation->hasAnimation(idle))
                     idle = "idle";
-                numLoops = 1 + Misc::Rng::rollDice(4);
+
+                // play until the Loop Stop key 2 to 5 times, then play until the Stop key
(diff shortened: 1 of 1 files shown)


Allofich:
One of the travis builds was stuck, so I just reset the builds.

PR #1037, opened 2016-08-28
Visual Studio 2015 support

ace13:
This should work for AppVeyor as well, though I expect to have to run at least one or two iterations of that before all the kinks are ironed out.

Also blocked out a few extra warnings that VS2015 spammed at me.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -679,7 +679,7 @@ if [ ! -z $CI ]; then
 fi
 
 if [ -z $VERBOSE ]; then
-	printf "- Configuring... "
+	printf -- "- Configuring... "
 else
 	echo "- cmake .. $CMAKE_OPTS"
 fi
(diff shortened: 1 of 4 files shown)


ace13:
I guess a good question before this can be seen as mergeable would be;

Do we want both 2013 and 2015 in the build matrix for AppVeyor?

scrawl:
> Do we want both 2013 and 2015 in the build matrix for AppVeyor?

Not sure. Is it common to get build issues in one version that wouldn't show up with the other version?

Well, the build matrix doesn't cost us anything, so why not.

> Also, do we want to hide 4456: declaration of 'variable' hides previous local declaration?

I think that is a useful warning. I've started fixing a few of them. Feel free to chip in.

Most of them are innocent, but [this one](https://github.com/OpenMW/openmw/blob/master/components/interpreter/defines.cpp#L176) for example revealed an actual bug.

PR #1039, opened 2016-08-29
Fix content file number issues.

Aesylwinn:
Some preliminary testing would suggest I've found the two larger issues:
- The RefNum.mContentFile field needed to be saved relative to the plugin
- The MasterData.index field wasn't being initialized

Admittedly the second fix is more of a hack by using a const_cast to modify the field. However, that is the same fix employed by the openmw engine in [esmstore.cpp](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwworld/esmstore.cpp#L49). To me, that would suggest that some of the loading code needs to be redesigned at some point.

### DIFF
--- a/apps/opencs/model/doc/savingstages.cpp
+++ b/apps/opencs/model/doc/savingstages.cpp
@@ -321,6 +321,10 @@ void CSMDoc::WriteCellCollectionStage::perform (int stage, Messages& messages)
                 {
                     CSMWorld::CellRef refRecord = ref.get();
 
+                    // Correct content file number to be relative to plugin
+                    refRecord.mRefNum.mContentFile = mDocument.getData().getPluginContentFile(
+                        refRecord.mRefNum.mContentFile);
+
                     // recalculate the ref's cell location
                     std::ostringstream stream;
                     if (!interior)
--- a/apps/opencs/model/doc/savingstages.cpp
+++ b/apps/opencs/model/doc/savingstages.cpp
@@ -321,9 +321,9 @@ void CSMDoc::WriteCellCollectionStage::perform (int stage, Messages& messages)
                 {
                     CSMWorld::CellRef refRecord = ref.get();
 
-                    // Correct content file number to be relative to plugin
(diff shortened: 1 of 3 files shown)


Aesylwinn:
I'm thinking I'll initialize the readers with the correct index instead of changing it when the cell reference is being saved. That's what I originally did, but I believe I changed it before realizing the MasterData indices weren't being initialized.

Aesylwinn:
As far as I can tell, I've solved the problem with content file numbers being invalid. Unfortunately, previously saved addons edited with opencs will still have incorrect values. I'm not sure how that should be dealt with. Perhaps a script should be made that changes all invalid content numbers to zero.

I noticed that in [refcollection.cpp](https://github.com/OpenMW/openmw/blob/master/apps/opencs/model/world/refcollection.cpp#L75), the content file number is being ignored for some reason. I suspect that will cause issues. I was wondering if someone else understood why it would be ignored.

zinnschlag:
Hm, this is a very complex topic and I would like to review it a bit more carefully; which unfortunately means not this week. I am kinda surprised that we still have problems in this area. I thought I fixed everything. Was there a specific error report that triggered your investigation?

Aesylwinn:
[Bug#3350](https://bugs.openmw.org/issues/3350) was the issue I was investigating. Basically, moving or deleting a CellRef not added by the plugin being edited creates a different CellRef. The easiest way to see this is to create an add-on that moves something in the base game (like a part of the bridge in Seyda Neen). You will see two objects when you run it in OpenMW.

I suspect part of the reason this has been overlooked so far is because OpenMW treats invalid content numbers as being added by the add-on.

Feel free to take as long as you need. I extrapolated what I think is the right pattern from using a hex editor and looking at some of OpenMW's code, so it would be best if someone who has knowledge of the load/save code examines it.

Aesylwinn:
Actually, I was investigating [bug#3427](https://bugs.openmw.org/issues/3427). I haven't tested bug#3350, so that may be caused by something else.

PR #1041, opened 2016-08-30
Don't allow casting Water Walking in deep water

Allofich:
For https://bugs.openmw.org/issues/3516. Somewhat related to https://bugs.openmw.org/issues/1138. This is a precursor to work on 1138, but does nothing about moving the player out of the water.

After a whole lot of testing I implemented the following behavior from the original engine:

- You will receive a "You cannot cast this effect right now." message if you try to cast water walking on yourself when your character is too deep in water. Generally, you seem to get this message if the z-position in water is below what would be the highest swimmable z-position if the fSwimHeightScale was 1 higher. Compare these results from the original engine, with z as reported with the "tdt" command in first-person. There were a couple oddities, possibly bugs in the original engine.

Value 0.9 (default) for fSwimHeightScale:
Dark Elf male can swim up to z of 6, can cast Water Walking down to z of -128.
High Elf male can swim up to z of 6, can cast Water Walking down to z of -141.
Wood Elf male can swim up to z of 5, can cast Water Walking down to z of -115.

Value 1.9 for fSwimHeightScale:
Dark Elf male can swim up to z of -127.
High Elf male can swim up to z of -140.
Wood Elf male can swim up to z of -114.

Value 1 for fSwimHeightScale:
Dark Elf male can swim up to z of -8, can cast Water Walking down to z of -141.
High Elf male can swim up to z of -8, can cast Water Walking down to z of -141. (no idea why)
Wood Elf male can swim up to z of -7, can cast Water Walking down to z of -126.

Value 2 for fSwimHeightScale:
Dark Elf male can swim up to z of -141, can cast Water Walking down to z of -273.
High Elf male can swim up to z of -155, can cast Water Walking down to z of -302.
Wood Elf male can swim up to z of -126, can cast Water Walking down to z of -260. (seems too low?)

Value 3 for fSwimHeightScale:
Dark Elf male can swim up to z of -274.
High Elf male can swim up to z of -301.
Wood Elf male can swim up to z of -246.

- The behavior in above, or something similar to it, also applies when you cast water walking on an NPC, in that it won't work if they are too deep in the water. You get no message, however. I decided to allow the message to show for OpenMW, since it seems like you should get the same feedback here as when you cast on yourself.

- When the message appears, you still lose the magicka, have no spell fail sound, and gain points in the magic skill.

- You can't cast water walking on water creatures (or slaughterfish, at least). 

- Even if you get the message that you cannot cast the effect right now, any other effects in the spell will cast as usual.

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -385,6 +385,7 @@ namespace MWBase
             ///Is the head of the creature underwater?
             virtual bool isSubmerged(const MWWorld::ConstPtr &object) const = 0;
             virtual bool isUnderwater(const MWWorld::CellStore* cell, const osg::Vec3f &pos) const = 0;
+            virtual bool isWaterWalkingCastableOnTarget(const MWWorld::ConstPtr &target) const = 0;
             virtual bool isOnGround(const MWWorld::Ptr &ptr) const = 0;
 
             virtual osg::Matrixf getActorHeadTransform(const MWWorld::ConstPtr& actor) const = 0;
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -393,7 +393,7 @@ namespace MWMechanics
 
                 if (!world->isWaterWalkingCastableOnTarget(target))
                 {
-                    if (castByPlayer)
+                    if (castByPlayer && caster == target)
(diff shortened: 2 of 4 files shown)


Allofich:
Changed my mind about allowing the "You cannot cast this effect right now." message with non-self-targeted spells. It's not technically accurate in that case, and could be misleading to people who don't understand the underlying mechanics. The message only shows for player self-targeted spells now, same as the original engine.

Allofich:
In the original engine you can cast Water Walking on water creatures as long as they are not fully in the water, so I implemented that.

PR #1044, opened 2016-09-02
Changes to multi-effect spells

Allofich:
For https://bugs.openmw.org/issues/3519.

This includes the following changes. All are to recreate behavior from the original engine.

1. Allow all the magic effect hit sounds of a spell to play when the spell hits a target.
2. Play all visual and sound effects when casting a spell with multiple magic effects.
3. Shoot all projectiles of spells with multiple magic effects.
4. Make the shot projectiles all have the same speed, which is the average of all the magic effects in the spell, including ones without their own projectiles (self-cast effects, etc.).
5. Use the last magic effect of a spell to determine the color of the magic effect on the hands when casting. The first magic effect, however, is used to determine what animation is played (on Self, on Touch or on Target). The hand effect is played for each magic effect, so the more magic effects, the thicker the hand effect looks.
6. Since I made all the projectiles shoot, I put all of the magic effects into the first projectile only, as otherwise the sounds of all of the magic effects would play for each projectile when they struck a target.
7. Fixed a bug with the spell explosions where a ranged effect could be exploded like a touched spell due to a missing check.

I had no problems with shooting different effects as different projectiles. They seem to all have the same collision and all disappear together when they strike something. In the original engine there is a swirling effect when there are multiple effects in a projectile spell, but I couldn't tell from testing whether that swirling effect has any effect on collision. I have not implemented this swirling effect.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -303,8 +303,8 @@ namespace MWMechanics
         if (count != 0)
             speed /= count;
 
-        std::string model;
-        std::vector<std::string> models;
+        std::string projectileID;
+        std::vector<std::string> projectileIDs;
         std::string sound;
         std::vector<std::string> sounds;
         ESM::EffectList projectileEffects;
@@ -320,10 +320,10 @@ namespace MWMechanics
             const ESM::MagicEffect *magicEffect = MWBase::Environment::get().getWorld()->getStore().get<ESM::MagicEffect>().find (
                 iter->mEffectID);
 
-            model = magicEffect->mBolt;
-            if (model.empty())
-                model = "VFX_DefaultBolt";
-            models.push_back(model);
+            projectileID = magicEffect->mBolt;
+            if (projectileID.empty())
+                projectileID = "VFX_DefaultBolt";
+            projectileIDs.push_back(projectileID);
 
             static const std::string schools[] = {
                 "alteration", "conjuration", "destruction", "illusion", "mysticism", "restoration"
@@ -335,16 +335,28 @@ namespace MWMechanics
             sounds.push_back(sound);
             projectileEffects.mList.push_back(*iter);
         }
+        
+        if (projectileEffects.mList.size() > 1) // add a VFX_Multiple projectile if there are multiple projectile effects
+        {
+            std::vector<std::string>::iterator it;
+            it = projectileIDs.begin();
+            char numstr[8];
+            sprintf(numstr, "%zd", (effects.mList.size()));
+            std::string ID = "VFX_Multiple";
+            ID = ID + numstr;
+            it = projectileIDs.insert(it, ID);
+        }
 
-            // Fall back to a "caster to target" direction if we have no other means of determining it
-            // (e.g. when cast by a non-actor)
-            if (!mTarget.isEmpty())
-                fallbackDirection =
-                   osg::Vec3f(mTarget.getRefData().getPosition().asVec3())-
-                   osg::Vec3f(mCaster.getRefData().getPosition().asVec3());
+        // Fall back to a "caster to target" direction if we have no other means of determining it
+        // (e.g. when cast by a non-actor)
+        if (!mTarget.isEmpty())
+            fallbackDirection =
+                osg::Vec3f(mTarget.getRefData().getPosition().asVec3())-
+                osg::Vec3f(mCaster.getRefData().getPosition().asVec3());
             
-            MWBase::Environment::get().getWorld()->launchMagicBolt(models, sounds, mId, speed,
-                                                       false, projectileEffects, mCaster, mSourceName, fallbackDirection);
+        if (!projectileEffects.mList.empty())
+            MWBase::Environment::get().getWorld()->launchMagicBolt(projectileIDs, sounds, mId, speed,
+                                                   false, projectileEffects, mCaster, mSourceName, fallbackDirection);
     }
 
     void CastSpell::inflict(const MWWorld::Ptr &target, const MWWorld::Ptr &caster,
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -356,7 +356,6 @@ namespace MWMechanics
 
         ESM::EffectList reflectedEffects;
         std::vector<ActiveSpells::ActiveEffect> appliedLastingEffects;
-        bool firstAppliedEffect = true;
         bool anyHarmfulEffect = false;
 
         // HACK: cache target's magic effects here, and add any applied effects to it. Use the cached effects for determining resistance.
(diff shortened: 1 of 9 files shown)


apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -305,6 +270,77 @@ namespace MWMechanics
...
+            // Only send the effects data with the first projectile, so we don't have the impact sounds
+            // playing multiple times.
+            if (isFirstProjectile)
+                MWBase::Environment::get().getWorld()->launchMagicBolt(model, sound, mId, speed,
+                                                       false, effects, mCaster, mSourceName, fallbackDirection);
+            else
+            {
+                ESM::EffectList empty;
+                MWBase::Environment::get().getWorld()->launchMagicBolt(model, sound, mId, speed,
+                                                       false, empty, mCaster, mSourceName, fallbackDirection);
```
scrawl: Launching a projectile with no effects seems like a hack. I think it'd make more sense to allow a projectile to have more than one sound / model (just like it can have multiple effects). I.e. you would supply a list of sounds & list of models to launchMagicBolt.

scrawl:
Do you know which NIF file is the swirling effect? Should be easy to implement it.

Allofich:
It might not be so easy. Looking at the files that are being accessed, it seems that it combines the magic_target_xxx.nif files and the vfx_pattern0x.nif files. Using vivec's_wrath from the default game it used vfx_pattern04.nif. Then I tried a double fireball spell I had made in the editor and it used vfx_pattern02.nif. I tried casting three or four times and it always used vfx_pattern02.nif, so the pattern that is used might be set. When I tried to replace a regular frost bolt's to use one of these pattern.nifs, the bolt was invisible in-game, so I wonder if they just act to provide places to stick the bolts on?

apps/openmw/mwmechanics/spellcasting.cpp
Allofich: I'll take a look at it later and see if I can do it that way.

Allofich:
Good idea. I'll try that but I won't get around to it until tomorrow.

Allofich:
The vfx_pattern (they are on "arrow" objects called VFX_Multiple1 through 8) that is used is based on the number of projectile effects in the spell.

From casting vivec's_wrath I get the following.
NiNode BASE VFX_Multiple4 (-9788,-71798,243)  // location of the spell projectile
	NiNode Dummy01 (0,0,0)
		NiNode COPY VFX_DestructBolt (0,0,0)
	NiNode Dummy02 (-2, 67, -3)
		NiNode COPY VFX_DestructBolt (0,0,0)
	NiNode Dummy03 (-16, -15, 20)
		NiNode COPY VFX_FrostBolt (0,0,0)
	NiNode Dummy04 (6,3,2)
		NiNode COPY VFX_DefaultBolt (0,0,0)
	NiPointLight (null) (0,0,0)

They don't seem to have any additional effects on them. A regular fireball or shock bolt will also have the NiPointLight (null) (0,0,0) show on it, but these child nodes do not have it.

The order of the projectiles attachment to the dummy nodes is the same as the order of the effects in the spell.

Allofich:
Here's an updated version, but it still needs work.

The sounds and models are now passed into one projectile as vectors of sound and model strings. The sound pointer in MagicBoltState is also now a vector of sound pointers that point to each of the sounds.

The save and load code currently just saves and loads the first of the sound and model lists. How should we handle saved data?

The invisible "VFX_Multiple" spell bolt is now added at the front of the list if there are multiple effects. The other effects are then attached with it. However, I don't know how to attach the effects to the Dummy bones of the bolt. I tried "addEffect()" in animation, but that crashes and doesn't work. Any advice? I'm not sure that I can do that part on my own.

I've also occasionally seen a slight delay when casting a spell, but I don't know if that is something particular to this PR or not, as I have also seen it with summoning spells before. Is the code in this PR being inefficient?

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -92,7 +94,18 @@ namespace MWWorld
...
 
-        mResourceSystem->getSceneManager()->getInstance(model, attachTo);
+        osg::ref_ptr<osg::Node> ptr = mResourceSystem->getSceneManager()->getInstance(model, attachTo);
+
+        if (state.mIdMagic.size() > 1)
+            for (size_t iter = 1; iter != state.mIdMagic.size(); ++iter)
+            {
+                char numstr[8];
+                sprintf(numstr, "%d", (int)iter);
+                std::string node = "Dummy0";
```
scrawl: Use Misc::StringUtils::toString(iter)

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -92,7 +94,18 @@ namespace MWWorld
...
+                node = node + numstr;
+                const ESM::Weapon* weapon = MWBase::Environment::get().getWorld()->getStore().get<ESM::Weapon>().find (state.mIdMagic.at(iter));
+                mResourceSystem->getSceneManager()->getInstance("meshes\\" + weapon->mModel, attachTo);
```
scrawl: The missing piece here is that you want to attach to the DummyX node instead of attachTo.

To find a node by name use the SceneUtil::FindByNameVisitor.

<pre>
SceneUtil::FindByNameVisitor findVisitor(node);
attachTo->accept(findVisitor);
// findVisitor.mFoundNode now holds a pointer to the node, or NULL if not found 
</pre>

scrawl:
> The save and load code currently just saves and loads the first of the sound and model lists. How should we handle saved data?

Hmm.. I guess saving the models/sounds is technically redundant because we can derive them from the effect list.

> I've also occasionally seen a slight delay when casting a spell, but I don't know if that is something particular to this PR or not, as I have also seen it with summoning spells before.

Probably because the models / sounds have to be loaded in. There is no preloading for spell effects implemented yet.

> However, I don't know how to attach the effects to the Dummy bones of the bolt so that they will spin. I tried "addEffect()" in animation, but that crashes and doesn't work.

See inline comment.

Allofich:
OK, I'll try and implement your notes.
Also, thank you for suggesting the showscenegraph command. I knew about it but hadn't actually used it in testing. It's already proving useful.

apps/openmw/mwworld/projectilemanager.cpp
Allofich: There doesn't seem to be any such method.

scrawl: Oops, you are right.

Then use the C++ way:

<pre>
std::ostringstream nodeName;
nodeName << "Dummy0" << iter;
std::string nodeNameStr = nodeName.str();
</pre>

scrawl: Or:

<pre>
std::ostringstream nodeName;
nodeName << "Dummy" << std::setw(2) << std::setfill('0') << iter;
</pre>

To handle the possibility of more than 9 magic effects.

Allofich:
Hey, it's working. :)

Only question now is what to do about save/load code. For now I will reconstruct the effects from the effect list if there is more than one projectile effect in the effect list.

Allofich:
Everything seems to be working. The model and sound data saved with magic projectiles is unused now, however.

scrawl:
> The model and sound data saved with magic projectiles is unused now, however.

That's fine.

The code for getting a projectile's sounds/models is now duplicated. Can we consolidate that into the ProjectileManager?

Allofich:
I think so. I'll have a look.

apps/openmw/mwworld/projectilemanager.hpp
```diff
@@ -49,9 +49,10 @@ namespace MWWorld
                 MWRender::RenderingManager* rendering, MWPhysics::PhysicsSystem* physics);
 
         /// If caster is an actor, the actor's facing orientation is used. Otherwise fallbackDirection is used.
-        void launchMagicBolt (const std::string& model, const std::string &sound, const std::string &spellId,
-                                     float speed, bool stack, const ESM::EffectList& effects,
-                                       const MWWorld::Ptr& caster, const std::string& sourceName, const osg::Vec3f& fallbackDirection);
+        void launchMagicBolt (const std::string &spellId, float speed, bool stack, const ESM::EffectList& effects,
+                              const MWWorld::Ptr& caster, const std::string& sourceName, const osg::Vec3f& fallbackDirection);
+
+        void ProjectileManager::getMagicBoltIDsAndSounds(std::vector<std::string>& projectileIDs, std::vector<std::string>& sounds, const ESM::EffectList& effects);
```
scrawl: Pretty sure this isn't going to build because of the extra ProjectileManager:: prefix.

Allofich:
There are a few more optimizations I'm working on.

apps/openmw/mwworld/projectilemanager.hpp
Allofich: Oops. I'll fix that.

scrawl: The function isn't being accessed from the outside, nor does it use any state of the ProjectileManager object, so we could put it in an anonymous namespace in projectilemanager.cpp .

Allofich:
Consolidated code so the effect list is only iterated through once. This means the speed of the magic projectiles is also now not taken from the save file. Also made some cleanups. The new function is now in the anonymous namespace.

This will have to be it from me, I can't spend any more time on this today. Everything seemed to be working in-game from a quick test.

Allofich:
Shoot. I just realized that the speed may not be correct if its recalculated. Give me one second.

MiroslavR:
I can no longer fire magic bolts after this PR (they seem to be hitting the player?). E.g. with the "daedric bite" spell.

MiroslavR:
After looking into it a bit further, I have found the actual issue. The projectile speed is not initialized.

Allofich:
I thought I had noticed some strangeness with the speed of projectiles. I'm adding the fix to https://github.com/OpenMW/openmw/pull/1051. Thanks.

PR #1045, opened 2016-09-03
Fix AI melee attack

Allofich:
Fixes https://bugs.openmw.org/issues/3521.

The AI currently does not use the melee attack chosen by chooseBestAttack() in aicombat.cpp most of the time, because its attack is being chosen by movement in character.cpp, and the timing of the movement from chooseBestAttack() isn't coinciding with the attack.

In the original engine the attacks of NPCs appear to be unrelated to their movement. They can do any of the attacks while standing still. This PR recreates that behavior to always allow the attack from chooseBestAttack() to play. From chooseBestAttack(), I set the attack of the NPC's charactercontroller. Another option would be to move the chooseBestAttack() method into the charactercontroller. I'm not sure which is better. I chose to leave chooseBestAttack() where it was, as the code for the attacktype selection would probably be expected to be found in aicombat.

I removed a clearing of the attack variable in the charactercontroller for this purpose, but I haven't found this to cause any problems.

I removed the movements that were associated with the attacks in chooseBestAttack(), because they are no longer serving that function and because they cause what I see as unwanted behavior, such as NPCs running into their target whenever they attempt a stab attack. Removing those movements produces behavior more similar to the original engine's melee movements.

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -687,21 +687,11 @@ std::string chooseBestAttack(const ESM::Weapon* weapon, MWMechanics::Movement &m
         //hand-to-hand deal equal damage for each type
         float roll = Misc::Rng::rollClosedProbability();
         if(roll <= 0.333f)  //side punch
-        {
-            movement.mPosition[0] = (Misc::Rng::rollClosedProbability() < 0.5f) ? 1.0f : -1.0f;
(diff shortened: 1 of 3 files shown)


Allofich:
I just realized that the hand-to-hand section of chooseBestAttack() is not doing anything, because character.cpp calls setAttackTypeRandomly() if a weapon's not equipped. I'll address that in a moment.

PR #1046, opened 2016-09-03
Editor fixes

Aesylwinn:
The first commit makes the editor more usable by adding the terrain mask to what is interactable. This means it is possible to drop objects on terrain. Additionally, clicking on the terrain will not select an unseen object beneath the terrain (and it stops tool tips from doing the same).

The second commit fixes [bug#3520](https://bugs.openmw.org/issues/3520). Turns out that the command line interpreter doesn't like unquoted paths with spaces.


Aesylwinn:
Just to note, I did test the second fix on a Windows machine.

PR #1048, opened 2016-09-05
Pathfinding unification v2

mrcheko:
This is re-request of #993 with fixes for issues mentioned by @scrawl:
- fixes rotation to target if target's and actor's positions exactly coincide (this was causing zero div);
- fix AiWander stuck after enter/leave building
- fix AiTravel not rebuilding path after passing through cells border. This also got fixed for other packages

### DIFF
--- a/apps/openmw/mwmechanics/aiactivate.cpp
+++ b/apps/openmw/mwmechanics/aiactivate.cpp
@@ -23,32 +23,28 @@ namespace MWMechanics
         return new AiActivate(*this);
     }
 
-    bool AiActivate::execute (const MWWorld::Ptr& actor, CharacterController& characterController, AiState& state, float duration)
+    bool AiActivate::execute(const MWWorld::Ptr& actor, CharacterController& characterController, AiState& state, float duration)
     {
-        ESM::Position pos = actor.getRefData().getPosition(); //position of the actor
         const MWWorld::Ptr target = MWBase::Environment::get().getWorld()->searchPtr(mObjectId, false); //The target to follow
 
         actor.getClass().getCreatureStats(actor).setDrawState(DrawState_Nothing);
 
-        if(target == MWWorld::Ptr() ||
-            !target.getRefData().getCount() || !target.getRefData().isEnabled()  // Really we should be checking whether the target is currently registered
+        if (target == MWWorld::Ptr() ||
+            !target.getRefData().getCount() || !target.getRefData().isEnabled()  // Really we should check whether the target is currently registered
                                                                                 // with the MechanicsManager
-                )
-            return true;   //Target doesn't exist
+            )
+        return true;   //Target doesn't exist
 
-        //Set the target desition from the actor
+        //Set the target destination for the actor
         ESM::Pathgrid::Point dest = target.getRefData().getPosition().pos;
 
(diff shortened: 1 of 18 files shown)


Allofich:
I was just testing the issue that existed before with bringing the follower Itermerel into Addamasartus and having slow downs. I didn't have that, but I did encounter something I've never seen in OpenMW before, where he got stuck running in circles. Load the attached save, and he will probably be running in circles, whereas with a build created from just before this merge he will successfully get out of it.

Interestingly, with this PR's code, if you then reload the save after loading it once, he should get out of the circle, which maybe is a clue for how to fix the problem. Any idea what's going on?
[Test.zip](https://github.com/OpenMW/openmw/files/456674/Test.zip)

mrcheko:
I just tested this in 'pathfinding' branch, my ideas are following: Itermerel has high running speed - that's why he can't turn to the pathpoint in time to reach it. The moment when he manages to get out of circling is the moment after he was hit with a throwing star - because while playing the hit animation he does not move hence has time to successfully turn to the pathpoint.
Though I'm curious if this was fixed before this PR. Will test that in a bit.

Also just in case you didn't know, there is a whole topic dedicated for 'running in circles' problem:
http://forum.openmw.org/viewtopic.php?f=6&t=2916&p=32904&hilit=pathfind#p32904

Allofich:
I did read that topic once a fairly long time ago, before getting involved with the project.

> The moment when he manages to get out of circling is the moment after he was hit with a throwing star

I did just see this happen, yes, but he also will reliably get out of the circling from the second time I load the game onward during a single session of OpenMW, which is interesting to me.

Allofich:
He got stuck again, and is circling a pathgrid point, which sounds like that forum topic. I'll try the code from before the PR.

mrcheko:
The reason why this works in pre-PR is that LOS-ray for shortcut possibility check was casted between actors' half-heights there. And now the ray is casted between positions (actor's position is located just under feet). In case of your saved game: in pre-PR Itermerel instantly shortcuts directly to the enemy and in post-PR the LOS-check fails. So the actual issue is presented in both pre-PR and post-PR, it's just can be revealed more easily in latter. I'll take an effort to fix that finally then.

Allofich:
There's what seems to be a fairly rare issue of NPCs becoming unable to attack due to failing to find their animation which I have tracked down to this PR, as it doesn't occur with the code from directly before it.

Try the attached save. I have only seen this happen with the ordinators in Ald Daedroth, who fight with bound weapons, so that may be related.
[fail_to_find_animation.zip](https://github.com/OpenMW/openmw/files/472727/fail_to_find_animation.zip)

scrawl:
The animation error looks to be a conflict with PR #1045. The mAttackType in the character controller is empty. Either the AI is not setting an attack type at all, or it's setting an empty one.

Allofich:
This PR seems to be the cause of https://bugs.openmw.org/issues/3646. I don't have the problem with https://github.com/OpenMW/openmw/commit/b8fa555260b4c501b9e475ef265d0b9b53f344e7, which is the commit right before it, but I do with https://github.com/OpenMW/openmw/commit/05e4542d332df09c47816e74ab70cc6d6c4b46ab, the merge commit.

MiroslavR:
Also, I could be very wrong about this, but I have my suspicions this PR is also causing https://bugs.openmw.org/issues/3615.

PR #1049, opened 2016-09-07
Create area effect visual for non-area spells

Allofich:
In the original engine, projectiles that do not affect a range will still create a silent scale 1 "Area Effect" visual. This PR implements this for OpenMW.

In the original engine it seems like if the projectile has multiple effects, only the first effect will show if the projectile hits the terrain or water, but all of them will show if it hits a static or an activator. I haven't copied this because it seems like it's probably just an unintentional limitation of the original engine, so all the effects will show even if the projectile hits the terrain or water.


Allofich:
Please don't merge yet. The effect erroneously appears when hitting an actor.

PR #1051, opened 2016-09-08
Use spell effect particle textures

Allofich:
For https://bugs.openmw.org/issues/3534.

This implements using the particle texture of some spell effects to override the texture.

I tested (I think) all effects in the original engine, and based on that, the following ESM::Statics and ESM::Weapons display differently depending on the particle texture of their magic effect.

VFX_DestructBolt
VFX_DefaultBolt

VFX_DefaultHit
VFX_MysticismHit
VFX_SoulTrapHit

VFX_DefaultArea

VFX_DefaultCast
VFX_ShieldCast

VFX_Hands

Those are implemented in this PR, so this PR works for vanilla game data. However, what seems to really decide whether the texture is used is whether the .NIF file has, directly under the top node, a NITexturingProperty node with a NiSourceTexture node under it defining a texture. I determined this from various tests and looking at the files with NifSkope. It also looks like the original engine only overrides one of these textures, I think maybe the latter one. For example, for e\SoulTrapHit.NIF and e\magic_cast_S.NIF, which both have two NiTexturingProperty nodes with textures under the top node, only part of the effect is overridden by the particle texture.

w\magic_target_myst.NIF has two NiTexturingProperty nodes directly under the top node, but does not seem to show a difference from using the particle texture. Maybe the texture that is being overridden is unused or is hard to notice.

I'm not sure how to determine from the .NIF file whether or which nodes to override with the particle texture, so I'm leaving that for future work or, if scrawl or someone can offer some guidance, maybe it could be put in this PR. The NiTexturingProperty node is unnamed according to NifSkope, and I wasn't able to locate it and override it using SceneUtil::FindByNameVisitor. Also, if we want to override only the latter texture (assuming that is what the original is doing), I'm not sure if FindByNameVisitor is usable for that.

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -6,6 +6,7 @@
 #include <osg/Array>
 #include <osg/LOD>
 #include <osg/TexGen>
+#include <osg/ValueObject>
 
 // resource
 #include <components/misc/stringops.hpp>
@@ -368,13 +369,32 @@ namespace NifOsg
             return created;
         }
 
-        void applyNodeProperties(const Nif::Node *nifNode, osg::Node *applyTo, SceneUtil::CompositeStateSetUpdater* composite, Resource::ImageManager* imageManager, std::vector<int>& boundTextures, int animflags)
+        void applyNodeProperties(const Nif::Node *nifNode, osg::Node *applyTo, SceneUtil::CompositeStateSetUpdater* composite, Resource::ImageManager* imageManager, std::vector<int>& boundTextures, int animflags, bool isRootNode)
         {
             const Nif::PropertyList& props = nifNode->props;
-            for (size_t i = 0; i <props.length();++i)
+            bool foundFirstRootTexturingProperty = false;
+            for (size_t i = 0; i <props.length(); ++i)
             {
                 if (!props[i].empty())
+                {
+                    // store the recIndex of the NiTexturingProperty. Used for spells when overriding textures.
+                    // Get the lowest numbered one for the root node. This is what is overridden when a spell
+                    // effect "particle texture" is used. For non-root nodes we keep setting until we have the highest
+                    // numbered one, which is the one that displays in the game and can be overridden if it matches the
+                    // lowest one on the root.
+                    if (!foundFirstRootTexturingProperty && isRootNode && props[i].getPtr()->recType == Nif::RC_NiTexturingProperty)
+                    {
+                        int index = props[i].getPtr()->recIndex;
+                        applyTo->setUserValue("overrideIndex", index);  
+                        foundFirstRootTexturingProperty = true;
+                    }
+                    else if (props[i].getPtr()->recType == Nif::RC_NiTexturingProperty)
+                    {
+                        int index = props[i].getPtr()->recIndex;
+                        applyTo->setUserValue("NiTexturingPropertyIndex", index);                
+                    }
                     handleProperty(props[i].getPtr(), applyTo, composite, imageManager, boundTextures, animflags);
+                }              
             }
         }
 
@@ -631,7 +651,7 @@ namespace NifOsg
 
             osg::ref_ptr<SceneUtil::CompositeStateSetUpdater> composite = new SceneUtil::CompositeStateSetUpdater;
 
-            applyNodeProperties(nifNode, node, composite, imageManager, boundTextures, animflags);
+            applyNodeProperties(nifNode, node, composite, imageManager, boundTextures, animflags, node == rootNode);
 
             if (nifNode->recType == Nif::RC_NiTriShape && !skipMeshes)
             {
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -369,7 +369,7 @@ namespace NifOsg
             return created;
         }
 
-        void applyNodeProperties(const Nif::Node *nifNode, osg::Node *applyTo, SceneUtil::CompositeStateSetUpdater* composite, Resource::ImageManager* imageManager, std::vector<int>& boundTextures, int animflags, bool isRootNode)
+        void applyNodeProperties(const Nif::Node *nifNode, osg::Node *applyTo, SceneUtil::CompositeStateSetUpdater* composite, Resource::ImageManager* imageManager, std::vector<int>& boundTextures, int animflags)
         {
             const Nif::PropertyList& props = nifNode->props;
             bool foundFirstRootTexturingProperty = false;
@@ -382,7 +382,7 @@ namespace NifOsg
                     // effect "particle texture" is used. For non-root nodes we keep setting until we have the highest
                     // numbered one, which is the one that displays in the game and can be overridden if it matches the
                     // lowest one on the root.
-                    if (!foundFirstRootTexturingProperty && isRootNode && props[i].getPtr()->recType == Nif::RC_NiTexturingProperty)
+                    if (!foundFirstRootTexturingProperty && nifNode->parent == NULL && props[i].getPtr()->recType == Nif::RC_NiTexturingProperty)
                     {
                         int index = props[i].getPtr()->recIndex;
                         applyTo->setUserValue("overrideIndex", index);  
@@ -651,7 +651,7 @@ namespace NifOsg
 
             osg::ref_ptr<SceneUtil::CompositeStateSetUpdater> composite = new SceneUtil::CompositeStateSetUpdater;
 
-            applyNodeProperties(nifNode, node, composite, imageManager, boundTextures, animflags, node == rootNode);
+            applyNodeProperties(nifNode, node, composite, imageManager, boundTextures, animflags);
 
             if (nifNode->recType == Nif::RC_NiTriShape && !skipMeshes)
             {
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -277,11 +277,13 @@ namespace NifOsg
     public:
         /// @param filename used for warning messages.
         LoaderImpl(const std::string& filename)
-            : mFilename(filename)
+            : mFilename(filename), mFirstRootTextureIndex(-1), mFoundFirstRootTexturingProperty(false)
         {
 
         }
         std::string mFilename;
+        size_t mFirstRootTextureIndex;
+        bool mFoundFirstRootTexturingProperty;
 
         static void loadKf(Nif::NIFFilePtr nif, KeyframeHolder& target)
         {
@@ -372,26 +374,21 @@ namespace NifOsg
         void applyNodeProperties(const Nif::Node *nifNode, osg::Node *applyTo, SceneUtil::CompositeStateSetUpdater* composite, Resource::ImageManager* imageManager, std::vector<int>& boundTextures, int animflags)
         {
             const Nif::PropertyList& props = nifNode->props;
-            bool foundFirstRootTexturingProperty = false;
             for (size_t i = 0; i <props.length(); ++i)
             {
                 if (!props[i].empty())
                 {
-                    // store the recIndex of the NiTexturingProperty. Used for spells when overriding textures.
-                    // Get the lowest numbered one for the root node. This is what is overridden when a spell
-                    // effect "particle texture" is used. For non-root nodes we keep setting until we have the highest
-                    // numbered one, which is the one that displays in the game and can be overridden if it matches the
-                    // lowest one on the root.
-                    if (!foundFirstRootTexturingProperty && nifNode->parent == NULL && props[i].getPtr()->recType == Nif::RC_NiTexturingProperty)
+                    // Get the lowest numbered recIndex of the NiTexturingProperty root node.
+                    // This is what is overridden when a spell effect "particle texture" is used.
+                    if (nifNode->parent == NULL && !mFoundFirstRootTexturingProperty && props[i].getPtr()->recType == Nif::RC_NiTexturingProperty)
                     {
-                        int index = props[i].getPtr()->recIndex;
-                        applyTo->setUserValue("overrideIndex", index);  
-                        foundFirstRootTexturingProperty = true;
+                        mFirstRootTextureIndex = props[i].getPtr()->recIndex;
+                        mFoundFirstRootTexturingProperty = true;
                     }
                     else if (props[i].getPtr()->recType == Nif::RC_NiTexturingProperty)
                     {
-                        int index = props[i].getPtr()->recIndex;
-                        applyTo->setUserValue("NiTexturingPropertyIndex", index);                
+                        if (props[i].getPtr()->recIndex == mFirstRootTextureIndex)
+                            applyTo->setUserValue("overrideFx", 1);                
                     }
                     handleProperty(props[i].getPtr(), applyTo, composite, imageManager, boundTextures, animflags);
                 }              
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1235,10 +1235,8 @@ bool CharacterController::updateWeaponState()
                     MWMechanics::CastSpell cast(mPtr, NULL);
                     cast.playSpellCastingEffects(spellid);
 
-                    const ESM::Spell *spell = store.get<ESM::Spell>().find(spellid);
-                    
-                    const ESM::ENAMstruct &lastEffect = spell->mEffects.mList.at(spell->mEffects.mList.size() - 1);
-
+                    const ESM::Spell *spell = store.get<ESM::Spell>().find(spellid);                   
+                    const ESM::ENAMstruct &lastEffect = spell->mEffects.mList.back();
                     const ESM::MagicEffect *effect;
 
                     effect = store.get<ESM::MagicEffect>().find(lastEffect.mEffectID); // use last effect of list for color of VFX_Hands
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -541,12 +541,17 @@ namespace MWMechanics
                     else
                         castStatic = MWBase::Environment::get().getWorld()->getStore().get<ESM::Static>().find ("VFX_DefaultHit");
 
+                    std::string texture = "";
+
+                    if (!magicEffect->mParticle.empty() && !(magicEffect->mData.mFlags & ESM::MagicEffect::Harmful))
+                        texture = magicEffect->mParticle;
+
                     // TODO: VFX are no longer active after saving/reloading the game
                     bool loop = (magicEffect->mData.mFlags & ESM::MagicEffect::ContinuousVfx) != 0;
                     // Note: in case of non actor, a free effect should be fine as well
                     MWRender::Animation* anim = MWBase::Environment::get().getWorld()->getAnimation(target);
                     if (anim)
-                        anim->addEffect("meshes\\" + castStatic->mModel, magicEffect->mIndex, loop, "");
+                        anim->addEffect("meshes\\" + castStatic->mModel, magicEffect->mIndex, loop, "", texture);
                 }
             }
         }
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -30,7 +30,6 @@
 
 namespace MWMechanics
 {
-
     ESM::Skill::SkillEnum spellSchoolToSkill(int school)
     {
         std::map<int, ESM::Skill::SkillEnum> schoolSkillMap; // maps spell school to skill id
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -543,6 +543,7 @@ namespace MWMechanics
 
                     std::string texture = "";
 
+                    // Use particle textures for non-harmful effects
                     if (!magicEffect->mParticle.empty() && !(magicEffect->mData.mFlags & ESM::MagicEffect::Harmful))
                         texture = magicEffect->mParticle;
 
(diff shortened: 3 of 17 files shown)


scrawl:
The nif loader will convert a NiTexturingProperty to one or more osg::Texture2D's and add them to the parent node's StateSet.

So your condition would be something like: if the model's root node has a StateSet with a texture attribute:  override that texture attribute

Not sure if the same condition applies to all other uses of overrideTexture, if not, the new condition will have to be made optional via a function parameter. I think the blood effect also uses overrideTexture.

> w\magic_target_myst.NIF has two NiTexturingProperty nodes directly under the top node, but does not seem to show a difference from using the particle texture. Maybe the texture that is being overridden is unused or is hard to notice.

The NiTexturingProperty's on magic_target_myst.NIF's root node have no effect because the actual particle systems define their own NiTexturingProperty. Children can override their parent's properties, it is that simple.

Having two NiTexturingProperties on the same node actually makes no sense, because the second one overrides the first one.

scrawl:
I never noticed wrong effect colors - does this PR have any impact with the default game files? Do I need to edit MagicEffect colors to see a difference?

scrawl:
> The NiTexturingProperty's on magic_target_myst.NIF's root node have no effect because the actual particle systems define their own NiTexturingProperty. 

Hmm, but this is also happening in files that (according to your list) do get magic effect textures. Looks like we need more research.

nifskope1

Allofich:
There are a number of spells with wrong colors in the default game. Here are some examples:
"daedric bite", "weakness to frost", "drain acrobatics" (all use the destruction bolt)
"shockball" (uses the default bolt. It has a purple tinge in the original engine)
"reflect," "spell absorption," "telekinesis," or the hit effect for "absorb health" (these use the mysticism hit)

Allofich:
What's a good thing to look for for more research? My conclusion was just made because it seemed that all of these effects in which the particle texture was used had a texture defined at the top (the root?), whereas other ones I checked that did not have the effect had textures further down in the children nodes. I thought that maybe the texture overriding with the particle texture only looks so far as the top level of the node hierarchy.

scrawl:
Have you tried swapping out models (i.e. so that a fireball uses magic_target_myst.NIF for its bolt)? Then we should know if the coloring comes from the model or from some other (hardcoded?) place.

Allofich:
Yes, I tried swapping out the ice bolt, which doesn't use the particle texture, with the fire bolt. But whichever effect used the fire bolt NIF used the particle texture.

It also appears to be unrelated to magic effect flags, as effects with the exact same flags differed on whether they used the texture or not. Spell school was also unrelated.

scrawl:
Well, I can't see too many differences w.r.t. NiTexturingProperty usage between magic_target_myst.nif and e.g. soultraphit.nif.

You could try taking two nif files that are not too dissimilar, but behave differently w.r.t. particle textures, then edit one file step by step so that it (in the end) exactly matches the second file - you would see at which step along the way the particle behaviour changed.

Allofich:
I'm going to try changing around the mysticism bolt NIF's textures and see if I can learn anything.

Allofich:
Looks like the mysticism bolt is affected, it's just very subtle. There is a slight tinting at the very front of the bolt of the color of the particle texture I gave it. It's for the first texture listed in NifSkope, the "vfx_shield_glow.tga" file. If you change this to vfx_grnflare.tga, the difference is obvious. So it seems to depend on the existing texture, not completely replace it, I guess.

Allofich:
Never mind, the "strength" of the replacement seems to not depend on the texture it is replacing after all. I think it just looked like that because I had changed the other texture in the mysticism bolt as well, and I suppose it didn't cover up the first texture so much, so it looked like the first texture was showing more strongly.

Also, I confirmed that with the soul trap hit as well, it is only the first of the textures that is overridden.

Allofich:
> The NiTexturingProperty's on magic_target_myst.NIF's root node have no effect because the actual particle systems define their own NiTexturingProperty. Children can override their parent's properties, it is that simple.

When I modify a texture with NifSkope, whether it is a texture higher up on the node hierarchy or in a child, it automatically sets the other textureproperty to match, so I can't check whether or not there is overriding going on.

> Having two NiTexturingProperties on the same node actually makes no sense, because the second one overrides the first one.

In the case of magic_target_myst.nif, each of those NiTexturingProperties is "linked" with a different "superspray" particle emitter. If you change one it changes one of the particle emitters's textures, and if you change the other it changes the other particle emitter. I've confirmed that both textures are being used in-game, one for the head of the bolt and one for the tail.

I tried adding a NiTexturingProperty with a NiSourceTexture to the root of the frost bolt model, which doesn't show the particle texture, but the particle texture still didn't show. So, I thought maybe that an "NiBSParticleNode" has to exist, so I tried adding a texture to the root of the conjuration bolt model, which has a NiBSParticleNode, but that didn't work either.

I notice, though, that changing the texture in a child node or changing the texture on the parent node I added does not cause the other to change, so maybe that is the problem: They have to be linked somehow. Unfortunately I can't find anything in NifSkope that I've set differently from the way that the textures are that come with the game. Well, I'll take another look again later.

scrawl:
What you call "linking" is simply the two nodes pointing to the same instance of NiTexturingProperty. The NIF format allows this, although our nifosg loader won't recognize that the NiTexturingProperty are the same, it'll simply create new StateSet/TextureAttributes for each NiTexturingProperty "link" that is encountered. So you'll have to get creative in your implementation I'm afraid.

> Unfortunately I can't find anything in NifSkope that I've set differently from the way that the textures are that come with the game, so I'm not sure how to do the linking. 

It is simple, you need to enter the record index of the NiTexturingProperty to link.

nifskope_link
.

Allofich:
Yeah, I realized that later. I think I've mostly figured out how the textures are chosen for child nodes, and therefore whether a "Particle Texture" from a magic effect will show or not on a particular node of a .nif file. I want to double check one more time, but I'm a little burned out from testing right now.

I know of at least one case that seems like it will be problematic if we want to copy the original engine. That is magic_target.nif (the shock bolt). It uses the same texture on all its nodes, but some of the child nodes have their own NiTexturingProperty, rather than pointing to the root node one. Apparently because of this they do not inherit the texture from the root node but instead "override" it, even though it's the same texture. Because of this the Particle Texture, which replaces the first NiTexturingProperty in the root node, does not override the textures for those child nodes. You can see the result: the purplish shock bolt has white particles trailing behind it that aren't made purple like the rest of the bolt.

If we lose the NiTexturingProperty information in the conversion, I'm not sure whether it's possible to implement this kind of case.

scrawl:
One more problem, a root node having *two* NiTexturingProperties will also lose information in the conversion (the second NiTexturingProperty overwrites the first one, so only the second one survives).

If, in the future, we implement a model optimizer that removes redundant state, both NiTexturingProperty's on the root node may be removed even.

A solution I could think of, is to build the information you need into the scene graph via UserData (objects derived from osg::Object have a UserDataContainer you can store whatever you want in). E.g. you write a NiTexturingProperty's record index to the user data container of the Texture state attribute. . Then you add a userData to the root node that specifies which NiTexturingProperty index was the first NiTexturingProperty of that root node, and should be overridden (e.g. ("overrideStateIndex", X)). That should be all the information you need. It's a dirty workaround, but I can't think of a better way to implement this super weird feature.

We're already employing a similar workaround with UserData for keyframe controllers, where we store the original translation/rotation/scale as UserData on the node, because it's not possible to get the individual transformations back from the node's transformation matrix. These kinds of workarounds will be necessary where the NIF scenegraph model doesn't work exactly like the OSG scenegraph does. That's the kind of trade-offs we have to make when going with a premade framework. (To be fair, Ogre3D was *way* more limiting than OSG).

Allofich:
> E.g. you write a NiTexturingProperty's record index to the user data container of the Texture state attribute. Then you add a userData to the root node that specifies which NiTexturingProperty index was the first NiTexturingProperty of that root node, and should be overridden

Is the second part necessary? If you have the record indices, and which NiTexturingPropertys were on which nodes, it seems like that's enough.

I just tried changing the order of the two NiTexturingPropertys in the mysticism bolt so that index 3 comes after 4 in NifSkope (I also turned off the "auto sanitize before save" feature that would correct this), but the index 3 NiTexturingProperty was still the one that was overridden in-game, so it seems like the index is used for determining order.

The changes to the nifloader sound like a fairly hefty job, maybe something that won't get done soon. What should I do about this PR? Are you interested in it as a temporary solution? I get the impression you don't like having special cases in the code.

If you do want to merge the PR, I'll make a small change to the TODO comments to be more accurate.

Allofich:
Well I might as well make the change to the TODO comments now I guess.

Allofich:
Also updated the comment about statics. I've done some testing with that and statics do seem to always create a casting vfx, it's just under the terrain sometimes and not visible.

Allofich:
Sorry, I guess maybe the whole reason you explained about the nifosg loader idea, was you were suggesting I do it for this PR, right? I'm not very confident, as I haven't touched that part of the code at all, but I'll take a look later and see if I can make any sense of it.

Allofich:
> Is the second part necessary? If you have the record indices, and which NiTexturingPropertys were on which nodes, it seems like that's enough.

Never mind about this question. I said it without thinking enough. We need the first index because, as you said, if there are more than one NiTexturingPropertys in the root, the first one gets overwritten, so we wouldn't have access to it anymore. The same will happen at every level of the node hierarchy, so only the NiTexturingProperty for the texture that is actually being used on that node will be saved. Then looking for any instances of the root's first NiTexturingProperty index throughout the node tree and overriding each of those textures, is what we do, if my understanding is correct.

I started attempting an implementation but it is slow going.

Allofich:
Implementation using NiTexturingProperty is up. This works with all the spell effects I have tried, as well as blood effects.

Allofich:
I noticed that permanent effect enchanted items were only playing the visual and audio effects of the first effect, unlike the original engine, so I added a fix for that in.

components/nifosg/nifloader.cpp
```diff
@@ -368,13 +369,32 @@ namespace NifOsg
             return created;
         }
 
-        void applyNodeProperties(const Nif::Node *nifNode, osg::Node *applyTo, SceneUtil::CompositeStateSetUpdater* composite, Resource::ImageManager* imageManager, std::vector<int>& boundTextures, int animflags)
+        void applyNodeProperties(const Nif::Node *nifNode, osg::Node *applyTo, SceneUtil::CompositeStateSetUpdater* composite, Resource::ImageManager* imageManager, std::vector<int>& boundTextures, int animflags, bool isRootNode)
```
scrawl: ```isRootNode``` is redundant - you could use ```nifNode->parent == NULL```

Allofich: Fixed.

scrawl:
That is pretty much what I had in mind. I would have named the user data in a non-NIF specific way (i.e. NiTexturingPropertyIndex -> textureIndex).

I just thought of a possibly better implementation (sorry), that would allow handling everything within the nifloader, with no need to parse user data on the outside. The osg StateSet system allows you to set StateAttributes with the StateAttribute::PROTECTED flag, which means that the StateAttribute can't be overriden by a StateAttribute in a parent StateSet that have the StateAttribute::OVERRIDE flag. The nif loader would set the ::PROTECTED flag on the texture attributes that don't match the index of the root node's first NiTexturingProperty. Then the texture override function just has to set the texture on the root node with OVERRIDE flag, no need to parse the graph or check user data.

Allofich:
OK, I'll take a look at doing it that way. I agree that sounds like a simpler, and therefore better implementation.

Allofich:
If you have a moment, scrawl, I could use some help.
I've tried setting the protected flag in nifloader.cpp like so
`if (texprop->recIndex != mFirstRootTexturingProperty)`
`stateset->setTextureAttributeAndModes(texUnit, texture2d, osg::StateAttribute::ON | osg::StateAttribute::PROTECTED);`

`else
stateset->setTextureAttributeAndModes(texUnit, texture2d, osg::StateAttribute::ON);`

but this is not enough. It looks like many of the spell effects have both the original and the overwriting texture. I looked at http://forum.openscenegraph.org/viewtopic.php?t=4673&view=next and I think maybe something similar is happening here, where `overrideTexture()` in util.cpp is adding a new texture attribute, and the PROTECTED flag doesn't stop the node from inheriting that new one in addition to the existing one. If the PROTECTED flag is not set, only the overriding texture will be present. 

In `overrideTexture()`, instead of creating a new osg::Texture2D as is being done currently, I tried to get a pointer to the existing texture attribute and modify it (this was just kind of experimental, so I might be doing something stupid), like 

`osg::ref_ptr<osg::Texture2D> tex = dynamic_cast<osg::Texture2D*>(stateset->getTextureAttribute(0, osg::StateAttribute::TEXTURE));`

`osg::Image* image = resourceSystem->getImageManager()->getImage(correctedTexture);`

`tex->setImage(image);`

but `tex->setImage(image)` crashes. Any ideas?

Allofich:
It's a little hard to read the code in that post, so here's a link to the WIP code if you care to look.
https://github.com/Allofich/openmw/commit/58143d4c9ab13049a08b2bc34f3a752b3cbfd22b?diff=split

scrawl:
> I've tried setting the protected flag in nifloader.cpp like so but this is not enough. 

Apparently there is a bug in OSG's state sharing mechanism. If you comment out line 395 in scenemanager.cpp,

<pre>
            //osgDB::Registry::instance()->getOrCreateSharedStateManager()->share(loaded.get());
</pre>
the PROTECTED flag works as expected. Looks like the state sharing code is not honoring the PROTECTED/OVERRIDE/INHERIT flags. I'll try to get a fix for this submitted. We can copy the osgDB::SharedStateManager into our own code with the fix applied so that we don't have to depend on a new OSG version with the fix.

> I tried to get a pointer to the existing texture attribute and modify it (this was just kind of experimental, so I might be doing something stupid), like

That is wrong. The existing texture attribute is a shared resource you shouldn't be modfiying (i.e. changing it might be inadvertently changing other meshes that use the same texture). The existing (shared) texture might be submitted for rendering already, in which case a crash may happen... or a background loading thread might be reading that texture's properties, in which case you get a crash on modifying also. But your crash is likely not even one of these (rare) race conditions, probably just a stateset with tex == NULL which your code does not seem to check against.

Allofich:
> probably just a stateset with tex == NULL which your code does not seem to check against.

Yeah, I noticed that and tried checking for NULL, but it still crashed, so it must have been something else. Anyway if that method is wrong I won't try it anymore.

I'll try the fix you suggested.

Allofich:
Seems to be working fine after commenting out that line. Good thing you figured that out. I don't even know what most of that line means.

I'll do some cleanup and then put the new version on the PR here.

Allofich:
We also need to handle blood effects somehow, since they don't follow the same rule for texture overriding. Maybe we need to parse through the nodes and take off the protected flags?

scrawl:
FYI: the SharedStateManager optimizes scene graphs by globally sharing state. For example, if 2 NIF's used the same texture file, with the same wrap/filter settings, the SharedStateManager will share them (i.e. replace the second NIF's state attribute with a pointer to the first NIF's state attribute). Additionally, if two StateSets end up having the same contents, the second StateSet is replaced by a pointer to the first StateSet. This is done as a post-load step. The sharing improves rendering performance & memory efficiency.

> We also need to handle blood effects somehow, since they don't follow the same rule for texture overriding. Maybe we need to parse through the nodes and take off the protected flags?

Hm, that's annoying... yes, removing the protected flags should work, but to be absolutely sure not to break shared stuff, you'll need to make a copy of the StateSet again. 

There is a similar problem with the cloud meshes, where OVERRIDE is used to set the cloud texture according to the current weather, as well as the night skydome mesh.

scrawl:
Hm, so if we need these workarounds in other parts of the code just to make the OVERRIDE function as it was previously intended, then perhaps the last solution with the user data was better after all.

Sorry for that going back & forth, sometimes I need to see different solutions implemented to be able to decide on one.

It's still nice that we experimented with the OVERRIDE/PROTECTED stuff, as we found & fixed that bug in OSG :)

Could you make a small change to the original solution, instead of setting overrideIndex & NiTexturingPropertyIndex, just set "overrideFx" to "0" on the TextureAttribute that shall not be overridden (i.e. recIndex != mFirstRootTextureIndex). That simplifies the work of the TextureOverrideVisitor a bit.

Allofich:
OK, well give me a little time to sort through this all, and I'll have it up before too long, hopefully.

Allofich:
I was actually setting the NiTexturingPropertyIndex (I'll rename that, btw) on nodes, rather than on the TextureAttributes, as it was more convenient, but do you want it on the TextureAttributes?

scrawl:
Actually, setting user data on the TextureAttribute might interfere with the sharing of that TextureAttribute. Nodes are never shared, so don't have that problem. Node is better.

Allofich:
Ok, it's up now, putting the data on nodes rather than texture attributes.

scrawl:
<pre>/home/scrawl/Dev/openmw/components/nifosg/nifloader.cpp:285:42: warning: non-static data member initializers only available with -std=c++11 or -std=gnu++11 [enabled by default]
         size_t mFirstRootTextureIndex = -1;
                                          ^
/home/scrawl/Dev/openmw/components/nifosg/nifloader.cpp:286:49: warning: non-static data member initializers only available with -std=c++11 or -std=gnu++11 [enabled by default]
         bool mFoundFirstRootTexturingProperty = false;
</pre>

Allofich:
Sorry, carelessness from being tired.

Allofich:
Left in an iostream now. Just a second.

Allofich:
Deleted some trailing whitespace I accidentally added.

scrawl:
Casting Veloth's Benison crashes the game for me.

Btw, if you are looking for related things to implement (in new PRs), I noticed that:
- Divine Intervention plays a different Hit effect in vanilla MW
- The Light explosion effect is much larger than in vanilla MW (e.g. 'light' spell). There is something about the light effect or its NIF file (magic_area_ill.nif) that causes the area scaling to be ignored.

Allofich:
No crash for me.

Allofich:
I don't see anything special about that spell. It's just a single on-self restore health effect.

scrawl:
Oops, the crash was my fault. It was the fix to the SharedStateManager I had applied to OSG that contained a crash bug.

Allofich:
Oh, OK. That's a relief, lol.

Allofich:
Yeah, I'll probably take a look at those spells, btw. It doesn't seem like they would be that hard to fix, but I don't want to promise anything.

Allofich:
> The Light explosion effect is much larger than in vanilla MW (e.g. 'light' spell). There is something about the light effect or its NIF file (magic_area_ill.nif) that causes the area scaling to be ignored.

It's the NIF file, not the effect. There will be a huge area vfx if you give the Light effect a NIF that can scale.

I think I have figured out what's happening, but I don't know how to fix it in OpenMW. I think what's happening is that vanilla MW's effect scaling does not change particle size. Instead the the "size" value from the NiParticleSystemController is always used. OpenMW also uses this size value, but applies effect scaling to it. "magic_summon.nif" is another effect that won't show scaling in vanilla MW if you use it as an area vfx for a spell, and it also uses particles.

Another example is "magic_cast_dst.nif", the destruction cast model. Try selecting the silt strider in Seyda Neen and making it cast a destruction spell through the console. The effect is scaled to be huge, but the particles themselves are not larger than usual.

As for what _is_ scaling in the casting effect, I think it's the NiTriShapes. "magic_area_ill.nif" has no NiTriShape. "magic_summon.nif" does have an NiTriShape, but it's just a single box, and I think that maybe the box changing size wouldn't affect the positioning of the particle effect.

Allofich:
Maybe we need to again set a userdata to mark NiBSParticleNodes as not to be scaled, and apply the scaling in addEffect() with a visitor that checks for this value?

scrawl:
To keep the particles un-scaled, I believe you can do

<pre>
            partsys->setParticleScaleReferenceFrame(osgParticle::ParticleSystem::WORLD_COORDINATES);
</pre>
instead of
<pre>
            partsys->setParticleScaleReferenceFrame(osgParticle::ParticleSystem::LOCAL_COORDINATES);
</pre>

But, I assume that I used the LOCAL_COORDINATES for a reason, so we may need more research on when particles should or should not be scaled. One possibility is that particleFlags contains a flag for this we haven't discovered. How does the magic_area_ill.nif look when you place it as a normal mesh in the game and scale it? (if nothing at all shows up, you might need to set the AutoPlay flag on the particle node to get the effect to play by default, or change the controller function).

In the original game, when I scale a candle in the Balmora mages guild (using the setscale command), the particles definitely scale with it. In fact, the particles seem to scale even more than the object itself, which I suspect is the result of an inaccurate computation in the billboarding code (i.e. leaving out the sqrt() to improve performance, at the cost of some accuracy).

scrawl:
With the osgParticle::ParticleSystem::WORLD_COORDINATES, magic_area_ill.nif still doesn't look correct. Even though the individual particles are small, the simulation itself is still scaled, so the particles are further apart and move faster than in the original game.

Allofich:
> How does the magic_area_ill.nif look when you place it as a normal mesh in the game and scale it?

It can be placed through the construction set, but it is stuck at the initial animation frame. You can't select it by clicking on it, but you can scale it by referring to it by name through the console, like "VFX_IllusionArea->setscale #," and it does indeed change size. I'm not sure how you set an "autoplay" flag. I couldn't see a way to add such a flag in nifSkope, nor any useful information about it online.

I can confirm one thing, if you look at the scene graph when the effect is playing when the light spell hits, it says the various particle nodes are being scaled by 0.01666, in other words 1/60, the opposite of the 60 scale from the spell effect. So it seems like something is correcting it back to original size. Unfortunately I haven't been able to find out anything else useful from testing right now.

Allofich:
As far as I can see, the difference between spell effects that are reported scaling their particle nodes back to 1 by the showscenegraph command and those that aren't, is whether they have an NiAutoNormalParticles node or not. However, even in effects that aren't reported like this and lack that node, (levitate cast effect, restoration cast effect), the particles stay at scale 1. I tested many spell effects, but I can't find any in which particles actually end up with a changed size after scaling, I've only been able to scale the candle like you said, and the freely placed vfx static.

PR #1052, opened 2016-09-09
Marker Size From User Prefs

ZenTux:
Referencing Bug: 3526 
https://bugs.openmw.org/issues/3526

Allows marker size to be set from a new section: "Marker Size" in user prefs, which is preferable to the currently hard-coded size values. This can be an Interim solution until full autoscaling can be developed. 

Tested by loading a cell, selecting small objects and positing with the newly sized markers.


scrawl:
To scale an object to the screen use the osg::AutoTransform

<pre>
diff --git a/apps/opencs/view/render/object.cpp b/apps/opencs/view/render/object.cpp
index eb90e9d..2834289 100644
--- a/apps/opencs/view/render/object.cpp
+++ b/apps/opencs/view/render/object.cpp
@@ -4,6 +4,7 @@
 #include <iostream>
 
 #include <osg/Group>
+#include <osg/AutoTransform>
 #include <osg/PositionAttitudeTransform>
 
 #include <osg/ShapeDrawable>
@@ -31,11 +32,10 @@
 
 
 const float CSVRender::Object::MarkerShaftWidth = 30;
-const float CSVRender::Object::MarkerShaftBaseLength = 70;
+const float CSVRender::Object::MarkerShaftBaseLength = 150;
 const float CSVRender::Object::MarkerHeadWidth = 50;
 const float CSVRender::Object::MarkerHeadLength = 50;
 
-
 namespace
 {
 
@@ -213,7 +213,7 @@ osg::ref_ptr<osg::Node> CSVRender::Object::makeMoveOrScaleMarker (int axis)
 {
     osg::ref_ptr<osg::Geometry> geometry (new osg::Geometry);
 
-    float shaftLength = MarkerShaftBaseLength + mBaseNode->getBound().radius();
+    float shaftLength = MarkerShaftBaseLength;
 
     // shaft
     osg::Vec3Array *vertices = new osg::Vec3Array;
@@ -295,17 +295,20 @@ osg::ref_ptr<osg::Node> CSVRender::Object::makeMoveOrScaleMarker (int axis)
 
     geometry->getOrCreateStateSet()->setMode (GL_LIGHTING, osg::StateAttribute::OFF);
 
-    osg::ref_ptr<osg::Geode> geode (new osg::Geode);
-    geode->addDrawable (geometry);
+    osg::ref_ptr<osg::AutoTransform> autoTransform (new osg::AutoTransform);
+    autoTransform->setAutoRotateMode(osg::AutoTransform::NO_ROTATION);
+    autoTransform->setAutoScaleToScreen(true);
+
+    autoTransform->addChild (geometry);
 
-    return geode;
+    return autoTransform;
 }
 
 osg::ref_ptr<osg::Node> CSVRender::Object::makeRotateMarker (int axis)
 {
     const float Pi = 3.14159265f;
 
-    const float InnerRadius = mBaseNode->getBound().radius();
+    const float InnerRadius = 100;
     const float OuterRadius = InnerRadius + MarkerShaftWidth;
 
     const float SegmentDistance = 100.f;
@@ -376,10 +379,13 @@ osg::ref_ptr<osg::Node> CSVRender::Object::makeRotateMarker (int axis)
 
     geometry->getOrCreateStateSet()->setMode (GL_LIGHTING, osg::StateAttribute::OFF);
 
-    osg::ref_ptr<osg::Geode> geode = new osg::Geode();
-    geode->addDrawable (geometry);
+    osg::ref_ptr<osg::AutoTransform> autoTransform (new osg::AutoTransform);
+    autoTransform->setAutoRotateMode(osg::AutoTransform::NO_ROTATION);
+    autoTransform->setAutoScaleToScreen(true);
+
+    autoTransform->addChild (geometry);
 
-    return geode;
+    return autoTransform;
 }
 
 osg::Vec3f CSVRender::Object::getMarkerPosition (float x, float y, float z, int axis)
</pre>

Though, the auto-scaling doesn't seem that usable to me. The controls get huge when you zoom out and too tiny (inside the object) when you zoom in. I think I prefer the bounding-box based version.

apps/opencs/model/prefs/state.cpp
```diff
@@ -166,6 +166,16 @@ void CSMPrefs::State::declare()
         setTooltip ("When using next/previous functions at the last/first item of a "
         "list go to the first/last item");
 
+    declareCategory ("Marker Size");
```
Aesylwinn: I think it would be better to call this category "3D Scene Marker". We have several small categories that deal with random scene options. Perhaps these should all be consolidated?

apps/opencs/model/prefs/state.cpp
```diff
@@ -166,6 +166,16 @@ void CSMPrefs::State::declare()
...
+    declareDouble ("MarkerShaftWidth", "Marker Shaft Width", 30.0).
+        setRange (0.001, 100.0);
+    declareDouble ("MarkerShaftBaseLength", "Marker Shaft Base Length", 70.0).
+        setRange (0.001, 100.0);
+    declareDouble ("MarkerHeadWidth", "Marker Header Width", 50.0).
+        setRange (0.001, 100.0);
+    declareDouble ("MarkerHeadLength", "Marker Header Length", 50.0).
```
Aesylwinn: Also, the defaults can probably be set smaller. The original values before I changed them were: 10, 50, 30, 30 in that order. It seems to be a good middle ground, so I would go with those.

Aesylwinn:
I tried out your patch, scrawl, and I agree that there are some issues with auto-scaling. I'm not sure if much could be done to fix that.

ZenTux:
From the responses, I'm a bit unclear on how you want me to proceed with this PR, specifically. Can you guys clarify?

Aesylwinn:
Axis selection will continue to be an issue at larger distances. However, if shortcuts for moving along an axis are added, then it becomes less of an issue.

Aesylwinn:
We could also keep a user setting that acts as a multiplier. Thoughts anyone?

Aesylwinn:
@zinnschlag If you have the time, could you give your input?

Aesylwinn:
Additionally, the option to hide the marker could be added.

zinnschlag:
Difficult. I tried various variations when I first implemented the markers and was not able to come up with something better than the current implementation (including some experimentation with AutoTransform). I am not quite sure what we should do about it, but I can provide some more input:

The idea of scaling markers to the bounding box sounds interesting at first, but that would make it very bothersome to move small objects over large distances, because you would need to be zoomed out for that which would make the markers very small.

Here is a crazy idea: Let's do both (scaling to BB and auto-scaling). Have a toggleable setting (keyboard shortcut and button menu item) that can switch between the two.

Another idea would be bounded size: Do either auto-scaling our bb-scaling, but do not allow the markers to shrink or grow beyond a specific (configurable) size.

I see no point in an option to hide the markers completely. However a bindable key that hides the markers while pressed down sounds kinda useful. That could also work as a toggle, in which case we should supplement the shortcut with a toggleable button menu item (either mode oder sub-mode button, not sure yet).

Aesylwinn:
> Here is a crazy idea: Let's do both (scaling to BB and auto-scaling). Have a toggleable setting (keyboard shortcut and button menu item) that can switch between the two.

Crazy, yes. However, I still like it. Instead of having it be a toggleable setting though, I would have it act similar to a modifier. While you press the key, it changes to auto-scaled mode. When that key is released, it draws the bounding box again.

> Another idea would be bounded size: Do either auto-scaling our bb-scaling, but do not allow the markers to shrink or grow beyond a specific (configurable) size.

If you set a limit, you're still going to eventually run into the same problem. It's just at a larger distance. If we want to change the growth, I would rather change the growth function than to set an upper limit. For example, making it scale logarithmically instead of linearly (though you would still need a lower limit). However, it's still more of a quick band-aid than a long-term solution.

> I see no point in an option to hide the markers completely.

I think it would be useful for detail work, when you don't need the axes and only want to make small adjustments.

> However a bindable key that hides the markers while pressed down sounds kinda useful. That could also work as a toggle, in which case we should supplement the shortcut with a toggleable button menu item.

That's interesting. I support this option if you still don't think a button to hide the markers completely is necessary.

Aesylwinn:
I may have misunderstood your last point.

zinnschlag:
Regarding the hiding: I thought we were speaking about I user settings option. That is what I am opposed to. But a button isn't really an option either, because we are running out of UI space.

Regarding the limits: You are not going to zoom in arbitrarily close to anything. For any given project there is a limit how small meshes get. A reasonable default (with the option to adjust it) should do the job.
Still if the crazy idea works we may have no need for this.

Aesylwinn:
The button for the marker could be added to the existing UI that controls scene visibility.

Anyway, it seems we've come up with a plan. @ZenTux Are you interested in implementing any of this? Since you started this, you get first dibs. :)

ZenTux:
It would take me too long to come up to speed and be able to effectively contribute to this specific enhancement. I'll continue to lurk and contribute to this project as my time and schedule allows.

jeffreyhaines:
@ZenTux if you are no longer planning to finish this issue, could you close your pull request please. Thanks. Hope to see you continued support and work on other issues.

jeffreyhaines:
But the action of free the bug for someone else to work on is why we'd close it. You can link the old PR in the bug report and mention the failure to complete. It makes the chore of ensuring everyone has a bug report and things are getting done hard if we leave stall bug reports in the code.

@zinnschlag what do you think?

scrawl:
If a bug was assigned to someone or they opened a PR, but haven't done anything in X months then its fair to say the task is up for grabs for whoever wants it.

jeffreyhaines:
In that process though we should close the old PR and attach a link to the ticket. Otherwise we'll have lots of stale PRs.. Leaving stale PRs makes it hard for anyone who's trying to apply conformance to our bug reporting guidelines and testing procedures.

scrawl:
We are getting off topic. I will move this to the forum, feel free to continue there.

jeffreyhaines:
Great. Thanks

jeffreyhaines:
@ZenTux are you still planning to work on this issue? If so let us know or close the PR. Thanks

scrawl:
Please dont ask anyone to close PRs until we have finished the forum discussion.

In this case OP already made it clear that he isnt going to do any more work. If we wanted the PR closed we would do so ourselves.

In the future before you post boilerplate comments on X different PR's / bug tickets please start a forum discussion first; my concern is these type of comments spam everyone's feed while not having much value.

jeffreyhaines:
@scrawl understood, my apologies.

scrawl:
Can anyone check how Blender does it?

akortunov:
FYI: the [bug #3526](https://bugs.openmw.org/issues/3526) was rejected after markers were made transparent.

PR #1053, opened 2016-09-10
NPC drowning

gugus:
fixes https://bugs.openmw.org/issues/1374

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -148,6 +148,10 @@ namespace MWMechanics
         if (actor.getClass().getCreatureStats(actor).isDead())
             return true;
 
+		//if the actor has to go up for air, then no other actions is taken
+		if (preventDrowning(actor))
+			return false;
+
         MWWorld::Ptr target = MWBase::Environment::get().getWorld()->searchPtrViaActorId(mTargetActorId);
         if (target.isEmpty())
             return false;
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -148,9 +148,9 @@ namespace MWMechanics
         if (actor.getClass().getCreatureStats(actor).isDead())
             return true;
 
-		//if the actor has to go up for air, then no other actions is taken
-		if (preventDrowning(actor))
-			return false;
+        //if the actor has to go up for air, then no other actions is taken
+        if (preventDrowning(actor))
+            return false;
 
         MWWorld::Ptr target = MWBase::Environment::get().getWorld()->searchPtrViaActorId(mTargetActorId);
         if (target.isEmpty())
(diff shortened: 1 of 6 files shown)


Svetomech:
Great way to name a commit.

gugus:
yeah sorry I got confused by the github gui interface (why am I even using it? ^^)

Capostrophic:
For future: you shouldn't use tabs. Read [code formatting conventions.](https://wiki.openmw.org/index.php?title=Code_Formatting_Conventions)

gugus:
Argh I forgot to reset my editor with tabs, thanks, I'll correct that

Allofich:
I just tested this code out, and I think maybe you aren't making the NPC go up until too late. It's only safe to wait that long if you are near the surface of the water. I tested in the original game, and NPCs seem to go up for breath at about the time that the breath meter is halfway down. This makes sense, too, since it would usually keep an NPC from diving down further than they can safely go.

You can rename that earlier commit by doing the following from a git bash prompt
1. type `git rebase -i ~2`
2. next to the earlier commit change "pick" to "r" or "reword" (in my case, I have to press insert to start editing text, then esc to exit edit mode, then type :wq to save changes)
3. it should start rebasing and automatically bring up the commit message for you to change. After you do that you are finished.

jeffreyhaines:
@gugus just wondering how're you doing on this issue and if you intend to continue working on it. Appreciate your response.

akortunov:
> I tested in the original game, and NPCs seem to go up for breath at about the time that the breath meter is halfway down

We can achive this by modifying preventDrowning function:

    bool MWMechanics::AiPackage::preventDrowning(const MWWorld::Ptr& actor)
    {
        static const float fHoldBreathTime = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fHoldBreathTime")->getFloat();
    
        const MWWorld::Class& actorClass = actor.getClass();
        if (actorClass.isNpc())
        {
            if (actorClass.getNpcStats(actor).getTimeToStartDrowning() < fHoldBreathTime / 2)
            {
                //go up
                //another option: create a new AiTravel package with coordinates at water level. 
                //or maybe cast a spell if available
                actor.getClass().getMovementSettings(actor).mPosition[1] = 1;
                smoothTurn(actor, -180, 0);
                return true;
            }
        }
        return false;
    } 

Also I found a bug. NPC will drown if a destination point is blocked (e.g. by another NPC) instead of reaching surface in another point.

During the test three guards from Seyda Neen used the same destination point to reach the surface - above PC. One of guards drowned, because destination point was blocked by two other guards (who tried to swim down, but their paths were blocked by the first guard).

Allofich:
> Also I found a bug. NPC will drown if a destination point is blocked (e.g. by another NPC) instead of reaching surface in another point.

A bug in this PR or in the original engine?

I think in the original engine NPCs just move upward with no other consideration when they need air. When I was testing the behavior I put an NPC below a boat and it just floated up against the boat bottom when it needed air, got stuck there and drowned.

akortunov:
> A bug in this PR or in the original engine?

In this PR.

akortunov:
Are there any problems with this PR except of rebasing and preventDrowning() go up time? Maybe we can create another PR since author is inactive over half of year?

PR #1054, opened 2016-09-10
AI dodging

Allofich:
For https://bugs.openmw.org/issues/3539 and https://bugs.openmw.org/issues/3540.

Some notes:

1. In the original engine, NPCs/bipedal creatures using a ranged weapon will back up if they are in their opponent's weapon range (unless the opponent has a ranged weapon, in which case they don't back up.) I added a note about this to the source, but I didn't implement this for now. Maybe we should have an option at some point to select between original and enhanced AI combat behavior to handle cases like this?

2. Although I have enabled dodging for bipedal creatures as in the original game, they hardly move for some reason. I haven't made any changes that should cause this, so the problem must lie elsewhere.

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -330,7 +330,7 @@ namespace MWMechanics
             mCombatMove = true;
         }
         // dodge movements (for NPCs and bipedal creatures)
-        else if (isNpc || ((actor.get<ESM::Creature>()->mBase->mFlags & ESM::Creature::Bipedal) != 0))
+        else if (actor.getClass().isBipedal(actor))
         {
             // get the range of the target's weapon
             float rangeAttackOfTarget = 0.f;
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -69,7 +69,7 @@ namespace MWMechanics
         mMovement()
         {}
 
-        void startCombatMove(bool isNpc, bool isDistantCombat, float distToTarget, float rangeAttack);
+        void startCombatMove(bool isNpc, bool isDistantCombat, float distToTarget, float rangeAttack, const MWWorld::Ptr& target);
         void updateCombatMove(float duration);
         void stopCombatMove();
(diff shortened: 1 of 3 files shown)


Allofich:
Also changed creature weapon range to use fHandtoHandReach, just like NPCs or the player using hand-to-hand, because I observed that in the original engine increasing that value increased the range of a mudcrab's attacks.

Allofich:
By the way I'm confident that in the original game the biped flag is what determines the dodge behavior for creatures, as giving it to a scamp caused it to start doing the behavior. It also causes animation errors. Give it a try if you want to see a scamp with its head spinning around.

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -323,25 +322,49 @@ namespace MWMechanics
...
     {
         if (mMovement.mPosition[0] || mMovement.mPosition[1])
         {
             mTimerCombatMove = 0.1f + 0.1f * Misc::Rng::rollClosedProbability();
             mCombatMove = true;
         }
-        // dodge movements (for NPCs only)
-        else if (isNpc && (!isDistantCombat || (distToTarget < rangeAttack / 2)))
+        // dodge movements (for NPCs and bipedal creatures)
+        else if (isNpc || ((actor.get<ESM::Creature>()->mBase->mFlags & ESM::Creature::Bipedal) != 0))
```
scrawl: Use ```else if (actor.getClass().isBipedal(actor))```

apps/openmw/mwmechanics/aicombataction.cpp
```diff
@@ -491,17 +491,9 @@ namespace MWMechanics
...
-            else
-            {
-                static float fHandToHandReach =
-                    MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fHandToHandReach")->getFloat();
-
-                return fHandToHandReach * fCombatDistance;
-            }
+            static float fHandToHandReach =
+                MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fHandToHandReach")->getFloat();
+            return fHandToHandReach * fCombatDistance;
```
scrawl: mIsNpc can be removed

PR #1055, opened 2016-09-10
prevent running in circles around path points

mrcheko:
Basically circling around a pathpoint happens rarely. But recently the problem was observed again in the quest of escorting Itermerel due to his high running speed. Easiest way to watch this is by downloading the save
https://github.com/OpenMW/openmw/files/456674/Test.zip (provided by Allofich [here](https://github.com/OpenMW/openmw/pull/1048#issuecomment-244936686)). Note: may require to load more than once to observe.

So I wrote a function which checks beforehand if circling will occur (the more speed actor has the more probability that it will happen). This function also minimizes any unnecessary possible loops around pathpoints when reaching them.
Explanation how the `isReachableRotatingOnTheRun` works:
can_rotate_on_the_run 
_On the left: running in circles will happen, on the right: no_
P - actor's position
dir - actor's orientation/looking direction
D - destination pathpoint (has a smaller circle around, radius = 32 (`PathTolerance`))
C - center of actor's smallest moving circle (depends on his speed; bigger circle, radius = 53 for Itermerel)
So if function returns false - then actor is not moving while continues to rotate to the pathpoint. When orientation is sufficient for actor to be able to rotate in time then he starts to move. If actor's speed is not as high as e.g. Itermerel's, then the old behaviour is preserved - actor rotates on the run.

@scrawl 
I have some doubts though: why do we use negative axes to determine actor's orientation (https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwworld/scene.cpp#L46)? Is that because of our angles vs OSG ones clockwise / counter-clockwise mismatch?
I managed to get correct results only with negative axes.

### DIFF
--- a/apps/openmw/mwmechanics/aipackage.cpp
+++ b/apps/openmw/mwmechanics/aipackage.cpp
@@ -19,12 +19,15 @@
 #include "actorutil.hpp"
 #include "coordinateconverter.hpp"
 
+#include <osg/Quat>
+
 MWMechanics::AiPackage::~AiPackage() {}
 
 MWMechanics::AiPackage::AiPackage() : 
     mTimer(AI_REACTION_TIME + 1.0f), // to force initial pathbuild
     mIsShortcutting(false),
-    mShortcutProhibited(false), mShortcutFailPos()
+    mShortcutProhibited(false), mShortcutFailPos(),
+    mRotateOnTheRunChecks(0)
 {
 }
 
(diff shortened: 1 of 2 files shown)


Allofich:
Very nice providing a diagram like that with your explanation.

scrawl:
> I have some doubts though: why do we use negative axes to determine actor's orientation (https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwworld/scene.cpp#L46)?

Because that is how the rotation data in the ESM files has to be interpreted.

Aesylwinn:
I believe it has to do with a change in handedness for the coordinate systems.

PR #1058, opened 2016-09-12
Don't play blood effects for completely resisted hits

Allofich:
For https://bugs.openmw.org/issues/3549.

The check for playing the blood effect was being done before the weapon resist effect was applied, so the blood effect was showing when it shouldn't. Test with any creature or npc by giving them the "Immune to Normal Weapons" spell and hitting them with a weapon that doesn't get through this effect.

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -323,7 +323,7 @@ namespace MWClass
         victim.getClass().onHit(victim, damage, healthdmg, weapon, ptr, hitPosition, true);
     }
 
-    void Creature::onHit(const MWWorld::Ptr &ptr, float damage, bool ishealth, const MWWorld::Ptr &object, const MWWorld::Ptr &attacker, osg::Vec3f hitPosition, bool successful) const
+    void Creature::onHit(const MWWorld::Ptr &ptr, float damage, bool ishealth, const MWWorld::Ptr &object, const MWWorld::Ptr &attacker, const osg::Vec3f &hitPosition, bool successful) const
     {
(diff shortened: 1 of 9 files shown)


Allofich:
Actually, I think it might be better to move the playing of the blood effect into onHit. I'm going to rewrite this.

Allofich:
New version up.

Old version is at https://github.com/Allofich/openmw/tree/onhit2. It made onHit() return damage (after resistance was applied to it) as a float, which was then used to determine whether to play the blood effect, but I think the new version is better.

apps/openmw/mwclass/creature.cpp
```diff
@@ -318,15 +318,12 @@ namespace MWClass
...
-            MWBase::Environment::get().getWorld()->spawnBloodEffect(victim, hitPosition);
-
         MWMechanics::diseaseContact(victim, ptr);
 
-        victim.getClass().onHit(victim, damage, healthdmg, weapon, ptr, true);
+        victim.getClass().onHit(victim, damage, healthdmg, weapon, ptr, hitPosition, true);
     }
 
-    void Creature::onHit(const MWWorld::Ptr &ptr, float damage, bool ishealth, const MWWorld::Ptr &object, const MWWorld::Ptr &attacker, bool successful) const
+    void Creature::onHit(const MWWorld::Ptr &ptr, float damage, bool ishealth, const MWWorld::Ptr &object, const MWWorld::Ptr &attacker, osg::Vec3f hitPosition, bool successful) const
```
scrawl: hitPosition could be passed by const reference: ```, const osg::Vec3f& hitPosition, ```

Allofich:
Let me also make them pass as reference.

PR #1062, opened 2016-09-15
apps/wizard: (trivial) fix misplaced arguments in installFile for installFiles

stil-t:
Arguments `keepSource` and `single` were misplaced in `Wizard::UnshieldWorker::installFile`.

### DIFF
--- a/apps/wizard/unshield/unshieldworker.cpp
+++ b/apps/wizard/unshield/unshieldworker.cpp
@@ -302,7 +302,7 @@ bool Wizard::UnshieldWorker::copyDirectory(const QString &source, const QString 
 
 bool Wizard::UnshieldWorker::installFile(const QString &fileName, const QString &path, Qt::MatchFlags flags, bool keepSource)
 {
-    return installFiles(fileName, path, flags, true, keepSource);
+    return installFiles(fileName, path, flags, keepSource, true);
 }
 
 bool Wizard::UnshieldWorker::installFiles(const QString &fileName, const QString &path, Qt::MatchFlags flags, bool keepSource, bool single)


scrawl:
The reversed parameters are definitely a bug, but I do not agree with the "trivial" label: the solution still needs testing. After all, this bug could be masking another bug, or the reversed parameters might be reversed again in another place.

I did a quick test run of the install wizard with this PR, didn't notice any problems, so will merge.

stil-t:
>The reversed parameters are definitely a bug, but I do not agree with the "trivial" label: the solution still needs testing. After all, this bug could be masking another bug, or the reversed parameters might be reversed again in another place.

About application flow impact: `Wizard::UnshieldWorker::installFile` called only by `Wizard::UnshieldWorker::installComponent` ([[1]](https://github.com/OpenMW/openmw/blob/master/apps/wizard/unshield/unshieldworker.cpp#L616), [[2]](https://github.com/OpenMW/openmw/blob/master/apps/wizard/unshield/unshieldworker.cpp#L624), [[3]](https://github.com/OpenMW/openmw/blob/master/apps/wizard/unshield/unshieldworker.cpp#L657), [[4]](https://github.com/OpenMW/openmw/blob/master/apps/wizard/unshield/unshieldworker.cpp#L670), [[5]](https://github.com/OpenMW/openmw/blob/master/apps/wizard/unshield/unshieldworker.cpp#L682)), and all of this calls use only first two args leaving rest by default (`keepSource = false`). On the other side `installFile` calls `Wizard::UnshieldWorker::installFiles` with

 * before fix: `single = false`, `keepSource = true` and
 * after fix: `single = true`, `keepSource = false`

There are two moments: `single` and `keepSource`.

Fixing `single` didn't change anything, before fix `installFiles` treated `fileName` as template (something like `.esm`) but `installComponent` always supplied full file names, so it doesn't matter.

Fixing `keepSource` affected `Wizard::UnshieldWorker::copyFile`, where it's passed by `installFiles`. Behavior changed, but again it doesn't matter, because at parent level (`Wizard::UnshieldWorker::installComponent`) OpenMW creates own temp folder for unpacking, so keeping or deleting files from temp folder don't affect user.

In summary it doesn't matter, but still it was a bug.

--------
I got your point, next time I'll be more verbose 🙂

PR #1063, opened 2016-09-16
Drain and Fortify fixes

Allofich:
Fixes https://bugs.openmw.org/issues/3049, https://bugs.openmw.org/issues/2796 and https://bugs.openmw.org/issues/2680.

Adds a new variable mDrainAndFortifyModifier to the health, magicka and fatigue stats, which is used to track the sum of drain and fortify modifiers, which are applied to the current values, rather than the modified values as they are now.

This new variable could be saved in saved games. Instead, right now when a game is loaded I set it to a value that would never be used to indicate that a saved game was just loaded and the variable should be updated to the correct value once without applying the modifiers.

I also took out some unnecessary processing on dead actors. Otherwise the drained/fortified modifiers would be reversed on character death (since the effects were done), and show a positive health if the player opened the inventory during the death animation after dying from a drain effect.

Note that if the code from this PR is used on an existing saved game, any drain and fortify effects should be removed first or else the stats won't be correct.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -546,7 +546,7 @@ namespace MWMechanics
         float diff = (static_cast<int>(magickaFactor*intelligence)) - magicka.getBase();
         float currentToBaseRatio = (magicka.getCurrent() / magicka.getBase());
         magicka.setModified(magicka.getModified() + diff, 0);
-        magicka.setCurrent(magicka.getBase() * currentToBaseRatio);
+        magicka.setCurrent(magicka.getBase() * currentToBaseRatio, false, true);
         creatureStats.setMagicka(magicka);
     }
 
@@ -577,8 +577,14 @@ namespace MWMechanics
         float normalizedEncumbrance = ptr.getClass().getNormalizedEncumbrance(ptr);
         if (normalizedEncumbrance > 1)
             normalizedEncumbrance = 1;
+ 
+        // Current fatigue can be above base value due to a fortify effect.
+        // In that case stop here and don't try to restore.
+        DynamicStat<float> fatigue = stats.getFatigue();
+        if (fatigue.getCurrent() >= fatigue.getBase())
+            return;
 
-        // restore fatigue
+        // Restore fatigue
         float fFatigueReturnBase = settings.find("fFatigueReturnBase")->getFloat ();
         float fFatigueReturnMult = settings.find("fFatigueReturnMult")->getFloat ();
         float fEndFatigueMult = settings.find("fEndFatigueMult")->getFloat ();
@@ -586,7 +592,6 @@ namespace MWMechanics
         float x = fFatigueReturnBase + fFatigueReturnMult * (1 - normalizedEncumbrance);
         x *= fEndFatigueMult * endurance;
 
-        DynamicStat<float> fatigue = stats.getFatigue();
         fatigue.setCurrent (fatigue.getCurrent() + 3600 * x);
         stats.setFatigue (fatigue);
     }
@@ -598,16 +603,20 @@ namespace MWMechanics
 
         MWMechanics::CreatureStats& stats = ptr.getClass().getCreatureStats (ptr);
 
-        int endurance = stats.getAttribute (ESM::Attribute::Endurance).getModified ();
+        // Current fatigue can be above base value due to a fortify effect.
+        // In that case stop here and don't try to restore.
+        DynamicStat<float> fatigue = stats.getFatigue();
+        if (fatigue.getCurrent() >= fatigue.getBase())
+            return;
 
-        // restore fatigue
+        // Restore fatigue
+        int endurance = stats.getAttribute(ESM::Attribute::Endurance).getModified();
         const MWWorld::Store<ESM::GameSetting>& settings = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>();
         static const float fFatigueReturnBase = settings.find("fFatigueReturnBase")->getFloat ();
         static const float fFatigueReturnMult = settings.find("fFatigueReturnMult")->getFloat ();
 
         float x = fFatigueReturnBase + fFatigueReturnMult * endurance;
 
-        DynamicStat<float> fatigue = stats.getFatigue();
         fatigue.setCurrent (fatigue.getCurrent() + duration * x);
         stats.setFatigue (fatigue);
     }
@@ -723,7 +732,7 @@ namespace MWMechanics
         for(int i = 0;i < 3;++i)
         {
             DynamicStat<float> stat = creatureStats.getDynamic(i);
-            stat.setModifier(effects.get(ESM::MagicEffect::FortifyHealth+i).getMagnitude() -
+            stat.setCurrentModifier(effects.get(ESM::MagicEffect::FortifyHealth+i).getMagnitude() -
                              effects.get(ESM::MagicEffect::DrainHealth+i).getMagnitude(),
                              // Magicka can be decreased below zero due to a fortify effect wearing off
                              // Fatigue can be decreased below zero meaning the actor will be knocked out
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -697,6 +697,19 @@ namespace MWMechanics
             }
         }
 
+        // dynamic stats
+        for (int i = 0; i < 3; ++i)
+        {
+            DynamicStat<float> stat = creatureStats.getDynamic(i);
+            stat.setCurrentModifier(effects.get(ESM::MagicEffect::FortifyHealth + i).getMagnitude() -
+                effects.get(ESM::MagicEffect::DrainHealth + i).getMagnitude(),
+                // Magicka can be decreased below zero due to a fortify effect wearing off
+                // Fatigue can be decreased below zero meaning the actor will be knocked out
+                i == 1 || i == 2);
+
+            creatureStats.setDynamic(i, stat);
+        }
+
         // attributes
         for(int i = 0;i < ESM::Attribute::Length;++i)
         {
@@ -726,19 +739,6 @@ namespace MWMechanics
                     }
                 }
             }
-        }
-
-        // dynamic stats
-        for(int i = 0;i < 3;++i)
-        {
-            DynamicStat<float> stat = creatureStats.getDynamic(i);
-            stat.setCurrentModifier(effects.get(ESM::MagicEffect::FortifyHealth+i).getMagnitude() -
-                             effects.get(ESM::MagicEffect::DrainHealth+i).getMagnitude(),
-                             // Magicka can be decreased below zero due to a fortify effect wearing off
-                             // Fatigue can be decreased below zero meaning the actor will be knocked out
-                             i == 1 || i == 2);
-
-            creatureStats.setDynamic(i, stat);
         }
 
         // AI setting modifiers
--- a/apps/openmw/mwmechanics/stat.cpp
+++ b/apps/openmw/mwmechanics/stat.cpp
@@ -5,11 +5,11 @@
 namespace MWMechanics
 {
     template<typename T>
-    Stat<T>::Stat() : mBase (0), mModified (0) {}
+    Stat<T>::Stat() : mBase (0), mModified (0), mCurrentModified (0) {}
     template<typename T>
-    Stat<T>::Stat(T base) : mBase (base), mModified (base) {}
+    Stat<T>::Stat(T base) : mBase (base), mModified (base), mCurrentModified (base) {}
     template<typename T>
-    Stat<T>::Stat(T base, T modified) : mBase (base), mModified (modified) {}
+    Stat<T>::Stat(T base, T modified) : mBase (base), mModified (modified), mCurrentModified (modified) {}
 
     template<typename T>
     const T& Stat<T>::getBase() const
@@ -22,23 +22,42 @@ namespace MWMechanics
     {
         return std::max(static_cast<T>(0), mModified);
     }
+
+    template<typename T>
+    T Stat<T>::getCurrentModified() const
+    {
+        return mCurrentModified;
+    }
+
     template<typename T>
     T Stat<T>::getModifier() const
     {
         return mModified-mBase;
     }
+
+    template<typename T>
+    T Stat<T>::getCurrentModifier() const
+    {
+        return mCurrentModified - mModified;
+    }
+
     template<typename T>
     void Stat<T>::set (const T& value)
     {
+        T diff = value - mBase;
         mBase = mModified = value;
+        mCurrentModified += diff;
     }
+
     template<typename T>
     void Stat<T>::setBase (const T& value)
     {
         T diff = value - mBase;
         mBase = value;
         mModified += diff;
+        mCurrentModified += diff;
     }
+
     template<typename T>
     void Stat<T>::setModified (T value, const T& min, const T& max)
     {
@@ -57,7 +76,15 @@ namespace MWMechanics
 
         mModified = value;
         mBase += diff;
+        mCurrentModified += diff;
     }
+
+    template<typename T>
+    void Stat<T>::setCurrentModified(T value)
+    {
+        mCurrentModified = value;
+    }
+
     template<typename T>
     void Stat<T>::setModifier (const T& modifier)
     {
@@ -65,16 +92,23 @@ namespace MWMechanics
     }
 
     template<typename T>
+    void Stat<T>::setCurrentModifier(const T& modifier)
+    {
+        mCurrentModified = mModified + modifier;
+    }
+
+    template<typename T>
     void Stat<T>::writeState (ESM::StatState<T>& state) const
     {
         state.mBase = mBase;
-        state.mMod = mModified;
+        state.mMod = mCurrentModified;
     }
     template<typename T>
     void Stat<T>::readState (const ESM::StatState<T>& state)
     {
         mBase = state.mBase;
-        mModified = state.mMod;
+        mModified = state.mBase;
+        mCurrentModified = state.mMod;
     }
 
 
@@ -98,6 +132,12 @@ namespace MWMechanics
     {
         return mStatic.getModified();
     }
+    template<typename T>
+    T DynamicStat<T>::getCurrentModified() const
+    {
+        return mStatic.getCurrentModified();
+    }
+
     template<typename T>
     const T& DynamicStat<T>::getCurrent() const
     {
@@ -127,14 +167,21 @@ namespace MWMechanics
             mCurrent = getModified();
     }
     template<typename T>
-    void DynamicStat<T>::setCurrent (const T& value, bool allowDecreaseBelowZero)
+    void DynamicStat<T>::setCurrentModified(T value)
+    {
+        mStatic.setCurrentModified(value);
+    }
+    template<typename T>
+    void DynamicStat<T>::setCurrent (const T& value, bool allowDecreaseBelowZero, bool allowIncreaseAboveModified)
     {
         if (value > mCurrent)
         {
             // increase
-            mCurrent = value;
-
-            if (mCurrent > getModified())
+            if (value <= getModified() || allowIncreaseAboveModified)
+                mCurrent = value;
+            else if (mCurrent > getModified())
+                return;
+            else
                 mCurrent = getModified();
         }
         else if (value > 0 || allowDecreaseBelowZero)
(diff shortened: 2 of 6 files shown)


Svetomech:
Yay! That issue bothered me a great lot in the original game.
*dissapears into the shadows again*

Allofich:
Glad to hear you're happy. If you mean original Morrowind, though, it shouldn't have the problems that this PR fixes. This changes OpenMW to act more like the original game.

apps/openmw/mwmechanics/stat.cpp
```diff
@@ -1,4 +1,5 @@
 #include "stat.hpp"
+#include "limits.h"
```
MiroslavR: Why don't you use the C++ way?

`#include <limits>` and `std::numeric_limits<int>::max()`

Svetomech:
Oh wait, I thought (judging only by a title) the PR fixes [this bug](https://forum.openmw.org/viewtopic.php?f=2&t=3796).

Allofich:
Oh, well, I didn't know about that bug, but it doesn't seem like it should be a problem.

apps/openmw/mwmechanics/stat.cpp
Allofich: OK, fixed. I also changed it to the minimum, rather than the maximum, since it seems very slightly less likely to be used. Probably it should be stored in the saved game file, though, since this PR is already a little bit problematic with existing save games, as I mentioned above. But I'm cautious about adding to the save data.

scrawl:
As you said, the PR doesn't properly work on existing savegames, so it probably shouldn't be merged in this form. As well, I am reluctant to make changes to the stats save format for one particular issue. Fixing one issue at a time will not get us far. What we need is an exhaustive list of all the stats system differences / incompatibilities to vanilla MW (there are definitely more on the bugtracker), discuss which of these changes are intended/bugfixes, and which of these should be changed. The compatibility with .ess imported files may be a factor to consider as well. Then design the new stats system, implement and playtest it, and (ideally) find a way to apply these changes in a way that is compatible with the old save files.

Allofich:
Updated. The new implementation is cleaner and doesn't introduce any new variables. It won't fix stats that are wrongly changed under the current drain/fortify implementation, though, as I couldn't think of a way to do that. My opinion is that it's easy enough to fix with the console, and maybe we can put a warning for the next release that players should remove all drain and fortify effects before using the new version.

Basically this works by using `mMagicEffects` in `creaturestats,` which I found is updated through `adjustMagicEffects` in `actor,` but lags behind the most recent effects, located in `mEffects` of `Spells.` I compare the two and apply any differences between them like instant spell effects, the same way the check for FortifyMaximumMagicka was already supposed to be working. In effect drain and fortify are applied once at spell start, then in reverse at spell end, like in vanilla.

As it turned out, the check for FortifyMaximumMagicka was not working, and in fact in master branch a fortify maximum magicka spell will last forever! The same problem was preventing my attempt for the drain and fortify effects from working. I finally tracked it down to `ExpiryVisitor,` which was prematurely updating the `mMagicEffects` in `creaturestats,` causing the comparisons when effects to not work because no difference between `mMagicEffects` and `mEffects` would be found. This fixed the problem with FortifyMaximumMagicka not working.

I also made some minor changes to how magicka is recalculated to make it work more like vanilla. So far as I've been able to test, everything seems to be working the same as vanilla as far as fortify and drain on health, magicka, fatigue and maximum magicka.

Allofich:
Moved the fortify maximum magicka check out of modifyMagickaEffects and into adjustMagicEffects with the drain and fortify checks, which let me get rid of the bool parameter I added.

Allofich:
@scrawl, just in case you didn't notice, I have updated this and am ready for review.

Regarding your previous post about issues on the bugtracker, I looked at all open bugs on the bugtracker. The ones relevant to drain and fortify effects on health, magicka and fatigue are:

https://bugs.openmw.org/issues/2680
https://bugs.openmw.org/issues/2796
https://bugs.openmw.org/issues/3049

and these are all fixed.

Similar bug reports about wrong magicka amounts in GCDLean:

https://bugs.openmw.org/issues/2471 - can't reproduce. Maybe it's fixed already.
https://bugs.openmw.org/issues/3334 - Fixed
https://bugs.openmw.org/issues/3522 - Fixed

There are also some bug reports about how attributes are handled with fortify and drain. This PR doesn't change anything about attributes.

The only problem I'm aware of is that saves made with the current OpenMW with fortify or drain on will have wrong values after being loaded. Unless I'm mistaken this is unavoidable, though. For example, the following situation, each starting with 100/100 magicka and no drain effect: 

1) Using master branch, you lower magicka to 97/97 through a -3 drain effect, cast a 7 magicka spell and save the game with 90/97 magicka.

2) Using this PR's code, you use "setmagicka 97" to lower magicka to 97/97, cast a 4 magicka spell to lower to 93/97, and then lower to 90/97 with a -3 drain effect. 

In this case you would only want to fix the maximum value back to 100 when loading the save from master branch, but I don't know of any data we can use that differentiates the two cases. I think maybe we just have to take the low-tech approach and advise players to remove drain and fortify effects from their character or any companions before using the new code. Not sure how we should do that, though.

scrawl:
This is a difficult topic, I need to find some time to dive in deep and think about it.

The idea behind the current code was that we can recompute modifiers at any time by starting from zero and just summing all magic effects together. Your changes break this approach by applying persistent changes to the current value as an effect starts or ends. The surrounding code wasn't written with this in mind so there is a chance for new bugs that we may have missed.

Regarding save files, you could increment the version (ESM::SavedGame::sCurrentFormat) and do a conversion when loading older version saves.

Allofich:
That's fine, take your time. Good to hear there is a version number for the save file. I'll see if I can implement a way to correct older saves.

Also, regarding the issues fixed, 

https://bugs.openmw.org/issues/2680
https://bugs.openmw.org/issues/2796
https://bugs.openmw.org/issues/3049

are basically all the same thing, the issue being that the modified value is affected by drain effects rather than the current value, so I think when I have some time later I'll close two of these as duplicates on the bugtracker.

Regarding the GDCLean issues, I think the fix for those can probably be done on its own, so I'll see about setting that up as a separate PR, which maybe can be merged sooner than this one.

Allofich:
> Regarding the GCDLean issues, I think the fix for those can probably be done on its own, so I'll see about setting that up as a separate PR, which maybe can be merged sooner than this one.

Never mind I believe these changes are only relevant if drain and fortify are also changed.

Allofich:
Updated. Changes are:

1. Old saves are repaired now. Saved game version is changed from 3 to 4. Health and Fatigue modifier are given a modifier of 0, while magicka is given one that takes out any fortify/drain magicka modifier while leaving the fortify maximum magicka change.

2. Allow restoring drained attributes and skills. This fixes bug 2446. I had to change the mModifier of attributes from an int to a float for this. For example, the spell "Restore Strength" which is an "effectTick" spell will do nothing otherwise. Older saves are accounted for and work fine. Vanilla behavior was observed and matched for all the cases I tested.

3. General cleanup. I corrected some earlier commits. I thought "Correctly handle setting stat value over modified" had fixed the GCD bugs but actually they were fixed by the "Fix drain and fortify to work more like vanilla" commit.

Allofich:
@scrawl This should be ready for review now. It's working for all my test cases:

1. Opening older save file with active drain effect
2. Opening older save file with active drain magicka effect and fortify magicka effects
3. "Drain magicka" situations with GCDLean active as described in GCDLean bug reports.

And same as vanilla behavior in all of these:
3. Drain/fortify on health, fatigue and magicka affects only current value.
4. Drain can be restored for health, fatigue and magicka, attributes and skills.
5. Drain can be restored before the effect ends or while a constant effect item is equipped. Drain/fortify will reverse on end. If the net drain and fortify modifier is negative, when the effect reverses it can go no higher than 0. meaning it will never reverse you into being "fortified" when it ends. If you have both drain and fortify active and the net modifier is positive, nothing will happen when you try to restore the drained stat or when the drain effect ends. It won't apply in reverse nor will it be "capped" to 0 (which would return it from positive to 0). It could be considered a bug in vanilla that you can't restore a drain effect when the net modifier is positive.
6. There are no such restrictions on fortify lowering or ending. So if you equip a constant effect fortify by 20, then drain the fortified value by 10, since the drain cannot be restored and will not reverse back to 20 even when it ends, when you unequip the constant effect item you will have -10 for the value, since the reversal of the fortify effect is not capped. This could also be considered a bug in vanilla.

Regarding the potential vanilla bugs, I think that, considering the existence of a complex stat mod like GCD, for compatibility and also because these aren't necessarily bugs, we should first of all emulate vanilla.

Allofich:
There was one more fix to make (seems like there always is). I wasn't setting the parameter to allow current health, fatigue and magicka to be set above the modified/base value when loading an older save, so older saves with active fortify effects were losing the effect.

Allofich:
Doing some more work on this after all.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -557,28 +556,25 @@ namespace MWMechanics
...
-            MWWorld::Ptr mActor;
-            float mDuration;
+            private:
+            bool unused;
 
-        public:
-            ExpiryVisitor(const MWWorld::Ptr& actor, float duration)
-                : mActor(actor), mDuration(duration)
+            public:
+            std::vector<std::tuple<float, float, MWMechanics::EffectKey>> mEffectsByRemainingTime;
```
MiroslavR: As much as it is convenient, we cannot use tuples until we switch to C++11 or higher...

Allofich:
I made some more changes. I noticed that when using the rest command, the order that effects would expire was not being adhered to. Specifically, when a fortify maximum magicka effect should be expiring before a fortify maximum effect, the resulting magicka after resting was as if the fortify magicka effect had ended first. Vanilla seems to handle this correctly.

I used a visitor class for this. I had no use for a parameter but EffectSourceVisitor wouldn't accept the visitor class I made unless I gave it a parameter, so I initialized it with a dummy variable, but am I doing something wrong?

apps/openmw/mwmechanics/actors.cpp
Allofich: Do you have a suggestion? I wanted to pull those three pieces of information out from the visitor class.

MiroslavR: A plain old struct?

Allofich:
Hmm, the problem with the order of expiry of those effects didn't happen now when I tried with a build from master. Maybe it was from something I did in this PR, or just a fluke? I didn't see anything to indicate to me that the implementation as it was was respecting the order of expiry... Well I will return to this later.

apps/openmw/mwmechanics/actors.cpp
Allofich: OK, well I'll give that a shot when I next come back to this.

Allofich:
Updated again. Hopefully the CI builds will work. I'm not aware of any problems with in-game functionality at the moment.

Allofich:
So this should be ready for review now, finally.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -595,15 +610,32 @@ namespace MWMechanics
             // in case duration > remaining time of effect.
             // One case where this will happen is when the player uses the rest/wait command
             // while there is a tickable effect active that should expire before the end of the rest/wait.
-            ExpiryVisitor visitor(ptr, duration);
-            creatureStats.getActiveSpells().visitEffectSources(visitor);
+
+            // Sort effects by remaining time, so effects that should end first will be applied first
+            SortByRemainingTimeVisitor sorter(true);
+            creatureStats.getActiveSpells().visitEffectSources(sorter);
+            std::sort(sorter.mEffectsByRemainingTime.begin(), sorter.mEffectsByRemainingTime.end());
```
MiroslavR: I would put this in a SortByRemainingTimeVisitor::sort() method.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -488,33 +552,48 @@ namespace MWMechanics
...
+        {
+        return (mRemainingTime < str.mRemainingTime);
+        }
+
+    };
+
+    class SortByRemainingTimeVisitor : public EffectSourceVisitor
+    {
+            private:
+            bool dummy;
```
MiroslavR: What's the point of this?

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -488,33 +552,48 @@ namespace MWMechanics
 
         float x = fFatigueReturnBase + fFatigueReturnMult * endurance;
 
-        DynamicStat<float> fatigue = stats.getFatigue();
         fatigue.setCurrent (fatigue.getCurrent() + duration * x);
         stats.setFatigue (fatigue);
     }
 
-    class ExpiryVisitor : public EffectSourceVisitor
+    struct effectStruct
```
MiroslavR: Pretty sure this goes against our naming conventions. Personally I would name it just 'Effect' and put it under the 'SortByRemainingTimeVisitor' namespace.

Allofich:
Thanks for your comments, let me fix/respond to them.

apps/openmw/mwmechanics/actors.cpp
Allofich: I had no use for a parameter, but when I tried to make SortByRemainingTimeVisitor have no parameter, I received an error of

a reference of type "MWMechanics::EffectSourceVisitor &" (not const-qualified) cannot be initialized with a value of type "MWMechanics::SortByRemainingTimeVisitor ()"

The only way I could find to fix it was this way. I will gladly remove it if there is a better way.

Allofich:
MiroslavR I tried to address your suggestions. I'm not completely sure I made the fixes correctly so if you could confirm I would appreciate it. Also I'm still using the dummy variable, I haven't been able to find out how to get around it.

apps/openmw/mwmechanics/actors.cpp
Allofich: I don't understand what's going on there, but I just wrote up and will upload a new implementation that doesn't use a dummy and is, hopefully, better overall.

Allofich: I decided to change it back. The new implementation I had was messier in its own way. However, I didn't need to actually make a "dummy" variable, just pass in something as a parameter to the visitor, and the compiler accepted it.

jeffreyhaines:
@Allofich This thread is dense, could you do me a favor? Could you please list and set yourself as the assignee for all the bugs related to this issue you're working on. Appreciate it.

Allofich:
Yeah sure. For at least one of them there was already an assignee, but he doesn't seem to be working on it, so I guess it's OK to put myself there.

scrawl:
There's a problem with the save repairing logic, I think. Won't work for actors in inactive cells, will it?
1. load game (with format = old)
2. stats in active cells get fixed, stats in inactive cells dont get fixed yet
3. save game
4. load game (with format = current)

Do you think we could do the repairing pass within the esm loader?

Allofich:
You were right, the save repair would not work when following those steps. I updated the repair commit and that situation works for me now.

Allofich:
I don't think it's necessary to change the saved game format, actually. Let me update this again.

Allofich:
Updated the saved game fix again. I think that all that needs to be done is to load mModified as mBase for health, magicka and fatigue. If there is a fortify maximum magicka effect, the magicka mModified value will be recalculated anyway.

scrawl:
Sorry for the wait, I haven't had a chance to review the new version in depth. I can't help but think there may be a less complicated way. Perhaps we could extend the stats class to keep track of a "currentModifier", that would allow for a cleaner fix at the expense of a few bytes per object. Save games shouldn't need any special treatment that way since both the modifier and currentModifier would be recomputed every frame. Feel free to experiment with such implementation but you can keep the current version in a branch in case we'll need it.

Allofich:
Just commenting to let you know that I saw your response. I've done some work on a new version but it's not ready yet.

Allofich:
New attempt up. I tried to do it using the concept of the "currentModifier."

As for saved games, I re-purposed mMod in DynamicStat to store the currentModifier. This part should be interchangeable with older saves.

I also added an mRestorationModifier for attributes and skills, though, which will make saves incompatible with older versions. I couldn't think of a way to allow restoring drained/absorbed attributes/skills without saving a restoration modifier. The only way I could think of would be like my old implementation: looking for a change in the magic effects on the actor and applying an effect in reverse when it disappears. But if I did that then we might as well do that for the dynamic stats as well, so that would be the old implementation again.

scrawl:
Right, it's great to have both versions up for comparison. Which one would you prefer? I'm not sure personally. In a way, the currentModifier approach seems cleaner but it's slightly more code and somewhat hack-ish in some places as well.

For the first version, how much of an impact would there be if we just keep storing mMod as an integer?  That way saves would be compatible both ways, too.

Allofich:
I think I prefer the older way. For me the concept is easier to understand, and I also suspect it's closer to what the original engine does. Also I don't think the .ess files keep a "restoreModifier" for the skills and attributes, so adding that was maybe my bad idea, since we should think about conversions from .ess files, but I couldn't see how to avoid it with the newer approach (maybe there is a way). I also think it's good to have more of the code about how spell effects work in spellcasting.cpp rather than in actors.cpp or stat.cpp.

> For the first version, how much of an impact would there be if we just keep storing mMod as an integer? 

Well I need to check the code again to be sure, but I guess we could just cast it to a float, use it, and cast back to an integer when saving. I don't think it would be a huge problem. Maybe it would result in a drained/absorbed stat that had been partially restored changing 1 up or down on save and load, but once the stat was fully restored I think it would be at the normal value again.

Allofich:
I put up the older method again, with mMod saved as an integer.

I also slightly modified how the magic effect update in adjustMagicEffects() in actors.cpp is done. Instead of merging the current update's magic effects list (the "now" list) and the previous list, I just go through each of them, keeping an integer array of the effect keys that were already checked in the first one (the "now" list). Seems less wasteful than merging the lists together (which includes the magnitude values) as I was doing before, when all I really needed was the effect keys.

https://github.com/Allofich/openmw/tree/drain3 has the "mCurrentModifier" implemenation.
https://github.com/Allofich/openmw/tree/drain2 is the older implementation before these new changes.

Allofich:
@scrawl Looks like https://bugs.openmw.org/issues/3049 was just re-targeted for 0.43. I was kind of hoping we could get this PR in for 0.42, but I guess you aren't happy with it?

Allofich:
Updated. I rewrote parts and removed changes not directly related to fixing drain and fortify.

akortunov:
Looks like we totally forgot about this PR. Is there any issues?

scrawl:
Oh, this PR already had its anniversary. Oops.

Well, I'm reluctant to merge it because:

* The code is (inevitably) kind of a mess and I wasn't sure which of the 2 versions I prefer or if there could be another option. 
* The drain/fortify issue seems like a minor thing, we don't receive a lot of complaints about it. Compare this to the amount of code changes/complexity that was required to fix it.

Btw, since this PR was made another issue about the attribute system popped up on the bugtracker: [Attributes and skills are not stored as floats](https://bugs.openmw.org/issues/4021)

Allofich:
When I have some time I'll see if I can clean this up a little more.

Btw, since this PR was made another issue about the attribute system popped
> up on the bugtracker: Attributes and skills are not stored as floats
> <https://bugs.openmw.org/issues/4021>


I think this PR used to change them to floats,because otherwise effects
like "Restore strength" do nothing, because they only add a tiny amount
(less than 1) on every tick, or at least the potion I was testing was like
that. But when I last touched this PR up some months ago I took out changes
to attributes and skills to try to narrow the changes and make the PR
simpler. I think skills and attributes could be fixed in a separate PR from
this.

Allofich:
Updated again, and back to the "currentModifier" approach suggested by scrawl.

I had preferred the other approach before, because I thought it was excessive having "current", "base", "modified" and "currentModified", but this approach is easier to follow code-wise and I think I prefer it now. It was difficult to follow the order of execution and get make sure that everything was being updated in the right order with the other approach, but this way doesn't have that problem.

This only includes fixes for drain and fortify effects on health, magicka and fatigue, nothing with skills or attributes. It includes allowing magicka and fatigue current values to exceed modified due to fortify effects.

Allofich:
Added a fix for https://bugs.openmw.org/issues/4231

Thunderforge:
@zinnschlag Are there any remaining concerns for this PR? I'd hate for it to get to a two year anniversary!

Allofich:
It's fine for merge as far as I'm concerned.

Back when I tested it this worked fine and even if someone has a better idea for implementation (which I doubt there is anyone who has been working on this or thinking about a different way to do it), I think this should be merged to gain the functionality, and then later replace the implementation if someone has a better way.

I don't have an OpenMW development environment set up anymore (I was having trouble using the script for preparing MSVC dependencies last I tried) so I may not be able to respond quickly to any requests for changes.

Capostrophic:
Another month won't make this PR any younger. @zinnschlag I hope you don't mind pinging.

zinnschlag:
Missing changelog entries.

Allofich:
Updated with changelog entries.

PR #1065, opened 2016-09-17
Testing build times for possible packaging

ace13:
Should currently build a similar setup to the current nightlies, though without the LTO I apply on my local builds;
VS2013, x86 + x86_64, Release

Not for merging, just to see what build times look like.
Going to run another test on VS2015 as well to see if there's an appreciable difference.


ace13:
I get the feeling that AppVeyor have **really** beefed up their build machines, considering the fact that they're able to build full Release packages with LTO enabled.

ace13:
Well, that worked. Sort of.

Wonder what's causing every one of the builds to not start.

ace13:
Problem found; It's the Unity builds...

scrawl:
Feel free to open a new one whenever you need more testing.

PR #1067, opened 2016-09-19
Use correct vfx for teleport spells

Allofich:
Fixes https://bugs.openmw.org/issues/2048.

Recreates following behavior observed from original game
1. Divine and Almsivi Intervention remove the spell's "on hit" vfx on the player after teleporting and create a scale 1.0 "VFX_Summon_end" at the post-teleport position of the player. The VFX_Summon_end is stationary and does not follow the player.
2. Recall removes the spell's "on hit" vfx on the player after teleporting but does not add any effects post-teleport. If no Mark spell has been cast, the spell plays the on hit vfx on the player like a normal spell and doesn't do anything. This is also what happens if an NPC casts Recall, Almsivi or Divine Intervention.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -646,14 +646,28 @@ namespace MWMechanics
             if (target != getPlayer())
                 return false;
 
+            MWRender::Animation* anim = MWBase::Environment::get().getWorld()->getAnimation(mCaster);
+
             if (effectId == ESM::MagicEffect::DivineIntervention)
             {
                 MWBase::Environment::get().getWorld()->teleportToClosestMarker(target, "divinemarker");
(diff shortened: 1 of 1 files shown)


Allofich:
I just noticed that the VFX_Summon_end effect after teleporting  seems shorter and smaller when in 1st person mode than in 3rd person mode in OpenMW. Any idea why this would happen?

In the original game it seems to be the same in both 1st and 3rd. You can see the difference in that in the original game, if you cast Divine or Almsivi Intervention in first person and don't move after teleporting, the effect can be seen extending on to the screen, while in OpenMW you won't see it at all unless you back up or look down, etc.

Allofich:
Shoot, looks like I made a mistake, too. It seems the effect _does_ follow the player after teleporting. I could have sworn it did not. I'll make a change for that.

Allofich:
And fixing that also seems to have fixed the above issue. I'm getting the effect showing on the screen just like in the original game now.

PR #1069, opened 2016-09-19
Add unarmed attack selection to chooseBestAttack()

Allofich:
Fixes the animation error mentioned in https://github.com/OpenMW/openmw/pull/1048. You can use the attached save game there to test.

I don't have a good understanding of the AI combat routines, so I'm not sure whether this is the best way to fix this issue, and perhaps @mrcheko or @scrawl would like to offer a different solution.

I have only seen the animation problem mentioned in the other PR with NPCs using bound weapons. I think what might have been happening is that the ESM::Weapon pointer being passed to chooseBestAttack() was inaccurate (pointing to a no longer equipped weapon? emptied?) because the weapon changed due to a bound weapon spell. The problem doesn't always occur, so it might be tied to something like an animation being interrupted by a hit or a spell casting, etc. I addressed the problem by getting the weapon from the active weapon slot immediately before calling chooseBestAttack(), to be sure the weapon pointer is accurate.

I also noticed that MSVC was giving a warning about chooseBestAttack() not giving a return value in the case of a NULL weapon, so to silence that I added random attack selection for that case instead of relying on the random attack selection in the character controller. Adding this in actually would have masked this animation problem, though. I verified that this PR fixes the animation problem through updating the weapon pointer, not through the random attack selection.

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -362,9 +362,9 @@ namespace MWMechanics
             }
         }
 
-        // Original engine behavior seems to be to back up during ranged combat
-        // according to fCombatDistance or opponent's weapon range, unless opponent
-        // is also using a ranged weapon
+        // Below behavior for backing up during ranged combat differs from vanilla.
+        // Vanilla is observed as backing up only as far as fCombatDistance or
+        // opponent's weapon range, or not backing up if opponent is also using a ranged weapon
         if (isDistantCombat && distToTarget < rangeAttack / 4)
         {
             mMovement.mPosition[1] = -1;
@@ -468,6 +468,8 @@ std::string chooseBestAttack(const ESM::Weapon* weapon)
         else
             attackType = "chop";
     }
+    else
+        MWMechanics::CharacterController::setAttackTypeRandomly(attackType);
 
     return attackType;
 }
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1312,9 +1312,9 @@ bool CharacterController::updateWeaponState()
                     mAttackType = "shoot";
                 else
                 {
-                    if (isWeapon)
+                    if(mPtr == getPlayer())
                     {
-                        if(mPtr == getPlayer())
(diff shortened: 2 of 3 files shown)


Allofich:
> (pointing to a no longer equipped weapon? emptied?) 

As far as I can see, it's always happening with a NULL weapon pointer.
I also just found it happen in OpenMW master with an NPC without a bound weapon, but who used a potion. I think maybe its using potions, not bound weapons (or maybe both?) that is related.

mrcheko:
Your assumption about potions causing the issue were correct. Check #1070.
And about moving random attack selection - I think it is a right choice. Makes things more clear.

Allofich:
After your PR is merged, I'll test and make sure the change here isn't necessary, then change this PR to just add the random attack selection to combat AI. It is a small amount of code duplication from the character controller, but yes I agree it does make it clearer what is happening, and less likely for someone changing the code to break something.

Allofich:
Looks like mrcheko's PR https://github.com/OpenMW/openmw/pull/1070 has taken care of the issue. So this is just to add the unarmed selection to chooseBestAttack() in aicombat now, for the reasons mentioned above, if scrawl agrees it should be added.

Allofich:
While I'm at it, I rewrote a difficult to understand comment regarding original Morrowind combat AI behavior I added in recently.

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -468,7 +468,16 @@ std::string chooseBestAttack(const ESM::Weapon* weapon)
...
+    else
+    {
+        float random = Misc::Rng::rollProbability();
+        if (random >= 2/3.f)
+        attackType = "thrust";
+        else if (random >= 1/3.f)
+        attackType = "slash";
+        else
+        attackType = "chop";
+    }
```
mrcheko: The better option would be to rewrite the method `CharacterController::setAttackTypeRandomly()` as static and use that here and in the `CharacterController::updateWeaponState` to avoid code duplication. Character.hpp is already included in AiCombat.

Allofich: Thanks for the tip. I'll do that.

PR #1070, opened 2016-09-19
fix using potion as weapon

mrcheko:
fixes broken ai combat issue from https://github.com/OpenMW/openmw/pull/1048#issuecomment-247075756


scrawl:
Please add a new virtual method e.g. ```Action::isAttack``` to replace the dynamic_cast.

PR #1074, opened 2016-09-22
Implement using telekinesis with touch spells

Allofich:
For https://bugs.openmw.org/issues/3560.

In original Morrowind, when telekinesis is active you can use touch spells on things effected by telekinesis. In practice, this means you can use the "Lock" and "Open" touch spells at range by combining them with telekinesis.

This opened up another small issue, in that it would allow combining telekinesis with area effect explosions from touch spells to remotely hit NPCs by casting on nearby objects. While it might be a kind of interesting gameplay mechanic, original Morrowind seems to intentionally not allow this. In original Morrowind, if you cast a touch spell with an area explosion on an non-actor object, normally the game does play a silent explosion effect. This silent explosion also happens if you combine a touch spell with telekinesis. But it won't create the explosion if its near an NPC, so you can't, as far as I can see, use telekinesis to remotely explode touch spells and hit NPCs with the area damage.

We could put in this exception but the silent explosions are odd anyway and have no gameplay use it would seem. For now at least, I just took out touch effects causing an explosion in OpenMW unless they are used on an actor.

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -3187,6 +3187,9 @@ namespace MWWorld
             if (effectIt->mRange != rangeType || (effectIt->mArea <= 0 && !ignore.isEmpty() && ignore.getClass().isActor()))
                 continue; // Not right range type, or not area effect and hit an actor
 
+            if (effectIt->mRange == ESM::RT_Touch && (!ignore.isEmpty()) && (!ignore.getClass().isActor() && !ignore.getClass().canBeActivated(ignore)))
+                continue; // Don't play explosion for touch spells on non-activatable objects
+
             // Spawn the explosion orb effect
             const ESM::Static* areaStatic;
             if (!effect->mArea.empty())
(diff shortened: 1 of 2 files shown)


Allofich:
Hold on this please. Looks like it _is_ possible to hit NPCs with area damage like I said, as long as the spell area is big enough.

Allofich:
Changed the second commit. Explosions occur with area touch spells, but only on actors or on activatable objects, which is the same as the original engine. It doesn't really make sense that you could create an explosion on an activatable object but not a non-activatable one, but I'm not sure what to do so for now I'm just copying the original engine. If we went the other direction I suppose we should make explosions from touch spells occur on terrain, which they currently don't.

Allofich:
I will touch this up a little bit later today.

Allofich:
A small change to allow telekinesis + area effect touch spell to also work on doors that load into another cell. In the original engine this can't be done, but that is probably just a leftover from using the telekinesis targeting that disallows interacting with cell-loading doors. There isn't really any reason to disallow those doors for touch spells while allowing other doors, since there isn't the issue of the player "teleporting" over to and through the door like would happen with activation.

PR #1075, opened 2016-09-23
Flavor Text support

kunesj:
Adds in "Flavor Text" text field in OpenMW-CS for all item types. 
- Items that don't have this value empty will display tooltip with flavor text next to standard tooltip (after specified delay).
- Flavor text will be displayed only in "Inventory mode" (right click, merchants, etc...)
- Width of flavor text tooltip and delay between displaying standard tooltip and flavor text tooltip can be changed with new options in settings-default.cfg

Screenshots: https://imgur.com/a/7yGT9
Forum thread: https://forum.openmw.org/viewtopic.php?f=3&t=3802

Note: 
I am still not sure if I completely understand how code for OpenMW-CS works, so someone who does should probably check that I did not forget to do something important.

EDIT:
bug ticket: https://bugs.openmw.org/issues/3688

### DIFF
--- a/apps/opencs/model/world/columns.cpp
+++ b/apps/opencs/model/world/columns.cpp
@@ -330,6 +330,8 @@ namespace CSMWorld
             { ColumnId_WeatherName, "Type" },
             { ColumnId_WeatherChance, "Percent Chance" },
 
+            { ColumnId_FlavorText, "Flavor Text" },
+
             { ColumnId_UseValue1, "Use value 1" },
             { ColumnId_UseValue2, "Use value 2" },
             { ColumnId_UseValue3, "Use value 3" },
--- a/apps/opencs/model/world/columns.hpp
+++ b/apps/opencs/model/world/columns.hpp
@@ -329,6 +329,8 @@ namespace CSMWorld
             ColumnId_WeatherName = 295,
             ColumnId_WeatherChance = 296,
 
+            ColumnId_FlavorText = 297,
+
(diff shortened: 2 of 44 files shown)


scrawl:
Zini should weigh in on this, but from what I remember, the plan was to keep the ESM format compatible with the vanilla game until after version 1.0, so the changes to the format would be out of the question for now.

kunesj:
I was under the impression that editor saved addons as *.omwaddon because they were not compatible...

scrawl:
.omwaddon/.omwgame in theory can be renamed to .esp/.esm and should work in the vanilla game (barring any bugs in the editor).

Aesylwinn:
Yes, in theory it should work, but honestly, no one tests that... Anyway, it looks like the Morrowind engine ignores data it doesn't use. Converting an omwaddon with the added flavor text to an esp file doesn't cause it to outright crash.

kcat:
Any omwaddon/game that uses scripts won't immediately work as esp/m files since they're not compiled.

jeffreyhaines:
@kunesj do you have an active bug ticket for this issue? One needs to be created before we proceed to merge and/or close this pull request. Please post a link to that new ticket here, thanks

kunesj:
Here is the bug ticket. https://bugs.openmw.org/issues/3688

jeffreyhaines:
Thanks :)

zinnschlag:
Post 1.0 feature. Sorry.

PR #1076, opened 2016-09-23
New option in settings allowing to add multiple identical effects in spellcreation

kunesj:
This small pull request adds option in settings to disable checking for identical effects in spell creation.

This makes it possible to for example create spells with both:
- firedamage 30-50pts 1s
- firedamage 1-4pts 10s  

old forum thread: https://forum.openmw.org/viewtopic.php?f=2&t=3545

EDIT:
bug ticket: https://bugs.openmw.org/issues/3687

### DIFF
--- a/files/settings-default.cfg
+++ b/files/settings-default.cfg
@@ -145,6 +145,10 @@ difficulty = 0
 # Show duration of magic effect and lights in the spells window.
 show effect duration = false
 
+# Do not allow to add same effect multiple times in spell creation
+# Example: [fire damage 10pts for 1s]+[fire damage 1-2pts for 10s] is possible if disabled
+unique spell effects = true
+
 [General]
 
 # Anisotropy reduces distortion in textures at low angles (e.g. 0 to 16).
--- a/apps/openmw/mwgui/spellcreationdialog.cpp
+++ b/apps/openmw/mwgui/spellcreationdialog.cpp
@@ -5,6 +5,7 @@
 
 #include <components/esm/records.hpp>
 #include <components/widgets/list.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/windowmanager.hpp"
 #include "../mwbase/soundmanager.hpp"
@@ -498,6 +499,8 @@ namespace MWGui
         mAddEffectDialog.eventEffectRemoved += MyGUI::newDelegate(this, &EffectEditorBase::onEffectRemoved);
 
         mAddEffectDialog.setVisible (false);
+
+        mUniqueEffectsOnly = Settings::Manager::getBool("unique spell effects", "Game");
     }
 
     EffectEditorBase::~EffectEditorBase()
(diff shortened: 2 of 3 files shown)


kcat:
Is this something that should have a settings option? Seems to me it either should always or never be allowed.

kunesj:
OpenMW before 1.0 is supposed to be replacement of vanilla engine, so I needed to make this behavior optional. That's why it's in settings.

If we were already over 1.0, then I would have advocated to just remove the part of code that checks for identical effects altogether. Because its only purpose seems to be to force player to create fireballs that freeze/electrocute enemy over time...

kcat:
OpenMW post-1.0 is still going to be a vanilla engine replacement. It's not as if after 1.0 we're going to do anything to break Morrowind or significantly alter support beyond improving what it's capable of (playing Morrowind unmodded, OpenMW post-1.0 should mechanically be the same as 1.0). It's just a matter of encouraging people to put the effort into working on extra features (pre-1.0, people should focus on getting the missing vanilla features implemented to get to 1.0 faster, instead of taking time with new features). Given that you've already put the effort into coding it though, in my opinion the question becomes if it's a feature OpenMW wants at all and if it's good enough to put in (in particular, does it break balancing? are multiple identical effects cheaper than one effect with the same combined strength?).

IMO, if it's good enough to have on by default after 1.0, it's good enough to have on by default before 1.0, given that it's already coded. If it's not good enough to have on by default after 1.0, then I don't think a settings option is the right way to go about it regardless.

kunesj:
Spell_one:
* Fire Damage 100 pts for 3 sec on Touch

Cost: 100, Chance: 25, Price: 787

Spell_two:
* Fire Damage 50 pts for 3 sec on Touch
* Fire Damage 50 pts for 3 sec on Touch

Cost: 100, Chance: 25, Price: 788

Spell_eight (max number):
* Fire Damage 12 pts for 3 sec on Touch
* Fire Damage 12 pts for 3 sec on Touch
* Fire Damage 12 pts for 3 sec on Touch
* Fire Damage 12 pts for 3 sec on Touch
* Fire Damage 13 pts for 3 sec on Touch
* Fire Damage 13 pts for 3 sec on Touch
* Fire Damage 13 pts for 3 sec on Touch
* Fire Damage 13 pts for 3 sec on Touch

Cost: 101, Chance: 24, Price: 795

Tested with Willpower=100, Destruction=50

Thankfully it doesn't break balancing at all. I have also not noticed any bugs that this would create while testing it.

files/settings-default.cfg
```diff
@@ -145,6 +145,10 @@ difficulty = 0
 # Show duration of magic effect and lights in the spells window.
 show effect duration = false
 
+# Do not allow to add same effect multiple times in spell creation
+# Example: [fire damage 10pts for 1s]+[fire damage 1-2pts for 10s] is possible if disabled
+unique spell effects = true
```
MiroslavR: Indeed. If one still has 0.35 settings-default.cfg, that means they have installed 0.41 improperly.

Also, `Settings::Manager::getBool("unique spell effects", "Game")` will throw, not return false, when the option is missing:
```
    throw std::runtime_error(std::string("Trying to retrieve a non-existing setting: ") + setting
                             + ".\nMake sure the settings-default.cfg file was properly installed.");
```

jeffreyhaines:
@kunesj do you have an active bug ticket for this issue? One needs to be created before we proceed to merge and/or close this pull request. Thanks

kunesj:
here it is https://bugs.openmw.org/issues/3687

jeffreyhaines:
Thank you :)

scrawl:
Post 1.0 feature.

PR #1077, opened 2016-09-24
OpenMW-CS: Only display editable nested tables (Bug #3277)

rcutmore:
This is an idea for fixing [bug 3277](http://bugs.openmw.org/issues/3277). Instead of disabling the table this sets it to disallow edits, selection, and focus. The background color is also changed (not doing so  causes the table items to not appear disabled). Doing this instead of disabling the table allows the user to still scroll through the table items.

### DIFF
--- a/apps/opencs/view/world/dialoguesubview.cpp
+++ b/apps/opencs/view/world/dialoguesubview.cpp
@@ -578,40 +578,31 @@ void CSVWorld::EditWidget::remake(int row)
                         fixedRows = true;
                 }
 
-                NestedTable* table =
-                    new NestedTable(mDocument, id, mNestedModels.back(), this, editable, fixedRows);
-                table->resizeColumnsToContents();
-                if (!editable)
+                // Create and display nested table only if it's editable.
+                if (editable)
                 {
-                    // Disallow table editing, selection, and focus.
-                    // This allows user to scroll through table but not edit items.
-                    // Disabling table doesn't allow user to scroll.
-                    table->setEditTriggers(QAbstractItemView::NoEditTriggers);
-                    table->setSelectionMode(QAbstractItemView::NoSelection);
-                    table->setFocusPolicy(Qt::NoFocus);
-                    table->setToolTip("Currently read-only");
+                    NestedTable* table =
+                        new NestedTable(mDocument, id, mNestedModels.back(), this, editable, fixedRows);
+                    table->resizeColumnsToContents();
+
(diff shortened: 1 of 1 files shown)


zinnschlag:
Not sure about this one. Manual colour changes sound problematic, because OpenMW-CS is used in a wide variety of desktop environments with very different colour schemes.

I think the best solution would be to copy what the regular (non-nested) tables do.

rcutmore:
Do you know off-hand where non-nested tables get disabled but are still scrollable?

I removed the color change and set a tooltip. The table header text is disabled so there is some indication that the table itself is disabled. This is how that looks:
screenshot at 2016-09-25 06 29 42

zinnschlag:
> Do you know off-hand where non-nested tables get disabled but are still scrollable?

I don't think they get disabled. They are just configured to not accept changes in cells.

As for tooltips and other hints: I don't like the approach. We don't lock individual tables. Either the whole document is locked for editing or not. Any hint on individual UI elements is just piecemeal. Maybe instead we need a general (Window-wide) indication that the document can't be edited. I haven't given this much thought yet. Here are some random ideas:
* indicate it somehow in the titlebar (icon or text)
* add a horizontal bar at the top of the window that says "Document locked for editing" or something

rcutmore:
@zinnschlag I'm not sure I understand, I might be misinterpreting. Are you maybe suggesting to change the nested table header to indicate it's locked for editing? So in my previous screenshot the headers would read *NPC Attributes (locked)* and *NPC Skill (locked)* (or something similar)?

The issue is that only those two nested tables are locked (because of the *Auto Calc* field being selected) while the rest of the object attributes are editable. That's why the headers for those two tables are greyed out in my screenshot, while the one above is not. It seems like it would be confusing to have some sort of window-wide indication as opposed to just indicating the locked elements.

zinnschlag:
Oh dear. It seems we are speaking about completely different topics here. I thought we were discussing the global locking mechanism. Sorry for wasting your time.

IIRC there is no auto calc in the editor currently, right? We have an issue about that on the tracker. There is a good chance that we will never get this feature (because of post 1.0 changes, discussed elsewhere). I think the easiest option would be to just remove the table when auto calc is enabled. Having GUI elements in the dialogue that one can not interact with and that do not provide any useful information is pointless.

rcutmore:
@zinnschlag No problem, I should have been clearer when I first opened this PR.

I pushed a change to avoid adding the nested table when it's uneditable. Let me know if this isn't what you had in mind.

PR #1080, opened 2016-09-27
add support for GLES, when compiling on armhf/armv7 devices

psi29a:
Includes the necessary flags for compiling against GLES instead of GL. This is a noop on most systems, but required on armhf.

This should not impact Android builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -511,7 +511,7 @@ endif(WIN32)
 # Extern
 add_subdirectory (extern/osg-ffmpeg-videoplayer)
 add_subdirectory (extern/oics)
-if (USE_QT)
+if (BUILD_OPENCS)
     add_subdirectory (extern/osgQt)
 endif()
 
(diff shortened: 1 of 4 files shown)


scrawl:
This is wrong. We don't explicitely include GL, OSG does it for us. Your compile issue is likely solved by building OSG with the correct compiler flags.

scrawl:
Feel free to open new PR for the USE_QT / BUILD_OPENCS bit.

scrawl:
Btw, have you tested if OpenMW runs correctly under GLES? From what I remember, sandstranger encountered rendering bugs. I don't see a point in working on packaging support for a nonfunctioning configuration.

psi29a:
This is in relation to this thread here where I asked for your input: 
https://forum.openmw.org/viewtopic.php?f=6&t=3814

The OSG build is from Debian itself for armhf.
This is how it is built here:
https://anonscm.debian.org/cgit/pkg-osg/openscenegraph-3.4.git/tree/debian/rules

They specifically use GLESv2 and only GLESv2 when building OSG-3.4

So any help you can provide to make this clearer would be appreciated.

PR #1081, opened 2016-09-28
make BUILD_OPENCS a dependency for building osgQt

psi29a:


scrawl:
Wrong commit message.

psi29a:
Better?

PR #1082, opened 2016-09-29
Fix for segmentation fault on entering a cell

Gladdy:
OpenMW 0.40 crashes whenever you enter a cell in which there is an NPC attacking you. 

## How to reproduce:
- Attack a NPC indoors
- move towards a door to an external cel
- as you're being attacked, activate the door
- on reentering the cell, the game will crash with a segmentation fault (SIGNAL 11)

## Reason: 
dereferencing a null pointer. ( "currentAction->isAttackingOrSpell()" in aicombat.cpp:235)

## Crash log: 
[mwcrash.txt](https://github.com/OpenMW/openmw/files/501906/mwcrash.txt)

## Fix
- Check whether the shared_ptr is NULL by using std::shared_ptr::operator bool, if so, prevent the attack by returning from AiCombat::attack(...).
- This only seems to happen on the very first entry to a new cell and does not interfere with combat.

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -229,6 +229,9 @@ namespace MWMechanics
         osg::Vec3f vAimDir = MWBase::Environment::get().getWorld()->aimToTarget(actor, target);
         float distToTarget = MWBase::Environment::get().getWorld()->getHitDistance(actor, target);
 
+        if (!currentAction)
+            return;
+
         storage.mReadyToAttack = (currentAction->isAttackingOrSpell() && distToTarget <= rangeAttack);
         
         // can't fight if attacker can't go where target is.  E.g. A fish can't attack person on land.


Allofich:
I can't reproduce the segmentation fault on the OpenMW 0.40 release or with the latest code. I'm running on Windows 7.

Gladdy:
[Try_entering_the_council_club.omwsave.zip](https://github.com/OpenMW/openmw/files/504340/Try_entering_the_council_club.omwsave.zip)

I'm running Ubuntu 16.04, I've attached a save file on which this is happening. Could you give that a go? 
I'm basing my change on commit 85c7d014d32df3c0d8a291854c4a6b7362e5643d.

scrawl:
I haven't tried replicating your issue but I will merge this PR as the added nullpointer check looks like a harmless change.

Allofich:
I couldn't replicate the issue by following the steps above, but with that save file I did get a crash. I'll test with the new master now to see if this PR fixed it for me on Windows.

Allofich:
It's fixed, nice work.

PR #1083, opened 2016-10-01
Make AI response to spell hits more like original MW

Allofich:
For https://bugs.openmw.org/issues/3576

There is one other difference I noticed between AI reaction to hostile spells in OpenMW and original Morrowind. In OpenMW a "friendly hit" voice will play for characters hit by spells, while in original Morrowind no voice will play, they will only do so when hit physically. I decided, for now at least, that OpenMW's behavior is probably better on that point, so I didn't change it. All of the on-hit voices still make sense if hit by a spell.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -411,6 +411,10 @@ namespace MWMechanics
             float magnitudeMult = 1;
             if (magicEffect->mData.mFlags & ESM::MagicEffect::Harmful && target.getClass().isActor())
             {
+                // Notify the target actor they've been hit
+                if (target != caster && !caster.isEmpty())
+                    target.getClass().onHit(target, 0.0f, true, MWWorld::Ptr(), caster, osg::Vec3f(), true);
+
                 if (absorbed)
                     continue;
 
@@ -443,10 +447,6 @@ namespace MWMechanics
                 // If player is attempting to cast a harmful spell, show the target's HP bar
                 if (castByPlayer && target != caster)
                     MWBase::Environment::get().getWindowManager()->setEnemy(target);
-
-                // Notify the target actor they've been hit
-                if (target != caster && !caster.isEmpty())
-                    target.getClass().onHit(target, 0.0f, true, MWWorld::Ptr(), caster, osg::Vec3f(), true);
             }
 
             if (magnitudeMult > 0 && !absorbed)
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -412,7 +412,7 @@ namespace MWMechanics
             if (magicEffect->mData.mFlags & ESM::MagicEffect::Harmful && target.getClass().isActor())
             {
                 if (absorbed)
-                    break;
+                    continue;
 
                 // Try reflecting
                 if (!reflected && !caster.isEmpty() && caster != target && !(magicEffect->mData.mFlags & ESM::MagicEffect::Unreflectable))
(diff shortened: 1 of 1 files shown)


apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -433,22 +426,28 @@ namespace MWMechanics
                                     "meshes\\" + reflectStatic->mModel, ESM::MagicEffect::Reflect, false, "");
                         reflectedEffects.mList.push_back(*effectIt);
                         magnitudeMult = 0;
+                        break;
```
scrawl: Shouldn't this read ```continue;```? If an effect is reflected, the rest of the effects list is not evaluated, which seems wrong. The ```if (absorbed)``` statement seems to have the same issue, although that was already there before the PR.

Allofich: Yes. Just to be sure, I tested in the original game and confirmed that some effects of a spell can be absorbed or reflected while others aren't.

Allofich: Specifically, effects further on the list continue to be checked even after one is absorbed/reflected.

Allofich:
Travis warning:
/home/travis/build/OpenMW/openmw/apps/openmw/mwmechanics/spellcasting.cpp:428:25: warning: Value stored to 'magnitudeMult' is never read
                        magnitudeMult = 0;

Allofich:
Stupid me... got the commit message backwards.

Allofich:
Another small thing this PR does. In OpenMW the health bar would show for an actor you cast upon even when the effect was reflected (but not when absorbed). This doesn't happen in original Morrowind so I changed it.

Allofich:
Sorry, that line was being used it seems. The warning was for when breaks were being used. Though it didn't seem to break anything by removing it.

Allofich:
Well never mind it's _not_ being used, so I'll take it out again. The following section in
`if (magnitudeMult > 0 && !absorbed)`
 is running, and it seems like an actor reflecting something would still get affected by the spell, but instead it's working correctly for some reason. I want to study the code a bit more.

Allofich:
Well, some time has passed and I thought about this again, and as often happens I've changed my mind about the implementation.

I had decided to bring the behavior largely in line with that of original Morrowind, but I don't like that with this PR you can cast a hostile spell at someone and the guards don't react if it was absorbed/reflected. This is especially because, as far as I know, there are no 100% absorption or reflection abilities in vanilla Morrowind, so you are always threatening someone when you cast a hostile spell at them, and it should be considered an assault. This also would be more consistent with physical attacks, where even if you miss it is considered an assault.

There is a minor issue with the on-hit voices for friendly hits. They are a mixture of warnings ("Careful with that, friend.") and pain sounds. The pain sounds don't make much sense for reflected or absorbed spells unless you imagine those as hurting. This is an issue in original MW, too, though, because the pain sounds will play even when you miss with a physical attack.

So, I'd like to put up a new implementation here considering both absorbed and reflected hostile effects as attacks.

It's also arguable whether changing the behavior to count each hostile effect as a separate attack, as original MW does, is better than counting by spell as OpenMW was doing. It seems fine to me, though. Usually a spell with multiple hostile effects would be more damaging, I would think, so it makes sense to more quickly cause a hostile response. It also makes things a little less uniform compared to it always taking exactly four hits to cause a follower to go hostile, so it's a bit more natural-seeming.

Allofich:
Uploaded. The previous implementation is at https://github.com/Allofich/openmw/tree/onhit2.

Allofich:
Yet another thing to consider: Do some players give a follower a spell absorption ability, and replenish their magicka by casting hostile spells at them? You could just use non-hostile spells to avoid the problem, but maybe one wouldn't be available. Even when using hostile spells, this situation could be seen as an agreement of trust between the player and their follower. For this case, maybe the original MW behavior would be better. Really not sure what to do here.

Allofich:
It's a few hours later, and coming back to this and thinking on it again, the "casting on a follower to replenish their magicka" is probably too edge case to really matter. So, it's more a question of whether we want to stick to original MW behavior or not.

PR #1093, opened 2016-10-05
Android crash fixes

sandstranger:
I will add again this fixes

### DIFF
--- a/apps/openmw/mwgui/mapwindow.cpp
+++ b/apps/openmw/mwgui/mapwindow.cpp
@@ -165,11 +165,7 @@ namespace MWGui
         , mCompass(NULL)
         , mChanged(true)
         , mFogOfWarToggled(true)
-#if defined (ANDROID) && !defined (OPENGL_ES)
-        , mFogOfWarEnabled(false)
-#else
         , mFogOfWarEnabled(fogOfWarEnabled)
-#endif
         , mMapWidgetSize(0)
         , mNumCells(0)
         , mCellDistance(0)
@@ -232,6 +228,9 @@ namespace MWGui
     bool LocalMapBase::toggleFogOfWar()
     {
         mFogOfWarToggled = !mFogOfWarToggled;
+#if defined (ANDROID)
+        mFogOfWarEnabled = false;
+#endif
         applyFogOfWar();
         return mFogOfWarToggled;
     }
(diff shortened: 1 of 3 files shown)


apps/openmw/mwgui/mapwindow.cpp
```diff
@@ -165,7 +165,11 @@ namespace MWGui
         , mCompass(NULL)
         , mChanged(true)
         , mFogOfWarToggled(true)
+#if defined (ANDROID) && !defined (OPENGL_ES)
+        , mFogOfWarEnabled(false)
```
kpp: And you get warning `unused argument`. Can you move this code where it is called? Outside ctor?

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -1290,8 +1290,11 @@ namespace MWPhysics
         mMovementResults.clear();
 
         mTimeAccum += dt;
+#if defined (ANDROID)
+        const float physicsDt = 1.f/30.0f;
```
kpp: Can you describe these magic numbers?

components/myguiplatform/myguirendermanager.cpp
```diff
@@ -128,8 +128,11 @@ class Drawable : public osg::Drawable {
             osg::Texture2D* texture = batch.mTexture;
             if(texture)
                 state->applyTextureAttribute(0, texture);
-
+#if defined (ANDROID) || defined (OPENGL_ES)
+            osg::GLBufferObject* bufferobject = 0;
```
kpp: Can you comment these lines? Why can't we create `vbo->getOrCreateGLBufferObject` even if `state->isVertexBufferObjectSupported`?

apps/openmw/mwphysics/physicssystem.cpp
sandstranger: arm processors is shit , so it need to increase physics step

components/myguiplatform/myguirendermanager.cpp
sandstranger: it is bug in osg in gles render , for example mali t760 supports vbo , but  game crashes inside osg sources

kpp: Will you comment it inside the code?

sandstranger:
I close it , because it not actual anymore

PR #1094, opened 2016-10-06
Fix reversed use of alembics and retorts

Allofich:
For https://bugs.openmw.org/issues/3574


Allofich:
I tested in original MW and OpenMW and correct numbers are coming out for positive and negative effects from potions now. I can provide a sample save for convenience of testing if wanted.

PR #1095, opened 2016-10-08
Make actors with non-portable lights in inventory glow

MiroslavR:
Adding non-portable lights (such as "blade light" or "yellow light") to an actor's inventory should cause him to glow. This only applies to creatures and NPCs, not to containers.

Implements:
* https://bugs.openmw.org/issues/2042
* https://bugs.openmw.org/issues/3338

### DIFF
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -23,7 +23,7 @@ add_openmw_dir (mwrender
     actors objects renderingmanager animation rotatecontroller sky npcanimation vismask
     creatureanimation effectmanager util renderinginterface pathgrid rendermode weaponanimation
     bulletdebugdraw globalmap characterpreview camera localmap water terrainstorage ripplesimulation
-    renderbin
+    renderbin actoranimation
     )
 
 add_openmw_dir (mwinput
--- a/apps/openmw/mwrender/actoranimation.cpp
+++ b/apps/openmw/mwrender/actoranimation.cpp
@@ -1,0 +1,155 @@
+#include "actoranimation.hpp"
+
+#include <utility>
+
+#include <osg/Node>
+#include <osg/Group>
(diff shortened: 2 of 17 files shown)


psi29a:
:+1:

scrawl:
I have tested with the trueflame (nerevarblade_01_flame). In the original engine, the light is stronger, and less directional. Looking at the showSceneGraph output, it appears that the ambient component of the light gets set to (1,1,1), in addition to the diffuse component specified by the light record.

Also, in the original engine, the light source will not affect the actor itself. There isn't a proper way to place the light source without it lighting the actor in a weird way. Right now, the light is placed at the actor's feet (same as the original engine, I think) which looks especially weird.

To implement this, perhaps the LightListCallback could be given a list of LightSources to ignore.

MiroslavR:
Changes:
* The light source no longer illuminates the actor. This is done via a list of ignored LightSources in LightListCallback.
* The ambient component of the light is now set to (1, 1, 1, 1).

In case you're wondering about _objectRootReset()_: since switching view mode recreates the object root along with the LightListCallback, ignored LightSources must be readded.

scrawl:
I made a few changes. The pointer to the LightListCallback is now stored in the Animation, which eliminates the need for dynamic_cast. Also, when the object root is recreated, the previously used LightListCallback will be reused, so we no longer need the objectRootReset() notifier.

Finally, there was a bug when saving and reloading the game, the getIgnoredLightSources() were not being set, as the ActorAnimation constructor completes before the NpcAnimation sets the ObjectRoot. This was solved by creating the LightListCallback in advance in the Animation constructor.

https://github.com/scrawl/openmw/commit/1893617ec98491c5729c9d5bd59d24814d7567e8

Please give it a test to make sure I'm not breaking anything, and then we can merge.

MiroslavR:
LGTM. :+1:

PR #1097, opened 2016-10-08
Don't let intimidate bring disposition below 0 (Fixes #3584)

Allofich:
Failed intimidate attempts had no check to prevent them from bringing disposition below 0. Testing in original MW indicates that disposition doesn't go below 0 there.

### DIFF
--- a/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
+++ b/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
@@ -448,6 +448,12 @@ namespace MWDialogue
     {
         MWBase::Environment::get().getWindowManager()->removeGuiMode(MWGui::GM_Dialogue);
 
+        // Clamp permanent disposition change so that final disposition doesn't go below 0 (could happen with intimidate)
+        
+        float curDisp = static_cast<float>(MWBase::Environment::get().getMechanicsManager()->getDerivedDisposition(mActor, false));
+        if (curDisp + mPermanentDispositionChange < 0)
+            mPermanentDispositionChange = -curDisp;
+
         // Apply disposition change to NPC's base disposition
         if (mActor.getClass().isNpc())
         {


Allofich:
Don't merge yet, please.

scrawl:
In theory, the disposition could also go out of the 0-100 range if fPerTempMult was != 1 (e.g. 0.5 or 2).

Perhaps the cap should be applied [here](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwdialogue/dialoguemanagerimp.cpp#L537) instead.

scrawl:
This might help you in testing:
> You can see these values in the console by using ToggleDialogStats before persuading.

https://wiki.openmw.org/index.php?title=Research:Disposition_and_Persuasion#Disposition_Change

Allofich:
Thanks, I know about that command and I'm using it. Right now this PR is using the wrong value in the calculation. I've fixed that locally already but after a break I'll spend some more time testing things.

psi29a:
You're doing a great job Allofich, thank you! :)

Allofich:
psi29a, thanks.

I've done some testing with the tds and getDisposition commands and adjusting the fPerTempMult value. The fPerTempMult value doesn't seem to be relevant here, the permanent disposition adjustment doesn't seem to be affected by it. The permanent disposition adjustment can go below 0 in original MW, seemingly infinitely (I tested as far as -500). It won't go above 100, though.

What's most relevant to this PR, is that the final disposition is clamped to 0 to 100 at the end of dialogue, so even if the permanent disposition adjustment was -500, you would end with 0. This is different from OpenMW, where even though the disposition meter shows 0, it will actually still be in negative territory after a dialogue finishes if you brought it down that low (which is what caused the bug report). Right now I'm working on an implementation that just clamps the disposition when dialogue is closed.

Allofich:
As far as I can see, everything seems to be working fine with this new implementation.

Allofich:
Changed it to only clamp to avoid going below 0, as that is all that is necessary.

MiroslavR:
The code you've added should be under `if (mActor.getClass().isNpc())` because creatures do not have disposition (NpcStats).

For instance, if you talk to Creeper in "Caldera, Ghorak Manor" and close the dialogue window, an exception propagates up to the frame listener:
```
Error in framelistener: class does not have NPC stats
```

Seems harmless, but for creatures that also provide travel services this can have game-breaking impact -- see https://forum.openmw.org/viewtopic.php?f=40&t=4068&p=44952.

PR #1100, opened 2016-10-11
Recognize special faction ID "FFFF" during topic info verification

MiroslavR:
Faction "FFFF" means that the speaker must not belong to a faction.

Fixes: https://bugs.openmw.org/issues/3564


Allofich:
Sounds like this would also fix the bug I reported.
https://bugs.openmw.org/issues/3380

MiroslavR:
> Sounds like this would also fix the bug I reported.
> https://bugs.openmw.org/issues/3380

Indeed, I wasn't aware that this bug existed.

psi29a:
Two birds, one commit! \o/

PR #1102, opened 2016-10-13
Change several spell effects from instant to non-instant

Allofich:
Fixes https://bugs.openmw.org/issues/2054.

Several spells applied instantly and then removed in OpenMW actually are applied over a duration of 1 in original Morrowind.

This is true for Cure Common Disease, Cure Blight Disease, Cure Corprus Disease, Cure Poison, Cure Paralyzation and Dispel. Remove Curse was also being treated as instant but it can receive a variable duration.

You can confirm this in original Morrowind by casting a spell or drinking a potion with Cure Common Disease. Select yourself in the console and type "addspell chills" to give yourself the disease immediately afterwards, while the icon still is showing, and the disease will be removed even though it is after the cure effect was applied. The ToggleMagicStats command will also show that the cure effect is repeatedly being applied. I also tested this for Cure Poison and Cure Paralyzation, so I assume it to be true for all of them.

I put these effects alongside an effect explicitly set to duration 1 and they all finished at the same time, so duration 1 seems to be correct. Duration 1 is also what the TES Construction Set shows.

I was going to explicitly give these effects a duration of 1 in OpenMW's code, but they seem to all get 1 anyway just by reading the duration from the spell effects.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -463,7 +463,8 @@ namespace MWMechanics
                 }
                 else // target.getClass().isActor() == true
                 {
-                    if (effectIt->mDuration == 0)
+                    bool hasDuration = !(magicEffect->mData.mFlags & ESM::MagicEffect::NoDuration);
+                    if (hasDuration && effectIt->mDuration == 0)
                     {
                         // duration 0 means apply full magnitude instantly
                         bool wasDead = target.getClass().getCreatureStats(target).isDead();
@@ -479,7 +480,10 @@ namespace MWMechanics
                         ActiveSpells::ActiveEffect effect;
                         effect.mEffectId = effectIt->mEffectID;
                         effect.mArg = MWMechanics::EffectKey(*effectIt).mArg;
-                        effect.mDuration = static_cast<float>(effectIt->mDuration);
+                        if (!hasDuration)
+                            effect.mDuration = 1.0f;
+                        else
+                            effect.mDuration = static_cast<float>(effectIt->mDuration);
                         effect.mMagnitude = magnitude;
 
                         targetEffects.add(MWMechanics::EffectKey(*effectIt), MWMechanics::EffectParam(effect.mMagnitude));
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -463,8 +463,7 @@ namespace MWMechanics
                 }
                 else // target.getClass().isActor() == true
                 {
-                    bool hasDuration = !(magicEffect->mData.mFlags & ESM::MagicEffect::NoDuration);
-                    if (hasDuration && effectIt->mDuration == 0)
+                    if (effectIt->mDuration == 0)
                     {
                         // duration 0 means apply full magnitude instantly
                         bool wasDead = target.getClass().getCreatureStats(target).isDead();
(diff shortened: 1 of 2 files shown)


scrawl:
Might be worth looking into durations of 0. Not sure if they exist in the vanilla game, but could always exist in mods. With this new code, a divine intervention spell with duration 0 would have no effect, but e.g. cure disease for 0 seconds would have an effect. Is that correct?

Allofich:
Well, it's not possible to give a cure disease or divine intervention effect a duration of 0 using the TES Construction Set. It's always a grayed out 1. It's not possible with the in-game spellmaker, either.

I just tried making the code treat all spells as 0 duration, and yes, this makes cure disease apply and end instantly and divine intervention do nothing. Is this a case we need to worry about?

scrawl:
In that case, why not cap the duration to 1 in the code? There is still the possibility of someone hex-editing the file or using unsupported editors (e.g. MWEdit), so that the duration could be 0.

Allofich:
I updated the code a few hours ago but didn't write anything here. I set any spell effects with the "no duration" flag to use a duration of 1. Did you perhaps want 1 to only be set as a minimum? If we want to support things like long-lasting cure disease spells, though, maybe we should do that by allowing editing of the "no duration" flag of spell effects in OpenCS.

Svetomech:
> long-lasting cure disease spells

Equals immunity to the said disease for some time. That would actually be a neat addition to the game, Especially if modders take use of it.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -1149,6 +1123,27 @@ namespace MWMechanics
...
+            actor.getClass().getCreatureStats(actor).getSpells().purgeCommonDisease();
+            break;
+        case ESM::MagicEffect::CureBlightDisease:
+            actor.getClass().getCreatureStats(actor).getSpells().purgeBlightDisease();
+            break;
+        case ESM::MagicEffect::CureCorprusDisease:
+            actor.getClass().getCreatureStats(actor).getSpells().purgeCorprusDisease();
+            break;
+        case ESM::MagicEffect::Dispel:
+            actor.getClass().getCreatureStats(actor).getActiveSpells().purgeAll(magnitude);
```
Allofich: I just tested in vanilla, and it does apply over a duration there. You can see it with togglemagicstats (tms) and verify by casting it and then having a dispel-affected effect be applied to you within the duration to you. For example, have a non-actor cast "ash feast" (drain health) on you. It will be dispelled even though it is after the cast. So we should implement it that way I would think. If you want to fix it feel free, or I'll give it a shot when I have some time.

MiroslavR: Okay, that might be the case, but it appears to apply `magnitude` instead of `magnitude*dt` despite what TMS says. I haven't confirmed that via reverse-engineering, but testing shows that 100% Dispel always works (immediately) in vanilla MW, whereas in OpenMW not so much.

Allofich: It might be interesting to see how vanilla handles a non-100% version. I've only ever used the 100% one in testing. I'll check that out later. I'm kind of busy at work so it may be a few days before I work on a fix for this but when I start I will assign the issue to myself on the bug tracker. In case you want to do it then please do likewise and we can avoid overlapping work. Speaking of which, I have what I think should be a fix in the works for the "aquatic creatures don't stop combat with the player" bug, so I'll go mark myself as working on that.

Allofich: Or well it's if it's just a one-line fix or something I'll do it sooner. I'll take a quick look at it tomorrow.

Allofich: Well, dispel seems to maybe be bugged in vanilla. I tested with some different magnitudes, and 50% seems to always succeed in removing all effects, while 40% and lower seem to never succeed. 45% seems sometimes succeed and sometimes not.

The magic effect description in the TES Construction Set is
"This effect removes magicka-based spell effects from the subject. It does not affect abilities, disease, curses, or constant magic item effects. The magnitude is the chance an effect is removed."
which makes it sound like maybe the % chance is supposed to be rolled separately for each effect, and with 45% I have been able to get a situation where some effects were removed while others weren't.

Allofich: I'm having a 30% dispel now work sometimes and sometimes not after limiting my framerate to 30 in morrowind.ini (it was limited to 60 in the earlier tests.) I've never played Morrowind on Xbox but based on footage on Youtube it was a lower framerate than I was used to seeing on my PC version, so maybe this is more the kind of framerate Morrowind was designed around.

PR #1107, opened 2016-10-16
Fix autoEquip to better match vanilla

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3590

This is the reversed code for autoEquip:
```python
# Order is important. Clothing must precede armor.
for item in npc.items:
    unarmoredRating = (fUnarmoredBase1 * npc.unarmoredSkill) * (fUnarmoredBase2 * npc.unarmoredSkill)

    if item is armor:
        armorRating = item.getArmorRating(npc)
        found = npc.findEquippedArmorWithType(item.armor.type) # ESM::Armor::Type

        if npc.race.isBeast and cantWear:
            continue

        if found:
            cond = armor_rating > found.getArmorRating()
        else:
            cond = armor_rating > unknownFloat # This appears to always be 0.

        if cond:
            if armorRating > unarmoredRating:
                pass
            else:
                continue
        else:
            continue
    elif item is clothing:
        found = npc.findEquippedClothingWithType(item.clot.type) # ESM::Clothing::Type

        if npc.race.isBeast and cantWear:
            continue

        if found and found.value > item.value:
            continue
    else:
        continue

    npc.equipItem(item)
```

I am by no means experienced in reversing, so I could have missed something. Also, there could be some minor differences to my OpenMW implementation (such as vanilla searching by types instead of inventory slots). From the testing I've done, though, it seems to work fine.

### DIFF
--- a/apps/openmw/mwworld/inventorystore.cpp
+++ b/apps/openmw/mwworld/inventorystore.cpp
@@ -216,21 +216,32 @@ MWWorld::ContainerStoreIterator MWWorld::InventoryStore::getSlot (int slot)
 
 void MWWorld::InventoryStore::autoEquip (const MWWorld::Ptr& actor)
 {
+    if (!actor.getClass().isNpc())
+        // autoEquip is no-op for creatures
+        return;
+
+    const MWBase::World *world = MWBase::Environment::get().getWorld();
+    const MWWorld::Store<ESM::GameSetting> &store = world->getStore().get<ESM::GameSetting>();
+    MWMechanics::NpcStats& stats = actor.getClass().getNpcStats(actor);
+
+    static float fUnarmoredBase1 = store.find("fUnarmoredBase1")->getFloat();
+    static float fUnarmoredBase2 = store.find("fUnarmoredBase2")->getFloat();
+    int unarmoredSkill = stats.getSkill(ESM::Skill::Unarmored).getModified();
+
+    float unarmoredRating = static_cast<int>((fUnarmoredBase1 * unarmoredSkill) * (fUnarmoredBase2 * unarmoredSkill));
+
     TSlots slots_;
     initSlots (slots_);
 
     // Disable model update during auto-equip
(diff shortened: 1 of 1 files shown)


Allofich:
> Equipping weapons is handled by AiCombat.

You may not want to concern yourself with it for this PR, but I think that to address https://bugs.openmw.org/issues/3562 (which I can confirm as valid report) it may be necessary for NPCs to equip their weapons even when not in combat.

Also, if you don't mind my asking, how do you reverse-engineer the code like that? You also reverse-engineered the code for the level-up image, which impressed me.

MiroslavR:
> You may not want to concern yourself with it for this PR, but I think that to address https://bugs.openmw.org/issues/3562 (which I can confirm as valid report) it may be necessary for NPCs to equip their weapons even when not in combat.

Thanks, I will look into it.

> Also, if you don't mind my asking, how do you reverse-engineer the code like that?

Using debugging tools like [OllyDbg](http://ollydbg.de/) and [x64dbg](http://x64dbg.com/). [IDA](https://www.hex-rays.com/products/ida/) is probably the most advanced tool for the job, which I believe Hrnchamd used for our research, but it's not free.

You will need to have general knowledge of the x86 architecture, the x86 assembly language (and probably also how compilers tend to translate C++ code - things like virtual tables and the ECX register being used to hold _this_), and a lot of dedication. I usually just search for references to strings and certain values, place some code/memory breakpoints in hopes to find the code responsible for e.g. auto-equipping. Then it's just a stepping fest, instruction by instruction, to make sense of it all (the mentioned tools also feature some form of code analysis and allow you to label addresses and write comments). The source code of MWSE as well as MGE-XE has been a helpful resource to me for deciphering some of the internal data structures.

Allofich:
That was what I thought the answer might be, debugging assembly. I would have to study a lot before I could hope to do start doing that. Well, thanks for answering.

PR #1108, opened 2016-10-18
Auto-equip weapons

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3562

This also removes some owner checks preventing merchants from equipping armor or clothing that is sold to them. I couldn't find any such checks in vanilla MW, and testing proved they can indeed equip items sold to them.

Reversed code for auto-equipping weapons:
```python
weaponSkills = [5, 6, 7, 22, 23, 4]
skillVisited = [False, False, False, False, False, False]

for _ in range(len(weaponSkills)):
    max = 0
    maxWeaponSkill = None

    for i in range(len(weaponSkills)):
        skill_val = npc.skills[weaponSkills[i]]

        if skill_val > max and skillVisited[i] == False:
            max = skill_val
            maxWeaponSkill = i

    if maxWeaponSkill is None:
        break

    max = 0
    stack = None
    for itemStack in inventory:
        item = itemStack.item
        if item is not weapon: # Unlike in OpenMW, ammunition (arrows and bolts) is not classified as weapon.
            continue

        if weaponTypeToSkill(item.weapon.type) == weaponSkills[maxWeaponSkill]:
            if item.weapon.chopMax >= max:
                max = item.weapon.chopMax
                stack = itemStack

            if item.weapon.slashMax >= max:
                max = item.weapon.slashMax
                stack = itemStack

            if item.weapon.thrustMax >= max:
                max = item.weapon.thrustMax
                stack = itemStack

    if stack is not None:
        if stack.something08 != 0: # No idea what this is.
            npc.equipItem(stack.item, stack.something08.something00, 0, 0)
        else:
            npc.equipItem(stack.item, 0, 0, 0)
        break

    skillVisited[maxWeaponSkill] = True
```

### DIFF
--- a/apps/openmw/mwworld/inventorystore.cpp
+++ b/apps/openmw/mwworld/inventorystore.cpp
@@ -6,6 +6,7 @@
 #include <components/esm/loadench.hpp>
 #include <components/esm/inventorystate.hpp>
 #include <components/misc/rng.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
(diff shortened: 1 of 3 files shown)


scrawl:
The owner check is to replicate an MCP feature. I know of at least one person who would complain if this feature was removed. But you are right that the default behaviour should be vanilla. We can add a new setting to settings.cfg [Game] section for the time being I guess.

Allofich:
About the owner check. I haven't actually tested it but I believe it would cause the end of the Imperial Legion quest line to not work as intended. At the end of those quests you are asked by the quest giver to go get some powerful equipment, which the quest giver auto-equips when you bring it back to him. In a quest after that you fight him. In the official strategy guide it mentions that the quest giver will use the equipment you brought against you in the fight, so I think this can be considered the intended, or at least expected, behavior by Bethesda.

If we have the check, as an option for example, maybe there should be a check of whether the NPC is a merchant. That's not a perfect solution but that's all I can think of.

MiroslavR:
> The owner check is to replicate an MCP feature. I know of at least one person who would complain if this feature was removed. But you are right that the default behaviour should be vanilla. We can add a new setting to settings.cfg [Game] section for the time being I guess.

Ah, ok. I thought those checks were based on an incorrect assumption. Anyway, done.

PR #1110, opened 2016-10-20
Save controls state

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3598

### DIFF
--- a/apps/essimporter/converter.hpp
+++ b/apps/essimporter/converter.hpp
@@ -278,7 +278,7 @@ public:
         PCDT pcdt;
         pcdt.load(esm);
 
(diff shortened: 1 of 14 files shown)


Allofich:
Are you aware of this issue?
https://bugs.openmw.org/issues/3420

Looks like you are partly implementing that feature.

MiroslavR:
I wasn't aware. Let's fix/implement it fully then.

Also, the flags mentioned in the report are wrong.
> 0x0100 not sure, I have two saves with this set after I played around with ForceJump and ForceRun, neither of which actually affect the player)

This is most likely `PlayerFlags_SpellDrawn`.

PR #1112, opened 2016-10-25
Move actors out of the water if there's room for them. Fixes #1138

Assumeru:
Based on the diff by Stanislav Zhukov, with a check thrown in to make sure there's room.

I did manage to get stuck when I cast it under a sloped roof (Hlormaren, Sewers), but [806](https://bugs.openmw.org/issues/806) would fix that.

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -1309,11 +1309,15 @@ namespace MWPhysics
         PtrVelocityList::iterator iter = mMovementQueue.begin();
         for(;iter != mMovementQueue.end();++iter)
         {
+            ActorMap::iterator foundActor = mActors.find(iter->first);
+            if (foundActor == mActors.end()) // actor was already removed from the scene
+                continue;
+            Actor* physicActor = foundActor->second;
+
             float waterlevel = -std::numeric_limits<float>::max();
             const MWWorld::CellStore *cell = iter->first.getCell();
             if(cell->getCell()->hasWater())
                 waterlevel = cell->getWaterLevel();
-
 
             const MWMechanics::MagicEffects& effects = iter->first.getClass().getCreatureStats(iter->first).getMagicEffects();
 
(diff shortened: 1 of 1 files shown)


apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -1318,11 +1318,34 @@ namespace MWPhysics
...
-                    && !world->isUnderwater(iter->first.getCell(),
-                                           osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
-                waterCollision = true;
+            if (cell->getCell()->hasWater() && effects.get(ESM::MagicEffect::WaterWalking).getMagnitude())
+            {
+                if (!world->isUnderwater(iter->first.getCell(), osg::Vec3f(iter->first.getRefData().getPosition().asVec3())))
+                    waterCollision = true;
+                else
+                {
+                    Actor *actor = getActor(iter->first);
```
Assumeru: Righty, moved it all the way up.

Assumeru:
On a somewhat related note, shouldn't waterlevel be initialized as std::numeric_limits<float>::min() instead of negative max?

kcat:
> On a somewhat related note, shouldn't waterlevel be initialized as std::numeric_limits::min() instead of negative max?

No, because unlike integer types, ```std::numeric_limits<float>::min()``` (and other floating point types) is the minimum positive non-0 value.

Assumeru:
Huh, weird.
Learnt something new then, thanks.

Assumeru:
Should I be doing anything else to get this merged?

scrawl:
Yes. I wrote a comment about line 1334, but apparently it was never published. The new "review" feature on github is confusing.

Anyway: I believe there's a mistake with the positions you're passing to doTrace. Because of how Bullet works, you need to account for the offset between the actor's origin (i.e. actor->getPosition()) and the center of the collision object. Adding actor->getHalfExtents().z to the positions should correct this.

Assumeru:
Added it to destinationPosition.
Can there ever be anything other than the actor between its origin and the center of its collision object?

scrawl:
You'll also need to add halfExtents.z to actorPosition. As it stands, you're tracing from slightly below the actor towards the water surface, this could lead to false collisions if there's some coliision object just below the actor.

PR #1113, opened 2016-10-30
Prevent NPC from chosing farther pathgrid node.

kurnevsky:
It prevents guard from changing the direction as I mentioned here https://bugs.openmw.org/issues/3587
This happens because the guard uses shortcut and then bumps into an obstacle and returns to the nearest pathgrid node.
Though I'm not sure that this patch is the best way to fix this behaviour (maybe guard should always use pathgrid instead of shortcut?). But anyway it seems that this patch might be useful in general.

### DIFF
--- a/apps/openmw/mwmechanics/pathfinding.cpp
+++ b/apps/openmw/mwmechanics/pathfinding.cpp
@@ -231,14 +231,22 @@ namespace MWMechanics
         {
             mPath = pathgridGraph.aStarSearch(startNode, endNode.first);
 
-            // If startNode (the closest node to startPoint) is farther from destination than
-            // startPoint and second path node is closer to destination then
-            // drop startNode and use the next path node.
             if (mPath.size() > 1) {
-                float endTo1stNodeLength2 = DistanceSquared(mPathgrid->mPoints[startNode], endPointInLocalCoords);
-                float endTo2ndNodeLength2 = DistanceSquared(*(++mPath.begin()), endPointInLocalCoords);
-                if (endTo1stNodeLength2 > startToEndLength2 && endTo2ndNodeLength2 < endTo1stNodeLength2)
-                    mPath.pop_front();
+                ESM::Pathgrid::Point secondNode = *(++mPath.begin());
+                osg::Vec3f firstNodeVec3f = MakeOsgVec3(mPathgrid->mPoints[startNode]);
+                osg::Vec3f secondNodeVec3f = MakeOsgVec3(secondNode);
+                osg::Vec3f toSecondNodeVec3f = secondNodeVec3f - firstNodeVec3f;
+                osg::Vec3f toStartPointVec3f = startPointInLocalCoords - firstNodeVec3f;
+                float cos = (toSecondNodeVec3f * toStartPointVec3f) / (toSecondNodeVec3f.length() * toStartPointVec3f.length());
+                if (cos > 0) {
+                    ESM::Pathgrid::Point temp(secondNode);
+                    converter.toWorld(temp);
+                    bool isPathClear = !MWBase::Environment::get().getWorld()->castRay(
+                        static_cast<float>(startPoint.mX), static_cast<float>(startPoint.mY), static_cast<float>(startPoint.mZ),
+                        static_cast<float>(temp.mX), static_cast<float>(temp.mY), static_cast<float>(temp.mZ));
+                    if (isPathClear)
+                        mPath.pop_front();
+                }
             }
 
             // convert supplied path to world coordinates
--- a/apps/openmw/mwmechanics/pathfinding.cpp
+++ b/apps/openmw/mwmechanics/pathfinding.cpp
@@ -231,6 +231,16 @@ namespace MWMechanics
         {
             mPath = pathgridGraph.aStarSearch(startNode, endNode.first);
 
+            // If startNode (the closest node to startPoint) is farther from destination than
+            // startPoint and second path node is closer to destination then
+            // drop startNode and use the next path node.
+            if (mPath.size() > 1) {
+                float endTo1stNodeLength2 = DistanceSquared(mPathgrid->mPoints[startNode], endPointInLocalCoords);
+                float endTo2ndNodeLength2 = DistanceSquared(*(++mPath.begin()), endPointInLocalCoords);
+                if (endTo1stNodeLength2 > startToEndLength2 && endTo2ndNodeLength2 < endTo1stNodeLength2)
+                    mPath.pop_front();
+            }
+
             // convert supplied path to world coordinates
             for (std::list<ESM::Pathgrid::Point>::iterator iter(mPath.begin()); iter != mPath.end(); ++iter)
             {
--- a/apps/openmw/mwmechanics/pathfinding.cpp
+++ b/apps/openmw/mwmechanics/pathfinding.cpp
@@ -237,8 +237,7 @@ namespace MWMechanics
                 osg::Vec3f secondNodeVec3f = MakeOsgVec3(secondNode);
                 osg::Vec3f toSecondNodeVec3f = secondNodeVec3f - firstNodeVec3f;
                 osg::Vec3f toStartPointVec3f = startPointInLocalCoords - firstNodeVec3f;
-                float cos = (toSecondNodeVec3f * toStartPointVec3f) / (toSecondNodeVec3f.length() * toStartPointVec3f.length());
-                if (cos > 0) {
+                if (toSecondNodeVec3f * toStartPointVec3f > 0) {
                     ESM::Pathgrid::Point temp(secondNode);
                     converter.toWorld(temp);
                     bool isPathClear = !MWBase::Environment::get().getWorld()->castRay(


Allofich:
> It prevents guard from changing the direction

I don't see this happen, using the latest code. You mean the guard who is in the ship, right?

It might not be a good idea to cut a pathgrid point just based on the distance. There could be a complex area with a pathgrid where an NPC needs to first go to a pathgrid point that is farther from the target in order to get around an obstacle or to move along a bridge, etc. In this situation I think the pathfinding AI is probably not good enough to find the way to the second node if the first one was omitted.

kurnevsky:
> I don't see this happen, using the latest code. You mean the guard who is in the ship, right?
Yes. Here is the video of this process:
[out.zip](https://github.com/OpenMW/openmw/files/560620/out.zip)

Allofich:
Thank you for the video. That does not happen for me, but I did see behavior just like that when work was being done some time ago on the AI pathfinding. I'm testing using the latest code from the OpenMW master, are you also?

kurnevsky:
Yes, I it's the latest master.

Allofich:
Strange. Does it happen always for you? I've tested three or more times now but it hasn't happened.

kurnevsky:
Almost always. Differs only how many times the guard returns to pathgrid node, and where exactly it happens :)

Allofich:
No mods, right? Well I don't know why we have different behavior. Maybe @scrawl or @mrcheko, who worked on the pathfinding code recently, can give an answer.

kurnevsky:
I just tested without mods - it doesn't happen... Give me a moment - I'll figure out what mod gives such behaviour.

kurnevsky:
It was http://www.nexusmods.com/morrowind/mods/44001 - even without esp activation.

kcat:
Maybe it's frame rate dependent? If the time it takes to handle a frame isn't constant, perhaps the NPC has a chance to overshoot the intended node and get confused about which to go to.

Allofich:
Maybe so. That video file has choppiness to it. Is it an accurate recreation of your frame rate?

kurnevsky:
It runs with 60 fps all the time in both cases with/without this mod. (But I recorded the video with intel video card - might be less fps in it)

Allofich:
I can confirm similar behavior when using the meshes from that mod (I didn't bother installing the textures as well.) The fps was smooth but it happened anyway. My guess is that a difference in the models allows a ray trace or something used for deciding to attempt a shortcut to succeed with the mod, while it doesn't with the vanilla models.

scrawl:
Perhaps the shortcutting test should use a sphere cast instead of a raycast, to account for the fact that NPC's are not infinitely thin.

scrawl:
Or we could add a "risk/reward" type of enhancement to the shortcutting logic, where if the length of the pathgrid path is not much longer than the shortcut, then don't risk taking that shortcut.

Allofich:
If the sphere can be made to be about equal to the actor's collision, that seems like a good idea. I think even if there isn't much distance to be saved by taking a shortcut, it could still be nice to have if it makes more natural-looking movement than the kind of jerky movement from walking exactly from point to point on the pathgrid, and it might also help stop NPCs from walking into each other at pathgrid points.

Allofich:
On the other hand, I can imagine it might not be good to always have a shortcut. For example, NPCs walking right next to lava or something else you would normally want to stay away from, just because their sphere cast made it through.

kurnevsky:
ok, then how about removing first node only if angle between vectors 1st_node-start_point and 1st_node-2nd_node is acute (about 5-10°)? It should prevent only such odd returns.

kurnevsky:
Even better we can calculate the distance between npc and line between first and second node. And if it's not big - drop first node. Unlikely it can break something, but can give more adequate behaviour in such rare cases.

Allofich:
I just did some testing of the ingame AI, and it could definitely use an improvement. Get an NPC attacking you in town, and when they are "shortcutting" towards you because of having direct line of sight, if you move behind an obstacle like a house, the NPC will run to the nearest pathgrid point, often backwards or sideways or in a small circle.

If we only dropped the node for very acute angles, I don't think it would help for this case. What about dropping the first node and attempting to shortcut to the second node? Maybe the shortcut code can be reused like that.

kurnevsky:
> If we only dropped the first node for very acute angles, I don't think it would help for this case.
Yes, I tested it - it works in rare cases only. I also noticed that getWorld->castRay often returns true (path is not clear) for (startPoint, firstPathgridNode), I don't know why. Maybe I do something wrong :)

mrcheko:
To the initial problem: guard goes backward there because of shortcut ray fails. The direct reason seems to be those planks on the floor. Can't look why it's ok with vanilla models right now.

Anyway I propose not to cut any nodes from the built path just because it will look smoother in some rare cases (like this one). My option would be to change rules for applying shortcuts - check even more conditions before apply. But the solution should be as generic as possible to not introduce problems in other places. You know pathfinding with these pathgrids is so fragile:)

I personally liked scrawl's idea:
>if the length of the pathgrid path is not much longer than the shortcut, then don't risk taking that shortcut.

Or you can treat it like: if a shortcut goes near a pathgrid then better stick to that pathgrid.

Or maybe skip shortcuts entirely for scripted sequences: in cases like NPC should travel somewhere due to a script.

Allofich:
Even if the conditions before attempting a shortcut are improved, I think we still should improve the behavior when a shortcut fails, as currently there is the behavior I described in my above post when a pursuing NPC loses line of sight with the target it is shortcutting to. I don't know if cutting the first node will satisfactorily solve that but it's one idea.

kurnevsky:
Updated it slightly - now use angle and castRay. But current version doesn't fix the behaviour of the guard with Better Ships mod because castRay returns true for the second node of pathgrid.

Allofich:
I tested your code, and it seems to be working better than the current AI for pursuing. They don't run backwards to a pathgrid point after losing LOS. I can confirm the guard walking in the intro with Better Ships still goes back.

Also about changing shortcut code and the suggestions by scrawl, though I said I liked the sphere cast idea I don't object to either idea. It's hard to say without actually seeing the ideas in practice which one is better.

Anyway, I think we probably need to clarify what the goal of this PR is. If it is to fix the guard with Better Ships, well, it is not doing that so I don't think it will be merged unless more changes are made. kurnevsky do you want to try to do the suggestions to improve the shortcut decision-making? If the PR goal is limited to improving AI pursuing, if no problems are added by this PR I think it would be OK to merge for that.

> Also I'm not sure if it make sence to use the full version of shortcutting here because NPC may start rushing between first and second nodes (if he will encounter obstacles).

I'm sorry I don't understand not sure what you mean.

kurnevsky:
I think that this change is useful even if it doesn't fix the guard behavior. It's possible to fix it in such way only without additional checks with castRay.
>> kurnevsky do you want to try to do the suggestions to improve the shortcut decision-making?
If I'll do it - i"ll open another pull request (maybe we should open low priority bug for it?).
One thing I'm not sure of - shall I move shortcutPath (or part of it) to pathfinding.cpp and use it instead of castRay? I meant that the result of checkWayIsClearForActor may be different depending on the locatiobn of NPC. So theoretically NPC can chose second node when checkWayIsClearForActor=true then first node when checkWayIsClearForActor=false then second node again and so on.

Allofich:
During testing I just saw an NPC running around in circles during combat, probably because he was getting LOS with his target then losing it and running back to a pathgrid point, then getting it again, or something that like. I do think we should fix this, and when I tested this earlier I found the problem was improved. Why don't we make that the focus of this PR and move towards a merge?

jeffreyhaines:
This open pull request sounds pretty familiar to the bug I recently reported. https://bugs.openmw.org/issues/3660

Where a summoned scamp can't path to me and just runs around in circles.

Allofich:
I just looked at your video, and yes it does look like the same thing is happening where the scamp keeps running back to the last pathgrid point after losing line of sight with you. Although, it should be trying to go up the ramp like the NPC did instead of running straight at you, but that's another issue.

jeffreyhaines:
Yeah it is a little difficult to see whats going on. I'll try to film a better video for you. Adding it to my list..

Allofich:
By the way I'm testing the same room in vanilla right now and the scamp is having trouble there, too. Might be a tricky bit of terrain for some reason.

Allofich:
Regardless, there's no doubt that the "lose line-of-sight with target, run back to last pathgrid point" behavior exists, whether it was the reason in this particular case or not.

Allofich:
Did we have an update here?
kurnevsky do you know of any problems? I just tested and the issue of NPCs retreating to their last pathgrid node whenever they lose LOS with their target is still in master.

kurnevsky:
@Allofich I only rebased my changes. I didn't see problems with this changes when played myself. But there might be some objections. Firstly it doesn't fix the original issue with the guard in better ships mod. Secondly it seems that vanilla morrowind doesn't have such check if NPS should move towards the second pathgrid node. But at least this fix doesn't break anything.
Other solutions were offered here but I'm not sure if I'm able to implement them :)

akortunov:
So which bug this PR should fix in its current state?

PR #1114, opened 2016-10-30
Fix shadowing warnings

Allofich:
Fixes most of the remaining MSVC shadowing warnings. The only ones left I saw being mentioned by AppVeyor were in Boost, and in instances of QT "foreach" statements, which I'm not familiar with so I left as they were for now.

In apps/openmw/mwscript/locals.cpp "index2" (renamed from "index") is a char taking a value from a function returning an int. Is this correct?

I also didn't fix the shadowing of item in updateButton() in the wizard's componentselectionpage.cpp. The function parameter is also "item" and isn't being used. It looks definitely wrong but I'm not familiar with that code area so I didn't touch it for now.

### DIFF
--- a/apps/launcher/maindialog.cpp
+++ b/apps/launcher/maindialog.cpp
@@ -326,7 +326,7 @@ bool Launcher::MainDialog::setupGameSettings()
     foreach (const QString &path, paths) {
         qDebug() << "Loading config file:" << path.toUtf8().constData();
 
-        QFile file(path);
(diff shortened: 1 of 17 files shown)


Allofich:
Scrawl do you have any comment about the questions I had about the char and int, and the unused function parameter?

scrawl:
I have now checked in the two changes to locals.cpp and componentselectionpage.cpp.

PR #1121, opened 2016-11-05
Allow creature opening only when death animation is finished.

kurnevsky:
It prevents disposing of corpse until death counter is not incremented.

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -422,7 +422,10 @@ namespace MWClass
             return action;
         }
 
-        if(getCreatureStats(ptr).isDead())
+        const MWMechanics::CreatureStats& stats = getCreatureStats(ptr);
+        // Allow creature opening only when death animation is finished because
+        // we increment death counter only after it.
+        if(stats.isDead() && stats.isDeathAnimationFinished())
             return boost::shared_ptr<MWWorld::Action>(new MWWorld::ActionOpen(ptr, true));
         if(ptr.getClass().getCreatureStats(ptr).getAiSequence().isInCombat())
             return boost::shared_ptr<MWWorld::Action>(new MWWorld::FailedAction(""));
--- a/apps/openmw/mwclass/npc.cpp
+++ b/apps/openmw/mwclass/npc.cpp
@@ -839,7 +839,10 @@ namespace MWClass
             return action;
         }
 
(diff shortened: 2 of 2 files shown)


apps/openmw/mwclass/creature.cpp
```diff
@@ -422,7 +422,10 @@ namespace MWClass
             return action;
         }
 
-        if(getCreatureStats(ptr).isDead())
+        MWMechanics::CreatureStats& stats = getCreatureStats(ptr);
```
scrawl: <pre>const MWMechanics::CreatureStats& stats</pre>

scrawl:
What about NPCs?

How does this relate to the feature that you are able to loot summoned creatures before they disappear? I think we want to keep that.

kurnevsky:
> How does this relate to the feature that you are able to loot summoned creatures before they disappear? I think we want to keep that.

It is very similar to a bug. Is there something that uses that? Vanilla engine lets loot summoned creatures for a few moments after the animation of death and before creature disappearing. So one possible solution - add some delay after this animation.

Allofich:
> So one possible solution - add some delay after this animation.

I think that we are trying to keep the game as close to the original as possible when we can, and not add in new things.

In https://bugs.openmw.org/issues/3528 there is discussion about this issue. If a character is in combat with anyone when they die, you cannot loot their body until it finishes the death animation. If they weren't in combat, you can loot the body during the death animation. As far as I know, this will only happen with a follower/summoned creature who needs a few hits before they go hostile. If you kill them within that number of hits, you can loot the body during the animation. It could also happen if you use a console command like "sethealth 0," or I suppose if a script did so.

Even if you kill a non-follower/non-summoned creatures in one hit, they still seem to go into combat and cannot be looted.

I don't know if that was an intentional design decision or not, but it seems to me that we should do it the same way. Not allowing looting the bodies during the animation could be added as an optional setting.

This should fix https://bugs.openmw.org/issues/3528 and https://bugs.openmw.org/issues/3513 (I don't know about the Umbra thing mentioned in the comments). The only remaining potential problem would be for counting deaths of followers/summoned creatures, or if scripts were used to kill characters. That should be a problem in original MW, too. To protect against these edge cases, we could have the "Dispose of Corpse" button be ineffective until the death animation ends.

scrawl:
Superseded by #1350

PR #1123, opened 2016-11-06
Fix dereference of null pointer warning

Allofich:
This fixes the Clang warning about a dereference of a null pointer.

Is this warning just a false positive? Anyway, putting in an assert statement seems to have quieted it.

### DIFF
--- a/apps/essimporter/importer.cpp
+++ b/apps/essimporter/importer.cpp
@@ -51,6 +51,7 @@ namespace
         {
             for (int x=0; x<128; ++x)
             {
+                assert(image->data(x,y));
                 *(image->data(x,y)+2) = *it++;
                 *(image->data(x,y)+1) = *it++;
                 *image->data(x,y) = *it++;


apps/essimporter/importer.cpp
```diff
@@ -51,6 +51,7 @@ namespace
         {
             for (int x=0; x<128; ++x)
             {
+                assert(image->data(x,y)+2);
```
EmperorArthur: Good idea, but you can probably drop the `+2` from that assert.

[Here's the documentation for osg::Image](http://trac.openscenegraph.org/documentation/OpenSceneGraphReferenceDocs/a00382.html#a9a1861cdedf0d8edc2d1e96f55cae837).

Reading the code it's copying the R,G,B values from the file into the image.  I'd recommend using [setColor](http://trac.openscenegraph.org/documentation/OpenSceneGraphReferenceDocs/a00382.html#ab08a188ade29125510cb3d1cedd8c0a0) instead.

Something like this might work instead:

```
image->setColor({*it++,*it++,*it++,0},{x,y});
```

Allofich: I spent some time trying that out, but I had no luck. That line as it is won't work, setColor wants a vector of floats. The data being read from the image in the .ess file is unsigned chars. I tried giving setColor this data converted to float (and rearranged into BGR order) but didn't get a correct image from it. Maybe I'm just doing something wrong.

The +2 was unnecessary, you are right, so I'll remove that. I'm inclined to just leave it at that, as all I was trying to do with this PR was silence the warning.

PR #1124, opened 2016-11-09
Create convert_bump_mapped_mods.rst

lysol90:
A tutorial to how normal mapping works with different mods in Morrowind and how they work in OpenMW. It also covers how to convert the mods made for Morrowind to be used with OpenMW.

See [this](https://forum.openmw.org/viewtopic.php?f=40&t=3922) thread. This is the same tutorial but reformatted to ReST-format and cleaned up a bit.

My first PR, so I hope I did everything right.

### DIFF
--- a/docs/source/openmw-mods/convert_bump_mapped_mods.rst
+++ b/docs/source/openmw-mods/convert_bump_mapped_mods.rst
@@ -35,13 +35,13 @@ So let's get on with it.
 Normal Mapping in OpenMW
 ------------------------
 
-Normal mapping in OpenMW works in a very simple way: The engine just looks for a texture with a "_n.dds" suffix, and you're done.
+Normal mapping in OpenMW works in a very simple way: The engine just looks for a texture with a *_n.dds* suffix, and you're done.
 
 So to expand on this a bit, let's take a look at how a model seeks for textures.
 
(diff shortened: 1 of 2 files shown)


psi29a:
Nice! Looking good. :) 

Can you also modify openmw/docs/source/openmw-mods/index.rst to include convert_bump_mapped_mods to the list of entries under the ToC (Table of Contents). This way Sphinx will pick up on it when rendering the documentation to html/pdf/latex/what-ev

lysol90:
Oops, I realize my latest edit might have become a pull request. Wasn't my intention. I'm going to make a few more changes first.

psi29a:
Nice work so far, you satisfied with the PR? Is it ready for review/merge?

lysol90:
Yeah, sure. It's fine to me.

psi29a:
LGTM :)

psi29a:
https://openmw.readthedocs.io/en/master/  <-- not bad :)  

I think the 1,2,3,4 shouldn't show up here like this:
https://openmw.readthedocs.io/en/master/openmw-mods/index.html

lysol90:
Nice! The documentation works great I see.

Hmm, no you're right. The tutorial in itself was good with the numbers, but when seen with the rest of the documentation, it doesn't really fit. I can remove it if you want. Also, I'm not sure the subcategories (e.g. the 1, 2, 3, 4) should be in the ToC as if they were their own articles. It's probably because I used the = header for those. I should do a - header instead. If I do this, I guess the problem with the numbers showing up in the ToC disappears too.

Sounds good or do you have another suggestion?

psi29a:
Give it a try.  If you would like to preview how it would look, give sphinx a try locally.

PR #1125, opened 2016-11-09
Further updating of normal map conversion documentation

lysol90:
Ok, so another update then. Sorry for another one so soon after the first PR, but I think (or hope) that this will be the last edit for some time. It looks good to me now.

Mainly, I changed changed some "=" headers to "-", and while at it, I fixed some typing errors and just edited some text to read better.

### DIFF
--- a/docs/source/openmw-mods/convert_bump_mapped_mods.rst
+++ b/docs/source/openmw-mods/convert_bump_mapped_mods.rst
@@ -6,7 +6,7 @@ Index:
 
 - `General introduction to normal map conversion`_
     - `Normal Mapping in OpenMW`_
-        - `Activating normal mapping shaders in OpenMW`_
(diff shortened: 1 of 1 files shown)


psi29a:
^5  looks great
https://openmw.readthedocs.io/en/master/

Feel free to spread the news on the forums :)

psi29a:
Spoke too soon: https://openmw.readthedocs.io/en/master/openmw-mods/index.html

'Index:' shouldn't be there, probably because Index: is an anchor? You can click on it, I think you can drop the Index: honestly, does it add anything?

psi29a:
Fixed it. :)  Great job and thank you!

lysol90:
> Fixed it. :)

Haha, that explains it. I just though "what index?" when I clicked the link. Thanks for fixing it for me. No, the index does not need to be there.

PR #1129, opened 2016-11-13
Fix various build issues with Visual Studio environment

PhillipAnd:
This fixes several issues with building in Visual Studio, including:
- Debug builds don't install because the install script is still looking for files in the Release directory
- Several library dependencies related to Qt are missing (imm32.lib, winmm.lib, version.lib)
- Also added the /bigobj flag for openmw-cs


CMakeLists.txt
```diff
@@ -429,27 +431,34 @@ if(WIN32)
         "${OpenMW_SOURCE_DIR}/Docs/license/DejaVu Font License.txt"
         "${OpenMW_BINARY_DIR}/settings-default.cfg"
         "${OpenMW_BINARY_DIR}/gamecontrollerdb.txt"
-        "${OpenMW_BINARY_DIR}/Release/openmw.exe"
         DESTINATION ".")
+	INSTALL(FILES "${OpenMW_BINARY_DIR}/Debug/openmw.exe" DESTINATION "." CONFIGURATIONS Debug)
+	INSTALL(FILES "${OpenMW_BINARY_DIR}/Release/openmw.exe" DESTINATION "." CONFIGURATIONS Release;RelWithDebInfo;MinSizeRel)
```
scrawl: I don't like this duplication of code. At the very least you should set a variable to "Debug" or "Release" depending on current configuration, then use that variable in the install commands. But perhaps there is a better way to check where the executable resides? Have you tried using RUNTIME_OUTPUT_DIRECTORY of the target's properties, or maybe CMAKE_CURRENT_BINARY_DIR ?

apps/bsatool/CMakeLists.txt
```diff
@@ -18,3 +18,7 @@ if (BUILD_WITH_CODE_COVERAGE)
   add_definitions (--coverage)
   target_link_libraries(bsatool gcov)
 endif()
+
+if(MSVC)
+  target_link_libraries(bsatool imm32.lib winmm.lib version.lib)
```
scrawl: Is there a reason you are adding these, according to you, QT-related libraries to the non QT projects? We only use QT for the launcher, wizard and opencs.

Might be worth looking into why QT isn't adding those libraries on its own. Our other Windows developers don't seem to have a problem with these. What version of Qt are you using? dynamic or static libraries? Ideally the Qt cmake script should be handling all the needed dependencies.

apps/opencs/CMakeLists.txt
```diff
@@ -225,6 +225,15 @@ if (WIN32)
     target_link_libraries(openmw-cs ${Boost_LOCALE_LIBRARY})
 endif()
 
+if (MSVC)
+    # Debug version needs increased number of sections beyond 2^16
+    if (CMAKE_CL_64)
+        set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /bigobj")
+    endif (CMAKE_CL_64)
+    add_definitions("-D_USE_MATH_DEFINES")
```
scrawl: Is the addition of USE_MATH_DEFINES really necessary? Looks to me it was inadvertently copy-pasted.

CMakeLists.txt
```diff
@@ -419,8 +419,10 @@ IF(NOT WIN32 AND NOT APPLE)
 ENDIF(NOT WIN32 AND NOT APPLE)
 
 if(WIN32)
-    FILE(GLOB dll_files "${OpenMW_BINARY_DIR}/Release/*.dll")
-    INSTALL(FILES ${dll_files} DESTINATION ".")
+	FILE(GLOB dll_files_debug "${OpenMW_BINARY_DIR}/Debug/*.dll")
```
scrawl: You've got tabs here and in other places, please configure your editor to spaces only.

apps/bsatool/CMakeLists.txt
scrawl: The targets using SDL would be launcher, openmw and components.

Then it should be handled in the SDL2 cmake script. Actually I see this in FindSDL2.cmake:
<pre>

if(WIN32)
    list(APPEND SDL2_TARGET_SPECIFIC winmm imm32 version msimg32)
endif()

set(SDL2_PROCESS_LIBS SDL2_TARGET_SPECIFIC)

libfind_process(SDL2)
</pre>

Then, somewhere inside the libfind_process macro it's doing something with the SDL2_PROCESS_LIBS.

Clearly, the find script already attempts to handle the libraries you are missing. We should look into why that is not working.

Does reverting this PR help? https://github.com/OpenMW/openmw/pull/1027

CMakeLists.txt
PhillipAnd: `CMAKE_CURRENT_BINARY_DIR` is set to the same as `OpenMW_BINARY_DIR`, so if you really want to avoid the kind of thing that I did, I think it's best to `INSTALL TARGETS` rather than `INSTALL FILES`.  Will get to work on that, let me know if you want to do something different

CMakeLists.txt
PhillipAnd: Oops, new editor!  Fixed in latest commits.

CMakeLists.txt
PhillipAnd: I just pushed a commit that moves the install commands for the binaries to the target-specific CMakeLists.  This completely fixes the problem with the install target only working for release builds.  Now there are only a few configuration-specific install commands left in the main file, mostly dealing with DLL's.  I don't know of any way to eliminate those.

PhillipAnd: One question - is the `Plugin_MyGUI_OpenMW_Resources.dll` still relevant?  It isn't in the official windows build and it breaks my build unless I disable `BUILD_MYGUI_PLUGIN`.

apps/bsatool/CMakeLists.txt
PhillipAnd: Okay, I tracked down the problem.  Changing `${SDL_LIBRARY}` to `${SDL_LIBRARIES}` in the CMakeList for the `components` target fixes everything.  I have done this and removed all of my dirty work-arounds in the latest commit.

scrawl:
@Ace13 could you give this a test?

CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ if(EXISTS ${PROJECT_SOURCE_DIR}/.git)
     else(NOT EXISTS ${PROJECT_SOURCE_DIR}/.git/shallow)
         message(STATUS "Shallow Git clone detected, not attempting to retrieve version info")
     endif(NOT EXISTS ${PROJECT_SOURCE_DIR}/.git/shallow)
-endif(EXISTS ${PROJECT_SOURCE_DIR}/.git)
+endif(EXISTS ${PROJECT_SOURCE_DIR}/.git)
```
psi29a: Why the extra space/padding?

ace13:
Tried packaging both debug and release packages and they seem functional on my side.

Nice to see someone else willing to spend some time on Windows packaging, never even thought of the need to package/install debug builds when I started writing the CPack stuff back in the day.

PR #1131, opened 2016-11-14
Make water walking mechanics closer to original MW

Allofich:
When testing the recent water walking PR, I noticed that if you attempted to cast without room for the spell to move you out of the water, the spell would still succeed and you would have the visual "on hit" effect played, before the effect was then removed in applyQueuedMovement(), where the check for enough room was.

In original Morrowind if there is not room to be moved out of the water, you will get the "You cannot cast this effect right now." message and you don't have the on hit visual effect. This even happens if collision is disabled. This PR replicates this, and the effect doesn't have to be removed later any more since it won't be applied in the first place because the check for enough room is done before applying the effect.

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -985,6 +985,18 @@ namespace MWPhysics
         }
     }
 
+    bool PhysicsSystem::canMoveToWaterSurface(const MWWorld::ConstPtr &actor, const float waterlevel)
+    {
+        const Actor* physicActor = getActor(actor);
+        const float halfZ = physicActor->getHalfExtents().z();
+        const osg::Vec3f actorPosition = physicActor->getPosition();
+        const osg::Vec3f startingPosition(actorPosition.x(), actorPosition.y(), actorPosition.z() + halfZ);
+        const osg::Vec3f destinationPosition(actorPosition.x(), actorPosition.y(), waterlevel + halfZ);
+        ActorTracer tracer;
+        tracer.doTrace(physicActor->getCollisionObject(), startingPosition, destinationPosition, mCollisionWorld);
+        return (tracer.mFraction >= 1.0f);
+    }
+
     osg::Vec3f PhysicsSystem::getHalfExtents(const MWWorld::ConstPtr &actor) const
     {
         const Actor* physactor = getActor(actor);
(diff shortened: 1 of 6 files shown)


scrawl:
What would happen if you equipped an item with permanent effect "water walking"?

Allofich:
I just checked, and in original Morrowind you will get the "You cannot cast this effect right now." message and the water walking effect won't be applied. Even if you then move to a position where you can cast water walking, the effect won't be applied unless you unequip and re-equip the item.

In OpenMW with this PR, unfortunately, the character will be positioned up into the level geometry. I think equipping magic items bypasses the checking that is done for spells, and since I removed the checking in the movement update, the player is positioned without any checking.

If we want to emulate original MW behavior, I think constant effects from equipped magic items need to go through the same checks about whether they can be cast that spells do. This might also apply for bound equipment effects, I haven't checked.

Also, I think there may be another issue. I'm not completely sure about this, but I think that constant magic effects from items in OpenMW are constantly reapplied. In original Morrowind, they are just applied once at the time of equipping, then applied in reverse when unequipping. (This is also relevant for the drain and fortify effects, which I hope to get around to and update my PR on that)

Well, when I have some time I'll take a look at addressing the above issues and see how far I get.

Allofich:
Actually, I think the way the current OpenMW master handles constant effect items with water walking is good. It keeps trying to apply it so you don't have to re-equip the item after you get out of the uncastable spot. Also the "you cannot cast" message is a bit strange for a constant effect item, anyway, since it's not what I think of as "casting." I haven't fully decided but I'll probably just add in back the check during the movement update, rather than try to copy original MW behavior here.

Allofich:
Updated. I changed my mind and decided to mimic vanilla behavior. Otherwise, we have a situation where you could equip a constant effect water walking item while in deep water, which turns the effect "on" even though you aren't yet teleported to the top, and then when you swim to a shallow enough area, would you be teleported to the top from there or not? I think it wouldn't look good to suddenly have the player teleport to the surface with no accompanying sound effect or anything, but if we didn't it would be kind of inconsistent. I decided to follow what I see as the vanilla logic: you can never have the "water walking" effect active while you are in the water.

So, this now mimics vanilla behavior. If you cast water walking or use a constant effect water walking item either in too deep water or in shallow water but with something blocking you from teleporting up, you get the "cannot cast" message . If it was a constant effect item you will have to unequip it and re-equip it in an area where you can cast water walking in order to get the effect. 

This PR also fixes an issue, btw. With current master branch if you equip a constant effect water walking item, it will teleport you to the water surface no matter how deep you are, I think because the check for whether you can do this only checks whether there is a collision object in the way of you and the surface.

Also, since I made all effects from constant effect equipment go through the checkEffectTarget() function, I think this would also fix being able to use a constant levitate effect item in a no-levitate area.

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -1326,25 +1338,10 @@ namespace MWPhysics
...
-                        waterCollision = true;
-                        physicActor->setPosition(osg::Vec3f(actorPosition.x(), actorPosition.y(), waterlevel));
-                    }
-                    else
-                    {
-                        //Remove the effect to remove the performance hit of casting in a weird spot
-                        //probably makes that Tribunal quest where the water rises a bit safer
-                        iter->first.getClass().getCreatureStats(iter->first).getActiveSpells().purgeEffect(ESM::MagicEffect::WaterWalking);
-                    }
+                    physicActor->setPosition(osg::Vec3f(actorPosition.x(), actorPosition.y(), waterlevel));
```
scrawl: FYI, you forgot to set 'waterCollision' = true here :/

https://bugs.openmw.org/issues/3672

Allofich:
Ah, sorry. Thanks for fixing it in master.

PR #1132, opened 2016-11-16
Implement fleeing AI

MiroslavR:
Implements: https://bugs.openmw.org/issues/1118

There are some differences to the original behavior:
* As said in the report, OpenMW's existing AiCombat decision making (prepareNextAction()) differs from vanilla's. Probably not a big deal, though.
* Target switching: From what I can tell, vanilla calls canFight(actor, target) every frame on all combat targets to determine who to fight. The actor flees if no suitable target is found.
  * My implementation calls canFight within the AiCombat package, which is bound to a single target. I'm not sure if OpenMW supports target switching...

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -560,6 +560,12 @@ namespace MWBase
             virtual void removeContainerScripts(const MWWorld::Ptr& reference) = 0;
 
             virtual bool isPlayerInJail() const = 0;
+
+            /// Return terrain height at \a worldPos position.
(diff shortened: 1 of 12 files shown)


scrawl:
Target switching happens in AiSequence::execute. To be honest, the way we handle combat targets as an AiCombat package isn't optimal, because not all targets are supposed to be persistent (e.g. bug 1930).

Allofich:
Nice work! Interesting that the AI will run after summoning if it has no weapon, I didn't know that happened but I just tested in vanilla Morrowind and it does indeed happen there.

PR #1134, opened 2016-11-18
Show starting abilities, powers and spells in stat-review window

MiroslavR:
Fixes: https://bugs.openmw.org/issues/2410

### DIFF
--- a/apps/openmw/mwgui/charactercreation.cpp
+++ b/apps/openmw/mwgui/charactercreation.cpp
@@ -132,6 +132,12 @@ namespace MWGui
             mReviewDialog->configureSkills(major, minor);
     }
 
+    void CharacterCreation::onFrame(float duration)
+    {
+        if (mReviewDialog)
+            mReviewDialog->onFrame(duration);
+    }
+
     void CharacterCreation::spawnDialog(const char id)
     {
         try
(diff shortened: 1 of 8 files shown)


scrawl:
The spells section is not complete. In addition to the race and birthsign spells, you may get additional starting spells depending on major/minor skills chosen (see buildPlayer() function).

MiroslavR:
Interesting, I had no idea you get auto-calculated starting spells.

MiroslavR:
Done.

scrawl:
* Shouldn't the spells be sorted alphabetically for display? 
* What is the reason for aggregating the spells manually? Wouldn't just using the player's current spells result in shorter code?
* What is the reason for adding onFrame()? Why not do updates right away?

MiroslavR:
> Shouldn't the spells be sorted alphabetically for display?

They're not in vanilla.

> What is the reason for aggregating the spells manually? Wouldn't just using the player's current spells result in shorter code?

I was trying to preserve the vanilla display order (generated spells->racial spells->birth sign spells). Player's spells are stored in a map. Also I wanted to keep the "Review" class unaware of the player.

> What is the reason for adding onFrame()? Why not do updates right away?

Because otherwise updateSkillArea() would get called 30 times (in a single frame) after changing the character, which results in a noticeable freeze.

scrawl:
Makes sense.

Allofich:
There's some Clang warnings you may not have noticed.
https://travis-ci.org/OpenMW/openmw/jobs/177323057

MiroslavR:
Yeah, gcc doesn't have that warning. Thanks for letting me know.

PR #1136, opened 2016-11-20
Allow deleting spells that have the "always succeeds" flag

Allofich:
I tried three spells, two that have the always succeed flag already, and manually adding it to one, but in all cases I could delete it in original Morrowind but not in OpenMW. I'm curious why this condition was added in the first place.

### DIFF
--- a/apps/openmw/mwgui/spellwindow.cpp
+++ b/apps/openmw/mwgui/spellwindow.cpp
@@ -11,6 +11,7 @@
 #include "../mwworld/inventorystore.hpp"
 #include "../mwworld/class.hpp"
 #include "../mwworld/esmstore.hpp"
+#include "../mwworld/player.hpp"
 
 #include "../mwmechanics/spellcasting.hpp"
 #include "../mwmechanics/spells.hpp"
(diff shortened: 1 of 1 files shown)


MiroslavR:
Even if the spells come from a race/birth sign?

Allofich:
Those are all "powers," right? Those are still checked against. Is there some specific example you're thinking of?

MiroslavR:
No, they're not all powers.

From the bug report:
`It looks like the same spell effect attached to the Ritual birthsign.`

Allofich:
Ah, no, there is, for example, Beggar's Nose from the Tower that you can't delete in vanilla. It is listed under spells.

Allofich:
Posted at the same time :). Well I'm looking around at it. Feel free to suggest a solution if you have one.

Allofich:
Well even after removing the "always succeeds" flag from Beggar's Nose, it still can't be deleted in vanilla, so I think we can say for sure that it is not that flag. I can't yet find out how the decision is made, though.

Allofich:
Looks like you can delete the birthsign spells as long as they don't come from your birthsign.

Allofich:
This currently allows deleting spells with "always succeed" but not racial spells.

Unfortunately, after looking around in the code it seems like the player's birthsign is not being stored anywhere, that it is gotten rid of after character creation. Am I just missing it, or is it something that needs to be added to saved games?

scrawl:
> it seems like the player's birthsign is not being stored anywhere, that it is gotten rid of after character creation

It is saved in MWWorld::Player.

Allofich:
Oh, I guess I just overlooked it. Well I'll keep working on this, then.

Allofich:
I don't know how I managed to miss a function called getBirthSign(). Anyway, it seems to all be working now.

MiroslavR:
How about auto-generated starting spells? Can you delete those?

Allofich:
You mean from your class? Yes, I believe you can. I was able to delete things from a spell-casting class I set through enablestatreviewmenu. Just to be safe I'll go through and create a character the normal way and make sure.

Allofich:
Yes, I can delete them.

Allofich:
Sorry for reuploading again. Just a style thing, I put the & for the const std::string& signId next to string rather than next to the variable name, to keep the code in this PR consistent. There doesn't seem to be any consistency throughout the codebase, though.

MiroslavR:
> Yes, I can delete them.

I have also tested now and can confirm that.

PR #1138, opened 2016-11-22
Rebuild NPCs properly after undoing werewolf transformation

MiroslavR:
Calling UndoWerewolf on an NPC unequips all of the NPC's items before modifying the NpcStats werewolf state, which makes NpcAnimation not rebuild the NPC thinking that the NPC is still a werewolf (mNpcType hasn't changed). This results in a floating werewolf head being shown instead of an actual NPC body.


scrawl:
I went for a slightly different fix that keeps the order of setWerewolf/unequip consistent no matter if it's NPC or player (first setwerewolf, then change equipment). The comment that unequip needed to be first appears to be irrelevant as the NpcStats::setWerewolf() doesn't cause the NpcAnimation to be rebuilt.

MiroslavR:
Fair enough.

PR #1141, opened 2016-11-27
Implement water sounds

MiroslavR:
Implements: https://bugs.openmw.org/issues/451

The water volume factor has been reverse-engineered. However, the "Water Layer" sound (ESM::Sound volume = 102 or 0.4) is much quieter in OpenMW than in vanilla MW. I think this may be due to the exponential nature of this formula: https://github.com/OpenMW/openmw/blob/be14ce414e53cdc84513c84200418ccd9174d943/apps/openmw/mwsound/soundmanagerimp.cpp#L143.

This can also be observed with `PlaySound "Water Layer"`.

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -34,8 +34,9 @@
 
 namespace MWSound
 {
-    SoundManager::SoundManager(const VFS::Manager* vfs, bool useSound)
+    SoundManager::SoundManager(const VFS::Manager* vfs, const std::map<std::string,std::string>& fallbackMap, bool useSound)
         : mVFS(vfs)
+        , mFallback(fallbackMap)
         , mOutput(new DEFAULT_OUTPUT(*this))
         , mMasterVolume(1.0f)
         , mSFXVolume(1.0f)
@@ -60,6 +61,13 @@ namespace MWSound
         mVoiceVolume = std::min(std::max(mVoiceVolume, 0.0f), 1.0f);
         mFootstepsVolume = Settings::Manager::getFloat("footsteps volume", "Sound");
         mFootstepsVolume = std::min(std::max(mFootstepsVolume, 0.0f), 1.0f);
+
+        mNearWaterRadius = mFallback.getFallbackInt("Water_NearWaterRadius");
+        mNearWaterPoints = mFallback.getFallbackInt("Water_NearWaterPoints");
+        mNearWaterIndoorTolerance = mFallback.getFallbackFloat("Water_NearWaterIndoorTolerance");
+        mNearWaterOutdoorTolerance = mFallback.getFallbackFloat("Water_NearWaterOutdoorTolerance");
+        mNearWaterIndoorID = mFallback.getFallbackString("Water_NearWaterIndoorID");
+        mNearWaterOutdoorID = mFallback.getFallbackString("Water_NearWaterOutdoorID");
 
         mBufferCacheMin = std::max(Settings::Manager::getInt("buffer cache min", "Sound"), 1);
         mBufferCacheMax = std::max(Settings::Manager::getInt("buffer cache max", "Sound"), 1);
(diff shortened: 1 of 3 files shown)


apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -796,6 +804,96 @@ namespace MWSound
...
     }
 
+    void SoundManager::updateWaterSound(float /*duration*/)
+    {
+        MWBase::World* world = MWBase::Environment::get().getWorld();
+        const MWWorld::ConstPtr player = world->getPlayerPtr();
+        osg::Vec3f pos = player.getRefData().getPosition().asVec3();
+
+        float volume = 0.0f;
+        std::string soundId = player.getCell()->isExterior() ? mNearWaterOutdoorID : mNearWaterIndoorID;
```
kcat: This should probably be a ```const std::string&```

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -796,6 +804,96 @@ namespace MWSound
...
+                                float height = world->getTerrainHeightAt(
+                                            osg::Vec3f(pos.x() - mNearWaterRadius + x*step, pos.y() - mNearWaterRadius + y*step, 0.0f));
+
+                                if (height < 0)
+                                    underwaterPoints++;
+                            }
+                        }
+                    }
+
+                    volume *= underwaterPoints * 2.0f / (mNearWaterPoints*mNearWaterPoints);
```
kcat: This looks weird. The maximum possible volume depends on ```mNearWaterPoints```, such that when ```mNearWaterPoints == 10``` this can't be any higher than ```0.2```, and when ```mNearWaterPoints == 20``` it can't be any higher than ```0.1```. Perhaps ```(underwaterPoints*underwaterPoints) / (mNearWaterPoints*mNearWaterPoints)``` would work better?

kcat: Or wait, no, sorry. I misread the code and missed that ```underwaterPoints``` goes from ```0``` to ```mNearWaterPoints*mNearWaterPoints```, not ```mNearWaterPoints```.

kcat:
> I think this may be due to the exponential nature of this formula: https://github.com/OpenMW/openmw/blob/be14ce414e53cdc84513c84200418ccd9174d943/apps/openmw/mwsound/soundmanagerimp.cpp#L143.

That formula comes from vanilla. Since Morrowind used DirectSound, and DirectSound works using a millibel scale (decibel*100), it makes sense that vanilla's volume properties would be on a logarithmic scale (which we need to convert to linear, by applying an exponential formula). The 3348 scalar is kind of odd, but it's what I observed being passed to DirectSound for a given sound's volume (0 = -3348mB, 128 = -1667mB, 255 = 0mB). Converting millibels to linear gain is done by ```pow(10, mB/2000.0)```, and putting it all together results in that line.

MiroslavR:
Changes:
- used `const std::string&`
- prevented division by zero in `underwaterPoints * 2.0f / (mNearWaterPoints*mNearWaterPoints)`

> That formula comes from vanilla. Since Morrowind used DirectSound, and DirectSound works using a millibel scale (decibel*100), it makes sense that vanilla's volume properties would be on a logarithmic scale (which we need to convert to linear, by applying an exponential formula). The 3348 scalar is kind of odd, but it's what I observed being passed to DirectSound for a given sound's volume (0 = -3348mB, 128 = -1667mB, 255 = 0mB). Converting millibels to linear gain is done by pow(10, mB/2000.0), and putting it all together results in that line.

Thanks for the explanation.

It appears that with full master and effects volume in both vanilla and OpenMW, the real volume is roughly the same. With half master and effects volume, however, "Water Layer" (esmVolume = 102) sounds much quieter in OpenMW than vanilla while "Weapon Swish" (esmVolume = 255) sounds much quieter in vanilla than OpenMW...

I have delved a bit deeper into vanilla's sound subsystem and it looks like this is how volume is calculated and passed on to DirectSound:
```python
# "PlaySound" impl
vol = int(float(masterVolume * effectsVolume / 250) * (esmVolume / 255.0))
vol = -3333 - int(vol * (-13.0706)) # conversion to mB

# Water sound impl
vol = int(float(masterVolume * effectsVolume / 250) * waterVolumeFactor)
vol = int(float(esmVolume) * float(vol) / 250.0) # bug?
vol = -3333 - int(vol * (-13.0706)) # conversion to mB
```

where:
- masterVolume, effectsVolume are ints in <0, 250>
- esmVolume is an int in <0, 255>
- waterVolumeFactor is a float in <0.0, 1.0>

MiroslavR:
There's still the issue that all of our volumes should be on a logarithmic scale to mimic vanilla behavior. Currently the SoundManager formula only applies to ESM::Sound volume. But I guess that can be tackled separately if it turns out to be a huge problem.

PR #1143, opened 2016-12-03
Crash fix when item is disabled before it casts spell.

kurnevsky:
Madd leveler has item with such script:

Begin Madd_Leveler_Effect
	Disable
	Short doOnce
	If ( doOnce == 0 )
		Set doOnce To 1
		Cast "mlEffectSpell" Player
	EndIf
End

If add this item with PlaceAtPC openmw will crash.


psi29a:
Nice find!

PR #1144, opened 2016-12-04
Feature #3523:  Light source on magic projectiles

Aussiemon:
[Feature #3523](https://bugs.openmw.org/issues/3523)

Attached light to magic bolt projectiles.

[Thread discussion of this pull request](https://forum.openmw.org/viewtopic.php?f=6&t=3972)

### DIFF
--- a/apps/openmw/mwworld/projectilemanager.cpp
+++ b/apps/openmw/mwworld/projectilemanager.cpp
@@ -173,10 +173,12 @@ namespace MWWorld
         {
             // Add magic bolt light
             osg::ref_ptr<osg::Light> projectileLight(new osg::Light);
-            projectileLight->setDiffuse(osg::Vec4(0.95f, 0.71f, 0.25f, 1.0f));
-            projectileLight->setAmbient(osg::Vec4(0.32f, 0.08f, 0.01f, 1.0f));
-            projectileLight->setSpecular(osg::Vec4(0, 0, 0, 0));
-            projectileLight->setLinearAttenuation(0.5f / 15);
+            projectileLight->setAmbient(osg::Vec4(1.0f, 1.0f, 1.0f, 1.0f));
+            projectileLight->setDiffuse(osg::Vec4(0.814f, 0.682f, 0.652f, 1.0f));
+            projectileLight->setSpecular(osg::Vec4(30.0f, 30.0f, 30.0f, 1.0f));
+            projectileLight->setConstantAttenuation(0.f);
+            projectileLight->setLinearAttenuation(0.1f);
+            projectileLight->setQuadraticAttenuation(0.f);
             projectileLight->setPosition(osg::Vec4(pos, 1.0));
             
             // Add magic bolt light source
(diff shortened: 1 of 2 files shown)


apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include <iomanip>
 
+#include <osg/Light>
+#include <osg/LightSource>
```
scrawl: Correct include is components/sceneutil/lightmanager.hpp

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -136,7 +138,7 @@ namespace MWWorld
     };
 
 
-    void ProjectileManager::createModel(State &state, const std::string &model, const osg::Vec3f& pos, const osg::Quat& orient, bool rotate, std::string texture)
+    void ProjectileManager::createModel(State &state, const std::string &model, const osg::Vec3f& pos, const osg::Quat& orient, bool rotate, bool isMagic, std::string texture)
```
scrawl: isMagic should be renamed to createLight to reflect the variable's real purpose.

Aussiemon:
Retrieval of colors from the relevant magic effects is complete. Testing on several spells shows that the blended diffusion values match the projectile colors in vanilla Morrowind.

scrawl:
> Retrieval of colors from the relevant magic effects is complete. Testing on several spells shows that the blended diffusion values match the projectile colors in vanilla Morrowind.

That sounds very.. scientific ;)

MiroslavR:
@scrawl, can we agree not to use C-style casts anymore?

scrawl:
Oh, I didn't notice the casts.

A ```static_cast<float>``` would be preferred over ```(float)```, but these are a very minor issue that I'm not going to bother bringing up in review.

Regarding this one:
<pre>
int numberOfEffects = ((MagicBoltState&)state).mEffects.mList.size();
</pre>
Is a problem, the function will crash if for some reason createLight is set to true for a non-magic projectile.

@Aussiemon: could you make a new PR updating the code to not use C-style casts, and remove the MagicBoltState cast? It's probably easiest to just compute the light color before hand and then pass it to createModel.

MiroslavR:
> A static_cast<float> would be preferred over (float), but these are a very minor issue that I'm not going to bother bringing up in review.

I was primarily referring to non-trivial types. The ability to fall back to reinterpret_cast sounds like an accident waiting to happen. C++ casts express their intent much more clearly.

scrawl:
Yes, that is my stance. For non-trivial types, we should always use C++ casts.

In this case, the cast shouldn't be there in the first place though.

MiroslavR:
> In this case, the cast shouldn't be there in the first place though.

Agreed.

Aussiemon:
Thanks for the review of my work. I'll definitely take this into account in the future. 

I've created a [pull request](https://github.com/OpenMW/openmw/pull/1151) with the recommended changes to casts. If there's a better way to make these fixes, I'd be happy to learn.

PR #1145, opened 2016-12-05
Bug #3647: Reduced loudness of multiple projectiles to a sum total of volume 1

Aussiemon:
[Issue tracker link](https://bugs.openmw.org/issues/3647)

Related to [feature #3519](https://bugs.openmw.org/issues/3519)

Playing every magical effect's audio simultaneously increases the total volume of the projectile cast by log10(number of sounds) [{citation}](https://soundphysics.ius.edu/?page_id=807). This is making a few assumptions about their wave properties, but is at least a reasonable approximation. Calculating the volume with 1 / (log10(number of sounds) + 1) counteracts this effect.

1 sound = 100% volume per sound
2 sounds = 77% volume per sound
8 sounds = 53% volume per sound

Testing this fix, the volumes now seem equivalent, at least to my ears.

Please let me know if I can make any style improvements. Still a lot to learn here!

### DIFF
--- a/apps/openmw/mwworld/projectilemanager.cpp
+++ b/apps/openmw/mwworld/projectilemanager.cpp
@@ -284,7 +284,7 @@ namespace MWWorld
         float volumeCalculation = 1.0f;
         if (state.mSoundIds.size() > 1)
         {
-            volumeCalculation = (1.0f / (std::log10(state.mSoundIds.size() + 1.0f)));
+            volumeCalculation = (1.0f / (std::log10(state.mSoundIds.size()) + 1.0f));
         }
         for (size_t it = 0; it != state.mSoundIds.size(); it++)
         {
@@ -578,7 +578,7 @@ namespace MWWorld
             float volumeCalculation = 1.0f;
             if (state.mSoundIds.size() > 1)
             {
-                volumeCalculation = (1.0f / (std::log10(state.mSoundIds.size() + 1.0f)));
+                volumeCalculation = (1.0f / (std::log10(state.mSoundIds.size()) + 1.0f));
             }
             for (size_t soundIter = 0; soundIter != state.mSoundIds.size(); soundIter++)
             {
--- a/apps/openmw/mwworld/projectilemanager.cpp
+++ b/apps/openmw/mwworld/projectilemanager.cpp
@@ -281,9 +281,14 @@ namespace MWWorld
         createModel(state, ptr.getClass().getModel(ptr), pos, orient, true, true, texture);
 
         MWBase::SoundManager *sndMgr = MWBase::Environment::get().getSoundManager();
+        float volumeCalculation = 1.0f;
+        if (state.mSoundIds.size() > 1)
+        {
+            volumeCalculation = (1.0f / (std::log10(state.mSoundIds.size() + 1.0f)));
+        }
         for (size_t it = 0; it != state.mSoundIds.size(); it++)
         {
-            state.mSounds.push_back(sndMgr->playSound3D(pos, state.mSoundIds.at(it), 1.0f, 1.0f, MWBase::SoundManager::Play_TypeSfx, MWBase::SoundManager::Play_Loop));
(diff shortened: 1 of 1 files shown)


Aussiemon:
Related to this: Was submitting the bug report, then pull request, the correct way of handling the issue? Should I have waited a bit? Submitted the pull request without the bug report, maybe?

Allofich:
Hi Aussimon, I was going to comment on your bug report later, but I'll do it here.

I'm aware of an issue involving the sound effects of spells playing. That is, that in vanilla, my testing suggests that an object (NPC, door, etc.) can only play one instance of a given sound at a time. Playing the sound again cancels the first instance and starts it over. Easy way to test is to go up to a locked door and activate it a bunch in a row. In vanilla the sound will keep restarting, while in OpenMW it will play on top of itself.

I believe this causes the spells with multiple effects to sound loud when the same sound effect is played on itself like this. Have you seen evidence that the volume itself is different, though?

Allofich:
https://bugs.openmw.org/issues/3490 Here is the relevant bug report (R.D. is me). I posted a bunch there and went back and forth on the issue (kind of embarrassing) until I figured out it seemed to be a general sound issue, not just something for spells. I alleviated the issue I was having by upgrading my OpenAL and ended up writing off the issue, but now I'm kind of leaning towards wanting to mimic the vanilla behavior, because I still think some spells with the restoration sound effect doubled up are unpleasant to listen to.

Is this the same issue you are talking about?

Allofich:
> Was submitting the bug report, then pull request, the correct way of handling the issue? Should I have waited a bit? Submitted the pull request without the bug report, maybe?

I'm not the authority on this or anything, but I think it's probably good to open a bug report as you did, because the people who make the change log when an OpenMW release is made use the bug reports that were closed as a guide. If you don't file a bug report your fix may not make it into the change log.

Aussiemon:
If the sounds are identical and play at the same time (identical waves, frequencies, and phase), their volumes would be combined (at least according to the page I linked) with a total volume of 1 + log10(number of sounds). That could definitely be why they seem to be playing louder than usual in the case of the door and spell sounds.

So, really, I think my fix would be obsoleted by preventing the same sound from playing multiple times. I don't know if that's an eventual goal or not though, so this calculation could still be useful in the meantime. 

Thanks for the advice regarding when to submit bug reports. It would be pretty cool to make it into the change log, after all.

kcat:
> Playing every magical effect's audio simultaneously increases the total volume of the projectile cast by log10(number of sounds)

When multiple audio samples are played simultaneously, the linear amplitude adds together. So if you play the same exact sample twice at the same time (no phase difference), the volume is doubled on a linear scale (the same scale of OpenMW's volume parameters). Halving the volume (specifying a volume of 0.5 for both sounds) would counteract that and create a result that's the same as playing one sample at full volume. In which case, you'd calculate the volume as ```1.0f / (float)number_of_sounds```.

If you're playing two different samples, it gets a bit trickier. The linear amplitudes still add together, but the energy is spread out into different frequencies by different amounts. The unique frequencies in each sound will be perceived at normal volume, even though the total energy was increased with the two sounds added together (it's just not piling up on the same frequencies). In these cases, there's no simple way to scale the volume to keep the unique frequencies normal without allowing non-unique frequencies to increase.

Given that, what Allofich said about vanilla Morrowind only allowing one instance of a given sound on an object makes more sense, rather than explicit scaling based on the number of sounds.

Aussiemon:
> When multiple audio samples are played simultaneously, the linear amplitude adds together. So if you play the same exact sample twice at the same time (no phase difference), the volume is doubled on a linear scale (the same scale of OpenMW's volume parameters). Halving the volume (specifying a volume of 0.5 for both sounds) would counteract that and create a result that's the same as playing one sample at full volume. In which case, you'd calculate the volume as 1.0f / (float)number_of_sounds.

That was actually my original implementation. I hadn't realized Morrowind was using a linear scale.

> Given that, what Allofich said about vanilla Morrowind only allowing one instance of a given sound on an object makes more sense, rather than explicit scaling based on the number of sounds.

I agree. It would be the simplest and truest method, I think.

PR #1146, opened 2016-12-05
Fix fortify maximum magicka expiration

Allofich:
This fixes the problem of temporary fortify maximum magicka effects lasting forever.

However, I would like to get the opinion of @MiroslavR on this, because this PR removes a piece of code he added. The code was added in https://github.com/OpenMW/openmw/commit/80f2ae0ca7258bef6742d468aae7fa0828195b2e
and I don't fully understand it or if any possible problems could arise from removing this line.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -512,8 +512,8 @@ namespace MWMechanics
                 if (magnitude > 0 && remainingTime > 0 && remainingTime < mDuration)
                 {
                     CreatureStats& creatureStats = mActor.getClass().getCreatureStats(mActor);
-                    effectTick(creatureStats, mActor, key, magnitude * remainingTime);
-                    creatureStats.getMagicEffects().add(key, -magnitude);
+                    if (effectTick(creatureStats, mActor, key, magnitude * remainingTime))
+                        creatureStats.getMagicEffects().add(key, -magnitude);
                 }
             }
     };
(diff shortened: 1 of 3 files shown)


Allofich:
The problem that occurs with the line I removed, is that it sets the magnitude of a fortify maximum magicka effect back to 0 when it ends before `creatureStats.modifyMagicEffects`, which is called in `Actors::adjustMagicEffects`, can check for a changed magnitude and set the flag to recalculate magicka. It compares two `MagicEffects` classes but due to the line I removed setting the magnitude to 0, one of them is updated too soon and so when they are compared they are the same and the flag doesn't get set.

MiroslavR:
The reason why ExpiryVisitor was added was to prevent tickable effects (such as Restore Health) from being applied across the entire duration (referring to the calculateCreatureStatModifiers() parameter) if this duration exceeded the effect's remaining duration. ExpiryVisitor aimed to fix this by correcting the magnitude supplied to effectTick().

For example, imagine you had a Restore Health effect that was about to expire in 3 seconds. By waiting 1 in-game hour (120 RL seconds with default timescale IIRC), the effect would be applied for whole 120 seconds instead of the remaining 3.

The line you are removing prevented the loop following ExpiryVisitor from applying the effect magnitude again for 120 seconds. By removing it you are making matters worse because now ExpiryVisitor will apply mag\*3 and the following loop will apply mag\*120.

Of course I did not realize this would have an impact on Fortify Maximum Magicka, which is not even a tickable effect. Probably not the ideal solution, but I guess one way to fix this would be to have effectTick() return a bool indicating whether something was applied, similarly to applyInstantEffect(), and then modify ExpiryVisitor::visit() to:

```c++
                if (magnitude > 0 && remainingTime > 0 && remainingTime < mDuration)
                {
                    CreatureStats& creatureStats = mActor.getClass().getCreatureStats(mActor);
                    if (effectTick(creatureStats, mActor, key, magnitude * remainingTime))
                        creatureStats.getMagicEffects().add(key, -magnitude);
                }
```

Allofich:
Thanks for the explanation.I implemented your suggestion, and it works.

I hope you don't mind that I also added a comment that the expiryvisitor code is relevant to when using the rest/wait command before an effect ends. I hadn't realized this was happening, so it might help other people also understand what is happening here.

MiroslavR:
@scrawl, if there are no objections from your side, I think we can merge.

PR #1147, opened 2016-12-05
Implement NiSphericalCollider (Closes #3644)

MiroslavR:
Implements: https://bugs.openmw.org/issues/3644

A few notes:

* I'm not completely sure of the significance of the vector **u**. I guess the `(u.length2() < mSphereInParticleSpace.radius2())` check could be replaced with `discriminant > 0` and the discriminant formula could be modified to not use **u**. I decided to leave it in to make sure the behavior matches vanilla's.
* Tunneling should be prevented by `(k < dt)`, although the particle is not moved to the point of contact. Edit: Particles inside the sphere can escape if given large initial velocities and/or bounceFactor, but that is also true in vanilla MW.

### DIFF
--- a/components/nifosg/particle.cpp
+++ b/components/nifosg/particle.cpp
@@ -376,4 +376,73 @@ void PlanarCollider::operate(osgParticle::Particle *particle, double dt)
     }
 }
 
+SphericalCollider::SphericalCollider(const Nif::NiSphericalCollider* collider)
+    : mBounceFactor(collider->mBounceFactor),
+      mSphere(collider->mCenter, collider->mRadius)
+{
+}
+
+SphericalCollider::SphericalCollider()
+    : mBounceFactor(1.0f)
+{
+
+}
+
+SphericalCollider::SphericalCollider(const SphericalCollider& copy, const osg::CopyOp& copyop)
+    : osgParticle::Operator(copy, copyop)
+    , mBounceFactor(copy.mBounceFactor)
+    , mSphere(copy.mSphere)
+    , mSphereInParticleSpace(copy.mSphereInParticleSpace)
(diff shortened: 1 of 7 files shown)


components/nifosg/particle.cpp
```diff
@@ -376,4 +376,71 @@ void PlanarCollider::operate(osgParticle::Particle *particle, double dt)
     }
 }
 
+SphericalCollider::SphericalCollider(const Nif::NiSphericalCollider* collider)
+    : mBounceFactor(collider->mBounceFactor),
+      mSphere(collider->mCenter, collider->mRadius)
+{
+}
+
+SphericalCollider::SphericalCollider()
```
scrawl: Initialize mBounceFactor here, or else coverity will complain.

components/nifosg/particle.cpp
```diff
@@ -376,4 +376,71 @@ void PlanarCollider::operate(osgParticle::Particle *particle, double dt)
...
+{
+
+}
+
+SphericalCollider::SphericalCollider(const SphericalCollider& copy, const osg::CopyOp& copyop)
+    : osgParticle::Operator(copy, copyop)
+    , mSphere(copy.mSphere)
+    , mSphereInParticleSpace(copy.mSphereInParticleSpace)
```
scrawl: Forgot to copy mBounceFactor?

components/nifosg/particle.cpp
MiroslavR: Fixed.

components/nifosg/particle.cpp
MiroslavR: Fixed.

PR #1148, opened 2016-12-06
Require line of sight for AI attacks

Allofich:
Seems to be working correctly.

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -60,30 +60,30 @@ namespace MWMechanics
             FleeState_RunToDestination
         };
         FleeState mFleeState;
-        bool mFleeLOS;
-        float mFleeUpdateLOSTimer;
+        bool mLOS;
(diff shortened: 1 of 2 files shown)


scrawl:
LOS should be updated on a timer, too expensive to do every frame for (potentially) every actor.

Allofich:
OK, thanks. I'm trying now to copy the code that MiroslavR used for the LOS checks in AI Flee, which use a timer, to be shared LOS code.

Allofich:
I borrowed the LOS code from MiroslavR's flee code and made it the general use LOS code for AI combat. @MiroslavR, I generalized some LOS variables that were unique to the flee code, but are you OK with these changes? I think you worked pretty hard to get an accurate recreation of the vanilla flee code and I don't want to mess it up. You were resetting the AI timer in stopFleeing() but I removed that because it's being shared now. Should the flee code continue to have its own timer that is always reset to 0 at the end?

MiroslavR:
Looks OK to me.

Allofich:
Good to hear, thanks for looking at it.

PR #1149, opened 2016-12-06
A couple movement solver (micro) optimizations.

logzero:
Don't really help with the worst case performance though.

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -223,7 +223,7 @@ namespace MWPhysics
                 collisionWorld->rayTest(from, to, resultCallback1);
 
                 if (resultCallback1.hasHit() &&
-                        ( (toOsg(resultCallback1.m_hitPointWorld) - tracer.mEndPos).length() > 35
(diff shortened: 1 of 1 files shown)


scrawl:
Do those make a measurable difference? I'm sure the bottleneck are the bullet convex casts.

logzero:
Yeah, the casts blow everything out of the water. Still no reason to waste cpu cycles on needless calculations.

Btw I think it might be possible to reduce the number of casts by taking hitpoint position into account.

scrawl:
I appreciate you looking into this, but we normally don't accept this type of micro optimization that isn't proven to make a measurable difference. We like to keep changes to a minimum to avoid merge conflicts and not risk new bugs.

Also, your changes make the code slightly less readable when it comes to the meaning of maxSlope (i.e. what was < maxSlope is now > maxSlope, which doesn't make much sense). Not normalizing the normal I am not sure if that is safe, the bullet doc's don't say if its already normalized.

> Btw I think it might be possible to reduce the number of casts by taking hitpoint position into account.

I am not sure what you mean exactly, but it sounds worth looking into.

PR #1150, opened 2016-12-07
[Incomplete] Bug #2777: Fix for disintegration efffects eliminated by framerate

Aussiemon:
Bug: [#2777](https://bugs.openmw.org/issues/2777)

Idea:

- Only one `mCharge`, a float value, in `esm::CellRef`.

- `setCharge(float)` is used everywhere, replacing both `setCharge(int)` and `setChargeFloat(float)`. It changes `mCharge` if `(std::abs(newCharge - oldCharge) > .00001f)`. which should be enough precision for 1 point of disintegration spread out over 100k ticks; if I've done my math correctly. Maybe this is overkill.

- `int getCharge()` (for usual functions) will now return the rounded charge value as an int. `float getChargeFloat()` (for lights and disintegration effects) will return the unrounded value.

Testing:

The implementation seems to work fine, except that about 1 in 10 disintegration points aren't being applied. A constant effect amulet with 1 point of disintegration worked without visible issue. A spell that applies ten points of disintegration applied 9-10 (0 points before this fix). A spell that applies 100 points of disintegration applied 92-99 (64-80 points before this fix). NPC's seem to take normal durability damage. Lights still extinguish in the proper amount of time.

If there are any mistakes or problems with my code, please don't hesitate to let me know! I'm particularly worried that this will effect savegames in some negative way.

### DIFF
--- a/apps/esmtool/esmtool.cpp
+++ b/apps/esmtool/esmtool.cpp
@@ -269,7 +269,7 @@ void loadCell(ESM::Cell &cell, ESM::ESMReader &esm, Arguments& info)
         std::cout << "    Faction: '" << ref.mFaction << "'" << std::endl;
         std::cout << "    Faction rank: '" << ref.mFactionRank << "'" << std::endl;
         std::cout << "    Enchantment charge: '" << ref.mEnchantmentCharge << "'\n";
-        std::cout << "    Uses/health: '" << ref.mChargeInt << "'\n";
+        std::cout << "    Uses/health: '" << ref.mCharge << "'\n";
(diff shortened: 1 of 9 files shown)


Aussiemon:
Need to make changes to OpenCS files as well, it seems.

scrawl:
You can't change the esm format like that. That breaks plugin files and save files.

Aussiemon:
> You can't change the esm format like that. That breaks plugin files and save files.

I see what you mean. I've written a [more lightweight fix](https://github.com/OpenMW/openmw/issues/1152) that shouldn't affect savegames and plugins.

PR #1152, opened 2016-12-10
Bug #2777: Fix for disintegration efffects eliminated by framerate (attempt #2)

Aussiemon:
Bug: [#2777](https://bugs.openmw.org/issues/2777)
Original pull request: [#1150](https://github.com/OpenMW/openmw/pull/1150)

Idea:

- Add a float to `ESM::CellRef` called `mChargeIntRemainder`, representing the unsubtracted remainder of subtractions applied to `mChargeInt`.

- Add a method to `MWWorld:CellRef` called `applyChargeRemainderToBeSubtracted(float)`. This method increases `mChargeIntRemainder` by the given value, and applies the subtraction when `mChargeIntRemainder > 1`.

Disintegration is the only use of the new method and variable right now. Everything else should be unaffected, if I'm reading the code correctly.

Testing:

Disintegration works well, with 98-99% of the effect applied, regardless of magnitude. There seem to be no save or plugin compatibility issues on my end with this much-more lightweight fix. The remainder will not be saved or loaded, but I think this is fine to maintain compatibility, and should not have much effect.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -980,8 +980,10 @@ namespace MWMechanics
                 if (charge == 0)
                     return false;
 
-                // FIXME: charge should be a float, not int so that damage < 1 per frame can be applied.
-                // This was also a bug in the original engine.
+                // Store remainder of disintegrate amount (automatically subtracted if > 1)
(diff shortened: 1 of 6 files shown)


scrawl:
Looks good to me, the fix is a bit hack-ish but it's the best we can do given the circumstances. Discarding the remainder shouldn't be a big issue.

PR #1156, opened 2016-12-13
Minor grammar and sentence changes

jeffreyhaines:


docs/source/openmw-cs/tour.rst
```diff
@@ -3,8 +3,8 @@ A Tour through OpenMW CS: making a magic ring
 
 In this first chapter we will create a mod that adds a new ring with a simple
 enchantment to the game. The ring will give its wearer a permanent Night Vision
-effect while being worn. You don't need prior knowledge about modding
-Morrowind, but you should be familiar with the game itself. There will be no
+effect while being worn. You do not need prior Morrowind modding experience,
```
psi29a: You pull apart a contraction don't->do not, but you left prior?  Prior is a pompous term when previous would do just fine. :)

docs/source/openmw-cs/tour.rst
```diff
@@ -13,7 +13,7 @@ installed and playable.
 Adding the ring to the game's records
 *************************************
 
-In this first section we will define what our new ring is, what it looks like
+In this first section we will define what our new ring is, what it looks like,
```
psi29a: That extra comma needs to go, it should not be there. The APA and MLA can go get stuffed, the rest of the English speaking world is against the serial comma.

docs/source/openmw-cs/tour.rst
jdh6660: I agree.

docs/source/openmw-cs/tour.rst
jdh6660: I think most people prefer the Oxford comma.

docs/source/openmw-cs/tour.rst
```diff
@@ -58,7 +58,7 @@ detached panel can be re-attached to a window by dragging it by the title bar
 on top of the window.
 
 Now let's look at the panel itself: we have a filter text field, a very large
-table and a status bar. The filter will be very useful when we want to find an
+table, and a status bar. The filter will be very useful when we want to find an
```
psi29a: You added a comma when it wasn't needed and left 'an' before word beginning with a consonant. :/

docs/source/openmw-cs/tour.rst
```diff
@@ -95,7 +95,7 @@ holding Shift to edit it (this is a configurable shortcut), but there is a
 better way: right-click the row of our new record and chose *Edit Record*, a
 new panel will open.
 
-We can right-click the row of our new record and chose *Edit Record*, a
+We can right-click the row of our new record and choose *Edit Record*, a
```
psi29a: Like you did above, perhaps select is a better word choice.

docs/source/openmw-cs/tour.rst
jdh6660: Are you talking about "an entry"? I'm pretty sure e is a vowel.

docs/source/openmw-cs/tour.rst
jdh6660: I agree.

docs/source/openmw-cs/tour.rst
jeffreyhaines: I agree.

docs/source/openmw-cs/tour.rst
jeffreyhaines: Are you referring to the "an entry"?

docs/source/openmw-cs/tour.rst
psi29a: The Oxford comma is dead.

 http://www.adweek.com/galleycat/oxford-comma-dropped-by-university-of-oxford/34361?red=as

docs/source/openmw-cs/tour.rst
psi29a: You're quite right, I must have been reading the following line incorrectly.

The unnecessary Oxford comma still stands though. :)

psi29a:
A few fixups, then we can merge this in. Thank you again for reviewing the text. I'm glad there are more eyes on it! :)

psi29a:
We should have a style-guideline... reading: No Oxord Commas. ;)

jeffreyhaines:
Idk, I think you are just against it. I would think the best way to go about this is to pull the community, since that's who these guides are for anyway. Lets create a pull :)

jeffreyhaines:
Removed the Oxford comma. At least we have the proper usage of choose and chose.

docs/source/openmw-cs/tour.rst
jeffreyhaines: Oooooo 2 to 1 Oxford ;)

psi29a:
Small fix during merge conflict, merging now.  Thank you! :)

Yes, I (and my journalist wife) are biased against the serial/oxford comma as it adds nothing to a list aside from cluttering it up.

kcat:
In my mind, it helps to clearly delineate each thing in the list. After all, it wouldn't make sense to have a choice between three things like this: cereal, bacon and eggs and ham and cheese. It helps the statement to flow better with it, as each list entry is separated clearly and consistently, and identifies it as being a list rather than some kind of identification like this: the target, an Ascended Sleeper and Dagoth Velos (are the Ascended Sleeper and Dagoth Velos the target, or is the target something else that's with them?).

psi29a:
@kcat You hit the exception, called a cluster which happens is most languages. The use of the comma is used to help differentiate between these and clusters. Even the BBC makes an exception here which would otherwise eschew any extraneous commas.

Example: http://content.tfl.gov.uk/tfl-editorial-style-guide.pdf
> Note: A comma is not required before and at the end of a list unless one of the listed items includes another ‘and’: We ordered chocolate cake, cheese and biscuits, and ice cream

kcat:
@psi29a Right, but that's what I mean. Why only have it around for exceptions, when a situation that calls for it may not be immediately obvious? Or where one person may intuit the meaning without the extra comma while someone else may not? Why not just be consistent and avoid the problem entirely?

Additionally, having it adds structure. Not including the extra comma in my eye is similar to having a list like this:
* Entry one
* Entry two

Entry three. It visually removes the final item from the otherwise comma-separated list. Should we kill the Oxford bullet-point too? :P

psi29a:
Considering that Oxford killed their own comma, we just need to wait around long enough for the rest to come full circle. :P

If you want to be internally consistent with how bullets are used, then drop the 'and' all together and use only commas exclusively to indicate individual items.

The Shatner and Walken commas are perfectly acceptable within context. ;)
http://liveoakcommunications.com/wp-content/uploads/2016/03/walken.png

jeffreyhaines:
Let's follow what Oxford recommends, and use the comma when it removes ambiguity.

https://www.ox.ac.uk/sites/files/oxford/media_wysiwyg/University%20of%20Oxford%20Style%20Guide.pdf#page14

psi29a:
I'm OK with this. They even go to say:

> Use semicolons in place of commas in a complicated list or sentence if it will improve clarity, particularly if list items already include commas.

This is inline what I posted above as well.

jeffreyhaines:
Yeah I agree with what Oxford is recommending. But lets not use that as an excuse to write lists with semicolons. That should really only happen once if ever. We should be concerned with correct spelling, good writing and then technicalities.

psi29a:
Agreed. :)

PR #1157, opened 2016-12-14
Multiple spelling fixes

lkrieg:
Fixed multiple spelling mistakes. Most of the changes only affect comments, but there are also two modified variable names (not referenced from anywhere else). I double-checked everything and made sure the code is building without errors.

Files with changes to variable names are:
- apps/opencs/model/world/idtree.cpp
- apps/wizard/inisettings.cpp

NOTE: I'm not quite sure if I should open separate pull requests for each fix since there are quite a few. If you would like to have them seperate, just say so and I'll resubmit.

This is my first pull request, please be gentle! :-)

### DIFF
--- a/apps/essimporter/importscri.hpp
+++ b/apps/essimporter/importscri.hpp
@@ -13,7 +13,7 @@ namespace ESM
 namespace ESSImport
 {
 
-    /// Local variable assigments for a running script
+    /// Local variable assignments for a running script
     struct SCRI
     {
         std::string mScript;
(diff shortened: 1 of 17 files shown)


psi29a:
Thank you for doing this.

scrawl:
Thanks, but could you squash these into one commit, no need to make a commit for every word changed.

lkrieg:
Of course! I'll do that as soon as I get back home.

lkrieg:
Done!

PR #1158, opened 2016-12-14
[RFC] Movement solver experiments

logzero:
This is my most recent modification set. The last commit is probably the most interesting, skips redundant step up check, and also step over/down when previous step over already hit an obstacle.

It is possible to skip even more step up checks whenever there was no change in position from one to next iteration. 

I've also been looking at using hitPointWorld from the cast. But the hit point reported is not always reliable (despite a high hit point, step over still can succeed), need to figure out what is going on there before trying to use it.

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -419,7 +419,7 @@ namespace MWPhysics
 
                     if ((newVelocity-velocity).length2() < 0.01)
                         break;
-                    if ((velocity * origVelocity) <= 0.f)
+                    if ((newVelocity * origVelocity) <= 0.f)
                         break; // ^ dot product
 
                     velocity = newVelocity;
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -402,17 +402,18 @@ namespace MWPhysics
                     reflectdir.normalize();
 
                     osg::Vec3f newVelocity = slide(reflectdir, tracer.mPlaneNormal)*movelen;
+
+                    // Do not allow sliding upward if there is gravity.
+                    // Stepping will have taken care of that.
+                    if(!(newPosition.z() < swimlevel || isFlying))
+                        newVelocity.z() = std::min(newVelocity.z(), 0.0f);
+
                     if ((newVelocity-velocity).length2() < 0.01)
                         break;
                     if ((velocity * origVelocity) <= 0.f)
                         break; // ^ dot product
 
                     velocity = newVelocity;
-
-                    // Do not allow sliding upward if there is gravity. Stepping will have taken
-                    // care of that.
-                    if(!(newPosition.z() < swimlevel || isFlying))
-                        velocity.z() = std::min(velocity.z(), 0.0f);
                 }
             }
 
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -395,6 +395,13 @@ namespace MWPhysics
                     break;
                 }
 
+                // We are touching something.
+                if (tracer.mFraction < 1E-9f)
+                {
+                    // Try to separate by backing off slighly to unstuck the solver
+                    const osg::Vec3f backOff = (newPosition - tracer.mHitPoint) * 1E-3f;
+                    newPosition += backOff;
+                }
 
                 // We hit something. Check if we can step up.
                 float hitHeight = tracer.mHitPoint.z() - tracer.mEndPos.z() + halfExtents.z();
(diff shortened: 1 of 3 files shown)


logzero:
It turns out that the hit point reported by the cast is working just fine.

My confusion about it was due to the fact that the initial forward trace is rejected even when it is partially successful fraction < 1. This will make the step up/over seem successful when it actually isn't.

logzero:
So I've done a bit testing in Balmora, running into obstacles, and so far have not been able to get my framerate below 30. With vanilla openmw I can get below 10 fps.

There are still a cases where the solver fails though.

psi29a:
What are those use-cases?

logzero:
One general issue is that when you are in contact with something casts will fail independent of their direction. The back off "hack" helps here, but still fails sometimes.

At times the solver just fails to find a good direction due to a shallow contact angle or bad contact normal on step down. It might be possible to use the hit point position to figure out a better new direction.

And there are cases where a vertical velocity component is involved, which aren't handled that well in general. But they seem to be pretty rare.

logzero:
> Does this fix sliding along walls?

No, it is focusing mostly on performance.

The sliding depends on the wall contact normal. What happens I think is that bullet reports wall (triangle mesh) edge collisions when it should not. But this is just a guess from me atm, would need testing to verify.

One thing I've observed is that using a capsule shape for actors instead of cylinder helps with sliding, so it might be dependent on the collision detection algorithm.

Unfortunately capsules somehow break jumping in some cases. I don't know why yet.

0tt:
I hate capsules, it makes platforming impossible because you end up sliding
down a drop off even if you're not completely off. Would using a box be out
of the question?

wareya:
>One thing I've observed is that using a capsule shape for actors instead of cylinder helps with sliding, so it might be dependent on the collision detection algorithm.

This is natural. AABBs get stuck in walls if the collision data allows the AABB's surface to collide with the walls' collision's  edges, and uses the AABB's surface's normal to derive the contact normal when that happens. With a cylinder, you don't get stuck on walls because the contact normal always corresponds to the direction the collision is in.

0tt:
> This is natural. AABBs get stuck against walls if the collision data allows the AABB's surface to collide with the walls' collision's edges, and uses the AABB's surface's normal to derive the contact normal when that happens. With a cylinder, you don't get stuck on walls because the contact normal always corresponds to the direction the collision is in, so you "roll" over any microscopic surface anomalies, much like a 45 degree rotated AABB.

But OpenMW uses cylinders. It's in that quote you quoted: 

> One thing I've observed is that using a capsule shape for actors instead of cylinder helps with sliding, so it might be dependent on the collision detection algorithm."

The normals of the player object are irrelevant. All that matters is the normal of the wall. Simply trace a hull in the direction of the move vector. If it hits something, move the remainder of the move vector perpendicular to that normal. The shape of the player's hull only determines when it hits something, not how it moves after hitting something.

wareya:
If the shape of the hull never determines that direction you go after a contact, then the system is definitely doing something wrong. You don't want the normal as the direction of the collision or the normal of the surface region hit, you want the normal of the contact area, which depends on the features of both objects.

0tt:
playerhulls

wareya:
image

0tt:
You have "perpendicular" mixed up with "opposite" but thank you for pointing out this edge case.
How would this work in an algorithm?

wareya:
Sorry, I meant to write "contact plane" since that's what I had in the graphics, but I guess I forgot :P

>How would this work in an algorithm without changing the behavior of my cases and your first case?

I don't know; for the 2d case, it's a matter of preferring plane normals for the contact normal regardless of which object they come from. For 3d where you have edges and vertices and edge-edge collisions, it's more complicated, which is why so many games use AABBs.

FWIW, every 3d movement solver I've seen from goldsource to quakeworld to source engine to unreal to thief gold has edge cases when dealing with disoriented geometry.

scrawl:
I've given this PR some thorough testing, am very pleased with the results. On top of the performance improvements, sliding works much better than it used to, and I don't get stuck as much. Great work @logzero !

scrawl:
> One thing I've observed is that using a capsule shape for actors instead of cylinder helps with sliding, so it might be dependent on the collision detection algorithm.
> 
> Unfortunately capsules somehow break jumping in some cases. I don't know why yet.

FYI, I've moved to capsules here: https://github.com/OpenMW/openmw/commit/91939c4687da020d9c24ad723c680bcd2b72cea3

Noticed two issues so far but nothing deal breaking.
- Jumping while walking up a slope and holding Forward seems to propel you a little further than it should
- Some slopes are behaving a little glitchy (only noticed so far with the steps leading up to 'vivec, palace of vivec')

On the palace steps, turning around on a particular spot sometimes cause the character to switch between 'on ground' and 'in air' state every other frame for some reason. I don't know why yet.

scrawl:
> Some slopes are behaving a little glitchy (only noticed so far with the steps leading up to 'vivec, palace of vivec')
> On the palace steps, turning around on a particular spot sometimes cause the character to switch between 'on ground' and 'in air' state every other frame for some reason.

This one's fixed by upgrading to bullet 2.86. I presume the collision margin fixes for btCapsuleShape [here](https://github.com/bulletphysics/bullet3/commits/master/src/BulletCollision/CollisionShapes) were related.

PR #1159, opened 2016-12-14
More spelling fixes

lkrieg:
Fixed spelling mistakes in one string and multiple comments.


Allofich:
Nice work! One thing, though, grid->mPoints was correct and should not be changed to grid>mPoints. It means the variable mPoints of the pointer grid, not grid is greater than mPoints.

lkrieg:
Oops, thank you - that was unintentional. It's fixed it now.

PR #1160, opened 2016-12-15
[Do not merge yet] Added check before attempting to remove actor's collision object

Aussiemon:
Bullet 2.8.5 and Windows 8.1 x64:

Attempting to start debug builds of OpenMW fails on an assert at line 260 of [btCollisionWorld](https://github.com/bulletphysics/bullet3/blob/master/src/BulletCollision/CollisionDispatch/btCollisionWorld.cpp).

This is caused by the `updateCollisionMask()` method during actor initialization. The method attempts to remove the actor's collision object when the collision object does not yet exist in the world. 

Not only does this fail the assert in debug builds, it also causes the collision world to search for the actor's collision object manually.

### DIFF
--- a/apps/openmw/mwphysics/actor.cpp
+++ b/apps/openmw/mwphysics/actor.cpp
@@ -47,7 +47,7 @@ Actor::Actor(const MWWorld::Ptr& ptr, osg::ref_ptr<const Resource::BulletShape> 
     updateScale();
     updatePosition();
 
-    updateCollisionMask();
+    addCollisionMask(getCollisionMask());
 }
 
 Actor::~Actor()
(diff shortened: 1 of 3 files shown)


Aussiemon:
Encountered same issue with water collision object.

The failing CI checks seem to indicate that these issues won't be a problem until OpenMW is updated to support Bullet 2.8.5, so I've marked this as "Do not merge" for now.

scrawl:
We should not rely on Bullet implementation details like 'getWorldArrayIndex' either way. Could you fix this in a different way?

For example, updateCollisionMask() could use some helper functions like getCollisionMask() and addCollisionObject(mask) and then the actor constructor would just do addCollisionObject(getCollisionMask()) instead of updateCollisionMask().

With the waterCollisionObject I would just delete the collision object after it is removed, that way removing it twice would never be an issue.

Aussiemon:
That's a good idea. I'll try it that way.

Aussiemon:
Okay, Actor's constructor is now set up to only add the collision object initially. This should prevent the collision object from being removed before its addition, and the assert isn't failing now so far.

As for the water object, I added a reset() call to destruct the object and set the pointer to null. I set this within the check to see if water is disabled. If the water collision object exists, but water was disabled, the early return here could've led to an attempt to remove the object twice. This should be prevented now, and I couldn't get the assert to fail again.

I can't see any problems in my own release and debug builds after a bit of testing.

PR #1163, opened 2016-12-16
Add bounty when player follower commits murder (#2852)

lkrieg:
Fixes [#2852](https://bugs.openmw.org/issues/2852) by allowing player followers to commit crimes and adding the bounty to the player.

Tested by attacking a NPC (which adds 40 bounty), then summoning a Golden Saint and letting them fight to the death. When the NPC dies, a bounty of 1000 is added to the player (which makes a total bounty of 1040).

**UPDATE1**: If the player is invisible/hidden while a player follower commits murder, there is no bounty added. This differs from vanilla morrowind behaviour where there is always a bounty added even if the player is not in alarm range. (This happened because I originally only changed the actorKilled function to allow player followers and passed the player to commitCrime.)

**UPDATE2**: Changed commitCrime function to also allow player followers as offenders. Alarm radius calculations are now based on the offender, and the player receives the bounty. This means the player still receives a bounty even if he's hidden while the follower commits a crime.

**UPDATE3**: Oh dang it, I missed that this task was already assigned to Scrawl, sorry.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -990,18 +990,25 @@ namespace MWMechanics
         }
     }
 
-    bool MechanicsManager::commitCrime(const MWWorld::Ptr &player, const MWWorld::Ptr &victim, OffenseType type, int arg, bool victimAware)
+    bool MechanicsManager::commitCrime(const MWWorld::Ptr &offender, const MWWorld::Ptr &victim, OffenseType type, int arg, bool victimAware)
     {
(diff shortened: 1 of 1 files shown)


scrawl:
Seems to work for physical attacks but does not work with a summoned flame atronach, any idea?

lkrieg:
I would guess that the problem is somewhere in the CastSpell::inflict function in mwmechanics/spellcasting.cpp. I'll have a look and update this when finished.

lkrieg:
Looks like I also missed that NeveHanter has an open pull request on this. I really should check this better. How is the status on this pull request? I don't want to get into somebody's way.

PR #1165, opened 2016-12-17
Handle SDL event 0x304 by doing nothing (#3670)

lkrieg:
Fixes [#3670](https://bugs.openmw.org/issues/3670) by recognizing the SDL_KEYMAPCHANGED event without acting on it. Changing the keyboard layout while playing should mostly happen on accident and should (as far as I can tell) not be acted upon.

Tested by running openmw from the terminal and then changing the keyboard-layout via shortcut.
After tabbing back out of the game there will be no warning "Unhandled SDL event of type 0x304".


scrawl:
Layout changes shouldn't be an issue because we map controls to scancodes, not keycodes.

scrawl:
Oops, this needs to be guarded by a version check. I don't have SDL_KEYMAPCHANGED in SDL 2.0.3 so I'm guessing it was introduced with 2.0.4. Could you fix this and open a new PR.

PR #1168, opened 2016-12-18
ESSImporter: Import mark location

MiroslavR:
This implementation has one limitation -- a mark at the (0, 0) exterior cell won't be imported correctly. I couldn't find a good way of determining whether the mark cell belongs to an interior or an exterior worldspace. I use the fact that for interior cells the grid coordinate data is set to (0, 0).


Allofich:
Looks like you identified various unknown values, too!

PR #1170, opened 2016-12-19
Enchanted arrows don't explode when hitting the ground (Bug #3617)

NeveHanter:
Fixed bug https://bugs.openmw.org/issues/3617 by allowing touch and target enchantments from ranged weapons and their projectiles to explode even when colliding with non-activable objects, terrain, water slab or when shoot underwater.

Also allowed projectiles to fly through the dead bodies as in vanilla.


Capostrophic:
> Also allowed projectiles to fly through the dead bodies

Is replicating vanilla behavior actually worth it in this case, though? On a local community forum where I mentioned this pull request as "possible to be merged soon" callbacks were very... mixed.

> But why? 0_o

> I guess "just because".

> Will there be an option to disable that in future? I thought they don't focus on 100% replicating original behavior in those controversial cases, do they?

> Like, replicating this absolutely insignificant trash does not make sense considering [quoted Zini's OpenMW 1.0 ramble follows].

MiroslavR:
I don't see how this could possibly be a source of controversy, lmao. I guess anything is these days. And if it's insignificant, why even argue about it? I haven't seen a single argument against sticking to vanilla behavior. Generally, we stick to vanilla behavior unless there is a good reason not to.

scrawl:
Colliding with a dead actor wouldn't work very well because the collision shape used is intended for actors standing upright. If we had ragdoll collision then sure...

There are many more cases where at first glance, the behavior may not make sense but once you look closer there is a very good reason the original game did it this way. That's why we have a general policy of following vanilla closely, especially when it comes to things so insignificant they're not even worth arguing about.

MiroslavR:
Seems to work well. One difference I have noticed is that in vanilla MW, 0-area "On Target" arrows (e.g. "cruel frostarrow") do not create an explosion effect upon impact. Neither do 0-area "On Touch" arrows, but your implementation already seems to mimic that behavior.

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -406,29 +406,27 @@ namespace MWWorld
             // TODO: use a proper btRigidBody / btGhostObject?
             MWPhysics::PhysicsSystem::RayResult result = mPhysics->castRay(pos, newPos, caster, 0xff, MWPhysics::CollisionType_Projectile);
 
+            bool hitDeadActor = !result.mHitObject.isEmpty() && result.mHitObject.getClass().isActor() && result.mHitObject.getClass().getCreatureStats(result.mHitObject).isDead();
```
scrawl: The collision mask of the Actor already excludes CollisionType_Projectile when the actor is dead, see https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwphysics/actor.cpp#L88). I think you just need to change the mask for the castRay to CollisionType_Projectile instead of 0xff and all should work as intended.

scrawl: From a quick test everything seems to be working, so you can delete this change.

scrawl:
Any updates on this? I think all that's missing is to remove the redundant check for ```hitDeadActor``` as I explained.

NeveHanter:
@scrawl I've removed the check and everything works as expected. I've also squashed my branch for cleaner PR

PR #1171, opened 2016-12-20
Implemented "paying" for travelling followers

NeveHanter:
Player now pays for the following actors when travelling with the exception of the first follower who travels for free, by an addition replaced two versions of getFollowers by getActorsFollowing and getActorsSidingWith methods in MechanicsManager & Actor classes.

### DIFF
--- a/apps/openmw/mwgui/travelwindow.cpp
+++ b/apps/openmw/mwgui/travelwindow.cpp
@@ -74,27 +74,14 @@ namespace MWGui
             price = static_cast<int>(d / gmst.find("fTravelMult")->getFloat());
         }
 
-        // Add price for the followers in range
+        price = MWBase::Environment::get().getMechanicsManager()->getBarterOffer(mPtr, price, true);
+
+        // Add price for the travelling followers
         std::set<MWWorld::Ptr> followers;
-        MWBase::Environment::get().getMechanicsManager()->getActorsFollowing(player, followers);
-
-        int travellingFollowers = 0;
-        for(std::set<MWWorld::Ptr>::iterator it = followers.begin();it != followers.end();++it)
-        {
-            MWWorld::Ptr follower = *it;
-
-            std::string script = follower.getClass().getScript(follower);
-            if (!script.empty() && follower.getRefData().getLocals().getIntVar(script, "stayoutside") == 1)
-                continue;
-
-            if ((follower.getRefData().getPosition().asVec3() - player.getRefData().getPosition().asVec3()).length2() <= 800*800)
-                ++travellingFollowers;
-        }
+        MWWorld::ActionTeleport::getFollowersToTeleport(player, followers);
 
         // Apply followers cost, in vanilla one follower travels for free
-        price *= std::max(1, travellingFollowers);
-
-        price = MWBase::Environment::get().getMechanicsManager()->getBarterOffer(mPtr, price, true);
+        price *= std::max(1, static_cast<int>(followers.size()));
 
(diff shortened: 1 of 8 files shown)


apps/openmw/mwgui/travelwindow.cpp
```diff
@@ -70,14 +70,34 @@ namespace MWGui
...
+            std::string script = follower.getClass().getScript(follower);
+            if (!script.empty() && follower.getRefData().getLocals().getIntVar(script, "stayoutside") == 1)
+                continue;
+
+            if ((follower.getRefData().getPosition().asVec3() - player.getRefData().getPosition().asVec3()).length2() <= 800*800)
+                ++travellingFollowers;
+        }
+
+        // Apply followers cost, in vanilla one follower travels for free
+        price *= travellingFollowers;
```
MiroslavR: Are you sure _price_ should be 0 when the player travels alone? That seems wrong.

NeveHanter: Good catch, fixed, thanks!

apps/openmw/mwgui/travelwindow.cpp
```diff
@@ -70,14 +70,34 @@ namespace MWGui
...
+            std::string script = follower.getClass().getScript(follower);
+            if (!script.empty() && follower.getRefData().getLocals().getIntVar(script, "stayoutside") == 1)
+                continue;
+
+            if ((follower.getRefData().getPosition().asVec3() - player.getRefData().getPosition().asVec3()).length2() <= 800*800)
+                ++travellingFollowers;
+        }
+
+        // Apply followers cost, in vanilla one follower travels for free
+        price *= std::max(1, travellingFollowers);
```
MiroslavR: From my testing, it seems this line should be swapped with the getBarterOffer() line.

Gaerzi:
I feel like this should be subject to a GMST or some other way to make it optional since it diverges from vanilla behavior.

MiroslavR:
> it diverges from vanilla behavior.

What does? AFAIK, this is exactly how vanilla behaves. See https://wiki.openmw.org/index.php?title=Research:Trading_and_Services#Travel_costs.

apps/openmw/mwgui/travelwindow.cpp
```diff
@@ -70,14 +70,34 @@ namespace MWGui
...
+        {
+            MWWorld::Ptr follower = *it;
+
+            std::string script = follower.getClass().getScript(follower);
+            if (!script.empty() && follower.getRefData().getLocals().getIntVar(script, "stayoutside") == 1)
+                continue;
+
+            if ((follower.getRefData().getPosition().asVec3() - player.getRefData().getPosition().asVec3()).length2() <= 800*800)
+                ++travellingFollowers;
+        }
```
scrawl: Please reuse the code that is in ActionTeleport, e.g. by moving it to a function ActionTeleport::getFollowersToTeleport or something like that

jeffreyhaines:
I saw this https://bugs.openmw.org/issues/1501

Is there a ticket for this bug in the tracker?

apps/openmw/mwworld/actionteleport.hpp
```diff
@@ -23,9 +24,12 @@ namespace MWWorld
 
         public:
 
-            ActionTeleport (const std::string& cellName, const ESM::Position& position, bool teleportFollowers);
             ///< If cellName is empty, an exterior cell is assumed.
```
MiroslavR: Please remove the "<" as it's now showing in the generated docs.

MiroslavR: Please remove the "<" as it's now showing in the generated docs.

PR #1172, opened 2016-12-21
Add _USE_MATH_DEFINES define to make msys2/mingw happy,

logzero:
else the compiler will complain about undefined M_PI


scrawl:
Fixed in a slightly different way. M_PI is not part of the C or C++ standard so it's best to not even use it.

PR #1173, opened 2016-12-22
Clean-up saves by dropping references with invalid RefNums

MiroslavR:
Fixes: https://bugs.openmw.org/issues/1956
Relevant thread: https://forum.openmw.org/viewtopic.php?f=40&t=4016

I gave it some thought and couldn't come up with a situation where doing this would be harmful. I'm a bit skeptical, though, because of this comment on the bug report:
> With a changed RefNum, the connection to the object in the savegame is lost, so there's really nothing we can do.


scrawl:
Makes sense to me, @zinnschlag any objections?

PR #1174, opened 2016-12-23
Adjustments to AI combat engaging and disengaging

Allofich:
(Fixes [#3690](https://bugs.openmw.org/issues/3690))

This fixes this issue by making enemies start combat with player allies on sight.

I also did a general cleanup of the function. I avoided touching the part that is modified in another PR https://github.com/OpenMW/openmw/pull/921, so there shouldn't be any merge conflict there.

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -19,6 +19,7 @@
 #include "aicombataction.hpp"
 #include "combat.hpp"
 #include "coordinateconverter.hpp"
+#include "actorutil.hpp"
 
 namespace
 {
@@ -210,13 +211,14 @@ namespace MWMechanics
         else
         {
             timerReact = 0;
-            attack(actor, target, storage, characterController);
+            if (attack(actor, target, storage, characterController))
+                return true;
         }
 
         return false;
     }
 
-    void AiCombat::attack(const MWWorld::Ptr& actor, const MWWorld::Ptr& target, AiCombatStorage& storage, CharacterController& characterController)
+    bool AiCombat::attack(const MWWorld::Ptr& actor, const MWWorld::Ptr& target, AiCombatStorage& storage, CharacterController& characterController)
     {
         const MWWorld::CellStore*& currentCell = storage.mCell;
         bool cellChange = currentCell && (actor.getCell() != currentCell);
@@ -231,7 +233,10 @@ namespace MWMechanics
             storage.stopAttack();
             characterController.setAttackingOrSpell(false);
             storage.mActionCooldown = 0.f;
-            forceFlee = true;
+            if (target == MWMechanics::getPlayer())
+                forceFlee = true;
+            else
+                return true;
         }
 
         const MWWorld::Class& actorClass = actor.getClass();
@@ -243,7 +248,7 @@ namespace MWMechanics
         if (!forceFlee)
         {
             if (actionCooldown > 0)
-                return;
+                return false;
 
             if (characterController.readyToPrepareAttack())
             {
@@ -258,7 +263,7 @@ namespace MWMechanics
         }
 
         if (!currentAction)
-            return;
+            return false;
 
         if (storage.isFleeing() != currentAction->isFleeing())
         {
@@ -266,7 +271,7 @@ namespace MWMechanics
             {
                 storage.startFleeing();
                 MWBase::Environment::get().getDialogueManager()->say(actor, "flee");
-                return;
+                return false;
             }
             else
                 storage.stopFleeing();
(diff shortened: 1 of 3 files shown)


Allofich:
I want to look at this again, so please don't merge.

Allofich:
Thanks for the label. Hopefully it won't be too hard to adjust this in the way I'm thinking.

Allofich:
It's updated and a bit different now. The following observed behaviors from vanilla are recreated.

1. Enemies start combat with player followers or escorters. They have to be able to see the followers or escorters to do so. The aggression check acts as if the follower/escorter is the player, in that the player's persuasion still affects it.

2. Creatures with the "walks" movement flag can attack water creatures if they are within attack range. OpenMW was using an environment compatibility check, which returns false for a walking creature if the target is in the water. I changed it to use canFight(), which includes the environment check but returns true if the target is in range of an attack.

3. AI characters disengage combat if they cannot reach each other (canFight() is false), rather than fleeing as they may do when fighting the player. With the two above fixes but not this one, summoned creatures with the "walks" flag will tend to run away when combat starts with a water creature because canFight() will be false as long as the creature is out of range.

Allofich:
An addition: The AI will also target the player if it targets the player's follower or escorter.

Allofich:
Another addition: AI combat caused by follower/escorter relationships does not need LOS. The effects of this can be seen in morrowind.esm if you go to "Ald Daedroth: Antechamber." There are two NPCs out of LOS but with an AIFollow package on another NPC in a fight. In vanilla the 2 NPCs will go to help but in OpenMW they would not because they can't see the enemy of their follow target.

Allofich:
This should be it from me for a while, so it's OK to merge if you like.

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -231,7 +232,13 @@ namespace MWMechanics
             storage.stopAttack();
             characterController.setAttackingOrSpell(false);
             storage.mActionCooldown = 0.f;
-            forceFlee = true;
+            if (target == MWMechanics::getPlayer())
+                forceFlee = true;
+            else
+            {
+                CreatureStats& creatureStats = actor.getClass().getCreatureStats(actor);
+                creatureStats.getAiSequence().stopCombat();
```
MiroslavR: This sounds very dangerous to do while the package is still being processed. You should have AiPackage::execute() return true if you want it to be removed correctly.

Allofich: Fixed.

Allofich:
Fixed the issue noted by MiroslavR. The issues this PR set out to fix are fixed, but there is another issue. Currently AI followers/escorters will attack any enemies that enter combat with the player. In vanilla, they only do so if the enemy attacks you or them first, not just goes into combat.

This behavior is already present before this PR and could be a problem for escort quests since the escorted AI will try to fight all aggressive opponents which could get it killed. This PR makes enemies target followers/escorters, so the difference will show up even a bit more. Researching and implementing the combat engaging behavior will take some time, and I'm fine with it being a separate PR.

scrawl:
> Fixed the issue noted by MiroslavR. The issues this PR set out to fix are fixed, but there is another issue. Currently AI followers/escorters will attack any enemies that enter combat with the player or themselves. In vanilla, they only do so if the enemy attacks you or gets up next to the follower/escorter first (i.e. in attack range, probably), not just goes into combat.

AFAIR I coded this difference on purpose because I was so annoyed of my summons not helping in the fight right away. But you have a point there could be a problem with quest followers getting themselves killed even moreso than they do in vanilla. Feel free to go ahead and change the attack condition in a new PR.

Armada651:
> But you have a point there could be a problem with quest followers getting themselves killed even moreso than they do in vanilla. Feel free to go ahead and change the attack condition in a new PR.

@scrawl Yeah, I just ran into this problem where a follower just bolted away from me seeming randomly just to fight an enemy that didn't even have a line-of-sight.

Allofich:
I've done some work to recreate behavior more like vanilla, but it may be a while before it's ready to submit. I was thinking of trying to add an option to keep the behavior of followers attacking enemies without waiting for a hit to be exchanged first, but requiring line of sight.

scrawl:
As far as I recall, the vanilla condition is to attack as soon as the enemy has attacked the player.

How about extending this condition to also attack when the player attacks the enemy? That way followers won't get themselves killed on their own accord, but would still help in fights that the player chooses.

Allofich:
Actually vanilla works that way. A few days ago I was working on this and testing vanilla behavior. A follower will go into combat if either you or they are attacked, or if you attack an enemy, or (I think, it's a little hard to test this one) if an enemy is in range of their weapon attack. Hmm, maybe I can test that last one by giving a follower a ranged weapon.

But yeah, I assume Bethesda's design decision was to allow the player to decide when to engage enemies, with the follower not starting fights on their own unless an enemy attacked first. If the player's followers engage on sight with the enemy, it will be the same as the summons of AI characters, which don't require an attack attempt first. I was thinking of allowing this as an option.

PR #1175, opened 2016-12-23
Fix for "PC Rank" info condition

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3689

In vanilla MW, "PC Rank" condition uses the speaker's faction if "PC Faction" is not given.

TES Construction Set help actually says:
> PCRank
This will only give dialogue if the player is of a certain rank. If Faction is left empty, the player can be of this rank in any faction.

But that's not correct, as far as I can tell.


werdanith:
With regards to issue #3689 , this commit does the trick for me, I can't reproduce it anymore.

PR #1179, opened 2016-12-28
OpenAL_Output: When a source is finished, rewind it instead of stopping it.

Armada651:
By calling refillQueue() twice at the start of a stream we are throwing away the first few samples of the stream.

I blamed the issue to commit 83721092f2c6c11243587a54833ee35cd97dc235.

### DIFF
--- a/apps/openmw/mwsound/openal_output.cpp
+++ b/apps/openmw/mwsound/openal_output.cpp
@@ -437,7 +437,9 @@ bool OpenAL_SoundStream::process()
             alGetSourcei(mSource, AL_SOURCE_STATE, &state);
             if(state != AL_PLAYING && state != AL_PAUSED)
             {
+                // Ensure all processed buffers are removed so we don't replay them.
                 refillQueue();
+
                 alSourcePlay(mSource);
             }
         }
@@ -906,7 +908,10 @@ void OpenAL_Output::finishSound(MWBase::SoundPtr sound)
     ALuint source = GET_PTRID(sound->mHandle);
     sound->mHandle = 0;
 
-    alSourceStop(source);
+    // Rewind the stream instead of stopping it, this puts the source into an AL_INITIAL state,
+    // which works around a bug in the MacOS OpenAL implementation which would otherwise think
+    // the initial queue already played when it hasn't.
+    alSourceRewind(source);
     alSourcei(source, AL_BUFFER, 0);
 
     mFreeSources.push_back(source);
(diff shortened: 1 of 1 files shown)


MiroslavR:
@kcat, could you review this?

kcat:
I think this may be due to a bug in Apple's OpenAL. ```refillQueue``` shouldn't remove any unplayed samples since it only unqueues buffers that have been processed/played (if some buffers haven't played yet, there's nothing to unqueue). The reason for the second call is in case of an underrun during the first ```refillQueue``` call in the ```process``` method... the source may underrun after querying the processed buffer count and before queueing the next set of buffers, leaving the source in a stopped state with processed buffers at the start of the queue and the rest unprocessed. The second ```refillQueue``` call is intended to ensure all processed buffers are removed so the following ```alSourcePlay``` call doesn't cause it to replay them, which creates a smoother underrun recovery (it won't audibly "skip backward", and instead pick right back up where it left off).

What's probably happening is Apple's OpenAL naively returns all buffers as processed when a source is in an AL_STOPPED state, regardless of any new buffers that may have been queued since it stopped, causing the second ```refillQueue``` to erroneously clear the first set of buffers as its told they're all processed.

Perhaps a better workaround would be to change ```OpenAL_Output::finishSound``` and ```OpenAL_Output::finishStream``` to call ```alSourceRewind``` instead of ```alSourceStop```. That will put the source into an AL_INITIAL state (as when the source is first created), preventing the start of the stream from thinking the initial queue already played when it hasn't.

Armada651:
@kcat Sounds like a good workaround, we should definitely document the workaround and the second call to `refillQueue` though.

Armada651:
Updated the PR and removed the incorrect description.

I am doing a couple more tests, but it seems to work. What @kcat is saying also explains why it only happens after 30-40 seconds, because the bug would only be triggered when a voice file is played twice.

kcat:
> What @kcat is saying also explains why it only happens after 30-40 seconds, because the bug would only be triggered when a voice file is played twice.

More specifically, when a previously-used source ID is reused for a stream. When a source ID is used that was never used before, the issue wouldn't occur.

Armada651:
Ah, so every sound file gets a new source ID until all fee source IDs have been used up, then it will start reusing source IDs and the audio gets cut off. That makes sense.

MiroslavR:
Can we get another Mac user to test? @jeffreyhaines?

Armada651:
Might be good to mention that `alSourceRewind` also stops the source even if it is already playing. Thus it is simply an expanded version of `alSourceStop` that set the `AL_INITIAL` state.

scrawl:
Thanks for the detective work Armada651 & kcat !

I don't think we currently have anyone else building from master on OS X. Might as well merge now and see how things are in the next release.

Armada651:
It seems there is a rare occasion where this bug still happens. Maybe once every hour or so an NPC talks with the first part cutoff.

sbenner:
it's always the case for me on Sierra (10.12.3) openmw 0.41.0

Armada651:
@sbenner This fix isn't part of the 0.41.0 release.

PR #1181, opened 2017-01-03
Fix teleportation being unreachable

Assumeru:
#1180 made `else if(target.getClass().isActor() && target == getPlayer()`) unreachable by adding `else if(target.getClass().isActor())` above it.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -613,13 +613,10 @@ namespace MWMechanics
                 return true;
             }
         }
-        else if (target.getClass().isActor())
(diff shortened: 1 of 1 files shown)


MiroslavR:
Perhaps I'm nitpicking, but I think it would be more readable to unite those cases under a single `else if (target.getClass().isActor())`:

```c++
        else if (target.getClass().isActor())
        {
            if (effectId == ESM::MagicEffect::Dispel)
            {
                target.getClass().getCreatureStats(target).getActiveSpells().purgeAll(magnitude);
                return true;
            }

            if (target != getPlayer())
                return false;

            MWRender::Animation* anim = MWBase::Environment::get().getWorld()->getAnimation(mCaster);

            ...
        }         
```

Allofich:
Careless of me in #1180. Good catch.

PR #1184, opened 2017-01-06
Disable fixed time steps.

logzero:
[Do not merge]

My testing so far shows an overall more stable frame rate with higher min fps. It will cause a reduced max fps, once you go above 60, but that could be 'fixed' too by skipping steps.

From what I see nothing really requires fixed dt. Is that correct?

A proper fix would probably need a full movement solver rewrite, optimized for fixed dt, that maybe doesn't use tracing at all due to its high cpu costs.

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -1346,7 +1346,7 @@ namespace MWPhysics
         mMovementResults.clear();
 
         mTimeAccum += dt;
-        const float physicsDt = 1.f/60.0f;
+        const float physicsDt = dt;//1.f/60.0f;
 
         const int maxAllowedSteps = 20;
         int numSteps = mTimeAccum / (physicsDt);


kcat:
The fixed timesteps are to help keep the physics stable. Dynamic timesteps all too often cause oddities in the simulation when the user's framerate is fluctuating, especially if the framerate gets extremely high or extremely low.

Ultimately, I think the desire is to do fixed-step logic frames (physics, AI, etc), using interpolation so that the visual results can be as smooth as possible without extraneous logic processing. Bullet can interpolate between physics frames, but it relies on properly integrating our actor movement code into Bullet's pipeline. I tried to do that once, but I had no idea what I was doing and it resulted in really bad performance. I'm sure a better job could be done.

logzero:
> The fixed timesteps are to help keep the physics stable. Dynamic timesteps all too often cause oddities in the simulation when the user's framerate is fluctuating, especially if the framerate gets extremely high or extremely low.

Can you provide examples for this oddities in openmw?

From what I see the only effect they have is an increase of movement steps with dropping framerate (>60), making step length fps independent. But does it actually matter?

scrawl:
> Can you provide examples for this oddities in openmw?

You can search the bug tracker - there were many reports of players getting stuck in rocks, falling through vivec cantons, etc. before the framerate independence was added, though personally I was never able to reproduce such issues. Really depends on the framerate you get I suppose.

logzero:
> getting stuck in rocks, falling through vivec cantons
Thanks. I'll have a look. My current min fps are about 30. I'll try to do a few test runs with the cpu downclocked.

scrawl:
You probably know this but just for reference: http://bulletphysics.org/mediawiki-1.5.8/index.php/Framerate_Independence

> The benefit of framerate independence is that the simulation behaves more predictably. Many physics engines have trouble with Tunnelling when the step size is too large, and behave very unrealistically. Framerate independence keeps the step size at a value that is known to work well for the velocities and objects typically found in the simulation. Also, Determinism is probably impossible without framerate independence.

I think the convex casts in bullet use Continuous Collision Detection so I'm not sure actually why framerate would be an issue - might be an issue in our code rather than bullet.

Either way once we add fun stuff like ragdolls, constraints, and rigid bodies after version 1.0 we will definitely need the fixed framerate again.

logzero:
Yeah, convex cast calculates time of impact, one of the things that makes it slow. So theoretically it should just work.

It might be possible to use bullet dynamics for actor movement too, slide along contact surfaces. Stepping would need some extra logic though.

logzero:
I haven't been able to replicate the falling through vivec cantons issue.

Recently I've been playing around with a different approach, based on the back off idea. Instead of avoiding contacts one could just advance and resolve them. Stepping on obstacles is somewhat of an issue. My approach is to try to push upwards first when dealing low contacts (below max step height).

I'll create a new pull request if I get something that works.

Xandaros:
Are there any big issues with this? If it means I can actually play again, I might apply this when I play again.
Last time I played I had like 5 seconds per frame in some modded areas, but even tel mora sometimes went down to like 3 fps. It usually recovered within 20-30 seconds, but it was still half a minute of slide show...
Before the move to fixed time steps, I had no such issues and it is physics events causing it. (In particular, slave cages)

logzero:
@Xandaros It works for me. Although I've never ran into cases where the fps would drop that low.

I think omw will eventually want to use bullet physics which requires fixed steps anyway. The simpler back off approach should work better here (if I get it to work).

kcat:
> Are there any big issues with this?

Yes, the physics simulation becomes more unpredictable and it makes determinism impossible. It becomes harder to ensure correct physics are applied if the simulation can and will vary between frames, and no two systems will have matching timing.

logzero:
@kcat there is no physics simulation in omw currently.

kcat:
Eh? It's very simplistic, but physics simulation is how physics is applied. If there wasn't, using dynamic time steps instead of fixed would have no effect.

Xandaros:
Well, I said big problems. I am fine with physics not being deterministic if it means I can actually go to these areas.

PR #1186, opened 2017-01-08
Creatures now auto-equip shields (Fixes #3704)

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3704

Makes creatures auto-equip their shields according to item health. Also, adding/removing items from a creature's inventory will no longer call `autoEquip(...)`.

@Allofich My previous code had the item health comparison condition reversed. Can you please retest?

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -164,7 +164,7 @@ namespace MWClass
             getContainerStore(ptr).fill(ref->mBase->mInventory, ptr.getCellRef().getRefId());
 
             if (hasInventory)
-                getInventoryStore(ptr).autoEquip(ptr);   
+                getInventoryStore(ptr).autoEquipShield(ptr);
         }
     }
 
--- a/apps/openmw/mwworld/inventorystore.cpp
+++ b/apps/openmw/mwworld/inventorystore.cpp
@@ -139,7 +139,7 @@ MWWorld::ContainerStoreIterator MWWorld::InventoryStore::add(const Ptr& itemPtr,
 
     // Auto-equip items if an armor/clothing or weapon item is added, but not for the player nor werewolves
     if (actorPtr != MWMechanics::getPlayer()
-            && !(actorPtr.getClass().isNpc() && actorPtr.getClass().getNpcStats(actorPtr).isWerewolf()))
+            && actorPtr.getClass().isNpc() && !actorPtr.getClass().getNpcStats(actorPtr).isWerewolf())
     {
         std::string type = itemPtr.getTypeName();
         if (type == typeid(ESM::Armor).name() || type == typeid(ESM::Clothing).name())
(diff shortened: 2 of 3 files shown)


Allofich:
I've done some pretty thorough testing. It looks good! Your code is also the same as vanilla in that if there are multiple shields with the same health, the one that is listed first in the creature's inventory in the Construction Set (the one that was added first) will be chosen for equipping. There are a couple picky differences, though.

1. In vanilla, it is possible for the creature (I'm testing a skeleton) to switch to another shield after the first one is destroyed, but this doesn't seem to happen with your code.

2. In vanilla, as far as I can see NPCs do not equip items added to their inventory with additem, while in OpenMW they are, both before this PR's code and with this PR's code, they do.

Regarding 1, there is a brief moment between the switch of shields in vanilla. I don't know if that was specifically coded in or is just a pause before an update is called. Also, strangely, in my testing the skeleton will only switch to a daedric shield for some reason. I have a skeleton with a bunch of 4 health shields: an iron shield (listed first, so equipped), another iron shield, the daedric shield, a dwemer shield and a dreugh shield. Regardless of the order of the daedric shield it is chosen after the iron shield is destroyed, and without the daedric shield none of the others are chosen as a second shield, not even the second iron shield. No idea why, any ideas? But anyway, since it's possible for the creature to re-equip, how about calling the equip function again if the shield is destroyed? If you are up for it, when this happens, I think a brief pause like vanilla has would look better than instantly switching the shields.

Allofich:
Looks like the skeleton will equip an ebony shield as a replacement after losing the first one. Maybe it just has to have a high enough armor rating?

MiroslavR:
Sorry, I'm a bit lost. So you're saying that auto-equipping during the initial creation of a creature is working as expected? Because that is really the focus of this PR.

As for replacing broken shields, my theory is that some other function than `autoEquipShield(...)` is called there, but I haven't confirmed that yet. If there is a timer, it could even be part of the AI controller. Have you tried disabling the AI before disintegrating their shield?

Allofich:
Yes, sorry for the disorganization. The PR works for solving the bug. I agree, replacing the broken shield seems like it could be a different function. I haven't looked at other potential instances of replacing broken equipment and don't know what the state of OpenMW is there, but it could be worked on later.

Allofich:
> Have you tried disabling the AI before disintegrating their shield?

I tried now. I toggled off AI, then used placeatpc to spawn the skeleton, and it had it's initial shield equipped. I then hit on it over and over until its shield broke, and it did not equip a new shield. When I then toggled AI on, it equipped its weapon and a new shield.

Allofich:
So, just to be clear,

> So you're saying that auto-equipping during the initial creation of a creature is working as expected? 

Yes.

MiroslavR:
OK, thanks for testing. :+1: 

> Same thing if I spawned it with AI on and let it go into combat mode and draw its weapon first. I then toggled AI off and hit it until the initial shield broke, and it didn't equip a new one. After toggling AI on again, it equipped the new shield.

I think that should be handled separately (perhaps when we reimplement vanilla decision making). I wonder if replacing shields is part of AiCombat. That could be tested with setting yourself invisible (`player->setinvisible 1`) instead of disabling AI.

Allofich:
> That could be tested with setting yourself invisible (player->setinvisible 1) instead of disabling AI.

I'm not sure what you mean, even if I am invisible if I hit the skeleton it sends it into combat mode (although it starts fleeing). I don't know how to damage its shield other than hitting it. I tried now and it did equip the replacement shield. It did so after arriving at a flee pathgrid point destination and then turning to face in my direction again. During the interval between when the first shield broke and it arrived at the pathgrid point, it was running without any shield.

It's hard to describe, so I made a video. I managed to get a rather extreme example of the skeleton taking a long time between losing the shield and re-equipping, usually it's much shorter.
[shield re-equip.zip](https://github.com/OpenMW/openmw/files/694207/shield.re-equip.zip)

Allofich:
I tried the new test you suggested, and the skeleton didn't equip the second shield. I then hit it to put it into combat mode and it equipped the second shield.

PR #1187, opened 2017-01-08
Added fonts and paths documents.

rhtucker:
Went ahead and ported fonts and paths. We want paths within RTD so if we change them the versioning properly shows where they are.


scrawl:
Did you forget to add fonts to the index?

PR #1191, opened 2017-01-11
Stop combat when adding an AI package to an actor

Allofich:
Fixes https://bugs.openmw.org/issues/3722.

Not only Command spells, but giving an actor a package like AIWander or AIFollow through the console should cancel their combat.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -93,6 +93,7 @@ public:
     }
 };
 
+// Check for command effects having ended and remove package if necessary
 void adjustCommandedActor (const MWWorld::Ptr& actor)
 {
     CheckActorCommanded check(actor);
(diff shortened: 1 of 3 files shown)


Allofich:
Whoops, this will stop combat when AI tries to avoid a door.

Allofich:
Avoiding doors, the "face" instruction, and crime pursuit should all not remove combat, so `isActualAiPackage` should be sufficient here I think.

Allofich:
I found another issue. Currently the code checks periodically checks for whether an actor is commanded, and stacks a command (follow) package if the actor is commanded but doesn't have such a package. But for ending combat, this is insufficient, because combat should end every time a command spell is successfully cast on an actor, even if a previous command effect is still active.

For example, if you command an actor, then make them startcombat with someone, then hit them with another command spell, it should take them out of combat to be like vanilla. To recreate this, I changed the code to add the command package (and stop combat) in `inflict()` in `spellcasting.cpp.` `adjustCommandedActor()` in `actors.cpp` now just removes the package when the effect is over.

I think maybe the package removal should be redone, too, because currently there can be a slight delay where the actor will continue to follow you for a split-second if you rest to advance time. The package removal should ideally be more immediate.

Allofich:
Updated with some checking on the spell caster and to allow AI actors to cast command on each other, as they can in vanilla if you use the "cast" command.

scrawl:
<pre>
apps/openmw/mwmechanics/spellcasting.cpp:495:87: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
                         if ((effectIt->mEffectID == ESM::MagicEffect::CommandHumanoid && target.getClass().isNpc()
</pre>

Allofich:
Warning fixed.

PR #654, opened 2015-07-07
C++11

zinnschlag:
Let's see if travis is okay with it.

PR #1195, opened 2017-01-26
added ability to reload shaders from game console via command "reloadshaders"

allcreater:
With possibility to reload shaders modern graphics development will be easy because no need to reload the game after changing something at shaders. But I'm not sure that my adding of a new command (which, as I understand, will be supported at scripts, and it's definitely excessively ) is good idea, so maybe there is better way to implement it :)

Discussion on forum: https://forum.openmw.org/viewtopic.php?f=3&t=4076&sid=f3d7b4a496155034985d55ccf6633b39#p45376

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -567,6 +567,9 @@ namespace MWBase
 
             /// Return physical or rendering half extents of the given actor.
             virtual osg::Vec3f getHalfExtents(const MWWorld::ConstPtr& actor, bool rendering=false) const = 0;
+
+            /// Trying to reload all loaded shaders, could be very useful for developlent purposes
+            virtual void reloadShaders() = 0;
     };
 }
 
--- a/apps/openmw/mwscript/miscextensions.cpp
+++ b/apps/openmw/mwscript/miscextensions.cpp
@@ -1251,6 +1251,18 @@ namespace MWScript
                 MWBase::Environment::get().getWorld()->createOverrideRecord(listCopy);
             }
         };
+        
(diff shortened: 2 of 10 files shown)


scrawl:
I'm afraid we won't accept any new script command until after 1.0. A hotkey (e.g. F4) would be possible. I am not sure though of a clean way to expose the shader backend to the input manager.

Personally I don't see much of a need for the reload function, as for me just restarting the game takes less than 2 seconds. You may be interested in the ```--skip-menu``` and/or ```--load-savegame``` commandline arguments to skip clicking through menus.

Btw, we can't use auto/C++11 yet (see #654 ).

allcreater:
@scrawl , thank you very much for response! 
It seems I was hurried to made a pull request, because I already was warned at forum, that scripts extension is not planned at nearest time =(

Sorry about c++11. I hope it will be possible to use it in future, but it seems the project have some reasons to use elder standard. Seems it's better for me to read that topic on forum :) 

How do you think, is it will be ok if we will find another way to call the reloading method and remove C++ 11? I have many plans about advanced graphics (but of course firstly will try to discuss it at forum), so it's very useful for me.

Maybe I do something wrong, but for me debug version of game is loading long, something like a minute :( And of course I use --skip-menu option :)

I will do my best to make something useful for a beloved game :)

scrawl:
Are you on windows? Debug builds on windows are known to be very slow. On Linux my debug build starts in less than 2 seconds.

Anyway let's do it with a hotkey then. You should be able to get the shadermanager from engine.cpp as follows ```mResourceSystem->getSceneManager()->getShaderManager()``` and pass that to the constructor of InputManager.

PR #1196, opened 2017-01-27
Fix MSVC warnings

Allofich:
Not a lot going on right now, so here's a few fixes for warnings I see when compiling with Visual Studio 2015.

Warnings fixed are:

openmw\extern\oics\tinyxml.cpp(51): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data

openmw\components\terrain\buffercache.cpp(26): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
openmw\components\terrain\buffercache.cpp(223): note: see reference to function template instantiation 'osg::ref_ptr<osg::DrawElementsUShort> `anonymous-namespace'::createIndexBuffer<osg::DrawElementsUShort>(unsigned int,unsigned int)' being compiled

openmw\components\terrain\buffercache.cpp(78): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)

openmw\components\esmterrain\storage.cpp(158): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)

storage.cpp.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library

### DIFF
--- a/components/esmterrain/storage.cpp
+++ b/components/esmterrain/storage.cpp
@@ -155,7 +155,7 @@ namespace ESMTerrain
                                             osg::ref_ptr<osg::Vec4Array> colours)
     {
         // LOD level n means every 2^n-th vertex is kept
-        size_t increment = 1 << lodLevel;
+        size_t increment = static_cast<size_t>(1) << lodLevel;
 
         osg::Vec2f origin = center - osg::Vec2f(size/2.f, size/2.f);
 
--- a/components/terrain/buffercache.cpp
+++ b/components/terrain/buffercache.cpp
@@ -23,7 +23,7 @@ osg::ref_ptr<IndexArrayType> createIndexBuffer(unsigned int flags, unsigned int 
 
     bool anyDeltas = (lodDeltas[Terrain::North] || lodDeltas[Terrain::South] || lodDeltas[Terrain::West] || lodDeltas[Terrain::East]);
 
-    size_t increment = 1 << lodLevel;
+    size_t increment = static_cast<size_t>(1) << lodLevel;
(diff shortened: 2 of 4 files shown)


components/esmterrain/storage.cpp
```diff
@@ -155,7 +155,7 @@ namespace ESMTerrain
                                             osg::ref_ptr<osg::Vec4Array> colours)
     {
         // LOD level n means every 2^n-th vertex is kept
-        size_t increment = 1 << lodLevel;
+        size_t increment = static_cast<int64_t>(1) << lodLevel;
```
scrawl: int64_t is not the same as size_t. Think about 32 bit systems.

Wouldn't this work?
<pre>
static_cast<size_t>(1) << lodLevel;
</pre>

Allofich: That won't work because it wants a type specifier within <>.

scrawl: size_t is a type specifier. I think there was a problem with the markdown formatting before, could you look at my edited comment?

Allofich: Do you know if casting it to <unsigned long long> would be okay?

Allofich: Oh it wasn't showing your new or edited comment before. I'll try that now.

Allofich: Seems to satisfy the compiler and no obvious problem in-game.

PR #1198, opened 2017-02-01
Make combat engagement logic more like vanilla

Allofich:
(Fixes #2678, Fixes #3705)

This is my attempt to make the combat engagement AI similar to vanilla behavior, based on observations. Much of it involves how AI characters allied to each other through Follow or Escort packages engage in combat.

I tried to recreate the following:
1) Enemies will stop combat with the player or a player ally if no hit attempt (either from the player/player ally or from the enemy) has occurred yet, and they can't reach the player/player ally any more (such as from the player leaving the water). This fixes the "water creatures won't stop combat" bug.

2) AI that is allied with other AI or with the player through Follow or Escort packages will start combat with anyone who exchanges a hit attempt with one of their allies, regardless of LOS.

3) AI that is allied with other AI will go into combat with anyone one of their allies is in combat with even before a hit attempt, but in this case they have to be aware of them (LOS, etc.)

4) If AI that is engaged in combat by two or more enemies who are allied with each other, it will engage in combat back with them regardless of LOS. This doesn't apply for player allies, though, they wait for an attack to be exchanged first.

5) Guards will fight against any creatures in combat, except for ones that are running follow or escort packages.

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -325,22 +325,38 @@ namespace MWClass
 
     void Creature::onHit(const MWWorld::Ptr &ptr, float damage, bool ishealth, const MWWorld::Ptr &object, const MWWorld::Ptr &attacker, const osg::Vec3f &hitPosition, bool successful) const
     {
-        // NOTE: 'object' and/or 'attacker' may be empty.
+        MWMechanics::CreatureStats& stats = ptr.getClass().getCreatureStats(ptr);
 
-        if (!attacker.isEmpty() && !ptr.getClass().getCreatureStats(ptr).getAiSequence().isInCombat(attacker))
-            getCreatureStats(ptr).setAttacked(true);
+        // NOTE: 'object' and/or 'attacker' may be empty.        
+        if (!attacker.isEmpty() && !stats.getAiSequence().isInCombat(attacker))
+            stats.setAttacked(true);
 
         // Self defense
         bool setOnPcHitMe = true; // Note OnPcHitMe is not set for friendly hits.
         
         // No retaliation for totally static creatures (they have no movement or attacks anyway)
         if (isMobile(ptr) && !attacker.isEmpty())
-        {
             setOnPcHitMe = MWBase::Environment::get().getMechanicsManager()->actorAttacked(ptr, attacker);
+
+        // Attacker and target store each other as hitattemptactor if they have no one stored yet
+        if (!attacker.isEmpty() && !ptr.isEmpty())
+        {
+            MWMechanics::CreatureStats& statsAttacker = attacker.getClass().getCreatureStats(attacker);
+            // First handle the attacked actor
+            if (stats.getHitAttemptActor().isEmpty()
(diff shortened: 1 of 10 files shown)


apps/openmw/engine.cpp
```diff
@@ -140,6 +138,7 @@ void OMW::Engine::frame(float frametime)
             mEnvironment.getMechanicsManager()->update(frametime,
                 guiActive);
         }
+        mEnvironment.getWorld()->markCellAsUnchanged();
```
scrawl: Why did you move markCellAsUnchanged() ?

scrawl:
Do we need to store mHitAttemptActor in save games? If so, you could use the actorId.

apps/openmw/engine.cpp
Allofich: I did that because I wanted to clear the player's `mHitAttemptActor` (I've now renamed it `mHitAttemptActorId)` when changing cells. Otherwise a follower in the new cell could still target an enemy from an earlier cell that attacked the player. When I checked `MWBase::Environment::get().getWorld()->hasCellChanged(),` though, it was always returning false, and when I traced it back to engine.cpp, I saw that `markCellAsUnchanged()` was being called before actors were updated, so that's why I reordered it.

Allofich:
The original engine does not seem to save whatever its equivalent to mHitAttemptActor is. To verify:

1. Give yourself a summon spell and get hit by an enemy
2. Distance yourself somewhat from the enemy
3. If you cast the summon spell at this point, the summon will go and attack the enemy that hit you. But if you save your game, load the game and then cast, they won't.

It's best if a saved game recreate the game state as exactly as possible, though, so I added it in, using the actorId like you suggested.

apps/openmw/engine.cpp
scrawl: Okay, but I don't see hasCellChanged used anywhere in these changes? If you scrapped the idea, the markCellAsUnchanged move should be undone too. That flag is used by scripts so we'd need to do a careful review nothing was broken.

Allofich: I used it in actors.cpp.

bool cellChanged = MWBase::Environment::get().getWorld()->hasCellChanged();
 
if (cellChanged)
          player.getClass().getCreatureStats(player).setHitAttemptActorId(-1);

Maybe another way to do it is to just check if the actor is located in a different cell, if there is a function handy for that. I'll take a look at it again later.

Allofich: I went ahead and looked at it now, and used `isInCell()` instead, which seems to work. I undid the change to engine.cpp. I pushed it to the branch but GitHub seems to be slow to update right now.

scrawl:
I decided to disable the saving code for now so we don't break save games for anyone who may be switching back and forth between stable and master. We can enable it the next time another change is made to save games or for the next release.

Allofich:
> I decided to disable the saving code for now so we don't break save games for anyone who may be switching back and forth between stable and master. We can enable it the next time another change is made to save games or for the next release.

That's fine with me.

PR #1200, opened 2017-02-11
Fix trying to access stats on non-actors

Allofich:
There were a few spots where framelistener errors would happen with traps due to trying to access creaturestats on a non-actor. Part of this (when the player activates a trap) was due to my recent PR, but there was also a pre-existing problem when AI activated traps. You can try the attached save. An AI actor will use a trapped chest right after it is loaded. Without this PR it may crash, and lower or turn off your sound volume first. With this PR it should be fine.

This also fixes AI actors having the attacked flag set when hit by a trap. They shouldn't, according to a test I did with the original engine. After using the trap the NPC should still give his normal greeting when you activate him instead of the "go away" one that happens if he has been attacked.

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -328,7 +328,7 @@ namespace MWClass
         MWMechanics::CreatureStats& stats = ptr.getClass().getCreatureStats(ptr);
 
         // NOTE: 'object' and/or 'attacker' may be empty.        
-        if (!attacker.isEmpty() && !stats.getAiSequence().isInCombat(attacker))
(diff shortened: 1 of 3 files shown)


Allofich:
[trap_test.zip](https://github.com/OpenMW/openmw/files/768349/trap_test.zip)

scrawl:
I'm not seeing a crash, just the game freezing however I can still bring up the menu and load another game or quit. If there was an actual crash I'd be interested in the stack trace.

Allofich:
OK, it did crash once for me but after that it didn't, so maybe the crash was caused by something else.

PR #1201, opened 2017-02-11
Prevent AI actors from hitting unintended targets

Allofich:
Fixes https://bugs.openmw.org/issues/3254.

This is not a 100% complete fix though. AI doing melee attacks that don't hit the intended target will do nothing, rather than passing through the unintended target like the original engine. For touch spells I was able to address this by casting a new ray from the point where the attempt hit an unintended target, but for melee attacks it wasn't clear to me how to do this.

Regardless I still think it's better than the AI whacking their allies in the back all the time so I am submitting this for merge. As far as I know all other types of unintended hit work the same as in the original engine. These are:

On-touch spell hits unintended target
Spell projectile hits unintended target
Non-spell projectile hits unintended target

In all cases, the attack acts as if the unintended target was not there and passes through it.

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -7,10 +7,11 @@
 #include <components/esm/esmwriter.hpp>
 #include <components/esm/cellid.hpp>
 
+#include <components/misc/resourcehelpers.hpp>
 #include <components/misc/rng.hpp>
 
 #include <components/files/collections.hpp>
-#include <components/misc/resourcehelpers.hpp>
+
 #include <components/resource/resourcesystem.hpp>
 
 #include <components/sceneutil/positionattitudetransform.hpp>
@@ -1064,7 +1065,7 @@ namespace MWWorld
         return osg::Matrixf::translate(actor.getRefData().getPosition().asVec3());
     }
 
-    std::pair<MWWorld::Ptr,osg::Vec3f> World::getHitContact(const MWWorld::ConstPtr &ptr, float distance)
+    std::pair<MWWorld::Ptr,osg::Vec3f> World::getHitContact(const MWWorld::ConstPtr &ptr, float distance, std::vector<MWWorld::Ptr> &targets)
     {
         const ESM::Position &posdata = ptr.getRefData().getPosition();
 
@@ -1084,7 +1085,7 @@ namespace MWWorld
             distance += halfExtents.y();
         }
 
-        std::pair<MWWorld::Ptr,osg::Vec3f> result = mPhysics->getHitContact(ptr, pos, rot, distance);
+        std::pair<MWWorld::Ptr,osg::Vec3f> result = mPhysics->getHitContact(ptr, pos, rot, distance, targets);
         if(result.first.isEmpty())
             return std::make_pair(MWWorld::Ptr(), osg::Vec3f());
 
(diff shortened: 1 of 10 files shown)


Allofich:
I just realized this fix won't work for targeted spells cast through the "cast" command. Please don't merge.

Allofich:
Fixed the above issue for projectiles cast from objects. This does not work for projectiles cast from actors with the "cast" command, which will erroneously pass through the target right now because the actor is not in combat with the target. However the cast command when used with actors needs to be worked on anyway, as right now they just shoot in the direction they are facing. They should instead aim at the specified target, and then then this target should be accessible for using in this check.

Projectiles shot from non-actors hit whatever they come into contact with, which is the same as the original engine.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -2681,18 +2684,30 @@ namespace MWWorld
...
+            if (!Settings::Manager::getBool("allow AI to hit unintended targets", "Game"))
+            {
+                // Don't let AI actors hit unintended targets. If an unintended target is hit, continue casting the ray past them
+                while (actor != MWMechanics::getPlayer() && !target.isEmpty() && target.getClass().isActor() && !stats.getAiSequence().isInCombat(target))
+                {
+                    result1 = mPhysics->castRay(result1.mHitPos, dest, target, MWPhysics::CollisionType_Actor);
+                    if (result1.mHit)
+                        target = result1.mHitObject;
+                    else
+                        target = NULL;
```
scrawl: I wouldn't rely on a cast from the hit point to work. How do you know you're not just going to hit the same shape again? Worst case this would be an infinite loop.

Instead, I would use a single cast and filter unintended targets via the ResultCallback.


I'd remove the setting for now; really no reason to enable at the moment with the AI being as bad as it is.

Allofich: I thought I had accounted for the infinite loop (I actually wrote this a while ago, not today), but maybe not.

> I would use a single cast and filter unintended targets via the ResultCallback.

I'm not quite sure what that involves at the moment but I'll see if I can figure it out.

> I'd remove the setting for now; really no reason to enable at the moment with the AI being as bad as it is.

I thought people might want it anyway but I'll go ahead and remove it then. I agree it's really a problem at the moment and I'm kind of surprised I haven't seen it brought up more.

Allofich:
New version mostly works, including allowing melee hits to go pass through unintended targets.

Unfortunately, there is a crash when a projectile sent by an AI actor in combat hits the terrain. I think the problem is in addSingleResult() in ClosestNotMeRayResultCallback, in

            `if (mTarget && (mTarget != rayResult.m_collisionObject))
            {
                PtrHolder* holder = static_cast<PtrHolder*>(rayResult.m_collisionObject->getUserPointer());
                if (!holder->getPtr().isEmpty() && holder->getPtr().getClass().isActor())
                    return 1.f;
            }`

but I'm not sure what it is. isEmpty() is used elsewhere to check whether a pointer is terrain, so I think that's OK, but the crash seems to happen around there. Maybe I can't use getUserPointer() like this?

scrawl:
That is missing a check for ```holder == NULL``` as far as I can tell.

Allofich:
So simple, but that seems to have been it. Fixed.

Also the problem of projectiles cast from actors using the "cast" script command not hitting targets is no longer happening (as long as the actor is not in combat)t.

One more thing is that I found that in the original engine the AI will hit any actors it is in combat with when it shoots a spell. So if it is in combat with both the player and an NPC and you have it cast a projectile at the player but it contacts the NPC, it will hit the NPC. I'll try to add this in.

Allofich:
New version up. AI attacks now hit any actors they are in combat with.

PR #1202, opened 2017-02-12
Make Book records display skill IDs instead of attribute IDs for teached skills. (Fixes Bug #3746.)

MAtahualpa:
Makes Book records display skill IDs instead of attribute IDs for teached skills. Please note that everything else already was in place -- the member variable mSkillID, the getData function, and the setData function. The verifier does not check the entry in the first place.

Related issue:
- Bug #3746: OpenMW-CS: Book records show attribute IDs instead of skill IDs for teached skills entry. (https://bugs.openmw.org/issues/3746)

Tests:
I tested the changes by creating a new .omwaddon file and changing some of the skill books' teached skills (including one which I changed to "none") in OpenMW-CS. Loading the add-on in OpenMW resulted in the correct skills increasing.

### DIFF
--- a/apps/esmtool/record.cpp
+++ b/apps/esmtool/record.cpp
@@ -494,7 +494,7 @@ void Record<ESM::Book>::print()
     std::cout << "  Weight: " << mData.mData.mWeight << std::endl;
     std::cout << "  Value: " << mData.mData.mValue << std::endl;
     std::cout << "  IsScroll: " << mData.mData.mIsScroll << std::endl;
-    std::cout << "  SkillID: " << mData.mData.mSkillID << std::endl;
+    std::cout << "  SkillId: " << mData.mData.mSkillId << std::endl;
     std::cout << "  Enchantment Points: " << mData.mData.mEnchant << std::endl;
     if (mPrintPlain)
     {
--- a/apps/essimporter/converter.hpp
+++ b/apps/essimporter/converter.hpp
@@ -202,7 +202,7 @@ public:
         bool isDeleted = false;
 
         book.load(esm, isDeleted);
-        if (book.mData.mSkillID == -1)
+        if (book.mData.mSkillId == -1)
             mContext->mPlayer.mObject.mNpcStats.mUsedIds.push_back(Misc::StringUtils::lowerCase(book.mId));
 
         mRecords[book.mId] = book;
--- a/apps/opencs/model/world/refidadapterimp.cpp
+++ b/apps/opencs/model/world/refidadapterimp.cpp
@@ -316,7 +316,7 @@ QVariant CSMWorld::BookRefIdAdapter::getData (const RefIdColumn *column,
         return record.get().mData.mIsScroll!=0;
 
     if (column==mSkill)
-        return record.get().mData.mSkillID;
+        return record.get().mData.mSkillId;
 
     if (column==mText)
         return QString::fromUtf8 (record.get().mText.c_str());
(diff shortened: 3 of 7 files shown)


MAtahualpa:
Ah, I see. Pushing more commits to the same branch automatically adds them to the open pull request. Should have thought of this before. Sorry for the inconvenience.

Anyway, the first commit regarding "refidcollection.cpp" is for the actual issue (Bug #3746), while the other five commits are pure cosmetics (changing Book records' member variable "mSkillID" to "mSkillId").

Allofich:
Don't worry about it, I had trouble with Git and GitHub for a few days at first, and sometimes still do.

I guess from the "Update --" messages that you are making changes through GitHub? I think that's the default commit message it gives. I'm not sure if it can be done through GitHub but if you use the command line program GitBash, you can "squash" your commits together, so you could make all of these into one commit, or two commits, or whatever.

MiroslavR:
```
/home/travis/build/OpenMW/openmw/apps/openmw/mwworld/actionread.cpp: In member function ‘virtual void MWWorld::ActionRead::executeImp(const MWWorld::Ptr&)’:
/home/travis/build/OpenMW/openmw/apps/openmw/mwworld/actionread.cpp:47:66: error: ‘const struct ESM::Book::BKDTstruct’ has no member named ‘mSkillID’
         if (ref->mBase->mData.mSkillId >= 0 && ref->mBase->mData.mSkillID < ESM::Skill::Length
```

You forgot to rename mSkillID here... BTW, I am strongly against these cosmetic changes. They create the potential for merge conflicts for absolutely no gain.

Allofich:
Yeah, looks like you renamed the later references to mSkillID, but you did not change the name of the "original" one.

If you just made the changes through GitHub, it's a good idea to do your changes in a coding program so you can make sure it compiles before you submit a pull request.

MAtahualpa:
> [...] it's a good idea to do your changes in a coding program so you can make sure it compiles without errors before you submit a pull request.

I did that actually: I'm using MS Visual Studio 2015 for testing my changes and, then, apply them on GitHub. I did the latter one manually because that's the way GitHub initially offered me. Anyway, I forgot the second "mSkillID" in "actionread.cpp", line 47. Fixed now.

Allofich:
> Does the chronological order of commits matter for the checks? I changed "loadbook.hpp" at the end -- however, mSkillId is declared there.

The checks are done using all of the commits put together, so it doesn't matter.

Allofich:
Looks like there's another one to change.

I'm also using Visual Studio 2015. If you don't know about it, one feature that I use very frequently is "Find All References." Right click on something you want to find, like mSkillID, and select "Find All References." It will show you everywhere that "mSkillID" appears in the solution file.

MAtahualpa:
Yeah, I frequently use the "Find All References" option, paired with a plain search in case of doubt. However -- like you said -- the record.cpp file isn't in the VS build. (Why?!)

I have edited the corresponding line; hopefully, Travis won't complain this time. :-/
(I'll be more careful with cosmetic changes in the future -- but this one seemed so easy. Yes...)

zinnschlag:
You want to make a separate feature branch for every set of changes (usually branching of from master). That avoids the problem we are seeing here. But never mind. The one time glitch is not an issue. Reviewing ...

zinnschlag:
Also, you don't want to do a commit for every file change. We are not too strict about that, but generally when you do a commit the code should be in a state that it can be build.
Merging ...

MAtahualpa:
Okay. I'll try to do better the next time. Thanks for your help, guys!

PR #1204, opened 2017-02-13
Optional breathe dependency

rhtucker:
I've included an extra conf.py file to remove the breathe dependency. When building with sphinx, use -c "path-to-/nobreathe". A couple other formatting fixes.


scrawl:
Did you delete font.rst on purpose?

rhtucker:
Yes and no, it should have been changed to fonts instead of font. It got the deletion but didn't make the addition of the new file. I've fixed that and the optional conf.py that apparently wasn't showing up either. I've been fiddling around with PyCharm and learning it's version control integration, so that's probably why. We also need to make sure once this is merged that the normal conf.py still works. I deleted a few different paths and things that were causing errors with Sphinx, but I tested it with the nobreathe version because I don't want to deal with doxygen if I don't have to.

psi29a:
I like the motivation, but perhaps we an make this a lot simpler. Hold off on this commit for now. :)

PR #1207, opened 2017-02-14
Replace "Scroll" check box with a combo box (fixes #3748)

MAtahualpa:
Replaces the "Scroll" check box in Book records with a "Book Type" combo box.

Related issue:
- Fixes #3748: OpenMW-CS: Replace "Scroll" check box in Book records with "Book Type" combo box. (https://bugs.openmw.org/issues/3748)

Tests:
The changes were successfully tested in OpenMW-CS by manipulating several Book records. Please note that the actual logic behind this entry is not implemented yet: Books which are of type "Scroll" can have an enchantment attached, normal books ("Book") cannot.

### DIFF
--- a/apps/opencs/model/world/refidadapterimp.cpp
+++ b/apps/opencs/model/world/refidadapterimp.cpp
@@ -313,7 +313,7 @@ QVariant CSMWorld::BookRefIdAdapter::getData (const RefIdColumn *column,
         data.getRecord (RefIdData::LocalIndex (index, UniversalId::Type_Book)));
 
     if (column==mBookType)
-        return record.get().mData.mIsScroll!=0;
+        return record.get().mData.mIsScroll;
 
     if (column==mSkill)
         return record.get().mData.mSkillId;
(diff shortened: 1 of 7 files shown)


MAtahualpa:
I somehow managed to shift from web-based editing to local editing via a remote repository -- only 1 commit this time. :-)
Hopefully, I didn't screw up during the process of pushing my changes from the local branch.

apps/opencs/model/world/refidadapterimp.cpp
```diff
@@ -312,7 +312,7 @@ QVariant CSMWorld::BookRefIdAdapter::getData (const RefIdColumn *column,
     const Record<ESM::Book>& record = static_cast<const Record<ESM::Book>&> (
         data.getRecord (RefIdData::LocalIndex (index, UniversalId::Type_Book)));
 
-    if (column==mScroll)
+    if (column==mBookType)
         return record.get().mData.mIsScroll!=0;
```
MAtahualpa: Yes, I wasn't sure about what do change now and what to leave for later. As of now, everything relies on boolean behaviour. I'll adjust the code.

MAtahualpa:
Okay, I simply deleted the "!=0" check which leaves us with the integer values "0" for books (mIsScroll == 0) and "1" for scrolls (mIsScroll == 1) respectively. Please note that we should consider changing the name of the member variable, if we want to add more book types in the future.

PR #1209, opened 2017-02-16
Added ConstContainerStoreIterator (Task #3092)

dhustkoder:
Added a template base class for ContainerStoreIterator.
Also added setContainerStore method to ConstPtr class.

### DIFF
--- a/apps/openmw/mwclass/container.cpp
+++ b/apps/openmw/mwclass/container.cpp
@@ -145,7 +145,7 @@ namespace MWClass
         // make key id lowercase
         std::string keyId = ptr.getCellRef().getKey();
         Misc::StringUtils::lowerCaseInPlace(keyId);
-        for (MWWorld::ContainerStoreIterator it = invStore.begin(); it != invStore.end(); ++it)
+        for (MWWorld::ConstContainerStoreIterator it = invStore.cbegin(); it != invStore.cend(); ++it)
         {
             std::string refId = it->getCellRef().getRefId();
             Misc::StringUtils::lowerCaseInPlace(refId);
--- a/apps/openmw/mwclass/container.cpp
+++ b/apps/openmw/mwclass/container.cpp
@@ -135,7 +135,7 @@ namespace MWClass
         const std::string trapActivationSound = "Disarm Trap Fail";
 
         MWWorld::Ptr player = MWBase::Environment::get().getWorld ()->getPlayerPtr();
-        MWWorld::InventoryStore& invStore = player.getClass().getInventoryStore(player);
+        const MWWorld::InventoryStore& invStore = player.getClass().getInventoryStore(player);
 
         bool isLocked = ptr.getCellRef().getLockLevel() > 0;
         bool isTrapped = !ptr.getCellRef().getTrap().empty();
--- a/apps/openmw/mwworld/containerstore.hpp
+++ b/apps/openmw/mwworld/containerstore.hpp
@@ -119,13 +119,13 @@ namespace MWWorld
 
             virtual ContainerStore* clone() { return new ContainerStore(*this); }
 
-            ContainerStoreIterator begin (int mask = Type_All);
-
-            ContainerStoreIterator end();
-
             ConstContainerStoreIterator cbegin (int mask = Type_All) const;
-            
             ConstContainerStoreIterator cend() const;
+            ConstContainerStoreIterator begin (int mask = Type_All) const;
+            ConstContainerStoreIterator end() const;
+            
+            ContainerStoreIterator begin (int mask = Type_All);
+            ContainerStoreIterator end();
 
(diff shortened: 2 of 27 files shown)


scrawl:
I was thinking of a template <i>class</i> to further reduce redundancy. You can actually implement templates in the .cpp file, as long as all instantiations are declared in advance.

dhustkoder:
@scrawl 
Ok, then it would be something like:
```c++
using ContainerStoreIterator = ContainerStoreIteratorTemplate<ContainerStore*>;
using ConstContainerStoreIterator = ContainerStoreIteratorTemplate<const ContainerStore*>;
```
Is that ok?

dhustkoder:
@scrawl changed it, take a look see if is more or less what you have in mind.

scrawl:
Looks good to me. Could you add ContainerStore::cbegin() and cend() and port over a piece of code to the const version, to test if this is actually working as intended?

dhustkoder:
Yes, I'll finish it and then test it as you said.

apps/openmw/mwclass/container.cpp
```diff
@@ -145,7 +145,7 @@ namespace MWClass
         // make key id lowercase
         std::string keyId = ptr.getCellRef().getKey();
         Misc::StringUtils::lowerCaseInPlace(keyId);
-        for (MWWorld::ContainerStoreIterator it = invStore.begin(); it != invStore.end(); ++it)
```
dhustkoder: @scrawl  Testing it here, steping through GDB, all fine.
The only thing that bothers me, is when you try to convert ConstIterator to Iterator you get a linking error instead of an compiling error. But that keeps the code smaller.

scrawl: > The only thing that bothers me, is when you try to convert ConstIterator to Iterator you get a linking error instead of a compiling error. But that keeps the code shorter.

That sounds worth fixing. How much longer would the code have to be?

dhustkoder: Right now I could create the actual Iterator classes inheriting from Base, and then duplicate the constructors on them, because we can't inherit constructors without C++11. Or to not duplicate all contructors I could change some code in ContainerStore and create fewer constructors on the Iterator classes that takes the Base as argument, But I'm afraid that would ending with more calls to copy and slowing things down.

If you want, you could review as it is now and merge, and then I update my fork and solve this issue later with a better solution (maybe theres a template trickery I'm missing now) and I could also port all other parts of code which should be using the const iterator.

apps/openmw/mwworld/containerstore.hpp
```diff
@@ -187,96 +200,143 @@ namespace MWWorld
...
+        bool incIterator ();
+        ///< Increment iterator for selected type.
+        ///
+        /// \return reached the end?
 
         public:
+            template<class T>
+            ContainerStoreIteratorBase (const ContainerStoreIteratorBase<T>& other)
+            {
+                char CANNOT_CONVERT_CONST_ITERATOR_TO_ITERATOR[IsConvertible<T, PtrType, void>::value ? 1 : -1];
```
dhustkoder: @scrawl Got the linking error issue solved with this.

Error for this code:
```c++
// I didn't use cbegin() to make sure we can convert from nonconst to const
// and cend() to make sure we can compare const with nonconst
for (MWWorld::ConstContainerStoreIterator it = invStore.begin(); it != invStore.end(); ++it)
{
    MWWorld::ContainerStoreIterator nonconst = it; // error
    /* ... */
}
```
GCC:
gcc_error

CLANG:
clang_error

scrawl: Nice. Want to go ahead and port over more code to the const version then?

dhustkoder: Yeah, done. Ported simple loops that required very small change to code with no conversions (if you consider RVO / NRVO).

For example:
```c++
void Actor::block(const MWWorld::Ptr &ptr) const
{
    MWWorld::InventoryStore& inv = getInventoryStore(ptr);
    MWWorld::ContainerStoreIterator shield = inv.getSlot(MWWorld::InventoryStore::Slot_CarriedLeft);
    /* ... */
}
```
Changing the iterator to const version would trigger a conversion that can't be optimized by the compiler with RVO/NRVO. The right thing seems to create a const version for the method getSlot, but the internal representation is of course non const, then the conversion will be triggered anyway.

dhustkoder:
@scrawl Is there something else you'd like me todo in this PR before merging ?

scrawl:
> Changing the iterator to const version would trigger a conversion that can't be optimized by the compiler with RVO/NRVO.

I think we should go ahead and change those too. Don't worry about any conversion overhead, that should be negligible.

scrawl:
Are we done here?

dhustkoder:
@scrawl  Yes

PR #1210, opened 2017-02-16
Fix frame rate drop during actor update

Allofich:
This fixes the frame rate drop caused by my earlier PR that changed how combat decisions are made.

The frame rate drop can be seen clearly at Tel Branora, where there are several NPCs following another one. With this PR there I don't see a noticeable frame rate drop here anymore on my system.

There was unnecessary processing going on when finding an actors allies (those in follow or escort relationships). This fix works by first finding the "top" of the follow/escort hierarchy, then working down from there. It should also work for cases where actors are following/escorting each other.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -280,7 +280,7 @@ namespace MWMechanics
         }
     }
 
-    void Actors::engageCombat (const MWWorld::Ptr& actor1, const MWWorld::Ptr& actor2, bool againstPlayer)
+    void Actors::engageCombat (const MWWorld::Ptr& actor1, const MWWorld::Ptr& actor2, std::map<const MWWorld::Ptr, const std::set<MWWorld::Ptr> >& cachedAllies, bool againstPlayer)
     {
         CreatureStats& creatureStats1 = actor1.getClass().getCreatureStats(actor1);
         if (creatureStats1.getAiSequence().isInCombat(actor2))
@@ -306,7 +306,8 @@ namespace MWMechanics
         // Get actors allied with actor1. Includes those following or escorting actor1, actors following or escorting those actors, (recursive)
         // and any actor currently being followed or escorted by actor1
         std::set<MWWorld::Ptr> allies1;
-        getActorsSidingWith(actor1, allies1);
+
+        getActorsSidingWith(actor1, allies1, cachedAllies);
 
         // If an ally of actor1 has been attacked by actor2 or has attacked actor2, start combat between actor1 and actor2
         for (std::set<MWWorld::Ptr>::const_iterator it = allies1.begin(); it != allies1.end(); ++it)
@@ -328,10 +329,10 @@ namespace MWMechanics
                 aggressive = true;
         }
 
-        std::set<MWWorld::Ptr> playerFollowersAndEscorters;
-        getActorsSidingWith(MWMechanics::getPlayer(), playerFollowersAndEscorters);
+        std::set<MWWorld::Ptr> playerAllies;
+        getActorsSidingWith(MWMechanics::getPlayer(), playerAllies, cachedAllies);
 
-        bool isPlayerFollowerOrEscorter = std::find(playerFollowersAndEscorters.begin(), playerFollowersAndEscorters.end(), actor1) != playerFollowersAndEscorters.end();
+        bool isPlayerFollowerOrEscorter = std::find(playerAllies.begin(), playerAllies.end(), actor1) != playerAllies.end();
 
         // If actor2 and at least one actor2 are in combat with actor1, actor1 and its allies start combat with them
         // Doesn't apply for player followers/escorters        
@@ -341,7 +342,9 @@ namespace MWMechanics
             if (actor2.getClass().getCreatureStats(actor2).getAiSequence().isInCombat(actor1))
             {
                 std::set<MWWorld::Ptr> allies2;
-                getActorsSidingWith(actor2, allies2);
+
+                getActorsSidingWith(actor2, allies2, cachedAllies);
+
                 // Check that an ally of actor2 is also in combat with actor1
                 for (std::set<MWWorld::Ptr>::const_iterator it = allies2.begin(); it != allies2.end(); ++it)
                 {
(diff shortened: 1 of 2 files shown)


scrawl:
Even with the PR applied, I am still seeing a framerate drop in that location every second. Maybe there's an improvement compared to master but I don't think this is the right approach, anyway, this just complicates things for little benefit but does nothing about the core issue that is the cubic time complexity of this update, i.e.:

<pre>
get_followers(x): # O(n)
   ...
   foreach actor:
      if actor follows x: add(actor)

engage(actor1,actor2): O(n)
   get_followers(actor1)
   ...

main: O(n^3)
foreach actor1:
   foreach actor2:
      engage(actor1, actor2)

# example: 10 actors = 1000 checks = 10000 iterations within get_followers
</pre>

To properly solve this I think we'll need to somehow cache results so that the followers for a particular actor are never computed more than once per update.

wareya:
Wouldn't it also work to link following in both directions? Then you don't have to traverse other actors in the first place.

Allofich:
Sorry for the slow response. I've been occupied by other things. I guess we should try to get this fixed before the next release, though.

> I am still seeing a framerate drop in that location every second. 

For me this PR makes the drop no longer be a noticeable problem, but maybe it wouldn't be so in other cases.

scrawl, or anyone, really, who wants to "cut me off" and do a better fix for this, feel free to do so, I won't mind. Otherwise, I'll try to fix it when I can.

wareya:
0.42 has already been tagged, but any improvement would be good for 0.42 if it can still get crammed in. The performance problem is currently just so bad that it makes Tel Branora basically unplayable until/unless you kill the relevant NPCs.

Allofich:
Just noting that I'm working on a fix for this and so far it seems to have resolved the issue. I still need to do some more work on it.

Allofich:
Ok, uploaded.

Allofich:
@zinnschlag Assuming this fix is OK, is it possible to get this into 0.42 or is it too late?

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -306,7 +306,25 @@ namespace MWMechanics
...
+            // Cache ptrs and their sets of allies so we don't search for them again the next time engageCombat is called
+            cachedAllies.insert(std::make_pair(actor1, allies1));
+
+            for (std::set<MWWorld::Ptr>::const_iterator it = allies1.begin(); it != allies1.end(); ++it)
+            {
+                search = cachedAllies.find(*it);
+                if (search == cachedAllies.end())
+                    cachedAllies.insert(std::make_pair(*it, allies1));
+            }
+        }
```
scrawl: * Don't use auto (C++11)
* I see this code repeated 3 times. Probably best to move it into a function.

Allofich:
Done. Also fixed a problem that Travis complained about.

apps/openmw/mwmechanics/actors.cpp
Allofich: Done. Also fixed a problem that Travis complained about.

MiroslavR:
I can still notice hitching on my end, though it's _much_ better. I would say this is a must-have for 0.42.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -1591,6 +1596,30 @@ namespace MWMechanics
                 getActorsSidingWith(*it, out);
     }
 
+    void Actors::getActorsSidingWith(const MWWorld::Ptr &actor, std::set<MWWorld::Ptr>& out, std::map<const MWWorld::Ptr, const std::set<MWWorld::Ptr> >& cachedAllies) {
+        std::list<MWWorld::Ptr> followers = getActorsSidingWith(actor);
```
scrawl: Eh, this line looks to be misplaced.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -1591,6 +1596,30 @@ namespace MWMechanics
...
+
+        // If we have already found actor's allies, use the cache
+        std::map<const MWWorld::Ptr, const std::set<MWWorld::Ptr> >::const_iterator search = cachedAllies.find(actor);
+        if (search != cachedAllies.end())
+            out = search->second;
```
scrawl: I think this is wrong, as the method is supposed to <i>append</i> to out, not replace it.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -1591,6 +1596,30 @@ namespace MWMechanics
...
+        else
+        {
+            for (std::list<MWWorld::Ptr>::iterator it = followers.begin(); it != followers.end(); ++it)
+                if (out.insert(*it).second)
+                    getActorsSidingWith(*it, out);
```
scrawl: I think we can use the caching variant of the function here (if the append, not overwrite issue is fixed)

apps/openmw/mwmechanics/actors.cpp
Allofich: Done.

apps/openmw/mwmechanics/actors.cpp
Allofich: I don't think it matters, because you shouldn't come across anything in the cache while iterating through getActorsSidingWith(*it, out) anyway, right? Again, sorry if I'm misunderstanding.

apps/openmw/mwmechanics/actors.cpp
Allofich: Are you sure? Maybe I'm misunderstanding, but I think we don't want to append to out if we find it in the cache. If the actor is in the cache, we should be able to return its associated set. We want to append to out while doing the recursive function call, but I don't think it's possible to come across an actor already in the cache while we're doing that.

Sorry it's difficult to explain in words. Am I overlooking something?

Allofich:
Well I'm too tired to think about it more at the moment, but I'm trusting what you said scrawl and made the changes you suggested. It's hard to tell but there might be a slight frame rate improvement from it.

scrawl:
> is it possible to get this into 0.42

Done.

psi29a:
Nice work! :D

Allofich:
Thanks!

PR #1211, opened 2017-02-16
Fix Clang and OS X Travis CI warnings

Allofich:
This is a suggestion for fixing the warnings in the Clang and OS X Travis CI builds about overridden virtual functions in animation.cpp.

I don't really know what the effects of this might be, so this is just a suggestion since it seems to satisfy the compilers. The game ran without obvious problems in a quick test.

### DIFF
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -179,15 +179,6 @@ namespace
         {
         }
 
-        virtual void apply(osg::Group& node)
-        {
-            traverse(node);
-        }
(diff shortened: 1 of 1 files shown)


scrawl:
Although slightly redundant, that's not too bad. Would it be possible to remove those functions from the base class 'RemoveVisitor' now?

Allofich:
> Would it be possible to remove those functions from the base class 'RemoveVisitor' now?

Well, I'll give it a go and see what Travis-CI says. I'll also build locally and make sure the game runs.

Allofich:
No obvious problems running after building locally.

PR #1212, opened 2017-02-16
Replace light emitter check boxes with a combo box (fixes #3752)

MAtahualpa:
Replaces the four emitter check boxes in Light records with an "Emitter Type" combo box.

Related issue:
- Fixes #3752: Editor: Replace emitter check boxes in Light records with "Emitter Type" combo box (https://bugs.openmw.org/issues/3752)

Tests:
The changes were successfully tested in OpenMW-CS by manipulating several Light records. Please note that this fix also prevents users from erroneously assigning two or more emitter types at the same time. (I don't know which one would prevail in that case.)

### DIFF
--- a/apps/opencs/model/world/refidadapterimp.cpp
+++ b/apps/opencs/model/world/refidadapterimp.cpp
@@ -628,6 +628,25 @@ QVariant CSMWorld::LightRefIdAdapter::getData (const RefIdColumn *column, const 
     if (column==mColumns.mSound)
         return QString::fromUtf8 (record.get().mSound.c_str());
 
+    if (column == mColumns.mEmitterType)
+    {
+        int mask = ESM::Light::Flicker | ESM::Light::FlickerSlow | ESM::Light::Pulse | ESM::Light::PulseSlow;
+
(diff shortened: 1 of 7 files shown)


apps/opencs/model/world/refidadapterimp.cpp
```diff
@@ -653,6 +672,21 @@ void CSMWorld::LightRefIdAdapter::setData (const RefIdColumn *column, RefIdData&
         light.mData.mColor = value.toInt();
     else if (column==mColumns.mSound)
         light.mSound = value.toString().toUtf8().constData();
+    else if (column == mColumns.mEmitterType)
+    {
+        int mask = ~(ESM::Light::Flicker | ESM::Light::FlickerSlow | ESM::Light::Pulse | ESM::Light::PulseSlow);
+
+        if (value.toInt() == 0)
```
zinnschlag: For future reference, this is a candidate for a switch-case or even a simple array. But never mind for now.

MAtahualpa: Hm, okay. I'll keep that in mind.

PR #1213, opened 2017-02-16
Change AI equip logic

Allofich:
Fixes https://bugs.openmw.org/issues/3754.

There are two issues there. One NPC, Val, is choosing to not equip armor instead of equip her armor due to equal armor rating. In the original engine this seems to be the reverse, with equipping armor preferred even if the armor rating is the same.

The other issue is that the NPC Nargol gra-Uftharz is equipping her bracers rather than her gauntlets in OpenMW, while she equips the gauntlets in the original engine. The bracers and gauntlets have equal armor rating. From testing in the original engine, it looks to me like NPCs always favor gauntlets over bracers, regardless of stats. I'm not quite sure how to handle this. It seems bug-like, but it also affects the appearance of an NPC, in this case... I decided to recreate the original engine behavior by default and allow setting the AI to treat bracers and gauntlets equally as an option.

### DIFF
--- a/apps/openmw/mwworld/inventorystore.cpp
+++ b/apps/openmw/mwworld/inventorystore.cpp
@@ -291,9 +291,15 @@ void MWWorld::InventoryStore::autoEquip (const MWWorld::Ptr& actor)
                 {
                     if (old.getTypeName() == typeid(ESM::Armor).name())
                     {
-                        if (old.getClass().getEffectiveArmorRating(old, actor) >= test.getClass().getEffectiveArmorRating(test, actor))
-                            // old armor had better armor rating
+                        if (old.get<ESM::Armor>()->mBase->mData.mType < test.get<ESM::Armor>()->mBase->mData.mType)
                             continue;
+
+                        if (old.get<ESM::Armor>()->mBase->mData.mType == test.get<ESM::Armor>()->mBase->mData.mType)
+                        {
+                            if (old.getClass().getEffectiveArmorRating(old, actor) >= test.getClass().getEffectiveArmorRating(test, actor))
+                                // old armor had better armor rating
+                                continue;
+                        }
                     }
                     // suitable armor should replace already equipped clothing
                 }


scrawl:
No setting please. That seems way too minor for anyone to even care or notice.

Allofich:
In that case, I guess we should have the AI treat them equally. I'll remove the second commit.

scrawl:
Either way is fine with me. I'd tend towards copying the original.

Allofich:
Ok, well I'll redo the second commit without a setting, then. Changing the behavior could be part of a larger setting for "smarter" equipping or something.

apps/openmw/mwworld/inventorystore.cpp
```diff
@@ -272,7 +272,7 @@ void MWWorld::InventoryStore::autoEquip (const MWWorld::Ptr& actor)
         }
 
         if (iter.getType() == ContainerStore::Type_Armor &&
-                test.getClass().getEffectiveArmorRating(test, actor) <= std::max(unarmoredRating, 0.f))
+                test.getClass().getEffectiveArmorRating(test, actor) < std::max(unarmoredRating, 0.f))
```
MiroslavR: The problem is elsewhere. See https://github.com/OpenMW/openmw/pull/1215 for details.

Allofich: Ok, I'll remove that part of the PR and also the "fixes" message.

scrawl:
Did you check if the gauntlet/bracers change is still necessary as of #1215?

Allofich:
Yes, it is. Of course, if MiroslavR wants to suggest something else, that is fine with me.

MiroslavR:
There are two main differences between vanilla and OpenMW auto-equip:
* Different iteration order. Vanilla appears to sort the inventory list. I haven't found the responsible code, but it seems to be ordered by record magic number (not sure how exactly, but CLOT always precedes ARMO) and type (in descending order).
* When searching for equipped items, vanilla uses the ESM::Armor/Clothing types instead of slots. See https://github.com/OpenMW/openmw/pull/1107 for the reverse-engineered code. Bracers and gauntlets share the same inventory slot, but have different types (ESM::Armor::Type).

This is the exact order in which Nargol gra-Uftharz's items are processed:

|#  |Magic| Type | Name | Action |
|---|-----|------|------|--------|
|1 | WEAP| 8| Steel Battle Axe| skip
|2 | ARMO| 10| Iron Right Bracer| equip
|3 | ARMO| 9| Iron Left Bracer| equip
|4 | ARMO| 7| Iron Right Gauntlet| equip
|5 | ARMO| 6| Iron Left Gauntlet| equip
|6 | ARMO| 5| Iron Boots| equip
|7 | ARMO| 4| Iron Greaves| equip
|8 | ARMO| 3| Iron Right Pauldron| equip
|9 | ARMO| 2| Iron Left Pauldron| equip
|10| ARMO| 1| Iron Cuirass| equip
|11| ARMO| 0| Steel Helm| equip

So yes, I think gauntlets are always preferred over bracers, regardless of which one has better armor rating. Armor rating (in case of armor) and value (in case of clothing) comparisons are only performed for items of the same type.

The changes of this PR are probably okay, but maybe you could generalize them a bit by comparing type values instead:
```c++
                        if (old.get<ESM::Armor>()->mBase->mData.mType < test.get<ESM::Armor>()->mBase->mData.mType)
                            continue;

                        if (old.get<ESM::Armor>()->mBase->mData.mType == test.get<ESM::Armor>()->mBase->mData.mType)
                        {
                            if (old.getClass().getEffectiveArmorRating(old, actor) >= test.getClass().getEffectiveArmorRating(test, actor))
                                // old armor had better armor rating
                                continue;
                        }
```

Allofich:
Thanks for the information, it's interesting to see how the original engine works with it. I put in the code you suggested, and it works fine.

PR #1219, opened 2017-02-21
Verification results are ordered by Type in alphabetical order.

FedeWar:
The ReportModel::data method didn't know why it was called, now the proxy model sets the role to Qt::UserRole when the data is being sorted. So if UserRole is passed to data a QString (i.e. the Type name) is returned instead of the Type id.

### DIFF
--- a/apps/opencs/model/tools/reportmodel.cpp
+++ b/apps/opencs/model/tools/reportmodel.cpp
@@ -37,7 +37,7 @@ int CSMTools::ReportModel::columnCount (const QModelIndex & parent) const
 
 QVariant CSMTools::ReportModel::data (const QModelIndex & index, int role) const
 {
-	if (role!=Qt::DisplayRole && role!=Qt::UserRole)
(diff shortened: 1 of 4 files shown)


zinnschlag:
It seems the the initial sorting when opening a report is still not considering alphabetical order. Any idea about that?

FedeWar:
When Verify is clicked CSMTools::Tools::verifierMessage gets the reports but doesn't sort them, then the ReportView shows them without knowing if something changed.

FedeWar:
I'm a newbie in this project, but as far as I can see it is not possible. It is not possible without doing any big change.

FedeWar:
This is the cleanest fix I could make, now the elements are sorted as they are inserted.

zinnschlag:
After evaluating this fix I have come to the conclusion that it is correct, but confusing to the user, especially with slower report building processes. From a usability point of view starting with an unordered list is actually better.
Therefore I will merge this pull request in its original form. Please discard the last commit.

PR #1223, opened 2017-03-07
Fix for incorrect gun sounds with the Clean Hunter Rifles mod

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3781
Mod: http://mw.modhistory.com/download-90-10875 
Forum post: https://forum.openmw.org/viewtopic.php?f=40&t=4193&p=46232

The mod overrides crossbow sounds like so (script is attached to the weapon):
```
    If ( Player -> HasItemEquipped "A_3PJM_GUN_Xs" )
        If ( GetSoundPlaying "crossBOWPULL" == 1 )
            Player -> stopsound "crossBOWPULL"
            Player -> playsound "apjm_revolver_slow"
        endif
        
        if ( getsoundplaying "crossbowshoot" == 1 )
            player-> stopsound "crossbowshoot"
            Playsound "APJM_gunshot"
        endif
    endif
```
Apparently GetSoundPlaying called on an equipped item should also look for sounds played by the equipping actor. Don't ask me why. It's also super inconsistent - e.g. same does not apply to StopSound. Why the mod's author didn't just use "Player->GetSoundPlaying" I have no clue either.

### DIFF
--- a/apps/openmw/mwscript/soundextensions.cpp
+++ b/apps/openmw/mwscript/soundextensions.cpp
@@ -12,6 +12,9 @@
 #include "../mwbase/soundmanager.hpp"
 #include "../mwbase/windowmanager.hpp"
 
+#include "../mwworld/inventorystore.hpp"
+#include "../mwworld/class.hpp"
+
 #include "interpretercontext.hpp"
 #include "ref.hpp"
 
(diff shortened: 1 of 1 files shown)


psi29a:
Is this a moment where we want to fix this?

scrawl:
We have a similar workaround for ```GetDistance```. Could that be generalized in some way? Maybe there are more script commands affected by that issue, too?

MiroslavR:
They're quite different, actually. For example, this one is restricted to equipped items only. Also, GetDistance replaces containees with their containers to get position and this one takes properties of both into account.

psi29a:
@MiroslavR apologies, scrawl more or less answered it (i.e. we've done it elsewhere, why stop here).   My question is moot now, but I was wondering if this is an instance where the fix shouldn't be implemented because it re-implements bad behaviour that we shouldn't be trying to replicate.

PR #1224, opened 2017-03-12
fix skill window update

schwitzerm:
invoke winMgr->updateSkillArea() in mechanicsmanagerimp.cpp after looping over ESM::Skill::SkillEnum if any updates have occurred.

### DIFF
--- a/apps/openmw/mwgui/statswindow.cpp
+++ b/apps/openmw/mwgui/statswindow.cpp
@@ -237,7 +237,18 @@ namespace MWGui
                 nameWidget->setSize(nameWidget->getWidth() - (widthAfter-widthBefore), nameWidget->getHeight());
             }
 
-            if (value.getBase() < 100) {
+            if (value.getBase() < 100)
+            {
+                nameWidget->setUserString("Visible_SkillMaxed", "false");
+                nameWidget->setUserString("UserData^Hidden_SkillMaxed", "true");
+                nameWidget->setUserString("Visible_SkillProgressVBox", "true");
+                nameWidget->setUserString("UserData^Hidden_SkillProgressVBox", "false");
+
+                valueWidget->setUserString("Visible_SkillMaxed", "false");
+                valueWidget->setUserString("UserData^Hidden_SkillMaxed", "true");
+                valueWidget->setUserString("Visible_SkillProgressVBox", "true");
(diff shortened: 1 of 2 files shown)


scrawl:
I removed that update on purpose because the rebuild of all widgets in updateSkillArea() is a performance bottleneck and the skill progress for e.g. athletics changes very often as you run around.

The skill progress should still be updated, though, by  ```StatsWindow::setValue(int parSkil, SkillValue&)```. Can you look into why that's not working?

schwitzerm:
Right, the issue is that the logic to update the window for the skill is inside updateSkillArea(). This is not called anywhere, and was not called anywhere but where I put it back. It could probably go inside of StatsWindow::setValue(SkillEnum, SkillValue&), but the logic in updateSkillArea() iterates over all skills, which I am assuming you do not want.

I can pull the logic out of there and apply it individually for the widget in another function and call it from StatsWindow::setValue(SkillEnum, SkillValue&) if you like.

schwitzerm:
On closer inspection it appears to be that the new functions use mSkillWidgetMap while the old mSkillWidgets is still the main object being used throughout the UI. Will update.

MiroslavR:
Quoting my bugtracker post:
> Confirmed. The problem is that only the value widget is updated, while the name widget is not. Also, I think the "Maximum proficiency reached" message is not being updated at all after that change...

schwitzerm:
Latest push fixes the display for the progress bar, will double check for your note about "Maximum proficency reached" notification.

apps/openmw/mwgui/statswindow.cpp
```diff
@@ -240,6 +240,16 @@ namespace MWGui
             if (value.getBase() < 100) {
                 setSkillProgress(nameWidget, value.getProgress(), parSkill);
                 setSkillProgress(valueWidget, value.getProgress(), parSkill);
+            } else {
+                nameWidget->setUserString("Visible_SkillMaxed", "true");
```
MiroslavR: What if you set your skill below 100 again? The "Maximum proficiency reached" message should not stay visible.

Also, please respect our code style, that is:
```c++
if (...)
{
}
```
not
```c++
if (...) {
}
```

schwitzerm: pardon me, will update styling & add missing lines

PR #1227, opened 2017-03-18
Editor: Limit pathgrid creation to valid cells (Bug #3345)

rcutmore:
[Bug #3345](https://bugs.openmw.org/issues/3345)

The issue is described as cloned/added pathgrids being lost after reopening saved omwgame file. When looking into it more, it appears it's an issue with adding a pathgrid for a cell that doesn't exist.

This addresses the problem by using a combo box in the pathgrid creator to limit user input only to existing cells that don't have a pathgrid yet. The combo box is repopulated when background data changes to handle added/removed cells/pathgrids. Also, the create button isn't enabled until it verifies that the selected cell doesn't have a pathgrid already and that the cell exists.

### DIFF
--- a/apps/opencs/view/world/pathgridcreator.cpp
+++ b/apps/opencs/view/world/pathgridcreator.cpp
@@ -10,6 +10,7 @@
 #include "../../model/world/idtable.hpp"
 
 #include "../widget/droplineedit.hpp"
+#include "idvalidator.hpp"
 
 std::string CSVWorld::PathgridCreator::getId() const
 {
@@ -40,6 +41,7 @@ CSVWorld::PathgridCreator::PathgridCreator(
     CSMWorld::ColumnBase::Display displayType = CSMWorld::ColumnBase::Display_Cell;
     mCell = new CSVWidget::DropLineEdit(displayType, this);
     mCell->setCompleter(completionManager.getCompleter(displayType).get());
+    mCell->setValidator(new IdValidator(relaxedIdRules, this));
     insertBeforeButtons(mCell, true);
 
     connect(mCell, SIGNAL (textChanged(const QString&)), this, SLOT (cellChanged()));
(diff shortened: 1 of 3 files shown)


zinnschlag:
The idea is fundamentally okay, but I have an issue with GUi aspect. The combo box now provides a huge list. Unfortunately on most desktops combo box lists don't scale all that well (who ever came up with the idea of adding down and up buttons on an overly long popup menu instead of a scroll bar deserves to be skinned alive).
I think we need to find an alternative. Normal text box with auto complete menu maybe? That feature is already available in the creator toolbar. Pathgrids for non-existent cells can still be rejected by disabling the create button.

rcutmore:
@zinnschlag I had the same reservation about the combo box. One problem with using a normal text input is that certain characters aren't allowed, which would prevent the user from adding a pathgrid for existing cells that have *#*, *-*, *'*, and possibly other characters, in the ID (there's a separate bug on the tracker related to that [here](https://bugs.openmw.org/issues/2838)).

I'll work on the change, but should we also at some point update *IdValidator* to allow any characters used in IDs from the base game? I haven't looked at that code very closely so I'm not sure what the potential negative consequences would be.

zinnschlag:
The problem with the additional characters is that we will most likely forbid them once OpenMW 1.1. comes around. We are disallowing them already in the editor to stop people from creating dirty content.
That being said, it would make sense to modify the id-validator to allow these characters. But only when the validator is used for referencing to existing IDs. Creating new records with these bad IDs should still be forbidden (pathgrid is a special case here, since the pathgrid record ID actually references the cell ID).

rcutmore:
@zinnschlag I pushed some changes to switch from using a combo box for the input. The input is now a drop target from the cells table and has auto-completion.

One thing I noticed is that all creator classes that use a DropLineEdit don't have ID validation set up. This is set up in *GenericCreator* but only for the original input. I added it here but I'm thinking the ID validation changes should be tackled in a separate PR (make sure ID validation is set up for each creator class and allow some of the extra characters mentioned previously). Let me know what you think.

zinnschlag:
Agreed on everything.

PR #1228, opened 2017-03-19
Options to show hidden weapon parameters

akortunov:
1) Option to show damage of arrows and bolts through tooltip ([Feature #2923](https://bugs.openmw.org/issues/2923)).
Disabled by default. Add 
`show projectile damage = true`
to [Game] section in settings.cfg to enable it.

2) Option to show attack speed and reach of melee weapons through tooltip
This option is not needed for clean Morrowind, but it useful for plugins ([Weapon Range Balance](http://www.nexusmods.com/morrowind/mods/6051/) for example).
Disabled by default. Add 
`show melee info = true`
to [Game] section in settings.cfg to enable it.

### DIFF
--- a/apps/openmw/mwclass/weapon.cpp
+++ b/apps/openmw/mwclass/weapon.cpp
@@ -1,6 +1,7 @@
 #include "weapon.hpp"
 
 #include <components/esm/loadweap.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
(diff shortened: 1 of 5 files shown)


scrawl:
Can you do a rebase to remove the last two commits?

akortunov:
Rebased.

scrawl:
Although I'm not a big fan of merging tons more options that (ideally) would be addressed with enhanced modding capabilities in the future instead, this one is simple enough that I'll make an exception.

But first, can you update the documentation too?
https://github.com/OpenMW/openmw/blob/master/docs/source/reference/modding/settings/game.rst

akortunov:
I updated the documentation.

PR #1241, opened 2017-03-26
Editor: Change creator ID validation

rcutmore:
[Bug #2838](https://bugs.openmw.org/issues/2838)
[Previous Discussion](https://github.com/OpenMW/openmw/pull/1227)

This contains a few changes related to ID validation:

- Documents why ID validation isn't used in some cases
- Changes mislabeled input and makes labels match style of other labels
- ~~Allows '.' in IDs~~
- Adds creator subclass for body parts to allow first person parts to be added

After thinking about it more, I believe it makes sense to not have an input validator set up for certain creator subclasses. These classes are restricted to existing record IDs so the input validation is important where those records are created but not in these particular subclasses. We just need to make sure it matches an existing ID plus any other rules specific to the creator subclass.

### DIFF
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -70,6 +70,7 @@ opencs_units (view/world
     cellcreator pathgridcreator referenceablecreator startscriptcreator referencecreator scenesubview
     infocreator scriptedit dialoguesubview previewsubview regionmap dragrecordtable nestedtable
     dialoguespinbox recordbuttonbar tableeditidaction scripterrortable extendedcommandconfigurator
+    bodypartcreator
     )
 
 opencs_units_noqt (view/world
--- a/apps/opencs/view/world/bodypartcreator.cpp
+++ b/apps/opencs/view/world/bodypartcreator.cpp
@@ -1,0 +1,54 @@
+#include "bodypartcreator.hpp"
+
+#include <QCheckBox>
+
+#include "../../model/world/data.hpp"
+#include "../../model/world/universalid.hpp"
+
+std::string CSVWorld::BodyPartCreator::getId() const
+{
+    std::string id = CSVWorld::GenericCreator::getId();
+
+    if (mFirstPerson->isChecked())
+    {
+        id += ".1st";
+    }
+
+    return id;
+}
+
+CSVWorld::BodyPartCreator::BodyPartCreator(
+    CSMWorld::Data& data,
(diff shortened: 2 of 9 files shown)


zinnschlag:
Allowing . in IDs is a no go. Makes it impossible to use them in scripts without putting them into "". That is exactly the situation we want to avoid with the validator.

I think the rest looks acceptable. But we need to be very careful about not moving further in this direction, because the premise of OpenMW-CS is that data consistency is not enforced on GUI level, but by the verifier function (plenty of use-cases in the old TES-CS where this kind of enforcement became a problem).

rcutmore:
@zinnschlag Is there a more appropriate approach for the issue in the bug tracker or is it an issue that shouldn't be fixed? Relaxed ID rules could be used for body parts but that doesn't seem to fit the intended purpose for the relaxed rules, at least from my understanding. It would allow "." in the IDs for body parts but it would also allow many other characters as well.

zinnschlag:
I think we are more or less okay with what we have here. As for body parts: I am not too familiar with the body part record type. I have seen the .1st name a few times in Morrowind.esm. Do you know if that is hardcoded and has a special meaning? If yes, then we obviously need to treat body parts as a special case. For every other new ID . should still be forbidden.

rcutmore:
I'm not too familiar with body parts but ".1st" in the ID seems to signify first person meshes. I searched a bit and one example where we use this fact can be found [here](https://github.com/OpenMW/openmw/blob/73b6c34a238da2a729716860cbed3672045845d0/apps/openmw/mwgui/race.cpp#L301).

I removed the commit allowing "." in IDs from this PR and I'll work on a creator subclass for body parts. If the rest of this PR is fine as is, it can be merged and I'll submit the remaining changes separately or I can add them to this PR.

zinnschlag:
Just add them. We don't need a separate PR for that.

rcutmore:
@zinnschlag I added a BodyPartCreator class to this. A checkbox is used to indicate that a part should be saved as a first person part but let me know if you think a different widget should be used to control this.

One thing I wasn't sure about is whether we should check that a normal body part ID exists before allowing the first person version to be added. For example, if the user is trying to add a first person body part that will end up with the ID "test_gauntlets.1st", should "test_gauntlets" already exist? Currently this just checks to make sure a duplicate doesn't exist.

zinnschlag:
No need. The user is free to add the body parts in the reverse order. Merging ...

PR #1242, opened 2017-03-26
Fix AppVeyor warnings

Allofich:
Fixes two AppVeyor warnings.

One is a warning in the msvc 2013 environment about a double definition of LAND_SIZE.

The other is a warning in the msvc 2015 environment about a function parameter being overridden.

### DIFF
--- a/apps/openmw/mwrender/objects.cpp
+++ b/apps/openmw/mwrender/objects.cpp
@@ -149,9 +149,9 @@ void Objects::removeCell(const MWWorld::CellStore* store)
 
             if (ptr.getClass().isNpc() && ptr.getRefData().getCustomData())
             {
-                MWWorld::InventoryStore& store = ptr.getClass().getInventoryStore(ptr);
-                store.setInvListener(NULL, ptr);
-                store.setContListener(NULL);
+                MWWorld::InventoryStore& invStore = ptr.getClass().getInventoryStore(ptr);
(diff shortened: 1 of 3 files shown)


scrawl:
That LAND_SIZE definition was required for me to solve a linking error. Let's see if travis builds that.

Allofich:
Travis is catching up on the merged PRs at the moment but I'll come back to see what happened.

On the subject of warnings, have you seen this in the OS X build?

/Users/travis/build/OpenMW/openmw/apps/openmw/mwrender/globalmap.cpp:140:49: warning: shifting a negative signed value is undefined [-Wshift-negative-value]
                                y2 = (SCHAR_MIN << 4) / 2048.f;

Allofich:
Why does ESM::Land::LAND_SIZE need that second definition when ESM::Land::REAL_SIZE and ESM::Land::LAND_TEXTURE_SIZE don't?

I tried adding loadland.hpp as an include but the Travis builds still failed. I'm not sure what else to try.

This is just a minor cosmetic thing mostly but it would be nice to figure it out.

Allofich:
I was able to stop the linking error by casting LAND_SIZE to an integer in two spots in Storage::fillVertexBuffers(). They were in std::min() assigning to an integer, with the other parameter being cast to integer, and I guess somehow the problem was related to that.

PR #1243, opened 2017-03-26
Journal improvements

akortunov:
This commit blocks journal when console enabled (fixes [bug #2351](https://bugs.openmw.org/issues/2351)).
Also adds "book close" sound on journal closing, as in vanilla game.


Allofich:
Sounds good. I haven't tested this PR but I can confirm there is no book closing sound when closing the journal in OpenMW, but there is the in the original engine.

PR #1247, opened 2017-04-02
Ini-importer progress bar fills the whole width of the ui element now

pvdk:
Changed the progress bar on the settings tab of the launcher to fill the whole width of the ui element it resides in.

### DIFF
--- a/files/ui/settingspage.ui
+++ b/files/ui/settingspage.ui
@@ -131,14 +131,17 @@
           </property>
          </widget>
         </item>
-        <item row="3" column="0">
-         <widget class="QProgressBar" name="progressBar">
-          <property name="maximum">
-           <number>4</number>
-          </property>
-         </widget>
-        </item>
        </layout>
+      </item>
+      <item>
+       <widget class="QProgressBar" name="progressBar">
+        <property name="maximum">
+         <number>4</number>
+        </property>
+        <property name="textVisible">
+         <bool>false</bool>
+        </property>
+       </widget>
       </item>
      </layout>
     </widget>


pvdk:
You know what, I'd rather rip out the whole progress bar and add a messagebox on completion (or even a label under the button or something if you really want to avoid using a messagebox).

zinnschlag:
Is there a specific issue with the progress bar that makes you want to take it out?

psi29a:
Is it even necessary? For something that should take, at most a few seconds, but most of the time happens in a blink of an eye shouldn't warrant one.  KISS (keep it simple stupid)

zinnschlag:
Okay. Let's try to think out of the box. How about this:

A light besides the button.
* red: no ini import has been performed yet; fallback values are missing in openmw.cfg and running OpenMW won't work properly
* yellow (and blinking): an ini import is in progress
* green: fallback values are present and running OpenMW should work without problems
And of course tooltips for the light that explains the meaning.

That assumes we won't get default fallback values in the near feature. With default values, the red state would be redundant and we might want to just change the text on the button to "in Progress" or something and after an import has completed add "(done)" to the button text.

zinnschlag:
Thought some more about this issue, but came up with nothing. I suggest we simply accept this PR and then review the situation again when we have defaults for the fallback values. Everyone okay with that?
@pvdk 
@psi29a

psi29a:
I'm OK with this. :)

pvdk:
What about this:
screenshot_20170407_124511
screenshot_20170407_124426
screenshot_20170407_124324

psi29a:
LGTM

When we get default values we can modify this further to be 'optional'. :)

PR #1249, opened 2017-04-03
Remove evil "implicit GetJournalIndex" hack from expression parser

MiroslavR:
I'm assuming the hack was added to silence syntax errors in these Bloodmoon scripts:
* WereChange2Script
* wereDreamScript2
* wereDreamScript3

...which used `if ( BM_BearHunt1 == 0 )` instead of `if ( GetJournalIndex BM_BearHunt1 == 0 )`. Thing is, these scripts are unused. Moreover, they produce corrupt bytecode and result in runtime errors in vanilla MW:
```
Script Error: EXPRESSION in testscript
Left eval
```

This implicit GetJournalIndex behavior seems to be completely made up by us, so I'm getting rid of it and blacklisting the unused scripts instead. Another reason why this hack is evil is that it breaks when parsing the following expression, thinking -> is unexpected:
```
( "AA_Laurenna"->GetMysticism )
```
Where "AA_Laurenna" is an NPC ID clashing with a journal ID.

### DIFF
--- a/components/compiler/exprparser.cpp
+++ b/components/compiler/exprparser.cpp
@@ -311,22 +311,6 @@ namespace Compiler
                 return true;
             }
 
-            // die in a fire, Morrowind script compiler!
-            if (const Extensions *extensions = getContext().getExtensions())
-            {
-                if (getContext().isJournalId (name2))
-                {
-                    // JournalID used as an argument. Use the index of that JournalID
-                    Generator::pushString (mCode, mLiterals, name2);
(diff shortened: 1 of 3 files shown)


zinnschlag:
I don't remember where this came from. But your arguments sound reasonable. Guess we will see if this causes any unexpected breakage in mods once we do the release.
Can you please add an issue to the tracker? I would like to have this change on record.

MiroslavR:
> Can you please add an issue to the tracker? I would like to have this change on record.

[Done.](https://bugs.openmw.org/issues/3816)

PR #1250, opened 2017-04-03
Hide WorldButton when in game mode

resilar:
The patch improves the map window by hiding the world/local button in game mode, i.e., "non-GUI" mode. In my opinion, this makes a pinned map window a better replacement for the minimap (the world button is useless in game mode because it cannot be pressed anyway). Illustration: http://i.imgur.com/WDZbztj.png

This is my first contribution to the OpenMW project, so please review the patch (it's only a few lines of code). Basically, the hiding functionality is implemented by overriding `WindowBase::setVisible()` in `mapwindow.cpp`. An alternative approach would have been to implement `MapWindow::showButton()` method and call it from `windowmanagerimp.cpp`, but this would have required more code and logic. However, I'm not really a C++ programmer (I write mostly C), so I can't tell if overriding a virtual base class method is considered a bad practice in this context.

### DIFF
--- a/apps/openmw/mwgui/mapwindow.cpp
+++ b/apps/openmw/mwgui/mapwindow.cpp
@@ -774,6 +774,12 @@ namespace MWGui
         mLastScrollWindowCoordinates = currentCoordinates;
     }
 
+    void MapWindow::setVisible(bool visible)
+    {
+        WindowBase::setVisible(visible);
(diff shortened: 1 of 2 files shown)


scrawl:
Nice little detail, I like it!

While the code isn't the cleanest (it relies on the fact that we call setVisible, even when the visibility of the window hasn't changed), I'm not really sure of a better way, so I'll merge it.

PR #1251, opened 2017-04-04
Editor: reset camera after cell drag and drop

rcutmore:
[Bug #3484](https://bugs.openmw.org/issues/3484)

When a cell is initially rendered a default camera position gets set [here](https://github.com/OpenMW/openmw/blob/master/apps/opencs/view/render/scenewidget.cpp#L312). If that cell is replaced by another via drag and drop the camera position isn't changed because *mCamPositionSet* is still set to true. When replacing cells, this change flags the camera position as not being set so it gets updated.

### DIFF
--- a/apps/opencs/view/render/cameracontroller.cpp
+++ b/apps/opencs/view/render/cameracontroller.cpp
@@ -632,6 +632,11 @@ namespace CSVRender
         getCamera()->getViewMatrix().orthoNormal(getCamera()->getViewMatrix());
     }
 
+    void OrbitCameraController::reset()
+    {
+        mInitialized = false;
+    }
+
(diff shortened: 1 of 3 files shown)


zinnschlag:
I am seeing a problem here. If you drop a new interior cell into a scene sub view the camera jumps to a new position when you perform the first camera movement. A missing update problem maybe?

rcutmore:
@zinnschlag You're right. I'm only seeing that issue with the orbiting camera, is that the same for you?

zinnschlag:
Correct.

rcutmore:
@zinnschlag The orbit cam controller needed to be re-initialized to properly center on the new cell, this contains the necessary change now.

PR #1254, opened 2017-04-12
Remove unnecessary line

Allofich:
I noticed that the new god mode code introduced a shadowed variable warning on MSVC.


scrawl:
Seems like the whole line can be removed.

Allofich:
Yeah, you're right. It's the same as the earlier line.

PR #1255, opened 2017-04-13
Fixed magicka calculation on chargen (bug #3694)

akortunov:
See [this](https://forum.openmw.org/viewtopic.php?f=8&t=4035) forum thread.
This fix just recalculates dynamic stats on review dialog whatever we changed intelligence from default 30, or no.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -73,6 +73,8 @@ namespace MWMechanics
         MWMechanics::CreatureStats& creatureStats = ptr.getClass().getCreatureStats (ptr);
         MWMechanics::NpcStats& npcStats = ptr.getClass().getNpcStats (ptr);
 
+        npcStats.setNeedRecalcDynamicStats(true);
+
         const ESM::NPC *player = ptr.get<ESM::NPC>()->mBase;
 
         // reset


Allofich:
Nice work.

It won't matter normally, but I just found that original Morrowind actually updates the magicka after the birthsign is selected. You can check with player->getmagicka in the interval between when you select the birthsign and when the review menu comes up.

Allofich:
You could put it in CharacterCreation::onBirthSignDialogDone, to be the same as Morrowind, although I don't know if there are any mods that would depend on this. Seems it would only matter if they delayed or skipped the review dialog for some reason.

akortunov:
> Morrowind actually updates the magicka after the birthsign is selected
You are right. I can modify CharacterCreation::onBirthSignDialogDone. But maybe we should recalculate magicka after class and race dialogs too as fallback?

Allofich:
I don't think it matters for class, does it? As for race, it looks like Morrowind does not even update the character attributes when race is chosen. That also happens at birthsign choice.

Well... I guess you could put in a recalculation for the race menu, then "EnableRaceMenu" called mid-game would work as may be expected (although different from original Morrowind). I see that OpenMW already recalculates the racial powers and abilities when EnableRaceMenu is called, unlike Morrowind, so this would be consistent with what OpenMW already does.

So I guess I would suggest that, and we can change it to Morrowind behavior if there is a problem with a mod.

akortunov:
Ok, we are recalculate magicka after race and birthsign choose now.

scrawl:
Better place is: MechanicsManager::buildPlayer(), I think.

PR #1256, opened 2017-04-13
Adjustments to RemoveItem behavior

Allofich:
Fixes https://bugs.openmw.org/issues/3796.

The problem is caused because in OpenMW NPCs would always try to equip a replacement when RemoveItem was used. The mod relies on the behavior of the original engine, which is that a replacement is not equipped. A replacement _is_ equipped in the original engine when you click to remove items from the inventory of a dead body or companion.

This also fixes the issue of the loud equipping noise. Only the player has equip sounds in the original engine.

This PR doesn't do anything about the
`Failed to load 'meshes/add world art': Resource 'meshes/add world art' not found, using marker_error.nif instead`
messages I mentioned in the comments there. That happens because the mod uses items that have no art files attached.

### DIFF
--- a/apps/openmw/mwscript/containerextensions.cpp
+++ b/apps/openmw/mwscript/containerextensions.cpp
@@ -144,7 +144,8 @@ namespace MWScript
                         if (::Misc::StringUtils::ciEqual(iter->getCellRef().getRefId(), item))
                             itemName = iter->getClass().getName(*iter);
 
-                    int numRemoved = store.remove(item, count, ptr);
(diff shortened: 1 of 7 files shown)


scrawl:
I think that instead of adding a new parameter to ```remove```, we can just call unequip before removing and that way a replacement won't be equipped. (but don't do so if itemCount is > toRemoveCount, e.g. for arrows you can have more than one equipped at the same time).

scrawl:
On the other hand, we can't unequip before casting to InventoryStore so I guess that'd be a little clunky.

PR #1258, opened 2017-04-16
Fixed drowning widget initialization with non-default "fHoldBreathTime" GMST value

akortunov:
If you have non-default "fHoldBreathTime" GMST value, drowning widget is displayed on a new game start.

To fix this we should set mTimeToStartDrowning from actual GMST value instead of using a hardcoded 20 sec value.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -328,13 +328,13 @@ namespace MWMechanics
                 winMgr->setValue(fbar, stats.getFatigue());
             }
 
-            static const float fHoldBreathTime = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>()
-                        .find("fHoldBreathTime")->getFloat();
-
             float timeToDrown = stats.getTimeToStartDrowning();
 
             if(timeToDrown != mWatchedTimeToStartDrowning)
             {
+                static const float fHoldBreathTime = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>()
+                        .find("fHoldBreathTime")->getFloat();
+
                 mWatchedTimeToStartDrowning = timeToDrown;
 
                 if(timeToDrown >= fHoldBreathTime || timeToDrown == -1.0) // -1.0 is a special value during initialization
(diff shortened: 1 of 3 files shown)


scrawl:
This change makes the assumption that World will be fully initialized by the time an NpcStats is created.

Maybe it's better to use a special value like -1 on initialization, then check against this during the update.

akortunov:
> check against this during the update

Actors::updateDrowning(const MWWorld::Ptr& ptr, float duration) is the suitable place for this check?

akortunov:
And another question: why are not we check that a player is actually underwater
(see [mechanicsmanagerimp.cpp](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp#L331)). Performance issues?

akortunov:
Ok, attempt #2:

1. mTimeToStartDrowning initialized by special -1 value, actual value sets in Actors::updateDrowning() during actors update.
2. mWatchedTimeToStartDrowning comparison is not sufficient since we are not update value of this variable when context changes (savegame loaded, new game started etc.).
 I suggest to add World::isUnderwater() check.

scrawl:
> mWatchedTimeToStartDrowning comparison is not sufficient since we are not update value of this variable when context changes (savegame loaded, new game started etc.).
I suggest to add World::isUnderwater() check.

I don't get it. mWatchedTimeToStartDrowning always reflects what is set on the WindowManager; if the two match, no update should be needed.

akortunov:
Sorry, my fault. We just need to check against "special value" in a widget update too.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -328,12 +328,16 @@ namespace MWMechanics
                 winMgr->setValue(fbar, stats.getFatigue());
             }
 
-            if(stats.getTimeToStartDrowning() != mWatchedTimeToStartDrowning)
-            {
-                const float fHoldBreathTime = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>()
+            static const float fHoldBreathTime = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>()
```
scrawl: Scope can be reduced.

akortunov: Yes, you are right. Fixed.

akortunov:
I reduced scope of fHoldBreathTime variable. Let me know if I need to do rebase.

PR #1259, opened 2017-04-19
Fix a few Cppcheck warnings

Allofich:
This fixes a few warnings that I saw after using the code analysis tool Cppcheck.

Fixes:
Boolean value assigned to floating point variable
physicssystem.cpp
Lines 1497, 1506

The scope of the variable 'isRangedCombat' can be reduced.
aicombat.cpp
Line 503

The scope of the variable 'targetTime' can be reduced.
animation.cpp
Line 1061

Variable 'current' is modified but its new value is never used
scene.cpp
Line 545

Suspicious calculation. Please use parentheses to clarify the code.
alchemy.cpp
Line 174, 176

Also, it shows

Unused private function: 'MovementSolver::reflect'
physicssystem.cpp
Line 233

but I left it there in case it is maybe still wanted.

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -500,8 +500,7 @@ namespace MWMechanics
         {
             // get the range of the target's weapon
             float rangeAttackOfTarget = 0.f;
-            bool isRangedCombat = false;
-            MWWorld::Ptr targetWeapon = MWWorld::Ptr();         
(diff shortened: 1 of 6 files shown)


scrawl:
> Unused private function: 'MovementSolver::reflect'
physicssystem.cpp
Line 233
but I left it there in case it is maybe still wanted.

Please do remove it. We have the git history if we ever need it again.

Allofich:
OK, I just wanted to be sure.

PR #1262, opened 2017-04-20
Fix more warnings

Allofich:
Fixed the following warnings from CPPCheck

> The interator 'it' is invalid after the element it pointed to has been erased. Dereferencing or comparing it with another iterator is invalid operation.

For aisequence.cpp, line 108

This was in AISequence::erase. Also, this function is used in two spots. One doesn't take the return value, and the other was doing the same thing as AISequence::stopCombat, so in addition to fixing the warning, I changed that call to stopCombat and made AISequence::erase a void return function.

> Parameter --- is passed by value. It could be passed as a (const) reference which is usually faster and recommended in C++

This showed for several files. I changed them all to const references.

And from MSVC, I fixed two shadowing warnings from recent commits.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -114,9 +114,7 @@ void adjustCommandedActor (const MWWorld::Ptr& actor)
     }
 
     if (!check.mCommanded && hasCommandPackage)
-    {
         stats.getAiSequence().erase(it);
-    }
 }
 
 void getRestorationPerHourOfSleep (const MWWorld::Ptr& ptr, float& health, float& magicka)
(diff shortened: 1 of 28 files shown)


apps/openmw/mwmechanics/actors.cpp
```diff
@@ -845,10 +834,7 @@ namespace MWMechanics
             }
         }
         else
-        {
-            static const float fHoldBreathTime = world->getStore().get<ESM::GameSetting>().find("fHoldBreathTime")->getFloat();
-            stats.setTimeToStartDrowning(fHoldBreathTime);
-        }
+            stats.setTimeToStartDrowning(world->getStore().get<ESM::GameSetting>().find("fHoldBreathTime")->getFloat());
```
scrawl: Can we keep the ```static const float``` variable for fHoldBreathTime? I don't think we want to retrieve this GMST for every actor, every frame.

apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -96,16 +96,16 @@ std::list<AiPackage*>::const_iterator AiSequence::end() const
...
+void AiSequence::erase(std::list<AiPackage*>::const_iterator package)
 {
     // Not sure if manually terminated packages should trigger mDone, probably not?
     for(std::list<AiPackage*>::iterator it = mPackages.begin(); it != mPackages.end(); ++it)
     {
         if (package == it)
         {
-            AiPackage* packagePtr = *it;
-            delete packagePtr;
-            return mPackages.erase(it);
```
scrawl: I don't understand why cppcheck flagged this as a problem. list::erase returns 'An iterator pointing to the element that followed the last element erased by the function call'.

Either way, I'm all for simplifying things. Carry on.

apps/openmw/mwmechanics/actors.cpp
Allofich: I've put it back, while getting rid of the later redeclaration

apps/openmw/mwmechanics/aisequence.cpp
Allofich: The difference I saw between this and other similar functions in the code was that this one created a pointer (packagePtr) for the iterator value and stored it in there rather than just accessing it directly, so I think maybe cppcheck didn't like lines 106 and 107. There's not any reason to store the value in a new pointer before deleting it is there? I guess it was done for readability? Anyway I made the change for your other comment but didn't change anything here.

PR #1264, opened 2017-04-20
Run scripts for voiced dialogue

Allofich:
For https://bugs.openmw.org/issues/3787.

### DIFF
--- a/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
+++ b/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
@@ -171,7 +171,7 @@ namespace MWDialogue
 
                     MWScript::InterpreterContext interpreterContext(&mActor.getRefData().getLocals(),mActor);
                     win->addResponse (Interpreter::fixDefinesDialog(info->mResponse, interpreterContext));
-                    executeScript (info->mResultScript);
+                    executeScript (info->mResultScript, mActor);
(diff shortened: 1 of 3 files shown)


apps/openmw/mwdialogue/dialoguemanagerimp.cpp
```diff
@@ -647,6 +647,9 @@ namespace MWDialogue
             if (!info->mSound.empty())
                 sndMgr->say(actor, info->mSound);
         }
+
+        mActor = actor;
```
scrawl: It's incorrect to overwrite mActor. That is the actor that the dialogue window is opened for, but ```say``` is completely independent of that window. I suggest to add a new parameter (actor) to executeScript.

Allofich: Fixed.

apps/openmw/mwdialogue/dialoguemanagerimp.cpp
```diff
@@ -647,6 +647,9 @@ namespace MWDialogue
             if (!info->mSound.empty())
                 sndMgr->say(actor, info->mSound);
         }
+
+        if (!info->mResultScript.empty())
```
MiroslavR: Info is not checked for NULL, which is causing crashes on my end. Please fix it.

Allofich: Sorry, new PR up.

PR #1265, opened 2017-04-21
Use "sLevel" GMST instead of hardcoded text in load game menu

akortunov:
Since ComboBox texts are not translated, we can translate text manually before adding it to ComboBox.

### DIFF
--- a/apps/openmw/mwgui/savegamedialog.cpp
+++ b/apps/openmw/mwgui/savegamedialog.cpp
@@ -169,7 +169,7 @@ namespace MWGui
                         className = "?"; // From an older savegame format that did not support custom classes properly.
                 }
 
-                title << " (#{sLevel} " << it->getSignature().mPlayerLevel << " " << className << ")";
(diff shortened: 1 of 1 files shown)


apps/openmw/mwgui/savegamedialog.cpp
```diff
@@ -168,9 +169,9 @@ namespace MWGui
                         className = "?"; // From an older savegame format that did not support custom classes properly.
                 }
 
-                title << " (Level " << it->getSignature().mPlayerLevel << " " << className << ")";
+                title << " (#{sLevel} " << it->getSignature().mPlayerLevel << " " << className << ")";
```
scrawl: It's possible that ```className``` contains a markup sequence that needs to be escaped (e.g. '#' )

Therefore, use:
```
                title << " (#{sLevel} " << it->getSignature().mPlayerLevel << " " << MyGUI::TextIterator::toTagsString(className) << ")";
```

akortunov: Reworked. Tested with # in class name, works fine.

PR #1266, opened 2017-04-21
Improved hands bodyparts selection fallback

akortunov:
Fixes  [bug #2594](https://bugs.openmw.org/issues/2594).

If we can not found 1st-person hand bodypart for current gender, we should try fallbacks in this order:
1. Try to use 3d person skin for same gender
2. Try to use 1st person skin for opposite gender
3. Try to use 3d person skin for opposite gender

We already have similar logic for armor/clothing hands.

In current implementation we select only first bodypart that matchs **any** of these three conditions, even if better substitution will be found later in list.

UPDATE:
Females can use male bodyparts as fallback, but males can not use female bodyparts as fallback.

### DIFF
--- a/apps/openmw/mwrender/npcanimation.cpp
+++ b/apps/openmw/mwrender/npcanimation.cpp
@@ -726,6 +726,19 @@ void NpcAnimation::removePartGroup(int group)
     }
 }
 
+bool NpcAnimation::isFirstPersonPart(const ESM::BodyPart* bodypart)
+{
+    return (bodypart->mId.size() >= 3)
+        && bodypart->mId[bodypart->mId.size()-3] == '1'
+        && bodypart->mId[bodypart->mId.size()-2] == 's'
+        && bodypart->mId[bodypart->mId.size()-1] == 't';
+}
+
+bool NpcAnimation::isFemalePart(const ESM::BodyPart* bodypart)
+{
+    return bodypart->mData.mFlags & ESM::BodyPart::BPF_Female;
+}
+
 bool NpcAnimation::addOrReplaceIndividualPart(ESM::PartReferenceType type, int group, int priority, const std::string &mesh, bool enchantedGlow, osg::Vec4f* glowColor)
 {
     if(priority <= mPartPriorities[type])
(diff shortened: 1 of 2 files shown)


apps/openmw/mwrender/npcanimation.cpp
```diff
@@ -1096,37 +1109,71 @@ const std::vector<const ESM::BodyPart *>& NpcAnimation::getBodyParts(const std::
...
+                        if(!parts[bIt->second])
+                            parts[bIt->second] = &*it;
+
                         ++bIt;
                     }
                 }
-                continue;
             }
 
+            // We do not need such fallback for non-hand bodyparts
```
scrawl: What has this check to do with hands?

apps/openmw/mwrender/npcanimation.cpp
```diff
@@ -1096,37 +1109,71 @@ const std::vector<const ESM::BodyPart *>& NpcAnimation::getBodyParts(const std::
...
+                    BodyPartMapType::const_iterator bIt = sBodyPartMap.lower_bound(BodyPartMapType::key_type(bodypart.mData.mPart));
+                    while(bIt != sBodyPartMap.end() && bIt->first == bodypart.mData.mPart)
+                    {
+                        // If we have substitution from p.3 and better substitution from p.2 found
+                        if(parts[bIt->second] && !isSameGender && isFemalePart(parts[bIt->second]) != female)
+                            parts[bIt->second] = &*it;
+
+                        if(!parts[bIt->second])
+                            parts[bIt->second] = &*it;
+
```
scrawl: This code path looks strange. From what I can tell, what is assigned here is always overwritten later by the block on line 1168 or 1183.

apps/openmw/mwrender/npcanimation.cpp
```diff
@@ -1096,37 +1109,71 @@ const std::vector<const ESM::BodyPart *>& NpcAnimation::getBodyParts(const std::
             if (!Misc::StringUtils::ciEqual(bodypart.mRace, race))
                 continue;
 
-            bool partFirstPerson = (bodypart.mId.size() >= 3)
-                    && bodypart.mId[bodypart.mId.size()-3] == '1'
-                    && bodypart.mId[bodypart.mId.size()-2] == 's'
-                    && bodypart.mId[bodypart.mId.size()-1] == 't';
-            if(partFirstPerson != (firstPerson))
```
scrawl: Maybe we can make the code a bit easier to read. For example:

```
for (int i=0; i<4; ++i)
{
   if i == 0:
      accept no fallback
   if i == 1:
      accept 1st/3rd fallback
   if i == 2:
      accept gender fallback
   if i == 3:
      accept 1st/3rd gender fallback
   
   BodyPartMapType::const_iterator bIt = ....
}
```

akortunov: Good suggestion, but we will have 4 iterations instead of one. If it is OK, I will rewrite this function.

apps/openmw/mwrender/npcanimation.cpp
akortunov: Hmm... It seems this comment is misleading.
I meaned that if a current bodypart is a hand, it is handled by fallback logic above. Otherwise we should skip this bodypart in this check.

apps/openmw/mwrender/npcanimation.cpp
scrawl: Don't worry about that, the normal case (no fallback) is what we should optimize for.

akortunov:
Updated. Males can not use female bodyparts as fallbacks.

Any suggestions about refactoring without performance loss?

PR #1267, opened 2017-04-21
Treat =< and => as == in scripts

Allofich:
For https://bugs.openmw.org/issues/3744

### DIFF
--- a/components/compiler/scanner.cpp
+++ b/components/compiler/scanner.cpp
@@ -408,6 +408,11 @@ namespace Compiler
                     special = S_cmpEQ;
                 else if (c=='=')
                     special = S_cmpEQ;
+                else if (c == '>' || c == '<')  // Treat => and =< as ==
+                {
+                    special = S_cmpEQ;
+                    mErrorHandler.warning (std::string("invalid operator =") + c + ", treating it as ==", mLoc);
+                }
                 else
                 {
                     special = S_cmpEQ;


Allofich:
I just found that `=<<` also works as "equals," and so does `=<<>,` but `=<<>6` fails, so it looks like you you can have an arbitrary number of < or > in there as long as it doesn't have anything else. But I don't think we need to worry about these cases.

Allofich:
Also see my comment in the bug. Should this be fixed?

Capostrophic:
IMO OpenMW basically shouldn't crap pants when it stumbles upon those typos. Ignoring them (say, to copy Morrowind engine behavior of "i dunno so let it be "equal"" like your solution) instead of outright dropping script execution doesn't really encourage bad plugin design. However, yes, Kortunov is right, there definitely should be a warning - at least for OpenMW-CS' script verifier.

Allofich:
Updated it with a warning message

components/compiler/scanner.cpp
```diff
@@ -408,6 +408,11 @@ namespace Compiler
                     special = S_cmpEQ;
                 else if (c=='=')
                     special = S_cmpEQ;
+                else if (c == '>' || c == '<')  // Treat => and =< as ==
+                {
+                    special = S_cmpEQ;
+                    mErrorHandler.warning ("reversed <= or >= operator, treating it as ==", mLoc);
```
scrawl: To be more specific, I'd use:
"invalid operator =" + c + ", treating it as =="

Allofich: OK I used your suggestion.

PR #1269, opened 2017-04-21
Don't issue errors for unused scripting arguments

Allofich:
For https://bugs.openmw.org/issues/3762.

### DIFF
--- a/components/compiler/parser.cpp
+++ b/components/compiler/parser.cpp
@@ -72,7 +72,7 @@ namespace Compiler
     bool Parser::parseInt (int value, const TokenLoc& loc, Scanner& scanner)
     {
         if (!(mOptional && mEmpty))
-            reportSeriousError ("Unexpected numeric value", loc);
+            reportWarning ("Unexpected numeric value, ignoring", loc);
         else
             scanner.putbackInt (value, loc);
 
@@ -87,7 +87,7 @@ namespace Compiler
     bool Parser::parseFloat (float value, const TokenLoc& loc, Scanner& scanner)
     {
         if (!(mOptional && mEmpty))
-            reportSeriousError ("Unexpected floating point value", loc);
+            reportWarning ("Unexpected floating point value, ignoring", loc);
         else
             scanner.putbackFloat (value, loc);
 
@@ -103,7 +103,7 @@ namespace Compiler
         Scanner& scanner)
     {
         if (!(mOptional && mEmpty))
-            reportSeriousError ("Unexpected name", loc);
+            reportWarning ("Unexpected name, ignoring", loc);
         else
             scanner.putbackName (name, loc);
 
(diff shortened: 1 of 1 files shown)


Allofich:
Note that I haven't actually been able to test the mod mentioned in the bug report. I've been trying to download it but can't seem to get the right files. But the script command in the bug report and the one suggested in the comments "tgm 1" work as expected.

Allofich:
Don't merge, this causes problems with script parsing.

Allofich:
It may take me a while to figure this out. A script in Bloodmoon, ColonySeler, causes problems. It has an incorrectly formed PositionCell command, with the cell name as the first argument followed by the coordinates, but the coordinates are supposed to come first. I don't think this command does anything to the NPC it's on in the original engine, and if used on the player it puts you in the 0 coordinate outside cell. It seems like if a string is passed in an integer parameter the original engine interprets it as 0.

I'll return to this later. Any advice is welcome.

akortunov:
Hmm... It seems Morrowind handles command arguments similar to Bash shell. It is just forms an array of arguments (string values) and passes them to command handler.

Handler itself takes care about validation, missing arguments, type casting etc. 
There are no errors when too many arguments - they are just unused elemenths of array.
There are no errors when argument has wrong type - they are all strings. When handler cast string to another type (integer, for example), it can just take default value.

When I typed

`player->additem "Gold_100" aaa`

in Morrowind console, I just got a script error - 'count "0" is not allowed for "Gold_100"'.

scrawl:
> It may take me a while to figure this out. A script in Bloodmoon, ColonySeler, causes problems.

We have a very specific workaround for that [here](0737d6cb72fd9ac15109e50acfe4ca8960877d30).

Allofich:
The link doesn't work. Based on the URL it looks like it was linking to https://github.com/OpenMW/openmw/commit/0737d6cb72fd9ac15109e50acfe4ca8960877d30, but that doesn't seem to be related to this.

Allofich:
I think I found the workaround you are talking about, in lineparser.cpp 

                        // workaround for broken positioncell instructions.
                        /// \todo add option to disable this
                        if (Misc::StringUtils::lowerCase (loc.mLiteral)=="positioncell")
                            errorDowngrade.reset (new ErrorDowngrade (getErrorHandler()));

and

                        // Ignore argument exceptions for positioncell.
                        /// \todo add option to disable this
                        if (Misc::StringUtils::lowerCase (loc.mLiteral)=="positioncell")
                        {
                            SkipParser skip (getErrorHandler(), getContext());
                            scanner.scan (skip);
                            return false;
                        }

Allofich:
The script parsing is too complicated for me to understand any time soon, so I'm going to close this and remove myself from the issue on the bug tracker. There's another scripting issue that I saw in the forums and will open a bug report for.

PR #1270, opened 2017-04-21
Check for NULL before running voice dialogue scripts

Allofich:
Requested fix.


Allofich:
Fixes https://bugs.openmw.org/issues/3834

PR #1271, opened 2017-04-22
Correct handling of bodyparts with multiple NiTriShapes 

akortunov:
Fixes bug [#3835](https://bugs.openmw.org/issues/3835).

Tested with Robert's Bodies, Better Bodies and New Argonian Bodies.

### DIFF
--- a/components/misc/stringops.hpp
+++ b/components/misc/stringops.hpp
@@ -59,6 +59,25 @@ public:
         return std::lexicographical_compare(x.begin(), x.end(), y.begin(), y.end(), ci());
     }
 
+    static bool ciLessSubstring(const std::string &x, const std::string &y)
+    {
+        int xlen = x.length();
+        int ylen = y.length();
+
+        std::string::const_iterator xit = x.end();
+        std::string::const_iterator yit = y.end();
+        if (xlen > ylen)
+        {
+            xit = x.begin() + ylen;
+        }
+        else if (xlen < ylen)
+        {
+            yit = y.begin() + xlen;
+        }
+
+        return std::lexicographical_compare(x.begin(), xit, y.begin(), yit, ci());
+    };
+
     static bool ciEqual(const std::string &x, const std::string &y) {
         if (x.size() != y.size()) {
             return false;
@@ -112,6 +131,14 @@ public:
         bool operator()(const std::string& left, const std::string& right) const
         {
             return ciLess(left, right);
+        }
+    };
+
+    struct CiStartsWithComp
+    {
+        bool operator()(const std::string& left, const std::string& right) const
+        {
+            return ciLessSubstring(left, right);
         }
     };
 };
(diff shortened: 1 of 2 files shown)


scrawl:
1. That compare needs to be case-insensitive.
2. Can we do better than linear time? This code path runs very often. 
Some data (time spent in isReservedName for openmw --start=balmora): master=0.857ms, PR= 5.88ms

akortunov:
I will try to implement a binary search. How can I measure the time?

scrawl:
Use osg::Timer:
```
            osg::Timer timer;
            static double total=0;
            for (std::set<std::string, Misc::StringUtils::CiComp>::iterator it = reservedNames.begin() ; it != reservedNames.end(); ++it)
            {
                std::string boneName = *it;

                if ((name.size() >= boneName.size() && name.compare(0, boneName.size(), boneName) == 0))
                {
                    total += timer.time_m();
                    return true;
                }
            }
            total += timer.time_m();
            OSG_ALWAYS << "total " << total << std::endl;
            return false;
```

Maybe you can keep using set::find and simply adjust the comparator function (currently CiComp), though it might be a problem that arguments can be passed in any order.

akortunov:
> simply adjust the comparator function (currently CiComp)

Comparator is just for ordering items in set, not for searching.
Maybe we need to cache result of isReservedName(), if this function is often used?

scrawl:
> Comparator is for ordering items in set (lesser-greater),

It's also used for [find](http://www.cplusplus.com/reference/set/set/find/):

_"Two elements of a set are considered equivalent if the container's comparison object returns false reflexively (i.e., no matter the order in which the elements are passed as arguments)."_

That's why the current code reservedNames.find(...) is effectively a case-insensitive search.

> Also I found that sometimes bodypart name is empty. Should I return false in this case?

Yes, an empty name is a trivial false result.

I couldn't observe any speed-up though from changing
<pre>
reservedNames.find(name) != reservedNames.end() 
</pre>
to
<pre>
!name.empty() && reservedNames.find(name) != reservedNames.end()
</pre>

akortunov:
I committed a new implementation. I got roughly an equal performance with this PR and current master now, but measurement is imprecise.

Tested with "openmw --start=balmora".

scrawl:
Well, crap. That doesn't work correctly:
<pre>
reservedNames.find("Hea") // incorrectly matches 'Head'
</pre>

akortunov:
Got it. Feel free to close this PR. Sorry for bothering you.

akortunov:
Update: I sligtly optimized my first, linear search version.
You can try to use it until someone suggests a better solution.

scrawl:
How annoying that the STL is useless here.

This solution should be better: https://github.com/OpenMW/openmw/commit/813af876f6508bd2d17cdac6eabc3f5a2f24cda4

Numbers (```--start=balmora```):
<pre>
akortunov/bodypartfix = 6 ms
813af87~1 = 5 ms
813af87 = 3.6 ms
</pre>

PR #1272, opened 2017-04-22
Service menus layout fixes

akortunov:
Fixed layout of service menus to looks line vanilla ones.
Windows affected:
1) Alchemy
2) Enchanting
3) Travelling
4) Spell buying
5) Spell creation
6) Merchant repair

Note:
Maximal text input length in these fields is capped by 31 symbols now, as in Morrowind (see [bug #3831](https://bugs.openmw.org/issues/3831)). I will revert it, if we do not need this feature.

### DIFF
--- a/apps/openmw/mwgui/spellbuyingwindow.cpp
+++ b/apps/openmw/mwgui/spellbuyingwindow.cpp
@@ -26,8 +26,6 @@ namespace MWGui
         , mLastPos(0)
         , mCurrentY(0)
     {
-        setCoord(0, 0, 450, 300);
-
         getWidget(mCancelButton, "CancelButton");
         getWidget(mPlayerGold, "PlayerGold");
         getWidget(mSpellsView, "SpellsView");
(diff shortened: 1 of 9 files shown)


scrawl:
> Maximal text input length in these fields is capped by 31 symbols now, as in Morrowind (see bug #3831). I will revert it, if we do not need this feature.

Please revert it. We try to avoid arbitrary limits. If the concern is that item titles would be too long to fit the tooltip, we have fancy [scrolling text](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwgui/tooltips.cpp#L538) ;)

akortunov:
Reverted 31 symbols cap.

Sorry for off-topic, but can we use any features from [this thread](https://forum.openmw.org/viewtopic.php?f=6&t=4269)?

scrawl:
> Sorry for off-topic, but can we use any features from this thread?

If you can somehow address this: _There is a small problem in tool select menu: there is no suitable GMST for a caption_ then, sure.

akortunov:
Recharge widget has no this problem - I simply use GMST from enchantment dialog.
For repair widget I can only use English text or remove a caption now.

There are no problems with spell sorting and enchant chance (my second post).

PR #1273, opened 2017-04-23
Keep an item health when enchant it

akortunov:
Current implementation of enchanting in OpenMW does not keeps an item condition when we enchant it - the item will be fully repaired after enchanting.

This PR keeps the old item condition, if it has it (weapon, armor etc.).

### DIFF
--- a/apps/openmw/mwmechanics/enchanting.cpp
+++ b/apps/openmw/mwmechanics/enchanting.cpp
@@ -88,9 +88,22 @@ namespace MWMechanics
         const ESM::Enchantment *enchantmentPtr = MWBase::Environment::get().getWorld()->createRecord (enchantment);
         std::string newItemId = mOldItemPtr.getClass().applyEnchantment(mOldItemPtr, enchantmentPtr->mId, getGemCharge(), mNewItemName);
 
-        // Add the new item to player inventory and remove the old one
-        store.remove(mOldItemPtr, 1, player);
+        // Add the new item to player inventory
         store.add(newItemId, 1, player);
+
+        // We should keep the item condition
+        MWWorld::Ptr newItemPtr = store.search(newItemId);
+
+        const bool hasOldItemHealth = mOldItemPtr.getClass().hasItemHealth(mOldItemPtr);
+        const bool hasNewItemHealth = newItemPtr.getClass().hasItemHealth(newItemPtr);
+        if(hasOldItemHealth && hasNewItemHealth)
+        {
+            int itemHealth = mOldItemPtr.getClass().getItemHealth(mOldItemPtr);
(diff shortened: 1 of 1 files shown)


Allofich:
I just checked and the original engine also restores items to full health.

I agree keeping the old condition does make sense, but when "magic" is involved you can argue anything, like that the enchanting magic also repairs the item. I think we should be careful about changing from original game behavior. Maybe there should be a post on the forum or a feature request on the bug tracker to get some opinions.

akortunov:
I created the feature request: https://bugs.openmw.org/issues/3837

akortunov:
> "magic" is involved 
No magic, just usual problem with item "upgrade" in various games, when game replaces one item for other, do not bothering itself to keep the old item state (health, ammo in clip, etc.).

Allofich:
Yes, I know it's probably just because of replacing the item with a new one, not intended to be "the enchanting repairs it." It's probably fine but people can be unhappy with even small changes so I thought it may be good to get a few opinions.

akortunov:
> it may be good to get a few opinions

Yes, you are right. Waiting for feedback from another users.

MiroslavR:
> No magic, just usual problem with item "upgrade" in various games, when game replaces one item for another, do not bothering itself to keep the old item state (health, ammo in clip, etc.).

That's just your opinion. There is no way to tell what was intended, so let's not stray away from vanilla behavior.

akortunov:
I think we should tag this PR as "Discussion" at least.

Capostrophic:
Feedback about the change by mere mortals from local TES community forum:
— This absolutely should be introduced.
— Sounds reasonable.
— Also supporting the feature.
— Great fix, keep it please.

rhtucker:
If it is not difficult to maintain, I think it would be ideal to preserve complete vanilla behavior, while also allowing the option to fix these kinds of questionable bugs. I imagine people would prefer something like MCP where you can customize your preferences with a list of check boxes.

akortunov:
Well, we have three options at least:
1) Accept PR.
2) Close PR.
3) Make an option in settings.cfg.

Allofich:
Making it an option seems fine to me.

Ghostbird:
I'm in favour of making this an option (preferably enabled by default), or just adding this right away. It seems to have been an engine limitation in the original Morrowind.

scrawl:
Seems like a niche feature. I doubt the average user would care or even notice. It's easy to explain the repair as 'magic' as someone put it. To not make the settings file and maintenance a complete mess, we should only add settings that have a clear value. IMO the proper way to handle things like this is with dehardcoding / scripting; which we're not quite ready for yet.

In terms of options that have a clear value, maybe we should start with some of what is provided by MCP  (with their permission of course) as they already put in a lot of work in RD & testing.

PR #1274, opened 2017-04-23
Fix wrong incrementation

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3838 (regression from 0e429ae).

A premature `continue` would not trigger `++i`.


MiroslavR:
The regression was introduced in 0.41. It should not cause crashes, but can lead to effects disappearing or showing incorrect information (only visually). Do we want to cherry-pick the fix into 0.42?

scrawl:
cherrypicked 1c5c82dd9f27688e053ebe8f70f37ed56cd0afda

PR #1275, opened 2017-04-24
Use osgDB::ifstream and osgDB::ofstream

Allofich:
Replacing all usage of std::ifstream or std::ofstream, or boost::filesystem::ifstream or boost::filesystem::ofstream with the osgDB equivalents gets rid of all LNK4006 warnings in MSVC builds, which are about ignoring second definitions of already defined symbols.

The game and saving and loading, etc. seems to work in both OpenMW and OpenMW-CS.

### DIFF
--- a/apps/opencs/editor.cpp
+++ b/apps/opencs/editor.cpp
@@ -292,7 +292,7 @@ bool CS::Editor::makeIPCServer()
         mPid /= "openmw-cs.pid";
         bool pidExists = boost::filesystem::exists(mPid);
 
-        mPidFile.open(mPid);
+        mPidFile.open(mPid.string().c_str());
 
         mLock = boost::interprocess::file_lock(mPid.string().c_str());
         if(!mLock.try_lock())
--- a/apps/opencs/editor.hpp
+++ b/apps/opencs/editor.hpp
@@ -4,7 +4,8 @@
 #include <memory>
 
 #include <boost/interprocess/sync/file_lock.hpp>
-#include <boost/filesystem/fstream.hpp>
+
+#include <osgDB/fstream>
 
 #include <QObject>
 #include <QString>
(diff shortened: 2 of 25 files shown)


scrawl:
I don't want to change dependencies just to get rid of a warning (which there may be another way of solving). Also, I am not sure if osgDB::ifstream is functionally equivalent to boost::filesystem::ifstream when it comes to the UTF8 path issues on windows.

Allofich:
OK, well, it doesn't sound like you'll change your mind, but I thought I remembered someone saying that we intend to to eventually stop using Boost, and so I thought this might be OK as it's a small step in that direction. I haven't been able to find any other way to get rid of the warnings.

The OSG docs at http://trac.openscenegraph.org/documentation/OpenSceneGraphReferenceDocs/a01351.html claim that it converts between UTF-8 and UTF-16, and say to "Always use one of these classes (osgDB::open, osgDB::ofstream and osgDB::ifstream) in any OpenSceneGraph code instead of the STL equivalent."

I'm using Windows, so I could test the UTF-8 conversion if there is a particular situation that we need to account for. But if you're sure you don't want the changes, I'll close the PR.

scrawl:
> The OSG docs at http://trac.openscenegraph.org/documentation/OpenSceneGraphReferenceDocs/a01351.html claim that it converts between UTF-8 and UTF-16, and say to "Always use one of these classes (osgDB::open, osgDB::ofstream and osgDB::ifstream) in any OpenSceneGraph code instead of the STL equivalent."

Nope, it only does that when enabling a build option that's disabled by default:

<pre>
IF (WIN32)
    OPTION(OSG_USE_UTF8_FILENAME "Set to ON to use a UTF8 locale for filenames instead of the default locale." OFF)
    MARK_AS_ADVANCED(OSG_USE_UTF8_FILENAME)
ENDIF()
</pre>

Allofich:
Oh well. Moving on.

PR #1276, opened 2017-04-26
OpenMW-CS: Add block commenting/uncommenting action in script editor (Feature #3274)

PlutonicOverkill:
https://bugs.openmw.org/issues/3274
Adds a context menu option to block comment/uncomment code in the script editor. Also adds a keyboard shortcut which can be changed in the "Key Bindings" section of the preferences panel. I'm not sure if it should be given its own category, but it doesn't seem to fit in any of the others. There's probably also a better way to add the menu item/shortcut.

### DIFF
--- a/apps/opencs/model/prefs/state.cpp
+++ b/apps/opencs/model/prefs/state.cpp
@@ -326,6 +326,10 @@ void CSMPrefs::State::declare()
     declareShortcut ("orbit-roll-right", "Roll Right", QKeySequence(Qt::Key_E));
     declareShortcut ("orbit-speed-mode", "Toggle Speed Mode", QKeySequence(Qt::Key_F));
     declareShortcut ("orbit-center-selection", "Center On Selected", QKeySequence(Qt::Key_C));
+
+    declareSubcategory ("Script Editor");
+    declareShortcut ("script-editor-comment", "Comment Selection", QKeySequence());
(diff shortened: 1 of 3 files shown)


zinnschlag:
This seems to collide with the existing context menu (still rendering for me, but all old menu items disabled).

PlutonicOverkill:
Fixed. It turns out that the context menu is generated differently depending on where createStandardContextMenu() is called.

zinnschlag:
I am not sure how Undo/Redo actually work there. But if these functions manipulate the text directly that would break OpenMW-CS, since it bypasses the real Undo/Redo system. At a minimum these menu items need to be disabled. Ideally they should be removed.

PlutonicOverkill:
Added a somewhat hacky fix removing the context menu items; I think the Undo/Redo shortcuts are all going through the proper system now. Also enclosed the comment/uncomment actions in an edit block so undo/redo applies to the whole action.

zinnschlag:
Thanks.

Question: On our tracker the issue was marked as resolved by an account named Capostrophic. Is that you?

PlutonicOverkill:
No, that wasn't me. As a matter of fact, I did try to register for the bugtracker, but I still haven't received the confirmation email.

zinnschlag:
That should happen instantly. You may want to PM  lgromanowski on the forum about this issue.

PlutonicOverkill:
Nevermind, I got the email. It must have just got lost in my inbox somewhere.

PR #1246, opened 2017-03-28
Menu improvements

akortunov:
1. Load game menu now have a confirmation dialog.
2. Added dynamic resize of confirmation dialogs
3. Disabled transparency for save/load menu, confirmation dialog widget (as in vanilla).
4. Count window: layout changes, disabled transparency (this widget is nearly unusable with transparency).

PR #1277, opened 2017-04-26
Fixed count widget buttons alignment

akortunov:
Fixes regression in [PR #1246](https://github.com/OpenMW/openmw/pull/1246) (see last 4 comments).

Since we calculate widget width in the code, I had to calculate buttons position in code too.
@scrawl, this PR was created by your request.

### DIFF
--- a/files/mygui/openmw_count_window.layout
+++ b/files/mygui/openmw_count_window.layout
@@ -16,13 +16,15 @@
             <Property key="Page" value="1"/>
             <Property key="WheelPage" value="1"/>
         </Widget>
-        <Widget type="HBox" skin="" position="0 91 140 24" align="Center Bottom">
-            <Widget type="AutoSizedButton" skin="MW_Button" position="0 0 53 24" align="Center Top" name="OkButton">
+        <Widget type="HBox" skin="" position="0 91 592 24" align="Center Bottom HStretch">
+            <Widget type="Spacer" />
+            <Widget type="AutoSizedButton" skin="MW_Button" position="0 0 53 24" align="Left Top" name="OkButton">
                 <Property key="Caption" value="#{sOk}"/>
             </Widget>
-            <Widget type="AutoSizedButton" skin="MW_Button" position="0 0 86 24" align="Center Top" name="CancelButton">
+            <Widget type="AutoSizedButton" skin="MW_Button" position="0 0 86 24" align="Right Top" name="CancelButton">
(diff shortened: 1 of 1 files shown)


scrawl:
You're not supposed to position buttons manually, the HBox will do that to its children whenever it's resized. If you want to center the buttons, AFAIK you can do that by putting a spacer widget on the left and one on the right.

akortunov:
> AFAIK you can do that by putting a spacer widget on the left and one on the right.

Unfortunately, in this case spacer widget on the left just takes the all available space, and widget on the right has 0 width.

I used this spacing widget:
`                    <Widget type="Widget">
                <UserString key="HStretch" value="true"/>
            </Widget>`

> The HBox would need to set align="HStretch"

I also tried this alignment, but without success. Parent widget size is setted by engine on runtime, and HBox is not resized properly.

> the HBox will do that to its children whenever it's resized

When the engine shrinks the counting widget window from 600px to 330px or so, HBox size is not changed, so child buttons are not centered.

Tested with stock MyGUI v3.2.2.

scrawl:
Can you upload that here so I can check what went wrong? I know there is a bug with the HBox which you may or may not be running into.

akortunov:
> Can you upload that here so I can check what went wrong?

Done.

scrawl:
I haven't tested it yet, but the width of the HBox should match the width of the parent window. HStretch will only change the size relatively on resizes of the parent window, but it won't make it the same width.

scrawl:
We should definitely write some documentation on GUI layouting because it's a little unintuitive (even aside from the "hack" that HBox/VBox was made as, which has some bugs left I think).

scrawl:
We could also add some kind of ContentAlign setting to HBox/VBox which would eliminate the need for using spacers in many cases.

akortunov:
> the width of the HBox should match the width of the parent window

Even in this case first spacing widget tries to fit 600px width and buttons are not visible.

scrawl:
I added a new Spacer class to master that works properly. Use ```<Widget type="Spacer"/>```, change the HBox width to 600px and it should work.

akortunov:
I uploaded a fixed version.

In my testing it worked fine even with sOk and sCancel of non-default length.

Note: For some reason HBox can not be placed over left border, but can be placed over right border. I have to use 592px (that's 600px - 2*borderWidth) instead of 600px width to fix it.

PR #1278, opened 2017-04-28
OpenMW-CS: Highlight occurrences when cursor is placed over a name (Feature #2509)

PlutonicOverkill:
https://bugs.openmw.org/issues/2509
Adds a setting to the script editor to enable highlighting other occurrences of a selected name. Currently it highlights all types of tokens, but it could be modified to highlight only variables. I had to use a flag in ScriptEdit::markOccurrences() to ensure it is only running once for each ScriptEdit instance, because otherwise QTextCursor::select() sends the changeHighlighting signal and it ends up being called recursively. I could have used a different signal to fix this, such as cursorPositionChanged, but that would probably impact performance more.
I also had to do some rebasing after I accidentally pushed the previous pull request from master branch, so I hope everything's still OK for a merge.

### DIFF
--- a/apps/opencs/model/prefs/state.cpp
+++ b/apps/opencs/model/prefs/state.cpp
@@ -152,6 +152,7 @@ void CSMPrefs::State::declare()
         setRange (0, 10000);
     declareInt ("error-height", "Initial height of the error panel", 100).
         setRange (100, 10000);
+    declareBool ("highlight-occurrences", "Highlight other occurrences of selected names", true);
     declareSeparator();
     declareColour ("colour-int", "Highlight Colour: Integer Literals", QColor ("darkmagenta"));
     declareColour ("colour-float", "Highlight Colour: Float Literals", QColor ("magenta"));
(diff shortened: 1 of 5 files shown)


zinnschlag:
The problem with this feature is that it only takes effect when the cursor is placed via mouse clicks. Moving it via cursor keys does nothing. That is a very inconsistent user experience.

PlutonicOverkill:
Sorry for the delayed response. I've changed it so that it rehighlights every time the cursor is moved, and only for variable names. I also removed the bool member and resorted to disconnecting and reconnecting each time. I don't know if this is noticeably slower (I'm testing using Windows debug builds, which are not known for their speed) but it is probably more readable.

PlutonicOverkill:
There's already an option to disable it. Turning it off disables highlighting for both the script editor and the highlighter, so that if it is turned off, it shouldn't need to rehighlight everything every time the cursor is moved, and the performance difference should be equivalent to checking a bool (in the editor and the highlighter).

PR #1279, opened 2017-04-29
Fix actors not returning to their wander locations

Allofich:
Fixes https://bugs.openmw.org/issues/3656

This simplifies AIWander by just using mInitialActorPosition, which is stored in saved games, in place of mReturnPosition, which is not stored in saved games, seemed redundant and was only set when a combat or pursuit package was added.

### DIFF
--- a/apps/openmw/mwmechanics/aisequence.cpp
+++ b/apps/openmw/mwmechanics/aisequence.cpp
@@ -295,21 +295,6 @@ void AiSequence::stack (const AiPackage& package, const MWWorld::Ptr& actor)
     if (actor == getPlayer())
         throw std::runtime_error("Can't add AI packages to player");
 
-    if (package.getTypeId() == AiPackage::TypeIdCombat || package.getTypeId() == AiPackage::TypeIdPursue)
-    {
-        // Notify AiWander of our current position so we can return to it after combat finished
-        for (std::list<AiPackage *>::const_iterator iter (mPackages.begin()); iter!=mPackages.end(); ++iter)
-        {
-            if((*iter)->getTypeId() == AiPackage::TypeIdCombat && package.getTypeId() == AiPackage::TypeIdCombat
-                && (*iter)->getTarget() == (&package)->getTarget())
-            {
-                return; // already in combat with this actor
-            }
-            else if ((*iter)->getTypeId() == AiPackage::TypeIdWander)
-                static_cast<AiWander*>(*iter)->setReturnPosition(actor.getRefData().getPosition().asVec3());
-        }
-    }
-
     // Stop combat when a non-combat AI package is added
     if (isActualAiPackage(package.getTypeId()))
         stopCombat();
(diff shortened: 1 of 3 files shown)


scrawl:
I think they're supposed to return to where they last wandered to, not to where they started wandering.

Allofich:
I'll try testing the original engine later.

Allofich:
It looks to me like what happens is that they resume pathing to whatever their destination was, whether there is a pathgrid there or not. The original engine doesn't seem to save this data (or if it does it doesn't restore it), so saving and restoring it would be an improvement over the original.

So the fix will need to be reworked and involve adding to the save file.

Allofich:
Another issue is that in original Morrowind if an NPC stops to greet the player while wandering, when they start moving again they continue to walk toward their original destination. In OpenMW they pick a new destination.

Allofich:
Actually looks like original Morrowind does save and restore the destination if its a point on the pathgrid, but not if the actor is wandering to a destination in an area without a pathgrid.

Allofich:
New version up.

scrawl:
The change itself looks ok, but I think we've reached a point where forwards-incompatible changes to the save format can be too much of an inconvenience for the user. I'd like to mitigate such breakage for small-ish features like this one.

There could be a number of ways to handle this, e.g.:
- When running git versions, provide a warning upon (over)writing a save that the resulting save will not be compatible with the stable release. 
- Introduce reading of the relevant state in release X, then enable the writing of state in a later release X+1 (i.e. that way, as long as no one's moving by more than one version at a time, they're fine - that would mean the last two releases are always 'supported')

Any other ideas?

Allofich:
> Introduce reading of the relevant state in release X, then enable the writing of state in a later release X+1 (i.e. that way, as long as no one's moving by more than one version at a time, they're fine - that would mean the last two releases are always 'supported')

The problem with that is that it would be a long time before the feature was fully implemented, which also means less time for it to get thorough testing.

> When running git versions, provide a warning upon (over)writing a save that the resulting save will not be compatible with the stable release.

I prefer this option. People should hopefully be ready for potential problems anyway when using a git version. The warning message would probably need to be an in-game dialog, though, not just a console message, in order for everyone to notice it before saving a game. And we would need to go in and remove it whenever a release happens, which is a small inconvenience. I'm sure this isn't the only feature that needs to change the save file format, though, so if we start doing this it opens up new possibilities for fixing things.

> Any other ideas?
Not at the moment.

If you agree I'll try to add a warning message to this PR.

Allofich:
I've redone the fix, simplifying it and removing the changes to save games. This means it's not perfect, but all that is missing is that on loading a save, rather than actors resuming their old wander destination as I observed vanilla to do, a new wander point will be picked, as is the case already. NPCs exiting combat will resume their old destination. Also, stationary NPCs will properly return to their position after loading a save, which I think satisfies the bug report.

scrawl:
Seems if you stand too close to the NPC, they'll go into greeting state and abort their return path. Any way to fix that?

Allofich:
That should be fixed now.

AIWander is kind of a mess in parts and I'm interested in trying to clean it up. I know there was just a discussion about cleaning files. Not only formatting, but I think there are things like unnecessary variables in there and possibly out-of-date comments. Would it be OK to submit a PR that tries to clean it up without changing functionality?

scrawl:
> AIWander is kind of a mess in parts and I'm interested in trying to clean it up. I know there was just a discussion about cleaning files and avoiding to stop merge conflicts. It's not just formatting, but I think there are possibly things like unnecessary variables or out-of-date comments in AIWander. Would it be OK to submit a PR that tries to clean it up without changing functionality?

Unnecessary variables or out-of-date comments: sure. Even no comment is better than a misleading comment. Formatting, only if its really bad or is next to code you already changed.

PR #1280, opened 2017-04-29
Make guards that are attacked try to arrest player

Allofich:
Fixes https://bugs.openmw.org/issues/3706

### DIFF
--- a/apps/openmw/mwmechanics/aisequence.cpp
+++ b/apps/openmw/mwmechanics/aisequence.cpp
@@ -121,6 +121,16 @@ bool AiSequence::isInCombat() const
     return false;
 }
 
+bool AiSequence::hasPackage(int typeId) const
+{
+    for (std::list<AiPackage*>::const_iterator it = mPackages.begin(); it != mPackages.end(); ++it)
+    {
+        if ((*it)->getTypeId() == typeId)
+            return true;
+    }
+    return false;
+}
+
 bool AiSequence::isInCombat(const MWWorld::Ptr &actor) const
 {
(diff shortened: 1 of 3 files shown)


scrawl:
With this change, guards are often receiving more than one AIPursue package. Although this doesn't appear to create any issues with the vanilla game, it should still be avoided I think.

Allofich:
Shouldn't happen now.

scrawl:
Doesn't seem to work for me now (merged against commit 63f101e0cdfd58e0a88f7676b299a915faae4200). An attacked guard attacks back rather than arresting me.

apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -121,6 +121,16 @@ bool AiSequence::isInCombat() const
...
+bool AiSequence::isPursuing() const
+{
+    for (std::list<AiPackage*>::const_iterator it = mPackages.begin(); it != mPackages.end(); ++it)
+    {
+        if ((*it)->getTypeId() == AiPackage::TypeIdPursue)
+            return true;
+    }
+    return false;
+}
+
```
scrawl: Can we refactor this to a hasPackage(typeid) function maybe?

Allofich:
Hi scrawl, I plan to revisit this and my other open PRs before too long and hopefully get them up to speed. I'll take a look at your suggestion.

Allofich:
It should work now, and I refactored the isPursuing() to hasPackage() like you suggested. I looked at replacing calls to isInCombat() with calls to hasPackage() but there are quite a few of them and it involves including aipackage.hpp in those files, so I did not do that.

PR #1282, opened 2017-04-30
Do not apply underwater effects to player's voice

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3850


MiroslavR:
Misclick, sorry!

PR #1283, opened 2017-04-30
Don't make NPCs or player speak while underwater

Allofich:
Fixes https://bugs.openmw.org/issues/3851.

In the original engine, NPCs and the player actually won't speak when they are about waist-deep. I could have used isWading() for a knee-deep check, or created a new function for a waist-deep check, but I went for using isSwimming(). This seems to be enough to keep characters from talking while they're mouth is underwater, while still allowing them to talk when waist-deep.

### DIFF
--- a/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
+++ b/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
@@ -633,6 +633,12 @@ namespace MWDialogue
             return;
         }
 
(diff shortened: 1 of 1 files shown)


scrawl:
Would it be better to add the check in DialogueManager::say()?

Allofich:
Yes, that works fine. I was worried about whether it would still be possible to make NPCs or the player speak underwater with the "Say" scripting command, but DialogueManager::say() doesn't affect that.

Allofich:
The PR now changes DialogueManager::say()

PR #1284, opened 2017-05-02
OpenMW-CS: Highlight each variable name in a unique colour (Feature #2508)

PlutonicOverkill:
https://bugs.openmw.org/issues/2508
Adds a setting in the preferences menu to highlight each different variable name in a unique colour. The colours generated seem to work well but could be easily tweaked if improvements were needed.
(By the way, what is the convention for spelling "colour" in OpenMW? I've seen both "color" and "colour" used, but "colour" seems more common.)

### DIFF
--- a/apps/opencs/model/prefs/state.cpp
+++ b/apps/opencs/model/prefs/state.cpp
@@ -152,6 +152,7 @@ void CSMPrefs::State::declare()
         setRange (0, 10000);
     declareInt ("error-height", "Initial height of the error panel", 100).
         setRange (100, 10000);
+    declareBool ("unique-name-colour", "Highlight each variable name in a unique colour", false);
     declareSeparator();
     declareColour ("colour-int", "Highlight Colour: Integer Literals", QColor ("darkmagenta"));
     declareColour ("colour-float", "Highlight Colour: Float Literals", QColor ("magenta"));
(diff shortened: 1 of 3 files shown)


PlutonicOverkill:
Just noticed that this feature only works for variables six letters long or more. Not sure if this is something to do with the script parser or with the hash function.

PlutonicOverkill:
Fixed the problem by switching from boost::hash to qHash.

zinnschlag:
Hm, that is an obscure feature. I had to look into the source to understand what it is supposed to do. I am not convinced that this is a good idea. Using a hash of the string is clearly a no-go, because that could end up with unreadable text (foreground/background colour combinations). Even if we can make this work reliably it will just make the script look like an angry fruit salad. There are only so many different colours the human eye can easily distinguish between.

Your current code also highlights any name that is not an ID or a keyword, which makes the situation worse. I am not convinced that this feature is even possible with the framework we have.

This issue must have slipped through somehow. If it hadn't I probably would have rejected it. Unless you have any other argument in its favour, I am still inclined to do that.

PlutonicOverkill:
I didn't think it would be all that useful myself but it seems that someone wants it for whatever reason. The color combinations shouldn't be a problem as they're calculated using HSL with the hue value depending on the name, so they're all lighter colours and readability isn't a problem. Also, the only names that are highlighted are variable names and strings, and that can't be changed without modifying the parser. That said, if there isn't any demand for this feature I don't mind if it gets rejected.

PR #1285, opened 2017-05-06
Update Windows CI scripts to use ffmpeg 3.2.4

ace13:
Should fix the AppVeyor issue for now - apparently the old 3.0.1 build was archived and no longer downloadable from there.


psi29a:
Thanks for looking into this.  Cheers!

PR #1287, opened 2017-05-08
Spellbuying menu improvements

akortunov:
Difference for user:
1) No need for scrolling from begin of list after every bought spell
2) Spells in list now sorted by name

I also tried to implement this via MyGUI::ListBox, but got stuck with tooltips, see [this](https://github.com/akortunov/openmw/tree/spellslist) branch, if you are interested.

### DIFF
--- a/apps/openmw/mwgui/spellbuyingwindow.cpp
+++ b/apps/openmw/mwgui/spellbuyingwindow.cpp
@@ -22,7 +22,6 @@ namespace MWGui
 
     SpellBuyingWindow::SpellBuyingWindow() :
         WindowBase("openmw_spell_buying_window.layout")
-        , mLastPos(0)
         , mCurrentY(0)
     {
         getWidget(mCancelButton, "CancelButton");
@@ -37,13 +36,20 @@ namespace MWGui
         MWBase::Environment::get().getWindowManager()->removeGuiMode(GM_SpellBuying);
     }
 
-    void SpellBuyingWindow::addSpell(const std::string& spellId)
+    bool SpellBuyingWindow::sortSpells (const ESM::Spell* left, const ESM::Spell* right)
+    {
+        std::string leftName = Misc::StringUtils::lowerCase(left->mName);
+        std::string rightName = Misc::StringUtils::lowerCase(right->mName);
+
+        return leftName.compare(rightName) < 0;
+    }
+
+    void SpellBuyingWindow::addSpell(const ESM::Spell& spell)
     {
         const MWWorld::ESMStore &store =
             MWBase::Environment::get().getWorld()->getStore();
 
-        const ESM::Spell* spell = MWBase::Environment::get().getWorld()->getStore().get<ESM::Spell>().find(spellId);
-        int price = static_cast<int>(spell->mData.mCost*store.get<ESM::GameSetting>().find("fSpellValueMult")->getFloat());
+        int price = static_cast<int>(spell.mData.mCost*store.get<ESM::GameSetting>().find("fSpellValueMult")->getFloat());
         price = MWBase::Environment::get().getMechanicsManager()->getBarterOffer(mPtr,price,true);
 
         MWWorld::Ptr player = MWMechanics::getPlayer();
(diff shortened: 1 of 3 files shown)


Allofich:
In original Morrowind, and also in OpenMW currently, it looks like spells are organized by school. Is having them sorted alphabetically better?

My feeling is that it might be a little more convenient for someone who knows the name of the spell or spells they want to buy, but for a player who is newer to the magic system having the spells grouped by school might be better, as it puts the spells of similar function, or that they will be good or bad at casting because of their school, close to each other.

Personally I would prefer to keep the sorting by school. I do appreciate all your work, though.

akortunov:
Just checked original Morrowind. Spells sorted neither by name nor by school. In OpenMW too, but spells order in list differs from vanilla game. Tested with "Balmora, Guild of Mages", russian localization.

We can sort spells by school, by name, or combine order (school first, name last). Any suggestions?

Allofich:
I had tested with Masalinie Merian (teleport service) in the Balmora Mages Guild, whose spells are ordered the same in original Morrowind and in OpenMW, and look like they are sorted by school.

But the order of the spells sold by Sharn gra-Muzgob (orc) is a little different and Dispel is among the restoration spells, and the spells of Marayn Dren (male dark elf) are ordered very differently, and Fire Shield is apart from the other alteration spells. It does seem that in general the spells are sorted by school, but not completely. Just enough to trick me into thinking they were.

I wonder what the difference for the exceptions is. Just a guess, but I think maybe Dispel and Fire Shield are both starting spells. That might be the difference.

As for what we do in OpenMW, school first, name last sounds fine to me, but I'm not sure.

Allofich:
Or, try to figure out the sorting in original Morrowind, which we seem to already be close to, and replicate that. I'm not sure that it is really useful for anything in this case, though.

rhtucker:
I went through and analyzed several spell orders. Since there is no pattern whatsoever in terms of the name, school, or effect, I began to suspect you could set the order in the CS. Turns out it's sorted in reverse order of the lists in the CS, which is in turn decided only by the order you add the spells to the NPC. I tried editing this order, but it seems readding them is the only way to alter it. I see no reason to duplicate this behavior, and would indeed prefer sorting by school, then by name, or simply by name. It's not like half the spell names make any sense anyway. Or OpenMW could preserve the exact behavior, but in OpenCS make it possible to edit the order. I suppose there are some who like that level of control over their modding.

Thunderforge:
I would be in favor of not replicating the vanilla Morrowind order, since it seems nonsensical. Ideally, I'd like for a way to switch the ordering, perhaps a dropdown menu called "Sort Order" that lets you pick alphabetically, by school, and perhaps by cost or magicka needed. 

If that's an impossibility, then I would be in favor of either school then name or alphabetically for all, with a slight preference towards the former.

akortunov:
> If that's an impossibility

It is possible, but we do not have GMSTs to localize such dropdown menu.

Thunderforge:
@akortunov What does GMST stand for? [Acronym Finder isn't much help](http://www.acronymfinder.com/GMST.html).

akortunov:
> What does GMST stand for?

"Game Setting (GMST) records are entries that contain game information required by Morrowind. Some of them contain numerical values for armor and weapons, movement and abilites, game data relating to view distance and gamma settings, while others simply contain the names of skills or spells."

Morrowind uses GMSTs to store GUI strings too.
See [this](https://forum.openmw.org/viewtopic.php?f=6&t=4300) forum topic about GUI localization.

kcat:
GMST is a Game State record in the game data files. It's a const/read-only value defined by the esm/esp that the engine uses for a particular reason (such as "Ok" or "Cancel" button label strings).

akortunov:
Rebased.

I just noticed that setTextColour() prevent focused widget highlighting. Is there any way to avoid this behaviour?

akortunov:
I uploaded a simplified implementation. Only position saving and spells sorting applied.

scrawl:
To be honest, the sorting to me seems like too minor of a thing to warrant an excursion from the vanilla behavior.

> I have just noticed that setTextColour() prevents focused widget highlighting. Is there any way to avoid this behaviour?

No. By setting one color explicitely, you're not going to get different colors for highlighting. You would need to use widget states instead.

Thunderforge:
> To be honest, the sorting to me seems like too minor of a thing to warrant an excursion from the vanilla behavior.

It may be minor, but I think it would be a welcome change that would make the game just a bit more enjoyable to play.

akortunov:
OpenMW is already sorting spells in order different from vanilla game.

scrawl:
> OpenMW is already sorting spells in order different from vanilla game.

Fair enough then.

PR #1288, opened 2017-05-08
Fixed Travis CI build (removed tab from test_stringops.cpp)

akortunov:
Just that.


psi29a:
Thanks for spotting that.

PR #1289, opened 2017-05-08
Improved repair and recharge widgets

akortunov:
See discussion and screenshots in [this](https://forum.openmw.org/viewtopic.php?f=6&t=4269) forum topic.


scrawl:
The untranslated string 'Repair tools' is a problem. How about using 'sSelect' or 'sRepair' ?

akortunov:
Modified to use sRepair GMST. After 1.0 release we need to implement a flexible localization system anyway.

PR #1290, opened 2017-05-09
Added hotkey for local/world map switching

mrkkrp:
I am not the author of the PR, I requested this featrue here: https://forum.openmw.org/viewtopic.php?f=3&t=4290, and @akortunov implemented it, but he did not open a PR for it, so I decided I'll try to do it and maybe it'll get through.

I tested the feature, it's a great time saver.

### DIFF
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -276,6 +276,7 @@ namespace MWBase
 
             virtual bool getPlayerSleeping() = 0;
             virtual void wakeUpPlayer() = 0;
+            virtual void toggleMap() = 0;
 
             virtual void showCompanionWindow(MWWorld::Ptr actor) = 0;
             virtual void startSpellMaking(MWWorld::Ptr actor) = 0;
--- a/apps/openmw/mwgui/mapwindow.cpp
+++ b/apps/openmw/mwgui/mapwindow.cpp
@@ -917,17 +917,7 @@ namespace MWGui
 
     void MapWindow::onWorldButtonClicked(MyGUI::Widget* _sender)
     {
-        mGlobal = !mGlobal;
-        mGlobalMap->setVisible(mGlobal);
-        mLocalMap->setVisible(!mGlobal);
-
-        Settings::Manager::setBool("global", "Map", mGlobal);
-
-        mButton->setCaptionWithReplacing( mGlobal ? "#{sLocal}" :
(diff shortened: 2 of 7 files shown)


mrkkrp:
Is this ignored on purpose? Should I request this feature of issue tracker first or what?

scrawl:
I'm not in favor of this:
- Because this hotkey isn't in the original game, it will show as untranslated (I know some other settings already do, but we should try not to add too many more)
- When the map window isn't visible or pinned, this hotkey is a no-op which may lead to confusion.
- Adding hotkeys for minor GUI actions seems like a recipe for a bloated menu.

mrkkrp:
> Because this hotkey isn't in the original game, it will show as untranslated (I know some other settings already do, but we should try not to add too many more)

Valid point, but if we to make something more powerful and more handy than vanilla (which I understand we want to), this is inevitable. Proper solution is probably to add a way to localize custom things in OpenMW itself?

> When the map window isn't visible or pinned, this hotkey is a no-op which may lead to confusion.

I don't see how this may lead to confusion, if you don't have a map, you just don't see the effect of switching...

> Adding hotkeys for minor GUI actions seems like a recipe for a bloated menu.

Can only say that this one is incredibly handy.

OK, I'll keep rebasing the change against master and play that version, because I find playing without this feature annoying now.

Thunderforge:
I would be in favor of seeing this change, as it does seem incredibly handy. OpenMW is getting to the point where it is superior to the vanilla engine, and a feature like this would be a great timesaver. 

Yes, we should have a way to localize new things in OpenMW, but that seems like something we'll already need to do, and I don't see that as a reason to reject this improvement.

Greendogo:
> Proper solution is probably to add a way to localize custom things in OpenMW itself?

> Yes, we should have a way to localize new things in OpenMW, but that seems like something we'll already need to do, and I don't see that as a reason to reject this improvement.

Another alternative is to wait on approving this PR until such time that a development strategy for features needing localization is implemented.

scrawl:
Too minor to warrant a new option.

A more generic way to bind any GUI button/action to any key would be better, though I'm not too sure in what form that should be implemented.

scrawl:
For what it's worth, #1481 adds a keyboard-only way to do this, though it could be more convenient. From game mode, it's `Menu Key`, `Space`, `Menu Key` (you'll also need to press `Tab` once to focus the world map button, unless you've already clicked on it)

PR #1291, opened 2017-05-13
Prevent respawned references from being added to the scene twice

MiroslavR:
Fixes: https://bugs.openmw.org/issues/3864

Reproduction steps:
- Suppose there is a respawnable guard in Gnaar Mok.
- Move him outside of his original cell to a faraway cell, relative to which the original cell is inactive, and kill him there.
- Unload the cell where the guard died (e.g. by entering some interior).
- Wait for the guard to respawn.
- COC to Gnaar Mok to load the original cell.

What happens afterwards is that the guard is added to the scene twice - once by the respawn routine (which calls `MWWorld::moveObject` - moving from an inactive cell to an active cell adds the guard to the scene), and then by `InsertVisitor`. Now if a reference is added to the scene twice, certain oddities may happen such as the crash described in the bug report:

```c++
void Objects::insertNPC(const MWWorld::Ptr &ptr)
{
    insertBegin(ptr);
    ptr.getRefData().getBaseNode()->setNodeMask(Mask_Actor);

    osg::ref_ptr<NpcAnimation> anim (new NpcAnimation(ptr, osg::ref_ptr<osg::Group>(ptr.getRefData().getBaseNode()), mResourceSystem));

    ptr.getClass().getInventoryStore(ptr).setInvListener(anim.get(), ptr);
    ptr.getClass().getInventoryStore(ptr).setContListener(anim.get());

    mObjects.insert(std::make_pair(ptr, anim));
}
```

If insertNPC is called the second time, `anim` is unref'd upon leaving the function scope, as `mObjects.insert` is no-op if the `ptr` key is already present. The listeners will then point to a deleted object.

### DIFF
--- a/apps/openmw/mwrender/objects.cpp
+++ b/apps/openmw/mwrender/objects.cpp
@@ -38,6 +38,8 @@ Objects::~Objects()
 
 void Objects::insertBegin(const MWWorld::Ptr& ptr)
 {
+    assert(mObjects.find(ptr) == mObjects.end());
+
     osg::ref_ptr<osg::Group> cellnode;
 
     CellMap::iterator found = mCellSceneNodes.find(ptr.getCell());
@@ -90,9 +92,8 @@ void Objects::insertCreature(const MWWorld::Ptr &ptr, const std::string &mesh, b
     else
         anim = new CreatureAnimation(ptr, mesh, mResourceSystem);
 
-    ptr.getClass().getContainerStore(ptr).setContListener(static_cast<ActorAnimation*>(anim.get()));
-
-    mObjects.insert(std::make_pair(ptr, anim));
+    if (mObjects.insert(std::make_pair(ptr, anim)).second)
+        ptr.getClass().getContainerStore(ptr).setContListener(static_cast<ActorAnimation*>(anim.get()));
 }
 
 void Objects::insertNPC(const MWWorld::Ptr &ptr)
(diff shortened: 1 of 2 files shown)


scrawl:
I remember debugging a similar issue a while back. Maybe we should not mark the cell as active until we're finished inserting it? Also, can we improve insertNPC so that it doesn't crash when called twice?

MiroslavR:
Done, although I can't tell with 100% certainty that none of the calls in `Scene::loadCell` depend on the cell being already marked as active.

scrawl:
> Maybe we should not mark the cell as active until we're finished inserting it? 

Oops... Actually, scratch that. If there is an exception mid-loading, the cell needs to be marked active so that what was already loaded can be unloaded.

MiroslavR:
Can you provide an example of such an exception occurring and where it is handled? I don't see it.

scrawl:
It would be handled as a failure to load a game / start a new game (e.g. [here](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwstate/statemanagerimp.cpp#L535)) and drop you back to the main menu.

I don't have an example. In an ideal world, faulty content files would not cause a cell to fail loading, but bugs can always happen (and certainly have in the past).

MiroslavR:
Fair enough. I have reverted to my previous fix.

scrawl:
Hm, that won't work properly, I think. The logic e.g. in Npc::respawn() doesn't actually add the NPC to the scene.

MiroslavR:
> Hm, that won't work properly, I think. The logic e.g. in Npc::respawn() doesn't actually add the NPC to the scene.

Which case won't work? If the respawning NPC is currently residing in:
- the cell that is being loaded, it will be added to the scene via `insertCell`.
- an inactive cell, it will be added to the scene via `Npc::respawn`->`World::moveObject` .(see https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwworld/worldimp.cpp#L1178).

scrawl:
Case:
- Actor is disposed of (count = 0) in its original cell. The player enters this cell.
- insertCell does not add this actor to the scene due to count = 0.
- respawn() sets the count back to 1 but does not add to the scene.

MiroslavR:
I see, will look for another solution.

MiroslavR:
Haven't had much time to look into this, unfortunately. I am freeing this so other people can work on it if they wish.

MiroslavR:
> Oops... Actually, scratch that. If there is an exception mid-loading, the cell needs to be marked active so that what was already loaded can be unloaded.

How about a RAII object that marks the cell active upon destruction/leaving scope?

scrawl:
Ok, good idea. That, or catch the exception, mark cell as active then re-throw.

MiroslavR:
This is ready for review now.

scrawl:
I think this will break levelled creature spawning (i.e. a newly spawned creature won't be added to the scene).

MiroslavR:
Right, `CreatureLevList::insertObjectRendering()` requires that the cell being loaded is already marked as active.

I don't see any other way to fix this issue other than not allowing to add objects to the scene twice, which would not fix it at its core.

scrawl:
Can we in the meantime add a workaround for the object being added twice (e.g. logging a warning)? Anything is better than crashing.

MiroslavR:
This should do for the time being.

PR #1292, opened 2017-05-15
Converted all settings files to semantic linefeeds.

rhtucker:
Also:
Corrected default segments in Windows settings.
Removed all default declarations in the prose of the setting descriptions except where clarification or additional information was also provided.
Some minor, but not thorough grammar corrections.


psi29a:
Had a quick read through and I approve! :)

PR #1295, opened 2017-05-20
Abilities affecting attributes add to base value rather than modified value

raglnd:
  - [http://bugs.openmw.org/issues/1751](http://bugs.openmw.org/issues/1751).
- Summary of the changes made
  - Modified the birthsign part of the character creation code to add attribute fortification to the base value of the attribute rather than the modified value. Also changed Spells::rebuildEffects() to give a magnitude of 0 for abilities that modify attributes so that the ability isn't double counted.
  - Original engine did not count birthsigns like The Lady and The Steed as normal fortify spells, but rather as an actual increase to the character's attributes. This change duplicates the original behavior. 
  - Went through character creation and verified that The Lady and the Steed functioned as in the original engine - values are increased without being highlighted, and the fortify attribute effect still shows up in the effects list.
  - Did not check to see if this change interfered with effects from corprus, or anything else that might have been counted as an ability, rather than a spell.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -1733,6 +1733,7 @@ namespace MWMechanics
 
     void Actors::updateMagicEffects(const MWWorld::Ptr &ptr)
     {
+        ptr.getClass().getCreatureStats(ptr).updateAbilityAttributeState();
         adjustMagicEffects(ptr);
         calculateCreatureStatModifiers(ptr, 0.f);
         if (ptr.getClass().isNpc())
--- a/apps/openmw/mwmechanics/creaturestats.cpp
+++ b/apps/openmw/mwmechanics/creaturestats.cpp
@@ -668,4 +668,40 @@ namespace MWMechanics
     {
         return mSummonGraveyard;
     }
+
+    void CreatureStats::updateAbilityAttributeState()
+    {
+        // Make abilities affect base attribute.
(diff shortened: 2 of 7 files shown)


scrawl:
What about abilities not originating from a birthsign? Looks like they would not function correctly with this change. Not sure if these exist in the vanilla game, but its easy to add one with scripting.

raglnd:
They would not function with this change. 

Should those be implemented as affecting the base value or as affecting the modification value?

raglnd:
Most recent commit allows for abilities not originating from a birthsign to function correctly.

scrawl:
> Should those be implemented as affecting the base value or as affecting the modification value?

I don't know. Can you check in the vanilla engine?

raglnd:
Looks like abilities affecting attributes add to the base value - checked using Gaenor's Abilities.

Will need to find a way to implement that instead of what is currently here.

raglnd:
Now works like the vanilla engine.

raglnd:
Loading a save game ignores setting the magnitude to 0. Need to investigate why. (Double counting the ability.)

raglnd:
Fixed issue when saving and loading the game. Now works as expected.

scrawl:
I don't feel comfortable merging this PR due to the extent of the changes and how awkwardly they are tacked on top of the existing code. Maybe we need to take a step back and rethink our whole attribute system but I could not tell you how to do it, short of me diving in and basically doing it all myself.

In any case, thanks for your work. I have added a link to this PR to the bug report for future reference.

PR #1296, opened 2017-05-21
Fixed jump fatigue loss formula (bug #3843)

akortunov:
See [bug #3843](https://bugs.openmw.org/issues/3843) for description.
Note: we also need to change formula on wiki page.


MiroslavR:
I took a look at the assembly and can confirm this.

PR #1297, opened 2017-05-21
 MWWorld::Cells: make cache size configurable

akortunov:
This PR consists of two parts:
1) Makes mIdCache size in cells.cpp configurable (implements TODO from cells.cpp).
2) Makes World::searchPtr() check cache first when activeOnly = false.

In current OpenMW master searchPtr() first iterates through active cells (slow), only then checks cache and inactive cells.

This PR also allows to workaround [bug #3242](https://bugs.openmw.org/issues/3242) with
`pointers cache size = 400`

EDIT: second part was reverted.

### DIFF
--- a/apps/openmw/mwworld/cells.cpp
+++ b/apps/openmw/mwworld/cells.cpp
@@ -7,6 +7,7 @@
 #include <components/esm/defs.hpp>
 #include <components/esm/cellstate.hpp>
 #include <components/loadinglistener/loadinglistener.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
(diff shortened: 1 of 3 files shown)


zinnschlag:
The first part is obviously okay. But I see a problem with the second.

When checking all cells (activeOnly) previously the active cells were still given priority. The new code will use what ever is in the cache, no matter if active or not. This is a change in behaviour and it may break things.

akortunov:
> When checking all cells (activeOnly) previously the active cells were still given priority.

Got it. If the vanilla game works in the same way, we need to find an another solution.

By the way, [this](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwworld/cells.cpp#L178) line is a bottleneck.

CellStore::search() does a linear search in std::vector<LiveCellRefBase*>, right? Can we refactor this to use a binary search?

zinnschlag:
> CellStore::search() does a linear search in std::vector<LiveCellRefBase*>, right? Can we refactor this to use a binary search?

I don't see how since that would require sorting first and the order of references actually matters.

I am not even sure if there is any need for improvement. So far we have only a single case where this bottleneck is a problem and this plugin is badly written. It needs to be fixed. Even the current cache is only there because there is a equally broken script in Morrowind.esm and we obviously can not tell people to fix their script in this case. But generally we are under no obligation to make junk scripts run fast.

If you really want to go all out on this task you could do a separate cache check at the beginning that ignores cached IDs in cells that are currently not active.

akortunov:
> I am not even sure if there is any need for improvement. 

In this case we can take only first part (configurable size) of PR and reject a bug. And player can play with larger cache as temporary workaround until this plugin will be fixed (fps is low, but still playable).

zinnschlag:
Sounds reasonable to me.

akortunov:
The second part of PR is reverted.

PR #1298, opened 2017-05-23
Fix for merchants not overstocking

raglnd:
  - [https://bugs.openmw.org/issues/2473](https://bugs.openmw.org/issues/2473)
- Summary of the changes made
  - Reworked how MWWorld::ContainerStore::restock() functions. Overstocking merchants now mostly works like it does in vanilla.
  - Current issues.
    - Unsure if the code removed is doing something with containers that I did not notice - I did not know what to look for to verify one way or the other.
    - Some merchants still sell the same item in two stacks to begin with, rather than in one stack. Example: Nalcarya of White Haven sells 2 stacks of 3 diamonds, rather than 1 stack of 6 diamonds. (Only one is initially restockable.)
    - The wiki says that there should be an option for vanilla behavior and for current OpenMW behavior, but I don't know how to implement options like that. If there is something that I could look at, I can implement this as an option.
    - Comments in code are not finalized - some are leftover from the older code.
  - Brings results of buying/selling closer to vanilla functionality. As an aside, all this should allow for is saving users from buying restockable items 5 at a time.
  - Bought and sold items from Ajira and Nalcarya, while comparing the results to what happened in the vanilla engine.

### DIFF
--- a/apps/openmw/mwworld/containerstore.cpp
+++ b/apps/openmw/mwworld/containerstore.cpp
@@ -580,7 +580,16 @@ void MWWorld::ContainerStore::restock (const ESM::InventoryList& items, const MW
             //Restocking static item - just restock to the max count
             int currentCount = count(itemOrList);
             if (currentCount < std::abs(it->mCount))
+            {
                 addInitialItem(itemOrList, owner, -(std::abs(it->mCount) - currentCount), true);
+            }
+            // Don't need to restock, but should update the amount to be restocked.
+            else if (currentCount > std::abs(it->mCount))
+            {
+                // Removing const here. Unavoidable without major reworks, but mCount was never itself declared const.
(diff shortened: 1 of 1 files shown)


raglnd:
When combined with pull request #1299 , vanilla trading implemented in OpenMW. Separated into two pull request because they fix different things.

Still not sure how to do this as an option, but it should be relatively easy to either allow or disallow this behavior.

Capostrophic:
@raglnd, please, register on bugtracker and if you already did tell me your name there so I can assign the relevant bug reports (2473 and 2842) to you.

raglnd:
I have signed up on the big tracker as raglnd

scrawl:
The const is there for a reason - you can not modify the ESM records because the changes won't be retained when you save/load the game.

PR #1299, opened 2017-05-24
Fix multiple stacks of same item

raglnd:
  - [https://bugs.openmw.org/issues/2473](https://bugs.openmw.org/issues/2473)
- Summary of the changes made
  - Cleaned up MWWorld::ContainerStore::stacks() and made it not return false when one item had an owner, and the second item did not. Result is a much cleaner trade window.
  - Example: Nalcarya of White Haven sold 2 stacks of 3 diamonds, rather than 1 stack of 6 diamonds. This has now been fixed so that she sells 1 stack of 6 diamonds.
  - Original engine had only one stack per item, and brings the correct behavior to OpenMW.
  - Bought and sold items from Ajira and Nalcarya in Balmora, while comparing the results to what happened in the vanilla engine.

### DIFF
--- a/apps/openmw/mwworld/containerstore.cpp
+++ b/apps/openmw/mwworld/containerstore.cpp
@@ -211,36 +211,67 @@ MWWorld::ContainerStoreIterator MWWorld::ContainerStore::restack(const MWWorld::
 
 bool MWWorld::ContainerStore::stacks(const ConstPtr& ptr1, const ConstPtr& ptr2) const
 {
+    // an item never stacks onto itself
+    if (ptr1 == ptr2)
+        return false;
+
+    // check IDs
+    const MWWorld::CellRef& cRef1 = ptr1.getCellRef();
+    const MWWorld::CellRef& cRef2 = ptr2.getCellRef();
+
+    // must have same ref ID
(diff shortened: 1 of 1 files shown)


xirsoi:
I think this behavior is intended, because it prevents the stack without an owner from being "corrupted", and given the same owner.

This was an issue in Morrowind that made thieving a lot harder than it needed to be.

raglnd:
Can you provide an example of this corruption?

I checked the original engine and all of the diamonds are owned by Nalcarya (3 of them are not specifically owned by her until you take them from the chest she owns upstairs).

Additionally, I do not see how this change would give the stack without an owner the same owner as the other stack.

xirsoi:
If you have an object that is unowned and merge it with a stack of items that are owned by an npc, the entire stack is regarded as stolen property by guards, in the base game anyway.

If your implementation will not cause that behavior, awesome 👍

raglnd:
Just verified that this does not cause the entire stack to be regarded as stolen property.

scrawl:
stacks() actually does affect the internal storage (see e.g. ContainerStore::add), so by returning true for different owners will cause the items to be merged internally with one of the owners being lost. We want to merge on the display level only.

PR #1305, opened 2017-06-01
Show success chance when self-enchanting

akortunov:
Implements [feature #3492](https://bugs.openmw.org/issues/3492).

### DIFF
--- a/apps/openmw/mwgui/enchantingdialog.cpp
+++ b/apps/openmw/mwgui/enchantingdialog.cpp
@@ -8,6 +8,7 @@
 #include <MyGUI_ScrollView.h>
 
 #include <components/widgets/list.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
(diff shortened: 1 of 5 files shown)


scrawl:
Alteration of vanilla features need to be made optional and disabled by default.

akortunov:
I made this feature optional.

Capostrophic:
> Show or not show a probability of success will be shown in enchantment menu.

==> Whether or not the chance of success will be displayed in the enchanting menu.

akortunov:
Thanks. I modified the .rst-file.

scrawl:
At least for the english game, sEnchantmentMenu6 seems to be a better label (Chance vs. Spell Chance).

What's the point of a tooltip that just repeats the label?

akortunov:
I can remove tooltip, if it isn't necessary.
About label: sEnchantmentMenu6 is ok, just sSpellmakingMenu1 localized as "Success chance" in russian version, that misled me.

akortunov:
The label is changed, the tooltip is removed, three commits rebased to one.

mrkkrp:
I like this one, but there seems to be no consistency in categorizing of the settings. For example, `show melee info` and `show projectile damage` go in the `Game` section, as well as `show effect duration`. I think you would agree that these features are quite similar to the one this PR implements. Yet `show enchant chance` was placed in the `GUI` section.

akortunov:
Which category would be better for these three settings?

mrkkrp:
It looks like the GUI section is about purely visual things, while the Game section includes things that affect gameplay itself (such as whether to highlight items that belong to some NPC or give additional information about projectile damage, etc.). So I think we should just move `show enchant chance` to the Game section (because it influences gameplay, i.e. it's not purely visual setting) and it'll be OK.

akortunov:
> we should just move show enchant chance to the Game section

@scrawl, should we do it?

scrawl:
Makes sense to me.

PR #1306, opened 2017-06-01
Fixed quickload with menus open

Nialsy:
[Bug #3879](https://bugs.openmw.org/issues/3879)

Summary: Added a check to see if the guiMode was false. 

Testing: Attempted quickloading in various menus and in the situation mentioned.

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -309,6 +309,7 @@ namespace MWWorld
 
         mDoorStates.clear();
 
+        mGoToJail = false;
         mTeleportEnabled = true;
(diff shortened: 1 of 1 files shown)


scrawl:
I'd prefer that we fixed the issue with the goToJail flag not being cleared instead.

Nialsy:
I felt the other way was more in line with how vanilla morrowind did things. Not too crazy with all the additions for something so simple, but I didn't feel it would fit being cleared in the functions used to load the game.

scrawl:
The best place is World::clear(). Also, can you do a git rebase to remove the commits for the previous solutions?

Nialsy:
Cleared the flag in World::clear() and rebased the commits.

PR #1308, opened 2017-06-03
Spell priority: dispel usage improvements

akortunov:
This PR:
1) Allows combat AI not only to dispel negative spells from self, but also to dispel positive effects from an enemy.
2) Takes in account both numbers of negative and positive effects.
For example NPC will not dispel one negative effect if a spell also dispels five positive effects.
3) Takes in account spell magnitude

Note: default target "purge magic" spell has only 5% chance of success and AI will use it only when the enemy has a large amount of positive effects (about 10-15). AI will use 50% chance spells when the enemy has 2-3 positive effects.

Golden saints will use dispel only when they have only two and more negative effects (see [bug #3441](https://bugs.openmw.org/issues/3441)).

### DIFF
--- a/apps/openmw/mwmechanics/aicombataction.cpp
+++ b/apps/openmw/mwmechanics/aicombataction.cpp
@@ -61,7 +61,7 @@ float suggestCombatRange(int rangeTypes)
     }
 }
 
-int numEffectsToCure (const MWWorld::Ptr& actor, int effectFilter=-1)
+int numEffectsToDispel (const MWWorld::Ptr& actor, int effectFilter=-1, bool negative = true)
 {
     int toCure=0;
     const MWMechanics::ActiveSpells& activeSpells = actor.getClass().getCreatureStats(actor).getActiveSpells();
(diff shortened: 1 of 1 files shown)


apps/openmw/mwmechanics/aicombataction.cpp
```diff
@@ -400,14 +405,49 @@ namespace MWMechanics
...
+                diff = numPositive - numNegative;
+
+                // if rating < 0 here, the spell will be considered as negative later
+                rating *= -1;
+            }
+
+            if (diff <= 0)
+                return 0.f;
+
+            int magnitude = (effect.mMagnMin + effect.mMagnMax) / 2;
```
scrawl: apps/openmw/mwmechanics/aicombataction.cpp:438:17: warning: unused variable ‘magnitude’ [-Wunused-variable]
             int magnitude = (effect.mMagnMin + effect.mMagnMax) / 2;

akortunov:
Sorry about unused variable. Removed.

PR #1310, opened 2017-06-04
Removed empty space before dialogue entry (bug #3892)

akortunov:
See bug [description](https://bugs.openmw.org/issues/3892).


scrawl:
Shouldn't 'choice' result in a new paragraph?

choice_2

akortunov:
Yes, you are right. It seems Morrowind just handles responses differently from greetings. I will try to fix it.

PR #1311, opened 2017-06-05
Pseudo random

uramer:
std::rand() is known for very poor value distribution, and it shows in gameplay (I routinely succeed 15 times in a row with a 64% cast chance spell or fail 4 times with a 95% one in my current playthrough). This a rather primitive implementation of pseudo random, but it should work reasonably well to at least make the most ridiculous coincidences appear less.
I scanned through all the uses of Misc::Rng and haven't found anything that could break from such a change (I can't think of anything that could rely on silly std::rand() distribution anyway), but if anyone more familiar with the project has any concerns that I should more thoroughly test, please let me know.

### DIFF
--- a/components/files/linuxpath.cpp
+++ b/components/files/linuxpath.cpp
@@ -3,6 +3,7 @@
 #if defined(__linux__) || defined(__FreeBSD__) || defined(__FreeBSD_kernel__) || defined(__OpenBSD__)
 
 #include <pwd.h>
+#include <unistd.h>
 #include <boost/filesystem/fstream.hpp>
 
 #include <components/misc/stringops.hpp>


akortunov:
I think we should do a test with large number of dices (100000 or more). Random numbers generation is crucial since a game mechanics is dice-based.

HumanG33k:
Maybe you can use /dev/urandom /dev/random. 
https://unix.stackexchange.com/questions/324209/when-to-use-dev-random-vs-dev-urandom

https://stackoverflow.com/questions/11216885/getting-a-float-from-dev-urandom

akortunov:
Windows does not have /dev/random.

uramer:
I'm running some tests, and my selection of a distribution function is indeed terrible, I'll use something else.
And yes, anything involving /dev/random will not be cross-platform, so ultimately pointless.

uramer:
Would be nice to use new and fancy functions from C++11, but AFAIK there is no complete agreement in OpenMW team about moving on to that standard yet.

akortunov:
Can we use a [shuffle bags](https://gamedevelopment.tutsplus.com/tutorials/shuffle-bags-making-random-feel-more-random--gamedev-1249)? Also see [this](http://archive.gamedev.net/archive/reference/design/features/randomness/) article.

The main idea is use "card deck" instead of dice.

uramer:
well, that kind of technique will actually give even more extreme "lucky/unlucky streaks"
when I say distribution in this context, I don't literally mean "how many 0s and 1s" we get, but rather how often similar values are next to eachother. In fact, std::rand() has almost ideal distribution in terms of how equally probably numbers are, but what I'm trying to achieve is similar to what Dota 2 did: the more succeed, the less likely you are to succeed going forwards, and vice versa. overall it provides more predictable gameplay: if you have a 66% chance to cast a spell, you  are very likely to cast it twice in 3 tries, and almost impossible to fail all 3, with the overall chance staying 66%

akortunov:
> the more succeed, the less likely you are to succeed going forwards

How do you tell to generator that it was "success"?

About Dota 2: do you mean solution referenced in [this](http://wiki.teamliquid.net/dota2/Pseudo_Random_Distribution) article?
It seems Dota 2 modifies chance of activation for every ability separately instead of using a single random generator.

uramer:
That's the point, I don't think it's a good idea to try and change every mechanic in the game to support pseudo-randomness.
And this is why I didn't go with Dota2's implementations: instead of counting successes/fails I increase random values after low values and decrease values after high values.
I've ran some tests, this is what the current commit does (on million values).
As you can see, it doesn't really change the average value (by less than 1%, and that will compensate over time and with different rand() seeds). It decreases the amount of 0.1-0.2 and 0.8-0.9 values by increasing the amount of 0-0.1 and 0.9-1.0 values. I don't like that, as it effectively changes cast chances of some spells and other effects. I should probably go back to a linear function over a cubic one.
[log.txt](https://github.com/OpenMW/openmw/files/1052092/log.txt)

akortunov:
I do not see how it may to help to player ingame since a game has a single random generator for every purpose:
Some values from your log with possible actions (just example):
0.388165; Player casts a spell with 50% success. Fail.
0.687613; Enemy1 casts a spell with 10% success. Success.
0.81402; Enemy2 casts a spell with 20% success. Success.
0.428449; Player casts a spell with 50% success. Fail.
0.90051; Enemy1 casts a spell with 10% success. Success. Player throws a keyboard into the window.

uramer:
Yeah, I guess.
I still think it's better for technical purposes, but probably ultimately not worth the change.

psi29a:
Heads up, C++11 branch is ready for merge. It just needs another review (and merge) and OpenMW will be a C++11 project.  We can take advantage of this for example:
http://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution

uramer:
Great to hear! That would be optimal, and maybe at some point I (or someone else) will bother with proper pseudo random on per-function (or, rather, player-only per-action) basis.

psi29a:
Feel free to use my branch: https://github.com/OpenMW/openmw/pull/1017

You can base your solution off my branch and wait for my branch to be merged.

PR #1312, opened 2017-06-08
Split greeting from topic response

akortunov:
Fixes a regression in [PR1310](https://github.com/OpenMW/openmw/pull/1310).

The main idea is splitting greating records from response records since Morrowind handles these record types differently.

### DIFF
--- a/apps/openmw/mwgui/dialogue.cpp
+++ b/apps/openmw/mwgui/dialogue.cpp
@@ -107,23 +107,22 @@ namespace MWGui
 
     // --------------------------------------------------------------------------------------------------
 
-    Response::Response(const std::string &text, const std::string &title)
-        : mTitle(title)
+    Response::Response(const std::string &text, const std::string &title, bool needMargin)
+        : mTitle(title), mNeedMargin(needMargin)
     {
         mText = text;
     }
 
     void Response::write(BookTypesetter::Ptr typesetter, KeywordSearchT* keywordSearch, std::map<std::string, Link*>& topicLinks) const
     {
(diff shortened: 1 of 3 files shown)


scrawl:
A new class seems overkill. I would just add a bool variable to the existing Response class.

apps/openmw/mwgui/dialogue.cpp
```diff
@@ -201,17 +198,37 @@ namespace MWGui
...
+
+    Response::Response(const std::string &text, const std::string &title)
+        : mTitle(title)
+    {
+        mText = text;
+    }
+
+    void Response::write(BookTypesetter::Ptr typesetter, KeywordSearchT* keywordSearch, std::map<std::string, Link*>& topicLinks) const
+    {
+        BookTypesetter::Style* title = typesetter->createStyle("", getDialogueTextColour("header"));
```
scrawl: While you're at it, this line should go inside the ```if (mTitle != "")``` block, shouldn't it?

scrawl:
I wonder what would happen if a script runs forceGreeting while the user is in, or after, a choice?

akortunov:
> I wonder what is supposed to happen if a script runs forceGreeting while the user is in, or after, a choice?

**After choice**: new greeting displayed after last response without empty line. (According to 
[rot tor](https://bugs.openmw.org/issues/3892) some mods use this technique).

**Inside choice**: 
Morrowind: new greeting displayed after red choices, but player still has to click to one of them.
OpenMW: red choice disappeared, new greeting displayed after last response.
I do not know how we should handle this case.

> just add a bool variable to the existing Response class.

Sounds ok for me.

PR #1313, opened 2017-06-09
Make combat AI to do not cast target spells under water

akortunov:
In current implementation NPCs try to use target spells underwater and spells obviously fail. Fixed.

Also sorry for offtoping, but see [this](https://github.com/OpenMW/openmw/blob/openmw-42/apps/openmw/mwmechanics/aicombataction.cpp#L828) line. Why are we check mArea when we should check mRange? Backward compatibility or a bug? Also getMaxAttackDistance() is affected.


MiroslavR:
> Backward compatibility or a bug?

Bug, it should be mRange.

akortunov:
Since magic projectiles collide with water, added a check for enemy too.

PR #1314, opened 2017-06-09
Deboosting

psi29a:
This is an attempt to de-boost the code by using std::* equivalents. This has the effect of decreasing compile times without lose of functionality.

### DIFF
--- a/components/fallback/fallback.cpp
+++ b/components/fallback/fallback.cpp
@@ -1,9 +1,11 @@
 #include "fallback.hpp"
 
-#include <boost/lexical_cast.hpp>
-
 namespace Fallback
 {
+    bool stob(std::string const& s) {
+        return s != "0";
+    }
+
     Map::Map(const std::map<std::string,std::string>& fallback):mFallbackMap(fallback)
     {}
 
@@ -22,7 +24,7 @@ namespace Fallback
         if(fallback.empty())
             return 0;
         else
-            return boost::lexical_cast<float>(fallback);
+            return std::stof(fallback);
     }
     int Map::getFallbackInt(const std::string& fall) const
     {
@@ -30,7 +32,7 @@ namespace Fallback
         if(fallback.empty())
             return 0;
         else
-            return boost::lexical_cast<int>(fallback);
+            return std::stoi(fallback);
     }
 
     bool Map::getFallbackBool(const std::string& fall) const
@@ -39,7 +41,7 @@ namespace Fallback
         if(fallback.empty())
             return false;
         else
-            return boost::lexical_cast<bool>(fallback);
+            return stob(fallback);
     }
     osg::Vec4f Map::getFallbackColour(const std::string& fall) const
     {
@@ -55,7 +57,7 @@ namespace Fallback
                 else if (sum[i] != ' ') ret[j]+=sum[i];
             }
 
-            return osg::Vec4f(boost::lexical_cast<int>(ret[0])/255.f,boost::lexical_cast<int>(ret[1])/255.f,boost::lexical_cast<int>(ret[2])/255.f, 1.f);
+            return osg::Vec4f(std::stoi(ret[0])/255.f,std::stoi(ret[1])/255.f,std::stoi(ret[2])/255.f, 1.f);
         }
     }
 
--- a/components/sceneutil/lightmanager.cpp
+++ b/components/sceneutil/lightmanager.cpp
@@ -1,14 +1,8 @@
 #include "lightmanager.hpp"
 
-#include <stdexcept>
-
-#include <osg/NodeVisitor>
-
 #include <osgUtil/CullVisitor>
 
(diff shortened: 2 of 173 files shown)


Allofich:
Hope this works out. It's nice to see a PR that removes more lines than it adds.

psi29a:
Me too. ;)

I did my best to test the functionality that I replaced to make sure the std::* functions work as intended. However more eyes on this branch, not just code but gameplay, that things play normally.

Allofich:
That was quick. Do you plan to phase out Boost usage until it's all out?

components/widgets/numericeditbox.cpp
```diff
@@ -30,15 +28,15 @@ namespace Gui
...
+            mValue = std::stoi(newCaption);
             int capped = std::min(mMaxValue, std::max(mValue, mMinValue));
             if (capped != mValue)
             {
                 mValue = capped;
                 setCaption(MyGUI::utility::toString(mValue));
             }
         }
-        catch (boost::bad_lexical_cast&)
+        catch (...)
```
scrawl: Can we avoid using catch-all?

components/sceneutil/lightmanager.cpp
```diff
@@ -220,7 +214,8 @@ namespace SceneUtil
         // possible optimization: return a StateSet containing all requested lights plus some extra lights (if a suitable one exists)
         size_t hash = 0;
         for (unsigned int i=0; i<lightList.size();++i)
-            boost::hash_combine(hash, lightList[i]->mLightSource->getId());
+            hash = hash ^ (lightList[i]->mLightSource->getId() << 1); // or use boost::hash_combine
```
scrawl: Is this really as good as hash_combine? Have you tested for a collision?

components/fallback/fallback.cpp
```diff
@@ -22,15 +24,15 @@ namespace Fallback
         if(fallback.empty())
             return 0;
         else
-            return boost::lexical_cast<float>(fallback);
+            return std::stof(fallback);
```
scrawl: It's not the same: g++ test.cpp -o test -std=c++11
```
#include <iostream>
#include <string>

#include <boost/lexical_cast.hpp>

int main()
{
	setlocale(LC_NUMERIC, "de_DE.UTF-8");

	float a = std::stof("1.5");
	float b = std::stof("1,5");

	std::cout << " a" << a << " b " << b << std::endl;

	try
	{
        	a = boost::lexical_cast<float>("1.5");
	}
	catch(...)
	{
		a = 0;
	}
	try
	{
        	b = boost::lexical_cast<float>("1,5");
	}
	catch(...)
	{
		b = 0;
	}

        std::cout << " a" << a << " b " << b << std::endl;

	return 0;
}
```
output:
<pre>
 a1 b 1.5
 a1.5 b 0
</pre>

1. error handling: boost throws an exception, stof silently ignores parsing errors
2. locale: boost::lexical_cast does not use current locale, stof does use it. We don't want to use it. (The default is 'C' usually, but AFAIK QT changes it).

scrawl:
I don't think there's any reasonable replacement for boost::filesystem.

components/sceneutil/lightmanager.cpp
psi29a: It was faster doesn't attempt to prevent  the low-end and high-end bits from colliding.

We can just do what boost does, use the golden ratio to prevent (slow down) this collision.

```
template <class T>
inline void hash_combine(std::size_t& seed, const T& v)
{
    std::hash<T> hasher;
    seed ^= hasher(v) + 0x9e3779b9 + (seed<<6) + (seed>>2);
}
```

Do we want good enough or do we want 'the best' but slower.

scrawl: We want whatever was working before and has not caused problems.

components/widgets/numericeditbox.cpp
psi29a: What do you suggest as a replacement?

components/fallback/fallback.cpp
psi29a: I can revert this unless we can find something else that does work.

psi29a:
I haven't touched boost::filesystem for that reason. :/

components/widgets/numericeditbox.cpp
scrawl: Both of these, I guess?
<pre>
If no conversion could be performed, an invalid_argument exception is thrown.

If the value read is out of the range of representable values by an int, an out_of_range exception is thrown.
</pre>

psi29a:
I'll get a PR ready this evening or tomorrow morning at the latest.

components/fallback/fallback.cpp
psi29a: Thank you for catching that, I would not have thought to test that. Outside the US that the , is used instead of the .

PR #1316, opened 2017-06-10
Doors usage improvements

akortunov:
1. AI can use keys to open doors now.
2. A door opening sound will be played when AI opens a door.
3. Search for key starts only if this door can be unlocked by key and ends if the key found.

Note: I took search from door.cpp. Maybe we should create a separate function for this search?

### DIFF
--- a/apps/openmw/mwmechanics/aipackage.cpp
+++ b/apps/openmw/mwmechanics/aipackage.cpp
@@ -182,7 +182,7 @@ void MWMechanics::AiPackage::evadeObstacles(const MWWorld::Ptr& actor, float dur
         if (getTypeId() != TypeIdWander && !door.getCellRef().getTeleport() && door.getCellRef().getTrap().empty()
                 && door.getCellRef().getLockLevel() <= 0 && door.getClass().getDoorState(door) == 0)
         {
-            MWBase::Environment::get().getWorld()->activateDoor(door, 1);
+            MWBase::Environment::get().getWorld()->activate(door, actor);
         }
     }
     else // any other obstacle (NPC, crate, etc.)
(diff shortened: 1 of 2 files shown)


kcat:
This sounds like something that could easily break the game, with NPCs opening doors that the player isn't supposed to be able to get through.

akortunov:
Actually, AI can use keys in vanilla Morrowind.

apps/openmw/mwmechanics/aipackage.cpp
```diff
@@ -179,10 +180,33 @@ void MWMechanics::AiPackage::evadeObstacles(const MWWorld::Ptr& actor, float dur
     if (door != MWWorld::Ptr())
     {
         // note: AiWander currently does not open doors
-        if (getTypeId() != TypeIdWander && !door.getCellRef().getTeleport() && door.getCellRef().getTrap().empty()
-                && door.getCellRef().getLockLevel() <= 0 && door.getClass().getDoorState(door) == 0)
+        if (getTypeId() != TypeIdWander && !door.getCellRef().getTeleport() && door.getClass().getDoorState(door) == 0)
         {
-            MWBase::Environment::get().getWorld()->activateDoor(door, 1);
+            if ((door.getCellRef().getTrap().empty() && door.getCellRef().getLockLevel() <= 0 ))
+                MWBase::Environment::get().getWorld()->activate(door, actor);
```
scrawl: Looks like you call activate() twice for a door that needs a key but is already unlocked?

akortunov: Yes, you are right. I missed a return statement. Sorry.

PR #1318, opened 2017-06-11
Fix deboosting

psi29a:
Fixed 3 things that @scrawl mentioned here: https://github.com/OpenMW/openmw/pull/1314

### DIFF
--- a/components/fallback/fallback.cpp
+++ b/components/fallback/fallback.cpp
@@ -1,5 +1,8 @@
 #include "fallback.hpp"
 
+#include <boost/lexical_cast.hpp>
+
+
 namespace Fallback
 {
     bool stob(std::string const& s) {
@@ -23,8 +26,9 @@ namespace Fallback
         std::string fallback=getFallbackString(fall);
         if(fallback.empty())
             return 0;
-        else
-            return std::stof(fallback);
+        else {
+            return  boost::lexical_cast<float>(fallback);
+        }
     }
     int Map::getFallbackInt(const std::string& fall) const
     {
--- a/components/fallback/fallback.cpp
+++ b/components/fallback/fallback.cpp
@@ -21,15 +21,16 @@ namespace Fallback
         }
         return it->second;
     }
+
     float Map::getFallbackFloat(const std::string& fall) const
     {
         std::string fallback=getFallbackString(fall);
         if(fallback.empty())
             return 0;
-        else {
-            return  boost::lexical_cast<float>(fallback);
-        }
+        else
+            return boost::lexical_cast<float>(fallback);
     }
+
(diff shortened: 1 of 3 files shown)


scrawl:
What's this formatting 'fix' for? Aside from being unnecessary, I think 99% of our code uses Allman style brackets not K&R style.

psi29a:
:( 
Is this documented in our style guideline?

psi29a:
@scrawl Hopefully better?  fallback.cpp had some sloppy (no style) formatting that I cleaned up as well.

scrawl:
No, I don't think there should be any cleanup at all. That's just asking for trouble for no reason (conflicts) if someone else wants to edit the same file.

Allofich:
I can understand avoiding cleanup on files that are frequently worked on, but according to fallback.cpp's history the last time it was worked on outside of psi29a's current work was over a year ago, so it doesn't really seem like a problem to me...

Allofich:
Well, on the other hand, I guess you can never predict when someone would want to edit the file... and since it's being worked on now, it might be more likely than usual that someone would want to... Well, never mind me.

psi29a:
@scrawl I thought we agreed that we never 'cleanup for sake of cleaning up' unless we editing the file in question.

In this case, fallback.cpp needed some love. I don't see any PR that touches the file in question so there shouldn't be any conflicts. 

Would you rather I do this in another PR?

scrawl:
I think formatting should be limited to and around lines that you were already changing anyway; changing one line doesn't justify giving the whole file a makeover. Fewer changes = less risk of conflict.

> I don't see any PR that touches the file in question so there shouldn't be any conflicts.

I suppose in this case (file is old and self contained) the risk seems low, but still we shouldn't go around assuming such things. Someone may be working on a PR they haven't submitted yet or have something in their stash. Then we have two forks already (tes3mp and the oblivion one) who the hell knows what they might have changed; and just because there aren't changes now doesn't mean there won't be any in the future (from users still on old code for whatever reason).
Making no exceptions seems like the best strategy to me, that way we are consistent and don't waste any time thinking about whether something should or should not be an exception.

> I can understand avoiding cleanup on files that are frequently worked on, but according to fallback.cpp's history the last time it was worked on outside of psi29a's current work was over a year ago, so it doesn't really seem like a problem to me...

You could turn that around and ask, if a file is no longer worked on and forgotten then why is the 'bad' formatting even an issue?

psi29a:
I see your point about tes3mp, I made a revert commit. :)

scrawl:
Great, now just rebase it and I'll be happy. You can merge yourself then.

psi29a:
done and one, i'll wait for last round of ci

PR #1320, opened 2017-06-11
Parsing else if as elseif

uramer:
Solution for https://bugs.openmw.org/issues/3006?next_issue_id=3906
See discussion in the bug report for more details
Haven't properly tested yet due to current master commit not compiling


zinnschlag:
Sorry, but Nope. This isn't going to work. You can't just add a new token to the scanner. That will break many things. This must be implemented on parser level only.

PR #1323, opened 2017-06-13
Fixes for building in MSVC

Allofich:
Since the deboosting commits, I haven't been able to compile the source locally on Visual Studio 2015. This PR fixes the compiler issues for me and the CI builds work.

Current master is working in AppVeyor, and no one else has said anything about not compiling (except for uramer in https://github.com/OpenMW/openmw/pull/1320), so I would appreciate it if uramer or anyone building in Visual Studio can confirm they have the same issue.

The compile problems are that "min" and "max" are not members of std, and that std::isalnum expects 2 arguments but only 1 is provided.

### DIFF
--- a/apps/openmw/mwgui/containeritemmodel.cpp
+++ b/apps/openmw/mwgui/containeritemmodel.cpp
@@ -1,5 +1,7 @@
 #include "containeritemmodel.hpp"
 
+#include <algorithm>
+
 #include "../mwworld/containerstore.hpp"
 #include "../mwworld/class.hpp"
 
--- a/apps/openmw/mwscript/skyextensions.cpp
+++ b/apps/openmw/mwscript/skyextensions.cpp
@@ -1,5 +1,7 @@
 #include "skyextensions.hpp"
 
+#include <algorithm>
+
 #include <components/compiler/extensions.hpp>
 #include <components/compiler/opcodes.hpp>
 
--- a/apps/openmw/mwsound/ffmpeg_decoder.cpp
+++ b/apps/openmw/mwsound/ffmpeg_decoder.cpp
@@ -4,6 +4,7 @@
 
 #include <stdexcept>
 #include <sstream>
+#include <algorithm>
 
 #include <components/vfs/manager.hpp>
 
--- a/apps/openmw/mwsound/loudness.cpp
+++ b/apps/openmw/mwsound/loudness.cpp
@@ -2,6 +2,7 @@
 
 #include <stdint.h>
 #include <limits>
+#include <algorithm>
 
 #include "soundmanagerimp.hpp"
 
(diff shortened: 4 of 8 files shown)


psi29a:
Thank you for this. I'm curious what you mean since the two appveyor CI builds came back green. That was MSVC2013 and MSVC2015.

I'm thinking that the min/max/isalnum is being clobbered by windows.h which defines them differently.

psi29a:
https://support.microsoft.com/en-gb/help/143208/prb-using-stl-in-windows-program-can-cause-min-max-conflicts

Allofich:
> I'm curious what you mean since the two appveyor CI builds came back green. That was MSVC2013 and MSVC2015.

I know, so I hesitated to open this PR, but I couldn't seem to get compilation to work. I tried building 32-bit, which I think the AppVeyor builds are, rather than 64-bit like I usually do, but it still failed. The AppVeyor builds are Debug builds, so I'll try that next. Otherwise I don't know what else I can do to make my build environment more like AppVeyor's.

I was searching about this issue and found that link as well. It seems to be for an older version of Visual Studio. But NOMINMAX is already being applied in CMakeLists.txt on line 182, right? I tried adding #define NOMINMAX before #include <windows.h> in main.cpp, but that didn't seem to help anything.

psi29a:
I'll call in the expert to get his opinion, otherwise I'll merge them in. :)

@ace13 Are you having issues with current master?

Allofich:
Just FYI, I'm building a Debug build of master right now and it seems to be working. I also specified -u (Unity build) for CI/before_script.msvc.sh, just to be closer to what AppVeyor was doing. I'll try removing that next.

Allofich:
The debug build compiled successfully. I removed the -u switch and it still seems to be compiling correctly, so I think it was a case of Release builds not compiling but Debug builds, which AppVeyor does, were compiling. Should AppVeyor be doing Debug builds? It seems like it should do whatever type is the most likely to fail compiling, but I assume there's a reason it's doing Debug builds.

Allofich:
Oh, also ace13 posted on the forums that he is away from home for a while, so he may not be able to comment. Also his automatic builds are apparently failing anyway because of a lack of disk space, so he probably wouldn't know if the recent commits caused problems.

https://forum.openmw.org/viewtopic.php?f=20&p=47741&sid=e96930b3fd3486759924066f34baa32d#p47741

Allofich:
In the meantime, I set AppVeyor running the current master in all configurations.
https://ci.appveyor.com/project/Allofich/openmw/build/568

It will probably be a few hours before it's all done.

psi29a:
Merging then, because we'll need release builds eventually. :)

Thank you for following this up! I had hoped to remove any stale includes that might slow down compile time.

Allofich:
Great, well, I expect the AppVeyor test builds I'm running will just confirm the Debug works, Release doesn't hypothesis, but I'll write something here if they show anything unexpected.

Allofich:
Well the AppVeyor builds all worked, so I guess there is some difference I didn't discover between the AppVeyor build environment and my own.

PR #1324, opened 2017-06-14
Improved doors detection

akortunov:
This PR provide a way to detect doors using rendering ray casting (as with spell projectiles) istead of iterating over all doors in an actor cell.

Allows to fix issues such as "Combat AI sometimes cannot open the door".
See my posts in [this](https://bugs.openmw.org/issues/3227) thread.

Note: maybe should we also do a rendering raycasting as fallback?

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -10,6 +10,7 @@
 #include "../mwworld/ptr.hpp"
 
 #include "../mwrender/rendermode.hpp"
+#include "../mwphysics/physicssystem.hpp"
 
 namespace osg
 {
@@ -298,6 +299,9 @@ namespace MWBase
 
             virtual bool castRay (float x1, float y1, float z1, float x2, float y2, float z2) = 0;
             ///< cast a Ray and return true if there is an object in the ray path.
+
+            virtual MWPhysics::PhysicsSystem::RayResult castRayTest (float x1, float y1, float z1, float x2, float y2, float z2) = 0;
+            ///< cast a rendering ray and return ray result.
 
             virtual bool toggleCollisionMode() = 0;
             ///< Toggle collision mode for player. If disabled player object should ignore
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -10,7 +10,6 @@
 #include "../mwworld/ptr.hpp"
 
 #include "../mwrender/rendermode.hpp"
-#include "../mwphysics/physicssystem.hpp"
 
 namespace osg
 {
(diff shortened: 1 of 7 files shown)


psi29a:
What is the performance cost of this solution?

akortunov:
> What is the performance cost of this solution?

What kind of speed test should I do?

akortunov:
Do not merge it, usage of rendering ray can lead to problems with non-solid doors.
Now I am testing usage of physics castRay instead (based by bounding boxes). No problems so far.

akortunov:
Using physics ray now. But I am not sure

akortunov:
Using activation distance instead of hardcoded value now.
Also door AI is enabled in exteriors.
Tested on the gates outside the Census and Excise Office in Seyda Neen.

akortunov:
Rebased 4 commits into 1.

scrawl:
I think a raycast is overkill, and they are rather expensive. We know that actors are always facing horizontally, so it should be relatively simple to determine if they're facing a door without using a raycast. Maybe the blocking angle code in combat.cpp helps for reference.

akortunov:
> Maybe the blocking angle code in combat.cpp helps for reference.

I do not see any help from blocking in this case:

With blocking we already know both attacker's and target's ID's and position, so can calculate angle and distantion between attacker and target easily.

With door detection we know only actor ID and position. Now we have to iterate over all doors in the cell to determine a distance between the door and actor. This algorithm is linear, so we had to disable it in exteriors. Also we have a similar check in Actors::getObjectsInRange().

How can I detect an object which NPC is facing without raycasting and iterating over all objects in the cell?

Or a current OpenMW implementation is a fast enough to enable it in exteriors, and add the angle check here?

scrawl:
Iterating all doors in the cell should be much cheaper than a raycast. I have no idea why the code was disabled for exteriors.

akortunov:
> Iterating all doors in the cell should be much cheaper than a raycast. 

Ok, I will try to improve a current OpenMW implementation.

akortunov:
I uploaded a my first attempt to improve the current doors detection algorithm.

Notes:
1. I took an angle formula from Actors::updateHeadTracking().
2. I do not know how to handle Z offset properly. In current master AI sometimes fails to open a door because of difference between door Z position and actor Z position.
3. The doors detection algorithm is enabled in exteriors now.
4. On a short distances the angle between an actor position and a door position can be up to PI/2, even if the actor is facing the door directly.

Tested with these doors:
1. Gates outside the Census and Excise Office in Seyda Neen.
2. A wooden door in "Ald-ruhn, Arobar Guard Quarters".
3. A wooden door near entrance in Addamasartus.

In master implementation AI fails to open first two of these doors, with this PR AI can open all these doors.

scrawl:
> On a short distances the angle between an actor position and a door position can be up to PI/2, even if the actor is facing the door directly.

Right, because the door position is only a point. To be more accurate maybe we could use the door's bounding box, or construct a plane through the door position based on its orientation?

PR #1326, opened 2017-06-16
Combat AI improvements [feedback needed]

akortunov:
1. AI will not use range weapons under water
2. AI will take in account enemy normal weapon resistance and silver weapons damage bonus (against werewolves).
3. AI will not use bows/crossbows if best arrow/bolt rating is 0.
4. AI will choose target by best action priority, not only by distance (see [this](https://bugs.openmw.org/issues/3921) feature).
For example, warrior with steel longsword will not attack summoned Golden Saint (100% normal weapons resistance), he will attack another enemies (player-summoner, for example) instead.

Notes: 
1. Strange, but both vanilla and OpenMW engines allow to kill actors with immunity to normal weapons (e.g. ghosts) with bare hands. Should we alter vanilla behaviour or leave the current implementation?
2. If actor can do nothing with all enemies, he will attack nearest enemy with bare hands.

Also I have a question: when I open a PR to fix a not-registered bug, should I create a bugreport too?

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -535,6 +535,10 @@ namespace MWMechanics
         // opponent's weapon range, or not backing up if opponent is also using a ranged weapon
         if (isDistantCombat && distToTarget < rangeAttack / 4)
         {
+            // actor should not back up into water
+            if (MWBase::Environment::get().getWorld()->isUnderwater(MWWorld::ConstPtr(actor), 0.5f))
+                return;
+
             mMovement.mPosition[1] = -1;
         }
     }
--- a/apps/openmw/mwmechanics/aicombataction.cpp
+++ b/apps/openmw/mwmechanics/aicombataction.cpp
@@ -854,6 +854,79 @@ namespace MWMechanics
         return bestAction;
     }
 
+    float getBestActionRating(const MWWorld::Ptr &actor, const MWWorld::Ptr &enemy)
+    {
+        Spells& spells = actor.getClass().getCreatureStats(actor).getSpells();
+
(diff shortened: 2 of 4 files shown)


akortunov:
> For example, NPC without suitable weapons/spells attacks summoned Golden Saint with bare hands instead of attacking a summoner.

Since AI attacks nearest target, I can make AI flee if it can do nothing to current target (bestActionRating == 0). When bestActionRating will be > 0 (AI switched to another target, e.g. summoner get close, or conjuration spell expired), AI will attack.

Capostrophic:
> Strange, but both vanilla and OpenMW engines allow to kill such actors (e.g. ghosts) with bare hands. Should we alter vanilla behaviour?

No, at least not right now. Leave this for post-1.0 mods.

> when I open a PR to fix a not-registered bug, should I create a bugreport too?

Preferably, but not necessarily.

Allofich:
> Strange, but both vanilla and OpenMW engines allow to kill such actors (e.g. ghosts) with bare hands. Should we alter vanilla behaviour?

You can also do this in other Elder Scrolls games (Daggerfall, at least). From a gameplay perspective it's probably to allow hand-to-hand characters like monks to be able to get through the game.

> when I open a PR to fix a not-registered bug, should I create a bugreport too?

I think it's preferred, because the bug tracker issues are used for writing up a list of changes when a new version is released. There may be other reasons, too.

akortunov:
Also I can make non-vanilla improvements optional if we need it.

Allofich:
> Also I can make non-vanilla improvements optional if we need it.

To me that's the ideal solution. I am "R.D." on the bugtracker, and I gave this opinion in the comments of https://bugs.openmw.org/issues/3406.

PR #1327, opened 2017-06-19
build.msvc.sh msbuild configuration fix

tlmullis:
msbuild was always defaulting to building the Debug configuration so it was impossible to build any other configurations with build.msvc.sh


scrawl:
@ace13 is this PR ok?

ace13:
@scrawl LGTM

Ended up not using this script myself, turned out to be easier than it looked at first to get AppVeyor to discover the generated solution.

PR #1329, opened 2017-06-28
Combat AI: make default spell priority calculation formula close to vanilla

akortunov:
Fixes [#3937](https://bugs.openmw.org/issues/3937), partially fixes [#3406](https://bugs.openmw.org/issues/3406) (Ash Slave spell priority issue) and [#3916](https://bugs.openmw.org/issues/3916) (still need to adjust ranged weapons priority).

What this PR does:
1) Takes in account spell area
2) Target spells have +50% rating
3) Normalizes spells rating (divide to 40)

Note: original Morrowind just takes spell magicka cost, but I took a magicka cost calculation formula to handle non-autocalculated spells correctly.


apps/openmw/mwmechanics/aicombataction.cpp
```diff
@@ -594,13 +592,17 @@ namespace MWMechanics
...
             }
         }
-        else
-        {
-            rating *= (effect.mMagnMin + effect.mMagnMax)/2.f;
-        }
 
-        if (!(magicEffect->mData.mFlags & ESM::MagicEffect::NoDuration))
-            rating *= effect.mDuration;
+        rating *= ((effect.mMagnMin + effect.mMagnMax) * (effect.mDuration > 0 ? effect.mDuration : 1) + effect.mArea);
```
scrawl: Why should we use magnMin / magnMax for a spell with NoMagnitude? Same question for duration, too. I wouldn't expect these variables to have a sane default value if the respective 'don't use' flag is set.

akortunov: NoMagnitude spells have magnMin and magnMax = 1. I just took vanilla magicka cost calculation formula from autocalcspell.cpp and spellcasting.cpp. If we do not have a sane default value, we need to fix these files too.

akortunov:
Magicka cost calculations formula was moved to separate function to aviod a code duplication.

apps/openmw/mwmechanics/spellcasting.hpp
```diff
@@ -4,6 +4,7 @@
 #include <components/esm/loadskil.hpp>
 
 #include "../mwworld/ptr.hpp"
+#include "../mwworld/esmstore.hpp"
```
scrawl: Seems like we are including much more than necessary here. components/esm/effectlist.hpp would be enough, right?

apps/openmw/mwmechanics/spellcasting.hpp
```diff
@@ -21,6 +22,8 @@ namespace MWMechanics
 
     ESM::Skill::SkillEnum spellSchoolToSkill(int school);
 
+    float calcEffectCost(ESM::ENAMstruct effect);
```
scrawl: non-primitive types should always be passed by const reference. If you don't, cppcheck will complain (and the resulting code might be slower, though maybe some compilers will optimize it?)

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -44,26 +44,44 @@ namespace MWMechanics
...
+        if (!(magicEffect->mData.mFlags & ESM::MagicEffect::NoDuration))
+            duration = effect.mDuration;
+
+        if (!(magicEffect->mData.mFlags & ESM::MagicEffect::UncappedDamage))
+            duration = std::max(1, duration);
+
+        static const float fEffectCostMult = MWBase::Environment::get().getWorld()->getStore()
+            .get<ESM::GameSetting>().find("fEffectCostMult")->getFloat();
+
+        float cost = (minMaxMagn * (duration + 1) + std::max(1, effect.mArea)) * magicEffect->mData.mBaseCost * fEffectCostMult * 0.05;
```
scrawl: Best to avoid refactoring game mechanics code, renaming its variables, rearranging statements etc. It will become harder to verify that the code matches the [research page](https://wiki.openmw.org/index.php?title=Research:Trading_and_Services#Spell_merchant). 

Code duplication should always be avoided though so its OK that you made a new function.

akortunov:
Committed changes mentioned in review.

According to our [research](https://wiki.openmw.org/index.php?title=Research:Magic), magicka cost formula uses "duration+1" multiplier, but success chance calculation formula uses just "duration" multiplier. 
Does original Morrowind have this difference too?

scrawl:
> According to our research, magicka cost formula uses "duration+1" multiplier, but success chance calculation formula uses just "duration" multiplier.
> Does original Morrowind have this difference too?

I have no clue.

akortunov:
There is indeed a difference in formula. I had to use different functions for magicka cost and chance formulas.

Anyway some additional testing needed.

scrawl:
What in particular needs additional testing? Did you add some print statements for each formula to make sure results are the same before and after PR?

akortunov:
> What in particular needs additional testing?

This PR affects basically all spells and magic items. It fixes an Ash Slave behaviour and enchanted weapons priority, but can break some other spells priority.
I personally did not encountered such issues during testing.

> Did you add some print statements for each formula to make sure results are the same before and after PR?

Yes, I did this, results were the same **in my testing**.

scrawl:
> It fixes an Ash Slave behaviour and enchanted weapons priority, but can break some other spells priority.

I guess the only way we will really find out is having the users test for us.

PR #1330, opened 2017-07-07
minimal height adjusted to the size of the contained image

RolandDH:
closes bug #3910

### DIFF
--- a/files/ui/mainwindow.ui
+++ b/files/ui/mainwindow.ui
@@ -58,7 +58,14 @@
       </property>
       <layout class="QVBoxLayout" name="verticalLayout">
        <item>
-        <widget class="QStackedWidget" name="pagesWidget"/>
+        <widget class="QStackedWidget" name="pagesWidget">
+         <property name="minimumSize">
+          <size>
+           <width>0</width>
+           <height>375</height>
+          </size>
+         </property>
+        </widget>
        </item>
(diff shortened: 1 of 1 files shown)


scrawl:
Which platform did you test on? The image is still slightly cut off for me on Linux. Also, it would be nice to implement this feature in a non-redundant way if possible.

RolandDH:
On Samstag, 8. Juli 2017 22:26:31 CEST scrawl wrote:
> Which platform did you test on?

Linux, me too. Did look OK here.

> The image is still slightly cut off for me

That's odd. 375 is exactly the height in pixels of
:/images/playpage-background.png. To my eyes that PNG has even some 
transpartent margin within that height.

> Also, it would be nice to implement this feature in a
> non-redundant way if possible.

I assume that would require a redesign of the ui files. I agree that that would 
be nice, but seemed to much effort just for my very first PR. ;-) Perhaps I'll 
pick another launcher bug and then try that too.

-- 
Roland Hautz

akortunov:
> The image is still slightly cut off for me on Linux. 

Maybe it depends on Qt version or other third-party libraries. For some reason I can not reproduce this bug on Linux with Qt 5.7.1.

Thunderforge:
PR #1349 also attempts to fix this bug.

PR #1331, opened 2017-07-10
Prevent GUI sounds distortion

akortunov:
This PR adds a playSound() function to WindowManager, similar to playVideo() one.
This function uses Play_NoEnv mode and played sounds will not be distorted underwater.
Allow to partially fix [#3942](https://bugs.openmw.org/issues/3942).

As alternative solution, we can change default mode for playSound() in SoundManager from Play_Normal to Play_NoEnv.

Or we can just use both solutions.

Note: we also have at least two cases to fix:
1) Sound played by console commands without position (such as "PlaySound", "PlayLoopSound") should not be distorted.
2) Action sound (equip item, drink potion, etc.) should not be distorted too.

### DIFF
--- a/apps/openmw/mwgui/tradewindow.cpp
+++ b/apps/openmw/mwgui/tradewindow.cpp
@@ -10,7 +10,6 @@
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
-#include "../mwbase/soundmanager.hpp"
 #include "../mwbase/windowmanager.hpp"
 #include "../mwbase/mechanicsmanager.hpp"
 #include "../mwbase/dialoguemanager.hpp"
@@ -211,7 +210,7 @@ namespace MWGui
     {
         const ItemStack& item = mTradeModel->getItem(mItemToSell);
         std::string sound = item.mBase.getClass().getDownSoundId(item.mBase);
-        MWBase::Environment::get().getSoundManager()->playSound (sound, 1.0, 1.0);
+        MWBase::Environment::get().getWindowManager()->playSound(sound);
 
         TradeItemModel* playerTradeModel = MWBase::Environment::get().getWindowManager()->getInventoryWindow()->getTradeModel();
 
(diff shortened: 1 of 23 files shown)


apps/openmw/mwgui/tradewindow.cpp
```diff
@@ -356,9 +355,7 @@ namespace MWGui
         MWBase::Environment::get().getWindowManager()->getDialogueWindow()->addResponse(
             MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("sBarterDialog5")->getString());
 
-        std::string sound = "Item Gold Up";
-        MWBase::Environment::get().getSoundManager()->playSound (sound, 1.0, 1.0);
-
+        MWBase::Environment::get().getWindowManager()->playSound("Item Gold Up");
```
Thunderforge: This is for grabbing gold, right? Would it make sense for grabbing gold underwater to make a different sound, since the coins would clink together in a distorted way?

apps/openmw/mwgui/travelwindow.cpp
```diff
@@ -158,7 +157,7 @@ namespace MWGui
 
         if (!mPtr.getCell()->isExterior())
             // Interior cell -> mages guild transport
-            MWBase::Environment::get().getSoundManager()->playSound("mysticism cast", 1, 1);
+            MWBase::Environment::get().getWindowManager()->playSound("mysticism cast");
```
Thunderforge: Why is casting Mysticism spells immune from underwater sound effects, but not other types of spells?

apps/openmw/mwgui/tradewindow.cpp
akortunov: Inventory sounds are not distorted in vanilla game.

apps/openmw/mwgui/travelwindow.cpp
akortunov: It is not mysticism, it is for travelling GUI window.

PR #1332, opened 2017-07-11
Decouple projectile speed calculations from aiming code, direction finding from magic bolt code

h313:
This allows issues [#3948](https://bugs.openmw.org/issues/3948) and [#3488](https://bugs.openmw.org/issues/3488) to be worked on


scrawl:
I think the merging of the PR should wait until one of these features is actually included.

Anyway, I don't understand what you changed launchMagicBolt() for. AFAIK, the fallbackDirection isn't used by combat AI; it's for effects casted from a script.

h313:
So where is the code that AI uses to cast target spells?

PR #1335, opened 2017-07-16
GUI fixes

akortunov:
1) Fixes quick keys menu width calculation and make it non-transparent ([screenshot](http://i.imgur.com/P4PPVSr.jpg))
2) Removes "Cost/Change" and "Cost/Charge" labels from quick key magic selection dialog and make it non-tranparent ([screenshot](http://i.imgur.com/TMRD4bi.jpg))
3) Makes character selection popup in game load menu non-transparent ([screenshot](http://i.imgur.com/z1smMFQ.jpg)).

### DIFF
--- a/files/mygui/openmw_resources.xml
+++ b/files/mygui/openmw_resources.xml
@@ -114,6 +114,24 @@
         </Widget>
     </Resource>
 
+    <Resource type="ResourceLayout" name="MW_ComboBoxNoTransp" version="3.2.0">
+        <Widget type="Widget" skin="" position="65 10 100 26" name="Root">
+            <Property key="MaxListLength" value="200"/>
+            <Property key="SmoothShow" value="true"/>
+            <Property key="ModeDrop" value="true"/>
+
+            <Widget type="Widget" skin="MW_Box" position="0 0 100 26" align="Stretch">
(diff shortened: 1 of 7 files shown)


uramer:
aren't those menus supposed to be transparent? At least I liked them that way. Good job on removing those silly bugs though!

akortunov:
> aren't those menus supposed to be transparent?

No. Almost all widgets placed over other widgets are not transparent in vanilla game.

files/mygui/openmw_resources.xml
```diff
@@ -114,6 +114,24 @@
         </Widget>
     </Resource>
 
+    <Resource type="ResourceLayout" name="MW_ComboBoxNoTransp" version="3.2.0">
```
scrawl: As far as I recall, ComboBox is only used once so you can just change the original skin instead.

akortunov: It is worth to create a separate PR for that?

scrawl: Sure.

akortunov: Just a note: combobox is also used in settings menu (e.g. in water tab).

PR #1339, opened 2017-07-22
Add Visual Studio 2017 support (Feature #3952)

PlutonicOverkill:
https://bugs.openmw.org/issues/3952
Adds Visual Studio 2017 support to the build script. This uses the same configuration settings as for VS2015 but uses a different generator. I removed the old TOOLSET and XP_TOOLSET variables as they didn't seem to be used anywhere, but they can be added back if they do turn out to be needed.
Should build.msvc.sh be upgraded as well? I'm not sure what its purpose is exactly but it's probably something most users won't need to run.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -215,9 +215,7 @@ fi
 case $VS_VERSION in
 	15|15.0|2017 )
 		GENERATOR="Visual Studio 15 2017"
-		XP_TOOLSET="v140xp"
-		TOOLSET="v140"
-		BOOST_TOOLSET="vc140"
+		TOOLSET="vc140"
 		MSVC_VER="14"
 		MSVC_YEAR="2015"
 		MSVC_DISPLAY_YEAR="2017"
@@ -225,9 +223,7 @@ case $VS_VERSION in
 
 	14|14.0|2015 )
 		GENERATOR="Visual Studio 14 2015"
-		XP_TOOLSET="v140_xp"
-		TOOLSET="v140"
-		BOOST_TOOLSET="vc140"
+		TOOLSET="vc140"
 		MSVC_VER="14"
 		MSVC_YEAR="2015"
 		MSVC_DISPLAY_YEAR="2015"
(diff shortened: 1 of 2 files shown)


scrawl:
@ace13 is this PR ok?

ace13:
That's a good question actually, Visual Studio 2013 and 2015 at least both call their toolsets `v<version>`, so not sure if they'll agree with `vc<version>`.

I'll do some tests and report back.

ace13:
Yeah, this won't work for VS2013 or 2015.
devenv_2017-07-22_13-05-35

ace13:
Or actually, come to think of it, I think I redid those parts, so nevermind my comments.

LennyPenny:
When I manually updated the vs2015 projects to 2017 the new toolset (v141) was chosen and everything compiled just fine. 



Our guess is that v140 and v141 are binary compatible

CI/before_script.msvc.sh
```diff
@@ -213,20 +213,28 @@ if [ -z $VS_VERSION ]; then
 fi
 
 case $VS_VERSION in
+	15|15.0|2017 )
+		GENERATOR="Visual Studio 15 2017"
+		TOOLSET="vc140"
```
LennyPenny: Try this with 141, it should work

ace13:
That's nice of them, don't think I've ever heard of Visual Studio being forward-compatible in that manner before.

But yeah, I think we'd still want to build some proper v141 packages down the line.

Until then, I guess it LGTM

CI/before_script.msvc.sh
PlutonicOverkill: No, this variable is intended to be vc140. The only reason it is used is to make sure CMake looks for the Boost v140 binaries instead of trying to look for non-existent v141 binaries. I guess this variable should be renamed to something less misleading.

PlutonicOverkill:
I don't know for sure if it was this pull that caused it, but recently I haven't been able to run built executables, always getting a runtime_error instead. It might be a good idea to hold off merging this at least until I find out exactly what's caused the problem.

PlutonicOverkill:
Arrgh, nevermind, it's just the old $(ProjectDir)/$(OutDir) thing that I forgot to set again. So looks like this feature works OK after all.
(Might be a good idea to add something about this [here](https://gist.github.com/ace13/de7f30454e99b3952405)/on the wiki somewhere, so that other people don't waste hours trying to figure out why it won't run.)

CI/before_script.msvc.sh
LennyPenny: But in the end the 141 toolset will be still be used?

PlutonicOverkill: Yes, that is correct.

PR #1342, opened 2017-07-23
Fade screen in on a new game start

akortunov:
Implements a missing case from [Feature #824](https://bugs.openmw.org/issues/824).

And what kind of work should we do to close this feature request?

Also I found a difference in fading behaviour: in vanilla Morrowind screen fades BEFORE location change (loading bar), in OpenMW - AFTER location change.

@scrawl, have we a description about how screen fading should behave?


scrawl:
IMO the rest of that feature isn't really needed, because there are no loading screens usually we would have to add some artificial delay for the screen to properly fade and that is just wasting everyone's time.

akortunov:
> IMO the rest of that feature isn't really needed

Can we cut fading from cell transitions then? Our current implementation is broken anyway.

PR #1344, opened 2017-07-24
Fix dialogue window performance

julianko:
Fixes [Bug #3126](https://bugs.openmw.org/issues/3126).

There were 2 problems with the dialogue window:

1. MyGUI doesn't seem to have separate events for resizing and moving, so when dragging the window it went through the whole resize routine which recreates the dialogue history etc.. Now it only does that when the size has actually changed.
2. During (re)writing the dialogue history, the text colours were retrieved by letting MyGUI parse a generated string over and over again hundreds of times per frame which caused the majority of the slowdown. The text colours are now cached in MWGui::WindowManager (they can't change at runtime right?) and can be used directly.

### DIFF
--- a/apps/openmw/mwgui/dialogue.cpp
+++ b/apps/openmw/mwgui/dialogue.cpp
@@ -22,18 +22,9 @@
 
 #include "widgets.hpp"
 #include "bookpage.hpp"
+#include "textcolours.hpp"
 
 #include "journalbooks.hpp" // to_utf8_span
-
-namespace
-{
-
-    MyGUI::Colour getDialogueTextColour (const std::string& type)
-    {
-        return MyGUI::Colour::parse(MyGUI::LanguageManager::getInstance().replaceTags("#{fontcolour=" + type + "}"));
-    }
-
-}
 
 namespace MWGui
 {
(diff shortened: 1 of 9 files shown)


scrawl:
Nice find. I only have two nitpicks:
* The build error should be rebased. 
* Please don't use ```auto``` in this way (or at all), using it opens a can of worms when it comes to abuse and I don't want to waste any time arguing on whether or not a particular use is acceptable.

julianko:
alright, i've rebased the whole thing into a single commit and removed the auto usages.
just for educational purposes: what is the "abuse" with auto?

scrawl:
It can harm the readability of the code, because I have to use an IDE to know what the actual type is and tools like 'grep' or 'search for all uses' will no longer work properly. Code is more often read than written, so there's no excuse for being lazy.

Also, it is breaking consistency with the existing codebase. Try to use similar style/conventions to the code that you're changing. If you are writing an entirely new component though, then the style is up to you.

apps/openmw/mwgui/dialogue.cpp
```diff
@@ -300,8 +293,12 @@ namespace MWGui
 
     void DialogueWindow::onWindowResize(MyGUI::Window* _sender)
     {
+        // if the window has only been moved, not resized, we don't need to update
+        if (mCurrentWindowSize == _sender->getSize()) return;
```
julianko: MyGUI::IntSize has a default constructor that initializes with (0, 0).

PR #1348, opened 2017-07-25
Editor: Fix handling of cell color field

rcutmore:
This fixes the cell colour field issue mentioned in [bug 3148](https://bugs.openmw.org/issues/3148).

The conversion from QColor to QVariant to int and vice versa was broken. This PR switches back to using a colour field and explicitly performs the conversion between types.

### DIFF
--- a/apps/opencs/view/world/util.cpp
+++ b/apps/opencs/view/world/util.cpp
@@ -181,9 +181,7 @@ QWidget *CSVWorld::CommandDelegate::createEditor (QWidget *parent, const QStyleO
     // (the third parameter of ColorEditor's constructor)
     else if (display == CSMWorld::ColumnBase::Display_Colour)
     {
-        int colorInt = index.data().toInt();
-        QColor color = QColor(colorInt & 0xff, (colorInt >> 8) & 0xff, (colorInt >> 16) & 0xff);
-        return new CSVWidget::ColorEditor(color, parent, true);
+        return new CSVWidget::ColorEditor(index.data().toInt(), parent, true);
     }
     return createEditor (parent, option, index, display);
 }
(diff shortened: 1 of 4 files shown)


apps/opencs/view/world/util.cpp
```diff
@@ -179,7 +181,9 @@ QWidget *CSVWorld::CommandDelegate::createEditor (QWidget *parent, const QStyleO
     // (the third parameter of ColorEditor's constructor)
     else if (display == CSMWorld::ColumnBase::Display_Colour)
     {
-        return new CSVWidget::ColorEditor(index.data().value<QColor>(), parent, true);
+        int colorInt = index.data().toInt();
+        QColor color = QColor(colorInt & 0xff, (colorInt >> 8) & 0xff, (colorInt >> 16) & 0xff);
```
zinnschlag: This line of code is repeated rather often. Wouldn't it be better to move it to the ColorEditor class? An overlaoded constructor, maybe?

rcutmore: Good point, I'll be removing some of the code duplication.

rcutmore:
@zinnschlag I moved both conversions, QColor to int and int to QColor, into the `ColorEditor` class.

zinnschlag:
Will merge. But we have a bit of a record-keeping issue here. Are you planning to also address the other two problems in Bug 3148? If no, please file a separate issue on the tracker.

rcutmore:
I plan to but I submitted this now because I might not be able to right away. I'll make a separate issue on the tracker for this and update 3148 just to be safe.

rcutmore:
New ticket is here: [Bug 3971](https://bugs.openmw.org/issues/3971)

PR #1349, opened 2017-07-26
UI adjustments to prevent parchment scroll from being cut off (fixes #3910)

Thunderforge:
I figure that the margin is unnecessary since the parchment scroll PNG has a built in margin. Increased the window size a bit to properly fit the parchment.

This is what is the minimum window size with these changes:

Launcher window with changes

Fixes [#3910](https://bugs.openmw.org/issues/3910)

### DIFF
--- a/files/ui/mainwindow.ui
+++ b/files/ui/mainwindow.ui
@@ -7,13 +7,13 @@
     <x>0</x>
     <y>0</y>
     <width>635</width>
-    <height>575</height>
+    <height>565</height>
    </rect>
   </property>
   <property name="minimumSize">
    <size>
     <width>635</width>
-    <height>535</height>
+    <height>565</height>
    </size>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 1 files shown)


Thunderforge:
Looks like the check failed because “AppVeyor was unable to build non-mergeable pull request”. But GitHub says that there are no conflicts with the base branch. Did I do something wrong?

zinnschlag:
I have no idea what is going on here. I do not see anything suspicious. It could be a glitch. Maybe close this PR and send a new one? That's a power-cycle-level dumb solution, but worth a shot anyway.

@scrawl Any idea?

Thunderforge:
PR #1330 also attempts to fix this bug.

scrawl:
This one fixes it for me, while the other PR did not. However, there is a new warning:
<pre>
files/ui/settingspage.ui: Warning: The name 'horizontalSpacer' (QSpacerItem) is already in use, defaulting to 'horizontalSpacer1'.
</pre>

scrawl:
Actually, that warning started with #1345 it seems.

Thunderforge:
@scral I've fixed the duplicate name reference with PR ##1358.

PR #1350, opened 2017-07-26
Do not allow to loot fighting actors during death animation (bug #3528)

akortunov:
Allows to fix [bug #3528](https://bugs.openmw.org/issues/3528).

Also adds an option to control lootiong behaviour:
1) True (default) - player can loot only non-fighting actors during death animation (as in vanilla Morrowind).
2) False - player can not loot all actors during death animation.

Tested with normal NPCs, summoned creatures and spriggans.

### DIFF
--- a/apps/openmw/mwclass/npc.cpp
+++ b/apps/openmw/mwclass/npc.cpp
@@ -885,6 +885,14 @@ namespace MWClass
         if(stats.getAiSequence().isInCombat())
             return std::shared_ptr<MWWorld::Action>(new MWWorld::FailedAction(""));
 
+        if(getCreatureStats(actor).getStance(MWMechanics::CreatureStats::Stance_Sneak)
+            || ptr.getClass().getCreatureStats(ptr).getKnockedDown())
+            return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionOpen(ptr)); // stealing
+
+        // Can't talk to werewolfs
+        if(ptr.getClass().isNpc() && ptr.getClass().getNpcStats(ptr).isWerewolf())
+            return std::shared_ptr<MWWorld::Action> (new MWWorld::FailedAction(""));
+
         return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionTalk(ptr));
     }
 
--- a/apps/openmw/mwclass/npc.cpp
+++ b/apps/openmw/mwclass/npc.cpp
@@ -7,6 +7,7 @@
 #include <components/esm/loadmgef.hpp>
 #include <components/esm/loadnpc.hpp>
 #include <components/esm/npcstate.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
@@ -863,16 +864,27 @@ namespace MWClass
             return action;
         }
 
-        if(getCreatureStats(ptr).isDead())
-            return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionOpen(ptr, true));
-        if(ptr.getClass().getCreatureStats(ptr).getAiSequence().isInCombat())
-            return std::shared_ptr<MWWorld::Action>(new MWWorld::FailedAction("#{sActorInCombat}"));
-        if(getCreatureStats(actor).getStance(MWMechanics::CreatureStats::Stance_Sneak)
-                || ptr.getClass().getCreatureStats(ptr).getKnockedDown())
-            return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionOpen(ptr)); // stealing
-        // Can't talk to werewolfs
-        if(ptr.getClass().isNpc() && ptr.getClass().getNpcStats(ptr).isWerewolf())
+        const MWMechanics::CreatureStats& stats = getCreatureStats(ptr);
+
+        if(stats.isDead())
+        {
+            bool canLoot = Settings::Manager::getBool ("can loot during death animation", "Game");
+
+            // by default user can loot friendly actors during death animation
+            if (canLoot && !stats.getAiSequence().isInCombat())
+                return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionOpen(ptr, true));
+
+            // otherwise wait until death animation
+            if(stats.isDeathAnimationFinished())
+                return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionOpen(ptr, true));
+
+            // death animation is not finished, do nothing
             return std::shared_ptr<MWWorld::Action> (new MWWorld::FailedAction(""));
+        }
+
+        if(stats.getAiSequence().isInCombat())
+            return std::shared_ptr<MWWorld::Action>(new MWWorld::FailedAction(""));
+
         return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionTalk(ptr));
     }
 
@@ -1019,7 +1031,11 @@ namespace MWClass
             return true;
 
         const NpcCustomData& customData = ptr.getRefData().getCustomData()->asNpcCustomData();
-        return !customData.mNpcStats.getAiSequence().isInCombat() || customData.mNpcStats.isDead();
+
+        if (customData.mNpcStats.isDead() && customData.mNpcStats.isDeathAnimationFinished())
+            return true;
+
+        return !customData.mNpcStats.getAiSequence().isInCombat();
     }
 
     MWGui::ToolTipInfo Npc::getToolTipInfo (const MWWorld::ConstPtr& ptr, int count) const
--- a/apps/openmw/mwmechanics/summoning.hpp
+++ b/apps/openmw/mwmechanics/summoning.hpp
@@ -22,7 +22,7 @@ namespace MWMechanics
                             float magnitude, float remainingTime = -1, float totalTime = -1);
 
         /// To call after all effect sources have been visited
-        void process();
+        void process(bool cleanup);
 
     private:
         MWWorld::Ptr mActor;
--- a/docs/source/reference/modding/settings/game.rst
+++ b/docs/source/reference/modding/settings/game.rst
@@ -65,6 +65,22 @@ the type of attack is determined by the direction that the character is moving a
 The default value is false.
 This setting can be toggled with the Always Use Best Attack button in the Prefs panel of the Options menu.
 
+can loot during death animation
+-------------------------------
+
+:Type:		boolean
+:Range:		True/False
+:Default:	True
+
+If this setting is true, the player is allowed to loot actors (e.g. summoned creatures) during death animation, if they are not in combat.
+However disposing corpses during death animation is not recommended - death counter may not be incremented, and this behaviour can break quests.
+This is how original Morrowind behaves.
+
+If this setting is false, player has to wait until end of death animation in all cases.
+This case is more safe, but makes using of summoned creatures exploit (looting summoned Dremoras and Golden Saints for expensive weapons) a lot harder.
+
+The default value is true. This setting can only be configured by editing the settings configuration file.
+
 difficulty
 ----------
 
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -4,6 +4,7 @@
 
 #include <components/esm/loadcrea.hpp>
 #include <components/esm/creaturestate.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwmechanics/creaturestats.hpp"
 #include "../mwmechanics/magiceffects.hpp"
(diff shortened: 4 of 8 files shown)


scrawl:
Wasn't there some exploit in the original game that allowed you to loot summoned creatures if you pressed the button at the correct time? How does this PR affect that?

akortunov:
> Wasn't there some exploit in the original game that allowed you to loot summoned creatures if you pressed the button at the correct time? How does this PR affect that?

Yes, there was an exploit in vanilla game.
By default, you can use this exploit in OpenMW now.
With

`can loot during death animation = false`

this exploit is blocked.

scrawl:
I mean that you could still loot them even if they <i>were</i> fighting you, if you timed it right. Or am I misremembering?

akortunov:
I tested a russian version of MTB.
How it behaves:
1. If actor is in combat _with anyone_, you can not loot him, because interaction with this actor is disabled until death animation ends.
2. If you kill a friendly actor with single attack, a game still treats this actor as "fighting" and you can not loot hit until death animation ends.
3. Excepts of followers: you need to hit the follower 4 times to make him hostile. So you can kill this actor with 3 hits and loot him during death animation, even he is summoned.
There is an exploit: you can summon Dremora or Golden Saint, kill this creature with 3 hits and loot a daedric/ebony weapon when creature's health is 0.

And more: when I tried to reproduce this exploit at 2004, I got a stable crash when taking any item from summoned creature's inventory. But I can reproduce this exploit under Wine now with same installation.

docs/source/reference/modding/settings/game.rst
```diff
@@ -65,6 +65,22 @@ the type of attack is determined by the direction that the character is moving a
...
+can loot during death animation
+-------------------------------
+
+:Type:		boolean
+:Range:		True/False
+:Default:	True
+
+If this setting is true, the player allowed to loot actors (e.g. summoned creatures) during death animation, if they are not in combat. 
+However disposing corpses during death animation is not recommended - death counter may not be incremented, and this behaviour can break quests.
+This is how original Mororwind behaves.
```
Capostrophic: A typo here.

docs/source/reference/modding/settings/game.rst
```diff
@@ -65,6 +65,22 @@ the type of attack is determined by the direction that the character is moving a
...
+
+If this setting is false, player have to wait until end of death animation in all cases.
```
Capostrophic: player has

docs/source/reference/modding/settings/game.rst
akortunov: "Mororwind"?

docs/source/reference/modding/settings/game.rst
```diff
@@ -65,6 +65,22 @@ the type of attack is determined by the direction that the character is moving a
...
 This setting can be toggled with the Always Use Best Attack button in the Prefs panel of the Options menu.
 
+can loot during death animation
+-------------------------------
+
+:Type:		boolean
+:Range:		True/False
+:Default:	True
+
+If this setting is true, the player allowed to loot actors (e.g. summoned creatures) during death animation, if they are not in combat.
```
Capostrophic: the player is allowed

akortunov:
Thank you, @Capostrophic. I updated a documentation.

scrawl:
http://en.uesp.net/wiki/Morrowind:Summon

> There is a short period of time after a summoned creature is killed during which you can loot its corpse just before it vanishes. [...] When exploiting this glitch, it is suggested to save, since the game will sometimes freeze when you try to take the item off the summoned creature.

akortunov:
Yes, this is the same exploit. You can use it in OpenMW by default, or disable it via settings file (similar to MCP fix).

And again, you can not loot creature during death animation if it was in combat.

scrawl:
> this is the same exploit

Sure about that? The UESP site doesn't mention anything about them having to not be in combat. As I recall, there was a short window of time after the death animation has finished but before the creature is deleted. I remember spamming the spacebar button as the creature would tumble to the ground, in hopes of timing it right. Is my memory failing me?

akortunov:
> there was a short window of time after the death animation has finished but before the creature is deleted

Hmm. It seems you are right. OpenMW now does not have a delay between end of death animation and corpse despawning. There is an only difference between Morrowind and OpenMW behaviour with this PR.

akortunov:
Take a look at [this](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/summoning.cpp#L107) code. Can we add a delay between creature's death and erasing it from graveyard?

scrawl:
> Take a look at this code. Can we add a delay between creature's death and erasing it from graveyard?

Yes, we can. In many cases, the graveyard persists for a longer time anyway (e.g. if you leave the cell, then wait outside - creature can't be deleted until you re enter the cell)

The easiest way would be to just run the cleanup code every N seconds, that way we don't have to track how long a creature has been dead for.

> If yes, should this delay be optional?

IMO, no.

akortunov:
And where to can we move a code from UpdateSummonedCreatures::process?

Thunderforge:
> Also adds an option to control lootiong behaviour:

> True (default) - player can loot only non-fighting actors during death animation (as in vanilla Morrowind).
> False - player can not loot all actors during death animation.

Why make it an option at all? Allowing looting of actors during the death animation can break quests, which can put the player in a zombie state. Why give users a noose to hang themselves with, so to speak?

akortunov:
> Why make it an option at all?

 I think, because we aimed to "keep vanilla game behaviour as default".

In vanilla game there are two separate bugs that allow you to to loot summoned creatures:
1. When actor is not combat, you can loot it immediately its health is 0, before death events are triggered (MCP fixes this bug), main OpenMW branch behaves in the same way for all actors.
Usually game counts fighting actors, in quests, where you need to kill something.
2. Morrowind has a small delay between end of summoned creature death animation and corpse despawn (optimization, maybe). This behaviour seems safe. OpenMW removes corpses immediately.

When I created this PR, I failed to reproduce the second bug (I was not fast enough), so I made a switch between ganilla behaviour and fixed. 
Actually, if we do not need to recreate the first bug, I will remove this option, but this is how #1121 behaves.

P.S. As for me, I am not in favor of recreating vanilla game bugs and exploits, but not all OpenMW players and developers share my opinion.

Thunderforge:
What would be the reason that someone would want to keep the first bug? Speedrunning? How big of a priority is that for OpenMW?

As for the second bug, do we know if that was ever an issue in OpenMW? I don't think it's a good idea to introduce a bug that never existed, so I'm curious to know if it's worth keeping the switch around if it can't happen.

akortunov:
@scrawl, in which form should we recreate the vanilla exploit? Both cases or only second? If latter, we do not need an option.

scrawl:
Just do the vanilla behavior in both cases for now, we can adjust it later if someone asks for it.

Thunderforge:
So if we keep the vanilla behavior and allow looting during the death animation (which allows disposing during the death animation), how will we be preventing that from breaking quests, per [bug #3528](https://bugs.openmw.org/issues/3528)?

scrawl:
That only affects non-hostile actors, and most will become hostile the second you attack them. Do we know if there are any quests potentially affected by that?

We could also just increment the kill count when a dying actor is being deleted. The chances of that breaking something should be slim, right?

akortunov:
> We could also just increment the kill count when a dying actor is being deleted.

IMO, in this case we can get a double incrementing.

Thunderforge:
> That only affects non-hostile actors, and most will become hostile the second you attack them. Do we know if there are any quests potentially affected by that?

Ah, I missed the part about it only being hostile actors. Do they still become hostile if you attack them and kill them in one hit, even if they didn't know you were there beforehand? If not, that would definitely be a problem for high level characters dealing sneak attack damage.

Two situations I can think of where a non-hostile actor can die are a scripted death and falling damage. I seem to recall that Oblivion had a scripted death with Uriel Septim so that he dies even if you prevent his assassins from reaching him. But I don't know of any scripted deaths to non-hostile actors in Morrowind. 

As for falling damage, the only planned one I can think of is Tarhiel falling from the sky. I guess if you were super fast and looted his corpse before he finishes his death animation, you would prevent the journal entry from appearing. It's minor, but it's something.

akortunov:
> Do they still become hostile if you attack them and kill them in one hit, even if they didn't know you were there beforehand? 

Yes, this behaviour was already mentioned by me and Allofich.

Thunderforge:
For what it's worth, [MCP](http://www.nexusmods.com/morrowind/mods/19510/?) seems to be disabling looting/disposing during death animations for all actors:

> - Safe dispose corpse. Actors are not fully treated as dead until their death animation is complete and they can no longer move. The final processing of an actor includes script triggers. Prevents a corpse from being disposed until all game mechanics on the actor are resolved.

Thunderforge:
@akortunov I assumed that the MCP fix I quoted was the same thing that we were implementing in this PR. If it was actually something different, I don't think we need to hold up this PR for that, given that it's one of the more serious bugs that we're currently addressing.

I think all my concerns have been answered for now, so I'm fine with this PR.

crumblingstatue:
Regardless of the resolution of this, I believe https://bugs.openmw.org/issues/3528 is an issue that doesn't exist in vanilla morrowind. I don't remember ever breaking kill quests by disposing of corpses in Vanilla morrowind. So even if we want to allow the vanilla exploits, a fix for #3528 should still be implemented.

Interestingly, it's marked as resolved, even though discussion about the fix is still ongoing.

Thunderforge:
@crumblingstatue The [bug reporting guidelines](https://wiki.openmw.org/index.php?title=Bug_Reporting_Guidelines#Explanation_of_issue_states) say that "Resolved" means

> A developer has fixed the issue, but the fix has not been reviewed by others and is not included in the official main repository (yet).

We don't really have a different state between "not reviewed" and "under active discussion", so I think that "Resolved" is the proper state.

crumblingstatue:
Applying this commit breaks pickpocketing. Attempting to pickpocket always triggers the dialogue, even when the stealth icon appears.

apps/openmw/mwclass/npc.cpp
```diff
@@ -863,16 +864,27 @@ namespace MWClass
             return action;
         }
 
-        if(getCreatureStats(ptr).isDead())
-            return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionOpen(ptr, true));
-        if(ptr.getClass().getCreatureStats(ptr).getAiSequence().isInCombat())
-            return std::shared_ptr<MWWorld::Action>(new MWWorld::FailedAction("#{sActorInCombat}"));
-        if(getCreatureStats(actor).getStance(MWMechanics::CreatureStats::Stance_Sneak)
```
crumblingstatue: I don't see this branch handled in the new code, could be why this commit breaks pickpocketing.

akortunov:
> Applying this commit breaks pickpocketing. 

Ok, I will investigate it.

apps/openmw/mwclass/npc.cpp
akortunov: Yes, you are right. I took this code from creature.cpp. My fault.

crumblingstatue:
This diff "fixes" pickpocketing, but I'm not guaranteeing it doesn't introduce additional bugs.

```diff
diff --git a/apps/openmw/mwclass/npc.cpp b/apps/openmw/mwclass/npc.cpp
index ed1f12b31..6ce44cec4 100644
--- a/apps/openmw/mwclass/npc.cpp
+++ b/apps/openmw/mwclass/npc.cpp
@@ -885,6 +885,10 @@ namespace MWClass
         if(stats.getAiSequence().isInCombat())
             return std::shared_ptr<MWWorld::Action>(new MWWorld::FailedAction(""));
 
+        if(getCreatureStats(actor).getStance(MWMechanics::CreatureStats::Stance_Sneak)
+                || ptr.getClass().getCreatureStats(ptr).getKnockedDown())
+            return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionOpen(ptr)); // stealing
+
         return std::shared_ptr<MWWorld::Action>(new MWWorld::ActionTalk(ptr));
     }
 

```

apps/openmw/mwclass/npc.cpp
akortunov: Also werewolf check was removed.

akortunov:
Thanks, @crumblingstatue. Pickpocketing and werewolf check should be fixed now.

scrawl:
> Delayed corpse despawning is out of my experience yet

What is the issue? The code is here: https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/summoning.cpp#L121 
We just need to add a check to only run that cleanup every N seconds. There are already several such timers in Actors class (for combat engagement, etc).

akortunov:
> There are already several such timers in Actors class

Ok, I found these timers. But what about [#2955](https://bugs.openmw.org/issues/2955)?
Looks like summoned creatures should be spawned instantly, but despawning has a delay (0.1 second is enough).
In this case we can split UpdateSummonedCreatures::process() to two separate functions.

scrawl:
> In this case we can split UpdateSummonedCreatures::process() to two separate functions.

Right, or add an argument e.g. ```UpdateSummonedCreatures::process(bool cleanup)```

akortunov:
I added a delay (0.2 sec). Should work as in vanilla game now.

apps/openmw/mwmechanics/summoning.hpp
```diff
@@ -22,7 +22,7 @@ namespace MWMechanics
                             float magnitude, float remainingTime = -1, float totalTime = -1);
 
         /// To call after all effect sources have been visited
-        void process();
+        void process(bool cleanup);
```
Thunderforge: Could you please add some documentation for `cleanup`? I think that it will not be immediately obvious to readers unfamiliar with this issue what it does and why it is necessary.

akortunov: Ok, I will add it, if my last commit is OK, and I will not have to looking for another implementation.

scrawl: I don't think we really need documentation for a function that's not used (or not going to be used) more than once.

PR #1352, opened 2017-07-28
remove breath/doxygen autodoc

psi29a:
This resolves the issue where RTD was failing to generate documentation due to hitting 1GiB memory constraints. We removed the autodoc/doxygen/breath which is redundant considering it is generated for us with every merge to master for us.


scrawl:
What about 'Indices and Tables' section, that can be removed too right?

psi29a:
Of course, if you think it unnecessary. There is a search on the left hand side as well.

psi29a:
I will create a special openmw-0.42 branch based on the tag and cherry-pick these commits so that we can have documentation for 0.42, but only after this MR is merged.

PR #1354, opened 2017-07-28
Add policy on original engine "fixes"

scrawl:
Since the topic came up again on #1350, I figured to write down this policy. It is what I gathered from many discussions in the past and some "bugs" that we had fixed then later reverted when complaints came in. Do we all agree on this?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,3 +33,25 @@ Furthermore, we advise to:
 * Feel free to submit incomplete pull requests. Even if the work can not be merged yet, pull requests are a great place to collect early feedback. Just make sure to mark it as *[Incomplete]* or *[Do not merge yet]* in the title.
 * If you plan on contributing often, please read the [Developer Reference](https://wiki.openmw.org/index.php?title=Developer_Reference) on our wiki, especially the [Policies and Standards](https://wiki.openmw.org/index.php?title=Policies_and_Standards).
 * Make sure each of your changes has a clear objective. Unnecessary changes may lead to merge conflicts, clutter the commit history and slow down review. Code formatting 'fixes' should be avoided, unless you were already changing that particular line anyway.
+
+Policy on original engine "fixes"
+=============================
+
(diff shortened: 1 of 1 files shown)


kcat:
Not sure that really clarifies it much. This line in particular:

"OpenMW, in its default configuration, is meant to be a faithful reimplementation of Morrowind, minus things like crash bugs, stability issues and design errors. However, we try to avoid touching anything that affects the core gameplay, the balancing of the game or introduces incompatibilities with existing mod content."

When people provide a patch to fix a replicated "bug", it would be because they see it as a design error and often doesn't really effect the core gameplay (e.g. looting summoned creatures, which is also a potential stability issue as it can demonstrably break the game balance). In the end, it still comes down to personal interpretation on a case-by-case basis, which doesn't really give a clear policy on the matter.

My personal take on it would be to fix any clearly unintended behavior (that is, things which can be logically interpreted as a flaw due to the way the vanilla engine works, or things which make no sense to occur given the established behaviors of the game), and provide config options to re-enable the more popular exploits that people can't live without.

zinnschlag:
Obviously there is a lot of room for interpretation on how we handle these kinds of bugs. I think this policy text captures the spirit of this reasonably well. Therefore a +1 from me.

I would make one minor correction:

> In the future, we may offer additional settings to turn each "bug" or exploit on/off, or allow modders to do so by means of scripting. 

These content settings are at least as likely to go into a GMST as into a script. Or it could be something completely different. I think it would best not to mention any specific implementation path in this text.

psi29a:
Looks good to me as well. Nice to have something in black and white on paper.

scrawl:
> In the end, it still comes down to personal interpretation on a case-by-case basis, which doesn't really give a clear policy on the matter

Right, that's the point. By writing this down though, we have some point of reference when, inevitably, a new controversy is started and we can apply those points to whatever is being discussed, instead of starting from scratch again.

On that note, I changed the title to "guidelines" which is probably a better name than "policy".

> I think it would best not to mention any specific implementation path in this text.

I removed that section.

PR #1362, opened 2017-08-01
Combat AI: use WhenUsed enchantments [Feedback needed]

akortunov:
Implements [feature #3953](https://bugs.openmw.org/issues/3953).

Known issues:
1. Some creatures still can not use magic items (e.g. Dagoth creatures can not use Six House amulets).

2. When actors use magic item (e.g. "ring of fireballs"), sometimes they change a stance for a half-second or so after cast, as if an item was added or removed from actor's inventory.
**FIXED**

3. When actors use items with combat enchantments (e.g. "ring of fireballs"), they will spam enchantments until a used item will be runned out of charges. This is ok from logic side, but do not look good since ActionEnchantedItem has no animation.
Maybe just allow to use only full-charged offensive items? It is similar how to vanilla Morrowind behaves.

4. A delay between casts is longer than in vanilla - OpenMW does not take in account that ActionEnchantedItem has no animation, so a round duration is shorter than with spells.
**FIXED**

Tested with:
1) Furius Acilius with his "demon katana".
2) A Golden saint with added "first barrier belt" and "ring of fireballs".
3) Dagoth Vaner - he can not use any magic items yet.

Any feedback would be greatly appreciated.

### DIFF
--- a/apps/openmw/mwmechanics/aicombataction.cpp
+++ b/apps/openmw/mwmechanics/aicombataction.cpp
@@ -71,6 +71,8 @@ namespace MWMechanics
     {
         const ESM::Enchantment* enchantment = MWBase::Environment::get().getWorld()->getStore().get<ESM::Enchantment>().find(mItem->getClass().getEnchantment(*mItem));
         int types = getRangeTypes(enchantment->mEffects);
+
+        isRanged = (types & RangeTypes::Target) | (types & RangeTypes::Self);
         return suggestCombatRange(types);
     }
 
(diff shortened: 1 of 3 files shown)


akortunov:
About 1: setSelectedEnchantItem() can be executed only for NPC with inventory store.
In OpenMW only creatures with ESM::Creature::Weapon flag (e.g. Dremora, Golden Saint) have inventory store and can use magic items. A vanilla game does not have this restriction.

PR #1363, opened 2017-08-01
Adding tutorial for placing object in game world

Thunderforge:
I couldn't create a picture for moving the ring because I was affected by [bug #3968](https://bugs.openmw.org/issues/3968).


docs/source/manuals/openmw-cs/tour.rst
```diff
@@ -215,9 +215,57 @@ in a chest and put it somewhere in plain sight. To this end we will have to
...
+Adding to a chest
 =====================
 
+*To be written…*
 
 
+Placing in plain sight
+=====================
 
+Let's hide the Ring of Night vision in the the cabin of the [Ancient Shipwreck]
```
scrawl: 'the the'

scrawl:
The images are kinda large files. Would it make sense to store the manual images in another repo? Don't see a reason why someone just building the code would need to download them.

Thunderforge:
I can probably reduce the file size to about 200 KB. Looks like a lot of the file size is because the screenshots were taken on a retina display, which provides 144 dpi.

Regardless, I'd be fine with moving the entire docs to a new repository. Even without my changes, the docs folder is 1 MB, and it will only grow as we add more pictures.

scrawl:
It might be a better idea to keep the documentation text in the main repository, so they benefit from OpenMW version tagging (as you can see the version selector on openmw.readthedocs.io), and just outsource the images. Does readthedocs support referencing external images?

Thunderforge:
> It might be a better idea to keep the documentation text in the main repository, so they benefit from OpenMW version tagging (as you can see the version selector on openmw.readthedocs.io), and just outsource the images. 

Wouldn't we run into the same problem if we needed to version the images (e.g. the user interface for OpenMW-CS changed between versions)? I would think it would be pretty easy to just up the version of the docs whenever we up the version of OpenMW.

> Does readthedocs support referencing external images?

I can't find anything in the documentation about this, and every existing docs I found uses local images. I suppose we could check in something and try it out.

scrawl:
> Wouldn't we run into the same problem if we needed to version the images (e.g. the user interface for OpenMW-CS changed between versions)?

In that case we can just add a new image with a new filename.

> I can't find anything in the documentation about this, and every existing docs I found uses local images. I suppose we could check in something and try it out.

@psi29a is there anyway we can test that without cluttering up master? Can we set up a temporary branch or 'version' to build an RTD for?

loriel2:
Existing image files are about 100KB to 200KB, and could be reduced further by converting to jpegs without losing too much quality. These should be similar.

psi29a:
@scrawl sorry, I didn't see this. I would advise against adding binaries to a repo, if necessary we can setup an apart repo just for images and other resources and have documentation link to that. I'll research if that is possible.

The other option is to have an apart repo for documentation...

psi29a:
Well, we've already added PNGs into our documentation.
I'm not thrilled about that... *shrugs*

Thunderforge:
I really screwed up this PR somehow. Replacing with #1598.

PR #1366, opened 2017-08-02
Merge redundant input events

julianko:
This is somewhat related to [Bug #3126](https://bugs.openmw.org/issues/3126).

If event handling for input, for whatever reason (recreating UI widgets etc.), takes longer than the time between SDL input events, the events will queue up. This results in an exponential increase in events being handled per frame (as long as input events are generated) because the longer frame time will result in even more input events being queued up.

To avoid this problem, this PR merges all SDL_MOUSEMOTION events together into one if there are no other events (like button presses) in between.


MiroslavR:
After this change, I can no longer reliably look around with my mouse when the _--no-grab_ option is enabled because the camera rotation keeps resetting. Reverting this change fixes the issue for me.

PR #1368, opened 2017-08-03
Trade window improvements

akortunov:
This PR:
1. Fixes [bug #3982](https://bugs.openmw.org/issues/3982).
2. Allow traders to restock items after deal (as in Morrowind), not when trade window is opening.

In vanilla Morrowind when you move an item to trader's inventory, * Up sound is played, when move to player's inventory - both * Up and * Down sounds are played. Also both sounds are played when the trader refuses to buy an item.
In OpenMW only * Down sounds are played in all cases.

With this PR * Up sound will be played when you move an item in both directions. When a trader refuses to take an item, * Down sound will be played. This behaviour is closer to vanilla than our current behaviour.

### DIFF
--- a/apps/openmw/mwgui/inventorywindow.cpp
+++ b/apps/openmw/mwgui/inventorywindow.cpp
@@ -320,7 +320,7 @@ namespace MWGui
     {
         ensureSelectedItemUnequipped(count);
         const ItemStack& item = mTradeModel->getItem(mSelectedItem);
-        std::string sound = item.mBase.getClass().getDownSoundId(item.mBase);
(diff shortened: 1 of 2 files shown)


crumblingstatue:
This "soft-breaks" old savegames that bought restockable items from a trader, and saved after, and opened that save in a new version. The trader won't restock until the next trade.

I don't know how acceptable this break is, but I'm personally fine with it.

akortunov:
> The trader won't restock until the next trade.

But after next trade the trader will restock, right?
We can add a note about this behaviour in 0.43 release.

PR #1370, opened 2017-08-03
Correcting https://bugs.openmw.org/issues/3906.

Siimacore:
Sorry for redoing a PR but I feel it's a bit cleaner.


scrawl:
Found a related bug, that was already the case before your PR though: https://bugs.openmw.org/issues/4009

PR #1371, opened 2017-08-04
Do not allow to use journal in a main menu

akortunov:
Fixes [bug #3991](https://bugs.openmw.org/issues/3991).

Note: in vanilla game you can not open a journal from some other windows (e.g. book, trade), but IMO OpenMW behaviour is better.

### DIFF
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -1022,6 +1022,7 @@ namespace MWInput
             return;
 
         if(MWBase::Environment::get().getWindowManager()->getMode() != MWGui::GM_Journal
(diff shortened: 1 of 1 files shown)


apps/openmw/mwinput/inputmanagerimp.cpp
```diff
@@ -1022,6 +1022,7 @@ namespace MWInput
             return;
 
         if(MWBase::Environment::get().getWindowManager()->getMode() != MWGui::GM_Journal
+                && MWBase::Environment::get().getWindowManager()->getMode() != MWGui::GM_MainMenu
                 && MWBase::Environment::get().getWindowManager ()->getJournalAllowed())
```
EmbraceUnity: Wouldn't you want to move the Main Menu check into the getJournalAllowed method?

PR #1377, opened 2017-08-06
Fade out music before change occurs

LennyPenny:
Fixes https://bugs.openmw.org/issues/3665
First real PR so please review :smile:

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.hpp
+++ b/apps/openmw/mwsound/soundmanagerimp.hpp
@@ -114,9 +114,15 @@ namespace MWSound
         MWBase::SoundStreamPtr playVoice(DecoderPtr decoder, const osg::Vec3f &pos, bool playlocal);
 
         void streamMusicFull(const std::string& filename);
+        void advanceMusic(const std::string& filename);
+
         void updateSounds(float duration);
         void updateRegionSound(float duration);
         void updateWaterSound(float duration);
+        void updateMusic(float duration);
+
+        float mMusicFader;
+        std::string mNextMusic;
 
         float volumeFromType(PlayType type) const;
 
--- a/apps/openmw/mwsound/soundmanagerimp.hpp
+++ b/apps/openmw/mwsound/soundmanagerimp.hpp
@@ -121,7 +121,6 @@ namespace MWSound
         void updateWaterSound(float duration);
         void updateMusic(float duration);
 
-        float mMusicFader;
         std::string mNextMusic;
 
         float volumeFromType(PlayType type) const;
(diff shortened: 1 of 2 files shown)


scrawl:
@kcat , is this code correct? I don't think we can call updateStream in the main thread when the source is also being used in the streaming thread. Also, the begin/end update isn't necessary, is it?

LennyPenny:
Well it didn't work without it

kcat:
updateStream is correct. It's just updating the volume and other related playback parameters, which are fine to do asynchronously with the streaming thread (which is just making sure the buffer queue stays full).

The start/finishUpdate calls aren't strictly necessary in this case, but are kind of recommended. It helps avoid redundant internal updates for OpenAL. Though in saying that, it may be better to avoid having to do it twice per sound update (once in updateSounds and again in updateMusic).

LennyPenny:
Okay I can confirm the working without begind/end 

Though if it's recommeded, should I just wrap all the sound updaters in a begin update?

apps/openmw/mwsound/soundmanagerimp.hpp
```diff
@@ -114,9 +114,15 @@ namespace MWSound
...
 
         void streamMusicFull(const std::string& filename);
+        void advanceMusic(const std::string& filename);
+
         void updateSounds(float duration);
         void updateRegionSound(float duration);
         void updateWaterSound(float duration);
+        void updateMusic(float duration);
+
+        float mMusicFader;
```
scrawl: mMusicFader needs to be initialized in the constructor. Even if a variable isn't supposed to be used until it's set, it is easy to make a mistake and end up with uninitialised values somewhere. (Also, tools like coverity/cppcheck will always scream at you for members not initialized in constructor)

Actually, why is this even a member variable when it's only used as a constant?

scrawl:
> Though if it's recommeded, should I just wrap all the sound updaters in a start/end?

Sounds good.

kcat:
> Though if it's recommeded, should I just wrap all the sound updaters in a start/end?

The main issue there is that the update methods won't necessarily do anything on certain updates, and calling start/finishUpdate without actually changing anything can be wasteful.

A better idea would probably be to restructure the update methods so that all necessary changes can occur within ```updateSounds```'s start/finish, and avoiding it when there's nothing to update.

apps/openmw/mwsound/soundmanagerimp.hpp
LennyPenny: Oh this is a leftover from when I thought setFadeout worked differently
Will fix

LennyPenny:
Decided to move updateMusic into the 30 times a second area because other sounds are also faded there (so no impact on listening experience) and it avoids keeping track of state when updateMusic already started an update

PR #1378, opened 2017-08-06
Implement GetPCTraveling

LennyPenny:
Fixes 3788

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -90,6 +90,13 @@ namespace MWBase
                 float x, y; // world position
                 ESM::CellId dest;
             };
+            
+            struct TravelTarget
+            {
+                ESM::Position pos;
+                std::string cellname;
+                bool interior;
+            };
 
             World() {}
 
@@ -565,6 +572,8 @@ namespace MWBase
 
             virtual void setPlayerTraveling(bool traveling) = 0;
             virtual bool isPlayerTraveling() const = 0;
+
+            virtual void travel(TravelTarget destination) = 0;
 
             /// Return terrain height at \a worldPos position.
             virtual float getTerrainHeightAt(const osg::Vec3f& worldPos) const = 0;
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -563,6 +563,9 @@ namespace MWBase
 
             virtual bool isPlayerInJail() const = 0;
 
+            virtual void setPlayerTraveling(bool traveling) = 0;
+            virtual bool isPlayerTraveling() const = 0;
+
             /// Return terrain height at \a worldPos position.
             virtual float getTerrainHeightAt(const osg::Vec3f& worldPos) const = 0;
 
--- a/apps/openmw/mwgui/travelwindow.cpp
+++ b/apps/openmw/mwgui/travelwindow.cpp
@@ -166,6 +166,9 @@ namespace MWGui
         npcStats.setGoldPool(npcStats.getGoldPool() + price);
 
         MWBase::Environment::get().getWindowManager()->fadeScreenOut(1);
+
+        MWBase::Environment::get().getWorld()->setPlayerTraveling(true);
+
         ESM::Position pos = *_sender->getUserData<ESM::Position>();
         std::string cellname = _sender->getUserString("Destination");
         bool interior = _sender->getUserString("interior") == "y";
@@ -187,6 +190,8 @@ namespace MWGui
         // Teleports any followers, too.
         MWWorld::ActionTeleport action(interior ? cellname : "", pos, true);
         action.execute(player);
+
+        MWBase::Environment::get().getWorld()->setPlayerTraveling(false);
 
         MWBase::Environment::get().getWindowManager()->fadeScreenOut(0);
         MWBase::Environment::get().getWindowManager()->fadeScreenIn(1);
--- a/apps/openmw/mwgui/travelwindow.cpp
+++ b/apps/openmw/mwgui/travelwindow.cpp
@@ -170,39 +170,7 @@ namespace MWGui
 
         MWBase::Environment::get().getWorld()->setPlayerTraveling(true);
 
-        mTravelTarget = TravelTarget{ *_sender->getUserData<ESM::Position>(), _sender->getUserString("Destination"), _sender->getUserString("interior") == "y" };
-    }
-
-    void TravelWindow::travel()
-    {
-        if (!MWBase::Environment::get().getWorld()->isPlayerTraveling())
-            return;
-
-        MWWorld::Ptr player = MWMechanics::getPlayer();
-
-        if (!mTravelTarget.interior)
-        {
-            ESM::Position playerPos = player.getRefData().getPosition();
-            float d = (osg::Vec3f(mTravelTarget.pos.pos[0], mTravelTarget.pos.pos[1], 0) - osg::Vec3f(playerPos.pos[0], playerPos.pos[1], 0)).length();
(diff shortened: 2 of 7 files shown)


scrawl:
This implementation makes no sense because there are no scripts running in between the click on the location and the location having loaded, so the getPcTravelling function could never return true. Do we know how this flag actually works in the vanilla engine?

LennyPenny:
Oh that makes sense 
I could find 3 references of this function
http://www.fliggerty.com/phpBB3/viewtopic.php?f=30&t=4462
http://harryboeck.dyndns.org/Spiele/Morrowind/Scripting%20for%20Dummies/2006-03-xx/Positioning.html
https://pastebin.com/b48xaKTY

Indicating that it returns true when in jail too. 
Maybe you can deduce from those two scripts when the function should return true while scripts are running?

LennyPenny:
Okay some further research shows that GetPCTraveling is true even after the button click but before the new region starts to load. 
Is this even possible to implement with current travel behavior? As you said no scripts run in OpenMW between clicking the button and the new cell loading

akortunov:
Can we run global scripts manually during travel, e.g. in ActionTeleport::execute()? 

We are already have something similar [here](https://github.com/OpenMW/openmw/blob/openmw-42/apps/openmw/mwworld/worldimp.cpp#L258).

scrawl:
> We are already have something similar here.

That's just a hack to run the start script before any cell is loaded, so we don't have to load any other cell before loading the (script-controlled) starting cell.

> Can we run global scripts manually during travel, e.g. in ActionTeleport::execute()?

Why just global scripts, what about local scripts?

Anyway, it seems wrong to run the scripts without running an actual frame. The frame duration (getSecondsPassed) wouldn't be correct in that case. And maybe there are scripts that somehow break if they're called twice in the same frame.

I wonder, what should happen if a script teleports you somewhere in the first frame that pcTravelling is true? Would you end up at that location or would you still end up where you wanted to travel? 
 
> Ok after some research in vanilla morrowind its true until like a frame after the game resumes.

'like' as in you're not sure if its 1 or more frames?

LennyPenny:
I made a two test scripts: 
The first one displays a messagebox (with clickable options so it doesn't disappear immediately) the first time getpcinjail is true
The second script displays a messagebox the first time getpcinjail is false after it was true
(note that only one script was enabled at a time)

The first script displayed a messagebox directly after I pressed the button to travel somwhere (even before the loading of the new cell started - so before the teleport)

The second script displayed the messagebox after the loading screen had finished. Now after some thinking this doesn't really prove that the first frame after the teleport still has it true - I mean the last frame before loading started could have had it as true and the first frame after as false and the messagebox would still only have displayed at the end. 
Do you have a suggestion for a better research script? 

Right now the only thing that is certain is that GetPCTraveling is true between clicking the button and loading and that scripts actually run between those.

LennyPenny:
Okay actually as I was totally wrong, GetPCTraveling is only true before teleportation happens. It is false as soon as you arrive. Made a video to prove my research [here](https://upl.io/i/4ct8te.mp4)

Is there a clean way to maybe delay the actual teleportation until the screen has fully faded to black? That would allow a few frames to run the game and the script logic as usual. That's maybe how it is in vanilla (I can't record at a high framerate right now, but someone might be able to catch a bit of the idle antimation while the screen is fading?)

scrawl:
I'd be fine with delaying the teleportation for one frame, but not longer. Fading transitions are just a waste of time for the player.

Still curious about this though:

> I wonder, what should happen if a script teleports you somewhere in the first frame that pcTravelling is true? Would you end up at that location or would you still end up where you wanted to travel?


Also, I wonder if the 'delay' is just a side product of cell transitions being handled differently. The way OpenMW handles it (changing cell instantly) produced some odd side effects, e.g. since teleport spells run during the middle of the animation update, the cell would be changed mid-update,  which lead to other bugs that had to be worked around. Maybe 'change cell after the next script run' should just be the default behavior for whenever a cell change is requested? Can we verify if the vanilla engine works this way for other cell change triggers like spell effects?

akortunov:
Just a guess: scripts are still executing during screen fading (there is an artifitial delay before location change).

LennyPenny:
>begin travel_tester
If (GetPCTraveling == 1)
	MessageBox "GetPCTraveling is true right now"
Endif
end travel_tester

I ran this test script and experiences some weir behavior in morrowind: 
Traveling from vivec to balmora in any order resulted in 3 messageboxes appearing (suggesting three ticks of the script engine)
Traveling from vivec to seyda need resulted in 2 messageboxes
Seyda neen to suran again 2 messageboxes
Suran to vivec 2 messageboxes
Vivec to suran 1 messagebox

Does anyone have in idea why this might be the case (maybe it's distance?)

Also those fade ins and outs were probably and active stylistic/immersive choice at least when fast traveling. It gives you more of a feel for that will happen before you arrive somewhere instead of it just feeling like an instant teleport.

scrawl:
> Does anyone have in idea why this might be the case (maybe it's distance?)

Possible. AFAIK the game passes time during travel, just like when waiting. Maybe one script tick per game hour passed?

> Also those fade ins and outs were probably an active stylistic/immersive choice at least when fast traveling. It gives you more of a feel for that will happen before you arrive somewhere instead of it just feeling like an instant teleport.

Fair point.

apps/openmw/mwgui/travelwindow.cpp
```diff
@@ -26,6 +26,7 @@ namespace MWGui
     TravelWindow::TravelWindow() :
         WindowBase("openmw_travel_window.layout")
         , mCurrentY(0)
+        , mTravelTarget{}
```
LennyPenny: This is probably not how it's supposed to be initialized. Any pointers?

LennyPenny:
@scrawl is this fine?

scrawl:
Small issue, I think: If one pressed Quickload in the frame after travel was selected then they would travel to the new location in the savegame that they just loaded. So the playerTravelling flag needs to be reset in World::clear(). Also I'd probably move the travel destination and the travel function as a whole into World as that makes more sense if we're also storing the traveling flag there.

apps/openmw/mwbase/world.hpp
```diff
@@ -566,6 +573,8 @@ namespace MWBase
             virtual void setPlayerTraveling(bool traveling) = 0;
             virtual bool isPlayerTraveling() const = 0;
 
+            virtual void travel(TravelTarget destination) = 0;
```
scrawl: inb4 cppcheck complains that TravelTarget can be passed by (const) reference.
Anyway it might be better to use ```travel(const CellID&, const ESM::Position&)```. That way we don't need to declare that TravelTarget struct, and by using CellID we won't have to change this code when extra worldspaces become supported in the future.

apps/openmw/mwgui/travelwindow.cpp
```diff
@@ -170,47 +170,18 @@ namespace MWGui
...
-        MWBase::Environment::get().getWindowManager()->fadeScreenIn(1);
-    }
-
-    void TravelWindow::onFrame()
-    {
-        if (!MWBase::Environment::get().getWorld()->isPlayerTraveling())
-            return;
-        
-        travel();
+        MWBase::Environment::get().getWorld()->travel(mTravelTarget);
```
scrawl: I would remove the onFrame() and just do all this from the World's frame update.

LennyPenny: But this would require storing const CellID&, const ESM::Position& on the WorldImp that would be rather ugly imo. 
Are you sure you want it that way?

scrawl: Not sure I understand, what's ugly about storing 2 variables?

PR #1380, opened 2017-08-07
Move menu check for journal to getJournalAllowed()

akortunov:
As EmbraceUnity suggested in #1371.


scrawl:
Don't see a reason to.

PR #1382, opened 2017-08-08
Add style for clang-format

compiletime:
Adds a style file for use with clang-format that is a close approximation to the current style throughout the codebase.

[clang-format](http://clang.llvm.org/docs/ClangFormat.html) is a refactoring tool for formatting in a custom style. It is useful because it enables other refactoring tools which are not formatting-aware, such as [clang-tidy](http://clang.llvm.org/extra/clang-tidy/) (which can safely update code to use C++11 idioms).

If this format was applied to the entire repository, there would be 85426 insertions and 87970 deletions across 1546 files. Most of this churn cannot be alleviated with an alternate style due to the following inconsistencies:
1. Namespaces are sometimes not indented.
2. Functions are sometimes spaced before parameters or arguments.
3. Arithmetic and logical operators are sometimes not spaced.

Note this commit doesn't apply the format. Developers could apply the format incrementally or it can be applied across the codebase with:
```bash
find . -iname *.hpp -o -iname *.cpp | xargs clang-format -i
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,14 @@
+---
+Language: Cpp
+BasedOnStyle: LLVM
+AccessModifierOffset: -4
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortIfStatementsOnASingleLine: true
+BreakBeforeBraces: Allman
+ColumnLimit: 0
+DerivePointerAlignment: true
+IndentCaseLabels: true
+IndentWidth: 4
+NamespaceIndentation: All
+SortIncludes: false
+...


nikolaykasyanov:
Also, `git clang-format` command can be used to format only staged changes (actual changes, not even the whole files).

scrawl:
> 85426 insertions and 87970 deletions

*gasps*

Nope, we don't accept mass changes because they lead to merge conflicts, among other things.

> Also, git clang-format command can be used to format only staged changes (actual changes, not even the whole files).

That would make a mess if you were working in a file that did not strictly adhere to our style guidelines. Using a mishmash of styles then will make it a pain to read. The general rule is to stick with whatever style is used in the file you are working on. For that you just need basic common sense while you type, nothing else.

Anyway, IMO code style is too insignificant of a problem to even waste time discussing it further. I do not see any reason for this file to be in the repository, when we won't use in any useful way.

PR #1383, opened 2017-08-08
Enable transparency for interactive messageboxes and tooltips in game mode

akortunov:
In vanilla game tooltips and most of messageboxes do not have transparency only in GUI mode. 
This behaviour makes sense: in GUI mode these widgets can be placed over other widgets.

### DIFF
--- a/apps/openmw/mwgui/messagebox.cpp
+++ b/apps/openmw/mwgui/messagebox.cpp
@@ -196,7 +196,7 @@ namespace MWGui
 
 
     InteractiveMessageBox::InteractiveMessageBox(MessageBoxManager& parMessageBoxManager, const std::string& message, const std::vector<std::string>& buttons)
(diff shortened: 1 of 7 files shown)


scrawl:
> Note: I do not know how to apply transparency to "owned" tooltip.

AFAIK the transparent texture (that is updated according to user's transparency in-game setting) is black, so you can't make it red (or whatever the 'owned' color is supposed to be), since we can only adjust colors by multiplication.
However, if we made the actual texture white, and then create the 'black' look for normal uses by setting the widget's color to black, we could use that texture for the owned background also. You might need to refactor a lot of skins though so I'm not sure if it's worth it. We can also just create another variant of the transparent bg texture that's white instead of black.

akortunov:
Added transparency to "owned" tooltip.

PR #1384, opened 2017-08-09
Wizard: Changing "New Installation" to "Retail CD/DVD" for clarity

Thunderforge:
References #3032, since it addresses the last point of that bug report. With this change, the "Select Installation Method" page now looks like this:

select installation page


PlutonicOverkill:
What about 'Retail Disc' instead of 'Retail CD'? Some versions of Morrowind (GOTY edition) came on DVD.

Thunderforge:
@PlutonicOverkill Originally I had "Retail CD/DVD", but I was unsure if it was ever distributed on DVD. Good to know that it was. Personally, I prefer that over "Retail Disc" (or "Disk"). Is there a strong preference for one or the other?

akortunov:
@Thunderforge, your build failed due to
`methodselectionpage.cpp(12): error C2065: 'newLocationRadioButton': undeclared identifier`
error.

It seems you replaced a newLocationRadioButton to retailCdRadioButton not in all cases.

Thunderforge:
@akortunov Thanks. That code block was collapsed in CLion, so I didn't see it. Fixing it now.

PlutonicOverkill:
IMO, naming doesn't matter too much, but I would use 'disc' for optical media rather than 'disk', which usually refers to magnetic storage.

Thunderforge:
@PlutonicOverkill Huh, didn't know that. Apple of all companies felt the new to [write a developer article](https://support.apple.com/en-us/HT201697) weighing in on it, which agrees with what you said.

Here's the UI after I modified that page.

select installation page v2

akortunov:
IMO, you can rebase your commits to one when you will finish your work.

Thunderforge:
Well, I seem to have thoroughly messed up this PR when I tried to squash my commits. Stay tuned for a new PR.

PR #1385, opened 2017-08-09
Wizard: Changing "New Installation" to "Retail CD/DVD" for clarity

Thunderforge:
References #3032, since it addresses the last point of that bug report. 

Before this change:

selection installation page before

After this change:

select installation page v2

### DIFF
--- a/apps/wizard/componentselectionpage.cpp
+++ b/apps/wizard/componentselectionpage.cpp
@@ -2,7 +2,6 @@
 
 #include <QDebug>
 #include <QPushButton>
-#include <QAbstractButton>
 #include <QMessageBox>
 
 #include "mainwizard.hpp"
@@ -26,7 +25,7 @@ Wizard::ComponentSelectionPage::ComponentSelectionPage(QWidget *parent) :
 
 void Wizard::ComponentSelectionPage::updateButton(QListWidgetItem*)
 {
-    if (field(QLatin1String("installation.new")).toBool() == true)
+    if (field(QLatin1String("installation.retailDisc")).toBool() == true)
         return; // Morrowind is always checked here
 
     bool unchecked = true;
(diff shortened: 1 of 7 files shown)


Allofich:
Seems a little strange to me to have it say "Select how you would like to install The Elder Scrolls III: Morrowind." It sounds like it's a given that you don't have Morrowind installed yet when you run the OpenMW wizard.

akortunov:
Maybe we can use something like "Select The Elder Scrolls 3: Morrowind data files source".

Gaerzi:
"Select installation of _The Elder Scrolls III: Morrowind_ to use with OpenMW"
- New installation (from retail disc)
- Existing installation (on hard drive)

Perhaps there could also be an option to install Morrowind from GOG or Steam?

Thunderforge:
> It sounds like it's a given that you don't have Morrowind installed yet when you run the OpenMW wizard.

@Allofich I assumed that would be the case for most users (it certainly would be for Mac/Linux users, since they can't run the vanilla installer). Do you think that's not the case?

> Maybe we can use something like "Select The Elder Scrolls 3: Morrowind data files source".

@akortunov The current name of the page is "Select Installation Method", which I've left unchanged. If we were to change the message (which may be a good idea), I'm guessing we would need to rename this page too. What would we change it to?

> Perhaps there could also be an option to install Morrowind from GOG or Steam?

@Gaerzi GOG and Steam are going to be added separately (I plan to work on Steam specifically in the near future). I don't think that "New Installation" and "Existing Installation" will be sufficient options once we add those.

psi29a:
@Thunderforge thanks for tackling this!

Keep in mind also that we also want to be able to setup OpenMW without Morrowind. The idea is that we would want to install the Example-Suite (or whatever we want to call it later) instead. This will require that we provide 'sane' defaults in the settings.cfg file so that importing of morrowind.ini isn't necessary unless you want to play Morrowind.

:)

Thunderforge:
@psi29a Glad you like it! I envisioned the Example Suite being installed a different way, given that CD/DVD, and eventually Steam and GOG.com, are Morrowind-only. [These mockups](https://forum.openmw.org/viewtopic.php?f=20&t=4372) show my vision for that. So I don't think that these changes will impact that path, since we'll be designing a new system.

psi29a:
Oh cheers then! :)  Full speed ahead. :)

Allofich:
> I assumed that would be the case for most users (it certainly would be for Mac/Linux users, since they can't run the vanilla installer). Do you think that's not the case?

I don't know. I used the vanilla installer on Windows first, then installed OpenMW after that, so I just know it doesn't apply for the way I did things, so it stuck out to me.

Thunderforge:
@Allofich What if we changed the terminology to "Load" instead of "Install"? Would that address your concerns?

> **Select Load Method**  
> Select how you would like to load *The Elder Scrolls III: Morrowind*.

scrawl:
Load sounds more confusing to me. I think we are good for now.

PR #1387, opened 2017-08-10
Check actual movement for GetPCRunning and GetPCSneaking [Feedback needed]

akortunov:
Upstream OpenMW checks only stance for these functions, Morrowind checks also actual moving.

But in upstream OpenMW getStance() can return true for both sneak and running. This behaviour leads to odd bugs. For example, if you sneak with Always run enabled, both PC Athletics and Sneak will be increased (a game treats PC as sneaking and running in the same time).

We have two options:
1) Modify getStance()
2) Use additional checks outside getStance()
Which option would be better?

### DIFF
--- a/apps/openmw/mwscript/controlextensions.cpp
+++ b/apps/openmw/mwscript/controlextensions.cpp
@@ -9,12 +9,14 @@
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/inputmanager.hpp"
+#include "../mwbase/mechanicsmanager.hpp"
 #include "../mwbase/world.hpp"
 
 #include "../mwworld/class.hpp"
 #include "../mwworld/ptr.hpp"
 
 #include "../mwmechanics/npcstats.hpp"
+#include "../mwmechanics/movement.hpp"
 
 #include "interpretercontext.hpp"
 #include "ref.hpp"
@@ -167,7 +169,11 @@ namespace MWScript
                 virtual void execute (Interpreter::Runtime& runtime)
                 {
                     MWWorld::Ptr ptr = MWBase::Environment::get().getWorld ()->getPlayerPtr();
-                    runtime.push (ptr.getClass().getCreatureStats(ptr).getStance(MWMechanics::CreatureStats::Stance_Run));
+                    const MWWorld::Class &cls = ptr.getClass();
+
+                    bool isRunning = MWBase::Environment::get().getMechanicsManager()->isRunning(ptr);
+
+                    runtime.push (isRunning && cls.getCreatureStats(ptr).getStance(MWMechanics::CreatureStats::Stance_Run));
                 }
         };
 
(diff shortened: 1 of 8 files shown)


akortunov:
I uploaded an implementation with second option used. This PR can fix [bug #4017](https://bugs.openmw.org/issues/4017) now.

apps/openmw/mwscript/controlextensions.cpp
```diff
@@ -167,7 +168,11 @@ namespace MWScript
                 virtual void execute (Interpreter::Runtime& runtime)
                 {
                     MWWorld::Ptr ptr = MWBase::Environment::get().getWorld ()->getPlayerPtr();
-                    runtime.push (ptr.getClass().getCreatureStats(ptr).getStance(MWMechanics::CreatureStats::Stance_Run));
+                    const MWWorld::Class &cls = ptr.getClass();
+
+                    bool isMoving = std::abs(cls.getMovementSettings(ptr).mPosition[0]) > .5 || std::abs(cls.getMovementSettings(ptr).mPosition[1]) > .5;
```
scrawl: Is there any reason to use > .5 instead of > 0 ? What if we implement a slower movement (e.g. for joysticks) at some point?

akortunov: I took this check from CharacterController::update. If > 0 is better, I will update PR.

scrawl:
What happens if you are overencumbered and try to run? Or if you press a movement key while jumping? Then the getPcRunning should return false, right? I don't think it does that in the current state of the PR. Maybe it would be best to just ask the character controller if the movement state is 'running'?

akortunov:
> Maybe it would be best to just ask the character controller if the movement state is 'running'?

I had theat idea yesterday, but for now I do not know how to access character controller from player pointer.

Also isSneaking() and isRunning() checks in CharacterController based on animation state. Will it be enough?

scrawl:
> I had theat idea yesterday, but for now I do not know how to access character controller from player pointer.

The character controller is mostly an implementation detail so you can't just get it from anywhere. I guess we could add a function like isRunning(ptr) to the MechanicsManager which would then pass that request to Actors and to the right CharacterController.

> Also isSneaking() and isRunning() checks in CharacterController based on animation state. Will it be enough?

I am not sure what you mean by that.

akortunov:
> I am not sure what you mean by that.

I do not know how GetPCRunning and GetPCSneaking check works in Morrowind - based on animation state or player position change.
I will do additional tests soon.

akortunov:
Ok, some test results:

> Or if you press a movement key while jumping? Then the getPcRunning should return false, right?

No, it returns true in Morrowind.

> What happens if you are overencumbered and try to run?

GetPCRunning is still returns true - a first frame of run animation played in Morrowind.
In OpenMW movement keys in this case just NOPs.

akortunov:
Personally I prefer to check a movement state, but this implementation has two differences from vanilla:
1) GetPCRunning returns false when jumping. I have no idea why it returns true in Morrowind.
2) GetPCRunning returns false when PC is overencumbered - due to overencumberance implementation in OpenMW.

akortunov:
FYI: I tested this PR with Stormrider mod.
It seems boat control works fine.

PR #1389, opened 2017-08-12
Create autosave when waiting

akortunov:
Fixes a recently reopened [bug #1452](https://bugs.openmw.org/issues/1452).


akortunov:
Looks like Bullet download links on http://www.lysator.liu.se are dead.
@ace13, can you check what's going wrong, please?

ace13:
Right, looking through the LysKOM logs it seems like the name server has died at Lysator.  
That's a little bit bothersome.

I'll poke some people and see what's happening.

PR #1391, opened 2017-08-13
AiFollow: add a threshold when turning to target

akortunov:
Fixes [bug #3749](https://bugs.openmw.org/issues/3749).

Note: in current implementation followers turn faster than in Morrowind. Is there any way to fix it?

### DIFF
--- a/apps/openmw/mwmechanics/aifollow.cpp
+++ b/apps/openmw/mwmechanics/aifollow.cpp
@@ -22,8 +22,15 @@ struct AiFollowStorage : AiTemporaryBase
 {
     float mTimer;
     bool mMoving;
+    float mTargetAngleRadians;
+    bool mTurnActorToTarget;
 
-    AiFollowStorage() : mTimer(0.f), mMoving(false) {}
+    AiFollowStorage() :
+        mTimer(0.f),
+        mMoving(false),
+        mTargetAngleRadians(0.f),
+        mTurnActorToTarget(false)
+    {}
 };
 
 int AiFollow::mFollowIndexCounter = 0;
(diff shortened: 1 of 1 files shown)


apps/openmw/mwmechanics/aifollow.cpp
```diff
@@ -171,6 +213,16 @@ std::string AiFollow::getFollowedActor()
     return mActorRefId;
 }
 
+void AiFollow::turnActorToFaceTarget(const osg::Vec3f& actorPosition, const osg::Vec3f& targetPosition, AiFollowStorage& storage)
+{
+    osg::Vec3f dir = targetPosition - actorPosition;
+
+    float faceAngleRadians = std::atan2(dir.x(), dir.y());
```
scrawl: We will get NaN here if direction is zero (actor and target in the same position)

scrawl:
> Note: in current implementation followers turn faster than in Morrowind. Is there any way to fix it?

Don't we have a speed argument to zTurn? Or do you mean there is a delay before they start turning at all?

Is there any reason we can't use the threshold function built in to zTurn ('epsilonRadians') ?

apps/openmw/mwmechanics/aifollow.cpp
akortunov: Yes, I got It.
But we have the same code in AiWander, AiFace and pathfinding.cpp (I just took already existing formula).
Should we change these files too?

akortunov:
I mean angular velocity.
Just run in circles around any follower in vanilla game and OpenMW.
Maybe I'm wrong, but looks like in OpenMW followers have higher angular velocity.

> Don't we have a speed argument to zTurn? 

I tried to use it, but without success. Even with very small values there is no noticeable difference.

apps/openmw/mwmechanics/aifollow.cpp
akortunov: Also, IIRC, std::atan2(0, 0) == 0 (atan2 should return NaN only if dir.x() is NaN or dir.y() is NaN).

scrawl:
> > Don't we have a speed argument to zTurn?
> I tried to use it, but without success. Even with very small values there is no noticeable difference

Then we should probably find out what's wrong with it, rather than adding a new method on top.

What about this one?  "Is there any reason we can't use the threshold function built in to zTurn ('epsilonRadians') ?"

akortunov:
I rewrote this PR to use epsilonRadians.

Notes:
1. I did not change pathTo() function because we need a threshold only in AiFollow.
2. Looks like followers in Morrowind turn around without playing animation. I am not sure if we want to replicate this behaviour.

PR #1393, opened 2017-08-14
Fix trading of stolen items

LennyPenny:
Works on https://bugs.openmw.org/issues/3944

### DIFF
--- a/apps/openmw/mwgui/tradeitemmodel.cpp
+++ b/apps/openmw/mwgui/tradeitemmodel.cpp
@@ -24,6 +24,18 @@ namespace MWGui
         return mItems[index];
     }
 
+    ItemModel::ModelIndex TradeItemModel::getIndex(ItemStack item)
+    {
+        size_t i = 0;
+        for (std::vector<ItemStack>::iterator it = mItems.begin(); it != mItems.end(); ++it)
+        {
+            if (*it == item)
+                return i;
+            ++i;
+        }
+        return -1;
+    }
+
     size_t TradeItemModel::getItemCount()
     {
         return mItems.size();
@@ -78,6 +90,7 @@ namespace MWGui
     {
         ItemStack item = source->getItem(itemIndex);
         item.mCount = count;
+        item.mType = ItemStack::Type_Barter;
         borrowImpl(item, mBorrowedToUs);
     }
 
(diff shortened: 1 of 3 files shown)


apps/openmw/mwgui/tradeitemmodel.cpp
```diff
@@ -24,6 +24,18 @@ namespace MWGui
...
+    ItemModel::ModelIndex TradeItemModel::getIndex(ItemStack item)
+    {
+        size_t i = 0;
+        for (std::vector<ItemStack>::iterator it = mItems.begin(); it != mItems.end(); ++it)
+        {
+            if (*it == item)
+                return i;
+            ++i;
+        }
+        return -1;
```
PlutonicOverkill: Why not just return std::find() - std::begin(), or -1 if not found? IMO this would make it more obvious what is happening here.

LennyPenny: This method was just copied over from all the other getIndex implementations around in the codebase. 
Would probably better to fix all these in a seperate PR (if it is even agreed that it is better) and fix all the code duplication while at it

kcat: ```std::distance(std::begin(), std::find())``` would be more appropriate. It compiles to the same thing for ```std::vector```, but it also handles other container types that don't use contiguous memory.

akortunov:
As I understood, this implementation only moves items, but does not decrement stolen items count.
What if I stole one item from trader, and try to sell ten items with the same ID to him?
Will he confiscate them all?
And after I pay fee and try to sell to him another ten items, he will confiscate them again, right?

P.S. If you are interested, I'm testing an alternative implementation:
https://github.com/akortunov/openmw/tree/tradeconfiscatefix
In this implementation a trader will take no more items, than you stole from him.

LennyPenny:
Oh I didn't even know that mStolenItems was a thing 
Yes currently my implementation has the problem that I don't remove stuff from mStolenitems and I also forgot to remove all items that the seller borrowed to me 

I looked at your implementation too and it has the problem that still only the first item you tried to sell to the merchant will disappear, as the loop in your implementation returns after the first iteration. 

I do not know which broken behavior you mean. but I think the policy is stated in that one document which I can't find right now. It was something along the lines of as long as its not a game breaking thing it should be carried over. 

I will just make a pull request to your branch to fix the problem of only the first itemstack being removed

akortunov:
> still only the first item you tried to sell to the merchant will disappear

Yes, and only one "This-item-belongs-to-me" messagebox will appear, as in Morrowind.
Or it would be better take ALL stacks containing stolen items?

LennyPenny:
Yes in morrowind all stolen items are taken - let me just test again

LennyPenny:
Never mind I'm dumb it's still only the first one - your implementation is entirely correct

LennyPenny:
To be honest it would be cool to give an option or something to fix all these bugs that are clearly defying logic

PR #1397, opened 2017-08-15
add transport-command

Florianjw:
The idea behind this is to provide an alternative to player->positioncell that also teleports any followers.

Having encountered two mods that involve teleport rings which add an NPC so that they can use the travel-command in order to also transport the followers, I came to the following conclusions:

* That approach is a very ugly hack
* A better solution would be to add a command that does that.
* Being able to do that nice would be yet another openmw-only feature that could help to convert people.

The implementation below is the result of a bit of experimenting but seems to do it's intended job reasonably well. One thing that should however be discussed is the opcode that I assigned for thsi command: I was unable to discover any kind of logic behind how they were assigned and therefore just picked something that was free. If I overlooked something there please adjust that part.

Of course I'll be happy to get some feedback on this before you merge it (if you plan to do so).

### DIFF
--- a/apps/openmw/mwscript/cellextensions.cpp
+++ b/apps/openmw/mwscript/cellextensions.cpp
@@ -199,6 +199,36 @@ namespace MWScript
                 }
         };
 
+        /// A command to transport the player and all his followers to make nice teleport-rings possible
+        // args: 'cell' x y z orientation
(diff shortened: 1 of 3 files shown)


zinnschlag:
Sorry, but extensions to the scripting language will have to wait until 1.0.

PR #1398, opened 2017-08-16
Get only text from input fields (bug #4025)

akortunov:
Fixes [bug #4025](https://bugs.openmw.org/issues/4025).

Now color tags should not be placed in items/spells names.

### DIFF
--- a/apps/openmw/mwgui/alchemywindow.cpp
+++ b/apps/openmw/mwgui/alchemywindow.cpp
@@ -64,7 +64,7 @@ namespace MWGui
 
     void AlchemyWindow::onCreateButtonClicked(MyGUI::Widget* _sender)
     {
-        MWMechanics::Alchemy::Result result = mAlchemy->create (mNameEdit->getCaption ());
+        MWMechanics::Alchemy::Result result = mAlchemy->create (mNameEdit->getOnlyText());
         MWBase::WindowManager *winMgr = MWBase::Environment::get().getWindowManager();
 
         switch (result)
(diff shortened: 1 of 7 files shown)


scrawl:
Did you test what happens now if you put a color tag in a name, i.e. name your savegame ```save #1``` ? If that causes issues then we need to use MyGUI::TextIterator::toTagsString to escape # characters before we set a label.

Or we could revert the PR, and change the copy/paste function so that it never copies (or never pastes) color codes. Is there any use in copying colors?

https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwgui/windowmanagerimp.cpp#L1989

akortunov:
> Is there any use in copying colors?

IMO only if a game supports text coloring. For example, if you copy green text from one widget to another, result will be green too.
But I do not see any use in copying colors in OpenMW for now. Perhaps later, if we implement a text formatting (similar to NWN2, for example).

> Did you test what happens now if you put a color tag in a name, i.e. name your savegame 

My fault, I missed this case. '#1' will not be in savegame.

> Or we could revert the PR, and change the copy/paste function so that it never copies (or never pastes) color codes

I prefer this option. When I created this PR, I just did not know about WindowManager can handle a clipboard. 

I'll try to experiment with the clipboard, and will report about results.

akortunov:
A question about MyGUI::TextIterator::toTagsString(): how can I split color tags (#RGB) from regular text ("Chest #100", for example)? We should escape only regular text.

I can just cut some chars from beginning, but it looks like a dumb solution.

PR #1399, opened 2017-08-17

h313:
This may fix #4024, which may be a result of the only guard against repetitive music being not letting the same song play twice. The memory overhead is minimal, with a mere 5 extra kb used in a music library with 1300 songs.

You may want to squash all commits when merging.

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -311,7 +311,6 @@ namespace MWSound
         if(!mOutput->isInitialized())
             return;
         std::cout <<"Playing "<<filename<< std::endl;
-        mLastPlayedMusic = filename;
         try {
             stopMusic();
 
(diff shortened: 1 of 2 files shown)


apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -367,23 +366,23 @@ namespace MWSound
...
-        }
+        //Do a Fisher-Yates shuffle
+        if(mMusicFiles.size() == 0)
+            for(int it = 0; it < filelist.size(); it++)
+                mLastPlayedMusic.push_back(it);
 
-        advanceMusic(filelist[i]);
+        int i = Misc::Rng::rollDice(mLastPlayedMusic.size());
+        advanceMusic(filelist[mLastPlayedMusic[i]]);
+        mLastPlayedMusic.erase(mLastPlayedMusic.begin()+i-1);
```
kcat: This will crash if ```i == 0``` (if the RNG happens to roll 0, or ```mLastPlayedMusic```/```mMusicFiles``` are 0 size. I don't think that ```-1```should be there. Also, ```mLastPlayedMusic``` seems to be a misnomer since it's a list of indices for playlist entries that haven't yet played.

PR #1400, opened 2017-08-17

h313:
This may fix #4024, which may be a result of the only guard against repetitive music being not letting the same song play twice. The memory overhead is minimal, with a mere 5 extra kb used in a music library with 1300 songs.

You may want to squash all commits when merging.


scrawl:
Seems to me that with this PR, whenever the list is empty and is repopulated, it is possible for the last played song to be played again, which is what we were trying to avoid in the first place.

h313:
@scrawl This should be fixed now. Do you see any other problems?

PR #1401, opened 2017-08-17

h313:
This may fix #4024, which may be a result of the only guard against repetitive music being not letting the same song play twice. The memory overhead is minimal, with a mere 5 extra kb used in a music library with 1300 songs.You may want to squash all commits when merging.

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.hpp
+++ b/apps/openmw/mwsound/soundmanagerimp.hpp
@@ -49,6 +49,7 @@ namespace MWSound
 
         // Caches available music tracks by <playlist name, (sound files) >
         std::map<std::string, std::vector<std::string> > mMusicFiles;
+        std::vector<int> mMusicToPlay; // The list of music files not played yet
         std::string mLastPlayedMusic; // The music file that was last played
 
         float mMasterVolume;
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -271,7 +271,6 @@ namespace MWSound
         return sound;
     }
 
-
     // Gets the combined volume settings for the given sound type
     float SoundManager::volumeFromType(PlayType type) const
     {
@@ -297,7 +296,6 @@ namespace MWSound
         }
         return volume;
     }
-
 
     void SoundManager::stopMusic()
     {
@@ -367,7 +365,9 @@ namespace MWSound
             }
 
             mMusicFiles[mCurrentPlaylist] = filelist;
-
+            mMusicToPlay.reserve(mMusicToPlay.size() + filelist.size());
+            for(int it = 0; it < filelist.size(); it++)
+                mMusicToPlay.push_back(it);
         }
         else
             filelist = mMusicFiles[mCurrentPlaylist];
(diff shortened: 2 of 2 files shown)


scrawl:
Why do you close and open a new PR for every commit update? That's not needed.

Code looks fine to me now. Just some minor issues with the commit history:

- The 'update to latest openmw' commit is redundant. Crash fix should be rebased (we don't want any commits that dont run). Best to just squash all commits into one. 
- Somehow your commits aren't linked to your github account. You might want to fix that, check https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

h313:
Could you just use squash commit? It currently runs for me, and the changes are only in the `soundmanagerimp` files.

apps/openmw/mwsound/soundmanagerimp.hpp
```diff
@@ -49,7 +49,8 @@ namespace MWSound
 
         // Caches available music tracks by <playlist name, (sound files) >
         std::map<std::string, std::vector<std::string> > mMusicFiles;
-        std::string mLastPlayedMusic; // The music file that was last played
+        std::vector<int> mMusicToPlay; // The music file that was last played
```
kcat: This comment is wrong.

h313:
Fixed the comment

logzero:
As a performance optimization you could replace erase with swap and pop, assuming that the order of the elements doesn't matter:

mMusicToPlay[i] = mMusicToPlay.back();
mMusicToPlay.pop_back();


And maybe reserve space before filling mMusicToPlay:

mMusicToPlay.reserve(mMusicToPlay.size() + filelist.size());

h313:
I added them. There doesn't seem to be any problem with it

scrawl:
> Could you just use squash commit? It currently runs for me, and the changes are only in the soundmanagerimp files.

Hm, there's now a "squash and merge" option on github, but I don't know what it would do to the commit message, and whether it creates a "merge PR#xxx" merge commit as usual. Can you just do the squash on your end and I'll merge it then.

Actually I'm not sure, what would squashing a merge commit do (since your 1st commit was a merge commit)? Will that make the squashed commit also a merge commit or will it just rebase so there isn't a merge commit?

h313:
That should do it. A squash takes the name of the farthest commit I think.

akortunov:
Looks like this shuffle causes crashes.
I attached a log at [bug #4024](https://bugs.openmw.org/issues/4024) topic.

akortunov:
A problem appears when a game tries to change current playlist.

akortunov:
Also what happens after tracklist switch (explore->combat->explore, for example)? I can get the same Explore track?

Looks like we need a separate mMusicToPlay for each type of playlist (for Explore and Combat track lists for now).

LennyPenny:
Maybe this has something to do with how I implemented advanceMusic in #1377?

h313:
I think the crash occurs in line 368, of `soundmanagerimp.cpp`, where `mMusicToPlay.reserve(mMusicToPlay.size() + filelist.size()); ` would increase each time a playlist switched. I'm not sure why it happens on a reload though.

akortunov:
> Maybe this has something to do with how I implemented advanceMusic in #1377

I can reproduce this crash only with latest commit in master branch.

akortunov:
> I'm not sure why it happens on a reload though.

Not only on reload. Sometimes game crashes with explore->combat transitions too.

h313:
It must be that line then. Why do you think it has to do with `advanceMusic` @LennyPenny ?

LennyPenny:
Because that was also a recent change to the soundmanager and I did not specifically test what happens when you load/save during music fades. 
But it seems that @akortunov has confirmed that advanceMusic isn't the problem

h313:
I think that #1410 should fix this

PR #1403, opened 2017-08-18
Get only text from input fields (alternative implementation)

akortunov:
Implement fix on a clipboard level (should work for all widgets now).
Adding '#' symbol to save or item name should work now too.

If this implementation is ok, we can revert my [67d59be](https://github.com/OpenMW/openmw/commit/67d59bead51ce4948c0e0256ee014bdab3bc8b5f) commit.

P.S. I am not sure about comment at lines 1987-1988. Maybe we should change it...

### DIFF
--- a/apps/openmw/mwgui/alchemywindow.cpp
+++ b/apps/openmw/mwgui/alchemywindow.cpp
@@ -64,7 +64,7 @@ namespace MWGui
 
     void AlchemyWindow::onCreateButtonClicked(MyGUI::Widget* _sender)
     {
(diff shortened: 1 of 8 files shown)


scrawl:
Wait, what? Why does UString(...).asUTF8() change anything? I don't get it.

Shouldn't we just remove that if test entirely?

scrawl:
> If this implementation is ok, we can revert my 67d59be commit.

You can add the revert commit to this PR.

akortunov:
> Shouldn't we just remove that if test entirely?

Removed. Also I added a rollback commit for my previous solution.

PR #1404, opened 2017-08-18
Owned crosshair feature improvements [discussion]

akortunov:
This PR is aimed to fix [bug #2789](https://bugs.openmw.org/issues/2789).

Since "owned crosshair" feature does not present in vanilla game, I am not sure about which behaviour we can treat as correct.

Anyway, in vanilla game there are no fines for activating owned doors or activators. What we can do here:
1) Check if an owned door is locked or trapped (unlocking a door is a Trespassing crime).
2) Check if owned activator is a bed, but I do not know how to implement this check properly.

Any ideas?

### DIFF
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -7,6 +7,8 @@
 #include <set>
 #include <stdint.h>
 
+#include "../mwworld/ptr.hpp"
+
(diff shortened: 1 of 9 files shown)


scrawl:
Yeah, I don't think there's any better way to detect beds. I guess we could scan the script for enableRestMenu but that seems overkill and isn't even guaranteed to work, so let's leave it.

PR #1405, opened 2017-08-18
Do not allow player to change weapon/spell during attack or spellcast

akortunov:
Fixes [bug #2445](https://bugs.openmw.org/issues/2445). 
Also partially fixes [bug #2835](https://bugs.openmw.org/issues/2835) (since you will have to wait until end of switching animation, your movement will be much slower).

As it works:
1) If player attacking someone, or casting spell, or switches between weapons, etc., disable weapon/spell cycling keys and weapon/spell stance toggling keys.
2) If player activates a hotkey during animation, a hotkey action will be delayed until animation ends.
If player activates many hotkeys during animation, only last action will be executed.

### DIFF
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -221,6 +221,7 @@ namespace MWBase
             virtual void keepPlayerAlive() = 0;
 
             virtual bool isReadyToBlock (const MWWorld::Ptr& ptr) const = 0;
+            virtual bool isAttackingOrSpell(const MWWorld::Ptr &ptr) const = 0;
 
             virtual void confiscateStolenItems (const MWWorld::Ptr& player, const MWWorld::Ptr& targetContainer) = 0;
 
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -212,7 +212,10 @@ namespace MWBase
             virtual void setSpellVisibility(bool visible) = 0;
             virtual void setSneakVisibility(bool visible) = 0;
 
-            virtual void activateQuickKey  (int index) = 0;
+            /// activate selected quick key
+            virtual void activateQuickKey (int index) = 0;
+            /// update activated quick key state (if action executing was delayed for some reason)
+            virtual void updateActivatedQuickKey () = 0;
 
             virtual std::string getSelectedSpell() = 0;
             virtual void setSelectedSpell(const std::string& spellId, int successChancePercent) = 0;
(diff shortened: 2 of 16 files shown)


scrawl:
Is it correct to fully disable quick keys? If I remember correctly you are supposed to be able to drink potions still, e.g. a cure paralysis while you are paralysed. Disabling spells cycling probably isn't needed either, since the spell that is actually cast is predetermined the moment you start casting. 

I wonder if the code would be cleaner if, instead of the GUI/Inputmanager checking if the action can be executed, we just ask the character controller to do it for us as soon as it is possible. This way we could apply the same restrictions to players and NPCs alike.

Another question is, if I press un-ready spell while casting a spell, should that be queued for later or should it be a no-op? And what if I pressed twice? I am not sure what is the most intuitive behavior here. How does the vanilla game do it?

scrawl:
> disable weapon/spell cycling keys

Note that changing the weapon via the inventory will still work, unless we added another check for that. If we do, then we would want to remove this line from the [features](https://wiki.openmw.org/index.php?title=Features#Controls) page "The annoying "Finish your current attack before equipping a new weapon" error has been removed and instead plays the weapon equip animation." (not that I'd be too sad about losing this "feature")

akortunov:
>  if I press un-ready spell while casting a spell, should that be queued for later or should it be a no-op?

No-op.

> Is it correct to fully disable quick keys?

Am I disabled quick keys? Quick key action will be delayed until end of animation. 

> you are supposed to be able to drink potions still, e.g. a cure paralysis while you are paralysed.

Only via inventory. Quick keys in Morrowind are no-ops when you are paralyzed.
But we can allow to use potions and ingredients via quck keys during animation or paralyze.
In this PR I did not add a paralyze check for now.

> Note that changing the weapon via the inventory will still work

For now - yes. I have no idea which behaviour is better - Morrowind or OpenMW one.

> we just ask the character controller to do it for us as soon as it is possible

Ok, I'll try to implement it.

scrawl:
By the way, in the current PR it seems a bug that we do not clear mActivatedIndex in WindowManager::clear() (e.g. in case another game is loaded)

akortunov:
I uploaded a slightly improved implementation.

> we do not clear mActivatedIndex in WindowManager::clear()

Should be fixed now.

> Note that changing the weapon via the inventory will still work, unless we added another check for that.

This check was added.

> Disabling spells cycling probably isn't needed either

Cycling works for both spells and magic items, including weapons with When Used enchantment (e.g. Demon Katana). If we keep it enabled, player will equip such weapons.

> you are supposed to be able to drink potions still

For now you can use any item excepts weapon via hotkey, if player is busy.

PR #1407, opened 2017-08-19
Add an EditorConfig file to force compatible editors to use the correct indentation style

AnyOldName3:
A bunch of IDEs and Text editors now support EditorConfig either natively or via a plugin. Adding this file should force all those programs to use the correct indentation style when people are working on OpenMW, even if the default setting is something else.

More information is here: http://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*.cpp]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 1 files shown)


AnyOldName3:
Something weird happened with the OSX CI build and Qt, even though this shouldn't have had any effect on it. Someone else might have to figure out what and why.

PR #1408, opened 2017-08-19
Add VS2017 aps files to gitignore

AnyOldName3:
VS2017 now creates APS files when loading OpenMW. This adds them to the gitignore file to prevent them being accidentally committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ Doxygen
 .settings
 .directory
 .idea
+files/windows/*.aps
 ## qt-creator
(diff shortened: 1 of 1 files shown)


LennyPenny:
Are you sure these *.aps files are in the git root? 
Would make more sense if they were created in the solution root which is MSVC*/ depending on your config

AnyOldName3:
That's where they appear on my machine, and I definitely didn't put them there myself. It might be a better idea to just ignore anything with the file extension.

PR #1410, opened 2017-08-20
fix extensive memory usage in mMusicToplay

h313:
This bug may have caused the segfault in #1401

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -365,7 +365,7 @@ namespace MWSound
             }
 
             mMusicFiles[mCurrentPlaylist] = filelist;
-            mMusicToPlay.reserve(mMusicToPlay.size() + filelist.size());
+            mMusicToPlay.reserve(filelist.size());
             for(int it = 0; it < filelist.size(); it++)
                 mMusicToPlay.push_back(it);
         }


akortunov:
Look like it fixes crash on reload. But shuffle still conflicts with music fading:

`MWSound::SoundManager::advanceMusic (this=0x4d94490, filename=<error reading variable: Cannot create a lazy string with address 0x0, and a non-zero length.>)
`
`at /home/andrei/Downloads/openmw-local/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:337
Cannot create a lazy string with address 0x0, and a non-zero length.>`
`in MWMechanics::Actors::update (this=this@entry=0xa9607a0, duration=duration@entry=0.0319060013, paused=paused@entry=false) at /home/andrei/Downloads/openmw-local/openmw/apps/openmw/mwmechanics/actors.cpp:1286`

I got this crash randomly if I attack someone (in my testing - Ranis Athrys).

But I do not know if there is an issue in music fading or shuffle itself.

h313:
@akortunov Weird. I'm not getting this problem when I attack someone.

Could you try adding `std::cout << filelist[mMusicToPlay[i]] << std::endl;` in front of `advanceMusic(filelist[mMusicToPlay[i]]);` on line 393 in `apps/openmw/mwsound/soundmanagerimp.cpp` and send in the output?

scrawl:
:reviewfail:  :S

I suppose this line ``` if(mMusicFiles.size() == 0)``` was probably supposed to be ```if(mMusicToPlay.size() == 0)``` as I do not think the former could be true at all (by the way please use .empty() rather than size()==0 or a certain tool will scream at you)

Ignoring that, the code still doesn't make sense. mMusicToPlay is never being cleared and the fact there is more than one playlist isn't accounted for.

I've reverted it all for now so you will have to revert my revert commit or start new.

LennyPenny:
I can confirm the game sometimes crashing when attacking someone

h313:
I will need to close this and create a new pull request because of the conflicts

PR #1412, opened 2017-08-20
Improved shuffle

h313:
Changes from previous commit #1401 :

- Use a `std::map` so that each playlist will have unique values
- Fixed the `mMusicToPlay` growing larger each time `startRandomTitle()` was called
- Documentation

I tested entering and going out of combat and did not encounter any crashes.

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -271,7 +271,6 @@ namespace MWSound
         return sound;
     }
 
-
     // Gets the combined volume settings for the given sound type
     float SoundManager::volumeFromType(PlayType type) const
     {
@@ -297,7 +296,6 @@ namespace MWSound
         }
         return volume;
     }
-
 
     void SoundManager::stopMusic()
     {
@@ -368,6 +366,12 @@ namespace MWSound
 
             mMusicFiles[mCurrentPlaylist] = filelist;
 
+            // Build mMusicToPlay for this playlist
+            std::vector<int> temp;
+            temp.reserve(filelist.size());
+            for(int it = 0; it < filelist.size(); it++)
+                temp.push_back(it);
+            mMusicToPlay.insert(std::make_pair(mCurrentPlaylist, temp));
         }
         else
             filelist = mMusicFiles[mCurrentPlaylist];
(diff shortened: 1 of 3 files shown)


kcat:
It's better to use ```std::unordered_map``` instead of ```std::map```, as the former can use a hash lookup and better pack entries. Might also be a good idea to use references or iterators to avoid doing multiple lookups in succession. Also, perhaps it would be better to rewrite the patch(es) instead of reverting a revert, which makes it confusing about what commits are actually applied (scrawl or Zini can comment on what they prefer).

h313:
Squashing this patch should work. Also, I switched it to a `std::unordered_map` and use a reference to get to `mMusicToPlay[mCurrentPlaylist]`. Thanks for the feedback!!

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -368,22 +366,39 @@ namespace MWSound
 
             mMusicFiles[mCurrentPlaylist] = filelist;
 
+            // Build mMusicToPlay for this playlist
+            std::vector<int> temp;
+            temp.reserve(filelist.size());
+            for(int it = 0; it < filelist.size(); it++)
```
akortunov: GCC complains about "comparison between signed and unsigned" here. IMO "it" should be unsigned int.

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -368,22 +366,39 @@ namespace MWSound
...
+        int i = Misc::Rng::rollDice(tracklist.size());
 
-        // Don't play the same music track twice in a row
-        if (filelist[i] == mLastPlayedMusic)
+        // Repopulate if playlist is empty
+        if(tracklist.empty())
         {
-            i = (i+1) % filelist.size();
+            tracklist.reserve(filelist.size());
+            for (int it = 0; it < filelist.size(); it++)
```
akortunov: Also I got "comparison between signed and unsigned" here.

akortunov:
Looks like a game does not crash with this PR.

apps/openmw/mwsound/soundmanagerimp.cpp
kcat: Probably should be ```size_t``` actually (or ```std::size_type```, but it's the same thing and less typing).

kcat:
An int can handle 2147483647, which is plenty. FWIW, the "C++ way" to fill the vector like you're trying to do is:
```
temp.resize(filelist.size());
std::iota(temp.begin(), temp.end(), 0);
mMusicToPlay.insert(std::make_pair(mCurrentPlaylist, std::move(temp)));
```
Or maybe simpler:
```
auto &list = mMusicToPlay[mCurrentPlaylist];
list.resize(filelist.size());
std::iota(list.begin(), list.end(), 0);
```
That way you don't have to worry about implementing the loop properly with the right types and such, ```std::iota``` will do it for you.

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -368,22 +368,34 @@ namespace MWSound
...
         }
         else
             filelist = mMusicFiles[mCurrentPlaylist];
 
         if(filelist.empty())
             return;
 
-        int i = Misc::Rng::rollDice(filelist.size());
+        // Do a Fisher-Yates shuffle
+        int i = Misc::Rng::rollDice(tracklist.size());
```
scrawl: Seems to me the roll should be done after the 'Repopulate if playlist is empty', or else the first track in the playlist will be more likely to play since it will always play just before a re-shuffle.

h313:
Is this PR ready for merging?

h313:
@scrawl is this PR ready to be merged?

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -368,22 +368,35 @@ namespace MWSound
 
             mMusicFiles[mCurrentPlaylist] = filelist;
 
+            // Build mMusicToPlay for this playlist
+            tracklist.resize(filelist.size());
+            std::iota(tracklist.begin(), tracklist.end(), 0);
```
scrawl: Aren't these two lines redundant, since we do the exact same thing later on line 384?

h313:
Fixed.

PR #1414, opened 2017-08-22
[Incomplete] Implement terrain texture selection

PlutonicOverkill:
Again, sorry that this has taken so long, but here's what I've done so far regarding terrain selection. This mostly seems to work, except that it crashes in SceneUtil::LightManager::accept() with the error "vector iterator not dereferencable". I don't know what could be causing this, especially since lighting is disabled here. I also get crashes relating to deleted records, but these occur in OpenCS anyway, so it probably isn't anything here that's causing it.

Previously I had a more elaborate implementation that automatically handled conversions between different types of selection (texture, vertex, etc.) without losing information, but when I started running into problems (mainly to do with keeping everything synced between cells) I decided to scrap it and start again. I'm still not sure that this is the right way to do it, since it works completely differently to all the other selection code, but hopefully something like this should be OK.


PlutonicOverkill:
I forgot to mention, the visual part of the selection (in TerrainTextureSelection::update()) is mostly just temporary, and can be made a lot more efficient. This was just for testing purposes.

PlutonicOverkill:
Looks like constexpr is not fully supported in VS2013. Apparently the Windows builds are [currently being built with VS2013](http://forum.openmw.org/viewtopic.php?p=48830#p48830), but this might change in the future, so it is OK to keep it for now?

apps/opencs/view/render/terraintextureselection.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+void CSVRender::TerrainTextureSelection::update()
+{
+    const ESM::Land::LandData* landData {getLandData()};
+
+    if (!landData)
+        return;
+
+    mVertices->clear();
```
Aesylwinn: I believe it is preferred to create a new vertex array instead of modifying the old one, to avoid issues with multithreaded osg.

scrawl: Indeed, you can't just modify vertices. Check https://wiki.openmw.org/index.php?title=Rendering_Architecture#Threading_considerations for explanation and solutions.

scrawl:
> This mostly seems to work, except that it crashes in SceneUtil::LightManager::accept() with the error "vector iterator not dereferencable". I don't know what could be causing this, especially since lighting is disabled here.

LightManager::accept() will usually be contained in the callstack since it's near the root of the scene graph, but that doesn't mean the crash has anything to do with it. The stack frames will basically resemble a path through the scene graph, with the lowest frame being the node currently being processed. What stack frame is it crashing on, or what line? Do you have a debug stack trace?

> I also get crashes relating to deleted records, but these occur in OpenCS anyway, so it probably isn't anything here that's causing it.

As in, also occuring in master?

PlutonicOverkill:
OK, this is pretty weird. I still get a crash in OSG even in a build from master branch, so it can't be anything I'm doing. I've pushed a commit so that Drawables in the scene graph aren't modified, although I can't test it due to the aforementioned crash. Is this the sort of thing you meant?

I also read on the wiki page that Geodes are deprecated, so how is geometry supposed to be added to the scene graph instead?

> LightManager::accept() will usually be contained in the callstack since it's near the root of the scene graph, but that doesn't mean the crash has anything to do with it. The stack frames will basically resemble a path through the scene graph, with the lowest frame being the node currently being processed. What stack frame is it crashing on, or what line? Do you have a debug stack trace?

Yeah, the crash didn't occur here directly, but I'm using OSG binaries built without debug symbols so I couldn't tell what function it actually occurred in. Maybe it's just an unstable OSG version that's causing it?
> As in, also occuring in master?

Can confirm that it occurs in master. Steps to reproduce: select "Create a new game" from the menu, go to World->Cells and add a new exterior cell. Right click on the record and click "View", and then click "Delete record". Finally change the editing mode and OpenCS will crash. The issue is caused by Cell::setSelection() calling Pathgrid::clearSelected() when mPathgrid is nullptr.

>Well, we can't merge anything the CI won't build.

VS2013 is getting a bit old now, and if it won't compile standard C++11, it might be time to think about dropping support (as long as nobody is using it for development). And since VS2017 is now supported, it could be added to the Appveyor builds to replace VS2013.

apps/opencs/view/render/terraintextureselection.cpp
```diff
@@ -0,0 +1,109 @@
...
+    mGeode = new osg::Geode{};
+    mGeode->addDrawable(geometry);
+    getBaseNode()->addChild(mGeode);
+}
+
+void CSVRender::TerrainTextureSelection::addToSelection(osg::Vec3d worldPos)
+{
+    const TerrainPos localPos {toTextureCoords(worldPos)};
+
+    if (std::find(mSelection.begin(), mSelection.end(), localPos) == mSelection.end())
```
Aesylwinn: Two of the travis builds are failing here with 
/home/travis/build/OpenMW/openmw/apps/opencs/view/render/terraintextureselection.cpp:33:65: error: no matching function for call to ‘find(std::vector<std::pair<int, int> >::iterator, std::vector<std::pair<int, int> >::iterator, const TerrainPos&)’

apps/opencs/view/render/terraintextureselection.cpp
```diff
@@ -0,0 +1,109 @@
...
+{
+    const ESM::Land::LandData* landData {getLandData()};
+
+    if (!landData)
+        return;
+
+    const auto geometry = static_cast<const osg::Geometry*>(mGeode->getDrawable(0));
+    const osg::ref_ptr<osg::Geometry> newGeometry {
+        geometry
+        ? new osg::Geometry{*geometry, osg::CopyOp::SHALLOW_COPY}
```
Aesylwinn: You're creating a shallow copy and then later retrieving and modifying the drawArrays from your shallow copy. I'm pretty sure you need a new drawArrays too, though I may be wrong. Personally, I would move all geometry and mesh creation to this function, since it would simplify your logic.

PlutonicOverkill: I took this line from [here](https://github.com/OpenMW/openmw/blob/ada85aa1ee24e32435083cf644faece988ba9a27/apps/openmw/mwrender/characterpreview.cpp#L94) where it does a shallow copy. About doing geometry creation in this function, I thought it would be too inefficient, but I can change it if it makes it clearer.

apps/opencs/view/render/terraintextureselection.cpp
PlutonicOverkill: Oops, looks like I forgot to `#include <algorithm>`.

apps/opencs/view/render/terraintextureselection.cpp
Aesylwinn: I'd say it's probably best to start simple and optimize only after a performance issue is noticed. Anyway, going by the name, a shallow copy would only copy the pointer, so if you go that route a new drawArrays needs to be created instead of accessing the old one.

Aesylwinn:
Regarding the osg crashes, you could see if a clean build fixes it. Otherwise you probably want to get a more stable version of osg.

PlutonicOverkill:
OK, thanks for the help. I've got mock exams next week but if I get time I'll try to compile my own build of OSG with debug symbols so I can find out what's going on with the crashes.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <osg/Group>
+
+#include "../../model/world/cellcoordinates.hpp"
+
+#include "worldspacewidget.hpp"
+
+namespace
+{
+    constexpr int cellSize {ESM::Land::REAL_SIZE};
```
akortunov: Looks like "constexpr" is not supported by Visual Studio 2013.

PlutonicOverkill: [Already spotted](https://github.com/OpenMW/openmw/pull/1414#issuecomment-323972038) :)
I'll probably just change it to 'const' for now, but I guess VS2013 support might have to be dropped sooner or later, especially if OpenMW ever upgrades to C++14.

scrawl:
> I also read on the wiki page that Geodes are deprecated, so how is geometry supposed to be added to the scene graph instead?

Added directly. The same way you added a Geode.

apps/opencs/view/render/terraintextureselection.cpp
scrawl: > Anyway, going by the name, a shallow copy would only copy the pointer, so if you go that route a new drawArrays needs to be created instead of accessing the old one.

This is correct.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    if (isInCell(hit)) {
+        toggleSelection(hit.worldPos);
+        update();
+    }
+}
+
+void CSVRender::TerrainSelection::activate()
+{
+    mParentNode->addChild(mBaseNode);
```
scrawl: Is there any way activate() could potentially be called twice? In that case we need to add a check to make sure mBaseNode won't be added to the scene twice. Even if that doesn't happen, we should probably still add the check just to be sure.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void CSVRender::TerrainSelection::deactivate()
+{
+    mParentNode->removeChild(mBaseNode);
+}
+
+CSVRender::TerrainSelection::TerrainPos CSVRender::TerrainSelection::toTextureCoords(osg::Vec3d worldPos) const
```
scrawl: Non-primitive types should be passed by const reference.

apps/opencs/view/render/terrainselection.cpp
PlutonicOverkill: From <osg/Group> header:
> If node is not NULL and is not contained in Group then increment its reference count, add it to the child list and dirty the bounding sphere to force it to recompute on next getBound() and return true for success. Otherwise return false.

It seems that if a node is already contained within a Group, adding it again is just a no-op. Does this mean that no check is needed?

apps/opencs/view/render/terrainselection.cpp
PlutonicOverkill: Vec3d is a small class, only consisting of three doubles, so I assumed it would be OK for passing by value. I will change it if this is a problem though.

apps/opencs/view/render/terrainselection.cpp
scrawl: Well, that documentation appears to be wrong. If you look at the [source](https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osg/Group.cpp#L77), it will only check for duplicates if ENSURE_CHILD_IS_UNIQUE is defined, which is never defined anywhere.

PlutonicOverkill:
I finally managed to get an OSG build working with OpenMW, and I can't reproduce the crashes anymore. So hopefully I should be able to get back to working on this soon.

zinnschlag:
Haven't seen any activity for quite a while. Is this PR still active? I am asking because this is an important feature that blocks most progress on terrain editing.

akortunov:
Is this PR still needed?

akortunov:
Superseded by PR1769.

PR #1415, opened 2017-08-22
Escape hash characters in content file names so the loading screen wi…

AnyOldName3:
…ll display them correctly

Since I tweaked the config file loading so content files with a hash in their name could actually be loaded, the loading screen text was interpreting any hash characters as a command to change the text colour. Now they're displayed properly.

### DIFF
--- a/apps/openmw/mwgui/loadingscreen.cpp
+++ b/apps/openmw/mwgui/loadingscreen.cpp
@@ -1,5 +1,7 @@
 #include "loadingscreen.hpp"
 
+#include <boost/algorithm/string/replace.hpp>
+
 #include <osgViewer/Viewer>
 
 #include <osg/Texture2D>
@@ -88,7 +90,8 @@ namespace MWGui
     {
         mImportantLabel = important;
 
-        mLoadingText->setCaptionWithReplacing(label);
+        std::string escaped = boost::replace_all_copy(label, "#", "##");
+        mLoadingText->setCaptionWithReplacing(escaped);
         int padding = mLoadingBox->getWidth() - mLoadingText->getWidth();
         MyGUI::IntSize size(mLoadingText->getTextSize().width+padding, mLoadingBox->getHeight());
         size.width = std::max(300, size.width);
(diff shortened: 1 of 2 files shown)


akortunov:
Maybe you should use built-in MyGUI toTagsString() function instead of boost? 

Something like:
`mLoadingText->setCaptionWithReplacing(MyGUI::TextIterator::toTagsString(label));`

AnyOldName3:
That's probably a better idea. I didn't know that function existed, so I went with what I knew.

scrawl:
This is broken, as there are some uses of setLabel that expect to be able to use escape sequences, e.g. #{sLoadingMessage3}

The line
```
mLoadingText->setCaptionWithReplacing(MyGUI::TextIterator::toTagsString(label));
```

is weird in itself, as 'setCaptionWithReplacing' suggests you intend to use escape sequences, but then you escape any escape sequences that may found in the given label so actually they can't be used.

AnyOldName3:
I've only modified the bit of code which concerns displaying esp/esm/omwaddon/omwgame file names as they're loaded when OpenMW is launched, so nothing should ever be displayed by this code that wants to be formatted.

The `setCaptionWithReplacing` probably should never have been there in the first place, and in an earlier version of this change, I tried changing it to the regular `setCaption` but that wasn't enough to fix the issue (as MyGUI always interprets a hash followed by six characters as a command to change the colour), so it needed other things changing too. The original version which used Boost would have preserved stuff like new lines, so the with replacing version would still have helped.

If you can point me at where something like `#{sLoadingMessage3}` is passed to the same code, I'll tweak my change.

scrawl:
Well, #{sLoadingMessage3} is literally what you see when you launch the game with this PR. There may be other such messages. I suggest we just escape the content filenames specifically. 

https://github.com/OpenMW/openmw/search?utf8=%E2%9C%93&q=%23%7BsLoadingMessage3%7D&type=

AnyOldName3:
I see the problem. Because I didn't see anything else with Visual Studio's 'View Call Hierarchy' feature, I assumed it was never called anywhere else. I can fix this easily.

PR #1417, opened 2017-08-23
Editor: Fix wrong folder name

Aesylwinn:
Videos are in the "video" folder and not in the "videos" folder for me. Seems strange that this wasn't noticed for so long. Did the folder name change at some point?


zinnschlag:
The reason for nobody noticing this issue is probably that the editor is still not fully functional and therefore not widely used. No wonder then that nobody ever used something as peripheral as video resources.

PR #1419, opened 2017-08-23
AiWander: add offset when forwarding only if destination is occupied

akortunov:
This PR is aimed to fix [#3638](https://bugs.openmw.org/issues/3638), but IMO we need a discussion first.

A solution, applied in this PR, is a quite simple: just add an offset only if destination pathgrid is already occupied by another actor.

Also we can use a more complex, but more safe solution: iterate over all allowed pathgrids, and find a free one. If there are no such pathgrids, just leave wandering actor in place (uploaded [here](https://github.com/akortunov/openmw/tree/forwardingfix2)). This solution allows to avoid moving actor to random position.

Which option would be better?

### DIFF
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -191,6 +191,9 @@ namespace MWBase
             virtual void getObjectsInRange (const osg::Vec3f& position, float radius, std::vector<MWWorld::Ptr>& objects) = 0;
             virtual void getActorsInRange(const osg::Vec3f &position, float radius, std::vector<MWWorld::Ptr> &objects) = 0;
 
+            /// Check if there are actors in selected range
+            virtual bool isAnyActorInRange(const osg::Vec3f &position, float radius) = 0;
+
             ///Returns the list of actors which are siding with the given actor in fights
             /**ie AiFollow or AiEscort is active and the target is the actor **/
             virtual std::list<MWWorld::Ptr> getActorsSidingWith(const MWWorld::Ptr& actor) = 0;
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -1621,6 +1621,17 @@ namespace MWMechanics
         }
     }
 
+    bool Actors::isAnyObjectInRange(const osg::Vec3f& position, float radius)
+    {
+        for (PtrActorMap::iterator iter = mActors.begin(); iter != mActors.end(); ++iter)
+        {
+            if ((iter->first.getRefData().getPosition().asVec3() - position).length2() <= radius*radius)
+                return true;
+        }
+
+        return false;
+    }
+
     std::list<MWWorld::Ptr> Actors::getActorsSidingWith(const MWWorld::Ptr& actor)
     {
         std::list<MWWorld::Ptr> list;
--- a/apps/openmw/mwmechanics/actors.hpp
+++ b/apps/openmw/mwmechanics/actors.hpp
@@ -115,14 +115,16 @@ namespace MWMechanics
             bool isRunning(const MWWorld::Ptr& ptr);
             bool isSneaking(const MWWorld::Ptr& ptr);
 
-        void forceStateUpdate(const MWWorld::Ptr &ptr);
+            void forceStateUpdate(const MWWorld::Ptr &ptr);
 
-        bool playAnimationGroup(const MWWorld::Ptr& ptr, const std::string& groupName, int mode, int number, bool persist=false);
-        void skipAnimation(const MWWorld::Ptr& ptr);
-        bool checkAnimationPlaying(const MWWorld::Ptr& ptr, const std::string& groupName);
-        void persistAnimationStates();
+            bool playAnimationGroup(const MWWorld::Ptr& ptr, const std::string& groupName, int mode, int number, bool persist=false);
+            void skipAnimation(const MWWorld::Ptr& ptr);
+            bool checkAnimationPlaying(const MWWorld::Ptr& ptr, const std::string& groupName);
+            void persistAnimationStates();
 
             void getObjectsInRange(const osg::Vec3f& position, float radius, std::vector<MWWorld::Ptr>& out);
+
+            bool isAnyObjectInRange(const osg::Vec3f& position, float radius);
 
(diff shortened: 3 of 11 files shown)


scrawl:
> Also we can use a more complex, but more safe solution: iterate over all allowed pathgrids, and find a free one. If there are no such pathgrids, just leave wandering actor in place (uploaded here). This solution allows to avoid moving actor to random position.

In that case any cell with more (or as many) actors as there are pathgrid nodes would be unlikely to see any AI fast-forwarding, right?.

scrawl:
Instead of adding an offset in a random XY-direction, maybe we could move a small amount towards a neighbouring pathgrid node. That way if there is a neighbouring pathgrid node upstairs, the actor will also move up on the Z axis to account for the stairs. If this isn't random enough, we could combine it with an additional offset that's orthogonal to the chosen path.
Unfortunately, this solution somewhat relies on the pathgrid nodes matching the incline of the stairs, and I'm not sure if that is always the case.

akortunov:
> In that case any cell with more (or as many) actors as there are pathgrid nodes would be unlikely to see any AI fast-forwarding, right?.

Yes, you are right. That's why I uploaded here an another variant.

akortunov:
Anyway, I am not sure, that moving an actor to a random position in complex 3D-world is a very good idea (especially in interiors). For now we can not guarantee that wandering actor will not be placed inside another object ([bug #3852](https://bugs.openmw.org/issues/3852), for example).

> we could move a small amount towards a neighbouring pathgrid node

Also we can move a small amount towards **random** neighbouring pathgrid node. IMO that should be enough.

> the actor will also move up on the Z axis

It is a quite hard to determine an actual Z offset amount that should be added. Just add an arbitrary high value?

akortunov:
Also can we **move** an actor instead of **teleporting** him? This is how script commands such as SetPos work in Morrowind.

scrawl:
> Anyway, I am not sure, that moving an actor to a random  position in complex 3D-world is a very good idea (especially in  interiors). For now we can not guarantee that wandering actor will not  be placed inside another object (bug #3852, for example).

Pathgrid nodes tend to have a bit of space around them, since actors have to be able to walk past easily. The only thing that's causing problems here is a change in altitude, right?

As a crazy idea, we could "simulate" a certain amount of wandering by running the actual AI packages and physics simulation. Seems like way too much work though, and it would break the ability for NPCs to go past doors in their fast-forwarding.

> Also can we move an actor instead of teleporting him? 

Where's the difference?

> It is a quite hard to determine an actual Z offset amount that should be added. Just add an arbitrary high value?

The position could be determined based on following the path to the next node, using a linear interpolation. For example, if the actor lands on a node at (0,0,0), the next node is (100,0,50), and we moved the actor 10% of the way to that next node, then the new position would be (10,0,5).

akortunov:
> Where's the difference?

Moving takes in account objects collisions, teleporting just changes object position.
But yes, looks like this is an overkill for this case.

Just a note: some Morrowind mods with interactive transport (Stormrider, Abot's Silt striders, etc.) use SetPos to emulate collisions and gravity. This trick does not work in OpenMW for now since SetPos in OpenMW teleports a target, not move.

akortunov:
I tried to use a linear interpolation, but it does not work as intended. What am I doing wrong?

akortunov:
Ok, a slightly improved uploaded.

Known issues: 
1. For now I have to add Z offset manually due to faulty underground pathgrids in Maar Gan.
Morrowind suffers from [bug #3638](https://bugs.openmw.org/issues/3638) too.
Should we consider this as a resource bug?
2. Looks like OpenMW does not change NPC position immediately, so isAnyActorInRange() check does not work properly.
How can I check if another actor was moved to the target position just now?
Note: Morrowind does not prevent overcrowding and places NPC's exactly in the target pathgrid position.

scrawl:
> Looks like OpenMW does not change NPC position immediately, so isAnyActorInRange() check does not work properly.

Are you sure? Fast-forwarding uses moveObject() which should be instant.

scrawl:
>  For now I have to add Z offset manually due to faulty underground pathgrids in Maar Gan.
>     Morrowind suffers from bug #3638 too.
>   Should we consider this as a resource bug?

If it's an exterior, would it help to cap by the terrain height?

A small offset should be ok. 80 seems a bit high though. Are you sure there won't be problems with getting stuck in the ceiling (in interiors) in that case?

akortunov:
> If it's an exterior, would it help to cap by the terrain height?

We are already do this, but a ladder with [this](https://i.imgur.com/c0JzrpS.jpg) problematic pathgrid     is a static.
It seems an engine fails to resolve [this](https://i.imgur.com/BM2pEtz.jpg) case. An actor sometimes falls through the ladder even if you place him exactly in the faulty pathgrid location. This behaviour is different from a common one ([bug #3803](https://bugs.openmw.org/issues/3803)).
Can we do something on the physics side instead of adding an arbitrary Z offset?

> Pathgrid nodes tend to have a bit of space around them

Normally yes, but there are some exceptions.

akortunov:
> Are you sure? Fast-forwarding uses moveObject() which should be instant.

And mActors in actors.cpp should be changed instantly too?

scrawl:
> Can we do something on the physics side instead of adding an arbitrary Z offset?

I suppose it's possible to attempt moving up in small increments until we are 'free' (basically the opposite of 'tracing down' where we move down until there is a collision, it's just that tracing down can be implemented more efficiently as bullet supports it natively without having to use small increments). 

> And mActors in actors.cpp should be changed instantly too?

Changed how? It just holds a pointer to the actor, that is aware of his own position, so that should be updated, yes.

akortunov:
Works quite well now. How it behaves:
1. Selects a random allowed destination path node
2. Checks if the destination point is busy
2.1. If no, move to the destination path node
2.2. If yes:
2.2.1. Get a list of neighboring nodes for the destination
2.2.2. If the list is empty, return 
2.2.3. Try to move to the random neighboring node (for 5,10, 15% of distance)
2.2.4. Checks if the destination node is busy
2.2.5. If no, move to the modified destination point
2.2.6. If yes, remove that neighboring node from list and repeat from 2.2.2

Notes:
1. [#3638](https://bugs.openmw.org/issues/3638) is not fixed perfectly, but it is much rare now.
2. We can make the algorithm more complex if we need: we can repeat it untill we will find a free allowed destination path node.

scrawl:
Sorry, forgot about this one. I will test now and merge if all is well.

PR #1420, opened 2017-08-24
Editor: Change default topic info values

Aesylwinn:
Going off the original morrowind data, it appears that having Rank, Gender, and PCRank unset is a more common case than having them set to 0, Male, and 0. These defaults should also be more user friendly. [Bug #4037](https://bugs.openmw.org/issues/4037) would suggest these are the defaults for the Morrowind Editor.


Aesylwinn:
I tested creating new Topic/Journal infos and cloning them as well.

PR #1424, opened 2017-08-29
Display negative values on fatigue progress bars

akortunov:
Vanilla Morrowind shows negative fatigue values on scroll bars, in OpenMW minimum value is 0.
Now it matches vanilla behaviour.


MiroslavR:
I think you've missed the tooltip on the HUD fatigue bar.

akortunov:
> I think you've missed the tooltip on the HUD fatigue bar.

Fixed now.

Note: a code might be cleaner, but MyGUI::ProgressBar treats negative values as full bar, not empty (at least in MyGUI 3.2.2).

PR #1426, opened 2017-08-31
Allow to play only one copy of given sound at time (bug #3647)

akortunov:
Fixes [bug #3647](https://bugs.openmw.org/issues/3647).

According to @Allofich (in #1145), only one copy of given sound can be played on single object at time. When a new copy is started, a game stops a previous copy.

Note: we have **two** playSound3D functions:
1. Play a sound on given object.
2. Play a sound at given position (e.g. for area spells).
I modified only first one.

Tested with imperial shrines and locked doors.


scrawl:
How's this affect footstep sounds? Are they or are they not supposed to overlap when moving very fast?

akortunov:
Is there any problems with footsteps?
I did not notice a difference between Morrowind and this PR.
A sound **still** will be played after every step, but **previous** sound copy will be stopped.
And if I remembered correctly, there are different sounds for left and right steps.

PR #1427, opened 2017-09-01
Editor: Land and Land Texture Tables

Aesylwinn:
This is what I've got so far. Basic table functionality is implemented for the LTEX tables. You can clone/add LTEX records and change the name/texture. You can also "touch" LAND records, so that they will use the current plugin land textures. Viewing the changes in the scene widgets does not work yet, but the effects can be seen by running openmw with the addon loaded.

[Forum post](http://forum.openmw.org/viewtopic.php?f=7&t=4322)
[LTEX feature](https://bugs.openmw.org/issues/933)
[LAND feature](https://bugs.openmw.org/issues/936)


Aesylwinn:
Loading land data later doesn't play nicely with the editor at the moment. The land load code may need some significant refactoring.

Aesylwinn:
My current list of what remains to be done:
1. Figure out the best way to solve the problem with land data delayed loading.
2. Connect and handle signals to update the terrain in the scene view.
3. Import land textures when cloning lands.
4. Add the ability to merge multiple land textures into one.

I'll be taking a break from this for a few days.

Aesylwinn:
I'm getting some strange artifacts with the terrain system. I think it may have to do with colors.
[](https://postimg.org/image/sqsim82sl/)

Aesylwinn:
Seems to only be an issue with cloned/touched lands which both copy the data.

Aesylwinn:
Turned out to be an array bounds issue.

Aesylwinn:
Land texture merging can be implemented later with the automatic land texture deletion feature, since both will require indexing the textures used in each land texture.

Aesylwinn:
Just a heads up, I'm likely to lose power/internet access at some point tomorrow, and it may take a few days to come back up, so I may not respond to feedback in a reasonable amount of time. Also, I'll admit I was rushing this toward the end, so I would suggest taking time with the reviewing process, especially in loadland.cpp and columnimp.cpp. I'll review it myself in a day or two assuming I'm able to.

zinnschlag:
No hurry. Extremely busy right now, so I won't get around to do the review for a couple of days. Entirely possible that you will beat me to it with your own final review.

psi29a:
I'm so glad you have taken this task on. I'll review this as well in the coming days.

Aesylwinn:
Alright, take your time.

components/terrain/world.cpp
```diff
@@ -75,4 +75,10 @@ void World::updateTextureFiltering()
     mTextureManager->updateTextureFiltering();
 }
 
+void World::clearAssociatedCaches()
+{
+    mTextureManager->clearCache();
```
scrawl: I don't think we need to clear mTextureManager when Land/LandTexture changes. It caches plain Textures, not LandTextures.

Aesylwinn: You're correct.

apps/opencs/model/world/columnimp.cpp
```diff
@@ -1,28 +1,261 @@
...
+        }
+    }
 
-    record.setModified(record2);
-}
+    void LandMapLodColumn::set(Record<Land>& record, const QVariant& data)
+    {
+        QByteArray array = data.toByteArray();
+        const signed char* rawData = reinterpret_cast<const signed char*>(array.data());
+        assert (array.count() == Land::LAND_GLOBAL_MAP_LOD_SIZE);
```
zinnschlag: That looks problematic. We shouldn't shut down the editor in such a case, because it would result in data loss. An exception would be a better option.

apps/opencs/model/world/columnimp.hpp
```diff
@@ -60,6 +63,20 @@ namespace CSMWorld
         }
     };
 
+    template<>
+    inline QVariant StringIdColumn<Land>::get(const Record<Land>& record) const
+    {
+        const Land& land = record.get();
+        return QString(Land::createUniqueRecordId(land.mX, land.mY).c_str());
```
zinnschlag: IIRC that is not the correct way to convert between std::string and QString (at least not for Qt4). There should be plenty of examples in the code on how to convert properly.

apps/opencs/model/world/columnimp.hpp
```diff
@@ -2417,7 +2434,125 @@ namespace CSMWorld
             return true;
         }
     };
-    
+
+    template<typename ESXRecordT>
```
zinnschlag: I don't understand this whole section. Why are these templates?

apps/opencs/model/world/columnimp.hpp
```diff
@@ -2417,7 +2434,125 @@ namespace CSMWorld
...
+    struct TextureHandleColumn : public Column<ESXRecordT>
```
zinnschlag: Haven't finished studying the code so I am not quite sure yet what TextureHandleColumn is used for. But it sounds like that is the texture resource ID (basically the filename)? If yes, we should call it that, because I don't think we use the terminology "handle" anywhere else.

apps/opencs/model/world/commands.cpp
```diff
@@ -15,6 +16,173 @@
...
+}
+
+CSMWorld::ImportLandTexturesCommand::ImportLandTexturesCommand(IdTable& landTable,
+    IdTable& ltexTable, QUndoCommand* parent)
+    : QUndoCommand(parent)
+    , mLands(landTable)
+    , mLtexs(ltexTable)
+    , mOldState(0)
+{
+    setText("Import land textures");
```
zinnschlag: Command descriptions are visible to the user. The terminology import seems strange here. It will most likely cause confusion.

apps/opencs/model/world/commands.cpp
```diff
@@ -15,6 +16,173 @@
...
+}
+
+void CSMWorld::ImportLandTexturesCommand::redo()
+{
+    int pluginColumn = mLands.findColumnIndex(Columns::ColumnId_PluginIndex);
+    int oldPlugin = mLands.data(mLands.getModelIndex(getOriginId(), pluginColumn)).toInt();
+
+    // Original data
+    int textureColumn = mLands.findColumnIndex(Columns::ColumnId_LandTexturesIndex);
+    mOld = mLands.data(mLands.getModelIndex(getOriginId(), textureColumn)).toByteArray();
```
zinnschlag: Just noticed that you are using ByteArray all over the place. Is there a specific reason not to simply use std::vector instead? We generally prefer C++ STL containers over Qt containers, unless there is a very good reason to do otherwise.

components/esm/loadland.cpp
```diff
@@ -175,6 +175,48 @@ namespace ESM
 
     }
 
+    void Land::blank()
+    {
+        if (mLandData)
+        {
+            delete mLandData;
```
zinnschlag: Best to set mLandData to nullptr here so we avoid a dangling pointer in case of an exception.

zinnschlag:
Here is my first review pass. Besides a couple of minor issues that I commented on inline I see one major problem. In the LandTexture table there are multiple records with the same texture index (sometimes even more than 2). That is confusing to the user and is not useful. For all intents and purposes LTEX records that are not part of the currently edited plugin don't exist.
Furthermore when I touched a Land record from base, the resulting LandTexture records had very high texture indices (4 digits or so, while everything else seems to be barely above 100. That looks odd and may require further investigation.

components/esm/loadland.cpp
kcat: Also you don't need to check for null before calling ```delete```. Like ```free```, it's a no-op to delete null pointers (some code checkers will complain about a redundant null check).

apps/opencs/model/world/columnimp.hpp
Aesylwinn: [Qt5](http://doc.qt.io/qt-5/qstring.html#QString-7)
[Qt4](http://doc.qt.io/qt-4.8/qstring.html#QString-8)
I can use the another method in case we plan on defining QT_NO_CAST_FROM_ASCII at some point.

apps/opencs/model/world/columnimp.hpp
Aesylwinn: I made them templates because that was what I observed the other columns doing (even for non-generic data), and I assumed there was a reason for it. For example, mUseValue is only in one esm structure (a class called Skill), but also has a template based column.

apps/opencs/model/world/columnimp.hpp
Aesylwinn: From my understanding, it is only the name displayed to the user.

apps/opencs/model/world/commands.cpp
Aesylwinn: As far as I know, you can't store an STL container in a QVariant.

Aesylwinn:
> In the LandTexture table there are multiple records with the same texture index (sometimes even more than 2). That is confusing to the user and is not useful.

Should we just not show Base LandTextures then?

> Furthermore when I touched a Land record from base, the resulting LandTexture records had very high texture indices (4 digits or so, while everything else seems to be barely above 100.

When importing textures that conflict with existing indices (and have a different texture or texture handle/name), a new index is chosen. I spread out these indices because it should reduce the number of future conflicts and I saw no downsides.

components/esm/loadland.cpp
Aesylwinn: Looking at how it is used, I don't even need to delete it. If it exists then I can reuse it.

apps/opencs/model/world/commands.cpp
zinnschlag: There is probably a way to do that, but that is actually a good enough reason to not use STL containers.

zinnschlag:
> Should we just not show Base LandTextures then?

That sounds reasonable.

> When importing textures that conflict with existing indices (and have a different texture or texture handle/name), a new index is chosen. I spread out these indices because it should reduce the number of future conflicts (when selecting new indices) and I saw no downsides.

Fair enough. But have you considered the case of multiple editing operations that require creating new LTEX indices. If this is done again and again over multiple content files, would this trick still help with avoiding conflicts and are we at risk of running out of indices eventually.

apps/opencs/model/world/columnimp.hpp
zinnschlag: Just looked it up. Here is an example from our code base:

> QString::fromUtf8 (mText.c_str())

apps/opencs/model/world/columnimp.hpp
zinnschlag: Hm, you are right. We have almost all columns as templates. I started out this way because I didn't know in advance which columns I would have to re-use for other record types. Probably a bit wasteful. Should have looked more closely and avoided making columns templates that clearly only belong to one table.
Nvm, then. We can look at that another time as part of a general code review.

apps/opencs/model/world/columnimp.hpp
zinnschlag: Yeah, it seems so. Doesn't make any sense to me since LTEX records are identified by Index and not by ID. Looks redundant. Maybe we are missing something here.

@scrawl Do you have any further input regarding this topic.

apps/opencs/model/world/idtable.cpp
```diff
@@ -281,3 +304,89 @@ CSMWorld::CollectionBase *CSMWorld::IdTable::idCollection() const
...
+            const LandTexture& oldLtex = dynamic_cast<const Record<LandTexture>&>(idCollection()->getRecord(oldRow)).get();
+            const LandTexture& newLtex = dynamic_cast<const Record<LandTexture>&>(idCollection()->getRecord(newRow)).get();
+            if (oldLtex.mId == newLtex.mId && oldLtex.mTexture == newLtex.mTexture)
+            {
+                // It's a match
+                results.recordMapping.push_back(std::make_pair(id, newId));
+                break;
+            }
+
+            // Determine next index. Spread out the indices to reduce conflicts.
```
Aesylwinn: We are indeed at risk of running out of indices. However, that limit is 65534. The looping scheme I came up with iterates over all of these indices, though it does not increment how you would traditionally. The gcd of 65534 and 8191 is 1, thus the lcm of the two numbers is 65534 * 8191. Each iteration, 8191 gets added to the index. If you start with an index of 0, then it shouldn't be too difficult to convince yourself that it iterates through all indices.

Aesylwinn: I'm debating if I should construct a map allowing reverse lookups. The benefit is that matching existing LandTexture records will be faster with huge data sets. Also, it would be more accurate for matching LandTexture records from different plugins (an oversight that I just realized). However, I have concerns with keeping the map up to date. It is possible to access the IdCollection directly from the Data class, completely bypassing the IdTable, which could cause some future issues. I suppose the best course of action would be to subclass IdCollection and store the reverse lookup map there. For picking indices, there is no reason to spread out the indices if a reverse lookup map is being used to find matches.

apps/opencs/model/world/commands.cpp
Aesylwinn: I've been looking into it a bit, and it is indeed possible. I had previously come to the mistaken conclusion that QVariant was very limited with what could be passed. I'll make an attempt to use an STL container, but I can at least get rid of the need for reinterpret_cast.

apps/opencs/model/world/idtable.cpp
zinnschlag: Are there performance issues? If not then there is really no reason to introduce error-prone optimisations.

Aesylwinn: Say plugin 1 has an LTEX with grass.dds and index 10 while plugin 2 has an LTEX with grass.dds and index 30. With the current implementation, these will most likely not be matched. I'd say the justification to change it would be for that reason, and not for performance reasons.

Aesylwinn: I'm assuming the above problem I mentioned is not acceptable. 

I think I've come up with a compromise. The importTextures function is only called once for each touch/clone command and is passed all the ids to import for one of these operations. Thus, building a map each time is not entirely wasteful, and more optimal than the alternative which is to iterate over every LTEX record for each passed id. I can test the performance by touching every LAND record, and if it is not satisfactory then we can further discuss this.

apps/opencs/model/world/columnimp.hpp
scrawl: As far as I remember, yes, this field is redundant.

apps/opencs/model/world/idtable.cpp
zinnschlag: Sounds reasonable.

psi29a: \o/

Aesylwinn:
Not sure if their is a good way to hide the base records. I have made it so they appear grayed out for now.

For containers, I've chose to use a QVector over an std::vector because the QVector should work better with how QVariants work (lots of copies). It would not be difficult to change though.

The performance using a map for texture inputs was decent. On my laptop with an i5-5200u processor it takes 1-2 seconds to touch all land records with Morrowind, Tribunal, and Bloodmoon loaded.

Should the TextureHandle column (which I renamed to TextureNickname) be removed or kept?

Aesylwinn:
Are the textures always case insensitive regardless of the filesystem? I noticed that there are some duplicate textures as a result of having a different case, and if that never matters then we should ignore the case. If it does matter, then there is an issue with the auto completion which seems to always suggests lower case names.

zinnschlag:
Everything in Morrowind is case-insensitive (stupid Windows!). If OpenMW treats anything non case-sensitive then that is a bug.

psi29a:
wait, what?  Shouldn't that read:
If OpenMW treats anything case-sensitive then that is a bug.

Everything should be case-insensitive, thanks to being handicapped by Windows. Right?

zinnschlag:
Oops. Yes, you are right. The last sentence was incorrect.

Aesylwinn:
I'm getting some errors in the console, but I think they are the result of recent changes and not mine.

>Error: RigGeometry rendering with no skeleton, should have been initialized by UpdateVisitor
>before Glyph::subload(): detected OpenGL error: invalid value
>Warning: detected OpenGL error 'invalid value' at after RenderBin::draw(..)

Aesylwinn:
I get the same errors on the master branch. Regarding the hiding of base LandTexture records, perhaps that could done using the existing search/sort mechanisms?

zinnschlag:
>  Regarding the hiding of base LandTexture records, perhaps that could done using the existing search/sort mechanisms?

Don't see how that works since what is displayed in the table is not determined by search/sort. User a filter may be an option if it doesn't interfere with other filtering functions. The only other option I see is to manually change the mapping between records in the collections and rows in the table model.

Aesylwinn:
> The only other option I see is to manually change the mapping between records in the collections and rows in the table model.

I don't think we want to do that since we only want to hide the records from the user, and the tables can be accessed for other uses. I'll see if I can figure out a way to make filtering work.

Aesylwinn:
I think everything has been addressed now.

zinnschlag:
I'll merge that for now. I don't see anything wrong with it, but we definitely need more testing.

I think we still need to think over the "texture nickname" column. It does not make any sense to me. And AFAIK this field isn't used anywhere. Maybe we should just hide it? Or maybe get some input from other modders on the forum first?

With this done the next task will be terrain texture selection. @PlutonicOverkill has done some work on it but I don't know how complete it is. @Aesylwinn Do you want to continue to work on the terrain feature? Unless PlutonicOverkill wants to go on with it, maybe you could take that over his work (https://github.com/OpenMW/openmw/pull/1414).

psi29a:
\o/

This is great news for the next release!

Aesylwinn:
I've gone ahead and posted a [thread](http://forum.openmw.org/viewtopic.php?f=7&p=50205#p50205) on the forums asking for opinions about the unused field. I'm somewhat busy at the moment, so I probably wouldn't be able to continue with the LAND/LTEX related features for at least a month, if PlutonicOverkill does not wish to continue with it.

@psi29a I'm afraid this was more grounds work for terrain editing, so there isn't much you can/should do with it. Touching land records and changing the textures referred to should not be preferred over creating a texture with the same name for mod compatibility reasons.

PlutonicOverkill:
Hi, I am moving house next week, and since this is a very busy time of year, I can't do any work on the terrain selection feature during the next few weeks. I've got OSG built, but still haven't managed to get everything correctly linked and running, and until that happens I won't be able to continue with it. I also have to admit, I get a bit lost whenever a discussion comes up surrounding LAND and LTEX records, so I'll need to research those properly as well.

Aesylwinn:
The files components/esm/loadland.hpp and components/esm/loadltex.hpp are good places to start. I believe there is a small discussion about LTEX records in the forum post linked in the opening post. You probably also want to look at apps/opencs/model/world/columnimp.hpp for the way to access and set the relevant data. Looking at how this PR "imports" LTEX records would probably also be beneficial. (files are off the top of my head, hopefully they are correct).

Seeing how PlutonicOverkill wants to continue, I'll start working on some of the recent issues that have popped up once I have some time available.

PR #1432, opened 2017-09-06
Update alchemy effects after every created potion (#4079)

akortunov:
Fixes [bug #4079](https://bugs.openmw.org/issues/4079).

A question: should we bother to use needsUpdate or can just update potion effects every time?


akortunov:
A first build was failed due to undeclared mClient variable. I have no idea why it worked before.

MiroslavR:
> I have no idea why it worked before.

I assume due to https://github.com/MyGUI/mygui/commit/b88ca1b124ab5ad7010d087ed4ff4517d49af063. Not sure if overriding `mClient` was intended or not.

PR #1438, opened 2017-09-10
make move/rotate/scale-marker-width proportional to object-radius

Florianjw:
This makes the width of the edit-markers in the opencs-object-mode proportional to the radius of the object (width = 0.05 * radius). The motivation behind this change is that the markers used to completely hide small objects so that moving and especially rotating them became pretty much guesswork. The same is even true for selecting them: Place ten candles in a row and try to select them in any order but front to back.

With this change all those things become easy by having the markers very small for very small objects, so that those are not obscured, but still very thick for huge objects.

Things to consider:

* Provide seperate markers for move and scale so that it is obvious which is active
* set minimum width or disable the current minimum length for the move/scale-arrows
* possibly a different scale-factor, though the current one seems to work fine.

### DIFF
--- a/apps/opencs/view/render/object.cpp
+++ b/apps/opencs/view/render/object.cpp
@@ -30,10 +30,7 @@
 #include "mask.hpp"
 
 
-const float CSVRender::Object::MarkerShaftWidth = 30;
 const float CSVRender::Object::MarkerShaftBaseLength = 70;
-const float CSVRender::Object::MarkerHeadWidth = 50;
-const float CSVRender::Object::MarkerHeadLength = 50;
 
 
 namespace
@@ -215,6 +212,10 @@ osg::ref_ptr<osg::Node> CSVRender::Object::makeMoveOrScaleMarker (int axis)
 
     float shaftLength = MarkerShaftBaseLength + mBaseNode->getBound().radius();
 
+    const float MarkerShaftWidth = mBaseNode->getBound().radius() * 0.05;
+    const float MarkerHeadWidth = MarkerShaftWidth * (5.0 / 3.0);
+    const float MarkerHeadLength = MarkerHeadWidth * 1.5;
+
     // shaft
     osg::Vec3Array *vertices = new osg::Vec3Array;
 
@@ -305,6 +306,7 @@ osg::ref_ptr<osg::Node> CSVRender::Object::makeRotateMarker (int axis)
 {
     const float Pi = 3.14159265f;
 
+    const float MarkerShaftWidth = mBaseNode->getBound().radius() * 0.05;
     const float InnerRadius = mBaseNode->getBound().radius();
     const float OuterRadius = InnerRadius + MarkerShaftWidth;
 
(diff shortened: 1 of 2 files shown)


Aesylwinn:
>The motivation behind this change is that the markers used to completely hide small objects so that moving and especially rotating them became pretty much guesswork.

Your changes certainly fixes that problem. However, one issue with this is that it becomes impossible to select an axis to move an item when the camera is at farther distances. There is a forum post somewhere that has discussed the markers before and some of the pitfalls with different approaches.

In any case, I think we should merge this. Being able to perform detailed work with smaller objects is more important than being able to easily move them across multiple cells. In the latter case, people can workaround it by removing the instance and re-adding it in the other cell.

Aesylwinn:
Personally I'd go for a larger scaling factor, because you don't have to be far at all from the object before you cannot select an axis to rotate/move/scale along. Perhaps you could make those magic numbers user configurable so that people could change them as they wish.

zinnschlag:
> Being able to perform detailed work with smaller objects is more important than being able to easily move them across multiple cells.

I am pretty sure you are only saying that because you never had to execute this task at a larger scale.

Sorry, that won't do. We need a smarter solution that covers both working with small objects and with objects from a distance. I am not aware of the state of said discussion/forum thread (probably happened while I was semi-away). But IMO we would want something that scales with distance (meaning the size of the markers in pixels does not depend on the distance between camera and object) but still follows certain size constraints (meaning when the camera moves closer the markers don't get so small that they disappear inside the object). Unfortunately I have no idea how to do that. I think there is an OSG node that can handle the former. But that doesn't address the later.

If we can not find a way to do the above (or any other better solution, because this is just a quick off-the-cuff idea), we could offer the user both options (the original one and the one in this PR). 
This would have to be done at runtime of course, which would require replacing all markers on a mode change. I am thinking of a pop-up menu on the object sub-mode buttons (move, scale, rotate) that contains a single tickable menu-item ("Scaling markers" or something).

Aesylwinn:
> I am pretty sure you are only saying that because you never had to execute this task at a larger scale.

I'm arguing that working with smaller stuff in detail is a more common case than moving things large distances. And it's not that the latter is impossible, it's just difficult to use the axis to do so. You are right though that I lack the experience, so I could be wrong. I was thinking of this solution as a stop gap measure, until someone comes up with a better solution that works everywhere. 

Anyway, providing the option for both is just another workaround, but it should temporarily solve the issue for now. It also shouldn't take a considerable amount of effort to accomplish.

scrawl:
> The motivation behind this change is that the markers used to completely hide small objects so that moving and especially rotating them became pretty much guesswork. The same is even true for selecting them: Place ten candles in a row and try to select them in any order but front to back.

How about instead making the markers semi-transparent, or make the selected object always render in front of the marker?

Aesylwinn:
Interesting, I tried making the marker render on top of everything before, but didn't try transparency. It's worth a shot.

rhtucker:
I don't know if this has been discussed elsewhere before, but I never saw a problem with using the x, y, and z keys to limit the axis of translation as the original CS had. If there's a hotkey to hide the markers, then you'd still be able to move objects at any distance.

Aesylwinn:
Related [feature request](http://bugs.openmw.org/issues/3501).

zinnschlag:
Transparency sounds like a very good idea. Of course with a configurable alpha.

scrawl:
#1543

PR #1439, opened 2017-09-10
Take in account elemental shields for GetResist and SetResist script commands 

akortunov:
Fixes [bug #4093](https://bugs.openmw.org/issues/4093).

### DIFF
--- a/apps/openmw/mwscript/statsextensions.cpp
+++ b/apps/openmw/mwscript/statsextensions.cpp
@@ -1189,6 +1189,14 @@ namespace MWScript
                 if (mNegativeEffect != -1)
                     currentValue -= effects.get(mNegativeEffect).getMagnitude();
 
+                // GetResist* should take in account elemental shields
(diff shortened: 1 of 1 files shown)


scrawl:
Not sure if a special case is good. In this case SetResist(GetResist()) would result in changing the value, right?

akortunov:
Well, a test case from vanilla Morrowind:
1. A Dark Elf PC (75% Resist Fire)
2. "fire shield" spell (10 pts Fire Shield, so +10% Resist Fire)

Testing:
1. Check PC Resist Fire:
player->GetResistFire
75

2. Cast Fire Shield, and check PC Resist Fire again:
player->GetResistFire
85

3. Set Resist Fire to 85, check again:
player->SetResistFire 85
player->GetResistFire
85

4. Wait until Fire Shield expires (just rest or dispel), chack again:
player->GetResistFire
75

So both GetResist and SetResist should take in account elemental shields.
OpenMW should work in the same way with this PR now.

I do not know how to achieve this behaviour without special case for now. Any ideas?

PR #1442, opened 2017-09-10
JL - Various Patches

Langerz82:
https://bugs.openmw.org/issues/4092
New Spell Random Priority Generator - Designed to avoid defensive spell spamming.

### DIFF
--- a/apps/openmw/mwmechanics/aicombataction.cpp
+++ b/apps/openmw/mwmechanics/aicombataction.cpp
@@ -2,10 +2,12 @@
 
 #include <components/esm/loadench.hpp>
 #include <components/esm/loadmgef.hpp>
+#include <components/misc/rng.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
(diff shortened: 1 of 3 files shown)


akortunov:
Generally, you should use "one feature - one pull request" rule.
Also mixing different features/fixes in a single commit is a very bad practice.

And I am not sure that adding a random generator to combat action selection is a good idea. This change can lead to odd behaviour, also it will make combat AI harder to debug.

apps/openmw/mwmechanics/aicombataction.cpp
```diff
@@ -230,18 +233,48 @@ namespace MWMechanics
             }
         }
 
+
+        // New Spell Random Priority Generator.
```
Capostrophic: This line only makes sense in the context of this pull request. I suggest to remove it and the line 277 as well..

scrawl:
Closed as not a pull request. Please separate your changes and explain the reasoning, testing, issues etc. (see contributing.md file)

PR #1443, opened 2017-09-10
[discussion] Improve hotkey usage for 'close container' and 'take all items' actions

akortunov:
This is a proof of concept for now.

Allows to fix [bug #1555](https://bugs.openmw.org/issues/1555), [bug #4016](https://bugs.openmw.org/issues/4016), and partially implements [feature #4092](https://bugs.openmw.org/issues/4092).

A summary of changes:
1. Use Activate button instead of hardcoded Spacebar to close a container window.
2. Use ToggleWeapon button instead of hardcoded Enter to take all items from the container window.
3. Resolve these actions in InputManager, so actions, assigned to Spacebar and Enter, should not be triggered now.

Questions:
1. I had to modify setPlayerControlsEnabled(). Have we a better option?
2. Have we a better button to use instead of ToggleWeapon to take all items?
3. If you loot a corpse, should "take all" action also dispose it?

### DIFF
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -196,6 +196,8 @@ namespace MWBase
             ///< show extra info in item tooltips (owner, script)
 
             virtual bool getFullHelp() const = 0;
+            virtual void closeContainer() const = 0;
+            virtual void takeAllItemsFromContainer() const = 0;
 
             virtual void setActiveMap(int x, int y, bool interior) = 0;
             ///< set the indices of the map texture that should be used
--- a/apps/openmw/mwgui/container.cpp
+++ b/apps/openmw/mwgui/container.cpp
@@ -46,7 +46,6 @@ namespace MWGui
 
         mDisposeCorpseButton->eventMouseButtonClick += MyGUI::newDelegate(this, &ContainerWindow::onDisposeCorpseButtonClicked);
         mCloseButton->eventMouseButtonClick += MyGUI::newDelegate(this, &ContainerWindow::onCloseButtonClicked);
-        mCloseButton->eventKeyButtonPressed += MyGUI::newDelegate(this, &ContainerWindow::onKeyPressed);
         mTakeButton->eventMouseButtonClick += MyGUI::newDelegate(this, &ContainerWindow::onTakeAllButtonClicked);
 
         setCoord(200,0,600,300);
@@ -156,14 +155,6 @@ namespace MWGui
         setTitle(container.getClass().getName(container));
     }
 
-    void ContainerWindow::onKeyPressed(MyGUI::Widget *_sender, MyGUI::KeyCode _key, MyGUI::Char _char)
-    {
-        if (_key == MyGUI::KeyCode::Space)
-            onCloseButtonClicked(mCloseButton);
-        if (_key == MyGUI::KeyCode::Return || _key == MyGUI::KeyCode::NumpadEnter)
-            onTakeAllButtonClicked(mTakeButton);
-    }
-
     void ContainerWindow::resetReference()
     {
         ReferenceInterface::resetReference();
@@ -251,6 +242,11 @@ namespace MWGui
 
             mPtr = MWWorld::Ptr();
         }
+    }
+
+    void ContainerWindow::takeAll()
+    {
+        onTakeAllButtonClicked(mTakeButton);
     }
 
     void ContainerWindow::onReferenceUnavailable()
--- a/apps/openmw/mwgui/container.hpp
+++ b/apps/openmw/mwgui/container.hpp
@@ -34,6 +34,7 @@ namespace MWGui
         ContainerWindow(DragAndDrop* dragAndDrop);
 
         void openContainer(const MWWorld::Ptr& container, bool loot=false);
+        virtual void takeAll();
         virtual void close();
 
         virtual void resetReference();
--- a/apps/openmw/mwgui/windowmanagerimp.cpp
+++ b/apps/openmw/mwgui/windowmanagerimp.cpp
@@ -1143,6 +1143,16 @@ namespace MWGui
         return mToolTips->getFullHelp();
     }
 
+    void WindowManager::closeContainer() const
+    {
+        mContainerWindow->exit();
+    }
+
+    void WindowManager::takeAllItemsFromContainer() const
+    {
+        mContainerWindow->takeAll();
+    }
+
     void WindowManager::setWeaponVisibility(bool visible)
     {
         mHud->setWeapVisible (visible);
--- a/apps/openmw/mwgui/windowmanagerimp.hpp
+++ b/apps/openmw/mwgui/windowmanagerimp.hpp
@@ -226,6 +226,8 @@ namespace MWGui
     virtual bool toggleFogOfWar();
     virtual bool toggleFullHelp(); ///< show extra info in item tooltips (owner, script)
     virtual bool getFullHelp() const;
+    virtual void closeContainer() const;
+    virtual void takeAllItemsFromContainer() const;
 
     virtual void setActiveMap(int x, int y, bool interior);
     ///< set the indices of the map texture that should be used
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -162,12 +162,11 @@ namespace MWInput
 
     void InputManager::setPlayerControlsEnabled(bool enabled)
     {
-        int nPlayerChannels = 18;
-        int playerChannels[] = {A_Activate, A_AutoMove, A_AlwaysRun, A_ToggleWeapon,
+        int nPlayerChannels = 16;
+        int playerChannels[] = {A_AutoMove, A_AlwaysRun, A_Use, A_Journal,
                                 A_ToggleSpell, A_Rest, A_QuickKey1, A_QuickKey2,
                                 A_QuickKey3, A_QuickKey4, A_QuickKey5, A_QuickKey6,
-                                A_QuickKey7, A_QuickKey8, A_QuickKey9, A_QuickKey10,
-                                A_Use, A_Journal};
+                                A_QuickKey7, A_QuickKey8, A_QuickKey9, A_QuickKey10};
 
         for(int i = 0; i < nPlayerChannels; i++) {
             int pc = playerChannels[i];
@@ -234,8 +233,12 @@ namespace MWInput
                 break;
             case A_Activate:
                 resetIdleTime();
-                if (!MWBase::Environment::get().getWindowManager()->isGuiMode())
-                    activate();
+                if (MWBase::Environment::get().getWindowManager()->getMode() == MWGui::GM_Container)
+                {
+                    MWBase::Environment::get().getWindowManager()->closeContainer();
+                    break;
+                }
+                activate();
                 break;
             case A_Journal:
                 toggleJournal ();
@@ -247,6 +250,11 @@ namespace MWInput
                 toggleWalking ();
                 break;
             case A_ToggleWeapon:
+                if (MWBase::Environment::get().getWindowManager()->getMode() == MWGui::GM_Container)
+                {
+                    MWBase::Environment::get().getWindowManager()->takeAllItemsFromContainer();
+                    break;
+                }
                 toggleWeapon ();
                 break;
             case A_Rest:
@@ -1078,7 +1086,10 @@ namespace MWInput
 
     void InputManager::activate()
     {
-        if (mControlSwitch["playercontrols"])
+        if (!mControlSwitch["playercontrols"])
+            return;
+
+        if (!MWBase::Environment::get().getWindowManager()->isGuiMode())
             mPlayer->activate();
     }
 


scrawl:
I implemented the hotkey using MyGUI keys mainly because it's a royal pain to route everything manually through the InputManager to every relevant GUI widget, adding accessors makes the code a mess etc. But I see the current behavior isn't ideal from a user perspective.

setPlayerControlsEnabled(), as far as I remember is mainly a hack in order to prevent one keypress triggering two different events. This is, again, something that MyGUI does properly as only one widget can consume a given event.

A solution in my mind is:
* have a visual indicator for the key-focused button
* support left/right keys to change button focus
* improve widget behavior so that e.g. clicking on an item does not change key focus, and other related key focus issues (the former I think is easy by setting 'NeedKey' to false on the widget)

This way we support all 3 actions with no need to configure anything:
* ```esc``` closes container
* ```space``` disposes
* ```right``` ```space``` takes all

(2 and 3 can be reversed, if people prefer 'take all' as default, or the menu could even remember which button you had focused last)

This may require some MyGUI changes to implement properly, and I'm not sure what the 'focused highlight' should look like. The good news is that we can also leverage these changes to improve the game controller UI.

I understand there's a motive for the 'Activate' key to also activate buttons even if Activate isn't bound to Space, so we can implement that separately through InputManager (as a one-time exception). But I don't see any point in using Ready Weapon key for something else - weapons have nothing to do with this and I don't see how a user would even get the idea.

akortunov:
> But I don't see any point in using Ready Weapon key for something else - weapons have nothing to do with this and I don't see how a user would even get the idea.

Me too. But for now we have a hardcoded Enter button, so I asked for a better option.
Using arrows on the keyboard to navigate over buttons (instead of shortcuts) does not look as a good option too.

scrawl:
> Using arrows on the keyboard to navigate over buttons (instead of shortcuts) does not look as a good option too.

Why not? Is it such a problem to press two keys instead of one?

The key focus method scales better in applying to any window. Why are we only talking about the container window? What about closing the dialogue window, for example, or any other window where it would be convenient (one that frequently takes you out of gameplay)?

kcat:
> Why not? Is it such a problem to press two keys instead of one?

The arrow keys tend to be out of the way. With the default WASD+mouse controls, you'll either have to take your right hand off the mouse or move your left hand across the keyboard. The Ready Weapon and Activate keys are within easy reach of WASD, though.

scrawl:
Ok, good point. What about Tab / Shift-tab then? That's very close to WASD, and serves the same function as Right / Left in most GUI's (with the added bonus that it also works for shifting between text input boxes)

rhtucker:
I actually think @akortunov's solution as first described is best for now. I think the rest of these options are going to open a can of worms we don't want to deal with pre-1.0. I think the button focus approach is not really acceptable for either keyboard or gamepad.

For keyboard, anything short of allowing people to customize every single key for every window is going to just make people complain about something else. I don't think we need to make that happen now, but I think it will inevitably come about. If you have to cycle through a bunch of different button focuses, you might as well have just used the mouse. In the near term, the vast majority of the population is probably used to Vanilla with MCP, which makes spacebar the default to close out of all windows. We can meet in the middle for now by making it the activate key. I think regardless of the other functions in each window, they all share the need to close, which is also the most used function in most cases. ToggleWeapon is fine for Take All, and shift-ToggleWeapon could work as Dispose of Corpse.

For gamepad, I can't remember the last time I played a game that didn't just have a button assigned to different GUI functions. Every game that I can recall uses the far right button, B on XBox, to close out of menus, so that is what people will expect. Vanilla on XBox had a button assigned to Dispose of Corpse, Take All, and Close.  As with keyboard, cycling through GUI buttons is annoying when there is already a simple way to achieve it. Regardless, any serious attempt to improve the gamepad GUI should probably wait until post-1.0 as well.

scrawl:
> I actually think @akortunov's solution as first described is best for now. 

Maybe it's better than what we have now, but that doesn't mean it should be merged (i.e. if people get used to 'Ready weapon' having a certain function, and we change it later, they'll be annoyed and ask for it back).

> ToggleWeapon is fine for Take All, and shift-ToggleWeapon could work as Dispose of Corpse.

ToggleWeapon is obscure enough, who would ever think of pressing Shift-ToggleWeapon? I do not think there's much of a reason to add these hotkeys when most users would never discover them. Sure, you can write documentation, but most people won't read it. Or you can add to the keybinding window, but this menu is cluttered enough as it is and context-specific bindings don't really belong there, in my opinion.

> In the near term, the vast majority of the population is probably used to Vanilla with MCP

We can only speculate about that. I've heard of many who are playing Morrowind for the first time through OpenMW, and even people who have played it may or may not have used MCP.

Personally I do not think MCP doing things is any indication we should do the same. Some of its features are a little kludgy (i.e. instead of adding a new system, they'll repurpose an existing mechanic). This is, of course, understandable due to the way it has to reverse-engineer and hack the executable, which limits you in doing new things.

> For keyboard, anything short of allowing people to customize every single key for every window is going to just make people complain about something else.

That was kind of my point. It makes little sense to add specific hotkeys for specific windows on an individual basis, whether they are configurable or not. On the other hand, adding configurability for every button in every window seems way overkill.

> For gamepad, I can't remember the last time I played a game that didn't just have a button assigned to different GUI functions. Every game that I can recall uses the far right button, B on XBox, to close out of menus, so that is what people will expect. Vanilla on XBox had a button assigned to Dispose of Corpse, Take All, and Close. 

Certainly, yes, the gamepad UI should assign buttons. Preferably in a (semi-)automatic way so we don't have to go around doing it for the entire UI. For example, any button labelled 'sClose' or 'sCancel' would automatically assign to B. All of the improved gamepad UI should be built automatically from the GUI we already have, whenever possible. No one wants to maintain two different GUIs that are mostly the same.

> Regardless, any serious attempt to improve the gamepad GUI should probably wait until post-1.0 as well.

Why should it?

> I think the button focus approach is not really acceptable for either keyboard or gamepad.
> If you have to cycle through a bunch of different button focuses, you might as well have just used the mouse. 

I can't think of much of a problem with it. 'Esc' and 'Space' already cover 2/3rd of the cases, and the last case 'Tab' 'Space' is relatively accessible too. Using the mouse may require you to move it from a totally different region of the screen, which takes longer.

rhtucker:
> if people get used to 'Ready weapon' having a certain function, and we change it later, they'll be annoyed and ask for it back

Fair enough, but that would be true of any key binding we provide.

>  I do not think there's much of a reason to add these hotkeys when most users would never discover them. Sure, you can write documentation, but most people won't read it. Or you can add to the keybinding window, but this menu is cluttered enough as it is and context-specific bindings don't really belong there, in my opinion.

This would also be true of any solution involving key bindings, but it doesn't make shifting button focus any less of a clunky solution itself. The only people who are complaining about not having a Take All hotkey are people who would probably dig around in key bindings anyway.

> That was kind of my point. It makes little sense to add specific hotkeys for specific windows on an individual basis, whether they are configurable or not. On the other hand, doing this for every button in every window seems way overkill.

I agree with that, but at the very least we should have an in-game key map and an in-menu key map. I can't remember which game it was that I played recently, but it had different tabs in the key mapping interface that were for each set of GUIs they had and I really liked that.

Part of what I'm thinking of is that I use space to get out of dialogue, containers, probably other windows, but for dialogue boxes with Ok/Cancel, it defaults to OK. This is different from how we're used to using controllers where cancel and close are the same button. This gives me a slightly different idea. 

Would having a "reverse keybinding" menu be a good solution? What I mean is that instead of selecting the key for a function, what if we had a "default" key that we could select the default function for each window? Correct me if I'm wrong, but people who use Take All probably always use Take All, so that is what they want a key for. But I close out of containers more often because I'll search every crate quickly, but most of them have random useless stuff. I also talk to every NPC I don't know to see if they have anything interesting to say, but if all they have is Solstheim and rumors, etc, I immediately close out. So for both of those my default behavior is close. I hate using escape for these "quick checks" because I'm trying to go through things quickly, so moving my finger forces my hand away from WASD resting position. Anything besides default there is really no reason to have a hotkey for. I'm sure you'll disagree because you're a programmer, but if our user base had more programmers that liked keyboard only interfaces, we'd have more developers.

>Maybe it's better than what we have now, but that doesn't mean it should be merged

My point is that I think a universal interface will require a lot of work, which it seems like we had been saving for post-1.0. I don't necessarily think we should be adhering strictly to the vanilla interface. This would solve some interim problems until we get to that point. If this isn't true, then we should probably announce an overall plan on the forums for input rather than speculating about what users want here.

scrawl:
> Fair enough, but that would be true of any key binding we provide.

Correct. The point is to do it right the first time, so there won't be another change down the road. Using a totally unrelated keybinding does not seem like the best way to ensure that.

> This would also be true of any solution involving key bindings, but it doesn't make shifting button focus any less of a clunky solution itself. 

My reasoning is the function of tab key is well known by users, performing the same function in pretty much any GUI application. It's also the typical way to use your computer if you misplaced your mouse. Or am I wrong and only 'power users' use tab? I couldn't imagine logging into a website any other way than ```name<tab>pass<enter>```

> The only people who are complaining about not having a Take All hotkey are people who would probably dig around in key bindings anyway.

Right, but wouldn't it be even better if the people who didn't complain (and don't dig around in keybinding menus) can also benefit from the new feature? 

Personally, I like having as few options as reasonably possible. Anything I don't use is wasting my attention just by having to read it or trying to understand what it does.

> If this isn't true, then we should probably announce an overall plan on the forums for input rather than speculating about what users want here.

Valid point, though input isn't necessarily as good as testing (i.e. I think I'd prefer X, but after trying both approaches I like Y better). 

> My point is that I think a universal interface will require a lot of work

We can do it in steps, one improvement at a time.

> What I mean is that instead of selecting the key for a function, what if we had a "default" key that we could select the default function for each window? Correct me if I'm wrong, but people who use Take All probably always use Take All, so that is what they want a key for. But I close out of containers more often because I'll search every crate quickly, but most of them have random useless stuff.

That's essentially what I meant by the 'menu could even remember which button you had focused last', except the default would be changed simply by using that button instead of going to the options menu. So if I come across a room of useless crates, and after that a room of treasures, my default choice naturally changes without interrupting gameplay, and without having to know in advance if the loot here is 'good' or 'bad'. We just need to visually indicate the highlighted button or else it'd be too confusing what is happening.

> I also talk to every NPC I don't know to see if they have anything interesting to say, but if all they have is Solstheim and rumors, etc, I immediately close out. 

Hm, we don't have spacebar-closing of the dialogue window yet? That sounds like an easy addition worth making while we're still busy talking about the 10000 possibilities for container closing and universal interfaces.

rhtucker:
Hmm, well I've been thinking about it and I'm feeling a bit more comfortable with your proposal. We agree that buttons can be assigned to the controller, which was a large part of my hang up with this. I think at this point, I'm concerned with intuitiveness and interface elegance.

I agree that tab really makes the most sense (although you would be shocked at how few people use it for even logging in, but I think our users are a little more savvy). There aren't enough options to warrant needing a bidirectional focus shift and tab is used that way already. I just don't think anyone would guess that it's even possible to do any of this without directly telling them, because as you say, no one is going to read documentation on controls. Perhaps the interface element will bring enough attention to it, which isn't necessarily a good thing.

Which brings me to how it will look. A large part of what I have a problem with is I'm imagining something like your browser tab where it puts little boxes around things or highlights them. I think in game this would look really tacky and out of place. Maybe we could have an "o" sized dot next to the words inside the button. This would be a clear marker, but be less intrusive. Perhaps adding some scrollwork to the button border. It really doesn't matter now I suppose as we can change the element later, the functionality to get rid of those bugs and improve controls is most important.

A few additional thoughts:
>Valid point, though input isn't necessarily as good as testing

I agree, iteration is often more productive than extensive planning, but I'm sure you'll agree we should still have a pretty good idea in mind. That being said, perhaps this means the urgency of a universal interface is higher than we had planned. If anything does change significantly, it would be better to do it during alpha stages now while people are "expecting" large changes. Post 1.0 would raise significantly more complaints.

>Hm, we don't have spacebar-closing of the dialogue window yet? That sounds like an easy addition worth making while we're still busy talking about the 10000 possibilities for container closing and universal interfaces.

I just checked to make sure and no, or it's not working for me. This would be greatly appreciated! I would also appreciate if the spacebar could also select OK on prompts.

akortunov:
> Hm, we don't have spacebar-closing of the dialogue window yet?

Yes, we don't. Judging by code, we have this feature only for containers and interactive messageboxes (and it seems the latter is broken).
You can close a dialogue window by pressing ESC button instead.

scrawl:
> There aren't enough options to warrant needing a bidirectional focus shift and tab is used that way already. I just don't think anyone would guess that it's even possible to do any of this without directly telling them, because as you say, no one is going to read documentation on controls.

With enough hints hopefully people will figure it out. There are already several (obvious) ways in which OpenMW's GUI behaves more similar to a traditional GUI than the original engine does. For example, we can navigate through list items with Up/Down keys and accept with Enter. Left/Right keys should not be a big leap from that, and once that is figured out the user might think to myself 'if only there was a more conveniently located key that may have the same function' and press 'Tab' by semi-accident. Hopefully, anyway :)

> Which brings me to how it will look. A large part of what I have a problem with is I'm imagining something like your browser tab where it puts little boxes around things or highlights them. I think in game this would look really tacky and out of place. Maybe we could have an "o" sized dot next to the words inside the button. This would be a clear marker, but be less intrusive. Perhaps adding some scrollwork to the button border.

I like the idea of a thicker button border. Alternatively, how about a semi-transparent (say 10%) white rectangle that covers the entire button? To make it more interesting, this rectangle could be textured with something.

rhtucker:
openmw-default-button-proof
I was thinking something along these lines. Making the border thicker means the button takes up more room and is out of line with the rest of the buttons. I initially was thinking pattern as well, but I didn't want to hinder legibility. Would this be relatively simple to implement akortunov? Ideally it's something possible with a GUI texture already present, but if you need me to make something for it, it's pretty easy. Just let me know exactly what you need.

>Yes, we don't. Judging by code, we have this feature only for containers and interactive messageboxes (and it seems the latter is broken).
You can close a dialogue window by pressing ESC button instead.

Would you like me to make a bug report and feature request respectively?
As a side note, the message box buttons show almost no horizontal padding compared to every other button that I was able to find when I was poking around. Is this intended behavior? I can make a bug report for this as well. I don't know what vanilla was like, but it seems it would be best if the GUI was visually consistent.

scrawl:
Hm... the rounding looks a little out of place, with every other aspect of the GUI being rectangular. What about just using the mouse-over or mouse-down state of the button (different font color)?

By the way, we have another type of button - image buttons, used in the main menu and books/journals.

> As a side note, the message box buttons show almost no horizontal padding compared to every other button that I was able to find when I was poking around. Is this intended behavior? I can make a bug report for this as well. I don't know what vanilla was like, but it seems it would be best if the GUI was visually consistent.

That's because the messagebox sizes its button manually (code was written before auto-sized buttons became a thing). Trivial fix, might as well make a PR instead of a report.

https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwgui/messagebox.cpp#L209

needs to match

https://github.com/OpenMW/openmw/blob/master/components/widgets/box.cpp#L84

Note, the first one is padding and the latter is 2*padding, so, correct values would be 12 and 24 instead of 5 and 24.

akortunov:
> Would this be relatively simple to implement akortunov?

If you mean things on your screenshot:
1. This solution may require adding an additional texture and can break some GUI mods
2. This solution may require ability to change button skin on the fly (I am not sure if MyGUI have it).

We can highlight a button label instead (as "over" or "pressed").

akortunov:
> By the way, we have another type of button - image buttons

IIRC, image buttons usually have three images: idle, over, and pressed ones.

scrawl:
The 'cleanest' way would be widget states, though IIRC they're a bit limited in functionality (just changing font color). Colors are defined [here](https://github.com/OpenMW/openmw/blob/master/files/mygui/openmw_button.skin.xml#L85) and then the state is changed by calling ```_setWidgetState("disabled");```. Note the state will be overwritten if we mouse-over this widget.

Here's how a proof of concept could work (evidently, it's going to be a bit hack-ish pending additions to upstream MyGUI):

* For every MyGUI::Window, we keep track of the current key-focused widget (this is in contrast to MyGUI::InputManager, which only keeps track of <i>one</i> key-focused widget globally)
* When the top Window changes, the key-focused widget of that window (as kept track of above) is brought into the MyGUI::InputManager's key focus.
* The default key-focus widget (for when a window is opened the first time) could be based on the widget order, or on their labels (e.g. preferring 'sOk'/'sYes'/''sNo'/'sClose'/'sCancel'/'sOk' in that order).
* Keys Space/Enter, Left/Right and Tab/Shift-Tab are intercepted on a manager level and perform the discussed functions.

A few notes:
* When opening the container GUI, we'll need to ensure the container window is the top window (which makes sense anyway, we prefer seeing what's in the container and not the player inventory).
* [ImageButton](https://github.com/OpenMW/openmw/tree/master/components/widgets/imagebutton.cpp) will need additional treatment because it currently only responds to mouse events and does not use the widget state system.
* Widgets could be destroyed at any time, so we need to make sure we're not going to hold deleted widget pointers in our manager. This can be addressed with a MyGUI::WidgetManager::registerUnlinker so that we get a callback before a widget is deleted, and can clean up all our pointers to it.
* Usually we'll want to go with the 'Windows' [layer](https://github.com/OpenMW/openmw/blob/master/files/mygui/openmw_layers.xml) to get the top window from. But the Console and Journal are on a separate layer, respectively. This is not an important issue and can be fixed later, or maybe is not an issue at at all.

Related functions:
* MyGUI::LayerManager::upLayerItem(window) brings a window to the top.
* MyGUI::ILayer::getLayerNodeAt (size_t _index) with index = 0 tells us the window at the top.
* MyGUI::InputManager::get/setKeyFocusWidget (note: setKeyFocusWidget should not be used, always use the MWGui::WindowManager::setKeyFocusWidget wrapper instead which updates SDL text input state).

rhtucker:
>Hm... the rounding looks a little out of place, with every other aspect of the GUI being rectangular. What about just using the mouse-over or mouse-down state of the button (different font color)?

Yes, I see what you mean. I think if we were able to pull in an existing texture, that would hopefully make it easier to implement as well as help visually tie it in. I only liked it because it requires no size change at all to the button and sticks out enough to draw attention to the fact that it's different. I'm not opposed to using a different state, but I think it should be distinct from mouse-over and mouse-down. It will look weird to have the default button and the currently moused-over button to have the same state.

The image buttons are more of a problem. I take it they have prerendered states? This is kind of going down a rabbit hole, but perhaps it would make sense to eliminate image buttons entirely if that's the case. It seems like prerendered buttons make a lot of things difficult, localization and extensibility of the interface for instance.

What would make the most sense to me is having the default button with some kind of border so the font color of mouse-over and mouse-down can still be seen as distinct. Or perhaps just bolding the defaulted button?

>Trivial fix, might as well make a PR instead of a report.

Unfortunately, I don't have a dev environment set up to test changes and I'm more likely to screw something up than fix anything so I'll leave it to you guys.

scrawl:
> This is kind of going down a rabbit hole, but perhaps it would make sense to eliminate image buttons entirely if that's the case. It seems like prerendered buttons make a lot of things difficult, localization and extensibility of the interface for instance.

And replace them with what? We'll want something that closely replicates the look of the original game.

I had a closer look at the MyGUI side of things. I had hoped we could work around any limitations, but this is starting to look harder and harder. Might be a better idea to work with upstream, push for a MyGUI 3.2.3 release and use it as soon as we can. In no particular order, here's what I found:

First of all, MyGUI master happens to have a new feature that looks useful to us: https://github.com/mygui/mygui/commit/632d0074 . Buttons will now use the 'highlighted' (=mouseover) visual state if they have key focus. As the commit states, by default this can only be achieved manually using 'setKeyFocusWidget' in the code. But one can also set the Button skin's 'NeedKey' property to true to make buttons receive key focus when clicked on. This is basically what I had in mind.

Then I attempted to find a way to make our ImageButtons use widget states. There's a 'ModeImage' value in MyGUI::Button which looks useful on first glance. Unfortunately it only works with images that are in the same texture, and the image switching on state-switch is done by giving new coordinates inside that texture. Like most of MyGUI, this feature was designed with the assumption that GUI textures are atlased for efficiency, which is not something Morrowind does, at all. Sadly, I couldn't spot a clean way to use widget states in our custom class either (_setWidgetState is private and non-virtual, Button's _setState is, too, and an equivalent _getWidgetState doesn't even exist). Looks like we'll have to duplicate some code from Button.

My last problem is with how MyGUI sets key focus. Currently, if I click on an empty space or on a widget that doesn't accept key focus, the key focus changes to that widget's first parent that accepts key focus, or if there's no such widget, the key focus resets to 'none'. This is not how other GUI's work and seems problematic. I'd like to change it to 'widget's first parent that accepts key focus, and if there's none then retain the previous focus', which would be a simple fix to MyGUI in itself. There's another related issue, though: MyGUI::Window is set to accept key focus (accompanied by a 'fixme' comment), so any click in a window would still trash our key focus. I suspect this has to be the case due to MyGUI's 'root key focus' handling. This root key focus appears to be required, among other things, to make Menu Controls disappear if I click anywhere outside that menu, and currently, root key focus can only get set if the clicked-on widget or any of its parents accept key focus. A fix might be to raise root key focus events even for widgets not accepting key focus ([related code](https://github.com/MyGUI/mygui/blob/master/MyGUIEngine/src/MyGUI_InputManager.cpp#L401)). I'm really not sure about this, though.

rhtucker:
> And replace them with what? We'll want something that closely replicates the look of the original game.

Obviously we want something closely matching the originals. All I can recall without looking is that the image buttons have a glow. I don't know how difficult that would be to add (I imagine rather difficult if there is no similar functionality, so it's probably out of the question), but there may be some embossed effect too, which would be even more difficult. Perhaps we should just implement this for non image buttons at the moment? That only leaves out the journal for in-game widgets correct?

It sounds like someone would have to custom build the code for the image buttons, which is unfortunate. If we had states for image buttons, the main menus could simply be tabbed through and selected gamepad style, which I think would be sufficient. I don't see why we would want to include the "default" feature in the main menus. You have to open the in-game menu via ESC, so your finger is already there if you want to resume quickly. I think the journal image buttons might need to be replaced though. IIRC they don't have all the fancy visual effects that the main menu has, just a glow. I don't think anyone would seriously miss it if we just had them change color so we could more easily add the default feature.

And thanks for the fix akortunov!

scrawl:
What is your objective with replacing the image buttons? I'm not sure what that has to do with this feature. They have prerendered normal, highlighted and pressed states, of which we can use the highlighted state to indicate they're in keyboard focus.

I decided to hack up MyGUI for my purposes for now, and worry later about how to get this upstream cleanly. This temporarily resolves the issues from my last post, and I massaged the ImageButtons into working (without changes to MyGUI) by adding overrides for key focus events. I found a new issue though: the way I intended to get the top window from the MyGUI::LayerManager isn't going to work, at all, and I'm not too keen on hacking MyGUI to do this (doesn't look like it was meant to). Instead, I decided to try using our WindowManager's 'GuiMode' stack to store key focus per GUI mode, instead of per window. We'll see how that goes. From my limited testing so far, I really like the new feature, though not necessarily for container windows as the main usecase. It's nice to be able to jump into my last game with 'Enter' 'Enter' from the main menu, and even better to skip that obnoxious 'Your plugins have changed' warning with another 'Enter'.

akortunov:
Can this PR be closed now?

PR #1444, opened 2017-09-10
Spell Ctrl instead of Ctl in OpenMW-CS shortcuts

Capostrophic:
I'm not sure why did Aesylwinn use Ctl, since Ctrl is much more common contraction for Control key. I doubt that "Ctl+whatever" may confuse users, but IMO that looks odd.


zinnschlag:
Good catch!

PR #1446, opened 2017-09-11
Do not allow to equip a replacement, when take all items from a dead corpse

akortunov:
Fixes [bug #4095](https://bugs.openmw.org/issues/4095).

Equipping in this case is useless - you will take all items anyway.
Also this PR makes exceptions in InventoryItemModel::removeItem() a bit more informative.

### DIFF
--- a/apps/openmw/mwgui/container.cpp
+++ b/apps/openmw/mwgui/container.cpp
@@ -11,6 +11,7 @@
 #include "../mwmechanics/actorutil.hpp"
 
 #include "../mwworld/class.hpp"
+#include "../mwworld/inventorystore.hpp"
 
 #include "../mwmechanics/pickpocket.hpp"
 #include "../mwmechanics/creaturestats.hpp"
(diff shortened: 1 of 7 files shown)


scrawl:
Maybe we can make the 'equip a replacement' thing a little more explicit? Where is that actually needed? The only case I can think of is the companion window.

akortunov:
> Where is that actually needed? 

IIRC, in works in vanilla game every time when you take an item from corpse (or put an item to the corpse).
However, looks like it does not work when you take ALL items from corpse.

But I am not sure if some mods using this "feature" (maybe mannequin ones?).

scrawl:
Ah. I forgot you can dress up corpses. (Was on my mind actually, but I discarded it as 'nah that can't be right'. Yeah, this game is weird).

So we have:
- dressing up corpses
- taking items from companion

which are actually sharing the same code, anyway. So I wonder if the re-equipping should be done in the GUI code. That way the 'equipReplacement' flag can be removed from ContainerStore, and we don't need to manually set it to false when we don't want this totally-intrusive behavior.

akortunov:
> the re-equipping should be done in the GUI code.

Basically, done. I'll try to refactor this code a bit.

akortunov:
I can move autoequipping to the MWWorld::InventoryStore::remove() function.
Should I do it?

akortunov:
Ok, I uploaded a new version.
I think, this PR is ready for review.

apps/openmw/mwgui/container.cpp
```diff
@@ -229,11 +229,14 @@ namespace MWGui
                 const ItemStack& item = mModel->getItem(i);
 
                 if (!onTakeItem(item, item.mCount))
-                    break;
-
-                mModel->moveItem(item, item.mCount, playerModel);
+                {
+                    MWBase::Environment::get().getWindowManager()->removeGuiMode(GM_Container);
+                    return;
+                }
```
scrawl: Looks like there is a change in behavior here. If you 'take all' with the pickpocket window, successfully roll the first few items but fail on a later one, no items are added to your inventory. Is this correct?

akortunov: Sorry, my fault. How can I workaround this bug?

akortunov: As a simple solution, I can just add equipReplacement flag to ItemModel::moveItem() and ItemModel::removeItem().
Then I will rollback changes in container.cpp.

scrawl: I'd prefer not complicating the item model API for something that doesn't apply to all item models.

How about this?
<pre>
if container is NPC:
   for every equipped item in model:
      unequip item
   update model
take all items
</pre>

It's still hack-ish, but at least the impact is limited to one file.

akortunov: Hmm... How is this code supposed to help us? And I did not get where we should place it.
We have no problems with **unequipping**, we have a problem with **autoequipping**.
When we use for() loop to remove all items one by one, we should not touch items stacks (as autoEquip does).
As a simpliest, but a dumb solution, we can use a while() loop instead of for(), and update model after every removed stack.
But in this case we can get an endless loop, if something will go wrong, so we will have to add some failsafe break.

scrawl: > How is this code supposed to help us? 

By un-equipping everything first, the problematic code path in InventoryStore::remove() that re-equips an alternative item when an equipped item is removed should no longer trigger (because the removed item wasn't equipped). Makes sense?

> while() loop instead of for(), and update model after every removed stack.

I've thought about this. The quadratic complexity could be a problem, though (i.e. for every item, we need to update the model, looking at every item again). How fast is this for a container with say, 1000 items? Good point with the failsafe, too.

akortunov: > Makes sense?

Yes. It is implemented now.

PR #1447, opened 2017-09-11
Clear player fields upon reload (bug #2639)

akortunov:
Should fix [bug #2639](https://bugs.openmw.org/issues/2639), if it still exists.

What's wrong here:
1. We keep mAttackingOrSpell flag between reloads.
2. With mAttackingOrSpell == true an engine treats a player as aiming.
3. This can lead to unintended behaviour, such as animation bugs, accidental spellcasting, etc.

So I suggest just ensure that mAttackingOrSpell will be reset after reload.

### DIFF
--- a/apps/openmw/mwworld/player.cpp
+++ b/apps/openmw/mwworld/player.cpp
@@ -276,6 +276,29 @@ namespace MWWorld
         mAutoMove = false;
         mForwardBackward = 0;
         mTeleported = false;
+        mAttackingOrSpell = false;
+        mCurrentCrimeId = -1;
+        mPaidCrimeId = -1;
+        mLastKnownExteriorPosition = osg::Vec3f(0,0,0);
+
+        for (int i=0; i<ESM::Skill::Length; ++i)
+        {
+            mSaveSkills[i].setBase(0);
+            mSaveSkills[i].setModifier(0);
+        }
+
+        for (int i=0; i<ESM::Attribute::Length; ++i)
+        {
+            mSaveAttributes[i].setBase(0);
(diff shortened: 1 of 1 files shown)


apps/openmw/mwworld/player.cpp
```diff
@@ -276,6 +276,9 @@ namespace MWWorld
         mAutoMove = false;
         mForwardBackward = 0;
         mTeleported = false;
+
+        // Reset mAttackingOrSpell flag to avoid animation glitches upon reload
+        MWBase::Environment::get().getWorld()->getPlayer().setAttackingOrSpell(false);
```
scrawl: Why go through World? Why not just:

<pre>
mAttackingOrSpell = false;
</pre>

Also we should reset mCurrentCrimeId, mPaidCrimeId, mLastKnownExteriorPosition

We can also reset mSaveSkills, mSaveAttributes and mMarkedPosition, just for good measure (even if there isn't a bug resulting from not clearing them).

Alternatively, we remove the clear() method and just create a new player object instead. That's less code, but inconsistent with the way clearing works elsewhere.

akortunov:
Reworked.

> We can also reset mSaveSkills, mSaveAttributes and mMarkedPosition

How can I reset them? Iterate over all arrays ind assign 0?
And anyway, we reset these variables when read data from savegame.

scrawl:
> How can I reset them? Iterate over all arrays and assign 0?

Yes.

> And anyway, we reset these variables when read data from savegame.

True, but not when starting a new game.

akortunov:
> Yes.

Done.

PR #1448, opened 2017-09-12
Manual modding - extend mod installation instructions

loriel2:
Add section on .bsa files
Extend plugins to include .esm files
Fix typo propper->proper.


rhtucker:
I'm glad we have someone working on our CS guide, thanks @loriel2! I think I want to massage the esm and bsa additions into the manual slightly differently, but go ahead with the merge anyway as I don't know when or how I want to do it.

It's entirely probable that I just am incompetent at understanding GitHub, but is there a reason why it says only two files have changed? It looks like there have been a great deal more than that.

loriel2:
I am a novice at using github, and was making two sets of changes to the same repository - CS Manual (PR 1449) and Modding Reference (this one, 1448). I then split off two separate branches, and in each branch  reversed the unwanted changes to produce two independent sets of changes. Ugly, but it finished up with what I wanted. Next time I will create the branch(es) first which should reduce the number of confusing changes in the PR.

PR #1449, opened 2017-09-12
CS Manual - extend tour ring tutorial

loriel2:
Add a section for giving the ring to an npc - both as shopkeeper and for simple interaction.

Images are reasonably small files, but style isn't the same as previous images in the tour, so please consider them as placeholders until I can replicate the style

### DIFF
--- a/docs/source/reference/modding/mod-install.rst
+++ b/docs/source/reference/modding/mod-install.rst
@@ -19,7 +19,6 @@ Install
 #.	Find or search for ``data=``. This is located very near the bottom of the file. If you are using Morrowind, this first entry should already point to your Morrowind data directory, ``Data Files``; otherwise it will point to your game file, ``.omwgame``.
 #.	Create a new line underneath and type: ``data="path/to/your/data folder"`` Remember, the *data folder* is where your mod's plugin files are. The double quotes around this path name are *required*.
 
-
 .. note::
 	Some text editors, such as TextEdit on Mac, will autocorrect your double quotes to typographical "curly" quotes instead of leaving them as the propper neutral vertical quotes ``""``.
 


psi29a:
How long would it take to replicate the style? Is it really that important?

loriel2:
I don't think it's essential, but it would look more professional if the style is consistent.

Two issues:
Colour of top bar/border - red in mine, grey in others
"Shadow" effect on others, missing from mine

Don't yet know how to achieve these, so don't know how long it will take to do.

rhtucker:
I agree that the style should be consistent. However, I don't necessarily like styling them at all because it means in order to be consistent, everyone contributing has to do the exact same thing. @psi29a, would we be able to add a shadow and outline via the Sphinx templates? Ideally that would all be taken care of via CSS once the document is generated. Granted this would then require us to re-screenshot the pics that are already there.

If that is going to be more trouble than it's worth, then we should decide on exactly what we want the style to be and create a very brief step by step guide on how to achieve it. Slightly more annoying would be having a single person in charge of doing the styling and just have them go through every image we use.

The OS styling for windows and window borders may end up being too difficult to realistically make everything uniform anyway so maybe all of this is moot.

psi29a:
RTD has their own style and I'm not sure how easy that is to modify. I'm also not the person to ask about modifying CSS. ;)

My only question was trying to limit the amount of binaries introduced into the repo. If these icons are placeholders, then they will be dead weight that gets carried around and never used.

loriel2:
Being a github novice, I hadn't appreciated that - I assumed the placeholders would be deleted/overwritten, but presumably they have to be retained to provide a full back history.
I know there are file size questions on the other PR in this area (#1363), leading to a question of "Where should the images be kept?". Would it be sensible to commit both of these with no images, pending resolution of the various questions?

scrawl:
Yes, better to remove the images for now - both for file size and so we can possibly re-shoot them all in a consistent style.

loriel2:
Closing this PR - now replaced by  #1451

PR #1453, opened 2017-09-14
ESS-Importer: Convert ballistic projectiles

MiroslavR:
Partially implements: https://bugs.openmw.org/issues/2320

This PR implements conversion of ballistic projectiles. Magic projectiles need further investigation - at first sight it looks like we might be missing context for magic effects (though I don't understand why OpenMW must serialize those effects in the first place - aren't they redundant?)

Due to the need for indexedRefId -> actorId conversion, I also had to implement static assignment of actor IDs. Previously, all actor instance records had them set to -1 (which means assign on demand at runtime). I'm surprised we haven't needed to reference concrete actors within converted savefiles until now.

I also ran into a bug where actors wouldn't fight me back after attacking them (the code for calming witnesses always triggered even if I didn't pay a fine). Turned out to be an uninitialized variable problem affecting the converted player record.

### DIFF
--- a/apps/essimporter/CMakeLists.txt
+++ b/apps/essimporter/CMakeLists.txt
@@ -16,6 +16,7 @@ set(ESSIMPORTER_FILES
     importjour.cpp
     importscri.cpp
     importscpt.cpp
+    importproj.cpp
     importercontext.cpp
     converter.cpp
     convertacdt.cpp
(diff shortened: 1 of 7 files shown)


scrawl:
> Magic projectiles need further investigation - at first sight it looks like we might be missing context for magic effects (though I don't understand why OpenMW must serialize those effects in the first place - aren't they redundant?)

I think they technically are redundant, just need a bit of code refactoring so we can determine the effects on impact from the enchantment/spell ID.

PR #1458, opened 2017-09-17
Use default 0 precision in the float formatting (bug #4096)

akortunov:
Fixes [bug #4096](https://bugs.openmw.org/issues/4096).

Looks like Morrowind uses default 0 precision in the "%0.f" format.

### DIFF
--- a/components/misc/messageformatparser.cpp
+++ b/components/misc/messageformatparser.cpp
@@ -34,21 +34,19 @@ namespace Misc
 
                         if (i < m.size())
                         {
-                            int precision = 0;
-                            bool precisionSet = false;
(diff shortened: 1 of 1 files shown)


kcat:
Perhaps this should only be the case for empty size. That is, ```%f``` should still have the default precision, but ```%.f``` has 0 precision. Basically the same as it was, but ```precisionSet``` is set to true once the ```.``` is found even if there's no numeric characters. Does vanilla use 0 precision with a plain ```%f```?

akortunov:
> Does vanilla use 0 precision with a plain %f?

No. 
For now I use 0 once "." found, otherwise I use -1.

PR #1460, opened 2017-09-17
Check death counts are for valid actors before loading them

AnyOldName3:
Should resolve the issue discussed here: https://forum.openmw.org/viewtopic.php?f=3&t=4637


scrawl:
Does this actually work? I think you still need to read the ```reader.getHNT(count, "COUN");``` or else it'll mess up the reader position for the next iteration.

AnyOldName3:
I never said it wouldn't create a zillion regressions. Sorry about that.

PR #1463, opened 2017-09-18
Allow to use GetDetected without a reference

akortunov:
Fixes [bug #3110](https://bugs.openmw.org/issues/3110).

Works as scrawl wrote:

> Testing with the console in vanilla MW, 'getDetected, player' without an implicit reference returns true if any NPC is detecting the player. The same thing happens when invoking the function with a non-NPC implicit reference (e.g. a creature).

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -1086,6 +1086,39 @@ namespace MWMechanics
         }
     }
 
+    bool Actors::isActorDetected(const MWWorld::Ptr& actor, const MWWorld::Ptr& observer)
+    {
+        if (!actor.getClass().isActor())
+            return false;
+
+        // If an observer is NPC, check if he detected an actor
+        if (!observer.isEmpty() && observer.getClass().isNpc())
+        {
+            return
+                MWBase::Environment::get().getWorld()->getLOS(observer, actor) &&
+                MWBase::Environment::get().getMechanicsManager()->awarenessCheck(actor, observer);
+        }
+
+        // Otherwise check if any actor in AI processing range sees the target actor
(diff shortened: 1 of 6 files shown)


apps/openmw/mwmechanics/actors.cpp
```diff
@@ -1086,6 +1086,39 @@ namespace MWMechanics
...
+            return
+                MWBase::Environment::get().getWorld()->getLOS(observer, actor) &&
+                MWBase::Environment::get().getMechanicsManager()->awarenessCheck(actor, observer);
+        }
+
+        // Otherwise check if any actor in AI processing range sees the target actor
+        std::vector<MWWorld::Ptr> actors;
+        osg::Vec3f position (actor.getRefData().getPosition().asVec3());
+        getObjectsInRange(position, aiProcessingDistance, actors);
+        for(std::vector<MWWorld::Ptr>::iterator it = actors.begin(); it != actors.end(); ++it)
```
kcat: Just a stylistic comment. It would probably be better to use a range-for loop:
```for(MWWorld::Ptr &other : actors)```
Functionally equivalent, but less verbose and less error-prone.

akortunov: I have never seen this style in OpenMW before.
If we will use it, we should also refactor loops in other files.

kcat: It's a feature added by C++11, so it's only been available to use in OpenMW recently. It also only works when you want to go from begin to end (can't modify the container itself while iterating, i.e. remove or add entries, can't reverse iterate, and you can't start some point after begin). I also don't think it's generally accepted to change code for the sake of changing it, but if you're working on some code anyway then it may as well be fixed up.

PR #1465, opened 2017-09-20
Bug 4077 - Enchanted items are not recharged if they are not in the player's inventory

Langerz82:
This should recharge items once a player has used it and has come back to an active cell. 

Note - Hope I have done the pull request right this is my first pull request. The code was applied from a manual patch so hopefully all changes are in there and correct.


Langerz82:
There are extra code i need to remove.

PR #1468, opened 2017-09-20
Improve item sorting in inventory and containers

akortunov:
Should fix [bug #4112](https://bugs.openmw.org/issues/4112), and, maybe, implements [feature #4080](https://bugs.openmw.org/issues/4080).
For now we sort items in inventory only by type and name.

How sorting works with this PR:
1. By type
2. By name
3. By enchantment 
(items with constant effect first, then items by charge percent, then non-enchanted items)
4. By condition
5. By remaining usage time (e.g., lights)
6. By value (cost)
7. By weight
8. By ID

Notes: 
1. As you can see, self-made potion with default names will be sorted by 6-8.
2. Soulgem sorting by soul amount makes no sense since soul affect soul value.
3. Enchantment sorting is a quite complicated for now.

Any ideas how to improve sorting, or refactor this code?

### DIFF
--- a/apps/openmw/mwgui/sortfilteritemmodel.cpp
+++ b/apps/openmw/mwgui/sortfilteritemmodel.cpp
@@ -96,8 +96,6 @@ namespace
                         leftChargePercent = (left.mBase.getCellRef().getEnchantmentCharge() == -1) ? 100
                             : static_cast<int>(left.mBase.getCellRef().getEnchantmentCharge() / static_cast<float>(ench->mData.mCharge) * 100);
                 }
-                else
-                    std::cerr << "Warning: Can't find enchantment '" << leftName << "' on item " << left.mBase.getCellRef().getRefId() << std::endl;
             }
 
             if (!rightName.empty())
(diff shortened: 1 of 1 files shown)


apps/openmw/mwgui/sortfilteritemmodel.cpp
```diff
@@ -59,15 +59,95 @@ namespace
...
+                if (ench)
+                {
+                    if (ench->mData.mType == ESM::Enchantment::ConstantEffect)
+                        leftChargePercent = 101;
+                    else
+                        leftChargePercent = (left.mBase.getCellRef().getEnchantmentCharge() == -1) ? 100
+                            : static_cast<int>(left.mBase.getCellRef().getEnchantmentCharge() / static_cast<float>(ench->mData.mCharge) * 100);
+                }
+                else
+                    std::cerr << "Warning: Can't find enchantment '" << leftName << "' on item " << left.mBase.getCellRef().getRefId() << std::endl;
```
scrawl: I'd get rid of this warning. It'll be printed every time the inventory is sorted, multiple times even. We already have similar warnings elsewhere that should be sufficient.

Ideally, any invalid references would be checked when loading the game, then replaced with null references, so the actual game logic doesn't have to deal with invalid references, and warnings are only printed once. That's how vanilla behaves, AFAIK. We'll need quite a bit of refactoring to get there, though.

PR #1469, opened 2017-09-20
water shader improvements (partially fixes #3365)

drummyfish:
bugfix

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -279,11 +279,14 @@ namespace MWRender
         mViewDistance = Settings::Manager::getFloat("viewing distance", "Camera");
         mFieldOfView = Settings::Manager::getFloat("field of view", "Camera");
         mFirstPersonFieldOfView = Settings::Manager::getFloat("first person field of view", "Camera");
-        updateProjectionMatrix();
         mStateUpdater->setFogEnd(mViewDistance);
 
         mRootNode->getOrCreateStateSet()->addUniform(new osg::Uniform("near", mNearClip));
         mRootNode->getOrCreateStateSet()->addUniform(new osg::Uniform("far", mViewDistance));
+
+        mUniformNear = mRootNode->getOrCreateStateSet()->getUniform("near");
+        mUniformFar = mRootNode->getOrCreateStateSet()->getUniform("far");
+        updateProjectionMatrix();
     }
 
     RenderingManager::~RenderingManager()
@@ -889,6 +892,9 @@ namespace MWRender
         if (mFieldOfViewOverridden)
             fov = mFieldOfViewOverride;
         mViewer->getCamera()->setProjectionMatrixAsPerspective(fov, aspect, mNearClip, mViewDistance);
+
+        mUniformNear->set(mNearClip);
+        mUniformFar->set(mViewDistance);
     }
 
     void RenderingManager::updateTextureFiltering()
--- a/apps/openmw/mwrender/renderingmanager.hpp
+++ b/apps/openmw/mwrender/renderingmanager.hpp
@@ -83,6 +83,9 @@ namespace MWRender
         SceneUtil::UnrefQueue* getUnrefQueue();
         Terrain::World* getTerrain();
 
+        osg::Uniform* mUniformNear;
+        osg::Uniform* mUniformFar;
+
         void preloadCommonAssets();
 
         double getReferenceTime() const;
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -112,7 +112,7 @@ class ClipCullNode : public osg::Group
             }
 
             // move the plane back along its normal a little bit to prevent bleeding at the water shore
-            const float clipFudge = -5;
+            const float clipFudge = -6;
             modelViewMatrix->preMultTranslate(mCullPlane->getNormal() * clipFudge);
 
             cv->pushModelViewMatrix(modelViewMatrix, osg::Transform::RELATIVE_RF);
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -112,7 +112,7 @@ class ClipCullNode : public osg::Group
             }
 
             // move the plane back along its normal a little bit to prevent bleeding at the water shore
-            const float clipFudge = -30;
+            const float clipFudge = -5;
             modelViewMatrix->preMultTranslate(mCullPlane->getNormal() * clipFudge);
 
             cv->pushModelViewMatrix(modelViewMatrix, osg::Transform::RELATIVE_RF);
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -112,7 +112,7 @@ class ClipCullNode : public osg::Group
             }
 
             // move the plane back along its normal a little bit to prevent bleeding at the water shore
-            const float clipFudge = -5;
+            const float clipFudge = -30;
             modelViewMatrix->preMultTranslate(mCullPlane->getNormal() * clipFudge);
 
             cv->pushModelViewMatrix(modelViewMatrix, osg::Transform::RELATIVE_RF);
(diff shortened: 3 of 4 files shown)


scrawl:
Increasing the fudge has its downsides. These are best seen with the player waist-deep in the water; the player's reflection will be 'cut' at some point. The current value of 5 makes this effect not too noticable, but anything higher is problematic. We'd better find another way to fix the shoreline issues (maybe some 'foam' on top that hides it?).

drummyfish:
You're right, I didn't check third person, my bad. I'll try to look into the foam idea - maybe the water depth that's used for refractions could be used in the fragment shader to add the foam to shallow areas.

drummyfish:
Okay, turns out that can't be done because water depth is computed from refraction map which is the source of the problem. I'll have to think of something else.

waterbug

drummyfish:
Alright I may have a solution: if I leave out the normal map, I can get the real water depth.

waterbug2

drummyfish:
So I found the problem was at bumpy reflections. I set a limit for water depth at which I suppress the normal map for reflections which fixes the bug. I'll format the code a bit and will submit it.

waterbug3

psi29a:
Nice! :)

scrawl:
That's an interesting idea. 

What about the change to setWaterLevel? Did you mean to leave that in?

Would it be possible to reduce the fudge factor now, is it still needed at all?

drummyfish:
Yes, the change is intentional, it does the same transformation with only two operations instead of three.

I tried setting the fudge to zero and there were still very small (1px) artifacts - I figured they were probably caused by refractions, so I did the same thing with refractions as I did with reflections, i.e. I suppressed the bump map at shallow depth. The artifacts then got even smaller, however they still persist, I don't know why exactly. Here is what it looks like:

waterbug4

I think we'll be able to decrease the fudge to much lower value than 5 now though.

drummyfish:
So a few things I discovered:

- If refractions are turned off in settings, reflection depth map is unavailable in the shader and so the water depth can't be computed from it so this fix doesn't work with refractions turned off. We could force the refraction pass even with refractions off just for this fix to be possible.
- I experimented with the fudge value and in normal situations lower values than 5 can be set, but at more extreme cases (camera just above the water level) the artifacts still do appear from time to time, so I ended up even increasing it to 6 to further suppress it. So the bug isn't 100% fixed, but is hardly ever noticeable now.

I'll commit my changes so far and will be looking into further improvements.

scrawl:
> Yes, the change is intentional, it does the same transformation with only two operations instead of three.

Is it really? The original moves the camera to the waterlevel, mirrors it, then moves it back, i.e. mirrors the scene at (0,0,waterlevel). I don't get where 2xwaterlevel would come from. Have you tested this code somewhere with waterlevel != 0 ? Keep in mind that matrix multiplications are not commutative.

Anyway, our usual rule is to not optimize code if it doesn't show up in the profiler.

> I tried setting the fudge to zero and there were still very small (1px) artifacts - I figured they were probably caused by refractions, so I did the same thing with refractions as I did with reflections, i.e. I suppressed the bump map at shallow depth. The artifacts then got even smaller, however they still persist, I don't know why exactly.

Might be because the textures are rendered at a lower resolution.

I had another idea that might work. In your pictures showing the perturbed and non-perturbed refraction depth, we can see exactly where the artifacts are. Anything that's maximum depth (red) on the perturbed image but isn't red on the non-perturbed image is going to be a refraction artifact. The same should apply to the reflection image, since it's cut off by the shore in the same way. It just uses different perturbation coordinates, so we need to do another sample of the refraction map with the perturbed reflection coordinates. Once we've identified these 'artifact' pixels, the fix is as simple as using the un-perturbed image for these pixels.

drummyfish:
I think the water transformation should work:

watertrans

But I only tested in Seyda Neen... any idea where I could test the water level != 0?

I just thought a shorter line was nicer, but I can undo this and take a look at how it performs later.

scrawl:
I object to the last commit. The main use case for turning off refractions is for performance gain. If we render them anyway when disabled there won't be any performance benefit. We'll just have to accept the fix doesn't work with refractions disabled.

> I think the water transformation should work:

Hm.. maybe it does. The diagram makes sense. But you should never push code without testing it. Try any interior cell with water (ex. beshara) and play with the water levels in the console (`setwaterlevel <value>`). Anyway, unless we've got any evidence this is performance critical code, I still consider this change wasted work (though now it's already done, there's no need to undo it I guess). 

Any interest in trying out the idea from my last post? Seems like it might work even better, possibly no longer require the fudge value, and perturbations can still show on shallow water.

drummyfish:
Okay, I'll revert the last commit back and will test the water levels. Also I think I need to look into how the water depth is computed - camera frustum, i.e. near and far values, should be taken into account. Also I think the depth values are in logarithmic space and need to be linearized.

>I had another idea that might work...

I'll try this out too.

drummyfish:
Tested changing water levels and it works:

water levels

I tried detecting the artifacts with the difference of distorted and undistorted water depth, as you suggested, but it also detects places that aren't artifacts, such as edges of underwater hills. So I kept my solution for now.

drummyfish:
I also looked at the profiler and there is no change in performance, so the change in water.cpp is only about readability, which I think is better now.

Other than that I don't know what else to do for this PR, could it maybe be merged now? I'd like to further look into the water shader, maybe the water droplet effects during rain etc., but that's another issue so I don't want to start it now.

scrawl:
BUMP_SUPPRESS_DEPTH is defined as a percentage of the far clip distance (currently 3% ?). Shouldn't we use an absolute value instead? We don't want the water to look different depending on how far the user has his view distance set. As a test, I set my view distance to 100 times the default (as I might do to use 'distant terrain'), which made the water reflection completely flat.

With the default view distance, everything seemed fine and there were no artifacts, though there were a few occasions where the lack of perturbation near shores seemed noticable (i.e. the low resolution of the texture suddenly becomes apparent).

> I tried detecting the artifacts with the difference of distorted and undistorted water depth, as you suggested, but it also detects places that aren't artifacts, such as edges of underwater hills.

Can't we test only for 'maximum depth' somehow (i.e. only catch fragments that have nothing rendered on them)?

Or maybe we could combine both tests? Like, if we're in shallow water and the difference between depths is large, there must be an artifact?

drummyfish:
I tried different view distances and you're right, but the same happens when I set it to absolute value. I think it should be set as a percentage though, because AFAIK the water depth is computed from the depth map which depends on the frustum (near/far), so in the shader we're not getting world-space water depth but rather water depth in terms of depth buffer. I also noticed that the water transparency in different water depths differs with view distance, which it shouldn't, so maybe the code that linearizes depth values (the one I moved to linearizeDepth function) is buggy?

drummyfish:
So the code I've moved to `linearizeDepth`  is clearly copied from [here](https://stackoverflow.com/questions/6652253/getting-the-true-z-value-from-the-depth-buffer), but the values that go in are in wrong form I think. The formula expects negative near and far values (zNear, zFar, mentioned in the comments), but the supplied values in our code seem to be positive. Also the input depth is supposed to be in 0 to 1 range while we supply `depthPassThrough` which is not in this range (it's a value passed from vertex shader and does not undergo perspective divide). I'll be looking into this occasionally.

drummyfish:
Took me some work but I managed to obtain frustum-independent water depth :) This will hopefully lead to solving the issue.

waterbug8

drummyfish:
Pushed, could you guys review how the water looks now? As a part of reworking the water depth I also set underwater visibility to lower value, which looks better I think (the water was so clear I sometimes couldn't even notice it was there), but can be reset back.

drummyfish:
I am also experimenting with adding the rain ripple effect to the shader - [here](https://www.youtube.com/watch?v=jesMMSiPElY&feature=youtu.be) is a preview of what I made so far. I'll be glad for opinions on how to make the effect look better.

scrawl:
I'll review it tomorrow and merge if all is good.

The ripples look pretty good to me as they are. If we used a cube map for reflections like the original game did, it'd be easy to get a high quality ripples effect by changing the normals alone. Alas, planar reflections don't work for that so well. Even GTA V has somewhat mediocre looking rain ripples IIRC.

scrawl:
Looks like there are artifacts for larger distances.

screenshot003

Still better than master, so I'll just merge now.

drummyfish:
Thanks, I'll look into it plus I'll try to implement the basic rain ripples.

PR #1470, opened 2017-09-20
ESS-Importer: Convert magic projectiles

MiroslavR:
Implements: https://bugs.openmw.org/issues/2320

This PR imports and converts magic projectiles, and partially imports active spells (SPLM record). To make this conversion possible, OpenMW save format was slightly changed. Particularly, these redundant MPRJ fields were removed:
* sourceName (can be determined from spellId)
* effects (can be determined from spellId)
* stack (was always false)
* sound (was unused)

The interface for launching magic projectiles was also refactored to remove redundancy.

Strange things noticed during development:
* `ESMStore::find(const std::string &id)` does not seem to work for spell records,
* Magic projectile speed stored by vanilla is 1000 times greater than OpenMW, e.g. `1250` vs `1.25` for spell "fireball" (content file defines `1.25` speed for "Fire Damage" effect). I'm not sure where that factor comes from.

### DIFF
--- a/apps/essimporter/CMakeLists.txt
+++ b/apps/essimporter/CMakeLists.txt
@@ -17,6 +17,7 @@ set(ESSIMPORTER_FILES
     importscri.cpp
     importscpt.cpp
     importproj.cpp
+    importsplm.cpp
     importercontext.cpp
     converter.cpp
     convertacdt.cpp
--- a/apps/essimporter/converter.cpp
+++ b/apps/essimporter/converter.cpp
@@ -1,12 +1,12 @@
 #include "converter.hpp"
 
 #include <stdexcept>
+#include <algorithm>
 
 #include <osgDB/WriteFile>
 
 #include <components/esm/creaturestate.hpp>
 #include <components/esm/containerstate.hpp>
-#include <components/esm/projectilestate.hpp>
 
 #include "convertcrec.hpp"
 #include "convertcntc.hpp"
@@ -461,11 +461,7 @@ namespace ESSImport
             if (!pnam.isMagic())
             {
                 ESM::ProjectileState out;
-                out.mId = pnam.mArrowId.toString();
-                out.mPosition = pnam.mPosition;
-                out.mOrientation.mValues[0] = out.mOrientation.mValues[1] = out.mOrientation.mValues[2] = 0.0f;
-                out.mOrientation.mValues[3] = 1.0f;
-                out.mActorId = convertActorId(pnam.mActorId.toString(), *mContext);
+                convertBaseState(out, pnam);
 
                 out.mBowId = pnam.mBowId.toString();
                 out.mVelocity = pnam.mVelocity;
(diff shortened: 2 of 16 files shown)


scrawl:
As stated on another PR I can't remember, I am slightly concerned about forwards-incompatible savegame changes at this point. They'll inconvenience people that are hopping between master and the latest release, for example. Until we have a better way to deal with this fallout, changes should be limited whenever possible.

In this case, saving the game while a projectile is in-air seems unlikely; but there's a caveat: 'rogue' projectiles need to be deleted at some point if they never hit a target, e.g. if they're further away than X units from the player. This isn't currently done (and probably should be done, regardless of this PR).

MiroslavR:
Well, I'm stumped. I thought we didn't even guarantee _backwards_ compatibility until 1.0 is released (but we just did that out of the kindness of our hearts). And now I'm hearing about maintaining forwards compabitility as well... Ugh, that sounds like a massive PITA. Is it worth it to inconvenience developers / possibly hurt development for satisfying a few version hoppers (I wonder if there even are any such people)? If you want to enforce that, it should at least be mentioned somewhere (other than in some random PR).

In any case, I don't know what to do with this PR then. It kinda relies on these changes in the save format.

scrawl:
Not really a hard policy, just something to keep in mind when judging the pros and cons of a changeset. 

>  Is it worth it to inconvenience developers / possibly hurt development for satisfying a few version hoppers 

I'd take inconveniencing developers over inconveniencing users every time. There are more users than developers.

>  (I wonder if there even are any such people)?

We can't assume that there aren't, unless we made a poll or something. I'd imagine it's a common scenario to play with the latest release while occasionally trying master to check on bugs, test new features etc. I seem to recall at least one error report on the forum that appeared to be the result of forward incompatibility (that was a while ago, but so was our last format change).

Having a savegame you can no longer load is effectively lost data. Worse, it'll just fail with a cryptic error that provides no hint as to what you did 'wrong'. Sure, you can go back to master but no one should have to use unreleased versions to be able to continue playing. Even as a developer, I've sometimes been annoyed when using bisect or using an older release to test something that most of my 'testing' savegames would no longer load. Maybe we should add some way for malformed records to be ignored, so the game can still be loaded albeit with some things possibly missing?

> it should at least be mentioned somewhere

Yes, we should.

>  I don't know what to do with this PR then.

I think this PR should be fine to merge if we can add the cleanup for far-away projectiles that never hit a target (something that should probably be done anyway, lest more and more of them accumulate) - then it should be unlikely for any particular savegame to contain projectile state, which reduces the impact of the forward compatibility issue. Even if a save does contain projectile state, I save it in master, go back to last release to find it won't load - I can 'fix' the save by going back to master and re-saving in another location with no projectiles. That seems acceptable to me.

MiroslavR:
I'm worried about this new policy blocking progress towards 1.0. It's very possible we will run into problems where breaking forward compatibility is inevitable.

As far as rogue projectiles go, they seem quite unlikely to occur. All projectiles are cleared when sleeping/waiting and upon changing the worldspace. Ballistic projectiles should also hit the water plane at some point after surpassing the active cells (unless the worldspace has no water). Same can't be said about magic projectiles as their trajectory is linear - I guess rogue projectiles only become a problem when you fire magic bolts at the sky?

When do you think would be the appropriate time to clear them - during the loading phase or periodically during simulation?

scrawl:
> I'm worried about this new policy blocking progress towards 1.0. It's very possible we will run into problems where breaking forward compatibility is inevitable.

It's not a concrete policy. I would judge everytime based on the benefits of the change vs the severity of the break. Of course, there will be cases where a break is inevitable or hard to avoid then we'll just have to accept it.

> seem quite unlikely to occur

Oh, you're right.

> When do you think would be the appropriate time to clear them - during the loading phase or periodically during simulation?

During the simulation, I'd say.

MiroslavR:
> During the simulation, I'd say.

Okay, done.

> Magic projectile speed stored by vanilla is 1000 times greater than OpenMW, e.g. 1250 vs 1.25 for spell "fireball" (content file defines 1.25 speed for "Fire Damage" effect). I'm not sure where that factor comes from.

Found out that factor is fTargetSpellMaxSpeed (1000 by default).

PR #1471, opened 2017-09-21
first pass on optimization of nif parsing functions from the file stream

crussell187:
This pull request could be used as is but I also want to start the discussion on further optimization.

I've done a few things in this changeset to help performance:
1. Hoisted all of the functions from nifstream.cpp to nifstream.hpp so that they can be considered for inlining by the compiler.
2. Removed the manual little endian massaging of the stream and instead do direct casting of the buffer read to the target data type.  This will break portability to big endian architectures.  Not sure if that was a target but I can always put it as a fallback if not on an x86 architecture.
3.  Pre-read the size of the entire structure from the buffer to reduce calls to inp->read().
4.  Doubled the size of the file buffer.

Using google benchmark on multiple iterations of parsing base_anim.nif, I get this from the original code:
---------------------------------------------------------
Benchmark                  Time           CPU Iterations
---------------------------------------------------------
^[[m^[[0;32mBM_NifRead_mean   ^[[m^[[0;33m  28403072 ns   28390625 ns ^[[m^[[0;36m        25^[[m
^[[m^[[0;32mBM_NifRead_median ^[[m^[[0;33m  28286460 ns   28125000 ns ^[[m^[[0;36m        25^[[m
^[[m^[[0;32mBM_NifRead_stddev ^[[m^[[0;33m    769761 ns     811667 ns ^[[m^[[0;36m        25^[[m
^[[m

After my optimizations, I get this:
---------------------------------------------------------
Benchmark                  Time           CPU Iterations
---------------------------------------------------------
^[[m^[[0;32mBM_NifRead_mean   ^[[m^[[0;33m  21612806 ns   21594238 ns ^[[m^[[0;36m        32^[[m
^[[m^[[0;32mBM_NifRead_median ^[[m^[[0;33m  21431406 ns   21484375 ns ^[[m^[[0;36m        32^[[m
^[[m^[[0;32mBM_NifRead_stddev ^[[m^[[0;33m    580247 ns     639873 ns ^[[m^[[0;36m        32^[[m

Just taking the mean, we get about a 30% speedup for this.  

Other things I want to consider is not using the traditional istream read and instead provide a funciton to move data directly from the internal 8K buffer into the target data type (uint16_t,uint32_t,uint64_t,osg::Vec3f,osg::Quat).

This will cut the move requirements from:
istream buffer -> intermediate buffer -> target data type
to this:
istream buffer -> target data type

Also thought using some SIMD intrinsics or even a direct memcopy may be quicker than some of these go between functions we use for the types.  These efforts might eek out another 10% or so.

Thoughts?


psi29a:
This will break on Debian and Ubuntu big endian systems, the packages for those systems are built for without asking. We want to be as portable as we can reasonably be. 

Please add a fallback as you suggest.

kcat:
> Removed the manual little endian massaging of the stream and instead do direct casting of the buffer read to the target data type. This will break portability to big endian architectures. Not sure if that was a target but I can always put it as a fallback if not on an x86 architecture.

I'd suggest not to break it. I don't know what the status of OpenMW on big-endian is, but it'd be good to not break it if possible. You should be able to make an inline wrapper function which does a simple casting on little-endian systems and leaves the byte swapping to only occur on big-endian using a compile-time check.

> Pre-read the size of the entire structure from the buffer to reduce calls to inp->read().

Certain structures have different sizes depending on values in the structure. There may be flags indicating the presence (or absence) of fields, and some contain a count of items to read.

components/nif/nifstream.hpp
```diff
@@ -2,7 +2,7 @@
 
 #ifndef OPENMW_COMPONENTS_NIF_NIFSTREAM_HPP
 #define OPENMW_COMPONENTS_NIF_NIFSTREAM_HPP
-
+#include <xmmintrin.h>
```
kcat: This isn't standard. Do you really need SSE?

components/nif/nifstream.hpp
```diff
@@ -26,10 +26,44 @@ class NIFStream {
...
-    float read_le32f();
+    uint8_t read_byte() {
+        uint8_t byte;
+        inp->read((char*)&byte, 1);
+        return byte;
+    }
+
+    uint16_t read_le16() {
+        alignas(2) uint8_t buffer[2];
+        inp->read((char*)buffer, 2);
```
kcat: Seems MSVC2013 doesn't handle ```alignas```. It's not necessary here anyway, you can do:
```
uint16_t ret;
inp->read(reinterpret_cast<char*>(&ret), 2);
return ret;
```
The same for le32, le64, and le32f. Also, these should be marked as inline if you want them inlined.

scrawl:
Regarding the benchmark, animation files like base_anim are probably a worst-case scenario, with many small fields to read, but this file is only loaded once by the engine. Typical mesh files with vertices/triangles can be read in big chunks with not much effort. Any chance we could get a benchmark for an actual game session?

crussell187:
> This will break on Debian and Ubuntu big endian systems, the packages for those systems are built for without asking. We want to be as portable as we can reasonably be.

What are the list of architectures we need support?  I'll explore some ways to enable it on ARM if I detect the compilation target for ARM is little endian.  In the current practical use case,  a significant portion of our use base is x86 and will get the benefits.  

> This isn't standard. Do you really need SSE?

This was included to get both compilers to include __m128.  Now with more research, I am finding that __m128 is not supported universally so I will find another portable replacement.  This data type is to hint to the compiler to use that big 128 bit register file in the machine sitting there doing nothing to move data larger than 64 bits (Vector3, Vector4, Matrix, etc.)

> Certain structures have different sizes depending on values in the structure. There may be flags indicating the presence (or absence) of fields, and some contain a count of items to read.

I agree with you but a lot of calls I saw in nif code were of the following:
if(nif->getInt()){
   foo = nif->getFloat();
   foo2 = nif->getInt();
   foo3 = nif->getMatrix3();
}
All of these can be amortized with buffer read.  That being said for readability of the nif code I would prefer to go this route to reduce the overhead of the istream read calls():

> Other things I want to consider is not using the traditional istream read and instead provide a function to move data directly from the internal 8K buffer into the target data type (uint16_t,uint32_t,uint64_t,osg::Vec3f,osg::Quat).

crussell187:
> Regarding the benchmark, animation files like base_anim are probably a worst-case scenario, with many small fields to read, but this file is only loaded once by the engine. Typical mesh files with vertices/triangles can be read in big chunks with not much effort. Any chance we could get a benchmark for an actual game session?

I'll do some isolated benchmarks first with the other file types.  Although those nifs are streaming more uniform data, I still expect some improvements due to the little endian massaging, lack of inlining, and lack of istream read amortization for Vector3 calls.

Do we have a good way of benchmarking in game?  I would like to perform a worst case asset load.  What part of morrowind is the worst? :)

psi29a:
To give you an example, here is the openal-soft (from kcat/chris) package that I maintain on Debian (and Ubuntu):
https://buildd.debian.org/status/package.php?p=openal-soft
That's 23 different archs. 

https://buildd.debian.org/status/package.php?p=openmw
Not all are supported because we're not yet in main. That will happen soon.

Please don't just drop BE support because of gaining a few ms.

kcat:
> This was included to get both compilers to include __m128. Now with more research, I am finding that __m128 is not supported universally so I will find another portable replacement.

__m128 is part of the SSE intrinsics, and not all systems support SSE. Anything beyond 64-bit data types requires system- and compiler-specific extensions, which is likely to be an unnecessary optimization (too much work/maintenance for too little gain). I bet you could find more speed improvements by loading the whole file into memory first, then read the individual elements from there (kind of like how we handle save games, but in reverse).

scrawl:
> Do we have a good way of benchmarking in game? I would like to perform a worst case asset load. What part of morrowind is the worst? :)

Running with `--skip-menu --start=balmora` might be a good test.

Note that most of the NIF loading happens from background threads and is most likely going to be on-time anyway. Therefore I'm not overly keen on dodgy optimizations that make the code harder to read and maintain. SSE sounds like overkill, and from what I've heard doesn't gain you much for code that isn't data-oriented. We should stick to changes that have a measured benefit. You've measured the overall speedup from this branch, but can you list, for the 4 improvements in your original post, the impact of each change respectively?

crussell187:
> Note that most of the NIF loading happens from background threads and is most likely going to be on-time anyway. Therefore I'm not overly keen on dodgy optimizations that make the code harder to read and maintain. SSE sounds like overkill, and from what I've heard doesn't gain you much unless you're using it to batch calculations. We should stick to changes that have a measured benefit. You've measured the overall speedup from this branch (using a release build I hope), but can you list, for the 4 improvements in your original post, the impact of each change respectively?

Latest gcc with -O3.  I wouldn't care to test this on debug (inlining wouldn't even be considered there.)

I would like to replace the intent of SSE specifically with just SIMD.  If the inlining is done correctly and call overhead is removed, you can do faster data movement from the file buffer to your data structures with the SIMD moves.  The only thing faster will be a direct mem copy for large data structures.  There is a portable way to do this and I will update my pull to reflect that.

I'll prepare some benchmarks with each separate change.

crussell187:
So I'll eat my words on my conclusions about SIMD.  I don't think the operations made a difference to the code.  This new commit has all of the SIMD taken out and a portable ifdef to only rely on little endian alignment on an x86 architecture.  Otherwise, it does the big endian conversion from the nif.  Below is the benchmark for the base_anim.nif with the new code:

---------------------------------------------------------
Benchmark                  Time           CPU Iterations
---------------------------------------------------------
BM_NifRead_mean     20961216 ns   20947266 ns         32
BM_NifRead_median   20771219 ns   20996094 ns         32
BM_NifRead_stddev     612575 ns     652292 ns         32

Slightly faster than my initial optimization.  Hopefully this code seems much more portable now.  I'll post some more benchmarks soon.

psi29a:
yay, thank you!

scrawl:
Great! The optimization even removes more lines than it added :)

I wouldn't call bloodsplat.nif a typical file though (it's a particle effect, which use lots of fragmented data also).

components/nif/nifstream.hpp
```diff
@@ -21,16 +21,57 @@ namespace Nif
...
+{
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386) || defined(_M_IX86)
+    pIStream->read((char*)dest, numInstances * sizeof(T));
+#else
+    char buffer[numInstances * sizeof(T)];
+    pIStream->read((char*)buffer, numInstances * sizeof(T));
+    /*
+    Due to the loop iterations being known at compile time,
+    this nested loop will most likely be unrolled
+    */
```
scrawl: Looks like a wrongly copy-pasted comment.

components/nif/nifstream.hpp
```diff
@@ -41,6 +41,25 @@ template <uint32_t numInstances, typename T> inline void readLittleEndianBufferO
...
+    for (uint32_t i = 0; i < numInstances; i++)
+    {
+        dest[i] = 0;
+        for (uint32_t byte = 0; byte < sizeof(T); byte++)
+            dest[i] |= ((T)buffer[i * sizeof(T) + byte]) << (byte * 8);
```
kcat: Does this actually work for the ```osg::Vec*``` types?
```
osg::Vec3f *dest;
...
dest[i] |= ((osg::Vec3f)buffer[i * sizeof(osg::Vec3f) + byte]) << (byte * 8);
```
looks extremely sketchy.

components/nif/nifstream.hpp
```diff
@@ -41,6 +41,25 @@ template <uint32_t numInstances, typename T> inline void readLittleEndianBufferO
 #endif
 }
 
+template <typename T> inline void readLittleEndianDynamicBufferOfType(Files::IStreamPtr &pIStream, T* dest, uint32_t numInstances)
+{
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386) || defined(_M_IX86)
+    pIStream->read((char*)dest, numInstances * sizeof(T));
```
kcat: Can we assume the ```osg::Vec*``` types are packed?

components/nif/nifstream.hpp
crussell187: This template should be used sparingly.  It only works for guaranteed packed objects.  The developer of OSG did a good job defining these objects.  If you look at the definition, the data for the individual floats is stored as an array.  It's guaranteed to be packed.  If we start deriving from osg::Vec3f, the object may have more storage and then this would break but I don't see examples of that in openmw that i'm aware of.  We could put static assertions that the osg vec types equal the size of the nif counterpart for them.

components/nif/nifstream.hpp
crussell187: Yes.  Please see my previous comment.

components/nif/nifstream.hpp
crussell187: Yes it is.  I'll fix it on the dynamic size template.

MiroslavR:
Can you make the endianness checks reusable (e.g. a macro in endian.hpp)?

crussell187:
Latest updates include fixes so that the template functions only read non POD types (float, uint32_t, uint64_t, etc.).  Also validated that both code paths work in game.

crussell187:
> Can you make the endianness checks reusable (e.g. a macro in endian.hpp)?

There's no true endianness check that's portable (something provided by the compiler that says __ARCHITECTURE_LITTLE_ENDIAN vs. __ARCHITECTURE_BIG_ENDIAN).  I think some architectures can even change modes at runtime between the 2.  The best defines we can make are ARCHITECTURE_KNOWN_TO_BE_LITTLE_ENDIAN and ARCHITECTURE_KNOWN_TO_BE_BIG_ENDIAN.  The x86 check I've currently provided can fall into the ARCHICTECTURE_KNOWN_TO_BE_LITTLE_ENDIAN.

psi29a:
Doesn't endianess checks happen during cmake?

crussell187:
Seems you are right for cmake 3.0 forward:
https://stackoverflow.com/questions/3877344/detect-endianness-with-cmake

Learn something new every day :).

Although, taking ARM specifically as an example, you can set both OS/system level endianness through a system control register and process level endianness through a process specific control register:
https://community.arm.com/processors/f/discussions/3313/how-and-where-do-i-change-endianness-in-cortex-a9

Not sure if other architectures try to give this configurability but since ARM would be a major cross-platform target, it's something to keep in mind.  Mind you, if someone actively changes endianness after compile time then we may just say they are on their own.

crussell187:
Just checking in.  What else would you like to see from me before this is pulled in?  I've validated both paths in game.  Do you want me to delete nifstream.cpp as it is effectively an empty file now?  More benchmarks?

psi29a:
I don't see the need for nifstream.cpp now that you've emptied it. One less file for the compiler...

scrawl:
Looks good to me.

scrawl:
FYI, if you're thinking of replacing the `vector<osg::Vec*f>` with `osg::Vec*Array` to get rid of the copying when we load them into osg: I already tried that, had to [revert](https://github.com/OpenMW/openmw/commit/aa8459b5c71d07ef46dbb5f269d4f4001cdc6b2b) later because it caused rare crashes. The VecArrays also store a BufferObject, this gets assigned by osg::Geometry and breaks our multithreaded resource loading (i.e. crashes if two threads try loading the same nif file into OSG at the same time, or something like that). I don't remember the full details but the issue was extremely nasty and withstood all my attempts of working around it :(

scrawl:
Another FYI, looks like your commits aren't being recognized as yours by github (invalid email most likely). Might want to fix that for next time!

PR #1473, opened 2017-09-21
Allow distant terrain to be seen when enabled

kcat:
Currently when distant terrain is enabled, the viewing distance clips it preventing it from being seen. This automatically extends the view distance when distant terrain is enabled, and also adjusts the fog so that it's not blocked out against the background. The viewing distance option is effectively ignored when distant terrain is enabled, and when distant terrain is disabled everything should still behave like normal.

The fog adjustments are based on MGE XE's default fog parameters when its distant land is enabled. Currently these are not modifiable by users/modders, but it wouldn't be difficult to add a settings.cfg section for the base fog values if it's desirable (the per-weather modifiers would probably need to wait until post-1.0 for omwaddons to specify weather settings). A settings section for these things might be useful anyway for future options to use exponential curves and atmospheric scattering and the like.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -53,7 +53,6 @@
 
 namespace
 {
-    float DLRenderDistance;
     float DLLandFogStart;
     float DLLandFogEnd;
     float DLUnderwaterFogStart;
@@ -231,7 +230,6 @@ namespace MWRender
 
         mWater.reset(new Water(mRootNode, sceneRoot, mResourceSystem, mViewer->getIncrementalCompileOperation(), fallback, resourcePath));
 
-        DLRenderDistance = Settings::Manager::getFloat("distant viewing distance", "Camera");
         DLLandFogStart = Settings::Manager::getFloat("distant land fog start", "Fog");
         DLLandFogEnd = Settings::Manager::getFloat("distant land fog end", "Fog");
         DLUnderwaterFogStart = Settings::Manager::getFloat("distant underwater fog start", "Fog");
@@ -312,10 +310,8 @@ namespace MWRender
         mFirstPersonFieldOfView = Settings::Manager::getFloat("first person field of view", "Camera");
         mStateUpdater->setFogEnd(mViewDistance);
 
-        mFarClip = mDistantTerrain ? DLRenderDistance : mViewDistance;
-
         mRootNode->getOrCreateStateSet()->addUniform(new osg::Uniform("near", mNearClip));
-        mRootNode->getOrCreateStateSet()->addUniform(new osg::Uniform("far", mFarClip));
+        mRootNode->getOrCreateStateSet()->addUniform(new osg::Uniform("far", mViewDistance));
 
         mUniformNear = mRootNode->getOrCreateStateSet()->getUniform("near");
         mUniformFar = mRootNode->getOrCreateStateSet()->getUniform("far");
@@ -680,7 +676,7 @@ namespace MWRender
         rttCamera->setRenderOrder(osg::Camera::PRE_RENDER);
         rttCamera->setReferenceFrame(osg::Camera::ABSOLUTE_RF);
         rttCamera->setRenderTargetImplementation(osg::Camera::FRAME_BUFFER_OBJECT, osg::Camera::PIXEL_BUFFER_RTT);
-        rttCamera->setProjectionMatrixAsPerspective(mFieldOfView, w/float(h), mNearClip, mFarClip);
+        rttCamera->setProjectionMatrixAsPerspective(mFieldOfView, w/float(h), mNearClip, mViewDistance);
         rttCamera->setViewMatrix(mViewer->getCamera()->getViewMatrix());
         rttCamera->setViewport(0, 0, w, h);
 
@@ -945,10 +941,10 @@ namespace MWRender
         float fov = mFieldOfView;
         if (mFieldOfViewOverridden)
             fov = mFieldOfViewOverride;
-        mViewer->getCamera()->setProjectionMatrixAsPerspective(fov, aspect, mNearClip, mFarClip);
+        mViewer->getCamera()->setProjectionMatrixAsPerspective(fov, aspect, mNearClip, mViewDistance);
 
         mUniformNear->set(mNearClip);
-        mUniformFar->set(mFarClip);
+        mUniformFar->set(mViewDistance);
     }
 
     void RenderingManager::updateTextureFiltering()
@@ -1008,8 +1004,6 @@ namespace MWRender
             else if (it->first == "Camera" && it->second == "viewing distance")
             {
                 mViewDistance = Settings::Manager::getFloat("viewing distance", "Camera");
-                if(!mDistantTerrain)
-                    mFarClip = mViewDistance;
                 if(!mDistantFog)
                     mStateUpdater->setFogEnd(mViewDistance);
                 updateProjectionMatrix();
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -51,6 +51,20 @@
 #include "terrainstorage.hpp"
 #include "util.hpp"
 
+namespace
+{
+    // These values are what MGE XE uses by default when distant land is enabled, and are specified
+    // in cells (8192 units each). Should make them settings somewhere? Or wait to expose them
+    // "properly"?
+    const float DLRenderDistance = 5.0f;
+    const float DLLandFogStart = 2.0f;
+    const float DLLandFogEnd = 5.0f;
+    const float DLUnderwaterFogStart = -0.5f;
+    const float DLUnderwaterFogEnd = 0.3f;
+    const float DLInteriorFogStart = 0.0f;
+    const float DLInteriorFogEnd = 2.0f;
+}
+
 namespace MWRender
 {
 
@@ -173,10 +187,12 @@ namespace MWRender
         , mResourceSystem(resourceSystem)
         , mWorkQueue(workQueue)
         , mUnrefQueue(new SceneUtil::UnrefQueue)
-        , mFogDepth(0.f)
+        , mLandFogStart(0.f)
+        , mLandFogEnd(std::numeric_limits<float>::max())
+        , mUnderwaterFogStart(0.f)
+        , mUnderwaterFogEnd(std::numeric_limits<float>::max())
         , mUnderwaterColor(fallback->getFallbackColour("Water_UnderwaterColor"))
         , mUnderwaterWeight(fallback->getFallbackFloat("Water_UnderwaterColorWeight"))
-        , mUnderwaterFog(0.f)
         , mUnderwaterIndoorFog(fallback->getFallbackFloat("Water_UnderwaterIndoorFog"))
         , mNightEyeFactor(0.f)
         , mDistantTerrain(false)
@@ -289,7 +305,7 @@ namespace MWRender
         mFirstPersonFieldOfView = Settings::Manager::getFloat("first person field of view", "Camera");
         mStateUpdater->setFogEnd(mViewDistance);
 
-        mFarClip = mDistantTerrain ? 8192.0f*5.0f : mViewDistance;
+        mFarClip = mDistantTerrain ? DLRenderDistance*8192.0f : mViewDistance;
 
         mRootNode->getOrCreateStateSet()->addUniform(new osg::Uniform("near", mNearClip));
         mRootNode->getOrCreateStateSet()->addUniform(new osg::Uniform("far", mFarClip));
@@ -482,14 +498,44 @@ namespace MWRender
     {
         osg::Vec4f color = SceneUtil::colourFromRGB(cell->mAmbi.mFog);
 
-        configureFog (cell->mAmbi.mFogDensity, mUnderwaterIndoorFog, color);
+        if(mDistantTerrain)
+        {
+            float density = std::max(0.2f, cell->mAmbi.mFogDensity);
+            mLandFogStart = (DLInteriorFogEnd*(1.0f-density) + DLInteriorFogStart*density) * 8192.0f;
+            mLandFogEnd = DLInteriorFogEnd * 8192.0f;
+            mUnderwaterFogStart = DLUnderwaterFogStart * 8192.0f;
+            mUnderwaterFogEnd = DLUnderwaterFogEnd * 8192.0f;
+            mFogColor = color;
+        }
+        else
+            configureFog(cell->mAmbi.mFogDensity, mUnderwaterIndoorFog, 1.0f, 0.0f, color);
     }
 
-    void RenderingManager::configureFog(float fogDepth, float underwaterFog, const osg::Vec4f &color)
+    void RenderingManager::configureFog(float fogDepth, float underwaterFog, float dlFactor, float dlOffset, const osg::Vec4f &color)
     {
-        mFogDepth = fogDepth;
+        if(mDistantTerrain)
+        {
+            mLandFogStart = dlFactor * (DLLandFogStart - dlOffset*DLLandFogEnd) * 8192.0f;
+            mLandFogEnd = dlFactor * (1.0f-dlOffset) * DLLandFogEnd * 8192.0f;
+            mUnderwaterFogStart = DLUnderwaterFogStart * 8192.0f;
+            mUnderwaterFogEnd = DLUnderwaterFogEnd * 8192.0f;
+        }
+        else
+        {
+            if(fogDepth == 0.0)
+            {
+                mLandFogStart = 0.0f;
+                mLandFogEnd = std::numeric_limits<float>::max();
+            }
+            else
+            {
+                mLandFogStart = mViewDistance * (1 - fogDepth);
+                mLandFogEnd = mViewDistance;
+            }
+            mUnderwaterFogStart = mViewDistance * (1 - underwaterFog);
+            mUnderwaterFogEnd = mViewDistance;
+        }
         mFogColor = color;
-        mUnderwaterFog = underwaterFog;
     }
 
     SkyManager* RenderingManager::getSkyManager()
@@ -520,23 +566,15 @@ namespace MWRender
             float viewDistance = mViewDistance;
             viewDistance = std::min(viewDistance, 6666.f);
             setFogColor(mUnderwaterColor * mUnderwaterWeight + mFogColor * (1.f-mUnderwaterWeight));
-            mStateUpdater->setFogStart(viewDistance * (1 - mUnderwaterFog));
-            mStateUpdater->setFogEnd(viewDistance);
+            mStateUpdater->setFogStart(mUnderwaterFogStart);
+            mStateUpdater->setFogEnd(mUnderwaterFogEnd);
         }
         else
         {
             setFogColor(mFogColor);
 
-            if (mFogDepth == 0.f)
-            {
-                mStateUpdater->setFogStart(0.f);
-                mStateUpdater->setFogEnd(std::numeric_limits<float>::max());
-            }
-            else
-            {
-                mStateUpdater->setFogStart(mViewDistance * (1 - mFogDepth));
-                mStateUpdater->setFogEnd(mViewDistance);
-            }
+            mStateUpdater->setFogStart(mLandFogStart);
+            mStateUpdater->setFogEnd(mLandFogEnd);
         }
     }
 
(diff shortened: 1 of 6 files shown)


scrawl:
Where does the value of `8192.0f*5.0f` for the far plane come from? What if I wanted to see further than that? I think the 'auto' or 'recommend' value should be an additional option (possibly the default), but we shouldn't take away the option for a manually set distance, right?

I should also mention there could be merits to enabling 'distant terrain' even if I don't want to see much further (say, only 1 additional cell). The non-distant path just renders terrain with no LOD whatsoever, which is probably slower than 'distant terrain' is (though the distant path needs to load cells outside of the viewing frustum, which possibly increases loading times).  I guess the name of the option is a bit of a mis-nomer. 'terrain paging' or 'terrain lod' might have been better. If we were able to make the paged terrain not load cells outside of the viewing frustum, we could probably enable paging by default with it no longer having any downsides.

kcat:
> Where does the value of 8192.0f*5.0f for the far plane come from? What if I wanted to see further than that?

That's the maximum fog distance possible. The largest fog end value is 5 (DLLandFogEnd), and the largest fog factor is 1 (Clear weather), and given the calculation ```dlFactor * (1.0f-dlOffset) * DLLandFogEnd * 8192.0f``` equals ```5.0f*8192.0f``` i.e. 5 cells. That also happens to be MGE XE's default draw distance for distant land.

Putting these distant-land-related values into a new section of settings.cfg can work, so the user can adjust them. I wouldn't know what to do with the current ```distant terrain``` option though since it seems wrong to have it in a different section, but moving it would disable it for users that already enabled it.

> I should also mention there could be merits to enabling 'distant terrain' even if I don't want to see much further (say, only 1 additional cell). The non-distant path just renders terrain with no LOD whatsoever, which is probably slower than 'distant terrain' is (though the distant path needs to load cells outside of the viewing frustum, which possibly increases loading times).

In my experience the non-distant path is faster. At least on a system that's tight on memory and has a relatively weak CPU, there's little reason to enable distant terrain unless you want to have a farther view distance despite the extra cost.

kcat:
Updated to fix a merge conflict. I also made the render distance with distant terrain a (```const```) variable instead of a magic number. Like I said, I can put these as settings in settings.cfg if you want, although I think it would be better to wait until we can allow mods to define each weather's fog factor and offset so that the amount of fogging that modded weather properties would create isn't dependent on user settings.

scrawl:
I suppose my concerns with this (or more generally, any changes to the settings) are:
* Taking away options that were there before (distant land with view distance > 5 cells).
* Changing the meaning of settings should be avoided (will make users wonder why their settings file doesn't do what it used to)
* Settings that implicitely change the value (or meaning) of another setting should be avoided, if the same can be achieved by just changing the latter setting manually. It's confusing to then be stuck with another setting that doesn't do anything anymore. If we want to have a set of 'presets' to use (like for MGE's fog) perhaps that could be handled by the [proposed](https://bugs.openmw.org/issues/4054) settings configuration window?

kcat:
> Taking away options that were there before (distant land with view distance > 5 cells).

That only worked by breaking fog. I can make the render distance configurable so it can go beyond 5 cells again, but without also being able to tweak the fog, you don't get much by rendering beyond the fog end. We'd need a more advanced fog implementation that doesn't have a strict 'fog end' to really see anything. As well, we need to be careful about over-extending the view frustum until we can improve the depth range distribution.

> Changing the meaning of settings should be avoided (will make users wonder why their settings file doesn't do what it used to)

I don't think the meaning of any settings is being changed, just separating settings that aren't useful together. The viewing distance isn't useful with distant terrain given its limits and use in fog calculations, and had to be extended far beyond its intended limit to make distant terrain practical which also broke certain aspects of rendering. This necessitates separating distant terrain from viewing distance.

Personally I wouldn't mind removing the viewing distance setting, since it's merely replicating a vanilla option that was needed with weak hardware (what was considered weak in 2002). I seriously doubt it's necessary these days, particularly with the range it has, and doesn't really give anything useful to the users.

> Settings that implicitely change the value (or meaning) of another setting should be avoided, if the same can be achieved by just changing the latter setting manually. It's confusing to then be stuck with another setting that doesn't do anything anymore.

As long as it's mentioned that the setting has no effect when another setting is enabled or disabled (or that enabling or disabling a setting makes another have no effect), I don't think it's that confusing. There's plenty of other settings that work this way, from the preload settings, normal/specular maps, small feature culling, hrtf, etc.

> If we want to have a set of 'presets' to use (like for MGE's fog) perhaps that could be handled by the proposed settings configuration window?

The main issue with that is that the vanilla fog isn't useful with distant terrain (it would obscure distant terrain), and the MGE fog isn't very useful without distant terrain (the fog wouldn't be very noticeable). A user would have to toggle two options to properly enable or disable a single feature, when a single toggle would do just as well. This would be particularly true if we want to expand the MGE fog with more options, such as exponential distance curves and atmospheric scattering shaders, which is more stuff that's only useful with distant terrain.

scrawl:
Sorry, I don't like where this is going. You're arguing what the game should be like or what's useful and what isn't, but that's not the point of having settings. Not everyone wants fog and not everyone cares about the Z-flickering enough to make them not use larger distances. Removing view distance slider is also a no go. Assuming what users want or need will end badly.

> A user would have to toggle two options to properly enable or disable a single feature, when a single toggle would do just as well.

In addition to the list of settings there'd be a 'presets' section, and selecting a preset will change the value of several settings at once with one click. We could think of the actual settings file as something like a web browser's about:settings section that most users won't deal with, they deal with a GUI that has presets of safe / recommended values.

If we can leave 'viewing distance' alone, add new settings for configuring fog (and/or separating it from the viewing distance), and 'distant terrain' only changes the terrain implementation and nothing else, I'd merge that in a heartbeat.

kcat:
Updated. Added a new ```distant viewing distance``` option which is used in place of ```viewing distance``` when ```distant terrain``` is enabled. Also added a new ```[Fog]``` section with a ```use distant fog``` option that enables the extended fog parameters separately from distant terrain. So enabling distant terrain sets the far clip plane to ```distant viewing distance```, but uses ```viewing distance``` as normal for the fog, and enabling ```use distant fog```  uses the farther fog parameters  regardless of distant terrain being used or not. The start and end points for the distant fog are also exposed.

I'm not sure exactly how to document the fog parameters in a way that a user would understand what they do (and why they shouldn't be modified if you don't know what you're doing). They'd also end up being largely boiler plate, repeating what the option is named (```Sets the distant fog start distance for exterior land```, ```Sets the distant fog end distance for underwater```, etc). What would you recommend to do for them?

alkis05:
I'm a player and just want to thank kcat for this commit. I've compiled it into my setup and apperantly everything works. I can see far way now, control the for distance. If I may sugest, it would be nice to have some in-game scroll bars to setup the clipping distance and fog configuration without having to edit manually openmw.cfg file.

Thunderforge:
@alkis05 There’s [a forum discussion](https://forum.openmw.org/viewtopic.php?f=3&t=4947) about adding distant terrain settings to the launcher, but it will be dependent on if this PR gets merged.

Thunderforge:
@kcat @scrawl What still needs to happen for this PR to get merged or rejected? There is interest in making distant terrain more user-friendly and easily accessible, which will depend on this PR.

scrawl:
Judging by the forum discussion, we seem to have some disagreements on the intended uses of distant terrain and fog. I think the best way forward would be for the settings system to not make any such assumptions nor couple different settings together, and have common use-cases/presets/etc. be handled by either a GUI that generates the config file, or by scripts (in the future that is, with extended post 1.0 scripting functions).
That's kind of what my requested updates to this PR have brought it to, with the exception of the separated 'viewing distance' and 'distant viewing distance' which don't make much sense to me from a low level perspective. That's basically my only gripe atm.

kcat:
> the separated 'viewing distance' and 'distant viewing distance' which don't make much sense to me from a low level perspective.

In my mind, this is necessary to avoid the range problems with the current `viewing distance` option. That has a predefined range that the in-game GUI moves it between, which is tailored for non-distant terrain. There is otherwise no defined value for the distant terrain's far plane, which leaves the user disadvantage for how to properly set it up. Even if a launcher preset handled it or the comments in the settings file instructed them on what to set, it still leaves the in-game GUI with a setting that can ruin it if it's accidentally interacted with, requiring the user to leave the game to fix it. At best, you'd have to alter the in-game GUI to treat the `viewing distance` value differently depending on the distant terrain option, but that's adding quite a bit of complexity to the settings system, particularly if the distant terrain option is added to the in-game GUI post-1.0.

scrawl:
> still leaves the in-game GUI with a setting that can ruin it if it's accidentally interacted with, requiring the user to leave the game to fix it

This is true, however, the alternative of just disabling the slider leaves me at a disadvantage if I wanted to quickly test the performance difference between the larger and the normal value.

How about, if the user has manually increased the viewing distance over its default maximum, the slider will adapt to make that value the new maximum.

> There is otherwise no defined value for the distant terrain's far plane, which leaves the user disadvantage for how to properly set it up.

It should be simple to set up if you want to see X times farther than before, you set it accordingly. The only thing that might make you guess is how to set it if you want to see the whole world, but perhaps that's better addressed by adding a recommendation to the settings file, or even making an 'auto' mode where it's set based on the size of the world.

kcat:
> How about, if the user has manually increased the viewing distance over its default maximum, the slider will adapt to make that value the new maximum.

Sounds like it's just asking for complaints. A user sets it really high, and they'll have a difficult time making finer adjustments since the slider covers a much larger range than normal. Plus, the value of the slider then changes depending on what it was when launched (a user manually sets a really high value, then in-game adjusts it down to say 70% or 80%, then the next time it's run they see the slider back at 100% despite nothing changing). Whether or not the user can put it back up depends on whether they restarted the game since turning it down or not. It makes the options feel inconsistent, like you don't know what you're getting with what you see in the GUI, and whether or not you can set what you want depends on what it was when the game started.

What I had envisioned is having a separate in-game view slider (with an appropriately different label) that replaces the normal one when distant terrain is enabled, and has a more appropriate range. Two separate widgets affecting two separate settings. The distant view slider can then also be modified as the engine changes (e.g. altering the range, or allowing "infinite" as the highest setting) without affecting the non-distant view slider. This could also be paired with some kind of toggle option to enable or disable distant terrain in-game, which would switch which slider is seen. Presuming you don't want both sliders visible at the same time anyway, which could be confusing and wasteful of UI real-estate. Either way, that would be a post-1.0 thing as it needs new UI text labels, but until then, that's something the launcher's advanced settings can deal with.

> It should be simple to set up if you want to see X times farther than before, you set it accordingly.

To a user, they'll just want to see distant land. They don't have a clue for "x times farther" or "x number of cells", they enable distant land because they want to see as much in the distance that they're able to. Ideally, they should be able to see into infinity (fog and curvature not withstanding). Asking them to set anything else is just asking them to set an arbitrary limit that they don't know the full consequences of. Having an option there for people who do understand it to play around with is one thing, but requiring people to know what to set (when the reason is partly technical and partly arbitrary) is a bit much IMO.

scrawl:
> To a user, they'll just want to see distant land. They don't have a clue for "x times farther" or "x number of cells", they enable distant land because they want to see as much in the distance that they're able to. Ideally, they should be able to see into infinity (fog and curvature not withstanding). Asking them to set anything else is just asking them to set an arbitrary limit that they don't know the full consequences of. Having an option there for people who do understand it to play around with is one thing, but requiring people to know what to set (when the reason is partly technical and partly arbitrary) is a bit much IMO.

But there isn't one perfect value, imo. Sure, seeing everything may be useful, but will also create more z-fighting artifacts, at least until we fix those. Some users will argue that rendering the outlines behind the end of fog is wasteful, and others still will complain that seeing too far ruins the illusion of a big world. So I think that some experimentation and personal preference is going to be involved, which would be an argument for adding the preliminary in-game slider, even if it's not perfect. The slider could also be set for a maximum of 2 times what's in the settings file, and/or use some hardcoded maximum for now, so it's possible to increase the value you had in the settings file. If the slider becomes too fiddly at large ranges we could make it so the left/right buttons can move it by a miniscule amount.

kcat:
> But there isn't one perfect value, imo. Sure, seeing everything may be useful, but will also create more z-fighting artifacts, at least until we fix those.

Exactly, which is why I said the setting is partly arbitrary and partly technical. As a result, you'd be asking the user to come up with their own precise value without knowing the consequences it has on depth precision, or without being able to see what it looks like or the performance impact it has. Most users should be happy with defaults (that's the point of defaults, something that works as-is for the majority of users), but users that want to change it can do so and they are the ones that should have to understand the full effects of what they're changing, rather than asking it of everyone.

> So I think that some experimentation and personal preference is going to be involved, which would be an argument for adding the preliminary in-game slider, even if it's not perfect.

Modifying the in-game UI isn't something I'm currently able to do (I've never messed about with the code for the settings GUI, and don't know where to start). I do agree that having an in-game slider would be useful, I just disagree that distant and non-distant terrain should be controlled by the same view distance setting since it would cause problems (both for consistency of meaning in the GUI, and with making users learn the technical aspects of the setting).

Consequently, I think this PR can be merged so @Thunderforge can add the new settings to the launcher, and the in-game slider can be dealt with separately.

Thunderforge:
@kcat I would be happy to add a setting to the Launcher if this were merged. Like the other settings, I can make it so that it still allows a user to manually put a custom value into the settings file and not have it overridden by the Launcher. That way, the majority of users who want recommended settings can just use the Launcher while advanced users wanting more control can edit the file directly.

scrawl:
Well, I still don't think that's a good idea to have two view distance settings in the backend. I think most of my concerns come from a developer perspective, but I'm sure they would affect some users too. For one, pulling this change would break existing config files in that the view distance they set for distant terrain is no longer used, probably resulting in a bit of head-scratching and/or confused bug reports. Second, I think the (already quite bloated) settings system should be made simpler when possible, and might want to avoid adding cases where a setting is used in one circumstance but not used in another circumstance. So whenever I want to change my view distance I need to remember whether I had distant terrain enabled or not, and find the correct setting to change. That's just asking for mistakes. Third, because distant terrain is still very much experimental and we don't know how many users are using it, it's probably best to not make it cross-contaminate other settings as that will again make it harder to retain backward compatibility if distant terrain / statics / fog / etc is ever overhauled or replaced by something else.
For the issue with some users enabling distant terrain without increasing their view distance and wondering why nothing happens, hopefully those actions (or lack thereof) were just caused by the poorly worded documentation, which I've fixed now (look in the repo because RTD currently fails to update).

> Exactly, which is why I said the setting is partly arbitrary and partly technical. As a result, you'd be asking the user to come up with their own precise value without knowing the consequences it has on depth precision, or without being able to see what it looks like or the performance impact it has. Most users should be happy with defaults (that's the point of defaults, something that works as-is for the majority of users), but users that want to change it can do so and they are the ones that should have to understand the full effects of what they're changing, rather than asking it of everyone.

If we extend the launcher as Thunderforge proposed, that would probably take care of having a nice default.

kcat:
> For one, pulling this change would break existing config files in that the view distance they set for distant terrain is no longer used, probably resulting in a bit of head-scratching and/or confused bug reports.

Distant terrain is still a feature-in-progress, so users should expect things about it to change. Plus, I doubt the view distance most users currently have set for distant terrain is in any way proper, just random large values they made up that seems to work or copied from someone else who made something up that seems to work, so providing a clean break with an official default setting would help get everyone on the same page. For the few users that have a reason to use a particular distance, it can still be set (just this time, without having to break the normal `view distance` setting).

> Second, I think the (already quite bloated) settings system should be made simpler when possible, and might want to avoid adding cases where a setting is used in one circumstance but not used in another circumstance.

Adding a new option isn't really complicating the settings system. It's just a new option that is read as needed, like any other option. It's what the system is designed to do. I actually think having `view distance` pull double-duty (where it has one target range when distant terrain is off, and another target range with distant terrain on) adds more complication, since the appropriate values for settings would change depending on the value of other settings.

> So whenever I want to change my view distance I need to remember whether I had distant terrain enabled or not, and find the correct setting to change.

I'd expect that if you have a particular view distance in mind to set, you'll know whether it involves distant terrain or not. Not much reason to use far distances without distant terrain, nor a reason to use near distances with it (of course, there's nothing stopping someone from doing it, but it's an atypical case so they should know what they're doing). Otherwise, if I want to turn distant terrain off or on, I have to remember if I need to change the view distance too, and remember what a good view distance is for it being off or on.

> Third, because distant terrain is still very much experimental and we don't know how many users are using it, it's probably best to not make it cross-contaminate other settings as that will again make it harder to retain backward compatibility if distant terrain / statics / fog / etc is ever overhauled or replaced by something else.

Isn't that a good reason to use a different setting? So if distant terrain is overhauled or replaced, they haven't contaminated the normal `view distance` setting with a large value and can simply turn distant terrain off to put it back to a normal state.

scrawl:
> Distant terrain is still a feature-in-progress, so users should expect things about it to change. 

Personally, when I use software I really despise seeing hardly necessary changes. I even get riled up by stuff like buttons moving elsewhere. Having a setting that I myself configured to a certain value be reset to something else with no indication of why or where to look to restore it would absolutely fall into that category. 

> Plus, I doubt the view distance most users currently have set for distant terrain is in any way proper, just random large values they made up that seems to work or copied from someone else who made something up that seems to work, so providing a clean break with an official default setting would help get everyone on the same page.

I think it's a bit of a stretch to say that that value would be better than the values people arrived at through testing and personal preference, and I don't see the point of getting everyone's settings on the same page.

> Not much reason to use far distances without distant terrain, nor a reason to use near distances with it (of course, there's nothing stopping someone from doing it, but it's an atypical case so they should know what they're doing).

On the other hand there is me coming from a developer perspective often not caring about whether distant terrain is on or not because I don't look at the view but to test specific code. For that reason I might leave distant terrain on with a low view distance or vice versa.

If you edit the settings file yourself, yes you should know what you're doing. That's what we're aiming for. Other people use the new shiny launcher menu.

When I said that distant terrain could be removed or replaced, part of what I meant is that having two code paths are not optimal. The non-distant terrain is kinda dumb because it doesn't do any LOD, but the distant terrain is dumb in that it loads stuff even if it's far beyond the view distance. If we can fix the latter thing, we could probably remove the distant terrain switch altogether and have it always on. In that case, if we actually had the separate view distance setting, we'd be looking at another config file break and a dead setting that will remain in user's configs to be mistakenly edited. Does the concern of contaminating settings make more sense with that in mind?

kcat:
> I think it's a bit of a stretch to say that that value would be better than the values people arrived at through testing and personal preference

How many people have done testing with personal tweaking? Rather than just bumping it up to 666666 to see really far because the default max was 6666, saw that it worked, and just left it there?

But alright. I still stand by what I said about a separate distant viewing option, but for now I've reverted it to always using the regular view distance option for the far clip plane. I've also rebased to current master. I don't know what wording you would prefer for the option descriptions regarding the view distance with distant terrain though, suggesting 40960 as a decent distant view distance (5 cells, also the max fog distance in clear weather with distant fog).

scrawl:
> +        bool mDistantFog : 1;
>+        bool mDistantTerrain : 1;
>+        bool mFieldOfViewOverridden : 1;

What's the point of initializing values in the header if the constructor sets them to a different value later?  Also, 'mDistantTerrain' seems unneeded now.

kcat:
Those aren't initialization, they're width specifiers. It packs the bools together into one bit each so they all fit into a single byte, rather than wasting 8 (or 32) bits per on/off flag. It's essentially a simple way to do bitfields (the compiler automatically applies the appropriate masks and shifts when accessing the individual member).

scrawl:
Huh, TIL. Then my question changes to this. That class is instantiated exactly once, so all we're doing is adding a few bytes to the code to save about 3 bytes in the executable. Is that worth it?

If I might suggest a place where this may actually be worth the time, there are a whole bunch of bools in MMechanics::NpcStats and that class instantiates thousands of time during a game.

Why can't the compiler do that packing by itself, btw? Is it because 'bool's don't really exist and are just a fancy word for byte?

kcat:
> That class is instantiated exactly once, so all we're doing is adding a few bytes to the code to save about 3 bytes in the executable. Is that worth it?

On its own, probably not. But it's a good practice to get into for places where it is more worth it. Generally speaking, using some cheap op codes (like masks and shifts) to save on memory throughput is a performance benefit, since cache is at a premium and pulling in data from RAM is rather slow compared to how fast data can be processed.

> If I might suggest a place where this may actually be worth the time, there are a whole bunch of bools in MMechanics::NpcStats and that class instantiates thousands of time during a game.

If I get the time, I could take a look through the code to see where some flags can be packed better.

> Why can't the compiler do that packing by itself, btw? Is it because 'bool's don't really exist and are just a fancy word for byte?

Sort of. A type's default storage has to be a multiple of a byte, even if it doesn't utilize all the bits or all combination of bits. True and false are the only valid values you can store in a bool (trying to store anything else, like a pointer or integer, implies a conversion to true or false), but its default size needs to be some multiple of a byte. So if you define `n` bools, it will take some multiple of `n` bytes, as each one is a unique object. The width specifiers are a way around that by packing POD types together, assigning a specific number of bits to each one, but the compiler can't do that automatically because it places additional restrictions on them (you can't take the address of such members, for instance).

PR #1477, opened 2017-09-24
AiWander: move back only after combat or pursue

akortunov:
Fixes [bug #3997](https://bugs.openmw.org/issues/3997), [bug #4251](https://bugs.openmw.org/issues/4251) and [bug #4432](https://bugs.openmw.org/issues/4432).

A basic idea: when the wandering actor (or actor without AI packages) starts combat or pursuit, I create the AiTravel package with his current position. After combat/pursuit this AiTravel package will return the actor to this position.

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -59,7 +59,7 @@ namespace MWMechanics
         float mTargetAngleRadians;
         bool mTurnActorGivingGreetingToFacePlayer;
         float mReaction; // update some actions infrequently
-        
+
         AiWander::GreetingState mSaidGreeting;
         int mGreetingTimer;
 
@@ -70,7 +70,7 @@ namespace MWMechanics
 
         bool mIsWanderingManually;
         bool mCanWanderAlongPathGrid;
-        
+
         unsigned short mIdleAnimation;
         std::vector<unsigned short> mBadIdles; // Idle animations that when called cause errors
 
@@ -86,7 +86,7 @@ namespace MWMechanics
 
         float mDoorCheckDuration;
         int mStuckCount;
-        
+
         AiWanderStorage():
             mTargetAngleRadians(0),
             mTurnActorGivingGreetingToFacePlayer(false),
@@ -111,7 +111,7 @@ namespace MWMechanics
             mIsWanderingManually = isManualWander;
         }
     };
-    
+
     AiWander::AiWander(int distance, int duration, int timeOfDay, const std::vector<unsigned char>& idle, bool repeat):
         mDistance(distance), mDuration(duration), mRemainingDuration(duration), mTimeOfDay(timeOfDay), mIdle(idle),
         mRepeat(repeat), mStoredInitialActorPosition(false), mInitialActorPosition(osg::Vec3f(0, 0, 0)), mHasDestination(false), mDestination(osg::Vec3f(0, 0, 0))
@@ -201,7 +201,18 @@ namespace MWMechanics
             stopWalking(actor, storage);
             currentCell = actor.getCell();
             storage.mPopulateAvailableNodes = true;
-            mStoredInitialActorPosition = false;
+        }
+
+        // Here we should reset an initial position, if a current cell was REALLY changed
+        // We do not store AiStorage in a savegame, so cellChange is not help us in this case
+        // TODO: find a more simple and fast solution, or do not store the mInitialActorPosition at all
+        if (mStoredInitialActorPosition)
+        {
+            int cx,cy;
+            MWBase::Environment::get().getWorld()->positionToIndex(mInitialActorPosition.x(),mInitialActorPosition.y(),cx,cy);
+            MWWorld::CellStore* cell = MWBase::Environment::get().getWorld()->getExterior(cx,cy);
+            if (cell != currentCell)
+                mStoredInitialActorPosition = false;
         }
 
         mRemainingDuration -= ((duration*MWBase::Environment::get().getWorld()->getTimeScaleFactor()) / 3600);
@@ -223,7 +234,7 @@ namespace MWMechanics
             if (mPathFinder.isPathConstructed())
                 storage.setState(Wander_Walking);
         }
-        
+
         doPerFrameActionsForState(actor, duration, storage, pos);
 
         playIdleDialogueRandomly(actor);
@@ -298,13 +309,6 @@ namespace MWMechanics
         if(mDistance && cellChange)
             mDistance = 0;
 
-        // For stationary NPCs, move back to the starting location if another AiPackage moved us elsewhere
-        if (mDistance == 0 && !cellChange
-            && (pos.asVec3() - mInitialActorPosition).length2() > (DESTINATION_TOLERANCE * DESTINATION_TOLERANCE))
-        {
-            returnToStartLocation(actor, storage, pos);
-        }
-
         // Allow interrupting a walking actor to trigger a greeting
         WanderState& wanderState = storage.mState;
         if ((wanderState == Wander_IdleNow) || (wanderState == Wander_Walking))
@@ -321,7 +325,7 @@ namespace MWMechanics
                 {
                     setPathToAnAllowedNode(actor, storage, pos);
                 }
-            } 
+            }
         } else if (storage.mIsWanderingManually && mPathFinder.checkPathCompleted(pos.pos[0], pos.pos[1], DESTINATION_TOLERANCE)) {
             completeManualWalking(actor, storage);
         }
@@ -330,8 +334,8 @@ namespace MWMechanics
     }
 
     bool AiWander::getRepeat() const
-    { 
-         return mRepeat; 
+    {
+         return mRepeat;
     }
 
 
@@ -348,27 +352,6 @@ namespace MWMechanics
         }
         // if get here, not yet completed
         return false;
-    }
-
-    void AiWander::returnToStartLocation(const MWWorld::Ptr& actor, AiWanderStorage& storage, ESM::Position& pos)
-    {
-        if (!mPathFinder.isPathConstructed())
-        {
-            mDestination = mInitialActorPosition;
-            ESM::Pathgrid::Point dest(PathFinder::MakePathgridPoint(mDestination));
-
-            // actor position is already in world coordinates
-            ESM::Pathgrid::Point start(PathFinder::MakePathgridPoint(pos));
-
-            // don't take shortcuts for wandering
-            mPathFinder.buildSyncedPath(start, dest, actor.getCell(), getPathGridGraph(actor.getCell()));
-
-            if (mPathFinder.isPathConstructed())
-            {
-                storage.setState(Wander_Walking);
-                mHasDestination = true;
-            }
-        }
     }
 
     /*
@@ -497,7 +480,7 @@ namespace MWMechanics
         }
     }
 
-    void AiWander::onWalkingStatePerFrameActions(const MWWorld::Ptr& actor, 
+    void AiWander::onWalkingStatePerFrameActions(const MWWorld::Ptr& actor,
         float duration, AiWanderStorage& storage, ESM::Position& pos)
     {
         // Is there no destination or are we there yet?
@@ -873,7 +856,7 @@ namespace MWMechanics
 
         state.moveIn(new AiWanderStorage());
 
-        MWBase::Environment::get().getWorld()->moveObject(actor, static_cast<float>(dest.mX), 
+        MWBase::Environment::get().getWorld()->moveObject(actor, static_cast<float>(dest.mX),
             static_cast<float>(dest.mY), static_cast<float>(dest.mZ));
         actor.getClass().adjustPosition(actor, false);
     }
@@ -914,7 +897,7 @@ namespace MWMechanics
             // get NPC's position in local (i.e. cell) coordinates
             osg::Vec3f npcPos(mInitialActorPosition);
             CoordinateConverter(cell).toLocal(npcPos);
-            
+
             // Find closest pathgrid point
             int closestPointIndex = PathFinder::GetClosestPoint(pathgrid, npcPos);
 
@@ -945,7 +928,7 @@ namespace MWMechanics
         storage.mPopulateAvailableNodes = false;
     }
 
-    // When only one path grid point in wander distance, 
+    // When only one path grid point in wander distance,
     // additional points for NPC to wander to are:
     // 1. NPC's initial location
     // 2. Partway along the path between the point and its connected points.
@@ -969,7 +952,7 @@ namespace MWMechanics
         delta.normalize();
 
         int distance = std::max(mDistance / 2, MINIMUM_WANDER_DISTANCE);
-        
+
         // must not travel longer than distance between waypoints or NPC goes past waypoint
         distance = std::min(distance, static_cast<int>(length));
         delta *= distance;
@@ -1041,4 +1024,3 @@ namespace MWMechanics
         init();
     }
 }
-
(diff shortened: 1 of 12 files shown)


apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -200,7 +200,18 @@ namespace MWMechanics
             stopWalking(actor, storage);
             currentCell = actor.getCell();
             storage.mPopulateAvailableNodes = true;
-            mStoredInitialActorPosition = false;
```
akortunov: We do not store an AiStorage in a savegame, and I can determine an old cell only by mInitialActorPosition after save load.
Actually, I have never seen cellChange == true during my testing (a batch of console commands from [this](https://bugs.openmw.org/issues/4010) topic):
First execute(): cellChange is false, because there is no currentCell in the AiState yet.
Second execute(): cellChange is false, because actor.getCell() == currentCell already.
Any ideas how to check if a current cell was REALLY changed after last update?

scrawl:
Might I suggest taking a step back and question why the code is even there? IIRC the 'return to start position' was mainly added to get actors to return after a fight. If that's all we need it for, maybe the code should be restricted so it only runs in that case.

scrawl:
Regarding the cellChange stuff, I honestly have no idea why that is there.

akortunov:
> If that's all we need it for, maybe the code should be restricted so it only runs in that case.

And how can we catch "after fight" moment? In AiCombat::execute()?

scrawl:
I have no idea. I'm sure there are several options.

akortunov:
> maybe the code should be restricted so it only runs in that case.

Ok, I managed to do this.
A basic idea: when the wandering actor starts combat, I create the AiTravel package with his current position. After combat this AiTravel package will return actor to this previous position.

Main problems:
1. I had to modify the AiTravel to do not cancel previous packages. If this behaviour is intended, I may make my changes optional.
2. I still have to keep the mInitialActorPosition for wanderNearStart().

Related issues:
1. [Carnius Magius](https://bugs.openmw.org/issues/4010)
2. [Almalexia](https://bugs.openmw.org/issues/3997)
3. [Korst Wind-Eye](https://forum.openmw.org/viewtopic.php?f=2&t=4766&p=51165)

scrawl:
> I create the AiTravel package with his current position. After combat this AiTravel package will return actor to this previous position.

That's an interesting idea.

> Main problems:
> 1. I had to modify the AiTravel to do not cancel previous packages. If this behaviour is intended,

Yes, it's intended. Every time an AI package is called from the script/console the previous package is cancelled. I guess we could move this behavior from AiSequence to the script functions.

akortunov:
>  I guess we could move this behavior from AiSequence to the script functions.

I do not like this idea much: a lot of chenges, a lot of places to break something.
For now I added the parameter to AiSequence::stack() to control if we need to cancel other packages, so I did not touch the default behaviour.

scrawl:
> I create the AiTravel package with his current position. After combat this AiTravel package will return actor to this previous position.
That's an interesting idea.

> Main problems:
> 1. I had to modify the AiTravel to do not cancel previous packages. If this behaviour is intended,

Yes, it is. Every time an AI package is called from the script/console the previous package is cancelled. I guess we could move this behavior from AiSequence to the script functions.

apps/openmw/mwmechanics/aiwander.cpp
```diff
@@ -201,7 +201,18 @@ namespace MWMechanics
...
+        }
+
+        // Here we should reset an initial position, if a current cell was REALLY changed
+        // We do not store AiStorage in a savegame, so cellChange is not help us in this case
+        // TODO: find a more simple and fast solution, or do not store the mInitialActorPosition at all
+        if (mStoredInitialActorPosition)
```
scrawl: Is this still needed?

akortunov: > Is this still needed?

For now the initial position is still used in the wanderNearStart() function.

akortunov:
Hey, welcome back!

> Has there been a regression?

Very probably. I can confirm this issue.


So how this feature behaves in vanilla game:
1. NPCs return only after combat: NPC can be teleported via console (ra, setpos, etc.) and will not return to the initial position. Instead he will return to the position, where he was before combat.
2. If I start combat, then reset the NPC AiWander package, NPC still returns to the position before combat.

I can guess that "Return to the initial position" is not a part AiWander package, but a hack from combat mode.

akortunov:
An another issue, with AiTravel package itself: looks like if you teleport the NPC from exterior to interior or back, the target coordinates will be wrong. In this case we should freeze or cancel AiTravel package.

Allofich:
> Hey, welcome back!
Thanks.

> NPCs return only after combat: NPC can be teleported via console (ra, setpos, etc.) and will not return to the initial position. Instead he will return to the position, where he was before combat.

I can confirm for a still NPC (caravaner in Seyda Neen), who I think has Wander of 0 distance, but not for two other NPCs I tried who move around with wander of more than 0 distance. I tested this when I worked on a PR for this earlier, and again now, and I think the behavior is that after combat the NPC will resume walking to whatever its destination was before combat, except with 0 distance. For 0 distance, the NPC returns to whatever spot it was at before combat started.

Test case:
Give NPC AIWander 2000 0 0 0. They start to walk to a distant spot.
While they are on the way, use startcombat player.
Lead them out of the way somewhere.
Either stopcombat (they resume walking to their _destination,_ not to where they were before combat) or AIWander 0 0 0 0 (they walk to the spot they were at before combat, and this is their new location to return to)

Allofich:
A guess:
Maybe in the original game, every Wander move saves both the location for the start of a Wander move, and the destination. Normally the start of the move is unused? but if the "distance" of the package is 0, desination position is set to equal the start position, perhaps to skip over code to calculate a destination.

akortunov:
An interesting thing: NPCs without AI packages return to their before-combat position too.

1. AiTravel 0 0 0
2. getcurrentaipackage (shows -1)
3. startcombat
4. stopcombat

Allofich:
You're right, and while returning, it still gives -1 for the AI package.

I did another test:
1. AiTravel 0 0 0
2. getcurrentaipackage (shows -1)
3. startcombat
4. AIWander 0 0 0 0
NPC returns returns to the pre-combat spot as their wander location.
5. AITravel 0 0 0 (reset to no AI package)
6. startcombat (make the NPC move)
7. AITravel 0 0 0 (reset to no AI package in new location)
8. startcombat
9. AIWander 0 0 0 0
The AI returns to the location from step 7.

Allofich:
A little more testing. Same behavior for returning with no AI package (AITravel 0 0 0, startcombat, stopcombat) as with AIWander 0 0 0 0. The NPC returns to whatever its last location was, including if the location is changed.

I guess combat code must store the location the NPC was in at the beginning of combat. There was some code like this before, but I removed it in a PR a while ago about this issue. I thought I could remove that code and just use returning to the AIWander package's destination, but it looks like if the NPC does not have an AIWander package with distance > 0, then the pre-combat location is used to return to instead.

Allofich:
Here you can see the code I removed in aisequence.cpp. It might be useful again.
https://github.com/OpenMW/openmw/commit/4f08084e7921603e2b590a645b1dbaa8d0777d87#diff-daf7c11ce0cfe23a9a8ba750df987bec

akortunov:
So what we have:
1. "Return to initial position" is handled outside AiWander.
2. AiWander with Distance > 0 and current destination should move to that destination, in other cases NPC should move to the position before combat.
3. Looks like we can not store that data in the AiCombat - we should close this package immediately after combat.

So what we can to do:
1. Create the AiTravel package if the NPC is wandering or does not have AI packages.
2. "Hide" created AiTravel (getcurrentaipackage should not show it)
3. If NPC is wandering before combat, we should read the AiWander package destination and use it as AiTravel destination. If the NPCS does not have the destination, we should use his current position.

As for me, result will be pretty close to vanilla.

Allofich:
> Create the AiTravel package if the NPC is wandering or does not have AI packages.
If "not" wandering, right? Yes, I agree this solution (hiding an AI travel, etc.) sounds like it will replicate vanilla behavior.

Allofich:
Another bit of information. I tried saving a game in the original engine when an NPC I have done either
1) AIWander 0 0 0 0
startcombat player

or

2) AITravel 0 0 0
Startcombat player

and saved the game while the NPC is attacking me. In both cases, the NPC is no longer in combat with me when the game loads (bug?) but the interesting part is that for case 1 the NPC returns to their pre-combat location but in case 2, they do not, they stay in the spot they were at when the game is loaded.

It probably doesn't matter, but it suggests that in the AIWander with distance 0 case, data from the AIWander package is accessed, but perhaps only if it is available. I think the solution you said is probably fine, though. This is just extra information.

akortunov:
> "npc->getCurrentAIPackage" as the NPC is returning to their location returns 0 (Wander package).
Do the same in OpenMW with this PR, and getCurrentAIPackage returns 1 (Travel package).

I guess the only thing left.
We can, of cource, store a some kind of flag, e.g. isHidded for internal packages, but I do not want to touch a savegame format. Any ideas?

Allofich:
> but I do not want to touch a savegame format. 

Hmm., I think it will be impossible to fully replicate original behavior without doing that, since we need the position before combat somehow... If we can figure out how the .ess files store this data, since it seems to be in there somewhere, maybe we can rework the solution to that, and also update the save file format for it. If we don't change the save file format, though, I guess we're stuck with a less-than-perfect solution, and npc->getCurrentAIPackage will return 1 after loading a game. It probably would be very rare for it to cause a problem, at least.

Allofich:
I wonder if it would be possible to use an actor's position in liveCellRef? This seems to be the value referred to by the "resetAllActors" command.

Is it necessary to know the initial position of an actor for anything except resetAllActors? What I thought of:
- Repurpose the "mPos" in liveCellRef to use as an AI's "return-to" location for the behaviors we have described here. This could maybe replace "mInitialActorPosition" in AIWander as well.
- Change "resetAllActors" to load the position directly from the .esm/.esp data.
- When making an actor return, use a method that will not save an AITravel package to the save file. Either use the mHidden field and a mDontSave field, or maybe make it a new AIReturn package with those properties.
- Actors check in an update loop somewhere whether they are away from their mPos and don't have an overriding behavior (AIWander with distance >0, combat package, AIFollow, etc.)

I don't know if it's OK to change mPos like this, and I'm sure the things I suggested would be a lot of work, but is this a possible solution?

Allofich:
Also maybe anything in AIWander involving a 0-distance package can be replaced with common code for having no package? It depends if there are any differences in behavior. I don't know of any.

Capostrophic:
What's the state of this?

akortunov:
> What's the state of this?

Basically it works. The only known issue that we do not store the mHidden flag in savegame, so getCurrentAipackage can behave differently from vanilla, if there was save/load afterstart of combat.

Allofich:
> Basically it works. The only known issue that we do not store the mHidden flag in savegame, so getCurrentAipackage can behave differently from vanilla, if there was save/load afterstart of combat.

I wondered in the comment above if we could use `mPos` to as the saved information for the return point (and therefore avoid saving a package with the new flag, just reapply it anew on loading the game), but I'm not sure that `mPos` is actually saved and loaded (or if it is OK to change it for this purpose)? So that approach might cause a change to the saved game format as well.

It seems like a problem to me, though, if saved game format concerns are getting in the way of properly fixing problems, or holding up a PR such as this (if that is indeed the reason). Is it not possible to get a more robust save/load system where save records can be added, and maybe even then removed or adjusted, without causing load errors? As for forwards compatibility of saves, which scrawl mentioned in another PR, that seems like too much to me. I'm not keeping up with all the latest software but I had never even heard of that before, only backwards compatibility.

akortunov:
> Is it not possible to get a more robust save/load system where save records can be added, and maybe even then removed or adjusted, without causing load errors?

According to community testing, Morrowind just ignores unknown records in the savegame and prints warnings instead of cancel the whole savegame loading.

> I wondered in the comment above if we could use mPos to as the saved information for the return point 

What the mPos supposed to store? Initial coordinates from ESM/ESP file? In this case it is not for us - we need to return actor to his position before combat, not to his spawn position.

Allofich:
> What the mPos supposed to store? Initial coordinates from ESM/ESP file? In this case it is not for us - we need to return actor to his position before combat, not to his spawn position.

My thinking was that we could get that information again from the ESP file. But I didn't think about NPCs placed through script (Although I don't know if they use mPos).

Anyway, that was just an idea I had when looking for some data we could use without adding to the save file. It's probably not the best idea.

akortunov:
I suppose this feature will be cleaner a bit if I create a new AiReturn package (inherit from AiTravel) instead if messing with mHidden field.

Greendogo:
Does this work? I was going to test it out, but saw that it wasn't merged yet.

akortunov:
> Does this work? I was going to test it out, but saw that it wasn't merged yet.

It worked last time I checked.

psi29a:
@zinnschlag this has been sitting in the wings for awhile, what are your thoughts?

zinnschlag:
The flag not saved to the saved game file seems to be an issue.

Note that contrary to what is stated above we have the framework in place for handling extensions to the saved game format. There is a version number that is checked ( ESM::SavedGame::sCurrentFormat), resulting in a proper rejection of newer format files loaded into older versions of OpenMW.
For adding new fields we would increase this value by one. We have done that successfully in the past.

Once this issue is resolved I am okay with merging.

akortunov:
> resulting in a proper rejection of newer format files loaded into older versions of OpenMW.

So newer saves will not work if player will swith back from nightlies to latest stable release or will try to do bisect?
IIRC, vanilla game in this case just skips unknown records and prints warning to player.

scrawl suggested an another approach to modify savegame format:
1. Add ability to read new tags
2. After next stable release start to write them to files

zinnschlag:
Not in favour of this approach. In the long run it isn't feasible anyway. While we will retain the ability to read old saves, there will be substantial changes to the format after 1.0 that will break compatibility. That is unavoidable (because of the planned de-hardcoding). Simple ignoring the difference in the file format when loading in older OpenMW versions would eventually lead to failed loads. That is more undesirable than limiting the compatibility of saved game files in one direction.

akortunov:
A first arrempt to store the flag in savegame.

I encountered an another bug in upstream master: we do not store mLastAiPackage in savegame, so getCurrentAiPackage() returns -1 if there were no **actual** AI packages executed after save load.
You can encounter this issue if the actor is fighting or pursuing someone upon save load.

akortunov:
Figured out that wandering actors and actors without AI packages should return after pursuit too.

akortunov:
`MSBUILD : error MSB4017: The build stopped unexpectedly because of an unexpected logger failure.`

A quite strange error.

psi29a:
Not the first time, I kicked it in the head to restart the build.

akortunov:
> Not the first time, I kicked it in the head to restart the build.

I wonder if it caused by my recent CI changes: we do not use unity builds now and build sub-projects.
If build is parallel, probably there is a some kind of deadlock.

psi29a:
So far it is only with msvc2015, this is the 2nd time in a week. Haven't seen this issue with msvc2017. We'll keep track of it, but not a deal breaker by any means.

zinnschlag:
The save code looks okay to me. I recently saw another PR that will also require an increase in the saved game version. @akortunov That was your's too, right? If so, we should try to merge them both at the same time, so we don't have to change the format twice.

zinnschlag:
Found it: https://github.com/OpenMW/openmw/pull/1559

In this PR (#1477) the changelog changes are still missing. And the CI checks on the other PR haven't run yet. Once that has been taken care of, we can have a merge.

akortunov:
> In this PR (#1477) the changelog changes are still missing.

Added it.

PR #1478, opened 2017-09-25
Add an option to use third person animations in first person view (Feature #2666)

Capostrophic:
Implements [feature #2666](https://bugs.openmw.org/issues/2666) based on scrawl's 8 lines. I've added a configuration setting, by default the behavior is (of course) not changed.

### DIFF
--- a/apps/openmw/mwrender/camera.cpp
+++ b/apps/openmw/mwrender/camera.cpp
@@ -1,8 +1,11 @@
 #include "camera.hpp"
 
 #include <osg/Camera>
+#include <osg/PositionAttitudeTransform>
 
 #include <components/sceneutil/positionattitudetransform.hpp>
+
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/windowmanager.hpp"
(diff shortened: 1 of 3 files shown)


akortunov:
An interesting feature, but non-ready for gameplay:
1. Camera moves repeatedly when player is breathing.
2. A camera is too high to show short weapons (e.g. daggers, probes) and player hands during spellcasting.
3. A shooting looks really strange (especially crossbow reload animation).

I wonder if we can solve these issues (on resource side or on engine side).

Capostrophic:
Well, this is just a generic solution which acts just like "immersive first person view" mod for Skyrim and some kind of "realistic camera" from Oblivion Reloaded mod. All of these points are unavoidable with third person animations used instead of first person animations. Maybe I should add a warning or something. 
Modifying animations is definitely viable, however there is no differentiation between first person and third person animations.

Enhanced Camera OBSE/FOSE/NVSE/SKSE plugin manages to use first person hands with third person body so there is no gameplay impact, but the implementation seems to be too complicated for such a minor feature.

scrawl:
I think it's a no from me, not in this form anyway. We don't add options that are half broken. People will inevitably complain, no matter how many warnings or whatever we put next to the setting.

IIRC, the camera offset didn't work properly for all races. Even if it did, it shouldn't be hardcoded. Adding another setting for the offset? Eh.. that's just more bloat.

If we want to continue exploring this, I'd suggest a tag in the animation file for the engine to use or something like that, this feature IMO goes in the realm of modding not a simple on/off switch.

PR #1480, opened 2017-09-26
Bug 1875-  NPC/Creatures should regenerate when resting.

Langerz82:
Let me know if anything is missing.


scrawl:
This issue has absolutely nothing to do with the savegame format. I have no idea why you change it. Also there are many more unnecessary and broken changes. Best to start over from scratch.

Langerz82:
It saves the hurt creatures/NPC because in my understanding the cells get unloaded once a player leaves it and it needs to keep track of the ones that are hurt or fatigued. When the player enters an active cell it retrieves those affected. If it did not save them a player could save then re-load and it would drop the creatures in the list thereby exploiting the bug still.

PR #1481, opened 2017-09-27
 WindowManager overhaul & improved keyboard support

scrawl:
This long-overdue refactoring of the WindowManager makes <s>me no longer want to kill myself whenever I look at it</s> it more concise and flexible. Code duplication was removed and replaced by more generic window management. Basic keyboard navigation is now supported, which was the main motivator that led to the refactoring. Some inconsistencies in the GUI were fixed.

Related discussion: [pull request for container hotkeys](https://github.com/OpenMW/openmw/pull/1443)

# Keyboard navigation

<i>If you want to test the PR, I suggest doing so before you read this 'how it works' section, so we can get feedback on how intuitive the new behavior is.</i>

The new support for keyboard navigation intends to bring our GUI more in line with typical GUIs and improve usability. Here's how it works:

- The button under key focus is visually highlighted (just like when it's hovered with the mouse).
- (`Shift`) `Tab` cycles through buttons.
- `Arrow` keys or movement buttons (default `WASD`) focus the next button in that direction.
- `Space`, (keypad)`Enter` and the key bound to 'Activate' trigger the focused button (I'll refer to these keys as `Accept`).
- The focused button is remembered for each GUI mode.
- Several windows are now more keyboard friendly and focus a button by default.
- Repeat key events are accepted by the GUI (i.e. holding down a key will, after an initial delay, repeatedly produce events).
- When entering GUI mode, the mouse cursor is initially hidden until you use it (to prevent it from being in the way). 
- The keypad can now be used to rotate the camera. This binding is not visible in the controls menu and will not be used if the keypad is already bound to something else.

## Use examples
- From the main menu, `Accept` `Accept` loads your last game. Another `Accept` skips the plugins changed warning.
- Read through books, scrolls and journals with the `UpArrow`/`DownArrow` keys.
- Make a bazillion potions by holding down `Accept` in the alchemy screen.
- Find the optimal haggle price using `UpArrow`, `DownArrow` and `Enter`.
- Spamming persuasion attempts is now just spamming `Accept`, no need to line up the two windows exactly and then spam mouse clicks.
- Loot (or check without looting) multiple containers faster with `Activate` `Activate`.

## Potential issues
- One could argue the more efficient controls break the game balance (e.g. by making the 'broken' alchemy system even more broken). If this is an issue, we can address it by limiting the key repeat speed.
- Users could accidentally dismiss message boxes before reading them. I doubt this is much of an issue though (most of the dialogs are predictable, but even if there's an unexpected dialog, it's still unlikely for the user to be pressing space/enter for no reason, or 'spam' activate on something).
- Accidentally deleting savegames. I attempted to mitigate this by making it not possible to focus the 'Delete' button with Tab, instead you have to press `Delete` key on the keyboard. If there are other similarly destructive buttons, we might want to treat them this way too.
- Possible regressions caused by refactoring. Fixed all that I noticed, there are probably some more.

## Future work

The gamepad mode could be updated to use the button navigation - should be a great improvement over the current 'axis controls mouse' mode.

Not everything is fully keyboard-accessible yet (mainly dialogue topics and the settings window), and more windows have not been tested. I hope we can get help from the community with filling in some missing pieces as they are discovered during play. Here's how to fix the most simple issues:

* Add `<Property key="NeedKey" value="true"/>` to the skin of a button to make it visible to the key navigation. 
* To set a default button, add `MWBase::Environment::get().getWindowManager()->setKeyFocusWidget(the_button);` somewhere in the window's `onOpen` or `setPtr` methods.

# Solved bug/feature tickets

- Menu closes instantly ([3978](https://bugs.openmw.org/issues/3978))
- Unable to close menu during drag and drop ([1692](https://bugs.openmw.org/issues/1692), [3099](https://bugs.openmw.org/issues/3099))
- Check if dragged item is deleted ([3097](https://bugs.openmw.org/issues/3097))
- Add keyboard shortcuts for menus ([3400](https://bugs.openmw.org/issues/3400))
- Container closing doesn't work after touching an item ([1555](https://bugs.openmw.org/issues/1555))
- Closing menus with spacebar will still send certain assigned actions through afterwards ([4016](https://bugs.openmw.org/issues/4016))
- Reject 'MCP keys' ([4092](https://bugs.openmw.org/issues/4092))


MiroslavR:
I like it, nice job. One thing I have noticed is that the focused button in the "plugins changed" dialog is not highlighted.

kcat:
> (```Shift```) ```Tab``` cycles through buttons.

Could this cause issues if the user has Tab bound to open/close the inventory? The later games use Tab for that by default, and for someone coming from those games could prefer that over right-click.

scrawl:
> Could this cause issues if the user has Tab bound to open/close the inventory? The later games use Tab for that by default, and for someone coming from those games could prefer that over right-click.

The precedence of hotkeys/bindings works líke this: keypress is injected to the GUI; if no widget accepts it, the keypress is then passed to the controls system. So opening the inventory should work. But closing won't work, I think, because the World Map button accepts the tab key. This could be quite easily fixed by removing key focus for the 'World Map' button. To be honest, keyboard navigation for the inventory is kinda difficult anyway, with it being 4 different windows.

MiroslavR:
> One thing I have noticed is that the focused button in the "plugins changed" dialog is not highlighted.

Another thing: The level-up dialog allows attribute value textboxes to be tabbed through. I assume that's because their skin is set to "SandTextButton".

akortunov:
A couple if issues found (tested with MyGUI 3.2.2):
1. You can close an interactive messagebox by pressing ESC button. In this case a most of in-game input will be frozen (an engine still awaits an answer from messagebox).
2. A cycling in the dialogue window is a bit strange.
When a player clicks on any topic or service, the label will be marked as focused. I'd expect one of these behaviours:
  a) The topic is focused, but I can use arrows to navigate in the topics list.
  b) A focus returns to the "Goodbye" button.

scrawl:
Fixed most of the mentioned issues and a few more I noticed.

>  A cycling in the dialogue window is a bit strange.
> When a player clicks on any topic or service, the label will be marked as focused. I'd expect one of these behaviours:
>   a) The topic is focused, but I can use arrows to navigate in the topics list.
>   b) A focus returns to the "Goodbye" button.

Right, it is kinda weird. Initially I had it behave like b), but then I wanted to be able to retain focus on the service menu items to do things like the repeated persuasion. I couldn't go all the way though because MWList class doesn't support key focusing properly. Currently, we have two classes for lists and none of them are that great: 

MWList:
- No keyboard navigation support
- API is built around the list being completely rebuilt with each update, this is bad as key focus will be lost
- Activates items by single click

MyGUI::ListBox:
- Has keyboard navigation
- No word-wrap support 
- Activates items by double click

We've also had issues with Disabled items not being able to have tooltips set (for both lists, I think).

I am not quite sure if it's better to extend MWList or to extend (or subclass) MyGUI::ListBox to add the things we need. While developing this feature I actually found about a half dozen issues in MyGUI that should probably be improved so I don't have to workaround them anymore. That further drives me to the 'try to fix MyGUI' path but I'm still not sure.

scrawl:
> A cycling in the dialogue window is a bit strange.

Changed to no longer retain focus for topics. That should reduce the confusion at least. Service items still retain focus to facilitate multiple uses of the same service.

kcat:
This seems to have broke the Travel option for NPCs. Also, it's a bit disconcerting to see two highlighted widgets, one for the keyboard input and one for the mouse. Shouldn't there be only one highlighted widget at a time? Either always follow the mouse and only allow the keyboard to cycle/change it when the mouse isn't moving over one (i.e. when the mouse is stationary or is moving over non-activatable widgets), or don't have the highlight follow the mouse cursor until the user clicks. Those seem to be the common UI interaction methods when mouse and keyboard are both usable.

akortunov:
Also "Take" button is broken in the book window: you can take book only via keyboard, not via mouse.

akortunov:
I am not sure, if it is related, but an engine does not update labels on local map after recent updates.
There are no labels after game start.

akortunov:
Also now I can talk with creatures, e.g. scribs.

akortunov:
Also companion window does not show player's inventory now.

scrawl:
Last two issues should be fixed.

> I am not sure, if it is related, but an engine does not update labels on local map after recent updates.
There are no labels after game start.

Can you be more specific please? What labels do you mean, door markers? On the map window or the HUD map?

I don't recall changing anything in local map code. Maybe there is some naming conflict with onFrame(dt) now being virtual? Not sure how though.

akortunov:
screenshot_20171004_085933

This is Balmora.

akortunov:
A game calls LocalMapBase::updateDoorMarkers(), creates marker widgets, but does not show them for some reason.

akortunov:
screenshot_20171004_102056

This is a Balmora, Guild of Mages.

akortunov:
"Detect X" spells markers are affected too.

scrawl:
For me the map was not updated when it was pinned. Opening Inventory Mode updates it. This is fixed now, along with the other issue (the crash on the bugtracker).

akortunov:
Just noticed that the training window behaviour is different from other service windows: when you press ESC, a main menu appears. But ESC closes other service windows.

scrawl:
Seems like I added that behavior in https://github.com/OpenMW/openmw/commit/843106fc616488ca174cd9d30d504254546fe631 and the intention was to not be able to skip the training progress bar. Probably this
<pre>
 bool exit() { return false; }
</pre>
should actually read
<pre>
return !mTimeAdvancer.isRunning()
</pre>
But I can't test right now. Feel free to make a PR if it works.

scrawl:
On the topic of escaping windows, do you think it would make sense to handle the Menu action (right click by default) as an alternative to Esc? Seems like that would be a lot more accessible. But are there any cases where accidentally exiting a window would be harmful?

akortunov:
> do you think it would make sense to handle the Menu action (right click by default) as an alternative to Esc?

Makes sense, as for me. But I suggest to create a forum topic first.

PR #1482, opened 2017-09-27
Inherit owner, if an item is in container (regression #4128)

akortunov:
Fixes [bug #4128](https://bugs.openmw.org/issues/4128).
In 0.42 isAllowedToUse() accepted ownerCellRef, in 0.43 accepts a pointer, so if an item is in container, we should provide a container pointer instead of item pointer.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -1012,14 +1012,17 @@ namespace MWMechanics
 
         MWWorld::Ptr victim;
 
+        bool isAllowed = true;
         const MWWorld::CellRef* ownerCellRef = &item.getCellRef();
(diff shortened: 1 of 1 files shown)


scrawl:
I think there's another issue in `ToolTips::checkOwned`. From what I can tell, it'll display items in owned containers as non-owned.

akortunov:
> I think there's another issue in ToolTips::checkOwned. From what I can tell, it'll display items in owned containers as non-owned.

That's because most of items in containers have no owner. You can check this behaviour in vanilla game via TFH console command. Only thing I can suggest here is to check if an items is stored in container in isAllowedToUse(). Have we an easy way to get a container pointer via item pointer?

scrawl:
Not directly, but there's a World::findContainer function or something, that works by searching through all active cells. I hope it won't be too slow.

PR #1483, opened 2017-09-28
water shader improvements plus rain ripple effect (#452)

drummyfish:
What's been done:

- Added rain induced water ripple effect in water fragment shader. Tested:
  - with and without refractions
  - above and under water
  - exteriors and interiors (the effect is turned off in interiors)
  - fading in and out when rain starts/stops
- Small water shader refactoring (moved copy/paste code into a function).
- Slightly decreased water transparency when refractions are turned off, as the water was too clear compared to "refraction version" and also the ripples were hardly noticeable.

[Here](https://www.youtube.com/watch?v=vJMNW58Yhy0&feature=youtu.be) is a new video preview (ripples with refraction, ripples without refraction and ripples gradually fading out as the rain stops).

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -251,6 +251,10 @@ namespace MWRender
         sceneRoot->setNodeMask(Mask_Scene);
         sceneRoot->setName("Scene Root");
 
+        mUniformRainIntensity = new osg::Uniform("rainIntensity",(float) 0.0);
+        
+        mRootNode->getOrCreateStateSet()->addUniform(mUniformRainIntensity);
+
         mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager()));
 
         source->setStateSetModes(*mRootNode->getOrCreateStateSet(), osg::StateAttribute::ON);
@@ -799,6 +803,7 @@ namespace MWRender
     {
         mEffectManager->clear();
         mWater->clearRipples();
+        mUniformRainIntensity->set((float) 0.0);  // for interiors
     }
 
     void RenderingManager::clear()
--- a/apps/openmw/mwrender/renderingmanager.hpp
+++ b/apps/openmw/mwrender/renderingmanager.hpp
@@ -85,6 +85,7 @@ namespace MWRender
 
         osg::Uniform* mUniformNear;
         osg::Uniform* mUniformFar;
+        osg::Uniform* mUniformRainIntensity;
 
         void preloadCommonAssets();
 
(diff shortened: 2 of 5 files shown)


akortunov:
Awesome work!
But I see an issue: ripples look fine on a short distance, but on a long distance (e.g. when you have a Distant Terrain enabled), ripples look like a white dots. If a player moves, it looks like a white noise, and it is tiring for player eyes.
Can we do not render ripples on a large distance?

drummyfish:
I didn't test distant terrain, thanks for the feedback, will try to fix this.

drummyfish:
@akortunov Could you please post a screenshot just so that I know what exactly you're talking about?

drummyfish:
Is it this issue right here?

ripplebug

These are caused by a little "trick" I used and will probably have to unuse again, turning ripples off at high distance won't be needed.

akortunov:
Sure: [screenshot](https://image.ibb.co/he6FrG/Screenshot_20170928_223114.jpg).

akortunov:
> Is it this issue right here?

Yes, it is.

akortunov:
Thanks. Works fine now.

akortunov:
But what if a water shader is disabled? Looks like there are ripples in this case.

drummyfish:
You can see the **rain** ripples with the shader disabled? If I disable the shader, I can see none, only the ripples caused by the player moving in the water, but these are different. Custom shader shouldn't be used for simple water at all, it's just a plane with animated texture.

akortunov:
> You can see the rain ripples with the shader disabled?

No. There was a typo in my last comment, which I fixed later.
Looks like a vanilla engine handles rain ripples in the same way, as ripples from swimming actors.

drummyfish:
Yes, this is a planned [feature](http://bugs.openmw.org/issues/3537), not implemented yet. The "swimming ripples" are now implemented as OSG emitters, not in the shader.

akortunov:
I can only suggest to use OSG emitters for both swimming and rain ripples, if water shader is disabled.
Otherwise use shaders.

drummyfish:
That would be possible but maybe bad for performance with so many emitters considering the simple water is supposed to be a "better performance" option. And it would probably not look very good either. I think that the simple water doesn't need rain ripples, but we can have a discussion about that of course, maybe in the bugtracker as a new feature request.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -251,6 +251,10 @@ namespace MWRender
         sceneRoot->setNodeMask(Mask_Scene);
         sceneRoot->setName("Scene Root");
 
+        mUniformRainIntensity = new osg::Uniform("rainIntensity",(float) 0.0);
+        
+        mRootNode->getOrCreateStateSet()->addUniform(mUniformRainIntensity);
```
scrawl: Could we look into setting the uniform on the water node rather than the root, so it doesn't clog the rest of the scene graph?

drummyfish: I'm getting back to this right when I'm done with the weather particles. I have to enable the water ripples for snow so I will look into this too.

PR #1487, opened 2017-10-01
[discussion] attempt at solving rain/snow moving with player (issue #2698)

drummyfish:
I've been trying to fix [bug #2698](https://bugs.openmw.org/issues/2698). **Do not merge** this please, there are bugs. I'm posting this to get some feedback.

I've done this:

- Reparented the rain emitter from sky node to its parent, so that the rain drops do not move with the player.
- Created a subclass of `BoxPlacer` (`RainPlacer`) to place particles relatively to camera position.

So now the emitter is always placed at [0,0] in the world and the rain particles are spawned relatively to the camera, but are not parented to it, so they move independently. Questions:

1. Is this a good idea? I had to increase the number of particles greatly in order to cover greater area than just the camera immediate surrounding. Is it OK to parent the rain node outside the sky subtree?
2. Could the emitter stay parented to the sky node, but ignore its position somehow? I am a newbie with OSG and don't know if this could be done. (I tried `emitter->setReferenceFrame(osgParticle::ParticleProcessor::ABSOLUTE_RF)`, didn't work).
3. Any ideas for different approaches?

Also snow is broken now and behaves weirdly, I didn't really look much into the reason yet, but I think it uses a different particle system object to which I'd have to apply the same approach.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -255,7 +255,7 @@ namespace MWRender
         
         mRootNode->getOrCreateStateSet()->addUniform(mUniformRainIntensity);
 
-        mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager()));
+        mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager(), this));
 
         source->setStateSetModes(*mRootNode->getOrCreateStateSet(), osg::StateAttribute::ON);
 
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -40,6 +40,7 @@
 #include <components/sceneutil/statesetupdater.hpp>
 #include <components/sceneutil/controller.hpp>
 #include <components/sceneutil/visitor.hpp>
+#include <components/nifosg/particle.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
@@ -1340,22 +1341,49 @@ protected:
     osg::Uniform* mRainIntensityUniform;
 };
 
-class RainPlacer : public osgParticle::BoxPlacer
+class CameraPlacerDecorator : public osgParticle::Placer
 {
   public:
-    RainPlacer(MWRender::RenderingManager *renderingManager): BoxPlacer()
+    CameraPlacerDecorator(osgParticle::Placer *decorated, MWRender::RenderingManager *renderingManager): Placer()
     {
         mRendering = renderingManager;
+        mDecorated = decorated;
     }
 
     virtual void place(osgParticle::Particle *P) const override
     {
-        BoxPlacer::place(P);
+        mDecorated->place(P);
         P->setPosition(P->getPosition() + mRendering->getCameraPosition());
+    }
+
+    virtual osg::Vec3 getControlPosition() const override
+    {
+        return mDecorated->getControlPosition();
+    }
+
+    virtual osg::Object* cloneType() const override
+    {
+        return mDecorated->cloneType();
+    }
+
+    virtual osg::Object* clone(const osg::CopyOp& op) const override
+    {
+        return mDecorated->clone(op);
+    }
+
+    osgParticle::Placer *getDecorated()
+    {
+        return mDecorated;
     }
 
   protected:
     MWRender::RenderingManager *mRendering;
+    osgParticle::Placer *mDecorated;
+
+    ~CameraPlacerDecorator()
+    {
+        mDecorated->unref();
+    }
 };
 
 void SkyManager::createRain()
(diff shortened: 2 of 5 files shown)


scrawl:
The Emitter and ParticleSystem can be placed under different nodes. To have particles emit from the camera but not move with it, you would put the Emitter node under the camera-relative node and the actual particle system above it. In theory, anyway. It's entirely possible this won't work because of how the camera-relative node is implemented, or shortcomings in osgParticle / the emitter class.

drummyfish:
Thanks for the suggestion, I'll try this out tomorrow.

drummyfish:
I'm experimenting with it and it doesn't seem to work, because the placer objects (used by the emitter) place the new particles relatively to a certain center point set with a function ([OSG source](https://github.com/openscenegraph/OpenSceneGraph/blob/4bc13207091e923bb5c57abba7278a57cfae46b4/include/osgParticle/BoxPlacer#L169)), i.e. the emitter doesn't operate on world coordinates. But I might create a custom placer that would place the particles according to its own world position, which would be similar to what I have done but more elegant, since I wouldn't have to pass the camera position to it.

drummyfish:
Now I'm looking into how for example the snow is created and it's a particle system loaded from a NIF file. So its placer will be defined in the file. So I think instead of creating my own placer I will create a decorator for the placer that will offset the created particles and I will simply decorate all the particle systems I need.

drummyfish:
Done, now even the snow works independently of the camera, but its emitter parameters aren't set correctly for this setup yet, I'll have to work some more on that.

Also are snow and rain the only particle systems this issue involves?

scrawl:
> I'm experimenting with it and it doesn't seem to work,  because the placer objects (used by the emitter) place the new particles  relatively to a certain center point set with a function (OSG source), i.e. the emitter doesn't operate on world coordinates.

The position given by a Placer is relative to the Emitter. So if the Emitter is placed under the camera, and the placer tells me to place a particle at (0,0,0) it should spawn right at the camera, regardless of where the Particle System is placed (emitter coordinates are converted to Particle System coordinates [here](https://github.com/OpenMW/openmw/blob/master/components/nifosg/particle.cpp#L307)). Again, this is only in theory and might be broken by some details of how the camera relative node is implemented. Not saying your solution is wrong but maybe we can give another look why it is that the conceptually 'clean' solution doesn't work.

scrawl:
Okay, I think I know what it is. The [camera relative node](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwrender/sky.cpp#L271) works by assuming a world-view matrix is passed to computeLocalToWorld (and this is the case during cull traversal), then sets the transform of that matrix to zero to essentially strip out the 'view' component. But the Emitter class [calls](https://github.com/OpenMW/openmw/blob/master/components/nifosg/particle.cpp#L270) computeLocalToWorld with a world matrix that doesn't include a view to begin with, so trying to strip out the view does nothing.

scrawl:
Something like this might work:

```
computeLocalToWorldMatrix(Matrix, NodeVisitor)
   if visitor is cullVisitor:
      eyePoint = cullVisitor's eyePoint
      set matrix translation to 0
   else:
      translate matrix by eyePoint
```
This is assuming the particle emitters are called sometime during the cull traversal (I believe they are). We've already stored the eyePoint for another purpose so this isn't much more of a hack than the existing code.

scrawl:
Second idea:

Change this line in Emitter
```
 const osg::Matrix& ltw = getLocalToWorldMatrix();
```

to something like

```
ltw = _cullVisitor->getModelViewMatrix() * inverse(cullVisitor->getViewMatrix())
```

so we're effectively no longer calling the CameraRelativeNode's computeLocalToWorldMatrix.

drummyfish:
Okay, I get where the problem with the camera, but not really the solution. Where do I get `_cullVisitor` in `Emitter::emitParticles(...)`? Sorry, this is probably just my lack of knowledge about OSG. Also: `osgParticle::ModularEmitter` is now being used for rain whereas `NifOsg::Emitter` is used for other particle systems (snow) - is there any difference between those? Is `NifOsg::Emitter` just a fork of OSG emitter so that we can modify it?

scrawl:
> Where do I get _cullVisitor in Emitter::emitParticles(...)?

There's a _current_nodevisitor member in ParticleProcessor which Emitter inherits from. Just need to cast from NodeVisitor to osgUtil::CullVisitor.

> Is NifOsg::Emitter just a fork of OSG emitter so that we can modify it?

Yes.

drummyfish:
Thank you, `_current_nodevisitor` is private so I can't access it, but I'll try to deal with it somehow, will report back.

drummyfish:
So I at least tried to make the other weather particles (snow, blizzard, ashes, ...) to work with my current solution, but I'm getting a lot of weird behavior and errors, I think I'll need some time to study how these things work.

To sum up I've been able to fix the rain, but I wouldn't include the changes until we find a unified solution to all particle systems, so I'll close this for now.

scrawl:
> private

Ah, too bad. Ok, then we can scratch the idea from [comment](https://github.com/OpenMW/openmw/pull/1487#issuecomment-333584079). [other comment](https://github.com/OpenMW/openmw/pull/1487#issuecomment-333581033) might still work.

One day, I'm going to tear out the mess that is osgParticle and replace it with something else... but it's not this day.

Have you considered keeping the particle system attached where it is, then just manually move each particle based on camera movement? We might also consider a 'wrap around' effect so that particles don't totally disappear if the camera is moving too fast.

drummyfish:
> Have you considered keeping the particle system attached where it is, then just manually move each particle based on camera movement?

Yes, but updating particles like that each frame would be slower than just offsetting them at start, so I just scrapped that.

> We might also consider a 'wrap around' effect

Yes, I've been thinking about this too, that would be elegant and also greatly reduce the number of particles to update.

drummyfish:
Maybe I could simply subclass `ParticleSystem` and make it offset & duplicate the particles [right before rendering](https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osgParticle/ParticleSystem.cpp#L322) to achieve camera inependent placement & wrap around effect. But I dunno if I could keep the particles sorted that way, but perharps knowing the camera direction I could.

drummyfish:
I'm testing it now - looks good and I can't even notice any performance drop with the rain.

PR #1489, opened 2017-10-03
Use an "owned crosshair" feature for items in containers

akortunov:
Requested by scrawl in #1482.

If a focused item is in container, we should check if the player can take items from this container, instead of owned common check.

How it works:
1. If the focused item is in container, get ContainerStore for current opened container.
2. Check if the container store is the same, as for focused item.
3. If yes, check isAllowedToUse() for opened container instead of focused item itself.

### DIFF
--- a/apps/openmw/mwgui/containeritemmodel.cpp
+++ b/apps/openmw/mwgui/containeritemmodel.cpp
@@ -6,7 +6,10 @@
 #include "../mwworld/class.hpp"
 
 #include "../mwbase/world.hpp"
+#include "../mwbase/mechanicsmanager.hpp"
 #include "../mwbase/environment.hpp"
+
+#include "../mwmechanics/actorutil.hpp"
 
 namespace
 {
@@ -45,6 +48,19 @@ ContainerItemModel::ContainerItemModel(const std::vector<MWWorld::Ptr>& itemSour
 ContainerItemModel::ContainerItemModel (const MWWorld::Ptr& source)
 {
     mItemSources.push_back(source);
+}
+
+bool ContainerItemModel::allowedToUseItems() const
+{
+    if (mItemSources.size() == 0)
+        return true;
+
+    MWWorld::Ptr ptr = MWMechanics::getPlayer();
+    MWWorld::Ptr victim;
+
+    // Check if the player is allowed to use items from opened container
+    MWBase::MechanicsManager* mm = MWBase::Environment::get().getMechanicsManager();
+    return mm->isAllowedToUse(ptr, mItemSources[0], victim);
 }
 
 ItemStack ContainerItemModel::getItem (ModelIndex index)
--- a/apps/openmw/mwgui/tooltips.cpp
+++ b/apps/openmw/mwgui/tooltips.cpp
@@ -120,7 +120,7 @@ namespace MWGui
                     if (info.caption.empty())
                         info.caption=mFocusObject.getCellRef().getRefId();
                     info.icon="";
-                    tooltipSize = createToolTip(info, true);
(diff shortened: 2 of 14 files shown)


scrawl:
Nooo, you're undoing my GUI refactoring *sobs*

Can we use the ItemModel maybe? The tooltip for container items is created from [here](https://github.com/akortunov/openmw/blob/fd225460f6ad5b5251d7378c0693aa7eec393a53/apps/openmw/mwgui/tooltips.cpp#L199) where we have access to ItemModel. A new function like ```ItemModel::isAllowedToUse(item)``` might do the trick.

Side question: from what I can tell, item tooltips in the trade window will be colored as 'owned' since the merchant owns the item, but that hint is not useful when we're buying legimitately, anyway. Right?

akortunov:
> Example

A good idea. I'll try to implement it.

akortunov:
> The tooltip for container items is created from here where we have access to ItemModel.

For some reason an engine uses only method in ItemModel class, but does not use overrides from derived classes. What am I doing wrong?

scrawl:
Most views actually use a SortFilterItemModel (ProxyItemModel) on top of the source model to handle filtering / sorting. So you'll have to implement ProxyItemModel::isAllowedToUse forwarding the call to the source model.

akortunov:
> ContainerItemModel::isAllowedToUse -> check allowedToUse according to item and container's owner

Actually, containers use InventoryItemModel instead of ContainerItemModel. Is it a bug or feature?

scrawl:
No reason in particular I think. Using ContainerItemModel should work too.

akortunov:
Should work now.

apps/openmw/mwgui/tooltips.cpp
```diff
@@ -351,35 +353,40 @@ namespace MWGui
...
+        MWWorld::Ptr target = mFocusObject;
+
+        // if a focused object is stored in container, check if the player can take items from this container instead if owner check
+        if (mFocusObject.getContainerStore() &&
+            MWBase::Environment::get().getWindowManager()->getMode() == GM_Container)
         {
-            return false;
+            // just make sure that focused object is stored in the current opened container
+            MWWorld::Ptr container = MWBase::Environment::get().getWindowManager()->getContainerWindow()->getPtr();
+            if (!container.isEmpty() && mFocusObject.getContainerStore() == &container.getClass().getContainerStore(container))
```
akortunov: This solution is a bit unclean, but at least we do not have to iterate over all containers in all active cells.

akortunov:
For now I have keep checkOwned() function: I am not sure how to handle tooltips and crosshair in non-GUI mode.

apps/openmw/mwgui/containeritemmodel.cpp
```diff
@@ -47,6 +53,29 @@ ContainerItemModel::ContainerItemModel (const MWWorld::Ptr& source)
     mItemSources.push_back(source);
 }
 
+bool ContainerItemModel::allowedToUseItem(const MWWorld::Ptr& target) const
```
scrawl: Should be as simple as ```isAllowedToUse(player, mItemSources[0])``` unless I'm missing something.

scrawl:
> For now I have keep checkOwned() function: I am not sure how to handle tooltips and crosshair in non-GUI mode.

That's fine.

apps/openmw/mwgui/containeritemmodel.cpp
akortunov: Yes, it seems you are right.

akortunov:
I uploaded a slightly simplified version.

PR #1491, opened 2017-10-03
Reverse pickpocketing

akortunov:
This PR is aimed to implement [feature #1336](https://bugs.openmw.org/issues/1336).

I have a lot of questions:
1. Can this feature be implemented before 1.0?
2. Should a victim use autoequipping if a player places clothing/armor to the victim's inventory?
3. Can we merge this feature with pickpocketing mechanics fix and use it as "advanced pickpocketing" config option?
4. Should we decrement a count of stolen items, if a player previously stole "reversed" item from the same victim?
5. Should the player be able to "overburden" victim by reverse pickpocketing?

A relevant forum discussion is [here](https://forum.openmw.org/viewtopic.php?f=6&t=4477).

### DIFF
--- a/docs/source/reference/modding/settings/game.rst
+++ b/docs/source/reference/modding/settings/game.rst
@@ -51,6 +51,18 @@ Whether or not the chance of success will be displayed in the enchanting menu.
 
 The default value is false. This setting can only be configured by editing the settings configuration file.
 
+advanced pickpocketing
+----------------------
+
+:Type:		boolean
+:Range:		True/False
+:Default:	False
+
+If this setting is true, the engine will use the the weight of an item instead of the value to calculate the chance of successful pickpocketing.
+Also the "reverse pickpocketing" will be allowed in this case.
+
+The default value is false. This setting can only be configured by editing the settings configuration file.
+
 best attack
 -----------
 
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -238,6 +238,7 @@ namespace MWBase
             /// Has the player stolen this item from the given owner?
             virtual bool isItemStolenFrom(const std::string& itemid, const std::string& ownerid) = 0;
 
+            virtual bool isBoundItem(const MWWorld::Ptr& item) = 0;
             virtual bool isAllowedToUse (const MWWorld::Ptr& ptr, const MWWorld::Ptr& target, MWWorld::Ptr& victim) = 0;
 
             /// Turn actor into werewolf or normal form.
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -250,6 +250,7 @@ namespace MWBase
             virtual void cleanupSummonedCreature(const MWWorld::Ptr& caster, int creatureActorId) = 0;
 
             virtual void confiscateStolenItemToOwner(const MWWorld::Ptr &player, const MWWorld::Ptr &item, const MWWorld::Ptr& victim, int count) = 0;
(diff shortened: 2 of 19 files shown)


akortunov:
I decided to move some logic from container.cpp to item models.

akortunov:
Known issues:
1. I am not sure about how to handle stolen items counters without adding a new API.
2. Bound items check during reverse pickpocking should be BEFORE crime reporting.

akortunov:
I think this PR is ready for testing and review.

docs/source/reference/modding/settings/game.rst
```diff
@@ -51,6 +51,18 @@ Whether or not the chance of success will be displayed in the enchanting menu.
...
 
+advanced pickpocketing
+----------------------
+
+:Type:		boolean
+:Range:		True/False
+:Default:	False
+
+If this setting is true, an engine will use an item weight instead of value to calculate chance of successful pickpocketing.
+Also a "reverse pickpocketing" feature will be enabled in this case.
```
Capostrophic: "**the** engine", "**the** item weight"/"**the** weight of an item", "**the** value", "**the** chance" and "**the** <...> feature" on this and previous lines. Definite article.

drummyfish:
I think Skyrim/Oblivion had reverse pickpocketing, didn't they? We could look at how it worked there regarding the overburdening etc. I'll try to research this a bit.

akortunov:
Well, then this PR should work as Skyrim without perk for now. The poisoning perk would require to implement poisons first.

lysol90:
Maybe we should get @zinnschlag to comment on this..? It's been a while now since this PR was done.

zinnschlag:
I commented on this issue on the forum already. But anyway:

This change seems to be outside of the scope of the OpenMW. It should be up to mod developers to make changes of this kind. I expect OpenMW to gain the necessary modding capabilities relatively quickly once we enter the post-1.0 de-hardcoding development phase. Hopefully 1.0 isn't too far off anymore (we made some substantial progress in the last couple of months). Now if 1.0 were still many years away I might consider this PR as a stopgap solution. But our current situation I just don't see the point.

akortunov:
Did anyone review/test this PR?

Amount of changes were made:
1. Move model-specific logic from container.cpp to item models (no more "if dynamic_cast") - about 80% of changes
We definely can use these changes before 1.0
2. Reverse pickpocketing itself, just provides an option and overrides one method - about 15% of changes. Requires 1.
It's hard to tell when the project will reach the **end** of post-1.0 de-hardcoding development phase.
So maybe we will have to use some "stopgap solutions" for a couple of years.
3. Change in success change formula - about 5% of changes
I can revert it, if you do not like it. We indeed can provide a way to tweak the formula after 1.0

scrawl:
My 2 cents:

* Reverse pickpocket function vs. the changed success formula are kinda unrelated - inevitably someone would ask to use one but not the other. Would not put them in the same option.
* I don't see much of a use case for the reverse pickpocket feature at the moment, unless someone created a quest to depend on it; in which case, the setting would rather belong to the content file and/or flag specific NPCs as being able to be reverse-pickpocketed.
* I don't see the current system as broken, it's just intentionally hard. Obviously, if I'm carrying very valuable things, I'm going to pay more attention to my surroundings, so using the item value kinda makes sense. If you think pickpocketing is too hard in general, you can mod 'fPickPocketChance' if you want.
* I agree with Zini that this is at best a stopgap solution until these things are moved to scripts.

akortunov:
Well, can we use item models refactoring at least?

zinnschlag:
Not opposed to the idea. @scrawl ?

Of course that would require stripping out the game mechanics changes first. General tips: Separate changes belong into separate PRs. Mixing up a feature change and a refactoring pass in the same PR definitely has hurt the review attention you were getting.

akortunov:
Feel free to close this PR. If someone needs the alternative pickpocketing mechanics, he can take it from  [this](https://github.com/akortunov/openmw/tree/pickpocket) branch.

PR #1492, opened 2017-10-04
fix rain/snow moving with player (issue #2698)

drummyfish:
I'm making this as a continuation of [this discussion](https://github.com/OpenMW/openmw/pull/1487).

This time I simply decorated the particle system draw callback that offsets the particles based on camera position and then draws them 9 times at different positions to achieve wrap-around effect. Advantages:

- kind of simple and straightforward solution
- no need to modify scene graph
- no need to modify the parameters of particle systems, they stay as they are
- no need to increase number of spawned particles as with the previous approach
- wrap-around effect (you can move quickly without getting away from the particles)
- no need to edit stuff all around the codebase, just the sky code (EDIT: not quite true after a few more commits)

Issues:

- Performance? I'm drawing the particle system 9 times each frame plus updating the position of each particle. I didn't notice an FPS drop though. Could you test?
- In the draw callback I don't care about the draw order for transparency. Should I?
- This is incomplete and only works for rain and snow now, I just want to discuss this early.
- The particles still move with the camera and don't wrap in Z-direction, I guess this won't be noticeable normally.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -256,6 +256,7 @@ namespace MWRender
         mRootNode->getOrCreateStateSet()->addUniform(mUniformRainIntensity);
 
         mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager()));
+        mSky->setCamera(mViewer->getCamera());
 
         source->setStateSetModes(*mRootNode->getOrCreateStateSet(), osg::StateAttribute::ON);
 
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -255,7 +255,7 @@ namespace MWRender
         
         mRootNode->getOrCreateStateSet()->addUniform(mUniformRainIntensity);
 
-        mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager(), this));
+        mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager()));
 
         source->setStateSetModes(*mRootNode->getOrCreateStateSet(), osg::StateAttribute::ON);
 
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -255,7 +255,7 @@ namespace MWRender
         
         mRootNode->getOrCreateStateSet()->addUniform(mUniformRainIntensity);
 
-        mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager()));
+        mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager(), this));
 
         source->setStateSetModes(*mRootNode->getOrCreateStateSet(), osg::StateAttribute::ON);
 
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -39,12 +39,15 @@
 #include <components/sceneutil/statesetupdater.hpp>
 #include <components/sceneutil/controller.hpp>
 #include <components/sceneutil/visitor.hpp>
+#include <components/nifosg/particle.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
 
 #include "vismask.hpp"
 #include "renderbin.hpp"
+
+#define PARTICLE_WIDTH 600.0
 
 namespace
 {
@@ -1091,8 +1094,9 @@ private:
     }
 };
 
-SkyManager::SkyManager(osg::Group* parentNode, Resource::SceneManager* sceneManager)
+SkyManager::SkyManager(osg::Group* parentNode, Resource::SceneManager* sceneManager, MWRender::RenderingManager *renderingManager)
     : mSceneManager(sceneManager)
+    , mRendering(renderingManager)
     , mAtmosphereNightRoll(0.f)
     , mCreated(false)
     , mIsStorm(false)
@@ -1337,6 +1341,43 @@ protected:
     osg::Uniform* mRainIntensityUniform;
 };
 
+class WeatherParticleDrawCallback : public osg::Drawable::DrawCallback
+{
+public:
+    WeatherParticleDrawCallback(MWRender::RenderingManager *renderingManager) : osg::Drawable::DrawCallback()
+    {
+      mRendering = renderingManager;
+    }
+
+    virtual void drawImplementation(osg::RenderInfo& renderInfo, const osg::Drawable *drawable) const
+    {
+        osg::Vec3 cameraPos = mRendering->getCameraPosition();
+        osg::Vec3 cameraOffset = osg::Vec3(
+          PARTICLE_WIDTH - fmod(cameraPos.x(), PARTICLE_WIDTH / 2),
+          PARTICLE_WIDTH - fmod(cameraPos.y(), PARTICLE_WIDTH / 2),
+          0);
+
+        osgParticle::ParticleSystem *ps = (osgParticle::ParticleSystem *) drawable;
+
+        for (int xOff = 0; xOff < 3; xOff++)
+            for (int yOff = 0; yOff < 3; yOff++)
+                {  
+                    osg::Vec3 offset = cameraOffset + osg::Vec3(-1 * xOff * PARTICLE_WIDTH, -1 * yOff * PARTICLE_WIDTH,0);
+
+                    for(int i = 0; i < ps->numParticles(); i++)
+                        ps->getParticle(i)->setPosition(ps->getParticle(i)->getPosition() + offset);
+
+                    ps->drawImplementation(renderInfo);
+
+                    for(int i = 0; i < ps->numParticles(); i++)
+                        ps->getParticle(i)->setPosition(ps->getParticle(i)->getPosition() - offset);
+                }
+    }
+
+protected:
+    MWRender::RenderingManager *mRendering;
+};
+
 void SkyManager::createRain()
 {
     if (mRainNode)
@@ -1345,6 +1386,8 @@ void SkyManager::createRain()
     mRainNode = new osg::Group;
 
     mRainParticleSystem = new osgParticle::ParticleSystem;
+    mRainParticleSystem->setDrawCallback(new WeatherParticleDrawCallback(mRendering));
+
     mRainParticleSystem->setParticleAlignment(osgParticle::ParticleSystem::FIXED);
     mRainParticleSystem->setAlignVectorX(osg::Vec3f(0.1,0,0));
     mRainParticleSystem->setAlignVectorY(osg::Vec3f(0,0,1));
@@ -1370,8 +1413,8 @@ void SkyManager::createRain()
     emitter->setParticleSystem(mRainParticleSystem);
 
     osg::ref_ptr<osgParticle::BoxPlacer> placer (new osgParticle::BoxPlacer);
-    placer->setXRange(-300, 300); // Rain_Diameter
-    placer->setYRange(-300, 300);
+    placer->setXRange(-PARTICLE_WIDTH / 2, PARTICLE_WIDTH / 2); // Rain_Diameter
+    placer->setYRange(-PARTICLE_WIDTH / 2, PARTICLE_WIDTH / 2);
     placer->setZRange(300, 300);
     emitter->setPlacer(placer);
 
@@ -1559,6 +1602,12 @@ void SkyManager::setWeather(const WeatherResult& weather)
 
             SceneUtil::DisableFreezeOnCullVisitor disableFreezeOnCullVisitor;
             mParticleEffect->accept(disableFreezeOnCullVisitor);
+
+            osgParticle::ParticleSystem *ps = (osgParticle::ParticleSystem *)
+              (mParticleEffect->asGroup()->getChild(1)->asGroup()->getChild(0)
+               ->asGroup()->getChild(2));
+
+            ps->setDrawCallback(new WeatherParticleDrawCallback(mRendering));
         }
     }
 
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -1361,24 +1361,38 @@ public:
 
         std::vector<osg::Vec3> positionBackups;
 
+        osg::Matrix toWorld, toLocal;
+
+        toWorld.makeIdentity();
+        toLocal.makeIdentity();
+
+        std::vector<osg::Matrix> worldMatrices = drawable->getWorldMatrices();
+
+        if (!worldMatrices.empty())
+            {
+                toWorld = worldMatrices[0];
+                toLocal.invert(toWorld);
+            }
+
         for (int i = 0; i < ps->numParticles(); i++)
-          {
-             osgParticle::Particle *particle = ps->getParticle(i);
+            {
+                 osgParticle::Particle *particle = ps->getParticle(i);
+                 positionBackups.push_back(particle->getPosition());
+                 particle->setPosition(toWorld.preMult(particle->getPosition()));
+                 particle->setPosition(particle->getPosition() - cameraOffset);
 
-             positionBackups.push_back(particle->getPosition());
+                 if (particle->getPosition().x() > mRangeX / 2.0)        // wrap-around effect
+                     particle->setPosition(particle->getPosition() - osg::Vec3(mRangeX,0,0)); 
+                 else if (particle->getPosition().x() < -mRangeX / 2.0)
+                     particle->setPosition(particle->getPosition() + osg::Vec3(mRangeX,0,0)); 
 
-             particle->setPosition(particle->getPosition() - cameraOffset);
+                 if (particle->getPosition().y() > mRangeY / 2.0)  
+                     particle->setPosition(particle->getPosition() - osg::Vec3(0,mRangeY,0)); 
+                 else if (particle->getPosition().y() < -mRangeY / 2.0)
+                     particle->setPosition(particle->getPosition() + osg::Vec3(0,mRangeY,0)); 
 
-             if (particle->getPosition().x() > mRangeX / 2.0)        // wrap-around effect
-                 particle->setPosition(particle->getPosition() - osg::Vec3(mRangeX,0,0)); 
-             else if (particle->getPosition().x() < -mRangeX / 2.0)
-                 particle->setPosition(particle->getPosition() + osg::Vec3(mRangeX,0,0)); 
-
-             if (particle->getPosition().y() > mRangeY / 2.0)  
-                 particle->setPosition(particle->getPosition() - osg::Vec3(0,mRangeY,0)); 
-             else if (particle->getPosition().y() < -mRangeY / 2.0)
-                 particle->setPosition(particle->getPosition() + osg::Vec3(0,mRangeY,0)); 
-          }
+                 particle->setPosition(toLocal.preMult(particle->getPosition()));
+             }
 
         ps->drawImplementation(renderInfo);
 
@@ -1388,6 +1402,7 @@ public:
 
 protected:
     float mRangeX, mRangeY;
+
 };
 
 void SkyManager::createRain()
@@ -1432,8 +1447,7 @@ void SkyManager::createRain()
     emitter->setPlacer(placer);
 
     osg::ref_ptr<osgParticle::ConstantRateCounter> counter (new osgParticle::ConstantRateCounter);
-//    counter->setNumberOfParticlesPerSecondToCreate(600.0);
-    counter->setNumberOfParticlesPerSecondToCreate(2000);
+    counter->setNumberOfParticlesPerSecondToCreate(600.0);
     emitter->setCounter(counter);
 
     osg::ref_ptr<RainShooter> shooter (new RainShooter);
@@ -1507,8 +1521,8 @@ void SkyManager::update(float duration)
         osg::Quat quat;
         quat.makeRotate(osg::Vec3f(0,1,0), mStormDirection);
 
-    if (mParticleNode)
-      mParticleNode->setAttitude(quat);
+        if (mParticleNode)
+            mParticleNode->setAttitude(quat);
 
         mCloudNode->setAttitude(quat);
     }
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -18,6 +18,8 @@
 #include <osg/PolygonOffset>
 #include <osg/observer_ptr>
 
+#include <OpenThreads/ReadWriteMutex>
+
 #include <osgParticle/ParticleSystem>
 #include <osgParticle/ParticleSystemUpdater>
 #include <osgParticle/ModularEmitter>
@@ -39,7 +41,6 @@
 #include <components/sceneutil/statesetupdater.hpp>
 #include <components/sceneutil/controller.hpp>
 #include <components/sceneutil/visitor.hpp>
-#include <components/nifosg/particle.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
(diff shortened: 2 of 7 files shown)


apps/openmw/mwrender/sky.cpp
```diff
@@ -1559,6 +1602,12 @@ void SkyManager::setWeather(const WeatherResult& weather)
 
             SceneUtil::DisableFreezeOnCullVisitor disableFreezeOnCullVisitor;
             mParticleEffect->accept(disableFreezeOnCullVisitor);
+
+            osgParticle::ParticleSystem *ps = (osgParticle::ParticleSystem *)
+              (mParticleEffect->asGroup()->getChild(1)->asGroup()->getChild(0)
```
drummyfish: This is ugly and only works for snow. Is there a way I can find a `ParticleSystem` node in the subtree?

drummyfish:
Mkay, so now all particle effects work, but I noticed a new issue (visible with weather 6 and 9): the direction the particles travel in is affected by the player movement for some reason.

drummyfish:
I replaced `PositionAttitudeTransform` node of the particle system with `Group` and the problem is fixed (why was it even there?). But now I'm noticing I can see the weathers 6 and 9 wrapping in discrete "jumps". The ideal solution would be gradual fading, but I think this is mostly cause by the fixed width constant I added (`PARTICLE_WIDTH`).

scrawl:
Hm, my first thought was 'evil, evil' when I see how particles are modified just before being rendered. But then I like the list of advantages and I like that this has little impact on the design (so we can take it out again with little consequences if we think of something else).

You should be able to get the current Camera and its position from the osg::RenderInfo& so that there's no need to add the RenderingManager dependency.

If you use a cull callback instead of a draw callback, like `node->addCullCallback(mycallback)` you can also get to the active camera and this might even make sorting work.

> I replaced `PositionAttitudeTransform` node of the particle system with `Group` and the problem is fixed (why was it even there?). 

Ash storm particles need to be rotated so they appear to come from the Red Mountain.

> Performance? I'm drawing the particle system 9 times each frame plus updating the position of each particle. I didn't notice an FPS drop though. Could you test?

There's really no need to worry about doing something 'per particle' because we already do many times (sorting them, updating position by velocity, etc.). Shouldn't affect performance in a detectable way.

Drawing X particle systems rather than one - it depends. If all of them are visible at all times, it's more efficient to use 1 particle system with X times the particles. If some are culled, splitting may make sense.

For Ash storms, I highly doubt that rendering more of it is going to be viable. Even one copy of it already covers a large area and tends to make FPS drop on weaker PCs.

drummyfish:
Thank you again @scrawl for your generous feedback :) I'll definitely take use of `RenderInfo` then and will also look at the other stuff like keeping the attitude transform plus the fading I mentioned. Hopefully tomorrow.

drummyfish:
It looks quite good now except for the blizzard, that one for some reason changes direction even up/down depending on player's rotation etc. Weird.

drummyfish:
[Here](https://www.youtube.com/watch?v=kxSe6A-zztM) I'm posting a footage for convenience, including issues to be resolved.

drummyfish:
Instead of rendering 9 times I'm now trying to modify the particles inside `NifOsg::ParticleSystem` and then render only once, but `Particle_vector` is defined differenly in different version of OSG (in mine they are dynamically allocated and the vector holds pointers to particles while in others it's a vector of static objects), which complicates things.

drummyfish:
OK, my next idea is to simply apply the wrap around effect per-particle.

scrawl:
> blizzard, that one for some reason changes direction even up/down depending on player's rotation etc. Weird.

If I had to make a guess, it's that the blizzard particle system is parented to some transform node that rotates it, while the other particle systems are not rotated. This transformation would need to be taken into account when you move the particle system in its local space but really expect this to move it in world (or camera) space.

> OK, my next idea is to simply apply the wrap around effect per-particle.

You're probably looking for osgParticle::Operator. One can implement a sub-class of Operator, add it to the particle system and then this custom function gets called every frame for every particle. From there you can easily convert from local space to world space too (example [here](https://github.com/OpenMW/openmw/blob/76f6964583b88f6560517996c429d365dd2fab5a/components/nifosg/particle.cpp#L372))

drummyfish:
> blizzard particle system is parented to some transform node that rotates it, while the other particle systems are not rotated

That makes sense, will check that, thanks :)

> You're probably looking for osgParticle::Operator.

I did it my way before I read this, so if this proves a good solution I'll reimplement using the Operator class, which of course is more elegant.

drummyfish:
I dunno if I'm doing the transform to world space correctly but the issue with blizzard seems to be gone. I think I could also apply the wrap effect in Z-direction now.

drummyfish:
Although wrapping in Z-direction wouldn't be good, the camera could be placed at the spawn level and see the particles appearing out of thin air. Would be better to just move the emitter in Z-direction and keep the spawned particles placed absolutely.

drummyfish:
Could some of you please test this when you have time and tell me how it looks? Thanks a lot :)

@scrawl I could move the code to `osgParticle::Operator` now, but would have to get the dependency on `RenderingManager` back as I wouldn't have `RenderInfo` to get the camera position from. So do you think I should do that?

drummyfish:
I could use custom `osgParticle::Operator` like this

```
restore particle positions;               // the code I have now
wrap-around particles in XY directions;   // the code I have now

// handle Z-direction:

for each particle:
  if (particle is newly spawned) // how to check this though?
    particle Z position += camera Z position  // place newly spawned particles relatively to camera

  particle Z position -= camera Z position  // move to world space before rendering
```

This should achieve the "emitter tied to camera but particles moving independently" in Z-direction, BUT: It seems we're not using `osgParticle::ParticleEffect` anywhere? It's needed to get access to `osgParticle::ModularProgram` object in order to use the operator.

Or should we leave things as they're now? The Z-direction movement isn't really noticeable during normal gameplay.

scrawl:
I've had a test. To see the effects better, I slowed down the time in engine.cpp (e.g. replace 'simulationTime += dt' by 'simulationTime += dt/5'). This slows the particle systems but not the player movement, ideal for testing. 

The rain seems OK, but ashstorms don't work like I think they should. Moving with the direction of the storm doesn't make the particles move 'slower' relatively to the player. Looks like your code assumes there is only <i>one</i> particle system and attaches the callback to the one found. This is a wrong assumption for the ash storm file which has many particle systems.

I'm not sure if using the BoxPlacer's area (where particles are initially placed) for the 'wrap around' radius is going to work for the non-rain particles. This works for rain because rain spawns in a random XY position and then only moves down. Ash particles are different, they mostly spawn in the same place then start moving horizontally. I think we can use the particle system's initial bounding box (set [here](https://github.com/OpenMW/openmw/blob/master/components/nifosg/nifloader.cpp#L826)) or even the current bounding box.

Regarding using osgParticle::Operator, we already have an example in our codebase (PlanarCollider).

>  if (particle is newly spawned) // how to check this though?
>    particle Z position += camera Z position  // place newly spawned particles relatively to camera
>  particle Z position -= camera Z position  // move to world space before rendering

I can't make sense of this snippet. We can handle Z axis the same way as XY axis, can't we?

> @scrawl I could move the code to osgParticle::Operator now, but would have to get the dependency on RenderingManager back as I wouldn't have RenderInfo to get the camera position from. So do you think I should do that?

We can always just set the Camera explicitely (i.e. RenderingManager calls mSky->setCamera(...) ) to cut the dependency on RenderingManager. That still fixes the sky to one camera, but that's not really a concern as long as we're only using one (actually we do use more cameras for water reflections, it's just the weather particles aren't shown in there IIRC).

drummyfish:
I'll try to slow down the time as you did, thank you for the tip and great feedback.

> I can't make sense of this snippet. We can handle Z axis the same way as XY axis, can't we?

If we do handle Z-direction in the same way (let's assume rain), then there will be levels at fixed heights (for example 0, 100, 200 etc. if we wrap by 100) in the world at which rain particles spawn, which is bad, because if the camera is near that Z-position, it will be able to see the particles spawning. The solution is to wrap the particles around, but spawn them relatively to the camera (let's say always 100 units above it in world coordinates), which is not the case now (the particles are placed in local space of the particle system). To place them in world coordinates I'll probably decorate the particle effect placer.

ACTUALLY now that I'm thinking about it, this will be a problem for the "horizontally spawned" particle systems. Actually blizzard has a weird behavior now that I think is this very issue. Ash storm would be affected too but as you pointed out my current solution isn't applied to it yet so it wasn't visible.

SO, to draw a conclusion:

- I **have to** make the particles spawn relatively to the camera.
- Then the wrap-around will stay as it is (with the updates of the boundary computations you mentioned of course).
- I will treat all directions equally, so no movement of particles with the camera whatsoever in any direction.

scrawl:
Sorry, I still don't get it. I think my hang up is with 

> make the particles spawn relatively to the camera.

Isn't that already the case now, with the emitter parented to CameraRelativeNode?

drummyfish:
That is the case without wrap-around, but with wrap-around I am offsetting the particles according to camera position to cancel this out, so effectively each particle's position is now absolute. I'll make a picture of what I mean later today to make this clear.

drummyfish:
I have wrapping in all directions now (without the correct spawning of particles so far), behaves basically as I expected - rain and blizzard looks kinda OK, ashstorm is underground and then like 100 meters in the sky (that's what should be fixed by the relative spawning). Snow looks super weird somehow, it's very sparse and non-uniformly distributed - maybe the bounding box is too big to be used as the wrapping range.

drummyfish:
Can mods define new weather particle systems? If not, I could be separately fine tuning each of the weathers I'm dealing with now. Otherwise I'll have to find general solution of course. I think I'll also go take a close look at vanilla Morrowind now.

drummyfish:
Summing up:

- Storms should be just left parented to camera like in vanilla.
- Other systems (rain and snow) seem to be wrapping in all directions without relative spawning. The issue I illustrated with my picture probably isn't very apparent, at least not with rain which I tested. So
  - Rain can be considered fixed (though it may be broken in latest commit).
  - Snow is behaving very weird. Just after creation it looks good but over time it starts to make dense chunks and sparse areas. I tried setting the wrap bounds to bounding box, boxplacer area (1024 x 1024) and randomly to other values but that didn't fix it. I'm gonna take a closer look at the structure of the system to see what's going on. Also when looking vertically at the snowflakes they rotate with camera, but that is also vanilla behavior.

drummyfish:
OK, fixed, may I ask for opinions now?

scrawl:
I'll test it in a bit.

By the way, we had (shader) swim ripples in the past, last seen in openmw-0.34 before commit  https://github.com/OpenMW/openmw/commit/dbd4abd6fe3ae3b3f1e812c8069b094bad50846e. They weren't that great I think, might still be useful as a reference.

apps/openmw/mwrender/sky.cpp
```diff
@@ -39,16 +41,22 @@
...
 #include "../mwbase/world.hpp"
 
 #include "vismask.hpp"
 #include "renderbin.hpp"
 
+#define RAIN_WIDTH 600.0
+#define RAIN_HEIGHT 600.0
+
+#define WEATHER_WRAP_WIDTH 1024
+#define WEATHER_WRAP_HEIGHT 800
```
scrawl: WEATHER_WRAP_HEIGHT is unused.

apps/openmw/mwrender/sky.cpp
```diff
@@ -1559,6 +1646,16 @@ void SkyManager::setWeather(const WeatherResult& weather)
 
             SceneUtil::DisableFreezeOnCullVisitor disableFreezeOnCullVisitor;
             mParticleEffect->accept(disableFreezeOnCullVisitor);
+
+            SceneUtil::FindByClassVisitor findEmitterVisitor(std::string("Emitter"));
```
scrawl: Why not just find the ParticleSystem(s) directly?

apps/openmw/mwrender/sky.cpp
```diff
@@ -1559,6 +1646,16 @@ void SkyManager::setWeather(const WeatherResult& weather)
...
+            mParticleEffect->accept(findEmitterVisitor);
+
+            for (unsigned int i = 0; i < findEmitterVisitor.mFoundNodes.size(); i++)
+            {
+                NifOsg::Emitter *emitter = (NifOsg::Emitter *) findEmitterVisitor.mFoundNodes[i];
+                NifOsg::ParticleSystem *ps = (NifOsg::ParticleSystem *) emitter->getParticleSystem();
```
scrawl: osgParticle::ParticleSystem would be sufficient.

scrawl: Prefer static_cast over c-style casts.

components/nifosg/particle.hpp
```diff
@@ -225,6 +225,7 @@ namespace NifOsg
 
         void setShooter(osgParticle::Shooter* shooter);
         void setPlacer(osgParticle::Placer* placer);
+        osgParticle::Placer *getPlacer();
```
scrawl: No longer needed?

apps/openmw/mwrender/sky.cpp
```diff
@@ -1337,6 +1345,81 @@ class RainFader : public AlphaFader
...
+        std::vector<osg::Vec3> positionBackups;
+
+        osg::Matrix toWorld, toLocal;
+
+        toWorld.makeIdentity();
+        toLocal.makeIdentity();
+
+        std::vector<osg::Matrix> worldMatrices = drawable->getWorldMatrices();
+
+        if (!worldMatrices.empty())
```
scrawl: Weird indentation style... is this done on purpose or a case of IDE's gone bad?

apps/openmw/mwrender/sky.cpp
```diff
@@ -39,16 +41,22 @@
 #include <components/sceneutil/statesetupdater.hpp>
 #include <components/sceneutil/controller.hpp>
 #include <components/sceneutil/visitor.hpp>
+#include <components/nifosg/particle.hpp>
```
scrawl: Include can be removed after my other comments are addressed.

apps/openmw/mwrender/sky.cpp
scrawl: Prefer variables over #define's, and put them where they are actually used.

apps/openmw/mwrender/sky.cpp
```diff
@@ -1337,6 +1345,81 @@ class RainFader : public AlphaFader
     osg::Uniform* mRainIntensityUniform;
 };
 
+class WeatherParticleDrawCallback : public osg::Drawable::DrawCallback
+{
+public:
+    WeatherParticleDrawCallback(osg::Vec3 wrapRange) : osg::Drawable::DrawCallback()
```
scrawl: `const osg::Vec3& wrapRange`

apps/openmw/mwrender/sky.cpp
```diff
@@ -1337,6 +1345,81 @@ class RainFader : public AlphaFader
...
+          fmod(cameraPos.x(), mWrapRange.x()),
+          fmod(cameraPos.y(), mWrapRange.y()),
+          fmod(cameraPos.z(), mWrapRange.z()));
+
+        std::vector<osg::Vec3> positionBackups;
+
+        osg::Matrix toWorld, toLocal;
+
+        toWorld.makeIdentity();
+        toLocal.makeIdentity();
```
scrawl: Matrix is initialized as identity matrix by default.

apps/openmw/mwrender/sky.cpp
drummyfish: my bad

drummyfish:
Thanks you, I was lazy to clean up after my experiments, sorry for that :) If there's anything else, let me know.

scrawl:
Everything seems to be working fine.

However, I'm not sure if the mutex stuff is really correct. Your lock should be a WriteLock, because you write to the particles. And what if after the unlock(), but before the drawImplementation() another thread locks the mutex and writes over your particles (i.e. runs a simulation step)? Technically, we would want to keep the mutex locked the whole time.

(Frankly, the "mutex" design in osgParticle looks pretty broken anyway. The mutex works around the worst issues like crashes, but does nothing to prevent the rendering thread from jumping a frame ahead if the next frame is already prepared. But we can't fix that here.)

Another potential issue is culling: when we wrap the particles in drawImplementation, and let's say all particles are currently off-screen, the drawable is culled and drawImplementation() is not invoked. I don't know how likely this is to happen in practice (extremely high movement speeds I guess?), but it's worth thinking about.

So can we give the osgParticle::Operator way a try maybe? Shouldn't have to deal with anything threading related that way, and it should fix the culling issue.

drummyfish:
You're right about the mutex, this needs fixing.

Culling shouldn't be a problem because if the draw callback isn't invoked, nothing bad happens, right? The particles are only temporarily moved during the draw callback and then moved back, so outside the draw callback they keep living in the local space of the particle system.

So yes, your suggestion of operator will probably be the right solution.

drummyfish:
I can't find any `ParticleEffect` instance in the scene to assign the operator to though. I though it was `SkyManager::mParticleEffect` but now that I checked it is `Group`. I searched with my `FindByClassVisitor` but nothing found. Sorry for my noobiness :(

drummyfish:
Maybe now we're good?

apps/openmw/mwrender/sky.hpp
```diff
@@ -161,6 +163,8 @@ namespace MWRender
 
         void listAssetsToPreload(std::vector<std::string>& models, std::vector<std::string>& textures);
 
+        void setCamera(MWRender::Camera *camera);
```
scrawl: I was thinking of passing the osg::Camera actually. Smaller dependency, I guess.

apps/openmw/mwrender/sky.hpp
```diff
@@ -9,6 +9,8 @@
 #include <osg/ref_ptr>
 #include <osg/Vec4f>
 
+#include <apps/openmw/mwrender/camera.hpp>
```
scrawl: Just for future reference (with this line probably being removed due to my last comment), I would have used `#include "camera.hpp"`. More concise, less likely to break if we ever change the directory structure, and consistent with the rest of the codebase.

apps/openmw/mwrender/sky.cpp
```diff
@@ -27,6 +27,10 @@
 #include <osgParticle/ConstantRateCounter>
 #include <osgParticle/RadialShooter>
 
+#include <osgParticle/Operator>
+#include <osgParticle/ParticleEffect>
```
scrawl: Include to ParticleEffect can be removed.

scrawl:
So apart from those minor style issues I think we're fine. I've yet to do a final test I'm assuming for now that nothing was broken by the switch to Operator.

What do you think we should do with the commit history? It is a bit long but I don't want to squash into one commit either. Maybe we can squash the smaller fixes maybe and the two commits that were immediately reverted?

scrawl:
By the way, github still thinks you're a 'First time contributor' - looks like your commits aren't recognized due to a wrong email address in git.

drummyfish:
>What do you think we should do with the commit history?

Sure, I can make it smaller, just give me a hint what git command I should use :)

>By the way, github still thinks you're a 'First time contributor'

Thanks, gonna check that.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -256,6 +256,7 @@ namespace MWRender
         mRootNode->getOrCreateStateSet()->addUniform(mUniformRainIntensity);
 
         mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager()));
+        mSky->setCamera(mCamera->getOsgCamera());
```
scrawl: New method isn't needed. Use mViewer->getCamera()

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -256,7 +256,10 @@ namespace MWRender
         mRootNode->getOrCreateStateSet()->addUniform(mUniformRainIntensity);
 
         mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager()));
-        mSky->setCamera(mCamera->getOsgCamera());
+
+        std::vector<osg::Camera*> cameras;
+        mViewer->getCameras(cameras,true);
```
scrawl: Eh, that works too, but mViewer->getCamer<b>a</b>() is easier no? :)

drummyfish: I appreciate the patience you have with me :)

drummyfish:
Thanks to these imperfections I at least get to practice the squashing I just learned.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -257,6 +257,9 @@ namespace MWRender
 
         mSky.reset(new SkyManager(sceneRoot, resourceSystem->getSceneManager()));
 
+        std::vector<osg::Camera*> cameras;
```
scrawl: Remove the unused `std::vector<osg::Camera*> cameras;`

apps/openmw/mwrender/sky.cpp
```diff
@@ -1337,6 +1340,87 @@ class RainFader : public AlphaFader
     osg::Uniform* mRainIntensityUniform;
 };
 
+void SkyManager::setCamera(osg::Camera *camera)
+{
+  mCamera = camera;
```
scrawl: Strangely, the identation here is wrong while it's fine in all of the other added code.

apps/openmw/mwrender/sky.hpp
```diff
@@ -8,6 +8,7 @@
 
 #include <osg/ref_ptr>
 #include <osg/Vec4f>
+#include <osg/Camera>
```
scrawl: It's usually preferred to use a forward declaration if possible.

apps/openmw/mwrender/sky.cpp
```diff
@@ -1337,6 +1340,87 @@ class RainFader : public AlphaFader
...
 
+void SkyManager::setCamera(osg::Camera *camera)
+{
+    mCamera = camera;
+}
+
+class WrapAroundOperator : public osgParticle::Operator
+{
+public:
+    WrapAroundOperator(osg::Camera *camera, osg::Vec3 wrapRange): osgParticle::Operator()
```
scrawl: To avoid a redundant copy, this variable can be passed by const reference:

`..., const osg::Vec3& wrapRange)`

I usually don't care about performance issues in places where it's not important at all, but this particular one is something cppcheck (a static analyser) will complain about so I prefer to get them fixed everywhere and keep the code consistent.

The same applies to using 'i++' instead of '++i' because the former (in theory) creates an extra temporary variable (according to cppcheck, though I highly doubt that modern compilers don't already optimize this in all cases).

drummyfish: Okay, I will get cppcheck too so I can also see these things.

drummyfish: Should I pre-increment everywhere in the code then?

scrawl: > Should I pre-increment everywhere in the code then?

Sure.

Admittedly, It's been a while since I last cppchecked. If it no longer flags those things, feel free to keep the code as is - but it still might be nice to for the sake of consistency.

scrawl: (That was the last nitpick, I promise. Will test now and merge when I come back)

drummyfish: Cppcheck doesn't complain but I will change to pre-increment anyway. It found some memory leaks in other files BTW. I'm OK with this nitpicking, I'm learning a lot.

scrawl: We also have Coverity Scan, but it's not been run for quite a while. Updated the branch and:

>  834 files changed, 27339 insertions(+), 9312 deletions(-)

This should be interesting.

scrawl:
Thanks for this feature, it really makes a big difference.

drummyfish:
I thank you for your guidance, you make a great job at always putting your criticism in a positive way, it makes this project really feel enjoyable and fun.

apps/openmw/mwrender/sky.cpp
scrawl: Had to disable a bunch of stuff to get the build under the time limit, now Coverity Scan seems to have completed and found some new defects.

However, I can't actually view them as I seem to have locked myself out of the OpenMW project on Coverity when I changed my github email. @psi29a you have an account there right, can you let me in?

PR #1493, opened 2017-10-06
Regression fixes

akortunov:
1. Fixes [bug #4146](https://bugs.openmw.org/issues/4146).
2. An engine displays magic items in the player's spellwindow as owned. Fixed.


psi29a:
I wish there was a way we could unit-test this...  you know, to prevent this sort of thing in the future?

PR #1494, opened 2017-10-07
[Editor] Fix region colors

Aesylwinn:
This fixes the regression reported in [Bug#4119](https://bugs.openmw.org/issues/4119). I have tested the following:
1. Setting region colors in both the regular table and nested table views
2. Setting the the map color in the Cell nested table, even though it seems to be an unused field. 
3. Setting colors in Preferences->Scripts

Is there anywhere else that uses colors and could be affected?

### DIFF
--- a/apps/opencs/model/world/columnimp.hpp
+++ b/apps/opencs/model/world/columnimp.hpp
@@ -692,32 +692,23 @@ namespace CSMWorld
         }
     };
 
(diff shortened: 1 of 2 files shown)


zinnschlag:
> Is there anywhere else that uses colors and could be affected?

Nothing that I am aware of. At least not yet. There will be more colour editing in a certain part of the terrain editing feature, but that hasn't even been started yet.

PR #1495, opened 2017-10-08
Add showsInInventory() check to the ContainerItemModel

akortunov:
For some reason ContainerItemModel do not have this check.

For example, you can add Light item without CanCarry flag, and this light will be shown in the items list.


akortunov:
Looks like something wrong with the OpenAL download link.

PR #1498, opened 2017-10-11
Make CMake copy the runtime resources to the correct location

AnyOldName3:
Previously, CMake only copied the runtime resources (e.g. shaders, config files, all the MyGUI stuff) to the correct directory for non-MSVC environments. The CI script was being co-opted into copying them from the solution directory to the actual directories the executables requiring them reside in.

This resolves this, meaning that now runtime resources are automatically copied to where they need to go if they've changed since the last debugging session in Visual Studio.

Resolves https://bugs.openmw.org/issues/4152

Hopefully, this will be less of a mess than my previous CMake-based commit. I've actually tested it on my own machine this time.


scrawl:
My hope is there would be a way for CMake to tell us where the actual directory for the executables is, so we don't have to add tests like 'if MSVC' or list the build configurations. Could you check what the target's LIBRARY_OUTPUT_DIRECTORY is set to? See 

https://stackoverflow.com/questions/13429656/cmake-how-to-copy-contents-of-a-directory-into-build-directory-after-make
https://stackoverflow.com/questions/7747857/in-cmake-how-do-i-work-around-the-debug-and-release-directories-visual-studio-2

To remove the duplicate function in the current version of this PR, we could probably just move the git version stuff into OpenMWMacros.

AnyOldName3:
`LIBRARY_OUTPUT_DIRECTORY` doesn't have the configuration name appended, and therefore is not any help here. The documentation says that 

> Multi-configuration generators (VS, Xcode) append a per-configuration subdirectory to the specified directory unless a generator expression is used.

I think maybe what we're supposed to be doing is something involving `file(GENERATE ...)` as that happens for each configuration type. However, I'm struggling to think of a viable way to use it that doesn't still require at least a little bit of `if (MSVC)` to toggle inclusion of the `/$<CONFIG>` bit.

I was hoping that the value `CMAKE_LIBRARY_OUTPUT_DIRECTORY_<CONFIG>` I found in the documentation would be a generator expression, and I could use it as an argument to `file(GENERATE ...)`, but it's just their way of saying that `CMAKE_LIBRARY_OUTPUT_DIRECTORY_DEBUG` and the equivalent variables for the other possible configurations exist. If it had existed, then everything would have been really easy (provided that `file(GENERATE ...)` sets up the right dependencies to ensure the resources actually got recopied when required.

That wouldn't have worked for `resources/version`, though, as certain things need setting up for that at configuration time instead of generation time. However, I found this function that might well have served as a workaround.

https://cmake.org/pipermail/cmake/2015-April/060464.html

tl;dr: I'm not convinced there's actually a nicer way to do this, although I think that the CMake people could have made it possible with a bit of foresight. An `All possible 'CMAKE_LIBRARY_OUTPUT_DIRECTORY_<CONFIG>' values in a list that I can send to a loop construct` variable would have made this easier.

AnyOldName3:
As for what to do going forward, there are a couple of options.

* Just merge this (once I've sorted the duplicated macro definition issue).

* Adjust everything so that we don't have separate debug and release folders and prevent Visual Studio users having the ability to have a debug and release build at the same time. This is not something I want to happen, as I often switch back and forth when testing shadows as the debug builds are so much slower but have debug info.

* As above, but adopt a suffix for debug executables, like OSG has. This probably requires more work than is obvious, because I'm assuming the launcher (and other sub-projects that call each other) have a string in them somewhere for the OpenMW executable name, and suddenly they will need to become configuration-dependent.

* Fork CMake, add the features we want, get it merged upstream, and then bump up OpenMW's required CMake version. This is also silly.

* Experiment with making an explicit runtime resources target which everything else depends on and hope that that presents us with a neater way of doing this. This may not work out, but it's also not mutually exclusive with merging this as an interim solution.

I think this PR is better than what's in master right now, even if I don't like manually listing the configurations CMake might emit.

scrawl:
Okay, so if CMake is being unhelpful then let's go with this version.

But is there any reason we need to copy the file to all possible configurations? Why not just the one that is currently being built?

scrawl:
Wait, doesn't this help?

> 
CMAKE_CFG_INTDIR 
    meta-variable! Please note that this is an important variable, since on multi-configuration generators it will be generated into dynamically switched content based on the configuration that the user currently selected within the generated build environment. Indicates the name of the current configuration (~ directory) for the project. May be used for any purpose which needs per-configuration-dynamic switching of strings, not just OutputDir configuration. For multi-configuration generators (e.g. MSVC) the resulting strings are typically some of "Debug", "Release", "RelWithDebInfo", or "MinSizeRel". For other compiler generators (single-configuration ones) it is typically ".", as they don't use MSVC-style configuration directories. 

https://cmake.org/Wiki/CMake_Useful_Variables

scrawl:
Or

 CMAKE_CONFIGURATION_TYPES 
    When using a multi-configuration generator, such as the one for Visual Studio, this variable contains a list of the available configurations.

AnyOldName3:
I'm not entirely sure how I missed those, as I read through the useful variables wiki page. I guess I must just have forgotten about them in between reading and writing.

AnyOldName3:
I've put `CMAKE_CONFIGURATION_TYPES` in as that was a really simple fix.

If we use `CMAKE_CFG_INTDIR` I'm worried that we'll have to manually set up more of the dependencies, whereas they're done automatically as-is. We'll have to either create a resources target which everything else depends on or add them separately to each project (the latter of these might cause superfluous rebuilds). However, it will give the freedom that we're only copying the resources to the specific configuration folder that needs it. I'm not sure that advantage is worth the effort of actually implementing the idea.

AnyOldName3:
That change should fix the CI failures. It should also make the change less dependent on the generator actually being Visual Studio.

cmake/GitVersion.cmake
```diff
@@ -26,4 +26,6 @@ else ()
     message(WARNING "Failed to get valid version information from Git")
 endif ()
 
-configure_file(${VERSION_IN_FILE} ${VERSION_FILE})
+include(${MACROSFILE})
```
scrawl: What's the value in passing the file as a variable? Why not just include(OpenMWMacros.cmake)?

components/CMakeLists.txt
```diff
@@ -2,22 +2,26 @@ project (Components)
...
             -DPROJECT_SOURCE_DIR=${PROJECT_SOURCE_DIR}
             -DVERSION_IN_FILE=${VERSION_IN_FILE}
-            -DVERSION_FILE=${VERSION_FILE}
+            -DVERSION_FILE_PATH_BASE=${VERSION_FILE_PATH_BASE}
+            -DVERSION_FILE_PATH_RELATIVE=${VERSION_FILE_PATH_RELATIVE}
             -DOPENMW_VERSION_MAJOR=${OPENMW_VERSION_MAJOR}
             -DOPENMW_VERSION_MINOR=${OPENMW_VERSION_MINOR}
             -DOPENMW_VERSION_RELEASE=${OPENMW_VERSION_RELEASE}
             -DOPENMW_VERSION=${OPENMW_VERSION}
+            -DMSVC=${MSVC}
```
scrawl: `MSVC=` not needed anymore, is it?

scrawl:
If we're not using `copy_all_files` anymore after this, it should probably be removed.

cmake/GitVersion.cmake
AnyOldName3: It can't find the file to include if it's done any other way.

CMake searches for includes relative to `CMAKE_SOURCE_DIR`, but when `GitVersion.cmake` is run (at least via Visual Studio), that takes a value which points to the location of `git-version.vcxproj`, which is dependent on where the user put their build directory.

components/CMakeLists.txt
AnyOldName3: If the current build passes CI, then it can go, but I've got a sneaking suspicion that it won't.

AnyOldName3:
Did I not get rid of `copy_all_files`? I thought I'd already done that. We definitely aren't using it.

scrawl:
CMake runs without errors, but the version file was not copied in my test.

AnyOldName3:
Did you build the components target or something that depends on the components target? The git version target is only listed as a dependency on that, so the version file is set up only when `components` is built unless you manually build the `git-version` target.

scrawl:
Ah, I see. I only ran cmake didn't (re)build anything. That explains it.

PR #1499, opened 2017-10-13
Bugfix #4155: NPCs don't equip a second ring in some cases

lukago:
https://bugs.openmw.org/issues/4155
Now NPC will autoequip any second ring if he has a free hand. NPC will also autoequip most expensive rings if has more than two.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ Makefile
 makefile
 build*
 prebuilt
+cmake-build-debug/
 
 ##windows build process
 /deps
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@ Makefile
 makefile
 build*
 prebuilt
-cmake-build-debug/
 
 ##windows build process
 /deps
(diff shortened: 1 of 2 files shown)


apps/openmw/mwworld/inventorystore.cpp
```diff
@@ -266,6 +266,7 @@ void MWWorld::InventoryStore::autoEquip (const MWWorld::Ptr& actor)
     // Autoequip clothing, armor and weapons.
     // Equipping lights is handled in Actors::updateEquippedLight based on environment light.
 
+    // the main loop iterating through all items in inventory
```
scrawl: Please don't litter the code with redundant comments. Comments that only describe what the code does are not helpful you might as well just read the code.

.gitignore
```diff
@@ -7,6 +7,7 @@ Makefile
 makefile
 build*
 prebuilt
+cmake-build-debug/
```
scrawl: I have never seen this directory if it's specific to your setup then it probably shouldn't be in our .gitignore. Just call your build directory 'build' and it will be ignored.

lukago:
Thanks for review. Unnecessary comments have been removed. 
About .gitignore: CLion default output directories for cmake builds are cmake-build-* so I think it can be left, otherwise anyone using this IDE will need to change default project settings.

PR #1500, opened 2017-10-14
Implement ExtraSpell magic effect

akortunov:
Partially implements [feature #2906](https://bugs.openmw.org/issues/2906).
Some mods can use this effect.

Note: I do not know how to implement the autoequipping after spell expiration fo now.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -536,6 +536,13 @@ namespace MWMechanics
 
                         appliedLastingEffects.push_back(effect);
 
+                        // Unequip all items, if a spell with the ExtraSpell effect was casted
+                        if (effectIt->mEffectID == ESM::MagicEffect::ExtraSpell && target.getClass().hasInventoryStore(target))
+                        {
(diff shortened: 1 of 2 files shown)


apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -536,6 +536,13 @@ namespace MWMechanics
 
                         appliedLastingEffects.push_back(effect);
 
+                        // Unequip all items, if a spell with the ExtraSpell effect was casted
+                        if (effectIt->mEffectID == ESM::MagicEffect::ExtraSpell && target.getClass().isNpc())
```
MiroslavR: Are you sure this is limited to NPCs? What about creatures with inventory store?

akortunov: Not much difference since creatures can not wear clothing or armor.
For shields and weapons we have autoequipping.

But I can replace isNpc() check to the hasInventoryStore(), if we need it.

MiroslavR: I don't think creatures equip their shields during combat (only autoequip them when they are first constructed). So it would make a difference if you can remove their shields.

PR #1501, opened 2017-10-14
Editor: Land and LandTexture Merging

Aesylwinn:
This fixes [bug#4074](https://bugs.openmw.org/issues/4074). In testing, there seems to be an issue with the distant lands feature accessing invalid texture data. However, I am unable to produce the same issue when distant lands is disabled, even when flying around for several minutes at ludicrous speed (player->SetSpeed 10000).

### DIFF
--- a/apps/opencs/model/tools/mergeoperation.cpp
+++ b/apps/opencs/model/tools/mergeoperation.cpp
@@ -41,6 +41,7 @@ CSMTools::MergeOperation::MergeOperation (CSMDoc::Document& document, ToUTF8::Fr
     appendStage (new PopulateLandTexturesMergeStage (mState));
     appendStage (new MergeLandStage (mState));
     appendStage (new FixLandsAndLandTexturesMergeStage (mState));
+    appendStage (new CleanupLandTexturesMergeStage (mState));
 
     appendStage (new FinishMergedDocumentStage (mState, encoding));
 }
--- a/apps/opencs/model/tools/mergeoperation.cpp
+++ b/apps/opencs/model/tools/mergeoperation.cpp
@@ -38,9 +38,9 @@ CSMTools::MergeOperation::MergeOperation (CSMDoc::Document& document, ToUTF8::Fr
     appendStage (new MergeRefIdsStage (mState));
     appendStage (new MergeReferencesStage (mState));
     appendStage (new MergeReferencesStage (mState));
-    appendStage (new ListLandTexturesMergeStage (mState));
-    appendStage (new MergeLandTexturesStage (mState));
+    appendStage (new PopulateLandTexturesMergeStage (mState));
     appendStage (new MergeLandStage (mState));
+    appendStage (new FixLandsAndLandTexturesMergeStage (mState));
 
     appendStage (new FinishMergedDocumentStage (mState, encoding));
 }
--- a/apps/opencs/model/tools/mergestages.cpp
+++ b/apps/opencs/model/tools/mergestages.cpp
@@ -180,5 +180,36 @@ void CSMTools::FixLandsAndLandTexturesMergeStage::perform (int stage, CSMDoc::Me
 
         CSMWorld::TouchLandCommand cmd(landTable, ltexTable, id);
         cmd.redo();
+
+        // Get rid of base data
+        const CSMWorld::Record<CSMWorld::Land>& oldRecord =
+            mState.mTarget->getData().getLand().getRecord (stage);
+
+        CSMWorld::Record<CSMWorld::Land> newRecord(CSMWorld::RecordBase::State_ModifiedOnly,
+            nullptr, &oldRecord.get());
+
+        mState.mTarget->getData().getLand().setRecord(stage, newRecord);
+    }
+}
+
+CSMTools::CleanupLandTexturesMergeStage::CleanupLandTexturesMergeStage (MergeState& state)
+    : mState (state)
+{
+}
+
+int CSMTools::CleanupLandTexturesMergeStage::setup()
+{
+    return 1;
+}
+
+void CSMTools::CleanupLandTexturesMergeStage::perform (int stage, CSMDoc::Messages& messages)
+{
+    auto& landTextures = mState.mTarget->getData().getLandTextures();
+    for (int i = 0; i < landTextures.getSize(); )
+    {
+        if (!landTextures.getRecord(i).isModified())
+            landTextures.removeRows(i, 1);
+        else
+            ++i;
     }
 }
--- a/apps/opencs/model/tools/mergestages.cpp
+++ b/apps/opencs/model/tools/mergestages.cpp
@@ -8,7 +8,9 @@
 #include "mergestate.hpp"
 
 #include "../doc/document.hpp"
+#include "../world/commands.hpp"
 #include "../world/data.hpp"
+#include "../world/idtable.hpp"
 
 
 CSMTools::StartMergeStage::StartMergeStage (MergeState& state)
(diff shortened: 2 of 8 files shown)


Aesylwinn:
I was looking into the aforementioned issue, and found that it does occurs even when distant land is disabled, and that it seems to be an issue with some flags.

Aesylwinn:
Since isDataLoaded() was apparently dead code when I decided to use it, there is no issue with me changing its functionality.

zinnschlag:
This is a bit hard to follow. Maybe it just need a bit more of high level documentation of the merge process. But these are the issues I am seeing:

1. Why is there a cleanup stage? Apparently it removed base records from the target. But there shouldn't ever be any base records in the target to begin with, because target is a single content file document, so there simply is no base.

2. In PopulateLandTexturesMergeStage you are iterating over all land texture records (ignoring only the deleted ones). There is the problem that this writes to target base, which should not happen (see 1), but also this could result in duplicate indices. 

3. Since the base land records now (after the merge) reference the new land texture record list, wouldn't you have to fully remap all of them? I can not see any place where you are doing that.

Hm, TouchLandCommand probably. That took some digging (apparently I did not remember from my last code review). Sorry, but this stuff needs a bit more documentation. At least a short description what the new land-related command classes are doing, because TouchLandCommand changing indices is really not obvious.

Aesylwinn:
My apologies, I should have documented what was happening in the former PR. Hopefully, the intent of the code is more obvious now. I reused the TouchLandCommand to avoid code duplication.

zinnschlag:
Hm, I am getting an error here, when trying to merge Morrowind, Tribunal and Bloodmoon.

> Loading failed: ESM Error: End of file while reading record header

Can you replicate the issue?

zinnschlag:
It appears the same happens with master. Strange. Did we break the merge feature at some point?

Aesylwinn:
Wasn't able to reproduce it in my branch. I was able to reproduce it after fetching/merging the current master. I would guess it has something to do with the following console line, which is where it is getting saved to.

> Adding data directory "?userdata?data"

Aesylwinn:
The changes in #1502 seem to have broken the token replacement.

AnyOldName3:
It wasn't the token replacement that I broke (although the thing I broke in turn broke the token replacement) but either way, this PR fixes it: https://github.com/OpenMW/openmw/pull/1507

zinnschlag:
Problem persists on my end. No change.

AnyOldName3:
https://www.youtube.com/watch?v=Pw2sex1mJNI

If I can't replicate it, I might need you to try a version with some output printed so I can see what's going on.

Aesylwinn:
Is it a specific file name that's causing it?

Aesylwinn:
Hm.. Once I deleted all of the *.proj files, I was able to get the following error when creating a new addon.

> Loading failed: ESM Error: End of file while reading record header File: ... Record TES3 Subrecord: Offset: 0x4

This happens when going to the commit before AnyOldName3's changes. There was a change to the Document system to provide a single VFS per document and a few changes to the esm load/save files, so I'll look into those.

AnyOldName3:
If it helps, I've had the same error but with the offset being `0x0` since 0.42 (or perhaps even earlier, but I don't have a way of testing that easily as I only have 0.42 installed) whenever I try and do anything with the CS.

AnyOldName3:
Also, I'm going to maybe have very limited time to investigate this further for a few hours (or maybe days, but I'm hoping hours) so if anyone else wants to do some digging, that would be helpful.

Aesylwinn:
> If it helps, I've had the same error but with the offset being 0x0 since 0.42

This doesn't make much sense to me. I've been creating new addons for a while now with the master branch, and never had an issue until recently.

AnyOldName3:
I've added some print statements to the version at https://github.com/AnyOldName3/openmw/tree/zini-test, and it might help me track this down if anyone who still has issues sends me the output to the console when the CS is started.

Aesylwinn:
The console output, if it helps.
```
qt5ct: using qt5ct plugin
Loading settings file: ./openmw-cs.cfg
Loading settings file: /home/aesylwinn/.config/openmw/openmw-cs.cfg
Loading config file: /home/aesylwinn/.config/openmw/openmw.cfg... done.
Loading config file: ./openmw.cfg... done.
Entering: void ConfigurationManager::processPaths(Files::PathContainer& dataDirs, bool create)
/home/aesylwinn/.openmw/Data Files
/home/aesylwinn/.openmw/Data Files
/home/aesylwinn/.openmw/Data Files
/home/aesylwinn/.openmw/NOM
/home/aesylwinn/.openmw/NOM
/home/aesylwinn/.openmw/NOM
?global?data
/usr/local/share/games/openmw/data
/usr/local/share/games/openmw/data
./data
./data

Entering: void ConfigurationManager::processPaths(Files::PathContainer& dataDirs, bool create)
?userdata?data
/home/aesylwinn/.local/share/openmw/data
/home/aesylwinn/.local/share/openmw/data
Entering: void ConfigurationManager::processPaths(Files::PathContainer& dataDirs, bool create)
/home/aesylwinn/.openmw/Data Files
/home/aesylwinn/.openmw/Data Files
/home/aesylwinn/.openmw/Data Files
/home/aesylwinn/.openmw/NOM
/home/aesylwinn/.openmw/NOM
/home/aesylwinn/.openmw/NOM
?global?data
/usr/local/share/games/openmw/data
/usr/local/share/games/openmw/data
./data
./data

Entering: void ConfigurationManager::processPaths(Files::PathContainer& dataDirs, bool create)
?userdata?data
/home/aesylwinn/.local/share/openmw/data
/home/aesylwinn/.local/share/openmw/data
Adding BSA archive /home/aesylwinn/.openmw/Data Files/Morrowind.bsa
Adding BSA archive /home/aesylwinn/.openmw/Data Files/Tribunal.bsa
Adding BSA archive /home/aesylwinn/.openmw/Data Files/Bloodmoon.bsa
Adding data directory /home/aesylwinn/.openmw/Data Files
Adding data directory /home/aesylwinn/.openmw/NOM
Adding data directory /usr/local/share/games/openmw/data
Adding data directory /home/aesylwinn/.local/share/openmw/data
```

For me, the issue first appears somewhere between commit 548e90a7bc and commit cf1106f3eb. This is a bit puzzling to me, since the only changes between those are changes to CMake. Perhaps it is being caused by a generated file?

AnyOldName3:
I tried checking what happens if I delete the entire contents of the build directory and then cmake/make everything from scratch. It did not change what happened.

Aesylwinn:
It looks like the generated resources/defaultfilters isn't being properly created.

AnyOldName3:
I'm investigating what happens when I roll back to earlier commits, and so far I'm seeing the issue at 548e90a, despite @Aesylwinn not having it in that commit.

Just to double check that we're all on the same page, the issue we're all seeing is the one where you open the CS, choose to create a new addon, select `Morrowind.esm` as the base file, come up with a name for the project file (I've been using `test`), then click create, then it loads `Morrowind.esm` fine, but as soon as it gets to the new file it spits out `Loading failed: ESM Error: End of file while reading record header File: /home/<my username was here>/.local/share/openmw/projects/test.omwaddon.project Record: TES3 Subrecord: Offset: 0x4`

If not, can I please get exact steps to replicate it?

Aesylwinn:
There might be multiple bugs at play here. I believe you are still getting the issue though because the editor doesn't seem to recreate the project file (it really should). Try removing the test.omwaddon.project file before creating it again, or use a different file name.

AnyOldName3:
That could well be why I'm having trouble and also managed to replicate the issue in 0.42.

I've also noticed that the `resources/defaultfilters` is screwed up on Windows, too, so I don't have to use my Ubuntu VM for this. That was an absolute pain because I'm pretty sure my Makefile is screwy - everything seems to be rebuilt when I make instead of just things that have changed or things which depend on things I changed.

AnyOldName3:
The issue seems to be a complicated typo-like issue which CMake didn't tell me about because all functions have variable argument counts even when they don't. I really don't like languages without static type checking.

AnyOldName3:
Fixed it: https://github.com/OpenMW/openmw/pull/1510

Aesylwinn:
Thank you AnyOldName3 for fixing that problem.

@zinnschlag You will either need to merge #1511 or manually delete your project files beforehand though.

PR #1502, opened 2017-10-15
Launcher and wizard ampersand escape support

AnyOldName3:
Before this pull request:

* The launcher wouldn't display .es[pm] files in folders whose path contained an ampersand (and may or may not have deactivated them), even though they would load fine in OpenMW itself.
* The installation wizard would produce an invalid `openmw.cfg` file when selecting an install in a path containing an ampersand character, as discovered by this user: https://forum.openmw.org/viewtopic.php?f=8&t=4676

This PR handles the escaping of quotation marks and ampersands in `data=...` lines more in line with how `boost::program_options` does, meaning the launcher and wizard produce config files which the engine and editor can read correctly and can read these config files correctly themselves.

With this pull request:

* The launcher displays all content files which the engine and editor can actually see, and definitely won't erroneously deactivate any (although I didn't check if it was doing that in the first place anyway).
* The installation wizard produces a valid `openmw.cfg` file when processing an installation with an ampersand in the path.

I'm still salty about the decision to use two different systems to handle interaction with the configuration files, but at least they should have achieved parity now.

I submit this to the mercy of Travis and AppVeyor and hope my offering pleases them more than my last one.

### DIFF
--- a/components/config/gamesettings.cpp
+++ b/components/config/gamesettings.cpp
@@ -55,7 +55,6 @@ void Config::GameSettings::validatePaths()
 
     for (Files::PathContainer::iterator it = dataDirs.begin(); it != dataDirs.end(); ++it) {
         QString path = QString::fromUtf8(it->string().c_str());
-        path.remove(QChar('\"'));
 
         QDir dir(path);
         if (dir.exists())
@@ -64,6 +63,11 @@ void Config::GameSettings::validatePaths()
 
     // Do the same for data-local
     QString local = mSettings.value(QString("data-local"));
+    if (local.at(0) == QChar('\"'))
+    {
+        local.remove(0, 1);
+        local.chop(1);
+    }
 
     if (local.isEmpty())
         return;
@@ -76,7 +80,6 @@ void Config::GameSettings::validatePaths()
 
     if (!dataDirs.empty()) {
         QString path = QString::fromUtf8(dataDirs.front().string().c_str());
-        path.remove(QChar('\"'));
 
         QDir dir(path);
         if (dir.exists())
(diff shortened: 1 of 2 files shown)


AnyOldName3:
Also, this should theoretically make it possible for users to load content from paths that they've somehow managed to get a quotation mark into. On Windows, this isn't an option, and I've not got a quick way of testing such a niche use-case as I've not got my Ubuntu laptop handy.

Aesylwinn:
Just a heads up, this seems to have broken the recognition of tokens like "?userdata?" for both the game and the CS, on linux at least.

Aesylwinn:
The tokens are supposed to be filled in with the relevant path, but they are not.

AnyOldName3:
The function in charge of that is `ConfigurationManager::processPaths(Files::PathContainer& dataDirs, bool create)` in https://github.com/OpenMW/openmw/blob/master/components/files/configurationmanager.cpp and I'm pretty sure that still does what it did before except now it doesn't remove quotes from paths as it shouldn't have to. Nothing I did _should_ have broken it. I'll try and get a dev environment set up in an Ubuntu VM so I can test this.

Just in case something else has somehow broken this, have you done a before and after test with this PR?

Aesylwinn:
> Just in case something else has somehow broken this, have you done a before and after test with this PR?

Yes, it first occurs here. Perhaps when escaping, the token is getting changed to \"datadir\" or something?

AnyOldName3:
It looks like it's broken on Windows, too. It also seems I already had an Ubuntu VM with an OpenMW dev env on it.

AnyOldName3:
It looks like it only broke the `?userdata?` token specifically, and not because of any of the reasons I initially thought. I'm submitting another PR to fix it.

AnyOldName3:
PR is here: https://github.com/OpenMW/openmw/pull/1507

PR #1505, opened 2017-10-16
Make snow create water ripples (fixes #4156)

drummyfish:
Fixes [this](http://bugs.openmw.org/issues/4156) issue - snow now creates water ripples as well as rain, just as in [vanilla](https://youtu.be/07J-MLQ1cVs?t=3m46s). I also moved the rain intensity uniform to the water node as requested by @scrawl.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -502,9 +502,6 @@ namespace MWRender
             mWater->update(dt);
         }
 
-        if (!mSky->isEnabled() || !mSky->hasRain())
-            clearRainRipples();
-
         mCamera->update(dt, paused);
 
         osg::Vec3f focal, cameraPos;
@@ -803,12 +800,6 @@ namespace MWRender
     void RenderingManager::notifyWorldSpaceChanged()
     {
         mEffectManager->clear();
-        mWater->clearRipples();
-    }
-
-    void RenderingManager::clearRainRipples()
-    {
-        mWater->getRainIntensityUniform()->set((float) 0.0);
     }
 
     void RenderingManager::clear()
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -251,12 +251,9 @@ namespace MWRender
         sceneRoot->setNodeMask(Mask_Scene);
         sceneRoot->setName("Scene Root");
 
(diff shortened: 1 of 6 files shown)


apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -506,7 +503,7 @@ namespace MWRender
         }
 
         if (!mSky->isEnabled() || !mSky->hasRain())
```
scrawl: If the weather is snow, this !hasRain() condition will fire and call clearRainRipples(), but things still happen to work because AlphaFader called in the Update phase runs later and resets the uniform to its expected (non zero) value. So from what I can tell, things will work but the code is odd - in its current form it could be reduced to clearRainRipples() every frame and still work. Could we move the clearing to Sky::setEnabled() maybe so that all changes to the uniform happen within sky.cpp?

scrawl:
Can you test starting the game with the water shader disabled, then switch it on? I think the code has a bug that causes the Sky class to not be assigned the uniform in this case.

apps/openmw/mwrender/renderingmanager.cpp
drummyfish: Oh yes I'll change it, I put this condition there because if I `coc`ed into an interior with water, the ripples stayed on, but I don't even know why I had the `!hasRain` there.

drummyfish:
Will test.

drummyfish:
I did some quick changes and will come back tomorrow.

apps/openmw/mwrender/water.cpp
```diff
@@ -418,11 +418,13 @@ Water::Water(osg::Group *parent, osg::Group* sceneRoot, Resource::ResourceSystem
 
     mRainIntensityUniform = NULL;
```
scrawl: Probably want to remove this line now.

scrawl:
@drummyfish FYI: 

> After PR1505 rain does not produce water ripples. The water shader is enabled.
https://bugs.openmw.org/issues/4169

I can confirm this. Turning the shader off then on again makes ripples appear.

drummyfish:
I'm on it. While testing it I found this [bug](https://www.youtube.com/watch?v=leWeThNxaUo&feature=youtu.be), could it be caused by PR #1506?

scrawl:
> While testing it I found this bug, could it be caused by PR #1506?

CC @akortunov

PR #1474, opened 2017-09-22
Implement missing swim animations

akortunov:
Should fix [#2704](https://bugs.openmw.org/issues/2704).

A full list of used animations:
SwimAttack1-3
SwimKnockdown
SwimKnockout
SwimTurnLeft
SwimTurnRight
SwimDeathKnockDown
SwimDeathKnockOut
SwimHit

Maybe I missed something, so any feedback would be helpful.
You can test this PR with horkers (BM_Horker) and any NPC (just type "modcurrentfatigue -10000" in console)

PR #1506, opened 2017-10-17
Use SwimTurnLeft/Right animations correctly

akortunov:
Fixes a bug in PR #1474. Due to wrong "if" statement, these animations were not used.
@scrawl, thanks for spotting that.

Note: If you turn with a very small speed, a player character will be twitchy (due to constant switching between idle and turning animations). Both OpenMW and vanilla Morrowind are affected by this bug. Can we do something with it? Or will we have to treat it as a bug in resource files?

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1874,7 +1874,7 @@ void CharacterController::update(float duration)
                                          : (sneak ? CharState_SneakBack
                                                   : (isrunning ? CharState_RunBack : CharState_WalkBack)));
             }
-            else if(rot.z() != 0.0f && !inwater && !sneak && !(mPtr == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson()))
+            else if(rot.z() != 0.0f && !sneak && !(mPtr == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson()))
             {
                 if(rot.z() > 0.0f)
                     movestate = inwater ? CharState_SwimTurnRight : CharState_TurnRight;


apps/openmw/mwmechanics/character.cpp
```diff
@@ -1874,7 +1874,7 @@ void CharacterController::update(float duration)
                                          : (sneak ? CharState_SneakBack
                                                   : (isrunning ? CharState_RunBack : CharState_WalkBack)));
             }
-            else if(rot.z() != 0.0f && !inwater && !sneak && !(mPtr == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson()))
+            else if(rot.z() != 0.0f && !sneak && !(mPtr == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson()))
```
akortunov: I am not sure about isFirstPerson() check. Is it really needed? There is an only usage of this function in the character manager.

scrawl: Don't recall exactly but the isFirstPerson() was probably added for a reason. Try 'git blame'. Also check for marksman aiming and weapon idle animations - these should not interrupt by turning.

scrawl:
> Note: If you turn with a very small speed, a player character will be twitchy (due to constant switching between idle and turning animations).

If you know a better solution, feel free to propose it. But I think very slow turning will look odd no matter what, when the animation has to slow down that much.

drummyfish:
> Note: If you turn with a very small speed, a player character will be twitchy (due to constant switching between idle and turning animations).

I made this an [issue](http://bugs.openmw.org/issues/4177#change-21071) because vanilla behavior differs from other than idle animations.

PR #1507, opened 2017-10-18
Fix the loading of 'data-local' lines which I broke

AnyOldName3:
Fixes a regression which I caused that @Aesylwinn noticed.


Aesylwinn:
That was fast! This fixes the issue.

PR #1508, opened 2017-10-18
Mark worldimp functions as override to stop distracting Travis CI war…

AnyOldName3:
…nings and prevent future typos or other issues that the override keyword was added to prevent.

I think only the OSX Travis builds emitted these warnings, but they were annoying me.

Alternatively, we could turn off the warning, but there's a reason the warning was implemented in the first place.

### DIFF
--- a/apps/openmw/mwworld/worldimp.hpp
+++ b/apps/openmw/mwworld/worldimp.hpp
@@ -137,7 +137,7 @@ namespace MWWorld
             MWWorld::Ptr getFacedObject(float maxDistance, bool ignorePlayer=true);
 
     public: // FIXME
-            void removeContainerScripts(const Ptr& reference);
+            void removeContainerScripts(const Ptr& reference) override;
     private:
             void addContainerScripts(const Ptr& reference, CellStore* cell);
             void PCDropped (const Ptr& item);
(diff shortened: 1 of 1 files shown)


kcat:
FWIW, I would personally prefer to omit the ```virtual``` keyword when ```override``` is specified. The latter implies the former (```override``` is invalid on a non-```virtual``` function), and it makes for less typing and uses less space.

kcat:
Also, it may be a good idea to mark the implementation class as ```final```, i.e.
```
class World final : public MWBase::World {
...
```
since they're not to be overridden. That will help with compiler optimizations since it will know the implementation's member functions can't be overridden again, giving it more opportunities to avoid calling through the vtable.

scrawl:
> FWIW, I would personally prefer to omit the virtual keyword when override is specified. The latter implies the former (override is invalid on a non-virtual function), and it makes for less typing and uses less space.

That's a good idea. It's also inconsistent to have a bunch of code that uses virtual for overriden functions and some code that doesn't. Using `override` but not `virtual` <i>everywhere</i> should make things a lot more readable.

scrawl:
I'm curious, why is World causing warnings for this when other classes like [MechanicsManager]( https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/mechanicsmanagerimp.hpp) do not?

kcat:
I saw similar warnings for a couple different class methods (which I fixed as part of an earlier PR), and was similarly confused as to why none of the other virtual interface classes were causing warnings. My conclusion was maybe the compiler recognized the virtual interface construct and knew it couldn't be a problem to omit the ```override``` keyword for the implementation (all the base virtual methods were pure (```= 0```), so if they weren't overridden it wouldn't allow the class's instantiation anyway). Perhaps something changed with the ```MWBase::World``` class that the compiler no longer recognizes it as a pure virtual interface and thinks it's worth warning about.

Either way though, it's probably good to be consistent about marking overrides as ```override``` regardless, even if the compiler believes it's no problem to omit the keyword.

AnyOldName3:
I wasn't sure that `MWWorld::World` would definitely not be subclassed ever, so didn't make that kind of change.

I also read somewhere that there was a semantic difference between functions marked virtual and override versus just override, but some more official documentation doesn't seem to imply this, so I'll tweak that.

scrawl:
On some functions there is a leftover space in place of the removed `virtual` so now the indentation is a bit odd.

AnyOldName3:
Good spot. I hadn't noticed that I'd done that.

AnyOldName3:
Also, while I'm changing basically every line in this file anyway, should I make the line breaks consistent? Some lines are split at seemingly arbitrary points and some are long but aren't split.

Would it be better to either keep things on a single line or at least standardise the splits at a single position? I'm more in favour of the first of these as basically every text editor supports word wrap, so people who like a split can get it without disrupting those who don't.

scrawl:
I don't see line breaks as an issue.

PR #1509, opened 2017-10-18
Make loading window non-resizable (fixes #4163)

drummyfish:
Hopefully this two-liner will be my way to editor development as well :) Fixes [this](http://bugs.openmw.org/issues/4163).

### DIFF
--- a/apps/opencs/view/doc/loader.cpp
+++ b/apps/opencs/view/doc/loader.cpp
@@ -22,7 +22,7 @@ CSVDoc::LoadingDocument::LoadingDocument (CSMDoc::Document *document)
     setWindowTitle (QString::fromUtf8((std::string("Opening ") + document->getSavePath().filename().string()).c_str()));
     setWindowFlags (Qt::MSWindowsFixedSizeDialogHint | Qt::Dialog);
 
-    int minimumWidth = 400;
+    const int minimumWidth = 400;
 
     mLayout = new QVBoxLayout (this);
 
--- a/apps/opencs/view/doc/loader.cpp
+++ b/apps/opencs/view/doc/loader.cpp
@@ -20,6 +20,7 @@ CSVDoc::LoadingDocument::LoadingDocument (CSMDoc::Document *document)
 : mDocument (document), mAborted (false), mMessages (0), mTotalRecords (0)
 {
     setWindowTitle (QString::fromUtf8((std::string("Opening ") + document->getSavePath().filename().string()).c_str()));
+    setWindowFlags (Qt::MSWindowsFixedSizeDialogHint);
 
     setMinimumWidth (400);
 
@@ -71,6 +72,7 @@ CSVDoc::LoadingDocument::LoadingDocument (CSMDoc::Document *document)
 
     show();
 
+    setFixedSize (this->size());
     connect (mButtons, SIGNAL (rejected()), this, SLOT (cancel()));
 }
 
--- a/apps/opencs/view/doc/loader.cpp
+++ b/apps/opencs/view/doc/loader.cpp
@@ -20,9 +20,9 @@ CSVDoc::LoadingDocument::LoadingDocument (CSMDoc::Document *document)
 : mDocument (document), mAborted (false), mMessages (0), mTotalRecords (0)
 {
     setWindowTitle (QString::fromUtf8((std::string("Opening ") + document->getSavePath().filename().string()).c_str()));
-    setWindowFlags (Qt::MSWindowsFixedSizeDialogHint);
+    setWindowFlags (Qt::MSWindowsFixedSizeDialogHint | Qt::Dialog);
 
-    setMinimumWidth (400);
+    int minimumWidth = 400;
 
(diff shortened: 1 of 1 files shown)


AnyOldName3:
If this effects the window that I think it does, then you may have to do something a little more complex, as currently the only way to read error messages is to resize that window. Me, @Zini and @Aesylwinn have been busy doing that all day today because I broke something twice and both breakages managed to get into master. The error messages still need to be readable.

If you want to test the error messages, conveniently, if you try and create a new addon file before this PR (https://github.com/OpenMW/openmw/pull/1510) gets merged, an error is produced.

Aesylwinn:
Preventing the user from being able to obtain more information is undesirable. If the window size is going to be fixed, then it needs to resize itself whenever its content does not fit. Both error messages and file names could expand beyond the window borders. You should be able to resize the window, either using signals or manually doing so for the relevant functions.

The issue doesn't seem to have been discussed at all, so I'll just mention that another solution would be to change the layout to better suit a fullscreen window.

drummyfish:
Oh okay, I didn't know the window also displayed error messages. Thank you for the info, I'll look into it.

drummyfish:
Looks good on my OS now, layout resized the window according to content but manual resize is not possible. Someone should also take a look at Windows I guess.

AnyOldName3:
Error messages seem to fit just fine on Windows.

For future reference, I'd recommend that you never make changes to your fork's master branch if you want them to end up upstream. When the PR gets merged, if you try and update your version to match upstream, it'll make a mess.

drummyfish:
You mean it would make conflicts or something? I've been fine so far. Sorry I'm not that skilled with contributing yet :) But I'll make separate branches from now on, it's better anyway.

AnyOldName3:
If your fork's master branch isn't completely up to date when the PR gets merged, it will contain your changes as commits based on a certain revision, but upstream will have the same changes as different commits based on a newer revision. This won't merge as the histories are different, so you have to do messy rewriting of history to make it work.

drummyfish:
That makes sense. I've been using git for quite some time but only for solo project before now, I think I never even used multiple branches :) Thanks for sharing the knowledge.

apps/opencs/view/doc/loader.cpp
```diff
@@ -20,8 +20,9 @@ CSVDoc::LoadingDocument::LoadingDocument (CSMDoc::Document *document)
 : mDocument (document), mAborted (false), mMessages (0), mTotalRecords (0)
 {
     setWindowTitle (QString::fromUtf8((std::string("Opening ") + document->getSavePath().filename().string()).c_str()));
+    setWindowFlags (Qt::MSWindowsFixedSizeDialogHint | Qt::Dialog);
 
-    setMinimumWidth (400);
+    int minimumWidth = 400;
```
Aesylwinn: The mimimumWidth variable should probably be declared as constant.

Aesylwinn:
@zinnschlag Your thoughts on this?

zinnschlag:
I do not see what the point of this change is. From my point of view the loading window works exactly as it is supposed to. I didn't veto the previous fixed size change, because that window was kinda a special case with it fixed layout and (almost) no elements that possibly could change size. But generally all windows being resizable in OpenMW-CS is intentional. Accepting this change would make the loading window inconsistent with any other resizable window and it has no functional benefits.
The aesthetic downsides are marginal IMO. If anything, we maybe should remove the maximise button (if that is doable consistently across all platforms).

zinnschlag:
btw. welcome to the team. Let me know if you need help with finding a task. You seem to find your way around in Qt pretty quickly, so there are options.

drummyfish:
AFAIK these informative pop-ups are usually non-resizable in pretty much any software, resizing is for "workspace" windows. Keeping the GUI behavior consistent with usual behavior helps user-friendliness. But if there is a consensus all windows should be resizable in the editor, I'm OK with it.

zinnschlag:
> AFAIK these informative pop-ups are usually non-resizable in pretty much any software, Keeping the GUI behavior consistent with usual behavior helps user-friendliness.

Consistency across applications is important, but if no one ever improves on consistently used bad ideas we keep being stuck with bad idea.

Would you like to give the no-maximaize button variant a go or can we close this here?

drummyfish:
I think hiding maximize button is just a half-solution that isn't really useful - if it's gonna be resizable, we should keep the resizing buttons. So I'm closing this for now, it's really a minor issue. Maybe I'd like to hear what other people think about this though, so I may ask others at the forums, just to see if I'm looking at this in a wrong way. I'm just curious and have no problems accepting this idea if it's justified. Thank you for the feedback.

AnyOldName3:
There is actually an advantage to this PR over master. At least under Windows, without this, the loading box doesn't expand to fit error messages, but it does with it. That part of this should still be merged in my opinion, regardless of any other complaints.

Also, while someone's already looking at the popup, it would be nice if the error messages could be highlighted so they could be copied and pasted into a search engine, as I'd imagine that's the first thing a user encountering an error is likely to try and do.

PR #1512, opened 2017-10-19
fixed new clothing replacing old clothing of same value

Rhiyo:
Found on the bug http://bugs.openmw.org/issues/4165.

In original Morrowind, new clothing of the same value wouldn't replace old clothing. Tested with common and expensive clothing by selling to merchants and using the AddItem console command in original Morrowind. 

In OpenMW, before this change, new clothing of the same value would replace old clothing, tested with the same methods used above.


Rhiyo:
Also, this is my first time working contributing to a public repository this large, please tell me if I didn't follow the right procedures in contributing (I read the contribute section).

akortunov:
But what about rings? There is a separate check for rings since #1499.

PR #1513, opened 2017-10-19
Feature #4082: use many ingredients/potions instantly

lukago:
Explanation: http://bugs.openmw.org/issues/4082

### DIFF
--- a/apps/openmw/mwgui/inventorywindow.cpp
+++ b/apps/openmw/mwgui/inventorywindow.cpp
@@ -1,5 +1,7 @@
 #include "inventorywindow.hpp"
 
+#include <stdexcept>
+
 #include <MyGUI_Window.h>
 #include <MyGUI_ImageBox.h>
 #include <MyGUI_RenderManager.h>
@@ -225,8 +227,8 @@ namespace MWGui
         std::string sound = item.mBase.getClass().getDownSoundId(item.mBase);
 
         MWWorld::Ptr object = item.mBase;
-        bool shift = MyGUI::InputManager::getInstance().isShiftPressed();
         int count = item.mCount;
+        bool shift = MyGUI::InputManager::getInstance().isShiftPressed();
 
         if (MyGUI::InputManager::getInstance().isControlPressed())
             count = 1;
@@ -525,7 +527,6 @@ namespace MWGui
         if (mDragAndDrop->mIsOnDragAndDrop)
         {
             MWWorld::Ptr ptr = mDragAndDrop->mItem.mBase;
-            int itemType = ptr.getContainerStore()->getType(ptr);
 
             mDragAndDrop->finish();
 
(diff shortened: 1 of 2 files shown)


apps/openmw/mwgui/inventorywindow.cpp
```diff
@@ -228,9 +225,9 @@ namespace MWGui
         std::string sound = item.mBase.getClass().getDownSoundId(item.mBase);
 
         MWWorld::Ptr object = item.mBase;
+        bool shift = MyGUI::InputManager::getInstance().isShiftPressed();
```
AnyOldName3: Why has this line been repositioned? Usually, you want to avoid changing anything you don't have to as it reduces the chance of merge conflicts.

apps/openmw/mwgui/inventorywindow.cpp
```diff
@@ -528,14 +525,26 @@ namespace MWGui
...
 
             if (mDragAndDrop->mSourceModel != mTradeModel)
             {
                 // Move item to the player's inventory
                 ptr = mDragAndDrop->mSourceModel->moveItem(mDragAndDrop->mItem, mDragAndDrop->mDraggedCount, mTradeModel);
             }
+
             useItem(ptr);
+
+            // If item is ingredient or potion don't stop drag and drop to simplify action of taking more than one 1 item
```
scrawl: How about a more generic implementation: if after `useItem` the item's count decreased by one and is greater than zero, restore the drag action. This way, the code doesn't need to be concerned with what is an ingredient or potion and if new actions/item types are added in the future then this will work for those too.

apps/openmw/mwgui/inventorywindow.cpp
```diff
@@ -528,14 +525,26 @@ namespace MWGui
         if (mDragAndDrop->mIsOnDragAndDrop)
         {
             MWWorld::Ptr ptr = mDragAndDrop->mItem.mBase;
+            int itemType = ptr.getContainerStore()->getType(ptr);
```
scrawl: I would have added check against ContainerStore being null to be sure, or just get the type from the MWWorld::Ptr to begin with.

Of course, if we don't check for types anymore and this line gets removed, then this comment is moot.

apps/openmw/mwgui/inventorywindow.cpp
```diff
@@ -528,14 +525,26 @@ namespace MWGui
...
             }
+
             useItem(ptr);
+
+            // If item is ingredient or potion don't stop drag and drop to simplify action of taking more than one 1 item
+            if ((itemType == MWWorld::ContainerStore::Type_Ingredient
+                    || itemType == MWWorld::ContainerStore::Type_Potion)
+                && mDragAndDrop->mDraggedCount > 1)
+            {
+                dragItem (nullptr, mDragAndDrop->mDraggedCount - 1);
```
scrawl: Here it is assumed that mSelectedItem is the item being dragged. I don't think that is the case if the drag started from an external window, i.e. pick an item from a container directly onto the character portrait.

apps/openmw/mwgui/inventorywindow.cpp
lukago: I was editing this method at first, looks like i forgot to revert this line

apps/openmw/mwgui/inventorywindow.cpp
lukago: The question is if we want to have this behaviour for other item types as well. Here: http://bugs.openmw.org/issues/4082 the author of the topic mentioned only ingredients and potions in the topic and last comment. I also think such behaviour for stacked clothes for example would be pointless.

scrawl: > I also think such behaviour for stacked clothes for example would be pointless.

Hm, true. Go ahead then.

apps/openmw/mwgui/inventorywindow.cpp
```diff
@@ -1,7 +1,5 @@
 #include "inventorywindow.hpp"
 
-#include <stdexcept>
```
scrawl: Why did you remove this include? `<stdexcept>` is one that haunts me in particular because apparently some versions of GCC implicitely include `<stdexcept>` from some other header but other versions don't, leading to easy-to-miss errors that only pop up later breaking the build for a subset of platforms.

Used on line 316:
<pre>
throw std::runtime_error("Can't find restacked item");
</pre>

lukago:
reverted remove of `<stdexcept>`

PR #1517, opened 2017-10-22
Updating OSG packages to 3.4.1

ace13:
Note that I've just only gotten the packaging done so far, not had any time to do a test build with these packages yet so would love to hear from anyone if they actually work before merging.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -330,9 +330,9 @@ if [ -z $SKIP_DOWNLOAD ]; then
 		"OpenAL-Soft-1.17.2.zip"
 
 	# OSG
-	download "OpenSceneGraph 3.4.0-scrawl" \
-		"http://www.lysator.liu.se/~ace/OpenMW/deps/OSG-3.4.0-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z" \
-		"OSG-3.4.0-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z"
+	download "OpenSceneGraph 3.4.1-scrawl" \
+		"http://www.lysator.liu.se/~ace/OpenMW/deps/OSG-3.4.1-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z" \
+		"OSG-3.4.1-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z"
 
 	# Qt
 	if [ -z $APPVEYOR ]; then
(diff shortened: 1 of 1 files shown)


scrawl:
Well, the build appears to work. Do we need to test running, too? Would it work for someone to grab one of the current nightly builds & replace the OSG libs with these ones, or did I put some binary incompatibility in my osg fork that prevents that?

PlutonicOverkill:
Of course I had to see this the day after I wasted hours trying to compile my own OSG build :(
Tested it and it seems to work. I no longer get a crash occurring in OSG when selecting terrain in OpenMW-CS, which looks like a good sign, but now I just get a different crash occurring [here](https://github.com/OpenMW/openmw/blob/0c6ef17fb56babe42343bff7a3bc5d4f611b0767/apps/opencs/model/prefs/shortcuteventhandler.cpp#L198) with the error `invalid comparator`. I also get the message `Error: RigGeometry rendering with no skeleton, should have been initialized by UpdateVisitor` appearing in the console. Oh well, I'll keep trying to find the cause of the issue.

Aesylwinn:
Neither of those are problems with osg. The crash is apparently because sort requires strictly weak ordering.

ace13:
@PlutonicOverkill If you want to do your own packages sometime, https://github.com/bjornblissing/osg-3rdparty-cmake is a great help.

Ran through some of my own builds with 3.4.1 now, can't see anything that's horribly wrong but I also run it rather plain so might not be touching all the areas.

PlutonicOverkill:
Thanks @ace13, I've already got OSG built, it's just the linker errors now. Also I spoke too soon, the crash with terrain editing still occurs in the new OSG version.

CI/before_script.msvc.sh
```diff
@@ -330,9 +330,9 @@ if [ -z $SKIP_DOWNLOAD ]; then
 		"OpenAL-Soft-1.17.2.zip"
 
 	# OSG
-	download "OpenSceneGraph 3.4.0-scrawl" \
-		"http://www.lysator.liu.se/~ace/OpenMW/deps/OSG-3.4.0-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z" \
-		"OSG-3.4.0-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z"
+	download "OpenSceneGraph 3.4.1-scrawl" \
+		"http://www.lysator.liu.se/~ace/OpenMW/deps/OSG-3.4.1-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z" \
+		"OSG-3.4.1-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z"
```
scrawl: Using a OSG_VERSION variable here might be nice in the future, so we don't have to change 11 different lines just to bump the version.

scrawl:
@ace13 do I need to merge this to openmw-43 too, or do you use a different script for releases anyway?

scrawl:
Nevermind, I'm going to merge anyway for the other 2 fixes.

PR #1518, opened 2017-10-22
Moving up in water

thegriglat:
Moving up player when idle in water .

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1999,6 +1999,8 @@ void CharacterController::update(float duration)
 
     if (mFloatToSurface && cls.isActor() && cls.getCreatureStats(mPtr).isDead() && cls.canSwim(mPtr))
         moved.z() = 1.0;
+    if (cls.isActor() && cls.canSwim(mPtr) && mPtr == getPlayer() && world->isSwimming(mPtr))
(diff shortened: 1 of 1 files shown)


kcat:
Does vanilla do this?

akortunov:
> Does vanilla do this?

No.
Another drawbacks:
1. An engine handles player differently from other actors.
2. Underwater battles become a bit annoying: the player is still moving up.
3. The player will have to move constantly to solve the Puzzle Canal quest (when you need to drown the player).

Anyway, what's the reason for this change?

thegriglat:
Yes, in vanilla there is no such behaviour.
But I think it more physically based behavior (as in real life), than just "fly" in the water and adds dynamic in water fights.
It will be even more physically if flotation of player depends on his weight. E.g. heavy armored player will sink.
P.S.: Really I don't know the best place to insert corresponding code (as it is a first my effort to change something in openmw code), at the moment it just work.
BTW, it is OK to close PR if you don't see it applicable, it is just idea.

akortunov:
> But I think it more physically based behavior (as in real life)

Do you mean buoyant force? A human with a heavy equipment will sink fastly instead of moving up.
And yes, a Morrowind behaviour is not realistic.
IMO, a realistic physics is a far more complex question than just two lines of code, executed for player only.

PR #1520, opened 2017-10-24
switch to C++14 and more!

psi29a:
switch to C++14, enforce min cmake version to 3.1, prefer Qt5 by default, removed commented code and general cmake cleanup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ os:
  - osx
 osx_image: xcode8.3
 language: cpp
+install: export CXX="g++-5"
 sudo: required
 dist: trusty
 branches:
@@ -24,7 +25,7 @@ addons:
       - llvm-toolchain-precise-3.6
     packages: [
       # Dev
-      clang-3.6, libunshield-dev, libtinyxml-dev,
+      g++-5, clang-3.6, libunshield-dev, libtinyxml-dev,
       # Tests
       libgtest-dev, google-mock,
       # Boost
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ addons:
       # FFmpeg
       libavcodec-dev, libavformat-dev, libavutil-dev, libswscale-dev,
       # Audio & Video
-      libsdl2-dev, libqt4-dev, libopenal-dev,
+      libsdl2-dev, libqt5opengl5-dev, libopenal-dev,
       # The other ones from OpenMW ppa
       libbullet-dev, libswresample-dev, libopenscenegraph-3.4-dev, libmygui-dev
     ]
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ addons:
       # FFmpeg
       libavcodec-dev, libavformat-dev, libavutil-dev, libswscale-dev,
       # Audio & Video
-      libsdl2-dev, libqt5opengl5-dev, libopenal-dev,
+      libsdl2-dev, libqt5opengl5-dev, libqt4-dev, libopenal-dev,
       # The other ones from OpenMW ppa
       libbullet-dev, libswresample-dev, libopenscenegraph-3.4-dev, libmygui-dev
     ]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,17 +1,17 @@
 # Apps and tools
-option(BUILD_OPENMW             "build OpenMW" ON)
-option(BUILD_BSATOOL            "build BSA extractor" ON)
-option(BUILD_ESMTOOL            "build ESM inspector" ON)
-option(BUILD_LAUNCHER           "build Launcher" ON)
-option(BUILD_MWINIIMPORTER      "build MWiniImporter" ON)
-option(BUILD_ESSIMPORTER        "build ESS (Morrowind save game) importer" ON)
-option(BUILD_OPENCS             "build OpenMW Construction Set" ON)
-option(BUILD_WIZARD             "build Installation Wizard" ON)
+option(BUILD_OPENMW "build OpenMW" ON)
+option(BUILD_BSATOOL "build BSA extractor" ON)
+option(BUILD_ESMTOOL "build ESM inspector" ON)
+option(BUILD_LAUNCHER "build Launcher" ON)
+option(BUILD_MWINIIMPORTER "build MWiniImporter" ON)
+option(BUILD_ESSIMPORTER "build ESS (Morrowind save game) importer" ON)
+option(BUILD_OPENCS "build OpenMW Construction Set" ON)
+option(BUILD_WIZARD "build Installation Wizard" ON)
 option(BUILD_WITH_CODE_COVERAGE "Enable code coverage with gconv" OFF)
-option(BUILD_UNITTESTS          "Enable Unittests with Google C++ Unittest" OFF)
-option(BUILD_NIFTEST            "build nif file tester" OFF)
-option(BUILD_MYGUI_PLUGIN       "build MyGUI plugin for OpenMW resources, to use with MyGUI tools" ON)
-option(BUILD_DOCS               "build documentation." OFF )
+option(BUILD_UNITTESTS "Enable Unittests with Google C++ Unittest" OFF)
+option(BUILD_NIFTEST "build nif file tester" OFF)
+option(BUILD_MYGUI_PLUGIN "build MyGUI plugin for OpenMW resources, to use with MyGUI tools" ON)
+option(BUILD_DOCS        "build documentation." OFF )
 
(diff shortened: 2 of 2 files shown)


scrawl:
Too many changes in one, can we split this up to separate PR's?

* Minimum cmake version: while I do like the idea of using the same version across the board, Ubuntu 14.04 (still supported until 2019) only has CMake 2.8. I'd be open to simplify the code to `if (apple) require 3.1 else require 2.8` as a middle ground between what we have now and your PR.
* Qt5: unfortunately, our current OSG+Qt integration does not support threaded rendering with Qt5. Therefore it might be preferable to default to Qt4 until this is resolved?
* Whitespace changes: no thanks.
* C++14: what do we gain from this? Is there a particular feature you wanted to use? I don't like upgrading just for the sake of it. Anyway, this is more of a question for the community to make sure everyone's compiler has support and the CI/dependencies server can be updated. (which sounds like wasted work if we don't have any specific reasons to)
* Removing commented MyGUI plugin: this code is unlikely to make a comeback in its current form so I'm open to removing it. In that case we want to remove everything, the cmake lines and the plugins/ directory.

kcat:
> C++14: what do we gain from this? Is there a particular feature you wanted to use? I don't like upgrading just for the sake of it.

C++14 is largely an iteration on C++11. Very few 'must have' features, but it cleans up C++11. It's a Quality of Life update, plugging some holes, adding some overlooked qualifiers, includes minor but useful features, etc. Unfortunately cppreference.com doesn't seem to have a page detailing all the changes made by a particular standard version, so it's difficult to get an overview of everything it does.

scrawl:
Regarding threaded rendering with Qt5/OSG, pvdk sent me this: https://github.com/anormann1974/osgviewerQt5 . I haven't tested if it works.

zinnschlag:
Considering that I am still on gcc 4.8 c++14 would mean for me to install a new compiler. For what little c++14 offers that is an unreasonable large amount of effort.
I agree with scrawl here. Our current setup works fine and we have little to gain with these upgrades. That looks indeed like upgrading for the sake of upgrading.

psi29a:
This PR came out of the discussion here:
https://forum.openmw.org/viewtopic.php?f=6&t=4719

Where I wanted to go to C++17, but was told to do a baby step first to C++14. The end goal was C++17 so that we can take advantage of filesystem symantics (based on boost) so that we can get rid of more boost in our codebase.

So upgrading for the sake of upgrading isn't what is happening here.

Qt4 is being removed from Ubuntu and Debian, we should follow the trend and make Qt5 default and leave Qt4 for backwards compability. If there are any problems with OpenMW and Qt5, it should be made a priority to fix otherwise there is no more OpenMW for those two distros in the next release.

The cmake and whitespace issues I'll fix up in a bit.

scrawl:
> Where I wanted to go to C++17, but was told to do a baby step first to C++14.

I'm afraid I don't get the 'baby steps' idea. If it turns out that C++17 isn't desirable for us yet (for whatever reason), then the C++14 step was (mostly) a waste of time. But if it turns out we want C++17, then we might as well jump to 17 directly. Doing that 'baby step' in between is just unnecessary work for packagers. Am I missing something here?

> The end goal was C++17 so that we can take advantage of filesystem symantics (based on boost) so that we can get rid of more boost in our codebase.

That is a noble cause I can get behind. I would just remove other uses of Boost first and then consider the switch to C++17 when boost::filesystem is the last thing still binding us to Boost. The practical benefit of replacing boost::filesystem will be low if we're still going to depend on boost for other things. For a start, it would be great to write a replacement for boost::program_options that doesn't use [weird escape characters](https://bugs.openmw.org/issues/3243) and doesn't [precede files in the wrong order](https://bugs.openmw.org/issues/2976).

> If there are any problems with OpenMW and Qt5, it should be made a priority to fix otherwise there is no more OpenMW for those two distros in the next release.

The lack of threading is just a minor performance issue for OpenCS that would be nice to fix, it's not going to prevent us from being available in distros in the meantime.

> Trusty has support for gcc5

Eh, using a PPA called 'test' is not what I'd call supported.

For the record, I don't use Trusty, but I'm sure someone out there does, and it's not our call to say no one should be using it when it's still officially supported by Ubuntu. Providing support for whatever distro's are currently supported is a nice and easy rule.

> The cmake, mygui plugin and whitespace issues I'll fix up in a bit.

A separate PR for these changes would be preferred.

psi29a:
Chris and Ace suggested it, I would have rather gone directly to C++17

Yes, I've been trying to get rid of boost for ages now. We finally have the ability to do so and make a (hopefully) final push and take full advantages in the advancement of C++.

OpenMW packages for Trusty is my problem as a package manager. ;)  I consider that a support issue and not a development issue. I would rather have developers using the latest Debian or Ubuntu releases for development, we shouldn't be held behind because someone wants to development (not just play openmw) on an older distro.

I'll split this up into seperate PRs, keep this PR alive (don't close) and I'll rebase this for one of those commits.

kcat:
FWIW, my comment was to point out that C++17 is probably too new to use. Considering we held off on C++11 for so long because it was too new and only relatively recently adopted it, if C++14 is also too new, C++17 wouldn't be a consideration.

But in terms of C++14, like I said, it's a Quality of Life improvement over C++11. There aren't any big must-have features like C++11 and C++17 add, but that doesn't mean it's not a worthless step to take when it's viable to. As an example, C++14 adds ```std::make_unique```, analogous to ```std::make_shared``` (added in C++11), which safely creates a ```std::unique_ptr``` (the typical way of doing ```std::unique_ptr<Type>(new Type(args...));``` can leak if ```unique_ptr```'s constructor throws). Or adding ```constexpr``` to some functions that lacked it in C++11. Or ```std::integer_sequence``` for generating compile-time integer sequences. Nothing we can't do without, but the lack of them can create small annoyances we may occasionally trip over.

psi29a:
By the way, Ubuntu Trusty supports cmake (2.8) and cmake3 (3.5), so we don't have to worry about Trusty development on that ground. Again, why do we need to bother with this stuff? Update your development environment already!

AnyOldName3:
The weird escape character issue (https://bugs.openmw.org/issues/3243) Scrawl mentioned is actually from `boost::filesystem`, not `boost::program_options`, so actually might go away were we to purge that. It's actually not a terrible escape character to use, though, as backslashed are in WIndows paths

zinnschlag:
C++ 17 is an absolute no go IMO. As kcat said, it is too new. I would guess that none of us are really familiar with this new standard yet and the new pitfalls to inevitably come with it. A project of OpenMW's scope is no place to experiment with new language features and especially not now that we getting in on the home stretch to 1.0.

psi29a:
We can play, it doesn't mean we have to merge. I have no problem holding on to a long-running branch. Would that work? It's what I did for C++11.

zinnschlag:
Sure, play is fine as long as it doesn't negatively affect our progress.

psi29a:
@zinnschlag that's the whole point, we want to help us move forward, not create headaches for developers and users. :)

BTW, c++17 is moving along... but shit it's a lot of work replacing all the boost::filesystem crap... I've had to dig into some bits that it did for us automatically that got split out into others bits like std::if/ofstream and std::chrono and other things... I've been compiling and testing along the way looking for regressions, but things look great and compile times are coming down.

scrawl:
Out of curiosity, how does C++ filesystem handle the dreaded 'utf8 paths on windows' issue? Is there a solution similar to [boost::path::imbue](https://github.com/OpenMW/openmw/blob/7a389f7ab117acad15d813a33f63877f45da98ef/components/files/windowspath.cpp#L35) or is this handled in some easier way even?

psi29a:
I don't have a windows environment to test, but I did do some digging. It turns out that there was a work-around using C++11 instead of Boost's imbue however this was deprecated in C++17, though I'm unsure if they just replaced it or baked it into std::filesystem

Here is how you handle utf8 in std::filesystem::path via std::filesystem::u8path
http://rextester.com/PXRH65151

Original post:
https://stackoverflow.com/questions/45572209/getting-a-boostfilesystempath-as-an-utf-8-encoded-stdstring-on-windows

Additional information about deprecation:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0618r0.html

Is there something I am missing here?

scrawl:
That's the same info I found... weird thing to deprecate a function before the actual replacement is available.

I'll close it for the misc changes have been split off and the C++ change is continued in #1526

psi29a:
ok, thanks :)

psi29a:
PS: I still think moving to C++14 is the right decision since it is stable and gains us many additional features.

PR #1521, opened 2017-10-24
Windows: Update CPack configuration for NSIS packages

ace13:
Now grabs generated files from the configuration directory where they're generated, instead of trying to read them directly from the build directory.


ace13:
Apparently I've lost my ability with git, going to have to rebase this properly

PR #1525, opened 2017-10-25
make Qt5 the default instead of 4

psi29a:
In reference to https://github.com/OpenMW/openmw/pull/1520

This does not remove Qt4 support, but makes Qt5 default because Ubuntu and Debian are purging Qt4 from future releases. Other distros are following.

The only issue preventing us from removing Qt4 support entirely is a slight performance issue with OpenMW-CS that involves threading. There are possible workarounds, but that is for another PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ addons:
       # FFmpeg
       libavcodec-dev, libavformat-dev, libavutil-dev, libswscale-dev,
       # Audio & Video
-      libsdl2-dev, libqt4-dev, libopenal-dev,
+      libsdl2-dev, libqt5opengl5-dev, libopenal-dev,
       # The other ones from OpenMW ppa
       libbullet-dev, libswresample-dev, libopenscenegraph-3.4-dev, libmygui-dev
     ]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ else()
 endif()
 
 if (USE_QT)
-    set(DESIRED_QT_VERSION 4 CACHE STRING "The QT version OpenMW should use (4 or 5)")
+    set(DESIRED_QT_VERSION 5 CACHE STRING "The QT version OpenMW should use (4 or 5)")
     set_property(CACHE DESIRED_QT_VERSION PROPERTY STRINGS 4 5)
 endif()
 


scrawl:
In my opinion, the default choice should be the one that provides the best functionality and is well tested, not necessarily the 'most modern' version.

If there are issues that some distro's don't ship qt4, we could make it so the script automatically choses the available Qt version if the default choice isn't available. How about that?

> removing Qt4 support

Woah there, getting ahead of ourselves? Supporting Qt4 isn't much effort nor harmful, I don't think.

psi29a:
If someone is willing to maintain the support, that is fine. I'm of the opinion that we focus on only the libraries and versions we need instead of multiple. In this case the world is moving on and away from Qt4.

I'm ok with defaulting to the one found, just need to figure out how best to do that. Prefer Qt5 over Qt4... even if both are found on a system.

zinnschlag:
> I'm ok with defaulting to the one found, just need to figure out how best to do that. Prefer Qt5 over Qt4... even if both are found on a system.

That is problematic. Is there actually still a distribution around that doesn't have some kind of Qt5? Probably not. So that would defeat the purpose of this test. If anything we would need a "prefer Qt4 over Qt5 if both are present" function.

psi29a:
Why? Shouldn't we prefer the one we want to target and not just support?

zinnschlag:
I still consider Qt4 our main target. That's definitely still the version I will be using for the time being.

To quote scrawl:
> In my opinion, the default choice should be the one that provides the best functionality and is well tested, not necessarily the 'most modern' version.

There is also the issue that users who don't have the most current distribution will have older versions of Qt5 that are even less well tested and have a higher chance to cause problems with OpenMW-CS.

IMO the removal of Qt4 on some of the distros is a mistake. Qt5 simply hasn't been mature enough for a long enough time for such a step.

psi29a:
Thankfully we're not removing Qt4, at least not in this PR.

That downstream is going to drop Qt4 is not something we can control, but that we can improve Qt5 support is something we can control.

zinnschlag:
Just to make sure that there is no misunderstanding:

> IMO the removal of Qt4 on some of the distros is a mistake.

Here I meant what the distros are doing, not the OpenMW packagers.

>  but that we can improve Qt5 support is something we can control.

Sure. Eventually we will make the move to Qt5, so any improvements to the Qt5 compatibility are welcome. Its just that it is too early to make Qt5 our main focus.

scrawl:
Closing as 'no action needed', there can be a new PR to automatically pick the found version of Qt.

psi29a:
... erm, why can't I just update this PR to reflect that? What is the reason to close, I was still working on it. ;)

PR #1526, opened 2017-10-25
C++17 [DO NOT MERGE YET; Needs MacOS support]

psi29a:
Let's see what works and what doesn't...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 os:
  - linux
  - osx
-osx_image: xcode9.0
+osx_image: xcode9.1
 language: cpp
 install: export CXX="g++-7"
 sudo: required
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 os:
  - linux
  - osx
-osx_image: xcode9
+osx_image: xcode9.1
 language: cpp
 install: export c
 sudo: required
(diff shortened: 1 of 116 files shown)


psi29a:
Win... C++17 is ready to go.

Going to experiment now with some de-boosting.

psi29a:
@ace13 would you be able to help get msvc2017 up and running? I would love to ditch as much boost as possible, that includes boost::any which has a std::any replacement, but is not available in msvc2015...

psi29a:
@nikolaykasyanov can you take a look at the osx side? I tried bumping it to use xcode9.1 but it complains that clang is 'broken' and it can't compile. Is there something I'm doing wrong?

scrawl:
Just to make it clear, there are no plans of merging C++17 any time soon (see #1520), and anyone working on this is doing so out of their own curiosity.

psi29a:
What @scrawl said, this is just to figure things and see what it gains us. Many of the functions are still in the std::experiment namespace.

zinnschlag:
Technically this is a post-1.0 PR too, right?

psi29a:
This is me playing around, hence the do not merge yet.  Not so much post-1.0 (unless that is just right around corner), but it can't be merged until supported by macOS. Believe or not, GCC/Clang/MSVC all support this... but macOS does not and no one knows when they will add std::filesystem (let alone std::experimental::filesystem into XCode/Clang.

PR #1529, opened 2017-10-28
Stop character landing animation when turning (issue #4177)

drummyfish:
Attempt at solving issue [#4177](http://bugs.openmw.org/issues/4177) I opened. Should I apply this to other anims such as attack too?

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -372,29 +372,28 @@ void CharacterController::refreshJumpAnims(const WeaponInfo* weap, JumpingState 
             }
         }
 
-        if(mJumpState == JumpState_InAir)
+        if (!mCurrentJump.empty())
         {
             mAnimation->disable(mCurrentJump);
-            mCurrentJump = jumpAnimName;
-            if (mAnimation->hasAnimation("jump"))
-                mAnimation->play(mCurrentJump, Priority_Jump, jumpmask, false,
+            mCurrentJump.clear();
+        }
+
+        if(mJumpState == JumpState_InAir)
+        {
+            if (mAnimation->hasAnimation(jumpAnimName))
+            {
+                mAnimation->play(jumpAnimName, Priority_Jump, jumpmask, false,
                              1.0f, (startAtLoop?"loop start":"start"), "stop", 0.0f, ~0ul);
+                mCurrentJump = jumpAnimName;
+            }
         }
         else if (mJumpState == JumpState_Landing)
         {
-            if (startAtLoop) 
-                mAnimation->disable(mCurrentJump);
-
-            if (mAnimation->hasAnimation("jump"))
+             if (mAnimation->hasAnimation(jumpAnimName))
+             {
                 mAnimation->play(jumpAnimName, Priority_Jump, jumpmask, true,
                              1.0f, "loop stop", "stop", 0.0f, 0);
-        }
-        else    // JumpState_None
-        {
-            if (mCurrentJump.length() > 0)
-            {
-                mAnimation->disable(mCurrentJump);
-                mCurrentJump.clear();
+                mCurrentJump = jumpAnimName;
             }
         }
     }
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -380,7 +380,7 @@ void CharacterController::refreshJumpAnims(const WeaponInfo* weap, JumpingState 
                 mAnimation->play(mCurrentJump, Priority_Jump, jumpmask, false,
                              1.0f, (startAtLoop?"loop start":"start"), "stop", 0.0f, ~0ul);
         }
-        else
+        else if (mJumpState == JumpState_Landing)
         {
             if (startAtLoop) 
                 mAnimation->disable(mCurrentJump);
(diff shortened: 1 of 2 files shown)


scrawl:
1. mJumpAnimName is the same as mCurrentJump, isn't it?
2. Can you place this code closer to the rest of the jump handling code?

drummyfish:
mCurrentJump didn't work as it gets cleared after landing. I don't know if it *has* to be cleared, just didn't want to break anything.

> Can you place this code closer to the rest of the jump handling code, or maybe to the code that starts the turning animation?

I'll try plus I'll look at the other animations in vanilla too.

drummyfish:
Regarding the attack animation, vanilla does flicker as well as OpenMW, so should we leave it like that? I think letting the attack animation play when turning would look better.

scrawl:
> mCurrentJump didn't work as it gets cleared after landing. I don't know if it has to be cleared, just didn't want to break anything.

It doesn't need to be cleared, I don't think. Others like mCurrentWeapon aren't cleared in this way (when the animation naturally ends), either.

What is the issue with attack animations precisely?

drummyfish:
> What is the issue with attack animations precisely?

The attack animation's flickering when slowly turning at the same time, [here's](https://youtu.be/M4LAlUgEVTk) a video. Although playing frame by frame now it looks to me like it's not flickering into turning animation so it may be issue that lies elsewhere in the code.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -382,8 +382,9 @@ void CharacterController::refreshJumpAnims(const WeaponInfo* weap, JumpingState
         }
         else
         {
-            mAnimation->disable(mCurrentJump);
-            mCurrentJump.clear();
+            if (startAtLoop)
```
scrawl: I don't quite understand why the `if (startAtLoop)` was added.

Also, aren't we missing an `mCurrentJump = jumpAnimName` here?

drummyfish: `mCurrentJump = jumpAnimName` was never (it was cleared instead) there so I just kept it that way as I only wanted to fix the issue at hand. Actually the clearing of `mCurrentJump` had its purpose (see below).

`if (startAtLoop)` is there to only disable the current animation at the frame the landing animation is to be started. Previously this was achieved by clearing `mCurrentJump` (so that the next updates call disable on empty string animation) but as I prevent it from being cleared now, I have to test it this way. Without this test the landing animation gets stopped in the next update.

scrawl: Hm, the current code is clearly a little odd. Seemingly the JumpState_Landing only gets set for one frame and then reset to JumpState_None despite the landing animation still playing. And then we call play() with an empty animation name. 

May I suggest try to refactor the code so the states are set properly and mCurrentJump always refers to currently played animation?

scrawl:
> The attack animation's flickering when slowly turning at the same time, here's a video. Although playing frame by frame now it looks to me like it's not flickering into turning animation so it may be issue that lies elsewhere in the code.

It probably is the turning animation in one way or another. IIRC with other races the effect is less pronounced so it might just be an issue with the animations we can't do anything about.

apps/openmw/mwmechanics/character.cpp
drummyfish: >  Seemingly the JumpState_Landing only gets set for one frame

Yes it does. Okay, I'll try to refactor this, at least I'll get more understanding of the animation system.

drummyfish:
So `mJumpState` and `mCurrentJump` are reflecting the landing animation now, you can try printing them in the update method. I really hope I'm not breaking anything else though - if anyone's reading this, I'll be glad if you help me test.

drummyfish:
I think it's done, someone throw at me some criticism :)

scrawl:
I've made a few more edits which removes the now unnecessary `if (startAtLoop)` and improves consistency. How does this work for you?

Additionally I've wondered if non-turning movement (e.g. run forward) should also interrupt the landing animation. This is harder to see than with turning, but when I slowed down the time to test I was able to see the same effect if you just tap the movement key, the switching between landing and movement animation looks odd. What do you think?

<pre>
diff --git a/apps/openmw/mwmechanics/character.cpp b/apps/openmw/mwmechanics/character.cpp
index 2e821b4..4cef8cb 100644
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -372,29 +372,28 @@ void CharacterController::refreshJumpAnims(const WeaponInfo* weap, JumpingState
             }
         }
 
-        if(mJumpState == JumpState_InAir)
+        if (!mCurrentJump.empty())
         {
             mAnimation->disable(mCurrentJump);
-            mCurrentJump = jumpAnimName;
-            if (mAnimation->hasAnimation("jump"))
-                mAnimation->play(mCurrentJump, Priority_Jump, jumpmask, false,
+            mCurrentJump.clear();
+        }
+
+        if(mJumpState == JumpState_InAir)
+        {
+            if (mAnimation->hasAnimation(jumpAnimName))
+            {
+                mAnimation->play(jumpAnimName, Priority_Jump, jumpmask, false,
                              1.0f, (startAtLoop?"loop start":"start"), "stop", 0.0f, ~0ul);
+                mCurrentJump = jumpAnimName;
+            }
         }
         else if (mJumpState == JumpState_Landing)
         {
-            if (startAtLoop)
-                mAnimation->disable(mCurrentJump);
-
-            if (mAnimation->hasAnimation("jump"))
+            if (mAnimation->hasAnimation(jumpAnimName))
+            {
                 mAnimation->play(jumpAnimName, Priority_Jump, jumpmask, true,
                              1.0f, "loop stop", "stop", 0.0f, 0);
-        }
-        else    // JumpState_None
-        {
-            if (mCurrentJump.length() > 0)
-            {
-                mAnimation->disable(mCurrentJump);
-                mCurrentJump.clear();
+                mCurrentJump = jumpAnimName;
             }
         }
     }
</pre>

drummyfish:
I can only notice the run/land animation problem when I slow it down, it's really a very minor issue, but I'd really like to fix the attack/turn animation because that one is kinda annoying and so we might as well both of these issues. Or maybe there is even more of them, I'll take a closer look.

PR #1530, opened 2017-10-29
Trigger OnActivate when item is picked up through the inventory (Issue #3726)

rexelion:
Trying to solve issue #3726. Right now, when you click on a cursed item nothing happens until you close the inventory. After that OnActivate is triggered and the item is added to the player (because of the script, if I understand correctly). This happens every time you try to pick it up by dragging.

### DIFF
--- a/apps/openmw/mwworld/actiontake.cpp
+++ b/apps/openmw/mwworld/actiontake.cpp
@@ -14,6 +14,9 @@ namespace MWWorld
 
     void ActionTake::executeImp (const Ptr& actor)
     {
+        //No need to do anything if moving items from the player's inventory back into player's inventory
+        if (actor == MWBase::Environment::get().getWorld()->getPlayerPtr() && getTarget().getCell() == 0)
+            return;
         MWBase::Environment::get().getMechanicsManager()->itemTaken(
                     actor, getTarget(), MWWorld::Ptr(), getTarget().getRefData().getCount());
         MWWorld::Ptr newitem = *actor.getClass().getContainerStore (actor).add (getTarget(), getTarget().getRefData().getCount(), actor);
--- a/apps/openmw/mwworld/actiontake.cpp
+++ b/apps/openmw/mwworld/actiontake.cpp
@@ -14,9 +14,6 @@ namespace MWWorld
 
     void ActionTake::executeImp (const Ptr& actor)
     {
-        //No need to do anything if moving items from the player's inventory back into player's inventory
-        if (actor == MWBase::Environment::get().getWorld()->getPlayerPtr() && getTarget().getCell() == 0)
-            return;
         MWBase::Environment::get().getMechanicsManager()->itemTaken(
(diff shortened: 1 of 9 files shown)


scrawl:
We can't just check the script text for keywords to take that as the function being used. The function might be in a if condition that's never even triggered.

rexelion:
How about checking Flag_SuppressActivate? Seems like it can only be set to true by OnActivate.

scrawl:
Yes, that's better.

rexelion:
I found a way to make it so that you can still drag the item into your inventory, instead of it just teleporting, but it creates a duplicate of the item: first, the item is moved from the world into the inventory, and after that the item inside the inventory is activated by the script, which triggers ActionTake. It adds a new object to the player's inventory, but doesn't delete the original. Can be fixed by adding a check inside ActionTake::executeImp to see if the source and destination locations are the same, in which case it should do nothing.

scrawl:
Is the 'activate on closing inventory' thing really something the engine itself does? Might also be a result of a statement like `if (MenuMode == 1) return` in the item's script.

rexelion:
I made it so because someone on the issue tracker (issue #1428) said that this is how vanilla Morrowind does it. I've just tested it, and in my game you can still pick up an item through the inventory, and activation happens as soon as you start dragging it.

scrawl:
Okay. Are you going to update the PR so it activates immediately, then?

rexelion:
Yes, I'll do that.

apps/openmw/mwworld/actiontake.cpp
```diff
@@ -14,6 +14,9 @@ namespace MWWorld
 
     void ActionTake::executeImp (const Ptr& actor)
     {
+        //No need to do anything if moving items from the player's inventory back into player's inventory
+        if (actor == MWBase::Environment::get().getWorld()->getPlayerPtr() && getTarget().getCell() == 0)
```
scrawl: I have no idea why this check was added, and it doesn't look correct to me either.

scrawl:
Hm, this seems wrong to me in that items which are not intended to be picked up (i.e. the script calls onActivate, but doesn't call 'activate' afterwards once onActivate is set) can incorrectly be picked up through the inventory mode. Another example you may test with is the Bitter Cup, which shows a message box with 2 choices when you activate it.

Might I suggest:
1. If calling the item's activate() succeeds (i.e. there's no supressActivate flag), pick it up as normal (and as before this PR).
2. Otherwise, do nothing. The script will respond to the activation in the next frame. Because we don't know what the script will do, there's no easy way (or no way?) to start a drag-and-drop action.

scrawl:
> Because we don't know what the script will do, there's no easy way (or no way?) to start a drag-and-drop action.

Actually, we could, if we added some code to 'ActionTake' so that if the inventory mode is open, we start a new drag for the taken item.

PR #1532, opened 2017-10-30
Create LICENSE

drummyfish:
Let's add the license file so that GitHub officially registers it and displays it next to the project.


scrawl:
We already have `docs/license/*.txt` and it's probably not a good idea to have two copies of the same file.

How about moving docs/license/GPL3.txt to LICENSE , then move docs/license/Deja Vu Font License into files/mygui where the font file is?

drummyfish:
In the diffs I'm seeing we don't have some stuff filled in in the license file. Should I do this?

- *{one line to give the program's name and a brief idea of what it does.}* - OpenMW, free and open-source Morrowind engine reimplementation
-

drummyfish:
Accidentally clicked close, sorry :D

LICENSE
```diff
@@ -631,8 +631,8 @@ to attach them to the start of each source file to most effectively
 state the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
-    <one line to give the program's name and a brief idea of what it does.>
-    Copyright (C) <year>  <name of author>
+    {one line to give the program's name and a brief idea of what it does.}
```
drummyfish: Should I fill this in? How about *OpenMW, free and open-source Morrowind engine reimplementation*

LICENSE
```diff
@@ -631,8 +631,8 @@ to attach them to the start of each source file to most effectively
...
+    Copyright (C) {year}  {name of author}
```
drummyfish: What about this? Is it the year the project started? What about the author? Maybe the person who started the project plus mention that additional authors are in AUTHORS.md?

drummyfish: [Here](https://www.gnu.org/licenses/gpl-howto.en.html) they say we should add year for each release and it's okay to use a range, plus we should mention all people who contributed code, so maybe:

*2008-2017, Nicolay Korslund and others (see AUTHORS.md)*?

LICENSE
drummyfish: Or better yet *engine **and editor***.

drummyfish: I feel like a lawyer now.

LICENSE
scrawl: That's fine, I would just leave out the link to AUTHORS.md so the license file is valid on its own if someone forgets to package authors.md.

scrawl: "Nicolay Korslund and Contributors" maybe.

LICENSE
scrawl: I'll give @psi29a a ping since pretending to be a lawyer is one of his expertises ;)

scrawl: Your description works for me.

scrawl:
There are some references to the license files in CMakeLists.txt that need to be updated. Also, on Windows platforms we might want to rename the file to '.txt' when the package is created because Windows doesn't recognize it as a text file otherwise.

drummyfish:
I could only test cmake on Linux but the changes concern Windows, so please someone try cmake on Win.

scrawl:
I've just had a closer read and I don't think the placeholders were supposed to be filled in at all. They're <i>under</i> the 'End of terms and conditions' where some guidance on how to apply the GPL to new programs (not OpenMW) is given. Can you revert that part?

PR #1535, opened 2017-11-01
Korean encoding added

netnk9151:
With these, we can play openmw in Korean with Korean plugins.


kcat:
Presuming Korean wasn't an officially supported language, wouldn't it be better to add a UTF-8 encoding to handle every other currently unhandled language, instead of adding individual codepages?

netnk9151:
kcat// This is not my work. A friend of mine did it and i'm not able to get his help right now.

zinnschlag:
> wouldn't it be better to add a UTF-8 encoding to handle every other currently unhandled language

Maybe I am missing something here, but what do you mean by that? What would have to be added? In principle we should already support Korean (minus missing characters in fonts maybe).

kcat:
> Maybe I am missing something here, but what do you mean by that? What would have to be added?

As far as I understand, the text from the esm/esp files (for dialog, books, messages, etc) is interpreted according to one of the few codepage you set in the config file. So if you specify the Latin-1 codepage for example, all the text from the game files is interpreted as Latin-1 and you can't write any text that's not in the Latin-1 character set. We support the same codepages that the official translations support, but that's it. If someone wants to make a mod to unofficially translate the game to Korean or Japanese or something (or make a new game in said language), it can't work since we don't support any codepages that those character sets need.

This pull request adds a new option for a Korean-specific codepage so that Korean text can be recognized and interpreted correctly. However, rather than adding support for more individual codepages for specific languages, we could instead just add an option for a UTF-8 "codepage" to handle everything Unicode can. Fan-made translations (or new games) can then work with any language without us having to specifically support it.

akortunov:
> we could instead just add an option for a UTF-8 "codepage" to handle everything Unicode can

How can we achieve this? IIRC, game files use some national encodings (cp1251, cp1252, cp949, etc.).
Now we have to convert the text from such encoding to the UTF-8 to use it properly.

kcat:
The current codepages would stay as they are. But if someone were to provide a translation mod, they could say "Set the codepage to utf8". In the future, we could add a way for the individual omwgame/omwaddon files to specify their own text encoding (internally we convert it all to Unicode anyway AFAIK, this just affects how to interpret the game files' text to do the conversion).

zinnschlag:
> As far as I understand, the text from the esm/esp files (for dialog, books, messages, etc) is interpreted according to one of the few codepage you set in the config file. So if you specify the Latin-1 codepage for example, all the text from the game files is interpreted as Latin-1 and you can't write any text that's not in the Latin-1 character set. We support the same codepages that the official translations support, but that's it. If someone wants to make a mod to unofficially translate the game to Korean or Japanese or something (or make a new game in said language), it can't work since we don't support any codepages that those character sets need.

Yes, but that is only true for esm/esp files. These would have to be created with the original TES-CS. I would be somewhat surprised if that can handle Korean.

OpenMW-CS can only write the new file formats and text in omwgame and omwaddon files is UTF8 anyway, so IMO there is no point in adding code for Korean legacy encodings.

zinnschlag:
Actually, I am only 99% sure that the new file formats use UTF-8. Need to check later to make sure I did not misremember that.

kcat:
It might be that OpenMW-CS outputs omwgame/omwaddon files using UTF-8 (I don't know), but it's also necessary for OpenMW itself to read omwgame/omwaddon files as UTF-8, rather than converting using the codepage like it has to for esm/esp files. As far as I know, OpenMW makes no distinction between omwgame/omwaddon and esm/esp files in regards to the text encoding.

zinnschlag:
Code looks messy. There are comments in there that indicate that it is doing the right thing, but the code itself doesn't seem to agree with that.
Guess we will have to have another pass at that. Anyway, using UTF-8 without any legacy encodings is the intended way of supporting additional languages. Therefore I am closing this PR.

PR #1536, opened 2017-11-01
Wait a few seconds before getting up (fixes #3884)

rexelion:
When characters are knocked down, they wait a few seconds before getting up, even if their fatigue raises above 0. Gives the knockdown animation time to finish.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -248,15 +248,14 @@ void CharacterController::refreshHitRecoilAnims()
     bool knockdown = mPtr.getClass().getCreatureStats(mPtr).getKnockedDown();
     bool block = mPtr.getClass().getCreatureStats(mPtr).getBlock();
     bool isSwimming = MWBase::Environment::get().getWorld()->isSwimming(mPtr);
-    MWWorld::TimeStamp currentTime = MWBase::Environment::get().getWorld()->getTimeStamp();
-    float timeScale = MWBase::Environment::get().getWorld()->getTimeScaleFactor();
     if(mHitState == CharState_None)
     {
         if ((mPtr.getClass().getCreatureStats(mPtr).getFatigue().getCurrent() < 0
                 || mPtr.getClass().getCreatureStats(mPtr).getFatigue().getBase() == 0)
                 && mAnimation->hasAnimation("knockout"))
         {
-            mKnockoutTime = MWBase::Environment::get().getWorld()->getTimeStamp();
+            mTimeToWake = time(NULL);
+            mTimeToWake += rand() % 2 + 1; // Wake up after 1 to 3 seconds
             if (isSwimming && mAnimation->hasAnimation("swimknockout"))
             {
                 mHitState = CharState_SwimKnockOut;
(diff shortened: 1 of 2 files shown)


scrawl:
It's almost always preferred to use real time, not game time. For example, if someone, for whatever reason, turns back the clock while someone is knocked down, the knock down will last much longer. That doesn't make sense.

scrawl:
Also, could you add the randomization aspect? According to a bug commenter it's 2-3 seconds, in my experience it was 1-2 seconds. We could settle on 1-3?

apps/openmw/mwmechanics/character.cpp
```diff
@@ -254,6 +254,8 @@ void CharacterController::refreshHitRecoilAnims()
                 || mPtr.getClass().getCreatureStats(mPtr).getFatigue().getBase() == 0)
                 && mAnimation->hasAnimation("knockout"))
         {
+            mTimeToWake = time(NULL);
+            mTimeToWake += rand() % 2 + 1; // Wake up after 1 to 3 seconds
```
PlutonicOverkill: This gives a time of 1-2 seconds, not 1-3. Also I'm not sure that the low-order bits of rand() are very good for generating random numbers.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -255,7 +255,7 @@ void CharacterController::refreshHitRecoilAnims()
                 && mAnimation->hasAnimation("knockout"))
         {
             mTimeToWake = time(NULL);
-            mTimeToWake += rand() % 2 + 1; // Wake up after 1 to 3 seconds
+            mTimeToWake += (int)( (float)rand() / (float)RAND_MAX * 2) + 1; // Wake up after 1 to 3 seconds
```
PlutonicOverkill: This still has not fixed the problem. This expression only gives 3 when rand() == RAND_MAX, which is an extremely unlikely occurrence. You need to obtain a random number in the range [0,2] and add 1 to it to get the desired result. Also see [here](http://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution) for random integers in C++11.

kcat:
OpenMW already has RNG functions in ```components/misc/rng.hpp```. ```Misc::Rng::rollDice(3)+1``` would get you a value between 1 and 3 inclusive.

akortunov:
> Also, could you add the randomization aspect? According to a bug commenter it's 2-3 seconds, in my experience it was 1-2 seconds

I guess that Morrowind just updates knockout state for all actors every 3 seconds.
You can modify NPC fatigue via console:
`modcurrentfatigue -1000`
And NPC will stay a couple of seconds, then will lay prone.

rexelion:
I've just tested it, in vanilla Morrowind actors don't get up as long as their fatigue is negative.

rexelion:
In my case, in vanilla Morrowind actors get knocked out as soon as I use the command.

rexelion:
I don't think the performance impact is that big, it is initialised only during the frame a character gets knocked out, but I change it to rollDice and modify Rng to use functions from <random>.

scrawl:
1. Please use the existing `Misc::Rng`, for consistency if nothing else.
2. Don't use `time()` (sorry, my advice to use 'real time' was too literal). You have to add the `duration` whenever `CharacterController::update` is called or else time will still run while the game is paused.

scrawl:
That looks good, I just have two minor style issues:
* mTimeUntilWake should be initialized in the constructor, just in case.
* I see no reason to make `updateKnockoutTimer` its own function as just a single statement.

PR #1537, opened 2017-11-02
Check the real distance to target in AIPursue (issue #4184)

drummyfish:
Should be fixing [this](http://bugs.openmw.org/issues/4184#change-21169).

### DIFF
--- a/apps/openmw/mwmechanics/aipursue.cpp
+++ b/apps/openmw/mwmechanics/aipursue.cpp
@@ -55,7 +55,7 @@ bool AiPursue::execute (const MWWorld::Ptr& actor, CharacterController& characte
     float pathTolerance = 100.0;
 
     if (pathTo(actor, dest, duration, pathTolerance) &&
-        MWMechanics::distance(dest,aPos.pos[0],aPos.pos[1],aPos.pos[2]) < pathTolerance)  // check the true distance in case the target is far away in Z-direction
+        abs(dest.mZ - aPos.pos[2]) < pathTolerance)      // check the true distance in case the target is far away in Z-direction
     {
         target.getClass().activate(target,actor).get()->execute(actor); //Arrest player when reached
         return true;
--- a/apps/openmw/mwmechanics/aipursue.cpp
+++ b/apps/openmw/mwmechanics/aipursue.cpp
@@ -50,8 +50,13 @@ bool AiPursue::execute (const MWWorld::Ptr& actor, CharacterController& characte
 
     //Set the target desition from the actor
     ESM::Pathgrid::Point dest = target.getRefData().getPosition().pos;
+    ESM::Position aPos = actor.getRefData().getPosition();
 
-    if (pathTo(actor, dest, duration, 100)) {
+    float pathTolerance = 100.0;
+
+    if (pathTo(actor, dest, duration, pathTolerance) &&
+        MWMechanics::distance(dest,aPos.pos[0],aPos.pos[1],aPos.pos[2]) < pathTolerance)  // check the true distance in case the target is far away in Z-direction
+    {
         target.getClass().activate(target,actor).get()->execute(actor); //Arrest player when reached
         return true;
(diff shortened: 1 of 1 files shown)


apps/openmw/mwmechanics/aipursue.cpp
```diff
@@ -50,8 +50,13 @@ bool AiPursue::execute (const MWWorld::Ptr& actor, CharacterController& characte
 
     //Set the target desition from the actor
     ESM::Pathgrid::Point dest = target.getRefData().getPosition().pos;
+    ESM::Position aPos = actor.getRefData().getPosition();
 
-    if (pathTo(actor, dest, duration, 100)) {
+    float dist = MWMechanics::distance(dest,aPos.pos[0],aPos.pos[1],aPos.pos[2]);
+
+    if (pathTo(actor, dest, duration, 100) &&
```
drummyfish: Tried using `MWMechanics::PathFinder::PathTolerance` but it was too small, so I set it to 100.

apps/openmw/mwmechanics/aipursue.cpp
```diff
@@ -50,8 +50,13 @@ bool AiPursue::execute (const MWWorld::Ptr& actor, CharacterController& characte
...
+        dist < 100.0)     // check the true distance in case the target is far away in Z direction
```
drummyfish: Tried using `MWMechanics::PathFinder::PathTolerance` but it was too small, so I set it to 100.

akortunov: Can we use an activation distance instead (192 by default)?

MWBase::Environment::get().getWorld()->getMaxActivationDistance();
We use this distance in our AI calculations.

apps/openmw/mwmechanics/aipursue.cpp
```diff
@@ -50,8 +50,13 @@ bool AiPursue::execute (const MWWorld::Ptr& actor, CharacterController& characte
 
     //Set the target desition from the actor
     ESM::Pathgrid::Point dest = target.getRefData().getPosition().pos;
+    ESM::Position aPos = actor.getRefData().getPosition();
 
-    if (pathTo(actor, dest, duration, 100)) {
+    float dist = MWMechanics::distance(dest,aPos.pos[0],aPos.pos[1],aPos.pos[2]);
+
+    if (pathTo(actor, dest, duration, 100) &&
+        dist < MWBase::Environment::get().getWorld()->getMaxActivationDistance())  // check the true distance in case the target is far away in Z-direction
```
scrawl: We can't use activationDistance for the one check and 100 for the other check. If someone changes `iActivateDist` to < 100 then this if condition will never complete, right?

So what value should we use, anyway? 100 vs. 196 is a pretty big difference. The latter would make it much harder to escape arrest so, I'd prefer to keep the original value unless we know the larger one is actually what vanilla uses.

scrawl:
The `pathTo` only actually keeps moving the actor if it's determined to not have reached its target, so now I'm worried that there might be a case where `distIgnoringZ < 100` but `dist > 100` (which can normally happen on a small slope) and then the actor would just stop moving but not arrest the player.

drummyfish:
That happens when you're high in the air and the guard gets right below you, right? He then stops moving and just stands there, but as soon as you move somewhere else he starts running after you again. So It's OK I think.

Previously I tried checking the real distance right in `pathTo`, but that caused the guard run below you and then run in circles and it was funny as hell. I don't think that should be the correct behavior.

In vanilla BTW the guard runs below you and then kinda realizes he can't reach you and starts running all around the place (but not in a funny way I mentioned above, i.e. in a 1m diameter circle).

scrawl:
No, what I refer to might happen for example if a guard was trying to chase the player up a hill.

drummyfish:
You're a living debugger scrawl, I see what you're saying now.

range

Should we do fix B or C?

PR #1538, opened 2017-11-03
[Discussion] Body visible in first person (feature #2666)

drummyfish:
I'm playing around with the idea @scrawl made a mockup of [here](http://bugs.openmw.org/issues/2666). To test:

- Add `first person body = true` to settings.cfg under `[Camera]`.
- Play, use `enableracemenu` in the console to change races.

What do you think?


scrawl:
See #1478 for my suggested way forward.

drummyfish:
Looks like it won't be that simple then, but I'll try to experiment some more. I'll try to hide the "body hands" and instead enable the "first person hands", am very curious about the result. Also note to self: disable helmet as well.

Anyway I Iove how swimming looks now, I am one happy Argonian :)

PR #1540, opened 2017-11-05
NifSkope 1.1.3

tmbutterworth:
Added info for downloading NifSkope 1.1.3 for Linux.


tmbutterworth:
Added info for downloading NifSkope 1.1.3 for Linux.

scrawl:
I would prefer if we can put this on the wiki so I don't have to accept new PRs whenever nifskope decides to bring out a new version that does or doesn't compile on linux.

tmbutterworth:
Are you thinking of any particular place on the Wiki? I was looking for a spot on their for this info earlier.

scrawl:
[Tools](https://wiki.openmw.org/index.php?title=Tools#NifSkope) mentions NifSkope. We can make a dedicated article if there is enough to say.

akortunov:
IIRC, NifSkope 2.0 Dev 7 works on Linux.
Also info on Wiki indeed would be easier to edit than adding commits on GitHub.
Maybe just provide a link to Tools page in documentation?

psi29a:
I agreed with @akortunov we should link only to projects and pages that don't change often, not to versions. The wiki-page is ideal for fast editing, for example version changes.

Capostrophic:
This should be closed.

psi29a:
Thanks.

PR #1541, opened 2017-11-05
Fixed escaping @ in boost program options filter

astillich:
Resolves [bug 4189](http://bugs.openmw.org/issues/4189) Openmw Doesn't Accept @ in names of BSAs, ESP/ESMs, Textures, Meshes, Ect.

Fixed handling of @ characters in escape_hash_filter::get(), fixed parameter mixup in EscapeHashString::processString(). 

The mod loads with the fix.

### DIFF
--- a/components/files/escape.cpp
+++ b/components/files/escape.cpp
@@ -1,6 +1,6 @@
 #include "escape.hpp"
 
-#include <boost/algorithm/string/replace.hpp>
+#include <components/misc/stringops.hpp>
 
 namespace Files
 {
@@ -8,7 +8,7 @@ namespace Files
     const int escape_hash_filter::sEscapeIdentifier = 'a';
     const int escape_hash_filter::sHashIdentifier = 'h';
 
-    escape_hash_filter::escape_hash_filter() : mNext(), mPrevious(), mSeenNonWhitespace(false), mFinishLine(false)
+    escape_hash_filter::escape_hash_filter() : mSeenNonWhitespace(false), mFinishLine(false)
     {
     }
 
@@ -26,9 +26,14 @@ namespace Files
 
     std::string EscapeHashString::processString(const std::string & str)
     {
-        std::string temp = boost::replace_all_copy<std::string>(str, std::string() + (char)escape_hash_filter::sEscape + (char)escape_hash_filter::sHashIdentifier, "#");
-        auto format = std::string(1, (char)escape_hash_filter::sEscape);
-        boost::replace_all(temp, std::string() + (char)escape_hash_filter::sEscape + (char)escape_hash_filter::sEscapeIdentifier, format);
+        std::string temp = str;
+
+        static const char hash[] = { escape_hash_filter::sEscape,  escape_hash_filter::sHashIdentifier };
+        Misc::StringUtils::replaceAll(temp, hash, "#", 2, 1);
+
+        static const char escape[] = { escape_hash_filter::sEscape,  escape_hash_filter::sEscapeIdentifier };
+        Misc::StringUtils::replaceAll(temp, escape, "@", 2, 1);
+
         return temp;
     }
 
--- a/components/files/escape.cpp
+++ b/components/files/escape.cpp
@@ -27,7 +27,8 @@ namespace Files
     std::string EscapeHashString::processString(const std::string & str)
     {
         std::string temp = boost::replace_all_copy<std::string>(str, std::string() + (char)escape_hash_filter::sEscape + (char)escape_hash_filter::sHashIdentifier, "#");
-        boost::replace_all(temp, std::string() + (char)escape_hash_filter::sEscape + (char)escape_hash_filter::sEscapeIdentifier, std::string((char)escape_hash_filter::sEscape, 1));
+        auto format = std::string(1, (char)escape_hash_filter::sEscape);
+        boost::replace_all(temp, std::string() + (char)escape_hash_filter::sEscape + (char)escape_hash_filter::sEscapeIdentifier, format);
         return temp;
     }
 
--- a/components/files/escape.hpp
+++ b/components/files/escape.hpp
@@ -30,7 +30,6 @@ namespace Files
 
     private:
         std::queue<int> mNext;
-        int mPrevious;
 
         bool mSeenNonWhitespace;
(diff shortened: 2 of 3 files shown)


scrawl:
I don't think that Meshes and Textures were really affected. Only the bsa/esp go through boost::program_options.

As to the actual fix, I'll have to ping @AnyOldName3 for a review.

AnyOldName3:
Oops.
(Regarding the argument mismatch)

Oops.
(Regarding the writing two @ characters to the stream)

It looks like this makes the ```C++
            else if (mPrevious == sEscape)
            {
                mNext.push(sEscape);
                mNext.push(sEscapeIdentifier);
            }``` block unnecessary, which in turn makes the whole `mPrevious` field, and hence anything writing to it, unnecessary, too. You should probably remove those before this gets merged as it just makes an already confusing function more confusing if they're there.

components/files/escape.cpp
```diff
@@ -27,7 +27,8 @@ namespace Files
     std::string EscapeHashString::processString(const std::string & str)
     {
         std::string temp = boost::replace_all_copy<std::string>(str, std::string() + (char)escape_hash_filter::sEscape + (char)escape_hash_filter::sHashIdentifier, "#");
-        boost::replace_all(temp, std::string() + (char)escape_hash_filter::sEscape + (char)escape_hash_filter::sEscapeIdentifier, std::string((char)escape_hash_filter::sEscape, 1));
+        auto format = std::string(1, (char)escape_hash_filter::sEscape);
+        boost::replace_all(temp, std::string() + (char)escape_hash_filter::sEscape + (char)escape_hash_filter::sEscapeIdentifier, format);
```
psi29a: Can we re-write this to not use boost?  Either with std::replace_copy, or std::replace ?
http://www.martinbroadhurst.com/replacing-all-occurrences-of-a-character-in-a-stdstring.html

psi29a: http://en.cppreference.com/w/cpp/string/basic_string/replace

```
std::string& replaceAll(std::string& context, std::string const& from, std::string const& to)
{
    std::size_t lookHere = 0;
    std::size_t foundHere;
    while((foundHere = context.find(from, lookHere)) != std::string::npos)
    {
          context.replace(foundHere, from.size(), to);
          lookHere = foundHere + to.size();
    }
    return context;
}
```

astillich: I'll look into removing mPrevious and replacing the boost stuff when I get home. Didn't wan to be too intrusive, so I just applied the minimal fixes.

psi29a: If you can, that would be great. Anything that lessens our dependency on boost.
Something that fixes this instance in this PR.

Then a second PR that then uses the same technique, but everywhere boost::replace_all is used.

Unless @scrawl or @zinnschlag disagrees. ;)

astillich: I'm all for removing boost, but I won't do it over the whole code base. Should be a separate issue, IMHO.

psi29a: Well... baby-steps. :)

astillich: Yeah, excatly. Sorry, got wound up in some after-work stuff. ;) Will get to it on wednesday, I hope.

psi29a:
Nice!  I did the same thing but used C++11's std::regex_replace

https://github.com/psi29a/openmw/blob/C%2B%2B17/apps/openmw/mwgui/settingswindow.cpp#L162

Less code... not sure if it as performant.

astillich:
I think compiling the regex every time might kill you performance-wise. Although the code in question isn't exactly critical. ;)

psi29a:
Yay, less boost!  I'll rebase my C++17 branch

scrawl:
@AnyOldName3, any idea if this should be backported to the 0.43 branch? Did the issue with @'s affect 0.43 or any release before it?

AnyOldName3:
Yeah, this should be backported. Something that worked in 0.39 (using `@` symbols in the config files e.g. to refer to data folders or content files) starts working again because of this, so it's a regression fix, and Debian etc. won't like us making a release that has known regressions.

Ideally, we'd backport this to make a 0.40.1, a 0.41.1 and a 0.42.1 release, too, but given that it took this long for the issue to be noticed, I'm not sure that anyone cares enough to justify the effort.

scrawl:
Thanks for the background. I'm on the fence about whether or not to backport because if this went unnoticed for 3 releases then it's unlikely the issue will cause considerable inconvenience for 0.43 users, and there's always a remote chance that the fix creates a new issue that we haven't discovered yet. On the other hand, I've already backported some other things so we'll probably want a new round of testing anyway. Ok, backport it is.

scrawl:
Regarding backports to older releases, I think we've so far been successful in making sure most OpenMW changes are uncontroversial which means that pretty much everyone* will be using the latest release at the time anyway, making backports to any but the latest version mostly a waste of time. That said, we can backport things like easy build fixes against newer libraries to the older release branches (but without making a new release/tag) just for the purpose of being able to regression test.

* I know Debian/Ubuntu only do patch updates for a particular distro release. That's their problem and I don't recommend anyone to use these packages.

psi29a:
:(  You make me sad @scrawl

scrawl:
@psi29a sorry, didn't see you there. For the record, with the 'I don't recommend....' I meant (old) OpenMW packages in particular. 

I think stable updates are generally a good idea and work reasonably well for most things, just not in the context of a game, much less multiplayer games, or other types of software that are by nature always unstable.

For, let's say, a web browser, stable releases are great. This is why I like Firefox over Chrome (Firefox has a LTS series so that I don't have to be annoyed by buttons moving back and forth every other update).

With OpenMW, maybe after 1.0 is reached we could think about a stable series. ATM I don't see much of a point.

Of course, when upstream developers don't do a stable series, but Ubuntu/Debian wants to host 'stable' packages, this presents a bit of a dilemma. Ubuntu/Debian can either keep hosting the old release (to the dismay of the upstream maintainers, who will receive bug reports on things long fixed) or they can host the newest release (to the dismay of users, who find that thing X is gone or doesn't work anymore after they update their system). From what I've seen, usually it will be old releases, but there are exceptions for security critical packages like Chromium. Is that correct or is it up to the package maintainer?

PR #1542, opened 2017-11-06
[Incomplete] #2229 Pathfinding based on any angle A* and physics system

elsid:
The idea is to use any angle **A*** to generate transitions graph on the fly by adding new points based on collisions with world objects. To check collisions **Bullet** is used. The chain of transitions without collisions leads from initial point to goal. The number of iterations (to handle one transition) and depth of search (length of the transitions chain) are limited. New algorithm doesn't replace old. Shortcutting is still working, and if path is not found old pathfinder uses. There are functional tests with specific scenario, tests with randomly generating environment and unit tests for some functions.

TODO:
* Improve performance (reduce the number of collision checks)
* Find path around holes
* ???

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,6 @@ addons:
     packages: [
       # Dev
       cmake, clang-3.6, libunshield-dev, libtinyxml-dev,
-      # Tests
-      libgtest-dev, google-mock,
       # Boost
       libboost-filesystem-dev, libboost-program-options-dev, libboost-system-dev,
       # FFmpeg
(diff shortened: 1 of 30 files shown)


scrawl:
Woah. Hi.

Usually we open a discussion about possible approaches before jumping knee-deep into one, but I guess that is too late here. There were talks of using Recast to build navmeshes, I'm just not sure how well that would adapt to changing environments (like scripts disabling objects) as your solution presumably does (?)

Anyway, this will probably take a long time for me to review. Feel free to carry on in the meantime but don't take this as my approval of this approach - I'll need to look at it some more before I can form an opinion.

elsid:
Hi.

I should thought about open a discussion, but also I wanted to try some ideas. Job is not done. So, we could start a discussion now. I'm not familiar with Recast, but will look now how it could fit into this solution or replace it. This one is adopted to changing environments by using btCollisionWorld to get all information about the world except actor.

Capostrophic:
This should probably be closed.

psi29a:
Leave it to @elsid , he is still active.

PR #1543, opened 2017-11-07
Transparent object markers

Aesylwinn:
This is a partial implementation for making object markers transparent, as suggested by scrawl in #1438 . The transparency is not yet configurable, since doing so will require some tedious chaining of functions. If people could test this and provide feedback on whether or not transparency is the way to go, it would be greatly appreciated.

### DIFF
--- a/apps/opencs/view/render/worldspacewidget.cpp
+++ b/apps/opencs/view/render/worldspacewidget.cpp
@@ -51,6 +51,7 @@ CSVRender::WorldspaceWidget::WorldspaceWidget (CSMDoc::Document& document, QWidg
     , mToolTipPos (-1, -1)
     , mShowToolTips(false)
     , mToolTipDelay(0)
+    , mInConstructor(true)
 {
     setAcceptDrops(true);
 
@@ -114,6 +115,8 @@ CSVRender::WorldspaceWidget::WorldspaceWidget (CSMDoc::Document& document, QWidg
 
     CSMPrefs::Shortcut* abortShortcut = new CSMPrefs::Shortcut("scene-edit-abort", this);
     connect(abortShortcut, SIGNAL(activated()), this, SLOT(abortDrag()));
+
+    mInConstructor = false;
 }
 
 CSVRender::WorldspaceWidget::~WorldspaceWidget ()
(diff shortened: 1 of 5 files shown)


Aesylwinn:
The forum post with some pictures is available [here](http://forum.openmw.org/viewtopic.php?f=7&p=50977).

Aesylwinn:
There doesn't seem to be anyone against this method, so it is probably the way to go.

apps/opencs/view/render/worldspacewidget.cpp
```diff
@@ -128,6 +131,17 @@ void CSVRender::WorldspaceWidget::settingChanged (const CSMPrefs::Setting *setti
         mDragWheelFactor = setting->toDouble();
     else if (*setting=="3D Scene Input/drag-shift-factor")
         mDragShiftFactor = setting->toDouble();
+    else if (*setting=="3D Scene Input/object-marker-alpha" && !mInConstructor)
+    {
+        float alpha = setting->toDouble();
+        // getSelection is virtual, thus this can not be called from the constructor
```
Aesylwinn: I'm not satisfied with this workaround. Is there any better solution?

Aesylwinn:
Looks like there's some more input coming in on the forum, so we should hold off on merging this until that's fully discussed.

Aesylwinn:
@zinnschlag If you have the time, could you chime in with your thoughts on the recent discussion taking place in the forum post?

scrawl:
Is this OK to merge? The forum discussion seems to have derailed a bit. I think the transparent markers are a good bet regardless of whatever additional marker features may be added.

Aesylwinn:
I don't see any problem with merging this. It shouldn't cause any future problems. I can open a new pull request when I find the time to continue working on the markers.

Aesylwinn:
Actually, let me retract that statement. I forgot that there was an issue with smaller rotation markers. Let me see what I can quickly do about that.

Aesylwinn:
There's actually an issue with the draw order. Is there any way to reorder based on distance to the camera? If not then I guess we should just wait on this.

scrawl:
Like this:
<pre>
stateset->setRenderingHint(osg::StateSet::TRANSPARENT_BIN);
</pre>

Aesylwinn:
That fixed it.

PR #1544, opened 2017-11-07
use fProjectileMaxSpeed for ranged weapons distance (fixes #4192)

rexelion:
Now the range is the same as in vanilla Morrowind.

### DIFF
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -504,13 +504,8 @@ namespace MWMechanics
                 targetWeapon = *weaponSlot;
         }
 
-        std::shared_ptr<Action> targetWeaponAction(new ActionWeapon(targetWeapon));
-
         bool isRangedCombat = false;
-        if (targetWeaponAction.get())
-        {
-            rangeAttackOfTarget = targetWeaponAction->getCombatRange(isRangedCombat);
-        }
+        rangeAttackOfTarget = ActionWeapon(targetWeapon).getCombatRange(isRangedCombat);
         
         if (mMovement.mPosition[0] || mMovement.mPosition[1])
         {
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -503,8 +503,8 @@ namespace MWMechanics
                 targetWeapon = *weaponSlot;
         }
 
-        bool isRangedCombat = false;
-        float rangeAttackOfTarget = ActionWeapon(targetWeapon).getCombatRange(isRangedCombat);
+        bool targetUsesRanged = false;
+        float rangeAttackOfTarget = ActionWeapon(targetWeapon).getCombatRange(targetUsesRanged);
         
         if (mMovement.mPosition[0] || mMovement.mPosition[1])
         {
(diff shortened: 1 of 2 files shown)


akortunov:
It would be nice to test also the backing up distance (take a look at [AiCombatStorage::startCombatMove](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/aicombat.cpp#L536)).
It depends on combat distance.

rexelion:
What is the desired value for the backing up distance? After this fix it was increased as well.

akortunov:
> What is the desired value for the backing up distance? 

About 190 game units without mods. But it depends on some GMSTs, and I do not know the exact formula.

akortunov:
But we can roughly take something like fCombatDistance * 1.5 * target.weapon.mReach, if the target uses a melee weapon or bare hands.

rexelion:
Should NPC be able to distinguish between Touch and Target spells?

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -490,6 +490,28 @@ namespace MWMechanics
...
+                targetWeapon = *weaponSlot;
+        }
+
+        std::shared_ptr<Action> targetWeaponAction(new ActionWeapon(targetWeapon));
+
+        bool isRangedCombat = false;
+        if (targetWeaponAction.get())
+        {
+            rangeAttackOfTarget = targetWeaponAction->getCombatRange(isRangedCombat);
+        }
```
PlutonicOverkill: Something seems odd here. If `targetWeaponAction` is allocated with `new`, won't it always be a non-null pointer and thus `targetWeaponAction.get()` always evaluate to `true`? Or am I missing something?
And if it doesn't need to be shared with anything, maybe it should be changed to a `unique_ptr` instead?

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -490,6 +490,28 @@ namespace MWMechanics
...
+        if (targetClass.hasInventoryStore(target))
+        {
+            MWMechanics::WeaponType weapType = WeapType_None;
+            MWWorld::ContainerStoreIterator weaponSlot =
+                MWMechanics::getActiveWeapon(targetClass.getCreatureStats(target), targetClass.getInventoryStore(target), &weapType);
+            if (weapType != WeapType_PickProbe && weapType != WeapType_Spell && weapType != WeapType_None && weapType != WeapType_HandToHand)
+                targetWeapon = *weaponSlot;
+        }
+
+        std::shared_ptr<Action> targetWeaponAction(new ActionWeapon(targetWeapon));
```
PlutonicOverkill: Should this be changed to a `unique_ptr` since it is not being shared with anything?

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -490,6 +490,28 @@ namespace MWMechanics
...
+
+        bool isRangedCombat = false;
+        if (targetWeaponAction.get())
```
PlutonicOverkill: Something seems odd here. Since `targetWeaponAction` is being allocated with `new`, won't it always be a non-null pointer and thus this condition always evaluate to `true`? Or am I missing something?

apps/openmw/mwmechanics/aicombat.cpp
kcat: Also, if it should be a ```shared_ptr``` for some reason, the proper way to instantiate it is with ```make_shared```:
```
std::shared_ptr<Action> targetWeaponAction = std::make_shared<ActionWeapon>(targetWeapon);
```

PlutonicOverkill: Looking more closely at this, why does this need to be a pointer at all? At first I assumed that it needed to be so that it could call `getCombatRange()` polymorphically, but it looks like it can only be an `ActionWeapon`.

apps/openmw/mwmechanics/aicombat.cpp
rexelion: I think you are right, and if using a pointer is not necessary, should I just use ActionWeapon instead?

PlutonicOverkill: I suppose so. I haven't tested it, but you could probably just use simplify this all down to
`rangeAttackOfTarget = ActionWeapon{targetWeapon}.getCombatRange(false);`

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -490,6 +490,23 @@ namespace MWMechanics
 
     void AiCombatStorage::startCombatMove(bool isDistantCombat, float distToTarget, float rangeAttack, const MWWorld::Ptr& actor, const MWWorld::Ptr& target)
     {
+        // get the range of the target's weapon
+        float rangeAttackOfTarget = 0.f;
```
PlutonicOverkill: Just a minor nitpick, but you don't need to initialise this only to change its value immediately afterwards.

scrawl:
I have tested and functionally the changes seem fine, but the code could be cleaned up a bit more. The comment that says how OpenMW differs from vanilla needs to be removed or updated, and the 'isRangedCombat' variable name could be more descriptive (as it is, the reader will wonder what the difference to 'isDistantCombat' would be).

PR #1545, opened 2017-11-07
Fix version info layout (bug #4195)

akortunov:
Fixes [bug #4195](https://bugs.openmw.org/issues/4195).


scrawl:
This is intended so you can copy&paste the version info if you want. I suggest to reject the bug report.

akortunov:
> This is intended so you can copy&paste the version info

Interesting. I did not know that. 
If it works as intented, feel free to close this PR and reject related bugreport.

PR #1546, opened 2017-11-07
Do not artificially limit viewing angle when on preview mode

turnipsi:
On third person view there is no angle limit on pitch, and thus there is more freedom from which point of view player can be seen.  However, when tab-key is held pressed (so that it is possible to see the player from the front, for example), there is a technical limit on the angle on pitch. I fail to see a technical reason for this, this change makes the angle limit behaviour on third person view and "preview mode" equivalent.

### DIFF
--- a/apps/openmw/mwrender/camera.cpp
+++ b/apps/openmw/mwrender/camera.cpp
@@ -294,6 +294,12 @@ namespace MWRender
         const float epsilon = 0.000001f;
         float limit = osg::PI_2 - epsilon;
 
+        // Apply a slightly stricter limit in preview mode so that
+        // it would not be possible to zoom in from below and look inside
+        // the player.
+        if (mPreviewMode && angle >= 0 && mPreviewCam.offset <= 100.f)
+                limit *= std::max(0.5f, mPreviewCam.offset / 100.f);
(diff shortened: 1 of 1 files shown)


scrawl:
IMO this is too exploitable, by moving the camera high above the player you can see things you shouldn't, like peek over a mountain, into an enemy camp, etc. That is probably why game limited the angle in the first place.

turnipsi:
Okay, that is a valid point of view.  On the other hand, I would think that the "zoom out" amount is more relevant to such exploitability, and the angle amount is free on third person view anyway and with maximum "field of view" setting that provides some such exploitability already.

I am guessing that the angle limit on the original game is more due to the fact that without it it is possible to see the insides of the player, when looked up from feet with high zoom in.  This is also a problem with the commit I made.

If OpenMW wants to follow the original game (that I have not played) regarding this, I can understand that.  However, if there still might be some interest in this patch, I think I could pursue this further and try to solve the above mentioned "see player insides"-problem as well.

zinnschlag:
For 1.0 we want to stay as close as possible to the original. We may consider making this configurable after 1.0, but closing for now.

PR #1548, opened 2017-11-09
360° screenshots

drummyfish:
I thought this feature would have potential (Morrowind street maps at UESP wiki? ([demo](https://360.vizor.io/v/dwr), [some action](https://360.vizor.io/v/y79)), nice wallpapers, FB 360 uploads (tested), VR, ...). Not much attention was given to this at the forums so maybe people aren't reall interested but I already have some working implementation so I'm giving it a try.

How to use:

- ~~You have to delete input_v3.xml in your settings folder (~/.config/openmw) because I added a new key. (Is there a way to keep compatibility with the old file? Should I make input_v4.xml?)~~
- ~~360 screenshot key is binded to F8 by default and the screenshots are saved in the same folder as regular ones (~/.local/share/openmw).~~ The screenshots are taken with the same key as regular ones, you just have to change the screenshot type in settings.
- You can configure some settings in settings.cfg (under Video):
  - ~~`s360 width = ...`~~
  - ~~`s360 height = ...`~~
  - ~~`s360 mapping = ...` - 0 (spherical, default), 1 (cylindrical), 2 (small planet), 3 (cubemap)~~
  - ~~`s360 cubemap size = ...` - will be chosen automatically if not specified~~
  - new way: only one setting `screenshot type = type resX resY cubeSize`, where type is `spherical`, `cylindrical`, `planet` or `cubemap` and remaining parameters are optional. E.g. `screenshot type = spherical 1600 1024`.

Notes/issues:

- ~~The input configuration file issue mentioned above.~~
- ~~I disabled reflections and refractions for water in these screenshots ATM because I had trouble with it (the reflected/refracted screen was the same in each cubemap side and so was only correct in front side, dunno why). This also causes reflections/refractions to disappear for a while when taking the 360 screenshot.~~
- ~~All is implemented on CPU. I wanted to use GLSL but OSG doesn't have much support for that. I thought it was gonna be slow but it's not (about half a second).~~
- I made it so that in 3rd person the player is captured in the screenshot and in 1st person they're not.
- The resolution of the screenshot is the screen resolution by default but can be changed in settings. The small planet screenshots are always saved in same width as height.
- The settings are more of a hint, for example small planet and cubemap mappings ignore the screenshot height setting. Cubemap mapping will also set the width of the final image to 6 times the width of cubemap to make it fit. Is this OK, or do settings have to force the values?
- ~~Taking screens in vanity mode messes them up ATM.~~ (disabled this)
- This also solves the bug of wrong water reflections in savegame screenshots.

Some screens (more [here](https://imgur.com/a/KWo9K)):


kcat:
Would it be possible to have an option to save the cubemap itself? Either with each face being a separate image file, or all six faces stitched together into one large image (each face at full size). That way, users would have the option to manipulate the 3D screenshot as they wished.

kcat:
Also, how is the cubemap face size dictated? Could that be made an option to have lower or higher quality screenshots?

drummyfish:
> Also, how is the cubemap face size dictated?

Empirically set to half the size of the final screenshot width. I'll add a settings option for this then.

> Would it be possible to have an option to save the cubemap itself?

Good idea, I'll do this.

drummyfish:
Done. The cubemap is selectable as a new mapping and is saved as an image consisting of the six cubemap sides in a row.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -768,6 +779,9 @@ namespace MWRender
             Settings::Manager::tryGetInt("s360 mapping","Video",SphericalScreenshot::MAPPING_SPHERICAL));
         int cubeWidth = Settings::Manager::tryGetInt("s360 cubemap size","Video",screenshotWidth / 2);                                    
 
+        if (mapping == SphericalScreenshot::MAPPING_CUBEMAP)
+            cubeWidth = screenshotWidth / 6;                  // the image will consist of 6 cube sides in a row
```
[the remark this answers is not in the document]
drummyfish: > screenshotWidth = cubeWidth * 6

Sounds more reasonable, will fix.

> 3x high and 4x wide

That layout is better for viewing, but wastes a lot of space. 6 sides in a row layout is better for automatic processing as you very easily get the cubemap coordinates out of image coordinates (cubeSide = x / cubeSize; x' = x - cubeSide * cubeSize; y' = y).

drummyfish:
I'd really love to get the water reflections/refractions working BTW, but I have no clue why they're messed up. I think I could really use some help here. Details:

- There's a special camera for taking screenshot for savegames (`rttCamera`) which I'm using for also taking the shots for each cubemap side.
- I rotate the camera accordingly to the side being rendered.
- All is OK except for the reflection/refraction texture which is always the same as in front view:
screenshot006

It's like the reflection/refraction objects are excluded from being rendered for `rttCamera` and so they keep their texture from the pass for the main camera? However I tried setting the mask to 1 for `mReflection` and `mRefraction` in `mWater`, but that didn't work.

@scrawl could you point me in the right direction when you have a minute please?

scrawl:
The solution to the water reflections issue is interesting. Do you think this would also work for fixing the water reflections in savegame screenshots? Currently, the reflections in savegame screenshots don't look correct when the screen's aspect ratio is different from the screenshot's.

Anyway, this PR as is seems like too much added code complexity for a very niche feature. I might be more open to merging it if it can be refactored to not touch existing (renderingmanager) code as much, i.e. a new component in sceneutil/ or something.

On the UX side, I think the least intrusive way would be a setting like `screenshot type = <line>` (defaults to empty = standard screenshot, or could be set to something like 'spherical 512 512'), and then we don't need two keybindings.

drummyfish:
> Do you think this would also work for fixing the water reflections in savegame screenshots?

Seems to be working :)

> On the UX side, I think the least intrusive way would be a setting like screenshot type = <line>

Great idea, I'm already working on it. You're right I messed with a lot of files, I just wanted to get the feature up but the code needs to be polished.

drummyfish:
I've made it simpler to use and am touching fewer files now. Back to testing and cleanups. I have to commit often but will squash the commits in the end.

drummyfish:
What bothers me now is the flickering of water effects when taking the 360 screenshot. Is there a way to not render the main camera, i.e. keep it "frozen", when rendering the screenshot?

apps/openmw/mwinput/inputmanagerimp.cpp
```diff
@@ -1014,10 +1016,33 @@ namespace MWInput
...
+            mScreenCaptureHandler->captureNextFrame(*mViewer);
+            MWBase::Environment::get().getWindowManager()->messageBox("Screenshot saved");
+        }
+        else
+        {
+            osg::ref_ptr<osg::Image> screenshot (new osg::Image);
+
+            if (MWBase::Environment::get().getWorld()->screenshot360(screenshot.get(),settingStr))
+                (*mScreenCaptureOperation) (*(screenshot.get()),0);
+                // mScreenCaptureHandler->getCaptureOperation() causes crash for some reason
```
drummyfish: Any idea why `mScreenCaptureHandler->getCaptureOperation()` causes segfault? If I was able to do it I'd be able to modify fewer files. Now I have to awkwardly pass the operation itself.

drummyfish: It's not evel that `getCaptureOperation()` returns 0, it's the call itself that causes the crash (`mScreenCaptureHandler` isn't 0 either). Is it OSG bug then?

drummyfish: So this should be my last issue ATM.

drummyfish: There is a [const and non-const](https://github.com/openscenegraph/OpenSceneGraph/blob/3b1212fed59b1024aab289bf7837056c559913e1/src/osgViewer/ScreenCaptureHandler.cpp) version of `getCaptureOperation`, neither works.

scrawl:
> What bothers me now is the flickering of water effects when taking the 360 screenshot.

But it shouldn't be flickering, should it? If we run things in the right order, anyway:
1. update water reflection for screenshot camera
2. render screenshot
3. update water reflection for main scene
4. render main scene

scrawl:
By the way, on this:

> All is implemented on CPU. I wanted to use GLSL but OSG doesn't have much support for that. I thought it was gonna be slow but it's not (about half a second).

I am not sure why there would be a problem with GLSL. OSG should support all of it. And as a bonus, rendering on the GPU will automatically provide filtering.

apps/openmw/mwinput/inputmanagerimp.cpp
scrawl: Just look at the (rather strange) [implementation](https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osgViewer/ScreenCaptureHandler.cpp#L54) and you'll see. Using ->_defaultCaptureOperation instead should work.

scrawl: No, we can't use that because it's in a private callback. Too bad. I don't see any solution.

drummyfish:
> I am not sure why there would be a problem with GLSL.

Maybe not really a problem, I just thought there would be a magical class I could use. Maybe the problem is I don't know very much how to do this with OSG :)

Anyway if I implemented cubemaps on GPU, could we use them for other things, like water reflections?

Should I try to do this now or leave it for another PR? I wouldn't like to push too much stuff at once.

drummyfish:
> If there's any flickering when saving the game, then we have a problem.

You're right, we have a problem, I gotta fix this. I could simply pass a bool which would keep the old behavior when saving the game, but that would be nasty and also leave us with the savegame screenshot bug you talked about.

> But it shouldn't be flickering, should it? If we run things in the right order, anyway:

I printed the scene and here's a part of it:

```
Group(), mask = 4294967295
  Group(GUI Root), mask = 8192
  LightManager(Scene Root), mask = 4096
    PositionAttitudeTransform(Water Root), mask = 4294967295
      Geometry(), mask = 64
      Geometry(), mask = 128
    Group(Terrain Root), mask = 256
  Camera(ReflectionCamera), mask = 32768
  Camera(RefractionCamera), mask = 32768
  Camera(), mask = 65536
  Camera(), mask = 0
  Camera(), mask = 0
```

`ReflectionCamera` and `RefractionCamera` are parented right to the root node, shouldn't they be parented to the main camera instead in order to inherit it's transform? I don't get how it works this way.

drummyfish:
I'm super-stuck right now, can only think of making a separate water with its own Reflection/Refraction camera for the screenshot camera.

drummyfish:
Ok so I seem to have fixed the flickering issue and even have the rendering on GPU! Will give it some testing now.

PlutonicOverkill:
> even have the rendering on GPU!

Does this mean that the 360 effect could be displayed in real time? Maybe this would not be an ideal way to play, but then you would at least be able to see what the screenshot would look like before you took it.

drummyfish:
> Does this mean that the 360 effect could be displayed in real time?

Well, it would be very nice, but I don't know about that. The scene still has to be captured six times plus one additional pass for stitching it all together, plus the main camera is still being rendered along. So the FPS would go down rapidly, and I don't even know how the preview mode would work. But it's better now because:

- It's a little faster.
- We have the filtering you were talking about.
- There's less code as I could scrap the software implementation of cubemap class.

I'd like to leave it at that and hunt down that last issue with the sun and maybe make the code a little nicer then. Unfortunately the OSG bugs are making this kinda hard.

drummyfish:
The issue seems to be solved. I added additional switch node above the scene root to be able to turn off rendering of the scene for the main camera. If I simply set the scene root mask to 0, it wouldn't be rendered for the screenshot camera either. I tried setting the main camera masks too, but that kept freezing the whole program for some reason.

drummyfish:
Also I wonder how this feature would go with the shadows that are in the making, considering the water effects were causing problems. So I'm leaving a note to myself here: try to merge with shadows and test.

scrawl:
I suspect that the way you turn off the main camera's rendering could cause a problem if double buffering is used. Swapping a buffer without having rendered to it will cause the buffer of the frame before the last one to be used, which might be visible as some kind of 'flickering' effect if you save the game while moving. Note that whether or not double (or triple) buffering is used is implementation defined, but there's a good chance it will be used if VSync is enabled.

drummyfish:
It sounds like you're right, though I've tried now and the issue doesn't show. I thing double buffering should be on as it's a norm now, also I tried with VSync. Anyway I'll look into alternative ways, though I think I've tried them all now :) Maybe draw a fullscreen quad textured with captured screen? I'll see if I can do that.

scrawl:
> Maybe draw a fullscreen quad textured with captured screen?

There's an example in loadingscreen.cpp doing just that.

drummyfish:
```
// Copy the current framebuffer onto a texture and display that texture as the background image
// Note, we could also set the camera to disable clearing and have the background image transparent,
// but then we get shaking effects on buffer swaps.
```

Think I found it, thanks :)

drummyfish:
Looks good now, screenshot taking time also got back to shorter time, nice.

drummyfish:
So now that I've fixed the issues, may I ask for opinions again? Maybe someone on Windows could check if this work, please? Thank you :)

EmperorArthur:
While I can't comment on the code quality itself, I can say those 360 screenshots you linked are nice.  Good work.

drummyfish:
Thank you :-) I was recently thinking about saving Exif metadata to the screenshot, i.e. when and where exactly it was taken in the world which would be useful for putting it on the map later, but the OSG readerwriter doesn't support that sadly.

nhydock:
With Exif metadata and 360 screenshots I now want there to be an OpenStreetMaps project for Vvardenfell

drummyfish:
> With Exif metadata and 360 screenshots I now want there to be an OpenStreetMaps/Cam project for Vvardenfell

I'd like this too, it's already been done for GTA IV I think. Best if someone from UESP wiki started the project, as they already have pretty good maps.

nhydock:
how would exif metadata work if you're in an interior cell?  Aren't they on their own world coordinate system?

drummyfish:
> how would exif metadata work if you're in an interior cell? Aren't they on their own world coordinate system?

You're right, we might be able to only have the name of the interior, or we could try to compute the coordinates as the entrance coordinates in the world plus the coordinates within the interior (but of course there may be multiple entrances etc., so maybe average them... would be non-trivial problem, probably for a separate pull request).

mupfelofen-de:
Hello,
for some reason I can't get it to work. Every screenshot type, except for cubemap, looks similar to this: http://misc.mupfelofen.de/distortion.jpg

Any clue? Thanks, Michael

drummyfish:
Hey, I've replied to your message on the forums, thank you very much for the feedback. It's probably caused by shaders not being enabled. I'll have to deal with this somehow in the code (e.g. show a message so that users know what's going on and how to solve this).

drummyfish:
I just checked and it works for me even with shaders disabled, so I'll have to investigate further.

@mupfelofen-de if it's still not working for you and you need it, you can:
- Checkout old commit with software rendering (like [this](https://github.com/OpenMW/openmw/pull/1548/commits/823218bb611adbbe07a1a4727a6f381cc7b896b4) one).
- Render the raw cubemap which you said worked for you and use some other software to convert it to spherical-mapped image. You could for example write a simple script with Python + PIL to do that.

drummyfish:
After an IRC session with @mupfelofen-de we've made some improvements. He's been able to make some [super crispy images](https://vizor.io/mupf/vivec-pillow-canton-on-ashfall). @scrawl may I ask what you think of this PR now?

mupfelofen-de:
I'm sorry. The link has changed.
Here are some of the _super crispy_ images I've taken:

https://vizor.io/mupf/argus-house
https://vizor.io/mupf/vivec-pillows-at-night
https://vizor.io/mupf/vivec-pillow-canton

@scrawl

TES3MP-TeamFOSS:
Default settings. Nice! So it's a fully optional feature now?
Hope we see it in the master branch soon. Maybe even before TES3MP 0.7.0 is released?

drummyfish:
I've added the `screenshot type` setting to the file that keeps the default settings and also documents them at a basic level, which I should have done earlier, but only learned about today :) I have things set up on my part, but I need feedback from the project leaders, we want to keep the code nice and not break anything, so let's not rush :)

Thunderforge:
@scrawl @zinnschlag Did you have any additional feedback for this? Is there anything that needs to be done for it to be merged, aside from having merge conflicts resolved?

drummyfish:
This is complete I guess, I was just waiting for the maintainers. I'll try to resolve the conflicts at least.

scrawl:
Basically it's in limbo because I'm not sure if a (probably) niche feature like this is worth whatever code cleanup is still necessary and the ongoing maintenance. Personally, I would probably use something like this once, say 'hah cool' then never use it again. I've also had an idea that a similar feature could probably be done with scripting post-1.0 and developed as a third-party mod.

scrawl:
(Don't take that as a definite no, just the explanation for why I've <s>been putting it off</s>not had the time for it over other PR's)

drummyfish:
I actually agree with you, I'm not really sure this should be merged to master, but I'd still like to have this code available for people who want to use this, make their own "OpenMW on steroids" fork etc. Is it OK to keep this PR open for this reason, or should it be closed? Maybe it could be marked with a special label? I dunno.

DBHHeroesJojy:
This is definitely a feature I want to see remain, even if it has to be it's own seperate branch.

Thunderforge:
While I agree it's niche, I do think that it's a nifty feature and an impressive example of "this is something OpenMW does that vanilla+MGE doesn't". Sometimes, features that are more for the wow factor than actual usefulness are still worth including.

My first choice is to merge into master anyways. This feature doesn't look especially more complicated than other features we've included, so I am less concerned about maintenance. 

Should that not be desirable, I don't think that having an "OpenMW on steroids" fork is the way to go because I'm concerned about the long-term viability of that. My concern is that it will ultimately be forgotten. 

A plugin would be a great compromise, but I don't know enough about how plugins will work post-1.0 to know if it's even technically feasible. Should it be possible, I'd be okay with that as a solution. If not, then I would prefer it being part of the main binary.

Thunderforge:
@zinnschlag @psi29a @davidcernat This PR has been in limbo for about seven months. It appears to be feature complete, but no decision has been made about whether to merge it or not.

I'm not sure that there is value in keeping it in limbo any longer. Can we make a decision to either merge it, close it, or set a definitive goal for when it will be merged (e.g. "after 1.0")?

psi29a:
I have no objection against, aside from a changelog entry. :)

Can this be a stepping-stone to OpenMW-VR?

@zinnschlag : something we can add now (it looks well baked) or hold off until after 1.0?

apps/openmw/mwinput/inputmanagerimp.cpp
```diff
@@ -1034,8 +1036,40 @@ namespace MWInput
...
+        bool screenshotTaken = false;
+
+        std::string settingStr;
+
+        try
+        {
+            settingStr = Settings::Manager::getString("screenshot type","Video");
+            regularScreenshot = settingStr.size() == 0 || settingStr.compare("regular") == 0;
+        }
+        catch (std::runtime_error)
```
zinnschlag: This looks suspicious. We just don't ignore exceptions (unless they are related to an error that has already been handled elsewhere). What are we trying to catch here?

zinnschlag:
No objections per se, exception for one code issue I commented on inline.

Scrawl's concerns about additional code complexity is valid though. The whole thing does look a bit messy, But this is a general problem. People are putting too much functionality into the facade classes. We will have to address this issue at some point. But that can wait until 1.0 is out. As much as I itch to do some major refactoring, now is not the time for it.

Once the exception issue has been resolved I am okay with merging.

zinnschlag:
IIRC @drummyfish moved away from GitHub.

drummyfish:
Alright, let me see here - since I am currently transitioning to GitLab workflow with another PR and I don't want to further complicate things, I'll make an exception and will try to address finishing this PR here.

psi29a:
+1 for being pragmatic  :)

drummyfish:
Well yeah, I didn't delete the account because I have some unfinished work, but no new PRs from here :)

I think the `try catch` is because of [this](https://github.com/OpenMW/openmw/blob/7a389f7ab117acad15d813a33f63877f45da98ef/components/esm/loadgmst.hpp#L40) probably?

drummyfish:
Actually [this](https://github.com/OpenMW/openmw/blob/7a389f7ab117acad15d813a33f63877f45da98ef/components/settings/settings.cpp#L392).

So what to do with it? Should I catch and handle it somehow or not catch?

psi29a:
If we catch it, should we log it and continue on?  It's not critical to the functioning of OpenMW right? There are no real side-effects that would effect the stability of the engine?

drummyfish:
E.g. [here](https://github.com/OpenMW/openmw/blob/c7f60a6dc87db0f59a064415ba844917a394af78/apps/openmw/engine.cpp#L454) nothing is being catched, so I guess just do the same?

Now I'm remembering, I think it threw the exception when the value "screenshot type" wasn't present in the config. But since I added it to settings-default.cfg later, it should pull it from there now? I'm gonna try.

drummyfish:
Okay, it was so, I tested it and it's okay without the block, so I removed it.

zinnschlag:
Changelog entry is still missing. Otherwise, ready to merge (well, once CI has caught up with us)

apps/openmw/mwinput/inputmanagerimp.cpp
```diff
@@ -1033,8 +1035,34 @@ namespace MWInput
...
+            screenshotTaken = true;
+        }
+        else
+        {
+            osg::ref_ptr<osg::Image> screenshot (new osg::Image);
+
+            if (MWBase::Environment::get().getWorld()->screenshot360(screenshot.get(),settingStr))
+            {
+                (*mScreenCaptureOperation) (*(screenshot.get()),0);
+                // mScreenCaptureHandler->getCaptureOperation() causes crash for some reason
```
Capostrophic: FIXME should probably be here.

psi29a:
done and done

drummyfish:
This was a long run, thanks to everyone involved :)

PR #1549, opened 2017-11-10
[Testing needed] Item models reworking

akortunov:
Splits refactoring code from #1491 from the reverse pickpocketing related stuff.

1) Moves model-specific stuff from container.cpp to item models:
1.1 Pickpocketing mechanics
1.2 Put items to containers conditions

2) Moves the bound items check to the mechanics manager

3) Changes an item owner and stolen items counter when the player steals the item via pickpocketing.

Item models now have three additional methods where you can put a custom mechanics:
1. onClose() - when closing the window
2. onDropItem() - when the player puts an item to the item model
3. onTakeItem() - when the player takes an item from the item model

An additional testing would be helpful, especially with the "owned crosshair" and "tfh".

### DIFF
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -238,6 +238,7 @@ namespace MWBase
             /// Has the player stolen this item from the given owner?
             virtual bool isItemStolenFrom(const std::string& itemid, const std::string& ownerid) = 0;
 
+            virtual bool isBoundItem(const MWWorld::Ptr& item) = 0;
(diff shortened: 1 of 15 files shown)


scrawl:
I get this error when taking an owned item from the world:
<pre>Error in framelistener: Cannot get class of an empty object</pre>

And it looks like the `allowedToInsertItems()` function is completely redundant now that you added `onDropItem`.

akortunov:
> I get this error when taking an owned item from the world

Should be fixed now.

> And it looks like the allowedToInsertItems() function is completely redundant now

It seems you are right. We can return false in onDropItem() if we want to block the ability to put items to the model.

An additional question: maybe onPutItem() would be a better name that the onDropItem()?

scrawl:
We don't really use the term 'put' anywhere else so I prefer the current naming.

PR #1550, opened 2017-11-11
Do not track a nearest actor during combat and pursue

akortunov:
Fixes [bug #4179](https://bugs.openmw.org/issues/4179).

Just do not track nearest actor if the current actor:
1. Is knocked down
2. Is in combat
3. Has a Pursue AI package


apps/openmw/mwmechanics/actors.cpp
```diff
@@ -1230,17 +1230,23 @@ namespace MWMechanics
                             float sqrHeadTrackDistance = std::numeric_limits<float>::max();
                             MWWorld::Ptr headTrackTarget;
 
+                            MWMechanics::CreatureStats& stats = iter->first.getClass().getCreatureStats(iter->first);
```
scrawl: Tip for next time, better use `const` whenever you can.

PR #1552, opened 2017-11-17
Add projectile - projectile collision (Issue #4201) [incomplete]

rexelion:
The idea is to add projectile object to the physics system whenever someone fires a projectile/casts a spell, but spell projectile meshes don't have collision boxes. I can't figure out how to use bounding box instead of normal collision mesh.


rexelion:
I'm thinking about creating a new btBoxShape, like it is done [here](https://github.com/OpenMW/openmw/blob/master/components/nifbullet/bulletnifloader.cpp#L78-L85). Will this be fine? If so, should it become a member of Resource::BulletShape, or should it be created inside PhysicsSystem::addObject, and then the reference passed to PhysicsSystem::mCollisionWorld::addCollisionObject()?

PR #1555, opened 2017-11-19
Reworked trade window

akortunov:
1. Use AutoSizedTextBox instead of TextBox (should fix the trade window issue from [bug #4215](https://bugs.openmw.org/issues/4215)).
2. Increased minimum window width to [handle more long labels](https://i.imgur.com/XL4Ggm3.jpg) (have we a better solution? Ideally, buttons should not overlap each others.)
3. Every line is wrapped to the HBox, so PlayerGold and MerchantGold can use a "Right" alignment.

### DIFF
--- a/files/mygui/openmw_trade_window.layout
+++ b/files/mygui/openmw_trade_window.layout
@@ -3,7 +3,7 @@
 <MyGUI type="Layout">
     <Widget type="Window" skin="MW_Window" layer="Windows" position="0 0 600 360" name="_Main">
         <Property key="Visible" value="false"/>
-        <Property key="MinSize" value="380 245"/>
+        <Property key="MinSize" value="428 245"/>
 
         <!-- Categories -->
         <Widget type="HBox" position="8 8 566 24" align="Left Top HStretch" name="Categories">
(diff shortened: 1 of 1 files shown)


scrawl:
Can we not increase the width of the number edit box? It's already hideously wide as it is. If at all, the width should match the english label which is what most players would see.

I tested some alternate layouts to use that extra horizontal space for the +/- buttons, but I don't really like them. The first one is too hard to click on and the other ones you have to move your mouse more to go from + to -.

layout1
layout2
layout3

akortunov:
I thought about alternative layouts too:




But text labels may have different length on different languages.

akortunov:
Or something like:

scrawl:
The last one is quite nice. Actually using '>>' for 'Max Sale' would solve an issue with the german version (or all localised versions?) where they forgot to translate the label.

But the one thing I don't like about it is how the '+' and '-' don't match up vertically. I guess that can't be fixed as it's a problem with the font.

mychaelo:
> But the one thing I don't like about it is how the '+' and '-' don't match up vertically. I guess that can't be fixed as it's a problem with the font.

Depends on which character was used. The usual 'fake minus' hyphen is not supposed to be lined up with the plus, but the U+2212 code point 'MINUS SIGN' must do so.

See Wiki pic for comparison: https://en.wikipedia.org/wiki/File:Plus_Minus_Hyphen-minus.svg

akortunov:
> but the U+2212 code point 'MINUS SIGN' must do so.

I tried to use this character previously, but OpenMW does not support it (? is displayed).
I do not know if there is an issue on the engine side, or on the font side.

PR #1558, opened 2017-11-20
Cyrillic journal index (bug #2485)

akortunov:
Related [forum discussion](https://forum.openmw.org/viewtopic.php?f=6&t=4773).


apps/openmw/mwgui/journalviewmodel.cpp
```diff
@@ -305,18 +305,46 @@ struct JournalViewModelImpl : JournalViewModel
...
         MWBase::Journal * journal = MWBase::Environment::get().getJournal();
 
         for (MWBase::Journal::TTopicIter i = journal->topicBegin (); i != journal->topicEnd (); ++i)
         {
-            if (i->first [0] != Misc::StringUtils::toLower(character))
-                continue;
+            unsigned char byte1 = i->first[0];
+            // First, check for two-byte UTF-8 symbols, e.g. Cyrillic ones
+            // TODO: check which language journal index is using
+            if ((byte1 == 0xd0 || byte1 == 0xd1) && i->first.length() >= 2)
```
scrawl: Wouldn't it be easier to convert the topic name to utf32 instead of messing with individual bytes?

scrawl: https://github.com/OpenMW/openmw/blob/11c4aed4e568f0b88d5a0d746351fc90c6ddd15e/components/misc/utf8stream.hpp

https://github.com/OpenMW/openmw/blob/7a389f7ab117acad15d813a33f63877f45da98ef/components/fontloader/fontloader.cpp#L25

akortunov: Do you mean the MyGUI::UString::utf32string?

scrawl:
> I need a way to handle multibyte characters int the createHotStyle().
> For now OpenMW uses the InteractiveId type (which is actually the intptr_t).
> Now I just store the letter position in alphabet, but I can't tell which alphabet to use.

You could store the unicode32 character. It should be relatively safe to assume the intptr_t will be large enough to hold that (it's 32 bit on 32 bit architectures or 64 bit on 64 bit architectures).

scrawl:
> I need somehow to determine which index to use (the Russian or the default Latin one).

I guess the easiest way is to pass the encoding parameter from engine to the gui.

Currently, the [windowmanager](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwgui/windowmanagerimp.cpp#L134) knows the encoding because it's also used to load fonts.

akortunov:
An another question: looks like the journal index shows greetings as themes, e.g. "Greeting 1" or "Greeting 5". Is it a bug or a feature?

akortunov:
I uploaded the improved version.

Note: I do not know how to initialize the Utf8Stream from std::string / hex array correctly, so initialization code is not clean for now.

apps/openmw/mwgui/journalbooks.cpp
```diff
@@ -217,25 +219,37 @@ book JournalBooks::createQuestBook (const std::string& questName)
...
         char ch = 'A' + i;
-
         char buffer [32];
-
         sprintf (buffer, "( %c )", ch);
 
+        char buffer2 [32];
+        sprintf(buffer2, "%c", ch);
+        Utf8Stream stream (buffer2);
+        uint32_t first = stream.peek();
```
scrawl: Wouldn't it be enough to just cast `ch` to uint32? I would expect that ASCII values are the same in Unicode32.

apps/openmw/mwgui/windowmanagerimp.cpp
```diff
@@ -2033,6 +2034,11 @@ namespace MWGui
         return mTextColours;
     }
 
+    ToUTF8::FromType WindowManager::getEncoding()
+    {
+        return mEncoding;
+    }
```
scrawl: We could pass the encoding to the constructor of JournalWindow so as to not have to change WindowManager's external interface.

apps/openmw/mwbase/windowmanager.hpp
```diff
@@ -11,6 +11,8 @@
 
 #include "../mwgui/mode.hpp"
 
+#include <components/to_utf8/to_utf8.hpp>
```
scrawl: This can be removed, yes?

components/misc/stringops.hpp
```diff
@@ -56,6 +57,23 @@ class StringUtils
...
+        // Cyrillic YO character
+        if (ch == 0x0451)
+            ch -= 0x50;
+
+        // Latin alphabet
+        if (ch >= 0x61 && ch < 0x80)
+            ch -= 0x20;
+
+        return ch;
+    }
```
scrawl: I think I would have put this function somewhere locally in the journal code, because it's not complete enough to be generally useful and because the inconsistency with StringUtils::toLower() (which deals with ASCII) is confusing.

scrawl:
I couldn't properly test due to lack of data files, but I'll take your word for it.

scrawl:
> An another question: looks like the journal index shows greetings as themes, e.g. "Greeting 1" or "Greeting 5". Is it a bug or a feature?

The Latin index? That's strange. There's no call to the journal in [startDialogue](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwdialogue/dialoguemanagerimp.cpp#L151), I don't understand how this would have happened.

PR #1559, opened 2017-11-21
Store integer actor ID in AiFollow and AiEscort (bug #4036)

akortunov:
Fixes [bug #4036](https://bugs.openmw.org/issues/4036).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.45.0
 ------
     Bug #2835: Player able to slowly move when overencumbered
+    Bug #4036: Weird behaviour of AI packages if package target has non-unique ID
     Bug #4221: Characters get stuck in V-shaped terrain
     Bug #4293: Faction members are not aware of faction ownerships in barter
     Bug #4327: Missing animations during spell/weapon stance switching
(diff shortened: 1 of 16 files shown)


akortunov:
> What sort of feedback are you looking for?

Mainly about savegame format changes. I got this feedback in #1592: I can read data from savegame, write it only in future releases.
I suppose this PR is ready for review now.

scrawl:
Isn't the  ```MWWorld::Ptr AiEscort::getTarget() const``` content copied from another function? If so can you put that somewhere shared, e.g. AiPackage, to avoid the code duplication?

akortunov:
> If so can you put that somewhere shared, e.g. AiPackage, to avoid the code duplication?

But in this case getTarget() will be a default behaviour for all AI packages, right?

zinnschlag:
Had another look at this. IMO it is almost ready to merge. The save code needs to be un-out-commented (is that actually a word?) and the saved game file version number needs to be increased by 1.

I am not comfortable with the new public variables in the AiPackage class. I have seen that the class already has public variables and that is something we will need to clean up at some point. But we shouldn't make the situation worse.

As a starting point, how about a new AiPackage constructor that takes the actor as an argument and fills in both mTargetActorRefId and mTargetActorId?

akortunov:
> As a starting point, how about a new AiPackage constructor that takes the actor as an argument and fills in both mTargetActorRefId and mTargetActorId?

There are situations, when we do not know actor, but know only actor ID (integer) or actor ref Id (string), e.g. when read data from ESM.
So we need more constructors.

PR #1563, opened 2017-11-23
Make thrown projectiles rotate (#4216)

drummyfish:
Attempt to make thrown projectiles rotate, like they do in vanilla ([video of the bug](https://www.youtube.com/watch?v=dyviwk83dTU&feature=youtu.be)). I added ~~two fields~~ one field to projectile state and added code to save and load the new state to/from savegames, but I need someone to check I did it right.

We could discuss:
- speed of rotation - [here](https://www.youtube.com/watch?v=bN9sxXD6VJo)'s how it looks now, see video description for footage of different speeds
- center of rotation: The projectile now rotates around its tip, but it's not much noticeable because it rotates slowly, I could shift the center of rotation but I would probably need to choose some constant (or maybe center of bounding box?)

### DIFF
--- a/apps/openmw/mwworld/projectilemanager.cpp
+++ b/apps/openmw/mwworld/projectilemanager.cpp
@@ -567,6 +567,9 @@ namespace MWWorld
             state.mVelocity = it->mVelocity;
             state.mAttackStrength = it->mAttackStrength;
 
+            state.mThrown = it->mThrown;
+            state.mTime = it->mTime;
+
             state.save(writer);
 
             writer.endRecord(ESM::REC_PROJ);
@@ -604,6 +607,8 @@ namespace MWWorld
             state.mVelocity = esm.mVelocity;
             state.mIdArrow = esm.mId;
             state.mAttackStrength = esm.mAttackStrength;
+            state.mThrown = esm.mThrown;
+            state.mTime = esm.mTime;
 
             std::string model;
             try
--- a/apps/openmw/mwworld/projectilemanager.cpp
+++ b/apps/openmw/mwworld/projectilemanager.cpp
@@ -317,7 +317,6 @@ namespace MWWorld
         state.mCasterHandle = actor;
         state.mAttackStrength = attackStrength;
         state.mThrown = projectile.get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::MarksmanThrown;
-        state.mTime = 0.0;
 
         MWWorld::ManualRef ref(MWBase::Environment::get().getWorld()->getStore(), projectile.getCellRef().getRefId());
         MWWorld::Ptr ptr = ref.getPtr();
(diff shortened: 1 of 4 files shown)


akortunov:
Looks like this PR changes the savegame format (as a #1559), so old game versions will not load new savegames.

drummyfish:
Yes it changes the format, the new version still loads the old savegames though.

Can't we make it so that the game skips the unknown records when loading so that this isn't an issue next time we modify the format?

akortunov:
> Can't we make it so that the game skips the unknown records when loading so that this isn't an issue next time we modify the format?

A good question. But I think OpenMW does not do that for a reason.

drummyfish:
Actually I can use `mEffectAnimationTime` for the animation, gotta fix that.

drummyfish:
Actually I'm doing this all wrong, there already is `RotateCallback` I should probably use.

scrawl:
Forward compatibility break would be OK here, same as #1470: Projectiles are short-lived and you're unlikely to save your game during battle. In all other cases there are no such records in the save-game.

With that said, do we need to store this state at all? We can just check the type of weapon on loading a game.

scrawl:
By the way, here's a somewhat related issue: https://bugs.openmw.org/issues/3778

drummyfish:
> We can just check the type of weapon on loading a game.

Done.

drummyfish:
I think the vanilla developers didn't bother very much with thrown weapons, throwind stars rotate like this (about the red axis):

image

My projectiles are rotated the "correct" way. Is it OK or should I copy the original behavior?

drummyfish:
[Here](https://www.youtube.com/watch?v=VJcT9Tj3Ao4) is how I'm rotating them now (in the video I'm trying to center the pivot point according to bounding sphere somehow works except for stars).

drummyfish:
All projectiles now rotate nicely around the center of their bounding box :)

drummyfish:
[Looks kinda good now](https://www.youtube.com/watch?v=Uzi-5Wa4wa8&feature=youtu.be) I think :)

scrawl:
LGTM. I'll just give it a test now.

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -202,6 +203,12 @@ namespace MWWorld
 
         osg::ref_ptr<osg::Node> projectile = mResourceSystem->getSceneManager()->getInstance(model, attachTo);
 
+        osg::ref_ptr<osg::ComputeBoundsVisitor> boundVisitor = new osg::ComputeBoundsVisitor();
```
scrawl: FYI, visitors can be created on the stack, so that line can be shortened to:

osg::ComputeBoundsVisitor boundVisitor;

You can do that with many other OSG objects, as long as you never create a ref_ptr of them.

PR #1564, opened 2017-11-24
Display 0 chance for spell if player does not have enought magic energy

thegriglat:
Issue in bugtracker: https://bugs.openmw.org/issues/4223

In vanilla spell chance is displayed as zero if player hasn't enought magic points for spell. In OpenMW displayed chance doesn't depend on magic points.

### DIFF
--- a/apps/openmw/mwgui/spellmodel.cpp
+++ b/apps/openmw/mwgui/spellmodel.cpp
@@ -60,7 +60,7 @@ namespace MWGui
             {
                 newSpell.mType = Spell::Type_Spell;
                 std::string cost = std::to_string(spell->mData.mCost);
-                std::string chance = std::to_string(int(MWMechanics::getSpellSuccessChance(spell, mActor)));
+                std::string chance = std::to_string(int(MWMechanics::getSpellSuccessChance(spell, mActor, NULL, true, true)));
                 newSpell.mCostColumn = cost + "/" + chance;
             }
             else
--- a/apps/openmw/mwgui/spellmodel.cpp
+++ b/apps/openmw/mwgui/spellmodel.cpp
@@ -60,7 +60,7 @@ namespace MWGui
             {
                 newSpell.mType = Spell::Type_Spell;
                 std::string cost = std::to_string(spell->mData.mCost);
-                std::string chance = std::to_string(int(MWMechanics::getSpellSuccessChance(spell, mActor)));
+                std::string chance = std::to_string((stats.getMagicka().getCurrent() >= spell->mData.mCost) ? int(MWMechanics::getSpellSuccessChance(spell, mActor)) : 0);
                 newSpell.mCostColumn = cost + "/" + chance;
             }
             else
--- a/apps/openmw/mwgui/spellmodel.cpp
+++ b/apps/openmw/mwgui/spellmodel.cpp
@@ -8,6 +8,7 @@
 
 #include "../mwmechanics/creaturestats.hpp"
 #include "../mwmechanics/spellcasting.hpp"
+#include "../mwmechanics/actorutil.hpp"
 
 #include "../mwworld/esmstore.hpp"
 #include "../mwworld/inventorystore.hpp"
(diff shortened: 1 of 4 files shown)


scrawl:
We might want to think about this issue in a wider sense. I noticed a few more inconsistencies:

- The HUD icon (once a spell is selected) always shows the success chance regardless of available magicka. This could be seen as a bug or a feature. You could argue it's a feature, because there isn't another way to know the 'difficulty' chance of the spell if you don't have the magicka. If you were to argue this way, you could optionally argue that the spell window should work the same way for consistency.
- If God Mode is on, both Vanilla and OpenMW show '0' chance in the spell window even though no magicka is required for casting spells in god mode.
- If a spell is set to Always Succeed (example: 'blood of the north'), but the character does not have enough magicka, vanilla MW's spell window still displays '100' (which seems wrong). OpenMW (with your patch) correctly shows '0'.

thegriglat:
1. In vanilla MW HUD icon shows success chance regardless of available magicka. I think it is useful (as you mentioned).
2. Vanilla MW shows 100% in HUD and 0% in spell window if player hasn't enough magicka when God Mode enabled (seems to vanilla bug) and 100% in both cases if player has enough magicka when God Mode enabled.
OpenMW shows it in the same way (with this PR, will be fixed) and in right way (100% HUD/window in both cases) without PR.

Capostrophic:
At this point readability begins to suffer. Would it be reasonable to move the new checks to GetSpellSuccessChance?

thegriglat:
@Capostrophic, good remark.
I tried the followind
```
         if (spell->mData.mFlags & ESM::Spell::F_Always)
             return 100;
 
+        if (stats.getMagicka().getCurrent() < spell->mData.mCost  && !godmode)
+            return 0;
+
         if (godmode)
         {
             return 100;
```

After tests with *getSpellSuccessChance()* it seems for me that it is called after spell cast. Is it true?
So if player has 150 magicka and spell requires 120 then spell cast will fail as *stats.getMagicka().getCurrent()* returns 30 which is less than 120.

Capostrophic:
I find this more logical.

        /* local variable declarations */
        if (godmode)
        {
            return 100;
        }

        if (stats.getMagicEffects().get(ESM::MagicEffect::Silence).getMagnitude())
            return 0;

        if (spell->mData.mType == ESM::Spell::ST_Power)
            return stats.getSpells().canUsePower(spell) ? 100 : 0;

        if (spell->mData.mType != ESM::Spell::ST_Spell)
            return 100;

        if (stats.getMagicka().getCurrent() < spell->mData.mCost)
            return 0;

        if (spell->mData.mFlags & ESM::Spell::F_Always)
            return 100;
        /* code */


>getSpellSuccessChance is called after spell cast

Uh... no? I don't think that would be the case.

thegriglat:
@Capostrophic, I mean that `stats.getMagicka().getCurrent()` in this case will be equal *magicka after player cast spell*.

thegriglat:
Moved all checks to openmw/mwmechanics/spellcasting.cpp,

Also I moved decreasing of mana to CastSpell::cast(const ESM::Spell*) -- now mana will be descreased *after* effort to cast spell (not *before* as it is in master).

scrawl:
> Also I moved decreasing of mana to CastSpell::cast(const ESM::Spell*) -- now mana will be descreased after effort to cast spell (not before as it is in master).

Are you sure that's a good idea? It sounds like only a workaround for the problem.

According to [Research page](https://wiki.openmw.org/index.php?title=Research:Magic#Chance_of_successful_spell_cast) there should be a flag passed to getSuccessChance to control whether or not magicka should be checked.

thegriglat:
OK, I will investigate this more. 
The main problem for me is that mana decreased in `World::startSpellCast()` before `CastSpell::Cast()` call. So `if (magicka.getCurrent() < spell->mData.mCost)` in `getSpellSuccessChance()` will return `false` for heavy spells because mana has been already decreased.

Also is it logical that mana decreased before cast? It seems more logical if mana will decrease after cast in both (failed/success) cases.

The following table describes chance and magicka.getCurrent() flow if I don't move mana reducing code.
initial magicka = 250 and spell cost = 135

| Event                                                          | magicka.getCurrent() | chance |
|----------------------------------------------------------------|:--------------------:|:------:|
| check chance in spell window via getSpellSuccessChance()       |          250         |   100  |
| player start casting spell                                   |                      |        |
| check possibility to cast in startSpellCast() (magicka < cost) |          250         |   100  |
| decreasing mana in startSpellCast()                            |         -135         |        |
| trying to cast spell                                         |                      |        |
| check chance in CastSpell::Cast() via getSpellSuccessChance()  |          115         |    0   |
| Spell fails as chance is 0 (but should success)                |                      |        |
| check chance in spell window via getSpellSuccessChance()       |          115         |    0   |

scrawl:
> Also is it logical that mana decreased before cast? It seems more logical if mana will decrease after cast in both (failed/success) cases.

That's just your opinion. We're not going to change how the game works if there's not a very good reason.

You could fix your problem by adding a parameter like 'checkMagicka' to getSpellSuccessChance that's set to true for the spell window but false when the spell is cast (because we already checked and consumed magicka in startSpellCast())

thegriglat:
OK, I added a new parameter and moved mana reducing code back.

PR #1565, opened 2017-11-24
Do not interrupt idle animations during attack (bug #4122)

akortunov:
1. Do not enter to the "attacking or spell" mode" if the player's weapon is holstered.
2. Do not interrupt actor's idle state during attack.

### DIFF
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -232,7 +232,10 @@ namespace MWInput
         if (mControlSwitch["playercontrols"])
         {
             if (action == A_Use)
-                mPlayer->setAttackingOrSpell(currentValue != 0);
+            {
+                MWMechanics::DrawState_ state = MWBase::Environment::get().getWorld()->getPlayer().getDrawState();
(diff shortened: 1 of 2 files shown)


scrawl:
I think this PR introduces a problem, compare with:

1. In first person mode, draw a sword
2. Wait for the part of the idle animation where the weapon moves around
3. Make an attack

Result: when the attack ends, the weapon flickers from one position to a totally different one.

akortunov:
If I understood you correctly, the original game suffers from this issue too, when a player is in the Sneak mode. We can still interrupt idle animations for player in the first-person mode to fix this issue.

Also we can do not interrupt only CharState_IdleSneak and CharState_IdleSwim states in other cases.

scrawl:
I didn't test sneak mode. I only tested in first person which was an issue with this PR but wasn't an issue in the original engine.

> We can still interrupt idle animations for player in the first-person mode to fix this issue.

> Also we can do not interrupt only CharState_IdleSneak and CharState_IdleSwim states in other cases.

Sounds good.

akortunov:
Should work fine now: we keep only CharState_IdleSneak and CharState_IdleSwim and only of the actor is not a player in the first-person mode.

PR #1568, opened 2017-11-25
Implement filtering in the spells window

akortunov:
Allows to filter spells by given substring. Useful when the player knows a lot of spells.
screenshot_20171125_134658
screenshot_20171125_134714

The search is case-insensitive for now (for Latin, Cyrillic and Deutch characters at least).

### DIFF
--- a/apps/openmw/mwgui/journalbooks.cpp
+++ b/apps/openmw/mwgui/journalbooks.cpp
@@ -244,7 +244,7 @@ BookTypesetter::Ptr JournalBooks::createLatinJournalIndex ()
         const MWGui::TextColours& textColours = MWBase::Environment::get().getWindowManager()->getTextColours();
         BookTypesetter::Style* style = typesetter->createHotStyle (body, textColours.journalTopic,
                                                                    textColours.journalTopicOver,
-                                                                   textColours.journalTopicPressed, (uint32_t) ch);
+                                                                   textColours.journalTopicPressed, (Utf8Stream::UnicodeChar) ch);
 
         if (i == 13)
             typesetter->sectionBreak ();
@@ -274,7 +274,7 @@ BookTypesetter::Ptr JournalBooks::createCyrillicJournalIndex ()
         sprintf(buffer, "( %c%c )", ch[0], ch[1]);
 
         Utf8Stream stream ((char*) ch);
-        uint32_t first = stream.peek();
+        Utf8Stream::UnicodeChar first = stream.peek();
 
         const MWGui::TextColours& textColours = MWBase::Environment::get().getWindowManager()->getTextColours();
         BookTypesetter::Style* style = typesetter->createHotStyle (body, textColours.journalTopic,
--- a/apps/openmw/mwgui/journalviewmodel.cpp
+++ b/apps/openmw/mwgui/journalviewmodel.cpp
@@ -7,7 +7,6 @@
 
 #include <components/translation/translation.hpp>
 #include <components/misc/stringops.hpp>
-#include <components/misc/utf8stream.hpp>
 
 #include "../mwbase/world.hpp"
 #include "../mwbase/journal.hpp"
@@ -307,37 +306,20 @@ struct JournalViewModelImpl : JournalViewModel
         visitor (toUtf8Span (topic.getName()));
     }
 
-    void visitTopicNamesStartingWith (uint32_t character, std::function < void (const std::string&) > visitor) const
+    void visitTopicNamesStartingWith (Utf8Stream::UnicodeChar character, std::function < void (const std::string&) > visitor) const
     {
         MWBase::Journal * journal = MWBase::Environment::get().getJournal();
 
         for (MWBase::Journal::TTopicIter i = journal->topicBegin (); i != journal->topicEnd (); ++i)
         {
             Utf8Stream stream (i->first.c_str());
-            uint32_t first = toUpper(stream.peek());
+            Utf8Stream::UnicodeChar first = Misc::StringUtils::toLowerUtf8(stream.peek());
 
-            if (first != character)
+            if (first != Misc::StringUtils::toLowerUtf8(character))
                 continue;
 
             visitor (i->second.getName());
         }
-    }
-
-    static uint32_t toUpper(uint32_t ch)
-    {
-        // Russian alphabet
-        if (ch >= 0x0430 && ch < 0x0450)
-            ch -= 0x20;
-
-        // Cyrillic IO character
-        if (ch == 0x0451)
-            ch -= 0x50;
-
-        // Latin alphabet
-        if (ch >= 0x61 && ch < 0x80)
-            ch -= 0x20;
-
-        return ch;
     }
 
     struct TopicEntryImpl : BaseEntry <MWDialogue::Topic::TEntryIter, TopicEntry>
--- a/apps/openmw/mwgui/journalviewmodel.hpp
+++ b/apps/openmw/mwgui/journalviewmodel.hpp
@@ -6,6 +6,8 @@
 #include <functional>
 #include <stdint.h>
 
+#include <components/misc/utf8stream.hpp>
+
 namespace MWGui
 {
     /// View-Model for the journal GUI
@@ -76,7 +78,7 @@ namespace MWGui
         virtual void visitTopicName (TopicId topicId, std::function <void (Utf8Span)> visitor) const = 0;
 
         /// walks over the topics whose names start with the character
-        virtual void visitTopicNamesStartingWith (uint32_t character, std::function < void (const std::string&) > visitor) const = 0;
+        virtual void visitTopicNamesStartingWith (Utf8Stream::UnicodeChar character, std::function < void (const std::string&) > visitor) const = 0;
 
         /// walks over the topic entries for the topic specified by its identifier
         virtual void visitTopicEntries (TopicId topicId, std::function <void (TopicEntry const &)> visitor) const = 0;
--- a/apps/openmw/mwgui/keyboardnavigation.cpp
+++ b/apps/openmw/mwgui/keyboardnavigation.cpp
@@ -14,6 +14,9 @@ namespace MWGui
 
 bool shouldAcceptKeyFocus(MyGUI::Widget* w)
 {
+    if (w && w->getUserString("IgnoreTabKey") == "y")
+        return false;
+
     return w && !w->castType<MyGUI::Window>(false) && w->getInheritedEnabled() && w->getInheritedVisible() && w->getVisible() && w->getEnabled();
 }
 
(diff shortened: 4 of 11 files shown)


akortunov:
Known issues:
1. The purpose of edit box is not clear when it is empty:
screenshot_20171125_134929

2. If there are no spells found, the window does not look good too:
screenshot_20171125_135549

3. Found substring highlighting breaks the current spell highlighing. Cut off this feature?
screenshot_20171125_134758

4. Selected filter persists between reloads. I am not sure if it should behave this way.


Any suggestions how to improve design?

thegriglat:
Best feature!

Ideas:
1. Put filter input at the bottom of the spell window. In that case if filter is clear it is not so conspicuous.
2. Maybe lock typing in the filter if no spells found? IMHO -- empty spell list is also correct result.
3. Don't highlight substring at all. It's clear that spells are filtered in window.

Possibly bug -- if you type "J" (or even "О" in russian) in filter then journal will be opened.

akortunov:
> Possibly bug -- if you type "J" (or even "О" in russian) in filter then journal will be opened.

I can confirm this bug, but only if OpenMW was started with a Russian keyboard layout.

akortunov:
> Put filter input at the bottom of the spell window. In that case if filter is clear it is not so conspicuous.

I do not see a much difference.

Find a text input in the screenshot:
screenshot_20171125_153847

akortunov:
Or something like:
mockup
mockup2

Warning: these pictures are mockups.

thegriglat:
In 1) I mean something like this

screenshot_2017-11-25_13-58-57

akortunov:
> I mean something like this

Yes, it looks better than two similar fields on the top.

akortunov:
How about ascending/descending switch?

screenshot_20171125_212030

scrawl:
How about an overlay search (or 'type ahead find') like the one GTK has? Example:

http://ubuntu-news.ru/sites/default/files/nautilus-type-ahead-find-trusty.png

Or, more generally, a search bar that only appears as you type.

I think the general issues with this feature are:

* The inventory key bound to 'Tab' (as Chris brought up before). The current PR breaks this because if there is any key-focusable widget in the inventory, Tab will focus that widget instead of closing the inventory.
* Key input for the inventory mode might be awkward because of the 4 different windows. If we also wanted to add a search to the inventory window, you have to keep track which of the 4 windows is in focus. And I'm not sure if 'Tab' should cycle between controls in one window or different windows.
* Some way to remove focus from the search bar so that controls like J for journal can be used.

akortunov:
> How about an overlay search

I thought about it, but the "Key input for the inventory mode might be awkward because of the 4 different windows."

akortunov:
I can suggest only hide the search layout by default:

mockup7

In this case we can add additional parameters (e.g. filtering by school or target type).

akortunov:
Try to determine what this filter means (greyed button are disabled, we can use transparency instead):
mockup8

akortunov:
> The current PR breaks this because if there is any key-focusable widget in the inventory, Tab will focus that widget instead of closing the inventory.

Looks like the keyboard navigation has a higher priority than key bindings. So if there are items to navigate, TAB binding is not working.

Kjarrigan:
:+1: for this featured. Played MW recently and it's always a pain to scroll through the spells....

I prefer the version with the bar at the bottom. To clarify what this thing does - couldn't you just add an :mag: symbol in front. Most should recognize this as a search. The second option (if available) would be to write some default text in the box which is deleted if you click in it (what most textinputs in browsers have these days) so something like: 'Type here to filter spell list'

> In this case we can add additional parameters (e.g. filtering by school or target type).
Don't know - never felt the need to show all spells by school BUT a little icon before the spell name indicating the school would be nice, so I can direclty see why my sucess chance is so bad (e.g. because my alteration is so bad)

akortunov:
> 🔍 symbol

I doubt that Morrowind fonts support it. These fonts do not support even the 'Minus' sign.

> 'Type here to filter spell list'

We have no way to localize such text. I can suggest to use something like sName GMST. Does MyGUI support the text hint inside the EditBox?

> BUT a little icon before the spell name indicating the school would be nice

What kind of icons? IIRC, skill icons have 32x32 size.  Even with large size it is hard to determine skill by magic skill icon, and if we will resize icons to 16x16, readability will become even worse.
As a possible replacement, we can add a column with icons that indicate spell effects.

akortunov:
Is there any way to force the EditBox to ignore the TAB button?

akortunov:
> Some way to remove focus from the search bar after you're done searching so that controls like J for journal can be used.

How to determine this "after done"? Would press ENTER/ESC be enough?

akortunov:
> Key input for the inventory mode might be awkward because of the 4 different windows. If we also wanted to add a search to the inventory window, the user has to keep track which of the 4 windows is in focus.

IMO, if we ever will want to implement the key input for these windows (for navigation or filtration), we should show only one window at time. In this case we can implement a some kind of Skyrim-style GUI for keyboard and gamepad.

kcat:
s/Skyrim-style GUI/SkyUI-style GUI/. Skyrim's unmodded GUI is horrible, even for controllers.

akortunov:
> s/Skyrim-style GUI/SkyUI-style GUI/.

Yes, I actually meaned SkyUI. I described some ideas [here](https://forum.openmw.org/viewtopic.php?f=20&t=4787).

scrawl:
> Is there any way to force the EditBox to ignore the TAB button?

No, but it shouldn't be difficult to add that. You can add a UserString to the widget and check for that string in keyboardnavigation.cpp.

> How to determine this "after done"? Would press the ENTER/ESC be enough?

ESC is fine.

akortunov:
> You can add a UserString to the widget and check for that string in keyboardnavigation.cpp.

Thanks. Binding TAB to inventory works now, also the search field does not get the focus from the TAB key now.

akortunov:
You can remove focus by clicking outside the search window.
Possible improvements:
1. Intercept the ESC key press and remove focus instead of closing window.
2. Remove focus when opening the inventory.

akortunov:
@thegriglat, can you retest the latest version of this PR?

akortunov:
And I still think there should be a some kind of prompt when the filter is not active.
For example (here I use the sName GMST):
screenshot_20171129_230043

Any ideas?

thegriglat:
It seems filter field has absolute (not float) position in spell window. The filter is sometimes invisible when you resize spell window.

screenshot000
screenshot001
screenshot002

akortunov:
> The filter is sometimes invisible when you resize spell window.

Should be fixed now.

EmperorArthur:
> I doubt that Morrowind fonts support it. These fonts do not support even the 'Minus' sign.
Yeah, that sounds like an icon that the project itself would have to provide.  Mind you, there are plenty of places where a search box would be handy, and it is the universal method of denoting one. So it wouldn't be a bad idea for someone to come up with an appropriate icon.

Capostrophic:
Yep, I think this is a nifty feature that at the same time looks good in implementation and is worthy to be added pre-1.0. What's keeping it from being merged?

zinnschlag:
Looks okay to me. Anyone any objections? @kcat @psi29a

akortunov:
I am not sure when I should clear spell filter: after exit from inventory, after reload, or do not clear it at all?

zinnschlag:
After exit from inventory is probably not a good idea Lots of use-cases where this would fail. Normally I would say just add a new user setting, but we already have too many unlocalised user settings that will be a pain in the arse to clean up later. After reload/new game seems a reasonable solution for now.

akortunov:
> After reload/new game seems a reasonable solution for now.

Ok, I will try to implement it.

akortunov:
>  I'd strongly suggest doing case-insensitive searches if it's not done already.

No, it was not done yet - filter does not take care about content encoding for now.

zinnschlag:
> If I search for something, it's not likely that I'll need to retain that search after exiting and reopening the inventory

I would say it is equally likely as it is unlikely. However if the user has to repeatedly type in a search string that is more annoying than to have to remove the search string after it is not needed anymore

Good point about the case-insensitivity though. Morrowind ain't Linux and your average user will be surprised by a case-sensitive search.

akortunov:
About case-insensitive search: we are already using similar approach for [journal](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwgui/journalviewmodel.cpp#L326),
so I can just move this function in more common place.

akortunov:
I uploaded the case-insensitive search.
Since both comparing strings are in UTF8, I had to implement a custom lowercase function for UTF8 characters.
Feel free to correct me if I re-invented a wheel and we have a better option.

About filter clearing: I suggest to do not clear it automatically for now (current behaviour) and collect a feedback from users.

psi29a:
How much work do you think it would be to support other character sets? I thought there was a chinese or japanese translation of Morrowind as well. (Not for this PR, but just to get an idea if anyone ever complains.)

akortunov:
> How much work do you think it would be to support other character sets? I thought there was a chinese or japanese translation of Morrowind as well.

For Latin/Cyrillic-based languages there should not be much work.

As for chinese or japanese languages, I even do not know if these languages have upper/lower case.
IIRC, japanese translation is not official and requires MCP hacks.

akortunov:
> One thing that imho would help a lot locating spells is having the icon of the spell to the left of its name.

First of all, there is no such thing as "spell icons". There are only effect icons. Spell/item can have up to 8 effects.
Second, some neighbouring items in list will have the same (or similar) icon with low resolution, e.g. Restore, and Summon spells.
Third, it is outside of scope of this PR since it has nothing common will spell filtering.

zinnschlag:
> As for chinese or japanese languages, I even do not know if these languages have upper/lower case.

To the best of my knowledge neither the Kanji, nor the Hiragana nor the Katakana writing systems have a concept of upper/lower case characters. We work with UTF-8 internally, so we should already fully support Japanese and probably Chinese too, since Kanji is based on that.

Rendering in the GUI and font support are different issues, no idea about their state.

What stops OpenMW from supporting these languages in practice is that our omgame/omwaddon readers/writers (not sure about save game format) are still using legacy encodings instead of UTF-8. This will change immediately after 1.0.

kcat:
> To the best of my knowledge neither the Kanji, nor the Hiragana nor the Katakana writing systems have a concept of upper/lower case characters.

They don't. At best you could match specific Kanji to Hiragana characters, but that's quite a bit more complex than case smashing (you need to check one Kanji character against one or more Hiragana characters, and I don't think there's any pragmatic way to know what Hiragana characters a given Kanji character should match; you'd probably just need a massive table).

zinnschlag:
Not familiar enough with the topic, but would a Japanese user even expect such a matching system for his software?

kcat:
No idea. I'd probably not worry about it unless enough people complain about it missing, though.

psi29a:
@zinnschlag mergable?

zinnschlag:
Probably. I want to give it a quick test-drive before I merge. Its in my test/review-queue for today.

zinnschlag:
We don't have an issue on the tracker for this one, right? My search is coming up empty.

PR #1569, opened 2017-11-26
Toggleborders console command/instruction (#1421)

drummyfish:
I'm trying to implement a new instruction that was in vanilla - toggle rendering of cell borders. It seems like in vanilla these were rendered with separate primitives - lines. I'm trying to make it look similar but instead use terrain fragment shader, as it's more elegant in my opinion, and probably even looks nicer. Comparison (venilla left, openMW right):

image

I didn't implement rendering of the borders in the minimap yet (as seen in the screens).

I need some review here, will leave some notes in the code of what I'm not sure about.

notes:
- Also works when wireframe is enabled.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -61,7 +61,6 @@ namespace MWRender
             : mFogStart(0.f)
             , mFogEnd(0.f)
             , mWireframe(false)
-            , mBorders(false)
         {
         }
 
@@ -80,16 +79,6 @@ namespace MWRender
             }
             else
                 stateset->removeAttribute(osg::StateAttribute::POLYGONMODE);
-
-            osg::Uniform *borderUniform = stateset->getUniform("borders");
-
-            if (!borderUniform)
-            {
-                borderUniform = new osg::Uniform("borders",(int) 0);
-                stateset->addUniform(borderUniform);
-            }
-
-            borderUniform->set((int) mBorders);
         }
 
         virtual void apply(osg::StateSet* stateset, osg::NodeVisitor*)
@@ -136,27 +125,12 @@ namespace MWRender
             return mWireframe;
         }
 
-        void setBorders(bool borders)
-        {
-            if (mBorders != borders)
-            {
-                mBorders = borders;
-                reset();
-            }
-        }
-
-        bool getBorders()
-        {
-            return mBorders;
-        }
-
     private:
         osg::Vec4f mAmbientColor;
         osg::Vec4f mFogColor;
         float mFogStart;
         float mFogEnd;
         bool mWireframe;
-        bool mBorders;
     };
 
     class PreloadCommonAssetsWorkItem : public SceneUtil::WorkItem
@@ -485,12 +459,6 @@ namespace MWRender
             bool wireframe = !mStateUpdater->getWireframe();
             mStateUpdater->setWireframe(wireframe);
             return wireframe;
-        }
-        else if (mode == Render_Borders)
-        {
-            bool borders = !mStateUpdater->getBorders();
-            mStateUpdater->setBorders(borders);
-            return borders;
         }
         else if (mode == Render_Water)
         {
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -40,8 +40,6 @@
 #include <components/fallback/fallback.hpp>
 
 #include "../mwworld/cellstore.hpp"
-#include "../mwbase/environment.hpp"
-#include "../mwbase/windowmanager.hpp"
 
 #include "sky.hpp"
 #include "effectmanager.hpp"
@@ -450,7 +448,6 @@ namespace MWRender
     {
         mBorders = !mBorders;
         mTerrain->setBordersVisible(mBorders);
-        MWBase::Environment::get().getWindowManager()->setCellBordersVisible(mBorders);
         return mBorders;
     }
 
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -220,9 +220,10 @@ namespace MWRender
                                              Settings::Manager::getBool("auto use terrain specular maps", "Shaders"));
 
         if (distantTerrain)
-            mTerrain.reset(new Terrain::QuadTreeWorld(sceneRoot, mRootNode, mResourceSystem, mTerrainStorage, Mask_Terrain, Mask_PreCompile));
+            mTerrain.reset(new Terrain::QuadTreeWorld(sceneRoot, mRootNode, mResourceSystem, mTerrainStorage, Mask_Terrain, Mask_PreCompile, Mask_Debug));
         else
-            mTerrain.reset(new Terrain::TerrainGrid(sceneRoot, mRootNode, mResourceSystem, mTerrainStorage, Mask_Terrain, Mask_PreCompile));
+            mTerrain.reset(new Terrain::TerrainGrid(sceneRoot, mRootNode, mResourceSystem, mTerrainStorage, Mask_Terrain, Mask_PreCompile, Mask_Debug));
+
         mTerrain->setDefaultViewer(mViewer->getCamera());
 
         mCamera.reset(new Camera(mViewer->getCamera()));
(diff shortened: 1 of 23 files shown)


apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -1374,6 +1388,7 @@ namespace MWScript
             interpreter.installSegment5 (Compiler::Misc::opcodeRemoveFromLevItem, new OpRemoveFromLevItem);
             interpreter.installSegment3 (Compiler::Misc::opcodeShowSceneGraph, new OpShowSceneGraph<ImplicitRef>);
             interpreter.installSegment3 (Compiler::Misc::opcodeShowSceneGraphExplicit, new OpShowSceneGraph<ExplicitRef>);
+            interpreter.installSegment5 (Compiler::Misc::opcodeToggleBorders, new OpToggleBorders);
```
drummyfish: What's the difference between `installSegment5`, `installSegment3` etc.?

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -79,6 +80,16 @@ namespace MWRender
             }
             else
                 stateset->removeAttribute(osg::StateAttribute::POLYGONMODE);
+
+            osg::Uniform *borderUniform = stateset->getUniform("borders");
+
+            if (!borderUniform)
+            {
+                borderUniform = new osg::Uniform("borders",(int) 0);
+                stateset->addUniform(borderUniform);
```
drummyfish: I tried adding the uniform only to the terrain node in ../components/terrain/world.cpp and then get it here with `getUniform` from the stateset, but I found this method is only called for the sceneRoot node, so I always got NULL from it. So now I'm simply setting this uniform for the whole sceneRoot node, but I'd like to know if there is a better way, i.e. only have the uniform on the terrain. Or is this OK?

drummyfish:
Just found the border lines on the map are bugged in vanilla, at least in Wine, so that's another thing we can fix with OpenMW. Screens for reference:
tb reference 2

scrawl:
I don't think we should use shaders for this.

* Shaders are off by default. As it is, the new command won't do anything unless the user manually enabled shaders.
* The more you add to a shader the more of an entangled mess it becomes. I would not use shaders for a feature that can be done without shaders.
* It will be harder to disable the markers for certain render targets, if we want to do that (e.g. water reflections). The easier way is to make a separate debug geometry with its own node mask (I think we already have a 'Debug' node mask too).

apps/openmw/mwscript/miscextensions.cpp
scrawl: I think 3 is for functions with optional arguments.

drummyfish:
Well, that's true, so back to square one I guess :)

drummyfish:
OK now I'm using geometry to draw the borders, and somehow they are even rendered at the minimap too, how come? :D

image

Am I doing it right? (I have yet to remove the old code etc. of course.)

drummyfish:
I get it now, the minimap is rendered at runtime, I thought it was prerendered in some file for some reason. I need to force its re-rendering when `tb` command is issued though, gotta find out how.

drummyfish:
It's difficult to rerender the local map, I tried doing it manually but it didn't work for some reason, it only rerenders when I move and change cells, and even then only some cells are rerendered, so I'm left with exactly the same bug I showed above in the vanilla screenshot. Even if I somehow manage to forcefully rerender all the cells, it will take a very long time.

So maybe I should abandon this rerendering approach and should just put some overlay on the local map instead?

drummyfish:
Also I found that if you toggle wireframe and cells get reloaded, the local map gets affected like this:

image

Will we consider it a bug?

drummyfish:
One hack to show the borders on the local map without creating new widgets could be to decrease the size of ImageBox widgets which results in lines exactly at cell borders:

image

But the background is transparent and I don't know how to change it :)

drummyfish:
Would this be ok?

image

drummyfish:
Or this is even better.

image

drummyfish:
I'm having problems with MyGUI. If I use any function that affects the image (setImageRect, setImageCoord, setImageTile) on the ImageBox after it's been created, it simply disappears.

drummyfish:
Solved by using setSize of the widget class instead, but the borders don't show on the minimap (only in the big local map window).

Also note to self: disable in interiors.

drummyfish:
So now it works quite well, except:

- No borders in the small minimap.
- Borders are not visible through water with refractions (they are visible underwater) because they have the debug mask set and so are not rendered in the refraction texture. Should I address this?

Also I can squash these commits if you tell me to.

drummyfish:
Now it's all working except for the water issue (no. 2 mentioned above), but I think that's a separate bug for another PR? It affects all geometry with debug mask, so for example collision boxes aren't visible through water either.

So could this be it for this PR?

scrawl:
Hi, sorry for the wait. I completely forgot about this PR. Maybe those people suggesting we do a purge of all the abanded PRs to get better visibility of the good ones were right.

It sounds like (have not tested yet) this is functionally ok. The function name just has a typo (toggleBordorders). Also +233 lines seems like an awful lot of code for a small debug function, probably in part due to the map rendering. Is it really necessary to see the borders on the map? Does the vanilla game show them there? If, with the naive implementation, we're bothered by borders either showing or not showing based on when the map was rendered, could that be solved by just never showing borders by way of excluding them in the map's node mask?

(Sorry for all the nitpicking:)

> Also I found that if you toggle wireframe and cells get reloaded, the local map gets affected like this:

I'd say that's a bug. Should be easy to fix if you add an osg::PolygonMode attribute in OVERRIDE mode to the map node.

scrawl:
> Borders are not visible through water with refractions (they are visible underwater) because they have the debug mask set and so are not rendered in the refraction texture. Should I address this?

Both is ok for me. I'll leave it to you.

drummyfish:
Typo fixed.

It's been a while but looking at the diff, the map rendering is only about 40 lines, most of the changes is the management around creating the border geometry. So if you want, I could take it away, but it wouldn't help that much. Vanilla had minimap border rendering, but they were buggy, this code actually works. I guess I won't address the water visibility issues, it's a minor thing and I don't want to start adding even more code.

scrawl:
I count a bit more than 40 but anyway, I guess I wasn't so much concerned about the size but the added complexity. The map code is, thanks to MyGUI's horrid layer system, complicated enough as it is.

Unfortunately the map changes in this PR break some things for me. I can't move the map with the mouse anymore and at one point I got some yellow gibberish instead of a map, although I can't reproduce the latter thing now. I'm not too keen on spending more time debugging this, so let's just revert the map part. Actually there may be a different way to do it. Let the map render include the border meshes from the scene, and make sure to re render the maps each time that 'toggleBorders' is called. That's assuming the border meshes are big enough to show in the map in any proper fashion, which I'm not sure of. If this is too big a hassle, just ignore the map then. Like you said, it's broken in vanilla anyway.

drummyfish:
Fair enough. I tried to revert the local map borders, hope it's OK now, please test.

scrawl:
Cool, now there's less code I have to tear apart ;)

osg/Switch - if it stays at all, which it probably won't when I'm done with my rant - should have been forward declared. But I think the switch is unnecessary. Just nuke the meshes when borders are turned off, it's not too expensive to recreate them. Similarly we don't need to create the border meshes before they're used, that's imo a big no with debugging code, it should not have any effect until you enable it.

The only method of the Terrain class you're using is the getHeightAt, which is already public, so I'm not sure if inside Terrain is the best place to put this debugging code. I usually prefer debugging code to be placed in a separate class because that kind of code is not something I would be concerned with when I'm reading through a component trying to understand it. It's just fairly boilerplate-ish code that distracts from the main thing. To add to that the Terrain isn't necessarily concerned with loading/unloading cells, at least the distant terrain path which will probably be made the one and only path in the future once it's improved a bit more.

We could make a new class similar to the MWRender::Pathgrid just for the border rendering.

drummyfish:
Well, seems like there's some work to be done :) I'll take a look at it when I have some spare time, am a little busy right now, but I agree with what you're saying and will fix it.

drummyfish:
Closed because I no longer use GitHub, but will try to continue [here](https://gitlab.com/OpenMW/openmw/merge_requests/1733).

PR #1572, opened 2017-12-01
[Feedback needed] AI: open doors before collision

akortunov:
In our current implementation AI tries to open doors only when it stucks at obstacle for several seconds. In vanilla Morrowind AI opens doors faster.
With this PR AI tries to open the door within activation distance every AI_REACTION_TIME seconds (0.25 sec).

I am aware of performance cost of this solution, and I possibly broke something, so any additional testing/feedback would be helpful.

P.S. It would be nice to deterimine the door opening direction and modify AiAvoidDoor: an actor generally should evade a door only when it is opening to him. Otherwise he can just wait until the door will be fully opened. Any ideas how to implement it?


apps/openmw/mwmechanics/aipackage.cpp
```diff
@@ -182,45 +184,43 @@ void MWMechanics::AiPackage::evadeObstacles(const MWWorld::Ptr& actor, float dur
...
+        }
+
+        const std::string keyId = door.getCellRef().getKey();
+        if (keyId.empty())
+            return;
+
+        MWWorld::ContainerStore &invStore = actor.getClass().getContainerStore(actor);
+        MWWorld::Ptr keyPtr = invStore.search(keyId);
+
+        if (keyPtr != MWWorld::Ptr())
```
zinnschlag: This should be just

> if (keyPtr)

Let's not try to be unnecessarily verbose.

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -47,7 +47,7 @@ namespace
         for (typename MWWorld::CellRefList<T>::List::iterator iter (list.mList.begin());
              iter!=list.mList.end(); ++iter)
         {
-            if (Misc::StringUtils::ciEqual(iter->mBase->mId, id2))
+            if (Misc::StringUtils::ciEqual(iter->mBase->mId, id2) && iter->mData.getCount())
```
zinnschlag: Wow! Major change. I understand why you did put it in there. Let's hope it doesn't break anything else.

apps/openmw/mwmechanics/obstacle.cpp
```diff
@@ -50,6 +50,15 @@ namespace MWMechanics
             const MWWorld::LiveCellRef<ESM::Door>& ref = *it;
 
             osg::Vec3f doorPos(ref.mData.getPosition().asVec3());
+
+            MWWorld::Ptr doorPtr = MWWorld::Ptr(&const_cast<MWWorld::LiveCellRef<ESM::Door> &>(ref), actor.getCell());
```
zinnschlag: Ugh! const_cast. Not yours, I see. But still. At least leave the FIXME comment in.

apps/openmw/mwmechanics/aipackage.cpp
```diff
@@ -120,6 +120,8 @@ bool MWMechanics::AiPackage::pathTo(const MWWorld::Ptr& actor, const ESM::Pathgr
 
     if (!isDestReached && mTimer > AI_REACTION_TIME)
     {
+        openDoors(actor);
```
Capostrophic: No bipedal check for this call.

apps/openmw/mwmechanics/aipackage.cpp
```diff
@@ -209,45 +211,43 @@ void MWMechanics::AiPackage::evadeObstacles(const MWWorld::Ptr& actor, float dur
...
-        if (getTypeId() != TypeIdWander && !door.getCellRef().getTeleport() && door.getClass().getDoorState(door) == 0)
-        {
-            if ((door.getCellRef().getTrap().empty() && door.getCellRef().getLockLevel() <= 0 ))
-            {
-                MWBase::Environment::get().getWorld()->activate(door, actor);
-                return;
-            }
+        openDoors(actor);
+    }
+    else // any other obstacle (NPC, crate, etc.)
```
Capostrophic: This comment is not entirely accurate which is why it was removed.

apps/openmw/mwmechanics/aipackage.cpp
akortunov: Good catch. This PR is a quite old.

PR #1573, opened 2017-12-01
[Do not merge yet] Play spellcasting effects from objects

akortunov:
Partially implements [feature #1645](https://bugs.openmw.org/issues/1645).

The basic idea: if the caster has no animation to attach the effect, I just spawn the effect via MWBase::Environment::get().getWorld()->spawnEffect() at the caster's position.

The only thing left is the effect scaling to the caster object size.
How can I calculate the mesh dimesions? Via osg::ComputeBoundsVisitor?

Note: looks like Morrowind uses the following formula to calculate the scale:

`effect.scale = max(caster.size.x/effect.size.x, caster.size.y/effect.size.y, caster.size.z/effect.size.z)`

so the effect will keep its proportions.


Allofich:
> The only thing left is the effect scaling to the caster object size.

One thing I noticed when I looked at this problem a long time ago (I never had a good solution so I never sent a PR) was that the scale of the cast effect seemed to scale very exactly with the player. A dark elf male, which has height and weight of scale 1.0.,  had a cast effect also of scale 1.0, as checked with the command "showscenegraph". A character with a height or weight not of 1.0, like 1.1, had the effect match that scale. It made me wonder if a 1.0 scaled character model was the base for the scaling calculation. The effect scales very huge, although the particles stay small, if you have a large object like a building or silt strider cast a spell.

Allofich:
scale
Here's an example of what I meant. The male nord has height 1.06, and you can see the effect has m_WorldScale 1.06 as well.

akortunov:
> The effect scales very huge, although the particles stay small

I wonder how we can achieve such behaviour.

Allofich:
> I wonder how we can achieve such behaviour.

I don't know. scrawl, help! :)

akortunov:
> partsys->setParticleScaleReferenceFrame(osgParticle::ParticleSystem::WORLD_COORDINATES)

Yes, it works. But how to determine when we need to scale particles, and when not...

scrawl:
>  How can I calculate the mesh dimesions? Via osg::ComputeBoundsVisitor?

Yes, that would work. Although for actors, the bounding box would vary based on the current animation pose, whether they're carrying a torch that emits particles 2 meters above them, etc. We might want a different solution for actors.

> But how to determine when we need to scale particles, and when not...

Just ignore this for now, does not seem important to me.

akortunov:
I uploaded a scaling demo.
Known issues:
1. Vanilla engine calculates Z coordinate for particle a bit strange. I did not figured out how it works yet, so I had to take an arbitrary size as 1.0 scale and place effect at minimum Z coordinate of casting object.
2. Vanilla engine also rotates casting effect to fit casting object rotation.
3. We still need a way to tell the particle system to do not scale particles size.
4. For some reason an engine sets the effect size for creatures as `casterSize*effectScale*casterScale`, so i had to use a dirty workaround. For other objects effect size is the `casterSize*effectScale`.

akortunov:
> Yes, that would work. Although for actors, the bounding box would vary based on the current animation pose, whether they're carrying a torch that has particles 2 meters above them, etc. We might want a different solution for actors

For NPCs I just use race height as scale factor. Not sure how to handle other actors.

akortunov:
Reworked feature a bit.
I get mesh from resourse manager, so mesh dimensions should not depend on animation state.
Also this approach allows to play casting effect, if the caster is disabled

scrawl:
Technically, you should use getTemplate() here not getInstance(). There's no need to create an instance if you're not doing any modifications to it. The problem with this is that ComputeBoundsVisitor (osg::NodeVisitor) doesn't support const objects, though. This is where I feel a const_cast would be legitimately warranted.

akortunov:
Mostly works now. Any suggestions or testing would be welcome.

zinnschlag:
The initial description says the issue only partially resolved. But the issue is flagged as resolved. Is there still something missing?

akortunov:
> The initial description says the issue only partially resolved. But the issue is flagged as resolved. Is there still something missing?

I forgot to modifiy description.
About issues: basically the feature is complete, but there are issues with particle scaling ([bug #3559](https://bugs.openmw.org/issues/3559)).
Also I am not sure that effect scaling works is the same way in vanilla game for large objects (e.g. daedric lords statues).

akortunov:
> This PR is unlikely to break anything new, right?

Well, nothing obvious at least, but community community testing would be nice.

apps/openmw/mwrender/objects.cpp
```diff
@@ -138,6 +138,14 @@ bool Objects::removeObject (const MWWorld::Ptr& ptr)
     return false;
 }
 
+void Objects::updateEffects(float duration)
+{
+    for(PtrAnimationMap::iterator iter = mObjects.begin();iter != mObjects.end();)
```
scrawl: We should really avoid iterating all objects every frame because there are thousands of them in a normal scene. I remember we had a performance issue caused by the same thing in PhysicsSystem::animateCollisionShapes and it was fixed by tracking animated shapes in a separate container.
Usually the osg::UpdateCallback is a very nice way to do updates every frame. Maybe the updateEffects should be ported to an UpdateCallback?

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1535,6 +1535,11 @@ namespace MWWorld
         }
     }
 
+    osg::ref_ptr<osg::Node> World::getInstance (const std::string& modelName)
```
scrawl: World isn't supposed to expose rendering details. Can we remove that function and move whatever uses it inside the rendering system?

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -1009,18 +1016,72 @@ namespace MWMechanics
...
+                const ESM::Race *race =
+                    esmStore.get<ESM::Race>().find(npc->mRace);
 
-                std::string texture = effect->mParticle;
+                scale = npc->isMale() ? race->mData.mHeight.mMale : race->mData.mHeight.mFemale;
+            }
+            else
+            {
+                std::string casterModel = mCaster.getClass().getModel(mCaster);
+                osg::ref_ptr<osg::Node> model = MWBase::Environment::get().getWorld()->getInstance(casterModel);
```
scrawl: As I said before we don't need to create an instance just to get the bounds of the model, that's wasted resources. Template is sufficient. Try something like this:
```
osg::ref_ptr<const osg::Node> node = sceneMgr->getTemplate(name);
computeBoundsVisitor.accept(const_cast<osg::Node*>(node.get()));
```
The const_cast is unfortunate, but I'll prefer it over a loss of performance.

apps/openmw/mwrender/effectmanager.cpp
```diff
@@ -26,6 +26,13 @@ EffectManager::~EffectManager()
 }
 
 void EffectManager::addEffect(const std::string &model, const std::string& textureOverride, const osg::Vec3f &worldPosition, float scale, bool isMagicVFX)
+{
+    osg::Vec3f scale3f (scale, scale, scale);
+
+    addEffect(model, textureOverride, worldPosition, scale3f, isMagicVFX);
```
scrawl: Why add the overload for non-uniform scales when it's not being used (as far as I can tell)?

PR #1576, opened 2017-12-04
Journal index fixes

akortunov:
1. Fixes [bug #4242](https://bugs.openmw.org/issues/4242). Now we add only dialogue responses to the journal index.
2. A small fix for cyrillic journal index - it supposed to do not display cyrillic hard/soft signs, but I broke this feature after refactoring. Sorry, I was blind.

### DIFF
--- a/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
+++ b/apps/openmw/mwdialogue/dialoguemanagerimp.cpp
@@ -146,7 +146,6 @@ namespace MWDialogue
                         // TODO play sound
                     }
 
-
                     MWScript::InterpreterContext interpreterContext(&mActor.getRefData().getLocals(),mActor);
                     callback->addResponse("", Interpreter::fixDefinesDialog(info->mResponse, interpreterContext));
                     executeScript (info->mResultScript, mActor);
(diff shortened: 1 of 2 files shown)


apps/openmw/mwdialogue/dialoguemanagerimp.cpp
```diff
@@ -401,15 +400,18 @@ namespace MWDialogue
...
 
-                    // Make sure the returned DialInfo is from the Dialogue we supplied. If could also be from the Info refusal group,
-                    // in which case it should not be added to the journal.
-                    for (ESM::Dialogue::InfoContainer::const_iterator iter = dialogue->mInfo.begin();
-                        iter!=dialogue->mInfo.end(); ++iter)
+                    if (dialogue->mType == ESM::Dialogue::Topic)
                     {
-                        if (iter->mId == info->mId)
+                        // Make sure the returned DialInfo is from the Dialogue we supplied. If could also be from the Greeting group,
+                        // in which case it should not be added to the journal.
```
scrawl: This comment should not be changed - the loop below it is still there to filter out Info Refusal topics.

akortunov: If you say so...

PR #1578, opened 2017-12-05
[Feedback needed] Do not change z-coordinate of the camera during rotation (1stPerson)

thegriglat:
An effort to fix bug #4241 (https://bugs.openmw.org/issues/4241)

position.z() in 1st person mode is set to z-coordinate of mTrackingPtr.getRefData().getBaseNode() instead of mTrackingNode.
position.x() and position.y() are taken from mTrackingNode as before.

z-coordinate of the camera seems to be close to z of eyes.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -1280,6 +1280,8 @@ namespace MWMechanics
                                 {
                                     if (it->first == iter->first)
                                         continue;
+                                    if (iter->first == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson())
+                                        continue;
                                     updateHeadTracking(iter->first, it->first, headTrackTarget, sqrHeadTrackDistance);
                                 }
                             }
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -1276,14 +1276,13 @@ namespace MWMechanics
                                 !stats.getAiSequence().isInCombat() &&
                                 !stats.getAiSequence().hasPackage(AiPackage::TypeIdPursue))
                             {
-                                for(PtrActorMap::iterator it(mActors.begin()); it != mActors.end(); ++it)
-                                {
-                                    if (it->first == iter->first)
-                                        continue;
-                                    if (iter->first == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson())
-                                        continue;
-                                    updateHeadTracking(iter->first, it->first, headTrackTarget, sqrHeadTrackDistance);
-                                }
+                                if (!(iter->first == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson()))
+                                    for(PtrActorMap::iterator it(mActors.begin()); it != mActors.end(); ++it)
+                                    {
+                                        if (it->first == iter->first)
+                                            continue;
+                                        updateHeadTracking(iter->first, it->first, headTrackTarget, sqrHeadTrackDistance);
+                                    }
                             }
 
                             iter->second->getCharacterController()->setHeadTrackTarget(headTrackTarget);
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -2421,10 +2421,6 @@ void CharacterController::updateHeadTracking(float duration)
     if (!head)
         return;
 
-    // Disable head tracking for player in First person mode
-    if (mPtr == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson())
-        return;
-
     float zAngleRadians = 0.f;
     float xAngleRadians = 0.f;
 
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -2421,6 +2421,10 @@ void CharacterController::updateHeadTracking(float duration)
     if (!head)
         return;
 
+    // Disable head tracking for player in First person mode
+    if (mPtr == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson())
+        return;
+
     float zAngleRadians = 0.f;
     float xAngleRadians = 0.f;
 
--- a/apps/openmw/mwrender/camera.cpp
+++ b/apps/openmw/mwrender/camera.cpp
@@ -3,6 +3,7 @@
 #include <osg/Camera>
 
 #include <components/sceneutil/positionattitudetransform.hpp>
+#include <components/esm/loadcell.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/windowmanager.hpp"
@@ -101,6 +102,8 @@ namespace MWRender
             position.z() += mHeight * mHeightScale;
         else
         {
+            if (!mTrackingPtr.getCell()->getCell()->hasWater())
+                return position;
             const osg::Node* baseNode = mAnimation->getNode("Right Foot");
             if (!baseNode || baseNode->getParentalNodePaths().empty())
                 return position;
--- a/apps/openmw/mwrender/camera.cpp
+++ b/apps/openmw/mwrender/camera.cpp
@@ -98,10 +98,13 @@ namespace MWRender
         osg::Vec3d position = worldMat.getTrans();
         if (!isFirstPerson())
             position.z() += mHeight * mHeightScale;
-        else {
-            if (mTrackingPtr.getRefData().getBaseNode() && !mTrackingPtr.getRefData().getBaseNode()->getParentalNodePaths().empty()){
-                osg::Vec3d position_base = osg::computeLocalToWorld(mTrackingPtr.getRefData().getBaseNode()->getParentalNodePaths()[0]).getTrans();
-                position.z() = mHeight * mHeightScale + position_base.z();
+        else
+        {
+            const osg::Node* baseNode = mTrackingPtr.getRefData().getBaseNode();
+            if (baseNode && !baseNode->getParentalNodePaths().empty())
+            {
+                const osg::Vec3d positionBase = osg::computeLocalToWorld(baseNode->getParentalNodePaths()[0]).getTrans();
+                position.z() = mHeight * mHeightScale + positionBase.z();
             }
         }
         return position;
--- a/apps/openmw/mwrender/camera.cpp
+++ b/apps/openmw/mwrender/camera.cpp
@@ -9,6 +9,7 @@
 
 #include "../mwworld/ptr.hpp"
 #include "../mwworld/refdata.hpp"
+#include "../mwworld/cellstore.hpp"
 
 #include "npcanimation.hpp"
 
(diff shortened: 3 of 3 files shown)


drummyfish:
for others: [forum thread](https://forum.openmw.org/viewtopic.php?f=6&t=4798) with some investigation and [vanilla footage](https://www.youtube.com/watch?v=b7S90TjZeHs&feature=youtu.be)

apps/openmw/mwrender/camera.cpp
```diff
@@ -98,6 +98,12 @@ namespace MWRender
         osg::Vec3d position = worldMat.getTrans();
         if (!isFirstPerson())
             position.z() += mHeight * mHeightScale;
+        else {
```
drummyfish: It's OK to have the code formatted like this now, but when it's about to get merged it should be formatted according to [this](https://wiki.openmw.org/index.php?title=Code_Formatting_Conventions), i.e. opening/closing brackets at the same level, camelcase (use `positionBase` instead of `position_base` etc.).

apps/openmw/mwrender/camera.cpp
```diff
@@ -98,6 +98,12 @@ namespace MWRender
...
+            if (mTrackingPtr.getRefData().getBaseNode() && !mTrackingPtr.getRefData().getBaseNode()->getParentalNodePaths().empty()){
```
drummyfish: You could use local variable to store `mTrackingPtr.getRefData().getBaseNode()` so that you don't have to repeat it twice here and once below.

drummyfish:
Interesting solution BTW :) Though the camera now seems to be placed exactly at the water-level surface. The original bug is gone but sometimes when I resurface the underwater effect stays on and doesn't change if I rotate the camera, even though I'm considered being above water (there's no breath bar etc.).

apps/openmw/mwrender/camera.cpp
thegriglat: Ok

drummyfish:
OK, thanks to my new super uber [library](https://github.com/drummyfish/osgdebug) I think I know what's causing the bug:

- The camera position is set to position of the head (via `mTrackingNode`).
- Head is attached to the neck and translated a bit up and forward against it.
- When you look up/down, the neck rotates to make the head rotate up/down (I've tried this by vatching the transforms of all parent nodes of head, and neck is the only node changing when looking up/down).
- This causes the head node to go up and down like this:
image
- In extreme cases (looking completely up or completely down) the head is lower than in normal situation (looking forward), which causes the head go under the water surface (taking the camera with it).

As a bonus here are the parents of the head node:

```
PARENTS OF NODE:
  Camera (): 1 children
    Group (): 9 children
      LightManager (Scene Root): 14 children
        PositionAttitudeTransform (Player Root): pos = Vec3: 5185337101024084505919488.00000 0.00000 0.00000 , rot = (-132.81438 178.52333 0.64508), piv = Vec3: 0.00000 inf 0.00000 
          Skeleton (FileRoot): 3 children
            MatrixTransform (ModelRoot): transform (TRS): Vec3: 0.00000 0.00000 0.00000 , (0.00000 0.00000 -0.00000), Vec3: 1.00000 1.00000 1.00000 
              MatrixTransform (Bip01): transform (TRS): Vec3: -0.00000 -0.00000 76.72200 , (87.46908 0.24835 0.00549), Vec3: 1.00000 1.00000 1.00000 
                MatrixTransform (Bip01 Pelvis): transform (TRS): Vec3: 0.00000 0.00000 -0.00000 , (-90.00001 0.00009 -87.99992), Vec3: 1.00000 1.00000 1.00000 
                  MatrixTransform (Bip01 Spine): transform (TRS): Vec3: 7.60739 -0.00672 0.26553 , (-1.94934 1.24046 -2.10471), Vec3: 1.00000 1.00000 1.00000 
                    MatrixTransform (Bip01 Spine1): transform (TRS): Vec3: 8.43644 -0.00894 0.00190 , (-0.00100 -11.99992 0.09403), Vec3: 1.00000 1.00000 1.00000 
                      MatrixTransform (Bip01 Spine2): transform (TRS): Vec3: 11.47311 -0.00829 -0.00000 , (-3.00000 0.00011 0.14092), Vec3: 1.00000 1.00000 1.00000 
                        MatrixTransform (Bip01 Neck): transform (TRS): Vec3: 10.43088 -0.00394 -0.00000 , (9.47124 -0.09413 -0.93814), Vec3: 1.00000 1.00000 1.00000 
                          MatrixTransform (Bip01 Head): transform (TRS): Vec3: 4.96451 0.00000 -0.00000 , (2.72822 9.58463 -1.09590), Vec3: 1.00000 1.00000 1.00000 
                            MatrixTransform (Head): transform (TRS): Vec3: 5.63873 1.29681 0.00000 , (-92.87163 -87.12826 nan), Vec3: 1.00000 1.00000 1.00000 
```

So what's the solution? I'm not sure, but maybe rotate head instead of neck. Or maybe don't rotate nor head nor neck in first person.

thegriglat:
It is really super uber library! It saves a lot of time for debugging!
I tried do not rotate head/neck or rotate just head but in this case the camera is placed exactly at the x,y-center of the player (what is unrealistic) and you see hands from the inside of the player.
I think we definitely need camera translated together with a neck, as it is in life :-)

Maybe we can just limit rotation angle for the neck? For human it is unreal to rotate neck at an angle of 90 degrees to the body (cases 2) and 3) from comment above)

thegriglat:
Ideally, we should rotate neck and then (after some limit) head (tested on my head) ...

drummyfish:
I tried disabling the neck rotation too and yes, it does look weird :)

> Ideally, we should rotate neck and then (after some limit) head (tested on my head) ...

We can't see the neck and the head in FP so anatomically correct behavior is not needed I think, I'd prefer simpler solution that fixes the bug and still keeps the hands in first person looking good. I would like to examine how exactly this head rotation is working in vanilla but we can't since there's no source code, so we have to just experiment I guess. Maybe we could keep it like it is and just limit the Z position to not go below it's initial Z-position?

drummyfish:
OK, take a look at [this](https://www.youtube.com/watch?v=pKwA-cAiMXs&feature=youtu.be) - that's how the head behaves in first person. It's definitely rotating in a weird way, for example if you look completely up and then keep turning left/right, the head keeps going above/under water. I think it's the quaternions which change roll of the rotation unpredictably. Also I find it weird that [here](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwrender/camera.cpp#L117) the up vector is transformed with `orient`, I think up should always be (0,0,1). I'll try to play around with that.

drummyfish:
[Here](https://www.youtube.com/watch?v=ZzyL2FcUvk4&feature=youtu.be) I have captured the problem of head animation in first person:

- The numbers in console are the camera position.
- I'm not moving my mouse (except when obviously I am).
- In the first part I'm partially in the wall to make you see the head twitching on its own.
- In the second part I'm being attacked by slaughterfish in the water - note that the camera position changes more significantly here as the head keeps rotating to face the enemies.

I'd say let's try to fix this :)

akortunov:
> note that the camera position changes more significantly here as the head keeps rotating to face the enemies.

[Here](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/actors.cpp#L1284) you can add the check if the tracking actor is a player and the view is a first-person.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -2421,6 +2421,10 @@ void CharacterController::updateHeadTracking(float duration)
     if (!head)
         return;
 
+    // Disable head tracking for player in First person mode
+    if (mPtr == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson())
+        return;
```
drummyfish: Maybe add setting the head to non-rotated state here before return (I'm guessing with `mAnimation->setHeadPitch` and `mAnimation->setHeadYaw`). Because this only stops rotating the head, but if it's already rotated, e.g. switching to FP during combat, it will stay that way.

akortunov: You can check how I did for unconscious actors in the actors.cpp - I call the setHeadTrackTarget() with empty target.

drummyfish: [Here](https://github.com/OpenMW/openmw/pull/1550/files), right?

akortunov: Yes.

drummyfish:
I just checked, using my lib again, and the head tracking bug seems to be solved, so good job, you've got one issue fixed I think. The head is still twitching a bit though, but it's caused by the player root node position twitches, it may be something with physics, which could be left for another pull request. All parents of the head node all the way up to the player root is stationary in first person (I also checked without this fix and the head indeed was tracking).

apps/openmw/mwmechanics/character.cpp
thegriglat: So should I move code to another place or not?

drummyfish: I'm a little drunk now so... but I guess whichever is simpler that achieves what I said in the first comments, it's up to you.

thegriglat:
There is an issue with commit f4a0ede -- after switching from 1st to 3rd view player's head starts to rotate to NPC from default (forward) position.

thegriglat:
strange, AppVeyor build failed because of downloading bullet

drummyfish:
That happens sometimes, nothing to worry about.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -1280,6 +1280,8 @@ namespace MWMechanics
                                 {
                                     if (it->first == iter->first)
                                         continue;
+                                    if (iter->first == getPlayer() && MWBase::Environment::get().getWorld()->isFirstPerson())
```
drummyfish: If I understand the code correctly, this condition would be better moved to the "if" above (i.e. after line 1277), because if this is false, then the whole for loop doesn't have to be run.

thegriglat:
A few comments about the previous commit

1. Using "Right Foot" is a bit strange, but I tried to use baseNode and faced with a problem. z() of baseNode depends on player's state -- idle or weapon-ready. "Right foot" also depends but less.
2. After some load cycles game segfaults on `mTrackingPtr.getCell()->getWaterLevel()` line. I tried to check `mTrackingPtr.getCell()` but it didn't help.
Why? How to check that cell is fully loaded? And should I check that cell has water (as I understand via `hasWater()` from `components/esm/loadcell.hpp`)?

thegriglat:
I have also a question -- as I can see camera starts to update (via `updateCamera()`) even before save game is fully loaded. Is it expected behavior?
In fact, this causes a number of (maybe unnecessary) checks in `getFocalPoint()`.

thegriglat:
@drummyfish @akortunov, could you please review commits again?

drummyfish:
I'll try to have a look today.

apps/openmw/mwrender/camera.cpp
```diff
@@ -98,6 +100,23 @@ namespace MWRender
         osg::Vec3d position = worldMat.getTrans();
         if (!isFirstPerson())
             position.z() += mHeight * mHeightScale;
+        else
+        {
+            if (mTrackingPtr.getCell()->getState() != MWWorld::CellStore::State_Loaded)
+                return position;
+            if (!mTrackingPtr.getCell()->getCell()->hasWater())
```
drummyfish: Could be put together with the above condition with ||.

drummyfish:
It works, though the solution again doesn't address the source of the bug, only the resulting behavior. But I don't know what else to do now, so... @scrawl could you take a look at this?

thegriglat:
Is there a way to re-launch automated checks (checks failed due to bullet downloading...)
And is there a chance to merge this?

wareya:
related to https://gitlab.com/OpenMW/openmw/issues/4448

akortunov:
Just a guess: can we do not rotate neck in the first person mode, just rotate camera directly?

wareya:
Vanilla doesn't reposition the camera **at all** in first person mode.

akortunov:
"Bip01 Head" has two children: "Camera" and "Head".
Currently we rotate neck, so both "Camera" and "Head" nodes are affected.
Just "do not reposition camera" will not help us since we still need to rotate hands at least.

akortunov:
Since this PR does not solve the root of this issue, I see no point to keep it.

PR #1579, opened 2017-12-05
Stop guards trying to arrest player when calm

Allofich:
Simple fix for https://bugs.openmw.org/issues/3863.

There is still a minor issue you may see, but it's unrelated. If you have a high enough bounty to make guards attack you, when you talk to a guard that you have placed a Calm effect on, the guard will keep starting to enter combat and then stopping. That issue is not guard-related, though. You can see the same thing happen if you calm Furius Acilius (who is a Warrior class, not Guard) and talk to him. That issue seems to be with StartCombat when it's in a dialogue script, like it keeps being executed over and over.

In the original engine you get a "This character is in combat" message if you try to talk to a calmed guard with the high bounty or to calmed Furius Acilius after his dialogue script has run.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -1653,6 +1653,12 @@ namespace MWMechanics
 
     bool MechanicsManager::isAggressive(const MWWorld::Ptr &ptr, const MWWorld::Ptr &target)
     {
+        // Don't become aggressive if a calm effect is active, since it would cause combat to cycle on/off as
+        // combat is activated here and then canceled by the calm effect
+        if ((ptr.getClass().isNpc() && ptr.getClass().getCreatureStats(ptr).getMagicEffects().get(ESM::MagicEffect::CalmHumanoid).getMagnitude() > 0)
+            || (!ptr.getClass().isNpc() && ptr.getClass().getCreatureStats(ptr).getMagicEffects().get(ESM::MagicEffect::CalmCreature).getMagnitude() > 0))
+            return false;
+
         int disposition = 50;
         if (ptr.getClass().isNpc())
             disposition = getDerivedDisposition(ptr, true);
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -1009,7 +1009,8 @@ namespace MWMechanics
             if (player.getClass().getNpcStats(player).isWerewolf())
                 return;
 
-            if (ptr.getClass().isClass(ptr, "Guard") && creatureStats.getAiSequence().getTypeId() != AiPackage::TypeIdPursue && !creatureStats.getAiSequence().isInCombat())
+            if (ptr.getClass().isClass(ptr, "Guard") && creatureStats.getAiSequence().getTypeId() != AiPackage::TypeIdPursue && !creatureStats.getAiSequence().isInCombat()
+                && creatureStats.getMagicEffects().get(ESM::MagicEffect::CalmHumanoid).getMagnitude() == 0)
             {
                 const MWWorld::ESMStore& esmStore = MWBase::Environment::get().getWorld()->getStore();
                 static const int cutoff = esmStore.get<ESM::GameSetting>().find("iCrimeThreshold")->getInt();


akortunov:
> the guard will keep starting to enter combat and then stopping

That's because the NPC starts fighting in the engageCombat(), but then immediately stops in the calculateCreatureStatModifiers().

Allofich:
Do you know why they start again and again? It's as if the "StartCombat" in the dialogue script is run repeatedly.

In the case of the guard, the script attached to the "death warrant" dialogue has as a condition that the player bounty is >= 5000, and you can stop the guard entering/leaving combat with SetPCCrimeLevel to less than 5000. It seems like the script is being executed repeatedly.

akortunov:
> Do you know why they start again and again? It's as if the "StartCombat" in the dialogue script is run repeatedly.

In my testing, I can just "setfight 100" to any NPC and then calm him to reproduce this bug.
The actor is hostile, so he starts a combat in every actor update. But he has a Calm effect, so he stops combat in the every spell effects update.

I am not sure how to fix it properly.

Allofich:
Ok, I also recreated it. For me at least, I need to calm the NPC first, then setfight 100, to get the behavior. If I use setfight 100 first (and the NPC starts attacking) and then I use calm, it doesn't happen.

Allofich:
Original game has the same issue, except that the npc won't try to draw their weapon. They will keep saying their combat taunt and the combat music will briefly start/stop.

I didn't notice it with the guard because the "This character is in combat" message prevented me from talking to him, which prevented the "death warrant" dialogue's script from setting his fight to 100. But if I manually set any calmed NPC to fight 100, they do the behavior.

It's easy to fix, just add a check for the calm effect in isAggressive() in mechanicsmanagerimp.cpp. But should we fix it? Seems like maybe we should. Only downside I can think of is it allowing you to get by hostile NPCs using the calm spell without them briefly starting combat, which could maybe affect some stealth mechanic in a mod or something, but that seems like a stretch.

Allofich:
I think we should fix it. I did a bit more testing, and it seems like normally you won't get this behavior. It only happens if the fight value is very high during a calm effect, which normally won't happen.

Even if a character has 100 fight, if you use the standard "Calm Humanoid" spell (-30 fight), they will not have the "on/off" behavior in the original engine with 70 fight and the active calm effect. It happens if I force the value to 99, 100, etc. using "setfight". And anyway, it looks and sounds buggy, so I think we should fix it. I'll go ahead and add it here.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1653,6 +1653,12 @@ namespace MWMechanics
 
     bool MechanicsManager::isAggressive(const MWWorld::Ptr &ptr, const MWWorld::Ptr &target)
     {
+        // Don't become aggressive if a calm effect is active, since it would cause combat to cycle on/off as
+        // combat is activated here and then canceled by the calm effect
+        if ((ptr.getClass().isNpc() && ptr.getClass().getCreatureStats(ptr).getMagicEffects().get(ESM::MagicEffect::CalmHumanoid).getMagnitude() > 0)
+            || (!ptr.getClass().isNpc() && ptr.getClass().getCreatureStats(ptr).getMagicEffects().get(ESM::MagicEffect::CalmCreature).getMagnitude() > 0))
+            return false;
```
akortunov: This solution does not look good for me. It seems calmed actor will not fight back at all, what looks like an exploit.

Allofich: Yes, I agree it seems like an exploit. The same exploit exists in the original game and in OpenMW master branch. If you mean this code you pointed to here causes it, I think it's actually caused by the check in actors.cpp to stop combat if an actor has any calm effect. This code just removes the "stuttering" effect of entering combat and immediately ending it.

If we change the calm effect, I think it should be a different PR. 
An example of how calm could work:
- After reducing fight value according to the spell magnitude (as is done now), instead of necessarily making the actor stop combat, have them re-evaluate whether to do combat according to fight level, disposition, etc.
- If an actor with the calm effect is hit, remove the calm effect.

This way different magnitudes of calm will have meaning, and the exploit will be removed. it seems to me more like the way the spell should work.

If it was done this way, the "stuttering" fix (second commit) could be removed. Do you think I should just remove it now?

akortunov: > The same exploit exists in the original game

Strange. I thought if you attack a calmed NPC, he will fight you. But it seems you are right.
So your fix is fine, if we want to replicate vanilla behaviour.

PR #1580, opened 2017-12-06
Add setting option to scale water refractions

drummyfish:
[forum thread](https://forum.openmw.org/viewtopic.php?f=6&t=4810)

Adds option to make water scale down objects for better realism:

It's off by default.

Another screenshot, note the oar:

### DIFF
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -272,6 +272,12 @@ public:
 
     void setWaterLevel(float waterLevel)
     {
+        const float refractionScale = std::min(1.0f,std::max(0.0f,
+            Settings::Manager::getFloat("refraction scale", "Water")));
+
+        setViewMatrix(osg::Matrix::scale(1,1,refractionScale) *
+            osg::Matrix::translate(0,0,(1.0 - refractionScale) * waterLevel));
+
         mClipCullNode->setPlane(osg::Plane(osg::Vec3d(0,0,-1), osg::Vec3d(0,0, waterLevel)));
     }
 
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -272,7 +272,7 @@ public:
 
     void setWaterLevel(float waterLevel)
     {
-        const float refractionScale = std::min(1.0f,std::max(0.0f,
+        const float refractionScale = std::min(1.0f,std::max(0.01f,     // 0.0 crashes the game
             Settings::Manager::getFloat("refraction scale", "Water")));
 
         setViewMatrix(osg::Matrix::scale(1,1,refractionScale) *
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -272,7 +272,7 @@ public:
 
     void setWaterLevel(float waterLevel)
     {
-        const float refractionScale = std::min(1.0f,std::max(0.01f,     // 0.0 crashes the game
+        const float refractionScale = std::min(1.0f,std::max(0.0f,
             Settings::Manager::getFloat("refraction scale", "Water")));
 
         setViewMatrix(osg::Matrix::scale(1,1,refractionScale) *
--- a/docs/source/reference/modding/settings/water.rst
+++ b/docs/source/reference/modding/settings/water.rst
@@ -88,3 +88,20 @@ This setting will have no effect if the shader setting is false,
 or the 'small feature culling' (in the 'Camera' section) is disabled.
 
 This setting can only be configured by editing the settings configuration file.
+
+refraction scale
+----------------
+
+:Type:		floating point
+:Range:		0 to 1
+:Default:	1.0
+
+Simulates light rays refracting when transitioning from air to water, which causes the space under water look scaled down
+in height when viewed from above the water surface. Though adding realism, the setting can cause distortion which can
+make for example aiming at enemies in water more challenging, so it is off by default (i.e. set to 1.0).
+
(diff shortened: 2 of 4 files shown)


thegriglat:
Nice feature!
Maybe better implement the option in in-game menu (e.g. in Water tab) instead of in configuration file?

drummyfish:
I was thinking that too, will try adding a small slider.

Capostrophic:
@thegriglat, technically every in-game slider has a configuration option in settings.cfg which would have to be implemented anyway, a slider just changes the value of its respective setting from the recommended minimum to the recommended maximum.
But not every setting has a slider, because it's much more easy to add a setting to the text files than to the in-game options.
@drummyfish, make sure to add the description of the setting to the documentation.

drummyfish:
> @drummyfish, make sure to add the description of the setting to the documentation.

Thanks, I added some basic docs, will complete it after I decide whether or not to add the GUI slider.

drummyfish:
[done](https://www.youtube.com/watch?v=Cy6y3wIN8uo&feature=youtu.be)

Capostrophic:
This looks weirdly satisfying.

thegriglat:
I mean on/off switcher ...

drummyfish:
More configuration possibilities is always better I guess, would be nice to know the exact value for air/water though. Here, if you're bored, compute the value 1 - X, use [Snell's law](https://en.wikipedia.org/wiki/Snell%27s_law) and refracting indices of air (1) and water (1.33):

image

scrawl:
> So the value should be set to 0.75 to look like in reality.

If that's how it is, do we need an option at all?

Did you try this with extreme angles (e.g. looking straight down?)

Are there any other games that do this? Are we sure that this isn't going to be another 'awkwardly realistic' thing (where the end result looks awkward because some parts are realistic but others aren't) like [snell's window](https://bugs.openmw.org/issues/816)?

Note that using non-uniform scaling has the potential to conflict with code that isn't properly accounting for it, e.g. a lighting shader, and more generally just using a model matrix instead of the inverse transpose of the matrix. Although, if the scale is small-ish (like 0.75) then any artifacts may or may not be noticable. I know that character models (which are scaled non-uniformly by factors like 1.1) appear to work fine, but I haven't done any testing with larger scales or other models/features.

drummyfish:
> If that's how it is, do we need an option at all?

Well I don't wouldn't want to have any hardcoded values, some players might want to go for a compromise between realism and distortion, modders might want to do other fluids than water etc. - what we might do could be have the float value in settings and in GUI only have a checkbox, instead of slider, that would turn on/off the 0.75 value. Advanced users could just go to the settings file.

> Did you try this with extreme angles (e.g. looking straight down?)

I think so, I kinda liked it so I was testing it in a lot of places and am playing with it turned on now all the time.

> Are there any other games that do this?

I don't know, let me see.

> Are we sure that this isn't going to be another 'awkwardly realistic' thing (where the end result looks awkward because some parts are realistic but others aren't) like snell's window?

Well unlike the snell's window this is completely optional and off by default, so I guess no one will complain.
> Note that using non-uniform scaling has the potential to conflict with code that isn't properly accounting for it, e.g. a lighting shader, and more generally just using a model matrix instead of the inverse transpose of the matrix.

I haven't noticed any problems but will try to test this more.

lysol90:
I still think this feature is great, and if some people think it looks strange, then make it an optional one. I think a slider feels a bit too much however, so I'd be happy with just a toggle that turns on or off the 0.75 value in the cfg file.

scrawl:
I've tried it out (with 0.7) and to be honest, don't like it. For me the effect is just disorienting but doesn't do anything to improve visual fidelity.

We can still have the optional setting but, the in-game slider should be removed imo. Noobs will play with it, set an absurd value, then later complain that their game is broken.

Also, the documentation should state the recommended 0.7 value or whatever it was you arrived at.

Thunderforge:
> We can still have the optional setting but, the in-game slider should be removed imo. Noobs will play with it, set an absurd value, then later complain that their game is broken.

> Also, the documentation should state the recommended 0.7 value or whatever it was you arrived at.

@scrawl Would you still have the same concern if the slider had a range from 0.0 to 0.75 (low end is no refraction,  high end is real world refraction), with 0.0 being the default? That way, the only way to set an absurd value that exceeds real-world is by editing the setting file manually. I don't see any reason really to set this value higher than real-world, so making that be the high end makes sense.

If it would still be a concern, could you elaborate how new players would say their game is "broken" if they deliberately changed the value? I'm not understanding what you mean by that.

scrawl:
Hm, right, that sounds like a more user-friendly slider, although the conversion from 'user friendly' to 'actual value' gets confusing then (something like remap to 0-1, invert, then map to 0-0.75?)

Regardless, let's keep the menu simple for now. I think it's best to introduce new things as a settings file only-option and then potentially move them to the ingame menu once they've undergone more testing and seen as very useful by the community.

I will just give this PR one more test then merge it.

Thunderforge:
> Hm, right, that sounds like a more user-friendly slider, although the conversion from 'user friendly' to 'actual value' gets confusing then (something like remap to 0-1, invert, then map to 0-0.75?)

I agree, having numbers is unintuitive. What I would probably recommend would be a slider like with View Distance where you don't have numbers visible to the user, just two labels:

screen shot 2018-01-02 at 9 01 48 pm

Basically, one end would be "None", the other would be "Real-world" or something like that, then the user would just pick some value along that slider. Under the hood, the numbers would be in the range [0.0, 0.75], but the user would never see that.

> Regardless, let's keep the menu simple for now. I think it's best to introduce new things as a settings file only-option and then potentially move them to the ingame menu once they've undergone more testing and seen as very useful by the community.

I'm fine with this.

PR #1582, opened 2017-12-07
CS: Camera improvements

drummyfish:
I added a new preference section for Rendering ([#1310](http://bugs.openmw.org/issues/1310)) in which I put setting for camera FOV. Additionally I tried to fix an issue with orbiting camera mode which changed its roll and was practically unusable, at least to me. It now orbits only in yaw and pitch, while keeping the roll set with Q/E keys.


zinnschlag:
> CS: Add rendering prefs and camera FOV

Hm, wondering. Would it make sense to have separate FOV settings for each camera mode?

> CS: Make orbit camera not change roll

Maybe bind this feature to a new user setting (on by default)?

drummyfish:
> Hm, wondering. Would it make sense to have separate FOV settings for each camera mode?

I don't think that would really be useful for anything, can't see any use case, but you made me think maybe each separate 3D view could have a different FOV? That would make sense I guess. The FOV setting I've added now would be the default FOV of each new view, and then in each view you could adjust it differently (But how? Some key + mouse wheel? Or add a visual element for it?).

Also now when we're elaborating on this, I think I should also add the orthographic camera option, which could then also be toggled per-view. That would allow for really nice multi-view settings we know from 3D editors (e.g. [this](https://upload.wikimedia.org/wikipedia/commons/thumb/e/ef/Blender3D_4Views_3DWindow-2.49.png/600px-Blender3D_4Views_3DWindow-2.49.png)).

> Maybe bind this feature to a new user setting (on by default)?

Sorry I don't understand this, could you explain more?

drummyfish:
Finding more bugs:

- When free camera is reset, the roll can't be changed.
- In orbit mode, when I get to limit top position with W key, the camera starts shaking and then if I press the S key, I sometimes "roll over the top" to the other side.

(Don't know if it's just my branch, will test.)

drummyfish:
Another thing: the pref strings ("Rendering/camera-fov" etc.) are repeated over in multiple files so you have to make potential changes at multiple places => bug prone. Would be better to have a separate header with these pref string definitions.

drummyfish:
I have the ortho camera now.

image

But I think 3D picking isn't working now, so that's the next thing.

drummyfish:
So I've added the option to turn on the old orbit behavior, which is off by default. I'm kinda happy with the camera now.

drummyfish:
Could maybe the parameter settings for individual views look like this?

image

zinnschlag:
Nope. That would mix up modes and mode settings. Do we actually need this in the scene UI. I thought the regular user settings UI would be doing fine. I can't come up with a use case where you would want to change FOV settings on the fly.
However if there are use cases that I missed, the right way to do it would be via the button context menu. There are a couple of examples with other buttons on how to do this.

drummyfish:
> Nope. That would mix up modes and mode settings.

Yep I thought that would be a problem, I just couldn't come up with anything else. I'm scratching that.

> Do we actually need this in the scene UI. I thought the regular user settings UI would be doing fine.

You mean the preferences that I have now? Yes, they're good for setting the default values, but I was still thinking about having the additional ability to adjust the parameters per each opened 3D view. I guess I'm still thinking about it from 3D editor point of view - in Blender I usually have multiple views and keep changing settings in each one separately on the fly based on what I'm doing, e.g. I might do view1: top ortho, view2: side ortho, view3: perspective. This way I can do very precise aligning and selection and have very high space awareness. Then I may close the first two views and keep only the perspective in bigger resolution to preview the scene. But:

- I don't know if how I'm working with Blender is how CS is supposed to work, it's a different type of program.
- I might leave it for another PR if there is discussion to be had, for example about the exact GUI for this.

> I can't come up with a use case where you would want to change FOV settings on the fly.

Changing FOV on fly may be rarely used, but changing between ortho and perspective would be pretty frequent I imagine, so for completeness I would simply allow all (3) camera parameters that can be set in preferences to be also changed per-view on-the-fly, if we were to do that.

zinnschlag:
Not entirely convinced that this is practical, but I don't have any objections.

drummyfish:
We might still leave it for later though, I just found that I may not have much time in the following days and I'd like to ask away at forums first anyway, so let's keep it this way maybe. Should I do anything else for this PR then? What do you think about the code?

zinnschlag:
Looks okay, but I will give it a more thorough review within the next few days.

apps/opencs/view/render/cameracontroller.cpp
```diff
@@ -668,28 +668,47 @@ namespace CSVRender
...
     void OrbitCameraController::rotateVertical(double value)
     {
         osg::Vec3d eye, center, up;
         getCamera()->getViewMatrixAsLookAt(eye, center, up);
+        osg::Vec3d absoluteUp = osg::Vec3(0,0,1);
 
         osg::Vec3d forward = center - eye;
-        osg::Quat rotation = osg::Quat(value, up ^ forward);
+
+        osg::Vec3d axis = absoluteUp ^ forward;
```
Aesylwinn: This would cause issues if absouteUp and forward were to ever be equal. Have you considered this edge case?

drummyfish: True, let me fix that.

apps/opencs/view/render/cameracontroller.cpp
```diff
@@ -668,28 +668,43 @@ namespace CSVRender
...
+            up = rotation * up;
+
         getCamera()->setViewMatrixAsLookAt(mCenter + newOffset, mCenter, up);
     }
 
     void OrbitCameraController::rotateVertical(double value)
     {
         osg::Vec3d eye, center, up;
         getCamera()->getViewMatrixAsLookAt(eye, center, up);
+        osg::Vec3d absoluteUp = osg::Vec3(0,0,1);
```
zinnschlag: I am getting a warning about an unused variable here. Is that just a left-over or did you forgot something?

drummyfish: I missed it, sorry.

zinnschlag:
Gave it another test-drive. Looks okay. But there is still some room for improvements in regards to the user settings.

* Do we really need doubles for the numeric values? Looks unnecessary cluttered IMO. Plus, I don't remember ever seeing a FOV slider that allowed changes of 0.1 or even smaller.
* Should there be a tooltip for "Orthographic projection size parameter? Not sure about this one.
* The 3d Scene Input page is starting to look a bit messy. Actually had to look into the source to locate the new option. Maybe sorting options by camera type would help. Or splitting camera controls into a separate page?

btw. just noticed that there is an "Object Marker Transparency" option in "3D Scene Input" that really belongs on your new Rendering page instead. Not really part of this task, but while your at it could you move it over?

drummyfish:
I tried to address some of these issues. However, does `declareSeparator()` do anything? I can't really see any separators in the menu. Also should I make near and far clip planes configurable as well? (I already tried but changing the values somehow did nothing for some reason, don't know if I should continue the effort.)

zinnschlag:
> However, does declareSeparator() do anything?

Introduces are small vertical gap. Maybe it is a bit too subtle.

> Also should I make near and far clip planes configurable as well? (I already tried but changing the values somehow did nothing for some reason, don't know if I should continue the effort.)

Not sure. @scrawl Any input?

scrawl:
OSG autocomputes near/far by default to match the current viewport. You could turn that off and use manual values, but I don't see the point tbh.

drummyfish:
Okay, I'd leave it at that. So anything else here? :)

PR #1583, opened 2017-12-08
Don't throw exceptions from the audio decoder

kcat:
Should fix #4252 (and #4133).

### DIFF
--- a/apps/openmw/mwsound/ffmpeg_decoder.cpp
+++ b/apps/openmw/mwsound/ffmpeg_decoder.cpp
@@ -176,16 +176,16 @@ size_t FFmpeg_Decoder::readAVAudioData(void *data, size_t length)
     return dec;
 }
 
-bool FFmpeg_Decoder::open(const std::string &fname)
+void FFmpeg_Decoder::open(const std::string &fname)
 {
     close();
+    mDataStream = mResourceMgr->get(fname);
+
+    if((mFormatCtx=avformat_alloc_context()) == NULL)
+        throw std::runtime_error("Failed to allocate context");
+
     try
     {
-        mDataStream = mResourceMgr->get(fname);
-
-        if((mFormatCtx=avformat_alloc_context()) == NULL)
-            throw std::runtime_error("Failed to allocate context");
-
         mFormatCtx->pb = avio_alloc_context(NULL, 0, 0, this, readPacket, writePacket, seek);
         if(!mFormatCtx->pb || avformat_open_input(&mFormatCtx, fname.c_str(), NULL, NULL) != 0)
         {
@@ -250,16 +250,8 @@ bool FFmpeg_Decoder::open(const std::string &fname)
         mOutputChannelLayout = (*mStream)->codec->channel_layout;
         if(mOutputChannelLayout == 0)
             mOutputChannelLayout = av_get_default_channel_layout((*mStream)->codec->channels);
-
-        return true;
     }
-    catch(std::exception &e)
-    {
-        std::cerr<< "Could not open audio file: "<<e.what() <<std::endl;
-    }
-
-    if(mFormatCtx)
-    {
+    catch(...) {
         if(mStream)
             avcodec_close((*mStream)->codec);
         mStream = NULL;
@@ -273,8 +265,8 @@ bool FFmpeg_Decoder::open(const std::string &fname)
         mFormatCtx->pb = NULL;
 
         avformat_close_input(&mFormatCtx);
+        throw;
     }
-    return false;
 }
 
 void FFmpeg_Decoder::close()
@@ -316,13 +308,10 @@ std::string FFmpeg_Decoder::getName()
     return mFormatCtx->filename;
 }
 
-bool FFmpeg_Decoder::getInfo(int *samplerate, ChannelConfig *chans, SampleType *type)
+void FFmpeg_Decoder::getInfo(int *samplerate, ChannelConfig *chans, SampleType *type)
 {
     if(!mStream)
-    {
-        std::cerr<< "No audio stream info" <<std::endl;
-        return false;
-    }
+        throw std::runtime_error("No audio stream info");
 
     if(mOutputSampleFormat == AV_SAMPLE_FMT_U8)
         *type = SampleType_UInt8;
@@ -330,6 +319,11 @@ bool FFmpeg_Decoder::getInfo(int *samplerate, ChannelConfig *chans, SampleType *
         *type = SampleType_Int16;
     else if(mOutputSampleFormat == AV_SAMPLE_FMT_FLT)
         *type = SampleType_Float32;
+    else
+    {
+        mOutputSampleFormat = AV_SAMPLE_FMT_S16;
+        *type = SampleType_Int16;
+    }
 
     if(mOutputChannelLayout == AV_CH_LAYOUT_MONO)
         *chans = ChannelConfig_Mono;
@@ -378,18 +372,10 @@ bool FFmpeg_Decoder::getInfo(int *samplerate, ChannelConfig *chans, SampleType *
                           0,                              // logging level offset
                           NULL);                          // log context
         if(!mSwr)
-        {
-            std::cerr<< "Couldn't allocate SwrContext" <<std::endl;
-            return false;
-        }
+            throw std::runtime_error("Couldn't allocate SwrContext");
         if(swr_init(mSwr) < 0)
-        {
-            std::cerr<< "Couldn't initialize SwrContext" <<std::endl;
-            return false;
-        }
+            throw std::runtime_error("Couldn't initialize SwrContext");
     }
-
-    return true;
 }
 
 size_t FFmpeg_Decoder::read(char *buffer, size_t bytes)
--- a/apps/openmw/mwsound/ffmpeg_decoder.cpp
+++ b/apps/openmw/mwsound/ffmpeg_decoder.cpp
@@ -3,18 +3,13 @@
 #include <memory>
 
 #include <stdexcept>
-#include <sstream>
+#include <iostream>
 #include <algorithm>
 
 #include <components/vfs/manager.hpp>
 
 namespace MWSound
 {
-
-void FFmpeg_Decoder::fail(const std::string &msg)
-{
-    throw std::runtime_error("FFmpeg exception: "+msg);
-}
 
 int FFmpeg_Decoder::readPacket(void *user_data, uint8_t *buf, int buf_size)
 {
@@ -33,7 +28,8 @@ int FFmpeg_Decoder::readPacket(void *user_data, uint8_t *buf, int buf_size)
 
 int FFmpeg_Decoder::writePacket(void *, uint8_t *, int)
 {
-    throw std::runtime_error("can't write to read-only stream");
+    std::cerr<< "can't write to read-only stream" <<std::endl;
+    return -1;
 }
 
 int64_t FFmpeg_Decoder::seek(void *user_data, int64_t offset, int whence)
@@ -180,40 +176,40 @@ size_t FFmpeg_Decoder::readAVAudioData(void *data, size_t length)
     return dec;
 }
 
-void FFmpeg_Decoder::open(const std::string &fname)
+bool FFmpeg_Decoder::open(const std::string &fname)
 {
     close();
-    mDataStream = mResourceMgr->get(fname);
-
-    if((mFormatCtx=avformat_alloc_context()) == NULL)
-        fail("Failed to allocate context");
-
-    mFormatCtx->pb = avio_alloc_context(NULL, 0, 0, this, readPacket, writePacket, seek);
-    if(!mFormatCtx->pb || avformat_open_input(&mFormatCtx, fname.c_str(), NULL, NULL) != 0)
-    {
-        // "Note that a user-supplied AVFormatContext will be freed on failure".
-        if (mFormatCtx)
-        {
-            if (mFormatCtx->pb != NULL)
-            {
-                if (mFormatCtx->pb->buffer != NULL)
-                {
-                    av_free(mFormatCtx->pb->buffer);
-                    mFormatCtx->pb->buffer = NULL;
-                }
-                av_free(mFormatCtx->pb);
-                mFormatCtx->pb = NULL;
-            }
-            avformat_free_context(mFormatCtx);
-        }
-        mFormatCtx = NULL;
-        fail("Failed to allocate input stream");
-    }
-
     try
     {
+        mDataStream = mResourceMgr->get(fname);
+
+        if((mFormatCtx=avformat_alloc_context()) == NULL)
+            throw std::runtime_error("Failed to allocate context");
+
+        mFormatCtx->pb = avio_alloc_context(NULL, 0, 0, this, readPacket, writePacket, seek);
+        if(!mFormatCtx->pb || avformat_open_input(&mFormatCtx, fname.c_str(), NULL, NULL) != 0)
+        {
+            // "Note that a user-supplied AVFormatContext will be freed on failure".
+            if (mFormatCtx)
+            {
+                if (mFormatCtx->pb != NULL)
+                {
+                    if (mFormatCtx->pb->buffer != NULL)
+                    {
+                        av_free(mFormatCtx->pb->buffer);
+                        mFormatCtx->pb->buffer = NULL;
+                    }
+                    av_free(mFormatCtx->pb);
+                    mFormatCtx->pb = NULL;
+                }
+                avformat_free_context(mFormatCtx);
+            }
+            mFormatCtx = NULL;
+            throw std::runtime_error("Failed to allocate input stream");
+        }
+
         if(avformat_find_stream_info(mFormatCtx, NULL) < 0)
-            fail("Failed to find stream info in "+fname);
+            throw std::runtime_error("Failed to find stream info in "+fname);
 
         for(size_t j = 0;j < mFormatCtx->nb_streams;j++)
         {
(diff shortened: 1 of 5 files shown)


scrawl:
* Might it be simpler to keep the decoders as is, but catch the exception in SoundManager (playSound3D, etc.)?
* As far as I remember from previous OpenMW versions, there are some cases where it might not be a good idea to keep logging errors. For example, with a script like `if (!getSoundPlaying(sound)) playSound(sound)` we would get a new error every frame, right?

kcat:
> Might it be simpler to keep the decoders as is, but catch the exception in SoundManager (playSound3D, etc.)?

Exceptions aren't free, and I also don't think they should be used for general error handling. It can be difficult to control the code flow in larger blocks if exceptions can fire from anywhere within several calls deep (as we've seen with unexpected exceptions that abort the frame processing at random points, causing the game to appear locked up and unresponsive, or make random things not work). You'll end up with levels upon levels of ```try { }``` blocks as they all attempt to either clean up after a potential exception fires before rethrowing it, or try to prevent the exception from propagating further up.

IMO, an exception is basically the code's way of panicking with "Oh god, everything's broken, I don't know what to do!", where an exception handler can hopefully step in and say "No, no, it's okay. I got this." and continue on as best it can (or there isn't a handler, the system agrees everything's broken, and terminates the process). If the code could instead say "I'm sorry, I can't do that." it tends to work better. Compiler extensions, C++17, and static analysis would allow us to find places where a function could return an error that's not checked, but not properly handling potential exceptions is a lot harder to detect.

> As far as I remember from previous OpenMW versions, there are some cases where it might not be a good idea to keep logging errors. For example, with a script like if (!getSoundPlaying(sound)) playSound(sound) we would get a new error every frame, right?

This could be fixed by substituting a dummy silent sound if the desired one fails to load. This would occur in the output backend, where it can synthesize a small silent buffer upon detecting a failure from the decoder. That way you don't even keep triggering an error, let alone keep logging it (and as far as the sound manager is concerned, everything's working smoothly). Having an error logged at least once would still be useful, though.

kcat:
Updated to load silence if a decoder fails to provide usable audio, which should avoid excessive log spam when continually trying to play a sound that can't be loaded.

zinnschlag:
I am with scrawl on this one. Exceptions aren't free, but they are generally cheap, often even cheaper than other error handling methods.

>  I also don't think they should be used for general error handling.

I am afraid you are pretty alone with this position, considering that that is how we have used exceptions in OpenMW right from the beginning of the project.

kcat:
> Exceptions aren't free, but they are generally cheap, often even cheaper than other error handling methods.

But as we regularly see, they tend to be harder to use correctly, especially given that OpenMW has an exception handler in the frame processor providing a false safety net. It's a lot harder to tell where exception handling is missing compared to where return values aren't checked, and the source of an exception is harder to identify in the code versus the source of an error return value.

> I am afraid you are pretty alone with this position, considering that that is how we have used exceptions in OpenMW right from the beginning of the project.

If I could make a suggestion then, please remove the exception handler in ```OMW::Engine::frame```. I have never encountered an instance where I was glad that it caught something, but plenty of times where I wish it hadn't. If the process is allowed to terminate from an unhandled exception we get a backtrace that gives the exact file and line number for the offending ```throw``` statement, rather than having to look through logs for a 'frame error' message and then grepping the source trying to match the message (hoping it's not a dynamically generated string, or duplicated in multiple places). It would also mean the bug is better exposed, instead of hiding as a message in a log while leaving behind unknown effects on the game state.

Also, it would be a good idea to make customized exception types (simply derive from related standard exception types, and throw/catch those). That way exception handlers can appropriately respond to specific problems, rather than assuming all exceptions are equal.

scrawl:
I'm with Zini that exceptions are usually a good thing. Actually, in this instance, the script exiting because the sound was not found might even be considered intended behavior, because if the script depended on the sound playing for a given amount of time or something like that then it will certainly break. But I guess that's not how vanilla works, so we'll want to just ignore the missing sound.

> If I could make a suggestion then, please remove the exception handler in OMW::Engine::frame. I have never encountered an instance where I was glad that it caught something, but plenty of times where I wish it hadn't. If the process is allowed to terminate from an unhandled exception we get a backtrace that gives the exact file and line number for the offending throw statement, rather than having to look through logs for a 'frame error' message and then grepping the source trying to match the message (hoping it's not a dynamically generated string, or duplicated in multiple places). It would also mean the bug is better exposed, instead of hiding as a message in a log while leaving behind unknown effects on the game state.

I don't think it would be a good idea to make a runaway exception just exit the game, mostly because you will lose your progress in the game. Exceptions caught in frame() aren't nice (sometimes they just freeze the game if the same one gets thrown over and over), but at least you should be able to hit F5 to save your game still.

But yes, if an exception gets caught in frame() it should probably have been caught somewhere below instead, which should be considered a bug. Maybe we can even change the exception message to 'please report a bug' or something.

> Also, it would be a good idea to make customized exception types (simply derive from related standard exception types, and throw/catch those). That way exception handlers can appropriately respond to specific problems, rather than assuming all exceptions are equal.

I've never realized a need for that, but if you have an idea where this can be used then go ahead.

scrawl:
> I don't think it would be a good idea to make a runaway exception just exit the game, mostly because you will lose your progress in the game.

Hm, maybe I'd be ok with it if we can make the game attempt to create an autosave before exiting.

I suppose that behavior might result in nicer bugreports, e.g. the report would be 'hey the game crashed and told me XYZ' rather than a 'ermagerd everything is frozen my character cant move!' where we have to follow up and tell them to post the log file before we can see anything.

kcat:
> Actually, in this instance, the script exiting because the sound was not found might even be considered intended behavior, because if the script depended on the sound playing for a given amount of time or something like that then it will certainly break.

A sound failing to play isn't usually cause for a script to abort, because it's otherwise possible to play the game without sound enabled. If a script can't play a sound, then it *may* be a problem depending on the script (most won't have a real problem, it'll just keep trying to play the sound each frame after seeing it's not playing), but if the script is killed it certainly will be a problem since the script is expected to run and do something.

> I don't think it would be a good idea to make a runaway exception just exit the game, mostly because you will lose your progress in the game. Exceptions caught in frame() aren't nice (sometimes they just freeze the game if the same one gets thrown over and over), but at least you should be able to hit F5 to save your game still.

Is it really a good idea to save the game with it in such a state? Since only part of the frame is being processed, the state isn't consistent so it can leave non-obvious problems. Some objects may be updated and others not, some objects may have X state updated and Y state not, when X and Y are assumed to update together each frame, etc. And if the resulting save turns out bad, it wouldn't be a good idea to overwrite the quicksave or autosave since it could be their latest (or only) save.

> I've never realized a need for that, but if you have an idea where this can be used then go ahead.

Well, my idea was that identifying specific problems would help in correctly dealing with it. Something along the lines of ```ScriptExecutionException```, which would get thrown if there was a problem processing the script (not because a script function had a problem, but the script processing itself, like a missing reference to call a function with for example), and the script manager would look for that before killing the script. Other exceptions would be left unhandled to cause an abort that we can track down with a crash log/stacktrace.

> I suppose that behavior might result in nicer bugreports, e.g. the report would be 'hey the game crashed and told me XYZ' rather than a 'ermagerd everything is frozen my character cant move!' where we have to follow up and tell them to post the log file before we can see anything.

Yeah, that's part of the reason behind the concept of "fail early and fail loud". Don't try to hide an error if it can fester and create other problems later on, as it'll just obfuscate the real issue and cause unnecessary work for both developers and the reporter. If it's a real problem that can cause continued misbehavior, then crash/abort and get a proper error report out of it.

That's also my reasoning behind relegating exceptions to serious "Oh Crap" errors, or obvious "I Said Don't Do That" errors. The fewer ```try```/```catch``` blocks you have, the easier it is to correctly handle exceptions (including being left unhandled if it should abort).

kcat:
Restored exception throwing to the decoder. The exception is still handled in the output backend to substitute a silent buffer for files that don't provide usable audio, though.

zinnschlag:
> Hm, maybe I'd be ok with it if we can make the game attempt to create an autosave before exiting.

> I suppose that behavior might result in nicer bugreports, e.g. the report would be 'hey the game crashed and told me XYZ' rather than a 'ermagerd everything is frozen my character cant move!' where we have to follow up and tell them to post the log file before we can see anything.

Absolutely do not agree with that. OpenMW is in wide use of people who not just test it, but play it. For a game a shutdown/crash is the worst possible outcome, even if the game state isn't lost. It should be avoided by any means possible.

kcat:
> For a game a shutdown/crash is the worst possible outcome

I'd say state corruption and lockups are a lot worse. With the former, you leave a problem that can end up getting corrupted data into the save game that can undo hours or even days of progress once it manifests. With the latter, you're just making the user wait until they realize nothing's happening and then have them do a forced shutdown (which may not be easy depending on the system and nature of the lockup).

kcat:
> If we auto save on crash (and we would need to do that in this situation) we would get the corruption anyway.

I don't see why we would need to save in that situation, but I'm also not a big fan of a save-on-crash idea. I suppose, though, a save could be useful for developers working on certain crash-related issues to get a closer look at the game state beyond what a backtrace provides. Although I would not overwrite the autosave or quicksave (a separate "crashsave" perhaps?) and I'd leave it with the caveat to the user that it could be corrupt. In fact, I bet we could add a flag to such crashsaves that make OpenMW display a warning when trying to load it:

"This savegame was created because of a crash and may contain corrupt save data. Continue at your own risk.
Continue / Cancel"

For developers that are expecting to load these often, there could be a config option to suppress the warning (or suppress the warning when loading from the command line?). The save could probably also be colored differently in the UI, giving it a red background or tint to indicate potential danger.

> Your Skyrim story doesn't surprise me hugely. We all know that Bethesda's software is fragile as hell.

Sure, but the principle remains. All software has bugs, and in this instance there was a bug that wasn't noticed until well after the point of no return because the engine tried to be robust. I wouldn't be surprised either if the main reason Bethesda's engine is so buggy is because over the years they kept trying to hide errors and keep the game running despite them, rather than facing them and tackling the root cause. QA and unit tests can only get you so far, especially on PC, and hiding what you don't expect to happen from users is just asking for trouble.

> For OpenMW in particular I don't see that as a problem. Our handling of the game state seems rather robust. And even if we had such an event, there would be options to salvage the situation because we have access to the source and know the file format.

Ironically, the most error-prone code tends to be error handlers since they receive the least amount of testing (you don't want an error to fire so you try to avoid hitting that code path, and thus have less confidence it'll work correctly when it's inevitably hit). IMO, it's not good to assume anyone's error handling is all-around sturdy and robust, so the more help we can get with error handling, the better.

With return values, we have tools to help make sure there's error handling in place since we can specify attributes on functions to require callers to check return values (or else get a compile-time warning), while with exceptions there's no way to pragmatically determine the intended code path (who's responsible for a given function call throwing an exception? was the direct caller supposed to handle it, or is someone further up the callstack missing an exception handler? or was it handled correctly and that was intended behavior?). Ensuring there's an error handler correctly in place is the first step to testing that error handler.

There are always exceptions (pun not intended), so it's not a hard and fast rule that everything should have a return value for errors. Sometimes it doesn't matter if a call wasn't successful, and sometimes an unsuccessful call would be so severe that there's no reasonable way to handle it. Sometimes it only matters if a string of calls was successful, rather than specific ones giving an error. The best case is that it's impossible for a function to fail except when everything's broken and there's nothing worth salvaging, but that's not always possible. It's all about using the correct tool for the job while ensuring the job is being done correctly.

zinnschlag:
This is a situation where both approaches have advantages and disadvantages and it comes down to a call of judgement, which makes it at least partially subjective.

Regarding shutdown on bug or not:
* Shutdown: Lower (but not zero) risk of data corruption; interruption of game play on encountering a bug
* No shutdown: Slight higher risk of corruption; continued game play on encountering a bug

Since the corruption case is very rare and recoverable bugs are common, I see this clearly as in favour of no shutdown.

Regarding exceptions vs error return codes: Both have advantages and disadvantages regarding clarity of code and performance. Deciding in an entirely objective and general way which outweighs the other is impossible. But from my personal experience the picture is very clear.
I have written code entirely without exceptions (because the bloody compiler I had to use didn't support them properly) and I have written code which relies heavily on exceptions. And what I have experienced is this:
Code that uses exceptions heavily (including handling of all kinds of errors) is overwhelmingly more stable, more reliable and more readable than code that uses error codes. My conclusion is that exceptions should be used generously and error codes should be reserved for special situations.

I don't think we will come to agreement on this issue.

kcat:
> Regarding shutdown on bug or not:
> * Shutdown: Lower (but not zero) risk of data corruption; interruption of game play on encountering a bug
> * No shutdown: Slight higher risk of corruption; continued game play on encountering a bug
> Since the corruption case is very rare and recoverable bugs are common, I see this clearly as in favour of no shutdown.

There is a distinction to make with the severity of a bug. An audio file failing to open/decode isn't that severe and it should be possible to continue on without issue, but an exception getting back to the main frame processing function is severe with a higher than normal chance of causing corruption because it breaks the intended/expected flow processing loop.

If something like the former causes the latter that means some error handling is missing and there's something causing an error, which needs to be looked at and rectified rather than hidden on the hopes that maybe the next frame can continue on "correctly" (without knowing if it can or not, it's basically the equivalent of putting your head in the sand and hoping the problem goes away on its own with no lingering effects).

> Code that uses exceptions heavily (including handling of all kinds of errors) is overwhelmingly more stable, more reliable and more readable than code that uses error codes. My conclusion is that exceptions should be used generously and error codes should be reserved for special situations.

Exceptions work a lot better when you're heavily RAII driven, and the success of individual functions is less important than the success of a long block (and by extension, you can deal with the fact that a block of code stopped partway through before getting to the end).

As far as I understand though, exceptions are a big no-no in commercial software development. Also, throwing across DLL boundaries is bad because of the way types work in C++ and symbol visibility (not currently much of a concern for us, but if we ever go modular it's something to think about).

scrawl:
> For OpenMW in particular I don't see that as a problem. Our handling of the game state seems rather robust.

Agreed, in all the years of testing/bug reports I don't recall many instances where a savegame got "corrupted" beyond the point of repair. You always have the console to fix things.

> Although I would not overwrite the autosave or quicksave (a separate "crashsave" perhaps?)

Yes, a separate crashsave is what I would suggest **if** we do this at all.

I feel like going from logging errors to just crashing the game would be a too big step though. As a middle ground how about we open an in-game messagebox saying 'an error occured: <exception text>' with choices 'Ignore', 'Ignore All' (IIRC, the original engine has such messages, though they usually crash the game shortly afterwards anyway).

kcat:
> Agreed, in all the years of testing/bug reports I don't recall many instances where a savegame got "corrupted" beyond the point of repair. You always have the console to fix things.

Generally because we're pretty good about not letting exceptions get back to the main loop, and most of the time when it does, it happens continuously creating a soft lock situation (where the only option is to shutdown anyway). It also tends to be something that requires immediate attention, and with OpenMW's player base being relatively small currently, that means it's less likely for people to be seriously stung before a fix is available.

> I feel like going from logging errors to just crashing the game would be a too big step though. As a middle ground how about we open an in-game messagebox saying 'an error occured: ' with choices 'Ignore', 'Ignore All' (IIRC, the original engine has such messages, though they usually crash the game shortly afterwards anyway).

I'd have the choices as "Ignore" and "Quit" (or "Abort"). So you have something like:
```
try {
    ... normal frame update stuff here...
}
catch(std::exception &e) {
    handle_uncaught_exception(e.what());
}
```
where ```handle_uncaught_exception``` would contain it's own loop for handling the error message box. Clicking "Ignore" would make the function return and continue on like nothing's wrong, while clicking "Quit"/"Abort" would rethrow the exception and generate a crashlog from an uncaught exception (and if the function itself somehow manages to throw an exception, it won't get caught in an infinite loop). An "Ignore All" doesn't make much sense to me because if it happens again, it's likely to be a soft lock and you've just taken away the ability for the user to easily quit with a crashlog.

This would also be a good use-case for custom exception types, since they can contain their own methods and data storage (e.g. a ```ScriptExecutionError``` could contain the name of the script, what object that script instance is connected to, if any, as well as the message about what happened).

scrawl:
I like that idea, the ability to get a stack trace would be nice too.

But I still think we might want to have an 'Ignore all' button because there are some errors that won't be persistent. Like, an error that only happens when you point at a particular object and goes away when you look elsewhere.

> (e.g. a ScriptExecutionError could contain the name of the script, what object that script instance is connected to, if any, as well as the message about what happened).

Should script execution errors really pop up a message though? I thought the 'intrusive' error boxes would be for engine-specific errors, things that are (probably) the fault of the engine. For mod-related errors, showing every single error could be very annoying.

scrawl:
> For mod-related errors, showing every single error could be very annoying.

... as I'm sure there are many people who play with broken-ass mod setups that work 'good enough' for them, but they don't want to spend the time to dive in and make them error-free, more so when they don't have the permission to upload those 'fixed' mods for others to use.

And we have the verifier function in the editor, which is more convenient anyway (and unlike the game, detects all errors, not just the ones in currently loaded cells).

kcat:
> But I still think we might want to have an 'Ignore all' button because there are some errors that won't be persistent. Like, an error that only happens when you point at a particular object and goes away when you look elsewhere.

Presuming the exception happens after input has been polled and movement applied, or else once you look at something you're stuck looking at it forever.

> Should script execution errors really pop up a message though? I thought the 'intrusive' error boxes would be for engine-specific errors, things that are (probably) the fault of the engine.

Maybe. I think it should be, since a script failing to execute means a script won't do what it was supposed to, leaving the game in an unintended state. Although I wouldn't be surprised if some mods use hacks that expect script execution to fail in some cases (in Oblivion, modders would sometimes use a faulty script that would abort the script in vanilla but work with OBSE, and whether or not the script completes would then be used to test if OBSE was present or not).

But, script execution errors was just an example. I'm sure there are other places where passing along extra information with the exception could be useful to log, even if it's not displayed in a message box, or to do better error recovery given the nature of the error.

PR #1588, opened 2017-12-14
[Testing needed] Make StartScript behaviour consistent (bug #3823)

akortunov:
CAUTION! Since this change affects script engine basics, we should test it carefully.

This PR is aimed to fix [bug #3823](https://bugs.openmw.org/issues/3823).

In master branch, a newly started script will be executed in the current frame basically depends on where it's inserted in the map respective to the currently running script, which depends on the name of the script.

With this PR, any script started via StartScript will start execution only from the next frame.
StopScript prevents the script from running in the current frame, if it was not executed yet.
ScriptRunning checks if the script was marked to stop/run.

How it works:
I added a new variable to script mWillBeRunning, which describes, if an engine should run the script in the next frame.
StartScript set this variable to true, StopScript - to false, ScriptRunning checks its value.
Before every run of script subsystem, I set the mRunning to mWillBeRunning.
Also I update mRunning when writting it to the savegame, so we do not need to store the mWillBeRunning in the savegame.


apps/openmw/mwscript/globalscripts.cpp
```diff
@@ -125,8 +137,10 @@ namespace MWScript
         return mScripts.size();
     }
 
-    void GlobalScripts::write (ESM::ESMWriter& writer, Loading::Listener& progress) const
+    void GlobalScripts::write (ESM::ESMWriter& writer, Loading::Listener& progress)
     {
+        updateState();
```
zinnschlag: What is this update supposed to do in the write function? Looks wrong.

akortunov: > Also I update mRunning when writting it to the savegame, so we do not need to store the mWillBeRunning in the savegame.

zinnschlag: Still makes no sense to me. Why would you have to store mWillBeRunning to begin with? Its not as if you can save in the middle of script execution or even a script execution cycle.

akortunov: > Its not as if you can save in the middle of script execution or even a script execution cycle.

Then I can just update state only only once in the script execution cycle, and write() changes are indeed redundant.

Allofich:
akortunov, can you suggest what kind of testing would be helpful?

Allofich:
I found a strange issue when using this PR.

Here is a test .esp I was using. The name is "noweight.esp" but that is because I was re-using it after testing 0-weight containers.

I modified the Main script. In vanilla, the Main script calls

StartScript Startup
StartScript VampireCheck

I added MessageBox commands to these scripts and to Main (after the StartScript calls) to show their names. I also added StopScript VampireCheck to VampireCheck because it would continuously show its message box otherwise. I then start a New Game and watch the message boxes appear.

At this point, the behavior was the same between the original engine and master branch. Both showed message boxes in the order 1. Main 2. Startup 3. VampireCheck (I thought maybe what this PR fixed was this? But master branch showed the correct order)

I remembered that there was mention of the script names affecting order in OpenMW, and Main->Startup->VampireCheck is alphabetical, so this time I added AjiraScript with a MessageBox and StopScript as an additional script to call after VampireCheck. Here is where the strange behavior occurred.

I can only see 3 message boxes at a time, so I don't see Main anymore. In the original engine, I see
1. Startup 2. VampireCheck 3. AjiraScript
as I would expect.
In master I see
1. VampireCheck 2. AjiraScript 3. AjiraScript
For some reason AjiraScript's message appears twice.
But with this PR, I see AjiraScript continously. The StopScript is not being called for some reason.

Any idea about what might be happening? I also tried using "StartScript ajiraScript" in the console rather than starting a New Game. In this case the message showed continuously in all cases: the PR, master branch, and the original engine.
[noweight.zip](https://github.com/OpenMW/openmw/files/1565823/noweight.zip)

Allofich:
Sorry that it's a bit messy and hard to explain. The easiest thing to do is to load the attached .esp in the original engine, master branch and this PR and see the difference in message boxes.

Allofich:
I noticed that the "StopScript" command in ajiraScript came after
if ( CellChanged == 0 )
	Return

If I move it before that command, master branch and original engine behave the same, and the message box only shows once when called StartScript ajiraScript from the console.

However, with this PR, I then get a different order of message boxes.
1. Main 2. ajiraScript 3. StartUp 4. VampireCheck

In the original engine and master branch I get
1. Main 2. StartUp 3. VampireCheck 4. ajiraScript
as expected.

Allofich:
Here is a simpler test .esp file.

[scriptOrderTest.zip](https://github.com/OpenMW/openmw/files/1567977/scriptOrderTest.zip)

There are three new scripts: TestScript1, TestScript2, TestScript3. Each announces itself with a message box.

TestScript1 contains:

> StartScript TestScript3
StartScript TestScript2
MessageBox, "TestScript1"

The order of messages in the original engine when using "StartScript TestScript1" from the console is:
1, 3, 2
In both master branch and this PR, it is
1, 2, 3

Just in case "MessageBox" was unreliable for showing order, I also added a new global variable called TestVar. This is set by TestScript1 to 1, by TestScript2 to 2, and by TestScript3 to 3. Checking it with the "sv" command after running the scripts shows that TestVar is 2 in the original engine, and 3 in master branch and with this PR.

Allofich:
I've done some more refined testing and found a pattern in original script execution.
Here is the .esp file
[scriptOrderTest2.zip](https://github.com/OpenMW/openmw/files/1571369/scriptOrderTest2.zip)

As above, scripts are testScript1, testScript2 and testScript3. Test with "startscript testScript1". It calls testScript3 and then testScript2. All scripts add to TestVar, and return when TestVar == 4.

By resetting TestVar after each run with "set TestVar to --", you can see the order of script execution. TestVar starts at 0, but by resetting after each run with "set TestVar to --", such as "set TestVar to -1", you can add or subtract the number of script executions that run until the limit of 4 is reached. I stepped through it this way and saw that the script execution in the original engine is:

testScript1
testScript3
testScript1
testScript2
testScript1
testScript3
testScript1
testScript2
etc.

So the behavior looks like when a script calls other scripts, first it finishes itself (ignoring further StartScript calls), then it runs the called script, and when it encounters the next StartScript it again finishes itself first, then runs that script, etc.

Mingun:
I think, that `startscript` command just schedule execution of the script. Also, all scripts run in the one thread. This algorithm was tracked from your previous comment.

I do not know that `stop itself` means, but it seems that it simply finishes the current frame. Queue of outstanding scripts at the same time is not cleaned. So I decode you example as:

frame 0: testScript1 -- schedule 3, 2; queue: [3, 2]
frame 0: testScript3 -- stop frame 0; queue: [2]
frame 1: testScript1 -- schedule 3, 2; queue: [2, 3, 2]
frame 1: testScript2 -- scheduled at frame 0; queue: [3, 2]
frame 1: testScript3 -- scheduled at frame 1; stop frame 1; queue: [2]
frame 2: testScript1 -- schedule 3, 2; queue: [2, 3, 2]
...

frame 0: startScript testScript3 -- schedule 3; queue: [3]
frame 0: startScript testScript2 -- schedule 2; queue: [3, 2]
frame 0: startScript testScript3 -- schedule 3; queue: [3, 2, 3]
frame 0: execution of 3; finish frame 0; queue: [2, 3]
frame 1: startScript testScript3 -- schedule 3; queue: [2, 3, 3] or [2, 3] if engine not schedule the same script twice (may be it do this to prevent queue overflow)
frame 1: startScript testScript2 -- schedule 2; queue: [2, 3, 3, 2] or [2, 3, 2]
frame 1: startScript testScript3 -- schedule 3; queue: [2, 3, 3, 2, 3] or [2, 3, 2, 3]
frame 1: execution of 2 scheduled at frame 0; queue: [3, 3, 2, 3] or [3, 2, 3]
frame 1: execution of 3; finish frame 1; queue: [3, 2, 3] or [2, 3]
...

Allofich:
> I do not know that stop itself means

I meant the script calls "StopScript" on itself, like TestScript3 containing "StopScript TestScript3".

> queue: [2, 3, 3, 2, 3]

I don't think the scripts build up a growing queue. If they did, when "StartScript" is no longer being called (in my tests, because "testVar" has reached >=4), the built-up queue would then be processed, right? Even if a script uses StopScript on itself, it will still finish processing that frame (in my test, add 1 to testVar). If I initialize testVar as something like -10 so that a large queue should be created before 4 is reached, testVar should be added to by each script in the queue and become large, but it doesn't.

> I do not know that stop itself means, but it seems that it simply finishes the current frame

Yes, it seems like it stops scripts that frame, and I think it also removes the target of StopScript from the queue. Then my examples I gave could be decoded as follows:
Ex.1) Both TestScript2 and TestScript3 call "StopScript" on themselves every time they are run.

frame 0: testScript1 -- schedules 3, 2; queue: [3, 2] ("1" is also in the queue, but not writing it here)
frame 0: testScript3 -- removes 3 from queue, stops remaining scripts this frame; queue: [2]
frame 1: testScript1 -- schedules 3, 2; queue: [2, 3] ("2" not added as it is a duplicate)
frame 1: testScript2 -- removes 2 from queue, frame stopped, queue: [3]
frame 2: testScript1 --  schedules 3, 2; queue: [3, 2] ("3" not added as it is a duplicate)
frame 2: testScript3 -- removes 3 from queue, frame stopped; queue: [2]
frame 3: testScript1 --  schedules 3, 2; queue: [2, 3]
.. etc.

Ex.2) TestScript3 calls "StopScript" on itself every time it is run. TestScript2 does not.

frame 0: testScript1 -- schedules 3, 2; queue: [3, 2]
frame 0: testScript3 -- removes 3 from queue, frame stopped, queue: [2]
frame 1: testScript1 -- schedules 3, 2; queue: [2, 3] ("2" not added as it is a duplicate)
frame 1: testScript2 -- no removal or scheduling, queue: [2, 3]
frame 1: testScript3 -- removes 3 from queue, frame stopped, queue: [2]
frame 2: testScript1 -- schedules 3, 2; queue: [2, 3] ("2" not added as it is a duplicate)
frame 3: testScript2 -- no removal or scheduling, queue: [2, 3]
frame 3: testScript3 -- removes 3 from queue, frame stopped, queue: [2]
.. etc.

I did another somewhat complex test, and the above assumptions held.

Ex.3) TestScript1 calls TestScript2, TestScript2 calls TestScript3 and stops self, TestScript3 stops self. Each add 1 to TestVar. TestScrip1 stops itself if TestVar >= 4.

TestVar starts at 2.
frame 0: testScript1, TestVar == 3, Schedules 2, Queue [1, 2] (I'm showing "1" in the queue this time)
frame 0: testScript2, TestVar == 4, Schedules 3, Removes 2, frame stopped, Queue [1, 3]
frame 1: testScript1, TestVar == 5, Schedules 2, Removes 1, frame stopped. Queue [3, 2]
frame 2: testScript3, TestVar == 6, Removes 3, frame stopped. Queue [2]
frame 3: testScript2, TestVar == 7, Schedules 3, Removes 2, frame stopped. Queue [3]
frame 4: testScript3, TestVar == 8, Removes 3, frame stopped. Queue [3]

The frame count is a bit of a guess, but the script behavior seems to follow the above rules.

Allofich:
Anyway, since this PR is about StartScript, perhaps I got a bit off track, since my test examples involve StopScript. But if we're only looking at StartScript, it seems like the correct behavior is to push the target script onto the back of the queue of scripts to run, unless the script already exists in the queue.

I just tested "ScriptRunning" a bit, and it seems to return 1 if the script is in the queue, and 0 if not. So if called immediately after a "StartScript Test" statement, "ScriptRunning Test" will return 1, and immediately after a "StopScript Test" statement, it will return 0.

Allofich:
From the bug report comments:

> How should it behave with StopScript? Should a script still potentially run on the frame it was stopped, even if it hasn't run before the call? Or can it be assumed the stopped script will not run after the call?

> Right, good point. We need to test how StopScript is supposed to behave. (I'd assume it stops right away)

Was this referring to immediately calling StopScript on a script that was just the target of StartScript? Like

StartScript testScript
StopScript testScript

If so, "testScript" is not run at all in this case.

Allofich:
Getting back to the root of the problem, this PR's solution of waiting until the next frame for execution seems to be incorrect.

Test case using TestScript1 and TestScript2:

TestScript1:

> StartScript  TestScript2
> 
>set TestVar to (TestVar + 1)
>MessageBox, "TestScript1 %G", TestVar
> 
>if (TestVar >= 3)
>	StopScript TestScript1
>	StopScript TestScript2
>endif

TestScript2

> set TestVar to (TestVar + 1)
> MessageBox, "TestScript2 %G", TestVar
> StopScript TestScript2


Original Morrowind output for "StartScript Testscript1":

TestScript1, 1
TestScript2, 2
TestScript1, 3

This PR's output:

--
TestScript1, 2
TestScript2, 3
TestScript1, 4

Master branch gives the same output as the original engine.

akortunov:
> Maybe when StartScript is used execution does indeed wait until the next frame, but the position in the script queue is retained from the previous frame.

I did not touch the script queue ordering. Looks like Morrowind uses a vector to store global scripts instead of map, so script running order independent from script names.

> Master branch gives the same output as the original engine.

I wonder if results will be the same, if scripts will have opposite names (scripts in OpenMW are ordered by name).

Allofich:
After switching names and running "StartScript TestScript2" from the console, original Morrowind, master branch and PR all return

TestScript2 1
TestScript1 2
TestScript2 3

Contents of scripts:

TestScript1
> Begin TestScript1
> 
> set TestVar to (TestVar + 1)
> MessageBox, "TestScript1 %G", TestVar
> 
> StopScript TestScript1
> End

TestScript2
> Begin TestScript2
> 
> StartScript TestScript1
> 
> set TestVar to (TestVar + 1)
> MessageBox, "TestScript2 %G", TestVar
> 
> if (TestVar >= 3)
> 	StopScript TestScript2
> 	StopScript TestScript1
> endif
> End

Allofich:
@akortunov (or anyone who cares to comment) what are your thoughts? I think a different approach is needed, because this PR improves some cases but could (theoretically) cause regressions in other cases, as the testing above showed.

If we change the global scripts to be sorted according to the order scripts were added, and when processing "StartScript" abort script execution until the next frame, while saving the index of the currently executing script in the list, it seems like it should give the correct results. Would changing the global scripts container cause problems with loading saved games?

Also I didn't test whether StartScript would have an effect on the execution of local scripts in a frame, do we need that?

akortunov:
> I think a different approach is needed

The different approach is to store global scripts as vector instead of map. This approach will slow a search by name a bit (e.g. via ScriptRunning console command).

But I am not sure how to handle StopScript in this case. Should we completely remove the script instance, or just store it somewhere else?

Allofich:
> But I am not sure how to handle StopScript in this case. Should we completely remove the script instance, or just store it somewhere else?

Would creating a script that, for example, iterates a variable every time it runs, StopScripts at some value, then starting it again and checking the value of the variable (whether it had reset to its initial value or not) be enough to answer that question?

akortunov:
> Would creating a script that, for example, iterates a variable every time it runs

Yes, it would be nice.

Allofich:
Ok, I'll try to test it.

akortunov:
So looks like we need a global script storage (mScripts map, I suppose) and running script queue (std::vector, I suppose).
**StartScript** should search for sctipt in mScripts or create a new GlobalScriptDesc instance, add it to end of the queue and break the global scripts execution loop. 
**StopScript** should remove the GlobalScriptDesc instance from queue. I am not sure if it should break the global scripts execution loop.
**ScriptRunning** should search for script in the queue.
We do not need the isRunning flag in this case.

Allofich:
I did a little bit of testing. 

> StartScript should ... break the global scripts execution loop. 

The script that calls it should first finish executing for that frame (I tested with Animated Morrowind, putting a "if cellchanged == 1" in the calling script, below the StartScript call, and it fired. That's a local script, not a global script as I have been testing, but hopefully it is the same) The next frame, execution will continue from the next script in the queue.

> StopScript ...I am not sure if it should break the global scripts execution loop.

I'm not sure, I can't think of a way right now to test it. The only particular behavior of StopScript that I have noticed, is that if a script stops itself, it seems that after that script finishes its current execution, scripts are restarted from the beginning of the queue.

Example:
TestScript1, TestScript2, TestScript3 all iterate TestVar by 1 every execution. Order of execution is TestScript1, TestScript2, TestScript3. TestScript2 will stop itself if TestVar == 5.

Result of TestVar is:
TestScript1, 1
TestScript2, 2
TestScript3, 3
TestScript1, 4
TestScript2, 5
TestScript1, 6
TestScript3, 7
..etc.

Allofich:
> I think it will then go to the next frame, where execution will continue from the next script in the queue. It's possible, though, that all the scripts in the queue first execute for the initial frame, and only when the newly added script is reached in the queue, that the frame is advanced. 

In an attempt to test this, I made TestScript1, which starts TestScript2, so we have them executing in order TestScript1, TestScript2.

When cellchanged == 1, they will both show a messagebox, but TestScript1 will also StartScript TestScript3, which also shows a messagebox if cellchanged == 1.

I expected that TestScript3 would not show its messagebox. I thought this is why the script in the bug report does not change the bard's music to explore music. But the result was that they each showed their messagebox, in the order 1, 2, 3. Any ideas?

Allofich:
Maybe it's due to the order of global scripts and local scripts. Do we know the order they are run in original Morrowind?

In my test in the comment above, all were global scripts, but in the bug report, the script using "StartScript" is a local script ("local script" means a script on an object like an NPC, right?). If local scripts are executed after global scripts, then maybe when the local bard script adds the global script with StartScript, the global scripts are already finished for that frame, so the added script is not run until the next frame, and cellchange == 0. There might not be any "breaking script execution," then.

akortunov:
> Do we know the order they are run in original Morrowind?

Run order according to russian Morrowind community:
1. Local scripts, alphabetically ordered by target ID
2. Global scripts, ordered by their index in the queue array.

Allofich:
So, the opposite of what I suggested. I'll try to test and confirm that when I can, hopefully we can figure this out. If anyone wants to step in with their own test results, reverse-engineering, ideas, whatever, please do.

Allofich:
> Run order according to russian Morrowind community:

> 1. Local scripts, alphabetically ordered by target ID
> 2. Global scripts, ordered by their index in the queue array.

My testing shows that the local script only executes first if it uses StartScript to start the global script, and the global script isn't already running. If a global script is already running when the cell is entered, the global script runs first.

_Except_ that local scripts run first once regardless, with `cellChanged` as 0, when the cell is first entered by the Player.

Keep the above in mind, and look at the results below of testing script order in "Animated Morrowind". I'll try to make it as clear as possible, but it is kind of complicated.

Assumption: Player has never entered the relevant cell `Mournhold, The Winged Guar` before, and the conditions for enabling the bard are met when the player enters the cell for the first time. The two relevant scripts are `AM_Bard2_Script` and `AM_StopMusic_Script`.

1. Neither script is running, player enters  `Mournhold, The Winged Guar` for the first time.
2. `AM_Bard2_Script` (attached to NPC `AM_Bard2`) runs, `cellChanged` is 0. (Special case when cells are first entered. Local scripts are run once with `cellChanged` at 0.) `AM_Bard2_Script` tries to start streaming bard music.
3. `AM_StopMusic_Script`, called by `AM_Bard2_Script`, runs, `cellChanged` is 1. It tries to start explore music.
4. `AM_Bard2_Script` runs again, `cellChanged` is 1. (This is the "Normal" run of the script that always happens on entering cell, not just first time) `AM_Bard2_Script` contains code to reset its `PlayOnce` variable if `cellChanged` is 1. It again starts streaming bard music.
5. `AM_StopMusic_Script` runs, `cellChanged` is 0, so it does not stop the music.
...
(If cell exited)
6. `AM_StopMusic_Script` runs, `cellChanged` is 1, music changed to explore music.


Note that `AM_StopMusic_Script` calls `StopScript AM_StopMusic_Script` when `cellChanged` is 1. When a script has been stopped with `StopScript` then when it is started again by `StartScript`, as when the player re-enters  `Mournhold, The Winged Guar`, `cellChanged` is 0. Perhaps there it is "pushed" to the next frame, like we have discussed above. Because it has been stopped and it is restarted by `AM_Bard2_Script`, script execution order for the player's second visit onward to `Mournhold, The Winged Guar` is:

1. `AM_Bard2_Script`, `cellChanged` is 1
2. `AM_StopMusic_Script`, `cellChanged` is 0
3. `AM_Bard2_Script`, `cellChanged` is 0
...

However, if `AM_StopMusic_Script` did _not_ call `StopScript AM_StopMusic_Script` (and if you do not call it yourself from the console) the execution order is

1. `AM_StopMusic_Script`, `cellChanged` is 1 
2. `AM_Bard2_Script`, `cellChanged` is 1
2. `AM_StopMusic_Script`, `cellChanged` is 0
3. `AM_Bard2_Script`, `cellChanged` is 0

As mentioned above, though, `AM_Bard2_Script` contains code to reset its `PlayOnce` variable if `cellChanged` is 1, so it will still work in this case.

akortunov:
Closing this PR since we do not really understand how it should work.

PR #1590, opened 2017-12-17
Do not allow to sell items from containers with zero capacity

akortunov:
Fixes [regression #4268](https://bugs.openmw.org/issues/4268).

Actually, vanilla game does not check the "organic" flag, instead it checks if the container capacity is > 0.
Most of organic containers in vanilla game have zero capacity, so it misled me. Sorry.


Allofich:
> Most of organic containers in vanilla game have zero capacity, so it misled me. Sorry.

Don't worry :) Other people on the bug tracker were also talking about organic containers, so you were misled.

PR #1591, opened 2017-12-18
Updating controller mappings to latest version

Thunderforge:
[Fixes #3708](https://bugs.openmw.org/issues/3708). I can now use my Xbox One wireless controller on a Mac. Note that this version requires SDL 2.0.6+, which the OpenMW Mac Deps [are using](https://github.com/OpenMW/openmw-deps-mac/commit/c40905f3c4afe05a9a4bbaa732e21e081f446fda). This will likely fail if OpenMW is compiled with earlier versions.


scrawl:
Quote from SDL_GameControllerDB repo:

> gamecontrollerdb.txt
> 
> For games or engines using the SDL >= 2.0.6 format. This is the most recent version.
gamecontrollerdb_205.txt
> 
> For games or engines using the SDL >= 2.0.5 format. There are no range or inversion modifiers (+,-,~).
gamecontrollerdb_204.txt
> 
> For games or engines using the SDL 2.0.4 format. GUIDs are different depending on the platform. Note that SDL > 2.0.4 can still read this format.

What an awful mess. If I pull this PR, controllers will stop working for anyone using 2.0.5 or 2.0.4, (including [Ubuntu users](https://packages.ubuntu.com/xenial/libsdl2-dev)), right?

Thunderforge:
> What an awful mess. If I pull this PR, controllers will stop working for anyone using 2.0.5 or 2.0.4, (including [Ubuntu users](https://packages.ubuntu.com/xenial/libsdl2-dev)), right?

That's correct. With our current build system, it's going to be broken for somebody. Right now, 2.0.4 is broken for Mac users, but works for Ubuntu users. If we upgrade to 2.0.6, it will be fixed for Mac users, but will break for Ubuntu users.

I'm surprised that Ubuntu only supports up to SDL 2.0.4. That was released [nearly two years ago](https://www.phoronix.com/scan.php?page=news_item&px=SDL-2.0.4-Released), and SDL 2.0.5 was released [a year and two months ago](https://discourse.libsdl.org/t/sdl-2-0-5-released/42). SL 2.0.6 was released [in September](https://discourse.libsdl.org/t/sdl-2-0-6-released/23109) of this year.

I do see [2.0.6 with the "artful" label](https://packages.ubuntu.com/artful/libsdl2-dev) and [2.0.7 with the "bionic" label](https://packages.ubuntu.com/bionic/libsdl2-dev). I'm not familiar with Ubuntu package terminology, but I'm guessing that means they are unsupported? If they were magically supported, would that address your concerns?

Thunderforge:
Also, it doesn't look like controller fixes were backported. The [2.0.6 version](https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt) has 381 lines, while the [2.0.4 version](https://github.com/gabomdq/SDL_GameControllerDB/blob/master/older_releases/gamecontrollerdb_204.txt) has 244 lines. So my Xbox One Bluetooth controller won't work even if we use that version.

Capostrophic:
>I do see 2.0.6 with the "artful" label and 2.0.7 with the "bionic" label. I'm not familiar with Ubuntu package terminology, but I'm guessing that means they are "alpha" and "beta" equivalents?

Not really. artful corresponds to Artful Aardvark, 17.10 (that is, current "stable" release of Ubuntu from October 2017), while bionic corresponds to Bionic Beaver (18.04) scheduled to be released in April 2018, it's the system of release code names (recently restarted from the beginning of the alphabet).

However, 14.04, 16.04 (being long term support releases), and 17.04 are not EOL and the support for at least 16.04 which only has SDL 2.0.4 (and this won't be changing) can't be just dropped. Till 2021.

scrawl:
I'm not a fan of breaking something to fix another something, so I think the correct solution would be either of:
* Add multiple copies of the gamecontrollerdb (like in the upstream repo) then load the one based on the SDL version we built with.
* Try converting the file at runtime.
* Improve our input binding workflow so that controller buttons/axes can still be manually bound if the controller is not supported by SDL_GameController. This would be nice to have in any case.

Thunderforge:
> Add multiple copies of the gamecontrollerdb (like in the upstream repo) then load the one based on the SDL version we built with.

This seems like the right solution in the interim; while allowing manual configuration for unsupported controllers would be nice, I don't think I'm ready to tackle that yet.

I'm thinking that implementing that would be a matter of calling [SDL_GetVersion](https://wiki.libsdl.org/SDL_GetVersion) and then changing [this line](https://github.com/OpenMW/openmw/blob/36f4f0ef85ffe74ab27fabe68f02f7c569cc1559/apps/openmw/engine.cpp#L480) and the one below it to load different files based on the result.

The only real downside of this approach is that it means we might get bug reports where certain controllers work for some users, but not others (e.g. works for Bionic Beaver, but fails for Xenial Xerus), since the controller mappings aren't being backported. But if we're okay with that, then this will probably work.

scrawl:
Come to think of it, we don't need to choose at <i>runtime</i>, we can check the SDL_VERSION in cmake and then copy the correct file to the build folder.

scrawl:
... but that will break if someone upgrades SDL without recompiling OpenMW. Clearly I'm stupid.

kcat:
> ... but that will break if someone upgrades SDL without recompiling OpenMW.

Newer versions of SDL will still read the older db file. It would only be downgrading SDL without recompiling that would be a problem. Though it would probably be better to select at run-time if possible anyway, so that upgrading SDL will also automatically switch to the updated db.

Thunderforge:
I pushed a fix that adds all three controller mappings and a check for them. However, it looks like I also wound up incorrectly merging commits from upstream so now they also appear as commits in this PR. Is there an easy way to fix it or is just creating a new PR a better solution?

Allofich:
> I pushed a fix that adds all three controller mappings and a check for them at runtime depending on the version of SDL. However, it looks like I also wound up incorrectly merging commits from upstream so now they also appear as commits in this PR. Is there an easy way to fix it or is just creating a new PR a better solution?

I tried pulling your branch to check it out, and the duplicate commits were automatically removed from the copy of your branch I received. It might be that they would be automatically removed if this PR was merged.

But to be safe, why not try doing the same thing?

git branch newBranch
git checkout newBranch
git pull origin bugfix/updating-controller-mappings
git push origin HEAD:bugfix/updating-controller-mappings --force

You could also try rebasing and squashing the "merge" commit if you want. Probably a good idea to first copy your work over into a new branch, first, in case you mess something up. But I think scrawl prefers to avoid having those types of commits in the commit history.

Thunderforge:
@Allofich I tried the instructions you gave, but it didn't seem to do anything. Here's my output:

```
$ git branch newBranch
$ git checkout newBranch
Switched to branch 'newBranch'
$ git pull origin bugfix/updating-controller-mappings
From https://github.com/Thunderforge/openmw
 * branch                bugfix/updating-controller-mappings -> FETCH_HEAD
Already up-to-date.
$ git push origin HEAD:bugfix/updating-controller-mappings --force
Everything up-to-date
```
Is there something obvious I'm missing?

Allofich:
Sorry, there was an unwritten assumption that you would start from the "master" branch. I think you started from the "bugfix/updating-controller-mappings" branch.

Try this:
git checkout master
git pull upstream master (update to latest master)
git branch newBranch -D (delete the "newBranch" you made)
git branch newBranch (make a new "newBranch" that is a copy of master branch)

then continue as the example was, starting from "git checkout newBranch"

Thunderforge:
@Allofich I appreciate you helping me to walk through this. I'm okay with using Git until I start running into problems (usually related to merge conflicts), so having some help resolving them is very much appreciated!

After going through the steps you provided me, it looks like 93d6f05 has my changes in one commit, but there are still more commits than intended associated with this PR. Any ideas at this point?

Allofich:
My best guess is that you're not starting from an updated master branch. As long as you've got all the commits from master branch already, I don't think they should be added in when you pull your PR branch.

Try:
- git checkout master (go to master branch on your machine)
- git pull upstream master (pull latest code from OpenMW repository to your master branch. If you don't have OpenMW's repository as your upstream, use "git pull https://github.com/OpenMW/openmw.git master"
- git branch newBranch -D (delete your old "NewBranch". Make sure you see a message confirming it was deleted. Or just make an entirely new branch "NewBranch2" or something)
- git checkout newBranch 
- git pull origin bugfix/updating-controller-mappings
- git push origin HEAD:bugfix/updating-controller-mappings --force (I think you will need to use "--force" for it to accept the push command)

If this doesn't work, I don't know what's wrong. I've never had a problem with commits from master branch appearing in my PRs.

Thunderforge:
@Allofich Still couldn't get it to work. I've replaced this PR with #1594

PR #1593, opened 2018-01-01
fix building on mips machines

psi29a:
Related to discussion here:
https://forum.openmw.org/viewtopic.php?f=6&t=4869&


AnyOldName3:
Your new name is snake case, but the developer reference says OpenMW uses camel case!

psi29a:
@AnyOldName3 corrected ;)

psi29a:
Going once... going twice... merged!

PR #1594, opened 2018-01-02
Updating controller mappings to latest SDL version

Thunderforge:
_Originally #1591, which started to have Git issues_ 

[Fixes #3708](https://bugs.openmw.org/issues/3708). I can now use my Xbox One Bluetooth controller on a Mac with SDL 2.0.6+. 

Note that this will use different mappings depending on if the linked SDL version is <= 2.0.4, 2.0.5, or >= 2.0.6. We kept this behavior because Ubuntu Xenial uses 2.0.4 and we didn't want to lock that out. Some controllers in newer mappings, like the Xbox One Bluetooth controller, are not present in older mappings.


apps/openmw/engine.cpp
```diff
@@ -477,8 +477,20 @@ void OMW::Engine::prepareEngine (Settings::Manager & settings)
...
 
     // find correct path to the game controller bindings
-    const std::string localdefault = mCfgMgr.getLocalPath().string() + "/gamecontrollerdb.txt";
-    const std::string globaldefault = mCfgMgr.getGlobalPath().string() + "/gamecontrollerdb.txt";
+    // File format for controller mappings is different for SDL <= 2.0.4, 2.0.5, and >= 2.0.6
+    SDL_version linkedSdlVersion;
+    SDL_GetVersion(&linkedSdlVersion);
+    std::string controllerFileName;
+    if (linkedSdlVersion.major == 2 && linkedSdlVersion.minor == 0 && linkedSdlVersion.patch <= 4) {
+        controllerFileName = "gamecontrollerdb_204.txt";
```
Mingun: If you already do such checks, maybe to make more generic solution: get the current SDL version, make a name of the appropriate file and to try to load it and if it is absent, use default file name? It will allow to avoid the same problems in the future.

Thunderforge: SDL 2.0.6 and 2.0.7 have the same format for controller mappings. Hopefully that means we are out of the woods for controller formats and can drop the old ones in the future. But even if 2.0.8 comes out and changes it again, then we'll have to do some logic to have one file for both 2.0.6 and 2.0.7.

Also, it's been a long time since I worked with files in C++ and all I remember with that was that it was a real pain. Knowing that the format is the same for two consecutive versions, do you still recommend going that route? If so, could you give me a pointer on how to go about doing it?

scrawl:
Could you remove the 'controllers' subdirectory? I think it's confusing that there is such subdirectory in the source repo, but none in the build or install folder.

scrawl:
I don't have a controller, can anyone test theirs still work in this PR?

Thunderforge:
> Could you remove the 'controllers' subdirectory? I think it's confusing that there is such subdirectory in the source repo, but none in the build or install folder.

Done. One advantage of this is that when we eventually drop SDL 2.0.4 and 2.0.5 support, we'll be back down to one file, and then a `controllers` subdirectory wouldn't make sense.

> I don't have a controller, can anyone test theirs still work in this PR?

If you wanted a controller, I would seriously buy one for you :-)

psi29a:
The Steam Controller is pretty slick, or so I've heard.

AnyOldName3:
It depends. It's great sometimes and is worse than a regular XInput one in other situations. In most of the situations where it beats a standard XInput controller, a mouse would beat the steam controller. There's also the problem of the massive rigmarole to set it up per application through Steam. All that said, it's my default controller when I decide to use a controller. Oh, also, it sips batteries so slowly that it makes my other wireless controllers look terrible.

Thunderforge:
@scrawl I posted [this topic](https://forum.openmw.org/viewtopic.php?f=6&t=4879) on the OpenMW forums requesting testers. One person reports that their DualShock 4 works with this PR.

PR #1595, opened 2018-01-02
Remove the 'Screenshot saved' message

akortunov:
Fixes [bug #4191](https://bugs.openmw.org/issues/4191).

In current master we display the "Screenshot saved" message before we actually take the screenshot. That's a quite odd.

EDIT: Remove the message for now due to possible threading conflicts.

### DIFF
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -1021,8 +1021,6 @@ namespace MWInput
     {
         mScreenCaptureHandler->setFramesToCapture(1);
         mScreenCaptureHandler->captureNextFrame(*mViewer);
-
-        MWBase::Environment::get().getWindowManager()->messageBox ("Screenshot saved");
     }
 
     void InputManager::toggleInventory()


scrawl:
Could you verify if the CaptureOperation is called from the main thread? It might be called from the graphics thread, in which case calling the window manager from there would be unsafe.

akortunov:
How can I verify it?

scrawl:
You could place a breakpoint in the debugger and check the call stack.

akortunov:
I suppose OpenMW run this code in the graphics thread, so maybe it would be better to look for another solution:

screenshot_20180103_215027

scrawl:
Yes please. Calling a non-thread safe function from another thread is most definitely going to lead to a crash at some point.

akortunov:
So ideally we need a way to tell the main thread if the screenshot was successfully saved.

It would be much simplier just to print message to the console instead of generating a messagebox.

kcat:
That could also help with the complaint of the "Screenshot saved" messages getting into consecutive screenshots. Perhaps there could be a separate HUD widget to inform a screenshot was saved, which is separate from and disappears quicker than a messagebox.

scrawl:
Or we remove the (rather unhelpful, as not even the location of the screenshot is included) message altogether. In my experience, the game usually freezes for a split-second when taking screenshots, which should be enough of a hint that a screenshot was taken.

akortunov:
The message is removed for now.

> Perhaps there could be a separate HUD widget to inform a screenshot was saved

Looks like an overkill for me.

PR #1601, opened 2018-01-22
use OSG instead of OpenGL directly

psi29a:
Replaces OpenGL specific code for OSG managed code.

This gives us the benefit of using GLES(1,2,3) instead of having to deal with a wrapper, important for GLES only platforms such as the RPi.

This only partially works, the MyGUI loading displays, but the rest is garbled. Any help getting this working would be appreciated.


psi29a:
@scrawl @AnyOldName3  Any ideas?

AnyOldName3:
I'm assuming you've found OSG's list of OSG equivalents to common GL calls, so there's not much to worry about there.

One thing I'd double check is that MyGUI and OSG are using the same layouts for the data they're giving to OpenGL in terms of things like vertex attribute sizes (e.g. 2D versus 3D versus 4D texture coordinates), interleaving and stride (are things aligned weirdly to get around a bug on some obscure hardware, for example) and whether OSG is setting additional state automatically when you tell it to set certain state which might interfere with MyGUI.

psi29a:
Well... I can get into a game and it seems to work, just some of the mygui elements sometimes load and sometimes not. However when Morrowind first loads, it tries to playback the intro and I can hear it but it doesn't display properly.

psi29a:
Ah hah! The last commit solves the issues I was having. It works locally here on linux x86_64 and compiles against gles2/3 in qmeu armhf environment. 

Please review and test on your machines please.

Yes, fixed_function stuff is disabled on environments where fixed_function is not supported. Fact of life until we figure out how to do it all with shaders instead.

AnyOldName3:
If things don't support fixed-function, then how is the sky working? That doesn't use shaders.

psi29a:
They don't. On GLES v2 and v3, there is no fixed-function support.  That part of the engine will be broken.

psi29a:
That was incorrectly worded, from the wiki:

> OpenGL ES 2.0 was publicly released in March 2007. It is based roughly on OpenGL 2.0, but it eliminates most of the fixed-function rendering pipeline in favor of a programmable one in a move similar to transition from OpenGL 3.0 to 3.1.

So OSG might be taking advantage of other techniques that do allow fixed-function rendering. However the use of glLightfv was removed.

Allofich:
Just wondering (not a criticism or anything), is the main motivation/benefit of this that it will allow OpenMW to run on additional devices?

psi29a:
That is correct. Currently OpenMW can't be built on armhf/armel on Debian and Ubuntu. That is my prime motivation. I managed to get this working a few years ago by getting opengl_es (1.1) supported, but that has since been dropped by these two distros.  In addition to this, on these distros, on these platforms, Qt is compiled against gles2 (not gles1) and openmw-cs, launcher and wizard could never be shipped.

OSG supports opengl_es 2 and 3, and is compiled against them currently on Debian/Ubuntu on armhf/armel, Qt is the same, so the only holdout is OpenMW itself. The code i've ifdeffed is exactly what is already done in OSG, @scrawl just copied the lighting code out of OSG and put it in OpenMW. So no magick going on there.

The end result is OpenMW will now be playable on platforms with ES2 and ES3, but without the pretty flickering lights.

The rest is untouched.

scrawl:
> now be playable on platforms with ES2 and ES3, but without the pretty flickering lights.

... and without the sky, which doesn't support shaders yet.

My opinion is the cmake options for gles1/2/3 should possibly be removed and replaced with the ones that are already set by OSG, because I can see no reason why one would compile OpenMW against a different API than OSG.

Regarding the vertex pointer calls in myguirendermanager, IIRC I made them direct calls so that the state tracker overhead can be avoided, and maybe even avoid some redundant GL calls. I'd preferably like some tests with apitrace to make sure there isn't a performance regression from this.

The direct calls we have to glLight are unlikely to be removed, because replacing with osg::Light would result in way more calls for things we don't even use like spotlights.

xyzz:
I tried this on android and it unfortunately doesn't work at all (solid screen). I then ran it under an opengles tracer and apparently it's trying to draw without ever assigning a shader program which wouldn't work on glesv2:

image

(you can see in the debugger it's funny rainbow but on the device just a solid color)

psi29a:
> > now be playable on platforms with ES2 and ES3, but without the pretty flickering lights.
> ... and without the sky, which doesn't support shaders yet.

That's just a fact of life for platforms that only support ES2/3 until lights and sky are shaders.

> My opinion is the cmake options for gles1/2/3 should possibly be removed and replaced with the ones that are already set by OSG, because I can see no reason why one would compile OpenMW against a different API than OSG.

This is reasonable, I'll look into this and see if we can get rid of the findgles files.

> Regarding the vertex pointer calls in myguirendermanager, IIRC I made them direct calls so that the state tracker overhead can be avoided, and maybe even avoid some redundant GL calls. I'd preferably like some tests with apitrace to make sure there isn't a performance regression from this.

I'll give apitrace a run through, no experience with that, to see if there is a perf hit. The reason I did this was because this functions are not supported in GLES2/3 doesn't support them. OSG's state tracker does however with another implementation that does work with GLES2/3.

> The direct calls we have to glLight are unlikely to be removed, because replacing with osg::Light would result in way more calls for things we don't even use like spotlights.

I know why you did it, that's fine. I just copied the #ifdef as well. :)
https://github.com/openscenegraph/OpenSceneGraph/blob/OpenSceneGraph-3.4.1/src/osg/Light.cpp#L122

I haven't tested on Android. I've been focusing on RPi and x86_64.

xyzz:
So it works on rpi using glesv2? Can you capture apitrace for me?

psi29a:
Here are two apitraces comparing usg_osg and master branches:
https://www.dropbox.com/s/p1mlyzcxs2xzs7n/traces.tar.bz2?dl=0

psi29a:
If I do a diff between the two, the only differences I see are memory locations and the following:

> glDisable(GL_DEPTH_TEST)
> glEnable(GL_BLEND)
> glViewport(0, 0, 1280, 720)
> glEnableClientState(GL_VERTEX_ARRAY)
> glEnableClientState(GL_TEXTURE_COORD_ARRAY)
> glEnableClientState(GL_COLOR_ARRAY)
> glBindTexture(GL_TEXTURE_2D, 5)
> glBindBuffer(GL_ARRAY_BUFFER, 1)
> glEnableClientState(GL_VERTEX_ARRAY)
> glVertexPointer(3, GL_FLOAT, 24, NULL)
> glEnableClientState(GL_COLOR_ARRAY)
> glColorPointer(4, GL_UNSIGNED_BYTE, 24, 0xc)
> glEnableClientState(GL_TEXTURE_COORD_ARRAY)
> glTexCoordPointer(2, GL_FLOAT, 24, 0x10)
> glDrawArrays(GL_TRIANGLES, 0, 6)
> glBindTexture(GL_TEXTURE_2D, 6)
> glBindBuffer(GL_ARRAY_BUFFER, 2)
> 

Looks like the ordering is just different, but the same calls are made.

psi29a:
@scrawl did you have a chance to review the apitraces?

As for using OSG to figure out which GL lib to link against... if only it was so simple, OSG does the same thing here:
https://github.com/openscenegraph/OpenSceneGraph/blob/master/CMakeLists.txt

However, this isn't shipped/distributed when installed. It is also not set in GLDefines.

We'd have to go through the process of eyeballing the shared library, on all platforms, and seeing what it is linked against so that we can link against it too. That's a lot of fucking work.

Why not leave it to the developer to set the right -D cmake flag, when linking (ld), it will complain anyway.

We currently default to desktop GL anyway.

scrawl:
> However, this isn't shipped/distributed when installed. It is also not set in GLDefines.

Check https://github.com/openscenegraph/OpenSceneGraph/blob/429aef3b08e68f5718851d7626300d8c1cb82df2/src/osg/GL.in

You could parse the installed osg/GL header. Kinda annoying to have to do that, but I don't have a better idea.

> Don't forget that OSG can be compiled against GL, GLES2 and GLES3 (and EGL) at the same time

I don't think so.

> We currently default to desktop GL anyway.

That's a good point, when the other GL's are still broken there's no need to attempt to use them. I could raise the same point for this PR as a whole, though..

psi29a:
Ah, thanks... I'll see what I can do.

My expectation was incorrect, here is RPi/Raspbian/Sid:
root@WhiteQueen:/opt/OpenMW/build# ldd /usr/lib/arm-linux-gnueabihf/libosg.so.131
	libOpenThreads.so.20 => /usr/lib/arm-linux-gnueabihf/libOpenThreads.so.20 (0xf65d7000)
	librt.so.1 => /lib/arm-linux-gnueabihf/librt.so.1 (0xf65c1000)
	libdl.so.2 => /lib/arm-linux-gnueabihf/libdl.so.2 (0xf65ae000)
	libGLESv2.so.2 => /usr/lib/arm-linux-gnueabihf/libGLESv2.so.2 (0xf6590000)
	libstdc++.so.6 => /usr/lib/arm-linux-gnueabihf/libstdc++.so.6 (0xf6484000)
	libm.so.6 => /lib/arm-linux-gnueabihf/libm.so.6 (0xf6407000)
	libgcc_s.so.1 => /lib/arm-linux-gnueabihf/libgcc_s.so.1 (0xf63dd000)
	libc.so.6 => /lib/arm-linux-gnueabihf/libc.so.6 (0xf62e9000)
	/lib/ld-linux-armhf.so.3 (0xf6fd6000)
	libpthread.so.0 => /lib/arm-linux-gnueabihf/libpthread.so.0 (0xf62c4000)
	libGLdispatch.so.0 => /usr/lib/arm-linux-gnueabihf/libGLdispatch.so.0 (0xf6246000)

Only GLESv2, but I thought I saw it also supporting GLESv3. Oh well, moving on.

darkbasic:
Great work, it would be awesome to support platforms like RPi!

psi29a:
Still looking into this, using cmake magic to auto detect what OSG uses.

psi29a:
Closing this, it is redundant with the fixes already available forcing OpenGL to be used on armel/armhf instead of GLESv1/2/3.

PR #1603, opened 2018-01-29
Adding panel for advanced settings.cfg options

Thunderforge:
Adds [Feature #4054](https://bugs.openmw.org/issues/4054). Many useful features that can only be set by manually editing `settings.cfg` can now be set through the Launcher, with the notable exception being settings related to distant terrain (see below).

# The Window
There is now a new panel with for configuring a number of options that are so far only editable in `Settings.cfg`.

advanced panel

The window's minimum width is 85 pixels wider (was 635, now 720) in order to accommodate the additional panel icon. The minimum height is unchanged (I made it bigger in the screenshot above to show all the options).

# Tooltips
Each option shows a tooltip when you hover over it (the mouse pointer isn't visible in the screenshot).

tooltip

Tooltip text is taken from the [OpenMW ReadTheDocs](http://openmw.readthedocs.io/en/master/reference/modding/settings/index.html). Unfortunately, there isn't a way to directly update the tooltips when the docs are updated.

# Settings chosen
I included all of the settings I could find except for ones that:

 - Are already available in-game and thus seemed like a waste of time to add (e.g. [`best attack`](http://openmw.readthedocs.io/en/master/reference/modding/settings/game.html#best-attack))
 - I deemed to not be useful for an end user to manipulate (e.g. the [size of the inventory window](http://openmw.readthedocs.io/en/master/reference/modding/settings/windows.html#inventory))
 - Could conceivably be on the Graphics panel (e.g. [`small feature culling pixel size`](http://openmw.readthedocs.io/en/master/reference/modding/settings/water.html#small-feature-culling-pixel-size)), which I think should be a separate PR
 - Are related to Distant Terrain, which I think should be a separate PR, given that it's a more complex feature (most of the disagreement in the Feature Request was over how to handle those settings)

I admit that this list is pretty subjective, but since it took me several minutes to add each setting, I had to draw the line somewhere.

# Value handling
If the user manually edits `settings.cfg` and sets a setting to an invalid value (e.g. setting the boolean [`show projectile chance`](http://openmw.readthedocs.io/en/master/reference/modding/settings/game.html#show-projectile-damage) to `banana`), the UI will show the default option, but will not save it unless the user interacts with the UI element.

The [`show owned`](http://openmw.readthedocs.io/en/master/reference/modding/settings/game.html#show-owned) setting accepts values `0`, `1`, `2`, and `3`. This isn't very useful for an end user, so the Launcher converts them into more readable options.

show owned options

There was particular concern about the [`screenshot format`](http://openmw.readthedocs.io/en/master/reference/modding/settings/general.html#screenshot-format) setting, given that a custom build of OpenSceneGraph could handle new image formats. The default values in the dropdown are `JPG`, `PNG`, and `TGA`. Should the user manually change the value in `settings.cfg` to a different value, it will appear in the dropdown. If the user changes to a different value and closes the launcher, it will not reappear in future sessions.

custom screenshot format


apps/launcher/advancedpage.cpp
```diff
@@ -0,0 +1,89 @@
...
+    loadSettingBool(grabCursorCheckBox, "grab cursor", "Input");
+    loadSettingBool(toggleSneakCheckBox, "toggle sneak", "Input");
+
+    // Other Settings
+    loadSettingBool(timePlayedCheckbox, "timeplayed", "Saves");
+
+    QString screenshotFormatString = QString::fromStdString(mEngineSettings.getString("screenshot format", "General")).toUpper();
+    if (screenshotFormatComboBox->findText(screenshotFormatString) == -1)
+        screenshotFormatComboBox->addItem(screenshotFormatString);
+    screenshotFormatComboBox->setCurrentText(screenshotFormatString);
```
akortunov: Looks like build fails with some versions of Qt (Qt4, I suppose):
`‘class QComboBox’ has no member named ‘setCurrentText’`

You can try to use the setItemText() instead.

lysol90:
I was fine with only using the .cfg-file, but I think _many_ other users will appreciate this a lot. 

Pity the tooltip text cant update automatically.

zinnschlag:
Not a huge fan of this approach. If we merge this change then we need to be aware that this is only a stopgap and will be replaced entirely post 1.0 (partially though in-game settings, partially through GMSTs). We are close enough to 1.0 now that I am more inclined to reject stopgap solutions.

@scrawl Got an opinion on this?

Thunderforge:
@zinnschlang I view this as something that provides a lot of polish for a 1.0 release by making it more user-friendly. I am 100% okay with having it be replaced when we are ready for a new approach.

Besides, it will take time for alternatives to be created. For instance, in-game settings will require a new localization system and volunteers who can translate into supported languages. I think that a solution like this would have a shelf life, since developing a replacement takes time.

apps/launcher/advancedpage.cpp
Thunderforge: Do we really need to support Qt4? The last version [ended support in December 2015](https://blog.qt.io/blog/2015/05/26/qt-4-8-7-released/).

AnyOldName3:
Personally, I'm in favour of having these settings available from the GUI at the 1.0 release. Whether it's implemented as a stopgap until something better gets implemented or we rush the permanent solution to get it done before 1.0, it's something that makes OpenMW feel a lot more finished and less like a demo. Not being finished is one of the biggest reasons I hear for people preferring not to use OpenMW, and if there's the sentiment that we're just sticking an arbitrary number on a release without making things 'noob friendly' I worry that some people may claim that people should ignore the 1.0 version number as we're still not ready.

lysol90:
I understand Zini's position on this, since I guess he really just don't want to encourage people to work on things that won't get us further to 1.0 by accepting PR's that won't get us further to 1.0, BUT:

I still agree with @Thunderforge and @AnyOldName3 that it just makes the engine more polished for 1.0, even though it gets replaced later.

Like my opinion matters. But just my two cents. :)

apps/launcher/advancedpage.cpp
scrawl: I've heard this narrative many times and I'm tired of it.

Developer 1: Hey, this breaks X.
Developer 2: X is old, do we still need to support X?

If you break it, you fix it. We don't remove support for something because you were too lazy to update one line. If anything, we remove support after a poll, discussion, etc, but that's hardly related to this PR.

In this case, there is an outstanding issue with Qt5 + OSG not allowing threaded rendering. It's been discussed before in other places. And then there was [this](https://stackoverflow.com/questions/40863661/how-do-i-get-the-system-qfiledialog-in-qt5) - on Ubuntu, with Qt5 you can either get native file dialogs *or* menus, but not both :(

IMO, their end of support for Qt4 was very premature. Anyway, when the time comes, we should probably update our default choice to Qt5 *before* our support for Qt4 is removed, to give users time to adapt.

scrawl:
While I'm not opposed to adding a simple thing that can be improved later, I think that now having the setting documentation in two different places could be a PITA for maintenance.

Thunderforge:
@scrawl It seems that we have a couple of options regarding that:

 1. Leave it as-is, requiring updates to the docs to be copied over
 2. Create some mechanism for generating tooltips based from the documentation
 3. Remove tooltips altogether, thus meaning there is no duplication

Maintainable, simple, or user-friendly, pick any two.

How often does the documentation for these settings actually change? If the answer is "hardly ever", then I'm not sure that there is a huge concern with copying them over, especially if this is a feature that will eventually be replaced. Heck, I'll volunteer to re-copy them before releases. In that case, option 1 would be the best choice.

I will say that I'd really hate to lose the tool-tips though, since having them goes a long way to making the feature noob-friendly option.

AnyOldName3:
This might be a little convoluted, but you could potentially get CMake to pull the tooltips from the RST files and then place it into the C++. If you did it right, then every time the affected files are built, it would automatically check if the RST files had been changed since last time, and if so, redo the copying. I think you'd do this with `configure_file` or maybe if there's something similar specifically for source files, that function instead. I'm not exactly sure how you'd get CMake to grab the documentation out of the RST file.

Thunderforge:
@AnyOldName3 The tooltip text isn't in the C++ code right now, but rather in the `.ui` files used by Qt. This definitely makes things more complicated.

I think it is possible to perhaps place tooltips into a text file or something and then have the C++ dynamically add them, but at some point you have to question if all that's worth the effort to do that, especially for a "temporary" feature.

AnyOldName3:
Have CMake shove them into the `.ui` files, then. We manage to stick the git hash into the version file, and that's not C++, for example.

apps/launcher/advancedpage.cpp
Thunderforge: @scrawl Sorry if that came across poorly. I was unaware that there had been prior discussion about it, and was merely wanting to raise the topic. My plan is still to fix this line when I get back to my home computer, and possibly raise a topic on the forums.

montreal91:
How much work it will take to make these parameters available through the in-game menu? Approximately

Thunderforge:
Qt5-only code is now replaced with Qt4 code, and the automated builds are now passing.

I took a look into how to convert the RSTs into a format that could be used for tooltips. The closest thing that I could find was [doxygen2qtcreator](https://github.com/mmmarcos/doxygen2qtcreator), but that appears to just convert to a file that is read by QtCreator, rather than the `.ui` files. I'll keep looking for options, but I'm not too optimistic about a way to automatically use the RST files.

@montreal91 As I understand it, the two things preventing these features from being added to the in-game menus are:

 - The lack of a localization framework (but one is planned for post-1.0)
 - The general desire to limit the number of non-vanilla features added before 1.0

So it's not really a matter of work and more just a matter of waiting for a more appropriate time.

Thunderforge:
I looked through the source code history and found that in the last year, the setting descriptions were changed exactly one time after they were added (d555c21), and even then it was almost exclusively to fix line feeds and other formatting errors. Therefore, I think the risk of having the tooltips be out of sync with the docs is low. 

Given that the setting descriptions change so rarely, I think that adding these settings, and having tooltips, outweighs any hypothetical headaches from them being changed in a significant capacity, especially if it will be replaced in the coming year.

zinnschlag:
@Thunderforge I think this argument is reasonable and I also get the point of delivering a more polished 1.0. Since this is only a stopgap solution that will be removed soon-ish after 1.0 anyway, I don't see a problem here.

I would request that a statement is added to the new tab indicating its transient nature. Maybe the text at the top can be adjusted? I don't think we need the "Use at your own risk" statement. These settings look risk-free to me.

Is that agreeable to everyone?

Thunderforge:
@zinnschlag So just reduce the message to "These settings are not available in game."?

zinnschlag:
That is not what I meant. The statement that the settings are not available in-game is a bit pointless anyway. Would anyone really expect there to be two different GUIs for the same function? But we need to indicate to the user that this tab is only an intermittent solution so that they don't get used to it and then later come complaining when we remove it. Seeing it happening with the data path systems already. Not keen on a repeat.

Thunderforge:
@zinnschlag Could you give an example of what you think it should say?

zinnschlag:
Hm, maybe something like this:

"This page is a stopgap solution for making cfg-file-only settings available via a GUI. It will be removed after 1.0 and settings will be made available in-game instead either as user settings or through other appropriate means."

Wording could be a bit more compact, I guess.

Thunderforge:
@zinnschlag It seems to me that the key things you want to get across are:

 - This page is temporary
 - It will be removed in the future, post-1.0
 - The settings will be available in the game

Here is my compact version that I think encapsulates that.

> This temporary page contains new settings that will be available in-game in a post-1.0 release of OpenMW.

shortened page description

zinnschlag:
Close enough.

Capostrophic:
Enchant chance checkbox does not work properly, it never has a value loaded in - instead melee info checkbox gets enchant chance value loaded in first and then it gets immediately replaced with melee info setting value.

Thunderforge:
@Capostrophic Fixed in PR #1609

PR #1604, opened 2018-02-09
Fix for https://bugs.openmw.org/issues/4319

uramer:
Tested with both the testing mesh provided by Greatness7 and problematic meshes in Tamriel Data by Project Tamriel.

### DIFF
--- a/components/nifbullet/bulletnifloader.cpp
+++ b/components/nifbullet/bulletnifloader.cpp
@@ -245,11 +245,6 @@ void BulletNifLoader::handleNiTriShape(const Nif::NiTriShape *shape, int flags, 
 {
     assert(shape != NULL);
 
-    // Interpret flags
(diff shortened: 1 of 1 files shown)


Capostrophic:
Why bother leaving the code in place if it's never going to be used again?

uramer:
I guess would be more reasonable to remove it, will change in a sec
Also I've found some bugs in the current master branch when I was testing this, should I report them and if yes, where?

akortunov:
> if yes, where?

On the bugtracker, obviously.

scrawl:
Can you make a test mesh that has no dedicated collision mesh, then set one of it's meshes to 'Hidden' to see if the hidden one then shows in the collision mesh or not?

uramer:
Ok, just done it, the result is the same as in vanilla: the hidden mesh doesn't draw, but still collides

PR #1607, opened 2018-02-13
Fix advanced settings enchant chance check box

Capostrophic:
Previously it couldn't be used for its intended purpose due to a copypaste (?) issue.


Capostrophic:
Closing as per [#1607](https://github.com/OpenMW/openmw/pull/1609#issuecomment-366865073)

PR #1608, opened 2018-02-18
Stop triggering "Journal has been updated" messagebox after Journal script function calls for the same quest stage

Capostrophic:
Currently every time you trigger a response which updates the quest stage you get the message even though you obviously don't get a duplicate entry in the journal. This doesn't look quite right.
I _think_ this is a regression from some time before 0.43.0, but I don't know exactly when this behavior has started.

### DIFF
--- a/apps/openmw/mwdialogue/journalimp.cpp
+++ b/apps/openmw/mwdialogue/journalimp.cpp
@@ -83,8 +83,10 @@ namespace MWDialogue
             if (i->mTopic == id && i->mInfoId == infoId)
             {
                 if (getJournalIndex(id) < index)
+                {
                     setJournalIndex(id, index);
-                MWBase::Environment::get().getWindowManager()->messageBox ("#{sJournalEntry}");
+                    MWBase::Environment::get().getWindowManager()->messageBox ("#{sJournalEntry}");
+                }
                 return;
             }
 


scrawl:
That change was part of 3d0094bd2e7cb4e0ee5929f3f1d909c18d4d33ff , and intended to mimic the original game which in my testing behaved like that. Feel free to test again though. With that said, I'm not saying it's the best behavior, but I would ask these questions:

* What if a quest relies on that message to convey that something happened, even though there hasn't been a new journal entry?
* What if there are stages 1 and 2, I update to 2, and then to 1? Should I get the entries in reverse order or just the entry for 2?

Ultimately, updating a quest stage like that might be more of a problem with the quest itself that should be brought to its authors attention, and the message box might be one way of doing that.

scrawl:
Oh, and feel free to put the of->if to a separate PR, because that's already irritated me countless times in fact.

Capostrophic:
I figured that dealing with false positives like this produces, well, false negatives - particularly in case of Meet Sul-Matuul quest with Nibani Maesa's dialogue - when I looked at it again and haven't come up with a different solution. Yet. I still want to do some more research on this and conjure a compromise.

What's with of->if? You've fixed the typo.

Capostrophic:
Ok, just tested Vodunius Nuccius quest (silt strider caravaneer's dialogue updates the quest stage, as evident). Master and Morrowind behaviors differ:
[Morrowind](https://pp.userapi.com/c841332/v841332669/6f80b/U5TkT5MFYi4.jpg)/[OpenMW](https://pp.userapi.com/c841332/v841332669/6f815/30VsKSCRnkU.jpg)

Still not sure what Morrowind actually does.

Capostrophic:
Just realized my fix is fine and in fact this trigger of the messagebox I moved behind a check is not necessary, as is the trigger of setJournalIndex, because it's unlikely the same topic could somehow trigger an update to a different quest stage... unless it was modded, so it's nice to leave a failsafe.

PR #1609, opened 2018-02-20
Fixing "Show Enchant Chance" checkbox setting melee info instead

Thunderforge:
Fixes issue reported [here](https://github.com/OpenMW/openmw/pull/1603#issuecomment-365290566) where the enchant chance option got mixed up with the melee info option.


akortunov:
The same thing as #1607, right?

Thunderforge:
@akortunov Yes, it looks like it is. I hadn't noticed that a PR had already been filed.

PR #1611, opened 2018-02-24
Round up encumbrance value in the inventory window (Feature #1786)

Capostrophic:
For rounding up, considering the encumbrance displayed is always a non-negative number, I use std::ceil from cmath class. It returns an integer value, so the static cast isn't necessary.


Capostrophic:
Oops. I just realized there are a few issues with this solution. Fixes incoming.

Capostrophic:
I've ran into a) some troubles with Git and b) messed it up too. Let's try again.

PR #1613, opened 2018-02-24
INI importer clean-up

Capostrophic:
See the commit message for clarification.

Some notes: Morrowind engine appears to support up to 8 slots for blood textures that can be made use of in mods - Diverse Blood utilizes them. Both OpenMW and OpenMW-CS lack such support; they're hardcoded to only assign three blood textures.

Even though OpenMW supports level up messages for levels past 20 now, you must add them manually to openmw.cfg, because INI importer only imports messages for levels from 1 to 20 and the default message. It's unclear how many level up messages should ini importer import.

### DIFF
--- a/apps/mwiniimporter/importer.cpp
+++ b/apps/mwiniimporter/importer.cpp
@@ -22,8 +22,6 @@ MwIniImporter::MwIniImporter()
     const char *fallback[] = {
 
         // light
-        "LightAttenuation:UseConstant",
-        "LightAttenuation:ConstantValue",
         "LightAttenuation:UseLinear",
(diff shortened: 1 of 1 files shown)


scrawl:
Aren't there plenty more settings that aren't being used, some with the intention they might be used in the future? If we have to re-add a setting later, everyone is going to have to re-run the importer, so I think It's better to be on the safe side and keep it just in case that we end up using the setting after all.
With that in mind, I am not sure what the goal of this cleanup would be.

Capostrophic:
Hm. Well, indeed.

Though there's still the matter of blood settings being listed twice for no reason.

PR #1614, opened 2018-02-25
Autoequip ammo when using ranged weapon

akortunov:
Currently autoequipping during actor creation ignores ammo.

With this PR:
1. If NPC wants to autoequip the ranged weapons, equip the ammo with best damage.
2. If there is no suitable ammo, do not equip the ranged weapon.

Can be useful if we will implement weapon sheathing and quivers.

### DIFF
--- a/apps/openmw/mwworld/inventorystore.cpp
+++ b/apps/openmw/mwworld/inventorystore.cpp
@@ -376,6 +376,34 @@ void MWWorld::InventoryStore::autoEquip (const MWWorld::Ptr& actor)
 
     bool weaponSkillVisited[weaponSkillsLength] = { false };
 
+    // give arrows/bolt with max damage by default
+    int arrowMax = 0;
+    int boltMax = 0;
+    ContainerStoreIterator arrow(end());
+    ContainerStoreIterator bolt(end());
+
+    for (ContainerStoreIterator iter(begin(ContainerStore::Type_Weapon)); iter!=end(); ++iter)
+    {
+        const ESM::Weapon* esmWeapon = iter->get<ESM::Weapon>()->mBase;
+
+        if (esmWeapon->mData.mType == ESM::Weapon::Arrow)
+        {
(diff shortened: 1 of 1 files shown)


scrawl:
I don't see a point in pulling this if the other PR won't be accepted, so it should probably be a part of that PR.
Anyway, won't equipping the ammo cause merchants to no longer sell it?

akortunov:
> Anyway, won't equipping the ammo cause merchants to no longer sell it?

Only if the merchant:
1. Has the ranged weapon equipped (bows for arrows, crossbow for bolts).
2. Weapon and ammo is in his inventory
And with "prevent merchant autoequipping" feature he will autoequp weapon and ammo only if he was an original owner of these items.
It's a quite hard to tell which equipment is for trading and which for personal use.

Also I noticed that creatures do not autoequip weapons, they equip only shields. It is a bug or a feature?

scrawl:
> 1. Has the ranged weapon equipped (bows for arrows, crossbow for bolts).
> 2. Weapon and ammo is in his inventory

That sounds like a common occurence still. 

Even if we get quivers, is it really necessary to visible carry them before any kind of combat starts? We can easily imagine that they're stashed away somewhere just like all the rest of the things that NPCs tend to carry. At the very least, I would disable quivers equipping for merchants before combat.

Anyway, this should probably be merged into the other PR so we have only one hanging PR instead of two.

> Also I noticed that creatures do not autoequip weapons, they equip only shields. It is a bug or a feature?

Probably neither, because it never matters if they're not visible.
Theoretically, you could have a humanoid creature merchant which should or should not be selling its weapon. There's nothing like that in the game from what I know, but anyway it might be an interesting research question.

akortunov:
> Probably neither, because it never matters if they're not visible.

Matters for #1615, that's why I asking.

akortunov:
Closing this PR for now since quivers are not ready for implementation. I suggest to create a forum discussion first.

PR #1615, opened 2018-02-25
Weapon sheathing

akortunov:
Implements [feature #4673](https://gitlab.com/OpenMW/openmw/issues/4673).
Related topic is [here](https://forum.openmw.org/viewtopic.php?f=6&t=4973).
Here are WIP animations for this feature: [Animations.zip](https://github.com/OpenMW/openmw/files/2430573/Animations.zip)

[Link](https://ci.appveyor.com/api/buildjobs/c4ycvfcx67c4lbhu/artifacts/OpenMW_MSVC2015_x64_1615_92e45507d8bba9321163e3bfd3099cd92ac9de20.zip) to archive with current build.

Basic idea: add some new bones to humanoid skeleton and attach weapons/quivers/scabbards/etc to these bones.

Data format is compatible with the [Weapon Sheathing mod](https://www.nexusmods.com/morrowind/mods/46069), so you can use meshes from this mod.
xbase_anim_sh.nif is required, other files are optional.

Sheathed models for now are separate meshes with _sh suffix. If there is no scabbard for current weapon mesh, an engine will show only holstered weapon. Also scabbards and quivers do not have enchanting glow (but arrows in quiver do).

The feature is disabled by default. You should enable it in config file:

[Game]
weapon sheathing = true
use additional anim sources = true

Now this feature supports:
1. Weapon holstering.
Holstered weapons also has glow when enchanted.
2. Scabbards for any type of weapon excepts of throwing.
Scabbard itself does not have a glow.
3. Quivers for bows and crossbows.
Quivers itself do not have a glow, but arrows have when enchanted.
Count and type of rendered arrows depends on type and count of equipped arrows.

Notes:
1. I had to rework autoequipping
2. Shield holstering is not supported for now - there is a lot of issues with resources (we will need to add new animation groups, there is a lot of clipping with movement animations, issues with mods that use pseudo-shields, etc).

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -166,7 +166,7 @@ namespace MWClass
             getContainerStore(ptr).fill(ref->mBase->mInventory, ptr.getCellRef().getRefId());
 
             if (hasInventory)
-                getInventoryStore(ptr).autoEquipShield(ptr);
+                getInventoryStore(ptr).autoEquip(ptr);
         }
     }
 
--- a/apps/openmw/mwrender/actoranimation.cpp
+++ b/apps/openmw/mwrender/actoranimation.cpp
@@ -1,5 +1,4 @@
 #include "actoranimation.hpp"
-
 #include <utility>
 
 #include <osg/Node>
@@ -9,10 +8,21 @@
 #include <components/esm/loadligh.hpp>
 #include <components/esm/loadcell.hpp>
 
+#include <components/resource/resourcesystem.hpp>
+#include <components/resource/scenemanager.hpp>
+
+#include <components/sceneutil/attach.hpp>
 #include <components/sceneutil/lightmanager.hpp>
 #include <components/sceneutil/lightutil.hpp>
+#include <components/sceneutil/visitor.hpp>
 
 #include <components/fallback/fallback.hpp>
+
+#include <components/misc/stringops.hpp>
+
+#include <components/settings/settings.hpp>
+
+#include <components/vfs/manager.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
@@ -43,6 +53,8 @@ ActorAnimation::ActorAnimation(const MWWorld::Ptr& ptr, osg::ref_ptr<osg::Group>
 
     // Make sure we cleaned object from effects, just in cast if we re-use node
     removeEffects();
+
+    mWeaponSheathing = Settings::Manager::getBool("weapon sheathing", "Game");
 }
 
 ActorAnimation::~ActorAnimation()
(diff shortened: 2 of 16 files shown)


apps/openmw/mwrender/npcanimation.cpp
```diff
@@ -959,6 +959,44 @@ void NpcAnimation::updateHolsteredWeapon()
...
     }
+
+    MWWorld::ConstContainerStoreIterator ammo = inv.getSlot(MWWorld::InventoryStore::Slot_Ammunition);
+    if (ammo != inv.end() && ammo.getType() == MWWorld::ContainerStore::Type_Weapon)
+    {
+        if (weapon == inv.end()) return;
+
+        const std::string &type = weapon->getClass().getTypeName();
+
+        // Show quiver only if the ranged weapon with the suitable ammo is equipped.
```
kcat: IMO, it should show the quiver as long as the ammo is equipped, regardless of if you have a bow/crossbow equipped or not.

akortunov: No, you need the bow/crossbow equipped (line 980).
Also there is no much review for now - the code is in the early stage, and I even do not know if my approach is acceptable.

kcat: > No, you need the bow/crossbow equipped (line 980).

That's what I mean. In my opinion, it shouldn't bother with that check and always show the quiver as long as the ammo is equipped.

akortunov: > That's what I mean. In my opinion, it shouldn't bother with that check and always show the quiver as long as the ammo is equipped.

We tried that, but quiver collides with two-handed weapon, so we decided to add this check, as in Skyrim, for now.

Capostrophic: Skyrim doesn't have such check. Quiver is always shown if ammo is equipped.
Screenshot

akortunov:
Some experiments:

screenshot_20180225_205552
screenshot_20180225_210053

Of course later modders can change layout to fit draw/holster animations.

apps/openmw/mwrender/npcanimation.cpp
```diff
@@ -264,6 +264,15 @@ static NpcAnimation::PartBoneMap createPartListMap()
     result.insert(std::make_pair(ESM::PRT_LPauldron, "Left Clavicle"));
     result.insert(std::make_pair(ESM::PRT_Weapon, "Weapon Bone"));
     result.insert(std::make_pair(ESM::PRT_Tail, "Tail"));
+
+    // These ones supposed to store the holstered weapons, as in later TES games, but without scabbard
```
Capostrophic: "These are for storing sheathed weapons (no scabbard).

I'd avoid copyrighted IP references, considering OpenMW isn't supposed to be Morrowind-only. :P

Side note: scabbard support in later TES games is pluginless: the engine searches for a shape named scb (I assume "scabbard" alias would also work if it could be implemented in openmw) in the mesh and if it finds one, it assumes it's the scabbard and leaves it on its place when the player unsheathes the weapon. Could be a cool feature to replicate, while you are still working on this.

apps/openmw/mwrender/npcanimation.cpp
```diff
@@ -893,6 +904,101 @@ void NpcAnimation::addControllers()
...
+
+        if (needToShowQuiver)
+        {
+            osg::Vec4f glowColor = getEnchantmentColor(*ammo);
+            std::string ammoMesh = ammo->getClass().getModel(*ammo);
+            if (ammoMesh.empty()) return;
+
+            std::string quiverMesh;
+            quiverMesh.append(ammoMesh.begin(),ammoMesh.end()-4);
+            quiverMesh.append("_quiver.nif");
```
Capostrophic: What about osgt, the native mesh format™®©?

akortunov: > "Just a proof of concept for now."

I used a stopgap solution to show a possibility to implement such feature in OpenMW.
I guess there should be better possibilities than using suffix.

apps/openmw/mwrender/npcanimation.cpp
akortunov: So in this case I need a way to attach just a node instead of whole model. Interesting.

apps/openmw/mwrender/npcanimation.cpp
akortunov: I wonder how quivers are implemented in other games (e.g. in Skyrim).

Capostrophic:
Concerning quivers: in Skyrim, it's a bit complex. There are two meshes you need for setting a ranged weapon up: one is for quiver with five static arrows (their IDs in meshes are numbered so the engine knows which not to enable if the character only has 4 or less arrows) and one arrow which the character always gets out from the quiver, and one is for a single arrow. In content files there are two records for an arrow/bolt: one (AMMO) defines arrows of a certain type in general and another (PROJ) defines the flying arrow projectile.

Thunderforge:
> That's all - no scripting, no every NIF conversion, no ESP.

> Just install two meshes as a common ESP-less mod and move around the world.

So this is going to be a mod eventually? I'm confused as to why this needs to be a PR for the game.

Would it make sense to configure it somehow so that it can be a `settings.cfg` setting?

kcat:
> So this is going to be a mod eventually? I'm confused as to why this needs to be a PR for the game.

It will need a mod to work, but it also needs engine-side support. The engine needs to know that if an NPC has certain nodes, it should render equipped weapons on them when sheathed. The default nifs don't contain those nodes so users won't see any change automatically -- someone will need to make a mod with the necessary changes to see a difference (provided separately, like on Nexus).

akortunov:
After additional testing I noticed that crossbowmans in Morrowind use a some kind of ammo belt instead of quivers, judging by reload animation.

So maybe I will have to split quivers and ammo belts to separate feature.

akortunov:
Also can anyone explain why it does not work for beast races and bipedal creatures?
I thought the engine supposed to load xbase_anim.nif for them too, but the NodeMap does not contain new nodes.

kcat:
> Also can anyone explain why it does not work for beast races and bipedal creatures?

IIRC, beast races use a different skeleton, one which includes tail bones that don't exist in the normal biped skeleton. They have the suffix "_kna" (Khajiit 'n Argonian).

For biped creatures, I'm not sure. The model and animation handling for creatures and NPCs are in different classes, but since biped creatures share some NPC-like behavior with regard to some equipment like weapons, I don't know how exactly they're handled.

akortunov:
I uploaded the updated version:
1. **Creatures support.** Unfortunately you need to edit mesh for any creature  you want to suppport weapon sheathing. Also I noticed that creatures do not autoequip weapons after spawn, they equip only shield. But it is a separate bug.
2. **Refactoring**
3. **Drop quivers support** for now since we have not design yet.

testman42:
> Would it make sense to configure it somehow so that it can be a `settings.cfg` setting without a mod?

If this is going to be part of OpenMW codebase, user should be able to toggle this behavior in settings / in launcher.

akortunov:
> If this is going to be part of OpenMW codebase, user should be able to toggle this behavior in settings / in launcher.

Again, this setting will do nothing if the player does not have modded assets (xbase_anim* at least).
This feature required support by both assets and engine.

akortunov:
Also I got stuck with Golden Saints. They do not have the xCreatureName.nif.
Is there any way to force an engine to load some nodes from xbase_anim.nif?

scrawl:
I haven't really reviewed this, so just a general comment for now:
The main concern with a feature like this, adding new meanings for node names or filenames has an ever so small chance of something triggering that name by accident. Just for that reason e.g. the 'auto normal maps' feature is also disabled by default. 
In any case we should be sure the feature is perfect before we roll it because people will start making mods and any further updates we make will probably break those mods.
It's a bit comparable to introducing new scripting functions before 1.0 which, for the record, Zini seems to be against.
Do you think the feature could be cleaner if e.g. the ESM format was updated so you can make new body part types?

Re: the last comment, I don't see another way than editing models because I'm sure the location you want at depends on the model.

akortunov:
> Re: the last comment, I don't see another way than editing models because I'm sure the location you want at depends on the model.

What kind of model? Weapon or creature ones? For example, sheathed weapon may have different position for humanoid NPCs and beast ones.

akortunov:
A bit offtopic question: why do not add an ability to load animations from per-group .kf files? Performance issues?

It would be nice to provide a couple of modded holster animations for weapon sheathing without replacing the whole xbase_anim.kf

akortunov:
Update:
1. Make this feature optional
2. Squash commits
3. Merge upstream changes from master.

apps/openmw/mwworld/inventorystore.cpp
```diff
@@ -383,7 +392,16 @@ void MWWorld::InventoryStore::autoEquip (const MWWorld::Ptr& actor)
 
         for (int j = 0; j < static_cast<int>(weaponSkillsLength); ++j)
         {
-            int skillValue = stats.getSkill(static_cast<int>(weaponSkills[j])).getModified();
+            int skillValue = 0;
+            if (actor.getClass().isNpc())
+            {
+                MWMechanics::NpcStats& stats = actor.getClass().getNpcStats(actor);
+                skillValue = stats.getSkill(ESM::Skill::Unarmored).getModified();
```
MiroslavR: Why did you replace `weaponSkills[j]`? I'm pretty sure I checked that against the game's assembly when I implemented it.

akortunov: Sorry, my fault. Thanks for spotting that.

akortunov:
How it works for scabbards:

image

akortunov:
Added initial quivers support:

Quiver is a part of bow, similarly to scabbard.
Type and count of arrows in quiver depends on count and type of equipped arrows.

Known issues:
1. Autoequipping ignores ammo, so quivers will be empty for NPCs before first combat.
If we force NPCs to autoequip ammo, it may alter item list for some traders.
2. I need find a way to update the quiver after equipped ammo unstacking/restacking .

akortunov:
>  it may alter item list for some traders

I modified autoequipping, but the issue with trader still exists.
How can we workaround this? Ignore restockable ammo?

Florianjw:
Oh, and having shields shouldered when you are not in combat would also be awesome.

Capostrophic:
It has been thought of before. Even though it's not difficult to implement, ultimately injecting and distributing quivers in leveled lists requires too much user intervention and heavily alters the game mechanics.

Florianjw:
Leveled lists are by far the thing that I believe would have the fewest problems: If you take a look [here](http://en.uesp.net/wiki/Morrowind:L_m_enchantitem_temple_rank0_1#l_n_wpn_missle_Series), `l_n_wpn_missle_bow` is the only list that really would have to be touched, and the simplest way to do so would be to replace it's entries by pairs of bow and quiver.

NPCs are much more likely to pose a problem here, but even for them it should be possible to simply ad the item via a basic mod. (I don't see why openmw shouldn't come with some mod that fixes minor details, especially when the alternative thing is to [hard-code](https://github.com/OpenMW/openmw/blob/e197f5318b7a2f820694096e82964142635ad28c/apps/openmw/mwscript/statsextensions.cpp#L228) bug-fixes into the codebase.)

akortunov:
No, separate slot for quivers it is not an option since it will break landmass mods (e.g. Tamriel Rebuilt).
Also making quiver a part of bow will make modeller's life easier - he need to prevent overlapping with single bow, not with any weapon in game.

akortunov:
Here is an example how we can render shield on the spine:
screenshot_20180520_132529

However, we do not have shield [un]equipping animations. Also there is overlapping with hands during walking or running.

akortunov:
Updates:
1. New archive with testing resources, including new animations.
2. Upsteram changes from master, so you can test this feature with these animations.

Any feedback is welcome.

akortunov:
@zinnschlag, do you have something about quivers and scabbards in your post-1.0 plans?

zinnschlag:
I do not (yet). Still that does sound like a post-1.0 feature. This we do have a procedure for handling post-1.0 PRs now, I nominate this PR.

psi29a:
For now, shouldn't we mark this as [post-1.0] in the title?  I just hope this doesn't collect cruft later on and merge conflicts.

zinnschlag:
Well, the github PR list is kinda messy. Cleaning it out seems like a good idea to me.

zinnschlag:
I retract my comment about the github PR list being messy. I just found the option to sort PRs by the date of recent updates. Had I have known about it a few months ago that would have saved me a ton of trouble.

The option was just right outside of the browser window (had to scroll only a bit to the right). Fuck fixed-width layouts! There would have been space for that option in my browser window, if GitHub wouldn't insist on this nonsense. Good thing that I am not a person who normally rage-quits, because otherwise that would be exactly what I would do with GitHub right now.

Okay. End of rant. Marking it as post-1.0 is fine then.

akortunov:
I updated data format. Now it should be compatible with MWSE sheathing feature.

apps/openmw/mwclass/creature.cpp
```diff
@@ -166,7 +166,13 @@ namespace MWClass
             getContainerStore(ptr).fill(ref->mBase->mInventory, ptr.getCellRef().getRefId());
 
             if (hasInventory)
-                getInventoryStore(ptr).autoEquipShield(ptr);
+            {
+                bool weaponSheathing = Settings::Manager::getBool("weapon sheathing", "Game");
+                // we need a weapon autoequipping for weapon sheathing feature.
+                // Theoretically, it can affect bipedal creatures-traders from mods, so disable it if not needed.
+
+                weaponSheathing ? getInventoryStore(ptr).autoEquip(ptr) : getInventoryStore(ptr).autoEquipShield(ptr);
```
zinnschlag: Very creative use of the ?: operator ;)

apps/openmw/mwrender/actoranimation.cpp
```diff
@@ -51,6 +64,260 @@ ActorAnimation::~ActorAnimation()
...
     }
+
+    mScabbard.reset();
+}
+
+PartHolderPtr ActorAnimation::getWeaponPart(const std::string& model, const std::string& bonename, bool enchantedGlow, osg::Vec4f* glowColor)
+{
+    osg::Group* parent = getBoneByName(bonename);
+    if (!parent)
+        return NULL;
```
zinnschlag: Use nullptr please.

apps/openmw/mwrender/npcanimation.cpp
```diff
@@ -1185,4 +1201,9 @@ void NpcAnimation::setAccurateAiming(bool enabled)
     mAccurateAiming = enabled;
 }
 
+bool NpcAnimation::isArrowAttached()
```
zinnschlag: Shouldn't this be const?

akortunov:
Before actual review I'd ask if this feature shuld be embedded to the engine (in the hardcoded way), or it should by implemented by Lua scripting.

apps/openmw/mwworld/containerstore.hpp
```diff
@@ -68,6 +68,9 @@ namespace MWWorld
 
             static const std::string sGoldId;
 
+        protected:
```
zinnschlag: Why is this protected?

zinnschlag: Never mind. My bad.

psi29a:
Honestly, this seems like something we would want in the engine itself.

I would like to think that there are some mods that eventually make it into the engine itself for the benefit of all games based on the engine. In this case, is this something that is applicable for TES4 and TES5?

apps/openmw/mwrender/actoranimation.cpp
akortunov: Unfortunately, we already use NULL in animation code everywhere.
Should we create a separate pull request to replace all NULLs in OpenMW to nullptr?

akortunov:
Reworked implementation of outo-equipping.
Summary of changes:
1. We equip weapons and ammo only when "sheathing" is enabled.
2. We do not autoequip weapon and ammo when actors sells weapons.
3. I had to split the autoEquip() function to two separate ones - for weapon and for armor - to do not wrap big loops to "if" statements.
4. We do not need the autoEquipShield() anymore, so I removed it.

akortunov:
Added support for throwing weapons.

MiroslavR:
> 4\. We do not need the autoEquipShield() anymore, so I removed it.

Why? Where's the code now and did you keep the previous behavior (pick the shield with the most health)? Please make sure you don't regress https://gitlab.com/OpenMW/openmw/issues/3704.

akortunov:
> Why? Where's the code now and did you keep the previous behavior (pick the shield with the most health)? Please make sure you don't regress https://gitlab.com/OpenMW/openmw/issues/3704.

Shields are handled inside the autoEquipArmor(). Actors choose shields with the highest effective armor rating, so creatures still equips theri shields during creation.
Note: autoequipping code does not take in account items condition at all. Is that intended behaviour intended (since all items have full health during character creation), or it is better to change it?

akortunov:
Ok, since creatures do not use the damage mitigation, it would be better to keep the custom logic for them (choose shield based on its health) and explain why this logic is needed in comment.

akortunov:
Should be fine now.

apps/openmw/mwworld/inventorystore.cpp
```diff
@@ -363,161 +535,79 @@ void MWWorld::InventoryStore::autoEquip (const MWWorld::Ptr& actor)
...
+    mUpdatesEnabled = false;
+
+    bool weaponSheathing = Settings::Manager::getBool("weapon sheathing", "Game");
+    if (weaponSheathing)
+    {
+        // We should not auto-equip weapons and ammo if actor is a trader and sells them
+        int services = actor.getClass().getServices(actor);
+        bool sellsWeapon = services & (ESM::NPC::Weapon|ESM::NPC::MagicItems);
+        if (!sellsWeapon)
+            autoEquipWeapon(actor, slots_);
```
MiroslavR: Another regression? Why is this limited to weapon sheathing being enabled?

MiroslavR: IIRC, auto-equipping weapons is desired even if they are not shown visually. They still have impact on gameplay - in particular, you should not be able to pickpocket equipped weapons. I think there is a (resolved) bug report regarding that but I can't find it right now.

akortunov: Ok, I'll change it in this case.

akortunov:
Ok, some additional investigation about how weapon autoequipping works in Morrowind:
1. NPCs autoequip both their weapons and ammo. If there in no ammo for ranged weapon, they do not equip it.
2. Creatures do not equip their weapons and ammo, but we need it for weapon sheathing.
In theory, the only case when it matters - if the creature sells weapon, so we should not use autoequipping for such actors. In other cases it should be safe to use it.

apps/openmw/mwworld/inventorystore.cpp
akortunov: For now I use weapon and ammo autoequipping for all NPCs, as in Morrowind, also for creatures which not sell weapons. Is it OK?

MiroslavR: Sounds good to me.

iillyyaa2033:
Btw is shield on spine protecting your back?

akortunov:
Resolved conflicts with Capostrophic's changes.

kurnevsky:
I get "this creature has no inventory store" error on this brach when script calls `RemoveItem` on creature.

akortunov:
> I get "this creature has no inventory store" error on this brach when script calls RemoveItem on creature.

Ok, I will investigate it.

akortunov:
> I get "this creature has no inventory store" error on this brach when script calls RemoveItem on creature.

Should be fixed now.

lysol90:
So with 0.45 around the corner, what the status of this now?

akortunov:
> So with 0.45 around the corner, what the status of this now?

Well, scabbards and quivers should work, but I did not get much feedback yet.
Shields hostering is not implemented - it will require to re-desingn of how we update shields state. I'd prefer to implement it in a separate PR later.
And I am not sure if this feature should be in the 0.45 release since it was not tested properly yet.

iillyyaa2033:
> And I am not sure if this feature should be in the 0.45 release since it was not tested properly yet.

What should be tested?

akortunov:
> What should be tested? Can I help with testing?

Ideally, there should be a gameplay test with this feature enabled and disabled to make sure that it does not break anything and does not cause crashes.

iillyyaa2033:
Got it. I'll test & report during next week.

psi29a:
@zinnschlag if the plan is to push back RC to December because of blockers and timing, can this be merged into master for more exposure (testers)?

iillyyaa2033:
Just compiled openmw at latest commit from 'holstered_weapons' branch.

I have problems with inventory - items added to player but not deleted from chest.

Here's screenshot of console output: image

And this is video: https://youtu.be/r7evSKLnoAQ

This is my config: [settings.cfg.zip](https://github.com/OpenMW/openmw/files/2517983/settings.cfg.zip)

Also Sellus Gravius gives me nothing, so I cant finish fist quest and make savegame.

Steps to reproduce:
1. Compile;
2. Start new game;
3. Open chest in second room or open barrel with Fargoth's ring.

akortunov:
Try again.

zinnschlag:
@psi29a There is no plan to push back the release. it seems we may have a fix for the blocker now. Still waiting for a confirmation. If that works out we will start the RC phase either today or tomorrow.

akortunov:
> Still waiting for a confirmation. If that works out we will start the RC phase either today or tomorrow.

I suggest to do not hurry - it seems Capostrophic broke the HUD weapon condition indicator.

zinnschlag:
Already taken care of.

iillyyaa2033:
I've added `weapon sheathing = true` and `use additional anim sources = true` to the `[Game]` section of `settings.cfg`, unzipped `Animations/` folder into my Morrowind data, but no equipped weapons shown:
screenshot000

Am I doing something wrong?

akortunov:
> Am I doing something wrong?

Are you sure that you installed meshes from Weapon Sheathing MWSE mod?

iillyyaa2033:
No. There's nothing in first post about it.

Okay, I'm going to install it.

psi29a:
Ready for 0.46? :)

akortunov:
I suppose we mostly fixed the RecastNavigation mess, right?

psi29a:
The dust host mostly settled.
@AnyOldName3 You ready? :)

AnyOldName3:
I don't think I have any particular objections to merging this.

lysol90:
@AnyOldName3 I think he meant if you're ready with the fixes required after the recastnav-mess... Right..?

iillyyaa2033:
I haven't faced with bugs. But I've played without mods.

psi29a:
Just for the record, this PR also resolves one of the last warnings being triggered by openmw in msvc 2015/2017.
#2019#issuecomment-435651599

akortunov:
So what did we decide? Are there issues which I need to address, or we can merge this PR?

apps/openmw/mwworld/inventorystore.cpp
```diff
@@ -246,25 +251,194 @@ MWWorld::ContainerStoreIterator MWWorld::InventoryStore::findSlot (int slot) con
...
-    static float fUnarmoredBase2 = store.find("fUnarmoredBase2")->mValue.getFloat();
-    int unarmoredSkill = actor.getClass().getSkill(actor, ESM::Skill::Unarmored);
-
-    float unarmoredRating = (fUnarmoredBase1 * unarmoredSkill) * (fUnarmoredBase2 * unarmoredSkill);
+    float unarmoredRating = 0;
+    int unarmoredSkill = 0;
+    if (actor.getClass().isNpc())
+    {
+        MWMechanics::NpcStats& stats = actor.getClass().getNpcStats(actor);
+        unarmoredSkill = stats.getSkill(ESM::Skill::Unarmored).getModified();
```
Capostrophic: You can just use actor.getClass().getSkill(actor, ESM::Skill::Unarmored) here.

apps/openmw/mwworld/inventorystore.cpp
```diff
@@ -246,25 +251,194 @@ MWWorld::ContainerStoreIterator MWWorld::InventoryStore::findSlot (int slot) con
...
     const MWWorld::Store<ESM::GameSetting> &store = world->getStore().get<ESM::GameSetting>();
 
-    static float fUnarmoredBase1 = store.find("fUnarmoredBase1")->mValue.getFloat();
-    static float fUnarmoredBase2 = store.find("fUnarmoredBase2")->mValue.getFloat();
-    int unarmoredSkill = actor.getClass().getSkill(actor, ESM::Skill::Unarmored);
-
-    float unarmoredRating = (fUnarmoredBase1 * unarmoredSkill) * (fUnarmoredBase2 * unarmoredSkill);
+    float unarmoredRating = 0;
+    int unarmoredSkill = 0;
+    if (actor.getClass().isNpc())
```
Capostrophic: This method is never supposed to be called for creatures, is it? You could put this check to the beginning to discard everything else for invalid class actor arguments.

docs/source/reference/modding/settings/game.rst
```diff
@@ -171,9 +171,18 @@ followers attack on sight
...
 This setting can be toggled in Advanced tab of the launcher.
 
+weapon sheathing
+----------------
+
+:Type:		boolean
+:Range:		True/False
+:Default:	False
+
+If this setting is true, OpenMW will render holstered weapons (including quivers and scabbards), if the player has assets with this feature support.
```
Capostrophic: If this setting is true, OpenMW will utilize weapon sheathing-compatible assets to display holstered weapons.

To make use of this, you need to have an xbase_anim_sh.nif file with weapon bones that will be injected into the skeleton. Additional _sh suffix models are not essential for weapon sheathing to work but will act as quivers or scabbards for the weapons they correspond to.

files/settings-default.cfg
```diff
@@ -239,6 +239,9 @@ barter disposition change is permanent = false
 # 2 means werewolves are ignored) 
 strength influences hand to hand = 0
 
+# Render holstered weapon, requires modded assets
```
Capostrophic: holstered weapons

apps/openmw/mwrender/actoranimation.cpp
```diff
@@ -51,6 +63,302 @@ ActorAnimation::~ActorAnimation()
...
+
+    SceneUtil::FindByNameVisitor findVisitor (boneName);
+    mObjectRoot->accept(findVisitor);
+
+    return findVisitor.mFoundNode;
+}
+
+std::string ActorAnimation::getHolsteredWeaponBoneName(const MWWorld::ConstPtr& weapon)
+{
+    std::string boneName = "";
```
Capostrophic: You don't need to assign a string to "" to keep it empty. Simply declare it like `std::string boneName;`. That will be a little faster and cleaner. Same thing in the getHolsteredWeaponBoneName(weaponType).

akortunov: We use the same approach in many other places. If we want to change it, I suppose we need to replace other usages too (better as a separate PR).

Capostrophic: Zini would have vetoed such a PR for very good reasons.

akortunov: OK, I replaced it, but it does not make situation better - we still widely "" in OpenMW code.

psi29a: We've done clean-ups in the past in PRs, like the push to C++11. 

I believe that whole-sale cleanups, like code re-formating, would be rejected because it adds nothing and masks git history. 

So a separate PR simply dropping empty string decelerations is in between the two, however in the grand scheme of things isn't that bad but more welcome.

psi29a:
Last call before merging... any other comments or reviews?

akortunov:
> Last call before merging... any other comments or reviews?

I suppose there was enough time to get feedback.

psi29a:
Merged!

apps/openmw/mwrender/actoranimation.cpp
```diff
@@ -51,6 +63,302 @@ ActorAnimation::~ActorAnimation()
     {
         mInsert->removeChild(iter->second);
     }
+
+    mScabbard.reset();
```
elsid, line 67: Is there a purpose to call reset here? C++ compilers generate code to autmatically destruct all members.

PR #1616, opened 2018-02-25
Ownership by dead actors is not cleared from picked items (Bug #4328)

Capostrophic:
This solution should suffice for ~~two cases~~ ~~one case~~ damn if I know what from the [report](http://bugs.openmw.org/issues/4328): if the owner of an item is dead at the time the player picks a certain item up. However, I'm not sure how to go around clearing the item from StolenItemsMap upon game loading. From what I can tell, this requires:

- messing with save loading
- getting the actor state during save loading
- clearing the items with dead owners *every time* a save is loaded
- probably more awkwardness

Any more practical ideas?

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -940,7 +940,8 @@ namespace MWMechanics
         if (!cellref.getOwner().empty())
         {
             victim = MWBase::Environment::get().getWorld()->searchPtr(cellref.getOwner(), true);
-            if (victim.getClass().getCreatureStats(victim).isDead() && isOwned == true)
+            const bool ownerIsDead = victim.getClass().getCreatureStats(victim).isDead();
+            if (ownerIsDead && isOwned)
                 isOwned = false;
         }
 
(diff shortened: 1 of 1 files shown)


kcat:
> However, I'm not sure how to go around clearing the item from StolenItemsMap upon game loading.

I personally wouldn't bother. Loading a save should be transparent to the game being loaded, and doesn't provide any behavior the game can rely on. It can also make bug hunting harder, if saves provided by players exhibit different behavior when loaded than when they made the save.

I'm actually a bit more curious about it losing the stolen tag when dropping and picking the item back up after the owner died. It's almost certainly a bug because the behavior makes no sense, but it's hard to say whether the game meant for it to auto-clear after the owner died, or stay stolen if they were alive when taken. I don't think that should affect this PR though, just musing for future stuff.

Capostrophic:
>Loading a save should be transparent to the game...

Yeah, that makes sense.

Hm. Whoops.

Capostrophic:
Apparently losing the stolen tag is not the case, according to the reporter.
I didn't test the change yet. Probably doesn't work.

Capostrophic:
Apparently (again) taking items from dead actors is still a crime, but the items themselves are not considered stolen. So this needs a completely different solution.

PR #1617, opened 2018-02-26
Ownership by dead actors is not cleared from picked items (Bug #4328)

Capostrophic:
[Tracker](http://bugs.openmw.org/issues/4328)

Prevent mStolenItems for specific item ID from increasing in this case.

Note: the stolen tag is not cleared from such items upon loading a saved game.

### DIFF
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -1120,7 +1120,7 @@ namespace MWMechanics
         if (!Misc::StringUtils::ciEqual(item.getCellRef().getRefId(), MWWorld::ContainerStore::sGoldId))
         {
             const MWWorld::Ptr victimRef = MWBase::Environment::get().getWorld()->searchPtr(item.getCellRef().getOwner(), true);
-            if (!victimRef.getClass().getCreatureStats(victimRef).isDead())
+            if (victimRef.isEmpty() || !victimRef.getClass().getCreatureStats(victimRef).isDead())
                 mStolenItems[Misc::StringUtils::lowerCase(item.getCellRef().getRefId())][owner] += count;
         }
         if (alarm)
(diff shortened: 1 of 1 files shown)


akortunov:
I wonder how the game supposed to behave if the owner is not inique NPC, e.g. guard.

scrawl:
I think you forgot to check if victimRef was not found, which because you set activeOnly could easily happen if they moved to another cell (but I'm not saying activeOnly should be false because that search is a whole lot more expensive). Also there are items in the starting cell owned by 'a shady smuggler' that don't actually exist.

akortunov's question is interesting, sure.

Capostrophic:
Setting activeOnly to true was an accident... but I guess the owner of an item the player tries to steal is in the same cell as that item or in a adjacent cell in case of, say, towns in most cases. 
As it works now, the engine will essentially try to find a dead reference of the unique owner in active cells or grab a random reference of not unique owner and learn if it's dead and increment the stolen items counter otherwise, if I understand what am I doing. ~~I don't~~

Capostrophic:
Really I should be checking the count of the references in refdata but in this case the counter won't get increased for alive owners outside the active cells.

scrawl:
That if test bothers me a bit. Isn't it the same as empty || !dead ?

scrawl:
> 

A nuance I didn't mention: for performance reasons, OpenMW will always consider owners that don't have a dead reference in the cells that were loaded at the time you were taking an item alive for the purpose of increasing the stolen item counter - this includes NPCs like "a shady smuggler" which don't have a reference in the world at all, and regular NPCs that happened not to be loaded while you were taking the item.

> <b>Coming to think of it, this includes actors which were cleared during corpse clearing.</b>

I'm not sure if that's so. Did you test? Deleted or 'cleared' objects can still be accessible. (Yes, very unintuitive, but that was required to make a broken-ass script for the gateway ghost quest work..)

Capostrophic:
Nope, I just took a wild guess, went through the corpse clearing code and it, um, looked that way. Well, good to know.

Capostrophic:
Just tested it. Unless I'm imagining things, it appears not to work when an item is stolen from a container owned by a dead NPC lying nearby, but it works fine otherwise.
...That's because I used item.getCellRef().getOwner() instead of ownerCellRef.getOwner(). Silly me.

scrawl:
I think the intention of show owned is to indicate whether taking is a crime, which it still is with this PR if I understood right. So we don't need to change anything else.

scrawl:
Could you rebase the commit that doesn't build (empty/isEmpty), and maybe a few others while already at it.

Capostrophic:
Done.

PR #1618, opened 2018-02-28
Adding max quicksaves to launcher

Thunderforge:
Adding the multiple quicksaves setting introduced in #1597 to the Launcher's Advanced page.

maximum quicksaves in launcher

I moved *Add "Time Played" to Saves* from the "Other" category to a newly formed "Saves" category.


Capostrophic:
max quicksaves is added to settings.cfg redundantly when the launcher is started even if it's not changed. Wouldn't it be possible to add an integer variant of load/saveSettingBool?

Thunderforge:
@Capostrophic Fixed in #1621.

PR #1619, opened 2018-03-01
Add unequip animation during stance switching (bug #4327)

akortunov:
Fixes [bug #4327](https://bugs.openmw.org/issues/4327) and [bug #2835](https://bugs.openmw.org/issues/2835).

Since CharacterController::updateWeaponState() has more than 500 lines of code, I am not able to test all possible cases for my own.
Also I noticed that there is no unequipping sound in Morrowind during weapon->spell switching. Should I add it?

Note: I am not sure if I should show torches and/or shields during stance change.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -749,6 +749,7 @@ void CharacterController::playRandomDeath(float startpoint)
 
 CharacterController::CharacterController(const MWWorld::Ptr &ptr, MWRender::Animation *anim)
     : mPtr(ptr)
+    , mWeapon(MWWorld::Ptr())
     , mAnimation(anim)
     , mIdleState(CharState_None)
     , mMovementState(CharState_None)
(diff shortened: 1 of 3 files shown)


apps/openmw/mwmechanics/character.cpp
```diff
@@ -1184,43 +1184,47 @@ bool CharacterController::updateWeaponState()
...
+            mAnimation->play(weapgroup, priorityWeapon,
+                              MWRender::Animation::BlendMask_All, true,
+                              1.0f, "unequip start", "unequip stop", 0.0f, 0);
+            mUpperBodyState = UpperCharState_UnEquipingWeap;
         }
-        else
+
+        float complete;
+        bool animPlaying = mAnimation->getInfo(mCurrentWeapon, &complete);
+        if (animPlaying && complete < 0.95f)
```
akortunov: I do not know how to play two animation groups one by one seamlessly, so I have to use a hack - play the equipping animation if more than 95% on unequipping animation was played.
Is there any better solution?

akortunov:
Also OpenMW misses unequip sound during weapon->hth switch.

akortunov:
I suppose this PR is ready for review/testing now.

scrawl:
> I do not know how to play two animation groups one by one seamlessly, so I have to use a hack - play the equipping animation if more than 95% on unequipping animation was played.
Is there any better solution?

That's an ongoing issue affecting many other animations, and for attack animations in particular we have an open bug.

In this PR, it seems that the lack of continuity actually causes a visible glitch rather than just delays/stuttering. The 95% number attempts to prevent this, but will not do so with lower frame rates. Maybe you could set the first animation to not auto-disable so that if there is a frame where the first animation is finished but the second is not started yet, we don't glitch to a position that's related to neither of the two animations. When that's done, the 95% 'hack' could be removed. 

I just noticed you already made a change to that. I'd still prefer a less hacky one, though.

akortunov:
> I'd still prefer a less hacky one, though.

Can you tell what's wrong? I'll try to rework it.

scrawl:
I do not like the 'Disable default idle state during weapon [un]equipping to avoid body desync', it looks like a hack and is placed in a totally unrelated section of the code.

My suggestion was to set the 'bool autodisable' of playAnimation to false. Not sure if it's practical, but worth a try.

akortunov:
> My suggestion was to set the 'bool autodisable' of playAnimation to false.

Done.
Note: hands position during unequipping on the run is a quite weird, but I suppose this is a resource bug.
Both OpenMW (with or without this PR) and Morrowind are affected by this bug.

Capostrophic:
Works well for me.

psi29a:
@zinnschlag can you cast your eye here. :)

apps/openmw/mwmechanics/character.cpp
```diff
@@ -1181,34 +1188,51 @@ bool CharacterController::updateWeaponState()
...
         {
-            if (!isWerewolf || mWeaponType != WeapType_Spell)
+            // Note: we do not disable unequipping animation automatically to avoid body desync
+            getWeaponGroup(mWeaponType, weapgroup);
+            mAnimation->play(weapgroup, priorityWeapon,
+                              MWRender::Animation::BlendMask_All, false,
+                              1.0f, "unequip start", "unequip stop", 0.0f, 0);
+            mUpperBodyState = UpperCharState_UnEquipingWeap;
+
+            if(!downSoundId.empty() && !isStillWeapon)
```
scrawl: !isStillWeapon is already checked above this.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -594,9 +594,9 @@ void getWeaponGroup(WeaponType weaptype, std::string &group)
 }
 
 
-MWWorld::ContainerStoreIterator getActiveWeapon(CreatureStats &stats, MWWorld::InventoryStore &inv, WeaponType *weaptype)
+MWWorld::ContainerStoreIterator getActiveWeapon(CreatureStats &stats, MWWorld::InventoryStore &inv, WeaponType *weaptype, bool getRealWeapon)
 {
-    if(stats.getDrawState() == DrawState_Spell)
+    if(!getRealWeapon && stats.getDrawState() == DrawState_Spell)
```
scrawl: The 'getRealWeapon' parameter seems a bit messy, is there any way to refactor the code so we don't need it?

apps/openmw/mwmechanics/character.hpp
```diff
@@ -152,6 +152,7 @@ struct WeaponInfo;
 class CharacterController : public MWRender::Animation::TextKeyListener
 {
     MWWorld::Ptr mPtr;
+    MWWorld::Ptr mPreviousWeapon;
```
scrawl: I think mWeapon would be a better name because this is the up to date weapon more often than it isn't.

akortunov:
I uploaded reworked version (changes, requested by scrawl).

An additional question: should we use weapon unequipping sound during weapon->spell switch?
It is missing in vanilla, but it should be here by common sence.

akortunov:
Are there any complaints or suggestions?

PR #1620, opened 2018-03-02
add more precise float-spinbox and use it for rotations

Florianjw:
Currently when editing instances with the CS, the precision of rotations is quite low, which can in some cases lead to visible gaps when entering them by hand.

This commit adds another type of edit that double the number of available decimals and uses them for rotations.

I should note that this is the change that appeared the least invasive to me, though I would consider the following a better solution: Instead of entering the rotations directly, involving the use of transcendental numbers and rounding, consider to normalize rotations to multiples of `2*pi`. That way a 180°-rotation would be written as 0.5, 90° would be 0.25, 45° 0.125 and so on. I believe that this would be a much more natural way of describing rotations. Alternatively degrees could be used, which are less beautifull from a mathematical perspective, but also allow things like 60° to be easy enough. The change for that would certainly be larger and users would have to change the way the use the CS, but I think in the long run it would be well worth it. I would be willing to implement it, if you are fine with that, but I cannot promise any timeframe (so you might want to merge this either way).

### DIFF
--- a/apps/opencs/model/world/columnbase.hpp
+++ b/apps/opencs/model/world/columnbase.hpp
@@ -88,6 +88,7 @@ namespace CSMWorld
             Display_UnsignedInteger8,
             Display_Integer,
             Display_Float,
+            Display_Double,
             Display_Var,
             Display_GmstVarType,
             Display_GlobalVarType,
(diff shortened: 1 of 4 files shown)


psi29a:
This is great, thank you! :)

PR #1622, opened 2018-03-06
Combine stacks of similar items when vendor sells items (bug #2842)

DanielVukelich:
A fix proposal for [bug #2842](https://bugs.openmw.org/issues/2842)

When merchants have an item in an owned container, and an identical item
on their person, those stacks do not appear combined in the trading
screen.  This is because the items in the container may not have their
ownership set on a per-item basis, but the items on the vendor's person
do.  When trying to combine stacks, the code would see that the owners
didn't match, and would not join them.

Since the code to build the container model for the trader already
checks that the items are owned by the vendor, it is safe for the
model to assume that if it is trading, then item ownership can be
ignored for the sake of stacking.

### DIFF
--- a/apps/openmw/mwgui/containeritemmodel.cpp
+++ b/apps/openmw/mwgui/containeritemmodel.cpp
@@ -18,23 +18,23 @@
 namespace
 {
 
-    bool stacks (const MWWorld::Ptr& left, const MWWorld::Ptr& right)
+    bool stacks (const MWWorld::Ptr& left, const MWWorld::Ptr& right, bool respectOwnership)
     {
         if (left == right)
             return true;
 
         // If one of the items is in an inventory and currently equipped, we need to check stacking both ways to be sure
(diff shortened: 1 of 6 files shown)


scrawl:
I'm gonna need to have a long look at this PR because it is touching historically very delicate code that likes to break when you just look at it the wrong way.

For starters though, I'm pretty sure this breaks the InventoryStore::stacks as it is no longer overriding the base method.

akortunov:
Do we need an ownership check in the GUI stacking at all? What it supposed to do? This check was added by scrawl as a part of large commit in 2014.

In the ContainerModel we do not need to respect ownders, since we track container owner instead of item owners. The same thing for pickpocketing model.
For inventory and companion windows there is no point to respect owners too.
Trading model is a wrapper for inventory one.

akortunov:
Closing this PR since if is incomplete, abandoned and buggy. We need a discussion first about how the stacking should work (including owners system).

PR #1623, opened 2018-03-06
Add testing options to the Settings page

Thunderforge:
[Implements #4345](https://bugs.openmw.org/issues/4345), providing a GUI for the three most commonly used [testing options](https://wiki.openmw.org/index.php?title=Testing#Starting_the_Game_with_Special_Arguments). A number of new users have indicated that command line parameters and editing files in the data folders are cumbersome. This provides a user-friendly alternative.

testing options

Additionally, the "Start default character at" field grays out if the "Skip menu and generate default character" setting is unchecked, meaning that there is no confusion over why characters aren't appearing at that location. Note that the cell itself itself is preserved.

start default character at disabled

And if the "Start default character at" field is blank, the words "default cell" appear as placeholder text to clarify what will happen.

start default character at default cell

The startup script field can be edited with either text or with the "Browse…" button. 

(Note that both the starting cell field and the startup script field are unfortunately affected by [Bug #3911](https://bugs.openmw.org/issues/3911), causing double characters when typing on macOS).

### DIFF
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -8,6 +8,7 @@
 #include <QMenu>
 #include <QSortFilterProxyModel>
 #include <thread>
+#include <mutex>
 
 #include <apps/launcher/utils/cellnameloader.hpp>
 #include <components/files/configurationmanager.hpp>
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -335,7 +335,7 @@ void Launcher::DataFilesPage::slotAddonDataChanged()
         // Loading cells for core Morrowind + Expansions takes about 0.2 seconds, which is enough to cause a
         // barely perceptible UI lag. Splitting into its own thread to alleviate that.
         std::thread loadCellsThread(&DataFilesPage::reloadCells, this, selectedFiles);
-        loadCellsThread.join();
+        loadCellsThread.detach();
     }
 }
 
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -356,5 +356,5 @@ void Launcher::DataFilesPage::reloadCells(QStringList selectedFiles)
     CellNameLoader cellNameLoader;
     QStringList cellNamesList = QStringList::fromSet(cellNameLoader.getCellNames(selectedFiles));
     std::sort(cellNamesList.begin(), cellNamesList.end());
-    emit signalSelectedFilesChanged(cellNamesList);
+    emit signalLoadedCellsChanged(cellNamesList);
 }
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -37,11 +37,14 @@ Launcher::DataFilesPage::DataFilesPage(Files::ConfigurationManager &cfg, Config:
 
     connect(mProfileDialog->lineEdit(), SIGNAL(textChanged(QString)),
             this, SLOT(updateOkButton(QString)));
-    connect(mSelector, SIGNAL(signalSelectedFilesChanged(QStringList)),
-            this, SLOT(slotSelectedFilesChanged(QStringList)));
 
     buildView();
     loadSettings();
+
+    // Connect signal and slot after the settings have been loaded. We only care about the user changing
+    // the addons and don't want to get signals of the system doing it during startup.
+    connect(mSelector, SIGNAL(signalAddonDataChanged(QModelIndex,QModelIndex)),
+            this, SLOT(slotAddonDataChanged()));
 }
 
 void Launcher::DataFilesPage::buildView()
@@ -322,7 +325,11 @@ bool Launcher::DataFilesPage::showDeleteMessageBox (const QString &text)
     return (msgBox.clickedButton() == deleteButton);
 }
 
-void Launcher::DataFilesPage::slotSelectedFilesChanged(QStringList selectedFilesChanged)
+void Launcher::DataFilesPage::slotAddonDataChanged()
 {
-    emit signalSelectedFilesChanged(selectedFilesChanged);
+    QStringList selectedFiles = selectedFilePaths();
+    if (previousSelectedFiles != selectedFiles) {
+        previousSelectedFiles = selectedFiles;
+        emit signalSelectedFilesChanged(selectedFiles);
+    }
 }
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -7,7 +7,9 @@
 #include <QCheckBox>
 #include <QMenu>
 #include <QSortFilterProxyModel>
+#include <thread>
 
+#include <apps/launcher/utils/cellnameloader.hpp>
 #include <components/files/configurationmanager.hpp>
 
 #include <components/contentselector/model/esmfile.hpp>
@@ -330,6 +332,17 @@ void Launcher::DataFilesPage::slotAddonDataChanged()
     QStringList selectedFiles = selectedFilePaths();
     if (previousSelectedFiles != selectedFiles) {
         previousSelectedFiles = selectedFiles;
-        emit signalSelectedFilesChanged(selectedFiles);
+        // Loading cells for core Morrowind + Expansions takes about 0.2 seconds, which is enough to cause a
+        // barely perceptible UI lag. Splitting into its own thread to alleviate that.
+        std::thread loadCellsThread(&DataFilesPage::reloadCells, this, selectedFiles);
+        loadCellsThread.join();
     }
+}
+
+void Launcher::DataFilesPage::reloadCells(QStringList selectedFiles)
+{
+    CellNameLoader cellNameLoader;
+    QStringList cellNamesList = QStringList::fromSet(cellNameLoader.getCellNames(selectedFiles));
+    std::sort(cellNamesList.begin(), cellNamesList.end());
+    emit signalSelectedFilesChanged(cellNamesList);
 }
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -142,6 +142,17 @@ void Launcher::DataFilesPage::saveSettings(const QString &profile)
     mGameSettings.setContentList(fileNames);
 }
 
+QStringList Launcher::DataFilesPage::selectedFilePaths()
+{
+    //retrieve the files selected for the profile
+    ContentSelectorModel::ContentFileList items = mSelector->selectedFiles();
+    QStringList filePaths;
+    foreach(const ContentSelectorModel::EsmFile *item, items) {
+        filePaths.append(item->filePath());
+    }
+    return filePaths;
+}
+
 void Launcher::DataFilesPage::removeProfile(const QString &profile)
 {
     mLauncherSettings.removeContentList(profile);
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -339,8 +339,16 @@ void Launcher::DataFilesPage::slotAddonDataChanged()
     }
 }
 
+// Mutex lock to run reloadCells synchronously.
+std::mutex _reloadCellsMutex;
+
 void Launcher::DataFilesPage::reloadCells(QStringList selectedFiles)
 {
+    // Use a mutex lock so that we can prevent two threads from executing the rest of this code at the same time
+    // Based on https://stackoverflow.com/a/5429695/531762
+    std::unique_lock<std::mutex> lock(_reloadCellsMutex);
+
+    // The following code will run only if there is not another thread currently running it
     CellNameLoader cellNameLoader;
     QStringList cellNamesList = QStringList::fromSet(cellNameLoader.getCellNames(selectedFiles));
     std::sort(cellNamesList.begin(), cellNamesList.end());
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -18,6 +18,7 @@
 
 #include <components/config/gamesettings.hpp>
 #include <components/config/launchersettings.hpp>
+#include <iostream>
 
 #include "utils/textinputdialog.hpp"
 #include "utils/profilescombobox.hpp"
@@ -47,6 +48,8 @@ Launcher::DataFilesPage::DataFilesPage(Files::ConfigurationManager &cfg, Config:
     // the addons and don't want to get signals of the system doing it during startup.
     connect(mSelector, SIGNAL(signalAddonDataChanged(QModelIndex,QModelIndex)),
             this, SLOT(slotAddonDataChanged()));
+    // Call manually to indicate all changes to addon data during startup.
+    slotAddonDataChanged();
 }
 
 void Launcher::DataFilesPage::buildView()
--- a/apps/launcher/datafilespage.cpp
+++ b/apps/launcher/datafilespage.cpp
@@ -37,6 +37,8 @@ Launcher::DataFilesPage::DataFilesPage(Files::ConfigurationManager &cfg, Config:
 
     connect(mProfileDialog->lineEdit(), SIGNAL(textChanged(QString)),
             this, SLOT(updateOkButton(QString)));
+    connect(mSelector, SIGNAL(signalSelectedFilesChanged(QStringList)),
+            this, SLOT(slotSelectedFilesChanged(QStringList)));
 
     buildView();
     loadSettings();
@@ -318,4 +320,9 @@ bool Launcher::DataFilesPage::showDeleteMessageBox (const QString &text)
     msgBox.exec();
 
     return (msgBox.clickedButton() == deleteButton);
+}
+
+void Launcher::DataFilesPage::slotSelectedFilesChanged(QStringList selectedFilesChanged)
+{
+    emit signalSelectedFilesChanged(selectedFilesChanged);
 }
--- a/apps/launcher/CMakeLists.txt
+++ b/apps/launcher/CMakeLists.txt
@@ -8,6 +8,7 @@ set(LAUNCHER
     settingspage.cpp
     advancedpage.cpp
 
+    utils/cellnameloader.cpp
     utils/profilescombobox.cpp
     utils/textinputdialog.cpp
     utils/lineedit.cpp
@@ -24,6 +25,7 @@ set(LAUNCHER_HEADER
     settingspage.hpp
     advancedpage.hpp
 
+    utils/cellnameloader.hpp
     utils/profilescombobox.hpp
     utils/textinputdialog.hpp
     utils/lineedit.hpp
@@ -39,6 +41,7 @@ set(LAUNCHER_HEADER_MOC
     settingspage.hpp
     advancedpage.hpp
 
+    utils/cellnameloader.hpp
     utils/textinputdialog.hpp
     utils/profilescombobox.hpp
     utils/lineedit.hpp
--- a/apps/launcher/advancedpage.cpp
+++ b/apps/launcher/advancedpage.cpp
@@ -1,7 +1,12 @@
 #include "advancedpage.hpp"
 
 #include <components/config/gamesettings.hpp>
+#include <components/config/launchersettings.hpp>
+#include <apps/launcher/utils/cellnameloader.hpp>
 #include <QFileDialog>
+#include <QCompleter>
+#include <components/contentselector/view/contentselector.hpp>
+#include <components/contentselector/model/esmfile.hpp>
 
 Launcher::AdvancedPage::AdvancedPage(Files::ConfigurationManager &cfg,
                                      Config::GameSettings &gameSettings,
@@ -15,6 +20,19 @@ Launcher::AdvancedPage::AdvancedPage(Files::ConfigurationManager &cfg,
     setupUi(this);
 
     loadSettings();
+}
+
+void Launcher::AdvancedPage::loadCellsForAutocomplete(QStringList filePaths) {
+    CellNameLoader cellNameLoader;
+    QStringList cellNamesList = QStringList::fromSet(cellNameLoader.getCellNames(filePaths));
+    std::sort(cellNamesList.begin(), cellNamesList.end());
+
+    // Set up an auto-completer for the "Start default character at" field
+    auto *completer = new QCompleter(cellNamesList);
+    completer->setCompletionMode(QCompleter::PopupCompletion);
+    completer->setCaseSensitivity(Qt::CaseSensitivity::CaseInsensitive);
+    startDefaultCharacterAtField->setCompleter(completer);
+
 }
 
 void Launcher::AdvancedPage::on_skipMenuCheckBox_stateChanged(int state) {
--- a/apps/launcher/advancedpage.cpp
+++ b/apps/launcher/advancedpage.cpp
@@ -5,6 +5,7 @@
 #include <apps/launcher/utils/cellnameloader.hpp>
 #include <QFileDialog>
 #include <QCompleter>
+#include <thread>
 #include <components/contentselector/view/contentselector.hpp>
 #include <components/contentselector/model/esmfile.hpp>
 
@@ -166,5 +167,8 @@ void Launcher::AdvancedPage::saveSettingBool(QCheckBox *checkbox, const std::str
 
 void Launcher::AdvancedPage::slotSelectedDataFilesChanged(QStringList selectedFiles)
 {
-    loadCellsForAutocomplete(selectedFiles);
+    // Loading cells for core Morrowind + Expansions takes about 0.2 seconds, which is enough to cause a
+    // barely perceptible UI lag. Splitting into its own thread to alleviate that.
+    std::thread loadCellsThread(&AdvancedPage::loadCellsForAutocomplete, this, selectedFiles);
+    loadCellsThread.join();
 }
--- a/apps/launcher/advancedpage.cpp
+++ b/apps/launcher/advancedpage.cpp
@@ -1,10 +1,14 @@
 #include "advancedpage.hpp"
 
-#include <components/files/configurationmanager.hpp>
+#include <components/config/gamesettings.hpp>
+#include <QFileDialog>
 
-Launcher::AdvancedPage::AdvancedPage(Files::ConfigurationManager &cfg, Settings::Manager &engineSettings, QWidget *parent)
+Launcher::AdvancedPage::AdvancedPage(Files::ConfigurationManager &cfg,
+                                     Config::GameSettings &gameSettings,
+                                     Settings::Manager &engineSettings, QWidget *parent)
         : QWidget(parent)
         , mCfgMgr(cfg)
+        , mGameSettings(gameSettings)
         , mEngineSettings(engineSettings)
 {
     setObjectName ("AdvancedPage");
@@ -13,8 +17,45 @@ Launcher::AdvancedPage::AdvancedPage(Files::ConfigurationManager &cfg, Settings:
     loadSettings();
 }
 
+void Launcher::AdvancedPage::on_skipMenuCheckBox_stateChanged(int state) {
+    startDefaultCharacterAtLabel->setEnabled(state == Qt::Checked);
+    startDefaultCharacterAtField->setEnabled(state == Qt::Checked);
+}
+
+void Launcher::AdvancedPage::on_runScriptAfterStartupBrowseButton_clicked()
+{
+    QString scriptFile = QFileDialog::getOpenFileName(
+            this,
+            QObject::tr("Select script file"),
+            QDir::currentPath(),
+            QString(tr("Text file (*.txt)")));
+
+
+    if (scriptFile.isEmpty())
+        return;
+
+    QFileInfo info(scriptFile);
+
+    if (!info.exists() || !info.isReadable())
+        return;
+
+    const QString path(QDir::toNativeSeparators(info.absoluteFilePath()));
+
+}
+
 bool Launcher::AdvancedPage::loadSettings()
 {
+    // Testing
+    bool skipMenu = mGameSettings.value("skip-menu").toInt() == 1;
+    if (skipMenu) {
+        skipMenuCheckBox->setCheckState(Qt::Checked);
+    }
+    startDefaultCharacterAtLabel->setEnabled(skipMenu);
+    startDefaultCharacterAtField->setEnabled(skipMenu);
+
+    startDefaultCharacterAtField->setText(mGameSettings.value("start"));
+    runScriptAfterStartupField->setText(mGameSettings.value("script-run"));
+
     // Game Settings
     loadSettingBool(canLootDuringDeathAnimationCheckBox, "can loot during death animation", "Game");
     loadSettingBool(followersAttackOnSightCheckBox, "followers attack on sight", "Game");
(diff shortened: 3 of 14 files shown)


scrawl:
Well, that looks nice, but I'm gonna have to play devil's advocate as usual.

I guess the main concern would be that this gets used for actual play as people discover 'yay, I can skip the stupid tutorial for my new characters' and only later discover that because the start script didn't run a few other things don't run, some characters are not disabled as they should, maybe some quests don't work, etc. IIRC that kind of stuff has been reported as bugs in the past, even while the 'skip-menu' option was still a cfg-only option. The 'Testing' label is a good start to communicate that this is, well, for testing. Maybe it should be under 'Advanced' as well with another disclaimer under it?

Personally, while I often use --start / --skip-menu, I don't think I would use the launcher to toggle it. For me its faster to edit the cfg file. The launcher wouldn't offer anything on top of that, as it can't detect beforehand if a cell exists or is correctly spelled, either.

Worth noting the 'Run OpenMW' option in OpenMW-CS is probably a good alternative to this feature if you're developing a mod, while at the same time being obviously just for testing and nothing more.

Thunderforge:
> The 'Testing' label is a good start to communicate that this is, well, for testing. Maybe it should be under 'Advanced' as well with another disclaimer under it?

@scrawl I had thought about putting it under the Advanced tab, but it was my understanding that that would be going away post-1.0 when we are able to add all those options as in-game settings. Obviously, we can't add this as an in-game setting, so putting it on a tab that would stay post-1.0 seemed like the better option. However, if you really want to put it in the Advanced menu, we can, but that probably will mean we need to talk about what to do post-1.0.

I agree that adding the "Testing" label does help. What if I added a quick message above the settings? Something like "These settings are intended for testing mods and may cause issues if used for normal gameplay."

> Personally, while I often use --start / --skip-menu, I don't think I would use the launcher to toggle it. For me its faster to edit the cfg file.

There's definitely demand for doing it in the Launcher (I added a few anecdotes from new users in the bug report). Still, I don't think there's a downside to having it available both ways.

> The launcher wouldn't offer anything on top of that, as it can't detect beforehand if a cell exists or is correctly spelled, either.

I'd thought about that and should this PR be merged, I'd like to look into if it's possible to have an auto-complete list of cells in the game.

> Worth noting the 'Run OpenMW' option in OpenMW-CS is probably a good alternative to this feature if you're developing a mod, while at the same time being obviously just for testing and nothing more.

True, that would be an alternative. My main use case has been testing new features, or downloading a mod and testing to see if it's compatible with OpenMW. Those wouldn't benefit from OpenMW-CS testing.

Thunderforge:
I've added a message to the Testing block like so:

testing message block

scrawl:
The current 'Settings' tab is not really for settings. A more accurate label may be 'Game Installation Setup'. It allows you to pick language, re-run installation wizard and re-import the INI. The new proposed testing options do not really fit here.

The disclaimer on the 'Advanced' page, IMO it's ill-worded, confusing and not very interesting anyway. Just remove it then. As a user, I couldn't give a crap if X will be Y in the future, I care about the current version. Even if those settings do move in-game at some point (which probably not all of them will, because some things are difficult or impossible to toggle once the game is running), I'm sure they will be easy enough to find without a prior announcement.

Btw, putting the Testing options on the Settings tab causes a new scrollbar to appear, whereas the Advanced tab already has a scrollbar anyway.

"and may cause issues" should be "and will cause issues" because --skip-menu most definitely will.

Thunderforge:
> The current 'Settings' tab is not really for settings. A more accurate label may be 'Game Installation Setup'. It allows you to pick language, re-run installation wizard and re-import the INI. The new proposed testing options do not really fit here.

@scrawl Fair enough. I can move it to the Advanced Page. But…

> The disclaimer on the 'Advanced' page, IMO it's ill-worded, confusing and not very interesting anyway. Just remove it then. As a user, I couldn't give a crap if X will be Y in the future, I care about the current version. 

The disclaimer phrase was [specifically requested](https://github.com/OpenMW/openmw/pull/1603#issuecomment-362899089) by @zinnschlag, who thought it pretty important to get across. While I'm fine removing it, I'm feeling uneasy about having conflicting instructions between the top two committers.

@zinnschlag If the nature of the Advanced page is changing to include things that can't be added in-game, such as testing options, are you okay with me removing the disclaimer?

> Btw, putting the Testing options on the Settings tab causes a new scrollbar to appear, whereas the Advanced tab already has a scrollbar anyway.

Actually, I added the scrollbar, but if I move it to Advanced, then I'll put it back to be scrollbar-less.

> "and may cause issues" should be "and will cause issues" because --skip-menu most definitely will.

I'll change that.

Thunderforge:
@scrawl Since Zini hasn't responded in a week, I've gone ahead and made the change. The Settings page has been restored. My new "Testing" options are now on the Advanced page, and the disclaimer on that page has been removed.

testing options on advanced page

zinnschlag:
Sorry about that. This PR completely escaped my attention. Was a very busy week for me indeed and apparently this PR drowned in the flood of OpenMW notifications I got. I will have a closer look at it later today.

zinnschlag:
Okay. I have read through the PR. I do not see a use case for this feature at all. Engine developers would certainly run OpenMW from the console (better access to log messages, quicker/less button pressing).

And a content developer would run OpenMW from within OpenMW-CS. Even when testing for compatibility of an existing mod with OpenMW that is the reasonable thing to do, because the first step of any test should be to run the verifier. Testing a specific piece of content from within OpenMW-CS is much more convenient anyway. Or it will be, once we have found a volunteer to create some default debug profiles (really surprised that this task has not been taken yet).

I don't think we can dismiss scrawl's devil's advocacy either. The launcher is meant as a tool for players. Not for engine or content developers. Mixing that up will result in confusion, no matter how we structure the launcher.

Thunderforge:
@zinnschlag There are two main groups of users that would benefit from this feature:

 - Content developers who are using the vanilla CS instead of OpenMW-CS.
 - Users testing mods to see if they are compatible with OpenMW

Users from the first group are the ones who made the statements quoted in the issue that prompted me to create this PR (they were Tamriel Rebuilt developers). As for mod testers from the second group, I don't think the reality is that mod testers run the verifier in OpenMW-CS. I know that for all the mods I tested, I sure didn't. 

It's all well and good to say that they _ought_ to go through OpenMW-CS, but the reality is that's not the workflow that people are using. (As an aside, there is [Bug #3932: OpenMW-CS: Debug -> Run OpenMW doesn't do anything](https://bugs.openmw.org/issues/3932), which I myself created after finding the OpenMW-CS workflow to be too confusing).

> Testing a specific piece of content from within OpenMW-CS is much more convenient anyway.

Judging from my own experience, and the reports from users on Reddit and Discord, there is some disagreement on that, especially with mods not originally created with OpenMW-CS.

> Engine developers would certainly run OpenMW from the console (better access to log messages, quicker/less button pressing).

Well, I have dabbled as an engine developer, so I'm an engine developer who would prefer this option. I use an IDE, so logs are already available, and there is no button pressing if I have this option set once.

> The launcher is meant as a tool for players. Not for engine or content developers.

What about players that want to test a mod real quick to see if they like it? For instance, if I as a player decide that I want to load up a Sadrith Mora graphics replacer mod, I would rather quickly start the game in Sadrith Mora to check out the graphics, rather than start a new game. If I like it, then I'll load up a real character. But for a quick "Do I like this mod?" check, this is something that's desirable.

And what about non-developers? Before I got a development environment set up and started submitting code to OpenMW, I started by just trying to reproduce bugs. For instance, I took some time to reproduce [Bug #3929: Leveled list merchant containers respawn on barter](https://bugs.openmw.org/issues/3929), and I wish that I would have had these Launcher options available rather than try to muck around with `openmw.cfg`.

>  Mixing that up will result in confusion, no matter how we structure the launcher.

I reached out to a TR developer and asked about if the Launcher provided more confusion. His response:

> Nope! No confusion at all! I just want to test my damn mod. All the other methods, especially whatever the hell they're saying to do with their CS, are *way* more confusing than just having a setting right there in the launcher. I already have my mod loaded, I change my mod a whole bunch in the regular CS, and I just want to pick which city I start in.

When I asked if he thought it would be confusing to players, he said:

> Why would players be confused by this if it clearly says it's for testing mods?

MortimerMcMire:
Hi there, I'm a TR dev who's been making a huge amount of scripting/quests/dialogue in the past few months. Some of those comments above are mine, and as the use case in mind I thought I'd share why this is a good feature.

I do all of my development in the original Morrowind Construction Set. Nothing personal against OpenMW-CS, it just isn't as fully featured yet for me to make the switch. The user options are buried in the documents folder (why?) and none of the startingcell options are printed in the default .cfg you're given, which means even figuring out how to use the feature was a labor in itself (thankfully Thunderforge helped me out and updated the documentation). 

I've tested none of the in-development TR release with OpenMW yet, and I suspect very few others have either. Hell, many people probably don't know StartingCell is an option. This feature puts you a leg up on vanilla and saves me many clicks. 

I also realized this lets you run a script after setup which is awesome, currently I have to run a dummy esp that adds default generic dialogue/joins guilds as a startscript.

OpenMW is a great engine with a lot of features, but if no one knows the features exist or knows how to use them then its frustrating.

AnyOldName3:
As an engine developer, I don't start the engine from the command line. If I wanted to start OpenMW in a specific cell, though, I'd probably:

* Put the setting in the VS debugger launch arguments (effectively the same as adding the setting to the command line).
* Put the setting in `openmw.cfg`, which I have quick and easy access to because I set up a link when I was testing something that required changing it a lot.

That means I wouldn't benefit from this feature.

However, I saw the original discussion with the TR developer who inspired this feature, and there were certain things that seemed immediately obvious to me and Thunderforge that were difficult/confusing/tedious for the TR guy. I doubt that he's an idiot, but things that obvious to people familiar with OpenMW and its guts aren't obvious to other people, and even if they know a feature exists, if it needs activating without using a GUI, there's room to break it.

zinnschlag:
First off the statement that OpenMW-CS isn't ready yet is of course correct.

However the way to deal with this issue is to work on improving OpenMW-CS, not adding stuff in semi-random places that compensates for the missing parts.

@Thunderforge: I closed 3932 since it is a duplicate.

> What about players that want to test a mod real quick to see if they like it? For instance, if I as a player decide that I want to load up a Sadrith Mora graphics replacer mod, I would rather quickly start the game in Sadrith Mora to check out the graphics, rather than start a new game. If I like it, then I'll load up a real character. But for a quick "Do I like this mod?" check, this is something that's desirable.

That might work for graphics replacers, but could break for many other things that depend on the player character being properly initialised.

@MortimerMcMire: Thanks for your input. I understand the points you are making. But essentially they all come down to OpenMW-CS isn't ready.

MortimerMcMire:
So I have to wait an indefinite amount of time for OpenMW-CS changes and then decide to make the switch to using it? Even if OpenMW-CS gets finished I don't think relying on people loading up all of their mods in OpenMW-CS and testing it through there is a good option for me. 

the only things that depend on character initialization are generic dialogue entries (background, latest rumors etc). I have my own esp that adds them, but the rest works great from what I've experienced.

I understand that you yourself don't see the value in this, but as a content developer I do, and ultimately this is for people like me who aren't willing to jump through a flaming hoop to figure it out.

lysol90:
While I understand (and would also) appreciate this PR, I get where Zini is coming from. When the CS is more finished, it makes so much sense to me to launch the game from there when testing a mod in development. Once we are actually using the OpenMW-CS to make mods, testing them via the launcher is going to be the slow and clunky way compared to launching directly in the OpenMW-CS.

In the meantime, this PR is very useful as a temporary solution of course. But then there's the risk of people getting mad when the feature gets removed.

mattwla:
Just wanted to ditto what Mort said from a content creator perspective. It would be nice to have a GUI with testing options obviously presented. On the other hand I get why people would rather keep all that in the CS, and have the launcher player focused.

 On the other other hand there is a not insignificant portion of the player base who never touches the CS, but is interested in the power to debug mods (I would say most my MW playing friends fall into this category). In which case a player-facing mod debug GUI sounds good.

Not really sure if I have contributed anything useful to the discussion. In the end the GUI is good and useful but I get why people would think it belongs elsewhere (but also get why having it in the launcher makes sense).

zinnschlag:
Hm, it seems we have two entangled discussions here. Let's try to untangle them:

1. OpenMW-CS isn't ready and until then people want a workaround with a GUI.

Sorry, but I don't agree with this position. All the functionality is there, even if using the command line/config files is necessary. That is a sufficient workaround until we get OpenMW-CS to 1.0.

2. There are some use-cases for having basic content debugging tools in the launcher even with a fully featured OpenMW-CS.

Personally I don't see these as significant. But that is just my personal opinion and I certainly wouldn't try to force it on other people.

However if we do add this feature as a permanent fixture, then it should be properly designed and not just a GUI front-end for some command line options.

I haven't given it much thought yet. But here are some things to consider:

* The debug settings (at least the start cell and the script) are specific to the content file list or at least to the omwgame file. If I were to switch to a different content list I would not want to use the same debug settings and in fact those settings from the other list could stop OpenMW from starting properly.
* If we have debug settings bound to content file lists then the right place for the GUI is the Data Files tab. Question is, can we design a suitable UI that is basically invisible for a user not interested in these features (i.e. it doesn't take any significant amount of screen-space and doesn't add visual clutter)?
* Is manually entering the cell a good idea? Maybe we should scan the content files for a cell list and present a drop-down menu? Remember that this also needs to take unnamed exterior cells into consideration.
* Can we document in the interface that the start script is not a proper script but just a list of instructions send to the console?

MortimerMcMire:
-vanilla morrowind behavior is to default to a balmora exterior cell (the location you end up in with 'coc balmora', not sure if theres any significance to that location) if the targeted cell isn't found

-manually entering a cell works fine, scanning content files for cell names sounds like overkill, just printing "cell not found, defaulting to balmora" (or cell 0,0, or whatever you decide) into the console would work.

zinnschlag:
Defaulting to Balmora is not an option, since there isn't necessary a Balmora. Remember, OpenMW is not just for running Morrowind.esm! But I just tested it. If the cell in the start argument is not found OpenMW defaults to the main menu with an error message. That is good enough.

Regarding manually entering vs list: Manual sounds a bit half-assed to me. Having a list would provide better usability, especially since we would get a basic search function without any additional work (part of the Qt drop down widget). But let's see what Thunderforge has to say about this topic, since he would be the one doing the work.

Thunderforge:
> Regarding manually entering vs list: Manual sounds a bit half-assed to me. Having a list would provide better usability, especially since we would get a basic search function without any additional work (part of the Qt drop down widget). But let's see what Thunderforge has to say about this topic, since he would be the one doing the work.

Scrawl had asked about this earlier in the thread. I would like to add it, but it felt like something more appropriate for a separate PR. If you'd rather have it now, I can look into implementing it as part of this PR.

zinnschlag:
Yes, please. Go ahead.

scrawl:
Probably you could make a new command like ```openmw --list-cells``` that the launcher calls in the background. But given that:
* You need to load the content files which takes a few seconds,
* Unnamed cells like "-4, -8" will have little context for me to know which one to choose,
* The editor already works much better for choosing cells as you can preview them or use the cell grid.
Is that added complexity really worth it?

MortimerMcMire:
I've never tried to start in an unnamed region exterior, only a town or
interior. From a user standpoint. I don't know if startingcell in openmw
handles exteriors yet. I remember getting some error when I tried that

Thunderforge:
> I've never tried to start in an unnamed region exterior, only a town or interior. From a user standpoint. I don't know if startingcell in openmw handles exteriors yet. I remember getting some error when I tried that.

OpenMW allows this since [Feature #3941](https://bugs.openmw.org/issues/3941) was implemented in August 2017. The reason I had requested it was that I was testing some stuff around the Shrine of Boethiah (-8, -4) and didn't have an easy way to start the game there.

scrawl:
To add to my list above:

OpenMW isn't a command line application. If you want to extend it to act like one you'll either be copy-pasting lots of code to make a completely new application, or have to substantially refactor OpenMW so that the required initialization routines can run without creating a game window. Or you could try adapting ```esmtool``` , i.e. add a mode that prints a list of cells given a list of content files.

If you really want to do this, I'd try the esmtool route first. I'm still not convinced if all of this is worth it, though.

Artemex:
Why not add bsa support to the launcher?

zinnschlag:
That would be more placeholder stuff that then needs to be replaced later. We are going to handle resource quite differently after 1.0.

scrawl:
> That would be more placeholder stuff that then needs to be replaced later. We are going to handle resource quite differently after 1.0.

Morrowind files will always remain Morrowind files, though. Or do you have another plan to handle those better?

kcat:
I think the `data` lines would be more important for the launcher, and the BSAs are a special-case of that. I'd actually consider handling BSAs as a `data` source (e.g. `data=/path/to/morrowind.bsa`). Or perhaps auto-load BSAs when their related content file is loaded (e.g. `content=morrowind.esm` would implicitly load `morrowind.bsa` for the same directory; BSAs would always be lower priority than `data` directories, and would be ordered according to their plugin order).

zinnschlag:
@Thunderforge I am having a hard time to come to a decision here, because there are equally good arguments for and against accepting this PR. Maybe @scrawl can weight in on that too?

But before we throw the towel, may I ask what made the task so difficult? Doesn't sound that bad too me. The way I would have done it:

- Hook a signal into opening the dropdown menu
- Check if you already have a list of cells and if the content file list hasn't changed
- If yes, do nothing, otherwise:
- Run a small specialised loader that skips over all records that aren't cells and build a new list from that.

Thunderforge:
@zinnschlag Sorry that I didn't give enough information, and I do appreciate your faith in being able to figure it out. I guess I was figuring that I didn't want to put a lot of work into something that might not make it. Still, maybe we can figure this out with a bit of help.

> *  Run a small specialised loader that skips over all records that aren't cells and build a new list from that.

This is the part that I found difficult, rather than the UI parts about the dropdown. From what I figured out, I had to create a new `CSMWorld::Data::Data` object, figure out how to load each file, and then call `getCells()`. This is obviously very inefficient, and that was my concern over the mods changing; that it would take several seconds to reload everything. I briefly thought about creating my own `ESM::ESMReader` and calling `open()` on each file, but I got stuck with how to actually parse the records and only keep the cell names.

Am I on the right track at least? Or is there a much easier way that I'm not seeing?

Thunderforge:
@zinnschlag Thanks, that was the hint I needed. My solution so far is this:

```c++
ESM::ESMReader esmReader;
esmReader.open(filePath);
ESM::Cell cell;

while(esmReader.hasMoreRecs())
{
        ESM::NAME n = esmReader.getRecName();
        esmReader.getRecHeader();
        if (n.intval == REC_CELL) {
              bool isDeleted = false;
              cell.loadNameAndData(esmReader, isDeleted);
              cellNamesSet.insert(cell.mName);
        }
}
```

Now it's just a matter of tying that in to the rest of the code. I'm a C++ novice (I usually work in Java), so I can't guarantee the code will be pretty, but I think I've gotten past the bottleneck and can get a solution made.

Thunderforge:
Alright! I got it to autocomplete names for cells!

autocomplete demo

The only catch is that I'm stuck with how to actually get the file names for all the content. The above screenshot is generated when I hardcode the full pathname of some content.

```c++
esmReader.open("~/Library/Application Support/Steam/SteamApps/common/The Elder Scrolls III - Morrowind/Data Files/Morrowind.esm");
```

Of course, I can't hardcode file paths, I have to do it programmatically. I found `mGameSettings.getContentList()`, but that just returns the names of the files (e.g. `Morrowind.esm`), rather than the whole path. Unfortunately, calling `esmReader.open("Morrowind.esm")` throws an exception.

Is there any way that I can get the full pathnames of all loaded content?

zinnschlag:
The launcher should have an instance of ContentSelectorView::ContentSelector somewhere. This class has a selectedFiles function, which returns a collection of EsmFiles, which have a filePath function.

Thunderforge:
Autocomplete is now working. `ContentSelectorView::ContentSelector` was really tightly coupled with the `DataFiles` class, so it wasn't practical to use. Instead, I created a new `selectedFilePaths()` method in `DataFiles` and just did it that way.

Right now, it only loads the files checked on startup. I'm guessing that a signal/slot system is the best way to update that? Unfortunately, it looks like I have to write my own logic for that because there isn't any existing code to respond to checkbox changes.

lysol90:
Nice work!

zinnschlag:
> Right now, it only loads the files checked on startup. I'm guessing that a signal/slot system is the best way to update that? Unfortunately, it looks like I have to write my own logic for that because there isn't any existing code to respond to checkbox changes.

I would look at the event function and see if you can hook into that. With Qt if there isn't a suitable signal, then most of the time an event is the way to go.

scrawl:
How long does it take to scan all files? If it's noticable, we should move that to a background thread. Note that to test the worst case, you have to test with your filesystem cache cleared (e.g. reboot your PC first).

Actually, a nice side effect of this PR could be that the launcher will load the content files into the OS's cache, which will speed up starting the game once you hit Play.

zinnschlag:
Good point. I am very interested in the performance results. @Thunderforge Please also post your hardware specs so we have something to compare to.

> Actually, a nice side effect of this PR could be that the launcher will load the content files into the OS's cache, which will speed up starting the game once you hit Play.

If it is implemented as I suggested (only scan when opening the menu) then this would not happen in most use-cases. But lets wait for the performance results. If the performance is bad and the function needs to be moved to a separate thread then we probably should scan at startup instead.

Thunderforge:
I've finally gotten a chance to work on this PR again. I've gotten to where the autocomplete list of cells can be dynamically changed, but I can't figure out how to tie it in with the user checking or unchecking a mod.

To see the dynamic change in action:

1. Go to the Data Files page
1. Enable or disable a mod (e.g. Bloodmoon)
1. Right-click on a mod to bring up the context menu (this is my workaround to get a hook to the data model being changed)
1. Go to the Advanced page
1. Search for a cell that exists in that mod (e.g. "Fort Frostmoth")

If the mod is enabled, autocomplete will see it. If the mod is disabled, autocomplete won't see it.

Now I need some help. How do I hook into the user eanbling or disabling the mod? I was hoping there would be some sort of `slotAddonChanged` or something, but I couldn't find anything like that. Is there something obvious that I'm missing?

AnyOldName3:
It's possible that there isn't a slot and signal because whoever implemented it in the first place didn't think we'd ever need to know about that type of change anywhere else. I'd try and track down what gets run when the checkbox is clicked for a plugin and if it doesn't emit any signals, create a new one.

Ferk:
It's possible I'm entirely missing the point here... but for the case of a player wanting to test if a mod works:

Wouldn't it make more sense if rather than having options in the launcher, there were userfriendly console commands you can input as soon as the game starts to quickly teleport to whatever cell to do the testing or to set the character to whichever state is convenient for it?
It would be convenient in general for teleporting while in game being able to auto-complete the name of the cell. No idea if this type of autocompletion is supported by the GUI framework within the engine, but it would be handy.

AnyOldName3:
@Ferk We already have all that.

Ferk:
@AnyOldName3 Really? there's already an in-game GUI for it?
Doesn't CenterOnCell require you to provide the exact Cell id?

zinnschlag:
The Model of the content selector is based on QAbstractTableModel. It has a dataChanged slot.

Thunderforge:
> The Model of the content selector is based on QAbstractTableModel. It has a dataChanged slot.

Thanks @zinnschlag, that was exactly the piece I needed!

The current code now correctly updates the autocomplete list with every checkbox (and only does so when the selected files have changed, as opposed to some other type of data change). The only drawback is there is a barely perceptible sluggishness when you check a box; probably the 0.2 second file load [previously reported](https://github.com/OpenMW/openmw/pull/1623#issuecomment-380314283). I'll look into using a thread to alleviate that.

Thunderforge:
Resetting the autocomplete is now delegated to a new thread (thanks to @AnyOldName3 for helping me get that part figured out), and there is no longer a lag in the UI. This involved changing where the cell loading was happening so that we could still update the Qt components on the main thread.

I believe the feature is now complete. The testing options are available and the "Start default character at" field now autocompletes to whatever cells are currently loaded with the selected addons. It updates whenever the user changes their addons. Please let me know if there are any issues with the code that need to be addressed before it is merged.

Thunderforge:
@zinnschlag Do you have any additional requests for this PR? If not, could it be merged in time to be part of 0.44?

psi29a:
@zinnschlag what do you think?
I'm OK with this.

zinnschlag:
Yeah, most likely. Want to do one final review and a test-run. Probably tomorrow.

apps/launcher/datafilespage.cpp
```diff
@@ -308,3 +326,23 @@ bool Launcher::DataFilesPage::showDeleteMessageBox (const QString &text)
...
+
+void Launcher::DataFilesPage::slotAddonDataChanged()
+{
+    QStringList selectedFiles = selectedFilePaths();
+    if (previousSelectedFiles != selectedFiles) {
+        previousSelectedFiles = selectedFiles;
+        // Loading cells for core Morrowind + Expansions takes about 0.2 seconds, which is enough to cause a
+        // barely perceptible UI lag. Splitting into its own thread to alleviate that.
+        std::thread loadCellsThread(&DataFilesPage::reloadCells, this, selectedFiles);
+        loadCellsThread.join();
```
MiroslavR: I fail to see how this is any different than simply calling `reloadCells(selectedFiles);`. You spawn a thread and wait until it finishes while still blocking the UI...

Thunderforge: The intention was definitely for `reloadCells(selectedFiles()` to be called in a separate thread while not blocking UI. Is there a better way that I could do this so that UI is not blocked?

MiroslavR: You can remove the `join` call and the main thread won't be blocked.

However, I see a potential problem with `reloadCells` running in the background. If I understand correctly, if the user changes mods multiple times in quick succession, multiple threads will run simultanenously and there is no telling in which order the `signalSelectedFilesChanged` signal will be emitted. Therefore the list of cells might end up in an inconsistent state.

Thunderforge: @MiroslavR I tried removing `join`, but got `SIGABRT (Signal 6)` and the app crashed. In looking through C++ documentation, I'm guessing that instead I want to replace `join` with [`detatch`](http://www.cplusplus.com/reference/thread/thread/detach/). Is that correct?

Good point about the race condition with regards to results. I couldn't trigger one on my own, but with a large number of mods, it's certainly possible if a big one (e.g. Tamriel Rebuilt) is enabled or disabled in quick succession. Is implementing a queue of some sort the best way to handle this, or is there a cool C++ feature that I can use that would be better?

MiroslavR: > @MiroslavR I tried removing join, but got SIGABRT (Signal 6) and the app crashed. In looking through C++ documentation, I'm guessing that instead I want to replace join with detatch. Is that correct?

I edited my post.

> Is implementing a queue of some sort the best way to handle this, or is there a cool C++ feature that I can use that would be better?

Honestly I'm not sure what the best approach would be. I suppose you could also disable content selection for the duration of the cell loading thread. Or get rid of threading altogether if the delay is barely noticeable. Maybe @zinnschlag or @scrawl could chime in?

Thunderforge: I replaced `join` with `detach` so that the UI will not be blocked.

zinnschlag:
Ouch! Good catch, but ouch! Merging this into 0.44 looks unlikely now. There are issues that need to be addressed first.

The problem of multiple running threads finishing in an unpredictable order needs to be addressed. Also, you are emitting a signal from a different thread. Qt-signals are not per se thread-safe. They can be configured to work across thread boundaries and if you send them from a different QT thread this is done automatically (or rather if you connect a signal and a slot across threads it is done automatically). But I am not sure if that work with std threads too.

As with all things multi-threading, if you are looking at it and don't feel a sense of awe and fear, you haven't understood it properly. The simplest fix I can think of is to have a boolean variable (in the main thread) that is set to true when the worker thread is started and set to false by the slot handling the signal. Don't start the thread if it is set to true. That could theoretically miss some updates, but the chances of it breaking anything should be minimal.

Thunderforge:
In Java world, there is a [`synchronized { }` keyword](https://javamex.com/tutorials/synchronization_concurrency_synchronized1.shtml) that forces a block to run synchronously. After some searching, I found [an equivalent with mutexes](https://stackoverflow.com/a/5429695/531762) in C++11.

The order of operations should now be:

 1. User clicks checkbox
 1. Through signals and slots, eventually we get to `std::thread loadCellsThread(&DataFilesPage::reloadCells, this, selectedFiles);  loadCellsThread.detach();`, causing the `reloadCells(selectedFiles)` method to be called on a new thread, no longer blocking UI.
 1. The `reloadCells(selectedFiles)` method effectively executes synchronously, meaning that only one thread can be running it at once and subsequent threads have to wait their turn.
 1. At the end of `reloadCells(selectedFiles)`, a signal is emitted to change the autocomplete settings, which is picked up by the main thread.
 1. When `reloadCells(selectedFiles)` is complete, the mutex lock is released and the next thread, if any, can run its own `reloadCells(selectedFiles)`.

I tried some console logging and a whole lot of rapid clicking and unclicking, both with a keyboard and mouse, and after nearly two hundred attempts could not produce a race condition. Therefore, I believe that the race condition concern is fixed.

Thunderforge:
@zinnschlag Glad that you agree my fix addresses the race condition situation. 

Could you please clarify what the issue is with the signal-slot problem? I had thought that you were talking about the race condition I just fixed, so I'mnot sure I understand what you are referring to or why it would be a problem.

AnyOldName3:
Zini's point was that in `reloadCells` you're emitting a signal despite not being in the same thread as the rest of Qt's event handling. If any slots connected to it aren't re-entrant, this has the potential to break things as, for example, Qt might start redrawing the same part of the UI from two different threads. You need to do something to ensure that any slots connected to the signal you emit are executed on the main worker thread instead of your temporary thread.

Thunderforge:
@AnyOldName3 This is my understanding of the issue you are presenting.

 1. In the main thread, I create a threads A and B to load cells
 1. Thread A starts work
 1. Thread B waits until Thread A is finished
 1. Thread A completes and emits a signal
 1. Thread B starts work
 1. Thread B completes and emits a signal
 1. The corresponding slot receives the event and continues execution

Which thread does the slot pick up the signal? My understanding of the issue is that you are saying that threads A or B will receive the event and run the slot code. 

If so, that doesn't seem to happen: the main thread always picks up the slot. I added some couts to print [the hash of the thread](https://stackoverflow.com/a/15813873/531762) and got the following output:

```
Main thread is 5131613941657782972
Loading cells from thread 14972012352158077675
Emmitting signal from thread 14972012352158077675
Slot thread is 513161394165778297
```

Besides, Qt UI elements must be updated on the main thread. If another thread tries to update them, you get an exception (probably to avoid the very issue that you were describing). Since UI elements are being updated, it must be on the main thread, which is consistent with what I've described above.

zinnschlag:
Odd. It shouldn't work that way. Maybe Qt somehow manages to detect that the signal moves across the thread barrier and configures the connection correctly? I find it disturbing, if something works and you don't know why; especially if multi-threading is involved.
No idea how to figure it out, so I guess we have to roll with it. Just to make sure, please flag the connection for that cross-thread-boundary signal as Qt::QueuedConnection and then we can merge.

psi29a:
^-- threading is great until things work and you don't know why... you should then feel a cold sweat...

zinnschlag:
Just to make sure there is no miscommunication: the connection should still be explicitly flagged as queued. I don't want to depend on Qt detecting a cross thread connection automatically in this case. Peace of mind and all.

Thunderforge:
@zinnschlag I've explicitly declared the signal/slot relationship and as a `Qt::QueuedConnection`.

psi29a:
Aside from a changelog entry, I think it's ready. :)

Thunderforge:
@psi29a Are you referring to [this file](https://github.com/OpenMW/openmw/blob/master/CHANGELOG.md)? I didn't know that I needed to add to it. Now that it's merged, do I need to create a separate PR to modify it?

psi29a:
Yeah, I think Zini jumped the gun but that's fine. Yeah, create a new PR that references this one and we'll merge it right in.  Can you do the same for https://github.com/OpenMW/openmw/pull/1744 ?

Two changelog entries in one PR I mean.

Thunderforge:
@psi29a I'll do that from now on. I've created PR #1745 to add this feature to `CHANGELOG.md`. Looks like Zini already got PR #1744 with [this commit](https://github.com/OpenMW/openmw/commit/dbc87e7c7dc9d7adebda34d0458b972a682038ec#diff-4ac32a78649ca5bdd8e0ba38b7006a1e)

PR #1629, opened 2018-03-12
Use middle gray fallback color instead of black in case a fallback color wasn't found (bug #2841)

Capostrophic:
[Bug report](http://bugs.openmw.org/issues/2841)

Apparently there *was* fallback for cases the fallback wasn't defined... which resulted in the behavior the reporter experienced, making the world uniformly black with some light source influence if the whole weather sections of Morrowind.ini were missing. I changed the color to middle gray to make things a) visible b) look not worse than they were with pure black lighting and c) look generic. This also means that a text which uses a missing fallback setting will be gray instead of black.

I've also added a warning in case a fallback value wasn't found.

### DIFF
--- a/components/fallback/fallback.cpp
+++ b/components/fallback/fallback.cpp
@@ -1,5 +1,7 @@
 #include "fallback.hpp"
 
+#include <iostream>
+
 #include <boost/lexical_cast.hpp>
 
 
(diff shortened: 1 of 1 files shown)


scrawl:
Usually I like default colors that are 'in your face' so you know that something is wrong. For example, textures that are not found will show up pink.

With that said, the previous default (black) is not really 'in your face' either, so this change isn't much of a change. I'll just merge it.

PR #1632, opened 2018-03-13
Fix memory leak

elsid:
There is no delete for TextFormat objects in PageDisplay destructor.


psi29a:
@scrawl what do you think?

PR #1634, opened 2018-03-17
Rework torch usage for NPCs

akortunov:
Fixes [bug #4334](https://bugs.openmw.org/issues/4334).

Summary of changes:
1. NPC will hide light sources during bad weather, as they do in vanilla game.
The only difference here - in vanilla this feature does not work well during weather transitions, so I decided to check if there are precipitation particles in scene (rain, snow, etc.).
2. Since we do not want to unequip shields/torches when using two-handed weapons, there is no point to keep such limitation for NPCs. They will not unequp their weapons when they equip torches at night.

### DIFF
--- a/apps/openmw/mwworld/weather.cpp
+++ b/apps/openmw/mwworld/weather.cpp
@@ -520,6 +520,7 @@ WeatherManager::WeatherManager(MWRender::RenderingManager& rendering, const Fall
     , mSecunda("Secunda", fallback)
     , mWindSpeed(0.f)
     , mIsStorm(false)
+    , mPrecipitation(false)
     , mStormDirection(0,1,0)
     , mCurrentRegion()
     , mTimePassed(0)
@@ -660,6 +661,10 @@ void WeatherManager::update(float duration, bool paused)
     mWindSpeed = mResult.mWindSpeed;
     mIsStorm = mResult.mIsStorm;
 
+    // For some reason Ash Storm is not considered as a precipitation weather in game
+    mPrecipitation = !(mResult.mParticleEffect.empty() && mResult.mRainEffect.empty())
+                                    && mResult.mParticleEffect != "meshes\\ashcloud.nif";
+
     if (mIsStorm)
     {
         osg::Vec3f playerPos (player.getRefData().getPosition().asVec3());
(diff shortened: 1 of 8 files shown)


apps/openmw/mwworld/weather.cpp
```diff
@@ -660,6 +660,10 @@ void WeatherManager::update(float duration, bool paused)
     mWindSpeed = mResult.mWindSpeed;
     mIsStorm = mResult.mIsStorm;
 
+    // For some reason Ash Storm is not considered as a precipitation weather in game
```
Capostrophic: Well, there's not much difference between a blight storm and an ash storm mechanically, so why even have a special case for ash storms - merely for replicating the original perhaps buggy behavior?

akortunov: > merely for replicating the original perhaps buggy behavior?

Yes. But I do not know if it is buggy or intended behaviour, so I decided to replicate it for now.
If we do not want this case, we just can remove it.

scrawl:
Given that the only use of ```isDark()``` and ```isPrecipitation()``` is for checking if torches are required, can you consolidate them into one function?

akortunov:
> Given that the only use of isDark() and isPrecipitation() is for checking if torches are required, can you consolidate them into one function?

Sure.

scrawl:
Ok that's better. Just a minor suggestion for WeatherManager::useTorches. You could reduce the amount of typing and dependencies a bit. exterior/quasiExterior can be checked in World::useTorches (has nothing to do with the weather, anyway). Then get the timestamp in World, and change the other function to ```WeatherManager::useTorches(float hour)```. And voila, the WeatherManager won't need to know anything about World, at least not in this instance.

scrawl:
```mPrecipitation``` should be initialized in the constructor.

akortunov:
> And voila, the WeatherManager won't need to know anything about World

Weather manager uses the World at least in the playerTeleported() and update(float duration, bool paused). Alter these methods too?

scrawl:
Ok sure, let's do that while we're at it.

akortunov:
Done. A lot of changes to eliminate one line...

PR #1635, opened 2018-03-17
Update magic effect particles after building new animation

akortunov:
Fixes [bug #2254](https://bugs.openmw.org/issues/2254).
To reapply effects, I should make sure that the target actor already has the animation object.
Also active effects are stored in the CreatureStats, so this approach will not work for non-actor objects.

Note: I restore only looping particles effects (e.g. elemental shields and levitation).

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -491,6 +491,8 @@ namespace MWBase
             virtual void launchProjectile (MWWorld::Ptr actor, MWWorld::ConstPtr projectile,
                                            const osg::Vec3f& worldPos, const osg::Quat& orient, MWWorld::Ptr bow, float speed, float attackStrength) = 0;
 
+            virtual void applyLoopingParticles(const MWWorld::Ptr& ptr) = 0;
+
             virtual const std::vector<std::string>& getContentFiles() const = 0;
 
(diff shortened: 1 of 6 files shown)


scrawl:
I don't like that there are two nearly-identical updateParticles() methods. Can you try using visitEffectSources()?

scrawl:
And what about spells from InventoryStore - you could have e.g. an enchanted item with permanent shield.

akortunov:
> Can you try using visitEffectSources()?

Which of them?

akortunov:
Also visitEffectSources() does not contain the pointer to target actor, so I can not apply particles.

akortunov:
> And what about spells from InventoryStore - you could have e.g. an enchanted item with permanent shield.

I tested with Ebony Mail, and it worked.

scrawl:
> Also visitEffectSources() does not contain the pointer to target actor, so I can not apply particles.

Yes you can. You make a new class that contains the visit function and give that class the actor as a member variable.

Doing something for every effect is precisely what the visitor pattern is meant for; seems pretty obvious to use it here. You can see many examples in the source tree.

akortunov:
Done.
The only issue left is calling visitor in three different places.
I need to call it somewhere after creation of animation, because I need to apply particles to it.
Now I do this after we add an NPC, creature or player to scene.

PR #1636, opened 2018-03-18
Restrict opening doors to bipedal actors (Bug #4313)

Capostrophic:
Non-bipedal creatures will attempt to evade the door.

#1572 in its current state breaks this.


akortunov:
> #1572 in its current state breaks this.

What do you mean?

Capostrophic:
Well, added openDoors call in a pathTo doesn't care if the actor is bipedal or not. openDoors code shouldn't be triggered for non-bipedal actors.

akortunov:
So I will just move this check in another place, if your PR will be merged.

PR #1639, opened 2018-03-21
[discussion] Restore projectiles from game world

akortunov:
Related [forum thread](https://forum.openmw.org/viewtopic.php?f=6&t=5022). Of course, this feature has limitations, but it would be nice to use it.

The basic idea: rewrite the moveProjectiles() function to:
1. Use mesh dimensions instead of collision meshes for projectile collisions, we already do similar thing for magic bolts.
2. If the projectile hit a non-actor object, spawn an item in hit position and move it a bit.

screenshot_20180321_095417

### DIFF
--- a/apps/openmw/mwmechanics/combat.cpp
+++ b/apps/openmw/mwmechanics/combat.cpp
@@ -4,6 +4,8 @@
 
 #include <components/sceneutil/positionattitudetransform.hpp>
 
+#include <components/settings/settings.hpp>
+
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
 #include "../mwbase/mechanicsmanager.hpp"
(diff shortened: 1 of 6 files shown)


scrawl:
Interesting idea. I suppose you could, but there could be unexpected problems with just adding to random nodes like that, for example if the node is shared. Also you need to handle skinned meshes specially, figure out which is the affected bone and attach to that instead of the skinned mesh's parent.

akortunov:
Any suggestions how to improve this feature?

zinnschlag:
Given this some thought. IMO this is a post 1.0 feature. Projectile handling should be controlled by the content, not by user settings. A mod developer may want to bind the chance of having a recoverable projectile to a skill or limit it to specific projectile types or control it in some other way.
As mentioned this feature can affect the balance and there is at least the possibility that it could break some particular content. That is not something for the user settings and definitely not for 1.0.
We may want to re-use this branch at a (significant) later time. The question is what do we want to do with this PR? Either we mark it as post-1.0 or we close it.

akortunov:
I suppose it will be better to close it to keep PR's list cleaner. I will keep the branch, so anyone interested can use it.

psi29a:
Added [post-1.0] so we can better track it, that it wasn't just closed because of rejection but for review later. Good idea?

akortunov:
> Good idea for other such PR/MRs?

Looks good for me.

PR #1640, opened 2018-03-23
Typofix: updateCrimePersuit -> updateCrimePursuit

unelsson:
typing error in mwmechanix/actors.cpp and mwmechanix/actors.hpp
fixed updateCrimePersuit -> updateCrimePursuit

My first pull request. Am I doing this right?


psi29a:
Nice find!

PR #1644, opened 2018-03-24
Implement sBookSkillMessage GMST

Capostrophic:
sBookSkillMessage is displayed when you read a skill book, in the same messagebox as "skill increased" message. It was previously unused.


scrawl:
If we use GMSTs that Morrowind doesn't use, there is a chance that in some versions of Morrowind translators haven't bothered to translate that text because it was, well, unused. Also we could stumble on new 'missing in unpatched MW 1.0' GMST where the GMST is totally missing and either shows nothing or crashes the game if the code is not equipped to handle the missing GMST.

Capostrophic:
Valid points. BookSkillMessage is used in the original game for sure, though. Could I leave it?

PR #1646, opened 2018-03-25
[WIP] Consider actor dead after their death animation is finished (bug #4291 and #4307)

Capostrophic:
I feel as if something different is broken, but I dunno what is. This seems to work fine.


Capostrophic:
Yep, something different is broken and this breaks it further. Will investigate more.

Capostrophic:
Right, so OpenMW considers guys with zero health automatically dead. I need to disable this and move the rest of death events processing to the end of the death animation somehow not breaking persistent corpses.

Capostrophic:
Thankfully mTimeofDeath is solely in creaturestats. What is it even used for?

Capostrophic:
Scrapping this branch, my local experiments don't get here anyway. Will submit another pull request later (probably much later).

PR #1647, opened 2018-03-26
Improve the logic of trading formulae (bug #2222)

Capostrophic:
This is my proposed solution for the infamous bug. buyTerm is no longer used in sellingPrice calculations, instead the merchant no longer offers a price lower than the 3/4 of the base price for an item they sell or a price higher than the 3/4 of the base price for an item they buy, though the player can still haggle them to do that.
Haggling was also modified, to fix PC stats that are much lower than NPC stats presenting an increase in the chance of winning the haggle roll.


scrawl:
I would plot some graphs for the buying and selling price modifiers of the old function vs. the new one. If they are similar enough, we could see the change as non-controversial.

Also, would be nice to now what is the fix that MCP went for.

Capostrophic:
default
x axis is pcTerm; values of buyTerm and sellTerm are calculated based on an npcTerm of 100. It appears that with the lower limit of 1 in an attempt to prevent the "sell above buying price" exploit (which the original formula also was preventing in a way) the buyTerm bumps into it too early and the buying price is too high. However, if the lower limit of buyTerm were 0.75 and the upper limit of sellTerm were 0.75, higher Mercantile and Fatigue would no longer present an advantage.

Capostrophic:
From my testing, it looks like MCP simply removes the vanilla engine equivalent of "std::min(buyTerm, sellTerm)" and it's possible to sell an item for more than it costs to buy, so replicating MCP is a no-go in my opinion if we don't want to introduce the exploit the original behavior dealt with.

akortunov:
But what about haggling (trading.cpp)? Judging by [code](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/trading.cpp#L59), it is affected too. There should be something like `int(buying ? npcTerm - pcTerm : pcTerm - npcTerm)` instead.

Capostrophic:
Yes, there are issues with using an absolute value of (pcTerm - npcTerm) which increases the chance to win the haggling roll when using negative values. The researched haggling mechanics seem to not be doing something they were intended to do and OpenMW is just replicating that odd behavior in the best way it can.
AFAIK there's already an early-out there.
I remember having reviewed that change in your trading patch but discarding it. I'll give this part another look, but I mainly wanted to have some feedback on offered price changes specifically.
I wonder if fDispositionMod should also be used for offered price calculations.

Capostrophic:
Alright, if I got my maths right... 
x is the roll winning chance. fBargainOfferMulti (by default is -4) changes the winning chance according to the delta of offered prices, fBargainOfferBase (50%) is the base chance.
pcTerm-npcTerm is the difference of PC and NPC stats.
When the absolute delta of the stats is used, it never lowers the winning chance if the player stats are very low compared to NPC stats, on the contrary, it increases the chance when the player stats are both greatly lower and higher than the merchant stats.
It must not depend on the transaction type, because it makes no sense for the player to have the better chance to win a roll the higher price they ask for a sold item.

Capostrophic:
Hm. I suppose. I'll try to write one later.

psi29a:
Honestly, this looks clean. It's just a pure formula without any if/thens which I'm happy about. 

I'm starting to get into the mindset that we're no longer a strict Morrowind/Vanilla replacement engine, but really a game-engine that happens to also play Morrowind with things that we've fixed along the way to improve the experience and not emulate pure Vanilla exactly. I believe there is some wiggle-room so long it isn't an overarching change.

zinnschlag:
> I'm starting to get into the mindset that we're no longer a strict Morrowind/Vanilla replacement engine, but really a game-engine that happens to also play Morrowind

From my point of view that is really what the project was about all along (remember I joined because I needed an engine to run Redemption on, not Morrowind).

We do have a bit of wiggle-room, but we have been getting a lot of use out of that recently and I do think staying as closely as possible to vanilla is desirable for default un-modded content.

Still, I am only mildly disfavouring this PR. If you really want to merge it, I don't have a problem with that.

psi29a:
I'm in favour of merging this.

However, we should go through our open tickets and really start sorting things that should go post 1.0 since this PR was based on the idea that the issue was for pre 1.0

That, or another classification than pre/post 1.0, but things that 'would be nice'.

zinnschlag:
pre/post 1.0 yes. Don't see much point in "would be nice".

psi29a:
OK, that's for future issue sorting. This ticket is done and mergable right? :)

psi29a:
If you want, create a feature on the bug tracker for post-1.0

psi29a:
Merged

PR #1648, opened 2018-03-26
Classic reflected absorb attribute behavior option (bug #4135)

Capostrophic:
Make Morrowind behavior the default and the MCP-like behavior optional so that nobody complains.


Capostrophic:
dammit git

PR #1649, opened 2018-03-27
Classic reflected absorb attribute behavior option (bug #4135)

Capostrophic:
Make Morrowind behavior the default and ~~the proper~~ previous behavior optional so that nobody complains.


lysol90:
Too many times I've heard people say "OpenMW sucks ass because I can't use absorb on enemies with reflect to become totally OP and ruin the challenge anymore!!". This will hopefully shut them up. :)

SashaTheRed:
Having this would‘ve saved me some headaches when fighting Gaenor!

Capostrophic:
Yeah, I only tested on Dremoras, but this should work against him.

PR #1651, opened 2018-03-27
remove native GLESv* support, as it never worked; make things less co…

psi29a:
…mplicated as result, fixed lighting.glsl to work with gl2es thanks to ptitSeb


CI/before_script.msvc.sh
```diff
@@ -613,8 +613,7 @@ fi
 			SUFFIX=""
 		fi
 
-		add_runtime_dlls "$(pwd)/bin/lib"{EGL,GLESv2}${SUFFIX}.dll \
-			"$(pwd)/bin/Qt5"{Core,Gui,Network,OpenGL,Widgets}${SUFFIX}.dll
+		add_runtime_dlls "$(pwd)/bin/Qt5"{Core,Gui,Network,OpenGL,Widgets}${SUFFIX}.dll
```
xyzz: You sure that's not needed? that's some Qt stuff - https://doc.qt.io/qt-5/windows-requirements.html

psi29a: That is for EG/GLESv2 stuff, Qt still uses OpenGL which is pulled in in line 616.

@ananace can you verify, just in case?

psi29a: By default, Qt will determine whether the system's opengl32.dll provides OpenGL 2 functions. If these are present, opengl32.dll is used, otherwise the ANGLE libraries (libEGL.dll and libGLESv2.dll) will be used. In case the ANGLE libraries are missing or initialization fails for some reason, an additional fallback is attempted by trying to load opengl32sw.dll.

ananace: According to Qt's packaging documentation;
> If ANGLE or dynamic OpenGL (the default) is used, you additionally need to include both libEGL.dll and libGLESv2.dll from Qt's lib directory as well as the HLSL compiler from DirectX. The HLSL compiler library, d3dcompiler_XX.dll, where XX is the version number that ANGLE (libGLESv2) was linked against.

The d3dcompiler DLL isn't packaged, so they probably have never actually even been loadable...

I'll do a few tests after work just to be sure though.

psi29a: @ananace you get around to any tests?

ananace: Sorry, got swamped with work that was due before easter.

Done tests now and there should be no loss of functionality from removing the EGL and GLESv2 DLLs, as expected they couldn't be used for ANGLE anyway.

psi29a: Great, thanks for checking this.

PR #1654, opened 2018-03-28
Allow jumping when you're stuck on a slope (bug #4221)

Capostrophic:
Apparently may or may not have side effects depending on how the character got stuck. Suggestions welcome.


zinnschlag:
This is an interesting idea. I am inclined to merge. Any objections anyone?

psi29a:
This is a better option than being stuck and having to break immersion by TCLing.

PR #1655, opened 2018-03-29
Switch playlists if tracks don't exist

h313:
Fixes #4134 by editing `playRandomTrack()` so that when either Combat or Explore is chosen but contains no tracks, and when the other playlist does contain tracks, the playlist is switched to the other.

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -395,6 +395,7 @@ namespace MWSound
     {
         std::vector<std::string> filelist;
         auto &tracklist = mMusicToPlay[mCurrentPlaylist];
+        // Find all tracks within the playlist
         if (mMusicFiles.find(mCurrentPlaylist) == mMusicFiles.end())
         {
             const std::map<std::string, VFS::File*>& index = mVFS->getIndex();
@@ -417,8 +418,20 @@ namespace MWSound
         else
             filelist = mMusicFiles[mCurrentPlaylist];
 
+        // If there are no tracks, switch to the other playlist if it exists
         if(filelist.empty())
-            return;
+        {
+            if(mCurrentPlaylist == "Combat" && tracklist.find("Explore") != tracklist.end())
+            {
+                mCurrentPlaylist == "Explore";
+                startRandomTitle();
+            }
+            else if (mCurrentPlaylist == "Explore" && tracklist.find("Combat") != tracklist.end())
+            {
+                mCurrentPlaylist == "Combat";
+                startRandomTitle();
+            }
+        }
 
         // Do a Fisher-Yates shuffle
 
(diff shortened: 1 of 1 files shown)


apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -417,8 +418,20 @@ namespace MWSound
...
+            if(mCurrentPlaylist == "Combat" && tracklist.find("Explore") != tracklist.end())
+            {
+                mCurrentPlaylist == "Explore";
+                startRandomTitle();
+            }
+            else if (mCurrentPlaylist == "Explore" && tracklist.find("Combat") != tracklist.end())
+            {
+                mCurrentPlaylist == "Combat";
+                startRandomTitle();
+            }
```
kcat: This won't work since you're using `==` to test for equality, where you mean to use `=` to assign.

But even with that fixed, this will go into an infinite loop if both Combat and Explore don't have any tracks. I'm not keen on hard-coding these playlist names, either. A better solution may be to simply not change to the new playlist if it doesn't have any tracks.

h313: What if it chooses the next playlist that still contains songs to play from instead?

kcat: What do you mean by "next playlist"? There isn't any real order to the playlists, one is just selected based on the current game state (which currently is just 'if player is in combat -> Combat, else -> Explore'). In the future we'll want to have mod-defined playlists that have their own state checks for more intricate music definitions.

Did you test these changes, by the way? This can't compile since `tracklist` is a reference to a vector of ints, not a map with a string key.

h313: I meant the next playlist containing songs found in `tracklist`. Would it be better to just default everything back to Explore then, like option #2 discussed in bug [#4134](https://bugs.openmw.org/issues/4134)?

kcat: `tracklist` is a vector of ints, which is indices for potential songs to play in the current playlist. Each playlist has its own tracklist, which gets refilled with new indices periodically.

IMO, it would be better to simply not change the current playlist (in `playPlaylist`) if the requested one has no songs.

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -417,8 +418,20 @@ namespace MWSound
         else
             filelist = mMusicFiles[mCurrentPlaylist];
 
+        // If there are no tracks and the other playlist is populated, switch to the other playlist if it exists
         if(filelist.empty())
-            return;
+        {
+            if(mCurrentPlaylist == "Combat" && tracklist.find("Explore") != tracklist.end() && !tracklist["Explore"].empty())
```
akortunov: tracklist is a std::vector<int>, so you can not search for string. Also std::vector does not have the find() function. Do you mean the std::find?

apps/openmw/mwsound/soundmanagerimp.cpp
Capostrophic: >it would be better to simply not change the current playlist...

I agree.

PR #1659, opened 2018-03-30
Extend taking resistance to magic into account to all resistable magic effects (bug #4309)

Capostrophic:
As most magic effects don't have a resistance effect assigned, as seen [here](https://github.com/OpenMW/openmw/blob/openmw-43/components/esm/loadmgef.cpp#L272), there's no need for a Harmful flag check. This allows certain Illusion effects to take resistance/weakness to magicka of the actor into account. I edited getEffectResistance() and inflict() to account for that, making sure NPCs will still only react aggressively to harmful effects.

I also improved the behavior of fully resisted harmful effects -- they won't make the health bar visible.


apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -488,13 +489,14 @@ namespace MWMechanics
                     else if (castByPlayer)
                         MWBase::Environment::get().getWindowManager()->messageBox("#{sMagicTargetResisted}");
                 }
-
-                // If player is attempting to cast a harmful spell, show the target's HP bar
-                if (castByPlayer && target != caster)
+                else if (magicEffect->mData.mFlags & ESM::MagicEffect::Harmful && castByPlayer && target != caster)
```
akortunov: How about non-AI effects with Self range (e.g. Light or Chameleon)? These effects will be resisted too?

Capostrophic: Oops, I haven't thought of them.

Capostrophic:
Is there anything I missed?

zinnschlag:
Looks like the kind of change that is most likely okay, but could have unintended side-effects we didn't think of. Only way to find out is to throw it out to the wider audience and see if anything explodes. I am inclined to merge. Any objections?

akortunov:
Commit 2dca389a3b00dd32697bd16703c47e29074dea6d breaks some things for me since now we resist all effects:
1. Spells with Self range effects stopped to work
2. Open spells stopped to work since target does not have creature stats.

Probably we need to limit spell resisting only for non-harmful illusion spells with non-Self range.

PR #1660, opened 2018-03-30
[Discussion] Improve glyph-replacing hacks (bug #1987)

Capostrophic:
Considering the first halves of Win-1250, Win-1251 and Win-1252 are all very similar and fonts and encodings are equally broken on all localizations of the game, I've removed the special treatment of Win-1252 encoding.

Notable changes, kudos to akortunov: much more vanilla-based character replacements, duplication and typo fixes, limited font-based Cyrillic letter ё support.

### DIFF
--- a/components/fontloader/fontloader.cpp
+++ b/components/fontloader/fontloader.cpp
@@ -339,52 +339,76 @@ namespace Gui
                                + MyGUI::utility::toString((fontSize-data[i].ascent)));
             code->addAttribute("size", MyGUI::IntSize(static_cast<int>(data[i].width), static_cast<int>(data[i].height)));
 
-            // More hacks! The french game uses several win1252 characters that are not included
-            // in the cp437 encoding of the font. Fall back to similar available characters.
-            if (mEncoding == ToUTF8::CP437)
+            // Fall back from unavailable Windows-1252 encoding symbols to similar characters available in the game fonts
+            std::multimap<int, int> additional; // fallback glyph index, unicode
+            additional.insert(std::make_pair(156, 0x00A2)); // cent sign
+            additional.insert(std::make_pair(89, 0x00A5)); // yen sign
+            additional.insert(std::make_pair(221, 0x00A6)); // broken bar
+            additional.insert(std::make_pair(99, 0x00A9)); // copyright sign
+            additional.insert(std::make_pair(97, 0x00AA)); // prima ordinal indicator
+            additional.insert(std::make_pair(60, 0x00AB)); // double left-pointing angle quotation mark
+            additional.insert(std::make_pair(45, 0x00AD)); // soft hyphen
+            additional.insert(std::make_pair(114, 0x00AE)); // registered trademark symbol
+            additional.insert(std::make_pair(45, 0x00AF)); // macron
+            additional.insert(std::make_pair(241, 0x00B1)); // plus-minus sign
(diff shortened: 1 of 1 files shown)


components/fontloader/fontloader.cpp
```diff
@@ -340,51 +340,48 @@ namespace Gui
             code->addAttribute("size", MyGUI::IntSize(static_cast<int>(data[i].width), static_cast<int>(data[i].height)));
 
             // More hacks! The french game uses several win1252 characters that are not included
-            // in the cp437 encoding of the font. Fall back to similar available characters.
-            if (mEncoding == ToUTF8::CP437)
+            // in the font. Fall back to similar available characters.
+            std::multimap<int, int> additional; // <cp437, unicode>
```
akortunov: IMO, CP437 comment is not needed now.
Also Morrowind supports the − U+2212 (MINUS SIGN) - some book mods use it.
Also it would be nice to add comments to all pairs, not just for 5 ones.

components/fontloader/fontloader.cpp
```diff
@@ -340,51 +340,48 @@ namespace Gui
...
+            additional.insert(std::make_pair(44, 0x201E));
+            additional.insert(std::make_pair(43, 0x2020));
+            additional.insert(std::make_pair(43, 0x2021));
+            additional.insert(std::make_pair(94, 0x02C6));
+            additional.insert(std::make_pair(37, 0x2030));
+            additional.insert(std::make_pair(83, 0x0160));
+            additional.insert(std::make_pair(60, 0x2039));
+            additional.insert(std::make_pair(79, 0x0152));
+            additional.insert(std::make_pair(90, 0x017D));
+            additional.insert(std::make_pair(39, 0x2019));
```
akortunov: We already have the 2019 above. IMO, there should be the 2018 (LEFT SINGLE QUOTATION MARK).

akortunov:
For testing, we can put all these signs in the messagebox, for example:
`MessageBox "−–—“”‘’" "Ok"`
For some reason, console does not display these signs, but they work in-game.

Capostrophic:
That's very helpful, thanks.

components/fontloader/fontloader.cpp
Capostrophic: We probably don't want to add fallbacks to the entire Unicode character table, though.

scrawl:
Most of those glyph replacements are ASCII, which most definitely are the same in the different win encodings. I'm not sure about the non-ASCII (>128) codes though. Did you check if those refer to the same glyphs?

> Note: while fonts like Pelagiad which include some of the missing characters exist, there isn't a way to detect if a glyph for a character actually exists, at not yet.

Such (TrueType) fonts are unaffected by the code path you're editing. They don't go through FontLoader at all.

> For some reason, console does not display these signs (probably there are no glyphs in console font),

You can add glyphs by editing the ranges in openmw_font.xml, provided that the font actually contains those glyphs.

Capostrophic:
>Such (TrueType) fonts are unaffected by the code path you're editing. They don't go through FontLoader at all.
Right, I realized that today and forgot to edit the description of this pull request.
>Did you check if those refer to the same glyphs?
They refer to the same or pretty much the same glyphs in Russian font at least, according [to some data I got](http://pic.fullrest.ru/KkvNNS8q.png). Not sure about Polish.

PR #1663, opened 2018-04-01
try xenial on travis

psi29a:


zinnschlag:
This was present as a merge request on GitLab before IIRC and now it isn't. What happened?

psi29a:
travis ci doesn't support xenial... they have a container but you can't do anything with the container. If we continue with travis-ci, we're stuck with trusty. Closing this. GL on the other hand allows you to roll your own containers in which you can do things...

PR #1666, opened 2018-04-04
Reduce twitching for player turning animations

akortunov:
Fixes [bug #4574](https://gitlab.com/OpenMW/openmw/issues/4574). Related [forum topic](https://forum.openmw.org/viewtopic.php?f=6&t=5039).

During turning for small angles anim speed is near 0, so mTurnAnimationThreshold almost make no sense. As workaround, we use a smoothTurn function for NPCs, but we can not use it for player.

The main idea: play turning animation a bit for player after he stops to turn.
As for me, it is not ideal solution in its current state, but it definely better than our current behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@
     Bug #4563: Fast travel price logic checks destination cell instead of service actor cell
     Bug #4565: Underwater view distance should be limited
     Bug #4573: Player uses headtracking in the 1st-person mode
+    Bug #4574: Player turning animations are twitchy
     Feature #2606: Editor: Implemented (optional) case sensitive global search
     Feature #3083: Play animation when NPC is casting spell via script
     Feature #3103: Provide option for disposition to get increased by successful trade
--- a/apps/openmw/mwgui/loadingscreen.cpp
+++ b/apps/openmw/mwgui/loadingscreen.cpp
@@ -170,11 +170,17 @@ namespace MWGui
         // We are already using node masks to avoid the scene from being updated/rendered, but node masks don't work for computeBound()
         mViewer->getSceneData()->setComputeBoundingSphereCallback(new DontComputeBoundCallback);
 
+        mShowWallpaper = visible && (MWBase::Environment::get().getStateManager()->getState()
+                == MWBase::StateManager::State_NoGame);
+
+        if (!visible)
+        {
+            draw();
+            return;
+        }
+
         mVisible = visible;
         mLoadingBox->setVisible(mVisible);
-
-        mShowWallpaper = mVisible && (MWBase::Environment::get().getStateManager()->getState()
-                == MWBase::StateManager::State_NoGame);
 
         setVisible(true);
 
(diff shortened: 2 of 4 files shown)


scrawl:
Isn't your solution framerate dependent? For the same turning speed, ```rot``` will be larger if your framerate is low.

Capostrophic:
When turning animation is still playing and you jump the character might briefly continue to play turning animation and then immediately transition into jumping pose. It's jarring.

zinnschlag:
Both scrawl and Capostrophic have reasonable concerns. Can these be addressed? If yes, I would be okay with a merge.

Capostrophic:
There's a related [hack](https://github.com/OpenMW/openmw/blob/openmw-44/apps/openmw/mwmechanics/aiwander.cpp#L482) in AiWander. Something should be done about it.

akortunov:
The only thing left is do determine when turning animations should be interrupted (as for greetings in AiWander), and when should not.

Capostrophic:
I finally figured out what's wrong with spellcasting stance turning animation. Apparently there *are* unique animation groups, but they have uncommon naming format: they're SpellTurnLeft and SpellTurnRight, while OpenMW tries to find TurnLeftSpell and TurnRightSpell (and fails) as if they were usual weapon shortgroup turning animation groups.

akortunov:
> I finally figured out what's wrong with spellcasting stance turning animation. 

Will you create a separate bugreport and PR, or it is up to me to fix this issue?

Capostrophic:
Hm. I'll do it by myself.

akortunov:
Since we already have smoothing for NPCs, I suggest to use a new behaviour only for player.

akortunov:
Reworked a bit:
1. Use new behaviour only for player.
2. Fix regression in check if we need to reset the mTurnAnimationThreshold.
3. Limit turning speed multiplier from 1.0 to 1.5 - with slow turning spead multiplier was near 0 and mTurnAnimationThreshold was pretty useless - we did not actually play turning animation anyway.
@Capostrophic, can you retest this PR?

Capostrophic:
I'll be able to do that some time next week.

Capostrophic:
Tested. Seems to work fine, although turning animation continues if you jump during it and land again, which looks a bit odd. Possibly related to my changes to same-movement-anim handling.

Another, not exactly related to this PR, but annoying issue is that turning animations are framerate dependent (happens with traditional behavior too), which causes all kinds of issues during turning if vertical synchronization is turned off and there is no framerate limit. If the framerate is too high, it causes the turning animation to loop from the start endlessly without actually turning the character, and while in this case it's a bit better than the original behavior (because the animation has to run for a bit), it still looks bad.

akortunov:
> Another, not exactly related to this PR, but annoying issue is that turning animations are framerate dependent (happens with traditional behavior too), which causes all kinds of issues during turning if vertical synchronization is turned off and there is no framerate limit. If the framerate is too high, it causes the turning animation to loop from the start endlessly without actually turning the character, and while in this case it's a bit better than the original behavior (because the animation has to run for a bit), it still looks bad.

Because animation speed is tied to rotation amount. With very high FPS rotation amount is near 0 and we almost do not play turning animation.

akortunov:
> turning animation continues after you jump during it and land again

As in vanilla game. Looks like movement animation blending do not work well with landing animation.

akortunov:
I increased priority of jumping animations. Landing should work fine now.

zinnschlag:
Missing changelog entry.

PR #1668, opened 2018-04-05
Remove getFallbackString missing value warnings

Capostrophic:
Ultimately these are too spammy, because there are a lot of weather settings that are tried to be retrieved "just in case", and the warnings appear every startup.

### DIFF
--- a/components/fallback/fallback.cpp
+++ b/components/fallback/fallback.cpp
@@ -1,7 +1,5 @@
 #include "fallback.hpp"
 
-#include <iostream>
-
 #include <boost/lexical_cast.hpp>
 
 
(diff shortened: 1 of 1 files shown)


zinnschlag:
Was about to state that I don't see any warnings, but then I noticed that my master branch was out of date. Looks like a semi-recent change. Removing the warning messages seems a bit radical. Can we identify what changed and improve it instead?

akortunov:
> Can we identify what changed and improve it instead?

002ad9ae1b85e98086b82004bb1a62fb91614619. This message was added recently.
As a better solution, we can add a severity for warnings.
This approach can get rid of spam in log (e.g. about empty texture layers or out of bounds UV).

Capostrophic:
Hm. If levels of debug for log messages are going to be added, they would probably the best way for working this situation around. I don't see any better way to workaround this, so... I'm going to close this then.

PR #1670, opened 2018-04-07
[Testing needed] Unhardcode sunset and sunrise settings

akortunov:
Fixes [bug #1990](https://bugs.openmw.org/issues/1990).
Implements settings, mentioned in the bug desctiption.

It would be nice to have an independent testing of this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.45.0
 ------
 
+    Bug #1990: Sunrise/sunset not set correct
     Bug #2835: Player able to slowly move when overencumbered
     Bug #3374: Touch spells not hitting kwama foragers
     Bug #3591: Angled hit distance too low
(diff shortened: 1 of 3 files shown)


akortunov:
> Слушай, не думали над созданием дискорд канала?

OpenMW already has the [Discord channel](https://discordapp.com/invite/wd3eSas).

Artemex:
thanks, man!

scrawl:
I think we need to be more specific on what needs to be tested lest this PR becomes stale. Why do you want an independent test? Is it because different installations may have a different INI? And what needs to be tested, I suppose that sunset and sunrise are at the same hour as before and/or as in vanilla MW?

zinnschlag:
This looks like another PR that we may want to merge early in the development cycle of 0.45 to get maximum testing. Any objections?

Capostrophic:
I don't have any.

psi29a:
<3 how everyone is just blazing through the remaining issues, 0.44 is not even yet out the door and we're doing great on 0.45!  Great job everyone! :)

PR #1672, opened 2018-04-09
<random>-based RNG

Capostrophic:
Now that OpenMW is a C++11 project, perhaps it's time to get rid of std::rand.

I'm not too knowledgeable of C++'s random stuff and classes in general, so please let me know if there are major flaws in the logic of this (there usually are).

Note: I only just found #1311 discussion, having forgotten about it earlier.

### DIFF
--- a/components/misc/rng.cpp
+++ b/components/misc/rng.cpp
@@ -1,28 +1,31 @@
 #include "rng.hpp"
-#include <cstdlib>
-#include <ctime>
+
+#include <chrono>
+#include <random>
 
 namespace Misc
 {
 
+    std::mt19937 Rng::generator = std::mt19937();
+
     void Rng::init()
     {
-        std::srand(static_cast<unsigned int>(std::time(NULL)));
+        generator.seed(static_cast<unsigned int>(std::chrono::high_resolution_clock::now().time_since_epoch().count()));
     }
 
     float Rng::rollProbability()
     {
-        return static_cast<float>(std::rand() / (static_cast<double>(RAND_MAX)+1.0));
+        return std::uniform_real_distribution<float>(0, 1 - std::numeric_limits<float>::epsilon())(generator);
     }
 
     float Rng::rollClosedProbability()
     {
-        return static_cast<float>(std::rand() / static_cast<double>(RAND_MAX));
+        return std::uniform_real_distribution<float>(0, 1)(generator);
     }
 
     int Rng::rollDice(int max)
     {
-        return static_cast<int>((std::rand() / (static_cast<double>(RAND_MAX)+1.0)) * (max));
+        return std::uniform_int_distribution<int>(0, max - 1)(generator);
     }
 
 }
--- a/components/misc/rng.hpp
+++ b/components/misc/rng.hpp
@@ -2,6 +2,7 @@
 #define OPENMW_COMPONENTS_MISC_RNG_H
 
 #include <cassert>
+#include <random>
 
 namespace Misc
 {
@@ -12,6 +13,9 @@ namespace Misc
 class Rng
 {
 public:
+
+    /// create a RNG
+    static std::mt19937 generator;
 
     /// seed the RNG
     static void init();


kiljacken:
Re-initializing the random number generator for each roll is probably a bad idea, god knows how much work it has to do each time. You should probably save the `std::uniform_real_distribution` as a field in `Rng` and then possibly add an additional function to `Rng` that gets a sample from the already init'ed distribution.

components/misc/rng.hpp
```diff
@@ -14,7 +16,7 @@ class Rng
 public:
 
     /// generate a random number
-    static double init();
+    static double getRandomNumber() { return dis(seed); }
```
akortunov: Static functions can not use non-static fields.

components/misc/rng.cpp
```diff
@@ -1,28 +1,24 @@
...
     {
-        return static_cast<float>(std::rand() / static_cast<double>(RAND_MAX));
+        Rng generator;
+        return static_cast<float>(generator.getRandomNumber() / static_cast<double>(RAND_MAX));
     }
 
     int Rng::rollDice(int max)
     {
-        return static_cast<int>((std::rand() / (static_cast<double>(RAND_MAX)+1.0)) * (max));
+        return int(rollProbability() * max);
```
elsid: Also here:
```c++
return std::uniform_int_distribution<int>(0, max)(seed);
```

components/misc/rng.cpp
```diff
@@ -1,28 +1,24 @@
...
-    void Rng::init()
-    {
-        std::srand(static_cast<unsigned int>(std::time(NULL)));
-    }
-
     float Rng::rollProbability()
     {
-        return static_cast<float>(std::rand() / (static_cast<double>(RAND_MAX)+1.0));
+        Rng generator;
+        return static_cast<float>(generator.getRandomNumber() / (static_cast<double>(RAND_MAX)+1.0));
```
elsid: Use custom distribution here:
```c++
return std::uniform_real_distribution<float>(0, 1 - std::numeric_limits<float>::epsilon())(seed);
```

components/misc/rng.cpp
```diff
@@ -1,28 +1,24 @@
...
     }
 
     float Rng::rollClosedProbability()
     {
-        return static_cast<float>(std::rand() / static_cast<double>(RAND_MAX));
+        Rng generator;
+        return static_cast<float>(generator.getRandomNumber() / static_cast<double>(RAND_MAX));
```
elsid: Also here:
```c++
return std::uniform_real_distribution<float>(0, 1)(seed);
```

components/misc/rng.cpp
```diff
@@ -1,28 +1,24 @@
...
 
-    void Rng::init()
-    {
-        std::srand(static_cast<unsigned int>(std::time(NULL)));
-    }
-
     float Rng::rollProbability()
     {
-        return static_cast<float>(std::rand() / (static_cast<double>(RAND_MAX)+1.0));
+        Rng generator;
```
elsid: Every call for this method will produce initialization for random_device. It is will cause on perfomance.

Capostrophic:
There are some very odd fatal issues with loading/starting a game now.

elsid:
Change distribution high border to `max - 1` for rollDice. `std::uniform_int_distribution` produce random values in interval `[a, b]`. Otherwise it could be overflow [here](https://github.com/OpenMW/openmw/blob/7a389f7ab117acad15d813a33f63877f45da98ef/apps/openmw/mwmechanics/levelledlist.hpp#L61) and some other places.

components/misc/rng.cpp
```diff
@@ -1,28 +1,24 @@
...
+#include <random>
 
 namespace Misc
 {
 
-    void Rng::init()
-    {
-        std::srand(static_cast<unsigned int>(std::time(NULL)));
-    }
+    std::minstd_rand Rng::generator = std::minstd_rand();
```
elsid: Global object in multithreaded application without any synchronization if it is being used from different threads will produce race condition.

Capostrophic:
>Otherwise it could be overflow here and some other places.

Oh, that was it, now RNG doesn't break everything.

components/misc/rng.cpp
Capostrophic: Ouch. Could I just add a lock here?

elsid: Yes, I think, `std::mutex` and `std::lock_guard` should be fine. But make private both generator and mutex.

scrawl: IIRC all uses of Rng are from the main thread anyway. Our threads are mostly used for deterministic loading tasks where you can't possibly need anything random. Just declare the class not thread safe and be done with it.

scrawl:
Would be nice to have a unit test to make sure nothing broke. Of course if the test fails it will only fail randomly, but if you make a long enough for loop it should fail (or not fail) somewhat reliably.

scrawl:
On the other hand, the only thing left to test really is probability vs. closedProbability, which would surely take millions of runs to discover a difference, so let's skip that. Merging.

scrawl:
Can you elaborate why you chose minstd_rand? Isn't mersenne twist better randomness-wise?

Capostrophic:
I tried to keep things as "drop-in" as possible, so I used minstd_rand because it is LCG like rand() and seems to be a bit faster than mersenne twister. Though it might be "savings on matches" as a saying goes.

scrawl:
> minstd_rand because it is LCG like rand()

Then what's the point of replacing rand() if the alternative isn't any better? I'm not a fan of making code 'modern' just for the sake of it. Don't change a running system and all that.

scrawl:
Can you move init() back to the cpp file so that any file that includes the header (quite a few) won't needlessly include ```<chrono>```?

Capostrophic:
Ok, done, nothing broke.

PR #1673, opened 2018-04-11
Fix white artifacts when MSAA is used with shader lighting (bug #4143)

Capostrophic:
> The centroid and sample (the latter requires OpenGL 4.0 or ARB_gpu_shader5) qualifiers affect interpolation, but they are somewhat special, grammatically speaking. [...]
> 
> They may not technically be interpolation qualifiers, but they do control aspects of interpolation. **They only have an effect when Multisampling is being used.**
> 
> During multisampling, if centroid is not present, then the written value can be interpolated to to an arbitrary position within the pixel. This may be the pixel's center, one of the sample locations within the pixel, or an arbitrary location. Most importantly of all, **this sample may lie outside of the actual primitive** being rendered, since a primitive can cover only part of a pixel's area.

— from [here](https://www.khronos.org/opengl/wiki/Type_Qualifier_(GLSL)#Interpolation_qualifiers)

So, I added that. Seems to work just right.

### DIFF
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -47,9 +47,9 @@ varying float depth;
 #define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
 
 #if !PER_PIXEL_LIGHTING
-varying vec4 lighting;
+centroid varying vec4 lighting;
 #else
-varying vec4 passColor;
+centroid varying vec4 passColor;
 #endif
 varying vec3 passViewPos;
 varying vec3 passNormal;
--- a/files/shaders/objects_vertex.glsl
+++ b/files/shaders/objects_vertex.glsl
@@ -38,9 +38,9 @@ varying float depth;
 #define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
 
 #if !PER_PIXEL_LIGHTING
-varying vec4 lighting;
+centroid varying vec4 lighting;
 #else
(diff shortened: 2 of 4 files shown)


akortunov:
Things are definely better now, but this fix does not work with per-pixel lighting.
Also it does not fix [this](https://bugs.openmw.org/attachments/download/3049/Screenshot_Pelagiad.jpg) issue.

files/shaders/objects_fragment.glsl
```diff
@@ -47,7 +47,7 @@ varying float depth;
 #define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
 
 #if !PER_PIXEL_LIGHTING
-varying vec4 lighting;
+centroid varying vec4 lighting;
 #else
 varying vec4 passColor;
```
akortunov: Adding "centroid" to passColor in both shaders fixed the issue with per-pixel lighting too.

scrawl:
The declarations in vertex and pixel shaders need to match. As is, this looks like undefined behavior.

akortunov:
> The declarations in vertex and pixel shaders need to match

Done, as I see.

Capostrophic:
Hm. AppVeyor couldn't download OpenAL-Soft for some reason.

As far as I can understand, using centroid qualifier on any variable used for the lighting pass of the shader fixes these artifacts by making the renderer "do it right" and don't just pick a random sample for that part. There shouldn't be downsides to this.

PR #1674, opened 2018-04-12
Don't prefer a DDS file if the originally requested file exists (bug #1392)

Capostrophic:
Fallback to a DDS file if the requested texture path doesn't point to an existing file, not vice versa.

### DIFF
--- a/components/misc/resourcehelpers.cpp
+++ b/components/misc/resourcehelpers.cpp
@@ -63,27 +63,15 @@ std::string Misc::ResourceHelpers::correctResourcePath(const std::string &topLev
 
     std::string origExt = correctedPath;
 
-    // since we know all (GOTY edition or less) textures end
-    // in .dds, we change the extension
-    bool changedToDds = changeExtensionToDds(correctedPath);
-    if (vfs->exists(correctedPath))
+    if (vfs->exists(origExt)
+     || (changeExtensionToDds(correctedPath) && vfs->exists(correctedPath)))
         return correctedPath;
-    // if it turns out that the above wasn't true in all cases (not for vanilla, but maybe mods)
-    // verify, and revert if false (this call succeeds quickly, but fails slowly)
-    if (changedToDds && vfs->exists(origExt))
-        return origExt;
 
     // fall back to a resource in the top level directory if it exists
-    std::string fallback = topLevelDirectory + "\\" + getBasename(correctedPath);
-    if (vfs->exists(fallback))
+    std::string fallback = topLevelDirectory + "\\" + getBasename(origExt);
+    if (vfs->exists(fallback)
+     || (changeExtensionToDds(fallback) && vfs->exists(fallback)))
         return fallback;
-
-    if (changedToDds)
-    {
-        fallback = topLevelDirectory + "\\" + getBasename(origExt);
-        if (vfs->exists(fallback))
-            return fallback;
-    }
 
     return correctedPath;
 }


scrawl:
The code was written this way for performance reasons. Can we get some profiling results before and after? I'm sure our new VFS is much faster than the slow Ogre resources system that this code basically worked around, but just to be sure.

Capostrophic:
>some profiling results
Hm. Do you mean the in-game profiler info or something entirely different? No difference there. At least I don't see anything significant. No noticeable changes in IO department either.

scrawl:
Whatever difference there might be would be visible in the startup time, mostly. You could use a fancy profiler like gprof or callgrind or just simple printf:

<pre>
// Let's profile myFunction
void myFunction(){
static double timer = 0.0;
osg::Timer t;

// ... do the actual function here ...

timer += t.time_s();
std::cout << "time: " << timer << "s" << std::endl;
}
</pre>

components/misc/resourcehelpers.cpp
```diff
@@ -61,30 +61,16 @@ std::string Misc::ResourceHelpers::correctResourcePath(const std::string &topLev
...
         correctedPath = prefix1 + correctedPath;
 
-    std::string origExt = correctedPath;
-
-    // since we know all (GOTY edition or less) textures end
-    // in .dds, we change the extension
-    bool changedToDds = changeExtensionToDds(correctedPath);
-    if (vfs->exists(correctedPath))
+    if (vfs->exists(correctedPath)
+     || changeExtensionToDds(correctedPath) && vfs->exists(correctedPath))
```
kcat: You should probably add some parenthesis here, to clarify the grouping:
```
if (vfs->exists(correctedPath)
 || (changeExtensionToDds(correctedPath) && vfs->exists(correctedPath)))
```
Some compilers will actually complain otherwise, since it's not obvious that `&&` binds stronger than `||`.

Capostrophic: Right. GCC does complain about this.

Capostrophic:
@unelsson,
in this case changeExtensionToDds(path) acts as a condition - it tries to change the path to one with .dds at the end and returns a boolean-type value if that was successful (i.e. the path wasn't already leading to a DDS file). It does return something that can be compared with something (in this case with `true`), so it's not nonsense to use it.
The logic goes like this: if the path leads to an existing file OR the path with dds at the end leads to (AND) an existing file, return this path. Although some compilers may understand the current code as "if the file exists OR the path can be changed to one with .dds at the end, AND the file with dds at the end exists" which obviously leads to issues.

scrawl:
Ok, but with profiling I mean hard results, such as: 'when loading balmora before this change, all calls to correctResourcePath() take 8ms combined, now they take 9ms'. Can have that?

Capostrophic:
I just tried to log handleSourceTexture() and it seems that every call of correctTexturePath (== correctResourcePath) there takes about 2-3 microseconds (i.e. 1 thousandth of a millisecond) and 4-5 microseconds without and with the pull request respectively. There were about 5000 calls when I teleported from another place to Balmora and walked around, ultimately they result in a 20ms overhead compared to the original 14ms - over a couple of minutes.

scrawl:
Okay. That is a steep increase percentage wise, but not too significant in absolute terms, especially considering that loading usually runs unnoticably in a background thread anyway. I will merge it.

scrawl:
I think there is an error in the code though. In the "// fall back to a resource in the top level directory if it exists" part, you intend to test both original extension and dds, but you've already run changeExtensionToDds on correctedPath before so you end up testing against the dds version twice and never the original extension.

Capostrophic:
Fixed, I think.

PR #1677, opened 2018-04-15
Opencs: Lamoot's new icon set

unelsson:
Some minor issues with missing icons
https://forum.openmw.org/viewtopic.php?f=7&t=3864&p=53852#p53852

Didn't yet connect new texture-edit, land shape edit etc. icons yet to avoid merge conflicts while developing those features.


zinnschlag:
What is the status on this? Still missing the journal/info stuff and some of the new icons for terrain editing?

unelsson:
I think the iconset is quite complete, but there are some things that might be issues. I'd say ready for merge, but it's a good idea that this PR is still properly checked. This PR doesn't do anything to raster/scalable icons, it doesn't remove unused icons, and it also adds some unused special journal & topic icons to resources.qrc. While icons for edit modes are also added to resources, the actual code using these editmode-icons is mostly unchanged, as there's development going on with editmodes and I wanted to avoid conflicts with editmode-related code. In the future, I anticipate some need for special brush buttons with land shape editing in the future, but how lamoot made the icon, it's possible to use circle-square-point-custom -icons or just change these slightly to get more customized brush icons for land editing.

zinnschlag:
It seems the fog button is broken. Don't know if that was the case before. And the run button seems to still use the old icon.

Other than that everything looks fine.

zinnschlag:
Regarding the fog button: Apparently a user-error. I tried it on an exterior cell which doesn't have fog, which is why the button is faded out. Interior works normally. However the new icon does not work well with being faded out. Maybe it would better to remove this button on exterior cells?

unelsson:
Run button seems to be missing from the icons, so if Lamoot makes one, I can add it here.

I'd vote to set a separate PR for removing fog icon on exterior cells, it's not a big task, but not trivial either.

PR #1678, opened 2018-04-15
Display a message if the player tries to use an unavailable spell via a quick key (bug #4391)

Capostrophic:


scrawl:
From what I see, this is a new feature that none of the other quick key types have. We generally leave the binding intact case the item comes back (or the player buys the spell again). Why do this unassigning only for spells? Is it worth to break the consistency here?

Capostrophic:
Huh? Doesn't [this](https://github.com/Capostrophic/openmw/blob/quickkeys/apps/openmw/mwgui/quickkeysmenu.cpp#L540) part unassign the quick key if the (magic) item with requested ID wasn't found in the inventory?

scrawl:
That code only runs when loading a save game, and I think the unassign() is just for technical reasons because I can't call onAssignItem() if I don't have the MWWorld::Ptr. Technically, this is wrong, but I can't think of an easy fix to this without refactoring a lot of code.
Line [318](https://github.com/Capostrophic/openmw/blob/quickkeys/apps/openmw/mwgui/quickkeysmenu.cpp#L318) shows a "you no longer have that item" message when you try to use a non existent item, but the item will not be unassigned. This is intended because you could get another copy of the item from elsewhere. The use case for spells is probably a bit different, but again I'm not sure if breaking consistency is worth it. Maybe the current spells behavior would not feel like a bug if there was a similar "you no longer have X" message when trying to activate the key.

scrawl:
> I think the unassign() is just for technical reasons because I can't call onAssignItem() if I don't have the MWWorld::Ptr.

And it is in case the base item no longer exists in the content files, in which case references to it should be cleared as part of the save clearing mechanism.

Capostrophic:
Hm. Maybe use sMagicInvalidEffect ("You cannot cast this effect right now.")
 as a message [here](https://github.com/Capostrophic/openmw/blob/quickkeys/apps/openmw/mwgui/quickkeysmenu.cpp#L332) then?

scrawl:
Not specific enough imo. What's wrong with sQuickMenu5 "You no longer have %s"?

Capostrophic:
Well, nothing, I suppose... This displays "You no longer have %spellname%".

scrawl:
This construct ```x = search(...); x->``` doesn't make sense. search() can return 0, so you'd need to check spell for 0 before using it. If you're sure the spell is always there then better use find() instead. readRecord() already checks if the spell still exists in the content files before adding the binding, so it should always be found in activate() anyway.

PR #1679, opened 2018-04-15
OpenCS: Add brush settings to terrain edit

unelsson:
(clean) rebase of https://github.com/OpenMW/openmw/pull/1665
- Brush settings window for terrain texture editing mode


apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+    mBrushTextureLabel = "Brush: " + mBrushTexture;
+    selectedBrush = new QLabel(QString::fromUtf8(mBrushTextureLabel.c_str()), this);
+
+    const std::string& iconPoint = ":scenetoolbar/brush-point";
+    const std::string& iconSquare = ":scenetoolbar/brush-square";
+    const std::string& iconCircle = ":scenetoolbar/brush-circle";
+    const std::string& iconCustom = ":scenetoolbar/brush-custom";
+
+    TextureBrushButton *buttonPoint = new TextureBrushButton(QIcon (QPixmap (iconPoint.c_str())), "", this);
```
zinnschlag: That is not how you convert between std::string and QString (at least not in Qt4). Messes up the encoding. There is an fromUtf8 function.

apps/opencs/view/render/terraintexturemode.hpp
```diff
@@ -0,0 +1,108 @@
...
+        public:
+            BrushSizeControls(const QString &title, QWidget *parent);
+
+        private:
+            QSlider *brushSizeSlider;
+            QSpinBox *brushSizeSpinBox;
+            QHBoxLayout *layoutSliderSize;
+    };
+
+    class TextureBrushButton : public QPushButton
```
zinnschlag: Is there are reason why you are using a custom class for this?

zinnschlag:
Did another review (final test run still missing). I made some inline comments. Seems mostly okay. But I noticed that you don't handle drops of textures from the texture resource table.

unelsson:
How would the intended behavior for dropping texture from resource table be? Creating a new index to landtextures-table that points to the texture filename? There's a possible problem with that regarding Terraintexture primary edit #1680 branch - if user uses touchlandcommand after using a texture from resource table, it might not handle overlapping indexes properly.

apps/opencs/view/render/terraintexturemode.cpp
unelsson: Hm. Okay. I changed it to QString::fromStdString as QString::fromUtf8 takes only const char *str.

zinnschlag:
Finally got around to test this. Does not work for me. The texture mode button is black and no other buttons (or other GUI elements) are shown.

unelsson:
My guess is that mode button is black because there is no icon, I didn't include icons here, as there's a separate [https://github.com/OpenMW/openmw/pull/1677](pull request) that modifies same files quite extensively. Brush button window opens by pressing MMB.

unelsson:
No need to keep this open, as development of https://github.com/OpenMW/openmw/pull/1680 is ongoing and it has everything this has.

zinnschlag:
Yeah, the whole GUI is button based. Clicking in the scene should not open any GUI.

PR #1681, opened 2018-04-16
Some PVS-Studio and cppcheck fixes

tri4ng1e:
This PR fixes some warnings from static analysis tools PVS-Studio and cppcheck. Warning list is in commit description.

I'm very new to openMW, so if I missed something, please point me to that.

There is a bunch of warnings like `components/esm/esmreader.cpp 22 err V730 Not all members of a class are initialized inside the constructor. Consider inspecting: mCtx, mHeader`. Should I fix it?

And also there is a lot of trailing whitespaces - 193 files with 730 entries. I can commit that thing too.

**UPD:**
I found some strange thing in `apps/openmw/mwworld/cellstore.cpp : 688`, it's commented in PR. And I just found another thing like that few lines above in `CellStore::loadState` (line 677). Should I remove both of it or it's important?

### DIFF
--- a/apps/esmtool/record.cpp
+++ b/apps/esmtool/record.cpp
@@ -694,7 +694,7 @@ void Record<ESM::Dialogue>::print()
     // loads, rather than loading and then dumping. :-( Anyone mind if
     // I change this?
     ESM::Dialogue::InfoContainer::iterator iit;
-    for (iit = mData.mInfo.begin(); iit != mData.mInfo.end(); iit++)
+    for (iit = mData.mInfo.begin(); iit != mData.mInfo.end(); ++iit)
         std::cout << "INFO!" << iit->mId << std::endl;
 }
 
@@ -1123,7 +1123,7 @@ void Record<ESM::Pathgrid>::print()
 
     int i = 0;
     ESM::Pathgrid::PointList::iterator pit;
-    for (pit = mData.mPoints.begin(); pit != mData.mPoints.end(); pit++)
+    for (pit = mData.mPoints.begin(); pit != mData.mPoints.end(); ++pit)
     {
         std::cout << "  Point[" << i << "]:" << std::endl;
         std::cout << "    Coordinates: (" << pit->mX << ","
@@ -1135,7 +1135,7 @@ void Record<ESM::Pathgrid>::print()
     }
     i = 0;
     ESM::Pathgrid::EdgeList::iterator eit;
-    for (eit = mData.mEdges.begin(); eit != mData.mEdges.end(); eit++)
+    for (eit = mData.mEdges.begin(); eit != mData.mEdges.end(); ++eit)
     {
         std::cout << "  Edge[" << i << "]: " << eit->mV0 << " -> " << eit->mV1 << std::endl;
         if (eit->mV0 >= mData.mData.mS2 || eit->mV1 >= mData.mData.mS2)
--- a/apps/opencs/model/filter/parser.cpp
+++ b/apps/opencs/model/filter/parser.cpp
@@ -579,7 +579,7 @@ bool CSMFilter::Parser::parse (const std::string& filter, bool allowPredefined)
     }
     // We do not use isString() here, because there could be a pre-defined filter with an ID that is
     // equal a filter keyword.
-    else if (token.mType==Token::Type_String && allowPredefined)
+    else if (token.mType==Token::Type_String)
     {
         if (getNextToken()!=Token (Token::Type_EOS))
         {
--- a/apps/opencs/model/prefs/boolsetting.cpp
+++ b/apps/opencs/model/prefs/boolsetting.cpp
@@ -22,6 +22,8 @@ CSMPrefs::BoolSetting& CSMPrefs::BoolSetting::setTooltip (const std::string& too
 
 std::pair<QWidget *, QWidget *> CSMPrefs::BoolSetting::makeWidgets (QWidget *parent)
 {
+    if (mWidget != nullptr)
+        delete mWidget;
     mWidget = new QCheckBox (QString::fromUtf8 (getLabel().c_str()), parent);
     mWidget->setCheckState (mDefault ? Qt::Checked : Qt::Unchecked);
 
--- a/apps/opencs/model/prefs/boolsetting.cpp
+++ b/apps/opencs/model/prefs/boolsetting.cpp
@@ -22,8 +22,6 @@ CSMPrefs::BoolSetting& CSMPrefs::BoolSetting::setTooltip (const std::string& too
 
 std::pair<QWidget *, QWidget *> CSMPrefs::BoolSetting::makeWidgets (QWidget *parent)
 {
-    if (mWidget != nullptr)
-        delete mWidget;
     mWidget = new QCheckBox (QString::fromUtf8 (getLabel().c_str()), parent);
     mWidget->setCheckState (mDefault ? Qt::Checked : Qt::Unchecked);
 
--- a/apps/opencs/model/prefs/shortcuteventhandler.cpp
+++ b/apps/opencs/model/prefs/shortcuteventhandler.cpp
@@ -49,7 +49,7 @@ namespace CSMPrefs
         ShortcutMap::iterator shortcutListIt = mWidgetShortcuts.find(widget);
         if (shortcutListIt != mWidgetShortcuts.end())
         {
-            std::remove(shortcutListIt->second.begin(), shortcutListIt->second.end(), shortcut);
+            shortcutListIt->second.erase(std::remove(shortcutListIt->second.begin(), shortcutListIt->second.end(), shortcut), shortcutListIt->second.end());
         }
     }
 
--- a/apps/opencs/model/tools/pathgridcheck.cpp
+++ b/apps/opencs/model/tools/pathgridcheck.cpp
@@ -29,7 +29,7 @@ void CSMTools::PathgridCheckStage::perform (int stage, CSMDoc::Messages& message
     CSMWorld::UniversalId id (CSMWorld::UniversalId::Type_Pathgrid, pathgrid.mId);
 
     // check the number of pathgrid points
-    if (pathgrid.mData.mS2 > static_cast<int>(pathgrid.mPoints.size()))
+    if (pathgrid.mData.mS2 < static_cast<int>(pathgrid.mPoints.size()))
         messages.add (id, pathgrid.mId + " has less points than expected", "", CSMDoc::Message::Severity_Error);
     else if (pathgrid.mData.mS2 > static_cast<int>(pathgrid.mPoints.size()))
         messages.add (id, pathgrid.mId + " has more points than expected", "", CSMDoc::Message::Severity_Error);
--- a/apps/opencs/model/world/refcollection.cpp
+++ b/apps/opencs/model/world/refcollection.cpp
@@ -24,7 +24,7 @@ void CSMWorld::RefCollection::load (ESM::ESMReader& reader, int cellIndex, bool 
     bool isDeleted = false;
 
     // hack to initialise mindex
-    while (!(mref.mRefNum.mIndex == 0) && ESM::Cell::getNextRef(reader, ref, isDeleted, true, &mref))
+    while (!(mref.mRefNum.mIndex = 0) && ESM::Cell::getNextRef(reader, ref, isDeleted, true, &mref))
     {
         // Keep mOriginalCell empty when in modified (as an indicator that the
         // original cell will always be equal the current cell).
--- a/apps/opencs/model/world/refcollection.cpp
+++ b/apps/opencs/model/world/refcollection.cpp
@@ -24,7 +24,7 @@ void CSMWorld::RefCollection::load (ESM::ESMReader& reader, int cellIndex, bool 
     bool isDeleted = false;
 
     // hack to initialise mindex
-    while (!(mref.mRefNum.mIndex = 0) && ESM::Cell::getNextRef(reader, ref, isDeleted, true, &mref))
+    while (!(mref.mRefNum.mIndex == 0) && ESM::Cell::getNextRef(reader, ref, isDeleted, true, &mref))
     {
         // Keep mOriginalCell empty when in modified (as an indicator that the
         // original cell will always be equal the current cell).
--- a/apps/opencs/model/world/refidadapterimp.cpp
+++ b/apps/opencs/model/world/refidadapterimp.cpp
@@ -1373,10 +1373,8 @@ QVariant CSMWorld::CreatureAttackRefIdAdapter::getNestedData (const RefIdColumn 
 
     if (subColIndex == 0)
         return subRowIndex + 1;
-    else if (subColIndex < 3) // 1 or 2
+    else // 1 or 2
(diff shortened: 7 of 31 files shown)


apps/openmw/mwgui/journalwindow.cpp
```diff
@@ -558,7 +558,7 @@ namespace
             if (mAllQuests)
             {
                 SetNamesInactive setInactive(list);
-                mModel->visitQuestNames(!mAllQuests, setInactive);
+                mModel->visitQuestNames(false, setInactive);
```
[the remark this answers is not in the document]
Capostrophic: Anything under `if (mAllQuests)` is only run when mAllQuests is true.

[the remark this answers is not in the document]
tri4ng1e: `mAllQuests` is always `true` because of `if` check, so do we really need to compute NOT? Maybe compiler optimizes it itself, but...

apps/openmw/mwworld/cellstore.cpp
```diff
@@ -685,8 +685,8 @@ namespace MWWorld
     {
         state.mId = mCell->getCellId();
 
-        if (mCell->mData.mFlags & ESM::Cell::Interior && mCell->mData.mFlags & ESM::Cell::HasWater)
-            state.mWaterLevel = mWaterLevel;
+        // if (mCell->mData.mFlags & ESM::Cell::Interior && mCell->mData.mFlags & ESM::Cell::HasWater)
```
jvoisin: Why comments instead of deleting?

tri4ng1e: Read first message please, I added info about this. I'm just not sure.

apps/esmtool/record.cpp
```diff
@@ -694,7 +694,7 @@ void Record<ESM::Dialogue>::print()
     // loads, rather than loading and then dumping. :-( Anyone mind if
     // I change this?
     ESM::Dialogue::InfoContainer::iterator iit;
-    for (iit = mData.mInfo.begin(); iit != mData.mInfo.end(); iit++)
+    for (iit = mData.mInfo.begin(); iit != mData.mInfo.end(); ++iit)
```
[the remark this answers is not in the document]
tri4ng1e: It was marked as (performance) by cppcheck. `Prefer prefix ++/-- operators for non-primitive types.`

apps/opencs/model/prefs/boolsetting.cpp
```diff
@@ -22,6 +22,8 @@ CSMPrefs::BoolSetting& CSMPrefs::BoolSetting::setTooltip (const std::string& too
 
 std::pair<QWidget *, QWidget *> CSMPrefs::BoolSetting::makeWidgets (QWidget *parent)
 {
+    if (mWidget != nullptr)
+        delete mWidget;
```
psi29a: Does makeWidgets run more than once?

apps/esmtool/record.cpp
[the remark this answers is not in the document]
Capostrophic: cppcheck: (performance) Prefer prefix ++/-- operators for non-primitive types.
I dunno, seems reasonable to me.

apps/opencs/model/filter/parser.cpp
```diff
@@ -579,7 +579,7 @@ bool CSMFilter::Parser::parse (const std::string& filter, bool allowPredefined)
     }
     // We do not use isString() here, because there could be a pre-defined filter with an ID that is
     // equal a filter keyword.
-    else if (token.mType==Token::Type_String && allowPredefined)
+    else if (token.mType==Token::Type_String)
```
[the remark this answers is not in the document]
Capostrophic: allowPredefined is always true here because if it's false then the previous part is run (line 557).

apps/opencs/model/tools/pathgridcheck.cpp
```diff
@@ -29,7 +29,7 @@ void CSMTools::PathgridCheckStage::perform (int stage, CSMDoc::Messages& message
     CSMWorld::UniversalId id (CSMWorld::UniversalId::Type_Pathgrid, pathgrid.mId);
 
     // check the number of pathgrid points
-    if (pathgrid.mData.mS2 > static_cast<int>(pathgrid.mPoints.size()))
+    if (pathgrid.mData.mS2 < static_cast<int>(pathgrid.mPoints.size()))
```
psi29a: Nice catch! :)

apps/opencs/model/filter/parser.cpp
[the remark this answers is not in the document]
tri4ng1e: `allowPredefined` is always `true`. I don't know about changing the comment, but logically this part is exactly the same as before fix, but without unnecessary computations.

apps/esmtool/record.cpp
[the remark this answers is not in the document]
psi29a: Isn't that something we want the compiler handle for us? We shouldn't be worrying about these kinds of things.

apps/opencs/model/prefs/boolsetting.cpp
tri4ng1e: As I pointed in the first message, I'm stiil a newbie here. In my personal projects I'm very strict against memory allocations. Maybe in future there will be tons of makeWidgets calls. :smile: If there is a possibility of a memory leak, then it can happen sometimes.

apps/esmtool/record.cpp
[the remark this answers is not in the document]
tri4ng1e: I think cppcheck guys know their work better than me, so I just impemented it.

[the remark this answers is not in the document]
psi29a: Profile it! :)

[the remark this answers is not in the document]
tri4ng1e: From Quora
> Using a postfix operator on a non-primitive types may require the old value of the object to be copied into another location, which is both expensive and confusing. The compiler can probably optimize this away if the copy is unnecessary, but avoiding postfix operators eliminates this confusion, and eliminating uncertainty is almost always good.

apps/opencs/model/prefs/boolsetting.cpp
psi29a: If you run makeWidgets, then what are you planning to do with the old ones? Do we need to disconnect the widget first from Qt? Would that too be a memory leak to leave a connection to a deleted pointer?

kcat: FWIW, it's safe to call `delete` with a nullptr.
```
if(x != nullptr)
    delete x;
x = nullptr;
```
is a redundant null check.

apps/openmw/mwgui/widgets.hpp
```diff
@@ -315,9 +315,6 @@ namespace MWGui
             virtual void initialiseOverride();
             void repeatClick(MyGUI::Widget* _widget, MyGUI::ControllerItem* _controller);
 
-            bool mEnableRepeat;
-            float mRepeatTriggerTime;
-            float mRepeatStepTime;
```
scrawl: These members were added to MyGUI::ScrollBar in 3.2.2. OpenMW still supports MyGUI 3.2.1 which doesn't have them. Ergo, we can't remove this code. Could add an #if guard, but tbh I'd prefer leaving everything as is.

apps/opencs/model/world/refcollection.cpp
```diff
@@ -24,7 +24,7 @@ void CSMWorld::RefCollection::load (ESM::ESMReader& reader, int cellIndex, bool
     bool isDeleted = false;
 
     // hack to initialise mindex
-    while (!(mref.mRefNum.mIndex = 0) && ESM::Cell::getNextRef(reader, ref, isDeleted, true, &mref))
+    while (!(mref.mRefNum.mIndex == 0) && ESM::Cell::getNextRef(reader, ref, isDeleted, true, &mref))
```
scrawl: Should have read above: 'hack to initialise mindex'. Assignment is intended.

apps/openmw/mwmechanics/pathfinding.cpp
```diff
@@ -314,17 +314,14 @@ namespace MWMechanics
         {
             const ESM::Pathgrid::Point oldStart(*getPath().begin());
             buildPath(startPoint, endPoint, cell, pathgridGraph);
-            if (mPath.size() >= 2)
```
scrawl: This check doesn't look redundant to me because buildPath() modifies mPath at which point it could have a totally different size.

apps/openmw/mwscript/interpretercontext.cpp
```diff
@@ -406,7 +406,7 @@ namespace MWScript
         const MWWorld::ESMStore &store = world->getStore();
         const ESM::Faction *faction = store.get<ESM::Faction>().find(factionId);
 
-        if(rank < 0 || rank > 9)
+        if(rank < 0)
```
scrawl: I don't understand. Why can rank never be > 9 here?

apps/openmw/mwworld/cellstore.cpp
scrawl: I think the correct fix is to remove the unconditional state.mWaterLevel = mWaterLevel; . The loadState function above has the same issue btw.

apps/opencs/model/prefs/boolsetting.cpp
scrawl: AFAIK this is wrong you never delete any QWidget, they are deleted by their parent.

components/fontloader/fontloader.cpp
```diff
@@ -424,7 +424,7 @@ namespace Gui
             MyGUI::FontCodeType::Enum type;
             if(i == 0)
                 type = MyGUI::FontCodeType::Selected;
-            else if (i == 1)
+            else
```
scrawl: In cases like this can we at least leave the condition as a comment?

```else // if (i == 1)```

File: apps/openmw/mwscript/interpretercontext.cpp (hunk 10 above)
Comment (Capostrophic) [reply apps/openmw/mwscript/interpretercontext.cpp]: A speculation: due to these lines. https://github.com/tri4ng1e/openmw/blob/static-analysis/apps/openmw/mwscript/interpretercontext.cpp#L402

File: apps/opencs/model/prefs/boolsetting.cpp (hunk 4 above)
Comment (zinnschlag) [reply apps/opencs/model/prefs/boolsetting.cpp]: This change is indeed wrong. The mWidget variable does not own the Widget. It is managed outside of this class. Please revert.

File: apps/openmw/mwgui/journalwindow.cpp (hunk 1 above)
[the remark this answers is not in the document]
Comment (zinnschlag) [reply apps/openmw/mwgui/journalwindow.cpp]: Agreed. The original version is slightly easier to read and the compile should be able to optimise the variable access away.

File: apps/openmw/mwstate/quicksavemanager.cpp [hunk 12]
```diff
@@ -1,8 +1,8 @@
 #include "quicksavemanager.hpp"
 
 MWState::QuickSaveManager::QuickSaveManager(std::string &saveName, unsigned int maxSaves)
+  : mSaveName(saveName)
```
zinnschlag: This is a good change, because the initialiser list is preferable to initialising in the body of the constructor. However this change also seems to be incomplete, since there are several in-body initialisations left.

apps/wizard/unshield/unshieldworker.cpp
```diff
@@ -472,8 +472,8 @@ bool Wizard::UnshieldWorker::setupComponent(Component component)
 
                 if (morrowindFound) {
                     // Check if we have correct archive, other archives have Morrowind.bsa too
-                    if ((tribunalFound && bloodmoonFound)
-                            || (!tribunalFound && !bloodmoonFound)) {
+                    if (tribunalFound == bloodmoonFound)
```
zinnschlag: Nice catch!

apps/opencs/model/prefs/shortcuteventhandler.cpp
```diff
@@ -49,7 +49,7 @@ namespace CSMPrefs
         ShortcutMap::iterator shortcutListIt = mWidgetShortcuts.find(widget);
         if (shortcutListIt != mWidgetShortcuts.end())
         {
-            std::remove(shortcutListIt->second.begin(), shortcutListIt->second.end(), shortcut);
+            shortcutListIt->second.erase(std::remove(shortcutListIt->second.begin(), shortcutListIt->second.end(), shortcut), shortcutListIt->second.end());
```
zinnschlag: This looks like another Qt-ownership issue.

apps/opencs/model/world/refidadapterimp.cpp
```diff
@@ -1373,10 +1373,8 @@ QVariant CSMWorld::CreatureAttackRefIdAdapter::getNestedData (const RefIdColumn
 
     if (subColIndex == 0)
         return subRowIndex + 1;
-    else if (subColIndex < 3) // 1 or 2
+    else // 1 or 2
```
zinnschlag: I would prefer to keep the condition. If there every arises the need to add more columns the removal of the condition because a possible source of a bug.

apps/openmw/mwgui/widgets.cpp
```diff
@@ -532,12 +532,11 @@ namespace MWGui
...
-        MWScrollBar::MWScrollBar()
-            : mEnableRepeat(true)
-            , mRepeatTriggerTime(0.5f)
-            , mRepeatStepTime(0.1f)
-            , mIsIncreasing(true)
+        MWScrollBar::MWScrollBar() : mIsIncreasing(true)
         {
+            mEnableRepeat = true;
+            mRepeatTriggerTime = 0.5f;
+            mRepeatStepTime = 0.1f;
```
zinnschlag: This change looks pointless.

apps/opencs/model/prefs/shortcuteventhandler.cpp
kcat: `std::remove` doesn't actually remove anything, it just moves the matching entries to the end of the container and returns an iterator to the first of the entries to remove. It would seem quite odd if this intended to only move the given widget's shortcuts to the end of the container but leave them in it. In either case, this isn't deleting anything, just removing the pointer(s) from the widget's map entry.

apps/openmw/mwscript/interpretercontext.cpp
jvoisin: Because of this, at line 403-404.

```C++
         // if we are already at max rank, there is no next rank
         if (rank > 9)
             rank = 9;
```

apps/opencs/model/prefs/shortcuteventhandler.cpp
zinnschlag: Hm, right. I misread that.

apps/openmw/mwgui/widgets.cpp
tri4ng1e: This variables were in parent MyGUI class and duplicated in MWScrollBar class. I removed them, and since I can't init them with member init-list anymore, I did it in ctor. Now I'm reverting this because of back-compatibility, as @scrawl mentioned.

tri4ng1e:
I updated PR with some reverts and changes.

scrawl:
Merged, just rebase did something weird so the PR isnt auto closing for some reason.

PR #1682, opened 2018-04-17
Fix build with Qt 5.11_beta3 (dropping qt5_use_modules)

a17r:


psi29a:
Is this backwards compatible with older Qt5 releases?

a17r:
Yes, it is perfectly fine with previous versions of Qt5.

psi29a:
Great, thank you! :)

PR #1683, opened 2018-04-18
Force quick load to use the last quicksave instead of any last save

akortunov:
Fixes [bug #4394](https://bugs.openmw.org/issues/4394).

Just search for the first save with the "Quicksave" desctiption for current character and load it. If there are no quick saves, do not load anything.

### DIFF
--- a/apps/openmw/mwstate/statemanagerimp.cpp
+++ b/apps/openmw/mwstate/statemanagerimp.cpp
@@ -557,11 +557,20 @@ void MWState::StateManager::loadGame (const Character *character, const std::str
 
 void MWState::StateManager::quickLoad()
 {
-    if (Character* currentCharacter = getCurrentCharacter ())
+    if (const Character* currentCharacter = getCurrentCharacter())
     {
-        if (currentCharacter->begin() == currentCharacter->end())
+        // Search for a newest quick save
+        Character::SlotIterator it;
+        for (it = currentCharacter->begin(); it != currentCharacter->end(); ++it)
+        {
+            if (it->mProfile.mDescription == "Quicksave")
+                break;
+        }
+
+        if (it == currentCharacter->end())
             return;
-        loadGame (currentCharacter, currentCharacter->begin()->mPath.string()); //Get newest save
+
+        loadGame (currentCharacter, it->mPath.string());
     }
 }
 


Capostrophic:
Well, Morrowind and other Bethesda games quickload behavior is to load the quicksave, and since it's a reference point OpenMW should follow...

kcat:
When it comes to gameplay and general presentation, sure. But OpenMW does make the odd quality-of-life improvement (see the alchemy window, or options menu). If having Quickload load the last save improves the user experience, why purposely change it to make it a worse experience?

Capostrophic:
The user that reported this to akortunov was displeased with the current behavior. Such controversial divergencies should still be kept to a minimum - and if Morrowind behavior is followed instead, we are not to blame.

akortunov:
> I gave a reason why the current behavior is preferable. 

Hmm... Where? This is hardly a QoL improvement, there are just two different approaches for quick saves. One games use first, other games - last.

> Is the new behavior the actual standard most other games follow? Is the new behavior more widely favored among the player-base?

Well, we can also ask these questions about current behaviour.
What we can do here:
1. Check how quick save behaves in Bethesda games
2. Make changes optional

kcat:
> Hmm... Where?

"If the last save I made was a hard save (either because something important happened or because I stopped playing then started again), if I Quickload I'd expect the latest save to be loaded, not necessarily the latest quicksave which may have been farther back."

A quickload is just that: load quickly. Same as a quicksave is save quickly. It's not its own separate system, it's just quick access to the existing system (you can manually load a quick save, so why not quickly load a manual save if it's the latest one)? If I have to remember if the last save I made was with the quicksave function or was manually made, that unnecessarily complicates the system and slows it down (as I'll fallback to the manual load menu just to be sure).

Capostrophic:
>Check how quick save behaves in Bethesda's games

It's behavior is consistent across the games. Even Skyrim and Fallout load the *quick*save when the quickload button is pressed.

scrawl:
Aside from kcat's argument which I agree with, changing the behavior now would be disruptive in that openmw users have to get used to the change. Wouldn't be surprised if someone would then report another bug on that.
The fact that it took years for anyone to notice the difference to bethesda and report the bug means that either no one noticed, no one cared or they actually prefer the new behavior.
If you want to change my mind on this, I suggest to make a user poll.

PR #1685, opened 2018-04-21
Fix crash when rollDice is called with 0.

kurnevsky:
After introducing new random number generator rollDice function crashes if called with 0. I fixed these places (not sure though if it's possible for iNumberCreatures to be 0) and simplified code a bit.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -286,7 +286,7 @@ void CharacterController::refreshHitRecoilAnims()
         }
         else if (recovery)
         {
-            std::string anim = isSwimming ? chooseRandomGroup("swimhit") : chooseRandomGroup("hit");
+            std::string anim = chooseRandomGroup("swimhit");
             if (isSwimming && mAnimation->hasAnimation(anim))
             {
                 mHitState = CharState_SwimHit;
--- a/components/misc/rng.cpp
+++ b/components/misc/rng.cpp
@@ -25,7 +25,7 @@ namespace Misc
 
     int Rng::rollDice(int max)
     {
-        return std::uniform_int_distribution<int>(0, max - 1)(generator);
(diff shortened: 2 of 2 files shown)


elsid:
My suggest is to do it much simplier. Just change [rollDice](https://github.com/OpenMW/openmw/blob/48467814d4745cfe27835bd59cd1be6a98f29d58/components/misc/rng.cpp#L26-L29) function to restore behavior as it was before 48467814d4745cfe27835bd59cd1be6a98f29d58 .
```c++
int Rng::rollDice(int max)
{
        return max > 0 ? std::uniform_int_distribution<int>(0, max - 1)(generator) : 0;
}
```
There are some other places where crashes could be:
https://github.com/OpenMW/openmw/blob/7a389f7ab117acad15d813a33f63877f45da98ef/components/interpreter/miscopcodes.hpp#L167-L173 (`limit` could be 0)
https://github.com/OpenMW/openmw/blob/d310d36ea3940f3feff1a0de20353246352d811c/apps/openmw/mwworld/worldimp.cpp#L3282-L3283 (`iNumberCreatures` could be 0)

kurnevsky:
ok, restored previous behavior. And just one more thing - removed useless comparison.

Capostrophic:
That part of refreshRecoilHitAnims could be simplified further.
``` cpp
std::string anim = isSwimming ? chooseRandomGroup("swimhit") : chooseRandomGroup("hit");
if (mAnimation->hasAnimation(anim))
{
    mHitState = isSwimming ? CharState_SwimHit : CharState_Hit;
    mCurrentHit = anim;
    mAnimation->play(mCurrentHit, Priority_Hit, MWRender::Animation::BlendMask_All, true, 1, "start", "stop", 0.0f, 0);
}
```

kurnevsky:
@Capostrophic it can't be simplified in the way you proposed because currently if there is no swimming animation usual will be used.

PR #1686, opened 2018-04-21
Don't transition to empty music playlists (bug #4134)

Capostrophic:
As per kcat's suggestion in #1655. Seems to work fine when I delete the battle music folder or its contents. Explore playlist simply continues to play without any interruptions, which is probably better than both the master and vanilla behavior.

I moved the initialization of playlist file list from startRandomTitle to playPlaylist and added two checks for the cases the playlist is changed to the same playlist or is empty even after its guaranteed initialization.

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -393,32 +393,8 @@ namespace MWSound
 
     void SoundManager::startRandomTitle()
     {
-        std::vector<std::string> filelist;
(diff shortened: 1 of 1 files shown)


Capostrophic:
Oops, I didn't add a failsafe for the case the playlist didn't change (which could happen at the end of combat).

scrawl:
I don't like that this fix duplicates code and it depends on the mMusicFiles not being initialized yet. Start again please.

Capostrophic:
Alright. I'll try another thing then.

Capostrophic:
Is this better?

PR #1689, opened 2018-04-26
Implicit Morrowind.esm dependency for dependency-less content addons (bug #2829)

Capostrophic:
Add a hack to content selector dependency sort: dependency-less content addons that are not game files are assumed dependent on Morrowind.esm and are sorted after it.

### DIFF
--- a/components/contentselector/model/contentmodel.cpp
+++ b/components/contentselector/model/contentmodel.cpp
@@ -512,7 +512,9 @@ void ContentSelectorModel::ContentModel::sortFiles()
             //dependencies appear.
             for (int j = i + 1; j < fileCount; j++)
             {
-                if (gamefiles.contains(mFiles.at(j)->fileName(), Qt::CaseInsensitive))
+                if (gamefiles.contains(mFiles.at(j)->fileName(), Qt::CaseInsensitive)
+                 || (!mFiles.at(i)->isGameFile() && gamefiles.isEmpty()
+                 && mFiles.at(j)->fileName().compare("Morrowind.esm", Qt::CaseInsensitive) == 0)) // Hack: implicit dependency on Morrowind.esm for dependency-less files
                 {
                         mFiles.move(j, i);
 


akortunov:
How does file sorting work? According to file modify time? In this case it will break manual mods ordering.

zinnschlag:
From what I remember we decided not to support plugins without dependencies. After 1.0 we most likely would have to drop support anyway.

>  dependencies are always supposed to load before the content files that depend on them

This is correct. Still, we should not automatically reorder the content files (unless the user explicitly asks for it).

Capostrophic:
> How does file sorting work? According to file modify time? In this case it will break manual mods ordering.

No, it won't. sortFiles() doesn't "re-sort" files in this case. If a content file with a specific dependency is loaded before its dependency for any reason, sortFiles() changes their order (but it doesn't mess up the manual ordering that didn't have any mistakes). If not, nothing happens, and it's not concerned with file dates, so it works even if you deliberately mess them up before e.g. using ini-importer.
If it didn't work like this, it wouldn't have resolved 3926.

>After 1.0 we most likely would have to drop support anyway.

Any particular reason for that?

>Still, we should not automatically reorder the content files

Well, the user will proceed to do the exactly the same thing this workaround is handling for them, as the game won't launch if the dependencies are loaded after the plugins that depend on them. Why not remove some headache?

zinnschlag:
> Well, the user will proceed to do the exactly the same thing this workaround is handling for them, as the game won't launch if the dependencies are loaded after the plugins that depend on them. Why not remove some headache?

We already indicate if there is a known problem with the content file order. That is sufficient in my opinion.

> Any particular reason for that?

A plugin without dependency can (legally) do very little, because it can not reference anything in another content file. That isn't a complete show stopper with vanilla MW, because there are still some hardcoded parts that could be interacted with. But with the de-hardcoding after 1.0 most of that will go away, which means using a dependency-less plugin with an omwgame file that takes full advantage of the de-hardcoding and is noticeably different from Morrowind.esm will most likely break.

I see only one way to maintain the support of dependency-less plugins: assume an implicit dependency on Morrowind.esm. This will avoid future problems resulting from de-hardcoding. But it is kinda a hack and I am also not sure if it is all that useful.

Capostrophic:
Hm. Should bug 2829 be rejected then?

But alright. I currently cooperate with tri4ng1e on fixing the load order supplied by the ini importer for good so that bug 3926 is actually fixed, not worked around. I just was worried that considering dependency sort was already available in contentmodel adding new sorting was redundant and was trying out ways to employ it at work. ini importer should really not bypass the content selector when creating a content file list if that's even possible...

zinnschlag:
> Hm. Should bug 2829 be rejected then?

Not entirely sure. The more I think about it the more I like the implicit Morrowind.esm dependency solution. We are unlikely to get dependency-less files for anything other than Morrowind.esm anyway.

scrawl:
Can you give an example e.g. before and after screenshots? My memory of the bug is fading and the bug description is a little vague.

Why does this need to include 'Morrowind.esm' btw? Won't total conversions remain unfixed then?

Capostrophic:
I checked for whether there is a game file after a non-game one earlier, but apparently such fix would have to be removed post-1.0 anyway and Zini asked for a Morrowind.esm-dependent hack instead (even though it also would have to be removed under that logic).

When a content list is created directly through the content selector, sortFiles sorts all the files found in the data folders based on their dependencies. But if a dependency-less plugin that is not a game file is placed before the main game file, it is ignored during that sorting, so the final content list is incorrect.

zinnschlag:
> Why does this need to include 'Morrowind.esm' btw? Won't total conversions remain unfixed then?

The problem with the universal plugin is that it that we can not guarantee that it will work with TCs that take advantage of the post 1.0 de-hardcoding and without any dependency information there is no way for the launcher to detect that a plugin is unsuitable for a TC.

Note that we are only dealing with legacy content files here and it seems to be a reasonable assumption that that most (if not all of them) are aimed at Morrowind.esm. For future content files that is not an issue, because OpenMW-CS does not allow the creation of dependency-less addon files.

components/contentselector/model/contentmodel.cpp
```diff
@@ -512,7 +512,9 @@ void ContentSelectorModel::ContentModel::sortFiles()
             //dependencies appear.
             for (int j = i + 1; j < fileCount; j++)
             {
-                if (gamefiles.contains(mFiles.at(j)->fileName(), Qt::CaseInsensitive))
+                if (gamefiles.contains(mFiles.at(j)->fileName(), Qt::CaseInsensitive)
+                 || (!mFiles.at(i)->isGameFile() && gamefiles.isEmpty()
+                 && mFiles.at(j)->fileName().compare("Morrowind.esm", Qt::CaseInsensitive) == 0)) // Hack: implicit dependency on Morrowind.esm for dependency-less files
```
zinnschlag: This line looks odd, considering that Morrowind.esm can only be ever found at index 0.

Capostrophic: Uh... during the sorting it's not necessarily at index 0. Aren't the files added alphabetically in addFiles? And isn't the "point" of the bug that masterless plugins are ignored during the sorting and never get ordered after Morrowind.esm if they happen to be placed earlier than it in the order sortFiles() got?

zinnschlag: Hm, isn't that part of the problem? I haven't looked at the code recently, but the omwgame file (which is the type a dependency less esm should be mapped to) must always be at index 0. Maybe that isn't a problem, because it is sorted to the correct position later. But this surely looks odd.

zinnschlag:
I'll go ahead and merge this anyway, because it doesn't break anything that isn't already broken and provides a useful workaround. But I am starting to get the feeling that at some point we need to take a fresh look at the whole content list handling.

PR #1690, opened 2018-04-29
Handle NiLookAtController

akortunov:
Implements [feature #4407](https://bugs.openmw.org/issues/4407).
Some models in Arktwend have NiLookAtController (I guess, imported ones).
According to @Capostrophic, vanilla engine just ignores this node.

With this PR user will get a warning about unhandled controller, but at least meshes will be rendered.

### DIFF
--- a/components/nif/controller.cpp
+++ b/components/nif/controller.cpp
@@ -101,6 +101,18 @@ namespace Nif
         data.post(nif);
     }
 
+    void NiLookAtController::read(NIFStream *nif)
+    {
+        Controller::read(nif);
+        data.read(nif);
+    }
+
+    void NiLookAtController::post(NIFFile *nif)
+    {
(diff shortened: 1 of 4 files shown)


Capostrophic:
This might be the same situation as is with NiPathController, for which there are no signs of being used for any purpose in Morrowind even though it ends up in some meshes.

psi29a:
Since these things are no-ops, In such situations shouldn't we let someone know that while OpenMW can read these, they aren't wired up to do anything. In this case "Handled" is a misnomer. ;)

scrawl:
> In such situations shouldn't we let someone know that while OpenMW can read these, they aren't wired up to do anything

Erm, so it does, supposedly:

> With this PR user will get a warning about unhandled controller, but at least meshes will be rendered.

PR #1692, opened 2018-05-01
Prevent PlaySound overlapping

akortunov:
If we play a new sound, we should stop the previous copy of this sound. This is easily to check: execute "playsound X" several times in console in vanilla game.
We already use similar approach for 3D sounds: we allow to play only one copy of given sound per object, so this PR just makes sound playing consistent.

As a side effect, this PR fixes a nightmare scene in Arktwend, so it should not crash now.

### DIFF
--- a/apps/openmw/mwgui/windowmanagerimp.cpp
+++ b/apps/openmw/mwgui/windowmanagerimp.cpp
@@ -1919,16 +1919,12 @@ namespace MWGui
             mInventoryWindow->cycle(next);
     }
 
-    void WindowManager::playSound(const std::string& soundId, bool preventOverlapping, float volume, float pitch)
+    void WindowManager::playSound(const std::string& soundId, float volume, float pitch)
     {
         if (soundId.empty())
             return;
 
-        MWBase::SoundManager *sndmgr = MWBase::Environment::get().getSoundManager();
-        if (preventOverlapping && sndmgr->getSoundPlaying(MWWorld::Ptr(), soundId))
-            return;
-
-        sndmgr->playSound(soundId, volume, pitch, MWSound::Type::Sfx, MWSound::PlayMode::NoEnv);
+        MWBase::Environment::get().getSoundManager()->playSound(soundId, volume, pitch, MWSound::Type::Sfx, MWSound::PlayMode::NoEnv);
     }
 
     void WindowManager::updateSpellWindow()
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -337,7 +337,7 @@ namespace MWBase
             /// Cycle to next or previous weapon
             virtual void cycleWeapon(bool next) = 0;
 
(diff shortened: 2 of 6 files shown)


psi29a:
I'm OK with this, but it touches on sound so i'll defer to @kcat

apps/openmw/mwgui/windowmanagerimp.cpp
```diff
@@ -1919,16 +1919,12 @@ namespace MWGui
...
     {
         if (soundId.empty())
             return;
 
-        MWBase::SoundManager *sndmgr = MWBase::Environment::get().getSoundManager();
-        if (preventOverlapping && sndmgr->getSoundPlaying(MWWorld::Ptr(), soundId))
-            return;
-
-        sndmgr->playSound(soundId, volume, pitch, MWSound::Type::Sfx, MWSound::PlayMode::NoEnv);
+        MWBase::Environment::get().getSoundManager()->playSound(soundId, volume, pitch, MWSound::Type::Sfx, MWSound::PlayMode::NoEnv);
```
kcat: Won't this change now cause the UI sound to get cutoff if it's played again while already playing, rather than the apparently intended behavior of letting the original sound finish without interruption?

kcat:
My only other real complaint is that it basically does `loadSound(Misc::StringUtils::lowerCase(soundId));` twice when playing a sound now. First for `playSound`, then again in `stopSound`.

apps/openmw/mwgui/windowmanagerimp.cpp
akortunov: Should we treat GUI sound differently from other ones?

psi29a: How much work would it be to treat GUI sounds differently?

akortunov: > How much work would it be to treat GUI sounds differently?

Almost none. We just should decide which behaviour is intended:
1. Allow to stack most of GUI sounds, with special case for journal wheel scrolling to prevent overlapping.
It is our current upstream implementation.
2. Play old copy to end and do not launch new copy to prevent overlapping.
3. Take the PR implementation.

Honestly, I can not hear much difference between all three variants (excepts of loudness in first case), so I can not advice which behaviour is better.

akortunov:
> My only other real complaint is that it basically does loadSound(Misc::StringUtils::lowerCase(soundId)); twice when playing a sound now. First for playSound, then again in stopSound.

How do you suggest to avoid it?

akortunov:
I added a new stopSound function, which accept Sound_Buffer.

akortunov:
> My only other real complaint is that it basically does loadSound(Misc::StringUtils::lowerCase(soundId)); twice when playing a sound now. First for playSound, then again in stopSound.

Should be fixed now. Any other issues?

psi29a:
@kcat mergable? :)

kcat:
Looks alright at a glance. Assuming it works and no one else has objections, go for it.

psi29a:
Merged, this should make Arktwend fans happy. :)
Thanks for the PR!

PR #1693, opened 2018-05-02
Fix undefined behaviour in CreatureAttackRefIdAdapter::getNestedData

mimi1vx:
bug #4409

### DIFF
--- a/apps/opencs/model/world/refidadapterimp.cpp
+++ b/apps/opencs/model/world/refidadapterimp.cpp
@@ -1368,13 +1368,15 @@ QVariant CSMWorld::CreatureAttackRefIdAdapter::getNestedData (const RefIdColumn 
 
     const ESM::Creature& creature = record.get();
 
-    if (subRowIndex < 0 || subRowIndex > 2 || subColIndex < 0 || subColIndex > 2)
-        throw std::runtime_error ("index out of range");
-
     if (subColIndex == 0)
         return subRowIndex + 1;
     else if (subColIndex < 3) // 1 or 2
         return creature.mData.mAttack[(subRowIndex * 2) + (subColIndex - 1)];
+
+    if (subRowIndex < 0 || subRowIndex > 2 || subColIndex < 0 || subColIndex > 2)
+        throw std::runtime_error ("index out of range");
+    
+    return 0;
 }
 
 void CSMWorld::CreatureAttackRefIdAdapter::setNestedData (const RefIdColumn *column,


apps/opencs/model/world/refidadapterimp.cpp
```diff
@@ -942,7 +942,8 @@ QVariant CSMWorld::NpcAttributesRefIdAdapter::getNestedData (const RefIdColumn *
     const ESM::NPC::NPDTstruct52& npcStruct = record.get().mNpdt52;
 
     if (subColIndex == 0)
-        return subRowIndex;
+        return subRowIndex;  if (subRowIndex < 0 || subRowIndex > 2 || subColIndex < 0 || subColIndex > 2)
```
Capostrophic: What are you trying to do here?

mimi1vx: typo .. github integrated editor, I now pushing without this unwanted change

apps/opencs/model/world/refidadapterimp.cpp
```diff
@@ -1368,13 +1369,13 @@ QVariant CSMWorld::CreatureAttackRefIdAdapter::getNestedData (const RefIdColumn
...
 
-    if (subRowIndex < 0 || subRowIndex > 2 || subColIndex < 0 || subColIndex > 2)
-        throw std::runtime_error ("index out of range");
-
     if (subColIndex == 0)
         return subRowIndex + 1;
     else if (subColIndex < 3) // 1 or 2
         return creature.mData.mAttack[(subRowIndex * 2) + (subColIndex - 1)];
+    
+    if (subRowIndex < 0 || subRowIndex > 2 || subColIndex < 0 || subColIndex > 2)
```
Capostrophic: What's the difference?

zinnschlag:
Can you explain what this change is about? Where is there undefined behaviour?

The end of the function seems nonsensical to me. If the if construct fails to return (because subColIndex is out of range) the if condition below (for the exception) will always be true, which means the if statement is pointless and the final return 0 will never be reached.

mimi1vx:
@zinnschlag this isn't true. Throw exception can be optimized out by compiler... And in this case if .. else if and what after this two conditions?

zinnschlag:
> Throw exception can be optimized out by compiler.

Nope. They can't. The optimizer is not allowed to change the resulting behaviour of the code.

>  And in this case if .. else if and what after this two conditions?

The first if else block (line 1371-1374) will check if subColIndex is 0, 1 or 2. If any of this is the case a return statement will be executed and the function does not continue. Meaning if the function continues, subColIndex can not be 0, 1 or 2, which then results in the second if statement always being executed and the final return 0 can never be reached.

zinnschlag:
I noticed the warning message when compiling this file. I guess that is what triggered this PR?

The warning is a false positive. The code is perfectly fine and apparently the compiler is not smart enough to notice that.

But we should try to silence the warning anyway. I suggest the following approach:
* Split the Index check into two separate checks for row and column (included adjusted exception messages)
* Then move the column check into an else clause below line 1377.
* Since we already checked the column index, there is no need for another if statement here and you can just throw the exception in the else branch. That will hopefully silence the compiler and is also a bit more robust than the original code.

I am closing this PR since it goes into the wrong direction. Feel free to make a new one.

mimi1vx:
@zinnschlag  

[gcc's -f-no-exceptions](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_exceptions.html)

and in newer compiler this warning will be fatal

zinnschlag:
Exceptions are mandatory for OpenMW. The compiler is not supposed to produce a working binary when invoked with -f-no-exceptions.

PR #1694, opened 2018-05-03
Load default markers definitions

akortunov:
Fixes [bug #4410](https://bugs.openmw.org/issues/4410).
If markers were not found in game files, load default definitions to ESM store.

### DIFF
--- a/apps/openmw/mwworld/store.cpp
+++ b/apps/openmw/mwworld/store.cpp
@@ -1070,12 +1070,10 @@ namespace MWWorld
             if (search(marker.first) == 0)
             {
                 ESM::Static newMarker = ESM::Static(marker.first, marker.second);
-                mStatic.insert(std::make_pair(marker.first, newMarker));
-
-                std::map<std::string, ESM::Static>::iterator found = mStatic.find(marker.first);
-                if (found != mStatic.end())
+                std::pair<typename Static::iterator, bool> ret = mStatic.insert(std::make_pair(marker.first, newMarker));
+                if (ret.first != mStatic.end())
                 {
-                    mShared.push_back(&found->second);
+                    mShared.push_back(&ret.first->second);
                 }
             }
         }
--- a/apps/openmw/mwworld/store.cpp
+++ b/apps/openmw/mwworld/store.cpp
@@ -1053,6 +1053,34 @@ namespace MWWorld
         }
     }
 
+    template<>
+    void Store<ESM::Static>::setUp()
+    {
+        // Load default marker definitions, if game files do not have them for some reason
+        std::pair<std::string, std::string> markers[] = {
+            std::make_pair("divinemarker", "marker_divine.nif"),
+            std::make_pair("doormarker", "marker_arrow.nif"),
+            std::make_pair("northmarker", "marker_north.nif"),
+            std::make_pair("templemarker", "marker_temple.nif"),
+            std::make_pair("travelmarker", "marker_travel.nif")
+        };
+
+        for (const std::pair<std::string, std::string> marker : markers)
+        {
+            if (search(marker.first) == 0)
+            {
+                ESM::Static newMarker = ESM::Static(marker.first, marker.second);
+                mStatic.insert(std::make_pair(marker.first, newMarker));
+
+                std::map<std::string, ESM::Static>::iterator found = mStatic.find(marker.first);
+                if (found != mStatic.end())
+                {
+                    mShared.push_back(&found->second);
+                }
+            }
+        }
+    }
+
     template <>
     inline RecordId Store<ESM::Dialogue>::load(ESM::ESMReader &esm) {
         // The original letter case of a dialogue ID is saved, because it's printed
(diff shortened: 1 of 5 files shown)


apps/openmw/mwworld/store.cpp
```diff
@@ -1057,44 +1057,23 @@ namespace MWWorld
...
-        if (ptr == 0)
+        std::pair<std::string, std::string> markers[] = {
+            std::make_pair("divinemarker", "marker_divine.nif"),
+            std::make_pair("doormarker", "marker_arrow.nif"),
+            std::make_pair("northmarker", "marker_north.nif"),
+            std::make_pair("templemarker", "marker_temple.nif"),
+            std::make_pair("travelmarker", "marker_travel.nif")
+        };
+
+        for (std::pair<std::string, std::string> marker : markers)
```
kcat: This will copy both strings of each element since it's taken by value. You can do `const auto &marker : markers` instead to take a reference to each pair.

akortunov: Done.

zinnschlag:
Did you check if the same fix also needs to be applied to OpenMW-CS?

akortunov:
> Did you check if the same fix also needs to be applied to OpenMW-CS?

OpenMW-CS shows DivineMarker as white cube in cell and does not complain about errors, even when there is no such object under Objects tab.

apps/openmw/mwworld/store.cpp
```diff
@@ -1053,6 +1053,30 @@ namespace MWWorld
...
+            std::make_pair("travelmarker", "marker_travel.nif")
+        };
+
+        for (const std::pair<std::string, std::string> marker : markers)
+        {
+            const ESM::Static *ptr = search(marker.first);
+            if (ptr == 0)
+            {
+                ESM::Static *newMarker = new ESM::Static(marker.first, marker.second);
+                mStatic.insert(std::make_pair(marker.first, *newMarker));
```
zinnschlag: Memory leak! You shouldn't use new here.

apps/openmw/mwworld/store.cpp
```diff
@@ -1053,6 +1053,30 @@ namespace MWWorld
...
+            std::make_pair("doormarker", "marker_arrow.nif"),
+            std::make_pair("northmarker", "marker_north.nif"),
+            std::make_pair("templemarker", "marker_temple.nif"),
+            std::make_pair("travelmarker", "marker_travel.nif")
+        };
+
+        for (const std::pair<std::string, std::string> marker : markers)
+        {
+            const ESM::Static *ptr = search(marker.first);
+            if (ptr == 0)
```
zinnschlag: And while we are doing random code improvements, here is another one: These two lines could be written in a less verbose form:

if (!search (marker.first))

zinnschlag:
The white cube is how OpenMW-CS handles missing meshes. I guess that is okay-ish. But it wouldn't hurt to provide the same fix for OpenMW-CS.

akortunov:
> But it wouldn't hurt to provide the same fix for OpenMW-CS.

Is there a common ESM-loading code where I can put markers initializaton to avoid code duplication?

apps/openmw/mwworld/store.cpp
akortunov: I had to rewrite iteration since I get the "bad alloc" error without "new".
Now I put the structure to mStatic, get a persisted pointer from mStatic and put it to mShared.
Is this approach OK?

zinnschlag:
Unfortunately not, since OpenMW and OpenMW-CS use different data structures to store records. Would you like to try it anyway or should we go ahead with the merge?

apps/openmw/mwworld/store.cpp
```diff
@@ -1053,6 +1053,34 @@ namespace MWWorld
...
+            std::make_pair("templemarker", "marker_temple.nif"),
+            std::make_pair("travelmarker", "marker_travel.nif")
+        };
+
+        for (const std::pair<std::string, std::string> marker : markers)
+        {
+            if (search(marker.first) == 0)
+            {
+                ESM::Static newMarker = ESM::Static(marker.first, marker.second);
+                mStatic.insert(std::make_pair(marker.first, newMarker));
```
zinnschlag: For future reference, insert returns an interator (actually a pair with first being the iterator) to the new element. Therefore the find below is unnecessary.

akortunov:
> Would you like to try it anyway or should we go ahead with the merge?

I want to try to fix it.

akortunov:
Looks like the end of data loading is [here](https://github.com/OpenMW/openmw/blob/master/apps/opencs/model/world/data.cpp#L986).
Do I need to add some records to mReferenceables somehow?

zinnschlag:
> Do I need to add some records to mReferenceables somehow?

Correct. But be careful there. Multi-Threading involved.

akortunov:
I managed to add entries to editor:
screenshot_20180507_222231

OpenMW-CS now considers default markers as entries from base file:
screenshot_20180507_222007
Verifier now does not complain about missing entries.

PR #1695, opened 2018-05-03
Don't hide topic/journal info ID columns (bug #4356)

Capostrophic:
[Report.](https://bugs.openmw.org/issues/4356)
So usually records have their full IDs shown in a column to the left of the state column in their tables. Journal and topic info records don't, and so do land texture records (but displaying their IDs would have been redundant in their case). Apparently their respective ID columns were hidden, making their full IDs only available from their info tables. I enabled those columns, as was requested.


zinnschlag:
Closing this one because this is not a bug and a proper fix for the problem requires more than just unhiding the ID-column. I also updated the issue.

PR #1696, opened 2018-05-04
Rename "Model" column to "Model/Animation" (feature #2694)

Capostrophic:
[Report.](https://bugs.openmw.org/issues/2694)

Yep. Just that.


psi29a:
LGTM

PR #1697, opened 2018-05-05
Improve sorting of imported load order (bug #3926)

tri4ng1e:
This PR changes sorting in iniimporter. It sorts content files from Morrowind.ini by timestamp and then tries to resolve dependency problems. And it hard-codes Morrowind - Tribunal - Bloodmoon order (if all 3 elements are in list) because Bloodmoon doesn't have a hard dependency on Tribunal.

I used a variation of depth-first topological sort to place dependencies before files that use them.

[Original bug report](https://bugs.openmw.org/issues/3926)

Also there was a little bug with `ESMReader::close` - mHeader was not cleared. So I added `mHeader.blank()` inside it.

### DIFF
--- a/apps/mwiniimporter/importer.cpp
+++ b/apps/mwiniimporter/importer.cpp
@@ -3,6 +3,7 @@
 #include <iostream>
 #include <sstream>
 #include <components/misc/stringops.hpp>
+#include <components/esm/esmreader.hpp>
 
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
@@ -824,33 +825,73 @@ void MwIniImporter::importArchives(multistrmap &cfg, const multistrmap &ini) con
     }
 }
 
-void MwIniImporter::importGameFiles(multistrmap &cfg, const multistrmap &ini, const boost::filesystem::path& iniFilename) const {
-    std::vector<std::pair<std::time_t, std::string> > contentFiles;
+void MwIniImporter::dependencySortStep(std::string& el, MwIniImporter::deplist& src, std::vector<std::string>& ret)
+{
+    auto it = std::find_if(src.begin(), src.end(), [&el](std::pair< std::string, std::vector<std::string> >& o)
+    {
+        return o.first == el;
+    });
+    if (it != src.end())
+    {
+        auto o = std::move(*it);
+        src.erase(it);
+        for (auto name : o.second)
+        {
+            MwIniImporter::dependencySortStep(name, src, ret);
+        }
+        ret.push_back(std::move(o.first));
+    }
+}
+
+std::vector<std::string> MwIniImporter::dependencySort(MwIniImporter::deplist src)
+{
+    std::vector<std::string> ret;
+    while (!src.empty())
+    {
+        MwIniImporter::dependencySortStep(src.begin()->first, src, ret);
+    }
+    return ret;
+}
+
+std::vector<std::string>::iterator MwIniImporter::findString(std::vector<std::string>& v, const std::string& s)
+{
+    return std::find_if(v.begin(), v.end(), [&s](const std::string& str)
+    {
+        return Misc::StringUtils::ciEqual(str, s);
+    });
+}
+
+void MwIniImporter::importGameFiles(multistrmap &cfg, const multistrmap &ini, const boost::filesystem::path& iniFilename) const
+{
+    std::vector<std::pair<std::time_t, boost::filesystem::path>> contentFiles;
     std::string baseGameFile("Game Files:GameFile");
     std::string gameFile("");
     std::time_t defaultTime = 0;
+    ToUTF8::Utf8Encoder encoder(mEncoding);
 
     // assume the Game Files are all in a "Data Files" directory under the directory holding Morrowind.ini
     const boost::filesystem::path gameFilesDir(iniFilename.parent_path() /= "Data Files");
 
     multistrmap::const_iterator it = ini.begin();
-    for(int i=0; it != ini.end(); i++) {
+    for (int i=0; it != ini.end(); i++)
+    {
         gameFile = baseGameFile;
         gameFile.append(this->numberToString(i));
 
         it = ini.find(gameFile);
-        if(it == ini.end()) {
+        if(it == ini.end())
             break;
-        }
 
-        for(std::vector<std::string>::const_iterator entry = it->second.begin(); entry!=it->second.end(); ++entry) {
+        for(std::vector<std::string>::const_iterator entry = it->second.begin(); entry!=it->second.end(); ++entry)
+        {
             std::string filetype(entry->substr(entry->length()-3));
             Misc::StringUtils::lowerCaseInPlace(filetype);
 
-            if(filetype.compare("esm") == 0 || filetype.compare("esp") == 0) {
+            if(filetype.compare("esm") == 0 || filetype.compare("esp") == 0)
+            {
                 boost::filesystem::path filepath(gameFilesDir);
                 filepath /= *entry;
-                contentFiles.push_back(std::make_pair(lastWriteTime(filepath, defaultTime), *entry));
+                contentFiles.push_back({lastWriteTime(filepath, defaultTime), filepath});
             }
         }
     }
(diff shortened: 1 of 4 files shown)


apps/mwiniimporter/importer.cpp
```diff
@@ -834,35 +864,78 @@ void MwIniImporter::importGameFiles(multistrmap &cfg, const multistrmap &ini, co
...
 
-    // this will sort files by time order first, then alphabetical (maybe), I suspect non ASCII filenames will be stuffed.
+    // sort by timestamp
     sort(contentFiles.begin(), contentFiles.end());
-    for(std::vector<std::pair<std::time_t, std::string> >::const_iterator iter=contentFiles.begin(); iter!=contentFiles.end(); ++iter) {
-        cfg["content"].push_back(iter->second);
+
+    MwIniImporter::deplist unsorted_files;
+
+    ESM::ESMReader reader;
```
zinnschlag: Running reader without setting an encoding. Not sure if that would work properly.

apps/mwiniimporter/importer.cpp
```diff
@@ -834,35 +864,78 @@ void MwIniImporter::importGameFiles(multistrmap &cfg, const multistrmap &ini, co
...
+        }
+        unsorted_files.emplace_back(boost::filesystem::path(reader.getName()).filename().string(), deps);
+        reader.close();
     }
+
+    auto sorted_files = dependencySort(unsorted_files);
+
+    // hard-coded dependency Morrowind - Tribunal - Bloodmoon
+    {
+        auto find = [&sorted_files](const std::string& f)
```
zinnschlag: Lambda much? I get that people are excited about using C++ features. But two nested lambdas seems a bit much. Not helping readability. I suggest to turn the top level lambda into a regular function.

apps/mwiniimporter/importer.cpp
```diff
@@ -834,35 +864,78 @@ void MwIniImporter::importGameFiles(multistrmap &cfg, const multistrmap &ini, co
...
+
+        if(find("Morrowind.esm") != sorted_files.end())
+        {
+            auto foundTribunal = find("Tribunal.esm");
+            auto foundBloodmoon= find("Bloodmoon.esm");
+
+            if (foundBloodmoon != sorted_files.end() && foundTribunal != sorted_files.end())
+            {
+                auto s = *foundTribunal;
+                sorted_files.erase(foundTribunal);
```
zinnschlag: What happens if Tribunal appears before Bloodmoon? Would that invalidate the foundBloodmoon interator?

tri4ng1e:
I pushed some changes based on @zinnschlag review, please have a look. I still don't sure about hard-code part, but for now it seems to be OK. Also I fixed naming style for variables.

zinnschlag:
Better. For future references, please avoid cryptic abbreviations like dst.

If possible I would like another set of eyes on this or at least another person doing some testing, before I merge. @psi29a maybe?

tri4ng1e:
Sorry for that. I'll change them in the next commit in a couple of minutes.

zinnschlag:
@psi29a Did you get around to test it on an actual Morrowind.ini file? I am getting some strange results here. The new version of the importer fails completely for me (content file list unchanged). No error message either.

Capostrophic:
Very odd, it works just fine for me before or after the latest commit.

tri4ng1e:
Very strange.

Output of `./openmw-iniimporter -g path/to/Morrowind.ini test.cfg`:
```
load ini file: "/home/sharro/mediaZone/openmw/openmw-data/Morrowind.ini"
Warning: ignored empty value for key 'General:Beta Comment File'.
load cfg file: "test.cfg"
content file: "path/to/Bloodmoon.esm" timestamp = (1262293200) 01/01/10 00:00:00
content file: "path/to/Tribunal.esm" timestamp = (1293829200) 01/01/11 00:00:00
content file: "path/to/Morrowind.esm" timestamp = (1325361600) 01/01/12 00:00:00
write to: test.cfg
```

Morrowind.ini: (order is broken specially for tests, same result woth normal order):
```
...
[Game Files]
GameFile2=Morrowind.esm
GameFile1=Tribunal.esm
GameFile0=Bloodmoon.esm
```

test.cfg:
```
content=Morrowind.esm
content=Tribunal.esm
content=Bloodmoon.esm
...
```

zinnschlag:
Okay, running from command line now. Apparently the launcher, when running the importer, does not check for error codes and does not report errors. That is something we need to address at some point.

Anyway, I found the culprit, but the whole thing is a bit strange, because a quick look at the code suggests that this should have never worked in the first place.

The error I am getting is: ERROR: Failed to open 'Data Files/Morrowind.esm' for reading: No such file or directory

Which is correct. I don't have a Data Files directory. Problem is that the Data Files part is hardcoded. The correct way to handle this would be to check the locations specified by data settings in the openmw.cfg files.

tri4ng1e:
I noticed that thing too, but I thought it's normal because a typical Morrowind installation always contains it.

I will try to implement this, but I think it's not exactly in the scope of this PR. Should I create a new one, or just push new commits here? Perhaps it would be better to file a separate bug report for consistency.

(And can I get a developer role on the bug tracker ~~and in AUTHORS~~? I can't assign myself to any bug)

What exactly should happen if `openmw-iniimporter` is run before `openmw-wizard` so no data folders are set? Should the logic be "if data locations are set, then use them, otherwise fallback to Data Files subfolder in the same folder as Morrowind.ini"?

zinnschlag:
>  will try to implement this, but I think it's not exactly in the scope of this PR. Should I create a new one, or just push new commits here? Perhaps it would be better to file a separate bug report for consistency.

Since the importer worked for a non-standard setup before (still not entirely sure why) and it doesn't anymore with your changes, I would prefer if you fix the issue first before we merge. A separate issue on the tracker is okay though.

> (And can I get a developer role on the bug tracker and in AUTHORS? I can't assign myself to any bug)

I updated your status on the tracker. As you apparently already have noticed people only get an entry into AUTHORS after their first PR is merged. That hasn't happened yet as far as I am aware (at least I didn't merge anything from you).

> What exactly should happen if openmw-iniimporter is run before openmw-wizard so no data folders are set? Should the logic be "if data locations are set, then use them, otherwise fallback to Data Files subfolder in the same folder as Morrowind.ini"?

Interesting question. My first thought was to simple consider this error and do not continue running. But using a fallback doesn't hurt so we might as well do it.

akortunov:
> That hasn't happened yet as far as I am aware (at least I didn't merge anything from you).

#1681, just for reference. Take a look at last scrawl's comment.

zinnschlag:
It looks like scrawl forgot to include your name. Well, shit happens. Sorry about that. Fixed now.

tri4ng1e:
Please test it, should work now.

What to do if file not found? Current behaviour is to print a warning and ignore that file. Should I throw an exception?

psi29a:
Kicking the OSX build in the head... (restarting)

zinnschlag:
How did this all get so complicated :( Didn't work at first. Had to set the cfg in the command line explicitly to the one with the data directory where the esm files are located. Haven't checked which cfg file the launcher is using (when invoking the importer), but we probably can't rely on it picking the right one.

I see two options how to resolve this:
1. Implement the full procedure in the importer (load all cfg files in the right order, check both data and data-local entries in these files)
2. Return the importer to a dumb state and let the launcher handle it. After all the Morrowind.ini file is supposed to list the files in the correct order and one could argue that the importer shouldn't mess with that.

Capostrophic:
>Morrowind.ini file is supposed to list the files in the correct order

But it lists them in an alphabetical order, not what Morrowind launcher actually displays. I doubt this is the intended load order.

zinnschlag:
Hm, right. Can the vanilla Morrowind engine be any more garbage?

I guess then we are left with option 1 :(

tri4ng1e:
Sorry for the long delay, I didn't have much free time this week.

I found that when the wizard invokes `openmw-iniimporter` it doesn't save the settings to a file, so there is no `data` or `data-local` entries for the importer. I just called `writeSettings()` before invoking the importer and it worked.

zinnschlag:
Works for me too. Merging, so that we can get the wider OpenMW community in on the testing.

PR #1698, opened 2018-05-05
A bunch of C++11-related updates

Capostrophic:
There were some leftovers from the dark pre-C++11 past, as suggested by the comments to them. I cleaned them up.


psi29a:
LGTM, thanks.

PR #1699, opened 2018-05-05
Editor verifier updates

Capostrophic:
Made some changes to record checking as suggested [here](https://forum.openmw.org/viewtopic.php?f=3&p=54213#p54213):

- Removed zero duration of a carryable light source light warning (such light sources are not broken).
- Removed empty class description warning (while the name is important for dialogue to work, the description is only used for playable classes and I'd say the mod maker would make sure to add a description for such classes so it's rather annoying as a warning).
- Fixed sound range warning message
- also made a bunch of minor spelling and style fixes


psi29a:
Thanks! :)

scrawl:
> the description is only used for playable classes

How about only warning for playable classes then instead of removing the warning?

PR #1701, opened 2018-05-08
Do not apply queue movement for standing actors

akortunov:
Fixes [bug #3803](https://bugs.openmw.org/issues/3803), improves physics performance.
Related forum discussion is [here](https://forum.openmw.org/viewtopic.php?f=6&t=5139).

This PR can have side effects, so it would be nice to have an additional testing, especially on CPU-limited hardware (I have the GPU-limited one).

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -683,6 +683,7 @@ namespace MWPhysics
         , mWaterEnabled(false)
         , mParentNode(parentNode)
         , mPhysicsDt(1.f / 60.f)
+        , mIdleUpdateTimer(0)
     {
         mResourceSystem->addResourceManager(mShapeManager.get());
 
@@ -737,6 +738,18 @@ namespace MWPhysics
         delete mCollisionConfiguration;
         delete mDispatcher;
         delete mBroadphase;
+    }
+
+    void PhysicsSystem::updateIdle()
+    {
+        for (ActorMap::iterator it = mActors.begin(); it != mActors.end(); ++it)
+        {
+            osg::Vec3f pos(it->second->getCollisionObjectPosition());
+
+            RayResult result = castRay(pos, pos - osg::Vec3f(0, 0, it->second->getHalfExtents().z() + 2), it->second->getPtr(), std::vector<MWWorld::Ptr>(), CollisionType_World|CollisionType_HeightMap|CollisionType_Door);
+
+            it->second->setIdle(result.mHit);
+        }
     }
 
     void PhysicsSystem::setUnrefQueue(SceneUtil::UnrefQueue *unrefQueue)
@@ -1045,6 +1058,11 @@ namespace MWPhysics
         return physactor && physactor->getOnGround();
     }
 
+    bool PhysicsSystem::isIdle(const MWWorld::Ptr &actor)
+    {
+        Actor* physactor = getActor(actor);
+        return physactor && physactor->getIdle();
+    }
     bool PhysicsSystem::canMoveToWaterSurface(const MWWorld::ConstPtr &actor, const float waterlevel)
     {
         const Actor* physicActor = getActor(actor);
@@ -1337,6 +1355,10 @@ namespace MWPhysics
             cmode = !cmode;
             found->second->enableCollisionMode(cmode);
             found->second->enableCollisionBody(cmode);
+
+            if (cmode)
+                queueObjectMovement(MWMechanics::getPlayer(), osg::Vec3f(0, 0, -0.1f));
+
             return cmode;
         }
 
@@ -1455,6 +1477,13 @@ namespace MWPhysics
     {
         for (std::set<Object*>::iterator it = mAnimatedObjects.begin(); it != mAnimatedObjects.end(); ++it)
             (*it)->animateCollisionShapes(mCollisionWorld);
+
+        mIdleUpdateTimer -= dt;
+        if (mIdleUpdateTimer <= 0.f)
+        {
+            mIdleUpdateTimer = 0.5f;
+            updateIdle();
+        }
 
 #ifndef BT_NO_PROFILE
         CProfileManager::Reset();
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -249,7 +249,7 @@ namespace MWPhysics
                 // Check if we actually found a valid spawn point (use an infinitely thin ray this time).
                 // Required for some broken door destinations in Morrowind.esm, where the spawn point
                 // intersects with other geometry if the actor's base is taken into account
-                btVector3 from = toBullet(position);
+                btVector3 from = toBullet(position + offset);
                 btVector3 to = from - btVector3(0,0,maxHeight);
 
                 btCollisionWorld::ClosestRayResultCallback resultCallback1(from, to);
--- a/apps/openmw/mwscript/miscextensions.cpp
+++ b/apps/openmw/mwscript/miscextensions.cpp
@@ -700,6 +700,8 @@ namespace MWScript
 
                 virtual void execute (Interpreter::Runtime& runtime)
                 {
+                    MWWorld::Ptr ptr = R()(runtime);
+                    MWBase::Environment::get().getWorld()->queueMovement(ptr, osg::Vec3f(0, 0, -0.1f));
                 }
         };
 
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1314,8 +1314,6 @@ namespace MWWorld
         if (pos.z() < terrainHeight)
             pos.z() = terrainHeight;
 
-        pos.z() += 20; // place slightly above. will snap down to ground with code below
-
         if (force || !isFlying(ptr))
         {
             osg::Vec3f traced = mPhysics->traceDown(ptr, pos, 500);
@@ -2155,6 +2153,11 @@ namespace MWWorld
     bool World::isOnGround(const MWWorld::Ptr &ptr) const
     {
         return mPhysics->isOnGround(ptr);
+    }
+
+    bool World::isIdle(const MWWorld::Ptr &ptr) const
+    {
+        return mPhysics->isIdle(ptr);
     }
 
     void World::togglePOV()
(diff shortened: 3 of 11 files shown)


psi29a:
Can you sum up the side-effects?

psi29a:
Do we see anything like this in Morrowind (vanilla)?

akortunov:
> Do we see anything like this in Morrowind (vanilla)?

I do not really understand the question.
If you mean the situation when standing actor should be slightly inside obstacle, then only in mods, I guess.
If you mean the situation when the tracer places such actor to underground because of there are no obstacles under him, it was the OpenMW-specific issue.

psi29a:
Thanks for answering both. 
So 1)  we can live with, nothing really 'breaks' badly enough to cause bug reports.
2) 'was' an issue that you solved.

@zinnschlag en @kcat: how do you both feel about this? I do not mind this, this simplifies things for us and takes quite a lot of load off the physics side, especially in an area with may actors.

psi29a:
Tested on my laptop and nothing breaks that I can see. Since this helps 'un-break' mods and helps relieve pressure on the CPU side thanks work done on the physics side.

Merging. We'll get more people testing this out and reporting if anything breaks.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1314,8 +1314,6 @@ namespace MWWorld
         if (pos.z() < terrainHeight)
             pos.z() = terrainHeight;
 
-        pos.z() += 20; // place slightly above. will snap down to ground with code below
```
scrawl: This line was first added with commit c41f119ba6d4734a8cea85ee4bbada750fa2d386 "Added new game button". I think that offset needs to be there to fix the player's position in the starting boat. Did you test a new game?

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -249,7 +249,7 @@ namespace MWPhysics
                 // Check if we actually found a valid spawn point (use an infinitely thin ray this time).
                 // Required for some broken door destinations in Morrowind.esm, where the spawn point
                 // intersects with other geometry if the actor's base is taken into account
-                btVector3 from = toBullet(position);
+                btVector3 from = toBullet(position + offset);
```
scrawl: This kind of change (and the other related ones) make me anxious because we had a lot of problems with various spawn points in Morrowind.esm not working correctly. If you're going to modify the spawn placement then please go through all the related issues we had on the bug tracker (git blame will probably find most of them) and test if they don't reoccur.

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -1045,6 +1058,11 @@ namespace MWPhysics
         return physactor && physactor->getOnGround();
     }
 
+    bool PhysicsSystem::isIdle(const MWWorld::Ptr &actor)
```
scrawl: 'isIdle' seems like a misnomer. What you're really using it for is on ground status. And if you'd called it onGround, you might have noticed that we already have a PhysicActor::mOnGround. Why not use that?

apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -700,6 +700,8 @@ namespace MWScript
 
                 virtual void execute (Interpreter::Runtime& runtime)
                 {
+                    MWWorld::Ptr ptr = R()(runtime);
+                    MWBase::Environment::get().getWorld()->queueMovement(ptr, osg::Vec3f(0, 0, -0.1f));
```
scrawl: Why the need to implement 'fall'? As far as I can tell falling actors should still be detected anyway because of the ray test in updateIdle().

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -739,6 +740,18 @@ namespace MWPhysics
...
+    void PhysicsSystem::updateIdle()
+    {
+        for (ActorMap::iterator it = mActors.begin(); it != mActors.end(); ++it)
+        {
+            osg::Vec3f pos(it->second->getCollisionObjectPosition());
+
+            RayResult result = castRay(pos, pos - osg::Vec3f(0, 0, it->second->getHalfExtents().z() + 2), it->second->getPtr(), std::vector<MWWorld::Ptr>(), CollisionType_World|CollisionType_HeightMap|CollisionType_Door);
+
+            it->second->setIdle(result.mHit);
+        }
```
scrawl: This function has a lot of problems:
1) It's unnecessary to check actors out of physics processing range.
2) Even with 1) fixed, an extra raytest for every actor every frame is too heavy performance wise. I'm sure you would notice a difference in the profiling graph.
3) A ray test is not very accurate for this. If the actor is standing on a slope, the center of its collision base may be elevated and falsely detected as not on ground. A convex cast using the actor's collision shape would be more accurate, like in the code that updates onGround status in PhysicsSystem::move. Then again, I have no idea why we would want to duplicate that code in the first place.

I think there's a much easier way to achieve what you want. In MovementSolver::move, make it so sGroundOffset is not added to the movement result if the actor is not moving. Physics can stay enabled that way and you can delete all the code you added except for the adjustPosition changes.

scrawl:
> takes quite a lot of load off the physics side

As I pointed out before, that's only the case if you have mods such as the sitting NPCs or modified Idle animations, because practically all of the vanilla idle animations contain a small amount of movement in them that will activate physics anyway. With vanilla files, I'm pretty sure this PR will make things slower because of the extra ray tests for every actor.

apps/openmw/mwscript/miscextensions.cpp
akortunov: Just for sure.
IMHO, keeping placehoilders is not a good thing too.

psi29a:
The discussion and testing done indicated that the PR make nothing slower for vanilla.

@akortunov is that correct?

akortunov:
> @akortunov is that correct?

Correct. Animations indeed contain a small amount of movement, **but not every frame**.
Actually, it is a quite rare case, so I still get a speedup in areas with a lot of standing actors.

I compared results with vanilla animation files and with Animation Compilation (with integrated fixes from Gretness7), results are pretty the same.

> I'm pretty sure this PR will make things slower because of the extra ray tests for every actor.

In theory, it should not be so bad: we use raycasting only twice per second.
If someone have a hundred of actors in scene, he has other things to care about.

But anyway, I asked other people to compare results, but I did not get any feedback yet.

psi29a:
We'll see shortly in 0.44 ;)

akortunov:
> We'll see shortly in 0.44 ;)

Or probably will not see since scrawl reverted PR from master.

apps/openmw/mwphysics/physicssystem.cpp
akortunov: > an extra raytest for every actor every frame is too heavy performance wise.

Why every frame? An engine should update it only twice per second. Or I did something wrong?

> It's unnecessary to check actors out of physics processing range.

Got it.

> A ray test is not very accurate for this. If the actor is standing on a slope, the center of its collision base may be elevated and falsely detected as not on ground. A convex cast using the actor's collision shape would be more accurate, like in the code that updates onGround status in PhysicsSystem::move. Then again, I have no idea why we would want to duplicate that code in the first place.

Actually, we do not need very accurate results here: there are cases, when an object (e.g. chair) inside actor's collision box. In this case result would be unpredictable.

If something really went wrong, there is a Fall console command now.

apps/openmw/mwworld/worldimp.cpp
akortunov: > Did you test a new game?

Yes, and did not see any issues.
IIRC, there was some magic for small objects on ground, and this magic caused issues with player position during new game. This line was added to workaround this issue.

Later this magic was removed, so now this line does not affect new game.

apps/openmw/mwphysics/physicssystem.cpp
akortunov: A quick test did not show anything critical.
Actually, this change fixed placing actors on slopes for me (remember issues in Maar Gan?).
Now we cast ray from actor basement down, so such raytesting sometimes fails to detect ladders and traces actor down to terrain. If the cell is interior, actor can fall infinitely.
This PR uses an actor center as ray origin and detects slopes correctly.
Also this change required for sitting actors - raytest from basement ofter fails to detect furniture:
https://i.imgur.com/6jEvCcG.jpg

Again, my testing results are subjective.

apps/openmw/mwphysics/physicssystem.cpp
akortunov: > we already have a PhysicActor::mOnGround. Why not use that?

IIRC, to update mOnGround, we should run physics calculations for actor, but the point is not to run physics calculations for actor.

About misnomer: I agree, but can not find a good name. Any suggestions?

scrawl:
> Again, my testing results are subjective.

That's why I asked for a test against the existing problem cases we encountered so far. IIRC the tracer was made to start from the base for a reason. git blame should find you the relevant issue number(s). You won't find those areas in a quick test because they were quite specific dungeons.

I found these 4 issues in a quick search, but there's probably more (I didn't check git blame)

https://bugs.openmw.org/issues/1679
https://bugs.openmw.org/issues/1881
https://bugs.openmw.org/issues/1633
https://bugs.openmw.org/issues/1623

> IIRC, to update mOnGround, we should run physics calculations for actor, but the point is not to run physics calculations for actor.

That's precisely what you're doing though. The raycast is a physics calculation and it's just as expensive as the convex cast that the normal physics loop does.

> An engine should update it only twice per second

Oops, didn't read. That only moves the problem elsewhere though. Frame times will be unstable, we have a new magic number, and we'll noticably have floated for half a second if that code ever does what it intended to do.

> About misnomer: I agree, but can not find a good name. Any suggestions?

Well, we don't need to think of a new name if it ends up being removed. 

> Correct. Animations indeed contain a small amount of movement, but not every frame.

Are you sure? I vaguely recall logging the movements once and they did move every frame. The only animation that doesn't move is the generic 'Idle' animation, but that only plays on the player, idle followers or on NPCs that don't have any AI package, a quite small proportion of a typical scene. Note if you've disabled AI for testing (ToggleAI) NPCs don't play the specific idle animations anymore, so none of them will have movement. This doesn't occur in normal gameplay, of course.

I'm actually ok with deactivating physics, but that should be a separate PR from the sitting animations fix because there's really no need to couple those things if you go for the alternate route I suggested:
> In MovementSolver::move, make it so sGroundOffset is not added to the movement result if the actor is not moving.

If you want to pursue deactivating physics my suggestion is:
* Don't use that "just in case" gravity check every 1/2 seconds because having that safety net will make bugs in the code harder to detect, yet the end result is still slightly wrong/delayed. I'm sure you would notice floating for half a second, but you wouldn't care enough to make a bug report.
* If an object someone is standing on (easy to determine with the standing object tracker) is disabled by a script, activate gravity for that actor. We don't strictly have to do this because Morrowind doesn't and there is the Fall command, but it's so easy to do that I don't see why we shouldn't do it.
* Check if things like scripted lifts need special treatment, or if they even work right now (I think there's a bug report somewhere).
* Test a lot and don't apply the change in a hurry just before a release.

PR #1703, opened 2018-05-09
Update MapWindow every frame, not just when it's open (bug #4279)

davidcernat:
[Link to bug #4279](https://bugs.openmw.org/issues/4279)

In OpenMW 0.42, the tiles on your global map were filled in as you went through exterior cells, even if you never checked your global map. The GUI restructuring in OpenMW 0.43 made it so no map tiles were filled in as you went through cells, except for when you pressed right click to enter your "menu mode" and make your MapWindow visible, in which case – with default OpenMW settings – only your 2 most recent tiles were filled in.

This commit brings back the 0.42 behavior of updating the MapWindow every frame, with a visibility check that ensures it is not updated twice in the same frame.

### DIFF
--- a/apps/openmw/mwgui/windowmanagerimp.cpp
+++ b/apps/openmw/mwgui/windowmanagerimp.cpp
@@ -913,6 +913,9 @@ namespace MWGui
 
         updateMap();
 
+        if (!mMap->isVisible())
+            mMap->onFrame(frameDuration);
+
         mHud->onFrame(frameDuration);
 
         mDebugWindow->onFrame(frameDuration);


zinnschlag:
I haven't been able to reproduce the original problem, which is an issue because that means I can verify that this PR fixes anything. Is there a procedure for reproducing this bug?

Capostrophic:
David says in a comment to the issue:
> As a demonstration, I started a new game in each, give myself 1000 Acrobatics and Speed using the console, jumped straight north from Seyda Neen while running, and then checked my world map.

davidcernat:
You can replicate it by travelling far enough between two exterior cells without ever opening up your inventory/world map. It's most noticeable if your "exterior cell load distance" is set to its default value of 1.

Capostrophic:
Just wanted to say master does have the issue for me while master with PR's changes doesn't.

PR #1704, opened 2018-05-09
[Feedback needed] Support for per-group KF-animation files

akortunov:
Allows to use per-group KF-files to add new animation groups or override existing animation groups instead of replacing the file with all animation groups (xbase_anim.kf, for example).

For example, I have WIP weapon [un]equipping animations to support weapon sheathing:
[Animations.zip](https://github.com/OpenMW/openmw/files/1988735/Animations.zip)
If I want to use them in OpenMW, I need to merge these animations in xbase_anim.kf via external program, e.g. via AnimKit, which is Windows-only executable and requires Morrowind registry keys.
If I want to use animations with some other animation replacers (e.g. with Animation Compilation), I need to merge them too.
WIth this PR I can use new animations as a common ESP-less mod and do not touch original files.

The main idea: when OpenMW adds animation source to NPC (e.g. xbase_anim.kf), it loads also files from Animations folder (Animations/xbase_anim/* for xbase_anim.kf). Animations from these files override animations from base file.

This feature is disabled by default.

Notes: 
1. Unfortunately, source animations from AnimKit are not supposed to be loaded to game directly, at least they have keyframes for beast-specific bones.
2. Maybe we can reduce performance hit by caching merged animations in memory (probably in KeyframeManager).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
     Bug #4293: Faction members are not aware of faction ownerships in barter
     Bug #4426: RotateWorld behavior is incorrect
     Bug #4433: Guard behaviour is incorrect with Alarm = 0
+    Feature #4444: Per-group KF-animation files support
 
 0.44.0
 ------
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -31,6 +31,8 @@
 #include <components/sceneutil/skeleton.hpp>
 #include <components/sceneutil/positionattitudetransform.hpp>
 
+#include <components/settings/settings.hpp>
+
 #include <components/fallback/fallback.hpp>
 
 #include "../mwbase/environment.hpp"
@@ -466,6 +468,8 @@ namespace MWRender
             mAnimationTimePtr[i].reset(new AnimationTime);
 
         mLightListCallback = new SceneUtil::LightListCallback;
+
+        mUseAdditionalSources = Settings::Manager::getBool ("use additional anim sources", "Game");
     }
 
     Animation::~Animation()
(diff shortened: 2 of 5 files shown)


akortunov:
Looks like GCC 4.8 does not support regex (it was added to GCC 4.9). Probably I will need to find another solution.

Capostrophic:
Looks like mac travis couldn't download Qt for some reason, but it's all good now.

psi29a:
Yeah, Played around with https://wandbox.org and 

```
#include <iostream>
#include <string>
#include <regex>

int main(void)
{
    std::string model = "test";
    std::string animationPath = std::regex_replace(model, std::regex("meshes"), "Animations");
    std::cout << "DEBUG: " << animationPath << std::endl;
    return 0;
}
```

Never got it to work right with 4.8.x, so this is the right move here.

psi29a:
Kicked the travis-ci osx build, it's now building...

scrawl:
@kcat: since you wrote most of the animation code originally, what do you think of this PR?

kcat:
Conceptually it's fine. Although it's been a long time since I last worked with this code (before the change to OSG, IIRC), so I can't say if simply loading more animation files to overwrite animation keys/frames will work as intended.

On a more practical level, I'm not sure if blindly auto-loading `animations/[mesh_name_base]/*` is the best route. At least in Oblivion, I remember seeing separation animation files put into sub-directories and sub-sub-directories to improve sorting and categorization.

Also, how does this work with multiple mod directories? Say I have Mod1 (low priority) with `animations/xbase_anim/foo.kf` which has `CustomIdle1` and `CustomIdle2`, then Mod2 (high priority) has `animations/xbase_anim/foo.kf` with `CustomIdle1` and  `animations/xbase_anim/bar.kf` with `OtherCustomIdle`.

Logically, this should result in `CustomIdle1` and `OtherCustomIdle` from Mod2, and no `CustomIdle2` (since foo.kf in Mod2 "overwrote" the Mod1 file, and the Mod2 version lacks `CustomIdle2`). But a user or mod author may expect `CustomIdle2` from Mod1, and `CustomIdle1` and `OtherCustomIdle` from Mod2. What happens here? What should happen here?

akortunov:
> What happens here? 

For now this feature just uses existing VFS, so yes, an engine will take both foo.kf and bar.kf from folder with high priority.

> What should happen here?

It is a good question. Actually, I just use the modified code from font loading for now.
Any suggestions about how to improve animation merging would be welcome.

zinnschlag:
If we want to hand out this change to a wider audience this would be the right time to do it, because we are at the beginning of 0.45 development. Opinions about going ahead with the merge?

psi29a:
I think this has great potential. Would be nice to kick it into the dailies/nightlies for people to play with. :)

psi29a:
Changelog entry then merge please. :)

akortunov:
I added the changelog entry.

psi29a:
Merging. Thanks!

PR #1705, opened 2018-05-12
Use getNormalizedEncumbrance where applicable (bug #4413)

Capostrophic:
At least these two locations in the code directly used the result of dividing encumbrance by inventory capacity, which caused issues when both of them were zero: namely, you couldn't jump when you had zero strength yet zero encumbrance and your fatigue would have been fully depleted if you tried to run in this situation. getNormalizedEncumbrance does exactly the same thing but it accounts for zero capacity (returning 1) so there isn't a division by zero.

I've edited getNormalizedEncumbrance to always return 0 for zero encumbrance so that the behavior of 0/0 encumbrance is actually an assumption of 0% encumbrance, not 100%.


akortunov:
Looks good to me.

PR #1706, opened 2018-05-16
Fix compiler warning in getNestedData()

akortunov:
Fixes [bug #4409](https://bugs.openmw.org/issues/4409) (which is not really a bug).

Just split subRowIndex and subColIndex checks and complete "if" statement. It is enough to make compilers happy.


psi29a:
Thanks for catching this. :)

PR #1708, opened 2018-05-16
fix GCC warnings

akortunov:
1. Generally we should avoid of negative values shifting, because result can depend on platform.
SCHAR_MIN is -127 or less, depending on platform.
Since we divide result anyway, probably we do not need shifts here at all.
2. GCC complains about possibly uninitialized 'format' variable, it is more likely a false-positive.
I guess it is safe to initialize this variable just for sure.

### DIFF
--- a/apps/openmw/mwrender/globalmap.cpp
+++ b/apps/openmw/mwrender/globalmap.cpp
@@ -125,7 +125,7 @@ namespace MWRender
                     {
                         for (int cellX=0; cellX<mCellSize; ++cellX)
                         {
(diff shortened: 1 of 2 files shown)


psi29a:
I'm OK with this.

@kcat  any thing trigger you?

kcat:
May as well initialize `srate` as well. Otherwise, looks alright to me.

akortunov:
> May as well initialize srate as well.

Done.

psi29a:
Waiting on CI...

Capostrophic:
One of Travis builds was successful.

psi29a:
@Capostrophic no patience? :P

PR #1709, opened 2018-05-16
Rework hitbox usage

akortunov:
Should fix [bug #3591](https://bugs.openmw.org/issues/3591) and [bug #3374](https://bugs.openmw.org/issues/3374), also makes compatibility patch for Morrowind Rebirth obsolete.
Related forum topic is [here](https://forum.openmw.org/viewtopic.php?f=6&t=5154).

Basically, this PR is aimed to fix hit detection for melee combat and touch spells.
I still do not know what to do with flying creatures, so I did not touch them.
Also my observations could be incomplete or partially wrong, so feel free to test this PR or join discussion.

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1083,7 +1083,7 @@ namespace MWWorld
             // general case, compatible with all types of different creatures
             // note: we intentionally do *not* use the collision box offset here, this is required to make
             // some flying creatures work that have their collision box offset in the air
-            pos.z() += halfExtents.z() * 2 * 0.75;
+            pos.z() += halfExtents.z();
         }
 
         // the origin of hitbox is an actor's front, not center
@@ -3468,7 +3468,7 @@ namespace MWWorld
     osg::Vec3f World::aimToTarget(const ConstPtr &actor, const MWWorld::ConstPtr& target)
     {
         osg::Vec3f weaponPos = actor.getRefData().getPosition().asVec3();
-        weaponPos.z() += mPhysics->getHalfExtents(actor).z() * 2 * 0.75;
+        weaponPos.z() += mPhysics->getHalfExtents(actor).z();
         osg::Vec3f targetPos = mPhysics->getCollisionObjectPosition(target);
         return (targetPos - weaponPos);
     }
@@ -3477,7 +3477,7 @@ namespace MWWorld
     {
         osg::Vec3f weaponPos = actor.getRefData().getPosition().asVec3();
         osg::Vec3f halfExtents = mPhysics->getHalfExtents(actor);
-        weaponPos.z() += halfExtents.z() * 2 * 0.75;
+        weaponPos.z() += halfExtents.z();
 
         return mPhysics->getHitDistance(weaponPos, target) - halfExtents.y();
     }
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -2761,64 +2761,66 @@ namespace MWWorld
     {
         MWMechanics::CreatureStats& stats = actor.getClass().getCreatureStats(actor);
 
-        // Get the target to use for "on touch" effects, using the facing direction from Head node
-        MWWorld::Ptr target;
-        float distance = getActivationDistancePlusTelekinesis();
-
-        osg::Vec3f hitPosition = actor.getRefData().getPosition().asVec3();
-        osg::Vec3f origin = getActorHeadTransform(actor).getTrans();
-
-        osg::Quat orient = osg::Quat(actor.getRefData().getPosition().rot[0], osg::Vec3f(-1,0,0))
-                * osg::Quat(actor.getRefData().getPosition().rot[2], osg::Vec3f(0,0,-1));
-
-        osg::Vec3f direction = orient * osg::Vec3f(0,1,0);
-        osg::Vec3f dest = origin + direction * distance;
-
         // For AI actors, get combat targets to use in the ray cast. Only those targets will return a positive hit result.
         std::vector<MWWorld::Ptr> targetActors;
         if (!actor.isEmpty() && actor != MWMechanics::getPlayer())
             actor.getClass().getCreatureStats(actor).getAiSequence().getCombatTargets(targetActors);
 
-        // For actor targets, we want to use bounding boxes (physics raycast).
-        // This is to give a slight tolerance for errors, especially with creatures like the Skeleton that would be very hard to aim at otherwise.
-        // For object targets, we want the detailed shapes (rendering raycast).
-        // If we used the bounding boxes for static objects, then we would not be able to target e.g. objects lying on a shelf.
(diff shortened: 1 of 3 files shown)


apps/openmw/mwworld/worldimp.cpp
```diff
@@ -2743,64 +2752,66 @@ namespace MWWorld
...
-        if (!target.isEmpty() && target.getClass().isActor() && target.getClass().getCreatureStats (target).getAiSequence().isInCombat()) 
-        {
-            distance = std::min (distance, getStore().get<ESM::GameSetting>().find("fCombatDistance")->getFloat());
-            if (distance < dist1)
-                target = NULL;
+            osg::Quat orient = osg::Quat(actor.getRefData().getPosition().rot[0], osg::Vec3f(-1,0,0))
+                    * osg::Quat(actor.getRefData().getPosition().rot[2], osg::Vec3f(0,0,-1));
+
+            osg::Vec3f direction = orient * osg::Vec3f(0,1,0);
+            float distance = getMaxActivationDistance();
```
Capostrophic: Telekinesis no longer gives a bonus (the distance was previously getActivationDistancePlusTelekinesis()). Was that intentional? Shouldn't on-touch spells benefit from telekinesis?

akortunov: Actually, gives a bonus, when we use the focused object, as in vanilla game.
From my testing, telekinesis does not give distance bonus against other actors (use hitboxes here).

Capostrophic: Alright. Otherwise everything looks and works fine for me.

Capostrophic: Though the original change was very much intentional, see [this Allofich's PR](https://github.com/OpenMW/openmw/pull/1074).

akortunov: I do not understand what "explosions for touch spell" means, but for normal touch spells telekinesis still give you range bonus for actors out of combat, because we use the faced object as a target, if you can activate it (tooltip is shown).

Capostrophic: Oh, I misparsed the code again. Carry on.

Allofich: > I do not understand what "explosions for touch spells" means,

"Explosion" means the effect that happens when a spell has an `Area` value specified and the spell collides with something, like when a fireball hits something. I found that with a touch-range spell, if the spell has a non-0 Area value and is used on some objects (for example, clothes or weapons, or doors that do not transport the player to a new cell), there will also be such an explosion, and the explosion can damage actors. If a Telekinesis effect is active, although you cannot use a touch-range spell on an actor directly, you can use the explosion effect to damage them from range this way.

 > for normal touch spells telekinesis still give you range bonus for actors out of combat
In the original game telekinesis does not give you a range bonus for actors out of combat. Probably so you can't "talk" to them from far away. Because of this, you also can't use a touch-range spell + telekinesis combination on them directly.

akortunov: > In the original game telekinesis does not give you a range bonus for actors out of combat. Probably so you can't "talk" to them from far away

Nevermind, ``getFacedObject()`` does not take in account telekinesis anyway if ``facedObject.getClass().allowTelekinesis(facedObject)`` returns false.
So this PR should replicate described behaviour for now:

> we use the faced object as a target, if you can activate it (tooltip is shown).

psi29a:
I like this, but we need more testers...

akortunov:
> I like this, but we need more testers...

I doubt that many users check PRs out and make long-time gameplay tests. Usually they use nightly builds for that.

psi29a:
What is your target audience then? If necessary, ask your fellow devs to help validate.

akortunov:
> What is your target audience then? If necessary, ask your fellow devs to help validate.

I asked for help on russian forum, let see what will happen. Probably people will ask for binary windows builds to test, since the build instruction for Windows is a quite complicated for average Morrowind player/modmaker, even with script.

psi29a:
What if I told you... that we could archive the build artifacts in appveyor? You could post the URL here to the zip file containing the build?  Would that help?

psi29a:
For example, like I do for WildMIDI: 
https://ci.appveyor.com/project/psi29a/wildmidi/build/job/ea6rku62onmclbam/artifacts

psi29a:
Did you read the topic? The infrastructure is there but the artifacts are debug builds and thus very slllllooooowwww...  However, it is better than nothing.

psi29a:
I can fix us up with something around lunch or later tonight. If you want to take things into your own hands, please do. You can use this as an example:
https://github.com/Mindwerks/wildmidi/blob/master/appveyor.yml

Push your changes to this PR so that appveyor can produce zip files and you can then post that link for others to test.

akortunov:
Well, AppVeyor now preserves builds. But I suppose that we do not need to pack pdb files to archive.

psi29a:
Correct, all of that needs to go into the zip as well as I doubt end-users of the zip will have those libs in their system path.

psi29a:
Are you going to mark this as 100%/finished?
https://bugs.openmw.org/issues/4015

akortunov:
> Are you going to mark this as 100%/finished?

I suppose for that I need to:
1. Move PDB-files do separate package.
2. Pack dependencies, probably into separate package too.

psi29a:
The dependencies probably won't be collected as they are external to the build. The deps are available in the scripts themselves however, so you could post them along with the debug builds and tell the end-user to extract them either into OpenMW directory or to their system path.

psi29a:
As for the 7zip exclude pdb... perhaps this might work?  

`'-xr!*.pdb'`

I think the scripts run in powershell, thus we have to use single quotes '

Can you try the above?

akortunov:
> Can you try the above?

Unfortunately, it did not work.

psi29a:
Looks like a success to me...  you agree?

akortunov:
> Looks like a success to me... you agree?

I managed to run debug build on Windows, but it is **really** sloooow (5-10 fps), so I'd rather consider this as failure (sigh).
Well, negative result is a result too.

I suggest to split CI changes to another PR if we want to continue investigationg.
I wonder if we can teach AppVeyor to make release builds on demand (this could be useful for long-standing branches like shadows or navmeshes).

psi29a:
It would be interesting to see what the compiles times are now for release builds... there is apparently a large speedup between msvc2015 (25m) and msvc2017(18m).

zinnschlag:
Seems the discussion got a bit off-topic. What is the state on this? If we want more testing, then we should merge it into master soon. However it seems travis doesn't like it anymore.

akortunov:
> What is the state on this?

It is finished, I tested it for my own and it works for me.
This PR did not get the review from scrawl yet.

> However it seems travis doesn't like it anymore.

Travis sometimes fails do download dependencies for some reason. This error is random and can happen with any PR.

psi29a:
Fixed the merge conflict.  I'm OK with this.

PR #1710, opened 2018-05-17
Rename Original Creature field into Parent Creature (bug #2897)

Capostrophic:
[Original report.](https://bugs.openmw.org/issues/2897)
*I think* this variant is good enough.

Yet a tiny change. I'll look into more serious editor stuff later.


Capostrophic:
Something odd with AppVeyor.

PR #1711, opened 2018-05-17
[Post-1.0] Add rain collision and muffle weather sounds when under cover

ghost:
Early PR for code review and feedback, these commits add collision for rain particles via PhysicsSystem::castRay.

Option for this is currently placed under [Water] in config files and in the Settings UI, currently defaulting to disabled.

The source for the PhysicsSystem::castRay call is an extrapolated particle origin somewhere high in the sky as castRay will not report a hit if both the source and destination vectors are inside an object (frequent when in a Vivec canton).

Rain sounds are now also pitched-down to sound more muffled depending on how many of the surrounding weather particles are being culled in the collision check, resulting in muffled sounding rain when standing under cover.

### DIFF
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -1119,6 +1119,7 @@ SkyManager::SkyManager(osg::Group* parentNode, Resource::SceneManager* sceneMana
     , mWindSpeed(0.f)
     , mEnabled(true)
     , mSunEnabled(true)
+    , mRainCoverage(1.0f)
 {
     osg::ref_ptr<CameraRelativeTransform> skyroot (new CameraRelativeTransform);
     skyroot->setName("Sky Root");
@@ -1450,10 +1451,11 @@ protected:
 class WorldCollisionOperator : public osgParticle::Operator
 {
 public:
-    WorldCollisionOperator(osg::Camera *camera, MWPhysics::PhysicsSystem* physicsSystem): osgParticle::Operator()
+    WorldCollisionOperator(osg::Camera *camera, MWPhysics::PhysicsSystem* physicsSystem, float* rainCoverage): osgParticle::Operator()
     {
         mCamera = camera;
         mPhysicsSystem = physicsSystem;
+        mRainCoverage = rainCoverage;
     }
 
     virtual osg::Object *cloneType() const override
@@ -1473,18 +1475,31 @@ public:
     virtual void operateParticles(osgParticle::ParticleSystem *ps, double dt) override
     {
         if(mPhysicsSystem->getRainCollisionEnabled()) {
+            int r = 0;
             osg::Vec3 cameraPosition = getCameraPosition();
 
-            for (int i = 0; i < ps->numParticles(); ++i) {
+            int numParticles = ps->numParticles();
+
+            for (int i = 0; i < numParticles; ++i) {
                 osgParticle::Particle *p = ps->getParticle(i);
 
                 osg::Vec3f pos = cameraPosition + p->getPosition();
                 osg::Vec3f extrapolatedOrigin = pos - (p->getVelocity() * 100);
 
-                MWPhysics::PhysicsSystem::RayResult result = mPhysicsSystem->castRay(extrapolatedOrigin, pos);
+                MWPhysics::PhysicsSystem::RayResult result = mPhysicsSystem->castRay(extrapolatedOrigin, pos, MWWorld::ConstPtr(), std::vector<MWWorld::Ptr>(), MWPhysics::CollisionType_World);
                 if (result.mHit)
+                {
                     p->kill();
+                    r++;
+                }
             }
+
+            if(numParticles > 0)
+                *mRainCoverage = 1.0f - ((r / (float)numParticles) / 2);
+            else
+                *mRainCoverage = 1.0f;
+        }else{
+            *mRainCoverage = 1.0f;
         }
     }
 
@@ -1495,6 +1510,7 @@ protected:
     {
         return mCamera->getInverseViewMatrix().getTrans();
     }
+    float* mRainCoverage;
 };
 
 
@@ -1550,8 +1566,8 @@ void SkyManager::createRain()
     updater->addParticleSystem(mRainParticleSystem);
 
     osg::ref_ptr<osgParticle::ModularProgram> program (new osgParticle::ModularProgram);
+    program->addOperator(new WorldCollisionOperator(mCamera, mPhysicsSystem, &mRainCoverage));
     program->addOperator(new WrapAroundOperator(mCamera,rainRange));
-    program->addOperator(new WorldCollisionOperator(mCamera, mPhysicsSystem));
     program->setParticleSystem(mRainParticleSystem);
     mRainNode->addChild(program);
 
@@ -1952,6 +1968,11 @@ void SkyManager::listAssetsToPreload(std::vector<std::string>& models, std::vect
 void SkyManager::setWaterEnabled(bool enabled)
 {
     mUnderwaterSwitch->setEnabled(enabled);
+}
+
+float SkyManager::getRainCoverage() const
+{
+    return mRainCoverage;
 }
 
 }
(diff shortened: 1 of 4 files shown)


tcotys:
Nice,
just to say that, in order to respect real physics (I'm not a real programmer, just a physicist), you cannot change the pitch of the sound (exept when the player is moving fast).
You have to apply a low pass filter around 200-300 Hz (maybe add some reverb).
The result might be better (or at least similar).

There is an example with 
https://github.com/OpenMW/openmw/commit/edfba68eb583948db714d2144be261a8800f65f1

akortunov:
Nice feature!

Note: unfortunately, rain water ripples are part of water shader, so an engine still renders them during rain (under bridges over rivers, fro example).

apps/openmw/mwrender/sky.cpp
```diff
@@ -1445,6 +1448,72 @@ class WrapAroundOperator : public osgParticle::Operator
...
+
+            int numParticles = ps->numParticles();
+
+            for (int i = 0; i < numParticles; ++i) {
+                osgParticle::Particle *p = ps->getParticle(i);
+
+                osg::Vec3f pos = cameraPosition + p->getPosition();
+                osg::Vec3f extrapolatedOrigin = pos - (p->getVelocity() * 100);
+
+                MWPhysics::PhysicsSystem::RayResult result = mPhysicsSystem->castRay(extrapolatedOrigin, pos, MWWorld::ConstPtr(), std::vector<MWWorld::Ptr>(), MWPhysics::CollisionType_World);
```
akortunov: Keep in mind that not all game objects have precise collision mesh. Probably rendering raycasting will give better results.

kcat:
I don't think we want to add collision to the rain particles. Given the number of particles moving around the player, it won't be particularly efficient since our physics aren't very optimized, and the way each particle handles multiple raindrops or snowflakes, with the inexact collision meshes, you'll end up with precipitation still going through some cover or being blocked by seemingly nothing.

I'd still prefer something like [this idea](https://forum.openmw.org/viewtopic.php?f=38&t=2516#p28317) to prevent rain and snow from falling through roofs.

I'd probably also hold off on muffling the rain sound until we have a more generic solution for sound occlusion/obstruction, which can be applied to most sounds blocked by geometry (and as mentioned, you don't want to do it by simply lowering the pitch; it's better to use a low-pass filter).

16bitint:
Thanks for the comments guys.

I've been running this on my local branch for a little while, just while casually playing the game and I've sunk about 20 hours into playing with this turned on, as well as some more directed performance testing. In my testing I've not been able to bring about any noticable performance drops by enabling this feature, though my PC is probably toward the higher-end of systems OpenMW targets.

Theoretically there should be no performance drops at all when the option is turned off, apart from the comparison in WorldCollisionOperator::operateParticles against mRainCollision, which I assume would be totally negligible on any targeted system.

I was initially using rendering raycasts as opposed to collision raycasts, but that definitely did have a huge impact on performance, unplayably so, therefore I went with collision raycasts instead. Throughout my own playtesting the collision meshes seem to be more than adequate for this, though this is pretty anecdotal and others' experience may vary.

There's certainly some room for optimization here too. The first idea in my head is to not update particles that aren't in the camera's view, though that would invalidate the rain coverage calculations (at least in its current implementation).

These calculations and rain sound processing were put in after the collision, it's definitely less of my focus here. I had initially thought about low-pass filtering but changed my mind when I saw just how easy a pitch-down could be implemented. I'm certainly no physicist, and I thought that, given that the rain sound's main element is similar enough to white noise, a pitch-down was a fairly nice approximation of low-pass filtering.

Also, in terms of the implementation itself, can anyone see any obvious issues? I'm not very experienced with C++ so it's quite possible I've gone about something the wrong way here, and I'd appreciate any feedback regarding the actual code changes in these commits.

Again, thanks for the comments everyone. Even if this idea gets shelved it's really cool to be having this discussion and getting feedback.

akortunov:
> That isn't a blocker and will take a bit more work to fix that as well.

I did not say that it is a blocker.

About possible issues: in which thread does WorldCollisionOperator work? If it works in own OSG thread, there could be problems with multithreading.

zinnschlag:
This is an advanced feature that is not required for 1.0. Therefore I nominate it for putting it on hold until 1.0.

psi29a:
It still needs a bit of work I believe, like ray-casting changes and the use of a low-pass filter before this can be considered for inclusion.

nalal:
I hate bringing up drama but it kinda needs to be recognized in this situation as I feel affiliations need to be disclosed. I'm sure most folks know about Malseph but for those who don't, they were a server manager that was removed from the TES3MP community for rather substantial and consistent incidents of targeted harassment towards specific individuals in his own server community and has been alleged to be responsible for inter-server sabotage as well as silencing nonconforming voices within his community.
 
When Mal was finally banned from TES3MP he proceeded to deny all allegations even though substantial evidence had been leveled against him and began a petty flame war as well as trying to highjack the project via forking the repository and not clearly labeling it as a distinct project separate from TES3MP (which is a violation of the GPL that TES3MP is distributed under). As Malseph himself is not familiar with C++ development he requisitioned the aid of the individual responsible for this commit, 16bitint, who was not only complicit in previously stated GPL violation but also the instigator of it.
 
**CITATION:**
https://github.com/Malseph/openmw-tes3mp/commit/8071a82056bba5c219bf0a7c4f653b969ecc1bb3#diff-04c6e90faac2675aa89e2176d2eec7d8R8
 
From what I've learned on my end, 16bitint apologized halfheartedly and moved to claim ignorance on the subject which is why the dialogue had been closed until now. However after recently (as in within the last hour of posting this)  joining Mal's discord server it proudly displays him as the only (as far as I can tell that is) developer working on Malseph's TES3MP distribution.
 
**CITATION:**
https://i.imgur.com/AwO5a1k.png
 
I have been brought to believe that that 16bitint is likely the same type of person Mal is, that being inherently manipulative and socially predatory towards those around them. As such I feel that accepting contributions from him might not be the wisest thing to do when the context of the project's stability and longevity are in question.

16bitint:
@nalal,
I would've certainly appreciated if this had been brought up with me first, especially given that you already found me on Discord. Maybe we could've foregone these comments. In the interest of killing drama as quickly as possible, I'll try to keep this short.

I haven't made any development effort regarding TES3MP since that linked commit, or indeed development regarding OpenMW as a whole. While on the topic of the linked commit, looking at Git hunk @@ -20,7 +20,8 @@ for README.md, you can see that I made a conscious effort to specify that the repo was forked, with credit to the upstream developers coming before my own. The Discord role was assigned to me by Malseph the day he made the release announcement, and assuming you're still in that discord server you can see that I haven't actually sent any messages on there since. 

While your comments insinuating that I'm "inherently manipulative and socially predatory" are simply inflammatory and unnecessary, I'm more unhappy with the claim that I "apologized halfheartedly." I had a conversation with David after the matter in which we both accepted fault and apologized for the disagreement which had resulted in me being removed from the TES3MP project. After that conversation, dialogue had been, as you said, closed until now. I assumed it was because we had come to an understanding, so I’m hoping that this was just your opinion and not by David's own account.

If you'd like to discuss this further, please message me on Discord.

nalal:
@16bitint 
I apologize for not going to you first, now that things have been cleared up I have removed my comment.

davidcernat:
On a technical level, this should be redone in an entirely different and much more optimized way that is also usable for dust and snow and potentially other types of weather particles that can get added in post-1.0. If you like, I'm fine with doing it myself once 1.0 is out, perhaps based on Chris' ideas.

On an ethical level, OpenMW shouldn't turn a blind eye to the fact that the creator of this pull request is the only coder who [teamed up](https://github.com/Malseph/openmw-tes3mp/commit/8071a82056bba5c219bf0a7c4f653b969ecc1bb3) with far-right extremist Malseph to do a [GPL-violating hostile takeover of the TES3MP name](https://www.reddit.com/r/tes3mp/comments/9s0go1/how_malseph_behaves_in_world_of_warcraft_for/e8o1hi0/) by releasing something that they had no right to call TES3MP 0.6.3, unless OpenMW wants to create a precedent where [anyone taking over the OpenMW name](https://i.imgur.com/Hc6mQWk.jpg) is also tolerated, while also harming relations with TES3MP at a time when it is seriously being considered for inclusion in its parent project, all for the sake of having a framerate-halving take on rain collision.

PR #1713, opened 2018-05-20
Don't reset sneaking camera offset while in GUI (bug #4420)

Capostrophic:
There's no real reason to reset the offset, but if updatePlayer gets called while in a GUI and then it gets called when the player is no longer in a GUI, it could cause a momentary flickering back into sneaking point of view.

I don't see this anymore when this is applied.

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1621,7 +1621,7 @@ namespace MWWorld
         if (!paused)
             doPhysics (duration);
 
-        updatePlayer(paused);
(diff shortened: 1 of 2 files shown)


scrawl:
I think the paused check was intended so you can't toggle sneaking while paused. What happens now if you try to do that?

Capostrophic:
Nothing. [You can't toggle sneaking in an interface.](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwinput/inputmanagerimp.cpp#L1155)
Just checked that. If you press the sneak button, nothing happens when you exit the interface.

PR #1714, opened 2018-05-20
Switch Appveyor CI to x64 VS2015 / 2017

ananace:
Just to help debug the apparent CI failures on Windows

PR #1715, opened 2018-05-20
Add OSG 3.6 binaries to CI script

ananace:
Will have to drop Visual Studio 2013 support for this, don't have the ability to build up new 2013 binaries for 3.6.

Because of this, #1714 is probably going to be required to switch the CI from VS2013 + VS2015 to VS2015 + VS2017

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -333,7 +333,7 @@ if [ -z $SKIP_DOWNLOAD ]; then
 	# OSG
 	download "OpenSceneGraph 3.6.0-scrawl" \
 		"https://www.lysator.liu.se/~ace/OpenMW/deps/OSG-3.6.0-scrawl-msvc${MSVC_DISPLAY_YEAR}-win${BITS}.7z" \
-		"OSG-3.4.1-scrawl-msvc${MSVC_DISPLAY_YEAR}-win${BITS}.7z"
+		"OSG-3.6.0-scrawl-msvc${MSVC_DISPLAY_YEAR}-win${BITS}.7z"
 
 	# Qt
 	if [ -z $APPVEYOR ]; then
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -332,8 +332,8 @@ if [ -z $SKIP_DOWNLOAD ]; then
 
 	# OSG
 	download "OpenSceneGraph 3.6.0-scrawl" \
-		"https://www.lysator.liu.se/~ace/OpenMW/deps/OSG-3.6.0-scrawl-msvc${MSVC_YEAR}-win${BITS}.7z" \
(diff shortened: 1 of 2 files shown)


ananace:
This should probably not be merged before it's been tested - and verified - by at least two people, seeing as OSG is a rather vital component of OpenMW.

ananace:
Right, the new OSG deps don't actually have any zlib DLL, the compression's statically linked.

ananace:
I've uploaded new packages with the missing DLLs now

AnyOldName3:
It built and runs.

psi29a:
@AnyOldName3 runs is good, did you run/fly around the landmass to see if there were any crashes with all the loading and unloading of cells (actors and statics)? That is a good stress test.

AnyOldName3:
I had one crash nearly immediately but can't replicate it with ten other runs doing the exact same thing.

akortunov:
Not sure if it 3.6-specific, but sometimes I get the crash in SceneManager::createInstance().
Log complains about dynamic casts [here](https://github.com/OpenMW/openmw/blob/master/components/sceneutil/clone.cpp#L35).

akortunov:
> I don't understand how those dynamic casts could break. 

I even more do not understand.
[crash.log](https://github.com/OpenMW/openmw/files/2103568/crash.log)

OpenMW can crash randomly in very rare case upon save load or cell change.

This part is common in all cases:

> #3  0x00007f80bfbcd743 in __dynamic_cast () from /usr/lib/gcc/x86_64-pc-linux-gnu/6.4.0/libstdc++.so.6
No symbol table info available.
#4  0x00005627b6bc6c88 in SceneUtil::CopyOp::operator()(osg::Node const*) const ()
No symbol table info available.
#5  0x00007f80c6539cbd in osg::Group::Group(osg::Group const&, osg::CopyOp const&) () from /usr/lib64/libosg.so.156
No symbol table info available.
#6  0x00005627b6bcd1f1 in SceneUtil::Skeleton::Skeleton(SceneUtil::Skeleton const&, osg::CopyOp const&) ()
No symbol table info available.
#7  0x00005627b6bcdf95 in SceneUtil::Skeleton::clone(osg::CopyOp const&) const ()
No symbol table info available.
#8  0x00005627b6ba1083 in osg::Node* osg::clone<osg::Node>(osg::Node const*, osg::CopyOp const&) ()
No symbol table info available.
#9  0x00005627b6b9e257 in Resource::SceneManager::createInstance(osg::Node const*) ()

akortunov:
> OpenMW can crash randomly in very rare case upon save load or cell change.

I have this crash on 3.4.1 too, so it is not 3.6-specific. Weird.

akortunov:
This PR is very outdated, and we are not going to use 3.6.x by default soon anyway because of its bug. So I see no point to keep this PR opened.

PR #1716, opened 2018-05-21
Soul gem value rebalance

declan-millar:
This branch incorporates the 'Soul gem value rebalance' formula from the Morrowind Code Patch as this was the only major economy-breaker that couldn't be resolved with mods. It's a very small tweak, for which I've added a checkbox in the Advanced page of the launcher.

In a niche case, I also chose to add the value of the empty Azura's Star to the total filled value as it is uniquely not destroyed during enchanting/recharging.

I thought it might be of interest to other people to help balance their playthroughs with OpenMW, but I totally understand if the developers want to reject this merge in order to avoid bloating the settings/launcher (or because my implementation is poor).

P.S. I've also posted a topic related to this merge on the developer application board of the forum.

### DIFF
--- a/apps/openmw/mwclass/misc.cpp
+++ b/apps/openmw/mwclass/misc.cpp
@@ -88,7 +88,7 @@ namespace MWClass
             if (creature)
             {
                 int soul = creature->mData.mSoul;
-                if (Settings::Manager::getBool("rebalance soulgem values", "Game"))
+                if (Settings::Manager::getBool("rebalance soul gem values", "Game"))
                 {
                     // use soulgem value rebalance formula from morrowind code patch 
                     float soulValue = 0.0001 * pow(soul, 3) + 2 * soul;
--- a/apps/openmw/mwclass/misc.cpp
+++ b/apps/openmw/mwclass/misc.cpp
@@ -87,10 +87,10 @@ namespace MWClass
             const ESM::Creature *creature = MWBase::Environment::get().getWorld()->getStore().get<ESM::Creature>().search(ref->mRef.getSoul());
             if (creature)
             {
+                int soul = creature->mData.mSoul;
                 if (Settings::Manager::getBool("rebalance soulgem values", "Game"))
                 {
                     // use soulgem value rebalance formula from morrowind code patch 
-                    int soul = creature->mData.mSoul;
                     float soulValue = 0.0001 * pow(soul, 3) + 2 * soul;
                     
                     // for Azura's star add the unfilled value
@@ -100,7 +100,7 @@ namespace MWClass
                         value = soulValue;
                 }
                 else
-                    value *= creature->mData.mSoul;
+                    value *= soul;
             }
         }
 
--- a/apps/openmw/mwclass/misc.cpp
+++ b/apps/openmw/mwclass/misc.cpp
@@ -90,7 +90,7 @@ namespace MWClass
                 int soul = creature->mData.mSoul;
                 if (Settings::Manager::getBool("rebalance soul gem values", "Game"))
                 {
-                    // use soulgem value rebalance formula from morrowind code patch 
+                    // use the 'Soul gem value rebalance' formula from morrowind code patch 
                     float soulValue = 0.0001 * pow(soul, 3) + 2 * soul;
                     
                     // for Azura's star add the unfilled value
--- a/apps/openmw/mwclass/misc.cpp
+++ b/apps/openmw/mwclass/misc.cpp
@@ -89,13 +89,13 @@ namespace MWClass
                 
                 // use soulgem value rebalance formula from morrowind code patch 
                 int soul = creature->mData.mSoul;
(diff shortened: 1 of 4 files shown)


psi29a:
@zinnschlag is this a case where we just want to 'fix' it or leave it selectable but **on** by default?

files/settings-default.cfg
```diff
@@ -216,6 +216,9 @@ followers attack on sight = false
 # Can loot non-fighting actors during death animation
 can loot during death animation = true
 
+# Makes the value of filled soul gems dependent only on soul magnitude (with formula from the Morrowind Code Patch)
+rebalance soul gem values = false
```
psi29a: Honestly, perhaps we should make this true since this is something that should have been fixed in Morrowind but wasn't.

zinnschlag:
Not hugely excited about making changes to game mechanics that really should be left to content developers. But since I don't have anything about soulgems in my post 1.0 de-hardcoding plan, I am okay with it.

However we definitely need to make this optional, because it is a clear deviation from the original game. I have no objections against setting the default to true.

apps/openmw/mwclass/misc.cpp
```diff
@@ -85,7 +86,22 @@ namespace MWClass
...
             const ESM::Creature *creature = MWBase::Environment::get().getWorld()->getStore().get<ESM::Creature>().search(ref->mRef.getSoul());
             if (creature)
-                value *= creature->mData.mSoul;
+            {
+                int soul = creature->mData.mSoul;
+                if (Settings::Manager::getBool("rebalance soul gem values", "Game"))
+                {
+                    // use the 'soul gem value rebalance' formula from the Morrowind Code Patch 
+                    float soulValue = 0.0001 * pow(soul, 3) + 2 * soul;
+
```
Capostrophic: Superfluous whitespace here.

psi29a: In C++, all whitespace is superfluous. ;)
One line doesn't hurt readability.

psi29a:
Thanks for this! Merging.

psi29a:
@MiroslavR thanks, updated.

scrawl:
I strongly disagree with the default being 'on'. Compare the mission statement in Contributing.md:

> OpenMW, in its default configuration, is meant to be a faithful reimplementation of Morrowind,

Faithful means we don't modify game mechanics. As soon as we make noticable changes we get a different game that hasn't been playtested as much as Morrowind has.
For example, you can get a lot of money in the early game by stealing Galbedir's soul gems. Changing the amount of this starting boost will definitely change the game experience.
IMO, attempting to 'balance' is problematic in itself. Having some unbalanced things isn't necessarily bad because I as a player can be excited when I discover an 'unbalanced' thing that's much better than the average thing. A world where everything I can do is equally powerful would feel artificial to me. That's just my opinion though, and we can't possibly predict every single opinion out there, so the safest thing to do is to just leave everything the way it is in Morrowind, since people do like that game a lot apparently. 

I'm also not sure how I feel about this kind of setting being available in the launcher. Isn't it kind of immersion breaking to be able to change the value of your items mid-playthrough just by ticking a box? Seems to me this would belong in mods, not settings.

scrawl:
(I'm torn between reverting the whole thing or just changing the default setting, if it wasn't obvious. Suggestions welcome)

scrawl:
> bug

Citation needed. I've never heard of this 'bug' in my Morrowind playing career. 

OP didn't call it a bug and even MCP itself lists this under the 'Feature' section. And as far as I know, 'Feature' in MCP is another word for 'We don't know if everyone will like or need this change, so it's not enabled by default.'

scrawl:
So what?

> tweaked [...] for personal use

Implying it's their own personal opinion.

> the only major economy-breaker that couldn't be resolved with mods.

This implies there are more economy breakers (and yes, there are many). A balanced economy isn't a feature in Morrowind, so it being 'broken' in some instances isn't a bug.

You haven't given me an actual argument aside from 'someone said so', which isn't an argument, especially if that isn't what they said. If you want to deviate from the previous, known working and tested behavior, the onus is on you to provide the reasoning for it. I haven't seen any, so will revert at least the default config for now.

kcat:
> A balanced economy isn't a feature in Morrowind, so it being 'broken' in some instances isn't a bug.

That's a rather circular argument, isn't it? I can't think the developers intended to have an unbalanced economy, so behaviors that unbalance the economy are clearly a bug.

Obviously what to do about the bug is debatable. Given its gameplay implications, I'd agree that it should at least default to vanilla behavior. But given it being an option in MCP that some users use to help keep the economy in balance, and the lack of any immediate plans for allowing modders to do something about it after 1.0, replicating the MCP option seems sensible if it's not unduly intrusive on the codebase.

psi29a:
>  I haven't seen any, so will revert at least the default config for now.

If it makes you happy. ;)

PR #1717, opened 2018-05-22
Render default land texture for Wilderness cells with distant terrain

akortunov:
Partially fixes [bug #4405](https://bugs.openmw.org/issues/4405).
There are still missing textures outside of world map bounds (looks like quad world is not infinite, or I just missed something).
Also default textures are still missing it the editor - looks like it uses a different approach.


psi29a:
Doing more with less is great. ;) 

Merging.

PR #1718, opened 2018-05-24
Preserve AppVeyor artifacts

akortunov:
[Feature #4015](https://bugs.openmw.org/issues/4015).
Some discussion can be found at #1709.

Just enable release builds and pack binaries to ZIP-archive, so Windows users in theory can test complex PRs without full environment setup.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -360,6 +360,8 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL GNU OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-but-set-parameter")
     endif()
 elseif (MSVC)
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS} /Zi /bigobj")
+    set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} /DEBUG /OPT:REF /OPT:ICF /INCREMENTAL:NO")
     # Enable link-time code generation globally for all linking
     if (OPENMW_LTO_BUILD)
         set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /GL")
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,8 +18,8 @@ platform:
     - x64
 
 configuration:
-    - Debug
-#   - Release
+#    - Debug
+   - Release
 
 # For the Qt, Boost, CMake, etc installs
 #os: Visual Studio 2017
@@ -56,8 +56,8 @@ build_script:
     - cmd: msbuild %build%\OpenMW.sln /t:Build /p:Configuration=%configuration% /m:2 /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
 
 after_build:
-    - cmd: if %PLATFORM%==x64 7z a OpenMW_MSVC%msvc%_x64.zip %APPVEYOR_BUILD_FOLDER%\MSVC%msvc%_64\Debug\ -xr"!*.pdb"
-    - cmd: if %PLATFORM%==x64 7z a OpenMW_MSVC%msvc%_x64_pdb.zip %APPVEYOR_BUILD_FOLDER%\MSVC%msvc%_64\Debug\*.pdb
+    - cmd: if %PLATFORM%==x64 7z a OpenMW_MSVC%msvc%_x64.zip %APPVEYOR_BUILD_FOLDER%\MSVC%msvc%_64\Release\ -xr"!*.pdb"
+    - cmd: if %PLATFORM%==x64 7z a OpenMW_MSVC%msvc%_x64_pdb.zip %APPVEYOR_BUILD_FOLDER%\MSVC%msvc%_64\Release\*.pdb
 
 test: off
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -55,6 +55,10 @@ build_script:
     - cmd: if %PLATFORM%==x64 set build=MSVC%msvc%_64
     - cmd: msbuild %build%\OpenMW.sln /t:Build /p:Configuration=%configuration% /m:2 /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
 
+after_build:
+    - cmd: if %PLATFORM%==x64 7z a OpenMW_MSVC%msvc%_x64.zip %APPVEYOR_BUILD_FOLDER%\MSVC%msvc%_64\Debug\ -xr"!*.pdb"
+    - cmd: if %PLATFORM%==x64 7z a OpenMW_MSVC%msvc%_x64_pdb.zip %APPVEYOR_BUILD_FOLDER%\MSVC%msvc%_64\Debug\*.pdb
+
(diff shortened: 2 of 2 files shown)


psi29a:
Release builds are built faster than debug builds...
Environment: msvc=2015, APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2015
18 min 7 sec
Environment: msvc=2017, APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2017
18 min 31 sec

there was probably something rotten going on with MSVC2013 that caused that to not be true...

psi29a:
How is the perf of these builds?

psi29a:
Oh... can we tell msvc make pdb files for release builds? That would be perfect...

LennyPenny:
@psi29a https://docs.microsoft.com/en-us/cpp/build/reference/how-to-debug-a-release-build

seems to be possible, just pass these compiler flags

akortunov:
> How is the perf of these builds?

A quite well.
About PDBs:
1. I guess that PDB generation will take some additional time.
2. We need to enable them only for AppVeyor, not for develpoers.

psi29a:
@akortunov what do you mean? appveyor is purely for CI, we almost never used the builds. The only reason they were set to debug was because with msvc2013 and earlier, it took longer to build release builds in comparison to debug.

So adding pdb files along with release builds for people to test out the PR before we move is ideal. That way testers and developers can use the same builds and if something bad happens, devs have the pdb files on hand.

akortunov:
> seems to be possible, just pass these compiler flags

Where to?

psi29a:
I suspect some CMake magic like we do for gcc, but for msvc. On linux/unix, we create release builds with debug info. Might as well do the same for msvc, add an additional compiler flag option.

https://stackoverflow.com/questions/14148374/how-do-you-set-properties-in-a-vcproj-file-with-cmake

psi29a:
What could possibly go wrong...

AnyOldName3:
The correct way to get PDB files for a release build is to build the RelWithDebInfo configuration instead of the Release configuration. It should be exactly the same process as changing from a debug build to a release build in the first place was.

psi29a:
We do that already in cmake:

```
# If the user doesn't supply a CMAKE_BUILD_TYPE via command line, choose one for them.
IF(NOT CMAKE_BUILD_TYPE)
    SET(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
        "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel."
        FORCE)
    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS None Debug Release RelWithDebInfo MinSizeRel)
ENDIF()
```

cmake and/or msvc apparently doesn't play along... so we have to add flags to get msvc to produce pdb files. Or are we going about it the wrong way?

AnyOldName3:
The Appveyor YAML file specifies that the configuration should be `Release` in a bunch of places.

* https://github.com/OpenMW/openmw/pull/1718/files#diff-180360612c6b8c4ed830919bbb4dd459R21 This should be changed from `Release` to `RelWithDebInfo`
* https://github.com/OpenMW/openmw/pull/1718/files#diff-180360612c6b8c4ed830919bbb4dd459R59 This should be changed from `%APPVEYOR_BUILD_FOLDER%\MSVC%msvc%_64\Release\` to `%APPVEYOR_BUILD_FOLDER%\MSVC%msvc%_64\RelWithDebInfo\`
* https://github.com/OpenMW/openmw/pull/1718/files#diff-180360612c6b8c4ed830919bbb4dd459R60  This needs the same change as the line above.
* https://github.com/akortunov/openmw/blob/48e15dabc3c70a5bfcc35c1277d188818759c42b/appveyor.yml#L51 We need to set the configuration here as the CI prebuild script defaults to Debug.
* https://github.com/OpenMW/openmw/blob/master/CI/build.msvc.sh#L54 This does nothing unless the CI prebuild script is called with the configuration argument set to RelWithDebInfo.
* The CI script always sets the `CMAKE_BUILD_TYPE ` and defaults to Debug. This code never gets run.
```CMAKE
# If the user doesn't supply a CMAKE_BUILD_TYPE via command line, choose one for them.
IF(NOT CMAKE_BUILD_TYPE)
    SET(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
        "Choose the type of build, options are: None(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel."
        FORCE)
    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS None Debug Release RelWithDebInfo MinSizeRel)
ENDIF()
```
* https://github.com/OpenMW/openmw/pull/1718/files#diff-af3b638bc2a3e6c650974192a53c7291R363 You know the copypasta about the guy's grandpa quitting smoking so he could see his grandchildren graduate university, but dying of lung cancer anyway and the grandchild being upset and never smoking but regretting it because reading the above post gave him cancer anyway? </overlyharsh> This is Very Bad CMake. We already have a lot of Very Bad CMake, but overriding one of the things it's already getting right is a bad thing.

I think that with all of this fixed, we should start getting unadulterated RelWithDebInfo builds without turning Release builds into a duplicate copy of RelWithDebInfo.

AnyOldName3:
By the way, the artefacts as they're produced right now don't run. As the CI prebuild script is being called with the wrong configuration, it's only putting the dependencies' DLLs in the Debug folder.

A Windows OpenMW dev should probably have been consulted about this before it got merged.

psi29a:
Paging dr. @ananace ...

@AnyOldName3 you sure about that? The build directory doesn't contain any DLL/libs, never did. When @akortunov tested, he said they did run. I think you missed the conversation where we talked about this and the fact that if anyone wants to test the artifacts, that they must installed the necessary libs themselves.  We can package them together in a later PR.

ananace:
Well, this should really have been a `RelWithDebInfo` build to get the PDBs, modifying build flags on the `Release` builds is not at all what we want. Especially since release builds apply destructive optimizations that make the PDBs all but useless anyway. And https://github.com/OpenMW/openmw/blob/master/CI/before_script.msvc.sh#L712-L713 needs to be followed as well, otherwise DLLs won't be copied over for the packaging, and so be useless for the non-developers - a.k.a. the people who would be most helped by having pre-built artefacts.

And of course, because the builds are built as unity builds in order to fit in the time limit, they might not even run anyway.

AnyOldName3:
@psi29a 
> the fact that if anyone wants to test the artifacts, that they must installed the necessary libs themselves. 

Installing library DLLs in Windows involves meddling with things that you're not supposed to and has the potential to break things. It's possible to do it, but it's very bad practice. Windows doesn't really do shared libraries, and all Windows software is supposed to come with all the libraries it uses.

psi29a:
@AnyOldName3 honestly... that is really a none issue as far as I'm concerned. You know the risks, I know them, those that want to test out feature branches know them as well because they are either fellow developers or those that want to help.

That doesn't mean we shouldn't try to package up the libs in the artifacts zip as well to make things easier on testers. This again makes installation of the libs yourself kinda irrelevant. Nothing a new PR can't fix.

akortunov:
We already pack DLLS to artifacts in #1721.
Also I suggest to continue discussion in opened PR thread.

AnyOldName3:
I was just mentioning to Psi that shared libraries on Windows should never be attempted by anyone, no matter how clever and organised they think they are. This is only really a tangential discussion and isn't relevant to the PR as we've already resolved the issue, but there's a chance that it's something Psi would encounter again in the future somehow, so I just thought it best to mention it.

PR #1720, opened 2018-05-26
Rework spell VFX management

akortunov:
Continue of #1573.
Implements [feature #1645](https://gitlab.com/OpenMW/openmw/issues/1645).

The only thing left is to optimize objects update:
> We should really avoid iterating all objects every frame because there are thousands of them in a normal scene. I remember we had a performance issue caused by the same thing in PhysicsSystem::animateCollisionShapes and it was fixed by tracking animated shapes in a separate container.
Usually the osg::UpdateCallback is a very nice way to do updates every frame. Maybe the updateEffects should be ported to an UpdateCallback?

So the main idea: when we add spell effect to mesh, attach the NodeCalback to it. This callback should update VFX effect state, then mark itself as finished.
Once per frame we remove finished callbacks via visitor inside Animation::runAnimation().
But this implementation has drawback: we do not run runAnimation() for non-animated objects, so we do not remove finished callbacks for statics objects yet.

Note: particle scaling is incorrect due to [bug #3559](https://gitlab.com/OpenMW/openmw/issues/3559).

### DIFF
--- a/apps/openmw/mwmechanics/aicast.cpp
+++ b/apps/openmw/mwmechanics/aicast.cpp
@@ -42,7 +42,19 @@ bool MWMechanics::AiCast::execute(const MWWorld::Ptr& actor, MWMechanics::Charac
             return false;
         }
 
-        osg::Vec3f dir = target.getRefData().getPosition().asVec3() - actor.getRefData().getPosition().asVec3();
+        osg::Vec3f targetPos = target.getRefData().getPosition().asVec3();
+        if (target.getClass().isActor())
+        {
+            osg::Vec3f halfExtents = MWBase::Environment::get().getWorld()->getHalfExtents(target);
+            targetPos.z() += halfExtents.z() * 2 * 0.75f;
+        }
+
+        osg::Vec3f actorPos = actor.getRefData().getPosition().asVec3();
+        osg::Vec3f halfExtents = MWBase::Environment::get().getWorld()->getHalfExtents(actor);
+        actorPos.z() += halfExtents.z() * 2 * 0.75f;
+
+        osg::Vec3f dir = targetPos - actorPos;
+
         bool turned = smoothTurn(actor, getZAngleToDir(dir), 2, osg::DegreesToRadians(3.f));
         turned &= smoothTurn(actor, getXAngleToDir(dir), 0, osg::DegreesToRadians(3.f));
 
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -24,6 +24,7 @@
 #include "../mwworld/inventorystore.hpp"
 
 #include "../mwrender/animation.hpp"
+#include "../mwrender/vismask.hpp"
 
 #include "npcstats.hpp"
 #include "actorutil.hpp"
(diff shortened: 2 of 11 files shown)


apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -1009,18 +1015,63 @@ namespace MWMechanics
...
-                    castStatic = store.get<ESM::Static>().find ("VFX_DefaultCast");
+                const ESM::Race *race =
+                    esmStore.get<ESM::Race>().find(npc->mRace);
 
-                std::string texture = effect->mParticle;
+                scale = npc->isMale() ? race->mData.mHeight.mMale : race->mData.mHeight.mFemale;
+            }
+            else
+            {
+                osg::BoundingBox bounds = MWBase::Environment::get().getWorld()->getObjectBounds(mCaster);
```
scrawl: That's a bit better than before, but is still a rendering detail. Would it be possible to pass the object or its model name to 'RenderingManager::spawnEffect' and then compute the scale in there?

scrawl:
> >     Usually the osg::UpdateCallback is a very nice way to do updates every frame. Maybe the updateEffects should be ported to an UpdateCallback?

> But I do not know how to implement it yet. Any ideas?

There are plenty of examples for UpdateCallback. You derive from that class and then add the callback to an osg::Node.

A small problem is that callbacks can't remove themselves when they're done, because an object can't delete itself. We could remove the finished callbacks in runAnimation. For objects that don't call runAnimation the finished callbacks will remain there, but that's not a huge issue I think. They're not very expensive and will be removed anyway once you leave the scene.

apps/openmw/mwmechanics/spellcasting.cpp
akortunov: Looks Animation::addEffect() does not use RenderingManager, so this code should be somewhere in more common place. How about sceneManager?

akortunov:
To which node I should apply the callback? To VFX one?

akortunov:
I wonder how to remove effect correctly.
I tried to use something like that:

```
    class UpdateVfxCallback : public osg::NodeCallback
    {
    public:
        UpdateVfxCallback(EffectParams params)
        : mParams(params)
        {
        }

        virtual void operator()(osg::Node* node, osg::NodeVisitor* nv)
        {
            traverse(node, nv);

            double duration = nv->getFrameStamp()->getSimulationTime();

            mParams.mAnimTime->addTime(duration);
            if (mParams.mAnimTime->getTime() >= mParams.mMaxControllerLength)
            {
                if (mParams.mLoop)
                {
                    // Start from the beginning again; carry over the remainder
                    // Not sure if this is actually needed, the controller function might already handle loops
                    float remainder = mParams.mAnimTime->getTime() - mParams.mMaxControllerLength;
                    mParams.mAnimTime->resetTime(remainder);
                }
                else
                {
                    // Remove effect
                    mParams.mObjects.reset();
                }
            }
        }

    private:
        EffectParams mParams;
    };
```

But I can not just remove node inside traversal due to iterator invalidation.
How can I remove the node correctly?

akortunov:
I uploaded a reworked version, but I need a review from someone familiar with OSG.
@AnyOldName3, can you take a look, please?

akortunov:
To handle non-animated objects, I guess I can use another NodeCallback, attached to root node (mInsert?).
I will try do add this callback when we add effect to object.

apps/openmw/mwmechanics/aicast.cpp
```diff
@@ -42,7 +42,19 @@ bool MWMechanics::AiCast::execute(const MWWorld::Ptr& actor, MWMechanics::Charac
             return false;
         }
 
-        osg::Vec3f dir = target.getRefData().getPosition().asVec3() - actor.getRefData().getPosition().asVec3();
+        osg::Vec3f targetPos = target.getRefData().getPosition().asVec3();
+        if (target.getClass().isActor())
+        {
+            osg::Vec3f halfExtents = MWBase::Environment::get().getWorld()->getHalfExtents(target);
+            targetPos.z() += halfExtents.z() * 2 * 0.75f;
```
AnyOldName3: Is this definitely supposed to be `... * 2 * 0.75`? The int is just going to get promoted to a float, so it's exactly equivalent to `... * 1.5`, isn't it?

akortunov: We already use such notation. It means "0.75 of actor's height".
If we do not want to use it, I can change it in all cases.

AnyOldName3: I was just making sure it wasn't a typo, for example if you'd meant `... * 2 + 0.75`.

zinnschlag:
Missing changelog entry.

akortunov:
Added changelog entry.

akortunov:
Is there any objections?

Capostrophic:
scrawl would most probably have had some, but he's not here. As far as I can tell the code looks good -- but I can't tell much. I'll test this just in case the moon phase is not right today.

Capostrophic:
Ok, seems to work good. All objects have the proper VFX and look like in vanilla, AI spellcasting didn't break, no crashes occur, at least I couldn't find any.

Proper scaling of the VFX when a netch casts a spell, compared to master, is nice.

PR #1721, opened 2018-05-26
[Do not merge yet] AppVeyor: build with RelWithDebInfo configuration

akortunov:
Try to build artifacts with RelWithDebInfo configuration.


AnyOldName3:
The current AppVeyor issue is probably because this workaround in the top-level `CMakeLists.txt` hasn't been applied to as many situations as it needs to be:

```CMAKE
if (BUILD_OPENMW)
        # Very specific issue this, only needed on 32-bit VS2015 during unity builds.
        if (MSVC_VERSION GREATER 1800 AND CMAKE_SIZEOF_VOID_P EQUAL 4 AND OPENMW_UNITY_BUILD)
            set_target_properties(openmw PROPERTIES COMPILE_FLAGS "${WARNINGS} ${MT_BUILD} /bigobj")
        else()
            set_target_properties(openmw PROPERTIES COMPILE_FLAGS "${WARNINGS} ${MT_BUILD}")
        endif()
    endif()
```

AnyOldName3:
Only turn it on when we're doing a unity build. Unity builds combine a bunch of `.cpp` files into one big one, so lead to more stuff ending up in each `.obj` file, and passing the limit, but when we're not doing a unity build, we're not near the limit and `/bigobj` will probably slightly increase build times on Windows.

AnyOldName3:
After all, I've been doing all my shadow development with non-unity RelWithDebInfo builds and have never hit the `.obj` size limit.

akortunov:
> Only turn it on when we're doing a unity build. 

Now better?

AnyOldName3:
Once AppVeyor is done, if it passes and the artefact runs without any extra meddling (including manually adding DLLs) this should be fine. I'll do a thorough check later.

akortunov:
> (including manually adding DLLs)

There are no DLLs in package yet. From where can I grab them?

ananace:
@akortunov https://github.com/OpenMW/openmw/blob/master/CI/before_script.msvc.sh#L712-L714

akortunov:
> https://github.com/OpenMW/openmw/blob/master/CI/before_script.msvc.sh#L712-L714

So there should be something like this?
`if [ -z $CI ] || [ -n $APPVEYOR ]; then`

ananace:
`CI` and `APPVEYOR` are the same thing for CI builds, just remove the if statement entirely. That code was originally disabled because the DLL copying takes a while, and the CI builds were really tight for time.

AnyOldName3:
The important changes have been made and despite being unity builds, the artefacts appear to run just fine. I'm assuming that the only potential issue is if something gets defined in two different places and they end up in the same combined file, but I'd have thought that would already cause a multiple definition error in AppVeyor.

The one change I'd still consider making is giving the artefact file name some information about what was built such as the commit hash and/or the PR title. If users are going to be downloading a bunch of these it should help keep things organised.

akortunov:
> The one change I'd still consider making is giving the artefact file name some information about what was built such as the commit hash and/or the PR title.

As for me, PR title can be too long to put it to file names. IMO, PR number + commit hash is a good start.
According to [documentation](https://www.appveyor.com/docs/environment-variables/), AppVeyour has many suitable environmental variables. Any ideas?

ananace:
@AnyOldName3 Multiple definitions - usually of global data - indeed results in compilation errors for the unity builds, not that it's too hard to fix that particular issue. The resulting artefacts used to be completely broken though, my assumption was that the larger compile units caused some failure in the linker, resulting in code that wouldn't work due to "dead" code removal having thrown away live code. Might be that 2015 and 2017 are both better with the whole thing, hence the working builds.

@akortunov My old posts when I was doing these CI builds at first have extensive benchmarks on the whole thing. There's still code in the CI script for doing subproject builds, as just compiling the OpenMW components was almost enough to hit the compile time limit without Unity builds back then.  
The builds probably run a lot faster today, because of newer MSVC versions, and because of the stronger CI hardware AppVeyor got their hands on.

Still, expect a 50-200% build time increase if disabling unity builds, as MSVC sucks time-wise at creating compilation units.

akortunov:
> Still, expect a 50-200% build time increase if disabling unity builds

Well, I'd say 35-40%.
MSVC2017 needs nearly 25 milutes for normal build, MSVC2015 needs nearly 25 minutes for unity build, 35 minutes for normal build.

Last time we checked, MSVC2015 needed more time to puild PDB files than MSVC2017.

akortunov:
Well, an expected results: without PDBs both normal builds have duration about 24-25 minutes.
With unity build duration is 17-18 minutes for both MSVC2015 and MSVC2017.

AnyOldName3:
@akortunov If I'm testing something, it's probably something I've just built myself, so they wouldn't help me much, but a few other projects that I'm familiar with release them, so there must be some advantage.

akortunov:
@ananace, @psi29a, is 25 minutes for normal release build acceptable for us?

psi29a:
For me, 25m is fine as it is the whole project.  Appveyor is quite generous and has a default timeout of just over an hour, I would never want us to push that limit. I am happy we're under 30m. We'll investigate again if and when we get there.

psi29a:
@ananace merge?

ananace:
I don't see any issue with merging it, the builds are working after all.

If anything's missing there's always the possibility of opening another PR to fix that afterwards.

PR #1722, opened 2018-05-26
Improve the handling of empty class name and description in verifier

Capostrophic:
Restore empty description warning, but put it behind a "class is playable" check, as suggested by scrawl in #1699.


lysol90:
Huh, back already @scrawl or just a temporary visit? :)

PR #1724, opened 2018-06-01
Add pursue package only when crime was reported

akortunov:
Fixes [bug #4433](https://bugs.openmw.org/issues/4433).

The main idea: iterate over neighbors twice. In the first run check if the crime was reported by anyone, in the second check update AI status.

To test this PR you can execute "setalarm 0" in any shop on both guard and shopkeeper and try to steal something.


zinnschlag:
For future reference: We are currently in the process of making a change to the PR procedure (just test-run, will be reviewed again after the 0.45 release). For any new PR (related to the issue), please add the issue description to the CHANGELOG.md file.

PR #1725, opened 2018-06-01
[Regression] Fix regressions in spell resistance

akortunov:
Fixes some regressions from commit 2dca389a3b00dd32697bd16703c47e29074dea6d:

1. Fixes casting spells on non-actor objects.
2. Fixes casting in GodMode: discard only harmful effects.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -221,7 +221,6 @@ namespace MWMechanics
         if (effects)
             magicEffects = effects;
 
-        float resisted = 0;
         // Effects with no resistance attribute belonging to them can not be resisted
         if (ESM::MagicEffect::getResistanceEffect(effectId) == -1)
             return 0.f;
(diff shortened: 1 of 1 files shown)


Capostrophic:
>Fixes Open and Lock spells by checking if spell target has a creature stats.

Wouldn't putting the check for whether there is a resistance effect at all earlier suffice?

>Fixes casting in GodMode: discard only harmful effects.

I'd argue this is not necessary, since vanilla ignores beneficial effects.

I would have taken a closer look at !absorbed condition, I think I consolidated that incorrectly.

akortunov:
> I'd argue this is not necessary, since vanilla ignores beneficial effects.

?
In upstream master player can not receive **any** effect in god mode, even from potions.
I doubt that vanilla ignores beneficial effects in this way.

akortunov:
Updated. No we do not try to reflect/resist if the target is non-actor.
To test this you can try to unlock door with Open spell or just throw fireball to it.

Capostrophic:
>I doubt that vanilla ignores beneficial effects [even from potions]
Hm. True, it doesn't.

Ferk:
> Fixes casting spells on non-actor objects.

I don't know if it's related, but I noticed I cannot cast "Open" spells on Vivec's palace door when the spell is set "on Target" (it does work fine with "on touch").

akortunov:
> I don't know if it's related, but I noticed I cannot cast "Open" spells on Vivec's palace door when the spell is set "on Target" (it does work fine with "on touch").

There is an issue with collisions near Vivec palace ([bug #2849](https://gitlab.com/OpenMW/openmw/issues/2849)). IIRC, touch spells just use focused object as target instead, so they work.

PR #1727, opened 2018-06-02
Make weapon durability short unsigned (bug #4435)

Capostrophic:
[Bug #4435](https://bugs.openmw.org/issues/4435). Looks like only weapon durability has short type.


akortunov:
Works for me. 
I wonder why BS programmers used fields of different size to store the same thing.

psi29a:
Thanks, merging. Would anyone be willing to go through and validate other fields?

PR #1731, opened 2018-06-07
Add easily understood error messages to the prebuild script instead o…

AnyOldName3:
…f vague/silent failures.

This should make it harder to break the script by accident.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -1,5 +1,23 @@
 #!/bin/bash
 
+MISSINGTOOLS=0
+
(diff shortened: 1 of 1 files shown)


AnyOldName3:
If the pre-3.9 CMake change needed a changelog entry, does this mean we need one for this, too?

psi29a:
Was there a bug  file for this?

AnyOldName3:
Lots of things with the script, including this, are covered in https://bugs.openmw.org/issues/4442

psi29a:
Then once that bug is closed, it can be added to the changelog... it's up to you if you wish to add a comment not related to a bug/issue. I'm not going that anal-retentive about it.

PR #1732, opened 2018-06-07
Use link to OpenMW Wiki page instead of direct NifSkope page

akortunov:
Created as replacement of #1540.

The main idea: provide a link to OpenMW Wiki [page](https://wiki.openmw.org/index.php?title=Tools#NifSkope) instead of direct NifSkope download page.
In this case we can store additional info on that Wiki page instead of project docs.


psi29a:
Thanks.

PR #1733, opened 2018-06-07
fix #3137

wareya:
This fixes http://bugs.openmw.org/issues/3137 but another jumping-related issue is relevant so I'm going to go over what I see:

http://bugs.openmw.org/issues/3137 actually has two causes:

1) Being clamped to the ground by the stair/slope code because the actor encountered a wall
2) Having the vertical part of motion cancelled because another part of the code assumed it as accounted for by the stair/slope code

... But the wall was completely vertical so the stair/slope code didn't accomplish anything.

This commit addresses those two causes. Addressing them probably causes other problems, this movement solver seems very elaborate. I have no idea what those two changes will do to steep sloped walls, for instance.

http://bugs.openmw.org/issues/2256 happens the way it does because of the stair stepping code is effectively responsible for placing the actor back on the "ground". If you entirely comment out the block that runs the "stair" code, actors will take an extremely long time to register that they landed on the ground, if they do so at all. This is why my check for not running the stair/slope code also checks for vertical motion direction too, not just whether the actor is in the air.

Also, there are probably cases when the actor **should** be able to climb up a "stair" while in the air, unless vanilla Morrowind doesn't do so. The reason that the "stair" code isn't run when moving up in the air in this commit is because the stair code behaves incorrectly when running into perfectly vertical walls, not because actors shouldn't be able to climb up stairs while in the air. As such, this commit is almost certainly the wrong fix, but I don't know where in the stairs code the problem is.

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -168,7 +168,7 @@ namespace MWPhysics
             mTracer.doTrace(mColObj, tracerPos, tracerPos + toMove, mColWorld);
             if(mTracer.mFraction < std::numeric_limits<float>::epsilon())
                 return false; // didn't even move the smallest representable amount
-            
+
             // FIXME: non-levitating aerial actors should not step down to a lower location than their initial location
             /*
              * Try moving back down sStepSizeDown using stepper.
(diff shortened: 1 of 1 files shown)


wareya:
Side note, I tested out cylinders (instead of capsules) and they worked fine, even for climbing most stairs. I understand the motivation for capsules, but vanilla Morrowind seems to use a shape with a flat bottom, and I think it's worth reconsidering at least as a configuration option or something. I thought about tapered cylinders or cylinders with semispheres on the ends but it looks like they're difficult to make in bullet, without using a pointcloud or something.

psi29a:
Is there any measurable performance change?

Capostrophic:
The switch to capsule shapes was for a reason; with cylinder collision shapes, actors could float above visible geometry and couldn't pass through some staircase passageways.

wareya:
That makes sense. Vanilla's movement solver is very buggy and has a lot of weird behaviors (like sliding off of thin posts).

wareya:
Fix a related issue involving GroundOffset that would interfere with these fixes in certain situations, as well as cause the player to get stuck in certain situations even in the current master. Also improved the fix for not sliding up steep slopes that the actor can't walk on. Ready for review.

wareya:
I'm rewriting this PR.

PR #1736, opened 2018-06-07
Make dialogue choices trigger goodbye if Goodbye option is present (bug #3897)

Capostrophic:
[Bug report 3897.](https://bugs.openmw.org/issues/3897)

The window closes and nothing happens in this case.

I added Goodbye to the script text of "did you bring the mushrooms" topic info of Ajira. On master Ajira answers the choice and Goodbye option is still present to be chosen, with PR the dialogue window closes immediately upon any choice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 0.45.0
 ------
-    Feature #3897: Have Goodbye give all choices the effects of Goodbye
+    Bug #3897: Have Goodbye give all choices the effects of Goodbye
     Bug #4293: Faction members are not aware of faction ownerships in barter
     Bug #4426: RotateWorld behavior is incorrect
     Bug #4433: Guard behaviour is incorrect with Alarm = 0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 0.45.0
 ------
+    Feature #3897: Have Goodbye give all choices the effects of Goodbye
     Bug #4293: Faction members are not aware of faction ownerships in barter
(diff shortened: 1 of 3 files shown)


MiroslavR:
Are you sure the scripts associated with those choices are not supposed to be executed?

Capostrophic:
Apparently the report is erroneous.

If a Goodbye command follows Choice in the dialogue response, it should practically be ignored - no option gets added and the choices don't result in a goodbye event, and if Choice follows Goodbye, then no choices should get added into the list.

Capostrophic:
Apparently vanilla Goodbye-Choice relationship is completely different from the one described in the report and it's erroneous. I'll try again tomorrow.

zinnschlag:
Can we close this PR then?

Capostrophic:
Well, go ahead, though I wanted to repurpose it. The feature request should be rejected, though, because that's definitely not what's happening in the vanilla. Goodbye option and choices are mutually exclusive there: the one earlier in the dialogue response will appear and the latter won't. If Goodbye is present in script text later than choices, choices won't suddenly change their function; it's practically ignored, and if choices are present later, they won't get added and the player is forced to choose goodbye.

zinnschlag:
Is there a reason why you re-opened this PR?

Capostrophic:
OpenMW-CS-created plugin was giving incorrect results in the original engine, thus making the behavior invalid - Morrowind didn't handle the return symbols used in the edited script text properly. See 4443 discussion. rot tor's plugins work properly; also, it seems that the choice scripting is not handled either.

PR #1737, opened 2018-06-08
Improve MRK NiStringExtraData handling (bug #4419)

akortunov:
Fixes [bug #4419](https://bugs.openmw.org/issues/4419).

Instead if ignoring any node with MRK NiStringExtraData, ignore only NiTriShape with name, starting with "Tri EditorMarker" (case-insensitive).


zinnschlag:
This looks like the kind of stuff we need to throw to a wider user base to test it properly. I am not familiar enough with MW animation to review this properly, but unless someone else finds a problem, I am inclined to merge.

psi29a:
Agreed.

PR #1739, opened 2018-06-08
fix #3876 and #3993

wareya:
Apparently vanilla Morrowind doubles the resolution of the blendmap. This fixes #3993.

Doubling the resolution of the blendmap makes it possible to nudge the blendmap texture uv transformation matrix without causing seams. Doing this without the doubling causes visible texturing seams. This fixes #3876.

Comparison images: https://imgur.com/a/3aYHR1i

### DIFF
--- a/components/esmterrain/storage.cpp
+++ b/components/esmterrain/storage.cpp
@@ -434,19 +434,14 @@ namespace ESMTerrain
         const int imageScaleFactor = 2;
         const int blendmapImageSize = blendmapSize * imageScaleFactor;
         
-        
-        const bool largeImage = true;
-
         for (int i=0; i<numBlendmaps; ++i)
         {
             GLenum format = pack ? GL_RGBA : GL_ALPHA;
 
             osg::ref_ptr<osg::Image> image (new osg::Image);
-            if(!largeImage)
-                image->allocateImage(blendmapSize, blendmapSize, 1, format, GL_UNSIGNED_BYTE);
-            else
-                image->allocateImage(blendmapImageSize, blendmapImageSize, 1, format, GL_UNSIGNED_BYTE);
+            image->allocateImage(blendmapImageSize, blendmapImageSize, 1, format, GL_UNSIGNED_BYTE);
             unsigned char* pData = image->data();
+            
             for (int y=0; y<blendmapSize; ++y)
             {
                 for (int x=0; x<blendmapSize; ++x)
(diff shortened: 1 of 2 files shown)


components/esmterrain/storage.cpp
```diff
@@ -430,15 +430,23 @@ namespace ESMTerrain
 
         // Second iteration - create and fill in the blend maps
         const int blendmapSize = (realTextureSize-1) * chunkSize + 1;
+        // We need to upscale the blendmap 2x with nearest neighbor sampling to look like Vanilla
+        const int imageScaleFactor = 2;
+        const int blendmapImageSize = blendmapSize * imageScaleFactor;
+        
+        
+        const bool largeImage = true;
```
Capostrophic: Well, this boolean is not very useful.

components/esmterrain/storage.cpp
```diff
@@ -448,14 +456,26 @@ namespace ESMTerrain
...
+                    int alpha = (blendIndex == i) ? 255 : 0;
+                    
+                    if(!largeImage)
+                        pData[((blendmapSize - y - 1)*blendmapSize + x)*channels + channel] = alpha;
                     else
-                        pData[(blendmapSize - y - 1)*blendmapSize*channels + x*channels + channel] = 0;
+                    {
+                        int realY = (blendmapSize - y - 1)*imageScaleFactor;
+                        int realX = x*imageScaleFactor;
+                        if(true)
```
Capostrophic: Neither does this condition have any worth. I presume they were for debug.

wareya: This one's for text alignment. Is that bad?

akortunov: Usually we do not use redundant conditions.
And I have a suspicion that this block really can be a some kind of "if-then-else".

wareya: Now that I'm fully awake again I realized that the conditions aren't even necessary as long as blendmapImageSize is a multiple of blendmapSize. I wrote them when I was still trying to make blendmapImageSize = blendmapSize*2-1 work.

components/esmterrain/storage.cpp
```diff
@@ -430,15 +430,18 @@ namespace ESMTerrain
 
         // Second iteration - create and fill in the blend maps
         const int blendmapSize = (realTextureSize-1) * chunkSize + 1;
-
+        // We need to upscale the blendmap 2x with nearest neighbor sampling to look like Vanilla
+        const int imageScaleFactor = 2;
+        const int blendmapImageSize = blendmapSize * imageScaleFactor;
+
```
Capostrophic: Redundant whitespace is the enemy.

components/esmterrain/storage.cpp
```diff
@@ -430,15 +430,18 @@ namespace ESMTerrain
...
         for (int i=0; i<numBlendmaps; ++i)
         {
             GLenum format = pack ? GL_RGBA : GL_ALPHA;
 
             osg::ref_ptr<osg::Image> image (new osg::Image);
-            image->allocateImage(blendmapSize, blendmapSize, 1, format, GL_UNSIGNED_BYTE);
+            image->allocateImage(blendmapImageSize, blendmapImageSize, 1, format, GL_UNSIGNED_BYTE);
             unsigned char* pData = image->data();
-
+
```
Capostrophic: also here

components/esmterrain/storage.cpp
```diff
@@ -448,14 +451,18 @@ namespace ESMTerrain
                     int layerIndex = textureIndicesMap.find(id)->second;
                     int blendIndex = (pack ? static_cast<int>(std::floor((layerIndex - 1) / 4.f)) : layerIndex - 1);
                     int channel = pack ? std::max(0, (layerIndex-1) % 4) : 0;
-
-                    if (blendIndex == i)
-                        pData[(blendmapSize - y - 1)*blendmapSize*channels + x*channels + channel] = 255;
-                    else
-                        pData[(blendmapSize - y - 1)*blendmapSize*channels + x*channels + channel] = 0;
+
```
Capostrophic: here

components/esmterrain/storage.cpp
```diff
@@ -448,14 +451,18 @@ namespace ESMTerrain
...
                     int blendIndex = (pack ? static_cast<int>(std::floor((layerIndex - 1) / 4.f)) : layerIndex - 1);
                     int channel = pack ? std::max(0, (layerIndex-1) % 4) : 0;
-
-                    if (blendIndex == i)
-                        pData[(blendmapSize - y - 1)*blendmapSize*channels + x*channels + channel] = 255;
-                    else
-                        pData[(blendmapSize - y - 1)*blendmapSize*channels + x*channels + channel] = 0;
+                    
+                    int alpha = (blendIndex == i) ? 255 : 0;
+
```
Capostrophic: here

components/esmterrain/storage.cpp
```diff
@@ -448,14 +451,18 @@ namespace ESMTerrain
...
-                    if (blendIndex == i)
-                        pData[(blendmapSize - y - 1)*blendmapSize*channels + x*channels + channel] = 255;
-                    else
-                        pData[(blendmapSize - y - 1)*blendmapSize*channels + x*channels + channel] = 0;
+                    
+                    int alpha = (blendIndex == i) ? 255 : 0;
+                    
+                    int realY = (blendmapSize - y - 1)*imageScaleFactor;
+                    int realX = x*imageScaleFactor;
+
```
Capostrophic: and here

psi29a:
Nice :)

zinnschlag:
For future reference: Please include an entry to the changelog in your PR

PR #1740, opened 2018-06-09
Decompress cursors using SDL software renderer on Mac or if OSG >= 3.5.8 or if OPENMW_DECOMPRESS_TEXTURES is set

nikolaykasyanov:
This change fixes [#4424](https://bugs.openmw.org/issues/4424).
It requires OSG 3.6 though & has to be tested on other platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
     Bug #4433: Guard behaviour is incorrect with Alarm = 0
     Bug #4443: Goodbye option and dialogue choices are not mutually exclusive 
     Feature #4444: Per-group KF-animation files support
-    Bug #4424: [macOS] Cursor is either empty or garbage when compiled against macOS 10.13 SDK
 
 0.44.0
 ------
@@ -109,6 +108,7 @@
     Feature #4423: Rebalance soul gem values
     Task #4015: Use AppVeyor build artifact features to make continuous builds available
     Editor: New (and more complete) icon set
+    Bug #4424: [macOS] Cursor is either empty or garbage when compiled against macOS 10.13 SDK
 
 0.43.0
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,7 @@
     Bug #4412: openmw-iniimporter ignores data paths from config
     Bug #4413: Moving with 0 strength uses all of your fatigue
     Bug #4420: Camera flickering when I open up and close menus while sneaking
+    Bug #4424: [macOS] Cursor is either empty or garbage when compiled against macOS 10.13 SDK
     Bug #4435: Item health is considered a signed integer
     Bug #4441: Adding items to currently disabled weapon-wielding creatures crashes the game
     Feature #1786: Round up encumbrance value in the encumbrance bar
@@ -108,7 +109,6 @@
     Feature #4423: Rebalance soul gem values
     Task #4015: Use AppVeyor build artifact features to make continuous builds available
     Editor: New (and more complete) icon set
-    Bug #4424: [macOS] Cursor is either empty or garbage when compiled against macOS 10.13 SDK
 
 0.43.0
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
     Bug #4433: Guard behaviour is incorrect with Alarm = 0
     Bug #4443: Goodbye option and dialogue choices are not mutually exclusive 
     Feature #4444: Per-group KF-animation files support
+    Bug #4424: [macOS] Cursor is either empty or garbage when compiled against macOS 10.13 SDK
 
 0.44.0
 ------
--- a/components/sdlutil/sdlcursormanager.cpp
+++ b/components/sdlutil/sdlcursormanager.cpp
@@ -221,9 +221,7 @@ namespace SDLUtil
     }
 
 #if OPENMW_USE_SOFTWARE_CURSOR_DECOMPRESSION
-    typedef std::unique_ptr<SDL_Surface, decltype(&SDL_FreeSurface)> SDLSurfacePtr;
-
-    SDLSurfacePtr decompress(osg::Image* source, int rotDegrees)
+    SurfaceUniquePtr decompress(osg::Image* source, int rotDegrees)
     {
         int width = source->s();
         int height = source->t();
@@ -265,7 +263,7 @@ namespace SDLUtil
         SDL_FreeSurface(cursorSurface);
         SDL_DestroyRenderer(renderer);
 
-        return SDLSurfacePtr(targetSurface, SDL_FreeSurface);
+        return SurfaceUniquePtr(targetSurface, SDL_FreeSurface);
     }
 #endif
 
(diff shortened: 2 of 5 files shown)


components/sdlutil/sdlcursormanager.cpp
```diff
@@ -25,133 +27,57 @@ USE_GRAPHICSWINDOW()
...
-        glReadPixels(0, 0, width, height, GL_RGBA, GL_UNSIGNED_BYTE, resultImage->data());
-
-        geom->releaseGLObjects();
-        source->releaseGLObjects();
-        texture->releaseGLObjects();
+        resultImage->setFileName(source->getFileName());
+        resultImage->allocateImage(width, height, source->r(), GL_RGBA, GL_UNSIGNED_BYTE);
+        assert(resultImage->getImageSizeInBytes() == (targetSurface->pitch * targetSurface->h));
+        memcpy(resultImage->data(), targetSurface->pixels, resultImage->getImageSizeInBytes());
+        SDL_FreeSurface(targetSurface);
```
nikolaykasyanov: We may actually return the surface instead of an `osg::Image` to avoid another conversion in `_createCursorFromResource`.

psi29a:
I would love to merge this, but we're stuck with OSG-3.4 for just a bit longer. Downstream still only supports OSG-3.4... I will feel more comfortable when there is at least one release where they ship OSG-3.6.

Anyway you can hold on to this and cherry-pick for OSX release for now?

nikolaykasyanov:
@psi29a I'd keep the old implementation and guard the new one with OSG version check macro & `__APPLE__` macro, what do you think? I'm using our OSG fork for macOS release builds anyway and it [includes](https://forum.openmw.org/viewtopic.php?f=6&t=5202#p54996) the required OSG commit.

components/sdlutil/sdlcursormanager.cpp
```diff
@@ -25,133 +27,57 @@ USE_GRAPHICSWINDOW()
...
+        SDL_Texture *cursorTexture = SDL_CreateTextureFromSurface(renderer, cursorSurface);
+
+        SDL_RenderCopyEx(renderer, cursorTexture, NULL, NULL, rotDegrees, NULL, SDL_FLIP_NONE);
+
+        SDL_DestroyTexture(cursorTexture);
+        SDL_FreeSurface(cursorSurface);
+        SDL_DestroyRenderer(renderer);
 
         osg::ref_ptr<osg::Image> resultImage = new osg::Image;
+        resultImage->setFileName(source->getFileName());
```
nikolaykasyanov: That's not necessary.

psi29a:
I'm OK with that too. :)

nikolaykasyanov:
@psi29a here we go. Still needs testing on other platforms though.

akortunov:
> Still needs testing on other platforms though.

Which platforms we need to check?

Capostrophic:
Linux and Windows presumably.

components/sdlutil/sdlcursormanager.cpp
```diff
@@ -22,6 +25,13 @@
 USE_GRAPHICSWINDOW()
 #endif
 
+#if OSG_VERSION_GREATER_OR_EQUAL(3, 5, 8) || defined(__APPLE__)
```
psi29a: It seems that only Apple has this issue right, so why are we applying this if OSG >= 3.5.8 OR APPLE?

Shouldn't this be && instead of || ?

akortunov:
Does not compile on Linux for me with OSG 3.6.0.
Compiler complains about undefined "unique_ptr" type.
After adding "#include \<memory\>" to the beginning of file OpenMW compiles and runs fine. Cursor is rendered fine too.

psi29a:
@akortunov feel free (if you can) to push to this branch.

We do allow collaborative commits if it is unobtrusive.

@AnyOldName3 can you test on windows?

AnyOldName3:
On OSG 3.6 or 3.4.1? Both are our fork, and I *think* that we backported DTX decompression before the last OSG 3.4.1 build @ananace put out was made. He'll have to say whether or not that's the case as I can only link against his binaries and I can't look at the source for those or tell which commit they were built from.

Separately to this, we have an environment variable which users can set to declare that OSG was built with the DTX commit despite the version number being too low, meaning that we should probably check that, too.

psi29a:
@AnyOldName3 keep in mind that openmw is sometimes packaged downstream and doesn't have the backported fixes for osg-3.4

AnyOldName3:
I see the following possible situations for a build:

* It uses OSG >= 3.5.8. It has DTX decompression. The environment variable is set or unset, which shouldn't matter. The new code path should be taken. I can test this.
* It uses OSG < 3.5.8. It doesn't have DTX decompression. The environment variable is unset. The old code path should be taken. I might be able to test this depending on how new my OSG-on-steroids 3.4.1 binary is.
* It uses OSG < 3.5.8. It has backported DTX decompression. The environment variable is set. Ideally, the new code path should be taken, but that's not how this is set up right now. I might be able to test this depending on how new my OSG-on-steroids 3.4.1 binary is.
* It uses OSG < 3.5.8. It has backported DTX decompression. The environment variable is not set. The old code path should be taken. I might be able to test this depending on how new my OSG-on-steroids 3.4.1 binary is.

I need @ananace's input to know if I can test the second case or if I can test the third and fourth cases. It's also possible that we only care about one or two of these tests.

My previous post was asking:
* Which of the test cases do we care about.
* If the person who implemented the `#if OSG_VERSION_GREATER_OR_EQUAL(3, 5, 8) || defined(__APPLE__)` check was aware that they should probably also be checking `OPENMW_DECOMPRESS_TEXTURES` at runtime.
* Which of the test cases I'm actually capable of testing.

nikolaykasyanov:
@AnyOldName3 thanks for the feedback, I didn't know about the environment variable, I'll change the code to do a runtime check.

nikolaykasyanov:
@akortunov Capostrophic is right, Windows & Linux, big-endian Linux would be great too.

nikolaykasyanov:
@AnyOldName3 all test cases look valid, how old is your OSG-on-steroids build (or what commit is it)?

nikolaykasyanov:
@AnyOldName3 here we go.

psi29a:
This looks cleaner. :)

zinnschlag:
There was a request to backport this into 0.44.0. What are people's opinions about that? Ready? Advisable?

psi29a:
I'm OK with this going into 0.44

Capostrophic:
Changelog needs to be updated then, with the entry moved to 0.44.0 section.

nikolaykasyanov:
Done!

CHANGELOG.md
```diff
@@ -108,6 +108,7 @@
     Feature #4423: Rebalance soul gem values
     Task #4015: Use AppVeyor build artifact features to make continuous builds available
     Editor: New (and more complete) icon set
+    Bug #4424: [macOS] Cursor is either empty or garbage when compiled against macOS 10.13 SDK
```
Capostrophic: It was and it still is out of order, it'd look nicer if you put it with other bug-type issues before the two bugs that have higher number.

psi29a:
@Capostrophic fixed :)

zinnschlag:
Whelp. I thought I knew how to do multi-commit cherry-picking. Apparently not. My attempt went completely wrong. Don't have the time to dig into it today. Will have another go tomorrow. If someone else would like to try, be my guest.

zinnschlag:
Uhm, that doesn't help. Makes it worse, actually. Because that could possible collide with the cherry pick when and if I get it done.

psi29a:
why is that? 0.44 can be re-tagged, but we branch off from where it is currently tagged
nothing is lost and no need for rebase

zinnschlag:
If you cherry-pick the commit gets a new commit ID. Therefore the same commits now (after cherry-picking) exist twice with different IDs in the repo. I don't know if it causes problems, but it is at least dirty.

zinnschlag:
Hm, apparently that did work. I don't understand why.

psi29a:
Magic... blame Linus. ;)

CHANGELOG.md
nikolaykasyanov: @Capostrophic my bad, now I’m a bit more familiar with how we do changelogs. Thanks!

@psi29a thanks for fixing!

nikolaykasyanov:
Thanks, gentlemen!

Somehow I cannot edit the issue in Redmine.

psi29a:
> Somehow I cannot edit the issue in Redmine.

Good! :)

https://gitlab.com/OpenMW/openmw/issues/4424
^-- our new home

PR #1741, opened 2018-06-09
Do not reset mUpperBodyState for weapon->weapon switch

akortunov:
Fixes regression [#4446](https://bugs.openmw.org/issues/4446).

Currently our weapon->weapon switch has no animations.
With my recent changes, actor stucks in UpperCharState_Nothing in this case.


psi29a:
Should this go into 0.44?

akortunov:
> Should this go into 0.44?

No, it is 0.45 regression.

zinnschlag:
Looks good to me (for 0.45). Only changelog is missing.

akortunov:
> Only changelog is missing.

It is a regression, regressions do not have a target version, so there is no point to add them to changelog.

psi29a:
@akortunov I agree, regression fixes within a milestone doesn't really deserve a changelog entry.

PR #1743, opened 2018-06-09
Add NPC validation to esmstore

akortunov:
Should fix [bug #2772](https://bugs.openmw.org/issues/2772).

The main idea: after we load ESM store, we can iterate over objects in it, and make sure that they are valid.
We can do it once, during game start. Example of output:
```
NPC 'wr_moreri_s1' (Eriel) has nonexistent faction 'wr_morFaction1', ignoring it.
NPC 'arktdunklegestalt02' (Dark Figure) has nonexistent class 'Menschentöter', using 'warrior' class as replacement.
```

We probably have a better place to store validation, so any suggestions would be welcome.

Notes: 
1. I can not use the Store::setUp() since I have to check content of other tables.
2. This approach is different from one, suggested by scrawl:
> A different solution might be replacing all ID fields by a 'smart' data field that holds both an ID and pointer, with the pointer being retrieved from the game's store on first use, and further uses just use the stored pointer.

### DIFF
--- a/apps/openmw/mwworld/esmstore.cpp
+++ b/apps/openmw/mwworld/esmstore.cpp
@@ -120,7 +120,7 @@ void ESMStore::load(ESM::ESMReader &esm, Loading::Listener* listener)
     }
 }
 
-void ESMStore::setUp()
+void ESMStore::setUp(bool validateRecords)
 {
     mIds.clear();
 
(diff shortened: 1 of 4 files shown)


apps/openmw/mwworld/esmstore.cpp
```diff
@@ -142,6 +142,40 @@ void ESMStore::setUp()
...
+    {
+        ESM::NPC& npc = const_cast<ESM::NPC&>(*it); // FIXME: cast
+
+        const std::string npcFaction = npc.mFaction;
+        if (!npcFaction.empty())
+        {
+            const ESM::Faction *fact = mFactions.search(npcFaction);
+            if (fact == NULL)
+            {
+                std::cout << "NPC " << npc.mName << " has unexisting faction '" << npc.mFaction << "', ignore faction." << std::endl;
```
Capostrophic: s/unexisting/nonexistent/g

ignoring it

Might be useful to use the ID instead of/with the name.

akortunov: Done.

akortunov:
I managed to get rid of cast - now I erase old entries from store and put fixed ones here.

akortunov:
This PR should be enough to fix issue with with missing class or faction.

psi29a:
I approve of this, because I wrote up the bug, but would still like @zinnschlag thoughts.

This would fix a whole class of issues regarding TC mods, removing the need for a openmwaddon fix.

apps/openmw/mwworld/esmstore.cpp
```diff
@@ -142,6 +142,55 @@ void ESMStore::setUp()
...
+    for (SharedIterator<ESM::NPC> it = mNpcs.begin(); it != mNpcs.end(); ++it)
+    {
+        ESM::NPC npc = *it;
+        bool changed = false;
+
+        const std::string npcFaction = npc.mFaction;
+        if (!npcFaction.empty())
+        {
+            const ESM::Faction *fact = mFactions.search(npcFaction);
+            if (fact == NULL)
```
zinnschlag: Please avoid NULL. Its pretty useless anyway and with C++11 we have nullptr.

zinnschlag:
Looks alright, though we may end up with a dangling class reference to Farmer now. Wasn't there another issue in regards to implicitly adding a farmer class? Would make sense to combine these fixes.

psi29a:
Not sure, but there is this https://bugs.openmw.org/issues/2246 and it was rejected. It had the same issue, but AoA should load now thanks to this.

akortunov:
I suggest to do not hurry.

I did a some testing. Looks like the "Farmer" is not default class in vanilla.
I created a test plugin with new class, assigned this class to player and NPC.
Then I removed plugin from list of plugins.
For player Morrowind assigned a Warrior class as a **persistent** replacement. I do not know what will happen if the Warrior class does not exist.
For NPC Morrowind just takes previous available class from list (ordered by ID).

My current approach does not work for player since the player does not have the ESM record (savegame just won't load). Should we silently replace player class as in vanilla engine?

zinnschlag:
That sounds reasonable.

akortunov:
For now I just use the first flass from store (usually it is Warrior). In most cases it replicates vanilla behaviour and it is enough to load savegame.

I still do not know what to do with player: if player's class does not exist, savegame won't load.
Replace class silently, as in vanilla or do not load save?

zinnschlag:
I don't see how you could reasonably recover from the player's class being missing. My vote would be for rejecting the file.

akortunov:
> My vote would be for rejecting the file.

So we keep our current behaviour for player for now.
Then this PR is ready for review.

psi29a:
fixed merge conflict

zinnschlag:
Oops. Was just merging that. Guess I have to close this PR manually afterwards.

psi29a:
Sorry!  (This is easier on GL ;) )

psi29a:
@akortunov are you able to close the issue on GL?

akortunov:
> @akortunov are you able to close the issue on GL?

It is closed already: https://gitlab.com/OpenMW/openmw/issues/2772

psi29a:
I think GL did that automagically for us... @zinnschlag is that correct?

psi29a:
OK, cool. :)

PR #1744, opened 2018-06-10
Adding version number of macOS build of OpenMW

Thunderforge:
Fixes [#4047](https://bugs.openmw.org/issues/4047).

Using the "Get Info" command on a Mac, here is the Get Info dialog before:

openmw mac get info no version

And here is the Get Info dialog after:

openmw mac get info fixed version


zinnschlag:
For future reference: Please add an entry to the changelog when you fix an issue.

PR #1746, opened 2018-06-10
Make WakeUpPC interrupt waiting if it was supposed to be (bug #3629)

Capostrophic:
If there was an interrupt waiting hour set during startWaiting(), WakeUpPC will trigger random creature spawning even if the necessary hour hasn't come yet. Thus both an assassin and a creature might spawn if Tribunal script interferes into player's resting, it's pretty rare, but possible now.

Time advancer is already stopped in stopWaiting(), triggered in the event too.


Capostrophic:
One of AppVeyor builds failed due to an internal server error.

PR #1747, opened 2018-06-10
[Don't merge yet] Replace 'reflect actors' with a broader range of reflection detail settings

wareya:
I heard there were objections to this, but I think that the performance gain from not reflecting animated models, or not drawing statics, is simply too big to overlook, especially considering that vanilla Morrowind doesn't do true reflection at all (it uses some kind of dynamic cubemap) and doesn't have to worry about the performance impact of it. The default setting is equivalent to the former default setting.

In interiors, statics are always reflected.

https://imgur.com/a/4geOVfL

### DIFF
--- a/apps/openmw/mwgui/settingswindow.cpp
+++ b/apps/openmw/mwgui/settingswindow.cpp
@@ -181,6 +181,7 @@ namespace MWGui
         getWidget(mKeyboardSwitch, "KeyboardButton");
         getWidget(mControllerSwitch, "ControllerButton");
         getWidget(mWaterTextureSize, "WaterTextureSize");
+        getWidget(mWaterReflectionDetail, "WaterReflectionDetail");
 
 #ifndef WIN32
         // hide gamma controls since it currently does not work under Linux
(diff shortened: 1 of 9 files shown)


wareya:
Oops, this breaks statics rendering in the topdown map.

wareya:
Fixed.

wareya:
This change is much simpler than shadows but it's almost certainly going to conflict with them, so it should go in after shadows do.

akortunov:
"Statics" provides reflection artifacts in Vivec.
screenshot_20180611_083908

wareya:
Looks normal for me.
screenshot209

akortunov:
> Looks normal for me.

It seems the Windows Glow plugin, which I use, replaces some statics to activators, so with "reflect only statics" can be holes in reflections.

wareya:
Yeah, that would turn them into non-statics. Interior reflections might need to be forced to nonactors at minimum instead.

akortunov:
I wish we had static LODs. In this case we could use low-poly models for reflections.

akortunov:
> What about animated world models like the silt striders?
I do not know how to filter them during culling.

akortunov:
> That reminds me, I don't know what setting in this PR will cause interactable objects (loose loot, loose equipment, etc) to become visible in reflections.

"Small feature culling" will handle them anyway.

wareya:
screenshot265

akortunov:
So actually "Statics" renders all not-animated objects.
I suggest to leave it only for not-animated statics and activators.
All items should use a "Model" mask in this case.

Note: for now we consider all activators as animated. Not sure if we can do anything here.

akortunov:
After a testing a bit I found that we should render doors under "Statics" section too.
So safe reflection sets are:
1. Terrain and sky.
2. + statics, activators and doors.
3. + items.
4. + actors.

wareya:
Looks good.

akortunov:
I wonder if there is a way to limit distance of water reflection and refraction and if it will affect performance.

wareya:
You could give the reflection a different far plane, or give it a different set of cutoff distances for actors etc. Good ideas for hidden/advanced (not in ingame UI) settings.

akortunov:
Any progress with this PR?

wareya:
It's "complete" as-is, aside from uncertainty about whether its specific set of settings makes sense (i.e. reflecting statics but not activators or doors).

wareya:
This or akortunov's version of it (https://github.com/OpenMW/openmw/pull/2095) can be revived now that shadows are merged.

wareya:
Replaced by #2095

PR #1748, opened 2018-06-11
Do not show any book text after last <BR> tag

akortunov:
Fixes last case from [bug #4215](https://bugs.openmw.org/issues/4215).
Vanilla game does not show any text after last "\<BR\>" tag (probably a bug in the trailing newlines removal).
OpenMW renders this text, so there are trailing symbols in some books in the German version of  game.

If we do not want to replicate this quirk, we can just close this bug and describe that our current behaviour is intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
     Bug #3997: Almalexia doesn't pace
     Bug #4036: Weird behaviour of AI packages if package target has non-unique ID
     Bug #4047: OpenMW not reporting its version number in MacOS; OpenMW-CS not doing it fully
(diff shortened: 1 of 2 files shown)


psi29a:
What exactly are the symbols? Are they QA marks?

akortunov:
> What exactly are the symbols? Are they QA marks?

I do not know what the "QA mark" is. A quotation mark?
In German version - yes, but vanilla engine ignores ANY text after last BR tag.
If there are no tags, the book is rendered as empty one.

psi29a:
Quality Assurance, like a marker to say this text has been approved.

psi29a:
OK, so this is a problem only seen in the German translation. Merging.

PR #1751, opened 2018-06-13
Fix #4452

wareya:
I had to work around a couple different problems (underwater terrain was broken because of screen fog, statics nearly parallel to the terrain had halos - both worked around) so this probably needs to be tested on computers other than my own. Also, terrain rendering is broken when shaders fail to compile with or without this patch (without the patch it looks like https://i.imgur.com/iPMGzA2.jpg, with the patch it's pure black), so I didn't bother to update the "shaders fail to compile" terrain material codepath.

See http://bugs.openmw.org/issues/4452 and https://imgur.com/a/nzGazLY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
     Bug #4429: [Windows] Error on build INSTALL.vcxproj project (debug) with cmake 3.7.2
     Bug #4432: Guards behaviour is incorrect if they do not have AI packages
     Bug #4433: Guard behaviour is incorrect with Alarm = 0
+    Bug #4452: Default terrain texture bleeds through texture transitions
     Feature #4345: Add equivalents for the command line commands to Launcher
     Feature #4444: Per-group KF-animation files support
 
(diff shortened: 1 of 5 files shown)


psi29a:
@lysol90 you're up I think. :)

unelsson:
I tested this without mods, spent some time flying around Morrowind and Solstheim, and also visited Mournhold although likely not affected by this. Looks good to me, but I also can't tell if mods are affected.

Capostrophic:
They work just fine.

fyi psi29a diffuse map is just the color map, if you see that terrain is not black then it's there

zinnschlag:
This looks ready for a merge.

psi29a:
Fixed merge conflict, waiting on CI to catch up for a final rebuild.

Any last comments?

@wareya you ready to take off the WIP in the title?

zinnschlag:
@wareya It seems you are not registered on GitLab yet. Could you please fix that so we can assign the issue to you on our new tracker?

psi29a:
LGTM, merging. Thanks!

PR #1753, opened 2018-06-13
Ignore broken items when search for replacement

akortunov:
Mostly fixes [bug #4453](https://bugs.openmw.org/issues/4453).

The only issue left is to update button tooltips on quick keys menu opening, but I do not how to implement it yet.


zinnschlag:
You are implementing (almost) the same algorithm twice. Wouldn't it make sense to unify these two code segments and then make a new function in the ContainerStore class from them?

akortunov:
> Wouldn't it make sense to unify these two code segments and then make a new function in the ContainerStore class from them?

Done.

psi29a:
https://github.com/OpenMW/openmw/blob/master/CHANGELOG.md  <-- I see 4454 (AI door thingy) bu tno 4453.

akortunov:
https://github.com/OpenMW/openmw/pull/1755

PR #1756, opened 2018-06-13
Make wandering actors not play sneaking idle animations (bug #4284)

Capostrophic:
If an actor has Wander AI package, they'll use standard idle animations instead of sneaking animations when they are in sneaking stance. Works fine in my testing: after ForceSneak Eldafire used standard idle animations while she was standing in place and used actual sneaking animations while moving.


akortunov:
error: ‘class MWMechanics::AiPackage’ has no member named ‘mWander’

Capostrophic:
I was aware of that, I was just trying out a way to do the thing I want to do. Removed the test commit.
I need some way to know if the actor is going to play additional idles (idle2-idle9) with a non-zero chance. So far I know that the info I seek is in AiWanderStorage mIdle vector, but I'm not aware of a method to recover it.

Capostrophic:
Ready for review.

PR #1757, opened 2018-06-14
Handle exception when try to load non-music file

akortunov:
This PR is aimed to fix [bug #4416](https://bugs.openmw.org/issues/4416).

Currently he have unhandled exception, which causes the skip of frame update.
Should mostly work now, the only issue left is whan player does not have valid tracks in playlist, only invalid ones.

@kcat, can you take a look at this PR?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
     Bug #4327: Missing animations during spell/weapon stance switching
     Bug #4368: Settings window ok button doesn't have key focus by default
     Bug #4393: NPCs walk back to where they were after using ResetActors
+    Bug #4416: Handle exception if we try to play non-music file
     Bug #4419: MRK NiStringExtraData is handled incorrectly
     Bug #4426: RotateWorld behavior is incorrect
     Bug #4429: [Windows] Error on build INSTALL.vcxproj project (debug) with cmake 3.7.2
--- a/apps/openmw/mwsound/ffmpeg_decoder.cpp
+++ b/apps/openmw/mwsound/ffmpeg_decoder.cpp
@@ -251,21 +251,24 @@ void FFmpeg_Decoder::open(const std::string &fname)
         if(mOutputChannelLayout == 0)
             mOutputChannelLayout = av_get_default_channel_layout((*mStream)->codec->channels);
     }
-    catch(...) {
+    catch(...)
+    {
         if(mStream)
             avcodec_close((*mStream)->codec);
         mStream = NULL;
 
-        if (mFormatCtx->pb->buffer != NULL)
+        if (mFormatCtx != NULL)
         {
-            av_free(mFormatCtx->pb->buffer);
-            mFormatCtx->pb->buffer = NULL;
-        }
-        av_free(mFormatCtx->pb);
-        mFormatCtx->pb = NULL;
+            if (mFormatCtx->pb->buffer != NULL)
+            {
+                av_free(mFormatCtx->pb->buffer);
+                mFormatCtx->pb->buffer = NULL;
+            }
+            av_free(mFormatCtx->pb);
+            mFormatCtx->pb = NULL;
 
-        avformat_close_input(&mFormatCtx);
-        throw;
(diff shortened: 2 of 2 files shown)


kcat:
I'm more a proponent of a blacklist, honestly (extensions we know we don't want), since I'd prefer to allow new formats to be implicitly supported as they become available in FFmpeg. In some cases, it may even be desirable to allow video files, just reading its audio track. At the very least we'd want to also support `ogg` and `flac`, though.

But more than that, I'd rather find out what's causing the crash upon trying to load a file it can't play. We ostensibly already handle the case where a file can't be played, so it shouldn't matter if a non-audio file is found and this is just hiding a deeper problem that can bite us elsewhere. FFmpeg tries to load a file, finds no audio stream in the file, nothing plays -- then next time, it sees nothing is playing and goes to the next file/track. If it's a problem with FFmpeg, we may want to find out what and see if upgrading or downgrading what we use on macOS helps.

akortunov:
I just managed to reproduce the crash on Linux.
Debugger complains about [this](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwsound/ffmpeg_decoder.cpp#L259) line - mFormatCtx and other variables a just empty.

kcat:
Yes, it looks like if it gets thrown [here](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwsound/ffmpeg_decoder.cpp#L208), `mFormatCtx` will be `NULL` and crash in the exception handler.

akortunov:
I tried to add a check for null, now it gives the "Error in frame: Failed to allocate input stream" and tries to play the same file again and again.
Probably we should catch exception and remove file from playlist instead of skipping the whole frame update.

akortunov:
@kcat, take a look at new update.
Can we make open() to return bool (success/fail)? In this case we can drop faulty file from playlist.

kcat:
Having the `open` method return a bool for success/fail would probably work. You'd have to go through the code and make sure all uses of the call are handled, though (being aware it's a virtual function, so may be called through the base class too).

akortunov:
Manipulations with playlists are out if my skills yet (I do not know OpenMW sound system well), and out of original report.

Mingun:
> I'm more a proponent of a blacklist, honestly (extensions we know we don't want), since I'd prefer to allow new formats to be implicitly supported as they become available in FFmpeg.

Perhaps it is simple to use an option with the list of all supported formats? With such option it will be possible even to organize a priority of loading of files -- at first we load those expansions which are specified in the list earlier.

zinnschlag:
A list of supported formats is exactly what we want to avoid (see discussion above).

Not sure about this one. Taking the non-working files out of the list, seems a pretty central part of this issue. I don't know how it is handled either. Maybe @kcat has some input about that? If anything else fails, you could just implement a blacklist of broken files.

kcat:
I haven't tested it, but it should work as it is now. If it finds a non-music file in the list, it will try to play it but won't be able to. Then on the next update it should see music isn't playing and go to the next waiting file in the list. Is that not what happens?

akortunov:
> on the next update it should see music isn't playing and go to the next waiting file in the list.

Basically, this PR behaves in this way. The only issue is when all files in playlist are not valid. In this case there will be no sound and spam in console because OpenMW still will use playlist with invalid sound files.

kcat:
I don't think that's needed. There normally isn't any non-music files in the music folders, though even if there are, there would be vastly more audio files than non-audio. The only way to have a playlist folder without audio files is by the user directly messing with/deleting the vanilla files, so they'd already know (though they really shouldn't do that anyway; if you want to silence them, make a mod that replaces them with silence).

akortunov:
> if you want to silence them, make a mod that replaces them with silence).

Or just rename Battle or Explore folder. Or disable music completely in game settings.

psi29a:
So we're OK with this as-is? :)

kcat:
I'm okay with it.

psi29a:
Looks good (and clean) to me too.

PR #1759, opened 2018-06-15
Refactor SDL loading in Launcher to fix macOS errors

Thunderforge:
Fixes both [Bug #2862](https://bugs.openmw.org/issues/2862) and [Bug #3911](https://bugs.openmw.org/issues/3911).

As suspected in the bug reports, SDL and Qt were interfering with each other, which caused issues on macOS. Since we only need SDL for the graphics tab, I limited the scope to that. This prevents issues with things outside of the graphics panel due to the SDL/Qt conflict, and everything works great now.


psi29a:
Great work, thanks! Merging.

nikolaykasyanov:
That's great, thanks!

PR #1760, opened 2018-06-15
Add murder bounty when a player follower commits murder

akortunov:
Should fix [bug #2852](https://bugs.openmw.org/issues/2852).

Created as replacement of #921.

The main idea: if the murderer is the player's follower, assign murder bounty to player in MechanicsManager::actorKilled().

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -760,11 +760,14 @@ namespace MWMechanics
         {
             // The actor was killed by a magic effect. Figure out if the player was responsible for it.
             const ActiveSpells& spells = creatureStats.getActiveSpells();
-            bool killedByPlayer = false;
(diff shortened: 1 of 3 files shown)


apps/openmw/mwmechanics/actors.cpp
```diff
@@ -782,9 +786,14 @@ namespace MWMechanics
...
+                    {
+                        if (caster == player || playerFollowers.find(caster) != playerFollowers.end())
+                        {
+                            killedByPlayer = true;
+                            break;
+                        }
+                    }
                 }
             }
             if (killedByPlayer)
```
akortunov: Not sure if we really need to increase werewolf kills here since werewolves can not cast spells.

Capostrophic: There's the case when the player could have transformed into a werewolf while the magic effect that killed the NPC was still in action, I think.

PR #1761, opened 2018-06-15
Implement unlockable locks with 'lock 0' console command

akortunov:
Should fix [bug #4431](https://bugs.openmw.org/issues/4431).

Note: std::numeric_limits<int>::max() value can depend on target platform, so saves from other platform can cause issues.

### DIFF
--- a/apps/openmw/mwclass/container.cpp
+++ b/apps/openmw/mwclass/container.cpp
@@ -239,7 +239,8 @@ namespace MWClass
         info.caption = ref->mBase->mName;
 
         std::string text;
-        if (ptr.getCellRef().getLockLevel() > 0)
+        int lockLevel = ptr.getCellRef().getLockLevel();
+        if (lockLevel > 0 && lockLevel != ESM::UnbreakableLock)
             text += "\n#{sLockLevel}: " + MWGui::ToolTips::toString(ptr.getCellRef().getLockLevel());
         else if (ptr.getCellRef().getLockLevel() < 0)
             text += "\n#{sUnlocked}";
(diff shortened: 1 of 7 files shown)


apps/openmw/mwclass/container.cpp
```diff
@@ -239,7 +239,8 @@ namespace MWClass
         info.caption = ref->mBase->mName;
 
         std::string text;
-        if (ptr.getCellRef().getLockLevel() > 0)
+        int lockLevel = ptr.getCellRef().getLockLevel();
+        if (lockLevel > 0 && lockLevel != std::numeric_limits<int>::max())
```
psi29a: why not just use -1 ?

akortunov: -1 is a negative level. Locks with negative levels are Unlocked.

zinnschlag:
> std::numeric_limits::max() value can depend on target platform

So does the size of the int type and that doesn't bother us (apparently). Using numeric_limits doesn't make the situation worse.

apps/openmw/mwclass/container.cpp
psi29a: Derp, ok.

psi29a:
Should we define a max that we can use across the board, an impossibly high number in game?

Thunderforge:
> Should we define a max that we can use across the board, an impossibly high number in game?

Isn't that what `std::numeric_limits<int>::max()` is?

psi29a:
@Thunderforge yes and break save games across platforms, by defining it ourselves then we make sure that the value is consistent across all platforms.  std::numeric_limits< type >:: is not portable.

Thunderforge:
Huh, I had no idea that the numeric limits varied by platform.

According to [this](https://stackoverflow.com/a/30689570/531762), you can request things like the maximum unsigned 16-bit number via `std::numeric_limits<uint16_t>::max()`, which will be consistent across platforms because there is only one mathematically correct definition of that maximum. The catch is that it’s optional for implementations to include this, so we *still* may have cross-platform issues.

I’m guessing the best thing would be to just define our own constant that we are sure will be supported on all platforms?

kcat:
Platform compatibility isn't an issue. OpenMW itself, and many support libs, already assume specific sizes (and layouts) for integers. OpenAL, for instance, requires ints to be 2's compliment 32-bit, shorts to be 2's compliment 16-bit, char/bytes to be 8-bit, and floats/doubles to be 32/64-bit IEEE-754 (this is required by the spec). I think OpenGL is the same, actually.

If there's a platform where this isn't true, OpenMW will have many other problems to worry about than the maximum integer value.

psi29a:
Fair enough, I would expect int to be 4-bytes anyway regardless of platforms. Where I saw that `std::numeric_limits<uint16_t>::max()` wasn't consistent was that it was bug. One care less. :)

akortunov:
>  Is there still something missing or did you just forgot to remove the WIP?

Known issues from bug description are still valid.

PR #1762, opened 2018-06-16
[Feedback needed] Quick load inconsistent with original

MocquillonCedric:
I know that this bug need feeback but I tried this one to start with the base code.
The two first commits are enough to fix the bug but other add code consistency and (small) improvements (const correctness, variable construction avoided, ...)

### DIFF
--- a/apps/openmw/mwstate/quicksavemanager.cpp
+++ b/apps/openmw/mwstate/quicksavemanager.cpp
@@ -1,5 +1,31 @@
 #include "quicksavemanager.hpp"
 
+namespace MWState {
+    class NextSlotFinder {
+        std::string mSaveName;
+        unsigned int mMaxSaves;
+        unsigned int mSlotsVisited;
+        const Slot *mOldestSlotVisited;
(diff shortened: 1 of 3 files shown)


psi29a:
Wasn't this discussed in the forums? I thought the consensus was that it was good how it was now? Is there a issue/bug to track this on our tracker?

Capostrophic:
>Is there a issue/bug to track this on our tracker?

[There is](https://bugs.openmw.org/issues/4394), and there was even a perfectly fine solution from akortunov - well, it was rejected.

MocquillonCedric:
Sorry for that, I missed it! I'm ok with the conclusion to do not change the current behavior.

MocquillonCedric:
I found the PR from akortunov (PR #1683), the code is effectively simpler but it does not work on all cases (when the number of quicksave is not set to one (the default), it load the first found, not necesserally the latests).
Nevertheless it does not change the fact that there is no concessus to change the current behavior.

PR #1763, opened 2018-06-16
Forbid actors to use teleporting doors

akortunov:
Fixes [bug #2562](https://bugs.openmw.org/issues/2562).

Currently if NPC activates teleporting door, it can cause crash in both vanilla and OpenMW.
Disable such activations for now.

After 1.0, if we teach AI to move between cells properly (e.g. via doors and teleportation), we will re-enable them.


psi29a:
Thank you for this! LGTM; Merging.

PR #1764, opened 2018-06-16
Ignore lights without CanCarry flags when NPC selects torch

akortunov:
Fixes [bug #4457](https://bugs.openmw.org/issues/4457).

The main idea: check if the light source can be equipped when we select light source, not when we try to equip it.


psi29a:
Thanks.

PR #1765, opened 2018-06-16
Fix arguments parsing for AiWander console command

akortunov:
Fixes [bug #4458](https://bugs.openmw.org/issues/4458).

In upstream code AiWander console command forms idleList of 9 elements (chances for Idle1-9), then drops last element, but treats array as chances for Idle2-9 for some reason.
Vanilla drops chance for Idle1 instead.


psi29a:
Should we be dropping Idle1? Is that not a bug in Morrowind?

Capostrophic:
It's not. Having chance for a default idle is nonsense, and there are only 8 idle chance arguments for AiWander.

psi29a:
OK, thanks.  Merging.

PR #1766, opened 2018-06-16
Started migrating installation guide from the wiki.

rhtucker:
Fixed a bunch of errors that popped up on my local build.


rhtucker:
So I realize some of this should probably be on separate PRs, but is there another way of having separate PRs than maintaining multiple branches?

loriel2:
See for example https://dictionary.cambridge.org/grammar/british-grammar/as-because-or-since - 
"As, because and since are conjunctions. As, because and since all introduce subordinate clauses. They connect the result of something with its reason."

"Since" "= conjunction
"you're probably like me when I started helping and don't know wtf ReST is" = reason
,"it's an incredibly simple language that is easy to read in plain text form that can then be converted automatically into different types of documents like PDFs and html for webpages." = result (?)

The "result" part of the sentence does not seem to me to be a consequence of with the "reason" part.

Another alternative would be to reword the second part, to turn it into a related result - eg "you will be glad to discover that it's an incredibly ...". But even with this change, I think it reads better without the leading "Since"

rhtucker:
Hmm, yes, I see what you mean. Removing "since" would make it two independent clauses though so I just split them up. It was a very long sentence anyway. Glad you're helping with docs!

zinnschlag:
Should we merge that or have it as an ongoing PR that is merged later?

psi29a:
@rhtucker If a PR is ongoing, please set `WIP` in the beginning of your subject so we know not to merge it.

Merging now.

PR #1767, opened 2018-06-18
Allow partial matches in NotCell condition (bug #4459)

Capostrophic:
[Bug report.](https://gitlab.com/OpenMW/openmw/issues/4459)

Extended the condition to return false on partially matching strings. Seems to work just fine with the plugin supplied by rot tor ([only available on redmine](https://bugs.openmw.org/issues/4459)) -- on master the greeting does appear in interiors of Balmora, with the change it does not. It should resolve the reporter's issue too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
     Bug #4454: AI opens doors too slow
     Bug #4457: Item without CanCarry flag prevents shield autoequipping in dark areas
     Bug #4458: AiWander console command handles idle chances incorrectly
+    Bug #4459: NotCell dialogue condition doesn't support partial matches
     Feature #4256: Implement ToggleBorders (TB) console command
     Feature #3276: Editor: Search- Show number of (remaining) search results and indicate a search without any results
     Feature #4222: 360° screenshots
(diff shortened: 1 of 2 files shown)


akortunov:
Are you sure that it just uses string comparison?
For example, we have two exterior cells: "Cell" and "Cell2". In your case "NotCell Cell" is true for both cells. Is this behaviour valid?
Another example: you have two interiors - "Cell, Interior" and "Cell, Interior2". "NotCell 'Cell, Interior'" is true for both cells.

Maybe it checks cell blocks, separated by comma?

Capostrophic:
I do almost exactly the same thing as [GetPCCell](https://github.com/OpenMW/openmw/blob/openmw-44/apps/openmw/mwscript/cellextensions.cpp#L109).
However, OpenMW doesn't check if the cell argument is valid, so it appears that our understanding of this function is incomplete.

zinnschlag:
So that means more research is needed before we can continue here?

Capostrophic:
Just tested. Morrowind works almost exactly the same, but it also checks if the tested cell does exist.
I created a cell named toddtes and went to toddtest. GetPCCell returns 1 for toddtes argument in this case.
Since NotCell condition is supposed to be consistent with GetPCCell behavior, string comparison should work fine. A check if the tested cell actually exists can be implemented later, if that's necessary.

PR #1768, opened 2018-06-18
TTF font scaling

akortunov:
Fixes [bug #3288](https://gitlab.com/OpenMW/openmw/issues/3288), as scrawl suggested (option 1).
You can twik the font size by settings:
```
[GUI]
font size = 16
ttf resolution = 96
```
Also there is no need to edit system-wide openmw_font.xml: you can define your own openmw_font.xml in addition to basic one (an example is [here](https://pastebin.com/Ly91jkrL)).
Just create a Fonts folder near openmw.cfg and place your openmw_font.xml and TTF/OTF files here.

Example if output is [here](https://imgur.com/a/9ojboWg) (before/after, font size = 17, scaling factor = 2, 1920x1080, Pelagiad).

Basically works, but there are issues with large font size, since our UI supposed to work with 16-17px fonts.
Also there could be some refactoring improvements, so any suggestions are welcome.

For testing you can use these fonts: [Fonts.zip](https://github.com/OpenMW/openmw/files/2347975/Fonts.zip) - just place the Fonts folder to your OpenMW configs folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
     Bug #3059: Unable to hit with marksman weapons when too close to an enemy
     Bug #3072: Fatal error on AddItem <item> that has a script containing Equip <item>
     Bug #3249: Fixed revert function not updating views properly
+    Bug #3288: TrueType fonts are handled incorrectly
     Bug #3374: Touch spells not hitting kwama foragers
     Bug #3486: [Mod] NPC Commands does not work
     Bug #3533: GetSpellEffects should detect effects with zero duration
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -516,7 +516,7 @@ void OMW::Engine::prepareEngine (Settings::Manager & settings)
     MWGui::WindowManager* window = new MWGui::WindowManager(mViewer, guiRoot, mResourceSystem.get(), mWorkQueue.get(),
                 mCfgMgr.getLogPath().string() + std::string("/"), myguiResources,
                 mScriptConsoleMode, mTranslationDataStorage, mEncoding, mExportFonts, mFallbackMap,
-                Version::getOpenmwVersionDescription(mResDir.string()));
+                Version::getOpenmwVersionDescription(mResDir.string()), mCfgMgr.getUserConfigPath().string());
     mEnvironment.setWindowManager (window);
 
     // Create sound system
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -290,6 +290,8 @@ namespace MWBase
             /// Warning: do not use MyGUI::InputManager::setKeyFocusWidget directly. Instead use this.
             virtual void setKeyFocusWidget (MyGUI::Widget* widget) = 0;
 
+            virtual void loadUserFonts() = 0;
+
             virtual Loading::Listener* getLoadingScreen() = 0;
 
             /// Should the cursor be visible?
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -219,6 +219,7 @@ namespace MWBase
             virtual const MWWorld::Ptr& getSelectedEnchantItem() const = 0;
             virtual void setSelectedWeapon(const MWWorld::Ptr& item) = 0;
             virtual const MWWorld::Ptr& getSelectedWeapon() const = 0;
+            virtual int getFontHeight() const = 0;
             virtual void unsetSelectedSpell() = 0;
             virtual void unsetSelectedWeapon() = 0;
 
--- a/apps/openmw/mwgui/bookpage.cpp
+++ b/apps/openmw/mwgui/bookpage.cpp
@@ -264,18 +264,29 @@ struct TypesetBookImpl::Typesetter : BookTypesetter
     {
     }
 
-    Style * createStyle (char const * fontName, const Colour& fontColour)
+    Style * createStyle (const std::string& fontName, const Colour& fontColour)
     {
-        if (strcmp(fontName, "") == 0)
-            return createStyle(MyGUI::FontManager::getInstance().getDefaultFont().c_str(), fontColour);
+        const std::string templateName = "Journalbook ";
+        std::string bookFont;
+        if (fontName.empty())
+        {
+            bookFont = MyGUI::FontManager::getInstance().getDefaultFont();
+            bookFont = templateName + bookFont;
+            return createStyle(bookFont, fontColour);
+        }
+
+        if (fontName.compare(0, templateName.size(), templateName) == 0)
+            bookFont = fontName;
+        else
+            bookFont = templateName + bookFont;
 
         for (Styles::iterator i = mBook->mStyles.begin (); i != mBook->mStyles.end (); ++i)
-            if (i->match (fontName, fontColour, fontColour, fontColour, 0))
+            if (i->match (bookFont.c_str(), fontColour, fontColour, fontColour, 0))
                 return &*i;
 
-        MyGUI::IFont* font = MyGUI::FontManager::getInstance().getByName(fontName);
+        MyGUI::IFont* font = MyGUI::FontManager::getInstance().getByName(bookFont);
         if (!font)
-            throw std::runtime_error(std::string("can't find font ") + fontName);
+            throw std::runtime_error(std::string("can't find font ") + bookFont);
 
         StyleImpl & style = *mBook->mStyles.insert (mBook->mStyles.end (), StyleImpl ());
         style.mFont = font;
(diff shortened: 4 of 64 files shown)


psi29a:
Well... sad news is, scrawl was also 'admin' on mygui gh probject. By leaving, he is no longer admin so we can't really make a release. Does this mean that we have to fork mygui as well and self-host?

psi29a:
Can you open a PR for https://github.com/MyGUI/mygui please?

akortunov:
> Can you open a PR for https://github.com/MyGUI/mygui please?

I suggest to solve issue with PageDisplay first.

Altren:
Instead of loading font as a complete resource you should create MyGU::ResourceTrueTypeFont in runtime, using your scale during generation. Here is example code to create it from scratch (from FontEditor):
`		std::string resourceCategory = MyGUI::ResourceManager::getInstance().getCategoryName();
		MyGUI::ResourceTrueTypeFont* font = MyGUI::FactoryManager::getInstance().createObject<MyGUI::ResourceTrueTypeFont>(resourceCategory);

		font->setResourceName(fontName);

		font->setSource(_data->getPropertyValue("Source"));
		font->setSize(_data->getPropertyValue<float>("Size"));
		font->setResolution(_data->getPropertyValue<int>("Resolution"));
		font->setHinting(_data->getPropertyValue("Hinting"));
		font->setAntialias(_data->getPropertyValue<bool>("Antialias"));
		font->setTabWidth(_data->getPropertyValue<float>("TabWidth"));
		font->setOffsetHeight(_data->getPropertyValue<int>("OffsetHeight"));
		font->setSubstituteCode(_data->getPropertyValue<int>("SubstituteCode"));
		font->setDistance(_data->getPropertyValue<int>("Distance"));

		std::string ranges = _data->getPropertyValue("FontCodeRanges");
		std::vector<std::string> values = MyGUI::utility::split(ranges, "|");
		for (size_t index = 0; index < values.size(); index ++)
		{
			MyGUI::IntSize size = MyGUI::IntSize::parse(values[index]);
			font->addCodePointRange(size.width, size.height); // о да
		}

		font->initialise();

		MyGUI::ResourceManager::getInstance().addResource(font);`

Altren:
Your current patch is not close to something that can be accepted, but I like the idea of scalable font resource.

psi29a:
@Altren thanks for taking the time to comment. :)

akortunov:
> @akortunov is this an acceptable solution?

I do not know yet. Currently we do not parse XML at all - we allow MyGUI to do it.
Also usually OpenMW user does not have TTF fonts at all - OpenMW uses bitmap fonts by default.
What can we do here? Take all fonts from resource manager, dynamic_cast them to TTF font, then modify fields according to our UI scaling?

psi29a:
@akortunov send a PR to MyGUi, @Altren is active there so there is collaboration to improving your patch for up-streaming. He is interested. :)

akortunov:
I spent a lot of time for testing.
Unfortunately, workaround, suggested by scrawl, does not work well for bitmap fonts.
I do not know if we can do anything from OpenMW side to emulate TTF font scaling.

A root of issue is how MyGUI handles TTF fonts: IFont interface does not allow to to get glyphs of different size from single font, even if this font is based on vector graphics and can be scaled without quality loss.
I do not familiar enough with MyGUI to implement such thing. Sorry.

psi29a:
@Altren could this be a feature request? Seems we're out of our depth in trying to implement something like that necessary in MyGUI.

akortunov:
> This is important, but I don't see MyGUI resolving its related issue soon. What are you going to do?

Honestly, I have no idea. Even if the issue with font height will be fixed on MyGUI side, this PR more likely will not work with old MyGUI versions.

akortunov:
Figured out that we can set font height manually from code:
```
editText->setFontHeight(16);
```

I wonder if we can override a MyGUI factory to do it automatically every time we create an EditBox instance (or instance of derived class).

akortunov:
I figured out how to avoid to patch MyGUI. Unfortunately, books are broken for now.

wareya:
What kind of broken? Not rendering, or still blurry?

akortunov:
> What kind of broken? Not rendering, or still blurry?

Still blurry and big, if you increase the font Resolution parameter..

akortunov:
Book scaling is mostly fixed now, but there is an another issue - book formatter can split text line to two pages.

akortunov:
Should work now, but configuration is not handly - a player still needs to edit XML files to change font size or resolution.

akortunov:
Ok, I added settings to configure font size and resolution.
The only known issue - journal index (at least, russian one) is broken with font size > 17.

akortunov:
@zinnschlag, @psi29a, @Capostrophic, can you take a look on this PR?

Capostrophic:
I'm on it.

Capostrophic:
When the resolution is a ridiculous value (like 1), MyGUI fails to set the font resolution to such value and OpenMW doesn't launch. You may want to put sanity checks in place like with font size, with possibly changing the resolution back to a reasonable value in the config itself. Though somehow the resolution can be 0 or negative in which case MyGUI will fallback to something? Still, in this case seems that it's better to use (96 * scaling factor) as a fallback value.

Currently books use the same resolution as everything else, even though they are scaled separately, so with a very high scaling factor it's noticeable how the font becomes too sharp, and it feels like the book text is *a bit* too blurry on scaling factor 1? (it looks ideal close to scaling factor 2, but I guess it depends on the ratio which was applied to the books to scale them to player's resolution). This could be a limitation, not sure if it's possible to fix.

akortunov:
> Currently books use the same resolution as everything else, even though they are scaled separately, so with a very high scaling factor it's noticeable how the font becomes too sharp, and it feels like the book text is a bit too blurry on scaling factor 1? (it looks ideal close to scaling factor 2, but I guess it depends on the ratio which was applied to the books to scale them to player's resolution). This could be a limitation, not sure if it's possible to fix.

Book and journal are completely custom widgets, and I suppose that their actual scaling factor depends on screen resolution, and their actual text size too.
Since MyGUI can not handle TTF fonts properly, in this case we need do setup separate fonts for book and journal.
Now we have this limitation too and I see no way to workaround it.

akortunov:
Added a separate font scaling for books and journal depending on resolution.
When player changes resoultion in-game, OpenMW reloads TrueType fonts.

akortunov:
The only my real complaint: player still need to modify openmw_font.xml to setup fonts after every update.

akortunov:
Added support for loading fonts from configs folder.

psi29a:
Ready for review/merge or still wip?

akortunov:
> Ready for review/merge or still wip?

Ready for review. Also it would be nice to test it on Windows and on 4K-display.

akortunov:
It seems we need to shut up the C4643 warning for MSVC.

psi29a:
@AnyOldName3 do you have a 4K display?

akortunov:
Also probably we can ping @Thunderforge here (IIRC, he uses MacOS).

AnyOldName3:
I'm on my third day of using a 1440p monitor, if that helps. It's set to 100% display scaling, though, as the pixel size is almost the same as a standard 23" 1080p monitor. What am I supposed to be doing, and is 1440p enough to do it with?

Capostrophic:
>is 1440p enough to do it with?

Yes.

>What am I supposed to be doing?

AFAIK you need to install a TTF font like Pelagiad (preferably by dropping its openmw_font.xml to fonts directory in the openmw local directory (in my games/openmw)) and check how well the interface and journal and books in particular are scaled.

akortunov:
> AFAIK you need to install a TTF font like Pelagiad (preferably by dropping its openmw_font.xml to fonts directory in the openmw local directory (in my games/openmw)) and check how well the interface and journal and books in particular are scaled.

I attached the archive with configured TTF fonts to the PR description. Just place the Fonts folder from this archive to openmw local config directory.
I can not tell if this feature works on Windows for now, so we need testing.

Thunderforge:
@akortunov I do have a Mac. How do I test this? Just checkout the branch and load the fonts in the zip file in my OpenMW config folder?

akortunov:
> @akortunov I do have a Mac. How do I test this? Just checkout the branch and load the fonts in the zip file in my OpenMW config folder?

Basically, yes:
1. Checkout and build this branch.
2. Copy the Fonts folder (with all its content) to your OpenMW configs folder.
3. Since Pelagiad glyphs visually are quite small, you may have increase "font size" option value to 17 or 18.
Also you can change the "ttf resolution" value depending on your display DPI.

akortunov:
Any progress with testing?

Thunderforge:
@akortunov Sorry that I haven't had the chance to test this. I've had some Real Life stuff come up. Is testing on a Mac the only blocker? I hope to be able to test it this week, but if someone else with a Mac wants to test, they can go ahead.

akortunov:
> Is testing on a Mac the only blocker?

No, this PR still was not tested on Windows too.

akortunov:
I asked for testing on Windows and MacOS two weeks ago and still got no response.
So I suggest to assume it working on these platforms and allow users to fill bugreports if it is not true.
@zinnschlag, what do you think?

akortunov:
I managed to test this PR on another PC with Windows 10. Works fine for me.

PR #1769, opened 2018-06-18
[Review phase] Terrain texture selection

unelsson:
Terrain texture selection, based on https://github.com/OpenMW/openmw/pull/1414

Feedback and review is greatly appreciated. Plutonicoverkill's code had some reviews, but I don't clearly understand what are the reasonings behind those comments.


zinnschlag:
And here we reach the point where not having scrawl around really hurts, because I am not familiar with this stuff either.

Anyway, I did a test run and ended up with a segfault on attempting to select terrain: https://gist.github.com/zinnschlag/0916aa134ae01a2484ed24f7f5faef9d

psi29a:
Get in the habit of using WIP in stead of Don't Merge :)

psi29a:
I know that @kcat has some experience with terrain... perhaps he can help review?

unelsson:
I haven't got any segfaults, was there anything specific you were doing, or just clicking primary select? I don't really know how to read that gistfile, but looks like activate-function doing something with Qt.

I really don't know... Version mismatch? I'm using Qt version 5.11.0

I'll try to click everywhere and see if I can get a segfault somehow.

zinnschlag:
Just random clicking. Still on Qt 4, but from the stack backtrace it is pretty clear that this is not a Qt problem.

zinnschlag:
> #0  0x00000000008a7451 in ESM::Land::loadData (this=this@entry=0x7fffffffb1c0, flags=flags@entry=2, 
    target=target@entry=0x0) at /home/marc/OpenMW/openmw/components/esm/loadland.cpp:220
> #1  0x00000000008a7c62 in ESM::Land::getLandData (this=0x7fffffffb1c0, flags=flags@entry=2)
    at /home/marc/OpenMW/openmw/components/esm/loadland.cpp:349

This looks highly suspicious. Just saying.

kcat:
> I know that @kcat has some experience with terrain... perhaps he can help review?

I'm not particularly familiar with terrain-related stuff. Especially on the editor side of things.

psi29a:
@zinnschlag what file(s) are you loading to trigger this?  That stacktrace looks like something in the loadland code.

zinnschlag:
Morrowind.esm. I doubt that it is the load code itself. Notice the this value? Very unusual. Could be a multi-threading problem or maybe an object life-time problem. Only guessing here.

psi29a:
Qt4 uses threaded renderer whilet Qt5 uses single-threaded renderer... not sure if that makes a difference?

zinnschlag:
Not really. I was speaking about threading in our terrain system.

zinnschlag:
Apparently nobody (including me) noticed the elephant in the room: Why is it trying to load terrain when I select a piece of terrain? Obviously the terrain was already loaded. Otherwise it could not have been rendered which means I couldn't have clicked on it to select it.

Aesylwinn:
Could be your click was interpreted as a being in one of the unloaded cells. Maybe try moving away from loaded cells and try reproducing it. I think the worldpos returned defaults to a set distance when nothing gets clicked on. (Unloaded cells just aren't rendered iirc).

zinnschlag:
Lol. I compiled your code again and then noticed the warning:

> /home/marc/OpenMW/openmw/apps/opencs/view/render/terrainselection.cpp:17:67: warning: a temporary bound to ‘CSVRender::TerrainSelection::mCoords’ only persists until the constructor exits [-Wextra]
     :mCoords {coords}, mEsmLand {esmLand}, mParentNode {parentNode}

So, yeah. One of my guesses turned out to be right. Object life-time issue.

unelsson:
I'm trying to reproduce or understand the issue, but I don't get any errors, warnings or crashes, and furthermore, I'm clueless why mCoords would persist only until the constructor exits, as to my understanding it's defined in terrainselection.hpp correctly. If this is clear, I'd be really grateful if someone could explain this to me.

I've stopped development until this is resolved. I don't know how to approach this problem, as Zini's comments are only info I have of this, and I can't understand a problem in the code. There are some other glitches too, but crashing is so fundamental that it's probably better fixed first. There's a noticeable offset from selection grid to actual world texture grid, and cell borders don't operate/render correctly. Drag-select is also still missing.

psi29a:
What platform are you guys on? Windows, Linux or MacOS?

unelsson:
Linux *** 4.9.107-2-MANJARO #1 SMP PREEMPT x86_64 GNU/Linux

zinnschlag:
That you don't get the crash isn't surprising, because undefined behaviour. However that you don't get a warning message is surprising.

Anyway, an easy fix would be to change mCoors from const CSMWorld::CellCoordinates& to simply CellCoordinates. Reference as member variables are rather strange to begin with. Not outright wrong, but most of the time a bad idea. Still trying to track down the source of the problem would be a good idea.

zinnschlag:
> I'm clueless why mCoords would persist only until the constructor exits

Just to make this clear: It is not mCoors (because mCoors is a reference). The compiler is complaining about a temporary that is bound to mCoors.

unelsson:
To be sure, I compiled it again from the start, and searched the log for that warning, and keywords (terrainselect.cpp) etc. but found no warnings related to this. I don't really understand when using references or pointers is good/bad, but I didn't change anything that didn't work.

Thanks for clarification anyway, it gives me at least one place to look.

Aesylwinn:
Compiler bug: https://stackoverflow.com/questions/10509603/why-cant-i-initialize-a-reference-in-an-initializer-list-with-uniform-initializ

I'm guessing Zini is using gcc 4. Use regular initialization to work around it.

psi29a:
@zinnschlag if you're toolchain is buggy, time to upgrade ;)

zinnschlag:
4.8.4. And not upgrading yet (too much stuff that would break and absolutely no time to deal with that).

unelsson:
The code is not written by me, and I'm still fairly beginner in C++, so I'm going to milk some knowledge out of you to fix this, and learn.

So, am I understanding this correctly? Because of C++ compiler bug/feature, there's some undefined behavior occurring from list-initialization in format: variable {parameter}

and that would be fixed by using "default initialization" in format: variable (parameter)

The former should work in C++11, but not necessarily in C++ versions before that? Does this mean that list-initialization cannot be used at all in OpenMW?

In addition, I don't see any advantages for using initializer-list with only one parameter. Is this true?

And fundamentally, I'm understanding that initializer list (member initializer list) is a different thing than list-initialization? Former is the thing happening after : in constructor, and latter is something like variable {}.

To the problem at hand, the problem seems to be affecting mCoords, as it's initialized with a parameter that is a reference. In this case, CSMWorld::CellCoordinates& coords instead of CSMWorld::CellCoordinates coords? I also don't see any advantages in using a reference here.

The full initialization list where the problem occurs also initializes variables mEsmLand {esmLand} and mParentNode {parentNode}, of which mEsmLand is a reference. Shouldn't that be changed also?

Do I need to change both - the list-initialization to default initialization, and references to default variables?

zinnschlag:
> So, am I understanding this correctly? Because of C++ compiler bug/feature, there's some undefined behavior occurring from list-initialization in format: variable {parameter}

> and that would be fixed by using "default initialization" in format: variable (parameter)?

I am not entirely sure if that is what is happening here (since the error message is noticeable different from what I am getting). But it is worth a try.

> The former should work in C++11, but not necessarily in C++ versions before that

The former is a C++11 feature. It did not exist before C++11.

> In addition, I don't see any advantages for using initializer-list with only one parameter. Is this true?

The initiliser-list is nearly always preferable over initilaising in the function body. Its cleaner, for some types it avoids redundant initialisation and for other types it is mandatory.

> To the problem at hand, the problem seems to be affecting mCoords, as it's initialized with a parameter that is a reference. In this case, CSMWorld::CellCoordinates& coords instead of CSMWorld::CellCoordinates coords? I also don't see any advantages in using a reference here.

The problem is not the parameter. The problem is the variable itself. mCoors should not be a reference.

Let me reiterate that: Having a reference as a class member variable is usually a bad idea. It is not outright wrong (there are certainly situation where I have used them in the past), but there is typically little benefit to it and a couple of potential disadvantages (objects of the class become un-assignable for example). I would also argue that it is somewhat counter-idiomatic.

> The full initialization list where the problem occurs also initializes variables mEsmLand {esmLand} and mParentNode {parentNode}, of which mEsmLand is a reference. Shouldn't that be changed also?

You can not simply remove the reference, because that would make a copy and that would be wrong in this case. However using a pointer instead of a reference would express the intend more clearly.

C++ does not have the concept of value and reference types (as in languages like C#). But CellCoordinates comes pretty close to a value type, while ESM::Land (in this context) is pretty close to a reference type.

Aesylwinn:
Initializer list is an ambiguous term. https://en.cppreference.com/w/cpp/utility/initializer_list

unelsson:
I ran into error message "no match for 'operator*" when backtracking and changing the reference of mCoords to a pointer, in mcoordinates in cell.cpp line 147. ( https://github.com/unelsson/openmw/blob/5419f4427b66302c2a3f37be751981804fd33b77/apps/opencs/view/render/cell.cpp#L147 ). There's still a reference as a member variable though.

Anyway, changed the {} to () and tested, on my computer it works without crashing as it did before.

unelsson:
While digging into what's a pointer and what's a reference and what to do, I encountered another thing I didn't understand: What would be the point in auto and static_cast at [terrainselection.ccp lines 59-60 & 67-68](https://github.com/unelsson/openmw/blob/5419f4427b66302c2a3f37be751981804fd33b77/apps/opencs/view/render/terrainselection.cpp#L59-L60)?

```
    const auto x = static_cast<int>(std::floor(toCellCoords(worldPos).first));
    const auto y = static_cast<int>(std::floor(toCellCoords(worldPos).second));
```

Couldn't that be as well just normal conversion from double to int?

```
    const int x = (int)(toCellCoords(worldPos).first + 0.5);
    const int y = (int)(toCellCoords(worldPos).second + 0.5);
```

kcat:
> What would be the point in auto and static_cast

`static_cast` is the proper C++ way to cast things. A C-style cast, like `(int)(...)`, is a lot more forgiving in the casts it can do and so may end up with unexpected results. A C-style cast is basically forces the compiler to accept what you tell it; it disables a lot of potential checks and warnings, and will only fail if the cast is literally impossible (e.g. casting an integer to a structure). `static_cast` will more readily fail for problematic casts.

As for `auto`, a good general coding principle to follow is to avoid repeating yourself. If you've already specified the result as in `int` (via the cast), it's unnecessary to also declare the variable as an `int` since it's supposed to be the same. In this way you avoid a potential issue in the future, such as someone just changing the variable's type but neglecting the cast, or changing the cast's type but neglecting the variable. With `auto`, it ensure the variable will always have the same type as the cast.

unelsson:
I guess passing CellCoordinates as const reference might have been nice, but I didn't know how to do it, so I just reverted back to passing it as value in the first place. When testing, I don't see anything new, but how about Zini, does this fix the crash?

Aesylwinn:
You didn't have to change the function params from const ref to copy. That part was correct. Zini only wanted you to change the class member variable.

unelsson:
Thanks.

Changing only the member variable, but not the function parameters, resulted in some random compile errors that I tried to fix for a good time without any luck. I can try again, but probably need to lose my sanity first.

unelsson:
TerrainTextureMode needs to know which button is used in order to choose correct function, e.g. whether it's primary edit or primary select. What would be the preferred way for achieving this?

EditMode class has functions distinguishing between InteractionType:s for the _start of the drag_, but not during the drag. InteractionType is stored in worldspacewidget, in int mDragMode, as private variable.

1) using CSVRender::TerrainTextureMode::drag for all InteractionType's, but using another int mDragMode to store and choose the correct interaction?

2) using CSVRender::TerrainTextureMode::drag for all InteractionType's, getting the mDragMode from worldspacewidget by implementing CSVRender::TerrainTextureMode::getInteractionType-function?

3) Creating separate functions for all drag's, e.g. CSVRender::TerrainTextureMode::primaryEditDrag, CSVRender::TerrainTextureMode::primarySelectDrag, and calling those functions from worldspacewidget.cpp? This requires altering the base EditMode class.

zinnschlag:
The button state needs to be examined at the beginning of the drag. We don't want to allow changing it in the middle of the drag. You are supposed to store any required additional information locally in the edit mode class when the drag begins. So, option 1.

unelsson:
It seems that dragging out of one cell causes a segfault.

unelsson:
Also, something strange going on with terrain texture selection size and actual texture location. In the screenshot, it looks like textures start from the edge in Y direction, but X is cut by... maybe half a texture?
https://i.imgur.com/p61S5K0.png

unelsson:
Getting texture coordinates exactly right is crucial for this feature. 

edit 2: Over here https://github.com/OpenMW/openmw/blob/908af3720f4198b36c1247d1fd0b3c6177bd38ca/components/terrain/material.cpp#L32 , there are some tricks which translate and scale textures, and it should be possible to apply a fix to ltex coordinates here, but this requires some investigating whether this "bug" occurs in Vanilla too, or is it only OpenMW specific.

zinnschlag:
Yeah, again a moment were losing scrawl really hurts. I don't have the slightest idea and can't provide any useful input. Sorry about that.

@AnyOldName3 Didn't you mention that you had worked a bit with terrain? any input about this?

AnyOldName3:
The parts of terrain I've worked on don't really overlap with what's being changed here, but from reading other PRs and bug reports in the past, I do remember that the weird `y` position of things is in vanilla Morrowind, and we're at least attempting to do it the same. I don't recall if we've got it exactly the same as vanilla, but I know that searching through past PRs and bug reports and the forums should answer that kind of question fairly thoroughly.

apps/opencs/view/render/terrainselection.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include <utility>
+#include <vector>
+
+#include <osg/Vec3d>
+#include <osg/ref_ptr>
+#include <osg/PositionAttitudeTransform>
+
+#include <components/esm/loadland.hpp>
+#include "../../model/world/cellcoordinates.hpp"
```
AnyOldName3: I think this should probably be a full path instead of a `..`-based one.

zinnschlag: Nope, that is how we normally do includes.

unelsson:
1/4th of a cell :D.. Not cell, but 1/4th of a landtexture.

I guessed the nudging formula kinda correct to land heights calculations, so this seems to work when selecting landtextures in the middle of cells. Note though that this will still crash when selecting textures crossing cell edges, as the mHeights goes out of boundary. Drawing selection borders needs a big overhaul anyway, so the current code is pretty much just proof of concept.

unelsson:
Behavior with segfaults is strange - dragging over cell edge slowly doesn't trigger a segfault, but dragging normal/fast always causes a segfault. Commenting out the vertices->push_back -lines will get rid of segfaults. Even simplified code like "vertices->push_back(osg::Vec3f(toWorldCoords(x) + nudgeOffset, toWorldCoords(y) - nudgeOffset, 250));" causes a segfault, but only when dragging normal/fast!

https://github.com/unelsson/openmw/blob/010822b959acd43995180ff515d9723d9d413695/apps/opencs/view/render/terraintextureselection.cpp#L86

AnyOldName3:
```C++
const osg::ref_ptr<osg::Vec3Array> vertices {new osg::Vec3Array{}};
```

I'd have done

```C++
const osg::ref_ptr<osg::Vec3Array> vertices(new osg::Vec3Array{});
```

or

```C++
const osg::ref_ptr<osg::Vec3Array> vertices = new osg::Vec3Array{};
```

I think the problem is related to you treating the `ref_ptr` to an array as if it was an array and initialising it weirdly. It probably works sometimes because C++ arrays are really just pointers, but doesn't work in other cases because it's wrong and only working by accident.

wareya:
Yeah, vanilla has a weird offset. I don't know why, I suspect it's due to however they implemented the upscaling-like thing that keeps the painting from being blurry.

unelsson:
Trying different ref_ptr arrangements doesn't fix the segfault, although it also works similarly. There were some earlier comments about hyperthreading, and I'm _guessing_ if this could be something related to ref_ptr:s trying to simulatenously reach to a common osg object. Reasoning is that the problem occurs only when dragging quickly between cells, and therefore there could be two cells using two different terrainselection's both accessing same osg object, or something?

unelsson:
I believe this feature has now been developed to a point, where extra planning is required.

This PR is based on the idea that terrain selection is cell-specific instead of worldwide. This approach faces some programming challenges when working with areas spanning multiple cells. On the other hand, iterating through selected areas in any given cell should be easy. I wonder if this is the correct approach in the first place?

I've encountered a problem when drawing the selection at cell edges, because terrainselection class implemented like this can only know the height data of it's own cell, but not the neighboring cells. Terrain textures at the edge do however extend beyond the cell edge. In order to draw the selection area correctly, one needs to know the height and selection data from neighboring cells. This is a programming challenge, and the solution is not obvious to me. Perhaps moving the drawing code out from terrain selection class, and iterating once through all terrain selections could do the trick, but where should this happen?

Also, how should the selection look like? Lines at the edge, but maybe also some sort of layer effect, texture effect?

These questions don't necessarily need an answer right now, but worth thinking anyway: Which classes need to know about the selection? How should the selection data be passed around?

Aesylwinn:
I think a global terrain editor makes more sense than working with multiple editors that each track their own selection. It's a more natural abstraction that fits with how the user is going to work. It does conflict though with how other editor operations work. Arguably those other commands could also be made to work at the global level. I'm curious as to why they were designed to work at the cell level. Perhaps to handle interior/exterior cells?

Anyway, is it not possible to store which points are selected on a per cell basis? If you are going for operations like flatten/smooth, then you could broadcast whatever information is needed for that operation (like target/average height, center of selection, etc) to all loaded cells.

zinnschlag:
Not sure either why the code originally stored selections on a per-cell basis. If you want to switch it over to a worldspace/scene wide implementation I am all in favour of it.

> Also, how should the selection look like? Lines at the edge, but maybe also some sort of layer effect, texture effect?

I don't have any specific appearance in mind. Use what ever works. Play around with a few variations, if you want to.

> These questions don't necessarily need an answer right now, but worth thinking anyway: Which classes need to know about the selection? How should the selection data be passed around?

Make a list of all functionality that requires access to the selection and then determine where this functionality is located?

Off the top of my head:
- Modification of what area is selected (via mouse click or button menu)
- Applying texture to selected area (via click or drop)
- Setting brush from selection

There shouldn't be much passing around of the selection. It gets set and it gets queried. I would expect a vector with tile locations would do the job.

For copy and paste we don't have to handle just the locations but also the texture at the locations. This may also be relevant for the missing brush mode. That's something to keep in mind.

unelsson:
Merged terrainselection and terraintextureselection into one, and changed the cell-based drawing into pagedworldspacewidget-based (scenewide) drawing. This destroyed most functionality in the process, and required some rewriting, but now it's time to choose the final data structure for selection.

What's planned now:
mTerrainSelection as a member of PagedWorldSpaceWidget

mTerrainSelection holds a get-function that returns the data.

It's time to choose the data format, so perhaps something like A, B or C?

A) std::vector<std::tuple<int, int, int, int>> mSelection; // CellX, CellY, localX, localY

B) std::vector<std::tuple<CSMWorld::CellCoordinates, int, int>> mSelection; // CellCoordinates, localX, localY

C) std::vector<std::pair><CSMWorld::CellCoordinates, std::pair<int, int>> // CellCoordinates, LocalCoordinates

unelsson:
Now, this makes me slightly mad, as it would be much easier to detect cell edges if all selection coordinates were stored in a big map holding everything, like std::pair<int,int>... This on the other hand, requires some inner logic, e.g. texture_selection_coordinate 0,0 would be in cell 0,0, but texture_selection_coordinate 16, 16 would be in cell 1,1...

unelsson:
I've understood that selection feature should support both terrain vertexes (height, color, normals) and texture grid. I've thought of these as two different "modes", therefore resolution of selection would need to jump from low to high and back. Converting the selection from high res (grid of vertexes) to low res (grid of textures) will lose data, no matter the algorithm. It could be possible to solve this by storing only high res selection (vertexes), and do conversion only temporarily when selection is used to modify textures. How does this sound?

unelsson:
The basic framework for terrain texture selection should be done now. It's just the ability to select, and no features that use selection yet. Graphical implementation is just very basic box-like selection, I might improve/change that later. It might be a good time to review this, it has been a good learning process for me, and quite challenging for my programming skills.

AppVeyor builds seems to fail randomly every now and then.

I didn't yet commit the conversion functions to CellCoordinates class, mainly because I haven't yet had much use of that class, and I don't properly understand how it should be done. One idea is below, but I haven't tested it properly. Probably should be landTextureSize(16) instead of landSize(64) in terrainTextureSelectionToCellId calculations too.

```
std::pair<int, int> CSMWorld::CellCoordinates::terrainTextureSelectionToLocalCoordinates (std::pair<int, int> globalPos)
{
    int x (globalPos.first);
    int y (globalPos.second);
    int cellX(0);
    int cellY(0);
    int localX(0);
    int localY(0);

    if (x >= 0)
    {
        cellX = std::floor(x / (landTextureSize - 1));
        localX = x - cellX * (landTextureSize - 1);
    }
    if (y >= 0)
    {
        cellY = std::floor(y / (landTextureSize - 1));
        localY = y - cellY * (landTextureSize - 1);
    }
    if (x < 0)
    {
        cellX = std::trunc(x / (landTextureSize - 1)) - 1;
        localX = x - cellX * (landTextureSize - 1);
    }
    if (y < 0)
    {
        cellY = std::trunc(y / (landTextureSize - 1)) - 1;
        localY = y - cellY * (landTextureSize - 1);
    }

    std::cout << localX << ", " << localY << std::endl;
    return std::make_pair(localX, localY);
}

std::string CSMWorld::CellCoordinates::terrainTextureSelectionToCellId (int x, int y)
{
    int cellX(0);
    int cellY(0);

    if (x >= 0)
    {
        cellX = std::floor(x / (landSize - 1));
    }
    if (y >= 0)
    {
        cellY = std::floor(y / (landSize - 1));
    }
    if (x < 0)
    {
        cellX = std::trunc(x / (landSize - 1)) - 1;
    }
    if (y < 0)
    {
        cellY = std::trunc(y / (landSize - 1)) - 1;
    }

    std::string cellId = "#" + std::to_string(cellX) + " " + std::to_string(cellY);

    std::cout << cellId << std::endl;

    return cellId;
}
```

zinnschlag:
I'll have a closer look at your code within the next couple of days.

Regarding the conversion functions:

The 2nd one looks odd. The idea (if I understood your original statement correctly) was to covert between local and global texture tile coordinates. The 2nd functions doesn't really do that. The return type for the first function (global to local) is also a bit odd. Shouldn't it return something like std::pair<CellCoordiantes, std::pair<int, int> > ?

The math of both functions (but especially the first one) looks very odd. Unfortunately it is much too hot here for me to do any math right now. May have another look at it when the weather is somewhat reasonable and I can think clearly again. Feel free to look over it yourself. I would expect there to be some modulo, maybe even some abs and sign (less sure about the last two).

Also, can we have shorter names, please? For example textureGlobalToLocal instead of terrainTextureSelectionToLocalCoordinates. This isn't specific to the selection anyway, so there is no point in mentioning in the function's name.

unelsson:
Yeah, just

unelsson:
Yeah, just to help reviewing (whenever you have time), some points here. I had to implement a new function CSVRender::WorldspaceWidget* CSVRender::EditMode::getWorldspaceWidgetPointer() even when there was already CSVRender::WorldspaceWidget& CSVRender::EditMode::getWorldspaceWidget(). I wanted a pointer so that no copies would be made, but I don't know if same could have been achieved with passing the reference. This is C++ stuff that still sometimes confuses me, although I've read many tutorials of using pointers.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+            // check adjacent terrain squares to see if there should be a border between them
+            const auto north = std::find(mSelection.begin(), mSelection.end(), std::make_pair(x, y + 1));
+            const auto south = std::find(mSelection.begin(), mSelection.end(), std::make_pair(x, y - 1));
+            const auto east = std::find(mSelection.begin(), mSelection.end(), std::make_pair(x + 1, y));
+            const auto west = std::find(mSelection.begin(), mSelection.end(), std::make_pair(x - 1, y));
+            const int textureSizeToLandSizeModifier = 4; // To-do: Calculate modifier according to landSize and landTextureSize
+
+            // Nudge selection by 1/4th of a texture size, similar how blendmaps are nudged
+            #define NUDGEPERCENTAGE (0.25)
```
zinnschlag: Please do not use the preprocessor for consts.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,251 @@
...
+            const int landHeightsNudge = (cellSize / landSize)/64; // Does this work with all land size configurations?
+
+            // calculate global vertex coordinates at selection box corners
+            int x1 = x * textureSizeToLandSizeModifier + landHeightsNudge;
+            int x2 = x * textureSizeToLandSizeModifier + textureSizeToLandSizeModifier + landHeightsNudge;
+            int y1 = y * textureSizeToLandSizeModifier - landHeightsNudge;
+            int y2 = y * textureSizeToLandSizeModifier + textureSizeToLandSizeModifier - landHeightsNudge;
+
+            // Draw straigth edges
+            int landHeightX1Y1 = calculateLandHeight(x1, y1),
```
zinnschlag: Please don't declare multiple variables that way. We have no shortage of ints. Might as well explicitly write it down for every variable.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,251 @@
...
+            if (west == mSelection.end()) {
+                vertices->push_back(osg::Vec3f(toWorldCoords(x) + nudgeOffset, toWorldCoords(y) - nudgeOffset, landHeightX1Y1+5));
+                vertices->push_back(osg::Vec3f(toWorldCoords(x) + nudgeOffset, toWorldCoords(y + 1) - nudgeOffset, landHeightX1Y2+5));
+            }
+        }
+    }
+
+    newGeometry->setVertexArray(vertices);
+
+    const auto drawArrays = new osg::DrawArrays(osg::PrimitiveSet::LINES);
```
zinnschlag: Not very familiar with osg but this looks fishy (allocating with new and storing in a raw local pointer). Shouldn't that be the osg ref pointer thing?

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,251 @@
...
+{
+    return cellSize * static_cast<double>(pos) / landTextureSize;
+}
+
+size_t CSVRender::TerrainSelection::landIndex(int x, int y)
+{
+    return y * landSize + x;
+}
+
+void CSVRender::TerrainSelection::update()
```
zinnschlag: This is a pretty long function with lots of repetition in it. Might be worth looking at again for a bit of refectoring/simplification.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,251 @@
...
+        cellX = std::trunc(x / (landSize - 1)) - 1;
+        localX = x - cellX * (landSize - 1);
+    }
+    if (y < 0)
+    {
+        cellY = std::trunc(y / (landSize - 1)) - 1;
+        localY = y - cellY * (landSize - 1);
+    }
+
+    std::string cellId = "#" + std::to_string(cellX) + " " + std::to_string(cellY);
```
zinnschlag: Please avoid constructing cell IDs manually.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -105,10 +116,18 @@ void CSVRender::TerrainTextureMode::primaryEditPressed(const WorldspaceHitResult
 
 void CSVRender::TerrainTextureMode::primarySelectPressed(const WorldspaceHitResult& hit)
 {
+    if(hit.hit == true)
```
zinnschlag: Why the == true? Looks redundant.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    const auto x = static_cast<int>(std::floor(xd));
+    const auto y = static_cast<int>(std::floor(yd));
+
+    return std::make_pair(x, y);
+}
+
+std::pair<int, int> CSVRender::TerrainSelection::toVertexCoords(osg::Vec3d worldPos) const
+{
+    const double xd {worldPos.x() * landTextureSize / cellSize+(landTextureSize / cellSize)/4};
```
zinnschlag: I suggest we avoid this method of initialising variables and stay with the = again. There aren't any real advantages to {} and we are just introducing new inconsistencies into the code base.

zinnschlag:
Added some inline comments. One thing I noticed is that there seems to be two ways to indicate an empty selection: mSelection being empty and the mode not having a selection object.

Still need to do a test run. Still building (can only spare one core for OpenMW right). Will report back.

zinnschlag:
Okay done. Seems to work okay. Obviously this is just a very early iteration of the feature. For this stage the rendering of the selection is good enough. But we need to improve it (and make it visually more consistent with the object selection) before we can merge.

apps/opencs/view/render/terrainselection.cpp
unelsson: What's the preferred way to do it then? Remove const or remove #define? Just using the number directly in the formula?

apps/opencs/view/render/terrainselection.cpp
unelsson: Do you mean that I should write "int" in front of each variabte, like.. int landHeightX1Y1; int ...X1Y2; and so on?

apps/opencs/view/render/terrainselection.cpp
unelsson: There has to be a way to convert global terraintexture-coordiantes x and y to a cell ID. What would the option? Would you prefer all this moved to CellCoordinates class?

apps/opencs/view/render/terraintexturemode.cpp
unelsson: If I recall correctly, there were some cases where hits were invalid, but I can't remember where that happened. It's possible this is unnecessary here.

AnyOldName3: The suggestion is that you change `if(hit.hit == true)` to the much simpler `if (hit.hit)`, not that you remove the check altogether (although if it's unnecessary anyway, then obviously it should be removed).

apps/opencs/view/render/terrainselection.cpp
AnyOldName3: It should be an `osg::ref_ptr` just for the sake of being sure. In this specific case, OSG will take responsibility for deleting things when they're done with (as the object's reference count won't have been incremented when it was constructed, so it won't be asking not to be deleted when the geometry is deleted and deletes its internal vector of `ref_ptr`s), but there's also a good chance that in another similar-looking situation it won't work like that.

In general, it's good to follow the rule that for every `new` a function contains, it also contains a `delete` unless the new thing is held in some kind of smart pointer.

apps/opencs/view/render/terrainselection.cpp
zinnschlag: const float NudgePercentage = 0.25f;

apps/opencs/view/render/terrainselection.cpp
zinnschlag: Yes. Declaring multiple variables in a single statement is kinda frowned upon in modern C++.

apps/opencs/view/render/terrainselection.cpp
zinnschlag: Yes.

apps/opencs/view/render/terrainselection.cpp
kcat: For C++11, `constexpr` may be better since it will treat it as a constant expression, rather than a constant variable (the former allowing for more aggressive optimizations). Basically the equivalent of a typed macro value.

apps/opencs/view/render/terrainselection.cpp
unelsson: Btw, is initialising with () okay? In commit 80ee08c I had an oversight and changed {} to (), although you specifically wrote =.

unelsson:
Work and life keeps me busy, so development is slowing down on my part. Anyway, made some fixes according to the reviews. Still to-do: Improving the visuals of the selection (how?), and moving cell Id construction and coordinates handling to mCoordinates class. Also Zini mentioned about rethinking how CSVRender::TerrainSelection::update() is written, which would surely be a good idea, but I have no idea when I'd have time to work on that. Also, there's the issue of having both CSVRender::WorldspaceWidget* CSVRender::EditMode::getWorldspaceWidgetPointer() AND CSVRender::WorldspaceWidget& CSVRender::EditMode::getWorldspaceWidget(), although the latter should suffice, but I had problems making it work.

zinnschlag:
> Improving the visuals of the selection (how?)

Instance selection already implements it. Maybe look t the code? Not sure if it can be used directly, but it should give you an idea at least.

> Also, there's the issue of having both CSVRender::WorldspaceWidget* CSVRender::EditMode::getWorldspaceWidgetPointer() AND CSVRender::WorldspaceWidget& CSVRender::EditMode::getWorldspaceWidget(), although the latter should suffice, but I had problems making it work.

Still very busy, so the amount of mentoring I can provide is limited here. Still, if you describe these problems maybe some help can be provided.

unelsson:
Here goes the question! :smile: 

Functionality of the TerrainSelection needs the ability to read land height data because the terrain textures selection is drawn by lines on the surface of the land. [In current implementation](https://github.com/OpenMW/openmw/blob/80ee08c189537b2ed7d5a7974e6386263f8e9cce/apps/opencs/view/render/terrainselection.cpp#L244) I'm getting the land heights data from WorldspaceWidget->getDocument() through OpenMW's column data structures. In [terraintexturemode.cpp](https://github.com/OpenMW/openmw/blob/80ee08c189537b2ed7d5a7974e6386263f8e9cce/apps/opencs/view/render/terraintexturemode.cpp#L58) I'm currently using **CSVRender::EditMode::getWorldspaceWidgetPointer()** although I would like to use **CSVRender::WorldspaceWidget& CSVRender::EditMode::getWorldspaceWidget()**. I don't understand how passing references works, and [how the function should be declared](https://github.com/OpenMW/openmw/blob/80ee08c189537b2ed7d5a7974e6386263f8e9cce/apps/opencs/view/render/terrainselection.cpp#L25) if not using a pointer. I'm confused where & symbol should go. How should I change the code in order to pass the reference function instead of the pointer?





I want to use the following function:


... to pass WorldspaceWidget

zinnschlag:
Quick and dirty solution: Replace the call to getWorldspaceWidgetPointer() with &getWorldspaceWidget(). This will give you a pointer just like the currently used (and now redundant) function.

Cleaner approach: Go through the code that passes the widget pointer along and replace every * with a & in every type declaration. Note that this will leave you with a reference member in your class. As mentioned before this isn't particular great style, but it is also not a major issue and in this case it won't hurt at all. Also note that now you have a reference instead of a pointer, so you usually will have to access members via . instead of ->

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,251 @@
...
+    const auto x = static_cast<int>(std::floor(xd));
+    const auto y = static_cast<int>(std::floor(yd));
+
+    return std::make_pair(x, y);
+}
+
+std::pair<int, int> CSVRender::TerrainSelection::toVertexCoords(osg::Vec3d worldPos) const
+{
+    const double xd (worldPos.x() * landTextureSize / cellSize+(landTextureSize / cellSize)/4);
+    const double yd (worldPos.y() * landSize / cellSize+(landTextureSize / cellSize)/4);
```
akortunov: Are you sure that here should be the landSize, not the landTextureSize?

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+void CSVRender::TerrainSelection::deactivate()
+{
+    mParentNode->removeChild(mGeode);
+}
+
+std::pair<int, int> CSVRender::TerrainSelection::toTextureCoords(osg::Vec3d worldPos) const
+{
+    const double xd {worldPos.x() * landTextureSize / cellSize+(landTextureSize / cellSize)/4};
+    const double yd {worldPos.y() * landTextureSize / cellSize+(landTextureSize / cellSize)/4};
```
akortunov: It does not look right too.
Technically, you use something like this:
```
xd = landTextureSize / cellSize * (worldPos.x() + 0.25f)
```
I wonder why you need to add the 0.25 to world coordinate.

zinnschlag: I think the original author mentioned something about needing this offset to make texture tiles line up. But I could be remembering incorrectly. It does look odd. Might be a symptom of a problem elsewhere?

apps/opencs/view/render/terrainselection.cpp
zinnschlag: Clearly a bug.

apps/opencs/view/render/terrainselection.cpp
unelsson: I haven't figured out a good implementation for handling texture selection and coordinates. This might not be a good way to do it. The offset is handled in another place, this "/ 4", if I remember correctly, is because vertex coordinates (land geometry) and texture coordinates (land textures) are sized in ratio of 1:4 (16 textures to 64 (+ 1 shared) vertexes).

apps/opencs/view/render/terrainselection.cpp
unelsson: This should be landSize, when calculation vertex coordinates, but landtexturesize in function toTextureCoords.

apps/opencs/view/render/terrainselection.cpp
unelsson: Sorry, zinnschlags explanation was the correct one. Texture nudging is done somewhere by -0.25f and +0.25f, and this needs to be accounted here. Vertex coordinate calculations probably shouldn't be nudged though. (fixed this in latest commit).

lysol90:
Welcome back!

unelsson:
It would be great to have the terrain selection drawn with osgFX::Scribe (adds wireframe), similar as objects, but I wonder if this can be done. Alternatively, drawing the selection lines with GL_depth could be useful, to get the selection drawn as an overlay on top of terrain, but I couldn't figure out how to do this.

unelsson:
My research and thoughts on terrain selection rendering:
1) It's possible to dig out deep primitive data with `osgUtil::LineSegmentIntersector::Intersection`, used in worldspacewidget.cpp. Something like  `osg::Geometry* geom = dynamic_cast<osg::Geometry*>(it->drawable.get());` can be used, where "it" is the LineSegment-Intersection. I find it troublesome to selectively choose triangles, as the intersector returns a chunk, not a primitive triangle.

2) It might be possible to use something along: `osg::Node* nodeTerrain = mTerrain->mChunkManager->getChunk;` at cell.cpp to directly alter `std::unique_ptr<Terrain::TerrainGrid> mTerrain` after it is updated. This could be used to draw a some sort of selection-effect (e.g. changing the color of selected faces without altering the actual terrain data). This doesn't work straight away as mChunkManager is protected, and I wonder if there should be a separate function for getting a pointer of mChunkManager, or if a friend class could be somehow used.

3) Easiest implementation is to draw the selection separately, not altering the drawn terrain directly. Current implementation draws lines at the edges of the selection, slightly above the terrain.

unelsson:
Maybe, but `hit.hit && hit.tag == 0` means that it's an untagged hit, likely a terrain hit, not an object or cell arrow. It's probably not a good idea to assume that everything untagged is terrain. I wonder if that tag-test is needed in the first place? I think that something related to masks is needed. Maybe just taking out the Mask_Reference from the TerrainTextureMode definition? `CSVRender::TerrainTextureMode::TerrainTextureMode (WorldspaceWidget *worldspaceWidget, osg::Group* parentNode, QWidget *parent) : EditMode (worldspaceWidget, QIcon {":scenetoolbar/editing-terrain-texture"}, Mask_Terrain | Mask_Reference, "Terrain texture editing", parent)`

unelsson:
terrainselection.cpp also uses deprecated osg::Geode. I tried to find alternate ways, but I either get some hyperthreading problems (altering live drawables) or I just keep drawing new lines without erasing the old ones. This is another place where I could use help: What's the proper way to draw and erase lines in terrainselection.cpp (without osg::Geode)?

AnyOldName3:
Where you're adding geometry to the geode, just don't do that - you can add geometry to the scene graph directly now (which is why geodes are deprecated).

akortunov:
Just for reference: Git allows you to keep your commits locally and push them at once, so you will not trigger unnecessary CI builds.

unelsson:
> Just for reference: Git allows you to keep your commits locally and push them at once, so you will not trigger unnecessary CI builds.

Thanks for the tip, and sorry for that. I was wondering if I should do that. It's just a habit of pushing commits when I get something useful done. It's quite often that I think I'm finished for now, then push, but decide to code one more small addition...

Btw, travis CI build on mac keeps failing: "Error: cmake: "cxx11" is not a recognized standard". I understand it's a problem of the build, not a problem of this PR?

Anyway, terrain texture selection is pretty much functional now. Some features are lacking, here's a video for ease of reviewing.
https://youtu.be/D7Ya1McDwVw

These issues (features) are still missing:
https://gitlab.com/OpenMW/openmw/issues/3873 
https://gitlab.com/OpenMW/openmw/issues/3869

akortunov:
> Btw, travis CI build on mac keeps failing: "Error: cmake: "cxx11" is not a recognized standard". I understand it's a problem of the build, not a problem of this PR?

Update your branch by commits from master (by either rebase or merge). Ths issue was fixed today.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    mGeometry->setVertexArray(vertices);
+
+    osg::ref_ptr<osg::DrawArrays> drawArrays = new osg::DrawArrays(osg::PrimitiveSet::LINES);
+
+    drawArrays->setCount(vertices->size());
+
+    mGeometry->addPrimitiveSet(drawArrays);
+
+    mSelectionNode->addChild(mGeometry);
```
akortunov: As for me, too much empty lines here. Can you group code lines here?

akortunov:
An additional thoughts:
1. It seems you missed an entry in the CHANGELOG.md ([this feature](https://gitlab.com/OpenMW/openmw/issues/3871), I suppose).
2. Once this PR vill be revieved, maybe it is worth to squash commits, especially if you have ones which do not build.
3. Does #1414 have anything useful which is not present here? Maybe it is the time to close that PR.

unelsson:
This PR is far ahead of #1414, so imho that should be closed. I left out the switch terrainselectiontype, which is still in #1414, but it isn't functional, and implementing that should be easy whenever terrain shape/vertice edit is developed. I didn't work a whole lot on terrain vertice selection, but I think this code is a good base for that later.

I'll add changelog and group those few lines.... I hope it won't trigger another builds though.

About squashing commits, I can't remember if there are commits that wont build. All have worked on my end. Should I just squash them all? Never done that though, so I'd better be careful not to mess up my whole git.

apps/opencs/view/render/terrainselection.cpp
unelsson: Done.

unelsson:
TODO:
- [ ] [The shape of the area affected by the selection change is defined by the state of the brush button.]
- [ ] [Modify changelog.]

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,271 @@
...
+    update();
+}
+
+void CSVRender::TerrainSelection::toggleSelect(const std::vector<std::pair<int, int>> localPositions, bool toggleInProgress)
+{
+    if (toggleInProgress == true)
+    {
+        for(auto const& localPos: localPositions)
+        {
+            auto iter = std::find(mSelection.begin(), mSelection.end(), localPos);
```
Capostrophic: iter is never used if itertemp is found, and std::find complexity is relatively high. You may want to put it into the if outer body.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,271 @@
...
+            if (west == mSelection.end())
+            {
+                for(int i = 1; i < (textureSizeToLandSizeModifier + 1); i++)
+                {
+                    float drawPreviousY = CSMWorld::CellCoordinates::textureSelectionToWorldCoords(y)+(i-1)*(cellSize / (landSize - 1));
+                    float drawCurrentY = CSMWorld::CellCoordinates::textureSelectionToWorldCoords(y)+i*(cellSize / (landSize - 1));
+                    vertices->push_back(osg::Vec3f(CSMWorld::CellCoordinates::textureSelectionToWorldCoords(x) + nudgeOffset, drawPreviousY - nudgeOffset, calculateLandHeight(x1, y1+(i-1))+2));
+                    vertices->push_back(osg::Vec3f(CSMWorld::CellCoordinates::textureSelectionToWorldCoords(x) + nudgeOffset, drawCurrentY - nudgeOffset, calculateLandHeight(x1, y1+i)+2));
+                }
+            }
```
Capostrophic: Is there really no hope for code duplication reduction here? :(

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -372,21 +444,99 @@ void CSVRender::TerrainTextureMode::editTerrainTextureGrid(const WorldspaceHitRe
...
+        selections.emplace_back(texCoords);
+    }
+
+    if (mBrushShape == 1)
+    {
+        for(int i = texCoords.first - r; i <= texCoords.first + r; ++i)
+        {
+            for(int j = texCoords.second - r; j <= texCoords.second + r; ++j)
+            {
+                selections.emplace_back(std::make_pair(i, j));
```
Capostrophic: You don't need to use std::make_pair with emplace_back here and after.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -372,21 +444,99 @@ void CSVRender::TerrainTextureMode::editTerrainTextureGrid(const WorldspaceHitRe
...
+    if (mBrushShape == 2)
+    {
+        for(int i = texCoords.first - r; i <= texCoords.first + r; ++i)
+        {
+            for(int j = texCoords.second - r; j <= texCoords.second + r; ++j)
+            {
+                int distanceX = abs(i - texCoords.first);
+                int distanceY = abs(j - texCoords.second);
+                int distance = std::round(sqrt(pow(distanceX, 2)+pow(distanceY, 2)));
+                if (distance < r) selections.emplace_back(std::make_pair(i, j));
```
Capostrophic: I noticed of course that this syntax is used elsewhere, but I wonder if these loops could be rewritten like
```cpp
        for (int i = -r; i <= r; i++)
        {
            for (int j = -r; j <= r; j++)
            {
                osg::Vec2f coords(i,j);
                if (std::round(coords.length()) < r)
                    selections.emplace_back(i + texCoords.first, j + texCoords.second);
            }
        }
```

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -418,18 +568,21 @@ void CSVRender::TerrainTextureMode::createTexture(std::string textureFileName)
...
-        } catch (const std::exception& e)
+        }
+        catch (const std::exception& e)
         {
             newId = CSMWorld::LandTexture::createUniqueRecordId(0, counter);
             freeIndexFound = true;
         }
-    } while (freeIndexFound == false);
+    }
+    while (freeIndexFound == false);
```
Capostrophic: You probably shouldn't touch the formatting of this function, it's not that outrageous.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -372,21 +444,99 @@ void CSVRender::TerrainTextureMode::editTerrainTextureGrid(const WorldspaceHitRe
...
+                }
+                else
+                {
+                    int cellXDifference = std::floor(1.0f*(xHitInCell + value.first)/landTextureSize);
+                    int cellYDifference = std::floor(1.0f*(yHitInCell + value.second)/landTextureSize);
+                    int xInOtherCell = xHitInCell + value.first - cellXDifference * landTextureSize;
+                    int yInOtherCell = yHitInCell + value.second - cellYDifference * landTextureSize;
+
+                    std::string cellId = CSMWorld::CellCoordinates::generateId(cellX+cellXDifference, cellY+cellYDifference);
+                    if (allowLandTextureEditing(cellId)==true)
```
Capostrophic: This comparison is always true.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -91,7 +108,7 @@ void CSVRender::TerrainTextureMode::primaryEditPressed(const WorldspaceHitResult
     CSMWorld::IdCollection<CSMWorld::LandTexture>& landtexturesCollection = document.getData().getLandTextures();
     int index = landtexturesCollection.searchId(mBrushTexture);
 
-    if (index != -1 && !landtexturesCollection.getRecord(index).isDeleted() && hit.hit == true)
+    if (index != -1 && !landtexturesCollection.getRecord(index).isDeleted() &&  hit.hit && hit.tag == 0)
```
Capostrophic: Extra space.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -534,9 +688,41 @@ void CSVRender::TerrainTextureMode::setBrushSize(int brushSize)
...
+        }
+    }
 }
 
 void CSVRender::TerrainTextureMode::setBrushTexture(std::string brushTexture)
 {
     mBrushTexture = brushTexture;
 }
+
+CSVRender::PagedWorldspaceWidget& CSVRender::TerrainTextureMode::getPagedWorldspaceWidget()
```
Capostrophic: This function doesn't seem to be used anymore.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -534,9 +688,41 @@ void CSVRender::TerrainTextureMode::setBrushSize(int brushSize)
...
+        selectionCenterX = selectionCenterX / selectionAmount;
+        selectionCenterY = selectionCenterY / selectionAmount;
+
+        mCustomBrushShape.clear();
+        std::pair<int, int> differentialPos {};
+        for(auto const& value: terrainSelection)
+        {
+            differentialPos.first = value.first - selectionCenterX;
+            differentialPos.second = value.second - selectionCenterY;
+            mCustomBrushShape.push_back(differentialPos);
```
Capostrophic: Maybe use something like
```cpp
        mCustomBrushShape.clear();
        for(auto const& value: terrainSelection)
            mCustomBrushShape.emplace_back(value.first - selectionCenterX, value.second - selectionCenterY);
```

apps/opencs/view/render/terraintexturemode.cpp
unelsson: I want to put a std::pair, holding values of i and j, in a vector. Is there another option?

apps/opencs/view/render/terrainselection.cpp
unelsson: I haven't found a way yet. Any ideas?

apps/opencs/view/render/terrainselection.cpp
unelsson: Good point!

apps/opencs/view/render/terraintexturemode.cpp
unelsson: Probably they could. That looks okay. I wrote it on multiple lines in this syntax to cut the process into parts, to make formulas easier to understand, but I don't think your proposition is unclear. Umm.. Is this the way how pairs can be made with emplace_back?

apps/opencs/view/render/terraintexturemode.cpp
unelsson: Precise reviewing. Thanks, will fix.

apps/opencs/view/render/terraintexturemode.cpp
Capostrophic: Yes.

apps/opencs/view/render/terraintexturemode.cpp
unelsson: True that. Removed.

apps/opencs/view/render/terraintexturemode.cpp
unelsson: That's nice. Does const& ref here avoid making a copy of the variable?

apps/opencs/view/render/terraintexturemode.cpp
unelsson: Yeh. Not very significant maybe, but changed it back.

apps/opencs/view/render/terraintexturemode.hpp
```diff
@@ -18,22 +19,34 @@
...
+#include "terrainselection.hpp"
+
 namespace CSVWidget
 {
     class SceneToolTextureBrush;
 }
 
 namespace CSVRender
 {
+    class PagedWorldspaceWidget;
```
Capostrophic: This forward declaration is no longer necessary.

apps/opencs/view/render/terraintexturemode.hpp
```diff
@@ -18,22 +19,34 @@
 #include "../../model/world/landtexture.hpp"
 #endif
 
+#include "terrainselection.hpp"
```
Capostrophic: I think it should be possible to forward-declare TerrainSelection instead of including this header.

unelsson: I tried this, then added missing include osg/Group, but this still results in some hefty compile errors. "/usr/include/c++/8.2.1/bits/unique_ptr.h:79:16: error: invalid application of ‘sizeof’ to incomplete type ‘CSVRender::TerrainSelection’" Doesn't ring any obvious bells how to fix this.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,271 @@
...
+    {
+        mDraggedOperationFlag = false;
+        mTemporarySelection.clear();
+    }
+    update();
+}
+
+void CSVRender::TerrainSelection::activate()
+{
+    mParentNode->addChild(mSelectionNode);
```
Capostrophic: So as scrawl said you should add a `if (!mParentNode->containsNode(mSelectionNode))` check here.

unelsson: Right.. And `if (mParentNode->containsNode(mSelectionNode))` for removeChild?

Capostrophic: No, removing non-existent child is safe, i.e. no-op.

apps/opencs/view/render/terraintexturemode.hpp
```diff
@@ -6,6 +6,8 @@
 #include <string>
 #include <memory>
 
+#include <osg/Group>
```
Capostrophic: Forward-declare it.

apps/opencs/view/render/terraintexturemode.cpp
Capostrophic: Well, yes, since it's a reference to the element.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,271 @@
...
+    {
+        mSelection.emplace_back(value);
+    }
+    update();
+}
+
+void CSVRender::TerrainSelection::addSelect(const std::pair<int, int> localPos)
+{
+    if (std::find(mSelection.begin(), mSelection.end(), localPos) == mSelection.end())
+            mSelection.emplace_back(localPos);
```
Capostrophic: Four extra spaces here.

Is update call necessary if the selection wasn't changed?

apps/opencs/view/render/terraintexturemode.hpp
```diff
@@ -66,6 +79,9 @@ namespace CSVRender
             /// \brief Handle brush mechanics, maths regarding worldspace hit etc.
             void editTerrainTextureGrid (const WorldspaceHitResult& hit);
 
+            /// \brief Handle brush mechanics for texture selection
+            void selectTerrainTextures (std::pair<int, int>, unsigned char, bool);
```
Capostrophic: Since the coords pair isn't modified, it should be passed by reference, not a copy. The arguments should be named here for clarity of their purpose too.

apps/opencs/view/render/terraintexturemode.hpp
Capostrophic: OK, never mind, it can't be done cleanly.

Capostrophic:
When I extend the selection with Ctrl-Click, this happens when the selections overlap: 
изображение
Pretty odd.

apps/opencs/view/render/terraintexturemode.hpp
unelsson: Ok.

apps/opencs/view/render/terrainselection.cpp
unelsson: I'll put update and emplace_back inside {} to prevent unnecessary update calls.

apps/opencs/view/render/terraintexturemode.hpp
unelsson: True, I'll fix this.

unelsson:
> When I use secondary select, this happens when the selections overlap:
> изображение
> Pretty odd. What is it supposed to do?

I tried several versions of this during development. Now it toggles selected to unselected, and unselected to selected. This is the best toggle-type functionality I could think of.

AnyOldName3:
Might it be sensible to have a button to toggle between AND, OR and XOR instead of hardcoding in XOR?

unelsson:
You mean AND/OR/XOR how secondary edit works? Extra button adds UI complexity, so I'm leaning towards hardcoding one functionality, though I also thought of possibility to do on/off toggle based on status of the center of click. Either way, doesn't seem like the best UI imho, but I understood Zini was planning some sort of toggle.

I'd personally see it handy to have only add-to-selection and remove-from-selection, maybe tied straight to primary and secondary selection. Anyhow, if consensus or clear plan is lacking, this can also be changed later, it's somewhat easy to change.

unelsson:
Almost forgot - should texture selection select textures only from loaded cells? Now it selects everything, regardless if cell is loaded or not.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -34,14 +36,18 @@
 #include "pagedworldspacewidget.hpp"
 #include "mask.hpp"
 #include "object.hpp" // Something small needed regarding pointers from here ()
+#include "terrainselection.hpp"
```
Capostrophic: Since terraintexturemode.hpp already depends on this header, this include is redundant.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,272 @@
...
+    return mSelection;
+}
+
+void CSVRender::TerrainSelection::onlySelect(const std::vector<std::pair<int, int>> localPositions)
+{
+    mSelection.clear();
+    for(auto const& value: localPositions)
+    {
+        mSelection.emplace_back(value);
+    }
```
Capostrophic: You're copying the vector, right? Can't something like ```mSelection = localPositions``` be done instead?

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,272 @@
...
+    if (std::find(mSelection.begin(), mSelection.end(), localPos) == mSelection.end())
+    {
+        mSelection.emplace_back(localPos);
+        update();
+    }
+}
+
+void CSVRender::TerrainSelection::toggleSelect(const std::vector<std::pair<int, int>> localPositions, bool toggleInProgress)
+{
+    if (toggleInProgress == true)
```
Capostrophic: It could be just me, but I still feel like this (and the lower check) is oververbose syntax.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,272 @@
...
+    drawArrays->setCount(vertices->size());
+    mGeometry->addPrimitiveSet(drawArrays);
+    mSelectionNode->addChild(mGeometry);
+}
+
+void CSVRender::TerrainSelection::drawShapeSelection(const osg::ref_ptr<osg::Vec3Array> vertices)
+{
+    if (!mSelection.empty())
+    {
+        for (std::pair<int, int> localPos : mSelection)
```
Capostrophic: It shouldn't be necessary to copy stuff here (especially considering that localPos only gets used for declaring x and y).

```
for (const std::pair<int, int> &localPos : mSelection)
```

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,272 @@
...
+    if (!mSelection.empty())
+    {
+        // Nudge selection by 1/4th of a texture size, similar how blendmaps are nudged
+        const float nudgePercentage = 0.25f;
+        const int nudgeOffset = (cellSize / landTextureSize) * nudgePercentage;
+        const int landHeightsNudge = (cellSize / landSize) / (landSize - 1); // Does this work with all land size configurations?
+
+        const int textureSizeToLandSizeModifier = (landSize - 1) / landTextureSize;
+
+        for (std::pair<int, int> localPos : mSelection)
```
Capostrophic: Also here.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -148,47 +176,91 @@ bool CSVRender::TerrainTextureMode::secondaryEditStartDrag (const QPoint& pos)
...
-    int index = landtexturesCollection.searchId(mBrushTexture);
+        CSMWorld::IdCollection<CSMWorld::LandTexture>& landtexturesCollection = document.getData().getLandTextures();
+        int index = landtexturesCollection.searchId(mBrushTexture);
 
-    if (index != -1 && !landtexturesCollection.getRecord(index).isDeleted())
-    {
-         undoStack.endMacro();
+        if (index != -1 && !landtexturesCollection.getRecord(index).isDeleted())
+        {
+             if (mIsEditing == true)
```
Capostrophic: Also too verbose.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -372,21 +444,98 @@ void CSVRender::TerrainTextureMode::editTerrainTextureGrid(const WorldspaceHitRe
...
+                }
+                else
+                {
+                    int cellXDifference = std::floor(1.0f*(xHitInCell + value.first)/landTextureSize);
+                    int cellYDifference = std::floor(1.0f*(yHitInCell + value.second)/landTextureSize);
+                    int xInOtherCell = xHitInCell + value.first - cellXDifference * landTextureSize;
+                    int yInOtherCell = yHitInCell + value.second - cellYDifference * landTextureSize;
+
+                    std::string cellId = CSMWorld::CellCoordinates::generateId(cellX+cellXDifference, cellY+cellYDifference);
+                    if (allowLandTextureEditing(cellId)==true)
```
Capostrophic: Also too verbose.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,272 @@
...
+{
+    deactivate();
+}
+
+std::vector<std::pair<int, int>> CSVRender::TerrainSelection::getTerrainSelection() const
+{
+    return mSelection;
+}
+
+void CSVRender::TerrainSelection::onlySelect(const std::vector<std::pair<int, int>> localPositions)
```
Capostrophic: onlySelect calls will copy the argument. :(

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,272 @@
...
+{
+    mSelection.clear();
+    for(auto const& value: localPositions)
+    {
+        mSelection.emplace_back(value);
+    }
+    update();
+}
+
+void CSVRender::TerrainSelection::addSelect(const std::pair<int, int> localPos)
```
Capostrophic: Another copy.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -372,21 +443,130 @@ void CSVRender::TerrainTextureMode::editTerrainTextureGrid(const WorldspaceHitRe
...
+        }
     }
+}
 
+bool CSVRender::TerrainTextureMode::isInCellSelection(const int& globalSelectionX, const int& globalSelectionY)
+{
+    if (CSVRender::PagedWorldspaceWidget *paged = dynamic_cast<CSVRender::PagedWorldspaceWidget *> (&getWorldspaceWidget()))
+    {
+        CSMWorld::CellSelection selection = paged->getCellSelection();
+        if (selection.has (CSMWorld::CellCoordinates::fromId(
```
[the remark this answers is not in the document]
unelsson: Yeah, this is one of the somewhat long if sentences, and I don't really know how they should be written. Is there a style guide or something I could refer to?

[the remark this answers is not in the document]
Capostrophic: I don't think there's a style guide but maybe you can do it like
```cpp
std::pair<int, int> textureCoords = std::make_pair(globalSelectionX, globalSelectionY);
std::string cellId = CSMWorld::CellCoordinates::textureGlobalToCellId(textureCoords);
return paged->getCellSelection().has(CSMWorld::CellCoordinates::fromId(cellId).first);
```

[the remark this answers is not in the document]
unelsson: Yeaah, that is more readable. Committed.

apps/opencs/model/world/cellcoordinates.hpp
```diff
@@ -49,10 +49,10 @@ namespace CSMWorld
...
             ///Converts worldspace coordinates to global texture selection, taking in account the texture offset.
-            static std::pair<int, int> toTextureCoords(osg::Vec3d worldPos);
+            static std::pair<int, int> toTextureCoords(const osg::Vec3d& worldPos);
 
             ///Converts worldspace coordinates to global vertex selection.
-            static std::pair<int, int> toVertexCoords(osg::Vec3d worldPos);
+            static std::pair<int, int> toVertexCoords(const osg::Vec3d& worldPos);
 
             ///Converts global texture coordinate to worldspace coordinate that is at the upper left corner of the selected texture.
             static float textureSelectionToWorldCoords(int);
```
Capostrophic: Here and in further declarations it's odd that the arguments are not named in the header specifically, it would result in extra reading for a future contributor to tell what does the function actually accept as input.

apps/opencs/view/render/terrainselection.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+CSVRender::TerrainSelection::TerrainSelection(osg::Group* parentNode, WorldspaceWidget *worldspaceWidget, TerrainSelectionType type):
+mParentNode(parentNode), mWorldspaceWidget (worldspaceWidget), mDraggedOperationFlag(false), mSelectionType(type)
+{
+    mGeometry = new osg::Geometry();
+
+    mSelectionNode = new osg::Group();
+    mSelectionNode->addChild(mGeometry);
+
+    activate();
```
Capostrophic: Are activate() and deactivate() used anywhere other than TerrainSelection constructor and destructor?

unelsson: Not yet, I don't have a clear idea when to delete and when to store the terrain selection. They are not necessary at the moment, but I kept those functions to accommodate easy development.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -148,47 +175,91 @@ bool CSVRender::TerrainTextureMode::secondaryEditStartDrag (const QPoint& pos)
...
+                 undoStack.endMacro();
+                 mIsEditing = false;
+             }
+        }
     }
 }
 
-void CSVRender::TerrainTextureMode::dragAborted() {
+void CSVRender::TerrainTextureMode::dragAborted()
+{
```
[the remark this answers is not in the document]
unelsson: Thanks for pointing this out, but I don't think it's worth it to cut this one into a separate PR, nor is it worth it to make a separate commit to revert this. The same with the other ones you mentioned.

Capostrophic:
Might be worth checking if std::set logarithmic search and better duplicate handling goodie can be traded for the std::vector contiguous memory iteration speed in case of mSelection and related stuff.

unelsson:
> Might be worth checking if std::set logarithmic search and better duplicate handling goodie can be traded for the std::vector contiguous memory iteration speed in case of mSelection and related stuff.

I've done some small tests with different methods, but haven't got too far into optimization. I'd rather leave it from the scope of this PR. Thanks for pointing that out anyway.

akortunov:
Which issues are left here excepts of merge conflict?

unelsson:
> Which issues are left here excepts of merge conflict?

I'm not aware of any issues.

unelsson:
Tried to rebase, got into trouble with merge conflicts, which resulted in total destruction.

PR #1770, opened 2018-06-19
Make Open spells casted by anything trigger player crime event (bug #4461)

Capostrophic:
[Bug 4461](https://gitlab.com/OpenMW/openmw/issues/4461)

~~I really like rot tor's game mechanics bug reports, they're simple to fix~~

If anything has managed to cast an Open spell successfully, the player is blamed for that instead. Situation from the bug report seems to work fine.

I noticed that OpenMW crashes if you try to use ExplodeSpell on a door in a different cell. It's not related to this specific change, it also happens on master. Looking at the crashlog, it seems to fail to add spellcasting glow on the object? (the last call is MWRender::Animation::addSpellCastGlow)
Edit: the crash also happens on OpenMW 0.43.0. I suggest to file a bug report.
Edit 2: I managed to fix the crashes, apparently the animation pointer would have been null in this case. Not sure if it's worth it to track every small optimization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
     Bug #4457: Item without CanCarry flag prevents shield autoequipping in dark areas
     Bug #4458: AiWander console command handles idle chances incorrectly
     Bug #4459: NotCell dialogue condition doesn't support partial matches
+    Bug #4461: "Open" spell from non-player caster isn't a crime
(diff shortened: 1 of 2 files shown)


zinnschlag:
Go ahead!

Capostrophic:
Fixed the crashes.

psi29a:
Thanks! Merging.

PR #1771, opened 2018-06-19
Retrieve SDL window settings instead of using magic numbers

Xenkhan:
FIXME asked for getting values through SDL instead of magic numbers, provided way to.


Capostrophic:
Works fine for me.

kcat:
Does this actually work as intended (get the in-use values from the window, rather than what was set by the app)? Also those `reinterpret_cast`s are kind of sketchy looking (they always are). If possible, it would be better to use actual `int` variables that get copied into the traits separately.

Xenkhan:
Can do! I'll attempt that tomorrow when I have time

psi29a:
Looks good (and clean) to me.

psi29a:
Maybe a changelog entry near the bottom:
```
FIXME: Retrieve SDL window settings instead of using magic numbers
```
?

zinnschlag:
That should get a proper issue on our GitLab tracker.

psi29a:
Created an issue to track it here:
https://gitlab.com/OpenMW/openmw/issues/4471

Just update the changelog and we'll merge it in. :)

Xenkhan:
Done! Updated it correctly I think.

psi29a:
Cheers!

nikolaykasyanov:
It's crashing for me:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x3b0)
  * frame #0: 0x00007fff331fce2d libGL.dylib`glGetString + 15
    frame #1: 0x0000000102d4a98e libSDL2-2.0.0.dylib`SDL_GL_GetAttribute_REAL + 1022
    frame #2: 0x0000000102c673ca libSDL2-2.0.0.dylib`SDL_GL_GetAttribute + 42
    frame #3: 0x00000001003b948a openmw`OMW::Engine::createWindow(Settings::Manager&) + 1466
    frame #4: 0x00000001003ba912 openmw`OMW::Engine::prepareEngine(Settings::Manager&) + 242
    frame #5: 0x00000001003bd2d3 openmw`OMW::Engine::go() + 2291
    frame #6: 0x00000001003ad21f openmw`main + 1855
    frame #7: 0x00007fff5168c015 libdyld.dylib`start + 1
    frame #8: 0x00007fff5168c015 libdyld.dylib`start + 1
```

both local builds & nightlies from Travis.

nikolaykasyanov:
It seems that `glGetString(GL_VERSION)` [requires](https://lists.apple.com/archives/quartz-dev/2010/Jul/msg00042.html) an existing OpenGL context on macOS, otherwise it's undefined behaviour (crash in our case).

I performed the following experiment: I put a `glGetString(GL_VERSION)` call right before the first `SDL_GL_GetAttribute` calls and `glGetString` crashed. Then I reverted this PR and put the `glGetString` call right after `new SDLUtil::GraphicsWindowSDL2(traits)` line to verify that it wouldn't crash (it didn't).

Conclusion: it's an SDL bug, either in documentation (`SDL_GL_GetAttribute` should explicitly require OpenGL context, its current documentation doesn't) or in implementation (`SDL_GL_GetAttribute` shouldn't rely on `glGetString(GL_VERSION)` call).

My suggestion would be to roll this PR back and to report the issue to SDL.

nikolaykasyanov:
The `glGetString(GL_VERSION)` requiring an OpenGL context doesn't seem platform-specific though: https://stackoverflow.com/questions/6288759/why-could-glgetstringgl-version-be-causing-a-seg-fault.

Xenkhan:
If it doesn't work for macOSX then I agree it'd be best to roll it back for now. Don't have a mac to test it so I'm not sure whether it'd work on it or not.

psi29a:
So the issue is between 391 and 397, that for whatever reason... we don't have permission to copy from that data structure? This is via SDL2?

nikolaykasyanov:
@psi29a could you elaborate please? `SDL_GL_GetAttribute` calls cause the issue because SDL uses `glGetString(GL_VERSION)` (there's a [branch](https://github.com/SDL-mirror/SDL/blob/f136438f7133f9e8ee408ebeaa81a364f50abe22/src/video/SDL_video.c#L2890) based on OpenGL version) in its implementation and there's no OpenGL context when `SDL_GL_GetAttribute` is called, thus `glGetString(...)` call crashes.

@Xenkhan thanks for the response, no worries at all, it's unrealistic to expect everyone to test on all platforms.

psi29a:
So it's a bug in SDL2? Should we revert this temporarily until it is fixed upstream?

kcat:
I'm not sure it's a bug, it's just how it works designed. `SDL_GL_GetAttribute` would call `glGetString` to get the GL version from the current context so that it knows what can be queried, but there is no context. It's ultimately a chicken and egg problem, and the reason why the code is like it was in the first place. It needs a context to get the current attributes, but it needs the attributes to set up the context. So it just specifies the same values that were given to SDL in the hopes that it's good enough to what the window eventually gets.

psi29a:
So to fix this on all platforms is to revert and modify the FIXME to say _why_ we did it this way. We can even point to this PR.

Capostrophic:
[>_>](https://bugs.openmw.org/issues/4465)

Damn users using the **read-only** tracker which has been deprecated for days.

nikolaykasyanov:
@kcat `SDL_GL_GetAttribute`'s documentation could at least state that it requires an OpenGL context.

nikolaykasyanov:
@psi29a I'm on it.

PR #1772, opened 2018-06-20
Fast-forward death animation to end if death animation was finished earlier

akortunov:
Fixes [bug #4468](https://gitlab.com/OpenMW/openmw/issues/4468).

Note: it is a regression, so there is no changelog entry.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -2026,10 +2026,11 @@ void CharacterController::update(float duration)
     {
         // initial start of death animation for actors that started the game as dead
         // not done in constructor since we need to give scripts a chance to set the mSkipAnim flag
-        if (!mSkipAnim && mDeathState != CharState_None && mCurrentDeath.empty() && cls.isPersistent(mPtr))
+        if (!mSkipAnim && mDeathState != CharState_None && mCurrentDeath.empty())
         {
-            // Fast-forward death animation to end for persisting corpses
-            playDeath(1.f, mDeathState);
+            // Fast-forward death animation to end for persisting corpses or corpses after end of death animation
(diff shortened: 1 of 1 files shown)


psi29a:
It still has a bug # attributed to it, so it still needs to be marked in the changelog.

A changelog-less PR/MR is only for regressions within a release (we are working on 0.45 release now).

This PR fixes a regression from 0.43, changelog entry please. :)

akortunov:
> This PR fixes a regression from 0.43, changelog entry please. :)

W-what? This regression was added only yesterday by my PR #1749.

psi29a:
Sorry, have I misread something?  
https://gitlab.com/OpenMW/openmw/issues/4468   <-- that's the bug this will fix right?

> Since 0.43.00+git201806191146~ubuntu16.04.1 anything that you kill will be up after reloading. Dead but up their feet.

This implies that it's in 0.44, and also in our RCs.

Since it has an issue associated with it, there should be a changelog entry too right?

Capostrophic:
>Either way, since it has an issue associated with it, there should be a changelog entry too right?
Um, no, it's a regression. Regressions do not deserve a changelog entry. That's the point of regressions.

akortunov:
I do not know we name Ubuntu nighlies, but I suppose we still use 0.43 in their names.
git201806191146 - it means "master branch, June 19", right?

Capostrophic:
>git201806191146 - it means "master branch, June 19", right?

It does.

psi29a:
> Um, no, it's a regression. Regressions do not deserve a changelog entry. That's the point of regressions.

But it has an issue associated with it, if a ticket is closed, then it deserves to go in the changelog for that release, regression or not.

psi29a:
I see your point, changelog isn't a dumping ground for closed tickets... only things changed from release to release.  My apologies for being grumpy.

PR #1775, opened 2018-06-22
[Discussion] Option to discard small amount of movement from animations

akortunov:
This PR is not aimed to be merged upstream yet.
Attempt to fix the cause of [bug #1298](https://gitlab.com/OpenMW/openmw/issues/1298).

The main idea: add an option to treat amount of resulting movement in animations below given threshold as non-intended and discard it. Zero by default (vanilla behaviour).
Can be useful when player can not fix animations himself.

Note: "discard movement threshold = 0.1" allows to workaround most of animation issues for scamps.

### DIFF
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -614,6 +614,7 @@ namespace MWRender
         , mPtr(ptr)
         , mResourceSystem(resourceSystem)
         , mAccumulate(1.f, 1.f, 0.f)
+        , mDiscardThresholdSquare(0)
         , mTextKeyListener(NULL)
         , mHeadYawRadians(0.f)
         , mHeadPitchRadians(0.f)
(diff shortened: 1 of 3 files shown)


kcat:
This will almost certainly create weird looking animations for such idles, like sliding feet (the hips are supposed to move left and right while the feet remain stationary, but the hips get locked in place with the movement node so the feet end up moving right and left relative to the stationary hips).

wareya:
They already have sliding feet.

akortunov:
> Merge conflict... :(

Please, do not touch this PR yet.

psi29a:
Just saying...

zinnschlag:
That is a difficult one. This is now my third attempt at writing a reply (discarded the other two as bad ideas). 

I don't think there is a good solution that is available to us right now. We could use either option 1 or 2, but this would have to be decided on a per-NPC basis (configurable in the content file). That is something we can't do pre-1.0.

Unless someone can come up with a really clever idea how to automatically determine to which NPCs a fix should be applied, the only option I see is to blame the resources and add more features to deal with them after 1.0.

akortunov:
Well, since we do not have good options pre-1.0, closing this PR now.

PR #1776, opened 2018-06-22
Apply weapon reload animations only for upper body

akortunov:
Allows to fix a lot of problems with crossbowmans.

Currently in both vanilla and OpenMW there is a twitching when actor starts or stops walking during crossbow reload animation. It looks very weird, so I decided to apply such animation only for upper body.
I managed to solve issues with missing 1th-person mode crossbow animations: I use the 1h animations as fallback.
As for me, it looks decent.

Can we fix this issue from engine side, or it is a strictly up to content developers?

An another approach would be to stop shooter until the end of reload, but I suppose players will not be happy about it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
     Bug #2852: No murder bounty when a player follower commits murder
     Bug #2862: [macOS] Can't quit launcher using Command-Q or OpenMW->Quit
     Bug #2971: Compiler did not reject lines with naked expressions beginning with x.y
+    Bug #3249: Fixed revert function not updating views properly
     Bug #3374: Touch spells not hitting kwama foragers
     Bug #3486: [Mod] NPC Commands does not work
     Bug #3591: Angled hit distance too low
@@ -59,7 +60,7 @@
     Bug #4480: Segfault in QuickKeysMenu when item no longer in inventory
     Bug #4489: Goodbye doesn't block dialogue hyperlinks
     Bug #4490: PositionCell on player gives "Error: tried to add local script twice"
-    Bug #3249: Fixed revert function not updating views properly
+    Bug #4495: Crossbow animations blending is buggy
     Feature #2606: Editor: Implemented (optional) case sensitive global search
     Feature #3276: Editor: Search- Show number of (remaining) search results and indicate a search without any results
     Feature #3641: Editor: Limit FPS in 3d preview window
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1627,16 +1627,18 @@ bool CharacterController::updateWeaponState()
                 break;
         }
 
+        // Note: apply reload animations only for upper body since blending with movement animations can give weird result.
+        // Especially noticable with crossbow reload animation.
         if(!start.empty())
         {
             mAnimation->disable(mCurrentWeapon);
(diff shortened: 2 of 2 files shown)


psi29a:
So long as reloading a weapon doesn't involve the whole body, this is a nice solution that fixes an issue from vanilla as well.

Realistically one will stay still while reloading a crossbow...  but TES was never really realistic and people tend to FPS it up.

akortunov:
> Does anyone have issues with 1st-person crossbow animations during movement in upstream OpenMW?

I suppose I have issues because Morrowind does not have movement animations for crossbow for some reason, and out current approach (apply basic movement animation with BlendMask_LowerBody) does not work well for some reason for 1st-person animations.

TaxAkla:
I use the Per-KF Animation Feature. Let's say that I get the Almalexia Casting (Three Files) from Animation Compilation and put them to overwrite the Beast ones. What I get now is Beasts castings Almalexia style spells with human legs animation only when standing still but with Beast legs while moving around.

Question: If merged will this pull give us Beasts with Almalexia Castings but with the correct legs for them?

I attached a example mod to easier tests.
[Beasts with Almalexia Casting Animations.zip](https://github.com/OpenMW/openmw/files/2127942/Beasts.with.Almalexia.Casting.Animations.zip)

akortunov:
I managed to solve issues with missing 1th-person mode crossbow animations: I use the 1h animations as fallback.
It is not 100% vanilla behaviour (for example, vanilla does not handle sneaking pose with crossbows well), but it is much better than our current behaviour.

akortunov:
> Question: If merged will this pull give us Beasts with Almalexia Castings but with the correct legs for them?

Nope, you are out of luck. You will need to edit cast animations anyway.

akortunov:
Any suggestions or issues with this PR?

psi29a:
@akortunov nothing more to discuss then I propose to merge, please remove discussion from title when you're ready. :)

akortunov:
I'll add the issue on bugtracker and changelog entry then.

akortunov:
> I'll add the issue on bugtracker and changelog entry then.

Done.

Capostrophic:
Seems to be relevant to this PR: when spellcasting stance is used, the rotating animation in OpenMW is not correct. In Morrowind it's much closer to the idle position of the legs and much less twitchy. Maybe 1h animations should be used for spellcasting stance too as fallback. I'll try that out and report back.

akortunov:
> I'll try that out and report back.

I already tried that several days ago and it did not work for me.
But probably you will achieve better results.

Capostrophic:
It's somewhat better, but it's still twitchy and inaccurate. What does Morrowind even utilize for spellcasting animations..?

Capostrophic:
Does anything break when you change the short group of crossbow weapon type to 1h instead (line 216)?

akortunov:
> Does anything break when you change the short group of crossbow weapon type to 1h instead of using a workaround (line 216)?

I think it is a bad idea - it will replace ALL crossbow animations instead of just missing jumping and idle ones and only when they are missing.

Capostrophic:
Already realized that and deleted the comment.

Maybe Morrowind uses some kind of different animation blending for spellcasting stance rotating animation specifically. I couldn't find any similar animation group. Maybe I'm missing something.

akortunov:
I suggest to try to fix turning animations in #1666 instead.

psi29a:
Going once... going twice... merged!

Thank you @akortunov

PR #1777, opened 2018-06-23
Consider <p> tag when discarding post-EOL tag text (regression #4473) 

Capostrophic:
[Regression #4473](https://gitlab.com/OpenMW/openmw/issues/4473).

Seems to work fine. On master the text in Note to Hrisskar is discarded, after the changes it isn't.


psi29a:
Thanks for catching this. :)

zinnschlag:
If I am not mistaken this is a regression that was introduced only after the 0.44.0 release. If that is the case we shouldn't have an entry in the changelog for 0.45.0, because this fix isn't a fix related to 0.44.0 but to 0.45.0. We already have a changelog entry (#4215). This is merely a patch for it.

PR #1779, opened 2018-06-25
Fix some code coverity issues

akortunov:
Attempt to fix some issues, found by Coverity Scan (mostly missing fields initialization).


psi29a:
Thank you for this, this also shuts up some clang and gcc warnings.

PR #1780, opened 2018-06-27
Ignore movement from scripted animations

akortunov:
Fixes [bug #4475](https://gitlab.com/OpenMW/openmw/issues/4475).
Abot's Silt Striders should work fine now.


psi29a:
Thanks!

PR #1781, opened 2018-06-27
[Discussion] Refactor AiState usage

akortunov:
Fixes [bug #4464](https://gitlab.com/OpenMW/openmw/issues/4464).

The main idea: since we do not handle AiState in Actor code, we can move it into AiSequence and clean storage when we cancel AI packages.

Can we move AiState to target AI package instead?
I use this approach [here](https://github.com/akortunov/openmw/tree/removeAiState).
Since the storage is a part of AI package here, we do not need to clean it manually after AI package expiration, also AI API is cleaner.


zinnschlag:
That doesn't look like a bad idea. Would like to have someone else (who is more familiar with our current AI implementation) to have another look at this.

zinnschlag:
Nobody has been yelling at us, so I will go ahead and merge.

PR #1782, opened 2018-06-27
Specialization/attribute/skill randomization in class creation widget

wareya:
The button text is defined in the layout file (like it is for a bunch of the settings menus) since I have no idea where else to put it.

### DIFF
--- a/apps/openmw/mwgui/class.cpp
+++ b/apps/openmw/mwgui/class.cpp
@@ -12,7 +12,7 @@
 
 #include "../mwworld/esmstore.hpp"
 
-#include "../../../components/misc/rng.hpp"
+#include <components/misc/rng.hpp>
 
 #include "tooltips.hpp"
 
--- a/apps/openmw/mwgui/class.cpp
+++ b/apps/openmw/mwgui/class.cpp
@@ -16,6 +16,9 @@
 
 #include "tooltips.hpp"
 
+// for std::sort
+#include <algorithm>
+
 namespace
 {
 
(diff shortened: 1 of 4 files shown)


wareya:
Note: the way random skills are selected is still buggy, I just wanted to get the PR up here.

image

Note Short Blade being in both columns.

wareya:
Alright that should fix it.

apps/openmw/mwgui/class.cpp
```diff
@@ -12,8 +12,13 @@
 
 #include "../mwworld/esmstore.hpp"
 
+#include "../../../components/misc/rng.hpp"
```
Capostrophic: Why not include it like ``<components/misc/rng.hpp>`` like anything else does?

files/mygui/openmw_chargen_create_class.layout
```diff
@@ -76,6 +76,9 @@
         <!-- Dialog buttons -->
         <Widget type="HBox" position="0 158 482 24">
             <Widget type="Spacer"/>
+            <Widget type="AutoSizedButton" skin="MW_Button" name="RandomButton">
+                <Property key="Caption" value="Random"/>
```
Capostrophic: Non-localizable strings are not good.

apps/openmw/mwgui/class.cpp
wareya: Oh, I didn't know the include paths were like that. I'll do that.

files/mygui/openmw_chargen_create_class.layout
wareya: Where are openmw-specific localizable strings stored?

akortunov: Nowhere. Zini said that custom localization is only post-1.0.

zinnschlag:
Not in favour of this change (pre 1.0). That's just more stuff we have to fix up once the de-hardcoding is under way. But a neat idea for post-1.0.

zinnschlag:
I am going to close this PR. By the time we are ready to merge, OpenMW will have changed so much that most of the code will have to be re-written anyway.

PR #1783, opened 2018-06-27
Altered sentence under packaging

LordInsane:
Altered a sentence to try to keep the issue clear while avoiding seeming as if any OS is denigrated.


zinnschlag:
I have restrained myself from doing any stabs against Windows throughout a 25k word document. That took some serious effort. Allow me this (very moderate) one, please.

PR #1784, opened 2018-06-27
Revert "Retrieve SDL window settings instead of using magic numbers"

nikolaykasyanov:
Reverts #1771, fixing crash to desktop on macOS.

Thanks @kcat for the quote I put instead of FIXME.


psi29a:
Thanks!

PR #1786, opened 2018-06-28
[WIP] Update before_script.msvc.sh

psi29a:
This has working GL Win10 MSVC updates, should be cross-compatible with appveyor.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -419,12 +419,12 @@ fi
 		BOOST_SDK="$(real_pwd)/Boost"
 		CWD_DRIVE_ROOT=`powershell -command '(get-location).Drive.Root'`  # get the current working drive's root
 
-		if [ -d Boost ] && grep "BOOST_VERSION 106100" Boost/boost/version.hpp > /dev/null; then
+		if [ -d Boost ] && grep "BOOST_VERSION 106700" Boost/boost/version.hpp > /dev/null; then
 			printf "Exists. "
 		elif [ -z $SKIP_EXTRACT ]; then
 			rm -rf Boost
-            "${DEPS}/boost-1.67.0-msvc${MSVC_YEAR}-win${BITS}.exe" //DIR="${CWD_DRIVE_ROOT}Boost_temp" //VERYSILENT //NORESTART //SUPPRESSMSGBOXES //LOG="boost_install.log"			
-			mv "${CWD_DRIVE_ROOT}Boost_temp" ${BOOST_SDK}
+			"${DEPS}/boost-1.67.0-msvc${MSVC_YEAR}-win${BITS}.exe" //DIR="${CWD_DRIVE_ROOT}Boost_temp" //VERYSILENT //NORESTART //SUPPRESSMSGBOXES //LOG="boost_install.log"			
+			mv "${CWD_DRIVE_ROOT}Boost_temp" "${BOOST_SDK}"
 
 		fi
 
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -x
 
 MISSINGTOOLS=0
 
@@ -235,7 +236,7 @@ case $VS_VERSION in
 		TOOLSET="vc140"
 		TOOLSET_REAL="vc141"
 		MSVC_REAL_VER="15"
-		MSVC_VER="14"
+		MSVC_VER="14.1"
 		MSVC_YEAR="2015"
 		MSVC_DISPLAY_YEAR="2017"
 		;;
@@ -245,7 +246,7 @@ case $VS_VERSION in
 		TOOLSET="vc140"
 		TOOLSET_REAL="vc140"
 		MSVC_REAL_VER="14"
-		MSVC_VER="14"
+		MSVC_VER="14.0"
 		MSVC_YEAR="2015"
 		MSVC_DISPLAY_YEAR="2015"
 		;;
@@ -255,7 +256,7 @@ case $VS_VERSION in
 		TOOLSET="vc120"
 		TOOLSET_REAL="vc120"
 		MSVC_REAL_VER="12"
-		MSVC_VER="12"
+		MSVC_VER="12.0"
 		MSVC_YEAR="2013"
 		MSVC_DISPLAY_YEAR="2013"
 		;;
@@ -325,9 +326,9 @@ if [ -z $SKIP_DOWNLOAD ]; then
 
 	# Boost
 	if [ -z $APPVEYOR ]; then
-		download "Boost 1.61.0" \
-			"https://sourceforge.net/projects/boost/files/boost-binaries/1.61.0/boost_1_61_0-msvc-${MSVC_VER}.0-${BITS}.exe" \
-			"boost-1.61.0-msvc${MSVC_YEAR}-win${BITS}.exe"
+		download "Boost 1.67.0" \
+			"https://sourceforge.net/projects/boost/files/boost-binaries/1.67.0/boost_1_67_0-msvc-${MSVC_VER}-${BITS}.exe" \
+			"boost-1.67.0-msvc${MSVC_YEAR}-win${BITS}.exe"
 	fi
 
 	# Bullet
@@ -365,8 +366,8 @@ if [ -z $SKIP_DOWNLOAD ]; then
 			QT_SUFFIX=""
 		fi
 
-		download "Qt 5.7.2" \
-			"https://download.qt.io/official_releases/qt/5.7/5.7.0/qt-opensource-windows-x86-msvc${MSVC_YEAR}${QT_SUFFIX}-5.7.0.exe" \
+		download "Qt 5.7.0" \
+			"https://download.qt.io/archive/qt/5.7/5.7.0/qt-opensource-windows-x86-msvc${MSVC_YEAR}${QT_SUFFIX}-5.7.0.exe" \
 			"qt-5.7.0-msvc${MSVC_YEAR}-win${BITS}.exe" \
 			"https://www.lysator.liu.se/~ace/OpenMW/deps/qt-5-install.qs" \
 			"qt-5-install.qs"
@@ -403,7 +404,7 @@ echo
 
 # Boost
 if [ -z $APPVEYOR ]; then
-	printf "Boost 1.61.0... "
+	printf "Boost 1.67.0... "
 else
 	if [ $MSVC_VER -eq 12 ]; then
 		printf "Boost 1.58.0 AppVeyor... "
@@ -421,11 +422,12 @@ fi
 			printf "Exists. "
 		elif [ -z $SKIP_EXTRACT ]; then
 			rm -rf Boost
-			"${DEPS}/boost-1.61.0-msvc${MSVC_YEAR}-win${BITS}.exe" //dir="$(echo $BOOST_SDK | sed s,/,\\\\,g)" //verysilent
+			"${DEPS}/boost-1.61.0-msvc${MSVC_YEAR}-win${BITS}.exe" //DIR="${SYSTEMDRIVE}\boost" //VERYSILENT //NORESTART //SUPPRESSMSGBOXES //LOG="boost_install.log"			
+			mv "${SYSTEMDRIVE}\boost" ${BOOST_SDK}
 		fi
 
 		add_cmake_opts -DBOOST_ROOT="$BOOST_SDK" \
-			-DBOOST_LIBRARYDIR="${BOOST_SDK}/lib${BITS}-msvc-${MSVC_VER}.0"
+			-DBOOST_LIBRARYDIR="${BOOST_SDK}/lib${BITS}-msvc-${MSVC_VER}"
 		add_cmake_opts -DBoost_COMPILER="-${TOOLSET}"
 
 		echo Done.
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -422,7 +422,7 @@ fi
 			printf "Exists. "
 		elif [ -z $SKIP_EXTRACT ]; then
 			rm -rf Boost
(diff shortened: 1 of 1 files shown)


psi29a:
Does it blend...

psi29a:
@ananace  do I have your blessing here?

ananace:
I'd suggest at least extracting the drive the script is running on instead of the system drive if you really can't extract to the current folder.

If you hardcode the script to dump data to the system drive then it's going to break for lots of people, as the system drive on modern Windowses has *really* strict access controls.

psi29a:
I tried TEMP and TMP, but it just dumped to C:\local when given to the boost exe for extraction, however bash takes TEMP/TMP to mean your user temp directory... that's not right.  If that had worked, would you have been OK with that?

Do you have another environment variable I can use?

psi29a:
Going to try the power-shell trick to get the current working drive. I assume that this drive will have enough space, at least 3GiB free.

The real issue is the innoinstaller

psi29a:
GL is happy:
https://gitlab.com/psi29a/openmw/-/jobs/78281749

CI/before_script.msvc.sh
```diff
@@ -415,17 +413,23 @@ fi
...
 		BOOST_SDK="$(real_pwd)/Boost"
+		CWD_DRIVE_ROOT=`powershell -command '(get-location).Drive.Root'`  # get the current working drive's root
 
 		if [ -d Boost ] && grep "BOOST_VERSION 106100" Boost/boost/version.hpp > /dev/null; then
 			printf "Exists. "
 		elif [ -z $SKIP_EXTRACT ]; then
 			rm -rf Boost
-			"${DEPS}/boost-1.61.0-msvc${MSVC_YEAR}-win${BITS}.exe" //dir="$(echo $BOOST_SDK | sed s,/,\\\\,g)" //verysilent
+            "${DEPS}/boost-1.67.0-msvc${MSVC_YEAR}-win${BITS}.exe" //DIR="${CWD_DRIVE_ROOT}Boost_temp" //VERYSILENT //NORESTART //SUPPRESSMSGBOXES //LOG="boost_install.log"			
+			mv "${CWD_DRIVE_ROOT}Boost_temp" ${BOOST_SDK}
```
ananace: You might still want to have quotes around `${BOOST_SDK}`, don't want to break on paths with spaces in it.

psi29a:
AV is also happy...  you guys happy?

@AnyOldName3 and @ananace

CI/before_script.msvc.sh
```diff
@@ -415,17 +413,23 @@ fi
 	if [ -z $APPVEYOR ]; then
 		cd $DEPS_INSTALL
 
+        # Boost's installer is still based on ms-dos API that doesn't support larger than 260 char path names
+        # We work around this by installing to root of the current working drive and then move it to our deps
+        
 		BOOST_SDK="$(real_pwd)/Boost"
+		CWD_DRIVE_ROOT=`powershell -command '(get-location).Drive.Root'`  # get the current working drive's root
 
 		if [ -d Boost ] && grep "BOOST_VERSION 106100" Boost/boost/version.hpp > /dev/null; then
```
ananace: You'll need to update the value in this check, otherwise it's going to replace the boost copy every time the script is run, even if it's already current.

psi29a:
AV/GI are both happy...  
@ananace another round? Good to go? :)

CI/before_script.msvc.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -x
```
ananace: Is this supposed to still be here?  
Perhaps it could at least be a `[ -n "$CI" ] && set -x` in that case?

We probably don't want to spam people with incomprehensible data when they just want to get a development environment set-up after all.

CI/before_script.msvc.sh
```diff
@@ -415,17 +413,23 @@ fi
 	if [ -z $APPVEYOR ]; then
 		cd $DEPS_INSTALL
 
+        # Boost's installer is still based on ms-dos API that doesn't support larger than 260 char path names
+        # We work around this by installing to root of the current working drive and then move it to our deps
```
ananace: Nit; Indent this one step to be on the same line as the code.

CI/before_script.msvc.sh
```diff
@@ -415,17 +413,23 @@ fi
...
+        
 		BOOST_SDK="$(real_pwd)/Boost"
+		CWD_DRIVE_ROOT=`powershell -command '(get-location).Drive.Root'`  # get the current working drive's root
```
ananace: Nit; Use `$()` instead of ` `` `, to stay consistent

CI/before_script.msvc.sh
```diff
@@ -415,17 +413,23 @@ fi
...
 
-		if [ -d Boost ] && grep "BOOST_VERSION 106100" Boost/boost/version.hpp > /dev/null; then
+		if [ -d Boost ] && grep "BOOST_VERSION 106700" Boost/boost/version.hpp > /dev/null; then
 			printf "Exists. "
 		elif [ -z $SKIP_EXTRACT ]; then
 			rm -rf Boost
-			"${DEPS}/boost-1.61.0-msvc${MSVC_YEAR}-win${BITS}.exe" //dir="$(echo $BOOST_SDK | sed s,/,\\\\,g)" //verysilent
+			"${DEPS}/boost-1.67.0-msvc${MSVC_YEAR}-win${BITS}.exe" //DIR="${CWD_DRIVE_ROOT}Boost_temp" //VERYSILENT //NORESTART //SUPPRESSMSGBOXES //LOG="boost_install.log"
```
ananace: Perhaps the `//VERYSILENT //SUPPRESSMSGBOXES` could be set only when `[ -n "$CI" ]` as well? That way they shouldn't hide errors or warnings from non-CI users of the script.

psi29a: The whole thing doesn't work (as a script) if we don't use //VERYSILENT, so that has to stay. The rest we can hide behind a switch.

psi29a:
AV is almost happy... 2015 is good, 2017 is building.  GL is also happy.

@ananace  Are all then nits picked? :)

ananace:
I'll have another look as soon as I'm back home 🙂

CI/before_script.msvc.sh
AnyOldName3: @ananace We can't support spaces whatever we do because Qt refuses to go anywhere with spaces.

ananace:
So maybe we should have a check that aborts with a very visual warning
about that fact before coming to the dependency work?

AnyOldName3:
I still have some nitpicking to do.

* We should check to see if the boost temporary directory exists before installing Boost to it. Just aborting and printing an error message is probably enough, but I'd still feel happier if there was some kind of check.
* There are some places where backslashes could sneak in, for example where PowerShell is used. They happen not to be breaking anything right now, but it's safer to change them to forward slashes. I think there's some code in the script that uses `sed` to sort this kind of thing out.

AnyOldName3:
@ananace is
> So maybe we should have a check that aborts with a very visual warning about that fact before coming to the dependency work?
a reply to
> @ananace We can't support spaces whatever we do because Qt refuses to go anywhere with spaces.

If so, there's already an error message. I added it a few weeks ago when there was a big hoohah about the script being unusable.

ananace:
Indeed it was @AnyOldName3, response methods are limited when on a phone with email.

And if there is an error message then of course we don't have *another* one, though I still consider quoting the string the right thing to do, it makes life so much saner.

CI/before_script.msvc.sh
```diff
@@ -415,17 +413,24 @@ fi
 	if [ -z $APPVEYOR ]; then
 		cd $DEPS_INSTALL
 
+		# Boost's installer is still based on ms-dos API that doesn't support larger than 260 char path names
+		# We work around this by installing to root of the current working drive and then move it to our deps
+        
 		BOOST_SDK="$(real_pwd)/Boost"
+		CWD_DRIVE_ROOT=$(powershell -command '(get-location).Drive.Root')  # get the current working drive's root
```
ananace: Nit; I'd love to see quotes around this too, even if there's no way that a drive root could contain a space today.

AnyOldName3: Personally, I think quotes **and** running it through `sed` to turn the backslash into a forward slash is the best option.

AnyOldName3:
I still think we should `sed`ify the backslash into a forward slash.

psi29a:
We can do a wider scope cleanup via another PR if you wish?

AnyOldName3:
I've had a more detailed look, and it seems that the PowerShell bit is the only place a backslash could sneak in. As that's been added here, it should probably be fixed here. I think that this line https://github.com/OpenMW/openmw/blob/2bf0d598cfb138dc5d7c234d95ce4a19b2c64901/CI/before_script.msvc.sh#L119 can be copied and applied to the output from PowerShell and then everything's fine and we're not going to encounter any surprises.

psi29a:
> Boost 1.67.0... mv: cannot stat '/C/Boost_temp': No such file or directory

mv doesn't like that

AnyOldName3:
I'd try something based on

```bash
CWD_DRIVE_ROOT=$(powershell -command '(get-location).Drive.Root' | sed "s,\\\\,/,g")
```

which seems to do everything we actually want except put the result in quotes. I don't think we want the result in quotes, though, as when we use `CWD_DRIVE_ROOT`, we're enclosing it in quotes.

The problem with the previous version is that the final `sed` was changing `C:/` into `/C/` which the Boost installer was interpreting as `./C/` and `mv` was interpreting as the C drive. Also we were `echo`ing the output of a command, which is valid, but it's also silly.

psi29a:
mv doesn't like that either:

> Boost 1.67.0... mv: cannot stat 'C:/Boost_temp': No such file or directory

It has to be C:\Boost_temp, or mv complains.

AnyOldName3:
It's not `mv` that's getting upset by it, as `mv` expects Unix-style paths. It's Boost (which is probably less happy with them, but should work if they're not too Unixy). Do you know where Boost's been installing itself the times `mv` has failed?

My concern is that at some point we might try changing `"${CWD_DRIVE_ROOT}Boost_temp"` into something where the backslash could do something bad, such as replacing `Boost_temp` with a variable which doesn't end up getting expanded because its `$` ends up right after a backslash, or we end up passing `CWD_DRIVE_ROOT` through another processing stage that gets upset by backslashes. If that's not something that could happen anyway, it's not a major concern and you can probably skip trying to fix it, but my instincts tell me not to trust backslashes in Bash scripts to be maintainable.

One further note is that we're still not double checking to see if `"${CWD_DRIVE_ROOT}Boost_temp"`  already exists and displaying an error before we overwrite it. It's conceivable that such a folder might already exist and as it's not within the OpenMW directory, it's likely users wouldn't expect us to meddle with it.

psi29a:
@AnyOldName3 I've seen this before but can't explain it, if I use: 
```
CWD_DRIVE_ROOT=$(echo "$(powershell -command '(get-location).Drive.Root')" | sed "s,\\\\,/,g" | sed "s,\(.\):,/\\1,") 
```
and then:
```
"${DEPS}/boost-1.67.0-msvc${MSVC_YEAR}-win${BITS}.exe" //DIR="${CWD_DRIVE_ROOT}Boost_temp" //VERYSILENT //NORESTART
```

innoinstaller then dumps it to C:\local\boost_1_67_0

This is why mv can't find:
```
Boost 1.67.0... mv: cannot stat 'C:/Boost_temp': No such file or directory
```

AnyOldName3:
What was the install location with the version I suggested with just `sed "s,\\\\,/,g"`? I wouldn't have expected the one you posted to work (as we're giving Boost something that definitely isn't a Windows-style path), but I'm surprised my version didn't (as it should be a Windows-style path with `/` separators, which Windows is supposed to support).

psi29a:
```
$(echo "$(powershell -command '(get-location).Drive.Root')")
C:\
```

```
$(echo "$(powershell -command '(get-location).Drive.Root')" | sed "s,\\\\,/,g" | sed "s,\(.\):,/\\1,") 
/C/
```

I understand how and why it works, but innoinstaller apparently doesn't like the pathname /C/ and then defaults to `C:\local\boost_1_67_0`, my guess is that it can't handle that path name.

What I can do is:
```
CWD_DRIVE_ROOT=$(echo "$(powershell -command '(get-location).Drive.Root')boost_temp")
```
pass this to innoinstaller,

then before passing it to mv, we run it through it through sed?

Is this acceptable?

psi29a:
My assumption turned out correct, innoinstaller cannot handle `/` but only `\`, as a result, it dumps the contents of boost into `C:\local\boost_1_67_0`.

Going to try it another way that lets `\` through for the innoinstaller.

psi29a:
This works:

```
		CWD_DRIVE_ROOT="$(powershell -command '(get-location).Drive.Root')Boost_temp"
		CWD_DRIVE_ROOT_BASH=$(echo "$CWD_DRIVE_ROOT" | sed "s,\\\\,/,g" | sed "s,\(.\):,/\\1,")
		
			"${DEPS}/boost-1.67.0-msvc${MSVC_YEAR}-win${BITS}.exe" //DIR="${CWD_DRIVE_ROOT}" //VERYSILENT //NORESTART ${CI_EXTRA_INNO_OPTIONS}
			mv "${CWD_DRIVE_ROOT_BASH}" "${BOOST_SDK}"
```

Is this acceptable?

AnyOldName3:
I am very surprised that `C:/` didn't work, and I think it means that the installer is doing its own path separation and combination instead of using Windows or std functions to do it. The separate paths for Boost and `mv` approach seems to be the best one, given that Boost is being weird.

All we need now is to double check that `CWD_DRIVE_ROOT_BASH` doesn't exist before we do anything with it, and spit out an error message if it does, then this should be merged.

psi29a:
https://gitlab.com/psi29a/openmw/-/jobs/79089099  <-- CI is happy

psi29a:
@AnyOldName3 are the changes acceptable?

AnyOldName3:
Yes, I think so.

AnyOldName3:
In fact, I'm just going to merge this as someone's messaging me about the Qt link change again and it's easier to fix it than explain.

psi29a:
OK... clicking merge... no one shoot me! :)

psi29a:
Thanks!

PR #1787, opened 2018-06-28
Use the DejaVuLGCSansMono.ttf for profiler output

akortunov:
For now I get a lot of such errors in-game:
`Error reading file fonts/arial.ttf: file not found`

That's because osgViewer::StatsHandler (which we use as profilier) uses hardcoded arial.ttf from "fonts" folder.
But we do not provide arial.ttf with OpenMW, and an engine uses some kind of default font instead of TTF one.

The main idea: derive from osgViewer::StatsHandler and override used font - we can use the DejaVuLGCSansMono.ttf since we already provide it with OpenMW.
If the font was not found for some reason, OpenMW will use the internal OSG font as fallback.
This PR affects both profilier (F3) and stats monitor (F4).

Before merge, it would be nice to test this PR on Windows and Mac with both osg 3.4 and 3.6.
Example of output is [here](https://i.imgur.com/L1iZI0L.jpg).

### DIFF
--- a/components/resource/stats.cpp
+++ b/components/resource/stats.cpp
@@ -10,6 +10,8 @@
 #include <osgViewer/Viewer>
 #include <osgViewer/Renderer>
 
+#include <components/myguiplatform/myguidatamanager.hpp>
+
 namespace Resource
 {
 
@@ -19,7 +21,7 @@ StatsHandler::StatsHandler():
     _statsType(false),
     _statsWidth(1280.0f),
     _statsHeight(1024.0f),
-    _font("fonts/arial.ttf"),
+    _font(""),
     _characterSize(20.0f)
 {
     _camera = new osg::Camera;
@@ -28,6 +30,15 @@ StatsHandler::StatsHandler():
     _camera->setProjectionResizePolicy(osg::Camera::FIXED);
 
     _resourceStatsChildNum = 0;
+
+    _font = osgMyGUI::DataManager::getInstance().getDataPath("DejaVuLGCSansMono.ttf");
+}
+
+Profiler::Profiler()
+{
+    _font = osgMyGUI::DataManager::getInstance().getDataPath("DejaVuLGCSansMono.ttf");
+
+    setKeyEventTogglesOnScreenStats(osgGA::GUIEventAdapter::KEY_F3);
 }
 
 bool StatsHandler::handle(const osgGA::GUIEventAdapter &ea, osgGA::GUIActionAdapter &aa)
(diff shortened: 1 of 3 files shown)


akortunov:
After some investigation I found that OSG searches for fonts in these directories:
. (directory of OpenMW binary in our case)
/usr/share/fonts/ttf
/usr/share/fonts/ttf/western
/usr/share/fonts/ttf/decoratives

On my system, for example, arial.ttf is located in the /usr/share/fonts/corefonts directory.
I wonder why OSG does not check all /usr/share/fonts/* directories since font location depends on Linux distribution.

akortunov:
I re-purposed this PR.
Now I search for arial.ttf in the system font folders.

components/resource/stats.cpp
```diff
@@ -10,16 +14,67 @@
...
+    }
+    return false;
+}
+
+std::string getFont()
+{
+    std::string fontName = "arial.ttf";
+
+#if defined(_WIN32)
+    boost::filesystem::path fontPath = "C:/windows/fonts";
```
Mingun: May be use "%WINDOWS%/Fonts" (and code for expanding variables)? Users which installing system to non-standard locations won't be upset

akortunov: Maybe a good idea, but I do not use Windows and can not test if it will work or not.
Does boost::filesystem::path understand OS-specific environmental variables?

akortunov: Also I doubt that many people will complain - "C:/windows/fonts" is harcoded in OSG code and we used it for years.

Capostrophic: You can use SHGetFolderPathW with CSIDL_FONTS set as CSIDL for determining the path to user fonts folder on Windows, like windows-specific paths are [already](https://github.com/OpenMW/openmw/blob/master/components/files/windowspath.cpp) determined.

akortunov: I guess I'll leave it for Windows developers.

akortunov:
How about option to store font name in config? Then player can choose font to better fit his taste.

akortunov:
I added setting to specify profilier font.
As I said, we could just use the DejaVuLGCSansMono.ttf, but in this case we need to determine MyGUI resources location on the fly somehow.

psi29a:
Ready? :)

akortunov:
> Ready? :)

Should be ready, but I did not test it on all supported operation systems.

MiroslavR:
It's spelled _profiler_, not _profilier_...

akortunov:
> It's spelled profiler, not profilier...

Fixed.

psi29a:
Merging this to get more coverage from users.

PR #1788, opened 2018-06-29
Make unarmed creature attacks not affect armor condition (bug #2455)

Capostrophic:
[Bug 2455](https://gitlab.com/OpenMW/openmw/issues/2455).

Move the armor degradation code behind a check that the attacker is not an unarmed creature.
Initially I tried to make a neat reduceArmorCondition function with more sanity checks to get rid of code duplication but I ran into issues due to shield in combat.cpp being an iterator.


psi29a:
I like. :)   Thanks.

PR #1790, opened 2018-07-02
Migrating necessary parts of Wiki over to RTD.

rhtucker:


rhtucker:
Looking through the wiki, I don't really think there is much left that should be part of our official docs.

I've already migrated:
- Mod installation
- Settings
- Fonts
- and now Installation Instructions

loriel is working on the OpenMW-CS stuff

All I can think of that needs to be moved to RTD is the texture related stuff.

psi29a:
Are you cleaning up the old wiki too (removing migrated pages)?

rhtucker:
Not currently, but creating a forum post to discuss the migration as a whole momentarily. I don't want to spend too much time tidying up the existing wiki if we're trying to move the whole thing to GitLab.

rhtucker:
Sure, I'll go ahead and add links like we've done in the past, but beyond that I'll probably leave it alone. Do you think anything else is necessary? Should get the texture stuff done tonight which should allow this to get merged tomorrow. I'll add the links once they show up on RTD.

psi29a:
RTD will install the version of sphinx and other pythons libs from the requirements.txt files.
If I read the documentation correctly, the extension is: `sphinx.ext.autosectionlabel`
You then need to set autosectionlabel_prefix_document to True:
```
autosectionlabel_prefix_document = True
```
And it should work.

Likely to be done at the top of each document.

rhtucker:
> RTD will install the version of sphinx and other pythons libs from the requirements.txt files.
>If I read the documentation correctly, the extension is: `sphinx.ext.autosectionlabel`

Tried those and been digging for a while so posting a question on Stack Overflow.

loriel2:
I assume it is sensible to keep the Mod Compatibility pages on the wiki for a while longer.

If it's not already covered in the discussion above, I think GitLab needs adding as an alternative to GitHub - both in the "Edit on Git***" button(s) at top right, and in the recently added "So you want to help with documentation?" - I didn't see any obvious was to link to GitLab when I installed Pycharm, as I recall.

psi29a:
Look for plugins:
https://plugins.jetbrains.com/plugin/7447-gitlab-integration-plugin
https://plugins.jetbrains.com/plugin/7975-gitlab-projects

rhtucker:
> I assume it is sensible to keep the Mod Compatibility pages on the wiki for a while longer.

Actually I think the plan is to always keep that on the wiki. Reason being that it’s something the community at large should be contributing regularly to, so we want to keep the barrier of entry as low as possible. Plus the status may change at any time, so it’s just too difficult for their compatibility to be added to our “official” documentation.

In terms of PyCharm having GitLab support, it doesn’t come with a built-in plugin, but as psi29a sent above, there’s a few to choose from. I’m personally ok with the doc guide only having one, because this is just meant to get people started and comfortable with the barebones of contributing. However, if I got to choose, I’d rather it be GitLab because that’s where the wiki will be and it just seems to be our unofficially official main repo. If I’m doing the work to add GitLab though, we might as well keep GitHub in the doc.

If you guys figure out how to get the plugin to work, I would appreciate some help, because I was struggling to figure out the integration. Perhaps that would be better done on the forum though so other poor fools can benefit from it.

loriel2:
Agreed on the Mod Compatibility, for the foreseeable future. Possibly convert/transfer when activity drops to a low level.

I am working through adding the plugins, which doesn't seem too difficult but the wording of the accompanying instructions is a bit vague - specifically for the "GitLab Integration Plugin", the instruction "go to IDE preferences" actually means go to "File/Settings/Other Settings/GitLab Integration".

I am proposing putting in a short PR sometime soon to add GitLab as an alternative to GitHub in the docs - which I think is what you suggest above.

Alternatively we could open a Documentation Forum thread to discuss it, as you suggest.

Thunderforge:
For what it's worth, I think that the Mod Compatibility wiki pages have a number of drawbacks, and I think that we could probably come up with better solutions. I've [raised a forum topic](https://forum.openmw.org/viewtopic.php?f=43&t=5270) to discuss alternatives.

psi29a:
@Thunderforge the discussion needs to go on further in that thread, let's keep this PR purely on topic. :)

Is this PR still WIP?

Capostrophic:
>Is this PR still WIP?

I suppose it is, as I only see a single removed EOF line.

psi29a:
@rhtucker was the intent to hold a discussion here? Would it not be better in the forums?

rhtucker:
Sorry, away from my computer with everything on it so I'll try and respond as best as possible here. Should be back in town tomorrow evening to finish up the transfer. I don't _need_ the prefix_document config that we were talking about above, so I'm just tabling that until I get it figured out, but eventually that will be super handy.

@psi29a No, I don't think we should have any involved discussions on here, it would be better in the forum. But yes, still WIP.

@Thunderforge I'll try and respond soon to the thread. I have many thoughts on the subject lol.

>I am proposing putting in a short PR sometime soon to add GitLab as an alternative to GitHub in the docs - which I think is what you suggest above.

@loriel2 Yes! That would be great. Keep in mind this is for totally non-technical people so it will have to be as step-by-step as possible.

>I suppose it is, as I only see a single removed EOF line.

Hmm, I had hoped the submit PR button on PyCharm would combine all the necessary commit, push, PR actions into one, but obviously not! Never fear, I'm almost done, just got hung up on the autosection feature with Sphinx.

loriel2:
GitLab integration for Pycharm - see GitLab Merge Request 20 at https://gitlab.com/OpenMW/openmw/merge_requests/20. Zini has commented already, let's continue the discussion there.

I intend another one "sometime soon" to document the simpler system for contributions to documentation outlined by Psi29a on the Thunderforge's forum topic.

rhtucker:
So I finished the texture section, but looks like I accidentally let PyCharm merge rather than rebase. Is there something I can do to fix all that?

psi29a:
Yeah, just rebase. ;) It will clean up your history.

rhtucker:
Hooray! Finally figured the damn thing out! For future reference, this is a pretty good guide: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

PR #1791, opened 2018-07-02
Editor: limit FPS in 3D preview windows

akortunov:
Implements [feature #3641](https://gitlab.com/OpenMW/openmw/issues/3641).

Note: I am aware of threading model - I can not test it because I do not have Qt4.
In theory, CompositeViewer::update() still should work in the main thread, but rendering update itself inside frame() call should be multithreaded.


zinnschlag:
Seems to work just fine on Qt4.

PR #1792, opened 2018-07-04
 Make spellcasting stance transition more smooth (bug #4358)

Capostrophic:
[Bug 4358](https://gitlab.com/OpenMW/openmw/issues/4358)

Somewhat "experimental" fix I made accidentally: if a movement animation was identical to the previous one that was played, restart it from the point the previous animation ended (get the point of completion of the previous animation), instead of cutting it abrupt and restarting from loop start. Seems to work fine from my testing..? Feedback would be appreciated. @akortunov, does this seem OK to you?

It bothers me that I don't account for difference between start and loop start text keys, but when I actually try to do that, I break it again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
     Bug #4293: Faction members are not aware of faction ownerships in barter
     Bug #4307: World cleanup should remove dead bodies only if death animation is finished
     Bug #4327: Missing animations during spell/weapon stance switching
+    Bug #4358: Running animation is interrupted when magic mode is toggled
     Bug #4368: Settings window ok button doesn't have key focus by default
     Bug #4393: NPCs walk back to where they were after using ResetActors
     Bug #4416: Handle exception if we try to play non-music file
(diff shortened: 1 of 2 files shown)


akortunov:
Looks good to me, but I did not test it yet.

> It bothers me that I don't account for difference between start and loop start text keys, but when I actually try to do that, I break it again.

What did you actually try to do?

psi29a:
Would be great if you could 'film/gif/apng' the video of before and after changes in the future. I understand if it's too much work though.

Capostrophic:
>What did you actually try to do?
A thing I tried to do is to make start point dependent on the actual time between loop start and loop stop... wait, I just realized why it didn't work properly.
>Would be great if you could 'film/gif/apng' the video of before and after changes in the future.
OK, I'll try to record something.

Capostrophic:
Two videos. Imgur "optimized" them to hell and back, but the difference should be visible.
[Master](https://i.imgur.com/oPPUqTf.mp4)
[PR behavior](https://i.imgur.com/ElYwAfQ.mp4)

psi29a:
That's well a great improvement.

akortunov:
As alternative solution, we could try to do not stop movement animation at all, if a new one is the same as the old one.

PR #1793, opened 2018-07-05
Ensure forward-compatibility of death animations of pre-0.43.0 saves (bug #4274)

Capostrophic:
[Bug 4274](https://gitlab.com/OpenMW/openmw/issues/4274)
Better late than never.
Add a check that the character is actually swimming before applying (or fast-forwarding, as in the case of the report) two swimming death animations added in 0.43.0.


psi29a:
Thanks!

PR #1794, opened 2018-07-06
Movement solver changes/fixes

wareya:
This fixes several major problems with the current movement solver, which make it a lot harder to get stuck and much less common that you can't jump in certain locations unless you stand still.

No changelog additions yet and not ready to merge. Some of these changes can cause performance problems and bring some other bugs to attention (like capsules breaking some vanilla level design), so it shouldn't get merged until enough people look at it.

If you look at this, though, please test it, too, and compare its behavior ingame directly to current master. The problems it fixes are monumental.

This is certainly still not a perfect movement solver, but it's a lot better than what OpenMW has right now.

Problems fixed:

- Not mapping against what's hit unless the stair code successfully runs
- Lack of safety margin making it unsafe to do the above
- Problem with how "don't skid up steep slopes" worked, preventing jumping while running into walls
- Some bad interactions between jumping and stairstepping
- Getting stuck in simple acute crevices, including some hanging sloped walls and things like the double stairs in the ald'ruhn mage's guild (complex acute crevices still get stuck)
- The stair code freaking out in obscure cases involving steep slopes

Test cases:

- The bottom of the stair meshes in interior vivec areas (test these with cylinders and rounded cylinders too, see actor.cpp)
- The beds in the ald'ruhn mage's guild (their sides are steep slopes)
- The bedside tables in the ald'ruhn mage's guild (their sides are VERY steep slopes)
- The hanging sloped walls in the ald'ruhn mage's guild
- The big plant pot thing near the entrance of the ald'ruhn mage's guild
- Most walls anywhere
- The double stairs in the ald'ruhn mage's guild (falling between the railings, not between them where you could always fit and get stuck) (note: in vanilla you can just walk between them for some reason)
- Some of the pillars in the ald'ruhn mage's guild walls, where they intersect with the wall at an obtuse angle (as far as player collision is concerned)
- A place in the ald'ruhn mage's guild (near a merchant) where two tall pots meet eachother at an obtuse angle (as far as player collision is concerned) (try jumping)

Things that might need adjustment:

- ~~~There's a hack that ensures the stairs code works correctly when the player is jumping and moving upwards, but maybe the stairs code shouldn't work in this case at all.~~~
- There's a hack that makes slopes act like walls in some cases, to keep the character from stuttering and unable to jump when like walking into the sides of beds and tables in ald'ruhn, or some rocks in the wilderness. It doesn't help in some cases.

Related forum thread (linked for bookkeeping purposes): https://forum.openmw.org/viewtopic.php?f=6&t=5200


psi29a:
For those wanting to test along with you, can you provide save games to the places that should be tested?

wareya:
Figured it would be easy to coc into those areas, but here.

[saves.zip](https://github.com/OpenMW/openmw/files/2171204/saves.zip)

wareya:
Is there anything that can be done about the random appveyor build failures?

psi29a:
Yeah, remove msvc2015 since it only happens with that and save us 25 build wait in the process.

ananace:
I'd personally prefer not dropping MSVC2015, keeping support for the two latest versions is a good policy to have. But since it seems to be an AppVeyor logger issue - and therefore 100% out of our control - I'm not sure what we could do to avoid it.

psi29a:
Can we set it to allow-failure (similar to how we do it for clang on linux)?

AnyOldName3:
Has the bug been reported to AppVeyor?

wareya:
Oops, I can't leave the ground while levitating. Guess I broke one of the flying checks.

wareya:
Fixed.

wareya:
About this hack:

>There's a hack that ensures the stairs code works correctly when the player is jumping and moving upwards, but maybe the stairs code shouldn't work in this case at all.

Vanilla does something similar, so this should stay in. My worry was that it might not, since some games don't.

wareya:
Alright, I can't see any more major changes needed based on my own testing. Anyone want to try to break it ingame before I change the "request for comments" to "ready for review"?

GIuka:
What is the benefit to snapping to surfaces when the player is moving upwards? An issue this creates is that the player can jump against a rope fence with ~300 acrobatics and then immediately snap onto it, losing all momentum. Sorry if this has an obvious answer.

Players with high acrobatics can also jump beneath awnings and other overhead geometry and will maintain all upward momentum. This can cause all sorts of weirdness. I know this has been an issue in master for years so this might not be entirely relevant to your PR but it might be worth looking into while you're working in this area.

wareya:
The reason is that vanilla Morrowind does the same thing. That said, removing it is only a few lines of changes.

akortunov:
Is there any way to make tracer do not touch actors with side collisions? [This](https://i.imgur.com/u7qHayp.png) situation make physics crazy - it tries to move the actor up and down (what obviously does not help), causing a jittering. With cylinders situation is better a bit, but still is far from ideal.
This bug is easy to reproduce with Darknut's Dwemer Dwelling mod:
1. coc "Kazara, Living Quarters".
2. Mannequins on the second floor collide with wall behind them.

wareya:
They're colliding with the wall because they're overlapping it, right? Are they moving?

akortunov:
> They're colliding with the wall because they're overlapping it, right? Are they moving?

Overlapping, are not moving.
Also I encountered a similar issue with Arktwend - some traders overlap furniture a bit. Technically, they move due to their idle animations.

wareya:
I think the mod is constantly placing them slightly inside the world, then they're sliding outside of the world. I can't figure out what code is making them slide outside of the world, though.

akortunov:
> I can't figure out what part of the code is making them slide outside of the world, though.

It should be somewhere inside PhysicsSystem::move().
Looks like movement solver tries to move actor up to get him out of obstacle.

wareya:
I fixed the vibrating, but is this one supposed to be floating?
screenshot233

wareya:
I think that because they're in the wall, it breaks the slope check in the ground-finding code, which allows gravity to be applied. This makes it so they have inertia, so checking whether they have inertia before doing anything movement related doesn't work.

wareya:
Do you think anyone would complain if, like, we just made the actor collision cylinder just a little thinner? It would fix the problem with the double stairs in the aid'ruhn mage's guild too.

wareya:
This mod happens to be a good test for performance problems. It seems that my movement solver changes don't cause any performance problems (at least in normal situations e.g. actors aren't stuck inside of statics), but the choice of collision hull does.

Cylinders have a mild framerate hit because of the mannequins overlapping the wall. This is almost certainly a problem with bullet, probably the normals it's spitting out. There's no particular reason cylinders should be much slower than capsules.

Rounded cylinders have a terrible framerate hit.

Capsules are the fastest. But again, capsules are mildly incompatible with vanilla Morrowind's level design (e.g. Ebonheart's submerged stairs).

akortunov:
> I fixed the vibrating, but is this one supposed to be floating?

Just for reference: if actor with side collision is spawned in air, vanilla engine just applies gravity to him, ignoring side collisions.
I encountered such behaviour in Animated Morrowind: Expanded (a beggar near Ra'Virr's shop in Balmora is initially spawned in the air).

wareya:
It could work if collisions with statics the actor is already touching are ignored, but I don't know how to do that, and it's technically unrelated to the solver tweaks.

akortunov:
> It could work if collisions with statics the actor is already touching are ignored, but I don't know how to do that, and it's technically unrelated to the solver tweaks.

Then just leave this issue for another PR.

wareya:
Okay, so there are basically only two remaining things for me to work on with this.

- I need to add SOME kind of configuration option for cylinders vs capsules. Capsules perform better, so if you're playing on an extremely undepowered ARM laptop or android device, it makes sense that you might want to switch to them, just like using a very low draw distance or culling small objects. I don't know why. Blame bullet.

- Someone on IRC mentioned that scripted moving platforms can push the player upwards in vanilla. This works in this branch, but it's basically accidental, and it's speed-limited, unlike vanilla.. If there's an easy way to do it right I want to do it here too.

kcat:
Given the significance the actor shape can have on gameplay (breaking things, allowing you to do things you couldn't with the other shape or not allowing you to do things you could with the other shape), any configuration option should be part of the game data. Either in the model or the omwgame/addon. Having it as a config setting doesn't make logical sense to me when a game is designed for a specific shape and can be unbeatable if not set appropriately for the game in question.

AnyOldName3:
Also, given that your investigation has shown that Morrowind uses AABBs, is that not the best thing to use as the default?

wareya:
I think cylinders should be fine. Using cylinders instead of AABBs doesn't actually have a significant effect on gameplay, and cylinders are rotationally invariant. AABBs are wider or skinnier if you're looking in different directions, which means that, for example, the direction that a tree is rotated affects how you slide when you run into it, and they make it harder for NPCs to slip past other NPCs or obstacles.

Unlike capsules, there are no major obvious cases where using cylinders instead of AABBs should render something unnavigable. The only example I can think of would be a jump that's easier to make with AABBs because the hitbox's bottom surface is longer in certain directions. But jump distance is stat- and fatigue-dependent, and tight jumps are extremely bad first-person RPG design anyway. I'm sure that Morrowind has some, but no intentional ones.

Rather, cylinders might make previously UNnavigable things navigable, like thin corridors, but it's already the case that vanilla navigability is broken - bosmer can't walk under certain geometry that dunmer can, and high elves can walk under geometry (even INTO it) that dunmer can. If you ignore what vanilla is "supposed" to do and look at what it ACTUALLY does, then you'd run into all sorts of insane edge cases if you want to replicate its navigability exactly.

My philosophy is to stick to mimicking what's very impactful and doing what would have been the right thing for vanilla to do for the rest.

The movement solver is already very different from vanilla's, and in some ways it's different in ways that break vanilla assets (e.g. you can't walk between the double stairs in the ald'ruhn mages guild in OpenMW, you can in vanilla, the reason vanilla can is because it lets you penetrate geometry in certain situations), so I think making the choice to use cylinders instead of AABBs shouldn't be impactful.

kcat:
AABBs are rotationally invariant too. AA = axis-aligned, so the box has a fixed alignment on the global axis (the box's X axis is always on the global X axis). Regardless of how you turn/rotate, the bounding box doesn't change.

wareya:
I meant that if everything is rotating meshes and cylinders, then the world is rotationally invariant. If you rotate the world then collisions with AABBs give different results.

wareya:
I'm moving soon and haven't had the time to finish this up. If 0.45 approaches or this PR gets too old, feel free to branch it, tie up the loose ends, and make another PR. Or just merge it and fix up the loose ends. Whatever works. If it happens.

wareya:
Added some more changes, this should be as good as I can get it with bullet. I guess this is ready for review now.

wareya:
Probably affects bug 2274: https://gitlab.com/OpenMW/openmw/issues/2274

Capostrophic:
I haven't noticed any difference with master when trying to move the player with a platform moved with Move. Should it be tested with the current fix for 2274 reverted?

The character slides down from a slope on a rock weirdly when running on it and jumping. It's not good in master either, but the player outright vibrates while sliding with the changes.

wareya:
That's a combination of physics framerate and not stalling the player when they get "reflected" if they're in the air. The actor gets deflected when they hit surfaces and if they run into an obtuse concave corner they'll get deflected backwards in interchanging directions each frame.

image

The smaller the physics framerate is the smaller the vibration is, but the vibration is there even in big games with really good movement solvers like Quake 3. They usually work around it by having acceleration so the speed you have each frame is reduced until the vibration is so tiny that some kind of state can hide it, but Morrowind's movement mechanics have fixed velocities with no speeding-up or slowing-down curve.

I just made it so that getting fully deflected stops movement solving, but only on the ground. If it stopped movement solving in the air, it would make it very hard to jump up onto objects from places with oblique angled concave walls, like pairs of pots.

Here's a video of the same problem in Quake 3, plus an example of the special state it has to keep it from vibrating at all keeping me from running across the wall at an angle that works when not in that state: https://streamable.com/pnwv8

akortunov:
The only thing I do not like here is performance. For me physics update takes 3x more time than in master. I can live with it, but not all players have Core i7.

wareya:
Does changing the collision shape back to capsules improve performance?

wareya:
Okay, that brings the question of whether we should have a config option for the collision shape back onto the table.

akortunov:
Are there any possible optimizations? Do not calculate physics every frame for standing actors, for example?

akortunov:
Any thoughts about [bug #4476](https://gitlab.com/OpenMW/openmw/issues/4476)?
Gondoliers are twitchy because they try to collide with both water and player and have no enough free space. In theory, they should ignore collision with player since player is inside gondolier (as I understand, ActorOverlapTester should do this stuff).

wareya:
I'm not sure what's going on there so I can't give a very good response. You were right that vanilla doesn't push you off actors if you're just floating on them though, it only does so when falling and also inside of them.

akortunov:
> I'm not sure what's going on there so I can't give a very good response.

From what I can tell, we apply gravity for gondolier for some reason when he collides with player (despite he has a water walking), and script moves him back to Z=0 position. Such behaviour provides twitching.

EDIT:According to my testing, even if I completely turn off collisions with CollisionType_Actor mask (inside the ClosestNotMeConvexResultCallback), there is a twitching. But this twitching is gone when I turn off player's collisions via TCL.

AnyOldName3:
Does that suggest that we have incorrectly-flagged nodes?

wareya:
So what do I have to do to prepare a change this major for review?

akortunov:
Not related to preperation to review: I wonder if there is an ability to use different collision shape types for different distances for NPCs (for example, use cylinder shapes for nearby actors and use box or polyhedral shapes on large distance).

wareya:
That would cause problems with suddenly popping into or out of other objects when the distance changes. Reducing the physics framerate of distant actors might be more viable, but would come with its own lot of problems.

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -454,19 +653,42 @@ namespace MWPhysics
...
+                    {
+                        // note: ground can't be an actor so no need to pick safety margin
+                        if(tracer.mFraction*groundDistance > sSafetyMargin)
+                        {
+                            newPosition = tracer.mEndPos;
+                            newPosition.z() += sSafetyMargin;
+                        }
+
+                        // safely eject from ground (only if it's walkable; if it's unwalkable this makes us glide up it for a bit before gravity kicks back in)
+                        if (!isFlying && !isOnSlope)
```
Capostrophic: isFlying is false in this scope

wareya:
So should I merge master into this branch and deal with any conflicts that occur?

akortunov:
It seems your branch has no conflicts with master.

wareya:
Okay, I merged it into a copy of master just to double check and it compiles and works properly.

psi29a:
We ready to merge?

wareya:
Yeah. I'll be ready to deal with anything that goes wrong if anything got missed.

akortunov:
> We ready to merge?

As I said, this PR decreases performance of physics a lot (up to 3 times on my hardware when using cylinders).
I'd suggest to add a setting to choose a physics shape (with capsules as a default value).

wareya:
I can add an option for the physics shape if it's understood that it's a temporary thing for until we figure out why cylinders perform so much worse than capsules. OpenMW's current movement solver is basically very broken in certain ways, and fixing it will decrease performance at least a little, but it shouldn't be 3x, and a large part of that 3x is switching from capsules to cylinders.

psi29a:
Would two spheres inside a box be a solution instead of just a box or just a capsule?

Capostrophic:
From what wareya stated earlier it seems that box and spheres combined into a single collision shape perform even worse than cylinders.
There's a need of a cleanup pass. There's no need for commented code ('cause git) and debug remains.

apps/openmw/mwphysics/actor.cpp
```diff
@@ -1,6 +1,9 @@
 #include "actor.hpp"
 
 #include <BulletCollision/CollisionShapes/btCapsuleShape.h>
+#include <BulletCollision/CollisionShapes/btCylinderShape.h>
+#include <BulletCollision/CollisionShapes/btSphereShape.h>
+#include <BulletCollision/CollisionShapes/btMinkowskiSumShape.h>
```
Capostrophic: Capsule, sphere and minkowski sum includes are unused.

Capostrophic:
It's somewhat risky to merge a PR which hasn't been reviewed well that is able to degrade performance of the main bottleneck that even exists and what users complain about just before a new release.

psi29a:
@Capostrophic It is, that is why I'm asking if it is ready so that the issues can be addressed.

AnyOldName3:
I can't remember why we settled on capsules/cylinders when we know vanilla uses AABBs. If it's something like weird normals, isn't that something to report to Bullet rather than ignore?

Capostrophic:
They're not rotationally invariant, so the actor collides differently depending on what angle the surface is rotated under to the actor. But using them for objects, creatures *and* NPCs seems to be much more consistent and the irregular collision downside also seems to be much more minor to me than the performance issues of cylinders and Morrowind incompatibilities of capsules. And existing Morrowind content (like, the actual bounding boxes you see in NPC animations that can be adjusted on every axis unlike cylinders in the PR) was definitely built with AABBs and not cylinders in mind.

jvoisin:
This should be documented in the code, instead of pop'ing every 5 months as comments on pull-requests :/

AnyOldName3:
If every piece of vanilla content was created with non-rotationally-invariant collision shapes in mind, then using non-rotationally-invariant collision shapes in the physics system isn't a problem, and should, in fact, be what we aim for until we're able to let `omwgame` files specify which they want to use.

wareya:
AABBs have problems with actors passing each other in certain spiral staircases that vanilla doesn't have because vanilla's movement solver is a lot worse and lets you enter certain objects if you get close to them in the wrong way. Using cylinders is technically wrong, but works around a few cases where vanilla's AABBs work but OpenMW's wouldn't. That's my rationale, at least.

wareya:
Okay, did some cleanup. What should I do about the collision shape? I'm okay with changing to something other than cylinders for now, just to get the rest of the movement solver fixes ready, or an option if it's acceptable.

Capostrophic:
I still vote boxes. The performance difference between AABBs and cylinders is rather drastic for me, I find vanilla compatibility preferable in this case anyways and I'm willing to trade unwalkable stairs or weird bullet shenanigans to a bit more inconsistent angled collisions.
I don't see most of the issues with Morrowind design that switch to capsules worked around in the past without them anyways with the other changes now. Point to keep in mind: if capsules are got rid of, Bullet requirement can be downgraded back to version 2.83.

akortunov:
> When I tested stuff on my machine, AABBs were half way between capsules and cylinders in performance. But it might be different.

In my testing AABBs were much faster than capsules.

psi29a:
@xyzz Does the Android fork use AABBs to work around perf issues?

akortunov:
Link to commit, just for reference: https://github.com/xyzz/openmw/commit/6d5607d282a2d5dbd55ac855c5a2b2fbc263791a

jvoisin:
As an owner of low-end hardware with OpenMW running at 20fps, I'm perfectly fine if struggling in stairs is the price to pay to have the game not running a 7fps.

akortunov:
I wonder if Bullet has a hexagonal shape. Probably it could solve a part of issues with AABBs but perform faster than cylinders.

wareya:
The game Warsow actually uses octagonal prisms. I don't think Bullet has a hexagonal shape, but I don't see anything stopping us from using a mesh...?

wareya:
It looks like if we want to use AABBs we have to set mRotationallyInvariant to true, or they'll rotate with the actor, and make updateRotation a no-op, or they'll rotate to the actor's angle when they load into the game.

wareya:
Octagonal prisms work. The code is disgusting, but it's just for testing. If it's good I'll figure out a way to clean up the code.

Test performance please.

```c++
#include <BulletCollision/CollisionShapes/btConvexHullShape.h>

//...

    // the ratio between the side length and inradius of a regular octagon is 2tan(pi/8)
    static const double eir = tan(M_PI/8); // edge_inradius_ratio, divided by two
    btVector3 vertices[16];
    vertices[0]  = btVector3( 1.0,  eir,  1.0);
    vertices[1]  = btVector3( eir,  1.0,  1.0);
    vertices[2]  = btVector3(-eir,  1.0,  1.0);
    vertices[3]  = btVector3(-1.0,  eir,  1.0);
    vertices[4]  = btVector3(-1.0, -eir,  1.0);
    vertices[5]  = btVector3(-eir, -1.0,  1.0);
    vertices[6]  = btVector3( eir, -1.0,  1.0);
    vertices[7]  = btVector3( 1.0, -eir,  1.0);
    vertices[8]  = btVector3( 1.0, -eir, -1.0);
    vertices[9]  = btVector3( eir, -1.0, -1.0);
    vertices[10] = btVector3(-eir, -1.0, -1.0);
    vertices[11] = btVector3(-1.0, -eir, -1.0);
    vertices[12] = btVector3(-1.0,  eir, -1.0);
    vertices[13] = btVector3(-eir,  1.0, -1.0);
    vertices[14] = btVector3( eir,  1.0, -1.0);
    vertices[15] = btVector3( 1.0,  eir, -1.0);
    
    auto tempshape = new btConvexHullShape();
    
    for(int i = 0; i < 16; i++)
    {
        vertices[i] *= toBullet(mHalfExtents);
        tempshape->addPoint(vertices[i]);
    }

    mShape.reset(tempshape);
    mRotationallyInvariant = true;
```

screenshot272

akortunov:
As for me, octagonal prism is only a bit slower than boxes.
I wonder why TCB can not render it properly.

wareya:
It renders it as a series of lines following the points in the order the points were added to the shape. It could be forced to render it better by adding points redundantly; I don't know if that would have an effect on performance.

wareya:
Now that I think about it, ordering them like this would probably be better:

screenshot274

But the more specific the way they're ordered is the less likely it is I'll be able to clean the code up. Maybe I should just resign myself to the octagonal prism code being gross.

psi29a:
I like where this PR is going :)

wareya:
ps: What are we /supposed/ to do when mHalfExtents isn't square from a topdown perspective? As in, what would vanilla do? Give a non-square AABB?

wareya:
Okay, pushed changing it to an octagonal prism. I think that for non-square hulls it's going to be a stretched octagonal prism. If octagonal prisms are bad I'll just change it back to an AABB and call it done.

apps/openmw/mwphysics/actor.cpp
```diff
@@ -28,39 +26,46 @@ Actor::Actor(const MWWorld::Ptr& ptr, osg::ref_ptr<const Resource::BulletShape>
...
+    {
+        vertices[0]  = btVector3( 1.0,  eir,  1.0);
+        vertices[1]  = btVector3( eir,  1.0,  1.0);
+        vertices[2]  = btVector3(-eir,  1.0,  1.0);
+        vertices[3]  = btVector3(-1.0,  eir,  1.0);
+        vertices[4]  = btVector3(-1.0, -eir,  1.0);
+        vertices[5]  = btVector3(-eir, -1.0,  1.0);
+        vertices[6]  = btVector3( eir, -1.0,  1.0);
+        vertices[7]  = btVector3( 1.0, -eir,  1.0);
+        inited_octagon = false;
```
Capostrophic: This doesn't look correct.

apps/openmw/mwphysics/actor.cpp
```diff
@@ -28,39 +26,46 @@ Actor::Actor(const MWWorld::Ptr& ptr, osg::ref_ptr<const Resource::BulletShape>
     mHalfExtents = shape->mCollisionBoxHalfExtents;
     mMeshTranslation = shape->mCollisionBoxTranslate;
 
-    // Use capsule shape only if base is square (nonuniform scaling apparently doesn't work on it)
-    if (std::abs(mHalfExtents.x()-mHalfExtents.y())<mHalfExtents.x()*0.05 && mHalfExtents.z() >= mHalfExtents.x())
+    // the ratio between the side length and inradius of a regular octagon is 2tan(pi/8), which is equivalent to 2(sqrt(2)-1)
```
Capostrophic: The following value is equivalent to tan(pi/8) or sqrt(2)-1. Is the comment or the constant incorrect?

wareya: The variable is half the actual edge-inradius ratio, because it's measured from the axis, not the other vertex of the edge.

wareya: I can elaborate on the variable's comment if you want.

Capostrophic: Makes sense.

Capostrophic:
In Morrowind bounding boxes are rotated according to the actor rotation and don't use world-space axis.

wareya:
That's definitely not how they work when an actor is colliding with the world. Here: (still uploading at time of post) https://streamable.com/49qsj

akortunov:
Stairs near player from [this](https://i.imgur.com/3l8H4SP.jpg) screenshot do not perform well with this PR - actors often play landing animation repeatedly.

wareya:
Which ones?

akortunov:
> Which ones?

Ahead of player.

akortunov:
As for vanilla engine, we can not trust the TCL output - it does not render actual collisions here.

Capostrophic:
Meh. TCB being unreliable is unsurprising.

The performance seems to be pretty good.

wareya:
The situation with ebonheart stairs is weird - they're a single mesh rank going up to a flat surface with exact connection and it's still making something mess up. I don't get it.

It doesn't happen with btBoxShape. Or a btConvexHullShape containing the equivalent of a btBoxShape.

wareya:
Okay, my plan for mudcrabs:

Average the horizontal dimensions of the AABB together, forcibly center the scaled transform in updateCollisionObjectPosition(). This is probably a bad idea! It might break some kind of mod that, like, has a dummy actor with a very long (horizontally) AABB. What should I actually do here? Is there a way to "detect" dummy actors?

My plan for the weird collision glitch akortunov noticed:

Just use AABBs.

psi29a:
@wareya anyway to do bound checking for those 'very long' AABBs?

Mods that do weird crap are probably just leaning into Morrowind's engine idiosyncrasies anyway.

wareya:
What kind of bound checking? Do you mean like, if the AABB is 2x1, make a 2x2 AABB? Or do you mean something related to https://github.com/OpenMW/openmw/pull/1980 ?

wareya:
PS: I will resolve the new merge conflict soon.

wareya:
Morrowind has a behavior where, if overlapping an actor, you and them act like cylinders if you're on the ground, or if you're in the air and not falling. But if you're in the air and falling, then it pushes you outside of the actor, like OpenMW used to does without this PR when you stand on top of them.

I implemented the cylinder-like-overlapping-actor-collisions part in this PR, but not the pushing-outside-of-if-falling part, because it needs trace.cpp to know whether the currently-being-thrown actor is onGround or not (checking the vertical velocity alone will have false positive in some situations).

Capostrophic:
>Also, this PR probably breaks the fix for Bug #2274 unless the platform moves extremely slowly.

You keep saying that (I remember you've said something similar before) but I don't understand what would be the problem because the fix is not physics-related at all, it simply applies the platform position change to the position of actors that are standing on the platform.

With the latest changes mRotationallyInvariant boolean is redundant.

wareya:
My bad, I keep forgetting that we already went over that and that I
misunderstood how the fix worked.

About rotational invariance: Vanilla might actually use rotated creature
hulls for humanoid-creature collisions... I'm not sure. Just know that
generally speaking it doesn't rotate AABBs in the simplest cases I tested.
When jumping onto a paralyzed mudcrab from above I *think* rotation might
have affected actor-creature collisions. I'll have to test it again. Might
not be time to remove it yet.

psi29a:
So this is still [WIP] or [More Testing Needed] ?

wareya:
It's ready for code review.

wareya:
I'm getting that ebonheart phantom drop issue with AABBs, too,  on the vivec cantinas. Might need to make a change to the way that falling/landing is detected. As far as I can tell it happens in a span of one or zero frames, and of course there's no distance, so it should be possible to detect.

wareya:
Also, after merging in master on a private branch, I got the recast navigation stuff and toggleactorpaths renders the actors as cylinders. Is this how the navmesh generator sees actors? Should I figure out how to change it to AABBs?
screenshot294

psi29a:
Paging Dr. @elsid  :)

wareya:
Sheeeesh, this is so bad! How do I build bullet myself in use-double-precision-floats mode so I can see if it's the problem?

https://streamable.com/wyel7

wareya:
I finally managed to get bullet with double-precision floats compiled and working and linked into OpenMW

- It fixed the weird physics glitches on meshes with giant triangles e.g. in vivec and ebonheart
- It broke the tcb command in exteriors for some reason
- Now it freezes eventually in exteriors (after a minute or so) (interiors seem fine?)
- It was really annoying to get bullet to compile correctly, really

Bullet was compiled with:

```
cmake ../ -LA -G "Visual Studio 14 2015 Win64" -DCMAKE_BUILD_TYPE=Release -DUSE_DOUBLE_PRECISION=ON -DBUILD_CPU_DEMOS=OFF -DBUILD_PYBULLET=OFF -DBUILD_BULLET2_DEMOS=OFF -DBUILD_OPENGL3_DEMOS=OFF -DBUILD_BULLET3=OFF -DUSE_MSVC_RUNTIME_LIBRARY_DLL=ON
/c/Program\ Files\ \(x86\)/MSBuild/14.0/Bin/MSBuild.exe ALL_BUILD.vcxproj -p:Configuration=Release;Platform=x64
```

Made the following changes to OpenMW (after merging in master for recastnavigation):

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 606d27bfd..4a83c1a5c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,8 @@ if (USE_SYSTEM_TINYXML)
     include_directories(SYSTEM ${TinyXML_INCLUDE_DIRS})
 endif()
 
+add_definitions(-DBT_USE_DOUBLE_PRECISION)
+
 # Platform specific
 if (WIN32)
     if(NOT MINGW)
diff --git a/components/detournavigator/recastmeshbuilder.cpp b/components/detournavigator/recastmeshbuilder.cpp
index e325b7eaf..bf6dc47c2 100644
--- a/components/detournavigator/recastmeshbuilder.cpp
+++ b/components/detournavigator/recastmeshbuilder.cpp
@@ -141,28 +141,28 @@ namespace DetourNavigator
         aabbMin = transform(aabbMin);
         aabbMax = transform(aabbMax);
 
-        aabbMin.setX(std::max(mBounds.mMin.x(), aabbMin.x()));
-        aabbMin.setX(std::min(mBounds.mMax.x(), aabbMin.x()));
-        aabbMin.setY(std::max(mBounds.mMin.y(), aabbMin.y()));
-        aabbMin.setY(std::min(mBounds.mMax.y(), aabbMin.y()));
+        aabbMin.setX(std::max((double)mBounds.mMin.x(), aabbMin.x()));
+        aabbMin.setX(std::min((double)mBounds.mMax.x(), aabbMin.x()));
+        aabbMin.setY(std::max((double)mBounds.mMin.y(), aabbMin.y()));
+        aabbMin.setY(std::min((double)mBounds.mMax.y(), aabbMin.y()));
 
-        aabbMax.setX(std::max(mBounds.mMin.x(), aabbMax.x()));
-        aabbMax.setX(std::min(mBounds.mMax.x(), aabbMax.x()));
-        aabbMax.setY(std::max(mBounds.mMin.y(), aabbMax.y()));
-        aabbMax.setY(std::min(mBounds.mMax.y(), aabbMax.y()));
+        aabbMax.setX(std::max((double)mBounds.mMin.x(), aabbMax.x()));
+        aabbMax.setX(std::min((double)mBounds.mMax.x(), aabbMax.x()));
+        aabbMax.setY(std::max((double)mBounds.mMin.y(), aabbMax.y()));
+        aabbMax.setY(std::min((double)mBounds.mMax.y(), aabbMax.y()));
 
         const auto inversedTransform = transform.inverse();
 
         aabbMin = inversedTransform(aabbMin);
         aabbMax = inversedTransform(aabbMax);
 
-        aabbMin.setX(std::min(aabbMin.x(), aabbMax.x()));
-        aabbMin.setY(std::min(aabbMin.y(), aabbMax.y()));
-        aabbMin.setZ(std::min(aabbMin.z(), aabbMax.z()));
+        aabbMin.setX(std::min((double)aabbMin.x(), aabbMax.x()));
+        aabbMin.setY(std::min((double)aabbMin.y(), aabbMax.y()));
+        aabbMin.setZ(std::min((double)aabbMin.z(), aabbMax.z()));
 
-        aabbMax.setX(std::max(aabbMin.x(), aabbMax.x()));
-        aabbMax.setY(std::max(aabbMin.y(), aabbMax.y()));
-        aabbMax.setZ(std::max(aabbMin.z(), aabbMax.z()));
+        aabbMax.setX(std::max((double)aabbMin.x(), aabbMax.x()));
+        aabbMax.setY(std::max((double)aabbMin.y(), aabbMax.y()));
+        aabbMax.setZ(std::max((double)aabbMin.z(), aabbMax.z()));
 
         shape.processAllTriangles(&callback, aabbMin, aabbMax);
     }
```

psi29a:
To make things interesting... Debian ships both float32 and float64 libs (this holds true for Ubuntu as well).

https://packages.debian.org/sid/amd64/libbullet2.87/filelist

> /usr/lib/x86_64-linux-gnu/libBullet3Collision-float64.so.2.87
> /usr/lib/x86_64-linux-gnu/libBullet3Collision.so.2.87

Do we want to allow developers to chose between them or always prefer float64?

wareya:
I think the freeze might be related to the recastnavigation stuff not playing well with double-precision bullet, possibly because more changes have to be made than the little diff I posted up there that gets it compiling.

Just merging in master to this branch, still using single-precision bullet:

screenshot299

Navmeshes are coherent everywhere.

Merging in master to this branch AND using double-precision bullet:

screenshot296

Navmesh is cut off in weird places for no apparent reason.

wareya:
If you merge this into master you'll notice actors "tripping" when stepping down at the ends of staircases in Vivec or when stepping down from certain objects. That's because OpenMW (neither master nor the PR) does this weird "looney-tunes physics" thing that vanilla does, where if you walk off of steep things instead of ledges, it pretends you never left the ground (except for the fact that you can't jump): https://streamable.com/u0ydm

elsid:
> I got the recast navigation stuff and toggleactorpaths renders the actors as cylinders. Is this how the navmesh generator sees actors?

Recastnavigation supports only cylinders for actors so the navmesh generator.

elsid:
> Navmesh is cut off in weird places for no apparent reason.

There are could be a lot of reasons for this. Need to debug this branch. One way is to look how is geometry for recast is generated. There is an option to enable geometry dump into .obj format and which could be opened by recast demo. It shows geometry and allows to build navmesh.

elsid:
@wareya Using bullet with double precision doesn't work without changing openmw code. There is one problem at least. `btHeightfieldTerrainShape` is constructed from `float*`, but internally [handles](https://github.com/bulletphysics/bullet3/blob/cdd56e46411527772711da5357c856a90ad9ea67/src/BulletCollision/CollisionShapes/btHeightfieldTerrainShape.h#L90) it as `btScalar*` which is `double` with `BT_USE_DOUBLE_PRECISION` define. The size of element doesn't match. Using `PHY_FLOAT` doesn't means use `float` type, but to [use](https://github.com/bulletphysics/bullet3/blob/cdd56e46411527772711da5357c856a90ad9ea67/src/BulletCollision/CollisionShapes/btHeightfieldTerrainShape.cpp#L141) `btScalar`. And there is no support for `PHY_DOUBLE`. My suggestion is to use this [patch](https://github.com/OpenMW/openmw/pull/2049/commits/67339cc0c38a98f4efb401481b5ac47d2c4a2478). Probably there are some other places, but with [pr](https://github.com/OpenMW/openmw/pull/2049/commits/67339cc0c38a98f4efb401481b5ac47d2c4a2478) it works.

wareya:
Thank you so much!

Capostrophic:
The more this PR keeps going, the more I think about how it might be better to separate collision shape changes (and Bullet double precision experiments) into a different PR and leave the rest of movement solver changes alone so they can be reviewed and merged.

AnyOldName3:
The double precision stuff is already in its own PR.

wareya:
How should I best remove the movement hull changes from this branch?

One issue I can imagine might happen with the fix for overlapping actor-actor collisions assuming a collision shape with flat tops/bottoms, but if it's not actually a problem in testing then it's fine.

Aside from that I don't have any real issues in making switching to AABBs its own changeset.

wareya:
A new forum user is considering writing a bespoke collision system for OpenMW. If they do, then it would probably be a good idea for me to rewrite this PR, because having a more appropriate collision system would remove the need for certain hacks make it easier to do certain things.

https://forum.openmw.org/viewtopic.php?f=6&t=5372&start=60#p60337

If they don't do it and just disappear, I might do it myself, though I'd need help figuring out how transformed objects work.

terabyte25:
https://gitlab.com/Stomy/openmw/commits/dedicated-collision

wareya:
Nice!

psi29a:
Just pushed a fix for gitlab's debian runner issue and bullet. Their commits should trigger proper building now. :)

wareya:
If it would help it get into 0.46, I can make a new PR with only the most vital collision changes, and redo the other changes later. (This also serves as a repeated summary of what this PR does, logically speaking.)

Vital changes (the movement solver is "wrong" in significant ways without them): 

- Actually trace onto collision surfacse instead of sliding along them from afar (sliding is still done too, of course)
- Fixed actor-actor collisions to behave more like vanilla (make actors interact like a pair of cylinders with just-touching radiuses if they're overlapping)
- Fix the way that "not sliding up slopes" works so that it doesn't reduce vertical inertia that's already there

Important but not literally vital changes:

- Detect simple acutely-angled crevices and project inertia along their normals' cross product instead of getting bounced back and forth between them with 0 distance
- A change to how jumping and stairstepping interact (mapping the player onto steps they climb while in the jumping state, and cancelling out their vertical inertia, just like vanilla does)

Not-vital changes:

- Significantly altering the way the stair-stepping hack works (altering it is only necessary for collision hulls without round bottoms)
- Anything else I forgot (there are a number of very minor tweaks I made to make all the above changes work better with a wide variety of collision shapes, which won't be necessary if we get a bespoke collision system, or if we stick with cylinders until we get double-precision bullet)

You could also merge this PR as-is, but it makes a lot of changes that are hard to review when they're all added together because they all integrate together, so it was probably a bad idea for me to pile all of them into a single PR. Sorry about that!

There are also a couple things that vanilla does that this PR doesn't:

- "Looney tunes" physics if you run off cliffs that end in slopes instead of walls or overhangs (you're in a "walking" state, but in the air
- Treating other actors like floors that also push you out of them, rather than like cylinders, if you're in the air and falling
- Vanilla doesn't make objects pop out of "the world" when they move. Replicating this exactly the same way as Vanilla isn't a good idea, but replicating it for "one-way walls" (rear-facing collisions) is a good idea. Apparently OpenMW is supposed to ignore rear-facing collisions, but an issue somewhere in bullet (possibly a lack of internal edge rejection) is causing it to not work, I think.

psi29a:
If you could split this out into the three things you mentioned, that would make reviewing easier but is that easier? Are they dependent on each other? Do we still need to get the bullet double precision PR in first?

Capostrophic:
>Do we still need to get the bullet double precision PR in first?

Double precision is only a must for collision hull changes. That was the point in separation of changesets.

wareya:
Bullet double precision going in first isn't necessary if I change it back
to using capsules. I asked for advice on how to revert that code but didn't
get a response. I might just copy paste most of the old code for selecting
the actor hull shape back in.

Splitting the PR up into multiple parts would be quite a bit of work, but
if doing so helps the important parts get into 0.46 then I'd do it.

Capostrophic:
>I might just copy paste most of the old code for selecting the actor hull shape back in.

Plz no. I thought it was argued for a bunch of times why allowing to configure the hull is not reasonable. Just leave capsules alone meanwhile.

wareya:
I meant copy pasting *over* what I currently have, so that the file
basically goes back to looking how it did before I started messing with it.
So, manually reverting my own changes to that file, essentially. (I'm away
from my computer for two days, so I'm not very eloquent due to lack of real
keyboard.)

wareya:
I replaced the hull code with the code for capsules and did some minor comment maintenance.

psi29a:
One last round of reviews please...

apps/openmw/mwphysics/trace.cpp
```diff
@@ -9,15 +9,36 @@
 #include "actor.hpp"
 #include "convert.hpp"
 
+#include <iostream>
```
Capostrophic: Debug remnants

apps/openmw/mwphysics/trace.cpp
```diff
@@ -121,4 +176,5 @@ void ActorTracer::findGround(const Actor* actor, const osg::Vec3f& start, const
     }
 }
 
+
```
Capostrophic: Extra line break

Capostrophic:
Multiple overlapping capsule shapes still make performance take a huge blow, which oddly doesn't happen with AABBs, while in master they simply couldn't overlap and actors are drifting away in a jumping pose without much of an issue (other than falling through the ground sometimes). Could anything be done about this?

I'll try to take a closer look in the evening (of 27th, GMT).

Also, not sure how did the "actor collision hack not working properly with capsules" thing go ahead, maybe it's a bit too early to check stuff.

wareya:
>Also, not sure how did the "actor collision hack not working properly with capsules" thing go ahead, maybe it's a bit too early to check stuff.

Basically:

With AABBs, it makes actors act like hugging cylinders when they're overlapping, which is how vanilla does it. If you try to walk into the other actor, you slide around them in a circle. You slowly get pushed outside of them (until they're behind you) because you only move in tiny straight lines, and also because of the safety margin, but it feels just like rubbing against a cylinder.

With capsules, you get snagged for some reason. I don't know why. This is probably related to the issue you're describing.

wareya:
I've been thinking about possible reasons for a week but I still have no real idea why the actor-actor collision fix works worse with capsules than AABBs/cylinders.

psi29a:
Do we put this on pause for a bit and wait on stormy? Or is this 'loose' from his work?

wareya:
This PR changes how OpenMW interacts with bullet a little but it shouldn't cause any real problems for the collision branch I don't think.

psi29a:
Will this PR make thing worse for instances that still use float precision in bullet instead of double?

I ask this because then we have to discuss if we want to 'ban' building against libbullet if it builit with floats instead of doubles.

wareya:
I think it won't, because I changed the collision hull back to a capsule. Double-precision is needed for hulls with sharp edges to work properly against large collision surfaces like part of the exteriors in Vivec and Ebonheart. Capsules have no sharp edges.

psi29a:
OK, one last call for reviews. :)

@akortunov  @Capostrophic  @AnyOldName3: please sign off?

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -322,7 +366,7 @@ namespace MWPhysics
                 velocity = (osg::Quat(refpos.rot[2], osg::Vec3f(0, 0, -1))) * movement;
 
                 if ((velocity.z() > 0.f && physicActor->getOnGround() && !physicActor->getOnSlope())
-                 || (velocity.z() > 0.f && velocity.z() + inertia.z() <= -velocity.z() && physicActor->getOnSlope()))
+                 || (velocity.z() > 0.f && velocity.z() + inertia.z() <= -velocity.z() && physicActor->getOnSlope())) // note: this should be less tolerant now that it's harder to get stuck in acute seams
```
Capostrophic: I'm concerned about this hack of mine because false-positive situations it (being a temporary solution) can cause are not good, and if anything, its usage should be minimized.

wareya: It's still important since it's still pretty easy to get stuck with capsules, but it should probably be replaced with something that's harder to trigger. Maybe based on lack of travel rather than velocity. Might be a good target for a followup PR.

Capostrophic: Well, it's not critical I guess.

Capostrophic:
Other than ~~the performance~~ the odd ground rejection that also happens in master and Morrowind after a flying player finishes flying while being as close to the ground as possible, capsules not liking actor-actor collision fix and creature bounding boxes untouched rotation making large creature (like ogrim) movement jerky (that also happens in master) I can't find any significant physics issues.

akortunov:
I wrote earlier that physics is a twice slower with PR, so I see no point to merge it right now.

At least one of these things should happen first:
1. #2049 is finished and merged, so we can use boxes shapes to avoid performance loss.
2. Stomy finished [his work](https://gitlab.com/Stomy/openmw/commits/dedicated-collision) on physics optimizations and collision detection became at least 2x faster.

psi29a:
Thank you guys for the summation and review.

OK, putting this on hold for now. until the 2x slower physics is negated or resolved.

@wareya don't lose hope! :)

wareya:
Update to current master, needed changes despite lack of merge conflict.

wareya:
I haven't seen any activity on Stormy's collision system branch for a little over two months. Should I do anything?

StomyPX:
I didn't see this, or else I would've said not to wait for me. So I'll say that now: "Don't wait for me, I at least won't be ready for 0.46, and on the off chance that I finish and it is no better than before, you would have been waiting for nothing" so yeah go ahead and I'll just adapt the new system to meet this behavior as I write it

wareya:
Okay, so that means driving https://github.com/OpenMW/openmw/pull/2049 through so we can switch to AABBs for speed, or accepting the performance penalty of using the new movement solver with capsules temporarily.

wareya:
Guys, someone's gotta make an executive decision on whether or not this is going to be held back by changing the physics system (either double-precision bullet or stomy's custom system). If it's going to be held back then I'm going to make a simpler version to get some of the fixes in sooner, so please make a decision.

Capostrophic:
I guess it's going to be held back by a while.

wareya:
Okay, one more vague summary of changes here so far in case I flake out and don't make a miniature version for a while:

1) actually mapping against things the collider comes into contact with instead of just doing the inertia rejection part
2) lack of a safety margin, making it unsafe to do the above (this might be where a small part of the performance loss comes from)
3) fix the "don't skid up steep slopes" logic so that you can jump while "+forward"ing into walls
4) fix some differences between OpenMW and Morrowind stairstepping logic (i.e. when jumping onto a ledge, either inertia is mapped onto the "ground" afterwards or not, right now OpenMW and Morrowind behave differently regarding this)
5) handle collisions between overlapping actors as a special case where they act like cylinders that are just barely touching, instead of their actual collision shape (while this plays poorly with capsules for some reason, it's better than what OpenMW is currently doing at the moment)
6) additional stairstepping hacks to make up for certain bad Morrowind assets when using AABBs/cylinders
7) detection of acute crevices and following them instead of getting trapped until running out of "bounces" (this is probably where most of the performance losses are)

3, 4, and 5 can be implemented immediately. 6 is unnecessary until we switch to AABBs. 1 and 2 are absolutely vital, but might cause a minor performance loss. 7 is the most likely cause of performance losses but only matters in certain situations.

One thing I haven't tried yet is using a simple short-circuited movement solver when actors are exclusively touching "floors" and not "walls" or "ceilings". This is a situation where point 1, 2, and 7 don't need to come into play, so ideally they wouldn't be affecting performance in such situations. I've been working on something completely unrelated for a while, though, so I'll probably end up making a mini version of this that only implements 3, 4, and 5.

wareya:
Upon testing, the crevice detection logic actually *speeds up* the movement solver when it comes into play, because it causes the movement solver to exit in a decent amount of time instead of looping a lot unnecessarily.

wareya:
I put the extra stairstepping hacks and the extra safety margin trace behind const member variables.

psi29a:
@wareya I would like to revisit this in light of the work being done on over here:
https://gitlab.com/OpenMW/openmw/-/merge_requests/215 (multi-threaded physics).

#2049 has also been merged, so I'd say this is ripe for inclusion in 0.47

wareya:
I would love to work on this again, but the dependency setup script for msvc silently fails part way through on my system, and it's unclear why, so I can't set up a dev environment with up-to-date dependency versions right now.

ananace:
If you haven't rebased onto master recently then it may be because Qt retroactively disabled account-less installs a while back, which makes the official Qt installer unusable in CI scripts.
If you have rebased onto master recently though, then it may be because you're lacking Python - for the replacement Qt installer.

wareya:
I have python 3.8.2, is that new enough?

ananace:
Most definitely.

Could you try running the script with a `-V` as well, for verbose output, to see where it breaks.

AnyOldName3:
There was also a new failure message merged yesterday, so try pulling master. Windows comes with a stub version of Python that's *supposed* to tell you how to install Python, but actually just dies with a permissions error if you try and run it through MSYS bash. It *sometimes* puts itself earlier in the PATH than your actual Python install.

wareya:
Still doesn't work, it silently dies here:
```
Qt 5.15.0...
wareya@Toriaezu MINGW64 /c/openmwdev/openmw/CI (master)
$
```

AnyOldName3:
Related to https://github.com/jurplel/install-qt-action/issues/38?

psi29a:
@wareya if you can/want, pop on discord and we can try to work things out. I would hate that we leave this hanging just because windows doesn't want to play nice.

AnyOldName3:
@wareya I now have the same issue as you. Yay!

AnyOldName3:
@wareya Something in https://gitlab.com/OpenMW/openmw/-/merge_requests/240 should hopefully fix it. Even if it doesn't, the error should get printed.

psi29a:
Merged, I hope this does the trick. :)

wareya:
The prebuild script works now, thanks. I'll build master and poke around and see what's changed.

wareya:
Moving to #2937.

PR #1795, opened 2018-07-06
Make Goodbye block hyperlinks

Capostrophic:
It's possible to use hyperlinks while Goodbye is in effect, which leads to... some odd behavior. odd behavior

I tested Morrowind and it seems that hyperlinks do nothing when clicked with Goodbye in effect there. So I replicated that.

I consider this a part of bug 3897.

### DIFF
--- a/apps/openmw/mwgui/dialogue.cpp
+++ b/apps/openmw/mwgui/dialogue.cpp
@@ -629,6 +629,9 @@ namespace MWGui
 
     void DialogueWindow::onTopicActivated(const std::string &topicId)
     {
+        if (mGoodbye)
+            return;
+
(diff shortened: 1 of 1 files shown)


psi29a:
> closed on 2018-06-09 by Alexei Dobrohotov

So the bug was not properly fixed?

@zinnschlag let this one through or mark/dag in changelog?

Capostrophic:
>So the bug was not properly fixed?

The bug as in its initial wording was fixed (choices should have the effects of goodbye when it's present), but this inconsistency is very similar because it's related to the same Goodbye function and touches something very close to choices - topic hyperlinks, so basically the bug is "goodbye behavior is inconsistent".

zinnschlag:
Yeah, separate but related bug. Can you please file an issue, so we have the change on record? Anyway, will go ahead with the merge.

PR #1796, opened 2018-07-08
Fix check whether file name starts with x or X (bug #4497)

elsid:
Fixes [#4497](https://gitlab.com/OpenMW/openmw/issues/4497).

If path doens't contains / or \, then slashpos will be 0.
Therefore slashpos + 1 = 1 doesn't point to first symbol.
```
xmesh.nif
 ^
 slashpos + 1
```


psi29a:
Was there a bug for this?

elsid:
@psi29a I didn't find any. And I've found this not because of some problems in the game.

zinnschlag:
It would be highly unusual to find a mesh path without a /. Certainly not in vanilla content, but there is also a good chance it doesn't appear in any mods.

Still in theory this could go wrong and your code looks cleaner than the original. Please add an issue and a changelog entry and then we can merge.

PR #1797, opened 2018-07-08
Fix double call of addContainerScripts on player in moveObject (bug #4490)

Capostrophic:
[Bug 4490](https://gitlab.com/OpenMW/openmw/issues/4490)

When PositionCell is called on the player and it moves the player into an interior cell, moveObject adds scripts on items one time in changeToInteriorCell and the second time naturally, when it tries to change the cells the items expect to the new cell. I added another call of removeContainerScripts (because the cel changeToInteriorCell assigns the items to is no longer valid) on the player to prevent the warning.

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1166,7 +1166,10 @@ namespace MWWorld
             if (isPlayer)
             {
                 if (!newCell->isExterior())
+                {
                     changeToInteriorCell(Misc::StringUtils::lowerCase(newCell->getCell()->mName), pos, false);
+                    removeContainerScripts(getPlayerPtr());
+                }
                 else
                 {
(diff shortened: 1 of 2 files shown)


apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1187,9 +1190,8 @@ namespace MWWorld
                     mWorldScene->addObjectToScene(newPtr);
 
                     std::string script = newPtr.getClass().getScript(newPtr);
-                    if (!script.empty()) {
+                    if (!script.empty())
                         mLocalScripts.add(script, newPtr);
-                    }
```
psi29a: Is this in our style-guideline? 

I'm OK with have redundant {} since it doesn't hurt to have them anyway. For me it is easier to read in general.

Capostrophic: Ehh. Reverted.

PR #1798, opened 2018-07-09
Make Equip console command to bypass most of restrictions

akortunov:
Fixes [bug #4460](https://gitlab.com/OpenMW/openmw/issues/4460).

This solution is hackish, I know, but I do not see any way to make it clean.

The main idea: allow Equip console command to bypass most of restrictions (e.g. allow beast races to wear armor or repair items in combat).
Note: you still can not equip broken items via console command, as in vanilla game. I guess that vanilla game handles this restriction separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
     Bug #4457: Item without CanCarry flag prevents shield autoequipping in dark areas
     Bug #4458: AiWander console command handles idle chances incorrectly
     Bug #4459: NotCell dialogue condition doesn't support partial matches
+    Bug #4460: Script function "Equip" doesn't bypass beast restrictions
     Bug #4461: "Open" spell from non-player caster isn't a crime
     Bug #4464: OpenMW keeps AiState cached storages even after we cancel AI packages
     Bug #4469: Abot Silt Striders – Model turn 90 degrees on horizontal
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -146,7 +146,7 @@ namespace MWBase
             virtual MWGui::TradeWindow* getTradeWindow() = 0;
 
             /// Make the player use an item, while updating GUI state accordingly
-            virtual void useItem(const MWWorld::Ptr& item) = 0;
+            virtual void useItem(const MWWorld::Ptr& item, bool force=false) = 0;
 
             virtual void updateSpellWindow() = 0;
 
--- a/apps/openmw/mwclass/apparatus.cpp
+++ b/apps/openmw/mwclass/apparatus.cpp
@@ -124,10 +124,9 @@ namespace MWClass
         return info;
     }
 
-
-    std::shared_ptr<MWWorld::Action> Apparatus::use (const MWWorld::Ptr& ptr) const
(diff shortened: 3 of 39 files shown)


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@
     Bug #3997: Almalexia doesn't pace
     Bug #4036: Weird behaviour of AI packages if package target has non-unique ID
     Bug #4047: OpenMW not reporting its version number in MacOS; OpenMW-CS not doing it fully
-    Bug #4125: OpenMW logo cropped on bugtracker
```
psi29a: Why did this need to be removed? From what I understand, it fixed itself by us using GL instead of Redmine?

psi29a:
Question, why the check? If someone types 'equip', then shouldn't it just equip? Why do we need to check if it's a beast or not?

CHANGELOG.md
akortunov: It is not a bug in OpenMW itself at all, it does not have an entry in 0.45 milestone and does not have related code in 0.45 master branch.
So I do not know if the changelog entry is really needed here.
And logo is still cropped on the old bugtracker.

akortunov:
> Question, why the check? If someone types 'equip', then shouldn't it just equip? Why do we need to check if it's a beast or not?

Because we have may other restrictions. For example, you can not equip repair hammer or alchemy apparatus during combat, can not equip broken equipment, etc.

CHANGELOG.md
psi29a: I understand, I'm not sure either however... let us get in the habit of focusing only on what the PR is addressing instead of doing unrelated drive-by's. I assume that if it's a problem or irrelevant that Raevol will take care since it's his responsibility. ;) 

Please revert and we'll leave it for Raevol to decide. Good?

psi29a:
If someone types that into the console, I'm under the impression that you want to equip a repair hammer or alchemy apparatus. Why would we want to prevent that?

akortunov:
> Why would we want to prevent that?

Well, in some situations such actions (if we will equip broken items, for example) can lead to non-intended results.
Since vanilla engine keeps most of these restrictions, modmakers won't bother to handle such cases themselves and will leave them up to engine.

But if we do not want to bother to handle faulty situations too, I am fine with it (since I do not trust to scripted mods anyway).

akortunov:
> I'm just wondering if we couldn't make this something generic so that we don't have to dehardcode later in 1.0

If we do not want to handle faulty cases, I can add additional Force flag to class::use(ptr) function to completely ignore all checks in ActionEquip (or keep only condition check as more vanilla style).

akortunov:
Updated. For now I keep only condition check in the "force" mode.
I can get rid of casts, but I will need to change class.use() signatures in all files inside of mwclass folder.

zinnschlag:
@psi29a We should replicate the vanilla behaviour as closely as possible. I don't see how we can do any de-hardcoding now without breaking stuff. The equip instruction does need to check for invalid equipment configurations.

> I can get rid of casts, but I will need to change class.use() signatures in all files inside of mwclass folder.

Please do that. The whole point of the action class is to avoid special case code like that.

psi29a:
Sweet! :)

PR #1801, opened 2018-07-09
Update trainer skill cap based off modified skill instead of based skill

terabyte25:
http://en.uesp.net/wiki/Morrowind:Trainers
>If a Fortify Skill effect is cast on a trainer and makes that skill one of their three highest, it becomes one of the skills in which the trainer will offer training. Use this to make them train different skills and/or train to higher levels. For instance, casting a +100 Unarmored spell on any trainer will turn them into a master-level trainer in Unarmored for the duration of the spell.


zinnschlag:
Looks reasonable. Please update the changelog file. Also, is there an issue for this change on our tracker? If not please add one.

terabyte25:
@zinnschlag Done.

PR #1800, opened 2018-07-09
Fix build with MSVC

akortunov:
...and make AppVeyor builds great again.

EDIT: also fixes MSVC C4456 warnings (duplicate declarations).

PR #1802, opened 2018-07-09
Update CMake lists for Windows (task #2490)

Capostrophic:
[Task #2490](https://gitlab.com/OpenMW/openmw/issues/2490)
See also [feature #4487](https://gitlab.com/OpenMW/openmw/issues/4487), which is related and should be implemented later to make this change useful.

Let's see what AppVeyor will say now.

- Automatic command prompt logging has been disabled in Release builds on Windows
- Five especially spammy warnings were disabled

For that the subsystem of release builds was changed from CONSOLE to WINDOWS. These warnings are C4297 and C5032, which are fired for MyGUI and Boost respectively, and they can't be fixed from the side of OpenMW, even though AppVeyor constantly spams them, and C5039, C5045, C4768, which are also not OpenMW's fault. C5032, C5039, C5045, C4768 are also supposed to be disabled by default at least in VS2015 and later, but they're enabled again by Wall flag.

Note that I "redesigned" the original issue because it's, well, not really possible to make something set via precompiler directives and cmake magic a configuration option.

As far as I know main.cpp should already handle this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@
     Feature #4444: Per-group KF-animation files support
     Feature #4466: Editor: Add option to ignore "Base" records when running verifier
     Feature #4012: Editor: Write a log file if OpenCS crashes
+    Task #2490: Don't open command prompt window on Release-mode builds automatically
 
 0.44.0
 ------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -610,10 +610,10 @@ if (WIN32)
     endif()
 
     if (BUILD_OPENMW)
-        # Release builds use the debug console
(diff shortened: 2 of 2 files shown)


akortunov:
I'd suggest to investigate (and probably shut up) warnings C5039 and C5045 under VS2017 too.

akortunov:
@Capostrophic, once #1800 will be merged, you should update your branch.

Capostrophic:
Alright, updated. All these warnings were supposed to be disabled by default. Waiting for 1800 to be merged.

Capostrophic:
As you can see, I've rebased it already before your comment, but there's another AppVeyor build of this exact branch running that will probably fail because it's outdated. @psi29a can you cancel it?

psi29a:
Already ahead of you ;)

PR #1805, opened 2018-07-14
Use editor markers for lights and creatures levelled lists

akortunov:
Implements [feature #4512](https://gitlab.com/OpenMW/openmw/issues/4512).

Subject:
screenshot_20180714_180740

There are at least two unused markers in openmw-cs:
1. marker_creature.nif - marker for creatures levelled lists
2. marker_light.nif - default marker for light sources, which do not have own mesh.

Just use them instead of white boxes. Works for both scene and instance preview widgets.


zinnschlag:
Can't find anything. Just go ahead and add the issues?

PR #1808, opened 2018-07-17
Rework weapon animation blending

akortunov:
Fixes [bug #4575](https://gitlab.com/OpenMW/openmw/issues/4575) and [bug #4576](https://gitlab.com/OpenMW/openmw/issues/4576).

Summary of changes:
1. Apply reload animations for upper body only for crossbows.
2. Increase priority of 1st-person weapon animations instead. Basically, attack animations in the first-person mode will be the same as when steady.
3. Do not reset idle animations if we can not attack.

Improvements:
1. Try to shoot from bow in the 1st-person mode when steady.
There is a regression in my PR #1776 - there is a bow animation desync during shooting.
Does not happen with this PR.
2. Try to shoot from bow in the 1st-person mode during movement.
Shooting animation blending with movement animation provides a quite weird result in master branch - an arrow has large left offset.
With this PR arrow is rendered fine.
3. Take a longsword and prepare thrust attack in the 1st-person mode. Switch between walking/stopping.
There will be weapon twitching in upstream master, which does not happen with this PR.
4. Try to attack with ranged weapon without ammo or to cast Power several times.
Currently we reset idle animations, despite there is no real attack.
With this PR this should not happen.

Notes:
1. It would it would be nice to get some independent testing.
2. Second commit can break mods with 1st-person legs (if they are compatible). Make these changes optional?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@
     Bug #4563: Fast travel price logic checks destination cell instead of service actor cell
     Bug #4565: Underwater view distance should be limited
     Bug #4573: Player uses headtracking in the 1st-person mode
+    Bug #4575: Weird result of attack animation blending with movement animations
     Feature #2606: Editor: Implemented (optional) case sensitive global search
     Feature #3083: Play animation when NPC is casting spell via script
     Feature #3103: Provide option for disposition to get increased by successful trade
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@
     Bug #4565: Underwater view distance should be limited
     Bug #4573: Player uses headtracking in the 1st-person mode
     Bug #4575: Weird result of attack animation blending with movement animations
+    Bug #4576: Reset of idle animations when attack can not be started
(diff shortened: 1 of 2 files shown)


Capostrophic:
I'll try to test this today.

Capostrophic:
>Second commit can break mods with 1st-person legs (if they are compatible).

Vanilla handles first person view body mods poorly (legs and feet are not displayed there at all if I've not mistaken), not sure if it matters much.

These changes seems to work fine.

Unrelated: when the character tries to shoot from a ranged weapon and is not able to (because the arrows is missing), the 1st person idle animation restarts, which doesn't look well.

akortunov:
> Unrelated: when the character tries to shoot from a ranged weapon and is not able to (because the arrows are missing), the 1st person idle animation restarts, which doesn't look well.

Should be fixed now.
Note: we still reset idle animations when actos starts to move.

Capostrophic:
Great.

zinnschlag:
Missing changelog entry.

akortunov:
Added changelog entry.

PR #1809, opened 2018-07-17
Sort icons in the DataDisplayDelegate

akortunov:
Fixes [bug #4498](https://gitlab.com/OpenMW/openmw/issues/4498).
Just sorts mIcons by name, as mValues.
There is no changelog entry since this bug is a regression.


psi29a:
Failure not related (msvc15 logging), merging. Thank you!

PR #1810, opened 2018-07-19
Use fallbacks for missing weapon animations

akortunov:
Attempt to fix [bug #4470](https://gitlab.com/OpenMW/openmw/issues/4470).
Summary of changes:
1. If weapon animation does not contain "equip attach"/"unequip detach" keys, attach/detach weapon immediately.
2. If non-bipedal creature with Weapon & Shield flag has no special weapon group, use the "attack1-3" as fallback.
3. Allow non-bipedal creatures with Weapon & Shield flag to use idle animations between attacks.


Capostrophic:
Seems to work well for me, if only there weren't 4271. Increased speed of attack animations due to the invisible equipped shortsword looks funny.

PR #1811, opened 2018-07-19
Set movement speed to 0 when unconscious

akortunov:
Fixes bug [#4519](https://gitlab.com/OpenMW/openmw/issues/4519).
Just make sure that unconscious actors have no own movement (not from animations, but bases on animation state, stats and GMSTs).


psi29a:
Thanks!

PR #1812, opened 2018-07-20
Added OpenMW logos to the file. TBD if this will show up on RTD.

rhtucker:
I know we probably don't like having the same image file in two places in our source, but I don't think it will work otherwise. Let's at least merge and see if it works before we get too excited with figuring out a more elegant solution.


psi29a:
I thought we wanted to put our images for docs in another repo to keep this from happening?

psi29a:
If we merge, it's permanent. You can't back it out without re-writing history and that's a pain with so many people forked on github and gitlab. This is painful because we know we have depulication. 

There is where I suggest just having documentation exist in it's own repo (openmw-docs). Then suddenly, all these little issues go away and we can commit to our heart's content binary data since it has zero effect on the openmw repo itself.

How do you feel about that?  Would you like to be maintainer of this new repo (along with Zini and myself)? :)

rhtucker:
> If we merge, it's permanent. You can't back it out without re-writing history and that's a pain with so many people forked on github and gitlab. This is painful because we know we have depulication.

Ooo gross, didn't realize that. Then yes, we should keep it on the other repo. I would like to be a maintainer, but do we really want to keep aaaall the docs over there? Would we still be able to version things properly on this side, e.g. when a dev adds a new setting, they can add the doc entry themselves? If that's true, then I suppose I see no reason to have it otherwise.

psi29a:
Yeah... that's a pain-point... when someone tags a release in openmw, that tag's documentation is auto-generated for us.

It means that for every release (tag), we would manually have to do the same with a openmw-docs repo.

The only positive to this is that we are de-coupled from openmw if we ever had a typo that needed a quick fix.

rhtucker:
> The only positive to this is that we are de-coupled from openmw if we ever had a typo that needed a quick fix.

That’s actually probably a really good reason to go forward with this then. Although at this point I’m linking everything directly to “latest”. In the future that might not be best practice though.

I wouldn’t worry about doing this yet, most urgent is just keeping images off!

PR #1813, opened 2018-07-21
Added CSS to drop shadow of figure images

rhtucker:
Doesn't look like we can add logo and favicon, but all normal images can be hosted separately. Might be able to finagle conf.py to add them as a custom script, but it won't be easy.

In the meantime, I've moved all images to openmw-docs, removed the shadows they had on them, and got the custom CSS to do the drop shadow for us.


psi29a:
Why not port our docs to openmw-docs? :)

rhtucker:
Did we decide for sure that’s what we wanted to do? Will devs be annoyed about adding their settings to a separate repo?

psi29a:
we should probably discuss it first and get input from zini and crew

loriel2:
Moving docs to join the images in the new repo sounds like a good idea - shortens the long and complicated path from docs to images.

rhtucker:
@psi29a @loriel2 @Thunderforge @zinnschlag I've started a thread to discuss moving the docs on the forums: https://forum.openmw.org/viewtopic.php?f=43&t=5302

PR #1814, opened 2018-07-25
Fix training "can't teach more" message (bug #4494)

Capostrophic:
In PR #1801 when changing training behavior to utilize NPC modified skills instead of base values in training @terabyte25 missed that "can't teach more" message also accounted for base skill instead of modified skill in a comparison of NPC skill and PC skill, so it resulted in the fix not actually working. He realized this today and asked me to fix the line.

~~Changelog should not be updated because this is basically the second part of the fix for the bug 4494 which already has a line there.~~

Edit: though bug 4494 has incorrect number there. I think someone reported some of such issues earlier somewhere. Edit 2: fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@
     Bug #4480: Segfault in QuickKeysMenu when item no longer in inventory
     Bug #4489: Goodbye doesn't block dialogue hyperlinks
     Bug #4490: PositionCell on player gives "Error: tried to add local script twice"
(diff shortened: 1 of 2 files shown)


zinnschlag:
Hey, didn't you move your main OpenMW activity to GitLab? Or am I confusing you with someone else? Just wondering why so much GitHub activity again.

Capostrophic:
Nah, I don't even have openmw forked on GitLab. I was away from my PC for two weeks since July 11.

PR #1815, opened 2018-07-26
Do not optimize animated shapes

akortunov:
Attempt to fix [bug #3950](https://gitlab.com/OpenMW/openmw/issues/3950).

Tested with problematic meshes in Arktwend, Abot Boats, Ships of Imperial Navy and Darknut's Dwemer Dwelling.

The main idea: physics system reads list of animated collision shapes from NIF file, but works with optimized OSG node, in which some animated shapes can be removed/merged.
So I decided to do not optimize animated shapes for now.

This solutuon is far from ideal, I know.
As a proper solution, we can try to read list of animated of collision shapes from optimized OSG node (in the PhysicsSystem::animateCollisionShapes()), but it is probably out of my skills yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
     Bug #3876: Landscape texture painting is misaligned
     Bug #3897: Have Goodbye give all choices the effects of Goodbye
     Bug #3911: [macOS] Typing in the "Content List name" dialog box produces double characters
+    Bug #3950: FLATTEN_STATIC_TRANSFORMS optimization breaks animated collision shapes
     Bug #3993: Terrain texture blending map is not upscaled
     Bug #3997: Almalexia doesn't pace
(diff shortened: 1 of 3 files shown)


zinnschlag:
Did you check for a possible impact on performance?

akortunov:
> Did you check for a possible impact on performance?

I tested with scene with 50 galleons from Ships of Imperial Navy.
With these changes, FPS is near 25 on i7-7700.
Without these changes, FPS is 10-12 - OpenMW tries to find nonexistent nodes again and again.
So I do not know how to measure performance difference.

akortunov:
Updated:
1. Code cleanup
2. If animated shape was not found for some reason, remove it from list of animated shapes for this object. As result, there should be no spam in log - OpenMW will complain about missing node only once per object.

PR #1816, opened 2018-07-26
Do not use magic numbers in capacity calculations

akortunov:
Not much difference, but improves code readability.
In all other cases we use attribute IDs, so just make behaviour consistent.

A separate question: should we use the fEncumbranceStrMult for creatures too?


Capostrophic:
For some reason I have a distant memory that not using the GMST for creatures was "intended behavior" but I have no idea why shouldn't it be used and can't find any materials that could help. Let's utilize it, it makes more sense not to use a magic number just for creatures for that.

PR #1817, opened 2018-07-26
Resume game fully if the player was resurrected (bug #2626)

Capostrophic:
[Bug 2626.](https://gitlab.com/OpenMW/openmw/issues/2626)

I've added a method to state manager which is called in the Resurrect function if the game has ended and sets the game state back to "running". 

I'm not sure if this change would actually be useful, but currently even though openmw allows resurrecting the player and makes sure the player's health is not 0 after resurrection, the player is left off with a "broken" game. With this it returns to normal fully (including the menu, movement, music etc.), which looks somewhat better.

If we want the behavior of post-resurrected player game to be broken, then so be it.


Capostrophic:
AppVeyor msvc2015 is drunk again.

PR #1819, opened 2018-07-27
Use file selection dialogue result

akortunov:
Fixes [bug #4524](https://gitlab.com/OpenMW/openmw/issues/4524).
Previously local "path" variable was not used, so the file selection dialogue was useless.
Should work now (can not test on Windows by myself).


Capostrophic:
I don't think this should have a changelog line and be set on 0.45.0 milestone. It's an issue in the implementation of feature 4345, there's no such thing as "run script from text file" in 0.44.0 and previous releases.

PR #1821, opened 2018-07-28
Rework launcher icons

akortunov:
Summary of changes:
1. Provide icons for categories.
Currently, we read three icons from current theme, but provide other two, so launcher may look weird (for example, if player uses a monochrome theme).
Even more, some desktop environments do not have suitable icons. For example, my KDE5 setup does not have icon for "Advanced" button.
Existing "tango" icons do not fit other icons too (and do not work for me anyway).
New icons are from old Caledonia theme for KDE4.
2. Existing launcher icons were cleaned via "mogrify" tool from ImageMagick, so they will not cause notorious "libpng warning: iCCP: known incorrect sRGB profile" warning with libpng1.6+

screenshot_20180728_124829


zinnschlag:
Can't test this right now (launcher still broken for Qt4). Anyone else wanna give it a look before we merge?

Capostrophic:
Can't run the launcher either, but this seems to work just fine in Wine-ran Windows builds. :P

MiroslavR:
No problems here with Qt5 and the libpng warnings are indeed gone.

PR #1822, opened 2018-07-28
[Discussion] Make the camera the only sound listener

Capostrophic:
Like in Morrowind, in third person view the sound listener in OpenMW is the player character. So if you go into third person view, you'll hear any sound as your character would even if the camera is positioned on a large distance from the character and has a different direction than the character has.

This removes updateSoundListener method, moves the stuff it does into updateCamera and makes the only sound listener the camera itself in both first person and third person view, making the sound direction consistent from the player's point of view, because what the player sees is what the player gets, or hears. You'll hear the footsteps and the NPCs speaking exactly at the distance and direction they would sound relatively to the camera.

updateCamera may not be the suitable location for the sound listener stuff. It was just a quick experiment.

### DIFF
--- a/apps/openmw/mwrender/camera.cpp
+++ b/apps/openmw/mwrender/camera.cpp
@@ -5,7 +5,9 @@
 #include <components/sceneutil/positionattitudetransform.hpp>
 
 #include "../mwbase/environment.hpp"
+#include "../mwbase/soundmanager.hpp"
 #include "../mwbase/windowmanager.hpp"
+#include "../mwbase/world.hpp"
 
 #include "../mwworld/ptr.hpp"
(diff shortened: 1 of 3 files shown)


Capostrophic:
I assume NoPlayerLocal play mode will allow me to get the behavior I seek for 2D sounds played with PlaySound3D or?..

kcat:
3D sounds originating from the player are supposed to play local to the listener. Otherwise, the player's sounds for voices, footsteps, and weapons sound odd in first-person with non-stereo output.

Capostrophic:
Since I didn't receive any further feedback and this PR is no longer necessary to fulfill its original destiny, I'm closing this for better times.

PR #1825, opened 2018-07-28
Ignore stray function arguments (bug #3762)

Capostrophic:
[Bug 3762](https://gitlab.com/OpenMW/openmw/issues/3762).

Now this is a somewhat odd one. Upon encountering a stray function argument or token instead of throwing an exception and ending the script compilation there'll be a warning from the parser, but the instruction in question will still be ran.

I didn't add a line to the changelog for now because I'm not sure if this will actually get accepted.

### DIFF
--- a/components/compiler/parser.cpp
+++ b/components/compiler/parser.cpp
@@ -67,12 +67,12 @@ namespace Compiler
     // Handle an int token.
     // \return fetch another token?
     //
-    // - Default-implementation: Report an error.
+    // - Default implementation: Report a warning.
 
     bool Parser::parseInt (int value, const TokenLoc& loc, Scanner& scanner)
     {
         if (!(mOptional && mEmpty))
-            reportSeriousError ("Unexpected numeric value", loc);
+            reportWarning ("Unexpected numeric value", loc);
         else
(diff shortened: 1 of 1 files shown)


kcat:
This behavior has been intended, extra unexpected parameters cause script errors. The registration methods can declare a function may be given some number of optional parameters to be ignored, but should only be used when we know vanilla's behavior in all use-cases of the function.

For instance, it may be worth checking how vanilla is interpreting things like
```
if ( Player->HasSoulGem "Feyfolken" misc_soulgem_grand == 0 )
```
since `misc_soulgem_grand` isn't a valid part of the HasSoulGem call. Is it interpreting it as
```
if ( Player->HasSoulGem "Feyfolken" == misc_soulgem_grand )
```
or
```
if ( ( Player->HasSoulGem "Feyfolken" == misc_soulgem_grand ) == 0 )
```
or
```
if ( Player->HasSoulGem "Feyfolken" == (misc_soulgem_grand == 0 ) )
```
or
```
if ( Player->HasSoulGem "Feyfolken" == 0 )
```
or does the vanilla CS's compiler just produce a junk call that happens to not crash?

PR #1826, opened 2018-07-29
Handle RootCollisionNode, attached to non-root node

akortunov:
Should fix [bug #4311](https://gitlab.com/OpenMW/openmw/issues/4311).
For now, if a mesh has the RootCollisionNode, attached to non-root node, OpenMW will use the whole mesh as collision (including RootCollisionNode).

According to independent testing, Morrowind behaves in the same way as OpenMW, but vanilla "tcb" does not display autogenerated shapes.
Also 3DMax nif-exporter likes to wrap mesh to new root node, so RootCollisionNode becomes silently ignored.
What we can do here:
1. Consider root NiNode with only other NiNode child as a wrapper, not as actual root node.
2. If a node has the RootCollisionNode, assume that it should override collision for **this node and its children**, not for whole mesh.
3. Just print a warning, if RootCollisionNode is found inside node hierarchy, to allow modders to fix such meshes.

This PR implements a first option for now.

Tested with:
1. Attached guar cart mesh from issue description
2. Attached railroad mesh from issue description
3. Ships from Abot's Boats mod
4. Hlaalu bridges from Morrowind Rebirth ([bug #2133](https://gitlab.com/OpenMW/openmw/issues/2133))
5. Fan in the "Kazara, Living Quarters" from Darknut's Dwemer Dwelling mod


akortunov:
I tested a bit. There are a lot of meshes from replacers which have their root nodes wrapped.
It should be safe to handle collision nodes in this case.
But in other cases, we can not tell if the collision node should be used by design or not, so just print a warning on this case.

PR #1829, opened 2018-07-29
Implement ranged crits (feature #3703)

Capostrophic:
~~Yay, stealth archering!~~

[Feature 3703](https://gitlab.com/OpenMW/openmw/issues/3703).

Replicate the melee weapon code of ranged sneak criticals in projectile hit code. I'm not sure which GMST should really be used, but using critical strike multiplier instead of KO multiplier makes more sense to me, and it seems that this feature is very unfinished in Morrowind so it shouldn't be replicated bug-to-bug.

Quick testing shows that this works pretty good, if you actually manage to lend a hit on an unaware enemy. The character in the saves supplied with the report has really low Marksman skill.

I think this feature is not too "post-1.0" to be implemented and can potentially be useful before that point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@
     Feature #3083: Play animation when NPC is casting spell via script
     Feature #3276: Editor: Search- Show number of (remaining) search results and indicate a search without any results
     Feature #3641: Editor: Limit FPS in 3d preview window
+    Feature #3703: Ranged sneak attack criticals
     Feature #4222: 360° screenshots
     Feature #4256: Implement ToggleBorders (TB) console command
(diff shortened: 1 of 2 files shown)


zinnschlag:
It seems there are contradicting opinions in regard to this change. I am somewhat hesitantly okay with merging it. But we probably should get a 2nd opinion. @akortunov What is your position on this?

akortunov:
> What is your position on this?

I do not mind to merge it, despite it is a derivation from vanilla game mechanics.
Vanilla:
1. Has crits only for melee combat,  according to game guide.
2. Looks like it has crits for creatures too (I encountered it once several years ago).
3. Applies knock out multiplier for ranged combat if target is unawared, more likely as a coding error.

PR #1830, opened 2018-07-30
Rescale player avatar

akortunov:
Attempt to fix [bug #4539](https://gitlab.com/OpenMW/openmw/issues/4539).
The main idea: multiple texture size by UI scaling factor, so after scaling avatar should not look blurred.

I have no 4k display so I can not test it by myself.
[Here](https://i.imgur.com/CCrSikS.jpg) is the comparison (new/old, scaling factor 2.0) on my 1920x1080 display.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@
     Bug #4503: Cast and ExplodeSpell commands increase alteration skill
     Bug #4510: Division by zero in MWMechanics::CreatureStats::setAttribute
     Bug #4519: Knockdown does not discard movement in the 1st-person mode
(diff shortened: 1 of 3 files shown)


akortunov:
> On high scaling factors it's good. On low scaling factors the doll expectedly becomes blurry.

"Low" is < 1.0?

Capostrophic:
Yes.

akortunov:
Updated. 
Now I increase resolution only if scaling factor > 1, but do not decrease it when scaling factor < 1.

Capostrophic:
Seems to work A-OK now.

PR #1833, opened 2018-07-31
Add permanent barter disposition change option (feature #3103)

Capostrophic:
[Feature #3103](https://gitlab.com/OpenMW/openmw/issues/3103).

Add just that - when the option is true, in addition to temporary disposition change which is visible when exiting barter, permanent disposition change is modified, which is applied after the player closes the dialogue with the merchant.

I also added a checkbox to the Advanced tab of the launcher and modified the documentation.


zinnschlag:
It seems this is not merging cleanly.

Capostrophic:
Done.

PR #1837, opened 2018-08-01
Fix some issues, found by Coverity Scan

akortunov:
Monstly missing valiables initialization and missing checks for null.
Probably we will need to rescan after merge.


apps/openmw/mwgui/container.cpp
```diff
@@ -82,7 +82,7 @@ namespace MWGui
             dialog->eventOkClicked.clear();
             dialog->eventOkClicked += MyGUI::newDelegate(this, &ContainerWindow::dragItem);
         }
-        else
+        else if (mModel)
```
zinnschlag: That is an odd place to fix it. Had to look up the source to understand what is going on. IMHO the check mModel would be better placed inside of dragItem.

components/esm/loadcell.hpp
```diff
@@ -78,18 +78,29 @@ struct Cell
...
+      AMBIstruct() : mAmbient(0),
+                     mSunlight(0),
+                     mFog(0),
+                     mFogDensity(0.f)
+      {}
   };
 
   Cell() : mName(""),
            mRegion(""),
+           mData(DATAstruct()),
```
zinnschlag: These lines looks pointless. A struct will be initialised via the default constructor anyway, even if you don't list it in the initialiser list.

apps/openmw/mwgui/container.cpp
akortunov: Moved.
Now I check if model exists inside dragItem() and dragItem() calls.

components/esm/loadcell.hpp
akortunov: I removed this commit.

PR #1838, opened 2018-08-01
Combat AI priority improvements (features #4548, #4549, #4550)

Capostrophic:
Implements three ideas from [feature #3946](https://gitlab.com/OpenMW/openmw/issues/3946).

* Instead of using the raw skill value as another weapon rating multiplier, the actual hit chance is used.
* Ranged weapon bonus is now a distance-dependent percentage-based (150%) multiplier.
* Use adjustWeaponDamage to determine the actual weapon damage which is influenced by Strength attribute to use for the rating.

I found 150% multiplier to be indeed optimal. It's applied if the distance from the actor to their enemy is higher than the distance the enemy can attack from.

The second idea makes the behavior of Mulvisie Othril in Addamasartus (from [bug 3405](https://gitlab.com/OpenMW/openmw/issues/3405)) much more intelligent: instead of sticking solely to throwing stars, she will switch to her dagger at short distance (or if the player switches to a ranged weapon or spell) and use throwing stars otherwise if the player tries to walk away.

The first and third ideas are kind of hard to test, but it seems that the actor switches to another weapon (like fists) more promptly if it's hopeless to use a weapon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,15 +76,13 @@
     Bug #4519: Knockdown does not discard movement in the 1st-person mode
     Bug #4539: Paper Doll is affected by GUI scaling
     Bug #4545: Creatures flee from werewolves
-    Bug #4548: Weapon priority: use the actual chance to hit the target instead of weapon skill
-    Bug #4549: Weapon priority: use the actual damage in weapon rating calculations
-    Bug #4550: Weapon priority: make ranged weapon bonus more sensible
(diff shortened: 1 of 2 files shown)


zinnschlag:
Looks okay. But we need better record keeping. The issue you linked is pretty bad (collection of vaguely related stuff). At least file an issue for this PR, though ideally we would want split up 3946.

Capostrophic:
Done.

PR #1841, opened 2018-08-02
Replace 0 sound ranges with default distances separately (bug #4551)

Capostrophic:
To make sound attenuation for sounds that don't have both min and max range equal to 0 work properly, as mentioned by Chris under [bug 3674](https://gitlab.com/OpenMW/openmw/issues/3674).


zinnschlag:
No changelog entry for this?

Capostrophic:
there is now

PR #1842, opened 2018-08-03
Use log file for editor (feature #4012)

akortunov:
Implements [feature #4012](https://gitlab.com/OpenMW/openmw/issues/4012).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
     Feature #3276: Editor: Search- Show number of (remaining) search results and indicate a search without any results
     Feature #3641: Editor: Limit FPS in 3d preview window
     Feature #3703: Ranged sneak attack criticals
+    Feature #4012: OpenMW-CS: Create log file like with OpenMW
     Feature #4222: 360° screenshots
     Feature #4256: Implement ToggleBorders (TB) console command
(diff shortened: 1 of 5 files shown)


zinnschlag:
I like the way how you moved the code into a separate component. But there is still an awful lot of code repetition between the OpenMW and OpenMW-CS main files. Also, lots of platform-specific #ifdefs. I wonder if more of the code and the platform-specific parts could be moved to the component?

akortunov:
> I wonder if more of the code and the platform-specific parts could be moved to the component?

Moved. Now main() calls a wrapper function, which runs a common code, and runs an acutual application (runApplication() function).
Now we need to test it carefully.

CHANGELOG.md
```diff
@@ -79,6 +79,7 @@
     Feature #3276: Editor: Search- Show number of (remaining) search results and indicate a search without any results
     Feature #3641: Editor: Limit FPS in 3d preview window
     Feature #3703: Ranged sneak attack criticals
+    Feature #4012: OpenMW-CS: Create log file like with OpenMW
```
Capostrophic: Changelog already has this line, though in wrong oder. I'd prefer that you don't touch the changelog because then I'd have to solve merge conflicts in the neighboring PR again (I placed the line where it belongs there).

PR #1844, opened 2018-08-04
Make forcegreeting a non-op for non-actor objects

akortunov:
Fixes [bug #4553](https://gitlab.com/OpenMW/openmw/issues/4553).
I encountered it once in unpatched Arktwend.

Note: maybe we have a better way to print warning rather than cerr.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,10 +77,11 @@
     Bug #4539: Paper Doll is affected by GUI scaling
     Bug #4545: Creatures flee from werewolves
     Bug #4551: Replace 0 sound range with default range separately
+    Bug #4553: Forcegreeting on non-actor opens a dialogue window which cannot be closed
     Feature #2606: Editor: Implemented (optional) case sensitive global search
     Feature #3083: Play animation when NPC is casting spell via script
     Feature #3103: Provide option for disposition to get increased by successful trade
-    Feature #3276: Editor: Search- Show number of (remaining) search results and indicate a search without any results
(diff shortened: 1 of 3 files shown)


Capostrophic:
I've never seen cerr being used in [whatever]extensions. In scripting system you're supposed to throw an exception so it could be handled as a compilation error. For example,
```cpp
if (cell->getCell()->isExterior())
    throw std::runtime_error("Can't set water level in exterior cell");
```
is used in SetWaterLevel.

Capostrophic:
```the whole script will not fail in this case?```
It will. That's the idea. Anyway, seems that cerr is actually used in PositionCell, so it's fine.
```cpp
                    try
                    {
                        store = MWBase::Environment::get().getWorld()->getInterior(cellID);
                    }
                    catch(std::exception&)
                    {
                        // cell not found, move to exterior instead (vanilla PositionCell compatibility)
                        const ESM::Cell* cell = MWBase::Environment::get().getWorld()->getExterior(cellID);
                        int cx,cy;
                        MWBase::Environment::get().getWorld()->positionToIndex(x,y,cx,cy);
                        store = MWBase::Environment::get().getWorld()->getExterior(cx,cy);
                        if(!cell)
                        {
                            std::string error = "Warning: PositionCell: unknown interior cell (" + cellID + "), moving to exterior instead";
                            runtime.getContext().report (error);
                            std::cerr << error << std::endl;
                        }
```
note the getContext().report() usage, it allows to use to know where did the warning happen.

akortunov:
Thanks. Now I display a warning to OpenMW console too.

PR #1847, opened 2018-08-05
[PoC, Discussion] Improve console logging

akortunov:
Just a proof of concept for now.
For now, we have a quite simple logging system with these limitations:
1. No debug levels. It means a lot of annoying and not very useful messages in output, even without a lot of mods.
2. No coloured highlighting, even if current terminal supports it.
3. cout/cerr are not threadsafe, so output can be garbaged.

This PR for now provides a prototype which adresses issues 1 and 2.

For example, we have this block in source code:
```
    std::cout << Misc::Warning << "Something went wrong." << std::endl;
    std::cout << Misc::Error << "Something went REALLY wrong." << std::endl;
    std::cout << Misc::Info << "Just some additional info, move along." << std::endl;
    std::cout << "A message with unspecified level." << std::endl;
```

By default, there will be something like this on console output:
screenshot_20180805_161645
Note: since the default debug level is "Warning", info message will be ignored.

You can specify debug level manually:
```
OPENMW_DEBUG_LEVEL=INFO ./openmw
```
Then info message will not be filtered:
screenshot_20180805_161410

Or you can disable coloured output:
```
NO_COLOR=1 ./openmw
```
Result:
screenshot_20180805_161715

So my questions:
1. Can we use something like this?
2. If yes, can anyone suggest a better approach?
3. How can we implement a threadsafe logging?


kcat:
Personally I'd prefer something that can optimize a little better. With this suggestion, a trace like this:
```
std::cout << Misc::Info << "Just some info, about " << a.something() << " and " << b.something_else() << std::endl;
```
will have all that stuff called (including all the `operator<<` invocations) even when no info logging is going to be written out, unnecessarily slowing down execution. IMO, a better option might be something like this:
```
LOGOUT(Misc::Info) << "Just some info, about " << a.something() << " and " << b.something_else() << std::endl;
```
where
```
// or however the current log level is queryable
#define LOGOUT(Level) if(LogLevel >= Level) std::cout<<Level
```
which ensures nothing for the log message is called if no log message is being written. With certain compiler builtins, this could be improved a bit more:
```
#define LOGOUT(Level) if(__builtin_expect(LogLevel >= Level, false)) std::cout<<Level
```
which will optimize for the case where the log message isn't generated, ensuring the generated code runs better when nothing abnormal is going on. On the whole, you aren't going to generate many log messages since warning and error messages should happen relatively rarely, while a full info log is probably going to be more I/O bound than anything to really feel an effect from optimizations.

akortunov:
Thanks.
Can you suggest a way to avoid issues with multimple threads?
For example, now we have something like this:
```
// thread1
cerr << "Warning: " << "message 1" << std::endl;

// thread2
cerr << "Warning: " << "message 2" << std::endl;
```

So instead of:
```
Warning: message 1
Warning: message 2
```

We can get something like:
```
Warning: Warning: message 1
message 2
```

As a blunt solution, we can use stringstream to construct message, and only then send it to logger as a single chunk.

akortunov:
For now I decided to use a macro with stringstream to write the whole message as a single chunk.

akortunov:
Looks like MSVC does not understand the __builtin_expect.

akortunov:
Should be a bit cleaner now.

components/misc/debugging.cpp
```diff
@@ -0,0 +1,68 @@
+#include "debugging.hpp"
+
+namespace Misc
+{
+    DebugLevel Error = 0xff;
+    DebugLevel Warning = 0xfe;
+    DebugLevel Info = 0xfd;
+    DebugLevel Default = 0;
+    DebugLevel DebugLevelMarker = Info;
```
kcat: These would probably be better as `constexpr`, since they should be compile-time constants.

kcat:
> Can you suggest a way to avoid issues with multiple threads?

My over-engineering brain says you could use a temporary object wrapper with a lock. Something like

```
class LogWrapper {
    static std::mutex sLock;

    std::unique_lock<std::mutex> mLock;
public:
    // Locks a global lock while the object is alive
    LogWrapper(DebugLevel level) : mLock(sLock) { std::cout << level; }

    // Perfect forwarding wrappers to give the chain of objects to cout
    template<typename T>
    const LogWrapper& operator<<(T&& rhs) const
    { std::cout << std::forward<T>(rhs); return *this; }
    template<typename T>
    const LogWrapper& operator<<(const T& rhs) const
    { std::cout << std::forward<T>(rhs); return *this; }
};
```
Then used as:
```
#define LOG(L) if(Misc::CurrectDebugLevel >= L) LogWrapper(L)
...
// If the current log level is high enough, a LogWrapper object is constructed that
// takes the global log lock, and is passed along the << chain until the ; at the end
// of the statement where it goes out of scape, where it's destructed and releases
// the lock, thus ensuring only a single thread is writing to the log at a time.
LOG(Misc::Info)<< "My info message" <<std::endl;
```

This could be extended a bit too to have the `LogWrapper` destructor auto-append `std::endl` to `std::cout`, so callers don't have to:
```
class LogWrapper {
   ...
   ~LogWrapper() { std::cout << std::endl; }
   ...
};
...
LOG(Misc::Info)<< "My info message";
```

> Looks like MSVC does not understand the __builtin_expect.

Yes, it's a GCC thing (also supported by Clang).

zinnschlag:
I am not a fan of this kind of preprocessor magic. Looks like asking for trouble. The use of the preprocessor should be avoided whenever possible or at least minimised.

Also, why is the debug level taken from an environment variable? Sounds more like a case for a cfg-file/command line option.

Finally, I suggest replacing the enum class (Level) with a plain enum. That will save the Level:: part and make the log statements less verbose.

@kcat Sorry, but I don't agree with your statement on optimisation. Unless a log statement is placed somewhere in a tight inner loop (a bad idea to begin with because that means the log will get spammed to death if there actually is any logging to be done) optimisation is irrelevant here. The performance impact of a log statement in the startup code is non-existent. A log statement at top level of the main loop will do diddly squat to the performance. And most log statements are located in if or else branches that are only executed in an error situation anyway.

akortunov:
So basically I should:
1. Implement Log function instead of macro.
2. Use config setting for log level instead of environmental variable.
3. Use enum for log levels. But I did not get how it supposed to save the "Level::" part.

Right?

zinnschlag:
Regarding 1: Not sure. Logging is complicated. There may be no reasonable way around some macros. But at least we should avoid such acrobatics where the first half of a statement is generated by the macro and the second half is written in the source file.

Regarding 2: Correct.

Regarding 3:

``
enum class A { X }; enum B { X };
``

In the first case you reference X via A::X. In the 2nd case (old style enum) you reference X via X instead of B::X. Saves the B:: part.

kcat:
> Implement (inline?) Log function instead of macro.

You could just rename the `LogWrapper` class name and forgo the macro. Include a boolean field in the class that's set depending if the log level is valid for the request. Then just check that bool before writing to `std::cout`.

> Use config setting for log level instead of environmental variable.

I'd recommend leaving the environment variable. If I want to quickly check something and need it logged, it's a lot simpler to just run
```
OPENMW_DEBUG_LEVEL=INFO ./openmw
```
instead of editing openmw.cfg to enable the info logging option, running openmw, then remembering to edit openmw.cfg again to disable info logging. Might just be because I'm a heavy command line user, but for temporary settings it's easier to just set an environment variable like that when running than to mess about with config files.

> Use enum for log levels. But I did not get how it supposed to save the "Level::" part.

If you do:
```
namespace Debug {
   enum Level { // not enum class Level
      Info, Warning, ...
   };
}
```
you reference them like `Debug::Info`, `Debug::Warning`, etc. You can include the `Level::` portion when referencing them, but for a normal enum (non-class) it's optional.

zinnschlag:
> If I want to quickly check something and need it logged, it's a lot simpler

Not really, since all cfg options are also available as command line options.

./openmw --log-level=1 (or something like that) is quicker than dealing with environment variables.

kcat:
> Not really, since all cfg options are also available as command line options.

Fair enough, I forgot about that.

akortunov:
Use enum for log levels.
About configs: currently we setup logger before we read config files. An additional investigation needed.

zinnschlag:
It seems we only log two lines before reading config. On my system:

> Loading config file: /home/marc/.config/openmw/openmw.cfg... done.
> Loading config file: ./openmw.cfg... done.

That's not too bad. Maybe remove logging from that and log the loaded config files after logging has been set up?

akortunov:
Do we have command-line arguments for editor?

akortunov:
Looks like command-line arguments is not and option for us since only game itself supports it for now.
So for editor we still sticked with environmental variable.

zinnschlag:
Not sure what kind of comment you expect in regards to the colours. Pretty, maybe?

OpenMW-CS doesn't have command line arguments, mostly because so far we didn't need any. I think it would kinda make sense to add command line support to OpenMW-CS. However that is quite a bit of extra work. I don't insist on the command line solution, if you would prefer to not go out of your way to handle this task.

Speaking of extra work: The new log system seems mostly alright now, but it is of little use as long most logging is still using the old one. Do you plan change all logging to the new system when you are done setting it up? Lot's of additional work. I would be okay with merging this PR without it for now (might make a good beginner task).

kcat:
> Not sure what kind of comment you expect in regards to the colours. Pretty, maybe?

I guess he means if the colors seem appropriate for the debug levels. They seem fine to me.

akortunov:
> I guess he means if the colors seem appropriate for the debug levels.

Yes, your guess is correct.

> Do you plan change all logging to the new system when you are done setting it up?

Yes, but as a separate PR - this one is about logging system itself.
Also I will start a forum discussing to decide which messages should be enabled by default, and which should not.

akortunov:
Added configs for debug levels.
But I am still not sure if it is a good idea.
Since OpenMW-CS does not use command-line arguments at all, and the editor and the game itself use different implementations of settings, case "config file + command line argument" requires 4 different ways to configure just one variable, when old "environmental variable" case requires only one way, common for all apps.

apps/opencs/model/prefs/state.cpp
```diff
@@ -164,11 +164,18 @@ void CSMPrefs::State::declare()
     declareColour ("colour-comment", "Highlight Colour: Comments", QColor ("green"));
     declareColour ("colour-id", "Highlight Colour: IDs", QColor ("blue"));
 
-    declareCategory ("General Input");
+    declareCategory ("General");
     declareBool ("cycle", "Cyclic next/previous", false).
         setTooltip ("When using next/previous functions at the last/first item of a "
         "list go to the first/last item");
 
+    EnumValue defaultLogLevel ("Info", "Show infolmational messages");
```
Capostrophic: informational

zinnschlag:
Regarding the colours: As I said, pretty. Though you probably shouldn't ask me about anything related to that. I have a terrible taste/lack of taste when it comes to colours (see the editing script syntax-colouring scheme as an example).

The default debug level must be maximum logging. When we request logs from a user coming to us with a problem, we don't want to guide him through the process of changing the log level (not to mention bugs that aren't easily reproducible).

But I think we need to take a step back here. This is getting messy. The proposal was to use command line options. Not user settings. If you really don't want to go down this implementation path we can as well stick with environment variable only.

akortunov:
Reworked.
Enabled verbose level by default, configuring via environmental variable.

zinnschlag:
Looks good to me. Well, except for the macro. But I don't have any idea how to improve on that. Wanna try it or shall we go ahead with the merge?

akortunov:
> But I don't have any idea how to improve on that.

Should I replace this macro to inline function?

components/debug/debuglog.hpp
```diff
@@ -18,32 +18,64 @@ namespace Debug
...
+    Debug::Level mLevel;
+    bool mWritten;
+
+    void WriteDebugLevel()
+    {
+        if (!mWritten)
+        {
+            std::cout << static_cast<unsigned char>(mLevel);
+            mWritten = true;
+        }
```
kcat: I don't think this is necessary, (re)writing the log level for every `<<` call and needing to track if something was written for the destructor. Just checking if the log level checks out prior to writing anything should be good enough.

akortunov: I do not understand what "Just checking if the log level checks out" means.

kcat: Doing in the constructor:
```
if(mLevel <= Debug::CurrentDebugLevel)
    std::cout << mLevel;
```
For the destructor:
```
if (mLevel <= Debug::CurrentDebugLevel)
    std::cout << std::endl;
```
And for writing the objects:
```
if (mLevel <= Debug::CurrentDebugLevel)
    std::cout << std::forward<T>(rhs);
```
There's no need to keep writing the debug level for every object, nor any need to keep a track of a `mWritten` variable.

akortunov: Done.

components/debug/debugging.hpp
```diff
@@ -0,0 +1,129 @@
...
+    enum Color
+    {
+        Reset = 0,
+        LightGray = 37,
+        DarkGray = 90,
+        Red = 91,
+        Yellow = 93
+    };
+
+    static bool checkColorTerminal()
```
zinnschlag: This static looks odd. Shouldn't that be an inline instead?

akortunov:
> the two issues I mentioned above are taken care of.

Done.

psi29a:
Roger that, kicking travis in the head to restart.

PR #1848, opened 2018-08-05
Don't render NiTriShapes without NiTexturingProperty (bug #4483)

Capostrophic:
[Bug 4483](https://gitlab.com/OpenMW/openmw/issues/4483).

If the shape has no textures bound to it, it's nodemask is set to 0x1 so it's hidden from view.
I'm not sure if I didn't break something, so please test it. Sprite trees from Better flora from the report seem to work better now - the untextured shape doesn't flicker in and out of view.


AnyOldName3:
According to https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwrender/vismask.hpp, 0x1 is `Mask_UpdateVisitor`.

If you want something not to be rendered, set its mask to zero. Don't pick random values which may mean something somewhere else. :)

PR #1849, opened 2018-08-05
 Make AiActivate not stop upon target activation (bug #4456)

Capostrophic:
[Bug 4456](https://gitlab.com/OpenMW/openmw/issues/4456).

Aaaand now it's broken. Hooray.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
     Bug #4452: Default terrain texture bleeds through texture transitions
     Bug #4453: Quick keys behaviour is invalid for equipment
     Bug #4454: AI opens doors too slow
+    Bug #4456: AiActivate should not be cancelled after target activation
     Bug #4457: Item without CanCarry flag prevents shield autoequipping in dark areas
     Bug #4458: AiWander console command handles idle chances incorrectly
     Bug #4459: NotCell dialogue condition doesn't support partial matches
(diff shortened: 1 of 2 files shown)


akortunov:
Unfortunatly, it is not broken enough - vanilla AiActivate does not use pathfinding.
Also it would be nice to have optional flag to keep our current hehaviour.

Capostrophic:
>vanilla AiActivate does not use pathfinding.

I'm not entirely sure why should this be replicated. I suppose it's possible to try to use AiPackage::shortcutPath first and if it fails, fallback to pathfinding.

>it would be nice to have optional flag to keep our current behaviour

Zini implied that the proper implementation of AiActivate is post-1.0, and that perhaps "broken" AiActivate will be the prerogative of mwscript.

akortunov:
> I'm not entirely sure why should this be replicated.

Then this PR is not really useful. It even does not fix original OP issue since actor's destination is blocked.

Capostrophic:
>since actor's destination is blocked.

And how exactly would getting rid of pathfinding would help in this case? The problem is that the path tolerance is too low for large objects, but that ceased to be the point of this specific bug and it can't be fixed in proper before there is a way to recover the mesh dimensions.

akortunov:
> The problem is that the path tolerance is too low for large objects, but that ceased to be the point of this specific bug and it can't be fixed in proper before there is a way to recover the mesh dimensions.

It is up to modders to handle this case.

> And how exactly would getting rid of pathfinding would help in this case? 

Originally Mages Robes mod relied on fact that actor with AiTravel package should mindlessly walk to target by straight line.
When actor is close enough to target, a scipt should manually cancel AiTravel package.
In OpenMW actor can not get close enough since destination is blocked and AI can not use shortcutting in this case.

Capostrophic:
I'll close this for now then.

PR #1851, opened 2018-08-07
Init SDL2 before Qt4 to avoid crash on Linux

akortunov:
Should fix [bug #4529](https://gitlab.com/OpenMW/openmw/issues/4529).

The main idea: init SDL2 before Qt4, as we do in 0.44. For Qt5 keep our upstream behaviour to solve issues with MacOS ([bug #2862](https://bugs.openmw.org/issues/2862) and [bug #3911](https://bugs.openmw.org/issues/3911)). Actually, we changed bahaviour because of them.

Notes: 
1. I do not have MacOS and Qt4, so I can not test these changes by myself. 
2. If MacOS also uses Qt4, probably it would be better to check for OS instead of Qt version.

### DIFF
--- a/apps/launcher/CMakeLists.txt
+++ b/apps/launcher/CMakeLists.txt
@@ -1,6 +1,7 @@
 set(LAUNCHER
     datafilespage.cpp
     graphicspage.cpp
+    sdlinit.cpp
     main.cpp
     maindialog.cpp
     playpage.cpp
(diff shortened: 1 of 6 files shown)


Thunderforge:
@akortunov Sorry that I forgot about this! Thank you for pinging me to remind me.

When using Qt5 and macOS 10.13, everything seems to be working correctly. I am not seeing key duplication or any of the other bugs that were previously happening due to SDL issues. So I think that things are okay here on a Mac.

Qt4 is not compatible with newer versions of macOS, so Qt5 is the only one that I can test. We've never really set minimum requirements, so I don't know if you can use an older version of macOS and Qt4.

zinnschlag:
Looks like we can merge then?

akortunov:
> Looks like we can merge then?

Yes, if it does not crash with Qt4.

PR #1852, opened 2018-08-07
Cap underwater view distance

akortunov:
Fixes [bug #4565](https://gitlab.com/OpenMW/openmw/issues/4565).

By design, underwater view distance should be limited, but this limitation in OpenMW is a dead code for now.
[Here](https://imgur.com/a/5XJCH4s) is the comparison (old/new behaviour).
Not sure if it worth a separate bugreport and entry in changelog.


zinnschlag:
> Not sure if it worth a separate bugreport and entry in changelog.

Yes, it is.

akortunov:
> Yes, it is.

Done.

PR #1853, opened 2018-08-07
Check the actor cell instead of the destination in fast travel price logic (bug #4563)

Capostrophic:
[Bug 4563](https://gitlab.com/OpenMW/openmw/issues/4563)

Making interior NPCs explicitly use guild guide logic and exterior NPCs use caravaneer logic seems to follow the original behavior exactly. Fixes the prices guild guides charge for exterior cell destinations. Yes, this means that guild guide NPCs that were placed to an exterior will work like silt strider caravaneers or shipmasters. This is how it works in Morrowind.

Interestingly when scrawl added the missing spellcast sound its behavior was correct.


terabyte25:
What about for NPCs who aren't Mage's Guild members? How does that work then?

PR #1854, opened 2018-08-08
Remove unnecessary link to osgAnimation

akortunov:
Scrawl wiped out dependency on osgAnimation library a year ago, but forgot to remove it from CMake config file.


psi29a:
We'll need this in the future to support collada animation...

akortunov:
Then feel free to close this PR.

psi29a:
For example: https://github.com/sbarthelemy/daenim  <-- we could in theory cut & paste from this with a little re-work.

PR #1855, opened 2018-08-08
Update appveyor.yml

psi29a:
allow msvc2015 to fail


psi29a:
Look at that... msvc2015 failed just like on queue:
https://ci.appveyor.com/project/psi29a/openmw/build/417

psi29a:
And we're still passing!

PR #1858, opened 2018-08-08
Do not use headtracking in the 1st-person view

akortunov:
Fixes [bug #4573](https://gitlab.com/OpenMW/openmw/issues/4573).


JDGBOLT:
Just so you know the commit 126b2fd broke aggro for NPC's and Creatures, unless you hit them first, which I am sure is not how it's meant to work.

PR #1859, opened 2018-08-08
Fix a bunch of MSVC warnings

Capostrophic:
Three streams were declared twice in a new debugging.hpp file and MSVC (AppVeyor) doesn't like it. I'm not sure if the duplicate declarations were accidental, but I decided to get rid of them just in case.


akortunov:
Just for reference: #1847 handles these warning (and moves affected function to another file).

PR #1863, opened 2018-08-10
Use the correct spell projectile speed GMST in AI aiming (bug #3948)

Capostrophic:
[Bug 3948](https://gitlab.com/OpenMW/openmw/issues/3948).
See the issue new description.

From my testing AI aims much better than before and the projectile will guaranteedly hit the player if one doesn't stop movement in the current direction.

Edit: just in case, yes, target spells have constant speed. There's no "min" GMST.


Capostrophic:
Is there anything wrong? Should I make AiCast account for target movement?

psi29a:
Zini is away for a few days and I'm not familiar enough with this area of code.

psi29a:
It looks ok and clean :)

I'll go with it and see if we get any bug reports. ;)

PR #1865, opened 2018-08-11
Fix race condition (bug #4584)

elsid:
Values are accessed from main thread and cell preloader thread. Fixes [#4584](https://gitlab.com/OpenMW/openmw/issues/4584).

### DIFF
--- a/components/terrain/material.cpp
+++ b/components/terrain/material.cpp
@@ -12,66 +12,150 @@
 
 #include <components/shader/shadermanager.hpp>
 
+#include <mutex>
+
+namespace
+{
+    class BlendmapTexMat
+    {
+    public:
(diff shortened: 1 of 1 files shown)


Capostrophic:
Nice. I've been having weird crashes recently that originated somewhere in the blendmap code.

psi29a:
@zinnschlag can you help me eye this?

components/terrain/material.cpp
```diff
@@ -12,66 +12,150 @@
...
 
+#include <mutex>
 
-namespace Terrain
+namespace
 {
+    class BlendmapTexMat
+    {
+    public:
+        static const osg::ref_ptr<osg::TexMat>& value(const int blendmapScale)
```
Aesylwinn: You should probably document these methods with static variables, saying something along the lines of "not thread safe."

PR #1866, opened 2018-08-12
Ease AI sequence restrictions in AI allies logic (bug #4304)

Capostrophic:
[Bug 4304](https://gitlab.com/OpenMW/openmw/issues/4304).

~~Since evidently checking for whether the **currently active** "real" AI package is AiFollow is not something vanilla does when finding allies of an actor and nobody argued against lifting this restriction OpenMW has, here's something that lifts it.~~

Since without researching how exactly vanilla behaves in certain edge cases when the matter touches finding actor allies lifting the restrictions leads to different issues, for now AiWander is considered another exception in AI ally finding logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
     Bug #4286: Scripted animations can be interrupted
     Bug #4291: Non-persistent actors that started the game as dead do not play death animations
     Bug #4293: Faction members are not aware of faction ownerships in barter
+    Bug #4304: "Follow" not working as a second AI package
     Bug #4307: World cleanup should remove dead bodies only if death animation is finished
     Bug #4311: OpenMW does not handle RootCollisionNode correctly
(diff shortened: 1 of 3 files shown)


Capostrophic:
>If you Command a follower, should he attack you or initial leader?

In Morrowind, the follower attacks the leader. The same happens in OpenMW (with these changes at least). Not sure how guaranteed this is, but presumably the more recent AiFollow package takes precendence.

Interestingly I got "Error in frame: Can't add AI packages to player" errors in OpenMW when testing this...

Capostrophic:
> If you Command a leader, what his followers should do?

In Morrowind they become allied to the player.

akortunov:
Just for reference: there is no need to push every single commit (it triggers new CI builds and causes e-mail notifications).
Keep them locally and push them once when you will finish to work on this PR.

PR #1869, opened 2018-08-13
[Do not merge, Needs research] Rework animated collisions systems

akortunov:
Related to [bug #4437](https://gitlab.com/OpenMW/openmw/issues/4437).

I found these issues with mentioned model feom Abot's Ships:
1. Looks like we apply scaling twice, so result is [weird](https://i.imgur.com/IgbQcVx.jpg) - generated mesh is not even with rendering mesh.
2. We ignore world transformation for skinned shapes, result is [not good](https://gitlab.com/OpenMW/openmw/uploads/f787b26c04d1500c52618e17772609b1/Screenshot_20180603_160640.jpg).
3. We do not consider skinned shapes as animated when generate collision shape. Result is [expected](https://i.imgur.com/CqbfUQS.jpg).

In its current state, this PR allows to [fix](https://i.imgur.com/KG5pKt6.jpg) the first issue, but I am not sure if I did not break anything.


Capostrophic:
In [Pearl Palace](http://mw.modhistory.com/download-45-12090) the eponymous palace has broken collisions in OpenMW - it might be related to this and need a look too.

akortunov:
> In Pearl Palace the eponymous palace has broken collisions in OpenMW - it might be related to this and need a look too.

Can you specify what exactly is going wrong?

Capostrophic:
>Can you specify what exactly is going wrong?

The collisions of the "underwater" palace (which is presumably moving?) are completely misaligned with the geometry itself. When you enter it, you fall down into the void, and if you try to proceed further with TCL and then enable collisions again, in the center room you bump into invisible walls if you try to move around. Seeing that there's something very wrong with the collision shapes is easy when TCG is used.

akortunov:
There is really something strange with this mod, but I can not find anything suspicious with palace_int.nif - this file has a flat hierarchy, just plain NiNodes and NiTriShapes.
And it is not an optimizer issue.

AnyOldName3:
I have some familiarity with our skinning system, and you have an OSG bug that Robert Osfield claims is a feature, but is 100% unarguably a bug to thank for that.

We have two types of animated meshes. The first is RigGeometry, which is skinned stuff, like hands and legs which bend according to a skeleton, and the other is MorphGeometry, which is morphed stuff like heads (which have lips that move during speech and eyes that blink) with no skeleton. As well as that, we seem to be able to make regular osg::Geometries children of skeleton nodes so, for example, if a bow is a child of a forearm node, it'll move as the forearm moves.

The following refers to the bounds used for the cull traversal, not the collision ones, as I don't know where those come from.

* Bounds for a RigGeometry are computed each frame during the update traversal based on the bounding spheres of all bones which affect the RigGeometry. They're probably quite big in comparison to the actual mesh.
* Bounds for a MorphGeometry seem to be computed whenever OSG requests a bounding box for the MorphGeometry and the bounds haven't been computed since a new MorphTarget was added. The value comes from working out the maximum possible position any vertex could end up in (without considering which morphs might be used at the same time). These are probably quite tight, as blinking doesn't change the shape of your face much.
* I think the bounds for a Geometry childed to a bone are computed in the same way as a Geometry childed to a bunch of transform nodes. these bounds are likely to be as good as the bounds for anything else.

I don't know if we want to have the collision shaped for things be the full quality mesh. In [this screenshot](https://imgur.com/CqbfUQS) we seem to be using a full-quality mesh for collision with an NPC, but I thought that we were using capsules, cylinders or AABBs (depending on which one @wareya gets the best results with). Is this NPC not actually a real NPC, but instead an animated static designed to look like an NPC?

Animating collision would basically involve either copying-and-pasting the skinning code from `RigGeometry::cull` (which I'm unconvinced is a great idea as it's duplicated code, and I'm pretty sure it uses a non-trivial amount of CPU time) or computing a new collision shape based on the child geometries produced by the cull method. I'm assuming we do collision in the update traversal, so we'd either need to use the previous frame's result, or somehow change the skinning system so it occurs during the update traversal. Either of these approaches is likely to remove the need for hardware skinning in the future, as we'd need the data in main memory, and that was something that didn't look like a bad idea to implement.

What do the collision shapes for these meshes look like in vanilla Morrowind?

akortunov:
@AnyOldName3, any ideas how to fix [bug #4437](https://gitlab.com/OpenMW/openmw/issues/4437)?
In theory, skinning via RigGeometry should take TriShape transformations and transformations of NiSkinData in account, but I just do not know where to apply these transformations.

AnyOldName3:
What's the structure of the original Nif? I've not had to work out what happens when there are transformations between the skeleton and the RigGeometry, but if that's what happens and we're just ignoring it, it would explain your issues.

akortunov:
Original NIF do not have animations at all.
Modified NIF from mod has transformations on [NiTriShape](https://i.imgur.com/WUitXMN.png) and [NiSkinData](https://i.imgur.com/rpBKhqF.png). If I discard these transformations, NifSkope will render this model [as OpenMW renders it now](https://i.imgur.com/138lBxp.png).
Both models are attached to mentioned bugreport.

AnyOldName3:
What would happen if the transformation was applied to the geometry when it was loaded, so its effect would have been already applied to the RigGeometry's source Geometry? I'm not sure what order the transformations get applied, but I don't imagine this would be so hard if it turns out that it works and there's nothing that would change the transformation at runtime.

akortunov:
> What would happen if the transformation was applied to the geometry when it was loaded, so its effect would have been already applied to the RigGeometry's source Geometry? 

I tried to use this hack in the NifLoader::handleSkinnedTriShape() function, but did not get any difference, but probably I just do not do it right. I tried to use osg::MatrixTransform and PAT.

AnyOldName3:
You won't be able to do it like that. I'm meaning manually apply the matrix to all of the vertices when they're loaded in the first place.

akortunov:
> I'm meaning manually apply the matrix to all of the vertices when they're loaded in the first place.

Result as with my [original hack](https://gitlab.com/OpenMW/openmw/issues/4437#note_82159628) for mInvBindMatrix: [animation still works in opposite direction](https://i.imgur.com/Lw8Nvkb.jpg), also sails disappear under certain angles.

akortunov:
Also sails [look fine](https://i.imgur.com/bsEbERN.jpg) if I comment [this loop](https://github.com/OpenMW/openmw/blob/openmw-44/components/sceneutil/riggeometry.cpp#L215).
But obviously there are no animations in this case.

akortunov:
I re-purposed this PR to fix only first bug. Ready for review now.

PR #1870, opened 2018-08-13
Make sure bip01 root node is not optimized away (bug #3856)

Capostrophic:
[Bug 3856.](https://gitlab.com/OpenMW/openmw/issues/3856)

Implements the "easy" fix suggested by scrawl: don't optimize root nodes named bip01. I think this should do it for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
     Bug #3533: GetSpellEffects should detect effects with zero duration
     Bug #3591: Angled hit distance too low
     Bug #3629: DB assassin attack never triggers creature spawning
+    Bug #3856: Bip01 is not found on creatures, even if it is present
     Bug #3876: Landscape texture painting is misaligned
     Bug #3897: Have Goodbye give all choices the effects of Goodbye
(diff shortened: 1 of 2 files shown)


akortunov:
A quite strange behaviour since the bip01 has the reserved name and should not be optimized anyway.

Capostrophic:
>bip01 has the reserved name and should not be optimized anyway.

Here's the fun part: nif loader level optimizations aren't supposed to care about reserved names.

> Did you test this fix?

You got me, and it appears that the first statement here is not true anymore: I haven't noticed any difference in master and in a branch related to this when creating a worm mouth creature using the assets from the report, but in 0.43.0 it is definitely broken and the creature falls down through the terrain. It might have been the cause of the reports about creatures that clip through the terrain even though they definitely have a bounding box (I tested velks some time ago and they were just fine).

psi29a:
Is the bug still relevant?
Or is it no longer reproducible (won't fix) ?

akortunov:
According to Capostrophic, this bug was fixed in 0.44 release.

PR #1874, opened 2018-08-15
Fix some issues with attack animations

akortunov:
1. Fixes a recent regression, when player can not attack with melee weapon, if he shoot from bow before.
2. Fixes [bug #4591](https://gitlab.com/OpenMW/openmw/issues/4591), so attack strength calculation for player should be more corrent.
3. Since some attack animation contain movement, add a threshold when turning to target in melee combat - we use cone shapes anyway.
4. In vanilla turning animations do not reset idle ones. Replicate this behaviour.
3 and 4 together fix [bug #4371](https://gitlab.com/OpenMW/openmw/issues/4271).
5. Since we update mIdleState directly in character controller, we should update idle animations also when mIdleState == CharState_None. Fixes [bug #4531](https://gitlab.com/OpenMW/openmw/issues/4531).
6. We should not reset idle animations in the spellcasting stance.
7. Fix incorrect spelling of "preparing" word.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@
     Bug #4503: Cast and ExplodeSpell commands increase alteration skill
     Bug #4510: Division by zero in MWMechanics::CreatureStats::setAttribute
     Bug #4519: Knockdown does not discard movement in the 1st-person mode
+    Bug #4531: Movement does not reset idle animations
     Bug #4539: Paper Doll is affected by GUI scaling
     Bug #4545: Creatures flee from werewolves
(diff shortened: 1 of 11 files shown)


akortunov:
Goblins now behave better than in 0.44 too. Nice.

akortunov:
Tested by me and Capostrophic, no severe issues found yet.

psi29a:
Fixed a merge conflict... this one ready? :)

Capostrophic:
Unless you can find a regression somehow, go ahead.

PR #1876, opened 2018-08-18
[Discussion] Allow messageboxes arguments to have newline characters

akortunov:
Fixes [bug #3836](https://gitlab.com/OpenMW/openmw/issues/3836).
Some background:
TES-CS does not allow to put "\n" characters in your text output, so you can not make multiline messageboxes - editor will not compile your script. But there is a workaround:
1. Use supported character as separator ("|", for example).
2. Compile your script.
3. Save your mod.
4. Replace "|" to 0x10 via HEX-editor (in compiled bytecode too).
As result, Morrowind will display your multiline messagebox.
But OpenMW does not use this bytecode, and fails to compile such script.
Russian Tribunal addon uses this quirk for "A Star Is Born" quest (a theatre in Mournhold).
It means that this quest is broken for many players, and we need to solve this issue somehow since it affects clean installation without mods.

How this fix works:
1. Check if the current scanned name is a "messagebox" keyword.
2. If yes, set a special flag.
3. If this flag is true and newline character encountered, print a warning instead of error.
4. If the scanned name is another keyword, set this flag to false.

Add a "Discussion" label since I am not sure if there are better options to solve this issue.


AnyOldName3:
Maybe support for some kind of multi-line string or usage of `\n` within a string literal (as in a `\` and then an `n` rather than an actual line feed) would be a better idea. It would need a patch to be made for Russian Morrowind, but it feels less hacky to me.

akortunov:
Here is a better description: http://en.uesp.net/wiki/Tes3Mod:Formatted_MessageBoxes

zinnschlag:
Okay. So | then? In this case your fix wouldn't actually improve compatibility. If content files that use this trick don't standardise on |, maybe we could supply a list of characters in the openmw.cfg file that are used as replacement for newline?

As for the implementation, I believe there is a clearer way to do it that is also more in line with what we do elsewhere. But I need to look into that first, before I can comment on it. In 1-2 days maybe.

zinnschlag:
If you look at the Scanner class, you will see the functions enableStrictKeywords and enableTolerantNames. That is the way to do it. Add a public function to the scanner that sets a flag that is automatically unset after scanning. Then call the function from respective parser function.

akortunov:
> If you look at the Scanner class, you will see the functions enableStrictKeywords and enableTolerantNames. That is the way to do it. Add a public function to the scanner that sets a flag that is automatically unset after scanning. Then call the function from respective parser function.

Done. 
Note: obviously I can not reset added flag when \n encountered, because I need to treat it as a part of console command. Also there could be several \n characters in line.

zinnschlag:
Okay, this implementation looks reasonable. What about my other point? Checking for \n instead of either a fixed or configurable line separator seems a bit pointless, since it doesn't improve compatibility.

akortunov:
I tested how \n works. Morrowind really displays "\n", OpenMW treats it as an escape sequence for new line, at least on Linux. It seems other escape sequences do not work.
If it works in the same way on Windows and MacOS, we can print a warning if 0x0a is encountered in messagebox (what this PR does now) and suggest to use the \n instead.
Other separators we can treat as we treat them now since we can not really tell if we should display character or replace it to \n.

zinnschlag:
Sounds reasonable.

PR #1880, opened 2018-08-21
Play landing sound manually and ignore land soundgen textkeys (bug #2256)

Capostrophic:
Yet another "landing audio candidate fix"!

[Bug 2256](https://gitlab.com/OpenMW/openmw/issues/2256).

I'll just copy my recap on the origin of the discrepancy:

> Apparently Morrowind ignores the landing sound soundgen textkey and plays the sound manually anyways, *in addition* to not interrupting the landing animation by a new jumping animation. [This diff] manually plays the landing sound immediately upon the beginning of the landing state (NB: about as early as we can do it), making the delay between what GetPCJumping considers jumping and the landing sound 2 frames and rendering both 4462 and this a non-issue, but it doesn't make landing animation play simultaneously with a new jumping animation.
Doublechecked it. Land Soundgen is never used for playing a sound [...] when an NPC uses an animation with it [...]. Landing sound does play [upon landing] even if the landing animation is interrupted before it can even start playing. Dammit, Bethesda.

So, what I did: made the character controller play the landing manually instead when an actor changes to the landing state. This applies to both creatures and NPCs. The land animation textkey is ignored like Morrowind appears to do.

This reduces the delay between the GetPCJumping = 0 after a jump finishes and the landing sound to exactly two frametimes. I don't know if it's possible to reduce it any further.

I didn't replicate the quirk of not interrupting the landing animation with a jumping animation. I don't think it's necessary. It might be even harmful.

Suggestions welcome.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -913,8 +913,7 @@ void CharacterController::handleTextKey(const std::string &groupname, const std:
         sndMgr->playSound3D(mPtr, evt.substr(7), 1.0f, 1.0f);
         return;
     }
-    if(evt.compare(0, 10, "soundgen: ") == 0
-     && evt.compare(10, evt.size()-10, "land") != 0) // Morrowind ignores land soundgen for some reason
+    if(evt.compare(0, 10, "soundgen: ") == 0)
     {
         std::string soundgen = evt.substr(10);
 
@@ -939,11 +938,14 @@ void CharacterController::handleTextKey(const std::string &groupname, const std:
             }
         }
 
+        if (soundgen == "land") // Morrowind ignores land soundgen for some reason
+            return;
+
         std::string sound = mPtr.getClass().getSoundIdFromSndGen(mPtr, soundgen);
         if(!sound.empty())
         {
             MWBase::SoundManager *sndMgr = MWBase::Environment::get().getSoundManager();
-            if(evt.compare(10, evt.size()-10, "left") == 0 || evt.compare(10, evt.size()-10, "right") == 0)
+            if(soundgen == "left" || soundgen == "right")
             {
                 // Don't make foot sounds local for the player, it makes sense to keep them
                 // positioned on the ground.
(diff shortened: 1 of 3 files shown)


Capostrophic:
Actually, you know what... It's possible to do it cleaner.

Capostrophic:
All of them. Nothing plays.

zinnschlag:
@kcat What do you think? Should we merge this?

kcat:
Looks alright to me. You could perhaps move the `evt.compare(10, evt.size()-10, "land") != 0` check earlier (before looking up the `"land"` soundgen sound since the result won't be used anyway), but I probably wouldn't block the PR for that.

kcat:
How does vanilla behave when a land soundgen is given an optional volume and/or pitch value? This will now play them in that case.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -914,7 +914,7 @@ void CharacterController::handleTextKey(const std::string &groupname, const std:
         return;
     }
     if(evt.compare(0, 10, "soundgen: ") == 0
-     && evt.compare(10, evt.size()-10, "land") != 0) // Morrowind ignores land soundgen for some reason
+     && evt.compare(10, 4, "land") != 0) // Morrowind ignores land soundgen for some reason
```
kcat: This will catch anything that starts with "land", including "landing" or whatever else is possible for a soundgen entry.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -943,7 +943,7 @@ void CharacterController::handleTextKey(const std::string &groupname, const std:
         if(!sound.empty())
         {
             MWBase::SoundManager *sndMgr = MWBase::Environment::get().getSoundManager();
-            if(evt.compare(10, evt.size()-10, "left") == 0 || evt.compare(10, evt.size()-10, "right") == 0)
+            if(evt.compare(10, 4, "left") == 0 || evt.compare(10, 5, "right") == 0)
```
kcat: Same here, it'll catch anything starting with left or right. Also I'm not sure if it's safe to specify an offset+length that can go past the end of the `evt` string.

Capostrophic:
Oops. This won't work properly either, I'll try again. Sorry.

psi29a:
Merging, thanks!

PR #1881, opened 2018-08-21
[Regression] Fix regressions in CharacterController

akortunov:
Fixes some regressions in my recent animation changes:
1. Extend 1h-fallback for crossbows to work with swim animations too, so fix for [bug #4495](https://gitlab.com/OpenMW/openmw/issues/4495) should be complete now.
2. Actually select random attacks for creatures (attack1-3), if they do not have animations for suitable weapon (the fix for [bug #4470](https://gitlab.com/OpenMW/openmw/issues/4470)). This change was erroneously removed.
3. Do not modify the mIdleState directly, modify it only during idle animations update.
We need to do it because of "idle != mIdleState" check in refreshIdleAnims().
Allows to fix a regression with crossbow movement animations.
4. Use turning animations only for bipedal actors. It seems vanilla behave in this way.
Note: also only bipedal actors use animations for side movement (in combat).

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -243,7 +243,7 @@ std::string CharacterController::chooseRandomGroup (const std::string& prefix, i
     return prefix + toString(roll);
 }
 
-void CharacterController::refreshHitRecoilAnims()
+void CharacterController::refreshHitRecoilAnims(CharacterState& idle)
 {
     bool recovery = mPtr.getClass().getCreatureStats(mPtr).getHitRecovery();
     bool knockdown = mPtr.getClass().getCreatureStats(mPtr).getKnockedDown();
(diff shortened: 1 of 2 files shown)


akortunov:
It seems works fine.

Capostrophic:
It might be a stupid thought (it's 4 AM), but since turning animations are not used for non-bipedal actors now, maybe there shouldn't be a smoothTurn threshold in AiCombat::rotateActorOnAxis for them?

akortunov:
> It might be a stupid thought (it's 4 AM), but since turning animations are not used for non-bipedal actors now, maybe there shouldn't be a smoothTurn threshold in AiCombat::rotateActorOnAxis for them?

They already do not use it.

Capostrophic:
This should be taken a look on sooner than later.

zinnschlag:
@Capostrophic Not familiar enough with this part of OpenMW to make a solid judgement. What is your take on this? Ready for merge?

Capostrophic:
*Should* be fine.

PR #1882, opened 2018-08-22
[Discussion] Optimize skinning

akortunov:
Currently most of our culling time is actually a skinning (RigGeometry::cull() function).
Since this function is critical for performance, I tried to optimize it.
What I did:
1. Use floats for transform3x3 instead of doubles since we use Matrixf anyway.
2. Place the accumulateMatrix() inline since we call it for every vertex of every bone of every animated object every frame.
3. Cache getTraversalNumber() instead of retrieve it three times.

On my hardware (i7-7700) execution of this function now takes nearly 30% less time, and overall culling update phase takes about 15-25% less time, depending of count of actors in scene.

Is there additional ways to optimize this function?


psi29a:
How CPU intensive are these calculations, is an individual calculation 'heavier' than the overhead passing this workload off to a thread then waiting for a .join()?

We can setup a worker threadpool during startup, just for fun... <nproc threads>, just to have them ready for use.

I'm just worried that the overhead in using thread and passing values to and from is more costly than performance gained in splitting the work up across CPUs.

akortunov:
> How CPU intensive are these calculations, is an individual calculation 'heavier' than the overhead passing this workload off to a thread then waiting for a .join()?

I am not sure if we can just move individual calculation to thread - we update skinning during Cull phase, when cull visitor iterates over scene graph.

For me skinning takes more than 50% of whole culling update and takes more time than physics calculations.

akortunov:
> Perhaps not individual calculation, perhaps batched or even per animation or group of animations?

I still do not understand what do you mean by "batching" here.
Currently we do skinning when CullVisitor encounters the RigGeometry node (the RigGeometry::cull(osg::NodeVisitor* nv) call).
I asked if it is possible to do this stuff in the separate thread instead of Culling pass in the main thread.

AnyOldName3:
The problem of doing it anywhere other than the cull traversal is that we'd end up skinning things outside the viewing frustum, so potentially increase the workload. My personal preference would be to switch to hardware skinning and do it in a vertex shader, but if we need collision shapes to be skinned, too, that doesn't work as we'd still need to do the skinning on the CPU for that, often for the same mesh. It's possible that doing it with a compute shader will be faster than what we're doing right now, but then we're transferring things to and from the GPU more often, so it might not be.

When I looked at the implementation, it seemed ripe for vectorisation, but I don't know how good modern compilers are at detecting that and using SIMD instructions. It's also possible that some loop unrolling or something like that will help.

Also, while we're on the subject, do we know if the result of skinning is recycled when we have RTT versions in the same traversal (e.g. with water reflections and shadow maps)? I think I checked this and came up with an answer, but I can't remember what it was if I did.

akortunov:
I tried to use SSE, but did not get a performance gain. It seems GCC is already do this job for me.
I see no other possible optimizations for software skinning, so remove the Discussion label.

zinnschlag:
We need an issue for this PR (Task label).

akortunov:
> We need an issue for this PR (Task label).

https://gitlab.com/OpenMW/openmw/issues/4605

components/sceneutil/riggeometry.cpp
```diff
@@ -141,7 +141,7 @@ bool RigGeometry::initFromParentSkeleton(osg::NodeVisitor* nv)
     return true;
 }
 
-void accumulateMatrix(const osg::Matrixf& invBindMatrix, const osg::Matrixf& matrix, float weight, osg::Matrixf& result)
+inline void accumulateMatrix(const osg::Matrixf& invBindMatrix, const osg::Matrixf& matrix, const float& weight, osg::Matrixf& result)
```
elsid: Better put this function into anonymous namespace. C++ standard does not say how to make an assembly code with `call` or without for function with `inline` specifier. But for optimizers anonymous namespace tells that this function will not be accessed from any other translation unit. This function is declared and defined only in this file, so to avoid any name conflicts anonymous namespace is more usefull. You can try to use [compiler explorer](https://github.com/mattgodbolt/compiler-explorer) to check compiler assembly output and compare it for different approaches (for openmw only local setup can be used). However there is no difference between `inline` and anonymous namespace for clang 6 and gcc 8 output, so right now there is no any problem in this code.

Reference to float doesn't have any effect if function is already inlined. But for not inlined it leads to one additional `push` to save one more register. diff

akortunov: Done.

PR #1884, opened 2018-08-23
[WIP] Load animations from NIF files

akortunov:
Attempt to fix [bug #4564](https://gitlab.com/OpenMW/openmw/issues/4564).
Main idea: if the custom animation mesh is attached to NPC, but it does not have the KF-file, load animations from attached NIF file.

For now this PR works only for bodyparts, which have skeletons in KF files: animations itself work, but bodyparts without skeletons have [wrong orientation](https://i.imgur.com/iQ505N8.png). Especially noticable with heads and pauldrons.
@kcat, @AnyOldName3, any ideas, what am I doing wrong?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@
     Bug #4557: Topics with reserved names are handled differently from vanilla
     Bug #4558: Mesh optimizer: check for reserved node name is case-sensitive
     Bug #4563: Fast travel price logic checks destination cell instead of service actor cell
+    Bug #4564: OpenMW does not load animations from custom NIF files attached to NPC
     Bug #4565: Underwater view distance should be limited
     Bug #4573: Player uses headtracking in the 1st-person mode
     Bug #4574: Player turning animations are twitchy
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -717,30 +717,33 @@ namespace MWRender
         }
     }
 
-    void Animation::addAnimSource(const std::string &model, const std::string& baseModel)
+    void Animation::addAnimSource(const std::string &model, const std::string& baseModel, bool useKfFile)
     {
         std::string kfname = model;
         Misc::StringUtils::lowerCaseInPlace(kfname);
 
-        if(kfname.size() > 4 && kfname.compare(kfname.size()-4, 4, ".nif") == 0)
-            kfname.replace(kfname.size()-4, 4, ".kf");
-        else
-            return;
+        if (useKfFile)
+        {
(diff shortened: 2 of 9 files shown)


kcat:
I'm not sure I understand the problem. An NPC definition can specify a NIF file, in which it'll load extra/override animations for that NPC. What does bodyparts have to do with it?

kcat:
Custom animations should already be working for NPCs. Check the dancers in the House of Earthly Delights in Suran. The three stage dancers have a custom NIF defining a unique dance animation (which overrides one of the default idle animations, I believe). I can't tell what the added code is trying to do since the behavior should already be there.

akortunov:
> The three stage dancers have a custom NIF defining a unique dance animation (which overrides one of the default idle animations, I believe).

Dancers have KF-file, which is the source of animations. Mannequins from Tamriel Rebuilt have only NIF-file.
IIRC, currently OpenMW can load NPC animations only from KF files since NIF files have different structure.

For example, mannequin is an NPC with tr_death_armordummy_a.nif animation file.
Since there is no xtr_death_armordummy_a.kf, OpenMW ignores the tr_death_armordummy_a.nif and uses basic animation files from xbase_anim.kf.
Morrowind in this case loads animations directly from tr_death_armordummy_a.nif **and does not load animations from xbase_anim.kf**.

kcat:
Hmm, well the technical structure of a nif and kf is the same (you should be able to load a .kf in NifSkope). I'd need to see the node structure of the TR NIF in question to get a better idea, but are you sure specifying a NIF for the NPC blocks animations from xbase_anim.kf, and there's not some other issue with the custom animation not being used?

akortunov:
Here is the NIF file: [tr_death_armordummy_a.zip](https://github.com/OpenMW/openmw/files/2314881/tr_death_armordummy_a.zip)

I need somehow to load animation groups from this file and allow NPC to use these animations.

kcat:
Perhaps simply changing `Animation::addAnimSource` to use the given filename as-is if the kf file doesn't exist. Or only change the extension to kf if the given filename doesn't exist. Would need some testing to see how vanilla behaves with nif and/or kf files available, while the NPC is defined to use a custom nif or kf file.

akortunov:
> Perhaps simply changing Animation::addAnimSource to use the given filename as-is if the kf file doesn't exist.

It does not work. As I said, KF-files have structure, different from NIF-files.
Since I can not use the Loader::loadKf() here, I use the Loader::load(), fill the KeyframeHolder and use it. Result is [here](https://i.imgur.com/tWt965p.jpg).

akortunov:
Closing for now since I have no idea how to implement this feature properly.

PR #1886, opened 2018-08-24
Workaround a bug in the Rapture3D OpenAL driver

kcat:
There's no associated OpenMW bug with this since there's not any problem on OpenMW's side (it's a bug in an OpenAL driver a user put in place of the provided one, discussed [here](https://forum.openmw.org/viewtopic.php?f=8&t=5353)). It's also nothing a user would normally experience unless they go out of their way to replace the provided OpenAL DLL on Windows with a specific different one, so I'm not sure it warrants a changelog entry.


zinnschlag:
I don't agree that this does not warrant a changelog entry. Hardening OpenMW against external problems is still a significant change (should get the task label, instead of bug or feature).

kcat:
Added [task #4606](https://gitlab.com/OpenMW/openmw/issues/4606), updated with changelog entry.

PR #1888, opened 2018-08-24
 Fix error: invalid application of 'sizeof' to an incomplete type 'Video::VideoPlayer'

pineapplemachine:
Fixes compile error encountered on OSX 10.9 using g++

## The compilation error:

```
[ 55%] Building CXX object apps/openmw/CMakeFiles/openmw.dir/mwgui/windowmanagerimp.cpp.o
In file included from /Users/pineapple/git/openmw/apps/openmw/mwgui/windowmanagerimp.cpp:1:
In file included from /Users/pineapple/git/openmw/apps/openmw/mwgui/windowmanagerimp.hpp:10:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/stack:86:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/deque:159:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__split_buffer:7:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/algorithm:627:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:2424:27: error: invalid application of
      'sizeof' to an incomplete type 'Video::VideoPlayer'
            static_assert(sizeof(_Tp) > 0, "default_delete can not delete incomplete type");
                          ^~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:2625:13: note: in instantiation of
      member function 'std::__1::default_delete<Video::VideoPlayer>::operator()' requested here
            __ptr_.second()(__tmp);
            ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:2593:46: note: in instantiation of
      member function 'std::__1::unique_ptr<Video::VideoPlayer, std::__1::default_delete<Video::VideoPlayer> >::reset' requested here
    _LIBCPP_INLINE_VISIBILITY ~unique_ptr() {reset();}
                                             ^
[...]

PR #1890, opened 2018-08-25
 Fix error: invalid application of 'sizeof' to an incomplete type 'Video::VideoPlayer' (No include)

pineapplemachine:
Alternate solution to same problem as in https://github.com/OpenMW/openmw/pull/1888 added in response to @zinnschlag's comment https://github.com/OpenMW/openmw/pull/1888#issuecomment-415951250

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -153,6 +153,7 @@ Programmers
     Siimacore
     sir_herrbatka
     smbas
+    Sophie Kirschner (pineapplemachine)
     spycrab
     Stefan Galowicz (bogglez)
     Stanislav Bobrov (Jiub)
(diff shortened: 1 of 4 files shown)


zinnschlag:
This looks okay. Now to some formalities (sorry about that).

1. Please file an issue on our bug tracker for this change so we have it on record. You can create a single issue for both this and your other PR, since they both basically do the same job (improve compatibility for a specific compiler in regards to use of smart pointers and forward declarations).

2. We recently changed how we handle new developers (that is not documented anywhere yet): Please add a commit to this PR, in which you add your name to the AUTHORS.md file.

pineapplemachine:
@zinnschlag Ok, I added to AUTHORS.md and created an issue here: https://gitlab.com/OpenMW/openmw/issues/4613

akortunov:
Add entry to CHANGELOG.md too.

akortunov:
Looks like there is something wrong with CI itself.
And just for reference: prefer rebase over merge - additional merge commits are not good.

pineapplemachine:
My b. I will rebase next time.

PR #1891, opened 2018-08-26
Mass potion creation

akortunov:
For now mass potion creation is a quite boring - you need to spam Enter button on mouse clicks on Create button to create potions one by one.
This PR adds a numeric edit box which allows to specify a count of potions to create.
Chance of success and skill gain is calculated for every potion, so there should be no difference here.

### DIFF
--- a/apps/openmw/mwgui/alchemywindow.cpp
+++ b/apps/openmw/mwgui/alchemywindow.cpp
@@ -1,7 +1,5 @@
 #include "alchemywindow.hpp"
 
-#include <climits>
-
 #include <MyGUI_Gui.h>
 #include <MyGUI_Button.h>
 #include <MyGUI_EditBox.h>
@@ -323,8 +321,8 @@ namespace MWGui
     {
         int currentCount = mBrewCountEdit->getValue();
 
-        // prevent overflows, and prevent entering INT_MIN since abs(INT_MIN) is undefined
-        if (currentCount == INT_MAX || currentCount == INT_MIN+1)
+        // prevent overflows
+        if (currentCount == std::numeric_limits<int>::max())
             return;
 
         mBrewCountEdit->setValue(currentCount+1);
--- a/apps/openmw/mwgui/alchemywindow.cpp
+++ b/apps/openmw/mwgui/alchemywindow.cpp
@@ -1,5 +1,7 @@
 #include "alchemywindow.hpp"
 
+#include <climits>
+
 #include <MyGUI_Gui.h>
 #include <MyGUI_Button.h>
 #include <MyGUI_EditBox.h>
@@ -25,6 +27,9 @@
 
 namespace MWGui
 {
+    const float AlchemyWindow::sCountChangeInitialPause = 0.5f;
+    const float AlchemyWindow::sCountChangeInterval = 0.1f;
+
     AlchemyWindow::AlchemyWindow()
         : WindowBase("openmw_alchemy_window.layout")
         , mSortModel(NULL)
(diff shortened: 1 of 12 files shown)


zinnschlag:
I am somewhat sceptical about this change. Alchemy is already very overpowered and the chance that players will abuse this is 100%. Question is, should we care about that?

pineapplemachine:
Can we put the widget behind a configuration switch?

As someone who generally refrains from abusing alchemy, something like this would be nice to just make the alchemy UI less annoying. But I can see where some players wouldn't like how much easier abuse becomes.

akortunov:
I do not like an idea to add yet another option.
I'd prefer to do not touch current behaviour at all.

wareya:
The alchemy exploit is actually most overpowered when you drink one potion at a time between alchs. 

Something like this just makes it less annoying to make several potions at once, for example 50 health potions.

zinnschlag:
wareya has a good point. Technically this change could make it a bit harder to exploit alchemy, because it requires additional work to use the brew one/drink one pattern. Again, if that is desirable or not is a matter for debate. But it is also not very significant.

IMO the feature is okay, but it should be controlled by a user settings options, because for some usage patterns it is clearly a downgrade.

Capostrophic:
> for some usage patterns it is clearly a downgrade

A new count dialog is not necessarily the only and final way of implementing this. I previously suggested akortunov to make the dialog a part of potionbrewing window itself and have count 1 as the default; he said it was possible but also that it's better to wait until there's a clear say for whether this feature would be included at all.

zinnschlag:
> I previously suggested akortunov to make the dialog a part of potionbrewing window itself and have count 1 as the default

Now that is a really neat idea.

akortunov:
Default value in counting widget inside alchemy window is now 1, but I still can not get how it should improve a situation.

akortunov:
With #1915 player can not spam potion creation by holding the ENTER key.

zinnschlag:
Still have to give it a final test run, but from looking at the source this appears to be okay to me. Does anyone else has something to contribute to the discussion or can I go ahead and put it into my review queue?

zinnschlag:
Finally got around to taking it for a test drive. I find the count dialogue very annoying. IMO it makes sense to go for Capostrophic's suggestion instead.

akortunov:
> IMO it makes sense to go for Capostrophic's suggestion instead.

What do you mean? Implement a count input field, as in trading window, right?

Capostrophic:
>Implement a count input field, as in trading window, right?

Yeah. Carbon copy of the barter window field won't look well, though, and there's no clear enough caption GMST possible to use. (sUses? sXTimes?)

akortunov:
Using a numeric edit box now. Ready for testing and review.

zinnschlag:
Looks good. Almost ready to merge. There are a few minor issues with the code quality. Will comment inline.

apps/openmw/mwgui/alchemywindow.cpp
```diff
@@ -250,4 +279,63 @@ namespace MWGui
...
+    {
+        mBrewCountEdit->setValue(std::abs(value));
+    }
+
+    void AlchemyWindow::onIncreaseButtonTriggered()
+    {
+        int currentCount = mBrewCountEdit->getValue();
+
+        // prevent overflows, and prevent entering INT_MIN since abs(INT_MIN) is undefined
+        if (currentCount == INT_MAX || currentCount == INT_MIN+1)
```
zinnschlag: This is the C way of doing it. Let's go for the C++ version (std::numeric_limits).

apps/openmw/mwgui/alchemywindow.cpp
```diff
@@ -250,4 +279,63 @@ namespace MWGui
...
+
+        mBrewCountEdit->setValue(currentCount+1);
+    }
+
+    void AlchemyWindow::onDecreaseButtonTriggered()
+    {
+        int currentCount = mBrewCountEdit->getValue();
+        if (currentCount <= 1)
+            return;
+        else
```
zinnschlag: Else after return doesn't make much sense.

apps/openmw/mwgui/alchemywindow.cpp
akortunov: I took this code from the trading window. Change it thre too?

PR #1901, opened 2018-08-29
Optimize combat AI

akortunov:
[Task #4621](https://gitlab.com/OpenMW/openmw/issues/4621).

Main changes:
1. Do not build map if resistance/weakness effects every time we access it.
2. The same thing for magic schools map.
3. Use square distance, when actors selects a nearest target.
4. Optimize combat action iteration: do not get equipment slots for every item in inventory, do not search for every effect in store twice.

[Here](https://imgur.com/a/WTGJJzl) is the comparison (PR/master) with 20x20 actors combat.

Notes:
1. Most of these optimizations useful only for actors with known spells and with magic items, potions, etc.
The more spells/items - the more difference.
2. getWeaknessEffect() and getResistanceEffect() functions are not threadsafe now, but we use them only in game mechanics in the main thread anyway. Also we initialize other static maps in the same way.
3. There is almost no difference for clean game without mods.

### DIFF
--- a/apps/openmw/mwmechanics/aisequence.cpp
+++ b/apps/openmw/mwmechanics/aisequence.cpp
@@ -228,7 +228,7 @@ void AiSequence::execute (const MWWorld::Ptr& actor, CharacterController& charac
             {
                 if ((*it)->getTypeId() != AiPackage::TypeIdCombat) break;
 
-                MWWorld::Ptr target = static_cast<const AiCombat *>(*it)->getTarget();
+                MWWorld::Ptr target = (*it)->getTarget();
 
                 // target disappeared (e.g. summoned creatures)
                 if (target.isEmpty())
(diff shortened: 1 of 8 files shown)


Capostrophic:
MWMechanics::spellSchoolToSkill map, while not huge, can probably be made static too.

apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -242,11 +242,11 @@ void AiSequence::execute (const MWWorld::Ptr& actor, CharacterController& charac
 
                     const ESM::Position &targetPos = target.getRefData().getPosition();
 
-                    float distTo = (targetPos.asVec3() - vActorPos).length();
+                    float distTo = (targetPos.asVec3() - vActorPos).length2();
 
                     // Small threshold for changing target
                     if (it == mPackages.begin())
-                        distTo = std::max(0.f, distTo - 50.f);
+                        distTo = std::max(0.f, distTo - 250.f);
```
Capostrophic: I'm probably missing something, but 250 is not 50 squared. Was it chosen intentionally?

PR #1909, opened 2018-09-04
Fix all MSVC warnings left

akortunov:
Let see how CI will handle this PR.


akortunov:
Well, it works. There is no MSVC warnings left.

PR #1911, opened 2018-09-06
Implements vanilla's off-by-one error, fixing #4611

Assumeru:
Guess what I'm adding to my settings file if this gets merged.


Capostrophic:
calcEffectCost is only used in the spellmaking dialog because spells made in TESCS get their cost autocalculated there - you don't need to add a new argument and modify spellcreationdialog.cpp. Add the setting boolean retrieval to spellcasting.cpp instead. If you want to go full purist, you can just make duration integer 1 by default and add the not-no-duration effects' duration to that 1, to hell with damned settings and puny documentations and launcher advanced pages, we have formulas to break.

Assumeru:
Ah, you're right my setting didn't do what I thought it did. I kinda thought the function would be shared with the CS.

Note that it's not just no-duration spells that have duration 0 in TESCS:
Fire bite (15-30 fire damage, 1s) costs 6 in TESCS with autocalc. Creating the same spell at a spellmaker costs 11 because it uses 2s for the calculation. Which is why I wanted a setting so the CS's autocalc would be the same as the spellmaker's, but that should wait until #2887 is implemented so they can use the same function (with that default parameter I added).

files/settings-default.cfg
```diff
@@ -228,7 +228,7 @@ barter disposition change is permanent = false
 # Uses the MCP formula (damage * (strength / 40)) to factor Strength into hand-to-hand combat.
 # (0 means it does not factor it in, 1 means it factors into werewolves damage calculation and
 # 2 means werewolves are ignored) 
-strength influences hand to hand = 0
+strength influences hand to hand = 0
```
Capostrophic: I'm not sure what's different but this will mess up git blame.

Capostrophic:
You need to add a changelog entry.

PR #1912, opened 2018-09-08
Coverity Scan fixes

akortunov:
Fixes some issues, found by Coverity Scan (up to 8 ones).
Mostly unitialized fields again.


akortunov:
A separate question: we have five destructors that can throw the MyGUI::Exception. How should we handle them?

apps/essimporter/converter.hpp
```diff
@@ -526,7 +526,11 @@ class ConvertJOUR : public Converter
 class ConvertGAME : public Converter
 {
 public:
-    ConvertGAME() : mHasGame(false) {}
+    ConvertGAME()
+        : mHasGame(false)
+        , mGame(GAME())
```
zinnschlag: What is the GAME() part for? Looks redundant to me.

akortunov: Coverity scan complains about uninitialised struct members. Maybe there is a better way to do it.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1185,10 +1185,10 @@ namespace MWWorld
                 addContainerScripts (getPlayerPtr(), newCell);
                 newPtr = getPlayerPtr();
             }
-            else if (currCell)
+            else if (currCell && newCell)
```
zinnschlag: I do not understand this set of changes. They seem to change the program logic quite a bit. Also, the error messages don't make sense. There is no meaningful concept of "empty cells".

components/esm/loadcell.hpp
```diff
@@ -90,6 +90,8 @@ struct Cell
 
   Cell() : mName(""),
            mRegion(""),
+           mData(DATAstruct()),
```
zinnschlag: Again, the DATAstruct() part looks redundant.

components/files/constrainedfilestream.cpp
```diff
@@ -103,21 +103,23 @@ namespace Files
...
-        delete rdbuf();
+        delete mBuf;
     }
 
 
     IStreamPtr openConstrainedFileStream(const char *filename,
                                                        size_t start, size_t length)
     {
-        return IStreamPtr(new ConstrainedFileStream(filename, start, length));
+        ConstrainedFileStreamBuf* buf = new ConstrainedFileStreamBuf(filename, start, length);
```
zinnschlag: This looks mildly unsafe. If the second new happens to throw, the first one would be left without a matching delete, i.e. resources leak.

zinnschlag:
> A separate question: we have five destructors that can throw the MyGUI::Exception. How should we handle them?

Catch (within the destructor), log and ignore.

apps/openmw/mwworld/worldimp.cpp
akortunov: In this block we call the "moveTo" method by using the "->" syntax. If currCell or newCell is nullptr here, there will be a crash. In theory, currCell could be nullptr only for player during new game start. Instead error logging we can throw an exception here to provide a more informative message.

components/esm/loadcell.hpp
akortunov: Again, it is not redundant. It allows to make sure that all fiends of mData struct are initialided by zeroes. With our current behaviour mData contains old values since C++ does not fill memory by zeroes automatically. At least, GCC7 behaves in this way.

components/files/constrainedfilestream.cpp
akortunov: What can we do here?

akortunov:
> Catch (within the destructor), log and ignore.

Done.

components/files/constrainedfilestream.cpp
kcat: I'd recommend a `unique_ptr`, unless you can't get away with multiple owners.

apps/openmw/mwworld/worldimp.cpp
zinnschlag: Yes, I get that part. What I do not get is your change. The problem is dealing with newCell==nullptr. The old code didn't handle that properly. But the new one doesn't either (see line 1173).
IIRC calling this function with a nullptr would be an error. If that is the case (and if we really want to take care of it), the easiest way to handle the issue would be a check at the beginning of the function that indeed throws an exception.

And if I interpret the "// currCell == NULL should only happen for player, during initial startup" comment correctly, CurrCell==nullptr can only happen when the moved object is the player, then this can be handled the same way (with an additional isPlayer in the check for the exception).

apps/essimporter/converter.hpp
akortunov: What do you mean by "Fixing GAME"?

components/esm/loadcell.hpp
akortunov: What's wrong with DATAstruct? It is just a simple structure with three fields.

components/files/constrainedfilestream.cpp
akortunov: Actually, there are two owners for now: mBuf field and openConstrainedFileStream() function.
I tried to use the unique_ptr here, but got the crash.

apps/openmw/mwworld/worldimp.cpp
akortunov: Throw exceptions now.

akortunov:
The main issue left for now is how to initialize struct class fields properly, right?

components/files/constrainedfilestream.cpp
kcat: `openConstrainedFileStream` just passes it to the `ConstrainedFileStream` constructor and relinquishes ownership. Did you use `std::move` to pass it?

apps/essimporter/converter.hpp
kcat: The `GAME` (and `DATAstruct`) members are left uninitialized in the default constructor. You can change the struct to do something like
```
struct DATAstruct
  {
    int mFlags{0};
    int mX{0}, mY{0};
  };
```
to ensure the fields are constructed.

components/files/constrainedfilestream.cpp
akortunov: > Did you use std::move to pass it?

No, I did not. It seems it led to crash. With std::move it works.

apps/essimporter/converter.hpp
akortunov: Done.

PR #1914, opened 2018-09-09
Use sTo GMST in spellmaking menu (feature #4636)

Capostrophic:
[Feature 4636](https://gitlab.com/OpenMW/openmw/issues/4636)

Recover the sTo GMST value for max magnitude caption. On English Morrowind.esm it becomes "to %number%", I can't tell what will it be on other localizations but sTo is definitely used in spellmaking menu in the original engine. I decided to use a hyphen instead of "to" by default because if the GMST is missing we'd most likely want to use as "international" variant of it as possible.

### DIFF
--- a/apps/openmw/mwgui/spellcreationdialog.cpp
+++ b/apps/openmw/mwgui/spellcreationdialog.cpp
@@ -147,7 +147,9 @@ namespace MWGui
 
         mDurationValue->setCaption("1");
         mMagnitudeMinValue->setCaption("1");
-        mMagnitudeMaxValue->setCaption("- 1");
+        static const std::string &to = MWBase::Environment::get().getWindowManager()->getGameSettingString("sTo", "-");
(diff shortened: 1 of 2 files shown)


apps/openmw/mwgui/spellcreationdialog.cpp
```diff
@@ -147,7 +147,9 @@ namespace MWGui
 
         mDurationValue->setCaption("1");
         mMagnitudeMinValue->setCaption("1");
-        mMagnitudeMaxValue->setCaption("- 1");
+        static const std::string &to = MWBase::Environment::get().getWindowManager()->getGameSettingString("sTo", "-");
```
zinnschlag: This is technically correct since the lifetime of the string is extended to the lifetime of the reference. Still, looks odd and I can not see any benefit in using a reference here.

PR #1915, opened 2018-09-09
Disable repeating for Accept GUI action

akortunov:
Fixes [bug #4260](https://gitlab.com/OpenMW/openmw/issues/4260).
There is no need to hold ENTER button and apply the same action again and again.
An additional thing we can do here is to set focus to Cancel button every time we open a persuation window, but it will decrease usability of keyboard navigation feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
     Bug #4230: AiTravel package issues break some Tribunal quests
     Bug #4231: Infected rats from the "Crimson Plague" quest rendered unconscious by change in Drain Fatigue functionality
     Bug #4251: Stationary NPCs do not return to their position after combat
+    Bug #4260: Keyboard navigation makes persuasion exploitable
     Bug #4271: Scamp flickers when attacking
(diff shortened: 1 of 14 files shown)


wareya:
If this is what I think it is I used it when bartering to work around the lack of the "increase shopkeeper's stock" exploit.

akortunov:
Strange, but MyGUI still fires eventEditSelectAccept events in EditBox (our text input fields) until release of ENTER NUMPAD_ENTER buttons, even when I filter input events with "repeat" flag.

akortunov:
I figured out that MyGUI InputManager handles continuous press internally.
I found a workaround to disable such repeating - fire "release button" MyGUI event manually.

zinnschlag:
Does anyone else has anything to contribute to this discussion? If not, I would be okay with going ahead with the merge.

AmihaLiss:
Just for check. 
It's also disable repeating in alchemy menu?

akortunov:
> It's also disable repeating in alchemy menu?

Yes, I already wrote it.

PR #1919, opened 2018-09-13
Make coc and fixme console commands close to vanilla

akortunov:
[bug #4292](https://gitlab.com/OpenMW/openmw/issues/4292) and [bug #4217](https://gitlab.com/OpenMW/openmw/issues/4217).

Amount of changes:
1. Fixme:
1.1. Does not work with references and always teleports player, so you can use the "fixme" in console instead of "player->fixme".
1.2. Moves player 128 units away from its current position to unoccupied direction as in vanilla instead of moving hit 8000 units up and then tracing down.
1.3. Use ActionTeleport as for COE/COC to avoid fall damage, if you use fixme in the air.

2. COC:
2.1. Does not use "fixme", just adjusts z position instead.
2.2. Instead of teleporting to first teleporting door in interior, sorts all teleporting doors in cell by ID first, then by destination cell name in alphabet order and selects the first suitable one.
This approach may be not entirely correct, but in my testing Morrowind worked in the same way.
But I tested it on a Russian version of Morrowind, so I do not know how an English one works.

### DIFF
--- a/apps/openmw/mwscript/cellextensions.cpp
+++ b/apps/openmw/mwscript/cellextensions.cpp
@@ -46,18 +46,19 @@ namespace MWScript
 
                     ESM::Position pos;
                     MWBase::World *world = MWBase::Environment::get().getWorld();
+                    const MWWorld::Ptr playerPtr = world->getPlayerPtr();
 
                     if (world->findExteriorPosition(cell, pos))
                     {
-                        MWWorld::ActionTeleport("", pos, false).execute(world->getPlayerPtr());
-                        world->fixPosition(world->getPlayerPtr());
+                        MWWorld::ActionTeleport("", pos, false).execute(playerPtr);
+                        world->adjustPosition(playerPtr, false);
                     }
                     else
                     {
                         // Change to interior even if findInteriorPosition()
                         // yields false. In this case position will be zero-point.
                         world->findInteriorPosition(cell, pos);
-                        MWWorld::ActionTeleport(cell, pos, false).execute(world->getPlayerPtr());
+                        MWWorld::ActionTeleport(cell, pos, false).execute(playerPtr);
                     }
                 }
         };
@@ -76,14 +77,15 @@ namespace MWScript
 
                     ESM::Position pos;
                     MWBase::World *world = MWBase::Environment::get().getWorld();
+                    const MWWorld::Ptr playerPtr = world->getPlayerPtr();
 
                     world->indexToPosition (x, y, pos.pos[0], pos.pos[1], true);
                     pos.pos[2] = 0;
 
                     pos.rot[0] = pos.rot[1] = pos.rot[2] = 0;
 
-                    MWWorld::ActionTeleport("", pos, false).execute(world->getPlayerPtr());
-                    world->fixPosition(world->getPlayerPtr());
+                    MWWorld::ActionTeleport("", pos, false).execute(playerPtr);
+                    world->adjustPosition(playerPtr, false);
                 }
         };
 
--- a/apps/openmw/mwscript/transformationextensions.cpp
+++ b/apps/openmw/mwscript/transformationextensions.cpp
@@ -732,14 +732,13 @@ namespace MWScript
             }
         };
 
-        template <class R>
         class OpFixme : public Interpreter::Opcode0
         {
         public:
 
             virtual void execute (Interpreter::Runtime& runtime)
             {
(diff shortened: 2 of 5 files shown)


Capostrophic:
Russian version of Morrowind is a special case because cell IDs there are the same as in English version thanks to using .cel files for figuring the difference between the internally used name and the displayed name. This is different from other Morrowind localizations where all cells have translated IDs. Might be worth making some test cases to run in Morrowind, a non-Russian localized Morrowind and in openmw with PR behavior.

apps/openmw/mwscript/cellextensions.cpp
```diff
@@ -46,18 +47,19 @@ namespace MWScript
...
                     ESM::Position pos;
                     MWBase::World *world = MWBase::Environment::get().getWorld();
+                    const MWWorld::Ptr playerPtr = world->getPlayerPtr();
 
                     if (world->findExteriorPosition(cell, pos))
                     {
-                        MWWorld::ActionTeleport("", pos, false).execute(world->getPlayerPtr());
-                        world->fixPosition(world->getPlayerPtr());
+                        MWWorld::ActionTeleport("", pos, false).execute(playerPtr);
+                        playerPtr.getClass().adjustPosition(playerPtr, true);
```
Capostrophic: You should use World::adjustPosition method so you don't need to add a class.hpp dependency.

Capostrophic:
There's another fixPosition call that should be changed to adjustPosition - in World::startNewGame.

Capostrophic:
Actually, here's one complaint about fixme: it should use fade-in and fade-out to obscure the instant position and camera rotation change. Obviously it's partially broken in Morrowind (so you see how jerky the teleportation is) but I see no reason why shouldn't we replicate that.

Capostrophic:
fixme seems to snap the player to the ground in Morrowind.

akortunov:
> The coordination of the player doesn't seem to be accurate still.

It is not coordinates, it is a rotation. Does it really matter? Currently we just take rotation from door marker, what makes sense, but I see no reasonable explanation how Morrowind handles rotation.

> You should use World::adjustPosition method instead so you don't need to add a class.hpp dependency.

Done.

> Actually, here's one complaint about fixme: it should use fade-in and fade-out to obscure the instant position and camera rotation (it's reset in Morrowind to forward direction) change.

I see no reason to implement it since it does not work in Morrowind and does not have any useful functions here.
Fading supposed to make screen black during long loading process, but there is no cell transition here (simlarly to SetPos command).
But we can use fading in the COC/COE commands, if we really need it.

akortunov:
> fixme seems to snap the player to the ground in Morrowind.

In OpenMW it traces player down. I already described in #1701 why tracing up does not work.

Capostrophic:
About 3: what I meant was that openmw doesn't snap the player to the ground so if you decide to fixme yourself from a cliff y

Capostrophic:
I'm on mobile, and github is clunkier. I see that there is an attempt to trace down, but Morrowind traces down for a much higher distance.

akortunov:
> I see that there is an attempt to trace down, but Morrowind traces down for a much higher distance.

Should be fixed now.

P.S. I do not like a lot of cell size magic numbers. Can we define such widely used constants in one header and then just use it?

apps/openmw/mwscript/transformationextensions.cpp
```diff
@@ -732,14 +732,13 @@ namespace MWScript
...
 
-        template <class R>
         class OpFixme : public Interpreter::Opcode0
         {
         public:
 
             virtual void execute (Interpreter::Runtime& runtime)
             {
-                MWWorld::Ptr ptr = R()(runtime);
+                const MWWorld::Ptr ptr = MWMechanics::getPlayer();
```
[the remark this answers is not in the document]
akortunov: I see to reasons to change API here.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1352,13 +1352,37 @@ namespace MWWorld
 
     void World::fixPosition(const Ptr &actor)
```
Capostrophic: Arguably it's unnecessary to have an argument for this function since it can only be used for the player.

akortunov:
I suppose there will not be additional testing. As for me, this PR makes bevaviour close to vanilla enough to close these bugs (or remove them from 1.0 milestone, at least).

PR #1925, opened 2018-09-17
Use constants instead of widely used magic numbers

akortunov:
Implements [task #4645](https://gitlab.com/OpenMW/openmw/issues/4645).

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -9,6 +9,7 @@
 #include <components/esm/esmwriter.hpp>
 #include <components/esm/cellid.hpp>
 
+#include <components/misc/constants.hpp>
 #include <components/misc/resourcehelpers.hpp>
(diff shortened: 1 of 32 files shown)


apps/openmw/mwworld/worldimp.cpp
```diff
@@ -3245,9 +3244,7 @@ namespace MWWorld
 
     float World::feetToGameUnits(float feet)
     {
-        // Looks like there is no GMST for this. This factor was determined in experiments
-        // with the Telekinesis effect.
-        return feet * 22;
+        return feet * Constants::UnitsPerFoot;
```
akortunov: Now it is 21.33333333. Are we fine with it (since it is an arbitrary value, according to comment), or we desperately need a 22 here?

zinnschlag: Not sure. Do a test? The function is used in only a few places. Shouldn't be hard to figure out if there is a noticeable difference.

PR #1927, opened 2018-09-17
[Regression] Do not use book fonts for dialogue window

akortunov:
Fixes a small bug in the new font scaling system.
For now we use "book" fonts (fonts, scaled for JournalBook widget) for dialogue window too since it uses the same code as the JournalBook widget. As a result, dialogue window output looks slightly different from as it supposed to be.

Added an argument to createStyle() function to control if there is a need to use a "book" font or a common one.


akortunov:
There is an another bug found - scrolls use a common ScrollView widget and I need somehow to set default font here.

PR #1928, opened 2018-09-18
Improve optional tooltips output

akortunov:
1. Since light sources with -1 Duration are infinite, there is no point to show -1 duration on tooltip.
Another negative values will be displayed as 0 (since we treat them as 0).
2. I figured out that ranged weapons have an attack speed too, but for now we display it only for melee weapons. Fixed, but I decided to do not rename setting.

### DIFF
--- a/apps/openmw/mwclass/light.cpp
+++ b/apps/openmw/mwclass/light.cpp
@@ -161,7 +161,12 @@ namespace MWClass
         std::string text;
 
         if (Settings::Manager::getBool("show effect duration","Game"))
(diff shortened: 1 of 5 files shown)


Capostrophic:
I initially assumed that ranged weaponry has a speed multiplier and implemented it in rating for anything that's not an arrow or a bolt but then I looked at the code of the speed multiplier tooltip and decided that it was no-op, so currently AI thinks ranged weaponry can't have a speed multiplier :smile:
So you may want to modify [this](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwmechanics/weaponpriority.cpp#L122) line.

PR #1929, opened 2018-09-19
Ignore the rest of the line after else operator (bug #3006)

Capostrophic:
[Bug 3006](https://gitlab.com/OpenMW/openmw/issues/3006)

TESCS compiler actually ignores literally anything put after else operator, so so should our compiler. I put in more handling of IfElseJunkState so that the rest of the line is ignored, like the original bytecode of such scripts demonstrates. This allows mods which erroneously use "else if" instead of "elseif" in some scripts such as Sotha Sil Expanded and Farmer's Mod to compile properly, albeit with a potential loss of functionality, replicating vanilla behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
     Bug #2969: Scripted items can stack
     Bug #2987: Editor: some chance and AI data fields can overflow
+    Bug #3006: 'else if' operator breaks script compilation
     Bug #3282: Unintended behaviour when assigning F3 and Windows keys
     Bug #3623: Fix HiDPI on Windows
     Bug #3733: Normal maps are inverted on mirrored UVs
--- a/components/compiler/controlparser.cpp
+++ b/components/compiler/controlparser.cpp
@@ -179,6 +179,14 @@ namespace Compiler
             scanner.scan (mLineParser);
             return true;
         }
+        else if (mState==IfElseJunkState)
+        {
+            getErrorHandler().warning ("Ignoring extra text after else", loc);
+            SkipParser skip (getErrorHandler(), getContext());
+            scanner.scan (skip);
+            mState = IfElseBodyState;
+            return true;
+        }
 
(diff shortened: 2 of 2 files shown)


zinnschlag:
Its been a long time since I looked into this. Will have to do some serious digging to come up with an answer (the whole script workaround stuff is excessively obscure). I'll try to make time for it, but don't expect an answer anytime soon.

Capostrophic:
I guess "soon" timeframe hasn't passed by yet. :P

zinnschlag:
Yeah, sorry about that. I haven't forgotten you. But the various workarounds for broken vanilla scripts aren't documented anywhere other than the source and the interactions can be very subtle. I need to take 1-2 hours where I can dive uninterruptedly into the code. And that is not on the table right now. Too much stuff happening that is competing for my attention (OpenMW and otherwise).

psi29a:
@zinnschlag can we pull this into 0.46 for testing and exposure?

zinnschlag:
I am still kinda working on the assumption that his will break things. I just haven't figured out why. Therefore I would prefer not to merge at this time.

psi29a:
Roger that.

Capostrophic:
I tried to manually "decompile" some scripts which used this syntax and apparently the part with the condition after else simply gets ignored by the compiler and it handles the branch as a simple "else". So IfElseJunkState condition ignoring must be extended to keywords.

Capostrophic:
Now it's logging warnings. It spits out some stuff for Bloodmoon.esm. e.g. there was such construction in ColonyBaro script:
```cpp
if ( player->GetWeaponType == -1 )
set weaponout to 1
else set weaponout to 2
```
where set weaponout to 2 after else is ignored.
I checked the bytecode of the script in question and it looks like TESCS borks their code too. There are actually a bunch more scripts than this that have broken syntax. So this actually makes the behavior of OpenMW in regards to Bethesda master files closer to vanilla as well.

PR #1930, opened 2018-09-20
Use autogenerated collision boxes as fallback for creatures

akortunov:
Implements [feature #2787](https://gitlab.com/OpenMW/openmw/issues/2787).
Currently, we have no fallback if creature's model does not contain a bounding box.
The main idea here:
1. Try to retrieve collision mesh from basic model (without X)
2. If it did not help, try to generate collisions box using osg::ComputeBoundsVisitor.

Unfortunately, ComputeBoundsVisitor returns bounds in world axes, not in local ones, so bounding box can be aligned in the wrong way.

But even in its current state this PR fixes issue with Tyddy's replacers.
Is there a better way to generate collision box?


symbiogenesis:
Just to be clear, you are saying that vanilla morrowing does have such a fallback?

Capostrophic:
It does.

akortunov:
@zinnschlag, what do you think?

zinnschlag:
The ComputeBoundsVisitor issue sounds problematic. If I understand it correctly this affects step 2 only, correct? Not keen on introducing a potential erroneous behaviour, even if it improves the situation. Is there a fix for step 2? Can we maybe only do step 1 for now and deliver step 2 later, if a fix can be found?

akortunov:
> Is there a fix for step 2? 

Basically yes.

> Can we maybe only do step 1 for now and deliver step 2 later, if a fix can be found?

In this case there is a need for a separate bugreport and separate pull request, but it is possible.

zinnschlag:
Let's do that then.

akortunov:
Ok, I found an another approach: get bounding box for shape from shapeManager.
Result is similar to vanilla game.

Note: you can not use autogeneration just for any arbitrary creature: in both Morrowind and OpenMW this feature does not consider bones transformation and skinning.
For example, Tyddy does not use bones transformation in his replacers, so this feature works fine there.

In theory, there could be an additional tricky difference, but I doubt I can do anything here without reverse-engineering.

akortunov:
Looks like GitHub is drunk.
This PR does not even with my branch:
https://github.com/akortunov/openmw/tree/collisionbox

PR #1932, opened 2018-09-21
Optimize AI update loop

akortunov:
Difference especially noticable with increased "exterior cell loading distance".
Affected lines take a most of mechanics update time.

Also this code is a quite strange. As I understand, it supposed to reset the HitAttemptActorId for player, if target actor disappeared from cell for some reason. But we do not set HitAttemptActorId inside affected loop, only reset it, so there is no point to run this code for every actor in scene - it will use the same value or -1.
So run this check only once before AI update.

I decided to do not create a separate task since we already have one for AI optimization in 0.45 milestone.

Also this PR optimizes combat music update a bit.

A little performance comparison: time in milliseconds to update game mechanics (based on profilier output), compared with master on i7-7700 in Vivec:

exterior cell load distance = 4
Master: 12.5ms
PR: 3 ms

exterior cell load distance = 3
Master: 6ms
PR: 2ms

exterior cell load distance = 2
Master: 2ms
PR: 1ms

exterior cell load distance = 1
Master: 0.8ms
PR: 0.6ms

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -501,12 +501,6 @@ namespace MWMechanics
         }
     }
 
-    void Actors::updateNpc (const MWWorld::Ptr& ptr, float duration)
-    {
-        updateDrowning(ptr, duration);
-        calculateNpcStatModifiers(ptr, duration);
-    }
-
     void Actors::adjustMagicEffects (const MWWorld::Ptr& creature)
     {
         CreatureStats& creatureStats =  creature.getClass().getCreatureStats (creature);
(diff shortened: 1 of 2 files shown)


Capostrophic:
Boolean from isAIActive() is similarly recovered for each and every actor in updateCombatMusic() unnecessarily - and since the related code isn't even used if AI is disabled, the whole actor iterating loop there there could be put behind the check for the very same boolean. updateCombatMusic() is called every frame from the actor update loop.

akortunov:
I rewrote the updateCombatMusic(), but there is not noticable performance gain since this function does not use heavy operations (such as search).

Capostrophic:
Every bit helps.

akortunov:
Added optimization for drowning state update: previously we used a search inside mActors map to retrieve actor's character controller just to check if actor is knocked out. Just pass this flag as parameter now.

I think I did what I could do.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -926,31 +920,27 @@ namespace MWMechanics
         return ctrl->isSneaking();
     }
 
-    void Actors::updateDrowning(const MWWorld::Ptr& ptr, float duration)
+    void Actors::updateDrowning(const MWWorld::Ptr& ptr, float duration, bool isKnockedOut, bool isPlayer)
```
Capostrophic: Is passing isPlayer that necessary since the boolean is possible to be set in the function itself?

akortunov: I just do not like an idea to get reference to player for every actor.

PR #1935, opened 2018-09-22
Wrap up World::canRest() implementation by moving enemiesNearby() fro…

fukanchik:
…m InputManager::rest() to World::canRest().

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -438,12 +438,16 @@ namespace MWBase
 
             virtual void enableActorCollision(const MWWorld::Ptr& actor, bool enable) = 0;
 
-            virtual int canRest() = 0;
-            ///< check if the player is allowed to rest \n
-            /// 0 - yes \n
-            /// 1 - only waiting \n
-            /// 2 - player is underwater \n
-            /// 3 - enemies are nearby (not implemented)
+            enum RestPermitted
+            {
+                Rest_Allowed = 0,
+                Rest_OnlyWaiting = 1,
+                Rest_PlayerIsUnderwater = 2,
+                Rest_EnemiesAreNearby = 3
+            };
+
+            virtual RestPermitted canRest() = 0;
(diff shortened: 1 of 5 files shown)


apps/openmw/mwworld/worldimp.cpp
```diff
@@ -2297,7 +2297,7 @@ namespace MWWorld
         applyLoopingParticles(player);
     }
 
-    int World::canRest ()
+    World::RestPermitted World::canRest ()
```
Capostrophic: I haven't ever seen enum-returning things in openmw be implemented this way. Why can't this function return type (and the enum type itself) be an integer?

Capostrophic: Hey, I know what enum is. :<

Had a bit of a brainfart this night concerning enum types, sorry. It's just that something looked weird in this.

apps/openmw/mwbase/world.hpp
```diff
@@ -438,12 +438,16 @@ namespace MWBase
...
+            enum RestPermitted
+            {
+                Rest_Allowed = 0,
+                Rest_OnlyWaiting = 1,
+                Rest_PlayerIsUnderwater = 2,
+                Rest_EnemiesAreNearby = 3
+            };
+
+            virtual RestPermitted canRest() const = 0;
+            ///< check if the player is allowed to rest
```
zinnschlag: We don't use this style of comment (///<, comment after function) anymore, because is causes problems with some editors/IDE. Now we don't like people doing commits with purely stylistic changes either (because that increases the chance of merge conflicts), but since you are touching this line anyway, I suggest to fix it (remove the < and then move the comment to the top).

zinnschlag:
Looks okay. There are a few formalities that need to be taken care of before we can merge. Please do the following:

* File an issue on our tracker on GitLab and link it from the pull request description
* Add an entry to the CHNAGELOG.md file
* Please add your name to the AUTHOS.md file

PR #1936, opened 2018-09-23
Fix issues with sun specularity (bug #4527)

Capostrophic:
[Bug 4527](https://gitlab.com/OpenMW/openmw/issues/4527).

The idea of fixing the first subissue is applying time of day glare visibility to the sunlight specular color (not just the weather-dependent multiplier). This makes specular lighting change gradual depending on the difference of the current time and sun peak hour, and at nights it's completely disabled - granted it's not used for anything other than the sun, the directional light source, and no point light source has a specular color, and it's already disabled in rainy weather.

For the second subissue wareya's suggestion from the report was used; the shader was modified so that the specular uses the transparency of the sunlight like the color of the sunlight is used elsewhere in the shader.

Feedback would be appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,7 @@
     Bug #4503: Cast and ExplodeSpell commands increase alteration skill
     Bug #4510: Division by zero in MWMechanics::CreatureStats::setAttribute
     Bug #4519: Knockdown does not discard movement in the 1st-person mode
+    Bug #4527: Sun renders on water shader in some situations where it shouldn't
     Bug #4531: Movement does not reset idle animations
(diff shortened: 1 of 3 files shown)


zinnschlag:
Seems we are not getting any more feedback. Shall we go ahead with the merge?

Capostrophic:
OK, this should do it. At first I wanted to make separate calculations for specular fade, but it made no sense to have glare fade out completely just about when sunset starts - the sun is not under the horizon at that point yet.

psi29a:
Looks like the google testsuite was 'upgraded' to use C++11 (and higher) features. Will need to look into this.

PR #1937, opened 2018-09-23
Experimental animation regression fixes

Capostrophic:
This resolves some of the less severe issues that I've seen happening in the development code.

1. Try to change to spellcasting stance when the running animation has just started (hasn't got to the loop start point). The animation will be visually reset. Specifically, it's reset to the loop start key, because it's the same animation. But as far as I can tell we don't need to do it, because openmw already loops the animations at the correct keys. This is more of an addition to #1792.
2. Try to jump while running, land, and continue to run. The running animation will start from the wrong point. This happens because technically it starts exactly when the player lands, but the landing animation has higher priority (see #1666), so the running animation can only start when it ends. The idea is to reset the movement change in refreshJumpAnims like we do with the idle animations if the character is still jumping. This makes it somewhat more obvious that the character is sliding along the ground in the landing animation while moving, but I still suppose akortunov had to go with the lesser evil for this one.

Neither of these issues appear in 0.44.0, but they were in the code at that point and they became uncovered by the earlier animation fixes of 0.45.0. These are called experimental for a reason, so testing and feedback would be appreciated.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -353,7 +353,7 @@ void CharacterController::refreshHitRecoilAnims(CharacterState& idle)
 
 void CharacterController::refreshJumpAnims(const WeaponInfo* weap, JumpingState jump, CharacterState& idle, CharacterState& movement, bool force)
 {
-    if (!force && jump == mJumpState && movement == CharState_None)
+    if (!force && jump == mJumpState && idle == CharState_None && movement == CharState_None)
         return;
 
     if (jump != JumpState_None)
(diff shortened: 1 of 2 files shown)


Capostrophic:
akortunov informed me that the swim idle animations aren't reset properly. Did something and now it seems to be fine.

There's an issue with idle animations in first person view. For some reason when you land they momentarily flicker into a wrong position, which depends on the camera rotation: just test this out and you'll see it. The camera also flickers into the wrong position. The same issue appears on a smaller scale in master (presumably masked by the landing animation). If anyone knows why that happens, please let me know.

Capostrophic:
Pushed an ugly hack that allows to workaround the mentioned issue, so it works like it did before.

PR #1938, opened 2018-09-23
Restore dynamic stats for actors in inactive cells

akortunov:
Aimed to fix [bug #1875](https://gitlab.com/OpenMW/openmw/issues/1875).
I took code from respawn handling, and it seems it works.

But this code is a bit dirty yet - stats calculations are in Mechanics (Actors), but cells iteration is in World (Cells/Cellstore).

An approach is similar to vanilla: restore stats for actors in all loaded cells when sleeping, and restore fatigue for actors in active cells only when waiting.

### DIFF
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -90,6 +90,8 @@ namespace MWBase
             virtual void setPlayerClass (const ESM::Class& class_) = 0;
             ///< Set player class to custom class.
 
+            virtual void restoreDynamicStats(MWWorld::Ptr actor, bool sleep) = 0;
+
             virtual void rest(bool sleep) = 0;
             ///< If the player is sleeping or waiting, this should be called every hour.
             /// @param sleep is the player sleeping or waiting?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.45.0
 ------
 
+    Bug #1875: Actors in inactive cells don't heal from resting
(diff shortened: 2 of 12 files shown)


terabyte25:
Does this apply to both waiting and resting, or just resting (similar to how vanilla did it)?

akortunov:
> Does this apply to both waiting and resting, or just resting (similar to how vanilla did it)?

Only for resting. For waiting we only restore fatigue for actors in active cells.

apps/openmw/mwbase/mechanicsmanager.hpp
```diff
@@ -90,6 +90,10 @@ namespace MWBase
             virtual void setPlayerClass (const ESM::Class& class_) = 0;
             ///< Set player class to custom class.
 
+            virtual void restoreDynamicStats(MWWorld::Ptr actor, bool sleep) = 0;
+            ///< Restores actor's dynamic stats during player's rest
```
zinnschlag: We don't use this kind of comment anymore (///<, comment after function), because several developers complained that it screws with functions in their editor/IDE.

zinnschlag:
Do we have any performance data for this? A meaningful test would require a game state with a very large number of loaded cells.

akortunov:
> Do we have any performance data for this? A meaningful test would require a game state with a very large number of loaded cells.

I suppose for such test we need a save from Tamriel Rebuilt from a lot of visited cells.

zinnschlag:
Good enough

akortunov:
I did some additinal testing (for creatures, manually placed actors, etc.), nothing suspicious found.

PR #1940, opened 2018-09-24
Respect selected encoding in the content selector 

akortunov:
Fixes [bug #4467](https://gitlab.com/OpenMW/openmw/issues/4467).
Works for both launcher and editor.

Also I found an additional issue: maximal length of plugin description is capped in OpenMW for some reason (150 characters), so OpenMW cut descriptions for some plugins (probably because we treat it as a column name in metadata table). @zinnschlag, is there any way to fix it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@
     Bug #4460: Script function "Equip" doesn't bypass beast restrictions
     Bug #4461: "Open" spell from non-player caster isn't a crime
     Bug #4464: OpenMW keeps AiState cached storages even after we cancel AI packages
+    Bug #4467: Content selector: cyrillic characters are decoded incorrectly in plugin descriptions
     Bug #4469: Abot Silt Striders – Model turn 90 degrees on horizontal
     Bug #4470: Non-bipedal creatures with Weapon & Shield flag have inconsistent behaviour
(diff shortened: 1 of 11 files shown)


zinnschlag:
Odd. The description should be capped at 256 bytes. I have no idea where the 150 is coming from.

> probably because we treat it as a column name in metadata table

I have no idea what you mean by this. A field from a record should definitely not be a column name. And even if it where, there shouldn't be a size limitation either.

akortunov:
> Odd. The description should be capped at 256 bytes.

How much 1 character takes, one byte or two?

akortunov:
There is an issue with ESMReader::getString - encoder sometimes cut the string.

akortunov:
As I suspected, we convert 256-byte string to 256+ byte UTF-8 string and then try to assign it to 256-byte field inside loadtes3.cpp. We need to either remove this limit or decode strings in another place.
Probably other text fields are affected too.

akortunov:
[bug #4653](https://gitlab.com/OpenMW/openmw/issues/4653).
I suppose it is for another PR. This one is ready for review.

akortunov:
Added a fix for UTF-8 string length.
Now OpenMW displays metadat correctly, and generated omwaddon has the same metadata as original ESP file.

PR #1941, opened 2018-09-25
Do not check owner during stacking

akortunov:
Attempt to fix [bug #1933](https://gitlab.com/OpenMW/openmw/issues/1933).
There is no point to check owner during stacking check - we consider that all items in ContanerStore belong to owner of that container store anyway.

This PR should fix stacking in the trading window.
For pickpocketing and looting it will work only for newly added items - we do not use update stacks for these windows on the fly since it causes issues with stacking equippable items (such as throwing weapons and ammo).
For newly added items we just add them to existing item in NPC's inventory, just increase the count.
If we really need it, I can add update on the fly (excepts of stacking equippable items), but it can break something else.

Probably #1622 is not needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.45.0
 ------
 
+    Bug #1933: Actors can have few stocks of the same item
     Bug #1990: Sunrise/sunset not set correct
     Bug #2131: Lustidrike's spell misses the player every time
     Bug #2222: Fatigue's effect on selling price is backwards
--- a/apps/openmw/mwworld/containerstore.cpp
+++ b/apps/openmw/mwworld/containerstore.cpp
@@ -241,7 +241,6 @@ bool MWWorld::ContainerStore::stacks(const ConstPtr& ptr1, const ConstPtr& ptr2)
     }
 
     return ptr1 != ptr2 // an item never stacks onto itself
-        && ptr1.getCellRef().getOwner() == ptr2.getCellRef().getOwner()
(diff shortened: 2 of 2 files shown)


Capostrophic:
See #1299. You can't just remove the check.

akortunov:
> See #1299. You can't just remove the check.

Yes, I read this and still do not why we need to store all owners for items in container. When we use them?
For confiscation we track count of stolen items, also all items in player's inventory have empty owner field.

Capostrophic:
My guess was that owner contamination is obviously bad, but looking at the actual implementation of the so called "internal merging" - changing the count of an existing reference for making a stack - it looks like the need to stack different references can't be worked around from that side.

zinnschlag:
This has been sitting here for a while. Have we come to a conclusion?

sthalik:
In common sense terms, doesn't the merchant become the *owner* of the sold stolen items?

akortunov:
> In common sense terms, doesn't the merchant become the owner of the sold stolen items?

In terms of game mechanics too.
If you steal an item with ID=X from container with owner A, an entry "item X stolen from A" will be added to the stolen items map.
If you sell the X to trader B and steal it via pickpocketing then, an entry "item X stolen from B" will be added.
As a result, both A and B will claim the first item with ID=X which you will try to sell to them as theirs.
Also guards will try to confiscate two instances of X from player despite techically he stole the same item twice.

akortunov:
Closing due to lack for response.

PR #1942, opened 2018-09-26
Rename reflected absorb spells setting and add it to the launcher Advanced tab

Capostrophic:
And update the documentation a bit.


Capostrophic:
There's something odd with Travis.

PR #1221, opened 2017-03-04
[WIP] Let's see if release build will fit in Travis worker timeout

nikolaykasyanov:
If yes, I'll look into configuring automatic nightly builds for macOS.

PR #1944, opened 2018-09-27
Editor: NPC rendering

akortunov:
Implements [feature #1221](https://gitlab.com/OpenMW/openmw/issues/1221).
Based on [this](https://gitlab.com/OpenMW/openmw/merge_requests/25) PR by Aesylwinn.
He permitted me to finish this task since he have no enough time to finish it himself.
Summary of changes, added to Aesylwinn's work:
1. Use male bodyparts as fallback for females
2. Implement worn items priority for editor, so robes should be rendered fine now.
3. Added changelog entry.

Should be ready for testing and review.

Notes:
1. We do not render weapons for now (and it makes no sense anyway since we do not use autoequipping for weapon at all).
2. We do not use animations yet - it requires a lot of work and it better to leave it for another PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,7 @@
     Bug #4653: Length of non-ASCII strings is handled incorrectly in ESM reader
     Bug #4654: Editor: UpdateVisitor does not initialize skeletons for animated objects
     Feature #912: Editor: Add missing icons to UniversalId tables
+    Feature #1221: Editor: Creature/NPC rendering
     Feature #1617: Editor: Enchantment effect record verifier
     Feature #1645: Casting effects from objects
     Feature #2606: Editor: Implemented (optional) case sensitive global search
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -93,7 +93,7 @@ opencs_units (view/render
 
 opencs_units_noqt (view/render
     lighting lightingday lightingnight lightingbright object cell terrainstorage tagbase
-    cellarrow cellmarker cellborder pathgrid
+    cellarrow cellmarker cellborder pathgrid actor
     )
 
 opencs_hdrs_noqt (view/render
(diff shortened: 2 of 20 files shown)


zinnschlag:
In many cells (but not all) I see NPCs/creatures rendered like this:

screenshot-morrowind-anim2

Also, it seems that the stylistic issues I mentioned on the merge request on GitLab have not been addressed yet.

akortunov:
Are you sure that on your screenshot are creatures/NPCs? "Black monkey" is a vanilla placeholder for leveled creatures. Can you give me an example of such cell?

akortunov:
About stylistic issues on GitLab: I found only one, related to WeakCache location. Are there other issues?

akortunov:
> But I also don't see why we would want animation. Maybe in the preview, but definitely not in the scenes.

I encountered this issue with modded Golden saint:
screenshot_20180928_142128

Hair node in this model has a KeyframeController, which supposed to place this mesh in right place. It seems we just ignore such controllers.
The same thing in upstream code. TES CS renders this model just fine.

zinnschlag:
> "Black monkey" is a vanilla placeholder for leveled creatures.

Okay, I didn't know that. Probably because all my work with the original TES-CS avoided levelled list like the plague. I withdraw this comment then.

Regarding animation: Oh, Morrowind, why do you do this kind of crap? But agreed. Issue for another PR. Also, not very important. Blame the resources. I would be perfectly okay with releasing OpenMW-CS without a fix.

> About stylistic issues on GitLab: I found only one, related to WeakCache location. Are there other issues?

All issues are still listed on the MR. To sum it up:
* cache namespace does not conform to our naming guidelines
* WeakCache doesn't really need a component of its own (could as well go into Misc), which makes the first issue moot
* Functionality moved to components (meant for functionality shared by multiple applications/tools) but is used exclusively in OpenMW-CS

Possible resolutions for the 3rd issue are:
1. Keep it that way, because the functionality is general enough that it may become useful elsewhere
2. Move it back to OpenMW-CS
3. Use it to replace code in other applications (OpenMW primarily in this case) and thus reduce redundancies

akortunov:
WeakCache is moved to Misc. I suggest to keep it here for now.

zinnschlag:
Good enough for now.

zinnschlag:
Okay, resolved. Apparently only a short glitch.

PR #1945, opened 2018-09-27
[WIP] Implement distant terrain into launcher

terabyte25:
(Based off) [https://forum.openmw.org/viewtopic.php?f=3&t=4947]

### DIFF
--- a/apps/launcher/graphicspage.cpp
+++ b/apps/launcher/graphicspage.cpp
@@ -166,6 +166,12 @@ void Launcher::GraphicsPage::saveSettings()
     int cScreen = screenComboBox->currentIndex();
     if (cScreen != mEngineSettings.getInt("screen", "Video"))
         mEngineSettings.setInt("screen", "Video", cScreen);
+	bool cDistantTerrain = distantTerrainCheckBox->checkState();
+	if (cDistantTerrain != mEngineSettings.getBool("distant terrain", "Terrain"))
+		mEngineSettings.setBool("distant terrain", "Terrain", cDistantTerrain);
+	int cLoad = 1;
+	if (cWidth != mEngine.getInt("exterior cell load distance", "Cells"))
+		mEngineSettings.setInt("exterior cell load distance", "Cells", cLoad);
 }
 
 QStringList Launcher::GraphicsPage::getAvailableResolutions(int screen)


terabyte25:
Building it returns this:` /openmw/apps/launcher/graphicspage.cpp: In member function ‘bool Launcher::GraphicsPage::loadSettings()’:
 /openmw/apps/launcher/graphicspage.cpp:102:3: error: ‘distantTerrainCheckBox’ was not declared in this scope
   distantTerrainCheckBox->setCheckState(Qt::Checked);
   ^
 /openmw/apps/launcher/graphicspage.cpp: In member function ‘void Launcher::GraphicsPage::saveSettings()’:
 /openmw/apps/launcher/graphicspage.cpp:176:25: error: ‘distantTerrainCheckBox’ was not declared in this scope
  bool cDistantTerrain = distantTerrainCheckBox->checkState();
                         ^
 /openmw/apps/launcher/graphicspage.cpp:179:22: error: expected initializer before ‘->’ token
  int cellLoadComboBox->currentText().toInt();
                      ^
 /openmw/apps/launcher/graphicspage.cpp:180:6: error: ‘cLoad’ was not declared in this scope
  if (cLoad != mEngine.getInt("exterior cell load distance", "Cells"))
      ^
 /openmw/apps/launcher/graphicspage.cpp:180:15: error: ‘mEngine’ was not declared in this scope
  if (cLoad != mEngine.getInt("exterior cell load distance", "Cells"))
               ^
 /openmw/apps/launcher/graphicspage.cpp:182:27: error: ‘viewingDistanceSpinBox’ was not declared in this scope
  float cViewingDistance = viewingDistanceSpinBox->checkState();
                           ^
 /openmw/apps/launcher/graphicspage.cpp:185:30: error: ‘smallFeatureCullingCheckBox’ was not declared in this scope
  bool cSmallFeatureCulling = smallFeatureCullingCheckBox->checkState();
                              ^
 /openmw/apps/launcher/graphicspage.cpp:188:40: error: ‘smallFeatureCullingPixelSizeCheckBox’ was not declared in this scope
  float cSmallFeatureCullingPixelSize = smallFeatureCullingPixelSizeCheckBox->checkState();
                                        ^
apps/launcher/CMakeFiles/openmw-launcher.dir/build.make:194: recipe for target 'apps/launcher/CMakeFiles/openmw-launcher.dir/graphicspage.cpp.o' failed

`
Any ideas?

Capostrophic:
As I said to you earlier, *you need to actually add these new checkboxes to the ui file in Qt designer*, yet you said you know that you'll have to do that.

terabyte25:
Alright.

terabyte25:
Should I add in fog settings as well as exterior cell load distance into the menu?

jvoisin:
There was a super-lengthy discussion about this (distant terrain vs. fog) on the forum, I can't find it anymore, but I think that there was no consensus whether the fog should adjust automatically or not :/

terabyte25:
Sorry for the extraneous commits.

terabyte25:
Huh, why did it compile successfully for macOS and not Linux?

Thunderforge:
> Huh, why did it compile successfully for macOS and not Linux?

Looking at the logs, it seems to have failed when changing directories. That might be a fluke, although someone else should weigh in.

I think the intention with the "Enable" checkbox was that if it was checked, everything below it was accessible, but if unchecked, everything below it was grayed out. Check out the "Skip menu and generate default character" checkbox on the Advanced page for an example of what I'm talking about.

terabyte25:
So should that checkbox be tied to Distant Terrain being on or off?

zinnschlag:
> I think the intention with the "Enable" checkbox was that if it was checked, everything below it was accessible, but if unchecked, everything below it was grayed out. Check out the "Skip menu and generate default character" checkbox on the Advanced page for an example of what I'm talking about.

This sounds correct.

terabyte25:
Looks like I'll have to redo the UI visuals again because coordinate based placement isn't all that pretty.

terabyte25:
> I think the intention with the "Enable" checkbox was that if it was checked, everything below it was accessible, but if unchecked, everything below it was grayed out. Check out the "Skip menu and generate default character" checkbox on the Advanced page for an example of what I'm talking about.

Done.

akortunov:
Please, remove tabs from your PR.

terabyte25:
Fixed.

psi29a:
Everyone OK with merging this?

Thunderforge:
Can I see a screenshot of the final result?

files/ui/graphicspage.ui
```diff
@@ -124,7 +124,7 @@
         <item row="1" column="1">
          <widget class="QRadioButton" name="customRadioButton">
           <property name="text">
-           <string>Custom:</string>
+           <string>Custo&amp;m:</string>
```
Thunderforge: This doesn't look right. Did we get an extra character added when we didn't mean to?

files/ui/graphicspage.ui
```diff
@@ -146,6 +146,70 @@
...
+         <layout class="QHBoxLayout" name="smallFeatureBoxLayout"/>
+        </item>
+        <item row="0" column="0">
+         <widget class="QCheckBox" name="smallFeatureCullingCheckBox">
+          <property name="text">
+           <string>Enable</string>
+          </property>
+         </widget>
+        </item>
+        <item row="0" column="1">
```
Thunderforge: We have two items at `row="0" column="1"`. Those should have separate numbers.

files/ui/graphicspage.ui
```diff
@@ -146,6 +146,70 @@
...
+           <number>1</number>
+          </property>
+          <property name="maximum">
+           <number>16</number>
+          </property>
+         </widget>
+        </item>
+       </layout>
+      </item>
+      <item row="6" column="0">
```
Thunderforge: If the previous row is 0, the next should be 1, not 6.

files/ui/graphicspage.ui
```diff
@@ -146,6 +146,70 @@
...
+       <widget class="QLabel" name="viewingDistanceLabel">
+        <property name="text">
+         <string>Viewing Distance</string>
+        </property>
+       </widget>
+      </item>
+      <item row="6" column="1">
+       <widget class="QDoubleSpinBox" name="viewingDistanceSpinBox">
+        <property name="maximum">
+         <double>4294967296.000000000000000</double>
```
Thunderforge: How was this number chosen? Also, I know that viewing distance takes a float, but I'm wondering if we should make it an integer in the spinbox and just convert it to a double under the hood. Would users really want to type in a decimal for this?

Also, would it be better to offer several presets? For instance: `14285 (2 Cells)`. I'd ideally like for this to be something that people can use without having to consult the documentation.

akortunov: I suppose it is just a 2^32. I wonder why this value was chosen since spinbox value is the double anyway.

Thunderforge: [The documentation](https://openmw.readthedocs.io/en/master/reference/modding/settings/camera.html#viewing-distance) suggests up to 35924 (5 cells), so 2^32 seems like incredible overkill (it's over 500,000 cells).

akortunov: In theory, 10 cells should be enough. Actually, it is possible to just use an integer value in cells and then convert it to game units.

Thunderforge: I'd be in favor of changing this spinner to an integer number of cells, so long as it could handle the user putting a different value manually in their openmw.cfg file (perhaps as an "Other" value). For comparison, check out the screenshot extension code, which will respect the user manually putting in a different screenshot type in openmw.cfg aside from the expected ones.

files/ui/graphicspage.ui
terabyte25: No idea how that happened.

files/ui/graphicspage.ui
terabyte25: The second item is under the layout @173, and the first one is outside it.

files/ui/graphicspage.ui
terabyte25: What would be the maximum cell value for the SpinBox?

terabyte25: Right now I set whatever is put into the spin box will be multiplied by 8192, is this okay?

terabyte25:
screenshot_20181018_190109
@Thunderforge

terabyte25:
Maybe I should add 2 radio buttons for viewing distance in cells or pixels?

akortunov:
I see two possible ways to configure visibility distance with distant terrain:
1. **Provide a different setting for view distance when distant terrain is enabled.**
In this case in-game visibility slider should change this setting when distant terrain is enabled.
The setting should be available from launcher too.
2. **Use the same setting for visibility distance, but implement a new one - distant visibility multiplier.**
In this case we do not touch the visibility slider and actual viewing distance will be the 
`"viewing distance" * "visibility multiplier"` 
when distant terrain is enabled.

terabyte25:
I figure I'll just close this now and fix it in a new PR, while referring to this PR for what to do.

PR #1946, opened 2018-09-28
Combat AI: check for obstacle before back up

akortunov:
Fixes [bug #4656](https://gitlab.com/OpenMW/openmw/issues/4656).
This solution is far from ideal, I know, but it should be enough for pre-1.0.
Also there should not be much performance loss since usually there is not too many archers or mages in scene, which need to evade enemy right now.

Note: I do not use such approach for dodging since performance loss can be more noticable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,7 @@
     Bug #4649: Levelup fully restores health
     Bug #4653: Length of non-ASCII strings is handled incorrectly in ESM reader
     Bug #4654: Editor: UpdateVisitor does not initialize skeletons for animated objects
+    Bug #4656: Combat AI: back up behaviour is incorrect
     Bug #4668: Editor: Light source color is displayed as an integer
     Bug #4669: ToggleCollision should trace the player down after collision being enabled
     Bug #4671: knownEffect functions should use modified Alchemy skill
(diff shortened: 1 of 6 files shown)


zinnschlag:
Looks okay to me, but I would like a second opinion. @Capostrophic ?

Capostrophic:
It's OK I guess.

akortunov:
I suggest to test it a bit in different situations.

akortunov:
I tested it in Balmora and figured out that the World::castRay() ignores collisions with terrain (HeightMap). Is it intended?

akortunov:
Should be fine now.

Capostrophic:
This doesn't seem to be very effective right now. For me enemies are still happily running into the walls and back off from fort walls.

akortunov:
> This doesn't seem to be very effective right now. For me enemies are still happily running into the walls and back off from fort walls.

Can you give me a list of test cases?

zinnschlag:
I guess that was directed towards @Capostrophic ?

Capostrophic:
I gave him one, just not here.

Capostrophic:
There were issues with pulling the branch on my side. This works mostly OK. Although the case when the space behind the actor is empty but the actor can't pass through it can't be handled, a "proper" fix like using navmeshes for combat movement navigation is currently unfeasible.

AI is noticeably smarter at backing up anyways, and the performance impact if any is insignificant.

akortunov:
Resolved conflicts with master branch.

Capostrophic:
I think this should be a no-brainer to be merged?

psi29a:
Merging this no-brainer. ;)

PR #1948, opened 2018-09-28
Allow to copy paths to selected content files to clipboard in content selector (feature #2847)

Capostrophic:
[Feature #2847](https://gitlab.com/OpenMW/openmw/issues/2847).

The idea is to recover the real indexes of the selected files in the content file view and use QClipboard for manipulating the global system clipboard. For that I had to make *item(int row) in contentmodel public instead of private. "Copy path(s) to clipboard" is a new option in content file view context menu.

File paths of all selected items are copied to the clipboard separated by a newline. Content selector is used in the launcher and the editor; I think this option works pretty well.

Also I fixed a tooltip in data files page of the launcher (there was a weird typo and "profile" nomenclature is not even used anymore).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,7 @@
     Feature #1617: Editor: Enchantment effect record verifier
     Feature #1645: Casting effects from objects
     Feature #2606: Editor: Implemented (optional) case sensitive global search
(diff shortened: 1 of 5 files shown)


zinnschlag:
Okay, resolved. Apparently only a short glitch.

PR #1960, opened 2018-10-09
Replace all NULLs to nullptr

akortunov:
Since I was told many times to use nullptr instead of NULL despite most of the engine actually uses NULLs, I decided to turn all NULLs to nullptrs to avoid such confusion.

Also we may want to add the "Prefer nullptr over NULL" rule to our developer reference.


zinnschlag:
Normally we do not accept these repository wide cleanup PRs, because they have a high potential for creating merge conflict, so our policy is to only cleanup stuff you are changing anyway.

However I really like this change. It improves the code quality a lot. Unless someone has any objections I'll make the executive decision to merge this anyway (with my next batch of merges which will be tomorrow, most likely).

psi29a:
I'm OK with this. It needed to be done anyway.

PR #1962, opened 2018-10-09
Fix Dark Mode on macOS 

terabyte25:
[Bug 4655.](https://gitlab.com/OpenMW/openmw/issues/4655)

> Apps linked against macOS 10.14 or later should support both light and dark appearances. If you build your app against an earlier SDK, you can still enable Dark Mode support by including the NSRequiresAquaSystemAppearance key (with a value of false) in your app's Info.plist file. Do so only if your app's appearance looks correct when running in macOS 10.14 and later with Dark Mode enabled.

https://developer.apple.com/documentation/appkit/nsappearancecustomization/choosing_a_specific_appearance_for_your_app

### DIFF
--- a/files/mac/openmw-Info.plist.in
+++ b/files/mac/openmw-Info.plist.in
@@ -32,5 +32,7 @@
 	<string></string>
 	<key>NSHighResolutionCapable</key>
 	<true/>
+	<key>NSRequiresAquaSystemAppearance</key>
+	<true/>
 </dict>
 </plist>


terabyte25:
As I don't own a macOS system, could someone test this out? Thanks.

Thunderforge:
@terabyte25 Qt 4 does not work on macOS 10.11 or later (or something close to that), so it doesn’t support the Dark Mode introduced in 10.14 because it doesn’t run on that OS.

zinnschlag:
Is this ready to merge?

Capostrophic:
Definitely not, it wasn't actually tested.

terabyte25:
>Definitely not, it wasn't actually tested.

As I said before, I don't have a macOS system to test this on.

Thunderforge:
I'm slowly catching up on things, and hope to check this out in the near future (although other Mac users are welcome to do this sooner).

terabyte25:
@Capostrophic https://gitlab.com/OpenMW/openmw/issues/4655#note_110328968

PR #1964, opened 2018-10-12
Fix crash in the ESM reader, when list is empty

akortunov:
Fixes [bug #4677](https://gitlab.com/OpenMW/openmw/issues/4677).
The "mList.back().mCellName" makes an assumption that mList is not empty.
Plugin from mentioned bugreport has an "az_NPC_az_poor_m_06" record (this NPC is unused anyway) with malformed destination field. It seems original engine just ignores such fields, but OpenMW crashes instead.

So just print warning and ignore this field instead of crash. Works for game itself, editor and esmtool.


psi29a:
Nice, thanks! :)

PR #1966, opened 2018-10-12
Allow apps without logging system to display log messages

akortunov:
Just a small addition to my new logging system.
For now if the app has no logging system setup (currently we use it only for editor and game itself), it can not display messages via Log() function because of failing "if" statement.

This change assumes that if CurrentDebugLevel == Debug::NoLevel, the app uses just a simple cout and does not filter messages.
For example, esmtool now can display messages from ESM reader.


components/debug/debuglog.hpp
```diff
@@ -8,11 +8,11 @@ namespace Debug
 {
     enum Level
     {
-        NoLevel = 0,
         Error = 1,
         Warning = 2,
         Info = 3,
         Verbose = 4,
+        NoLevel = 5,
         Marker = Verbose
```
Mingun: Whether it is required to update value of a `Marker` also?

zinnschlag: However Mingun's question shows that the code isn't very clear. At first glance this does look like an oversight. I suggest to swap the position of NoLevel and Marker and to add an empty line between the two.

akortunov: Done.

Mingun: Nevertheless it seems to me, the code is still insufficiently clear. The idiom with a marker is often used for a mark of the maximum count of elements in enumaration, looking at a code, the same case is supposed here and it is unclear why there is an element which has value greater, than a `Marker`. As a result it seemed to me that there can be a mistake. I suggest to add the comment (preferred in Doxygen style) describing purpose of a `Marker` and why `NoLevel` has greater value than it

PR #1967, opened 2018-10-12
Re-fix water reflections while making a no-GUI screenshot (regression #4603)

Capostrophic:
[Regression #4603](https://gitlab.com/OpenMW/openmw/issues/4603).

drummyfish used a smart trick to fix water reflections in no-GUI screenshots which are used for 360 degree screenshots and savegame screenshots: he attached the reflection and refraction cameras to screenshot camera and used an invisible pseudo-loading screen to hide the fact that the main camera now rendered water reflections incorrectly. Unfortunately, it was partially broken after the freeze that appeared afterwards was fixed, so you could see the weird water reflections/refractions when you made saves or 360-degree screenshots - in order to not touch the GUI modes the return call for the invisible loading screen was placed too early and the pseudo-loading screen wasn't enabled so the rendering wasn't stopped. That was fixed.

akortunov didn't have any objections (while drummyfish appears to be MIA), so hopefully this should be good to go.

Also, I cleaned up redundant calls in screenshot(), so superfluous images aren't attached to the screenshot camera and some parameters are not set twice. This might improve the performance of making 360-degree screenshots and quicksaving a bit.

### DIFF
--- a/apps/openmw/mwgui/loadingscreen.cpp
+++ b/apps/openmw/mwgui/loadingscreen.cpp
@@ -169,19 +169,18 @@ namespace MWGui
         // We are already using node masks to avoid the scene from being updated/rendered, but node masks don't work for computeBound()
         mViewer->getSceneData()->setComputeBoundingSphereCallback(new DontComputeBoundCallback);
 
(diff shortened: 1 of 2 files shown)


psi29a:
Seems that AppVeyor is slow today...

PR #1969, opened 2018-10-14
Support soundgen calls for activators (feature #4285)

Capostrophic:
[Feature 4285](https://gitlab.com/OpenMW/openmw/issues/4285)

The idea is to iterate through creature store and look for the first creature with the same model as the one activator has, then get the played sound IDs for that creature as normal. It seems to work OK. Note that land soundgen goes unused for creatures and activators alike unlike vanilla after the changes needed for bug 2256: fixed in PR #1970.


psi29a:
Thanks!

PR #1970, opened 2018-10-14
Re-enable handling soundgen land for creatures

Capostrophic:
Apparently creatures are still supposed to play a sound when land soundgen is used in their animations. This fixes a regression: some creature death animations used land soundgen to play body fall impact sound.

It appears that soundgen: land textkeys in base animation files are used because biped creatures need them.


psi29a:
Thanks!

PR #1971, opened 2018-10-16
Try to use collisions from basic actor model if an animated one has no collisions

akortunov:
Implements [feature #4682](https://gitlab.com/OpenMW/openmw/issues/4682).

Originally a part of #1930.


psi29a:
LGTM, thanks!

PR #1974, opened 2018-10-19
Fix some issues found by Coverity Scan

akortunov:
Summary of changes:
1. Remove some redundant code (basically, unused field and constructor) from animation.cpp
2. Unify random number usage - the only place where we used the std::rand() was the particle.cpp and the Coverity Scan complained about it.
It is a bit strange since the `std::rand() / static_cast<float>(RAND_MAX)` generates a random float from 0 to 1. In other places we use a more advanced Misc::Rng::rollClosedProbability() for this approach, so it should be safe to use it here too.

Coverity Scan shows additionally two suspicious issues for me:
1. Possible integer overflow inside tinyxml. I suppose we can dismiss it since it is an external code and we usually use an external tinyxml anyway.
2. Passing struct by value [here](https://github.com/OpenMW/openmw/blob/openmw-44/components/esm/loadland.cpp#L327). It was added by [this](https://github.com/OpenMW/openmw/commit/b0641934d4a34b4e462932aa715cc0ca9d854cae) commit.
@zinnschlag, is there an error here, or this change is intended?

### DIFF
--- a/apps/opencs/editor.cpp
+++ b/apps/opencs/editor.cpp
@@ -7,6 +7,7 @@
 
 #include <components/debug/debuglog.hpp>
 #include <components/fallback/validate.hpp>
+#include <components/misc/rng.hpp>
 #include <components/nifosg/nifloader.hpp>
 
 #include "model/doc/document.hpp"
@@ -354,6 +355,8 @@ int CS::Editor::run()
 {
     if (mLocal.empty())
         return 1;
+
+    Misc::Rng::init();
 
     mStartup.show();
 
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -190,12 +190,6 @@ namespace
         {
         }
 
-        RemoveFinishedCallbackVisitor(int effectId)
-            : RemoveVisitor()
-            , mHasMagicEffects(false)
-        {
-        }
-
         virtual void apply(osg::Node &node)
         {
             traverse(node);
@@ -228,9 +222,6 @@ namespace
         virtual void apply(osg::Geometry&)
         {
         }
-
-    private:
-        int mEffectId;
     };
 
     class RemoveCallbackVisitor : public RemoveVisitor
--- a/components/nifosg/particle.cpp
+++ b/components/nifosg/particle.cpp
@@ -6,6 +6,7 @@
 #include <osg/Geometry>
 
 #include <components/debug/debuglog.hpp>
+#include <components/misc/rng.hpp>
 #include <components/nif/controlled.hpp>
 #include <components/nif/nifkey.hpp>
 #include <components/nif/data.hpp>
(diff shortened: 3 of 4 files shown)


kcat:
> Fix fossible null dereference inside the niffile.cpp: when we assign value to the r->recName, we do not check if r is null (we only have an assert in the debug mode).

The point of the assert is (should be) to not do the check in release build since it shouldn't ever fail. If it does, it's a bug elsewhere and we don't want to hide it there.

akortunov:
About tinyxml: the possible overflow is [here](https://github.com/OpenMW/openmw/blob/openmw-44/extern/oics/tinyxml.cpp#L1058).
I wonder which version of tinyxml we have - upstream ones have completely different "while" loop in this function.
Just an outdated version or Bentley-style?

akortunov:
Since our version already contains some fixes which are not part of upstream tinyxml, just backporting loop for now.

psi29a:
Can't we just use tinyxml2 (pull in to extern)?

We already use it in Debian and Ubuntu.

akortunov:
> Can't we just use tinyxml2 (pull in to extern)?
We already use it in Debian and Ubuntu.

It theory, yes, but I do not know if it will not cause issues.

akortunov:
Also tinyxml2 has the different file structure. Are you sure you use it without changes in OpenMW sources?

psi29a:
https://packages.ubuntu.com/cosmic/openmw
https://packages.debian.org/sid/openmw

> libtinyxml2.6.2v5

akortunov:
> libtinyxml2.6.2v5

It is the 1st tinyxml, not the tinyxml2.
For now we have the 2.5.3 with some fixes embedded.

psi29a:
OK, thanks for confirming that.

psi29a:
Here is the tinyxml2:
https://packages.debian.org/sid/libtinyxml2-6a

> TinyXML2 supersedes the previous TinyXML library, with various improvements:
> 
>  - Fewer memory allocations (1% - 10% compared to TinyXML)
>  - Uses less memory (about 40% of that used by TinyXML)
>  - Faster
>  - No STL requirement
>  - More modern C++, including a proper namespace
>  - Proper and useful handling of whitespace

akortunov:
> Here is the tinyxml2

Yes, and it seems it is not backward-compatible with the old one.
Basically, we have two options:
1. Update our embedded version to 2.6.2 and apply our fixes to it (probably as a separate pull request).
2. Just modify the buggy loop in this PR.

I'm fine with either options.

psi29a:
You've already done # 2, so this PR is wrapped up right?

I would leave # 1 for another PR as you suggest.

zinnschlag:
@akortunov The passing by value is intended. This is called the copy-and-swap idiom: https://stackoverflow.com/questions/3279543/what-is-the-copy-and-swap-idiom

apps/opencs/editor.cpp
```diff
@@ -355,6 +356,8 @@ int CS::Editor::run()
     if (mLocal.empty())
         return 1;
 
+    Misc::Rng::init();
```
zinnschlag: Are we even using Misc::Rng in the editor? Seems redundant. But if we in fact do use it, it probably should go into the constructor instead.

akortunov: We use RNG in components (in controllers and particles, for example). Since the editor renders scene too, I suppose it is better to init it.

akortunov:
So any objections here?

psi29a:
None from my side.

Looking forward to a future PR involving moving to tinyxml2. :)

PR #1975, opened 2018-10-19
Add collision on items (configurable)

terabyte25:
https://forum.openmw.org/viewtopic.php?f=6&t=4465&p=48339#p48339


terabyte25:
How did all those misc commits leak in?

PR #1976, opened 2018-10-19
Add collision for dropped items (configurable via settings)

terabyte25:
[#4687](https://gitlab.com/OpenMW/openmw/issues/4687)
https://forum.openmw.org/viewtopic.php?f=6&t=4465&p=48339#p48339

### DIFF
--- a/apps/openmw/mwclass/apparatus.cpp
+++ b/apps/openmw/mwclass/apparatus.cpp
@@ -1,6 +1,7 @@
 #include "apparatus.hpp"
 
 #include <components/esm/loadappa.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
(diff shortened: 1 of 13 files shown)


terabyte25:
Should this be configurable from the Advanced menu?

terabyte25:
Seems like tes3mp and OpenMW have 2 different ways of handling collision for an item. For tes3mp, actors can land normally, but in OpenMW the actors "land" on the items, and "levitate" as if they are jumping. This causes the physics system to shit itself whenever an actor is on an object.

terabyte25:
Here's the omwsave to test with for this PR. 
[save.zip](https://github.com/OpenMW/openmw/files/2497697/save.zip)

kcat:
I don't think this is the correct way to do this. The way objects have collisions is (mostly) consistent with vanilla, and I don't think a settings option is really the way to make it configurable. It should be handled by the game data since it has a significant effect on the game. It should be up to the content creator rather than the user.

davidcernat:
> Seems like tes3mp and OpenMW have 2 different ways of handling collision for an item. For tes3mp, actors can land normally, but in OpenMW the actors "land" on the items, and "levitate" as if they are jumping. This causes the physics system to shit itself whenever an actor is on an item.

No, they don't. That's just you giving the items actor-style collisions by default. Is there even a way to turn off collisions for items again in your implementation? It looks like you've made them have world collisions if the actor-style collisions are turned off. If you're going to copy-paste functionality from TES3MP, perhaps you should at least do it in a way that makes sense.

zinnschlag:
Can we please cut down on new settings? The more we add now the more we need to clean up later.

PR #1977, opened 2018-10-20
Fix #4665

terabyte25:
(Maybe) Fixes [#4665](https://gitlab.com/OpenMW/openmw/issues/4665).
This probably fixes it, but I've yet to test it. Apparently this line wasn't removed after #1771.


kcat:
I don't see what [4665](https://gitlab.com/OpenMW/openmw/issues/4665) has to do with the line you remove. The line also has a comment as to why its there, won't that break again?

PR #1978, opened 2018-10-20
Migrate to FFMPEG3

akortunov:
[Task #4686](https://gitlab.com/OpenMW/openmw/issues/4686).
Put it here to collect an eary feedback and suggestions.
For now CMake requires ffmpeg 3.2+
Unfortunately there is still no minimum version check on Windows.

### DIFF
--- a/extern/osg-ffmpeg-videoplayer/audiodecoder.cpp
+++ b/extern/osg-ffmpeg-videoplayer/audiodecoder.cpp
@@ -6,15 +6,9 @@
 
 extern "C"
 {
-
     #include <libavcodec/avcodec.h>
 
     #include <libswresample/swresample.h>
-
-    #if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(55,28,1)
-    #define av_frame_alloc  avcodec_alloc_frame
-    #endif
-
 }
 
 #include "videostate.hpp"
@@ -61,6 +55,7 @@ MovieAudioDecoder::MovieAudioDecoder(VideoState* videoState)
     , mFrameData(NULL)
     , mDataBufLen(0)
     , mFrame(av_frame_alloc())
+    , mGetNextPacket(true)
     , mAudioDiffAccum(0.0)
     , mAudioDiffAvgCoef(exp(log(0.01 / AUDIO_DIFF_AVG_NB)))
     /* Correct audio only if larger error than this */
@@ -68,10 +63,34 @@ MovieAudioDecoder::MovieAudioDecoder(VideoState* videoState)
     , mAudioDiffAvgCount(0)
 {
     mAudioResampler.reset(new AudioResampler());
+
+    AVCodec *codec = avcodec_find_decoder(mAVStream->codecpar->codec_id);
+    if(!codec)
+    {
+        std::string ss = "No codec found for id " +
+                            std::to_string(mAVStream->codecpar->codec_id);
+        throw std::runtime_error(ss);
+    }
+
+    AVCodecContext *avctx = avcodec_alloc_context3(codec);
+    avcodec_parameters_to_context(avctx, mAVStream->codecpar);
+
+// This is not needed anymore above FFMpeg version 4.0
+#if LIBAVCODEC_VERSION_INT < 3805796
+    av_codec_set_pkt_timebase(avctx, mAVStream->time_base);
+#endif
+
+    mAudioContext = avctx;
+
+    if(avcodec_open2(mAudioContext, codec, nullptr) < 0)
+        throw std::runtime_error(std::string("Failed to open audio codec ") + codec->long_name);
 }
 
 MovieAudioDecoder::~MovieAudioDecoder()
 {
+    if(mAudioContext)
+        avcodec_free_context(&mAudioContext);
+
     av_freep(&mFrame);
     av_freep(&mDataBuf);
 }
@@ -81,13 +100,13 @@ void MovieAudioDecoder::setupFormat()
     if (mAudioResampler->mSwr)
         return; // already set up
 
-    AVSampleFormat inputSampleFormat = mAVStream->codec->sample_fmt;
+    AVSampleFormat inputSampleFormat = mAudioContext->sample_fmt;
 
-    uint64_t inputChannelLayout = mAVStream->codec->channel_layout;
+    uint64_t inputChannelLayout = mAudioContext->channel_layout;
     if (inputChannelLayout == 0)
-        inputChannelLayout = av_get_default_channel_layout(mAVStream->codec->channels);
+        inputChannelLayout = av_get_default_channel_layout(mAudioContext->channels);
 
-    int inputSampleRate = mAVStream->codec->sample_rate;
+    int inputSampleRate = mAudioContext->sample_rate;
 
     mOutputSampleRate = inputSampleRate;
     mOutputSampleFormat = inputSampleFormat;
@@ -133,7 +152,7 @@ int MovieAudioDecoder::synchronize_audio()
         {
             int n = av_get_bytes_per_sample(mOutputSampleFormat) *
                     av_get_channel_layout_nb_channels(mOutputChannelLayout);
-            sample_skip = ((int)(diff * mAVStream->codec->sample_rate) * n);
+            sample_skip = ((int)(diff * mAudioContext->sample_rate) * n);
         }
     }
 
@@ -146,23 +165,31 @@ int MovieAudioDecoder::audio_decode_frame(AVFrame *frame, int &sample_skip)
 
     for(;;)
     {
-        while(pkt->size > 0)
+        /* send the packet with the compressed data to the decoder */
+        int ret = 0;
+        if (mGetNextPacket)
+            ret = avcodec_send_packet(mAudioContext, pkt);
+
+        /* read all the output frames (in general there may be any number of them */
+        while (ret >= 0)
         {
-            int len1, got_frame;
-
-            len1 = avcodec_decode_audio4(mAVStream->codec, frame, &got_frame, pkt);
-            if(len1 < 0) break;
-
-            if(len1 <= pkt->size)
+            ret = avcodec_receive_frame(mAudioContext, frame);
+            if (ret == AVERROR_EOF || ret == AVERROR(EAGAIN))
             {
-                /* Move the unread data to the front and clear the end bits */
-                int remaining = pkt->size - len1;
-                memmove(pkt->data, &pkt->data[len1], remaining);
-                av_shrink_packet(pkt, remaining);
+                // EAGAIN means that we need additional packages to decode this frame.
+                // AVERROR_EOF means the end of package.
+                mGetNextPacket = true;
+                break;
+            }
+            else if (ret < 0)
+            {
+                // Error encountered. Stop to decode audio stream.
+                av_packet_unref(&mPacket);
+                mGetNextPacket = true;
+                return -1;
             }
 
-            /* No data yet? Look for more frames */
-            if(!got_frame || frame->nb_samples <= 0)
+            if(frame->nb_samples <= 0)
                 continue;
 
             if(mAudioResampler->mSwr)
@@ -170,7 +197,7 @@ int MovieAudioDecoder::audio_decode_frame(AVFrame *frame, int &sample_skip)
                 if(!mDataBuf || mDataBufLen < frame->nb_samples)
                 {
                     av_freep(&mDataBuf);
-                    if(av_samples_alloc(&mDataBuf, NULL, av_get_channel_layout_nb_channels(mOutputChannelLayout),
+                    if(av_samples_alloc(&mDataBuf, nullptr, av_get_channel_layout_nb_channels(mOutputChannelLayout),
                                         frame->nb_samples, mOutputSampleFormat, 0) < 0)
                         break;
                     else
@@ -187,14 +214,16 @@ int MovieAudioDecoder::audio_decode_frame(AVFrame *frame, int &sample_skip)
             else
                 mFrameData = &frame->data[0];
 
-            mAudioClock += (double)frame->nb_samples /
-                           (double)mAVStream->codec->sample_rate;
+            int result = frame->nb_samples * av_get_channel_layout_nb_channels(mOutputChannelLayout) *
+                    av_get_bytes_per_sample(mOutputSampleFormat);
 
             /* We have data, return it and come back for more later */
-            return frame->nb_samples * av_get_channel_layout_nb_channels(mOutputChannelLayout) *
-                   av_get_bytes_per_sample(mOutputSampleFormat);
+            mGetNextPacket = false;
+            return result;
         }
-        av_free_packet(pkt);
+
+        av_packet_unref(&mPacket);
+        mGetNextPacket = true;
 
         /* next packet */
         if(mVideoState->audioq.get(pkt, mVideoState) < 0)
@@ -202,7 +231,7 @@ int MovieAudioDecoder::audio_decode_frame(AVFrame *frame, int &sample_skip)
 
         if(pkt->data == mVideoState->mFlushPktData)
         {
-            avcodec_flush_buffers(mAVStream->codec);
+            avcodec_flush_buffers(mAudioContext);
             mAudioDiffAccum = 0.0;
             mAudioDiffAvgCount = 0;
             mAudioClock = av_q2d(mAVStream->time_base)*pkt->pts;
--- a/extern/osg-ffmpeg-videoplayer/videostate.cpp
+++ b/extern/osg-ffmpeg-videoplayer/videostate.cpp
@@ -11,19 +11,7 @@ extern "C"
     #include <libavcodec/avcodec.h>
     #include <libavformat/avformat.h>
     #include <libswscale/swscale.h>
-
-    // From libavformat version 55.0.100 and onward the declaration of av_gettime() is
-    // removed from libavformat/avformat.h and moved to libavutil/time.h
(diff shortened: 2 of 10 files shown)


akortunov:
Rewrited audio decoding in the our videoplayer according to [this](https://www.ffmpeg.org/doxygen/3.4/group__lavc__encdec.html) article.
Now the "Multiple frames in a packet" warning should be fixed.

extern/osg-ffmpeg-videoplayer/audiodecoder.cpp
```diff
@@ -159,31 +160,48 @@ int MovieAudioDecoder::audio_decode_frame(AVFrame *frame, int &sample_skip)
...
+                continue;
+            }
+            else if (ret == AVERROR(EAGAIN))
+            {
+                // Additional frames were not decoded yet.
+                // Enter to the draining mode.
+                avcodec_send_packet(mAudioContext, NULL);
+                ret = 0;
+                continue;
+            }
```
kcat: This doesn't look right/logical. `EAGAIN` from `avcodec_receive_frame` means another packet needs to be sent before a frame can be decoded. Sending a NULL packet means the stream is over, so you get an `EOF` that's supposed to tell you to stop. It looks like you're abusing the stream-is-done message to signal to get another packet. `avcodec_flush_buffers` also flushes and restarts decoding, which would cause partial frames, that haven't been fully decoded yet, to drop.

More generally, it doesn't make sense for the decoder thread to be sending the encoded packets and receiving the decoded frames. For example, [note here](https://github.com/kcat/openal-soft/blob/master/examples/alffplay.cpp#L1320) how it holds in a loop while getting `EAGAIN`, signalling to the packet pump thread to send more packets and waiting for it to do so. When it gets a frame, it signals the pump thread to go and send another while it does its thing. When it gets `EOF`, the stream is over and it ends. Then [here](https://github.com/kcat/openal-soft/blob/master/examples/alffplay.cpp#L1523) the packet pump thread gets a packet from the input file, queues it, then sends whatever packets it can to the available stream handlers. If the combined queues hold enough compressed data, it simply waits until it can get rid of some before reading more (to avoid reading the whole file into memory).

This setup improves multithreading performance since it splits the workload across threads. With the decoder thread doing the sending and receiving, you're serializing the workload:
```
send_packet           receive_frame....send_packet           receive_frame....
          |           ^                          |           ^
          v           |                          v           |
          decode packet                          decode packet
```
Instead, you want to be able to interleave the work:
```
send_packet .. send_packet
          |              |
          v              v
          decode packet  decode packet
                      |                |
                      v                v
                      receive_frame .. receive_frame ..
```
That helps makes sure everything stays busy, avoiding wasted cycles while waiting for another thread to complete what you just told it to do.

akortunov:
>  This doesn't look right/logical. EAGAIN from avcodec_receive_frame means another packet needs to be sent before a frame can be decoded. Sending a NULL packet means the stream is over, so you get an EOF that's supposed to tell you to stop. It looks like you're abusing the stream-is-done message to signal to get another packet. avcodec_flush_buffers also flushes and restarts decoding, which would cause partial frames, that haven't been fully decoded yet, to drop.

Ok, then how should I handle the "Multiple frames in a packet" issue? 
I receive a packet, start do decode it and get the EAGAIN (which indicates that I need an additional packet to decode a frame).
Then I get a next packet, but avcodec_send_packet() complains about unread data from the previous one.

kcat:
> I receive a packet, start to decode it and get the EAGAIN (which indicates that I need an additional packet to decode a frame).
> Then I get a next packet, but avcodec_send_packet() complains about unread data from the previous one.

If `avcodec_receive_frame` returns `EAGAIN`, `avcodec_send_packet` should accept another packet. You should keep calling `avcodec_send_packet` for a given codec until it returns `EAGAIN`. Then you keep calling `avcodec_receive_frame` (and processing each received frame) for a given codec until it returns `EAGAIN`. When you're completely out of input packets and have nothing more to decode, call `avcodec_send_packet` on each codec with `nullptr` to finish input. `avcodec_receive_frame` will then return any final decoded frame data it has, before returning `EOF` which tells you there's nothing more to output.

[Here's a rough synopsis](https://blogs.gentoo.org/lu_zero/2016/03/29/new-avcodec-api/).
* You setup the decoder or the encoder as usual
* You feed data using the avcodec_send_* functions until you get a AVERROR(EAGAIN), that signals that the internal input buffer is full.
* You get the data back using the matching avcodec_receive_* function until you get a AVERROR(EAGAIN), signalling that the internal output buffer is empty.
* Once you are done feeding data you have to pass a NULL to signal the end of stream.
* You can keep calling the avcodec_receive_* function until you get AVERROR_EOF.
* You free the contexts as usual.

The "simple decoding loop" approach is as simple as it gets with replicating the old API. The "separated threads" approach is effectively what I described above and what I do with alffplay (the "feeding loop" goes on one thread to read from the file and marshal packets to the appropriate codecs, and the "decoding loop" is on another thread that handles the given stream).

akortunov:
> If avcodec_receive_frame returns EAGAIN, avcodec_send_packet should accept another packet. You should keep calling avcodec_send_packet for a given codec until it returns EAGAIN. Then you keep calling avcodec_receive_frame (and processing each received frame) for a given codec until it returns EAGAIN. When you're completely out of input packets and have nothing more to decode, call avcodec_send_packet on each codec with nullptr to finish input. avcodec_receive_frame will then return any final decoded frame data it has, before returning EOF which tells you there's nothing more to output.

Unfortunately, it did not help - I still get the "Multiple frames in a packet" warning inside the avcodec_send_packet().
Also I do not know how to wrap avcodec_send_packet() to loop properly since I have only reference to one packet at time - the mPacket field.

kcat:
> Unfortunately, it did not help - I still get the "Multiple frames in a packet" warning inside the avcodec_send_packet().

I don't think that message is significant. It means an encoded data packet resulted in multiple decoded frames, and we have no control over what's in a packet since we get it from ffmpeg's demuxer. The `avcodec_send_packet`/`avcodec_receive_frame` API takes care of frame reordering and such for us, we just need to keep sending packets for as long as the codec will accept them, and keep receiving frames for as long as the codec provides them. When either end receives `EAGAIN`, it means the other end is busy doing work.

> Also I do not know how to wrap avcodec_send_packet() to loop properly since I have only reference to one packet at time - the mPacket field.

The packet handling would need to be modified (simplified). [This class](https://github.com/kcat/openal-soft/blob/master/examples/alffplay.cpp#L183) is a fairly simple packet queue handler; it has a method to push a packet reference to the back of the queue, a method to view the packet at the front of the queue (i.e. the next one to send), and a method to pop said packet (for when it successfully sends). It also tracks the total size of the encoded packet data currently stored in the queue, to help avoid storing too much of the input file.

[This code](https://github.com/kcat/openal-soft/blob/master/examples/alffplay.cpp#L1523) is then responsible for reading packets from the input stream, queueing them into the appropriate stream's queue, and feeding them from each queue to its pairing codec as needed. Note that unlike currently, the same thread pulls the packets out of the queue that put them in, so the queues doesn't need to worry about synchronization anymore. The `send`/`receive` calls effectively move the data from one thread to another, so the synchronization only needs to protect against calling `send` on one thread as another calls `receive` (which has the nice side effect of providing a safe way to signal a condition variable to indicate more data is needed or ready, since it would happen around those calls).

akortunov:
I updated CI for Mac OS, but it seems linking fails because of missing symbols. Any ideas?

nikolaykasyanov:
@akortunov I'll look into this.

nikolaykasyanov:
@akortunov fixed & tested locally, intro video seems to be working fine. Figuring out how to get it uploaded to downloads.openmw.org...

akortunov:
Just for reference: I'd pereper to use the same ffmpeg version on different OSes.

nikolaykasyanov:
@akortunov I agree, what's the preferred version?

akortunov:
Any 3.x version should be fine, but since we already use the [3.2.4](https://github.com/OpenMW/openmw/commit/ae05c37a468aa22fd8cdb1bf2d6c3051a4b07651) on Windows more than year without complaints, I suggest to use this version for other OSes, if it is possible.

psi29a:
> @akortunov fixed & tested locally, intro video seems to be working fine. Figuring out how to get it uploaded to downloads.openmw.org...

Just upload it and ping me with the name of the file. I'll copy it over manually

nikolaykasyanov:
@akortunov https://downloads.openmw.org/osx/dependencies/openmw-deps-938adba.zip, this one uses FFmpeg 3.2.4.

psi29a:
sweet... no more ftp for uploads, pure scp

psi29a:
Oh.. and there should be ffmpeg3 debs in our Trusty repo with Xenial on the way. I hope it works...

psi29a:
Bingo! Linux builds are gold!  Good job so far! :)

akortunov:
OK, CI is up.

I wonder if there is a way to specify required ffmpeg version since there will be compile errors with ffmpeg 2.x.

akortunov:
Looks like Windows fails to fill "FFmpeg_*_VERSION" variables. I wonder, why...

kcat:
Looks like it only [gets them with pkgconfig](https://github.com/OpenMW/openmw/blob/master/cmake/FindFFmpeg.cmake#L40):
```
macro(_internal_find_component _component _pkgconfig _library _header)
    set(_package_component FFmpeg_${_component})
    libfind_pkg_detect(${_package_component} ${_pkgconfig}
        FIND_PATH ${_header}
            HINTS $ENV{FFMPEG_HOME}
            PATH_SUFFIXES include ffmpeg
        FIND_LIBRARY ${_library}
            HINTS $ENV{FFMPEG_HOME}
            PATH_SUFFIXES lib
    )
    set(${_package_component}_DEFINITIONS  ${${_package_component}_PKGCONF_CFLAGS_OTHER})
    set(${_package_component}_VERSION      ${${_package_component}_PKGCONF_VERSION})
    libfind_process(${_package_component})
endmacro()
```
I guess Windows/MSVC that doesn't use pkgconfig doesn't get `*_PKGCONF_VERSION` set? I use an [older version of FindFFmpeg](https://github.com/kcat/openal-soft/blob/master/cmake/FindFFmpeg.cmake), which appears to directly parse the version.h headers from ffmpeg to get the versions; perhaps the updated one using libfind doesn't and can only get the version through pkgconfig.

akortunov:
A small update. I tested this PR with different FFmpeg versions. 
With 3.1 and earlier there is no sound in videos, 3.2 or higher versions seem to work fine.
So I was forced to update minimum requirement to 3.2.

psi29a:
So MacOS needs to have their libs updated again?

akortunov:
> So MacOS needs to have their libs updated again?

I thought MacOS uses 3.2.4 now too:
https://github.com/OpenMW/openmw-deps-mac/pull/43

psi29a:
Apparently not?

> -- Found FFmpeg_AVCODEC 57.64.101
> -- Found FFmpeg_AVFORMAT 57.56.101
> -- Found FFmpeg_AVUTIL 55.34.101
> -- Found FFmpeg_SWSCALE 4.2.100
> -- Found FFmpeg_SWRESAMPLE 2.3.100
> -- Found FFmpeg: /tmp/openmw-deps/openmw-deps/lib/libavcodec.a;/tmp/openmw-deps/openmw-deps/lib/libavformat.a;/tmp/openmw-deps/openmw-deps/lib/libavutil.a;/tmp/openmw-deps/openmw-deps/lib/libswscale.a;/tmp/openmw-deps/openmw-deps/lib/libswresample.a  found components:  AVCODEC AVFORMAT AVUTIL SWSCALE SWRESAMPLE 
> -- libavformat is too old! (57.56.101, wanted 57.71.100)
> -- libavcodec is too old! (57.64.101, wanted 57.89.100)
> -- libavutil is too old! (55.34.101, wanted 55.58.100)
> -- libswscale is too old! (4.2.100, wanted 4.6.100)
> -- libswresample is too old! (2.3.100, wanted 2.7.100)
> CMake Error at CMakeLists.txt:195 (message):
>   FFmpeg version is too old, 3.2 is required

akortunov:
It seems the issue in my side.

akortunov:
Should work now.

psi29a:
Wooo... supporting FFMpeg4 now too? :)

akortunov:
> Wooo... supporting FFMpeg4 now too? :)

Capostrophic uses FFMpeg 4, and there is just a couple of additional warnings: 
two functions are now no-ops due to API changes, one field was replaced by more sophisticated one.
I took these changes from [here](https://github.com/illuusio/ffmpeg-example/blob/master/example1.c).

psi29a:
Super! Thanks for future-proofing this.

psi29a:
Do you consider this 'ready' for 0.45 or wait until 0.45 goes into RC for 0.46?

akortunov:
> Do you consider this 'ready' for 0.45 or wait until 0.45 goes into RC for 0.46?

I'd prefer to go to 0.46 - this PR touches a quite a lot of things and can break something.

nikolaykasyanov:
@akortunov could you please update macOS dependencies url once more? I've merged my PR and it's just changing the commit hash, basically, contents are the same. 
http://downloads.openmw.org/osx/dependencies/openmw-deps-1641627.zip

akortunov:
> https://downloads.openmw.org/osx/dependencies/openmw-deps-1641627.zip

Done.

nikolaykasyanov:
@akortunov I've updated the URL, now it complains that MR is unmergable.

nikolaykasyanov:
I could resolve it but not sure about your plans (merge? rebase?).

akortunov:
Using this URL for now: https://downloads.openmw.org/osx/dependencies/openmw-deps-7cf2789.zip

psi29a:
Ready for merging for 0.46?

extern/osg-ffmpeg-videoplayer/videostate.cpp
```diff
@@ -403,24 +402,29 @@ class VideoThread : public OpenThreads::Thread
...
+            while (!ret)
+            {
+                ret = avcodec_receive_frame(self->video_ctx, pFrame);
+                if (!ret)
+                {
+                    double pts = 0;
+                    if(packet->dts != AV_NOPTS_VALUE)
+                        pts = static_cast<double>(packet->dts);
+                    else if(pFrame->opaque && *(int64_t*)pFrame->opaque != AV_NOPTS_VALUE)
+                        pts = static_cast<double>(*(int64_t*)pFrame->opaque);
```
kcat: You should be able to use the frame's `best_effort_timestamp` field here, rather than trying to decide between the `dts` field and what was stored in the `opaque` pointer. That should also allow getting rid of the `our_get_buffer` hack that was needed for older ffmpeg versions.

akortunov: Using the best_effort_timestamp now. A quick test did not show any issues.

extern/osg-ffmpeg-videoplayer/videostate.cpp
```diff
@@ -340,29 +340,7 @@ double VideoState::synchronize_video(AVFrame *src_frame, double pts)
     return pts;
 }
 
-static void our_free_buffer(void *opaque, uint8_t *data);
-/* These are called whenever we allocate a frame
- * buffer. We use this to store the global_pts in
- * a frame at the time it is allocated.
- */
 static int64_t global_video_pkt_pts = AV_NOPTS_VALUE;
```
kcat: This is now also unnecessary (set but never used).

kcat:
Nothing more sticks out to me at a cursory glance.

akortunov:
> Nothing more sticks out to me at a cursory glance.

I suppose this PR should be basically OK them.
It would be nice if anyone could test if it work on Windows.

psi29a:
Waiting for Windows confirmation.

akortunov:
I convinced one guy to test this PR.
With MSVC2017 it works fine, but with MSVC2015 it crashes after Bethesda's logo.
@AnyOldName3, can you confirm the crash?

psi29a:
You're merging against master right?
https://travis-ci.org/OpenMW/openmw/jobs/448956339 <-- from your last build

akortunov:
>  You're merging against master right?

I did not merge anything, but I suppose CI did.
And if I will re-trigger this build, the crash should be fixed, right?

psi29a:
You want me to be optimistic or pessimistic? You can always retry in order to validate, there have been fixes recently including msvc2015 I believe.

akortunov:
> You want me to be optimistic or pessimistic? You can always retry in order to validate, there have been fixes recently including msvc2015 I believe.

I just hope that it is a MSVC2015-specific bug with RecastNavigation, that was fixed recently.

psi29a:
If you want, you can open a PR against openmw-45 branch to validate as well. Confirm the two builds. If the one against openmw-45 works and the one against master doesn't, then we can point the finger at recast.

akortunov:
If it is really a RecastNavigation stuff, it should not crash in the build which I triggered now.
There will be a need to retest artifacts before we will start investigation.

AnyOldName3:
> I convinced one guy to test this PR.
> With MSVC2017 it works fine, but with MSVC2015 it crashes after Bethesda's logo.
> @AnyOldName3, can you confirm the crash?

That looks like the Recast crash to me, so it should be gone if this is merged into master. I'll give it a try on my end.

AnyOldName3:
Yup, the crash is gone now.

akortunov:
> I convinced one guy to test this PR.
With MSVC2017 it works fine, but with MSVC2015 it crashes after Bethesda's logo.

For this guy this PR works fine on Windows too now.
Ready for merge?

psi29a:
Great, merging. 

Hold on to your butts, this guy is merge-happy! ;)

PR #1979, opened 2018-10-20
Utilize the default soundgen entries when necessary (bug #4689)

Capostrophic:
[Bug 4689](https://gitlab.com/OpenMW/openmw/issues/4689)

If a sound generator record isn't found for the specific creature ID, the entries which don't have an ID assigned but have the same soundgen type are used now, fixing footstep sounds for ash creatures, for example. The logic is replicated for activators too.

NPCs don't use sound generator records.

Addendum: the more apparent cases are that other than ash creatures, daedroths, corprus creatures, Flame and Frost Atronachs, bonelords, bonewalkers, Dagoth Ur and other Sixth House members, Dremoras and Nix-Hounds and Vivec (and probably more) don't play footstep sounds in master (or any sound generator sounds in Vivec's case) whatsoever.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,6 +144,7 @@
     Bug #4677: Crash in ESM reader when NPC record has DNAM record without DODT one
     Bug #4678: Crash in ESP parser when SCVR has no variable names
     Bug #4685: Missing sound causes an exception inside Say command
+    Bug #4689: Default creature soundgen entries are not used
     Feature #912: Editor: Add missing icons to UniversalId tables
     Feature #1221: Editor: Creature/NPC rendering
     Feature #1617: Editor: Enchantment effect record verifier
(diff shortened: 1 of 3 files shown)


Capostrophic:
Don't merge this yet. Apparently this was an intentional change to fix [bug 2228](https://gitlab.com/OpenMW/openmw/issues/2228). So hungers oddly use werewolf sounds now.

Capostrophic:
I haven't noticed any differences from vanilla soundgen sound fallback for ascended sleepers and hungers.

Capostrophic:
Yep. That wasn't a regression. Vanilla also plays the fallback werewolf-sounding sounds for ascended sleepers and hungers. scrawl broke that.

The hardcoded landing sound fallback that is currently used instead because scribs need it for extremely adorable thump idle animation isn't even correct.

PR #1980, opened 2018-10-22
Allow creatures to use the autogenerated collision box

akortunov:
Implements [feature #2787](https://gitlab.com/OpenMW/openmw/issues/2787).
The main idea: if collision box was not found for creature, take AABB from shape from ShapeManager.

Created as a replacement of #1930 since this PR is broken for me (comments and commits randomly disappears after page refresh).

Note: you can not use autogeneration just for any arbitrary creature: in both Morrowind and OpenMW this feature does not consider bones transformation and skinning.
For example, Tyddy does not use bones transformation in his creature replacers, so this feature works fine there.

In theory, there could be an additional tricky differences, but I doubt I can do anything here without reverse-engineering.


akortunov:
Is there a way to restart CI?

terabyte25:
@akortunov add a dummy commit?

psi29a:
Yeah, I have access to restart. But @terabyte25 is correct, an empty commit can re-trigger a build.

akortunov:
```
The following packages have unmet dependencies:
 libavcodec-dev : Depends: libavcodec57 (= 7:3.4.4-1~14.04.york0) but it is not going to be installed or
                           libavcodec-extra57 (= 7:3.4.4-1~14.04.york0) but it is not going to be installed
 libavformat-dev : Depends: libavformat57 (= 7:3.4.4-1~14.04.york0) but it is not going to be installed
```

Something fishy here on Clang. Why it tries to use the ffmpeg 3.4.4 header packages?

akortunov:
> Have a look at these warnings that clang gives us:
https://travis-ci.org/OpenMW/openmw/jobs/444633343

Almost all these warnings (excepts of googletest false-positive) are handled by my two other PR's.

psi29a:
So this PR is ready for review?

akortunov:
> So this PR is ready for review?

Yes, it is.

psi29a:
This looks OK to me. 

Should we warn the user that a collision mesh had to be made? That way users can pester modders to fix it, modders can test it out in openmw and openmw-cs and see that they forget to add one of their own.

akortunov:
> Should we warn the user that a collision mesh had to be made? That way users can pester modders to fix it, modders can test it out in openmw and openmw-cs and see that they forget to add one of their own.

In this case there will be a lot of spam in log. For now I log an entry only if autogeneration failed for some reason.

zinnschlag:
I agree with @akortunov. Eventually we will want to have such a check in the editor. In OpenMW it will produce too much spam. Going to merge this PR.

PR #1981, opened 2018-10-23
[Regression] Fix fixme behavior in interiors

Capostrophic:
This passed by unnoticed when the changes were tested. ActionTeleport was using an empty cell name instead of the current cell name and this only works properly for exteriors; in interiors, it caused the player to be teleported to 0, 0 because the absolute coordinates in interiors are very small.


terabyte25:
Confirmed working.

PR #1982, opened 2018-10-24
Avoid item condition and charge zero divisions

Capostrophic:
While normally this shouldn't be necessary, in the hypothetical situations the maximum values of these stats are 0 and the divided value is e.g. also 0 and you are able to use them the zero division results in a +inf value and all sorts of weird issues.

I actually ran into such situation while testing a custom weapon record in tes3mp that I forgot to set maximum condition to. It got 0/0 condition and the damage was infinite, which was pretty funny I admit. That weapon wasn't considered "broken" until I gave it away to a different player, seems that some kind of issue (probably not on openmw's side) didn't properly assign the charge to the item cell reference.

This doesn't actually affect anything in the frontend of openmw.

### DIFF
--- a/apps/openmw/mwclass/armor.cpp
+++ b/apps/openmw/mwclass/armor.cpp
@@ -297,7 +297,7 @@ namespace MWClass
     {
         const MWWorld::InventoryStore& invStore = npc.getClass().getInventoryStore(npc);
 
-        if (ptr.getCellRef().getCharge() == 0)
+        if (getItemHealth(ptr) == 0)
             return std::make_pair(0, "#{sInventoryMessage1}");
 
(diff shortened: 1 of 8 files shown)


Capostrophic:
I'll add getNormalizedHealth and getNormalizedEnchantmentCharge methods to the classes then.

Capostrophic:
Ouch. I accidentally triggered 5 redundant CI builds due to suddenly noticing issues that compile but break everything too late multiple times. :(

psi29a:
Fixing...

psi29a:
canceled the others, no worries... AV is still working on your "optimize actors processing"

PR #1983, opened 2018-10-26
Decrease dispostion of assault victim even if Alarm = 0

akortunov:
Fixes [bug #4683](https://gitlab.com/OpenMW/openmw/issues/4683).

Currently we decrease disposition [here](https://github.com/OpenMW/openmw/blob/openmw-44/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp#L1357). It means that we decrease disposition only for NPCs with high Alarm rating, which will attack player if he will commit crime.
However, vanilla game decreases disposition for any victim of assault, even if Alarm = 0 and crime was not reported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,6 +143,7 @@
     Bug #4674: Journal can be opened when settings window is open
     Bug #4677: Crash in ESM reader when NPC record has DNAM record without DODT one
     Bug #4678: Crash in ESP parser when SCVR has no variable names
+    Bug #4683: Hitting NPC with Alarm 0 should reduce disposition
     Bug #4685: Missing sound causes an exception inside Say command
     Bug #4689: Default creature soundgen entries are not used
     Feature #912: Editor: Add missing icons to UniversalId tables
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -1410,6 +1410,12 @@ namespace MWMechanics
                     // Mark as Alarmed for dialogue
                     observerStats.setAlarmed(true);
                 }
+                else if (type == OT_Assault && *it == victim)
+                {
+                    // Still decrease disposition for victim of attack, even if Alarm = 0
+                    NpcStats& observerStats = it->getClass().getNpcStats(*it);
+                    observerStats.setBaseDisposition(observerStats.getBaseDisposition() + static_cast<int>(dispTerm));
+                }
             }
         }
 


Capostrophic:
Are you sure this reduces disposition for reacting guards correctly? They should have -50 disposition hit like any observer yet not attack the player. I ran into this when I tried to make the code closer to the research. Also, the disposition hit should be reset for observers that are not the victim and don't have 100 fight rating (or guards with non-100 alarm rating)

akortunov:
> Are you sure this reduces disposition for reacting guards correctly? 

For guards we do not change disposition at all. Otherwise we should restore disposition for them when player pays fee.

Capostrophic:
>For guards we do not change disposition at all.

Too bad, because Morrowind does.

akortunov:
> Too bad, because Morrowind does.

Do you know where it stores a previous disposition?

akortunov:
If Morrowind restores disposition for actors if you pay your free, we should not touch base disposition - we should use the crime ID to modify only current disposition level on the fly.

Capostrophic:
>Also does Morrowind restore disposition for non-guard actors?

```
    if observer is not victim or observer.isGuard:
        dispTerm *= alarmTerm

    fightTerm += fFightDispMult * (50 - dispTerm)
    x = iFightDistanceBase - fFightDistanceMultiplier * distance(player, observer)
    fightTerm = alarmTerm * (fightTerm + x)

    if observer.fight + fightTerm > 100:
        fightTerm = 100 - observer.fight
    fightTerm = max(0, fightTerm)

    observer.fight += int(fightTerm)
    observer.disposition += int(dispTerm) <==

    - lots more crime system stuff occurs here -

    if (observer.alarm < 100 or not observer.isGuard) and observer.fight < 100:
        observer.fight is reset to the original value
        if observer is not victim: observer.disposition is reset to the original value <==
```

akortunov:
So actually behaviour here is inconsistent a bit:
1. If the victim is a guard, reduce disposition by 10 and restore it back if you pay fee.
2. If the victim is not a guard, reduce disposition by 10 permanently, even if Alarm = 0.
3. Any observer (not victim) get -50 disposition and isposition will be restored if you pay your fee.

For now we do not restore disposition at all and even do not decrease it for guards.

akortunov:
I suggest to decrease disposition of assault victim permanently, and move other disposition changes to the MechanicsManager::getDerivedDisposition(). Is it OK?

akortunov:
But in this case we should store not only crime ID, but a crime type too.

akortunov:
Closing this PR since the problem is a more complex an requires a more complex solution.

PR #1986, opened 2018-10-26
[Regression] Handle RootCollisionNode exactly as in original engine

akortunov:
My PR #1826 makes an assumption:
```
Consider root NiNode with only other NiNode child as a wrapper, not as actual root node.
```

It allows to solve many issues with mesh replacers, which use over-detailed collision meshes, but causes other issues.
For example, attached door mesh from [bug #4694](https://gitlab.com/OpenMW/openmw/issues/4694) has a bugged RootCollisionNode. Morrowind ignores it since it is not attached to root node, so mod author did not bother to fix it.

So suggested solution:
1. Revert #1826.
2. Display a warning about RootCollisionNode, as we do now.

In this case we will replicate vanilla behaviour, but still will warn players about buggy meshes.
It is either modder's or player's work to fix such meshes.


psi29a:
A good compromise. Modders need to be made aware that their mods are broken.

akortunov:
I wonder why Bullet tests are failing - I just reverted one commit, which was added after these tests were implemented.
It does not like how much times getFilename() was called?

akortunov:
Attempt number two.

akortunov:
OK, one failing test left:
```
TestBulletNifLoader.for_tri_shape_child_node_with_extra_data_string_mrk_and_other_collision_node_should_return_shape_with_triangle_mesh_shape_with_all_meshes
Found RootCollisionNode attached to non-root node in test.nif. Treat it as a common NiTriShape.
/home/travis/build/OpenMW/openmw/apps/openmw_test_suite/nifloader/testbulletnifloader.cpp:949: Failure
Expected equality of these values:
  *result
    Which is: Resource::BulletShape {nullptr, osg::Vec3f {0, 0, 0}, std::map<int, int> {}}
  expected
    Which is: Resource::BulletShape {&Resource::TriangleMeshShape {btVector3 {1, 1, 1}, 1, 1, {btVector3 {0, 0, 0}, btVector3 {1, 0, 0}, btVector3 {1, 1, 0}, }}, osg::Vec3f {0, 0, 0}, std::map<int, int> {}}
[  FAILED  ] TestBulletNifLoader.for_tri_shape_child_node_with_extra_data_string_mrk_and_other_collision_node_should_return_shape_with_triangle_mesh_shape_with_all_meshes (0 ms)
```

I wonder how this test is ever supposed to work - we can init the TriangleMeshShape only when handle NiTriShapes.

The file structure:

```
NiNode node
    NiNode node2
        NiRootCollisionNode node3
             NiTriShape shape
                 NiStringExtraData "MRK"
```

There is only one NiTriShape in file - the one with "MRK" data, so we should ignore it during collision calculation.
In master it works only because we consider "node" as a wrapper and just consider "node3" as a valid RCN, what we should not do.
Should I change this testcase?

psi29a:
Ready to be merged? Do we need more testing?

psi29a:
Is @elsid available on the forums? PM?

elsid:
@akortunov Yes, you should change it. Those tests are unit and only reflect current code logic, so if you change behaviour, some test should fail. My purpose was to add new logic without breaking old. I didn't test original Morrowind logic. And your change is ok.

psi29a:
Merging then. Thanks @elsid and @akortunov

PR #1988, opened 2018-10-26
Adjust magic light source linear attenuation (bug #3890)

Capostrophic:
[Bug 3890](https://gitlab.com/OpenMW/openmw/issues/3890).

Make the linear attenuation formula as close as possible to the original, where 1 point of Light magnitude corresponds to 1 foot of the light source range. Note that global attenuation settings are not used still. Note that cut-off distance is artificially increased because the cut-off was really obvious with the normal radius (radius adjustment doesn't affect the light source brightness). I cannot present comparison pictures at the moment.

### DIFF
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -1789,9 +1789,12 @@ namespace MWRender
         }
         else
         {
-            effect += 3;
-            float radius = effect * 66.f;
-            float linearAttenuation = 0.5f / effect;
+            // TODO: use global attenuation settings
+
+            // 1 pt of Light effect magnitude corresponds to 1 foot of light source radius, which is about 21.33 game units,
+            // but Morrowind uses imprecise value of foot for magic effects.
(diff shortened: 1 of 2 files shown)


apps/openmw/mwrender/animation.cpp
```diff
@@ -1789,9 +1789,12 @@ namespace MWRender
...
         else
         {
-            effect += 3;
-            float radius = effect * 66.f;
-            float linearAttenuation = 0.5f / effect;
+            // TODO: use global attenuation settings
+
+            // 1 pt of Light effect magnitude corresponds to 1 foot of light source radius, which is about 21.33 game units,
+            // but Morrowind uses imprecise value of foot for magic effects.
+            float radius = effect * 22.f;
```
akortunov: I'd prefer to use UnitsPerFoot here (or ceil(UnitsPerFoot), if we really need 22 here).
The same thing with comment - there is no need to write foot size in units here.

PR #1990, opened 2018-10-27
Avoid making expensive visitEffectSources calls if no spell absorption effect is active

Capostrophic:
For spell absorption effect to be multiplicative, the engine needs to iterate through each and every effect source to make sure all of them are accounted for. But it doesn't need to do this if there isn't any absorption effect active.


akortunov:
You can just modify the "if" statement, in this case amount of changes will be much smaller.

Capostrophic:
>You can just modify the "if" statement, in this case amount of changes will be much smaller.

I can't initialize creature stats reference before I know for sure that the target is in fact an actor and
a call like
```
target.getClass().getCreatureStats(target).getMagicEffects().get(ESM::MagicEffect::SpellAbsorption).getMagnitude()
```
from within a condition doesn't look very nice.

PR #1993, opened 2018-10-28
Center progress bar when there are active messageboxes

akortunov:
Fixes [bug #4691](https://gitlab.com/OpenMW/openmw/issues/4691).
The approach is very simple:
1. Check if there are active messageboxes on the screen.
2. If yes, place loading bar to center of screen instead of bottom.

Affects both save process and cell change process.

### DIFF
--- a/apps/openmw/mwbase/windowmanager.hpp
+++ b/apps/openmw/mwbase/windowmanager.hpp
@@ -285,6 +285,8 @@ namespace MWBase
 
             virtual void setEnemy (const MWWorld::Ptr& enemy) = 0;
 
+            virtual int getMessagesCount() const = 0;
+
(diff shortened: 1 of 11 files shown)


apps/openmw/mwbase/windowmanager.hpp
```diff
@@ -285,6 +285,8 @@ namespace MWBase
 
             virtual void setEnemy (const MWWorld::Ptr& enemy) = 0;
 
+            virtual int getMessagesCount() = 0;
```
zinnschlag: Shouldn't this function be const?

akortunov: Fixed.

PR #1996, opened 2018-10-28
Clean up redundant getClass() calls in classes

Capostrophic:
In lion's share of cases this must not be necessary because we already know what the class is. It makes no sense to make calls like `ptr1.getClass().doSomething(ptr2)` or `ptr.getClass().doSomething(nullptr)` elsewhere (not in the class itself), so we effectivelly make calls to the functions that are already present in the very same class. It especially not very useful to make such calls if we already made calls to stuff that would throw an exception if the input is invalid earlier.

So, for example, NPCs always have inventory store and there's no need to check for that since the relevant function will always return true.


psi29a:
Nice work! Thanks! :)

PR #1998, opened 2018-10-28
Set Shift-C and Shift-V default keybindings for View Record and Preview Record (feature #2845)

Capostrophic:
[Feature 2485](https://gitlab.com/OpenMW/openmw/issues/2845)

Shift-C short for Cell (since View opens a view of the instance in the cell) and Shift-V short for View (since P is pretty far from the left-side Shift which is used far more often and V fits the function anyways).

Yes, just that. I figured that other keybindings shouldn't be touched without a sheer necessity.


psi29a:
I see the usefulness here but I would rather have @zinnschlag 's OK first.

zinnschlag:
I think I explained my position towards keyboard shortcuts several time already, but I happily doing it again: This should be done last. We need to make a list of all functions (global and sub-view specific) and then see how we best arrange the keyboard shortcuts in a way that avoids collisions and is also consistent across the whole application.

That being said there is no harm in adding some shortcuts now and since this PR is already here, I can as well go ahead and merge it. However I suggest to hold back on adding further shortcuts in the future.

PR #1999, opened 2018-10-29
[Regression] Make sure we reset current weapon animation when resurrect actor

akortunov:
Not sure how to treat this bug.

Technically, it is a part of [bug #2626](https://gitlab.com/OpenMW/openmw/issues/2626) (which is claimed fixed in the 0.45), but we also have a separate [bugreport](https://gitlab.com/OpenMW/openmw/issues/4698) for it.
So I suggest to do not add a separate changelog entry and close the bug 4698 once this PR will be merged.

Note: player will play equipping animation after resurrection if the weapon was drawn. In the Morrowind behaviour is the same.


psi29a:
You're right on your approach, if the regression happens in master then no need for a new changelog entry. New issue gets marked as duplicate/closed and mentioned in old issue.

PR #2000, opened 2018-10-30
Add prison marker record fallback definition (bug #4701)

Capostrophic:
So this is the PR 2000. Huh.

[Bug #4701](https://gitlab.com/OpenMW/openmw/issues/4701).

I simply added PrisonMarker in the style of "framework" akortunov has already set up. This should resolve the problem the reporter is having.


akortunov:
IIRC, prison marker is a door, not a static.

Capostrophic:
You're right. Rewritten this.

PR #2001, opened 2018-10-30
Allow OSG objects to be sent to the debug log.

AnyOldName3:
This seems like an obvious oversight that should have been included from the start.

@akortunov I'm not doing something silly with your feature by doing this, am I?


AnyOldName3:
I've not added a changelog entry as there's no change in behaviour, just less hassle for someone (probably me) who might need this in the future.

akortunov:
> I'm not doing something silly with your feature by doing this, am I?

Honestly, I do not know if it will work.

AnyOldName3:
It worked just fine when I tested it **without** recastnavigation. I guess maybe that's including the same header, but it doesn't have include guards.

akortunov:
Can the issue be somewhere [here](https://github.com/OpenMW/openmw/blob/master/components/detournavigator/flags.hpp#L18), for example?

akortunov:
It seems the is a conflict between our basic logging system and [this](https://github.com/OpenMW/openmw/blob/master/components/detournavigator/debug.hpp) one.

akortunov:
Suggested solution:
1. Wipe the components/osghelpers/operators.hpp out.
2. Use the osg/io_utils in the components/detournavigator/debug.hpp instead of operators.hpp.

AnyOldName3:
Yeah, `components/osghelpers/operators.hpp` should not exist. We need more code review before merging big things.

AnyOldName3:
I'm also unconvinced that `components/detournavigator/debug.xpp` should exist, at least in its current form. It both logs to files and to `std::cout` instead of to the main logging system.

akortunov:
I suppose we missed one overloaded "<<" operator in the testbulletnifloader.cpp, line 120.

AnyOldName3:
Weird that AppVeyor let it pass... It's not like I can build this myself anyway, though, as I'm still knocked out from the build system changes.

akortunov:
```
In file included from /home/travis/build/OpenMW/openmw/apps/openmw_test_suite/detournavigator/navigator.cpp:1:0:
/home/travis/build/OpenMW/openmw/apps/openmw_test_suite/detournavigator/operators.hpp:6:47: fatal error: components/osghelpers/operators.hpp: No such file or directory
 #include <components/osghelpers/operators.hpp>
```

AnyOldName3:
Not being able to build OpenMW really makes PRs a lot more error prone.

psi29a:
@AnyOldName3 thanks for for merging this.

PR #1633, opened 2018-03-15
Use recastnavigation for pathfinding (#2229)

elsid:
Solves [#2229](https://gitlab.com/OpenMW/openmw/issues/2229).

Should replace https://github.com/OpenMW/openmw/pull/1542. 

To build use recastnavigation from [repository](https://github.com/recastnavigation/recastnavigation).

Tasks and decisions to be done:
- [x] When build `NavMesh`? On cell preload and cell load in background.
- [x] Load `Navigator` configuration from settings.cfg.
- [x] Use tiles to build navmesh `. Allows actors to start find path faster. Possibly to do in parallel. Allows caching.
- [x] Visualize `NavMesh`. `togglenavmesh` command.
- [x] Visualize actors paths. `toggleactorspaths` command.
- [x] Split `RecastMesh` into tiles. Reduces time to build single navmesh tile.
- [x] Limit number of tiles to add to `NavMesh`. Important when number of loaded cells more than 9. There is a physical limit in recastnavigation how much polygons and tiles could be in navmesh. Polygon identifier contains tile number and polygon number inside tile 22 bits total. Now 10 bits used for tile and 12 for polygon. Could be increased if recastnavigation would use 64 bits identifiers (need a lot of changes).
- [x] Support changing collision shapes. Navmesh updates when object changed or transformed.
  - [x] Animated
  - [x] Transformed (doors)
- [x] Support avoided areas (task [#1724](https://gitlab.com/OpenMW/openmw/issues/1724))
- [x] Support swim (water). Navmesh surface with specific water flag. Possible to forbid actors to swim:
  - [x] Exterior cells
  - [x] Interior cells
- [ ] Support doors. Allow actors to find path through closed doors:
  - [x] without teleports
  - [ ] with teleports (who can use it?)
- [ ] Use different navmesh with half extents for different actors:
  - [ ] Exterior cells (takes too much time to build navmesh, need optimization)
  - [x] Interior cells
- [ ] Different settings for interior and exterior cells. There is a restrictions for that. Tile size depends on `cell size`. Navmesh can't contain tiles with different sizes, so different navmeshes are required to use different `cell size` for interior and exterior. Probably will not be possible to make off mesh connections for teleports (needs research).
- [ ] Documentation for settings
  - [x] Comment for each option in `settings-default.cfg` file
  - [ ] docs/source/reference/modding/settings/navigator.rst document
- [x] Cache `NavMesh` tiles data
  - [x] Select key generation approach:
[...]

PR #2003, opened 2018-10-31
Recastnavigation build fix

elsid:


AnyOldName3:
Before this is merged, bear in mind that this sets the recastnavigation include and library directories to whichever build type the Windows CI script is run with first (as that ends up in the CMake cache and then gets used whatever ends up being set later) for all build types. I'm not convinced that you'll be able to build a Debug version of OpenMW if you run the CI script in Release mode first and vice versa. More work is needed on the CMake side.

elsid:
Is it problem with recastnavigation? Build directory [depends](https://github.com/OpenMW/openmw/pull/2003/commits/a0ff4dc1c7362365624cdbf9bf958ee68f69aaa5#diff-5732a79d0837fe4a1ca67ab6e6f8e0a5R654) on build_type. If there is no recastnavigation, how does it work? Does openmw put all kind of artifacts into same directory with different names?

AnyOldName3:
Before this change, you were outputting whichever build type's libs into `/deps/recastnavigation/build/libs`. This is bad because on Windows, the default build system is Visual Studio, and that's what CMake calls a multi-config generator, so should support Debug, Release, RelWithDebInfo and MinSizeRel all with one build directory. Recastnavigation uses the same lib names for all build types (instead of using a suffix, like OSG or Qt), so these can't all go into the same directory. Also, on Windows, you can easily have VS 2013, 2015 and 2017 all with an OpenMW setup at the same time, so anything specific to just one of these should go in an appropriate directory.

After this change, while the source stays in `/deps/recastnavigation`, the selected build type's libs go into `/MSVC<year><_64 if it's 64-bit>/deps/recastnavigation/<build type>/libs`, so the libs for different build types, architectures and compiler versions don't conflict with each other, just like the rest of our dependencies.

However, when CMake is run, it's only being given one value for `RecastNavigation_ROOT` at a time (and because of how it uses its cache, it remembers this one forever even if it's called later with an alternative). This means that whatever build type you select in Visual Studio's drop-down, it'll always use the recastnavigation build that CMake was told about first, even if it's not one which will link correctly.

Something needs to be done so that CMake knows Release builds should be linked against the contents of `MSVC<year><_64?>/deps/recastnavigation/Release/libs`, Debug with `MSVC<year><_64?>/deps/recastnavigation/Debug/libs` etc.

When everything is sorted out, a Windows user should be able to run
```bash
CI/before_script.msvc.sh -k -v 2015 -c Release -p Win64
CI/before_script.msvc.sh -k -v 2015 -c Debug -p Win64
CI/before_script.msvc.sh -k -v 2015 -c RelWithDebInfo -p Win64

CI/before_script.msvc.sh -k -v 2017 -c Release -p Win64
CI/before_script.msvc.sh -k -v 2017 -c Debug -p Win64
CI/before_script.msvc.sh -k -v 2017 -c RelWithDebInfo -p Win64
```

This should give them an `MSVC2015_64` directory containing one solution file and an `MSVC2017_64` directory containing a different solution file. For both of these, selecting a build type from the drop-down at the top and building OpenMW should cause it to link against a recastnavigation build of the same type.

Right now, we have the first part, but not the last part.

psi29a:
We _really_ shouldn't be doing this. I highly recommend we remove this from our CI scripts and pull recastnavigation into /extern then fix up cmake from there.

AnyOldName3:
Whatever we do, we need to achieve the thing I'm about to repost. I don't particularly care how we do it (although I don't think we should have maintained dependencies in extern unless they're submodules) but it needs to get done. Putting it in `/extern` would achieve this, but would be the only library of its status in extern. @zinnschlag what's your opinion on submodule versus pulling for something that's still maintained, so will receive updates?

# The End Goal

**When everything is sorted out (be it via cloning, pulling to extern or a submodule), a Windows user should be able to run**

```bash
CI/before_script.msvc.sh -k -v 2015 -c Release -p Win64
CI/before_script.msvc.sh -k -v 2015 -c Debug -p Win64
CI/before_script.msvc.sh -k -v 2015 -c RelWithDebInfo -p Win64

CI/before_script.msvc.sh -k -v 2017 -c Release -p Win64
CI/before_script.msvc.sh -k -v 2017 -c Debug -p Win64
CI/before_script.msvc.sh -k -v 2017 -c RelWithDebInfo -p Win64
```

**This should give them an MSVC2015_64 directory containing one solution file and an MSVC2017_64 directory containing a different solution file. For both of these, selecting a build type from the drop-down at the top and building OpenMW should cause it to link against a recastnavigation build of the same type.**

AnyOldName3:
I specifically didn't make my changes into a pull request because I thought something more maintainable than cloning would be preferable, though. I just wanted to be able to build OpenMW (which I can now, but it doesn't run, instead crashing with an attempt to dereference `0x0` with the following stack trace:
image)

akortunov:
Looks like we need to wipe the custom logging system out anyway.

akortunov:
@AnyOldName3, can you try to make the isEnabled() return false if mSink is null and see what will happen?

elsid:
`mSink` can't be null. It is a wrapper over unique_ptr. This method is designed to check guarded value for null. I wonder what is the assembly code msvc generated. Probably it missed some default initializations somewhere else.

elsid:
To make build system work correctly on windows, such environment is required. I don't have windows development environment. If all the problem is with artifacts suffixes, so we need to change recastnavigation cmake files to make it as we need.

elsid:
After https://github.com/OpenMW/openmw/pull/1633#issuecomment-434972441 and https://github.com/OpenMW/openmw/pull/1633#issuecomment-434786268 I consider we want to have all recastnavigation code inside openmw, so here is pr https://github.com/OpenMW/openmw/pull/2005  to make it such way.

psi29a:
@AnyOldName3 agreed and I believe @elsid is doing the right thing here with #2005 
I talked with @zinnschlag yesterday:

> (09:49:31) Zini: I looked through the email notifications from GitHub a few minutes ago. Seems there are problems. Still not a major issue so early in the development cycle of 0.46.
> (09:49:58) psi29a: It is something that should have been dumped into /extern
> (09:50:05) psi29a: we shouldn't be pulling from git
> (09:50:16) psi29a: that is worse than submodules
> (09:50:49) Zini: Oh, good old submodules. How I miss you ... not
> (09:50:59) Zini: But yeah, I agree
> (09:51:05) psi29a: I know, but now we're
> (09:51:42) psi29a: pulling from a 3rd party's master... fun
> (09:52:15) Zini: Dwarf Fortress kind of "fun".
... (dwarf fortress tangent)
> (09:57:04) psi29a: anyway.. I'm looking into pull retour lib into /extern, but if someone beats me to it... that's cool
> (09:57:29) Zini: Not me, that much is sure ;)

psi29a:
Closing this.

PR #2004, opened 2018-11-01
Ensure different Boost versions get different names

AnyOldName3:
If, like me, someone has separate VS2017 and VS2015 OpenMW builds, the change to Boost 1.67 accidentally made them share the same Boost installer, which would break one build when the user eventually got around to updating it. This ensures different files get different names again.

It was silly of Microsoft to give VC 14.0 (which comes with VS 2015) and VC 14.1 (which comes with VS 2017) both the year 2015, though. They shouldn't have done that.


psi29a:
No mess, no revert. :)

PR #2005, opened 2018-11-01
Add recastnavigation as subdirectory

elsid:
No changes in recastnavigation cmake files are required thanks to CMake 3. So we can update it later without any merge conflicts.


elsid:
It is based on https://github.com/recastnavigation/recastnavigation master.

psi29a:
Yes... but master is just a branch, do you mean:

> Latest commit 7bfd9a1  13 hours ago
?

Can you add this somewhere so we can track it?

elsid:
Yes, this one https://github.com/recastnavigation/recastnavigation/commit/7bfd9a1d4caccf61e0485b2b05b29966348a8b39 . Is file with commit hash at external/recastnavigation/.id alright?

akortunov:
Probably we should also exclude RecastNavigation from CoverityScan, otherwise we may get many defects in the scan output.

akortunov:
> Is 3 solved with PR #2006?

I do not know. We need to test it.

CMakeLists.txt
```diff
@@ -536,6 +536,9 @@ if(WIN32)
 endif(WIN32)
 
 # Extern
+set(RECASTNAVIGATION_DEMO OFF CACHE BOOL "Do not build RecastDemo")
+
+add_subdirectory (extern/recastnavigation)
```
psi29a: Can you also set to building statically by default? We should be pulling this into component.

elsid:
> Probably we should also exclude RecastNavigation library itself from CoverityScan, otherwise we may get many defects in the scan output.

Should it be done by changing some script of CI config?

CMakeLists.txt
elsid: Fixed

Capostrophic:
Coverity configuration is in .travis.yml file

psi29a:
> Error: Tab characters found!

Probably a good idea to set your IDE or editor to use 4 spaces instead of tabs.

AnyOldName3:
Even better, we could make the tab checking script look for `.editorconfig` files, then the OpenMW code will be checked against our `.editorconfig` file and the recastnavigation stuff will be checked against their `.editorconfig` file.

psi29a:
The last commit is wrong, recastnavigation needs to go in /extern, not components.

psi29a:
What I meant by 

> 1) recastlib needs to be built statically and linked into components, we don't want to carry any additional shared libraries

that the resulting static libs are to be linked into the components, which is also a static library. Not move recastnavigation directory into components. Is that doable?

AnyOldName3:
This PR:
* Builds.
* Doesn't make a big mess.
* Lets all build types and generators live in harmony.
These are good.
This PR also:
* Still crashes (I've not checked akortunov's other PR yet).
* Puts the recastnavigation source into the components directory instead of extern, which I feel isn't ideal (I think Psi commented on this while I was writing this).

AnyOldName3:
Looking at the linker settings in VS, I'm pretty sure `components` and `openmw` are both linked against the recastnavigation libs.

psi29a:
So this PR is mergable once recast is moved back to /extern :)

psi29a:
Merged here: https://gitlab.com/OpenMW/openmw/merge_requests/45

It is the same except for the last commit.

psi29a:
Closing this PR, thanks @elsid for getting this done. :)

Now, to solve why OpenMW is crashing on Windows.

PR #2006, opened 2018-11-01
[Discussion] Unify logging system for RecastNavigation

akortunov:
An attempt to get rid of navigation-only logging system.
Notes:
1. I see no point to use the second logging system.
If our main one does not support something, we can enhance it instead of doubling.
2. For now it will log any recast messages, so we will need to discuss how and to where we should log such messages.


elsid:
I’m ok with idea of this change, but I think we need support multiple log channels first, or adapt some existing solution, like I said here https://github.com/OpenMW/openmw/pull/1633#issuecomment-434809951 . Otherwise current log will be a mess for anyone who don’t develop detournavigator component.

AnyOldName3:
Are verbose logging messages enabled or disabled by default on release builds?

Another option is to have a Log level variable in the navigation system somewhere which is configured at load time based on the setting the old recast log used, then is used as the argument to all `Log(level)` calls. if we don't have such a thing already, adding `None` as a log level to discard things would be needed.

AnyOldName3:
image

This fixed *some* of the crash, but the same code is called elsewhere.

psi29a:
@elsid we really need to get to the bottom of this.

psi29a:
*sigh* Not only that... while OpenMW runs, I just tried a 'new' game... and the guard never comes down into the boat's hold and I can't move around. This is a problem.

elsid:
I've tried both appveyor artifacts for c1a3eec456fbca45c93ce2f7d726b30645922a16 . [MSVC2015](https://ci.appveyor.com/project/psi29a/openmw/builds/19971677/job/16n1hheb16572s6a/artifacts) result crashed, but [MSVC2017](https://ci.appveyor.com/project/psi29a/openmw/builds/19971677/job/3vqef6qblibfaqss/artifacts) is not. I've met problems before with default constructed types. And there is same [issue](https://stackoverflow.com/questions/32784399/possible-bug-in-handling-the-default-keyword-in-vs2015-c) on stackoverflow. There was even some kind of [ticket](https://connect.microsoft.com/VisualStudio/feedback/details/1849804) at MS connect, but they shutdown it (google cache and webarchive doesn't have it). I will try to make explicit call for all default constructors.

AnyOldName3:
The issue reporting system in VS 2015 tells you that you can only report issues for VS 2017, so to upgrade to that, check if the problem persists, and only then create an issue. I guess they have no intention of fixing it even if we do start complaining.

elsid:
So, this [issue](https://stackoverflow.com/questions/32784399/possible-bug-in-handling-the-default-keyword-in-vs2015-c) is applicable in [this](https://github.com/OpenMW/openmw/pull/2006#issuecomment-435091263) case. Pr https://github.com/OpenMW/openmw/pull/2011 fixes msvc2015 crashes, tested using appeveyor [artifact](https://ci.appveyor.com/project/elsid/openmw/builds/19995791/job/a0ith8ewq65lt41v/artifacts). Thanks microsoft for buggy compiler.

terabyte25:
Off topic; but have the recent changes been tested on macOS?

akortunov:
I suppose we need to discuss how we should log navigation-related stuff.

Personally, I am not in favour of separate logging system just for navigation. I'd prefer to enhance our current one (probably by implementing message categories).
Also I do not like an idea to use an external logging library - we do not need most of functionality which they provide.
An another issue: this change sooner or later will be in the TES3MP, and devs will need to adopt a new logging system too.

Basically, typical output looks like this:
```
process job for agent=29.279994964599609375 28.4799976348876953125 66.5
update NavMesh with mutiple tiles: agentHeight=2.3470590114593505859375 agentMaxClimb=0.60000002384185791015625 agentRadius=0.516705811023712158203125 changedTile=209 49 playerTile=208 49 changedTileDistance=1
cache updated for agent=29.279994964599609375 28.4799976348876953125 66.5 status=replaced generation=1 revision=40 time=0.077560000121593475341796875ms total_time=254.40301513671875ms
got 207 jobs
```

I suppose readability of such messages can be improved a bit by limiting output precision.

psi29a:
Ah, good and should probably be left at "Verbose" until 1.0 right?
I agree that we shouldn't have to be that precise. Do you think that two digits after the period is enough?

akortunov:
> Do you think that two digits after the period is enough?

Yes, I think two or three digits should be enough.

wareya:
>Actually, our default level is "Verbose", i.e. we display all debug messages by default.

>Ah, good and should probably be left at "Verbose" until 1.0 right?

Maybe there should be one more log level past "verbose", like "internal". That is, for people actively working on some part of OpenMW only, not for people testing it.

akortunov:
> Maybe there should be one more log level past "verbose", like "internal".

"Tracing"?

wareya:
Yeah that's better.

psi29a:
Tracing is good. :)

bozeyman9000:
Has anyone enabled a toggle for sneak?

jvoisin:
"Debug" instead of "tracing"?

psi29a:
@jvoisin One more level beyond verbose (aka Debug) will be Tracing.  
I understand, I'm used to: Info, Debug, Trace

AnyOldName3:
In a logging system with both debug and verbose levels, I'd expect debug to be more verbose than verbose is, and I think debug is closer to what it'll be used for than tracing is unless we expect to print something with every function call and branch. I like debug as a level name more than tracing. To me, verbose could be something a regular user might want, especially if they're doing something like making mods, but debug really implies it's only useful for hunting bugs and other developer-focused tasks.

akortunov:
An example ofmore readable output:
```
Process job for agent=(29.28 28.48 66.50)
Update NavMesh with multiple tiles: agentHeight=2.35 agentMaxClimb=0.60 agentRadius=0.52 changedTile=(202 55) playerTile=(208 49) changedTileDistance=8.49
Cache updated for agent=(29.28 28.48 66.50) status=add generation=1 revision=472 time=9.51ms total_time=1945.92ms
Got 61 navigator jobs
```

A bit strange stat in one placte there is a `std::setprecision(std::numeric_limits<float>::max_exponent10)`, but other places use a default precision.

psi29a:
So the types of logging could be:
Normal, Verbose, Debug and Tracing

With the first two something for the end-user?

psi29a:
What's the plan?

akortunov:
> What's the plan?

Well, first of all we need to decide which messages to keep and where to log them. Probably we need a feedback from @elsid here.
Second, we should decide if there is a need for external logging library, or our current logging implementation covers most of our needs.

akortunov:
Ok, [here](https://github.com/akortunov/openmw/tree/spdlog) is the first draft to use the spdlog. It seems it works, even with separate log file for Recast navigation.

My main complaints:
1. It seems the spdlog does not support the << operator, so I have so build output string and pass it to logger, even if this string will be filtered.
2. spdlog does not have a reliable way to override output colors, so I have to use a couple of hacks. Not sure if they will work on Windows.

Any ideas?

psi29a:
Pulling in @davidcernat for his thoughts, if we're going to pull in a 3rd party lib for logging, this might be interesting/useful to tes3mp.

davidcernat:
I'm going to ask @Koncord to comment here instead. He wrote both versions of TES3MP's logging system.

Koncord:
For first: why spdlog? If we are want to deboosting code and Boost.Log is not applicable. The best solution is use homebrew logger.

IMO, ``Trace() << "Message"`` better than ``Log(Trace) << "Message"``

akortunov:
So what we decided to do? 
Just add an another log level (Debug::Debug or Debug::Tracing), but keep the Debug::Verbose as a default log level?

Koncord:
Debug::Tracing should be displayed in Release and Debug, Debug::Debug only in Debug

elsid:
> why spdlog?

It has quite good performance and supports our cases. If there is any other same quality solution I would be glad to know.

> The best solution is use homebrew logger.

Why is it so? Any reason to write one more solution to write text to file?

akortunov:
And we still did not decide what to do:
1. Use an external logging library.
2. Extend our homebrew logging system to support different sinks and wipe out the navigator-specific logging system.
3. Add a new log level to our homebrew logging system and wipe out the navigator-specific logging system.
4. Keep two logging systems with different syntax.

Capostrophic:
I like 3 (KISS principle) but it doesn't seem like everyone shares such opinion.

AnyOldName3:
I *think* I was pro-3, but can't remember as it was ages ago.

psi29a:
Same with 3. Let's get everything under one logging system, then later deal with the best way to log.

Capostrophic:
>Is there any additional work for 3?

Yes. While recastnavigation-specific logging system was purged the additional debug level wasn't added and all messages pathfinding logs are currently under Verbose, and the debug level should be configurable in some way for elsid's (or any other person debugging recastnavigation) convenience.

Capostrophic:
It would make sense to get this over with sooner than later.

akortunov:
@elsid, do you still need a separate logging system, or just additional log level would be enough?

elsid:
Additional level is fine, but disabled by default and with option in settings to enable it.

psi29a:
One less logging system please. ;)

psi29a:
Sweet...

psi29a:
Ready or more discussion needed?

apps/openmw/mwmechanics/pathfinding.cpp
```diff
@@ -327,7 +327,6 @@ namespace MWMechanics
         }
         catch (const DetourNavigator::NavigatorException& exception)
         {
-            DetourNavigator::log("PathFinder::buildPathByNavigator navigator exception: ", exception.what());
             Log(Debug::Verbose) << "Build path by navigator exception: \"" << exception.what()
```
Capostrophic: Debug level should be updated here.

components/debug/debugging.hpp
```diff
@@ -82,7 +82,8 @@ namespace Debug
             mColors[Error] = Red;
             mColors[Warning] = Yellow;
             mColors[Info] = Reset;
-            mColors[Verbose] = DarkGray;
+            mColors[Verbose] = Reset;
```
Capostrophic: Are you sure the color should be changed for Verbose level?

akortunov: Why it should not?

Capostrophic: Why should it?

It's been dark gray forever -- well, since the time you implemented the new logging system -- and that has been just fine for all the messages it was used for. At least I myself got used to it.

akortunov: Can you suggest an another color for Debug level then?

psi29a:
Ready?

akortunov:
> Ready?

There is still one unresolved discussion about text colouring in console.

components/debug/debugging.hpp
Capostrophic: IMO dark gray for both levels would have been just fine.

akortunov: Done.

akortunov:
Should be ready for merging now, if there are no other objections.

Capostrophic:
There were practically no objections for literal months, so I believe that this is fine to be merged as it is now.

psi29a:
Thanks!

PR #2008, opened 2018-11-01
Fix isWithinMaxRange function in AiTravel [0.45.0 regression]

Capostrophic:
This fixes the major problem brought in (surprisingly) by actor processing optimization and not recastnavigation introduction: AiTravel is completely broken.

isWithinMaxRange was using incorrect type for AI processing range so it almost always returned false, breaking AiTravel. JDGBOLT discovered this.

This should be cherry-picked to 0.45.0 branch. I wanted to make some cleanup but it was too specific to recastnavigation pathfinding.


psi29a:
So this is a regression for 0.45, not caused by recastnavigation. Glad we got that cleared up. :)

psi29a:
cherry-picked to 0.45 branch

PR #2009, opened 2018-11-01
Use default tolerance to drop last point from path

elsid:
Fixes AiEscort. Before actor tried to reach target with 0 tolerance.
Back to logic like it was before 4fe764c3a53 and bbd82a743 commits.


JDGBOLT:
I can confirm that this gets rid of the circling problem that the guard escorting you to the exit of the ship when starting a new game, he correctly goes to where he is supposed to and tells you to get up to the deck.

psi29a:
What does this fix?

The regression for the guard not coming to unglue you from the boat's hold was apparently a regression  caused by "actor processing optimization" that was fixed here: #2008

JDGBOLT:
Yes, there was a second problem after that was fixed, where the guard would reach where he is supposed to go to, but then starts circling around that point and not behaving correctly.

PR #2010, opened 2018-11-02
Include missing Windows header before calling Windows-specific functions

AnyOldName3:
https://github.com/OpenMW/openmw/commit/d19839a66670f4410ebf97e7dcd42f1c89300114 bumped the minimum CMake version on Windows, but Windows relied on the old behaviour of CMP0043, which this switched to new. This PR switches the broken calls to use their new forms.

https://github.com/OpenMW/openmw/blob/master/components/debug/debugging.hpp#L55 calls a Windows function, but at some point, an indirect include must have changed, as it's no longer there. This restores the include, making the function available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -614,7 +614,6 @@ if (WIN32)
     if (BUILD_OPENMW)
         # Release builds don't use the debug console
         set_target_properties(openmw PROPERTIES LINK_FLAGS_RELEASE "/SUBSYSTEM:WINDOWS")
-        set_target_properties(openmw PROPERTIES COMPILE_DEFINITIONS $<$<CONFIG:Release>:_WINDOWS>)
         set_target_properties(openmw PROPERTIES LINK_FLAGS_MINSIZEREL "/SUBSYSTEM:WINDOWS")
     endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,7 +276,7 @@ if (APPLE)
 endif (APPLE)
 
 # Set up DEBUG define
-set_directory_properties(PROPERTIES COMPILE_DEFINITIONS_DEBUG DEBUG=1)
+set_directory_properties(PROPERTIES COMPILE_DEFINITIONS $<$<CONFIG:Debug>:DEBUG=1>)
 
 if (NOT APPLE)
     set(OPENMW_MYGUI_FILES_ROOT ${OpenMW_BINARY_DIR})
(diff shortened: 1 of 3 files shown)


akortunov:
We already have such code in the main.cpp:
```
#if defined(_WIN32)
// For OutputDebugString
#ifndef WIN32_LEAN_AND_MEAN
#define WIN32_LEAN_AND_MEAN
#endif
#include <windows.h>
// makes __argc and __argv available on windows
#include <cstdlib>
#endif
```

Can it conflict with your code?

akortunov:
> lolwat?

Which tinyXML CI uses? A bundled one or a system-wide one?

akortunov:
https://forums.gentoo.org/viewtopic-t-817732-start-0.html

https://sourceforge.net/p/tinyxml/discussion/42748/thread/2ae82eab/

wareya:
Dependency hell, woo!

psi29a:
Nothing has changed with Trusty (really old at this point).

psi29a:
AV has other errors now too... so it's all kinds of fun with this PR.

AnyOldName3:
Good thing I didn't merge it.

AnyOldName3:
I know why AppVeyor fails (although don't know why it didn't fail when I built it on my machine). One of the windows headers does
```C++
#ifndef min
#define min(a,b)            (((a) < (b)) ? (a) : (b))
#endif
```
so when we do
```c++
size_t toRead = std::min((mOrigin+mSize)-(mFile.tell()), sBufferSize);
```
the `min` gets replaced by the Windows version, and then the expression makes no sense. This is a stupid thing for the Windows header to do.

AnyOldName3:
> We already have such code in the main.cpp:
> 
> ```
> #if defined(_WIN32)
> // For OutputDebugString
> #ifndef WIN32_LEAN_AND_MEAN
> #define WIN32_LEAN_AND_MEAN
> #endif
> #include <windows.h>
> // makes __argc and __argv available on windows
> #include <cstdlib>
> #endif
> ```
> Can it conflict with your code?

It shouldn't conflict, as (if it's not completely stupid) `Window.h` *should* have include guards.

If we're only using it for `OutputDebugString`, though, it's a bit silly that it's lower in the file than the first use of `OutputDebugString`

CMakeLists.txt
```diff
@@ -276,7 +276,7 @@ if (APPLE)
 endif (APPLE)
 
 # Set up DEBUG define
-set_directory_properties(PROPERTIES COMPILE_DEFINITIONS_DEBUG DEBUG=1)
+set_directory_properties(PROPERTIES COMPILE_DEFINITIONS $<$<CONFIG:Debug>:DEBUG=1>)
```
AnyOldName3: This line used to do nothing. My 'fix' enabled it, and that's what's killing stuff.

AnyOldName3: This line used to do nothing. My 'fix' enabled it again, but it seems we want it dead.

AnyOldName3:
For some reason, adding the `_WINDOWS` compile definition to Release builds was removing the Debug-specific ones from Debug builds. I think it has something to do with `_WINDOWS` coming from somewhere else, too. I guess it's a CMake bug.

AnyOldName3:
I'm declaring victory now it's working for me and CI.

PR #2011, opened 2018-11-02
Use explicit default ctors call

elsid:
Special change for MSVC2015. Prevents crashes.


Capostrophic:
AppVeyor MSVC 2015 build finished fine but MSVC 2017 build is apparently still running, having taken 3 hours to get to the current point?!

PR #2013, opened 2018-11-02
[WIP] Try AV with Ubuntus

psi29a:
Just checking what we can do with AV.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -52,16 +52,19 @@ cache:
 clone_folder: C:\projects\openmw
 
 install:
-    - set PATH=C:\Program Files\Git\mingw64\bin;%PATH%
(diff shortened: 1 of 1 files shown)


akortunov:
Is there a reason to do it? IIRC, AppVeyor executes jobs one-by-one, so 4 tasks will spend near 2 hours of time.

psi29a:
A few reasons:
1) to see if it possible
2) we get xenial and bionic
3) we get artifacts and hosting for them
4) we can set it up for nightlies only

akortunov:
Any progress here?

psi29a:
Still WIP but sadly AV is still serial and not parallel. Which is all kinds of sad. I'm going to close this and if I need the code I can always comes back to it.

PR #2014, opened 2018-11-03
Improve toggleactorspaths console command

akortunov:
Summary of changes:
1. Do not render paths for dead actors.
2. Do not render path, if an actor (a wandering one, for example) reached the destination, but with destination tolerance.
3. Add a "tap" shortcut for this command.


psi29a:
Nice!

PR #2018, opened 2018-11-03
fix resolution on HiDPI displays under Windows

sthalik:
This particular Windows manifest bit fixes resolution choice on my Windows 2160p display. The operating system likes to confuse itself unless the particular piece of boilerplate is included in the executable.

Otherwise it's necessary to force "application scaling" in the executable's compatibility options.


Capostrophic:
Is this related to [bug 3623](https://gitlab.com/OpenMW/openmw/issues/3623) somehow?

sthalik:
Yes, at least in windowed mode.

psi29a:
Is there anyone on Window that can validate this?

lysol90:
I only have a 1440p monitor, and it is on my linux rig, so sorry no. But I think @AnyOldName3 had some nice phat resolution monitor..?

AnyOldName3:
I've got a 1440p monitor, What do I need to do?

AnyOldName3:
I have resolution scaling set to 100%, though, as it's not a small monitor.

akortunov:
> I've got a 1440p monitor, What do I need to do?

I suppose you should try to increase scaling factor in Windows settings (set it to something like 120% or 150%).

sthalik:
The following were broken for me:

- windowed native resolution without frames (was too big)
- fullscreen native resolution (display's OSD shows lower resolution)

psi29a:
Anyone else on windows to test this?

AnyOldName3:
I've tested this and it does indeed resolve the issues it's supposed to.

PR #2020, opened 2018-11-04
fix rain delay when exiting water

sthalik:
This fixes [bug 4540](https://gitlab.com/OpenMW/openmw/issues/4540). I've documented own failed attempts in the commit message.

Rain bubbles touching the surface still work, rain doesn't display through the water, and it shows immediately upon exiting the water. I don't know about switching cells, the issue is likely to remain.

Note, I've had to disable culling due to how the osg particle system works. The rationale is that there's barely any detail underwater, and the lack of culling won't degrade performance in any significant manner.

For fixing cell switch I'd go for a hack like increasing particle count *and* velocity by some large factor so that there's already plenty of rain in a fraction of a second. Either that or introducing an artificial delay for rain to show.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.46.0
 ------
 
+    Bug #4540: Rain delay when exiting water
     Feature #2229: Improve pathfinding AI
     Feature #3442: Default values for fallbacks from ini file
     Task #4686: Upgrade media decoder to a more current FFmpeg API
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -401,11 +401,15 @@ public:
     {
     }
 
-    virtual void operator()(osg::Node* node, osg::NodeVisitor* nv)
+    bool isUnderwater()
     {
         osg::Vec3f eyePoint = mCameraRelativeTransform->getLastEyePoint();
+        return mEnabled && eyePoint.z() < mWaterLevel;
+    }
 
-        if (mEnabled && eyePoint.z() < mWaterLevel)
+    virtual void operator()(osg::Node* node, osg::NodeVisitor* nv)
+    {
+        if (isUnderwater())
             return;
 
         traverse(node, nv);
@@ -1575,6 +1579,8 @@ void SkyManager::update(float duration)
             mRainIntensityUniform->set((float) mWeatherAlpha);
     }
 
+    switchUnderwaterRain();
+
     if (mIsStorm)
     {
         osg::Quat quat;
@@ -1624,6 +1630,15 @@ void SkyManager::updateRainParameters()
         mRainShooter->setVelocity(osg::Vec3f(0, mRainSpeed * windFactor, -mRainSpeed));
         mRainShooter->setAngle(angle);
     }
+}
+
+void SkyManager::switchUnderwaterRain()
+{
+    if (!mRainParticleSystem)
+        return;
+
+    bool freeze = mUnderwaterSwitch->isUnderwater();
+    mRainParticleSystem->setFrozen(freeze);
 }
 
 void SkyManager::setWeather(const WeatherResult& weather)
--- a/apps/openmw/mwrender/sky.hpp
+++ b/apps/openmw/mwrender/sky.hpp
@@ -180,6 +180,7 @@ namespace MWRender
 
         void createRain();
         void destroyRain();
+        void switchUnderwaterRain();
         void updateRainParameters();
 
         Resource::SceneManager* mSceneManager;


Capostrophic:
Changes look promising, I'll take a look later today.

lysol90:
Just wanted to chip in and say "welcome". Nice to see a new developer here. You won't see any code reviews from me, but you might see my useless fanboy comments from time to time...

sthalik:
v2: update changelog and rebase on master only.

AnyOldName3:
Wouldn't it have been enough to just put the `setFrozen` call in the existing cull callback rather than removing it completely? It only seems to be that call and swapping the nodemask (thereby disabling the update callback, too) that was actually changed here, but lots of other stuff got touched.

sthalik:
No, `setFrozen` by itself causes the rain to stand still underwater.

AnyOldName3:
But the `return` statement here https://github.com/OpenMW/openmw/pull/2020/files#diff-49768b509d6eb7ab9109d043edae5a9fL409 prevents the cull traversal from happening, so it won't get drawn, either.

sthalik:
The culling needs to run on `mRainParticleSystem` instead. Node mask needs to be set to zero, will the cull callback run again then? Other than that, the callback looks like more boilerplate and overhead. YMMV.

sthalik:
@AnyOldName3 

With regard to diff size, the callback class looks boilerplate-y C++98. YMMV. I elected to remove it entirely given it can't support the functionality by itself (nodemask inhibits culling). Having it present requires two places to check for the underwater rain transition: the callback and the update method to disable/enable rain, respectively. The functionality isn't contained, rather it smells of "action at distance".

FWIW, it would be interesting to make a node callback wrapper taking C++11 lambdas. Then the logic becomes more centralized while preserving osg's way of doing things.

AnyOldName3:
To clarify, I'm not saying that your solution isn't necessarily the best one, just that as you've not had anything merged yet and OSG can sometimes expect you to know how to do things without it being mentioned anywhere in their documentation, source code, mailing list or examples, I want to make sure you've properly considered all options rather than throwing random changes at the problem until it went away. Basically I want to check you know that:

* The callback's `operator()` is called every time an attached node is visited by the cull traversal.
* Returning from that call without traversing is the same as skipping the node during the cull traversal.
* Disabling the traversal via the nodemask also skips the other traversals (update/intersection/any custom traversal we do or might do in the future) and so might have far-reaching effects that might not even appear for a decade.
* The recommended way to do many things in OSG is with a callback, even ones that it isn't obviously a good idea to (although they're not a universal solution and can be really silly at times).
* *The eye point given by the `CameraRelativeTransform` comes from the last cull traversal, so is probably out of date when the sky manager's `update(float dt)` method is called, and you should probably only be using it during a cull traversal when it's still fresh, like the custom callback used to.*
* It might be safe to pause the particle system during a cull traversal so it doesn't get updated during the next update traversal, but it might not be.
* It might even be the case that there's no way to make everything safe and use the current frame's camera position without somehow giving the `update` method access to that information.

If you're already aware of all this, have properly considered all options, and are convinced that this PR is the best solution (and doesn't suffer from any of the out-of-date-eye-point issues that I'm worried about), it's probably ready to be merged.

sthalik:
> I want to make sure you've properly considered all options rather than throwing random changes at the problem until it went away.

Actually that's what I did (bang on the keyboard until it worked), in fact. My knowledge of OSG is minimal, and that was the first task I threw myself at.

Can you please point out whether there's an inherent performance disadvantage from avoiding the callback, as per the commit? If there's an issue like having to rebuild the scene graph due to direct manipulation that's definitely to be avoided.

While the issue you present is definitely valid, there are more basic problems at hand. I think having it done one or the other way pales in comparison to `SkyManager` having poor architectural design (a "god class", individual components not factored out for composition). Simple "bool" or "float" types not initialized inline C++11-style, and so on. I think in light of that we should think of practical concerns here (performance, correctness, stability) rather than getting it "right" for a component that's in bad shape as it is.

AnyOldName3:
>I think in light of that we should think of practical concerns here (performance, correctness, stability) rather than getting it "right" for a component that's in bad shape as it is.

Now I've had a close enough look at this, I'm pretty sure that you got rid of correctness by moving the camera position check to the update function rather than during the cull traversal. You should take a close look (potentially using something like APITrace so it doesn't have to be real-time) at the frame immediately after crossing the water plane when moving quickly upwards and downwards. If in both cases, you manage to avoid there being a frame with the wrong rain state, this PR is probably fine. If there are incorrectly rainy frames, I think it's probably an inherent flaw with doing it during the update function and this is a case where we *need* to use a cull callback.

sthalik:
@AnyOldName3, the code indeed had an off-by-one-frame error as per apitrace. I changed the approach and now transitioning both in and out of water has no errors.

It's possible that transitioning out of water state has a frame with the particle system frozen, given the relative eye position is one frame late. I haven't been able to see it on from my frame-by-frame dumps.

The droplets no longer show for one frame under water's surface, and there's no need to set the node mask anymore.

AnyOldName3:
> I changed the approach and now transitioning both in and out of water has no errors.

Great 😄 

> It's possible that transitioning out of water state has a frame with the particle system frozen, given the relative eye position is one frame late. I haven't been able to see it on from my frame-by-frame dumps.

If you can't see it when looking that closely, it's probably not an issue that people will see in-game if they're playing with a low framerate. If you want to be really pedantic, you could check by (doing something along the lines of) starting with the player underwater, then triggering a script that immediately moves the player to a fixed position above the water and keeps them there. As the player moves then stays still, you'll have a fixed camera position when comparing two rain positions.

If you don't want to be pedantic, this is probably okay to be merged.

sthalik:
@AnyOldName3 this can't be done at this layer. One "Z-above-water" value will be kept out of date. I tried to freeze inside the callback and it doesn't want to work.

psi29a:
> this is probably okay to be merged.

Right! Congrats @sthalik on your first merged PR. :)

PR #2021, opened 2018-11-04
Recast global allocator

elsid:
To fix https://github.com/OpenMW/openmw/pull/1633#issuecomment-435601824 . Recastnavigation uses a lot of small vectors to perform internal calculations like [this](https://github.com/OpenMW/openmw/blob/d3633bbada6ec44c9e6dc3e04404a1438ce71afa/extern/recastnavigation/Recast/Source/RecastContour.cpp#L904) one. It has allocator function customization which supports different allocation policies: permanent and temporary. By default both of them use `malloc`. `rcIntArray` allocates using temporary policy. To reduce `malloc` calls all temporary allocations should use preallocated buffer. This pr set custom allocator for recastnavigation with stack allocator for temporary allocations.

Here is memory profiles to compare:

[openmw-45](https://github.com/OpenMW/openmw/commit/de381d0d49c702b8e52a5208cb783144aec53216) branch:
openmw-45 summary
openmw-45 allocated

[master](https://github.com/OpenMW/openmw/commit/d3633bbada6ec44c9e6dc3e04404a1438ce71afa) branch:
master summary
master allocated

[this](https://github.com/OpenMW/openmw/commit/b477775e1602893c6586b0525ed104b1b383636e) pr:
with_allocator summary
with_allocator allocated

### DIFF
--- a/components/detournavigator/recastallocutils.hpp
+++ b/components/detournavigator/recastallocutils.hpp
@@ -1,0 +1,90 @@
+#ifndef OPENMW_COMPONENTS_DETOURNAVIGATOR_RECASTALLOCUTILS_H
+#define OPENMW_COMPONENTS_DETOURNAVIGATOR_RECASTALLOCUTILS_H
+
+#include <RecastAlloc.h>
+
+#include <cstdint>
+
+namespace DetourNavigator
+{
+    static_assert(sizeof(std::size_t) == sizeof(void*), "");
+
+    enum BufferType : std::size_t
+    {
+        BufferType_perm,
+        BufferType_temp,
+        BufferType_unused,
+    };
+
+    inline BufferType* tempPtrBufferType(void* ptr)
+    {
+        return reinterpret_cast<BufferType*>(static_cast<std::size_t*>(ptr) + 1);
+    }
+
+    inline BufferType getTempPtrBufferType(void* ptr)
+    {
+        return *tempPtrBufferType(ptr);
+    }
+
+    inline void setTempPtrBufferType(void* ptr, BufferType value)
+    {
+        *tempPtrBufferType(ptr) = value;
+    }
+
+    inline void** tempPtrPrev(void* ptr)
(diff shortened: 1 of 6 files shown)


psi29a:
Very nice profiling and comparisons. There will still be a memory bump to OpenMW but at least this can be more easily swallowed. :)

components/detournavigator/recastallocutils.hpp
```diff
@@ -0,0 +1,102 @@
...
+    inline void* getPermPtrDataPtr(void* ptr)
+    {
+        return static_cast<std::size_t*>(ptr) + 1;
+    }
+
+    // TODO: use std::align
+    inline void* align(std::size_t align, std::size_t size, void*& ptr, std::size_t& space) noexcept
+    {
+        const auto intptr = reinterpret_cast<std::uintptr_t>(ptr);
+        const auto aligned = (intptr - 1u + align) & - align;
```
akortunov: There is a something strange here.
Since "align" is unsigned, MSVC complains about unary minus operator which has no effect.
I wonder what should be here.

elsid:
I’ve just taken this function from libstdc++. Probably, there is better implementation somewhere else. We can even use standard one for windows.

psi29a:
> I’ve just taken this function from libstdc++. Probably, there is better implementation somewhere else. We can even use standard one for windows.

You  mean std::align? It's C++11, can't we just use that?

psi29a:
Just tested, std::align works just fine.

components/detournavigator/recastallocutils.hpp
psi29a: Try this one?  #2026

elsid:
> You mean std::align? It's C++11, can't we just use that?

We can't with gcc 4.8 . This comment is about future when we will use newer gcc.

psi29a:
I know, I'm working on that.

PR #2022, opened 2018-11-04
[WIP] prevent operating on invalidated iterators

sthalik:
Prevent incorrect usage of STL containers that can lead to use-after-free, crashes, etc. I haven't tested all of the code paths since there's so many of them in different places.

### DIFF
--- a/apps/opencs/model/prefs/shortcutmanager.cpp
+++ b/apps/opencs/model/prefs/shortcutmanager.cpp
@@ -31,7 +31,7 @@ namespace CSMPrefs
         {
             if (it->second == shortcut)
             {
-                mShortcuts.erase(it++);
+                it = mShortcuts.erase(it);
             }
             else
             {
--- a/apps/openmw/mwmechanics/activespells.cpp
+++ b/apps/openmw/mwmechanics/activespells.cpp
@@ -246,7 +246,7 @@ namespace MWMechanics
             }
 
             if (Misc::Rng::roll0to99() < chance)
-                mSpells.erase(it++);
+                it = mSpells.erase(it);
             else
                 ++it;
         }
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -1240,7 +1240,7 @@ namespace MWMechanics
             if((iter->first.isInCell() && iter->first.getCell()==cellStore) && iter->first != ignore)
             {
                 delete iter->second;
-                mActors.erase(iter++);
+                iter = mActors.erase(iter);
             }
             else
                 ++iter;
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -1109,7 +1109,7 @@ namespace MWMechanics
                 itemCount -= toRemove;
                 ownerIt->second -= toRemove;
                 if (ownerIt->second == 0)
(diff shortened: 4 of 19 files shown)


psi29a:
No changelog needed in this case as it doesn't fix a specific bug and is more refactor/code-clean-up.

Thanks, I'll merge this in. :)

sthalik:
@zinnschlag, you're right about `std::map`. I'm amending the pull request.

@psi29a almost all changes are untested. Something may break for a subtle reason.

sthalik:
After reading [the rules](https://stackoverflow.com/questions/6438086/iterator-invalidation-rules/6438087#6438087), nothing actually remains to be fixed for all STL iterator remove-while-iterating code.

@zinnschlag thanks for pointing it out.

kcat:
Thinking about it, I do wonder if things like `mShortcuts.erase(it++);` on maps are valid. Since this is a post-increment, the function call gets the old value as the iterator increments. Is there any guarantee the increment happens before the function call? IIRC, the increment is only guaranteed to occur by the time of reaching the `;`. C++ can be pretty tricky with when side-effects of that nature become visible (and the rules have slightly changed over the years) and who knows what the optimizer is going to do.

sthalik:
> Is there any guarantee the increment happens before the function call?

It's generally assumed that function arguments are evaluated before the functional expression itself.

kcat:
> It's generally assumed

You know what they say about assuming.

When it comes to compilers, they can get pretty aggressive with optimizations. Its task is to make code more efficient while staying within the behavior required by the spec. Reordering operations, as long as all visible side-effects are viewed as the spec requires, is rather common.

For example, prior to C++17, `x = x++;` is undefined. We would assume the right side is completed before the assignment to the left, but that's not quite how it works. The assignment on the left only needs the result of the operation on the right, which is known to be unmodified x, but the side effects of assigning x and incrementing x are unordered.
https://youtu.be/XEXpwis_deQ?t=1636 (see Example 9)

To bring it back, the call `mShortcuts.erase(it++);` would be similar. The assignment to the erase method's parameter only needs the unmodified `it`. And because `it` is a local variable, the compiler knows the function call can't modify it. Therefore, to initialize the parameter it only needs unmodified `it`, and the side effect of incrementing `it` only needs to be visible before its next read, which would be some time after the call. It is of course valid behavior to increment before the call, but it doesn't necessarily have to be done that way. Especially given that `erase()` would probably be (partly) inlined, the compiler may reorder various sub-operations in a way that it doesn't know is producing undesirable outer-operation order.

elsid:
When it comes to definition of some behavior. Read the rules from the standard, of it possible. Otherwise look at cppreference at least. https://en.cppreference.com/w/cpp/language/eval_order .
“3) When calling a function (whether or not the function is inline, and whether or not explicit function call syntax is used), every value computation and side effect associated with any argument expression, or with the postfix expression designating the called function, is sequenced before execution of every expression or statement in the body of the called function.”
.erase is a function call, so the rule is applicable here. Current code is correct.

Mingun:
Unless in this case `it++` is calculation of value of an argument? Value of an argument is `it`, and call order of `operator++` or `erase` isn't defined.

Mingun:
Anyway, irrespective of, I am right or you if the current code raises questions, then it is better to rewrite it

sthalik:
@Mingun, elsid's [comment](https://github.com/OpenMW/openmw/pull/2022#issuecomment-436194098) disproves that.

PR #2024, opened 2018-11-05
fix msvc warnings with recast

psi29a:
bump /extern/recastnavigation to commit 3087e805b02d5eb8fff7851234fa2b3f71290eba; fixes clang warning about missing delete: https://github.com/recastnavigation/recastnavigation/issues/359

should just be a no-op (good coding syle)


psi29a:
Reviewed by @Capostrophic

PR #2025, opened 2018-11-06
Use C++11 for close actor-iterating loops and avoid some redundant operations

Capostrophic:
This mostly uses range-based for loops from C++11 for syntactic sugar of actor-iterating loops in the mechanics, but other than that also reorders some conditions and avoids some redundant loops, calculations and declarations that could be done and result in no actual effect on the game other than worse performance with greater number of actors in some conditions.

However, I doubt this would bring any noticeable performance improvement. The point is to improve the readability and maintanability of the code (remember that one regression when `it` was confused with `iter` in the same condition?).

Unfortunately std::any_of couldn't really be used anywhere.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -491,8 +491,8 @@ namespace MWMechanics
         // If any of the above conditions turned actor1 aggressive towards actor2, do an awareness check. If it passes, start combat with actor2.
         if (aggressive)
         {
-            bool LOS = MWBase::Environment::get().getWorld()->getLOS(actor1, actor2);
-            LOS &= MWBase::Environment::get().getMechanicsManager()->awarenessCheck(actor2, actor1);
+            bool LOS = MWBase::Environment::get().getWorld()->getLOS(actor1, actor2)
+                    && MWBase::Environment::get().getMechanicsManager()->awarenessCheck(actor2, actor1);
 
             if (LOS)
                 MWBase::Environment::get().getMechanicsManager()->startCombat(actor1, actor2);
@@ -1205,8 +1205,8 @@ namespace MWMechanics
             if (neighbor == actor)
                 continue;
 
-            bool result = MWBase::Environment::get().getWorld()->getLOS(neighbor, actor);
-            result &= MWBase::Environment::get().getMechanicsManager()->awarenessCheck(actor, neighbor);
+            bool result = MWBase::Environment::get().getWorld()->getLOS(neighbor, actor)
+                       && MWBase::Environment::get().getMechanicsManager()->awarenessCheck(actor, neighbor);
 
             if (result)
                 return true;
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -309,6 +309,12 @@ namespace MWMechanics
     void Actors::updateHeadTracking(const MWWorld::Ptr& actor, const MWWorld::Ptr& targetActor,
                                     MWWorld::Ptr& headTrackTarget, float& sqrHeadTrackDistance)
     {
+        if (!actor.getRefData().getBaseNode())
+            return;
+
+        if (targetActor.getClass().getCreatureStats(targetActor).isDead())
+            return;
+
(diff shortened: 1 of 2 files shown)


apps/openmw/mwmechanics/actors.cpp
```diff
@@ -415,20 +415,24 @@ namespace MWMechanics
...
+                            MWBase::Environment::get().getMechanicsManager()->startCombat(ally1, actor2);
                         return;
                     }
                 }
             }
         }
 
+        // Stop here if target is unreachable
+        if (!canFight(actor1, actor2))
+            return;
```
AnyOldName3: Is this reasonable to move before the followers are told to attack? If the player can't attack, but the follower can, this would stop them.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -1200,17 +1197,16 @@ namespace MWMechanics
...
         {
-            if (*it == actor)
+            if (neighbor == actor)
                 continue;
 
-            bool result =
-                MWBase::Environment::get().getWorld()->getLOS(*it, actor) &&
-                MWBase::Environment::get().getMechanicsManager()->awarenessCheck(actor, *it);
+            bool result = MWBase::Environment::get().getWorld()->getLOS(neighbor, actor);
+            result &= MWBase::Environment::get().getMechanicsManager()->awarenessCheck(actor, neighbor);
```
AnyOldName3: This doesn't short-circuit any more, and there isn't a `&&=` operator in C++, so it can't be made to. Do we care?

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1209,36 +1206,20 @@ namespace MWMechanics
 
         // Did anyone see it?
         bool crimeSeen = false;
-        for (std::vector<MWWorld::Ptr>::iterator it = neighbors.begin(); it != neighbors.end(); ++it)
+        for (const MWWorld::Ptr &neighbor : neighbors)
         {
-            if (*it == player)
-                continue; // skip player
```
AnyOldName3: This check seems to have disappeared.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1209,36 +1206,20 @@ namespace MWMechanics
...
-                // Crime reporting only applies to NPCs
-                if (!it->getClass().isNpc())
-                    continue;
-
-                if (it->getClass().getCreatureStats(*it).getAiSequence().isInCombat(victim))
-                    continue;
-
-                if (playerFollowers.find(*it) != playerFollowers.end())
-                    continue;
-
```
AnyOldName3: In fact, most of the checks here seem to not have been replaced. There's a lot more red than green here.

apps/openmw/mwmechanics/actors.cpp
Capostrophic: Valid notion.

apps/openmw/mwmechanics/actors.cpp
Capostrophic: But &= is used in the same manner at line 495.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -129,10 +129,9 @@ namespace MWMechanics
                 npcStats.getSkill (i).setBase (5 + bonus);
             }
 
-            for (std::vector<std::string>::const_iterator iter (race->mPowers.mList.begin());
-                iter!=race->mPowers.mList.end(); ++iter)
+            for (const std::string power : race->mPowers.mList)
```
elsid: Any reason to copy each item?

Capostrophic: I don't think there is. Thanks.

psi29a:
Ready or is there more to come?

Capostrophic:
If nobody has any more objections, yes.

I'm still pretty sure !canFight check does the same thing it was supposed to do before, since the value of aggressive boolean is only evaluated in the end of the function and the operations under followersAttackOnSight check are unnecessary to make if that check succeeds.

psi29a:
No more comments since the weekend and it looks good to me. Merging.
Thanks!

PR #2026, opened 2018-11-06
use c++11 std::align from <memory>

psi29a:
* cleans up .travis-ci.yml
* uses named matrix that we can now extend
* clang static analysis done in two parts: openmw and openmw-cs
* replaces old precise clang (3.6) with trusty clang (6.0)
* replaces gcc 4.8 (Trusty default) with gcc 8 (for Trusty)
* replaces inline align with c++11 std::align


psi29a:
@akortunov @elsid 
This OK for you guys?

akortunov:
It seems the Clang does not have the std::align too. If we need to drop support of older compilers just for one function, maybe we should keep our current implementation?

psi29a:
The compilers have always been available for Trusty and gcc 4.8 and clang-3.6 are not fully C++11 complaint anyway. 

Trusty is EoL in April 2019 anyway.

So we'll be bumping compilers anyway when Travis-CI upgrades to Xenial which is now in Beta.

This is already the case in AppVeyor.

psi29a:
We've used clang's static analyzer forever now, there tons of flags we can set as well. We can investigate those in another PR.

Let's get this squashed and merged.

CI/before_script.linux.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh -e
+#!/bin/bash -ex
```
symbiogenesis: did this need to switch to bash specifically? Isn't it better to be agnostic?

psi29a: dash (default sh in ubuntu) doesn't support =~ operator that I needed when testing for clang, so bash it is.

However, if you can do the same and set it bash sh, please send a PR :)

Don't forget, this is for CI builds on Travis-CI and AppVeyor where bash is expected. Linux based releases are not built with these scripts, that is the job of the package managers.

PR #2027, opened 2018-11-07
Tweak the default chargen and level up messages

Capostrophic:
I didn't really like the wordings and the grammar of the character generation questions and level up messages psi29a has made up. While I didn't rewrite them from scratch again - I don't have *that* much imagination and free time - I decided to at least adjust some while keeping generally the same ideas.


psi29a:
Oh! I like these. :)

Thanks!

Capostrophic:
Travis didn't report that it has finished building for some reason.

Capostrophic:
We can't use Bethesda content because it's copyrighted.

i30817:
Might as fill fill them up with the other possible levels from this 'mod':
https://www.nexusmods.com/morrowind/mods/43431/

(it's just a rtf file). Gets pretty corny on the level 70 but everyone and their siblings can get to level 21+ with Tamriel Rebuilt.

psi29a:
@i30817 not without permission we won't. The authors of that file all have to agree to re-license their work under the GPL or other suitable license like CC-BY, all of them.

PR #2028, opened 2018-11-07
[Testing needed] Use predetermined cost for AI spell and enchanted item rating

Capostrophic:
In Morrowind, spell's predetermined cost is used for AI spell rating, and mod authors sometimes rely on that (for example, some spellcasters in Draugr Diversity rely on spells with huge cost but a very low duration to make something work). OpenMW currently always autocalculates the spell cost based on the real power of its effects. But it's possible to use enchantment and spell predetermined cost instead of doing that. Effect-specific no-ops, rating buffs, debuffs and whatnot are still used, the cost is just a multiplier to the final rating. Nothing has changed for potions and cast once enchantments.

I removed the racial ability check. Racial powers have zero cost and as such will never be used by NPCs anyways, the same is most probably the case in Morrowind too (MCP might only adjust the rating for zero-cost *powers* and not spells)

I'll probably have limited ability to work on anything starting with this weekend.

I didn't actually test this past testing if spellcasting hasn't been broken in general. But that seemed to be fine. Testing needs to be done before I file a tracker bug report.

### DIFF
--- a/apps/openmw/mwmechanics/spellpriority.cpp
+++ b/apps/openmw/mwmechanics/spellpriority.cpp
@@ -100,7 +100,7 @@ namespace MWMechanics
             return 0.f;
 
         const ESM::Potion* potion = item.get<ESM::Potion>()->mBase;
-        return rateEffects(potion->mEffects, actor, MWWorld::Ptr());
+        return rateEffects(potion->mEffects, actor, MWWorld::Ptr(), true);
     }
 
     float rateSpell(const ESM::Spell *spell, const MWWorld::Ptr &actor, const MWWorld::Ptr& enemy)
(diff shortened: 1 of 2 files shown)


akortunov:
This code does not seem correct since you multiply generated rating to spell cost - a resulting spell rating can be huge in this case.
Actually, we already take the spell cost in account, but always use the autogenerated one, even if spell has no AutoCalc flag.

Capostrophic:
The cost should be divided by the effect count, I guess.
I mean, yes, the available magicka and autocalculated NPC spellbook do depend on the predetermined cost of the spell, but that's kind of obvious and doesn't actually affect the action priority directly. Are there other instances of using the field I'm not aware of?

Capostrophic:
I won't be able to work on this for a while and I'll need to make another take on this when I am, so closing

PR #2030, opened 2018-11-08
Fix some compiler warnings

akortunov:
Related [forum thread](https://forum.openmw.org/viewtopic.php?f=6&t=5529&p=59456#p59456).

Note: I am still not sure about memory leaks in shortcuts.

@zinnschlag, @psi29a, what do you think?


Capostrophic:
Qt foreach is actually different from a normal range-based loop because it preserves something Qt needs -- look it up. I'm not sure if its usage causes the warning. You made another edit to the loop, right?

akortunov:
> You made another edit to the loop, right?

Currently we create a QListWidgetItem even if we do not add it to list (since the path is duplicated).

psi29a:
Zini approved, merging.

PR #2031, opened 2018-11-08
Fix some Clang 7 warnings

akortunov:
Let see what will happen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ matrix:
       sudo: required
       env:
         - MATRIX_EVAL="CC=clang-7 && CXX=clang++-7"
-        - ANALYZE="scan-build-7 --use-cc clang-7 --use-c++ clang++-7"
+        - ANALYZE="scan-build-7 --force-analyze-debug-code --use-cc clang-7 --use-c++ clang++-7"
         - BUILD_OPENMW_CS="OFF"
       compiler: clang
     - name: OpenMW (openmw-cs) on Ubuntu Trusty Clang-7 with Static Analysis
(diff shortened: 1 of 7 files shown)


akortunov:
Well, basically it works.
Of course, there is still a lot of issues with RecastNavigation, also there is a lot of warnings with possible memory leaks in the editor (but I think they are false-positives).

Now let's try to use a verbose output for scan-build.

extern/recastnavigation/Detour/Include/DetourAssert.h
```diff
@@ -22,7 +22,15 @@
...
-#ifdef NDEBUG
+#ifdef __clang_analyzer__
+
+#	include <assert.h>
+#	define dtAssert(expression) \
+		{ \
+			assert(expression); \
+		}
+
+#elif defined NDEBUG
```
psi29a: Will we try to get this upstreamed if it works?

akortunov: Unfortunately, it does not work - there is still 23 warnings.

akortunov:
I wonder if there is a way to take a look at generated analyzer report - in theory it should contain info about which if-branches analyzer took to generate warnings.

extern/recastnavigation/Detour/Include/DetourAssert.h
psi29a: ok ;)

psi29a:
If you want, I can run it locally and send you the zip file. I'll be out of town this weekend though.

akortunov:
> If you want, I can run it locally and send you the zip file. I'll be out of town this weekend though.

Yes, please, if it is possible. I tried to run it locally (with Clang 6), but it does not detect any issues for me.
Probably I just do something wrong.

akortunov:
Also I managed to shut up more than half of issues.
The approach used here:
1. Make Clang analyzer to use a common asserts instead of custom Recast/Detour ones.
2. Add some additional asserts to prevent lang from generating false-positives.

psi29a:
https://mega.nz/#!jYd32KpB!Tor4HFmR2Vyq236cmLGHtEmOepQTHtNCR_i27dvgY2k

here you go, should be about 42 bugs all from master (not your branch).

akortunov:
Ok, Detour-specific warnings are moved to separate branch. Only OpenMW-specific warnings left.

psi29a:
Sweet, thanks! :)

PR #2032, opened 2018-11-11
fix win32 app manifest copy-paste error

sthalik:
There's an error in the win32 manifest file I originally sent, referencing other application's name. Fix this.

In case any of you are worried–I'm very confident the manifest file doesn't fall under copyright. There's no violation by copying it over from another project. It's something from MSDN people reuse all the time to get win32 working correctly (UAC, DPI).

sh


psi29a:
Since this is copied from q2pro (I'm assuming here), it's GPL anyway:
https://github.com/skullernet/q2pro/blob/master/LICENSE

Not worried.

PR #2033, opened 2018-11-12
test this, should be ok

psi29a:


psi29a:
obviously not, -test apparently is newer than -ppa (which is now deprecated).  Good to know.

PR #2034, opened 2018-11-13
Improve GUI cleanup

akortunov:
Fixes [bug #4714](https://gitlab.com/OpenMW/openmw/issues/4714) and [bug #4715](https://gitlab.com/OpenMW/openmw/issues/4715).
Currently we cleanup tooltips by calling the 
```
mToolTips->setFocusObject(MWWorld::Ptr());
```
upon game load, but this method also tries to check if it needs to create new tooltips by reading data from focused widget.
Since we call it in the middle of GUI system cleanup, it may lead to undefined behaviour (if the focused widget already was cleaned up, for example).

A suggested solution: provide a clear() method for ToolTips class, which does only tooltips cleanup and does not try to create anything.

As for dialogue weirdness, there is just a missing check if actor pointer is empty.

Note: I did not create a separate PR since there will be merge conflicts in the CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
     Bug #4540: Rain delay when exiting water
     Bug #4701: PrisonMarker record is not hardcoded like other markers
     Bug #4714: Crash upon game load in the repair menu while the "Your repair failed!" message is active
+    Bug #4715: "Cannot get class of an empty object" exception after pressing ESC in the dialogue mode
     Feature #2229: Improve pathfinding AI
     Feature #3442: Default values for fallbacks from ini file
     Feature #4673: Weapon sheathing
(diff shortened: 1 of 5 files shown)


akortunov:
there is definely something wrong with Travis Clang builds.

psi29a:
We had issues with MSVC2015 on AV as well because no more output was generated, we had to mark that build as 'allowed to fail' to handle it.

Clang was also set to 'allowed to fail' as well until recently. If it keeps up, we'll switch it back.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
     Bug #3623: Fix HiDPI on Windows
     Bug #4540: Rain delay when exiting water
     Bug #4701: PrisonMarker record is not hardcoded like other markers
+    Bug #4714: Crash upon game load in the repair menu while the "Your repair failed!" message is active
```
psi29a: Is this something we need to worry about with 0.45?

psi29a: Same with Bug 4715?

wareya: I encountered the crash before but it only occurs when you're savescumming and you'll always have a save to fall back on when reopening the game, so it wouldn't be a blocker. It's also not a recent regression.

psi29a: OK

PR #2035, opened 2018-11-13
Fix many Coverity Scan warnings

akortunov:
I still suggest to filter autogenerated files and external dependencies again since they cause spam of warnings which we can not fix.


psi29a:
Working on it. ;)

psi29a:
Looks like I've been able to fix things. Again, caused by local build. I made the regex's position independent.

psi29a:
After this is merged, I'll run another round.

akortunov:
Build error is fixed.

psi29a:
Wonderful work, thank you! :)

Going to re-run coverity in a minute.

PR #2036, opened 2018-11-13
Update OpenAL-soft to 1.19.1

ananace:
As requested for the underwater audio bug


psi29a:
Going to merge this, but it would be great if we could get away from magic numbers (versions) and use a constant in the future. :)

PR #2037, opened 2018-11-14
Fix warnings in the OpenMW-CS and NIFTest

akortunov:
If it is OK, it should fix 7 or 8 Coverity Scan warnings.

### DIFF
--- a/apps/niftest/niftest.cpp
+++ b/apps/niftest/niftest.cpp
@@ -10,6 +10,7 @@
 #include <components/vfs/bsaarchive.hpp>
 #include <components/vfs/filesystemarchive.hpp>
 
+#include <boost/exception/all.hpp>
 #include <boost/program_options.hpp>
 #include <boost/filesystem.hpp>
 
@@ -128,14 +129,24 @@ std::vector<std::string> parseOptions (int argc, char** argv)
 
 int main(int argc, char **argv)
 {
-    std::vector<std::string> files = parseOptions (argc, argv);
+    std::vector<std::string> files;
+    try
+    {
+        files = parseOptions (argc, argv);
+    }
+    catch( boost::exception &e )
+    {
+        std::cout << "ERROR parsing arguments: " << boost::diagnostic_information(e) << std::endl;
+        exit(1);
+    }
 
 //     std::cout << "Reading Files" << std::endl;
     for(std::vector<std::string>::const_iterator it=files.begin(); it!=files.end(); ++it)
     {
-         std::string name = *it;
+        std::string name = *it;
 
-        try{
+        try
+        {
             if(isNIF(name))
             {
                 //std::cout << "Decoding: " << name << std::endl;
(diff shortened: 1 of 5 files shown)


apps/niftest/niftest.cpp
```diff
@@ -128,14 +129,24 @@ std::vector<std::string> parseOptions (int argc, char** argv)
...
-    std::vector<std::string> files = parseOptions (argc, argv);
+    std::vector<std::string> files;
+    try
+    {
+        files = parseOptions (argc, argv);
+    }
+    catch( boost::exception &e )
+    {
+        std::cout << "ERROR parsing arguments: " << boost::diagnostic_information(e) << std::endl;
+        exit(1);
```
zinnschlag: Don't use exit in C++ code. It is not guaranteed to do its job cleanly (some cleanup may be left undone). Doesn't look like that is the problem here, but  it is still kinda dirty. I suggest to use return 1 instead.

Also, is there actually a need to pull in boost exceptions? I would suspect that a std::exception should do the job just fine.

apps/opencs/model/world/commands.cpp
```diff
@@ -199,8 +199,10 @@ CSMWorld::ModifyCommand::ModifyCommand (QAbstractItemModel& model, const QModelI
     if (mIndex.parent().isValid())
     {
         CSMWorld::IdTree* tree = dynamic_cast<CSMWorld::IdTree*>(mModel);
-
-        assert(tree != nullptr);
+        if (tree == nullptr)
```
zinnschlag: This is very verbose (and all the code repetition makes it worse. I suggest to cast via reference. This will allow the build in C++ mechanics to take care of everything, e.g.:

CSMWorld::IdTree* tree = &dynamic_cast<CSMWorld::IdTree&>(*mModel);

If this cast fails it will automatically throw a suitable exception.

apps/niftest/niftest.cpp
akortunov: 1. NIFtest already uses an `exit(1)` in different places. If we do not like it, we need to refactor it.
2. Boost exceptions are not derived from std::exception, so std::exception catch more likely will not work, and we will need to catch them separately.
We already use a similar approach in the BSATool::parseOptions().

apps/opencs/model/world/commands.cpp
akortunov: Ok.

akortunov:
> Sorry, but not really happy with this PR. I suggest a revert. I'll comment inline about the problems I see.

Since not only affected code has these problems, I suggest to create a separate PR for refactoring - reverting will not make situation better.

apps/niftest/niftest.cpp
psi29a: niftest is a CLI tool and not a library.

psi29a:
I agree with @akortunov, reverting doesn't make much sense here. Another PR to address the issues is fine.

akortunov:
https://github.com/OpenMW/openmw/pull/2040

PR #2039, opened 2018-11-15
Handle last Coverity Scan issues left

akortunov:
CID 83467 - it is a false positive, can be dismissed.

CID 121666 - Coverity Scan complains about possible case when `comments` vector contains values, but the `commentStart == fileCopy.end()` is true. 
We can check iterator just for sure or dismiss this issue too as a false positive.
Not sure which option is better.

CID 134779 - here we parse a `pos` variable from input string and use it as a loop bound, despite actual string size can be lesser.
We can print a warning here, cap `pos` silently or throw an exception.

CID 184322 - related to redundant logging system for pathfinding, which probably should be removed, so there is no need to handle it right now.

### DIFF
--- a/apps/opencs/view/world/scriptsubview.cpp
+++ b/apps/opencs/view/world/scriptsubview.cpp
@@ -7,6 +7,8 @@
 #include <QSplitter>
 #include <QTimer>
 
+#include <components/debug/debuglog.hpp>
+
 #include "../../model/doc/document.hpp"
 #include "../../model/world/universalid.hpp"
 #include "../../model/world/data.hpp"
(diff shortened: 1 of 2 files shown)


akortunov:
Request a review from @zinnschlag.

apps/opencs/view/world/scriptsubview.cpp
```diff
@@ -210,18 +212,28 @@ void CSVWorld::ScriptSubView::useHint (const std::string& hint)
...
             QModelIndex index = mModel->getModelIndex (getUniversalId().getId(), mColumn);
             QString source = mModel->data (index).toString();
+            unsigned stringSize = source.length();
             unsigned pos, dummy;
             if (!(stream >> c >> dummy >> pos) )
                 return;
 
-            for (unsigned i = 0; i <= pos; ++i){
-                if (source[i] == '\n'){
+            if (pos >= stringSize)
```
zinnschlag: This looks like an off-by-one error to me, since it is possible to position the cursor past the last character of the script text.

akortunov: Note that loop uses a <= statement instead of <.
For example, with pos = 1000 stringSize should be > 1000 to handle 1000-th character in string.
So the situation pos = stringSize is invalid too.

Can you explain what do you mean?

zinnschlag: Lets make it simple. Assume a single line file with a single character in it. The string has the size 1. Legal cursor positions then would be 0 (on the character) and 1 (after the character). Note that s[s.length()]==0, so there is no need for a special case. At least that is the case for std::string. I just noticed that we are using a bloody QString here. I am 99% sure that it behaves in the same way as std::string. But it wouldn't hurt to check the documentation.

psi29a: > int QString::length() const
> Returns the number of characters in this string. Equivalent to size().
https://doc.qt.io/qt-5/qstring.html#length

> int QString::size() const
> Returns the number of characters in this string.
> 
> The last character in the string is at position size() - 1.
> 

Example:
```
QString str = "World";
int n = str.size();         // n == 5
str.data()[0];              // returns 'W'
str.data()[4];              // returns 'd'
```

akortunov: Interesting. I tried to use the `str.data()[1000000];`, but it just returns zeroes.
So in theory possible error should cause only redundant iterations, and we can just print warning here.

akortunov: So what did we decide to do here?

akortunov: Increased limit by 1 for sure.

psi29a:
Moving forward with this. Merging, thanks @akortunov

PR #2040, opened 2018-11-15
Requested refactoring

akortunov:
Changes, requested by @zinnschlag.

Summary of changes:
1. Modify dynamic_casts in the editor to throw exceptions automatically.
2. Refactor NIFTest to use `return` statements instead of `exit(1)`.
3. Do not use redundant boost exceptions handlers in ESMTool and NIFTest - according to my tests, std::exception handles a subset of used exceptions indeed.

### DIFF
--- a/apps/esmtool/esmtool.cpp
+++ b/apps/esmtool/esmtool.cpp
@@ -123,14 +123,9 @@ bool parseOptions (int argc, char** argv, Arguments &info)
 
         bpo::store(valid_opts, variables);
     }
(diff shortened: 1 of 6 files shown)


psi29a:
Where there additional boost::exceptions in other bits, like BSATool?

akortunov:
> Where there additional boost::exceptions in other bits, like BSATool?

I do not see any. There were boost exceptions only in esmtool and niftest.

psi29a:
Great, I'm glad that std::exception also does the job.

PR #1547, opened 2017-11-07
Shadows

AnyOldName3:
Scrawl suggested I actually make a PR for this so that it's easier to keep track of while the last few things are sorted out.

# Current issues (that I can think of):

## Blockers
- [x] ~~I've not decided which way to make the sky not receive shadows is best (https://github.com/AnyOldName3/openmw/tree/osgshadow-test-vdsm-sky-method-one versus https://github.com/AnyOldName3/openmw/tree/osgshadow-test-vdsm-sky-method-two)~~ Scrawl likes method two and I trust his judgement above my own.
- [x] The minimap needs shadows turning off.
- [x] The character preview needs shadows turning off.
- [x] Interiors need shadows to not be cast by walls etc. (as they use a 'fake' light source which is blocked everywhere)
- [x] Settings need to be changeable by users instead of hardcoded.
- [x] Description of settings needs adding to the documentation
- [x] Turning on shadows should ~~potentially somehow warn users that they might get issues with shaders off unless~~ force shaders on, but maybe the documentation should suggest they want to [do this](https://github.com/OpenMW/openmw/pull/1547#issuecomment-369657381)
- [x] Assorted parameter tuning to make things look as good as possible.
- [x] Move as much as possible out of `shadermanager.cpp`
- [x] Separate `MWShadow` into two classes - a shadow technique class based directly on `osgShadow::ViewDependentShadowMap` and a shadow manager class handling interaction with the rest of the engine.
- [x] ~~Work out why some Linux users are seeing issues that I'm not, and fix the problem.~~ (This wasn't Linux-specific after all, and was just that distant land was on and broke things.)
- [x] Ensure that shadow map projection keeps shadow map texels roughly 1:1 with screen pixels.
* - [x] Fix this when due to perspective errors.
* - [x] Fix this when due to a high viewing distance.
* - [x] Fix this when due to ~~similar light and camera angles~~ *that wasn't the issue* shadow maps being set up to cover non-existent objects under the ground.
   * - [x] Stop that fix from breaking first person meshes.
   * - [x] Stop that fix from sometimes making shadows disappear from near to the camera when the closest thing to the camera is terrain, not an object.
* - [x] Fix this when due to similar light and camera angles as this issue seems to have ~~reappeared~~ never actually been solved?
   * - [x] See how easily this can be fixed by messing around with Cascading Shadow Maps.
   * - [x] Implement a maintainable and configurable implementation of Cascading Shadow Maps.
[...]

PR #2042, opened 2018-11-19
[abandoned] Euclidean distance fog

sthalik:
It still looks suboptimally when the terrain is close to the fog distance limit. Results in white mountains outside the range. The view and fog distance must be small enough, or large enough to accomodate this.

Note that overall view distance uses frustum culling. It must be a bit than the fog distance to accomodate this difference. We could use the `discard` statement in GLSL but it has performance implications. Ideally we could cull using Euclidean distance prior to sending the vertices to the GPU.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -242,7 +242,19 @@ namespace MWRender
         resourceSystem->getSceneManager()->setAutoUseSpecularMaps(Settings::Manager::getBool("auto use object specular maps", "Shaders"));
         resourceSystem->getSceneManager()->setSpecularMapPattern(Settings::Manager::getString("specular map pattern", "Shaders"));
 
+
+        if (resourceSystem->getSceneManager()->getAccurateFog() &&
+            !resourceSystem->getSceneManager()->getForceShaders())
+        {
+            static bool ext = osg::isGLExtensionSupported(0, "GL_NV_fog_distance");
+
+            // force shader-based fog if lacking ffp extension
+            if (!ext)
+                resourceSystem->getSceneManager()->setForceShaders(true);
+        }
+
         osg::ref_ptr<SceneUtil::LightManager> sceneRoot = new SceneUtil::LightManager;
+
         sceneRoot->setLightingMask(Mask_Lighting);
         mSceneRoot = sceneRoot;
         sceneRoot->setStartLight(1);
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -222,6 +222,7 @@ namespace MWRender
         resourceSystem->getSceneManager()->setForceShaders(Settings::Manager::getBool("force shaders", "Shaders"));
         resourceSystem->getSceneManager()->setClampLighting(Settings::Manager::getBool("clamp lighting", "Shaders"));
         resourceSystem->getSceneManager()->setForcePerPixelLighting(Settings::Manager::getBool("force per pixel lighting", "Shaders"));
+        resourceSystem->getSceneManager()->setAccurateFog(Settings::Manager::getBool("accurate fog", "Fog"));
         resourceSystem->getSceneManager()->setAutoUseNormalMaps(Settings::Manager::getBool("auto use object normal maps", "Shaders"));
         resourceSystem->getSceneManager()->setNormalMapPattern(Settings::Manager::getString("normal map pattern", "Shaders"));
         resourceSystem->getSceneManager()->setNormalHeightMapPattern(Settings::Manager::getString("normal height map pattern", "Shaders"));
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -254,7 +254,6 @@ namespace MWRender
         }
 
         osg::ref_ptr<SceneUtil::LightManager> sceneRoot = new SceneUtil::LightManager;
-
         sceneRoot->setLightingMask(Mask_Lighting);
         mSceneRoot = sceneRoot;
         sceneRoot->setStartLight(1);
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -88,6 +88,7 @@ namespace MWRender
             : mFogStart(0.f)
             , mFogEnd(0.f)
             , mWireframe(false)
+            , mAccurateFog(false)
         {
         }
 
@@ -97,6 +98,8 @@ namespace MWRender
             stateset->setAttribute(lightModel, osg::StateAttribute::ON);
             osg::Fog* fog = new osg::Fog;
             fog->setMode(osg::Fog::LINEAR);
+            if (mAccurateFog)
+                fog->setUseRadialFog(true);
             stateset->setAttributeAndModes(fog, osg::StateAttribute::ON);
             if (mWireframe)
             {
@@ -152,12 +155,22 @@ namespace MWRender
             return mWireframe;
         }
 
+        void setAccurateFog(bool enable)
+        {
+            mAccurateFog = enable;
+        }
+
+        bool getAccurateFog() const
+        {
+            return mAccurateFog;
+        }
+
     private:
         osg::Vec4f mAmbientColor;
         osg::Vec4f mFogColor;
         float mFogStart;
         float mFogEnd;
-        bool mWireframe;
+        bool mWireframe, mAccurateFog;
     };
 
     class PreloadCommonAssetsWorkItem : public SceneUtil::WorkItem
@@ -310,6 +323,8 @@ namespace MWRender
         source->setStateSetModes(*mRootNode->getOrCreateStateSet(), osg::StateAttribute::ON);
 
         mStateUpdater = new StateUpdater;
+        mStateUpdater->setAccurateFog(resourceSystem->getSceneManager()->getAccurateFog());
+
         sceneRoot->addUpdateCallback(mStateUpdater);
 
         osg::Camera::CullingMode cullingMode = osg::Camera::DEFAULT_CULLING|osg::Camera::FAR_PLANE_CULLING;
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -528,6 +528,8 @@ void Water::createShaderWaterStateSet(osg::Node* node, Reflection* reflection, R
     // use a define map to conditionally compile the shader
     std::map<std::string, std::string> defineMap;
     defineMap.insert(std::make_pair(std::string("refraction_enabled"), std::string(refraction ? "1" : "0")));
+    bool accurateFog = mResourceSystem->getSceneManager()->getAccurateFog();
+    defineMap.insert(std::make_pair(std::string("accurateFog"), std::string(accurateFog ? "1" : "0")));
 
     Shader::ShaderManager& shaderMgr = mResourceSystem->getSceneManager()->getShaderManager();
     osg::ref_ptr<osg::Shader> vertexShader (shaderMgr.getShader("water_vertex.glsl", defineMap, osg::Shader::VERTEX));
--- a/components/shader/shadervisitor.cpp
+++ b/components/shader/shadervisitor.cpp
@@ -45,6 +45,7 @@ namespace Shader
         , mAllowedToModifyStateSets(true)
         , mAutoUseNormalMaps(false)
         , mAutoUseSpecularMaps(false)
+        , mAccurateFog(false)
         , mShaderManager(shaderManager)
         , mImageManager(imageManager)
         , mDefaultVsTemplate(defaultVsTemplate)
@@ -292,6 +293,7 @@ namespace Shader
         }
 
         defineMap["forcePPL"] = mForcePerPixelLighting ? "1" : "0";
+        defineMap["accurateFog"] = mAccurateFog ? "1" : "0";
         defineMap["clamp"] = mClampLighting ? "1" : "0";
 
         defineMap["parallax"] = reqs.mNormalHeight ? "1" : "0";
@@ -424,6 +426,11 @@ namespace Shader
     void ShaderVisitor::setSpecularMapPattern(const std::string &pattern)
     {
         mSpecularMapPattern = pattern;
+    }
+
+    void ShaderVisitor::setAccurateFog(bool enable)
+    {
+        mAccurateFog = enable;
     }
 
 }
--- a/components/terrain/chunkmanager.cpp
+++ b/components/terrain/chunkmanager.cpp
@@ -220,8 +220,14 @@ osg::ref_ptr<osg::Node> ChunkManager::createChunk(float chunkSize, const osg::Ve
         layer.mDiffuseMap = compositeMap->mTexture;
         layer.mParallax = false;
         layer.mSpecular = false;
-        geometry->setPasses(::Terrain::createPasses(mSceneManager->getForceShaders() || !mSceneManager->getClampLighting(), mSceneManager->getForcePerPixelLighting(),
-                                                    mSceneManager->getClampLighting(), &mSceneManager->getShaderManager(), std::vector<TextureLayer>(1, layer), std::vector<osg::ref_ptr<osg::Texture2D> >(), 1.f, 1.f));
+        geometry->setPasses(::Terrain::createPasses(mSceneManager->getForceShaders(),
+                                                    mSceneManager->getForcePerPixelLighting(),
+                                                    mSceneManager->getClampLighting(),
+                                                    mSceneManager->getAccurateFog(),
+                                                    &mSceneManager->getShaderManager(),
+                                                    std::vector<TextureLayer>(1, layer),
+                                                    std::vector<osg::ref_ptr<osg::Texture2D>>(),
+                                                    1.f, 1.f);
     }
     else
     {
--- a/components/terrain/chunkmanager.cpp
+++ b/components/terrain/chunkmanager.cpp
@@ -164,8 +164,13 @@ std::vector<osg::ref_ptr<osg::StateSet> > ChunkManager::createPasses(float chunk
 
     float blendmapScale = mStorage->getBlendmapScale(chunkSize);
 
-    return ::Terrain::createPasses(useShaders, mSceneManager->getForcePerPixelLighting(),
-                                     mSceneManager->getClampLighting(), &mSceneManager->getShaderManager(), layers, blendmapTextures, blendmapScale, blendmapScale);
+    return ::Terrain::createPasses(useShaders,
+                                   mSceneManager->getForcePerPixelLighting(),
+                                   mSceneManager->getClampLighting(),
+                                   mSceneManager->getAccurateFog(),
+                                   &mSceneManager->getShaderManager(),
+                                   layers, blendmapTextures,
+                                   blendmapScale, blendmapScale);
 }
 
 osg::ref_ptr<osg::Node> ChunkManager::createChunk(float chunkSize, const osg::Vec2f &chunkCenter, int lod, unsigned int lodFlags)
@@ -220,14 +225,14 @@ osg::ref_ptr<osg::Node> ChunkManager::createChunk(float chunkSize, const osg::Ve
         layer.mDiffuseMap = compositeMap->mTexture;
         layer.mParallax = false;
         layer.mSpecular = false;
-        geometry->setPasses(::Terrain::createPasses(mSceneManager->getForceShaders(),
+        geometry->setPasses(::Terrain::createPasses(mSceneManager->getForceShaders() || !mSceneManager->getClampLighting(),
                                                     mSceneManager->getForcePerPixelLighting(),
                                                     mSceneManager->getClampLighting(),
                                                     mSceneManager->getAccurateFog(),
                                                     &mSceneManager->getShaderManager(),
                                                     std::vector<TextureLayer>(1, layer),
                                                     std::vector<osg::ref_ptr<osg::Texture2D>>(),
-                                                    1.f, 1.f);
+                                                    1.f, 1.f));
     }
     else
     {
--- a/files/settings-default.cfg
+++ b/files/settings-default.cfg
@@ -108,6 +108,10 @@ distant interior fog start = 0
 
 distant interior fog end = 16384
 
+# When disabled, the visible distance is larger on screen edges. Enabling
+# the setting makes fog distance accurate all around the player.
+accurate fog = false
+
 [Map]
 
 # Size of each exterior cell in pixels in the world map. (e.g. 12 to 24).
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -42,9 +42,7 @@ uniform sampler2D specularMap;
 varying vec2 specularMapUV;
 #endif
 
-#if !@accurateFog
 varying float depth;
-#endif
 
 #define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
 
@@ -78,11 +76,9 @@ void main()
     vec3 viewNormal = gl_NormalMatrix * normalize(passNormal);
 #endif
 
-#if @parallax || @accurateFog
+#if @parallax
     vec3 cameraPos = (gl_ModelViewMatrixInverse * vec4(0,0,0,1)).xyz;
     vec3 objectPos = (gl_ModelViewMatrixInverse * vec4(passViewPos, 1)).xyz;
-#endif
-#if @parallax
     vec3 eyeDir = normalize(cameraPos - objectPos);
     vec2 offset = getParallaxOffset(eyeDir, tbnTranspose, normalTex.a, (passTangent.w > 0) ? -1.f : 1.f);
     adjustedDiffuseUV += offset; // only offset diffuse for now, other textures are more likely to be using a completely different UV set
@@ -153,9 +149,6 @@ void main()
 
     gl_FragData[0].xyz += getSpecular(normalize(viewNormal), normalize(passViewPos.xyz), shininess, matSpec);
 
-#if @accurateFog
-    float depth = length(objectPos - cameraPos);
-#endif
     float fogValue = clamp((depth - gl_Fog.start) * gl_Fog.scale, 0.0, 1.0);
     gl_FragData[0].xyz = mix(gl_FragData[0].xyz, gl_Fog.color.xyz, fogValue);
 }
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -42,7 +42,9 @@ uniform sampler2D specularMap;
 varying vec2 specularMapUV;
 #endif
 
+#if !@accurateFog
 varying float depth;
+#endif
 
 #define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
 
@@ -76,9 +78,11 @@ void main()
     vec3 viewNormal = gl_NormalMatrix * normalize(passNormal);
 #endif
 
-#if @parallax
+#if @parallax || @accurateFog
     vec3 cameraPos = (gl_ModelViewMatrixInverse * vec4(0,0,0,1)).xyz;
     vec3 objectPos = (gl_ModelViewMatrixInverse * vec4(passViewPos, 1)).xyz;
+#endif
+#if @parallax
     vec3 eyeDir = normalize(cameraPos - objectPos);
     vec2 offset = getParallaxOffset(eyeDir, tbnTranspose, normalTex.a, (passTangent.w > 0) ? -1.f : 1.f);
     adjustedDiffuseUV += offset; // only offset diffuse for now, other textures are more likely to be using a completely different UV set
@@ -149,6 +153,9 @@ void main()
 
     gl_FragData[0].xyz += getSpecular(normalize(viewNormal), normalize(passViewPos.xyz), shininess, matSpec);
 
+#if @accurateFog
+    float depth = length(objectPos - cameraPos);
+#endif
     float fogValue = clamp((depth - gl_Fog.start) * gl_Fog.scale, 0.0, 1.0);
     gl_FragData[0].xyz = mix(gl_FragData[0].xyz, gl_Fog.color.xyz, fogValue);
 }
--- a/files/shaders/objects_vertex.glsl
+++ b/files/shaders/objects_vertex.glsl
@@ -33,7 +33,9 @@ varying vec2 envMapUV;
 varying vec2 specularMapUV;
 #endif
 
+#if !@accurateFog
 varying float depth;
+#endif
 
 #define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
 
@@ -50,7 +52,9 @@ varying vec3 passNormal;
 void main(void)
 {
     gl_Position = gl_ModelViewProjectionMatrix * gl_Vertex;
+#if !@accurateFog
     depth = gl_Position.z;
+#endif
 
     vec4 viewPos = (gl_ModelViewMatrix * gl_Vertex);
     gl_ClipVertex = viewPos;
--- a/files/shaders/objects_vertex.glsl
+++ b/files/shaders/objects_vertex.glsl
@@ -33,9 +33,7 @@ varying vec2 envMapUV;
 varying vec2 specularMapUV;
 #endif
 
-#if !@accurateFog
 varying float depth;
-#endif
 
 #define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
 
@@ -52,13 +50,15 @@ varying vec3 passNormal;
 void main(void)
 {
     gl_Position = gl_ModelViewProjectionMatrix * gl_Vertex;
-#if !@accurateFog
-    depth = gl_Position.z;
-#endif
-
     vec4 viewPos = (gl_ModelViewMatrix * gl_Vertex);
     gl_ClipVertex = viewPos;
     vec3 viewNormal = normalize((gl_NormalMatrix * gl_Normal).xyz);
+
+#if !@accurateFog
+    depth = gl_Position.z;
+#else
+    depth = length(viewPos.xyz);
+#endif
 
 #if @envMap
     vec3 viewVec = normalize(viewPos.xyz);
--- a/files/shaders/terrain_fragment.glsl
+++ b/files/shaders/terrain_fragment.glsl
@@ -12,7 +12,9 @@ uniform sampler2D normalMap;
 uniform sampler2D blendMap;
 #endif
 
+#if !@accurateFog
 varying float depth;
+#endif
 
 #define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
 
@@ -45,9 +47,11 @@ void main()
     vec3 viewNormal = normalize(gl_NormalMatrix * passNormal);
 #endif
 
-#if @parallax
+#if @parallax || @accurateFog
     vec3 cameraPos = (gl_ModelViewMatrixInverse * vec4(0,0,0,1)).xyz;
     vec3 objectPos = (gl_ModelViewMatrixInverse * vec4(passViewPos, 1)).xyz;
+#endif
+#if @parallax
     vec3 eyeDir = normalize(cameraPos - objectPos);
     adjustedUV += getParallaxOffset(eyeDir, tbnTranspose, normalTex.a, 1.f);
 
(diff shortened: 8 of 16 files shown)


psi29a:
Is or could this eventually take weather into account? A clear sunny day vs a rain storms vs a ash storm for example?

AnyOldName3:
The gist of this looks sensible, but as I mentioned on Discord, it needs to be fenced off behind a setting. As there are situations where most of the scene is rendered in fixed-function mode with just a few objects using shaders, we need the old behaviour to be available so those few objects don't stand out. My recommendation would be to add a new setting in the `[Fog]` section, and when it's enabled, force-enable the `force shaders` setting as I've done here: https://github.com/OpenMW/openmw/pull/1547/files#diff-f1edf5c7ed145cbb380aad92dedbf3bbR223

Another thing to note is that you've left out the water shader.

sthalik:
@psi29a I was able to verify that visibility in Ald'Ruhn during an ash storm is minimal.

@AnyOldName3 is it enough to guard on `force shaders = yes`? There's no discernible performance impact, unlike shadows.

AnyOldName3:
It's probably best to have it as a separate setting - we want setting names to be descriptive, and `force shaders` doesn't say anything about improving the fog.

sthalik:
@AnyOldName3 is this more than an implementation detail? I'd consider the frustum-based fog distance discrepancy as a limitation of the fixed pipeline. I recommend against expecting no differences in fixed vs shader pipeline. In the end it's not my decision to make.

I tested this version for both cases and it *seems* to work.

There's some discrepancy in `ChunkManager` where `createPasses()` in one case forces shaders on depending on other settings but leaves them alone in the other overload. There's also no similar setting fixup for drawing objects. Maybe we could do the fixup in `ResourceManager` directly?

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -222,6 +222,7 @@ namespace MWRender
         resourceSystem->getSceneManager()->setForceShaders(Settings::Manager::getBool("force shaders", "Shaders"));
```
AnyOldName3: You're not force-enabling shaders here, so you're still going to end up with a bunch of objects using fixed-function mode.

AnyOldName3: Actually, it looks like you're doing it elsewhere, so I guess what you've got would work. I'm somewhat inclined to say it's best to have it here as then it's guaranteed to be applied everywhere, even if something new is added by someone who doesn't know about it.

components/shader/shadervisitor.cpp
```diff
@@ -250,7 +251,7 @@ namespace Shader
 
     void ShaderVisitor::createProgram(const ShaderRequirements &reqs)
     {
-        if (!reqs.mShaderRequired && !mForceShaders)
+        if (!reqs.mShaderRequired && !mForceShaders && !mAccurateFog)
```
AnyOldName3: As I mentioned elsewhere, it's probably best to have this set up in the rendering manager.

files/settings-default.cfg
```diff
@@ -280,6 +280,10 @@ force per pixel lighting = false
 # Setting this option to 'false' results in more realistic lighting.
 clamp lighting = true
 
+# Controls fog calculation. When disabled, the visible distance is larger on screen edges.
+# Enabling the setting makes fog distance accurate all around the player.
+accurate fog = false
```
AnyOldName3: I really think this setting should be in the `[Fog]` section rather than the `[Shaders]` section. It requires shaders in order to work, but it's the fog that it affects.

files/shaders/objects_fragment.glsl
```diff
@@ -42,9 +42,11 @@ uniform sampler2D specularMap;
 varying vec2 specularMapUV;
 #endif
 
+#if !@accurateFog
 varying float depth;
+#endif
 
-#define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
+#define PER_PIXEL_LIGHTING (@normalMap || @forcePPL || @accurateFog)
```
AnyOldName3: We certainly don't want to force per-pixel lighting when using your fog. Morrowind's assets are stupid and regularly look wrong if they're not done with vertex lighting, so we avoid PPL wherever possible. There's no reason why your fog would make it a requirement.

files/shaders/objects_vertex.glsl
```diff
@@ -33,9 +33,11 @@ varying vec2 envMapUV;
 varying vec2 specularMapUV;
 #endif
 
+#if !@accurateFog
 varying float depth;
+#endif
 
-#define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
+#define PER_PIXEL_LIGHTING (@normalMap || @forcePPL || @accurateFog)
```
AnyOldName3: PPL for no reason again

files/shaders/terrain_fragment.glsl
```diff
@@ -12,9 +12,11 @@ uniform sampler2D normalMap;
 uniform sampler2D blendMap;
 #endif
 
+#if !@accurateFog
 varying float depth;
+#endif
 
-#define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
+#define PER_PIXEL_LIGHTING (@normalMap || @forcePPL || @accurateFog)
```
AnyOldName3: PPL for no reason

files/shaders/terrain_vertex.glsl
```diff
@@ -1,9 +1,12 @@
 #version 120
 
 varying vec2 uv;
-varying float depth;
 
-#define PER_PIXEL_LIGHTING (@normalMap || @forcePPL)
+#define PER_PIXEL_LIGHTING (@normalMap || @forcePPL || @accurateFog)
```
AnyOldName3: PPL for no reason

AnyOldName3:
> I recommend against expecting no differences in fixed vs shader pipeline.

If it's a debate we don't need to have yet, there's not much point having it. It doesn't particularly make this task harder, and there have been a few times I've needed to compare FF and shaders directly, so if they're doing different things, the comparison isn't fair.

components/terrain/chunkmanager.cpp
```diff
@@ -164,8 +166,9 @@ std::vector<osg::ref_ptr<osg::StateSet> > ChunkManager::createPasses(float chunk
 
     float blendmapScale = mStorage->getBlendmapScale(chunkSize);
 
-    return ::Terrain::createPasses(useShaders, mSceneManager->getForcePerPixelLighting(),
-                                     mSceneManager->getClampLighting(), &mSceneManager->getShaderManager(), layers, blendmapTextures, blendmapScale, blendmapScale);
+    return ::Terrain::createPasses(useShaders,
+                                   mSceneManager->getForcePerPixelLighting(), mSceneManager->getClampLighting(), mSceneManager->getAccurateFog(),
+                                   &mSceneManager->getShaderManager(), layers, blendmapTextures, blendmapScale, blendmapScale);
```
AnyOldName3: @zinnschlag @psi29a How do we feel about switching around where line breaks are when it isn't strictly necessary?

sthalik: Good catch! Can these be refactored into a plain struct? Similarly for `SkyManager` and `ShaderVisitor`.

AnyOldName3: We're not the DirectX API. We don't put random structs that only get used in one place everywhere as it's the kind of abstraction which just makes things harder to read.

AnyOldName3:
This whole PR would be a lot cleaner if we had global shader defines. Maybe it's worth cherry picking the stuff that does that from the shadows PR into master and redoing this based on that. How do other people feel about that idea? Do we expect it to cause issues when the shadow PR is merged (I've barely ever cherry picked things, so don't know if it has risks).

components/terrain/chunkmanager.cpp
sthalik: Microsoft's VMS-like APIs are awful, yeah. But from my experience that mostly stems from *initializing* random structs, not passing them around. Also see that in `sky.hpp` there's already a random struct passed, then copied into that class' member variables.

apps/openmw/mwrender/water.cpp
```diff
@@ -528,6 +528,8 @@ void Water::createShaderWaterStateSet(osg::Node* node, Reflection* reflection, R
     // use a define map to conditionally compile the shader
     std::map<std::string, std::string> defineMap;
     defineMap.insert(std::make_pair(std::string("refraction_enabled"), std::string(refraction ? "1" : "0")));
+    bool accurateFog = mResourceSystem->getSceneManager()->getAccurateFog();
+    defineMap.insert(std::make_pair(std::string("accurateFog"), std::string(accurateFog ? "1" : "0")));
```
sthalik: This is way bad, especially given it's the third duplication.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -217,17 +217,26 @@ namespace MWRender
...
+            s.setNormalMapPattern(Settings::Manager::getString("normal map pattern", "Shaders"));
+            s.setNormalHeightMapPattern(Settings::Manager::getString("normal height map pattern", "Shaders"));
+            s.setAutoUseSpecularMaps(Settings::Manager::getBool("auto use object specular maps", "Shaders"));
+            s.setSpecularMapPattern(Settings::Manager::getString("specular map pattern", "Shaders"));
+
+            if (s.getClampLighting())
+                s.setForceShaders(true);
+            if (s.getAccurateFog())
+                s.setForceShaders(true);
+        }
```
AnyOldName3: Doing stuff like this isn't a good idea on big projects with lots of people. It might be a bit tidier now, but you've changed a bunch of stuff that isn't in the scope of this PR, made it harder to use git blame to work out where things came from and made merge conflicts with other stuff a lot more likely.

Also, `s` is not a good variable name, even for a context as small as this. `sceneManager` would have been much better.

psi29a:
> This whole PR would be a lot cleaner if we had global shader defines. Maybe it's worth cherry picking the stuff that does that from the shadows PR into master and redoing this based on that. How do other people feel about that idea? Do we expect it to cause issues when the shadow PR is merged (I've barely ever cherry picked things, so don't know if it has risks).

I call this 'architectural runway' , it is work that is done for the support of a future feature but is useful for other features so they can be worked on in parallel. 

It has the bonus effect of reducing the size of your shadow branch making it that much easier to review.

cherry-picking is just taking one commit and applying somewhere else, it will have a different commit id but the history is clean and you won't create merge conflicts with it when you go to merge your shadow branch in.

Make a new branch, cherry-pick the commits you want and create a new PR.

apps/openmw/mwrender/renderingmanager.cpp
sthalik: Hey, but this *is* self-documenting regardless of variable name length.

components/terrain/chunkmanager.cpp
sthalik: @AnyOldName3 what's worse, long getter chains or receiving a trivial settings struct by reference?

sthalik: To make it clear, I'm NOT telling you how to do things in your own project. Just that your way of doing things is pretty alien and I have trouble adjusting.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -222,12 +222,18 @@ namespace MWRender
...
         resourceSystem->getSceneManager()->setClampLighting(Settings::Manager::getBool("clamp lighting", "Shaders"));
         resourceSystem->getSceneManager()->setForcePerPixelLighting(Settings::Manager::getBool("force per pixel lighting", "Shaders"));
+        resourceSystem->getSceneManager()->setAccurateFog(Settings::Manager::getBool("accurate fog", "Shaders"));
         resourceSystem->getSceneManager()->setAutoUseNormalMaps(Settings::Manager::getBool("auto use object normal maps", "Shaders"));
         resourceSystem->getSceneManager()->setNormalMapPattern(Settings::Manager::getString("normal map pattern", "Shaders"));
         resourceSystem->getSceneManager()->setNormalHeightMapPattern(Settings::Manager::getString("normal height map pattern", "Shaders"));
         resourceSystem->getSceneManager()->setAutoUseSpecularMaps(Settings::Manager::getBool("auto use object specular maps", "Shaders"));
         resourceSystem->getSceneManager()->setSpecularMapPattern(Settings::Manager::getString("specular map pattern", "Shaders"));
 
+        if (resourceSystem->getSceneManager()->getClampLighting())
```
Capostrophic: Won't this cause shaders to be used when the lighting *is* clamped?

sthalik:
I'm wondering what - if any - approach should be taken to fix the same behavior with view distance. The glsl `discard` statement could be used as the most lazy fix. Can we avoid geometry on the frustum's bounds not getting culled, despite being out of Euclidean bounds?

apps/openmw/mwrender/renderingmanager.cpp
AnyOldName3: While we're discussing this, we only globablly applied the setting for terrain. The documentation for the setting says
```ini
# Restrict the amount of lighting that an object can receive to a maximum of (1,1,1).
# Only affects objects that render with shaders (see 'force shaders' option). Always affects terrain.
# Setting this option to 'true' results in fixed-function compatible lighting, but the lighting
# may appear 'dull' and there might be color shifts.
# Setting this option to 'false' results in more realistic lighting.
clamp lighting = true
```

The reason it's applied to *terrain* specifically is because if one terrain chunk uses the shader as it's got a normal map, we need to ensure there're no lighting seams with adjacent terrain chunks. This isn't something that can happen with objects as they're distinct things, not multiple things pretending to be one, so I'm not 100% sure if it's a good idea to apply everywhere.

If we *do* decide this approach is better, we need to make sure the documentation is updated.

AnyOldName3:
There are a few possible approaches that I can remember (as I had a similar issue with shadow maps, although I couldn't use most options as I needed to avoid depth writes, too).

* **`discard`:** This upsets tiled rasterisers, but if we benchmark on a device with a tiled rasteriser and it makes no measurable difference, we might not need to care.
* **Setting the fragment alpha to zero, but not alpha testing or blending:** This will definitely do a depth write, which mattered for shadows, but doesn't here, but might well replace the already-rendered skybox with a fully transparent pixel, which would be bad.
* **Setting the fragment alpha to zero and using an alpha test:** I think I read that using the fixed-function-era alpha test had exactly the same drawbacks as a `discard` and was basically implemented the same way on modern GPUs, so if your research comes up with the same result, it's not a better option than the easier `discard`-based one.
* **Setting the fragment alpha to zero and using alpha blending:** The results from this should be fine and I think the performance impact is minimal, but do your own research. It does a depth write, which is why I couldn't use this for shadows, but provided you don't accidentally make OSG put everything in the depth-sorted transparent renderbin, it should be fine.
* **Moving the fragment depth so it's behind the far plane:** Just like `discard` and alpha testing, this kills the early depth test.
* **Moving the clip coordinates of vertices in the vertex shader (with or without also changing the vertex position):** this can't work as we want as triangles can straddle the distance threshold.

At the very least, this should tell you what to google. You might also find https://www.khronos.org/opengl/wiki/Early_Fragment_Test helps.

apps/openmw/mwrender/renderingmanager.cpp
sthalik: If a person has hardware capable of GLSL, why would they run it selectively on terrain only? openmw is bottlenecked on the CPU anyway, and the fixed pipeline can be slower since no one cares about its implementation anymore.

I recommend centralizing GPU-related options changing other options. That makes sure they're always at consistent values, in addition to being more readable overall.

AnyOldName3: I *think* we still technically support hardware that *supports* GLSL, but incredibly slowly. Even on modern hardware, fixed-function isn't always slower than the mish-mash of deprecated fixed-function and shader-based stuff we do - when vendors optimise their shader-based paths, I doubt they put much effort into making sure the optimisations work when deprecated behaviour is in use, but that deprecated behaviour was mainstream when they last touched the fixed-function stuff, so provided they're not doing anything to make it worse, we can be better off doing an old and deprecated normal thing than a half-and-half weird thing.

AnyOldName3:
> > This whole PR would be a lot cleaner if we had global shader defines. Maybe it's worth cherry picking the stuff that does that from the shadows PR into master and redoing this based on that. How do other people feel about that idea? Do we expect it to cause issues when the shadow PR is merged (I've barely ever cherry picked things, so don't know if it has risks).
> 
> I call this 'architectural runway' , it is work that is done for the support of a future feature but is useful for other features so they can be worked on in parallel.
> 
> It has the bonus effect of reducing the size of your shadow branch making it that much easier to review.
> 
> cherry-picking is just taking one commit and applying somewhere else, it will have a different commit id but the history is clean and you won't create merge conflicts with it when you go to merge your shadow branch in.
> 
> Make a new branch, cherry-pick the commits you want and create a new PR.

So it looks like the global defines implementation isn't suitable for cherry picking as it was written as an improvement to a stupid approach instead of its own thing based on master. The options are either to write it again, merge that, deal with the trivial merge conflict with the shadow PR and rewrite this PR now, or to keep this PR in its current form and tidy up all the sort-of-global shader defines in one batch after shadows are merged. The latter is now seeming like less work.

sthalik:
Is [this stackoverflow answer](https://stackoverflow.com/questions/19125628/how-does-gl-clipvertex-work-relative-to-gl-clipdistance) relevant with regard to the terrain view distance effect?

The fragment about the requirement for a linear transformation left me puzzled though. Linear in regard to what? Position with regard to camera after perspective transform?

AnyOldName3:
In this context, I'm pretty sure it just means that the clip planes and clip vertex need to be specified in regular cartesian coordinates instead of perspective-transform-compatible homogeneous ones.

Regardless, you can't do what you want to with `gl_CipVertex` and clipping planes. For a start, you're not aiming to clip against a plane, but instead a sphere centred around the camera. Also, like I said a couple of posts ago:

> * **Moving the clip coordinates of vertices in the vertex shader (with or without also changing the vertex position):** this can't work as we want as triangles can straddle the distance threshold.

sthalik:
Given we have all the data in the three shader types, I recommend going for the simple solution and spending more effort only if it turns out to be necessary. It should be guarded behind a .cfg option. I'm going to send a PR once the fog issue's resolved. From my tests and hardware configuration, openmw remains heavily CPU-starved and actual rendering takes a mere fraction of frame time.

Speaking of the CPU bottleneck, are there some clean ways to deserialize more steps, without introducing intrusive fine-grained locking? I'm going to look into cell preloading code. It causes frame drops despite having CPU threads available.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -97,6 +97,7 @@ namespace MWRender
             stateset->setAttribute(lightModel, osg::StateAttribute::ON);
             osg::Fog* fog = new osg::Fog;
             fog->setMode(osg::Fog::LINEAR);
+            fog->setUseRadialFog(true);
```
AnyOldName3: You should probably have this controlled by the setting we're using elsewhere. The extension is proprietary to Nvidia, so no one else can implement it even if they wanted to.

This does mean, though, that if `osg::isGLExtensionSupported(state.getContextID(),"GL_NV_fog_distance");` returns true, you don't actually have to force-enable shaders when the setting is on.

sthalik: > The extension is proprietary to Nvidia, so no one else can implement it even if they wanted to.

That isn't true, Mesa implements [[1]](https://www.mail-archive.com/mesa-dev@lists.freedesktop.org/msg11988.html)[[2]](https://cgit.freedesktop.org/mesa/mesa/log/?qt=grep&q=NV_fog_distance) this setting for all Gallium drivers. See also [Apple's data](http://arek.bdmonkeys.net/GLInfo/GLInfo_1068.html) for their proprietary Radeon driver.

Which setting do you have in mind? Should it be "accurate fog" as in this PR?

AnyOldName3: > > The extension is proprietary to Nvidia, so no one else can implement it even if they wanted to.
> 
> That isn't true, Mesa implements [[1]](https://www.mail-archive.com/mesa-dev@lists.freedesktop.org/msg11988.html)[[2]](https://cgit.freedesktop.org/mesa/mesa/log/?qt=grep&q=NV_fog_distance) this setting for all Gallium drivers. See also [Apple's data](http://arek.bdmonkeys.net/GLInfo/GLInfo_1068.html) for their proprietary Radeon driver.

Usually when an extension says it's proprietary to one vendor in its text, other vendors have to pay a fee to implement it. I imagine Apple paid to be allowed to do it, and have no idea how Mesa is allowed (maybe Nvidia granted them blanket permissions to do whatever they want, or maybe Nvidia just doesn't sue them). I hadn't seen some of that when I wrote my post, though, so assumed no one had bothered to get permission. It's definitely not supported by AMD hardware on Windows, though. I tested that myself.

> Which setting do you have in mind? Should it be "accurate fog" as in this PR?

Yeah, it should come under the same setting

psi29a: > Regardless of the vendor prefix, vendor-specific OpenGL extensions may be implemented an/or licensed by multiple hardware vendors to ensure compatibility. This is very rare in practice.

For example: GL_NV_texture_barrier (implemented widely on AMD hardware)

https://mesamatrix.net/ <-- complete coverage of GL extensions for mesa supported hardware

psi29a: Did more research, here is the result:

> Name NV_fog_distance
> Name Strings GL_NV_fog_distance
> Notice Copyright NVIDIA Corporation, 1999, 2000, 2001.
> IP Status  NVIDIA Proprietary.
> Status  Shipping (version 1.0)
> Version NVIDIA Date: January 18, 2001
>  $Id: //sw/main/docs/OpenGL/specs/GL_NV_fog_distance.txt#14 $ 

http://developer.download.nvidia.com/opengl/specs/nvOpenGLspecs.pdf

However, it was implimented for Mesa in 2011 here:
https://lists.freedesktop.org/archives/mesa-commit/2011-November/033899.html
discussion about it:
https://www.mail-archive.com/mesa-dev@lists.freedesktop.org/msg11929.html

It is like I said above, their way of implimenting it is NVIDIA Proprietary, but if someone implements it themselves without looking at how NVIDIA does it, then it's OK. Just like S2TC.

sthalik:
@AnyOldName3 the code compiles vertex/frag shaders and checks for the return value. But is there a possibility to check for *linking* error code? I can't find it in OSG.

In case of shader link failure and shaders enabled we should probably either silently disable them, display an error message, halt and catch fire, etc. Right now graphics look half-broken when one of the glsl files has an error.

AnyOldName3:
I'm pretty sure OSG prints the error in situations where there's a linking error the first time the program is used. The code I found looks like this:

```C++
if( ! _isLinked )
    {
        OSG_NOTICE << "glLinkProgram "<<this<<"\""<< _program->getName() << "\" FAILED" << std::endl;

        std::string infoLog;
        if( getInfoLog(infoLog) )
        {
            OSG_NOTICE << "Program \""<< _program->getName() << "\" "
                                      "infolog:\n" << infoLog << std::endl;
        }

        return;
    }
```

AnyOldName3:
@sthalik Regarding the issue of a euclidean distance far plane, I've had a better idea than any of the existing ideas in my list of ideas. If we were to just render a transparent sphere onto the main framebuffer in between the skybox and the rest of the scene in such a way that the depth writes still happen, then nothing will get drawn behind the sphere. As the sphere's only one object, there wouldn't be a major drawback to making the sphere really just a fullscreen quad and setting the fragment depth in the shader rather than using a sphere mesh, and then the surface would be perfectly smooth instead of having noticeable polygon edges.

sthalik:
> If we were to just render a transparent sphere onto the main framebuffer in between the skybox and the rest of the scene in such a way that the depth writes still happen, then nothing will get drawn behind the sphere.

That's a good idea. I'm not aware of the GL specifics yet (alpha test vs blend), but can you point me to some OSG class to start researching that?

AnyOldName3:
The difference between alpha test and alpha blend is that alpha test is effectively the fixed-function equivalent of `discard`ing a fragment if the alpha is below some threshold, and an alpha blend blends the colour of the current fragment into the existing framebuffer colour. Alpha testing is deprecated (but still works with shaders despite being fixed-function behaviour with a GLSL equivalent) and has all of the problems discard does, such as disabling early depth tests and doesn't do a depth write. Alpha blending always does a depth write unless you turn off depth writes completely, so isn't suitable for shadow maps, but is perfect here as a completely transparent fragment still can't be drawn behind.

To get this working in OSG, you'd need to add a new Geometry which held a plane, and make it a child of the scene root (or some other node within the scene). You'd set the renderbin by calling `setRenderBinDetails(<some number that comes just after whatever number the skybox has>, "RenderBin")` on its StateSet. You'd enable alpha blending by adding an appropriate StateAttribute to the StateSet, and you'd enable a shader program by adding that as another StateAttribute. To generate the program in the first place, you'd either ask OpenMW's shader manager for one (which would load the shader from `resources/shaders`) or you'd have the shader source inline as a string and pass the string directly to OSG's Shader constructor and add those to an OSG Program.

The vertex shader would need to ensure `gl_Position` gets set to the four corners of homogeneous clip space and set a `varying` to the equivalent view space x and y position. The fragment shader would read the `varying` and then calculate the maximum depth buffer value we want to render to in that pixel, and put the value in `gl_FragDepth` and set `gl_FragData` to `vec4(0.0)`.

kcat:
> You'd enable alpha blending by adding an appropriate StateAttribute to the StateSet, and you'd enable a shader program by adding that as another StateAttribute.

You can also disable color writes. I don't recall exactly how to do this with OSG, though OpenGL would use `glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE);`. The color buffer will then not be modified during subsequent rendering, though the depth buffer would be as appropriate.

AnyOldName3:
I didn't know that was a thing, although I did know about there being other masks, so it's probably something I should have thought of. It's just a StateAttribute in OSG, so it's done just like how alpha blending or a shader program would be.

sthalik:
> OSG prints the error in situations where there's a linking error the first time the program is used. 

It's unfortunate that we can't reliably detect program link failures. In some cases the program was unable to be used but I don't know whether it's mwrender not checking for program availability or did it only happen at link time.

@AnyOldName3, @kcat: can we finish reviewing and merging this and then work on a complete solution? Obviously I have an interest in view distance working right, fully. But that involves some reading (yuck!) and further discussion.

AnyOldName3:
> > OSG prints the error in situations where there's a linking error the first time the program is used.
> 
> It's unfortunate that we can't reliably detect program link failures. In some cases the program was unable to be used but I don't know whether it's mwrender not checking for program availability or did it only happen at link time.

I'm pretty sure any checks we do to check if the shader is available trigger the code that prints any link errors, too. I've definitely seen OpenMW print link errors when I've broken the shaders.

Regarding the mergeability of this, there are still a few problems, including, but not necessarily limited to:
* On hardware which doesn't support the radial fog extension, you should force shaders when the setting is enabled.
* You're forcing shaders everywhere when clamp lighting isn't enabled, but that's only necessary for terrain.

AnyOldName3:
I don't think https://github.com/OpenMW/openmw/pull/2042/commits/cc87e2a4b9e5fff55b1d1c9bc13cf6cc482c0a3e is a good thing to include in this PR, and am also unconvinced that it's the best solution to the problem it causes either (as once shadows get merged and the global defines system becomes available, it'll need gutting as it'll be completely unnecessary).

Regarding the radial view distance implementation you've got here, as I've said elsewhere, it's worse in every way to the solution I recommended, including in terms of actual implementation difficulty, so you've done a chunk of work on something we couldn't merge when doing it properly would have been easier. It might be a better idea to do the radial view distance stuff as a separate PR so we can get the fog stuff (which is probably fine now, but difficult to review as there are a zillion other changes included with it now) merged.

Once those two issues are resolved, I might have some other comments, but the gist of things here should be fine.

sthalik:
> I don't think cc87e2a is a good thing to include in this PR, and am also unconvinced that it's the best solution to the problem it causes either (as once shadows get merged and the global defines system becomes available, it'll need gutting as it'll be completely unnecessary).

Fair enough. I'm also waiting for a solution to the accessor hell.

> Regarding the radial view distance implementation you've got here, as I've said elsewhere, it's worse in every way to the solution I recommended

I agree with your here. For now I don't have the time to learn enough about GL and OSG to implement your solution.

Once view distance is done, it should implicitly implement radial fog as well, correct?

AnyOldName3:
> Once view distance is done, it should implicitly implement radial fog as well, correct?

Nope. Radial fog is always going to be its own thing.

> For now I don't have the time to learn enough about GL and OSG to implement your solution.

I'm pretty sure you've already done more than it would take - it should be about ten lines to add one plane to the scene graph, give it a custom shader, disable colour writes and assign it a certain renderbin.

sthalik:
Are we fine now after the last rebase? I removed all refactor-related commits. Do you have more reservations?

components/terrain/chunkmanager.cpp
```diff
@@ -220,8 +220,14 @@ osg::ref_ptr<osg::Node> ChunkManager::createChunk(float chunkSize, const osg::Ve
...
-        geometry->setPasses(::Terrain::createPasses(mSceneManager->getForceShaders() || !mSceneManager->getClampLighting(), mSceneManager->getForcePerPixelLighting(),
-                                                    mSceneManager->getClampLighting(), &mSceneManager->getShaderManager(), std::vector<TextureLayer>(1, layer), std::vector<osg::ref_ptr<osg::Texture2D> >(), 1.f, 1.f));
+        geometry->setPasses(::Terrain::createPasses(mSceneManager->getForceShaders(),
+                                                    mSceneManager->getForcePerPixelLighting(),
+                                                    mSceneManager->getClampLighting(),
+                                                    mSceneManager->getAccurateFog(),
+                                                    &mSceneManager->getShaderManager(),
+                                                    std::vector<TextureLayer>(1, layer),
+                                                    std::vector<osg::ref_ptr<osg::Texture2D>>(),
+                                                    1.f, 1.f);
```
AnyOldName3: The extra line breaks make it harder to see what was changed here, but I see that `mSceneManager->getForceShaders() || !mSceneManager->getClampLighting()` got changed to `mSceneManager->getForceShaders()` which I suspect isn't right.

sthalik: Good catch!

Just to be clear, I'm only reformatting due to some lines in excess of 200 columns.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -222,13 +235,26 @@ namespace MWRender
...
+        {
+            static bool ext = osg::isGLExtensionSupported(0, "GL_NV_fog_distance");
+
+            // force shader-based fog if lacking ffp extension
+            if (!ext)
+                resourceSystem->getSceneManager()->setForceShaders(true);
+        }
+
         osg::ref_ptr<SceneUtil::LightManager> sceneRoot = new SceneUtil::LightManager;
+
```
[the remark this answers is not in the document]
sthalik: Yes!

components/terrain/chunkmanager.cpp
AnyOldName3: Some of us have wide monitors, so lines of 200 characters aren't always unreadable, and as word wrap is a thing, it's not the biggest concern even if there are. Sometimes, splitting is nice, though, but splitting for every individual argument can be excessive.

kcat: Word wrap doesn't normally produce good results. It helps with horizontal scrolling, but messes up formatting and layout cleanliness. Personally, I avoid lines over 100 wide, and use indented line contiunations. e.g.
```
geometry->setPasses(::Terrain::createPasses(mSceneManager->getForceShaders(),
    mSceneManager->getForcePerPixelLighting(), mSceneManager->getClampLighting(),
    mSceneManager->getAccurateFog(), &mSceneManager->getShaderManager(),
    std::vector<TextureLayer>(1, layer), std::vector<osg::ref_ptr<osg::Texture2D>>(),
    1.f, 1.f
));
```
Makes it easy to see without horizontal scrolling or excessive bunching up on the right side, and can easily find where a call's parameter list ends.

sthalik: In an IDE, I have always-visible project filesystem structure and file's functions listed, taking close to half of the screen. A terminal takes about a fifth of the screen to aid multitasking. Given a 4K display, these environments have about ~140 columns each for text editing.

In any case I reformatted only the pieces of code that were touched for other reasons.

Capostrophic:
I'm testing this on an Nvidia card (Fedora 29, 415.22 driver version) and while the fog looks fine on fixed-function pipeline (I understand that far objects are not in the scope of this PR),
screenshot028

this view with shaders forced doesn't look quite right. The positions are different, but the issue is visible.
screenshot029

When accurate fog option is disabled, objects are properly covered by fog again.

I believe I've pulled the latest revision of the PR branch.
The issue doesn't appear to be related to per-pixel lighting and lighting (un)clamping.

OSG extension-dependent radial fog and shaders may conflict somehow.

sthalik:
@Capostrophic radial fog and other ffp stuff is completely off when shaders are used, regardless of whether they're being set or not.

Is it that some objects aren't covered with fog while object with similar distance are? Can you compare without this patchset applied?

Capostrophic:
>Can you compare without this patchset applied?
No, it's not present in master if you're asking about that.

AnyOldName3:
> > Can you compare without this patchset applied?
> 
> No, it's not present in master if you're asking about that.

I think he's asking for an identical screenshot from master just for comparison's sake.

AnyOldName3:
> radial fog and other ffp stuff is completely off when shaders are used, regardless of whether they're being set or not.

If the extension is really old and no longer used, it's possible it's not been considered when other things got updated, so it's triggering weird driver bugs. It's not especially likely, but it's possible radial fog is causing wrong things to be sent to the shader.

Capostrophic:
Looks like this on master - same settings but accurate fog option obviously not present.
screenshot030

AnyOldName3:
It would be much easier to make comparisons if the screenshots were all of the same thing.

When I'm doing comparisons for shadows, I make a save, then load it without touching the mouse for each build I want to test.

Capostrophic:
>I make a save

Well, you can't do that on a clean branch without removing the rogue two lines. :P

AnyOldName3:
Fair point.

sthalik:
> Looks like it's broken on objects with alpha-blended parts.

What happens when fog is to be drawn with alpha-blended parts? Are the objects not foggy? Are the objects foggy in a wrong way?

>> I make a save

> Well, you can't do that on a clean branch without removing the rogue two lines. :P

Can you elaborate? I'm ignorant.

wareya:
I was referring to capostrophic's screenshot, at "this view with shaders forced doesn't look quite right".

Capostrophic:
> Can you elaborate? I'm ignorant.

A recent commit that set data variance of source geometries for skinning to dynamic fixed and broke some stuff, specifically saving in exteriors. Not related to this PR.

sthalik:
@Capostrophic try it now. I was able to reproduce the rendering problem. There's an issue with passed positions for objects with bump mapping or possibly other transforms applied. After the last commit it works on my end.

Capostrophic:
Works fine now.

files/settings-default.cfg
```diff
@@ -108,6 +108,10 @@ distant interior fog start = 0
 
 distant interior fog end = 16384
 
+# When disabled, the visible distance is larger on screen edges. Enabling
+# the setting makes fog distance accurate all around the player.
```
AnyOldName3: Maybe this could be worded a little more clearly. It should also be added to the REST documentation.

files/shaders/objects_vertex.glsl
```diff
@@ -50,12 +50,16 @@ varying vec3 passNormal;
...
-    depth = gl_Position.z;
-
     vec4 viewPos = (gl_ModelViewMatrix * gl_Vertex);
     gl_ClipVertex = viewPos;
     vec3 viewNormal = normalize((gl_NormalMatrix * gl_Normal).xyz);
 
+#if !@accurateFog
+    depth = gl_Position.z;
+#else
+    depth = length(viewPos.xyz);
```
AnyOldName3: Are we 100% sure that `viewPos` will always be in Cartesian space, or might it sometimes be in homogeneous space? If so, `length(viewPos.xyz / viewPos.w);` would be preferable.

AnyOldName3:
@sthalik Shadows are merged now, so if you come back to this, you're probably going to want to make use of the new global shader defines system.

sthalik:
Good to know, thanks.

psi29a:
Looking forward to this. :)

sthalik:
@psi29a, @AnyOldName3: Unfortunately I won't be working on #2042, #2044, or #2072. Thanks for your time reviewing my pull requests over the time I was involved. Best luck with your project :+1:! -sh

AnyOldName3:
It's probably a good idea to keep these PRs open for now (just probably with a title that shows they've been abandoned). It'll be easier for someone else to pick them up if they're not lumped in with the rejected ones.

AnyOldName3:
@psi29a can you do that, please? I'm not at my computer.

Capostrophic:
Superseded by [MR 101](https://gitlab.com/OpenMW/openmw/merge_requests/101)

PR #2044, opened 2018-11-22
[WIP] fix lake floor refracting in shallow water

sthalik:
This prevents shallow ponds and lakes reflecting their floor upon water's surface, when being underwater. This happens when refraction is enabled.

Issue: https://gitlab.com/OpenMW/openmw/issues/4728

### DIFF
--- a/files/shaders/water_fragment.glsl
+++ b/files/shaders/water_fragment.glsl
@@ -28,6 +28,7 @@ const float BUMP = 0.5;                            // overall water surface bump
 const float BUMP_RAIN = 2.5;
 const float REFL_BUMP = 0.10;                      // reflection distortion amount
 const float REFR_BUMP = 0.07;                      // refraction distortion amount
+const float BUMP_UNDERWATER = 8;                   // distortion increase coefficient when underwater, >1
 
 const float SCATTER_AMOUNT = 0.3;                  // amount of sunlight scattering
 const vec3 SCATTER_COLOUR = vec3(0.0,1.0,0.95);    // colour of sunlight scattering
@@ -235,6 +236,7 @@ void main(void)
     float fresnel = fresnel_dielectric(vVec, normal, ior);
 
     fresnel = clamp(fresnel, 0.0, 1.0);
+    fresnel *= 1 - isUnderwater;
 
 #if REFRACTION
     float normalization = frustumDepth / 1000;
@@ -247,6 +249,7 @@ void main(void)
     float shore = 1.0;
 #endif
     vec2 screenCoordsOffset = normal.xy * REFL_BUMP * shore;
+    screenCoordsOffset *= max(1, isUnderwater * BUMP_UNDERWATER);
 
     // reflection
     vec3 reflection = texture2D(reflectionMap, screenCoords + screenCoordsOffset).rgb;


sthalik:
The exact coefficient I chose for reflection reduction is pretty arbitrary. We may want to remove reflection entirely while underwater, or keep some barely-visible reflection visible.

sthalik:
Good catch. I'm gonna play around with different values and rebase afterward.

psi29a:
If you're not ready, then mark this PR with [WIP] in the title please.

sthalik:
I'm having problems with this. Removing the fresnel coefficient entirely increases the water transparency. I counteract this by increasing the bump mapping amount while underwater.

The problem is mostly trivial to solve if not for shaders requiring minimal branching.

Despite the visible change (which may be a no-go from the start), I'd like you to comment on the modified effect. For me it increases the feeling of being underwater.

kcat:
I'm not really on board with changing the underwater reflections, unless it's correcting something mathematically or physically wrong. Neither the bug report or this really say what's wrong, other than you don't think it was intended. IIRC, the main issue is really the lack of physically-correct underwater fog/scattering, missing caustics, and other things. I think there were also issues with the reflections being flatter than intended. But the reflections themselves were modeled on realistic behavior, and I'd prefer to keep that as long as it's not egregiously incorrect.

wareya:
It negatively affects vanilla gameplay. Physical accuracy is basically
worthless unless you're designing a game from scratch.

sthalik:
See https://gitlab.com/OpenMW/openmw/issues/4728 for an example image. It might take a while to load. Basically the water reflects/refracts upside down.

I'm going to spend more time on this and try to reproduce the original look without the downsides.

kcat:
As I said,

> IIRC, the main issue is really the lack of physically-correct underwater fog/scattering ... I think there were also issues with the reflections being flatter than intended.

sthalik:
We're long way before physically-accurate rendering. openmw right now uses GL 2.x and has compatibility with non-programmable GPUs. The shaders can't just switch versions forward either as they use obsolete GL 2 constructs that don't work in newer versions.

The PR is about fixing the visual artifact.

kcat:
> The shaders can't just switch versions forward either as they use obsolete GL 2 constructs that don't work in newer versions.

What do you mean?

> The PR is about fixing the visual artifact.

You asked for feedback on fixing (getting rid of) underwater reflections, I said they were intended and I'd prefer to keep them. If there's an question on purity, we already have the option of disabling water shaders to get the vanilla look back, and if you want the vanilla shader look, an option for them would be better than messing up the current shader's intended look, IMO.

sthalik:
>>    The shaders can't just switch versions forward either as they use obsolete GL 2 constructs that don't work in newer versions.
> What do you mean?

We're limited to what GL 2 can do, both in OSG and GLSL. In the latter, changing the `#version` pragma forward just breaks compilation.

> You asked for feedback on fixing (getting rid of) underwater reflections, I said they were intended and I'd prefer to keep them.

This isn't the final version. I'm going to arrive at something more tasteful, but *eventually*. The bump mapping effect is totally exaggerated.

AnyOldName3:
I just had a lengthy swim in Balmora's canal and couldn't find anything that looked wrong. Do you have a savegame exhibiting the issue you're trying to fix as soon as you load it?

wareya:
Excuse me for the incoming rant. In Vanilla you can't see through the water very well in EITHER direction. In OpenMW you can see through it easily but only from the top. The problem is ultimately that it's a lot harder to see through water from the bottom than the top. This is extremely weird and it makes people question the engine and notice weird things like underwater reflections. **There is a reason people keep complaining about it.**

Do you want it to be easy to see through the water from the top? Make it easy to see through the water from the bottom too. Do you want it to be hard to see through the water from the bottom? Then make it easy to see through it from the top.

Also, in vanilla, you can TECHNICALLY see through the bottom, due to the lack of reflections or distortion. It's just very hard. You can just barely make out a faint silhouette, assuming there's enough contrast on the other side. This is how the water shader reflecting things on the bottom "affects vanilla gameplay". If those faint sillhouettes are there but covered up by foggy reflections it would be way harder to see them no matter what.

gimp-2 10_2018-11-22_22-50-13
gimp-2 10_2018-11-22_22-52-13

(Of course, if the water shader is off, then it barely effects gameplay at all because you can just see through it. In both engines.)

AnyOldName3:
But it's a lot easier to see into water than out of it. Have you never been swimming?

If in vanilla, depending on shader settings, there's both the option of seeing through and not seeing through water, we should be able to take that to mean that seeing through water or not is not a major issue and should take a back seat to pretty graphics.

wareya:
Yes, I've gone swimming. As a kid I went swimming a lot. Thanks for asking.

Different people keep popping up out of nowhere and complaining about the underwater reflections looking weird/uncanny/wrong, and the author of this pull request even made an effort to make it look better.

In my opinion, people keep complaining about the underwater reflections because they stand out against the hazy opaque water surface outside of the tiny cone where you can see through the surface, which would start being a lot less of a problem if permissiveness above and below the water was more similar. I'm pretty sure that lines of visibility between two points should be consistent in both directions.

I'm sure there's some kind of reason for permissiveness being so much similar from above and below,but I think it's uncanny and probably makes the underwater reflections stand out too much.

This should make the uncanniness apparent:
https://user-images.githubusercontent.com/585488/48948319-0ead1f80-ef02-11e8-9228-725d2f7f5462.png
https://user-images.githubusercontent.com/585488/48948326-110f7980-ef02-11e8-8e49-7826478b5214.png

Bonus images:

Refraction disabled:
https://user-images.githubusercontent.com/585488/48947233-6f3a5d80-eefe-11e8-8fa1-7cefe09fda67.png
https://user-images.githubusercontent.com/585488/48947241-72cde480-eefe-11e8-9979-58d8aa314856.png

Refraction enabled:
https://user-images.githubusercontent.com/585488/48947247-77929880-eefe-11e8-9ab1-0dce0660858c.png
https://user-images.githubusercontent.com/585488/48947252-7a8d8900-eefe-11e8-90ae-ca99e824b56e.png

kcat:
> Different people keep popping up out of nowhere and complaining about the underwater reflections looking weird/uncanny/wrong

Be careful about that. For every person that pops up to complain about the reflections, how many are silently happy with them? You can't just use the number of complaints to determine what people at large think, you need to gather that information from a sizable portion of people while correcting for any sampling bias. GZDoom had similar issues recently, where if you went just by the number of people complaining about things the engine wanted to do, you'd think everyone was running on Windows XP 32-bit with a GL2 graphics card. But upon collecting information from a larger less biased sampling pool, the vast majority were perfectly fine.

> and the author of this pull request even made an effort to make it look better.

Better is subjective. Personally, I'd prefer to make the water look more physically accurate, but in lieu of that I'd rather keep them as they are than to move away from physical accuracy. If accuracy to vanilla is your aim, then as I said, turning off shaders gets you accurate vanilla water, and for vanilla shaders, a separate shader pack more closely mimicking vanilla's would probably be preferable.

sthalik:
My motivation for the PR is the floor being visible clearly upward. The rotation map ought to be either darkened or distorted in some other way.

AnyOldName3:
You've still not shared a save file for a situation where as soon as you load the save, it's immediately broken. I've still not seen anything that I'm convinced looks at all weird or wrong (with the exception that I'm not seeing the whole 180° FOV through Snell's window, just more-or-less whatever's directly in the direction I'm looking).

sthalik:
@AnyOldName3 give me a few days to formulate a proper response, along with a save file. Thanks.

wareya:
The shader's Snell's Window is just a blending thing.

i30817:
My god. The first thing i will disable now and forevermore is the shader. Worrying about 'snell window high accuracy' to the point you forget that most engines don't model such phenomena so it looks alien and inconsistent as hell in what's supposed to be a game for gamers who never went diving in their lives much less care about the finer details of a phenomenon that only happens in very specific circumstances.

Maybe take a step back sometimes and think about 'how much physical modelling is too muc'.

sthalik:
What is the physical behavior here? Do we see reflections of shallow water's bottom surface at water's top while submerged? Sadly I have no bathtub here. 😄

@AnyOldName3 WRT the save game, is it enough to tell you to go underwater in the Balmora's central canal? My game has quite a few mods and it probably won't work. Look up and down to see the reflection correspond to the bottom surface.

AnyOldName3:
I've been swimming in Balmora's canal, and I could see the bottom of the canal reflected in the top of the canal, except for a circle right above me where I could see out of the water (Snell's Window). This is roughly the correct behaviour, and while the reflection should probably be affected more by the waves on the water's surface and Snell's Window should show lensing, I don't think those are the changes this PR is trying to do.

sthalik:
@wareya probably knows best what to do about the lensing effect.

The reflection was what concerned me originally and that's still the case. Despite being an existing optics and physics phenomena, I think it has no place here. I understand I'm going on a limb here.

sthalik:
Please note, something starting with

```
fresnel *= 1 - .3 * isUnderwater;
```

makes the clear reflection less obnoxious.

PR #2045, opened 2018-11-23
[Testing needed] Fix regressions in the animation system

akortunov:
Fixes [bug #4729](https://gitlab.com/OpenMW/openmw/issues/4729).

Needs caruful testing since it can break something else.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -446,11 +446,10 @@ void CharacterController::refreshMovementAnims(const WeaponInfo* weap, Character
 
             if(!mAnimation->hasAnimation(movementAnimName))
             {
-                movemask = MWRender::Animation::BlendMask_LowerBody;
                 movementAnimName = movestate->groupname;
-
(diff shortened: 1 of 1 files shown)


zinnschlag:
@psi29a Sorry, not familiar enough with this part of OpenMW to contribute more than a basic sanity check on the code (nothing obvious wrong). This needs careful and preferably extensive testing, indeed. I suggest we wait a little longer and if no one else has anything to contribute we merge. That is still the best method to get more testing.

AnyOldName3:
IIRC @Capostrophic was the one who made the changes which caused the regression, so he's probably pretty familiar with this code. I've not meddled with anything here either, so, like Zini, the best I could do would be a sanity check.

akortunov:
> Oops, just noticed that this is meant of 0.45. That makes the lets merge it and see what breaks options less good.

Do not hurry with that. There is nothing gamebreaking in these issues.

Capostrophic:
In short, LGTM (for real this time).

psi29a:
This still needs to be cherry picked into openmw-45

PR #2049, opened 2018-12-01
Support bullet double precision

elsid:
To build and run game with or without double precision for bullet. Some tests is still failing, I will fix them.

### DIFF
--- a/apps/openmw/mwphysics/heightfield.hpp
+++ b/apps/openmw/mwphysics/heightfield.hpp
@@ -3,6 +3,10 @@
 
 #include <osg/ref_ptr>
 
+#include <LinearMath/btScalar.h>
+
+#include <vector>
+
 class btCollisionObject;
 class btHeightfieldTerrainShape;
 
@@ -27,6 +31,7 @@ namespace MWPhysics
         btHeightfieldTerrainShape* mShape;
         btCollisionObject* mCollisionObject;
         osg::ref_ptr<const osg::Object> mHoldObject;
+        std::vector<btScalar> mHeights;
 
         void operator=(const HeightField&);
         HeightField(const HeightField&);
--- a/apps/openmw_test_suite/detournavigator/recastmeshbuilder.cpp
+++ b/apps/openmw_test_suite/detournavigator/recastmeshbuilder.cpp
@@ -363,11 +363,11 @@ namespace
             AreaType_ground
         );
         const auto recastMesh = builder.create(mGeneration, mRevision);
-        EXPECT_EQ(recastMesh->getVertices(), std::vector<float>({
+        EXPECT_THAT(recastMesh->getVertices(), Pointwise(FloatNear(1e-5), std::vector<float>({
             1.41421353816986083984375, 0, 1.1920928955078125e-07,
             -1.41421353816986083984375, 0, -1.1920928955078125e-07,
             1.1920928955078125e-07, 0, -1.41421353816986083984375,
-        }));
+        })));
         EXPECT_EQ(recastMesh->getIndices(), std::vector<int>({0, 1, 2}));
         EXPECT_EQ(recastMesh->getAreaTypes(), std::vector<AreaType>({AreaType_ground}));
     }
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -393,9 +393,9 @@ if [ -z $SKIP_DOWNLOAD ]; then
 	fi
 
 	# Bullet
-	download "Bullet 2.86" \
-		"https://www.lysator.liu.se/~ace/OpenMW/deps/Bullet-2.86-msvc${MSVC_YEAR}-win${BITS}.7z" \
-		"Bullet-2.86-msvc${MSVC_YEAR}-win${BITS}.7z"
+	download "Bullet 2.87" \
+		"https://www.lysator.liu.se/~ace/OpenMW/deps/Bullet-2.87-msvc${MSVC_YEAR}-win${BITS}.7z" \
+		"Bullet-2.87-msvc${MSVC_YEAR}-win${BITS}.7z"
 
 	# FFmpeg
 	download "FFmpeg 3.2.4" \
@@ -533,15 +533,15 @@ fi
 cd $DEPS
 echo
 # Bullet
-printf "Bullet 2.86... "
+printf "Bullet 2.87... "
 {
 	cd $DEPS_INSTALL
 	if [ -d Bullet ]; then
 		printf -- "Exists. (No version checking) "
 	elif [ -z $SKIP_EXTRACT ]; then
 		rm -rf Bullet
-		eval 7z x -y "${DEPS}/Bullet-2.86-msvc${MSVC_YEAR}-win${BITS}.7z" $STRIP
-		mv "Bullet-2.86-msvc${MSVC_YEAR}-win${BITS}" Bullet
+		eval 7z x -y "${DEPS}/Bullet-2.87-msvc${MSVC_YEAR}-win${BITS}.7z" $STRIP
+		mv "Bullet-2.87-msvc${MSVC_YEAR}-win${BITS}" Bullet
 	fi
 	export BULLET_ROOT="$(real_pwd)/Bullet"
 	echo Done.
--- a/apps/openmw/mwphysics/heightfield.cpp
+++ b/apps/openmw/mwphysics/heightfield.cpp
@@ -7,12 +7,48 @@
 
 #include <LinearMath/btTransform.h>
 
+#include <type_traits>
+
+namespace
+{
+    template <class T>
+    auto makeHeights(const T* heights, float sqrtVerts)
+        -> std::enable_if_t<std::is_same<btScalar, T>::value, std::vector<btScalar>>
+    {
+        return {};
+    }
+
+    template <class T>
+    auto makeHeights(const T* heights, float sqrtVerts)
+        -> std::enable_if_t<!std::is_same<btScalar, T>::value, std::vector<btScalar>>
+    {
+        return std::vector<btScalar>(heights, heights + static_cast<std::ptrdiff_t>(sqrtVerts * sqrtVerts));
+    }
+
+    template <class T>
+    auto getHeights(const T* floatHeights, const std::vector<btScalar>&)
+        -> std::enable_if_t<std::is_same<btScalar, T>::value, const btScalar*>
+    {
+        return floatHeights;
+    }
+
+    template <class T>
+    auto getHeights(const T*, const std::vector<btScalar>& btScalarHeights)
+        -> std::enable_if_t<!std::is_same<btScalar, T>::value, const btScalar*>
+    {
+        return btScalarHeights.data();
+    }
+}
+
 namespace MWPhysics
 {
     HeightField::HeightField(const float* heights, int x, int y, float triSize, float sqrtVerts, float minH, float maxH, const osg::Object* holdObject)
+        : mHeights(makeHeights(heights, sqrtVerts))
     {
         mShape = new btHeightfieldTerrainShape(
-            sqrtVerts, sqrtVerts, heights, 1,
+            sqrtVerts, sqrtVerts,
+            getHeights(heights, mHeights),
+            1,
             minH, maxH, 2,
             PHY_FLOAT, false
         );
--- a/apps/openmw_test_suite/detournavigator/operators.hpp
+++ b/apps/openmw_test_suite/detournavigator/operators.hpp
@@ -20,22 +20,62 @@ namespace DetourNavigator
     }
 }
 
+namespace
+{
+    template <class T>
+    struct Wrapper {
+        const T& mValue;
+    };
+
+    template <class Range>
+    inline testing::Message& writeRange(testing::Message& message, const Range& range)
+    {
+        message << "{\n";
+        for (const auto& v : range)
+            message << Wrapper<typename std::decay<decltype(v)>::type> {v} << ",\n";
+        return message << "}";
+    }
+}
+
 namespace testing
 {
     template <>
+    inline testing::Message& Message::operator <<(const osg::Vec3f& value)
+    {
+        return (*this) << "osg::Vec3f(" << std::setprecision(std::numeric_limits<float>::max_exponent10) << value.x()
+            << ", " << std::setprecision(std::numeric_limits<float>::max_exponent10) << value.y()
+            << ", " << std::setprecision(std::numeric_limits<float>::max_exponent10) << value.z()
+            << ')';
+    }
+
+    template <>
+    inline testing::Message& Message::operator <<(const Wrapper<osg::Vec3f>& value)
+    {
+        return (*this) << value.mValue;
+    }
+
+    template <>
+    inline testing::Message& Message::operator <<(const Wrapper<float>& value)
+    {
+        return (*this) << std::setprecision(std::numeric_limits<float>::max_exponent10) << value.mValue;
+    }
+
+    template <>
     inline testing::Message& Message::operator <<(const std::deque<osg::Vec3f>& value)
     {
-        (*this) << "{\n";
-        for (const auto& v : value)
-        {
-            std::ostringstream stream;
-            stream << "osg::Vec3f("
-                   << std::setprecision(std::numeric_limits<float>::max_exponent10) << v.x() << ", "
-                   << std::setprecision(std::numeric_limits<float>::max_exponent10) << v.y() << ", "
-                   << std::setprecision(std::numeric_limits<float>::max_exponent10) << v.z() << ")";
-            (*this) << stream.str() << ",\n";
-        }
-        return (*this) << "}";
+        return writeRange(*this, value);
+    }
+
+    template <>
+    inline testing::Message& Message::operator <<(const std::vector<osg::Vec3f>& value)
+    {
+        return writeRange(*this, value);
+    }
+
+    template <>
+    inline testing::Message& Message::operator <<(const std::vector<float>& value)
+    {
+        return writeRange(*this, value);
     }
 }
 
--- a/apps/openmw_test_suite/nifloader/testbulletnifloader.cpp
+++ b/apps/openmw_test_suite/nifloader/testbulletnifloader.cpp
@@ -9,6 +9,8 @@
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
 
+#include <algorithm>
+
 namespace
 {
     template <class T>
@@ -29,6 +31,34 @@ namespace
         shape.getAabb(btTransform::getIdentity(), aabbMin, aabbMax);
         shape.processAllTriangles(&callback, aabbMin, aabbMax);
         return result;
+    }
+
+    bool isNear(btScalar lhs, btScalar rhs)
+    {
+        return std::abs(lhs - rhs) <= 1e-5;
+    }
+
+    bool isNear(const btVector3& lhs, const btVector3& rhs)
+    {
+        return std::equal(
+            static_cast<const btScalar*>(lhs),
+            static_cast<const btScalar*>(lhs) + 3,
+            static_cast<const btScalar*>(rhs),
+            [] (btScalar lhs, btScalar rhs) { return isNear(lhs, rhs); }
+        );
+    }
+
+    bool isNear(const btMatrix3x3& lhs, const btMatrix3x3& rhs)
+    {
+        for (int i = 0; i < 3; ++i)
+            if (!isNear(lhs[i], rhs[i]))
+                return false;
+        return true;
+    }
+
+    bool isNear(const btTransform& lhs, const btTransform& rhs)
+    {
+        return isNear(lhs.getOrigin(), rhs.getOrigin()) && isNear(lhs.getBasis(), rhs.getBasis());
     }
 }
 
@@ -157,7 +187,7 @@ static bool operator ==(const btCompoundShape& lhs, const btCompoundShape& rhs)
     for (int i = 0; i < lhs.getNumChildShapes(); ++i)
     {
         if (!compareObjects(lhs.getChildShape(i), rhs.getChildShape(i))
-                || !(lhs.getChildTransform(i) == rhs.getChildTransform(i)))
+                || !isNear(lhs.getChildTransform(i), rhs.getChildTransform(i)))
             return false;
     }
     return true;
@@ -165,13 +195,13 @@ static bool operator ==(const btCompoundShape& lhs, const btCompoundShape& rhs)
 
 static bool operator ==(const btBoxShape& lhs, const btBoxShape& rhs)
 {
-    return lhs.getLocalScaling() == rhs.getLocalScaling()
+    return isNear(lhs.getLocalScaling(), rhs.getLocalScaling())
         && lhs.getHalfExtentsWithoutMargin() == rhs.getHalfExtentsWithoutMargin();
 }
 
 static bool operator ==(const btBvhTriangleMeshShape& lhs, const btBvhTriangleMeshShape& rhs)
 {
-    return lhs.getLocalScaling() == rhs.getLocalScaling()
+    return isNear(lhs.getLocalScaling(), rhs.getLocalScaling())
         && lhs.usesQuantizedAabbCompression() == rhs.usesQuantizedAabbCompression()
         && lhs.getOwnsBvh() == rhs.getOwnsBvh()
         && getTriangles(lhs) == getTriangles(rhs);
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,8 +30,8 @@ configuration:
 clone_depth: 1
 
 cache:
-    - C:\projects\openmw\deps\Bullet-2.86-msvc2015-win32.7z
-    - C:\projects\openmw\deps\Bullet-2.86-msvc2015-win64.7z
+    - C:\projects\openmw\deps\Bullet-2.87-msvc2015-win32.7z
+    - C:\projects\openmw\deps\Bullet-2.87-msvc2015-win64.7z
     - C:\projects\openmw\deps\MyGUI-3.2.2-msvc2015-win32.7z
     - C:\projects\openmw\deps\MyGUI-3.2.2-msvc2015-win64.7z
     - C:\projects\openmw\deps\OSG-3.4.1-scrawl-msvc2015-win32.7z
--- a/components/detournavigator/recastmeshbuilder.cpp
+++ b/components/detournavigator/recastmeshbuilder.cpp
@@ -163,15 +163,15 @@ namespace DetourNavigator
 
         transformBoundingBox(transform, aabbMin, aabbMax);
 
-        aabbMin.setX(std::max(mBounds.mMin.x(), aabbMin.x()));
-        aabbMin.setX(std::min(mBounds.mMax.x(), aabbMin.x()));
-        aabbMin.setY(std::max(mBounds.mMin.y(), aabbMin.y()));
-        aabbMin.setY(std::min(mBounds.mMax.y(), aabbMin.y()));
+        aabbMin.setX(std::max(static_cast<btScalar>(mBounds.mMin.x()), aabbMin.x()));
+        aabbMin.setX(std::min(static_cast<btScalar>(mBounds.mMax.x()), aabbMin.x()));
+        aabbMin.setY(std::max(static_cast<btScalar>(mBounds.mMin.y()), aabbMin.y()));
+        aabbMin.setY(std::min(static_cast<btScalar>(mBounds.mMax.y()), aabbMin.y()));
 
-        aabbMax.setX(std::max(mBounds.mMin.x(), aabbMax.x()));
-        aabbMax.setX(std::min(mBounds.mMax.x(), aabbMax.x()));
-        aabbMax.setY(std::max(mBounds.mMin.y(), aabbMax.y()));
-        aabbMax.setY(std::min(mBounds.mMax.y(), aabbMax.y()));
+        aabbMax.setX(std::max(static_cast<btScalar>(mBounds.mMin.x()), aabbMax.x()));
+        aabbMax.setX(std::min(static_cast<btScalar>(mBounds.mMax.x()), aabbMax.x()));
+        aabbMax.setY(std::max(static_cast<btScalar>(mBounds.mMin.y()), aabbMax.y()));
+        aabbMax.setY(std::min(static_cast<btScalar>(mBounds.mMax.y()), aabbMax.y()));
 
         transformBoundingBox(transform.inverse(), aabbMin, aabbMax);
 


psi29a:
Is there a test to validate that the Bullet lib supports double precision? I can imagine that people thinking that have it installed, but don't, then turn on double precision and things don't work as expected.

elsid:
Bullet can't be build supporting both `double` and `float`. If we are talking about destribution for openmw user, we should provide both or one of versions.

This test implicitly is made by linker for openmw developers. Right now if you try to build openmw with wrong bullet (openmw uses `-DBT_USE_DOUBLE_PRECISION` but bullet doesn't and vice versa), linker produce an `undefined reference` error. It happens because openmw expects to have function with `double` parameter type overload using header declaration, but has only `float`.

psi29a:
I meant to test what is found by Cmake is indeed what we are asking for. No point in turning on double when the found lib and headers are float. Or is this irrelevant? 

Also... Keep in mind that scrawls osg on steroids uses floats instead of double for performance reasons.

psi29a:
Is this still WIP?

Anybody else have any issues with this?

AnyOldName3:
I thought the whole reason we were doing this was because single precision wasn't enough. Making it optional would be like adding an option called 'enable physics bugs'.

psi29a:
@AnyOldName3: this is exactly how I viewed it.

wareya:
> Obviously double precision float physics would have worse performance
Did you actually test this or are you just saying it because it seems like it makes sense? It probably won't have a measurable effect on modern desktop computers.

Capostrophic:
>are you just saying it because it seems like it makes sense

Yep. :smile:

If there aren't any noticeable problems with that then move along. I'm just worried about Android quasi-support.

psi29a:
plan of action: 
1) validate assumptions, test performance before and after this branch
2) should performance loss (if any) be acceptable, then this PR should drop the options and ifdefs and explicitly use double precision with the reason why in comments.

wareya:
Android is the real worry for performance here, yeah.

One thing maybe worth mentioning is that the primary reason to use double precision bullet in the first place is because single precision bullet's very bad at giving good results for collisions between hulls with sharp edges (i.e. anything other than spheres and capsules) and large meshes (like the ones in ebonheart and vivec), and double precision bullet improves that a lot. But IIRC, AABBs were faster than capsules when people tested my movement solver changes. So it might be a win-some-lose-some situation, not just a performance reduction.

psi29a:
> Android is the real worry for performance here, yeah.

Not worried about this, SoCs are just going to improve with time.

elsid:
> I meant to test what is found by Cmake is indeed what we are asking for.

I didn't find any.

> validate assumptions, test performance before and after this branch

I think we need a benchmark test for openmw engine. Load some save, take control over player, walk around with cell changing including exterior and interior, go into battle (probably something else). Check how much time it takes, what is fps, cpu, memory usage etc.

Fixed all tests. Ready to merge.

Capostrophic:
This is not something related to the PR, but @elsid, can you take a look at [this](https://gitlab.com/OpenMW/openmw/issues/4751) issue?

wareya:
By the way, there isn't a way to tell cmake "no, ONLY get double-precision bullet", is there? What if a packager screws up and releases a build that depends on single precision bullet without testing it? It uses the same headers.

psi29a:
@wareya I already asked:
>> I meant to test what is found by Cmake is indeed what we are asking for.

@elsid said:
>I didn't find any.

elsid:
> Please remove the option and just set things to double precision.

Ok, but need to change CI then.

> What if a packager screws up and releases a build that depends on single precision bullet without testing it?

Linker wouldn't make a binary for a static linkage with an `undefined reference` error. So packager just couldn't make a package. But for a dynamic linkage package will be fine but for a player game wouldn't start because of missing some symbols (all functions overloads with `btScalar` in arguments).

> Would this mean that builds would be 'broken' on 32-bits (not compilable) or would this mean that we just lose precision when converting between floats and doubles?

It will be broken if I will do [this](https://github.com/OpenMW/openmw/pull/2049/files#diff-af3b638bc2a3e6c650974192a53c7291R301) without condition (as I plan to).

psi29a:
OK... things are getting difficult again.

@elsid: hold off for now. This PR should be good to go as-is while stills supporting 32-bit builds.

@zinnschlag @AnyOldName3  @everyone_else: What should OpenMW's policy be about 32-bit? Should we keep 32-bit support or drop it all together?

Aside from armhf/armel, I don't see any new chips being put out for our 3 main target platforms that are purely 32-bit. aarch64 (arm64) is pretty much standard going forward for SoC (Android and RPi3)

AnyOldName3:
> What should OpenMW's policy be about 32-bit? Should we keep 32-bit support or drop it all together? I'm starting to see more problems than benefits to keeping 32-bit around.

It should be possible to build 32-bit stuff which uses double precision albeit (maybe) slowly - the floating-point ALU for most 32-bit processors should have a double-precision mode, even if actually copying data around would require two instructions rather than one, and where it doesn't, the compiler should just be making it use several instructions to simulate double precision.

If we're using static Bullet by default, it shouldn't matter that Debian is shipping it precompiled for single-precision on 32-bit machines, and if we're not, I don't see how we ever got things to work before this PR on 64-bit machines as I think most Bullet functions that take pointers take pointers to `btScalar` rather than having a version for `float` and a version for `double`.

psi29a:
Debian (and Ubuntu) builds and ships bullet differently depending on platform. 32-bit gets float and 64-bit gets double. Like I pointed out in the rule file above.

zinnschlag:
Regarding 32bit support: I am wondering if we have any statistics about that? Probably not on Linux. But are there any statistics about Windows/Mac downloads? I am generally inclined to drop 32bit builds. But if they are still in wider use, that would have a significant impact on our users.

psi29a:
According to some individuals we're only 32-bit. ;)

psi29a:
Just a brief/simple count of 2 weeks worth logs that I parsed from downloads.openmw.org, filtering out known 'bots':

> grep -ir 'win32' test  | wc -l
> 217

> grep -ir 'win64' test  | wc -l
> 576

kcat:
That may be unrepresentative if it's a period well after the last release. Most downloads occur soon after release, and blends in with background noise as time passes on. You also need to ask if people who downloaded the win32 version actually have a 32-bit OS, or if they're running the 32-bit version on a 64-bit OS (you would be surprised, sometimes..).

zinnschlag:
> I would rather have the 0.46 release cycle to have a poll where people could voice their support for and against to measure what people want as apposed to to download statistics.

+1

psi29a:
@AnyOldName3 Is this PR good to go; or do we not want to give people the option to build/link against incorrectly built bullet (one without double support)?

AnyOldName3:
I reckon it would probably be best to have double-precision bullet provided by the Windows prebuild script as soon as possible after this is merged, otherwise that's most Windows devs screwed for a while. The script is downloading Bullet from a location managed by @ananace, so he'd need to be involved.

psi29a:
@AnyOldName3 so put this PR on hold until we're sure that both mac and windows libs are validated to be built with double precision?

@ananace and @nikolaykasyanov : is bullet already built with double support, if not can we have new libs please? :)

ananace:
I've compiled up a new set of Bullet packages - built with both double precision as well as float. Only VS2015/2017 for now though, VS2013 is being irksome to reinstall.

If people want to test it before the merge;

http://www.lysator.liu.se/~ace/OpenMW/deps/Bullet-2.87-msvc2015-win32.7z
http://www.lysator.liu.se/~ace/OpenMW/deps/Bullet-2.87-msvc2015-win32_double.7z
http://www.lysator.liu.se/~ace/OpenMW/deps/Bullet-2.87-msvc2015-win64.7z
http://www.lysator.liu.se/~ace/OpenMW/deps/Bullet-2.87-msvc2015-win64_double.7z

elsid:
I've managed to build openmw with bullet double precision for windows https://ci.appveyor.com/project/elsid/openmw/builds/22909844 from different branch. Tested both Win32 and x64 (2015) versions and it seems to be working. And we still need packages for MacOS.

akortunov:
Ping @nikolaykasyanov here.
We need to test this PR on MacOS X with Bullet, built with double precision somehow.

nikolaykasyanov:
@akortunov on it.

nikolaykasyanov:
btw how are we going to ensure that both single and double precision configurations are building going forward? 2 x CI jobs?

nikolaykasyanov:
macOS build: https://www.dropbox.com/s/jblkratdi1hv2hm/OpenMW-bullet-double-precision.dmg?dl=1. Anything specific I should test?

@zinnschlag macOS build are 64 bit only.

akortunov:
So what shall we do here?

elsid:
Build and test openmw with double precision bullet on macos (windows and linux are tested). I'm not able to this testing. Then decide do we want to switch or add an option to use double precision. After that change CI scripts and package building for all os accordingly.

terabyte25:
Just wondering how this PR would work out in Android, as there many Android users are evenly divided between 32 bit and 64 bit.

akortunov:
Can we check Bullet precision during runtime (maybe by result of simple test with known result)?

kcat:
I doubt you can build against single-precision and run with double-precision (or vice-versa), given that it changes the structures and types used by various functions. You'd probably get an error saying some symbols/functions are missing when trying to run the executable.

akortunov:
> I doubt you can build against single-precision and run with double-precision (or vice-versa)

I do not expect it, I just want to get rid of additional configuration on the OpenMW side - maintainers may not know about this configuration switch.

terabyte25:
So for Android we should build 32 bit bullet for arm and x86, while for 64 bit we build 64 bit bullet for x86_64 and ARM64?

wareya:
If that's entirely asking about Android, sure. Just to make sure the situation is clear though, every desktop target should default to double-precision bullet. When it's supported properly by the hardware (i.e. not horrendously slow for no reason), there's no real reason *not* to use double precision collision math; it lets us use AABBs (which are faster than capsules) and it makes the movement solver less buggy-seeming (which means fewer collision iterations).

terabyte25:
So I build with 
```
USE_DOUBLE_PRECISION
```
for bullet and for openmw I build with
```
BULLET_USE_DOUBLE_PRECISION
```
?

terabyte25:
https://github.com/xyzz/openmw-android/pull/2 should add support for bullet's double precision on android.

AnyOldName3:
> So I build with
> 
> ```
> USE_DOUBLE_PRECISION
> ```
> for bullet and for openmw I build with
> 
> ```
> BULLET_USE_DOUBLE_PRECISION
> ```
> ?

Isn't there a system in CMake that when you build a library, it tags it with its exported CMake options, so anything built against it automatically inherits it? That should handle things on OpenMW's end (unless the option is there so that if you have both types of Bullet, it knows what to prefer).

akortunov:
> Isn't there a system in CMake that when you build a library, it tags it with its exported CMake options, so anything built against it automatically inherits it?

Strange. I can not find any info about it.

AnyOldName3:
It's apparently not CMake-specific, but it can do stuff with pkgconfig, which can be used to say which compiler settings something was built with, and I guess that includes stuff like the values of defines.

There's also the ability to add stuff to exported CMake targets, which everything *should* be doing, but effectively nothing actually *is* doing. https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/ If you set a define as an interface or public property on an exported target, it should get picked up by your dependencies. I know Boost does this kind of thing, and Recastnavigation at least mostly does, but I'm not sure if Bullet is in *enlightened modern CMake* mode or *everyone else's CMake is hot garbage and totally inconsistent, so lets make ours terrible, too, so it matches* mode.

akortunov:
Is there any progress here?

elsid:
> It's apparently not CMake-specific, but it can do stuff with pkgconfig, which can be used to say which compiler settings something was built with, and I guess that includes stuff like the values of defines.

Last [commit](https://github.com/OpenMW/openmw/pull/2049/commits/2414bed0b20edf1092655875bda23b96e969f788) adds check for bullet flags whether it has definition of double precision. It works on linux. All is requied bullet.pc file. For windows need to add pkg-config tool to CI and this file generation to bullet build. Apparently it is not supported in [latest](https://github.com/bulletphysics/bullet3/blob/25cc1fa386447771e3a19c1470578a995eb06f9e/CMakeLists.txt#L450-L457) bullet version at least for MSVC.

ananace:
On windows, pkg-config is only supported for when doing MinGW builds / cross-compiles. CMake won't even attempt to use it when generating a Visual Studio solution / MSVC / NMake build files.

akortunov:
So the whole point of this PR is just to define different height arrays when Bullet has double precision and when it has not?

If we do not have a cross-platform way to detect configuration, cam we run a runtime test to determine if we have a double precision or if we have not?

AnyOldName3:
We could use templates depending on `sizeof(btScalar)` and some SFINAE/`std::enable_if` stuff to make the extra stuff only appear when we need it. Depending on how we do it, we might need a bigger C++ version.

I'd rather we didn't have a runtime check as it's something we know at build time.

In fact, surely the headers shipped with a double-precision Bullet build should have a define in them for whether double-precision was enabled which we could use?

wareya:
>In fact, surely the headers shipped with a double-precision Bullet build should have a define in them for whether double-precision was enabled which we could use?

Other way around. BT_USE_DOUBLE_PRECISION must be defined during the build process (or before including bullet headers) if you are linking against double-precision bullet. Yes, this is super fishy.

AnyOldName3:
Have we complained to Bullet that that's stupid?

I also think CMake should start doing something like pkgconfig but that does CMake options and works on Windows. Their modern *everything works with targets* thing works really well when you're building your dependencies at the same time like we do with RecastNavigation, but with prebuilt libraries, all of those exported target properties just evaporate into the ether.

ananace:
You can still use CMake package config files on Windows, though without as nice automatic discovery support as what you'd get on Linux, as Windows doesn't have anything like a default path for things such as that.

(With CMake package config files I mean things like `/usr/lib/cmake/SDL2/sdl2-config.cmake`)
```cmake
# sdl2 cmake project-config input for ./configure scripts

set(prefix "/usr") 
set(exec_prefix "${prefix}")
set(libdir "/usr/lib64")
set(SDL2_PREFIX "/usr")
set(SDL2_EXEC_PREFIX "/usr")
set(SDL2_LIBDIR "/usr/lib64")
set(SDL2_INCLUDE_DIRS "${prefix}/include/SDL2")
set(SDL2_LIBRARIES "-L${SDL2_LIBDIR}  -lSDL2")
string(STRIP "${SDL2_LIBRARIES}" SDL2_LIBRARIES)
```

AnyOldName3:
They're not generated automatically when configuring, generating and installing a library, though, are they?

ananace:
Depends entirely on the library in question, all of the KDE Frameworks libraries for instance generate and install config files. [SFML](https://github.com/SFML/SFML/blob/master/cmake/SFMLConfig.cmake.in) also does this.
But yes, it is something that has to be explicitly done, it doesn't happen automatically - just like it doesn't with pkg-config.

elsid:
> So the whole point of this PR is just to define different height arrays when Bullet has double precision and when it has not?

The whole point is to make it possible to build openmw with double precision Bullet. With current master it is not possible without changes.

> If we do not have a cross-platform way to detect configuration, cam we run a runtime test to determine if we have a double precision or if we have not?

What kind of runtime check do you suggest?

Why do we need any additional check? Do you want to publish 2x versions of openmw with and without double precision? If double precision solves some physics related problem then it's just useless to have non double precision version. If we don't have any problems with physics without double precision then let's drop this pr.

AnyOldName3:
Double precision does solve or mitigate some issues, but we can't drop single-precision easily because lots of prebuilt bullet packages conflate 64-bit and double precision, so only provide x86 builds with single precision. We'd have to start providing our own bullet packages all over the place if we dropped single-precision support.

wareya:
Yet if we want to take advantage of double precision bullet's better accuracy, OpenMW built against those single-precision packages will be extremely glitchy, to the point of it being difficult to just walk around. It's not that simple of a decision to make.

akortunov:
> What kind of runtime check do you suggest?

Run a some kind of test with known result (result should be different for single and double precisions).

> Why do we need any additional check? 

Because we can not drop single-precision library support, since bullet packages in many distros use single precision by default.
So we will have to support both single and double precisions and we should somehow distingush them.

elsid:
> Because we can not drop single-precision library support, since bullet packages in many distros use single precision by default.
> So we will have to support both single and double precision and we should somehow distinguish them.

Here is how I see it.
1. We build and distribute 2 versions of openmw engine binary and all binaries that directly depending on Bullet. Each one is dynamically linked against corresponding Bullet precision version single or double.
2. Launcher does either of:
    2.1. Launcher looks for installed bullet package or files. If only a single precision exists then launcher runs single precision openwm binary, if there is only double then double, if there are both then makes a choice between preferring double by default. Probably double is more preferred if otherwise is not set by a user in the launcher settings.
    2.2. Launcher simply tries to launch both openmw binary with a specific flag to just exit from main. OS will allow to run that binary which will find all dynamically linked libraries. The same as in 2.1.: only single, only double, both -> preference or user choice.

I'd prefer 2.2 because we delegate all the work for finding and checking bullet libraries to OS that knows better anyway how to do it.

Why do we need 2 binaries?
There is no way AFAIK for C++ to be able make a choice in runtime without explicitly putting it in the code before what function to call `void f(float)` or `void f(double)`. If a translation unit is compiled including only one of them, there is no way to decide to call other version. When OS runs binary and there is no corresponding symbol in all linked libraries the binary will not run. And symbols in C++ for different overloads of functions are different. And if translation unit includes both overloads there is still a problem because OS will try to find all of them. If anyone else knows how we can avoid having two openmw binaries and can present a working solution I'd glad to see.

apps/openmw/mwphysics/heightfield.hpp
```diff
@@ -27,6 +32,9 @@ namespace MWPhysics
         btHeightfieldTerrainShape* mShape;
         btCollisionObject* mCollisionObject;
         osg::ref_ptr<const osg::Object> mHoldObject;
+#ifdef BT_USE_DOUBLE_PRECISION
+        std::vector<btScalar> mHeights;
```
akortunov: BTW, is it possible to use the `mHeights` in both precisions to avoid conditional compilation?

elsid: Yes, it's possible.

akortunov: Are there drawbacks here? Maybe it would be better indeed to use the BT_USE_DOUBLE_PRECISION only for unit tests.

elsid: Unnecessary copying of heights data for single precision.

akortunov:
> You can still use CMake package config files on Windows, though without as nice automatic discovery support as what you'd get on Linux, as Windows doesn't have anything like a default path for things such as that.

So the only issue left is the config autodection for Windows? In theory, usually on Windows developers use dependency libraries which we provide instead of using system-wide libraries as on Linux.

So on Windows we can use the `-DBT_USE_DOUBLE_PRECISION` do determine which Bullet libraries to download and use.

apps/openmw/mwphysics/heightfield.hpp
elsid: Or we can use SFINAE to avoid copying. But not sure is it better than conditional compilation here.

apps/openmw_test_suite/detournavigator/recastmeshbuilder.cpp
```diff
@@ -310,11 +310,19 @@ namespace
             AreaType_ground
         );
         const auto recastMesh = builder.create();
+#ifdef BT_USE_DOUBLE_PRECISION
+        EXPECT_EQ(recastMesh->getVertices(), std::vector<float>({
```
akortunov: Maybe it is worth to use the `EXPECT_NEAR` with something like the 1e-06 epsilon instead. In this case we do not need to use additional CMake flags at all.

elsid: `EXPECT_NEAR ` doesn't support containers, but `EXPECT_THAT` with `FloatNear` can do this. CMake flags and defines are still required because bullet headers don't know about used precision.

akortunov: > CMake flags and defines are still required because bullet headers don't know about used precision.

But you do not use these flags anyway now, and there is still no agreement about how to configure and handle `BT_USE_DOUBLE_PRECISION`.
I suggest to remove the `BT_USE_DOUBLE_PRECISION` from this PR and merge it to 0.46, and then continue discussion in the separate PR's thread, when we really will need this flag.

elsid: Done

akortunov:
I instealled Bullet with double precision.
It seems GCC does not like references to float versions of some Bullet functions, which we use:

```
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: CMakeFiles/openmw.dir/mwphysics/physicssystem.cpp.o: in function `MWPhysics::PhysicsSystem::castSphere(osg::Vec3f const&, osg::Vec3f const&, float)':
/home/andrei/Downloads/openmw/apps/openmw/mwphysics/physicssystem.cpp:876: undefined reference to `btCollisionWorld::convexSweepTest(btConvexShape const*, btTransform const&, btTransform const&, btCollisionWorld::ConvexResultCallback&, float) const'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: CMakeFiles/openmw.dir/mwphysics/physicssystem.cpp.o: in function `MWPhysics::PhysicsSystem::updateWater()':
/home/andrei/Downloads/openmw/apps/openmw/mwphysics/physicssystem.cpp:1442: undefined reference to `btStaticPlaneShape::btStaticPlaneShape(btVector3 const&, float)'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: CMakeFiles/openmw.dir/mwphysics/physicssystem.cpp.o: in function `MWPhysics::PhysicsSystem::getHitContact(MWWorld::ConstPtr const&, osg::Vec3f const&, osg::Quat const&, float, std::vector<MWWorld::Ptr, std::allocator<MWWorld::Ptr> >)':
/home/andrei/Downloads/openmw/apps/openmw/mwphysics/physicssystem.cpp:713: undefined reference to `btConeShape::btConeShape(float, float)'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: CMakeFiles/openmw.dir/mwphysics/actor.cpp.o: in function `MWPhysics::Actor::Actor(MWWorld::Ptr const&, osg::ref_ptr<Resource::BulletShape const>, btCollisionWorld*)':
/home/andrei/Downloads/openmw/apps/openmw/mwphysics/actor.cpp:60: undefined reference to `btCapsuleShapeZ::btCapsuleShapeZ(float, float)'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: CMakeFiles/openmw.dir/mwphysics/heightfield.cpp.o: in function `MWPhysics::HeightField::HeightField(float const*, int, int, float, float, float, float, osg::Object const*)':
/home/andrei/Downloads/openmw/apps/openmw/mwphysics/heightfield.cpp:54: undefined reference to `btHeightfieldTerrainShape::btHeightfieldTerrainShape(int, int, void const*, float, float, float, int, PHY_ScalarType, bool)'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: ../../components/libcomponents.a(bulletshapemanager.cpp.o):(.data.rel.ro._ZTVN8Resource17TriangleMeshShapeE[_ZTVN8Resource17TriangleMeshShapeE]+0x28): undefined reference to `btCollisionShape::getBoundingSphere(btVector3&, float&) const'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: ../../components/libcomponents.a(bulletshapemanager.cpp.o):(.data.rel.ro._ZTVN8Resource17TriangleMeshShapeE[_ZTVN8Resource17TriangleMeshShapeE]+0x38): undefined reference to `btCollisionShape::getContactBreakingThreshold(float) const'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: ../../components/libcomponents.a(bulletshapemanager.cpp.o):(.data.rel.ro._ZTVN8Resource17TriangleMeshShapeE[_ZTVN8Resource17TriangleMeshShapeE]+0x50): undefined reference to `btTriangleMeshShape::calculateLocalInertia(float, btVector3&) const'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: CMakeFiles/openmw.dir/mwphysics/trace.cpp.o: in function `MWPhysics::ActorTracer::doTrace(btCollisionObject const*, osg::Vec3f const&, osg::Vec3f const&, btCollisionWorld const*)':
/home/andrei/Downloads/openmw/apps/openmw/mwphysics/trace.cpp:69: undefined reference to `btCollisionWorld::convexSweepTest(btConvexShape const*, btTransform const&, btTransform const&, btCollisionWorld::ConvexResultCallback&, float) const'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: CMakeFiles/openmw.dir/mwphysics/trace.cpp.o: in function `MWPhysics::ActorTracer::findGround(MWPhysics::Actor const*, osg::Vec3f const&, osg::Vec3f const&, btCollisionWorld const*)':
/home/andrei/Downloads/openmw/apps/openmw/mwphysics/trace.cpp:107: undefined reference to `btCollisionWorld::convexSweepTest(btConvexShape const*, btTransform const&, btTransform const&, btCollisionWorld::ConvexResultCallback&, float) const'
```

I did a clean build with CMake re-run.

elsid:
What compiler flags have you used to build openmw? Without `-DBT_USE_DOUBLE_PRECISION` it will not work. Have you tried to build before?

akortunov:
> What compiler flags have you used to build openmw? Without -DBT_USE_DOUBLE_PRECISION it will not work.

I thought that we removed the `-DBT_USE_DOUBLE_PRECISION` flag since we do not use a conditional compilation now:
```
andrei ~/Downloads/openmw/build % cmake -DDESIRED_QT_VERSION=5 -DBT_USE_DOUBLE_PRECISION=True ../ 
  Manually-specified variables were not used by the project:

    BT_USE_DOUBLE_PRECISION    
```

wareya:
> BT_USE_DOUBLE_PRECISION must be defined during the build process (or before including bullet headers) if you are linking against double-precision bullet.

akortunov:
So we again returned back to our original issue - how to determine when we need to declare this flag?

ananace:
Perhaps we should add a [try_compile](https://cmake.org/cmake/help/latest/command/try_compile.html#id3) step in the CMake lists, I've seen other projects use that for similar reasons before.

akortunov:
> I've seen other projects use that for similar reasons before.

[OsgBullet](https://github.com/mccdo/osgbullet/blob/master/CMakeModules/FindBulletHelper.cmake) uses this approach, but I do not know if this approach is reliable.

elsid:
> So we again returned back to our original issue - how to determine when we need to declare this flag?

I thought we agreed to delay the answer for this question after 0.46 🤦‍♂:

> I suggest to remove the `BT_USE_DOUBLE_PRECISION` from this PR and merge it to 0.46, and then continue discussion in the separate PR's thread, when we really will need this flag.

Do we want to provide openmw in two versions: double and single precision? Let's answer this first and decide how this will work at distribution level on different platforms. After we can think about implementation in cmake.

AnyOldName3:
If double precision magically fixes a bazillion problems (which I think the whole point of the discussion was that it does) and doesn't force us to totally drop support for 32-bit machines (which it shouldn't, but package maintainers may have done dumb things) there's no particular reason to keep single-precision mode. If we have to keep two implementations of everything around, and one's shitty and full of weird workarounds, that's just going to turn into a maintenance nightmare. Also, if we pile a bazillion hacks and workarounds on top of the single-precision path to make it not suck, it might well end up slower than the double-precision path even on 32-bit machines, especially if it's soaking up dev time that could have gone to optimising the double-precision path.

psi29a:
Agreed, this is another thing I would like to try out early in 0.47 and see what that gives us.

akortunov:
I suppose it is a time to review this PR again.
The main question is which precision we want to support - only f64, or both f32 and f64.
It should allow us to determine how to declare the BT_USE_DOUBLE_PRECISION.

wareya:
I'm in the f64-only camp, but if it continues to be impossible to ensure that 32-bit builds use f64 bullet on certain linux distributions or whatever then I don't know what to suggest.

psi29a:
I see... so the 32-bit issue comes up again. Does anyone have any compelling reasons to keep 32-bit builds?

AnyOldName3:
Some Windows tablets are still 32-bit IIRC. I think we should try and keep 32-bit builds if we can. If downstream packagers have builds fail with single-precision Bullet on 32-bit targets, that's probably the best way of doing things. They can fix up the precision-and-word-size conflation or they can drop 32-bit themselves. If it's possible to successfully build something that won't work properly, that's bad.

psi29a:
This PR still has some conflicts that need to be resolved and I'm glad we have the ability to choose between float and double. :)

@elsid I have a few additional questions, do you (or anyone) know what the performance impact of switching to double over float?

What does switching to double solve for OpenMW?

If OSG is any indication, I would expect bullet (and thus openmw) to be less performant.

wareya:
Collision math is different from rendering math in that it generally tries to compensate for numerical precision problems with stuff that involves *running more code*, so it's not easy to make a clear assumption that double will be slower. If anything, getting more-accurate collision normals out of bullet might make things faster, because the movement solver will struggle less to find a solution. But the only way to know for sure is to do a direct comparison, and it would differ depending on what kinds of geometry you're throwing at eachother.

Memory usage will increase no matter what though.

wareya:
Also: I don't think that forcing windows builds to be f64-only will be a problem. There shouldn't be any issue with using f64 bullet on 32-bit windows. The problem was package managers that only package f32 32-bit bullet and f64 64-bit bullet, with no f64 32-bit bullet or f32 64-bit bullet, wasn't it?

akortunov:
> The problem was package managers that only package f32 32-bit bullet and f64 64-bit bullet, with no f64 32-bit bullet or f32 64-bit bullet, wasn't it?

It depends on OS. Some Linux distros build 64-bit Bullet libraries with single precision by default.

wareya:
Yeah, that too.

akortunov:
> What does switching to double solve for OpenMW?

It *may* solve:
1. Jittering when using box shapes for humanoids.
2. NPCs-fall-through-the ground issue (which happens with caravaners and Danso Indules).

wareya:
There are a lot of subtle problems with bullet's approach to continuous collision detection (i.e. it uses distance support functions and moves objects closer together until they're withing a collision margin rather than doing analytic/closed-form contact tracing) that are improved a lot by switching to doubles, especially with very large triangles and/or far away from the world's origin. It's more than just fixing jittering with AABBs, it would basically make it possible for me to revive my movement solver PR in general, with less hacks (but still some hacks because of badly-designed vanilla morrowind assets).

akortunov:
> especially with very large triangles

Yes, it is a presumed [cause](https://pybullet.org/Bullet/phpBB3/viewtopic.php?t=3524) of falling caravaners.

psi29a:
Then going double over float is a work-around for the issues we see then?

> far away from the world's origin

Can't this be solved by shifting the coordinate frame?

> especially with very large triangles

This can also be solved in the nifbulletloader by subdividing triangles that are too large. 

So if those are the main use-cases that double precision is supposed to improve (solves?), then I perhaps we should fix those other areas first?

I still don't believe that going from float to double will improve overall OpenMW performance, but would have the opposite effect. Can anyone prove otherwise with some testing?

wareya:
Double-precision bullet already saw primitive testing by me back when I was working on my movement solver fixes. Start [here](https://github.com/OpenMW/openmw/pull/1794#issuecomment-437138371) (make sure to load all hidden comments) and keep reading down until you get to the part where this PR (the double precision bullet one) was first posted. Yes, it starts with me struggling to get it compiling and ends with me giving up at fixing double precision support bugs, but that's exactly where this PR comes in.

At the time, I saw no dramatic performance drawbacks on my system (an i5 6600 non-k, and yes, physics is a bottleneck on my system) and the collisions returned by bullet were *so much* cleaner that it's possible to dramatically dial back how aggressive the movement solver is with collision margins and switch to AABB collision shapes (faster, accurate to morrowind, fewer weird collision edge cases than capsules, etc) because they're less buggy. These improvements happen everywhere, not just against giant triangles or far from the world's origin.

I would be very happy to see more people test double-precision bullet's performance and accuracy, and that certainly has to happen before this is merged, but ***please*** stop directing the conversation in directions that imply that no testing has been done whatsoever or that the belief that double-precision bullet is significantly more accurate is unfounded. It's not neutral, it's condescending.

----

Giant triangles and being far from the origin are just two cases where things are worsened *even more*. Single-precision bullet already gives horrible normals for primitive-vs-mesh collisions in general. It's part of why the movement solver can't easily be fixed without running into new problems, and why we use capsules instead of AABBs (vanilla-accurate, faster, etc).

> Can't this be solved by shifting the coordinate frame?

This is incredibly difficult and prone to causing bugs.

> but I don't know of any games that use double-precision coordinates.

Most game collision systems (at least ones that are not repurposed general-purpose physics engines like Bullet) attempt to use analytic solutions when performing continuous contact traces, or otherwise have specially-designed kinematic body collision systems.

Bullet doesn't. Bullet's designed in a way where numerical precision losses bubble out through all of its algorithms. You haven't felt dread until you've seen an axis-aligned AABB colliding with axis-aligned triangles giving collision normals pointing **entire degrees** away from where they *should* be pointing, and what's more, pointing in directions that cause vector rejection to direct the kinematic body's velocity vector *back into the wall the body just collided with*, what horror!

Bullet was not designed around kinematic motion, it was designed around rigidbody simulation, and it shows. The fact that double-precision bullet improves things so much is wonderful, but please do not mistake acknowledging this as somehow supporting the idea that games with collision systems *designed around* kinematic physics would benefit from switching to double precision collision math as much as OpenMW does.

Having said that, there *are* lots of high-profile games that use doubles all over the place, like Minecraft. It's not unusual.

>The chief reason is that double precision is much slower

Double-precision floats are not "much slower" on modern CPUs. Slightly slower, maybe. Perhaps they're much slower when you're filling up every SIMD register available or the extra memory usage makes you miss cache more, but those are auxiliary issues. If they really were horribly slower, I would have noticed it during my initial testing, since physics is one of the bottlenecks on my system.

>This can also be solved in the nifbulletloader by subdividing triangles that are too large.

This increases memory usage, and if we're talking about things like the geometry in Vivec exteriors, then it wouldn't necessarily be insubstantial. Now, this increase in memory usage wouldn't necessarily have a performance impact, but memory is also one of the places where doubles lose out. It goes both ways.

>Not only this, but bullet is also heavily tuned for SIMD and other low-level vector/matrix optimisations, and those don't work as well (or at all?) with 64-bit floats. Using a 64-bit processor doesn't help you there.

What???????? Basic 64-bit float math is in ***SSE2***. It's not 1999, it's 2020.

AnyOldName3:
The big space games (like Elite Dangerous and Star Citizen) use double-precision with origin shifting but they're dealing with humongous regions, so what they need will be different from what we need.

Also, I did some cursory googling, and it looks like SIMD extensions usually go at the same speed, but with half the data with double-precision floats. It looks like Wareya already addressed that while I was googling, though. IIRC, x86/AMD64 processors work out both single and double-precision floats with 80 bits of internal precision, so there shouldn't be any speed difference from the actual calculation, just the speed difference from vectors only being effectively half as wide.

psi29a:
@wareya I understand and can feel the existential dread in how you type about bullet. That is a feeling I think most people who work on OpenMW physics feel, it would explain why so few people have ever touched it. If you look at the same thread you posted, you will see me championing your work. It is also why I'm here in this thread as well, wondering if there aren't other things that could be done to improve the situation. 

I modified my post slightly to encourage _additional_ testing, not to presume that none has already been done. :)

I like that this PR treats it as optional, so we could pull it in in the 0.47 dev cycle and have more people test, in addition to playing around with other ways to help the situation.

elsid:
Tried to measure `physics_time_taken` in a scene with 16 guars wandering around stones for 10000 frames with 60 fps limit. Doesn't look like there is a significant difference:
bullet_physics_frame_time
Save: [aiwander_more_guars.zip](https://github.com/OpenMW/openmw/files/4631571/aiwander_more_guars.zip)
[stats.bullet_float_precision.log](https://github.com/OpenMW/openmw/files/4631580/stats.bullet_float_precision.log)
[stats.bullet_double_precision.log](https://github.com/OpenMW/openmw/files/4631568/stats.bullet_double_precision.log)

psi29a:
Let's do this.

akortunov:
Still does not build for me, when library with double precision is used.
OpenMW fails to link, and I suppose it is because the `BT_USE_DOUBLE_PRECISION` is missing.

elsid:
You previously suggested how to build with double precision https://github.com/OpenMW/openmw/pull/2049#issuecomment-551626173 . Change compiler flags for now. It’s enough for testing.

akortunov:
https://github.com/OpenMW/openmw/pull/2848

PR #2051, opened 2018-12-01
Add Invert X Axis option (feature #3610)

Capostrophic:
[Feature 3610](https://gitlab.com/OpenMW/openmw/issues/3610).

Adds another button to Controls section of the in-game settings. When it's on, the horizontal axis of the camera are inverted. It works very similar to how invert Y option works, so although I don't have a controller it should work fine for controllers too.

The button caption is not localized, so it obviously may look out of place on non-English localizations.


zinnschlag:
I see we are still adding stuff that is supposed to be added post-1.0. :( 

But all right. Makes sense.

PR #2052, opened 2018-12-01
fix moc achieving to reach unparsable boost code

mp3butcher:
note: I used moc from QT 5.9.7 (linux x86_64)


psi29a:
Has this been tested with Qt4 as well?

mp3butcher:
it can't harm even with qt4

PR #2053, opened 2018-12-02
Native animated containers support

akortunov:
I decided to place it here to collect an early feedback.
Just a basic implementation - add opening and closing animations with sounds.
Part of opening animation is played in the GUI mode, so player does not need to wait until end of opening animation.

Archive with resources for testing: [AC.zip](https://github.com/OpenMW/openmw/files/2636831/AC.zip)
[Link](https://ci.appveyor.com/api/buildjobs/353i8nf4958lpcu0/artifacts/OpenMW_MSVC2015_x64_2053_f3e4d6773a5c5e87f702a6747c34cc6d1b34626f.zip) to actual build.
ESM file contains only Sound records and does not modify any container records, so in theory it can be used with any total conversion or landmass mod, which uses vanilla assets.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -419,6 +419,43 @@ void CharacterController::refreshJumpAnims(const WeaponInfo* weap, JumpingState 
     }
 }
 
+bool CharacterController::onOpen()
+{
+    if (mPtr.getTypeName() == typeid(ESM::Container).name())
+    {
+        if (!mAnimation->hasAnimation("containeropen"))
+            return true;
+
+        if (mAnimation->isPlaying("containeropen"))
+            return false;
+
+        if (mAnimation->isPlaying("containerclose"))
+            return false;
+
+        mAnimation->play("containeropen", Priority_Persistent, MWRender::Animation::BlendMask_All, false, 1.0f, "start", "stop", 0.f, 0);
+        if (mAnimation->isPlaying("containeropen"))
+            return false;
+    }
+
+    return true;
+}
+
+void CharacterController::onClose()
+{
+    if (mPtr.getTypeName() == typeid(ESM::Container).name())
+    {
+        if (!mAnimation->hasAnimation("containerclose"))
+            return;
+
+        float complete, startPoint = 0.f;
+        bool animPlaying = mAnimation->getInfo("containeropen", &complete);
+        if (animPlaying)
+            startPoint = 1.f - complete;
+
+        mAnimation->play("containerclose", Priority_Persistent, MWRender::Animation::BlendMask_All, false, 1.0f, "start", "stop", startPoint, 0);
+    }
+}
+
 void CharacterController::refreshMovementAnims(const WeaponInfo* weap, CharacterState movement, CharacterState& idle, bool force)
 {
     if (movement == mMovementState && idle == mIdleState && !force)
@@ -1079,6 +1116,8 @@ void CharacterController::handleTextKey(const std::string &groupname, const std:
 
     else if (groupname == "shield" && evt.compare(off, len, "block hit") == 0)
         mPtr.getClass().block(mPtr);
+    else if (groupname == "containeropen" && evt.compare(off, len, "loot") == 0)
+        MWBase::Environment::get().getWindowManager()->pushGuiMode(MWGui::GM_Container, mPtr);
 }
 
 void CharacterController::updatePtr(const MWWorld::Ptr &ptr)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
     Feature #3442: Default values for fallbacks from ini file
     Feature #3610: Option to invert X axis
     Feature #4673: Weapon sheathing
+    Feature #4730: Native animated containers support
     Task #4686: Upgrade media decoder to a more current FFmpeg API
 
 0.45.0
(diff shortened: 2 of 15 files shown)


zinnschlag:
I am considering how we could include this feature pre-1.0. I don't think this will collide with any future enhancements, except for resources and IDs, both of which need to be taken care of. 

It seems the only IDs we are dealing with are the new sound IDs, right? If we move them into the sys namespace that should avoid future name collisions and be compatible with all future enhancements.

On the resource front we have sounds and meshes. Now the meshes are content file specific (a TC may or not may use these specific container meshes), so having an addon file makes sense. The Morrowind game can use it and any TC that reuses Morrowind assets can also adapt the addon for its own use (or bake it into their core resources). I would however use an omwaddon file instead of an esp to make clear that this is not for the vanilla Morrowind engine.

I am not so sure about the sound resources. These feel more like fallback items (similar to fallback values in cfg files). We don't have fallback resources yet. Question is should we start to introduce them? This needs a (probably longer) discussion.
Also, if we go with fallback sound resources, we can inject the sound records on loading (when loading a legacy version esm/omwgame file) like we are doing it with some others records already. In this case we would avoid the need for addon file altogether and the only thing the user would have to do is to drop the mesh replacer into his data folder.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -419,6 +419,47 @@ void CharacterController::refreshJumpAnims(const WeaponInfo* weap, JumpingState
...
+    if (mPtr.getTypeName() == typeid(ESM::Container).name())
+    {
+        if (!mAnimation->hasAnimation("containeropen"))
+            return true;
+
+        bool playing = mAnimation->isPlaying("containeropen");
+        if (playing)
+            return false;
+
+        playing = mAnimation->isPlaying("containerclose");
```
zinnschlag: Can we avoid this kind of variable re-use please? Doesn't look like we need a variable in the first place. The function call can go directly into the condition of the if statement.

zinnschlag:
> In current implementation sounds are handled in the open/close animations directly - we do not handle them separately on the engine level.

I need to look at that again. And the records in your esp file. Not entirely sure if I understand how that works.

>What kind of fallback resources are you talking about?

Fallback resources are a concept that I expected to show up for quite some time. But it hasn't yet. I'll give a quick explanation. But for more details I'll have to point you to the post-1.0 design doc, which describes fallbacks and handling of older content files in greater detail.
Basically the idea is that newer versions of the engine may need some values/resources that were not needed in previous version. An example would be a hardcoded value that has now become configurable from within the content files. In newer version game files these values/resources would then be mandatory.
This by itself would stop the engine from loading older game files (e.g. Morrowind.esm). Therefore we have a need for fallbacks that will be made available automatically in such a case.

> An interesting idea, but I am not sure if there is a reliable way to generate such records - a content developer may want to modify loudness, for example.

They are only fallbacks. Content based on a modern version omwgame file wouldn't even get to see them and content based on a legacy gamefile (esm or current omwgame format) could easily modify them via a addon.

akortunov:
> Not entirely sure if I understand how that works.

The current approach is very simple:
1. KF animation files contain Sound text key keys, for example
```
Sound: AC_sack_open
```
2. OpenMW just handles there text keys as any other similar text keys in animation files (tries to play sound with given ID).
3. If there are no suitable sounds, sound just will not be played - container will be opened\closed silently.
4. It is a content developer responsibility to add sound records (and sound files) which he uses in his mod (of course, only if these sounds are not present in default Morrowind assets).

akortunov:
Also there is a small issue with current implementation - most of opening animation is played in GUI mode, but currently I do not update animated shapes in GUI mode. 
You can reproduce this issue by enabling TCB and opening animated container.
Is it worth to fix, or we should advise to use RootCollisionNodes for better performance instead?

Capostrophic:
To clarify again: this feature doesn't concern any sound playback at all. Sound textkeys are existing animation functionality, as noted up here.
The feature simply makes `containeropen`/`containerclose` animation groups and `loot` textkey in container animation files automatically used if they are present. The container window doesn't open on its own if the container open animation is still playing, but `loot` textkey allows the modmaker to work around that -- the container will be updated during the GUI pause so it's still animated.

zinnschlag:
Thanks for the clarification.

> It is a content developer responsibility to add sound records (and sound files) which he uses in his mod 

This is generally true. I could make the point that this situation is different, but I am not going to press on with it, because I am not convinced either that this is a good enough reason to introduce the concept of fallback resources.

> Also there is a small issue with current implementation - most of opening animation is played in GUI mode, but currently I do not update animated shapes in GUI mode.
>You can reproduce this issue by enabling TCB and opening animated container.
> Is it worth to fix, or we should advise to use RootCollisionNodes for better performance instead?

Well, can you fix it without breaking something else?

zinnschlag:
Try it out and see if anything breaks? Its early enough in 0.46.0 that I would not be worried about the consequences of this approach.

akortunov:
In theory, there could be issues with heavily scripted containers, but I did not find any with vanilla scripts.
Also it would be nice to get a feedback from TES3MP developers.
@davidcernat, is this approach OK, or it will cause issues in multiplayer?

davidcernat:
I doubt this will cause issues in multiplayer. If it does, it will probably only require minimal adjustments.

akortunov:
> I doubt this will cause issues in multiplayer. If it does, it will probably only require minimal adjustments.

Well, in this case this PR should be ready for final review.

zinnschlag:
Looks good to me.

Capostrophic:
I don't see and haven't felt any issues with this either.

PR #2054, opened 2018-12-02
fix a rig bug visible with OSG_VERTEX_BUFFER_HINT=VAO

mp3butcher:


AnyOldName3:
After someone sent me a PM pointing something out to me and reminding me about the last sentence of https://wiki.openmw.org/index.php?title=Rendering_Architecture#Threading_considerations, I no longer think that this is an acceptable solution. It should be possible to use a VBO with `GL_DYNAMIC_DRAW_ARB` set and a VAO without setting the whole Geometry to dynamic, so a more surgical solution to the bug is the best approach. This PR should be reverted.

Regarding the original bug that this suppresses, what did it look like, exactly? Were things just not animated?

mp3butcher:
the purpose of this flag is not to tag buffer as dynamic but the vertexarraystate...so it is valid...

mp3butcher:
@AnyOldName3 
I'm curious what that PM was about?...

AnyOldName3:
> I'm curious what that PM was about?...

It contained some unrelated stuff that I'm not at liberty to publish or I'd share it.

I'll look into whether this is a valid exception to the rule in the threading guide in detail tomorrow. The person who reported this to me seems to have looked at the OSG code in detail, though, so one of you two missed something that the other has seen.

mp3butcher:
I wrote the first implementation of VAO support in osg
Robert was not happy with it but i follow its work on it...

mp3butcher:
@AnyOldName3
sorry to bother, but at first glance (perhaps i missed it),
 all your worker threads (SceneUtil::WorkItem) shared resources aren't protected by mutices.
A lot of crash may come from concurrent read/write..

psi29a:
@AnyOldName3 even with this:
https://github.com/openscenegraph/OpenSceneGraph/issues/655#issuecomment-446780837
that will not help us here with this commit right?

mp3butcher:
I'm afraid you don't understand the issue...
this commit is good to test osg  
Let this commit until we hacked a new way out of VAS Managment

Capostrophic:
Keeping a commit that literally renders the game unplayable (breaks saving) in master, even though it works around some minor rendering glitches, until something works out the kinks of the solution, doesn't make much sense. master branch is supposed to be at least *somewhat* stable - there are people that were playing happily on git builds for a while and harming their experience for no reason is not good.

And it harms contribution too. I'm not in the ability to solve the issue and applying manual patches removing the two lines every time I make a clean branch just to make saving work is inconvenient.

mp3butcher:
it's not openmw responsability

Capostrophic:
>it's not openmw responsability but osg

The current focus is on forked 3.4.1 and that won't change in foreseeable future because there's no reason it should. The declared minimal version dependency is not 3.4.0, even. Some distributions OpenMW supports (that are not EOL) don't even offer OSG 3.6 packages.

Oh, so you're confusing a different issue (random crashes with OSG 3.6.3 due to improper geometry removal) with this one.

mp3butcher:
yeah i'm totally confused that you're building against a hacked osg version instead of contributing to the main stream evolution

Capostrophic:
That's not the point.

mp3butcher:
your issues seams all related to the new VAS design i think. So I'd though that make openmw compliant with it would make your claims gains in credibility against osg community...

Capostrophic:
The crash happens *even on 3.4.1*, which *does not* have the broken memory leak fix in it.

mp3butcher:
strange?
did 3.4.1 use osg::Drawable::DataVariance somewhere?

psi29a:
@mp3butcher : I think you're taking things a bit personal here. 

The argument is this PR breaks our nightlies, how do we triage this? 

Leaving it in isn't an option because there isn't a solution forthcoming yet.

This effects _all_ versions of OSG-3.4, upstream and downstream from Debian and Ubuntu.

From a pragmatic point of view, reverting this PR is the 'right' thing to do to get OpenMW nightlies into users hands and usable.

The correct way forward is to re-enable your work when it works with OpenMW without crashing.

mp3butcher:
I don't say the contrary (further uploaded 3.6 binaries embed the bug)

psi29a:
> I don't say the contrary (further uploaded 3.6 binaries embed the bug)

OpenMW doesn't link against 3.6 in our nightlies nor our releases, only against OSG-3.4

mp3butcher:
On what did I work then?

psi29a:
We honestly don't know, it isn't very clear. Possible language barrier issues? I'm not trying to sound patronizing, but I can tell that this is hugely frustrating. :(

Perhaps we can start again? It seems that a better way forward to fixing the issues was talked about here:
https://github.com/OpenMW/openmw/pull/2054#issuecomment-446419840

AnyOldName3:
> yeah i'm totally confused that you're building against a hacked osg version instead of contributing to the main stream evolution

Every change in the fork was made as a PR to upstream OSG, too. Most of the changes ended up in OSG 3.5.x, but the biggest one was adding single-precision Quat support, and making quats single-precision by default. Robert disagreed with this, as it would obviously reduce accuracy when OSG was used for serious simulations, so it didn't get merged upstream. It gave about a 15% performance boost to OpenMW, though. It's possible that that improvement is lower now and the fork is unnecessary, but I don't think anyone's done a direct comparison recently.

Anyway... Regarding this change in particular, it's obvious it exposes another bug we've got somewhere else which should be fixed, but I'm still sceptical that the threading implications of using a drawable labelled as dynamic are as minor as you've suggested. I've not had time to look into it today, unfortunately, but I'll try to do so tomorrow. Theoretically, OpenMW should have enough armour in it that it's never necessary to label anything as dynamic - for example, the drawables that this change affects are double-buffered, so the instance being edited shouldn't be in flight at the same time.

My best guess is that there's an OSG bug where, when a VAO is in use, it doesn't check to see if `dirtyGLObjects` has been called, so doesn't know that the vertex array has been updated, and doesn't put the new data in the VBO. Working out if this is the case is yet another thing I need to look at tomorrow.

mp3butcher:
https://github.com/openscenegraph/OpenSceneGraph/blob/ab32dd859b587fe24510f7b3339c6a0d1cd46a3e/src/osg/Drawable.cpp#L663

AnyOldName3:
I've done some digging, and I can confirm that using any dynamic drawables at all definitely kills our threading. After the cull traversal, the render bins count up how many render leaves they have whose drawables have `osg::Object::DYNAMIC` set. This count is added up per render stage, and then that count goes to the scene view, and then to the state object. The viewer has a semaphore which causes the rendering traversals to block while contexts have states whose dynamic object counts haven't gone down to zero this frame, so the next traversal is delayed, effectively making everything sequential.

I reckon that the crash might have been caused because we have many threads doing many things, and OSG wasn't prepared for a thread to modify something it thought we shouldn't (even though, in practice, everything that might have problems is double-buffered or otherwise protected). Alternatively, because the save game screenshot rendering traversal happened when the viewer wasn't ready for it, it's possible that the semaphore wasn't initialised for this traversal, and we ended up doing something which that didn't like.

mp3butcher:
if you're mod your scene graph in other way than osg Callback, I'm afraid you're doomed to osgViewer::SingleThreaded.

AnyOldName3:
We use all [these](https://wiki.openmw.org/index.php?title=Rendering_Architecture#Threading_considerations) strategies, so it's never objects which OSG has access to that are being modified. For example, RigGeometry and MorphGeometry should be safe because the underlying geometry is double-buffered and there's a cull callback being used to select which one to cull on a frame-by-frame basis.

I don't have a stack trace of the crash this PR caused (I can't actually *build* OpenMW now due to other problems), but from what I remember I'm now fairly sure that it's because we dispatch another rendering traversal from either a thread or a time that OSG doesn't expect, not because we're editing something illegally. When there are one or more dynamic objects in the scenegraph, the viewer waits before executing the following stuff, so I'm pretty sure the crash was caused because on another thread, the viewer started accessing the stuff changed here:

```c++
    if (_releaseContextAtEndOfFrameHint && doneMakeCurrentInThisThread)
    {
        //OSG_NOTICE<<"Doing release context"<<std::endl;
        releaseContext();
    }

    if (getViewerStats() && getViewerStats()->collectStats("update"))
    {
        double endRenderingTraversals = elapsedTime();

        // update current frames stats
        getViewerStats()->setAttribute(frameStamp->getFrameNumber(), "Rendering traversals begin time ", beginRenderingTraversals);
        getViewerStats()->setAttribute(frameStamp->getFrameNumber(), "Rendering traversals end time ", endRenderingTraversals);
        getViewerStats()->setAttribute(frameStamp->getFrameNumber(), "Rendering traversals time taken", endRenderingTraversals-beginRenderingTraversals);
    }

    _requestRedraw = false;
```

It's probably a good idea for the extra rendering traversals to be dispatched in the way OSG expects, but it's not strictly necessary if we don't have any dynamic objects, which we shouldn't need.

AnyOldName3:
I've gained access to a stack trace, and it seems that OpenMW just hardlocks when it's waiting for the dynamic drawables to all have been drawn rather than crashes. It's even simpler than I thought. It's still caused by calling the rendering traversals when OSG isn't ready, but it looks like we're dispatching it without a cull traversal (as the results should be the same) so the dynamic drawable count hasn't been reinitialised, so there's nothing to wait for.

AnyOldName3:
> The purpose of Drawable Datavariance = DYNAMIC is here
> https://github.com/openscenegraph/OpenSceneGraph/blob/ab32dd859b587fe24510f7b3339c6a0d1cd46a3e/src/osg/Drawable.cpp#L663
> Perhaps there's a bug, my attention to what Robert do is relative

I've taken a closer look at this, and the VAS' `_requiresSetArrays` property has already been set by `dirtyGLObjects`, then gets cleared again because the drawable isn't dynamic. This definitely looks like a mistake to me - it should probably be ORing in the existing value.

Unfortunately, while tweaking just that line would solve the animation issue OpenMW's having, `_requiresSetArrays` never gets cleared anywhere else, so the vertices would get updated every frame, even if it was unnecessary. To resolve that, it would need to be set back to false once it had been checked, namely here https://github.com/openscenegraph/OpenSceneGraph/blob/ab32dd859b587fe24510f7b3339c6a0d1cd46a3e/src/osg/Geometry.cpp#L937 as an else, here https://github.com/openscenegraph/OpenSceneGraph/blob/ab32dd859b587fe24510f7b3339c6a0d1cd46a3e/src/osgTerrain/GeometryPool.cpp#L958 and in three similar places in `osgText`.

mp3butcher:
...err..dirtyGLObject and drawimplementation are totally different code paths..:/
I'm still thinking you believe drawable datavariance= DYNAMIC reupload arrays but it's not
> Unfortunately, while tweaking just that line would solve the animation issue OpenMW's having, 
If you're always talking of openmw with  VAO_VERTEX_BUFFER_HINT=VAO you should NOT remove that line (it invalidates underlying vao)
Please precise the bug you'r talking about 
 
You have to deal with the new VAS design.

AnyOldName3:
OpenMW calls `dirtyGLObjects` after it changes the vertices when doing skinning. Before VAOs were a thing, this was always enough to ensure OSG sent the updated vertices to the GPU. Since the VAO change, the code that handles that is fenced off behind a check which returns when `_requiresSetArrays` is false, so it never gets executed unless that's set, but `dirtyGLObjects` *does* set it, so everything should be fine. Unfortunately, it gets set back to false again for non-dynamic drawables, so that line should be changed so it doesn't make it false when it's already true, but does make it true if the drawable is dynamic.

It's possible that things are in a bigger mess than I think and there're multiple things that are getting skipped when that boolean is false, but this is definitely a starting point for the next bit of investigation.

mp3butcher:
openmw never call dirtyGLObjects but dirty on Arrays?!
if you're talking about the  character sliding issue I'm seeing 
I fixed it by swapping to SingleThreaded :/
 

It should

mp3butcher:
err ?! The sliding character bug misteriously disappear even with threaded viewer?!
I may have to something but don't know what

mp3butcher:
I think I have diagnose the OpenMW big problem:
 Performances are better in SingleThreaded viewer!!->Too much RenderBins
Do you have a RenderBin policy?
I'll try to remove them but I'm afraid to touch some inner conventions

mp3butcher:
Please, Can anyone tell me where in the openmw code nodes are added to the scene graph?

AnyOldName3:
> openmw never call dirtyGLObjects but dirty on Arrays?!

Looks like you're right there. It does seem like it's only the arrays that we dirty. What I discovered is probably still an OSG bug, but I guess not one that affects OpenMW. When you said

> The purpose of Drawable Datavariance = DYNAMIC is here
> https://github.com/openscenegraph/OpenSceneGraph/blob/ab32dd859b587fe24510f7b3339c6a0d1cd46a3e/src/osg/Drawable.cpp#L663

I started investigating that, and realised that the VAS might already have had `_requiresSetArrays` set, and if it was, for non-dynamic drawables, it would be cleared.

> updating data in cull is generally not a so good practice...openmw may comply with osganimation or use an update callback...

About 3/4 of animated things will be off-screen for any given frame, so don't actually need updating. There's not really any mechanism in OSG to determine what is and isn't off-screen during the update traversal, and as animation is close to being the biggest single time-sink for OpenMW, doing a bunch of it unnecessarily would be a really bad idea. It's not like OSG sends anything to the GPU before the cull callback has finished anyway, so there shouldn't be a performance impact for avoiding skinning until we know the object isn't culled.

We can't use osgAnimation as it doesn't support a bunch of things which Morrowind's animations do.

Animated stuff *does* already do some of its work in the update callback, though. The biggest things are updating the skeleton and updating the bounding volumes, as these are required to know which things are and aren't culled.

> if you're talking about the strange displaced meshes on rigs it's the vao issue this pr was about.

As far as I'm aware, we've only been discussing the one issue in this thread. I've not been able to experience or test it as OSG and OpenMW really don't like linking on Windows, so I'm basically at the mercy of when Ananace updates the OSG build provided by the prebuild script.

> * GPU charge is enormous for as few vertices

Morrowind was being made in the late nineties, and the things which are considered prohibitively expensive now were often considered releatively cheap back then. That means that, despite having few vertices, Morrowind's assets massively overuse things like alpha blending (meaning about half of all drawables end up in the depth-sorted bin and have to be submitted in a specific order, causing loads of state changes) and have lots of tiny objects each using their own unique texture and state. This makes the number of draw calls per vertex absolutely ridiculous by modern standards.

> Performances are better in SingleThreaded viewer!!

I've not tested this recently, but I'd be surprised - stuff is definitely being done while draw calls are beign submitted when I look at the profiler overlay.

> My guess is Too much RenderBins

OpenMW uses the following renderbins:

```c++
/// Defines the render bin numbers used in the OpenMW scene graph. The bin with the lowest number is rendered first.
    enum RenderBins
    {
        RenderBin_Sky = -1,
        RenderBin_Default = 0, // osg::StateSet::OPAQUE_BIN
        RenderBin_Water = 9,
        RenderBin_DepthSorted = 10, // osg::StateSet::TRANSPARENT_BIN
        RenderBin_OcclusionQuery = 11,
        RenderBin_FirstPerson = 12,
        RenderBin_SunGlare = 13
    };
```

* Sky is used to render the skybox before everything else.
* Default is the normal render bin that everything goes into.
* Water is used only when the refraction RTT is disabled so that the water gets drawn after the things behind it. It's not needed when the refraction RTT is in use as everything behind the water is rendered to that texture, so in that case, the water can go in the opaque bin.
* DepthSorted is OSG's standard depth-sorted renderbin that all the translucent stuff goes into when the rendering hint is set to `TRANSPARENT_BIN`.
* OcculsionQuery is used with occlusion queries to see how much of the sun is obscured behind the scene so we know how intense to make the sun glare. This is done as a cull callback, and I'm pretty sure it uses whichever result is available rather than blocking waiting for the result from the previous frame.
* FirstPerson is used for the player's arms and weapon etc. as that's drawn after everything else and after a depth clear.
* SunGlare is used to actually render the sun glare over the top of everything else.

As well as this, RTT stuff is usually done with `osg::Camera::setRenderOrder(osg::Camera::PRE_RENDER)`, but without a renderbin.

This doesn't strike me as a huge number of renderbins, especially as we've got the depth clear before rendering the first person stuff to contend with.

While we're on the subject of renderbins, do you happen to know if setting the renderbin for a stateset will override the rendering hint for anything using it? There's no mechanism to override the rendering hint, and I'm not sure whether or not translucent stuff is being depth sorted before being rendered into the shadow maps in my PR, but it's obviously wasted performance if it is.

AnyOldName3:
By the way, you edited and deleted a bunch of stuff, so I've only responded to what's here, but you've said some stuff about changing the threading model changing whether or not the bug this PR was supposed to solve happens. If any of the stuff you've removed was true, that might give me more to go on while investigating things.

mp3butcher:
Wow thanks for all taking the time:)
Concerning what I deleted: It was noise (mea culpa), stuff I hacked yielding problems
I also remove what I said about performances and bins (it shouldn't be a problem). My guess is that I have a huge gpu charge due to water shading (and damned GT440 for the moment)

> I've not been able to experience or test it as OSG and OpenMW really don't like linking on Windows, so I'm basically at the mercy of when Ananace updates the OSG build provided by the prebuild script.

What kind of problem?

> While we're on the subject of renderbins, do you happen to know if setting the renderbin for a stateset will override the rendering hint for anything using it?

Mmmh I know what You're trying to do :) I've not try but it seems like ss.BinMode=osg::StateSet::PROTECTED_RENDERBIN_DETAILS control that

mp3butcher:
perhaps using/reading conan script https://github.com/mp3butcher/osg_conan_recipe can help with depends

AnyOldName3:
> And don't hesitate to report osg issue . Robert is under linux....

The majority of interactions I've had with Robert have lead to him immediately dismissing whatever I said and giving my username as the justification for why, so I've tended to exhaust all other options first.

Anyhow, regarding the animation bug, I've had some more time for some more investigation.

It looks like, now VAOs are a thing, OSG skips updating dirtied arrays unless the `_requiresSetArrays` boolean has been set, either by making the drawable dynamic or by using `dirtyGLObjects`, which is a new method. With display lists, they got updated when the array was dirtied even if `dirtyDisplayLists` hadn't been called, which is what's lead to much of the confusion. This change may or may not be a bug.

Regarding https://github.com/openscenegraph/OpenSceneGraph/blob/ab32dd859b587fe24510f7b3339c6a0d1cd46a3e/src/osg/Drawable.cpp#L663, I no longer think this is doing a wrong thing. Critically, this line is *after* the drawable is drawn, so if the flag was set by `dirtyGLObjects`, the necessary change will have been applied, and we *want* it to get cleared, but if the drawable is dynamic, we want to ensure it's set again for the next frame.

This all means that we can fix the bug by simply adding a call to `dirtyGLObjects` to OpenMW and everything will be fine. I do, however, think it might be best to get some clarification from Robert as to whether the difference in requiring manual dirtying with display lists to VAOs is intentional or not, as we're not going to be the only project affected by this, and it would be preferable to not have a line be conditional on the OSG version.

AnyOldName3:
Actually, that might not be perfectly accurate - the rigging system uses VBOs instead of display lists, so it's possible it's only VBOs that don't require manual dirtying.

AnyOldName3:
I've had a cursory glance and it looks like display lists would probably have required manual dirtying, too, so it might have only ever been coincidence that made VBOs work.

AnyOldName3:
https://github.com/OpenMW/openmw/pull/2083 now exists, which should fix this issue without adding a single crash. I have no ability to test it myself, though.

AnyOldName3:
Contrary to what I said earlier, it looks like OpenMW is reporting that it uses 50 renderbins, at least with my shadow PR. I've had a look at the cull visitor, and it seems that any RTT camera gets its own set of renderbins (and setting it up via `osg::Camera::setRenderOrder(osg::Camera::PRE_RENDER)` seems to still do renderbins, just automatically). If lots of renderbins is always a bad thing, OpenMW might be optimisable in that respect.

mp3butcher:
arff....I was right...I'm a bit high sometime :)

PR #2056, opened 2018-12-04
esmtool cleanup

Capostrophic:
More minor readability improvements like in the PRs I previously have done - like using std::string::empty instead of `""` comparison, range-based for loops instead of many iterators and cleaning up redundant `""` empty string declarations and usage.

I admit I don't fully understand why there were references constructed for working with certain collections. If changing that is horribly wrong, please let me know.


psi29a:
Perhaps written many many moons ago? :)

davidcernat:
Looks fine to me.

PR #2059, opened 2018-12-06
Don't stack scripted items (bug #2969)

Capostrophic:
[Bug 2969](https://gitlab.com/OpenMW/openmw/issues/2969).

Make scripted items not stack, ensure scripted items added with AddItem don't stack and optimize an item name recovery loop in RemoveItem.

Scripted items must not stack because every stack is effectively a single reference which can have non-1 count but still have 1 script, but every script must be per-instance and not per-stack (or shit breaks due to script states being merged). This is something scrawl changed a while ago seemingly for no reason. Even then, ContainerStore::add happily creates stacks of anything added even if the added items shouldn't stack (they simply won't stack with existing instances of the same item ID if they can't stack).

Using a loop for ContainerStore::add is obviously not ideal for performance, but I can't think of something else that doesn't horribly break stacking.

### DIFF
--- a/apps/openmw/mwscript/containerextensions.cpp
+++ b/apps/openmw/mwscript/containerextensions.cpp
@@ -64,7 +64,19 @@ namespace MWScript
                             || ::Misc::StringUtils::ciEqual(item, "gold_100"))
                         item = "gold_001";
 
-                    MWWorld::Ptr itemPtr = *ptr.getClass().getContainerStore (ptr).add (item, count, ptr);
+                    MWWorld::ContainerStore& store = ptr.getClass().getContainerStore (ptr);
+                    // Create a Ptr for the first added item to recover the item name later
+                    MWWorld::Ptr itemPtr = *store.add (item, 1, ptr);
+                    if (itemPtr.getClass().getScript(itemPtr).empty())
+                    {
+                        store.add (item, count-1, ptr);
+                    }
+                    else
+                    {
+                        // Adding just one item per time to make sure there isn't a stack of scripted items
+                        for (int i = 1; i < count; i++)
+                            store.add (item, 1, ptr);
+                    }
 
                     // Spawn a messagebox (only for items added to player's inventory and if player is talking to someone)
                     if (ptr == MWBase::Environment::get().getWorld ()->getPlayerPtr() )
@@ -143,8 +155,13 @@ namespace MWScript
 
                     std::string itemName;
                     for (MWWorld::ConstContainerStoreIterator iter(store.cbegin()); iter != store.cend(); ++iter)
+                    {
                         if (::Misc::StringUtils::ciEqual(iter->getCellRef().getRefId(), item))
+                        {
                             itemName = iter->getClass().getName(*iter);
+                            break;
+                        }
+                    }
 
                     int numRemoved = store.remove(item, count, ptr);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.46.0
 ------
 
+    Bug #2969: Scripted items can stack
     Bug #2987: Editor: some chance and AI data fields can overflow
     Bug #3623: Fix HiDPI on Windows
     Bug #4411: Reloading a saved game while falling prevents damage in some cases
--- a/apps/openmw/mwworld/containerstore.cpp
+++ b/apps/openmw/mwworld/containerstore.cpp
@@ -246,7 +246,9 @@ bool MWWorld::ContainerStore::stacks(const ConstPtr& ptr1, const ConstPtr& ptr2)
 
         && ptr1.getClass().getRemainingUsageTime(ptr1) == ptr2.getClass().getRemainingUsageTime(ptr2)
 
-        && cls1.getScript(ptr1) == cls2.getScript(ptr2)
+        // Items with scripts never stack
+        && cls1.getScript(ptr1).empty()
+        && cls2.getScript(ptr2).empty()
 
         // item that is already partly used up never stacks
         && (!cls1.hasItemHealth(ptr1) || (
@@ -306,7 +308,7 @@ MWWorld::ContainerStoreIterator MWWorld::ContainerStore::add (const Ptr& itemPtr
     item.getCellRef().unsetRefNum(); // destroy link to content file
 
     std::string script = item.getClass().getScript(item);
-    if(script != "")
+    if (!script.empty())
     {
         if (actorPtr == player)
         {


apps/openmw/mwscript/containerextensions.cpp
```diff
@@ -143,8 +148,13 @@ namespace MWScript
 
                     std::string itemName;
                     for (MWWorld::ConstContainerStoreIterator iter(store.cbegin()); iter != store.cend(); ++iter)
+                    {
                         if (::Misc::StringUtils::ciEqual(iter->getCellRef().getRefId(), item))
+                        {
                             itemName = iter->getClass().getName(*iter);
+                            break;
```
psi29a: We should be doing this more often. Thanks!

psi29a:
Everyone cool with this? :)

i30817:
I'm not sure if I agree with this. One of my mods add scripts to items that can stack (scrolls), and while this should not affect it overmuch it'll probably make vendors annoying.

Also stacks of scrolls in the world...

Capostrophic:
Morrowind doesn't stack scripted items at all. I'm trying to replicate that.

A stack of items is technically a single reference which has a single script. So if you equip one item from the stack, *all carried instances of the item* have OnPCEquip set to 1 when it is equipped. Similar thing about other flags (OnActivate, OnPCDrop, OnPCRepair etc.). This causes various kinds of issues with scripts that expect the scripted item to be "alone".

Capostrophic:
>Like, do you seriously want scripted arrows to turn into 200 separate instances on the inventory all running a script (or scrolls etc)

If that's what Morrowind does (and it does), sure.

>Isn't this a Bethesda mod?

Arrows from Area Effect Arrows are enchanted, not scripted.

i30817:
I cannot say i approve of this because it makes my own mod kind of useless. But it's your habit to chase MW behavior instead of extending it i'm pretty sure that even if i had a great idea, you'd just go ahead.

My current best/easiest idea is:

MW 'system' operations on stack with a script like OnPCEquip, OnPCRepair etc are made illegal (only singular items), except for OnPCDrop that goes ahead as normal.
OnPCDrop that divides a stack into 2 clones the script and on merge the one in the inventory wins always.

Capostrophic:
>to chase MW behavior instead of extending it

for your information, is the universally agreed goal of 1.0.

>my own mod kind of useless.

Don't make OpenMW-specific mods intended for running with Morrowind assets. Make mods compatible with Morrowind engine so OpenMW can work with them too (or will work with them at some point).

i30817:
> Don't make OpenMW-specific mods intended for running with Morrowind assets. Make mods compatible with Morrowind engine so OpenMW can work with them too (or will work with them at some point).

Let me laugh at that, hah fucking hah. First the mod could work because i build the records manually instead of on cs, though i didn't try because i can't run the original. Second your own fileformat extension discourages running mods on morrowind so don't be sanctimonious about this..

i30817:
> Morrowind doesn't stack scripted items at all. I'm trying to replicate that.
> 
> A stack of items is technically a single reference which has a single script. So if you equip one item from the stack, _all carried instances of the item_ have OnPCEquip set to 1 when it is equipped. Similar thing about other flags (OnActivate, OnPCDrop, OnPCRepair etc.). This causes various kinds of issues with scripts that expect the scripted item to be "alone".

Okay, this doesn't really track with the link at https://gitlab.com/OpenMW/openmw/issues/2969

If it's a single instance, how is it printing twice in that script that guards against that? Isn't the main problem that things do get cloned, but that the stack separation is too late

Capostrophic:
Allow me make the behavior closer to *Morrowind* behavior to improve *Morrowind* mod compatibility in peace please, without overcomplicating things trying to adapt a different behavior to act the same.

psi29a:
@i30817 your frustration is understandable, but don't attack developers for it. Calm it down please.

Until OpenMW 1.0 has be released, don't assume that your OpenMW specific mods will keep working. The only 'official' API is the one that Morrowind itself has and the behaviour of MWScript. 

OpenMW has never made an implicit nor explicit promise of compatibility beyond MWScript. 

This will change once we have 'newscript' in place.

i30817:
Again, you didn't explain what happens with your 'separate scripts instances are different items' with stacks on the world already (from multiple mods). Does the script run for all of them (lol), or does it run '1' and when you pick them up they get cloned?

Capostrophic:
>Also the notion of 'enforcing compatibility' on a thing that doen't actually exist on the original...

Excuse me? I have already stated Morrowind doesn't stack scripted items for a very good reason.

>Good luck automatically converting all the old Bethesda copyrighted scripts to 'newscript' without being sued.

OpenMW has already been converting Morrowind scripts into bytecode automatically without knowing what the scripts are for *years* and somehow Bethesda hasn't freed an army of lawyers on anyone from the AUTHORS.md file yet. It's not going to be greatly different if and when mwscript scripts are converted into Lua scripts first.

Like, it's not actually necessary to redistribute anything.

lysol90:
@i30817 Please, try to work _with_ us instead of being sarcastic all the time. Right now you're being a dick and at the same time, you want the devs to do it your way. Want to get some influence in the project? Pro tip: Start with being polite. 

I'm not here to debate anything, just state my opinion in this.

i30817:
> > 
> 
> OpenMW has already been converting Morrowind scripts into bytecode automatically for _years_ and somehow Bethesda hasn't freed an army of lawyers on anyone from the AUTHORS.md file yet. It's not going to be greatly different if and when mwscript scripts are converted into Lua scripts first.

A transpiler is ofc fine. Editing the scripts so they work on a listener model of scripting is another thing entirely. Just look at the trouble scummvm goes through to fix SCI and others scripting languages. It's done with runtime patching of the compiled machine code (not even decompiled sources).

AnyOldName3:
> Meanwhile i gave you a alternative option: perform stacking only when the object instance deep equals match (make script equals only if their instances are of the same source and their local variables are equal), separating the stack creates a new object with a deep clone (and you need to make sure that script variables like 'OnPCEquip' are only set after performing a stack separation when you drop a stack on the paperdoll or a repair item on the stack).

We've actually discussed something like this and it's a vague possibility (and shouldn't be something we can't get done pre-1.0 if someone has time) but the critical distinction between this and that is that the implementation in this PR fixes a bug right away, but doing something more complicated leaves the bug in-place for a while as the new thing needs to be written and tested. Merging this now has to happen whether or not a new solution will be made later.

> I believe there is a very good reason scrawl diverged from morrowind

It's much more likely that he thought it was probably fine and did it without thinking through whether or not anything would break (as otherwise the change would have been accompanied by a bunch of discussion and someone would remember that). The guy was good, but only the pope has the superpower of infallibility (and I think he might be lying about that).

i30817:
Close bug, open (at least) 3? The second paragraph of my previous response still applies, and the rationale for pushing for a decision. Half measures here just make me think that you're inching to remove the ability altogether by making a decision that opens those bugs.

wareya:
>Close bug, open (at least) 3? 

Yes.

i30817:
Well, do what you want, i made my argument.

zinnschlag:
The idea behind this PR is correct. Items with scripts can not stack. That is because each item with a script has state associated with it (the local script variables). Merging items with scripts will result in lost state. This is hopelessly irrecoverably broken. If I had caught this change by scrawl I would have vetoed it.

Stacking only if the state matches is a possibility, as mentioned before but that is better left for post-1.0 development.

I am not so sure about the implementation though. As noted the loop isn't great, especially since it works on all items, even those that do not have scripts. A simple approach would be to check if the item doesn't have a script and then use the loop-less implementation.

Just to make sure do we know exactly how vanilla handles this: Does it actually unstack if you add a stack of items with a script? Another option would be to simple reduce the number of added items to 1 and print a warning message. But of course we should try to be as compatible as possible.

Capostrophic:
Does anything else need attention?

psi29a:
Seal of approval from Zini :)

Capostrophic:
It's been two months and Zini is yet to reappear. Are there any objections?

psi29a:
Executive decision, merging.

PR #2061, opened 2018-12-07
allow release builds to have a console

sthalik:
Right now it's possible to have a console after-the-fact using `editbin`. The user ought to be able to control its presence regardless of build type. Otherwise it's necessary to use a debug build type with release-type compiler and linker flags.

It's not typical for CMake projects to dispatch on the build type, which is after all an arbitrary name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -656,12 +656,6 @@ if (WIN32)
       set_target_properties(openmw PROPERTIES LINK_FLAGS_RELWITHDEBINFO "/SUBSYSTEM:WINDOWS")
     endif()
 
-    if (BUILD_OPENMW)
-        # Release builds don't use the debug console
-        set_target_properties(openmw PROPERTIES LINK_FLAGS_RELEASE "/SUBSYSTEM:WINDOWS")
-        set_target_properties(openmw PROPERTIES LINK_FLAGS_MINSIZEREL "/SUBSYSTEM:WINDOWS")
-    endif()
-
     # Play a bit with the warning levels
 
     set(WARNINGS "/Wall") # Since windows can only disable specific warnings, not enable them


kcat:
Won't this make release builds always have a console when run? I don't think that's desirable. OpenMW already echos what it prints to openmw.log, AFAIK, so if the user needs to check anything they can look there. More in-depth bug chasing should probably be done with a debug (or relwithdebinfo) build anyway.

sthalik:
> More in-depth bug chasing should probably be done with a debug (or relwithdebinfo) build anyway.

I create debug info with my release builds. It's not a problem. My personal project also uses release builds with shipped full `.pdb`s. The optimizations haven't interfered with postmortem debugging. The console output is useful when running under an IDE that slurps stderr/stdout.

The user ought to modify the variable controlling the console's appearance rather than rely on `CMAKE_BUILD_TYPE` doing it for them.

akortunov:
We disabled background console window for release builds [for reason](https://gitlab.com/OpenMW/openmw/issues/2490) - OpenMW is a GUI application, not a command line one.
As for me, it would be better to control subsystem via CMake variable, and by default there should be WINDOWS.
So for end users OpenMW will behave as any other 3D game, but developer can configure it as he wants.

AnyOldName3:
In general, I'm not a fan of making behaviour dependent on the build type, but hiding the console in Release builds is a reasonable strategy. We've been very naughty and not hidden it in RelWithDebInfo builds (which should be identical to Release builds except for a single compiler switch), but that's only because it would have delayed stuff like shadows by a long time if I didn't have console output in a build that runs in frames-per-second rather than the other way around.

There is, however, a real issue here. Ideally, when OpenMW is manually launched via a terminal, we want std out to go to that terminal, and right now, it doesn't. This is something which we know is possible, and almost certainly has a simple solution, but no one's got around to doing it. @sthalik If you want to do something about the current situation, make that work instead.

sthalik:
@AnyOldName3 please see #2064

PR #2063, opened 2018-12-07
Add NMake support to the Windows prebuild script.

AnyOldName3:
@sthalik discovered that it's actually possible to use NMake to build OpenMW with the dependencies the prebuild script grabs. This makes that something you can do without messing around.

To use this, launch Git Bash from a terminal with command-line C++ capability (e.g. a Developer Command Prompt or regular command prompt which has run VCVarsWhatever.bat) and run the script as normal but with an extra `-n` argument.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -31,6 +31,7 @@ SKIP_EXTRACT=""
 KEEP=""
 UNITY_BUILD=""
 VS_VERSION=""
+NMAKE=""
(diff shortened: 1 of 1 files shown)


AnyOldName3:
Huh, so that doesn't work yet.

AnyOldName3:
The manual for this is now available here: https://wiki.openmw.org/index.php?title=Development_Environment_Setup#NMake_2013.2F2015.2F2017_scripted_way

PR #2064, opened 2018-12-08
enable win32 debug console when launching from console

sthalik:
Replacing older PR.


components/debug/debugging.hpp
```diff
@@ -9,6 +9,9 @@
 #include <SDL_messagebox.h>
 
 #include "debuglog.hpp"
+#if defined _WIN32
+#   include "win32.hpp"
```
AnyOldName3: This should probably be included in the `.cpp` file instead to reduce compile times for things that include this file. It doesn't make much difference here, as the header is tiny, but it's best to stick to best practices even when it doesn't matter so you stay in the habit.

components/debug/win32.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
AnyOldName3: We tend to use define-based include guards instead of `#pragma once`. It seems to be ubiquitous these days, so maybe we should reconsider it, but where not absolutely impossible, we stick to standard C++.

@zinnschlag Any opinions?

AnyOldName3:
As well as the inline comments, I'm not super happy about the `win32.hpp` file name - it's not especially descriptive and it's easy to confuse with a Windows header if you're not already aware what it's doing.

components/debug/win32.hpp
sthalik: @AnyOldName3 I'd advise to give a free hand for use of either include guard style. Of course I'm abiding by either decision.

The only way `#pragma once` fails is when the same header is included multiple times via hard links, etc. It's not treated to a warning by GNU or Clang either.

components/debug/debugging.hpp
sthalik: Would you rather have me move it to the central debug file or rename "win32" to, say, "win32console"?

AnyOldName3: At risk of sounding a bit like KenM's ["I am only certified to diagnose item deficiency"](https://i.imgur.com/0C4JHZR.jpg), I'm not sure. Someone else might have a preference.

Capostrophic: IMO it makes sense to move win32 into the central debug file, since it's a mere init method that won't be used by anything else.

sthalik:
This header-in-namespace thing is scary. It might've worked if the header was already included somewhere before that line. It's more important than this PR by all means.

components/debug/debugging.cpp
```diff
@@ -2,8 +2,42 @@
 
 #include <components/crashcatcher/crashcatcher.hpp>
 
+#ifdef _WIN32
+#   undef WIN32_LEAN_AND_MEAN
```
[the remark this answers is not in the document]
sthalik: It's equivalent to setting it only if it's undefined, and saves on preprocessor directives. Also note that the original code setting it to `1` had good intentions, but propagates misinformation. It only needs to be defined.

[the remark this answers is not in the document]
AnyOldName3: > It's equivalent to setting it only if it's undefined

It isn't. It does the opposite of that. See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-undef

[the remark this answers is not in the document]
sthalik: Having

```
#undef FOO
#define FOO
```

is the moral equivalent of

```
#ifndef FOO
#    define FOO
#endif
```

provided that the value of FOO doesn't matter, which is the case with this win32 preprocessor identifier.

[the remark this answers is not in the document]
AnyOldName3: Ignore me, somehow I missed that you were redefining it immediately after.

sthalik:
Can we merge a separate fix for header-in-namespace in the stable branch?

AnyOldName3:
If this PR does what it says it does and someone other than me says everything is fine, it's probably mergeable. There shouldn't be a need to complicate and potentially delay the process by splitting things up.

PR #2065, opened 2018-12-08
fix NMake failure at configure-time

sthalik:
Fix `OSGDB_LIBRARY` not having a singular path on Windows resulting in CMake configure failure. I'd rather use some elegant solution but there's no `IMPORTED_LOCATION` available under any name. See `share/cmake/Modules` for osg bits.

If there's some elegant solution to extract the correct library path then naturally it should be used instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,7 +256,14 @@ set(USED_OSG_PLUGINS
                     osgdb_tga
                   )
 
-get_filename_component(OSG_LIB_DIR ${OSGDB_LIBRARY} DIRECTORY)
+if(OSGDB_LIBRARY_DEBUG)
+    set(osgdb-libpath "${OSGDB_LIBRARY_DEBUG}")
+elseif(OSGDB_LIBRARY_RELEASE)
+    set(osgdb-libpath "${OSGDB_LIBRARY_RELEASE}")
+else()
+    set(osgdb-libpath "${OSGDB_LIBRARY}")
+endif()
+get_filename_component(OSG_LIB_DIR ${osgdb-libpath} DIRECTORY)
 set(OSGPlugins_LIB_DIR "${OSG_LIB_DIR}/osgPlugins-${OPENSCENEGRAPH_VERSION}")
 
 if(OSG_STATIC)


AnyOldName3:
I think I understand the problem, but have no idea how it's only hit you. How did this fail for you exactly? Were any errors printed?

sthalik:
> I think I understand the problem, but have no idea how it's only hit you. How did this fail for you exactly? Were any errors printed?

Maybe it's because there's `osgDBd.lib` detected as `OSGDB_LIBRARY_DEBUG` and having both that and the release version set up, makes `OSGDB_LIBRARY` into an array like that.

My CMake version is 3.13.1 if that changes anything.

AnyOldName3:
There's probably a problem here, but I don't think we should leap into any particular solution until we've worked out why no one else can replicate it.

Do you maybe have multiple OSGs that CMake is detecting (e.g. because you've got a system-wide one registered via an environment variable)? We know that the Find OSG script included with CMake has issues, but they say that as OSG uses CMake, it's OSG's job to fix their CMake, and OSG are mean whenever I try and talk to them, so that kind of thing isn't going to be fixed for a while. If it's detecting another OSG, it may be taking a different code path which puts different stuff into the variable we're reading.

The CMake script can be found here if you want to try and work out why it's giving you a list and other people a single path - we wouldn't want to merge this fix only to find that the problem is bigger than anticipated and we need more armour against it. https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindOpenSceneGraph.cmake

sthalik:
> There's probably a problem here, but I don't think we should leap into any particular solution until we've worked out why no one else can replicate it.

If you set `OSGDB_LIBRARY_DEBUG` to the existing `osbDBd.lib` file it's gonna happen. For me the cache variable gets set on its own.

The full invocation is:

```
% msvc64 CI/before_script.msvc.sh -n -k -V -c Release -p Win64 -v 2017
```

The particular specified MSVC version doesn't matter, as long as it's 2015 or above. Using the NMake generator slurps the compiler specifics from PATH, LIB, LIBPATH, and INCLUDE.

> Do you maybe have multiple OSGs that CMake is detecting (e.g. because you've got a system-wide one registered via an environment variable)? 

No.

> OSG are mean whenever I try and talk to them, so that kind of thing isn't going to be fixed for a while. 

Figures. I already have to maintain some forks due to either obsolescence or general incompetence. It's interesting how they don't showcase this project on their site while including some crappy ones. It's also sad using OSG without having a healthy working relationship with the OSG project.

> The CMake script can be found here if you want to try and work out why it's giving you a list

I've already read that script (and several helpers). It's because both `OSGDB_LIBRARY_RELEASE` and `OSGDB_LIBRARY_DEBUG` get set by CMake when the CI script invokes it.

> we wouldn't want to merge this fix only to find that the problem is bigger than anticipated

If anything I'd rather look for a more proper solution since my proposed fix just stinks.

AnyOldName3:
> ```
> % msvc64 CI/before_script.msvc.sh -n -k -V -c Release -p Win64 -v 2017
> ```
> The particular specified MSVC version doesn't matter, as long as it's 2015 or above. Using the NMake generator slurps the compiler specifics from PATH, LIB, LIBPATH, and INCLUDE.

The options are still required because you need the right versions of the dependencies downloaded. AFAIK, there's not an easy to use command in the MSVC tools to query this stuff, so I left it up to the user to remember what they'd clicked.

>For me the cache variable gets set on its own.

That's usually because it's been set (either manually or automatically) on a previous CMake run. Could this have happened accidentally when you were investigating something?

sthalik:
>>     The particular specified MSVC version doesn't matter, as long as it's 2015 or above. Using the NMake generator slurps the compiler specifics from PATH, LIB, LIBPATH, and INCLUDE.
> The options are still required because you need the right versions of the dependencies downloaded. 

AFAIK, there's not an easy to use command in the MSVC tools to query this stuff, so I left it up to the user to remember what they'd clicked.

I'm referring to the `case` block inside the shell script. I'm disambiguating that whether 2017 Preview or 2019 Preview is used, right tools and generators are set on my end (as long as the .sln generator isn't used). So there's no issue in that part.

>    For me the cache variable gets set on its own.
> That's usually because it's been set (either manually or automatically) on a previous CMake run. Could this have happened accidentally when you were investigating something?

No, I hosed the entire build directory beforehand.

Overall it's pretty normal that some libraries set both debug CRT and release CRT versions. See Boost for instance. I don't see why you're confused, even though I'd like to get it solved more properly than this pull request. Maybe there's a different way of reaching the plugin path without using the library directory?

How about we add a CMake variable that's checked first for the osg plugin path? Then the CI script for MSVC can pass it along and the code remains relatively clean.

That there are several red flags, at least:

- insistence that only git-cola msys2 environment works
- handling multi-generators separately, even though they require no care

is more of a long-term process.

psi29a:
@sthalik just want to say thank for sticking with this and working with us. :)

AnyOldName3:
I've done more investigation. Even though they previously said they wouldn't update the OSG finding scripts, CMake updated how they find OSG packages as there was an issue preventing the release versions of libraries ever being detected. In doing so, they changed what `OSGDB_LIBRARY` means. The comments in `FindosgDB.cmake` say
> `# OSGDB_LIBRARY - The osgDB library OSGDB_LIBRARY_DEBUG - The osgDB`
> `# debug library`

so what we were doing before was a little weird, but still valid. With the new changes, `OSGDB_LIBRARY` is now a list of all `osgDB` libs found (even though the comment saying it isn't hasn't been changed).

Presumably, we only ever `GET_FILENAME_COMPONENT`ed `OSGDB_LIBRARY` specifically because CMake wasn't finding the release versions, so it's pretty likely that the best approach would be to use the new stuff to link and abandon the old code. We'll need to fence the new behaviour behind a CMake version check, though, as relying on the fixed behaviour will obviously only work on versions where it's actually been fixed.

AnyOldName3:
We need to come up with a proper fix for this ASAP as now I've upgraded CMake, I can't actually build OpenMW any more. 🙃 

My proposed solution which should be completely backward and forward compatible, and also allow the debug and release libs to end up in different places (as that seems to be a thing that CMake think might happen) is:
* In the top-level `CMakeLists.txt`, check if `OSGDB_LIBRARY` is a list. If it isn't, just fill `OSGPlugins_LIB_DIR` the same way we're already doing, otherwise split it, and put each path in `OSGPlugins_LIB_DIR`, making that a list, too.
* In `cmake/FindOSGPlugins.cmake` we just set another variable to the value of `OSGPlugins_LIB_DIR`, and that gets in `osg_find_library`, which we call, and that gives it to CMake's `find_library` as a directory hint, and that works just fine with semicolon-separated lists. This part is already fine and requires no extra work.
* In `CMakeLists.txt`, we later use `OSGDB_LIBRARY` for some OSX-specific stuff.
  * The first time, we use the value to construct a list of the absolute paths of all OSG plugins. This could be replaced by using the plugin names to get the names of the CMake variables that actually contain the real paths (e.g. `osgdb_dds` to `OSGDB_DDS_LIBRARY`) and then use those paths directly. On CMake 13.1 or higher, that'll give us a list, but it looks like the code as-is always uses the release library, so instead of `OSGDB_DDS_LIBRARY`, use `OSGDB_DDS_LIBRARY_RELEASE` to get that specific one if its set, and fall back to the old value otherwise.
  * The second time it's used, we use it to get the plugin directory name. We generate that earlier on as `osgPlugins-${OPENSCENEGRAPH_VERSION}"`, so doing the same again will yield the same results.
  * The final time, we use it to print an error message about the second time not working. That will no longer be necessary.

This set of changes should work on CMake 3.12 and earlier and also on CMake 13 and later. The only issue is that when people upgrade to CMake 3.13, it leaves a bunch of things cached inside the macros that got changed, which makes them spit out garbage, so we might want to somehow detect that the upgrade has happened and clear them.

sthalik:
Why not wrap what I sent into a function, stuff `set(${var-name} ... PARENT_SCOPE)` in it, and put it in `cmake/FindOSGPlugins.cmake`? Either solution is fine though.

sthalik:
@AnyOldName3 unless there's some better way, #2075 should take care of the issue without the need for further changes–on Windows at least. If you can pass the library directory from outside CMake then placing it somewhere in the `else if` chain will do the trick.

`OSG_DIR` is set by CI but I'm worried that hardcoded `lib/` isn't appropriate for all OS-en.

AnyOldName3:
I don't think that PR is a great solution. It doesn't allow for the plugins to be in different locations for debug and release, and if CMake thought that situation was likely enough that they decided to update a script they'd said they were never going to touch again, it's probably likely enough that we should be ready for it.

AnyOldName3:
https://github.com/OpenMW/openmw/pull/2078 now exists. Hopefully it's perfect.

PR #2066, opened 2018-12-08
Do not read font size just for every single widget

akortunov:
For now we read "font size" setting every time when we create a widget. It is a bit annoying.
Since we can not change font size during runtime anyway, I suggest to make font size a static variable.

### DIFF
--- a/components/widgets/fontwrapper.hpp
+++ b/components/widgets/fontwrapper.hpp
@@ -14,18 +14,10 @@ namespace Gui
         virtual void setFontName(const std::string& name)
         {
             T::setFontName(name);
-            T::setPropertyOverride ("FontHeight", mFontSize);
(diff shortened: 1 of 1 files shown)


components/widgets/fontwrapper.hpp
```diff
@@ -34,11 +26,22 @@ namespace Gui
...
             if (_key == "FontName")
             {
-                T::setPropertyOverride ("FontHeight", mFontSize);
+                T::setPropertyOverride ("FontHeight", getFontSize());
             }
         }
 
-        std::string mFontSize;
+    private:
+        int clamp(const int value, const int lowBound, const int highBound)
```
Mingun: I think this function can be static too

Capostrophic:
Seems to work fine.

psi29a:
I'm on board. Thanks.

PR #2068, opened 2018-12-08
 Make normal weapon resistance behavior closer to vanilla (bug #4384)

Capostrophic:
[Bug 4384](https://gitlab.com/OpenMW/openmw/issues/4384).

Previously onHit was the only where normal weapons resistance was applied, and it only checked one weapon to determine whether to modify the damage. In case of ranged weapons, that weapon was the projectile. However, in Morrowind *both* the bow/crossbow *and* the projectile have to be normal weapons for normal weapons resistance to work. The discrepancy caused player complaints, since their daedric bows were considered ineffective for shooting scamps by the game.

To add to that, resistNormalWeapon function applied werewolf silver weapon damage multiplier for some reason (which should really be in a different function) and used a complicated check to determine whether the weapon is normal. I moved both the damage mult functionality and normal weapon check into new functions. Since applyWerewolfDamageMult can only reasonably be used with NPC victims, creature class doesn't need it.

isNormalWeapon could easily be reused in projectileHit so that Morrowind behavior could be followed. 
This has made it a necessity to also move resistNormalWeapon out of onHit methods for its more flexible usage.

Unfortunately I didn't adjust the AI other than for applyWerewolfDamageMult continued usage, because the projectile and ranged weapon combination can be arbitrary in the given frame, we don't keep track of the current ammo in rateWeapon for bows/crossbows, and because it's not really necessary for resolving the bug. Maybe later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
     Bug #3733: Normal maps are inverted on mirrored UVs
     Bug #4329: Removed birthsign abilities are restored after reloading the save
     Bug #4383: Bow model obscures crosshair when arrow is drawn
+    Bug #4384: Resist Normal Weapons only checks ammunition for ranged weapons
     Bug #4411: Reloading a saved game while falling prevents damage in some cases
     Bug #4540: Rain delay when exiting water
     Bug #4701: PrisonMarker record is not hardcoded like other markers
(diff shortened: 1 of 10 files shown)


akortunov:
A quite questionable change - I got mostly negative feedback about this change. Can we make it configurable (maybe merge with the `enchanted weapons are magical`)?

psi29a:
Holding off on this, needs more feedback/testing.

psi29a:
Right, having an option to switch behaviour helps.

PR #2071, opened 2018-12-09
Don't interrupt sneak and swim idles in attack ready state in first person view (bug #4750)

Capostrophic:
[Bug 4750](https://gitlab.com/OpenMW/openmw/issues/4750).

Apparently Morrowind doesn't do that for a reason. Sneaking breaks otherwise.
As I said in the report, the side effect is that weapon weighing animation is not interrupted by an attack during sneaking and swimming, the same thing happens in Morrowind.


terabyte25:
Seems good to me.

PR #2072, opened 2018-12-10
Navmesh tiles cache item fix (bug #4751)

elsid:


psi29a:
@elsid Can you add a bugfix entry to the changelog for issue: 4751
https://gitlab.com/OpenMW/openmw/issues/4751

Once merged, the issue can be marked as resolved/closed.

akortunov:
@psi29a, since navmeshes were added only in the 0.46, I suppose we do not need a changelog entry.

psi29a:
You're right, my excuses.

PR #2073, opened 2018-12-11
various fixes

psi29a:
makes openmw.appdata.xml a template that accepts @OPENMW_VERSION@ so we do not have to update it all the time; fix deprecated warning for boost header; un-ignore openmw.appdata.xml


files/openmw.appdata.xml
```diff
@@ -43,7 +43,7 @@ Copyright 2017 Bret Curtis <psi29a@gmail.com>
   <category>RolePlaying</category>
  </categories>
  <releases>
-  <release version="0.46.0" date="2017-12-05"/>
+  <release version="@OPENMW_VERSION@" date="2017-12-05"/>
```
Mingun: Well, you still need update the `date` field manually so this change is not too useful

apps/launcher/graphicspage.cpp
```diff
@@ -1,6 +1,6 @@
 #include "graphicspage.hpp"
 
-#include <boost/math/common_factor.hpp>
+#include <boost/integer/common_factor.hpp>
```
akortunov: Looks like this change leads to compilation errors.

akortunov:
And since we chage the openmw.appdata.xml here anyway, I suggest to:
1. Update screenshots (especially the launcher one, which is from 0.11 release).
2. Update link to bugtracker (https://gitlab.com/OpenMW/openmw/issues)

files/openmw.appdata.xml
psi29a: @Mingun it's better than nothing. ;)

@akortunov release date is tricky, I would peg release_date to the day it was tagged. Then it would be yet another thing for someone who is about to tag a release to tag. Could we somehow pull the date from the last commit in git?

apps/launcher/graphicspage.cpp
psi29a: Yeah, math/common_factor.hpp is deprecated in later versions of boost. I can revert this for now, just wanted to get rid of issue. I'll see if we can find some common ground first. (Hence this PR anyway.)

psi29a:
> 1. Update screenshots, which may not reflect the current OpenMW state (especially the launcher one, which is from 0.11 release).

If someone wants to do this, feel free. My PR was to focus only on automating version and feature creep to date. The content itself I'm leaving alone.

> 2. Update the link to bugtracker (https://gitlab.com/OpenMW/openmw/issues)

I'll update this however. Thanks.

Capostrophic:
It's probably worth it to squash the commits for more clean git history (5 commits to adjust a few lines in a single file the names of which are not very informative is a little too much).

psi29a:
squash and merge is an option during a PR along with rebase and merge.

zinnschlag:
Looks good to me.

PR #2074, opened 2018-12-12
[abandoned] allow more generator setup in ci/msvc

sthalik:
Unconditionally use `CodeBlocks - NMake Makefiles` rather than `NMake Makefiles` since that's required by at least the CLion IDE. It used to be required by old versions of Qt Creator. Any IDE that hasn't yet implemented the "CMake server mode" protocol depends on the CodeBlocks aux generator.

Allow for using any arbitrary generator through the `-G` switch. This helps with IDEs that support arbitrary generators. For one, `Unix Makefiles` is unable to fully saturate the CPU due to forking off unnecessary processes and processing rules suboptimally. With `Ninja` I'm able to keep the cores fully saturated till it's done. This is more directly evident on projects with many trivial targets, but openmw as well.

While at it, add a Vim modeline since `:se noet` gets old after a while.

This affects building inside IDEs like Qt Creator or CLion. I can drop the second (and third) commit from the series but there should be no harm in keeping them.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -313,7 +313,7 @@ if [ ${BITS} -eq 64 ]; then
 fi
 
 if [ -n "$NMAKE" ]; then
-	GENERATOR="NMake Makefiles"
+	GENERATOR="CodeBlocks - NMake Makefiles"
 fi
 
 add_cmake_opts "-G\"$GENERATOR\""
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -32,6 +32,7 @@ KEEP=""
 UNITY_BUILD=""
 VS_VERSION=""
 NMAKE=""
+FORCE_GENERATOR=""
 PLATFORM=""
 CONFIGURATION=""
 
@@ -68,7 +69,12 @@ while [ $# -gt 0 ]; do
 				shift ;;
 
 			n )
-				NMAKE=true ;;
+				FORCE_GENERATOR="CodeBlocks - NMake Makefiles"
+				NMAKE=1 ;;
+
+			G )
+				FORCE_GENERATOR="$1"
+				shift ;;
 
 			p )
 				PLATFORM=$1
@@ -100,6 +106,8 @@ Options:
 		Choose the Visual Studio version to use.
 	-n
 		Produce NMake makefiles instead of a Visual Studio solution.
+	 G
+		Force using a specific CMake generator.
 	-V
 		Run verbosely
 EOF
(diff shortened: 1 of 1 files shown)


AnyOldName3:
It looks to me like this assumes any forced generator will be a single-configuration one, but I'm pretty sure CMake can spit out a multi-config Qt Creator project, too.

AnyOldName3:
Can you explain the Vim modeline? I've found enough to know it means `noexpandtab`, `softtabstop=0`, `tabstop=4` and `shiftwidth=4`, but what exactly do those mean?

I'm not even sure this file uses a consistent indentation/alignment style, so I'm not sure how helpful setting this even is. If it *is* worth doing, if we're using this whitespace style in all the prebuild scripts, it's probably better to set it up in the `.editorconfig` file as several text editors and IDEs have native support for that (instead of just Vim) and basically any decent text editor has a plugin for it these days (including Vim).

sthalik:
> t looks to me like this assumes any forced generator will be a single-configuration one, but I'm pretty sure CMake can spit out a multi-config Qt Creator project, too.

CMake doesn't create `qmake`-based projects. Qt Creator uses CMake projects natively now. It's been gradually improved upon since 4.5 or 4.6. Now it uses CMake server mode.

I'd be greatly inclined to assume that almost all generator types are single-config. Furthermore, other multi-config generators (iCode? 😉) may (or may not) use other directory structure so they won't work without further adjustments to the CI script and the build process anyway. Otherwise how about slapping a `case/esac` block for known multi-config generators, deciding where things get copied?

I'd however recommend doing DLL copying (if that's the only problem) as a build target rather than through CI directly. That's universal for all generators, and actually solves the problem of running directly from the build directory.

As a side note, I think using `OpenMW_{SOURCE,BINARY}_DIR` stems from misunderstanding `CMAKE_{{CURRENT_,}BINARY,SOURCE}_DIR`. The latter are less vague and appropriate from my build system snafu experience.

sthalik:
> Can you explain the Vim modeline? I've found enough to know it means noexpandtab, softtabstop=0,
> tabstop=4 and shiftwidth=4, but what exactly do those mean?

It's so that the file keeps tabs as indentation and for further-indented lines. `shiftwidth` is for `<>`. `tabstop` is how long the tabs appear on the terminal in the first place. I tried keeping the original file's indentation but that requires this complex invocation. I don't have either preference, either mixing tabs and spaces or `:retab` is fine to me. But note that `<>` will change tabs to spaces and vice versa depending on `expandtab` value.

> It looks to me like this assumes any forced generator will be a single-configuration one, but I'm pretty sure CMake can spit out a multi-config Qt Creator project, too.

Better yet, since CMake can detect generator and compiler type (say, MSVC multi- or not), we can gradually introduce copying from CI rather than in one sweeping change across all CI/generator stuff. That also fixes the installer bits if the dll files are copied from the source directory.

AnyOldName3:
> Furthermore, other multi-config generators (iCode? 😉) may (or may not) use other directory structure so they won't work without further adjustments to the CI script and the build process anyway.

I'm pretty sure that the directories are set by CMake rather than Visual Studio, so it should be the same for all multi-config generators.

> I'd however recommend doing DLL copying (if that's the only problem) as a build target rather than through CI directly. That's universal for all generators, and actually solves the problem of running directly from the build directory.

Doing this might be a good idea as long as it's a build step and not an install step.

psi29a:
@AnyOldName3 you feel like this is ready or needs more work?

AnyOldName3:
This was two months ago, so I don't really have any idea any more. I can see that I at least started a code review and then either forgot about it part way through or only had specific comments about one line and just didn't submit it.

I think a lot of the disagreement and discussion hasn't been resolved yet, so just merging it as-is is probably a bad idea.

PR #2076, opened 2018-12-12
factorize buffer object of all terrain tile

mp3butcher:
it can't hurt


akortunov:
Causes crashes for me on OSG 3.6.0 from OpenMW repo almost immediately when the Distant Terrain feature is enabled:
[openmw-crash.log](https://github.com/OpenMW/openmw/files/2672923/openmw-crash.log)

Note: stack trace is the pretty the same as in [this](https://github.com/openscenegraph/OpenSceneGraph/issues/655) issue.

mp3butcher:
will see that

mp3butcher:
@akortunov 
distant terrain seems to leak memory
[Terrain]
 distant terrain = true
[Cells]
pointers cache size = 666666
I reproduce your strange crash in debug and saw that there's mess in memory

PR #2077, opened 2018-12-12
Sanity check for text input in toggleWalking

terabyte25:
s a sanity check needed for text input in non-standard GUI windows such as the chat in TES3MP


davidcernat:
It makes sense to me.

PR #2078, opened 2018-12-13
Adapt to CMake 3.13's new meaning of OSGDB_LIBRARY

AnyOldName3:
Adapt to CMake 3.13's new meaning of OSGDB_LIBRARY (i.e. that it can now be a list) while allowing for the possibility that the found libraries may be in different directories when debug and optimised versions exist.

This is a third attempt at fixing the same thing as https://github.com/OpenMW/openmw/pull/2065 and https://github.com/OpenMW/openmw/pull/2075, but should be more flexible.

This should be backwards compatible with CMake 3.12.3 and earlier, but I've only actually tested it with 3.13.1. Also, the OSX bundling stuff hasn't been tested on a real Mac, but I've tested that the same code does what I expect it to on Windows when I put it in a different bit of the file. I don't know if Travis tests that, but if not, someone with a Mac should.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,8 +256,14 @@ set(USED_OSG_PLUGINS
                     osgdb_tga
                   )
 
-get_filename_component(OSG_LIB_DIR ${OSGDB_LIBRARY} DIRECTORY)
-set(OSGPlugins_LIB_DIR "${OSG_LIB_DIR}/osgPlugins-${OPENSCENEGRAPH_VERSION}")
+set(OSGPlugins_LIB_DIR "")
+foreach(OSGDB_LIB ${OSGDB_LIBRARY})
+	# Skip library type names
+	if(EXISTS ${OSGDB_LIB})
+		get_filename_component(OSG_LIB_DIR ${OSGDB_LIB} DIRECTORY)
+		list(APPEND OSGPlugins_LIB_DIR "${OSG_LIB_DIR}/osgPlugins-${OPENSCENEGRAPH_VERSION}")
+	endif()
+endforeach(OSGDB_LIB)
 
 if(OSG_STATIC)
     add_definitions(-DOSG_LIBRARY_STATIC)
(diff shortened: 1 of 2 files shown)


AnyOldName3:
Well a good thing is that the CI *does* test the Mac bundling, but I'm going to need someone who knows Macs to tell me what various types of library are and how they work. @Thunderforge, can you lend a hand here?

Thunderforge:
@AnyOldName3 Unfortunately, I don't think I'm going to be much help because I get a black screen when I compile OpenMW (I filed this issue in the [prebuilt dependencies](https://github.com/OpenMW/openmw-deps-mac/issues/45) repo), and so I won't know if any changes I make actually work. Custom-building the Mac dependencies result in some compile failures that I haven't been able to resolve.

@nikolaykasyanov I don't suppose this issue is easily testable by you?

sthalik:
> My first question is if find_package is capable of finding .so files on a Mac (under the hood it calls find_library)

You can link against `.dylib` files. Then you learn about `otool -L` and `install_name_tool` and wish you never dealt with redistributable OSX apps in the first place.

AnyOldName3:
That doesn't really answer the question. As part of the OSX bundling process, the `.so` files for the OSG plugins are copied from a subdirectory of OSG's lib directory to the bundle. With the new change, the lib directory is now a list of directories, but if the library finding code works on `.so` files, they can be grabbed by CMake and copied from where it finds them.

AnyOldName3:
Even if this new version passes CI, don't merge it. I've had a thought, but need to see what all the different values on all the different OSes are.

Basically, when making this PR, I commented out some stuff that caused warnings to be printed, but I'm thinking maybe this was actually necessary on OSX.

AnyOldName3:
If you're not getting DDS load errors, it's fixed. Now I just need to figure out how to make it work without causing warnings when linking against static OSG on Windows (even though no one cares about that as it's hard enough to link against regular OSG on Windows).

AnyOldName3:
I've decided that I'm actually going to ignore the Windows-only warnings with static OSG as it seems like the issue is caused by changing a variable from an empty string to a different empty string is the cause of the warning, and that's stupid. As the 'problem' was present before this PR, the fix doesn't need to be in this PR.

AnyOldName3:
Unless someone objects, I'm going to merge this as soon as CI completes. I'd really like to be able to build OpenMW again.

AnyOldName3:
I'm seeing CMake errors while Travis is assembling the bundle on OSX, but I'm also seeing them in other PRs, so I'm going to assume they're unrelated.

PR #2080, opened 2018-12-15
Recharge items outside of player inventory

akortunov:
Fixes [bug #4077](https://gitlab.com/OpenMW/openmw/issues/4077).

The main idea:
1. Implement cached lists of recharging items for CellStore, as for InventoryStore.
2. Recharge items in these lists every time when we advance time.
3. If we advance time for a large amount of time, assume that player is resting or training and we need to update items in all loaded cells. Otherwise update items only for active cells.

So basically in this approach I update items in active cells every frame and update items in other loaded cells during rest, travelling, training, etc.
There should not be noticable performance loss since lists are cached.
This approach should be superior to Morrowind behaviour which updates only items, actually used by player.

Basically this PR should work now and it is ready for testing.
If used approach is fine, I will refactor it a bit to reduce a code duplication.

Note: Morrowind recharges items using a real game time instead of world time, so even with `timescale = 0` the game recharges items as usual. I replicated this behaviour too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
     Bug #3778: [Mod] Improved Thrown Weapon Projectiles - weapons have wrong transformation during throw animation
     Bug #3812: Wrong multiline tooltips width when word-wrapping is enabled
     Bug #3894: Hostile spell effects not detected/present on first frame of OnPCHitMe
+    Bug #4077: Enchanted items are not recharged if they are not in the player's inventory
     Bug #4202: Open .omwaddon files without needing toopen openmw-cs first
     Bug #4240: Ash storm origin coordinates and hand shielding animation behavior are incorrect
     Bug #4270: Closing doors while they are obstructed desyncs closing sfx
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -287,16 +287,6 @@ namespace MWMechanics
         mWatched = ptr;
     }
 
-    void MechanicsManager::advanceTime (float duration)
-    {
-        // Uses ingame time, but scaled to real time
-        const float timeScaleFactor = MWBase::Environment::get().getWorld()->getTimeScaleFactor();
-        if (timeScaleFactor != 0.0f)
-            duration /= timeScaleFactor;
-        MWWorld::Ptr player = getPlayer();
-        player.getClass().getInventoryStore(player).rechargeItems(duration);
-    }
-
     void MechanicsManager::update(float duration, bool paused)
     {
         if(!mWatched.isEmpty())
--- a/apps/openmw/mwmechanics/recharge.cpp
+++ b/apps/openmw/mwmechanics/recharge.cpp
@@ -1,0 +1,92 @@
+#include "recharge.hpp"
+
+#include <components/misc/rng.hpp>
+
+#include "../mwbase/world.hpp"
+#include "../mwbase/environment.hpp"
+#include "../mwbase/windowmanager.hpp"
+
+#include "../mwworld/containerstore.hpp"
+#include "../mwworld/class.hpp"
+#include "../mwworld/esmstore.hpp"
+
+#include "creaturestats.hpp"
+#include "actorutil.hpp"
+
+namespace MWMechanics
+{
+
+bool rechargeItem(const MWWorld::Ptr &item, const float maxCharge, const float duration)
+{
+    float charge = item.getCellRef().getEnchantmentCharge();
+    if (charge == -1 || charge == maxCharge)
+        return false;
+
+    static const float fMagicItemRechargePerSecond = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find(
+                "fMagicItemRechargePerSecond")->mValue.getFloat();
+
+    item.getCellRef().setEnchantmentCharge(std::min(charge + fMagicItemRechargePerSecond * duration, maxCharge));
+    return true;
+}
+
+bool rechargeItem(const MWWorld::Ptr &item, const MWWorld::Ptr &gem)
+{
+    if (!gem.getRefData().getCount())
+        return false;
+
+    MWWorld::Ptr player = MWMechanics::getPlayer();
+    MWMechanics::CreatureStats& stats = player.getClass().getCreatureStats(player);
+
+    float luckTerm = 0.1f * stats.getAttribute(ESM::Attribute::Luck).getModified();
+    if (luckTerm < 1 || luckTerm > 10)
+        luckTerm = 1;
+
+    float intelligenceTerm = 0.2f * stats.getAttribute(ESM::Attribute::Intelligence).getModified();
+
+    if (intelligenceTerm > 20)
+        intelligenceTerm = 20;
+    if (intelligenceTerm < 1)
+        intelligenceTerm = 1;
+
+    float x = (player.getClass().getSkill(player, ESM::Skill::Enchant) + intelligenceTerm + luckTerm) * stats.getFatigueTerm();
+    int roll = Misc::Rng::roll0to99();
+    if (roll < x)
+    {
+        std::string soul = gem.getCellRef().getSoul();
+        const ESM::Creature *creature = MWBase::Environment::get().getWorld()->getStore().get<ESM::Creature>().find(soul);
+
+        float restored = creature->mData.mSoul * (roll / x);
+
+        const ESM::Enchantment* enchantment = MWBase::Environment::get().getWorld()->getStore().get<ESM::Enchantment>().find(
+                    item.getClass().getEnchantment(item));
+        item.getCellRef().setEnchantmentCharge(
+            std::min(item.getCellRef().getEnchantmentCharge() + restored, static_cast<float>(enchantment->mData.mCharge)));
+
+        MWBase::Environment::get().getWindowManager()->playSound("Enchant Success");
+
+        player.getClass().getContainerStore(player).restack(item);
+    }
+    else
+    {
+        MWBase::Environment::get().getWindowManager()->playSound("Enchant Fail");
+    }
+
+    player.getClass().skillUsageSucceeded (player, ESM::Skill::Enchant, 0);
+    gem.getContainerStore()->remove(gem, 1, player);
+
+    if (gem.getRefData().getCount() == 0)
+    {
+        std::string message = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("sNotifyMessage51")->mValue.getString();
+        message = Misc::StringUtils::format(message, gem.getClass().getName(gem));
+
+        MWBase::Environment::get().getWindowManager()->messageBox(message);
+
+        // special case: readd Azura's Star
+        if (Misc::StringUtils::ciEqual(gem.get<ESM::Miscellaneous>()->mBase->mId, "Misc_SoulGem_Azura"))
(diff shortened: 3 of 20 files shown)


wareya:
I liked the timestamp idea for unloaded items, personally.

akortunov:
Looks like we totally forgot about this PR.
@Capostrophic, @AnyOldName3, @psi29a, what do you think?

psi29a:
We need more testers here then. :)

apps/openmw/mwworld/cellstore.cpp
```diff
@@ -1024,4 +1061,83 @@ namespace MWWorld
...
+            Ptr ptr = getCurrentPtr(it->first);
+
+            if (ptr.getCellRef().getEnchantmentCharge() == -1
+                    || ptr.getCellRef().getEnchantmentCharge() == it->second)
+                continue;
+
+            static float fMagicItemRechargePerSecond = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find(
+                        "fMagicItemRechargePerSecond")->mValue.getFloat();
+
+            if (ptr.getCellRef().getEnchantmentCharge() <= it->second)
```
Capostrophic: This comparison is always true. Or at least the "equal" part of the "less or equal" is redundant.

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -406,6 +410,57 @@ MWWorld::ContainerStoreIterator MWWorld::ContainerStore::addNewStack (const Cons
...
+    for (TRechargingItems::iterator it = mRechargingItems.begin(); it != mRechargingItems.end(); ++it)
+    {
+        if (it->first->getCellRef().getEnchantmentCharge() == -1
+                || it->first->getCellRef().getEnchantmentCharge() == it->second)
+            continue;
+
+        static float fMagicItemRechargePerSecond = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find(
+                    "fMagicItemRechargePerSecond")->mValue.getFloat();
+
+        if (it->first->getCellRef().getEnchantmentCharge() <= it->second)
```
Capostrophic: Same thing here.

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -406,6 +410,57 @@ MWWorld::ContainerStoreIterator MWWorld::ContainerStore::addNewStack (const Cons
...
+                        enchantmentId);
+            if (!enchantment)
+            {
+                Log(Debug::Warning) << "Warning: Can't find enchantment '" << enchantmentId << "' on item " << it->getCellRef().getRefId();
+                continue;
+            }
+
+            if (enchantment->mData.mType == ESM::Enchantment::WhenUsed
+                    || enchantment->mData.mType == ESM::Enchantment::WhenStrikes)
+                mRechargingItems.push_back(std::make_pair(it, static_cast<float>(enchantment->mData.mCharge)));
```
Capostrophic:
```suggestion
                mRechargingItems.emplace_back(it, static_cast<float>(enchantment->mData.mCharge));
```

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -406,6 +410,57 @@ MWWorld::ContainerStoreIterator MWWorld::ContainerStore::addNewStack (const Cons
...
+}
+
+void MWWorld::ContainerStore::updateRechargingItems()
+{
+    mRechargingItems.clear();
+    for (ContainerStoreIterator it = begin(); it != end(); ++it)
+    {
+        if (it->getClass().getEnchantment(*it) != "")
+        {
+            std::string enchantmentId = it->getClass().getEnchantment(*it);
```
Capostrophic: It's minor, but it'd look a bit better (arguably) like this:
```cpp
std::string enchantmentId = it->getClass().getEnchantment(*it);
if (!enchantmentId.empty())
// ...
```

apps/openmw/mwworld/cellstore.cpp
```diff
@@ -1024,4 +1061,83 @@ namespace MWWorld
...
+        const ESM::Enchantment* enchantment = MWBase::Environment::get().getWorld()->getStore().get<ESM::Enchantment>().search(enchantmentId);
+        if (!enchantment)
+        {
+            Log(Debug::Warning) << "Warning: Can't find enchantment '" << enchantmentId << "' on item " << ptr.getCellRef().getRefId();
+            return;
+        }
+
+        if (enchantment->mData.mType == ESM::Enchantment::WhenUsed
+                || enchantment->mData.mType == ESM::Enchantment::WhenStrikes)
+            mRechargingItems.push_back(std::make_pair(ptr.getBase(), static_cast<float>(enchantment->mData.mCharge)));
```
Capostrophic: Same emplace_back suggestion here.

It'd be nice to get rid of code duplication in general.

apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
```diff
@@ -1,4 +1,5 @@
 #include "mechanicsmanagerimp.hpp"
+#include <iostream>
```
Capostrophic: Redundant include.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -3264,6 +3263,29 @@ namespace MWWorld
         mCells.rest();
     }
 
+    void World::rechargeItems(double hours)
+    {
+        // Convert timespan in game hours to real time seconds.
+        // If the difference is more than one hours, assume that player is resting, travelling, in prison, etc.
+        // In this case we will recharge items in all loaded cells, otherwise update only items in active cells.
+        bool activeOnly = hours < 1.f;
+        float duration = (hours * 3600) / getTimeScaleFactor();
```
Capostrophic: There should be a guarantee that getTimeScaleFactor isn't 0.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -3264,6 +3263,29 @@ namespace MWWorld
...
+
+        MWWorld::Ptr player = getPlayerPtr();
+        player.getClass().getInventoryStore(player).rechargeItems(duration);
+
+        if (activeOnly)
+        {
+            const Scene::CellStoreCollection& active = mWorldScene->getActiveCells();
```
Capostrophic: Why not just do
```cpp
for (auto &cell : mWorldScene->getActiveCells())
```

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -3264,6 +3263,29 @@ namespace MWWorld
         mCells.rest();
     }
 
+    void World::rechargeItems(double hours)
+    {
+        // Convert timespan in game hours to real time seconds.
+        // If the difference is more than one hours, assume that player is resting, travelling, in prison, etc.
+        // In this case we will recharge items in all loaded cells, otherwise update only items in active cells.
```
Capostrophic: than one hour
we assume
we recharge
we update
the items

Capostrophic:
Sorry, forgot about this: 
>If used approach is fine, I will refactor it a bit to reduce a code duplication.

I guess it's fine..?

apps/openmw/mwworld/cellstore.cpp
akortunov: I just did not touch the logic - we use the same check now.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -3285,6 +3283,31 @@ namespace MWWorld
...
     }
 
+    void World::rechargeItems(double hours)
+    {
+        // Convert timespan in game hours to real time seconds.
+        // If the difference is more than one hours, we assume that player is resting, travelling, in prison, etc.
+        // In this case we recharge items in all loaded cells, otherwise we update only the items in active cells.
+        bool activeOnly = hours < 1.f;
+        float duration = hours * 3600;
+        const float timeScaleFactor = MWBase::Environment::get().getWorld()->getTimeScaleFactor();
```
Capostrophic: You don't need to recover world to call this function.

akortunov:
The only thing left is a code duplication, I guess (if we will not decide to use timestamps, of course).

akortunov:
OK, I moved duplicated recharge logic to the new recharge.cpp file.

If there are no ideas how to better implement this feature, I claim this feature completed.
It is a time for a proper testing and review then.

apps/openmw/mwmechanics/recharge.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+    if (!gem.getRefData().getCount())
+        return false;
+
+    MWWorld::Ptr player = MWMechanics::getPlayer();
+    MWMechanics::CreatureStats& stats = player.getClass().getCreatureStats(player);
+
+    float luckTerm = 0.1f * stats.getAttribute(ESM::Attribute::Luck).getModified();
+    if (luckTerm < 1 || luckTerm > 10)
+        luckTerm = 1;
```
Capostrophic: Since everything is being moved anyway, maybe copying vanilla oversights like this one should be reconsidered?

apps/openmw/mwmechanics/recharge.cpp
```diff
@@ -0,0 +1,92 @@
...
+namespace MWMechanics
+{
+
+bool rechargeItem(const MWWorld::Ptr &item, const float maxCharge, const float duration)
+{
+    float charge = item.getCellRef().getEnchantmentCharge();
+    if (charge == -1 || charge == maxCharge)
+        return false;
+
+    static float fMagicItemRechargePerSecond = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find(
```
Capostrophic: This should either be "const float" or "static const float" because if it's static but not const the overhead of the record recovery will remain with the GMST remaining in memory for no reason and if we care about making GMSTs not read-only in future leaving it "const float" is enough.

Capostrophic:
You asked me for feedback but I'm not one of the people who argued against the approach of recharging items in the entire game world and I don't feel like thinking too deeply about it.

Capostrophic:
Getting rather stagnant now. Anyone has any thoughts on the approach and the implementation?

akortunov:
> Anyone has any thoughts on the approach and the implementation?

I suppose that no one reads the 9 months-old PR thread.

akortunov:
About timestamps approach: Morrowind uses a real time to update items instead of world time.
Now we do the same, and when we fast-forward time (e.g. when resting), we use a _current_ timescale to convert world hours to real seconds.
With such approach, timestamp-based fast-forwarding provides incorrect results when mods change current timescale or current timestamp. We already have the same issue with spells (see [bug #5165](https://gitlab.com/OpenMW/openmw/issues/5165)), and I see no reasons to make things worse.

CHANGELOG.md
```diff
@@ -17,10 +17,12 @@
     Bug #3778: [Mod] Improved Thrown Weapon Projectiles - weapons have wrong transformation during throw animation
     Bug #3812: Wrong multiline tooltips width when word-wrapping is enabled
     Bug #3894: Hostile spell effects not detected/present on first frame of OnPCHitMe
+    Bug #4077: Enchanted items are not recharged if they are not in the player's inventory
     Bug #4202: Open .omwaddon files without needing toopen openmw-cs first
     Bug #4240: Ash storm origin coordinates and hand shielding animation behavior are incorrect
     Bug #4270: Closing doors while they are obstructed desyncs closing sfx
     Bug #4276: Resizing character window differs from vanilla
+    Bug #4077: Enchanted items are not recharged if they are not in the player's inventory
```
Capostrophic: Oops.

Capostrophic:
Last call!

PR #2081, opened 2018-12-16
Remove unnecessary Recastnavigation headers & static libs from macOS package

nikolaykasyanov:
I suspect other platforms don't need them packaged too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -604,7 +604,7 @@ set(RECASTNAVIGATION_DEMO OFF CACHE BOOL "Do not build RecastDemo")
 set(RECASTNAVIGATION_STATIC ON CACHE BOOL "Build recastnavigation static libraries")
 set(RECASTNAVIGATION_TESTS OFF CACHE BOOL "Do not build recastnavigation tests")
 
-add_subdirectory (extern/recastnavigation)
+add_subdirectory (extern/recastnavigation EXCLUDE_FROM_ALL)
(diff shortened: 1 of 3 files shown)


nikolaykasyanov:
I have a fix, I just want to make sure that the check script does indeed fail when there are unexpected files. Unfortuatenly, Travis CI is experiencing issues, let's wait a bit.

CMakeLists.txt
```diff
@@ -604,7 +604,7 @@ set(RECASTNAVIGATION_DEMO OFF CACHE BOOL "Do not build RecastDemo")
 set(RECASTNAVIGATION_STATIC ON CACHE BOOL "Build recastnavigation static libraries")
 set(RECASTNAVIGATION_TESTS OFF CACHE BOOL "Do not build recastnavigation tests")
 
-add_subdirectory (extern/recastnavigation)
+add_subdirectory (extern/recastnavigation EXCLUDE_FROM_ALL)
```
psi29a: What does this do specifically? EXCLUDE_FROM_ALL?

nikolaykasyanov: As far as I understand, it excludes subdirectory contents from the default build & install targets. As OpenMW target references it explicitly, it would still be built but not installed.

https://cmake.org/cmake/help/v3.13/prop_dir/EXCLUDE_FROM_ALL.html
https://knzl.at/cmake-exclude-from-all/

psi29a: Noice. :D

PR #2082, opened 2018-12-16
Make magic teleportation attempts detectable even when teleportation is disabled (bug #3765)

Capostrophic:
[Bug 3765](https://gitlab.com/OpenMW/openmw/issues/3765)

Add teleportation effects to the list even if they're not going to work due to teleportation being disabled. Disabled teleportation handling was moved from checkEffectTarget to applyInstantEffect. This allows them to be detected by scripting, and also allows teleportation magic VFX to play properly even if teleportation is disabled, which is what I observe in Morrowind.

Also I reduced Intervention effects code duplication.

Works in my testing of a script that simply checks ```player->getEffect sEffectMark``` every frame to show a messagebox:
screenshot033


terabyte25:
H

PR #2083, opened 2018-12-18
Fix animation bug with VAOs

AnyOldName3:
Hopefully, this should fix the issue @mp3butcher reported. It's been discussed thoroughly at https://github.com/OpenMW/openmw/pull/2054, but there's no issue report for it yet, so one of those should be made before this is merged so it doesn't get left off the changelog.

### DIFF
--- a/components/sceneutil/morphgeometry.cpp
+++ b/components/sceneutil/morphgeometry.cpp
@@ -2,6 +2,8 @@
 
 #include <cassert>
 
+#include <osg/Version>
+
 namespace SceneUtil
 {
 
@@ -175,6 +177,10 @@ void MorphGeometry::cull(osg::NodeVisitor *nv)
     }
 
     positionDst->dirty();
+
+#if OSG_MIN_VERSION_REQUIRED(3, 5, 6)
+    geom.dirtyGLObjects();
+#endif
 
     nv->pushOntoNodePath(&geom);
     nv->apply(geom);
--- a/components/sceneutil/riggeometry.cpp
+++ b/components/sceneutil/riggeometry.cpp
@@ -3,6 +3,8 @@
 #include <stdexcept>
 #include <cstdlib>
 
+#include <osg/Version>
+
 #include <components/debug/debuglog.hpp>
 
 #include "skeleton.hpp"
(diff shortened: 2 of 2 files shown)


mp3butcher:
I think You have to determine the bug responsibility:
osg or openmw
openmw as client code may not have to interact with globject
as a regular osg contributor i'm against this commit as it's a hack

And What's your problem setting datavariance of drawable?!
If you've found a bug share it with osg community

AnyOldName3:
In much the same way as how if we'd been using display lists for animated things instead of VBOs, we'd have had to call `dirtyDisplayLists` to update the display lists, we now have to call `dirtyGLObjects` to update the VAOs. That call was added specifically for this exact type of situation. It only wasn't necessary to dirty the VBOs because OSG automatically did that when we dirtied the arrays. OSG added a new feature, and they added a new function to make it do certain things. OpenMW itself wasn't specifically requesting to use the new feature as it hadn't been made to handle the new feature, but it was being forced by an environment variable. There was no bug in either piece of software, per se, but OpenMW was missing compatibility with a new OSG feature, and was being forced to use that feature anyway.

I'm at least 95% sure that this isn't a hack, but is actually how OSG intends VAOs get used with geometry that gets updated without being dynamic.

The problem with using dynamic drawables is that, whatever threading model is in use, OSG won't let the next advance, update or cull traversal start until the current draw traversal has drawn all its dynamic drawables. This totally kills OpenMW's threading. Instead, we use update and cull callbacks to double-buffer drawables and statesets that get updated so that one can be updated while the other is being drawn, without any need to serialise anything that could otherwise have been multithreaded.

As I've said, there's (probably) no *bug*, but even if there was, I'd be reluctant to share it with OSG as Robert's been excessively abusive most times I've interacted with him.

mp3butcher:
be aware contributing to LGPL is not GPL... as grey is not white
It require some ethic and resiliance
osg is a software dependancy for lot of companies exploiting it for black and white stuff.
Robert can be rude but he's someone very reliable in osg community (until i began vsg:/)

AnyOldName3:
If you think there's an OSG bug here, you can make an issue report for it, and I can add some clarifications. If Robert agrees it's a bug, I won't merge this PR and I'll let him fix it. I don't think it's likely, though.

mp3butcher:
@AnyOldName3 apologies for being  dumb with vas stuff  
openmw don't operate like i do but it work quite well...
In a more general point of view, do openmw restricted to GL2 or is there anyway more modern stuff could used?

AnyOldName3:
OpenMW should always *work* on GL2, but if certain extensions are available, extra behaviour can be added that needs enabling via settings. It's not exactly the same thing, but we also have the option of being completely fixed-function provided you don't turn on shadows or use normal maps or anything like that.

PR #2084, opened 2018-12-18
[WIP] OSG PSSM shadow For SunLight

mp3butcher:
Parallel Split Shadow Mapping (fitted for Sun-outdoor- not for other sources unless very far)
test it against any osg version it should work
please give detailed feedbacks(with osg version) if it fails (but no crashlog it's too big)

### edit:
arbitrary choices made:
- shadowing op in shading is a COLout=mix(ambiantbias*COLin,COLin,shadowedratio), replace that if you're not happy
- shadowing shading is controlled with 2 extra StatSet components provided by shadowmanager ( a static unshadowed uniform and a  shadowed stateset to reuse shadowmaps)

fixes:
- strange behavior at night fixed by disabling backface culling : terrain was culled by shadowcams
- flickering shadows was related to strange numeric problem(sorry to misslead you with my previous mt  or multipass implication) fixed removing "Transform on LightSource" pssm feature
- Indoor sun disable

known bugs: 
- an numeric problem persists in shadow camera matrices in few light configurations (Warnings)

capture d ecran de 2018-12-23 16-51-58


Capostrophic:
> TODO test compatibility with openmw osg

It doesn't work well.
screenshot037

Even then, you haven't put any default values for the used settings, and I don't know what's supposed to be sensible. If you recover a non-declared setting anywhere, OpenMW will deem it as a game-breaking issue and close with a warning to the user. Putting placeholder zeroes to all of them "works" but see above.

mp3butcher:
[Shadows]
pssm shadow ambient= 0.7
pssm polygon offset units = -1
pssm polygon offset factor = 0.1
pssm distlight = 10000
pssm texunitoffset = 2
enable shadows = true
enable indoor shadows = true
number of shadow maps = 3
pssm textures resolution = 1024

JDGBOLT:
Just so you know, the Mesa Open source drivers hate your shaders. One error is at https://github.com/OpenMW/openmw/pull/2084/files#diff-6795e8e42f438d7482eb325d24479096R82 , this needs to be 1.0, as it complains that it's returning an Int. The other it's complaining about:
VERTEX glCompileShader "18" FAILED
VERTEX Shader "18" infolog:
0:10(2): error: unsized array index must be constant

Still trying to find out exactly what it's talking about. Since I think you mentioned you are using an Nvidia card, those are horrible to develop shaders on, as they let you get away with a lot of things that more strict shader compiler won't accept, such as the Mesa one.

This is what the shadows currently look like on my system, which has an AMD Fury card running the latest Mesa drivers: https://streamable.com/3hkmv

mp3butcher:
mmh i may not have been strict enough with glsl will try to findout too

JDGBOLT:
All right, I think I found the problem line, after going and commenting out various lines to try to find it.
It's complaining about this line: https://github.com/OpenMW/openmw/pull/2084/files#diff-7280b7f94028bfc5009e6c34ccf2ea68R10
As far as how to fix it, not well versed enough with glsl to say, but commenting out that line causes it to no longer complain.

mp3butcher:
if change the line if it persist try replace it with 
gl_TexCoord[i] = transpose(eyePlaneMat)*viewPos ;

JDGBOLT:
Still has issues, it looks like it specifically doesn't like referencing the gl_TexCoord with i, if I change the original line to gl_TexCoord[0] or whatever, it doesn't complain anymore.

mp3butcher:
okay done mesa is VERY strict

mp3butcher:
I developped on AMD and intel so know that for sure
I just didn't expect a mesa dude will try my stuff:)
@JDGBOLT  Did it work now?

JDGBOLT:
A preliminary attempt, of just replacing the shaders worked right, but then trying out the new one the shadows are all wonky. That change you made to the TexUnitOffset caused all of the shadows no longer to be in the right places, changing it back to 1 and recompiling made it work again.

mp3butcher:
oups

JDGBOLT:
For reference here is what it looks like now with the TexUnitOffset set back to 1: https://streamable.com/at1w8

It's certainly better at being stable based on where you are looking, though obviously some various things look off, the lighting looks way too bright, and sometimes there are parts of cells that are just shadowed or whatever. Also enabling the water shader makes the reflections in the water shader completely broken. Also noticed the player doesn't have a shadow.

mp3butcher:
great

mp3butcher:
@JDGBOLT  Which osg version please?

JDGBOLT:
The OSG I am using is a custom version, basically I took the OpenMW fork of OSG, which has some patches to it, and updated it to the latest commits of OpenSceneGraph-3.6 branch of upstream. Though I did apply your PR to make it not crash. You can find the fork on my github.

JDGBOLT:
Just so you know, I have tested with upstream openscenegraph 3.6 branch, with your pull request applied, and it basically seems to have the same behavior, so I don't think the fork is having any effect on the operation of the shadows.

mp3butcher:
just changed upstream default texoffset

terabyte25:
you should probably fix the CI for windows development and include osgShadows in dlls

mp3butcher:
...was fixing it while you write

Capostrophic:
Why don't you add the config settings into settings-default.cfg so the playability doesn't depend on the state of user settings.cfg?

mp3butcher:
@Capostrophic 
In fact I tried to comply at max with @AnyOldName3 without "stealing" its works

AnyOldName3:
Nearly everything outside of `mwshadowtechnique.hpp`, `mwshadowtechnique.cpp` and `shadow.cpp` should be independent of which shadow technique is used - most of it is either fixing bugs which only show up when shadows are in use or infrastructure changes which would be beneficial to any shadow system.

terabyte25:
Another thing, which arguably could be implemented into this specific PR is the ability to change shadow settings within the settings menu in game using MyGUI.

mp3butcher:
@terabyte25 here it is...forgot to add changes

JDGBOLT:
Hey, definitely some improvements here and there. Though some things that I do notice with them, it seems like the switch you added to turn the multi threaded viewer on and off doesn't seem to work. No matter what I set that setting to, the viewer is stuck in single threaded mode. Also it seems like the frame time isn't being updated with the shadows. If you press f3 several times for the OSG debug overlay, it's showing that the operation time should be like 25 ms for the draw/cull or whatever, and the framerate is showing 24 fps or whatever. Maybe there is just differences in how it calculates between this and the other shadow PR. Also, another thing I notice is that it seems the angle of the sun is stuck straight above, without having an angle. If you say do "set timescale to 1000" in the console, the shadow direction will never change, even when it goes to night and such.

JDGBOLT:
What I am referring to as far as the debug bars, here is what the AnyOldName3 shadows looks like when running:
screenshot209

And here are yours, seems to be a lot lower framerate than what is being shown, but possible you are doing calculations in different areas.
screenshot211

It's in multithreaded drawing mode, but for that I had to comment out the two lines that you did in engine.cpp to enable the singlethreaded mode. Also how the shadows seem wrong for the time of day, since as I said, I think it's just going from above, and never moving. Capostrophic was saying it's possible the changes you did to the directional light source initialization. Hopefully this helps.

mp3butcher:
Wow stats seems completely wrong...I don't know why
I haven't done any customizations yet to the osg method...just to show it works without any mod and hacks
Concerning mt setting i'm on it

AnyOldName3:
Buildings are definitely casting and receiving shadows in my implementation. They just show up in different places because I'm not having problems with the sun light position, and things are different levels of brightness because I'm not causing other lighting bugs.

mp3butcher:
@AnyOldName3 in all cases stats seems broken with your implementation
With the PRs I've done on osg I'll remove most of stuff i copy/paste, and you've got a clean and light pr to integrate in openmw

AnyOldName3:
My work is only for the directional sun light, not for the spot lights. There's no way to have both of these systems.

I've also never touched the stats system, but I've glanced at it while looking at nearby code. I'm pretty sure we just grab whatever OSG gives us.

JDGBOLT:
Well, just tried out the latest changes you put in, and it definitely does seem like they are better. Though one thing I have noticed is that the shadows at certain distances seem to disappear and come back into existence when getting closer or farther away. Here is a video of what I am talking about:
https://streamable.com/lbjtp

These are on the default settings. Though also another thing I have noticed is that the shader seems to be quite heavy when you up the shadow map resolution. When I can commonly use 2048 on AnyOldName3's branch, on this that commonly will cause 100%+ GPU utilization. But other wise it definitely is getting better bit by bit. It would be nice if we could get point lights working at some point, as that could make interiors that much more believable.

mp3butcher:
I haven't done anything that wasn't on osg upstream
customizing would be for an other prs

JDGBOLT:
Yeah, multithreading I think is activated. Whatever was the default.

JDGBOLT:
Ah, for that I enter "player -> setflying 500" into the console, which you can access via the ~ key. With your latest push a few minutes ago, even with the single threading on I think it still has some weirdness. Will try to get another video.

mp3butcher:
@JDGBOLT  try disable mt to check if different

JDGBOLT:
Yeah, with single threading set it does this: https://streamable.com/ew1ac
And I don't think it's something with the other shadow maps, even with just 1 enabled it still does the same thing.

mp3butcher:
@JDGBOLT 
try textureunit offset >2 or other pssm settings
 (do you have enough vram? diminish texture resolution )

JDGBOLT:
By the way, OpenMW has a discord and IRC channel if you wanted to join for quicker feedback of things. Discord is at https://discord.gg/wd3eSas and the IRC channel is #OpenMW on Freenode. And also will try out the various settings, and I have 4GB of vram, so I doubt it's running low on vram.

mp3butcher:
@JDGBOLT give me the min texunit you can set ...I'll make it default

JDGBOLT:
4 was the minimum I could set, 2 and 3 would exhibit the same problem. Though it's possible trying other scenarios might cause it to still be broken, but 4 seems to have fixed it in my limited testing.

JDGBOLT:
Do you happen to know what is causing the kind of distracting twitching of the shadows and such? Also the shadows on the edges definitely aren't what I would term "sharp," as they seem to have some sort of dithering or something like that, though maybe that's what it's supposed to do.

mp3butcher:
@JDGBOLT 
Yes Its the way trees are drawn in 2 pass Alpha Blend tricky mode

JDGBOLT:
Just going to ask, not that I don't appreciate all of the work that no doubt had to go into this, but do you think the OSG shadow implementation could be changed to be a system that could support drawing shadows from multiple light sources? Since AnyOldName3's shadows have been hyper focused on shadow from a single, sun like light source, and so indoor shadows are just kind of an afterthought. But if the OSG system could be used for dynamic shadows indoors from the light sources indoors, that would be the best of both worlds. Just a thought, and thought I would ask, since it is interesting seeing another shot at shadows being done for OpenMW. Currently indoors just kind of have a downward shadow for actors, but if we could have a system focused on the indoor cells, where a different approach would need to be used, we would get the best of both worlds. Not meaning to be rude or anything, just curious.

mp3butcher:
substituting technique won't be an issue

JDGBOLT:
Yeah, unfortunately I'm not too familiar with rendering or openmw/osg internals. Just can do testing and feedback, really. Morrowind is a bit of an odd case as in the game I don't think there are any directional lights other than the sun. All other placeable lights are spherical. It would be tough I would think to reach the level of refinement of AnyOldName3's shadows as far as quality goes. That has been in development for quite some time now, and is quite fast and accurate. But we currently have no ways of doing shadows for any of the lights that are indoors, so the quality of the shadows indoors are quite diminished compared to the outdoor ones. Just was curious whether you thought it possible to handle the indoor shadows. :)

JDGBOLT:
Another area that could be useful is currently OpenMW doesn't have a way of doing LOD(Level of Detail) for static objects. The way that shadows are created in MGE XE, which is a way to get better graphics for the original engine, is it has an LOD system for farther view distances, and it creates shadows from these objects. We currently have a system for the terrain, but nothing for simplifying and displaying the models. If you increase the view distance of OpenMW, the performance just tanks, as it tries to deal with a very large amount of geometry.

AnyOldName3:
> multiple passes drawables should be done with FBO not with two drawcalls...(Adding feature in dirty stuff is never good)

The only thing that takes multiple draw calls to draw is the terrain, and that's because each tile can have a zillion different textures on different bits of it, so we have to submit an extra draw call for each pair of texture and blendmap and then use additive blending. If we didn't, we'd have to bind more textures than is possible on some of the hardware we support.

mp3butcher:
> The only thing that takes multiple draw calls to draw is the terrain
@AnyOldName3 
Strange, so what's wrong with tree leafs? Do you have any idea?

AnyOldName3:
Nothing's *wrong* with them (unless you're referring to this PR specifically - I've not tested it). They just use alpha blending, so get depth-sorted.

AnyOldName3:
We're not going to second guess whether a nif file really needs alpha blending if it says it needs alpha blending. There *is* a mod that removes the alpha blending flag from tree nifs, and it sometimes yields a significant performance improvement, but usually it's minor - remember that alpha *testing* kills the early depth test, so isn't particularly fast, and is significantly slower on tiled rendering hardware (e.g. phones, anything from Nvidia since Maxwell), so it's capable of burning up all the time saved by not depth sorting.

The modern solution for approximating alpha testing is to use alpha blending *without* depth sorting, and set the fragment's alpha to either zero or one in the shader, but never anything in between. You still have to render these things after everything which is opaque, though, as there are still depth writes, but there's no ordering requirement other than that. We can't do this, though, as the nif format doesn't have any way of expressing that this is what we should do and it breaks fixed-function compatibility.

mp3butcher:
JDGBOLT yes indeed but i think i found the reason of flickering 
the light source position flicker during cull ... surely because of some openMW tree traversal...will see it
once I do that I'll see the sahding

JDGBOLT:
Well, tested out the new changes, and Shadows seem quite broken now. They seem to be coming perhaps from some strange direction, and don't seem to move with the changing time too much. Here is a link showing what I am seeing: https://streamable.com/9a0hf

mp3butcher:
@JDGBOLT oups forgot to comment other code....

JDGBOLT:
All right, after reverting the shading change commit so I could actually see the shadows, I can confirm that the shadows have much less flickering and twitching. They are much more stable. Though I do notice that the shadows when night falls completely become strange, with the shadows getting lit from below and such. But over all it definitely is a quality improvement, once the shader is fixed not to make the shadows impossible to see and have strange colors.

mp3butcher:
@JDGBOLT 
here we are I think

JDGBOLT:
Well, it sort of seems to work, though it definitely looks really strange with the shader changes and such. This is what I am experiencing: https://streamable.com/yxv92

Also even with the shader changes reverted it still kind of shades things strangely at night, spots of shadow where things are that travel up objects and such.

mp3butcher:
mmmh double pass stuff at night I fix it

psi29a:
Why did you remove the [WIP] prefix? Is this, according to you, ready to be merged? If not, we put the [WIP] in front to indicate that no one should review it for inclusion.

mp3butcher:
@psi29a 
we checked yesterday with @JDGBOLT 
ready to review I think

mp3butcher:
@Capostrophic 
Thanks for the feedback
I haven't figure that minimap was under shadowscene
Wil fix it and add a hack for "interior sun"

> I think you haven't explained the rendering manager initialization changes, and there's odd formatting there (please stick to 4 spaces per whitespace).
What? ahh! in Settings.... ok will see

Concerning

lysol90:
Just popping in to ask... how will we handle having _two_ more or less finished PR's with shadows?

mp3butcher:
mine's definetly cleaner and the other branch might be incrementally merged

AnyOldName3:
I haven't used trapezoidal shadow maps at all. The same problem is solved by cascading shadow maps and/or light-space perspective shadow mapping. I wouldn't have bothered with LiSPSM except there was already a nearly-working implementation in `osgShadow::ViewDependentShadowMap`, which my work is based on. I had to write the cascading shadow maps from scratch, though, as Robert had put in an implementation of what he thought parallel split shadow maps were, but because the split was done in light space instead of view space, it ended up just being equivalent to having more shadow map texels in one direction than the other, but slower as it needed more draw calls and more framebuffers.

CI/before_script.msvc.sh
```diff
@@ -761,3 +761,4 @@ if [ -z $VERBOSE ]; then
 	fi
 fi
 exit $RET
+
```
AnyOldName3: Unnecessary line break

apps/openmw/CMakeLists.txt
```diff
@@ -19,10 +19,10 @@ set(GAME_HEADER
 source_group(game FILES ${GAME} ${GAME_HEADER})
 
 add_openmw_dir (mwrender
-    actors objects renderingmanager animation rotatecontroller sky npcanimation vismask
+    actors objects renderingmanager animation rotatecontroller sky shadow npcanimation vismask
     creatureanimation effectmanager util renderinginterface pathgrid rendermode weaponanimation
     bulletdebugdraw globalmap characterpreview camera localmap water terrainstorage ripplesimulation
-    renderbin actoranimation landmanager navmesh actorspaths
+    renderbin  actoranimation landmanager navmesh actorspaths parallelsplitshadowmap
```
AnyOldName3: Unnecessary space.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -228,19 +229,49 @@ namespace MWRender
...
+        sceneRoot->getOrCreateStateSet()->setMode(GL_NORMALIZE, osg::StateAttribute::ON);
+        osg::ref_ptr<osg::Material> defaultMat (new osg::Material);
+        defaultMat->setColorMode(osg::Material::OFF);
+        defaultMat->setAmbient(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
+        defaultMat->setDiffuse(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
+        defaultMat->setSpecular(osg::Material::FRONT_AND_BACK, osg::Vec4f(0.f, 0.f, 0.f, 0.f));
+        sceneRoot->getOrCreateStateSet()->setAttribute(defaultMat);
+
+        scenegroup->setNodeMask(Mask_Scene);
+        scenegroup->setName("Scene Root");
```
AnyOldName3: You've still not explained what these changes do, how they do it, or why it's necessary.

apps/openmw/mwrender/shadow.cpp
```diff
@@ -0,0 +1,111 @@
...
+    definesWithShadows["texture_offset"] = std::to_string(mOutdoorShadowTechnique->getTextureOffset());
+    definesWithShadows["num_pssm_texture"] = std::to_string(mOutdoorShadowTechnique->getSplitCount());
+    definesWithShadows["pssm_texture_size"] = std::to_string(mOutdoorShadowTechnique->getTextureResolution());
+    return definesWithShadows;
+}
+
+Shader::ShaderManager::DefineMap getShadowsDisabledDefines()
+{
+    Shader::ShaderManager::DefineMap definesWithShadows;
+    definesWithShadows["shadows_enabled"] = "0";
```
AnyOldName3: This won't work - the preprocessor fails on undefined references, even if they're `#if`ed out.

components/shader/shadermanager.cpp
```diff
@@ -145,7 +146,7 @@ namespace Shader
             static unsigned int counter = 0;
             shader->setName(std::to_string(counter++));
 
-            shaderIt = mShaders.insert(std::make_pair(std::make_pair(shaderTemplate, defines), shader)).first;
+            shaderIt = mShaders.insert(std::make_pair(std::make_pair(shaderTemplate, localdefines), shader)).first;
         }
```
AnyOldName3: This won't work if the global defines are changed while OpenMW is running (which you should be ready for them to do as we have an in-game settings menu).

CI/before_script.msvc.sh
mp3butcher: @AnyOldName3 I obviously copy past your work :)
I'll redo this pr cleaner after bugs solved

files/settings-default.cfg
```diff
@@ -259,6 +259,9 @@ texture min filter = linear
 # Texture mipmap type.  (none, nearest, or linear).
 texture mipmap = nearest
 
+# change it to false to fix shadow flickering
+multi thread viewer = true
```
AnyOldName3: You've done something wrong if shadows flicker with multithreading.

apps/openmw/mwrender/renderingmanager.cpp
mp3butcher: just add a derivation to the tree in order to choose weither or not i cast shadow on

files/shaders/objects_fragment.glsl
```diff
@@ -146,9 +147,9 @@ void main()
     float shininess = gl_FrontMaterial.shininess;
     vec3 matSpec = gl_FrontMaterial.specular.xyz;
 #endif
-
     gl_FragData[0].xyz += getSpecular(normalize(viewNormal), normalize(passViewPos.xyz), shininess, matSpec);
 
+    gl_FragData[0].xyz = mix(ambientBias.y * gl_FragData[0].xyz, gl_FragData[0].xyz, shadowing);
```
AnyOldName3: This isn't a good way of doing things. Applying the shadow to just the light casting shadows is a better idea.

files/shaders/shadows_fragment.glsl
```diff
@@ -0,0 +1,83 @@
...
+        term+= (shadow3)*map0[3];
+#endif
+#if(@num_pssm_texture >4)
+        float shadowOrg4 = shadow2D( shadowTexture4, gl_TexCoord[@texture_offset +4].xyz+vec3(0.0,0.0,fZOffSet) ).r;
+        float shadow04 = shadow2D( shadowTexture4, gl_TexCoord[@texture_offset +4].xyz+vec3(-fTexelSize,-fTexelSize,fZOffSet) ).r;
+        float shadow14 = shadow2D( shadowTexture4, gl_TexCoord[@texture_offset +4].xyz+vec3(fTexelSize,-fTexelSize,fZOffSet) ).r;
+        float shadow24 = shadow2D( shadowTexture4, gl_TexCoord[@texture_offset +4].xyz+vec3(fTexelSize,fTexelSize,fZOffSet) ).r;
+        float shadow34 = shadow2D( shadowTexture4, gl_TexCoord[@texture_offset +4].xyz+vec3(-fTexelSize,fTexelSize,fZOffSet) ).r;
+        float shadow4 = ( 2.0*shadowOrg4 + shadow04 + shadow14 + shadow24 + shadow34)/6.0;
+        term+= (shadow4)*map0[4];
```
AnyOldName3: Copy and paste much?

apps/openmw/mwrender/renderingmanager.cpp
AnyOldName3: That only partially answers the first of those three questions.

files/settings-default.cfg
mp3butcher: This flag would allow to debug (it could be cool to diagnose end user issue)

files/shaders/objects_fragment.glsl
mp3butcher: I say it : simplistic I let you do what u want as i don't master the 2 passes terrain shading

files/shaders/shadows_fragment.glsl
mp3butcher: ...arf...straight forward

mp3butcher: ..but straightforward

files/settings-default.cfg
AnyOldName3: Or, as it says in this page I've sent you a link to many times, you could just use the OSG environment variable that does the same thing, as it's unlikely any *user* would ever need to do it, only a developer. https://wiki.openmw.org/index.php?title=Rendering_Architecture#Threading_considerations

files/shaders/objects_fragment.glsl
AnyOldName3: This has nothing to do with terrain shading at all. I can't tell what you're trying to say here.

components/shader/shadermanager.cpp
mp3butcher: What do you think of a int lock on globaldefines (preventing it's mod during run)?

AnyOldName3: A lock alone won't help with this in any way, shape or form. Take a look at the implementation in my PR.

files/settings-default.cfg
mp3butcher: i'll removed it

apps/openmw/mwrender/shadow.cpp
mp3butcher: thx i forgot them

components/shader/shadermanager.cpp
mp3butcher: I see what you did, i just made a simplier version that could be incrementally updated...
Further the feature of changing shaders global define during prog run should be for further pr: I guess it would be like adding gui shadow conf for the first pr..Not now
For that, I purpose to add an int guard to prevent globaldefine mod after shadowmanager updated them...
No?

files/shaders/objects_fragment.glsl
mp3butcher: so tell me where to put the line

apps/openmw/mwrender/renderingmanager.cpp
mp3butcher: arg why did all my comment have disappeared?
Have you had the time to read the answer i wrote here?

terabyte25:
I very much think most of the code here should be refractored in the chance there is anyway that Zini or psi29a wants to merge this

files/shaders/lighting.glsl
```diff
@@ -1,6 +1,6 @@
 #define MAX_LIGHTS 8
 
-vec4 doLighting(vec3 viewPos, vec3 viewNormal, vec4 vertexColor)
+vec4 doLighting(vec3 viewPos, vec3 viewNormal, vec4 vertexColor,float shadow)
```
AnyOldName3: This approach won't work without per-pixel lighting, which we recommend against as Morrowind's assets often look wrong with it (thanks, Todd Howard). Also, while better in theory, it's still being applied to all the lights, not just the sun light, and that's not good as Morrowind uses negative-intensity lights to simulate ambient occlusion, and this can lead to bright shadows. It's also not being applied to the specular component of any lights.

JDGBOLT:
Just so you know, the shaders are broken with the changes. The error message was:
0:28(14): error: `ambientBias' undeclared
0:28(14): error: type mismatch
0:28(14): error: operands to arithmetic operators must be numeric
I have a hunch it is this: https://github.com/OpenMW/openmw/pull/2084/commits/3576f45c940b0467f4cc002a538fc862ab8fd2fd#diff-688109f0da7d42b4c84a4de920f2ca99R28
Just based off the line number, but I could be wrong. Either way, the game now looks like this when running:

screenshot227

terabyte25:
Have you tried a git bisect?

JDGBOLT:
Shader updates to support the per pixel shader path now work as expected. :+1:

psi29a:
Stop messing with the title please. It's either [WIP] or it is not. Ask your questions in the PR please.

JDGBOLT:
I'm afraid there are still multiple issues that I see with playing around with the shadows. 
* Distant Land causes the quality of the shadows up close to diminish a fairly large amount. 
* The character model in the inventory is completely covered in shadow. 
* I'm not 100% sure if the map is getting updated correctly when traveling around. 
* I think there is something broken with the screenshot that gets generated when saving the game, as it appears far too dark and is blown out.

Those are the ones that I immediately notice when playing around with it. OpenMW in general seems not to accept first step PR's that get refined later on down the line for such large features, as both the Recast Navigation PR and the Shadow PR have gone through many layers of refinement through many months. Also there is the fact that there already is a PR that is quite far along for doing sun based shadows. This is one reason why I commented about interior shadows, and whether the OSG shadow system could be changed to a multiple light source based system. Sorry if you are getting a larger amount of pushback than you expected, but it's just kind of the environment.

mp3butcher:
@JDGBOLT 
surely the matrix I set back to what osg was
I just force pushed

mp3butcher:
more bugs?

JDGBOLT:
Hey, sorry, didn't notice the notification for this, will be testing now after compiling.

JDGBOLT:
Well, I can confirm that the Map seems to work as near as I can tell, though it might have been last time too. The screenshot on saving does work, and the character is no longer in shadow. But when using distant terrain there still is at least a bit of a loss in quality, and for whatever reason the textures on the distant terrain itself is broken, which doesn't happen in master or the other PR, which I think I mentioned to you before.

Here is what I am seeing with Distant Terrain:
https://streamable.com/6vqt8

But it definitely is improved from earlier. Is it perhaps possible to set a maximum distance that the shadows will draw at, or perhaps I think with the other PR there was a I think split point uniform logarithmic ratio or something to that effect that was able to clean that up a bit.

mp3butcher:
@JDGBOLT  how do you set max ram for distant terrain?

JDGBOLT:
I'm sorry, that I am not too sure about. I can do testing and poking around in the code, but I don't have a large amount of knowledge about how everything works. I'm not too sure if it's a ram issue for the distant terrain, or something else. I'm going to try playing around with the settings for the shadows just to see if there is something in the shadows that's clobbering the distant land terrain textures. Since it seems like it's perhaps trying to get it from perhaps clobbered bits of video memory or something. I know that AnyOldName had to fight for quite a while to make everything behave with the distant terrain. I thought I remember reading that the distant terrain was using different systems for the rendering or something, that was separate from how OSG handled things. But I could be wrong on that. Sorry I can't be of more help.

mp3butcher:
I know why it do that but it's strange design... (cv->getCurrentCamera()->getUserDataContainer())
is used as storage...

JDGBOLT:
Yeah, that distant terrain feature was the last big push that Scrawl put into OpenMW when he was doing a large amount of work on the project. Eventually the hope is that someone will step up and implement doing actual statics like rocks or buildings or whatever, with a LOD system, so we can have large view distances without a huge hit in performance.

mp3butcher:
https://developer.nvidia.com/gpugems/GPUGems3/gpugems3_ch21.html

JDGBOLT:
Distant statics along with a customizable shader pipeline are two of the big areas where the OpenMW engine loses graphically to the Vanilla engine with MGEXE. Unfortunately we don't have many developers who know the graphical side of things, so things move along slowly.

JDGBOLT:
I can confirm with the latest commit that distant terrain now doesn't experience the wrong textures that it had before. Just because I was curious I tried changing the shadow view distance from being tied to the view distance to being configurable, and it was kind of interesting. This definitely gave you better quality shadows closer to the player, as it's not trying to shadow regions outside that distance, and so can use more detail for things in view. Though I did notice that outside of this view distance everything was basically looking like a shadow covered all of it, so obviously not ideal. But as I said, I was just curious, and whether it would have any better performance when being used with say larger exterior cell load distances.

mp3butcher:
@JDGBOLT if I hard bind viewdistance to shadowdistance, there was a reason...

JDGBOLT:
Yep, I know, as I said, I was just curious what would happen if I did that, as for larger exterior cell load distances having the shadows enabled can tank the framerate pretty badly. I went from 23 to 32 fps with an exterior cell load distance of 3 and a view distance of 24000, and I did 8192 for the one that went to 32 fps.

mp3butcher:
@JDGBOLT In this case perhaps, I'd be preferable to find a compromise with settings :)
Optimization of the method may be done, (shadowprogram override, cull masks to optimise culling...),
but I'd say it would be preferable to keep that for other pr.

JDGBOLT:
Just so you know, I experienced a crash within OpenMW when sleeping from 7PM to 9 AM, after going the first hour or so, I think to 8PM, it segfaults. I think it's caused somehow by this PR as it doesn't happen without it. The backtrace, at least for OpenMW code, is:
https://gist.github.com/JDGBOLT/6022916f030765862e0bf3d472d93034

apps/openmw/mwrender/shadow.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+using namespace osgShadow;
+
+namespace MWRender
+{
+
+/*static */osg::ref_ptr<osg::Uniform> ShadowManager::_unshadowedAmbientBias=0;
+/*static */osg::Uniform* ShadowManager::getUnshadowedUniform()
+{
+    if(!_unshadowedAmbientBias.valid()) _unshadowedAmbientBias = new osg::Uniform("ambientBias",osg::Vec2(1,1));
```
AnyOldName3: This is definitely insufficient to avoid undefined behaviour with RTTs without shadows - you're still sampling from unbound textures even if you don't do anything with the result. This *will* produce OpenGL errors on conformant OpenGL implementations.

psi29a:
It still wouldn't be so bad to have another shadow technique in OpenMW...

mp3butcher:
I reopen the pr but I doubt it has a chance to be compatible with  @AnyOldName stuff.
Further, as PSSM is by nature ViewDependant I don't see the difference with what he has done.
If he follow a paper I would like to know which one..

mp3butcher:
I reopen the pr but I doubt it has a chance to be compatible with  @AnyOldName stuff.
Further, as PSSM is by nature ViewDependant I don't see the difference with what he has done.
If he follow a paper I would like to know which one..

apps/openmw/mwrender/shadow.cpp
mp3butcher: https://stackoverflow.com/questions/28411686/opengl-reading-from-unbound-texture-unit

psi29a:
This branch needs to be re-based a bit.

AnyOldName3:
> I reopen the pr but I doubt it has a chance to be compatible with @AnyOldName3 stuff.
> Further, as PSSM is by nature ViewDependant I don't see the difference with what he has done.
> If he followed a paper I would like to know which one..

My work took OSG's ViewDependentShadowMap class and fixed like a zillion bugs it had, added missing features, resolved incompatibilities OpenMW had with OSG features we weren't using before (mostly letting the cull visitor compute the near and far planes) and adding general infrastructure to make it all work. It's based on *that* class specifically because it was supposed to be OSG's most reliable and flexible shadow mapping implementation. However, it's also OSG's most complicated shadow mapping implementation by a ridiculous margin. Features it has, but that wouldn't be necessary for a balls-to-the-wall speed-focused implementation, include:

* Setting the initial light view and projection matrices using more complicated maths than appears strictly necessary.
* Computing the bounds of light space with a node visitor which traverses shadow casters.
* Computing the bounds of light space by using the near/far planes the cull visitor computes during the receiving scene's cull traversal, potentially overriding the existing compute near/far mode.
* Optional Parallel-split/cascaded shadow mapping with the split points chosen per-frame. (Weirdly, the original implementation called itself parallel-split, but wasn't what the original paper (or, indeed, any paper I've found) said to do, so I had to write that again from scratch.)
* Optionally meddling with the shadow projection matrix after the cull traversal based on the near/far planes computed during that traversal (which I disabled as implementing depth clamping worked much better and was faster).
* Optional Light-space Perspective Shadow Mapping with a bunch of extra twiddling tacked on that isn't in the original LiSPSM paper.
* Recomputing the bounds of light space after the cull traversal based on the bounds of render leaves collected into the renderbin.
* Normal-offset shadows to reduce shadow acne.
* Picking which shadow map(s) to sample from not based on depth, but instead which would look best while still containing all possible casters. (I intend to make this optional in a follow-up PR as it does more dynamic branching in the shader than I'd like and I realise now that no other game engine does it this way. It's mostly a holdover from trying to make the calls-itself-PSSM-but-isn't-really system work before I realised its flaws couldn't be worked around.)

A simpler "here's the view frustum, let's chop it up and render things that could cast shadows into each segment" without all the extra tricks to eek out small improvements in scene bound estimates would be faster on lower-end machines, so would likely be appreciated by at least some people.

mp3butcher:
ok so it's pssm with lispm shadowmap warping and it's better than simple pssm.
So no need to reopen this pr

akortunov:
Is this PR still maintained?

Capostrophic:
I guess not. Not sure if this is that good of a reference either. Closing for the time being, mp3butcher or AnyOldName3 or psi29a or anyone can reopen this later if they wish.

PR #2085, opened 2018-12-20
Native graphics herbalism support

akortunov:

You can take meshes from [this](https://www.nexusmods.com/morrowind/mods/46599) mod.

Related [forum topic](https://forum.openmw.org/viewtopic.php?f=6&t=5574).

How this feature can be implemented:
1. Mark some nodes in container's mesh as harvestable and hide them when container is activated.
2. Use NiSwitchNode in plants meshes, which have two child nodes: a basic mesh and a harvested mesh.
3. Use a separate mesh with suffix for harvested plants, as for sheathing feature.

In any case content of container is placed to the player's inventory immediately after activation.

This PR now uses a second approach, as in Skyrim.

### DIFF
--- a/apps/openmw/mwworld/actionharvest.cpp
+++ b/apps/openmw/mwworld/actionharvest.cpp
@@ -1,0 +1,93 @@
+#include "actionharvest.hpp"
+
+#include <sstream>
+
+#include <MyGUI_LanguageManager.h>
+
+#include <components/misc/stringops.hpp>
+
+#include "../mwbase/environment.hpp"
+#include "../mwbase/mechanicsmanager.hpp"
+#include "../mwbase/windowmanager.hpp"
+#include "../mwbase/world.hpp"
+
+#include "class.hpp"
+#include "containerstore.hpp"
+
+namespace MWWorld
+{
+    ActionHarvest::ActionHarvest (const MWWorld::Ptr& container)
+        : Action (true, container)
+    {
+        setSound("Item Ingredient Up");
+    }
+
+    void ActionHarvest::executeImp (const MWWorld::Ptr& actor)
+    {
+        if (!MWBase::Environment::get().getWindowManager()->isAllowed(MWGui::GW_Inventory))
+            return;
+
+        MWWorld::Ptr target = getTarget();
+        MWWorld::ContainerStore& store = target.getClass().getContainerStore (target);
+        MWWorld::ContainerStore& actorStore = actor.getClass().getContainerStore(actor);
+        std::map<std::string, int> takenMap;
+        for (MWWorld::ContainerStoreIterator it = store.begin(); it != store.end(); ++it)
+        {
+            if (!it->getClass().showsInInventory(*it))
+                continue;
+
+            int itemCount = it->getRefData().getCount();
+            // Note: it is important to check for crime before move an item from container. Otherwise owner check will not work
+            // for a last item in the container - empty harvested containers are considered as "allowed to use".
+            MWBase::Environment::get().getMechanicsManager()->itemTaken(actor, *it, target, itemCount);
+            actorStore.add(*it, itemCount, actor);
+            store.remove(*it, itemCount, getTarget());
+            takenMap[it->getClass().getName(*it)]+=itemCount;
+        }
+
+        // Spawn a messagebox (only for items added to player's inventory)
+        if (actor == MWBase::Environment::get().getWorld()->getPlayerPtr())
+        {
+            std::ostringstream stream;
+            int lineCount = 0;
+            const static int maxLines = 10;
+            for (auto & pair : takenMap)
+            {
+                std::string itemName = pair.first;
+                int itemCount = pair.second;
+                lineCount++;
+                if (lineCount == maxLines)
+                    stream << "\n...";
+                else if (lineCount > maxLines)
+                    break;
+
+                // The two GMST entries below expand to strings informing the player of what, and how many of it has been added to their inventory
+                std::string msgBox;
+                if (itemCount == 1)
+                {
+                    msgBox = MyGUI::LanguageManager::getInstance().replaceTags("\n#{sNotifyMessage60}");
+                    Misc::StringUtils::replace(msgBox, "%s", itemName.c_str(), 2);
+                }
+                else
+                {
+                    msgBox = MyGUI::LanguageManager::getInstance().replaceTags("\n#{sNotifyMessage61}");
+                    Misc::StringUtils::replace(msgBox, "%d", std::to_string(itemCount).c_str(), 2);
+                    Misc::StringUtils::replace(msgBox, "%s", itemName.c_str(), 2);
+                }
+
+                stream << msgBox;
+            }
+            std::string tooltip = stream.str();
+            // remove the first newline (easier this way)
+            if (tooltip.size() > 0 && tooltip[0] == '\n')
+                tooltip.erase(0, 1);
+
+            if (tooltip.size() > 0)
+                MWBase::Environment::get().getWindowManager()->messageBox(tooltip);
+        }
+
+        // Update animation object
+        MWBase::Environment::get().getWorld()->disable(target);
+        MWBase::Environment::get().getWorld()->enable(target);
+    }
+}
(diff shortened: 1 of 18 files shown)


akortunov:
I updated this PR to use NiSwitchNodes (as in Skyrim) and uploaded modified meshes for testing.

components/nifbullet/bulletnifloader.cpp
```diff
@@ -241,6 +241,11 @@ void BulletNifLoader::handleNode(const std::string& fileName, const Nif::Node *n
                 // Marker can still have collision if the model explicitely specifies it via a RootCollisionNode.
                 return;
             }
+            else if (sd->string == "HBN" && autogenerated)
```
Capostrophic: Looks like outdated code which should likely be removed.

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -420,6 +420,17 @@ int MWWorld::ContainerStore::remove(const std::string& itemId, int count, const
     return count - toRemove;
 }
 
+bool MWWorld::ContainerStore::isEmpty() const
```
Capostrophic: If someone is going to use this function in future (instead of doing something like comparing the begin iterator to the end iterator), there might be issues caused by considering not actually empty containers without items visible in the inventory in them non-empty. Maybe it should be renamed to reflect its purpose better.

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -420,6 +420,17 @@ int MWWorld::ContainerStore::remove(const std::string& itemId, int count, const
     return count - toRemove;
 }
 
+bool MWWorld::ContainerStore::hasVisibleItems() const
```
Capostrophic: And now it checks the truth of the fact that there *aren't* any visible items in the inventory. :P
The name is fine as it is but the returned values and the handling of the returned boolean should be inverted.

akortunov:
Should be pretty finished now. If we can use it, I'll add a geature request and changelog entry.
@psi29a, can we have a basic graphics herbalism support as an engine feature?

psi29a:
Worried about the save game format change as a result of this engine feature.

akortunov:
> Worried about the save game format change as a result of this engine feature.

For now I have no better ideas - we need to determine somehow if a container was harvested and store this state somewhere.

unelsson:
Would be great to have engine support for this feature! However does this feature require using NIF-format? I'm looking forward to creating things with latest blender, exporting to blender-supported format, and then using models straight in the game!

akortunov:
> However does this feature require using NIF-format?

I have no idea what do you mean. NiSwitchNode was a part of NIF format for years, so I suppose NIF plugins already should support it.

lysol90:
I guess he wonders if the feature works with other formats than NIF, like the osg-native format for example.

unelsson:
Yes, osg-native perhaps, but there was also some talk of Collada and glTF at http://forum.openmw.org/viewtopic.php?f=2&t=3724&p=60901&hilit=opengex#p60901 .

akortunov:
According to Greatness7, Morrowind reuses the `MNAM` record for containers to store the "isEmpty" flag for containers, fully emptied by player. This flag is used for optimization - Morrowind does not store empty containers with the same ID as separate references. Also Morrowind reset this flag if container's content was respawned (and when any other item is placed to the container, I suppose).

In theory, we can use a similar approach. In this case player will have an ability to enable/disable herbalism feature without any data loss.

akortunov:
Personally I prefer a herbalism-specific flag. It this case if we are going to add additional states, we can add them without changing a savegame format.

Should be ready for review and testing.

Capostrophic:
My most major concerns right now:
1. Savegame format change.  You've already explained in details why you desire it to be a thing, but other than this specific gimmick, there's really no reason to do it *now*. Once that reason (the said potential optimization) appears, *maybe* it could be reconsidered, but not for such a minor addition. It's bare-bones format support without any configurability unlike MWSE scripts, there isn't the capability for that. More pleasant behavior could be implemented in scripts (don't complain that'll happen in ten years from now).
That is to say there isn't any visual/auditory indication that the player wasn't successful in looting the container other than it becoming, well, visually harvested anyway.
2. Doesn't make sense to have this toggleable if it happened to be a thing. There isn't a setting for glowing windows support and it in theory can't possibly conflict with existing modifications that don't already depend on the equivalent functionality of say the MWSE-Lua script. Arguably it's still should be documented but it shouldn't be dummied out by default.

Minor concerns:
The way canBeActivated and canBeHarvested replacements were added into Container class is awkward, it's not done this way anywhere else and this particular part:
```cpp
    bool Container::canBeActivated(const MWWorld::Ptr& ptr) const
    {
        return MWClass::canBeActivated(ptr);
    }
```
screams "a crutch" to me.

Maybe something else I forget about.

akortunov:
> My most major concerns right now:

Well, technically there is a way to avoid these issues - instead of storing flag, check if a container has items when we add it to scene. In this case we do not need to change a savegame format.
Then initially empty containers will be considered harvested too, so we do not need the "visual/auditory indication" here.
Unfortunately, this way requires some additional work on plant meshes to look good.

akortunov:
OK, a change in current approach - make this feature completely stateless, so there is no need to change a savegame format or add new settings.

But this approach has a drawback - initially empty containers (e.g. when levelled list did not generate an item), so there is a need to re-design some meshes from mentioned mod.

akortunov:
I figured out that we accidentally generate content for every container in visited cell just to check if there are scripted items in it (in the AddContainerItemScriptsVisitor), but it is not right - Morrowind generates containers's content only when player opens it.

So I decided to modify the AddContainerItemScriptsVisitor to ignore containers without content.
With such approach we will generate container's content only when we really need to know its content - when player opens it, when player is in trade with container's owner or when we modify its content via scripts.

With this approach savegames size will grow more slow than now and the saving/loading will be faster.
Also this approach will allow a savescumming to get a better loot or trader's stock, as in Morrowind.

apps/openmw/mwworld/actionharvest.cpp
```diff
@@ -0,0 +1,90 @@
...
+                    stream << "\n...";
+                else if (lineCount > maxLines)
+                    break;
+
+                // The two GMST entries below expand to strings informing the player of what, and how many of it has been added to their inventory
+                std::string msgBox;
+                if (itemCount == 1)
+                {
+                    msgBox = MyGUI::LanguageManager::getInstance().replaceTags("\n#{sNotifyMessage60}");
+                    msgBox = boost::str(boost::format(msgBox) % itemName);
```
Capostrophic: Use the syntax introduced in [this commit](https://github.com/OpenMW/openmw/commit/8ecd0b82a4604d7bc5e34b663c218c9c005dd34b) instead so the Boost dependency can be avoided.

apps/openmw/mwworld/cellref.hpp
```diff
@@ -99,6 +99,7 @@ namespace MWWorld
         // For an unlocked door, it is set to -(previous locklevel)
         int getLockLevel() const;
         void setLockLevel(int lockLevel);
+
```
Capostrophic: Extra line.

akortunov:
If this approach is OK, I'll create a related feature request and will squash commits.

Capostrophic:
I have an idea regarding hasTooltip constness, will demonstrate it a bit later.

akortunov:
The only thing left - it is unclear how to handle console commands such as AddItem, RemoveItem and GetItemCount. These commands can generate container's content and can make container empty or not empty, but they do not update container's mesh state.
I suppose we can either update container's mesh state in the containerextensions.cpp or to leave the mesh update to content developers (they can update the state manually via Disable/Enable).

Capostrophic:
So:
1. MWSE herbalism can't handle scripted contents updates either (due to MWSE hook limitations), so that's likely not important at the moment.
2. Basic herbalism support is basically finished - at least objective flaws of the approach are not obvious - and possible improvements can be introduced later in 0.46.0 development cycle.
3. AnyOldName3 didn't give an opinion on this. David C. noted a fully scripted solution would be preferred but since that's not possible in pure openmw at the moment this is fine to be merged as a stopgap solution.

Hereby I invoke the awesome "merge pull request" powers.

PR #2089, opened 2018-12-21
Editor: Add basic info window

akortunov:
screenshot_20181221_184821
Implementation is based on the Help menu from qBitTorrent.

Also adds an "About Qt" window.
I have no idea if it will work on Qt4, also it would be nice if any English-speaker could check spelling.


Capostrophic:
Honestly I'd get rid of the description or replace it with something much shorter like "OpenMW-CS is a content file editor for OpenMW, a modern free and open source game engine", TESCS "About" window [doesn't have it](https://upload.wikimedia.org/wikipedia/ru/c/cd/The_Elder_Scrolls_IV_Construction_Set_screenshot.png).

I also suggest to replace (c) with ©, and show the used version of the binary, like TESCS does.

akortunov:
> and show the used version of the binary

It may be too complicated since for that I need to read OpenMW version in the C++ code.

Capostrophic:
Moreover, we'll have to edit the "current year" copyright date every year which might be cumbersome.

akortunov:
In theory, we can take data from openmw.appdata.xml, but I do not know how to use it.

akortunov:
I did what I could do here.
In theory, it is possible to generate an HTML file via CMake and use it as a Qt resource, but it is not for me.

psi29a:
This gets us further. If someone would like to improve on this, they can do it in another PR.

PR #2090, opened 2018-12-23
android_main.c: Replace nullptr back to NULL since there are no nullptr in C

xyzz:


lysol90:
Huh, so ressurections can actually happen then?

psi29a:
Welcome back!

PR #2092, opened 2018-12-23
Store attributes and skills values as floats

akortunov:
Fixes [bug #4021](https://gitlab.com/OpenMW/openmw/issues/4021).

Summary of changes:
1. Store skills and attributes values as floats instead of integers.
2. Modify attribute-related script functions to accept and return floats instead of integers.
3. Provide a fallback mechanism to read old saves.

Notes:
1. This PR needs a careful testing since it is a pretty intrusive.
2. It may not help to Halbe Null with his levelling mod since float variables can contain rounding errors.
3. It seems some formulas in Morrowind use integer values of skills and some use full values of skills.
Since to handle every case separately will take too much time, for now I use full values everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
     Bug #2311: Targeted scripts are not properly supported on non-unique RefIDs
     Bug #3676: NiParticleColorModifier isn't applied properly
     Bug #3714: Savegame fails to load due to conflict between SpellState and MagicEffects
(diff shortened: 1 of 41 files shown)


Capostrophic:

psi29a:
Conflict due to @Capostrophic refactor work.

psi29a:
Right,previous PR seemed to go down well. Time to take this pill and move on.

Thanks @akortunov

PR #2094, opened 2018-12-23
Cherry-pick cc9cii's minor editor improvements

Capostrophic:
This is an adaptation of certain minor editor improvements from cc9cii's branch (the one with support of new ESM/BSA formats) that can be used separately.

1. getModelIndex no longer throws an "object doesn't exist" exception if the ID doesn't exist in the table which could happen when you delete a record while its editing subview is open
2. Use std::to_string instead of ostringstream in refcollection. According to cc9cii it leads to a minor performance improvement.
3. Use internal message system instead of common logging system for a warning during exterior moved reference loading.

Not sure if all of these are valid.


akortunov:
Looks good to me.

PR #2095, opened 2018-12-23
Add more settings to water reflections

akortunov:
Based on old #1747 PR, but uses a different settings set:

1. Reflect terrain and sky.
2. Also reflect statics, activators and doors.
3. Also reflect items and containers.
4. Also reflect actors.

Also I tried to limit reflection distance to see if it will increase performance, but without success.

### DIFF
--- a/docs/source/reference/modding/settings/water.rst
+++ b/docs/source/reference/modding/settings/water.rst
@@ -62,17 +62,18 @@ reflection detail
 --------------
 
 :Type:		integer
-:Range:		0, 1, 2, 3
-:Default:	1
+:Range:		0, 1, 2, 3, 4
+:Default:	2
 
 Controls what kinds of things are rendered in water reflections.
 
-0: only terrain and the sky are reflected (acts like level 1 in interiors)
-1: statics, activators, and doors are also reflected
-2: items, containers, and particles are also reflected
-3: actors are also reflected
+0: only sky is reflected
+1: terrain is also reflected
+2: statics, activators, and doors are also reflected
+3: items, containers, and particles are also reflected
+4: actors are also reflected
 
-The player is always reflected in third-person mode.
+In interiors the lowest level is 2.
 This setting can be changed ingame with the "Reflection shader detail" dropdown under the Water tab of the Video panel in the Options menu.
 
 small feature culling pixel size
--- a/docs/source/reference/modding/settings/water.rst
+++ b/docs/source/reference/modding/settings/water.rst
@@ -58,17 +58,22 @@ This setting has no effect if the shader setting is false.
 
 This setting can be toggled with the 'Refraction' button in the Water tab of the Video panel of the Options menu.
 
-reflect actors
+reflection detail
 --------------
 
-:Type:		boolean
-:Range:		True/False
-:Default:	False
+:Type:		integer
+:Range:		0, 1, 2, 3
+:Default:	1
 
-This setting controls whether or not NPCs and creatures are drawn in water reflections.
-Setting this to true will enable actors in reflections and increase realism with a likely decrease in performance.
+Controls what kinds of things are rendered in water reflections.
 
-This setting can be toggled with the 'Reflect actors' button in the Water tab of the Video panel of the Options menu.
+0: only terrain and the sky are reflected (acts like level 1 in interiors)
+1: statics, activators, and doors are also reflected
+2: items, containers, and particles are also reflected
+3: actors are also reflected
+
+The player is always reflected in third-person mode.
+This setting can be changed ingame with the "Reflection shader detail" dropdown under the Water tab of the Video panel in the Options menu.
 
 small feature culling pixel size
 --------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
     Feature #4730: Native animated containers support
     Feature #4812: Support NiSwitchNode
     Feature #4836: Daytime node switch
+    Feature #4859: Make water reflections more configurable
     Feature #4887: Add openmw command option to set initial random seed
     Feature #4890: Make Distant Terrain configurable
     Task #4686: Upgrade media decoder to a more current FFmpeg API
--- a/apps/openmw/mwclass/activator.cpp
+++ b/apps/openmw/mwclass/activator.cpp
@@ -2,6 +2,7 @@
 
 #include <components/esm/loadacti.hpp>
 #include <components/misc/rng.hpp>
+#include <components/sceneutil/positionattitudetransform.hpp>
 
 #include "../mwbase/environment.hpp"
 #include "../mwbase/windowmanager.hpp"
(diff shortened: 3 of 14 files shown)


akortunov:
Request a review from @AnyOldName3.

AnyOldName3:
Regarding the cull masks, I have a few points to make:

* Is there enough of a distinction between objects and statics to justify having different node masks? My discussions with Scrawl made it seem like the only worthwhile distinction to make is actor versus object. This also means that if we come up with some kind of system in the future to determine which nodes are truly static and which can be modified by scripts (e.g. for distant statics), the obvious name is already taken.
* When I discussed it with Scrawl, he definitely thought ordering `vismask.hpp` by type rather than date added was a good idea, hence why I added my `Mask_Object` to the middle of the file in the shadows PR. As it says in the comment at the top of the file, nothing's dependent on the enum values staying the same, so clarity wins out over a small diff.
* There's still a reasonable chance that my shadows will be merged in 2018 (depending on how my time is split between family, sleep and OpenMW), so it's probably best not to have two vismask PRs in-flight and being reviewed at once, as they're obviously going to conflict.
* I've not looked in detail so may be wrong, but I suspect that with where the object and static nodemasks are applied, you can end up with objects and maybe statics as the child nodes of actors. I'm not sure if there are any especially negative consequences to this, but it probably isn't ideal.

wareya:
Statics are completely static, and can e.g. build the world entirely in interiors. Non-static objects can be animated, like silt striders. I found a very respectable performance increase in my own testing from disabling non-static objects, leaving statics rendering, in reflections. We discussed this in my earlier PR.

Capostrophic:
For reference Morrowind's water shader has terrain reflections toggleable - though given the cubemap is only updated once per every 20 seconds by default it doesn't make much of a difference.

akortunov:
> There's still a reasonable chance that my shadows will be merged in 2018 

Did I miss something?

psi29a:
So what is our position here? Sit on this until shadows are merged?

AnyOldName3:
Probably. If the definition of statics here is correctly implemented, sensible and useful (and we can all agree on a name for unanimated objects), it might be sensible to make shadows adapt to this new set of options, though.

terabyte25:
I'm guessing the [Do not merge yet] flag should be added?

akortunov:
Should be ready for review now.

akortunov:
@AnyOldName3, @wareya?

As about activators, the only possible inmprovement I see is to check if an activator really has the "x" mesh. If the activator has it, use Mask_Object for this activator, otherwise use the Mask_Static.
But since most of activators are usually just statics with script, there should not be much difference.

Also we can use the default "1" value instead of "2" by default to do not reflect items.

apps/openmw/mwclass/npc.cpp
```diff
@@ -417,6 +419,7 @@ namespace MWClass
     void Npc::insertObjectRendering (const MWWorld::Ptr& ptr, const std::string& model, MWRender::RenderingInterface& renderingInterface) const
     {
         renderingInterface.getObjects().insertNPC(ptr);
+        ptr.getRefData().getBaseNode()->setNodeMask(MWRender::Mask_Actor);
```
Capostrophic: I assume setNodeMask calls were moved from insertNPC and insertCreature to class classes for consistency with other objects, but is it really necessary? Shouldn't it be fine to assume any inserted creature and actor must have Mask_Actor node mask?

docs/source/reference/modding/settings/water.rst
```diff
@@ -58,17 +58,21 @@ This setting has no effect if the shader setting is false.
...
-Setting this to true will enable actors in reflections and increase realism with a likely decrease in performance.
+Controls which object should be rendered in the water reflections in exteriors when the water shader is enabled.
+If the setting is 0, only terrain and sky will be reflected.
+If the setting is 1, also statics, activators and doors will be reflected.
+If the setting is 2, also items, containers and particles will be reflected.
+If the setting is 3, also actors will be reflected.
+In interiors all objects are reflected, player is always reflected too.
 
-This setting can be toggled with the 'Reflect actors' button in the Water tab of the Video panel of the Options menu.
+This setting can be configured via "Reflection shader detail" dropdown menu under Water section in the in-game settings menu.
```
Capostrophic: This setting controls the level of detail of exterior shader water reflections.
Level 0: only terrain and sky are reflected;
Level 1: statics, activators and doors are reflected;
Level 2: items, containers and particles are reflected;
Level 3: actors are reflected.

Player character is always reflected, and interior water always reflects everything.
This setting can be configured via 'Reflection shader detail' dropdown menu in the Water tab of the Video panel of the Options menu.

wareya: Controls what kinds of things are rendered in water reflections.

0: only terrain and the sky are reflected (acts like level 1 in interiors)
1: statics, activators, and doors are also reflected
2: items, containers, and particles are also reflected
3: actors are also reflected

The player is always reflected in third-person mode. (?)
This setting can be changed ingame with the "Reflection shader detail" dropdown under the Water tab of the Video panel in the Options menu.

AnyOldName3:
As this splits things into more levels, is it reasonable to have a follow-up PR for the shadow system to add the extra drawable type as a caster?

akortunov:
> is it reasonable to have a follow-up PR for the shadow system to add the extra drawable type as a caster?

Or we can just add it in this PR.
You can either:
1. Add your commits here.
2. Explain what to do and I'll try to code it.

Capostrophic:
To be honest... does having a single setting with a few integer values really present any benefit in comparison to a few checkboxes which toggle the reflection of object types separately (like in Ogre OpenMW water shader options)?

wareya:
To some extent, yes. It doesn't make any sense to render activators without rendering statics, for example.

Capostrophic:
Ok.

Also, Morrowind allows to disable terrain reflections (but not the sky reflection obviously) for pixel shader water. Shouldn't an option to do the same be present?

wareya:
Sure, I guess.

akortunov:
> Also, Morrowind allows to disable terrain reflections (but not the sky reflection obviously) for pixel shader water. Shouldn't an option to do the same be present?

Is there a real reason to do it? TBH, water shader without reflections look really weird.

wareya:
Performance on machines where reflecting interior statics is expensive I guess.

akortunov:
> Performance on machines where reflecting interior statics is expensive I guess.

On such machines probably it is better to disable water shader at all.

wareya:
Not necessarily. The non-shaded water can trigger bugs in bad drivers because it's just a single huge alpha-shaded surface, like how it used to have broken fog on certain AMD card-driver combinations. It's also not necessarily a lot much more expensive to render the scene one more time if you're not including very many objects in the second render, like if you're only rendering the skybox.

akortunov:
A water shader without reflections and refraction:
screenshot_20190224_205921

Looks just as a white plane.

psi29a:
Need a rebase here.

akortunov:
> Need a rebase here.

Done.

PR #2097, opened 2018-12-24
Do not scale player's model in the 1st-person view depending on race

akortunov:
Fixes [bug #4383](https://gitlab.com/OpenMW/openmw/issues/4383).

Now Bosmers-archers should be playable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
     Bug #2987: Editor: some chance and AI data fields can overflow
     Bug #3623: Fix HiDPI on Windows
     Bug #4411: Reloading a saved game while falling prevents damage in some cases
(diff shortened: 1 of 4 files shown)


Capostrophic:
Not especially happy about camera.cpp becoming dependent on world to do something that's not directly related to the camera, but hey, if it works...

AnyOldName3:
I agree about camera depending on world being weird. Is there definitely no cleaner way of doing things?

akortunov:
> Is there definitely no cleaner way of doing things?

Well, I can update scale manually after every toggleViewMode() in six different places. I'd not consider this way as a "cleaner", though.

psi29a:
Executive override, merging.

PR #2098, opened 2018-12-24
loadingscreen: Fix UaF in loading screen.

xyzz:
When the CopyFramebufferToTextureCallback callback is called, in its operator() it resets setInitialDrawCallback by providing a NULL pointer.
However, this causes the callback to get deleted. In turn, the "this" pointer is invalidated.
When execution returns to DrawCallback::run, it accesses a _nestedCallback member of deleted "this" which is UB.

### DIFF
--- a/apps/openmw/mwgui/loadingscreen.cpp
+++ b/apps/openmw/mwgui/loadingscreen.cpp
@@ -141,10 +141,6 @@ namespace MWGui
             int w = renderInfo.getCurrentCamera()->getViewport()->width();
             int h = renderInfo.getCurrentCamera()->getViewport()->height();
             mTexture->copyTexImage2D(*renderInfo.getState(), 0, 0, w, h);
-
-            // Callback removes itself when done
-            if (renderInfo.getCurrentCamera())
-                renderInfo.getCurrentCamera()->setInitialDrawCallback(nullptr);
         }
 
     private:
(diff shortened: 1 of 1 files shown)


davidcernat:
I remember someone who couldn't start up OpenMW or TES3MP at all until I gave them a test build with the loading screen disabled. I never figured out their real problem, but installing the latest drivers for their Intel graphics card solved it. Perhaps _this_ was the real problem.

terabyte25:
@davidcernat can you still contact the guy with the original problem by chance?

AnyOldName3:
This is probably a sensible implementation. My only quibble is that it might be an idea to double check that OSG doesn't skip anything when it detects a non-null callback, if you've not done so already - there are definitely places where a callback totally replaces the default behaviour.

xyzz:
Seems only used here https://github.com/openscenegraph/OpenSceneGraph/blob/74462c46b97970f03fd26977210a41bb7cb6ab61/src/osgUtil/RenderStage.cpp#L1154 so should be safe if it's called repeatedly.

xyzz:
Btw, I just realized that it's another race condition in OSG, if the callback changes between the NULL check and invocation.

AnyOldName3:
https://wiki.openmw.org/index.php?title=Rendering_Architecture#Threading_considerations is your go-to for that kind of thing. If it's getting updated in the frame advance or update or cull traversal, the camera will probably need double buffering or something like that.

xyzz:
Yeah I'm not sure at which point in the frame render it's getting called; either way even with that theoretical race it's still better than a definite uaf AND a race.

AnyOldName3:
Looking at things, I'm pretty sure that the call happens during the frame advance, so there could definitely be a draw traversal in flight at the same time.

davidcernat:
@terabyte25, can you please not talk to me? You've previously called me a soyboy, a cuck, a gypsy and an unemployed loser on the TES3MP Discord, and I've seen you call other people far worse things. If you want to be on your absolute best behavior here, great, but I don't want to have to interact with you at all.

psi29a:
Merging.

PR #2099, opened 2018-12-24
Replace volatile bools with std::atomic<bool>

xyzz:
Volatiles shouldn't be used for thread synchronization.

### DIFF
--- a/apps/openmw/mwsound/openal_output.cpp
+++ b/apps/openmw/mwsound/openal_output.cpp
@@ -4,6 +4,7 @@
 #include <vector>
 #include <memory>
 #include <array>
+#include <atomic>
 
 #include <stdint.h>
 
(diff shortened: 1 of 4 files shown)


psi29a:
Pulling in @kcat , but otherwise OK!

kcat:
Looks like it should be okay. You could probably improve it by not using `memory_order_seq_cst` (the heavy-handed default for reading/writing atomic vars), but other than performance concerns it should be functionally fine.

psi29a:
What kind of performance concerns are we talking about here? Anything we can look at later?

kcat:
IIRC, std::memory_order_seq_cst employs a memory fence, while usually you just need acquire/release semantics to ensure correct ordering relative to other reads/writes (sometimes not even that). It shouldn't be a significant issue unless it's being done in a really hot loop, so it can be something to look at later.

psi29a:
Merged, thanks! :)

PR #2100, opened 2018-12-25
Do not restore birthsign abilities upon game load

akortunov:
Fixes [bug #4329](https://gitlab.com/OpenMW/openmw/issues/4329).

Morrowind does not have this hack.
As a proper solution, we need to implement import of active effects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,9 @@
 
     Bug #2987: Editor: some chance and AI data fields can overflow
     Bug #3623: Fix HiDPI on Windows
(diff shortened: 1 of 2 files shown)


Capostrophic:
Anyone who makes OpenMW-specific mods shouldn't rely on behavior that is subject to change.

But anyway, I just checked the mod scripts and it should still do what it's supposed to do: it will remove the ability spells altogether and set attribute base values to the last modified values they had instead upon finishing chargen. Although its script that removes the abilities after game load again will be rendered redundant...

PR #2105, opened 2018-12-29
[0.45.0 regression] Editor: Fix color picker oddities

Capostrophic:
The "color picking finished" signal was accidentally removed without adding an equivalent by Thunderforge in his summer color picker changes so the values of fields with Color display type weren't saved anymore.

Qt's getColor returns invalid color (literally invalid QColor, intended behavior) in reaction to cancel button press and that invalid color was assigned as the new color (black in practice), which could look very odd. Now the color simply doesn't change when Cancel is pressed - there's a check that the new color is valid.


psi29a:
Letting @zinnschlag make the final decision.

akortunov:
Well, basically it works.
An another strange case is when user pressed Cancel in the color picker. In this case the editor now selects the black color. I suppose in this case the editor should just to do not alter value, right?

akortunov:
Works for me.

PR #2106, opened 2018-12-29
Revert to lexical cast for string-to-float conversion

Capostrophic:
psi29a, c'mon, don't ignore [your own good advice](https://github.com/OpenMW/openmw/commit/7c758a629363d3a7b8fcd88fd2d40ab618c2591d) when reviewing.


terabyte25:
I'm just going to say this but c++11's specialized functions are much [better performance wise](https://stackoverflow.com/a/23591010) over boost's functions.

Capostrophic:
Added a comment, also added handling for another exception.

Being slightly better performance-wise means nothing if we can't support any localization of Morrowind.ini in existence.

psi29a:
> Game engines are impossible to actually cover with tests so it's a good idea to use them where they're so obvious.

I'm not in the game industry, but my gut feeling is that you're wrong or at least miss the point of unit, component and end-to-end testing, let alone play-testing.

True, you can't cover every conceivable 'issue'. The universe will always win. However, when things do happen, it is best to write a test for it so that we can prevent regressions instead of bug-hunting and wasting time.

This particular case is a prime example of a unit-test when dealing user provided data and wrangled in some way that has an impact on the the function of the software.

wareya:
Yes, that's exactly what I mean.

PR #2107, opened 2018-12-29
Move idle dialogue playback from AiWander (bug #4594)

Capostrophic:
Partially resolves [bug 4594](https://gitlab.com/OpenMW/openmw/issues/4594).

This allows actors without AiWander package to play idle dialogue as long as they are not in combat or are followers of other actors and still fit any other previously used conditions like non-0 Hello, not swimming, not being dead or knocked down, being closer than 3000 units away to the player etc.

Vanilla without MCP also limits idle dialogue playback to NPCs but I don't think that should be replicated.

Allowing actors without AiWander to greet the player is proving to need much more effort than this relatively minor (comparatively) change so I'm not keen on adding changelog entries earlier than necessary.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -8,6 +8,7 @@
 
 #include <components/sceneutil/positionattitudetransform.hpp>
 #include <components/debug/debuglog.hpp>
+#include <components/misc/rng.hpp>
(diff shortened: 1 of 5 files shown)


psi29a:
@zinnschlag need your input on this

Capostrophic:
Creatures can now play Attack voiceover like the MCP creature voiceover modding option allows. The restriction was largely artificial.

PR #2108, opened 2018-12-29
[Code Cleanup] Use case instead of elseif chain

terabyte25:


Capostrophic:
terra no

terabyte25:
what did I do wrong

terabyte25:
oh wait the variable is supposed to be arg instead of gw

psi29a:
Don't get it, so we're adding more lines by using a switch, but no default?

psi29a:
Someone _really_ wants to make a commit. ;)

apps/openmw/mwscript/guiextensions.cpp
```diff
@@ -189,15 +189,23 @@ namespace MWScript
                 }
                 else
                 {
-                    MWGui::GuiWindow gw = MWGui::GW_None;
-                    if (arg == 3)
+                    MWGui::GuiWindow gw;
+                    switch (arg) {
```
akortunov: Usually we place brackets to the separate line.

terabyte25: Right.

akortunov:
So what we decided to do? As for me, the switch here is an overkill. Since this function is not widely used, I suppose a couple of elseifs should be enough here.

psi29a:
This isn't really code clean up but adding more code for the sake of it.

I'm going to close this for now.

PR #2109, opened 2018-12-30
Simplify some world loops

Capostrophic:
Even more syntactic sugar (for monstrous cellstore and door list loops). Slight optimization of summoned creature updates. Some redundant code cleanup.

Adjustment of blood texture recovery in preparations for potential rethinking of blood texture handling (AFAIK we only handle three while Morrowind handles eight).

Fixes a potential issue in how damage is applied to actors standing in something harmful like lava or colliding with it while the player is standing with them in godmode (the loop skipped over the remaining actors).

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -387,10 +387,8 @@ namespace MWWorld
     void World::write (ESM::ESMWriter& writer, Loading::Listener& progress) const
     {
         // Active cells could have a dirty fog of war, sync it to the CellStore first
-        for (Scene::CellStoreCollection::const_iterator iter (mWorldScene->getActiveCells().begin());
-            iter!=mWorldScene->getActiveCells().end(); ++iter)
+        for (CellStore* cellstore : mWorldScene->getActiveCells())
         {
-            CellStore* cellstore = *iter;
             MWBase::Environment::get().getWindowManager()->writeFog(cellstore);
         }
 
@@ -546,22 +544,19 @@ namespace MWWorld
     {
         // first try named cells
         const ESM::Cell *cell = mStore.get<ESM::Cell>().searchExtByName (cellName);
-        if (cell != 0) {
+        if (cell)
             return cell;
-        }
 
         // didn't work -> now check for regions
-        const MWWorld::Store<ESM::Region> &regions = mStore.get<ESM::Region>();
-        MWWorld::Store<ESM::Region>::iterator it = regions.begin();
-        for (; it != regions.end(); ++it)
+        for (const ESM::Region &region : mStore.get<ESM::Region>())
         {
-            if (Misc::StringUtils::ciEqual(cellName, it->mName))
+            if (Misc::StringUtils::ciEqual(cellName, region.mName))
             {
-                return mStore.get<ESM::Cell>().searchExtByRegion(it->mId);
+                return mStore.get<ESM::Cell>().searchExtByRegion(region.mId);
             }
         }
 
-        return 0;
+        return nullptr;
     }
 
     const Fallback::Map *World::getFallback() const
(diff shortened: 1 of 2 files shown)


apps/openmw/mwworld/worldimp.cpp
```diff
@@ -3569,27 +3533,8 @@ namespace MWWorld
...
-            texture = getFallback()->getFallbackString("Blood_Texture_0");
-            break;
-        }
-
-        std::stringstream modelName;
-        modelName << "Blood_Model_";
-        int roll = Misc::Rng::rollDice(3); // [0, 2]
-        modelName << roll;
-        std::string model = "meshes\\" + getFallback()->getFallbackString(modelName.str());
+        std::string texture = getFallback()->getFallbackString("Blood_Texture_" + std::to_string(ptr.getClass().getBloodTexture(ptr)));
```
kcat: I don't know if this is significant, but this isn't the same. Previously, any type value that wasn't 0, 1, or 2 would result in `Blood_Texture_0`. Now, for example, type 3 results in `Blood_Texture_3` and type -1 results in `Blood_Texture_-1`. Are those other values possible, and what is vanilla's behavior if so?

Capostrophic: It physically can't be a value other than the ones in [0; 2] range at the moment, see getBloodTexture.
Vanilla supports blood textures in [0; 7] range but there's a need of record loading tweaking to make it so in OpenMW.

kcat: I don't see a problem, then

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -2753,44 +2732,43 @@ namespace MWWorld
...
-            if (0 != source)
+            if (source)
             {
                 // Find door leading to our current teleport door
                 // and use its destination to position inside cell.
-                const DoorList &destinationDoors = source->getReadOnlyDoors().mList;
-                for (DoorList::const_iterator jt = destinationDoors.begin(); jt != destinationDoors.end(); ++jt)
+                for (const MWWorld::LiveCellRef<ESM::Door>& destDoor : source->getReadOnlyDoors().mList)
                 {
-                    if ((*it)->getTeleport() &&
```
akortunov: Is the getTeleport() removal intended?

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -2567,19 +2556,18 @@ namespace MWWorld
...
             MWMechanics::CreatureStats& stats = actor.getClass().getCreatureStats(actor);
             if (stats.isDead())
                 continue;
 
             mPhysics->markAsNonSolid (object);
 
-            if (actor == getPlayerPtr() && MWBase::Environment::get().getWorld()->getGodModeState())
-                return;
+            if (actor == getPlayerPtr() && getGodModeState())
+                continue;;
```
akortunov: Redundant ";" sign.

akortunov:
I commented a couple of suspicious places, otherwise this PR should be fine.

apps/openmw/mwworld/worldimp.cpp
Capostrophic: Yes, there can't be non-teleport doors in the sortedDoors vector because non-teleport doors are skipped when the vector is created.

apps/openmw/mwworld/worldimp.cpp
Capostrophic: Fixed.

apps/openmw/mwworld/worldimp.cpp
akortunov: Ok.

psi29a:
Great, let's merge thsi and get it into the hands of testers. :)

Thank you @Capostrophic

PR #2112, opened 2018-12-31
[0.45.0 regression] Don't re-enable collision body for dead actors

Capostrophic:
[Bug 4761](https://gitlab.com/OpenMW/openmw/issues/4761).

I don't really like this solution, but it's something quick to remedy the issue before new year.


psi29a:
Has Zini really been awol for the past 22 days?

terabyte25:
Probably.

terabyte25:
Probably so.

akortunov:
Works for me, but maybe it is worth to check if it does not break mannequins collisions.

Capostrophic:
It does unfortunately. Pushed an alternative check.

PR #2114, opened 2019-01-02
Remove some redundant code

Capostrophic:
1. Replace mapSchoolToSkill with its optimized carbon copy spellSchoolToSkill.
2. Replace stream-based toString in character controller with std::to_string, which should have exactly the same performance
3. Purge unnecessary STL includes - for example, autocalcspell uses std::vectors instead of std::sets it used initially so including set is no longer necessary and C limits are no longer used anywhere. Some files which depend on autocalcspell header should compile a little bit faster.

### DIFF
--- a/apps/opencs/model/world/ref.cpp
+++ b/apps/opencs/model/world/ref.cpp
@@ -1,9 +1,5 @@
 #include "ref.hpp"
 
-#include <cmath>
-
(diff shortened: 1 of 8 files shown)


akortunov:
Looks good to me.

psi29a:
Great, merging.

PR #2117, opened 2019-01-06
Make Interiors of Illusion puzzle of Sotha Sil Expanded work (bug #4778)

Capostrophic:
[Bug 4778](https://gitlab.com/OpenMW/openmw/issues/4778)

1. GetSpellEffects, GetEffect and GetSpell return 0 if they're called on non-actors instead of throwing an exception due to non-actor objects lacking a magic effect list and creature stats. This allows the puzzle script to compile properly.
2. Actor AI data is `blank`ed when loading to make sure it consists of zeroes if there isn't an AIDT subrecord in the actor record instead of something weird.

In my experience the puzze is now solvable.

Cleaned up one unnecessary line from Creature record blank method.


psi29a:
Executive override, merging.

@zinnschlag: I assume that this is mwscript compatibility support.

PR #2118, opened 2019-01-07
Split physics update from world update in the profiler

akortunov:
Currently "Physics" tab measures time of the whole world update, including weather, navigator and scene update. Technically, such behaviour is not entirely corrent.
So I decided to split physics update from world update and measure its time separately.

Notes:
1. I had to update physics before weather, but it should be harmless.
2. I use short "Phys" and "Mech" labels since otherwise digits can be over timing bar, what makes digits not readable.


psi29a:
Executive override, merging.

PR #2119, opened 2019-01-07
Get rid of unnecessary string streams

akortunov:
In many cases we use string streams to just convert a digit value to string or to concat it with string.

Since now we use the C++11, we can use the std::to_string() to convert digit to string.

Note: most of string strem usages are fine (many arguments or custom formatting), so I did not touch them.

### DIFF
--- a/components/settings/settings.cpp
+++ b/components/settings/settings.cpp
@@ -8,48 +8,6 @@
 #include <boost/filesystem/fstream.hpp>
 #include <boost/algorithm/string.hpp>
 
-namespace
-{
-
-    bool parseBool(const std::string& string)
-    {
-        return (Misc::StringUtils::ciEqual(string, "true"));
-    }
-
-    float parseFloat(const std::string& string)
-    {
-        std::stringstream stream;
-        stream << string;
-        float ret = 0.f;
-        stream >> ret;
-        return ret;
-    }
-
-    int parseInt(const std::string& string)
-    {
-        std::stringstream stream;
-        stream << string;
-        int ret = 0;
-        stream >> ret;
-        return ret;
-    }
-
-    template <typename T>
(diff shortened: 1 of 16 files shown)


components/settings/settings.cpp
```diff
@@ -37,9 +37,7 @@ namespace
     template <typename T>
     std::string toString(T val)
     {
-        std::ostringstream stream;
-        stream << val;
-        return stream.str();
+        return std::to_string(val);
```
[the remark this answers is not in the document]
akortunov: Done.

components/settings/settings.cpp
```diff
@@ -393,17 +349,18 @@ std::string Manager::getString(const std::string &setting, const std::string &ca
 
 float Manager::getFloat (const std::string& setting, const std::string& category)
 {
-    return parseFloat( getString(setting, category) );
+    return std::atof(getString(setting, category).c_str());
```
Capostrophic: Using C atof instead of C++11 stof (atoi instead of stoi) will cause undefined behavior if the string is not a valid numerical value. I think it's better to use the C++ function and catch its two exceptions using 0 as a fallback to simulate the previous behavior of parse<type> functions.
Even then, stof is bad because it's locale-dependent, so you should use boost::lexical_cast<float> instead and catch its single exception.

akortunov: Use stoi/stof for now and handle possible exceptions.

psi29a:
Marked for merging. Improves (reduces) build time across all 3 platforms which reduces turn-around time on all CI/CD instances. A clear win. Will need additional testing of master to make sure there are no regressions which is acceptable in the beginning of the release cycle.

Future 'clean-ups' in the 2nd half of the release cycle will be discouraged however.

PR #2122, opened 2019-01-08
Avoid conflict between SpellState and MagicEffects

akortunov:
Fixes [bug #3714](https://gitlab.com/OpenMW/openmw/issues/3714).

The fix has two parts:
1. Apply a fix from mentioned thread to restore old savegames.
1. Use the "PEID" subrecord instead of "EFID" in the SpellState, so in newer savegames this bug should not appear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
     Bug #2987: Editor: some chance and AI data fields can overflow
     Bug #3623: Fix HiDPI on Windows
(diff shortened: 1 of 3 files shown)


AnyOldName3:
I don't know enough about how any of this works to review it...

Capostrophic:
I'm personally skeptical about this because the savegame format will likely have to change again in this area at a certain point (when the improper corprus implementation is replaced) and the corprus effects map could potentially not be a thing then. More subrecords -- more things to keep backwards compatibility for.

akortunov:
OK, there should be a discussion about how the corprus should work first...

PR #2123, opened 2019-01-08
Optimize skinning, part 2

akortunov:
For now we do not use key lookups in the mBone2VertexMap during skinning calcualtions. Instead we just iterate over the whole map.

So I decided to replace map to vector since it stores elements contiguously and accessing the next element is much cheaper.

As a result, culling phase now takes about 0.5-1ms less time per frame when there are some actors in scene with skinned bodyparts (noticable in the profiler under Cull line).

The same thing for mBoneSphereMap, but the difference is a lot smaller - about 0.1ms (noticable in the profiler under Update line).

### DIFF
--- a/components/sceneutil/riggeometry.cpp
+++ b/components/sceneutil/riggeometry.cpp
@@ -146,18 +146,14 @@ bool RigGeometry::initFromParentSkeleton(osg::NodeVisitor* nv)
             continue;
         }
 
-        mBoneSphereVector.emplace_back(bone, influencePair.second.mBoundSphere);
-
         const BoneInfluence& bi = influencePair.second;
+        mBoneSphereVector.emplace_back(bone, bi.mBoundSphere);
 
-        const std::map<unsigned short, float>& weights = influencePair.second.mWeights;
-        for (std::map<unsigned short, float>::const_iterator weightIt = weights.begin(); weightIt != weights.end(); ++weightIt)
+        for (auto& weightPair: bi.mWeights)
         {
-            std::vector<BoneWeight>& vec = vertex2BoneMap[weightIt->first];
+            std::vector<BoneWeight>& vec = vertex2BoneMap[weightPair.first];
 
-            BoneWeight b = std::make_pair(std::make_pair(bone, bi.mInvBindMatrix), weightIt->second);
-
-            vec.push_back(b);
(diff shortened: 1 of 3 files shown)


components/sceneutil/riggeometry.cpp
```diff
@@ -145,7 +146,7 @@ bool RigGeometry::initFromParentSkeleton(osg::NodeVisitor* nv)
             continue;
         }
 
-        mBoneSphereMap[bone] = it->second.mBoundSphere;
+        mBoneSphereVector.push_back(std::make_pair(bone, it->second.mBoundSphere));
```
Capostrophic: Use emplace_back instead, it allows to avoid a copy.

components/sceneutil/riggeometry.cpp
```diff
@@ -160,9 +161,16 @@ bool RigGeometry::initFromParentSkeleton(osg::NodeVisitor* nv)
...
     {
-        mBone2VertexMap[it->second].push_back(it->first);
+        bone2VertexMap[it->second].push_back(it->first);
+    }
+
+    mBone2VertexVector.clear();
+    for (auto &pair : bone2VertexMap)
+    {
+        mBone2VertexVector.push_back(pair);
     }
```
Capostrophic: I still think using assign is better than using clear() and a loop.

components/sceneutil/riggeometry.cpp
akortunov: Done.

components/sceneutil/riggeometry.cpp
akortunov: Done.

akortunov:
Use a vector for mInfluenceMap too now, so in theory cell transition should be a bit faster now.

akortunov:
Perfonmance difference with ten visible actors (with Better Clothes, Robert's Bodies and Better Morrowind Armor):

1. Master branch
screenshot_20190109_080713

2. This PR:
screenshot_20190109_080811

So the difference is about 8 FPS for me for such scene.

akortunov:
Optimize the RigGeometry::initFromParentSkeleton (~20% less time spent an the same scene), so cell transitions should be a bit faster.

psi29a:
LGTM, merging.

PR #2125, opened 2019-01-11
Share RigGeometry node data

akortunov:
For now we calculate a lot of stuff in the RigGeometry::initFromParentSkeleton(), but most of data is actually static (excepts of link to bone nodes). So I got an idea to calculate data in the NifLoader only once, and then re-use it, but we do not know actual skeleton here.

A basic approach:
1. Store string bone names instead of reference to OSG node.
2. Since data will become static, in theory we can use the same vector for original template and all instances
3. We can load actual bones once they will be available. Probably we can use the "boneName -> osg::Node" map or even just a vector of nodes.

@kcat, @AnyOldName3, what do you think?

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -1120,7 +1120,7 @@ namespace NifOsg
             const Nif::NodeList &bones = skin->bones;
             for(size_t i = 0;i < bones.length();i++)
             {
-                std::string boneName = bones[i].getPtr()->name;
+                std::string boneName = Misc::StringUtils::lowerCase(bones[i].getPtr()->name);
 
                 SceneUtil::RigGeometry::BoneInfluence influence;
                 const std::vector<Nif::NiSkinData::VertWeight> &weights = data->bones[i].weights;
--- a/components/sceneutil/riggeometry.cpp
+++ b/components/sceneutil/riggeometry.cpp
@@ -51,6 +51,8 @@ RigGeometry::RigGeometry(const RigGeometry &copy, const osg::CopyOp &copyop)
     : Drawable(copy, copyop)
     , mSkeleton(nullptr)
     , mInfluenceMap(copy.mInfluenceMap)
+    , mBone2VertexVector(copy.mBone2VertexVector)
+    , mBoneSphereVector(copy.mBoneSphereVector)
     , mLastFrameNumber(0)
     , mBoundsFirstFrame(true)
 {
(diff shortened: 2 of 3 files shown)


terabyte25:
I think 100-200mb is way too much to give up just for faster cell transitions. Perhaps have it as an option?

akortunov:
Use the same approach as for InfluenceMap. Memory usage for me as in master branch now, or sometimes even lesser.

Unfornunately, for now mBoneNameMap makes skinning a little slower. Is there a more fast way for boneName -> boneNode mapping?

akortunov:
Try to use vectors to store bones references. Should be a bit faster now.

AnyOldName3:
So do we think that the current state of this is that it increases or reduces memory usage, cell load speed and animation speed? IMO, actual animation speed should be the priority.

akortunov:
In its current current state this PR:
1. Does not change the speed of actual animation
2. Does not speed up the first mesh load - we need to calculate data in the first time anyway.
If we will preload bodyparts later, there will be a speedup too.
3. Speeds up later loads since they should use data, cached in the mesh template.
4. As for memory usage, things are bit complicated. From the one hand, we do not need to store skinning data for every mesh instance. From the other hand, we need to store vector of pointers anyway, also cached data will never be unloaded since we have no cache expiration mechanism.

Without actual tests I can not tell if this approach ever worth to use or just leave our current behaviour.

psi29a:
Do we have some performance numbers here?  What is the memory usage like?

akortunov:
> Do we have some performance numbers here? 

On my hardware RigGeometry::initFromParentSkeleton() now works 3-4 times faster than in master since we now just copy arrays instead of recalculating them every time, but overall difference during cell transition depends on count of RigGeometries in the scene.

As for memory usage, I have no idea how to measure it, but I see no much difference.

psi29a:
OK, but I was worried about this:

>  From the other hand, we need to store vector of pointers anyway, also cached data will never be unloaded since we have no cache expiration mechanism.

Is this such a problem that we'll need cache expiration? I'd imagine that someone doing a fly-over with player speed of 6000~10000 would cause OpenMW to eat up even more memory as a result or is this not a problem?

akortunov:
> Is this such a problem that we'll need cache expiration?

We will need a cache expiration anyway.
Also keep in mind that most of meshes which use RigGeometry are either bodyparts or creatures. In theory, these meshes are a quite small part of data.

As for me, we should find a way to measure size of caches in the SceneManager and ShapeManager to check how much memory they consume.
If they do not consume much memory, we have a some kind of memory leaks.

akortunov:
According to my testing:
1. Caches actually have an expiration mechanism. I was wrong here.
2. The issue is not with meshes caches themselves, but somewhere else.

> Valgrind?

I have no idea how to use it, and it is not related to this PR anyway.

psi29a:
Fair enough. 

BTW, I pinged @nikolaykasyanov about osx failures. (also unrelated)

PR #2126, opened 2019-01-13
Account for running/sneaking stances while in air (bug #4797)

Capostrophic:
[Bug 4797](https://gitlab.com/OpenMW/openmw/issues/4797)

Make GetPCRunning and GetPCSneaking return 1 if the player is in air while the respective stance is toggled on.
Make sneaking skill usage possible if sneaking stance is toggled while in air.

The code which updated sneaking indicator state was moved out into a separate function and was adjusted slightly.

Wine has many weird quirks for me when trying to run Morrowind to check what it does so any feedback would be appreciated (for example, you can try testing the mod referenced in the report). I had to make an assumption that jumping in place should also allow the functions to return 1.


akortunov:
Did you check if this PR does not break the Stormrider controls?

Capostrophic:
It does not.

PR #2127, opened 2019-01-13
Adjust static and constant GMST usage

Capostrophic:
GMSTs are currently static after their records are loaded, so we often use `static const` to make sure the value of the recovered GMST doesn't change at all and is kept in memory for future calls to the function which uses the GMST. However, this isn't the cause for some instances of static GMST declaration, so although std::map::find call time is constant, it is there and presents additional overhead every time the GMST is recovered (because it is still recovered if the variable is static but not constant, it's not marked as read-only and will simply not reset between calls).

So I changed that. It should help slightly in the often used awareness checks which are not cheap.

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -247,7 +247,8 @@ namespace MWClass
 
         MWMechanics::applyFatigueLoss(ptr, weapon, attackStrength);
 
-        float dist = gmst.find("fCombatDistance")->mValue.getFloat();
+        const float fCombatDistance = gmst.find("fCombatDistance")->mValue.getFloat();
+        float dist = fCombatDistance;
         if (!weapon.isEmpty())
             dist *= weapon.get<ESM::Weapon>()->mBase->mData.mReach;
 
--- a/apps/openmw/mwclass/npc.cpp
+++ b/apps/openmw/mwclass/npc.cpp
@@ -296,6 +296,7 @@ namespace MWClass
             gmst.iKnockDownOddsMult = store.find("iKnockDownOddsMult");
             gmst.iKnockDownOddsBase = store.find("iKnockDownOddsBase");
             gmst.fCombatArmorMinMult = store.find("fCombatArmorMinMult");
+            gmst.iVoiceHitOdds = store.find("iVoiceHitOdds");
 
             inited = true;
         }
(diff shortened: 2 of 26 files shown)


akortunov:
Looks good, but only in case if we will not allow to change GMST values on the fly via scripts.
Otherwise we propably should optimize search itself instead.

Capostrophic:
GMSTs are already recovered into static constants in the rest of the codebase, so all the GMST declarations will have to be patched anyway to allow runtime GMST record replacement whenever that happens (likely not soon).

kcat:
Not necessarily for this PR, but related. It may also be a good idea to move all GMSTs a particular function gets to the same place. When you do
```
static const float fRepairMult = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fRepairMult")->mValue.getFloat();
...stuff using fRepainMult...

static const std::string sgp = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("sgp")->mValue.getString();
...stuff using sgp...
```
Non-trivial static local variables like this need run-time checks to test if they've been initialized on a given call, and a mutex lock to prevent concurrent initialization attempts. The compiler has to add those checks/locks where it encounters the variable declaration, typically using a check-lock-recheck method, e.g.:
```
static const float fRepairMult;
if(!__static0_initialized)
{
    __static0_lock();
    if(!__static0_initialized)
        fRepairMult = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fRepairMult")->mValue.getFloat();
    __static0_initialized = true;
    __static0_unlock();
}
...stuff using fRepainMult...

static const std::string sgp;
if(!__static1_initialized)
...
```
However, if the declarations are grouped together:
```
static const float fRepairMult = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fRepairMult")->mValue.getFloat();
static const std::string sgp = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("sgp")->mValue.getString();

...stuff using fRepainMult...

...stuff using sgp...
```
the compiler can theoretically combine the initializations:
```
static const float fRepairMult;
static const std::string sgp;
if(!__static0_initialized)
{
    __static0_lock();
    if(!__static0_initialized)
    {
        fRepairMult = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fRepairMult")->mValue.getFloat();
        sgp = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("sgp")->mValue.getString();
    }
    __static0_initialized = true;
    __static0_unlock();
}
...
```
I don't know if compilers will actually do this, but the opportunity is there for the optimizer.

psi29a:
Another round of reviews please. :)

akortunov:
> Another round of reviews please. :)

In theory, it should be fine for now, but we will need to rewrite GMST usage in two cases anyway:
1. If we will allow to modify values on the fly.
2. If we will use GMSTs not only in the main thread.

Capostrophic:
Changed the approach and reverted some "not really" necessary changes. Existing GMST declarations are only touched to make their usage as constants more obvious; static GMSTs are touched if they can be grouped into one section of the function or are not declared as "static const".

Note that the reverts of the static declarations in World implementation are actually reverts of my earlier unnecessary recent changes.

psi29a:
Is this mostly cleanup or are there clear performance advantages to this?

Capostrophic:
>Is this mostly cleanup or are there clear performance advantages to this?

I explained it in the start post. Even though there are much less read-only GMSTs overall compared to the original PR now, the ones that were supposed to be read-only are still actually read-only now, awarenessCheck being the one function that seriously needs to be as fast as possible. Then there's the situation with static variable declaration overhead, which could possibly improve.

Capostrophic:
And ultimately this makes no real difference in anything and I worry too much about stuff.

PR #2128, opened 2019-01-13
Support all eight possible blood types (feature #4958)

Capostrophic:
Apparently the second byte of actor flags informs Morrowind and TESCS that the actor uses one of eight blood types (textures) that can be set up in Morrowind.ini (this is how Diverse Blood mod works). OpenMW currently only uses three. I determined the logic of this by analyzing the mod records. I asked NullCascade if it was accurate and he answered yes.

The mod works with these changes and OpenMW-CS/esmtool can display the blood type data accurately as long as a relevant Texture Name is present for the former (it uses the first blood type as a fallback for the fallback otherwise).

TESCS uses Texture Name setting for creating the blood type list in the actor editing window so if there isn't a Texture Name for a blood type TESCS doesn't list it. Since the fallback map became easier to use recently, OpenMW-CS makes use of the fallback settings (changes courtesy of akortunov).

### DIFF
--- a/apps/esmtool/labels.cpp
+++ b/apps/esmtool/labels.cpp
@@ -704,29 +704,25 @@ std::string creatureFlags(int flags)
 {
     std::string properties;
     if (flags == 0) properties += "[None] ";
-    if (flags & ESM::Creature::None) properties += "All ";
+    if (flags & ESM::Creature::Base) properties += "Base ";
     if (flags & ESM::Creature::Walks) properties += "Walks ";
     if (flags & ESM::Creature::Swims) properties += "Swims ";
     if (flags & ESM::Creature::Flies) properties += "Flies ";
     if (flags & ESM::Creature::Bipedal) properties += "Bipedal ";
     if (flags & ESM::Creature::Respawn) properties += "Respawn ";
     if (flags & ESM::Creature::Weapon) properties += "Weapon ";
-    if (flags & ESM::Creature::Skeleton) properties += "Skeleton ";
-    if (flags & ESM::Creature::Metal) properties += "Metal ";
     if (flags & ESM::Creature::Essential) properties += "Essential ";
-    int unused = (0xFFFFFFFF ^
-                  (ESM::Creature::None|
+    int unused = (0xFF ^
+                  (ESM::Creature::Base|
                    ESM::Creature::Walks|
                    ESM::Creature::Swims|
                    ESM::Creature::Flies|
                    ESM::Creature::Bipedal|
                    ESM::Creature::Respawn|
                    ESM::Creature::Weapon|
-                   ESM::Creature::Skeleton|
-                   ESM::Creature::Metal|
                    ESM::Creature::Essential));
     if (flags & unused) properties += "Invalid ";
-    properties += str(boost::format("(0x%08X)") % flags);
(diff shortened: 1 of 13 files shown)


apps/esmtool/labels.cpp
```diff
@@ -704,26 +704,44 @@ std::string creatureFlags(int flags)
...
     if (flags & ESM::Creature::Weapon) properties += "Weapon ";
-    if (flags & ESM::Creature::Skeleton) properties += "Skeleton ";
-    if (flags & ESM::Creature::Metal) properties += "Metal ";
+    int bloodmask = ESM::Creature::BloodType2|
+                    ESM::Creature::BloodType3|
+                    ESM::Creature::BloodType4|
+                    ESM::Creature::BloodType5|
+                    ESM::Creature::BloodType6|
+                    ESM::Creature::BloodType7|
+                    ESM::Creature::BloodType8;
```
Mingun: This construction repeated five times. May be it is better to create `ESM::Creature::BloodMask` enum соnstant?

components/esm/loadcrea.hpp
```diff
@@ -28,20 +28,21 @@ struct Creature
...
-        Weapon      = 0x004, // Has weapon and shield
-        None        = 0x008, // ?? This flag appears set for every creature in Morrowind.esm
-        Essential   = 0x080,
-
-        // Blood types
-        Skeleton    = 0x400,
-        Metal       = 0x800
+        Bipedal       = 0x0001,
+        Respawn       = 0x0002,
+        Weapon        = 0x0004, // Has weapon and shield
```
Mingun: I think you can also convert comments to doc comments

akortunov:
I suppose we need a way to fill list values by values from the fallback map.
Or an alternative solution: make a readonly table and fill it by values from the openmw.cfg.
Once we will have a new table, we can implement a new type of dropdown list, which will read all records of given type.
Probably we can use the same approach for weapon types dehardcoding.

akortunov:
Just for reference: [here](https://github.com/akortunov/openmw/tree/blood) I made the fallback map static (as a Settings::Manager), so it is possible to read blood types from openmw.cfg during UI initialization in the editor.
Of course, it needs a little refactoring to make reading approach more generic, also we can extend creatures verification to check if used blood type actually exists.

components/esm/loadcrea.hpp
Capostrophic: Not a priority.

components/esm/loadcrea.cpp
```diff
@@ -121,7 +124,7 @@ namespace ESM {
         esm.writeHNOCString("FNAM", mName);
         esm.writeHNOCString("SCRI", mScript);
         esm.writeHNT("NPDT", mData, 96);
-        esm.writeHNT("FLAG", mFlags);
+        esm.writeHNT("FLAG", ((mBloodType << 10) + mFlags));
```
akortunov: Did you check if it is a backward-compatible with existing format?

Capostrophic:
Any objections?

akortunov:
Not from me.

akortunov:
Seems to work well. Merging since there were no objections in last 4 months...

PR #2131, opened 2019-01-15
Allow stray special characters before Begin statement (bug #4803)

Capostrophic:
[Bug 4803](https://gitlab.com/OpenMW/openmw/issues/4803).

A warning is logged for any special character that is not a newline character. Unfortunately OpenMW-CS can't display it in verification log, as with any warnings of file parser, even though the engine itself can log it.


psi29a:
Another good candidate for a unit-test.

PR #2132, opened 2019-01-16
Fix an exception when try to load a particle node with empty sizes array

akortunov:
Fixes [bug #4804](https://gitlab.com/OpenMW/openmw/issues/4804).

Just use the same approach as for `colors` array, so OpenMW will load the mesh.

Note: it seems the `partcolor` variable (an initial particle color and alpha, I suppose) is not used at all. Is it a bug or a feature?

In theory, there should be something like 
```
created->setColorRange(osgParticle::rangev4(partcolor, partcolor));
```
to set initial colors at least. Not sure about how the `colors` array should be handled properly, though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
     Bug #4778: Interiors of Illusion puzzle in Sotha Sil Expanded mod is broken
     Bug #4800: Standing collisions are not updated immediately when an object is teleported without a cell change
     Bug #4803: Stray special characters before begin statement break script compilation
+    Bug #4804: Particle system with the "Has Sizes = false" causes an exception
     Feature #2229: Improve pathfinding AI
     Feature #3442: Default values for fallbacks from ini file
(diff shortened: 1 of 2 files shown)


psi29a:
> Note: it seems the `partcolor` variable (an initial particle color and alpha, I suppose) is not used at all. Is it a bug or a feature?

This still needs to be answered.

@AnyOldName3 @zinnschlag any idea?

AnyOldName3:
Did the `partcolor` thing ever get used? Is there some way to search through the commit history for anything that changed it? That might give us some context. Also, by 'not used at all' are we assigning a value to it (either default or loaded from the game data) or is it truly just dead code?

akortunov:
> Did the partcolor thing ever get used? Is there some way to search through the commit history for anything that changed it?

It seems this code in unchanged since initial particle system implementation, that's why I asked.
I'm fine with leaving this variable intact.

PR #2134, opened 2019-01-19
Weapon types de-hardcoding

akortunov:
Changes custom bones loading, so requires new skeleton meshes for weapon sheathing: [Sheath.zip](https://github.com/OpenMW/openmw/files/2775564/Sheath.zip)

Summary of changes:
1. Moves most of weapontype-related logic from a lot of switches and if-s to the table. 
In theory, later we can move this table to ESM or some kind of logic. This approach will allow modders to add new weapon types with new animations (one-handed spears or magic sword which requires Destruction skill instead of Long Blades, for example) or modify existing ones (move bow to the left hand or add a separate set of animations for short blades, for example).

2. Changes custom bones loading.
Instead of hardcoded xbase_anim_sh.nif OpenMW will scan nifs in the Animation folder and will load any node marked with "BONE" NiStringExtraData. Optimizer will not remove such nodes.
This approach allows to tweak new nodes differently for females, beast races, etc.
Here I just rotated sheathing bone for beast races only (left - old behaviour, right - new behaviour):
screenshot_20190119_171304

Related [forum topic](https://forum.openmw.org/viewtopic.php?f=6&t=5660&p=60730).

### DIFF
--- a/apps/openmw/main.cpp
+++ b/apps/openmw/main.cpp
@@ -6,6 +6,8 @@
 #include <components/debug/debugging.hpp>
 #include <components/misc/rng.hpp>
 
+#include "mwmechanics/weapontype.hpp"
+
 #include "engine.hpp"
 
 #if defined(_WIN32)
@@ -102,6 +104,9 @@ bool parseOptions (int argc, char** argv, OMW::Engine& engine, Files::Configurat
 
         ("script-blacklist", bpo::value<Files::EscapeStringVector>()->default_value(Files::EscapeStringVector(), "")
             ->multitoken(), "ignore the specified script (if the use of the blacklist is enabled)")
+
+        ("weapon-type", bpo::value<Files::EscapeStringVector>()->default_value(Files::EscapeStringVector(), "")
+            ->multitoken(), "override data for given weapon type")
 
         ("script-blacklist-use", bpo::value<bool>()->implicit_value(true)
             ->default_value(true), "enable script blacklisting")
@@ -222,6 +227,10 @@ bool parseOptions (int argc, char** argv, OMW::Engine& engine, Files::Configurat
     engine.setSkipMenu (variables["skip-menu"].as<bool>(), variables["new-game"].as<bool>());
     if (!variables["skip-menu"].as<bool>() && variables["new-game"].as<bool>())
         Log(Debug::Warning) << "Warning: new-game used without skip-menu -> ignoring it";
+
+    const std::vector<std::string>& weaponTypes = variables["weapon-type"].as<Files::EscapeStringVector>().toStdStringVector();
+    for (const std::string& type : weaponTypes)
+        MWMechanics::registerWeaponType(type);
 
     // scripts
     engine.setCompileAll(variables["script-all"].as<bool>());
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -641,7 +641,7 @@ osg::Vec3f AimDirToMovingTarget(const MWWorld::Ptr& actor, const MWWorld::Ptr& t
     const MWWorld::Store<ESM::GameSetting>& gmst = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>();
 
     // get projectile speed (depending on weapon type)
-    if (MWMechanics::getWeaponType(weapType)->mWeaponClass == MWMechanics::WeaponClass::Thrown)
+    if (MWMechanics::getWeaponType(weapType)->mWeaponClass == ESM::WeaponType::Thrown)
     {
         static float fThrownWeaponMinSpeed = gmst.find("fThrownWeaponMinSpeed")->mValue.getFloat();
         static float fThrownWeaponMaxSpeed = gmst.find("fThrownWeaponMaxSpeed")->mValue.getFloat();
--- a/apps/openmw/mwmechanics/aicombat.cpp
+++ b/apps/openmw/mwmechanics/aicombat.cpp
@@ -442,10 +442,9 @@ namespace MWMechanics
 
         if (targetClass.hasInventoryStore(target))
         {
-            MWMechanics::WeaponType weapType = WeapType_None;
-            MWWorld::ContainerStoreIterator weaponSlot =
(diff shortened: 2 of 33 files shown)


akortunov:
And if anyone is interested, here is the prototype of bow in the left hand: [Left.zip](https://github.com/OpenMW/openmw/files/2775575/Left.zip)
Unfortunately, since this feature is not configurable yet, you need to modify line for MarksmanBow in the weapontype.hpp and recompile OpenMW:
```
{ ESM::Weapon::MarksmanBow, "2b", "bowandarrow", "Item Weapon Bow", "Weapon Bone Left", "Bip01 MarksmanBow", ESM::Skill::Marksman, WeaponClass::Ranged, ESM::Weapon::Arrow, WeaponType::TwoHanded },
```
Short animation: "1h" -> "2b"
Attach bone: "Weapon Bone" -> "Weapon Bone Left"

Config:
```
[Game]
use additional anim sources = true
```

Basically, ZIP-archive contains new animations for bows and meshes with the "Weapon Bone Left" bone.
screenshot_20190119_175739
screenshot_20190119_175747

apps/openmw/mwrender/animation.hpp
```diff
@@ -233,6 +233,7 @@ class Animation : public osg::Referenced
     // The node expected to accumulate movement during movement animations.
     osg::ref_ptr<osg::Node> mAccumRoot;
 
+
```
Capostrophic: Extra line break

apps/openmw/mwmechanics/aicombat.cpp
```diff
@@ -442,10 +442,9 @@ namespace MWMechanics
 
         if (targetClass.hasInventoryStore(target))
         {
-            MWMechanics::WeaponType weapType = WeapType_None;
-            MWWorld::ContainerStoreIterator weaponSlot =
-                MWMechanics::getActiveWeapon(targetClass.getCreatureStats(target), targetClass.getInventoryStore(target), &weapType);
-            if (weapType != WeapType_PickProbe && weapType != WeapType_Spell && weapType != WeapType_None && weapType != WeapType_HandToHand)
+            int weapType = ESM::Weapon::None;
+            MWWorld::ContainerStoreIterator weaponSlot = MWMechanics::getActiveWeapon(target, &weapType);
+            if (weapType != ESM::Weapon::PickProbe && weapType != ESM::Weapon::Spell && weapType != ESM::Weapon::None && weapType != ESM::Weapon::HandToHand)
```
Capostrophic: I feel like there either could be a Lesser comparison with None type, or an opportunity for another flag for some of these types.

akortunov:
The main questions here:
1. Should we de-hardcode the table?
2. If yes, should we strictly change the ESM format, of we can just add settings to the openmw.cfg?
3. Can we add an ability to define custom weapon types?

akortunov:
OK, I allow to override data via openmw.cfg now. An example for left-handed bow:
```
weapon-type=9;Bow;2b;bowandarrow;Item Weapon Bow;Weapon Bone Left;Bip01 MarksmanBow;23;1;12;1
```

No editor support yet.

akortunov:
Added basic weapon types support for the editor.

akortunov:
All right, time for review and testing, especially for new weapon types.

As about code, I do not like much the weapon types enum with negative values, also a configuration in the openmw.cfg is not very user-friendly. Any ideas how to improve then?

akortunov:
Also if we are going to move these settings to ESM files later, I can try to setup a virtual ESM::Store for weapon types, but it will not work for editor, where we setup UI widgets only once (IIRC, even before we read data from ESM).

apps/openmw/mwworld/inventorystore.cpp
```diff
@@ -357,31 +357,21 @@ void MWWorld::InventoryStore::autoEquipWeapon (const MWWorld::Ptr& actor, TSlots
...
+            int ammotype = MWMechanics::getWeaponType(type)->mAmmoType;
+            if (ammotype == ESM::Weapon::Arrow)
             {
                 if (arrow == end())
                     hasAmmo = false;
                 else
                     slots_[Slot_Ammunition] = arrow;
             }
-            if (isCrossbow == true)
+            else if (type == ESM::Weapon::Bolt)
```
Capostrophic: I think this should check ammotype.

apps/openmw/mwrender/animation.cpp
```diff
@@ -238,6 +238,25 @@ namespace
...
+    public:
+        GetExtendedBonesVisitor()
+            : osg::NodeVisitor(TRAVERSE_ALL_CHILDREN)
+        {
+        }
+
+        void apply(osg::Node& node)
+        {
+            if (SceneUtil::hasUserDescription(&node, "CustomBone"))
+                mFoundBones.push_back(std::make_pair(&node, node.getParent(0)));
```
Capostrophic: emplace_back hallowed be thy name

apps/openmw/mwmechanics/character.cpp
```diff
@@ -678,76 +651,12 @@ void CharacterController::refreshCurrentAnims(CharacterState idle, CharacterStat
 }
 
 
-void getWeaponGroup(WeaponType weaptype, std::string &group)
+void getWeaponGroup(const int weaptype, std::string &group)
```
Capostrophic: This is not a class member. Do we care if this function still exists if it's possible to just recover the group directly in the same amount of code?

apps/openmw/main.cpp
```diff
@@ -223,6 +228,10 @@ bool parseOptions (int argc, char** argv, OMW::Engine& engine, Files::Configurat
     if (!variables["skip-menu"].as<bool>() && variables["new-game"].as<bool>())
         Log(Debug::Warning) << "Warning: new-game used without skip-menu -> ignoring it";
 
+    const std::vector<std::string>& weaponTypes = variables["weapon-type"].as<Files::EscapeStringVector>().toStdStringVector();
```
Capostrophic: Nobody is really sure if overriding weapon types via configs is the right way to dehardcode them. Before any abrupt decisions are acknowledged in the codebase and it doesn't take too much effort *not* to have this, I'd reconsider having a stopgap solution this intrusive.

apps/openmw/mwworld/inventorystore.cpp
akortunov: Nice catch!

akortunov:
Replaced by #2413.

PR #2137, opened 2019-01-20
Avoid useless read of navmeshes from navigator

elsid:
Prevents [deadlock](https://gitlab.com/OpenMW/openmw/issues/4777) if navmesh rendering is disabled (but not fixes). Actulally this is a little optimization because of [this](https://github.com/OpenMW/openmw/compare/master...elsid:navmesh_render?expand=1#diff-f1edf5c7ed145cbb380aad92dedbf3bbR1404) lock.


psi29a:
Thanks!

PR #2140, opened 2019-01-22
[Discussion] Make creature/activator screams Voice-type sounds

Capostrophic:
Instead of being affected by Effect volume slider, Moan, Roar and Scream sounds that were "generated" by soundgen keys will be affected by Voice volume slider, which makes more sense to me.
I removed redundant stopSound calls, the sounds are already stopped automatically in playSound3D.


akortunov:
Did you check which slider affects them in Morrowind?

Capostrophic:
It's Effects slider for whatever reason.

kcat:
I'm not completely sure about this. It seems to me that Voice should be for actual spoken dialog, while Moan/Roar/Scream sounds would be more general effects. The latter needn't be voiced sounds at all, but could be, for example, mechanical noise from automatons.

Capostrophic:
>could be, for example, mechanical noise from automatons.

Yeah, that makes sense, they are in vanilla game. Well, this was but a random idea not well thought out.

PR #2141, opened 2019-01-23
Support NiSwitchNode

akortunov:
Implements [feature #4812](https://gitlab.com/OpenMW/openmw/issues/4812).

In upstream master we handle the NiSwitchNode as a common NiNode (render all children), while vanilla Morrowind renders just its first child (and MGE improves its loading, so it is possible to switch between its children).

This PR creates an osg::Switch for every NiSwitchNode, so it is possible so set current active child:
```
switchNode->setSingleChildOn(index);
```

By default only first children is rendered.

Note: autogenerated collisions still will be based on all node children in both Morrowind and OpenMW, so it is recommended to use either NiRootCollisionNode or NCO for such meshes.

Switch nodes will allow us to implement some extended features (such as native herbalism or glowing windows) and it does not matter if they will be hardcoded or handled via Lua scripting with scene graph access (probably based on existing Lua bindings for OpenSceneGraph).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
     Feature #3610: Option to invert X axis
     Feature #4673: Weapon sheathing
     Feature #4730: Native animated containers support
+    Feature #4812: Support NiSwitchNode
     Task #4686: Upgrade media decoder to a more current FFmpeg API
 
 0.45.0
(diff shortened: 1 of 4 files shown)


lysol90:
Nice!

psi29a:
LGTM, sweet!

PR #2142, opened 2019-01-24
Add creature-with-the-same-model soundgen fallback (bug #4813)

Capostrophic:
[Bug 4813](https://gitlab.com/OpenMW/openmw/issues/4813).

Creatures will fallback to the sounds of the first creature that has the same model in the record store if the creature couldn't get any specific sounds before falling back to actual creature-independent fallback sounds.


Capostrophic:
One of Travis builds failed for a reason unrelated to the changes.

psi29a:
Thank you, I'll take it.

PR #2143, opened 2019-01-25
Optimize HUD update

akortunov:
We update HUD widgets every frame, and in upstream master we spend some time to update windows labels, even if they did not change and even if these windows are not visible (e.g. level widget, inventory window).

With this PR we update titles only when they were actually changed.
On my hardware game mechanics update now takes up to 0.1ms less time per frame.


psi29a:
Looks good!

PR #2144, opened 2019-01-25
Don't apply effects when spell absorption is successful (bug #4820)

Capostrophic:
[Bug 4820](https://gitlab.com/OpenMW/openmw/issues/4820).
This reverts an erroneous change by noob me from the time of 0.44.0 development so absorbed effects will be again skipped instead of being applied for whatever reason sometimes. Unlike 0.43.0, however, neither harmful nor beneficial effects will be applied - which actually follows Morrowind behavior.

The whole inflict implementation is messy but it's not the time to refurnish it.


psi29a:
It's still early in 0.46 development, you can always come back to it to refactor it in another MR/PR.

Thanks

PR #2145, opened 2019-01-25
Update jail state once instead of for every single hour

akortunov:
Fixes [bug #4823](https://gitlab.com/OpenMW/openmw/issues/4823).

Instead of updating world for every single hour in prison in the loop, update world just once with parameter.

Note: a prison progress bar is an artifitial thing without any practical value. In theory, we can completely remove it.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -130,23 +130,45 @@ void getRestorationPerHourOfSleep (const MWWorld::Ptr& ptr, float& health, float
     MWMechanics::CreatureStats& stats = ptr.getClass().getCreatureStats (ptr);
     const MWWorld::Store<ESM::GameSetting>& settings = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>();
 
-    bool stunted = stats.getMagicEffects ().get(ESM::MagicEffect::StuntedMagicka).getMagnitude() > 0;
     int endurance = stats.getAttribute (ESM::Attribute::Endurance).getModified ();
-
     health = 0.1f * endurance;
 
-    magicka = 0;
-    if (!stunted)
-    {
-        float fRestMagicMult = settings.find("fRestMagicMult")->mValue.getFloat ();
-        magicka = fRestMagicMult * stats.getAttribute(ESM::Attribute::Intelligence).getModified();
-    }
+    float fRestMagicMult = settings.find("fRestMagicMult")->mValue.getFloat ();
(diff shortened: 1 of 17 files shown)


Capostrophic:
Nice, I always thought there was something odd with "rest" implementation and calls in sequence, this is a sweet optimization.

akortunov:
So what to do with progress bar?

akortunov:
All right, I suggest to do not touch the progress bar itself for now.

akortunov:
Any objections?

Capostrophic:
None from me.

akortunov:
Also keep in mind that jail progress in 0.45 takes more time than in 0.44 because we update stats for actors in inactive cells, so this bug partly can be considered as a regression. Maybe it is worth to cherry-pick this fix to the 0.45 branch as well.
What do you think?

Capostrophic:
There is a concern that theoretical temporary Stunted Magicka effects will lead to the player not getting any magicka restoration after they expire while resting.

psi29a:
Is there a ticket for this? We can consider it a known-issue (noted in changelog).

akortunov:
> There is a concern that theoretical temporary Stunted Magicka effects will lead to the player not getting any magicka restoration after they expire while resting.

I added support for temporary Stunted Magicka effects here. Do you happy now?

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -555,12 +577,31 @@ namespace MWMechanics
...
 
             DynamicStat<float> stat = stats.getHealth();
-            stat.setCurrent(stat.getCurrent() + health);
+            stat.setCurrent(stat.getCurrent() + health * hours);
             stats.setHealth(stat);
 
-            stat = stats.getMagicka();
-            stat.setCurrent(stat.getCurrent() + magicka);
-            stats.setMagicka(stat);
+            // We should take in account Stunted Magicka effects.
```
Capostrophic:
```suggestion
            // Stunted Magicka effect should be taken into account.
```

It should be good to put all the stunted magicka stuff behind a check that the actor actually has an instance effect on them. :p

akortunov:
Any suggestions here?

Capostrophic:
LGTM

akortunov:
> Hm, still... do all magic effects act as they should for sure?

?

akortunov:
Any objections?

psi29a:
@Capostrophic what did you mean?

Capostrophic:
It should be fine.

PR #2146, opened 2019-01-25
Make quest status update handling closer to vanilla (bug #4815)

Capostrophic:
[Bug 4815](https://gitlab.com/OpenMW/openmw/issues/4815).

SetJournalIndex in Morrowind simply updates the quest index without doing anything else, so setIndex shouldn't have updated quest status. Journal instruction relied on setIndex to do this, but it can only increase the index when adding an entry, while Morrowind updates quest status regardless of whether the index changed or not. The solution is to move quest status update from setIndex to Quest::addEntry. This is safe to do because the used functions are not used anywhere else other than the script instruction and Journal class implementation.

I'm concerned about Journal::addEntry, though, since its "bailing out" behavior has now changed and I'm not sure what should it actually be.


psi29a:
Executive decision.

PR #2147, opened 2019-01-26
Apply UVControllers only for given UV Set

akortunov:
Should fix [bug #4827](https://gitlab.com/OpenMW/openmw/issues/4827).

Just applies UV matrix only to textures which use a given UV Set (usually 0).
You can specify used UV Set index in the NiUVController ("unknown short" in the NifSkope).

Meshes from Glow in the Dahrk should work fine now, but UV Set selection needs testing.


akortunov:
I got a confirmation from Colt17 that basically this PR works as intended.

psi29a:
Thanks, merging.

PR #2149, opened 2019-01-29
Update fonts setup guide

akortunov:
1. Fix a copy-paste error - font size is configured in the settings.cfg instead of openmw.cfg
2. Add a step to remove the unnecessary MonoFont entry from the user-defined openmw_fonts.xml.
3. Use digits in lists instead of lowercase Roman numerals.


psi29a:
Has this been thrown through sphinx to see if it renders correctly?

akortunov:
I checked rendering via this online editor: http://rst.ninjs.org/

PR #2150, opened 2019-01-29
Update magic effects VFX immediately after effects update

akortunov:
Fixes [bug #4828](https://gitlab.com/OpenMW/openmw/issues/4828).

Currently we update the MagicEffects in the updateActor() and update visual effects itself in the updateContinuousVfx() based on the MagicEffect state, so if there is an effect added between these calls, this effect sometimes is considered as expired and OpenMW removes VFX.

An example - when combat AI uses potions.

A simpliest way to fix this issue is to update VFX immediately after magic effects update.
In this case newly added effects just will be handled in the next frame.

As an alternative solution, we can update MagicEffects in the end of actors update, but in this case combat AI will use old values, so I prefer approach from this PR since it affects only spell VFX.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
     Bug #4804: Particle system with the "Has Sizes = false" causes an exception
     Bug #4820: Spell absorption is broken
     Bug #4827: NiUVController is handled incorrectly
+    Bug #4828: Potion looping effects VFX are not shown for NPCs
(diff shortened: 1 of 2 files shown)


psi29a:
Does this have an effect/impact on rendering performance?

akortunov:
> Does this have an effect/impact on rendering performance?

It should not affect perfromance since it changes only calls order.

psi29a:
Thanks, merging.

PR #2151, opened 2019-01-29
Account for inverted tangents on normal-mapped objects (bug #3733)

Capostrophic:
[Bug 3733](https://gitlab.com/OpenMW/openmw/issues/3733)

Normal maps didn't look right on objects where UV is mirrored because the shader didn't account for the case when tangents were actually inverted in the model. So I did a thing similar to [what is done in Unity](https://forum.unity.com/threads/symmetric-models-with-mirrored-normal-maps-shader-fix.65859/#post-420888). Now such objects should look fine.

Terrain cannot have inverted tangents.
Before: before

After: after
Perhaps more apparent:
Before: before

After: after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
     Bug #2987: Editor: some chance and AI data fields can overflow
     Bug #3623: Fix HiDPI on Windows
+    Bug #3733: Normal maps are inverted on mirrored UVs
     Bug #4329: Removed birthsign abilities are restored after reloading the save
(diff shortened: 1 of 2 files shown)


psi29a:
Looks good to me, thanks.

I read the article, apparently the real fix is with the mesh which needs to be properly exported. This just 'catches' things that happen to be done badly correctly?

Capostrophic:
I got the impression that this actually makes things that were done *properly* look correctly.

psi29a:
I'll take it anyway. ;)

PR #2152, opened 2019-01-30
[0.45 regression] Editor: hide hairs when helmet is used

akortunov:
Morrowind has a quirk - when you equip a helmet, it hides hairs even in case if it replaces only Head part.

OpenMW itself takes it in account, but when we added NPC's rendering to the editor, we forgot about it.


akortunov:
Technically, it is a bug in the OpenMW-CS NPC's rendering feature in 0.45. If we are going to create an another RC because of other found regressions, it would be nice to backport this patch as well. Otherwise I'm fine if this bug will be fixed in the 0.46.

PR #2153, opened 2019-01-30
Native glowing windows support

akortunov:
[Feature #4836](https://gitlab.com/OpenMW/openmw/issues/4836).

Adds an internal ESP-less glowing system for both game and editor. It does not spawn any objects and does not alter any records, just affects rendering, so it is a purely graphics feature.

An approach for game itself is a quite simple:
1. Calculate current daytime state once per frame during weather update using settings imported from Morrowind.ini.
2. Add a callback for every osg::Switch with given name. This callback is supposed to read data from weather manager and select active child node.

For edtor we just update windows state via visitor when we change lighting mode - for exterior cells windows will glow in the "night" mode.

This feature does not affect performance with default assets since they have no NiSwitchNodes.

For testing you can use meshes from Glow in the Dahrk. On my hardware, glowing windows decrease FPS a bit because of NiUVController's for flickering, but not because if this feature itself.

Since this feature is often used, can we add a fast and reliable implementaion to the engine?

Also I am not sure which callbacks are more suitable - Update or Cull ones.

### DIFF
--- a/apps/opencs/view/render/lighting.cpp
+++ b/apps/opencs/view/render/lighting.cpp
@@ -1,5 +1,38 @@
 #include "lighting.hpp"
 
 #include <osg/LightSource>
+#include <osg/NodeVisitor>
+#include <osg/Switch>
+
+#include <components/misc/constants.hpp>
+
+class DayNightSwitchVisitor : public osg::NodeVisitor
+{
+public:
+    DayNightSwitchVisitor(int index)
+        : osg::NodeVisitor(TRAVERSE_ALL_CHILDREN)
+        , mIndex(index)
+    { }
+
+    virtual void apply(osg::Switch &switchNode)
+    {
+        if (switchNode.getName() == Constants::NightDayLabel)
+            switchNode.setSingleChildOn(mIndex);
+
+        traverse(switchNode);
+    }
+
+private:
+    int mIndex;
+};
 
 CSVRender::Lighting::~Lighting() {}
+
+void CSVRender::Lighting::updateDayNightMode(int index)
+{
+    if (mRootNode == nullptr)
+        return;
+
+    DayNightSwitchVisitor visitor(index);
+    mRootNode->accept(visitor);
+}
--- a/apps/opencs/view/render/lightingbright.cpp
+++ b/apps/opencs/view/render/lightingbright.cpp
@@ -4,7 +4,7 @@
 
 CSVRender::LightingBright::LightingBright() {}
 
-void CSVRender::LightingBright::activate (osg::Group* rootNode)
+void CSVRender::LightingBright::activate (osg::Group* rootNode, bool /*isExterior*/)
 {
     mRootNode = rootNode;
 
@@ -20,6 +20,8 @@ void CSVRender::LightingBright::activate (osg::Group* rootNode)
     mLightSource->setLight(light);
 
     mRootNode->addChild(mLightSource);
+
+    updateDayNightMode(0);
 }
 
 void CSVRender::LightingBright::deactivate()
--- a/apps/opencs/view/render/lightingnight.cpp
+++ b/apps/opencs/view/render/lightingnight.cpp
@@ -4,7 +4,7 @@
 
 CSVRender::LightingNight::LightingNight() {}
 
-void CSVRender::LightingNight::activate (osg::Group* rootNode)
+void CSVRender::LightingNight::activate (osg::Group* rootNode, bool isExterior)
 {
     mRootNode = rootNode;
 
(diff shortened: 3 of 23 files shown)


psi29a:
> This feature does not affect performance with default assets since they have no NiSwitchNodes.

Great!

> For testing you can use meshes from Glow in the Dahrk. On my hardware, glowing windows decrease FPS a bit because of NiUVController's for flickering, but not because if this feature itself.

Is this solvable? What's do you mean by 'a bit' ;)

> Since this feature is often used, can we add a fast and reliable implementaion to the engine?

If this would solve the above issue, then I believe that it falls under this PR.

> Also I am not sure which callbacks are more suitable - Update or Cull ones.

@AnyOldName3 any advice here?

akortunov:
> Is this solvable? What's do you mean by 'a bit' ;)

1-2 FPS. It is because of flickering animation, so an only way to avoid it is to remove NiUVControllers from meshes at all. But a classic Windows Glow with NiFlipController's has a way worse performance (5-8 FPS loss for me in large cities).

apps/opencs/view/render/lighting.cpp
```diff
@@ -1,5 +1,38 @@
 #include "lighting.hpp"
 
 #include <osg/LightSource>
+#include <osg/NodeVisitor>
+#include <osg/Switch>
+
+#include <components/misc/constants.hpp>
+
+class SwitchWindowsGlowVisitor : public osg::NodeVisitor
```
psi29a: Can this be more generic? I assume Glowing is something that happens with not just Windows but perhaps with other meshes/statics? For example a Cave entrance, or an object underwater?

akortunov: You are right, but I am not very good with inventing names on English.

akortunov:
Technically, this feature is about day-night mode swtch. Meshes may not have a glow at all.
I suppose a more generic name should be used, but I have no idea, which ones.

Which switch indeces are used:
0 - a default mesh, used during day in exteriors or at night (or when the weather is not bright) in interiors.
1 - a night mode for exteriors.
2 - a daytime mode for interiors (may contain sunrays, for example, which should not be present in exteriors).

psi29a:
Not just AV but also RTD, github is dropping it's 'services' features. We'll have to step over to using webhooks, if that is possible.

AV still works fine on gitlab (because they do it correctly).

So... perhaps... maybe... you guys should start using gitlab and MRs instead of github please? :)

terabyte25:
Mmm, looking back at GitLab, MRs are much better than PRs, at least for me now.

akortunov:
I renamed variables to the NightDayMode ones since this feature is not limited to glowing windows.
In theory, you can implement any object which should look differently depending on time of day.

psi29a:
I see isExterior being used... does this mean that day/night cycle for internal areas too? For example... we're inside, would the windows be light up like it was sunny outside or dark if at night?

wareya:
>does this mean that day/night cycle for internal areas too?

>2 - a daytime mode for interiors (may contain sunrays, for example, which should not be present in exteriors).

psi29a:
Last call for review?  @Capostrophic @AnyOldName3

apps/opencs/view/render/lightingnight.cpp
```diff
@@ -20,6 +20,9 @@ void CSVRender::LightingNight::activate (osg::Group* rootNode)
     mLightSource->setLight(light);
 
     mRootNode->addChild(mLightSource);
+
+    // In the light mode windows in exterior cells should glow.
```
Capostrophic: I don't understand this comment. Well, I understand the meaning behind it, but "the light mode" doesn't make much sense without the context.

apps/opencs/view/render/lightingbright.cpp
```diff
@@ -4,7 +4,7 @@
 
 CSVRender::LightingBright::LightingBright() {}
 
-void CSVRender::LightingBright::activate (osg::Group* rootNode)
+void CSVRender::LightingBright::activate (osg::Group* rootNode, bool isExterior)
```
Capostrophic: Perhaps "isExterior" should be commented to clarify that it's not going to be used. (bool /*isExterior*/).
The whole implementation of updateDayNightMode is pretty clunky with all these arguments being forwarded to each lighting mode, but I don't know what would be better.

AnyOldName3:
> > Also I am not sure which callbacks are more suitable - Update or Cull ones.
> 
> @AnyOldName3 any advice here?

I think you were right to go with an update callback for this.

akortunov:
Any other objections?

Capostrophic:
I'll try actually testing this in-game first.

Capostrophic:
Nothing obvious comes up at the first glance.

psi29a:
Merging, thanks.

PR #2155, opened 2019-02-04
Fix macOS CI

nikolaykasyanov:
It looks like Homebrew has deprecated some syntax making CMake formula invalid, now a fork of the formula is used.


psi29a:
Thanks! :)

PR #2158, opened 2019-02-05
[Regression] Remove looping particles from dead actors

akortunov:
Related [forum topic](https://forum.openmw.org/viewtopic.php?f=8&t=5713).

Fixes a regression, added by #2150.

We need to update looping particles for dead actors as well.


psi29a:
Thanks! :)

PR #2162, opened 2019-02-09
[WIP/Testing needed] Get rid of out-of-range actor collision shenanigans

Capostrophic:
Something more in-depth than the previous regression fix attempt. This, however, is not intended for 0.45.0.
While this is very "raw" and needs some cleanup, I decided to submit it so that it haunts me for the following week and I don't forget to finish it.
This is something (arguably) better than the current awkward workaround for the original issue actor processing rework tried to avoid.

Original behavior: when an actor first gets in range and gets out of range after that, its collision modes are toggled off to prevent the player being able to hit them with magic and ranged combat projectiles, and there are some checks in place to prevent actors out of range getting hit by exploding spells. When the actor gets into the processing range again, the collisions are turned on again, *even if it's a dead actor which shouldn't have collisions*.
Current behavior: same thing but the collisions are not reenabled for actors with death animation finished.
PR behavior: collisions are not touched at all, instead projectiles go through actors when their distance to the player is beyond actor processing range and exploding spells similarly don't affect out-of-range actors purely based on their distance to the player.
As a reminder, out-of-range actors are invisible so it doesn't look that weird anyway. What I currently have now worked in my testing, but there may have been some cases messing with collision modes worked around some more issues I missed. Even then, I think it's more universal, will allow to avoid issues similar to the one that was fixed recently and that the current behavior (where collision modes are turned off and on) is something openmw shouldn't really do.

This doesn't need a changelog entry because there shouldn't be any visible difference to the end user, it's just a change of approach.

I have written too much.
tl;dr I don't like how we mess with actor collisions, this is a different, less laconic but more universal approach to avoiding the original issue but I may have missed something, needs testing.

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -312,7 +312,6 @@ namespace MWBase
 
             virtual bool castRay (float x1, float y1, float z1, float x2, float y2, float z2) = 0;
 
-            virtual void setActorCollisionMode(const MWWorld::Ptr& ptr, bool internal, bool external) = 0;
             virtual bool isActorCollisionEnabled(const MWWorld::Ptr& ptr) = 0;
 
             virtual bool toggleCollisionMode() = 0;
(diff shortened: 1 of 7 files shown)


Capostrophic:
This needs some rethinking.

PR #2163, opened 2019-02-09
AI data handling minor revisions

Capostrophic:
This changes several very close things in handling AI data and ratings.
Hello is now a 16-bit integer and we're down to 3 unknown values. If you check TESCS, these are the maximum values AI ratings can have (although in Fight/Flee/Alarm cases they're deliberately capped to 100, being probabilities) after saving. This also what is present in MWSE original structure info.
tescs
mHasAI boolean was never a necessary boolean, being just an indicator that an AIDT subrecord is not present. Only one record in Morrowind.esm lacks one (the player) and only one record in Tribunal.esm also lacks one (none in Bloodmoon). It's not really used for anything at this point, other than esmtool, because the AI data is guaranteed to be present anyways.
For whatever reason OpenMW-CS and other creature/NPC record saving functionality dropped AIDT subrecord if the ratings and services are all 0 This is not what TESCS does, and TESCS appears to always save an AIDT subrecord. This leads us to the following issue:
NPCs and creatures actually have non-zero defaults if AIDT subrecord is absent in both Morrowind and TESCS, and they are different for NPCs and creatures. For NPCs, it's Fight, Flee and Hello being 30, for creatures, it's Fight being 90 and Flee being 20. When OpenMW-CS drops the AIDT subrecord, the meaning of the AI data actually changes from 0, 0, 0, 0 to different values in TESCS and Morrowind.
Added a cap to Alarm, Flee and Fight saved values. You cannot input stuff out of 0-100 bounds anymore into the relevant fields.

Of course I may have missed something, so any insight would be appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
     Bug #4828: Potion looping effects VFX are not shown for NPCs
     Bug #4837: CTD when a mesh with NiLODNode root node with particles is loaded
     Bug #4860: Actors outside of processing range visible for one frame after spawning
+    Bug #4876: AI ratings handling inconsistencies
     Feature #2229: Improve pathfinding AI
     Feature #3442: Default values for fallbacks from ini file
     Feature #3610: Option to invert X axis
--- a/apps/esmtool/record.cpp
+++ b/apps/esmtool/record.cpp
@@ -660,7 +660,7 @@ void Record<ESM::Creature>::print()
 
     printTransport(mData.getTransport());
 
-    std::cout << "  Artifical Intelligence: " << mData.mHasAI << std::endl;
+    std::cout << "  Artificial Intelligence: " << std::endl;
     std::cout << "    AI Hello:" << (int)mData.mAiData.mHello << std::endl;
     std::cout << "    AI Fight:" << (int)mData.mAiData.mFight << std::endl;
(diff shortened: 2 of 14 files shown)


akortunov:
> So, here is a concern: should blank() methods of NPC and creature records revert the ratings to the default values or zeroes?

I suppose we should use the first option.

> Should TESCS behavior of capping the saved probabilities to 100 be mimicked?

In theory, we can cap them in the game itself, but I'd prefer to mimick Morrowind approach.

Capostrophic:
Done and done. The capping actually works better than I expected. When you try to input a value out of 0-100 bounds into a modified field, it gets replaced by 100 immediately.

psi29a:
We need more testing here?

Capostrophic:
No, not really. I've filed an issue on the tracker now.

Capostrophic:
Ok, for some reason player ratings get messed up still. Investigating.

psi29a:
So this fixes things so that a new game doesn't get a borked value but all other saves before this commit is borked and un-fixable.

Capostrophic:
Yeah, but at least it's just the player in a standard game, and it's not like the player character uses AI ratings. Unless it's a clone, but the cloned player is not reliable anyway.

PR #2164, opened 2019-02-09
Allow localization to use implicit keywords

akortunov:
Attempt to fix [bug #4841](https://gitlab.com/OpenMW/openmw/issues/4841)

My understanding of dialogue system is a quite limited, so I do not know if this check was initially added for reason.

@MiroslavR, can you tell why did you add this check?


MiroslavR:
I have no memory of the code in question as quite some time has passed since I wrote it. That said, looking at it now, I believe (not 100% sure on that) the behavior of ignoring implicit keywords in the Russian version of the game was also present before my changes in PR #331.

kcat:
My memory is fuzzy, but I know some versions use an external file to map specific words in dialog to various topics (since some languages don't work so well to map a word/phrase used in dialog to a listed topic with the same word/phrase). I believe scrawl (or maybe Zini?) had a hand in initially handling that.

I can't recall why or if it was necessary to suppress the auto-topic-links on such versions, but on those versions there's no need to have them at least (somebody would've mentioned missing topic links if they were expected to appear; the lookup table apparently handles all of what's needed).

Capostrophic:
The problem is that cel, top and mrk files are not always present. Russian mods which do not use them end up with no hyperlinks in the dialogue. if for example there are top files for Bethesda master files but none for the mods themselves. They were intended to be autodetected.

And yes, topic links were missing. Russian Morrowind 1.0 version appears to lack the necessary files and thus is impossible to play. akortunov got a Russian plugin that expected the hyperlinks to get autodetected.

kcat:
I (personally) wouldn't worry about Morrowind 1.0.  Especially for localizations where things may need to be rushed to make the 1.0 release, and then altered and cleaned up a bit when they have time for subsequent patches. I don't think we should worry about handling data differences from unpatched versions of any language.

Obviously, though, if the latest Russian version does handle auto-topic-links (in addition to the topic map), it would make sense for OpenMW to as well. Does anyone have that version to test?

akortunov:
> Does anyone have that version to test?

According to two different persons which have a Russian localization, it indeed uses topic map as an addition to auto-topic-links, not as a replacement.

You can create a new topic with Russian name and it will be highligted if there will be an exact match in the dialogue response. Not sure which system takes priority, though, put I guess the topic map has a higher priority.

Capostrophic:
This is kind of important.

akortunov:
> This is kind of important.

Yes, it would be nice to have this issue fixed in 0.46.

psi29a:
Is there feedback still needed?

Are there any possible regressions here?

akortunov:
> Is there feedback still needed?

In the first place, it is unclear why this check was added at all (it was added a very long time ago by MIA contributors), so it is hard to tell where to search for regressions.

Capostrophic:
It can only cause regressions on Russian localization of Morrowind, you'll have to trust us that it won't. :P

PR #2165, opened 2019-02-10
[DO NOT MERGE YET]Add distant terrain to launcher (try #2)

terabyte25:
Based off my previous work in #1945

### DIFF
--- a/apps/launcher/graphicspage.cpp
+++ b/apps/launcher/graphicspage.cpp
@@ -2,6 +2,7 @@
 
 #include <boost/math/common_factor.hpp>
 #include <csignal>
+#include <limits>
 #include <QDesktopWidget>
 #include <QMessageBox>
 #include <QDir>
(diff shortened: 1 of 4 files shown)


apps/launcher/graphicspage.cpp
```diff
@@ -44,7 +45,8 @@ Launcher::GraphicsPage::GraphicsPage(Files::ConfigurationManager &cfg, Settings:
     connect(fullScreenCheckBox, SIGNAL(stateChanged(int)), this, SLOT(slotFullScreenChanged(int)));
     connect(standardRadioButton, SIGNAL(toggled(bool)), this, SLOT(slotStandardToggled(bool)));
     connect(screenComboBox, SIGNAL(currentIndexChanged(int)), this, SLOT(screenChanged(int)));
-
+    connect(distantTerrainCheckBox, SIGNAL(toggled(bool)), this, SLOT(slotDistantTerrainChanged(bool)));
+    viewingDistanceDoubleSpinBox->setMaximum(std::numeric_limits<float>::max());
```
terabyte25: I'm not really sure if this is a good way of setting a maximum to the viewing distance spin box.

akortunov:
As I said in previous PR, visibility spinbox does not have much sense for now since it conflicts with in-game visibility slider.
What we can do here:
1. Use a separate visibility distance if Distant Terrain is enabled (via new setting).
In this case in-game slider probably should alter this setting when Distant Terrain is enabled, also spinbox in the launcher can use an integer value in cells rather than the float one in game units.
2. Add a visibility multiplier setting and apply it when Distant Terrain is enabled.

I'd prefer the first option.

terabyte25:
Should distant fog be taken into account too?

davidcernat:
terabyte25 has continued engaging in trolling elsewhere despite our requests that he stop, so we're closing this for the time being.

PR #2166, opened 2019-02-11
Rework the viewing distance slider

akortunov:
EDIT:

Just increases a maximum value for visiblity slider and display output value in cells.
I still think that it is a quite stupid idea since short and long viewing distances should be handled completely differently (shadows, fog, etc.) and in-game settings should not provide graphics artifacts with any combinations.

### DIFF
--- a/files/mygui/openmw_settings_window.layout
+++ b/files/mygui/openmw_settings_window.layout
@@ -376,17 +376,29 @@
                         <Widget type="TextBox" skin="NormalText" position="4 130 322 18" align="Left Top" name="RenderDistanceLabel">
                             <Property key="Caption" value="#{sRender_Distance}"/>
                         </Widget>
-                        <Widget type="MWScrollBar" skin="MW_HScroll" position="4 154 322 18" align="Left Top">
-                            <Property key="Range" value="4667"/>
-                            <Property key="Page" value="300"/>
+                        <Widget type="MWScrollBar" skin="MW_HScroll" position="4 154 322 18" align="Left Top" name="RenderingDistanceSlider">
+                            <Property key="Range" value="4609"/>
+                            <Property key="Page" value="128"/>
                             <UserString key="SettingType" value="Slider"/>
                             <UserString key="SettingCategory" value="Camera"/>
                             <UserString key="SettingName" value="viewing distance"/>
                             <UserString key="SettingValueType" value="Integer"/>
-                            <UserString key="SettingMin" value="2000"/>
-                            <UserString key="SettingMax" value="6666"/>
+                            <UserString key="SettingMin" value="2048"/>
+                            <UserString key="SettingMax" value="6656"/>
                             <UserString key="SettingLabelWidget" value="RenderDistanceLabel"/>
                             <UserString key="SettingLabelCaption" value="#{sRender_Distance} (%s)"/>
+                        </Widget>
+                        <Widget type="MWScrollBar" skin="MW_HScroll" position="4 154 322 18" align="Left Top" name="LargeRenderingDistanceSlider">
+                            <Property key="Range" value="79873"/>
+                            <Property key="Page" value="2048"/>
+                            <UserString key="SettingType" value="Slider"/>
+                            <UserString key="SettingCategory" value="Camera"/>
+                            <UserString key="SettingName" value="viewing distance"/>
+                            <UserString key="SettingValueType" value="Cell"/>
+                            <UserString key="SettingMin" value="2048"/>
+                            <UserString key="SettingMax" value="81920"/>
+                            <UserString key="SettingLabelWidget" value="RenderDistanceLabel"/>
+                            <UserString key="SettingLabelCaption" value="#{sRender_Distance} (x%s)"/>
                         </Widget>
                         <Widget type="TextBox" skin="SandText" position="4 178 332 18" align="Left Top">
                             <Property key="Caption" value="#{sNear}"/>
--- a/apps/openmw/mwgui/settingswindow.cpp
+++ b/apps/openmw/mwgui/settingswindow.cpp
@@ -11,9 +11,12 @@
 
 #include <SDL_video.h>
 
+#include <iomanip>
+
 #include <components/debug/debuglog.hpp>
 #include <components/misc/stringops.hpp>
 #include <components/misc/gcd.hpp>
+#include <components/misc/constants.hpp>
 #include <components/widgets/sharedstatebutton.hpp>
 #include <components/settings/settings.hpp>
 
@@ -127,13 +130,22 @@ namespace MWGui
                 MyGUI::ScrollBar* scroll = current->castType<MyGUI::ScrollBar>();
                 std::string valueStr;
                 std::string valueType = getSettingValueType(current);
-                if (valueType == "Float" || valueType == "Integer")
+                if (valueType == "Float" || valueType == "Integer" || valueType == "Cell")
                 {
                     // TODO: ScrollBar isn't meant for this. should probably use a dedicated FloatSlider widget
                     float min,max;
                     getSettingMinMax(scroll, min, max);
                     float value = Settings::Manager::getFloat(getSettingName(current), getSettingCategory(current));
-                    valueStr = MyGUI::utility::toString((int)value);
+
+                    if (valueType == "Cell")
+                    {
+                        std::stringstream ss;
+                        ss << std::fixed << std::setprecision(2) << value/Constants::CellSizeInUnits;
+                        valueStr = ss.str();
+                    }
+                    else
+                        valueStr = MyGUI::utility::toString(int(value));
+
                     value = std::max(min, std::min(value, max));
                     value = (value-min)/(max-min);
 
@@ -146,7 +158,8 @@ namespace MWGui
                     scroll->setScrollPosition(value);
                 }
                 scroll->eventScrollChangePosition += MyGUI::newDelegate(this, &SettingsWindow::onSliderChangePosition);
-                updateSliderLabel(scroll, valueStr);
+                if (scroll->getVisible())
+                    updateSliderLabel(scroll, valueStr);
             }
 
             configureWidgets(current);
@@ -170,6 +183,12 @@ namespace MWGui
         WindowBase("openmw_settings_window.layout"),
         mKeyboardMode(true)
     {
+        bool terrain = Settings::Manager::getBool("distant terrain", "Terrain");
+        const std::string widgetName = terrain ? "RenderingDistanceSlider" : "LargeRenderingDistanceSlider";
+        MyGUI::Widget* unusedSlider;
+        getWidget(unusedSlider, widgetName);
+        unusedSlider->setVisible(false);
+
         configureWidgets(mMainWidget);
 
         setTitle("#{sOptions}");
@@ -425,7 +444,7 @@ namespace MWGui
         {
             std::string valueStr;
             std::string valueType = getSettingValueType(scroller);
-            if (valueType == "Float" || valueType == "Integer")
+            if (valueType == "Float" || valueType == "Integer" || valueType == "Cell")
             {
                 float value = pos / float(scroller->getScrollRange()-1);
 
@@ -436,7 +455,15 @@ namespace MWGui
                     Settings::Manager::setFloat(getSettingName(scroller), getSettingCategory(scroller), value);
                 else
                     Settings::Manager::setInt(getSettingName(scroller), getSettingCategory(scroller), (int)value);
-                valueStr = MyGUI::utility::toString(int(value));
+
+                if (valueType == "Cell")
+                {
+                    std::stringstream ss;
+                    ss << std::fixed << std::setprecision(2) << value/Constants::CellSizeInUnits;
+                    valueStr = ss.str();
+                }
+                else
+                    valueStr = MyGUI::utility::toString(int(value));
             }
             else
             {
--- a/docs/source/reference/modding/settings/camera.rst
+++ b/docs/source/reference/modding/settings/camera.rst
@@ -46,7 +46,7 @@ viewing distance
 
 :Type:		floating point
 :Range:		> 0
-:Default:	6666.0
+:Default:	6656.0
 
 This value controls the maximum visible distance (also called the far clipping plane).
 Larger values significantly improve rendering in exterior spaces,
@@ -87,7 +87,7 @@ Enabling the distant terrain setting is an alternative to increasing exterior ce
 Note that the distant land setting does not include rendering of distant static objects,
 so the resulting visual effect is not the same.
 
-This setting can be adjusted in game from the ridiculously low value of 2000.0 to a maximum of 6666.0
+This setting can be adjusted in game from the ridiculously low value of 2048.0 to a maximum of 81920.0
 using the View Distance slider in the Detail tab of the Video panel of the Options menu.
 
 field of view
--- a/files/settings-default.cfg
+++ b/files/settings-default.cfg
@@ -21,9 +21,9 @@ small feature culling = true
 
 small feature culling pixel size = 2.0
 
-# Maximum visible distance (e.g. 2000.0 to 6666.0).  Caution: this setting
+# Maximum visible distance. Caution: this setting
 # can dramatically affect performance, see documentation for details.
-viewing distance = 6666.0
+viewing distance = 6656.0
 
 # Camera field of view in degrees (e.g. 30.0 to 110.0).
 # Does not affect the player's hands in the first person camera.


psi29a:
@Capostrophic tested?

akortunov:
It is still unclear what to do with distant fog. Are we OK if it will be configurable only via config?

AnyOldName3:
I thought I'd already said this, but I'm not seeing any comments from me, so maybe it got lost in the ether...

I'm not super keen on this approach. I think it would be much better to just change the limit for the slider for the existing view distance setting when distant terrain is enabled than having two different settings that do exactly the same thing, but are used in different situations. Also, viewing distance is inherently a camera-related setting, and that's where people would think to look, so having it get overridden by another version somewhere else in another section is bad.

akortunov:
> I think it would be much better to just change the limit for the slider for the existing view distance setting when distant terrain is enabled than having two different settings that do exactly the same thing, but are used in different situations.

I am fine with this approach, but in this case we will need a launcher setting, which will increase the viewing distance automatically when a player enables the distant terrain, and will decrease it back, when the player disables it.

Also it is still unclear how to handle the distant fog feature.

akortunov:
In theory, the `distant land fog end` usually should be equal to viewing distance and the `distant land fog start` should be lesser than the viewing distance, so we can use the percentage setting instead.

akortunov:
@Capostrophic, @kcat, what do you think?

akortunov:
I suggest to replace the `distant land fog *` settings by the `distant land fog ratio *` settings.
By default we can use the 0.5x of viewing distance as the fig start and the 1.0x of viewing distance as the fog end. Both ratios should be configured via lanucher.

As about in-game slider, I am not sure that we should show values like 163840. Value in cells here could be more human-readable, but I am not sure if MyGUI allows us to use the custom output format without separate setting.
Slider setup should be something like this:
Min value - 8192 (1 cell).
Max value - 163840 (20 cells).
Move step - 8192 (1 cell).

Also launcher should allow to set the viewing distance when the distant terrain is enabled (in cells too, I suppose).

kcat:
> In theory, the `distant land fog end` usually should be equal to viewing distance and the `distant land fog start` should be lesser than the viewing distance, so we can use the percentage setting instead.

I'm not sure I agree. IMO, the fog should be constant regardless of the viewing distance. Even when playing without distant land, I prefer the less in-your-face fog provided by distant fog. As it is, setting the viewing distance/far-clip plane to the fog end doesn't hide the seam anyway, and I like being able to make out details near the far plane, weather permitting. I know that's not how vanilla behaves, so when distant fog isn't enabled, whatever. But when distant fog is enabled, and especially when distant land is enabled too, I don't think the fog end and view distance should be tied together.

akortunov:
Well, at least we should disable distant fog when the distant terrain is disabled.
Now I have to edit several settings every time when I need just to disable distant terrain for testing.
Ideally, I'd like to edit just the `distant terrain` settings without additional work.
Also separate fog distances (which are not configurable in-game) render in-game visibility slider useless.

TaxAkla:
First, thanks the team for OpenMW. 
Yesterday I was playing with the latest nigtly with both shadows and distant land (3 cells) active. When I was leaving the Underskar a ashstorm was happening and was surprised that I could see perfectly the Ghostgate and the Red Mountain.
So, I came here to remember that the distant fog should be linked with the weather like MGE does.

akortunov:
> So, I came here to remember that the distant fog should be linked with the weather like MGE does.

Per-weather fog settings is a completely different feature.

akortunov:
If we **really** want to keep independent fog distances, the simpliest solution would be just to disable the in-game visibility slider at all when the Distant Land is used.

In this case the whole configuration is supposed to be done only via launcher or manually via config file.

Also if the Distant Fog is supposed to be used separately from Distant Land, the Distant Fog term is a misnomer.

AnyOldName3:
So I can follow the discussion, what do the current fog modes actually do? Based on what I've picked up, I *think*:

* Regular fog starts at a fixed distance and ends at the far clip plane.
* Distant fog starts at a distance set by one setting, and ends at a distance set by another?

terabyte25:
@AnyOldName3 Isn't the documentation for distant fog?

AnyOldName3:
That's a fair point. Back in a minute...

AnyOldName3:
There's no RTD documentation, but looking at `settings-default.cfg`, there're a few words basically confirming what I had guessed.

I think the setting isn't very well named, as it has nothing to do with extended viewing distances, and, in fact, completely decouples fog from the viewing distance.

akortunov:
How about the "fixed distance fog"?

kcat:
> If we really want to keep independent fog distances, the simpliest solution would be just to disable the in-game visibility slider at all when the Distant Land is used.

Ideally, enabling distant land would remove the visible far clip plane and use some kind of "infinite" view distance. Basically, more efficient use of the depth buffer so the far plane can be pushed out farther than anything is rendered. Though implementing something like that is beyond me.

> I think the setting isn't very well named, as it has nothing to do with extended viewing distances, and, in fact, completely decouples fog from the viewing distance.

It does have the effect of moving the fog to be more distant, and not completely disable it (as distant land and large view distances otherwise would).

Originally, the distant fog was going to be tied to distant land, but others wanted the two to be separate. So rather than the distant land setting implying distant fog behavior, it was made a separate toggle.

AnyOldName3:
> > If we really want to keep independent fog distances, the simpliest solution would be just to disable the in-game visibility slider at all when the Distant Land is used.
> 
> Ideally, enabling distant land would remove the visible far clip plane and use some kind of "infinite" view distance. Basically, more efficient use of the depth buffer so the far plane can be pushed out farther than anything is rendered. Though implementing something like that is beyond me.

It's kind of beyond *anyone*, really. Being able to see from one corner of Project Tamriel to the other would be more depth buffer range than is remotely feasible. Also, wanting more than the default viewing distance doesn't necessarily equate to wanting infinite viewing distance. You don't get infinite viewing distance on Earth because there's a horizon, and it's only a concern in Morrowind because Nirn is flat.

> > I think the setting isn't very well named, as it has nothing to do with extended viewing distances, and, in fact, completely decouples fog from the viewing distance.
> 
> It does have the effect of moving the fog to be more distant, and not completely disable it (as distant land and large view distances otherwise would).

That's wholly dependent on the setting values. If I enabled it with the default viewing distance and regular terrain, and set the fog start to 0 and end to 100, the fog would most certainly *not* be distant.

akortunov:
So what did we decide to do?

kcat:
> It's kind of beyond *anyone*, really. Being able to see from one corner of Project Tamriel to the other would be more depth buffer range than is remotely feasible.

https://outerra.blogspot.com/2012/11/maximizing-depth-buffer-range-and.html

The problem is the default depth calculation gives way more precision to the area nearest to the near-clip plane than is necessary. You can do things differently to get a better distribution, which can give you all that you would need.

BTW, the issue that post mentions with OpenGL can actually be more cleanly fixed than the workarounds it prescribes, but would need a relatively newer extension (`GL_ARB_clip_control`), which is where the difficultly partly comes in. The other is in ensuring the depth test modes are properly reversed given the reversed depth range.

> Also, wanting more than the default viewing distance doesn't necessarily equate to wanting infinite viewing distance. You don't get infinite viewing distance on Earth because there's a horizon, and it's only a concern in Morrowind because Nirn is flat.

Sure, there can also be a subtle curvature applied to the terrain (not something you could really notice with the size of the game world, even with TR), which would offer an absolute max distance we can work with for the generated default cells that surround the game world. That's kind of why I put "infinite" in quotes, because it's not technically infinite, but far enough out that no one could complain about it being too near.

And yeah, someone could want distant terrain and not infinite view distance, but I suspect that'd be a minority. Especially with fog obscuring everything past 40960 units. People who want it could still do it of course, but I don't think it would be necessary to cater to them to the point that it hinders the majority of people who use distant land.

> That's wholly dependent on the setting values. If I enabled it with the default viewing distance and regular terrain, and set the fog start to 0 and end to 100, the fog would most certainly not be distant.

Sure, just like if you enable distant land but keep the view distance low, you won't get distant land. Or if you enable specular/parallax, but change the postfix names that no mod uses. If you tweak config settings from their intended/default values, you should know what you're getting into.

AnyOldName3:
The real problem is just that the fog isn't quite the same colour as the skybox, so you don't actually see it how fog would look and see silhouettes at the clip plane. Really, the fog and skybox should blend together as if there were particulates in the air stopping you from seeing anything, and there shouldn't be a silhouette.

Another option, if we can guarantee that fog will be at 100% at the clip plane, is to use depth clamping to ensure there's the right silhouette for everything, no matter how far away. The depth sorting will be broken, but 100% fog is 100% fog, no matter what's in front of what. However, this eliminates the main benefit of a short viewing distance - the extra performance, as everything still gets drawn.

kcat:
> The real problem is just that the fog isn't quite the same colour as the skybox, so you don't actually see it how fog would look and see silhouettes at the clip plane. Really, the fog and skybox should blend together as if there were particulates in the air stopping you from seeing anything, and there shouldn't be a silhouette.

That only works when objects at the far-clip plane are small enough to not be above the horizon line. Without simulating a planetary curve, and especially with a low view distance, it's easy to have objects that get clipped out when they're against the colored sky and not obstructed by the atmosphere, creating the visible seam.

AnyOldName3:
I'm not sure I understand what you're saying. If the fog and skybox are the same colour, I don't see how you can have a seam.

kcat:
The skybox itself isn't a flat color, though. It has a blue sky and/or clouds, which fade to the fog color at the horizon. If an object is not below the horizon line when it's clipped, you'll see the fogged object cutting against the blue sky or clouds. [Example](https://en.uesp.net/wiki/File:MW-place-Molag_Amur.jpg). See the far clip plane where the terrain is obscured in fog, but the sky is fading from the blue above and the fog color below. Especially as you move forward and backward, that seam becomes much more apparent.

If you had a higher viewing angle and were looking downward, the seam may be more obscured by the fog and sky having the same color, but not against taller objects/terrain and certainly not at normal viewing height.

kcat:
That would also give the appearance of objects fading in/out translucently, rather than by fog. A better idea may be to raise the apparent vertical level that the skybox becomes the fog color, though that would certainly increase the perception of a constant nearby wall of fog (which some players may be fine with, but not others).

AnyOldName3:
I see what you were trying to say now. I was talking about a hypothetical solution where the fog and skybox were the same colour (however it was achieved), not the current situation where they're different colours because we use a constant fog colour and a texture for the skybox. As has been mentioned, what I was suggesting could be achieved by either using the skybox colour for the fog, or changing the skybox so it was always fog coloured near fogged objects.

I kind of feel that the skybox is trying to suggest that there's thick fog at the height the player is at (so the horizon is completely fogged out), but not much fog above the player (so you can see the sky fairly clearly). Maybe blending more fog into the skybox the lower the player goes could help and be fixed-function-compatible - if you're at the top of Red Mountain on a clear day, there isn't going to be anything fogged out above you, so it's fine to only have fog colour below the horizon, but if you're at sea level, you could be looking at many things higher than you, so the fog would need to be above the horizon by quite a way.



The depth-clamping-and-draw-everything-to-get-silhouettes idea would potentially give the impression that there was fog in the distance, but it gets less dense the further away from the player it is (otherwise all distant mountains wouldn't be visible at all).



Really, the best solution would be to stop caring about fixed-function compatibility and add an optional pretty fog shader that made fog height-dependent and realistic, and people who cared about fog blending would have to accept that they'd need a non-garbage GPU.

kcat:
> I kind of feel that the skybox is trying to suggest that there's thick fog at the height the player is at (so the horizon is completely fogged out), but not much fog above the player (so you can see the sky fairly clearly).

I think it's more trying to simulate atmospheric scattering. The atmosphere itself isn't 100% transparent, so the farther you look out to the horizon, the things farther out there become more obscured, even affecting how you see the distant sky. I think if it was trying to simulate thick fog on the skybox, the line where it becomes the fog color would be much higher and better blend with the fog at the far plane. As it is, if you use distant land with more distant fog settings, the skybox still generally fits.

AnyOldName3:
> I think it's more trying to simulate atmospheric scattering. The atmosphere itself isn't 100% transparent, so the farther you look out to the horizon, the things farther out there become more obscured, even affecting how you see the distant sky

That's kind of what I was meaning (or at least similar) - at lower altitudes, the atmosphere is denser and has more particulates, so if effectively thicker fog. I don't really see any point in distinguishing between atmospheric scattering and any other kind of fog.

Also, apparently Nirn might not be flat after all. The only definitive thing we have is that there's a globe in Daggerfall, but it's muddied by the number of times it's referred to as a plane and the number of times other things that aren't planar are also referred to as planes.

MaxMaslov:
> Also, apparently Nirn might not be flat after all. The only definitive thing we have is that there's a globe in Daggerfall, but it's muddied by the number of times it's referred to as a plane and the number of times other things that aren't planar are also referred to as planes.

Nirn is definitely a planet. Look at Oblivion's Orrery
https://en.uesp.net/wiki/File:OB-interior-Orrery_(labeled).jpg

kcat:
> Also, apparently Nirn might not be flat after all. The only definitive thing we have is that there's a globe in Daggerfall, but it's muddied by the number of times it's referred to as a plane and the number of times other things that aren't planar are also referred to as planes.

"Plane" can also mean realm or dimension, not necessarily a flat planar surface.

terabyte25:
Does OpenMW have to adhere to lore anyway?

terabyte25:
@kcat [Well Nirn is shown to be spherical.](https://aurbis.c0da.es/c0da/s/ksenia/apocalypse_v5_2560.png)

terabyte25:
[Taken from here.](https://forums.elderscrollsonline.com/en/discussion/464613/is-tamriel-flat-in-topological-space)

> Tamriel is a continent on a sphere, just as Masser and Secunda are spheres. The fact that Masser and Secunda show crescent forms in various phases demonstrates they are spheres.

psi29a:
Looks like some rebasing is in order...

akortunov:
> Looks like some rebasing is in order...

We need to determine which approach to use. There is still no agreement here...

akortunov:
So we have two main approaches here:
1. Hide the in-game visibility slider, when the Distant Terrain is enabled.
In this case we will keep fog settings unrelated to camera settings, but we will not have an ingame control for visibility distance.

2. Adjust visibility slider to have an another range, when the Distant Terrain is enabled.
In this case we will need to bind fog settings to the visibility distance, but will have an ingame control.

Also we may or may not use a separate setting for viewing distance for Distant Terrain.
1. Without the setting, we will need to implement a custom logic in the launcher, which will set viewing distance to 6666 every time when user disables the DIstant Terrain and set it to the larger value (the 40960, for example) every time when he enables the Distant Terrain.
 
2. With setting, we can keep separately configured values, so the launcher just will need to change the `distant terrain` setting value to enable/disable feature properly.

AnyOldName3:
I am incredibly against having two different settings (and if I have veto powers, I'll exercise them).

I'm in favour of having the visibility slider change its range based on whether distant terrain is enabled. Clamping the view distance to this range when the launcher or engine detects a mismatch should be fine.

Capostrophic:
I agree with AnyOldName3 here. It should be as simple as having 5x higher range without any consideration for different units (number of cells).

Though I'm not so sure about clamping the viewing distance if distant terrain is disabled. Having higher than one cell viewing distance is generally fine for certain debug even though the ocean is visible, and distant terrain is going to be hardcoded "on" at some point anyway once it's optimized enough to completely replace the standard terrain system without side effects, so there's no use in extra effort going towards limiting it.

akortunov:
> It should be as simple as having 5x higher range without any consideration for different units (number of cells).

I doubt it is a good idea now since we show the setting value near slider. Values such as 57344 are not human-readable.

kcat:
What about showing the values in cells, rather than units? Or perhaps both, like `57344 (7.0 cells)`. That'll make it a bit more human-readable.

In either case, I too would prefer being able to set a larger view distance even without distant terrain; I'm fine with pop-in if it means hiding the far-plane seam during movement.

akortunov:
> What about showing the values in cells, rather than units?

I doubt that our UI currently supports it.

akortunov:
All right, I just increase a maximum value for that slider now.

I still think that it is a stupid approach since we will get bugreports from end-users about broken rendering on large viewing distances, because the "viewing distance" setting does not adjust the distant terrain, distant fog, split point ratio for shadows, etc.

AnyOldName3:
We don't *have* to make the viewing distance slider not change the near clip distance, too. I'd be surprised if this wasn't the usual approach in other software.

Capostrophic:
>I still think that it is a stupid approach since we will get bugreports from end-users about broken rendering on large viewing distances, because the "viewing distance" setting does not adjust the distant terrain, distant fog, split point ratio for shadows, near clip to reduce the z-fighting, etc.

I thought it was pretty clear from AnyOldName3's idea that the limit should only be raised if distant terrain is enabled.

akortunov:
> I thought it was pretty clear from AnyOldName3's idea that the limit should only be raised if distant terrain is enabled.

[Sigh]

files/mygui/openmw_settings_window.layout
```diff
@@ -376,15 +376,27 @@
...
-                        <Widget type="MWScrollBar" skin="MW_HScroll" position="4 154 322 18" align="Left Top">
-                            <Property key="Range" value="4667"/>
-                            <Property key="Page" value="300"/>
+                        <Widget type="MWScrollBar" skin="MW_HScroll" position="4 154 322 18" align="Left Top" name="RenderingDistanceSlider">
+                            <Property key="Range" value="4609"/>
+                            <Property key="Page" value="128"/>
                             <UserString key="SettingType" value="Slider"/>
                             <UserString key="SettingCategory" value="Camera"/>
                             <UserString key="SettingName" value="viewing distance"/>
                             <UserString key="SettingValueType" value="Integer"/>
```
Capostrophic: Should this be Cell instead?

Capostrophic:
Of course there's no "cells" GMST. Maybe the number should have "x" added to it or something, or the English word even.

files/mygui/openmw_settings_window.layout
akortunov: Not sure about it. It has the 0.25-0.8 range.

PR #2167, opened 2019-02-12
[Discussion, don't merge] Editor: Transient land shape editing

unelsson:
Based on Terrain texture selection (https://github.com/OpenMW/openmw/pull/2517).

This PR edits ESMterrain::Storage on component-level. I'm not too sure how it should work, so reviewing especially the changes on the components would be helpful. Probably should be thoroughly reviewed and tested before even thinking of merging.

Additionally, terrain editing feature allows editing outside of the range of ESM file format. This PR now implements a workaround that highlights the broken land heights as red whenever the edit goes out of range. This is a workaround as it tells the user that editing is going too far, and saving would then corrupt the terrain data, but I'd really like to find another way to solve this.

Features:
- Terrain vertex selection.
- Terrain shape editing
- Transient land shape editing (press esc to undo). Changes made to land are held in float array until the drag end. Memory is not allocated runtime, but rather all cells constantly hold an extra float array. This feature overrides a function from Storage, shouldn't affect OpenMW game engine.

TODO:
- [x] Fix cell edges breaking, when cells are not pre-loaded.
- [x] Add drag-amount to heights change
- [x] Proper naming of variables and functions in alterheights and heightmap to alterheight and height
- [x] Limit land edits to steps of +-8, in order to avoid land tearing because of rounding when saving.
- [x] Limit slopes to fit land format (extreme changes will break the save)
- [x] Develop basic tools for land editing
- [x] Fix bug: saves break, investigate when and why (reason: missing ltex records)
- [x] Fix bug: limitHeightChanges function breaks land, remove duplicated code
- [x] Fix bug: limitHeightChanges still rarely breaks land corners and edges
- [x] Fix bug: Out-of-limits land-edits, strange hit values?
- [x] Fix bug: Land normals command is issued after land shape command (resulting in wrong normals)
- [ ] Fix bug: Tool size 1 doesn't do anything with circle/square brush
- [ ] Fix bug: Smooth next to a broken cell edge doesn't calculate average properly. Note: Other fixes made the land break less, and function fixEdges can sometimes be used as workaround.
- [x] Improve: Make smooth-tool properly paintable
- [x] Improve: limitHeightChanges follows strange logic, and it's unknown if it always works, it shouldn't have to fix cell edges. -> Fixed the logic, but still breaks some corners in special cases. Removed the whole limit feature, opting for broken land highlight and manual edit.
- [ ] Improve: Some cells (e.g. -1, -17 in Morrowind) seems to have land record, but no land is visible, even and terrain heights set to zero. They should also be editable in the future.
- [x] Improve: Highlight excessive height changes/broken land in view window
- [x] Improve: Add label to toolstrength-slider
- [ ] Improve: Ability to re-attach broken cell edges (already done: function fixEdges)
- [x] Improve: When creating new land next to existing land, seam cell edge propely
- [x] Improve: Painting height on empty cell (no land) or non-existing cell should create new land. Note: Fundamentally works, but some cells (e.g. -1, -17) still refuse to be visible.
- [ ] Extra: Bump -tool
- [x] Extra: Plateau/flatten -tool
- [x] Check, and then uncomment or remove commented stuff at terrain and terrainstorage.
- [ ] Testing and further design, planning of advanced tools
- [x] Fix changelog (no hurry with this one though, merge conflict prevents useless builds while updating this)

Planning/feedback needed for:
1) UI/UX, feedback on tools and general usability.
2) Transient land change implementation (terrainstorage.cpp), any objections or better ideas?
3)  C++ style in general
4) Memory management (I've tried to make it as automatic as possible). Should transient edits allocate memory dynamically, use std::vector instead of float[]?
5) Terrain vertex selection drawing, how does it look, tips on optimization
6) Limits and how they are implemented, currently limits are at terrainstorage (always steps of +-8 per change), and max height change is limited at command-phase (end of drag), limiting cuts extreme land edits to those that shouldn't break during mod save.
7) Currently circle brush is the only tool with soft edges, all other tools have hard edges. Hardness/softness can both be useful at times. Should there also be a user option for brush hardness, or some other default setting? Implementing hardness/softness for custom brush could be based on the distance from the center of the custom brush, or a gaussian blur on top of selection shape (using a gaussian blurred "opacity map" to determine the tool strength in any vertex).

https://youtu.be/YtEL6MMvw9w (old video, bugs seen here should be fixed by now)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -158,6 +158,8 @@
     Feature #4784: Launcher: Duplicate Content Lists
     Feature #4812: Support NiSwitchNode
     Feature #4836: Daytime node switch
+    Feature #4840: Editor: Transient terrain change support
+    Feature #????: Editor: Land shape editing, land selection
     Feature #4859: Make water reflections more configurable
     Feature #4882: Support for NiPalette node
     Feature #4887: Add openmw command option to set initial random seed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,6 +143,7 @@
     Feature #3025: Analogue gamepad movement controls
     Feature #3442: Default values for fallbacks from ini file
     Feature #3610: Option to invert X axis
+    Feature #3871: Editor: Terrain Selection
     Feature #3893: Implicit target for "set" function in console
     Feature #3980: In-game option to disable controller
     Feature #3999: Shift + Double Click should maximize/restore menu size
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -89,7 +89,7 @@ opencs_units (view/render
     scenewidget worldspacewidget pagedworldspacewidget unpagedworldspacewidget
     previewwidget editmode instancemode instanceselectionmode instancemovemode
     orbitcameramode pathgridmode selectionmode pathgridselectionmode cameracontroller
-    cellwater terraintexturemode actor
+    cellwater terraintexturemode actor terrainselection
     )
 
 opencs_units_noqt (view/render
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -82,14 +82,14 @@ opencs_units_noqt (view/world
 
 opencs_units (view/widget
     scenetoolbar scenetool scenetoolmode pushbutton scenetooltoggle scenetoolrun modebutton
-    scenetooltoggle2 scenetooltexturebrush completerpopup coloreditor colorpickerpopup droplineedit
+    scenetooltoggle2 scenetooltexturebrush scenetoolshapebrush completerpopup coloreditor colorpickerpopup droplineedit
     )
 
 opencs_units (view/render
     scenewidget worldspacewidget pagedworldspacewidget unpagedworldspacewidget
     previewwidget editmode instancemode instanceselectionmode instancemovemode
     orbitcameramode pathgridmode selectionmode pathgridselectionmode cameracontroller
-    cellwater terraintexturemode actor terrainselection
+    cellwater terraintexturemode actor terrainselection terrainshapemode
     )
 
 opencs_units_noqt (view/render
--- a/apps/opencs/model/prefs/state.cpp
+++ b/apps/opencs/model/prefs/state.cpp
@@ -170,7 +170,7 @@ void CSMPrefs::State::declare()
         "list go to the first/last item");
 
     declareCategory ("3D Scene Input");
-    
+
     declareDouble ("navi-wheel-factor", "Camera Zoom Sensitivity", 8).setRange(-100.0, 100.0);
     declareDouble ("s-navi-sensitivity", "Secondary Camera Movement Sensitivity", 50.0).setRange(-1000.0, 1000.0);
     declareSeparator();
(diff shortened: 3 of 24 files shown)


vakhoir:
Hi! I found a small bug. Looks like a literal corner case, the CS crashes when you finish editing land on a cell that's at the edge of the map. The crash is at openmw/apps/opencs/view/render/terrainshapemode.cpp line:286. Looks like landDownShapePointer is empty during one of the iterations (i=0, j=64).

I can send you the crash logs, but it's easily reproducible. Just create a new game file, add a single cell, and land, and try to shape it.

unelsson:
> Hi! I found a small bug. Looks like a literal corner case, the CS crashes when you finish editing land on a cell that's at the edge of the map. The crash is at openmw/apps/opencs/view/render/terrainshapemode.cpp line:286. Looks like landDownShapePointer is empty during one of the iterations (i=0, j=64).
> 
> I can send you the crash logs, but it's easily reproducible. Just create a new game file, add a single cell, and land, and try to shape it.

Thanks for reporting. This bug was present in the early versions of function that limits land heights. However, I fixed crash bugs I could find. Then I finally removed automatic use of that function, as the limiting function couldn't properly read height changes at Morrowind's edge cells (that for some reason don't identify landless, yet also don't have any land data). I can't reproduce this bug in latest version though, and there is no variable called landDownShapePointer in terrainshapemode.cpp any more. Have you tested this with the latest version?

vakhoir:
> I can't reproduce this bug in latest version though, and there is no variable called landDownShapePointer in terrainshapemode.cpp any more. Have you tested this with the latest version?

I'm a bit confused, what do you mean by latest version? Of OpenMW? I just pulled the master branch and tested again, and still get the error. I noticed you have 2 other PRs, but in both branches terrainshapemode.cpp does have a landDownShapePointer variable. Which one should I test?

unelsson:
uhh, right. It seems like I had a git hiccup at some point. I'll investigate.

> > I can't reproduce this bug in latest version though, and there is no variable called landDownShapePointer in terrainshapemode.cpp any more. Have you tested this with the latest version?
> 
> I'm a bit confused, what do you mean by latest version? Of OpenMW? I just pulled the master branch and tested again, and still get the error. I noticed you have 2 other PRs that are more recent, but in both branches terrainshapemode.cpp does have a landDownShapePointer variable. Which one should I test?

unelsson:
I see now the point where this bug should trigger, but for some reason I haven't been able to reproduce a crash. Could you tell me exact way you reproduce this? There's a fix pending, but didn't commit it yet as I'd really like to test first if it works.

vakhoir:
Yeah, like I mentioned I just created a new game file, added a single cell, and a land record and tried shaping the land. It crashes 100% of the time.

It stops crashing when I add cells around the first one, but they all need to have lands in them as well. If I try to shape land surrounded by empty cells CS still crashes.

unelsson:
> Yeah, like I mentioned I just created a new game file, added a single cell, and a land record and tried shaping the land. It crashes 100% of the time.
> 
> It stops crashing when I add cells around the first one, but they all need to have lands in them as well. If I try to shape land surrounded by empty cells CS still crashes.

It should indeed crash, but for me it doesn't, no matter what I do. Anyway, I've added a check for null land/cell in commit 1cf6469. Could you please check if it still crashes?

vakhoir:
Works perfectly now!

unelsson:
Recent changes to terrain system broke this PR, and apparently my git changes were incompatible with other changes. Rebasing with incompatible changes resulted in a mess that can probably only be resolved by more or less building a new commit history from the beginning.

unelsson:
Latest commits fixed this PR back into life, so it does compile, and worked okay in quick testing, but no guarantees on code quality.

PR #2168, opened 2019-02-15
Option to disable controller (feature #3980)

Capostrophic:
[Feature 3980](https://gitlab.com/OpenMW/openmw/issues/3980).

Adds a button into Controls menu which "toggles" the controller support, the respective option. When it's off, all, or at least most of the controller input events are ignored by the input manager so the detected controller is effectively disabled.

~~The major issue is that I don't have a controller and cannot tell if it works as it should.~~
Independent testing reports that this works as intended.

Controls section of the in-game settings looks like this at the moment. We're tight on space.

Note that in Morrowind the equivalent button is forced off when a joystick is not detected. However:
1. I don't think there's an easy-to-use way to have unique functionality for generic MyGUI-based setting checkboxes
2. While the setting has no effect when there isn't any controller in the system, it might be less confusing for the user when the button state explicitly shows whether controllers are used or not.

screenshot004

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
     Feature #2229: Improve pathfinding AI
     Feature #3442: Default values for fallbacks from ini file
     Feature #3610: Option to invert X axis
+    Feature #3980: In-game option to disable controller
     Feature #4209: Editor: Faction rank sub-table
     Feature #4673: Weapon sheathing
     Feature #4730: Native animated containers support
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -55,6 +55,7 @@ namespace MWInput
         , mInvertX (Settings::Manager::getBool("invert x axis", "Input"))
         , mInvertY (Settings::Manager::getBool("invert y axis", "Input"))
         , mControlsDisabled(false)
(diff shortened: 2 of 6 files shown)


terabyte25:
https://wiki.openmw.org/index.php?title=GMSTs_(status) should sEnableJoystick be changed to implemented once this PR gets merged?

Capostrophic:
Yes, obviously.

akortunov:
Can we detect if controller is actually available and hide this button (with Controller tab) if it is not available?

terabyte25:
@akortunov https://gamedev.stackexchange.com/questions/73374/sdl-gamecontroller-automatic-detection perhaps?

Capostrophic:
Pretty sure it could be possible to just track controllerAdded and controllerRemoved events somehow and relay them to the window manager and settings window which would then hide/show the redundant widgets. But is it even in the scope of this PR?

akortunov:
> But is it even in the scope of this PR?

Morrowind at least does not allow to enable this setting if it can not find controller.

wareya:
What if I want to change the setting when my controller isn't plugged in because of some weird driver problem?

terabyte25:
@wareya You modify the settings.cfg file.

wareya:
:|

psi29a:
Great, I'll take it.

PR #2169, opened 2019-02-16
Add resources to path list

Delphik:
The resources folder is required for the fonts section of the wiki, but the path isn't stated anywhere. I'm on linux and don't the paths for other OS'.

Made a typo in the heading, new to github, not sure how to fix without making new request

### DIFF
--- a/docs/source/reference/modding/paths.rst
+++ b/docs/source/reference/modding/paths.rst
@@ -13,6 +13,13 @@ Configuration files and log files
 :Mac:		``$HOME/Library/Preferences/openmw``
 :Windows:	``C:\Users\Username\Documents\my games\openmw``
 
+Screenshots
+-----------
+
+:Linux:		``/usr/share/games/openmw/resources/``
+:Mac:		``?``
+:Windows:	``?``
+
 Savegames
 ---------
 


terabyte25:
> Made a typo in the heading, new to github, not sure how to fix without making new request

Use git rebase, or edit the file in the branch.

Capostrophic:
> > Made a typo in the heading, new to github, not sure how to fix without making new request
> 
> Use git rebase, or edit the file in the branch.

You're not in the position to make recommendations in this area, you know.
_________________________________________________________
I'd rather not have any specific resource paths in the documentation. Reasons being:
1. On Windows the resources folder path is simply the installation folder and the user can tell straight-away where is it. I can't tell what is the situation on macOS.
2. On Linux, the path depends on OpenMW distribution *and* the path to user home folder. The path to resources depends on how OpenMW was installed (it could be a system-wide package or user-specific package, and even then the resources folder may as well be in the folder OpenMW was installed too or built in). In this case it's not even accurate. Even then, not every distribution repository has openmw, and the only truly official system-wide builds are for Ubuntu, while generic binaries exist that fit everything must be installed into a dedicated folder - resources folder will still not be in /usr/share.
3. **User should not mess with files in the resources folder in general case.** In case of fonts, you should put openmw_font.xml file and TTF/OTF files into Fonts folder of the configuration file path.

But that's just my opinion.

Capostrophic:
The change is going to be outdated regardless depending on how the discussion of PR 2321 goes on, this is basically a "broken" PR that doesn't appear to be worked on so I'm closing it.

PR #2170, opened 2019-02-16
Add option to disable DetourNavigator component (feature #4833)

elsid:
Allow to disable DetourNavigator by option `enable = false` in `Navigator` settings section. If set to `false`:
* no background threads will be launched
* no navmeshes are created and rendered
* `findPath` method will always return empty path which follows to fallback to old pathfinding with pathgrid
* `toggleactorspath` command still works


psi29a:
@Capostrophic @akortunov what do you guys think?

Capostrophic:
I mean, it looks fine, but I don't know enough about the navigator to tell what's a no-no.
I can only try and test if nothing broke on Sunday.

akortunov:
Also with `enable = false` these bugs do not appear:

https://gitlab.com/OpenMW/openmw/issues/4785
https://gitlab.com/OpenMW/openmw/issues/4832

Capostrophic:
And this issue too.
https://gitlab.com/OpenMW/openmw/issues/4814

Seems to be fine.

psi29a:
Great! :)

PR #2171, opened 2019-02-16
Fix update navmesh

elsid:
Every updated object should produce a set of changed tiles where it is placed. Before this change only current object tiles were updated. If object was moved to another set of tiles then navmesh were not changed in new tiles.

`TileCachedRecastMeshManager::updateObject` should add all new tiles if object was moved and remove all no more used tiles. Both new and old tiles should be marked as changed.

Also add tests to show desired result for add, update, remove.


psi29a:
Thanks!

PR #2172, opened 2019-02-17
Fix some issues, found by PVS-Studio

akortunov:
1. Do not throw by reference ([V1022](https://www.viva64.com/en/w/v1022/)).
2. Simplify some checks where we already check the same condition above.
3. Qt::CaseInsensitive is supposed to be an argument of endsWith() call for case-insensitive search.
4. Fix possible overrun of the ESM::Skill::sSkillNameIds array with skillId = 27 (could happen only in Release mode because of correct assert), simplify loop.
5. Init some missing variables
6. Previously the `playSound` function in the WindowsManager had the bool argument, now it has two optional float arguments instead, but I forgot to change a couple of calls. Fixed.
7. Since we init the `encoder` variable in the stack and pass reference to it to the mEncoder field, it can become invalid outside the Engine::go() call. A suggested solution: init encoder object in the heap and destroy it in the engine destructor.

### DIFF
--- a/apps/openmw/mwgui/hud.cpp
+++ b/apps/openmw/mwgui/hud.cpp
@@ -262,7 +262,7 @@ namespace MWGui
 
             if (mode == GM_Console)
                 MWBase::Environment::get().getWindowManager()->setConsoleSelectedObject(object);
-            else if ((mode == GM_Container) || (mode == GM_Inventory))
+            else //if ((mode == GM_Container) || (mode == GM_Inventory))
             {
                 // pick up object
                 if (!object.isEmpty())
--- a/apps/openmw/mwgui/journalwindow.cpp
+++ b/apps/openmw/mwgui/journalwindow.cpp
@@ -631,7 +631,7 @@ namespace
 
                 if (page+2 < book->pageCount())
                 {
-                    MWBase::Environment::get().getWindowManager()->playSound("book page", true);
+                    MWBase::Environment::get().getWindowManager()->playSound("book page");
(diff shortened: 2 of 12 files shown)


apps/openmw/mwgui/journalwindow.cpp
```diff
@@ -573,7 +573,7 @@ namespace
             if (mAllQuests)
             {
                 SetNamesInactive setInactive(list);
-                mModel->visitQuestNames(!mAllQuests, setInactive);
+                mModel->visitQuestNames(false, setInactive);
```
Capostrophic: https://github.com/OpenMW/openmw/pull/1681#discussion_r181964558

apps/openmw/mwgui/hud.cpp
```diff
@@ -262,7 +262,7 @@ namespace MWGui
 
             if (mode == GM_Console)
                 MWBase::Environment::get().getWindowManager()->setConsoleSelectedObject(object);
-            else if ((mode == GM_Container) || (mode == GM_Inventory))
```
Capostrophic: Maybe just comment the check out instead of completely removing it?

elsid:
Could we integrate PVS-studio check into our CI?

akortunov:
> Could we integrate PVS-studio check into our CI?

Currently it shows too many false-positives, especially for external code. I suppose we will need to tweak it somehow in this case.

psi29a:
Thanks!

PR #2173, opened 2019-02-17
[WIP] Optimize navmesh update (bug #4785)

elsid:
I've tried to do some simple optimization. Got about ~10% performance increase for navmesh update. Then I've added a profiling for `Engine::frame` with detailed report for each frame. All `Navigator` method are profiled. It produces text file when player quits game like [this](https://gist.github.com/elsid/9a393819832378239a7e6718bd2db526). Also I've made a [script](https://gist.github.com/elsid/532e461dfc687975a4d8f71d38a6acd6) to produce [report](https://gist.github.com/elsid/13e3320b36ffdd00a4b3bcb5d63bf45a) with some statistics and detailed profile for each frame with duration greater than 1/60 second exclude all frames with `Scene::loadCell` call.

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -69,6 +69,8 @@ namespace
 
 void OMW::Engine::executeLocalScripts()
 {
+    const ::ProfileScope profile("Engine::executeLocalScripts");
+
     MWWorld::LocalScripts& localScripts = mEnvironment.getWorld()->getLocalScripts();
 
     localScripts.startIteration();
@@ -85,6 +87,8 @@ bool OMW::Engine::frame(float frametime)
 {
     try
     {
+        const ::ProfileScope profile("Engine::frame");
+
         mStartTick = mViewer->getStartTick();
 
         mEnvironment.setFrameDuration(frametime);
@@ -752,6 +756,8 @@ void OMW::Engine::go()
 
     // Save user settings
     settings.saveUser(settingspath);
+
+    ::Profiler::instance().save();
 
     Log(Debug::Info) << "Quitting peacefully.";
 }
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -412,6 +412,8 @@ namespace MWInput
 
     void InputManager::update(float dt, bool disableControls, bool disableEvents)
     {
+        const ::ProfileScope profile("InputManager::update");
+
         mControlsDisabled = disableControls;
 
         mInputManager->setMouseVisible(MWBase::Environment::get().getWindowManager()->getCursorVisible());
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -27,6 +27,8 @@
 
 #include <components/sceneutil/positionattitudetransform.hpp>
 
+#include <components/debug/debuglog.hpp>
+
 #include "../mwrender/animation.hpp"
 
 #include "../mwbase/environment.hpp"
--- a/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
+++ b/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp
@@ -28,6 +28,8 @@
 #include "actorutil.hpp"
 #include "combat.hpp"
 
+#include <components/debug/debuglog.hpp>
+
 namespace
 {
 
@@ -298,6 +300,8 @@ namespace MWMechanics
 
     void MechanicsManager::update(float duration, bool paused)
     {
+        const ::ProfileScope profile("MechanicsManager::update");
+
         if(!mWatched.isEmpty())
         {
             MWBase::WindowManager *winMgr = MWBase::Environment::get().getWindowManager();
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -1239,6 +1239,8 @@ namespace MWPhysics
 
     const PtrVelocityList& PhysicsSystem::applyQueuedMovement(float dt)
     {
+        const ::ProfileScope profile("PhysicsSystem::applyQueuedMovement");
+
         mMovementResults.clear();
 
         mTimeAccum += dt;
--- a/apps/openmw/mwscript/scriptmanagerimp.cpp
+++ b/apps/openmw/mwscript/scriptmanagerimp.cpp
@@ -91,6 +91,8 @@ namespace MWScript
 
     void ScriptManager::run (const std::string& name, Interpreter::Context& interpreterContext)
     {
+        const ::ProfileScope profile("ScriptManager::run");
+
         // compile script
         ScriptCollection::iterator iter = mScripts.find (name);
 
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -663,6 +663,7 @@ namespace MWSound
                                      float volume, float pitch, Type type, PlayMode mode,
                                      float offset)
     {
+        const ::ProfileScope profile("SoundManager::playSound3D");
         if(!mOutput->isInitialized())
             return nullptr;
 
@@ -718,6 +719,7 @@ namespace MWSound
 
     void SoundManager::stopSound3D(const MWWorld::ConstPtr &ptr, const std::string& soundId)
     {
+        const ::ProfileScope profile("SoundManager::stopSound3D");
         Sound_Buffer *sfx = loadSound(Misc::StringUtils::lowerCase(soundId));
         if (!sfx) return;
 
@@ -1119,6 +1121,8 @@ namespace MWSound
 
     void SoundManager::update(float duration)
     {
+        const ::ProfileScope profile("SoundManager::update");
+
         if(!mOutput->isInitialized())
             return;
 
--- a/apps/openmw/mwstate/statemanagerimp.cpp
+++ b/apps/openmw/mwstate/statemanagerimp.cpp
@@ -596,6 +596,8 @@ MWState::StateManager::CharacterIterator MWState::StateManager::characterEnd()
 
 void MWState::StateManager::update (float duration)
 {
+    const ::ProfileScope profile("StateManager::update");
+
     mTimePlayed += duration;
 
     // Note: It would be nicer to trigger this from InputManager, i.e. the very beginning of the frame update.
--- a/apps/openmw/mwworld/scene.cpp
+++ b/apps/openmw/mwworld/scene.cpp
@@ -317,6 +317,7 @@ namespace MWWorld
 
     void Scene::update (float duration, bool paused)
     {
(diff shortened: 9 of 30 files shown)


terabyte25:
Huh, new PR format I guess.

akortunov:
Can you take a look at at attached save (or you can just use the "coc balmora" in upstream master)?

[world.omwsave.zip](https://github.com/OpenMW/openmw/files/2887989/world.omwsave.zip)

In this save World update takes near 10ms per frame because of World::updateNavigator(). It should be a recent regression (probably in your previous PR) since AnyOldName's shadows branch does not suffer from this issue.

elsid:
@akortunov  I've tried to load [world.omwsave.zip](https://github.com/OpenMW/openmw/files/2887989/world.omwsave.zip) save using [this](https://github.com/elsid/openmw/commit/943a13fa196efd670ddab981d64ac56c72f62dc6) commit. It adds frame profilng to upstream master (with merged shadows). I have following top 20 `World::updateNavigator` calls:
```
depth  name                                     duration, ns        
2      World::updateNavigator                   9638                
2      World::updateNavigator                   5613                
2      World::updateNavigator                   5353                
2      World::updateNavigator                   4446                
2      World::updateNavigator                   4140                
2      World::updateNavigator                   3885                
2      World::updateNavigator                   3797                
2      World::updateNavigator                   3457                
2      World::updateNavigator                   3436                
2      World::updateNavigator                   3259                
2      World::updateNavigator                   3194                
2      World::updateNavigator                   3156                
2      World::updateNavigator                   2896                
2      World::updateNavigator                   2791                
2      World::updateNavigator                   2363                
2      World::updateNavigator                   2360                
2      World::updateNavigator                   2359                
2      World::updateNavigator                   2277                
2      World::updateNavigator                   2239                
2      World::updateNavigator                   2213                
```
Full profile is [here](https://gist.github.com/elsid/6d2605f4b51cddc51747a7a3713ca5b8). There is nothing like 10ms, only 10us.

It seems you are using some mods. Probably it is a reason. I have [this](https://gist.github.com/elsid/5ddb8de08d9c9f037151e6b5f6ed7e49) openmw output. And the character has no visual chest body part.

Please, try [this](https://github.com/elsid/openmw/commit/943a13fa196efd670ddab981d64ac56c72f62dc6) commit and publish profile output.

akortunov:
> Please, try [this](https://github.com/elsid/openmw/commit/943a13fa196efd670ddab981d64ac56c72f62dc6) commit and publish profile output.

[profile.1550738650.txt](https://github.com/OpenMW/openmw/files/2888386/profile.1550738650.txt)

psi29a:
@akortunov what is your Morrowind loadout? Just the expansions or with mods?

elsid:
I've tried to run openmw with Animated Containers mod. It seems I've broken update for animated objects by da6df818ff56e794ecec1d57a6b4a223aff24eb2 . Now tiles are updating even if objects has not been changed. f13b9e62e5d2c2b32d9c9b2d82277453c8ea035d allow to see how many times tile has been changed by heat color.

Before [fix](https://github.com/OpenMW/openmw/commit/da6df818ff56e794ecec1d57a6b4a223aff24eb2) with animated containers:
before_fix_with_animated_containers
Before fix with animated containers and opened container:
before_fix_with_animated_containers_and_opened_container
After fix without animated containers:
after_fix_without_animated_containers
After fix with animated containers:
after_fix_with_animated_containers

Blue color means few changes. Green color is a lot more. Basically not blue tiles are buggy if there is no changed object.

elsid:
@akortunov Problem is fixed by https://github.com/OpenMW/openmw/pull/2179

akortunov:
I found an another case when the slowdown appears - when a player opens/closes non-teleporting doors.

Capostrophic:
Confirmed, tile rebuilding during door rotation has a ridiculous framerate impact.

elsid:
Please, report with more details. Do some profiling at least. I found two other bugs ([4880](https://gitlab.com/OpenMW/openmw/issues/4880), [4881](https://gitlab.com/OpenMW/openmw/issues/4881)) trying to reproduce non-teleporting doors problem, but still no success.

akortunov:
Which kind of details do you need?
I disabled all mods and replacers, tried to use default settings and OpenMW still spends a lot of time to update, and it is not a recent regression (presents at least since beginning of year).

akortunov:
Also I still have no idea why navmesh build after cell change decreases FPS so badly. I have a suspicion that both navmesh and main threads are assigned to the same CPU core on my PC for some reason (probably because of hyper-threading).

`htops` shows that only one core is fully utilized (which handles a main thread and probably navmesh thread).

psi29a:
Threading is one of those things... that the OS assigns threads, so it is entirely possible that two threads are assigned per physical CPU which sucks.

elsid:
@akortunov 

I can see FPS drop only when navmesh rendering is enabled. Same thing on Arch Linux and Windows systems with different hardware.


> Which kind of details do you need?
1. Does it happens on every open/close of same door? Usually after couple of iterations all possible state are cached, so there is no real navmesh update then.
2. What exactly is fps drop? Could you provide in-game profile graphic? Better to have it in dynamic (video or series of screenshots).
3. What system do you use to play?
4. I managed [this](https://gist.github.com/elsid/c89e1358bf9d009e7831db80b398e672) small program to determine approximate process quantum. It starts two threads assigned to same CPU and prints timestamp in infinite loop with own prefix for each thread. On my Arch Linux it is [about](https://gist.github.com/elsid/6054c410b7f1ead06ee61c63f51e61ea) 4-6ms. Try to run it. What number do you have?

> I have a suspicion that both navmesh and main threads are assigned to the same CPU core on my PC for some reason (probably because of hyper-threading).

It's absolutely fine to have multiple threads to run on one CPU core. Many years single-core cpu managed to run multiple processes simultaniously and still are able to. It is more about OS rather than hardware.

> Can the issue be because of using two different threading implementaions in the same application (OpenThreads and std::thread)?

They are just wrappers over pthreads on linux. This shouldn't be an issue.

akortunov:
Strange, but after software update I can not reproduce freezing issues. What could affect it:
1. GCC 7.3.0 -> 8.2.0 update
2. Patched OSG 3.6.3 -> "OSG-on-steroids" 3.6.0 change
3. nvidia-drivers 415.27 -> 418.43 update

I still can reproduce doors issue with my system-wide build, which was built with GCC 7.
@Capostrophic, you had some similar issues before. Which setup do you use?

Other issues I found:
1. There is a 
> Build path by navigator exception: "Navmesh polygon for start point is not found"

spam in the log file, happens for both swimming and flying actors. Can we suppress such warnings?

2. Large creatures (such as alits, kagouti or guars) move twitchy as if the navigator rebuilds path every frame (similar to [bug #4814](https://gitlab.com/OpenMW/openmw/issues/4814)). Does not happen when the navigator is disabled.

elsid:
> spam in the log file, happens for both swimming and flying actors. Can we suppress such warnings?

Yes. But probably it is not required to use navmesh to build paths for those actors at all. So we can avoid these warnings.

>2. Large creatures (such as alits, kagouti or guars) move twitchy as if the navigator rebuilds path every frame (similar to bug #4814). Does not happen when the navigator is disabled.

Didn't look at that bug yet. Probably somehow related to #2204.

akortunov:
> Patched OSG 3.6.3 -> "OSG-on-steroids" 3.6.0 change

This. I switched back to the 3.6.3 and issue with doors opening re-appeared.

elsid:
I found this. OpenSceneGraph [uses]() cpu affinity for some reason. OpenMW fork partially [disables](https://github.com/OpenMW/osg/commits/3.6/src/osgViewer/ViewerBase.cpp) it for version 3.6. So I think it is the reason. I use OpenMW osg fork but version 3.4.

psi29a:
@elsid you mean that default OSG sets CPU affinity and the OS can schedule other threads on the CPU already in use?

What are our options then? Do we then set CPU affinity for our threads too to make sure there is no collision or do we turn that off for OSG and let the OS do its job?

fr3dz10:
> I found this. OpenSceneGraph uses cpu affinity for some reason. OpenMW fork partially [disables](https://github.com/OpenMW/osg/commits/3.6/src/osgViewer/ViewerBase.cpp) it for version 3.6. So I think it is the reason. I use OpenMW osg fork but version 3.4.

They [solved this bug](https://github.com/openscenegraph/OpenSceneGraph/blob/7ebe56291fb032820522a193c43d49e658a0a65f/src/OpenThreads/pthreads/PThread.cpp#L134) later in the 3.6 branch (or so it seems, I can't test myself).

psi29a:
@fr3dz10 that's very interesting, it would explain why people see good performance with 3.6 without all the scrawl fixes.

Capostrophic:
>it would explain why people see good performance with 3.6 without all the scrawl fixes.

It's not something that would magically improve the performance... and the newly added system that toggles this behavior would need to actually be used from OpenMW's side. Quoting AnyOldName3:

>In theory, this was fine, as it was behind a check that OSG was in SingleThreaded mode, but that makes the assumption that all threads are OSG threads, which doesn't hold for OpenMW. Upstream now has a system to enable or disable this at runtime, so it doesn't need our fork's version.

psi29a:
But we make use of OSG's multithreaded renderer... right?

fr3dz10:
> @fr3dz10 that's very interesting, it would explain why people see good performance with 3.6 without all the scrawl fixes.

As I understand, it just fixes a regression that occured at the beginning of the 3.6 branch when trying to outsmart the OS scheduler. Now, by default there is no affinity set for any thread.

The improvement were rather in other, unrelated area.

elsid:
So what should we do with [this](https://gitlab.com/OpenMW/openmw/issues/4785) issue? This pr at current state don't add significantly better performance. I would prefer to close it. And the real problem is OSG version.

akortunov:
Just for reference - during savegame loading I get a lot of "Set existing cache value" exceptions in the NavMeshTilesCache::set(). It is a quite annoying during debugging. Is it an intended behaviour?

elsid:
> Just for reference - during savegame loading I get a lot of "Set existing cache value" exceptions in the NavMeshTilesCache::set(). It is a quite annoying during debugging. Is it an intended behaviour?

Basically it's ok but not good. It happens when there are more than one navmesh updater thread and one tile is changed twice within quite short time interval. First thread take job to process tile, then new job is pushed, then second thread start to process second job for same tile. Both can't get cached tile and do same work. At the end only one can set value for this tile. Better to avoid exceptions and work duplication here. Thanks for report.

akortunov:
> They solved this bug later in the 3.6 branch (or so it seems, I can't test myself).

From what I can tell, this patch is present in the 3.6.3 release (actually, the commit was added 3 years ago), which is affected by the affinity bug.
So I suppose this bug will be present in the 3.6.4 as well.

akortunov:
So the summary here:

> So what should we do with this issue? This pr at current state don't add significantly better performance. I would prefer to close it

1. Feel free to close this PR if you prefer to do it.
2. Do not touch this issue now. I'll check the bug is resolved in the 3.6.4 release (more likely it is not). If the 3.6.4 does not suffer from it, I will close the issue.

PR #2175, opened 2019-02-19
Include cleanup

Capostrophic:
Cut down on includes in header files where possible (in MWWorld and MWMechanics), use forward declarations and move includes into implementation files where necessary.


akortunov:
LGTM.

PR #2176, opened 2019-02-20
Add support for NiRollController

akortunov:
Implements [feature #4675](https://gitlab.com/OpenMW/openmw/issues/4675), so the Daedric Crescent mesh from Weapon Sheathing mod should work properly.

Note: from what I can tell, a controller's current value is the per-update rotation, not a per-second speed in Morrowind, so rotation will be faster with high FPS, but in a very tricky way. I see no reason to replicate such behaviour if we really do not need it.

It would be nice to have some meshes for testing.

### DIFF
--- a/components/nifosg/controller.cpp
+++ b/components/nifosg/controller.cpp
@@ -309,6 +309,46 @@ void VisController::operator() (osg::Node* node, osg::NodeVisitor* nv)
     traverse(node, nv);
 }
 
+RollController::RollController(const Nif::NiFloatData *data)
+    : mData(data->mKeyList, 1.f)
+{
+}
+
+RollController::RollController()
+{
+}
+
+RollController::RollController(const RollController &copy, const osg::CopyOp &copyop)
+    : osg::NodeCallback(copy, copyop)
+    , Controller(copy)
+    , mData(copy.mData)
+    , mStartingTime(0)
+{
+}
+
+void RollController::operator() (osg::Node* node, osg::NodeVisitor* nv)
+{
+    traverse(node, nv);
+
+    if (hasInput())
+    {
+        double newTime = nv->getFrameStamp()->getSimulationTime();
+        double duration = newTime - mStartingTime;
+        mStartingTime = newTime;
+
+        float value = mData.interpKey(getInputValue(nv));
+        osg::MatrixTransform* transform = static_cast<osg::MatrixTransform*>(node);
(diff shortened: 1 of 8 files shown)


components/nifosg/controller.cpp
```diff
@@ -309,6 +309,50 @@ void VisController::operator() (osg::Node* node, osg::NodeVisitor* nv)
...
+        float value = mData.interpKey(getInputValue(nv));
+        osg::MatrixTransform* transform = static_cast<osg::MatrixTransform*>(node);
+        osg::Matrix matrix = transform->getMatrix();
+
+        // Rotate around "roll" axis.
+        // Note: we set per-update speed, not per-second, so an actual rotation speed will depend on FPS
+        // Original engine behaves in the similar way, but actual rotation speed can be different from OpenMW.
+        osg::Quat rotate (value, osg::Vec3f(0,0,1));
+        osg::Quat orient = rotate * matrix.getRotate();
+        matrix.setRotate(orient);
```
Capostrophic: To close 4856: bzzt believes the external rotation of the node must be accounted for in orient (which makes sense, though I'm not sure what else could rotate the thing).

terabyte25: ... what?

akortunov:
A small status update:
1. NiRollController is really supposed to rotate node about local Z ("roll", R in the NifSkope) axis.
2. I made the controller framerate-independent for now. If later there will be complaints about it, we will change its behaviour.

kcat:
I would go as far as to say, even if there are complaints about it we should leave it framerate-independent. Any mod that tries to use framerate to control game behavior is doing a Bad Thing and will be unreliable, especially given the wide gulf there is between systems with low FPS (20~30 or lower) and high (120~140 or higher). To say nothing of G-Sync/FreeSync, which gives you variable framerates. It's a recipe for bugs to be framerate-dependent since it adds unpredictability, inhibiting proper testing and debugging.

akortunov:
Tested with this mesh. My new approach does not discard scale node scale.
[w_art_blade_crescent.zip](https://github.com/OpenMW/openmw/files/2888331/w_art_blade_crescent.zip)

akortunov:
Can anyone check if it works on Windows?

akortunov:
I suppose it means "no". Then I suggest an another approach: merge and see if anyone will complain about it.

Capostrophic:
It works just fine in the AppVeyor artifact under Wine.

PR #2177, opened 2019-02-21
 Handle initial actor's transparency

akortunov:
Fixes [bug #4860](https://gitlab.com/OpenMW/openmw/issues/4860).

The main idea:
1. Move visibility update to the separate function
2. Call it after adding actor to scene as well during actor's update.

Note: when we teleport player, we can not check distance to actor since player still has old coordinates.
I do not know how to handle this case - for now I just ignore the visibility check.


psi29a:
I'll take it, thank you.

PR #2178, opened 2019-02-21
Ignore extra text after a local variable declaration (bug #4867)

Capostrophic:
[Bug 4867](https://gitlab.com/OpenMW/openmw/issues/4867).

Special characters, keywords and names are ignored when End state is present and the rest of the line is skipped.


Capostrophic:
Pushed an alternative approach that logs warnings regarding said extra text.

PR #2179, opened 2019-02-22
Fix navmesh update for not changed objects

elsid:
When update method is called for not changed object befor this change all object tiles were considered as not object tiles and were removed. Also this marked those tiles as changed. This lead to alternation between remove and add each tile update method was called. Problem was detected by using Animated Containers mod.


akortunov:
Works for me.

PR #2182, opened 2019-02-22
Replace Boost format and replace_all where possible

Capostrophic:
Every time boost/format.hpp is included in a unique file, virtually [the entirety of the freaking Boost.Format module](https://github.com/boostorg/format/blob/develop/include/boost/format.hpp) with all the templates and complex functions gets put into the source to simply copy the name and count of something into a messagebox, which certainly doesn't improve the compilation times and it doesn't even have that good of runtime performance to begin with.

So I replaced boost::format with less complex "replace" alternative which does the same thing in the context to get rid of these dependencies in the engine itself. I used it in all or most instances of %s/%d tag replacement.

esmtool still uses boost::format but it is left alone for the time being.

replace_all from Boost was replaced with our own replaceAll which does the same thing, which also allows to remove some boost dependencies in book formatting.

### DIFF
--- a/apps/openmw/mwgui/recharge.cpp
+++ b/apps/openmw/mwgui/recharge.cpp
@@ -1,7 +1,5 @@
 #include "recharge.hpp"
 
-#include <boost/format.hpp>
-
 #include <MyGUI_ScrollView.h>
 #include <MyGUI_Gui.h>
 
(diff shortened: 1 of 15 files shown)


apps/openmw/mwgui/recharge.cpp
```diff
@@ -179,7 +177,9 @@ void Recharge::onItemClicked(MyGUI::Widget *sender, const MWWorld::Ptr& item)
     if (gem.getRefData().getCount() == 0)
     {
         std::string message = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("sNotifyMessage51")->mValue.getString();
-        message = boost::str(boost::format(message) % gem.getClass().getName(gem));
+        if (message.find("%s") != std::string::npos)
+            message.replace(message.find("%s"), 2, gem.getClass().getName(gem));
```
elsid: Why don't put this logic into separate function? Each `boost::format` call now is replaced by `if`. This increases algorithmic complexity of each function. And because there is no function you have to call find two times to avoid new variable definition. If eventually c++ will get `std::format` for string it would be much easier to just change our function implementation than to find all of this if-find-replace-find patterns.

Capostrophic: Hm, okay. I'll put something into StringUtils.

Capostrophic:
I did... something.

psi29a:
You guys are doing some serious de-boosting, however... be-careful with replace. I ran into issues a year or two ago when doing this.

Capostrophic:
>I ran into issues a year or two ago when doing this.

I did some brief testing with all the instances of formatting but didn't notice anything weird. What kind of issues?

PR #2183, opened 2019-02-23
[Feedback needed] Projectiles collisions

akortunov:
Should fix [bug #3372](https://gitlab.com/OpenMW/openmw/issues/3372).

A current design may be suboptimal, so it would be nice to get a feedback from other developers.
@AnyOldName3, @wareya, @elsid, what do you think?

The main idea:
1. Add collision object to the collision world for every projectile in scene.
2. Use an integer ID since projectiles have no persistent pointers.
3. During sweep tests for actors check if they collide with projectiles. If yes, apply projectile hit immediately and skip collision.

### DIFF
--- a/apps/openmw/mwphysics/closestnotmeconvexresultcallback.cpp
+++ b/apps/openmw/mwphysics/closestnotmeconvexresultcallback.cpp
@@ -2,6 +2,14 @@
 
 #include <BulletCollision/CollisionDispatch/btCollisionObject.h>
 
+#include <components/misc/convert.hpp>
+
+#include "../mwbase/world.hpp"
+#include "../mwbase/environment.hpp"
+
+#include "collisiontype.hpp"
+#include "projectile.hpp"
+
 namespace MWPhysics
 {
     ClosestNotMeConvexResultCallback::ClosestNotMeConvexResultCallback(const btCollisionObject *me, const btVector3 &motion, btScalar minCollisionDot)
@@ -10,10 +18,22 @@ namespace MWPhysics
     {
     }
 
-    btScalar ClosestNotMeConvexResultCallback::addSingleResult(btCollisionWorld::LocalConvexResult& convexResult,bool normalInWorldSpace)
+    btScalar ClosestNotMeConvexResultCallback::addSingleResult(btCollisionWorld::LocalConvexResult& convexResult, bool normalInWorldSpace)
     {
         if (convexResult.m_hitCollisionObject == mMe)
             return btScalar(1);
+
+        if (convexResult.m_hitCollisionObject->getBroadphaseHandle()->m_collisionFilterGroup == CollisionType_Projectile)
+        {
+            Projectile* projectileHolder = static_cast<Projectile*>(convexResult.m_hitCollisionObject->getUserPointer());
+            int projectileId = projectileHolder->getProjectileId();
+            PtrHolder* targetHolder = static_cast<PtrHolder*>(mMe->getUserPointer());
+            const MWWorld::Ptr target = targetHolder->getPtr();
+
+            osg::Vec3f pos = Misc::Convert::makeOsgVec3f(convexResult.m_hitPointLocal);
+            MWBase::Environment::get().getWorld()->manualProjectileHit(projectileId, target, pos);
+            return btScalar(1);
+        }
 
         btVector3 hitNormalWorld;
         if (normalInWorldSpace)
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -41,6 +41,8 @@
 
 #include "collisiontype.hpp"
 #include "actor.hpp"
+
+#include "projectile.hpp"
 #include "trace.h"
 #include "object.hpp"
 #include "heightfield.hpp"
@@ -58,6 +60,7 @@ namespace MWPhysics
         , mResourceSystem(resourceSystem)
         , mDebugDrawEnabled(false)
         , mTimeAccum(0.0f)
+        , mProjectileId(0)
         , mWaterHeight(0)
         , mWaterEnabled(false)
         , mParentNode(parentNode)
@@ -108,6 +111,11 @@ namespace MWPhysics
         }
 
         for (ActorMap::iterator it = mActors.begin(); it != mActors.end(); ++it)
+        {
+            delete it->second;
+        }
+
+        for (ProjectileMap::iterator it = mProjectiles.begin(); it != mProjectiles.end(); ++it)
         {
             delete it->second;
         }
@@ -262,7 +270,7 @@ namespace MWPhysics
             return (point - Misc::Convert::toOsg(cb.m_hitPointWorld)).length();
     }
 
-    PhysicsSystem::RayResult PhysicsSystem::castRay(const osg::Vec3f &from, const osg::Vec3f &to, const MWWorld::ConstPtr& ignore, std::vector<MWWorld::Ptr> targets, int mask, int group) const
+    PhysicsSystem::RayResult PhysicsSystem::castRay(const osg::Vec3f &from, const osg::Vec3f &to, const MWWorld::ConstPtr& ignore, std::vector<MWWorld::Ptr> targets, int mask, int group, int projId) const
     {
         btVector3 btFrom = Misc::Convert::toBullet(from);
         btVector3 btTo = Misc::Convert::toBullet(to);
@@ -293,7 +301,7 @@ namespace MWPhysics
             }
         }
 
-        ClosestNotMeRayResultCallback resultCallback(me, targetCollisionObjects, btFrom, btTo);
+        ClosestNotMeRayResultCallback resultCallback(me, targetCollisionObjects, btFrom, btTo, projId);
         resultCallback.m_collisionFilterGroup = group;
         resultCallback.m_collisionFilterMask = mask;
 
@@ -306,7 +314,17 @@ namespace MWPhysics
             result.mHitPos = Misc::Convert::toOsg(resultCallback.m_hitPointWorld);
             result.mHitNormal = Misc::Convert::toOsg(resultCallback.m_hitNormalWorld);
             if (PtrHolder* ptrHolder = static_cast<PtrHolder*>(resultCallback.m_collisionObject->getUserPointer()))
+            {
                 result.mHitObject = ptrHolder->getPtr();
+
+                if (group == CollisionType_Projectile)
+                {
+                    Projectile* projectile = static_cast<Projectile*>(ptrHolder);
+                    result.mProjectileId = projectile->getProjectileId();
+                }
+                else
+                    result.mProjectileId = -1;
+            }
         }
         return result;
     }
@@ -500,6 +518,19 @@ namespace MWPhysics
         }
     }
 
+    void PhysicsSystem::removeProjectile(const int projectileId)
+    {
+        ProjectileMap::iterator foundProjectile = mProjectiles.find(projectileId);
+        if (foundProjectile != mProjectiles.end())
+        {
+            delete foundProjectile->second;
+            mProjectiles.erase(foundProjectile);
+
+            if (projectileId == mProjectileId)
+                mProjectileId--;
+        }
+    }
+
     void PhysicsSystem::updateCollisionMapPtr(CollisionMap& map, const MWWorld::Ptr &old, const MWWorld::Ptr &updated)
     {
         CollisionMap::iterator found = map.find(old);
@@ -582,6 +613,17 @@ namespace MWPhysics
         }
     }
 
+    void PhysicsSystem::updateProjectile(const int projectileId, const osg::Vec3f &position)
+    {
+        ProjectileMap::iterator foundProjectile = mProjectiles.find(projectileId);
+        if (foundProjectile != mProjectiles.end())
+        {
+            foundProjectile->second->setPosition(position);
+            mCollisionWorld->updateSingleAabb(foundProjectile->second->getCollisionObject());
+            return;
+        }
+    }
+
     void PhysicsSystem::updateRotation(const MWWorld::Ptr &ptr)
     {
         ObjectMap::iterator found = mObjects.find(ptr);
@@ -638,6 +680,15 @@ namespace MWPhysics
 
         Actor* actor = new Actor(ptr, shape, mCollisionWorld);
         mActors.insert(std::make_pair(ptr, actor));
+    }
+
+    int PhysicsSystem::addProjectile (const osg::Vec3f& position)
+    {
+        mProjectileId++;
+        Projectile* projectile = new Projectile(mProjectileId, position, mCollisionWorld);
+        mProjectiles.insert(std::make_pair(mProjectileId, projectile));
+
+        return mProjectileId;
     }
 
     bool PhysicsSystem::toggleCollisionMode()
--- a/apps/openmw/mwphysics/projectile.cpp
+++ b/apps/openmw/mwphysics/projectile.cpp
@@ -1,0 +1,64 @@
+#include "projectile.hpp"
+
+#include <BulletCollision/CollisionShapes/btSphereShape.h>
+#include <BulletCollision/CollisionDispatch/btCollisionWorld.h>
+
+#include <components/sceneutil/positionattitudetransform.hpp>
+#include <components/resource/bulletshape.hpp>
+#include <components/debug/debuglog.hpp>
+#include <components/misc/convert.hpp>
+
+#include "../mwworld/class.hpp"
+
+#include "collisiontype.hpp"
+
+namespace MWPhysics
+{
+Projectile::Projectile(int projectileId, const osg::Vec3f& position, btCollisionWorld* world)
+    : mCollisionWorld(world)
+{
+    mProjectileId = projectileId;
+
+    mShape.reset(new btSphereShape(1.f));
+    mConvexShape = static_cast<btConvexShape*>(mShape.get());
+
+    mCollisionObject.reset(new btCollisionObject);
+    mCollisionObject->setCollisionFlags(btCollisionObject::CF_KINEMATIC_OBJECT);
+    mCollisionObject->setActivationState(DISABLE_DEACTIVATION);
+    mCollisionObject->setCollisionShape(mShape.get());
+    mCollisionObject->setUserPointer(this);
+
+    setPosition(position);
+
+    const int collisionMask = CollisionType_World | CollisionType_HeightMap |
+        CollisionType_Actor | CollisionType_Door | CollisionType_Water;
+    mCollisionWorld->addCollisionObject(mCollisionObject.get(), CollisionType_Projectile, collisionMask);
+}
+
+Projectile::~Projectile()
+{
+    if (mCollisionObject.get())
+        mCollisionWorld->removeCollisionObject(mCollisionObject.get());
+}
+
+void Projectile::updateCollisionObjectPosition()
+{
+    btTransform tr = mCollisionObject->getWorldTransform();
+   // osg::Vec3f scaledTranslation = mRotation * mMeshTranslation;
+   // osg::Vec3f newPosition = scaledTranslation + mPosition;
+    tr.setOrigin(Misc::Convert::toBullet(mPosition));
+    mCollisionObject->setWorldTransform(tr);
+}
+
+void Projectile::setPosition(const osg::Vec3f &position)
+{
+    mPosition = position;
+    updateCollisionObjectPosition();
+}
+
+osg::Vec3f Projectile::getPosition() const
+{
+    return mPosition;
+}
+
+}
--- a/apps/openmw/mwworld/projectilemanager.cpp
+++ b/apps/openmw/mwworld/projectilemanager.cpp
@@ -278,6 +278,7 @@ namespace MWWorld
         MagicBoltState state;
         state.mSpellId = spellId;
         state.mCasterHandle = caster;
+        state.mHit = false;
         if (caster.getClass().isActor())
             state.mActorId = caster.getClass().getCreatureStats(caster).getActorId();
         else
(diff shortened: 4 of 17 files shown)


AnyOldName3:
Any specific parts of this you want me to review? As far as I can see from a skim-read, it doesn't touch anything I know about.

akortunov:
> Any specific parts of this you want me to review?

I do not like things such as `markToRemove`, where I forced to iterate over two vectors and search by Id. Are there more effective ways to do it without rewriting the whole ProjectileManager?

The second question - it is not very clear which collision boxes projectiles should have. In my current implementation I try to calculate them automatically and if the calculation fail, use the 10x10x10-units box.

wareya:
>     4\. During sweep tests for actors check if they collide with projectiles. If yes, apply projectile hit immediately and skip collision.

This is probably safe, but it might cause code execution order problems. Doing a single piece of game logic (e.g. handling projectiles hitting actors) in multiple places makes me very suspicious.

This is going to sound crazy, but would it make sense to add a flag to the projectile saying that an actor ran into it, so the projectile can react to the actor during its normal update function? Basically delaying the "hit" until the engine is looping over projectiles again.

wareya:
> As I understand current logic each projectile can hit only one actor. What if there are multiple actors collided with a projectile? Who will be hit first? Or should projectile hit all of them?

The right way to do this is to use some kind of continuous contact solver, but OpenMW doesn't have one, and it would be an unreasonable complexity addition.

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -1201,6 +1236,19 @@ namespace MWPhysics
...
 
+    int PhysicsSystem::addProjectile (const std::string& mesh, const osg::Vec3f& position, const osg::Quat &orient)
+    {
+        osg::ref_ptr<const Resource::BulletShape> shape = mShapeManager->getShape(mesh);
+        if (!shape)
+            return -1;
+
+        mProjectileId++;
+        Projectile* projectile = new Projectile(mProjectileId, shape, position, orient, mCollisionWorld);
+        mProjectiles.insert(std::make_pair(mProjectileId, projectile));
```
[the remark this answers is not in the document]
akortunov: It is a copy-paste from addActor().

akortunov:
> As I understand current logic each projectile can hit only one actor. What if there are multiple actors collided with a projectile? Who will be hit first? Or should projectile hit all of them?

A first hit will be the first hit, obviously.

> This is going to sound crazy, but would it make sense to add a flag to the projectile saying that an actor (and specifically, a specific actor) ran into it, so the projectile can react to the actor during its normal update function? Basically delaying the "hit" until the engine is looping over projectiles again.

An interesting idea, but we will need to execute the code from the ProjectileManager inside collision detection anyway (at least, set target actorId and mark projectile as removed).

elsid:
> A first hit will be the first hit, obviously.

Ok, I didn't know that we always have a first actor in a game. Seriously is it first by distance to a projectile, or by place in some container, or by random? Collision check is placed in `ActorTracer::doTrace` and `ActorTracer::findGround` methods which are called multiple times from multiple places.

akortunov:
> Seriously is it first by distance to a projectile, or by place in some container, or by random?

First by update. Pretty the same thing as in our current implementation.

apps/openmw/mwphysics/projectile.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    mShape.reset(new btBoxShape(toBullet(mHalfExtents)));
+    mConvexShape = static_cast<btConvexShape*>(mShape.get());
+
+    mCollisionObject.reset(new btCollisionObject);
+    mCollisionObject->setCollisionFlags(btCollisionObject::CF_KINEMATIC_OBJECT);
+    mCollisionObject->setActivationState(DISABLE_DEACTIVATION);
+    mCollisionObject->setCollisionShape(mShape.get());
+    mCollisionObject->setUserPointer(static_cast<PtrHolder*>(this));
```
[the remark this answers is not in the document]
akortunov: I took a code from actor.cpp as example. In other places we use the same cast too.

apps/openmw/mwphysics/physicssystem.cpp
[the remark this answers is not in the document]
akortunov: Since all other places use the same approach, I suppose such possible leaks are outside of scope of this PR.

akortunov:
Issues left:
1. "Projectile-projectile" part does not work well for some reason.
2. ClosestNotMeConvexResultCallback applies hit, if a moving actor collides with projectile.
It is not clean code, but I see no ways to implement it without making code a way more complex.
3. Currently I apply hits to actors in two cases:
a) When I move projectiles in the ProjectileManager  - to the closest ray tracing result.
b) When I move actors in the PhysicsSystem via convexSweepTest - to the first actor, which collides with given projectile.
In the second case I do not know how to checks if the actor is really closest. Maybe it does not matter, though.

I'd consider 2 and 3 as non-blockers.

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -501,27 +529,153 @@ namespace MWWorld
...
+
+                MWWorld::Ptr caster = it->getCaster();
+                if (caster == target)
+                    return;
+
+                if (!isValidTarget(caster, target))
+                    return;
+
+                if (caster.isEmpty())
+                    caster = target;
```
elsid: Not sure I understand why the case of empty caster is not a logic error? There is another check for empty below that seems redundant now.

akortunov: I can not really tell - too many time passed, but I am pretty sure that was check was added to do not break Morrowind's game mechanics.

elsid: As I understand this is a copy of logic [here](https://github.com/OpenMW/openmw/pull/2183/files#diff-cc9df71b3061f0d452a113f4708feb57R528-R529) that was originally introduced by [this](https://github.com/OpenMW/openmw/commit/12de0afb03a382332f2da3d31b073fe9d49bb7ff#diff-d3b808d6f40e4c12e247457971def846R2294-R2295) commit. Looks like some kind of a workaround for an undescribed problem. And of course there is no trivial way to find a pr for the original commit to see the discussion.

elsid: Found the [pr](https://github.com/zinnschlag/openmw/pull/1279/files#diff-d3b808d6f40e4c12e247457971def846R2299-R2300). Had to look manually all pr around commit date. Sadly there is no any discussion.

apps/openmw/mwphysics/projectile.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    mCollisionObject.reset(new btCollisionObject);
+    mCollisionObject->setCollisionFlags(btCollisionObject::CF_KINEMATIC_OBJECT);
+    mCollisionObject->setActivationState(DISABLE_DEACTIVATION);
+    mCollisionObject->setCollisionShape(mShape.get());
+    mCollisionObject->setUserPointer(this);
+
+    setPosition(position);
+
+    addCollisionMask(getCollisionMask());
```
Capostrophic, line 38: The collision mask is constant (it can even be constexpr), so addCollisionMask and getCollisionMask can be removed in favor of only adding the collision object in the constructor. I don't suppose anything needs to work with a projectile's collision mask.

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -301,7 +302,9 @@ namespace MWWorld
         MWWorld::Ptr ptr = ref.getPtr();
 
         osg::Vec4 lightDiffuseColor = getMagicBoltLightDiffuseColor(state.mEffects);
-        createModel(state, ptr.getClass().getModel(ptr), pos, orient, true, true, lightDiffuseColor, texture);
+
+        const std::string model = ptr.getClass().getModel(ptr);
```
Capostrophic, line 306: This change and the identical added line below look redundant, and so do the formatting changes.

apps/openmw/mwphysics/closestnotmeconvexresultcallback.cpp
```diff
@@ -10,11 +16,23 @@ namespace MWPhysics
...
             return btScalar(1);
 
+        if (convexResult.m_hitCollisionObject->getBroadphaseHandle()->m_collisionFilterGroup == CollisionType_Projectile)
+        {
+            Projectile* projectileHolder = static_cast<Projectile*>(convexResult.m_hitCollisionObject->getUserPointer());
+            int projectileId = projectileHolder->getProjectileId();
+            PtrHolder* targetHolder = static_cast<PtrHolder*>(mMe->getUserPointer());
+            const MWWorld::Ptr target = targetHolder->getPtr();
+
+            osg::Vec3f pos(convexResult.m_hitPointLocal.x(), convexResult.m_hitPointLocal.y(), convexResult.m_hitPointLocal.z());
```
Capostrophic, line 31: Use Misc::Convert::makeOsgVec3f()

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -444,11 +461,14 @@ namespace MWWorld
...
 
+                if (result.mProjectileId >= 0)
+                    markToRemove(result.mProjectileId);
+
                 MWBase::SoundManager *sndMgr = MWBase::Environment::get().getSoundManager();
                 for (size_t soundIter = 0; soundIter != it->mSounds.size(); soundIter++)
                     sndMgr->stopSound(it->mSounds.at(soundIter));
 
-                mParent->removeChild(it->mNode);
+                cleanupMagicBolt(*it);
```
Capostrophic, line 471: cleanupMagicBolt() already handles stopping sounds, so the code above should be removed.

akortunov:
I had to remove "projectile-projectile" collision part since it did not work properly.
I suppose it is a time to review and test this PR again - it is a quite old.

akortunov:
The main mess so har is the "manual hit" logic - when we do a ConvexSweepTest for actor, actor can encounter a projectile shape. In this case we should somehow remove that projectile from the physics system and tell ProjectileManager that it will need to apply hit to given actor and remove projectile record.

Any ideas how to do it properly, especially if physics is supposed to work in another thread(s) in future?

fr3dz10:
> The main mess so far is the "manual hit" logic - when we do a ConvexSweepTest for actor, actor can encounter a projectile shape now. In this case we should somehow remove that projectile from the physics system (or change its collision flags to make sure that it will not collide with other actors) and tell ProjectileManager that it will need to apply hit to given actor and remove projectile record from mProjectiles/mMagicBolts.
> 
> Any ideas how to do it properly, especially if physics is supposed to work in another thread(s) in future?

To prevent future collision after a hit, it should be as simple as:
`projectileCollisionObject->getBroadphaseHandle()->m_collisionFilterMask ^=CollisionType_Projectile;`
or even 
`projectileCollisionObject->getBroadphaseHandle()->m_collisionFilterMask = 0;`

Now that I think of it, since projectiles are handled separately, `CollisionType_Projectile` could be removed altogether from `Actor::getCollisionMask()`.

What can't be done safely concurrently is `mCollisionWorld->removeCollisionObject(projectileCollisionObject);`
So in the hypothesis that this would run in parallel with physics, it would need to be either 
1. serialized (probably easier)
2. deferred till "after physics" (more performant)
Projectiles might be infrequent enough so that option 1 is sufficient. But we're not there yet :)

akortunov:
> Now that I think of it, since projectiles are handled separately, CollisionType_Projectile could be removed altogether from Actor::getCollisionMask().

Not sure if it is a good idea - when we move projectiles, they still can collide with actors. In this PR we just make it possible for actors to collide with projectiles when we move actors.

fr3dz10:
> > Now that I think of it, since projectiles are handled separately, CollisionType_Projectile could be removed altogether from Actor::getCollisionMask().
> 
> Not sure if it is a good idea - when we move projectiles, they still can collide with actors. In this PR we just make it possible for actors to collide with projectiles when we move actors.
Correct, it wouldn't work.

akortunov:
So what should we do here? I can either apply hit and remove projectile immediately or try to register hit in the physics system and apply hit during world update. Both approaches will conflict with async physics and multithreaded phyics.

fr3dz10:
I'm not sure what you mean by conflict in this context. To make it safe (non-racy) it is enough to use the wrapper around `addCollisionObject`, `removeCollisionObject` and `updateSingleAabb` provided by the physics scheduler instead of manipulating directly the collision world. If it is desirable for the whole projectile handling to be done in the same thread as the movement solver, it can be done without difficulty (even though there is obviously more work than a mere `s/mCollisionWorld/mPhysicsTaskScheduler/`).

akortunov:
> If it is desirable for the whole projectile handling to be done in the same thread as the movement solver

1. "Projectile handling" should be done in the main thread, as other game mechanics, while movement solver and its callbacks seem to work in the separate thread with async physics
2. If I understood correctly, async physics runs in one frame behind main simulation, while animations and game mechanics (which spawn new projectiles) work in the current frame, so there should be a some kind of double-buffering to do not send projectiles to the past.
3. If async physics is aimed to be multithreaded later, there can be a situation when several actors collide with the same projectile in the same time. It is unclear how to handle this case.
4. There is a case when we spawn projectile inside target's collision box. We track such cases by doing a raycasting when we spawn a projectile and applying the hit immediately. I do not know how this case is supposed to work if the target is still in the past (in the previous frame).

The separate issue is that the async physics can break scripts which move objects - script moves object from position X1 to position X2 and expects it to be in the X2, while the object is in the X0, because physics system works in another frame.

fr3dz10:
> 
> 
> > If it is desirable for the whole projectile handling to be done in the same thread as the movement solver
> 
>     1. "Projectile handling" should be done in the main thread, as other game mechanics, while movement solver and its callbacks seem to work in the separate thread with async physics

Not exactly. The movement solver run in a separate thread, other functionalities (hit detection for combat for example) run in the main thread. They need to query the collision world for ad-hoc collision detection so the scheduler act as a proxy that handle the synchronization details. 

>     2. If I understood correctly, async physics runs in one frame behind main simulation, while animations and game mechanics (which spawn new projectiles) work in the current frame, so there should be a some kind of double-buffering to do not send projectiles to the past.

There is only one copy of the collision world, each actor / object have only one position.

>     3. If async physics is aimed to be multithreaded, there can be a situation when several actors collide with the same projectile in the same time. It is unclear how to handle this case.

The same as now: whichever is first in the `mActors` container will be hit first. 

>     4. There is a case when we spawn projectile inside target's collision box. We track such cases by doing a raycasting when we spawn a projectile and applying the hit immediately. I do not know how this case is supposed to work if the target is still in the past (in the previous frame).

Same as now, for the same reason as 2.

> 
> The separate issue is that the async physics can break scripts which move objects - script moves object from position X1 to position X2 and expects it to be in the X2, while the object is in the X0, because physics system works in another frame.

No. If a script move an object from X1 to X2, it will be in X2. X2 then will be the new origin from which the movement solver will work, exactly as it is in synchronous case. 
What changes is that the end position of the actor might be closer to X2 as whatever the movement solver computed before the script ran will be discarded.

akortunov:
> There is only one copy of the collision world, each actor / object have only one position.

It is a very good source of heisenbugs then. 
If I move object via script in the N-th frame, a collisions detection for the N-1th frame should not be affected by this change.
Otherwise collision detection in the N-1th frame will depend on if CPU managed to finish collision detection for nearby actors 
 in N-1th frame or if it is not.
It may cause issues, which are not reproducable on another hardware.
There are plenty related complaints on Discord, by the way:
Screenshot_20200729_130028

For the same reason we have to use double-buffering (two copies of objects) in the rendering system, which runs in one frame behind of main thread, so changes from main thread should not affect that frame.

> The same as now

Not really. Now both game mechanics and physics use the same simulation time.
Since we run physics update after scripts and game mechanics update, we know in the physics update that the world is always in consistent state.
So I can remove projectile and apply hit on the fly since I know that no one else works with the physics scene in the current frame.
Whatever will happen, behaviour will be consistent on different setups. For the same reason we use a constant dt for physics system - to keep in consistent on different setups.
With async physics such assumption fails - part of game world lives in the N-th frame, part lives in the N-1th frame and behaviour depends on hardware and how many async threads are enabled.
And when arrow launched in the 20.03s simulation time hits an actor which is in the 20.0s simulation time (because used CPU did not finish processing for previous frame yet) at least strange.

fr3dz10:
> 
> 
> > There is only one copy of the collision world, each actor / object have only one position.
> 
> It is a very good source of heisenbugs then.
> If I move object via script in the N-th frame, a collisions detection for the N-1th frame should not be affected by this change.
> Otherwise collision detection in the N-1th frame will depend on if CPU managed to finish collision detection for nearby actors
> in N-1th frame or if it is not.

That's one way to look at it, another is: if I move object, it doesn't matter if it could have collide if I didn't move it.
OpenMW physics is not "real physics", there is no application of the "action-reaction law", no dynamics, no bouncing, no friction, etc. When a collision occurs, object simply stops (modulo the unstuck / slide logic that apply sometimes).


> It may cause issues, which are not reproducable on another hardware.
> There are plenty related complaints on Discord, by the way:
> Screenshot_20200729_130028

I'm aware of that, I don't know yet what causes this issue on Android.
Anyway, whatever is the problem, its solution won't change the principle.

> For the same reason we have to use double-buffering (two copies of objects) in the rendering system, which runs in one frame behind of main thread, so changes from main thread should not affect that frame.
> 
> > The same as now
> 
> Not really. Now both game mechanics and physics use the same simulation time.
> Since we run physics update after scripts and game mechanics update, we know in the physics update that the world is always in consistent state.
> So I can remove projectile and apply hit on the fly since I know that no one else works with the physics scene in the current frame.
Why can't you do the same if physics is async? 

> Whatever will happen, behaviour will be consistent on different setups. For the same reason we use a constant dt for physics system - to keep in consistent on different setups.
> With async physics such assumption fails - part of game world lives in the N-th frame, part lives in the N-1th frame and behaviour depends on hardware and how many async threads are enabled.
> And when arrow launched in the 20.03s simulation time hits an actor which is in the 20.0s simulation time (because used CPU did not finish processing for previous frame yet) at least strange.
Perhaps it will be a problem, perhaps no. At least I didn't experience issue with projectile due to collision shape being in a little bit different spot than what mechanics knows. The difference in position for actors is at most "dt / n * velocity", with velocity typically small and n between 10 and 20.

akortunov:
> Why can't you do the same if physics is async?

With current approach I know that other threads do not work with physics scene and know that result will be the same on the same scene on different setups, because there are no different timings which can affect it.
I can not make same assumptions with multithreaded and/or async physics and I can not guarantee that on-the-fly updates will work properly, even if they "seem to work" on my setup.
And judging by 
> Perhaps it will be a problem, perhaps no. 

you can not guarantee it as well.

In other parts of engine we have to use double buffering for objects which are used in different threads with different simulation time, otherwise heisenbugs on some setups arise, and it is hard to debug them.
An example of such approach is RigGeometry and statesets, which are updated in the main thread in the current frame, but their previous state is still used by Draw thread to render previous frame.

fr3dz10:
> And judging by
> 
> > Perhaps it will be a problem, perhaps no.
> 
> you can not guarantee it as well.

Of course I can't. But I prefer to start with the simplest implementation that seems to work, then add complexity if the need arise. It is more productive than introduce 2 copies of each collision objects (and maybe 2 collision worlds) to solve a problem that is still hypothetical.
I don't claim it will not be necessary to go "double buffering", just that for now there is no element to indicate that it is.

AnyOldName3:
We need double buffering with OSG stuff because we're producing the next frame while consuming it for the current frame. It's not an appropriate approach in the general case, though. It works for rendering because only one thread makes any changes and the other thread is just interested in the results. I'm under the impression that parts of physics need to provide feedback to game logic, so it's not an equivalent situation.

akortunov:
*sigh*

Here is a call stack from the ClosestNotMeConvexResultCallback, which I use:
Screenshot_20200730_100707

It is *definely* is not a main thread. Even if movement solver itself is now threadsafe and does not cause issues (actually, causes), the rest of engine is not threadsafe (including ProjectileManager), and movement solver does not exist in vacuum, it interacts with other parts of engine.
I need all callbacks to be in the main thread to work with ProjectileManager and other game mechanics.

Your stance:
1. Changes in order of job's subtasks execution do not affect job's result
2. Access to non-threadsafe objects from multithreading environment usually does not cause serious issues, so it is OK.

My stance:
1.  Changes in order of job's subtasks execution can affect job's result in unpredictable ways
2. Access to non-threadsafe objects from multithreading environment should be avoided at all since random sync issues can hurt project stability bad.

fr3dz10:
> 
> 
> _sigh_
> 
> Here is a call stack from the ClosestNotMeConvexResultCallback, which I use, with async physics:
> Screenshot_20200730_100707
> 
> It is _definely_ not a main thread. 

No it's not, but why do you think it is a problem? The callbacks are free function without internal state. Bullet takes care of its internal integrity if configured to do so at compile time, otherwise it is necessary to serialize calls to Bullet API. 

> Even if movement solver itself is now threadsafe and does not cause issues (actually, causes), the rest of engine is not threadsafe (including ProjectileManager), and movement solver does not exist in vacuum, it interacts with other parts of engine.
> I need all callbacks to be in the main thread to work with ProjectileManager and other game mechanics.


> Your stance:

Nope
 
>     1. Changes in order of job's subtasks execution do not affect job's result

It does change job result, but current behavior is not deterministic _at all_ : 
- collision is checked in the order actors are added to their container 
- internally Bullet uses a tree to represent collision objects that get rebalanced dynamically, so order of pair-checks change
- the solver is run only if "enough time" has passed, so faster machine call it less often
Each run will get a different result anyway.
 
>     2. Access to non-threadsafe objects from multithreading environment usually does not cause serious issues, so it is OK.

This is outright false.

akortunov:
> No it's not, but why do you think it is a problem?

Did you actually see the code of this PR before starting a discussion?

akortunov:
Closing since approach with `ClosestNotMeConvexResultCallback` will not work properly with async physics.

psi29a:
@akortunov perhaps @fr3dz10 can take this PR back up?

akortunov:
Probably, I do not know.

fr3dz10:
Modifying this PR to be "async-compatible" is not really a problem (I still don't see why ClosestNotMeConvexResultCallback is not tbh), but even then I think this PR doesn't go far enough: movement is still handled outside of physics with big fixed position increments, which leave the door open for clipping for fast projectiles. I can't reproduce the original report, but I believe that's what happen.
I'll extend it to handle projectiles movement the same way as actors if you don't want to do it.

PR #2184, opened 2019-02-23
[Testing needed] Always run startup script once at engine start (bug #4877)

elsid:


Capostrophic:
Won't this cause Bad Things™ to happen if the script gets ran while inside main menu? It probably shouldn't be ran when the game state is not Running. But then side effects of the script won't be carried over into the first loaded save.

elsid:
Yeah, it was bad idea to this in that way. Game just crashes in main menu.

psi29a:
@Capostrophic @akortunov you guys ok with this?

akortunov:
LGTM, if it really works (can not test it right now).

psi29a:
Waiting for more testing... (confirmation)

Capostrophic:
Appears to be fine. Though it obviously only works if the main menu is bypassed by the respective command line argument or by a saved game file command line argument.

PR #2185, opened 2019-02-23
[Discussion] Simple benchmark environment

elsid:
The idea is to do some manipulations with a player character and somehow measure performance. Tests should be stable and reproducible. I don't suggest how to measure performance but only how to setup environment. My idea is to use scripts as benchmarks scenarios. Benchmark script require a game mod with named script to be executed on each engine frame and startup script to start this benchmark script.

To control player character there are some commands like `setangle` and `move`. But they don't actually simulate player's actions. And we can't exit a game from script. Benchmark should be finite so we need a command to stop game process. Following commands should solve this:
* `setmovementposition <axis> <value>` - changes actor's movement [position](https://github.com/OpenMW/openmw/blob/da018199d21ea27b5783a4ca9294086d52069ecb/apps/openmw/mwmechanics/movement.hpp#L11);
* `setmovementrotation <axis> <value>` - changes actor's movement [rotation](https://github.com/OpenMW/openmw/blob/da018199d21ea27b5783a4ca9294086d52069ecb/apps/openmw/mwmechanics/movement.hpp#L12);
* `quit` - request game quit;

Simple benchmark script could be like this (omwaddon: [benchmark.zip](https://github.com/OpenMW/openmw/files/2897235/benchmark.zip)):
```
begin Benchmark

player->setflying 1
player->setspeed 1000

if (player->getpos z < 15000)
    player->setangle x 90
    player->setmovementposition y -1
elseif (player->getpos y < 21000)
    player->setangle z 270
    player->setmovementposition x 1
else
    StopScript Benchmark
    quit
endif

end Benchmark
```

And startup:
```
coc "Seyda Neen" ; required to make benchmark script work properly
StartScript Benchmark
```

To run it just add addon in lancher and then run openmw with `--script-run` and `----load-savegame` like this:
```
openmw --script-run start.txt --load-savegame mysave.omwsave
```
Or manually load some save and then execute start script in console.

Any ideas how we can do this better?

PS: this pr already partially resolves https://gitlab.com/OpenMW/openmw/issues/4875

### DIFF
--- a/apps/openmw/mwscript/docs/vmformat.txt
+++ b/apps/openmw/mwscript/docs/vmformat.txt
@@ -459,5 +459,9 @@ op 0x2000308: ToggleNavMesh
 op 0x2000309: ToggleActorsPaths
 op 0x200030a: SetNavMeshNumber
 op 0x200030b: Journal, explicit
+op 0x200030c: SetMovementPosition
+op 0x200030d: SetMovementPosition, explicit
+op 0x200030e: SetMovementRotation
+op 0x200030f: SetMovementRotation, explicit
 
-opcodes 0x200030c-0x3ffffff unused
+opcodes 0x2000310-0x3ffffff unused
--- a/apps/openmw/mwscript/docs/vmformat.txt
+++ b/apps/openmw/mwscript/docs/vmformat.txt
@@ -464,5 +464,7 @@ op 0x200030d: SetMovementPosition, explicit
 op 0x200030e: SetMovementRotation
 op 0x200030f: SetMovementRotation, explicit
 op 0x2000310: Quit
+op 0x2000311: SetCollisionMode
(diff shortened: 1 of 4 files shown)


Capostrophic:
Mildly concerned with mwscript additions. Those are vetoed.

akortunov:
As for me, it makes sense to implement such things only via Lua.

Capostrophic:
Yeah, same.

psi29a:
I don't think we are ready for this, mwscript is, for now, set in stone. It will not be evolving.

I think at this point, this PR should be applied to TES3MP (OpenMW fork) and their Lua.

akortunov:
So is there an agreement to do not touch MWScript? Can we close this PR then?

psi29a:
If possible, try to do this with Lua in TES3MP.

PR #2186, opened 2019-02-24
Make limit of max move steps depending on queue size (bug #4881)

elsid:
See https://gitlab.com/OpenMW/openmw/issues/4881 for more deails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
     Bug #4837: CTD when a mesh with NiLODNode root node with particles is loaded
     Bug #4860: Actors outside of processing range visible for one frame after spawning
     Bug #4876: AI ratings handling inconsistencies
+    Bug #4881: Slow PhysicsSystem::applyQueuedMovement for large multiplication of queue size and number of steps
     Feature #2229: Improve pathfinding AI
     Feature #3442: Default values for fallbacks from ini file
     Feature #3610: Option to invert X axis
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -1226,7 +1226,7 @@ namespace MWPhysics
 
         mTimeAccum += dt;
 
-        const int maxAllowedSteps = 20;
+        const int maxAllowedSteps = std::min(20, std::max(1, 40 / std::max(1, static_cast<int>(mMovementQueue.size()))));
         int numSteps = mTimeAccum / (mPhysicsDt);
         numSteps = std::min(numSteps, maxAllowedSteps);
 


akortunov:
IIRC, when I asked about it in the last time, I was told that physics subsystem in OpenMW was designed for fixed (60) FPS and this behaviour is intended.

Also I suppose this PR will cause weird sliding and teleporting with low FPS as with low value of OPENMW_PHYSICS_FPS environmental variable, which does the pretty same thing, but can be controllable by user.

The proper solution would be to optimize collision detection itself.

wareya:
No matter how much you optimize physics there's always going to be a case where someone can't run the physics at 60fps. If someone can't run the game's physics at 60fps then it ***absolutely definitely positively must not*** build up a large backlog of physics updates to handle. It needs to either fall back to delta time (to catch up), or slow down (and not catch up). Anything else is wrong.

The fact that the engine tries to catch up all at once in a single loop without falling back to delta time is ridiculous. I can only see that resulting in stop-and-go stuttering. If you can't run 60 physics updates per second, then trying to do 20 physics updates at once is just going to stall you for a minimum of 1/3 of a second before anything can render.

If you want to catch up without using delta time, you need to go one frame at a time until you exceed a certain amount of time, not a certain amount of physics steps, which will slow you down until you CAN catch up. Or you have to fall back to delta time and catch up that way. Or you have to give up on catching up at all and just let the game slow down.

"Fixed timestep" only works when you can actually run the physics at a fixed interval. If you can't then you have to fall back to something else.

elsid:
This change just makes already existing limit more adaptive to make game more responsive. Doing for example 3 steps of fixed intervals instead of 10 just will make smaller visible change per each frame. If there are more than 1 frame like this in a row, player may will see game being like a slow motion. But all controls, sound and rendering would be without freezing.

There is no way to optimize infinitely. Real time systems (I believe this engine is) should be responsive in a fixed time interval. Only one way to do this is to divide calculations into chunks with approximately same but much less than interval calculation time. And do them as they fit interval with consistent result of each one.

akortunov:
So are we OK with adaptive physics framerate instead of fixed one?

akortunov:
Casting @psi29a, @kcat, @Capostrophic and @AnyOldName3 here.

akortunov:
An another possible issue - every step handles the mPhysicsDt time, so if the physics update handles lesser time than requested (a frame duration), the mTimeAccum variable will grow from frame to frame.
As a result, the physics system will handle less time when needed (for example, the game session lasts for 60 seconds, but physics system updated only 10 seconds).
So instead of capping the number of steps, there should be an adaptive step duration (mPhysicsDt).
It is in case if we want to change our current approach at all, of course.

wareya:
"Adaptive step duration" is "delta time". Delta time causes problems for how consistent physics are, so when there aren't any performance problems, you want to use a fixed timestep, which is what OpenMW is doing right now.

But IIRC right now OpenMW doesn't fall back to anything gracefully when it has physics performance problems. It queues up more physics steps. Which is bad. It should settle for slowed-down physics (fixed step "duration", fewer steps per real time) or switch to delta time (adaptive step duration). In that sense (if I'm recalling the situation correctly) OpenMW's "current approach" is broken and needs to be changed *in some way*. It's just that there are multiple valid options.

kcat:
> So are we OK with adaptive physics framerate instead of fixed one?

I would say no. Adaptive physics framerate prevents physics from being deterministic, which can cause problems when trying to debug physics issues since the physics step is variable (larger steps may allow something to happen that smaller steps don't, and vice-versa). It can also actually increase CPU load when the physics is locked to FPS, because higher FPS means more physics updates that aren't otherwise necessary. There would also be smaller between-frame movements, which can create numeric stability problems.

And this is on top of other timing issues as we don't get accurate frame timing. Because we can start processing the next frame before the GPU has finished rendering and displaying the last frame, the actual amount of time since the last frame may be off (visually this can create so-called "micro-stutters", as the CPU may think it's running behind and missed a frame while the GPU is still on time, and vice-versa).

At best, I would say there could be some kind of "vsync" for physics (where it uses an integer multiple of a set rate), but I'd still be wary as it creates non-determinism, and the CPU and GPU timing may be desynced causing microstutters in the physics.

wareya:
Okay, so slowdown it is.

akortunov:
> It can also actually increase CPU load when the physics is locked to FPS, because higher FPS means more physics updates that aren't otherwise necessary.

We already have the 60 FPS cap in physics system, so there should not be such issues.

And we returned to the original question - what to do if the game is running with lesser FPS than 60 (especially because of slow physics update)?

wareya:
If physics can run at its intended tickrate but the game can only render 20 frames a second, it should still simulate its intended tickrate for physics, just not render all of them. (Of course, there's interpolation, so it's not "render a physics frame", but still.)

akortunov:
> If physics can run at its intended tickrate but the game can only render 20 frames a second, it should still simulate its intended tickrate for physics, just not render all of them. (Of course, there's interpolation, so it's not "render a physics frame", but still.)

I suppose you talk about case when the game is limited by the rendering thread, not by a main thread.
We need somehow to handle an another case - when the physics update takes a very long time because of large mount of simulation steps. We can not just blindly add additional steps - it decreases FPS and we need to add additional steps again and again (the "death spiral").

AnyOldName3:
> How do other game engines handle this situation?

Some of them will declare your computer to be below minimum system requirements and outright refuse to run, so the horrible stutter and freezing we have now isn't the *least* graceful thing to have ever been released.

Physics determinism is kind of a must-have for reproducing user issues on developer machines. If we sacrifice it, it *will* make things harder.

However, we could maybe simulate it by doing something clever. For example, if we guarantee it always runs at 60, 30, 20, 10 or 5 ticks/second (and write to the log when falling back to a lower value) at least developers will be able to get a log and save and manually cap the physics tick rate to what the user was experiencing. Maybe we could even have a super secret developer setting that lets only people who can figure out how to apply a patch to OpenMW and rebuild it change the physics tick rate with a hotkey to simulate a user's run. If we're going to be brainstorming esoteric solutions, we could even try some kind of replay system where a user's inputs, random seed and anything else that changes observable behaviour gets logged and then sent to a dev to be replayed.

wareya:
>Some of them will declare your computer to be below minimum system requirements and outright refuse to run

Not a single game I've ever played has actually refused to run because of something like this. The only things I've even seen are slowdown and variable step sizes.

>Physics determinism is kind of a must-have for reproducing user issues on developer machines. 

You realize that **falling back** to delta time when physics **cannot** run at full speed doesn't mean that normal gameplay is going to have nondeterministic physics, right? It's going to be limited **strictly** to situations that would cause slowdown (not stuttering, **slowdown**; or in the case of OpenMW right now, the "death spiral") otherwise.

>For example, if we guarantee it always runs at 60, 30, 20, 10 or 5 ticks/second 

You cannot attempt to many any guarantees of this sort with an engine like OpenMW. Different situations will have different levels of demandingness for different players. This isn't a 90s or early 2000s console game where the hardware and game content were both locked down. This is a moddable PC game engine.

>If we're going to be brainstorming esoteric solutions, we could even try some kind of replay system where a user's inputs, random seed and anything else that changes observable behaviour gets logged and then sent to a dev to be replayed.

**You guys can't try to beat around the bush and find a way around this problem. You have to pick between falling back to slowdown and falling back to variable physics step sizes. There are no other valid options.**

There is no brainstorming to be done. Every possible valid solution is either a form of slowdown or a form of variable physics timestep length, even if it does something like try to catch up for a few physics frames before changing modes or doing accelerations/integrations/curves at a higher internal fixed timestep but only doing collision calculations once.

AnyOldName3:
> > Some of them will declare your computer to be below minimum system requirements and outright refuse to run
> 
> Not a single game I've ever played has actually refused to run because of something like this. The only things I've even seen are slowdown and variable step sizes.

It's not that uncommon for games to refuse to launch if the CPU is below the minimum specification. Far Cry 4 caught a lot of flack for refusing to launch on a dual core, for example, and other games have followed in its feet since (such as Titanfall 2). As an earlier example, when my friend tried playing the first Witcher game on his laptop, we had to do something with configuration files to tell it that it was okay to run on that computer, even though it was below minimum spec (and it worked afterwards, just with stutters, so it wasn't something world-ending like missing GPU extensions).

> > Physics determinism is kind of a must-have for reproducing user issues on developer machines.
> 
> You realize that **falling back** to delta time when physics **cannot** run at full speed doesn't mean that normal gameplay is going to have nondeterministic physics, right? It's going to be limited **strictly** to situations that would cause slowdown (not stuttering, **slowdown**; or in the case of OpenMW right now, the "death spiral") otherwise.

There are computers that are always skirting the death spiral threshold right now (and as such, certain parts of Morrowind become totally unplayable). They're going to have nondeterministic physics for a significant percentage of time if that's the solution we go with. From a user's perspective, that's better than a death spiral, but if they see a bug in such a situation, it's going to be difficult to reproduce.

> > For example, if we guarantee it always runs at 60, 30, 20, 10 or 5 ticks/second
> 
> You cannot attempt to make any guarantees of this sort with an engine like OpenMW. Different situations will have different levels of demandingness for different players. This isn't a 90s or early 2000s console game where the hardware and game content were both locked down. This is a moddable PC game engine.

I think you're overthinking that idea. It's basically just course-grained delta time where the time step gets rounded up to the next 'permissible' value so we can attempt to use the same value on a developer's machine.

> > If we're going to be brainstorming esoteric solutions, we could even try some kind of replay system where a user's inputs, random seed and anything else that changes observable behaviour gets logged and then sent to a dev to be replayed.
> 
> **You guys can't try to beat around the bush and find a way around this problem. You have to pick between falling back to slowdown and falling back to variable physics step sizes. There are no other valid options.**
> 
> There is no brainstorming to be done. Every possible valid solution is either a form of slowdown or a form of variable physics timestep length, even if it does something like try to catch up for a few physics frames before changing modes or doing accelerations/integrations/curves at a higher internal fixed timestep but only doing collision calculations once.

I suggested two different ways we could have variable physics step sizes without making issue reproduction impossible. I'm not disagreeing that we're going to have to do something, but it would be much better if we don't sacrifice nice things when doing so.

wareya:
>It's not that uncommon for games to refuse to launch if the CPU is below the minimum specification.

This usually has to do with extensions like SSE4 and AVX. Cases where it's based on actual CPU power are extremely rare, and are usually imposed by abusive DRM, not the game's designers, and I've never run into them myself. Of course, if you're playing AAA games with a dual-core CPU today, you're doing something wrong, but OpenMW is not that kind of game engine, so it's not relevant.

>There are computers that are always skirting the death spiral threshold right now (and as such, certain parts of Morrowind become totally unplayable).

The death spiral has been a problem for long enough, worrying about what new minor problems you might introduce while fixing it is just bikeshedding.

>They're going to have nondeterministic physics for a significant percentage of time if that's the solution we go with.

Say they would be able to run it at 50fps instead of 60fps.

If your thing would switch between 30fps and 60fps frames in this situation, then IMO, it's probably going to be worse for physics consistency for physics to randomly jump between 60fps and 30fps, than it would be to run in a range between 45fps and 55fps.

And if you're talking about switching down to 30fps entirely when it can't maintain 60fps, you have to decide how to detect that, and that detection logic is going to run into possible problems too. Maybe someone will be in combat at 60fps, then the enemy AI summons an atronach and whenever that atronach walks it dips down to 30fps physics.

This might be fine, but it can cause problems. It's not a silver bullet. Players don't normally care about pure determinism, just that things don't warp around and that they don't suddenly become unable to walk through a hallway because the physics framerate dipped.

>I think you're overthinking that idea. It's basically just course-grained delta time where the time step gets rounded up to the next 'permissible' value so we can attempt to use the same value on a developer's machine.

Being able to test a small amount of physics timestep values is technically a good thing, but I think you're way overthinking how much of an impact it would have on bug reproduction.

>I suggested two different ways we could have variable physics step sizes without making issue reproduction impossible. I'm not disagreeing that we're going to have to do something, but it would be much better if we don't sacrifice nice things when doing so.

None of this really matters yet. Yes, there are ways to make variable physics timestep lengths less inconsistent, but you still have to pick a general approach and go with it before you can test out what techniques are actually going to have a real impact on making things more consistent.

Different techniques matter for different games and something that's good for one game will be bad for another. You won't know whether something is actually a problem or not until you implement it. You don't have to plan the specifics out so far in advance, in fact things are going to change when you go to implement it.

I just want to make it as clear as possible that any solution OpenMW goes with here is going to fall into one of two categories and they're both going to have certain minor problems no matter what. Fixing the death spiral is too important and it's so big of a problem that any problems you introduce in return are nothing in comparison.

Decide which nice things you're going to sacrifice. Do you want slowdown or do you want consistency problems? Pick one. You have to give up maintaining full speed or give up maintaining full consistency. When I read this thread I people trying to figure out how to make one of the two options better so that it's the obvious choice. But it's just not going to be an obvious choice. There's a degree of personal preference here. It's bikeshedding. Just pick something.

Here, I wrote something to try to keep this from turning into a circular debate: https://pastebin.com/raw/HjZaq0DA

Ultimately slowdown and variable timestep are both perfectly valid fallbacks for OpenMW. I personally prefer variable timestep, but only because slowdown is annoying. Just pick one and implement it.

AnyOldName3:
My personal preference would be delta time with some way to replay inputs using the same timesteps and random seed. This isn't just about reproducing physics bugs on dev machines - there were a bunch of times with shadows where I'd have loved to replay the exact same sequence before and after making changes, both to compare things like shimmering and to do useful performance benchmarks.

kcat:
> Say they would be able to run it at 50fps instead of 60fps.

With variable physics rates, it wouldn't be "50 fps", but more like 50.00321532 for one frame and 49.99975243 for another. Then 52.246543 for another and 47.75365 next. Because the deltas are different, calculations for things like movement will get different per-update scalars. This is particularly a problem since these values won't match the visual update rate, making movement subtly faster or slower (or varying) over time. A fixed delta prevents this since each update gets the same scalar.

Ultimately, the problem is that physics performance is not good. IMO, it would be better to fix the core of the performance problem rather than trying to find out how to compensate for poor performance. As previously pointed out, other games with far more complex physics don't have this issue, and even heavily modded vanilla Morrowind doesn't have a physics death spiral while maintaining a solid framerate.

wareya:
>With variable physics rates, it wouldn't be "50 fps" (...)

I know this. That's why I said "would be able to run", not "would be running". I understand the intricacies of this subject, both in theory and practice.

>A fixed delta prevents this since each update gets the same scalar.

You can just use a fixed delta but not try to catch up on missed physics frames and cause a performance death spiral like OpenMW currently does. Physics run in slow motion if they can't run at full speed, but that's perfectly okay, that's what happens when you don't fall back to a variable timestep.

>This is particularly a problem since these values won't match the visual update rate, making movement subtly faster or slower (or varying) over time. 

Nobody on the face of the planet is going to notice a 1~2ms physics latency shift, especially considering that the camera angle, the most latency-sensitive input, updates independently of the simulation.

On the other hand, microstuttering because of timer precision problems is a very realistic problem, and the choice between microstuttering and slowdown is a valid choice to take either option regarding.

>Ultimately, the problem is that physics performance is not good.

There's ***always*** going to be cases where a user can't run the game at full speed. That user's still going to exist somewhere if an angel comes out of the sky and gives OpenMW a bespoke physics engine, or we finally switch to double-precision bullet and AABBs. It's still going to be a problem.

>IMO, it would be better to fix the core of the performance problem rather than trying to find out how to compensate for poor performance.

There's no "rather than" here. Obviously OpenMW should do both. But optimization is very hard, so making it so the engine doesn't spiral off and "die" is the obvious first step. Spiraling off and dying is not a valid option. It ***must*** either slow down or fall back to a variable timestep.

You obviously don't like variable delta timesteps, so pick slowdown.

Slowdown is super easy to implement. Just limit the size of the physics queue to a small size (like 1 or 2). Do this and this entire problem will disappear until someone on reddit complains about the physics speed slowing down sometimes, which you can just ignore if/when it happens if you want.

kcat:
> You can just use a fixed delta but not try to catch up on missed physics frames and cause a performance death spiral like OpenMW currently does. Physics run in slow motion if they can't run at full speed, but that's perfectly okay, that's what happens when you don't fall back to a variable timestep.

Which I believe is what we've said (I have, at least). Alternatively, use a fixed delta that can be scaled by an integer multiple as needed, so it can go from 60 to 30 to 20 to 15, etc. I'd still be wary with that latter option, but it would be better than being fully variable.

wareya:
>Alternatively, use a fixed delta that can be scaled by an integer multiple as needed, so it can go from 60 to 30 to 20 to 15, etc. I'd still be wary with that latter option, but it would be better than being fully variable.

This sounds like a good idea but it usually causes worse problems than it fixes unless you carefully program everything to take advantage of it.

Nobody here is supporting being "fully variable" if by that you mean running physics at exactly the rendering framerate even if it's far above 60fps.

akortunov:
So what did we decide to do? Do not touch anything now and return to this question once the physics will be more optimized?

wareya:
That's the one choice of action that doesn't make any sense. The path of least resistance is to make physics slow down when they fall behind.

akortunov:
Closing this PR since in its current state causes issues instead of solving them and since there is no agreement about how to handle this bug. Feel free to start a forum discussion.

PR #2187, opened 2019-02-24
Option to set specific random seed for random number generator

elsid:
This one is more like debugging feature. It would be nice to have reproducible game situations. When random seed depends on time and engine makes a lot of decisions by rolling a dice this option can help. Probably it is not the best way to do it. I will listen other suggestions.

### DIFF
--- a/apps/openmw/main.cpp
+++ b/apps/openmw/main.cpp
@@ -3,6 +3,7 @@
 #include <components/files/escape.hpp>
 #include <components/fallback/validate.hpp>
 #include <components/debug/debugging.hpp>
+#include <components/misc/rng.hpp>
 
 #include "engine.hpp"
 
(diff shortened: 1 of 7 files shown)


components/misc/rng.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include <cassert>
 #include <random>
+#include <chrono>
```
Capostrophic: A bunch of stuff depends on rng.hpp and including chrono in it is something scrawl advised against.

elsid: Ok, I will hide default seed generation into a function and will place definition into .cpp file.

psi29a:
This still adds boost::chrono correct? Yet another external dependency... can we implement this without boost?

psi29a:
Right, so this adds std::chrono and this is abstracted away... I'm OK with this, waiting for @Capostrophic opinion as well.

components/misc/rng.hpp
psi29a: @Capostrophic is this OK?

apps/openmw/main.cpp
```diff
@@ -3,9 +3,12 @@
 #include <components/files/escape.hpp>
 #include <components/fallback/validate.hpp>
 #include <components/debug/debugging.hpp>
+#include <components/misc/rng.hpp>
 
 #include "engine.hpp"
 
+#include <chrono>
```
Capostrophic: This header doesn't seem to be necessary.

Capostrophic:
Seems somewhat clunky but I'm not sure what could be better as a solution.

apps/openmw/main.cpp
psi29a: Agreed, this is only really needed for rng.cpp

elsid: Fixed

elsid:
Incude is fixed, also changelog record is added.

PR #2189, opened 2019-02-26
Ignore global variable stray explicit reference calls

Capostrophic:
[Bug 4888](https://gitlab.com/OpenMW/openmw/issues/4888)

This makes the expression parser clear explicit reference calls before names (these can normally only be variables in the expression parser) and makes the line parser reset the state to SetState when a potential member variable modification turns out to be an explicit reference call. So global variables can now "support" explicit reference calls by ignoring them, replicating a quirk one of Sotha Sil Expanded scripts is relying on.

OpenMW-CS will report these mistakes and OpenMW will report them as well.


psi29a:
I'll take it, thanks for warning explicitly about the issue.

PR #2190, opened 2019-02-27
Make Distant Terrain configurable

akortunov:
Implements [feature #4890](https://gitlab.com/OpenMW/openmw/issues/4890).

Based on [0af77e529e1c9771b7ee8e6e09df8db1959098c8](https://gitlab.com/OpenMW/openmw/merge_requests/51/diffs?commit_id=0af77e529e1c9771b7ee8e6e09df8db1959098c8), but this PR is a more clean.

As for terrain optimization, basically it is implemented [here](https://gitlab.com/OpenMW/openmw/merge_requests/51/diffs?commit_id=0bc02727b842cbbe249d9586202e8e57e4f7f3e5). Unfortunately, with this commit terrain near border disappear sometimes (bzzt managed to fix this issue somewhere in later commits).

### DIFF
--- a/components/terrain/chunkmanager.hpp
+++ b/components/terrain/chunkmanager.hpp
@@ -35,6 +35,7 @@ namespace Terrain
         void setCullingActive(bool active) { mCullingActive = active; }
         void setCompositeMapSize(unsigned int size) { mCompositeMapSize = size; }
         void setCompositeMapLevel(float level) { mCompositeMapLevel = level; }
+        void setMaxCompositeGeometrySize(float maxCompGeometrySize) { mMaxCompGeometrySize = maxCompGeometrySize; }
 
         void reportStats(unsigned int frameNumber, osg::Stats* stats) const override;
 
@@ -59,6 +60,7 @@ namespace Terrain
 
         unsigned int mCompositeMapSize;
         float mCompositeMapLevel;
+        float mMaxCompGeometrySize;
 
         bool mCullingActive;
     };
--- a/components/terrain/chunkmanager.hpp
+++ b/components/terrain/chunkmanager.hpp
@@ -32,13 +32,15 @@ namespace Terrain
 
         osg::ref_ptr<osg::Node> getChunk(float size, const osg::Vec2f& center, int lod, unsigned int lodFlags);
 
+        void setCullingActive(bool active) { mCullingActive = active; }
+        void setCompositeMapSize(unsigned int size) { mCompositeMapSize = size; }
+        void setCompositeMapLevel(float level) { mCompositeMapLevel = level; }
+
         void reportStats(unsigned int frameNumber, osg::Stats* stats) const override;
 
         void clearCache() override;
 
         void releaseGLObjects(osg::State* state) override;
-
-        void setCullingActive(bool active);
 
     private:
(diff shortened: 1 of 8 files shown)


terabyte25:
I suppose #2165 needs to be changed in accordance to this if this gets merged.

akortunov:
Ready for review now.

components/terrain/quadtreeworld.cpp
```diff
@@ -285,7 +295,31 @@ void traverseToCell(QuadTreeNode* node, ViewData* vd, int cellX, int cellY)
...
+    int lod = Log2(int(node->getSize()));
+    if (vertexLodMod > 0)
+    {
+        lod = std::max(0, lod-vertexLodMod);
+    }
+    else if (vertexLodMod < 0)
+    {
+        float size = node->getSize();
+        // stop simplyfing at this level since smaller chunks are already less dense compared to the farther stuff in view relative terms
+        // the cutoff is 1 because at size = 1 a node has getCellVertices() (ie 65) vertices just like larger nodes
```
unelsson: Explanation (in comments) could be clarified altogether, but at least simplyfing -> simplifying. Thinking of better way to say "less dense compared to the farther stuff in view", but I wonder if I understand the code correctly... My personal preference wouble be to have the comment before the first "if" that is related to the comment (line 306 in this case), but not sure if that's actually better.

components/terrain/chunkmanager.hpp
```diff
@@ -32,14 +32,17 @@ namespace Terrain
 
         osg::ref_ptr<osg::Node> getChunk(float size, const osg::Vec2f& center, int lod, unsigned int lodFlags);
 
+        void setCullingActive(bool active) { mCullingActive = active; }
+        void setCompositeMapSize(unsigned int size) { mCompositeMapSize = size; }
+        void setCompositeMapLevel(float level) { mCompositeMapLevel = level; }
+        void setMaxCompositeGeometrySize(float maxCompGeometrySize) { mMaxCompGeometrySize = maxCompGeometrySize; }
+
```
unelsson: I haven't seen function definitions doing something in headers in OpenMW yet, is this the preferred way for such short functions?

unelsson:
I tested this PR, played with various settings, and as far as I know, it works as intended. Reviewed with limited understanding, not much to comment really. Good job both of you.

components/terrain/chunkmanager.hpp
akortunov: Actually, we use approach (in the `apps/openmw/mwgui/formatting.hpp`), for example.
From what I can tell, there is no strict rule here.

Capostrophic:
I think "Optimize terrain QuadTree build" commit probably doesn't belong here and moving setCullingActive doesn't change anything in particular (other than *maybe* making the implementation file more clean).

akortunov:
> I think "Optimize terrain QuadTree build" commit probably doesn't belong here

Sorry, it was moved here by accident.

Capostrophic:
Did some playtesting. Works nice, and the new value for 'max composite geometry size' makes gameplay with TR quite a bit smoother. Thumbs up for bzzt and akortunov.

PR #2191, opened 2019-02-27
Do not store Viewer reference in the ViewData

akortunov:
Based on [MR #51](https://gitlab.com/OpenMW/openmw/merge_requests/51).

Currently the mViewer field is needed only to dispose Viewer object after the related ViewData object was disposed.

The main idea is to store smart pointers in the links map instead of ViewData object, so unused viewer object will be removed once link will be removed from the map.

Can be use such refactoring change or we should tell bzzt to drop it?

### DIFF
--- a/components/terrain/viewdata.cpp
+++ b/components/terrain/viewdata.cpp
@@ -118,7 +118,6 @@ ViewData *ViewDataMap::getViewData(osg::Object *viewer)
     if (found == mViews.end())
     {
         ViewData* vd = createOrReuseView();
-        vd->setViewer(viewer);
         mViews[viewer] = vd;
         return vd;
     }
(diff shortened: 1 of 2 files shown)


akortunov:
@AnyOldName3, is it OK for you?

AnyOldName3:
If a `ref_ptr` is enough to ensure everything gets cleaned up properly, that's fine.

akortunov:
> If a ref_ptr is enough to ensure everything gets cleaned up properly, that's fine.

Actually, we already use the `ref_ptr` here.

psi29a:
Generally bad form to merge your own PR/MRs :)  

@Capostrophic has that ability as well. ;)

Capostrophic:
:thinking:

do I?

PR #2192, opened 2019-02-28
Port minor fixes from bzzt's branch

Capostrophic:
While large-scale additions appear in bzzt's branch in hopes to bring the performance in the best shape it can be, there also are much smaller improvements, that, despite their laconity, can be quite effective. So while bzzt disagrees with the notion that his MR can't be viewed upon not as a whole, I took the liberty of being a crow I've been criticized as being and picking out some very small stuff that deserves to be merged straightaway without waiting for the rest of the MR to get reviewed (somehow) in the following decade.

The commits were cherry-picked, which retains the author and his email intact, even though committeer changes, and GitLab and GitHub will count these as his own commits.

But we want some actual documentation to the commits, right?

1. Don't preload terrain when loading an interior save.
Terrain is preloaded based on the player's position when loading saves. However, said position will be near 0, 0, 0 if the player is in an interior, so terrain in that Ashlands daedric ruins cell (you know the one) will be preloaded redundantly. This should improve interior save loading performance somewhat.

2. Don't use MWRender namespace in common terrain components.
It was used there for some reason even though MW%stuff% namespaces are specific to the engine app itself. 

3. Use emplace instead of find-assign in some places.
std::map::emplace/std::map::insert check for duplicates automatically and doesn't insert anything if the key is used, so making a search is inefficient. The second element of the returned pair gives information whether the insert was successful. The commit was slightly modified; originally it used insert(std::make_pair()) which is slightly more verbose and less efficient.

4. Don't reallocate light list vector unnecessarily.
It is perfectly fine to assume a scene has a maximum of 8 lights at any time, so it doesn't make sense to reallocate the vector which contains light sources every time it gets an element before that point.

5. Fix creature model instance preloading.
The instance wasn't actually used according to bzzt, so he replaced the call.

6. Fix crash when incremental compile operation is not used.
The code in the scope assumes mIncrementalCompileOperation is always available. It may not be, and if it's null then bad things will happen.

7. Prune empty quad tree nodes.
If the assumed bounding box of a distant terrain node is invalid, it means that there aren't any valid children added. So all possible children are removed, with an assumption made that a QuadTreeNode can't have more than four children (obviously).

8. Delete composite map layers on demand.
Composite map layers were previously removed all at once, which wasn't really good for performance. Now they are removed at the moment they are compiled so getting rid of them is gradual instead of being sudden. This fixes the initial ridiculous freeze upon being outside when Tamriel Rebuilt is used and reduces the fatality of frame drops upon turning as akortunov has observed and is actually quite a game-changer.

9. Fix quad tree node bounding box dimensions.
scrawl made a mistake so they used the full size of the quad tree node as a radius/dimension (starting from the center) instead of a half of the size. Bounding boxes affect culling so reducing the size of the bounding box allows to reduce the amount of terrain rendered at a single time. According to bzzt, this leads to about 20% less terrain being rendered (the terrain which was outside of the camera viewport).

10. Don't discard object cache with uninitialized timestamp.
While this is a divergence from OSG, objects with cache timestamp being 0 could have never gotten the real timestamp initialized so the cache could get dropped in an instant. This could happen when the terrain was being loaded.

### DIFF
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -1436,7 +1436,7 @@ namespace MWRender
                 return sceneMgr->createInstance(found->second);
         }
         else
-            return sceneMgr->createInstance(model);
+            return sceneMgr->getInstance(model);
     }
 
     void Animation::setObjectRoot(const std::string &model, bool forceskeleton, bool baseonly, bool isCreature)
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -409,7 +409,8 @@ namespace MWWorld
                 if (getPlayerPtr().isInCell())
                 {
                     mWorldScene->preloadCell(getPlayerPtr().getCell(), true);
-                    mWorldScene->preloadTerrain(getPlayerPtr().getRefData().getPosition().asVec3());
+                    if (getPlayerPtr().getCell()->isExterior())
+                        mWorldScene->preloadTerrain(getPlayerPtr().getRefData().getPosition().asVec3());
                 }
                 break;
             default:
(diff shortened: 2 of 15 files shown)


davidcernat:
Thanks a lot, bzzt and Capo. You guys make a wonderful team.

psi29a:
Thanks @davidcernat , you just made it official. :)

PR #2194, opened 2019-02-28
Navigator settings documentation

elsid:
See result [here](https://elsid-openmw.readthedocs.io/en/navigator_doc/reference/modding/settings/navigator.html).


psi29a:
Great work, very thorough and professional.

PR #2195, opened 2019-02-28
[Discussion, don't merge] Editor: Cell edge fixing tool

unelsson:
A menu-item that opens a window, a tool, that can be used for fixing broken cell edges. The tool "glues" cells together. This PR is based on 2167, but only actually needs CSMWorld::CellCoordinates::generateId from it. In this branch, the cell edge fixing code is duplicated in CSVRender::TerrainShapeMode and CSMWorld::FixCellEdgeTool, but this PR tries to keep commits only relevant to cell edge fixing tool.

I've personally found this tool useful, when doing heavy land edits, but it's up to debate whether this should be in the master, and if so, where it should be.


unelsson:
It has been brought to my attention that some mods have land height conflicts that produce torn cell edges. This tool can bind those edges back into place. Is this the wanted solution? Should I develop this further?

akortunov:
> It has been brought to my attention that some mods have land height conflicts that produce torn cell edges. This tool can bind those edges back into place. Is this the wanted solution? Should I develop this further?

It is hard to tell what this PR really does, but if it just adds a brush tool to fix gaps between terrain heights in neighbouring cells, (so every conflict should be fixed manually), it is not very useful.
IMO, an automatic tool which can detect and fix such issues would be a better solution, but cells usually contain objects, so it would not be a trivial task.

unelsson:
> if it just adds a brush tool to fix gaps between terrain heights in neighbouring cells, (so every conflict should be fixed manually), it is not very useful.
It's an UI for a for-loop that goes through edges of given cell, and binds it's heights to those of neighboring cells. I don't think this is very useful as is, but it was easy to code it, so it's a quick workaround for cases where cell edges break. Maybe this could be expanded into automatic detect and fix -tool?

akortunov:
@psi29a, @zinnschlag, what can you suggest to do here?

> Is this the wanted solution? Should I develop this further?

Here is the main question. So far this PR is a WIP with more than year without updates, and someone should decide what to do with this PR.

zinnschlag:
This PR is very difficult to review. It seems to contain a lot of unrelated changes. Probably based on another PR that hasn't been merged yet?

From what I got, the idea of an edge fixing tool seems interesting. However the way the UI is implemented (apparently a separate window), is a complete no-go. We would have to find another (more consistent) way to integrate it into the UI.

In summary: Good idea, but would require a major cleanup and rework.

akortunov:
> Good idea, but would require a major cleanup and rework.

Closing then. I suppose that before inplementing such large feature, it should be discussed on forum first.
Also this PR is pretty outdated.

psi29a:
@unelsson Any thoughts on this, I know you've been busy with other things lately.

PR #2196, opened 2019-02-28
[Discussion, don't merge] Editor: Tool for scaling land

unelsson:
Menu-item, opening a window that holds UI for scaling tool. You choose a box of cells for the source, by setting two coordinates of the corners of the box, then you choose a box of cells for the target. Pressing scale reads from the source, and writes the data scaled to target. Height scaling uses nearest neighbor or bilinear filtering, textures are always scaled with nearest neighbor -method. It's up to debate whether this tool should be in the master, when it should be done and where it should go.

Currently this feature can't mirror, and values of corner A has to be smaller than those of corner B. Also it's not a good idea to make source and target area overlap, as that will likely corrupt the data at some point (this feature edits the data directly without reading everything to the buffer first).

This PR is based on 2167, using the CSMWorld::CellCoordinates::generateId, but nothing else from that.

Screenshot https://imgur.com/a/yueL6Og (the land tearing is fixed though, as well as some other bugs)

Known bugs:
* ~~A strange texture offsetting problem with this PR. It seems like land heights scale more or less how they should, but textures are offset slightly. I've tried various fixes, but I'm running in circles. I've tried various offsetting values and techniques, but nothing seems to fix this. Part of it is related to Morrowind's texture offset of +- 0.25f, and then something else with the algorithm.~~ ---> Major scaling bug was finally fixed through complete rethinking of scaling logic, although a minor offset thing still exists, likely because of +-0.25f offset used by Morrowind.
* ~~It looks like the scales of land heights and land textures do not match exactly.~~ --> (fixed)
* Land normals are calculated at the same time as everything else, while they correctly should be calculated in a separate loop later to avoid wrong values at cell edges.
* Land vertex colors are not scaled at all.


unelsson:
It has been brought to my attention that some mods have land height conflicts that produce torn cell edges. This tool can bind those edges back into place. Is this the wanted solution? Should I develop this further?

akortunov:
> It's up to debate whether this tool should be in the master, when it should be done and where it should go.

Are there use-cases for this feature? Usually cells contain objects, so terrain scaling makes no sense.

unelsson:
> > It's up to debate whether this tool should be in the master, when it should be done and where it should go.
> 
> Are there use-cases for this feature? Usually cells contain objects, so terrain scaling makes no sense.

I've found this useful when creating big landmasses for new games. I've used this to make a miniature of a world, sized 3x2 cells, then scaled it to 30x20 cells. There's plenty of work to make such a landmass, but creating a miniature first lets you tune heights of mountain ranges etc., and helps to understand the sizes of world regions.

This could be useful also for total overhauls or huge landmass mods, Tamriel Rebuilt maybe? There are also some modders who just want to experiment with huge Morrowind landmasses. Having a feature for scaling or moving objects would be helpful for that task, but it's not my piece of cake (who knows, maybe some coder out there wants to have a go for it later).

akortunov:
Looks like this PR did not get any feedback and did not get any updates for 15 months.
@psi29a, @zinnschlag, can you tell if such tool should be present in the OpenmW-CS 1.0, or it is an extended post-1.0 feature?

zinnschlag:
Exactly the same issues as with the other editor PR (https://github.com/OpenMW/openmw/pull/2195).

psi29a:
@unelsson thoughts?

PR #2197, opened 2019-03-01
Optimize terrain QuadTree build

akortunov:
1. QuadTree node can contain either 0 or 4 child nodes, so we can reserve space to make insert faster.
2. `osg::Group::addChild()` does many additional stuff: fixes deprecated data in geometries, checks if child nodes have custom callbacks, etc. In theory, we can just add child nodes without additional stuff.

On my hardware speedup is about 10% with vanilla landmass (0.24-0.25s vs 0.25-0.27s to build the terrain QuadTree).

### DIFF
--- a/components/terrain/quadtreenode.cpp
+++ b/components/terrain/quadtreenode.cpp
@@ -71,16 +71,6 @@ QuadTreeNode::~QuadTreeNode()
 {
 }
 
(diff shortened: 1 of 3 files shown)


Capostrophic:
Appears to be working fine.

psi29a:
Thanks!

PR #2199, opened 2019-03-02
Use C++11-style loops in the GUI instead of iterators

akortunov:
Currently despite we migrated to the C++11, most of our codebase still uses old-school iterators.

This PR replaces iterators in our GUI code to C++11-style loops to make code more readable.


psi29a:
Waiting on CI

akortunov:
CI is finished.

PR #2200, opened 2019-03-02
Replace boost GCD by the homebrew implementation

akortunov:
CI showed an another warning about undeclared BOOST_VERSION variable, and preprocessor stuff looks ugly.

Do we really need a boost here? We just need to determine an aspect ratio, so a homebrew implementation will be enough. 

If OpenMW will use the C++17 stuff, we can use the `std::gcd` here.

### DIFF
--- a/apps/launcher/graphicspage.cpp
+++ b/apps/launcher/graphicspage.cpp
@@ -1,11 +1,5 @@
 #include "graphicspage.hpp"
 
-#include <boost/version.hpp>
-#if BOOST_VERSION >= 106500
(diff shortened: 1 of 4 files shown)


Capostrophic:
I don't think the new file is going to be widely used in the code base at any point before C++17 comes. Maybe use lambdas instead?

psi29a:
I know, already done in my C++17 branch ;)  Sadly we aren't there yet, msvc was the big blocker here.

psi29a:
I'll this this over boost, cheers

PR #2201, opened 2019-03-02
Optimize terrain

akortunov:
Task [#4695](https://gitlab.com/OpenMW/openmw/issues/4695).

In my testing OpenMW now consumes 2GB of memory on start with the Tamriel Rebuilt and Stirk instead of 5.5Gb.

Also helps much with freezes during turning.
Freezes still may happen, but only with insane viewing distance (there are some fixes in bzzt's brach which can help here).

The only issue is default cells handling. Usually we render some default cells near player and it is enough when the Distant Terrain is disabled, otherwise player can see holes in terrain. Water shader allows to mitigate this issue.

What we should do here? Assigning default data for every cells (as we do in the 0.45) is too expensive). Suggest users to enable water shader with Distant Terrain?

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -371,8 +371,10 @@ namespace MWRender
 
         mNearClip = Settings::Manager::getFloat("near clip", "Camera");
         mViewDistance = Settings::Manager::getFloat("viewing distance", "Camera");
-        mFieldOfView = Settings::Manager::getFloat("field of view", "Camera");
-        mFirstPersonFieldOfView = Settings::Manager::getFloat("first person field of view", "Camera");
+        float fov = Settings::Manager::getFloat("field of view", "Camera");
+        mFieldOfView = std::min(std::max(1.f, fov), 179.f);
+        float firstPersonFov = Settings::Manager::getFloat("first person field of view", "Camera");
+        mFirstPersonFieldOfView = std::min(std::max(1.f, firstPersonFov), 179.f);
         mStateUpdater->setFogEnd(mViewDistance);
 
         mRootNode->getOrCreateStateSet()->addUniform(new osg::Uniform("near", mNearClip));
@@ -1200,7 +1202,9 @@ namespace MWRender
         mUniformFar->set(mViewDistance);
 
         // Since our fog is not radial yet, we should take FOV in account, otherwise terrain near viewing distance may disappear.
-        float distanceMult = std::cos(osg::DegreesToRadians(mFieldOfView)/2.f);
+        // Limit FOV here just for sure, otherwise viewing distance can be too high.
+        fov = std::min(mFieldOfView, 140.f);
+        float distanceMult = std::cos(osg::DegreesToRadians(fov)/2.f);
         mTerrain->setViewDistance(mViewDistance * (distanceMult ? 1.f/distanceMult : 1.f));
     }
 
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -1198,6 +1198,10 @@ namespace MWRender
 
         mUniformNear->set(mNearClip);
         mUniformFar->set(mViewDistance);
+
+        // Since our fog is not radial yet, we should take FOV in account, otherwise terrain near viewing distance may disappear.
+        float distanceMult = std::cos(osg::DegreesToRadians(mFieldOfView)/2.f);
+        mTerrain->setViewDistance(mViewDistance * (distanceMult ? 1.f/distanceMult : 1.f));
     }
 
     void RenderingManager::updateTextureFiltering()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
     Feature #4887: Add openmw command option to set initial random seed
     Feature #4890: Make Distant Terrain configurable
     Task #4686: Upgrade media decoder to a more current FFmpeg API
+    Task #4695: Optimize Distant Terrain memory consumption
 
 0.45.0
 ------
--- a/apps/openmw/mwrender/terrainstorage.cpp
+++ b/apps/openmw/mwrender/terrainstorage.cpp
@@ -22,6 +22,15 @@ namespace MWRender
         mResourceSystem->removeResourceManager(mLandManager.get());
     }
 
+    bool TerrainStorage::hasData(int cellX, int cellY)
+    {
+        const MWWorld::ESMStore &esmStore =
+             MWBase::Environment::get().getWorld()->getStore();
+
+        const ESM::Land* land = esmStore.get<ESM::Land>().search(cellX, cellY);
+        return land != nullptr;
+    }
+
     void TerrainStorage::getBounds(float& minX, float& maxX, float& minY, float& maxY)
     {
         minX = 0, minY = 0, maxX = 0, maxY = 0;
--- a/apps/openmw/mwrender/terrainstorage.hpp
+++ b/apps/openmw/mwrender/terrainstorage.hpp
@@ -23,6 +23,8 @@ namespace MWRender
         virtual osg::ref_ptr<const ESMTerrain::LandObject> getLand (int cellX, int cellY);
         virtual const ESM::LandTexture* getLandTexture(int index, short plugin);
 
+        virtual bool hasData(int cellX, int cellY) override;
+
         /// Get bounds of the whole terrain in cell units
         virtual void getBounds(float& minX, float& maxX, float& minY, float& maxY);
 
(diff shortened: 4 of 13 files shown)


AnyOldName3:
One idea I've discussed in the past would be to just reuse the same terrain drawable for everywhere with default terrain, just with a different transform matrix.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -1195,6 +1195,10 @@ namespace MWRender
 
         mUniformNear->set(mNearClip);
         mUniformFar->set(mViewDistance);
+
+        // Since our fog is not radial yet, we should take FOV in account, otherwise terrain near viewing distance may disappear.
+        float distanceMult = std::cos(osg::DegreesToRadians(mFieldOfView)/2.f);
+        mTerrain->setViewDistance(mViewDistance * (distanceMult ? 1.f/distanceMult : 1.f));
```
Capostrophic: As something I haven't thought about, on extremely high FOVs (beyond 180 degrees) the mult becomes negative and bad things happen. Either distant mult should be absolute, or be replaced with a cosine of (pi-mFieldOfView/2.f) before usage, or there simply should be a check that distance mult is positive.

wareya: me: "that doesn't work though!"

me: test it out ingame

me: https://i.imgur.com/HJooPAl.jpg

Capostrophic: Yeah. The problem is that with these changes the terrain disappears too. My bad, I wrote this part.
screenshot021

unelsson:
Tested this with two huge landmasses, and also Morrowind, all with view distance 350000, compared this to master, and the improvement is great. Framerates with huge landmasses are 250-450, whereas master branch crashes (OOM with 24gb ram). Ram usage with such settings goes somewhere between 3-4gb (in system setup with 2,1gb RAM used before running OpenMW, that is OpenMW adds another uses 1-2gb ram with view distance 350k). Flew around with speeds 50, 500, 5000 and 50k. Looked around landmass shadow rendering, compared it to master, and can't see any problems with that. Asked to test with high FOV: Indeed beyond 180 degrees it goes haywire, but FOV slightly less than 180 does work, although obviously doesn't look very nice on a normal screen.

AnyOldName3:
How are you setting the fov above 180? Projection matrices don't make sense at or above 180.

wareya:
editing the config manually

AnyOldName3:
If I stick -50 as the full screen resolution, I can put that in the config, but that doesn't mean it's going to work.

wareya:
yes, obviously

psi29a:
Do we need to put a min/max bound on FOV to prevent craziness? And also warn user about non-sane config value?

AnyOldName3:
I'd just assumed something would be spitting out an error message at some point whether it was us upon encountering the value or OSG upon being asked to use the value. If no error is printed, we should probably make it be.

I don't think it's reasonable to bother making the distant terrain system robust against FOVs `>=` 180°, as that's never going to be achievable by a single camera, but maybe we should be making considerations for things like 360° screenshots where several different viewing angles are used back to back.

apps/openmw/mwrender/renderingmanager.cpp
akortunov: And which cap we should use for FOV? 110, as in UI?

wareya: please no, there are valid reasons to play with unreasonably high FoVs, making the config file cap 179 or something just to keep it from being a broken value is fine

should also enforce a minimum of like 1 or something if we don't already

akortunov:
Did anyone test this PR with 360° screenshots?

Capostrophic:
https://vizor.io/capostrophic/test/

So, yeah, I don't think it's fair to rely on refraction to hide the missing seafloor. I think that one cell (there are more but they are loaded around the player) is missing from Tamriel Rebuilt and from Morrowind.esm and Firemoth should be there.

akortunov:
So what we decided to do here?

Capostrophic:
Not having something universal to substitute default terrain without ridiculous memory usage is pretty sad and I still think it's desirable to look into a way to add that. But I think everyone can agree the terrain viewing distance limit thing could probably be merged straightaway.

akortunov:
> Not having something universal to substitute default terrain without ridiculous memory usage is pretty sad and I still think it's desirable to look into a way to add that.

Even in upstream master, the landscape plane is not endless anyway - we are limited by the QuadTree size, and the border is visible in places such as Solstheim. I see no reliable way to generate endless terrain plain - transparent vanilla water just was not designed for large viewing distance.

I tried to theck how MGE handles default cells, but it seems there is no easy way to disable water shader in MGE (even `togglewater` does not work).

Can we somehow limit viewing distance for things behind the non-shader water?

Capostrophic:
Won't argue with the first bunch of statements, but if I remember correctly this drops *any* cells that don't have any data and not just those "out of the bounds of the actual terrain" like it happened in 0.43.0 and earlier which leads to ugly holes like on the screenshot above. While 0.43.0 method obviously would make memory usage higher, surely it wasn't that bad if the ram usage only went bonkers once the entire quad tree was loaded into the memory. Couldn't it and the new approach (load cells around the player) be combined somehow?

akortunov:
> but if I remember correctly this drops any cells that don't have any data and not just those "out of the bounds of the actual terrain" like it happened in 0.43.0

Nope. This PR uses a same approach to render default cells, as our common terrain system - renders grid near player. 
In 0.43 we indeed did not render cells without terrain at all with Distant Terrain.

lysol90:
> Can we somehow limit viewing distance for things behind the water? We already do it, but only when the player is underwater.

Yes please. This would make a big difference I think. In screenshots taken from high above, you see right down to the ocean floor as long as the water isn't in an oblique angle. Limiting the view distance to prevent this would look much better. If that is possible at the moment of course.

akortunov:
@AnyOldName3, can we implement it without shaders?

AnyOldName3:
If you do it with fog, the fixed function pipeline can do it. I'm not sure how exactly we can make that work, though. If refraction is on, we can set a different fog range for its RTT, but otherwise, I don't think there's any way to tell the FFP to use one fog distance for things below z=0 in world coordinates and another for things above z=0.

akortunov:
> I don't think there's any way to tell the FFP to use one fog distance for things below z=0 in world coordinates and another for things above z=0.

So it seems that the only way to make water with Distant Terrain to look good is to use refraction?

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -1195,6 +1197,10 @@ namespace MWRender
...
```
Capostrophic: I think it should still make sense to put an upper limit on the multiplier, something like 2 (110 degrees FOV) or 3 (140 degrees FOV). Or it'll be 114 if 179 degree FOV is used.

tcotys:
I might be a bit off topic but there is lenses that uses a FOV > 180°.
It depends on the type of the projection used for the camera :
- the usual orthographic FOV projection is limited at 180°
- but (for exemple) stereographic projection can get up to FOV = 360° (i.e. the "small planet" effet)
with a relatively small FOV both projections a quite the same

AnyOldName3:
That doesn't mean that the maths of projection matrices works like that, though. A projection matrix converts a frustum to a cube, and frustums don't make sense as a concept with an angle at or above 180°. When rendering things to look like they've got a FOV above 180°, it has to be done by rendering multiple times with a smaller FOV, facing in different directions, and combining the results together.

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -1200,6 +1202,8 @@ namespace MWRender
         mUniformFar->set(mViewDistance);
 
         // Since our fog is not radial yet, we should take FOV in account, otherwise terrain near viewing distance may disappear.
+        // Limit FOV here just for sure, otherwise viewing distance can be too high.
+        fov = std::min(fov, 140.f);
         float distanceMult = std::cos(osg::DegreesToRadians(mFieldOfView)/2.f);
```
Capostrophic: fov variable is not used.

akortunov:
I suggest to recommend users to use a refraction with Distant Terrain for now and fill the feature request about default cells handling.
But as I said, I see no reliable to limit the underwater viewing distance with the fixed function pipeline.

Any other objections about this PR?

psi29a:
Last call...

AnyOldName3:
I'm a little confused why we're using the terrain grid to render default cells instead of doing it with the same system as the rest of distant terrain.

akortunov:
> I'm a little confused why we're using the terrain grid to render default cells instead of doing it with the same system as the rest of distant terrain.

We can not use the same system here - the quad tree size is limited.
Ideally, we need to use a some kind of paging here to cover the whole visible range for both terrain and water (yes, our water plane is not infinite too), but it is outside of scope of this PR.

Capostrophic:
>yes, our water plane is not infinite too

It's not, but its origin moves with the player, which makes the fact only noticeable on a humongous viewing distance.

akortunov:
So what's up with this PR?

PR #2206, opened 2019-03-03
Loop title screen music (bug #4896)

Capostrophic:
[Bug 4896](https://gitlab.com/OpenMW/openmw/issues/4896)

Fixes two issues:
1. If title music file was missing, then the logo video (the one after bethesda softworks intro video) wouldn't play due to the exception. Since the exception caused by the video file being missing would actually be catched in VideoWidget playVideo, it's unnecessary to have a try-catch block for that.
2. The titular one. I've added a new hardcoded Title playlist with a single track in it which will play in the main menu.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
     Bug #4876: AI ratings handling inconsistencies
     Bug #4877: Startup script executes only on a new game start
     Bug #4888: Global variable stray explicit reference calls break script compilation
+    Bug #4896: Title screen music doesn't loop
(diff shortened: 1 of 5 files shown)


Capostrophic:
This may not be consistent. Please stand by.

Capostrophic:
Pushed a new approach. It seems to be very hacky to me, but it should work fine even if you're not in the main menu GUI mode (e.g. settings menu while not in-game).

Capostrophic:
Any objections? I know that you can actually enter main menu even if you never visited it in the session (if you decide to stop loading a save while being in a save that bypassed the main menu) and the music will not play but that's pretty minor and should be an issue regardless of these changes.

PR #2207, opened 2019-03-04
[Trivial] Reuse a previously-saved value

CyberShadow:
`ptr.getCellRef().getLockLevel()` was already saved to a variable. Since it cannot change between invocations, reuse its value instead of fetching it again.


akortunov:
LGTM.

akortunov:
Also you can add yourself to the AUTHORS.md.

akortunov:
> Also you can add yourself to the AUTHORS.md.

I added the entry in the #2219.

psi29a:
Thanks! :)

PR #2208, opened 2019-03-04
Add sort button to item views

CyberShadow:
I was encouraged [on the forum](https://forum.openmw.org/viewtopic.php?f=2&t=5769&p=61812#p61812) to submit this as a pull request.

I'm aware that the contributing guide discourages patches of this sort. I don't have an opinion on whether this functionality should be part of OpenMW or not. But, if the maintainers agree in its favor, I don't mind spending some time polishing this patch for inclusion.

Issues so far:
- [ ] Replace `ItemView` button with layout buttons
- [ ] Replace hard-coded strings with localizable ones
- [ ] Review list of sorting modes
- [ ] Add a way to reverse sorting direction?

This is the original implementation as posted on the forum, for discussion / initial review.

### DIFF
--- a/apps/openmw/mwgui/itemview.cpp
+++ b/apps/openmw/mwgui/itemview.cpp
@@ -13,6 +13,7 @@
 
 #include "itemmodel.hpp"
 #include "itemwidget.hpp"
+#include "sortfilteritemmodel.hpp"
 
 namespace MWGui
 {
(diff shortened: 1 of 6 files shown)


apps/openmw/mwgui/itemview.cpp
```diff
@@ -48,6 +50,26 @@ void ItemView::initialiseOverride()
         throw std::runtime_error("Item view needs a scroll view");
 
     mScrollView->setCanvasAlign(MyGUI::Align::Left | MyGUI::Align::Top);
+
+    {
+        static const int width  = 60;
```
terabyte25: Not a good idea for using hardcoded values.

CyberShadow:
I am not working on this any more.

PR #2209, opened 2019-03-04
[Regression] Fix magic glow update for sheathing weapons

akortunov:
When there is a no dedicated _sh mesh for weapon, a Weapon Sheathing feature will attach a generic weapon mesh to the sheathing bone. 
If the weapon mesh contains controllers, they will be copied as well. It may lead to weird side effects (such as bow playing attack animation repeatedly because of morph controller).
So I used a workaround to disable update traversal for sheathed weapons. As a result, I accidentally broke the magic glow update for autogenerated holstered nodes. Also it **may** be related to black textures for enchanted sheathed weapons which a couple of users have (no one bothered to create a bugreport with their hardware specs, though).

In this PR I use an another approach - since it is hard to tell how to interpret current controller's time, disable controllers for autogenerated sheathed weapons at all.
It seems we use the same approach for non-weapon bodyparts for NPCs.

Are there better ways to do it? Maybe limit this workaround only for KeyFrame and GeomMorpher controllers?

### DIFF
--- a/apps/openmw/mwrender/actoranimation.cpp
+++ b/apps/openmw/mwrender/actoranimation.cpp
@@ -194,15 +194,16 @@ void ActorAnimation::injectWeaponBones()
     }
 }
 
-// To make sure we do not run morph controllers for weapons, i.e. bows
-class EmptyCallback : public osg::NodeCallback
+void ActorAnimation::resetControllers(osg::Node* node)
 {
-    public:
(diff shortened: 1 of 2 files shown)


akortunov:
Ok, Greatness7 uses a similar approach for weapon sheathing - he does not copy controllers for ranged weapons at all.

PR #2210, opened 2019-03-04
Fix pathfinding when scene is changed by scripts (bug #4763) (2nd try)

elsid:
There are several problems fixed:
1. Pathfinding uses player half extents instead of actor in interior cells.
2. Player scaling makes unavailable to find navmesh in exterior cells (also in interior because of 1).
3. Actor scaling makes unvailable to find navmesh in interior cells (bug after 1 is fixed).
4. Local scaling is not checked when object is updated in recast mesh.
5. Scaled, moved, rotated objects by scripts do not trigger navmesh update.
6. Using `setnavmeshnumber` for invalid value than for valid disable navmesh render without getting back.

Update: check is actor in a cell before check is it in an exterior cell

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -906,6 +906,14 @@ namespace MWPhysics
             return osg::Vec3f();
     }
 
+    osg::Vec3f PhysicsSystem::getOriginalHalfExtents(const MWWorld::ConstPtr &actor) const
+    {
+        if (const Actor* physactor = getActor(actor))
+            return physactor->getOriginalHalfExtents();
+        else
+            return osg::Vec3f();
+    }
+
     osg::Vec3f PhysicsSystem::getRenderingHalfExtents(const MWWorld::ConstPtr &actor) const
     {
         const Actor* physactor = getActor(actor);
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1337,6 +1337,8 @@ namespace MWWorld
 
         if (mPhysics->getActor(ptr))
             mNavigator->addAgent(getPathfindingHalfExtents(ptr));
+        else if (const auto object = mPhysics->getObject(ptr))
+            mShouldUpdateNavigator = updateNavigatorObject(object) || mShouldUpdateNavigator;
     }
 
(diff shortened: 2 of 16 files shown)


apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -906,6 +906,14 @@ namespace MWPhysics
             return osg::Vec3f();
     }
 
+    osg::Vec3f PhysicsSystem::getOriginalHalfExtents(const MWWorld::ConstPtr &actor) const
+    {
+        if (const Actor* physactor = getActor(actor))
```
terabyte25: Use a ternary conditional for compactness.

apps/openmw/mwworld/worldimp.cpp
```diff
@@ -3787,4 +3810,12 @@ namespace MWWorld
         mRendering->setNavMeshNumber(value);
     }
 
+    osg::Vec3f World::getPathfindingHalfExtents(const MWWorld::ConstPtr& actor) const
+    {
+        if (actor.isInCell() && actor.getCell()->isExterior())
```
terabyte25: Same here.

apps/openmw/mwphysics/physicssystem.cpp
Capostrophic: Why though.

elsid: Why? I don't think it has better readability and it's harder to debug (set breakpoint on specific branch for example). Performance is same for optimized code (look at generated assembly https://godbolt.org/z/iW8hAR).

akortunov:
Make sure this issue will not re-appear: https://gitlab.com/OpenMW/openmw/issues/4901

elsid:
I've checked, [this](https://gitlab.com/OpenMW/openmw/issues/4901) bug is not reproduced. New game and load game works also.

apps/openmw/mwphysics/physicssystem.cpp
psi29a: Readability is always better than compactness.

apps/openmw/mwworld/worldimp.cpp
psi29a: No.

psi29a:
Looks good, nothing obvious strikes me. Thank you.

PR #2211, opened 2019-03-05
Small script verifier tweaks

Capostrophic:
Changes two minor things:

1. Journal now has an explicit variant. Explicit reference calls for Journal actually matter (in case the scripter wants to use a different actor for %Name and other similar "macros"), parser deciding they're redundant is unfair (though it may have worked fine before). This is mainly to silence the spammy warning about stray explicit references for the instruction.
2. Script verifier messages are restructured again, placing line and column after the message, they didn't get sorted well alphabetically. Also, line number is fixed. Lines start from 1, not 0.


psi29a:
This seems reasonable, thank you.

PR #2212, opened 2019-03-05
Do not build path by navigator for pure water and flying creatures

elsid:
They don't need to move by surfaces and to open/close doors. Also prevents spam to log because of failure to find path.


psi29a:
Love this right here, no more slaughterfish spam! :)

terabyte25:
Nice.

PR #2213, opened 2019-03-05
 Use min point tolerance to prevent face wrong direction (bug #4814) 

elsid:
When next path point is too close to actor and it has speed 0, it can face direction not pointing to path target.


akortunov:
Nice.
But unfortunately it did not fix the issue with large creatures twitchy movement - it seems to be a separate issue.

PR #2214, opened 2019-03-05
DRY skill widget updates in stats window

CyberShadow:
The skill widgets were configured in two places: initial creation, and updates. The former was redundant, and duplicated the functionality of the latter.

To fix this redundancy, replace the duplicate code with a call to `setValue`.


akortunov:
Seems to work.

psi29a:
Thank you!

PR #2216, opened 2019-03-06
Fix windows CI scripts

elsid:
1. Cache proper files.
2. Remove annoying white spaces and trailing spaces
3. Fix compare with empty value.

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -82,7 +82,7 @@ while [ $# -gt 0 ]; do
 
 			t )
 				TEST_FRAMEWORK=true ;;
-				
+
 			h )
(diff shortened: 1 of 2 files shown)


Capostrophic:
>Cache proper files.

Oh, cool. This will cut the building times, right?

elsid:
> This will cut the building times, right?

Yes, but not much.

elsid:
One of linux builds was timeouted (has no relation to these changes). Need retry I think, but I don't have access to do it.

psi29a:
Looks like build times for openmw-cs with clang has shot through the roof... very spammy output about possible memory leaks and more. Not related here, but should be looked at.

PR #2221, opened 2019-03-07
Use C++11-style loops in the game world instead of iterators

akortunov:
The same thing, as in my previous PR, but for game world at this time.

It is a bit smaller since we did not use much iterators here to just emulate foreach over vector.


unelsson:
Any idea which is faster? I just found out about these methods, and have been thinking of good reasons for choosing one above other, except that one is "newer" or perhaps simpler to write.

Capostrophic:
As long as the range-based for loop uses references instead of copies they should be exactly the same in terms of compiled code but range-based for loop is arguably much more human-readable.

PR #2222, opened 2019-03-07
Attempt to fix Clang warnings about hidden virtual methods

akortunov:
Let see if it will help us.


kcat:
Personally I'd suggest adding `override` instead of `virtual`, for the virtual overrides. That'll ensure properly matching prototypes, and not inadvertently create unneeded virtuals.

PR #2223, opened 2019-03-07
Fix raki creatures in Skyrim: Home of the Nords (bug #4810)

Capostrophic:
[Bug 4810.](https://gitlab.com/OpenMW/openmw/issues/4810)

So basically raki creature model has a bip01 node under the 'root bone' node and such construction somehow causes movement accumulation not to work properly and subsequently breaks animations. Apparently bip01 needs to have higher priority than root bone in the movement accumulation root node assignment code to make it work, so I switched the node names around and now movement accumulation for dear, sweet, precious rakis works fine without breaking movement accumulation for the original reason why 'root bone' node name was considered - Correct UV Mudcrabs mudcrab model - or any vanilla creature.

'Root bone' node cannot be the root node. Both OpenMW and Morrowind drop it from the model.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
     Bug #4800: Standing collisions are not updated immediately when an object is teleported without a cell change
     Bug #4803: Stray special characters before begin statement break script compilation
     Bug #4804: Particle system with the "Has Sizes = false" causes an exception
+    Bug #4810: Raki creature broken in OpenMW
     Bug #4813: Creatures with known file but no "Sound Gen Creature" assigned use default sounds
     Bug #4815: "Finished" journal entry with lower index doesn't close journal, SetJournalIndex closes journal
     Bug #4820: Spell absorption is broken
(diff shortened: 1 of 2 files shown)


psi29a:
Is this something we need documented somewhere?

Capostrophic:
I googled it and it appears that in Oblivion and later games there isn't a concept of a "root bone" named node which can be the root bone at all, instead bip01 is always the root bone.

I tried to add a parent node to cliff racer model named "bip01" (replacing the normal root node) and the cliff racer animations were broken in Morrowind, so it looks like bip01 became the actual root bone.

So as far as I can guess if bip01 is present, it will always be the root bone. However, in some cases 'root bone' will  substitute bip01 in this role, e.g. in the mudcrab model from correct uv mudcrabs and in cliff racer model in Morrowind.

So it appears that indeed bip01 has higher priority than root bone node when the root bone is detected.

psi29a:
OK, then we should also document this somewhere for people interested in creating assets for OpenMW.

PR #2224, opened 2019-03-07
Fix ODR violation for VDSMCameraCullCallback

elsid:
This class is also defined in OpenSceneGraph at global namespace.

ODR violation is an undefined behavior. See https://en.cppreference.com/w/cpp/language/definition .

ASAN report:
```
=================================================================
==14531==ERROR: AddressSanitizer: odr-violation (0x563832d5f620):
  [1] size=624 'vtable for VDSMCameraCullCallback' /home/elsid/dev/openmw/components/sceneutil/mwshadowtechnique.cpp
  [2] size=624 'vtable for VDSMCameraCullCallback' /home/elsid/dev/OpenSceneGraph/src/osgShadow/ViewDependentShadowMap.cpp
These globals were registered at these points:
  [1]:
    #0 0x56383141ecf1 in __asan_register_globals.part.13 (/home/elsid/dev/openmw/build/clang/asan/openmw+0xdffcf1)
    #1 0x56383220cabb in asan.module_ctor (/home/elsid/dev/openmw/build/clang/asan/openmw+0x1bedabb)

  [2]:
    #0 0x56383141ecf1 in __asan_register_globals.part.13 (/home/elsid/dev/openmw/build/clang/asan/openmw+0xdffcf1)
    #1 0x7f2b6636828b in asan.module_ctor (/home/elsid/dev/OpenSceneGraph/build/clang/asan/lib64/libosgShadow.so.130+0x1c228b)

==14531==HINT: if you don't care about these errors you may set ASAN_OPTIONS=detect_odr_violation=0
SUMMARY: AddressSanitizer: odr-violation: global 'vtable for VDSMCameraCullCallback' at /home/elsid/dev/openmw/components/sceneutil/mwshadowtechnique.cpp
==14531==ABORTING
```

This pr is just a fix for one simple problem. But I think `components/sceneutil/mwshadowtechnique.cpp` should be devided into multiple files and all symbols should be moved into some namespace. 2.5k lines is too much.

### DIFF
--- a/components/sceneutil/mwshadowtechnique.cpp
+++ b/components/sceneutil/mwshadowtechnique.cpp
@@ -25,6 +25,8 @@
 
 #include <sstream>
 
+namespace {
+
 using namespace osgShadow;
 using namespace SceneUtil;
 
@@ -330,6 +332,8 @@ void VDSMCameraCullCallback::operator()(osg::Node* node, osg::NodeVisitor* nv)
 
     _projectionMatrix = cv->getProjectionMatrix();
 }
+
+} // namespace
 
 MWShadowTechnique::ComputeLightSpaceBounds::ComputeLightSpaceBounds(osg::Viewport* viewport, const osg::Matrixd& projectionMatrix, osg::Matrixd& viewMatrix) :
     osg::NodeVisitor(osg::NodeVisitor::TRAVERSE_ACTIVE_CHILDREN)


AnyOldName3:
> But I think components/sceneutil/mwshadowtechnique.cpp should be devided into multiple files and all symbols should be moved into some namespace. 2.5k lines is too much.

Normally I'd support chopping it up (and that would have made working on it a lot easier) but because it's based on a file from OSG, it makes it hard to deal with stuff like keeping track of what's code we've written versus what's original OSG code and potentially attempts to upstream it if it's structured differently.

Namespace-mess-wise, things are the way they are because I didn't want to have to indent everything and ruin the git blame when I moved it from `osgShadow` to `SceneUtil` and a `using namespace SceneUtil;` was enough to make it compile.

I don't exactly *get* why this change doesn't add more problems, though. Now the `using`s are enclosed within an anonymous namespace, there are a bunch of function definitions just loose in the `.cpp` file with no `namespace { ... }` or `using namespace ...;` to tie those definitions to the declarations they go with. As it's passed CI, I'm assuming there's something about C++ linking that I don't know, so I'd like to know what it is.

I'm guessing that the root cause of the error here is that we've got two identical/nearly identical (I can't remember if any of the changes I made to that class got kept) classes with the same name, one in the `components` lib and one in the `osgShadow` lib. If so, I'm guessing that this fix works because it's now in an anonymous namespace, which is different from it being in no namespace at all.

elsid:
> Normally I'd support chopping it up (and that would have made working on it a lot easier) but because it's based on a file from OSG, it makes it hard to deal with stuff like keeping track of what's code we've written versus what's original OSG code and potentially attempts to upstream it if it's structured differently.

Ok, I agree.

> Namespace-mess-wise, things are the way they are because I didn't want to have to indent everything and ruin the git blame when I moved it from osgShadow to SceneUtil and a using namespace SceneUtil; was enough to make it compile.

That's why I didn't make an indent and broke our styleguide rule. Probably we can reconsider this rule for namespaces. Just make it less strict. Styleguides are fine but they shouldn't prevent from writing correct code instead of incorrect.

> I don't exactly get why this change doesn't add more problems, though. Now the usings are enclosed within an anonymous namespace, there are a bunch of function definitions just loose in the .cpp file with no namespace { ... } or using namespace ...; to tie those definitions to the declarations they go with. As it's passed CI, I'm assuming there's something about C++ linking that I don't know, so I'd like to know what it is.

There is already an answer on stackoverflow: https://stackoverflow.com/questions/8775563/using-namespace-statement-inside-an-anonymous-namespace .

> I can't remember if any of the changes I made to that class got kept

`VDSMCameraCullCallback` has some difference with OpenSceneGraph master.

> I'm guessing that this fix works because it's now in an anonymous namespace, which is different from it being in no namespace at all.

It's an idea of my solution.

PR #2225, opened 2019-03-08
Support explicit limit of navmesh tiles for scene

elsid:
To allow player limit number of tiles new option is added. Now it is possible to have number of tiles at scene equal to physical limit without lost tiles.

Condition to check whether tile should be added to navmesh is following:
```c++
std::ceil(osg::PI * osg::square(std::ceil(getDistance(tilePosition, playerTile)))) * 3
   <= maxTiles
```
 `maxTiles`  is a physical limit of navmesh tiles number (1024 with current default settings).
Multiplication by 3 is required to left some space for new tiles. If player is moving and new tile is added before any other is removed, add may fail when there is no more space in navmesh and tile will be lost.  This pr removes multiplication by 3 to avoid wasted space. Instead it adds job reposting to avoid lost tiles. If `dtNavMesh::addTile` rejected new tile because there is no more space, job is reposted into the queue with reduced priority. Number of reposts is limited.

### DIFF
--- a/apps/openmw_test_suite/detournavigator/navigator.cpp
+++ b/apps/openmw_test_suite/detournavigator/navigator.cpp
@@ -61,6 +61,7 @@ namespace
             mSettings.mMaxSmoothPathSize = 1024;
             mSettings.mTrianglesPerChunk = 256;
             mSettings.mMaxPolys = 4096;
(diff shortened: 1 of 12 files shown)


psi29a:
Thanks! :)

PR #2227, opened 2019-03-08
Fix twitchy movement for massive creatures (bug #4907)

elsid:
1. Drop next path point for massive creatures at larger distance.
2. Drop path points while tolerance allows.
3. Increase path step size for massive creatures.

Also fixes twitchy movement for fast moving actors.


Capostrophic:
It seems to work fine.

PR #2229, opened 2019-03-08
Use recastnavigation 64bit polyref

elsid:
This allows to have up to `2**28 = 268435456` navmesh tiles and `2**20 = 1048576` polygons per each tile on scene.

Allow to cover all cells with `exterior cell load distance = 3`: https://streamable.com/2i02r
And with some more tweaks `aitravel` works for large distances: https://streamable.com/td24o

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,9 @@ if (WIN32)
     add_definitions(-DNOMINMAX -DWIN32_LEAN_AND_MEAN)
 endif()
 
+# Use recastnavigation 64bit polygon reference
+add_definitions(-D DT_POLYREF64)
+
 if (NOT WIN32 AND BUILD_WIZARD) # windows users can just run the morrowind installer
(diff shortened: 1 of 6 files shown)


akortunov:
Do we really need it now? He do not update (and even do not render) actors outside the 7168 area anyway. Who is supposed to use these tiles?

akortunov:
So what did we decide to do with this PR?

psi29a:
What is the performance impact here?

akortunov:
As for me, it would be nice to have a some kind of simplified movement system (as in S.T.A.L.K.E.R., for example) for actors in inactive cells, rather than just use the huge scene for navmeshes (most of which will be unused 99% of time anyway).

akortunov:
So what should we do here?
Using the whole scene (outside of 3x3 cells grid) for navmeshes makes no sence.

akortunov:
There is no point to cover large areas with navmeshes at the moment.

PR #2230, opened 2019-03-09
Minor preloading improvements

akortunov:
1. Allow to interrupt terrain preloading if we set an mAbort flag for this cell.
2. Calculate bound volumes immediately when we cache instances to do not do it for all cached instances during first update traversal in the new cell.


psi29a:
Who would set the abort flag? :)

Is this just prep work for something to come?

akortunov:
> Who would set the abort flag? :)
Is this just prep work for something to come?

We already have this flag and set it by the `abort()` call for the TerrainPreloadItem. The point is just to cancel the `mWorld->preload()` call to make cleanup a bit faster.

PR #2231, opened 2019-03-09
Get rid of C-style limits in the shadows code

akortunov:
The only thing which I did not get, why there was a mix between DBL_MAX and FLT_MAX for the same double variable in the `adjustPerspectiveShadowMapCameraSettings`. 
@AnyOldName3, is it intended?


AnyOldName3:
I'll have a look later to see if it matters. It's possibly just weird inherited OSG code from when it used to be a float.

AnyOldName3:
I doubt very much that it matters. I'm clicking merge.

PR #2234, opened 2019-03-09
Some more improvements from bzzt's branch

Capostrophic:
I've grabbed some more mostly no-brainer stuff from bzzt's branch that is mainly related to esmterrain/storage.cpp file.

1. Vertex colors data type matches the source data. Right now they are expressed in floating point numbers in [0, 1] range even though the input is 4 bytes, one per each channel. Through OpenGL normalization byte array can be used instead.
2. A bunch of functions was inlined. According to bzzt it leads to 17% faster filling of vertex buffers (important in terrain chunk creation).

### DIFF
--- a/components/esmterrain/storage.cpp
+++ b/components/esmterrain/storage.cpp
@@ -158,7 +158,7 @@ namespace ESMTerrain
         normal.normalize();
     }
 
-    void Storage::fixColour (osg::Vec4f& color, int cellX, int cellY, int col, int row, LandCache& cache)
+    void Storage::fixColour (osg::Vec4ub& color, int cellX, int cellY, int col, int row, LandCache& cache)
     {
         if (col == ESM::Land::LAND_SIZE-1)
         {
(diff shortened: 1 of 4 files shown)


AnyOldName3:
* Normalized integers gets a thumbs up from me.
* Inlining is good in cases where it makes things faster. https://github.com/OpenMW/openmw/pull/2234/commits/051df95738c032aab2e986083715d950aff46c9a#diff-a69f664064c442a68c60f6e1dbfd7ac4R37 might want a line break, though.
* The other optimisation will get a look at some point that isn't right now.

elsid:
`inline` for declaration in header while the difinition still in .cpp file doesn't do anything. This keyword is not about optimization (read [this](https://en.cppreference.com/w/cpp/language/inline)).

> Since this meaning of the keyword inline is non-binding, compilers are free to use inline substitution for any function that's not marked inline, and are free to generate function calls to any function marked inline. Those optimization choices do not change the rules regarding multiple definitions and shared statics listed above. 

Compilers smart enough to inline functions without this specifier (look at [this](https://godbolt.org/z/nu98kX) example). But compiler can't inline function into other function if they are in a different translation units.

Capostrophic:
>This keyword is not about optimization.

I mean... Obviously.

>Compilers smart enough to inline functions without this specifier

Not always.

Updated.

kcat:
> This keyword is not about optimization (read this).
>
> [...]
>
> Compilers smart enough to inline functions without this specifier (look at this example).

Yes and no. Compilers use their own heuristics to determine if a function can be inlined, basing it on various factors like how many different places a function is likely to be called (non-static and non-anonymous-namespaced functions may or may not be called from various other translations units the compiler can't currently see), and how much the resulting code size would grow (there's icache performance to worry about, and inlining functions can have stack size implications). While the compiler is free to inline or not inline regardless of the keyword, specifying the `inline` keyword does tell the compiler to try a little harder or be more forgiving with inlining it; it can inline something it otherwise wouldn't have on its own.

IMO, it's a mistake for the standard to use the `inline` keyword to simply mean "allow multiple definitions", because there is still other behavior and functionality tied to it that's still useful, and if you want one but not the other it becomes a problem.

AnyOldName3:
> inline for declaration in header while the difinition still in .cpp file doesn't do anything
> But compiler can't inline function into other function if they are in a different translation units.

I remember thinking it was weird that there were some `inline`d functions that were declared and defined separately, but I noticed that they were all `private`. That means that the only places they could be called were the header and the `.cpp` file and the definition would always be in the same translation unit as anywhere a call could happen.

If it's only a same-translation-unit rule (and not a declaration-must-include-the-definition rule) then it should work as intended as it was when the PR was opened.

elsid:
Ok, if `inline` this makes difference to that code we can use it. I just want to see that difference. What if `LandObject::getData` and `LandObject::getPlugin` definitions are moved into class definition without adding `inline` keyword. Would it be the same assembly to code with `inline` keyword. Also too many inlined code doesn't mean better performance.

If method is marked `inline` in declaration only in class definition, compiler doesn't bother yourself to generate assembly for this method if it is not called from same transation unit. If it is not maked it generates assembly (look [this](https://godbolt.org/z/Pkd3CA) example, gcc, clang do this, but not msvc). This mean compiler don't have to make optimization for all kind of calls (from this and from other translation unit), basically it is a purpose of `inline` keyword. Since class may have friends, any private method may be called from other transation unit, but it's not the case. So I'd recommend to return that commit as is. Sorry for disturbance.

Capostrophic:
Dropped blending map generation rewrite (will need a thorough look later), put inlining commit back.

psi29a:
Needs additional testing/review?

Capostrophic:
Probably not. It seems that point 1 looks good to AnyOldName3 while Chris and elsid came to an agreement that point 2 was fine.

PR #2236, opened 2019-03-09
Fix EAGAIN error leading to packets getting lost (bugs #4906, #4909)

Capostrophic:
The snippet from Chris [here](https://gitlab.com/OpenMW/openmw/issues/4909) appears to work and resolve these two issues that arised after FFmpeg 3+ transition.


akortunov:
The main idea:
1. Handle a case when there are several frames in the sound packet.
2. Do not lose a last sound packet (it caused issues for short sounds, especially for files with single sound packet).

psi29a:
Sweet, thanks.

PR #2237, opened 2019-03-09
[Regression] Improve sound decoding (bug #4906, bug #4909)

akortunov:
Fixes [bug #4906](https://gitlab.com/OpenMW/openmw/issues/4906) and [bug #4909](https://gitlab.com/OpenMW/openmw/issues/4909).

The main idea:
1. Handle a case when there are several frames in the sound packet.
2. Do not lose a last sound packet (it caused issues for short sounds, especially for files with single sound packet).

Thanks to @kcat.


akortunov:
It seems Capostophic opened an another PR earlier.

PR #2238, opened 2019-03-09
Include osgdb_freetype in builds with statically linked OSG

Capostrophic:
This makes sure OSG FreeType plugin is included in builds where OSG is linked statically, so the profiler font is loaded correctly. So basically this should only affect official Linux and macOS builds, and it should be up to the packager to include the plugin in a 0.45.0 build -- it shouldn't need cherry-picking.


terabyte25:
This seems to break android compilation currently.

PR #2241, opened 2019-03-10
 Use selected object local variables in the console (feature #3893) 

Capostrophic:
[Feature 3893](https://gitlab.com/OpenMW/openmw/issues/3893).

This allows to output and change local variables of the selected object without stating the object ID explicitly. The idea is to recover local variables of the script the reference potentially has, then the compiler and interpreter will handle the instance's local variables properly automatically.

Unfortunately the console now depends on script manager and class headers.

Also, I removed an extraneous space in "ID does not have a script" message showvars can output.


Capostrophic:
Is there something that requires attention?

psi29a:
None, just needed time to review.

PR #2246, opened 2019-03-11
Merge upstream osgQt changes

Capostrophic:
There uh, weren't actually that many of them, mainly due to how our osgQt is stripped down.
Upstream changes:
* Remove extraneous QGLWidget include
* Fix *a different* Qt5 threading issue (see [6e4de176](https://github.com/openscenegraph/osgQt/commit/6e4de176c7c2b4aeb4547fa5dc48d29d92b9d1ca))
* Make sure the window is valid before handling movement and resizing events.

My own changes:
* NULL is replaced with nullptr in the added lines (only in them).
* Removed unnecessary QGesture includes. They can't be used anyway due to how Aesylwinn purged the entirety of osgQt input stuff.

I think they should be fine.

Edit: apparently it doesn't fix **the** multithreaded model issue.

### DIFF
--- a/extern/osgQt/GraphicsWindowQt
+++ b/extern/osgQt/GraphicsWindowQt
@@ -14,8 +14,6 @@
 #ifndef OSGVIEWER_GRAPHICSWINDOWQT
 #define OSGVIEWER_GRAPHICSWINDOWQT
 
-#include <QGLWidget>
-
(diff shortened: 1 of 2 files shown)


Capostrophic:
I don't think an upstream sync would hurt anyway, would it?

PR #2247, opened 2019-03-11
Allow to override texture records in the game

akortunov:
Fixes [bug #4736](https://gitlab.com/OpenMW/openmw/issues/4736).

The main idea:
1. Currently every plugin has its own index set for LandTexture records.
2. If the new plugin declares a texture record with existing index and ID, consider it as an override for base record.
3. If either ID or index differs, consider record from plugin as a new record.

Notes: 
1. It does not fix the [#4904](https://gitlab.com/OpenMW/openmw/issues/4904) - it requires a lot of work on the editor.
2. It would be nice to have a cleaner solution, but it would require to rewrite a lot of legacy stuff. I'd prefer to do not touch it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
     Bug #4720: Inventory avatar has shield with two-handed weapon during [un]equipping animation
     Bug #4723: ResetActors command works incorrectly
     Bug #4745: Editor: Interior cell lighting field values are not displayed as colors
+    Bug #4736: LandTexture records overrides do not work
     Bug #4746: Non-solid player can't run or sneak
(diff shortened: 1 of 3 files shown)


psi29a:
Would it be best to do the re-write for 0.47 or wait until 1.0 is released?

akortunov:
> Would it be best to do the re-write for 0.47 or wait until 1.0 is released?

Do you mean the editor part, or a refactoring part?

psi29a:
refactoring part

akortunov:
> refactoring part

It is not a blocker.

psi29a:
I know it isn't, just wondering when it was best to... 0.47 or post 1.0 :)

PR #2249, opened 2019-03-13
Implement generic caching system

akortunov:
Originally the ObjectCache was designed to cache files. Internally it uses a map with std::string key with `fileName` name.
Later ObjectCache was re-used in other places, so we have to construct string keys on the fly even when we do not really need them.

Based on bzzt's changes.

The main idea:
1. Implement a ResourceManager with a generic key instead of string
2. Use the custom keys in the ChunkManager and the LandManager rather than strings, constructed from the same data.

As a result, operations with cache (e.g. search, insert) become faster.
A difference is not noticable in-game since the search already takes nano-seconds, but it would be nice to use it anyway since we access the cache thousands and thousands times during game session.

For example, in the master branch every access to the ChunkManager cache objects takes about 5ns, with this PR - less than 1ns.

Note: an implementation was moved to header files because of it is hard to split definition from declaration from template files. Actually, a cache implementation did not change - the only difference is which type of key is used.
So if an everyone knows how to make this PR cleaner, I am listening.


psi29a:
This is great work and thanks for generalizing it.

components/resource/objectcache.hpp
```diff
@@ -32,71 +36,161 @@ namespace osg
...
           * This would typically be called once per frame by applications which are doing database paging,
           * and need to prune objects that are no longer required.
           * The time used should be taken from the FrameStamp::getReferenceTime().*/
-        void updateTimeStampOfObjectsInCacheWithExternalReferences(double referenceTime);
+        void updateTimeStampOfObjectsInCacheWithExternalReferences(double referenceTime)
+        {
+            OpenThreads::ScopedLock<OpenThreads::Mutex> lock(_objectCacheMutex);
+            for(typename ObjectCacheMap::iterator itr=_objectCache.begin(); itr!=_objectCache.end(); ++itr)
+            {
+                // if the timestamp is yet to be initialized, update it as well
```
Capostrophic: If the movement of declaration to the definition is unavoidable, you should still copy-paste the whole comment as it is from the implementation file, bzzt has omitted the original comments for whatever reason so I adjusted and included all of them.

akortunov:
Any objections?

psi29a:
Last call..

akortunov:
@AnyOldName3, @Capostrophic, what do you think?

Capostrophic:
It's fine, I guess.

psi29a:
That sounded very enthusiastic. ;)

AnyOldName3:
Are we 100% sure that moving things from the `cpp` to the header without the `inline` hint is enough to inline them? I'd be surprised if all compilers always inlined stuff when possible.

kcat:
If it's a method defined in the class/struct, `inline` is implied.

AnyOldName3:
Yeah, I've found that in the spec now.

PR #2250, opened 2019-03-13
Do not swap buffers for non-exposed windows

akortunov:
Fixes [bug #4911](https://gitlab.com/OpenMW/openmw/issues/4911).

The main idea is to do not swap buffers, if the window was not exposed to the window managing system yet.

I took the code from [here](https://github.com/anormann1974/osgviewerQt5/blob/master/GraphicsWindowQt5.cpp) (it is basically a Qt5-version of OSG viewer, take a look at the `swapBuffersImplementation()`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
     Bug #4876: AI ratings handling inconsistencies
     Bug #4877: Startup script executes only on a new game start
     Bug #4888: Global variable stray explicit reference calls break script compilation
+    Bug #4911: Editor: QOpenGLContext::swapBuffers() warning with Qt5
     Feature #2229: Improve pathfinding AI
     Feature #3442: Default values for fallbacks from ini file
     Feature #3610: Option to invert X axis
--- a/extern/osgQt/GraphicsWindowQt.cpp
+++ b/extern/osgQt/GraphicsWindowQt.cpp
@@ -17,6 +17,7 @@
 #include <osgViewer/ViewerBase>
 #include <QInputEvent>
 #include <QPointer>
+#include <QWindow>
 
 #if (QT_VERSION>=QT_VERSION_CHECK(4, 6, 0))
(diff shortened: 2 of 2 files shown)


psi29a:
I'll take it, already played around with this awhile back.

Capostrophic:
Maybe submit a PR to upstream osgQt too?

akortunov:
> Maybe submit a PR to upstream osgQt too?

It is unclear in which form osgQt will exist: https://github.com/openscenegraph/osgQt/pull/22

So I see no reason to create a PR right now, but feel free to open it if you want.

AnyOldName3:
I'd definitely say opening a PR against openscenegraph/osgQt/master is a good idea, as it's still the main branch and it still has this issue.

There's one problem I see with this PR - I'm not sure Qt pre-5.7 actually uses `QOpenGLContext` under the hood, so the check might be on the wrong Qt version. It might not matter, though, as there's not a huge danger of not swapping invisible buffers, but in that case, there's not really any need for a version check at all.

By the way, I'm assuming I can take the line that says
> QOpenGLContext::swapBuffers() called with non-exposed window, behavior is undefined
off my list of OpenMW tasks now?

akortunov:
> there's not really any need for a version check at all.

I placed a version check because the QWindow was added in the Qt5, so I can not tell if the `isExposed()` check is present in Qt4.
But it seems there is an another issue - "include QWindow" should be under the version check too.

akortunov:
#2251.

> my list of OpenMW tasks

Can I take a look at this list?

AnyOldName3:
It's almost entirely things on the issue tracker already. It's just a bunch of things that I might take a crack at.

akortunov:
> things that I might take a crack at

I'd suggest you to take a look at AMD-specific artifacts with the water shader, or at least fill a bugreport.
It is a pretty serious issue, as for me.

PR #2251, opened 2019-03-13
Place QWindow include under Qt version check

akortunov:
QWindow class was added only in the Qt 5.0 release.

### DIFF
--- a/extern/osgQt/GraphicsWindowQt.cpp
+++ b/extern/osgQt/GraphicsWindowQt.cpp
@@ -17,7 +17,10 @@
 #include <osgViewer/ViewerBase>
 #include <QInputEvent>
 #include <QPointer>
+
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 0, 0))
 #include <QWindow>
+#endif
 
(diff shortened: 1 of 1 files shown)


extern/osgQt/GraphicsWindowQt.cpp
```diff
@@ -17,7 +17,10 @@
 #include <osgViewer/ViewerBase>
 #include <QInputEvent>
 #include <QPointer>
-#include <QWindow>
+
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 0, 0))
+# include <QWindow>
```
Capostrophic: why the whitespace?

akortunov: Take a look at lines 26-28.

Capostrophic: OSG code style not being consistent due to third party contributions doesn't mean OpenMW's code style can't be consistent, and it's not said anywhere in the code formatting conventions there should be whitespace to indent preprocessor directives after the hash symbol.

The following defines aren't used anyway which is why I purged them in a neighboring PR, and you can see later how *another* contributor indents branching directives with 4 spaces.

PR #2254, opened 2019-03-15
Print SDL version at startup

Capostrophic:
This is put in the logs and can be useful for figuring out if something is actually an SDL issue in user reports.

e.g.:
```
Using Cyrillic font encoding.
OSG version: 3.4.1
SDL version: 2.0.9
Loading settings file: ./settings-default.cfg
```

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -651,6 +651,9 @@ void OMW::Engine::go()
     assert (!mContentFiles.empty());
 
     Log(Debug::Info) << "OSG version: " << osgGetVersion();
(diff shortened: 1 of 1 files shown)


terabyte25:
[Perhaps include the revision as well?](https://wiki.libsdl.org/SDL_GetRevision)

Capostrophic:
I don't see the value. Individual revisions of SDL are virtually never used.

psi29a:
> [Perhaps include the revision as well?](https://wiki.libsdl.org/SDL_GetRevision)

Not necessary.

PR #2258, opened 2019-03-17
Update coverity scan branch

Capostrophic:
It's been a while since November, so it's a good time to trigger a coverity build.


psi29a:
It will fail. Builds take too long now.

Capostrophic:
Hm. And disabling building of *all* additional apps but openmw will not work?

psi29a:
We've already done that I think, I'll check again.

akortunov:
> We've already done that I think, I'll check again.

Last time I checked, we enabled analyze for the OpenMW-CS as well.
From what I can tell, earlier we did not build the editor as well.

PR #2259, opened 2019-03-17
Update optimizer with upstream improvements

Capostrophic:
1. Typo fixes
2. mergeGroup method was refactored to avoid complexity issues with removing children of groups with a large number of children.
3. mergeGeometry method was restructured to be more readable.
4. Attributes of arrays of more types can be merged (though the new types don't seem to be used in openmw right now, they can get used at some point).

Ironically upstream optimizer could also benefit from changes in OpenMW's version of it.


psi29a:
Could always upstream it ;)

PR #2263, opened 2019-03-18
Do not recreate shaders to update transparency

akortunov:
Based on bzzt's changes.
Attempt to add a color uniform directly instead of doing the same thing via shader visitor to handle alpha level.
May not be a good idea if other color modes should be used sometimes, so requesting a review from @AnyOldName3.

### DIFF
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -1798,17 +1798,13 @@ namespace MWRender
                 material->setDiffuse(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,alpha));
                 material->setAmbient(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
                 stateset->setAttributeAndModes(material, osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
-
+                stateset->addUniform(new osg::Uniform("colorMode", 0), osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
                 mObjectRoot->setStateSet(stateset);
-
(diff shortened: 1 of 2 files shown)


AnyOldName3:
A colour mode of zero basically just means that no part of the material tracks the vertex colour. Is that definitely what you're trying to do?

akortunov:
> Is that definitely what you're trying to do?

I just try to determine if these bzzt's changes are reliable, or we should discard them.

Capostrophic:
The point is to make the material use the set transparency. Since diffuse/ambient/emission colors are being overridden anyway - as per the lines and fixme above - this change makes sense to me in scope of what bzzt said about shader visitor (added through recreateShaders) previously handling adding the uniform.

So basically LGTM.

akortunov:
> Though if we're going there, it makes sense to pick the neighboring commit that fixes the data race of only changing the alpha if the material already exists too.

I am not sure about this commit. If there are really two threads work with the stateset in the same time, it may just hide an issue instead of fixing it.

AnyOldName3:
> The point is to make the material use the set transparency. Since diffuse/ambient/emission colors are being overridden anyway - as per the lines and fixme above - this change makes sense to me in scope of what bzzt said about shader visitor (added through recreateShaders) previously handling adding the uniform.

The objection to setting the colour mode uniform was that it may well have existed already in a parent stateset. However, I'm now pretty sure that there'd be no need to have regenerated the shaders if there was no chance of the colour mode being changed back when it was a define, so setting it as a uniform now is probably the right thing to do.

psi29a:
Leaving this to @AnyOldName3 for him to review and merge at this desecration. :)

PR #2266, opened 2019-03-18
Vsync tweaks

Capostrophic:
* Try to use adaptive vsync if available.
* Don't use vsync if unavailable.

Using adaptive vsync means that the framerate won't be capped to 30 or 15 or whatever if it's too low for the user's display refresh rate, leading to lower input delay and higher framerate, but it requires some OpenGL extension not all cards have (it was introduced to Nvidia and AMD drivers in late 2011).

Not using vsync if unavailable may "fix" visual issues with Intel cards if they really don't support it. I doubt that it will, though.

### DIFF
--- a/components/sdlutil/sdlgraphicswindow.cpp
+++ b/components/sdlutil/sdlgraphicswindow.cpp
@@ -113,7 +113,7 @@ void GraphicsWindowSDL2::init()
         return;
     }
 
-    SDL_GL_SetSwapInterval(_traits->vsync ? 1 : 0);
+    setSwapInterval(_traits->vsync);
 
     SDL_GL_MakeCurrent(oldWin, oldCtx);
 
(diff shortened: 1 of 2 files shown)


components/sdlutil/sdlgraphicswindow.cpp
```diff
@@ -194,7 +211,24 @@ void GraphicsWindowSDL2::setSyncToVBlank(bool on)
...
+        if (!vsynccheck)
+        {
+            OSG_NOTICE << "Vertical synchronization unsupported, disabling" << std::endl;
+            SDL_GL_SetSwapInterval(0);
+        }
+    }
+    else
+    {
+        SDL_GL_SetSwapInterval(0);
+    }
```
Mingun: This code duplication can be extracted to small function

components/sdlutil/sdlgraphicswindow.hpp
```diff
@@ -80,6 +80,9 @@ class GraphicsWindowSDL2 : public osgViewer::GraphicsWindow
 
         SDL_Window *mWindow;
     };
+
+private:
+    void setSwapInterval(bool enable);
```
elsid: Why don't place this function into anonymous namespace in .cpp file? It don't use any of this class fields and is not used outside `sdlgraphicswindow.cpp` file.

wareya: The function "belongs" to the SDL window. Global state and stuff.

PR #2267, opened 2019-03-18
Stack items with and without owner if item holder is owner (bug #1933)

elsid:
Holder is an actor in which inventory or container item is placed. If player pick and item or item is sold to NPC, or stolen from NPC, then item owner is set to empty. If actor has other item in the same container and with same parameters but has different owner (actor itself) they can be stacked. At least I don't see why they can't be stacked.

This pr fixes [1933](https://gitlab.com/OpenMW/openmw/issues/1933) and doesn't break [1953](https://gitlab.com/OpenMW/openmw/issues/1953). Sold item is placed into merchant inventory and can be stolen from it. All sold apparel can be bought back.

Also removed some unused code.

### DIFF
--- a/apps/openmw/mwworld/containerstore.cpp
+++ b/apps/openmw/mwworld/containerstore.cpp
@@ -58,6 +58,13 @@ namespace
 
         return MWWorld::Ptr();
     }
+
+    bool shouldStackByHolder(const MWWorld::ConstPtr& withoutOwner, const MWWorld::ConstPtr& withOwner,
+        const MWWorld::ConstPtr& holder)
+    {
+        return withoutOwner.getCellRef().getOwner().empty()
+            && withOwner.getCellRef().getOwner() == holder.getCellRef().getRefId();
+    }
 }
 
 template<typename T>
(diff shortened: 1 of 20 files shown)


apps/openmw/mwworld/containerstore.cpp
```diff
@@ -241,7 +248,12 @@ bool MWWorld::ContainerStore::stacks(const ConstPtr& ptr1, const ConstPtr& ptr2)
     }
 
     return ptr1 != ptr2 // an item never stacks onto itself
-        && ptr1.getCellRef().getOwner() == ptr2.getCellRef().getOwner()
+        && (
+            ptr1.getCellRef().getOwner() == ptr2.getCellRef().getOwner()
```
akortunov: The `ContainerStore::stacks` checks if two items can be combined to the one CellRef with increased count of objects.
If these two items have different owners, one of them will be lost, but it does not matter with our current implementation (see #1941).
Issues will arise only if you aimed to keep original item owners to track stolen items, as you wanted.
In this case every item with different owner should be in the separate stack in the inventory, and we should combine them only in ItemStacks in item models. I tried to do it, but it does not work well for stackable eqipped items (such as throwing weapons and ammo).

akortunov: > Will it change save format?

Yes, it will.

> But still we do Morrowind.

Then we should decide if we want to keep an original system with its quirks or if we want to replace it later.
In the first case there is a little we can do here:
1. Merge stacks with different owners as in #1941.
2. Take in account Soul field when tracking stolen items (requires a savegame format change).
3. Add a workaround for evidence chests ([bug #3609](https://gitlab.com/OpenMW/openmw/issues/3609)).

Keeping the Owner field during item transfers is not a reliable approach here.

elsid:
We should keep original system quirks if it’s a part of gameplay. Only option in launcher should change it. Stolen flag is fine and it allows to keep original behavior. Players can complain a lot, but if you change something unconditionally, there are always will be another complainers to return it back.

I’m closing this pr because it doesn’t solve all stacking problems with given approach.

PR #2270, opened 2019-03-20
Fixes for werewolf animations

akortunov:
1. Fixes a 0.46 regression when werewolves sometimes used animations for bipedal actors.
The main idea is to set NPC type properly from beginning, rather than starting to construct a normal NPC and then rebuild him to vampire or werewolf.
2. Fixes [bug #4922](https://gitlab.com/OpenMW/openmw/issues/4922).
Currently we use the forceUpdate() when we rebuild actor from scratch in these cases:
2.1. For player object when we change the view mode
2.2. For all NPCs when we change the NPC type (e.g. from turn him from normal NPC to vampire or werewolf).
In the first case we do not allow to switch mode during attack, so this case is irrelevant here, so in theory it should be safe to reset flags which are responsible for attack state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
     Bug #4896: Title screen music doesn't loop
     Bug #4911: Editor: QOpenGLContext::swapBuffers() warning with Qt5
     Bug #4916: Specular power (shininess) material parameter is ignored when shaders are used.
+    Bug #4922: Werewolves can not attack if the transformation happens during attack
     Feature #2229: Improve pathfinding AI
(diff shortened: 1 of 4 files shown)


psi29a:
At some point, we need to make things like Werewolf and Vampire things that are dependent per game not something hard-coded in the engine itself. Could you create a de-hardcoding ticket to handle this case and link this PR for reference?

akortunov:
>  Could you create a de-hardcoding ticket to handle this case and link this PR for reference?

I see no gain from such ticket in the foreseen future. There is a need for a fully working editor and a forum discussion about design of such feature first.

PR #2271, opened 2019-03-20
Minor NpcAnimation cleanup

Capostrophic:
1. Replace esmstore iterator-based loops with range-based loops which look nicer.
2. Replace one instance of find-assign and one instance of pair insertion with emplace.
3. Missing werewolf head and hair records no longer cause an exception and unexpected behavior (like saves not loading), and are handled practically the same way as normal NPC heads.
4. Use a more clean way for detecting 1st person body part suffix.
5. Use an initializer list for initializing body part map.
6. Fixed a typo in a condition (bitwise AND was used instead of logical AND, which blocks shortcutting)


psi29a:
Very nice, I like it. It a lot more readable now.

PR #2272, opened 2019-03-20
Do not block loading while compiling composite maps

akortunov:
Currently we have two queues to compile composite maps in the draw thread:
1. mCompileSet - for low priority maps. Compile while some time left to avoid frame drops. If there is no time left, continue work in the next frame.
2. mImmediateCompileSet - for high priority maps. Compile all of them even if there will be a frame drop.

We manage these queues in the loading thread, so we use a mutex to block access to queues when we work with them.
Since compiling takes some time (about 1ms per map), the draw thread holds the mutex for a while.
As a result, if the background thread wants to add some items to the compile queue (e.g. during cell loading), it should wait for mutex from the draw thread.

The main idea here - since compiling itself does not work with queues, we can release the mutex while compiling (but not when we read data from queues), so the loading thread can push items to queue and continue execution.

If the loading thread moves current map from one queue to another via `setImmediate()`, the draw thread may try to compile the map in the second time, but it will compile only non-compiled drawables in this map, if there are any.

So this PR should increase the loading speed a bit.

Requesting a review from @elsid since he has more experience with threads sync than me - maybe there is a more clean way to do it.

### DIFF
--- a/components/terrain/compositemaprenderer.cpp
+++ b/components/terrain/compositemaprenderer.cpp
@@ -59,26 +59,29 @@ void CompositeMapRenderer::drawImplementation(osg::RenderInfo &renderInfo) const
 
     while (!mImmediateCompileSet.empty())
     {
-        CompositeMap* node = *mImmediateCompileSet.begin();
+        osg::ref_ptr<CompositeMap> node = *mImmediateCompileSet.begin();
         mCompiled.insert(node);
+        mImmediateCompileSet.erase(node);
 
+        mMutex.unlock();
         compile(*node, renderInfo, nullptr);
-
-        mImmediateCompileSet.erase(mImmediateCompileSet.begin());
+        mMutex.lock();
     }
 
     double timeLeft = availableTime;
 
     while (!mCompileSet.empty() && timeLeft > 0)
     {
-        CompositeMap* node = *mCompileSet.begin();
+        osg::ref_ptr<CompositeMap> node = *mCompileSet.begin();
 
+        mMutex.unlock();
         compile(*node, renderInfo, &timeLeft);
+        mMutex.lock();
 
         if (node->mCompiled >= node->mDrawables.size())
         {
             mCompiled.insert(node);
-            mCompileSet.erase(mCompileSet.begin());
+            mCompileSet.erase(node);
         }
     }
     mTimer.setStartTick();
--- a/components/terrain/compositemaprenderer.cpp
+++ b/components/terrain/compositemaprenderer.cpp
@@ -70,14 +70,17 @@ void CompositeMapRenderer::drawImplementation(osg::RenderInfo &renderInfo) const
     while (!mCompileSet.empty() && timeLeft > 0)
     {
         osg::ref_ptr<CompositeMap> node = *mCompileSet.begin();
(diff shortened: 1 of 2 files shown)


components/terrain/compositemaprenderer.cpp
```diff
@@ -59,26 +59,29 @@ void CompositeMapRenderer::drawImplementation(osg::RenderInfo &renderInfo) const
...
 
+        mMutex.unlock();
         compile(*node, renderInfo, &timeLeft);
+        mMutex.lock();
 
         if (node->mCompiled >= node->mDrawables.size())
         {
             mCompiled.insert(node);
-            mCompileSet.erase(mCompileSet.begin());
+            mCompileSet.erase(node);
```
elsid: If we have only one `osg::GraphicsThread` to call this method, then erase here is fine. But if we have more, then two threads can start compiling same composite map. I'm not sure do they need to do it for each `osg::GraphicsContext` or not.

components/terrain/compositemaprenderer.cpp
```diff
@@ -59,26 +59,29 @@ void CompositeMapRenderer::drawImplementation(osg::RenderInfo &renderInfo) const
 
     while (!mImmediateCompileSet.empty())
     {
-        CompositeMap* node = *mImmediateCompileSet.begin();
+        osg::ref_ptr<CompositeMap> node = *mImmediateCompileSet.begin();
         mCompiled.insert(node);
```
elsid: This insert seems to be useless. `mCompiled` is not used anywhere for reading.

akortunov: scrawl did not leave any comments, but it can be aimed just to keep references for handled nodes. Or it indeed can be redundant now.

components/terrain/compositemaprenderer.cpp
akortunov: @AnyOldName3, do we have several graphics threads?

akortunov: As a possible solution, we can try to remove composite map from the `mCompileSet` immediately as a draw thread starts to compile it. If we did not compile it fully (because of not enough time), place it back after processing. In this case graphics threads should not compile maps which we already compile in other threads.

akortunov:
Update a bit.
1. Remove composite maps from `mCompileSet` immediately after we start to compile them, so different draw threads will not try to compile them too.
If we do not have enought time to compile it fully, place it back to queue, so we can continue processing in the next frame or in other threads.
2. Remove the `mCompiled` set - we do not read data from it, and composite maps are already stored as user data in terrain chunks and use ref_ptr's. So I see no reason to fill the set to just clear it in the beginning of next update. It even is not under lock.

psi29a:
What exactly do we win here in terms of performance and stability?

akortunov:
> What exactly do we win here in terms of performance and stability?

A slightly better loading speed.

components/terrain/compositemaprenderer.cpp
AnyOldName3: We use the DrawThreadPerContext threading model, so as the engine only ever has one context active at a time, we've only got one graphics thread (assuming that's how it works).

PR #2275, opened 2019-03-21
Use auto for map value_type in range-based for loops

elsid:
To avoid implicit call of copy constructor for `pair<const K, V>` to `pair<K, V>` conversion.


elsid:
macos build is failed because of some network problems, need retry.

PR #2277, opened 2019-03-22
Camera-related fixes

akortunov:
Let's discuss an another set of changes from bzzt branch.

Summary of changes:
1. Use result of first traversal for other traversals with the same view point
2. Copy preloaded terrain views from the preloading thread to the mViewDataMap in the main thread (there was a relevant TODO in the OpenMW code).
3. Take the view point from main camera for reflection and reflaction cameras, so we will not have to use separate terrain LOD's for them.
4. Remove the DefaultView since we do not need it anymore.

As a result, cell transitions should be a bit faster.

Can we use these changes, or they do not worth it?
Which kind of clarification from bzzt we need here?

Casting @AnyOldName3 here.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -312,7 +312,6 @@ namespace MWRender
         else
             mTerrain.reset(new Terrain::TerrainGrid(sceneRoot, mRootNode, mResourceSystem, mTerrainStorage, Mask_Terrain, Mask_PreCompile, Mask_Debug));
 
-        mTerrain->setDefaultViewer(mViewer->getCamera());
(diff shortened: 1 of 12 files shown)


akortunov:
If we will have these changes, we can fix the [bug #4897](https://gitlab.com/OpenMW/openmw/issues/4897) by using the view point instead of eye point do determine if we need to freeze rain.

psi29a:
Ready?

akortunov:
I still get no response from @AnyOldName3.

akortunov:
Since there is no response in three weeks, I suggest to merge this PR and allow users to test it.

psi29a:
I'll bite this bullet.

PR #2279, opened 2019-03-22
Optimize map window update

akortunov:
Based on bzzt's branch.
Summary of changes:
1. Explore cells immediately, so we do not need to update map window when it is not visible.
Originally this delay was added to fix [bug #772](https://gitlab.com/OpenMW/openmw/issues/772) on OGRE version, but it seems this bug is not present with OSG (at least I failed to reproduce it with strongholds).
2. Instead of storing data in 4 differect vectors, define a struct and store a vector of structs.
3. Do not store both sets of new and old textures, replace old textures via `reset()` instead.
4. Update fog texture only if it was changed.

As a result, cell transitions work a bit faster (only fog textures update takes about 50 microseconds less time).

### DIFF
--- a/apps/openmw/mwgui/mapwindow.cpp
+++ b/apps/openmw/mwgui/mapwindow.cpp
@@ -213,8 +213,7 @@ namespace MWGui
                 map->setNeedMouseFocus(false);
                 fog->setNeedMouseFocus(false);
 
-                mMapWidgets.push_back(map);
-                mFogWidgets.push_back(fog);
+                mMaps.emplace_back(map, fog);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


apps/openmw/mwgui/mapwindow.cpp
```diff
@@ -213,8 +213,7 @@ namespace MWGui
                 map->setNeedMouseFocus(false);
                 fog->setNeedMouseFocus(false);
 
-                mMapWidgets.push_back(map);
-                mFogWidgets.push_back(fog);
+                mMaps.push_back(MapEntry(map, fog));
```
Capostrophic: *In emplace_back I trust.*

PR #2281, opened 2019-03-23
Minor water shader fixes

akortunov:
Based on bzzt's work.
Summary of changes:
1. Cleanup water shader a bit - do not use division by 1000.
2. Use the Incremental Compile Operation for water shader and RTT (now we seems to use it only for water geometry)

### DIFF
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -255,6 +255,7 @@ public:
         attach(osg::Camera::COLOR_BUFFER, mRefractionTexture);
 
         mRefractionDepthTexture = new osg::Texture2D;
+        mRefractionDepthTexture->setTextureSize(rttSize, rttSize);
         mRefractionDepthTexture->setSourceFormat(GL_DEPTH_COMPONENT);
         mRefractionDepthTexture->setInternalFormat(GL_DEPTH_COMPONENT24);
         mRefractionDepthTexture->setWrap(osg::Texture::WRAP_S, osg::Texture::CLAMP_TO_EDGE);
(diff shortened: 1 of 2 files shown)


AnyOldName3:
What about the water shader is getting incrementally compiled with this?

Also, I'm not 100% sure there's a good reason to set a minimum alpha value - when looking at water there *are* bits that are pretty clear.

akortunov:
> What about the water shader is getting incrementally compiled with this?

A response from bzzt:

> shader and textures will pre compile by ICO i measured by 1st togglewater framdedrop

Probably you will manage to get more info from him since you know what exactly to ask.

> Also, I'm not 100% sure there's a good reason to set a minimum alpha value - when looking at water there are bits that are pretty clear.

A related thing - what do you think about his changes in water transparency (a previous discussion in his MR)?

akortunov:
Conversation on GitLab:

> frame 1 = shader.. frame 2 = refracttexture.. frame 3 = reflecttexture etcc..

> If we need water in the first frame, does this delay it until the ICOs are done, or is there pop-in? I've not meddled with ICOs yet, so I don't know what happens if their results are needed before they're done

> than it will sync immedatly as if there is no ICO and a later ICO will do nohting

AnyOldName3:
I still had a maybe-objection.

Capostrophic:
About what specifically?

xyzz:
Seems to have broken shader water on android:

```
05-11 21:03:30.808 29234 29288 I Adreno  : ERROR: 0:42: 'const' :  non-matching types for const initializer  
05-11 21:03:30.808 29234 29288 I Adreno  : ERROR: 0:72: 'const' :  non-matching types for const initializer  
05-11 21:03:30.808 29234 29288 I Adreno  : ERROR: 2 compilation errors.  No code generated.
```

I think it wants an explicit `.0` at the end of const floats... (https://www.khronos.org/opengl/wiki/GLSL_:_common_mistakes#NVIDIA_and_Types)

AnyOldName3:
> To what specifically?

No idea as I can't see anything questionable in the shader changes. Maybe it got fixed with a force-push.

Capostrophic:
Right, akortunov removed the transparency changes after you objected to them.

PR #2282, opened 2019-03-23
Replace black blendmap layer with blending over black

Capostrophic:
Another terrain texture blending change from bzzt.

wareya has replaced "default texture" base layer with a black texture base layer earlier, but truth is, such a base layer is not necessary at all, and the layers can be simply blended over black. This further simplifies the terrain texture blending and gets rid of some awkward workarounds in order to keep engine from thinking the black layer is a real blendmap layer.

Needs review.

### DIFF
--- a/components/esmterrain/storage.cpp
+++ b/components/esmterrain/storage.cpp
@@ -352,11 +352,6 @@ namespace ESMTerrain
 
     std::string Storage::getTextureName(UniqueTextureId id)
     {
-        // Goes under used terrain blend transitions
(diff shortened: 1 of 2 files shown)


akortunov:
Does not compile in Windows.

akortunov:
On Linux as well:
```
{anonymous}::BlendFunc::BlendFunc()» is private within this context
```

Capostrophic:
Thought I had compiled it before pushing. Copypaste mistake.

Capostrophic:
Let's do it again.

PR #2284, opened 2019-03-24
[WIP] Make OpenMW version information easily available in-game (feature #2780)

Capostrophic:
[Feature 2780](https://gitlab.com/OpenMW/openmw/issues/2780).

The request is not for a script command that returns version information specifically (which we shouldn't need to implement anyway), but for a way to easily find openmw version information in the game in general. So the elegant solution I propose is to always show the version text normally available in the "main" main menu in the in-game main menu. Morrowind does this exact thing (note 1.6.1820 appearing in the lower left corner) and it works fine there.

This version text can be copied and pasted as usual.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
     Bug #4916: Specular power (shininess) material parameter is ignored when shaders are used.
     Bug #4922: Werewolves can not attack if the transformation happens during attack
     Feature #2229: Improve pathfinding AI
+    Feature #2780: Make OpenMW version information easily available in-game
     Feature #3442: Default values for fallbacks from ini file
     Feature #3610: Option to invert X axis
     Feature #3893: Implicit target for "set" function in console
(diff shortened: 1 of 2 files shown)


elsid:
Approach seems to be fine, but I'm not sure that visualization is good enough. Can you move text to avoid gui items overlapping or add more contrast?
openmw_version

Capostrophic:
How about this?
Version text

akortunov:
Does not look good IMO.The text just is too large and overlaps with UI.

Probably just show the version and revision hash under fatigue bar will be enough.

psi29a:
Instead of in-game, can this information be put into the console (as apart of the background)?

akortunov:
> Instead of in-game, can this information be put into the console (as apart of the background)?

An interesting idea, but it will not look good with either low or high background transparency.

psi29a:
Worth a try anyway? At the very least it should be there at the main menu.

It should also, eventually, be apart of our new lua system ;)

Capostrophic:
Needs more thinking through.

PR #2286, opened 2019-03-25
Do not re-render maps for all active cells during cell transitions

akortunov:
Based on bzzt's changes.

Currently during cell transitions we try to re-render local maps for all active cells, what is a bit redundant since we need to load local maps only for new cells.

With this PR we will re-render only cells, visible after transition.

### DIFF
--- a/apps/openmw/mwrender/localmap.cpp
+++ b/apps/openmw/mwrender/localmap.cpp
@@ -256,36 +256,30 @@ bool needUpdate(std::set<std::pair<int, int> >& renderedGrid, std::set<std::pair
     return false;
 }
 
-void LocalMap::requestMap(std::set<const MWWorld::CellStore*> cells)
+void LocalMap::requestMap(const MWWorld::CellStore* cell)
 {
-    std::set<std::pair<int, int> > grid;
-    for (const MWWorld::CellStore* cell : cells)
+    if (cell->isExterior())
     {
-        if (cell->isExterior())
-            grid.insert(std::make_pair(cell->getCell()->getGridX(), cell->getCell()->getGridY()));
-    }
+        int cellX = cell->getCell()->getGridX();
+        int cellY = cell->getCell()->getGridY();
 
-    for (const MWWorld::CellStore* cell : cells)
-    {
-        if (cell->isExterior())
-        {
-            int cellX = cell->getCell()->getGridX();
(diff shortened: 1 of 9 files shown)


apps/openmw/mwrender/localmap.cpp
```diff
@@ -256,35 +256,32 @@ bool needUpdate(std::set<std::pair<int, int> >& renderedGrid, std::set<std::pair
...
+    else
+        requestInteriorMap(cell);
+}
+
+void LocalMap::addCell(MWWorld::CellStore *cell)
+{
+    if (cell->isExterior())
+    {
+        std::pair<int, int> coords = std::make_pair(cell->getCell()->getGridX(), cell->getCell()->getGridY());
+        mCurrentGrid.insert(coords);
```
Capostrophic:
```cpp
if (cell->isExterior())
    mCurrentGrid.emplace(cell->getCell()->getGridX(), cell->getCell()->getGridY());
```
:smile:

akortunov: ?

terabyte25: Emplace over insert I guess is what he means.

akortunov:
Any thoughts?

AnyOldName3:
Aren't we re-rendering maps on purpose in case anything has been enabled/disabled?

akortunov:
> Aren't we re-rendering maps on purpose in case anything has been enabled/disabled?

Well, then we do it wrong since objects already will appear/disappear on the map only if the modified cell is unloaded and loaded again.

akortunov:
All right, from what I can tell, the main idea is actually to do not re-render the fog of war for everything every time we change the current cell.

akortunov:
A response from bzzt:

> a detailed epxlanation
ISSUE WITH CURENT IMPLEMEMTAION
render 9X9 MAPS at same time htat new cells are load = BIG FRAMEDDROP in drawthread
MY IMPLEMEMTATION
ender 1x MAP where player is
other MAPS rendered ONLY IF any map window will reveal it by a viewscrolling offset =i have a huge window OR i will mousedrag view to a far area
result = A LOT LESS maps rendering AND spread over more frames becaus a map window revealing happen at in different time than cells loading

PR #2287, opened 2019-03-26
Update paths.rst - Windows paths, shell friendly paths

alexkokkinos:
This is a change to the Paths documentation to generally make paths easier to copy/paste and use within a terminal or the Go menu.

If this sort of change makes things too weird/complicated or is otherwise undesired, my apologies! In particular throwing RST tables around can make this harder to edit in the future, but I think the information needed tables when presented this way.

I'm also worried that this might make the docs more complicated for non-technical users, but I also thought it was something of a problem that users couldn't just copy and paste directories.

- Adds path expansion based paths for Windows for easier copying and pasting
- Edited Windows paths with proper case (even though Windows is case-insensitive)
- Escapes Application Support in Mac to make the paths terminal and Finder Go menu friendly
- Places paths in tables

I have done zero testing beyond previewing the results in GitHub. I'm not incredibly familiar with how ReadTheDocs generates its pages. I did some reading but in the end didn't get my environment fully set up.

### DIFF
--- a/docs/source/reference/modding/paths.rst
+++ b/docs/source/reference/modding/paths.rst
@@ -9,20 +9,50 @@ The following describes the locations for the various OpenMW file paths for diff
 Configuration files and log files
 ---------------------------------
 
-:Linux:		``$HOME/.config/openmw``
-:Mac:		``$HOME/Library/Preferences/openmw``
-:Windows:	``C:\Users\Username\Documents\my games\openmw``
++--------------+------------------------------------------------------------------+
+| OS           | Location                                                         |
++==============+==================================================================+
+| Linux        | ``$HOME/.config/openmw``                                         |
++--------------+------------------------------------------------------------------+
+| Mac          | ``$HOME/Library/Preferences/openmw``                             |
++--------------+---------------+--------------------------------------------------+
+| Windows      | File Explorer | ``%USERPROFILE%\Documents\My Games\OpenMW``      |
+|              |               |                                                  |
+|              | PowerShell    | ``"$env:USERPROFILE\Documents\My Games\OpenMW"`` |
+|              |               |                                                  |
+|              | Example       | ``C:\Users\Username\Documents\My Games\OpenMW``  |
++--------------+---------------+--------------------------------------------------+
 
 Savegames
 ---------
 
-:Linux:		``$HOME/.local/share/openmw/saves``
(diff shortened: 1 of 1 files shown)


psi29a:
RTD uses sphinx to generate its documentation (ReST). This is slightly different than Markdown, at least the variant that GitHub uses. We'll need to test this first on RTD to make sure it holds up.

alexkokkinos:
@psi29a Would you like me to help test this out? Like I mentioned, I didn't spend a ton of time trying to get anything setup yet, but I could.

I still wonder how helpful this sort of PR is, but if you like the idea I'm glad to help test/edit it if needed.

psi29a:
Just did a check, sphinx rendered the tables just fine. Looks good and readable. Thanks :)

psi29a:
https://openmw.readthedocs.io/en/latest/reference/modding/paths.html

AnyOldName3:
In future, could people please ask the Windows developers for a review before merging misleading or wrong things?

While the original version before this PR was wrong, it's *still* wrong after this PR, but because it's been changed recently, it now looks like we think it's right.

Before this PR, the documentation was accurate when:
* The user was using Windows 8.1 or earlier and had not changed their documents directory location or user area location from the default.
* The user was using Windows 10, and changed from the default behaviour of keeping Documents within OneDrive to keeping them in their user area, but had not further modified the location of documents or their user area.

This PR makes the documentation correct when:
* Either of the two above cases applies.
* Where the user had moved their user area, but left everything within it as the default.

From my experience, it's much more likely to see someone move their documents directory out of their user area than move their whole user area. Also, given that the default in Windows 10 is to put documents in `<user area>\OneDrive\Documents` I really think we should support that.

The correct PowerShell command to get the OpenMW directory is
```powershell
Join-Path ([environment]::GetFolderPath("mydocuments")) "My Games\OpenMW"
```
The correct thing to put in Explorer is just
```
Documents\My Games\OpenMW
```

AnyOldName3:
If you want to use `cmd` instead of PowerShell, you either need to add an inline PowerShell call or use `reg query` and then parse the result.

alexkokkinos:
Thanks for letting us know of this case, I wasn’t aware of it at all especially since I’ve been so dismissive of OneDrive setup prompts. I didn’t realize Windows modified the path that much.

In my original PR I expressed some

AnyOldName3:
I've opened a new MR on GitLab to change it to what should work.

PR #2288, opened 2019-03-26
Validate spell effect argument (bug #4927)

Capostrophic:
[Bug 4927](https://gitlab.com/OpenMW/openmw/issues/4927)

Drop invalid arguments for spell effects that don't really need them to avoid triggering an exception upon effect key generation.


akortunov:
I wish we had a way to do it without additional includes.
Can we add this validation when we read data from ESM/ESP file (probably with warning) instead?

PR #2289, opened 2019-03-28
mingw: add fixes for MinGW static building

the-eater:
This PR does the following:

- Fixes some paths that are cased wrongly.
- Add an `OPENMW_STATIC` flag, which will add some extra libraries to link with openmw, which are needed to build staticly
- Add an  `OPENMW_NO_PKGCONFIG` to not use pkg config to resolve libraries or cflags
- Add an `OPENMW_CXX_FLAGS` to add CXX_FLAGS if needed only for openmw

### DIFF
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -119,6 +119,7 @@ include_directories(
 )
 
 target_link_libraries(openmw
+    ${OPENMW_CXX_FLAGS}
     ${OSG_LIBRARIES}
     ${OPENTHREADS_LIBRARIES}
     ${OSGPARTICLE_LIBRARIES}
@@ -141,7 +142,33 @@ target_link_libraries(openmw
     components
 )
 
-if (ANDROID)
+if (OPENMW_STATIC AND NOT OPENMW_NO_PKGCONFIG AND (NOT WIN32 OR MINGW))
+    find_package(PkgConfig)
+
+    pkg_check_modules(openmw_pkgconfig REQUIRED
+        libjpeg
+        openal
+        freetype2
+        sdl2
+        gl
+        libavcodec
+        libavdevice
+        libavfilter
+        libavformat
+        libavresample
+        libavutil
+    )
+
+    target_link_libraries(openmw
+        ${openmw_pkgconfig_STATIC_LIBRARIES}
+    )
+
+    target_compile_options(openmw PUBLIC
+        ${openmw_pkgconfig_STATIC_CFLAGS_OTHER}
+    )
+endif()
+
+if (ANDROID OR OPENMW_STATIC)
     set (OSG_PLUGINS
         -Wl,--whole-archive
     )
@@ -154,15 +181,20 @@ if (ANDROID)
     )
 
     target_link_libraries(openmw
-        EGL
-        android
-        log
         dl
         z
         ${OPENSCENEGRAPH_LIBRARIES}
         freetype
         jpeg
-    	png
+        png
+    )
+endif (ANDROID OR OPENMW_STATIC)
+
+if (ANDROID)
+    target_link_libraries(openmw
+        EGL
+        android
+        log
     )
 endif (ANDROID)
 
--- a/cmake/FindOSGPlugins.cmake
+++ b/cmake/FindOSGPlugins.cmake
@@ -28,10 +28,14 @@ foreach(_library ${OSGPlugins_FIND_COMPONENTS})
     set(_component OSGPlugins_${_library})
 
     set(${_library_uc}_DIR ${OSGPlugins_LIB_DIR}) # to help function osg_find_library
-    set(_saved_lib_prefix ${CMAKE_FIND_LIBRARY_PREFIXES}) # save CMAKE_FIND_LIBRARY_PREFIXES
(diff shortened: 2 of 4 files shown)


the-eater:
One does wonder what happened with `boost_zlib`

apps/openmw/CMakeLists.txt
```diff
@@ -141,7 +142,33 @@ target_link_libraries(openmw
     components
 )
 
-if (ANDROID)
+if (OPENMW_STATIC AND NOT OPENMW_NO_PKGCONFIG AND (NOT WIN32 OR MINGW))
```
kcat: It might be a better idea to instead redirect pkg-config to use the right locations. For cross-compiling, I have this in the toolchain file:
```
# set env vars so that pkg-config will look in the appropriate directory for
# .pc files (as there seems to be no way to force using ${HOST}-pkg-config)
set(ENV{PKG_CONFIG_LIBDIR} "${CMAKE_INSTALL_PREFIX}/lib/pkgconfig")
set(ENV{PKG_CONFIG_PATH} "")
```
(where `CMAKE_INSTALL_PREFIX` is the cross-compiler's directory for headers and libs). That way, pkg-config will find the .pc files for packages installed for mingw.

cmake/FindOSGPlugins.cmake
```diff
@@ -28,10 +28,14 @@ foreach(_library ${OSGPlugins_FIND_COMPONENTS})
     set(_component OSGPlugins_${_library})
 
     set(${_library_uc}_DIR ${OSGPlugins_LIB_DIR}) # to help function osg_find_library
-    set(_saved_lib_prefix ${CMAKE_FIND_LIBRARY_PREFIXES}) # save CMAKE_FIND_LIBRARY_PREFIXES
-    set(CMAKE_FIND_LIBRARY_PREFIXES "") # search libraries with no prefix
+    if (NOT MINGW)
+        set(_saved_lib_prefix ${CMAKE_FIND_LIBRARY_PREFIXES}) # save CMAKE_FIND_LIBRARY_PREFIXES
+        set(CMAKE_FIND_LIBRARY_PREFIXES "") # search also libraries with no prefix
+    endif()
```
kcat: This looks sketchy to me. Either the currently set prefixes should be used, or they shouldn't. Why do you not want to clear them for MinGW specifically, but do clear them for everything else?

apps/openmw/CMakeLists.txt
the-eater: What do you mean exactly? pointing to the correct pkg-config isn't really the responsibility here. it's more that if your set on not using `pkg-config` you can turn it off. and Win32 to my knowledge has no pkg-config.

Setting the correct pkg-config dir and path should as you've shown, happen in the toolchain file

cmake/FindOSGPlugins.cmake
the-eater: I am actually wondering why this is here in the first place, because theoretically this would break on default on Linux, as lib is always used as prefix.

MinGW follows Unix conventions, and with the current toolchain all osgPlugins get prefixed with `lib`, so it breaks with `CMAKE_FIND_LIBRARY_PREFIXES` set to ""

apps/openmw/CMakeLists.txt
kcat: > it's more that if your set on not using pkg-config you can turn it off. and Win32 to my knowledge has no pkg-config.

Hmm, well I could've sworn my MinGW came with x86_64-w64-mingw32-pkg-config. But either way, when installing libraries with MinGW they generally come with .pc files like for GCC. I don't see why you'd want to ignore them since they help get the correct paths and compiler flags.

> Setting the correct pkg-config dir and path should as you've shown, happen in the toolchain file

Right, which is why I'm not sure it's a good idea to ignore pkg-config with MinGW, since the toolchain file should set it up to work correctly.

the-eater: oooh, you're misreading the last expression :)

`NOT WIN32 OR MINGW`

it will use pkg-config things when it's not Win32 **or** uses MinGW.

So the only case where it's not doing pkg-config is when you're building for windows, but are not using MinGW

cmake/FindOSGPlugins.cmake
kcat: Looks like FindOSGPlugins.cmake is only used for static builds and OSX deployment. I can't really see why it should be cleared either since the libs should have the default prefix.

apps/openmw/CMakeLists.txt
kcat: Ah, sorry. My mistake.

cmake/FindOSGPlugins.cmake
the-eater: Yeah, I could put it under a flag, if you're not fine with this.

psi29a:
How are we doing here? :)

cmake/FindOSGPlugins.cmake
AnyOldName3: I think I tried meddling with this at one point, but can't remember if I worked out why it needed to be how it is in order to work. I might be able to remember stuff if prompted.

akortunov:

PR #2290, opened 2019-03-28
Relax CMake version requirements on macOS

nikolaykasyanov:


CMakeLists.txt
```diff
@@ -796,8 +796,8 @@ endif()
 
 # Apple bundling
 if (OPENMW_OSX_DEPLOYMENT AND APPLE AND DESIRED_QT_VERSION MATCHES 5)
-    if (${CMAKE_MAJOR_VERSION} STREQUAL "3" AND ${CMAKE_MINOR_VERSION} STREQUAL "13")
-        message(FATAL_ERROR "macOS packaging is broken in CMake 3.13.*, see https://gitlab.com/OpenMW/openmw/issues/4767. Please use an older version like 3.12.4")
+    if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.13 AND CMAKE_VERSION VERSION_LESS 3.13.4)
+        message(FATAL_ERROR "macOS packaging is broken in early CMake 3.13 releases, see https://gitlab.com/OpenMW/openmw/issues/4767. Please use at least 3.13.4 or an older version like 3.12.4")
```
psi29a: Doesn't this mean that we can't use versions later than 3.13.4?

Capostrophic: Looks fine to me.
```cpp
if (version >= 3.13 && version < 3.13.4)
   message("cmake brek sry");
```

Capostrophic:
VERSION_GREATER_EQUAL wasn't a thing in CMake 3.1. I assume even for macOS deployment the compatibility with it should be maintained.

CMakeLists.txt
psi29a: Gotcha this was to guard against broken shit. Thanks. Was tooooo late for me.

nikolaykasyanov:
@Capostrophic good point, thanks, I'll look into it.

PR #2291, opened 2019-03-28
Improve compiler messages

Capostrophic:
1. Make compiler messages and exceptions start from a capital letter consistently.
2. The fact that the erroneous syntax was ignored was stated far too often, even though it was pretty obvious. Now it's considered implicit.


akortunov:
LGTM.

PR #2292, opened 2019-03-29
Make fallback map static to simplify constructors

akortunov:
The main idea - make the fallback map static as the settings map.
New syntax looks like this:
```
const std::string& value = Fallback::Map::getString(settingName);
```
As a result, a code is a lot cleaner since we do not need to pass the map into every constructor and store it in different places.

Also now we read configs before GUI init in the editor, so we can populate enums and labels by values from openmw.cfg (can be useful for blood types or weapon types).

### DIFF
--- a/apps/openmw/mwrender/characterpreview.cpp
+++ b/apps/openmw/mwrender/characterpreview.cpp
@@ -168,16 +168,15 @@ namespace MWRender
         stateset->setAttributeAndModes(lightmodel, osg::StateAttribute::ON);
 
         osg::ref_ptr<osg::Light> light = new osg::Light;
-        const Fallback::Map* fallback = MWBase::Environment::get().getWorld()->getFallback();
-        float diffuseR = fallback->getFallbackFloat("Inventory_DirectionalDiffuseR");
-        float diffuseG = fallback->getFallbackFloat("Inventory_DirectionalDiffuseG");
-        float diffuseB = fallback->getFallbackFloat("Inventory_DirectionalDiffuseB");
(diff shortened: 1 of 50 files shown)


apps/openmw/mwrender/characterpreview.cpp
```diff
@@ -168,15 +168,14 @@ namespace MWRender
...
-        float ambientB = fallback->getFallbackFloat("Inventory_DirectionalAmbientB");
-        float azimuth = osg::DegreesToRadians(180.f - fallback->getFallbackFloat("Inventory_DirectionalRotationX"));
-        float altitude = osg::DegreesToRadians(fallback->getFallbackFloat("Inventory_DirectionalRotationY"));
+        float diffuseR = Fallback::Map::getFloat("Inventory_DirectionalDiffuseR");
+        float diffuseG = Fallback::Map::getFloat("Inventory_DirectionalDiffuseG");
+        float diffuseB = Fallback::Map::getFloat("Inventory_DirectionalDiffuseB");
+        float ambientR = Fallback::Map::getFloat("Inventory_DirectionalAmbientR");
+        float ambientG = Fallback::Map::getFloat("Inventory_DirectionalAmbientG");
+        float ambientB = Fallback::Map::getFloat("Inventory_DirectionalAmbientB");
+        float azimuth = osg::DegreesToRadians(180.f - Fallback::Map::getFloat("Inventory_DirectionalRotationX"));
```
Capostrophic: Since you're modifying this line anyway, it'd be nice if you removed the substraction from 180 degrees and put a minus before positionX definition.

akortunov:
Any progress with the review/testing? This blocks further progress on other PRs.

PR #2293, opened 2019-03-29
Make the default vertical FOV in line with vanilla (bug #4933)

Capostrophic:
[Bug 4933](https://gitlab.com/OpenMW/openmw/issues/4933).

Stating again:
I did some research and I think changing the default vertical FOV to 60 seems reasonable. Since Morrowind uses 75 horizontal FOV by default, which is a standard FOV for 4:3 aspect, the vertical FOV is supposed to be 60, but on widescreen resolutions the scene becomes cut off vertically using the same horizontal FOV. MGE using 91.3 degree FOV corrects the vertical FOV to be 60 again on 16:9, resulting in "wider" portion of the scene becoming visible.
Nicolay Korslund first hardcoded the vertical FOV to be 55 in 2008 based on unknown assumptions.
It was carried over faithfully since then.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
     Bug #4922: Werewolves can not attack if the transformation happens during attack
     Bug #4927: Spell effect having both a skill and an attribute assigned is a fatal error
     Bug #4932: Invalid records matching when loading save with edited plugin
+    Bug #4933: Field of View not equal with Morrowind
     Bug #4938: Strings from subrecords with actually empty headers can't be empty
     Bug #4942: Hand-to-Hand attack type is chosen randomly when "always use best attack" is turned off
     Bug #4945: Poor random magic magnitude distribution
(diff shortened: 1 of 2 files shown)


akortunov:
Did you check if we need to change the first person FOV as well? A setting description suggests to do not touch it.

Capostrophic:
Yes, we need to change it too, in both the reporter's and my own testing 55 degrees is too narrow compared to vanilla (in the same scene at the same point of view smaller portion of the hands is visible).

Pop000100:
Is there a reason why the vertical FOV isn't configurable(mind my ignorance i haven't worked on camera stuff)?

Capostrophic:
It's configurable. I changed its default value in this PR and the Field of View slider in the in-game settings controls the value of "field of view" setting. First person field of view, however, needs to be set manually in the config file.

Capostrophic:
Visual glitches akortunov mentioned - character's body being visible in the lower left corner - are actually noticeable even with the current default FOV, and are likely caused by the camera implementation being inaccurate to vanilla. I don't think this worsens the issue that much (considering many players don't even use the default FOV and instead use a higher vertical FOV without issue).

Capostrophic:
akortunov stated he's mostly OK with this being merged though his concerns above still hold.

AnyOldName3:
As mentioned on GitLab, this isn't *exactly* the original game's FOV - that's 59.84...°, which is noticeable in comparison screenshots, but not when playing.

PR #2295, opened 2019-03-30
Fix macOS CI build timeouts

nikolaykasyanov:
~The recent builds are taking too much time & being terminated by Travis a lot.
I'll look into ccache and if it helps, re-enable the step.~

This PR improves macOS build times by switching to the most recent macOS image & using ccache.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,9 +41,9 @@ addons:
     build_command: "make VERBOSE=1 -j3"
 matrix:
   include:
-    - name: OpenMW (all) on macOS Xcode 10.1
+    - name: OpenMW (all) on macOS Xcode 10.2
       os: osx
-      osx_image: xcode10.1
+      osx_image: xcode10.2
       if: branch != coverity_scan
     - name: OpenMW (all) on Ubuntu Xenial GCC-5
       os: linux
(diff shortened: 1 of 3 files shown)


nikolaykasyanov:
Doesn't help much, looking into ccache.

psi29a:
Some of the build timeouts is caused by openmw being built twice on macos... I think somehow one of the script variables is being set.

psi29a:
Looks like these builds do take longer... lots of checking going on?  Perhaps tell compiler not to do optimizations and set flags for building fast?

nikolaykasyanov:
@psi29a funnily, the last time I was improving build times, switching to release build helped, probably by reducing disk I/O. But yeah, that'll be on my list of the things to try.

.travis.yml
```diff
@@ -92,7 +92,7 @@ before_script:
     - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then ./CI/before_script.${TRAVIS_OS_NAME}.sh; fi
 script:
  - cd ./build
- - if [ "${COVERITY_SCAN_BRANCH}" != 1 ]; then ${ANALYZE} make -j3; fi
+ - if [ "${COVERITY_SCAN_BRANCH}" != 1 ] && [ ! -z "${ANALYZE}" ]; then ${ANALYZE} make -j3; fi
```
nikolaykasyanov: @psi29a does this make sense?

nikolaykasyanov: Nope, it doesn’t 😅

nikolaykasyanov:
Interesting, now it builds in 25 minutes without ccache. My guess is that Travis CI is throttling older macOS images somehow and switching to Xcode 10.2 helps.

PR #2296, opened 2019-03-30
Optimize blendmap generation

akortunov:
Based on bzzt's changes.

Currently for some cells blendmap generation takes about 3ms, what can lead to FPS drop during cell transitions. With this PR generation takes about 0.5ms (roughly 6x faster).

The main idea - do not blend with black texture, get rid of outer loop.

### DIFF
--- a/components/esmterrain/storage.cpp
+++ b/components/esmterrain/storage.cpp
@@ -387,25 +387,42 @@ namespace ESMTerrain
         const int blendmapImageSize = blendmapSize * imageScaleFactor;
 
         LandCache cache;
-        std::map<UniqueTextureId, int> textureIndicesMap;
+        std::map<UniqueTextureId, unsigned int> textureIndicesMap;
 
-        for (int y=0; y<blendmapSize; ++y)
+        for (int y=0; y<blendmapSize; y++)
         {
-            for (int x=0; x<blendmapSize; ++x)
+            for (int x=0; x<blendmapSize; x++)
             {
                 UniqueTextureId id = getVtexIndexAt(cellX, cellY, x+rowStart, y+colStart, cache);
-                std::map<UniqueTextureId, int>::iterator found = textureIndicesMap.find(id);
+                std::map<UniqueTextureId, unsigned int>::iterator found = textureIndicesMap.find(id);
                 if (found == textureIndicesMap.end())
                 {
-                    found = textureIndicesMap.insert(std::make_pair(id, textureIndicesMap.size())).first;
-                    layerList.push_back(getLayerInfo(getTextureName(id)));
-                    osg::ref_ptr<osg::Image> image (new osg::Image);
-                    image->allocateImage(blendmapImageSize, blendmapImageSize, 1, GL_ALPHA, GL_UNSIGNED_BYTE);
-                    unsigned char* pData = image->data();
-                    memset(pData, 0, image->getTotalDataSize());
-                    blendmaps.push_back(image);
+                    unsigned int layerIndex = layerList.size();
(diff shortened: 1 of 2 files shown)


akortunov:
Any objections here?

components/esmterrain/storage.cpp
```diff
@@ -385,72 +380,61 @@ namespace ESMTerrain
...
-        for (int i=0; i<numBlendmaps; ++i)
-        {
-            osg::ref_ptr<osg::Image> image (new osg::Image);
-            image->allocateImage(blendmapImageSize, blendmapImageSize, 1, GL_ALPHA, GL_UNSIGNED_BYTE);
-            unsigned char* pData = image->data();
+        LandCache cache;
+        std::map<UniqueTextureId, unsigned int> textureIndicesMap;
 
-            for (int y=0; y<blendmapSize; ++y)
+        for (int y=0; y<blendmapSize; ++y)
```
Capostrophic: Starting to nitpick. Making integer copies (4 bytes) is cheaper on 64-bit architecture than making integer references (8 bytes).

components/esmterrain/storage.cpp
```diff
@@ -385,72 +380,61 @@ namespace ESMTerrain
...
+                        if (layerList[i].mDiffuseMap == info.mDiffuseMap)
+                        {
+                            layerIndex = i;
+                            break;
+                        }
+                    }
 
-                    int realY = (blendmapSize - y - 1)*imageScaleFactor;
-                    int realX = x*imageScaleFactor;
+                    found = textureIndicesMap.insert(std::make_pair(id, layerIndex)).first;
```
Capostrophic: emplace

components/esmterrain/storage.cpp
```diff
@@ -385,72 +380,61 @@ namespace ESMTerrain
...
-                    pData[(realY+0)*blendmapImageSize + realX + 1] = alpha;
-                    pData[(realY+1)*blendmapImageSize + realX + 1] = alpha;
+                    if (layerIndex >= layerList.size())
+                    {
+                        osg::ref_ptr<osg::Image> image (new osg::Image);
+                        image->allocateImage(blendmapImageSize, blendmapImageSize, 1, GL_ALPHA, GL_UNSIGNED_BYTE);
+                        unsigned char* pData = image->data();
+                        memset(pData, 0, image->getTotalDataSize());
+                        blendmaps.push_back(image);
+                        layerList.push_back(info);
```
Capostrophic: can use emplace_back here

components/esmterrain/storage.cpp
```diff
@@ -385,72 +380,61 @@ namespace ESMTerrain
...
 
-                    int realY = (blendmapSize - y - 1)*imageScaleFactor;
-                    int realX = x*imageScaleFactor;
+                    found = textureIndicesMap.insert(std::make_pair(id, layerIndex)).first;
 
-                    pData[(realY+0)*blendmapImageSize + realX + 0] = alpha;
-                    pData[(realY+1)*blendmapImageSize + realX + 0] = alpha;
-                    pData[(realY+0)*blendmapImageSize + realX + 1] = alpha;
-                    pData[(realY+1)*blendmapImageSize + realX + 1] = alpha;
+                    if (layerIndex >= layerList.size())
```
Capostrophic: layerIndex will never be greater than the size of the list. In fact, the whole layer index handling in this scope is obscure. Not sure how exactly to adjust the surrounding code with that in mind, though.

akortunov: But it still can be **equal** to the size of list. It seems the > is here just for sure.

Capostrophic:
I'll bite the bullet.

PR #2298, opened 2019-03-30
Do not allocate eight dummy LightStateAttribute's for every stateset

akortunov:
Based on bzzt's work.

If I understand correctly, now we assign LightStateAttribute's for every stateset in the LightManager to reset lights to default when we pop stateset.

bzzt think that it is enough to allocate dummy attributes only once and re-use them for different statesets.
It may be a good idea if the dummies are really dummies, or may be a bad idea if we write data to them.


AnyOldName3:
The change looks valid to me, but I can't tell what the dummy `LightStateAttribute`s are actually achieving. When they have an empty light list (which the dummies do) they don't actually seem to do anything.

PR #2300, opened 2019-03-31
[Do not merge] Shadow-related fixes from bzzt's branch

akortunov:
Do not merge! It is just supposed to make easier to review or test these changes!
@AnyOldName3, it is specially for you, so you will not have to review the whole PR with 98 commits which change completely unrelated stuff.

Most noticable changes:
1. Custom ShadowsBin, which decreases a Draw thread loading when shadows are enabled.
Unfortunately, it increases the Cull time in the main thread, so this change may actually decrease FPS on some setups (including mine).
2. Shadows are added to the water refraction.
An interesting idea, but needs some tweaking.


akortunov:
Is not needed anymore.

PR #2306, opened 2019-04-04
Add various coordinate conversion functions

unelsson:
Based on my, plutonicoverkills (terrain texture selection pr) and akortunov's (terrainediting branch) work. Piece of terrain texture selection, terrain vertex selection and terrain shape edit. Cut apart for easier reviewing and a base feature needed by various other optional features (land scaling, proc. generation etc.)

### DIFF
--- a/apps/opencs/model/world/cellcoordinates.hpp
+++ b/apps/opencs/model/world/cellcoordinates.hpp
@@ -7,6 +7,8 @@
 
 #include <QMetaType>
 
+#include <osg/Vec3d>
+
 namespace CSMWorld
 {
     class CellCoordinates
--- a/apps/opencs/model/world/cellcoordinates.hpp
+++ b/apps/opencs/model/world/cellcoordinates.hpp
@@ -7,6 +7,8 @@
 
 #include <QMetaType>
 
+#include <osg/PositionAttitudeTransform>
+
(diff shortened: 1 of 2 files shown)


akortunov:
These functions are not used and I do not know what they are supposed to do, so it is a quite hard to tell if they are OK.

unelsson:
I added comments to header file to explain what they are supposed to do. While worldspace coordinates, local texture coordinates, vertex coodinates and cell ids are already used in OpenMW, global selection coordinates are new.

Global selection coordinates are similar to texture grid in cell, or vertex grid in cell, but a continuous line from zero rather than cell-based coordinates. For example, global texture coordinate 18, 0 is in cell 1, 0 at local coordinate 2, 0.

I understand the difficulty of testing though, as they are not used... It's a bit tricky to use them in same PR without making the whole PR very big.

apps/opencs/model/world/cellcoordinates.hpp
```diff
@@ -7,6 +7,8 @@
 
 #include <QMetaType>
 
+#include <osg/PositionAttitudeTransform>
```
AnyOldName3: Is this include needed? I don't see any `osg::PositionAttitudeTransform`s anywhere.

PR #2307, opened 2019-04-06
Reset old record, if its ID does not match with ID of new record

akortunov:
Fixes [bug #4932](https://gitlab.com/OpenMW/openmw/issues/4932).

Unfortunately, TES CS likes to shuffle RefNums, so we will have to reset most of records anyway if some  instance records were removed from plugin, but at least we will avoid false matches.

OpenMW-CS seems to do not have this issue - it just removes instance records from omwaddon files ithout shuffling RefNums.


Capostrophic:
It seems to work. Do you want anything specific to be tested?

akortunov:
> It seems to work. Do you want anything specific to be tested?

I wanted working Windows CI to send the build to original reporter.

akortunov:
Author of the bugreport claims that the fix works as intended too.

PR #2309, opened 2019-04-08
Add an option to use casting animations for magic items

akortunov:
Implements [feature #4962](https://gitlab.com/OpenMW/openmw/issues/4962).

Just refactors our spellcasting code to support both spells and magic items.
Option is disabled by default to replicate vanilla behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@
     Feature #4859: Make water reflections more configurable
     Feature #4887: Add openmw command option to set initial random seed
     Feature #4890: Make Distant Terrain configurable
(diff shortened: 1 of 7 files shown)


Capostrophic:
As though as I regret the current option-adding routine, this should get a launcher advanced tab checkbox for visibility.

akortunov:
> As though as I regret the current option-adding routine, this should get a launcher advanced tab checkbox for visibility.

I have no idea how to work with the launcher.

Capostrophic:
That explains it... well, remind me to do that manually later.

Seems to work OK. Merging.

PR #2311, opened 2019-04-08
Use dice rolls instead of chance for spell magnitude (bug #4945)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/4945).

Hopefully improves magic effect random magnitude distribution, now it's humanly possible to get the max magnitude of a spell/constant ability/enchantment applied to the character. This assumes that spell magnitudes are valid (i.e. max magnitude is higher than min magnitude). This doesn't do anything about the fact that spell magnitudes are not as pseudo-random as in Morrowind (i.e. can become mean average after continuous recalculations).

In spells and enchantments, I had to do something hacky to avoid modifying the saved spell parameters format.


Capostrophic:
CMAugust confirmed that the chance to get any magnitude is roughly 100/(max-min+1) percent with these changes versus (near) 0 chance to get the max magnitude on 0.45.0. So I think this should be good to go.

PR #2314, opened 2019-04-09
bump from C++11 to C++14

psi29a:
give it a spin...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -405,9 +405,9 @@ if (NOT WIN32 AND NOT APPLE)
 endif()
 
 # CXX Compiler settings
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
 if (CMAKE_CXX_COMPILER_ID STREQUAL GNU OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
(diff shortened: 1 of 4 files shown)


akortunov:
Just for reference - what the C++14 has for us?
Also, IIRC, MSVC2015 does not have the full C++14 support.

psi29a:
We'll find out as we start adding features.  MSVC19 is also out or almost out? If we add support for that, then we drop MSVC2015.

std::make_unique

akortunov:
>  MSVC19 is also out or almost out?

It is a Preview only. Before dropping 2015, we will need to make sure that 2019 builds OpenMW properly.

psi29a:
For now, we're OK with MSVC2015 so long as we don't use anything too esoteric. The moment people want to use things not supported by MSVC2015 is the moment we should look into moving on to MSVC2019.

AnyOldName3:
> > MSVC19 is also out or almost out?
> 
> It is a Preview only. Before dropping 2015, we will need to make sure that 2019 builds OpenMW properly.

I'm pretty sure the preview ended about a week ago and it's now been released properly.

I've only just transitioned from using 2015 to 2017 a few weeks ago and only have 2015 set up to use the same dependency versions as our releases, so if we can avoid dropping 2015 support for a bit, that might end up being useful.

It looks like the only missing features are a relaxation on what can be used as a `constexpr` and a minor change to how aggregates are initialised, so it might be a while before anything actually comes up that causes problems.

psi29a:
Well, I'm not going to do any more to this PR. So feel free to merge whenever someone feels like it.

psi29a:
Bwah hahahahahha!

PR #2315, opened 2019-04-09
Revise light source configuration (bug #4965)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/4965).

1. configureLight has been made a part of SceneUtil namespace and is used for Light effect light source.
2. Instead of supplying the arguments every time the functions in lightutil are called, they're only used once in the function that uses them - configureLight - greatly reducing some noticeable bloat.
3. Constant attenuation settings are now used. Sometimes mod guides and mods recommend certain attenuation settings, and constant attenuation is often an important part of getting the right look of the game.

Currently there's still the assumption that the light source never changes the cell for the purposes of switching linear/quadratic attenuation for a setting (which is off by default).

### DIFF
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -38,8 +38,6 @@
 
 #include <components/settings/settings.hpp>
 
-#include <components/fallback/fallback.hpp>
-
 #include "../mwbase/environment.hpp"
 #include "../mwbase/world.hpp"
 #include "../mwworld/esmstore.hpp"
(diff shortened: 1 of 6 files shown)


apps/openmw/mwrender/animation.cpp
```diff
@@ -1833,14 +1823,10 @@ namespace MWRender
...
         {
-            // TODO: use global attenuation settings
-
             // 1 pt of Light magnitude corresponds to 1 foot of radius
             float radius = effect * std::ceil(Constants::UnitsPerFoot);
-            const float linearValue = 3.f; // Currently hardcoded: unmodified Morrowind attenuation settings
-            float linearAttenuation = linearValue / radius;
 
-            if (!mGlowLight || linearAttenuation != mGlowLight->getLight(0)->getLinearAttenuation())
+            if (!mGlowLight || (radius * 3) != mGlowLight->getRadius())
```
akortunov: I'd prefer to keep the `linearValue` with comment. "3" is just a magic number here.

Capostrophic: It's not linearValue, it's the arbitrary multiplier chosen to reduce light pop-in as described below.

akortunov: Then create the variable with the `radius * 3` value, comment it and use it in both places.
Otherwise persons which will read this code will need to guess if both `3` are the same constant, or really different ones which accidentally got the same value in the current codebase.

akortunov:
Has a conflict...

psi29a:
resolved

PR #2316, opened 2019-04-10
Don't clear the depth buffer to render the first person model

sade1212:
I am not a programmer, I don't know C++ or OpenSceneGraph, and I have no idea what I'm doing, other than hopefully invoking some variation of Cunningham's Law. Sorry if this is a horrible waste of your time.

I wanted to use [reshade](https://reshade.me/) with OpenMW to inject ambient occlusion (until such a time as post-processing shaders are supported natively). However, that shader requires depth buffer access, which is [deliberately cleared in first-person mode in OpenMW so the hands/weapon models don't clip with the world](https://forum.openmw.org/viewtopic.php?t=3230).

Scrawl mentioned glPolygonOffset in the thread I linked, so I googled that and came up with this alternative (horrible and hacky) way of stopping the first person model from clipping, while also leaving all the depth information that reshade needs accessible. I had to set and unset the polygon offset depending on whether you're in first person mode because otherwise the third person player character would be rendered at an incorrect depth.

Ideally the GUI would also be output to the depth buffer as infinitely close so that Reshade's depth of field filters don't blur it, but I haven't been able to work out a way to do that with my layman's skills.

### DIFF
--- a/apps/openmw/mwrender/npcanimation.cpp
+++ b/apps/openmw/mwrender/npcanimation.cpp
@@ -370,18 +370,18 @@ private:
 
 void NpcAnimation::setRenderBin()
 {
-    osg::StateSet* stoteset = mObjectRoot->getOrCreateStateSet();
+    osg::StateSet* stateset = mObjectRoot->getOrCreateStateSet();
     osg::ref_ptr<osg::PolygonOffset> polyoffset = new osg::PolygonOffset;
     polyoffset->setFactor(1.0f);
     if (mViewMode == VM_FirstPerson)
     {
         polyoffset->setUnits(-1000000.0f);
-        stoteset->setAttributeAndModes(polyoffset, osg::StateAttribute::OVERRIDE | osg::StateAttribute::ON);
-        stoteset->setRenderBinDetails(RenderBin_FirstPerson, "DepthClear", osg::StateSet::OVERRIDE_RENDERBIN_DETAILS);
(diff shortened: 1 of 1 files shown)


AnyOldName3:
I'm not totally convinced that this is going to be completely reliable - I'd have thought that by the time you've shifted things enough that you're guaranteeing they don't clip, you're risking having issues with the near plane.

The alternative approach suggested in the linked thread - rendering to another framebuffer and copying across - is what I'd have gone for. Either doing something with the stencil buffer or by doing it in a shader and reading the alpha channel should let you transfer depth values for only the pixels with stuff in them.

kcat:
Another possibility is a two-pass approach. Render the main scene first, then disable color writing, invert the depth testing, enable depth writing, and render the first-person model. Then set color writing, depth testing, and depth writing as appropriate and render the first-person model again. The first render pass will fill in part of the depth buffer with values that are equal to or farther than the first-person model (essentially making a cutout in the depth buffer), and the second render will draw the first-person model within that cutout.

akortunov:
> The alternative approach suggested in the linked thread - rendering to another framebuffer and copying across - is what I'd have gone for.

> Another possibility is a two-pass approach. Render the main scene first, then disable color writing, invert the depth testing, enable depth writing, and render the first-person model. Then set color writing, depth testing, and depth writing as appropriate and render the first-person model again.

Judging by

> I am not a programmer, I don't know C++ or OpenSceneGraph, and I have no idea what I'm doing

PR's author seems to have no skills to implement it properly.

Capostrophic:
Closing for the time being.

PR #2321, opened 2019-04-12
Make data from Resources folder overridable

akortunov:
Summary of changes:
1. Inject data from the Resources folder to the VFS (mostly MyGUI XML files ands shaders), so users can override them ([feature #4974](https://gitlab.com/OpenMW/openmw/issues/4974)).
2. Provide custom textures for scrollbars, so they will not use a bookmark texture.
3. Since this PR changes a way to configure fonts anyway, provide TrueType fonts with OpenMW and add a way to configure them via launcher ([feature #4975](https://gitlab.com/OpenMW/openmw/issues/4975)).

As a result, shaders can be overridable by files in the `shaders` folder in VFS, MyGUI files under `mygui` folder, and new scrollbar textures - under `textures` folder.


psi29a:
How does this effect our documentation? Do we need modify existing documentation and add new bits?

akortunov:
> How does this effect our documentation? Do we need modify existing documentation and add new bits?

Depends on which configuration we already have.
I see no docs about shaders/UI layout tweaking.

Capostrophic:
I assume you've read the posts that were vetoing this kind of change already: https://forum.openmw.org/viewtopic.php?f=2&t=2774&start=30#p31504

Not happy about 3 in general.

akortunov:
> I assume you've read the posts that were vetoing this kind of change already: https://forum.openmw.org/viewtopic.php?f=2&t=2774&start=30#p31504
> 
> Not happy about 3 in general.

We already provide the DejaVuLGCSansMono.ttf for console.
As a result, we have a strange situation when we can not provide 3d-party fonts, but already provide a 3d-party font.
So we should either provide a full set of fonts (so for users it will be enough just to enable setting) or to do not provide them at all.
In the latter case we can use the Century Gothic font by default for console, and users will have to install all replacers manually.
Of course, sometimes they will fail to do that manually and we will have to spend additional time for technical support.

Capostrophic:
>So we should either provide a full set of fonts (so for users it will be enough just to enable setting) or to do not provide them at all.

Only the Sith deal in absolutes. It's not an either-or situation, a fine line has already been drawn, and I'm opposed to making *specific* fonts a hardcoded part of in theory and future a general-purpose engine.

>Of course, sometimes they will fail to do that manually

I sense the sarcasm. It's their problem. Adding `use ttf` setting obfuscates the situation further.

Think about how once a feature is added it can't be removed easily.

akortunov:
> Only the Sith deal in absolutes. 
```
> _ <
```

If you will veto this change, there is an alternative approach:
1. Host fonts on the openmw.org and place link to the fonts setup guide. In this case users can just install them as any other mod without manual messing with XML files.
2. Replace the "use ttf" checkbox in the launcher to the "scaling factor" spinbox.
3. Drop the "use ttf" setting.
4. Rename the "MonoFont" to the "Century Gothic", so the default Morrowind font will be used if a user breaks the openmw_font.xml or removes entry for console font from it for some reason.
5. Provide a reasonable explanation why we replace one resources just as we want (Fonts), but can not ship additional optional resources of the same type.

akortunov:
> How exactly does one go about adding more VFS layers with this? Do you just need to add multiple resources= lines to openmw.cfg?

No, just install data as any other mod. XML files should be in the MyGUI subfolder, shaders should be in the Shaders subfolder.

For example, player wants to replace fonts:

```
data="/share/Games/OpenMW/Mods/Fonts"
```

Then he should copy openmw_font.xml to the /share/Games/OpenMW/Mods/Fonts/MyGUI, modify it, and place new TTF/OTF files here. A modified openmw_font.xml will take higher priority than the basic file from the Resources folder.

Unfortunately, MyGUI does not support multiple resources folder out of box, so I had to add a workaround - make the DataManager to copy data from VFS rather than just read files directly from disk.

AnyOldName3:
I'm not sure having that kind of thing in the data directories is the best idea.

terabyte25:
I for one, support this. Currently it's pretty much impossible for the user on Android to even be able to configure anything with the resources folder, so this is great for the user to have custom windows and fog. 

However, I am not sure how this would work with the version file as OpenMW uses it for displaying commit hashes in the main menu, and it is located in the resources folder. Maybe blacklist it from being overridden?

akortunov:
> However, I am not sure how this would work with the version file as OpenMW uses it for displaying commit hashes in the main menu, and it is located in the resources folder. Maybe blacklist it from being overridden?

OpenMW still loads this file directly from disk.

> I'm not sure having that kind of thing in the data directories is the best idea.

Do you have a better idea how to persistently mod UI layout?

AnyOldName3:
> > I'm not sure having that kind of thing in the data directories is the best idea.
> 
> Do you have a better idea how to persistently mod UI layout?

Maybe have a separate VFS for resources? I'm not sure. At the least, I think maybe having stuff in an `openmw-resources` subdirectory of the data directory rather than just `MyGUI` or whatever might be better.

akortunov:
> Maybe have a separate VFS for resources?

What is it supposed to give us and end-users?

kcat:
Can we even guarantee that resources will be backwards compatible? If we ever use something else other than MyGUI, or if a MyGUI breaks compatibility, any modded xml and layout files could be a problem. Similarly, I don't think we can depend on the shaders not needing changes, which could break replacement shaders.

Meshes, textures, sounds, and other related stuff can be overridden because we guarantee they'll remain supported as-is. The only thing that can break them is another mod. But the UI and shaders have no such guarantee, which is likely why previous attempts/requests for making them overridable were rejected.

akortunov:
> Can we even guarantee that resources will be backwards compatible? If we ever use something else other than MyGUI, or if a MyGUI update breaks compatibility, any modded xml and layout files could be a problem.

Actually, it is a common issue with mods for games, which are in active developments.
We will have to deal with it later anyway, especially when we will start to improve modding abilities after 1.0. We can not guarantee _eternal_ backward compatibility, and it does not matter if we will replace OpenSceneGraph for UnrealEngine, Bullet to PhysX, or just will fix existing bugs in OpenMW. Even now we have no 100% compatibility with existing resources for Morrowind, and I doubt if it ever will be.

kcat:
> Actually, it is a common issue with mods for games, which are in active developments.
> We will have to deal with it later anyway, especially when we will start to improve modding abilities after 1.0.

That would mean then 1.0 would be blocked on getting a stable UI modding setup, if that's where the engine is considered stable. That doesn't sound like a very nice prospect.

> We can not guarantee eternal backward compatibility, and it does not matter if we will replace OpenSceneGraph for UnrealEngine, Bullet to PhysX, or just will fix existing bugs in OpenMW. Even now we have no 100% compatibility with existing resources for Morrowind, and I doubt if it ever will be.

We don't guarantee eternal 100% compatibility, but we can boast a very high rate of compatibility with well-reasoned explanations of what's incompatible. The system is set up in a way that's implementation agnostic -- mods should continue working regardless of what we'll do to the engine, because they're not reliant on OSG or Bullet or anything, and we understand them and how they fit into the engine. Unlike the UI which is not only dependent on MyGUI to understand and deal with, but a particular version of MyGUI (AFAIK, MyGUI doesn't guarantee existing layout and xml files will work when upgrading).

We know how to keep compatibility with models, textures, sounds, esm/esps, and the like, so we have a path forward as we update anything relating to these things in the engine. It was also a design goal from the start to be compatible with existing mods. This is not true of the UI, which was created as-needed using whatever tools we found to work, and could need a fundamental rework to expose proper future-proof modding capabilities.

If there's one thing I learned from long-running open source engines, exposing things to modding too early will cause many more headaches later on, with many mods breaking due to poor early decisions and short-sightedness that makes it impossible to maintain with improvements (which will get modders upset at us for breaking their mods, and cause users to stick to old versions where their mods work and we can't support them).

PR #2322, opened 2019-04-12
Fix frozen rain when water refractions are disabled

Capostrophic:
Due to another set of changes from bzzt recently introduced underwater switch callback can use the viewpoint inheritted from the main camera instead of the eye point reflection camera has so that the rain doesn't get frozen for no reason, fixing [this](https://gitlab.com/OpenMW/openmw/issues/4897) regression.


akortunov:
LGTM.

PR #2325, opened 2019-04-13
Optimize the Distant Terrain memory consumption

akortunov:
Adds a rest of changes in the from bzzt's branch (excepts of related to textures).
For me it slightly decreases memory usage by the Distant Terrain (usually about 50-150 MB depending on vieving distance and landmass size).

An interesting thing - on some setups Distant Terrain with vanilla viewing system gives more FPS than the common terrain system, but it costs additional memory:
in the master branch - 100MB with vanilla landscape, 200MB with TR, 350MB with TR+PC
with this PR - 50MB with vanilla landscape, 150MB with TR, 300MB with TR+PC

**Please, measure memory consumption and FPS with this PR and with latest nightlies in these situations:**
1. Vanilla viewing distance, the Distant Terrain is enabled
2. Vanilla viewing distance, the Distant Terrain is disabled
3. Increased viewing distance (40960, for example), the Distant Terrain is enabled

Depending on results, we should decide:
1. If it is even worth to spend time to improve code from these commits or we can just ignore it.
2. If the Distant Terrain really should be disabled by default. There is a non-zero chance that for most of setups it will be really faster.

### DIFF
--- a/components/terrain/quadtreeworld.cpp
+++ b/components/terrain/quadtreeworld.cpp
@@ -188,15 +188,14 @@ public:
 
         if (node->getSize() <= mMinSize)
         {
-            // We arrived at a leaf
-            float minZ,maxZ;
-            if (mStorage->getMinMaxHeights(size, center, minZ, maxZ))
-            {
-                float cellWorldSize = mStorage->getCellWorldSize();
-                osg::BoundingBox boundingBox(osg::Vec3f((center.x()-halfSize)*cellWorldSize, (center.y()-halfSize)*cellWorldSize, minZ),
-                                        osg::Vec3f((center.x()+halfSize)*cellWorldSize, (center.y()+halfSize)*cellWorldSize, maxZ));
-                node->setBoundingBox(boundingBox);
-            }
+            // We arrived at a leaf.
+            // Since the tree is used for LOD level selection instead of culling, we do not need to load an actual height data here.
+            float minZ = -std::numeric_limits<float>::max();
+            float maxZ = std::numeric_limits<float>::max();
+            float cellWorldSize = mStorage->getCellWorldSize();
+            osg::BoundingBox boundingBox(osg::Vec3f((center.x()-halfSize)*cellWorldSize, (center.y()-halfSize)*cellWorldSize, minZ),
+                                    osg::Vec3f((center.x()+halfSize)*cellWorldSize, (center.y()+halfSize)*cellWorldSize, maxZ));
+            node->setBoundingBox(boundingBox);
             return node;
         }
         else
@@ -360,7 +359,7 @@ void QuadTreeWorld::accept(osg::NodeVisitor &nv)
                 mRootNode->traverseTo(vd, 1, osg::Vec2f(x+0.5,y+0.5));
             }
             else
-                mRootNode->traverse(vd, &nv, cv->getViewPoint(), true, mViewDistance);
+                mRootNode->traverse(vd, cv->getViewPoint(), mViewDistance);
         }
         else
         {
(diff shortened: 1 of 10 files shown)


Capostrophic:
All testing done with pure vanilla files and assets, 1920x1080 resolution, shaders forced for everything, AI processing range set to 7168, Medium resolution water with refractions enabled and reflection level set to World, max filtering settings and vsync off, with shadows disabled and navigator enabled, in Sadrith Mora performance-intensive spot, after waiting for some time.

git:
1. FPS 105, 412 MB of RAM used by openmw process
2. FPS 110-115, 545 MB of RAM used
3. FPS 90-100, 579 MB of RAM used

PR:
1. FPS 105, 419 MB of RAM used
2. FPS 115, 460 MB of RAM used
3. FPS 95-100, 514 MB of RAM used

components/terrain/terraindrawable.cpp
```diff
@@ -4,13 +4,19 @@
...
 namespace Terrain
 {
 
 TerrainDrawable::TerrainDrawable()
 {
 }
 
+TerrainDrawable::~TerrainDrawable()
+{
+}
```
Capostrophic: I think the constructor and destructor can be declared as
```cpp
TerrainDrawable() = default;
~TerrainDrawable() = default;
```
in the header file and these empty declarations aren't necessary.

Capostrophic:
1. Definitely.
2. No, not right now.

components/terrain/quadtreeworld.cpp
```diff
@@ -400,21 +351,23 @@ void QuadTreeWorld::accept(osg::NodeVisitor &nv)
...
-    {
-        // view point is the same, but must still update visible status in case the camera has rotated
-        for (unsigned int i=0; i<vd->getNumEntries(); ++i)
         {
-            ViewData::Entry& entry = vd->getEntry(i);
-            entry.set(entry.mNode, !static_cast<osgUtil::CullVisitor*>(&nv)->isCulled(entry.mNode->getBoundingBox()));
+            osgUtil::IntersectionVisitor* iv = static_cast<osgUtil::IntersectionVisitor*>(&nv);
+            osgUtil::LineSegmentIntersector* lineIntsersector = dynamic_cast<osgUtil::LineSegmentIntersector*>(iv->getIntersector());
+            if (!lineIntsersector)
+                throw std::runtime_error("Can not update QuadTreeWorld - the LineSegmentIntersector expected");
```
Capostrophic: Cannot update QuadTreeWorld: node visitor is not LineSegmentIntersector

components/terrain/quadtreeworld.cpp
```diff
@@ -400,21 +351,23 @@ void QuadTreeWorld::accept(osg::NodeVisitor &nv)
...
-    }
-    else if (isCullVisitor)
-    {
-        // view point is the same, but must still update visible status in case the camera has rotated
-        for (unsigned int i=0; i<vd->getNumEntries(); ++i)
         {
-            ViewData::Entry& entry = vd->getEntry(i);
-            entry.set(entry.mNode, !static_cast<osgUtil::CullVisitor*>(&nv)->isCulled(entry.mNode->getBoundingBox()));
+            osgUtil::IntersectionVisitor* iv = static_cast<osgUtil::IntersectionVisitor*>(&nv);
+            osgUtil::LineSegmentIntersector* lineIntsersector = dynamic_cast<osgUtil::LineSegmentIntersector*>(iv->getIntersector());
```
Capostrophic: The typo here and later in the variable name should be corrected. "lineIntersector"

components/terrain/quadtreeworld.cpp
```diff
@@ -191,15 +185,14 @@ class QuadTreeBuilder
...
-            float minZ,maxZ;
-            if (mStorage->getMinMaxHeights(size, center, minZ, maxZ))
-            {
-                float cellWorldSize = mStorage->getCellWorldSize();
-                osg::BoundingBox boundingBox(osg::Vec3f((center.x()-halfSize)*cellWorldSize, (center.y()-halfSize)*cellWorldSize, minZ),
-                                        osg::Vec3f((center.x()+halfSize)*cellWorldSize, (center.y()+halfSize)*cellWorldSize, maxZ));
-                node->setBoundingBox(boundingBox);
-            }
+            // We arrived at a leaf.
+            // Since the tree is used for LOD level selection instead of culling, we do not need to load an actual height data here.
```
Capostrophic: Since I'm complaining about grammar anyway...
"we do not need to load the actual height data here".

PR #2329, opened 2019-04-17
Some options shuffling

Capostrophic:
1. Always account for every follower travelling in fast travel price calculations. Normally the first follower travels for free. The respective option has been removed.

There are a few reasons for this:
* The option was never documented, although it did get an Advanced tab checkbox
* It's a "default" fix in MCP, so at least Hrnchamd deems it an issue worthy enough not to be in a purist player's game and be not intended by Bethesda (it does seem pretty nonsensical)
* It shouldn't be controversial enough to make it the hardcoded default even before the Great Dehardcoding comes
* I asked about such a change some time ago, didn't get any opinion on this.

2. Make previously used MCP-like normalized race speed behavior optional.
3. Document that option and also `strength influences hand to hand` option.

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -2372,7 +2372,8 @@ void CharacterController::update(float duration, bool animationOnly)
     moved.x() *= scale;
     moved.y() *= scale;
 
-    if (mPtr.getClass().isNpc() && !Settings::Manager::getBool("normalise race speed", "Game"))
+    static const bool normalizeSpeed = Settings::Manager::getBool("normalise race speed", "Game");
+    if (mPtr.getClass().isNpc() && !normalizeSpeed)
     {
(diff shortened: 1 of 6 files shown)


apps/openmw/mwmechanics/character.cpp
```diff
@@ -2372,7 +2372,7 @@ void CharacterController::update(float duration, bool animationOnly)
     moved.x() *= scale;
     moved.y() *= scale;
 
-    if(mPtr.getClass().isNpc())
+    if (mPtr.getClass().isNpc() && !Settings::Manager::getBool("normalise race speed", "Game"))
```
akortunov: In there a reason to retrieve the setting every frame for every NPC?

Capostrophic: I see what you mean. Adjusted it.

akortunov:
LGTM, merging.

PR #2330, opened 2019-04-17
Only collect every unique magic bolt sound once (bug #4964)

Capostrophic:
[Bug 4964](https://gitlab.com/OpenMW/openmw/issues/4964).

Use a set instead of a vector for magic projectile sounds in getMagicBoltData to make sure there aren't any duplicate sound IDs, so no multiple identical sounds play at the exact same position, previously making an impression of a louder single sound, and e.g. fixing Wrath of Dagoth spell projectile audio volume, which has 7 magic effects with the exact same bolt sound. Now the volume is the same as in vanilla.


akortunov:
LGTM.

PR #2332, opened 2019-04-20
Apply looping particles when adding spell to existing actor

akortunov:
Fixes [bug #4918](https://gitlab.com/OpenMW/openmw/issues/4918).

The main idea - when we add spells with constant effects to existing actor, we should apply looping particles (but should not play sounds). The `creatureStats.getSpells().add(id)` can not do it itself (it just manages the list of known spells), so we will have to do it outside this function. 

This PR handles scripts and diseases. Other usages of "add" should not cause issues since we rebuild target actor anyway (e.g. when we spawn actor or select an another race or birthsign).

Note: there is a little inconsistence with Morrowind here - when using "addSpell", OpenMW adds particles immediately, but when using the "removeSpell", it removes effects during next actor update.
Morrowind does opposite - adds particles during next update, but removes them immediately. 
But I think it does not really matter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
     Bug #4896: Title screen music doesn't loop
     Bug #4911: Editor: QOpenGLContext::swapBuffers() warning with Qt5
     Bug #4916: Specular power (shininess) material parameter is ignored when shaders are used.
+    Bug #4918: Abilities don't play looping VFX when they're initially applied
     Bug #4922: Werewolves can not attack if the transformation happens during attack
     Bug #4927: Spell effect having both a skill and an attribute assigned is a fatal error
     Bug #4932: Invalid records matching when loading save with edited plugin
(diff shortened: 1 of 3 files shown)


Capostrophic:
The CHANGELOG.md issue name/"short description" should be updated to avoid misunderstandings with whoever writes the changelog.

On a somewhat related note, the scale of the shield VFX seems to be incorrect: dagoth ur
screenshot012
Is there anything possible to do about it?

akortunov:
> Is there anything possible to do about it? Looks like looping particle visitor doesn't handle the scale.

There is nothing to do with scale - it is already an 1.0 for Dagoth Ur.
There is an another issue - Morrowind seems to resize looping particles for creatures depending on collision box size. For NPCs Morrowind seems to take in account race height and weight factors.

akortunov:
> The CHANGELOG.md issue name/"short description" should be updated to avoid misunderstandings with whoever writes the changelog.

I do not understand you. The CHANGELOG.md simply contains IDs and names of bug/features from bugtracker. Feel free to rename bug to whatever you like, and I will update the entry.

PR #2333, opened 2019-04-20
Actually check if setting gamma level is supported

akortunov:
Related to [feature #3491](https://gitlab.com/OpenMW/openmw/issues/3491).

Currently we forcibly disable gamma slider for everything excepts Windows indifferently if it is supported or not.

The main idea of this PR is to check if the setting works on a target machine, and disable gamma slider if it does not for some reason indifferently from OS.

Note: this PR obviously does not implement a post-processing shader to adjust the brightness level.
Such kind of shader more likely is a post-1.0 feature, which will not work in the fixed pipeline mode anyway. 

But with this PR 3491 becomes a not 1.0 blocker, also this issue actually is not related to either Linux or SDL.

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -524,7 +524,8 @@ void OMW::Engine::prepareEngine (Settings::Manager & settings)
     else
         gameControllerdb = ""; //if it doesn't exist, pass in an empty string
 
-    MWInput::InputManager* input = new MWInput::InputManager (mWindow, mViewer, mScreenCaptureHandler, mScreenCaptureOperation, keybinderUser, keybinderUserExists, gameControllerdb, mGrab);
+    bool isGammaSupported = false;
+    MWInput::InputManager* input = new MWInput::InputManager (mWindow, mViewer, mScreenCaptureHandler, mScreenCaptureOperation, keybinderUser, keybinderUserExists, gameControllerdb, mGrab, isGammaSupported);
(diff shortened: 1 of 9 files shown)


kcat:
I'm not sure this is a good idea. As it is now, the monitor's gamma/contrast isn't changed if the player hasn't modified the sliders. This is so it maintains the system setting when the user hasn't changed anything in the game. With this change, the gamma/contrast will always be set to the game's config settings regardless of what the system is set to.

OpenMW shouldn't be setting the monitor gamma at all. It's an antiquated concept for individual apps to do because it's a global setting that wholly and only affects a given screen and nothing else... set a different gamma/contrast in game, make a screenshot, then look at the screenshot outside of the game and it will appear darker or brighter, with more or less contrast (or instead of a screenshot, try recording video), creating disparity between what a player sees in game vs what their screenshots or videos show. Or switch between fullscreen and windowed (is gamma/contrast still applied in windowed mode? neither answer is good). This isn't something that a modern engine on modern systems (which OpenMW is supposed to be aiming for) should be doing.

This really needs to be done with a post-process shader, with the current monitor gamma/contrast code taken out to avoid problems. Yes, the feature will require a shader, but OpenMW can't be held back by requiring everything to work with the FFP. The long past time that shader support is a divisive feature in hardware (even vanilla Morrowind made optional use of pixel shaders, 17 years ago).

PR #2336, opened 2019-04-21
Use the logging system for compiler errors

akortunov:
Currently error handlers for scripting system use cout/cerr directly instead of logging system.
As a result, there is no filtering and color selection for them.

With this PR, script errors use the Error level, and script warnings - Info level (since they are usually harmless and do not require actions from user).

Note: I can rename the StreamErrorHandler to the LogErrorHandler, if needed.

### DIFF
--- a/apps/openmw/mwdialogue/scripttest.cpp
+++ b/apps/openmw/mwdialogue/scripttest.cpp
@@ -28,8 +28,7 @@ void test(const MWWorld::Ptr& actor, int &compiled, int &total, const Compiler::
 
     MWScript::CompilerContext compilerContext(MWScript::CompilerContext::Type_Dialogue);
     compilerContext.setExtensions(extensions);
-    std::ostream errorStream(std::cout.rdbuf());
-    Compiler::StreamErrorHandler errorHandler(errorStream);
+    Compiler::StreamErrorHandler errorHandler;
     errorHandler.setWarningsMode (warningsMode);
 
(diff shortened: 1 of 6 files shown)


apps/openmw/mwdialogue/scripttest.cpp
```diff
@@ -84,8 +83,7 @@ void test(const MWWorld::Ptr& actor, int &compiled, int &total, const Compiler::
 
                 if (!success)
                 {
-                    Log(Debug::Warning)
-                        << "compiling failed (dialogue script)\n" << info->mResultScript << "\n\n";
+                    Log(Debug::Error) << "Error: script compiling failed (dialogue script): " << info->mResultScript;
```
Capostrophic: The line breaks are gone.

Stating that "script" compilation specifically was failed here and in other instances of such messages is probably redundant.

Alternatively, here it could be dealt with like "dialogue script compiling/compilation failed".

akortunov: > Stating that "script" compilation specifically was failed here and in other instances of such messages is probably redundant.

Do you suggest to remove this message at all and use the single message as in dialogue scripts?

> The line breaks are gone.

There is no much gain from them since it is the same message, not different ones. They can be misleading here.

Capostrophic: >Do you suggest to remove this message at all and use the single message as in dialogue scripts?

No, I meant that adding the word "script" is unnecessary, it's obvious from the context.

akortunov:
Any other objections?

psi29a:
None from me, I like this.

PR #2337, opened 2019-04-21
[Feedback needed] Keep AI packages order when using repeating

akortunov:
Fixes [bug #4955](https://gitlab.com/OpenMW/openmw/issues/4955).

The issue was introduced in PR #947.

Brallion has two AI packages - an AiWander one with Duration = 5 and AiFollow one.
AiWander is first, so it takes a higher priority. After 5 hours in Sadrith Mora we remove AiWander package and place a new AiWander package to the back of AI sequence. As a result, now AiFollow takes a higher priority because it is earlier in the sequence.

There is no such issue in Morrowind - it is either does not alter packages order or does not handle Duration field at all (more likely a first option). So a suggested fix - replace an old package by a new package without altering packages order.

@Allofich, can you tell if the placing a new package exactly at the back of AI sequence is required, or you just did not consider packages order at all?

### DIFF
--- a/apps/openmw/mwmechanics/aisequence.cpp
+++ b/apps/openmw/mwmechanics/aisequence.cpp
@@ -274,17 +274,20 @@ void AiSequence::execute (const MWWorld::Ptr& actor, CharacterController& charac
         {
             if (package->execute (actor, characterController, mAiState, duration))
             {
-                // Put repeating noncombat AI packages on the end of the stack so they can be used again
+                // AiPackage::execute() can queue an another AI package, so we can not just use the mPackages.mfront() here.
+                std::list<MWMechanics::AiPackage*>::iterator toRemove =
+                        std::find(mPackages.begin(), mPackages.end(), package);
+
+                std::list<MWMechanics::AiPackage*>::iterator toInsert = mPackages.erase(toRemove);
+
+                // Replace finished repeating noncombat AI package by the new one.
                 if (isActualAiPackage(packageTypeId) && (mRepeat || package->getRepeat()))
                 {
                     package->reset();
-                    mPackages.push_back(package->clone());
+                    MWMechanics::AiPackage* clone = package->clone();
+                    mPackages.insert(toInsert, clone);
                 }
-                // To account for the rare case where AiPackage::execute() queued another AI package
-                // (e.g. AiPursue executing a dialogue script that uses startCombat)
-                std::list<MWMechanics::AiPackage*>::iterator toRemove =
-                        std::find(mPackages.begin(), mPackages.end(), package);
-                mPackages.erase(toRemove);
+
                 delete package;
                 if (isActualAiPackage(packageTypeId))
                     mDone = true;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@
     Bug #4945: Poor random magic magnitude distribution
     Bug #4947: Player character doesn't use lip animation
     Bug #4948: Footstep sounds while levitating on ground level
+    Bug #4955: AI package repeating alters packages order
     Bug #4963: Enchant skill progress is incorrect
     Bug #4964: Multiple effect spell projectile sounds play louder than vanilla
     Bug #4965: Global light attenuation settings setup is lacking


apps/openmw/mwmechanics/aisequence.cpp
```diff
@@ -274,17 +274,20 @@ void AiSequence::execute (const MWWorld::Ptr& actor, CharacterController& charac
         {
             if (package->execute (actor, characterController, mAiState, duration))
             {
-                // Put repeating noncombat AI packages on the end of the stack so they can be used again
+                // AiPackage::execute() can queue an another AI package, so we can not just use the mPackages.mfront() here.
+                std::list<MWMechanics::AiPackage*>::iterator toRemove =
+                        std::find(mPackages.begin(), mPackages.end(), package);
```
elsid: Possibly at this point `mPackages` might be empty or `package` might not belong to the list. So find would return `end()`. `end()` can't be used for `erase`.

Allofich:
> @Allofich, can you tell if the placing a new package exactly at the back of AI sequence is required, or you just did not consider packages order at all?

I did consider packages order.

I just tested the NPC of that bug report, Brallion in Sadrith Mora. In original Morrowind he also runs the follow package.

1. Start/load a fresh game (no data saved yet for Brallion)
2. `coc "Sadrith Mora"`
3. Brallion's first package is a 5-hour Wander. Click him and `GetCurrentAIPackage` returns 0 (Wander).
4. Wait 5 game hours.
5. Brallion's next package is a 0-hour (unlimited) Follow. Click him and `GetCurrentAIPackage` returns 3 (Follow).
6. Brallion's package remains Follow forever.

AI with Follow will not move unless the Follow target gets close to them first. Brallion will follow the target (in my case, it was `telvani guard00000000`. There are also`telvani guard00000001`m `telvani guard00000002`, etc. Maybe the engine picks the first "Telvanni Guard" it found) when I teleport the guard close to him with `setpos` .

If there is a difference between the original engine and Morrowind, maybe it is the distance at which AI with Follow will begin to move. Or maybe OpenMW will allow Brallion to follow any "Telvanni Guard"? (In my test of the original engine he ignored guards besides `telvani guard00000000`). Perhaps a difference in pathfinding or physics lets the guard get closer, etc.

------------------------------------------

BTW, just as information to understand the original game's packages...
I think the current cycling packages of OpenMW is only "on the surface" like original Morrowind. As far as I can remember, I found that original Morrowind works by loading the AI packages from the .esm and .esp files every time and then saving only an index into the currently running package. There is an index, stored as a number, in .ess save files for NPCs running an editor-placed AI package. ("0" for first package, "1 for next package", etc.)

If the AI package is added through script/console command, then the package information is added to the NPC record, but if it is an editor-placed package, I think I found that it was only the index that was stored.

If you have a saved game with an NPC with an editor-placed AI package, then load that save with a different .esp file that changes the NPC's editor-placed packages, they will run the new .esp packages, unlike OpenMW. Ex.) If at the time of the saved game they were running the package at index 1, they will run whatever package is at index 1 according to the loaded .esm/.esp files when you load the game.

Allofich:
> I think I found that it was only the index that was stored.

I just tested.
1. Loaded a fresh game (no data saved yet for Brallion)
2. `coc "Sadrith Mora"`
3. Brallion's first package is a 5-hour Wander. Clicked him and `GetCurrentAIPackage` returned 0 (Wander).
4. Waited 4 game hours. Clicked him and `GetCurrentAIPackage` returned 0 (Wander). (Will end in 1 hour)
5. Saved game.
6. Loaded game I saved at step 5.
7. Waited 4 game hours. Clicked Brallion and `GetCurrentAIPackage` still returned 0 (Wander).

This matches what I remember. The original game does not save remaining duration in the save file, which is the only property of an editor-placed package that can change during the game. It only saves the index.

Also, while waiting, the Follow target guard "teleported" next to Brallion, and he began to follow him. So the bug report's behavior exists in the original game.

akortunov:
> Brallion in Sadrith Mora. In original Morrowind he also runs the follow package.

Hmmm... It seems you are right. Do you suggest to do not store remaining duration in the save file, or leave the current behaviour?

Allofich:
> Do you suggest to do not store remaining duration in the save file, or leave the current behaviour?

Well, I think the goal of a saved game file is to preserve as much information of the current game state as possible, right? It seems better to have that information than not.

However, if the AI package "rotating" as it is implemented in OpenMW could be replaced by simply keeping track of an index and the remaining duration of the current package, maybe that would allow simpler code? There would be the problem of breaking compatibility with existing saved game files, though.

Allofich:
Also, the original engine does leave AI package information (not sure about duration) if you add the package with script or console. Rather than saving information differently for editor-placed and script/console-placed, maybe it is better to save the package information in both cases as OpenMW does now.

Allofich:
> if, for example, OpenMW allows any Telvanni Guard to be followed (I haven't tested OpenMW behavior)

Testing with a fairly new OpenMW (January), this doesn't seem to be the case. Brallion didn't react to any of the 4 Telvanni Guards I found (Construction Kit says there are 6 guards total). Probably, as with the original engine, he will only react to one of them. (The one he chooses to follow might be different, though.)

Anyway, there doesn't seem to be any real problem.

PR #2341, opened 2019-04-22
Allow using functions in non-conditional expressions (bug #3725)

Capostrophic:
[Bug 3725](https://gitlab.com/OpenMW/openmw/issues/3725)

The idea: out of the console, when naked expressions are not allowed, code is still generated for functions so that their possible side effects - such as OnActivate disabling normal activation - can be handled without using the return value. This is similar to how instructions are handled. GetDisabled and GetDistance keywords that are not proper functions are safely ignored with logging a parser warning.

Obviously many of the functions are going to be no-op in non-conditional expressions but it's better than breaking the whole script and possibly mod.


Capostrophic:
False alarm, I looked at the wrong branch.

PR #2342, opened 2019-04-22
[Testing needed] Do not use the CPU affinity of OSG viewer

akortunov:
Fixes [bug #4785](https://gitlab.com/OpenMW/openmw/issues/4785).

Using custom affinity may lead to stuttering since several threads can be assigned to the same CPU core. OSG-on-steroids workarounds this bug, but it still present in the upstream OSG.

For example, with two navigator threads I get 25 FPS with master+OSG-3.6.3 and 90-100 FPS with PR+OSG-3.6.3 or master+OSG-on-steroids.

I tested it on Linux, but it still should be tested on Windows and MacOS X.

Note: the `setUseConfigureAffinity` was added in the OSG 3.5.5, so this approach will not work with older stock OSG versions.

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -674,6 +674,11 @@ void OMW::Engine::go()
     mViewer = new osgViewer::Viewer;
     mViewer->setReleaseContextAtEndOfFrameHint(false);
 
+#if OSG_VERSION_GREATER_OR_EQUAL(3,5,5)
+    // Do not try to outsmart the OS thread scheduler (see bug #4785).
+    mViewer->setUseConfigureAffinity(false);
+#endif
+
     mScreenCaptureOperation = new WriteScreenshotToFileOperation(mCfgMgr.getUserDataPath().string(),
         Settings::Manager::getString("screenshot format", "General"));
 


terabyte25:
Does this need to be tested on Android as well?

akortunov:
> Does this need to be tested on Android as well?

It would be nice, but Android users usually do not use navmeshes and preloading anyway.

terabyte25:
Most Android users do use navmeshes, they were only disabled in my builds.

terabyte25:
Without - 140 FPS

With - 150 FPS

akortunov:
> I never bothered to compile with that version.

Stock 3.6.3 without patches causes crashes with OpenMW.

Capostrophic:
3.6 branch in general isn't broken anymore in regards to onw.

I'm pretty sure the sun flare doesn't work properly there for one, though, so use 3.4.1 instead for Android builds.

terabyte25:
@Capostrophic 3.4.1's "osg on steriods" fork has broken white textures on Android.

akortunov:
> I'm pretty sure the sun flare doesn't work properly there for one

Works for me (on a 3.6.0, at least).

psi29a:
small request... can we link back to the bug (on gitlab) that triggered the need for this? That way when we go back through the code later, we know what we touched in why?

I would like that we get into the habit of doing this for bug fixes, that way we have context about why we have hacks/workarounds.

akortunov:
> can we link back to the bug (on gitlab) that triggered the need for this? 

Done.

PR #2343, opened 2019-04-22
Tweak object VFX scaling (bug #4989)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/4989)

I'm still not quite sure about object scaling but I think it should be fine as it is.
List of changes:
1. addEffect now handles effect scaling so that any VFX can be scaled according to the bounding box and not just spellcasting VFX. Most notably it fixes Dagoth Ur shield VFX rendering.
2. VFX are no longer scaled twice according to the NPC height (Morrowind doesn't do that and they already looked fine without the additional scaling because of the scaling of the NPC root node the effects are added to).
3. The scaling is more accurate to vanilla: instead of using arbitrarily chosen divisors added, actual divisors used by vanilla are used which are based on yards - this was determined by exploring the vanilla scenegraph.
4. The scaling code was made much shorter.

Objects without animation (disabled, invisible objects) share the code for creatures/non-actors with animation.

This should make the scaling behavior make much more sense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@
     Bug #4979: AiTravel maximum range depends on "actors processing range" setting
     Bug #4980: Drowning mechanics is applied for actors indifferently from distance to player
     Bug #4984: "Friendly hits" feature should be used only for player's followers
+    Bug #4989: Object dimension-dependent VFX scaling behavior is inconsistent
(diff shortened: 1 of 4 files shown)


akortunov:
LGTM. Merging for a more wide testing...

PR #2344, opened 2019-04-24
Add a button to delete spells from spellwindow

xyzz:
image

I've added a button in bottom right to delete currently active spell. The behavior's the same as shift-clicking a spell, so it would pop up a warning dialog, etc. This is mostly useful on Android where there's no shift key to press, but also on PC for people who don't know they can press shift to delete spells.

### DIFF
--- a/apps/openmw/mwgui/spellmodel.cpp
+++ b/apps/openmw/mwgui/spellmodel.cpp
@@ -153,8 +153,10 @@ namespace MWGui
         ModelIndex selected = -1;
         for (SpellModel::ModelIndex i = 0; i<int(getItemCount()); ++i)
         {
-            if (getItem(i).mSelected)
+            if (getItem(i).mSelected) {
                 selected = i;
+                break;
+            }
         }
         return selected;
     }
--- a/apps/openmw/mwgui/spellmodel.cpp
+++ b/apps/openmw/mwgui/spellmodel.cpp
@@ -148,6 +148,17 @@ namespace MWGui
         return mSpells.size();
     }
 
+    SpellModel::ModelIndex SpellModel::getSelectedIndex() const
+    {
+        ModelIndex selected = -1;
+        for (SpellModel::ModelIndex i = 0; i<int(getItemCount()); ++i)
+        {
+            if (getItem(i).mSelected)
+                selected = i;
+        }
+        return selected;
+    }
+
     Spell SpellModel::getItem(ModelIndex index) const
     {
         if (index < 0 || index >= int(mSpells.size()))
--- a/files/mygui/openmw_spell_window.layout
+++ b/files/mygui/openmw_spell_window.layout
@@ -16,7 +16,7 @@
         <Widget type="Widget" position="8 535 268 23" align="Left Bottom HStretch">
             <!-- Spell deletion button -->
             <Widget type="Button" skin="MW_Button" align="Right Bottom" position="150 0 118 24" name="DeleteSpellButton">
-                <Property key="Caption" value="Delete Spell"/>
+                <Property key="Caption" value="#{sDeleteSpell}"/>
                 <Property key="NeedKey" value="false"/>
             </Widget>
 
--- a/files/mygui/openmw_spell_window.layout
+++ b/files/mygui/openmw_spell_window.layout
@@ -13,17 +13,17 @@
         <Widget type="SpellView" skin="MW_SpellView" position="8 38 268 490" align="Left Top Stretch" name="SpellView">
         </Widget>
 
-        <Widget type="Widget" position="8 535 268 23" align="Left Bottom HStretch">
+        <Widget type="HBox" position="8 535 268 23" align="Right Bottom HStretch">
+            <Widget type="Spacer"/>
             <!-- Spell deletion button -->
-            <Widget type="AutoSizedButton" skin="MW_Button" align="Right Bottom" position="268 0 0 24" name="DeleteSpellButton">
-                <Property key="ExpandDirection" value="Left"/>
+            <Widget type="AutoSizedButton" skin="MW_Button" align="Right Bottom" position="8 535 0 23" name="DeleteSpellButton">
                 <Property key="Caption" value="#{sDelete}"/>
                 <Property key="NeedKey" value="false"/>
             </Widget>
+        </Widget>
 
-            <!-- Search box-->
-            <Widget type="EditBox" skin="MW_TextBoxEditWithBorder" position="0 0 100 23" align="Left Bottom HStretch" name="FilterEdit">
-            </Widget>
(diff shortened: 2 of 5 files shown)


files/mygui/openmw_spell_window.layout
```diff
@@ -13,8 +13,16 @@
         <Widget type="SpellView" skin="MW_SpellView" position="8 38 268 490" align="Left Top Stretch" name="SpellView">
         </Widget>
 
-        <!-- Search box-->
-        <Widget type="EditBox" skin="MW_TextBoxEditWithBorder" position="8 535 268 23" align="Left Bottom HStretch" name="FilterEdit">
+        <Widget type="Widget" position="8 535 268 23" align="Left Bottom HStretch">
+            <!-- Spell deletion button -->
+            <Widget type="Button" skin="MW_Button" align="Right Bottom" position="150 0 118 24" name="DeleteSpellButton">
+                <Property key="Caption" value="Delete Spell"/>
```
terabyte25: Use GMSTs instead of non localizable strings when you can. sDeleteSpell is applicable for the string value here, so use that instead.

apps/openmw/mwgui/spellmodel.cpp
```diff
@@ -148,6 +148,17 @@ namespace MWGui
         return mSpells.size();
     }
 
+    SpellModel::ModelIndex SpellModel::getSelectedIndex() const
+    {
+        ModelIndex selected = -1;
+        for (SpellModel::ModelIndex i = 0; i<int(getItemCount()); ++i)
+        {
+            if (getItem(i).mSelected)
+                selected = i;
```
akortunov: I suppose there should be "break". Or maybe you can just store selected index in the variable and update it in the SpellModel::update().

files/mygui/openmw_spell_window.layout
```diff
@@ -13,8 +13,16 @@
         <Widget type="SpellView" skin="MW_SpellView" position="8 38 268 490" align="Left Top Stretch" name="SpellView">
         </Widget>
 
-        <!-- Search box-->
-        <Widget type="EditBox" skin="MW_TextBoxEditWithBorder" position="8 535 268 23" align="Left Bottom HStretch" name="FilterEdit">
+        <Widget type="Widget" position="8 535 268 23" align="Left Bottom HStretch">
+            <!-- Spell deletion button -->
+            <Widget type="Button" skin="MW_Button" align="Right Bottom" position="150 0 118 24" name="DeleteSpellButton">
+                <Property key="Caption" value="#{sDeleteSpell}"/>
```
akortunov: sDeleteSpell is too long for such button, at least inRussian translation ("Удалить заклинание"). Probably the sDelete ("Удалить") will be enough here.

apps/openmw/mwgui/spellmodel.cpp
xyzz: I've added a break

files/mygui/openmw_spell_window.layout
xyzz: I've made it longer to fit the translation

files/mygui/openmw_spell_window.layout
```diff
@@ -15,13 +15,13 @@
 
         <Widget type="Widget" position="8 535 268 23" align="Left Bottom HStretch">
             <!-- Spell deletion button -->
-            <Widget type="Button" skin="MW_Button" align="Right Bottom" position="150 0 118 24" name="DeleteSpellButton">
+            <Widget type="Button" skin="MW_Button" align="Right Bottom" position="108 0 160 24" name="DeleteSpellButton">
```
akortunov: I did not mean that.
First of all, there is an AutoSizedButton widget, which adjusts its size depending on its label size.
Second, the sDeleteSpell is just too long for such button - it should be a quite small and non-intrusive.

terabyte25: So use sDelete instead and use an autoresize box?

akortunov: I am afraid that it will not be so trivial.

xyzz: so what do you want me to do?

akortunov: 1. Use the AutoSizedButton.
2. Use the sDelete instead of sDeleteSpell.
3. Adjust the FilterEdit width properly. More likely you will have to do it manually in the spellwindow.cpp.

akortunov:
I thought about simplier approach. Something like:

```
        <Widget type="HBox" position="8 535 268 23" align="Right Bottom HStretch">
            <Widget type="Spacer"/>
            <!-- Spell deletion button -->
            <Widget type="AutoSizedButton" skin="MW_Button" align="Right Bottom" position="8 535 0 23" name="DeleteSpellButton">
                <Property key="Caption" value="#{sDelete}"/>
                <Property key="NeedKey" value="false"/>
            </Widget>
        </Widget>

        <!-- Search box-->
        <Widget type="EditBox" skin="MW_TextBoxEditWithBorder" position="8 535 268 23" align="Left Bottom HStretch" name="FilterEdit">
        </Widget>
```

And:
```
    SpellWindow::SpellWindow(DragAndDrop* drag)
        : WindowPinnableBase("openmw_spell_window.layout")
        , NoDrop(drag, mMainWidget)
        , mSpellView(nullptr)
        , mUpdateTimer(0.0f)
    {
        ...

        setCoord(498, 300, 302, 300);

        // Adjust the spell filtering widget size because of MyGUI limitations.
        MyGUI::Widget* deleteButton;
        getWidget(deleteButton, "DeleteSpellButton");
        int filterWidth = mSpellView->getSize().width - deleteButton->getSize().width - 3;
        mFilterEdit->setSize(filterWidth, mFilterEdit->getSize().height);
    }
```

In theory, you need to set the size only once, and MyGUI should handle the rest automatically

xyzz:
Sure, that works.

akortunov:
Well, basically it works now. The only drawback I see - you can remove only currently selected spell via that button, but I do not know what to do with it.

Let see what other developers think about this PR.

zimmi:
Can the label be named "Forget Spell" instead? "Delete" sounds very database-y.

xyzz:
Is there a string resource for that?

terabyte25:
@zimmi @xyzz https://wiki.openmw.org/index.php?title=GMSTs_(status)

Pls.

terabyte25:
Just remembered controller people have no real way of deleting spells, so +1 from me.

zimmi:
I somehow thought this was OpenMW-only without a corresponding Morrowind feature / translation, sorry for the noise.

terabyte25:
@zimmi Yeah, sorry for being a bit rash.

PR #2345, opened 2019-04-25
More argument usage corrections

Capostrophic:
More research led to additional findings about some more functions that commonly "accept" extra arguments for no benefit. OpenMW freaks out and then the mods using the no-op syntax break. Sometimes catastrophically.

1. Someone used a cell string argument for AITravel.
2. Someone used a reference ID string argument for GetHealth.
3. Someone used faction string arguments for LowerRank and RaiseRank.
4. Someone used redundant numerical arguments with unknown meaning for ForceGreeting (the guess is that they attempted to force a certain greeting group to be used by the actor).
5. Position and PlaceItem accepted a zero (number) after them and that was used in some cases. Cell variants of the functions also accept the redundant argument for consistency (I think I've seen a case where PlaceItemCell was fed a zero).

These are all now ignored (with logging a parser warning).

6. Requested by akortunov: invalid axis arguments for various transformation instructions are now handled more gracefully. Morrowind simply does nothing if these are used, while we throw an exception (which makes the script broken). Now there isn't any exception but simply nothing happens (these commands are often run per-frame, so logging their invalid execution would be pretty spammy).

### DIFF
--- a/apps/openmw/mwscript/transformationextensions.cpp
+++ b/apps/openmw/mwscript/transformationextensions.cpp
@@ -108,8 +108,6 @@ namespace MWScript
                         MWBase::Environment::get().getWorld()->rotateObject(ptr,ax,angle,az);
                     else if (axis == "z")
                         MWBase::Environment::get().getWorld()->rotateObject(ptr,ax,ay,angle);
-                    else
-                        throw std::runtime_error ("invalid rotation axis: " + axis);
                 }
         };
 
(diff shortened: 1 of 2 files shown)


Capostrophic:
>the command silently returns data for object, to which the script is attached.

It's logged as a parser warning. Presumably Morrowind handles it just fine even though the function doesn't return what is intended to be returned.

akortunov:
LGTM.

PR #2346, opened 2019-04-25
Make sure forcePPL and clamp are defined in the editor

Capostrophic:
After somewhat recent changes to the initialization of these shader defines editor was logging "stuff undefined" errors for akortunov, which meant that the editor would fail to create shaders and fall back to fixed function pipeline. AnyOldName3 recommended to put the initialization of these defines for the editor here specifically. Now what gets used is the previous defaults.

### DIFF
--- a/apps/opencs/model/world/data.cpp
+++ b/apps/opencs/model/world/data.cpp
@@ -11,6 +11,7 @@
 #include <components/esm/cellref.hpp>
 
 #include <components/resource/scenemanager.hpp>
+#include <components/shader/shadermanager.hpp>
 #include <components/vfs/manager.hpp>
 #include <components/vfs/registerarchives.hpp>
 
@@ -74,6 +75,11 @@ CSMWorld::Data::Data (ToUTF8::FromType encoding, bool fsStrict, const Files::Pat
 
     mResourcesManager.setVFS(mVFS.get());
     mResourceSystem.reset(new Resource::ResourceSystem(mVFS.get()));
+
+    Shader::ShaderManager::DefineMap defines = mResourceSystem->getSceneManager()->getShaderManager().getGlobalDefines();
+    defines["forcePPL"] = "0";
+    defines["clamp"] = "1";
+    mResourceSystem->getSceneManager()->getShaderManager().setGlobalDefines(defines);
 
     mResourceSystem->getSceneManager()->setShaderPath((resDir / "shaders").string());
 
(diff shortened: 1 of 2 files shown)


akortunov:
Now it complains about "Undefined shadows_enabled".

akortunov:
And about "shadow_texture_unit_list"

akortunov:
Additional defines:
```
    defines["shadows_enabled"] = "0";
    defines["shadow_texture_unit_list"] = "";
    defines["shadowMapsOverlap"] = "0";
    defines["useShadowDebugOverlay"] = "0";
    defines["perspectiveShadowMaps"] = "0";
    defines["disableNormalOffsetShadows"] = "0";
    defines["shadowNormalOffset"] = "0.0";
```

Basically you need to get a getShadowsDisabledDefines() as well.

Capostrophic:
Yeah, yeah, done that independently.

akortunov:
Seems to work for me now, despite I'd prefer to avoid code duplication.

Capostrophic:
If I were to avoid code duplication, then it would be necessary to initialize the entirety of the shadow manager (not even possible with the data available in the scope) that as I said won't get used for anything but recovering these specific define names.

akortunov:
It is how we handle them now:

```
        Shader::ShaderManager::DefineMap shadowDefines = mShadowManager->getShadowDefines();
        Shader::ShaderManager::DefineMap globalDefines = mResourceSystem->getSceneManager()->getShaderManager().getGlobalDefines();

        for (auto itr = shadowDefines.begin(); itr != shadowDefines.end(); itr++)
            globalDefines[itr->first] = itr->second;

        globalDefines["forcePPL"] = Settings::Manager::getBool("force per pixel lighting", "Shaders") ? "1" : "0";
        globalDefines["clamp"] = Settings::Manager::getBool("clamp lighting", "Shaders") ? "1" : "0";

        // It is unnecessary to stop/start the viewer as no frames are being rendered yet.
        mResourceSystem->getSceneManager()->getShaderManager().setGlobalDefines(globalDefines);
```

akortunov:
> If I were to avoid code duplication, then it would be necessary to initialize the entirety of the shadow manager (not even possible with the data available in the scope) that as I said won't get used for anything but recovering these specific define names.

Well, LGTM then.

AnyOldName3:
I'd be happier making `getShadowsDisabledDefines()` static than having duplication. I can't really think of any reason why it would need to be non-static as even if we end up having two shadow systems, we're going to need the defines to disable all of them as they'll still be present in the shader source.

akortunov:
@Capostrophic, will you try to implement a static method for defines?

Capostrophic:
I am willing to do that. Where should it be put though? Should it be left in ShadowManager namespace?

AnyOldName3:
I'd leave it as a class member, yeah.

Capostrophic:
Does this look right?

akortunov:
> Does this look right?

Seems to work.

AnyOldName3:
LGTM

PR #2347, opened 2019-04-25
Make pinnable windows hiding persistent

akortunov:
Implements [feature #4994](https://gitlab.com/OpenMW/openmw/issues/4994).

Just use the same approach as for "Pinned" status - store it in the settings.cfg and restore during UI initialization.


Capostrophic:
Seems to work fine and looks good to me.

PR #2349, opened 2019-04-26
Allow to maximize windows via Shift + Double Click

akortunov:
Implements [feature #3999](https://gitlab.com/OpenMW/openmw/issues/3999).

Most of windows use the whole screen when maximized (excepts of inventory-related ones - they occupy only the half of screen).

### DIFF
--- a/apps/openmw/mwgui/exposedwindow.cpp
+++ b/apps/openmw/mwgui/exposedwindow.cpp
@@ -2,12 +2,12 @@
 
 namespace MWGui
 {
-    MyGUI::VectorWidgetPtr ExposedWindow::getSkinWidgetsByName (const std::string &name)
+    MyGUI::VectorWidgetPtr Window::getSkinWidgetsByName (const std::string &name)
     {
         return MyGUI::Widget::getSkinWidgetsByName (name);
     }
 
-    MyGUI::Widget* ExposedWindow::getSkinWidget(const std::string & _name, bool _throw)
+    MyGUI::Widget* Window::getSkinWidget(const std::string & _name, bool _throw)
     {
         MyGUI::VectorWidgetPtr widgets = getSkinWidgetsByName (_name);
 
--- a/apps/openmw/mwgui/inventorywindow.cpp
+++ b/apps/openmw/mwgui/inventorywindow.cpp
@@ -162,28 +162,39 @@ namespace MWGui
         mItemView->setModel(nullptr);
     }
 
+    void InventoryWindow::toggleMaximized()
+    {
+        std::string setting = getModeSetting();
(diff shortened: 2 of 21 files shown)


apps/openmw/mwgui/windowpinnablebase.cpp
```diff
@@ -1,6 +1,9 @@
 #include "windowpinnablebase.hpp"
 
-#include <MyGUI_Button.h>
+#include <MyGUI_InputManager.h>
+
+#include "../mwbase/windowmanager.hpp"
+#include "../mwbase/environment.hpp"
```
Capostrophic: These look redundant.

apps/openmw/mwgui/exposedwindow.cpp
```diff
@@ -2,12 +2,12 @@
 
 namespace MWGui
 {
-    MyGUI::VectorWidgetPtr ExposedWindow::getSkinWidgetsByName (const std::string &name)
+    MyGUI::VectorWidgetPtr Window::getSkinWidgetsByName (const std::string &name)
```
Capostrophic: Why all the renames?

akortunov: The point is to replace implementation for the basic Window class. Otherwise I need turn everything for ExposedWindows - a lot of mess with layout files.

Probably it is possible implement a custom button class instead of using custom layout properties, but I am not sure that is is better.

Capostrophic:
Some brief testing didn't trigger anything obvious.

Capostrophic:
"Maximized" setting is perhaps too trigger-happy. Specifically the added setBool call in onWindowResize. Even if you *never* used Shift-Click *any* position/scale change of the window will add the setting into the user config file. Maybe it should only be changed if it was there to begin with (that is, "true")?

apps/openmw/mwgui/inventorywindow.cpp
```diff
@@ -162,28 +162,42 @@ namespace MWGui
...
-            case GM_Inventory:
-            default:
-                setPinButtonVisible(true);
-                break;
-        }
+        std::string setting = getModeSetting();
+        if (mode == GM_Inventory)
+            setPinButtonVisible(true);
+        else
+            setPinButtonVisible(false);
```
Capostrophic: Maybe just go with
```cpp
setPinButtonVisible(mode == GM_Inventory);
```

akortunov:
> Maybe it should only be changed if it was there to begin with (that is, "true")?

OK, I reset setting now only if it is "true".

akortunov:
Any objections?

PR #2353, opened 2019-04-28
Make Drop behavior closer to vanilla (bug #4999)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/4999).

1. Drop unnecessary container support, make the instruction actor-exclusive.
2. Play down sound of the item on the actor. Sound manager dependency was added, since playing it as a 2D sound via window manager is not sufficient (the sound can as well play on slaves when their shackles are dropped).
3. Spawn an item stack manually if there isn't enough items in the actor's inventory. Scripted items are spawned in separate stacks. manualref dependency had to be added, since that's what is used for generating new references.

### DIFF
--- a/apps/openmw/mwscript/miscextensions.cpp
+++ b/apps/openmw/mwscript/miscextensions.cpp
@@ -15,6 +15,7 @@
 #include "../mwbase/environment.hpp"
 #include "../mwbase/windowmanager.hpp"
 #include "../mwbase/scriptmanager.hpp"
+#include "../mwbase/soundmanager.hpp"
 #include "../mwbase/world.hpp"
 
 #include "../mwworld/class.hpp"
@@ -23,6 +24,7 @@
 #include "../mwworld/inventorystore.hpp"
 #include "../mwworld/esmstore.hpp"
 #include "../mwworld/cellstore.hpp"
+#include "../mwworld/manualref.hpp"
 
 #include "../mwmechanics/aicast.hpp"
 #include "../mwmechanics/npcstats.hpp"
@@ -469,6 +471,9 @@ namespace MWScript
                     std::string gem = runtime.getStringLiteral (runtime[0].mInteger);
                     runtime.pop();
 
+                    if (!ptr.getClass().hasInventoryStore(ptr))
+                        return;
+
                     const MWWorld::ESMStore& store = MWBase::Environment::get().getWorld()->getStore();
                     store.get<ESM::Creature>().find(creature); // This line throws an exception if it can't find the creature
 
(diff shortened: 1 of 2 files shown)


apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -534,54 +536,65 @@ namespace MWScript
...
                     for (MWWorld::ContainerStoreIterator iter (store.begin()); iter!=store.end(); ++iter)
                     {
-                        if (::Misc::StringUtils::ciEqual(iter->getCellRef().getRefId(), item)
-                                && (!invStorePtr || !invStorePtr->isEquipped(*iter)))
+                        if (::Misc::StringUtils::ciEqual(iter->getCellRef().getRefId(), item) && !store.isEquipped(*iter))
                         {
                             int removed = store.remove(*iter, toRemove, ptr);
                             MWWorld::Ptr dropped = MWBase::Environment::get().getWorld()->dropObjectOnGround(ptr, *iter, removed);
-                            dropped.getCellRef().setOwner("");
+                            dropped.getCellRef().setOwner(std::string());
```
akortunov: It is incosistent with other setOwner usages - we use the "" at least in three different places.

akortunov:
What's about DropSoulGem?

apps/openmw/mwscript/miscextensions.cpp
Capostrophic: Ok.

Capostrophic:
Soulgem instructions don't work on non-actors and DropSoulgem doesn't work if there isn't a relevant soulgem in the actor's inventory, that's all I can say and have changed now.

apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -469,6 +471,9 @@ namespace MWScript
                     std::string gem = runtime.getStringLiteral (runtime[0].mInteger);
                     runtime.pop();
 
+                    if (!ptr.getClass().hasInventoryStore(ptr))
```
akortunov: According to my testing, this commands can be applied for any creature in Morrowind, not just for ones with inventory.
Also Morrowind seems to spawn items near player, not near target actor.

From what I can tell, Drop uses the same approach to place items as a "player->placeatme" in Morrowind, but I am not sure if it worth to replicate.

Capostrophic: >this commands can be applied for any creature in Morrowind, not just for ones with inventory.

Ok.
>Also Morrowind seems to spawn items near player, not near target actor.

Vanilla bug.
>but I am not sure if it worth to replicate.

Nuances, nuances. It's not the same as PlaceAt because there isn't a new reference generated in Morrowind if there *is* an instance of an item in the inventory. It could copy the first instance of the item in theory, but the understanding isn't clear enough.

akortunov:
LGTM.

PR #2355, opened 2019-04-29
Use the StateSetUpdater to modify alpha for Animation objects

akortunov:
Should fix [bug #5001](https://gitlab.com/OpenMW/openmw/issues/5001).

An approach, suggested by bzzt, leads to performance loss since we recreate the whole stateset every time we need to change the alpha level.

I use the StateSetUpdater instead, so performance is the same as in master branch.

### DIFF
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -460,6 +460,46 @@ namespace
 
 namespace MWRender
 {
+    class TransparencyUpdater : public SceneUtil::StateSetUpdater
+    {
+    public:
+        TransparencyUpdater(const float alpha)
+            : mAlpha(alpha)
+        {
+        }
+
+        void setAlpha(const float alpha)
+        {
+            mAlpha = alpha;
+        }
+
+    protected:
+        virtual void setDefaults(osg::StateSet* stateset)
+        {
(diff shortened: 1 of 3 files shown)


akortunov:
@xyzz, I suggest you to take a look too since you usually report such kind of issues.
Can you check actors fading with tsan?

AnyOldName3:
LGTM, but I haven't tested it.

apps/openmw/mwrender/animation.cpp
```diff
@@ -460,6 +460,46 @@ namespace
...
+        }
+
+        virtual void apply(osg::StateSet* stateset, osg::NodeVisitor* /*nv*/)
+        {
+            osg::Material* material = static_cast<osg::Material*>(stateset->getAttribute(osg::StateAttribute::MATERIAL));
+            material->setAlpha(osg::Material::FRONT_AND_BACK, mAlpha);
+        }
+
+    private:
+        float mAlpha;
```
xyzz: This isn't thread-safe, you should be using `std::atomic<float>` if you're updating and using the float from different threads

akortunov: The whole animation object is not supposed to be used in different threads, also we do not use atomic in other stateset updaters.

xyzz: If it's not used in different threads, then that's fine.

AnyOldName3:
@OpenMW/core last call

PR #2358, opened 2019-04-30
engine: fix window creation when SDL returns different-sized window

xyzz:
If sdl2 returns a different size from what was requested when creating a window, there would be inconsistency between size of the actual window and the opengl viewport used by osg. This resulted in graphical glitches on android.

Would appreciate if people could test it on different os/windowed/fullscreen combinations.


akortunov:
Seems to work on Linux in both windowed and fullscreen modes.

psi29a:
Getting it into master so that windows people, who follow nightlies, can test it.

PR #2360, opened 2019-05-01
Do not track resolution settings changes in the InputManager

akortunov:
Fixes [bug #4902](https://gitlab.com/OpenMW/openmw/issues/4902).

The InputManager::windowResized() is actually a side effect of resolution change, so we get unhandled "resolution x" and "resolution y" settings changes in the Settings::Manager::mChangedSettings. If a user changes an another setting after resolution change, OpenMW changes resolution again after Settings::Manager::apply() and generates an another set of "windowResized" events and so on.

So the main idea - just change setting here, without adding it to the mChangedSettings set.
In this case there will be no additional setting changes for the Settings::Manager::apply(), which may accidentally change resolution.

### DIFF
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -826,9 +826,6 @@ namespace MWInput
                                         Settings::Manager::getInt("resolution y", "Video"),
                                         Settings::Manager::getBool("fullscreen", "Video"),
                                         Settings::Manager::getBool("window border", "Video"));
-
-            // We should reload TrueType fonts to fit new resolution
-            MWBase::Environment::get().getWindowManager()->loadUserFonts();
         }
     }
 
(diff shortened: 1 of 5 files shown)


akortunov:
Note: if additional arguments are bad, I can add an overloaded apply() function to apply only one setting.

apps/openmw/mwinput/inputmanagerimp.cpp
```diff
@@ -1129,10 +1126,15 @@ namespace MWInput
 
     void InputManager::windowResized(int x, int y)
     {
-        Settings::Manager::setInt("resolution x", "Video", x);
-        Settings::Manager::setInt("resolution y", "Video", y);
+        // Note: it is a side affect of resolution change or window resize.
```
Capostrophic: "This is a side effect"

Capostrophic:
>I can add an overloaded apply() function to apply only one setting.

I dislike this but it's the lesser evil and you probably should.

akortunov:
> I dislike this but it's the lesser evil and you probably should.

Done.

components/settings/settings.cpp
```diff
@@ -414,6 +414,12 @@ void Manager::setBool(const std::string &setting, const std::string &category, c
     setString(setting, category, value ? "true" : "false");
 }
 
+void Manager::apply(const std::string &setting, const std::string &category)
```
Capostrophic: I guess I should have said it but such a function name is misleading. It doesn't actually apply anything, it removes pending setting changes from the tracker vector. I'm not sure what could be a better name, though.

akortunov: The same thing for existing "apply" function - it does not apply anything too.

PR #2368, opened 2019-05-05
Avoid using null pointer for player's cell in moveObjectImp

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5015)
Make sure isExterior is not called on a null player CellStore pointer, which should fix potential crashes.


akortunov:
I managed to reproduce this crash on 0.46 using these commands in a startup script:

```
player->Position, 2816.406, 7016.756, 4306.963, 180
hul->Position, 2816.406, 7016.756, 4306.963, 180
```

This PR seems to fix this issue.
But I can not reproduce the crash in the 0.45, so it seems this bug a regression.

akortunov:
LGTM.

PR #2372, opened 2019-05-07
Fix some thread sync issues

akortunov:
Summary of changes:
1. Fix possible data race when some threads try to modify a static local string variable.
2. OSG viewer keeps the reference to the WorkQueue, so we actually stop background (e.g. preloading) tasks after we destroy the ResourceManager, which is not right. 
The current solution is to destroy viewer before ResourceManager, so work items will be stopped before resources destroying.
3. We sort data in the Land store every time when we load savegame, but this store is supposed to be static to do not conflict with the global map rendering task ([bug #4844](https://gitlab.com/OpenMW/openmw/issues/4844)).
Since we do not change terrain on the fly, we can just sort data only once during game start before lauch the global map rendering task.
4. Setup a per-frame compile limit for ICO only once during setup (fixes [bug #5025](https://gitlab.com/OpenMW/openmw/issues/5025)). I suppose that initially the limit was designed to be dynamic depending on framerate, but this approach was changed anyway a long time ago.
5. Make integer fields of PacketQueue atomic to avoid data races - this struct is accessed from different threads and we already use atomic for `flushing` field.

### DIFF
--- a/components/esmterrain/storage.cpp
+++ b/components/esmterrain/storage.cpp
@@ -353,11 +353,11 @@ namespace ESMTerrain
     std::string Storage::getTextureName(UniqueTextureId id)
     {
         // Goes under used terrain blend transitions
-        static const std::string baseTexture = "textures\\tx_black_01.dds";
+        static constexpr char baseTexture[] = "textures\\tx_black_01.dds";
         if (id.first == -1)
             return baseTexture;
 
-        static const std::string defaultTexture = "textures\\_land_default.dds";
+        static constexpr char defaultTexture[] = "textures\\_land_default.dds";
         if (id.first == 0)
             return defaultTexture; // Not sure if the default texture really is hardcoded?
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
     Bug #4828: Potion looping effects VFX are not shown for NPCs
     Bug #4837: CTD when a mesh with NiLODNode root node with particles is loaded
     Bug #4841: Russian localization ignores implicit keywords
+    Bug #4844: Data race in savegame loading / GlobalMap render
     Bug #4847: Idle animation reset oddities
     Bug #4851: No shadows since switch to OSG
     Bug #4860: Actors outside of processing range visible for one frame after spawning
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@
     Bug #5001: Possible data race in the Animation::setAlpha()
     Bug #5004: Werewolves shield their eyes during storm
     Bug #5018: Spell tooltips don't support purely negative magnitudes
+    Bug #5025: Data race in the ICO::setMaximumNumOfObjectsToCompilePerFrame()
     Bug #5028: Offered price caps are not trading-specific
     Bug #5038: Enchanting success chance calculations are blatantly wrong
     Feature #1774: Handle AvoidNode
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -257,9 +257,9 @@ OMW::Engine::~Engine()
 
     mWorkQueue = nullptr;
 
-    mResourceSystem.reset();
-
     mViewer = nullptr;
+
+    mResourceSystem.reset();
 
     delete mEncoder;
     mEncoder = nullptr;
--- a/apps/openmw/mwgui/loadingscreen.cpp
+++ b/apps/openmw/mwgui/loadingscreen.cpp
@@ -165,11 +165,6 @@ namespace MWGui
         if (mMainWidget->getVisible())
             return;
 
-        if (mViewer->getIncrementalCompileOperation())
-        {
-            mViewer->getIncrementalCompileOperation()->setMaximumNumOfObjectsToCompilePerFrame(100);
-        }
-
         // Assign dummy bounding sphere callback to avoid the bounding sphere of the entire scene being recomputed after each frame of loading
         // We are already using node masks to avoid the scene from being updated/rendered, but node masks don't work for computeBound()
         mViewer->getSceneData()->setComputeBoundingSphereCallback(new DontComputeBoundCallback);
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -268,6 +268,7 @@ namespace MWRender
         {
             mViewer->setIncrementalCompileOperation(new osgUtil::IncrementalCompileOperation);
             mViewer->getIncrementalCompileOperation()->setTargetFrameRate(Settings::Manager::getFloat("target framerate", "Cells"));
+            mViewer->getIncrementalCompileOperation()->setMaximumNumOfObjectsToCompilePerFrame(100);
         }
 
         mResourceSystem->getSceneManager()->setIncrementalCompileOperation(mViewer->getIncrementalCompileOperation());
--- a/apps/openmw/mwworld/store.cpp
+++ b/apps/openmw/mwworld/store.cpp
@@ -500,7 +500,12 @@ namespace MWWorld
     }
     void Store<ESM::Land>::setUp()
     {
+        // The land is static for given game session, there is no need to refresh it every load.
+        if (mBuilt)
+            return;
+
         std::sort(mStatic.begin(), mStatic.end(), Compare());
+        mBuilt = true;
     }
 
 
--- a/apps/openmw/mwworld/store.hpp
+++ b/apps/openmw/mwworld/store.hpp
@@ -247,6 +247,8 @@ namespace MWWorld
 
         RecordId load(ESM::ESMReader &esm);
         void setUp();
+    private:
+        bool mBuilt = false;
     };
 
     template <>
--- a/extern/osg-ffmpeg-videoplayer/videostate.hpp
+++ b/extern/osg-ffmpeg-videoplayer/videostate.hpp
@@ -80,8 +80,8 @@ struct PacketQueue {
 
     AVPacketList *first_pkt, *last_pkt;
     std::atomic<bool> flushing;
-    int nb_packets;
-    int size;
+    std::atomic<int> nb_packets;
+    std::atomic<int> size;
 
     OpenThreads::Mutex mutex;
     OpenThreads::Condition cond;


components/esmterrain/storage.cpp
```diff
@@ -353,20 +355,18 @@ namespace ESMTerrain
     std::string Storage::getTextureName(UniqueTextureId id)
     {
         // Goes under used terrain blend transitions
-        static const std::string baseTexture = "textures\\tx_black_01.dds";
         if (id.first == -1)
-            return baseTexture;
+            return mBaseTexture;
```
kcat: Considering `baseTexture` is `static const`, there was no problem here or below with the same construct. Static variables are guaranteed to be initialized in a thread-safe manner, and as the variable's `const` with no mutable members, it can't be modified. It's only read from to make a copy, so there's no race (multiple threads can read the same memory at the same time unsynchronized just fine).

components/sceneutil/workqueue.cpp
```diff
@@ -32,6 +32,8 @@ WorkItem::WorkItem()
 
 WorkItem::~WorkItem()
 {
+    // Make sure that no one locked this work item.
+    OpenThreads::ScopedLock<OpenThreads::Mutex> lock(mMutex);
```
kcat: Honestly, if an object is deleted when it's still in use and needs locking, there's a more fundamental issue regarding the object's lifetime and ownership. Something still has an access handle to it even though it's being freed. Work on an object should be done before it's deleted, not during.

components/sceneutil/workqueue.cpp
```diff
@@ -19,10 +19,8 @@ void WorkItem::waitTillDone()
 
 void WorkItem::signalDone()
 {
-    {
-        OpenThreads::ScopedLock<OpenThreads::Mutex> lock(mMutex);
-        mDone.exchange(1);
-    }
+    OpenThreads::ScopedLock<OpenThreads::Mutex> lock(mMutex);
+    mDone.exchange(1);
```
kcat: I'm not sure this is necessary. I don't know the exact details of the code, but it's a more preferable paradigm to release the lock after setting the checked condition and before signalling the actual `Condition` variable. Like this:
```
// Thread 1
OpenThreads::ScopedLock<OpenThreads::Mutex> lock(mMutex);
while(!mDone)
    mCondition.wait(lock);

// Thread 2
{
    OpenThreads::ScopedLock<OpenThreads::Mutex> lock(mMutex);
    mDone = 1;
}
mCondition.broadcast();
```
This is completely safe (no unsynchronized access to `mDone`) and avoids extraneous wakeups from `broadcast()` causing Thread 1 to wakeup and try to retake the lock while Thread 2 still has the lock, causing Thread 1 to immediately go back to sleep, while Thread 2 then continues on and finally releases the lock. The OS's scheduler really doesn't like threads constantly waking up and going back to sleep like that, especially when it actually has something to do but couldn't because of that microsecond of contention.

components/detournavigator/navmeshmanager.hpp
```diff
@@ -22,6 +22,15 @@ namespace DetourNavigator
...
         NavMeshManager(const Settings& settings);
 
+        ~NavMeshManager()
+        {
+            mAsyncNavMeshUpdater.stop();
+            mCache.clear();
+            mChangedTiles.clear();
+            mPlayerTile.clear();
+            mLastRecastMeshManagerRevision.clear();
+        }
```
kcat: IIRC, a class's members are constructed in the order of declaration, and destructed in reverse order. So as long as `mAsyncNavMeshUpdater` is the last declared data member, it will be destructed first before `mCache` and the others.

akortunov: The idea is th `mAsyncNavMeshUpdater` **is not** the last, and attempt to just move it leads to crashes on exit.

components/sceneutil/workqueue.cpp
```diff
@@ -7,10 +7,10 @@ namespace SceneUtil
 
 void WorkItem::waitTillDone()
 {
+    OpenThreads::ScopedLock<OpenThreads::Mutex> lock(mMutex);
     if (mDone > 0)
```
kcat: `mDone` appears to be an atomic variable, so the lock is unnecessary here. In the worst case, `mDone` is first checked to be 0, causing it to continue on, take the lock, and be checked as non-0 then release the lock and return. Taking the lock earlier doesn't really help, and can actually inhibit slight improvements... changing this first read to use acquire semantics would be faster for cases where the work item has been long done (relatively speaking). Only in cases where it was very recently completed might it "erroneously" fail, but in that case the lock and subsequent recheck would clear it all up.

As it is, with the lock moved like this you may as well just remove this first check since it checks the same thing again immediately with nothing in between. You only removed the pre-lock early-out test.

components/detournavigator/navmeshmanager.hpp
kcat: Okay, fair enough.

components/sceneutil/workqueue.cpp
akortunov: Technically, since the doWork() is public, the `mDone` does not guarantee that the work item really does not do any job now.

akortunov: Can we somehow to make sure that we run every work item only once and only insode the WorkQueue, so the `mDone` is correct?

AnyOldName3:
Anything specific you need me to look at? I've not really done that much with (and in some cases haven't looked at) the stuff you've changed so far.

akortunov:
> Anything specific you need me to look at? I've not really done that much with (and in some cases haven't looked at) the stuff you've changed so far.

Here:
> A lot of messages about possible data race [here](https://github.com/OpenMW/openmw/blob/master/components/resource/objectcache.hpp#L61), if an another thread creates/removes reference to the same object, for which we try to get the referenceCount().

Do you have an idea if it is safe? In theory, `unref` should be threadsafe.

components/sceneutil/workqueue.cpp
kcat: I'm not sure what you mean. `mDone` is only ever set in `WorkItem::signalDone` from what I see. And given that `mDone` is an atomic variable, moving its early-out test under the lock doesn't improve anything since it was already safe to access (and instead just makes the test not early-out, and redundant with the following loop condition).

akortunov: There is nothing to prevent you from doing something like:

        mWorkItem = new CreateMapWorkItem(mWidth, mHeight, mMinX, mMinY, mMaxX, mMaxY, mCellSize, esmStore.get<ESM::Land>());
        mWorkQueue->addWorkItem(mWorkItem);
        mWorkItew->doWork();

In this case `mDone` just guarantees that both threads can update it without issues, It does not guarantee that access to task results is safe.

Also you can re-run already finished task manually. The `mDone` is still 1 in this case, despite the task is running and access to its results is not safe.

akortunov:
All right, let's try to solve it from the other side:

1. Every time when Sandstranger is asked about bugs in the Android port, he blames the engine that it has hundreds and thousands thread sync issues and lazy and incompetent OpenMW devs just ignore them for years, what give us a bad PR.
2. He uses a some kind of ThreadSanitizer to detect thread sync issues.
3. Roughly 75% of tsan error messages are related to our WorkItem class. This class makes analyzers to cry:
3.1. WorkItems can be stored and managed outside of background thread. For example, you can re-run a finished task while another thread reads results of previous execution.
3.2. As a result of 3.1, there is no real guarantee that task is not running during its destruction.
3.3. As a result of both, the work queue itself can not ensure its sanity - it relies on a sane client code.
On a practice, these issues do not lead to troubles because of our coding discipline - we seems to do not do bad things when using work items, but ananyzers are not smart enough to determine it - they spam data race warnings every time when we use work items.
So it is more a code quality issue - component X does not know anything about how it is used, especially if it is used in different threads, so it is nice if it can validate its state and can tell a client if it uses X in a wrong way.

So the main issue is what we can do here (if we should do anything at all).

components/sceneutil/workqueue.cpp
kcat: Calling `doWork` manually after the item's been added to the queue doesn't look at all safe, unless all implementations of `doWork` are guaranteed thread-safe (and guaranteed to not return until the first caller is finished; which is up to the implementation, not this base). The caller has a responsibility to use an API correctly, and if `doWork` isn't thread-safe no one should be calling it in multiple threads.

In either case, `mDone` is only set to non-0 by a call to `signalDone`, which should only be called after `doWork` returns. As such, `waitTillDone` was fine as it was since it can't return before `signalDone` is called, which is after `doWork` is done.

components/esmterrain/storage.cpp
akortunov: > Considering baseTexture is static const, there was no problem here or below with the same construct. 

GCC does not think so.

akortunov: Interesting - it does not compain if I initialize variable during compilation:

```
static constexpr const char* baseTexture = "textures\\tx_black_01.dds";
```

kcat:
> Every time when Sandstranger is asked about bugs in the Android port, he blames the engine that it has hundreds and thousands thread sync issues and lazy and incompetent OpenMW devs just ignore them for years, what give us a bad PR.

We can't fix bugs we don't know about. If Sandstranger knows for sure that the engine is at fault for something, he should be able to tell us clearly what's wrong (not vague statements like "thread sync issues", which could mean a million different things). If he knows what's actually wrong, he can be more specific. If he just notices a problem but doesn't know what, sees some tool spitting out warnings he doesn't understand, and just assumes they're related, that's not helpful. It could just as easily send us barking up the wrong tree and not actually fix anything.

> Roughly 75% of tsan error messages are related to our WorkItem class. This class makes analyzers to cry:
> 3.1. WorkItems can be stored and managed outside of background thread. For example, you can re-run a finished task while another thread reads results of previous execution.
> 3.2. As a result of 3.1, there is no real guarantee that task is not running during its destruction.

Every function has preconditions and postconditions. Things that must be true before you can call it, and things that will be true after it returns. The vast majority of them are up to the caller to ensure, as it becomes prohibitive for the function itself to check and recheck the caller did everything right. In fact, certain preconditions, like 'not thread-safe' is explicitly for absolving the implementation of having to check at runtime (if a function had to check to make sure it wasn't being used in a thread-unsafe manner, it may as well just be thread-safe; and making everything thread-safe would be way too costly).

> On a practice, these issues do not lead to troubles because of our coding discipline - we seems to do not do bad things when using work items, but ananyzers are not smart enough to determine it - they spam data race warnings every time when we use work items.

What kind of function annotation options does tsan have? Properly documenting functions with what you can and can't do with them would certainly be a benefit (if it's not done already), and if that documentation can also serve to tell tsan how the function is used to help silence it, the better.

components/esmterrain/storage.cpp
kcat: Probably because that causes the returned `std::string` to be constructed in the given thread from a const char array, rather than copy-constructing in the given thread from a separate `std::string` object. A static const/constexpr string can also be initialized at load time since it doesn't depend on runtime state, unlike a string which needs to be initialized when it's first executed since it does depend on runtime state (the allocator).

Given that the function returns a new string object, rather than a const reference, constructing it from a char array shouldn't be any worse (might even be better) than copy-constructing it from another string. In such a case, I'd recommend:
```
static constexpr char baseTexture[] = "textures\\tx_black_01.dds";
```
And similar for the other static const string.

akortunov:
> What kind of function annotation options does tsan have?

I did not find any excepts on "no_sanitize" attribute and the -fsanitize-blacklist parameter for Clang.

components/sceneutil/workqueue.cpp
akortunov: Actually, you use the same approach in the `OpenAL_Output::StreamThread` destructor in [this](https://github.com/OpenMW/openmw/commit/0f05ccf72a20b0a2a927509c58b30dabdb0520a0) commit.

components/sceneutil/workqueue.cpp
akortunov: I did not invent anything - in all other cases in the engine we do the broadcast under lock.

components/sceneutil/workqueue.cpp
kcat: `OpenAL_Output` owns the thread the mutex is used for helping signal. The mutex is used to ensure the controlled thread knows to end before `join()` is called, in the same destructor. It's different here since this isn't being used to signal or join anything, but to temporarily halt deletion of the object while another handle in another thread is calling a method on it. But if another handle exists in another thread when deletion occurs, there's an ownership/lifetime problem with the object.

components/sceneutil/workqueue.cpp
kcat: I didn't say you did invent anything. All I'm saying is it's preferable, when possible, to not call `broadcast()` or `notify()` with the lock held to avoid scheduling conflicts. And in the case here, it does not seem necessary to call `broadcast()` with the lock held, as long as the lock was held when `mDone` was set.

akortunov:
OK, current results of discussion:
1. Do not touch the WorkItem class and its false-positives yet.
2. Create tickets for other suspicious messages.
3. Leave only fixes for real issues here.

xyzz:
> Do you have an idea if it is safe? In theory, unref should be threadsafe.

It's because osg atomic isn't implemented the best way, it's using `__sync_synchronize` for when compiling with gcc builtins (getting side tracked here, but it also has a fallback which does NO synchronization whatsoever). Whether it's a false positive in asan, or these gcc builtins are used wrong / don't provide the required thread safety I can't say since I didn't bother looking deep into it. However, what you can do is patch osg and replace their Atomic impl with std::atomic, like so: https://github.com/xyzz/openmw-android/blob/master/buildscripts/patches/osg/0001-Replace-Atomic-impl-with-std-atomic.patch

AnyOldName3:
Robert Osfield is generally against OSG having conditional compilation, but switching OpenThreads to use C++11 `std` features might be something he'd make an exception for. If he's already allowing GCC bultins as a conditional path, actual standard language features might be an easier sell.

I've seen bits of OSG code that switch between `std::auto_ptr` and `std::unique_ptr` depending on the C++ version, so it's not as if they avoid C++11 features like the plague when there's a reasonable justification. If it turns out that there are genuine problems that are OpenThreads' fault, fixing those might count.

AnyOldName3:
While Robert Osfield doesn't like conditional compilation in OSG, if OpenThreads is already optionally using GCC builtins, something from `std` might be an easier sell.

I've seen bits of OSG code that switch between `std::auto_ptr` and `std::unique_ptr` depending on the C++ version, so it's not as if they avoid C++11 features like the plague when there's a reasonable justification. If it turns out that there are genuine problems that are OpenThreads' fault, fixing those might count.

AnyOldName3:
I've posted a reply to this twice now and it's not showing up.

AnyOldName3:
That's three.

akortunov:
Also @xyzz, which kind of issues did you try to solve with std::atomic?

AnyOldName3:
Help:

image

AnyOldName3:
Time seems to have gone back to chronological order now, so the conversation should be easy to follow again if anyone was waiting for that.

xyzz:
> Also @xyzz, which kind of issues did you try to solve with std::atomic?

Basically, the Atomic is broken. Any access to it would generate an error from tsan. I think you will see more of these if you compile osg with tsan as well.

AnyOldName3:
This is really a debate for OSG's forums rather than here. They seem to have chilled out since my initial visits. Just make sure to put in something that resembles a real name and to let their post template finish your post with *thanks, <name>* and you shouldn't be sent away.

xyzz:
I'm not really interested in filing an issue for osg, if you want to do that then go ahead.

xyzz:
Btw, I just realized if you're not compiling osg with tsan, you might see false positives. So that could be the reason why you see workitem errors, as tsan isn't tracking osg's mutex implementation properly.

akortunov:
I built OSG with -fsanitize=thread and indeed messages related to navmeshes and work item disappeared: [report_full_patched_osg.txt](https://github.com/OpenMW/openmw/files/3160664/report_full_patched_osg.txt)
There are a lot of messages about stateset sharing and IncrementalCompileOperation.

@AnyOldName3, it seems these messages are related to shadows, maybe they would be interesting for you: [report_shadows.txt](https://github.com/OpenMW/openmw/files/3160668/report_shadows.txt)

akortunov:
I suggest to continue discussion in [this](https://forum.openmw.org/viewtopic.php?f=6&t=5942) thread.

akortunov:
> Did you find any specific issue with it?

There was a tsan warning about search in the map which is under destroying.
I tried just to place the `mAsyncNavMeshUpdater` after the `mCache` first, but in this case I get stable crashes on exit without log.
The "stop processing before cleanup" soultion seems to work well - it does not produce crashes and warnings for me.

akortunov:
> While Robert Osfield doesn't like conditional compilation in OSG, if OpenThreads is already optionally using GCC builtins, something from std might be an easier sell.

IIRC, he wanted to move OSG 4.0 to the C++11 and remove the OpenThreads dependency from it at all.

akortunov:
Any objections about fixes, currently present here?

AnyOldName3:
I don't think we're *getting* OSG 4.0. VSG has taken over, and uses C++17.

elsid:
> > Did you find any specific issue with it?
> 
> There was a tsan warning about search in the map which is under destroying.
> I tried just to place the `mAsyncNavMeshUpdater` after the `mCache` first, but in this case I get stable crashes on exit without log.
> The "stop processing before cleanup" soultion seems to work well - it does not produce crashes and warnings for me.

Please provide thread sanitizer report with issue in the pr description. Just for convenience. I've found some links in comments but I'm not sure are they right.

akortunov:
> Please provide thread sanitizer report with issue in the pr description. Just for convenience. I've found some links in comments but I'm not sure are they right.

I got such messages sometimes:

```
WARNING: ThreadSanitizer: data race (pid=3277)
  Write of size 8 at 0x7b240029e390 by main thread:
    #0 operator delete(void*) /tmp/portage/sys-devel/gcc-8.2.0-r6/work/gcc-8.2.0/libsanitizer/tsan/tsan_new_delete.cc:74 (libtsan.so.0+0x74b0d)
    #1 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/shared_ptr_base.h:171 (openmw+0x4746be)
    #2 std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/shared_ptr_base.h:706 (openmw+0x4746be)
    #3 std::__shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem>, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/shared_ptr_base.h:1145 (openmw+0x4746be)
    #4 std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> >::~shared_ptr() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/shared_ptr.h:103 (openmw+0x4746be)
    #5 std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >::~pair() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_pair.h:198 (openmw+0x4746be)
    #6 void __gnu_cxx::new_allocator<std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::destroy<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >(std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/ext/new_allocator.h:140 (openmw+0x4746be)
    #7 void std::allocator_traits<std::allocator<std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > > >::destroy<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >(std::allocator<std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >&, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/alloc_traits.h:487 (openmw+0x4746be)
    #8 std::_Rb_tree<osg::Vec3f, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >, std::_Select1st<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::_M_destroy_node(std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_tree.h:661 (openmw+0x4746be)
    #9 std::_Rb_tree<osg::Vec3f, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >, std::_Select1st<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::_M_drop_node(std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_tree.h:669 (openmw+0x4746be)
    #10 std::_Rb_tree<osg::Vec3f, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >, std::_Select1st<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::_M_erase(std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_tree.h:1874 (openmw+0x4746be)
    #11 std::_Rb_tree<osg::Vec3f, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >, std::_Select1st<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::~_Rb_tree() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_tree.h:965 (openmw+0xcfa82f)
    #12 std::map<osg::Vec3f, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::~map() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_map.h:300 (openmw+0xcfa82f)
    #13 DetourNavigator::NavMeshManager::~NavMeshManager() /home/andrei/Downloads/openmw/components/detournavigator/navmeshmanager.hpp:20 (openmw+0xcfa82f)
    #14 DetourNavigator::NavigatorImpl::~NavigatorImpl() /home/andrei/Downloads/openmw/components/detournavigator/navigatorimpl.hpp:9 (openmw+0xcfa82f)
    #15 DetourNavigator::NavigatorImpl::~NavigatorImpl() /home/andrei/Downloads/openmw/components/detournavigator/navigatorimpl.hpp:9 (openmw+0xcfa82f)
    #16 MWWorld::World::~World() /home/andrei/Downloads/openmw/apps/openmw/mwworld/worldimp.cpp:549 (openmw+0x836095)
    #17 MWBase::Environment::cleanup() /home/andrei/Downloads/openmw/apps/openmw/mwbase/environment.cpp:183 (openmw+0xb28e49)
    #18 OMW::Engine::~Engine() /home/andrei/Downloads/openmw/apps/openmw/engine.cpp:253 (openmw+0xb44d9b)
    #19 OMW::Engine::~Engine() /home/andrei/Downloads/openmw/apps/openmw/engine.cpp:274 (openmw+0xb454e5)
    #20 std::unique_ptr<OMW::Engine, std::default_delete<OMW::Engine> >::~unique_ptr() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/unique_ptr.h:81 (openmw+0xb2c7e9)
    #21 runApplication(int, char**) /home/andrei/Downloads/openmw/apps/openmw/main.cpp:254 (openmw+0xb2c7e9)
    #22 wrapApplication(int (*)(int, char**), int, char**, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /home/andrei/Downloads/openmw/components/debug/debugging.cpp:137 (openmw+0xc7eb7a)
    #23 __libc_start_main ../csu/libc-start.c:308 (libc.so.6+0x2375a)

  Previous atomic read of size 1 at 0x7b240029e390 by thread T9 (mutexes: write M759836356693844880):
    #0 pthread_mutex_unlock /tmp/portage/sys-devel/gcc-8.2.0-r6/work/gcc-8.2.0/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:3813 (libtsan.so.0+0x4046b)
    #1 __gthread_mutex_unlock /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/x86_64-pc-linux-gnu/bits/gthr-default.h:778 (openmw+0xcfbd1a)
    #2 std::mutex::unlock() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/std_mutex.h:121 (openmw+0xcfbd1a)
    #3 std::unique_lock<std::mutex>::unlock() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/std_mutex.h:323 (openmw+0xcfbd1a)
    #4 std::unique_lock<std::mutex>::~unique_lock() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/std_mutex.h:232 (openmw+0xcfbd1a)
    #5 Misc::Locked<DetourNavigator::NavMeshCacheItem const>::~Locked() /home/andrei/Downloads/openmw/./components/misc/guarded.hpp:10 (openmw+0xcfbd1a)
    #6 DetourNavigator::AsyncNavMeshUpdater::processJob(DetourNavigator::AsyncNavMeshUpdater::Job const&) /home/andrei/Downloads/openmw/components/detournavigator/asyncnavmeshupdater.cpp:170 (openmw+0xcfbd1a)
```

I tired to move the `mAsyncNavMeshUpdater` declaration after `mCache` or `mLastRecastMeshManagerRevision` to make sure that we stop update threads before do a cleanup, but I get stable crashes without log in both cases. I suppose it should not happen, if these fields are really independent.
So I decided to stop update threads explicitely before really start to cleanup any related data.

elsid:
> I tired to move the `mAsyncNavMeshUpdater` declaration after `mCache` or `mLastRecastMeshManagerRevision` to make sure that we stop update threads before do a cleanup, but I get stable crashes without log in both cases. I suppose it should not happen, if these fields are really independent.

It's wrong to destruct `mAsyncNavMeshUpdater` before `mCache`, because elements from `mCache` use `mAsyncNavMeshUpdater`.

`mCache` contains `SharedNavMeshCacheItem` which is basically a wrapper over `NavMeshCacheItem`. `NavMeshCacheItem::mUsedTiles` contains `NavMeshTilesCache::Value` which calls `NavMeshTilesCache::releaseItem` using pointer to `AsyncNavMeshUpdater::mNavMeshTilesCache` in [destructor](https://github.com/OpenMW/openmw/blob/6857ad657f59c92c8ad701da6dc293fd8e6d7bdc/components/detournavigator/navmeshtilescache.hpp#L71). So if `mAsyncNavMeshUpdater` is destructed before, pointer become dangling.

I can't reproduce problem using clang. Current solution seems like a hack because there is no detailed description what really happens. I'm not sure is it gcc specific issue. I will try to investigate more.

akortunov:
> I can't reproduce problem using clang. Current solution seems like a hack because there is no detailed description what really happens. 

I can drop that commit in any moment if you are not happy about it, but design is still a quite strange - we destroy the navigator object while it still does job in other threads. A more safe approach would be to stop navigator tasks before we start to destroy data, IMO.

Capostrophic:
Status?

akortunov:
> Status?

Still waiting for reply from @elsid about f3f2d3f65e7c8994eb07bbbe03d5b5f9b487a4f3. As I said, I can drop this commit if it is not good enough.

elsid:
> I can drop that commit in any moment if you are not happy about it, but design is still a quite strange - we destroy the navigator object while it still does job in other threads. A more safe approach would be to stop navigator tasks before we start to destroy data, IMO.

Please drop it for now. I can't catch the problem by tsan yet. But I think I just need some time to find out what is wrong.

akortunov:
> Please drop it for now. I can't catch the problem by tsan yet. But I think I just need some time to find out what is wrong.

Done.

elsid:
> ```
> WARNING: ThreadSanitizer: data race (pid=3277)
>   Write of size 8 at 0x7b240029e390 by main thread:
>     #0 operator delete(void*) /tmp/portage/sys-devel/gcc-8.2.0-r6/work/gcc-8.2.0/libsanitizer/tsan/tsan_new_delete.cc:74 (libtsan.so.0+0x74b0d)
>     #1 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/shared_ptr_base.h:171 (openmw+0x4746be)
>     #2 std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/shared_ptr_base.h:706 (openmw+0x4746be)
>     #3 std::__shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem>, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/shared_ptr_base.h:1145 (openmw+0x4746be)
>     #4 std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> >::~shared_ptr() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/shared_ptr.h:103 (openmw+0x4746be)
>     #5 std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >::~pair() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_pair.h:198 (openmw+0x4746be)
>     #6 void __gnu_cxx::new_allocator<std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::destroy<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >(std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/ext/new_allocator.h:140 (openmw+0x4746be)
>     #7 void std::allocator_traits<std::allocator<std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > > >::destroy<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >(std::allocator<std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >&, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/alloc_traits.h:487 (openmw+0x4746be)
>     #8 std::_Rb_tree<osg::Vec3f, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >, std::_Select1st<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::_M_destroy_node(std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_tree.h:661 (openmw+0x4746be)
>     #9 std::_Rb_tree<osg::Vec3f, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >, std::_Select1st<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::_M_drop_node(std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_tree.h:669 (openmw+0x4746be)
>     #10 std::_Rb_tree<osg::Vec3f, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >, std::_Select1st<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::_M_erase(std::_Rb_tree_node<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >*) /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_tree.h:1874 (openmw+0x4746be)
>     #11 std::_Rb_tree<osg::Vec3f, std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > >, std::_Select1st<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::~_Rb_tree() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_tree.h:965 (openmw+0xcfa82f)
>     #12 std::map<osg::Vec3f, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> >, std::less<osg::Vec3f>, std::allocator<std::pair<osg::Vec3f const, std::shared_ptr<Misc::ScopeGuarded<DetourNavigator::NavMeshCacheItem> > > > >::~map() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/stl_map.h:300 (openmw+0xcfa82f)
>     #13 DetourNavigator::NavMeshManager::~NavMeshManager() /home/andrei/Downloads/openmw/components/detournavigator/navmeshmanager.hpp:20 (openmw+0xcfa82f)
>     #14 DetourNavigator::NavigatorImpl::~NavigatorImpl() /home/andrei/Downloads/openmw/components/detournavigator/navigatorimpl.hpp:9 (openmw+0xcfa82f)
>     #15 DetourNavigator::NavigatorImpl::~NavigatorImpl() /home/andrei/Downloads/openmw/components/detournavigator/navigatorimpl.hpp:9 (openmw+0xcfa82f)
>     #16 MWWorld::World::~World() /home/andrei/Downloads/openmw/apps/openmw/mwworld/worldimp.cpp:549 (openmw+0x836095)
>     #17 MWBase::Environment::cleanup() /home/andrei/Downloads/openmw/apps/openmw/mwbase/environment.cpp:183 (openmw+0xb28e49)
>     #18 OMW::Engine::~Engine() /home/andrei/Downloads/openmw/apps/openmw/engine.cpp:253 (openmw+0xb44d9b)
>     #19 OMW::Engine::~Engine() /home/andrei/Downloads/openmw/apps/openmw/engine.cpp:274 (openmw+0xb454e5)
>     #20 std::unique_ptr<OMW::Engine, std::default_delete<OMW::Engine> >::~unique_ptr() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/unique_ptr.h:81 (openmw+0xb2c7e9)
>     #21 runApplication(int, char**) /home/andrei/Downloads/openmw/apps/openmw/main.cpp:254 (openmw+0xb2c7e9)
>     #22 wrapApplication(int (*)(int, char**), int, char**, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /home/andrei/Downloads/openmw/components/debug/debugging.cpp:137 (openmw+0xc7eb7a)
>     #23 __libc_start_main ../csu/libc-start.c:308 (libc.so.6+0x2375a)
> 
>   Previous atomic re>ad of size 1 at 0x7b240029e390 by thread T9 (mutexes: write M759836356693844880):
>     #0 pthread_mutex_unlock /tmp/portage/sys-devel/gcc-8.2.0-r6/work/gcc-8.2.0/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:3813 (libtsan.so.0+0x4046b)
>     #1 __gthread_mutex_unlock /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/x86_64-pc-linux-gnu/bits/gthr-default.h:778 (openmw+0xcfbd1a)
>     #2 std::mutex::unlock() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/std_mutex.h:121 (openmw+0xcfbd1a)
>     #3 std::unique_lock<std::mutex>::unlock() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/std_mutex.h:323 (openmw+0xcfbd1a)
>     #4 std::unique_lock<std::mutex>::~unique_lock() /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/include/g++-v8/bits/std_mutex.h:232 (openmw+0xcfbd1a)
>     #5 Misc::Locked<DetourNavigator::NavMeshCacheItem const>::~Locked() /home/andrei/Downloads/openmw/./components/misc/guarded.hpp:10 (openmw+0xcfbd1a)
>     #6 DetourNavigator::AsyncNavMeshUpdater::processJob(DetourNavigator::AsyncNavMeshUpdater::Job const&) /home/andrei/Downloads/openmw/components/detournavigator/asyncnavmeshupdater.cpp:170 (openmw+0xcfbd1a)
> ```

@akortunov  This seems to be not a full message. Sanitizer usually aslo prints `Location is heap block of size <...> at <...> allocated by <...> thread`, `Thread <...> created by <...> thread at` and more. Please provide full report message.

PR #2376, opened 2019-05-11
GetWeaponType returns -2 for picks and -3 for probes (feature #5031)

Capostrophic:
See [feature request](https://gitlab.com/OpenMW/openmw/issues/5031).


akortunov:
LGTM.

PR #2377, opened 2019-05-11
Armor mitigation tweaks (feature #5033)

Capostrophic:
1. Don't account for armor rating in weapon priority. It's ultimately not very effective, since enemy armor rating is not actually related to the weapon used and most of the time this would simply reduce physical damage rating by a fixed multiplier, but it has a sometimes noticeable overhead every time an AI attack is done. So it was removed for optimization. For the end user there should be no observable difference.

2. Magic armor mitigation for creatures based on Shield effect magnitude (see [feature request](https://gitlab.com/OpenMW/openmw/issues/5033) for details). I checked up with Hrnchamd on the logic of the (rather miniscule) changes, which was already mostly "implemented".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,7 @@
     Feature #5000: Compressed BSA format support
     Feature #5010: Native graphics herbalism support
     Feature #5031: Make GetWeaponType function return different values for tools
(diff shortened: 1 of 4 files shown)


akortunov:
LGTM.

PR #2380, opened 2019-05-12
Allow scripted faction leaving using PCLowerRank (feature #5036)

Capostrophic:
[See feature request for details.](https://gitlab.com/OpenMW/openmw/issues/5036)

NpcStats::lowerRank has been adjusted to erase the faction entry from the player faction ranks map if the new rank value is below 0 to imitate the MCP addition.


akortunov:
Works a bit strange if player is expelled from faction.
In this case PC is still considered as a member of faction, even if PCLowerRank removed PC from faction.

Capostrophic:
Right, hm. Corrected.

PR #2383, opened 2019-05-13
Make enchanting make more sense (bug #5038)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5038)

1. Bring enchanting chance calculations in line with Hrnchamd's much more accurate research. It will now use the chance mult GMSTs properly, use the precise calculated enchantment strength, account for the player character's fatigue term and use the correct multipliers for the relevant stats. 
2. Clean up and repurpose getEnchantPoints for point 1. It returns the precise calculated value normally for the purposes of enchanter service price and enchanting chance calculation and sum of rounded down effect costs used for the actual enchantment base cost if the supplied argument is "false".
3. Clean up labels update of the enchanting window too because not all of them worked as intended after some of the introduced changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@
     Bug #5004: Werewolves shield their eyes during storm
     Bug #5018: Spell tooltips don't support purely negative magnitudes
     Bug #5028: Offered price caps are not trading-specific
+    Bug #5038: Enchanting success chance calculations are blatantly wrong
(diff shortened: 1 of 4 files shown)


akortunov:
Which kind of testing do you need here?

Capostrophic:
Making sure I didn't overlook something *again* since the changes are somewhat intrusive.

PR #2384, opened 2019-05-14
Fix RaiseRank/LowerRank script functions

akortunov:
Implements [feature #4316](https://gitlab.com/OpenMW/openmw/issues/4316).

Currently these functions do not work at all in OpenMW, while in Morrowind they work, but changes are not persistent (MCP fixes this issue).

This PR re-uses an existing `mFactionRank` field in NPC data to override a current rank for NPC via scripts. If we did not alter current rank for NPC, the `mFactionRank` is still empty and we use data from base NPC record as fallback. It allows us to do not bloat savegames.
Older OpenMW versions are still compatible - they still can read and store `mFactionRank` field, but they obviously ignore this field when they determine a current rank for NPC.

Now Ajira should get the next rank after her quests properly.

Notes: 
1. LowerRank will not exit NPC from faction (the minimum rank is still 0 for NPC's).
2. It would be nice to rewrite the whole faction-related API, but it is out of scope of this PR.

### DIFF
--- a/apps/openmw/mwclass/npc.cpp
+++ b/apps/openmw/mwclass/npc.cpp
@@ -1416,6 +1416,19 @@ namespace MWClass
 
     int Npc::getPrimaryFactionRank (const MWWorld::ConstPtr& ptr) const
     {
+        std::string factionID = ptr.getClass().getPrimaryFaction(ptr);
+        if(factionID.empty())
+            return -1;
+
+        // Search in the NPC data first
+        if (const MWWorld::CustomData* data = ptr.getRefData().getCustomData())
+        {
+            int rank = data->asNpcCustomData().mNpcStats.getFactionRank(factionID);
+            if (rank >= 0)
+                return rank;
+        }
+
+        // Use base NPC record as a fallback
         const MWWorld::LiveCellRef<ESM::NPC> *ref = ptr.get<ESM::NPC>();
         return ref->mBase->getFactionRank();
     }
--- a/apps/openmw/mwscript/statsextensions.cpp
+++ b/apps/openmw/mwscript/statsextensions.cpp
@@ -1019,7 +1019,19 @@ namespace MWScript
                     if (ptr == player)
                         return;
 
-                    ptr.getClass().getNpcStats(ptr).raiseRank(factionID);
+                    // If we already changed rank for this NPC, modify current rank in the NPC stats.
+                    // Otherwise take rank from base NPC record, increase it and put it to NPC data.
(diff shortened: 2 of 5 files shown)


apps/openmw/mwclass/npc.cpp
```diff
@@ -1414,8 +1414,19 @@ namespace MWClass
...
     }
 
-    int Npc::getPrimaryFactionRank (const MWWorld::ConstPtr& ptr) const
+    int Npc::getPrimaryFactionRank (const MWWorld::Ptr& ptr) const
     {
+        std::string factionID = ptr.getClass().getPrimaryFaction(ptr);
+        if(factionID.empty())
+            return -1;
+
+        // Search in the NPC data first
```
Capostrophic: The added getFactionRank method looks pretty much identical to the code following this line. I think something like 
```cpp
int rank = ptr.getClass().getNpcStats(ptr).getFactionRank();
if (rank != -1)
    return rank;
```
should be possible to do.

Or, to make getPrimaryFactionRank const-correct again (accept a ConstPtr), we can really assume NPCs without custom data (obviously) never have a faction ranks map set up, so it's also possible to do something like this:
```cpp
if (const MWWorld::CustomData* data = ptr.getRefData().getCustomData())
{
    int rank = data->asNpcCustomData().mNpcStats.getFactionRank();
    if (rank >= 0)
        return rank;
}
```

akortunov: > int rank = ptr.getClass().getNpcStats(ptr).getFactionRank();

I do not understand you. There is no getFactionRank() function without arguments in the NpcStats.

akortunov: Using the getFactionRank(factionID) now.

apps/openmw/mwscript/statsextensions.cpp
```diff
@@ -1019,7 +1019,19 @@ namespace MWScript
                     if (ptr == player)
                         return;
 
-                    ptr.getClass().getNpcStats(ptr).raiseRank(factionID);
+                    // If we already changed rank for this NPC, modify current rank in the NPC stats.
+                    // Otherwise take rank from base NPC record, increase it and put it to NPC data.
+                    int currentRank = ptr.getClass().getNpcStats(ptr).getFactionRank(Misc::StringUtils::lowerCase(factionID));
```
Capostrophic: NPC stats functions including the newly added one will normalize the faction ID automatically and other calls of them here and in LowerRank make the assumption that it will always happen so this lowerCase operation and the same operation in LowerRank is kind of redundant.

PR #2386, opened 2019-05-14
Exit waiting at the correct time when the player is in air

Capostrophic:
Remove awkward pointer handling and move the logic of the "solid ground" messagebox into setPtr so it can trigger at the correct time.


akortunov:
LGTM. Seems to be a regression, so there is no need for changelog entry.

PR #2389, opened 2019-05-15
Introduce a custom format() function

akortunov:
Allows to get rid of `boost::format` and hackish `Misc::StringUtils::replace`.

A current approach:
```
Misc::StringUtils::replace(message, "%d", std::to_string(mDays).c_str(), 2);
```
A new approach:

```
message = Misc::StringUtils::format(message, mDays);
```

Unfortunately, arguments of `std::string` type still should be converted to C-string, as with `replace` function. Simple types (such as `int`) can be used directly.

### DIFF
--- a/components/misc/stringops.hpp
+++ b/components/misc/stringops.hpp
@@ -1,7 +1,6 @@
 #ifndef MISC_STRINGOPS_H
 #define MISC_STRINGOPS_H
 
-#include <cstring>
 #include <string>
 #include <algorithm>
 
@@ -17,6 +16,19 @@ class StringUtils
             return toLower(x) < toLower(y);
         }
     };
+
+    // Allow to convert complex arguments to C-style strings for format() function
+    template <typename T>
+    static T argument(T value) noexcept
+    {
+        return value;
+    }
+
+    template <typename T>
+    static T const * argument(std::basic_string<T> const & value) noexcept
+    {
+        return value.c_str();
+    }
 
 public:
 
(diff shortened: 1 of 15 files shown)


components/misc/stringops.hpp
```diff
@@ -212,30 +215,18 @@ class StringUtils
...
-            withLen = strlen(with);
-
-        std::size_t found;
-        if ((found = str.find(what)) != std::string::npos)
-            str.replace(found, whatLen, with, withLen);
-
-        return str;
+        va_list args;
+        va_start (args, fmt);
+        size_t len = std::vsnprintf(NULL, 0, fmt.c_str(), args);
```
Capostrophic: should use nullptr here as usual

akortunov:
There is a warnings spam with VS2017:

```
stringops.hpp(222): warning C4840: non-portable use of class 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>' as an argument to a variadic function 
```

I'll try to fix it.

akortunov:
The simpliest way to avoid such warnings is to make the `fmt` a `const char*`, but in this case there will be additional `c_str()` calls.

akortunov:
OK, use the C++11-style variadic templates instead of C-style macro now.
It also allows to get rid of boring `c_str() ` conversions in every `format()` call.
Let see if all compilers are happy with it...

akortunov:
Also it may be worth to move functions implementations from stringops.hpp to stringops.cpp.

akortunov:
Compilers are happy.

components/misc/stringops.hpp
```diff
@@ -18,6 +17,19 @@ class StringUtils
         }
     };
 
+    // Allow to convert complex arguments to C-style strings for format() function
+    template <typename T>
+    static T Argument(T value) noexcept
```
elsid: Why east const here and below but not `char const*`? Also function name should start from lower case by our [conventions](https://wiki.openmw.org/index.php?title=Code_Formatting_Conventions).

components/misc/stringops.hpp
```diff
@@ -212,30 +224,26 @@ class StringUtils
...
 
-        std::size_t found;
-        if ((found = str.find(what)) != std::string::npos)
-            str.replace(found, whatLen, with, withLen);
+        return ret;
+    }
 
-        return str;
+    template <typename ... Args>
+    static std::string format(std::string& fmt, Args const & ... args)
```
elsid: `const std::string& fmt`? `fmt` should not change in this function.

components/misc/stringops.hpp
akortunov: > Why east const here and below but not char const*? 

I do not understand what do you mean.

akortunov: > Also function name should start from lower case by our conventions.

Resolved.

Also if you mean that we should use the `char const*` instead of `const char*`, the `const char*` is more widely used in OpenMW. If we want to change it, it should be changed over the whole engine, but it is outside of scope of this PR.

psi29a:
I like...

PR #2390, opened 2019-05-15
Launcher: Clone content list button (feature #4784)

Capostrophic:
[Feature request](https://gitlab.com/OpenMW/openmw/issues/4784).

Adds a button into the launcher which which allows the user to clone the existing content list under a new name — the name dialog has the same rules as for new content list dialog, i.e. the name is not empty and the content list with the same name doesn't already exist. The created list will have the current file selection as its contents. It will also become the current content list. The action has Ctrl-G keyboard shortcut.


akortunov:
Seems to work fine. Merging to get more testers.

PR #2391, opened 2019-05-15
Content selector: don't add file namesakes (bug #2395)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/2395).

Normally we only avoid adding files with the exact same file path into the content file list. To fix the issue, we need to avoid adding files with the same file name into the list. There were more issues, though:

1. The editor doesn't work the same as the launcher in regards to adding files into content selector dialog. It loads files in a direct order while the launcher loads them in a reverse order. So any duplicate dropping would be inconsistent between the two behaviors. I decided to make launcher behavior the main one, since that's the most commonly used behavior.
2. Unlike the test suite and the engine the launcher loaded the local data folder files *after* it loaded the rest of the files (which are in LIFO order) which caused... oddities. Now it has the highest priority and its files are loaded before any other folder to allow the fix to work properly here too.
3. The editor tried to load data folders twice in every instance openmw.cfg was reloaded. There are safeguards in content selector to prevent this from causing issues but it is a waste of CPU cycles.

### DIFF
--- a/components/contentselector/model/contentmodel.cpp
+++ b/components/contentselector/model/contentmodel.cpp
@@ -408,12 +408,6 @@ bool ContentSelectorModel::ContentModel::dropMimeData(const QMimeData *data, Qt:
 
 void ContentSelectorModel::ContentModel::addFile(EsmFile *file)
 {
-    for (int row = 0; row < mFiles.size(); row++)
-    {
-        if (!mFiles.at(row)->fileName().compare(file->fileName(), Qt::CaseInsensitive))
-            return;
-    }
-
     beginInsertRows(QModelIndex(), mFiles.count(), mFiles.count());
         mFiles.append(file);
     endInsertRows();
@@ -434,7 +428,7 @@ void ContentSelectorModel::ContentModel::addFiles(const QString &path)
     {
         QFileInfo info(dir.absoluteFilePath(path2));
 
-        if (item(info.absoluteFilePath()) != 0)
+        if (item(info.fileName()))
             continue;
 
         try {
--- a/components/contentselector/model/contentmodel.cpp
+++ b/components/contentselector/model/contentmodel.cpp
@@ -408,6 +408,17 @@ bool ContentSelectorModel::ContentModel::dropMimeData(const QMimeData *data, Qt:
 
 void ContentSelectorModel::ContentModel::addFile(EsmFile *file)
 {
+    for (int row = 0; row < mFiles.size(); row++)
+    {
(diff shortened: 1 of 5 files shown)


AnyOldName3:
I'm guessing it's not worth changing the data structure to something more quickly searchable?

Capostrophic:
I'm hesitant to do it because there might as well not bring a real benefit. Something that has logarithmic search complexity (QSet doesn't btw) would typically iterate elsewhere worse. Then a major part of content selector would have to be rewritten and that's a risk of overlooking something and breaking something important.

There's one thing I can do to make the loop potentially faster and more simple though.

akortunov:
I suppose the comparison should be case-insensitive (if it is not already).

Capostrophic:
Updated.

components/contentselector/model/contentmodel.cpp
```diff
@@ -408,6 +408,17 @@ bool ContentSelectorModel::ContentModel::dropMimeData(const QMimeData *data, Qt:
 
 void ContentSelectorModel::ContentModel::addFile(EsmFile *file)
 {
+    for (int row = 0; row < mFiles.size(); row++)
+    {
+        if (!mFiles.at(row)->fileName().compare(file->fileName(), Qt::CaseInsensitive))
+        {
+            beginRemoveRows(QModelIndex(), row, row);
+                mFiles.removeAt(row);
```
akortunov: 1. It seems you just remove file from list and do not delete it. It can lead to memory leaks.
2. Do we really need a such bulk construction to just replace a value in QList?

Capostrophic: To 2: the file in question is **not** "replaced", and it shouldn't be. There is a difference between a) removing an existing basically random item from the container and appending a new different one and b) replacing the item. And QList has no means to do a.

Capostrophic:
Apparently the launcher and editor load data folders in *different order* which makes things more difficult. This works as intended with the editor but not as intended with the launcher due to intricacies of QMap used for loading data folders in GameSettings and its LIFO order.

Capostrophic:
Pushed some experimental changes. I made the editor behavior consistent -- it also loads folders in reverse order and no longer loads the files in them twice in every instance of config reading. The content selector changes were adjusted to support this behavior. This means that files in folders with higher priority will appear earlier in the order (especially noticeable in the editor), like in the launcher earlier, I'm not sure if it's critical or not. Need some testing and feedback.

Capostrophic:
There isn't an extra loop anymore.

PR #2392, opened 2019-05-17
Get ride of some redundant GUI stuff

akortunov:
Since the minimum requirement is the 3.2.2 now, we can remove ControllerRepeatClick and MWScroll classes - the first one is a part of MyGUI since 3.2.2, and the stock Scroll uses it by default.

Note: there are some todo's about `setKeyFocusWidget`, but I do not understand what Scrawl meant here.


Capostrophic:
I did some research and apparently it's now safe to replace setKeyFocusWidget with the MyGUI method it wraps. Hidden widgets no longer retain key focus in upstream due to something scrawl did in August 2014 and onKeyFocusChanged is triggered by setKeyFocusWidget anyways. Note that the various plain setKeyFocusWidget(nullptr) calls need to be replaced with resetKeyFocusWidget.

akortunov:
> it's now safe to replace setKeyFocusWidget with the MyGUI method it wraps

If you say so...

PR #2394, opened 2019-05-18
Enhance optimizer to reduce number of drawables

mp3butcher:
add an loop until convergence to redundant free node graph.
It reduces drastically number of drawables and so on FPS

### DIFF
--- a/components/sceneutil/optimizer.cpp
+++ b/components/sceneutil/optimizer.cpp
@@ -87,13 +87,18 @@ void Optimizer::optimize(osg::Node* node, unsigned int options)
         RemoveEmptyNodesVisitor renv(this);
         node->accept(renv);
         renv.removeEmptyNodes();
-
-        RemoveRedundantNodesVisitor rrnv(this);
-        node->accept(rrnv);
-        rrnv.removeRedundantNodes();
-
-        MergeGroupsVisitor mgrp(this);
-        node->accept(mgrp);
+        
+        bool repass = true;
+        while(repass)
+        {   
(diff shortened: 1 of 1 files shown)


Capostrophic:
Have you measured the difference?

akortunov:
I do not see any noticable difference.

mp3butcher:
I have an old intel chipset on an old celeron...so it would'nt be significative to compare..
The nuber of drawable is divide by more than 2 so it rocks

akortunov:
> The nuber of drawable is divide by more than 2 so it rocks

On which scene? I tested it in Ebonheart (which is known to have badly optimized meshes).
In both master and this PR testing scene had near 3100 drawables.

akortunov:
> if you don't see any difference you're surely bound by the fragment cost

How the fragment cost affects resulting scene graph and count of drawables in it?

mp3butcher:
mmh perhaps I mod some datavariance else where...Let me check

akortunov:
> 12000

It is ain't right. Pofiler in Balmora shows about 2000-3000 drawables with default settings, and about 7000 with real-time shadows and water reflections for everything.

wareya:
Uh. Where in Balmora? The position and angle where I stand make a huge difference, from 1500 to 10k. FoV matters too.

mp3butcher:
@akortunov 
i set view distance to 66666 and watch in the worst direction and it give me 12k
(I don't know why I answer such trivial critic)
Let me the time to confirm I mod other stuff before yelling

PR #2395, opened 2019-05-18
remove an expensive call to normalize

mp3butcher:


files/shaders/lighting.glsl
```diff
@@ -9,7 +9,7 @@ void perLight(out vec3 ambientOut, out vec3 diffuseOut, int lightIndex, vec3 vie
 
     lightDir = gl_LightSource[lightIndex].position.xyz - (viewPos.xyz * gl_LightSource[lightIndex].position.w);
     lightDistance = length(lightDir);
-    lightDir = normalize(lightDir);
+    lightDir = lightDir / lightDistance;
```
Capostrophic: Zero division might cause issues.

mp3butcher: arf you may be right...

PR #2397, opened 2019-05-24
Analogue movement fixes

Capostrophic:
There were issues gamepad users complained about:
1. Always run setting interfered with analogue movement function. You would have (slow) running animation while moving at roughly walking speed. Apparently Xbox version doesn't have the equivalent of always run so I disabled always run completely on gamepads in favor of analogue movement.
2. Analogue movement didn't affect fatigue loss, so (especially noticeable when combined with the first issue) your fatigue would be consumed as if you were running at full speed even if you didn't really run. So I made analogue movement speed multiplier apply to the running/swimming/sneaking fatigue loss.

The fixes do not need changelog entries.

Edit: independent testing shows that this works as intended.

### DIFF
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -703,7 +703,7 @@ namespace MWInput
                     mOverencumberedMessageDelay = 0.f;
                 }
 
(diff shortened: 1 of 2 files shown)


akortunov:
> independent testing shows that this works as intended.

Merging then.

PR #2398, opened 2019-05-26
Make SayDone return 1 on the frame speech is started (bug #4879)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/4879).

Hrnchamd stated that SayDone uses the playback marker used for looking up the sound loudness for lip sync to determine whether the sound is being played, and that marker only fires off during the sound system update (which is later than when the scripts are updated) so it will start to return 0 with the following frame — however it will start to return 1 once the sound buffer end is reached no matter what.

I've added a say sound queue map that is copied into the active say sounds map during the sound system update to simulate this behavior.

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -994,6 +994,15 @@ namespace MWSound
 
     void SoundManager::updateSounds(float duration)
     {
+        // We update active say sounds map for specific actors here
+        // because for vanilla compatibility we can't do it immediately.
+        SaySoundMap::iterator queuesayiter = mSaySoundsQueue.begin();
+        while (queuesayiter != mSaySoundsQueue.end())
+        {
+            mActiveSaySounds[queuesayiter->first] = queuesayiter->second;
+            mSaySoundsQueue.erase(queuesayiter++);
+        }
+
         static float timePassed = 0.0;
 
         timePassed += duration;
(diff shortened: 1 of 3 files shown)


apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -554,13 +554,13 @@ namespace MWSound
...
 
-    bool SoundManager::sayDone(const MWWorld::ConstPtr &ptr) const
+    bool SoundManager::sayDone(const MWWorld::ConstPtr &ptr, bool ignoreInit) const
     {
         SaySoundMap::const_iterator snditer = mActiveSaySounds.find(ptr);
         if(snditer != mActiveSaySounds.end())
         {
             if(mOutput->isStreamPlaying(snditer->second))
-                return false;
+                return ignoreInit && !mOutput->getStreamOffset(snditer->second);
```
kcat: This doesn't look safe since the stream offset will increment whenever OpenAL processes some audio, which can be any time after the stream playback starts. That would cause `SayDone` to return 0 randomly on the same frame `Say` was called, depending on when the next internal mixing update is scheduled.

Capostrophic: Ok, I had another idea. Here's something somewhat different (no doubt equally crutchy).

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -1068,8 +1076,15 @@ namespace MWSound
...
         }
 
-        SaySoundMap::iterator sayiter = mActiveSaySounds.begin();
-        while(sayiter != mActiveSaySounds.end())
+        SaySoundMap::iterator sayiter = mSaySoundsQueue.begin();
+        while (sayiter != mSaySoundsQueue.end())
+        {
+            mActiveSaySounds[sayiter->first] = sayiter->second;
+            mSaySoundsQueue.erase(sayiter++);
+        }
```
kcat: Looks better, but the placement of this hand-off from the queue to the active list is behind the 30hz update check. That would mean there may or may not be multiple frames after `Say` before `SayDone` starts giving 0. Putting this loop before the `if(timePassed < (1.0f/30.0f))` check may fix that.

Capostrophic: OK, updated.

kcat:
Looks good to me

PR #2399, opened 2019-05-26
add Octree Occlusion Queries feature

mp3butcher:
A cheap way to improve render performance when there's lot of hidden objects...
Include a lazy optimisation for shadow:
I narrow the far plane of the view with OQ results splitting cull traversal in 3 traversal:
1) for OQ  (just to submit queries)
2) for not occluded (that give me the narrow  frustum with the nearest far plane possible)
3) for the rest (don't care the far plane of this frustum)
2&3 may be merged but **include terrain in far plane computation**  can be of some use for fine tuning

NB: 
- You can check drawcharge of the oq compromise with stats (it's part of the few quads)
- Indoor is treated as outdoor...(differents octree parameter -for other scene scale-should be used for indoor, but it wouldn't be very usefull as indoor dont lag)
- octree occlusion culling is hierarchic so keep  **max BVH OQ level count** low for more reactivity **SOLVED**
- bug still exists with hierarchical terrain+oq+shadow culling (some are drawn even if occluded when **include terrain in far plane computation** is true) **SOLVED**
- little bug with ingame shadows settings (can't enable it if shaders disabled) but switching shadow on/off ingame is so usefull that i let the bug be.
Video (outdated):
[Demo video](https://www.youtube.com/watch?v=DeWsFH9pdlg&hl=en&cc_lang_pref=en&cc_load_policy=1)
Capture d’écran de 2019-06-11 20-20-04
reference:


mp3butcher:
arf thread issue i haven't spot after some refactoring..!!

mp3butcher:
it works only for osg 3.6 for the moment

mp3butcher:
would require testing with osg 3.4

akortunov:
Does not compile with OSG 3.6.0 - it complains about undeclared `_validQueryGeometry` variable and missing `osg::QueryGeometry::QueryResult` type.

akortunov:
From what I can tell, `osg::QueryGeometry` was changed in 3.6.4, not in 3.6.0.

apps/openmw/mwrender/objects.cpp
```diff
@@ -66,22 +184,125 @@ void Objects::insertBegin(const MWWorld::Ptr& ptr)
...
+        }
+
+    }
+    bool recursivCellRemoveStaticObject(osg::OcclusionQueryNode & parent, osg::Node * childtoremove)
+    {
+        osg::Group * pchild; bool removed=false;
+        for(uint i=0; i< parent.getNumChildren(); ++i)
+        {
+            pchild=(parent.getChild(i)->asGroup());
+            if(removed = pchild->removeChild(childtoremove)) break;
```
akortunov: I suppose there should be comparison, not assignment.

mp3butcher: no ..i'll put parenthesis
...this breaks save feature will see that

mp3butcher: > Does not compile with OSG 3.6.0 - it complains about undeclared `_validQueryGeometry` variable and missing `osg::QueryGeometry::QueryResult` type.

this require lattest release candidate of 3.6...changes have been made two month ago

akortunov:
> this require lattest release candidate of 3.6...changes have been made two month ago

Then you should use the `OSG_VERSION_GREATER_OR_EQUAL(3,6,4)` version check, not the `#if OSG_VERSION_GREATER_OR_EQUAL(3,6,0)`.

Also it seems that this PR sligthly reduces amount of drawables in scene, but noticably increases a Cull time, what leads to performance drops when OpenMW is limited by the main thread, not by the drawing one (-10-20 FPS on my setup). Is there a way to avoid it?

mp3butcher:
I haven't notice such a cull increase...
The problem with this implementation is that i don't see any way to know cell bound before all objects are added..It can lead to deep scene graph in some case as i consider first added center as cell center
And concerning the breakage of the save feature it's as always due to the screen shot camera
I'll investigate but if you have any leadi'll take it

mp3butcher:
Even when i replace OQNs with groups, the screenshot crash persists
there's surely something coming from openmw policies that make it freezes
Insights would be required (light managemnt, stateset update..)

mp3butcher:
Even when i replace OQNs with groups before the screenshot, crash persists!
there's surely something coming from openmw policies that make it freezes (cullvisitor convention, light managemnt, stateset update..)
Insights would be required

mp3butcher:
replacing OQN with Group for screen shot camera seams to  work...will find better

mp3butcher:
.. nodemask tweak on querygeometry is cleaner to prevent the problem.
I find the screenshot hack tricky but
https://gitlab.com/OpenMW/openmw/commit/20633692 
might prevent this kind of WTF bug

JDGBOLT:
Interesting work, interesting looking at it functioning within the Old Ebonheart city within the Tamriel Rebuilt mod, which is really the area that this sort of function could be useful, as it's a fairly dense and packed city that is notorious for bad performance. Unfortunately I see the same that akortunov has, which is less performance compared to the feature being off. The draw time definitely does go down, but overall framerate is a good deal less. In old ebonheart I'm seeing it down to like 30-40 compared to a solid 60.

I took a video just since I was curious, this is without default settings, but over 60 frames for the object rather than 5 and 16 pixel visibility threshold. Mostly just took the video to see the effect. https://youtu.be/0Qhxz-7grWA

mp3butcher:
You both have big GPUs which seams to endure high drawcall charge so bottleneck for you isn't draw phase. perhaps (or not)  enabling shadow will make a difference as it multiplies by 4 drawcall counts .
You could also change octree parameters to find a good compromise
On lower end gpu, the gain is obvious both on the cull and draw

mp3butcher:
the bottle neck in old ebon heart seams in game mechanism (put on pause to see the difference)

AnyOldName3:
Am I right in thinking that you're delaying receiving the query results for a few frames so you don't introduce a hard pipeline stall? If so, this is going to add pop-in, and I'm not super happy about accepting wrong frames just so we get more frames.

akortunov:
> If so, this is going to add pop-in

Yes, this PR causes graphics artifacts for a couple of frames after savegame load.

kcat:
> OQ yields pop-in artifacts because that's how it works (asynchronously -so frame latency is not avoidable-)...Why aren't you happy with this fact?

Because pop-in is ugly, especially when it's close to the camera. Just because it's intended behavior doesn't make it any better.

mp3butcher:
if someone can explain these stats (right OQ on/left OQ off)...
Capture d’écran de 2018-12-25 03-55-01
That's really WTF....

wareya:
Are you using some kind of tolerance on moving objects to reduce false negatives?

wareya:
Oh, my bad, I thought more about how this would work and I think I get the problem now. Kind of annoying, I can't think of a way to add tolerance to this kind of occlusion system to make up for the pop-in that the one-frame latency causes...

Maybe it's possible to disable occlusion at times when it's going to be giving bogus data?

AnyOldName3:
The two approaches modern engines use to get around the pop in are:

* Using heuristics like oversized occlusion geometry (so a query also covers the region an object is likely to move into), ignoring query results for fast-moving objects or when the camera moves quickly (as those query results are likely to be wrong) and whatever else happens to be effective for the game.
* Using the query results in a compute shader having dispatched everything that might possibly be drawn in a few huge draw calls.

Both of these approaches are hard enough to get right that lots of games don't use occlusion queries at all, instead relying on CPU-based approaches.

For example, in Skyrim, when indoors, all objects in a room are parents of one node. Everything in the same room as the camera is submitted provided it passes view frustum culling. If a door object passes view frustum culling and is open, the room it connects to is added to the scene being considered.

Also in Skyrim, when outdoors, some objects (like buildings) have occlusion planes inside them. Occlusion planes encountered during the cull pass are remembered. Objects are tested against these planes, and if their bounding volume is entirely behind one, they don't get drawn. Remembering the planes can be done either by just storing them in a list and testing against them directly for every bounding volume, or by rasterising them in software to a low-resolution depth buffer and then rasterising bounding volumes (also in software) and seeing if they'd pass. This doesn't actually work out much slower than hardware occlusion queries but doesn't have the pop-in issue, and so is the approach used by lots and lots of games. OSG has some level of support for this (mp3butcher once accidentally disabled it when trying to make shadow volume stuff optional) but I've never got it to do anything when playing with the example.

(Also, because Bethesda is Bethesda, they accidentally broke this code in Skyrim Special Edition, so that only has view frustum culling.)

mp3butcher:
using an heuristik extra extents for query geometries is straight forward
using compute shader is prohibited for compatibility reason
using osg::OccluderVolume would introduce extra CPU cull charge 
I've also added but not commit few other options (like max hierarchic OQ)

My problem on this pr is that the gain of OQ is not as big as it should:
Could anyone explain these stats ?! (right OQ on/left OQ off)...
Capture d’écran de 2019-06-02 22-40-46

That's really WTF....same FPS for double cull&draw&gpu charge!!!!

Capostrophic:
It looks like on the left OSG takes its sweet time to count objects before culling them.

mp3butcher:
okay found the bootlneck : OQN::computebounds

mp3butcher:
over 20 fps in Old Ebonheart!!
Capture d’écran de 2019-06-03 19-59-38

JDGBOLT:
I can confirm that it definitely seems faster with than without now. Severely limiting my video card I can say that with it on it boosts the framerate in old ebonheart from like 40-50 fps to 60+. Also with shadows in old ebonheart, from around 30 or so fps to 40. There still seems to be a slight bit of overhead, as with a GPU time of 9ms or so I was getting 80 or so fps, but it's definitely a lot better than it was, seems overall to be faster in the cases I have tested, rather than before where it was slower usually. Good work! :)

akortunov:
Does not build (at least with OSG 3.4 ) because of undeclared `_validQueryGeometry` variable.

mp3butcher:
@JDGBOLT 
All OQs made for each cameras..In the case of shadow camera, there's not a lot of occlusion so the gain is not very obvious. It might even getting slower than without...
I've put an arbitrary node mask on OQ geometry for the moment (Mask_RenderToTexture) but i'm waiting for https://gitlab.com/OpenMW/openmw/commit/20633692 review to remove the hack in screenshot camera

akortunov:
After recend patches this PR started to work, but it still has some issues, which should be addressed:
1. Shadows are pretty broken with this PR - they flicker and disappear randomly.
2. Cell loading takes a lot more time now (to build the octree, I suppose). Can it be done during cell preloading?
3. Pop-ins are pretty bad. Is there a way to reduce them?

mp3butcher:
1 I don't have any problem with shadow with osg3.6.4
2 yes cell loading is a problem there's overcalling to computebound..will see that
3 you can reduce poppins tweaking parameters 

There still flicking artifact but nothing that can't be fixed (or minimized with parameters)

mp3butcher:
after all, i don't know if the last commit improve anything..:/

akortunov:
Does not build again. There is no `<osg/ContextData>` file.

akortunov:
> after all, i don't know if the last commit improve loading time..:/

No, it does not.

Also while shadows flickering seems to be fixed, culling itself is a bit broken now:

Screenshot_20190604_200814

It is the same place as on your last screenshot, but buildings are erroneusly culled. Also there is a lot of such messages in console:

```
osgOQ: QG: Invalid RQCB
```

mp3butcher:
> > after all, i don't know if the last commit improve loading time..:/
> 
> No, it does not.
> 
> Also while shadows flickering seems to be fixed, culling itself is a bit broken now:
> 
> Screenshot_20190604_200814
> 
> It is the same place as on your last screenshot, but buildings are erroneusly culled. Also there is a lot of such messages in console:
> 
> ```
> osgOQ: QG: Invalid RQCB
> ```

Yes the last commit prohibit the use of osg::OclusionQueryNode in sky.cpp

mp3butcher:
still flickering with recursive OQN...set max BVH OQ level count to 1

AnyOldName3:
> using osg::OccluderVolume would introduce extra CPU cull charge

While I'm not saying OSG's built-in CPU occlusion culling is necessarily good (or bad - as I said, I've not even got the official example to do anything) a good implementation of CPU-side occlusion culling should be able to *reduce* the cull traversal workload. Remember, we do CPU skinning during the cull phase, and have particle systems that change from frame to frame, so when a few actors and particles are occluded, you buy a decent amount of extra time, and you can cull more aggressively as you don't need heuristics to mask the fact that you're working with an old frame, buying yet more time. It wouldn't be widely-used in commercial games if it wasn't an improvement over a simpler approach, and we're going to catch lots of flak from MGE XE fans if we have pop-in that the original engine doesn't.

This PR touches the engine in fewer places than I'd expected so far, so provided you can keep it clean, it might well turn out to be mergeable. However, I'm not going to be keen on merging something that has the potential to turn into a maintenance burden especially if it has to rely on magic heuristics to avoid artefacts.

If you hadn't noticed, I'm not sold on hardware occlusion queries being the best approach, especially when we don't have the manpower to try implementing everything to see what's actually best.

> But shadow could benefit from occlusion culling if applied to terrain too (far plane culling should narrow shadow cameras view frustra)

Can you try rephrasing this as I'm not following. Are you just saying that we could reduce the depth range of shadow maps if we knew a previous frame's depth range?

By the way, I'm noticing a few usages of unsafe pointers instead of `ref_ptr`s. Do you have a good reason for this, as we really don't want to go inserting foot-guns into the code.

mp3butcher:
OQNs improves a lot culling and so on draw and gpu performance (still have problem with hierarchic OQN but that's not the point here)
I don't put actor in OQN, only static meshes and particle system...so there's no issue with the update-at-cull tweak used in openmw

The only heuristik in this work is cellsize but disabling highlevel OQN make this a priori hypothesis harmless for performance. All the bugs mitigation i talked about came from bug in osg OQNs, and i fixed most of them

When I said shadow could benefit from occlusion culling : OQ improve culling so if you put a distinct nodemask on OQgeometry you can get a smaller frustum volume
during shadow traversal you derive smaller shadow cameras frustra from this smaller view frustum
and then less draw work for shadow cameras

mp3butcher:
oh yeah :)

akortunov:
It seems the performance is noticably degraded since my previous testing (the Cull phase now takes 8ms per frame instead of 5.5ms in Old Ebonheart).

mp3butcher:
@akortunov 
please be more precise (which commits do you compare)

akortunov:
> which commits do you compare

97b69440be2c809dfb06d0967562d4304ebd3c72 and newer ones.

mp3butcher:
could you compare https://github.com/OpenMW/openmw/pull/2399/commits/ba95b833b200c7770d54e0448925cc8011586d11 and https://github.com/OpenMW/openmw/pull/2399/commits/ab1b3e4565e9604e19bd6c07d50b0c2fdd377f61 to see if it comes from the tweak to avoid flickering artefacts when OQNs go hierarchical

mp3butcher:
The last commit is hacky but it shows how to reduce far plane for view frustum in shadowcode
set high value for **max BVH OQ level count** (>=5 seams to work)

@AnyOldName3 
I recomputed the far plane in mwshadowtechnique but it doesn't help improve performance
It would require your insight on how you convert view frustum to shadowcams frusta

mp3butcher:
The last commit is hacky but it shows how to reduce far plane for view frustum in shadowcode
set high value for **max BVH OQ level count** (>=5 seams to work)

@AnyOldName3 
I recomputed the far plane in mwshadowtechnique but it doesn't help improve performance
It would require your insight on how you convert view frustum to shadowcams frusta

AnyOldName3:
That's a fairly complicated process, most of which came from Robert Osfield's implementation rather than me. I'm going to leave out stuff like what happens if the view frustum is orthographic as that's completely untested and so probably sets things on fire, or when settings I haven't exposed to the user get meddled with. The whole process (not all of which is relevant to what you're doing here - some of it is just coming up with clipping volumes, for example) is:

* Get the parameters for the main projection matrix.
* Set the compute near/far mode to near and far using primitives if the user's set it to do that (as there are issues with the bounding volumes of certain things, so there's no point doing it based on those).
* Traverse the scene.
* Get OSG to compute the near and far planes based on that traversal.
* Work around an OSG feature where it meddles with the projection matrix and moves the near plane away if the far plane is too far.
* Construct an object representing the main view frustum:
  - Get the view and projection matrices and computed near and far planes from the cull visitor.
  - Work out where the frustum corners are in world-space.
  - Work out where the eye position and other interesting parts of the frustum are in world-space.
* Make the considered near and far planes cover the smallest frustum allowed by both the initial matrices and the CV's computed values.
* Put the CV's compute near/far mode back to what it was before (always DO_NOT_COMPUTE...)
* Note down where the traversed scene had lights that are allowed to create shadows based on the shadow settings (always just the sun).
* Compute an `osg::Polytope` for the light view frustum based on the light and the main view frustum:
  - Initialise a polytope to match the main view frustum.
  - (Using the definition of *faces* that means planes face into the polytope) Any plane that faces within 90° of the same direction of the light (so if we pretend the sun was being treated as a point light, it would be behind the plane) gets discarded, and the rest get copied to the new polytope.
  - Fill in other planes so that you end up with a polytope representing the view frustum plus any space which could contain shadow casters.
* Compute initial values for the shadow camera:
  - Compute vectors `lightSide` and `lightUp` perpendicular to each other and the light direction. This depends on how close the light and view directions are to each other, so if they're not too close, one approach, compatible with LiSPSM is used, and if they are too close, another approach is used. These vectors form the basis for the provisional light view space.
  - Put the view frustum corners in light space and use this to compute their range.
  - Try and reduce this range based on the scene's bounding sphere, which never helps at all.
  - Use the vectors to form a lookat-style view matrix and the ranges to form an orthographic light projection matrix.
* Create a `ComputeLightSpaceBounds` visitor (which extends CullStack like a cull visitor does). Make it's clipping polytope the one we computed earlier, but in provisional light clip space. Use this to traverse the scene (only shadow casters) and work out its range in light space based on the bounding boxes of individual drawables. If a distant terrain node is encountered, set the range in all directions to the maximum as there were bugs with terrain bounds (which I think have been fixed now, so that should be tweaked).
* Use the results of this traversal to reduce the light projection matrix's range in x and y (I'm not sure if there's a good reason why it doesn't so it in z, too). Also convert the range back into main view space and if possible, reduce the considered near and far planes.
* For each shadow map:
  - Set the camera's view and projection matrices to the provisional light view and projection matrices.
  - Get a local version of the polytope, but in provisional light view space.
  - Work out where to chop up the frustum based on the practical split scheme in the original PSSM paper.
  - Ignore all the stuff calling itself parallel-split because Robert wrote it and it's not what the original PSSM paper or any other source says to do. Instead, do the cascaded stuff. (Although technically there's a setting to use the wrong path as people were complaining about the correct path.)
  - Crop the shadow camera to the main view frustum, considering only the subset for the current cascade:
    * Create a convex hull represented as edges from the view frustum.
    * Chop off the regions of space not in the current cascade based on its near and far plane.
    * Convert the hull into light clip space.
    * Get the clip space range of the hull. Use this to shrink the light projection matrix. Don't care that this doesn't cover casters outside the main view frustum as depth clamping will be on.
    * Convert the hull to light view space. Use this to generate more clipping planes for the cull visitor to use. Exclude one of the Z ones as we have depth clamping on.
  - Add the extra planes to the local polytope.
  - Construct a new VDSMCameraCullCallback and attach it to the camera. They should probably be recycled instead of reallocated per-camera per-frame, but we need to be sure they're not being used by the cull and draw traversal at the same time or something like that. I don't see how they could be used in the draw traversal at all, so I'm thinking we probably just need one per camera. This callback:
    * Attaches the local polytope to the CV so it is used for culling.
    * Traverses the scene.
    * Remembers the render stage and projection matrix (by pointer).
    * Doesn't compute the near and far planes as I disabled that when implementing depth clamping as the results would be wrong.
  - Traverse the scene's shadow casters.
  - If LiSPSM is on (which it always is, but I might make a setting for it):
    * Set up a uniform for the non-reprojected light space (as it's needed to work around a bug caused by an improvement I no longer like and might get rid of).
    * Do lots of magic to make LiSPSM, including cropping the considered volume based on the bounds of render leaves in the render stage. The weird decisions weren't my idea - I just fixed the maths so it actually worked.
    * Use the pointer version of the matrix got via the camera cull callback to replace the matrix in the render stage.
  - Stick the light view and projection matrices in the eye planes.
  - Mark the texture unit etc. as being in use.
* Set up OpenGL state for the stateset used and apply it.

There are lots of things in here that depend on each other, so if something's broken, it could be broken in lots of places. A good way to narrow it down is to make the check for `PERSPECTIVE_SHADOW_MAP` fail as LiSPSM is one of the most fragile parts of the chain.

mp3butcher:
is it this branch or distant terrain didn't work with VAO backend?

mp3butcher:
time for testing and review

mp3butcher:
Time for testing and review
Are you aware distant terrain fails with VAO backend?
`OSG_VERTEX_BUFFER_HINT=VAO ./openmw  --skip-menu`

Capostrophic:
>Are you aware distant terrain fails with VAO backend?

I am now, and it's probably very recent. Any leads?

akortunov:
It seems new in-game settings do not work properly:
1. Thre are a lot of 
```Shadow casting shader has not been set up. Remember to call setupCastingShader(Shader::ShaderManager &)``` messages in console if you disable shadows
2. Occlusion settings changes need a game restart, so adding them to the Settings window is pointless now.

**Comment by JDGBOLT** [comment]
Actually the occlusion settings do work, they just require you to change cells in order to start working, as it is only in use when the octree is generated on cell load.

**Comment by mp3butcher** [comment]
@JDGBOLT 
last commit should fix the bug (based on symptoms your stacktrace described)...Altough I haven't been able to reproduce so ...need testing

**Comment by mp3butcher** [comment]
@JDGBOLT
last commit should fix the bug (based on symptoms your stacktrace described)...Altough I haven't been able to reproduce so ...need testing

**Comment by JDGBOLT** [comment]
Correct me if I am wrong, but wasn't there something relating to the fact that dynamic data variance was bad for OpenMW, I think there were a number of discussions regarding it in this PR, among others: https://github.com/OpenMW/openmw/pull/2054 .

**Comment by mp3butcher** [comment]
I have reproduce the bug. seams datavariance is not reliable for this check, i use nodemask now...I could removed these datavariance. but in theory, datavariance for a transform states that it changes or not over time (to cache matrix or not). Can't remember if it's the case in practice in osg but in all cases it can't hurt i think..

**Comment by AnyOldName3** [comment]
The rule is that we don't have dynamic drawables and statesets, and use other approaches such as double-buffering instead. Dynamic drawables cause OSG to delay the start of the next frame's processing until all of them have been drawn, effectively going into single-threaded mode if the drawables happen to be in a late render bin. Dynamic transforms and switches etc. are okay, though, as that's just telling the optimiser not to flatten them.

**Comment by mp3butcher** [comment]
The terrain culling drive me mad:
You can see the fps problem facing a wall with OQ and Shadows enabled
and switching [Shadows] include terrain in far plane computation  true/false
the terrain bound seams not culled by the OQN but  terraindrawable are ?!

**Comment by JDGBOLT** [comment]
Would it possible to have some toggle or something to toggle the use of the octree occlusion within an indoor cell? It seems the places that have the most issues where things are removed or flicker most often happen indoors, probably due to the more complex and smaller geometry that is used indoors, with the various tile sets and the like. Like I notice within "Seyda Neen, Arrile's Tradehouse" cell, the when you enter the cell, the corner piece by the stairs will be flashing, along with the chandelier. I mostly notice things getting culled that shouldn't be indoors, and indoors is commonly not very hard on systems, and typically can go into the hundreds of fps, so having no occlusion within these cells is not as much of a big deal. I still notice popin if I turn the camera too quickly outdoors, but typically I don't notice too much of things flickering or other weirdness outdoors that much, other than say viewing through doorways sometimes things get culled that shouldn't.

**Comment by mp3butcher** [comment]
bugs fixed testing needed

**Comment by JDGBOLT** [comment]
There are some points where you forgot to remove references to max cell size, so by removing it from settings-default.cfg the game won't start now. The places I see it referenced is in apps/openmw/mwrender/objects.cpp, components/sceneutil/occlusionquerynode.hpp, and components/terrain/world.cpp

**Comment by mp3butcher** [comment]
oups

**Comment by mp3butcher** [comment]
Perhaps this last commit is not at its place but huge perf improvement, say if i remove it

**Comment by JDGBOLT** [comment]
Just so you know I believe with the "use isMobile" commit, I encounter a crash when loading a save game with this stack trace: https://pastebin.com/RjQN6wa0

**Comment by mp3butcher** [comment]
I moved this last commit in other pr https://github.com/OpenMW/openmw/pull/2421

**Comment by akortunov** [comment]
By the way, how do you handle semi-transparent meshes? Objecs behind them should not be culled.

**Comment by AnyOldName3** [comment]
Alpha testing discards fragments, so there's no depth write to worry about. I *think* we have depth writes off for alpha blended stuff, so they shouldn't be a concern, either.

**Comment by akortunov** [comment]
I just tested with default mesh (a semi-transparent yellow exclamation mark) and this PR tries to cull objects behind it, but does it inconsistently.

**Comment by mp3butcher** [comment]
i introduced bugs during cleaning i just fixed perhaps inconsistencies have been removed

**Comment by mp3butcher** [comment]
travis sucks

**Comment by akortunov** [comment]
Seems to be pretty broken right now (it is a Buckmoth Legion Fort):

Screenshot_20190619_084554

**Comment by akortunov** [comment]
Just for reference: it seems to be incompatible with logarithmic depth buffer.

Also pop-ins are still exist.

**Comment by mp3butcher** [comment]
@akortunov read the doc
set max BVH level =1 and queryframecount=1

**Comment by mp3butcher** [comment]
erf hope it's the last bug...

**Comment by akortunov** [comment]
> set max BVH level =1 and queryframecount=1 and increase maxpopin to 5000

Unfortunately, it does not really help, it only allows to reduce the severity of issue by sacrificing the part of performance gain. I suppose, to fully solve this issue, the `min pop in distance` should be equal to viewing distance, right? In this case player sacrifices 100% of performance gain.

Even more, with large values (a `min pop in distance = 5000`, for example) occlusion culling starts to **decrease** performance on my setup - additional culling workload spends more FPS than reducing count of drawables gives in many scenes (basically everywhere excepts of large TR cities).

**Comment by mp3butcher** [comment]
popin are due to latency of requests
latency is prorata of number of occlusion test
number of occlusion test is connected to BVH level and queryframecount
min pop in distance forces query result on distance criteria but don't make economy on them,
worse this parameter could put zfar away for shadow and so decrease perf

**Comment by AnyOldName3** [comment]
I have two questions right now.

* Are the occlusion queries happening against the oversized MSAA depth buffer, or the resolved depth buffer? If it's the latter, things might get weird when I get around to making alpha-to-coverage work, so it might be worth making sure it's the former.

* How are you handling camera rotations? My instinct is that the safest way to get maximum performance with minimum pop in is to:

1. Keep the `osg::Polytope` used for frustum culling for the occlusion frame around until we use the queries.
1. Query results are valid for anything completely contained within this polytope, so if those queries say we can skip drawing, it's safe.
1. Anything partially outside or wholly outside this polytope has invalid query results, as the depth buffer doesn't cover the region. This means that we should ignore the query result and draw anyway (provided it passes frustum culling for the current frame.
1. We can test for this by flipping all the planes in the remembered polytope and ignoring query results for volumes that pass that.

**Comment by mp3butcher** [comment]
@AnyOldName3 
> Are the occlusion queries happening against the oversized MSAA depth buffer, or the resolved depth buffer?

OQ are performed in the dedicated renderbin after all other OPAQUES for maximum z rejection and so it operates on the resolved depth buffer
I don't understand why i would have used costy MSAA depthtest...

> How are you handling camera rotations?

Same camera as viewer-> same matrices

There can't be dedicated OQ camera without copy viewer->getCamera depthbuffer to it before using it.
But if you see further development perhaps the better would be to integrate them in further pr

**Comment by AnyOldName3** [comment]
> > How are you handling camera rotations?
> 
> I don't have to...Same camera as its parent node so same matrices
> 
> There can't be dedicated OQ camera without copy parent node camera depthbuffer to it before using it.
> But if you see further development perhaps the better would be to integrate them in further pr

That's not what I asked. I want to know what you do about things that have invalid occlusion query results because the camera moved.

Let's say we drew a frame with the red frustum, and performed occlusion queries against its depth buffer, and we're now culling a new frame with a different camera angle represented by the blue frustum.

image

Three of the occlusion queries are represented by the black, green and orange boxes:

image

With the green box, we don't need to draw it whatever the occlusion query says because it fails frustum testing for the blue frustum, so can't be in the blue frame. That means that we shouldn't traverse this subgraph regardless of what the query says.

With the black box, we can probably trust the result of the occlusion query to be valid because it's entirely within the red frustum, so if it failed the query, it was definitely because it was occluded. That means that we should use the query result to decide to traverse the subgraph.

With the orange box, however, we can't trust the query to be valid. If the query failed, it could be because the object was entirely occluded, or it could be because the shaded region isn't in the red frustum, so couldn't have been drawn even if it wasn't occluded. This means that we should traverse this subgraph regardless of what the query says.

**Comment by AnyOldName3** [comment]
> > Are the occlusion queries happening against the oversized MSAA depth buffer, or the resolved depth buffer?
> 
> OQ are performed in the dedicated renderbin after all other OPAQUES for maximum z rejection and so it operates on the parent node camera depth buffer
> look at changes in shadow code and CullVisitor::apply(occlusoinqueriesnode&)
> I don't understand why i would have used costy MSAA depthtest...if there's a option in openmw for msaa depthbuffer it can slow down OQs

I don't think you're right here. It looks to me like occlusion queries use the same camera as the rest of the scene now I've looked at it, and so it should be the same unresolved depth buffer that everything else gets drawn into.

**Comment by mp3butcher** [comment]
>  It looks to me like occlusion queries use the same camera as the rest of the scene now I've looked at it, and so it should be the same unresolved depth buffer that everything else gets drawn into.

that's what i mean by parent node cameras depthbuffer...there's multiple camera in openmw and then mulitiple depth buffer

**Comment by mp3butcher** [comment]
> With the orange box, however, we can't trust the query to be valid. If the query failed, it could be because the object was entirely occluded, or it could be because the shaded region isn't in the red frustum, so couldn't have been drawn even if it wasn't occluded. This means that we should traverse this subgraph regardless of what the query says.

there's no such thing as "invalid" query ence it's ready else it's not. For Orange box the query has great chance not to be ready: in this case subtree is visited

**Comment by AnyOldName3** [comment]
Can you please read things before replying to them. If something I'm saying sounds stupid, consider that it might not be because it's stupid, but instead because you've not read it correctly.

When I'm describing a query as valid or invalid, this isn't a technical term from the OpenGL spec, but instead an English word. In this context, you can take it to mean *trustworthy*. I'm asking what metrics you're using to decide whether or not to trust a query's results, and outlining a method you can use to avoid making decisions based on untrustworthy results.

**Comment by mp3butcher** [comment]
I added code to provoke OQ when objects enter frustum, 
if you see improvement, i let the code

**Comment by AnyOldName3** [comment]
Please don't edit posts to add new things to them. Other people don't get notified about edits, so we don't know that you've written anything extra.

> for orange box, it's the same case as the black one (i really don't understand what u wrote about it) but it you put it out of red (t0) frustum it's more pertinent: in this case it would fails frustum culling at t0 and success it at t1. at t1 OQ must be send in order to know if occluded, waiting for the OQ result, it passes the test.

What I'm saying is that when drawing a frame t1 using occlusion queries against frame t0, for anything that was partially or wholly outside t0's view frustum, you should draw it normally instead of using the query results. This is because the query can only be correct about things wholly within the t0 view frustum. They're *occlusion* queries, not things-outside-the-view-frustum queries.

**Comment by mp3butcher** [comment]
@AnyOldName3
yes, but  this all conversation is pointless as this Pr already take into account what you talking about.
I really would like more pertinent review. 
@all 
I fixed bug linked to ingame shadow settings so please review it

**Comment by mp3butcher** [comment]
I changed my "oq provocation strategy"...as you can see in the video, too much oq kill oqs
now I provoke oq only for out of frustum object but not for previously occluded

File: components/sceneutil/occlusionquerynode.cpp [hunk 2]
```diff
@@ -12,6 +12,9 @@
 #include <osg/ContextData>
 #endif
 
+//ensure minimum popin for in frustum occluded but drastically increase OQ count per frame
+//#define PROVOK_OQ_4_PREVIOUSLY_OCCLUDED 1
```
mp3butcher: uncomment this to provoke oq for stuff already in frustum but masked previously (increase a lot oq so increase max BVH level)

mp3butcher: uncomment this to provoke oq for stuff already in frustum but masked previously (increase a lot oq so increase max BVH level)

mp3butcher:
still problem with hierarchic oqn latency but with max level to 1, there's better result

mp3butcher:
@AnyOldName3 there's definitly weird interaction of oq with shadow probihibiting me to track entering frustum for shadowcams (see last commit)...not a drama as oqs on shadow cam are pretty useless but nevertheless frustrating ...

components/sceneutil/mwshadowtechnique.cpp
```diff
@@ -553,7 +553,7 @@ MWShadowTechnique::ShadowData::ShadowData(MWShadowTechnique::ViewDependentData*
     // set up the camera
     _camera = new osg::Camera;
     _camera->setName("ShadowCamera");
-    _camera->setReferenceFrame(osg::Camera::ABSOLUTE_RF_INHERIT_VIEWPOINT);
+    _camera->setReferenceFrame(osg::Camera::ABSOLUTE_RF);
```
AnyOldName3: This will break terrain shadows with distant terrain enabled - the shadow cameras are often a few miles above Vvardenfell, so the terrain system would choose the lowest quality LOD level for those cameras and that's not enough detail to get terrain shadows from anything other than Red Mountain during late evening. To get correct terrain shadows with working self-shadowing, the same LOD nodes need to be used for the shadow cameras and main camera, and that means we need to inherit the viewpoint.

mp3butcher: my excuses i thought it was an error..i rollback this hacky change

mp3butcher: arf not so anodine in fact

components/terrain/chunkmanager.hpp
```diff
@@ -36,7 +36,6 @@ namespace Terrain
 
         osg::ref_ptr<osg::Node> getChunk(float size, const osg::Vec2f& center, unsigned char lod, unsigned int lodFlags);
 
-        void setCullingActive(bool active) { mCullingActive = active; }
```
mp3butcher: unused...

psi29a:
Is this still being worked on?

mp3butcher:
not really active on this but i spot a zfar computation weirdness when occlusionquerrynode gets hierarchical....setting default **include terrain in far plane computation** to false and **max BVH OQ level count** to 0 hide this weirdness and get correct shadow optimisation working... this misbehavior can be adress in a further pr ...

mp3butcher:
But it's not mergeable yet because of chnage in osg api
https://github.com/openscenegraph/OpenSceneGraph/issues/901#issuecomment-576175076

mp3butcher:
complying with osg 3.6.5 api i found a bug with secure pop distance...

mp3butcher:
complying with osg 3.6.5 api i fixed bugs with secure pop distance and terrain occlusion...

apps/opencs/view/render/cell.cpp
```diff
@@ -137,7 +137,11 @@ void CSVRender::Cell::updateLand()
             else
             {
                 mTerrain.reset(new Terrain::TerrainGrid(mCellNode, mCellNode,
+<<<<<<< HEAD
                     mData.getResourceSystem().get(), mTerrainStorage, Mask_Terrain, OQsettings));
+=======
+                    mData.getResourceSystem().get(), mTerrainStorage));
+>>>>>>> 2a31382e2014a79cf6f2e4a124239edf9e10e85e
```
psi29a: Ewww....

apps/openmw/mwrender/objects.cpp
```diff
@@ -45,7 +45,7 @@ void Objects::resetSettings()
     mOQNSettings.querymargin = Settings::Manager::getFloat("queries margin", "OcclusionQueries");
     mOQNSettings.maxBVHOQLevelCount = Settings::Manager::getInt("max BVH OQ level count", "OcclusionQueries");
     mOQNSettings.securepopdistance = Settings::Manager::getFloat("min pop in distance", "OcclusionQueries");
-    mOQNSettings.OQMask = Mask_OcclusionQuery;
+    mOQNSettings.OQMask = SceneUtil::Mask_OcclusionQuery;
```
psi29a, line 48: We are trying to get away from this...  see: 
https://gitlab.com/OpenMW/openmw/-/merge_requests/178

components/CMakeLists.txt
```diff
@@ -51,7 +51,11 @@ add_component_dir (shader
 add_component_dir (sceneutil
     clone attach visitor util statesetupdater controller skeleton riggeometry morphgeometry lightcontroller
     lightmanager lightutil positionattitudetransform workqueue unrefqueue pathgridutil waterutil writescene serialize optimizer
+<<<<<<< HEAD
     actorutil detourdebugdraw navmesh agentpath shadow mwshadowtechnique occlusionquerynode
+=======
+    actorutil detourdebugdraw navmesh agentpath shadow mwshadowtechnique vismask recastmesh
+>>>>>>> 2a31382e2014a79cf6f2e4a124239edf9e10e85e
```
psi29a, line 58: yuk...

psi29a:
@mp3butcher this is going to be painful... but, it probably wasn't such a good day to merge. Please read:
https://gitlab.com/OpenMW/openmw/-/merge_requests/178

I would back our your merge and wait a bit. Once the dust settles then we should be good to go and your PR will have a lot less conflicts.

mp3butcher:
> @mp3butcher this is going to be painful... but, it probably wasn't such a good day to merge. Please read:
> https://gitlab.com/OpenMW/openmw/-/merge_requests/178
> 
> I would back our your merge and wait a bit. Once the dust settles then we should be good to go and your PR will have a lot less conflicts.

was very tired when have done the merge... lot to clean
Further lot of 3.6 conflicts (in sky.cpp particularly) might be removed

components/CMakeLists.txt
mp3butcher: dont know by which magic it achieved to build

JDGBOLT:
It's been a while since I last took a look at this PR, and thought I would try my hand at merging it into current master now that we have object paging merged. I was able to get it for the most part working with current master at https://github.com/JDGBOLT/openmw/tree/master-occlusion, with some caveats. It appears that the cause, whatever it is, of the occlusion functionality being slower with it on rather than off is something with the interaction of OSG 3.6.5 and the https://github.com/OpenMW/openmw/pull/2399/commits/cd0777655a737fc9df311cde5b374c1a17a79ea8 commit. I'm not sure if the problem is with OSG 3.6.5 or with the commit, but what I notice is that there seems to be some sort of stall that occurs for a decent chunk of time between the update bar of the debug graph and the start of the cull bar on the graph. If I take the commit and apply it against 7370f26, with OSG 3.4 things still work just fine, but with 3.6.5 it seems to have the behavior I listed, which is causing lower framerates as there seems to be something stalling the cull thread. Also with the branch it seems to have a conflict with the bzzt water culling that was merged, as I had crashes around that general area that I "fixed" in the https://github.com/JDGBOLT/openmw/commit/209512712051c0ef5be74c82065ccc20c36fb7d4 commit of just disabling it. I have had one instance of several crashes while going into an indoor cell, something maybe related to detour, but after those several crashes in a row I haven't gotten another crash like it in a while of playing around with the branch. OSG 3.6.4 seems to be completely broken in the latest commits of the branch, with the culling functionality not seeming to work whatsoever.

Just for reference, here is that branch compiled against OSG 3.4 OpenMW fork and OSG 3.6.5:

3.4 OpenMW OSG Fork

psi29a:
While this research looks promising, it has fallen behind with quite a bit of merge conflicts.  Feel free to dust it off though. Perhaps even create a merge request here:

PR #2401, opened 2019-05-28
Log orphan nodes of a pathgrid record in a single message

Capostrophic:
Slightly improves pathgrid verifier. It's by far one of the least useful verifiers but its messages take the most space in the results, so I had an idea to mitigate this by logging orphan nodes in a single pathgrid record in a single message. Some pathgrid records in Bethesda files have a lot of orphan nodes.

I reworded the message for the issue to be more obvious. The message also logged the path point coordinates, which I dropped because it too is frankly useless — the important information the message conveys is that the point is not connected to anything, the point being located in a specific place is not important.

Now you'll see this instead of having 8 separate messages being dumped on you.
single message

### DIFF
--- a/apps/opencs/model/tools/pathgridcheck.cpp
+++ b/apps/opencs/model/tools/pathgridcheck.cpp
@@ -120,18 +120,27 @@ void CSMTools::PathgridCheckStage::perform (int stage, CSMDoc::Messages& message
         }
     }
 
+    std::vector<int> orphanList;
     // check pathgrid points that are not connected to anything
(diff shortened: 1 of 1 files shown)


akortunov:
Do point indeces really matter for users? Maybe it would be enough to display something like "N points are not connected to any other point".

Capostrophic:
I don't wish to stray too much from what exists already.

Though tbh there are more things that could be changed about pathgrid verifier and this isn't really a significant improvement if it is at all in its current form. I'll hold off from this for now.

PR #2402, opened 2019-05-29
Make console to do not hide other GUI windows

akortunov:
Fixes [bug #1515](https://gitlab.com/OpenMW/openmw/issues/1515).

The main idea is to handle console window separately, not as a part of GUI modes stack.
The console window becomes nearly unusable with transparent UI, but Morrowind has the same issue. 
This PR should be enough to close this ticket, despite there are some minor derivations from Morrowind:
1. I still hide pinned windows when opening console, otherwise it will be unusable with a semi-transparent UI as well.
2. I restore focus to the console when closing a background window.
3. I do not allow to exit a main menu or inventory directly while the console window is visible - it leads to weird effects with both OpenMW and Morrowind.

### DIFF
--- a/apps/openmw/mwgui/windowmanagerimp.cpp
+++ b/apps/openmw/mwgui/windowmanagerimp.cpp
@@ -445,7 +445,6 @@ namespace MWGui
         mConsole = new Console(w,h, mConsoleOnlyScripts);
         mWindows.push_back(mConsole);
         trackWindow(mConsole, "console");
-        mGuiModeStates[GM_Console] = GuiModeState(mConsole);
 
         bool questList = mResourceSystem->getVFS()->exists("textures/tx_menubook_options_over.dds");
         JournalWindow* journal = JournalWindow::create(JournalViewModel::create (), questList, mEncoding);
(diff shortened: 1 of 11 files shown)


apps/openmw/mwgui/windowmanagerimp.cpp
```diff
@@ -2083,6 +2088,25 @@ namespace MWGui
         SDL_free(text);
     }
 
+    void WindowManager::toggleConsole()
+    {
+        bool visible = !mConsole->isVisible();
```
Capostrophic: "visible = not visible"

The logic of this variable is a bit confusing and may hurt maintainability.

Capostrophic:
Unlike Morrowind, I can't open the console window when there's a modal window open.
Persuasion dialog specifically prevents any attempt to change the focus to the console.

There is a need to have a more generic solution (or workaround of MyGUI's shortcomings). :/

akortunov:
> Unlike Morrowind, I can't open the console window when there's a modal window open.

There is nothing I can do here - it is a PURPOSE of modal windows - they are placed over other windows and grab the input.

PR #2403, opened 2019-05-29
Improve info message about missing GDB

akortunov:
Attempt to fix [bug #4341](https://gitlab.com/OpenMW/openmw/issues/4341#).

Currently OpenMW claims GDB to be not istalled when fails to find it. It does not handle case when GDB is installed, but located in directory which is not listed in PATH environmental variable.

An old message:
```
Failed to create a crash report. Please install 'gdb' and crash again!
```

A new message:
```
Failed to create a crash report - 'gdb' is not found.
Make sure that it is installed and located in directory, listed in your PATH environmental variable.
The current PATH variable value: /sbin:/usr/sbin:/usr/lib/llvm/6/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/opt/bin
```

Additionally if GDB location is known on MacOS, we can add its location to PATH. @nikolaykasyanov, what do you think?

### DIFF
--- a/components/crashcatcher/crashcatcher.cpp
+++ b/components/crashcatcher/crashcatcher.cpp
@@ -176,7 +176,8 @@ static void gdb_info(pid_t pid)
         int ret = system(cmd_buf);
 
         if (ret != 0)
-            printf("\nFailed to create a crash report. Please install 'gdb' and crash again!\n");
+            printf("\nFailed to create a crash report. Please make sure that 'gdb' is installed and present in PATH then crash again."
+                   "\nCurrent PATH: %s\n", getenv("PATH"));
         fflush(stdout);
 
         /* Clean up */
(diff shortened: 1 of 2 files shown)


components/crashcatcher/crashcatcher.cpp
```diff
@@ -176,7 +176,9 @@ static void gdb_info(pid_t pid)
         int ret = system(cmd_buf);
 
         if (ret != 0)
-            printf("\nFailed to create a crash report. Please install 'gdb' and crash again!\n");
+            printf("\nFailed to create a crash report - 'gdb' is not found."
+                   "\nMake sure that it is installed and located in directory, listed in your PATH environmental variable."
+                   "\nThe current PATH variable value: %s\n", getenv("PATH"));
```
Capostrophic: Way too verbose. I suggest something like this:
```
Failed to create a crash report. Please make sure that 'gdb' is installed and present in PATH then crash again. <newline>
Current PATH: %s
```

nikolaykasyanov:
@akortunov there's no well-known gdb location on macOS, I doubt it even exists since the developer tools moved to LLDB some time ago.

akortunov:
Then I suggest just to modify the message.
In theory, we can use the LLDB or a custom crash catcher on MacOS later, but it is outside of scope of this PR.

nikolaykasyanov:
@akortunov fine with me!

PR #2404, opened 2019-05-30
Make framerate limit configurable in the launcher

Capostrophic:
Because apparently if you want something to be done right, you must do it yourself, I made framerate limit configurable in the launcher, with the range being 1-1000 and the "default" value being 300. The actual default value of the setting wasn't changed, so the limit will appear to be disabled by default.
It's a QDoubleSpinBox because the setting has a floating point value type. I decided to have 1 decimal precision for the value.

Settings manager was made locale-independent again (after akortunov's changes in January), being locale-dependent was a real issue in the launcher since the changed values were using wrong separators on my Russian locale system at least so the engine couldn't recover them properly. get methods work like in the fallback manager while set methods use string streams again. This isn't necessarily bad — now they always use the correct separators and only have significant digits included without trailing zeroes and will be properly loaded in both the Qt-based apps and the engine.

Yes, this means that getFloat is going to be using boost::lexical_cast. I don't believe there's a good way around it. Lexical cast works best for the purpose and it's best to be consistent between different config managers.

If this is merged the *cough* discussions on what the default framerate limit should be can proceed.

### DIFF
--- a/components/settings/settings.cpp
+++ b/components/settings/settings.cpp
@@ -355,8 +355,8 @@ float Manager::getFloat (const std::string& setting, const std::string& category
     const std::string value = getString(setting, category);
     try
     {
-        // We have to rely on Boost because std::stof from C++11
-        // uses the current locale for separators which we don't want and often silently ignores parsing errors.
+        // We have to rely on Boost because std::stof from C++11 uses the current locale
+        // for separators (which is undesired) and it often silently ignores parsing errors.
         return boost::lexical_cast<float>(value);
     }
     catch (boost::bad_lexical_cast&)
--- a/components/settings/settings.cpp
+++ b/components/settings/settings.cpp
@@ -7,6 +7,7 @@
 
 #include <boost/filesystem/fstream.hpp>
 #include <boost/algorithm/string.hpp>
+#include <boost/lexical_cast.hpp>
 
 namespace Settings
 {
@@ -354,12 +355,14 @@ float Manager::getFloat (const std::string& setting, const std::string& category
     const std::string value = getString(setting, category);
     try
     {
-        return std::stof(value);
+        // We have to rely on Boost because std::stof from C++11
+        // uses the current locale for separators which we don't want and often silently ignores parsing errors.
+        return boost::lexical_cast<float>(value);
     }
-    catch(const std::exception& e)
+    catch (boost::bad_lexical_cast&)
     {
         Log(Debug::Warning) << "Cannot parse setting '" << setting << "' (invalid setting value: " << value << ").";
-        return 0;
+        return 0.f;
     }
 }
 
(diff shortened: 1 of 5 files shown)


wareya:
>with the range being 1-999

I pride myself in being able to run certain quakeworld sourceports at over 1000fps with the right settings. Might as well make the limit in the UI be 1000.0.

>and the "default" value being 300. The actual default value of the setting wasn't changed, so the limit will appear to be disabled by default.

Why not enable the limit by default anyway? Nobody is going to be upset by a default framerate limit of 300, and there's reasons to have it, and no good reason not to. And it's not like a framerate limit of 300 is going to interact with vsync for anyone.

Capostrophic:
>Might as well make the limit in the UI be 1000.0.

Noted. Waiting for more suggestions.

>Why not enable the limit by default anyway?

Because I'd like for this PR to be merely a "support" PR for a certain MR which actually changes the default framerate limit and I don't want this to be another branch of *intense discussions*.

AnyOldName3:
A few questions:

* Is this ready? If I press merge right now to make the argument go away, are we expecting bad things to happen?
* Does the disabling and enabling of the spinner depend on what's in the user's settings when they start the launcher, or is it always going to start disabled?
* Does disabling the spinner put a zero in `settings.cfg` so that people just clicking a checkbox end up with no limit once MR 54 is merged?
* What's the general feeling around putting this in-game, too?
* Can we have a screenshot, as that makes reviewing UI changes much easier?

> no good reason not to

Don't start that here, too. There are good reasons, just they're niche, and become more niche once this PR is merged, so are less good than the reasons to have a limit.

Capostrophic:
1. It won't break anything though in-game settings aren't modified.
2. It depends on the user settings.
3. Yes, exactly.
4. I dunno, didn't try to add it into the game.
5. Graphics tab now looks like this.
изображение
The displayed separator is affected by the locale but the saved value is fine.

components/settings/settings.cpp
```diff
@@ -354,12 +355,14 @@ float Manager::getFloat (const std::string& setting, const std::string& category
     const std::string value = getString(setting, category);
     try
     {
-        return std::stof(value);
+        // We have to rely on Boost because std::stof from C++11
+        // uses the current locale for separators which we don't want and often silently ignores parsing errors.
```
AnyOldName3: This might not be the best place to break the line - because the first line is much shorter than the second, if someone isn't paying much attention, it could look like we're avoiding `std::stof` *because* it's from C++11 and that the second line of the comment is something unrelated.

AnyOldName3:
So if I click merge, will anything bad happen?

Capostrophic:
No, I believe not. And a technical regression will be fixed.

PR #2407, opened 2019-06-01
Add separate textures for scrollbars

akortunov:
Implements a [feature #5051](https://gitlab.com/OpenMW/openmw/issues/5051).

New files have a lowest priority in VFS, so modders can override them as any other texture.
Also scrollbars become scalable, so you in theory can use textures with different resolution.


Capostrophic:
Looks fine to me. Since our scrollbar is not really compatible with Morrowind assets due to how its length is not fixed, this doesn't violate how it was said explicitly in the past we shouldn't supply Morrowind replacement assets, so I think this can get a pass. Any objections?

psi29a:
I hate binaries in our repo...

first, are these assets properly licensed? What are their origins?

akortunov:
> What are their origins?

GIMP. Arrows are based on arrow textures from minimap window, but it is not a copy-paste.

psi29a:
So you created them? Can you provide, either here or in a text file somewhere that indicates under what license you would like these to fall under? I wish we had done this with other assets, for now we assume they _all_ fall under the GPLv3 which is absurd. 

I think have your comment here is enough that we can point back to it later saying: see, he licensed this under XXX.

akortunov:
> I think have your comment here is enough that we can point back to it later saying: see, he licensed this under XXX.

Just as the engine itself (GPLv3, I suppose) - these textures were created specially for OpenMW.

Capostrophic:
GPL isn't really suitable for non-software works, there's a reason why psi29a suggested to use a Creative Commons license for the new assets.

akortunov:
> GPL isn't really suitable for non-software works, there's a reason why psi29a suggested to use a Creative Commons license for the new assets.

Use any license whatever you like.

Capostrophic:
Then these textures are released into public domain, alright. :smile:

psi29a:
That doesn't work, France for example (and many others) do not recognize public domain. If you want it to be that 'free' then please use 
https://creativecommons.org/publicdomain/zero/1.0/

Is that OK? I need you publicly state it instead of saying "whatever you like". I know it is obnoxious, but I want to make sure that OpenMW has the best chances.

wareya:
If you want to license something under the public domain, yeah, just say that you license it under the CC0 license.

akortunov:
> If you want to license something as if it were public domain, yeah, just say that you license it under the CC0 license.

All right, let it be the CC0 then.

PR #2411, opened 2019-06-03
Make sure fade operations properly end (bug #4639)

Capostrophic:
Due to various reasons fader may not properly update the alpha, and when it happens to be too late to update it (for example, if the game was minimized and the GUI updates were paused for the time of the fade operation) fade operation will not finish properly and the alpha will remain whatever it was in the frame before the fade operation was finished.

I added a safe guard for the case it happens so if there's no time to finish the fade operation the final target alpha is set anyway just in case it's not equal to the current alpha (if the alpha is the same there won't be any difference from the current behavior because notifyAlphaChanged early-outs).

Also I fixed a erroneous check in fade operation queueing: non-zero delay was ignored if the fade operation was planned to be immediate.

The easiest way to trigger the [original issue](https://gitlab.com/OpenMW/openmw/issues/4639) in my experience is to minimize the game while the training fade in hasn't been finished and wait for some seconds before going back in. These changes fix the resulting black screen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
     Bug #4384: Resist Normal Weapons only checks ammunition for ranged weapons
     Bug #4411: Reloading a saved game while falling prevents damage in some cases
     Bug #4540: Rain delay when exiting water
+    Bug #4639: Black screen after completing first mages guild mission + training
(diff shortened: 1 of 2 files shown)


akortunov:
Despite I did not manage to reproduce this issue, I assume this PR working. At least, I did not encounter bugs during testing.

PR #2413, opened 2019-06-06
Refactor weapon types behaviour

akortunov:
Replacement of #2134.

Since we do not have a good way to de-hardcode settings right now, just de-hardcode the custom rigs loading and still move weapon settings to the table. Later we can add Lua bindings to it, if needed.

Even in its current state this PR allows to define different placement of sheathing bones for actors with different skeletions (males, females and beast races).
Testing data (animations + skeletons): [Data.zip](https://github.com/OpenMW/openmw/files/3260224/Data.zip)

Once this PR will be ready for merging, I'll contact with Greatness7 to update his mod and update documentation.

### DIFF
--- a/apps/openmw/mwclass/armor.cpp
+++ b/apps/openmw/mwclass/armor.cpp
@@ -20,6 +20,7 @@
 #include "../mwrender/objects.hpp"
 #include "../mwrender/renderinginterface.hpp"
 #include "../mwmechanics/actorutil.hpp"
+#include "../mwmechanics/weapontype.hpp"
 
 #include "../mwgui/tooltips.hpp"
 
@@ -332,21 +333,13 @@ namespace MWClass
             if(*slot == MWWorld::InventoryStore::Slot_CarriedLeft)
             {
                 MWWorld::ConstContainerStoreIterator weapon = invStore.getSlot(MWWorld::InventoryStore::Slot_CarriedRight);
-
-                if(weapon == invStore.end())
-                    return std::make_pair(1,"");
-
-                if(weapon->getTypeName() == typeid(ESM::Weapon).name() &&
-                        (weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::LongBladeTwoHand ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::BluntTwoClose ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::BluntTwoWide ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::SpearTwoWide ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::AxeTwoHand ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::MarksmanBow ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::MarksmanCrossbow))
+                if(weapon != invStore.end() && weapon->getTypeName() == typeid(ESM::Weapon).name())
                 {
-                    return std::make_pair(3,"");
+                    const MWWorld::LiveCellRef<ESM::Weapon> *ref = weapon->get<ESM::Weapon>();
+                    if (MWMechanics::getWeaponType(ref->mBase->mData.mType)->mFlags & ESM::WeaponType::TwoHanded)
+                        return std::make_pair(3,"");
                 }
+
                 return std::make_pair(1,"");
             }
         }
--- a/apps/openmw/mwclass/weapon.cpp
+++ b/apps/openmw/mwclass/weapon.cpp
@@ -17,6 +17,8 @@
 #include "../mwphysics/physicssystem.hpp"
 #include "../mwworld/nullaction.hpp"
 
+#include "../mwmechanics/weapontype.hpp"
+
 #include "../mwgui/tooltips.hpp"
 
 #include "../mwrender/objects.hpp"
@@ -64,8 +66,9 @@ namespace MWClass
     bool Weapon::hasItemHealth (const MWWorld::ConstPtr& ptr) const
     {
         const MWWorld::LiveCellRef<ESM::Weapon> *ref = ptr.get<ESM::Weapon>();
+        int type = ref->mBase->mData.mType;
 
-        return (ref->mBase->mData.mType < ESM::Weapon::MarksmanThrown); // thrown weapons and arrows/bolts don't have health, only quantity
+        return MWMechanics::getWeaponType(type)->mFlags & ESM::WeaponType::HasHealth;
     }
 
     int Weapon::getItemMaxHealth (const MWWorld::ConstPtr& ptr) const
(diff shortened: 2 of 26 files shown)


apps/openmw/mwclass/armor.cpp
```diff
@@ -332,21 +333,13 @@ namespace MWClass
...
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::BluntTwoWide ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::SpearTwoWide ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::AxeTwoHand ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::MarksmanBow ||
-                weapon->get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::MarksmanCrossbow))
+                if(weapon != invStore.end() && weapon->getTypeName() == typeid(ESM::Weapon).name())
                 {
-                    return std::make_pair(3,"");
+                    const MWWorld::LiveCellRef<ESM::Armor> *ref = ptr.get<ESM::Armor>();
+                    if (MWMechanics::getWeaponType(ref->mBase->mData.mType)->mFlags & ESM::WeaponType::TwoHanded)
```
Capostrophic: I don't think this is supposed to recover the *armor* record and its type.

apps/openmw/mwclass/weapon.cpp
```diff
@@ -64,8 +66,9 @@ namespace MWClass
     bool Weapon::hasItemHealth (const MWWorld::ConstPtr& ptr) const
     {
         const MWWorld::LiveCellRef<ESM::Weapon> *ref = ptr.get<ESM::Weapon>();
+        ESM::WeaponType::Class weapClass = MWMechanics::getWeaponType(ref->mBase->mData.mType)->mWeaponClass;
 
-        return (ref->mBase->mData.mType < ESM::Weapon::MarksmanThrown); // thrown weapons and arrows/bolts don't have health, only quantity
+        return (weapClass == ESM::WeaponType::Melee || weapClass == ESM::WeaponType::Ranged); // thrown weapons and arrows/bolts don't have health, only quantity
```
Capostrophic: Maybe it's better to assume that anything that isn't ammunition/thrown weapon has condition. I'm not sure.

apps/openmw/mwmechanics/weaponpriority.cpp
```diff
@@ -90,10 +92,21 @@ namespace MWMechanics
         }
         else if (weapon->mData.mType == ESM::Weapon::MarksmanCrossbow)
         {
-            if (boltRating <= 0.f)
-                rating = 0.f;
-            else
-                rating += boltRating;
+            int ammotype = MWMechanics::getWeaponType(weapon->mData.mType)->mAmmoType;
```
Capostrophic: Special handling of crossbow ammo type seems odd. Maybe instead of only checking if the weapon has crossbow type or bow type *any* ranged weapon should be checked for whether it has arrow ammo or bolt ammo? Basically putting this whole crossbow logic into the else path of "weapclass != ESM::WeaponType::Ranged" check.

apps/openmw/mwrender/npcanimation.cpp
```diff
@@ -275,7 +276,7 @@ static NpcAnimation::PartBoneMap createPartListMap()
     result.insert(std::make_pair(ESM::PRT_LLeg, "Left Upper Leg"));
     result.insert(std::make_pair(ESM::PRT_RPauldron, "Right Clavicle"));
     result.insert(std::make_pair(ESM::PRT_LPauldron, "Left Clavicle"));
-    result.insert(std::make_pair(ESM::PRT_Weapon, "Weapon Bone"));
+    result.insert(std::make_pair(ESM::PRT_Weapon, "Weapon Bone")); // A fallback value, an actual one will depend on current weapon type
```
Capostrophic:
```suggestion
    result.insert(std::make_pair(ESM::PRT_Weapon, "Weapon Bone")); // Fallback. The real node name depends on the current weapon type.
```

apps/openmw/mwworld/projectilemanager.cpp
```diff
@@ -317,7 +318,11 @@ namespace MWWorld
         state.mIdArrow = projectile.getCellRef().getRefId();
         state.mCasterHandle = actor;
         state.mAttackStrength = attackStrength;
-        state.mThrown = projectile.get<ESM::Weapon>()->mBase->mData.mType == ESM::Weapon::MarksmanThrown;
+        state.mThrown = false;
+
+        int type = projectile.get<ESM::Weapon>()->mBase->mData.mType;
+        if (MWMechanics::getWeaponType(type)->mWeaponClass == ESM::WeaponType::Thrown)
+            state.mThrown = true;
```
Capostrophic: I think
```
state.mThrown = MWMechanics::getWeaponType(type)->mWeaponClass == ESM::WeaponType::Thrown;
```
here and on line 613 should do and shouldn't be that long of a line.

apps/openmw/mwclass/weapon.cpp
akortunov: Or maybe add a "hasHealth" flag to weapon type?

akortunov:
@Capostrophic, @AnyOldName3, @kcat, @wareya, any objections or suggestions?

AnyOldName3:
No, but that doesn't necessarily mean I have any idea what's going on with this, as it's mostly `mwmechanics`, which I haven't done much with.

akortunov:
> No, but that doesn't necessarily mean I have any idea what's going on with this, as it's mostly mwmechanics, which I haven't done much with.

Well, first of all, there could be general codestyle issues.
Second, I use DYNAMIC data variance for injected rigs (we seems to use it for other rigs as well, but I can be wrong here). IIRC, you do not like such things.
Third, I wanted to make this table configurable in some way, but I am not sure if we have good pre-1.0 solutions here. But it would be nice to have a bow in left hand (I have an almost full animation set for this feature) or Skyrim-style blunt weapons sheathing (which requires a separate animation group for axes and maces).

AnyOldName3:
> Second, I use DYNAMIC data variance for injected rigs (we seems to use it for other rigs as well, but I can be wrong here). IIRC, you do not like such things.

Is it on a drawable or stateset? If not, it's a flag for the optimiser rather than a flag for rendering, so doesn't actually break things.

akortunov:
> Is it on a drawable or stateset?

On the drawable.

AnyOldName3:
Are you sure? The node is a Node pointer, not a Geometry pointer, so I'd be a little surprised if it was actually the drawable.

akortunov:
> Are you sure? The node is a Node pointer, not a Geometry pointer, so I'd be a little surprised if it was actually the drawable.

Well, at least on the loaded node itselt. I doubt that every bone node has its own stateset.

AnyOldName3:
When I say
> Is it on a drawable or stateset?

I'm asking is it on
* A drawable or a stateset.

or
* Some other kind of node.

If it's a Group or a Transform or whatever, it's not a drawable.

akortunov:
Anyway, I removed DYNAMIC. Users will need to use bones with reserved names.

AnyOldName3:
If it isn't on a drawable, there's no problem with using dynamic, so if it makes things better, don't remove it.

components/esm/loadweap.hpp
```diff
@@ -75,5 +81,34 @@ struct Weapon
     void blank();
     ///< Set record to default state (does not touch the ID).
 };
+
+struct WeaponType
```
Capostrophic: This struct is representation of a new record type (although it isn't exactly implemented in full right now). I'm not sure if it's wise to put it into the technically unrelated weapon record loading header if it will have to be moved anyway.

akortunov: I see no point to put it to the separate file. As you said, there is so such standalone ESM record yet.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -1243,34 +1146,23 @@ bool CharacterController::updateCreatureState()
     return false;
 }
 
-bool CharacterController::updateCarriedLeftVisible(WeaponType weaptype) const
+bool CharacterController::updateCarriedLeftVisible(const int weaptype) const
```
Capostrophic: This function seems kind of redundant to me now.

Capostrophic:
Nothing obvious coming up. I think it's time to prepare for this to be merged?

akortunov:
> I think it's time to prepare for this to be merged?

Which kind of preparation is needed?

apps/openmw/mwmechanics/objects.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "character.hpp"
 #include "movement.hpp"
+#include "weapontype.hpp"
```
Capostrophic: This seems to be unnecessary.

apps/openmw/mwmechanics/weaponpriority.cpp
```diff
@@ -76,24 +78,28 @@ namespace MWMechanics
...
+            else if (ammotype == ESM::Weapon::Bolt)
+            {
+                if (boltRating <= 0.f)
+                    rating = 0.f;
+                else
+                    rating += boltRating;
+            }
         }
 
         if (!weapon->mEnchant.empty())
```
Capostrophic: Projectile weapon type check added here should probably be updated.

PR #2415, opened 2019-06-07
Unix console hotkeys

Assumeru:
Implements ^W (delete word), ^U (delete line), ^LEFT (move past word), and ^RIGHT (move past word) in the in-game console.

Mostly just because I keep trying to use them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,7 @@
     Feature #5036: Allow scripted faction leaving
     Feature #5046: Gamepad thumbstick cursor speed
     Feature #5051: Provide a separate textures for scrollbars
+    Feature #5094: Unix like console hotkeys
     Task #4686: Upgrade media decoder to a more current FFmpeg API
     Task #4695: Optimize Distant Terrain memory consumption
     Task #4789: Optimize cell transitions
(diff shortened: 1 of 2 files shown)


Capostrophic:
Ctrl-Left Arrow and Ctrl-Right Arrow have been [held off](https://gitlab.com/OpenMW/openmw/issues/2648) because they're kind of supposed to be implemented on the side of MyGUI, so that every editable text field in the GUI can react to these combinations.

wareya:
MyGUI fork when?

Assumeru:
> 
> 
> MyGUI fork when?
https://github.com/Assumeru/mygui/commit/294c0c05a27e68a7732d9f3b07549c75870376c8

But I still haven't gotten OpenMW to compile (and run) with that so...

Assumeru:
g++ actually gave me a useful error message unlike Visual Studio, so #2424 fixes my compilation issue and https://github.com/MyGUI/mygui/pull/161 implements stuff in MyGUI.

Capostrophic:
Works fine.
Although this really should be implemented on MyGUI's side too, I'm merging this anyway since MyGUI development... isn't very active... and these shortcuts can always be moved there once it's revived some day.

PR #2418, opened 2019-06-08
Partially revert "Get rid of redundant setKeyFocusWidget() method"

Capostrophic:
Temporary workaround of weird player controls regressions caused by onKeyFocusChanged events no longer being fired when the focus widget didn't change so that the cause of problems with that can be determined isolated from the players. Some other stuff that wasn't necessary anymore and didn't cause issues after it was removed wasn't brought back.


Capostrophic:
I'll go ahead and merge this myself because this is basically how stuff worked a month ago and it can't get more broken than it was while the regression this fixes is rather significant.

PR #2419, opened 2019-06-09
Optimizer: Only merge pure osg::Groups

Capostrophic:
Avoid merging derivative classes because they often have specialized function and behave... badly when their children are stuffed into an unrelated group by the optimizer. Fixes missing lantern glass issues with Morrowind Optimization Patch in my testing.

I'm aware of the overhead of typeid but it's preferable to doing a dynamic cast to osg::LOD and checking if it resulted in nothing because it's a typeid check has both better "universality" and lower overhead. Also something similar is already done in empty node and redundant node visitors and OSG itself.


akortunov:
I suppose we do not need a changelog entry since we have another fixes for optimizer to handle LODs properly.

PR #2421, opened 2019-06-12
[ Is it usefull? ] Hierarchize transforms & Flattenstatic first transform based on isMobile flag

mp3butcher:
2 pr in one as i ran into numeric problem flattening transforms:
- push hierarchic transforms at cell levels (require deep testing, perhaps i missed special transform ajustement)
- add flatten statics objects feature

The latter can be subject of other pr, further it's not as pertinent as i thought on first glance, if you want i remove it:
 performance was due to flatten of animated!! (see screenshot)
perhaps there's improvement on lowerend gpus....don't know

flatten test is:

```
!animated (i forgot this one, it explains mesured perf )
&&!ptr.getClass().isMobile(ptr) 
&& !ptr.getClass().isActivator() 
&& !ptr.getClass().isDoor()
```

fok

### DIFF
--- a/apps/openmw/mwclass/activator.cpp
+++ b/apps/openmw/mwclass/activator.cpp
@@ -33,8 +33,7 @@ namespace MWClass
     {
         if (!model.empty())
         {
-            renderingInterface.getObjects().insertModel(ptr, model, true);
-            ptr.getRefData().getBaseNode()->setNodeMask(MWRender::Mask_Static);
+            renderingInterface.getObjects().insertModel(ptr, model, MWRender::Mask_Static, true);
         }
     }
 
--- a/apps/openmw/mwmechanics/combat.cpp
+++ b/apps/openmw/mwmechanics/combat.cpp
@@ -78,7 +78,7 @@ namespace MWMechanics
         float angleDegrees = osg::RadiansToDegrees(
                     signedAngleRadians (
                     (attacker.getRefData().getPosition().asVec3() - blocker.getRefData().getPosition().asVec3()),
-                    blocker.getRefData().getBaseNode()->getAttitude() * osg::Vec3f(0,1,0),
+                    static_cast<SceneUtil::PositionAttitudeTransform*>(blocker.getRefData().getBaseNode())->getAttitude() * osg::Vec3f(0,1,0),
                     osg::Vec3f(0,0,1)));
 
         const MWWorld::Store<ESM::GameSetting>& gmst = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>();
--- a/apps/openmw/mwmechanics/combat.cpp
+++ b/apps/openmw/mwmechanics/combat.cpp
@@ -75,6 +75,7 @@ namespace MWMechanics
         if (!blocker.getRefData().getBaseNode())
             return false; // shouldn't happen
 
+        assert(!blocker.getRefData().isBaseNodeFlatten());
         float angleDegrees = osg::RadiansToDegrees(
                     signedAngleRadians (
                     (attacker.getRefData().getPosition().asVec3() - blocker.getRefData().getPosition().asVec3()),
--- a/apps/openmw/mwrender/objects.cpp
+++ b/apps/openmw/mwrender/objects.cpp
@@ -5,6 +5,7 @@
 
 #include <components/sceneutil/positionattitudetransform.hpp>
 #include <osg/PositionAttitudeTransform>
+#include <osg/MatrixTransform>
 #include <components/sceneutil/unrefqueue.hpp>
 #include <components/settings/settings.hpp>
 
@@ -21,6 +22,8 @@
 #include "vismask.hpp"
 #include "renderbin.hpp"
 #include "components/sceneutil/optimizer.hpp"
+#include <osgUtil/Optimizer>
+#include <osgUtil/TransformAttributeFunctor>
 
 namespace MWRender
 {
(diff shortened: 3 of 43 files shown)


mp3butcher:
any advise to unhack it is welcome

apps/openmw/mwclass/activator.cpp
```diff
@@ -33,8 +33,7 @@ namespace MWClass
     {
         if (!model.empty())
         {
-            renderingInterface.getObjects().insertModel(ptr, model, true);
-            ptr.getRefData().getBaseNode()->setNodeMask(MWRender::Mask_Static);
+            renderingInterface.getObjects().insertModel(ptr, model, MWRender::Mask_Object, true);
```
mp3butcher: I hack the mask cause activator are mobile...

mp3butcher:
arg...force pushed..I forgot to commit some files

mp3butcher:
wait a little, there's a problem flatten visit doesn't work

mp3butcher:
AHHH Do you feel it my big intelligence damned machine!!!

mp3butcher:
require review and discussion from here

apps/openmw/mwrender/objects.cpp
```diff
@@ -34,54 +44,227 @@ Objects::~Objects()
...
+void Objects::insertModel(const MWWorld::Ptr &ptr, const std::string &mesh, unsigned int mask, bool animated, bool allowLight)
 {
-    insertBegin(ptr);
-    ptr.getRefData().getBaseNode()->setNodeMask(Mask_Object);
+    osg::Group *cellroot = insertBegin(ptr);
+    osg::ref_ptr<SceneUtil::PositionAttitudeTransform> transbasenode = (SceneUtil::PositionAttitudeTransform *) ptr.getRefData().getBaseNode();
+    osg::Group * basenode=transbasenode;
+    osg::ref_ptr<ObjectAnimation> anim;
+    osg::ref_ptr<osg::Group> sub=new osg::Group;
+    if(mask==Mask_Static&&!ptr.getClass().isMobile(ptr)&&!ptr.getClass().isActivator())
```
mp3butcher: require less silly condition...

akortunov:
> Big performance enhancement!!!

Where and how did you measure it? 
I see no noticable difference between this PR and master, especially when shadows are enabled.
The only difference I see - a slightly lesser amount of Drawables in scene when shadows are disabled (about 2-4% depending on scene) without noticable FPS change.

akortunov:
Also items, dropped to the ground, seem to disappear.

mp3butcher:
@akortunov 
I pass from 17 to 25 fps in old ebonheart
Your cpu is to strong you can see anything

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -690,19 +690,57 @@ namespace MWRender
...
+        SceneUtil::PositionAttitudeTransform* ptrans;
+        if(ptr.getRefData().getBaseNode()->getNumParents()>0)
+        {
+            ptrans=dynamic_cast<SceneUtil::PositionAttitudeTransform*>(ptr.getRefData().getBaseNode()->getParent(0));
+            if(ptrans)
+                cellorigin=SceneUtil::getCellOrigin(ptr.getCell()->getCell());
+            else
+            {
+                if(!ptr.getClass().isNpc())
+                    OSG_WARN<<"fuck"<<std::endl;
```
mp3butcher: strange stuff here

mp3butcher:
> Also items, dropped to the ground, seem to disappear.

@akortunov fixed but I may missed other specific adjustemetn

apps/openmw/mwrender/renderingmanager.cpp
mp3butcher: I've fixed this problem in https://github.com/OpenMW/openmw/pull/2421/commits/d4e0663ece86c6e51c41b422303ab0336c61931d
removing 
mObjectRoot->getParent(0)->removeChild(mObjectRoot);
in MWRender::Animation::setObjectRoot
Hope it doesn't mess with the previous purpose of that code

akortunov:
Why do you need to change node masks, by the way? It increases amount of changes a lot.

akortunov:
I now get a lot of "NPC without cell parent" messages in log. Is it intended?

apps/openmw/mwrender/renderingmanager.cpp
mp3butcher: ok it's the player that is initialize before any cell loading....perhaps player should be added in the scenegraph after cell loading

mp3butcher:
@akortunov 
i don't change node mask just harmonise it
I just find out why "NPC without cell parent" : it's the player (attached to mSceneRoot before any file loading)

mp3butcher:
arf have broken the feature

apps/openmw/mwmechanics/combat.cpp
```diff
@@ -75,10 +75,11 @@ namespace MWMechanics
         if (!blocker.getRefData().getBaseNode())
             return false; // shouldn't happen
 
+        assert(!blocker.getRefData().isBaseNodeFlatten());
         float angleDegrees = osg::RadiansToDegrees(
                     signedAngleRadians (
                     (attacker.getRefData().getPosition().asVec3() - blocker.getRefData().getPosition().asVec3()),
-                    blocker.getRefData().getBaseNode()->getAttitude() * osg::Vec3f(0,1,0),
+                    static_cast<SceneUtil::PositionAttitudeTransform*>(blocker.getRefData().getBaseNode())->getAttitude() * osg::Vec3f(0,1,0),
```
akortunov: These static casts are really bad. Probably it would be better to have the `getBaseTransform()` method (which should throw an exception if the value is nullptr).

akortunov:
> i don't change node mask just harmonise it

I do not understand what do you mean by "harmonise". If node masks remained the same, changes in classes are not needed at all (and they are not related to your feature anyway).

apps/openmw/mwclass/activator.cpp
```diff
@@ -33,8 +33,7 @@ namespace MWClass
     {
         if (!model.empty())
         {
-            renderingInterface.getObjects().insertModel(ptr, model, true);
-            ptr.getRefData().getBaseNode()->setNodeMask(MWRender::Mask_Static);
+            renderingInterface.getObjects().insertModel(ptr, model, MWRender::Mask_Static, true);
```
mp3butcher: @akortunov masks do stay the same insertModel set it to Object and it's hacked at exit settings it static...

mp3butcher:
do you want me to remove flatten stuff in order to adress only the cell hierarchization or do i close it and open a new one for this single purpose?

AnyOldName3:
Just to recap as it's been a month, this PR:

1. Makes it so that the root node of each cell has a transform from world space to cell-local space, and each object has a transform from cell-local space to model space (whereas previously each object just had a single transform from world space to model space).
2. Makes it so that the transform from cell-local space to model space gets baked into meshes before they're sent to the GPU if we don't think they're likely to move.

Am I correct?

If so, part 1 is a necessary prerequisite for the batching system I want to implement, so if it can be done tidily without causing problems, that would be good. Part 2 is potentially going to conflict with the future batching system, though, and because things that look like they shouldn't move can move, it might be hard to prove it's bug-free without ending up doing most of the work the proposed batching system would require.

I think we want a new PR for part 1, and to keep this one around but closed so it can be used as reference material later.

mp3butcher:
i agree

PR #2422, opened 2019-06-15
Make sure "Tri Shadow"-named shapes aren't visible (bug #5063)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5063)

This makes the engine and the editor behave mostly behave like TESCS in that shapes that are using this reserved name pattern and "Shadow" pattern in the beginning of the name are not handled even if they're not a part of an actor. I'm not sure this can be adjusted to behave 100% like Morrowind.exe without some awful hacks thrown into the way of how the NIF loader works.


Capostrophic:
This is minor enough that I feel I can merge this without repercussions.

PR #2423, opened 2019-06-15
fix transparent writing depth buffer

mp3butcher:
as transparent bin depth sorts drawables, it doesn't disable depth write...this behavior is implicit in osg and so depthwrite should be disable to comply with nif semantic of "blending" (in which depth write off is not specified: my guess is original engine treat unsorted blended in a peculiar fashion -surely depth peeling-). 
Arguments:
- sorting back2front with depth test don't garantee transparents don't occlude each others (sorting is coarse grain)
- enable depthwrite interferes with Occlusion Queries Node which are based on depth test : transparents write depth so queries coming from behind return not visible (which is not.. because "transparents")
Further it mitigates bug  https://gitlab.com/OpenMW/openmw/issues/3366 too
alphafix

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -1627,13 +1627,15 @@ namespace NifOsg
                         blendFunc = shareAttribute(blendFunc);
                         stateset->setAttributeAndModes(blendFunc, osg::StateAttribute::ON);
                         
-                        osg::ref_ptr<osg::Depth> depth = new osg::Depth(osg::Depth::LESS, 0.0, 1.0, false);
-                        depth = shareAttribute(depth);
-                        stateset->setAttributeAndModes(depth, osg::StateAttribute::ON);
 
                         bool noSort = (alphaprop->flags>>13)&1;
                         if (!noSort)
+                        {
+                            osg::ref_ptr<osg::Depth> depth = new osg::Depth(osg::Depth::LESS, 0.0, 1.0, false);
+                            depth = shareAttribute(depth);
+                            stateset->setAttributeAndModes(depth, osg::StateAttribute::ON);
                             stateset->setRenderingHint(osg::StateSet::TRANSPARENT_BIN);
+                        }
                         else
                             stateset->setRenderBinToInherit();
                     }
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -1626,7 +1626,10 @@ namespace NifOsg
                                                                                    getBlendMode((alphaprop->flags>>5)&0xf)));
                         blendFunc = shareAttribute(blendFunc);
                         stateset->setAttributeAndModes(blendFunc, osg::StateAttribute::ON);
-                        stateset->setAttributeAndModes(new osg::Depth(osg::Depth::LESS, 0.0, 1.0, false));
+                        
+                        osg::ref_ptr<osg::Depth> depth = new osg::Depth(osg::Depth::LESS, 0.0, 1.0, false);
+                        depth = shareAttribute(depth);
+                        stateset->setAttributeAndModes(depth, osg::StateAttribute::ON);
 
                         bool noSort = (alphaprop->flags>>13)&1;
                         if (!noSort)
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -1626,6 +1626,7 @@ namespace NifOsg
                                                                                    getBlendMode((alphaprop->flags>>5)&0xf)));
                         blendFunc = shareAttribute(blendFunc);
                         stateset->setAttributeAndModes(blendFunc, osg::StateAttribute::ON);
+                        stateset->setAttributeAndModes(new osg::Depth(osg::Depth::LESS, 0.0, 1.0, false));
 
                         bool noSort = (alphaprop->flags>>13)&1;
                         if (!noSort)


kcat:
Is this how vanilla behaves? NIFs control the depth-test/mode as well as the alpha blending, so if they intend a specific depth function, we probably shouldn't overwrite that when alpha blending's enabled.

mp3butcher:
as transparent bin depth sort drawable it don't disable depth write...this behavior come from osg

mp3butcher:
but i may be wrong

Capostrophic:
Now, I'm not too knowledgeable about intricacies of depth write, alpha sorting etc., but if this doesn't have any side effects, it should fix [this](https://gitlab.com/OpenMW/openmw/issues/3366) apparently. At least I've noticed a significant reduction in Z-fighting of transparent objects like crystals and Ghostfence.

wareya:
One thing that it does *at least* do is make transparency sorting problems *less obvious*.

akortunov:
> Now, I'm not too knowledgeable about intricacies of depth write, alpha sorting etc., but if this doesn't have any side effects, it should fix this apparently.

Maybe it helps with crystals (did not test it), but I do not see any improvements with other transparent meshes (e.g. lava and ice chunks).

akortunov:
Regarding ice chunks - can we use a custom depth test for water to avoid Z-fighting with objects near water surface (swamps, ice chunks, shore terrain, etc.)? Basically, water should have a lower priority than other objects.

wareya:
The exterior water surface is already offset slightly if that's what you mean.

akortunov:
> The exterior water surface is already offset slightly if that's what you mean.

I mean that the water z-fights with other objects, especially on large distances:

Screenshot_20190618_171254

In such cases we should render conflicting object over water surface.

mp3butcher:
in all the cases coarse sorting (TRANPARENT_BN) can violate the alpha composition function (ONE_MINUS_SRC_ALPHA - SRC_ALPHA can be one of a nearer fragment) ,as well  as DEPTH_TEST if drawable bounds overlap  SRC_ALPHA and "SRC_DEPTH" can be one of a nearer fragment...
no depth write is a compromise to reduce transparent zfighting and indesired occlusion

mp3butcher:
completely disabling depth test for transprent may actually be even better

akortunov:
I doubt that it is a good idea:

Screenshot_20190618_174232

mp3butcher:
here it is force pushed back

wareya:
Z-fighting between distant geometry is a different problem.

AnyOldName3:
Typically (as in, in most software, but not necessarily Morrowind), when alpha blending and depth sorting are on, depth writes are off. There are times when having them on makes sense (not that things making sense has any correlation with what Bethesda do), so we should confirm what Morrowind does. I've had a peek at some leaves in `meshes/f/flora_tree_01.nif` and they have a NiAlphaProperty but no NiZBufferProperty. It could be that:
* Turning on translucency/depth sorting with a NiAlphaProperty implicitly turns off depth writes. If so, we need to see how this interacts with an explicit NiZBufferProperty, and which of blending and depth sorting actually triggers this condition.
* Bethesda are Bethesda and just *had* to be weird, so wanted depth writes on for tree leaves, and we're supposed to obey only the NiZBufferProperty.

My expectation is that it's the first option, but Bethesda are Bethesda, so we need to check.

wareya:
If you do depth writes, do you also do depth tests? Like if you had transparent textures that were completely opaque, enabling both depth tests and depth writes would allow exactly the parts that completely opaque to occlude correctly, right? The transparent parts would always glitch out no matter what, if they were occluding in complex ways. It seems to me like it wouldn't be truly universal to disable depth writes when doing depth sorting.

kcat:
> Bethesda are Bethesda and just had to be weird, so wanted depth writes on for tree leaves, and we're supposed to obey only the NiZBufferProperty.

I think it's more that they just never bothered to actively turn it off. Since most renderers will handle translucent surfaces after opaque ones, and draw them sorted, the fact that depth-writes were left on rarely became an issue. All the opaque stuff is done, and all the translucent stuff in the back is done first, so what does it matter if the depth buffer gets written to. It would only crop up when translucent surfaces are intersecting, and that's going to be a problem either way you do it.

wareya:
Oh yeah, of course they still do depth tests, they have to occlude with opaque stuff too. Probably shouldn't post when I just woke up.

AnyOldName3:
> It would only crop up when translucent surfaces are intersecting, and that's going to be a problem either way you do it.

The artefacts with writes on are generally considered worse than those with writes off. I'm struggling to find a source that actually says why, though.

Short-term, we should double-check what Morrowind does and implement that.

Medium term, I'll get alpha-to-coverage working and for people using MSAA, things will magically start working without sorting.

Long term, we can look at order-independent solutions like multiplicative blending (which isn't exactly physically accurate, but doesn't really have artefacts if what you're trying to draw is mostly emissive or transmissive rather than reflective) or using modern GPU features like shader atomics to implement per-pixel linked-lists of transparent fragments, and sorting these on the GPU as an extra pass.

AnyOldName3:
Okay, so I have confirmation that the vanilla engine leaves depth writes on when drawing the leaves of `meshes/f/flora_tree_01.nif`, so disabling depth writes isn't implicit with depth-sorted alpha blending, and needs to be set separately with a `NiZBufferProperty`. That means that this PR changes behaviour to diverge from vanilla, and should probably be closed.

Unfortunately, that breaks occlusion queries in their current form. I guess the best solution there would be to issue queries after the opaque bin but before the translucent bin, as that should be relatively easy to do without breaking anything...

kcat:
> Unfortunately, that breaks occlusion queries in their current form. I guess the best solution there would be to issue queries after the opaque bin but before the translucent bin, as that should be relatively easy to do without breaking anything...

Until something gets drawn with alpha=1.0, or has a texture where a big part is alpha=1.0 (or some other really high value), where it's effectively opaque but not technically.

The proper fix would be to make a mod with models that avoid translucency, preferring alpha testing whenever possible, and setting correct `NiZBufferProperty` values where depth writes should be off. A similar issue exists with shadows, since the default tree model has leaves drawn with translucency and the shadow camera can't see them properly. Mods fix that by using replacement trees with alpha-tested leaves, so they generate correct depth information (which is not only more efficient, but works with occlusion queries and anything else that needs proper depth info).

AnyOldName3:
> Until something gets drawn with alpha=1.0, or has a texture where a big part is alpha=1.0 (or some other really high value), where it's effectively opaque but not technically.

Occlusion queries are already rough - you're testing against previous frames based on bounding boxes, so you're already drawing some things that are actually occluded (and potentially skipping drawing things that should be visible). It's not the end of the world if things can't be occluded by leaves.

> A similar issue exists with shadows, since the default tree model has leaves drawn with translucency and the shadow camera can't see them properly.

For shadows, alpha testing everything and blending nothing is the correct thing to do as you don't care about colour at all, just whether light mostly passes through or mostly gets blocked. Depth buffers only have one value per pixel, so partial transparency isn't even a thing you can do with shadow mapping. The threshold at which to keep things is the only debatable part, and 0.5 was picked as it showed leaf shapes the best.

kcat:
> For shadows, alpha testing everything and blending nothing is the correct thing to do as you don't care about colour at all, just whether light mostly passes through or mostly gets blocked. Depth buffers only have one value per pixel, so partial transparency isn't even a thing you can do with shadow mapping.

That's what I mean. A depth value can't be translucent, but a potential shadow caster can be. Trees, for example, use translucency for leaves, to create a see-through effect for the main camera. Like occlusion queries, a shadow camera can only see whether a there's been a pixel rendered (at some depth), completely ignoring visual/color translucency.

mp3butcher:
> Okay, so I have confirmation that the vanilla engine leaves depth writes on when drawing the leaves of `meshes/f/flora_tree_01.nif`, so disabling depth writes isn't implicit with depth-sorted alpha blending, and needs to be set separately with a `NiZBufferProperty`. That means that this PR changes behaviour to diverge from vanilla, and should probably be closed.

I'm not sure about nif of tree leaves but what i know is that leaves doesn't require blendfunc...alpha test is sufficient.
So This pr is valid

mp3butcher:
...and if I'm wrong and tree leaves use blendfunc then depth write must be disable for proper alpha composition

mp3butcher:
now you can't be unhappy with your strange tree leaves

mp3butcher:
@AnyOldName3 
my guess is original engine (vanilla?) treats unsorted blended object in a peculiar fashion -surely depth peeling-

AnyOldName3:
> @AnyOldName3
> 
> > Okay, so I have confirmation that the vanilla engine leaves depth writes on when drawing the leaves of `meshes/f/flora_tree_01.nif`, so disabling depth writes isn't implicit with depth-sorted alpha blending, and needs to be set separately with a `NiZBufferProperty`. That means that this PR changes behaviour to diverge from vanilla, and should probably be closed.
> 
> I'm not sure about nif of tree leaves but what i know is that leaves doesn't require blendfunc...alpha test is sufficient.

In theory, it's fine to draw tree leaves with just testing, but we're not trying to draw trees as fast as possible - we're trying to do exactly what `Morrowind.exe` does when it's drawing a particular Nif file, and when it's drawing `meshes/f/flora_tree_01.nif`, it sees it has a `NiAlphaProperty`, and sees that the `NiAlphaProptery` has blending enabled with the source set to source alpha and the destination set to one minus source alpha, alpha testing turned off, and sorting turned on, then it sees that there's no `NiZBufferProperty`, so it uses the default settings for a Nif file, which includes having depth writes on.

I've looked at what the DirectX rendering state is when `Morrowind.exe` renders those leaves, and it matches our (and NifSkope's) interpretation of what the Nif file means, so that's absolutely what we have to do when drawing the same mesh.

> ...and if I'm wrong and tree leaves use blendfunc then depth write must be disable for proper alpha composition
> how can blending can be right if leaves occlude each other?

When Morrowind was being designed, Michael Kirkbride said to Todd Howard “Man, this is really boring. How about we put it in a volcano with like giant bugs everywhere?”

Todd Howard misunderstood, and didn't realise Kirkbride was talking about insects, so made sure the game has many catastrophic bugs all over the place.

In all seriousness, Morrowind does lots and lots of weird things that (in many cases) no other game engine does, and often which didn't even make sense when the game was made, either. Unfortunately, many of the game's assets include workarounds for these issues, and when you try and fix things too much, the workarounds make things worse than the bugs they covered up. If we ignore what a Nif file says about how it should be rendered and do something else instead, we're creating more potential bugs.

If you don't like how Morrowind's assets are, the correct thing to do is to release a mod containing fixed versions of the assets, not to change the engine to try and patch things on-the-fly.

> i consider that non sorted blended (ex:tree leaves) are depthwrite(true) (don't care proper alphacomposition) but sorted blended are depthwrite(false) (proper alphacomposition)
> 
> hence I moved the code in the sorted case so you can't be unhappy with your strange tree leaves (but i'm not happy with this change as it doesn't minimize bad alpha composition as previously)

Tree leaves *are* depth sorted. They *are* blended. They still do depth writes. The *don't sort* flag doesn't control the depth writes. Only a `NiZBufferProperty` controls whether depth writes occur.

> @AnyOldName3
> my guess is original engine (vanilla?) treats unsorted blended object in a peculiar fashion -surely depth peeling- and so require depth write for one of the 2 passes

Depth peeling was first publicised by Nvidia at GDC 2001, and Morrowind released a whole fourteen months after that, so it wouldn't be technically impossible, but it's very rare for state-of-the-art techniques to make it into retail games that quickly. Either way, I've looked at how the leaves are drawn in APITrace, and `Morrowind.exe` definitely isn't doing depth peeling, and if it were, there'd be no reason for the *don't sort* flag to be unset as the depth peeling would make the sorting redundant.

The original engine just does what the Nif file says, and the Nif file says 
image
and has no `NiZBufferProperty` turning depth writes off.

mp3butcher:
would it be plausible that NiAlphaProperty have implicit NiZBufferProperty?

AnyOldName3:
As I said in the above comment, I looked with APITrace and determined that that isn't what the vanilla engine does.

PR #2426, opened 2019-06-18
Logarithmic depth buffer

akortunov:
Here are results of my experiments with these articles:

https://outerra.blogspot.com/2009/08/logarithmic-z-buffer.html
https://outerra.blogspot.com/2012/11/maximizing-depth-buffer-range-and.html
https://www.gamedev.net/blogs/entry/2006307-tip-of-the-day-logarithmic-zbuffer-artifacts-fix
https://stackoverflow.com/questions/18182139/logarithmic-depth-buffer-linearization

**Since it implemented via shaders, I suppose it will have no effect in the fixed pipeline mode.
Also you do not need to alter the "near clip" setting value.**

In my testing it worked quite good:

Screenshot_20190618_212105

And here is what we have now in master branch:
Screenshot_20190618_212127

Unfortunately, there are issues. What I found so far:
~~1. Minimap does not render some objects (the bug is visible on the first screenshot).~~
2. There is a white outline when looking horizontally, presumably because of water shader.
3. There is a quite large performance drop (under GPU section). Maybe there are ways to optimize calculations...
4. TCB does not work well - lines are black for some reason instead of white.

I can not fix these issues by myself (since I am not familiar with GLSL at all).
Casting @AnyOldName3, @kcat and @Capostrophic here. Can we improve this approach, or it is not suitable for us?

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -583,7 +583,7 @@ namespace MWPhysics
 
         if (mDebugDrawEnabled && !mDebugDrawer.get())
         {
-            mDebugDrawer.reset(new MWRender::DebugDrawer(mParentNode, mCollisionWorld));
+            mDebugDrawer.reset(new MWRender::DebugDrawer(mParentNode, mCollisionWorld, mResourceSystem));
             mCollisionWorld->setDebugDrawer(mDebugDrawer.get());
             mDebugDrawer->setDebugMode(mDebugDrawEnabled);
         }
--- a/apps/openmw/mwrender/bulletdebugdraw.cpp
+++ b/apps/openmw/mwrender/bulletdebugdraw.cpp
@@ -2,6 +2,9 @@
 
 #include <BulletCollision/CollisionDispatch/btCollisionWorld.h>
 
+#include <components/shader/shadermanager.hpp>
+#include <components/resource/scenemanager.hpp>
+
 #include <osg/Geometry>
 #include <osg/Group>
 
@@ -20,10 +23,11 @@ namespace
 namespace MWRender
 {
 
-DebugDrawer::DebugDrawer(osg::ref_ptr<osg::Group> parentNode, btCollisionWorld *world)
+DebugDrawer::DebugDrawer(osg::ref_ptr<osg::Group> parentNode, btCollisionWorld *world, Resource::ResourceSystem* resourceSystem)
     : mParentNode(parentNode),
       mWorld(world),
-      mDebugOn(true)
+      mDebugOn(true),
+      mResourceSystem(resourceSystem)
 {
 
     createGeometry();
@@ -35,6 +39,17 @@ void DebugDrawer::createGeometry()
     {
         mGeometry = new osg::Geometry;
         mGeometry->setNodeMask(Mask_Debug);
+
+        std::map<std::string, std::string> defineMap;
+
+        Shader::ShaderManager& shaderMgr = mResourceSystem->getSceneManager()->getShaderManager();
+        osg::ref_ptr<osg::Shader> fragmentShader (shaderMgr.getShader("tcb_fragment.glsl",defineMap, osg::Shader::FRAGMENT));
+        osg::ref_ptr<osg::Shader> vertexShader (shaderMgr.getShader("tcb_vertex.glsl", defineMap, osg::Shader::VERTEX));
+
+        osg::ref_ptr<osg::Program> program (new osg::Program);
+        program->addShader(fragmentShader);
+        program->addShader(vertexShader);
+        mGeometry->getOrCreateStateSet()->setAttributeAndModes(program, osg::StateAttribute::ON);
 
         mVertices = new osg::Vec3Array;
 
@@ -74,6 +89,7 @@ void DebugDrawer::step()
         mDrawArrays->setCount(mVertices->size());
         mVertices->dirty();
         mGeometry->dirtyBound();
+        mResourceSystem->getSceneManager()->recreateShaders(mGeometry);
     }
 }
 
--- a/apps/openmw/mwrender/bulletdebugdraw.hpp
+++ b/apps/openmw/mwrender/bulletdebugdraw.hpp
@@ -7,6 +7,8 @@
 
 #include <LinearMath/btIDebugDraw.h>
 
+#include <components/resource/resourcesystem.hpp>
+
 class btCollisionWorld;
 
 namespace osg
@@ -32,9 +34,11 @@ protected:
     void createGeometry();
     void destroyGeometry();
 
+    Resource::ResourceSystem* mResourceSystem;
+
 public:
 
-    DebugDrawer(osg::ref_ptr<osg::Group> parentNode, btCollisionWorld *world);
+    DebugDrawer(osg::ref_ptr<osg::Group> parentNode, btCollisionWorld *world, Resource::ResourceSystem* resourceSystem);
     ~DebugDrawer();
 
     void step();
--- a/apps/openmw/mwrender/localmap.cpp
+++ b/apps/openmw/mwrender/localmap.cpp
@@ -171,7 +171,7 @@ osg::ref_ptr<osg::Camera> LocalMap::createOrthographicCamera(float x, float y, f
 
     camera->setProjectionMatrixAsOrtho(-width/2, width/2, -height/2, height/2, 5, (zmax-zmin) + 10);
     camera->setComputeNearFarMode(osg::Camera::DO_NOT_COMPUTE_NEAR_FAR);
-    camera->setViewMatrixAsLookAt(osg::Vec3d(x, y, zmax + 5), osg::Vec3d(x, y, zmin), upVector);
+    camera->setViewMatrixAsLookAt(osg::Vec3d(x, y, zmax + 1000), osg::Vec3d(x, y, zmin), upVector);
     camera->setReferenceFrame(osg::Camera::ABSOLUTE_RF);
     camera->setRenderTargetImplementation(osg::Camera::FRAME_BUFFER_OBJECT, osg::Camera::PIXEL_BUFFER_RTT);
     camera->setClearColor(osg::Vec4(0.f, 0.f, 0.f, 1.f));
--- a/files/shaders/CMakeLists.txt
+++ b/files/shaders/CMakeLists.txt
@@ -7,6 +7,8 @@ set(SDIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(DDIRRELATIVE resources/shaders)
 
 set(SHADER_FILES
+    tcb_vertex.glsl
+    tcb_fragment.glsl
     water_vertex.glsl
     water_fragment.glsl
     water_nm.png
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -61,6 +61,8 @@ varying vec3 passNormal;
 
 void main()
 {
+    gl_FragDepth = (log(gl_TexCoord[6].z + 1.0) / log(1000000000.0 + 1.0));
+
 #if @diffuseMap
     vec2 adjustedDiffuseUV = diffuseMapUV;
 #endif
--- a/files/shaders/objects_vertex.glsl
+++ b/files/shaders/objects_vertex.glsl
@@ -53,6 +53,7 @@ varying vec3 passNormal;
 void main(void)
 {
     gl_Position = gl_ModelViewProjectionMatrix * gl_Vertex;
+    gl_TexCoord[6] = gl_Position;
     depth = gl_Position.z;
 
     vec4 viewPos = (gl_ModelViewMatrix * gl_Vertex);
--- a/files/shaders/tcb_fragment.glsl
+++ b/files/shaders/tcb_fragment.glsl
@@ -1,0 +1,6 @@
+#version 120
+
+void main()
+{
+    gl_FragDepth = (log(gl_TexCoord[6].z + 1.0) / log(1000000000.0 + 1.0));
+}
--- a/files/shaders/tcb_vertex.glsl
+++ b/files/shaders/tcb_vertex.glsl
@@ -1,0 +1,7 @@
+#version 120
+
+void main(void)
+{
+    gl_Position = gl_ModelViewProjectionMatrix * gl_Vertex;
+    gl_TexCoord[6] = gl_Position;
+}
--- a/files/shaders/terrain_fragment.glsl
+++ b/files/shaders/terrain_fragment.glsl
@@ -31,6 +31,8 @@ varying vec3 passNormal;
 
 void main()
 {
+    gl_FragDepth = (log(gl_TexCoord[6].z + 1.0) / log(1000000000.0 + 1.0));
+
     vec2 adjustedUV = (gl_TextureMatrix[0] * vec4(uv, 0.0, 1.0)).xy;
 
 #if @normalMap
--- a/files/shaders/terrain_vertex.glsl
+++ b/files/shaders/terrain_vertex.glsl
@@ -21,6 +21,7 @@ varying vec3 passNormal;
 void main(void)
 {
     gl_Position = gl_ModelViewProjectionMatrix * gl_Vertex;
+    gl_TexCoord[6] = gl_Position;
     depth = gl_Position.z;
 
     vec4 viewPos = (gl_ModelViewMatrix * gl_Vertex);
--- a/files/shaders/water_fragment.glsl
+++ b/files/shaders/water_fragment.glsl
@@ -242,7 +242,7 @@ void main(void)
 #if REFRACTION
     float depthSample = linearizeDepth(texture2D(refractionDepthMap,screenCoords).x);
     float depthSampleDistorted = linearizeDepth(texture2D(refractionDepthMap,screenCoords-(normal.xy*REFR_BUMP)).x);
-    float surfaceDepth = linearizeDepth(gl_FragCoord.z);
+    float surfaceDepth = linearizeDepth(gl_FragDepth);
     float realWaterDepth = depthSample - surfaceDepth;  // undistorted water depth in view direction, independent of frustum
     float shore = clamp(realWaterDepth / BUMP_SUPPRESS_DEPTH,0,1);
 #else
--- a/files/shaders/water_fragment.glsl
+++ b/files/shaders/water_fragment.glsl
@@ -153,13 +153,14 @@ float frustumDepth;
 
 float linearizeDepth(float depth)
   {
-    float z_n = 2.0 * depth - 1.0;
-    depth = 2.0 * near * far / (far + near - z_n * frustumDepth);
+    depth = exp(depth*log(1000000000.0+1.0)) - 1;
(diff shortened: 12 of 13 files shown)


akortunov:
From what I can tell, the performance drop is mostly because of `gl_FragDepth` since it disables many optimizations, but I need it to avoid artifacts (disappearing objects near camera). Have we a better solution?

mp3butcher:
what are you trying to fix with that: writing gl_FragDepth is not the bottleneck
it's the use of log for each fragment (even those not displayed)... that make this costy

mp3butcher:
a polygonoffset may be sufficient
https://github.com/OpenMW/openmw/pull/2427
Capture d’écran de 2019-06-19 01-06-31

mp3butcher:
you could get ride of gl_TexCoord[6] using builtin gl_FragPosition

kcat:
I  wouldn't use the shader approach, but what I [described in the forums](https://forum.openmw.org/viewtopic.php?f=2&t=4219&p=46595). Basically, with the GL_ARB_clip_control extension, map the depth range to 0...1 instead of OpenGL's default -1...+1, use a floating-point depth buffer, and reverse the depth values (so the near plane is 1 and the far plane is 0). Floating-point values have greater precision around 0, which neatly counteracts the reduced precision from distant 1/z values.

Or alternatively, use glDepthRangedNV (supported on AMD too) to reverse the depth range.

Or if you must go with the shader approach, the Outtera blog mentioned using GL_ARB_conservative_depth to restore some of the lost depth optimization. The extension is basically a way to promise that the new corrected depth value you write will always be farther (or closer) that the initial value it would have written, so if the new initial depth value is already farther (or closer) than what's in the depth buffer, the corrected value you write would still fail the depth test and the pixel still gets discarded early.

akortunov:
> Basically, with the GL_ARB_clip_control extension, map the depth range to 0...1 instead of OpenGL's default -1...+1, use a floating-point depth buffer, and reverse the depth values (so the near plane is 1 and the far plane is 0). 

I already googled about reverse float depth buffers, but implementations which I found rely on the NVidia's OpenGL extensions, which some hardware may not support (e.g. Intel GPU's, and mobile devices which use the GL ES). I would like to have a platform-independent solution instead.

> Or if you must go with the shader approach, the Outtera blog mentioned using GL_ARB_conservative_depth to restore some of the lost depth optimization.

From what I can tell, the conservative depth is the OpenGL 4.x feature (OpenGL 1.2 seems to do not support the `layout` instruction).

kcat:
> I already googled about reverse float depth buffers, but implementations which I found rely on the NVidia's OpenGL extensions, which some hardware may not support (e.g. Intel GPU's, and mobile devices which use the GL ES). I would like to have a platform-independent solution instead.

That's what GL_ARB_clip_control can do. Being an ARB extension, it's not an nVidia extension. In fact, one of the main purposes for that extension is to improve depth range with floating-point depth buffers like Direct3D can do by default. As per [the spec](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_clip_control.txt):
> This extension's goal is to 1) allow OpenGL to effectively match Direct3D's coordinate system conventions, and 2) potentially improve the numerical precision of the Z coordinate mapping.
> ...
> 14) Will using the GL_ZERO_TO_ONE clip control depth mode improve my depth precision?
>
> RESOLVED:  Yes, if you use a floating-point depth buffer and place the near and far values at 1.0 and 0.0 (reversed from the normal convention).

The spec even references the Outerra blog for logarithmic depth buffers along with a few others (see issue 7). It is a somewhat newer extension, but given reasonably up-to-date drivers, there's little reason it can't be supported on old hardware. All it's doing is letting the hardware work like it would under Direct3D, with regard to clip space. It's merely exposing what hardware already does to support both APIs.

mp3butcher:
Other idea: why don't you use a postrender to frame buffer camera with the player and let osg do his magic for the near plane narowing?

akortunov:
> Other idea: why don't you use a postrender to frame buffer camera with the player and let osg do his magic for the near plane narowing?

And how to do it?

mp3butcher:
discord

kcat:
> Other idea: why don't you use a postrender to frame buffer camera with the player and let osg do his magic for the near plane narowing for the scene?

That doesn't fix the issue, it just moves it. As long as there's anything close to the camera while the distance is also viewable, the problem will still crop up. Worse, making the near-plane dynamic will make the issue less consistent.

akortunov:
Are there ways to use the GL_ARB_conservative_depth with OpenGL 1.2?

kcat:
No because it's a shader thing. OpenGL 1.2 doesn't have shaders, so you can't change gl_FragDepth in a shader to make any difference anyway. Regardless, don't worry about OpenGL 1.2, given that it's not even at the D3D8-level capabilities vanilla used (D3D8 is probably closer to GL1.5, and D3D9 is closer to GL2.1, give or take a few extensions).

akortunov:
> Regardless, don't worry about OpenGL 1.2

What do you mean by "don't worry about OpenGL 1.2"?

kcat:
> What do you mean by "don't worry about OpenGL 1.2"?

I mean don't worry about devices that only support OpenGL 1.2. That version was released in early 1998, there's no reason to limit what we use to that.

akortunov:
All right, a different question - are there ways to use the GL_ARB_conservative_depth with GLSL 1.2, which OpenMW uses?

kcat:
If the extension exists in the `GL_EXTENSIONS` list, you can use it. If the extension is available, you just put
```
#extension GL_ARB_conservative_depth : enable

layout(depth_less) out float gl_FragDepth;
```
into the fragment shader to use it (changing `depth_less` if needed/wanted).

akortunov:
I tried this upproach yesterday, but it gives such results:

```
0(5) : error C0000: syntax error, unexpected '(', expecting "::" at token "("
```

I suspect that it does not know about what the `depth_less` is.

kcat:
Is `GL_ARB_conservative_depth` in the extensions list? It's possible there's another extension you need to explicitly enable for it to accept the `layout(...)` syntax. I don't know off-hand which one, though.

Capostrophic:
The extension *precisely* gives meaning to depth_less and other qualifiers. Though it normally requires OpenGL3+, not sure if it has to be supported on earlier versions.

akortunov:
> Though it normally requires OpenGL3+

More likely it is the root of issue.

akortunov:
> you could get rid of gl_TexCoord[6] using builtin gl_FragCoord

Unfortunately, gl_FragCoord != gl_Position.

mp3butcher:
> Unfortunately, gl_FragCoord != gl_Position.

gl_FragCoord.z/gl_FragCoord.w=gl_TexCoord[6].z

akortunov:
> gl_FragCoord.z*gl_FragCoord.w==gl_TexCoord[6].z

No, it does not improve performance. As I said, performance drop is not because of calculations. It is because of `gl_FragDepth` which disables many optimizations.

Can be considered to be only a refactoring change.

wareya:
I know I already said this but have we considered rendering to a framebuffer with a 32-bit depth buffer? I checked the ingame opengl state with apitrace and it showed the depth buffer format being 24-bit.

akortunov:
>  I checked the ingame opengl state with apitrace and it showed the depth buffer format being 24-bit.

I suppose we explicitely tell OSG to use a 24-bit depth buffer.

wareya:
Isn't that only in the refraction shader?

akortunov:
> Isn't that only in the refraction shader?

I have no idea, but according to docs, the `depth` field of osg::GraphicsContext::Traits represents a depth buffer format. Can you try to use depth = 32 in the engine.cpp and check with API trace will show different results?

wareya:
Ah, I see. I'll test that.

wareya:
GL_DEPTH_BITS is 24, and the actual depth attachment is GL_DEPTH_COMPONENT (i.e. implementation-defined default).

image
image

I'm not completely sure what the depth buffer's Actual Format is, but it seems to be 24-bit. Example values are 0.999962986 (exactly 16776594 / 2^24), 0.999962986 (exactly 16776595 / 2^24), and 0.999993205 (exactly 16777102 / 2^24).

mp3butcher:
depth buffer is packed with stencil buffer 24 and 8 bits, you can't change depth to 32 bits because nif use stencil test

mp3butcher:
> > gl_FragCoord.z*gl_FragCoord.w==gl_TexCoord[6].z
> 
> No, it does not improve performance. As I said, performance drop is not because of calculations. It is because of `gl_FragDepth` which disables many optimizations.
> 
> Can be considered to be only a refactoring change.
get rid of gl_Texcoord[6] is not about computation but a bandwidth saver

kcat:
> depth buffer is packed with stencil buffer 24 and 8 bits, you can't change depth to 32 bits because nif use stencil test

You can have a 32-bit depth buffer with 8-bit stencil. The depth/stencil values won't get packed with 32/8 like 24/8 can, but nowadays hardware doesn't pack it that way anyway, instead keeping the depth and stencil values separate (this could also mean 32-bit depth can end up being slightly faster, since packed 24-bit values have poor alignment properties; same reason we have 32-bit display buffers when there's only 24 bits of color).

The trick will be to make sure the 32-bit depth buffer is floating-point, rather than integer. I can't recall how that's done.

mp3butcher:
@kcat you're right ...i'm too used to packed 24/8

wareya:
>The trick will be to make sure the 32-bit depth buffer is floating-point, rather than integer. I can't recall how that's done.

I think that it's not possible to ensure it one way or another. AFAIK devices only have to support one format per bit depth. They might support 32i or 32f, but not the other, or something.

mp3butcher:
other lead :https://github.com/XenonofArcticus/OSG-Transparency-Tool

mp3butcher:
perhaps depth_peeling would absorb zfight :https://github.com/XenonofArcticus/OSG-Transparency-Tool

AnyOldName3:
> From what I can tell, the performance drop is mostly because of gl_FragDepth since it disables many optimizations, but I need it to avoid artifacts (disappearing objects near camera). Have we a better solution?

It's going to be worse for you than other people as you're on a post-Maxwell Nvidia card (IIRC a 1060 or 960) - with Maxwell, Nvidia switched to tiled rendering (like on a mobile GPU) which provides big performance improvements when you don't meddle with the fragment depth or alpha testing, but has a big performance hit if you do.

> > The trick will be to make sure the 32-bit depth buffer is floating-point, rather than integer. I can't recall how that's done.
> 
> I think that it's not possible to ensure it one way or another. AFAIK devices only have to support one format per bit depth. They might support 32i or 32f, but not the other, or something.

The built-in `GL_DEPTH_COMPONENT`, `GL_DEPTH_COMPONENT24` and `GL_DEPTH_COMPONENT32` are always fixed-point representations. Support for floating-point versions comes from extensions (or later OpenGL versions), and different extensions are required for different situations. For example, to use a floating-point depth buffer for the main framebuffer on Windows, you need `WGL_EXT_depth_float`, which gives `WGL_DEPTH_FLOAT_EXT` as a constant you can pass during OpenGL context creation (SDL will probably handle this for us in a platform-independent way). To use it on a different framebuffer, you can use something like `GL_ARB_depth_buffer_float` which gives `DEPTH_COMPONENT32F`, or the older Nvidia version of the extension (supported on some non-Nvidia hardware), or check for OpenGL 3 support.

wareya:
Well, apparently in GL 4.5, the situation is the other way around, GL_DEPTH_COMPONENT32F is required and GL_DEPTH_COMPONENT32 is not.

https://gamedev.stackexchange.com/questions/168241/is-gl-depth-component32-deprecated-in-opengl-4-5

akortunov:
@AnyOldName3, any ideas how to implement this feature properly?

AnyOldName3:
Logarithmic depth buffers require either really, really triangle-dense meshes (so you can't see interpolation artefacts) or meddling with things with shaders that disable a bunch of optimisations, so I'm more in favour of going for floating-point depth buffers.

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -583,7 +583,7 @@ namespace MWPhysics
 
         if (mDebugDrawEnabled && !mDebugDrawer.get())
         {
-            mDebugDrawer.reset(new MWRender::DebugDrawer(mParentNode, mCollisionWorld));
```
AnyOldName3: Ignore this review, I'm just testing a webhook.

AnyOldName3: Okay, what about now?

AnyOldName3: Maybe if I finish the review?

AnyOldName3:
Is it doing PR comments that aren't part of a review?

AnyOldName3:
And now?

akortunov:
> I'm more in favour of going for floating-point depth buffers.

Do you know how to implement it? The only example I found is an [old OSG example](https://github.com/openscenegraph/OpenSceneGraph/blob/master/examples/osgfpdepth/osgfpdepth.cpp) (which seems to do not use the reverse-Z), but it looks too complicated to me.
I suppose that it is not enough just attach depth texture to camera,
```
    osg::Texture2D *depthTex = new osg::Texture2D;
    depthTex->setTextureSize(width, height);
    depthTex->setSourceFormat(GL_DEPTH_COMPONENT);
    depthTex->setSourceType(GL_FLOAT);
    depthTex->setInternalFormat(internalFormat);
    depthTex->setFilter(osg::Texture2D::MIN_FILTER, osg::Texture2D::NEAREST);
    depthTex->setFilter(osg::Texture2D::MAG_FILTER, osg::Texture2D::NEAREST);
    depthTex->setWrap(osg::Texture::WRAP_S, osg::Texture::CLAMP_TO_EDGE);
    depthTex->setWrap(osg::Texture::WRAP_T, osg::Texture::CLAMP_TO_EDGE);
    camera->attach(osg::Camera::DEPTH_BUFFER, depthTex);
```
right?

AnyOldName3:
`internalFormat` would need to be something like `GL_DEPTH_COMPONENT32F`, and I don't think setting stuff like wrapping and filtering is necessary, but otherwise, that should work. Setting up reverse-Z might be as simple as setting some parameters on an `osg::Depth`.

akortunov:
> internalFormat would need to be something like GL_DEPTH_COMPONENT32F

Yes, it is the GL_DEPTH_COMPONENT32F, but I see no difference between floating buffer and out current implementation.

akortunov:
@AnyOldName3, I also tried to use [this](https://nlguillemot.wordpress.com/2016/12/07/reversed-z-in-opengl) article, but without success:

```
    osg::ref_ptr<osg::Camera> camera = mViewer->getCamera();
    camera->setGraphicsContext(graphicsWindow);
    camera->setViewport(0, 0, traits->width, traits->height);

    osg::ref_ptr<osg::Texture2D> depthTexture = new osg::Texture2D;
    depthTexture->setTextureSize(traits->width, traits->height);
    depthTexture->setSourceFormat(GL_DEPTH_COMPONENT);
    depthTexture->setSourceType(GL_FLOAT);
    depthTexture->setInternalFormat(GL_DEPTH_COMPONENT32F);
    camera->detach(osg::Camera::DEPTH_BUFFER);
    camera->attach(osg::Camera::DEPTH_BUFFER, depthTexture.get());
    camera->setClearDepth(0);
```

and 

```
    osg::ref_ptr<osg::Group> rootNode (new osg::Group);

    auto attr = new osg::ClipControl(osg::ClipControl::LOWER_LEFT, osg::ClipControl::ZERO_TO_ONE);
    rootNode->getOrCreateStateSet()->setAttributeAndModes(attr, osg::StateAttribute::ON);
    osg::ref_ptr<osg::Depth> depth = new osg::Depth(osg::Depth::GEQUAL, 1.0, 0.0);
    rootNode->getOrCreateStateSet()->setAttributeAndModes(depth, osg::StateAttribute::ON);

    mViewer->setSceneData(rootNode);
```

A custom depth buffer, attached to the main viewer camera, still seems to do nothing, while the depth buffer reversing breaks terrain rendering. Mentioned article also says that I need to update the projection matrix when I reverse the buffer, but I do not see related code in the OSG depth buffer example.
Can you tell what I do wrong? Do I need to create a framebuffer?

AnyOldName3:
It might be a better idea to ask SDL for the right depth format when the window is created than to try and switch the depth buffer after-the-fact.

akortunov:
> It might be a better idea to ask SDL for the right depth format when the window is created than to try and switch the depth buffer after-the-fact.

Do you know how to do it? The only thing I found is the `SDL_GL_SetAttribute(SDL_GL_DEPTH_SIZE,32);`, but it seems to be unrelated to our issue.

wareya:
I tried configuring the window's default depth buffer depth to 32-bit like so and it resulted in the same old 24-bit depth buffer I showed in the screenshots I posted a few weeks ago:

```
@@ -418,7 +418,7 @@ void OMW::Engine::createWindow(Settings::Manager& settings)
     traits->green = 8;
     traits->blue = 8;
     traits->alpha = 0; // set to 0 to stop ScreenCaptureHandler reading the alpha channel
-    traits->depth = 24;
+    traits->depth = 32;
     traits->stencil = 8;
     traits->vsync = vsync;
     traits->doubleBuffer = true;```

I don't know what other locations would need to be changed to make it actually use a 32-bit depth buffer.

**Comment by Capostrophic** [comment]
There's no implementation for traits->depth in our window so naturally there won't be any difference.  This line
```
checkSDLError(SDL_GL_SetAttribute(SDL_GL_DEPTH_SIZE, 24));
```
should actually be updated too.

wareya:
Still seems to be making the depth buffer for main scene rendering with GL_DEPTH_COMPONENT, giving me a 24 bit depth buffer with the characteristic distant z-fighting.

image
image
(ignore the exact event I stopped on, I just look around randomly until I find a state where the scene is mostly drawn but the depth buffer hasn't been cleared for viewmodel drawing yet)

AnyOldName3:
I've had a look, and it seems that while it should be possible to ask SDL to give us a 32-bit depth buffer, there's no way to specify that it needs to be floating point. Someone should report this to SDL as it seems like a major oversight. Otherwise, we'd need to either convince OSG/OpenGL to use a different depth buffer to the window one, which feels weird, or render to an FBO and copy to the main framebuffer at the end.

wareya:
Well right now it doesn't seem to be choosing a 32-bit depth buffer at all integer or not, at least on my system.

wareya:
Posted too quickly, sorry.

>or render to an FBO and copy to the main framebuffer at the end

This is probably a good idea in general. We should do it for the shaded path at least.

mp3butcher:
Just something i never tested but should work:
use stencil test and 8 cameras with 8 different splittted nearfar ranges and clear depth buffer after each camera pass...

AnyOldName3:
OSG has depth partitioning built-in as a feature without having to mess around setting it up ourselves. I *think* we even discussed setting it up in the past, but we decided against it because it's a lot of extra complexity (both code and execution time) for a small gain - doubling the number of partitions only helps as much as adding one extra depth bit, so if we were to have something work as well as a 32-bit depth buffer, even a boring integer one with nothing fancy added, we'd need 128 splits, which is ridiculous.

PR #2428, opened 2019-06-18
[Not for merging] add polygon offset on nif transparents

mp3butcher:
fix the lava z fight but could produce new artefacts...
...bad ressources:/


kcat:
That looks very hacky. I don't see the point in throwing in random offsets on random surfaces in the hopes it fixes some issues in another PR.

PR #2430, opened 2019-06-19
[Ready for review] use osg BoundingVolume Culling for near and far plane

mp3butcher:
It's literally a copy/paste from default osg ClampProjectionMatrixImplementation, with just a clamping of the near plane in a callback mean to be used for that....
(ref http://forum.openscenegraph.org/viewtopic.php?t=1332)
ClampCallback code might be simplifiied but it's a sensible code so let that for further tuning.

Near Plane Clamping  is the only thing preventing openmw from using BoudingVolume culling..
this approach have little cost and perhaps it might not be acceptable but benefits are all features brought by osgUtil::CullVisitor
- depth range narrowing (decreases zfight artefacts )
- correct frustum polytope, projectionmatrix, znear and zfar accessible in cull callbacks
- etc...


mp3butcher:
anything against this change?

Capostrophic:
Please read [policies and standards](https://wiki.openmw.org/index.php?title=Policies_and_Standards).

mp3butcher:
@Capostrophic  i use osg naming policy with osg objects
Doesn't it sound good to you to maintain naming convention across derivated class?

mp3butcher:
....now could i have pertinent critics

mp3butcher:
this pr interfere with shadow but if you want i can fix it (only few lines to remove in shadowtechnique)

AnyOldName3:
This looks like a reasonably large amount of code and it isn't where I expected it to be. I know there were things stopping OpenMW working with OSG's dynamic clipping planes, and I know I fixed lots of them as I needed them to mostly work to know where to put the shadow maps, but I don't know what else was broken. Could you elucidate, please?

mp3butcher:
it's literrally a copy/paste from default osg ClampProjectionMatrixImplementation, nothing to elaborate, just clamping near plane in a callback mean to be used for that

mp3butcher:
now nothing prevent openmw to use osg plane culling

mp3butcher:
@AnyOldName3
near plane is the only thing preventing openmw from using osg plane culling..
this approach have a the cost of a (dereferement +fx calling ) per drawables.
Perhaps it might not be acceptable but benefits are:

    AABB bounding box culling
    depth range computation
    frustum polytope maintained by osgUtil::CullVisitor
    and other culling features that doesn't come up right now...

AnyOldName3:
> this approach have a the cost of a (dereferement +fx calling ) per drawables.

Can you try rephrasing this? I'm not sure what you're saying.

mp3butcher:
i just say the additionnal cost of this approach is limited to calling the callback.

cost of call the callback =  pointer dereferencement + call to the function

all the rest is same behavior as before

mp3butcher:
@AnyOldName3  http://forum.openscenegraph.org/viewtopic.php?t=1332

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -362,14 +454,15 @@ namespace MWRender
             cullingMode |= osg::CullStack::SMALL_FEATURE_CULLING;
         }
 
-        mViewer->getCamera()->setCullingMode( cullingMode );
-
-        mViewer->getCamera()->setComputeNearFarMode(osg::Camera::DO_NOT_COMPUTE_NEAR_FAR);
```
AnyOldName3: Why can't we just remove this line and have OSG use the computed near and far planes? I don't get why we need a clamp projection matrix callback.

mp3butcher: near plane might be far compare to a hardcoded one...it allows to narrow depth range as it can be

AnyOldName3: You said

> it's literrally a copy/paste from default osg ClampProjectionMatrixImplementation, nothing to elaborate, just clamping near plane in a callback mean to be used for that

so I don't get why we're not just relying on that implementation.

mp3butcher: the ClampProjection matrix  callback is called at each leaf added to the stategraph.
it allows to give yourself constraints on the desired projection matrix
(like min near plane or max far plane)
more important, using bounding volume culling the cullvisitor use this clamped perspective matrix to maintain the frustum polytope...

mp3butcher: the near plane clamping is the only code change

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -192,7 +192,99 @@ namespace MWRender
...
+                    double zfarPushRatio = 1.02;
+                    double znearPullRatio = 0.98;
+        
+                    //znearPullRatio = 0.99;
+        
+                    double desired_znear = znear * znearPullRatio;
+                    double desired_zfar = zfar * zfarPushRatio;
+        
+                    // near plane clamping
+                    double min_near_plane = mZnear;
```
mp3butcher: this line is mine :-D

akortunov:
Does not seem to reduce the z-fighting at all. Are you sure this PR works as intended?

AnyOldName3:
If it works, this won't actually make all that much difference. OSG's near-far calculation is really designed for planet viewers, where you're not going to have anything nearby when looking at things a long way away, and aren't going to be able to see things a long way away when zoomed in. OpenMW always has things nearby unless you're flying.

It's not necessarily a bad thing to have a working version of this, but you have to half the far plane distance or double the near plane distance to get a single bit of extra depth precision, so it's not going to work miracles.

mp3butcher:
@akortunov 
it's not a matter of zfighting (but it can improve it in few cases)
It's a matter of using provided osg feature correctly
If we don't use bb for culling, a lot of culling stuff will not be used...For example, i strongly suspect that this is the root of all problems @AnyOldName3 had with the maintained frustum polytope in his  shadows impl

mp3butcher:
in fact thes change are not as useful as i though . I could merge it with https://github.com/OpenMW/openmw/pull/2433/commits/b9fd6bd657255d9f67b6f75ada8b914b29e92765 to clamp near to mNear but it would only fix z fight when view orientation is near vertical

psi29a:
Any additional work to be done here? Can BoudingVolume culling be implemented here? Do we have any figures when it comes to performance or drawbacks?

Feel free to dust it off though. Perhaps even create a merge request here:

PR #2432, opened 2019-06-21
Fix #4665

Assumeru:
From what I can tell the blank parts of the scrollbars are too big - they overlap the right/bottom button. This reorders the stack so those buttons are on top of the blank parts thereby ensuring they get click events instead.


Capostrophic:
Since this doesn't cause issues with the currently required and used 3.2.2 version but 3.2.2 version doesn't have any issues to begin with and this only fixes upstream MyGUI compatibility I'm going to merge this without requiring Assumeru to add a changelog entry.

PR #2439, opened 2019-07-01
Fix jumping mechanics framerate dependency (bug #4991)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/4991).

The way jump attempts work is, we cache the jumping state, feed physics system vertical movement every frame and see if it's going to be successful. If not (if the current framerate is above 60 or whatever the physics framerate is), we try it again in every character controller update. However, when we feed vertical movement to the physics system, we reduce fatigue and progress player's acrobatics skill each time. If we do in fact reset the jumping state, this is "fixed", but the player is unable to jump if the jumping movement is not going to be handled in this frame, so that's not a proper solution.

I work around this here by moving the fatigue decreasing and acrobatics progressing to the physics update, to the moment when the jumping state is in fact handled by the physics system and reset, so these operations are guaranteed to be only done once when a jump happens. Jumping state is now reset in the character controller if the vertical movement was cancelled by a certain conditions (sneaking, flying, knockdown state etc.) so these operations are not performed every time the player or an NPC attempts to jump. It's not a very satisfying solution due to how a part of the mechanics had to be moved to physics, but we already do that from time to time and it's better to have some working solution than a solution that isn't there at all. And there aren't any new dependencies other than the player class one necessary to set the jumping flag for GetPCJumping.

I hope I'm not going to hate myself for this.

### DIFF
--- a/apps/openmw/mwphysics/physicssystem.cpp
+++ b/apps/openmw/mwphysics/physicssystem.cpp
@@ -34,6 +34,7 @@
 
 #include "../mwworld/esmstore.hpp"
 #include "../mwworld/cellstore.hpp"
+#include "../mwworld/player.hpp"
 
 #include "../mwrender/bulletdebugdraw.hpp"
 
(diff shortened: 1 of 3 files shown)


wareya:
>It's not a very satisfying solution due to how a part of the mechanics had to be moved to physics, but we already do that from time to time and it's better to have some working solution than a solution that isn't there at all. 

This is just one of those situations where a physics event (the inertia change from jumping) and game mechanics (the effects on and dependencies on rpg values) are the same thing.

Capostrophic:
Well, if you put it that way...

apps/openmw/mwphysics/physicssystem.cpp
```diff
@@ -326,7 +327,30 @@ namespace MWPhysics
             if (movement.z() > 0 && ptr.getClass().getCreatureStats(ptr).isDead() && position.z() < swimlevel)
                 velocity = osg::Vec3f(0,0,1) * 25;
 
-            ptr.getClass().getMovementSettings(ptr).mPosition[2] = 0;
+            if (ptr.getClass().getMovementSettings(ptr).mPosition[2])
```
akortunov: Can there be a situation when the `mPosition[2]` is negative?

Capostrophic: Currently not, it's just a "vertical movement attempt" flag expressed in a number.

PR #2440, opened 2019-07-04
Fixed crash with sound disabled (bug #4600)

jlaumon:
[Bug #4600](https://gitlab.com/OpenMW/openmw/issues/4600)

It crashed inside openal code when either the sound is disabled with `--no-sound` or because there's no output device detected by the OS.

The reason was that `SoundManager::stopSound3D` forgot to check if `mOutput` was initialized before calling functions on it (the full callstack is in the bug report).

As it turned out, the call to stopSound3D that generated the crash wasn't even needed, so I removed it in the second commit (playSound3D already stops the sound).

### DIFF
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -983,7 +983,6 @@ void CharacterController::handleTextKey(const std::string &groupname, const std:
     if(evt.compare(0, 7, "sound: ") == 0)
     {
         MWBase::SoundManager *sndMgr = MWBase::Environment::get().getSoundManager();
-        sndMgr->stopSound3D(mPtr, evt.substr(7));
         sndMgr->playSound3D(mPtr, evt.substr(7), 1.0f, 1.0f);
         return;
     }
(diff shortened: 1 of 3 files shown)


Capostrophic:
Add a line into CHANGELOG.md.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -983,7 +983,6 @@ void CharacterController::handleTextKey(const std::string &groupname, const std:
     if(evt.compare(0, 7, "sound: ") == 0)
     {
         MWBase::SoundManager *sndMgr = MWBase::Environment::get().getSoundManager();
-        sndMgr->stopSound3D(mPtr, evt.substr(7));
```
akortunov: These stopSound3D calls were added for reason: e71aa53574f29860f7bf1e4baa254a44cde78add
Make sure that you do not regress [bug #2689](https://gitlab.com/OpenMW/openmw/issues/2689).

Capostrophic: playSound3D already stops the sound.

jlaumon: Yes it seems the stop was added a bit later (for a different bug) https://github.com/OpenMW/openmw/commit/1aaa8a76c5d9654669bf5c7ff0a408df5c3ef5a0

PR #2441, opened 2019-07-08
Particle System pointSprite, shader and lighting (fixing bug#3676)

mp3butcher:
This pr enables lighting of particle systems
and add a shaded path exploiting point_sprite opengl feature
it also fixes the alpha issue (https://gitlab.com/OpenMW/openmw/issues/3676) and https://gitlab.com/OpenMW/openmw/issues/4949

NB: 
- I would have let Ambient material on partsys but spec say alpha come from diffuse product.
https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glLightModel.xml

>             The alpha component of the resulting lighted color is set to the alpha value
>             of the material diffuse reflectance.

I only force particles normal to (0,0,-1) it appears to be quite like vanilla...

Capture d’écran de 2019-07-29 01-37-08
lighsmok


mp3butcher:
arg 3.4 problem

mp3butcher:
i can get rid of some shader parameters . very few are used for the moment in openmw so i can keep backward compatibility but changing texcoord of sprites could be usefull in the future to handle new effects...
Your call....

AnyOldName3:
We can't merge things that don't work with OSG 3.4.1 - our official releases still use our fork of it, and there are still issues that only occur with OSG 3.6.x. (Ideally, we should work out what they are in the next week or two, as Robert wants to tag the 3.6.4 release soon, and we don't know if the incompatibilities are because we're using OSG incorrectly, or just weirdly.)

mp3butcher:
should be compatible now

components/nifosg/nifloader.cpp
```diff
@@ -1028,6 +1031,13 @@ namespace NifOsg
                 trans->addChild(toAttach);
                 parentNode->addChild(trans);
             }
+
+            osg::StateSet *stateset = partsys->getOrCreateStateSet();
+            stateset->setMode(GL_LIGHTING,osg::StateAttribute::OFF);
+
+            osg::ref_ptr<osg::PointSprite> sprite = new osg::PointSprite;
```
mp3butcher: can't harm...even fixed pipeline should exploit this feature

mp3butcher:
I merged https://github.com/OpenMW/openmw/pull/2423 to fix particle behind transparents (ex:lanterns)

components/shader/shadervisitor.cpp
```diff
@@ -263,9 +265,79 @@ namespace Shader
...
+            {
+                state->getCurrentVertexArrayState()->assignNormalArrayDispatcher();
+                state->getCurrentVertexArrayState()->setNormalArray(*state, mNormalArray);
+            }else
+            {
+                state->getAttributeDispatchers().activateNormalArray(mNormalArray);
+            }
+#else
+            ///this doesn't work on 3.6
+            state->setNormalPointer(mNormalArray);
```
mp3butcher: osg 34 TEST NEEDED

components/nifosg/nifloader.cpp
```diff
@@ -1626,6 +1626,7 @@ namespace NifOsg
                                                                                    getBlendMode((alphaprop->flags>>5)&0xf)));
                         blendFunc = shareAttribute(blendFunc);
                         stateset->setAttributeAndModes(blendFunc, osg::StateAttribute::ON);
+                        stateset->setAttributeAndModes(new osg::Depth(osg::Depth::LESS, 0.0, 1.0, false));
```
kcat: As has been mentioned before, the nif files set the depth parameters themselves. Unless you can verify that vanilla does it, it's not up to the engine to override the depth settings the model says to use (including the defaults), even if they're not ideal.

mp3butcher: please I opened several pr in order to discuss of various mod
use this one https://github.com/OpenMW/openmw/pull/2423  to talk about blendfunc/deptmask

mp3butcher: and as i repeat to everyone: you MUST disable depthwrite for proper alpha blending...so i don't care what you say unless you have a proper answer to the problem of transparent occlusion

mp3butcher: I could remove this code by simply remove particle system from TRANSPARENT_BIN but again if we want correct blending of particles with other blended it's mandatory

kcat: > i don't care what you say unless you have a proper answer

The answer is, the engine doesn't second-guess what the game data tells it to do. Unless you can verify the vanilla engine does it too, it will be up to modders to create fixed models, not for the engine to alter models based on what it thinks it should do.

mp3butcher: you're wrong : the engine do want it want with game data
it's the principle of a blackbox
if someone have apitrace of a lantern in morrowing, i would really be surprised if they write depth for the glass as it prevent the drawing of the particlesystem

wareya: No. The engine does what the data says. You don't fix broken data on the engine side. Engines meant for general use are transparent by nature, not the black box you're talking about.

mp3butcher: I've made 3 engines so don't tell me how magic works

mp3butcher: and please discuss this problem in the dedicated pr #2423

mp3butcher: ...or on discord...

mp3butcher: alphafix

mp3butcher: I remove the merge to make u happy

mp3butcher: now lanterns  and foutain are broken (under certain view angle)
Capture d’écran de 2019-07-12 21-11-59
Capture d’écran de 2019-07-12 21-13-31

mp3butcher:
i removed #2423 
now lanterns  and foutain are broken (under certain view angle)
Capture d’écran de 2019-07-12 21-11-59
Capture d’écran de 2019-07-12 21-13-31

mp3butcher:
this should do the trick

components/nifosg/nifloader.cpp
AnyOldName3: Someone tell me a cell with a dodgy lantern and I'll have a look.

mp3butcher: https://github.com/OpenMW/openmw/pull/2441/commits/998900566e42bfae8e353e1856eeabaa122665f0 fix the problem

AnyOldName3:
In case it gets lost on Discord (conversation here https://discordapp.com/channels/260439894298460160/260439894298460160/599392927579635712), I've determined that particles are drawn in depth order with the other translucent drawables, so if you have two lanterns, you'll get flame, then glass, then flame, then glass. I don't know if this is done per-particle or per-particle-system. If it's the former, everything is terrible as we won't be able to combine whole systems into individual drawables any more.

mp3butcher:
particlesystem are sort coarsely with transparents then particles are sorted per particlesystem

mp3butcher:
merging this pr with https://github.com/OpenMW/openmw/pull/2447 works very well (i don't apply the cullcallback on particle system on purpose: this way there depth stays at their center)

AnyOldName3:
> particlesystem are sort coarsely with transparents then particles are sorted per particlesystem

How have you verified this? While we've discovered the original engine does a bunch of stupid things, I want to be sure it's definitely not doing other stupid things.

mp3butcher:
Yes I'm sure osg do that ..don't know about original engine

AnyOldName3:
Checking the original engine is what needs to happen, then.

mp3butcher:
@AnyOldName3 you've discussed with @Hrnchamd
apart the fact that vanilla sorts  on modelview coords ZthenYthenX and not on cartesian distance, the behavior seems pretty the same...so ... waiting your acknowledgement for others to review

mp3butcher:
problem: not compatible with scrawl version (no vertexarray path in particlesystem.cpp)

mp3butcher:
i mimic 3.6 behavior to fix lack of vertexArray support in osg scrawl...

files/shaders/objects_vertex.glsl
```diff
@@ -33,6 +33,12 @@ varying vec2 envMapUV;
 varying vec2 specularMapUV;
 #endif
 
+#if @pointsprite
+uniform float axisScale;
+uniform float visibilityDistance;
+varying vec3 basic_prop; // _alive, _current_size, _current_alpha
+#define SIZESCALE 1000
```
mp3butcher: If someone have a deterministic way of finding this scale factor i'm all heard

mp3butcher: dumb

mp3butcher:
I would have let Ambient material on partsys but there's strange lighting artefact with mat colormod =ambient (weirdier than transparent fading bug) like this one https://streamable.com/nw9kh
as a workaround i would choose to simulate ambient with diffuse: ambient approx= diffuse with normal pointing toward light foreach light...but as it can be done only via shader, i putted normal as pointing toward viewer:/(double workaround)

AnyOldName3:
If you want to disable pointsprite everywhere except the particle system, it might be better to do so in the global shader define map.

mp3butcher:
@AnyOldName3 u have a point...

mp3butcher:
arf dumb i am...can't handle two things at a time

mp3butcher:
@AnyOldName3 mmh didn't work...

AnyOldName3:
How so?

mp3butcher:
nothing messing with versions

components/shader/shadervisitor.cpp
```diff
@@ -283,8 +289,154 @@ namespace Shader
...
+            osg::State& state = *renderInfo.getState();
+            if(partsys->getParticleAlignment() == osgParticle::ParticleSystem::BILLBOARD)
+                mAxisScaling->set(osg::Matrix::transform3x3( state.getModelViewMatrix(), partsys->getAlignVectorX() * 0.5f * renderInfo.getCurrentCamera()->getViewport()->width()).length());
+            else
+                mAxisScaling->set(osg::Matrix::transform3x3( state.getModelViewMatrix(), partsys->getAlignVectorX() * 0.5f * renderInfo.getCurrentCamera()->getViewport()->width()).length2());
+
+            mAxisScaling->apply(state.get<osg::GLExtensions>(), state.getUniformLocation("axisScale"));
+#if OSG_MIN_VERSION_REQUIRED(3,6,0)
+            drawable->drawImplementation(renderInfo);
+#else //OSG Scrawl specific
```
mp3butcher, line 340: @akortunov shit begins here

mp3butcher:
now the "shit" can disappear. How magic!...sarcasm ad lib

Capostrophic:
That's not very nice.

mp3butcher:
@Capostrophic What's not nice is to always throw the rock on osg!!!

AnyOldName3:
Sometimes we think OSG might want their rocks back.

mp3butcher:
> Sometimes we think OSG might want their rocks back.

if you want, you can pr for it... no need to throw rock on anyone but you on this one, you have introduce bugs in your osg that wasn't present neither in 3.4 neither in 3.6

psi29a:
@mp3butcher take into account that you had to modify 26 files to fix a one-liner 'hack' that worked around an issue in OSG. https://github.com/OpenMW/osg/pull/14

Great, the original problem was fixed in OSG and the 'hack' can be removed. No need to be rude.
I think it is a good idea to take a step back to see that we are both attacking problems from different directions. OSG develops a library and has deep knowledge of OSG and has a good idea about the archtecture and how things 'should' work. While OpenMW (and oher applications) that use OSG are focused on how things 'really' work which isn't the same at all. So we're trying to understand why. It can be frustrating. Sometimes we really suspect a bug, and to be fair, it is. A refactor that changed behaviour or we are doing something that wasn't anticipated. That's just the nature of things.

We have a love/hate relationship with OSG and that's okay... **_we want OSG to become better_**. Otherwise we wouldn't be using it. So don't see it as throwing stones, rather we want to know why things don't work like we expect and it's good to get push-back. Right now that is double difficult because Robert is no longer focused on OSG.

mp3butcher:
If you want OSG to be better, I think the less you can do is making your OSGOS divergence working....There's no way ffp particle lighting to work in these condition
** ffp particle lighting works with 3.4, 3.6 but not with OSGOS**
Damned!

mp3butcher:
mmh excuse my anger.... OSGOS seems to have accepted my normal....
it's really strange it works now as i tried really hard in the past without achieving good result....
need testing

mp3butcher:
still a strange problem with antialiasing: particles get tinier with number of AA passes

AnyOldName3:
Is the point size pixels in the non-resolved buffer rather than the resolved one?

mp3butcher:
> Is the point size pixels in the non-resolved buffer rather than the resolved one?

I add code to take into account the only case posing problem(antialiasing = 16), but not proud of the implementation.
I though  gl_PointSize *= log2(antialiasingsetting) would do the trick but it's far more complex aa= [0..8] require no scaling but aa = 16 require scalingx2 !?
@AnyOldName3 if you have a lead on that I'm all heard

mp3butcher:
I capitulated with point_sprite *snif*

PR #2442, opened 2019-07-08
fix particle system alpha

mp3butcher:
it fixes the alpha issue https://gitlab.com/OpenMW/openmw/issues/3676

Capture d’écran de 2019-07-08 19-31-10

### DIFF
--- a/components/nifosg/particle.cpp
+++ b/components/nifosg/particle.cpp
@@ -152,7 +152,6 @@ void ParticleColorAffector::operate(osgParticle::Particle* particle, double /* d
     osg::Vec4f color = mData.interpKey(time);
 
     particle->setColorRange(osgParticle::rangev4(color, color));
-    particle->setAlphaRange(osgParticle::rangef(color.w(), color.w()));
 }
 
 GravityAffector::GravityAffector(const Nif::NiGravity *gravity)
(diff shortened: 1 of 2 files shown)


components/nifosg/particle.cpp
```diff
@@ -152,6 +152,7 @@ void ParticleColorAffector::operate(osgParticle::Particle* particle, double /* d
     osg::Vec4f color = mData.interpKey(time);
 
     particle->setColorRange(osgParticle::rangev4(color, color));
+    particle->setAlphaRange(osgParticle::rangef(color.w(), color.w()));
```
Capostrophic: This will just apply the transparency twice from my understanding of osgParticle. The fourth value of the color *is* used as transparency which is then multiplied by alpha range alpha.

mp3butcher: yes you're right

akortunov:
Does not work for me.

mp3butcher:
ok my reasonning is the following:
as Alpha and Color may have different functions the commit i removed is pertinent.
so there's an incoherence in osgParticle::PArticleSystem::drawimplementation
I 'll force push in order to take that into account

mp3butcher:
i've made that cleaner: https://github.com/OpenMW/openmw/pull/2444

PR #2444, opened 2019-07-08
fix particle system alpha and lighting for fixed pipeline

mp3butcher:
same as https://github.com/OpenMW/openmw/pull/2441 but only for FFP (easier to review)
it fixes the alpha issue https://gitlab.com/OpenMW/openmw/issues/3676

Capture d’écran de 2019-07-11 03-42-04


akortunov:
Just for reference: it works only in the fixed pipeline mode.

Capostrophic:
Particle lighting doesn't behave right in shader pipeline, there's a bug filed for that. This seems to be worth looking into since there's no bug in osgParticle itself.

mp3butcher:
@akortunov 
...that's the reason of the 2 differents prs...Making a simplier one (fixing only fixed pipeline) intend not to be overhelming for review

AnyOldName3:
Can you explain why the changes here help?

mp3butcher:
@AnyOldName3 
commit1 : disable lighting because material cause strange behavior with gl_Color alpha
commit2: separate function over time for Alpha and Color

AnyOldName3:
Can you be clear and specific about that, please? I've done some investigation into OpenGL material colours and how they interact with vertex colours in the past, so I might be able to find a better way of sorting that out without disabling lighting if you can tell me exactly what you've found, and I don't know how the colour and alpha are used in `osgParticle` exactly, so you might be able to enlighten me much faster than I could look through the source code.

mp3butcher:
Material seems to affect gl_Color preventing alpha control of particlesystem...and disabling material is not enough to prevent that, so i disabled light. (not a problem particlesystems have its own renderbin)
Perhaps it's linked to some normal bore effect as particles don't have normals (and so can't be used for lighting)
Don't know

AnyOldName3:
You know about `glColorMaterial` and what it does, right?

mp3butcher:
please... I know how to disable Material...

AnyOldName3:
Just checking - not everyone still knows how the FFP works.

mp3butcher:
i found that scrawl commit https://github.com/OpenMW/openmw/commit/0a411cbd7cd55cacb5c8c2c1efcbf5dddf8f862e#diff-d913374ea75d1e583d8c7a5d028e4f4f
but it seems this code was removed 
it can explain shy i have to disable lighting

mp3butcher:
seems the proper fix in fact

Capostrophic:
Pretty sure just removing that code causes severe rendering glitches with particles.

mp3butcher:
@Capostrophic 
give argument against this change not your assurance...you're never sure....

mp3butcher:
reopened just for a win32 test

components/nifosg/particle.cpp
```diff
@@ -191,6 +191,7 @@ void ParticleColorAffector::operate(osgParticle::Particle* particle, double /* d
 
     particle->setColorRange(osgParticle::rangev4(color, color));
     particle->setAlphaRange(osgParticle::rangef(alpha, alpha));
+
```
akortunov, line 194: Please discard this redundant change and squash commits.

mp3butcher: > 
> 
> reopened just for a win32 test
I'll clean that pr as test is pretty satisfying..don't know why i closed it

mp3butcher: ah yes i remember, someone nags me about stuff that wasn't in nif

PR #2445, opened 2019-07-09
move rig and morph update in update thread

mp3butcher:
cleaner for stats

### DIFF
--- a/components/sceneutil/morphgeometry.cpp
+++ b/components/sceneutil/morphgeometry.cpp
@@ -157,7 +157,7 @@ void MorphGeometry::update(osg::NodeVisitor *nv)
     if (mLastFrameNumber == nv->getTraversalNumber() || !mDirty)
     {
         osg::Geometry& geom = *getGeometry(mLastFrameNumber);
-        mLastFrameMutex.unlock();
+
         nv->pushOntoNodePath(&geom);
         nv->apply(geom);
         nv->popFromNodePath();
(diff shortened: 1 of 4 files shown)


akortunov:
I do not understand what do you try to achieve. From what I can tell, OpenMW performs both Update and Cull in the main thread (you can see it in the profiler), so moving rigs processing from Cull to Update does not improve situation at all. Also a separate mutex for EVERY rig in scene seems to be an overkill for me (mutex is a quite slow thing).

As about FPS: this PR drops FPS from 95 to 55 for me in testing scene with 10 NPC's. This drop is caused by DYNAMIC data variance in the MorphGeometry (the draw thread basically becomes idle when the main thread is busy).

So in its current state this PR is basically unusable.

mp3butcher:
sorry was i bit tired doing this pr...i fixed errors
the goal of this pr is to do update in update, it is cleaner for stats

AnyOldName3:
We intentionally do this in the cull thread because it means that a fairly time-hungry operation only happens for things that pass culling. This is a very common thing to do in game engines - Unreal doesn't do any skinning without the approval of an occlusion query against the previous frame, for example. We're not going to sacrifice performance just to make the profiler look tidier.

It also means that if we ever implement hardware skinning, it's a simpler task as things aren't shunted around as much.

mp3butcher:
as you like

PR #2446, opened 2019-07-10
remove hackish particle setup

mp3butcher:
don't know if i'm right but it doesn't seems to break anything
tell me if you know the problem adressed by previous code or find  broken stuff

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -942,17 +942,11 @@ namespace NifOsg
                 return;
             }
 
-            osgParticle::ParticleProcessor::ReferenceFrame rf = (animflags & Nif::NiNode::ParticleFlag_LocalSpace)
-                    ? osgParticle::ParticleProcessor::RELATIVE_RF
-                    : osgParticle::ParticleProcessor::ABSOLUTE_RF;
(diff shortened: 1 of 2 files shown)


components/nifosg/nifloader.cpp
```diff
@@ -942,17 +942,11 @@ namespace NifOsg
...
-                    ? osgParticle::ParticleProcessor::RELATIVE_RF
-                    : osgParticle::ParticleProcessor::ABSOLUTE_RF;
+            osgParticle::ParticleSystem::ParticleScaleReferenceFrame rf = (animflags & Nif::NiNode::ParticleFlag_LocalSpace)
+                    ? osgParticle::ParticleSystem::LOCAL_COORDINATES
+                    : osgParticle::ParticleSystem::WORLD_COORDINATES;
 
-            // HACK: ParticleSystem has no setReferenceFrame method
-            if (rf == osgParticle::ParticleProcessor::ABSOLUTE_RF)
-            {
-                partsys->getOrCreateUserDataContainer()->addDescription("worldspace");
```
akortunov: Keep in mind that this decription is used in the SceneManager.

mp3butcher: ...yes but it doesn't seems to change anything

PR #2447, opened 2019-07-12
[To merge] add cullcallback on depth sorted to better compute leafdepth

mp3butcher:
try to fix bug https://gitlab.com/OpenMW/openmw/issues/3366 and other frustrating transparent artefacts
but make proof sort order have to be reverted
https://streamable.com/by12v

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -40,6 +40,7 @@
 #include <components/sceneutil/skeleton.hpp>
 #include <components/sceneutil/riggeometry.hpp>
 #include <components/sceneutil/morphgeometry.hpp>
+#include <components/sceneutil/visitor.hpp>
 
 #include "particle.hpp"
 #include "userdata.hpp"
@@ -1030,6 +1031,7 @@ namespace NifOsg
             }
         }
 
+
         void triShapeToGeometry(const Nif::NiTriShape *triShape, osg::Geometry *geometry, osg::Node* parentNode, SceneUtil::CompositeStateSetUpdater* composite, const std::vector<int>& boundTextures, int animflags)
         {
             const Nif::NiTriShapeData* data = triShape->data.getPtr();
@@ -1068,7 +1070,11 @@ namespace NifOsg
             //   above the actual renderable would be tedious.
             std::vector<const Nif::Property*> drawableProps;
             collectDrawableProperties(triShape, drawableProps);
-            applyDrawableProperties(parentNode, drawableProps, composite, !data->colors.empty(), animflags, false);
+            if(applyDrawableProperties(parentNode, drawableProps, composite, !data->colors.empty(), animflags, false))
+            {
+                SceneUtil::AddRemoveTransparentCullCallback transvis(true);
+                geometry->accept(transvis);
+            }
         }
 
         void handleTriShape(const Nif::NiTriShape* triShape, osg::Group* parentNode, SceneUtil::CompositeStateSetUpdater* composite, const std::vector<int>& boundTextures, int animflags)
@@ -1559,11 +1565,11 @@ namespace NifOsg
                 found = sCache.insert(attr).first;
             return *found;
         }
-
-        void applyDrawableProperties(osg::Node* node, const std::vector<const Nif::Property*>& properties, SceneUtil::CompositeStateSetUpdater* composite,
+        bool applyDrawableProperties(osg::Node* node, const std::vector<const Nif::Property*>& properties, SceneUtil::CompositeStateSetUpdater* composite,
                                              bool hasVertexColors, int animflags, bool particleMaterial)
         {
             osg::StateSet* stateset = node->getOrCreateStateSet();
+            bool isTransparent = false;
 
             int specFlags = 0; // Specular is disabled by default, even if there's a specular color in the NiMaterialProperty
             osg::ref_ptr<osg::Material> mat (new osg::Material);
@@ -1639,8 +1645,10 @@ namespace NifOsg
                         stateset->setAttributeAndModes(blendFunc, osg::StateAttribute::ON);
 
                         bool noSort = (alphaprop->flags>>13)&1;
-                        if (!noSort)
+                        if (!noSort){
                             stateset->setRenderingHint(osg::StateSet::TRANSPARENT_BIN);
+                            isTransparent = true;
+                        }
                         else
                             stateset->setRenderBinToInherit();
                     }
@@ -1715,12 +1723,13 @@ namespace NifOsg
                     && mat->getSpecular(osg::Material::FRONT_AND_BACK) == osg::Vec4f(0.f, 0.f, 0.f, 0.f))
             {
                 // default state, skip
-                return;
+                return isTransparent;
             }
 
             mat = shareAttribute(mat);
 
             stateset->setAttributeAndModes(mat, osg::StateAttribute::ON);
+            return isTransparent;
         }
 
     };
--- a/apps/openmw/mwrender/animation.cpp
+++ b/apps/openmw/mwrender/animation.cpp
@@ -1859,9 +1859,16 @@ namespace MWRender
             osg::StateSet* stateset = mObjectRoot->getOrCreateStateSet();
             stateset->setRenderingHint(osg::StateSet::TRANSPARENT_BIN);
             stateset->setRenderBinMode(osg::StateSet::OVERRIDE_RENDERBIN_DETAILS);
+
+            SceneUtil::AddRemoveTransparentCullCallback transvis(true);
+            mObjectRoot->accept(transvis);
         }
         else if (osg::StateSet* stateset = mObjectRoot->getStateSet())
+        {
             stateset->setRenderBinToInherit();
+            SceneUtil::AddRemoveTransparentCullCallback transvis(false);
+            mObjectRoot->accept(transvis);
+        }
     }
 
     void Animation::setLightEffect(float effect)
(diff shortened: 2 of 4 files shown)


components/nifosg/nifloader.cpp
```diff
@@ -1030,6 +1030,78 @@ namespace NifOsg
...
+                            cv->pushStateSet(itr->first.get());
+                        }
+                    }
+                }
+
+                float depth = FLT_MAX;
+                if(bb.valid())
+                {
+                   for(unsigned int i=0; i<8; ++i)
+                       depth = std::min((float)distance(bb.corner(i), matrix), depth);
```
AnyOldName3: I'm pretty sure I saw a way somewhere in the cull visitor code of finding the nearest BB corner without having to loop through all of them. This isn't the biggest optimisation, but if there's tried and tested OSG code that does the same thing but faster, we might as well do the same thing.

components/nifosg/nifloader.cpp
```diff
@@ -1030,6 +1030,78 @@ namespace NifOsg
             }
         }
 
+        class PreciseLeafDepthCullCallback : public osg::DrawableCullCallback{
```
AnyOldName3: It's probably a good idea to declare this in a header as there are other places where things can become translucent. For example, when the chameleon effect is applied to an actor, they become translucent, so all of their drawables will need this callback adding.

AnyOldName3:
My main concern with this (that hasn't been mentioned in one of the inline comments) is how it might interfere with any other cull callbacks on drawables we might set elsewhere - we might not use any (so that should be checked) but if we do, we need to make sure that everything still works.

components/nifosg/nifloader.cpp
AnyOldName3: `osg::DrawableCullCallback` is `typedef`ed to `osg::Drawable::CullCallback`, and `osg::Drawable::CullCallback` existed in older OSG versions, so switching to that should help with the compilation issues with 3.4.1.

AnyOldName3: I see you fixed this while I was typing that out.

components/nifosg/nifloader.cpp
```diff
@@ -1030,6 +1030,78 @@ namespace NifOsg
...
+            inline osgUtil::CullVisitor::value_type distance(const osg::Vec3& coord,const osg::Matrix& matrix) const
+            {
+                return -((osgUtil::CullVisitor::value_type)coord[0]*(osgUtil::CullVisitor::value_type)matrix(0,2)+(osgUtil::CullVisitor::value_type)coord[1]*(osgUtil::CullVisitor::value_type)matrix(1,2)+(osgUtil::CullVisitor::value_type)coord[2]*(osgUtil::CullVisitor::value_type)matrix(2,2)+matrix(3,2));
+            }
+
+            virtual bool cull(osg::NodeVisitor* nv, osg::Drawable* drawable, osg::RenderInfo* renderInfo) const{
+                osgUtil::CullVisitor * cv = nv -> asCullVisitor();
```
AnyOldName3: `osg::NodeVisitor::asCullVisitor` didn't exist until OSG 3.5.2, so this won't work on 3.4.1. Your options are:

* A `dynamic_cast` - slowness that isn't necessary.
* Checking the type of visitor with `osg::NodeVisitor::getVisitorType() == osg::NodeVisitor::CULL_VISITOR` and then a `static_cast` if it's true - less slowness, but unnecessary if you know it's always going to be a cull visitor.
* Just doing a `static_cast` - if you can guarantee that cull callbacks only get called by cull visitors (which seems sensible) a static cast is fine. It'll also be faster than the virtual function call you've already got here.
* Checking the OSG version with a macro and using a different function depending on that - probably unnecessary and overcomplicated, but might avoid using a slow path on 3.6.x just becasue 3.4.x didn't have the fast path yet.

mp3butcher: static_cast is cool

components/nifosg/nifloader.cpp
mp3butcher: could be approx with distance(bb.center,matrix)-BoundingSphere(bb).radius but what's heck...there's not to much transparents

mp3butcher: ...i found the code...

mp3butcher:
mmmh don't know if it's valid it's from the billboard case in culvisitor

components/nifosg/nifloader.cpp
mp3butcher: ..it was for BILLBOARD..:(

mp3butcher:
> My main concern with this (that hasn't been mentioned in one of the inline comments) is how it might interfere with any other cull callbacks on drawables we might set elsewhere - we might not use any (so that should be checked) but if we do, we need to make sure that everything still works.

I added multiple callbacks support to prevent that...

mp3butcher:
sorry it take me the entire day to convince me but sorting order NEED to be reversed for correct result!!!!
again an excentrity from bowler plate manners....alpha might have been pre-inverted or something....

mp3butcher:
> sorry it take me the entire day to convince me (#2449 sorting order NEED to be reversed for correct result!!!!
> again an excentrity from bowler plate manners....alpha might have been pre-inverted or something...
> or i'm just asking to much to a coarse sorting

I'm wrong this is the best that can be done..:/

AnyOldName3:
I think the main thing we've learned from this is that Morrowind does things weirdly, and making assumptions about how transparency works isn't safe, so we need more testing to determine exactly how depth sorting is done in the vanilla engine before meddling with it. We need a test case where the rendering order would be different if the sort was based on the nearest or furthest point or the centre of a bounding box or sphere or the object's origin.

mp3butcher:
the lava and the smoke in chamber of akulakahn and the fountain of old ebonhaert are a good example arguing bak2front sorting require farrest point as depth
capture are done with depth as nearest
Capture d’écran de 2019-07-14 23-41-43
Capture d’écran de 2019-07-14 23-42-18

AnyOldName3:
I've consulted with someone who actually knows how copyright law works, and because you read about this in a Gamebryo header, someone else has to implement it instead of you. I'm going to have to close this, but if you explain the intended behaviour in English (or maybe pseudocode) but not real C++ on the issue tracker, someone should get around to doing it legally.

mp3butcher:
This behavior is not gamebryo behavior but just the boiler plate traditional back2front transparent composition!!!!!!!!!!!
Please reopen this pr!

AnyOldName3:
@psi29a Please mediate

mp3butcher:
@psi29a @AnyOldName3  Yes mediate this, you're attacking my public reputation with this kind of allegation..Delete this immediatly

AnyOldName3:
Dude, you looked at the Gamebryo header. You've said you looked at the Gamebryo header. We're not saying you *illegally* looked at the header (as whoever uploaded it might have had the legal right to do so), but people who've looked at a Gamebryo header can't submit related pull requests to OpenMW unless Gamebase Co., Ltd. releases the Gamebryo engine with a GPL-compatible licence.

mp3butcher:
as we've speak on discord this has nothing to do with gamebryo engine as it's the exact opposite behavior, so i demand you removed these out of context allegation

mp3butcher:
as we've speak on discord this has nothing to do with gamebryo engine as it's the opposite behavior:
what i oversee is that Gambryo engine seemed to sort front2back base on nearestpoint , here it's the exact opposite behavior sortback2front based on farrest point.
So i demand you removed these out of context allegations

AnyOldName3:
What do you think we're alleging that you did? The only things we've said you've done are look at a header file and create this PR.

Do you dispute that you looked at a Gamebryo header? That would be a silly thing to dispute, as you already said you did.

Do you dispute that you created this PR? That would be a silly thing to dispute, as you're clearly using the same account as created the PR.

If you agree that both of those things happened, then there are no untruthful allegations, as those two things are literally all we've said you've done.



The problem here is that the way copyright law works, anyone who's looked at any Gamebryo source code (including headers) that concerns alpha sorting *can't contribute code relating to alpha sorting to OpenMW*. It doesn't matter that your code doesn't do things in the same way. It doesn't matter that none of the code here came from Gamebryo. It doesn't matter how you found the Gamebryo header. You've not necessarily done anything illegal, but, as a GPL project, we can't accept this pull request - if we did, then *we'd* have done something illegal.

mp3butcher:
I looked at header an see with function name that misslead me and yield lot of non sens from me on discord
but it has nothing to do with this pr

mp3butcher:
further i'm more concern of the future of this pr than what you think i did or didn't read

AnyOldName3:
Hrnchamd has just confirmed what the original engine does anyway, and it's not this, so the discussion is moot anyway.

> it has nothing to do with this pr

This PR concerns alpha sorting. The header concerns alpha sorting. The details are all completely different, so from a people-who-know-what-they're-talking-about perspective, they don't have anything to do with each other. However, from a *legal* perspective, they're similar. As I keep saying, copyright law isn't sensible, as sensible laws wouldn't call this PR copying, but the laws we have are the laws that must be followed.

psi29a:
Don't forget, thanks to Oracle, headers and API are copyrighted and not fair-use.

https://www.itworld.com/article/2738675/the-impact-of-oracle-s-defense-of-api-copyrights.html

So in the United States, this is just plain copyright infringement. I'm still holding my breath waiting for this to shot back to hell. I personally want OpenMW to as free from liability as possible, this also includes doing 'our' best effort to make sure we are complying to applicable laws around the world. Open, transparent and if necessary, having to say no to otherwise good PRs.

PR #2449, opened 2019-07-13
[not for merging] Revert sort of Transparent

mp3butcher:
fix bug https://gitlab.com/OpenMW/openmw/issues/3366 and other frustrating transparent artefacts
https://streamable.com/79d00
but break a lot of stuff so coc punabi and check the krystals
If you want to be convinced compare this pr with https://github.com/OpenMW/openmw/pull/2447
only last commit is different
 it takes me the entire day to convince me sorting order NEED to be reversed for correct result!!!!
again an excentrity from bowler plate manners....alpha might have been pre-inverted or something....
but it can be simply me that ask to much and https://github.com/OpenMW/openmw/pull/2447  the best we could hope for


mp3butcher:
but i could be completely wrong

AnyOldName3:
As discussed on Discord, this is absolutely not what the original engine does.

PR #2450, opened 2019-07-14
[ready for review] Hierarchize transforms at cell level

mp3butcher:
following https://github.com/OpenMW/openmw/pull/2421 (for reference)
push hierarchic transforms at cell levels (require testing, perhaps i missed special transform ajustement)


mp3butcher:
arg the last commit broke stuff

apps/openmw/mwrender/objects.cpp
```diff
@@ -197,9 +213,14 @@ void Objects::updatePtr(const MWWorld::Ptr &old, const MWWorld::Ptr &cur)
...
         }
 
-    if (objectNode->getNumParents())
+    if (objectNode->getNumParents()>0)
+    {
         objectNode->getParent(0)->removeChild(objectNode);
-    cellnode->addChild(objectNode);
+    }
+
+    const float *f = cur.getRefData().getPosition().pos;
```
Capostrophic: Keep in mind getPosition().asVec3() is a thing, and it's not just here you can use it.

apps/openmw/mwrender/objects.hpp
```diff
@@ -6,7 +6,8 @@
 
 #include <osg/ref_ptr>
 #include <osg/Object>
-
+#include <osg/Vec3>
```
Capostrophic: This specific header doesn't use vectors in its declarations. Put this include into files that actually use vectors to keep compilation times shorter.

components/sceneutil/util.cpp
```diff
@@ -1,10 +1,20 @@
 #include "util.hpp"
-
 #include <osg/Node>
+#include <components/esm/loadcell.hpp>
+#include <components/esm/loadland.hpp>
 
 namespace SceneUtil
 {
 
+const float cellSize = static_cast<float>(ESM::Land::REAL_SIZE);
```
Capostrophic: Global variables are a no-no.

mp3butcher: any reason for it?

apps/openmw/mwrender/objects.cpp
mp3butcher: i just used the original syntax...but if you think it's better, i have nothing against it

apps/openmw/mwrender/objects.hpp
mp3butcher: which file do you think about?

Capostrophic: Um. The source files that fail to compile if the line is removed.

mp3butcher:
comply with @Capostophic review

apps/openmw/mwrender/renderingmanager.cpp
```diff
@@ -1152,7 +1153,8 @@ namespace MWRender
     {
         mPlayerAnimation = new NpcAnimation(player, player.getRefData().getBaseNode(), mResourceSystem, 0, NpcAnimation::VM_Normal,
                                                 mFirstPersonFieldOfView);
-
+        if(mPlayerNode->getNumParents()==0)
+            mObjects->getOrCreateCell(player.getCell())->addChild(mPlayerNode);
```
Capostrophic: This line causes a crash (null pointer dereferencing) for me. Player's cell can and will be null before the game is fully loaded. How did it get past your own testing?

mp3butcher: newgame was a case i haven't checked..I added a fix...it seems to do the trick

akortunov:
So which gains this PR is supposed to give?

mp3butcher:
> So which gains this PR is supposed to give?
What that must be done must be done....

No hierarchy -> huge transforms on each object leads to numeric problem if you want to flatten or merge them

psi29a:
Close this for now, unless more work is inbound. Obviously as time goes on, we have merge conflicts.

It would be nice if there was an explanation about what this change attempts to do and how it would benefit OpenMW.

Feel free to reopen here: https://gitlab.com/OpenMW/openmw

AnyOldName3:
The main motivation for me suggesting this was that I thought it would be a necessary prerequisite for what was then known as dynamic batching, i.e. what we now call object paging. It shouldn't be necessary any more now object paging is implemented.

psi29a:
It touches a lot of things around light, I'm wondering if was also trying to solve that issue too?

Capostrophic:
If you mean the light limit issue then I'm pretty sure that wasn't the point of changes in lightutil. They make sure a light source's position is adjusted properly taking the cell origin into account since light sources aren't handled like most objects.

PR #2452, opened 2019-07-22
Improve swish sounds handling (bug #5057)

akortunov:
A used formula:
```
volume = 0.98 + attackStrength * 0.02;
pitch = 0.75 + attackStrength * 0.4;
```

It is used for common attacks, werewolves and "miss" sounds.
Notes:
1. For non-bipedal creatures pitch is constant. I can not tell which value exactly is used in Morrowind, so use 0 for now, as we do in out upstream code.
2. I can not tell if "hit" sounds should use pitch which depends on attack strength or not, so I did not touch them.

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -377,7 +377,7 @@ namespace MWClass
         {
             // Missed
             if (!attacker.isEmpty() && attacker == MWMechanics::getPlayer())
-                MWBase::Environment::get().getSoundManager()->playSound3D(ptr, "miss", 1.0f, 1.0f);
+                MWBase::Environment::get().getSoundManager()->playSound3D(ptr, "miss", 0.98f, 0.75f);
             return;
         }
 
--- a/apps/openmw/mwmechanics/character.cpp
+++ b/apps/openmw/mwmechanics/character.cpp
@@ -1227,8 +1227,9 @@ bool CharacterController::updateCreatureState()
 
                 mAttackStrength = std::min(1.f, 0.1f + Misc::Rng::rollClosedProbability());
 
+                // Note: for non-bipedal creatures attack strength seems to be 0 during pitch calculation
                 if (weapType == WeapType_HandToHand)
-                    playSwishSound(0.0f);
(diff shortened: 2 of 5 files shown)


apps/openmw/mwclass/creature.cpp
```diff
@@ -377,7 +377,7 @@ namespace MWClass
         {
             // Missed
             if (!attacker.isEmpty() && attacker == MWMechanics::getPlayer())
-                MWBase::Environment::get().getSoundManager()->playSound3D(ptr, "miss", 1.0f, 1.0f);
+                MWBase::Environment::get().getSoundManager()->playSound3D(ptr, "miss", 0.98f, 0.75f);
```
Capostrophic: This isn't how it works. WeaponSwish sound gets lower pitch and volume due to a miss while miss sound doesn't get changed. WeaponSwish sound still plays if you miss.

apps/openmw/mwmechanics/character.cpp
```diff
@@ -1227,8 +1227,9 @@ bool CharacterController::updateCreatureState()
 
                 mAttackStrength = std::min(1.f, 0.1f + Misc::Rng::rollClosedProbability());
 
+                // Note: for non-bipedal creatures attack strength seems to be 0 during pitch calculation
```
Capostrophic: It's not 0, it's the same as attack strength (and is 0 if the creature misses and attack strength is 0), but Hrnchamd said swish sound is played on swing release in any instance, which is curious.

akortunov: Unfortunately, we do not know if attack will hit or miss when we start attack animation and start to play the swish sound.

Capostrophic:
This *does* improve swish sound playback but it doesn't fix the original issue. For which the following is necessary:
miss/no hit evaluation must be performed before the swish sound is played, and in the first case attack strength must be reset to 0.

Miss sound should be unrelated and it only plays for the player anyways.

akortunov:
Requires an additional research and refactoring then - we calculate the hit/miss in the NPC/Creature::onHit(), and if we are supposed to play swish sounds here, there will be no sounds when player hits the air at all.

PR #2453, opened 2019-07-25
Store raw data about fog of war instead of RGBA images

akortunov:
Fixes [bug #5108](https://gitlab.com/OpenMW/openmw/issues/5108) by implementing a TODO - using a raw data about "fog of war" textures instead of 32-bit RGBA image.

As a result, fog textures now occupy 4x lesser space in savegame.
With this PR my endgame savegame has 15% lesser size (26.3MB instead of 31MB) and loading takes about 3% less time (I use SSD, HDD may have a different speedup).

### DIFF
--- a/apps/openmw/mwrender/localmap.cpp
+++ b/apps/openmw/mwrender/localmap.cpp
@@ -9,8 +9,6 @@
 #include <osg/LightSource>
 #include <osg/PolygonMode>
 
-#include <osgDB/ReadFile>
-
 #include <components/debug/debuglog.hpp>
 #include <components/esm/fogstate.hpp>
 #include <components/esm/loadcell.hpp>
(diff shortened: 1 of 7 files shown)


apps/openmw/mwrender/localmap.cpp
```diff
@@ -3,22 +3,20 @@
 #include <stdint.h>
 
 #include <osg/Fog>
+#include <osg/ImageUtils>
```
Capostrophic: Is this used?

akortunov:
Are you sure that it is a good idea to merge it right now? This PR changes a savegame format.

Capostrophic:
Please revert!

psi29a:
I noticed that.  Will previous games work?

akortunov:
Just as after any savegame format update.

psi29a:
Apologies, I didn't realize we were going to save all save-game related issues for 0.47; reverted 

We could probably make use of [waiting for 0.47] in the title or somewhere in the text.

psi29a:
That makes it clear it is a save-game format change, one I was comfortable with. It was not indicated that it was on-hold for 0.47, I had to read about that else where, after it was brought to my attention.

https://forum.openmw.org/viewtopic.php?f=42&t=6373

 [on hold] and perhaps a short reason why (like posting link above) isn't unreasonable.

PR #2455, opened 2019-07-30
Revert questionable input manager decisions

Capostrophic:
These are some things people have been complaining about and some things I personally dislike about input changes introduced during gamepad behavior improvements. They weren't given much of a thought before implementation.

* Remove frankly unuseful Options Menu shortcut (I'm not sure if people are even aware of its existence).
* Revert to more pleasant 0.45.0 Escape behavior -- currently it purely toggles the main menu, like vanilla, but people complained about it not working as it did previously -- and it closed the current GUI mode. Any subsequent improvements must first be discussed as feature requests.
* Re-enable keyboard GUI arrow conversion feature -- while gamepads got DPAD arrows working for that, keyboards only got arrows for that again, but people liked that WASD or whatever set up as movement controls worked like GUI arrows. I don't have a gamepad so I can't test any possible improvements of gamepad behavior.
* Remove gamepad button release handling -- currently just dead code untouched for months triggering my non-existent OCD.

May need a review from gamepad people, because from as far as I can tell B button works kind of similarly to Escape now, and I'm unsure if test that should block DPAD from triggering button release twice works as intended. Also, maybe someone liked options menu being available directly on gamepads and I must put that back immediately.

### DIFF
--- a/apps/openmw/mwinput/inputmanagerimp.cpp
+++ b/apps/openmw/mwinput/inputmanagerimp.cpp
@@ -197,6 +197,11 @@ namespace MWInput
 
     void InputManager::handleGuiArrowKey(int action)
     {
+        // This is currently keyboard-specific code
+        // TODO: see if GUI controls can be refactored into a single function
+        if (mJoystickLastUsed)
+            return;
+
         if (SDL_IsTextInputActive())
             return;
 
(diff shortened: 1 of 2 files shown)


Capostrophic:
So far one gamepad user stated this works fine.

Capostrophic:
Taking two reports as a confirmation this works as intended, merging since these are fixes for things that have lots of complaints, or basically regression fixes.

PR #2456, opened 2019-07-30
Make sure getSpellSuccessChance() checks mana when necessary (bug #5112)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5112)

Make sure all getSpellSuccessChance calls work correctly -- return 0 to setSelectedSpell calls so that HUD shows an empty bar and inform AI that they have 0 chance to cast the spell so that AI doesn't accidentally decide not to flee for some reason. One unnecessary check in spell rating was removed and vanillaRateSpell behavior became more sensible.

Works fine in my testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,7 @@
     Bug #5104: Black Dart's enchantment doesn't trigger at low Enchant levels
     Bug #5105: NPCs start combat with werewolves from any distance
     Bug #5110: ModRegion with a redundant numerical argument breaks script execution
+    Bug #5112: Insufficient magicka for current spell not reflected on HUD icon
     Bug #5123: Script won't run on respawn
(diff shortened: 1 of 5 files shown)


akortunov:
Since now we use the `nullptr, true, true` almost everywhere, probably it would be better to change default argument values.

Capostrophic:
OK, I think all calls have been updated.

akortunov:
There are more of `getSpellSuccessChance`. Since we changed default argument values, are you sure that we do not need to modify these usages as well?

Capostrophic:
No? The rest is totally fine with default arguments. I inverted the edits, basically.

akortunov:
A merge conflict ...

Capostrophic:
Yeah one second.

PR #2460, opened 2019-07-30
Don't force-trace down swimming actors (bug #5089)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5089)

Added another check. Dreughs and slaughterfish now seem to be swimming fine at mid-water levels.


Capostrophic:
Yep, according to Hrnchamd it's what Morrowind does before tracing down the actor.

PR #2462, opened 2019-07-31
[Incomplete, do not merge] Issue 5097: Handle ints/floats passed as strings

ampaulg:
https://gitlab.com/OpenMW/openmw/issues/5097

For names in quotes, Scanner::scanName now also checks for whether they're ints or floats, and then calls the appropriate parser function as necessary.

Tested by using the modcurrenthealth console command on the player with the different possible argument formats.

### DIFF
--- a/components/compiler/scanner.cpp
+++ b/components/compiler/scanner.cpp
@@ -4,6 +4,7 @@
 #include <cctype>
 #include <sstream>
 #include <iterator>
+#include <regex>
 
 #include "exception.hpp"
 #include "errorhandler.hpp"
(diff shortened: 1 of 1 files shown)


Capostrophic:
I think this isn't wise. String parser can't parse numbers and this will probably cause something like messagebox "1.0" to fail.

ampaulg:
You're right, passing "1.0" to functions like messagebox and streammusic has errors because the incorrect type gets passed with this strategy.

Is it possible for the scanner or parser to determine the necessary types for the function, and then try to convert to the required type as needed?

Capostrophic:
See parseArguments(); different parsers parse different types of arguments. String parser parses strings. Expression parser parses numerical arguments, such as (in case of parseName) variables which correspond to numbers. As in the bug report's description, my thought was to make expression parser's parseName function understand not just variable names, but also quoted numbers.

ampaulg:
Thanks for the info, I'll take another look. I'll close this PR for now and put up a new one when it's redone.

PR #2468, opened 2019-08-01
Reset path after AiTravel is fast forwarded (bug #5037)

elsid:
Otherwise actor will keep going to the first path point that may be unreachable.


Capostrophic:
Seems to work OK.

PR #2472, opened 2019-08-04
Purge unnecessary recast stuff

Capostrophic:
Removes a few things:
1. RecastDemo application source code.
2. RecastNavigation documentation.
3. RecastDemo tests.
4. Other Recast-repo-specific stuff like CI files and some readme files.

CMake/gitignore files were adjusted accordingly.
35k removed lines overall, out of 65500 lines that were added initially.

In my understanding we only want to supply recastnavigation whole source code to be able to use any part of its libraries without the necessity to bring over the files from the original repo or adding it as a dependency. However, these specific parts of it are virtually unusable in OpenMW code itself (they aren't even built by default) and just bloat the code base. We don't want to actually have a pristine copy of recastnavigation present in the codebase, and anyone who wants to check out recast should just use the original repo to retrieve the necessary stuff for testing and development. Please correct me and turn down this PR if I'm wrong.


akortunov:
LGTM.

psi29a:
Once recastnavigation has debian/windows/mac pre-compiled libs, we can drop this external dep altogether...

PR #2474, opened 2019-08-06
[WIP, do not merge] Issue 5097: Handle ints/floats passed as strings

ampaulg:
https://gitlab.com/OpenMW/openmw/issues/5097

If a script passes an int or float in quotes, exprparser can now handle that in parseName by checking if the string can be converted to an int or float. Unlike my previous attempt, this one does not break functions that expect strings (eg. messagebox "1.0" still works).

One thing to note is that the CS will still give "unexpected name" warnings when numbers are passed this way even though they're now "valid."

Tested by running scripts and console functions with quotes around their number arguments.

### DIFF
--- a/components/compiler/exprparser.cpp
+++ b/components/compiler/exprparser.cpp
@@ -5,6 +5,8 @@
 #include <algorithm>
 #include <stack>
 #include <iterator>
+#include <regex>
+#include <string>
 
 #include <components/misc/stringops.hpp>
 
@@ -322,6 +324,22 @@ namespace Compiler
             if (mExplicit.empty() && getContext().isId (name2))
             {
                 mExplicit = name2;
+                return true;
+            }
+
+            // In case someone tried passing an int in quotes
+            if (regex_match(name, std::regex("[+-]?[0-9]+")))
+            {
+                pushIntegerLiteral (stof(name));
+                mTokenLoc = loc;
+                return true;
+            }
+
+            // In case someone tried passing a float in quotes
+            if (regex_match(name, std::regex("[+-]?[0-9]+.[0-9]+")))
+            {
+                pushFloatLiteral (stof(name));
+                mTokenLoc = loc;
                 return true;
             }
         }


Capostrophic:
Ok, this seems to be the right idea, although I suspect that stof doesn't actually understand separators on non-English locales. But no big deal. I'll check this out later today.

components/compiler/exprparser.cpp
```diff
@@ -5,6 +5,8 @@
 #include <algorithm>
 #include <stack>
 #include <iterator>
+#include <regex>
+#include <string>
```
Capostrophic: Redundant.

components/compiler/exprparser.cpp
```diff
@@ -324,6 +326,22 @@ namespace Compiler
                 mExplicit = name2;
                 return true;
             }
+
+            // In case someone tried passing an int in quotes
+            if (regex_match(name, std::regex("[+-]?[0-9]+")))
+            {
+                pushIntegerLiteral (stof(name));
```
Capostrophic: pushIntegerLiteral accepts an integer, not a floating point number, you should use stoi directly.

components/compiler/exprparser.cpp
```diff
@@ -324,6 +326,22 @@ namespace Compiler
                 mExplicit = name2;
                 return true;
             }
+
+            // In case someone tried passing an int in quotes
+            if (regex_match(name, std::regex("[+-]?[0-9]+")))
```
elsid: Why don't just catch stof exception? If string has invalid format or overflow happens `stof` throws one. With current approach overflow is not handled. Same thing with `stoi`.

akortunov: IIRC, Morrowind in such case just uses the default value (0 for digits).
For example Morrowind treats the
`player->additem "steel longsword" "aaa"`
as a
`player->additem "steel longsword" 0`

Capostrophic: Which means that it really does use atof. Oh woe.

ampaulg: How does the default value work? I just tried `player->setcurrentfatigue "shouldBeZero"` and vanilla thinks of it as a name and then does nothing rather than using 0.

numberParsing

Capostrophic: In-game and TESCS compilers might be slightly different. Can you verify that TESCS isn't in fact able to compile a script with such an instruction?

akortunov:
https://github.com/OpenMW/openmw/pull/2539

PR #2476, opened 2019-08-07
Use magic glow for enchanted arrows

akortunov:
Implements a [feature #5122](https://gitlab.com/OpenMW/openmw/issues/5122).

Summary of changes:
1. Move `getEnchantmentColor` from `Animation` to `Class` (there was a relevent TODO).
2. Move `GlowUpdater` and related stuff from `Animation` to `SceneUtils`.
3. Add glow to enchanted arrow/bolt when we attach it to actor.
4. Add glow to enchanted projectile when we launch it.
5. Do not use a separate GlowUpdater for every arrow in quiver.

Notes:
1. An amount of changes is a quite large just because I had to move some stuff from `Animation` to use it in the `ProjectileManager`
2. I did not get the point of `mGlowUpdater` variable - it stores only last GlowUpdater, created for given animation file. So I do not change this variable when create a GlowUpdater for attached arrows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -157,6 +157,7 @@
     Feature #5051: Provide a separate textures for scrollbars
     Feature #5094: Unix like console hotkeys
     Feature #5098: Allow user controller bindings
+    Feature #5122: Use magic glow for enchanted arrows
     Task #4686: Upgrade media decoder to a more current FFmpeg API
     Task #4695: Optimize Distant Terrain memory consumption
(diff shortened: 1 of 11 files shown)


Capostrophic:
Awesome. Works nice, looks good.

AnyOldName3:
Was the last commit here the thing that got reverted?

akortunov:
> Was the last commit here the thing that got reverted?

Yes, the b7c214b1bc395c5ae7291d1e14ef46b834652a3a.

PR #2478, opened 2019-08-08
[Regression] Make sure local path includes trailing slash

Capostrophic:
Launcher assumes it does but parent_path doesn't have one, so it breaks some stuff.

### DIFF
--- a/components/files/linuxpath.cpp
+++ b/components/files/linuxpath.cpp
@@ -87,7 +87,7 @@ boost::filesystem::path LinuxPath::getLocalPath() const
     {
         if (readlink(path, &binPath[0], binPath.size()) != -1)
         {
-            localPath = boost::filesystem::path(binPath).parent_path();
+            localPath = boost::filesystem::path(binPath).parent_path() / "/";
             break;
         }
     }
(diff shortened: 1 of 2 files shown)


AnyOldName3:
Is it definitely best to add the separator as a character instead of asking Boost for a magic platform-specific one that'll work everywhere?

Capostrophic:
I thought about it. There's boost::filesystem::preferred_separator in Boost::Filesystem, however Windows, BSD and Linux kernel based systems all accept forward slash as a separator so I decided to keep it simple.

psi29a:
The good news is... Windows (Microsoft), thanks to their WSL/WSL2 work have been making things easier for developers which includes little things like supporting '/' which makes things like magic_path stuff in boost::filesystem and std::filesystem unnecessary... so long as we don't support windowsXP of course. ;)

psi29a:
btw... just as we follow EoL for  Debian/Ubuntu and Mac, here is one for windows:
https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet

Client operating systems | End of mainstream support | End of extended support
Windows 8.1 | January 9, 2018 | January 10, 2023
Windows 7, service pack 1* | January 13, 2015 | January 14, 2020

PR #2479, opened 2019-08-08
normalize weights (fix bug #5125)

mp3butcher:
fix https://gitlab.com/OpenMW/openmw/issues/5125


components/sceneutil/riggeometry.cpp
```diff
@@ -355,7 +355,17 @@ void RigGeometry::setInfluenceMap(osg::ref_ptr<InfluenceMap> influenceMap)
             vec.emplace_back(std::make_pair(boneName, bi.mInvBindMatrix), weightPair.second);
         }
     }
-
+    /// normalize weights 
+    for (auto& vertexPair : vertex2BoneMap)
+    {
+        float norm = 0.0f;
+        for(std::vector<BoneWeight>::iterator wit = vertexPair.second.begin(); wit != vertexPair.second.end(); ++wit )
```
Capostrophic: Syntax ~~diabetes~~ sugar suggestion:
```cpp
// Normalize bone weight
float sum = 0.f;
for (const BoneWeight& weight : vertexPair.second)
    sum += weight.second;
if (sum > 0.f && sum != 1.f)
    for (BoneWeight& weight : vertexPair.second)
        weight.second /= sum;
```

mp3butcher: ...I'm C99 over conditionnated :)

components/sceneutil/riggeometry.cpp
```diff
@@ -355,7 +355,18 @@ void RigGeometry::setInfluenceMap(osg::ref_ptr<InfluenceMap> influenceMap)
             vec.emplace_back(std::make_pair(boneName, bi.mInvBindMatrix), weightPair.second);
         }
     }
-
+    
+    /// normalize weights 
+    for (auto& vertexPair : vertex2BoneMap)
```
akortunov: Do we really need a separate loop here? Why just to do not use a loop below?

Capostrophic:
Also, add a changelog entry.

components/sceneutil/riggeometry.cpp
mp3butcher: that's how normalization works

mp3butcher: an other method would be to deffer normalization to runtime and you don't want that

mp3butcher: perhaps you are right will see that tomorrow 'im to drunk right now

mp3butcher: i do that drunk....

mp3butcher:
plize check..I shouldn't code drunk

mp3butcher:
oh boy i mixed branches ...

components/sceneutil/riggeometry.cpp
Capostrophic: No, it's not resolved.

components/sceneutil/riggeometry.cpp
```diff
@@ -350,16 +361,17 @@ void RigGeometry::setInfluenceMap(osg::ref_ptr<InfluenceMap> influenceMap)
...
+            vec.first += weightPair.second;
+            vec.second.emplace_back(std::make_pair(boneName, bi.mInvBindMatrix), weightPair.second);
         }
     }
 
     Bone2VertexMap bone2VertexMap;
     for (auto& vertexPair : vertex2BoneMap)
-    {
-        bone2VertexMap[vertexPair.second].emplace_back(vertexPair.first);
+    {
```
Capostrophic: Redundant whitespace. Same on the end of the next line.

components/sceneutil/riggeometry.cpp
AnyOldName3: So I should click the unresolve button?

components/sceneutil/riggeometry.cpp
```diff
@@ -350,16 +352,22 @@ void RigGeometry::setInfluenceMap(osg::ref_ptr<InfluenceMap> influenceMap)
 
         for (auto& weightPair: bi.mWeights)
         {
-            std::vector<BoneWeight>& vec = vertex2BoneMap[weightPair.first];
-
-            vec.emplace_back(std::make_pair(boneName, bi.mInvBindMatrix), weightPair.second);
+            std::pair<float, std::vector<BoneWeight> >& vec = vertex2BoneMap[weightPair.first];
```
AnyOldName3: You've made a typedef for this a few lines above, and then not used it. Surely it's best to be consistent?

mp3butcher: arf forgot this one

components/sceneutil/riggeometry.cpp
Capostrophic: mp3butcher said something about how it doesn't work because BoneWeight is private but I'm not sure.

mp3butcher: @Capostrophic 
it's just that  can't use an inline function because of this....not a big deal

Capostrophic:
Ultimately this is not something Morrowind does and it can't fix all inconsistencies of the behavior compared to vanilla (which make artists' life harder), we'll have to implement the limitations of Morrowind's hardware skinning to properly fix the issue.

mp3butcher:
WTF
You don't even diagnose the bug that you close my fix!
Further I really would like a proof that morrowind don't normalize weights

Capostrophic:
I don't have any proof that it doesn't other than "I was told by multiple knowledgeable people who worked with modelling a lot that it doesn't", but do you have the proof to the contrary?

I did test this and multiple situations of the skinning issue. Its usage doesn't lead to the same behavior as in Morrowind in a different case, so it's better to take precautions to make sure we don't overcorrect the model format that isn't supposed to be ours.

akortunov:
> Further I really would like a proof that morrowind don't normalize weights

IIRC, NetImmerse (and a plenty of other game engines from 2000's) just inherited the "4 bones per vertex" limit from hardware skinning, provided by DirectX 8. And assets for NetImmerse/GameBryo consider that this limitation is present.

A quote from Ogre3D forum:

> DirectX 9 has a limit of 256 world matrices (bones) within one batch.
The Flexible Vertex Format (used for fixed function) datatype for bone blend indices is D3DDECLTYPE_UBYTE4, which gives 4 bone indices affecting each vertex, with a max of 256 possible bones.

OpenMW uses a software skinning, so we do not  have such limitation ATM, what may cause issues, if assets have garbage in their additional bone weight, which Morrowind just ignores.

Also there could be additional quirks, which Morrowind uses, so the "4 bones per vertex" limit is not a silver bullet (weight normalization too).

Capostrophic:
Yeah limiting skinning to 4 bones per vertex didn't work for me, but that isn't the only hardware skinning limitation.

AnyOldName3:
What do the failing meshes do in NifSkope? If they work there, then we have a reference implementation we can use.

Anyway, we know for sure that only the four most significant bone weights are considered per vertex (unless Bethesda cocked up and uses something like the first four instead, but that's quick to test), so we're not getting less accurate if we implement that. It should also improve performance slightly, and make fast hardware acceleration much more simple to implement in the future (which I think we should do, as it'll take a good chunk off the amount of data we upload to the GPU each frame, even if the actual skinning maths doesn't take long on the CPU).

There's one possibility I'm not sure anyone's considered, by the way. Could it be the case that Morrowind is normalising the four vertex weights it actually uses? Barring any other tricks, this would be a reasonable explanation for why normalising everything helps sometimes and pruning weights helps other times.

mp3butcher:
> Anyway, we know for sure that only the four most significant bone weights are considered per vertex 
this implies vertexweights selection (sort by weight then select) so normalizing (accum then divide) during this process is quasi free (and mandatory for user abstration)..i doubt vanilla doesn't do that. 

so If you are sure there's vertexweight selection then it's quasi sure there's normalization

AnyOldName3:
The later Bethesda games don't normalise the weights. There's no need, as it's something that can be done ahead of time by whatever mesh editor you're using, and it's conceivable that you might implement some effect by having a greater-than-one vertex weight. Whether or not it's a good idea also has never been known to have any bearing on if Bethesda Game Studios did a thing.

Also, sorting by weight and selecting the biggest four is an $`O(n\log(n))`$ operation, but it's faster to scan across the array once, keeping track of the four biggest so far as you go, as that's $`O(n)`$.

GitHub needs to add inline LaTeX support, as apparently those formulae would have been pretty on GitLab.

PR #2480, opened 2019-08-08
fix hand shield direction (bug #4783)

mp3butcher:
not sure
it seems to fix https://gitlab.com/OpenMW/openmw/issues/4783


Capostrophic:
No, it's completely wrong.

PR #2481, opened 2019-08-08
Handle NiTriStrips (feature #5121)

Capostrophic:
[Feature request](https://gitlab.com/OpenMW/openmw/issues/5121).

Implements just about everything about Morrowind's version of NiTriStrips: loading of data, handling of non-skinned and skinned shapes, morphs, direct rendering thanks to OpenGL's and OpenSceneGraph's support of triangle strip primitives and collision shape generation (triangulated). Some refactoring had to take place to make sure NiTriStrips can take place of NiTriShape seamlessly.

Examples of now working models (stripified Suzanne with collisions enabled and disabled, outfit from Sylgia the Seamstress)
screenshot016
screenshot013


Capostrophic:
Hm. Looks like strip lengths isn't a single value, but is rather set for each strip individually.

Capostrophic:
[Alternative branch](https://github.com/Capostrophic/openmw/tree/triangulation) where instead of using triangle strips directly they are triangulated upon loading. On my hardware it's faster (as fast as NiTriShape). At the very least it leads to less draw calls (multiple strips -- multiple draw calls normally). However, I'm not sure if it's the case just for me.

Here's a pack of Suzanna test meshes I got, check meshes with strips in their names.
[g7.zip](https://github.com/OpenMW/openmw/files/3483634/g7.zip)

akortunov:
Seems to work well. I suggest to merge it if there are no objections.

PR #2483, opened 2019-08-09
Revert "Do not use a separate GlowUpdater for every arrow in quiver"

akortunov:
It sometimes leads to
```
Warning: detected OpenGL error 'invalid operation' at after RenderBin::draw(..)
```

errors, so I have to revert it.


AnyOldName3:
I guess I'm going to have to take a look at this when the *revert the revert* PR is opened.

PR #2484, opened 2019-08-09
Allow SwimRunForward state WalkForward fallback (bug #5126)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5126)

Shuffled the relevant code a little to enable walk animation fallback for swimming animations. Should allow swimming creatures to use their WalkForward animation if it's present and running animation is not.

Right, works on a slaughterfish with running animation artificially removed.


akortunov:
IIRC, there was a similar issue with shark frol Lothavor's Legacy. Can you check if it is fixed as well?

Capostrophic:
The shark is swimming more or less correctly with the PR.

akortunov:
Seems to work well.

PR #2486, opened 2019-08-10
Rewrite flickering/pulsing to work like vanilla (bug #4952)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/4952)

Rewrote flickering/pulsing code completely according to research information received from NullCascade. It has been greatly simplified, our previous implementation was absurdly complex and inaccurate.

Cavern of the Incarnate and torch_infinite_time torch lighting looks fine.


akortunov:
LGTM.

PR #2488, opened 2019-08-10
fix Sunglare bug with osg 3.6

mp3butcher:


Capostrophic:
It just works.

AnyOldName3:
Can we not merge OSG stuff without someone who knows why it works reviewing it? This looks suspiciously like a hack.

mp3butcher:
@AnyOldName3 
all oq setup in sky.cpp is hacky because it circumvents classic usage
this pr is the only way for validgeometry to be true

AnyOldName3:
To be honest, when I first saw how the sky OQ was done, I was pretty surprised, as I'm sure it would have been less work for Scrawl to create a new node than to use the OSG OQ one the way he did.

Robert can be pretty keen on maintaining backwards compatibility, though, so if we get him in the right mood, it might be best to persuade him to restore the old behaviour in OSG 3.6.5, as OSG 3.6.4 has issues with OpenMW anyway (and he went and tagged the release before we'd sorted them out).

mp3butcher:
> To be honest, when I first saw how the sky OQ was done, I was pretty surprised, as I'm sure it would have been less work for Scrawl to create a new node than to use the OSG OQ one the way he did.
> 
as oqn is part of the framework, you HAVE to subclass it. and even subclassing is not a panacea (see oq pr). That's why scrawl hasn't created custom oqn.
What could be done is using regular geomety with a OQDrawCallback...but it's lot of effort for little benefit

mp3butcher:
I opened an issue https://github.com/openscenegraph/OpenSceneGraph/issues/810 if you want to add your arguments

AnyOldName3:
Implementing a custom drawable or geometry subclass that did the queries directly in its `drawImplementation` would have been completely consistent with how OSG does things, and a callback could have worked fine, too. The provided OSG class is only really intended for occlusion culling.

mp3butcher:
This pr purpose was to fix a bug, not to question chosen design....

PR #2490, opened 2019-08-12
fix PlaceAtXXX with possible collaterals  (bug #5095)

mp3butcher:
https://gitlab.com/OpenMW/openmw/issues/5095
Clamp placed objects to the floor;
as i touched SafePlaceObject there may be collaterals damage ... 
feedback and insights are very welcome

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1514,7 +1514,10 @@ namespace MWWorld
         ipos.rot[2] = referenceObject.getRefData().getPosition().rot[2];
 
         MWWorld::Ptr placed = copyObjectToCell(ptr, referenceCell, ipos, ptr.getRefData().getCount(), false);
-        adjustPosition(placed, true); // snap to ground
+
+        if (!placed.getClass().isActor())
+            adjustPosition(placed, true); // snap only objects to ground
+
         return placed;
     }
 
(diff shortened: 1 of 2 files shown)


apps/openmw/mwworld/worldimp.cpp
```diff
@@ -1514,7 +1514,10 @@ namespace MWWorld
         ipos.rot[2] = referenceObject.getRefData().getPosition().rot[2];
 
         MWWorld::Ptr placed = copyObjectToCell(ptr, referenceCell, ipos, ptr.getRefData().getCount(), false);
-        adjustPosition(placed, true); // snap to ground
+
+        if (!placed.getClass().isActor())
```
mp3butcher: actor have physics so no interest in snap them to the ground

mp3butcher: I removed it in https://github.com/OpenMW/openmw/pull/2490/commits/2f1fb27d18cb9e9c96df06995adb1b586d78aa56 but it's less fun than letting them fall and squash on the ground

akortunov:
According to my testing, placing does not trace actors at all - physics system does it during next update. Non-actor objects are not affected by physics at all.

mp3butcher:
@akortunov 
Please reopen this pr, i would want to comply with vanilla behavior, but community feature description have divergence:
It seams your experience diverges from what had been told here https://gitlab.com/OpenMW/openmw/issues/5095

Further, please read comments of this pr , I added
https://github.com/OpenMW/openmw/commit/2f1fb27d18cb9e9c96df06995adb1b586d78aa56 
just to comply with vanilla (removing the ability to squash invocated creatures on the floor)
If you say this is vanilla behavior, i would be glad to reverse this commit

PR #2491, opened 2019-08-14
Implement NiPalette support (feature #4882)

Capostrophic:
[Feature request](https://gitlab.com/OpenMW/openmw/issues/4882).

Implements NiPalette reading and handling.
screenshot032

The NiPalette data is loaded into a vector of unsigned integer color entries. Technically there's no entry limit but there can only be up to 256 usable colors due to the format restraints. When a palettized image is loaded, it's reconstructed into a RGB/RGBA image using the palette and the original color index data contained. When the palette has hasAlpha flag turned off, its colors are considered opaque and their alpha value is discarded. When there are less entries than 256, the rest of the palette is filled with 0 (black color).

Two more internal texture pixel formats are now handled, PAL8 and PALA8, with the relevant enum corrected. One has alpha and one doesn't (guess which is which).

I made a couple of fixes along the way:
1. The last three mipmap levels in mipmap data of internal textures were discarded for some reason, which caused issues when there were *just* three mipmap levels. There doesn't seem to be any adverse effect to loading the entire mipmap level data.
2. For consistency all unsigned integers in NiPixelData are now handled as, well, unsigned integers.

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -1276,9 +1276,11 @@ namespace NifOsg
             switch (pixelData->fmt)
             {
             case Nif::NiPixelData::NIPXFMT_RGB8:
+            case Nif::NiPixelData::NIPXFMT_PAL8:
                 pixelformat = GL_RGB;
                 break;
             case Nif::NiPixelData::NIPXFMT_RGBA8:
+            case Nif::NiPixelData::NIPXFMT_PALA8:
                 pixelformat = GL_RGBA;
                 break;
             default:
@@ -1293,7 +1295,7 @@ namespace NifOsg
             int height = 0;
 
             std::vector<unsigned int> mipmapVector;
-            for (unsigned int i=0; i<pixelData->mipmaps.size()-3; ++i)
+            for (unsigned int i=0; i<pixelData->mipmaps.size(); ++i)
             {
                 const Nif::NiPixelData::Mipmap& mip = pixelData->mipmaps[i];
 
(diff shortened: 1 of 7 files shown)


components/nifosg/nifloader.cpp
```diff
@@ -1319,10 +1327,43 @@ namespace NifOsg
                 return nullptr;
             }
 
-            unsigned char* data = new unsigned char[pixelData->data.size()];
-            memcpy(data, pixelData->data.data(), pixelData->data.size());
+            const std::vector<unsigned char>& pixels = pixelData->data;
+            if (pixelData->fmt != Nif::NiPixelData::NIPXFMT_PAL8)
+            {
+                unsigned char* data = new unsigned char[pixels.size()];
```
akortunov: IIRC, in C++11 vectors have the `data()` method to convert them to array, so maybe `copy()` is not needed here.

akortunov: Oh, never mind, OSG does not copy data in `setImage()` itself, it just stores a reference to the input array.

components/nifosg/nifloader.cpp
```diff
@@ -1319,10 +1327,43 @@ namespace NifOsg
...
+                std::copy(pixels.begin(), pixels.end(), data);
```
elsid: Not sure what do you mean saying "more efficient", but `std::copy` may [use](https://godbolt.org/z/0fZGFY) `memmove` for implementation that is actually [slower](https://stackoverflow.com/questions/1201319/what-is-the-difference-between-memmove-and-memcpy) way to copy memory when compare with `memcpy`. Avoid to copy any memory would be prefered way if it is possible.

components/nifosg/nifloader.cpp
```diff
@@ -1319,10 +1327,43 @@ namespace NifOsg
...
+                        Log(Debug::Info) << "Color index is not present in the palettized texture palette, ignoring texture";
+                        return nullptr;
+                    }
+                    converted.emplace_back((color >>  0) & 0xFF);
+                    converted.emplace_back((color >>  8) & 0xFF);
+                    converted.emplace_back((color >> 16) & 0xFF);
+                    converted.emplace_back((color >> 24) & 0xFF);
+                }
+                unsigned char* data = new unsigned char[converted.size()];
+                std::copy(converted.begin(), converted.end(), data);
```
elsid: Why do you need to copy? Just use `data` instead of `converted` variable for transformation destination.

components/nifosg/nifloader.cpp
Capostrophic: [This stackoverflow question answers imply otherwise.](https://stackoverflow.com/questions/4707012/is-it-better-to-use-stdmemcpy-or-stdcopy-in-terms-to-performance).
Perhaps it's better to let the compiler decide for the coder how exactly to copy stuff and not assume the coder knows best.

components/nifosg/nifloader.cpp
Capostrophic: OK.

components/nifosg/nifloader.cpp
akortunov: Just for reference: we use the `std::copy()` a lot more often in our codebase (excepts of `external` folder, IOCS and R&D use the `memcpy` internally) than `memcpy` since our general rule is to prefer C++ features over C ones.

But I'd recommend to do not touch the `memcpy` in this PR since it is a codestyle issue, not really related to paletted textures. IMO, it would be better to create a separate refactoring pull request to unify memory copying in OpenMW.

akortunov:
A quick test did not show any issues. It would be nice to have additional meshes to test this feature, though.

components/nifosg/nifloader.cpp
Capostrophic: Okay.

akortunov:
If there are no objections, I suggest to squash commits and merge this PR.

PR #2492, opened 2019-08-15
Use a common doors rotation code in the "lock" command handler

akortunov:
Fixes [bug #5134](https://gitlab.com/OpenMW/openmw/issues/5134) by using a common doors rotation code for scripting system.

Note: this PR does not fix an issue with bar doors (which are not doors at all).

### DIFF
--- a/apps/openmw/mwscript/miscextensions.cpp
+++ b/apps/openmw/mwscript/miscextensions.cpp
@@ -189,12 +189,6 @@ namespace MWScript
                     if (ptr.getTypeName() == typeid(ESM::Door).name() && !ptr.getCellRef().getTeleport())
                     {
                         MWBase::Environment::get().getWorld()->activateDoor(ptr, 0);
-
-                        float xr = ptr.getCellRef().getPosition().rot[0];
-                        float yr = ptr.getCellRef().getPosition().rot[1];
-                        float zr = ptr.getCellRef().getPosition().rot[2];
-
-                        MWBase::Environment::get().getWorld()->rotateObject(ptr, xr, yr, zr);
                     }
                 }
         };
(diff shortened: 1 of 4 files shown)


apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -190,11 +192,11 @@ namespace MWScript
...
 
-                        float xr = ptr.getCellRef().getPosition().rot[0];
-                        float yr = ptr.getCellRef().getPosition().rot[1];
-                        float zr = ptr.getCellRef().getPosition().rot[2];
+                        osg::Quat xr = osg::Quat(ptr.getCellRef().getPosition().rot[0], -osg::X_AXIS);
+                        osg::Quat yr = osg::Quat(ptr.getCellRef().getPosition().rot[1], -osg::Y_AXIS);
+                        osg::Quat zr = osg::Quat(ptr.getCellRef().getPosition().rot[2], -osg::Z_AXIS);
 
-                        MWBase::Environment::get().getWorld()->rotateObject(ptr, xr, yr, zr);
+                        MWBase::Environment::get().getWorld()->rotateWorldObject(ptr, zr * yr * xr);
```
Capostrophic: @elsid the navigator seems to ignore rotations performed through rotateWorldObject, so it may not perform as intended with RotateWorld instructions (it currently ignores script-locked doors with this fix). Any thoughts?

elsid: That's right. #2492 fixes this.

akortunov:
Ok, a used approach was changed - instead of rotationg door manually in the `lock` handler, just re-use a common door rotation code and allow the `World::activateDoor(const Ptr &door, int state)` do the job for us by fast-forwarding the door state.

Capostrophic:
Things to do:
1. Update the issue name since it's not in fact the issue of local/world coordinate dichotomy but that Lock-reset doors used a rotation order different from the ordinary door rotations and reset the rotations completely.
2. Resolve the changelog conflict.

akortunov:
Done.

PR #2494, opened 2019-08-15
[WIP] Open door when it is on the way to a next path point (bug #5073)

elsid:
This fixes the case described in the [bug](https://gitlab.com/OpenMW/openmw/issues/5073) (`Ildogesto` in `Galom Daeus, Entry`), but may introduce some other issues. However when I start a fight `Ildogesto` is able to open that door (`door_dwrv_inner00`).

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -306,8 +306,6 @@ namespace MWBase
 
             virtual bool castRay (float x1, float y1, float z1, float x2, float y2, float z2) = 0;
 
-            virtual MWWorld::Ptr castRay(const osg::Vec3f& from, const osg::Vec3f& to, int mask) const = 0;
-
             virtual void setActorCollisionMode(const MWWorld::Ptr& ptr, bool internal, bool external) = 0;
             virtual bool isActorCollisionEnabled(const MWWorld::Ptr& ptr) = 0;
 
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -306,6 +306,8 @@ namespace MWBase
 
             virtual bool castRay (float x1, float y1, float z1, float x2, float y2, float z2) = 0;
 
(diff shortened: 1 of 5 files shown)


Capostrophic:
The raycast is too costly. Vanilla behavior is opening the door on collision with it, which is better than 0.44.0 behavior of opening the door if the NPC has been stuck for too long. It'll be necessary to do what vanilla does to avoid user complaints anyway...

elsid:
I've tried to profile `AiPackage::isDoorOnTheWay` [function](https://github.com/OpenMW/openmw/pull/2494/commits/c32872fb16fe826c5409e31111bd50466595abf5#diff-04b7df27787fe5bb1ed28ca5b03f53b0R416-R421) excluding first `if`. Visited locations where started fights and run from actors closing doors multiple times:
* Addamasartus
* Ald-ruhn, Llether Vari: Enchanter
* Galom Daeus, Entry
* Vivec, Redoran Records
* Vivec, Redoran Prison Cells

Got these measurements:

| function | openDoors | getNearbyDoor | isDoorOnTheWay |
| --- | --- | --- | --- |
| number | 3106 | 2999 | 374 |
| min, ns | 385 | 258 | 1273 |
| max, ns | 47310744 | 311591 | 42414 |
| mean, ns | 54718.96 | 2535.88 | 13426.3 |
| median, ns | 3101.5 | 2421 | 13762 |
| 0.95-quantile, ns | 19065.5 | 4230.5 | 22645.25 |
| 0.98-quantile, ns | 30199.5 | 5230.76 | 23934.1 | 
| 0.99-quantile, ns | 50706.8 | 5688.44 | 24975 |
| sum, ns | 169957090 | 7605089 | 5021452 |

For a single frame isDoorOnTheWay is slower by ~5 times than getNearbyDoor but in total it takes ~1.5 times less time.

akortunov:
>  Vanilla behavior is opening the door on collision with it

Unfortunately, we do not store detected collisions to access them in the AI code.

elsid:
Sorry for merge, hit wrong button, reverted.

PR #2496, opened 2019-08-16
Change the default reverb preset to improve underwater FX transition

Capostrophic:
Living room preset seems good enough to make the transition to underwater reverb preset less jarring and bring peace to CMAugust's soul.

Not bothering to add a changelog entry since most users don't pay attention to this and there's technically no issue with FX transitions (OpenAL-Soft does everything just right).


akortunov:
Did you contact with CMAaugust to check if this change really brings peace to his soul?

Capostrophic:
Well, he *was* tagged to check in with this, so I think he'll come around soon in time for the appveyor artifact to be built.

akortunov:
BTW, personally I like this change.

Capostrophic:
CMAugust thinks it's good. Given that there are thumbs up from two people, I'm going to merge this.

PR #2507, opened 2019-08-27
Fix actors stuck in closing door (bug #5138)

elsid:
Two things are done:
1. Add more variation in changing direction for actor when it tries to avoid door using `AiAvoidDoor` package. It helps solve the issue but with some delay for actor to get out.
2. Rotate door in opposite direction when collision with actor is detected. Previously when collision with any actor is detected opening or closing door rotation was rolled back to begin of the frame. So until actor will find the right direction to move door will not rotate. If we start to rotate door in opposite direction it helps actor to find a way eventually. This may produce issues with visibly operlapping actors and doors. Probably need more detailed testing for this.

### DIFF
--- a/apps/openmw/mwmechanics/aiavoiddoor.cpp
+++ b/apps/openmw/mwmechanics/aiavoiddoor.cpp
@@ -1,5 +1,7 @@
 #include "aiavoiddoor.hpp"
 
+#include <components/misc/rng.hpp>
+
 #include "../mwbase/world.hpp"
 #include "../mwbase/environment.hpp"
 #include "../mwbase/mechanicsmanager.hpp"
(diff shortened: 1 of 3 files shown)


Capostrophic:
Improves [this](https://gitlab.com/OpenMW/openmw/issues/4083) (as in makes door touching behavior closer to vanilla) too but the door starts to vibrate violently when you touch it, following the player's idle animation oddly enough. Also the door still rotates backwards clipping into the character when you collide with its opposite side.

elsid:
> the door starts to vibrate violently when you touch it, following the player's idle animation oddly enough. Also the door still rotates backwards clipping into the character when you collide with its opposite side

Thanks for checking this. I haven't done such testing. Probably we should drop 555cc33 and use only 667c264 to have at least something.

elsid:
Removed door vibration. Changed direction selection to random instead of iteration.

PR #2508, opened 2019-08-27
Handle death event manually if needed before disposing a corpse

akortunov:
Implements [feature #5146](https://gitlab.com/OpenMW/openmw/issues/5146).

Just increment death counter and run script instantly if death animation was not finished yet, as in MCP.

Notes:
1. We will need to update the [MCP](https://wiki.openmw.org/index.php?title=MCP) page as well.
2. Probably the `can loot during death animation` setting is not needed at all and can be wiped out now.
3. In case if we still want to keep an original buggy behaviour, we can extend the `can loot during death animation` instead.


Capostrophic:
Looks ok to me.

PR #2510, opened 2019-08-27
Open door when it is on the way to a next path point (bug #5073)

elsid:
Follow up for https://github.com/OpenMW/openmw/pull/2494 .

Now different approach.
1. Check does actor has a path partially helps. If actor is not going anywhere there is no point to open door.
2. Quite less expensive raycasting is used to detect collision with door - `btRayAabb`. Simply detect ray collision with door bounding box.

### DIFF
--- a/apps/openmw/mwmechanics/aipackage.cpp
+++ b/apps/openmw/mwmechanics/aipackage.cpp
@@ -13,6 +13,8 @@
 #include "../mwworld/cellstore.hpp"
 #include "../mwworld/inventorystore.hpp"
 
+#include "../mwphysics/collisiontype.hpp"
+
 #include "pathgrid.hpp"
 #include "creaturestats.hpp"
 #include "movement.hpp"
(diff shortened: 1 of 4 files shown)


apps/openmw/mwmechanics/aipackage.cpp
```diff
@@ -402,3 +413,13 @@ DetourNavigator::Flags MWMechanics::AiPackage::getNavigatorFlags(const MWWorld::
 
     return result;
 }
+
+bool MWMechanics::AiPackage::isDoorOnTheWay(const MWWorld::Ptr& actor, const MWWorld::Ptr& door) const
+{
+    const auto world = MWBase::Environment::get().getWorld();
+    const auto halfExtents = world->getHalfExtents(actor);
+    const auto position = actor.getRefData().getPosition().asVec3() + osg::Vec3f(0, 0, halfExtents.z());
+    const auto destination = mPathFinder.getPath().front() + osg::Vec3f(0, 0, halfExtents.z());
```
Capostrophic: front() will cause undefined behavior if the path is empty, maybe that should be fixed to avoid issues if someone tries to call isDoorOnTheWay elsewhere (even though that's unlikely).

elsid: Thanks for noticing but I intetionally made it that way, because there is no reason other than that to make logic of this function more complex and as you noticed there is no UB until this method will be called somewhere else. Do you have any proposals how to fix it? One way is to put `if` at the beginning and either return false or throw an exception. Exception will not help much, because function is just wrongly used when it throws. `assert` may help. At least debug builds will report an error. Don't access mPathFinder from here, pass front point instead as argument. Move this code to `openDoors` function to make it impossible to call. Which way do you prefer or maybe there is another one?

elsid:
Bump

akortunov:
Not exactly related to this PR, but AI seems to move to locked door, even if there is an unlocked one nearby:



A screenshot is from Seyda Neen, the guard is the player's follower.

PR #2511, opened 2019-08-30
Water shader cleanup

Capostrophic:
1. Removed some unused stuff.
2. Moved light scattering calculations to refraction path since they're unused for reflection-only path, though it probably doesn't make much of a difference.
3. Fog is now applied after rain ripples completely to prevent weirdness when the fog covers them.
4. Condensed the shader in general.


akortunov:
I suppose that AON3 will not review this PR...

akortunov:
Works fine for me.

PR #2512, opened 2019-08-31
Show spell cost in spell buying window (feature #5147)

Capostrophic:
[Feature 5147](https://gitlab.com/OpenMW/openmw/issues/5147).

Works like it works in Morrowind with the MCP feature enabled.

The spell magicka cost is added after the spell name in braces and has a singular or plural point suffix depending on the value.

Like this.
screenshot021

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -177,6 +177,7 @@
     Feature #5131: Custom skeleton bones
     Feature #5132: Unique animations for different weapon types
     Feature #5146: Safe Dispose corpse
+    Feature #5147: Show spell magicka cost in spell buying window
     Task #4686: Upgrade media decoder to a more current FFmpeg API
     Task #4695: Optimize Distant Terrain memory consumption
     Task #4789: Optimize cell transitions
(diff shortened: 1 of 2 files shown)


akortunov:
Screenshot_20190831_162946
I'd suggest to add a space.

Capostrophic:
Ok, I suspected something like this will happen, but I tried to keep it true to MCP behavior. Updated.

akortunov:
Basically, LGTM (if we can afford to have such derivation from valilla game).

Capostrophic:
We already update the GUI in some measure (alchemy dialog, repair dialog, etc.), this seems to be in the spirit of that.

akortunov:
Personally I prefer this approach:
https://gitlab.com/OpenMW/openmw/merge_requests/144

Capostrophic:
For better or worse that approach modifies *all* spell tooltips while this only affects the specific window.

akortunov:
> that approach modifies all spell tooltips
It is fixable.

Capostrophic:
Alright, I was going to pass the banner anyway.

PR #2513, opened 2019-09-03
Make sure failed pick/trap attempts are a crime (bug #5149)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5149)

Move some lines to make sure the crime event is always triggered when necessary, even if the lock is too complex to be opened

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -712,9 +712,6 @@ namespace MWMechanics
                     if (target.getCellRef().getLockLevel() > 0)
                     {
                         MWBase::Environment::get().getSoundManager()->playSound3D(target, "Open Lock", 1.f, 1.f);
-                        if (!caster.isEmpty())
-                            MWBase::Environment::get().getMechanicsManager()->objectOpened(getPlayer(), target);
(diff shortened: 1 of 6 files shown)


apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -723,6 +720,11 @@ namespace MWMechanics
                 }
                 else
                     MWBase::Environment::get().getSoundManager()->playSound3D(target, "Open Lock Fail", 1.f, 1.f);
+
+                // Failed attempt is a crime too
+                if (!caster.isEmpty())
+                    MWBase::Environment::get().getMechanicsManager()->objectOpened(getPlayer(), target);
```
akortunov: `objectOpened` is a quite weird name here, especially if we will call it during failed unlock attempts.
Maybe something like `onUnlockAttempt` would be better.

akortunov:
Excepts of naming complaint seems to work.

apps/openmw/mwmechanics/spellcasting.cpp
Capostrophic: That's debatable, but whatever. Renamed.

PR #2514, opened 2019-09-05
Generate sCrimeMessageReport only once per frame

akortunov:
Fixes [bug #5012](https://gitlab.com/OpenMW/openmw/issues/5012).

The main idea:
1. Store previous bounty value in the WindowsManager
2. During GUI update compare new value with stored
3. If values are not the same, update stored value and generate messagebox.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@
     Bug #4999: Drop instruction behaves differently from vanilla
     Bug #5001: Possible data race in the Animation::setAlpha()
     Bug #5004: Werewolves shield their eyes during storm
+    Bug #5012: "Take all" on owned container generates a messagebox per item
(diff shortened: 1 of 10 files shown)


Capostrophic:
While the change is technically intentional, the intention really was to maintain script backwards compatibility, not to allow multiple speech lines to be queued in the same frame. And to be fair that should never happen... Could there be just one actual speech line being played but multiple subtitles being displayed? That I could see being a thing. Then adding an argument into sayDone() that allows it not to be vanilla-compatible or adding some kind of sayStarted() could be necessary.

akortunov:
I added a new `sayActive` function, which seeks in both `mSaySoundsQueue` and `mActiveSaySounds`.  The `sayDone()` is used only in scripting system now.

Capostrophic:
LGTM

PR #2515, opened 2019-09-09
Fix #5155

Assumeru:
Note that the check for canLock in security.cpp should have been redundant.


apps/openmw/mwclass/container.cpp
```diff
@@ -321,7 +321,8 @@ namespace MWClass
 
     bool Container::canLock(const MWWorld::ConstPtr &ptr) const
     {
-        return true;
+        const MWWorld::LiveCellRef<ESM::Container> *ref = ptr.get<ESM::Container>();
+        return !(ref->mBase->mFlags & ESM::Container::Organic);
```
akortunov: Actually, organic containers can be locked in Morrowind. It just is not possible to lock them via spell, so you should add check to the lock effect handling instead of `canLock`.

Capostrophic: canLock isn't checked for scripted locking/unlocking (it just throws if the object physically can't be locked), so technically this is valid, especially since with the check removed for pickLock the only place where canLock is used is spellcasting. Though I'm not sure why *should* it be removed for lockpicking.

Assumeru: Uhm... guess what? Morrowind allows lockpicking on activators. Yay?

Capostrophic: Hm, okay then.

Assumeru:
Oh yeah, this allows locking NPCs which is a difference from vanilla as it just crashes. Didn't seem worth reimplementing.

apps/openmw/mwworld/class.cpp
```diff
@@ -146,17 +146,21 @@ namespace MWWorld
...
     {
-        throw std::runtime_error ("class does not support unlocking");
+        int lockLevel = ptr.getCellRef().getLockLevel();
+        ptr.getCellRef().setLockLevel(-abs(lockLevel)); //Makes lockLevel negative
     }
 
     bool Class::canLock(const ConstPtr &ptr) const
     {
-        return false;
+        return hasToolTip(ptr);
```
Capostrophic: This allows the player to lock/unlock almost any object using a spell, which is probably invalid.

Assumeru: That's definitely invalid.

Capostrophic:
Looks good to me. @akortunov?

Capostrophic:
That's all hypothetical and unlikely though. What's apparent is that class-independent locking/unlocking is not something harmful right now and is necessary for Morrowind compatibility.

I still think lock/unlock should be moved to cellref.cpp/hpp. And there's a changelog conflict to resolve.

Capostrophic:
akortunov surrendered and doesn't mind merging this anymore.

PR #2519, opened 2019-09-11
Updated to note water shaders may not be turned on

jefhai:


Capostrophic:
...What's the use case?

jefhai:
> ...What's the use case?

Users are confused the game is not installed correctly and unaware of the ability to turn shaders on. Aware water shaders are wip.

Capostrophic:
I don't think a person that doesn't realize that water settings probably exist if water looks weird can realize that a documentation exists and manage to find the solution to their problem there.

Especially if said solution is put into a completely irrelevant place like Morrowind assets installation.

jefhai:
> I don't think a person that doesn't realize that water settings probably exist if water looks weird can realize that a documentation exists and manage to find the solution to their problem there.
> 
> Especially if said solution is put into a completely irrelevant place like Morrowind assets installation.

This is actually in the Morrowind install section not the assets install sections. This is the guide he said he was using and pointed me too. I'm pretty sure people are following this guide because this user followed the guide and made the suggestion to have this note in.

Either way its a small note that some will find helpful.

AnyOldName3:
The water shader has nothing to do with installing Morrowind, though.

jefhai:
Thats not the point of the note @AnyOldName3. The guide is helping others install the game and if they think the game install is broken because the water looks "bad/wrong" then we should hint in the guide that it is off by default.

jefhai:
Maybe it's better in common problems? https://openmw.readthedocs.io/en/master/manuals/installation/common-problems.html

But I think it should go along with the guide we tell users to use when installing and getting the game started.

akortunov:
Both default and shader water looks are "right". The only known case when it does not look "right" is the water shader incompatibility with some AMD GPUs.

PR #2521, opened 2019-09-12
Support target color in NiMaterialColorController (bug #5159)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5159)

Added the support for the target color defined in material controller flags, mimics what nifskope does. Seems to work like it should now.

### DIFF
--- a/components/nifosg/controller.cpp
+++ b/components/nifosg/controller.cpp
@@ -385,8 +385,9 @@ void AlphaController::apply(osg::StateSet *stateset, osg::NodeVisitor *nv)
     }
 }
 
-MaterialColorController::MaterialColorController(const Nif::NiPosData *data)
+MaterialColorController::MaterialColorController(const Nif::NiPosData *data, TargetColor color)
     : mData(data->mKeyList, osg::Vec3f(1,1,1))
+    , mTargetColor(color)
 {
 }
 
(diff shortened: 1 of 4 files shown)


Capostrophic:
Compared NiVertexColorProperty behavior to vanilla too, it *seems* to be fine.

components/nifosg/controller.cpp
```diff
@@ -385,8 +385,9 @@ void AlphaController::apply(osg::StateSet *stateset, osg::NodeVisitor *nv)
     }
 }
 
-MaterialColorController::MaterialColorController(const Nif::NiPosData *data)
+MaterialColorController::MaterialColorController(const Nif::NiPosData *data, int color)
```
elsid: Consider use `TargetColor` for `color` type. This makes object construction more explicit. A place where object is constructed may have more context about how to convert `int` to `TargetColor`. Otherwise some explanation why type is hidden and why is any `int` value allowed should be here to avoid mistakes in the future.

components/nifosg/controller.hpp
```diff
@@ -283,9 +283,17 @@ namespace NifOsg
     {
     private:
         Vec3Interpolator mData;
+        enum TargetColor
+        {
+            Ambient,
+            Diffuse,
+            Specular,
+            Emissive
```
elsid: Assign specific values to enum elements. Now it seems like the order is not important which actually is not a case because `(matctrl->flags >> 4) & 7` gives specific values.

Capostrophic: They're implicitly initialized to 0, 1, 2, 3 in that order though, and it is a standard C++ syntax, I'm not sure how would it seem the order is not important.

elsid: It is a standard syntax, but if order doesn't matter same approach could be used. And this might be confusing.

akortunov:
A merge conflict appeared.
Also do you have some meshes to test rhis PR?

PR #2522, opened 2019-09-13
Refactor NIF extra data and particle modifier handling

Capostrophic:
Some minor changes from my NIF branch that don't need the NIF loader to care what version the file is. They should make the future NIF format improvements more simple and in line with what NifSkope and more recent Gamebryo games do.

1. Objects no longer inherit from the extra data class.
    
    Someone thought it was a good idea to make extra data record loading and the loading of the part of NiObjectNET (Named in our loader) corresponding to extra data the same. It was not.
    
    While strictly speaking fixing it is not necessary for Morrowind files to be loaded "correctly", it inhibits the loading process for different NIF formats because whatever fields actual extra data may share will be shared with the named objects, but there's actually only one field named objects need — the first extra data object in the linked list. Also obviously named objects are not extra data, but you could cast it to be one. So I decoupled Named objects and extra data objects to avoid confusion.
    
    Gamebryo models have the entire extra data list in the named object information instead of the first extra data object in the list so our current setup is not reusable unlike NifSkope's.

2. "Controlled" abstraction no longer exists.
    
    Similar issue with NiParticleModifiers. They used this abstraction to have access to the next controlled object in the particle modifier linked list. This abstraction was also used for named objects to have access to the object's controller. Which means that technically speaking named objects were descending from both extra data and particle modifiers, which is weird and *also* is badly reusable for newer NIF formats.
    
    Files that contain particle modifier definitions are still named "controlled" because I couldn't come up with a better name and controlled still makes sense, kind of.

3. Introduced NiParticleModifier/NiParticleCollider abstractions
    
    To replace Controlled, here are some better generic record types. They're not very useful right now but they'll become useful once I somehow finish something in my NIF branch and they get updated, because particle modifiers/colliders get more shared fields.

4. Extra data size reading moved into the base read() method
    
    So that duplicate code can be removed from extra data record reading and can be easily disabled for newer extra data records in future.

5. Next extra data in the linked list is now named "next" and not "extra" in the previous extra data's data.
    
    So it's more clear that's what it is.

Had to update the tests and bullet nif loader slightly to account for the changes. Controllers now point to a named object as their target, this is in line with NifTools research.


AnyOldName3:
I tweaked your indentation as how it was before as making the headings and paragraphs the wrong way round.

Capostrophic:
Hm. Looks like tests still fail. Let's see...

PR #2523, opened 2019-09-14
[WIP] Lua scripting support

akortunov:
Based on MWSE-Lua and sol2 (unfortunately, sol3 uses the C++17). To build this branch, you need the LuaJIT installed on your machine.

**If you want to help with development, feel free to create pull requests to this branch.
There is a lot of stuff from MWSE and MWScript in missing. Just stick to features which do not require a full-scale refactoring of OpenMW engine.**

**Note: a resulting implementation is supposed to replace MWScript and old MWSE and should have feature parity with MWSE-Lua. It is not guaranteed to be backward combatible with MWSE-Lua due to differences in engines implementations (especially in regards of scene graph handling, GUI and filesystem support).**

Current state:
1. Lua state itself is embedded
2. Most of ESM objects are mapped to Lua
3. Part of functions from tes3 namespace is implemented
4. Events framework is partially implemented (you can not disable events and filter them by object at the moment)
5. Some events are implemented for testing purposes
6. Reference arguments are supported, but there is no CellRef stuff implemented.
7. String variables and vectors are supported (no matrices yet)
8. Added some bindings for scene graph nodes.
9. Added ability to assign custom tables to references (without savegame format change yet).

Known design flaws:
1. No sandboxing yet
2. No VFS support (I suppose we should use it instead of direct access to filesystem)
3. Events framework exposes Lua internals to other parts of the engine. I'd prefer to have a language-independent event manager. Also GCC fails to compile approach used to disable events in MWSE (define a static bool flag in base template class, declare it in events implementations).
4. sol2 does not work properly with osg::ref_ptr yet, so I have to use homebrew smart pointers from MWSE.
5. There are functions such as the `getWorld` in MWSE, which expose large parts of engine's internals to scripting system. I suggest to do not use this approach.
6. CI does not work since we have no LuaJIt in our libraries.
7. An outdated sol version is used since we do not use the C++17.
8. It is unclear if we should give users a write access to ESM record stores or should not.

To launch the game, you need the `init.lua` script. Unpack `scripts` folder from [attached archive](https://github.com/OpenMW/openmw/files/3612668/scripts.zip) to folder with OpenMW executable.

Code samples:
**1. Attach mesh to actor's node:**
```
local ref = omw.getReference("carnius magius")
local visual = omw.loadMesh("meshes\\w\\w_mace.nif")
local attachNode = ref.sceneNode:getObjectByName("Tri Right Hand");
attachNode:attachChild(visual)

for i = 1, #attachNode.children do
    print(attachNode.children[i].name)
end
```

**2. Print message when player loads the game:**
```
local function myLoadEvent(eventData)
    if (eventData.newGame == true)
    then
        omw.messageBox("New game launched")
    elseif (eventData.quickload == true)
    then
        omw.messageBox("Quicksave is loaded")
    else
        omw.messageBox("The game is loaded: "..eventData.filename)
    end
end

event.register("loaded", myLoadEvent)
```

**3. Implement throwing spears:**
```
local throwingSpearType = {}
throwingSpearType.id = 15
throwingSpearType.shortGroup = "1h"
throwingSpearType.longGroup = "throwweapon"
throwingSpearType.sound = "Item Weapon Spear"
throwingSpearType.attachBone = "Weapon Bone"
throwingSpearType.sheathBone = "Bip01 SpearTwoWide"
throwingSpearType.skill = 23
throwingSpearType.class = 2
throwingSpearType.twoHanded = false
throwingSpearType.hasHealth = false

omw.addWeaponType(throwingSpearType)
```

**4. Vector maths:**
```
local vec = tes3vector3:new(1, 2, 3)
local vec2 = tes3vector3:new(4, 5, 6)
print(vec:dot(vec2)) -- dot product
print(vec:negate())
```

**5. Example of timers usage:**
```
local function onTimerComplete()
    print("1 simulate second has passed!")
end

timer.start({ duration = 1, callback = onTimerComplete, iterations = 10, type = timer.simulate })
```

**6. Use custom data:**
```
local player = omw.getPlayer()
player.customData = { hunger = 18.5, thirst = 12.9 }
player.customData.sleep = 22.7
player.customData = nil
```


davidcernat:
Great job on a technical level, but starting an entire new implementation unilaterally and in secrecy goes against previous agreements.

I suppose you could call it a tactical victory and a strategic failure.

davidcernat:
I'm going to have to close this until an executive decision about it is made by Zini or psi29a, seeing as that decision will decide the entire future of the project.

PR #2525, opened 2019-09-16
Improve faction items handling

akortunov:
Fixes [bug #3609](https://gitlab.com/OpenMW/openmw/issues/3609) and [bug #5164](https://gitlab.com/OpenMW/openmw/issues/5164).

Summary of changes:
1. Consider using of evidence chest to be a crime against of chest owner even if it is technically allowed (a player has a high rank in the Imperial Legion, for example).
2. Improve tooltips handling with TFH - display a faction rank namy, display a warning about evidence chests.
3. Pass object's cellref faction to the commitCrime, so we can expell player when he commits crime against faction itself, not againts faction members.

### DIFF
--- a/apps/openmw/mwclass/container.cpp
+++ b/apps/openmw/mwclass/container.cpp
@@ -275,9 +275,11 @@ namespace MWClass
         if (ptr.getCellRef().getTrap() != "")
             text += "\n#{sTrapped}";
 
-        if (MWBase::Environment::get().getWindowManager()->getFullHelp()) {
-            text += MWGui::ToolTips::getCellRefString(ptr.getCellRef());
+        if (MWBase::Environment::get().getWindowManager()->getFullHelp())
+        {   text += MWGui::ToolTips::getCellRefString(ptr.getCellRef());
             text += MWGui::ToolTips::getMiscString(ref->mBase->mScript, "Script");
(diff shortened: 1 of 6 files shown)


apps/openmw/mwclass/container.cpp
```diff
@@ -282,9 +282,11 @@ namespace MWClass
         if (ptr.getCellRef().getTrap() != "")
             text += "\n#{sTrapped}";
 
-        if (MWBase::Environment::get().getWindowManager()->getFullHelp()) {
-            text += MWGui::ToolTips::getCellRefString(ptr.getCellRef());
+        if (MWBase::Environment::get().getWindowManager()->getFullHelp())
+        {   text += MWGui::ToolTips::getCellRefString(ptr.getCellRef());
             text += MWGui::ToolTips::getMiscString(ref->mBase->mScript, "Script");
+            if (Misc::StringUtils::ciEqual(ptr.getCellRef().getRefId(), "stolen_goods"))
+                text += "\nYou can not use evidence chests";
```
Capostrophic: What's the point of this warning?

akortunov: When user uses TFH + owned crosshair, tooltip shows ranks requirements, which player may pass, while crosshair always shows that using evidence chest is not permitted. It may confuse user and considered to be a bug.

Capostrophic:
LGTM

PR #2527, opened 2019-09-17
Continue scripts execution after player's death

akortunov:
Fixes [bug #5166](https://gitlab.com/OpenMW/openmw/issues/5166).

Just treat scripts as an any other part of game simulation (such as game mechanics, physics and world update).


Capostrophic:
I have been aware of this being a thing for a while but couldn't get a verification that Morrowind for sure does in fact always run scripts. I did experiment with the same code changes, and it didn't result in any apparent issues. So merging.

PR #2528, opened 2019-09-18
Use real time to update spell effects

akortunov:
Fixes [bug #5165](https://gitlab.com/OpenMW/openmw/issues/5165)

Makes the game playable with zero timescale.
Instead of storing the spell casting timestamp, store the remaining time of active effects. I update this time once per frame during actors update.

Notes:
1. Our previous approach is to update effects when we call the `ActiveSpells::getMagicEffects()` for current game timestamp, a new approach - update effects during actors update.
In theory, there should be no difference since we advance time after we run scripts, so `GetEffects` and `GetSpellEffects` already use effects from previous frame.
2. Since old saves have no remaining duration, use the full duration as fallback. Technically, it is not correct, but allows to avoid some additional hacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
     Bug #3676: NiParticleColorModifier isn't applied properly
     Bug #4774: Guards are ignorant of an invisible player that tries to attack them
     Bug #5108: Savegame bloating due to inefficient fog textures format
+    Bug #5165: Active spells should use real time intead of timestamps
     Bug #5358: ForceGreeting always resets the dialogue window completely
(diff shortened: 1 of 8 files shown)


Capostrophic:

psi29a:
Small merge conflict, but would like to merge.

akortunov:
Needs testing since I implemented it a year ago and have no idea if it works with upstream code.

testman42:
I tested this branch and it appears to be working well.
Spell effect timing did not misbehave no matter how much I saved/loaded and how much I abuse time-related console commands.

psi29a:
Thanks @testman42 and @akortunov

PR #2532, opened 2019-09-19
Fix undefined behaviour

Assumeru:
Loading an internal TR ESP triggered an assert because mShared was empty while mStatic was around 4000 items big.

### DIFF
--- a/apps/openmw/mwworld/store.cpp
+++ b/apps/openmw/mwworld/store.cpp
@@ -1091,6 +1091,18 @@ namespace MWWorld
         return RecordId(dialogue.mId, isDeleted);
     }
 
+    template<>
+    bool Store<ESM::Dialogue>::eraseStatic(const std::string &id)
+    {
+        auto it = mStatic.find(Misc::StringUtils::lowerCase(id));
+
+        if (it != mStatic.end() && Misc::StringUtils::ciEqual(it->second.mId, id)) {
+            mStatic.erase(it);
+        }
+
+        return true;
+    }
+
 }
 
 template class MWWorld::Store<ESM::Activator>;


apps/openmw/mwworld/store.cpp
```diff
@@ -279,7 +280,7 @@ namespace MWWorld
         if (it != mStatic.end() && Misc::StringUtils::ciEqual(it->second.mId, id)) {
             // delete from the static part of mShared
             typename std::vector<T *>::iterator sharedIter = mShared.begin();
-            typename std::vector<T *>::iterator end = sharedIter + mStatic.size();
+            typename std::vector<T *>::iterator end = sharedIter + std::min(mStatic.size(), mShared.size());
```
akortunov: Looks like a cheap hack to me.

With empty `mShared` the whole `while` loop is useless, so you can wrap the whole `mShared` handling by the `mShared.size() > 0` check. 

But I am not sure if the `mShared` really should be empty here, so the issue may be somewhere else - normally we assume that the `mShared.size() >= mStatic.size()`.

Capostrophic: Yeah, it really should never happen because mShared should have the same number of elements as mStatic and mDynamic combined. I don't suppose there can be a clear test case?

Assumeru: Yeah, agreed on it being a cheap hack. (I really just wanted to check #2529 yesterday.)

Anyway, apparently dialogue records don't get inserted into mShared while loading. Which makes sense because they get copied in after and there'd be no point in inserting them the first time if they're just going to get cleared again right after.

Capostrophic:
Not sure if I like this better, still one redundant check for deleted non-dialogue records. Maybe override eraseStatic for ESM::Dialogue records?

apps/openmw/mwworld/store.cpp
```diff
@@ -294,6 +294,18 @@ namespace MWWorld
...
+    template<>
+    bool Store<ESM::Dialogue>::eraseStatic(const std::string &id)
+    {
+        auto it = mStatic.find(Misc::StringUtils::lowerCase(id));
+
+        if (it != mStatic.end() && Misc::StringUtils::ciEqual(it->second.mId, id)) {
+            mStatic.erase(it);
+        }
+
+        return true;
```
akortunov: I suppose it would be better to move this override to other dialogue overrides (near line 1050).

apps/openmw/mwworld/store.cpp
```diff
@@ -282,7 +282,7 @@ namespace MWWorld
             typename std::vector<T *>::iterator end = sharedIter + mStatic.size();
 
             while (sharedIter != mShared.end() && sharedIter != end) {
-                if((*sharedIter)->mId == idLower) {
+                if ((*sharedIter)->mId == idLower) {
```
akortunov: Please avoid to touch lines if you really to do not modify it.

akortunov:
I suppose we also may need a bugreport and a changelog entry.

apps/openmw/mwworld/store.cpp
Assumeru: Maybe. I think there's something to be said for keeping them together so you can quickly see it's being overridden though.

apps/openmw/mwworld/store.cpp
Assumeru: Hmm, weird. I blame this one on my IDE.

apps/openmw/mwworld/store.cpp
Capostrophic: Move it! Some Store things are already being overridden for dialogue, no need to be inconsistent.

akortunov:
LGTM.

PR #2535, opened 2019-09-23
Implement vanilla-style AiActivate

akortunov:
Fixes [bug #4456](https://gitlab.com/OpenMW/openmw/issues/4456).

Now AiActivate uses Morrowind's approach - just move directly to target without pathfinding and activate it every frame once it is in activation range.

Note: according to Zini and Capostrophic, we should not implement a "reset" flag from MCP, so we will have only "broken" package version.


Capostrophic:
The fact that the movement isn't reset when the actor is near an object seemed suspicious to me, but apparently the vanilla behavior is hella inconsistent in this area so I guess it's fine.

PR #2537, opened 2019-09-24
Implement vanilla-style wind speed calculations

akortunov:
Fixes [bug #4449](https://gitlab.com/OpenMW/openmw/issues/4449)

Based on  Morrowind's logic, reverse-engeneered by Hrnchamd.

Basically, instead or reading the wind speed from config file, we simulate wind speed every frame for current weather, and next one when the weather transition is active.
The resulting wind speed is a linear inrterpolation between simulated wind speed for current weather and the wind speed for the next one. If there is no next weather, just take the simulated wind speed for current weather.

Note: As a side effect, raindrops now have a variable angle. Not sure if it is a desired behaviour - Morrowind seems to do not take the wind speed in account at all for raindrops.

### DIFF
--- a/apps/openmw/mwworld/weather.cpp
+++ b/apps/openmw/mwworld/weather.cpp
@@ -538,6 +538,8 @@ WeatherManager::WeatherManager(MWRender::RenderingManager& rendering, MWWorld::E
     , mMasser("Masser")
     , mSecunda("Secunda")
     , mWindSpeed(0.f)
+    , mCurrentWindSpeed(0.f)
+    , mNextWindSpeed(0.f)
     , mIsStorm(false)
     , mPrecipitation(false)
     , mStormDirection(0,1,0)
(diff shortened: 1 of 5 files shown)


apps/openmw/mwworld/weather.cpp
```diff
@@ -1180,7 +1223,11 @@ inline void WeatherManager::calculateTransitionResult(const float factor, const
     mResult.mFogDepth = lerp(current.mFogDepth, other.mFogDepth, factor);
     mResult.mDLFogFactor = lerp(current.mDLFogFactor, other.mDLFogFactor, factor);
     mResult.mDLFogOffset = lerp(current.mDLFogOffset, other.mDLFogOffset, factor);
-    mResult.mWindSpeed = lerp(current.mWindSpeed, other.mWindSpeed, factor);
+
+    mResult.mCurrentWindSpeed = calculateWindSpeed(mCurrentWeather, mCurrentWindSpeed);
+    mResult.mNextWindSpeed = calculateWindSpeed(mNextWeather, mNextWindSpeed);
+
+    mResult.mWindSpeed = lerp(mResult.mCurrentWindSpeed, mResult.mNextWindSpeed, factor);
```
unelsson: I'm not really familiar with this part of the code, but for the fun of reviewing: Why is this not something like lerp(current.mCurrentWindSpeed, other.mNextWindSpeed)? All other member variables seem to follow the same formula, but this is different from all the others.

apps/openmw/mwworld/weather.cpp
```diff
@@ -663,6 +665,40 @@ void WeatherManager::playerTeleported(const std::string& playerRegion, bool isEx
...
+    stormDirection.normalize();
+
+    return stormDirection;
+}
+
+float WeatherManager::calculateWindSpeed(int weatherId, float currentSpeed)
+{
+    float targetSpeed = std::min(8.0f * mWeatherSettings[weatherId].mWindSpeed, 70.f);
+    if (currentSpeed == 0.f)
+        currentSpeed = targetSpeed;
```
unelsson: This is probably okay for gameplay, but the stuff done in this function isn't exactly linear interpolation. Should this particular line from "zero to targetspeed" be smoother? How about 70.f? How come that is the minimum?

Capostrophic: This is just the values from [reverse-engineered vanilla research](https://wiki.openmw.org/index.php?title=Research:Weather#Wind_speed). Any constant names will have to be pulled out of the backside.

What do you mean "smoother"? There's no gameplay-side transition from 0 to the precalculated speed.

apps/openmw/mwworld/weather.cpp
akortunov: I do not understand the question.
During weather transitions Morrowind simulates wind speed for both next weather and current weather. So we do a linear interpolation between simulated values. The `current` and `other` have only values, taken from INI file. They are not simulated values.

akortunov: In theory, it is possible to pass a previous wind speed to the `calculateResult` as argument, but it looks like a hack anyway.

unelsson: Ok. I guess that's pretty clear. I'm pretty much just being a duck here. https://en.wikipedia.org/wiki/Rubber_duck_debugging

Capostrophic:
Seems to work and look good. Since we currently don't have enough information about how rain should work exactly (and it'll have to be reimplemented anyway), variable angle shouldn't be a discrepancy critical enough to resolve immediately. I'll merge this.

PR #2538, opened 2019-09-24
Use random attack strength if there's no wind up animation (bug #5059)

Capostrophic:
Use a ~~hack~~ workaround that compares the animation time like in another similar case instead of a check for whether it's a creature or not. Seems to work well enough according to my trusty infallible debug — doesn't break normal NPC attacks, riekling attacks and fixes the NPC test case from the [report](https://gitlab.com/OpenMW/openmw/issues/5059), both anim sets.

Not sure about the double damage issue, if it was a thing it's not obvious anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,7 @@
     Bug #5050: Invalid spell effects are not handled gracefully
     Bug #5055: Mark, Recall, Intervention magic effect abilities have no effect when added and removed in the same frame
     Bug #5056: Calling Cast function on player doesn't equip the spell but casts it
+    Bug #5059: Modded animation with combined attack keys always does max damage and can double damage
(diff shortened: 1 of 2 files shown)


akortunov:
Did you check crossbows in Morrowind, when you set variable damage to them?
An another test case - 1h-weapons attack animations from Animation Compilation.

There is a possibility that there is a special case for HtH animations.

Capostrophic:
I honestly don't think there could be a special case demonstrated in the report because Morrowind just uses random attack strength for all non-player actors and it works fine, but we shouldn't need to replicate that. But I tested Animation Compilation longsword wind up behavior and it seems that the player attack strength is random there. And so it will be for us.

Capostrophic:
It appears to be random for crossbows too. PR behavior is consistent with my observations.

PR #2539, opened 2019-09-25
Try to parse strings as number literals (bug #5097)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5097)

It has been a month since the last activity on the other relevant PR, so I decided to add the hack myself.

Using a stringstream to convert the string into a number and back (for the warning message) because it doesn't care about the locale. Also warning the user that something weird could have happened due to this.

### DIFF
--- a/components/compiler/exprparser.cpp
+++ b/components/compiler/exprparser.cpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <stack>
 #include <iterator>
+#include <sstream>
 
 #include <components/misc/stringops.hpp>
 
@@ -324,6 +325,21 @@ namespace Compiler
                 mExplicit = name2;
                 return true;
             }
+
+            // This is terrible, but of course we must have this for legacy content.
+            // Convert the string to a number even if it's impossible and use it as a number literal.
+            // Can't use stof/atof or to_string out of locale concerns.
+            float number;
+            std::stringstream stream(name2);
+            stream >> number;
+            stream.str(std::string());
+            stream.clear();
+            stream << number;
+
+            pushFloatLiteral(number);
+            mTokenLoc = loc;
+            getErrorHandler().warning ("Parsing a non-variable string as a number: " + stream.str(), loc);
+            return true;
         }
         else
         {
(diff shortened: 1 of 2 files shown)


components/compiler/exprparser.cpp
```diff
@@ -324,6 +325,21 @@ namespace Compiler
...
+            float number;
+            std::stringstream stream(name2);
+            stream >> number;
+            stream.str(std::string());
+            stream.clear();
+            stream << number;
+
+            pushFloatLiteral(number);
+            mTokenLoc = loc;
+            getErrorHandler().warning ("Parsing a non-variable string as a number: " + stream.str(), loc);
```
Mingun: What will happen if string do not represent valid `float`? If there no exceptions, I think, it is better also to log parsed number, for reducing possible misunderstandings

Capostrophic: No, there are no exceptions, tge argument gets parsed as 0 because vanilla appears to use atof. Stringstream does the same thing. I already convert the parsed number back to a string to log it in the warning, it's not useful to have the warning duplicate the weird name.

elsid: So you write a number to a stream just not to use the variable with weird name? Optimizers are not so smart, it's just extra work for CPU to do.

Capostrophic: Hey, I can't just use "name" variable because the converted number is not necessarily the same thing as the string (and the point is to show the actual number), and the editor and the launcher are locale-sensitive so if to_string is used script parser will not be accurate

elsid: Ok, I thought original value should be logged. Will be original value logged anywhere else? Assume there is a recond in log "Parsing a non-variable string as a number: 0". What is next? How to react on this?

Capostrophic: The converted number is logged for the case the really used value is a mistake and wasn't intended by the script author. So the editor will log it in the script record verifier and the engine will put it into the log if script warning logging is enabled.

The original value doesn't need to be logged. The warning message will say which line and position in the line of the script is faulty.

elsid:
Is it ready to merge?

Capostrophic:
Probably.

PR #2543, opened 2019-09-29
Revert to 0.45.0 comments-in-the-middle settings parsing behavior 

Capostrophic:
Use stringstreams again to avoid issues with stoi arising when comments are used that break users' games. A possible better solution is left up to further discussion.

Note that while error logging was removed it doesn't revert the intended "fix" for issue 4768.

There are actually less dependencies now.

### DIFF
--- a/components/fallback/fallback.cpp
+++ b/components/fallback/fallback.cpp
@@ -1,8 +1,8 @@
 #include "fallback.hpp"
 
-#include <components/debug/debuglog.hpp>
+#include <sstream>
 
-#include <boost/lexical_cast.hpp>
+#include <components/debug/debuglog.hpp>
 
 namespace Fallback
 {
@@ -28,16 +28,10 @@ namespace Fallback
         const std::string& fallback = getString(fall);
         if (!fallback.empty())
         {
-            try
-            {
-                // We have to rely on Boost because std::stof from C++11 uses the current locale
-                // for separators (which is undesired) and it often silently ignores parsing errors.
-                return boost::lexical_cast<float>(fallback);
-            }
-            catch (boost::bad_lexical_cast&)
-            {
-                Log(Debug::Error) << "Error: '" << fall << "' setting value (" << fallback << ") is not a valid number, using 0 as a fallback";
-            }
+            std::stringstream stream(fallback);
+            float number = 0.f;
+            stream >> number;
+            return number;
         }
 
         return 0;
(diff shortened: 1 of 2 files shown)


AnyOldName3:
`openmw.cfg` should already work just fine with comments partway along lines. That's what all the escape hash stuff is there for. What problems were you actually seeing?

components/fallback/validate.hpp
```diff
@@ -28,6 +28,14 @@ namespace Fallback
         for (std::vector<std::string>::const_iterator it = tokens.begin(); it != tokens.end(); ++it)
         {
             std::string temp = Files::EscapeHashString::processString(*it);
```
AnyOldName3: In fact, I'm pretty sure that because `temp` is an EscapeHashString-processed string, it's impossible for it to contain a hash.

Capostrophic: Derp.

Capostrophic:
In my testing fallback settings can contain hash signs in their value and their recovery will cause parsing errors in stoi:
```
Error: 'Moons_Secunda_Size' setting value (40#tralala) is not a valid number, using 0 as a fallback
```

AnyOldName3:
Well, I guess I'm going on a foray into the bonkers shit I wrote for my first (maybe second) pull request here.

AnyOldName3:
Update: I had this backwards. My change made it so that we *couldn't* have comments start partway along a line. By default, `boost::program_options` ignores anything following the first `#` of a line, but we wanted to be able to have hashes in data directory paths and content file names as `Morrowind.exe` allowed that. My changes made it so that `#` characters would be hidden from Boost so it would keep reading the rest of the line.

I guess it might have been a better idea to make it so that we could escape hash characters if we wanted them not to be comments, and if they weren't escaped, let Boost do with them as it wanted.

Capostrophic:
The idea is to at least not make it significantly worse than 0.45.0 behavior so that config files don't break upon an upgrade to 0.46.0. Any objections?

psi29a:
@Capostrophic this look possibly dangerous... can you have a look to see if we cover the issue we solved with this commit?
https://github.com/OpenMW/openmw/commit/7c758a629363d3a7b8fcd88fd2d40ab618c2591d

Apparently we have problems with accents (french).

Capostrophic:
This doesn't allow to use comma separators in openmw.cfg (ignores the part after the comma), but the previously used boost lexical cast doesn't seem to understand a comma separator on the default locale either (on 0.45.0 it throws an exception, on current 0.46.0 it falls back to 0).

Problems with accents? Please explain.

psi29a:
Sorry, not accents but...

> std::stof can not handle comma in us localization and can not handle period in french localization, using std::replace is not a solution, going back to boost::lexical_cast

apparently we had to use boost::lexical_cast because of a few issues...

Capostrophic:
And like I said lexical cast throws an exception upon trying to parse a separator that isn't expected for the locale (comma instead of a period) which broke parts of the game and isn't very user-friendly. I then adjusted it to catch the exceptions, but then it still discarded the values that look valid but aren't (like 1.0 # This is a setting) from lexical cast point of view, then I reverted Settings Manager to sonething resembling 0.45.0 system which doesn't care about things after the line and made fallback handling similar to that.

PR #2544, opened 2019-09-30
Rework corprus implementation

akortunov:
Fixes [bug #4623](https://gitlab.com/OpenMW/openmw/issues/4623) and [bug #3714](https://gitlab.com/OpenMW/openmw/issues/3714).

Our current corprus implementation is horribly incorrect.

When Morrowind worsens corprus, it just damages attributes by positive or negative amount. This PR replicates such behaviour. 
The `mPermanentEffects` map is redundant now, and I do not use data from it.

An additional bonus - added a workaround to load saves which failed to load due to design flaw of our old corprus implementation.

Note: the "Restore attribute to 1 point below base value" seems to be a rounding error. I see no reason to replicate such behaviour and restore attributes to their base values after curing the corprus.

Known Morrowind's quirks here:
1. It is possible to create a spell or enchantment with corprus effect in Morrowind. They should work too.
2. Morrowind seems to remember how many points of each attribute it needs to restore, but it does not store such info in savegame. As a result, when there is a save-load between corprus worsening and curing, atrributes may not be restored properly.
3. Morrowind does not apply corprus worsening for NPC's.
4. When temporary corprus effect from spell expires, Morrowind removes the whole spell.
5. Worsening should use effect's magnitude, applied to target, instead of 1.
6. Worsening does not happen in Morrowind during time fast-forwarding in prison.

### DIFF
--- a/apps/openmw/mwbase/mechanicsmanager.hpp
+++ b/apps/openmw/mwbase/mechanicsmanager.hpp
@@ -280,6 +280,8 @@ namespace MWBase
             virtual float getAngleToPlayer(const MWWorld::Ptr& ptr) const  = 0;
             virtual MWMechanics::GreetingState getGreetingState(const MWWorld::Ptr& ptr) const = 0;
             virtual bool isTurningToPlayer(const MWWorld::Ptr& ptr) const = 0;
+
+            virtual void restoreStatsAfterCorprus(const MWWorld::Ptr& actor, const std::string& sourceId) = 0;
     };
 }
 
--- a/components/esm/spellstate.cpp
+++ b/components/esm/spellstate.cpp
@@ -33,23 +33,36 @@ namespace ESM
             mSpells[id] = state;
         }
 
+        // Obsolete
         while (esm.isNextSub("PERM"))
         {
             std::string spellId = esm.getHString();
-
             std::vector<PermanentSpellEffectInfo> permEffectList;
-            while (esm.isNextSub("EFID"))
+
+            while (true)
             {
+                ESM_Context restorePoint = esm.getContext();
+
+                if (!esm.isNextSub("EFID"))
+                    break;
+
                 PermanentSpellEffectInfo info;
                 esm.getHT(info.mId);
-                esm.getHNT(info.mArg, "ARG_");
-                esm.getHNT(info.mMagnitude, "MAGN");
+                if (esm.isNextSub("BASE"))
+                {
+                    esm.restoreContext(restorePoint);
+                    return;
+                }
+                else
+                    esm.getHNT(info.mArg, "ARG_");
 
+                esm.getHNT(info.mMagnitude, "MAGN");
                 permEffectList.push_back(info);
             }
             mPermanentSpellEffects[spellId] = permEffectList;
         }
 
+        // Obsolete
         while (esm.isNextSub("CORP"))
         {
             std::string id = esm.getHString();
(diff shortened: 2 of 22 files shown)


Capostrophic:
> It is unclear if we should update corprus state for NPCS.

Corprus does affect Tanusea Veloth's attributes in Morrowind, so it (probably) makes sense to update the state for NPCs too.

>  It is unclear how to handle active spells with Corprus effect (or if we should handle them at all).


>>The corprus effect is closely linked to the corprus spell. Every 24 hours after the contraction of corprus, all other effects on the same spell as corprus, that also have effect flag 0x1000 (APPLIED_ONCE) are re-applied in a special case. This increases the magnitude of the effect, while correctly recording the amount applied so it can be removed later. Without other effects on the spell, corprus does nothing.

>>On removal, the drain spells are removed normally, while another special case interdicts any removal of positive attribute effects. Instead, a fortification is converted into a restore effect of the same magnitude.

Doesn't say anything about special cases for permanent spells.

akortunov:
> Corprus does affect Tanusea Veloth's attributes in Morrowind, so it (probably) makes sense to update the state for NPCs too.

Does "affects" mean the "corprus spell effects are applied to NPC" or "stats worsening happens every day for  NPC (even if he is in inactive cell)?

akortunov:
Since corprus can be applied via spell, active spell, enchantment or alchemy, I suppose the `CreatureStats` would be a better place to store corprus info than the `Spells`. The issue here is how to track adding/removal corprus effect in every possible place (`Spells`, `ActiveSpells` and `InventoryStore`, first two of them even do not contain reference to target actor).

akortunov:
Should be ready, but may require some refactoring.

Mainly I do not like how we handle game mechanics which should be applied when we add or remove magic effects (in the `calculateCreatureStatModifiers`), but it is outside of scope of this PR.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -1336,8 +1336,19 @@ namespace MWMechanics
             actor.getClass().getCreatureStats(actor).getSpells().purgeBlightDisease();
             break;
         case ESM::MagicEffect::CureCorprusDisease:
+        {
+            for (auto corprusSpell : actor.getClass().getCreatureStats(actor).getCorprusSpells())
```
Capostrophic: This will make a copy of the pair, should be `auto&`.

psi29a:
@akortunov do you have a follow up branch for calculateCreatureStatModifiers refactoring or at least an issue we can track for that work?

For the rest, it looks alright but needs some conflict resolution.

akortunov:
> do you have a follow up branch for calculateCreatureStatModifiers

Which kind of branch? With floating points stats? I did not include it here.

psi29a:
You said it would be the topic for another PR, do we have it on the issue tracker  or will it just be part of one of the two listed issues/bugs?

akortunov:
> You said it would be the topic for another PR, do we have it on the issue tracker or will it just be part of one of the two listed issues/bugs?

Techinically, it is not a bug, it is a code design, is not really related to corprus. We have no "effect expired" event, so we have track needed effects separately every frame (bound items, summons, corprus, etc.). I do not know if want to re-design this sytem before 1.0.

psi29a:
I don't think that's necessary either, but having the issue tracked is important. Not all things on the issue tracker are bugs, they can also be features, tasks, to-dos, follow-up work...

Capostrophic:

psi29a:
https://forum.openmw.org/viewtopic.php?f=42&t=6373

psi29a:
Merge conflicts then should be good to go.

akortunov:
A very old pull request, requires review and testing.

apps/openmw/mwbase/mechanicsmanager.hpp
```diff
@@ -17,6 +17,7 @@ namespace osg
 namespace ESM
 {
     struct Class;
+    struct Spell;
```
elsid, line 20: This forward declaration is not used.

components/esm/spellstate.cpp
```diff
@@ -33,23 +33,36 @@ namespace ESM
             mSpells[id] = state;
         }
 
+        // Obsolete
```
elsid, line 36: This is confusing. Reader may have impression that this code can be removed. If we don't plan to remove save compatibility this code is not obsolete. I hope we don't. I'd better to rewrite it in the obvious way that it just discards the result with a comment that this should not be removed until we keep compatibility with saves from <= 0.46.

akortunov: I just took the approach which we use in many other places. There is a lot of such "obsolete", "unused" and "deprecated" stuff, and it is not really corprus-specific.

> I hope we don't. 

I suppose we eventually will have to, and it is better to do it prior to 1.0 release.

psi29a:
Going to pull this is and have it go through a few nightlies for further testing.

Holding back no other save-game format changes until we can verify that this is alright.

PR #2545, opened 2019-10-02
Make map cameras and pending images update run once per frame again

Capostrophic:
Apparently this bzzt's change caused issues, probably like the issue in [this](https://gitlab.com/OpenMW/openmw/issues/5176) report.

cleanupCameras is supposed to be run once per frame (note how it gives time to draw thread to run and now the last few visited cells, the same number as there were frames, get dropped upon saving and loading). Due to the changes it was run once per cell exploration so pending images and cameras weren't updated properly, causing the last few explored cells not getting added into the overlay image and saved. It should probably be safe to just clean up cameras and not revert the commit completely, although cleanupCameras now gets called after cell exploration event fires, so I did that.

### DIFF
--- a/apps/openmw/mwgui/mapwindow.cpp
+++ b/apps/openmw/mwgui/mapwindow.cpp
@@ -856,13 +856,18 @@ namespace MWGui
 
     void MapWindow::cellExplored(int x, int y)
     {
-        mGlobalMapRender->cleanupCameras();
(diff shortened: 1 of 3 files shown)


AnyOldName3:
Which bzzt changes did we merge that caused this?

Capostrophic:
The first commit from #2279.

PR #2547, opened 2019-10-05
More minor NIF improvements

Capostrophic:
Some more small NIF changes ported from my NIF branch that apply to Morrowind version of the format.

1. Make sure UV set numbers and texture clamp modes are always unsigned values.
    This allows to get rid of some unnecessary casts to signed/unsigned integers.
2. Get rid of filter mode from textures.
    It corresponds to trilinear/bilinear/nearest neighbor/whatever filtering. But I assume we'll always use global filtering settings because doing otherwise is not very sensible except for very specific situations. Also it was unused.
3. Turn NiTexturingProperty texture array into a vector.
    This will be necessary in the future and possibly even now. The list of textures must not have a hardcoded size, so now the number of loaded textures depends on the declared number of textures in the property. Any textures beyond the normal 7 won't be handled atm. Morrowind seems to support more than one decal map but meshes with more than one decal texture simply won't be loaded atm, this remedies that slightly.
4. Simplify NiTriStripsData loading.
    It was a bit too complex for no reason.
5. More signedness consistency in texture-related locations.
6. Convert C-style typedefs to `using` directives to improve readability.

### DIFF
--- a/components/nif/controller.hpp
+++ b/components/nif/controller.hpp
@@ -112,7 +112,7 @@ class NiUVController : public Controller
 {
 public:
     NiUVDataPtr data;
-    int uvSet;
+    unsigned int uvSet;
 
     void read(NIFStream *nif);
     void post(NIFFile *nif);
--- a/components/nif/data.cpp
+++ b/components/nif/data.cpp
@@ -102,12 +102,12 @@ void NiTriStripsData::read(NIFStream *nif)
     std::vector<unsigned short> lengths;
     nif->getUShorts(lengths, numStrips);
 
+    if (!numStrips)
+        return;
+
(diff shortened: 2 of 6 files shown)


AnyOldName3:
Are there definitely no cases where something wants to override the global filtering settings e.g. a mod that adds Minecraft blocks, and so wants nearest neighbour filtering?

Capostrophic:
Well nobody complained about that in ten years there was no filter mode handling, and to the contrary I've observed a lot of situations where the default mode not being trilinear interpolation was undesired behavior. If someone ever complains we can just actually add it.

psi29a:
lgtm, taking it.

PR #2548, opened 2019-10-05
Improve items ownership handling

akortunov:
Yet another attempt to fix the [bug #1933](https://gitlab.com/OpenMW/openmw/issues/1933).

I still did not find any case where we actually use an Owner field for items in container store or when a container can have items with different owners.
So a suggested solution:
1. Do not take in account an Owner field when we check if items can stack.
2. Reset an item's owner every time when we move the item from one item model to another. It also allows to simplify an item copying API.

Also print an owner faction's name instead of ID when using TFH (if the name is empty, use the faction ID as fallback).

Notes:
1. If a trader has some sources of item with given ID and you buy some items, there is no guarantee that you will buy restocking ones first. Morrowind has this issue too. In theory, we may somehow to sort items from the `getItemsOwnedBy()` to move restocking items with given ID to beginning of list, but it is a non-vanilla improvent an outside of scope of this PR.
2. In theory, there can be cases when traders from old saves can still have some stacks of the same item.

### DIFF
--- a/apps/openmw/mwgui/tooltips.cpp
+++ b/apps/openmw/mwgui/tooltips.cpp
@@ -652,19 +652,22 @@ namespace MWGui
         std::string ret;
         ret += getMiscString(cellref.getOwner(), "Owner");
         const std::string factionId = cellref.getFaction();
-        ret += getMiscString(factionId, "Faction");
-        if (!factionId.empty() && cellref.getFactionRank() >= 0)
+        if (!factionId.empty())
         {
             const MWWorld::ESMStore &store = MWBase::Environment::get().getWorld()->getStore();
             const ESM::Faction *fact = store.get<ESM::Faction>().search(factionId);
             if (fact != nullptr)
             {
-                int rank = cellref.getFactionRank();
-                const std::string rankName = fact->mRanks[rank];
-                if (rankName.empty())
-                    ret += getValueString(cellref.getFactionRank(), "Rank");
-                else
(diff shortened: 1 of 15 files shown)


apps/openmw/mwgui/tooltips.cpp
```diff
@@ -652,19 +652,22 @@ namespace MWGui
...
             const ESM::Faction *fact = store.get<ESM::Faction>().search(factionId);
             if (fact != nullptr)
             {
-                int rank = cellref.getFactionRank();
-                const std::string rankName = fact->mRanks[rank];
-                if (rankName.empty())
-                    ret += getValueString(cellref.getFactionRank(), "Rank");
-                else
-                    ret += getMiscString(rankName, "Rank");
+                ret += getMiscString(fact->mName, "Owner Faction");
```
Capostrophic: The faction may not have a name. Then the ID should be used instead.

I think that the fact that the faction owns the item is implicit? Maybe leave the "Faction" caption alone?

Capostrophic:
Last call for feedback otherwise.

apps/openmw/mwgui/tooltips.cpp
akortunov: > I think that the fact that the faction owns the item is implicit? Maybe leave the "Faction" caption alone?

I have no idea, I just replicated text which Morrowind uses.

Capostrophic:
Guess no objections. Merging.

PR #2549, opened 2019-10-06
OpenMW-CS: Fix minimap corruption

unelsson:
fix bug [5177](https://gitlab.com/OpenMW/openmw/issues/5177)

mWnam wasn't copied in custom copy constructor


components/esm/loadland.cpp
```diff
@@ -340,6 +342,10 @@ namespace ESM
         std::swap (mContext, land.mContext);
         std::swap (mDataTypes, land.mDataTypes);
         std::swap (mLandData, land.mLandData);
+        for (int i = 0; i < LAND_GLOBAL_MAP_LOD_SIZE; ++i)
+        {
+            std::swap (mWnam[i], land.mWnam[i]);
```
Capostrophic: Equally sized arrays can be swapped without a loop going through all of their elements.

unelsson: I thought it was only std::arrays, so I used this method during bug-hunt to keep track of values. Anyway, fixed in 2fdaacf . Also tested that it really does work.

Capostrophic:
Update the changelog pls.

unelsson:
> Update the changelog pls.

Yeaah.. Done. And conflict.

PR #2550, opened 2019-10-07
Fix rebuild path for walking actors (Bug #5181)

elsid:
Fixes bug: https://gitlab.com/OpenMW/openmw/issues/5181

1. Ignore z coordinate for not swimming nor flying actors to calculate distance from actor destination to last path point. If walking actor destination point is floating above the ground then a point on navmesh may be too far away when z coordinate is included. In this case path will be rebuild on each `AI_REACTION_TIME`.
2. Increase default `recast scale factor`. To make `4` voxels fit into `sStepSizeUp = 34`: `"recast scale factor" = 4 * "cell size" / sStepSizeUp`

### DIFF
--- a/apps/openmw/mwmechanics/aipackage.cpp
+++ b/apps/openmw/mwmechanics/aipackage.cpp
@@ -22,6 +22,27 @@
 
 #include <osg/Quat>
 
+namespace MWMechanics
+{
+    static float distance(const osg::Vec2f& lhs, const osg::Vec2f& rhs)
+    {
+        return (lhs - rhs).length2();
+    }
+
+    static float distanceIgnoreZ(const osg::Vec3f& lhs, const osg::Vec3f& rhs)
+    {
+        return distance(osg::Vec2f(lhs.x(), lhs.y()), osg::Vec2f(rhs.x(), rhs.y()));
+    }
+
+    static float distance(const osg::Vec3f& lhs, const osg::Vec3f& rhs, const MWWorld::Ptr& actor)
+    {
+        const auto world = MWBase::Environment::get().getWorld();
+        if (world->isSwimming(actor) || world->isFlying(actor))
+            return distance(lhs, rhs);
+        return distanceIgnoreZ(lhs, rhs);
+    }
+}
+
 MWMechanics::AiPackage::~AiPackage() {}
 
 MWMechanics::AiPackage::AiPackage() :
@@ -133,7 +154,7 @@ bool MWMechanics::AiPackage::pathTo(const MWWorld::Ptr& actor, const osg::Vec3f&
 
         if (!mIsShortcutting)
         {
-            if (wasShortcutting || doesPathNeedRecalc(dest, actor.getCell())) // if need to rebuild path
+            if (wasShortcutting || doesPathNeedRecalc(dest, actor)) // if need to rebuild path
             {
                 const auto pathfindingHalfExtents = world->getPathfindingHalfExtents(actor);
                 mPathFinder.buildPath(actor, position, dest, actor.getCell(), getPathGridGraph(actor.getCell()),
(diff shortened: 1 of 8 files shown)


files/settings-default.cfg
```diff
@@ -642,8 +642,8 @@ enable = true
 # Scale of NavMesh coordinates to world coordinates (value > 0.0). Recastnavigation builds voxels for world geometry.
 # Basically voxel size is 1 / "cell size". To reduce amount of voxels we apply scale factor, to make voxel size
 # "recast scale factor" / "cell size". Default value calculates by this equation:
-# sStepSizeUp * "recast scale factor" / "cell size" = 3 (max climb height should be equal to 3 voxels)
-recast scale factor = 0.017647058823529415
+# sStepSizeUp * "recast scale factor" / "cell size" = 4 (max climb height should be equal to 4 voxels)
+recast scale factor = 0.023529411764705882
```
akortunov: I suppose the ``"recast scale factor" = (numVoxels * cellSize) / sStepSizeUp``.

Why just do not store a count of voxels as an integer setting instead of irrational number?

elsid: It's possible to have non integer number of voxels per `sStepSizeUp`. So far my assumption is that it's better to have this value aligned to `sStepSizeUp` because it is better with slope edge cases. I haven't done much testing for this. Basically this affects slopes and roofed locations.

In theory this is a binary classification in therms of machine learning. By changing the value we have following situations:
* Passable by physics and passable by navmesh (true positive)
* Passable by physics and impassable by navmesh (false negative)
* Impassable by physics and passable by navmesh (false positive)
* Impassable by physics and impassable by navmesh (true negative)

To get optimal value we need to run optimization over "all" possible slopes and roofed locations maximize [F-score](https://en.wikipedia.org/wiki/F1_score). Also we need to consider that by increasing `recast scale factor` real size of each tile decreases. Having limited amout of tiles this decreases area covered by navmesh. That's where [64bit navmesh](https://github.com/OpenMW/openmw/pull/2229) may give some benefit. It'd be much of fun to setup such an environment to make this optimization possible in openmw.

akortunov: > It's possible to have non integer number of voxels per sStepSizeUp.

Even in this case it is simplier so store it directly as a setting, rather than storing an arbitrary irrational multiplier to calculate the same value.

elsid: This value is not calculated anywhere except probably recastnavigation internals, search for mRecastScaleFactor to see.

elsid: Voxel size is a constant = 1.

apps/openmw/mwmechanics/aipackage.cpp
```diff
@@ -22,6 +22,27 @@
...
+        return distance(osg::Vec2f(lhs.x(), lhs.y()), osg::Vec2f(rhs.x(), rhs.y()));
+    }
+
+    static float distance(const osg::Vec3f& lhs, const osg::Vec3f& rhs, const MWWorld::Ptr& actor)
+    {
+        const auto world = MWBase::Environment::get().getWorld();
+        if (world->isSwimming(actor) || world->isFlying(actor))
+            return distance(lhs, rhs);
+        return distanceIgnoreZ(lhs, rhs);
+    }
```
Capostrophic: Maybe move these into pathfinding.hpp?

elsid: I thought about this. Wasn't sure what's better. Decided to have this code where it's only required. Maybe we should have some separate header for this stuff and other one about distances.

Capostrophic: We may need these overloads for something else, keeping them in one place would be more consistent.

e.g. the overload that takes a Ptr can be used in getFightDistanceBias().

elsid: Done

apps/openmw/mwmechanics/combat.hpp
```diff
@@ -55,6 +55,9 @@ void applyFatigueLoss(const MWWorld::Ptr& attacker, const MWWorld::Ptr& weapon,
 float getFightDistanceBias(const MWWorld::Ptr& actor1, const MWWorld::Ptr& actor2);
 
 bool isTargetMagicallyHidden(const MWWorld::Ptr& target);
+
+float getAggroDistance(const MWWorld::Ptr& actor, const osg::Vec3f& lhs, const osg::Vec3f& rhs);
```
Capostrophic: Redundant.

elsid: No.

What is redundant? Function, declaration or something else? Why? What do you want instead?

Capostrophic: Sorry, got confused in the intention behind the function. Maybe add a comment which describes its absolute difference from getPathDistance (it's easy to mix them up if you're not careful otherwise).

elsid:
It seems this fix is not so simple. If actor moves by z (e.g. teleports between floors) then attacking actors are confused. They don't change destination until attacked actor moves by other coordinates.

elsid:
e4d0af6a6de7d30cf8cd687ab808f4d5e5f7e20d fixes the problem based on assumption that it's not possible for any actor to be in the location with roof elevation lower than the actor height.

akortunov:
A yet another issue:
Screenshot_20191127_123304

PR #2552, opened 2019-10-10
Split greetings from AiWander

akortunov:
Should fix [bug #4594](https://gitlab.com/OpenMW/openmw/issues/4594).

Now actors with AiWander, AiTravel and without AI packages can greet player, as in Morrowind.

Notes:
1. In Morrowind when an actor with AiTraver greets player, he stops to move to its target. The only way to move him to his destination is the fast-forwarding during rest. I did not replicate such behaviour, and actors with AiTravel resume their walking after greeting.
2. Maybe we have better ways to determine a current AI package.
3. Morrowind seems to always play an endless "idle2" animation when greets player, even if actor has no AI packages, while in upstream master we use the "idle" animation during greeting. I replicated it.
4. Swimming actors in Morrowind do not greet player, even if their heads are above water level. I replicated it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
     Bug #4411: Reloading a saved game while falling prevents damage in some cases
     Bug #4456: AiActivate should not be cancelled after target activation
     Bug #4540: Rain delay when exiting water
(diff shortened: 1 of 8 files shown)


Capostrophic:
LGTM

PR #2555, opened 2019-10-10
Encode ID's in all places to UTF-8

akortunov:
Fixes [bug #3977](https://gitlab.com/OpenMW/openmw/issues/3977).

I did not encounter noticable slowdown during game loading.

Also improves situation with [#4598](https://gitlab.com/OpenMW/openmw/issues/4598) - escaping by `"` characters now should work (but MWScript parser still can not handle multibyte characters in general case properly, though).

### DIFF
--- a/components/esm/loadregn.cpp
+++ b/components/esm/loadregn.cpp
@@ -62,8 +62,10 @@ namespace ESM
                     break;
                 case ESM::FourCC<'S','N','A','M'>::value:
                 {
+                    esm.getSubHeader();
                     SoundRef sr;
-                    esm.getHT(sr, 33);
+                    sr.mSound.assign(esm.getString(32));
+                    esm.getT(sr.mChance);
                     mSoundList.push_back(sr);
                     break;
                 }
@@ -103,7 +105,10 @@ namespace ESM
         esm.writeHNT("CNAM", mMapColor);
         for (std::vector<SoundRef>::const_iterator it = mSoundList.begin(); it != mSoundList.end(); ++it)
         {
-            esm.writeHNT<SoundRef>("SNAM", *it);
+            esm.startSubRecord("SNAM");
+            esm.writeFixedSizeString(it->mSound, 32);
+            esm.writeT(it->mChance);
+            esm.endRecord("NPCO");
         }
     }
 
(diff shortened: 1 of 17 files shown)


Capostrophic:
>I did not encounter noticable slowdown during game loading.

IIRC the benefit of NAME32s is in decreasing the memory usage (because multibyte characters aren't used I guess). Waiting for others' input, I remember that NAME32 vs. normal strings discussion was a sensitive topic somehow.

akortunov:
> Waiting for others' input, I remember that NAME32 vs. normal strings discussion was a sensitive topic somehow.

We have no choice anyway if we still want to use UTF-8. Otherwise we will have to use legacy encodings instead.

Thunderforge:
I'm very much in favor of using UTF-8. It guarantees cross-platform compatibility (especially with Linux, macOS, and Android), and allows the engine to reach an international audience by supporting characters in non-English languages. And if we want OpenMW to be a general purpose engine, I think that supporting UTF-8 is essential.

psi29a:
Will this break the save game format?

akortunov:
> Will this break the save game format?

1. ESM still will use legacy encodings. The only difference is how we will treat identifiers during runtime.
2. We do not support mods which use non-ASCII identifiers ATM anyway.

So in theory we do not need to increase a savegame version.

psi29a:
2. is interesting... I like this, just not for this PR.

components/esm/loadregn.cpp
```diff
@@ -103,7 +105,10 @@ namespace ESM
         esm.writeHNT("CNAM", mMapColor);
         for (std::vector<SoundRef>::const_iterator it = mSoundList.begin(); it != mSoundList.end(); ++it)
         {
-            esm.writeHNT<SoundRef>("SNAM", *it);
+            esm.startSubRecord("SNAM");
+            esm.writeFixedSizeString(it->mSound, 32);
+            esm.writeT(it->mChance);
+            esm.endRecord("NPCO");
```
elsid, line 111: Why `SNAM` record ends as `NPC0`? endRecord has assert to check whether argument matches last record and it fails now in `esmtool clone` for Morrowind.esm.

Capostrophic: Copypaste mistakes do this to you.

PR #2557, opened 2019-10-12
Add Visual Studio 2019 support

Naratzul:
Updated before_script.msvc.sh to support Visual Studio 2019.
I thing it's also worth updating [setup environment wiki page](https://wiki.openmw.org/index.php?title=Development_Environment_Setup).


psi29a:
Thanks, I'll take it.

PR #2558, opened 2019-10-12
Rework 'prevent merchant equipping' feature

akortunov:
Summary of changes:
1. Restore functionality of feature after we decided to do not store owners for items in container stores. Now we really check if a target actor is a trader, if yes - allow to auto equip items only once, during initial autoequipping.
2. This settings now does not break autoequipping for player (e.g. when changing race or class via console) - we did not store owners for items in player's inventory anyway, so there were no items which player could autoequip.
3. Now we remove bound items after actor's death and re-equip weapons when spell expires (traders do not re-equip weapons when the setting is true).

Not sure if we need separate changelog entries since most of this PR fixes regressions.

Note: I had to disable light equipping for traders when the setting is active since in this case autoequipping is disabled for them.


akortunov:
An another thing: maybe it is worth to forbid traders to use light sources indifferently from setting value.

PR #2559, opened 2019-10-12
Dehardcode rain settings

akortunov:
Fixes [bug #4262](https://gitlab.com/OpenMW/openmw/issues/4262) by using related values from openmw.cfg.

### DIFF
--- a/apps/openmw/mwrender/sky.hpp
+++ b/apps/openmw/mwrender/sky.hpp
@@ -9,6 +9,8 @@
 #include <osg/Vec4f>
 #include <osg/Uniform>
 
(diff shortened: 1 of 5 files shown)


apps/openmw/mwrender/sky.hpp
```diff
@@ -9,6 +9,8 @@
 #include <osg/Vec4f>
 #include <osg/Uniform>
 
+#include <osgParticle/BoxPlacer>
```
Capostrophic: It shouldn't be necessary to include it here exactly, just forward-declare BoxPlacer and revert dependency changes in sky.cpp.

PR #2561, opened 2019-10-18
Moving OpenMW into the future with c++17 std::filesystem

psi29a:
Let's see what XCode10 has to say about c++17


akortunov:
> /Users/travis/build/OpenMW/openmw/apps/bsatool/bsatool.cpp:220:22: error: 'path' is unavailable: introduced in macOS 10.15

psi29a:
Yes, I know... I posted about this already in the forums. It has to do with the fact that they provide 10.15 SDK, but only on 10.14 images. Meaning they have headers there for new stuff, but the libstdc++ is still 10.14... we have no choice but to wait.

xyzz:
For the record, `std::filesystem` doesn't appear to be officially supported in Android NDK at the moment:
- https://github.com/android/ndk/issues/609
- https://android.googlesource.com/platform/ndk/+/master/docs/Roadmap.md#c_file-system-api

psi29a:
Thanks. Do you have an ETA for when that would be available?

akortunov:
It is unlikely that we will migrate to C++17 soon, so I see no point to keep this small testing PR opened.

We may re-open it once C++17 will be fully supported on MacOS X and Android.

PR #2562, opened 2019-10-19
Weather-related fixes (incl. bug #4783)

Capostrophic:
1. Simplify some calculations
    Some calculations added by akortunov to conform to the vanilla research were extraneous.
    In storm direction calculations, storm direction is a 1-unit length vector, so multiplied by wind speed its length would just be wind speed.  I removed that part from wind speed calculations.
   In rain angle calculation, both atan2 arguments are positive (mWindSpeed is non-negative), so it is possible to use atan and a single argument. This also allows to avoid a division by zero if the rain weather Morrowind.ini wind speed is somehow 0.

2. Fix Blizzard weather direction ([bug 4783](https://gitlab.com/OpenMW/openmw/issues/4783))
   Blizzard is no longer considered a glorified ash storm and moves from south to north instead of Red Mountain to player. Blizzard model is now rotated by 90 degrees clockwise like in Morrowind to look correct. The clouds rotation still uses the normal rotation quaternion to look as intended. Practical storm direction for hand shielding purposes is still correct. I also moved the whole storm direction calculations into weather update because it actually only calculated ash storm direction and not storm direction.

3. Fix clouds direction
  Now they should consistently move from south to north or follow the storm particle direction. Their direction could be reverse in some situations.

I decided to use model checks instead of weather ID checks to avoid confusion between different storm weathers during transitions and to make it easier to dehardcode clearly Morrowind-specific ash/blight weathers in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
     Bug #4768: Fallback numerical value recovery chokes on invalid arguments
     Bug #4775: Slowfall effect resets player jumping flag
     Bug #4778: Interiors of Illusion puzzle in Sotha Sil Expanded mod is broken
+    Bug #4783: Blizzard behavior is incorrect
     Bug #4787: Sneaking makes 1st person walking/bobbing animation super-slow
     Bug #4797: Player sneaking and running stances are not accounted for when in air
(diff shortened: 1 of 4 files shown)


psi29a:
lgtm, thanks.

PR #2567, opened 2019-10-21
add msvc2019 and ditch msvc2015

psi29a:


psi29a:
Let's flip the switch.

PR #2571, opened 2019-10-27
Reset idle animation only when play movement animation

akortunov:
Fixes [bug #5196](https://gitlab.com/OpenMW/openmw/issues/5196).

Just reset movement state, if there is no suitable animation found (including fallbacks).


Capostrophic:
LGTM

PR #2572, opened 2019-10-28
Rework prevent merchant equipping setting again

Capostrophic:
prevent merchant equipping started to work in general case after the previous rework, but there were issues with mods that triggered autoequip through scripting like Melchior's Mage Robes. MCP's fix doesn't cancel merchant autoequip in this case.

We (akortunov and myself) couldn't figure out the exact way the fix works (also, a Wine regression prevents us from checking it) but this seems to emulate it well enough according to third party testing.

TradeItemModel's item transfer, triggered when a barter deal successfully ends, now avoids triggering inventory store's autoequip while scripting doesn't. The approach isn't particularly beautiful (all item models that override copyItem had to be adjusted), but it resolves a release blocker and hopefully we can improve this once the great dehardcoding™ comes or earlier.

### DIFF
--- a/apps/openmw/mwgui/companionitemmodel.cpp
+++ b/apps/openmw/mwgui/companionitemmodel.cpp
@@ -25,12 +25,12 @@ namespace MWGui
     {
     }
 
-    MWWorld::Ptr CompanionItemModel::copyItem (const ItemStack& item, size_t count)
+    MWWorld::Ptr CompanionItemModel::copyItem (const ItemStack& item, size_t count, bool allowAutoEquip)
     {
         if (hasProfit(mActor))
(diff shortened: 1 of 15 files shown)


akortunov:
An additional note: according by SiberianCrab's testing, MCP behaves in the similar way - it prevents autoequipping only for trading window. So if you add equipment via `additem` or `equip` console command, merchants equip items from his inventory without restrictions. It is intended behaviour to prevent issues with mods such as Melchior's Mage Robes, which add items via scripts.

Capostrophic:
So, yeah, now the option should work even closer to MCP as NPCs no longer stop equipping armor and clothing they don't own like in 0.45.0 and earlier.

PR #2573, opened 2019-10-28
Attempt to fix a regression - crash on ARM

akortunov:
1. Do not use `pragma pack` for `SoundRef` struct since it now has a varying size
2. Remove outdated comments about struct size.

Requires testing on Android.

### DIFF
--- a/components/esm/loadregn.hpp
+++ b/components/esm/loadregn.hpp
@@ -33,14 +33,14 @@ struct Region
         // the engine uses mA as "snow" and mB as "blizard"
                 mA, mB;
     }; // 10 bytes
+#pragma pack(pop)
 
     // Reference to a sound that is played randomly in this region
     struct SoundRef
     {
         std::string   mSound;
         unsigned char mChance;
-    }; // 33 bytes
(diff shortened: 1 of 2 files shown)


treymerkley:
why would moving the (pop) before the SoundRef fix the regression? It doesn't look like SoundRef contains anything related to the `pragma pack()`.

Capostrophic:
I'm not sure if I understand the observation. The SoundRef struct is no longer packed (because std::strings have variable size), it was moved out from the pragma block just fine. So the alignment issue should be fixed.

treymerkley:
I'm *probably* reading it wrong. It looked like the pragma block was never actually inside of the SoundRef block, just a line below it, and was instead moved above the block. I mean, if it fixes it, it fixes it, but I wanted to make sure I was reading it right and understood what the change is

xyzz:
Everything between `#pragma pack(1)` and `#pragma pack(pop)` is packed, by moving `#pragma pack(pop)` above `struct SoundRef` it is no longer marked as packed.

PR #2574, opened 2019-10-28
Avoid copying strings in SearchVisitor and readReferenceCollection (#5195)

Capostrophic:
[Tracker page](https://gitlab.com/OpenMW/openmw/issues/5195)

As suggested by icecream95, just add a pointer variant of the cell reference getRefId() method for now and only use it in SearchVisitor and readReferenceCollection, to avoid string copy overhead. In my testing, a single record search is now at least 3 times faster, down from 15-45 microseconds to 5-10 microseconds on average. Loading a saved game of ~60 hours takes about 9% less time.

### DIFF
--- a/apps/openmw/mwworld/cellref.hpp
+++ b/apps/openmw/mwworld/cellref.hpp
@@ -34,6 +34,9 @@ namespace MWWorld
         // Id of object being referenced
         std::string getRefId() const;
 
+        // Pointer to ID of the object being referenced
+        const std::string* getRefIdPtr() const;
+
         // For doors - true if this door teleports to somewhere else, false
         // if it should open through animation.
         bool getTeleport() const;
--- a/apps/openmw/mwworld/cellref.cpp
+++ b/apps/openmw/mwworld/cellref.cpp
@@ -25,6 +25,11 @@ namespace MWWorld
         return mCellRef.mRefID;
     }
 
+    const std::string* CellRef::getRefIdPtr() const
+    {
+        return &mCellRef.mRefID;
+    }
+
     bool CellRef::getTeleport() const
     {
         return mCellRef.mTeleport;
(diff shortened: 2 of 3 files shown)


treymerkley:
Why create a new pointer constant? Why not convert getRefId() to a pointer? Wouldn't that get better performance in other areas as well?

treymerkley:
Oh, okay, makes sense. Thanks!

apps/openmw/mwworld/cellref.hpp
```diff
@@ -34,6 +34,9 @@ namespace MWWorld
         // Id of object being referenced
         std::string getRefId() const;
 
+        // Pointer to ID of the object being referenced
+        const std::string* getRefIdPtr() const;
```
elsid, line 38: Any reason why it's a pointer instead of a reference? This function can't return `nullptr`.

Capostrophic: As stated in the description that was the suggestion. It hasn't been given much more of a thought.

zinnschlag: Then it should be a reference at least. Still, pretty hacky. As a workaorund for now maybe okay. As a long term solution not so much. Scrawl and I once discussed a plan to implement something like unreal engines FName (https://docs.unrealengine.com/4.26/en-US/ProgrammingAndScripting/ProgrammingWithCPP/UnrealArchitecture/StringHandling/FName/) for IDs. I guess that never materialised?

psi29a: @zinnschlag There is a MR to optimise this: https://gitlab.com/OpenMW/openmw/-/merge_requests/1160

Moving away from this however is not in the issue tracker is it?

zinnschlag: > @zinnschlag There is a MR to optimise this: https://gitlab.com/OpenMW/openmw/-/merge_requests/1160
> 
> Moving away from this however is not in the issue tracker is it?

Not that I know of. This would require a fair bit of design work. Getting something like FName right isn't a simple task.

PR #2575, opened 2019-10-29
Support UTF-8 in script parser

akortunov:
Should fix the [bug #4598](https://gitlab.com/OpenMW/openmw/issues/4598).

A "verify" check in MTB, Myar Aranath and unpatched Arktwend ESM's did not show anything suspicious.

A separate question - do we want to use UTF-8 in scripting system at all? I suppose yes since we convert almost everything to UTF-8 during ESM loading, including scripts text.

About performance: script compilation now takes about 10-20% more time, but it is still very fast (on my hardware it takes about 0.15-0.16 sec to compile 1.5k scripts - all scripts in vanilla ESM's, on master it takes about 0.14 sec).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
     Bug #4456: AiActivate should not be cancelled after target activation
     Bug #4540: Rain delay when exiting water
     Bug #4594: Actors without AI packages don't use Hello dialogue
+    Bug #4598: Script parser does not support non-ASCII characters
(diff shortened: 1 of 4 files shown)


Capostrophic:
> do we want to use UTF-8 in scripting system at all?

I don't see why not. We're probably going to use practically the same script record format for anything that isn't mwscript, and that anything that isn't mwscript may need Unicode support, and there's no benefit in deliberately restricting mwscript to 8-bit encodings if we use Unicode for all ESM strings anyway.

PR #2578, opened 2019-10-30
Implement mouse wheel bindings (bug #2679)

akortunov:
Basically, cleaned up version of [MR97](https://gitlab.com/OpenMW/openmw/merge_requests/97) with resolved conflicts and a couple of bugfixes (Stomy is inactive ATM).

Known issue: mouse wheel bindings do not work for actions which require to hold button (e.g. movement - actionIsActive() does not work for mouse wheel movemnt bindings) - the channel value is 0, and it is unclear how to fix it. May be not a blocker, though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
     Bug #1515: Opening console masks dialogue, inventory menu
     Bug #1933: Actors can have few stocks of the same item
     Bug #2395: Duplicated plugins in the launcher when multiple data directories provide the same plugin
+    Bug #2679: Unable to map mouse wheel under control settings
     Bug #2969: Scripted items can stack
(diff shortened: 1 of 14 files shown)


psi29a:
Couldn't you have at least included Stomy's commits then your clean-up? Seems kind of harsh to not give credit where credit is due for his work.

akortunov:
> Couldn't you have at least included Stomy's commits then your clean-up?

I'd prefer to do not push them to master. There are 16 commits, which are a bit buggy and have a lot of random changes in modified files, so `git blame` will not be happy about these files.

The only thing I can do is to change author of commit from this PR.

psi29a:
That's is probably good enough, unless you want to squash his commits first, set him as the author then your own special sauce on top.

akortunov:
> That's is probably good enough

Done.

PR #2580, opened 2019-10-30
Store raw data about fog of war instead of RGBA images

akortunov:
Re-incarnation of #2453.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,7 @@
     Bug #5104: Black Dart's enchantment doesn't trigger at low Enchant levels
     Bug #5105: NPCs start combat with werewolves from any distance
     Bug #5106: Still can jump even when encumbered
+    Bug #5108: Savegame bloating due to inefficient fog textures format
     Bug #5110: ModRegion with a redundant numerical argument breaks script execution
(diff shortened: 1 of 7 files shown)


akortunov:
Just for reference - PNG with its palettes gives a better data compression (18.5% save instead of 15% with raw write). Not sure about performance, though.

Julix91:
The "on hold" going away - does that mean work on 0.47 is starting?

Managed to confirm 0.46 is in RC:

> [...] branched openmw-46... let the RC phase begin.

https://forum.openmw.org/viewtopic.php?f=20&t=5548&start=80#p67038
https://github.com/OpenMW/openmw/tree/openmw-46

In an old issue I see a reference to 0.47
https://github.com/OpenMW/openmw/pull/2453#issuecomment-547917228
but clicking on the link https://forum.openmw.org/viewtopic.php?f=42&t=6373 it says I don't have permission to view.

Is there a thread about scope etc for 0.47 that's publicly (normal reigstered user) visible?

Capostrophic:
There's no real scope for 0.47 in general.

psi29a:
Went with PNG variant, thanks @akortunov

PR #2581, opened 2019-10-31
Implement actors movement deceleration (feature #4544)

akortunov:
Implements [feature #4544](https://gitlab.com/OpenMW/openmw/issues/4544) by using related formula from Morrowind.

The formula is a bit strange, though (deceleration is a very fast, so on ranges < 450 movement speed almost always has the 0.7x multiplier).

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -144,6 +144,7 @@ namespace MWMechanics
 {
     static const int GREETING_SHOULD_START = 4; //how many updates should pass before NPC can greet player
     static const int GREETING_SHOULD_END = 10;
+    static const float DECELERATE_DISTANCE = 512.f;
 
     class GetStuntedMagickaDuration : public MWMechanics::EffectSourceVisitor
     {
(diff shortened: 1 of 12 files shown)


apps/openmw/mwmechanics/actors.cpp
```diff
@@ -422,6 +423,26 @@ namespace MWMechanics
...
+        float newSpeedFactor = 1.f;
+
+        CreatureStats &stats = actor.getClass().getCreatureStats(actor);
+        MWMechanics::AiSequence& seq = stats.getAiSequence();
+
+        if (!seq.isEmpty() && seq.getActivePackage()->useVariableSpeed())
+        {
+            osg::Vec3f targetPos = seq.getActivePackage()->getDestination();
+            osg::Vec3f actorPos = actor.getRefData().getPosition().asVec3();
+            float distance = (targetPos - actorPos).length();
```
Capostrophic: Maybe compare length2() result with DECELERATE_DISTANCE * DECELERATE_DISTANCE to avoid an unnecessary square root extraction for every actor every frame.

previousSpeedFactor should never be set to a value above 1, so a std::min can be removed.

Works good otherwise, apart from a few quirks not specific to the addition.

PR #2583, opened 2019-10-31
Vanilla-style rain

akortunov:
An attempt to implement a vanilla-style rain using a reverse-engeneered logic from Morrowind.
Should fix the [bug #3404](https://gitlab.com/OpenMW/openmw/issues/3404).

Unfortunately, a particles-based rain works faster than a drawables-based approach.
So we basically should decide - either we keep a backward compatibility here and sacrifice performance, or we claim our current approach to be a more advanced and close mentioned bug as WON'T FIX.

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -615,6 +615,8 @@ namespace MWBase
 
             /// Return physical half extents of the given actor to be used in pathfinding
             virtual osg::Vec3f getPathfindingHalfExtents(const MWWorld::ConstPtr& actor) const = 0;
+
+            virtual void emitWaterRipple(const osg::Vec3f& pos) = 0;
     };
 }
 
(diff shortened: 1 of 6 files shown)


psi29a:
This is one of those things... do we want to have 1:1 Morrowind or do we want a more modern Morrowind?

Are there mods that change the rain that would take advantage of this?

Is it done similar with Oblivion, Skyrim etc?

Pop000100:
Is the current implementation at default settings visually identical to Morrowind and will Particle based rain make no negative difference to modding(preventing more complex modifications).

If both of those statements are true then I'd say the current method is better.

Capostrophic:
Neither are true.

Pop000100:
Then how much slower is the mesh based way of doing things and will it make things like rain occlusion more complicated?

akortunov:
> Then how much slower is the mesh based way of doing things and will it make things like rain occlusion more complicated?

It is unknown. That's why this PR is marked as NEEDS TESTING.

akortunov:

PR #2584, opened 2019-10-31
Resolving 0.46.0 release shadow-related blockers

Capostrophic:
As requested by AnyOldName3, I've rewritten bzzt's shadow fading and max distance settings. I've documented them, adjusted the default values, reverted back to "dumb" shadow distance calculations out of performance concerns (there would have been very slight overhead for every pixel even if shadows aren't enabled, IMO shadow distance is very much related to fog and frustum clipping and it's not like it's that jarring) and added a global define which toggles shadow distance limit calculations on or off. Also I made some stylistic adjustments to keep them in line with the existing shadow code.
I changed the committeer to bzzt even though I didn't amend changes of his commit and rather did them "from scratch", although I suspect bzzt won't be entirely happy whether I make him the author of the commit or not.
As mentioned earlier, shadow distance limit improves the shadow quality and performance significantly if the viewing distance is very large at the cost of removing shadows from distant objects, while shadow fading just makes the transition from "shadows" to "no shadows" smooth.
Seems to work well enough in my testing. Here's how shadows look like with the settings deliberately exaggerated (max distance set to 1024, fade start set to 0.5, i.e. shadows start fading away at 512 units).
screenshot096

The viewing distance is 81920. You will notice that the shadows obviously look extremely fine up-close, especially with a higher shadow map resolution (here it's 2048).
screenshot097

I'm unable to add shadow settings into the game settings (they need specific runtime handling), so I made a compromise, adding them into the launcher like it was requested earlier. These are some of the most basic, non-complicated settings that are unlikely to confuse players and make them wonder where did their life go wrong if they mess something up. They have some ~~arbitrary~~ sensible default values set up and they also have tooltips. Here's how Graphics tab should look now. The two spin boxes are normally disabled unless shadow distance limit check box is enabled.

twotabs

I didn't adjust other default settings. They're probably fine as they are.

### DIFF
--- a/apps/launcher/graphicspage.cpp
+++ b/apps/launcher/graphicspage.cpp
@@ -45,6 +45,7 @@ Launcher::GraphicsPage::GraphicsPage(Files::ConfigurationManager &cfg, Settings:
     connect(standardRadioButton, SIGNAL(toggled(bool)), this, SLOT(slotStandardToggled(bool)));
     connect(screenComboBox, SIGNAL(currentIndexChanged(int)), this, SLOT(screenChanged(int)));
     connect(framerateLimitCheckBox, SIGNAL(toggled(bool)), this, SLOT(slotFramerateLimitToggled(bool)));
+    connect(shadowDistanceCheckBox, SIGNAL(toggled(bool)), this, SLOT(slotShadowDistLimitToggled(bool)));
 
 }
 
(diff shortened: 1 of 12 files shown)


Capostrophic:
Alright, here's how it should look now.

twotabs

akortunov:
LGTM.
A quick testing did not show anything suspicious.

Capostrophic:
CMAugust tried this out, seems to work OK for him too. I guess I'll add tooltips for everything anyway though.

AnyOldName3:
For future reference, I'd be happier if things like this were merged *after* I'd given them a final once-over, not before.

apps/launcher/graphicspage.cpp
```diff
@@ -185,6 +213,52 @@ void Launcher::GraphicsPage::saveSettings()
...
+        mEngineSettings.setInt("maximum shadow map distance", "Shadows", cShadowDist);
+    float cFadeStart = fadeStartSpinBox->value();
+    if (cShadowDist > 0 && mEngineSettings.getFloat("shadow fade start", "Shadows") != cFadeStart)
+        mEngineSettings.setFloat("shadow fade start", "Shadows", cFadeStart);
+
+    bool cActorShadows = actorShadowsCheckBox->checkState();
+    bool cObjectShadows = objectShadowsCheckBox->checkState();
+    bool cTerrainShadows = terrainShadowsCheckBox->checkState();
+    bool cPlayerShadows = playerShadowsCheckBox->checkState();
+    if (cActorShadows || cObjectShadows || cTerrainShadows || cPlayerShadows)
```
elsid, line 228: Not sure why it's done like this, but when all types of shadows are enabled to disable all shadows one need to uncheck all types. That's kind of annoying. Having `enable shadows` setting it would be better just to control it directly without any special logic like this.

psi29a:
Two questions:
Is this Bzzt's ShadowsBin PR?
Is this to be cherry-picked to 0.46?

PR #2585, opened 2019-11-03
Terrain selection fixes

unelsson:
Attempts to fix the bugs described here: https://gitlab.com/OpenMW/openmw/issues/5200

Problem: These fixes don't do anything to a visual glitch arising from rendering vertex selection at the edge of visible cells (aka cell selection edge). As for the global coordinates, x and y are handled in sizes of 64 x 64, whereas there is actually 65 x 65 grid when edges are taken in account. CellCoordinates calculations for global coordinates don't take in account the extra line, therefore anything at 65th row or column is seen as belonging to the next cell's vertex coordinate 0, which isn't always loaded). It's probably possible to fix this, yet terrain texture editing has a similar problem, but slightly more troublesome, as terrain textures overlap between cells and have a special offset. I intend to fix this (preferably all selection-related stuff in this PR), but in case this turns out to be more complex or time consuming, this PR may be reviewed and merged separately.

### DIFF
--- a/apps/opencs/view/render/terrainselection.cpp
+++ b/apps/opencs/view/render/terrainselection.cpp
@@ -242,10 +242,37 @@ void CSVRender::TerrainSelection::drawTextureSelection(const osg::ref_ptr<osg::V
     }
 }
 
+bool CSVRender::TerrainSelection::noCell(const std::string& cellId)
+{
+    CSMDoc::Document& document = mWorldspaceWidget->getDocument();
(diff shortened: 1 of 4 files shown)


Capostrophic:
Cosmetic changes commit doesn't really improve readability IMO...  Neither is it used in the most of the rest of the code base. But don't move return to a new line elsewhere, just drop the commit.

unelsson:
Okay. Dropped the cosmetic changes commit, and resolved problem in the first message with commit 8f6dc78 .

Capostrophic:
Seems to work good.

PR #2586, opened 2019-11-04
Allow selecting cell edges everywhere

unelsson:
Enables vertex selection at 64th row and column (at the cell edge), when the next cell isn't visible (aka in cell selection).

Explanation: Global CellCoordinates calculations don't take in account the 64th vertex, but always treat it as the 0th vertex of the next cell. Because of this, when the selection was at 64th vertex, the logic was to check if the next cell was loaded. This PR adds checks for these special cases, and therefore enables selecting 64th vertex even when the next cell isn't visible. I don't really know what's the real use case for this, but this makes vertex selection have more consistent behavior.

### DIFF
--- a/apps/opencs/view/render/terrainshapemode.cpp
+++ b/apps/opencs/view/render/terrainshapemode.cpp
@@ -1047,9 +1047,16 @@ void CSVRender::TerrainShapeMode::handleSelection(int globalSelectionX, int glob
         int moduloY = globalSelectionY % (ESM::Land::LAND_SIZE - 1);
         bool xIsAtCellBorder = moduloX == 0;
         bool yIsAtCellBorder = moduloY == 0;
-        if (isInCellSelection(globalSelectionX - 1, globalSelectionY) && xIsAtCellBorder && !yIsAtCellBorder) selections->emplace_back(globalSelectionX, globalSelectionY);
-        if (isInCellSelection(globalSelectionX, globalSelectionY - 1) && !xIsAtCellBorder && yIsAtCellBorder) selections->emplace_back(globalSelectionX, globalSelectionY);
-        if (isInCellSelection(globalSelectionX - 1, globalSelectionY - 1) && xIsAtCellBorder && yIsAtCellBorder) selections->emplace_back(globalSelectionX, globalSelectionY);
+        if (!xIsAtCellBorder && !yIsAtCellBorder)
+            return;
+        int selectionX = globalSelectionX;
+        int selectionY = globalSelectionY;
+        if (xIsAtCellBorder)
+            selectionX--;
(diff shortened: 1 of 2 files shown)


unelsson:
I wonder if these calculations should be in CellCoordinates instead of TerrainShapeMode... Maybe it's useful to know if a global selection coordinate is at the cell edge in some other cases?

apps/opencs/view/render/terrainshapemode.cpp
```diff
@@ -1038,14 +1038,29 @@ bool CSVRender::TerrainShapeMode::isInCellSelection(int globalSelectionX, int gl
...
+    if (isInCellSelection(globalSelectionX, globalSelectionY)) selections->emplace_back(globalSelectionX, globalSelectionY);
+    else
+    {
+        int moduloX = globalSelectionX % (ESM::Land::LAND_SIZE - 1);
+        int moduloY = globalSelectionY % (ESM::Land::LAND_SIZE - 1);
+        bool xIsAtCellBorder = moduloX == 0;
+        bool yIsAtCellBorder = moduloY == 0;
+        if (isInCellSelection(globalSelectionX - 1, globalSelectionY) && xIsAtCellBorder && !yIsAtCellBorder) selections->emplace_back(globalSelectionX, globalSelectionY);
+        if (isInCellSelection(globalSelectionX, globalSelectionY - 1) && !xIsAtCellBorder && yIsAtCellBorder) selections->emplace_back(globalSelectionX, globalSelectionY);
+        if (isInCellSelection(globalSelectionX - 1, globalSelectionY - 1) && xIsAtCellBorder && yIsAtCellBorder) selections->emplace_back(globalSelectionX, globalSelectionY);
```
Capostrophic: isInCellSelection is not a trivial check, yet it's done thrice for each vertex of the selection the first isInCellSelection fails for unnecessarily. Maybe this part can be rearranged as
```cpp
if (!xIsAtCellBorder && !yIsAtCellBorder)
    return;
int selectionX = globalSelectionX;
int selectionY = globalSelectionY;
if (xIsAtCellBorder)
    selectionX--;
if (yIsAtCellBorder)
    selectionY--;
if (isInCellSelection(selectionX, selectionY))
    selections->emplace_back(globalSelectionX, globalSelectionY);
```

Please avoid making lines that long, we're not at shortage of line breaks.

unelsson:
This PR still has a visual problem, where the vertex selection lines at a cell edge are drawn strangely when adjacent land is not found. I'm not sure if I should add a check for this to draw phase, because it complicates the logic, adding up extra checks for each vertex selection.

Appveyor build seems to have a problem with `Extracting archive: C:\projects\openmw\deps\Bullet-2.86-msvc2015-win64.7z`

PR #2587, opened 2019-11-04
Minor sheathing-related changes

Capostrophic:
* Add weapon and shield sheathing and use additional anim sources checkboxes into the Advanced tab of the launcher
* Remove outdated information from the documentation

The checkboxes are all in Game section of the tab for now. Maybe it'll become nicer and less cluttered later, like Graphics tab did.
Given that shield sheathing is now a thing, the relevant part of extended modding documentation became outdated. It likely needs to be actually written, but given that the documentation isn't included with OpenMW builds (and the rules on its maintenance are extremely loose anyway) it's not as important and it can be done at any time. At least it doesn't include blatantly incorrect information anymore.

### DIFF
--- a/apps/launcher/advancedpage.cpp
+++ b/apps/launcher/advancedpage.cpp
@@ -82,6 +82,9 @@ bool Launcher::AdvancedPage::loadSettings()
     loadSettingBool(requireAppropriateAmmunitionCheckBox, "only appropriate ammunition bypasses resistance", "Game");
     loadSettingBool(magicItemAnimationsCheckBox, "use magic item animations", "Game");
     loadSettingBool(normaliseRaceSpeedCheckBox, "normalise race speed", "Game");
+    loadSettingBool(weaponSheathingCheckBox, "weapon sheathing", "Game");
+    loadSettingBool(shieldSheathingCheckBox, "shield sheathing", "Game");
+    loadSettingBool(animSourcesCheckBox, "use additional anim sources", "Game");
 
     // Input Settings
     loadSettingBool(grabCursorCheckBox, "grab cursor", "Input");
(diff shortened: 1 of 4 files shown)


akortunov:
Sheathing checkboxes make no sense when the "use additional anim sources" are disabled.
Not sure what to do with it.

Capostrophic:
Updated. Went with the suggestion from the Discord server, updated the tooltips. The launcher already considers itself smarter than the user, I don't want to worsen that by toggling the checkbox automatically or something, so at least now it should be more obvious to the user why doesn't sheathing work with less effort to make the research.

akortunov:
The main issue of such tooltips is that almost no one actually reads them, especially if the checkbox name itself gives info about what this checkbox does.
User opens the launcher, sees a Weapon sheathing checkbox, checks it and lauches the game.
As a result, feature does not work for him and this user starts to complain on forums, send bugreports, etc. It is not intuitive that one checkbox has effect only when an another checkbox is set, if there is no visual indication.

A simplest solution would be to gray out sheathing checkboxes when "Use additional anim sources" is not set. We already use a similar approach for the Window border checkbox. In this case we should place animation checkbox above sheathing checkboxes, and maybe add some indent to sheathing checkboxes.

Capostrophic:
I'd argue it's not our problem that the user can't read but fine.

Capostrophic:
I'd argue it's not our problem that the user can't read but fine.

Capostrophic:
incorrect setup

akortunov:
Maybe it is worth to uncheck sheathing checkboxes when you uncheck the "Use additional animation sources" one.

Capostrophic:
What for?

akortunov:
> What for?

To disable these settings, of course. They will have no effect anyway without additional animation sources, but still may cause small performance drop due to additional checks.

Capostrophic:
There'll inevitably be the reverse complaints about how the launcher decides to reset settings for seemingly no reason then, but done.

akortunov:
LGTM.

PR #2590, opened 2019-11-09
UI textures scaling corrections

akortunov:
1. Restore normal size for Show All / Show Active buttons in journal.
2. Crop redundant texture space for them.
3. Implement scaling for quick keys widgets.

Notes:
1. Not sure if it possible to center mentioned buttons properly since on different localizations textures have different offset.
2. These buttons were tested only on English and Russian localizations.


Capostrophic:
Looks acceptable when German localization is used (GOG version), at least while the Show All/Show Active buttons (Alle and Offene buttons) are offset to the left it's exactly the same way in 0.45.0 for example.

PR #2591, opened 2019-11-09
Do not show shields in the 1st-person view when shield sheathing is enabled

akortunov:
Fixes a regression from #2569.
Since 1st-person skeleton has no sheathing bone, we disable this feature for 1st-person meshes, and show shield when weapon is sheathed, which is not right.
With this patch we still use shield sheathing for player in the 1st-person view, we just do not show holstered shield because of missing sheath bone.


Capostrophic:
Didn't test this yet, but make sure inventory doll behavior is correct.

akortunov:
> Didn't test this yet, but make sure inventory doll behavior is correct.

Inventory doll uses a 3d-person skeleton, so it has sheathing bones and it is not affected by this bug. I can add an IsInCell() check just for sure, but it should not make any difference.

PR #2592, opened 2019-11-09
Use transformations of NiSwitchNode

akortunov:
There is a bug with ex_nord_win_02 mesh from Glow in the Dahrk with OpenMW:

Screenshot_20191109_214848

It is because we ignore transformations of switch node itself.
With this patch we will create a MatrixTransform if the tranform is not identity.

I suppose it is not require a changelow entry since the NiSwitchNode support was added in 0.46.


Capostrophic:
NiLODNode is technically just a NiSwitchNode that can switch its state automatically. I suspect that it has the same transformation logic.

PR #2594, opened 2019-11-12
Minor summoned creature cleanup fixes

Capostrophic:
1. Summoned creatures of summoned creatures that recently became possible weren't properly cleared. I added a workaround for that.
2. I ignored my own good advice and forgot again that I should test the vanilla deaths using the normal damage instead of SetHealth 0. Summoned creatures are still supposed to be removed *after* the caster's death animation according to my testing. Also this fixes the issue where the despawn VFX are played twice (upon the summoned creature despawn and upon the caster's death animation end) in some situations.
3. If a summoned creature's despawn is postponed, it didn't trigger the linked spell effects removal from other actors like a normal cleanupSummonedCreature call. But the main purpose of the workaround I've added is to make sure all summoned creatures are removed even if said removal is postponed. There's a plan to expand on that, hm, later, as summoned creature mechanics are not exactly consistent yet.

### DIFF
--- a/apps/openmw/mwgui/container.cpp
+++ b/apps/openmw/mwgui/container.cpp
@@ -248,6 +248,12 @@ namespace MWGui
                         MWScript::InterpreterContext interpreterContext (&mPtr.getRefData().getLocals(), mPtr);
                         MWBase::Environment::get().getScriptManager()->run (script, interpreterContext);
                     }
+
+                    // Clean up summoned creatures as well
+                    std::map<MWMechanics::CreatureStats::SummonKey, int>& creatureMap = creatureStats.getSummonedCreatureMap();
+                    for (const auto& creature : creatureMap)
+                        MWBase::Environment::get().getMechanicsManager()->cleanupSummonedCreature(mPtr, creature.second);
+                    creatureMap.clear();
                 }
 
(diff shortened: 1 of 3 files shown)


akortunov:
> Summoned creatures are still supposed to be removed after the caster's death animation according to my testing.

What about the "safe corpse disposal"? We should remove actor's summons even if death animation was not finished yet.

akortunov:
Basically, seems to work well, but there is in another case in Morrowind when it de-spawns summoned creatures - when the target becomes out of active cell grid for some reason (e.g. it is teleported to inactive cell or just disabled via console).

A suggested approach - cleanup summons in the `Actors::removeActor()`. At least, it seems to work in this way in the Morrowind. In this case some your special handling may be not needed.

If you are not going to work on it in this PR, just fill a bugreport.

Capostrophic:
As I hinted in the first post, I don't plan to work on that here.

>Maybe we should introduce a some kind of mToDelete set in Actors, and remove actors from this vector after actors update

That was my idea too. But it's general design issue that doesn't concern summoning specifically.

akortunov:
I created a related [bugreport](https://gitlab.com/OpenMW/openmw/issues/5207) to finish the work later.

PR #2595, opened 2019-11-12
Default visual settings tweaks

Capostrophic:
I adjusted the default values of three certain settings to maintain a consistent visual quality level in 0.46.0 release:

1. composite map level was reverted from -2 to 0.
   Unfortunately negative values of the setting as they are lead to a currently unavoidable unpleasant shimmering of the terrain relatively close to the player since composite maps don't have filtering. While such values increase the performance, they worsen the look of the terrain proportionally and IMO noticeably. Maybe another time.

2. shadow front face culling is now disabled by default.
  It doesn't seem like it helps with anything in the area of shadow flickering (which is mostly absent now) but due to Morrowind's poor normals (probably?) most users would have seen pretty bad peter-panning in most situations. It probably shouldn't be removed just yet because it may improve performance in the situations where assets are more well done. So I simply disabled it.

3. tight scene bounds are computed by default.
  While it was observed in the past that while there *is* a certain performance impact when tight scene bounds are computed (5-8%), it's insignificant compared to other sources of shadows' infamous performance impact and to the quality improvement it brings in the general case (when the player looks at the ground). So I enabled it.

Both the shadow setting adjustments *do* make a noticeable difference with the default viewing distance.

### DIFF
--- a/docs/source/reference/modding/settings/shadows.rst
+++ b/docs/source/reference/modding/settings/shadows.rst
@@ -85,11 +85,11 @@ compute tight scene bounds
 
 :Type:		boolean
 :Range:		True/False
-:Default:	False
+:Default:	True
 
 With this setting enabled, attempt to better use the shadow map(s) by making them cover a smaller area.
 This can be especially helpful when looking downwards with a high viewing distance but will be less useful with the default value.
-The performance impact of this may be very large.
+May have a minor to major performance impact.
 
 shadow map resolution
 ---------------------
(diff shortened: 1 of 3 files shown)


akortunov:
> it's insignificant compared to other sources of shadows' infamous performance impact 

It depends on time of day. The performance impact actually may be a quite large at evening or morning, especially if user does not limit shadows distance.

Capostrophic:
But then there's the performance improvement when the player is looking at the ground and the water plane doesn't go into the shadow map, and the fact that shadow map distance would usually be limited.

AnyOldName3:
Have we had the thing from bzzt where the water plane doesn't ruin the scene bounds when we use the more performance-friendly bounding sphere version of scene bound computation in a mergeable form yet, or has it always been bundled with other crap? I'd be happier having the looser bounds computation that we don't offer yet on by default than having tight bounds computation on.

akortunov:
> Have we had the thing from bzzt where the water plane doesn't ruin the scene bounds

1. His patch did not improve anything.
2. I am not sure if we merged it.

Capostrophic:
We didn't. But it hasn't been supplied with a lot of unrelated changes for a while.

AnyOldName3:
I thought it *did* help, giving a performance boost over tight bounds and a quality boost over no computation, but wasn't perfect.

I'm assuming something else that's big like terrain tiles would also need switching from spheres to bounding boxes or something if the improvement was genuinely not there.

PR #2596, opened 2019-11-13
Fix some issues, found by CoverityScan

akortunov:
Should fix most of newly-detected issues. As for things left:
1. atan2 issues are false-positives.
2. Missing RefData move constructor (an old report from 2016 year) does not seem to be an issue, IMO.
3. A complaint about tainted value in OICS is not our issue too. There is just missing validation for values from XML, and I doubt that we will implement it.

2 and 3 more likely are just old discarded reports, which reappeared because we modified related files (when we implemented mouse wheel bindings, for example).

### DIFF
--- a/components/sceneutil/shadow.cpp
+++ b/components/sceneutil/shadow.cpp
@@ -66,6 +66,8 @@ namespace SceneUtil
     void ShadowManager::disableShadowsForStateSet(osg::ref_ptr<osg::StateSet> stateset)
     {
         int numberOfShadowMapsPerLight = Settings::Manager::getInt("number of shadow maps", "Shadows");
+        numberOfShadowMapsPerLight = std::max(1, std::min(numberOfShadowMapsPerLight, 8));
+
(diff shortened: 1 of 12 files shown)


psi29a:
lgtm; merging

components/sceneutil/shadow.cpp
```diff
@@ -66,6 +66,8 @@ namespace SceneUtil
     void ShadowManager::disableShadowsForStateSet(osg::ref_ptr<osg::StateSet> stateset)
     {
         int numberOfShadowMapsPerLight = Settings::Manager::getInt("number of shadow maps", "Shadows");
+        numberOfShadowMapsPerLight = std::max(1, std::min(numberOfShadowMapsPerLight, 8));
```
AnyOldName3: If we're putting a bound on this when *disabling* shadows, we should also do it when *enabling* shadows here: https://github.com/OpenMW/openmw/pull/2596/files#diff-ea4f77bf8ee802245edb11cf81aa0556R26

PR #2602, opened 2019-11-17
Fix clang-tidy issues

elsid:
Mostly related to virtual calls in constructors and destructors. Technically it's ok, but explicit calls are easier to understand. Also some virtual modifiers are not used, these are removed.

### DIFF
--- a/components/sceneutil/mwshadowtechnique.hpp
+++ b/components/sceneutil/mwshadowtechnique.hpp
@@ -261,18 +261,18 @@ namespace SceneUtil {
 
         float                                   _shadowFadeStart = 0.0;
 
-        class DebugHUD : public osg::Referenced
+        class DebugHUD final : public osg::Referenced
         {
         public:
             DebugHUD(int numberOfShadowMapsPerLight);
 
-            virtual void draw(osg::ref_ptr<osg::Texture2D> texture, unsigned int shadowMapNumber, const osg::Matrixd &matrix, osgUtil::CullVisitor& cv);
+            void draw(osg::ref_ptr<osg::Texture2D> texture, unsigned int shadowMapNumber, const osg::Matrixd &matrix, osgUtil::CullVisitor& cv);
 
-            virtual void releaseGLObjects(osg::State* state = 0) const;
+            void releaseGLObjects(osg::State* state = 0) const;
 
-            virtual void setFrustumVertices(osg::ref_ptr<osg::Vec3Array> vertices, unsigned int traversalNumber);
+            void setFrustumVertices(osg::ref_ptr<osg::Vec3Array> vertices, unsigned int traversalNumber);
         protected:
-            virtual void addAnotherShadowMap();
+            void addAnotherShadowMap();
 
             static const int sDebugTextureUnit = 0;
 
--- a/components/sceneutil/shadow.hpp
+++ b/components/sceneutil/shadow.hpp
@@ -19,15 +19,13 @@ namespace SceneUtil
 
         ShadowManager(osg::ref_ptr<osg::Group> sceneRoot, osg::ref_ptr<osg::Group> rootNode, unsigned int outdoorShadowCastingMask, unsigned int indoorShadowCastingMask, Shader::ShaderManager &shaderManager);
 
-        virtual ~ShadowManager() = default;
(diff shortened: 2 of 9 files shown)


components/sceneutil/mwshadowtechnique.hpp
```diff
@@ -261,18 +261,18 @@ namespace SceneUtil {
...
+        class DebugHUD final : public osg::Referenced
         {
         public:
             DebugHUD(int numberOfShadowMapsPerLight);
 
-            virtual void draw(osg::ref_ptr<osg::Texture2D> texture, unsigned int shadowMapNumber, const osg::Matrixd &matrix, osgUtil::CullVisitor& cv);
+            void draw(osg::ref_ptr<osg::Texture2D> texture, unsigned int shadowMapNumber, const osg::Matrixd &matrix, osgUtil::CullVisitor& cv);
 
-            virtual void releaseGLObjects(osg::State* state = 0) const;
+            void releaseGLObjects(osg::State* state = 0) const;
```
AnyOldName3: This is maybe an override. If it isn't, I maybe should have made the class extend `osg::Object` instead.

components/sceneutil/shadow.hpp
```diff
@@ -19,15 +19,13 @@ namespace SceneUtil
...
-        virtual void setupShadowSettings();
+        Shader::ShaderManager::DefineMap getShadowDefines();
 
-        virtual Shader::ShaderManager::DefineMap getShadowDefines();
+        void enableIndoorMode();
 
-        virtual void enableIndoorMode();
-
-        virtual void enableOutdoorMode();
+        void enableOutdoorMode();
```
AnyOldName3: This section of the diff is a mess which will make git blame harder to use. Is there some way you could regenerate it so that it's more obvious which lines came from where?

Also, the whole things-being-virtual approach was so that we could potentially have other shadow implementations as a drop-in replacement (e.g. the one MP3Butcher was working on and then abandoned), but I guess if that ever happens, we can just make them virtual again.

elsid: Not sure what do you mean. Blame looks fine: 
blame
Diff itself is not very good but it's because github interface does it. Git [doesn't store diffs](https://stackoverflow.com/questions/10398744/does-git-store-diff-information-in-commit-objects/10398894#10398894), it stores full files content. So how to generate diff is not in my power here.

About virtual I think it's better to add it when it's required. Also more recommended approach is to create class with pure virtual methods to define interface.

components/sceneutil/mwshadowtechnique.hpp
elsid: Deriving from `osg::Object` requires to add more stuff here that is actually  not required. Only one object of this class is created in openmw code and it is used just to call defined methods. I'm not even sure it should be derived from `osg::Referenced`. `std::unique_ptr` should be enough because it's not shared anywhere.

AnyOldName3: I didn't use `std::unique_ptr` because I've kept to OSG's code style in files taken from their repo.


While we're on the subject, IIRC there are some naked pointers nearby that should really be observer pointers, but aren't because nothing showed up when I googled for OSG weak pointer.

psi29a:
Last call?

We'll switch back to virtual if/when that needs to happen and we do indeed have a 2nd shadow implementation. I'm assuming any further work will be done improving the existing one.

components/sceneutil/mwshadowtechnique.hpp
AnyOldName3: Have the naked pointers been found and converted to observer pointers yet?

akortunov: IIRC, `osg::ref_ptr` and `std::unique_ptr` conflict with each other, so we use the former for OSG objects.

akortunov:
@AnyOldName3, do you have real complaints left?

AnyOldName3:
No, none about the actual content of the PR, and I can't actually find the naked pointers that I thought we could tidy up while in the area - all the ones that exist are from Robert, so I guess I sorted the issue out some other way.

psi29a:
Thanks @akortunov

PR #2605, opened 2019-11-19
Fix GCC8 warning in Recast

akortunov:
An attempt to shut up the [GCC8 warning](https://github.com/recastnavigation/recastnavigation/issues/363), which Recast developers are not going to fix.

### DIFF
--- a/extern/recastnavigation/Detour/Source/DetourNavMesh.cpp
+++ b/extern/recastnavigation/Detour/Source/DetourNavMesh.cpp
@@ -210,12 +210,7 @@ dtNavMesh::~dtNavMesh()
 {
 	for (int i = 0; i < m_maxTiles; ++i)
 	{
-		if (m_tiles[i].flags & DT_TILE_FREE_DATA)
-		{
-			dtFree(m_tiles[i].data);
-			m_tiles[i].data = 0;
-			m_tiles[i].dataSize = 0;
-		}
+		m_tiles[i].~dtMeshTile();
 	}
 	dtFree(m_posLookup);
 	dtFree(m_tiles);
@@ -234,13 +229,16 @@ dtStatus dtNavMesh::init(const dtNavMeshParams* params)
 	if (!m_tileLutSize) m_tileLutSize = 1;
 	m_tileLutMask = m_tileLutSize-1;
 	
-	m_tiles = (dtMeshTile*)dtAlloc(sizeof(dtMeshTile)*m_maxTiles, DT_ALLOC_PERM);
-	if (!m_tiles)
+	void* mem = dtAlloc(sizeof(dtMeshTile)*m_maxTiles, DT_ALLOC_PERM);
+	if (!mem)
 		return DT_FAILURE | DT_OUT_OF_MEMORY;
+
+	memset(mem, 0, sizeof(dtMeshTile)*m_maxTiles);
+	m_tiles = ::new(mem) dtMeshTile[m_maxTiles];
+
 	m_posLookup = (dtMeshTile**)dtAlloc(sizeof(dtMeshTile*)*m_tileLutSize, DT_ALLOC_PERM);
 	if (!m_posLookup)
 		return DT_FAILURE | DT_OUT_OF_MEMORY;
-	memset(m_tiles, 0, sizeof(dtMeshTile)*m_maxTiles);
 	memset(m_posLookup, 0, sizeof(dtMeshTile*)*m_tileLutSize);
 	m_nextFree = 0;
 	for (int i = m_maxTiles-1; i >= 0; --i)
--- a/extern/recastnavigation/Detour/Include/DetourNavMesh.h
+++ b/extern/recastnavigation/Detour/Include/DetourNavMesh.h
@@ -306,6 +306,19 @@ struct dtMeshTile
 	int dataSize;							///< Size of the tile data.
 	int flags;								///< Tile flags. (See: #dtTileFlags)
 	dtMeshTile* next;						///< The next free tile, or the next tile in the spatial grid.
+
+	dtMeshTile() {}
+
+	~dtMeshTile()
+	{
+		if(flags & DT_TILE_FREE_DATA)
+		{
+			dtFree(data);
+			data = 0;
+			dataSize = 0;
+		}
+	}
+
 private:
 	dtMeshTile(const dtMeshTile&);
 	dtMeshTile& operator=(const dtMeshTile&);


akortunov:
Casting @psi29a here.

extern/recastnavigation/Detour/Source/DetourNavMesh.cpp
```diff
@@ -240,7 +240,7 @@ dtStatus dtNavMesh::init(const dtNavMeshParams* params)
 	m_posLookup = (dtMeshTile**)dtAlloc(sizeof(dtMeshTile*)*m_tileLutSize, DT_ALLOC_PERM);
 	if (!m_posLookup)
 		return DT_FAILURE | DT_OUT_OF_MEMORY;
-	memset(m_tiles, 0, sizeof(dtMeshTile)*m_maxTiles);
+	memset(static_cast<void*>(m_tiles), 0, sizeof(dtMeshTile)*m_maxTiles);
```
kcat: It's not a false-positive when it's true. [It's undefined behavior to use `memset` on a non-trivially copyable type](https://en.cppreference.com/w/cpp/string/byte/memset), and [declaring non-defaulted and non-deleted copy methods makes it not trivially copyable](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable)). You don't know what the compiler may do to non-trivial structs, now or in the future, that can break this call.

psi29a: This is also 'extern' code. If there is a proper fix then we need to upstream it.

akortunov: According to Recast developers, declaring copy methods without implementation is a workaround for compilers, which do not support C++11:

> The structure being cleared is just a POD type with deleted copy-constructor and assignment operator.
It appears GCC8 does not recognize the old-style idiom used for deleting auto-generated members (where you just declare them privately without an implementation). We use this approach to support older compilers

kcat: > According to Recast developers, declaring copy methods without implementation is a workaround for compilers, which do not support C++11

Yes, but all the compiler sees is the copy method declarations in private scope, and needs to assume they could be defined elsewhere, even if they ultimately never are. It's a useful pre-C++11 trick to prevent the copy operators from being used, but it makes it not trivially copyable as a consequence (a fact the compiler can leverage however it wants, including ways that make memset invalid on it).

> It appears GCC8 does not recognize the old-style idiom used for deleting auto-generated members (where you just declare them privately without an implementation).

It behaves exactly as it always has. It doesn't truly delete them, it just makes them not callable from outside of the class (and causes a linker error if it manages to do that). But the compiler can't know there's no implementation when parsing the class, since the methods can be in another source file.

akortunov: Are there "proper" fixes for this issue?

kcat: Adding a `clear()` method to `dtMeshTile` (which sets all the members to 0/null), and calling it like:
```
std::for_each(m_tiles, m_tiles+m_maxTiles, std::mem_fn(&dtMeshTile::clear));
```
should work. Using proper construction and destruction would be better, but that would be more invasive changes.

akortunov: IIRC, the `std::mem_fn` is a C++11 thing, so it will not work. We need a fix which will work on all C++ versions. Unfortunately, developers tends to just disable the `-Wclass-memaccess`, so I did not find good alternatives so far.

kcat: Pre-C++11 has `std::mem_fun`.

akortunov: > Pre-C++11 has std::mem_fun.

Which is deprecated in C++11 and removed in C++17, so there will be if-defs, which Recast developers will not accept. Weird.

akortunov: How about something like that:
```
    for (int i = 0; i < m_maxTiles; ++i)
	{
        m_tiles[i].clear();
    }
```

and:
```
    void clear()
    {
        salt = 0;
        linksFreeList = 0;
        header = NULL;
        polys = NULL;
        verts = NULL;
        links = NULL;
        detailMeshes = NULL;
        detailVerts = NULL;
        detailTris = NULL;
        bvTree = NULL;
        offMeshCons = NULL;
        data = NULL;
        dataSize = 0;
        flags = 0;
        next = NULL;
    }
```

akortunov: An another approach:
```
void* mem = dtAlloc(sizeof(dtMeshTile)*m_maxTiles, DT_ALLOC_PERM);
memset(mem, 0, sizeof(dtMeshTile)*m_maxTiles);
m_tiles = (dtMeshTile*)mem;
```

akortunov: @elsid, what do you think about this issue?

kcat: > ```
> void* mem = dtAlloc(sizeof(dtMeshTile)*m_maxTiles, DT_ALLOC_PERM);
> m_tiles = (dtMeshTile*)mem;
> memset(mem, 0, sizeof(dtMeshTile)*m_maxTiles);
> ```

May as well just go the extra step:
```
void* mem = dtAlloc(sizeof(dtMeshTile)*m_maxTiles, DT_ALLOC_PERM);
memset(mem, 0, sizeof(dtMeshTile)*m_maxTiles);
m_tiles = ::new(mem) dtMeshTile[m_maxTiles];
```
That's the proper C++ way to create a new array in user-provided storage. To be extra proper, it should also run the destructor on the array before freeing:
```
for(int i = 0;i < m_maxTiles;++i)
    m_tiles[i].~dtMeshTile();
dtFree(m_tiles);
```

akortunov:
```
void* mem = dtAlloc(sizeof(dtMeshTile)*m_maxTiles, DT_ALLOC_PERM);
memset(mem, 0, sizeof(dtMeshTile)*m_maxTiles);
m_tiles = ::new(mem) dtMeshTile[m_maxTiles];
```

Does not compile:
```
no matching function for call to «dtMeshTile::dtMeshTile()»
```

akortunov:
```
for(int i = 0;i < m_maxTiles;++i)
    m_tiles[i].~dtMeshTile();
dtFree(m_tiles);
```

I suppose it is handled in this loop:
```
	for (int i = 0; i < m_maxTiles; ++i)
	{
		if (m_tiles[i].flags & DT_TILE_FREE_DATA)
		{
			dtFree(m_tiles[i].data);
			m_tiles[i].data = 0;
			m_tiles[i].dataSize = 0;
		}
	}
```

kcat: You need to define a default/empty constructor since the copy constructor inhibits the automatic one.

kcat: > I suppose it is handled in this loop

Technically, that if check and the extra `dtFree` call should be in `dtMeshTile`'s destructor. i.e.
```
dtMeshTile::~dtMeshTile()
{
    if(flags & DT_TILE_FREE_DATA)
        dtFree(data);
}

...

for(int i = 0;i < m_maxTiles;++i)
    m_tiles[i].~dtMeshTile();
dtFree(m_tiles);
```

akortunov: Are you sure the `::new(mem)` does not allocate new memory and uses the `mem` instead?

kcat: > Are you sure the `::new(mem)` does not allocate a new memory and uses the mem instead?

I'm positive, that's called placement new. It constructs the objects in-place and returns a pointer to the correct type.

akortunov: I uploaded a new commit, despite I am still not convinced that touching a destructior is a good idea - suggested changed do not fit the `dtMeshTile` usage by R&D itself.

elsid: First I'd try to suggest to remove private copy constructor and assignment operator. This type is a POD by intention. Actually to copy a POD is a correct behaviour since there is no destructor and user controls when to release any locked resources by an object of such type. They were added by https://github.com/recastnavigation/recastnavigation/commit/b7a83e0afbed5643600bad2b93febbad323333d1 that doesn't explicitly mentioned this in the commit message. Not sure is there any reasoning behind. IMO saying "it's a POD but we don't allow to copy it" is not consistent with C++.

If this will not be accepted, then I agree with kcat.

I've tried somehow to break code using such UB as memset on an non trivially copyable object. But didn't get any different results in assembly by different compilers. Probably compilers don't pay attention for it.

akortunov: So a suggested approach then:
1. Describe the issue to R&D devs (they disabled the `-Wclass-memaccess` via `premake` so far) and link this discussion, so they can make decision themselves - **done**.
2. Remove these private methods from our local copy of R&D, so we will silence this warning in OpenMW while upstream R&D does not have the fix.

kcat: > I've tried somehow to break code using such UB as memset on an non trivially copyable object. But didn't get any different results in assembly by different compilers. Probably compilers don't pay attention for it.

The problem is ultimately that the language allows a compiler to do unorthodox things with non-trivial types, and the compiler is warning you to not do it. A compiler may not right now do anything that could break with memset, but it could at any time in the future without further warning. Disabling the warning would also hide legitimate problems, like if someone adds a `std::string` to a previously trivial type that gets used with `memset` or `memcpy`.

C++ libraries really have no reason to cater to pre-C++11 compilers, unless it's targeting old programs that aren't updated anymore and need an ABI of a particular compiler version. C++11 is over 8 years old, and has way too many features to improve the readability, safety, and maintenance of code. I might even argue that C++11 is getting outdated, but C++14 doesn't really add too much over it and I'd agree C++17 is still a bit too new to require. But as compilers improve, they're going to look at doing more and more things that can help code optimizations and tooling, so clever tricks you could get away with before aren't always going to be the ideal solution.

psi29a: C++14 is just a bug-fix on top of C++11, for anyone moving from 03, they should step right into 14.

psi29a:
Have appveyor a kick

psi29a:
What's the verdict here? Are we going to sit on this until it is upstreamed or are we going to fork again?

akortunov:
> Can you put your changes to non extern/* in another PR and we can merge that, keep this PR open for fixes in navmesh?

Done.

> What's the verdict here? Are we going to sit on this until it is upstreamed or are we going to fork again?

We already created a bugreport, where we described the issue and provided several ways to fix it.
R&D's developers current approach is just disable `-Wclass-memaccess` warnings via premake - they do not treat such warnings as potential issues and they are not going to patch them, and there is little we can do about it from our side.

I'd suggest just to remove two "pseudo-deleted" methods declarations from our `extern` folder to suppress the warning and do not touch anything else.

akortunov:
So what should we do here?

akortunov:
Closing due to lack of feedback.

PR #2612, opened 2019-11-28
Fix most of deprecated API usage

akortunov:
There are many warnings in the editor and the launcher when you build them against Qt 5.13+ (as in our MacOS CI). This PR fixes most of them:

1. The `QGuiApplication::primaryScreen()->geometry();` and the `QApplication::desktop()->screenGeometry();` return the same QRect (at least, if you have a single display) and the former is the proper way to get screen dimensions in the Qt5.

2. The `tabStopWidth` and font width were renamed for some reason in Qt 5.11

It would be nice if someone could test this PR on multi-display setup.

Also probably it is worth to drop Qt4 support at all. In this case we should limit target Qt version in CMake and search for Qt5 by default (-DDESIRED_QT_VERSION).

### DIFF
--- a/apps/launcher/graphicspage.cpp
+++ b/apps/launcher/graphicspage.cpp
@@ -5,6 +5,10 @@
 #include <QMessageBox>
 #include <QDir>
 
+#if QT_VERSION >= QT_VERSION_CHECK(5,0,0)
+#include <QScreen>
+#endif
+
 #ifdef MAC_OS_X_VERSION_MIN_REQUIRED
 #undef MAC_OS_X_VERSION_MIN_REQUIRED
 // We need to do this because of Qt: https://bugreports.qt-project.org/browse/QTBUG-22154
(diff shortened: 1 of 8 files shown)


symbiogenesis:
Qt 5.0 was released in December of 2012. 

For comparison, in the hyper-conservative enterprise world, Microsoft has ended "mainstream support" for Windows Server 2012.

akortunov:
@psi29a, can you tell if we support multiscreen configurations for the editor?

Judging by docs, the `QGuiApplication::primaryScreen()->geometry();` returns geometry of active screen while the `QApplication::desktop()->screenGeometry();` returns overall geometry for all  screens.
Does the difference matter for us and which behaviour is preferred?

psi29a:
@akortunov yes, we should support multi-screen workflows.

akortunov:
> Yes, we should support multi-screen workflows.

Then there should be a some kind of specification about how OpenMW-CS is supposed to work in this workflow.

PR #2613, opened 2019-11-28
Fix GCC9 warnings about implicit declarations

akortunov:
Implicit declarations are deprecated in GCC now, so there is a spam in the build log.
Basically, GCC does not like when there is a user-defined copy constructor, but no copy assignment operator (or vice versa).

### DIFF
--- a/components/nifosg/particle.hpp
+++ b/components/nifosg/particle.hpp
@@ -78,6 +78,8 @@ namespace NifOsg
         ParticleShooter();
         ParticleShooter(const ParticleShooter& copy, const osg::CopyOp& copyop = osg::CopyOp::SHALLOW_COPY);
 
+        ParticleShooter& operator=(const ParticleShooter&) = delete;
+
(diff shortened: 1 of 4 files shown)


components/nifosg/particle.hpp
```diff
@@ -135,6 +137,8 @@ namespace NifOsg
         GrowFadeAffector();
         GrowFadeAffector(const GrowFadeAffector& copy, const osg::CopyOp& copyop = osg::CopyOp::SHALLOW_COPY);
 
+        GrowFadeAffector& operator=(const GrowFadeAffector&) = default;
```
elsid: This operator is used inside ctor like:
```c++
*this = copy;
```
This expression will call assignment operator for the base class which is probably not desired here. See [this](https://godbolt.org/z/jtNXyG) example. I suggest to mark assignment operator as deleted and change `*this = copy` to explicit members copy.

akortunov: Done.

PR #2615, opened 2019-11-29
Fix iterator usage when destroying cell borders

akortunov:
Fixes [bug #5218](https://gitlab.com/OpenMW/openmw/issues/5218).

Since we are going to remove all items from map anyway, just remove scene nodes in loop and clear the map.

### DIFF
--- a/components/terrain/cellborder.cpp
+++ b/components/terrain/cellborder.cpp
@@ -91,8 +91,9 @@ void CellBorder::destroyCellBorderGeometry(int x, int y)
 
 void CellBorder::destroyCellBorderGeometry()
 {
-    for (CellGrid::iterator it = mCellBorderNodes.begin(); it != mCellBorderNodes.end(); ++it)
-        destroyCellBorderGeometry(it->first.first,it->first.second);
+    for (const auto& v : mCellBorderNodes)
+        mRoot->removeChild(v.second);
+    mCellBorderNodes.clear();
 }
 
 }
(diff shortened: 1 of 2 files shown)


components/terrain/cellborder.cpp
```diff
@@ -76,23 +76,24 @@ void CellBorder::createCellBorderGeometry(int x, int y)
...
+    return mCellBorderNodes.erase(it);
 }
 
 void CellBorder::destroyCellBorderGeometry()
 {
-    for (CellGrid::iterator it = mCellBorderNodes.begin(); it != mCellBorderNodes.end(); ++it)
-        destroyCellBorderGeometry(it->first.first,it->first.second);
+    CellGrid::iterator it = mCellBorderNodes.begin();
+    while (it != mCellBorderNodes.end())
+        it = destroyCellBorderGeometry(it->first.first,it->first.second);
```
elsid: This implementation will call `map::find` to get the iterator that is already here. Just increases complexity from `O(n)` to `O(log n)`. I suggest to separate these operations. Leave single node `destroyCellBorderGeometry` as is but change for all to this:
```c++
void CellBorder::destroyCellBorderGeometry()
{
    for (const auto& v : mCellBorderNodes)
        mRoot->removeChild(v.second);
    mCellBorderNodes.clear();
}
```
It's much clear that we want to remove all of them. Call erase on each remove also may have an additional cost.

akortunov: Done.

PR #2616, opened 2019-11-29
Several languages for the launcher

Brouilles:
The integration of QTranslator into the Launcher. Because it's an element that is used by a lambda user, it seemed like a good idea to be able to translate it.

The application get the local language `fr-FR` for example and search a `fr.qm` file with the two first characters (So it's also working gor fr-CA). If the translation doesn't exist, the launcher is in English.

I have added a partial translation for french (just all tooltips are not translated) but the rest is.

### DIFF
--- a/apps/launcher/main.cpp
+++ b/apps/launcher/main.cpp
@@ -1,6 +1,7 @@
 #include <iostream>
 
 #include <QApplication>
+#include <QTranslator>
 #include <QTextCodec>
 #include <QDir>
 #include <QDebug>
@@ -25,6 +26,13 @@ int main(int argc, char *argv[])
 #endif
 
         QApplication app(argc, argv);
+
+        // Internationalization 
+        QString locale = QLocale::system().name().section('_', 0, 0);
+
+        QTranslator appTranslator;
+        appTranslator.load(":/translations/" + locale + ".qm");
+        app.installTranslator(&appTranslator);
 
         // Now we make sure the current dir is set to application path
         QDir dir(QCoreApplication::applicationDirPath());
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -74,6 +74,7 @@ Programmers
     Fil Krynicki (filkry)
     Finbar Crago (finbar-crago)
     Florian Weber (Florianjw)
+    Gaëtan Dezeiraud (Brouilles)
     Gašper Sedej
     Gijsbert ter Horst (Ghostbird)
     Gohan1989
--- a/apps/launcher/maindialog.cpp
+++ b/apps/launcher/maindialog.cpp
@@ -55,6 +55,7 @@ Launcher::MainDialog::MainDialog(QWidget *parent)
     iconWidget->setFlow(QListView::LeftToRight);
 
     QPushButton *playButton = new QPushButton(tr("Play"));
+    buttonBox->button(QDialogButtonBox::Close)->setText(tr("Close"));
     buttonBox->addButton(playButton, QDialogButtonBox::AcceptRole);
 
     connect(buttonBox, SIGNAL(rejected()), this, SLOT(close()));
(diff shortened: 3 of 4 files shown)


akortunov:
I am not sure if is is a good idea to do it right now. OpenMW is still in the pre-1.0 state, and we modify the launcher pretty often. I'd not like to update translations every time when we add a new checkbox to the Advanced tab.

Brouilles:
I understand, how you want!
You can close the PR if it's to early.

Capostrophic:
Right, this could increase the maintenance cost for an incomplete launcher significantly. Every new translation would require getting people that know the affected language(s) to help translate almost every single improvement the launcher gets.

However, would it be possible to just add the framework for the localization without actually adding any support for translation files or translation files at this moment? So it would be possible to easily add the translations in some future when the launcher is deemed sufficiently ready for that.

apps/launcher/main.cpp
```diff
@@ -1,6 +1,7 @@
 #include <iostream>
 
 #include <QApplication>
+#include <QTranslator.h>
```
Capostrophic: Shouldn't this just be `<QTranslator>`?

Brouilles: Yes!

Brouilles:
I have push a new commit. It's remove french translation. I just leave the .ts file for the example for the futur. Seems good for you?

Capostrophic:
Mhm. Yeah, I guess this looks mostly okay, though I'm not convinced the string translation file won't end up being too confusing or outdated over the time. Nice job on the Close button text fix, the inconsistency was killing me.

fr.qm file is not outcommented properly from the qrc file, causes a building error.

Brouilles:
I removed it! :)
Otherwise, you have done an incredible job with this project. Good job!

AnyOldName3:
I'm not super-against having translations done before we're settled on all our strings. We (i.e. MO2 developers) change things in MO2 far more often than we (i.e. us here) do in OpenMW, and MO2 manages to cope. Lots of people are just subscribed to Transifex notifications so find out whenever there are new strings, and are keen enough to translate them without any nagging. It's barely taken any nagging to get several languages translated for some plugins that aren't bundled.

From a technical perspective, this is how things work for MO2:
* Everything that displays a string to the user goes through a `tr` call so Qt's language tools can pick it up.
* CMake for each project is set to generate its base `.ts` file as a dependency of the actual binary.
* When committing something, you're supposed to commit any changes to the `.ts` file, too.
* Transifex occasionally grabs all the `.ts` files from all the repos. I'm not sure if this is manually triggered or on a timer, but it's not a post-commit webhook, even though I think it should be.
* When building MO2 with the right settings (either it's been enabled manually or you're building from a release branch), Transifex's API is queried for all translations more than 70% done.
* The `.ts` files for those languages are downloaded and turned into `.qm` files.
* The built `.qm` files plus the Qt-provided `.qm` files (necessary for standard buttons like 'Yes' and 'Okay') for those languages are set as files to install in CMake in a `translations` subdirectory, so they'll end up in the installer or in the directory CMake's been told to install to.
* When MO2 starts, it looks in its translations directory for `.qm` files and takes note of which languages are available.
* These languages are offered as options in the settings menu and when chosen, the correct language is set.

I'm generally pretty keen on this system. The only changes I'd make for OpenMW would be moving the base `.ts` file out of source control and make it a CI artefact that gets sent to Transifex after a successful master/release branch build and moving the code that does this from MO2's custom meta-meta build system to CMake. I don't think we should keep translated `.ts` files or compiled `.qm` files in source control, and I don't think we should hardcode the list of available languages.

akortunov:
I suppose such major change is a term of long discussion on forum.

As for this PR... @Brouilles, I'd suggest you to add yourself to the AUTHORS.md and squash commits.

Brouilles:
@akortunov Done!

PR #2618, opened 2019-11-30
Fix objects culling for recast mesh tiles (bug #5216)

elsid:
Mainly fixes bug [#5216](https://gitlab.com/OpenMW/openmw/issues/5216).

1. All objects except heightfields are culled per each triangle instead of optimized quad-tree based approach by Bullet. The reason is it's impossible to have bounding box in shape local coordinates that will cull the same set of polygons as bounding box in global coordianates in all corner cases with rotation. Applying this to heightfields is about 5x slower because they have most of the polygons in the scene. But they usually are not rotated, so local bounding box still works.
2. New [algorithm](http://dev.theomader.com/transform-bounding-boxes/) is used to transform bounding box for heightfields. At least this is a described solution somewhere.
3. To fix the issue with that bridge metioned in [#5216](https://gitlab.com/OpenMW/openmw/issues/5216) recast scale factor is increased and tile size to compensate decreased total navmesh area.

screenshot025

### DIFF
--- a/apps/openmw_test_suite/detournavigator/recastmeshbuilder.cpp
+++ b/apps/openmw_test_suite/detournavigator/recastmeshbuilder.cpp
@@ -361,9 +361,9 @@ namespace
         );
         const auto recastMesh = builder.create();
         EXPECT_EQ(recastMesh->getVertices(), std::vector<float>({
-            0.707107067108154296875, 0, -3.535533905029296875,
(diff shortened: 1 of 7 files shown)


akortunov:
I moved different cells and tested different types of slopes (e.g. bridges and ladders) and did not encounter issues yet.

I suppose that for proper testing we will need to merge this PR anyway.

PR #2619, opened 2019-11-30
Support recast mesh rendering

elsid:
Add a new command `togglerecastmesh` and corresponding settings option `enable nav mesh render` to enable recast mesh rendering. Essentially it's the same thing as `tbc` but transformed several times to be used for navmesh building. The idea is to use in combination with wireframe mode (`togglewireframe`) to reduce Z-fighting problem.

screenshot026

### DIFF
--- a/components/sceneutil/recastmesh.cpp
+++ b/components/sceneutil/recastmesh.cpp
@@ -1,0 +1,48 @@
+#include "navmesh.hpp"
+#include "detourdebugdraw.hpp"
+
+#include <components/detournavigator/settings.hpp>
+#include <components/detournavigator/recastmesh.hpp>
+
+#include <RecastDebugDraw.h>
+
+#include <osg/Group>
+
+namespace
+{
+    std::vector<float> calculateNormals(const std::vector<float>& vertices, const std::vector<int>& indices)
+    {
+        std::vector<float> result(indices.size());
+        for (std::size_t i = 0, n = indices.size(); i < n; i += 3)
+        {
+            const float* v0_ptr = &vertices[indices[i] * 3];
+            const float* v1_ptr = &vertices[indices[i + 1] * 3];
+            const float* v2_ptr = &vertices[indices[i + 2] * 3];
+            const osg::Vec3f v0(v0_ptr[0], v0_ptr[1], v0_ptr[2]);
+            const osg::Vec3f v1(v1_ptr[0], v1_ptr[1], v1_ptr[2]);
+            const osg::Vec3f v2(v2_ptr[0], v2_ptr[1], v2_ptr[2]);
+            const osg::Vec3f e0 = v1 - v0;
+            const osg::Vec3f e1 = v2 - v0;
+            osg::Vec3f normal = e0 ^ e1;
+            normal.normalize();
+            for (std::size_t j = 0; j < 3; ++j)
+                result[i + j] = normal[j];
+        }
+        return result;
+    }
+}
+
+namespace SceneUtil
+{
+    osg::ref_ptr<osg::Group> createRecastMeshGroup(const DetourNavigator::RecastMesh& recastMesh,
+        const DetourNavigator::Settings& settings)
+    {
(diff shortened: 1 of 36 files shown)


Capostrophic:
The Z-fighting seems to be quite aggressive to the point of unusability... no way around that?

elsid:
Do you mean Z-fighting of the recast mesh with itself or with rendered world or with wireframe? I'd say in close distances it's quite ok. Far usually are not used. Anyway it's a debugging tool and Z-fighting is not really important here.

components/sceneutil/recastmesh.cpp
```diff
@@ -0,0 +1,60 @@
...
+            {
+                d = 1.0f / d;
+                normal[0] *= d;
+                normal[1] *= d;
+                normal[2] *= d;
+            }
+        }
+        return result;
+    }
+}
```
AnyOldName3: This would probably be much easier to read if it used OSG vector types instead of floats.

AnyOldName3: Okay, I see that this gets passed to some external code from Recast. I *think* there's a nice, clean, cross-platform way of getting an OSG vector array or buffer or something as contiguous float memory (as otherwise there'd be no way to put it on a GPU), but if it turns out that accessing it is super-ugly, I can excuse doing vector operations on scalar types.

elsid: I had some concerns about using `osg::Vec3f` but it seems  there is no difference in performance: http://quick-bench.com/sssJs5iB8J_GkZlmq6X10iHdIK0
Also quite interesting see here clang 8 beating gcc 9.1: http://quick-bench.com/6ElyfweLhiS85UxO5AwN6gT09_Y

AnyOldName3: There's lots of inlining, so there's no practical difference once that's been applied.

components/sceneutil/recastmesh.cpp
```diff
@@ -0,0 +1,52 @@
...
+        {
+            const float* v0_ptr = &vertices[indices[i] * 3];
+            const float* v1_ptr = &vertices[indices[i + 1] * 3];
+            const float* v2_ptr = &vertices[indices[i + 2] * 3];
+            const osg::Vec3f v0(v0_ptr[0], v0_ptr[1], v0_ptr[2]);
+            const osg::Vec3f v1(v1_ptr[0], v1_ptr[1], v1_ptr[2]);
+            const osg::Vec3f v2(v2_ptr[0], v2_ptr[1], v2_ptr[2]);
+            const osg::Vec3f e0 = v1 - v0;
+            const osg::Vec3f e1 = v2 - v0;
+            osg::Vec3f normal(
```
MiroslavR: Why not use operator^ for cross product?

components/sceneutil/recastmesh.cpp
```diff
@@ -0,0 +1,48 @@
...
+    {
+        std::vector<float> result(indices.size());
+        for (std::size_t i = 0, n = indices.size(); i < n; i += 3)
+        {
+            const float* v0_ptr = &vertices[indices[i] * 3];
+            const float* v1_ptr = &vertices[indices[i + 1] * 3];
+            const float* v2_ptr = &vertices[indices[i + 2] * 3];
+            const osg::Vec3f v0(v0_ptr[0], v0_ptr[1], v0_ptr[2]);
+            const osg::Vec3f v1(v1_ptr[0], v1_ptr[1], v1_ptr[2]);
+            const osg::Vec3f v2(v2_ptr[0], v2_ptr[1], v2_ptr[2]);
```
AnyOldName3: I thought there was a constructor that just took one `float*`.

elsid: There is no https://github.com/OpenMW/osg/blob/3.4/include/osg/Vec3f#L41-L43 .

psi29a:
Kicked travis to rebuid...

PR #2622, opened 2019-12-01
Implement TestCells and TestInteriorCells console commands

akortunov:
Implements [feature #5219](https://gitlab.com/OpenMW/openmw/issues/5219).

Adds two commands - TestCells and TestInteriorCells.

Current limitations:
1. Testing still requires some additional memory (usually near 1 Gb for all interior and exterior cells). This RAM is used to keep cached assets to do not load them from disk for every tested cell.
I use a more short cache life time, otherwise memory usage will be larger.

2. Since I did not figure out how to keep loaded exterior and interior cells in the same time, I allow to use the TestCells only from main menu. It also allows to decrease memory usage.

3. Current implementation seems to do not detect scripting errors since we do not run scripts during testing, but OpenMW-CS already handles scripting errors pretty well. Maybe it is worth to add an another command to compile all scripts during runtime.

4. We do not update sound system during cell testing, so if the main theme ends, we restart it only AFTER testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -235,6 +235,7 @@
     Feature #5147: Show spell magicka cost in spell buying window
     Feature #5170: Editor: Land shape editing, land selection
     Feature #5193: Weapon sheathing
+    Feature #5219: Impelement TestCells console command
     Feature #5224: Handle NiKeyframeController for NiTriShape
     Task #4686: Upgrade media decoder to a more current FFmpeg API
     Task #4695: Optimize Distant Terrain memory consumption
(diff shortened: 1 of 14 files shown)


Capostrophic:
>Maybe I assigned a wrong opcode

Yes, that was an opcode that belonged to one of GetAttribute variants, that's why there were weird issues. This part of the scripting system is not horribly broken.

akortunov:
Also maybe it is worth to hide console when launching a test.

akortunov:
Any feedback?

Capostrophic:
Seems to work okay, although I wish there could be a way to stop cell testing in progress (maybe it could be aborted when Escape is pressed). However it doesn't seem like it's possible to make a loading operation finish abruptly due to keyboard input. Maybe later.

PR #2623, opened 2019-12-02
Add High DPI support for Windows (Launcher / Wizard / OpenMW-CS)

Brouilles:
The launcher, wizard and OpenMW-CS now support High DPI screen for Windows. And with multiple screens (the scale is automatic when the window pass to a High DPI screen to standard). Before, windows (so applications) could not be used because of scale issues.

Tested on Windows 10. But normally work on Windows Vista/7/8/8.1.

### DIFF
--- a/apps/launcher/CMakeLists.txt
+++ b/apps/launcher/CMakeLists.txt
@@ -15,6 +15,7 @@ set(LAUNCHER
     utils/lineedit.cpp
 
     ${CMAKE_SOURCE_DIR}/files/windows/launcher.rc
+    ${CMAKE_SOURCE_DIR}/files/windows/qt.exe.manifest
 )
 
 set(LAUNCHER_HEADER
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -1,5 +1,6 @@
 set (OPENCS_SRC main.cpp
     ${CMAKE_SOURCE_DIR}/files/windows/opencs.rc
+    ${CMAKE_SOURCE_DIR}/files/windows/qt.exe.manifest
     )
 
 opencs_units (. editor)
--- a/apps/wizard/CMakeLists.txt
+++ b/apps/wizard/CMakeLists.txt
@@ -17,6 +17,7 @@ set(WIZARD
 
 if(WIN32)
     list(APPEND WIZARD ${CMAKE_SOURCE_DIR}/files/windows/openmw-wizard.rc)
+    list(APPEND WIZARD ${CMAKE_SOURCE_DIR}/files/windows/qt.exe.manifest)
 endif()
 
 set(WIZARD_HEADER
--- a/files/windows/qt.exe.manifest
+++ b/files/windows/qt.exe.manifest
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3" manifestVersion="1.0">
+  <!-- Enable use of version 6 of the common controls (Win XP and later) -->
+  <dependency>
+    <dependentAssembly>
+      <assemblyIdentity type="win32"
+                        name="Microsoft.Windows.Common-Controls"
+                        version="6.0.0.0"
+                        processorArchitecture="*"
+                        publicKeyToken="6595b64144ccf1df"
+                        language="*" />
+    </dependentAssembly>
+  </dependency>
+
+  <!-- Indicate UAC compliance, with no need for elevated privileges -->
+  <!-- (if you need enhanced privileges, set the level to "highestAvailable" or "requireAdministrator") -->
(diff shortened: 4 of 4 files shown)


apps/launcher/main.cpp
```diff
@@ -12,13 +13,29 @@
 #define MAC_OS_X_VERSION_MIN_REQUIRED __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__
 #endif // MAC_OS_X_VERSION_MIN_REQUIRED
 
+#ifdef _WIN32
+#include <Windows.h>
+#endif
```
AnyOldName3: Can you get away with `WIN32_LEAN_AND_MEAN` here? It's supposed to shave a significant amount off compile times.

In fact, are you 100% sure you need to do this at all? Qt should have cross-platform DPI scaling stuff that doesn't need you to include anything OS-specific.

Brouilles: Good point for `#define WIN32_LEAN_AND_MEAN`. I forgot. 
Yes, it's the only true solution for support High DPI on Windows if we have multiple screens. To my knowledge. 

Because ` QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);` don't resize correctly some inputs (like checkbox) and the Windows window border stay very small. And QMessageBox are not resized correctly. It's working only if we have only one screen. Or multiple High DPI screens.

With this code. No problem. Other solution (but similar) is to add a parameter in the manifest of the Windows Application. It's do the same. But directly implement the code seems to me more comprehensive (because currently applications don't have it).

akortunov:
1. This PR in its current state does not build on Windows.
2. We need a cross-platform solution. If there are no such solutions, update a manifest indeed.

Brouilles:
1. It's work on my Windows 10 main machine. What is your Windows system?
2. The manifest is only for Windows too. See https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/mt846517(v%3Dvs.85)

I don't have linux machine for testing. For macOS it's working good because all devices are High DPI and macOS have a better support (QT by default enable High DPI support on this OS).

akortunov:
> For macOS it's working good because all devices are High DPI and macOS have a better support (QT by default enable High DPI support on this OS).

No, we just use the `NSHighResolutionCapable` flag in plist files.

akortunov:
>  I see appveyor don't find SetProcessDpiAwarenessT. I search why.

Google says that the `SetProcessDpiAwareness` was added only in Windows 8.1. I suspect that CI uses Windows 7.
If it is true, you can not use this function (I am even not sure if we officeally dropped Windows XP support).

Brouilles:
Ok. So the Manifest solution seems to be a better option. Because if the system don't support this options, it's just ignored by Windows. 

I know how to add Manifest with QT Creator but not QT with Visual Studio. I search and I'm come back.

Brouilles:
@akortunov Here the same result (fix the High DPI by screens) with Windows Manifest.
It's better actually. And no problem with Windows XP (just no High DPI support. But I do not even know if Windows XP supports it for the system).

AnyOldName3:
It looks like a network issue that killed CI. I don't know how to kick it, so either someone who does needs to do that, or another push is needed.

Brouilles:
I have make a new push for restart CI.

akortunov:
There is something wrong with your commit name - looks like you forgot to rename it during rebase process.

@AnyOldName3, do you have any complaints here?

Brouilles:
@akortunov It's better?

akortunov:
> It's better?

Yes, it is.

AnyOldName3:
I don't know much about high-DPI support or manifest files in Windows, so I can't be much help there, but I ran the AppVeyor artefact and noticed some things.

At 100% scaling, the launcher looked a little weird. It might just be that it's changed in master-branch since I last used it, but the graphics and settings pages looked much worse than in the 0.45 release build.

At 175% scaling, the launcher's layout broke:
image
This means that there's clearly more work that needs to be done to have high-DPI support than just telling Windows that we have high-DPI support. The flag isn't on by default because it usually requires changes to an application to make it not look ugly.

Brouilles:
I don't have try when changing window percentage scaling like you. But on my 5K and 1080p screen (dual screen) with this manifest, I have the good scale on the 5k screen. Whereas before no. And the good scale on the 1080p screen (even if I move the window between each other).

There is normally nothing more to do in the application itself.

AnyOldName3:
> There is normally nothing more to do in the application itself. The manifest do the same thing like set High DPI on the .exe file (right click -> Compatibility -> Change High DPI settings).

That depends on whether the application is setting sizes and positions manually and things like that - Qt will handle most things for us if we let it do everything, but I'm guessing we've done something custom to make the Play button look like it's on some parchment.

The reason you're probably not seeing issues when I am is that I made sure to test with a weird non-integer scaling level of 1.75, but you're on a neat 5x. Scaling bugs are much more likely to present themselves with non-integer values, so that's why I tested that value specifically.

akortunov:

PR #2625, opened 2019-12-03
[Regression] Attempt to fix idle state reset again

Capostrophic:
Guess I didn't look well enough.
Sneaking check in movement animation refresh didn't work properly because sneak idle state was never reset even if the animation was disabled. Caused issues with movement animation speed in first person view. Briefly hold the sneak button while moving and you'll notice that the running/walking animation becomes super fast, and so do the footsteps.

Now it's actually reset. I had to move idle reset to an earlier moment and reset both the pending and the current idle states to make sure the check returns a valid value. Obviously there could be something I overlooked...


Capostrophic:
Doesn't seem to reveal deeper-lying issues over prolonged testing.

PR #2631, opened 2019-12-12
Disallow resting if the fall height hasn't been reset (bug #4802)

Capostrophic:
[Bug report.](https://gitlab.com/OpenMW/openmw/issues/4802)

Compare the falling height with an epsilon of 0.0001 to make sure falling height is reset and falling damage is applied before the player can open the waiting dialog. I haven't been able to reproduce the issue with this, but it's pretty difficult to do anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@
     Bug #4787: Sneaking makes 1st person walking/bobbing animation super-slow
     Bug #4797: Player sneaking and running stances are not accounted for when in air
     Bug #4800: Standing collisions are not updated immediately when an object is teleported without a cell change
+    Bug #4802: You can rest before taking falling damage from landing from a jump
     Bug #4803: Stray special characters before begin statement break script compilation
     Bug #4804: Particle system with the "Has Sizes = false" causes an exception
     Bug #4805: NPC movement speed calculations do not take race Weight into account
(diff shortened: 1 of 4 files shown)


anotak:
I'm the original bug reporter, and although I'm not familiar with the internals of OpenMW, I have a question & opinion about this. It seems to me, based on inspecting the code and my experience with other programming, the core issue is that:
1) the player lands, whatever internal stuff is done to make `mPhysics->isOnSolidGround(player)` return true,
2) then the rest input is processed, allowing sleep,
3) then falling damage is applied.

While the epsilon check does fix this particular case, is there a strong reason that falling damage is not applied before, for example, processing inputs? That seems to me to be the larger inconsistency, as not doing that sooner means other currently-unknown bugs could creep in. I imagine, for example, if one opens the inventory, and does something from there that checks `mPhysics->isOnSolidGround(player)`, that it can still be done before falling damage is applied.

I can understand if there is some strong internal reason for this ordering. If that's the case, I would suggest that perhaps the [function in question](https://github.com/OpenMW/openmw/blob/bd18655d23faa99bcd2264792874195d91df706e/apps/openmw/mwphysics/physicssystem.cpp#L628) should be marked with a comment explaining that it may have unexpected behavior like this.

Capostrophic:
Mechanics update is done after GUI update and physics update which means that anything can happen.

PR #2635, opened 2019-12-15
Change rescaling to be more inline with vanilla (fixes #5214)

Assumeru:
Vanilla save files can contain XSCL records with values outside the [0.5, 2] range.

Interestingly SetScale 10 on Fargoth saved a value of 0.8, while SetScale 8 saved a value of 7. (At least, I think I used 8, might have actually been 7. Doesn't really matter either way.)

### DIFF
--- a/components/esm/cellref.cpp
+++ b/components/esm/cellref.cpp
@@ -145,7 +145,12 @@ void ESM::CellRef::save (ESMWriter &esm, bool wideRefNum, bool inInventory, bool
     }
 
     if (mScale != 1.0) {
-        esm.writeHNT("XSCL", mScale);
+        float scale = mScale;
+        if (scale < 0.5)
+            scale = 0.5;
+        else if (scale > 2)
+            scale = 2;
+        esm.writeHNT("XSCL", scale);
     }
 
     esm.writeHNOCString("ANAM", mOwner);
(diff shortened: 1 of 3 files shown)


components/esm/cellref.cpp
```diff
@@ -67,6 +67,10 @@ void ESM::CellRef::loadData(ESMReader &esm, bool &isDeleted)
                 break;
             case ESM::FourCC<'X','S','C','L'>::value:
                 esm.getHT(mScale);
+                if (mScale < 0.5)
+                    mScale = 0.5;
+                else if (mScale > 2)
+                    mScale = 2;
```
akortunov: Probably it would be better to cap scale during saving instead.

Assumeru: It'd have to be in addition to deal with content files containing uncapped scales - of course those shouldn't exist, but...

Think it'd be worthwhile to put the capping into its own function so it can be used in both places?

Capostrophic: I honestly don't think clamping needs to be its own function, the code will be readable enough anyways. We can't remove clamping upon loading as old saves don't necessarily have the scale clamped. Clamping upon saving won't break anything but it might be redundant anyway.

PR #2636, opened 2019-12-17
Remove FactionID editing from OpenCS

Assumeru:
The ability to edit a value that's never used seems pointless at best and confusing at worst.

### DIFF
--- a/apps/opencs/model/world/columns.cpp
+++ b/apps/opencs/model/world/columns.cpp
@@ -291,7 +291,7 @@ namespace CSMWorld
             { ColumnId_UChar, "Value [0..255]" },
             { ColumnId_NpcMisc, "NPC Misc" },
             { ColumnId_Level, "Level" },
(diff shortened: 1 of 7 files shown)


apps/opencs/model/world/columns.cpp
```diff
@@ -291,7 +291,7 @@ namespace CSMWorld
             { ColumnId_UChar, "Value [0..255]" },
             { ColumnId_NpcMisc, "NPC Misc" },
             { ColumnId_Level, "Level" },
-            { ColumnId_NpcFactionID, "Faction ID" },
+            // { ColumnId_NpcFactionID, "Faction ID" },
```
Capostrophic: Just remove the lines (here and in esmtool). It's not necessary to comment anything out.
Maybe rename mFactionId variable to something like "mUnknown" or "mPadding" and leave a comment saying "always 0".

PR #2637, opened 2019-12-17
Hostile followers travel with player bugfix

Br0keb:
[Bug #5101](https://gitlab.com/OpenMW/openmw/issues/5101)
As it is my first time contributing I wanted to make my change as much non-invasive as possible and that's the cleanest solution I could think of. 
1) getFollowersToTeleport() function **must** insert only viable followers into passed set as it is used not only by executeImp() function but also other functions like the one that counts price for travel. 
2) Since we want to stopCombat() only after the teleportation was made (and not, for example, after prices were counted) executeImp() needs to know the list of followers who were in combat with the player in order to stopCombat().

To do that, I added a null by default pointer to a set of followers that are currently in combat with the player as an argument in getFollowersToTeleport(). Obviously, now thanks to appropriate condition nothing will be inserted into this set if the set was not specified in the function call.
I tested it in-game and everything works as expected. It still is not Morrowind behavior though. The player will not have to pay for summoned creature nor will he travel with the summoned creature. But when he returns back the creature will disappear, which is intended OpenMW behavior that I shouldn't change.

### DIFF
--- a/apps/openmw/mwworld/actionteleport.cpp
+++ b/apps/openmw/mwworld/actionteleport.cpp
@@ -78,9 +78,7 @@ namespace MWWorld
             if (follower.getClass().getCreatureStats(follower).getAiSequence().isInCombat(world->getPlayerPtr()))
             {
                 if(inCombatSet)
-                {
                     inCombatSet->insert(follower);
-                }
                 continue;
             } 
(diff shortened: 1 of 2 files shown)


Capostrophic:
Sorry, [it's already being worked on](https://gitlab.com/OpenMW/openmw/merge_requests/137), albeit slowly. Consider leaving feedback to the MR, although there are issues with it the author is aware of.

PR #2638, opened 2019-12-19
Fix ESMTool build

akortunov:
It does not build in upstream master due to error in #2636.


AnyOldName3:
We have pretty much unlimited CI capacity on GitHub, so we should probably set it up for the less-used things.

PR #2640, opened 2019-12-21
Do not store gold-specific value for everything

akortunov:
The mGoldValue field is used only for gold stacks, for other objects it is always 0. So there is no point to store zero values for every object. It allows to decrease save size up to 5% and speedup saving/loading process a bit.
Does not require a savegame format change.

Note: for some reason we did not store mGoldValue = 1, while we init it by 0 by default, so the value became 0 after save/load. Not sure if it is right, more likely it was an oversight or typo.

### DIFF
--- a/components/esm/cellref.cpp
+++ b/components/esm/cellref.cpp
@@ -168,9 +168,8 @@ void ESM::CellRef::save (ESMWriter &esm, bool wideRefNum, bool inInventory, bool
     if (mChargeInt != -1)
         esm.writeHNT("INTV", mChargeInt);
 
-    if (mGoldValue != 1) {
+    if (mGoldValue > 1)
         esm.writeHNT("NAM9", mGoldValue);
-    }
 
     if (!inInventory && mTeleport)
     {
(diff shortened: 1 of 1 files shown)


components/esm/cellref.cpp
```diff
@@ -168,9 +168,8 @@ void ESM::CellRef::save (ESMWriter &esm, bool wideRefNum, bool inInventory, bool
     if (mChargeInt != -1)
         esm.writeHNT("INTV", mChargeInt);
 
-    if (mGoldValue != 1) {
+    if (mGoldValue != 0)
```
Capostrophic: Wouldn't it make sense to initialize mGoldValue by 1 instead? And reset it to 1 for anything that has it as 0. That would allow to skip saving this subrecord for anything that isn't a non-1 septim. Manual references are initialized with 1 for example even though they're often not gold. NAM9 subrecord actually has "number of items" semantics in Morrowind but we don't replicate that yet.

It won't break anything either as it stands atm.

akortunov: Done.

PR #2641, opened 2019-12-21
Attempt to fix MacOS CI

akortunov:
I do not know, why brew can not update CMake.


psi29a:
You have to remove Travis's cmake first, then use brew to install theirs.

nikolaykasyanov:
Yep, it used to be like that but was lost at some point: https://github.com/OpenMW/openmw/commit/956934911a3053442ae8876b8ae735eb2b551e07#diff-175c1386d71a1d8d7c98b614bd34203eL5

PR #2642, opened 2019-12-21
Add safety checks for door state

akortunov:
I noticed that sometimes DoorState for teleporting doors has garbage values (especially for chargen doors) in my saves. From what I can tell, we do not initialize mDoorState in the `ESM::DoorState`, if a door has no custom data, so we write a random unused value in save.

So I init `mDoorState` variables by zero just for sure, also I reset value and print a warning in case when door state has an invalid value.

It should be enough to clean up old saves (and reduce their size a bit), but if we will get these warnings on new saves, we have an another bug to track down and fix.

### DIFF
--- a/components/esm/doorstate.cpp
+++ b/components/esm/doorstate.cpp
@@ -3,6 +3,8 @@
 #include "esmreader.hpp"
 #include "esmwriter.hpp"
 
+#include <components/debug/debuglog.hpp>
+
 namespace ESM
 {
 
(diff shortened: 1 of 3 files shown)


components/esm/doorstate.cpp
```diff
@@ -12,12 +14,20 @@ namespace ESM
...
+            Log(Debug::Warning) << "Skipping an invalid door state (" << mDoorState << ") for door \"" << mRef.mRefID << "\"";
     }
 
     void DoorState::save(ESMWriter &esm, bool inInventory) const
     {
         ObjectState::save(esm, inInventory);
 
+        if (mDoorState < 0 || mDoorState > 2)
+        {
+            Log(Debug::Warning) << "Skipping an invalid door state (" << mDoorState << ") for door \"" << mRef.mRefID << "\"";
```
Capostrophic: Drop the "an" article in both of the warnings, it shouldn't be necessary (a door only has one state).

Maybe use "dropping" instead of "skipping".

Are warnings in ESM reading/writing code even a thing?

akortunov: > Are warnings in ESM reading/writing code even a thing?

Yes.

PR #2646, opened 2019-12-27
Allow to enchant multiple projectiles at once

akortunov:
Implements [feature #3517](https://gitlab.com/OpenMW/openmw/issues/3517).

Summary of changes:
1. Allow to enchant multiple projectiles at once (depending on soulgem charge), as in MCP. Disabled by default.
2. Do not create a new dynamic enchantment record, if there is an existing dynamic enchantment with same stats. We already use the same approach for potions.
3. Do not allow to create ConstantEffect and WhenUsed enchantments for projectiles, as in Morrowind.

Note: maybe it is worth to store an original ID as optional field for manually enchanted items. In this case if an original item ID, enchanted item's name and enchantment stats are the same, we can use an existing dynamic item record instead of creating a new one every time.
It would allow us to stack manually enchanted scrolls and projectiles with same stats.
Unfortunately, it will require a savegame format change.

### DIFF
--- a/apps/openmw/mwmechanics/enchanting.cpp
+++ b/apps/openmw/mwmechanics/enchanting.cpp
@@ -1,6 +1,7 @@
 #include "enchanting.hpp"
 
 #include <components/misc/rng.hpp>
+#include <components/settings/settings.hpp>
 
 #include "../mwworld/manualref.hpp"
 #include "../mwworld/class.hpp"
@@ -62,7 +63,6 @@ namespace MWMechanics
         const MWWorld::Ptr& player = getPlayer();
         MWWorld::ContainerStore& store = player.getClass().getContainerStore(player);
         ESM::Enchantment enchantment;
-        enchantment.mData.mCharge = getGemCharge();
         enchantment.mData.mAutocalc = 0;
         enchantment.mData.mType = mCastStyle;
         enchantment.mData.mCost = getBaseCastCost();
@@ -81,19 +81,26 @@ namespace MWMechanics
             mEnchanter.getClass().skillUsageSucceeded (mEnchanter, ESM::Skill::Enchant, 2);
         }
 
-        if(mCastStyle==ESM::Enchantment::ConstantEffect)
-        {
-            enchantment.mData.mCharge=0;
-        }
         enchantment.mEffects = mEffectList;
 
+        int count = getEnchantItemsCount();
+
+        if(mCastStyle==ESM::Enchantment::ConstantEffect)
+            enchantment.mData.mCharge = 0;
+        else
+            enchantment.mData.mCharge = getGemCharge() / count;
+
+        // Try to find a dynamic enchantment with the same stats, create a new one if not found.
+        const ESM::Enchantment* enchantmentPtr = getRecord(enchantment);
+        if (enchantmentPtr == nullptr)
+            enchantmentPtr = MWBase::Environment::get().getWorld()->createRecord (enchantment);
+
         // Apply the enchantment
-        const ESM::Enchantment *enchantmentPtr = MWBase::Environment::get().getWorld()->createRecord (enchantment);
         std::string newItemId = mOldItemPtr.getClass().applyEnchantment(mOldItemPtr, enchantmentPtr->mId, getGemCharge(), mNewItemName);
 
         // Add the new item to player inventory and remove the old one
-        store.remove(mOldItemPtr, 1, player);
-        store.add(newItemId, 1, player);
+        store.remove(mOldItemPtr, count, player);
+        store.add(newItemId, count, player);
 
         if(!mSelfEnchanting)
             payForEnchantment();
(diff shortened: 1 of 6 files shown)


apps/openmw/mwmechanics/enchanting.cpp
```diff
@@ -224,6 +280,9 @@ namespace MWMechanics
 
         float priceMultipler = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find ("fEnchantmentValueMult")->mValue.getFloat();
         int price = MWBase::Environment::get().getMechanicsManager()->getBarterOffer(mEnchanter, static_cast<int>(getEnchantPoints() * priceMultipler), true);
+        int count = getEnchantItemsCount();
+        if (count > 1)
+            price *= count * 0.125f;
```
Capostrophic: The 1/8 multiplier seems to be arbitrary. Maybe put it into a setting?

docs/source/reference/modding/settings/game.rst
```diff
@@ -282,3 +282,19 @@ equivalent to the one introduced by the equivalent Morrowind Code Patch feature.
...
 This setting can be controlled in Advanced tab of the launcher.
+
+projectiles enchant multiplier
+------------------------------
+
+:Type:		floating point
+:Range:		0.0 to 1.0
+:Default:	0.0
+
+This setting determines which percentage of soulgem charge can be used to enchant multiple projectiles (thrown weapons, arrows and bolts):
```
Capostrophic: The value of this setting determines how many projectiles (thrown weapons, arrows and bolts) you can enchant at once according to the following formula:

docs/source/reference/modding/settings/game.rst
```diff
@@ -282,3 +282,19 @@ equivalent to the one introduced by the equivalent Morrowind Code Patch feature.
...
+
+count = gemCharge / enchantmentPower * projectilesEnchantMultiplier
```
Capostrophic: `count = (soul gem charge * projectiles enchant multiplier) / enchantment strength`

docs/source/reference/modding/settings/game.rst
```diff
@@ -282,3 +282,19 @@ equivalent to the one introduced by the equivalent Morrowind Code Patch feature.
...
+
+Zero value means that you can enchant only one projectile at once.
```
Capostrophic: A value of 0 means that you can only enchant one projectile.

docs/source/reference/modding/settings/game.rst
```diff
@@ -282,3 +282,19 @@ equivalent to the one introduced by the equivalent Morrowind Code Patch feature.
...
+If you want to enchant multiple projectiles at once, it is recommended to use value near 0.25 (25% of charge, as in MCP).
```
Capostrophic: If you want to have Morrowind Code Patch-like count of projectiles being enchanted at once, set this value to 0.25 (i.e. 25% of the charge).

docs/source/reference/modding/settings/game.rst
```diff
@@ -282,3 +282,19 @@ equivalent to the one introduced by the equivalent Morrowind Code Patch feature.
...
+
+This setting can be controlled in Advanced tab of the launcher.
```
Capostrophic: Hm, no, no it can't.

files/settings-default.cfg
```diff
@@ -267,6 +267,11 @@ use magic item animations = false
 # Don't use race weight in NPC movement speed calculations
 normalise race speed = false
 
+# Percent of soul gem charge, which can be used to enchant multiple projectiles:
+# count = gemCharge / enchantmentPower * projectilesEnchantMultiplier
```
Capostrophic: Adjusts the number of projectiles you can enchant at once.
Same suggestions as above.

apps/openmw/mwmechanics/enchanting.cpp
```diff
@@ -282,13 +341,33 @@ namespace MWMechanics
...
         if (mCastStyle == ESM::Enchantment::ConstantEffect)
             x *= fEnchantmentConstantChanceMult;
 
         return static_cast<int>(x);
     }
 
+    int Enchanting::getEnchantItemsCount() const
+    {
+        int count = 1;
+        int enchantPoints = getEnchantPoints(false);
```
Capostrophic: Do we want to use inaccurate enchantment strength or?

apps/openmw/mwmechanics/enchanting.cpp
kcat: Wouldn't this make a count of 8 equal to 1 in price, and 2 to 7 be cheaper than 1?

Capostrophic: I guess that this multiplier is in place because enchanting multiple arrows would have been outrageously expensive otherwise.

kcat: It may be better to scale the price based on the item type then, rather than the count. Otherwise, enchanting 1 arrow will be much more expensive than enchanting 2. And if there were to ever be a QoL enhancement of enchanting multiple objects at once (at an equal cost to doing one at a time, just more efficiently), this count check would break it.

Capostrophic: Oh right, that seems to be wrong.

docs/source/reference/modding/settings/game.rst
akortunov: Sorry, copy-paste error.

PR #2647, opened 2019-12-28
Do not use the QStringList::fromSet, which is deprecated since Qt 5.14

akortunov:
Just that.


AnyOldName3:
Is there any reason for the version check? Surely the non-deprecated path would work fine with older versions.

akortunov:
> Is there any reason for the version check?

A quote from Qt docs:

> Since Qt 5.14, range constructors are available for Qt's generic container classes and should be used in place of this method.

I have no idea if these constructors are available for old Qt versions, especially for Qt 4.x.

AnyOldName3:
Looks like they weren't available before, so the version check makes sense if there's no alternative option that's been around for longer.

Capostrophic:
Doesn't seem like there is.

PR #2649, opened 2019-12-30
Yet more NIF adjustments

Capostrophic:
In a way, a severely stripped down version of my NIF branch. This introduces some helpful stuff that makes it possible to add countless version checks and a lot of version-specific code to everywhere in the NIF model loading process. **Actual version checks are absent and no versions apart from 4.0.0.0** (with an obligatory asterisk) **and 4.0.0.2 are actually supported.**

* getBoolean() replaces getInt() where it should according to NifTools spec, for now it is in fact just a wrapper for it. Why is it necessary? In later versions of the format boolean type is stored in a single byte instead of 4 bytes, so there'll be a version check that toggles between these.
* Some typo fixes.
* More typedef directives were turned into using directives.
* Removed a lengthy irrelevant notice in NIF header loading, it'll become redundant anyway if all goes according to the plan.
* Some chosen file version and Bethesda version numbers are labelled in new public enums that can be used from anywhere in the loader instead of just the header loading. Currently largely unused.
* Added version check functionality! The file, user and bethesda version can be checked from anywhere in the code that works with the NIF file/stream. Of course user version and bethesda version aren't loaded yet, so they'll always be zero for the time being. Unused, but should be used a lot down the line.
* Ways to recover lists of integers and strings were added to the stream class. They'll be helpful in the future for loading the header, although there are more get methods where it came from that are of use in loading myriads of extra data records for example.
* Empty string table with no way to use it was added to the NIF file thing. I have version-specific code for string tables in the other branch, but they are a post-Oblivion addition.
* Interpolation was refactored. Now it's actually usable to introduce something that isn't linear interpolation and it is (IMO) more readable.
* For example, constant interpolation. NifSkope implementation was used as a reference, however it's the only logical way to do it. It's the only interpolation type added here. Although I figured out the principle of Bezier and TCB interpolation for scalars and vector, these types of interpolation are considerably more difficult to add for quaternions properly.
* Interpolation types are properly enumerated.
* Extra data lists and byte key map types field types. They're unused, but they'll be used (especially the extra list one).
* Double negation is avoided per akortunov's request.

So yeah, the only thing that end users may notice is the presence of constant interpolation.
This helpful graph from somewhere else demonstrates how it is supposed to work.
graph

...It's kind of the least useful interpolation type.

### DIFF
--- a/components/nif/nifstream.cpp
+++ b/components/nif/nifstream.cpp
@@ -24,4 +24,21 @@ namespace Nif
         t.scale = getFloat();
         return t;
     }
+
+    ///Currently specific for 4.0.0.2 and earlier
+    bool NIFStream::getBoolean()
+    {
+        return !!getInt();
+    }
+
+    ///Read in a string, either from the string table using the index (currently absent) or from the stream using the specified length
+    std::string NIFStream::getString()
+    {
+        return getSizedString();
+    }
+
+    // Convenience utility functions: get the versions of the currently read file
+    unsigned int NIFStream::getVersion() { return file->getVersion(); }
+    unsigned int NIFStream::getUserVersion() { return file->getBethVersion(); }
+    unsigned int NIFStream::getBethVersion()  { return file->getBethVersion(); }
 }
--- a/components/nif/nifstream.cpp
+++ b/components/nif/nifstream.cpp
@@ -28,7 +28,7 @@ namespace Nif
     ///Currently specific for 4.0.0.2 and earlier
     bool NIFStream::getBoolean()
     {
-        return !!getInt();
+        return getInt() != 0;
     }
 
     ///Read in a string, either from the string table using the index (currently absent) or from the stream using the specified length
--- a/apps/openmw_test_suite/nifloader/testbulletnifloader.cpp
+++ b/apps/openmw_test_suite/nifloader/testbulletnifloader.cpp
@@ -266,9 +266,13 @@ namespace
         MOCK_CONST_METHOD0(numRecords, std::size_t ());
         MOCK_CONST_METHOD1(getRoot, Nif::Record* (std::size_t));
         MOCK_CONST_METHOD0(numRoots, std::size_t ());
+        MOCK_CONST_METHOD1(getString, std::string (std::size_t));
         MOCK_METHOD1(setUseSkinning, void (bool));
         MOCK_CONST_METHOD0(getUseSkinning, bool ());
         MOCK_CONST_METHOD0(getFilename, std::string ());
+        MOCK_CONST_METHOD0(getVersion, unsigned int ());
+        MOCK_CONST_METHOD0(getUserVersion, unsigned int ());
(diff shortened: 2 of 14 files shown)


psi29a:
I liked this, did a quick review and saw nothing obviously weird.

components/nif/nifstream.cpp
```diff
@@ -24,4 +24,21 @@ namespace Nif
         t.scale = getFloat();
         return t;
     }
+
+    ///Currently specific for 4.0.0.2 and earlier
+    bool NIFStream::getBoolean()
+    {
+        return !!getInt();
```
akortunov: Why `!` is used twice?

akortunov:
Since this PR touches the low level of animations system, maybe it is worth to compare performance using timers.

components/nif/nifstream.cpp
Capostrophic: This is one of the explicit ways to avoid int-to-bool implicit conversion warnings which are currently silenced. Another way is a comparison to zero. !!getInt() is a boolean (not an integer) expression which means that the integer is not equal to zero. I didn't change it to anything because that's the way it was in the NIF code already.

akortunov: Are there more readable ways to avoid such warnings?

Capostrophic:
> to compare performance using timers.

While it's difficult to measure time of interpolation with timers, there doesn't seem to be any measurable difference in Balmora. According to past scrawl notes, Update time should be measured. For me it's 0.8-1 ms both without and with the changes during clear weather.

psi29a:
So no noticeable changes in performance? I would like to merge this. :)

akortunov:
Since this PR touches pretty many things and does not give any noticable difference for users, I'd prefer to have it in 0.47, so there will be an additional couple of years to test it after merge.

Capostrophic:
> many things

Apart from constant interpolation, there aren't actually any changes in the existing logic, and there's nothing much to test.

psi29a:
Executive veto.

components/nif/nifstream.cpp
```diff
@@ -24,4 +24,21 @@ namespace Nif
...
+
+    ///Read in a string, either from the string table using the index (currently absent) or from the stream using the specified length
+    std::string NIFStream::getString()
+    {
+        return getSizedString();
+    }
+
+    // Convenience utility functions: get the versions of the currently read file
+    unsigned int NIFStream::getVersion() { return file->getVersion(); }
+    unsigned int NIFStream::getUserVersion() { return file->getBethVersion(); }
```
elsid, line 42: Should this be? Otherwise what's the point of this function?
```c++
unsigned int NIFStream::getUserVersion() { return file->getUserVersion(); }
```

Capostrophic: You can tell the intention by the existence of userVer/getUserVersion in NIFFile


I guess there accidentally was never a case in testing where the user version mattered and most files worked by accident.

PR #2651, opened 2020-01-02
Unify streams usage to support non-ASCII paths

akortunov:
Fixes [bug #5239](https://gitlab.com/OpenMW/openmw/issues/5239).

Allows us to use the editor, importer and recast debugging with non-ASCII paths since boost streams support Unicode paths.

Tested on Windows with Russian characters in user's name.

### DIFF
--- a/apps/essimporter/importer.cpp
+++ b/apps/essimporter/importer.cpp
@@ -2,6 +2,7 @@
 
 #include <iomanip>
 
+#include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
 
 #include <osgDB/ReadFile>
@@ -349,7 +350,7 @@ namespace ESSImport
 
         writer.setFormat (ESM::SavedGame::sCurrentFormat);
 
-        std::ofstream stream(mOutFile.c_str(), std::ios::binary);
+        boost::filesystem::ofstream stream(boost::filesystem::path(mOutFile), std::ios::out | std::ios::binary);
         // all unused
         writer.setVersion(0);
         writer.setType(0);
--- a/apps/opencs/model/doc/document.cpp
+++ b/apps/opencs/model/doc/document.cpp
@@ -1,7 +1,6 @@
 #include "document.hpp"
 
 #include <cassert>
-#include <fstream>
 
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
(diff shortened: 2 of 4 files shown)


apps/opencs/model/doc/document.cpp
```diff
@@ -289,19 +288,23 @@ CSMDoc::Document::Document (const Files::ConfigurationManager& configuration,
...
+        if (!destination.is_open())
+            throw std::runtime_error("Can not create project file: " + mProjectPath.string());
+        destination.exceptions(std::ios::failbit | std::ios::badbit);
+
+        if (!boost::filesystem::exists (filtersPath))
+            filtersPath = boost::filesystem::path(mResDir.string() + "/defaultfilters");
+
+        boost::filesystem::ifstream source(filtersPath, std::ios::in | std::ios::binary);
+        if (!source.is_open())
+            throw std::runtime_error("Can not read filters file file: " + filtersPath.string());
```
Capostrophic: Seems like the second "file" is unnecessary.

apps/opencs/model/doc/document.cpp
```diff
@@ -289,19 +288,23 @@ CSMDoc::Document::Document (const Files::ConfigurationManager& configuration,
...
-            destination << std::ifstream(customFiltersPath.string().c_str(), std::ios::binary).rdbuf();
-        }
-        else
-        {
-            destination << std::ifstream(std::string(mResDir.string() + "/defaultfilters").c_str(), std::ios::binary).rdbuf();
-        }
+        boost::filesystem::path filtersPath (configuration.getUserDataPath());
+        filtersPath /= "defaultfilters";
+
+        boost::filesystem::ofstream destination(boost::filesystem::path(mProjectPath.string().c_str()), std::ios::out | std::ios::binary);
```
Capostrophic: Why is it necessary to convert mProjectPath from a boost::filesystem::path to
1) std::string
2) then to C-string
3) back to boost::filesystem::path?

Is it not possible to cut down on conversions here and in mResDir case?

apps/essimporter/importer.cpp
```diff
@@ -349,7 +350,7 @@ namespace ESSImport
 
         writer.setFormat (ESM::SavedGame::sCurrentFormat);
 
-        std::ofstream stream(mOutFile.c_str(), std::ios::binary);
+        boost::filesystem::ofstream stream(boost::filesystem::path(mOutFile.c_str()), std::ios::out | std::ios::binary);
```
Capostrophic: mOutFile is a string, it shouldn't be necessary to convert it to C string to construct a boost::filesystem::path here either.

psi29a:
Since we want to eventually get rid of as much boost as possible... this PR goes in the opposite direction.

Quoting from https://forum.vcmi.eu/t/supporting-unicode-file-paths/1025/4
> Internally, the boost::filesystem::basic_ifstream::open() method (and the constructor) that takes a boost::filesystem::path as an argument, converts the path to a c-string and passes it to std::basic_ifstream::open(). For non-windows systems, it gets passed as an UTF-8 encoded char*, which works perfectly fine. However, windows does not support opening files through an UTF-8 name. So on MSVC, it gets passed as an UTF-16 encoded wchar_t*, to a non-standard overload of std::basic_ifstream::open(), which takes a wchar_t* as an argument instead of a char*. But that overload does not exist on MinGW. So there, it ends up passing an UTF-8 encoded char*, even though windows does not support it. The code compiles just fine, but you get an error every time you try to open a file with a non-ascii name.

If this is true, then we can do this ourselves instead of just handing it back to boost::filesystem which will help us remove yet another boost dependency.

akortunov:
> Since we want to eventually get rid of as much boost as possible... this PR goes in the opposite direction.

No, we want to get rid of Boost where it does not break the functionality.
In this particular case our current behaviour in the editor is incorrect and this PR just makes it consistent with our codebase.
I see no point to implement custom streams if C++17 can handle Unicode paths.

psi29a:
Correct and we do not break functionality if we replace boost's ostream with our own version to specifically handle Windows brokenness when it comes with utf8.

I do not see a problem with future-proofing our code, since std::ostream behaviour on windows will continue to be broken after C++17.

So while std::filesystem (in C++17) is a close copy of boost::filesystem, it is missing boost::filesystem::ofstream which is what I'm talking about.
https://github.com/OpenMW/openmw/pull/2651/files#diff-d45cbd1087e92eadb64d59715fd54cfdR293

So while this PR fixes things now, the work I mention above will still have to be done.

akortunov:
Any objections left?

PR #2652, opened 2020-01-02
Read parameters of NiPathController

akortunov:
NiPathController is just a simplified version of NiPathInterpolator from GameBryo.
So it uses pretty same parameters (but does not have its own flags).


Capostrophic:
I don't see the value of this PR. That's one thing, another thing is that NiPathController does have its own flags, while followAxis is not a flag (it's the axis that may or may not get used, the flag itself is in the controller flags).

PR #2653, opened 2020-01-03
Improve BetaComment handling

akortunov:
Implements [feature #4129](https://gitlab.com/OpenMW/openmw/issues/4129).

Summary of changes:
1. Print BetaComment output to cout and openmw.log
2. Print report time in the same format as for savegames (but in UTC instead of local time):
```
Report time: 2020.01.06 04:43:31 UTC
```
Morrowind prints report time too.
3. Use a C++11-syile approach in the loading menu as well to unify used API and simplify the code.

Notes:
1. An another approach would be to use a separate file for BetaComments, but not definely a better one.
2. If we ever add timestamps to log, we will need to print report time only to debug console.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -196,6 +196,7 @@
     Feature #3980: In-game option to disable controller
     Feature #3999: Shift + Double Click should maximize/restore menu size
     Feature #4001: Toggle sneak controller shortcut
+    Feature #4129: Beta Comment to File
     Feature #4209: Editor: Faction rank sub-table
     Feature #4255: Handle broken RepairedOnMe script function
     Feature #4316: Implement RaiseRank/LowerRank functions properly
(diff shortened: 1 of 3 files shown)


Capostrophic:
In US, month and day positions are swapped.  May lead to confusion. Perhaps the year and the day should be switched around.

Cramal:
Year.Month.Day Hour:Minute:Seconds is a good option for writing time stamps in logs.

One big advantage is that alphabetical sorting is the same as chronological sorting.

AnyOldName3:
I'd go for the ISO format with bigger numbers at the front, too. No one's going to be confused by it and it's still readable.

akortunov:
As  I said, I just took the same format, which we use in the loading menu.

Screenshot_20200104_102606

IIRC, no one complained about it so far, despite Morrowind uses an American-style date format.

Capostrophic:
> As I said, I just took the same format, which we use in the loading menu.

Well then it's a good time to change it to the international standard, to minimize misunderstandings.

PR #2655, opened 2020-01-04
Fix reported spellcasting discrepancies

Capostrophic:
[5241](https://gitlab.com/OpenMW/openmw/issues/5241), [5242](https://gitlab.com/OpenMW/openmw/issues/5242).

Absorb spells are still applied to the target if the caster is the target, but their magnitude is zero. Hopefully this will allow a certain scripted sequence from TR to work.

ExplodeSpell now mostly works like Cast with the difference that the spell's target is the casting object/actor.

Cast can now work with permanent abilities like in vanilla again. For whatever reason you're able to cast them even though akortunov assumed otherwise.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -554,16 +554,11 @@ namespace MWMechanics
 
                     // Avoid applying absorb effects if the caster is the target
                     // We still need the spell to be added
-                    if (caster == target)
+                    if (caster == target
+                        && effectIt->mEffectID >= ESM::MagicEffect::AbsorbAttribute
+                        && effectIt->mEffectID <= ESM::MagicEffect::AbsorbSkill)
                     {
-                        for (int i=0; i<5; ++i)
-                        {
-                            if (effectIt->mEffectID == ESM::MagicEffect::AbsorbAttribute+i)
-                            {
-                                effect.mMagnitude = 0;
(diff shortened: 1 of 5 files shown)


akortunov:
> ExplodeSpell now mostly works like Cast

In this case it would be better to use a shared code to reduce code duplication.

Capostrophic:
The code is not identical though, and there might be more nuances to each instruction we're not aware of that will need the code to be "duplicated back".

akortunov:
> there might be more nuances to each instruction we're not aware of that will need the code to be "duplicated back"

In this case it will not be hard to do a copy-paste and then handle such nuances.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -552,6 +552,20 @@ namespace MWMechanics
                     effect.mArg = MWMechanics::EffectKey(*effectIt).mArg;
                     effect.mMagnitude = magnitude;
 
+                    // Avoid applying absorb effects if the caster is the target
+                    // We still need the spell to be added
+                    if (caster == target)
+                    {
+                        for (int i=0; i<5; ++i)
```
akortunov: I suppose the loop is redundant here:

```
if (effectIt->mEffectID >= AbsorbAttribute && effectIt->mEffectID <= AbsorbSkill)
    effect.mMagnitude = 0;
```

PR #2656, opened 2020-01-05
Don't combine sneak idle with scripted/wander idles (bug #4284)

Capostrophic:
[Bug report.](https://gitlab.com/OpenMW/openmw/issues/4284)

Not switching the idle state seems to fix the issue in the general case. Not sure if it doesn't have side effects.


akortunov:
The `CharState_SpecialIdle` is used mostly for scripted animations. Did you check if the behaviour is correct here?

Capostrophic:
I did. Scripted animations aren't combined with the sneak idle.

PR #2657, opened 2020-01-05
BulletNifLoader: Reduce exposed loader API to a single function

ususdei:
This patch minimizes the exposed API and state of the `BulletNifLoader` in `bulletnifloader.hpp`.

Everything not intended for the exposed API (including a helper class to hold the state) is moved into the .cpp file.
Admittedly, this could be considered a style-change only, since it ~~does~~ should not change any observable behaviour (aside from some cases of brutal misuse like reusing the same loader instance in multiple threads e.g.)
If you would rather not accept PRs like this without observable fixes, feel free to just close it.

The `BulletNifLoader` is already testet extensively in `testbulletnifloader.cpp`, therefore I feel rather confident in not breaking anything after running these and playing the game for a while.


elsid:
The idea to keep class definition in header is to make it usable from anywhere. Putting it into a cpp file makes it harder to add new features and add new tests. Considering this class as an API makes things complicated. It's not an API. There is no any client for it except engine itself.

If you want to make an API there is no need to move definition into cpp file. Add a new header, define API classes and declare functions, add a new cpp file, put there functions definitions and include the header with a class/function you want to provide an API for. But I would ask you please don't do this for this specific class, there is simply no need for it unless you explain one that I don't see.

ususdei:
Thanks for your feedback, although I cannot really follow all of your points:

1) The class is not meant to be used directly - it is only there to hold temporary state during a single load.
2) I would consider this header to be a piece of the component-layer-API, intended for use by upper layers
3) This is now very similiar to the implementation of `NifOsg::Loader`

Nonetheless, I don't want to steal anybodys time by arguing about stylistic and architectural preferences, so:
PR be gone!

/closed due to lack of usefulness

PR #2659, opened 2020-01-06
Handle out-of-range actors' travel packages (bug #5212)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5212)

AiTravel and internal travel packages of actors that are out of processing range are now handled. The observable effect in-game is that travelling actors in all loaded cells will move while they are beyond the range and appear from "nowhere" when they become closer to the player. Their position changes as expected.

Of course their physics/animations/whatever are still updated, since they need to actually move... So most of the benefit from reducing AI processing range is nullified for such actors. But they're very rare.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -1728,9 +1728,14 @@ namespace MWMechanics
                 const float dist = (playerPos - iter->first.getRefData().getPosition().asVec3()).length();
                 bool isPlayer = iter->first == player;
                 CreatureStats &stats = iter->first.getClass().getCreatureStats(iter->first);
-                int packageId = stats.getAiSequence().getTypeId();
-                bool travelling = (packageId == AiPackage::TypeIdTravel) || (packageId == AiPackage::TypeIdInternalTravel);
-                bool inRange = isPlayer || dist <= mActorsProcessingRange || travelling;
+                // Actors with active AI should be able to move.
+                bool alwaysActive = false;
+                if (!isPlayer && isConscious(iter->first) && !stats.isParalyzed())
+                {
+                    MWMechanics::AiSequence& seq = stats.getAiSequence();
+                    alwaysActive = !seq.isEmpty() && seq.getActivePackage()->alwaysActive();
+                }
+                bool inRange = isPlayer || dist <= mActorsProcessingRange || alwaysActive;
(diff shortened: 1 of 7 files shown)


akortunov:
I suppose the docs about `actors processing range` setting should be updated as well.

apps/openmw/mwmechanics/actors.cpp
```diff
@@ -1722,7 +1727,10 @@ namespace MWMechanics
             {
                 const float dist = (playerPos - iter->first.getRefData().getPosition().asVec3()).length();
                 bool isPlayer = iter->first == player;
-                bool inRange = isPlayer || dist <= mActorsProcessingRange;
+                CreatureStats &stats = iter->first.getClass().getCreatureStats(iter->first);
+                int packageId = stats.getAiSequence().getTypeId();
+                bool travelling = (packageId == AiPackage::TypeIdTravel) || (packageId == AiPackage::TypeIdInternalTravel);
```
akortunov: I'd prefer to have a more generic "dependent on distance" flag in AI packages themselves.

akortunov:
```
In file included from /home/travis/build/OpenMW/openmw/apps/openmw/mwmechanics/aisequence.cpp:14:
/home/travis/build/OpenMW/openmw/apps/openmw/mwmechanics/aitravel.hpp:37:26: warning: 'MWMechanics::AiTravel::alwaysActive' hides overloaded virtual function [-Woverloaded-virtual]
            virtual bool alwaysActive() const { return true; }
                         ^
/home/travis/build/OpenMW/openmw/apps/openmw/mwmechanics/aipackage.hpp:108:26: note: hidden overloaded virtual function 'MWMechanics::AiPackage::alwaysActive' declared here: different qualifiers (none vs const)
            virtual bool alwaysActive() { return false; }
```

apps/openmw/mwmechanics/aipackage.hpp
```diff
@@ -104,6 +104,9 @@ namespace MWMechanics
 
             virtual osg::Vec3f getDestination() { return osg::Vec3f(0, 0, 0); }
 
+            // Return true if any loaded actor with this AI package must be active.
+            virtual bool alwaysActive() { return false; }
```
akortunov: I suppose you missed the `const` modifier here, what leads to warnings.

Capostrophic: Huh, how could I miss it. Fixing now.

PR #2660, opened 2020-01-09
Additional de-boosting

akortunov:
Summary of changes:
1. Implement strings trimming, splitting and substring replacement to get rid of boost::algorithm.
2. Do not link Boost threads library since we do not use it now.

### DIFF
--- a/components/misc/stringops.hpp
+++ b/components/misc/stringops.hpp
@@ -1,6 +1,7 @@
 #ifndef MISC_STRINGOPS_H
 #define MISC_STRINGOPS_H
 
+#include <cctype>
 #include <string>
 #include <algorithm>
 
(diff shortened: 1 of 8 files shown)


components/misc/stringops.hpp
```diff
@@ -245,6 +246,58 @@ class StringUtils
...
+        while (current != std::string::npos)
+        {
+            cont.push_back(str.substr(previous, current - previous));
+            previous = current + 1;
+            current = str.find_first_of(delims, previous);
+        }
+        cont.push_back(str.substr(previous, current - previous));
+    }
+
+    // TODO: use the std::string_view once we will use the C++17.
```
psi29a: Thank you! :)

PR #2662, opened 2020-01-10
Use shield body part model for creatures (bug #5250)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5250)

According to rot, both biped and non-biped creatures use the body part model of the shield in vanilla. 

So I added a look up for that model which will be used instead of the ground model when possible. Shields seem to look fine now.


akortunov:
I suppose you will also rework the `ActorAnimation::getShieldMesh()` function for shields holstering. 
Now it assumes that only NPC's use shield bodyparts. If it is not true, the `NpcAnimation::getShieldMesh()` should be moved to the `ActorAnimation`.

Capostrophic:
screenshot
Seems to work. Can't help code duplication, creatures don't have a gender.

PR #2665, opened 2020-01-12
Separate grass handling

akortunov:
This PR handles grass objects from enabled plugins separately from other objects.

Also there is an [another](https://github.com/akortunov/openmw/tree/grass_paging) implementation, which is based on ObjectPaging (unfortunately, grass animations do not work).

Summary of changes:
1. Attach grass nodes directly to cell node, without creating a cellref and animation object just for every grass shape. As a result, grass almost does not affect cell loading speed, does not have collisions and does not block objects (e.g. corpses).
2. Use a separate mask for grass nodes, so grass objects do not cast shadows (but still receive them) and do not present in reflections. It allows to use grass when shadows and/or object reflections are enabled.
3. Implemented a grass density setting. A density algorithm is deterministic, so the grass should be in the same place every time when the same density is used.
4. Implemented a grass rendering distance setting. Grass objects near rendering border fade away. Allows to increase performance and avoid pop-ins. Requires an AlphaProperty for groundcover objects.
5. Grass is animated via separate shader - it takes in account wind speed from openmw.cfg and player's position to emulate stomping.

It is recommended to test this PR with the [Aesthesia Groundcover](https://www.nexusmods.com/morrowind/mods/46377?tab=files).

An example of settings:

```
[Grass]
enabled = true
fade start = 0.75
density = 0.3
distance = 6144
animation = true
```

Known limitations:
1. There is no batching system, so large grass density still significally increases CPU loading and may lead to noticable performance drops on Draw-limited setups.
2. MGE uses custom lighting shaders (some grass replacers even use custom shaders), so some MGE-specific grass mods may do not work as intended. 
3. Some mods may add groundcover objects, which are not supposed to be animated (stones, mushrooms, etc). Such objects also have no alpha property, so fading does not work for them too.
4. Grass is rendered/animated only for active grid.

Screenshot_20200112_230045

The main question - can we use such system, or our current approach is to do nothing and wait until a skilled OpenGL developer implements a procedure-generated grass for us for free?

### DIFF
--- a/apps/openmw/mwrender/objects.cpp
+++ b/apps/openmw/mwrender/objects.cpp
@@ -8,6 +8,7 @@
 
 #include "../mwworld/ptr.hpp"
 #include "../mwworld/class.hpp"
+#include "../mwworld/cellstore.hpp"
 
 #include "animation.hpp"
 #include "npcanimation.hpp"
@@ -66,6 +67,24 @@ void Objects::insertBegin(const MWWorld::Ptr& ptr)
     insert->setScale(scaleVec);
 
     ptr.getRefData().setBaseNode(insert);
+}
+
+void Objects::insertGrass(MWWorld::CellStore* cell)
+{
+    osg::ref_ptr<osg::Group> cellnode;
+
+    CellMap::iterator found = mCellSceneNodes.find(cell);
+    if (found == mCellSceneNodes.end())
+    {
+        cellnode = new osg::Group;
+        cellnode->setName("Cell Root");
+        mRootNode->addChild(cellnode);
+        mCellSceneNodes[cell] = cellnode;
+    }
+    else
+        cellnode = found->second;
+
+    cell->insertGrass(cellnode, mResourceSystem);
 }
 
 void Objects::insertModel(const MWWorld::Ptr &ptr, const std::string &mesh, bool animated, bool allowLight)
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -153,6 +153,11 @@ void main()
 #endif
 
     float shadowing = unshadowedLightRatio(linearDepth);
+    if (isGrass)
+    {
+        if (euclideanDepth > @grassFadeStart)
+            gl_FragData[0].a *= 1.0-smoothstep(@grassFadeStart, @grassFadeEnd, euclideanDepth);
+    }
 
 #if !PER_PIXEL_LIGHTING
 
@@ -163,7 +168,8 @@ void main()
 #endif
 
 #else
-    gl_FragData[0] *= doLighting(passViewPos, normalize(viewNormal), passColor, shadowing, isGrass);
+    if(gl_FragData[0].a != 0.0)
+        gl_FragData[0] *= doLighting(passViewPos, normalize(viewNormal), passColor, shadowing, isGrass);
 #endif
 
 #if @envMap && !@preLightEnv
@@ -208,10 +214,4 @@ void main()
     gl_FragData[0].xyz = mix(gl_FragData[0].xyz, gl_Fog.color.xyz, fogValue);
 
     applyShadowDebugOverlay();
-
-    if (isGrass)
-    {
-        if (euclideanDepth > @grassFadeStart)
-            gl_FragData[0].a *= 1.0-smoothstep(@grassFadeStart, @grassFadeEnd, euclideanDepth);
-    }
 }
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -51,8 +51,6 @@ uniform mat2 bumpMapMatrix;
 
 uniform bool simpleWater;
 
-uniform bool isGrass;
-
 varying float euclideanDepth;
 varying float linearDepth;
 
@@ -153,11 +151,6 @@ void main()
 #endif
 
     float shadowing = unshadowedLightRatio(linearDepth);
-    if (isGrass)
-    {
-        if (euclideanDepth > @grassFadeStart)
-            gl_FragData[0].a *= 1.0-smoothstep(@grassFadeStart, @grassFadeEnd, euclideanDepth);
-    }
 
 #if !PER_PIXEL_LIGHTING
 
@@ -168,8 +161,7 @@ void main()
 #endif
 
 #else
-    if(gl_FragData[0].a != 0.0)
-        gl_FragData[0] *= doLighting(passViewPos, normalize(viewNormal), passColor, shadowing, isGrass);
+    gl_FragData[0] *= doLighting(passViewPos, normalize(viewNormal), passColor, shadowing, false);
 #endif
 
 #if @envMap && !@preLightEnv
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -163,7 +163,7 @@ void main()
 #endif
 
 #else
-    gl_FragData[0] *= doLighting(passViewPos, normalize(viewNormal), passColor, shadowing);
+    gl_FragData[0] *= doLighting(passViewPos, normalize(viewNormal), passColor, shadowing, isGrass);
 #endif
 
 #if @envMap && !@preLightEnv
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -51,6 +51,10 @@ uniform mat2 bumpMapMatrix;
 
 uniform bool simpleWater;
 
+#if @grassAnimation
+uniform bool isGrass;
+#endif
+
 varying float euclideanDepth;
 varying float linearDepth;
 
@@ -206,4 +210,10 @@ void main()
     gl_FragData[0].xyz = mix(gl_FragData[0].xyz, gl_Fog.color.xyz, fogValue);
 
     applyShadowDebugOverlay();
+
+    if (isGrass)
+    {
+        if (euclideanDepth > @grassFadeStart)
+            gl_FragData[0].a *= 1.0-smoothstep(@grassFadeStart, @grassFadeEnd, euclideanDepth);
+    }
 }
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -51,9 +51,7 @@ uniform mat2 bumpMapMatrix;
 
 uniform bool simpleWater;
 
-#if @grassAnimation
 uniform bool isGrass;
-#endif
 
 varying float euclideanDepth;
 varying float linearDepth;
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -21,7 +21,7 @@ add_openmw_dir (mwrender
     actors objects renderingmanager animation rotatecontroller sky npcanimation vismask
     creatureanimation effectmanager util renderinginterface pathgrid rendermode weaponanimation
     bulletdebugdraw globalmap characterpreview camera viewovershoulder localmap water terrainstorage ripplesimulation
-    renderbin actoranimation landmanager navmesh actorspaths recastmesh fogmanager objectpaging
+    renderbin actoranimation landmanager navmesh actorspaths recastmesh fogmanager objectpaging grass
     )
 
 add_openmw_dir (mwinput
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -454,7 +454,8 @@ namespace MWBase
             ///< Apply a health difference to any actors colliding with \a object.
             /// To hurt actors, healthPerSecond should be a positive value. For a negative value, actors will be healed.
 
-            virtual float getWindSpeed() = 0;
+            virtual float getBaseWindSpeed() const = 0;
+            virtual float getWindSpeed() const = 0;
 
             virtual void getContainersOwnedBy (const MWWorld::ConstPtr& npc, std::vector<MWWorld::Ptr>& out) = 0;
             ///< get all containers in active cells owned by this Npc
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -2,7 +2,9 @@
 
 #include <components/misc/stringops.hpp>
 
-#include <components/sceneutil/lightmanager.hpp>
+#include <components/resource/resourcesystem.hpp>
+#include <components/resource/scenemanager.hpp>
+
 #include <components/sceneutil/positionattitudetransform.hpp>
 
 #include <components/settings/settings.hpp>
@@ -13,6 +15,19 @@
 
 namespace MWRender
 {
+    Grass::Grass()
+    {
+        blank();
+        mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
+        mUseAnimation = Settings::Manager::getBool("animation", "Grass");
+    }
+
+    void Grass::blank()
+    {
+        mItems.clear();
+        mCurrentGrass = 0;
+    }
+
     void Grass::update()
     {
         osg::Vec3f playerPos(MWMechanics::getPlayer().getRefData().getPosition().asVec3());
@@ -21,8 +36,7 @@ namespace MWRender
             item.updateVisibility(playerPos);
         }
 
-        const static bool useAnimation = Settings::Manager::getBool("animation", "Grass");
-        if (!useAnimation)
+        if (!mUseAnimation)
             return;
 
         float windSpeed = MWBase::Environment::get().getWorld()->getBaseWindSpeed();
@@ -32,9 +46,21 @@ namespace MWRender
 
     void Grass::insertGrass(osg::Group* cellnode, Resource::ResourceSystem* rs)
     {
+        osg::ref_ptr<osg::Group> grassGroup = new osg::Group();
+        grassGroup->setName("CellGrass");
+        cellnode->addChild(grassGroup);
+
         for (MWRender::GrassItem& item : mItems)
         {
-            attachToNode(item, cellnode, rs);
+            attachToNode(item, grassGroup, rs);
+        }
+
+        rs->getSceneManager()->recreateShaders(grassGroup, "grass");
+
+        if (mUseAnimation)
+        {
+            osg::StateSet* stateset = grassGroup->getOrCreateStateSet();
+            stateset->addUniform(mWindSpeedUniform.get());
         }
     }
 
@@ -54,16 +80,10 @@ namespace MWRender
 
         rs->getSceneManager()->getInstance("meshes\\" + item.mModel, insert);
 
-        osg::StateSet* stateset = insert->getOrCreateStateSet();
-        // @grass preprocessor define would be great
-        stateset->addUniform(mIsGrassUniform);
-
-        const static bool useAnimation = Settings::Manager::getBool("animation", "Grass");
-        if(useAnimation)
+        if (mUseAnimation)
         {
-            // for some reason this uniform is added to other objects too? not only for grass
-            stateset->addUniform(new osg::Uniform("Rotz", (float) item.mPos.rot[2]));
-            stateset->addUniform(mWindSpeedUniform.get());
+            osg::StateSet* stateset = insert->getOrCreateStateSet();
+            stateset->addUniform(new osg::Uniform("Rotz", item.mPos.rot[2]));
         }
 
         item.mNode = insert;
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -53,7 +53,6 @@ namespace MWRender
             return;
 
         float windSpeed = MWBase::Environment::get().getWorld()->getBaseWindSpeed();
-        windSpeed *= 4.f; // Note: actual wind speed is usually larger than base one from config
         mWindSpeedUpdater->setWindSpeed(windSpeed);
     }
 
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -96,6 +96,17 @@ namespace MWRender
         stateset->addUniform(new osg::Uniform("colorMode", 0), osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
         stateset->setAttributeAndModes(defaultMat, osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
 
+        const static bool useGrass = Settings::Manager::getBool("animation", "Grass");
+        if(useGrass)
+        {
+            // @grass preprocessor define would be great
+            stateset->addUniform(new osg::Uniform("isGrass", true));
+            // for some reason this uniform is added to other objects too? not only for grass
+            mStormDirectionUniform = new osg::Uniform("WindDirection", (osg::Vec3f) osg::Vec3f(0.0, 0.0, 0.0));
+            stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
+            stateset->addUniform(mStormDirectionUniform.get());
+        }
+
         mNode = insert;
     }
 
@@ -126,6 +137,15 @@ namespace MWRender
         float fadeRatio = (dist - fadeStartDistance)/(fadeEndDistance - fadeStartDistance);
         if (fadeRatio > 0)
             visibilityRatio -= std::max(0.f, fadeRatio);
+
+        // having its own updateUniform function would be better, need to update it even for non visible grass
+        const static bool useGrass = Settings::Manager::getBool("animation", "Grass");
+        if (useGrass)
+        {
+            // TODO: actually support storm
+            float windSpeed = MWBase::Environment::get().getWorld()->getBaseWindSpeed();
+            mStormDirectionUniform->set((osg::Vec3f) osg::Vec3f(0, 0, windSpeed));
+        }
 
         visibilityRatio = std::min(1.f, visibilityRatio);
 
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -123,12 +123,13 @@ namespace MWRender
 
         rs->getSceneManager()->getInstance("meshes\\" + mModel, insert);
 
+        osg::StateSet* stateset = insert->getOrCreateStateSet();
+        // @grass preprocessor define would be great
+        stateset->addUniform(new osg::Uniform("isGrass", true));
+
         const static bool useAnimation = Settings::Manager::getBool("animation", "Grass");
         if(useAnimation)
         {
-            osg::StateSet* stateset = insert->getOrCreateStateSet();
-            // @grass preprocessor define would be great
-            stateset->addUniform(new osg::Uniform("isGrass", true));
             // for some reason this uniform is added to other objects too? not only for grass
             mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
             stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -118,8 +118,6 @@ namespace MWRender
             osg::Quat(mPos.rot[0], osg::Vec3(-1, 0, 0)));
 
         insert->setNodeMask(Mask_Grass);
-        osg::ref_ptr<SceneUtil::LightListCallback> lightListCallback = new SceneUtil::LightListCallback;
-        insert->addCullCallback(lightListCallback);
 
         rs->getSceneManager()->getInstance("meshes\\" + mModel, insert);
 
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -102,9 +102,9 @@ namespace MWRender
             // @grass preprocessor define would be great
             stateset->addUniform(new osg::Uniform("isGrass", true));
             // for some reason this uniform is added to other objects too? not only for grass
-            mStormDirectionUniform = new osg::Uniform("WindDirection", (osg::Vec3f) osg::Vec3f(0.0, 0.0, 0.0));
+            mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
             stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
-            stateset->addUniform(mStormDirectionUniform.get());
+            stateset->addUniform(mWindSpeedUniform.get());
         }
 
         mNode = insert;
@@ -142,9 +142,9 @@ namespace MWRender
         const static bool useGrass = Settings::Manager::getBool("animation", "Grass");
         if (useGrass)
         {
-            // TODO: actually support storm
             float windSpeed = MWBase::Environment::get().getWorld()->getBaseWindSpeed();
-            mStormDirectionUniform->set((osg::Vec3f) osg::Vec3f(0, 0, windSpeed));
+            windSpeed *= 4.f; // Note: actual wind speed is usually larger than base one from config
+            mWindSpeedUniform->set(windSpeed);
         }
 
         visibilityRatio = std::min(1.f, visibilityRatio);
@@ -178,9 +178,9 @@ namespace MWRender
                 // @grass preprocessor define would be great
                 stateset->addUniform(new osg::Uniform("isGrass", true));
                 // for some reason this uniform is added to other objects too? not only for grass
-                mStormDirectionUniform = new osg::Uniform("WindDirection", (osg::Vec3f) osg::Vec3f(0.0, 0.0, 0.0));
+                mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
                 stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
-                stateset->addUniform(mStormDirectionUniform.get());
+                stateset->addUniform(mWindSpeedUniform.get());
             }
             mNode->removeUpdateCallback(mAlphaUpdater);
             mAlphaUpdater = nullptr;
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -13,12 +13,21 @@
 
 namespace MWRender
 {
-    void Grass::updateGrassVisibility()
+    void Grass::update()
     {
         osg::Vec3f playerPos(MWMechanics::getPlayer().getRefData().getPosition().asVec3());
         for (MWRender::GrassItem& item : mItems)
         {
             item.updateVisibility(playerPos);
+        }
+
+        const static bool useAnimation = Settings::Manager::getBool("animation", "Grass");
+        if (!useAnimation)
+            return;
+
+        for (MWRender::GrassItem& item : mItems)
+        {
+            item.update();
         }
     }
 
@@ -70,44 +79,12 @@ namespace MWRender
         return false;
     }
 
-    void GrassItem::attachToNode(osg::Group* cellnode, Resource::ResourceSystem* rs)
+    void GrassItem::update()
     {
-        osg::ref_ptr<SceneUtil::PositionAttitudeTransform> insert (new SceneUtil::PositionAttitudeTransform);
-        cellnode->addChild(insert);
-
-        insert->setPosition(mPos.asVec3());
-        insert->setScale(osg::Vec3f(mScale, mScale, mScale));
-        insert->setAttitude(
-            osg::Quat(mPos.rot[2], osg::Vec3(0, 0, -1)) *
-            osg::Quat(mPos.rot[1], osg::Vec3(0, -1, 0)) *
-            osg::Quat(mPos.rot[0], osg::Vec3(-1, 0, 0)));
-
-        insert->setNodeMask(Mask_Grass);
-        osg::ref_ptr<SceneUtil::LightListCallback> lightListCallback = new SceneUtil::LightListCallback;
-        insert->addCullCallback(lightListCallback);
-
-        rs->getSceneManager()->getInstance("meshes\\" + mModel, insert);
-
-        osg::ref_ptr<osg::Material> defaultMat (new osg::Material);
-        defaultMat->setColorMode(osg::Material::OFF);
-        defaultMat->setAmbient(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
-        defaultMat->setDiffuse(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
-        osg::StateSet* stateset = insert->getOrCreateStateSet();
-        stateset->addUniform(new osg::Uniform("colorMode", 0), osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
-        stateset->setAttributeAndModes(defaultMat, osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
-
-        const static bool useGrass = Settings::Manager::getBool("animation", "Grass");
-        if(useGrass)
-        {
-            // @grass preprocessor define would be great
-            stateset->addUniform(new osg::Uniform("isGrass", true));
-            // for some reason this uniform is added to other objects too? not only for grass
-            mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
-            stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
-            stateset->addUniform(mWindSpeedUniform.get());
-        }
-
-        mNode = insert;
+        // having its own updateUniform function would be better, need to update it even for non visible grass
+        float windSpeed = MWBase::Environment::get().getWorld()->getBaseWindSpeed();
+        windSpeed *= 4.f; // Note: actual wind speed is usually larger than base one from config
+        mWindSpeedUniform->set(windSpeed);
     }
 
     void GrassItem::updateVisibility(osg::Vec3f& playerPos)
@@ -126,70 +103,38 @@ namespace MWRender
         }
         else
             mNode->setNodeMask(Mask_Grass);
-
-        static const float fadeStart = std::max(0.f, Settings::Manager::getFloat("fade start", "Grass"));
-        if (fadeStart >= 1.f)
-            return;
-
-        float visibilityRatio = 1.0;
-        float fadeStartDistance = grassDistance*fadeStart;
-        float fadeEndDistance = grassDistance;
-        float fadeRatio = (dist - fadeStartDistance)/(fadeEndDistance - fadeStartDistance);
-        if (fadeRatio > 0)
-            visibilityRatio -= std::max(0.f, fadeRatio);
-
-        // having its own updateUniform function would be better, need to update it even for non visible grass
-        const static bool useGrass = Settings::Manager::getBool("animation", "Grass");
-        if (useGrass)
-        {
-            float windSpeed = MWBase::Environment::get().getWorld()->getBaseWindSpeed();
-            windSpeed *= 4.f; // Note: actual wind speed is usually larger than base one from config
-            mWindSpeedUniform->set(windSpeed);
-        }
-
-        visibilityRatio = std::min(1.f, visibilityRatio);
-
-        setAlpha(visibilityRatio);
     }
 
-    void GrassItem::setAlpha(float alpha)
+    void GrassItem::attachToNode(osg::Group* cellnode, Resource::ResourceSystem* rs)
     {
-        if (mVisibility == alpha)
-            return;
+        osg::ref_ptr<SceneUtil::PositionAttitudeTransform> insert (new SceneUtil::PositionAttitudeTransform);
+        cellnode->addChild(insert);
 
-        mVisibility = alpha;
-        if (alpha != 1.f)
+        insert->setPosition(mPos.asVec3());
+        insert->setScale(osg::Vec3f(mScale, mScale, mScale));
+        insert->setAttitude(
+            osg::Quat(mPos.rot[2], osg::Vec3(0, 0, -1)) *
+            osg::Quat(mPos.rot[1], osg::Vec3(0, -1, 0)) *
+            osg::Quat(mPos.rot[0], osg::Vec3(-1, 0, 0)));
+
+        insert->setNodeMask(Mask_Grass);
+        osg::ref_ptr<SceneUtil::LightListCallback> lightListCallback = new SceneUtil::LightListCallback;
+        insert->addCullCallback(lightListCallback);
+
+        rs->getSceneManager()->getInstance("meshes\\" + mModel, insert);
+
+        const static bool useAnimation = Settings::Manager::getBool("animation", "Grass");
+        if(useAnimation)
         {
-            if (mAlphaUpdater == nullptr)
-            {
-                mAlphaUpdater = new MWRender::GrassUpdater(alpha);
-                mNode->addUpdateCallback(mAlphaUpdater);
-            }
-            else
-                mAlphaUpdater->setAlpha(alpha);
+            osg::StateSet* stateset = insert->getOrCreateStateSet();
+            // @grass preprocessor define would be great
+            stateset->addUniform(new osg::Uniform("isGrass", true));
+            // for some reason this uniform is added to other objects too? not only for grass
+            mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
+            stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
+            stateset->addUniform(mWindSpeedUniform.get());
         }
-        else
-        {
-            if (mNode->getStateSet())
-            {
-                // FIXME: probably there is a better way to reset stateset
-                mNode->setStateSet(nullptr);
-                osg::StateSet* stateset = mNode->getOrCreateStateSet();
-                // @grass preprocessor define would be great
-                stateset->addUniform(new osg::Uniform("isGrass", true));
-                // for some reason this uniform is added to other objects too? not only for grass
-                mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
-                stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
-                stateset->addUniform(mWindSpeedUniform.get());
-                osg::ref_ptr<osg::Material> defaultMat (new osg::Material);
-                defaultMat->setColorMode(osg::Material::OFF);
-                defaultMat->setAmbient(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
-                defaultMat->setDiffuse(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
-                stateset->addUniform(new osg::Uniform("colorMode", 0), osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
-                stateset->setAttributeAndModes(defaultMat, osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
-            }
-            mNode->removeUpdateCallback(mAlphaUpdater);
-            mAlphaUpdater = nullptr;
-        }
+
+        mNode = insert;
     }
 }
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -15,10 +15,23 @@
 
 namespace MWRender
 {
+    void WindSpeedUpdater::setDefaults(osg::StateSet *stateset)
+    {
+        osg::ref_ptr<osg::Uniform> windUniform = new osg::Uniform("windSpeed", 0.0f);
+        stateset->addUniform(windUniform.get());
+    }
+
+    void WindSpeedUpdater::apply(osg::StateSet *stateset, osg::NodeVisitor *nv)
+    {
+        osg::ref_ptr<osg::Uniform> windUniform = stateset->getUniform("windSpeed");
+        if (windUniform != nullptr)
+            windUniform->set(mWindSpeed);
+    }
+
     Grass::Grass()
     {
         blank();
-        mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
+        mWindSpeedUpdater = new WindSpeedUpdater;
         mUseAnimation = Settings::Manager::getBool("animation", "Grass");
     }
 
@@ -41,7 +54,7 @@ namespace MWRender
 
         float windSpeed = MWBase::Environment::get().getWorld()->getBaseWindSpeed();
         windSpeed *= 4.f; // Note: actual wind speed is usually larger than base one from config
-        mWindSpeedUniform->set(windSpeed);
+        mWindSpeedUpdater->setWindSpeed(windSpeed);
     }
 
     void Grass::insertGrass(osg::Group* cellnode, Resource::ResourceSystem* rs)
@@ -59,8 +72,7 @@ namespace MWRender
 
         if (mUseAnimation)
         {
-            osg::StateSet* stateset = grassGroup->getOrCreateStateSet();
-            stateset->addUniform(mWindSpeedUniform.get());
+            grassGroup->addUpdateCallback(mWindSpeedUpdater);
         }
     }
 
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -15,23 +15,30 @@
 
 namespace MWRender
 {
-    void WindSpeedUpdater::setDefaults(osg::StateSet *stateset)
+    void GrassUpdater::setDefaults(osg::StateSet *stateset)
     {
         osg::ref_ptr<osg::Uniform> windUniform = new osg::Uniform("windSpeed", 0.0f);
         stateset->addUniform(windUniform.get());
+
+        osg::ref_ptr<osg::Uniform> playerPosUniform = new osg::Uniform("playerPos", osg::Vec3f(0.f, 0.f, 0.f));
+        stateset->addUniform(playerPosUniform.get());
     }
 
-    void WindSpeedUpdater::apply(osg::StateSet *stateset, osg::NodeVisitor *nv)
+    void GrassUpdater::apply(osg::StateSet *stateset, osg::NodeVisitor *nv)
     {
         osg::ref_ptr<osg::Uniform> windUniform = stateset->getUniform("windSpeed");
         if (windUniform != nullptr)
             windUniform->set(mWindSpeed);
+
+        osg::ref_ptr<osg::Uniform> playerPosUniform = stateset->getUniform("playerPos");
+        if (playerPosUniform != nullptr)
+            playerPosUniform->set(mPlayerPos);
     }
 
     Grass::Grass()
     {
         blank();
-        mWindSpeedUpdater = new WindSpeedUpdater;
+        mGrassUpdater = new GrassUpdater;
         mUseAnimation = Settings::Manager::getBool("animation", "Grass");
     }
 
@@ -53,7 +60,8 @@ namespace MWRender
             return;
 
         float windSpeed = MWBase::Environment::get().getWorld()->getBaseWindSpeed();
-        mWindSpeedUpdater->setWindSpeed(windSpeed);
+        mGrassUpdater->setWindSpeed(windSpeed);
+        mGrassUpdater->setPlayerPos(playerPos);
     }
 
     void Grass::insertGrass(osg::Group* cellnode, Resource::ResourceSystem* rs)
@@ -71,7 +79,7 @@ namespace MWRender
 
         if (mUseAnimation)
         {
-            grassGroup->addUpdateCallback(mWindSpeedUpdater);
+            grassGroup->addUpdateCallback(mGrassUpdater);
         }
     }
 
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -181,6 +181,12 @@ namespace MWRender
                 mWindSpeedUniform = new osg::Uniform("windSpeed", 0.0f);
                 stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
                 stateset->addUniform(mWindSpeedUniform.get());
+                osg::ref_ptr<osg::Material> defaultMat (new osg::Material);
+                defaultMat->setColorMode(osg::Material::OFF);
+                defaultMat->setAmbient(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
+                defaultMat->setDiffuse(osg::Material::FRONT_AND_BACK, osg::Vec4f(1,1,1,1));
+                stateset->addUniform(new osg::Uniform("colorMode", 0), osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
+                stateset->setAttributeAndModes(defaultMat, osg::StateAttribute::ON|osg::StateAttribute::OVERRIDE);
             }
             mNode->removeUpdateCallback(mAlphaUpdater);
             mAlphaUpdater = nullptr;
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -170,17 +170,20 @@ namespace MWRender
         }
         else
         {
+            if (mNode->getStateSet())
+            {
+                // FIXME: probably there is a better way to reset stateset
+                mNode->setStateSet(nullptr);
+                osg::StateSet* stateset = mNode->getOrCreateStateSet();
+                // @grass preprocessor define would be great
+                stateset->addUniform(new osg::Uniform("isGrass", true));
+                // for some reason this uniform is added to other objects too? not only for grass
+                mStormDirectionUniform = new osg::Uniform("WindDirection", (osg::Vec3f) osg::Vec3f(0.0, 0.0, 0.0));
+                stateset->addUniform(new osg::Uniform("Rotz", (float) mPos.rot[2]));
+                stateset->addUniform(mStormDirectionUniform.get());
+            }
             mNode->removeUpdateCallback(mAlphaUpdater);
             mAlphaUpdater = nullptr;
         }
-
-        if (alpha != 1.f)
-        {
-            osg::StateSet* stateset = mNode->getOrCreateStateSet();
-            stateset->setRenderingHint(osg::StateSet::TRANSPARENT_BIN);
-            stateset->setRenderBinMode(osg::StateSet::OVERRIDE_RENDERBIN_DETAILS);
-        }
-        else if (osg::StateSet* stateset = mNode->getStateSet())
-            stateset->setRenderBinToInherit();
     }
 }
--- a/apps/openmw/mwrender/grass.cpp
+++ b/apps/openmw/mwrender/grass.cpp
@@ -34,8 +34,39 @@ namespace MWRender
     {
         for (MWRender::GrassItem& item : mItems)
         {
-            item.attachToNode(cellnode, rs, mWindSpeedUniform.get(), mIsGrassUniform.get());
+            attachToNode(item, cellnode, rs);
         }
+    }
+
+    void Grass::attachToNode(MWRender::GrassItem& item, osg::Group* cellnode, Resource::ResourceSystem* rs)
+    {
+        osg::ref_ptr<SceneUtil::PositionAttitudeTransform> insert (new SceneUtil::PositionAttitudeTransform);
+        cellnode->addChild(insert);
+
+        insert->setPosition(item.mPos.asVec3());
+        insert->setScale(osg::Vec3f(item.mScale, item.mScale, item.mScale));
+        insert->setAttitude(
+            osg::Quat(item.mPos.rot[2], osg::Vec3(0, 0, -1)) *
+            osg::Quat(item.mPos.rot[1], osg::Vec3(0, -1, 0)) *
+            osg::Quat(item.mPos.rot[0], osg::Vec3(-1, 0, 0)));
+
+        insert->setNodeMask(Mask_Grass);
+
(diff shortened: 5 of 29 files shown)


kcat:
I'm having trouble understanding how this conceptually works. The presence of a plugin makes me think the grass is hand-placed objects, while the Density config parameter makes it seem as though it's placed procedurally. What determines what grass appears where and by how much?

akortunov:
> What determines what grass appears where and by how much?

```
Now statics which ID's have the "grs" substring are considered to be a grass. 
We need a more reliable approach to determine grass objects.
```
```
Attach grass nodes directly to cell node, without creating 
a cellref and animation object just for every grass shape. 
```

kcat:
Yes, but you also mention a density parameter, and that there's no batching so high density means lower performance. If grass is instead a static object (which is detected with the "grs" substring and given a "fast path" of sorts), that means grass is hand-placed by the plugin, so where does the density come in and what does it do?

akortunov:
I do not understand the question.
I already told that I just skip part of grass objects during cell load.
I doubt that there is a simpler explanation, which would be shorter than code itself.

AnyOldName3:
I don't think we should be merging this, no matter the cool-factor, until at the minimum we're accepting changes to the ESM format, so could tag grass with its own field.

I don't see how we could be adding grass to the scene without it being capable of receiving lighting. , but I guess you could be skipping whatever system adds a light update callback to each object when you're skipping doing things.

kcat:
> so 50% means that every grass object has a 50% chance to be rendered

That's what I was unsure about. So `density = 0.5` means each placed grass object in the plugins has a 50% chance to "spawn"?

wareya:
The density is just a performance setting.

akortunov:
> So density = 0.5 means each placed grass object in the plugins has a 50% chance to "spawn"?

If you re-read my mesage, you may notice that it was the comparison with MGE: 

> I already told that I just skip part of grass objects during cell load (depending on selected grass density).

> MGE [in comparison] does pretty the same thing, but it uses a random generator instead (so 50% means that every grass object has a 50% chance to be rendered).

apps/openmw/mwrender/objects.cpp
```diff
@@ -109,6 +110,9 @@ void Objects::insertGrass(MWWorld::CellStore* cell)
 
         insert->setAttitude(makeObjectOsgQuat(grass.mPos));
 
+        osg::ref_ptr<SceneUtil::LightListCallback> lightListCallback = new SceneUtil::LightListCallback;
+        insert->addCullCallback(lightListCallback);
```
AnyOldName3: Definitely a cull callback, not an update callback?

akortunov: A quote from the lightmanager.hpp:

> To receive lighting, objects must be decorated by a LightListCallback. Light list callbacks must be added via
**node->addCullCallback(new LightListCallback).**
 Once a light list callback is added to a node, that node and all its child nodes can receive lighting.

AnyOldName3: Cool. I'd misremembered it as an update callback for some reason, but that obviously wouldn't work very well.

akortunov:
So what did we decide to do?

Capostrophic:
In my opinion this isn't the way to add grass, and it's no use to add it if we can't get on par with MGE XE:
1) The most common complaint of the users is performance, not how the grass looks. The performance is only changed slightly.
2) Even then, the grass still looks not that good.
3) The way the grass is integrated into ESM is awful,

akortunov:
> The performance is only changed slightly.

It is not true. Difference from master:
1. Grass density is configurable.
2. Grass rendering distance is configurable, and is not affected by the "exterior cell load distance" value.
3. Grass does not cast shadows.
4. Grass almost does not affect cell loading speed, while in master it cripples loading speed.
5. Memory consumption is lesser (about 300MB lesser consumption with the 100% density and about 750MB with the 30% density with the "exterior cell load distance = 1").

With 30% density and 8192 units distance grass costs about 4-6 FPS on GTX1050 while still looks a quite good.
With more powerful GPU the FPS drop with same settings more likely will be negligable, so it will be possible to use a higher density.

In the upstream master performance drop is MUCH larger, but many players still enable grass. Some modpacks for OpenMW even enable grass mods by default.

Capostrophic:
Couldn't come to an agreement in a discussion on the Fullrest Discord server. I cannot override potential executive decisions, but I stand by my opinion that we still need full parity with MGE XE before either us or some of the players stand advertising "grass support" that cannot satisfy all players that could move from Morrowind, and having a placeholder solution for *everything* in place because "some grass is better than no grass!" isn't good. I'll avoid continuing arguing.

psi29a:
I want to label this: technical debt

It's great you wanted to tackle this problem, but I'm not convinced that this will get us closer to having a working "Grass solution" and that we'll just end up ditching this code in the future. It just isn't a complete solution and doesn't help us carry OpenMW towards 1.0

So unless the majority of developers (stake holders?) think this is a good idea; I'm not keen on signing off on it. If you really think this is the way forward, you'll need to convince more developers that you're on to something.  I'm abstaining on this issue in either case.

akortunov:
> but I stand by my opinion that we still need full parity with MGE XE

We already violated that rule when merged a Distant Land which supports only terrain or shaders which do not implement all features which MGE shaders have.

AnyOldName3:
That's a false equivalence. There's no way to have a better terrain system than the terrain paging system, so we'll never have to replace it. The vertex and fragment shaders we use aren't remotely the same thing as post process shaders, and the vanilla engine uses vertex and pixel shaders under some circumstances, so even if they were related, it's not exclusively an mge thing

akortunov:
> There's no way to have a better terrain system than the terrain paging system, so we'll never have to replace it.

I do not understand the logic. 
From the one hand, it is OK to have a partially implemented Distant Land (which supports only terrain) without the "full parity with the MGE XE" just because no one provided an implementation which supports objects yet.
From the other hand, it is not OK to have an optimization for already existing and widely used assets because this optimization does not provide the "full parity with the MGE XE", despite it is even not claimed as a part MGE's features re-implementation.

Capostrophic:
Distant terrain was never intended to be a full replication of MGE XE's distant land functionality, it's a terrain system first and foremost.

akortunov:
> Distant terrain was never intended to be a full replication of MGE XE's distant land functionality, it's a terrain system first and foremost.

This PR was never intended to be a full replication of MGE XE's grass functionality, it's an optimization for already used assets first and foremost.

kcat:
IIRC, the "already used assets" are explicitly for MGE XE's grass system. The existing grass mods aren't supposed to be used on a running game, but instead are given to MGE's LOD generator and that handles it from there. So proper use of these assets would depend on a system like MGE's since that's what they're designed for. OpenMW's Distant Terrain, in contrast, is a system for handling terrain with far distances, providing similar functionality to MGE's, but in a different way that doesn't share resources/assets with it and can continue to be built upon separately.

I'll just reiterate what I said before about this PR, though. I don't mind it if it's functioning, and it doesn't get in the way of proper grass handling later. It would be a stop-gap to try and use some MGE grass mods to pretty up the terrain a bit, but shouldn't be considered the way forward with how to handle grass. It just depends if the added maintenance is worth it.

unelsson:
Any ideas how the grass batching should be done? That is to my understanding the main culprit for bad performance.

Can this PR be developed further to become _the solution_ or part of the solution that everyone wants?

I lack understanding of the technical details, but my five cents: Supporting grass mods for Morrowind is probably a good goal before extended features, but I agree that performance for the grass is most important goal. Animating grass has to be possible, but having a system that supports adding animation support later should be acceptable. Making procedural generation for grass shouldn't be a problem, but it's useless without performance.

kcat:
> Any ideas how the grass batching should be done? That is to my understanding the main culprit for bad performance.
>
> Can this PR be developed further to become the solution or part of the solution that everyone wants?

Unlikely. This PR depends on having a mod manually place each individual grass mesh, with some engine-side recognition of what objects count as "grass" to give it a more efficient render path. A more appropriate solution would likely depend on extending the esm format to define grass volumes (which themselves define the type/texture and height of the grass), and then the engine automatically generates the appropriate grass meshes on the ground in those volumes, batching as appropriate.

That's just my initial thoughts, though. It might be helpful to look into how other engines like Unity or UE handle dense dynamic foliage.

AnyOldName3:
Batching for legacy-style grass mods should be the same as batching for anything else. It's just one of the worst offenders right now for bazillions of tiny draw calls.

I don't think it's unreasonable to handle grass separately to other objects (e.g. to make shadows optional), either, just I think it should be handled by a flag in the ESP rather than a check in the name for a substring. Whether or not the grass is placed manually or procedurally, and whether or not the procedural placement is done at runtime or by the CS doesn't change that.

I can think of two potential approaches for procedural grass, and both would end up driven by grass density textures either in the ESP like how we currently have blendmaps for terrain textures, or inferred by assigning each terrain texture a grass density for each grass type. Either we pass the terrain tile to a geometry shader with the grass density texture bound and generate fully procedural grass based on that, or we define collections of grass meshes (e.g. we have one collection with six different tufts of green grass and one dandelion, and another with four tufts of ashlands grass) and have a density texture for each that gets used for each terrain tile, and place meshes when loading the world based on these textures. Both of these options are used elsewhere and can look good. They also both allow for an overall grass density setting that gets multiplied into the texture to reduce the density on lower-end machines.

akortunov:
> It might be helpful to look into how other engines like Unity or UE handle dense dynamic foliage.

IIRC, UE4 places vegetation objects randomly when you visit a location in the first time and stores position to the savegame. It allows trees/bushes/grass/etc to be in the same place when you visit the cell later.

AnyOldName3:
It surely makes more sense to use the cell coordinates as the seed to a deterministic RNG than to bake the results of the generator into the save.

Eli2:
This does not look very intrusive and might work fine as a stop-gap.
Does anyone have insights how later Beth engines do grass?

psi29a:
Coming back to this again, I like that most of it is self-contained. If someone wants to come up with a better way to handle existing mods and open things up for something better in the future, they can replace it or improve upon it.

With the inclusion of Object Paging and Active Paging, how does this work with this PR? Would the grass then become part of the merged mesh? Does grass at distance removed for being too small to render or is it 'cheap' enough to be added to the new mesh? How is performance with and without OP and with and without AG? 

Would batching improve performance even more or has then been made superfluous with OP/AG?  

Would an eventual LOD system to simplify meshes at distance improve the situation at all?

psi29a:
Also important, is this specific to just grass or ground cover in general? Should it not be made more generic than just 'grass' ?

psi29a:
Anyway we can split out the 'fade in/out' part to also support other objects? Like how how it is in done in Oblivion?

akortunov:
> With the inclusion of Object Paging and Active Paging, how does this work with this PR?

I wrote above that object paging with this PR just does not load grass objects.

AnyOldName3:
Don't you love it when it turns out the post you thought you'd posted the day before disappears (presumably because you never actually pressed send)?

Now I've thought of it, I'd be very surprised if object paging (active grid or otherwise) was capable of merging grass due to the translucency. There needs to be depth sorting to get it right, and that prevents merging. Maybe groundcover mod authors decide that accurate transparency doesn't matter much and turn it off, though, and maybe MGE XE turns it off despite what the Nif says. I'm not actually completely sure how object paging works, though, due to the technical writeup I said we needed before it was merged having never materialised.

I'd be much happier if most of the stuff in this PR could be turned on with options in the ESP rather than by detecting strings in Nifs, and might be better if made more generic. For example, if we support 'true statics' then we can avoid the cellref and animation stuff for those, too, and having a way to say that certain objects aren't worth considering as shadow casters without settings maxed out could be good too (although with how nodemasks work in OSG, it's only really simple if we have a dedicated groundcover nodemask like this PR adds).

psi29a:
Using the mod linked at the top, I did a series of three 4 tests across 3 saves. Keep in mind that all tests are without using this PR. My tests included the baseline of OP and AG being turned off and no Aesthesia Groundcover. This was the best performing. The rest are setting OP and AG on/off with Aesthesia Groundcover mod turned on.

```
by the water:
OP off, AG off grass off
55 fps

OP off, AG off grass on
20 fps

OP on, AG off, grass on
17 fps

OP on, AG on, grass on
50 fps


from the air:
OP off, AG off grass off
88 fps

OP off, AG off grass on
72 fps

OP on, AG off, grass on
77 fps

OP on, AG on, grass on
79 fps


over looking hill:
OP off, AG off grass off
84 fps

OP off, AG off grass on
49 fps

OP on, AG off, grass on
37 fps

OP on, AG on, grass on
60 fps
```

My conclusion is that without OP _and_ AG, adding grass and ground cover, like in the mod, will punish OpenMW, dropping 30fps right off in my examples. Turning on OP just makes the situation worse, cutting another 12fps off in two tests. However... turning on AG improves the situation dramatically to the point that we are playable again.

Grass should be fast enough as-is with OP _and_ AG enabled. There can be additional room for improvement though, such as if the NIF has collisions enabled which can be solved by toggling the collision in the NIF files. Any further optimisations are better done in a generic way that's not specific to grass, at least pre-1.0.

Post-1.0, there is a potential to paint grass/foliage layers in the CS, reducing the amount of data that needs to be stored in the plugin (think procedural generation).

Animations are best done with a custom shader, which is, again, not really specific to grass (think banners waving in the wind and etc). We can introduce an #if branch in our main shader which will be toggled on by a keyword found in the nif/osg/dae and other files, along with any uniforms to customise the animation speed. This could be done pre-1.0, no editor support required.

akortunov:
> Grass should be fast enough as-is with OP and AG enabled, even with shadows.

No, it should not. Object paging does not really solve issues which cause performance and shadow quality drop for foliage, but introduces lighting issues for active grid.

psi29a:
> > Grass should be fast enough as-is with OP and AG enabled, even with shadows.
> 
> No, it should not. Object paging does not really solve issues which cause problems with performance and shadows quality for foliage, but introduces lighting issues when it is used for active grid.

That's incorrect, the lighting issue has _always_ been there but made more noticeable through AG.

From my testing with the mod you provided, it all looked really damn good. Perf was good and the quality of the shadows was also pretty good. However this is subjective observation. I'm sure it could be improved upon further by updating/modifying the NIFs and perhaps even a way to flip a bit in the NIF somewhere so that OpenMW can key off of that to enable or disable shadows. We could even have a setting to enable or disable shadows for NIFs with that specific field/bit toggled. That would give modders a way to control how things should appear on screen and give users of OpenMW a way to turn shadows on/off for those objects in a nice generic way.

psi29a:
I updated my test post to include a datapoint asked for by @AnyOldName3 , to test with OP/AG on but without grass/groundcover, as another baseline. 

Here is a snippet again of just the first test by the water, it was a more demanding scene:

```
OP off, AG off grass off: 55 fps
OP off, AG off grass on: 20 fps
OP on, AG off, grass on :17 fps
OP on, AG on, grass on: 50 fps
OP on, AG on, grass off: 53 fps
```
We we see is that with grass enabled and disabled while AG is on is a difference of 3 fps. 

We see here though that without AG, having grass cover is just a slide-show and not very playable. Once AG is turned on, we get back near how it was _before_ grass was added!

Yet in another place:

```
over looking hill:
OP off, AG off grass off: 84 fps
OP off, AG off grass on: 49 fps
OP on, AG off, grass on: 37 fps
OP on, AG on, grass on: 60 fps
OP on, AG on, grass off: 75 fps
```
We see a 15fps drop between grass on at 60fps and grass off at 75fps.


All in all, still very playable.

psi29a:
> Now I've thought of it, I'd be very surprised if object paging (active grid or otherwise) was capable of merging grass due to the translucency. There needs to be depth sorting to get it right, and that prevents merging. Maybe groundcover mod authors decide that accurate transparency doesn't matter much and turn it off, though, and maybe MGE XE turns it off despite what the Nif says. I'm not actually completely sure how object paging works, though, due to the technical writeup I said we needed before it was merged having never materialised.

I believe a rough overview wouldn't be useful, as it wouldn't answer your specific question. A more detailed one would just be "TL;DR" for most purposes.

If you have a look over here: https://gitlab.com/OpenMW/openmw/-/blob/master/apps/openmw/mwrender/objectpaging.cpp#L616
You'll see that OP/AG does merge alpha blended objects with:
`optimizer.setMergeAlphaBlending(true);`

Even if it wasn't implemented, alpha blending wouldn't be a requirement for grass. Just alpha testing would look fine too.

> I'd be much happier if most of the stuff in this PR could be turned on with options in the ESP rather than by detecting strings in Nifs, and might be better if made more generic. For example, if we support 'true statics' then we can avoid the cellref and animation stuff for those, too, and having a way to say that certain objects aren't worth considering as shadow casters without settings maxed out could be good too (although with how nodemasks work in OSG, it's only really simple if we have a dedicated groundcover nodemask like this PR adds).

As it turns out, the 'defacto' modder standard is pushing everything into a 'grass' directory in mesh and textures, even if we're talking about bushes, trees and other foliage and then detecting that string in the NIF. I agree with @AnyOldName3 that having another more generic way to handle things would be idea.

Speaking of animations, could we just handles those in shader?

AnyOldName3:
I think that's only merging alpha-blended stuff for distant cells as it depends on the size. I don't know precisely what size refers to, but unless you're redoing the merge every frame (which is probably going to involve strictly more work than just not merging) you're going to end up with transparency ordering issues.

psi29a:
AG is just OP but for the active cells. But yes, size is a factor. Do not forget though that we first see how simple it is, if it not costly, we merge simple things automatically. Most grass meshes are super simple.

You can see grass being batched with the debug on:
https://media.discordapp.net/attachments/260443767482482688/728631702259499046/screenshot056.png?width=1124&height=587

akortunov:
I have already tried to use object paging for grass. As for me, result is pretty bad, especially with increased viewing distance (10-15 cells).

kcat:
> I'm sure it could be improved upon further by updating/modifying the NIFs and perhaps even a way to flip a bit in the NIF somewhere so that OpenMW can key off of that to enable or disable shadows.

In my opinion, the NIFs/models and ESPs shouldn't determine whether something handles shadows or not. At most, they should determine what kind of thing an object is, and the engine determines whether or not that type of thing deals with shadows with some user option. e.g. the model/esp says "this is grass", and the engine has an option "grass casts shadows" that a user can toggle given their hardware or performance target.

Either way for shadows, be careful of the distinction between receiving and casting them. We definitely want grass to receive shadows regardless of whether they cast them or not, otherwise it'll appear to glow in some situations.

For lighting in general, having a limit on the number of lights per draw while attempting to minimize draws will be a constant struggle. Batching/grouping by its nature will make the problem worse. Until something better is implemented, like deferred shading or clustered forward rendering that abstracts the lighting from draw calls, it won't go away. This of course will require shaders (IMO, the sooner we stop caring about the ancient fixed-function pipeline that's been dead for over a decade, since GL3 in 2008, the easier that will get; we can always simulate ffp-style vertex lighting with shaders as a backup for people who want original-style visuals).

psi29a:
> I have already tried to use object paging for grass. As for me, result is pretty bad, especially with increased viewing distance (10-15 cells).

Really? PT people have been partying and I myself have an exceptionally long viewing distance set. Even TR people have been cheering lately as well. I think OP/AG on improves things quite dramatically, is very playable. That being said, there still lots of room for improvement.

AnyOldName3:
> > I'm sure it could be improved upon further by updating/modifying the NIFs and perhaps even a way to flip a bit in the NIF somewhere so that OpenMW can key off of that to enable or disable shadows.
> 
> In my opinion, the NIFs/models and ESPs shouldn't determine whether something handles shadows or not. At most, they should determine what kind of thing an object is, and the engine determines whether or not that type of thing deals with shadows with some user option. e.g. the model/esp says "this is grass", and the engine has an option "grass casts shadows" that a user can toggle given their hardware or performance target.

We have more freedom to selectively misinterpret Nifs with `NiStringExtraData` nodes than we do with ESMs. Otherwise it doesn't particularly matter where shadow casting is decided.

> For lighting in general, having a limit on the number of lights per draw while attempting to minimize draws will be a constant struggle. Batching/grouping by its nature will make the problem worse. Until something better is implemented, like deferred shading or clustered forward rendering that abstracts the lighting from draw calls, it won't go away. This of course will require shaders (IMO, the sooner we stop caring about the ancient fixed-function pipeline that's been dead for over a decade, since GL3 in 2008, the easier that will get; we can always simulate ffp-style vertex lighting with shaders as a backup for people who want original-style visuals).

We don't need anything as fancy as clustered shading and deferred is vetoed. We can *probably* just have all the lights active for all the drawables and end up with reasonable performance. The overhead of doing something more complicated means it isn't worthwhile unless we have data saying otherwise.

floopfloop's work will only work with shaders on, so we probably won't be recommending active grid in FFP mode ever.

The main reason we're keeping the FFP is because most of the time, that's what the original engine uses, and it means when we're doing comparisons, it's easier to see if the difference comes from our shaders or the data we're putting onto the GPU. Totally abandoning pre-core OpenGL will be more work than maintaining the FFP path, too, as we're not really having to *do* anything to keep it working, but there's a *lot* of other deprecated stuff we rely on. Also, if we get native GLES working, GLES2 has a lot of the deprecated stuff still.

kcat:
> We can *probably* just have all the lights active for all the drawables and end up with reasonable performance. The overhead of doing something more complicated means it isn't worthwhile unless we have data saying otherwise.

I would doubt that, especially for per-pixel lighting. Given that CFR essentially involves uploading all active light data to the GPU, but still splits up the viewport so that each segment only handles a small number of those lights, there's clear benefits to avoid having all lights processed for every drawable even on modern games. With mods that add more light sources, enabling per-pixel lighting, and wanting to keep those light sources visible in the distance, it can easily overwhelm modern systems to throw all lights at every pixel.

> Totally abandoning pre-core OpenGL will be more work than maintaining the FFP path, too, as we're not really having to do anything to keep it working

I didn't say we should abandon pre-core OpenGL (right away, at least), just the ffp path. We're not really having to do anything to keep the ffp working because more drastic changes haven't been allowed. Granted it was a while ago, and the stance on shader-only features seems to have relaxed in recent times, but we still need to make sure the ffp works as the vanilla engine had it, maintain its existing options, and ensure any and all new options don't interfere with it. As new shader-only features get added requiring bigger changes to the renderer, the maintenance cost is just going to increase from here. Switching to GL3 would be work, yes, but not without its benefits, so having the option open for the future would be good.

akortunov:
> Really?

Really. With large viewing distance all grass objects from horizon to horizon take ages to load for me, even with SSD, use large amount to RAM and cause stuttering despite FPS counter shows playable values.

When the grass rendering distance is limited by active grid, an overhead is much smaller and does not become larger when with larger viewing distance.

psi29a:
> Really. With large viewing distance all grass objects from horizon to horizon take ages to load for me, even with SSD, use large amount to RAM and cause freezes and stuttering despite FPS counter shows playable values.
> 
> When the grass rendering distance is limited by active grid, an overhead is much smaller and does not become larger when with larger viewing distance.

You mean that with just OP that it freezes and stuttering, but with AG enabled it isn't as bad? Because I do not recommend OP with any grass mod unless they have AG enabled.

akortunov:
> You mean that with just OP that it freezes and stuttering, but with AG enabled it isn't as bad?

AG does not tolve issues that I described above.

AnyOldName3:
Do you still have weird non-default settings for OP from before it was merged? You shouldn't end up with grass from horizon to horizon as it's too small.

AnyOldName3:
> > We can _probably_ just have all the lights active for all the drawables and end up with reasonable performance. The overhead of doing something more complicated means it isn't worthwhile unless we have data saying otherwise.
> 
> I would doubt that, especially for per-pixel lighting. Given that CFR essentially involves uploading all active light data to the GPU, but still splits up the viewport so that each segment only handles a small number of those lights, there's clear benefits to avoid having all lights processed for every drawable even on modern games. With mods that add more light sources, enabling per-pixel lighting, and wanting to keep those light sources visible in the distance, it can easily overwhelm modern systems to throw all lights at every pixel.

Morrowind only *has* a small number of lights, and the new fancy forward techniques have significant overhead, hence why people did complicated deferred shading with significant overhead for years. You still need to attach all the lights to every shader and it adds a lot more dynamic branching. GPUs *really* don't like dynamic branching. A few years ago, they didn't even *have* dynamic branching - all branches were taken, then the results of all branches were multiplied by the branch condition and everything was summed together.

> > Totally abandoning pre-core OpenGL will be more work than maintaining the FFP path, too, as we're not really having to do anything to keep it working
> 
> I didn't say we should abandon pre-core OpenGL (right away, at least), just the ffp path. We're not really having to do anything to keep the ffp working because more drastic changes haven't been allowed. Granted it was a while ago, and the stance on shader-only features seems to have relaxed in recent times, but we still need to make sure the ffp works as the vanilla engine had it, maintain its existing options, and ensure any and all new options don't interfere with it. As new shader-only features get added requiring bigger changes to the renderer, the maintenance cost is just going to increase from here. Switching to GL3 would be work, yes, but not without its benefits, so having the option open for the future would be good.

With how pre-core GLSL works, you basically still have to do all the same setup as you would for the FFP, and then also write a shader. We access the FFP lights and materials and vertex attributes in our shaders, so all we need to do to have the FFP is have no shader bound and it Just Works™. Obviously some of that will change with the lighting rework, but not all that much. We're already vetoing non-vanilla features with the FFP and not being hugely fussy if vanilla ones don't make it there, either, so all that keeping the FFP entails is just not purposely getting rid of it.

akortunov:
> Do you still have weird non-default settings for OP from before it was merged?

I use defaults.

kcat:
> Morrowind only *has* a small number of lights, and the new fancy forward techniques have significant overhead, hence why people did complicated deferred shading with significant overhead for years.

Vanilla Morrowind may only have a small number of lights, but mods can and do easily increase that number, as would enabling distant land and letting lights remain visible in the distance. The "fancy forward techniques" were designed to get around the overhead of having every light affect every pixel; no engine has ever done 'apply all lights to everything' given a decent number of lights because the costs become untenable. Even the 8 light limit can be hard on older GPUs if you have per-pixel lighting, normal mapping, etc.

It was only relatively recently (recently as in 5 or so years ago) that the CFR technique we know today was developed. The original CFR technique was developed years prior, but worked in 2D space on the viewport. That had limitations as the number of lights allotted per segment was too low since each segment covered the entire depth range, causing artifacting without additional work to determine which lights actually affected a surface in that segment (which would indeed undercut the potential speed gains). However, improvements in GPU data streaming and extending CFR to work in 3D on the viewport has helped make it a viable option again without the complex light-on-surface-in-segment tests (basic bounding box tests will suffice). There's a reason Doom 2016, which has a focus on high framerates and fast movement, went with it over deferred or standard forward rendering.

> You still need to attach all the lights to every shader and it adds a lot more dynamic branching. A few years ago, they didn't even have dynamic branching - all branches were taken, then the results of all branches were multiplied by the branch condition and everything was summed together.

GL3 and DX10 introduced dynamic branching to shaders. Though yes, with a performance penalty.

Regardless, CFR doesn't depend on dynamic branching. There's one large buffer containing the light array data, and `num_segments*lights_per_segment` number of indices referring to that array, and this data remains constant for all draws in the frame. The number of segments and number of lights per segment never changes for the life of the program. For each frame, the light data gets updated (for changes in position, color, etc) and the indices get recalculated (using simple bounding box tests on the viewport).

AnyOldName3:
DOOM 2016 has 256 lights *per cluster* (which is more than I think would ever be in the view frustum in moderately modded Morrowind) giving up to 786432 lights total and does Cook-Torrance shading (which is *much* more expensive than the Lambertian shading that Morrowind uses for every light except the sun). There's far more work that can be saved with clustered shading than exists in OpenMW in the first place, so that makes the overhead of setting it up obviously worthwhile in Doom while it isn't obvious for us. I'm not saying it's not the long-term plan, as it *is* the long-term plan, but it's not the immediate plan.

Basic forward is *probably* fastest with anything less than a hundred to a thousand lights. Forward+ and friends make it viable for numbers closer to ten thousand, and clustered makes that closer to a hundred thousand. Once you get to a million lights, you're pretty much stuck with deferred again. Morrowind is much closer to the lower end of the scale, and uses a far simpler lighting model than the test scenes that were used to generate these numbers.

Branching has *always* been a thing in shaders. `if` has been part of GLSL and HLSL from their inception, and equivalents were available in their pseudo-assembly predecessors. Nothing really changed with GL3 and DX10, but OpenGL 4.0 and Direct3D 11 introduced a concept that OpenGL calls *Dynamically Uniform Expressions*, meaning that a dynamic branch only depends on the values of uniforms and constants, so might as well have been a preprocessor define. There's no requirement that branches based on dynamically uniform expressions are any *faster*, but it suddenly became legal to have undefined behaviour in the not-taken part of such branches, either as the GPU would implement a jump instruction and genuinely skip over it, or the driver would implicitly convert it to a preprocessor definition and compile out the bad code. Before that, undefined behaviour in not-taken branches was always allowed to infect the output. There's still no requirement that GPUs actually implement any kind of jump instruction even with OpenGL 4.6, Vulkan and Direct3D 12.

Anyhow, clustered forward *definitely* relies on dynamic branching. If you want to only work on the relevant cluster's data, you need to take a different path through the shader depending on which cluster you're in so you work on that cluster's light list and not all the others. The information you've posted about how light data is uploaded to the GPU is basically correct, but isn't relevant to what the shader needs to do with it once it's run.

kcat:
> DOOM 2016 has 256 lights *per cluster* (which is more than I think would ever be in the view frustum in moderately modded Morrowind) giving up to 786432 lights total

*Up to* 256 lights per cluster. Though it'd be more accurate to say it has up to 256 items per cluster, where each item could have a light, decal, and probe/cubemap index (the cluster having a variable count for each). The way Doom 2016 implemented it did use dynamic branching, as it would be more efficient to loop on a variable count and process only the lights that need processing, rather than force processing of all potential lights. Consequently, the main reason it could have up to 256 is because it used a simple byte for the light count, and not bothering to enforce an arbitrary limit beyond that.

And "only" 4096 total lights were possible. Each item had a 12-bit index into the light data array, which means no more than 4096 lights could be addressed in a frame.

> There's still no requirement that GPUs actually implement any kind of jump instruction even with OpenGL 4.6, Vulkan and Direct3D 12.

Regardless of how it's implemented, GL3 and D3D10 added the ability for shaders to do indeterminate jumps where it can't know ahead of time where execution will go and there's no explicit upper limit. Given loops, such as
```
int nNumSamples = (int)lerp( nMaxSamples, nMinSamples, dot( E, N ) );
...
int nCurrSample = 0;
while ( nCurrSample < nNumSamples )
{
    fCurrSampledHeight = NormalHeightMap.SampleGrad( ... ).a;
    if ( fCurrSampledHeight > fCurrRayHeight )
    {
        ...
        nCurrSample = nNumSamples + 1;
    }
    else
    {
        nCurrSample++;
        ...
    }
}
```
[source](https://www.gamedev.net/tutorials/_/technical/graphics-programming-and-theory/a-closer-look-at-parallax-occlusion-mapping-r3262/)
This wasn't possible in GL2/D3D9 shaders as they required loops to have a compile-time constant count (and were thus guaranteed to be fully unrollable and not need jumps), but could be done in GL3/D3D10. Short of probing and recompiling it on every invocation, that can't work without some kind of condition-based jump.

> Anyhow, clustered forward *definitely* relies on dynamic branching. If you want to only work on the relevant cluster's data, you need to take a different path through the shader depending on which cluster you're in so you work on that cluster's light list and not all the others.

It needs relative addressing, which has been a thing since the beginning, rather than branching. Roughly:
```
/* Presuming viewport goes [0...1), easily adjustable if it's something else. */
int cluster_idx = (int)(viewport.x*cluster_w) + (int)(viewport.y*cluster_h)*cluster_w
    + (int)(viewport.z*cluster_d)*cluster_wh;
int item_offset = cluster_idx * 8;
for(int i = 0;i < 8;++i)
{
    int light_idx = item[item_offset + i].light_idx;

    ... process light_data[light_idx] ...
}
```
Or with indeterminate looping/branching:
```
int cluster_idx = (int)(viewport.x*cluster_w) + (int)(viewport.y*cluster_h)*cluster_w
    + (int)(viewport.z*cluster_d)*cluster_wh;
int count = cluster[cluster_idx].light_count;
int item_idx = cluster[cluster_idx].item_offset;
while(count)
{
    int light_idx = item[item_idx++].light_idx;

    ... process light_data[light_idx] ...
    --count;
}
```

akortunov:
I think, this thread is not the right place to discuss lighting systems in other engines.

akortunov:
Here are some testing screenshots with 40000 viewing distance:
1. Object paging + paging for active grid, 3.3 GB RAM usage
Screenshot_20200706_140257

2. Object paging without active grid paging, 3.3 GB RAM usage:
Screenshot_20200706_140354

3. Separate grass handling, 100% density, 8192 rendering distance, 2.3 GB RAM usage:
Screenshot_20200706_140424

4. Separate grass handling, 50% density, 8192 rendering distance, 2.3 GB RAM usage:
Screenshot_20200706_140450

From what I can tell, object paging allow to decrease CPU usage, but it does not solve any other issues with GPU loading, RAM usage or loading speed. So do not tell me that it "just works", it does not.

psi29a:
I've had a look over your PR, as it exists here and did additional testing. It seems to be 'fighting' Object Paging and Active Grid, offering worse overall performance when OP/AG is enabled. That is simply not acceptable.

near shore:
```
OP off, AG off grass off, grass-handling off: 55 fps
OP off, AG off grass on, grass-handling off: 20 fps
OP on, AG off, grass on, grass-handling off: 17 fps
OP on, AG on, grass on, grass-handling off: 50 fps
OP on, AG on, grass off, grass-handling off: 53 fps
OP on, AG on, grass on, grass-handling on: 45 fps
OP on, AG off, grass on, grass-handling on: 35 fps
OP off, AG off, grass on, grass-handling on: 48 fps
```

over looking hill:
```
OP off, AG off grass off, grass-handling off: 84 fps
OP off, AG off grass on, grass-handling off: 49 fps
OP on, AG off, grass on, grass-handling off: 37 fps
OP on, AG on, grass on, grass-handling off: 60 fps
OP on, AG on, grass off, grass-handling off: 75 fps
OP on, AG on, grass on, grass-handling on: 43 fps
OP on, AG off, grass on, grass-handling on: 34 fps
OP off, AG off, grass on, grass-handling on: 58 fps
```

Note: keep in mind that AG will be turned on by default soon as it doesn't make sense to have it off when OP is on, so that setting will eventually disappear.

Even with OP/AG both off, this PR's performance comes within a few fps of when OP/AG are both on. The only benefits your PR has though is that it loads faster and saves several hundred megabytes of memory but only when OP and AG are disabled. That is a truly horrible drawback and makes this PR almost mutually exclusive to OP/AG. In its current state, it's unmergable.

By mutually exclusive, I mean this:
https://github.com/OpenMW/openmw/pull/2665/files#diff-440c2258ab5e948edc7c4e2e3b95e060R482
you took everything that would have 'grass' in their path out of Object Paging entirely.

Now if there is a way make this PR play nice with OP/AG by actually improving performance instead of decreasing it, then we can see about merging this PR in. Taking your work on grass density and then feeding that to OP would also increase performance, decrease load times and reduce the amount memory used.

akortunov:
> keep in mind that AG will be turned on by default soon

I doubt that lighting issues will be solved in the foreseen future, so there is no point to enable AG by default. And again, it is not a silver bullet, it decreases only CPU loading, while GPU loading becomes the same. When OP is enabled, many systems may become GPU-limited due to large amount of polygons in scene, especially with high-poly assets and/or large viewing distance. On such setups AG is pretty useless and just causes additional memory usage and lighting issues.

Also keep in mind that object paging was optimized for vanilla assets and level design. It simply was not designed to handle groundcover, which should be treated separately from generic static objects.

> That is a truly horrible drawback and makes this PR almost mutually exclusive to OP/AG. In its current state, it's unmergable.

In my testing I got pretty opposite result - OP/AG provides much worse result than this PR indifferently from settings, because my setup is GPU-limited with OP. Does it make OP/AG unmergable?

> you took everything that would have 'grass' in their path out of Object Paging entirely.

Yes, that's the point. Object paging in it current state can not handle groundcover objects properly and handles them just as any other statics, so from my side I can only make it to ignore groundcover objects when groundcover handling is enabled.

AnyOldName3:
no

Sisah2:
Here is my attempt to bring simple animation :)
Need some improvements.
[simpleanimation.txt](https://github.com/OpenMW/openmw/files/5046829/simpleanimation.txt)

akortunov:
Nice job.
Basically, it works and does not cause significant performance loss on PC.
But grass movement itself is twitchy since it uses reverse-engineered wind speed formula, which was not designed to animate grass at all.

I have a suggestion - take wind speed from openmw.cfg for current weather (and take in account weather transitions).

akortunov:
I uploaded a simplified version of animated grass for additional testing.
Main differences:
1. Use wind speed from openmw.cfg, so we do not need to smooth it.
2. Disable "storm" mode so far - it does not work properly anyway.

akortunov:
I'd appreciate some help with transparency updater. 
From what I can tell, when the grass object is far enough, I need to create a new stateset with new material and set alpha here.
Then when the object is close enough, I need to remove created stateset, otherwise grass does not look as expected.
Unfortunately, when I do it, grass animation stops to work. Do I need to recreate shaders or do something like this?

akortunov:
As for "storm" mode, probably it would be better to remove it at all so far and leave it to follow-up PR's.

Sisah2:
I think it need to add those uniforms to newly created stateset.
IsGrass, WindDirection and Rotz. I would try
Uniform bool isGrass;
Main(){
.
.
.
If(IsGrass) gl_FragData[0].x = 1.0;
}
 in fragment shader to see if its pased here.

Without "storm" mode some parts can be removed. Do you think preprocessor define can be created for it and set to 1 if animation is enabled and shader is running for grass?

akortunov:
> I think it need to add those uniforms to newly created stateset.

Thanks, it works.

> Do you think preprocessor define can be created for it and set to 1 if animation is enabled and shader is running for grass?

Maybe, similar to specular maps.

Sisah2:
This is shader without storm stuff, just need to change vec3 WindDirection from vec3 to float windSpeed in engine.
https://pastebin.com/xQ83jv8R

akortunov:
Done. Use a really simple animation from starting, a more complex one can be added later if needed.

The main question now is how to handle lighting for grass objects. Since MGE uses instancing for grass, I doubt that it uses light attenuation as common statics in scene. I do not know if there is a custom solution to light up the grass is used, or MGE just does not use lighting for grass.

vtastek:
MGE uses sun/ambient and backligthing. MGE's distant land does not do point lights, and MGE's grass is part of distant land.

akortunov:
> MGE uses sun/ambient and backligthing.

I suppose that the `LightListCallback` is not needed here - ambient is handled without it for the whole scene.
I do not know what the "backlighting" is and if OpenMW has it at all.

An only thing which I do not like is that without `LightListCallback` terrain will get point lights, but groundcover on this terrain will not:
Screenshot_20200809_230711

vtastek:
> An only thing which I do not like is that without `LightListCallback` terrain will get point lights, but groundcover on this terrain will not:

Is it a performance issue? I would like to see point lights on grass.
Backlighting is simply
```
 if(lambert < 0)
        lambert *= -0.3;
```

akortunov:
> Is it a performance issue?

It is a visual fidelity issue (there are seams and abrupt transitions with current lighting system).
Also it is harder to implement batching with point lights since whole grass batches will be treated as a single drawable.

vtastek:
> > Is it a performance issue?
> 
> It is a visual fidelity issue (there are seams and abrupt transitions with current lighting system).
> Also it is harder to implement batching with point lights since whole grass batches will be treated as a single drawable.

Then omitting it until a better lighting system implemented is the way to go. Backlighting can be added easily and it would be MGE level.

akortunov:
Can you experiment with it then? The main point on the first stage is to make grass to lokk more-or-less as in MGE to re-use MGE mods.

Point lights can be disable for groundcover by commenting out the `LightListCallback`:
```
insert->setNodeMask(Mask_Grass);
//osg::ref_ptr<SceneUtil::LightListCallback> lightListCallback = new SceneUtil::LightListCallback;
//insert->addCullCallback(lightListCallback);
```

akortunov:
About fading - it seems to does not work properly for meshes which use alpha blending (15084 flags for NiAlphaProperty in NifSkope). Not sure what do do here - disable blending on engine side or just document the limitation.

Sisah2:
I think it can be done in shaders too, something like
If(isGrass) 
    gl_FragData[0].a = 1.0-smoothstep(@grassFadeStart.0, @grassFadeEnd.0, euclideanDepth);

akortunov:
> I think it can be done in shaders too

It is with a = 1.0:
Screenshot_20200810_111547

With lesser values grass is just missing.

akortunov:
All right, we need to use *= instead of just =.

Sisah2:
There is also issue with remiros mushrooms animation, only "hat" is moving it look bad :(

akortunov:
Use shader to update transparency now.
Now we do not need to mess with statesets.

akortunov:
I noticed that lighting behaves wrong for some replacers:
Screenshot_20200810_105358

I have a suspicion that affected meshes have malformed normals, but MGE compensates it via backlighting.

akortunov:
Also I tried to experiment with lambert lighting, but it affects only front sides of objects:
Screenshot_20200810_134402

In MGE it looks like both sides of mesh get backlighting:
Morrowind_df_1_09 04_0001


 Any ideas how to do it?

akortunov:
A related fragment from MGE's shader:

```
    // Lighting for two-sided rendering, no emissive
    float lambert = dot(normal.xyz, -SunVec) * -sign(dot(eyevec, normal.xyz));
    // Backscatter through thin cover
    if(lambert < 0)
        lambert *= -0.3;
    
    // Ignoring vertex colour due to problem with some grass mods
    OUT.colour.rgb = SunCol * lambert + SunAmb;
```

Sisah2:
This is probably wrong
vec3 SunVec = gl_LightSource[0].position.xyz - (passViewPos.xyz * gl_LightSource[0].position.w);//??
vec3 eyeVec = passViewPos.xyz;//?? its probably not eyeVector
float lambert = dot(viewNormal.xyz, -SunVec) * -sign(dot(eyeVec, viewNormal.xyz)) ;
if(lambert < 0.0) lambert *= -0.3;
gl_FragData[0].rgb *= gl_LightSource[0].diffuse.xyz * lambert + gl_LightSource[0].ambient.xyz;

Vtastek probably know better.

Sisah2:
Doing this in perLight seems to lit mesh from both sides
If(isGrass)
diffuseOut = diffuse.xyz * gl_LightSource[lightIndex].diffuse.xyz * (max(dot(viewNormal.xyz, lightDir), 0.0) + max(dot(-viewNormal.xyz, lightDir), 0.0))  * illumination;
else
diffuseOut = diffuse.xyz * gl_LightSource[lightIndex].diffuse.xyz * max(dot(viewNormal.xyz, lightDir), 0.0) * illumination;

https://youtu.be/WMgJYue3W9M

files/shaders/objects_fragment.glsl
```diff
@@ -51,6 +51,10 @@ uniform mat2 bumpMapMatrix;
 
 uniform bool simpleWater;
 
+#if @grassAnimation
+uniform bool isGrass;
```
Capostrophic, line 55: This uniform used later doesn't exist if grass animation isn't enabled which causes issues.

akortunov:
> Doing this in perLight seems to lit mesh from both sides

No, there is still a lot of dark planes.

akortunov:
According to independent testing, there are dark planes for affected meshes in MGE too, but they are partially masked by post-processing shaders.

akortunov:
> There is also issue with remiros mushrooms animation

Remiros Groundcover uses custom shaders. I suppose it does not work even with default ones from MGE.

akortunov:
Also according to testing, MGE ignores materials, attached to grass meshes. I have a suspision that it uses default 1.0 values for everything, so grass looks brighter in MGE than in OpenMW when non-1.0 diffuse or ambient are used.

Sisah2:
What about this? I dont have windows so cant compare, but it may work.
if(isGrass)
{
    float lambert = dot(viewNormal.xyz, lightDir);
    if(lambert < 0.0)
        lambert *= -0.3;
    diffuseOut = gl_LightSource[lightIndex].diffuse.xyz * lambert * illumination;
}
else
    diffuseOut = diffuse.xyz * gl_LightSource[lightIndex].diffuse.xyz * max(dot(viewNormal.xyz, lightDir), 0.0) * illumination;

akortunov:
> What about this?

Does not behave any better.

akortunov:
If groundcover objects are not supposed to use bump/normal/specular/etc maps, maybe it would be better to define separate simple shaders fro groundcover objects. In this case we can avoid additional uniforms for all objects.

Sisah2:
That would be great, it also allow to force vertex lighting, which may give better performance.
Also skiping lighting for transparent pixels help gpu performance a lot, just like this
if(gl_FragData[0].a != 0.0)  gl_FragData[0] *= doLighting
At least with vtastek lighting and ppl.

akortunov:
> Also skiping lighting for transparent pixels

I suppose it would work only for per-pixel lighting.

Sisah2:
Yes it help only with perpixel lighting. to force vertex lighting it need separate grass shader or @grass define.

akortunov:
> to force vertex lighting

The point is that grass usually will use PPL only when "force per pixel lighting".
It is strange to force vertex lighting for valid object when player tells engine to force per-pixel lighting.

Sisah2:
Well, forcing vertex lighting for particles increase performance a lot, there it may be similar, its benefit for vtastek lighting where ppl is recomended, with openmw shaders ppl is not that usefull.

akortunov:
> with openmw shaders ppl is not that usefull.

That's why "force per pixel lighting" is disabled by default.

akortunov:
The main source of performance drop with grass is actually shadowing (the `unshadowedLightRatio()` function.
I wonder if there is a way to simplify shadowing for groundcover objects...

akortunov:
Ok, use separate simplified shaders for grass now.
On my setup it allows to reduce performance drop.

akortunov:
> maybe that shaderVisitor should be created only once?

No, it is created every time when you load an object from NIF-file.

> I have feeling there are some crashes now

I can not tell anything without context.

akortunov:
I use a unified way to recreate shaders now.
In theory, it *may* workaround crashes on Android if it was caused by missing `setAllowedToModifyStateSets(false)`, but I'd not really expect it.
But it is worth to check, though.

akortunov:
Also it may be not safe to update statesets directly (see [bug #5026](https://gitlab.com/OpenMW/openmw/-/issues/5026)), so use double-buffering now to avoid possible data races.

akortunov:
> // for some reason this uniform is added to other objects too? not only for grass

More likely it is because you define uniform as a variable for shader, not for some instances of objects. It is one of reasons why it may be better to use a separate shader for groundcover objects.

Sisah2:
Yep, this comment can be removed, good thing is no more crashes with yesterday change. one thing that should be improved is this line in shader.
float height = 1.0-(gl_ModelViewMatrix[0].z + gl_Vertex.z);
It can be actually only gl_Vertex.z, but problem is with remiros groundcover, some meshes seems to have axis.z in midle of that mesh which result in only upper part is animated, also if some paging for grass become a thing this will dont work at all, because its height of that whole batch.

akortunov:
> axis.z == 0 in midle of that mesh

It is called a "pivot point". Usually it is either a center of mesh bottom plane or just a center of mesh (as in your case).

akortunov:
Here is screenshot of stone from this mod:

Screenshot_20200820_143032

I suspect that it was done intentionally to do not animate such meshes.

akortunov:
Done. Stomping seems to work for small objects, but the effect looks weird when player is is the middle of large bush.
I wonder how MGE handles such case...

akortunov:
Ok, other users tell me that in MGE effect does not work properly for large objects as well.

akortunov:
Now the effect should not be that bad. At least, bushes with 2m height should not be too distorted.

akortunov:
So what's next?

Sisah2:
Kind of storm mode can be added, it just need smooth begining and end storm direction and add some extra animation, but it may look bad if that pivot point is in middle of grass mesh. And paging would be great :D

akortunov:
Paging is out of question so far.

As for "storm" mode, in its initial implementation it did not look good even when pivot point was in the bottom part of grass mesh. From what I can tell, shader moved bottom part of grass mesh instead of top one.
If this mode is really needed, it can be done in follow-up PR.

Sisah2:
It moved both, above pivot it moved in right direction, below pivot it moved in opossite direction. Something like this work better
    if(windSpeed.xy != vec2(0.0) && gl_Vertex.z > 0.0) {
        displacedVertex.z -= length(grassVertex)/3.14;
        grassVertex.xy += gl_Vertex.z*windSpeed.xy;
        //some sinewave to .z axis here for better look?
    }

Sisah2:
Also that max 10 is not needed it was also part of first "storm mode" it can be
vec2 grassVertex = grassDisplacement(worldPos, gl_Vertex.z);
Instead of
vec2 displ = grassDisplacement(worldPos, gl_Vertex.z); vec2 grassVertex = min(vec2(10.0), displ);

akortunov:
> Something like this work better

It does not seem to do anything.

Sisah2:
It should look like this.
https://pastebin.com/2GfT63Ay
windSpeed.xy is stormDirection uniform from engine.

akortunov:
Still looks meh, as for me, especially for relatively large objects.
So far this effects is not that good to mess with animation transitions to implement it.

akortunov:
@psi29a, so what's the conclusion?

psi29a:
I'm still not convinced that this is the way forward. There are more reports that even with animation support and fading, that performance is still an issue.

On a Intel i5 2500k, Nvidia GTX 970 system:
> [11:23 PM] CMAugust: Ok, so using grass PR, object paging with active grid, and using akortunov's grass settings at their defaults, I get 28fps.
> [11:23 PM] CMAugust: If I simply turn the special grass handling off so it's paged like other objects, the same scene gets 48fps.
> [11:23 PM] CMAugust: If I remove the grass mod altogether so there's no grass, the same scene gets 53fps.

I think it would be more interesting to get animation support in AG and fade-in/out in OP which would cover the two upsides of this PR.

akortunov:
> Nvidia GTX 970

It is a high-end GPU, despite a bit outdated.
Many users do not have such hardware (especially on mobile devices).

psi29a:
> > Nvidia GTX 970
> 
> It is a high-end GPU, despite a bit outdated.
> Many users do not have such hardware (especially on mobile devices).

Mobile hardware these days are catching up to or on par with older top-end hardware. 


> It would be more interesting to make OP/AG to work properly first (there is a list of existing OP/AG drawbacks in this thread), but I am not aware of any attempts to do it.

You could say that about many other parts of OpenMW as well.

Is someone stopping you from improving upon it? 

The only real drawback from AG's use is exposing a pre-existing flaw in our lighting situation. Everything else is finding a sweet spot for settings (RAM vs. CPU) and typical improvements/optimisations over time like any other part of OpenMW we have.

AnyOldName3:
Can we at least weed out the heresy without implying mobile GPUs are going to be anywhere close to a GTX 970 any time soon. I was going to link a performance comparison with a recent Snapdragon, but can't even find one as there don't seem to be benchmarks that run the same scene on Android and Windows.

cvuchener:
>On a Intel i5 2500k, Nvidia GTX 970 system:
>
>>  [11:23 PM] CMAugust: Ok, so using grass PR, object paging with active grid, and using akortunov's grass settings at their defaults, I get 28fps.
>>  [11:23 PM] CMAugust: If I simply turn the special grass handling off so it's paged like other objects, the same scene gets 48fps.

i5 9600k + gtx 970 on linux here, it is kind of the opposite or I am doing something wrong. Using remiros ground cover mod:
akortunov grass off, shadows off → 42 fps
akortunov grass off, shadows on → 33 fps
akortunov grass on, shadows off → 55 fps
akortunov grass on, shadows on → 50 fps
I have a nice fps boost, and shadows have a much lower impact.

Grass settings are

    [Grass]
    enabled = true
    fade start = 0.75
    density = 1.0
    distance = 6144
    animation = true

I am not sure about the active grid stuff, I have "object paging active grid = true" in the Terrain section, is that enough?

I am using this PR merged on top of ffed6acc49944124da130ac4575a7299624871b2, it may be a little outdated.

akortunov:
> it is kind of the opposite or I am doing something wrong.

It is hard to tell anything without full config. Viewing distance, shadow distance, and water reflection may affect testing result a lot (since in upsteam master grass is rendered on a full viewing distance, cast shadows and is used in reflections).

psi29a:
I did another test with this updated branch, I'll redo the test twice more (with same savegame/settings) on an ThinkPad 420s (nvidia/intel hybrid) again so that we can test a 10 year old laptop nvidia and Intel (3000) GPU.

I left everything default, so density is 1.0 and I used Remiros' Groundcover - OpenMW-46733-2-0-1577400885 this time around. I am also using a cell view distance of 9 cells since the point of OpenMW is an RPG and not a flight simulator. This is to give a more apples to apples comparison even though this PR allows for density to toned down a bit in add of performance. This is something else I would like to see available as an OP/AG param, if we can somehow key off specific data in a NIF that allows for density calculation in a more generic way just like any other modern engine (amount of foliage: none, low, medium, high).

I also really like the animation, that gives things such a nice feel to the static ground cover. It is definitely one thing we should have in AG and what makes this PR attractive to me at this point.


Current laptop (i9-9980HK w/ RTX 3000 Mobile)

master + OP (on) + AG (on) + 2665.patch
grass baseline no grass | with grass OP/AG only | with grass 2665 no anim | with grass 2665 with anim
fps:             81                |              65                    |                     55                |                53


Note: One of the cited factors was load times, however if you play with the settings in OP, you can reduce load times as well.

C3pa:
> if we can somehow key off specific data in a NIF that allows for density calculation in a more generic way just like any other modern engine (amount of foliage: none, low, medium, high).
> 
Well, from the content side it could be done using NiSwitchNode

akortunov:
> Well, from the content side it could be done using NiSwitchNode

A quite bad idea - you need to put *all* levels of density to the same NIF file and load them to scene, despite only one level will be used.

akortunov:
> i9-9980HK w/ RTX 3000 Mobile

It is nice that machine which costs about 5000-6000$ can handle run Morrowind with 60FPS, but most of users (and developers) do not have such hardware, and will not have it in the foreseen future.

psi29a:
> > i9-9980HK w/ RTX 3000 Mobile
> 
> It is nice that machine which costs about 5000-6000$ can run Morrowind with 60FPS, but most of users (and developers) do not have such hardware, and will not have it in the foreseen future.

Is you're comment supposed to sound condescending, because that is how it comes across. Please check your attitude problem at the door. 

It was around 2800 when it was bought, but it's a work laptop.  I haven't had a workstation or desktop in over a decade.  Another reason why I will test on my decade old ThinkPad, as I said in my post.

I miss my old Radeon laptop though, it was nice to be able to test on all 3 GPUs.

I'll get those results later.

akortunov:
> Is someone stopping you from improving upon it?

I told several times that it OP was designed to be a black box rather than readable and extendable module, so it is very hard to work with it. But if you understand how it works, you can help with it, though.

Here is a list of major culprits:
1. OP was designed for vanilla objects density and more-or less static view. 
   It means that if you plug-in a grouncover mod (which usually has about 10x more objects per cell than Morrowind.esm), paging for all these objects may take too many time, even if you use 30-50% density (it is still 4-6x slower than with clean game) and uses more memory, so it can cause stuttering, OOM crashes and other undesired behaviour. Also all these grass pages are used for the full view (which can be 5-10 cells), but it is an overkill - there is no real point to render grass outside of active grid (I got multiple reports that 4096-8192 distance is really enough for most of users).
  So instead of building a huge number of grass paging and relying that any user has GTX970 or higher, there should be a **way to tell OP to do not waste resources to handle groundcover objects outside of small (relatively to viewing distance) given range.** Also this way should be fast.
2. There is a way to define non-animated and partially animated groundcover objects an mesh level by moving mesh pivot point - only a part above pivot point is supposed to be animated. 
  It means that grass animation just will not work properly for pages. From what I can tell, MGE uses instances instead of paging, so shaders still have info about original meshes, so they can handle animations.
 So OP needs somehow to **tell shaders about Z position of each vertex from original, non-paged instances**.

Can OP handle these culprits?

psi29a:
Updated: https://github.com/OpenMW/openmw/pull/2665#issuecomment-702591343

Here you see this PR shine a bit, even if a Pyrrhic victory as even with grass off, it was nearly unplayable. 

To make things a bit more fare, I'll need to tweak the settings on my old laptop to get it more playable then try the test again. Anything in particular you would recommend @akortunov ?

Resolution is 1440x752, all shadows and all refractions are on.

akortunov:
> Anything in particular you would recommend

```
[Grass]
enabled = true
fade start = 0.75
density = 0.2
distance = 4096  # can be lower, if needed.
animation = true
```

psi29a:
I was referring to anything not OP/AG and Grass related so there could be a baseline comparison. I seem to remember enjoying good performance of OpenMW a decade ago... :rofl: :rofl:   (Time before refraction, water shader, distant terrain)

psi29a:
Round 2!  I turned off ALL shadows (wow, they look beautiful but it does have a high impact) and set the view distance to 3 cells (instead of 9).

master + OP (on) + AG (on) + 2665.patch
grassbaseline
```
no grass | with grass OP/AG only | with grass 2665 no anim | with grass 2665 with anim
i9 laptop fps: 350 |          300             |        90         |         76
i7 laptop fps fps: 81  |          65           |           55            |          53
```
300 fps with OP/AG and 90 fps using this patch. This PR killed performance by about 60% on  my i9 :eyes:
The same theme happened on my i7 laptop though the numbers were closer together, this PR cost a 15% performance drop.

akortunov:
```
# A maximum distance on which grass is rendered.
distance = 67614
```

Wait, what? The whole purpose of this PR is to reduce resource loading by limiting groundcover rendering distance. I'd expect to see 6144 or so.

psi29a:
Then 6144 should be applied to OP/AG to get any kind of comparable results.

akortunov:
> Then 6144 should be applied to OP/AG to get any kind of comparable results.

Actually, I asked a way to do it in one of my posts above, but you did not reply.

psi29a:
Lemme do another round real quick then

psi29a:
At 6144, I get with OP/AG only about 600 fps and with this PR it is about 220 fps. :eyes:

akortunov:
> At 6144, I get with OP/AG only about 600

Just for reference - if just decreased "viewing distance", it is not right.
The point is to have small viewing distance for groundcover objects while having large viewing distance.

psi29a:
> > At 6144, I get with OP/AG only about 600
> 
> Just for reference - if you just decreased "viewing distance" setting value, it is not right.
> The point is to have small viewing distance for groundcover objects while having large viewing distance.

I understand, you want to bypass OP/AG so that larger objects can be seen at distance and keep grass at a reasonable viewing distance to decrease memory usage and improve fps. However the system has performance issues (as shown) and adds additional code that isn't necessary.

Referencing your earlier comment about OP/AG being a block box, that's nonsense because the code is right there. You or someone else could also work with OP/AG to allow for the same technique of keying off the 'grass' directory name and giving a different (shorter) rendering distance than the camera view distance. I'm pretty confident that this would be more performant.

psi29a:
For the record,  I set the following:
```
[Camera]
viewing distance = 21903
[Grass]
# A maximum distance on which grass is rendered.
distance = 6144
```
That's 300 fps with OP/AG and about 140 fps with this PR.

akortunov:
> the code is right there

It does not really help.

> You or someone else could also work with OP/AG to allow for the same technique of keying off the 'grass' directory name and giving a different (shorter) rendering distance than the camera view distance.

I tried, but 

> OP was designed to be a black box rather than readable and extendable module, so it is very hard to work with it

As I said, I'd like to make OP to work with groundcover, but I did not find a way to do it without major rewrite. A major rewrite is out of question sinse no one excepts of bzzt knows about how this thing works and can modify it without breaking anything.

Also I still did not get an aswer how you are going to make grass animation to work with pages.

AnyOldName3:
The object paging MR doesn't actually change that much stuff. I'm pretty sure it's just leveraging OSG's database paging system, plus merging suitable geometries. The details shouldn't be too opaque.

I'd expect adding a special case for grass to probably work something like:
* Find the identical grass meshes in a cell.
* Use instanced rendering by having a position vector and Euler rotation vector for each of them instead of statically merging them.
* If they're past the fade distance, kill them in the vertex shader.
* The mesh origin is the same as before, so animation is the same as before.

All we really need is the loader to tag the grass meshes somehow to say that we want to use the special case. The instancing bit might not need that, as we can probably work out the threshold at which it's better to use instanced rendering than plain merging and have the merging function decide that generically.

akortunov:
> If they're past the fade distance, kill them in the vertex shader.

By using "discard"? I tries to do it, but it decreased performance a lot, so I do it on CPU now by setting node masks to 0.

AnyOldName3:
`discard` only works in the fragment shader, but I said to kill it in the vertex shader. If you know a vertex and every vertex it's connected to aren't going to be visible, you can set the location of the vertex outside the viewing frustum (or, even better, set its clip position to the wrong side of a clipping plane, or one of its clip distances negative), and it should get ignored much earlier in the pipeline. You kind of have to have the object's bounds in its vertex data, a uniform, or a texture, though, otherwise you can't determine the *every vertex it's connected to* until the geometry shader.

akortunov:
> you can set the location of the vertex outside the viewing frustum (or, even better, set its clip position to the wrong side of a clipping plane

Can you provide an example, please? Articles which I found either require newer GLSL version or suggest to cull objects on CPU instead.

An only thing which I managed to get working with GLSL 1.2 is such thing:
```
    if (euclideanDepth > @grassFadeEnd)
       gl_Position.z = 10000000;
```
While it works, it does not improve performance at all.

akortunov:
I only managed to find something like [this](https://www.programmersought.com/article/32552343631), but I did not find a way to assign an arbitrary mesh to the instanced geometry (probably we need somehow to convert the whole mesh to the vertex buffer and color buffer).

psi29a:
> 1. I did not manage to find any reusable instancing example for OSG, which doesn not require OpenGL 3.0 or higher.

Wouldn't this be a case where we are okay with this? We have stated before that for some features, we can fence them off based on GL extension.

We do not support GLES 2 nor 3 and make no claims to do so.

Static batching, works well but not with animations. So animated foliage is probably not going to be a thing in OP/AG unless we implement either dynamic batching or GPU instancing.

Dynamic batching works for very small meshes but is CPU intensive.

GPU Instancing is meant to be used by modern hardware. So this feature will not be supported by GL2 nor GLES2.

GPU Instancing is available on OpenGL >= 4.1+ or ES >= 3.0

akortunov:
> According to Unity (for example): GPU Instancing is available on OpenGL >= 4.1+ or ES >= 3.0

glDrawElementsInstanced / glDrawArraysInstanced should be available since OpenGL 3.1 and GLES 3.0, in OpenGL 2.1 it seems to be available via ARB extensions, but OSG does not handle them out of box.

> Static batching, works well

Unfortunately, only in some cases.

AnyOldName3:
I keep seeing people claim that OSG doesn't support instancing out of the box (including its own mailing list), but it actually lives in the exact same place as all the rest of the rendering. If you have any vertex attributes bound per-instance rather than per-draw-call or per-vertex, it automatically turns on instancing. It'll only work if the OpenGL version is high enough or the ARB or EXT extension exists, but they're present on some *seriously* old hardware. In the compatability database I'm looking at, most of the cards that are listed as incompatible are seriously old ATi cards with Mesa or Gallium drivers that are also listed as compatible when a newer driver is used.

I don't have too much data for how often mobile GPUs expose the extension if they're in ES 2.0 mode but are ES 3.0 capable, but as Psi said, we don't officially support it in the first place, so I think this is a decent candidate for a situation where the fast path is only available on reasonably modern hardware.

As for whether this should be Object Paging's job to handle, I don't see instancing as wildly different from merging objects into the same draw call any other way. It'll be a good tool for it to have in its arsenal, and if it turns out that a cell with the same rock fifty times can be drawn as fast with less memory pressure, that's a win we'll get for free by adding it.

akortunov:
@AnyOldName3, do you have ideas how to make animations to work with object paging?

So far the main blocker is that we do not need to animate weeds roots. Since they are usually near or below mesh pivot point, there is a good assumption "higher gl_Vertex.z -> higher animation speed". Unfortunately, it does not work for pages since technically grass page is one huge plant, partially located underground.

I need somehow to know a relative Z coordinate of each vertex from original mesh in grass shaders.

AnyOldName3:
That's one of the things that'll become easier if instancing is added as a type of batching to object paging. As far as (that part of) the vertex shader's concerned, it's the same mesh as it always was.

akortunov:
> that'll become easier if instancing is added as a type of batching to object paging

I know, but we do not have instancing, and it more likely will not work on all platforms where OpenMW exists.
I suppose we will need to use animations for batches anyway in such case.

AnyOldName3:
It'll work on every platform OpenMW *supports*, so I feel the discussion is academic as it's dumb to throw away performance to make things easier for unsupported platforms. Adding one extra float as a vertex attribute for the Z threshold when merging the buffers would be a solution, though.

psi29a:
Not sure if already talked about, but here is an OSG example:
https://github.com/openscenegraph/OpenSceneGraph/blob/master/examples/osgdrawinstanced/osgdrawinstanced.cpp
that is makes use of the GL_ARB_draw_instanced extension.

psi29a:
Digging deeper is this: https://github.com/openscenegraph/OpenSceneGraph/tree/master/examples/osggpucull

> - instanced rendering capable of drawing thousands of different objects with  almost no CPU intervention  ( cull and draw times are close to 0 ms ).
> - input objects may be sourced from any OSG graph ( for example - information about  object points may be stored in a PagedLOD graph. This way we may cover the whole  countries with trees, buildings and other objects ).  Furthermore if we create osgDB plugins that generate data on the fly, we may generate information for every grass blade for that country.
> - every object may have its own parameters and thus may be distinct from other objects of the same type.
> - relatively low memory footprint ( single object information is stored in a few vertex attributes ).
> - no GPU->CPU roundtrip typical for such methods ( method uses atomic counters  and glDrawArraysIndirect() function instead of OpenGL queries. This way information about quantity of rendered objects never goes back to CPU.   The typical GPU->CPU roundtrip cost is about 2 ms ).
> - this example also shows how to render dynamic objects ( objects that may change its position ) with moving parts ( like car wheels or airplane propellers ) . The obvious extension to that dynamic method would be the animated crowd 
> rendering.
> - rendered objects may be easily replaced ( there is no need to process the whole OSG graphs, because these graphs store only positional information ).

akortunov:
> instancing is added as a type of batching to object paging.

Honestly, I see no point to use instancing in object paging.
From what I can tell, the whole purpose of QuadTreeWorld is to be a terrain implementaion, which is able to load large shapes (chunks) from cached providers (e.g. ChunkManager or ObjectPaging), attach them to the game world and detach them when they are not needed (so such terrain can work with large viewing distance). Instancing works with original meshes and has no relation to chunks, so it is quite strange to use it in the ObjectPaging, which is just a *chunk provider*.

I see two real alternatives here:
1. Do not take in account Android port and add instancing to this PR. In this case we will get pretty the same implementation as in MGE, with MGE-like animations.
2. Use batching instead of instancing. In this case it is possible to use a batch of hacks to improve performance (for example, setup a separate QuadTreeWorld got groundcover objects with separate viewing distance), something like [this](https://github.com/akortunov/openmw/tree/grass_paging). But it is quite hard to make animations to work with this approach.

> the terrain heightmap could be bound as a texture

I suppose that it will not work since by design there can be not-animated meshes (e.g. stones, the whole mesh is below pivot point), or partially animated meshes (e.g. trees, bushes or mushrooms, a part of mesh is below pivot point).
Vertex attributes may work, but I did not find where OSG converts vertex transformations when it merges shapes. Probably it would be better to ask bzzt about it, but it is up to @psi29a.

AnyOldName3:
I think you're misunderstanding what instancing is. It's just another kind of batching, which, for things like grass, *might* give better performance than dumb stick-everything-into-the-same-vertex-buffer-pre-transformed batching, and will definitely use less memory. Object paging has two jobs, and one of them is putting things into batches.

akortunov:
> It's just another kind of batching

It is an another kind of batching, which does not need neither ChunkManager (and ObjectPaging is a ChunkManager) nor QuadTreeWorld (which works with chunks).

> Object paging has two jobs, and one of them is putting things into batches.

Chunk manager has only one job - provide QuadTreeWorld chunks (which are just osg::Node's to attach to scene) for given cells by its request.
The rest is just details of implementation (for example, basic ChunkManager generates chunks by using terrain storage, while ObjectPaging iterates over target cells and merges objects meshes to chunks). The point is that chunks are static, so they can be cached and used for a quite large amount of time. What do you suggest to use as chunks for QuadTreeWorld, when using it with instancing?

AnyOldName3:
What I'm suggesting is to add an extra feature to the optimiser part (which I think might live in sceneutil rather than object paging itself) to do the instancing if it makes more sense as a merge strategy than sticking everything into the same model space and putting it in one big vertex buffer, and then some other stuff either in the loader or object paging's create chunk function that sets up the grass-specific stuff.

akortunov:
@psi29a, if you do not like a check which is based on mesh folder, there is a possibility is just to register groundcover mods separately from normal content files, for example:
```
content=Morrowind.esm
groundcover=Grass Vanilla 1.esp
```
In this case any static object record, added by such content file, is treated as groundcover object (by setting the flag in the ESM::Static). In this case groundcover objects detection can become more reliable and fast. After OpenMW-CS 1.0 release this flag can become a part of ESM format and be modified via editor.

AnyOldName3:
I actually like that a lot more than I like the mesh path check. Everything's explicit and there's no chance of picking things up by mistake.

psi29a:
Actually yes, that is a lot nicer to deal with.

akortunov:
Superseded by #3010.

PR #2669, opened 2020-01-19
Generate WNAM subrecord upon saving terrain instead of upon editing it

Capostrophic:
While unelsson added terrain editing, the way it worked the low-detail heightmap used for unexplored global map was only regenerated when a land record in the plugin was modified by the user of OpenMW-CS. This lead to a complaint that it's impossible to fix the lack of land LOD introduced by TESAnnwyn easily without editing the entire land.

Here's my approach towards working around that. Low-detail heightmap generation is no longer handled by OpenMW-CS itself, and in fact any kind of working with that information from within OpenMW-CS was removed simply because it is redundant and isn't used for anything. Now the LOD is always regenerated every time a land record is saved - I put the original logic into ESM saving. It seems to be sufficiently fast, seeing as it's just a look up of 81 values to put into a record. Low-detail heightmap LOD usage for the map is very 2002, and hopefully we can come up with something better, while leaving the LOD generation up to the ESM writer as an artifact of the legacy format will allow us to maintain the backward compatibility with content files made for Morrowind.
Some could argue that we could allow the user to modify the low-detail heightmap manually through the CS, but it's very difficult to work with and it isn't that useful.
The user that complained about the lack of LOD regeneration tried out this branch and it worked fine for him.

You can try to resave Bethesda game and expansion files into omwgames/omwaddons for example to check if the map looks A-OK.

### DIFF
--- a/apps/opencs/model/world/columnimp.cpp
+++ b/apps/opencs/model/world/columnimp.cpp
@@ -76,53 +76,6 @@ namespace CSMWorld
         return false;
     }
 
-    /* LandMapLodColumn */
-    LandMapLodColumn::LandMapLodColumn()
-        : Column<Land>(Columns::ColumnId_LandMapLodIndex, ColumnBase::Display_String, 0)
-    {
-    }
-
-    QVariant LandMapLodColumn::get(const Record<Land>& record) const
-    {
-        const int Size = Land::LAND_GLOBAL_MAP_LOD_SIZE;
-        const Land& land = record.get();
-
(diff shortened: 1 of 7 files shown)


psi29a:
Can we get @unelsson to validate/review this?

unelsson:
I checked the code through, I couldn't find anything to comment so just gave it thumbs up. I haven't tested it though, not yet at least.

psi29a:
Thank you both! This is how it should be done.  :)

PR #2672, opened 2020-01-21
comply with new osg 3.6.5 OQN API

mp3butcher:


mp3butcher:
arf i missed initial bound...

mp3butcher:
arf i can't tell if there's a difference in behavior with previous version...
Is it normal that sun doesn't flash when entering the frustum?

AnyOldName3:
I see changes in the version less than path.

mp3butcher:
> I see changes in the version less than path.

@AnyOldName3 i don't understand
this pr is simply the consequence of https://github.com/openscenegraph/OpenSceneGraph/pull/902 ....

AnyOldName3:
A change has been made in OSG, so you've added a version check to do something different if a version with that change is present. However, you've changed the code that gets executed when the new change isn't present, too.

mp3butcher:
>  However, you've changed the code that gets executed when the new change isn't present, too.
@AnyOldName3 
I only factorized qg->getQueryGeometry() in 3.4 path, that's not a big deal...No?

apps/openmw/mwrender/sky.cpp
```diff
@@ -557,26 +558,33 @@ class Sun : public CelestialBody
...
         oqn->setQueriesEnabled(true);
-
+#if OSG_VERSION_LESS_THAN(3,6,5)
         // Make it fast! A DYNAMIC query geometry means we can't break frame until the flare is rendered (which is rendered after all the other geometry,
         // so that would be pretty bad). STATIC should be safe, since our node's local bounds are static, thus computeBounds() which modifies the queryGeometry
         // is only called once.
         // Note the debug geometry setDebugDisplay(true) is always DYNAMIC and that can't be changed, not a big deal.
-        oqn->getQueryGeometry()->setDataVariance(osg::Object::STATIC);
+        osg::Geometry* queryGeom = oqn->getQueryGeometry();
+        queryGeom->setDataVariance(osg::Object::STATIC);
```
AnyOldName3: You're not setting the data variance to static in the 3.6.5 path.

PR #2673, opened 2020-01-22
Make Wizard less shy (bug #4493)

Capostrophic:
Wizard will now log its repeating attempts to restart installation media selection dialog and annoy the user with in-the-face messageboxes a bit more. It will tell the user what it couldn't find in the requested folder if the user (either accidentally or deliberately) chooses an invalid installation media/folder for Morrowind or its expansion, e.g. like this:
изображение

### DIFF
--- a/apps/wizard/installationpage.cpp
+++ b/apps/wizard/installationpage.cpp
@@ -154,6 +154,14 @@ void Wizard::InstallationPage::showFileDialog(Wizard::Component component)
         name = QLatin1String("Bloodmoon");
         break;
     }
+    logTextEdit->appendHtml(tr("<p>Attempting to install component %1.</p>").arg(name));
+    mWizard->addLogText(tr("Attempting to install component %1.").arg(name));
+
+    QMessageBox msgBox;
+    msgBox.setWindowTitle(tr("%1 Installation").arg(name));
+    msgBox.setIcon(QMessageBox::Information);
+    msgBox.setText(QObject::tr("Select a valid %1 installation media.<br><b>Hint</b>: make sure that it contains at least one <b>.cab</b> file.").arg(name));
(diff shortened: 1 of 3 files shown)


apps/wizard/installationpage.cpp
```diff
@@ -154,6 +154,14 @@ void Wizard::InstallationPage::showFileDialog(Wizard::Component component)
         name = QLatin1String("Bloodmoon");
         break;
     }
+    logTextEdit->appendHtml(tr("<p>Attempting to install component %1.</p>").arg(name));
+    mWizard->addLogText(tr("Attempting to install component %1.").arg(name));
+
+    QMessageBox msgBox;
+    msgBox.setWindowTitle(tr("%1 Installation").arg(name));
+    msgBox.setIcon(QMessageBox::Information);
+    msgBox.setText(QObject::tr("Select a <b>valid</b> %1 installation media.").arg(name));
```
Mingun: I think, you need explain, what mean "valid", especially as it emphasised. This may look like a mockery of a user who tries to specify a path but doesn't understand what exactly to specify

psi29a:
"Complete" ?

apps/wizard/installationpage.cpp
Capostrophic: Is this better?
изображение

Mingun: Ideally, If you can specify exactly file(s), that is looked for by wizard, for example "folder, where morrowind.exe is located" or something. If indication of cab files is enough, then good (I don't known morrowind directory layout)

Capostrophic: It's not a folder where Morrowind is located (which is completely unrelated). It's the actual Morrowind CD.

PR #2674, opened 2020-01-26
improve storage of vertex colors in gpu buffers

psy-fidelious:
This patch reduces the storage of colors by using bytes instead of floats. This should result in some nice reduction in GPU memory use.

### DIFF
--- a/apps/opencs/model/prefs/setting.cpp
+++ b/apps/opencs/model/prefs/setting.cpp
@@ -61,6 +61,12 @@ double CSMPrefs::Setting::toDouble() const
     return mValues->getFloat (mKey, mParent->getKey());
 }
 
+float CSMPrefs::Setting::toFloat() const
+{
+    QMutexLocker lock(mMutex);
+    return mValues->getFloat(mKey, mParent->getKey());
+}
+
 std::string CSMPrefs::Setting::toString() const
 {
     QMutexLocker lock (mMutex);
--- a/apps/opencs/model/prefs/setting.hpp
+++ b/apps/opencs/model/prefs/setting.hpp
@@ -62,6 +62,8 @@ namespace CSMPrefs
 
             double toDouble() const;
 
+            float toFloat() const;
+
             std::string toString() const;
 
             bool isTrue() const;
--- a/apps/opencs/view/render/cellarrow.cpp
+++ b/apps/opencs/view/render/cellarrow.cpp
@@ -148,12 +148,13 @@ void CSVRender::CellArrow::buildShape()
 
     geometry->addPrimitiveSet (primitives);
 
-    osg::Vec4Array *colours = new osg::Vec4Array;
+    osg::ref_ptr<osg::Vec4ubArray> colours = new osg::Vec4ubArray;
+    colours->setNormalize(true);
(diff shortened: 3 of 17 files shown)


akortunov:
> This should result in some nice reduction in GPU memory use.

Maybe it depends on hardware. I do not see any difference on GTX 1050.
BTW, Morrowind's assets were designed when GPU's had about 32-64 MB of memory.
With HD replacers without shadows GPU's memory usage can be about 300 MB.
Shadows can use additional 300-600 MB of GPU memory, but I am not sure if this PR affects them.

So questions here are:
1. Which improvement exactly should be here (in percents or megabytes of memory)?
2. On which hardware?
3. Why this improvement should be present?

psy-fidelious:
Hi,

1. I agree that compared to texture memory use this probably won't amount to a lot of memory reduction overall. Whatever format the assets are stored in, I assume the original engine used bytes instead of floats to pass colors to the GPU since it was really important at the time to conserve GPU memory. Besides they even used stencil shadows so they only needed to pay the cost of one extra stencil buffer.
Only meshes that use vertex colors will be affected and I really don't know the percentage of them which do to make a proper estimate. I assume it will also be scene dependent. If you know of a good scene where lots of colored meshes are present let me know.

In a quick test while walking in Ald-Ruhn and using the windows task manager's details tab it was hard to say what the difference was because readings fluctuated. Probably the gain is no more than a few megabytes, but it really is hard to tell. I do have shadows enabled but my overall values were around 320MB. Without shadows I was getting something like 270MB, but I hope colors are not used when rendering shadows anyway. Those values were highly inconsistent though, and even reloading a save can alter them completely. Sometimes the unpatched version would win, sometimes the patched version would. Without a common benchmark and settings our readings will never agree.

2. This should affect all hardware, since it allocates smaller vertex buffers. I hope that we can agree that 4 bytes is less than 4 floats regardless of hardware. Now if there's a minimum space requirement for some buffers or whether the driver inserts some kind of padding, it's hard for me to know.

3. If you don't want less memory and memory bandwidth usage on your colored meshes feel free to drop the patch. For me it was a fun coding night.

AnyOldName3:
If vertex colours are floats in the actual Nif files, it's pretty much certain that the original engine uploaded them to the GPU as floats. Bethesda and best practice have never been in the same room as each other.

The original engine using stencil shadows (don't confuse this with shadow volumes, which is another shadow technique that heavily uses the stencil buffer) probably wasn't a memory saving decision. It used one shadow map per actor per light (although with the default settings, it only does this for one light and a few actors). There weren't enough texture slots on a GPU in those days to have that many shadow maps bound, so it *had* to be done after-the-fact instead of as part of the main lighting.

psy-fidelious:
Then keep it as floats sorry for bothering you.

AnyOldName3:
It doesn't necessarily mean this isn't a good change to make, as it shouldn't make anything worse (I doubt that Bethesda were manually specifying float colours that didn't start as ints) and will marginally improve performance. I'd potentially put more of it into the nif loader and less into everywhere else, though.

It's not letting me reopen this as you've deleted the branch.

PR #2676, opened 2020-01-27
comply with new OQN API

mp3butcher:


AnyOldName3:
Issue: https://gitlab.com/OpenMW/openmw/issues/5288

Another Fix: https://gitlab.com/OpenMW/openmw/-/merge_requests/162

apps/openmw/mwrender/sky.cpp
```diff
@@ -557,27 +558,33 @@ class Sun : public CelestialBody
     {
         osg::ref_ptr<osg::OcclusionQueryNode> oqn = new osg::OcclusionQueryNode;
         oqn->setQueriesEnabled(true);
-
+#if OSG_VERSION_LESS_THAN(3,6,5)
+        osg::Geometry* queryGeom = oqn->getQueryGeometry();
```
AnyOldName3: Always `ref_ptr` OSG stuff. Even if it's perfectly safe here, it's safer if other people start changing things later.

apps/openmw/mwrender/sky.cpp
```diff
@@ -557,27 +558,33 @@ class Sun : public CelestialBody
...
         queryGeom->setTexCoordArray(0, mGeom->getTexCoordArray(0), osg::Array::BIND_PER_VERTEX);
-        queryGeom->removePrimitiveSet(0, oqn->getQueryGeometry()->getNumPrimitiveSets());
+        queryGeom->removePrimitiveSet(0, queryGeom->getNumPrimitiveSets());
         queryGeom->addPrimitiveSet(mGeom->getPrimitiveSet(0));
 
+#if OSG_VERSION_LESS_THAN(3,6,5)
         // Hack to disable unwanted awful code inside OcclusionQueryNode::computeBound.
         oqn->setComputeBoundingSphereCallback(new DummyComputeBoundCallback);
         // Still need a proper bounding sphere.
         oqn->setInitialBound(queryGeom->getBound());
```
AnyOldName3: Are these definitely safe to remove?

mp3butcher: yes it's just a hack for previous design

mp3butcher: ....but it worth the check...(even with 3.4 i don't think the initialbound is required as there's a computeboundingshperecallback...it's seams redondant..but perhaps there's stuff in openmw that mandates this initialbound...)

mp3butcher: I just get back to what the code...sorry for previous noise
removing these line is more than safe, it gives better result: glow don't take few frame to display when sun enter frustum

mp3butcher:
ok

PR #2678, opened 2020-01-28
Editor: Delete instances via hotkey

unelsson:
https://gitlab.com/OpenMW/openmw/issues/5172

In scene view, delete selected instances via hotkey, delete button by default. Previously this was only possible via selection menu (right click on selectionmode button in sceneview), which was cumbersome.

I'm not absolutely sure if the shortcut is the correct method to implement this feature though, but seems like the most obvious method.

My concerns (should be tested before merge):
- Is selection cleared after delete, should it be? (done, clear after delete)
- Is function called multiple times without reason, is this a problem? (might be called during keypress multiple times, not a major problem)
- Clean-up (no need for braces in `for` and `if`) (done)

### DIFF
--- a/apps/opencs/view/render/instancemode.cpp
+++ b/apps/opencs/view/render/instancemode.cpp
@@ -10,6 +10,7 @@
 #include "../../model/world/idtree.hpp"
 #include "../../model/world/commands.hpp"
 #include "../../model/world/commandmacro.hpp"
+#include "../../model/prefs/shortcut.hpp"
 
 #include "../widget/scenetoolbar.hpp"
 #include "../widget/scenetoolmode.hpp"
@@ -96,6 +97,9 @@ CSVRender::InstanceMode::InstanceMode (WorldspaceWidget *worldspaceWidget, QWidg
 {
     connect(this, SIGNAL(requestFocus(const std::string&)),
             worldspaceWidget, SIGNAL(requestFocus(const std::string&)));
+
+    CSMPrefs::Shortcut* deleteShortcut = new CSMPrefs::Shortcut("scene-delete", worldspaceWidget);
+    connect(deleteShortcut, SIGNAL(activated(bool)), this, SLOT(deleteSelectedInstances(bool)));
 }
 
 void CSVRender::InstanceMode::activate (CSVWidget::SceneToolbar *toolbar)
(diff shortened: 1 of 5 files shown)


apps/opencs/view/render/instancemode.cpp
```diff
@@ -659,3 +663,24 @@ void CSVRender::InstanceMode::subModeChanged (const std::string& id)
...
+{
+    std::vector<osg::ref_ptr<TagBase> > selection = getWorldspaceWidget().getSelection (Mask_Reference);
+    if (selection.empty()) return;
+
+    CSMDoc::Document& document = getWorldspaceWidget().getDocument();
+    CSMWorld::IdTable& referencesTable = dynamic_cast<CSMWorld::IdTable&> (
+        *document.getData().getTableModel (CSMWorld::UniversalId::Type_References));
+    QUndoStack& undoStack = document.getUndoStack();
+
+    undoStack.beginMacro ("Delete Instances");
```
zinnschlag: This should be using CSMWorld::CommandMacro instead. Things like begin/end, open/close, lock/unlock are an anti-pattern in C++.

unelsson: Looks like CSMWorld::CommandMacro is the same thing, but a bit stricter when it comes to hanging leaving macro hanging open. Good advice! This rings a bell though, I've used beginMacro directly in terrainshapemode, and possible in terraintexturemode, so it's probably a good idea to update them to use CSMWorld::CommandMacro too. Great to see you back!

zinnschlag:
Oops. Completely missed that I had to click the approve button. This is actually the first time that I use the review feature on GitHub.

But I am still not able to test myself (or handle the merge). Getting swamped with stuff, which means I still haven't set up a dev environment for OpenMW. Anyway, this looks ready for merging IMO.

unelsson:
Btw. Delete-key is used in some shortcuts for table views too, therefore if multiple views are open, and have selected data, keyboard shortcuts will act on all views. E.g. you select a cell from Cell list, then an object from worldspacewidget, then press delete-key and you have just deleted the Cell as well as the Object.

This PR doesn't affect this shortcut behavior, but is that still an issue that needs a solution?

zinnschlag:
Crap. That is a major problem. I suggest you file an issue.

unelsson:
I opened an issue at https://gitlab.com/OpenMW/openmw/issues/5301 .

PR #2680, opened 2020-01-28
Fallback to straight path when navmesh and pathgrind are not available

elsid:
Fixes https://gitlab.com/OpenMW/openmw/issues/5262#note_277955138

### DIFF
--- a/apps/openmw/mwmechanics/pathfinding.cpp
+++ b/apps/openmw/mwmechanics/pathfinding.cpp
@@ -314,7 +314,9 @@ namespace MWMechanics
     {
         mPath.clear();
 
-        buildPathByNavigatorImpl(actor, startPoint, endPoint, halfExtents, flags, std::back_inserter(mPath));
+        // If it's not possible to build path over navmesh due to disabled navmesh generation fallback to straight path
+        if (!buildPathByNavigatorImpl(actor, startPoint, endPoint, halfExtents, flags, std::back_inserter(mPath)))
+            mPath.push_back(endPoint);
 
         mConstructed = true;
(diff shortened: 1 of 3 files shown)


apps/openmw/mwmechanics/pathfinding.cpp
```diff
@@ -344,7 +346,7 @@ namespace MWMechanics
             const auto world = MWBase::Environment::get().getWorld();
             const auto stepSize = getPathStepSize(actor);
             const auto navigator = world->getNavigator();
-            navigator->findPath(halfExtents, stepSize, startPoint, endPoint, flags, out);
+            return navigator->findPath(halfExtents, stepSize, startPoint, endPoint, flags, out).is_initialized();
         }
         catch (const DetourNavigator::NavigatorException& exception)
         {
```
akortunov: This path does not not return value, despite the function returns boolean value.

elsid: Thanks, missed this.

akortunov: Are you sure that it should return "true" instead of "false"? 
As I understood, if an attempt to build path leads to exception, the resulting path may be an incorrect, and in this case  it makes sense to fallback to straight path as well.

elsid: No, it should be true, because it indicates the fact that there was a try to find a path by navmesh. If the try is failed then the destination can’t be used for wandering actor. If it will be false then actor will wander to any destination and use of navmesh in AiWander will became useless.

PR #2682, opened 2020-01-29
Fix #5264 by putting current behaviour behind a setting

Assumeru:
Think there's value to changing the setting to an int so it can have 3 modes (vanilla, mcp, openmw)?


Assumeru:
Closing in favour of #2684

PR #2684, opened 2020-01-30
Make uncapped Damage Fatigue optional (bug #5264)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5264)

A competitor to #2682.

In comparison to Assumeru's approach, this:
1) Updates the documentation
2) Adds the setting into Advanced tab of the launcher
3) Makes Absorb Fatigue strictly vanilla-like
4) Probably has slightly better performance as the setting value is truly read-only
5) And basically acts as a 1:1 MCP option equivalent.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -1199,10 +1199,10 @@ namespace MWMechanics
         return false;
     }
 
-    void adjustDynamicStat(CreatureStats& creatureStats, int index, float magnitude)
+    void adjustDynamicStat(CreatureStats& creatureStats, int index, float magnitude, bool allowDecreaseBelowZero = false)
     {
         DynamicStat<float> stat = creatureStats.getDynamic(index);
-        stat.setCurrent(stat.getCurrent() + magnitude, index == 2);
+        stat.setCurrent(stat.getCurrent() + magnitude, allowDecreaseBelowZero);
         creatureStats.setDynamic(index, stat);
     }
 
@@ -1241,9 +1241,12 @@ namespace MWMechanics
 
         case ESM::MagicEffect::DamageMagicka:
         case ESM::MagicEffect::DamageFatigue:
-            adjustDynamicStat(creatureStats, effectKey.mId-ESM::MagicEffect::DamageHealth, -magnitude);
+        {
+            int index = effectKey.mId-ESM::MagicEffect::DamageHealth;
+            static const bool uncappedDamageFatigue = Settings::Manager::getBool("uncapped damage fatigue", "Game");
+            adjustDynamicStat(creatureStats, index, -magnitude, index == 2 && uncappedDamageFatigue);
             break;
-
+        }
         case ESM::MagicEffect::AbsorbHealth:
             if (magnitude > 0.f)
                 receivedMagicDamage = true;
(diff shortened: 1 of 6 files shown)


elsid:
Not sure I understand what's the idea behind PR competition. You could just leave a review asking to add the stuff you think is required for PR to be complete. What is the goal? Should we wait for more competitors and then vote for PRs? Or who will decide which one to merge?

akortunov:
Not sure about which kind of review to you want - I do not understand how this feature is supposed to work.

Capostrophic:
PR competition is a figure of speech. Telling Assumeru "You must rewrite *everything* tbh" doesn't seem to be a good idea.

@akortunov, basically check if actors stop getting knocked down with Absorb Fatigue and Damage Fatigue effects active and see if the new option changes Damage Fatigue to allow actors to get knocked down as intended. (Works in my testing.)

Assumeru:
> 
> 
> PR competition is a figure of speech. Telling Assumeru "You must rewrite _everything_ tbh" doesn't seem to be a good idea.

I wouldn't have minded, but this seems like a more efficient use of time.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -1199,10 +1199,10 @@ namespace MWMechanics
         return false;
     }
 
-    void adjustDynamicStat(CreatureStats& creatureStats, int index, float magnitude)
+    void adjustDynamicStat(CreatureStats& creatureStats, int index, float magnitude, bool capAt0 = true)
     {
         DynamicStat<float> stat = creatureStats.getDynamic(index);
-        stat.setCurrent(stat.getCurrent() + magnitude, index == 2);
+        stat.setCurrent(stat.getCurrent() + magnitude, !capAt0);
```
elsid: Use parameter name from [this](https://github.com/OpenMW/openmw/blob/d7989ac7782272628745eb7d920dc477d57c6809/apps/openmw/mwmechanics/stat.hpp#L101) method (`allowDecreaseBelowZero`) to remove boolean conversions.

apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -1199,10 +1199,10 @@ namespace MWMechanics
         return false;
     }
 
-    void adjustDynamicStat(CreatureStats& creatureStats, int index, float magnitude)
+    void adjustDynamicStat(CreatureStats& creatureStats, int index, float magnitude, bool allowDecreaseBelowZero = false)
     {
         DynamicStat<float> stat = creatureStats.getDynamic(index);
-        stat.setCurrent(stat.getCurrent() + magnitude, index == 2);
+        stat.setCurrent(stat.getCurrent() + magnitude, allowDecreaseBelowZero);
```
elsid: What about the case when `index == 2`? When this function is called  not for `DamageFatigue`? Now it always be not allowed to be negative. Not sure where it's expected.

You can make `allowDecreaseBelowZero  = true` by default. Change
```suggestion
        stat.setCurrent(stat.getCurrent() + magnitude, index == 2 && allowDecreaseBelowZero);
```
And call as:
```c++
adjustDynamicStat(creatureStats, effectKey.mId-ESM::MagicEffect::DamageHealth, -magnitude, uncappedDamageFatigue);
```

Capostrophic: adjustDynamicStat() is only called in effectTick(), and only for the following effects:
Restore effects (don't need to be uncapped)
Damage effects (already handled)
Absorb effects (they all need to be capped)
Sun damage (doesn't need to be uncapped, characters die when Health reaches 0)
Fire, Frost, Shock, Poison damage effects (don't need to be uncapped, characters die when Health reaches 0)

And we *are* only interested in uncapping for Damage Fatigue effect.

Limiting it for fatigue effects seems to be unnecessary, and I'm pretty sure assuming the "default" case is that no uncapping takes place is reasonable.

PR #2686, opened 2020-02-01
Use cmake method to enable interprocedural optimizations

Eli2:
[Bug #5265](https://gitlab.com/OpenMW/openmw/issues/5265)
Edit:
I have not tested this on windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,11 @@ endif()
 # set the minimum required version across the board
 cmake_minimum_required(VERSION 3.1.0)
 
+# for link time optimization, remove if cmake version is >= 3.9
+if(POLICY CMP0069)
+    cmake_policy(SET CMP0069 NEW)
+endif()
+
 project(OpenMW)
 
 # If the user doesn't supply a CMAKE_BUILD_TYPE via command line, choose one for them.
@@ -100,6 +105,7 @@ option(OSG_STATIC "Link static build of OpenSceneGraph into the binaries" FALSE)
 option(QT_STATIC "Link static build of QT into the binaries" FALSE)
 
 option(OPENMW_UNITY_BUILD "Use fewer compilation units to speed up compile time" FALSE)
+option(OPENMW_LTO_BUILD "Build OpenMW with Link-Time Optimization (Needs ~2GB of RAM)" OFF)
 
 # what is necessary to build documentation
 IF( BUILD_DOCS )
@@ -113,7 +119,6 @@ option(OPENMW_OSX_DEPLOYMENT OFF)
 
 if (MSVC)
     option(OPENMW_MP_BUILD "Build OpenMW with /MP flag" OFF)
-    option(OPENMW_LTO_BUILD "Build OpenMW with Link-Time Optimization (Needs ~2GB of RAM)" OFF)
 endif()
 
 # Set up common paths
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -27,6 +27,11 @@ endif()
 # set the minimum required version across the board
 cmake_minimum_required(VERSION 3.1.0)
 
+# for link time optimization, remove if cmake version is >= 3.9
+if(POLICY CMP0069)
+    cmake_policy(SET CMP0069 NEW)
+endif()
+
```
AnyOldName3: Why are we messing with the policy? IIRC, the history of the option was:

* It didn't exist. (This was pre-CMake-3.0, and I'm pretty sure we use CMake 3.0 stuff to link with RecastNavigation, so this reality isn't relevant to us.)
* It was added. Turning it on would enable LTO when compatible build tools were used, and silently fail (or maybe just log that it was incompatible) if incompatible tools were used.
* CMake 3.9 came and added policy CMP0069 and `check_ipo_supported`. If you turned the option on for an unsupported set of tools, it was now an error, so if you wanted to turn it into a warning or give a custom error message, you needed to do the new check.

To me, this means that the most sensible option (and what I *think* the snippet of CMake's own `CMakeLists.txt` does) is:
* Check the option is set, if it is, keep going, otherwise, skip over all of this.
* If CMake ≥ 3.9 is used and the policy is set to the new value (checking the policy on its own is fine as that implies the version is sufficient), then checking the new check is necessary, so do that, otherwise skip to setting the setting.
* If the check fails, yell at the user and don't set the option.
* Set the option.

CMakeLists.txt
```diff
@@ -411,6 +416,23 @@ endif()
...
+        include(CheckIPOSupported)
+        check_ipo_supported(RESULT HAVE_IPO OUTPUT HAVE_IPO_OUTPUT)
+        if(HAVE_IPO)
+            message(STATUS "LTO enabled")
+            set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+        else()
+            message(WARNING "Requested option OPENMW_LTO_BUILD not supported by this compiler: ${HAVE_IPO_OUTPUT}")
+        endif()
+    else()
+        message(WARNING "Requested option OPENMW_LTO_BUILD not supported by this cmake version: ${CMAKE_VERSION}")
```
AnyOldName3: I'm pretty sure it *will* be supported, just not necessarily for the current compiler.

hristoast:
>  I have not tested this on windows.

I've built it using VS2019 on Windows 10, everything checks out.

CMakeLists.txt
Eli2: > Why are we messing with the policy?
>     * It didn't exist. (This was pre-CMake-3.0 ...

Policy CMP0069 was introduced in cmake 3.9
https://cmake.org/cmake/help/v3.9/policy/CMP0069.html

> It was added. Turning it on would enable LTO ...
If i have understood correctly that is not what it does
`This policy takes effect whenever the IPO property is enabled. The OLD behavior for this policy is to add IPO flags only for Intel compiler on Linux. The NEW behavior for this policy is to add IPO flags for the current compiler or produce an error if CMake does not know the flags.`

So you want me to check against the cmake version 3.9 instead of checking if the policy that was added in this verson exists ?

Maybe the confusion is this:
The example posted in the ticket uses:
cmake_minimum_required(VERSION 3.1...3.15 FATAL_ERROR)
we use:
cmake_minimum_required(VERSION 3.1.0)
wich forces the policy to be unset.
https://cmake.org/cmake/help/v3.1/command/cmake_policy.html#setting-policies-by-cmake-version

CMakeLists.txt
Eli2: I am not sure how much work we should put into supporting:
icc users on linux with cmake < 3.9

AnyOldName3: Now I'm aware of how bad the setting used to be, I see your point. I don't necessarily agree that it's a good idea to do it quite like this. It's probably better to log that it *probably* isn't supported, and try setting it anyway.

AnyOldName3: E.g. if someone's stuck on a distro with ancient CMake, they could provide their own toolchain file for their tools, and include the LTO flags in that, and CMake would suddenly be able to set them.

CMakeLists.txt
```diff
@@ -411,6 +416,23 @@ endif()
 
 # CXX Compiler settings
 set(CMAKE_CXX_STANDARD 14)
+
+if(OPENMW_LTO_BUILD)
+    if(POLICY CMP0069)
```
AnyOldName3: Now we both have a better idea of what's going on, I'd not check the policy here, just the version. If the version is high enough, we can still do the check, even if we don't use the result of it to decide what to do, just what to log.

CMakeLists.txt
AnyOldName3: I'm pretty sure we also set the minimum version to other values in other places, so now you've reminded me that setting the minimum version also sets its policies, I'm not even sure this will stick throughout the whole CMake process.

My preference would be to make the place where we actually set it able to work no matter what the policy is set to. If it turns out that the old policy also uses CMake 3.8's restricted list of tools rather than just mandating the check, then we should set the new policy right before we do the thing that requires it.

akortunov:
Are there plans to continue development of this PR?

Eli2:
I consider this pr complete.

akortunov:
> I consider this pr complete.

I believe that it was not merged in February because of concerns left.

akortunov:
@AnyOldName3, can you take a look again?

akortunov:
Any update here?

AnyOldName3:
Superseded by https://gitlab.com/OpenMW/openmw/-/merge_requests/293

PR #2687, opened 2020-02-02
Re-work wandering outside of initial cell

akortunov:
Fixes [bug #5261](https://gitlab.com/OpenMW/openmw/issues/5261), [bug #5262](https://gitlab.com/OpenMW/openmw/issues/5261). Also partially fixes [bug #5217](https://gitlab.com/OpenMW/openmw/issues/5217) (the bug can still re-appear when actor is not in his initiall cell after save/load).

Summary of changes:
1. Do not reset wandering distance to 0 when actor in cell without pathgrid crosses cell borders (fixes 5261 for new saves).
2. Do not reset an initial position when actor crosses cell borders or enters combat (we reset temporary AI storage in this case) (fixes 5262).
3. Do not try to return actor to his pre-combat position, if he moved too far away from it (due to AiTravel limitations).

Note: we need to determine actor's initial cell and use it to build pathgrid (instead of actor's current cell) to fix the 5217 properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -195,6 +195,7 @@
     Bug #5249: Wandering NPCs start walking too soon after they hello
     Bug #5250: Creatures display shield ground mesh instead of shield body part
     Bug #5255: "GetTarget, player" doesn't return 1 during NPC hello
+    Bug #5261: Creatures can sometimes become stuck playing idles and never wander again
     Feature #1774: Handle AvoidNode
     Feature #2229: Improve pathfinding AI
     Feature #3025: Analogue gamepad movement controls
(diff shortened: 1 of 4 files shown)


elsid:
[#5261](https://gitlab.com/OpenMW/openmw/issues/5261) can't be fully fixed like this. Master logic breaks game save because zeroed wandering distance is saved. So after loading creature will never wander again.

akortunov:
> can't be fully fixed like this. Master logic breaks game save because zeroed wandering distance is saved. So after loading creature will never wander again.

If you mean that this PR does not fix existing saves, there is nothing we can do with them since affected characters are already altered.

elsid:
Why do we save wandering distance? What does it give us? Could we just use initial value and if for some reason we need to adjust it, we apply this logic every load.

akortunov:
> Why do we save wandering distance? What does it give us?

There are cases when actor get an AiWander package via script (e.g. followers), and such packages still need to be persistent, so we should store wandering distance in savegame as well.

elsid:
Ok, but only initial distance should be saved, not the modified one.

akortunov:
> Ok, but only initial distance should be saved, not the modified one.

There is no way to tell if the distance is initial or modified for old saves.
And since we do not alter distance with this PR, for new saves an initial one will be stored anyway.

PR #2689, opened 2020-02-02
Some more minor NIF support fixes

Capostrophic:
Although we "load" 4.0.0.0 NIF files since last Spring, the loading will break horribly in some specific cases, according to NifTools research:

1. An unused always active flag in NiGeomMorpherController was added in 4.0.0.2.
2. LOD center position was added in 4.0.0.2 revision of NiLODNodes.
3. NiSkinPartition links were added to 4.0.0.2 NiSkinData. They're not used.

I've added version checks for that because 4.0.0.0 is still not really a different NIF version and one day someone may naively export a file with morphing/skinning/LOD information in this format. Morrowind apparently can load 3.x files as well, but hopefully nobody tried to use them for a mod yet so I can get to that later.

LOD node center will just be (0,0,0) if it's not declared in the file explicitly.

Some more typedefs were converted to using directives.

Material color controller is now a target color field in the loaded record itself (in later revisions it's not a part of flags).

Internal flag in NiSourceTexture is now handled like in NifSkope. The internal texture will not be loaded if it's off. Other NiSourceTexture fields have the correct unsigned type.

A slightly more elegant way to compare generic version numbers with the file version has been added (NIFVersion enum becomes extremely bloated otherwise).

So with this OpenMW will finally be able to support 2 (!!!) NIF revisions.

### DIFF
--- a/components/nif/base.hpp
+++ b/components/nif/base.hpp
@@ -59,7 +59,7 @@ public:
         controller.post(nif);
     }
 };
-typedef Named NiSequenceStreamHelper;
+using NiSequenceStreamHelper = Named;
 
 } // Namespace
 #endif
--- a/components/nif/controlled.cpp
+++ b/components/nif/controlled.cpp
@@ -10,17 +10,18 @@ namespace Nif
         Named::read(nif);
 
         external = nif->getChar() != 0;
-        if(external)
+        bool internal = false;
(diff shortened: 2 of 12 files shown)


psi29a:
Nice work, code is straightforward and easier to read. Cheers.

PR #2691, opened 2020-02-04
Use status codes to handle navigator errors instead of exceptions

elsid:
For find path use case.


psi29a:
Thanks for the extra pair of eyes @akortunov 
@elsid I like that we remove some boost::optional :)

PR #2692, opened 2020-02-07
Editor: Implement brush outlines

unelsson:
Implements a tool outline for TerrainShapeMode and TerrainTextureMode.

Notes:
-  Point tool is represented by a small cross. For some reason, when I tried to use osg::PrimitiveSet::POINTS the pointer became invisible. All outlines use LINES in this PR, but on the other hand, TRIANGLES could be used to generate 3d meshes.
- Circle and Square tool are implemented as mathematical representations of size, but the actual edits that go to data are rounded to nearest, therefore outline might feel strange at some brush sizes. It might be more intuitive to have exact representations, but if the underlying data is changed (e.g. cell size is different, the exact representation has to take this into account).
- Custom tool shape is not implemented. I suggest to cut this feature out to a separate PR.
- Outline snaps to vertex or texture grid.
- Color of the outline is white, it's the same color as terrain selection. Is that correct?
- some if-oneliners (I will fix this)
- Updates terrainselection to use CSMWorld::CellCoordinates::textureGlobalXToWorldCoords and CSMWorld::CellCoordinates::textureGlobalYToWorldCoords in order to make it more readable, but also to be consistent on the method with selection and outline drawing.
- Updates terraintexturemode's variables to use same formatting as terrainshapemode, this is needed for brush shape so that brushdraw gets consistent values.

Lines-implementation https://imgur.com/a/2E4eR99

https://gitlab.com/OpenMW/openmw/issues/5201

### DIFF
--- a/apps/opencs/view/render/brushdraw.cpp
+++ b/apps/opencs/view/render/brushdraw.cpp
@@ -1,0 +1,136 @@
+#include "brushdraw.hpp"
+
+#include <osg/Group>
+#include <osg/Geometry>
+#include <osg/Array>
+#include <osg/BlendFunc>
+#include <osg/CullFace>
+#include <osg/Material>
+
+#include <osgUtil/LineSegmentIntersector>
+
+#include "mask.hpp"
+
+CSVRender::BrushDraw::BrushDraw(osg::Group* parentNode) :
+    mParentNode(parentNode)
+{
+    mBrushDrawNode = new osg::Group();
+    mGeometry = new osg::Geometry();
+    mBrushDrawNode->addChild(mGeometry);
+    mParentNode->addChild(mBrushDrawNode);
+}
+
+CSVRender::BrushDraw::~BrushDraw()
+{
+    if (mBrushDrawNode->containsNode(mGeometry)) mBrushDrawNode->removeChild(mGeometry);
+    if (mParentNode->containsNode(mBrushDrawNode)) mParentNode->removeChild(mBrushDrawNode);
+}
+
+float CSVRender::BrushDraw::getIntersectionHeight (const osg::Vec3d& point)
+{
+    osg::Vec3d start = point;
+    osg::Vec3d end = point;
+    start.z() += 8000.0f; // these numbers need fixing
+    end.z() -= 8000.0f;
+    osg::Vec3d direction = end - start;
+
+    // Get intersection
+    osg::ref_ptr<osgUtil::LineSegmentIntersector> intersector (new osgUtil::LineSegmentIntersector(
+        osgUtil::Intersector::MODEL, start, end) );
+    intersector->setIntersectionLimit(osgUtil::LineSegmentIntersector::NO_LIMIT);
+    osgUtil::IntersectionVisitor visitor(intersector);
+
+    visitor.setTraversalMask(Mask_Terrain);
+
+    mParentNode->accept(visitor);
+
+    for (osgUtil::LineSegmentIntersector::Intersections::iterator it = intersector->getIntersections().begin();
+         it != intersector->getIntersections().end(); ++it)
+    {
+        osgUtil::LineSegmentIntersector::Intersection intersection = *it;
+
+        // reject back-facing polygons
+        if (direction * intersection.getWorldIntersectNormal() > 0)
+        {
+            continue;
+        }
+
+        return intersection.getWorldIntersectPoint().z();
+    }
+    return 0.0f;
+}
+
+
+void CSVRender::BrushDraw::buildGeometry(const float& radius, const osg::Vec3d& point, int amountOfPoints)
+{
+    osg::ref_ptr<osg::Geometry> geom = new osg::Geometry();
+    osg::ref_ptr<osg::Vec3Array> vertices (new osg::Vec3Array());
+    osg::ref_ptr<osg::Vec4Array> colors (new osg::Vec4Array());
+    const float step ((osg::PI * 2.0f) / static_cast<float>(amountOfPoints));
+
+    for (int i = 0; i < amountOfPoints + 2; i++)
+    {
+        float angle (static_cast<float>(i) * step);
+        vertices->push_back(osg::Vec3d(
+            point.x() + radius * cosf(angle),
+            point.y() + radius * sinf(angle),
+            getIntersectionHeight(osg::Vec3d(
+                point.x() + radius * cosf(angle),
+                point.y() + radius * sinf(angle),
+                point.z()) )));
+        colors->push_back(osg::Vec4f(
+            50.0f,
+            50.0f,
+            50.0f,
+            100.0f));
+        angle = static_cast<float>(i + 1) * step;
+        vertices->push_back(osg::Vec3d(
+            point.x() + radius * cosf(angle),
+            point.y() + radius * sinf(angle),
+            getIntersectionHeight(osg::Vec3d(
+                point.x() + radius * cosf(angle),
+                point.y() + radius * sinf(angle),
+                point.z()) ) + 200.0f));
+        colors->push_back(osg::Vec4f(
+            50.0f,
+            50.0f,
+            50.0f,
+            100.0f));
+    }
+
+    geom->setVertexArray(vertices);
+    geom->setColorArray(colors, osg::Array::BIND_PER_VERTEX);
+    geom->addPrimitiveSet(new osg::DrawArrays(osg::PrimitiveSet::TRIANGLE_STRIP, 0, (amountOfPoints + 2) * 2 - 2));
+    mGeometry = geom;
+}
+
+void CSVRender::BrushDraw::update(osg::Vec3d point, int brushSize)
+{
+    if (mBrushDrawNode->containsNode(mGeometry)) mBrushDrawNode->removeChild(mGeometry);
+    mBrushDrawNode->setNodeMask (Mask_EditModeCursor);
+    float radius = static_cast<float>(brushSize * mLandSizeFactor);
+    int amountOfPoints = (osg::PI * 2.0f) * radius / 20;
+
+    buildGeometry(radius, point, amountOfPoints);
+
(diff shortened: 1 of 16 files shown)


apps/opencs/view/render/brushdraw.cpp
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        return intersection.getWorldIntersectPoint().z();
+    }
+    return 0.0f;
+}
+
+void CSVRender::BrushDraw::buildPointGeometry(const osg::Vec3d& point)
+{
+    osg::ref_ptr<osg::Geometry> geom = new osg::Geometry();
```
zinnschlag: Why is that a = while the next two are using () ?

apps/opencs/view/render/brushdraw.cpp
```diff
@@ -0,0 +1,308 @@
...
+    osg::ref_ptr<osg::Vec3Array> vertices (new osg::Vec3Array());
+    osg::ref_ptr<osg::Vec4Array> colors (new osg::Vec4Array());
+    const int amountOfPoints = (osg::PI * 2.0f) * radius / 20;
+    const float step ((osg::PI * 2.0f) / static_cast<float>(amountOfPoints));
+    const float brushOutlineHeight (1.0f);
+    osg::Vec4f lineColor(1.0f, 1.0f, 1.0f, 0.6f);
+
+    for (int i = 0; i < amountOfPoints + 2; i++)
+    {
+        float angle (static_cast<float>(i) * step);
```
zinnschlag: This static_cast should be unnecessary, since step is a float. int * float is a float.

apps/opencs/view/render/terrainshapemode.cpp
```diff
@@ -49,6 +50,12 @@ CSVRender::TerrainShapeMode::TerrainShapeMode (WorldspaceWidget *worldspaceWidge
 {
 }
 
+CSVRender::TerrainShapeMode::~TerrainShapeMode ()
+{
+    if (mBrushDraw)
+        mBrushDraw.reset();
```
zinnschlag: I don't understand the purpose of this code.

apps/opencs/view/render/terrainshapemode.hpp
```diff
@@ -57,6 +58,7 @@ namespace CSVRender
 
             /// Editmode for terrain shape grid
             TerrainShapeMode(WorldspaceWidget*, osg::Group* parentNode, QWidget* parent = nullptr);
+            ~TerrainShapeMode();
```
zinnschlag: This should have a virtual. Not strictly needed, but helps with clarity.

apps/opencs/view/render/terraintexturemode.cpp
```diff
@@ -328,11 +342,9 @@ void CSVRender::TerrainTextureMode::editTerrainTextureGrid(const WorldspaceHitRe
     std::string mBrushTextureInt = mBrushTexture.substr (hashlocation+1);
     int brushInt = stoi(mBrushTexture.substr (hashlocation+1))+1; // All indices are offset by +1
 
-    int rf = mBrushSize / 2;
-    int r = mBrushSize / 2 + 1;
-    int distance = 0;
+    int r = static_cast<float>(mBrushSize) / 2;
```
zinnschlag: This whole function is way too long and too repetitive. Major candidate for refactoring. But that is probably best done in a separate PR. Okay for now.

apps/opencs/view/render/terrainshapemode.cpp
unelsson: I wanted to make sure that the pointer would be dropped, but now to think of it, isn't std::unique_ptr exactly the kind of smart pointer that will be destroyed automatically anyway?

apps/opencs/view/render/terraintexturemode.hpp
```diff
@@ -102,14 +107,15 @@ namespace CSVRender
             bool allowLandTextureEditing(std::string textureFileName);
 
             std::string mCellId;
-            std::string mBrushTexture;
-            int mBrushSize;
-            int mBrushShape;
+            std::string mBrushTexture = "L0#0";
```
zinnschlag: Not sure which rules we have adopted for C+11 in this regard (that was probably while I was away). But IMO it is a bad idea to initialise variables this way, if the class also has a constructor. Use one or the other. Don't mix them.

apps/opencs/view/render/brushdraw.cpp
unelsson: Right... I've had some trouble occasionally with calculations having both float and int, static_cast was done here to enforce that I get a float calculation and I can be absolutely sure this works. Removing that should be anyway a minor optimization and easier code to read, so I'll do that.

apps/opencs/view/render/terraintexturemode.hpp
unelsson: For terrainshapemode, I got a review from Elsid, who suggested moving initialization here as for some C-standard (maybe C+11 standard) , but of course it should be consistently the same way. This was adopted for terrainshapemode, so this is the reason I'm doing the same here, as I was forced to do minor changes here anyway.

unelsson: Suggesting to fix this so that I remove these from terraintexturemode.cpp, and keep initialization only at the header. This is a matter of standard, and I have nothing to argue why one is better than the other, so will do either way.

apps/opencs/view/render/terrainshapemode.cpp
zinnschlag: Exactly.

apps/opencs/view/render/terraintexturemode.hpp
zinnschlag: I do. If you have it in the header, than any modification has the potential to trigger re-compiling of mulitple cpp files, while making a change to a cpp will only force re-compliation of a single cpp file.
We need to write something about that into our style guide, if we haven't already. IMO initialising members in the class definition should be limited to basic classes that are just collections of data (typically structs) and constants where feasible.

unelsson: I now removed the initializations from terraintexturemode header. Terrainshapemode is untouched, suggesting to change that to same style in another PR.

apps/opencs/view/render/terrainshapemode.hpp
unelsson: Removed it altogether in 66fba7c .

apps/opencs/view/render/terraintexturemode.hpp
elsid: Using in-class initializers is a part of C++ Core Guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers. Of course we can create our own different style guide rules but then we face [this](https://imgs.xkcd.com/comics/standards.png). When there is an existing standard covering some case better to use it rather than propose another standard. Also it's always possible to propose a change to a standard. Argument about compilation time should be based on some statistics (how often defaults change per commit). Because if all default initializations changes increase compilation time across all builds by insignificant percent then there is no much sense to have this argument. I think having less copy-paste and no forgotten default initalizations is a stronger argument.

zinnschlag: Admittedly the compile time argument is pretty weak. I only brought it up to show there is one, after unelsson wasn't able to find any.

I don't care hugely about other people's style guide, particularly in regards to c++11 which is still pretty new. Each project should define its own, but some conformity is certainly desirable. Overall the question of prefer constructor or prefer in class initialisation is again subjective. My main point here was that we shouldn't mix them in the same class. Because that is not subjective at all.

psi29a: I agree with this sentiment.

To be fair, we're on C++14 though switching to C++17 is possible, we're kneecapped by MacOS and Android not supporting nice things like std::filesystem (yet).

psi29a:
How are we doing here? Is this ready for 0.46 or should we see this merged for 0.47?

unelsson:
> How are we doing here? Is this ready for 0.46 or should we see this merged for 0.47?

With caution, I'm suggesting to merge this to 0.46. I've tested it myself, and I like it, but what do the reviewers think?

psi29a:
@zinnschlag any comments?

apps/opencs/view/render/terraintexturemode.cpp
psi29a: @unelsson have you created a separate issue to track this?

psi29a:
@hristoast whatcha think? :)

hristoast:
Looks good, works as expected.  :shipit:

apps/opencs/view/render/brushdraw.cpp
```diff
@@ -0,0 +1,308 @@
...
+            point.z()) ) + brushOutlineHeight));
+    colors->push_back(lineColor);
+
+    geom->setVertexArray(vertices);
+    geom->setColorArray(colors, osg::Array::BIND_PER_VERTEX);
+    geom->addPrimitiveSet(new osg::DrawArrays(osg::PrimitiveSet::LINES, 0, 4));
+    mGeometry = geom;
+}
+
+void CSVRender::BrushDraw::buildSquareGeometry(const float& radius, const osg::Vec3d& point)
```
akortunov, line 116: Just for future reference: there is no point to pass float by references, it does not make code faster.

apps/opencs/view/render/brushdraw.cpp
```diff
@@ -0,0 +1,308 @@
...
+{
+    osg::ref_ptr<osg::Geometry> geom (new osg::Geometry());
+    osg::ref_ptr<osg::Vec3Array> vertices (new osg::Vec3Array());
+    osg::ref_ptr<osg::Vec4Array> colors (new osg::Vec4Array());
+
+    const float brushOutlineHeight (1.0f);
+    float diameter = radius * 2;
+    int resolution = (diameter / mLandSizeFactor) * 2; //half a vertex resolution
```
akortunov, line 124: Are you sure that you really want to use 2 diameters instead of two radiuses here?

PR #2694, opened 2020-02-09
Make Show fallback to global variables when sensible (bug #5278)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/5278).

In case there isn't a local variable for the given object, Show will try to find a global namesake.

The way stringstream emptiness is detected is somewhat clunky, but I don't want to actually recover the string that is put into it before it's actually necessary.

### DIFF
--- a/apps/openmw/mwscript/miscextensions.cpp
+++ b/apps/openmw/mwscript/miscextensions.cpp
@@ -923,11 +923,7 @@ namespace MWScript
                 if (!ptr.isEmpty())
                 {
                     const std::string& script = ptr.getClass().getScript(ptr);
-                    if (script.empty())
-                    {
-                        output << ptr.getCellRef().getRefId() << " has no script " << std::endl;
-                    }
-                    else
+                    if (!script.empty())
                     {
                         const Compiler::Locals& locals =
                             MWBase::Environment::get().getScriptManager()->getLocals(script);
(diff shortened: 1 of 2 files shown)


apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -923,11 +923,7 @@ namespace MWScript
                 if (!ptr.isEmpty())
                 {
                     const std::string& script = ptr.getClass().getScript(ptr);
-                    if (script.empty())
-                    {
-                        output << ptr.getCellRef().getRefId() << " has no script " << std::endl;
```
elsid: How do a user see a difference is it a global variable or a script variable?

apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -941,13 +937,11 @@ namespace MWScript
                         case 'f':
                             output << ptr.getCellRef().getRefId() << "." << var << ": " << ptr.getRefData().getLocals().getFloatVar(script, var);
                             break;
-                        default:
-                            output << "unknown local '" << var << "' for '" << ptr.getCellRef().getRefId() << "'";
-                            break;
+                        // Do nothing otherwise
```
elsid: Useless comment. It's obvious that there is nothing else in this switch statement.

Capostrophic: Arguably it demonstrates to the reader that all possible switch cases were considered. Ok, I'll remove it.

apps/openmw/mwscript/miscextensions.cpp
Capostrophic: I changed the format to the way vanilla does it with slight deviations (a member separator, dot instead of ->, the case of the original variable name isn't preserved). Hopefully it's clear enough.

PR #2695, opened 2020-02-09
Add more destination checks for AiWander without pathgrid

elsid:
Should partially fix [#5260](https://gitlab.com/OpenMW/openmw/issues/5260).

1. Check for destination visibility by casting a ray. This should prevent actors collision on path crossing.
2. Test for aabb collision at destination with other actors using a sphere. Ray cast may not be sufficient to verify that destination area has enough space for actor to stand. Full physics test might be expensive so for now only actors.
3. Apply 1 and 2 in each actor reaction to minimize simultaneously wandering actors collision.
4. Add one more metric to obstacle checker.
5. Print actor base ptr on betacomment command. Makes debug easier when there are multiple actors with the same name and RefID on a scene.

### DIFF
--- a/apps/openmw/mwscript/miscextensions.cpp
+++ b/apps/openmw/mwscript/miscextensions.cpp
@@ -1254,6 +1254,7 @@ namespace MWScript
                     msg << "[Deleted]" << std::endl;
 
                 msg << "RefID: " << ptr.getCellRef().getRefId() << std::endl;
+                msg << "Memory address: " << ptr.getBase() << std::endl;
 
                 if (ptr.isInCell())
                 {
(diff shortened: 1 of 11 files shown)


apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -1254,6 +1254,7 @@ namespace MWScript
                     msg << "[Deleted]" << std::endl;
 
                 msg << "RefID: " << ptr.getCellRef().getRefId() << std::endl;
+                msg << "Ptr: " << ptr.getBase() << std::endl;
```
akortunov: The string like `Ptr: 0x55df90966fb0` is not informative enough for modmakers, IMO.
How about "Memory address" or something like that?

elsid: Done

PR #2697, opened 2020-02-11
Don't try to attach absent (empty path) shield models to NPCs

Capostrophic:
Morrowind seems to hide such shields instead of trying to find a model with no path given, so let's do the same thing.


akortunov:
Just for reference: there is a similar quirk for equippable Light objects in Morrowind (an actor hold an invisible torch in this case), but I am not sure if it can be implemented easily in OpenMW.

Capostrophic:
I believe this is the same code path? If it doesn't work still, it's probably not important anyway atm, as the complaint was about a regression with shields while lights were unaffected.

PR #2700, opened 2020-02-12
Move VisMask to components

akortunov:
In upstream master we declared enum for node masks in the MWRender namespace (with the separate enum for the editor), so when we need to use node masks inside of components code, we are forced to pass node masks as arguments and store them as class fields, what is weird. Also we use some magic numbers for node masks.

Summary of changes:
1. Declare nodemasks in components, so the whole OpenMW codebase can use them when needed.
2. Do not pass node masks as arguments, when it is not really needed. Allows to simplify API.
3. Define constants for default (0xffffffff) and empty (0x0) node masks to use them instead of magic numbers.
4. Drop the enum entry for fog since we do not use it anyway (fog has no its own scene nodes), hide a fog stub button in the editor.

Note: it is not very effective to use a separate node mask just for every editor's scene component, including controls, if there are many of them. Probably in this case it would be better to use the Mask_GUI with unique node names.


apps/opencs/view/render/object.cpp
```diff
@@ -140,7 +140,7 @@ void CSVRender::Object::update()
     if (light)
     {
         bool isExterior = false; // FIXME
-        SceneUtil::addLight(mBaseNode, light, Mask_ParticleSystem, Mask_Lighting, isExterior);
+        SceneUtil::addLight(mBaseNode, light, isExterior);
```
unelsson: Here you have also removed the Mask_ParticleSystem and Mask_Lighting. What's the purpose of that?

apps/opencs/view/render/pagedworldspacewidget.cpp
```diff
@@ -127,8 +126,8 @@ void CSVRender::PagedWorldspaceWidget::addVisibilitySelectorButtons (
     CSVWidget::SceneToolToggle2 *tool)
 {
     WorldspaceWidget::addVisibilitySelectorButtons (tool);
-    tool->addButton (Button_Terrain, Mask_Terrain, "Terrain");
-    tool->addButton (Button_Fog, Mask_Fog, "Fog", "", true);
+    tool->addButton (Button_Terrain, SceneUtil::Mask_Terrain, "Terrain");
+    //tool->addButton (Button_Fog, Mask_Fog, "Fog", "", true);
```
unelsson: While fog is not implemented, removing it is different than just moving mask to component level. I wonder what's the idea behind fog here in the first place, still, it's a different issue.

apps/opencs/view/render/unpagedworldspacewidget.cpp
```diff
@@ -304,8 +303,8 @@ void CSVRender::UnpagedWorldspaceWidget::addVisibilitySelectorButtons (
     CSVWidget::SceneToolToggle2 *tool)
 {
     WorldspaceWidget::addVisibilitySelectorButtons (tool);
-    tool->addButton (Button_Terrain, Mask_Terrain, "Terrain", "", true);
-    tool->addButton (Button_Fog, Mask_Fog, "Fog");
+    tool->addButton (Button_Terrain, SceneUtil::Mask_Terrain, "Terrain", "", true);
+    //tool->addButton (Button_Fog, Mask_Fog, "Fog");
```
unelsson: Same as above: Removing fog, is it needed?

apps/opencs/view/render/object.cpp
akortunov: Read the PR description. There are answers for all your questions.

apps/opencs/view/render/cell.cpp
```diff
@@ -134,7 +134,7 @@ void CSVRender::Cell::updateLand()
             else
             {
                 mTerrain.reset(new Terrain::TerrainGrid(mCellNode, mCellNode,
-                    mData.getResourceSystem().get(), mTerrainStorage, Mask_Terrain));
+                    mData.getResourceSystem().get(), mTerrainStorage));
```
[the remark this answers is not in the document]
akortunov: > Terrain should be in it's own mask in CS.

There are no reasons to keep separate masks for same things for OpenMW and the editor.

components/terrain/quadtreeworld.cpp
```diff
@@ -425,7 +426,7 @@ void QuadTreeWorld::enable(bool enabled)
     }
 
     if (mRootNode)
-        mRootNode->setNodeMask(enabled ? ~0 : 0);
+        mRootNode->setNodeMask(enabled ? ~SceneUtil::Mask_Disabled : SceneUtil::Mask_Disabled);
```
zinnschlag: Shouldn't ~SceneUtil::Mask_Disabled be Mask_Default instead? Since we are moving away from magic numbers anyway. And the naming is a bit weird. How about renaming Mask_Disabled to Mask_None and Mask_Default to Mask_All?

akortunov: > Shouldn't ~SceneUtil::Mask_Disabled be Mask_Default instead?

Well, if bitwise negation applied to 0 produces the 0xffffffff result on all supported platforms, it should be safe to use the Mask_Default here.

> How about renaming Mask_Disabled to Mask_None and Mask_Default to Mask_All?

I do not like the Mask_All here since in our current implementation we can not combine mask. So `Mask_All != Mask_Object|Mask_Terrain|Mask_Debug|... ` It just means that the node has the default node mask, without any special meaning.

AnyOldName3: I think I prefer `All` over` `Default`, but maybe `Any` is a better choice.

apps/openmw/mwrender/bulletdebugdraw.cpp
```diff
@@ -7,7 +7,8 @@
 
 #include <components/debug/debuglog.hpp>
 
-#include "vismask.hpp"
+#include <components/sceneutil/visitor.hpp>
```
unelsson: Just curious why do you add visitor.hpp here?

unelsson:
I went through all of the code, and compiled this PR. Tested OpenMW-CS behavior just in case. All works fine, good improvements in general. If there's a good reason for adding visitor.hpp to bulletdebugdraw.cpp (comment above), this PR is imho ready for a merge.

apps/openmw/mwrender/bulletdebugdraw.cpp
akortunov: An oversight, fixed.

unelsson: It should be components/sceneutil/vismask.hpp, right?

akortunov: GitHub still shows an old code to keep the context. There is an "Outdated" mark for reason.

unelsson: Oh, right, sorry. I thought it showed it changed to components/sceneutil/visitor.hpp, instead of original visitor.hpp...

The new one is indeed correct. So, merge-time?

PR #2707, opened 2020-02-20
Make PCSkipEquip and OnPCEquip behavior vanilla-like (bug #4141)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/issues/4141)

Replace our wacky hacks with vanilla wacky hacks based on the research of Hrnchamd and NullCascade.

Basically, this should be the case:
1. If PCSkipEquip is set to 1, OnPCEquip is set to 1 (for books as well) and equipment attempt is ignored. This is consistent with the behavior expected in bug 5182 report which I previously fixed.
2. If it's not and the items otherwise can be equipped (aren't broken or , OnPCEquip is set to 1 for everything except ingredients, books and repair tools. Ingredients are understandable because they are removed immediately, the repair hammer thing I'm not sure. Potions are removed as well, but they still have OnPCEquip set for some reason. I decided to copy all of this precisely.
3. For books, PCSkipEquip is set to 1, OnPCEquip stays 0. Don't ask why.

This fixes PCSkipEquip and OnPCEquip handling in scripts, so all things that rely on their state in the current frame should work fine, e.g. Morrowind Scripting for Dummies template which a bunch of TR scripts and many mods as well rely on and some other scripts that aren't based on this template, including one from the bug report and vanilla scripts, work exactly as in vanilla in my testing. Some book scripts still aren't fully compatible because there's some kind of an unclear issue with Activate understanding. OnActivate/Activate behavior is not obvious in vanilla scripting, as OnActivate can affect the script behavior merely by its presence anywhere in it. But their behavior is still much closer to vanilla.

As far as I can tell Morrowind doesn't run the script in the middle of the frame here like we did and doesn't save skipped equipment attempts in any way (it's unnecessary). So the behavior is actually more simple now.

### DIFF
--- a/apps/openmw/mwgui/inventorywindow.cpp
+++ b/apps/openmw/mwgui/inventorywindow.cpp
@@ -21,12 +21,10 @@
 #include "../mwbase/environment.hpp"
 #include "../mwbase/windowmanager.hpp"
 #include "../mwbase/mechanicsmanager.hpp"
-#include "../mwbase/scriptmanager.hpp"
 
 #include "../mwworld/inventorystore.hpp"
 #include "../mwworld/class.hpp"
 #include "../mwworld/actionequip.hpp"
-#include "../mwscript/interpretercontext.hpp"
 
 #include "../mwmechanics/actorutil.hpp"
 #include "../mwmechanics/creaturestats.hpp"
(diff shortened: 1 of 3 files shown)


apps/openmw/mwgui/inventorywindow.cpp
```diff
@@ -522,47 +530,32 @@ namespace MWGui
 
             if (!force)
             {
-                std::pair<int, std::string> canEquip = ptr.getClass().canBeEquipped(ptr, player);
+                const std::pair<int, std::string>& canEquip = ptr.getClass().canBeEquipped(ptr, player);
```
elsid: Why did you put a reference here? I can't find any difference in the generated assembly https://godbolt.org/z/icNb-q .

akortunov:
Did you check if the `bk_treasuryreport` works as intended?

Capostrophic:
Yes, it works exactly like in vanilla.

PR #2708, opened 2020-02-22
Limit frequency for navmesh updates

elsid:
This pr mitigates the issue with constantly transforming objects like sing in Seyda Neen described in [bug 4917](https://gitlab.com/OpenMW/openmw/issues/4917). The idea is to limit frequency of updates per navmesh tile. Added of removed objects from scene are not affected. Visible changes are:
* CPU usage. It drops from 141% to 105% on my machine for default settings with 60 fps limit and 250 ms navmesh update interval for the first 2000 frames since game is loaded in Seyda Neen location.  Save: [1.zip](https://github.com/OpenMW/openmw/files/4240150/1.zip).
* Navmesh change around opening/closing door.
  * master: https://streamable.com/8wf4w
  * pr: https://streamable.com/rq8ms

### DIFF
--- a/apps/openmw_test_suite/detournavigator/navigator.cpp
+++ b/apps/openmw_test_suite/detournavigator/navigator.cpp
@@ -73,6 +73,7 @@ namespace
             mSettings.mTrianglesPerChunk = 256;
             mSettings.mMaxPolys = 4096;
             mSettings.mMaxTilesNumber = 512;
+            mSettings.mMinUpdateInterval = std::chrono::milliseconds(50);
             mNavigator.reset(new NavigatorImpl(mSettings));
         }
     };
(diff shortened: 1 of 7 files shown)


AnyOldName3:
It might let us do more optimisation while keeping low latency and potentially improving realism if we make an object's bounds from the point of view of navmesh generation encompass everywhere it's been in the last few seconds. If something's doing the same thing in a loop, it's last-ten-second bounds aren't going to change, so navmesh regeneration can be skipped altogether, and people tend not to want to walk in the path of moving objects anyway.

elsid:
The idea is good, but there are some implementation problems. To have proper detection based only on time a some kind of poll based system is required. When object have a finite animation the last transformation may need to be handled with a delay. Simple example is to open a door and then close it right after it's opened. The whole animation will fit into 10 seconds or even in 3 seconds but last transformation may not be recognized as pushing object out of bounds for this interval. So it should be handled with a delay. But at that point object may not exist at the scene or may have another finite animation.

Another approach is to look at the last N transformations but this may cause false negatives (same problem as with only time based solution) if N is too big or false positives if N is too small.

elsid:
Here is a simple implementation using second approach a590d9510001cc6987e15e5d144158602e0afa3d and a video example how it can be abused: https://streamable.com/ugs1j

AnyOldName3:
I don't think you've quite understood what I was going for if the video was an attempt to implement what I was describing. The navmesh wouldn't go through the doorway until the door had stopped moving for a few seconds (because NPCs wouldn't want the madman playing with the door to hit them with the door).

What I'm suggesting is feeding the expensive part of navmesh generation a volume representing the union of all the places the door has been in the last few seconds. While you were messing with the door, it would just see it changing shape from a thin cuboid to a quarter of a cylinder, then staying that way until you left the door alone again, when it would shrink back to its original shape, either in the open or closed state depending on where you left it.

I *am* making some assumptions with this, though. The main one is that there's a way of computing an approximate convex hull of the last few hundred positions of an object and updating it to include new positions and exclude old ones that works out faster than navmesh generation (maybe something using signed distance fields that decay over time). It would probably need some kind of heuristic to detect things that were doing reciprocating movement, like signs, doors, and spinning blades like the Dwemer traps in Skyrim that will return to their original position versus things that stay still or go from one place to another and don't return.

elsid:
I think AnyOldName3 suggestion is orthogonal to the idea proposed by this pr. This pr suggests to limit navmesh update independent from possible ingame situations. And "generation of volume representing the union of all the places the door has been in the last few seconds" will make update rate depending on the scene. I thinks this approach is more about AI logic rather than about reducing CPU usage. For example if a single navmesh tile contains multiple doors placed at the same x, y but different z coordinate and they open, wait, close repeatedly and asynchronously. Update rate will depend on the number of doors. The situation seems to be unreal and ridiculous but it shows the difference between ideas. I excluded a590d9510001cc6987e15e5d144158602e0afa3d from the pr so it should not have the issue showed by https://streamable.com/ugs1j. I can try to implement AnyOldName3 idea but in the different pr. This one I suggest to merge as is if there are no issues.

psi29a:
I'll take it and we can address other issues in another PR.

PR #2712, opened 2020-02-25
Make search fields behave more consistently

Capostrophic:
1) There were keyboard focus issues due to how the spell window search field and the inventory search field are supposed to ignore Tab to avoid breaking Tab shortcut used as an alternative to Right Click pause shortcut. The used approach in spell window is problematic as it's a bit nuclear and can break mouse focus for the fields that have the user string, I switched it to a different solution where AcceptTab user string is set to true that works a bit differently - once you switch focus to the fields, Tab doesn't switch focus to anything in favor of other actions Tab is assigned to - but it should still be acceptable and even convenient somewhat. It's added to all search fields for consistency. IgnoreTabKey user string handling was dropped because it went unused.
2) Inventory search filter behaved oddly when the sort item model was replaced when the inventory window was reinitialized (after you load a save for example)
3) Companion window search filter wasn't cleaned up properly either.

I could have missed something.

### DIFF
--- a/apps/openmw/mwgui/spellwindow.cpp
+++ b/apps/openmw/mwgui/spellwindow.cpp
@@ -45,7 +45,7 @@ namespace MWGui
         getWidget(mEffectBox, "EffectsBox");
         getWidget(mFilterEdit, "FilterEdit");
 
-        mFilterEdit->setUserString("IgnoreTabKey", "y");
+        mFilterEdit->setUserString("AcceptTab", "true");
 
         mSpellView->eventSpellClicked += MyGUI::newDelegate(this, &SpellWindow::onModelIndexSelected);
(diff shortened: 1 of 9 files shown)


apps/openmw/mwgui/spellwindow.cpp
```diff
@@ -45,7 +45,7 @@ namespace MWGui
         getWidget(mEffectBox, "EffectsBox");
         getWidget(mFilterEdit, "FilterEdit");
 
-        mFilterEdit->setUserString("IgnoreTabKey", "y");
+        mFilterEdit->setUserString("AcceptTab", "true");
```
akortunov: Probably it would be better to set such strings in layout files rather than hardcoding them.

Capostrophic: Done.

PR #2713, opened 2020-02-27
Editor: Implement dropping objects with keypress in instancemode

unelsson:
G to drop objects to ground level, H to drop objects to next collision (e.g. on top of object below). If multiple objects are selected, their relative positions remain unchanged. Additionally implements two drop functions with different behavior that breaks the relative positions, dropping each item individually, these have no default assigned keyboard shortcut.

Notes:
- Implements osg::ref_ptr<osg::Group> CSVRender::Object::getRootNode() that is used to temporarily move object to SceneUtil::Mask_Disabled so that intersector doesn't collide with the dropped object itself.
- Implements also osg::ref_ptr<osg::Group> CSVRender::Object::getBaseNode() to calculate bounding boxes
- LineSegmentIntersector code is modified from similar code at CSVRender::WorldspaceWidget::mousePick
- The `LineSegmentIntersector` shoots straight down from the middle point of object, and often the lowest point is somewhere other than directly below the root node. This can result in a drop that looks wrong, but this should be good enough collision detection for OpenMW-CS.

Implemented:
- Bounding Box implementation
- Outdoor and indoor

### DIFF
--- a/apps/opencs/model/prefs/state.cpp
+++ b/apps/opencs/model/prefs/state.cpp
@@ -356,8 +356,10 @@ void CSMPrefs::State::declare()
         QKeySequence(Qt::ControlModifier | (int)Qt::MiddleButton));
     declareModifier ("scene-speed-modifier", "Speed Modifier", Qt::Key_Shift);
     declareShortcut ("scene-delete", "Delete Instance", QKeySequence(Qt::Key_Delete));
-    declareShortcut ("scene-instance-groundlevel", "Instance to Ground Level", QKeySequence(Qt::Key_G));
-    declareShortcut ("scene-instance-drop", "Drop Instance", QKeySequence(Qt::Key_H));
+    declareShortcut ("scene-instance-drop-terrain", "Drop to terrain level", QKeySequence(Qt::Key_G));
+    declareShortcut ("scene-instance-drop-collision", "Drop to collision", QKeySequence(Qt::Key_H));
+    declareShortcut ("scene-instance-drop-terrain-separately", "Drop to terrain level separately", QKeySequence());
+    declareShortcut ("scene-instance-drop-collision-separately", "Drop to collision separately", QKeySequence());
     declareShortcut ("scene-load-cam-cell", "Load Camera Cell", QKeySequence(Qt::KeypadModifier | Qt::Key_5));
     declareShortcut ("scene-load-cam-eastcell", "Load East Cell", QKeySequence(Qt::KeypadModifier | Qt::Key_6));
     declareShortcut ("scene-load-cam-northcell", "Load North Cell", QKeySequence(Qt::KeypadModifier | Qt::Key_8));
--- a/apps/opencs/model/prefs/state.cpp
+++ b/apps/opencs/model/prefs/state.cpp
@@ -356,6 +356,8 @@ void CSMPrefs::State::declare()
         QKeySequence(Qt::ControlModifier | (int)Qt::MiddleButton));
     declareModifier ("scene-speed-modifier", "Speed Modifier", Qt::Key_Shift);
     declareShortcut ("scene-delete", "Delete Instance", QKeySequence(Qt::Key_Delete));
+    declareShortcut ("scene-instance-groundlevel", "Instance to Ground Level", QKeySequence(Qt::Key_G));
+    declareShortcut ("scene-instance-drop", "Drop Instance", QKeySequence(Qt::Key_H));
     declareShortcut ("scene-load-cam-cell", "Load Camera Cell", QKeySequence(Qt::KeypadModifier | Qt::Key_5));
     declareShortcut ("scene-load-cam-eastcell", "Load East Cell", QKeySequence(Qt::KeypadModifier | Qt::Key_6));
     declareShortcut ("scene-load-cam-northcell", "Load North Cell", QKeySequence(Qt::KeypadModifier | Qt::Key_8));
--- a/apps/opencs/view/render/instancemode.cpp
+++ b/apps/opencs/view/render/instancemode.cpp
@@ -7,6 +7,7 @@
 
 #include "../../model/prefs/state.hpp"
 
+#include <osg/ComputeBoundsVisitor>
 #include <osg/Group>
 #include <osg/Vec3d>
 #include <osgUtil/LineSegmentIntersector>
@@ -704,7 +705,6 @@ void CSVRender::InstanceMode::dropInstance(DropMode dropMode, CSVRender::Object*
 
     osg::Vec3d start = point;
     osg::Vec3d end = point;
-    start.z() += 1.0f;
     end.z() = std::numeric_limits<float>::min();
 
     osg::ref_ptr<osgUtil::LineSegmentIntersector> intersector (new osgUtil::LineSegmentIntersector(
@@ -730,14 +730,12 @@ void CSVRender::InstanceMode::dropInstance(DropMode dropMode, CSVRender::Object*
     }
 }
 
-float CSVRender::InstanceMode::getDropHeight(DropMode dropMode, CSVRender::Object* object)
+float CSVRender::InstanceMode::getDropHeight(DropMode dropMode, CSVRender::Object* object, float objectHeight)
 {
     const osg::Vec3d& point = object->getPosition().asVec3();
-    osg::ref_ptr<osg::Group> objectNode = object->getRootNode();
 
     osg::Vec3d start = point;
     osg::Vec3d end = point;
-    start.z() += 1.0f;
     end.z() = std::numeric_limits<float>::min();
 
     osg::ref_ptr<osgUtil::LineSegmentIntersector> intersector (new osgUtil::LineSegmentIntersector(
@@ -754,9 +752,8 @@ float CSVRender::InstanceMode::getDropHeight(DropMode dropMode, CSVRender::Objec
          it != intersector->getIntersections().end(); ++it)
     {
         osgUtil::LineSegmentIntersector::Intersection intersection = *it;
-        ESM::Position position = object->getPosition();
-        float dropHeight = intersection.getWorldIntersectPoint().z();
-        return position.pos[2] - dropHeight;
+        float collisionLevel = intersection.getWorldIntersectPoint().z();
+        return point.z() - collisionLevel + objectHeight;
     }
 
     return 0.0f;
@@ -793,13 +790,25 @@ void CSVRender::InstanceMode::handleDropMethod(DropMode dropMode, QString comman
     CSMWorld::CommandMacro macro (undoStack, commandMsg);
 
     std::vector<osg::Node::NodeMask> oldMasks;
+    std::vector<float> objectHeights;
     for(osg::ref_ptr<TagBase> tag: selection)
     {
         if (CSVRender::ObjectTag *objectTag = dynamic_cast<CSVRender::ObjectTag *> (tag.get()))
         {
-            osg::ref_ptr<osg::Group> objectNode = objectTag->mObject->getRootNode();
-            oldMasks.emplace_back(objectNode->getNodeMask());
-            objectNode->setNodeMask(SceneUtil::Mask_Disabled);
+            osg::ref_ptr<osg::Group> objectNodeWithGUI = objectTag->mObject->getRootNode();
+            osg::ref_ptr<osg::Group> objectNodeWithoutGUI = objectTag->mObject->getBaseNode();
+
+            osg::ComputeBoundsVisitor computeBounds;
+            computeBounds.setTraversalMask(SceneUtil::Mask_EditorReference);
+            objectNodeWithoutGUI->accept(computeBounds);
+            osg::BoundingBox bounds = computeBounds.getBoundingBox();
+            float boundingBoxOffset = 0.0f;
+            if (bounds.valid()) boundingBoxOffset = bounds.zMin();
+
+            objectHeights.emplace_back(boundingBoxOffset);
+            oldMasks.emplace_back(objectNodeWithGUI->getNodeMask());
+
+            objectNodeWithGUI->setNodeMask(SceneUtil::Mask_Disabled);
         }
     }
 
@@ -811,11 +820,13 @@ void CSVRender::InstanceMode::handleDropMethod(DropMode dropMode, QString comman
             case Collision:
             {
                 float smallestDropHeight = std::numeric_limits<float>::max();
+                int counter = 0;
                     for(osg::ref_ptr<TagBase> tag: selection)
                         if (CSVRender::ObjectTag *objectTag = dynamic_cast<CSVRender::ObjectTag *> (tag.get()))
                         {
-                            float thisDrop = getDropHeight(dropMode, objectTag->mObject);
+                            float thisDrop = getDropHeight(dropMode, objectTag->mObject, objectHeights[counter]);
                             if (thisDrop < smallestDropHeight) smallestDropHeight = thisDrop;
+                            counter++;
                         }
                     for(osg::ref_ptr<TagBase> tag: selection)
                         if (CSVRender::ObjectTag *objectTag = dynamic_cast<CSVRender::ObjectTag *> (tag.get()))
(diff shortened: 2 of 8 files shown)


apps/opencs/model/prefs/state.cpp
```diff
@@ -356,6 +356,8 @@ void CSMPrefs::State::declare()
         QKeySequence(Qt::ControlModifier | (int)Qt::MiddleButton));
     declareModifier ("scene-speed-modifier", "Speed Modifier", Qt::Key_Shift);
     declareShortcut ("scene-delete", "Delete Instance", QKeySequence(Qt::Key_Delete));
+    declareShortcut ("scene-instance-groundlevel", "Instance to Ground Level", QKeySequence(Qt::Key_G));
+    declareShortcut ("scene-instance-drop", "Drop Instance", QKeySequence(Qt::Key_H));
```
zinnschlag: The wording is a bit strange. "Ground level" could be interpreted as the floor in a building, which is obviously not the case here. Let's call it what it is: Terrain. We are using this term consistently throughout the editor anyway.

I was about to comment that "Drop Instance" is inaccurate too. Should contain "selected" and it is rather instances than instance. But then I noticed that the same problem is present elsewhere. I guess we can go along with that for now and look at this issue again at a later time.

zinnschlag:
This is an interesting feature and I like the additional terrain option. However I see a usability issue. Assuming you have several instance selected and then drop them. They will all fall individually to the ground, breaking any existing formation. This is fine with unrelated collections of items. It is not fine when the instances form a single thing (in lack of better words).

A few examples:
* Back in the redemption days we had a smelter with several levers attached to it. These were separate objects that needed to be kept in the same position relative to each other.
* A table with some items on top of it.
* Walls with attached items (e.g. light sources)

Of course there are also situations where the current implementation would be desirable. So we really need four instead of two functions.

Note: If you consider these two new functions as a separate feature that is fine with me. Other than the issues I commented on inline this can go ahead IMO. In this case just make sure to file a separate issue for the two other functions.

unelsson:
Having each object drop to ground individually was a choice, but now that you mention it, there are situations where objects dropping as unit would be desirable. We can definitely implement the extra two functions, but we might eventually run out of keyboard shortcuts. I suggest implementing four functions, but in the end, leaving these two blank by default (empty keyboard shortcut). Default behavior should probably be dropping objects as one unit (non-breaking).

wareya:
You can drop groups individually by just dropping them one at a time, but there's no way to get group dropping unless it's handled as such, so I agree with making group dropping the default.

apps/opencs/view/render/instancemode.cpp
```diff
@@ -728,36 +735,106 @@ void CSVRender::InstanceMode::dropInstance(DropMode dropMode, CSVRender::Object*
...
 
-void CSVRender::InstanceMode::dropSelectedInstances(bool active)
+float CSVRender::InstanceMode::getDropHeight(DropMode dropMode, CSVRender::Object* object)
 {
-    std::vector<osg::ref_ptr<TagBase> > selection = getWorldspaceWidget().getSelection (SceneUtil::Mask_EditorReference);
-    if (selection.empty()) return;
+    const osg::Vec3d& point = object->getPosition().asVec3();
+    osg::ref_ptr<osg::Group> objectNode = object->getRootNode();
+    osg::Node::NodeMask oldMask = objectNode->getNodeMask();
+    objectNode->setNodeMask(SceneUtil::Mask_Disabled);
```
zinnschlag: This is a bit of a problem. What happens when there is an exception between setting the new mask and restoring the old mask? The same problem couldd be present in the previous commit. Guess I missed it during the review.

This needs to be made exception-safe. A try-catch with a re-throw would be an option. A better option would be something like the CommandMacro.

zinnschlag:
I am not terribly happy about leaving the shortcuts empty, because we currently have no other way to access the functions. But I always considered the current list of shortcuts a placeholder anyway. We need to review and re-assign shortcuts once we are feature complete (1.0) and have the full list of functions. Therefore I am okay with it for now.

In the long run we will need other ways to access these kinds of functions. I have some ideas about that (playing around with blender a bit lately, which is surprisingly inspiring). But that is a topic for another issue/PR.

apps/opencs/view/render/instancemode.cpp
unelsson: Ok, that's doable, but while testing this, I also found another issue: When dropping multiple objects, only one object at a time is set to Mask_Disabled, therefore dropped objects that are above each other tend to collide with each other. Correct behavior would be to set all dropped objects temporarily to Mask_Disabled, and everything in-between then done with try-catch.

unelsson:
Dropping is now implemented with one LineSegmentIntersector ray so that base node is set to the level of collision / terrain. It doesn't take in account the collision boxes of the objects, do you think that should be implemented before this feature is merged?

apps/opencs/view/render/instancemode.cpp
```diff
@@ -681,3 +697,164 @@ void CSVRender::InstanceMode::deleteSelectedInstances(bool active)
...
+        Log(Debug::Error) << "Error in dropping instance: " << e.what();
+    }
+
+    int counter = 0;
+    for(osg::ref_ptr<TagBase> tag: selection)
+    {
+        if (CSVRender::ObjectTag *objectTag = dynamic_cast<CSVRender::ObjectTag *> (tag.get()))
+        {
+            osg::ref_ptr<osg::Group> objectNode = objectTag->mObject->getRootNode();
+            objectNode->setNodeMask(oldMasks[counter]);
```
unelsson: This probably should be oldMasks.at(counter) instead of []

zinnschlag:
> Dropping is now implemented with one LineSegmentIntersector ray so that base node is set to the level of collision / terrain. It doesn't take in account the collision boxes of the objects, do you think that should be implemented before this feature is merged?

Yes, that would be desirable.

unelsson:
This feature seems to work quite often, but there are still some nasty bugs. Help and reviewing is cool.

unelsson:
Commands seem to be issued twice, at keypress and keyrelease.. This becomes an issue when you'd only want to drop once. I wonder if the correct solution is just to delete the implementation of QEvent::KeyPress here:
https://github.com/OpenMW/openmw/blob/bbca1f3d1d9456b8ace3f8e3598c1e192d33aea1/apps/opencs/model/prefs/shortcutsetting.cpp#L69

unelsson:
Alright, all known bugs fixed and bounding box functionality implemented. Review again please.

apps/opencs/view/render/instancemode.cpp
```diff
@@ -681,3 +698,180 @@ void CSVRender::InstanceMode::deleteSelectedInstances(bool active)
...
+                        objectTag->mObject->apply (macro);
+                        counter++;
+                    }
+            }
+                break;
+            default:
+                break;
+        }
+    }
+    catch (const std::exception& e)
```
zinnschlag: This is inconsistent with how we handle exceptions. Typically we let them propagate to the event loop, where they are handled and logged (not 100% sure of the logged part). We should keep this consistent (in case we want to change the error handling/logging behaviour at some point).

unelsson: I'm not quite sure what you mean. This is almost the same as https://github.com/OpenMW/openmw/blob/master/apps/opencs/view/render/scenewidget.cpp#L112 . There are other cases where a Qt signal is emitted at exception, is this what you meant?

zinnschlag: The code you pointed out is a special case, because it sits in a destructor. Destructors should not throw exceptions.

Our OpenMW-CS exception handling code is located in Application::notify (main.cpp). This is the central place where all exception handling happens. Code in event or signal handlers should not do their own exception handling other than local cleanup (if required). In this case the exception must be rethrown (via a simple throw; statement) so it can reach the main exception handling code.
Of course ideally code would just utilise RAII, in which case the cleanup would happen automatically without a try catch clause.

unelsson: I see, but.. I'm not really sure where the exception could occur anyway, and I've tried to use automatic clean-up whenever possible. Sure there are various OpenMW classes and osg smart pointers whose behavior might be hard to predict, but then, I don't see why exactly this code would require try-catch, and why other osg-using code wouldn't need that.

But I assume you have a reason to suggest this, so, instead of:

```
    {
        Log(Debug::Error) << "Error in dropping instance: " << e.what();
    }
```

should I do this?
```
    {
        throw std::runtime_error ("Error in dropping instance.");
    }
```

zinnschlag: It is reasonable to assume that in C++ (almost) anything can throw (almost) anything, unless you have a nothrow specifier, of course. Just because the code does not throw now doesn't mean it won't after a future change. That is a question of having maintainable and robust code.

Unfortunately the solution you posted won't work. First, you do not reach the cleanup code in case of an exception. And second, you throw the wrong exception. You should re-throw by using a

`
throw;
`

statement; that is, if you do not utilise RAII (see the CommandMacro class for an example).

There are two methods to implement this functionality:

1. Move the cleanup functionality into a function and call it both from within the catch statement and after the catch statement. This is the non-C++ way. Some other language have the finally keyword for this purpose, which is missing from C++, because that is non-idiomtic code in this language.

2. You write a little helper class (can be placed fully into the cpp file) that does the following:
- Do the initial change in the constructor
- Do the cleanup (reverting the change) in the destructor
- Delete the copy constructor and the assignment operator

This is the RAII method and it is the preferred approach in C++.

unelsson: I've been thinking of what variables should have their own class with custom destructors. I haven't found any better ways to do things yet.

Here are all the variables that are defined inside the default functionality of try-catch (except the CSVRender::InstanceMode::dropInstance, which is quite similar):

```in CSVRender::InstanceMode::handleDropMethod
float smallestDropHeight -> deleted when case-scope ends
int counter -> deleted when case-scope ends
osg::ref_ptr<TagBase> -> smart pointer (RAII?)
CSVRender::ObjectTag* -> pointer, deleted when scope ends (and dynamic_cast)
float thisDrop -> deleted when if-scope ends
osg::ref_ptr<TagBase> -> smart pointer (RAII?)
CSVRender::ObjectTag* -> pointer, deleted when scope ends (and dynamic_cast)
ESM::Position -> struct with floats, should be deleted when scope ends ( https://github.com/OpenMW/openmw/blob/dd67cdfa554c2bbd5aa073fa36a6bb736de4cca1/components/esm/defs.hpp#L38 )
objectTag->mObject->apply uses CommandMacro (RAII)

in CSVRender::InstanceMode::getDropHeight
const osg::Vec3d& point -> deleted when function-scope ends?
osg::Vec3d -> deleted when function-scope ends?
osg::ref_ptr<osgUtil::LineSegmentIntersector> -> smart pointer (RAII?)
osgUtil::IntersectionVisitor -> deleted when function-scope ends?
osgUtil::LineSegmentIntersector::Intersections::iterator it -> deleted when for-scope ends```

However, if something results in an exception, and the clean-up you are referring to is about the change of a mask, perhaps something like the code below would be the correct solution?

```
void CSVRender::InstanceMode::myCustomHelperClassChangesMask::myCustomHelperClassChangesMask()
{
    CHANGE MASK HERE
}

CSVRender::InstanceMode::myCustomHelperClassChangesMask::~myCustomHelperClassChangesMask()
{
    CHANGE MASK BACK HERE
}

void CSVRender::InstanceMode::handleDropMethod(DropMode dropMode, QString commandMsg)
{
    myCustomHelperClassChangesMask()
    try
    {
         DROPPING FUNCTIONALITY HERE
    }
    {
        throw;
    }
} // When this scope ends, the destructor of myCustomHelperClassChangesMask is called```

unelsson: edit note: Removed the `void` etc. crap from previous meta-code example...

elsid: > I've been thinking of what variables should have their own class with custom destructors.

It's possible to customize type by a callback using templates, `std::function` and lambdas:
```c++
template <class OnDestruct>
struct MyType {
    OnDestruct onDestruct;
    ~MyType() { onDestruct(this); }
};

template <class OnDestruct>
MyType<OnDestruct> makeMyType(OnDestruct&& onDestruct) {
    return MyType<std::decay_t<OnDestruct>> {std::forward<OnDestruct>(onDestruct)};
}

auto foo = makeMyType([] (auto self) { /*...*/ });
auto bar = makeMyType([] (auto self) { /*...*/ });
```

Also there is such generic type called `std::unique_ptr`.
```
struct MyType {};

MyType foo;
auto onFooDestruct = [] (auto self) {  };
std::unique_ptr<MyType, decltype(onFooDestruct)> fooWithCustomDestructor(&foo, onFooDestruct);
```

zinnschlag: @unelsson The code you posted looks nearly correct. You can delete the try-catch-clause here, because if the only thing your catch does is to re-throw, then it becomes a no-op.

@elsid Interesting approach, but probably a bit more than we need for this rather simple problem.

unelsson: Thanks for your great reviews. I used the constructor-destructor data handling method (made a custom helper class) and rebased this PR to the latest master.

zinnschlag: Functionally this looks good. There are a few stylistic issues. I will comment about them inline. Once these issues have been addressed this PR should be ready for a merge.

apps/opencs/view/render/instancemode.cpp
```diff
@@ -681,3 +697,183 @@ void CSVRender::InstanceMode::deleteSelectedInstances(bool active)
 
     getWorldspaceWidget().clearSelection (SceneUtil::Mask_EditorReference);
 }
+
+void CSVRender::InstanceMode::dropInstance(DropMode dropMode, CSVRender::Object* object, float objectHeight)
+{
+    const osg::Vec3d& point = object->getPosition().asVec3();
```
zinnschlag: This shouldn't be a reference. Strictly speaking the reference is okay, because it will extend the lifetime of the temporary. But nothing is gained by using a reference and it adds unnecessary complexity to the code.

apps/opencs/view/render/instancemode.cpp
```diff
@@ -681,3 +697,183 @@ void CSVRender::InstanceMode::deleteSelectedInstances(bool active)
...
+    start.z() += objectHeight;
+    osg::Vec3d end = point;
+    end.z() = std::numeric_limits<float>::lowest();
+
+    osg::ref_ptr<osgUtil::LineSegmentIntersector> intersector (new osgUtil::LineSegmentIntersector(
+        osgUtil::Intersector::MODEL, start, end) );
+    intersector->setIntersectionLimit(osgUtil::LineSegmentIntersector::NO_LIMIT);
+    osgUtil::IntersectionVisitor visitor(intersector);
+
+    if (dropMode == Terrain_sep) visitor.setTraversalMask(SceneUtil::Mask_Terrain);
```
zinnschlag: We don't really do this kind of one-liners for if-clauses in OpenMW. The body of the if statement belongs into the next line. I have seen this a bunch of times.

apps/opencs/view/render/instancemode.cpp
```diff
@@ -681,3 +697,183 @@ void CSVRender::InstanceMode::deleteSelectedInstances(bool active)
...
+        position.pos[2] = intersection.getWorldIntersectPoint().z() + objectHeight;
+        object->setPosition(position.pos);
+
+        return;
+    }
+}
+
+float CSVRender::InstanceMode::getDropHeight(DropMode dropMode, CSVRender::Object* object, float objectHeight)
+{
+    const osg::Vec3d& point = object->getPosition().asVec3();
```
zinnschlag: Same as above.

apps/opencs/view/render/instancemode.cpp
```diff
@@ -681,3 +697,183 @@ void CSVRender::InstanceMode::deleteSelectedInstances(bool active)
...
+                if (CSVRender::ObjectTag *objectTag = dynamic_cast<CSVRender::ObjectTag *> (tag.get()))
+                {
+                    dropInstance(dropMode, objectTag->mObject, dropObjectDataHandler.mObjectHeights[counter]);
+                    objectTag->mObject->apply (macro);
+                    counter++;
+                }
+        }
+            break;
+
+        default:
```
zinnschlag: This default does not serve a purpose. It can be removed.

apps/opencs/view/render/instancemode.hpp
```diff
@@ -29,13 +33,22 @@ namespace CSVRender
                 DragMode_Scale
             };
 
+            enum DropMode
+            {
+                Collision,
+                Terrain,
+                Collision_sep,
+                Terrain_sep
+            };
```
zinnschlag: This doesn't quite match our naming scheme. CollisionSep instead of Collision_sep. Or Collision_Sep, if you absolutely must.

Also, we normally prefix the individual enum names with the name of the enum itself (DropMode_Collision). Alternatively you may use a class enum, which would then require explicitly stating the enum on use (DropMode::Collision).

PR #2715, opened 2020-03-04
Remove custom data for disposed actors

akortunov:
Based on [task #5194](https://gitlab.com/OpenMW/openmw/issues/5194).

Summary of changes:
1. Do not write a custom data (inventory, current stats values, etc) to the savegame, if an actor is deleted (count = 0).
2. Adjust "check if item's owner is dead" logic to take in account the fact that we can not use CreatureStats for disposed actors.

According to my testing, my endgame save has about 400 records for dead actors (many records are for dead horkers from the Bloodmoon quest line).
A cleanup makes my endgame save about 550 KB smaller (about 2%), and it loads about 2-3% faster.

Notes:
1. This patch does not require a savegame format change, but older 0.47 builds rely on fact that we store data forever, so they can not handle the "check if item's owner is dead" logic well for cleaned saves.
2. If there is an additional obscure logic, which handles disposed actors (basically, accesses CreatureStats or ContainerStore), it should be updated as well.
As for script commands which check target's stats or inventory, they should not work for disposed actors as well (in the Morrowind they return default values, as if target was resurrected).

### DIFF
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -813,6 +813,12 @@ namespace MWClass
             return;
         }
 
+        if (ptr.getRefData().getCount() <= 0)
+        {
+            state.mHasCustomState = false;
+            return;
+        }
+
         const CreatureCustomData& customData = ptr.getRefData().getCustomData()->asCreatureCustomData();
 
         customData.mContainerStore->writeState (state2.mInventory);
--- a/apps/openmw/mwclass/npc.cpp
+++ b/apps/openmw/mwclass/npc.cpp
@@ -1331,6 +1331,12 @@ namespace MWClass
             return;
         }
 
+        if (ptr.getRefData().getCount() <= 0)
+        {
+            state.mHasCustomState = false;
+            return;
+        }
+
         const NpcCustomData& customData = ptr.getRefData().getCustomData()->asNpcCustomData();
 
         customData.mInventoryStore.writeState (state2.mInventory);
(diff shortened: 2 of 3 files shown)


psi29a:
For #2 what happens to already disposed corpses from existing saves? Will they just forever stay in those older saves? Will new saves remove them?

akortunov:
> For #2 what happens to already disposed corpses from existing saves?

Without save-game checks - nothing happens for them, they stay forever as in upstream master. With current PR state I just do not write data for removed actors to savegame.

Capostrophic:
Does the functionality where disposed actors present in the cell are counted as dead for the sake of not considering an item stolen still work?

akortunov:
> Does the functionality where disposed actors present in the cell are counted as dead for the sake of not considering an item stolen still work?

Can you tell where this functionality is located?

Capostrophic:
In itemTaken(). I mean that disposed actors should still be found and still be "dead".

akortunov:
> I mean that disposed actors should still be found and still be "dead".

Judging by code, this mechanics uses flags from CreatureStats to determine if the target is dead.
I am pretty sure that Morrowind does not use such approach since it reset victim's stats when player disposes the corpse.
I suspect that Morrowind uses GetDeadCount check here.

akortunov:
> Does the functionality where disposed actors present in the cell are counted as dead for the sake of not considering an item stolen still work?

Should work now.

akortunov:
Any suggestins here?

psi29a:
None, merged.

PR #2717, opened 2020-03-07
Use clone() instead of subimage copying for global map texture

akortunov:
Should fix [bug #5308](https://gitlab.com/OpenMW/openmw/issues/5308).

We handle global map in the master branch in this way:
1. We create map texture during game start, when we init UI.
2. We use this texture for new game, and bake it to the savegame in the PNG format.
3. During savegame loading we check if world dimensions and map resolution from the savegame are the same as for existing one. If they are, we **copy the whole overlay image data from texture, which we read from savegame, to the existing overlay image**.

EDIT: it should be enough just to use an assignment here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,6 +204,7 @@
     Bug #5269: Editor: Cell lighting in resaved cleaned content files is corrupted
     Bug #5278: Console command Show doesn't fall back to global variable after local var not found
     Bug #5300: NPCs don't switch from torch to shield when starting combat
+    Bug #5308: World map copying makes save loading much slower
     Feature #1774: Handle AvoidNode
     Feature #2229: Improve pathfinding AI
     Feature #3025: Analogue gamepad movement controls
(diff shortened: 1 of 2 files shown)


AnyOldName3:
Won't `DEEP_COPY_ALL` also do exactly the same copying of the image data as we're already doing?

akortunov:
> Won't DEEP_COPY_ALL also do exactly the same copying of the image data as we're already doing?

`copySubImage` have additional functionality, such as image rescaling.

AnyOldName3:
The closest thing to what we're doing now that avoids the `gluScaleImage` call would be to call `data()`  on the images to get the data pointers, `memcpy` the new image across, and then call `mOverlayImage.dirty()`.

I've had a look, though, and don't see any reason why we need a copy. We can just do `mDisplayImage = image` and nothing should break.

AnyOldName3:
Also, I saw a `&thing[0]` a few lines above this all, which should probably be changed to a `.data()` while we're in the area.

akortunov:
> I've had a look, though, and don't see any reason why we need a copy. We can just do mDisplayImage = image and nothing should break.

If you say so...

PR #2721, opened 2020-03-14
Support for radial fog

akortunov:
Implements [feature #4708](https://gitlab.com/OpenMW/openmw/issues/4708).
Allows to have a radial fog when shaders are enabled, and should not cause noticable FPS drops.

~~Note that it is not a [feature #4708](https://gitlab.com/OpenMW/openmw/issues/4708) - this PR does not have a radial clipping, which is a quite hard to implement without performance loss. Also the clipping itself may be not really important for us, if objects which it is supposed to hide will be covered by fog anyway.~~

Also according to AnyOldName3, shadows still expect linear depth to work properly, so define both in shaders when needed.

### DIFF
--- a/docs/source/reference/modding/settings/shaders.rst
+++ b/docs/source/reference/modding/settings/shaders.rst
@@ -135,3 +135,14 @@ apply lighting to environment maps
 Normally environment map reflections aren't affected by lighting, which makes environment-mapped (and thus bump-mapped objects) glow in the dark.
 Morrowind Code Patch includes an option to remedy that by doing environment-mapping before applying lighting, this is the equivalent of that option.
 Affected objects will use shaders.
+
+radial fog
+----------
+
+:Type:		boolean
+:Range:		True/False
+:Default:	False
+
+By default, the fog becomes thicker proportionally to your distance from the clipping plane set at the clipping distance, which causes distortion at the edges of the screen.
+This setting makes the fog use the actual eye point distance (or so called Euclidean distance) to calculate the fog, which makes the fog look less artificial, especially if you have a wide FOV.
+Note that the rendering will act as if you have 'force shaders' option enabled with this on, which means that shaders will be used to render all objects and the terrain.
--- a/files/settings-default.cfg
+++ b/files/settings-default.cfg
@@ -350,6 +350,10 @@ terrain specular map pattern = _diffusespec
 # Affected objects use shaders.
 apply lighting to environment maps = false
 
+# Determine fog intensity based on the distance from the eye point.
+# This makes fogging independent from the viewing angle. Shaders will be used to render all objects.
+radial fog = false
+
 [Input]
 
 # Capture control of the cursor prevent movement outside the window.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -225,6 +225,7 @@
     Feature #4544: Actors movement deceleration
     Feature #4673: Weapon sheathing
     Feature #4675: Support for NiRollController
+    Feature #4708: Radial fog support
     Feature #4730: Native animated containers support
     Feature #4784: Launcher: Duplicate Content Lists
     Feature #4812: Support NiSwitchNode
--- a/apps/opencs/model/world/data.cpp
+++ b/apps/opencs/model/world/data.cpp
@@ -82,6 +82,7 @@ CSMWorld::Data::Data (ToUTF8::FromType encoding, bool fsStrict, const Files::Pat
     defines["forcePPL"] = "0"; // Don't force per-pixel lighting
     defines["clamp"] = "1"; // Clamp lighting
     defines["preLightEnv"] = "0"; // Apply environment maps after lighting like Morrowind
+    defines["radialFog"] = "0";
     for (const auto& define : shadowDefines)
         defines[define.first] = define.second;
     mResourceSystem->getSceneManager()->getShaderManager().setGlobalDefines(defines);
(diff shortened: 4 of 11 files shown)


Capostrophic:
I'd still prefer an all-or-nothing setting, because even when "force shaders" is set to false objects are not going to be rendered purely in FFP, and then we see a glaring inconsistency in the look of FFP-rendered objects and the objects that are rendered with shaders - like normal-mapped objects that look conspiciously more pale on the edge of the screen.

And the shader pipeline is supposed to look exactly like FFP by default at the moment. It's just to lessen the headache when we're wondering why openmw looks different from vanilla.

> Note that it is not a feature #4708 

I'm pretty sure we decided not to implement radial clipping and that that issue should be repurposed to refer radial fog only.

akortunov:
> I'd still prefer an all-or-nothing setting

The only thing we can do here is to enable shaders automatically, when the setting is enabled, as we do for bump maps, shadows and some other features, but it is not really a "all-or-nothing setting".
> And the shader pipeline is supposed to look exactly like FFP by default at the moment.

No, we have a plenty of features, which do not work in FFP mode.

Capostrophic:
> No, we have a plenty of features, which do not work in FFP mode.

Which doesn't mean we shouldn't strive for a vanilla-compatible look.

>

AnyOldName3:
We have plenty of features that people agree make things prettier than vanilla, but they're all off by default. The defaults should be possible to achieve in vanilla.

akortunov:
OK, I added a setting.

wareya:
Sorry for butting in, but where can I read the discussion about giving up on radial clipping? I'm interested in the rationale.

akortunov:
On a Discord. An example:

> We've mostly gone off the idea. Every way of doing it is either hugely complicated or makes things much slower for a lot of people. There are good reasons why we literally couldn't find a case of any other game ever doing radial clipping.

psi29a:
What?

> DetourNavigatorNavigatorTest.update_remove_and_update_then_find_path_should_return_path

files/settings-default.cfg
```diff
@@ -350,6 +350,10 @@ terrain specular map pattern = _diffusespec
 # Affected objects use shaders.
 apply lighting to environment maps = false
 
+# Use Euclidean distance to determince fog intensity instead of plane.
+# Affected objects use shaders.
```
Capostrophic: Determine fog intensity based on the distance from the eye point. <newline>
This makes fogging independent from the viewing angle. Shaders will be used to render all objects.

docs/source/reference/modding/settings/shaders.rst
```diff
@@ -135,3 +136,13 @@ apply lighting to environment maps
...
+
+radial fog
+----------
+
+:Type:		boolean
+:Range:		True/False
+:Default:	False
+
+Use Euclidean distance to calculate fog level instead of plane to prevent distortion at the edges of the screen. The effect this prevents is more noticeable at high Field of View settings.
+Affected objects will use shaders.
```
Capostrophic: By default, the fog becomes thicker proportionally to your distance from the clipping plane set at the clipping distance, which causes distortion at the edges of the screen. This setting makes the fog use the actual eye point distance (or so called Euclidean distance) to calculate the fog, which makes the fog look less artificial especially if you have a wide FOV. <newline>
Note that the rendering will act as if you have 'force shaders' option enabled with this on, which means that shaders will be used to render all objects and the terrain.

akortunov:
> DetourNavigatorNavigatorTest.update_remove_and_update_then_find_path_should_return_path

Looks similar to [this](https://gitlab.com/OpenMW/openmw/-/merge_requests/163) issue, as for me.

docs/source/reference/modding/settings/shaders.rst
```diff
@@ -10,6 +10,7 @@ force shaders
 
 Force rendering with shaders. By default, only bump-mapped objects will use shaders.
 Enabling this option may cause slightly different visuals if the "clamp lighting" option is set to false.
+Also fog uses a Euclidean distance from camera instead of depth when this option is enabled.
```
Capostrophic: This line should be removed.

AnyOldName3:
One minor objection to this (that should be fixable in a few seconds and is just caused by copying something I must have missed with a previous PR): OpenMW uses British English for variable name spelling and documentation, but there are lots of `*ise` words spelled as `*ize` in this PR.

akortunov:
> OpenMW uses British English for variable name spelling and documentation

IIRC, OpenMW uses both British and American spellings (we use an "optimizer" word, for example). Also such requirements should be somewhere in [contributing guides](https://wiki.openmw.org/index.php?title=Naming_Conventions).

By the way, I used an American spelling for three years in OpenMW, and no one complained about it.

psi29a:
Not about complaining, it is about being consistent. Even as an American, but I would prefer that we stick to British names for consistency and international sake.

The contributing guides: https://wiki.openmw.org/index.php?title=Naming_Conventions
already imply British spelling because of how the document is written, take "neighbourhood" for example with the extra u that isn't found in the American spelling. It probably couldn't hurt to be explicit about it.

akortunov:
> The contributing guides already imply British spelling because of how the document is written, take "neighbourhood" for example

Keep in mind that many contributors are not English-speakers and do not know about such nuances.
If you really want to use this rule:
1. Explicitely add it to the contributing guide.
2. Assign a person (an English-speaker), who is supposed to review the code and find such stylistic errors.

As I said, for now we use both spellings in the OpenMW codebase, so maybe it is worth to do a refactoring as well - without it the rule does not really make sense.

AnyOldName3:
The last time this came up, I was able to link to something in our docs saying we used British English, so either we have it *somewhere*, or someone's removed it without telling everyone. Maybe we should be looking in the wiki edit history for harbours full of tea.

PR #2722, opened 2020-03-16
Rotate camera via gyroscope on mobile devices

akortunov:
Implements [feature #5311](https://gitlab.com/OpenMW/openmw/-/issues/5311).
Uses a [new SDL API](https://github.com/SDL-mirror/SDL/blob/master/test/testsensor.c) for reading data from sensors (including gyroscopes).
Gyroscope should return its current angular speed for three axis (in rad/sec). Requires SDL 2.0.9 or newer.

Settings:
```
[Input]
enable gyroscope = true
gyro horizontal axis = -x
gyro vertical axis = y
gyro input threshold = 0.01
gyro horizontal sensitivity = 1.0
gyro vertical sensitivity = 1.0
```

So far SDL supports only built-in gyroscopes for mobile devices, so this feature can be useful only for ports to mobile devices (to Android-based and iOS-based ones, for example).
There are efforts to implement an API for gamepad gyroscopes, but it is unclear if these efforts will be merged to upstream SDL.

A [testing video](https://www.youtube.com/watch?v=WVEykPdcVio).

The only known issue left - default gyroscope update interval is too high, so SDL needs to be patched for now to work properly. A related [bugreport](https://bugzilla.libsdl.org/show_bug.cgi?id=5052). Related patch will be present only in upcoming SDL release (2.0.13, I believe). So Android port's developers will need to apply this patch while they use earlier versions of SDL.


akortunov:
@xyzz, can you provide an APK file for Android with this patch, please?
There are users on 4pda.ru, which are interested in testing.

xyzz:
Since the patch requires SDL 2.0.9 it'd require a considerable amount of work updating the Android port from the current 2.0.8. Unfortunately SDL makes it their mission to break Android API every minor release (e.g. see https://github.com/xyzz/openmw-android/commit/0e90482454cc5bb66c43fdc21b3fa885cadf2312 which was update from 2.0.5 to 2.0.8). I'll try to get to it this weekend but cannot promise anything.

psi29a:
From a debian/ubuntu packagers perspective, sdl2 2.0.9 shouldn't be a problem.
https://packages.debian.org/buster/libsdl2-dev

Before this PR becomes mergable, we'll need to bump packages for Windows, Mac and Linux.

xyzz:
So the big thing it's missing is `SDL_INIT_SENSOR` in https://github.com/OpenMW/openmw/blob/ed06fc2aa8dbac262359d5aa55f7891dbe9f9523/apps/openmw/engine.cpp#L238 After adding it, the change does move the camera, but it doesn't work very well.

[omw_debug_0.46.0-32.zip](https://github.com/OpenMW/openmw/files/4364520/omw_debug_0.46.0-32.zip) (apk inside the zip because github wouldn't let me upload it otherwise)

Here's the source tree https://github.com/xyzz/openmw-android/tree/sdl-2.0.12, unfortunately update to 2.0.12 broke a bunch of other stuff, so I'll have to figure it out, but in the meantime feel free to compile and test the feature.

akortunov:
@xyzz, can you re-build an APK, please? I uploaded some bugfixes and added some debug measures.
From what I can tell, the main issue left is that `SDL_SENSORUPDATE` events seem to be quite rare, so camera transitions are abrupt.

akortunov:
A new approach - do not reset rotation speed after every UI update since `SDL_SENSORUPDATE` events a quite rare.

xyzz:
Sure, there you go: 
[omw_debug_0.46.0-32.zip](https://github.com/OpenMW/openmw/files/4365551/omw_debug_0.46.0-32.zip)

akortunov:
Judging by testing results, gyroscope update interval is about 100-200 ms (something like the SENSOR_DELAY_NORMAL), but I see such thing in the port's code:

```
public void enableSensor(int sensortype, boolean enabled) {
        // TODO: This uses getDefaultSensor - what if we have >1 accels?
        if (enabled) {
            mSensorManager.registerListener(this,
                            mSensorManager.getDefaultSensor(sensortype),
                            SensorManager.SENSOR_DELAY_GAME, null);
        } else {
            mSensorManager.unregisterListener(this,
                            mSensorManager.getDefaultSensor(sensortype));
        }
    }
```

SENSOR_DELAY_GAME is something like the 40ms interval, but it seems to do not work.
Is there a way to setup a sensor update interval on Android?

xyzz:
This code is only used for accelerometer, the gyroscope implementation in SDL is completely native and does not use any Java code.

akortunov:
I filled a [bugreport](https://bugzilla.libsdl.org/show_bug.cgi?id=5052) to the SDL team. Let see if we get a response...

xyzz:
FWIW, there is a FIXME comment present in SDL2 source here: https://github.com/spurious/SDL-mirror/blob/863c4bd26b51892864c6042ad2db474e11b08fed/src/sensor/android/SDL_androidsensor.c#L152

After adding 

```
    int delay = ASensor_getMinDelay(hwdata->asensor);
    if (delay) {
        ASensorEventQueue_setEventRate(hwdata->eventqueue, hwdata->asensor, delay);
    }
```

The gyroscope does appear much more responsive (as well as spamming the log a lot), though perhaps we should limit it to 60-120 Hz.

[omw_debug_0.46.0-32.zip](https://github.com/OpenMW/openmw/files/4365854/omw_debug_0.46.0-32.zip)

There's another problem, which is that we need to invert the gyroscope when the device is rotated. SDL2 doesn't do anything about it at the moment.

akortunov:
Thanks. Did you test the input yourself? Is it actually usable now?

xyzz:
It felt pretty good to me - though I haven't tried any other games which implement the feature and therefore not familiar with what sensitivity players expect. It definitely should be much lower than the current 1.0 - perhaps in the 0.01-0.05 range.

akortunov:
Summary of changes:
1. Disabled events spam.
2. Added a docs draft for new settings.

> There's another problem, which is that we need to invert the gyroscope when the device is rotated. SDL2 doesn't do anything about it at the moment.

For now I use only manual axis mappings, but did not implement an in-game way to change settings yet.

> though I haven't tried any other games which implement the feature and therefore not familiar with what sensitivity players expect.

I got a feedback from several users familiar with applications, which use gyroscopes (including Doom's port). According to that feedback, gyroscopes work pretty well in your latest APK.

psi29a:
I'm busy getting travis to use bionic, I should be able to backport sdl2 2.0.9 (or greater) to that (in our PPA)
Someone is going to have to bump for android. Hopefully homebrew and chocolately can help us on macos and windows...

psi29a:
@nikolaykasyanov Could you help with the macos deps for sdl2?

xyzz:
> Do you refer a portrait/landscape mode?
For now I use only manual axis mappings, but I did not implement an in-game way to change settings yet.

Landscape mode only. However, the user might still flip the phone 180°, so we should take care to automatically flip x/y when the phone is rotated. In fact, SDL2 already does it for accelerometer (https://github.com/spurious/SDL-mirror/blob/863c4bd26b51892864c6042ad2db474e11b08fed/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L2050) but not for gyroscope.

psi29a:
Right...  windows and ubuntu builds have SDL2 >= 2.0.9
Just waiting on macos

akortunov:
> the user might still flip the phone 180°, so we should take care to automatically flip x/y when the phone is rotated

In theory, in 2.0.9 an API for screen orientation was added: 
```
SDL_GetDisplayOrientation() to get the current display orientation
along with SDL_DISPLAYEVENT for changes to it.
```

So an attempt to get a screen orientation would be something like:

```
int screen = Settings::Manager::getInt("screen", "Video");
SDL_DisplayOrientation currentOrientation = SDL_GetDisplayOrientation (screen);
```

The main issue here would be to re-map axes properly for different screen.
We should do it during game start and after every SDL_DISPLAYEVENT.

I found [this](https://www.mathworks.com/help/supportpkg/android/ref/gyroscope.html) doc.
If it is realiable, it should be possible to re-map assigned axes on the fly.

akortunov:
@xyzz, I added an experimental axes re-mapping, using the SDL_DISPLAYEVENT.
In theory, OpenMW now should detect on the fly when device orientation is changed.
When orientation is changed, OpenMW should correct axes directions and print two messages to openmw.log:
1. The "Display orientation changed."
2. The "Setup gyroscope axes..."

My maths can be wrong, so feel free to change the InputManager::correctGyroscopeAxes(), if needed.
A new APK would be welcome (especially if OpenMW on Android supports a device orientation change).

xyzz:
FWIW I've added github actions integration to the openmw-android repo so you should be able to fork it, change the openmw commit/hash and have it produce working APKs.

akortunov:
As I understood, I should clone a repo and create a pull request to your repo?

xyzz:
> As I understood, I should clone a repo and create a pull request to your repo?

Looks like you've got it now, but yeah it shouldn't be required.

> And a separate question - is the sdl-2.0.12 branch feature-complete?

No, a ton of things are still broken: pause/stop sounds on minimize, touch controls somehow get randomly offset, and I haven't even tried any more advanced features we have like external gamepad support or custom resolution... it's why I prefer never touching SDL2, sigh.

psi29a:
Managed to backport sdl2 2.0.9 to trusty, took a bit, but we're good to go there.

akortunov:
Looks like SDL devs [decided](https://hg.libsdl.org/SDL/rev/06adf174772e) to poll sensors about 60 times per second on Android for now.

> No, a ton of things are still broken: pause/stop sounds on minimize, touch controls somehow get randomly offset, and I haven't even tried any more advanced features we have like external gamepad support or custom resolution... it's why I prefer never touching SDL2, sigh.

All right, is there a branch, where openmw-android is ported to from SDL 2.0.8 to SDL 2.0.9, and it contains a patch to reduce gyroscope update interval?
Or did you try to port it directly to 2.0.12?

nikolaykasyanov:
@psi29 I’ll try to have a look in coming days.

akortunov:
As for not working SDL_DISPLAYEVENT, I found that the
```enableSensor(Sensor.TYPE_ACCELEROMETER, true);```
is called and the SDL thread is created only after aplication's resume, not after application's start (and there is a related FIXME there).

**Comment by akortunov** [comment]
So there is no point to read a screen orientation every frame - SDL caches the value anyway after display initialization and updates it when the SDL_DISPLAYEVENT comes. Since this event does not work for us for some reason, there is no way to get a current display orientation ATM.

**Comment by akortunov** [comment]
@xyzz, can you tell why you removed such calls:
```enableSensor(Sensor.TYPE_ACCELEROMETER, true);```
from the SDLActivity.java?
Such change breaks display-related API in SDL. I restored the call in the handleResume(), and display orientation change started to [work](https://www.youtube.com/watch?v=WVEykPdcVio).

**Comment by akortunov** [comment]
The feature itself is complete and ready for review.

**Comment by xyzz** [comment]
IIRC, it was so that it didn't spam my profiling logs, as sensor wasn't used but the events arrived all the time. At that time it also didn't send orientation events (this is a separate feature introduced in https://github.com/spurious/SDL-mirror/commit/4ab7b3d88bc97d2db629ffc59f993a9ff4487a84).

I've tried enabling it, and somehow this enabled portrait mode, which is not supposed to happen as it's locked to `sensorLandscape` in AndroidManifest.xml - maybe the [`getRotation`](https://developer.android.com/reference/android/view/Display#getRotation()) API SDL ignores the manifest, I'm not sure...

**Comment by xyzz** [comment]
> All right, is there a branch, where openmw-android is ported from SDL 2.0.8 to SDL 2.0.9, and it contains a patch to reduce gyroscope update interval? Can I just drop the 2231e97 to get working 2.0.9?

I haven't tried 2.0.9 yet, I should see if it's working better, then again, it doesn't contain some of the fixes such as https://github.com/spurious/SDL-mirror/commit/382e3d8ad28874a4ea8b6010d2cb9408efcd647b

**Comment by akortunov** [comment]
> I've tried enabling it, and somehow this enabled portrait mode

Well, disabling it does not disable a portrait mode.

**Comment by xyzz** [comment]
> Well, disabling it does not disable a portrait mode.

Weird, it should - if the portrait mode is still working, then I guess it should be added to the ever growing pile of issues.

> BTW, you enabled a portrait mode for the main activity, which SDL seem to use.

The main activity is used for the settings menu, SDL's using the GameActivity. At least that's the case on the latest stable/nightly builds with 2.0.8.

**Comment by akortunov** [comment]
According to testing results, SDL_HINT_ORIENTATIONS does the trick:
```
 SDL_SetHint(SDL_HINT_ORIENTATIONS, "LandscapeLeft LandscapeRight");
```

**Comment by nikolaykasyanov** [comment]
Rebuilding macOS dependencies, will keep you posted.

Btw, would this also help to support gyroaiming for supported controllers like Nintendo Switch ones?

**Comment by akortunov** [comment]
> Btw, would this also help to support gyroaiming for supported controllers like Nintendo Switch ones?

1. Someone should port OpenMW to Nintendo switch first.
2. Someone should implement SDL_Sensor-related API for Nintendo switch for the SDL itself (I suppore it should be possible by using the Nintendo SDK API).

**Comment by nikolaykasyanov** [comment]
@akortunov Nintendo Switch Pro Controller can be connected to a PC or Mac.

**Comment by akortunov** [comment]
> Nintendo Switch Pro Controller can be connected to a PC or Mac.

It does not matter since external sensors are not supported in SDL so far.

**Comment by nikolaykasyanov** [comment]
@akortunov the link is final now 🎉

File: apps/openmw/mwinput/inputmanagerimp.hpp [hunk 1]
```diff
@@ -264,6 +296,7 @@ namespace MWInput
         void loadControllerDefaults(bool force = false);
 
         int mFakeDeviceID; //As we only support one controller at a time, use a fake deviceID so we don't lose bindings when switching controllers
+        SDL_Sensor* mGyroScope;
```
nikolaykasyanov: nitpick: why capital `S`?

akortunov:
I suppose the only real blocker left is the Android port itself - it was not fully ported to 2.0.9+ yet.

akortunov:
So the main question what to do with this feature - include it to 0.46 or to 0.47.

psi29a:
What are the benefits of having this in 0.46 as apposed to 0.47?

Are there any issues fixed because of the bump? Do we have better gamepad support?

akortunov:
> What are the benefits of having this in 0.46 as apposed to 0.47?

The only difference is that this feature will be present in the 0.46 stable APK (which is on the Play Market, I suppose) without additional efforts. Otherwise xyzz will need to apply the code from this PR as a very bulk 3d-party patch, or leave this feature only in 0.47 nightlies. Both options are not really good.

> Are there any issues fixed because of the bump?

We can use a single gamepads database (gamecontrollerdb.txt) instead of three different ones for different versions of SDL. Also we can get rid of several preprocessor directives, so code becomes cleaner.

> Do we have better gamepad support?

Because of SDL update? I have no idea. In theory, yes, if there are gamepad-related bugfixes between newer (which we are going to use) and older (which we used for previous OpenMW releases) SDL versions.

psi29a:
I'll take it for 0.46 then. Do we need a changelog entry for this as a 'feature' ?

akortunov:
> Do we need a changelog entry for this as a 'feature'

Yes, we are. I did not add it since I was not sure if this feature should be in 0.46 or 0.47.
I'd also ask English-speakers to verify docs for new settings.

docs/source/reference/modding/settings/input.rst
```diff
@@ -134,3 +134,76 @@ This setting has no effect on the camera rotation speed, which is controlled by
...
+This setting can only be configured by editing the settings configuration file.
+
+gyro horizontal axis
+--------------------
+
+:Type:      string
+:Range:     x, y, z, -x, -y, -z
+:Default:   -x
+
+Maps an axis of gyroscope to the horizontal in-game axis. Minus sign inverses left and right directions.
```
Capostrophic, line 157: This setting sets up an axis of the gyroscope as the horizontal camera axis. Minus sign swaps the positive and negative direction of the axis.

docs/source/reference/modding/settings/input.rst
```diff
@@ -134,3 +134,76 @@ This setting has no effect on the camera rotation speed, which is controlled by
...
+A setting is for the landscape mode, for other modes the game corrects data automatically.
```
Capostrophic, line 158: Keep in mind that while this setting corresponds to the landscape mode of the display, the portrait mode or any other mode will have this axis corrected automatically.

docs/source/reference/modding/settings/input.rst
```diff
@@ -134,3 +134,76 @@ This setting has no effect on the camera rotation speed, which is controlled by
...
+This setting can only be configured by editing the settings configuration file.
+
+gyro vertical axis
+------------------
+
+:Type:      string
+:Range:     x, y, z, -x, -y, -z
+:Default:   y
+
+Maps an axis of gyroscope to the vertical in-game axis. Minus sign inverses up and down directions.
```
Capostrophic, line 169: This setting sets up an axis of the gyroscope as the vertical camera axis. Minus sign swaps the positive and negative direction of the axis.

docs/source/reference/modding/settings/input.rst
```diff
@@ -134,3 +134,76 @@ This setting has no effect on the camera rotation speed, which is controlled by
...
+A setting is for the landscape mode, for other modes the game corrects data automatically.
```
Capostrophic, line 170: Like with horizontal axis, the used axis will be corrected automatically in the portrait mode.

docs/source/reference/modding/settings/input.rst
```diff
@@ -134,3 +134,76 @@ This setting has no effect on the camera rotation speed, which is controlled by
...
+This setting can only be configured by editing the settings configuration file.
+
+gyro input threshold
+--------------------
+
+:Type:		floating point
+:Range:		> 0
+:Default:	0.01
+
+This setting allows to filter a noise (very small values) of rotation speed.
```
Capostrophic, line 181: This setting determines the minimum value of the rotation that will be accepted.
It allows to avoid crosshair oscillation due to gyroscope "noise".

docs/source/reference/modding/settings/input.rst
```diff
@@ -134,3 +134,76 @@ This setting has no effect on the camera rotation speed, which is controlled by
...
+gyro horizontal sensitivity
+---------------------------
+
+:Type: float
+:Range: >0
+:Default: 1.0
+
+This setting controls the overall gyroscope horizontal sensitivity.
+The default sensitivity is 1.0, with smaller values requiring more device rotation,
+and larger values requiring less.
```
Capostrophic, line 194: The smaller this sensitivity is, the less visible effect the device rotation will have on the horizontal camera rotation, and vice versa.

docs/source/reference/modding/settings/input.rst
```diff
@@ -134,3 +134,76 @@ This setting has no effect on the camera rotation speed, which is controlled by
...
+gyro vertical sensitivity
+-------------------------
+
+:Type: float
+:Range: >0
+:Default: 1.0
+
+This setting controls the overall gyroscope vertical sensitivity.
+The default sensitivity is 1.0, with smaller values requiring more device rotation,
+and larger values requiring less.
```
Capostrophic, line 207: The smaller this sensitivity is, the less visible effect the device rotation will have on the vertical camera rotation, and vice versa.

files/settings-default.cfg
```diff
@@ -384,6 +384,24 @@ enable controller = true
 # Emulated gamepad cursor sensitivity.
 gamepad cursor speed = 1.0
 
+# Enable gyroscope support.
+enable gyroscope = false
+
+# gyroscope axis, used to turn around horisontal screen axis.
```
Capostrophic, line 390: Gyroscope axis that corresponds to horizontal camera axis.

files/settings-default.cfg
```diff
@@ -384,6 +384,24 @@ enable controller = true
...
+gyro horizontal axis = -x
+
+# gyroscope axis, used to turn around vertical screen axis.
```
Capostrophic, line 393: Gyroscope axis that corresponds to vertical camera axis.

files/settings-default.cfg
```diff
@@ -384,6 +384,24 @@ enable controller = true
...
+gyro vertical axis = y
+
+# Allows to ignore small amount of movement to avoid crosshair oscillation.
```
Capostrophic, line 396: The minimum gyroscope movement that is able to rotate the camera.

files/settings-default.cfg
```diff
@@ -384,6 +384,24 @@ enable controller = true
...
+gyro input threshold = 0.01
+
+# A giroscope horizontal sensitivity.
```
Capostrophic, line 399: Horizontal camera axis sensitivity to gyroscope movement.

files/settings-default.cfg
```diff
@@ -384,6 +384,24 @@ enable controller = true
...
+gyro horizontal sensitivity = 1.0
+
+# A giroscope vertical sensitivity.
```
Capostrophic, line 402: Vertical camera axis sensitivity to gyroscope movement.

Capostrophic:
By the way, aren't we supposed to downplay OpenMW's playability on non-desktop platforms in official PR as a part of Bethesda relationship? "Gyroscope input" sounds ok but "gyroscope input on Android" is a risky wording when it comes to something that's going to be present in the release announcement newspost.

psi29a:
> By the way, aren't we supposed to downplay OpenMW's playability on non-desktop platforms in official PR as a part of Bethesda relationship? "Gyroscope input" sounds ok but "gyroscope input on Android" is a risky wording when it comes to something that's going to be present in the release announcement newspost.
> 
> Wait, I'm the one writing the newspost anyway, never mind.

No, we're just not allowed to display videos or pictures showing Morrowind running on Android.

If we can show, for example, OpenMicroWave running the Example Suite... we're fine.

We'll also need this information as well if we ever get around to having anything VR related, right?

akortunov:
> Anyway... 'Gyroscopic input' is good enough as it is not just Android specific functionality right?

Well, it should also work on iOS, but I am not sure if it is better for Bethesda than Android.

mp3butcher:
you really want to play openmw in Portrait mode to break compatibility with ubuntu LTS (adopting sdl 2.0.9)...(sarcasm ad lib)

akortunov:
> you really want to play openmw in Portrait mode to break compatibility with ubuntu LTS (adopting sdl>2.0.8)...(sarcasm ad lib)

1. It is not about portrait mode (which is disabled), it is about dropping a deprecated API and using a new API for sensors.
2. OpenMW and its dependencies are in PPA anyway, so nothing is broken.

mp3butcher:
>     1. It is not about portrait mode (which is disabled), it is about dropping a deprecated API and using a new API for sensors.
> 
>     2. OpenMW and its dependencies are in PPA anyway, so nothing is broken.

I understand the utility of gyroscope to rotate camera but
AFAIK sdl 2.0.8 is not deprecated and you could have waited sdl LTS upgrade before merging this pr...
But what's done is done and I will not argue for such a thing

psi29a:
Apologies, @mp3butcher , I'm saving you some embarrassment by removing your statement. Feel free to talk to me in private.

PR #2723, opened 2020-03-17
Do not use dynamic casts when using ObjectState

akortunov:
A main idea - use as() methods for ObjectState, as OSG does and as we do for CustomData.
Allows to get rid of a lot of ugly casts, makes code safer and more readable without noticable difference in the saving/loading speed.

### DIFF
--- a/apps/openmw/mwclass/container.cpp
+++ b/apps/openmw/mwclass/container.cpp
@@ -317,7 +317,6 @@ namespace MWClass
     {
         if (!state.mHasCustomState)
             return;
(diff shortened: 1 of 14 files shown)


AnyOldName3:
OSG returns `NULL` in its default `as*()` implementations to get behaviour more like `dynamic_cast`, so we should probably be returning `nullptr`. If we're expecting casts to always succeed, they should be `static_cast`s instead of `dynamic_cast`s, and `as` methods are slower than those. I don't see much point in an exception-throwing cast versus either of these options.

akortunov:
A difference in performance is negligable even with dynamic casts. The main idea here is to make code more readable. I just took an existing code from our codebase as example.

PR #2725, opened 2020-03-18
Do not write empty trade time

akortunov:
There is no point to write trading timestamp just for every scrib - we set it only when player talks with actor, so the timestamp is empty for most of creatures.
So I just use the same approach as for `mTimeOfDeath`.
No savegame format change is required, since the "TIME" record in savegame is already optional.

The difference is not very noticable (maybe about 0.3-0.5% of save), but it is still a nice improvement.


psi29a:
I'll take it. :)

PR #2728, opened 2020-03-21
Trace down dead persistent actors underwater

akortunov:
Fixes [regression #5317](https://gitlab.com/OpenMW/openmw/-/issues/5317).

Persistent dead actors do not use "swim death" animations and do not float up, so we still should trace them down underwater. I do not know if the Mororwind behaves *exactly* in the same way, though.

As for non-persistent initially dead actors, a behaviour is buggy in Morrowind - dead actor starts to use a common swim animation underwater, as he is still alive. In OpenMW such actors use use "swim death" animations and float up instead.

### DIFF
--- a/apps/openmw/mwworld/worldimp.cpp
+++ b/apps/openmw/mwworld/worldimp.cpp
@@ -1424,7 +1424,9 @@ namespace MWWorld
 
         pos.z() += 20; // place slightly above. will snap down to ground with code below
 
-        if (force || !ptr.getClass().isActor() || (!isFlying(ptr) && !isSwimming(ptr) && isActorCollisionEnabled(ptr)))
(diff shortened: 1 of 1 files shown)


Capostrophic:
Hrnchamd didn't make an exact documentation of the relevant code because according to him it's fairly complicated, more than he would be comfortable with, so we can probably be liberal with it for now.

Though perhaps it'd be cleaner to put an adjustPosition() call with force set to true somewhere around [here](https://github.com/openmw/openmw/blob/master/apps/openmw/mwmechanics/character.cpp#L2331-L2332)?

akortunov:
> Though perhaps it'd be cleaner to put an adjustPosition() call with force set to true somewhere around here?

It will not be cleaner, the mess just will be in the different place.

PR #2730, opened 2020-03-21
QT5 link is fixed 

strikersix23:
Compilation now works on windows msvc

### DIFF
--- a/CI/before_script.msvc.sh
+++ b/CI/before_script.msvc.sh
@@ -421,7 +421,7 @@ if [ -z $SKIP_DOWNLOAD ]; then
 		fi
 
 		download "Qt 5.7.0" \
-			"https://download.qt.io/new-archive/qt/5.7/5.7.0/qt-opensource-windows-x86-msvc${MSVC_YEAR}${QT_SUFFIX}-5.7.0.exe" \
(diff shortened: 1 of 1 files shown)


CI/before_script.msvc.sh
```diff
@@ -421,7 +421,7 @@ if [ -z $SKIP_DOWNLOAD ]; then
 		fi
 
 		download "Qt 5.7.0" \
-			"https://download.qt.io/archive/qt/5.7/5.7.0/qt-opensource-windows-x86-msvc${MSVC_YEAR}${QT_SUFFIX}-5.7.0.exe" \
+			"http://download.qt.io/new_archive/qt/5.7/5.7.0/qt-opensource-windows-x86-msvc${MSVC_YEAR}${QT_SUFFIX}-5.7.0.exe" \
```
akortunov: Why do you use http instead of https?

psi29a: Agreed, if possible, switch back to https please.

lostmsu: Tested, works with https.

akortunov: Fixed upstream to do not waste a time for rebasing.

PR #2733, opened 2020-03-22
give sdl2 2.0.12 for windows a try

psi29a:


psi29a:
msvc2017 and 2019 seems to be green... done

PR #2738, opened 2020-03-23
[WIP] use c++11 (and up) to make code more readable

psi29a:
This is getting me back into looking at the code again, low hanging fruit really.

This changes no functionality but should increase readability and consistency.

Additional comments, thoughts and testing welcome.

### DIFF
--- a/apps/bsatool/bsatool.cpp
+++ b/apps/bsatool/bsatool.cpp
@@ -24,7 +24,7 @@ struct Arguments {
     bool fullpath{};
 };
 
-void replaceAll(std::string& str, const std::string& needle, const std::string& substitute)
+void replaceAll(std::string &str, const std::string &needle, const std::string &substitute)
 {
     size_t pos = str.find(needle);
     while (pos != std::string::npos)
@@ -34,7 +34,7 @@ void replaceAll(std::string& str, const std::string& needle, const std::string& 
     }
 }
 
-bool parseOptions(int argc, char** argv, Arguments& info)
+bool parseOptions(int argc, char **argv, Arguments &info)
 {
     bpo::options_description desc("Inspect and extract files from Bethesda BSA archives\n\n"
                                   "Usages:\n"
@@ -74,7 +74,7 @@ bool parseOptions(int argc, char** argv, Arguments& info)
                 .options(all).positional(p).run();
         bpo::store(valid_opts, variables);
     }
-    catch (std::exception& e)
+    catch (std::exception &e)
     {
         std::cout << "ERROR parsing arguments: " << e.what() << "\n\n"
                   << desc << std::endl;
@@ -141,11 +141,11 @@ bool parseOptions(int argc, char** argv, Arguments& info)
     return true;
 }
 
-int list(Bsa::BSAFile& bsa, Arguments& info);
-int extract(Bsa::BSAFile& bsa, Arguments& info);
-int extractAll(Bsa::BSAFile& bsa, Arguments& info);
+int list(Bsa::BSAFile &bsa, Arguments &info);
+int extract(Bsa::BSAFile &bsa, Arguments &info);
+int extractAll(Bsa::BSAFile &bsa, Arguments &info);
 
-int main(int argc, char** argv)
+int main(int argc, char **argv)
 {
     try
     {
@@ -169,17 +169,17 @@ int main(int argc, char** argv)
             return 1;
         }
     }
-    catch (std::exception& e)
+    catch (std::exception &e)
     {
         std::cerr << "ERROR reading BSA archive\nDetails:\n" << e.what() << std::endl;
         return 2;
     }
 }
 
-int list(Bsa::BSAFile& bsa, Arguments& info)
+int list(Bsa::BSAFile &bsa, Arguments &info)
 {
     // List all files
-    const Bsa::BSAFile::FileList& files = bsa.getList();
+    const Bsa::BSAFile::FileList &files = bsa.getList();
     for (auto file : files)
     {
         if (info.longformat)
@@ -198,7 +198,7 @@ int list(Bsa::BSAFile& bsa, Arguments& info)
     return 0;
 }
 
-int extract(Bsa::BSAFile& bsa, Arguments& info)
+int extract(Bsa::BSAFile &bsa, Arguments &info)
 {
     std::string archivePath = info.extractfile;
     replaceAll(archivePath, "/", "\\");
@@ -247,15 +247,15 @@ int extract(Bsa::BSAFile& bsa, Arguments& info)
     return 0;
 }
 
-int extractAll(Bsa::BSAFile& bsa, Arguments& info)
+int extractAll(Bsa::BSAFile &bsa, Arguments &info)
 {
     // Get the list of files present in the archive
     Bsa::BSAFile::FileList list = bsa.getList();
 
     // Iter on the list
-    for (auto& it : list)
+    for (auto &it : list)
     {
-        const char* archivePath = it.name;
+        const char *archivePath = it.name;
 
         std::string extractPath(archivePath);
         replaceAll(extractPath, "\\", "/");
--- a/apps/bsatool/bsatool.cpp
+++ b/apps/bsatool/bsatool.cpp
@@ -15,13 +15,13 @@ namespace bpo = boost::program_options;
 namespace bfs = boost::filesystem;
 
 struct Arguments {
-  std::string mode;
-  std::string filename;
-  std::string extractfile;
-  std::string outdir;
+    std::string mode;
+    std::string filename;
+    std::string extractfile;
+    std::string outdir;
 
-  bool longformat{};
-  bool fullpath{};
+    bool longformat{};
+    bool fullpath{};
 };
 
 void replaceAll(std::string& str, const std::string& needle, const std::string& substitute)
--- a/apps/bsatool/bsatool.cpp
+++ b/apps/bsatool/bsatool.cpp
@@ -14,54 +14,51 @@
 namespace bpo = boost::program_options;
 namespace bfs = boost::filesystem;
 
-struct Arguments
-{
-    std::string mode;
-    std::string filename;
-    std::string extractfile;
-    std::string outdir;
+struct Arguments {
+  std::string mode;
+  std::string filename;
+  std::string extractfile;
+  std::string outdir;
 
-    bool longformat{};
-    bool fullpath{};
+  bool longformat{};
+  bool fullpath{};
 };
 
 void replaceAll(std::string& str, const std::string& needle, const std::string& substitute)
 {
     size_t pos = str.find(needle);
-    while(pos != std::string::npos)
+    while (pos != std::string::npos)
     {
         str.replace(pos, needle.size(), substitute);
         pos = str.find(needle);
     }
 }
 
-bool parseOptions (int argc, char** argv, Arguments &info)
+bool parseOptions(int argc, char** argv, Arguments& info)
 {
     bpo::options_description desc("Inspect and extract files from Bethesda BSA archives\n\n"
-            "Usages:\n"
-            "  bsatool list [-l] archivefile\n"
-            "      List the files presents in the input archive.\n\n"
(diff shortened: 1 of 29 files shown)


apps/bsatool/bsatool.cpp
```diff
@@ -256,8 +256,8 @@ int extractAll(Bsa::BSAFile& bsa, Arguments& info)
     Bsa::BSAFile::FileList list = bsa.getList();
 
     // Iter on the list
-    for(Bsa::BSAFile::FileList::iterator it = list.begin(); it != list.end(); ++it) {
-        const char* archivePath = it->name;
+    for(auto & it : list) {
```
Capostrophic: Typically ampersands follow the type without an additional space in the code base.

apps/essimporter/importer.hpp
```diff
@@ -9,7 +9,7 @@ namespace ESSImport
     class Importer
     {
     public:
-        Importer(const std::string& essfile, const std::string& outfile, const std::string& encoding);
+        Importer(std::string  essfile, std::string  outfile, std::string  encoding);
```
Capostrophic: Double spaces here.

apps/niftest/niftest.cpp
```diff
@@ -18,106 +16,88 @@ namespace bpo = boost::program_options;
...
-            }
-            else if(isBSA(name))
-            {
-                if(!archivePath.empty() && !isBSA(archivePath))
-                {
+        try {
+            if (isNIF(name)) {
+                //           std::cout << "Decoding: " << name << std::endl;
+                Nif::NIFFile temp_nif(myManager.get(name), archivePath + name);
+            } else if (isBSA(name)) {
```
Capostrophic: This kind of if-else if formatting is inconsistent with Allman indentation style typically used in the code base.

apps/niftest/niftest.cpp
```diff
@@ -18,106 +16,88 @@ namespace bpo = boost::program_options;
...
 //                     std::cout << "Reading BSA File: " << name << std::endl;
-                    readVFS(new VFS::BsaArchive(archivePath+name),archivePath+name+"/");
+                    readVFS(new VFS::BsaArchive(archivePath + name), archivePath + name + "/");
 //                     std::cout << "Done with BSA File: " << name << std::endl;
                 }
             }
         }
-        catch (std::exception& e)
-        {
+        catch (std::exception &e) {
```
Capostrophic: yeah, pls don't do this

psi29a: Is it Allman or GTFO, or possibility of finally going through with a 'style' brush and cleaning up so that everything is consistent?

I'm partial to K&R, what's the consensus...

AnyOldName3: Allman.

apps/bsatool/bsatool.cpp
psi29a: Also Allman?

apps/niftest/niftest.cpp
```diff
@@ -18,106 +16,88 @@ namespace bpo = boost::program_options;
...
             std::cerr << "ERROR, an exception has occurred:  " << e.what() << std::endl;
         }
     }
 }
 
-bool parseOptions (int argc, char** argv, std::vector<std::string>& files)
-{
+bool parseOptions(int argc, char **argv, std::vector<std::string> &files) {
```
zinnschlag: That is not how we place our { normally. I am aware that the other style is present in some (mostly older) files and I assume you made this change to improve consistency. I don't agree with it. Consistency should be seen across the whole project and not just a single file.

apps/bsatool/bsatool.cpp
```diff
@@ -14,54 +14,51 @@
...
     bpo::options_description desc("Inspect and extract files from Bethesda BSA archives\n\n"
-            "Usages:\n"
-            "  bsatool list [-l] archivefile\n"
-            "      List the files presents in the input archive.\n\n"
-            "  bsatool extract [-f] archivefile [file_to_extract] [output_directory]\n"
-            "      Extract a file from the input archive.\n\n"
-            "  bsatool extractall archivefile [output_directory]\n"
-            "      Extract all files from the input archive.\n\n"
-            "Allowed options");
+                                  "Usages:\n"
```
zinnschlag: Is that an improvement? I think we are not entirely consistent with indenting in cases like this. But from what I remember the bulk of the code is using the style present before this change. So this would decrease consistency overall.

apps/bsatool/bsatool.cpp
```diff
@@ -74,69 +71,69 @@ bool parseOptions (int argc, char** argv, Arguments &info)
...
         std::cout << "ERROR parsing arguments: " << e.what() << "\n\n"
-            << desc << std::endl;
+                  << desc << std::endl;
         return false;
     }
 
     bpo::notify(variables);
 
-    if (variables.count ("help"))
+    if (variables.count("help"))
```
zinnschlag: Another case of not entirely consistent style. From what I remember the style with an additional space here is more common. I could be wrong about that though.

apps/niftest/niftest.cpp
psi29a: I'm going back though, it should be Allman style.

components/esm/creaturestats.hpp
```diff
@@ -36,7 +36,7 @@ namespace ESM
         std::vector<int> mSummonGraveyard;
 
         ESM::TimeStamp mTradeTime;
-        int mGoldPool;
+        unsigned int mGoldPool;
```
akortunov: The mGoldPool in the mwmechanics/creaturestats.hpp is still declared as int.

apps/bsatool/bsatool.cpp
psi29a: We had a talk about this with the team, the idea is to create a .clangformat that everyone can use so that we're all on the same page consistency. The point of my PR was not style, but it seems to have gone in that direction due my K&R leanings.  Moving into Allman style.

apps/bsatool/bsatool.cpp
psi29a: Yes, this is not consistent throughout the code base, sometimes with or without a space.

Perhaps we should talk about this in the forum and come up with a .clangformat that everyone can use?

components/esm/creaturestats.hpp
zinnschlag: Unsure about this one. Are we actually using unsigned int across the board for gold amounts? If not then this chance has the potential to introduce subtle bugs and inconsistencies.

psi29a: I haven't touched anything yet in apps/openmw yet.This was to fix an issue elsewhere.

psi29a: Not yet, as noted above by @akortunov, I haven't gotten to apps/openmw yet.

apps/bsatool/bsatool.cpp
psi29a: https://forum.openmw.org/viewtopic.php?f=42&t=6790  <-- for you @zinnschlag :D
(and @Capostrophic @akortunov )

psi29a: Allman doesn't address spaces all that much though.

components/esm/creaturestats.hpp
zinnschlag: It should stay int IMO. unsigned int tends to not follow the principle of least astonishment when used in calculation. Besides, if a 31-bit type (signed int) isn't big enough to hold a value, a 32-bit type is unlikely to be a sufficient improvement. There are really no significant benefits to using unsigned here.

psi29a: It was because of this:
https://github.com/OpenMW/openmw/blob/fc38467c1a9d0f76869b8eb3d26b2417f60264c8/apps/essimporter/importacdt.hpp#L45

```
#pragma pack(1)
    struct ACDT
    {
        // Note, not stored at *all*:
        // - Level changes are lost on reload, except for the player (there it's in the NPC record).
        unsigned char mUnknown[12];
        unsigned int mFlags;
        float mBreathMeter; // Seconds left before drowning
        unsigned char mUnknown2[20];
        float mDynamic[3][2];
        unsigned char mUnknown3[16];
        float mAttributes[8][2];
        float mMagicEffects[27]; // Effect attributes: https://wiki.openmw.org/index.php?title=Research:Magic#Effect_attributes
        unsigned char mUnknown4[4];
        unsigned int mGoldPool;
        unsigned char mCountDown; // seen the same value as in ACSC.mCorpseClearCountdown, maybe
                                  // this one is for respawning?
        unsigned char mUnknown5[3];
    };
```

zinnschlag: I agree. The question is, which one is correct (i.e. compatible with vanilla MW). And if it is unsigned int, should we use the opportunity to switch to a signed 64-bit type internally, meaning for anything that does not interact with legacy file formats. I bet there are some wacky players out there that are not far from breaking the 4 billion limit.

psi29a: Depending on your platform, int has been shifted to mean different things. Perhaps at some point we should consider moving to explicit types: uint64_t ?
https://en.cppreference.com/w/cpp/types/integer

zinnschlag: For data structures used to load/save definitely. I was under the impression that we are doing that already. But maybe I misremember.
Generally I prefer to use the basic types. In the 64 bit case however using one of the new explicit types is necessary, because Microsoft (long only 32 bit, grrr).
Again, I would caution against a move to unsigned. That has the potential to break things. For example I remember that item counts use out of band values (-1) in some places. Not sure if the same is true for gold values. But it could be. If we make such a change (that I am not in favour of) this would have to be checked carefully.

apps/bsatool/bsatool.cpp
unelsson: [I like spaces!](https://upload.wikimedia.org/wikipedia/en/thumb/1/14/Ralph_Wiggum.png/220px-Ralph_Wiggum.png)

components/esm/creaturestats.hpp
psi29a: Right, I'll revert this particular change for now and look into it again in another PR.

apps/essimporter/converter.cpp
```diff
@@ -139,25 +137,25 @@ namespace ESSImport
         image2->allocateImage(width, height, 1, GL_RGBA, GL_UNSIGNED_BYTE);
         memcpy(image2->data(), &data[0], data.size());
 
-        for (std::set<std::pair<int, int> >::const_iterator it = mContext->mExploredCells.begin(); it != mContext->mExploredCells.end(); ++it)
+        for (const auto & mExploredCell : mContext->mExploredCells)
```
elsid:
```suggestion
        for (const auto & exploredCell : mContext->mExploredCells)
```

apps/niftest/niftest.cpp
```diff
@@ -100,24 +96,24 @@ bool parseOptions (int argc, char** argv, std::vector<std::string>& files)
...
         }
         if (variables.count("input-file"))
         {
-            files = variables["input-file"].as< std::vector<std::string> >();
+            files = variables["input-file"].as<std::vector<std::string>>();
             return true;
         }
     }
-    catch(std::exception &e)
+    catch (std::exception& e)
```
elsid:
```suggestion
    catch (const std::exception& e)
```

apps/niftest/niftest.cpp
```diff
@@ -129,40 +125,38 @@ bool parseOptions (int argc, char** argv, std::vector<std::string>& files)
...
+    for (auto name : files)
     {
-        std::string name = *it;
-
         try
         {
-            if(isNIF(name))
+            if (isNIF(name))
+            {
+                // std::cout << "Decoding: " << name << std::endl;
```
akortunov, line 138: I suppose that these comments are not really needed.

psi29a: Jup. I plan on taking a bit more time in this PR, no style changes, purely functional changes. I'll take care of stuff like above to. I don't want this PR to rot and now that we're branched, we can see about getting this into master.

akortunov: You can take a look at the #2903, which does pretty the same thing, but without formatting changes.

PR #2740, opened 2020-03-25
Ignore unterminated empty strings

Assumeru:
Makes the script(s) in [this thread](https://forum.openmw.org/viewtopic.php?f=40&t=6781) compile.

This still fails in at least two **hypothetical** cases:
1. StartScript DN_MinionDrain_s" <- final whitespace
2. StartScript DN_MinionDrain_s" ;comment

Does anyone feel those are worth fixing as well?

### DIFF
--- a/components/compiler/scanner.cpp
+++ b/components/compiler/scanner.cpp
@@ -282,6 +282,8 @@ namespace Compiler
 
         if (!scanName (name))
             return false;
+        else if(name.empty())
+            return true;
 
         TokenLoc loc (mLoc);
         mLoc.mLiteral.clear();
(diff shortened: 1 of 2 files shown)


components/compiler/scanner.cpp
```diff
@@ -282,6 +282,8 @@ namespace Compiler
 
         if (!scanName (name))
             return false;
+        else if(name.empty())
```
Capostrophic: This shouldn't be necessary in the case you try to avoid (since "error" is not true), but if somehow an incomplete string that didn't pass your check below is empty it could make it possible for the parser to utilize empty strings in more cases than intended. Are you sure it's essential to fix the script?

components/compiler/scanner.cpp
```diff
@@ -366,6 +368,13 @@ namespace Compiler
                 {
                     if (mIgnoreNewline)
                         mErrorHandler.warning ("string contains newline character, make sure that it is intended", mLoc);
+                    else if (name.size() == 1 || name.size() == 2 && name[1] == '\r')
```
Capostrophic: This combination of || and && is ambiguous.

Assumeru: Why? && has higher precedence than ||

Capostrophic: It's still an undefined behavior. And it makes GCC emit a warning.

Assumeru: I'm very happy to hear it's not undefined behaviour. I know c++ is a bit loose sometimes, but that would've been scary.

Anyway, brackets added.

PR #2746, opened 2020-03-29
Remove "extra argument" warning from AiTravel

Assumeru:
AiTravel takes a reset flag per https://gitlab.com/OpenMW/openmw/-/issues/1465 and https://www.tamriel-rebuilt.org/content/tutorial-morrowind-scripting-dummies but OpenMW currently emits a warning when a script contains said flag.

### DIFF
--- a/components/compiler/extensions0.cpp
+++ b/components/compiler/extensions0.cpp
@@ -33,7 +33,7 @@ namespace Compiler
         {
             extensions.registerInstruction ("aiactivate", "c/l", opcodeAIActivate,
                 opcodeAIActivateExplicit);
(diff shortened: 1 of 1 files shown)


Capostrophic:
That'll break scripts that put a string after the arguments, it's not so simple.

Assumeru:
I was going to think of something else and made a test script. Turns out it doesn't break (any more).

`Warning: _travel line 5, column 28 ("Wilderness"): Parsing a non-variable string as a number: 0`

Capostrophic:
Hm, okay then. Since Morrowind doesn't actually care what the value of the 4th flag is as long as it's present (if it's present, then the behavior of the package is changed), it'll probably make the implementation easier.

PR #2747, opened 2020-03-30
[WIP] An attempt to fix context loss on Android

akortunov:
I try to use a slightly better approach than [global variable](https://github.com/xyzz/openmw-android/blob/master/buildscripts/patches/openmw/0010-android-fix-context-being-lost-on-app-minimize.patch).

Note: If we do not want to maintain android_main file for some reason, probably it would be better to drop it (Android port already fully replaces it).

### DIFF
--- a/apps/openmw/android_main.cpp
+++ b/apps/openmw/android_main.cpp
@@ -4,6 +4,9 @@ int stderr = 0; // Hack: fix linker error
 #include <SDL_gamecontroller.h>
 #include <SDL_mouse.h>
 #include <SDL_events.h>
+#include <SDL_hints.h>
+
+#include "mwbase/inputmanager.hpp"
 
 /*******************************************************************************
  Functions called by JNI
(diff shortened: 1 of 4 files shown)


akortunov:
@xyzz, @psi29a, what do you think?

psi29a:
I would personally love to have fork work with and sync with upstream; instead of it being like a long-living branch. I'm still not worried so much about any 'legal' issues.

xyzz:
It won't build like that, it's missing for example a patch adding the implementation for `g_viewer`.

psi29a:
@xyzz and now?

akortunov:
> and now?

Do not touch this PR for now.

akortunov:
Probably it is worth to investigate how work SDL_GL_CONTEXT_RELEASE_BEHAVIOR helper and SDL_APP_DIDENTERBACKGROUND/SDL_APP_DIDENTERFOREGROUND events.

PR #2749, opened 2020-03-31
Pause audio and video when the game is minimized

akortunov:
Implements [feature #4944](https://gitlab.com/OpenMW/openmw/-/issues/4944).

Summary of changes:
1. Turn `mPausedSoundTypes` into the map, so we can store a collection of different blockers and resume only sound types without blockers left.
2. Use ALC_SOFT_PAUSE_DEVICE to pause/resume playback when the game is minimized.
3. Use an existing API to pause/resume video playback as well.

According to testing results, this PR renders this [this patch](https://github.com/xyzz/openmw-android/blob/master/buildscripts/patches/openmw/0012-engine-pause-sounds-when-game-is-paused.patch) obsolete.

Note: this feature does not fix the [bug #5320](https://gitlab.com/OpenMW/openmw/-/issues/5320), so the game still may not be minimized properly on MacOS.

An example of usage of per-channel blockers:
```
sndMgr->pauseSounds("Overall pause"); // pause everything
sndMgr->pauseSounds("MyVoicePause", MWSound::Type::Voice); // pause only voices
...
sndMgr->resumeSounds("Overall pause"); // resume everything, voices are still paused
sndMgr->resumeSounds("MyVoicePause"); // resume voices
```

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -96,7 +96,12 @@ bool OMW::Engine::frame(float frametime)
         // If we are not currently rendering, then RenderItems will not be reused resulting in a memory leak upon changing widget textures (fixed in MyGUI 3.3.2),
         // and destroyed widgets will not be deleted (not fixed yet, https://github.com/MyGUI/mygui/issues/21)
         if (!mEnvironment.getInputManager()->isWindowVisible())
+        {
+            mEnvironment.getSoundManager()->pausePlayback();
             return false;
+        }
+        else
+            mEnvironment.getSoundManager()->resumePlayback();
 
         // sound
         if (mUseSound)
--- a/apps/openmw/mwbase/soundmanager.hpp
+++ b/apps/openmw/mwbase/soundmanager.hpp
@@ -14,6 +14,14 @@ namespace MWWorld
 
 namespace MWSound
 {
+    // Each entry excepts of MaxCount should be used only in one place
+    enum BlockerType
+    {
+        VideoPlayback,
+
+        MaxCount
+    };
+
     class Sound;
     class Stream;
     struct Sound_Decoder;
@@ -168,10 +176,10 @@ namespace MWBase
             ///< Is the given sound currently playing on the given object?
             ///  If you want to check if sound played with playSound is playing, use empty Ptr
 
-            virtual void pauseSounds(const std::string& blockerId, int types=int(Type::Mask)) = 0;
+            virtual void pauseSounds(MWSound::BlockerType blocker, int types=int(Type::Mask)) = 0;
             ///< Pauses all currently playing sounds, including music.
 
-            virtual void resumeSounds(const std::string& blockerId) = 0;
+            virtual void resumeSounds(MWSound::BlockerType blocker) = 0;
             ///< Resumes all previously paused sounds.
 
             virtual void pausePlayback() = 0;
--- a/apps/openmw/mwbase/soundmanager.hpp
+++ b/apps/openmw/mwbase/soundmanager.hpp
@@ -174,6 +174,9 @@ namespace MWBase
             virtual void resumeSounds(const std::string& blockerId) = 0;
             ///< Resumes all previously paused sounds.
 
+            virtual void pausePlayback() = 0;
+            virtual void resumePlayback() = 0;
+
             virtual void update(float duration) = 0;
 
             virtual void setListenerPosDir(const osg::Vec3f &pos, const osg::Vec3f &dir, const osg::Vec3f &up, bool underwater) = 0;
--- a/apps/openmw/mwbase/soundmanager.hpp
+++ b/apps/openmw/mwbase/soundmanager.hpp
@@ -168,10 +168,10 @@ namespace MWBase
             ///< Is the given sound currently playing on the given object?
             ///  If you want to check if sound played with playSound is playing, use empty Ptr
 
-            virtual void pauseSounds(int types=static_cast<int>(Type::Mask)) = 0;
+            virtual void pauseSounds(const std::string& blockerId, int types=int(Type::Mask)) = 0;
             ///< Pauses all currently playing sounds, including music.
 
-            virtual void resumeSounds(int types=static_cast<int>(Type::Mask)) = 0;
+            virtual void resumeSounds(const std::string& blockerId) = 0;
             ///< Resumes all previously paused sounds.
 
             virtual void update(float duration) = 0;
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -857,30 +857,29 @@ namespace MWSound
         return false;
     }
 
-    void SoundManager::pauseSounds(const std::string& blockerId, int types)
+    void SoundManager::pauseSounds(BlockerType blocker, int types)
     {
         if(mOutput->isInitialized())
         {
-            auto found = mPausedSoundTypes.find(blockerId);
-            if (found != mPausedSoundTypes.end() && found->second != types)
-                resumeSounds(blockerId);
+            if (mPausedSoundTypes[blocker] != 0)
+                resumeSounds(blocker);
 
             types = types & Type::Mask;
             mOutput->pauseSounds(types);
-            mPausedSoundTypes[blockerId] = types;
+            mPausedSoundTypes[blocker] = types;
         }
     }
 
-    void SoundManager::resumeSounds(const std::string& blockerId)
+    void SoundManager::resumeSounds(BlockerType blocker)
     {
         if(mOutput->isInitialized())
         {
-            mPausedSoundTypes.erase(blockerId);
+            mPausedSoundTypes[blocker] = 0;
             int types = int(Type::Mask);
-            for (auto& blocker : mPausedSoundTypes)
+            for (int currentBlocker = 0; currentBlocker < BlockerType::MaxCount; currentBlocker++)
             {
-                if (blocker.first != blockerId)
-                    types &= ~blocker.second;
+                if (currentBlocker != blocker)
+                    types &= ~mPausedSoundTypes[currentBlocker];
             }
 
             mOutput->resumeSounds(types);
@@ -1425,7 +1424,7 @@ namespace MWSound
             mUnusedStreams.push_back(sound);
         }
         mActiveTracks.clear();
-        mPausedSoundTypes.clear();
         mPlaybackPaused = false;
+        std::fill(std::begin(mPausedSoundTypes), std::end(mPausedSoundTypes), 0);
     }
 }
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -52,6 +52,7 @@ namespace MWSound
         , mListenerUp(0,0,1)
         , mUnderwaterSound(nullptr)
         , mNearWaterSound(nullptr)
+        , mPlaybackPaused(false)
     {
         mMasterVolume = Settings::Manager::getFloat("master volume", "Sound");
         mMasterVolume = std::min(std::max(mMasterVolume, 0.0f), 1.0f);
@@ -856,7 +857,6 @@ namespace MWSound
         return false;
     }
 
-
     void SoundManager::pauseSounds(const std::string& blockerId, int types)
     {
         if(mOutput->isInitialized())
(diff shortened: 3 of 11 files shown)


kcat:
If you want to pause all playback, you'd be better off using [`ALC_SOFT_pause_device`](https://openal-soft.org/openal-extensions/SOFT_pause_device.txt), which will stop all mixing-related processing (nicer on system resources when it's not in focus).

It might also be better to pause audio (if not the whole game) when the window loses focus, rather than on minimize. Perhaps a config option to specify if the game and audio should auto-pause on being minimized, loss of focus, or never.

I wouldn't recommend using string IDs and maps like that, since the flags are global control. The code should know if certain types of sounds are supposed to unpause before telling the sound system to unpause them, and spreading out that kind of control would make it difficult to keep track of (who is ultimately controlling when certain types of sounds should start playing again?).

akortunov:
> If you want to pause all playback, you'd be better off using ALC_SOFT_pause_device

Thanks, I'll take a look.

> It might also be better to pause audio (if not the whole game) when the window loses focus

No, it would not. For example, I use the "minimize on focus loss = false" since I do not want to top a game when I open a drop-down terminal (yakuake). When a game continue its running, there is no point to stop a sound system.

> The code should know if certain types of sounds are supposed to unpause before telling the sound system to unpause them

Actually, the advantage of separate IDs is that the code **should not** know about different code which wants to stop/resume the same sound type. I took [this snippet](https://docs.devexpress.com/eXpressAppFramework/DevExpress.ExpressApp.View.AllowEdit) as an example. An approach is pretty the same, but I use integers insteads of bool flags. 
The only drawback here is possible ID collision, but we do not use this function such widely.

akortunov:
From what I can tell, the ALC_SOFT_pause_device was added somewhere in 1.16, so we will have to limit a required OpenAL-Soft version to use it.

kcat:
> No, it would not. For example, I use the "minimize on focus loss = false" since I do not want to stop a game when I open a drop-down terminal (yakuake).

Similarly, if some window pops up and takes focus away from the game while I'm playing, I would want it to pause since I don't have control (get stuck walking, unable to fight enemies, etc). An option to pause on focus loss (the game and sound) would be good to have.

> There is no "control" here, here is just "asking". Different client may ask sound system to start/stop several types of sounds, but it is up to the sound system itself to dispatch such requests.

Well, I'd still at least recommend against using strings. A fixed set of IDs would work better since there would be no risk at runtime of a misspelled ID, and it doesn't need to be hashed for a key lookup. It can just be a fixed array that just ORs its elements to determine which types should be paused.

akortunov:
> An option to pause on focus loss

"minimize on focus loss" is already is supposed to tell if we need to minimize app (and stop all activities) or not. But according to SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS's docs, it works only in fullscreen mode.

kcat:
> From what I can tell, the ALC_SOFT_pause_device was added somewhere in 1.16, so we will have to limit a required OpenAL-Soft version to use it.

The way it's handled in GZDoom, if the extension exists, it's used to pause the device. It then also sets the listener gain to 0, effectively muting everything as a fallback. So if the extension exists, playback stops. If it doesn't, the game still goes quiet (1.16 was released almost 6 years ago, and I'd expect many other issues if using such an outdated version).

> "minimize on focus loss" is already is supposed to tell if we need to minimize app (and stop all activities) or not.

I wouldn't want to minimize on focus loss, that's even more jarring than something taking control. The window suddenly disappearing like that would actually make me think it crashed. Unless the game window is blocking the offending window that took focus, in which case yeah, it's best to ensure the focused window is visible (one point against using exclusive fullscreen modes).

akortunov:
> I wouldn't want to minimize on focus loss, that's even more jarring than something taking control. The window suddenly disappearing like that would actually make me think it crashed.

Again, this setting works only in fullscreen mode.

If you want to change how OpenMW handles focus loss events (we did not change it for years), fell free to start a new thread on forum. This one is about how we should handle sound and video playback when OpenMW is minimized, and we do not render or update anything.

akortunov:
I use the same approach as in GZDoom now, but I still need to keep per-channel blockers - we use such API for video (pause existing non-video sounds when we start to play video).

apps/openmw/engine.cpp
```diff
@@ -96,7 +96,12 @@ bool OMW::Engine::frame(float frametime)
         // If we are not currently rendering, then RenderItems will not be reused resulting in a memory leak upon changing widget textures (fixed in MyGUI 3.3.2),
         // and destroyed widgets will not be deleted (not fixed yet, https://github.com/MyGUI/mygui/issues/21)
         if (!mEnvironment.getInputManager()->isWindowVisible())
+        {
+            mEnvironment.getSoundManager()->pausePlayback();
             return false;
+        }
+        else
+            mEnvironment.getSoundManager()->resumePlayback();
```
kcat: Making this call every frame the window is visible seems wasteful. The following call to `SoundManager::update` can probably handle it without needing a redundant call through the interface.

akortunov: In this case we will need to call the `getInputManager()->isWindowVisible()` every frame in the
`SoundManager::update`, so performance will be pretty the same.

kcat: From what I can see here, `SoundManager::update` doesn't get called if `isWindowVisible` returns false. Is there some other place `SoundManager::update` gets called where that's not true? Alternatively, you could save the result of the `isWindowVisible` call and pass it as a parameter to `SoundManager::update`.

akortunov: > From what I can see here, SoundManager::update doesn't get called if isWindowVisible returns false.

But the SoundManager itself does not know about it, and people who read its code too.

akortunov:
Any other objections?

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -858,27 +857,36 @@ namespace MWSound
...
         if(mOutput->isInitialized())
         {
+            auto found = mPausedSoundTypes.find(blockerId);
+            if (found != mPausedSoundTypes.end() && found->second != types)
+                resumeSounds(blockerId);
+
             types = types & Type::Mask;
             mOutput->pauseSounds(types);
-            mPausedSoundTypes |= types;
+            mPausedSoundTypes[blockerId] = types;
```
kcat, line 870: Won't this unpause all sounds currently paused by `blockerId`, then repause what's being set? The given types should combine with what's already there for the given blocker, or at the very least, shouldn't unpause sounds that are going to get paused again.

akortunov: > Won't this unpause all sounds currently paused by blockerId, then repause what's being set?

It is for a quite specific case when user set a different mask for already used blocker. Yes, they the only readable approach I was able to get working here is to "remove an old blocker and set a new one".

kcat:
I still don't like using text strings for the IDs, since it makes it more error-prone with misspellings not being caught at compile-time. A set of enum IDs using a fixed-size array of paused types would work just as well, if not better, and have improved compile-time safety.

But other than that, nothing else is popping out at me.

akortunov:
> A set of enum IDs using a fixed-size array of paused types

Sorry, I did not get what you mean. If we create an enum with one entry (for video) and array with fixed size 1, it really will make no sense -  in this case we can just rename `pauseSounds / resumeSounds` to `pauseSoundsForVideo / resumeSoundsForVideo` to make it clear that developers can not use this API for other purposes.

kcat:
> > A set of enum IDs using a fixed-size array of paused types
> 
> Sorry, I did not get what you mean. If we create an enum with one entry (for video) and array with fixed size 1, it really will make no sense - in this case we can just rename `pauseSounds / resumeSounds` to `pauseSoundsForVideo / resumeSoundsForVideo` to make it clear that developers can not use this API for other purposes.

Except it would make it easier to add another ID if another use-case comes up. E.g.
```
// Might want to make these names more unique, or put them in a namespace or something.
enum {
    Video,

    MaxCount
};

...

int mPausedSoundTypes[MaxCount] = {};
```
If another use-case comes along, say wanting to pause certain sounds with the menu up, you can just add a new `Menu` ID under `Video` and use it.

akortunov:
Yes, exactly what I was talking about - I see no point to implement such thing.
By design, enums are needed to re-use their entries, but our case is pretty opposite - we want to DO NOT re-use them. Also enum with one entry is a real strange thing.

If you really care about performance here, it would be simpler to rename `pauseSounds / resumeSounds` as I suggested above to make sure that developers will not use it.
But such API is supposed to be used in quite rare cases, so performance is not critical here.

kcat:
> Yes, exactly what I was talking about - I see no point to implement such thing.
> By design, enums are needed to re-use their entries, but our case is pretty opposite - we want to DO NOT re-use them. Also enum with one entry is a really strange thing.

I'm not understanding what the goal is, then. It seems to me the whole point of having per-type blockers is that each blocker can be uniquely identified, with new ones being added as needed. If it's not supposed to be reused, that it's just for pausing non-video sound during video playback, why have per-type blockers instead of simply flagging 'a video is playing' in the sound manager that keeps all non-video streams paused?

> If you really care about performance here, it would be simpler to rename `pauseSounds / resumeSounds` as I suggested above to make sure that developers will not use it.

More about safety than performance. Say one piece of code calls `sndMgr->pauseSounds("Video")` and another calls `sndMgr->resumeSounds("video")`, you'll have no warning that anything is wrong. You'd only find out there's a problem when playing a game after watching a video that the sounds don't resume, and even then the actual problem (a misspelled ID) won't be immediately apparent. An enum would ensure compile-time correctness, catching the potential bug early, since a `video` enum wouldn't exist if it's `Video`.

akortunov:
> I'm not understanding what the goal is, then.

Read the discussion in the related feature description.
Basically, an existing API to pause/resume sounds is misleading and leads to bugs if you try to use it, and we have at least two ways to improve the situation:
1. Improve this API by using separate blockers, so it really can be used in different places.
2. Mark this API as a special case (by renaming related functions) and add a comment about found issues to make sure that developers will not use it if a need to pause/resume sounds appears somewhere.

psi29a:
> Yes, exactly what I was talking about - I see no point to implement such thing.
> By design, enums are needed to re-use their entries, but our case is pretty opposite - we want to DO NOT re-use them. Also enum with one entry is a really strange thing.

Depending on where we go with this in the future, it might not be just one. It just so happens that there is only one needed right now.

akortunov:
> stability/readability

Stability readability will be pretty the same in both cases.

akortunov:
Well, if you are not going to merge my current approach, I have no choice excepts if using enums, but I still strongly to not like it since it has a lot of drawbacks:

1. Every time when you change an enum, you need to rebuild everything which uses sound system. It is a most of files in the /apps/openmw.
2. Arrays are less informative in debugger than maps.
3. Arrays here look weird when there are many entries in enum (and we got a large array, which is always mostly filled by zeroes) and where are too low amount of entries in enum (and we get an array from one element).

Actually, an only advantage of array is a performance, which is totally irrelevant here.

apps/openmw/mwbase/soundmanager.hpp
```diff
@@ -14,6 +14,14 @@ namespace MWWorld
 
 namespace MWSound
 {
+    // Each entry excepts of MaxCount should be used only in one place
+    enum BlockerType
+    {
+        VideoPlayback = 0,
+
+        MaxCount = 1
```
kcat, line 22: You don't need to explicitly declare values for enums, and in this case it's not recommended to. They always start at 0 and each entry is one more than the previous, so `MaxCount` is automatically correct.

kcat:
> Actually, an only advantage of array is a performance, which is totally irrelevant here.

`unordered_map` can still be used with an enum key, if you're really set on maps. But my main issue is the use of strings which doesn't have any compile-time guarantee for correctness.

apps/openmw/mwbase/soundmanager.hpp
akortunov: Can you guarantee that enums in C++ use indices on all platforms and old compilers?
It will be not a good thing to break, for example, Clang or Android builds because they use powers of two instead?

kcat: [Positive](https://en.cppreference.com/w/cpp/language/enum):
"If the first enumerator does not have an initializer, the associated value is zero. For any other enumerator whose definition does not have an initializer, the associated value is the value of the previous enumerator plus one."

psi29a:
@akortunov would you like to see this in 0.46 ?

akortunov:
>  would you like to see this in 0.46 ?

Yes, it would be nice.

akortunov:
> If it doesn't, the game still goes quiet (1.16 was released almost 6 years ago, and I'd expect many other issues if using such an outdated version).

Well, at leats for one user on MacOS thiis feature did not work because of old OpenAL-Soft version. Intereseting that both functions (alcDevicePauseSOFT and alListenerf) did not work.

kcat:
It was Apple's OpenAL ("the framework provided by xcode"), not OpenAL Soft (which isn't built as a framework). I can't see why `alListenerf` wouldn't work, but I certainly wouldn't rule out a bug in Apple's OpenAL implementation.

psi29a:
Oh man, yeah I remember that with my WildMIDI project... had to brew install openal-soft because apple's version lacked features.

PR #2751, opened 2020-03-31
[WIP] In-game menu to tweak shadows

akortunov:
Mostly replicates menu from launcher. 

Known issues:
1. Shadows do not work on Android, so I have to disable shadows menu on that platform.
2. Attempts to disable shadows in interiors lead to graphics artifacts (because of `mShadowTechnique->disableShadows();` call).
3. Other settings require horrible hacks to tweak them in the runtime. Unfortunately, our shadowing system does not provide an API for runtime tweaks.

### DIFF
--- a/apps/openmw/mwgui/settingswindow.cpp
+++ b/apps/openmw/mwgui/settingswindow.cpp
@@ -6,6 +6,7 @@
 #include <MyGUI_ScrollView.h>
 #include <MyGUI_Gui.h>
 #include <MyGUI_TabControl.h>
+#include <MyGUI_TabItem.h>
 
 #include <SDL_video.h>
 
@@ -204,6 +205,7 @@ namespace MWGui
         getWidget(mControllerSwitch, "ControllerButton");
         getWidget(mWaterTextureSize, "WaterTextureSize");
         getWidget(mWaterReflectionDetail, "WaterReflectionDetail");
+        getWidget(mShadowTextureSize, "ShadowsTextureSize");
 
 #ifndef WIN32
         // hide gamma controls since it currently does not work under Linux
@@ -228,6 +230,7 @@ namespace MWGui
 
         mWaterTextureSize->eventComboChangePosition += MyGUI::newDelegate(this, &SettingsWindow::onWaterTextureSizeChanged);
         mWaterReflectionDetail->eventComboChangePosition += MyGUI::newDelegate(this, &SettingsWindow::onWaterReflectionDetailChanged);
+        mShadowTextureSize->eventComboChangePosition += MyGUI::newDelegate(this, &SettingsWindow::onShadowTextureSizeChanged);
 
         mKeyboardSwitch->eventMouseButtonClick += MyGUI::newDelegate(this, &SettingsWindow::onKeyboardSwitchClicked);
         mControllerSwitch->eventMouseButtonClick += MyGUI::newDelegate(this, &SettingsWindow::onControllerSwitchClicked);
@@ -261,21 +264,38 @@ namespace MWGui
         mTextureFilteringButton->setCaption(textureMipmappingToStr(tmip));
 
         int waterTextureSize = Settings::Manager::getInt("rtt size", "Water");
-        if (waterTextureSize >= 512)
-            mWaterTextureSize->setIndexSelected(0);
-        if (waterTextureSize >= 1024)
-            mWaterTextureSize->setIndexSelected(1);
         if (waterTextureSize >= 2048)
             mWaterTextureSize->setIndexSelected(2);
+        else if (waterTextureSize >= 1024)
+            mWaterTextureSize->setIndexSelected(1);
+        else if (waterTextureSize >= 512)
+            mWaterTextureSize->setIndexSelected(0);
 
         int waterReflectionDetail = Settings::Manager::getInt("reflection detail", "Water");
         waterReflectionDetail = std::min(4, std::max(0, waterReflectionDetail));
         mWaterReflectionDetail->setIndexSelected(waterReflectionDetail);
 
+        int shadowsTextureSize = Settings::Manager::getInt ("shadow map resolution", "Shadows");
+        if (shadowsTextureSize >= 4096)
+            mShadowTextureSize->setIndexSelected(3);
+        else if (shadowsTextureSize >= 2048)
+            mShadowTextureSize->setIndexSelected(2);
+        else if (shadowsTextureSize >= 1024)
+            mShadowTextureSize->setIndexSelected(1);
+        else if (shadowsTextureSize >= 512)
+            mShadowTextureSize->setIndexSelected(0);
+
         mWindowBorderButton->setEnabled(!Settings::Manager::getBool("fullscreen", "Video"));
 
         mKeyboardSwitch->setStateSelected(true);
         mControllerSwitch->setStateSelected(false);
+
+#ifdef ANDROID
+        // Shadows do not work on Android so far, so hide shadows tab
+        MyGUI::TabControl *tabControl;
+        getWidget(tabControl, "VideoTabs");
+        tabControl->removeItemAt(3);
+#endif
     }
 
     void SettingsWindow::onTabChanged(MyGUI::TabControl* /*_sender*/, size_t /*index*/)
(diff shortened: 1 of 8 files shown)


AnyOldName3:
The shadow system is *designed* to support runtime tweaks and will as soon as someone makes the UI, so I'd avoid using it not doing so right now as justification for any decisions. That said, this is one of the functions that would need to be called again when certain settings change, so this would only affect things a little bit.

I don't have any particular objections to this PR itself. As for it fixing Android, I'm struggling to think of how it could be causing issues without it just happening to expose a GL4ES bug. If it clashed with texture slots, we'd be seeing *some* kind of issue on desktops, too.

akortunov:
> The shadow system is designed to support runtime tweaks and will as soon as someone makes the UI

Can you document an API for such tweaks? When I tried to implement an in-game UI for shadows, I encountered a pretty the same issue - a shadow system is initialized only once during game start, and I did not find any good way to set different parameters for it during runtime.

AnyOldName3:
The vast majority of settings should Just Work™ if you call `ShadowManager::setupShadowSettings()` again after changing them. It might be reasonable to add individual methods to change them one at a time when adding a GUI, though. Some other settings will require calling `ShadowManager::disableShadowsForStateSet` for the relevant statesets again (specifically changing the number of shadow maps, and, after this PR, turning them on or off) and some different ones will require calling `ShadowManager::getShadowDefines()` and giving the results to the shader manager as we do here: https://github.com/OpenMW/openmw/blob/e80fbf4786d555a448e614f9ba79a4ec4c306297/apps/openmw/mwrender/renderingmanager.cpp#L251-L263. I even edited that shader manager method as part of the shadow PR so it wouldn't crash if we changed things at runtime.

There might be a *couple* of settings that don't change super gracefully, and the shadow map resolution is the most likely candidate. I know we generate any missing shadow maps and debug hud entries we encounter, so it's just a case of whether we verify the resolution is still good for the existing ones.

akortunov:
> The vast majority of settings should Just Work™ if you call shadowManager::setupShadowSettings()

Actually, only `enable shadows` works without additional efforts. It is what I would call the "our shadowing system does not support runtime tweaks anyway".

akortunov:
I tried to use something like this (I had to move masks setup to the shadowing system itself):

```
            else if (it->first == "Shadows")
            {
                bool forceShaders = Settings::Manager::getBool("radial fog", "Shaders") || Settings::Manager::getBool("force shaders", "Shaders") || Settings::Manager::getBool("enable shadows", "Shadows");
                mResourceSystem->getSceneManager()->setForceShaders(forceShaders);
                mShadowManager->setupShadowSettings();
                Shader::ShaderManager::DefineMap shadowDefines = mShadowManager->getShadowDefines();
                Shader::ShaderManager::DefineMap globalDefines = mResourceSystem->getSceneManager()->getShaderManager().getGlobalDefines();

                for (auto itr = shadowDefines.begin(); itr != shadowDefines.end(); itr++)
                    globalDefines[itr->first] = itr->second;

                mResourceSystem->getSceneManager()->getShaderManager().setGlobalDefines(globalDefines);

                if (mSky->isEnabled())
                    mShadowManager->enableOutdoorMode();
                else
                    mShadowManager->enableIndoorMode();
            }
```
Unfortunately, things that affect shaders still do not work (shadow map resolution, fadeout settings, etc.). Also indoor shadows switching is buggy.

AnyOldName3:
This is the first I'm hearing about any of this. I *did* actually test some of it myself when I was debugging things as it was faster to alter the shadow settings object than restarting the engine, and several prototype versions of `setGlobalDefines` were crash, so I had to test it until it wasn't. Maybe other people have broken things since then, or maybe it only works during certain phases of rendering the frame, or maybe there was threading weirdness.

I know for a fact that the indoor shadow switching works as we do that every time you go through a door, so someone would have reported it by now.

akortunov:
> Maybe other people have broken things since then, or maybe it only works during certain phases of rendering the frame, or maybe there was threading weirdness.

Well, you pass shadow masks to the constructor since [your original PR](https://github.com/OpenMW/openmw/pull/1547).
Also it has this code:

```
 // It is unnecessary to stop/start the viewer as no frames are being rendered yet.
        mResourceSystem->getSceneManager()->getShaderManager().setGlobalDefines(globalDefines);
```

So I suppose that you were aware of some kind of bugs.

AnyOldName3:
> Well, you pass shadow masks to the constructor since your original PR.

The shadow manager class isn't 100% ready for things just yet, but the actual complicated code in the shadow technique is ready. Just add a getter and setter to the manager and call the right indoor or outdoor enabler.

> ```
> // It is unnecessary to stop/start the viewer as no frames are being rendered yet.
> mResourceSystem->getSceneManager()->getShaderManager().setGlobalDefines(globalDefines);
> ```
> So I suppose that you were aware of some kind of bugs.

This is explained in the comment before `setGlobalDefines` - once rendering has started, you need to call `viewer->stopThreading()` first so that you don't accidentally start using the new defines for the previous frame.

akortunov:
> `viewer->stopThreading()`

I can not call it - there are sliders, which send events every frame whille player chenges their value.

AnyOldName3:
All of our settings that would benefit from sliders are handled CPU-side or by uniforms, so you don't need to update the global shader defines for those. There are some that would be checkboxes or a dropdown, and it's completely reasonable for those to have an asterisk added to their label and only change when Okay or Apply gets clicked, or maybe even when the checkbox is toggled. There are other settings we have that aren't spammed a bazillion times when the settings dialogue is open.

akortunov:
All right, I uploaded a code for runtime shadows configuring. It is possible to change masks and shadowmap resolution.

Attempts to disable shadows in interiors still lead to graphics artifacts:

Screenshot_20200402_161019

Other settings mostly require additonal hacks, so I see no point to handle them until shadow system exposes an API to configure it at runtime.

AnyOldName3:
What you're calling hacks *is* the API.

akortunov:
> What you're calling hacks is the API.

*sigh*. A person who develops UI windows, in a general case is not very familiar with low-level details of graphics system implementation. So when he needs to change the `number of shadow maps` setting, for example, he expects to do something like this:
```
for (Settings::CategorySettingVector::const_iterator it = changed.begin(); it != changed.end(); ++it)
{
   (it->first == "Shadows" && it->second == "number of shadow maps")
   {
       int numOfShadowMaps = Settings::Manager::getInt("number of shadow maps", "Shadows")
       mShadowManager->setNumOfShadowMaps(numOfShadowMaps);
   }
}
```

When he instead needs to manually reload statesets for sky, water and minimap, update shader defines, stop/resume rendering thread, reload shaders for everything, etc, it is not right.

By idea, such things should be incapsulated to the shadow system itself, if it is claimed as configurable during runtime.

apps/openmw/mwgui/settingswindow.cpp
```diff
@@ -272,10 +275,27 @@ namespace MWGui
...
 
+        int shadowsTextureSize = Settings::Manager::getInt ("shadow map resolution", "Shadows");
+        if (shadowsTextureSize >= 512)
+            mShadowTextureSize->setIndexSelected(0);
+        if (shadowsTextureSize >= 1024)
+            mShadowTextureSize->setIndexSelected(1);
+        if (shadowsTextureSize >= 2048)
+            mShadowTextureSize->setIndexSelected(2);
+        if (shadowsTextureSize >= 4096)
+            mShadowTextureSize->setIndexSelected(3);
```
elsid, line 286: Not sure is it ok to call setIndexSelected multiple times in a row, but still this looks odd. You can swap the order and then it will be only one call.
```suggestion
        if (shadowsTextureSize >= 4096)
            mShadowTextureSize->setIndexSelected(3);
        else if (shadowsTextureSize >= 2048)
            mShadowTextureSize->setIndexSelected(2);
        else if (shadowsTextureSize >= 1024)
            mShadowTextureSize->setIndexSelected(1);
        else if (shadowsTextureSize >= 512)
            mShadowTextureSize->setIndexSelected(0);
```

AnyOldName3:
I was told by Scrawl to not put the shadow manager in charge of stopping threading as it's the rendering manager's job to manage the global shader defines. Anything *other* than the global shader defines just needs a getter and setter adding to the shadow manager and potentially to call one of the methods that the shadow manager already has.

AnyOldName3:
My impression was that for most settings, they'd be read directly from the settings manager after any of them were updated, and only the expensive ones would be done differently.

apps/openmw/mwgui/settingswindow.cpp
akortunov: I just used a code for water RTT resolution as an example. Both should be fixed now.

akortunov:
Closing this PR since new way to handle node masks is horrible and using it in this PR will make that way even more horrible.

AnyOldName3:
You're already using the horrible new way of handling nodemasks in this PR. 😜

PR #2752, opened 2020-04-01
Reset errorhandler context

Assumeru:
Compiling scripts can result in errors/warnings being logged with the wrong script name, leading to cases where the log says "Warning: script_a[local variables]..." while the actual error is in script_b.

### DIFF
--- a/apps/openmw/mwscript/scriptmanagerimp.cpp
+++ b/apps/openmw/mwscript/scriptmanagerimp.cpp
@@ -171,7 +171,7 @@ namespace MWScript
         {
             Compiler::Locals locals;
 
-            Compiler::ContextRestore restore = mErrorHandler.setContext(name2 + "[local variables]", true);
+            const Compiler::ContextRestore&& restore = mErrorHandler.setContext(name2 + "[local variables]", true);
 
             std::istringstream stream (script->mScriptText);
             Compiler::QuickFileParser parser (mErrorHandler, mCompilerContext, locals);
--- a/apps/openmw/mwscript/scriptmanagerimp.cpp
+++ b/apps/openmw/mwscript/scriptmanagerimp.cpp
@@ -171,7 +171,7 @@ namespace MWScript
         {
             Compiler::Locals locals;
 
-            const Compiler::ContextRestore&& restore = mErrorHandler.setContext(name2 + "[local variables]", true);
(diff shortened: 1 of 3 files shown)


zinnschlag:
This change is problematic. First, you should not expose an internal variable like that. Breaks encapsulation. Second, the change you made to the getLocals function is not exception safe. If there is an exception between setting the context variable and resetting it later, the error handler will be left in an incorrect state.

I suggest looking at the class ErrorDowngrade in errorhandler.hpp for an example on how to handle this operation in an exception-safe way.

Assumeru:
I'm starting to wonder if removing the setContext call isn't the thing to do. Removing the End statement from the referenced script doesn't even trigger a message.

zinnschlag:
> I'm starting to wonder if removing the setContext call isn't the thing to do. Removing the End statement from the referenced script doesn't even trigger a message.

Could you elaborate more on what you mean by that?

zinnschlag:
Looked at the code and had a laugh. That is the most roundabout method of adding exception-safety I have ever seen. I guess it works and it is kinda ingenious, but I suggest to simplify it and get it in line with how this is normally done, so not to confuse other coders who read the code and are familiar with the pattern. That is if we are going to keep the feature at all, see comment above.

A more standard way to implement this feature would be to leave the setContext function alone. Make a OverrideContext class (similar to your ContextRestore class), that changes the context in the constructor and changes it back in the destructor (don't forget to mark copy constructor and assignment operator as deleted).

Assumeru:
> 
> Could you elaborate more on what you mean by that?

I've yet to find a way to get it to emit a warning/error for the replaced context (ie "x[local variables]"). The QuickFileParser just ignores everything thrown at it and even introducing an unexpected EOF didn't log anything in this part of the code. (It gets logged when the broken script in question gets parsed, of course.)

zinnschlag:
Yes, that is the purpose of the QuickFileParser. It just extracts the local variables, ignoring everything else (where possible). It should be able to produce errors when there are problems with local variables.

Assumeru:
It ignored `short a-b` and a missing `End`, but you're right. I finally got it to do something with `short` (no name).

zinnschlag:
Looks ready for a merge.

btw. you don't get an error on short a-b because the parser is extremely error tolerant. It has to because otherwise even half of the scripts in Morrowind.esm wouldn't compile.

PR #2753, opened 2020-04-02
These files aren't in the repo anymore

hristoast:


akortunov:
Thanks. I thought I wiped out all references to them.

PR #2757, opened 2020-04-04
Fix ESM::Land oversights

Capostrophic:
1. Don't operate on a null land data pointer, it'll lead to null pointer exception if there's no land data (noticed by akortunov). When there's no land data, the WNAM subrecord won't be saved (there doesn't seem to be a point to do that).
2. mLandData is set to nullptr instead of 0 where applicable.
3. mWnam is properly cleared when necessary just to be sure it's properly initialized.
4. std::fill is used where applicable.

### DIFF
--- a/components/esm/loadland.cpp
+++ b/components/esm/loadland.cpp
@@ -75,6 +75,8 @@ namespace ESM
         mContext = esm.getContext();
 
         mLandData = nullptr;
+        for (int i = 0; i < LAND_GLOBAL_MAP_LOD_SIZE; ++i)
+            mWnam[i] = 0;
 
         // Skip the land data here. Load it when the cell is loaded.
         while (esm.hasMoreSubs())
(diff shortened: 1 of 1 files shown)


components/esm/loadland.cpp
```diff
@@ -75,6 +75,8 @@ namespace ESM
         mContext = esm.getContext();
 
         mLandData = nullptr;
+        for (int i = 0; i < LAND_GLOBAL_MAP_LOD_SIZE; ++i)
+            mWnam[i] = 0;
```
elsid, line 79: General suggestion to work with containers: use algorithms when possible
```suggestion
        std::fill(std::begin(mWnam), std::end(mWnam), 0);
```
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es1-prefer-the-standard-library-to-other-libraries-and-to-handcrafted-code

components/esm/loadland.cpp
```diff
@@ -197,7 +194,7 @@ namespace ESM
         mPlugin = 0;
 
         for (int i = 0; i < LAND_GLOBAL_MAP_LOD_SIZE; ++i)
-            mWnam[0] = 0;
+            mWnam[i] = 0;
```
elsid, line 197: Very nice example why it's better to use standard algorithms.

components/esm/loadland.cpp
```diff
@@ -75,8 +75,7 @@ namespace ESM
         mContext = esm.getContext();
 
         mLandData = nullptr;
-        for (int i = 0; i < LAND_GLOBAL_MAP_LOD_SIZE; ++i)
-            mWnam[i] = 0;
+        std::fill(std::begin(mWnam), std::end(mWnam), 0);
```
psi29a, line 78: Much better :D

PR #2759, opened 2020-04-04
Don't use up light duration if the held light is hidden (bug #5352)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/-/issues/5352)

Use mShowCarriedLeft value from the player's animation object. This allows the duration not to decrease if the character is currently in a "two-handed" attack ready state (e.g. hand-to-hand attack stance, two-handed weapon attack stance, spellcasting stance).

### DIFF
--- a/apps/openmw/mwrender/creatureanimation.hpp
+++ b/apps/openmw/mwrender/creatureanimation.hpp
@@ -31,6 +31,8 @@ namespace MWRender
         virtual void equipmentChanged() { updateParts(); }
 
         virtual void showWeapons(bool showWeapon);
+
+        virtual bool getCarriedLeftShown() const { return mShowCarriedLeft; }
         virtual void showCarriedLeft(bool show);
 
         void updateParts();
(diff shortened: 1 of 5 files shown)


apps/openmw/mwrender/creatureanimation.hpp
```diff
@@ -24,13 +24,18 @@ namespace MWRender
     // Most creatures don't need weapons/shields, so this will save some memory.
     class CreatureWeaponAnimation : public ActorAnimation, public WeaponAnimation, public MWWorld::InventoryStoreListener
     {
+        bool mShowWeapons;
+        bool mShowCarriedLeft;
```
akortunov, line 28: I suppose it is not a right place for these variables. If you want to make them protected, move them to appropriate section.

Capostrophic: I want to leave them private but allow getCarriedLeftShown to work with them.

akortunov: > I want to leave them private

They are already private, so your change here should be just a no-op.

Capostrophic: I thought they'd be unreachable. Apparently not. Hmph, okay.

PR #2761, opened 2020-04-05
Check if the weapon is attached in GetWeaponDrawn (bug #4816)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/-/issues/4816)

Check if the equipped weapon is attached if there's one at all.
It seems to resolve the issue, but the behavior of the test case (the results the script returns) is heavily dependent on the frame rate and the chosen actor.

### DIFF
--- a/apps/openmw/mwscript/miscextensions.cpp
+++ b/apps/openmw/mwscript/miscextensions.cpp
@@ -35,6 +35,8 @@
 #include "../mwmechanics/spellcasting.hpp"
 #include "../mwmechanics/actorutil.hpp"
 
+#include "../mwrender/animation.hpp"
+
 #include "interpretercontext.hpp"
 #include "ref.hpp"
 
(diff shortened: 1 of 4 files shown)


apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -664,8 +666,32 @@ namespace MWScript
...
                     MWWorld::Ptr ptr = R()(runtime);
 
-                    runtime.push((ptr.getClass().hasInventoryStore(ptr) || ptr.getClass().isBipedal(ptr)) &&
-                                ptr.getClass().getCreatureStats (ptr).getDrawState () == MWMechanics::DrawState_Weapon);
+                    if (!ptr.getClass().hasInventoryStore(ptr) && !ptr.getClass().isBipedal(ptr))
+                    {
+                        runtime.push(0);
+                        return;
+                    }
+                    if (ptr.getClass().getCreatureStats(ptr).getDrawState() != MWMechanics::DrawState_Weapon)
```
akortunov, line 674: Not sure if this check should be here.

akortunov:
It seems to resolve the issue, but the behavior of the test case (the results the script returns) is heavily dependent on the frame rate and the chosen actor.

That's why I did not touch this bug - there is no clean task description to kane proper tests and determine if the bug is fixed or it is not.

Capostrophic:
I guess I'll close this for now.

PR #2763, opened 2020-04-06
Fix int/float conversions in the CSVRender::BrushDraw

akortunov:
Summary of changes:
1. Do not lose remainder during `mLandSizeFactor` calculation.
2. Explicitly cast a float result to integer during `resolution` value calculation.

Note that I do not know if the formula from step 2 really should use 2 diamaters instead of 2 radiuses.

### DIFF
--- a/apps/opencs/view/render/brushdraw.cpp
+++ b/apps/opencs/view/render/brushdraw.cpp
@@ -21,8 +21,9 @@ CSVRender::BrushDraw::BrushDraw(osg::ref_ptr<osg::Group> parentNode, bool textur
     mBrushDrawNode->addChild(mGeometry);
     mParentNode->addChild(mBrushDrawNode);
     if (mTextureMode)
-        mLandSizeFactor = ESM::Land::REAL_SIZE / ESM::Land::LAND_TEXTURE_SIZE;
-    else mLandSizeFactor = ESM::Land::REAL_SIZE / ESM::Land::LAND_SIZE;
+        mLandSizeFactor = static_cast<float>(ESM::Land::REAL_SIZE) / static_cast<float>(ESM::Land::LAND_TEXTURE_SIZE);
+    else
+        mLandSizeFactor = static_cast<float>(ESM::Land::REAL_SIZE) / static_cast<float>(ESM::Land::LAND_SIZE);
 }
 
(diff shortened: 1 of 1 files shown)


unelsson:
Hmm... `ESM::Land::REAL_SIZE` and `ESM::Land::LAND_TEXTURE_SIZE` are square numbers, so this conversion should be okay as long as those values aren't changed, but this obviously makes it work even if those are changed.. But how about `ESM::Land::LAND_SIZE` which is 65, but one row/column is shared. Did you test this one? It might be a proper fix, but my intuition makes me wonder if it should in fact be float calculation, but with 65 - 1, but I'm not sure, I'd have to think this through properly.

I also do not understand why the latter change with resolution-calculation is better. Something to do with different compilers or systems how they handle the conversion, or just another way to tell other programmers that we are converting that to int? I'll also double check whether it should be 2 diameters or radiuses.

akortunov:
> I also do not understand why the latter change with resolution-calculation is better. 

Some compilers may generate warnings if you assign float value to integer variable, so we usually tell compiler that the assignment is intended (by using type casts).

unelsson:
The correct value for mLandSizeFactor in height editing mode is indeed ESM::Land::LAND_SIZE - 1, and converting to float is required.
```    if (mTextureMode)
        mLandSizeFactor = float(ESM::Land::REAL_SIZE) / float(ESM::Land::LAND_TEXTURE_SIZE);
    else mLandSizeFactor = float(ESM::Land::REAL_SIZE) / float(ESM::Land::LAND_SIZE - 1);```

However, is there are reason why you are suggesting C-style cast instead of static_cast?

unelsson:
In the current code, `resolution` has to be the `diameter` doubled, as the `resAdjustedLandSizeFactor` is halved. Lines of the outline are drawn therefore to the length of half of the diameter of a vertex, so the overall resolution must be the double of the amount of vertexes. Half a vertex resolution is required, because the outline is drawn from the middle of a square, and half of a vertex is a distance to the next point where the land slope is changed. Therefore the current math is as intended, but there might be options for later optimizations and clearing it up.

I tested the changes, and works as intended. This PR improves the code by adding a type cast.

akortunov:
Changed code to use static casts.

PR #2764, opened 2020-04-07
Minor fix to OpenMW-CS tool outline

unelsson:
Adds `-1`


akortunov:
Can you make commit name a more descriptive, please?

elsid:
I'd expect some more explanation. Why `-1` is a fix, why not `+1`? What is broken?

Capostrophic:
Quoting unelsson:
>About the OpenMW-CS PR above, should I make another PR for having ESM::Land::LAND_SIZE - 1 instead of just ESM::Land::LAND_SIZE? It's really a minor thing, but the tool outline has now slightly wrong scale... The difference is something like 1/65 vs 1/64, so hardly visible, but still.

unelsson:
Yeah, it's -1 because there are 65 vertexes, of which one is shared with the next cell, therefore there are 64 bits between the vertexes.

psi29a:
Is this for 0.46 ?

Capostrophic:
Yeah it was.

psi29a:
cherry-picked; thanks

PR #2765, opened 2020-04-08
Increase water-camera Z axis fudge factor (bug #5351)

Fanael:
On ATI/AMD hardware, the current fudge factor is way too small, which results in serious artefacting when the camera gets too close to the water surface.

The value of 5 way chosen because smaller values, like 1 or 2, still produce visible artifacting at higher resolutions.

---

[Bug report](https://gitlab.com/OpenMW/openmw/-/issues/5351).

Be warned that I only tested this patch on Polaris (RX 550) on Linux, it's untested on any other GPU+OS combination.

### DIFF
--- a/apps/openmw/mwrender/water.cpp
+++ b/apps/openmw/mwrender/water.cpp
@@ -188,7 +188,7 @@ public:
     {
         osgUtil::CullVisitor* cv = static_cast<osgUtil::CullVisitor*>(nv);
 
-        const float fudge = 0.2;
+        const float fudge = 5;
         if (std::abs(cv->getEyeLocal().z()) < fudge)
         {
             float diff = fudge - cv->getEyeLocal().z();


AnyOldName3:
If I'm remembering the scale correctly, that's like 12 cm of fudge. Does it definitely not cause other artefacts?

Fanael:
None that I have noticed, but that's just me and on a single system, so that's not a particularly great sample size. Do you know what potential artefacting issues I should look for?

AnyOldName3:
I'd guess things like the reflections and refractions not matching up with the rest of the world, especially when the camera is close to the water.

Fanael:
To make it clear: I view this as a temporary workaround, not a permanent solution. Properly solving the problem *requires* a better water rending system.

I do not see any significant additional artifacting when comparing master to this PR. There is some slight distortion which is expected from changing the fudge factor by a factor of 25, and a bit of the original problem still remains — I was afraid of changing the fudge by *too much* — but the problem is certainly alleviated by *a lot*, which is worth it in my opinion, at least as a temporary solution until better water rendering that doesn't rely on huge quads is there, and making invasive changes to water rendering during the RC period seems unwise.

(As an alternative, it might be reasonable to change the fudging factor only for AMD cards by checking the OpenGL driver vendor?)

I've attached some screenshots for comparison: the first of each set is master on Nvidia, the second is the patch on AMD. Do bear in mind this is with the camera just above the water level, so this is the worst case scenario for AMD, and it's where we currently see an utter pixel hellscape. It looks better in the usual case where the player isn't actively trying to place the camera right at the water level.

Set 1:
1-master
1-patched

Set 2:
2-master
2-patched

jackleson:
here is an example on my r9 390 windows. nightly. https://www.youtube.com/watch?v=rP14G3Cjv94

psi29a:
I consider this something to be pulled into 0.46 branch, anyone disagree?

akortunov:
> anyone disagree?

Well, this PR does not really fix a bug, and it is unclear which side effects it has. I'd not hurry to merge it to 0.46.

psi29a:
It's to be understood that it's a work-around which is fine for now, it doesn't close the bug. The real fix, not for 0.46 of course, would be a water system re-write as suggested above.

AnyOldName3:
Does this affect the recent water regressions related to radial fog at all?

Capostrophic:
No.

Capostrophic:
Somehow this fixes [this](https://gitlab.com/OpenMW/openmw/-/issues/5365) regression for me on Nvidia, I guess because the depth is never small enough for the side effects of radializing show themselves. Which AMD users don't even get because of the artifacts this is supposed to fix... Win-win..?

Capostrophic:
Alas, it depends on the distance to the near clipping plane, so near clip set higher (to something like 3) makes the refraction issue reappear.

Which makes me question whether the AMD artifacts become more or less severe depending on the near clip distance.

Fanael:
Now that we're in 0.47 cycle, this dubious hack is unlikely to get anywhere because we can just implement a better water system instead. Closing.

PR #2766, opened 2020-04-10
Make extra sure groups with special behavior aren't optimized out

Capostrophic:
Since there are more static transforms now (for example, controller-less NiTriShapes), they could be merged with NiCollisionSwitches and make them lose their special node mask, which caused issues with GitD interior god rays. Setting data variance to dynamic makes sure NiCollisionSwitches aren't considered eligible for such a thing.

osg::LOD groups have their data variance set to dynamic too so as to prevent similar optimizer issues from appearing (e.g. if someone figures out how to handle transformations but forgets to change the data variance, or if RemoveRedundantNodesVisitor decides osg::LOD node should be purged when it really shouldn't).

Switch nodes are handled slightly differently, I didn't touch them.

This should go into 0.46.0 branch.

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -457,7 +457,7 @@ namespace NifOsg
             {
                 const Nif::NiLODNode* niLodNode = static_cast<const Nif::NiLODNode*>(nifNode);
                 node = handleLodNode(niLodNode);
-                dataVariance = osg::Object::STATIC;
+                dataVariance = osg::Object::DYNAMIC;
                 break;
             }
             case Nif::RC_NiSwitchNode:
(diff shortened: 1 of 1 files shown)


akortunov:
Are you sure that using dynamics here is a proper fix? Can we fix this issue on the optimizer's side?

Capostrophic:
> Are you sure that using dynamics here is a proper fix?

It's not going to mess up performance if you're about that. Data variance is only used as an optimizer hint, and CollisionSwitch groups are pretty clearly not static transforms that can be merged.

> Can we fix this issue on the optimizer's side?

I don't see a good way to do it.

AnyOldName3:
It's only dynamic drawables that stall the other threads, so Capo's right here.

psi29a:
cherry-picked into openmw-46

PR #2767, opened 2020-04-11
Make sure it's a crime to unlock owned doors

Capostrophic:
unlockAttempted must be called before the doors are actually unlocked because it's not a crime to unlock unlocked doors. Somehow in testing whether it was a crime to fail to unlock a door I overlooked checking the main situation where an attempt to unlock a door is successful. Ouch.

Should go into 0.46.0 branch as well.


psi29a:
cherry-picked into openmw-46

PR #2773, opened 2020-04-12
Only clear dialogue history if the dialogue window is closed (bug #5358)

Capostrophic:
The history is reset in a handful of ways the window is closed but it's not reset if the window's actor is changed without closing the window, which seems to resolve [the issue](https://gitlab.com/OpenMW/openmw/-/issues/5358).


Fanael:
This change stops the dialogue history from being cleared if the dialogue window was closed by, for example, using the mages guild guide for fast travel, or by pressing the escape key.

PR #2775, opened 2020-04-12
Only disarm traps with keys when the door/container is locked (bug #5370)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/-/issues/5370).

Removes the undesirable "disarm unlocked doors" behavior for keys.
Containers and doors behave the same way here in Morrowind in my testing.

The behavior was added in PR #995, but somehow it doesn't seem like it's present in the vanilla game (neither in my testing nor geist's). Maybe Allofich meant something different when he said "lock level 0" like locks that were locked with `Lock 0`?..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
     Bug #5364: Script fails/stops if trying to startscript an unknown script
     Bug #5367: Selecting a spell on an enchanted item per hotkey always plays the equip sound
     Bug #5369: Spawnpoint in the Grazelands doesn't produce oversized creatures
+    Bug #5370: Opening an unlocked but trapped door uses the key
     Feature #5362: Show the soul gems' trapped soul in count dialog
 
 0.46.0
--- a/apps/openmw/mwclass/container.cpp
+++ b/apps/openmw/mwclass/container.cpp
@@ -177,11 +177,10 @@ namespace MWClass
             }
         }
 
-        if ((isLocked || isTrapped) && hasKey)
+        if (isLocked && hasKey)
         {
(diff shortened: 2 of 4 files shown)


Allofich:
> Maybe Allofich meant something different when he said "lock level 0" like locks that were locked with Lock 0?

It was a long time ago but yes, I think I remember testing this by using "Lock 0" with the console, rather than "unlocking a door with other means and leaving the trap, then activating it (with the key in your inventory)" like the bug reporter said.

Capostrophic:
So yeah, `lock 0` doesn't behave the same way as `unlock` in vanilla, it sets the lock level to the largest positive value (which isn't displayed on the tooltip). That's the way it works in OpenMW as well (since 2018).

akortunov:
But what if the container was initially only trapped, but had an assigned key? I wonder how Morrowind handles this case.

Capostrophic:
> I wonder how Morrowind handles this case.

Oddly enough Morrowind behaves as if there's no trap at all and the door is simply opened as a normal unlocked door that wasn't trapped.

Capostrophic:
The latest change implements the "ignore trap" behavior noted above. It's one thing Morrowind does consistently. There's most likely something fishy about 0 lock level locks with keys assigned otherwise (as akortunov claims it's essential to get the unbreakable lock functionality in Arktwend to work) because the one situation I tested had the lock unlocked but it's not directly related to the issue at hand, so I didn't change it.

There's a feature in MCP that allows the scripter to check the cell reference's lock level instead of an actor's level using GetLevel that may help figure out the correct thing to do there, but that's a matter for another time.

Capostrophic:
Any objections?

psi29a:
I do not have any.

PR #2777, opened 2020-04-13
Make book layer have lower priority than the layer for other GUI windows

Capostrophic:
Should in theory solve [this](https://gitlab.com/OpenMW/openmw/-/issues/5349) painlessly. Books must have their own layer for scaling reasons, but there isn't a good reason to let them overlap all and any GUI window, and since books/scrolls/journal view is an exclusive GUI mode, it shouldn't be able to conflict with another GUI window. Except the free-floating console.

Apparently letting the console have its own layer with higher priority than both books and GUI windows still has its caveats, while this set up should avoid issues with it.

If it's merged it should go into 0.46 branch.


Capostrophic:
Oops. Mispush.

akortunov:
@psi29a, you can cherry-pick this PR to the 0.46 branch.

psi29a:
merged

PR #2781, opened 2020-04-16
Particle System shading and lighting (fixing bug#3676 and #4949)

mp3butcher:
This pr enables lighting of particle systems
it also fixes the alpha issue (https://gitlab.com/OpenMW/openmw/issues/3676) and https://gitlab.com/OpenMW/openmw/issues/4949

NB: 
- I would have let Ambient material on partsys but spec say alpha come from diffuse product.
https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glLightModel.xml

>             The alpha component of the resulting lighted color is set to the alpha value
>             of the material diffuse reflectance.

I only force particles normal to (0.3,0.3,0.3) following aon3 DX trace...

fountain
lighsmok

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 0.47.0
 ------
 
+    Bug #3676: NiParticleColorModifier isn't applied properly
+    Bug #4949: Incorrect particle lighting when force shaders = true
     Bug #5358: ForceGreeting always resets the dialogue window completely
     Bug #5363: Enchantment autocalc not always 0/1
     Bug #5364: Script fails/stops if trying to startscript an unknown script
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -1809,14 +1809,6 @@ namespace NifOsg
             if (specFlags == 0)
                 mat->setSpecular(osg::Material::FRONT_AND_BACK, osg::Vec4f(0.f,0.f,0.f,0.f));
 
-            // Particles don't have normals, so can't be diffuse lit.
-            if (particleMaterial)
-            {
-                // NB ignoring diffuse.a()
-                mat->setDiffuse(osg::Material::FRONT_AND_BACK, osg::Vec4f(0,0,0,1));
-                mat->setColorMode(osg::Material::AMBIENT);
-            }
-
             if (lightmode == 0)
             {
                 osg::Vec4f diffuse = mat->getDiffuse(osg::Material::FRONT_AND_BACK);
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -1078,6 +1078,8 @@ namespace NifOsg
                 trans->addChild(toAttach);
                 parentNode->addChild(trans);
             }
+            // create partsys stateset in order to pass in ShaderVisitor like all other Drawables
+            partsys->getOrCreateStateSet();
         }
 
         void triCommonToGeometry(osg::Geometry *geometry, const std::vector<osg::Vec3f>& vertices, const std::vector<osg::Vec3f>& normals, const std::vector<std::vector<osg::Vec2f>>& uvlist, const std::vector<osg::Vec4f>& colors, const std::vector<unsigned int>& boundTextures, const std::string& name)
--- a/components/nifosg/particle.cpp
+++ b/components/nifosg/particle.cpp
@@ -159,8 +159,11 @@ void ParticleColorAffector::operate(osgParticle::Particle* particle, double /* d
 {
     float time = static_cast<float>(particle->getAge()/particle->getLifeTime());
     osg::Vec4f color = mData.interpKey(time);
+    float alpha = color.a();
+    color.a() = 1.0f;
 
     particle->setColorRange(osgParticle::rangev4(color, color));
+    particle->setAlphaRange(osgParticle::rangef(alpha, alpha));
 }
 
 GravityAffector::GravityAffector(const Nif::NiGravity *gravity)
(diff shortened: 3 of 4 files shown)


mp3butcher:
lot more clean than https://github.com/OpenMW/openmw/pull/2441

AnyOldName3:
The APITrace from the vanilla engine I was messing around with the other day had the particle normals set to (0.3, 0.3, 0.3) in model-space. We might as well replicate that exactly seeing as we know the original value.

AnyOldName3:
Also, I don't understand why you're setting the normal in a draw callback instead of overall.

mp3butcher:
> Also, I don't understand why you're setting the normal in a draw callback instead of overall.
perhaps some rest of the point_sprite implementation

mp3butcher:
> > Also, I don't understand why you're setting the normal in a draw callback instead of overall.
> 
> perhaps some rest of the point_sprite implementation

arf no it's because particle system are drawable not geometry

mp3butcher:
> The APITrace from the vanilla engine I was messing around with the other day had the particle normals set to (0.3, 0.3, 0.3) in model-space. We might as well replicate that exactly seeing as we know the original value.

So you'll say   normal = invmodelview * osg::Vec3(0.3,0.3,0.3); ?

mp3butcher:
@AnyOldName3 in theory a particle have an omnidirectional normal, so tell me what it should be in practice

AnyOldName3:
If I'm understanding Direct3D8 correctly, it would be `gl_NormalMatrix * vec3(0.3, 0.3, 0.3)`

AnyOldName3:
It should be, provided the normal matrix is set up in the usual way.

mp3butcher:
> It should be, provided the normal matrix is set up in the usual way.

done, i would add additional shading but keep it for further PR

components/nifosg/nifloader.cpp
```diff
@@ -8,6 +8,7 @@
 #include <osg/Switch>
 #include <osg/TexGen>
 #include <osg/ValueObject>
+#include <osg/Version>
```
AnyOldName3, line 11: This include isn't used in this file

AnyOldName3:
LGTM once the unwanted include is gone.

components/nifosg/particle.cpp
```diff
@@ -159,8 +183,11 @@ void ParticleColorAffector::operate(osgParticle::Particle* particle, double /* d
 {
     float time = static_cast<float>(particle->getAge()/particle->getLifeTime());
     osg::Vec4f color = mData.interpKey(time);
+    float alpha = color.a();
+    color.a() = 1.0f;
 
     particle->setColorRange(osgParticle::rangev4(color, color));
+    particle->setAlphaRange(osgParticle::rangef(alpha, alpha));
```
Capostrophic, line 190: AFAIK ignoring alpha component of the used color isn't a part of upstream osgParticle (and somehow I can't find your PR where you tried to change that). While the way this is done won't break anything these few changes above don't seem to be necessary at the moment and aren't relevant to the issue.

components/nifosg/nifloader.cpp
mp3butcher: oups

components/nifosg/particle.cpp
mp3butcher: this is mandatory because  color and alpha are separated in osg, i don't ignore alpha just put it in the right channel

components/nifosg/particle.cpp
```diff
@@ -19,12 +20,16 @@ ParticleSystem::ParticleSystem()
...
 {
+    mNormalArray = new osg::Vec3Array(1); mNormalArray->setBinding(osg::Array::BIND_OVERALL);
+    (*mNormalArray.get())[0] = osg::Vec3(0.3, 0.3, 0.3);
 }
 
 ParticleSystem::ParticleSystem(const ParticleSystem &copy, const osg::CopyOp &copyop)
     : osgParticle::ParticleSystem(copy, copyop)
     , mQuota(copy.mQuota)
 {
+    mNormalArray = new osg::Vec3Array(1); mNormalArray->setBinding(osg::Array::BIND_OVERALL);
```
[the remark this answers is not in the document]
mp3butcher: ok

CHANGELOG.md
```diff
@@ -223,6 +223,8 @@
     Bug #5345: Dopey Necromancy does not work due to a missing quote
     Bug #5350: An attempt to launch magic bolt causes "AL error invalid value" error
     Bug #5352: Light source items' duration is decremented while they aren't visible
+    Bug #3676: NiParticleColorModifier isn't applied properly
+    Bug #4949: Incorrect particle lighting when force shaders = true
```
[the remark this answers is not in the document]
mp3butcher: haven't figured Bug was not fix chronologically ordered

CHANGELOG.md
```diff
@@ -118,6 +119,7 @@
     Bug #4945: Poor random magic magnitude distribution
     Bug #4947: Player character doesn't use lip animation
     Bug #4948: Footstep sounds while levitating on ground level
+    Bug #4949: Incorrect particle lighting when force shaders = true
```
Capostrophic, line 122: Again, these changes won't be a part of 0.46.0. Put them to 0.47.0 milestone.

mp3butcher: haven't understood...

psi29a:
Pinging @mp3butcher as it seems this PR causes this issue: https://gitlab.com/OpenMW/openmw/-/issues/5455

Any ideas?

Capostrophic:
Keep in mind the PR right above (#2823) was intended to resolve the regression but wasn't finished.

PR #2783, opened 2020-04-17
Refactor InputManager

akortunov:
Implements [task #5333](https://gitlab.com/OpenMW/openmw/-/issues/5333).

The main idea - instead of having a "god class" with 2000+ lines of code and dozens of fields, move the low-level stuff to separate classes and delegate execution to them.
As a result, InputManager becomes just a wrapper for a set of classes, which handle different aspects of input, and it has less than 300 lines of code now.

What new classes do:
1. ActionManager - contains handlers for in-game actions for hotkeys
2. BindingsManager - isolates OICS handling from the rest of the engine
3. ControllerManager - handles gamepad-related features 
4. ControlSwitch - a container to keep control switches and their logic
5. MouseManager - handles mouse-related features 
6. KeyboardManager - handles keyboard-related features 
7. SensorManager - handles sensor-related features (a gyroscope input for now)

Any feedback about design or bugs would be welcome.


psi29a:
Kicking msvc2019

psi29a:
Oef.. msvc 2019 really doesn't like this for some reason.

apps/openmw/mwinput/sdlmappings.cpp
```diff
@@ -0,0 +1,218 @@
...
+        keyMap[SDLK_LCTRL] = MyGUI::KeyCode::LeftControl;
+        keyMap[SDLK_RCTRL] = MyGUI::KeyCode::RightControl;
+#endif
+    }
+
+    MyGUI::KeyCode sdlKeyToMyGUI(SDL_Keycode code)
+    {
+        static std::map<SDL_Keycode, MyGUI::KeyCode> keyMap;
+        if (keyMap.empty())
+            initKeyMap(keyMap);
```
elsid, line 209: Do you think it could be better to replace map by switch-case from the performance point of view? Or at least initialize the map by a single statement.

akortunov: Not sure if it is needed. Input system is not a bottleneck in OpenMW.

apps/openmw/mwinput/bindingsmanager.cpp
```diff
@@ -0,0 +1,738 @@
...
+        descriptions[A_QuickKey4] = "sQuick4Cmd";
+        descriptions[A_QuickKey5] = "sQuick5Cmd";
+        descriptions[A_QuickKey6] = "sQuick6Cmd";
+        descriptions[A_QuickKey7] = "sQuick7Cmd";
+        descriptions[A_QuickKey8] = "sQuick8Cmd";
+        descriptions[A_QuickKey9] = "sQuick9Cmd";
+        descriptions[A_QuickKey10] = "sQuick10Cmd";
+        descriptions[A_AlwaysRun] = "sAlways_Run";
+        descriptions[A_QuickSave] = "sQuickSaveCmd";
+        descriptions[A_QuickLoad] = "sQuickLoadCmd";
```
elsid, line 465: This also could be replaced by switch-case.

apps/openmw/mwinput/bindingsmanager.cpp
```diff
@@ -0,0 +1,738 @@
...
+
+        if (mInputBinder->getJoystickAxisBinding(c, sFakeDeviceId, ICS::Control::INCREASE) != ICS::InputControlSystem::UNASSIGNED)
+            return sdlControllerAxisToString(mInputBinder->getJoystickAxisBinding(c, sFakeDeviceId, ICS::Control::INCREASE));
+        else if (mInputBinder->getJoystickButtonBinding(c, sFakeDeviceId, ICS::Control::INCREASE) != ICS_MAX_DEVICE_BUTTONS)
+            return sdlControllerButtonToString(mInputBinder->getJoystickButtonBinding(c, sFakeDeviceId, ICS::Control::INCREASE));
+        else
+            return "#{sNone}";
+    }
+
+    std::vector<int> BindingsManager::getActionKeySorting()
```
Capostrophic, line 520: This vector and controller vector below can be initialized straightaway, and they only need to be initialized once. That should reduce some overhead and save some lines of code.

psi29a:
I like what has been done here; it makes sense under the SDL2 gyro umbrella.

PR #2784, opened 2020-04-18
Disable specular lighting for 4.0.0.2 and earlier files

Capostrophic:
Morrowind doesn't actually support specular lighting, while scrawl made it purposefully enabled if there's a NiSpecularProperty with non-zero flag in the model. This sometimes makes the models made for Morrowind not look as intended. Since we strive for legacy compatibility when it comes to NIFs, specular lighting should be disabled in cases Morrowind has it disabled. However, NetImmerse and Gamebryo do support it, and seem to be designed to have it on by default unless NiSpecularProperty is added with its flags set to 0. So with this only 4.0.0.2 and earlier files — which Morrowind can read — will have specular lighting disabled as long as there isn't a specular map autodetected. 

This should improve compatibility with such mods as Better Meshes (don't know which one) or Westly's head replacer.

To underline: specular maps will be A-OK, it's the material specularity that will be murdered.

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -1705,7 +1705,8 @@ namespace NifOsg
         {
             osg::StateSet* stateset = node->getOrCreateStateSet();
 
-            int specFlags = 0; // Specular is disabled by default, even if there's a specular color in the NiMaterialProperty
+            // Specular lighting is enabled by default, but there's a quirk...
+            int specFlags = 1;
             osg::ref_ptr<osg::Material> mat (new osg::Material);
             mat->setColorMode(hasVertexColors ? osg::Material::AMBIENT_AND_DIFFUSE : osg::Material::OFF);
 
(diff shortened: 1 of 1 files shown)


psi29a:
So if someone wanted to take advantage of this feature, they would need to save their NIF to something > 4.0.0.2?

Capostrophic:
Technically NiSpecularProperty is not necessary at all, the modeller can just set specular color to black.

There's no support for newer files though. I don't know what could be documented. "Morrowind doesn't support specular lighting, so don't assume specular color in model properties does something if you export a Morrowind model".

akortunov:
> There's no support for newer files though.

Do you mean that there is no NiSpecularProperty in GameBryo?

Capostrophic:
No, Gamebryo does have it, but OpenMW can only load NIF 4.0.0.2 and 4.0.0.0 at the moment. That's what I meant.

psi29a:
> No, Gamebryo does have it, but OpenMW can only load NIF 4.0.0.2 and 4.0.0.0 at the moment. That's what I meant.

And in the future? (Forward thinking... )

Capostrophic:
As you can see newer files will still support material specularity with this.

PR #2785, opened 2020-04-18
Don't radialize underwater fog

Capostrophic:
There doesn't seem to be a good solution for the issue that doesn't depend on the near clipping distance at the moment.

Since it's not actually very noticeable that the underwater fog is not radial, it should be safe to disable radialization of water/refraction/whatever depth for the time being while allowing the release to carry on.


psi29a:
Is this meant for 0.46 then?

Capostrophic:
Yes.

PR #2786, opened 2020-04-19
[WIP] Editor: Update osgQt in extern

unelsson:
Very heavily WIP, but a PR for commenting and pointing out stuff early. Idea is to update osgQt in extern. It's used by the OpenMW-CS scene view. I understood it generated some errors, and it's probably a reason for a crash when exiting OpenMW-CS compiled with Qt5.

The PR is currently broken (probably should use osgQOpenGLWidget instead of Window), but it does compile. There might be additional unused clutter here too. I haven't really worked with CMakeLists or Qt MOC before.

What's done:
- scenewidget uses osgQOpenGLWindow by default instead of GraphicsWindowQt
- Added Qt5Gui to CMakeLists, because QOpenGLFunctions is part of it and osgQOpenGLWindow is derived from it.
- Added all cpp and header files from osgQt to extern
- Added namespace osgQt to osgQOpenGLWindow, osgQOpenGLWidget and OSGRenderer.
- Removed export macros from osgQt function declarations just to make stuff more clear for me, but maybe those are needed for some win compilers, I don't know (please help).
E.g. removed "OSGQOPENGL_EXPORT" present in https://github.com/openscenegraph/osgQt/blob/cf47af8c7608b1b2e0f6a70ba8b730db8a8cdff8/include/osgQOpenGL/osgQOpenGLWindow#L31 , but not in https://github.com/unelsson/openmw/blob/da16836ba3de1c5744b7b2e15437dd100386b612/extern/osgQt/osgQOpenGLWindow#L34

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,6 +160,7 @@ if (USE_QT)
         find_package(Qt5Core REQUIRED)
         find_package(Qt5Network REQUIRED)
         find_package(Qt5OpenGL REQUIRED)
+        find_package(Qt5Gui REQUIRED)
     # Instruct CMake to run moc automatically when needed.
     #set(CMAKE_AUTOMOC ON)
     endif()
--- a/apps/opencs/view/render/scenewidget.cpp
+++ b/apps/opencs/view/render/scenewidget.cpp
@@ -5,13 +5,17 @@
 #include <QTimer>
 #include <QLayout>
 
-#include <extern/osgQt/GraphicsWindowQt>
 #include <osg/GraphicsContext>
 #include <osgViewer/CompositeViewer>
 #include <osgViewer/ViewerEventHandlers>
 #include <osg/LightModel>
 #include <osg/Material>
 #include <osg/Version>
+#if OSG_VERSION_LESS_THAN(3,5,4)
+#include <extern/osgQt/GraphicsWindowQt>
+#else
+#include <extern/osgQt/osgQOpenGLWindow>
+#endif
 
 #include <components/debug/debuglog.hpp>
 #include <components/resource/scenemanager.hpp>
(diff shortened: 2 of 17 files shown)


akortunov:
Are there plans to continue development of this PR?

psi29a:
@unelsson not sure if you're still on this but... Perhaps a read here would help?
http://www.openscenegraph.org/index.php/documentation/guides/programming-guides/93-viewer-vs-compositeviewer

Basically... we need compositeviewer (a list of views) since OpenMW-CS can have multiple views (terrain, objects, etc.).  The 'viewer' was meant purely for one camera (or more slave cameras that you switch between), while the compositeviewer was to allow for multiple viewers to be shown at once.

The examples do indeed only show osgViewer::Viewer so it looks like we'll have to maintain our own osgViewer::CompositeViewer to have a list of osgViewer::Viewer instances in OpenMW-CS.

psi29a:
Looks like we're not the only ones to come across this problem:
https://github.com/rdiankov/openrave/pull/583/files

psi29a:
It seems that dropped osgQt and rolled their own using QOSGViewerWidget

Could be worth looking into.

psi29a:
```
- class ViewerWidget : public QWidget, public osgViewer::CompositeViewer
+ class QOSGViewerWidget : public QOpenGLWidget
```

unelsson:
I've been looking at this a bit. Seems that the old GraphicsWindowQt created a GLWidget 
`_widget = new GLWidget( traits2qglFormat( _traits.get() ), parent, shareWidget, flags );` which is `class GLWidget : public QGLWidget`.

It might be what OpenMW needs. QGLWidget is in QT5: "QGLWidget provides functionality for displaying OpenGL graphics integrated into a Qt application." Anyway, the process isn't just about updating to the new osgQt, but rather dropping that and making something that works. Gotta look into this later!

AnyOldName3:
QGLWidget was deprecated in favour of QOpenGLWidget with the 5.0 release. The whole point of this exercise is to migrate from the deprecated Qt4 approach to the non-deprecated Qt5 approach.

psi29a:
Closing this for now since work has stalled. Feel free to continue your work here: https://gitlab.com/OpenMW/openmw

PR #2787, opened 2020-04-19
Added setting to change the directory screenshots are stored in

p4r4digm:
Was annoyed that all my screnshots were clogging up my user data path so whipped this up.  

Uses boost filesystem for creating directories so should be cross platform and falls back to the user data directory if making the folder fails.  I was torn on putting this in the ConfigurationManager because although it thematically fits, it needs the settings string passed into it and returns by-value.  This seemed consistent enough with the behavior the configurationmanager already does in the constructor.  Let me know if you dont like it there, I'm completely new to this codebase (and also deeply prefer c-style structure)

Tested relative and absolute paths on windows only.

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -686,7 +686,7 @@ void OMW::Engine::go()
 #endif
 
     mScreenCaptureOperation = new WriteScreenshotToFileOperation(
-        mCfgMgr.getScreenshotPath(Settings::Manager::getString("screenshot path", "General")).string(),
+        mCfgMgr.getScreenshotPath().string(),
         Settings::Manager::getString("screenshot format", "General"));
 
     mScreenCaptureHandler = new osgViewer::ScreenCaptureHandler(mScreenCaptureOperation);
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -657,7 +657,6 @@ private:
 };
 
 // Initialise and enter main loop.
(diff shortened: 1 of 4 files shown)


akortunov:
If I remembered correctly, we avoid to store paths in the settings.cfg due to security/safety reasons (such settings may lead to path traversals, allow to overwrite existing files, etc.).
For same reasons we escape special characters in player's name.

p4r4digm:
yeah i was pretty torn on absolute paths support.  We could lock down the feature to just boolean 'store screenshots to a seperate folder in userdata' if that's more comfortable.  Even make it opt-out depending on the rules around settings.cfg back-compat.  Let me know.  

The core feature here is having an option to declutter the user data folder.

Capostrophic:
How about adding a new folder to user's local data directory called "screenshots" to put screenshots there?

akortunov:
> How about adding a new folder to user's local data directory called "screenshots" to put screenshots there?

A good idea, but it may be not what PR's author wants.

p4r4digm:
I figured that people would want this feature to be configurable but if not then there's no reason to navel-gaze on this.  I've updated the feature to save screenshots in a folder in userdata called screenshots.

psi29a:
> I figured that people would want this feature to be configurable but if not then there's no reason to navel-gaze on this. I've updated the feature to save screenshots in a folder in userdata called screenshots.

I like the cut of your jib.

PR #2788, opened 2020-04-20
Make sure non-carriable light sources can't be picked up

Capostrophic:
Name check would never have worked as intended in many cases either way, and now that objects without a name get an ID as that name, it's much better to check for the presence of a tooltip. This is consistent with how we early-out before activating objects directly.

Should be cherry-picked into 0.46 branch.


psi29a:
cherry picked

PR #2791, opened 2020-04-21
Consider the first person movement sneaking if it's actually sneaking

Capostrophic:
There seem to still be idle reset troubles: sneaking idle state somehow doesn't get reset when you get back from sneaking state while you have an idle fallback turned on. But I don't want to experiment with idle reset fixes anymore (especially when the symptom is in something completely unrelated to idles) and this is just something that will work with movement. Especially considering that the touched code is already a hack.

Should go into 0.46 branch.


psi29a:
cherry-picked

PR #2792, opened 2020-04-21
Fix outdated bug tracker links

Capostrophic:
Needless to say they're not exactly accurate right now.


psi29a:
0.46 material?

akortunov:
More likely somewhat 0.44 one.

psi29a:
cherry-picked

PR #2793, opened 2020-04-22
Pick the correct aggressive actor in fight distance bias calculation

Capostrophic:
I don't remember why I made it actor2, actor1 is the actual actor that the bias are being calculated for while actor2 is their target.

Should go into 0.46 branch.


psi29a:
*falls over*

PR #2795, opened 2020-04-22
Test

IISAKNIFE:


psi29a:
Looks like you have some merge conflicts...

> Conflicting files
> apps/openmw/mwgui/windowmanagerimp.cpp
> apps/openmw/mwinput/inputmanagerimp.cpp

From reading into the PR, looks like switch support. Cool!  Mind telling us a bit more? :)

components/resource/scenemanager.cpp
```diff
@@ -515,9 +515,13 @@ namespace Resource
             // share state
             // do this before optimizing so the optimizer will be able to combine nodes more aggressively
             // note, because StateSets will be shared at this point, StateSets can not be modified inside the optimizer
-            mSharedStateMutex.lock();
-            mSharedStateManager->share(loaded.get());
-            mSharedStateMutex.unlock();
+            static bool allowOptimizations = true;
+            if (allowOptimizations)
```
akortunov, line 519: If you do not want to merge nodes, you can just disable this feature via OPENMW_OPTIMIZE environmental variable.

apps/openmw/switch_startup.cpp
```diff
@@ -0,0 +1,211 @@
...
+            u128 uid = *(u128*)&st_out[8];
+            if (lres == 0)
+                out_id = uid;
+        }
+        return out_id;
+    }
+
+    void fatal(const char *fmt, ...)
+    {
+        FILE *f = fopen("fatal.log", "w");
```
akortunov, line 98: OpenMW writes its output to the openmw.log on different platforms. It would be better to use the same file name on Switch.

apps/openmw/switch_startup.cpp
```diff
@@ -0,0 +1,211 @@
...
+            );
+
+        Log(Debug::Info) << "Found Morrowind.ini, converting.";
+
+        // perform ini => cfg conversion and save the cfg
+        try
+        {
+            MwIniImporter importer;
+            // TODO: figure out how to detect this
+            importer.setInputEncoding(ToUTF8::WINDOWS_1252);
```
akortunov, line 144: On other platforms we set it manually in the importer.

apps/openmw/switch_startup.cpp
```diff
@@ -0,0 +1,211 @@
...
+        if (R_SUCCEEDED(rc))
+        {
+            AccountProfileBase pb;
+            rc = accountProfileGet(&profile, nullptr, &pb);
+
+            if (R_SUCCEEDED(rc))
+            {
+                // HACK: check if the username is printable ascii
+                // otherwise throw it out for the time being, since fs seems
+                // allergic to certain UTF-8 characters
```
akortunov, line 54: Google says that Switch uses the exFAT filesystem with UTF-16 encoding, not UTF-8. Probably the same thing for accounts. In this case you should convert your paths to UTF-8 by using `bconv`, as we do on Windows.

akortunov:
Also if Nintendo Switch have gyroscopes, you can write a provider for SDL sensors API (and sent it to SDL developers) to rotate camera via gyroscope.
OpenMW itself already supports this feature by listening the SDL_SENSORUPDATE event.

Capostrophic:
This doesn't seem to be the same person who worked on the port nor does it seem it was submitted with the permission of that person. Maybe close this?

psi29a:
> This doesn't seem to be the same person who worked on the port nor does it seem it was submitted with the permission of that person. Maybe close this?

Agreed. It's good that it's there but we don't have anyone who can test this and carry on the work. If the original author would want to come in and support it, great!

PR #2796, opened 2020-04-22
store rttcamera as renderingmanager attribute to fix screenshot

mp3butcher:


mp3butcher:
arg not enough changes

PR #2797, opened 2020-04-23
Remove issue that hasn't been solved properly from the changelog

Capostrophic:
Either something weird happened or I just tested it poorly but it seems to be reproducible anyways.

Yes, this should go into 0.46.0 branch.

At least the number of closed issues is round now (280 issues).


psi29a:
Woops... cherry-picked

PR #2798, opened 2020-04-23
Log shader template name on parse error

elsid:
That's not very convenient to see in a log something like:
```
Undefined normalMap
```
without any reference where it happened.


AnyOldName3:
It might have been better to make it clearer that the variable held the template name, not the template itself.

PR #2799, opened 2020-04-24
Fix simple water with radial fog enabled

Capostrophic:
Apparently recalculating euclidean depth on per-fragment basis works for simple water.

Things to note:
1) The way this is done means on older GPUs the length of the view position vector will be calculated once for every vertex and once per every fragment, while on modern GPUs the second calculation will be optimized out for tiles which lack simple water. I've considered doing it only once (once per fragment) for every object but that isn't ideal: while per-fragment calculation isn't really going to be avoidable, the per-vertex calculation cost should be extremely minor for water tiles, while on modern GPUs per-fragment calculation is clearly avoidable.

2) Initializers for uniforms are legal in GLSL 1.20 though I don't know if they work on Android.

3) Should go into 0.46.0 branch too.

### DIFF
--- a/components/sceneutil/waterutil.cpp
+++ b/components/sceneutil/waterutil.cpp
@@ -74,6 +74,9 @@ namespace SceneUtil
 
         stateset->setRenderBinDetails(renderBin, "RenderBin");
 
+        // Let the shader know we're dealing with simple water here.
+        stateset->addUniform(new osg::Uniform("simpleWater", true));
+
         return stateset;
     }
 }
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -49,6 +49,8 @@ uniform vec2 envMapLumaBias;
 uniform mat2 bumpMapMatrix;
 #endif
 
+uniform bool simpleWater = false;
+
 varying float euclideanDepth;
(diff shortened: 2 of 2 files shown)


AnyOldName3:
This is probably better for a make-the-release-branch-not-awful-in-a-hurry commit than a actually-let-it-onto-master commit.

akortunov:
> This is probably better for a make-the-release-branch-not-awful-in-a-hurry commit than a actually-let-it-onto-master commit.

Do you know better alternatives?

AnyOldName3:
I have a few ideas:
* Break the water up into multiple planes and it works like distant terrain with the quadtree. This might be a pain to do without doing a bazillion draw calls, which we've got too many of already.
* Do something else to submit a less stupid water plane mesh.
* Submit the plane in view space, then change the depth of the vertices in the vertex shader so we cover everywhere that's currently visible, but don't have huge triangles. This would require a custom vertex shader just for the water plane and wouldn't help the FFP, though.

akortunov:
So the real alternative is to "do something else"?

Capostrophic:
I'm fine with this only getting into 0.46 branch, experimenting with doing something smarter like paging is not something I want to do at the moment. Just like a proper fix of underwater fog radialization.

PR #2802, opened 2020-04-24
enable BB culling instead of primitive culling for shadow pass

mp3butcher:
less cpu overhead than primitive culling


psi29a:
This is already done in bzzt's MR:

https://gitlab.com/OpenMW/openmw/-/merge_requests/189

PR #2803, opened 2020-04-24
Add env variable to disable crash catcher

elsid:
Sometimes there is no need to have it but it's impossible to disable it without running a process under debugger.

### DIFF
--- a/components/crashcatcher/crashcatcher.cpp
+++ b/components/crashcatcher/crashcatcher.cpp
@@ -555,6 +555,9 @@ static bool is_debugger_present()
 
 void crashCatcherInstall(int argc, char **argv, const std::string &crashLogPath)
 {
+    if (const auto env = std::getenv("OPENMW_DISABLE_CRASH_CATCHER"))
+        if (std::atol(env) != 0)
+            return;
     if ((argc == 2 && strcmp(argv[1], crash_switch) == 0) || !is_debugger_present())
     {
         int s[5] = { SIGSEGV, SIGILL, SIGFPE, SIGBUS, SIGABRT };


components/crashcatcher/crashcatcher.cpp
```diff
@@ -555,6 +555,8 @@ static bool is_debugger_present()
 
 void crashCatcherInstall(int argc, char **argv, const std::string &crashLogPath)
 {
+    if (std::getenv("OPENMW_DISABLE_CRASH_CATCHER"))
```
kcat, line 558: You might want to check what it's actually set to, like `true` or `1`. If I set e.g. `OPENMW_DISABLE_CRASH_CATCHER=false` or `0` I'd expect it to be enabled.

elsid: I wanted something very simple, but ok. When there is a check for a specific value then there is a question how to classify string by true and false. Is there such thing as invalid value. For example how to handle `OPENMW_DISABLE_CRASH_CATCHER=foo` or `OPENMW_DISABLE_CRASH_CATCHER=yes` or `OPENMW_DISABLE_CRASH_CATCHER=0x0`.

kcat: Given that it's mostly a debug feature for developers, I'd consider a limited set of recognized values for `true`.
```
if(const char *value = getenv("OPENMW_DISABLE_CRASH_CATCHER"))
{
    if(strcasecmp(value, "true") == 0 || atoi(value) != 0)
        return;
}
```

elsid: `strcasecmp` is not supported by msvc. `strcasecmp` is a part of POSIX not C++ standard. But msvc has `_stricmp` instead.

mp3butcher: strncmp is less subject to bufferoverfow....please use it instead

kcat: `strncmp` is only useful if you know the length of the strings already, and either isn't or you don't know is null terminated. That won't be the case with strings returned by `getenv`; it is null terminated, as is the string literal. You'd just be doing more manual work and increase the risk of error.

elsid: `strncmp` doesn't allow case insensetive compare. Anyway if the purpose of this env variable can be fulfiled by ability to define a value by `0` or `1` why do we add a support for `true`. There are following use cases for this variable:
1. Don't use it because default behavior is fine.
2. Add it to disable crash catcher:
    ```bash
    env OPENMW_DISABLE_CRASH_CATCHER=1 openmw
    ```
3. Set value to `0` to enable crash catcher back with minimal change:
    ```bash
     env OPENMW_DISABLE_CRASH_CATCHER=0 openmw
    ```

Do you see any other use cases?

mp3butcher: ...and not case case :)

kcat: > Anyway if the purpose of this env variable can be fulfiled by ability to define a value by `0` or `1` why do we add a support for `true`.

It's just something I default to checking. If you don't think it's necessary and there's no complaints, feel free to simplify it to just `if(atoi(value) != 0)` or something.

AnyOldName3: Don't we have fast, simple string comparing functions in `misc` or `utils` or something with `#ifdefs` to use the most appropriate version for the current platform?

PR #2804, opened 2020-04-24
Fix warnings

elsid:


AnyOldName3:
Two points:
* I've not used `std::array` much, but it's raising a mental red flag that there's a hardcoded `7` as the array length. Surely the compiler should be able to infer the length parameter from the fact that we're initialising the array with `{{ stuff }}`.
* It might just be a better idea to ask `osgDB` which readers it has that can return `osg::Image`.

elsid:
Class template arguments deduction is a C++17 feature: https://en.cppreference.com/w/cpp/container/array/deduction_guides, so it's not possible to use it now. Also even with C++17 partial argument deduction (`std::array<std::string> v{"foo", "bar"}`) will not work and probably will never be added to C++ standard. If you're interested [here](https://youtu.be/7LPQWqAZZqs?t=4901) is explanation why.

psi29a:
We can bump to C++17 and see. So long as MacOS accepts (the rest shouldn't be a problem) it, then we're good to go. Right now, the only thing really holding us back was that std::filesystem wasn't yet supported, but that doesn't mean that other things haven't been implemented. 

Perhaps it's worth a PR to bump C++17 and give it at try?

AnyOldName3:
Android's still missing chunks of C++17

psi29a:
Yes, so is MacOS pre 10.15  

As I said, we can give it a try.

xyzz:
https://github.com/android/ndk/issues/609 is still open so that's at least one part that's missing. Not sure if there's anything else.

psi29a:
@xyzz could you test if this is available as of r21 ?

https://github.com/OpenMW/openmw/pull/2804#issuecomment-630430690

PR #2806, opened 2020-04-25
Add detournavigator test for multiple worker threads

elsid:
Basically verify that this works. Also log thread id. Fixed possible issue with `AsyncNavMeshUpdater::wait`. It didn't wait until the end for processing of all changes. Only until all jobs are taken from the queue.

### DIFF
--- a/apps/openmw_test_suite/detournavigator/navigator.cpp
+++ b/apps/openmw_test_suite/detournavigator/navigator.cpp
@@ -699,4 +699,71 @@ namespace
 
         EXPECT_FLOAT_EQ(distance, 85.260780334472656);
     }
+
+    TEST_F(DetourNavigatorNavigatorTest, multiple_threads_should_lock_tiles)
+    {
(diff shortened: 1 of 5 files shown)


psi29a:
Not related, but... it seems nightlies on the ubuntu side are broken through this:
https://launchpadlibrarian.net/477398362/buildlog_ubuntu-xenial-arm64.openmw_0.45.0+git202004281116~ubuntu16.04.1_BUILDING.txt.gz
Not sure if this was fixed?

elsid:
This pr is not about fixing the build. I assume current master is not broken because badges are green and latest runs for multiple compilers are successful. The log doesn't show commit id to verify when this problem was introduced.

elsid:
The compile error in the log seems to be gcc [bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61636) that is fixed in 7.1: https://godbolt.org/z/FLzj-c. The simple workaround is to add `this->` to call the method.

PR #2808, opened 2020-04-25
Clean up spellcasting

Capostrophic:
Not the actual refactoring, just something that would make that refactoring less bloated.

Some clean up and small corrections.

1. spellSchoolToSkill() and getSummonedCreature() have their map initialized in a shorter C++11-based way. operator[] with an out-of-range index is an UB for arrays, so I'm using at() which does a range check.
2. Unnecessary calcEffectCost overload has been removed, it's typically done elsewhere in the source the way it is now.
3. Range-based for loops are used where applicable.
4. scrawl made a typo in the formula for the spell chance calculation (making it inconsistent with the research), but it actually only changes the result very slightly. That typo has been fixed.
5. Some functions that accept spell IDs consider missing spells for an exceptional situation, but it's not. Now they'll ignore spells that don't exist.
6. Magic effect VFX won't be added if its model path is empty.
7. Spell absorption visitor initialization has been cleaned up.
8. Permanent VFX addition visitor ignores non-continuous magic effects straightaway.
9. Some CastSpell fields are initialized in the class instead of its only constructor because the passed arguments don't modify them.

### DIFF
--- a/apps/openmw/mwmechanics/spellcasting.cpp
+++ b/apps/openmw/mwmechanics/spellcasting.cpp
@@ -32,12 +32,12 @@ namespace MWMechanics
 {
     ESM::Skill::SkillEnum spellSchoolToSkill(int school)
     {
-        static const std::map<int, ESM::Skill::SkillEnum> schoolSkillMap
+        static const std::array<ESM::Skill::SkillEnum, 6> schoolSkillArray
         {
-            {0, ESM::Skill::Alteration}, {1, ESM::Skill::Conjuration}, {2, ESM::Skill::Destruction},
-            {3, ESM::Skill::Illusion}, {4, ESM::Skill::Mysticism}, {5, ESM::Skill::Restoration}
+            ESM::Skill::Alteration, ESM::Skill::Conjuration, ESM::Skill::Destruction,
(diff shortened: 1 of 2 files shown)


apps/openmw/mwmechanics/spellcasting.cpp
```diff
@@ -32,41 +32,23 @@ namespace MWMechanics
...
-    }
-
-    float calcEffectCost(const ESM::ENAMstruct& effect)
-    {
-        const ESM::MagicEffect* magicEffect = MWBase::Environment::get().getWorld()->getStore().get<ESM::MagicEffect>().find(effect.mEffectID);
-        return calcEffectCost(effect, magicEffect);
+            {0, ESM::Skill::Alteration}, {1, ESM::Skill::Conjuration}, {2, ESM::Skill::Destruction},
+            {3, ESM::Skill::Illusion}, {4, ESM::Skill::Mysticism}, {5, ESM::Skill::Restoration}
+        };
+        return schoolSkillMap.at(school);
```
elsid, line 40: `school` is expected to have a value from 0 to 5. Looks like an array or vector would do the same.

Capostrophic: Ok, updated.

PR #2810, opened 2020-04-26
Fix hidden node with NiVisController optimization

Capostrophic:
Geometry that such a node could have could be erroneously optimized out because the check for whether the node has a NiVisController controlling it assumes the node will have the NiVisController as its last controller.

This optimization will have to be redesigned anyway in the future because controllers can control any parent node, but for now it should just not be broken.

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -591,7 +591,13 @@ namespace NifOsg
             {
                 bool hasVisController = false;
                 for (Nif::ControllerPtr ctrl = nifNode->controller; !ctrl.empty(); ctrl = ctrl->next)
-                    hasVisController = (ctrl->recType == Nif::RC_NiVisController);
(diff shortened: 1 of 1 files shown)


akortunov:
Is it for 0.46 or for 0.47?

Capostrophic:
Either will do.

AnyOldName3:
Might this be simpler with an `|=`?

Capostrophic:
Wouldn't that mean we'll have to check every single controller of the node instead of just getting to the NiVisController?

Capostrophic:
I guess you mean leaving a check for hasVisController value and break in and putting hasVisController |= whatever before it. That saves one line, but does it really improve readability?

AnyOldName3:
🤷‍♂️

psi29a:
Merged and cherry-picked to 0.46

PR #2811, opened 2020-04-29
Allow particle emitters to be attached to an arbitrary node

Capostrophic:
Should fix [this](https://gitlab.com/OpenMW/openmw/-/issues/5332) issue in Morrowind Optimization Patch.

Since we use osgParticle and Morrowind modellers themselves didn't seem to put the emitter node later than the particle node in the scene graph, there was an assumption in place that the particle emitter is always placed before the particle system in the scene graph. That's what osgParticle normally expects, since particle processors operate on particles that have already been emitted, but apparently it's not guaranteed for every model.

This allows NIF loader to attach the emitter to any emitter node in the NIF hierarchy and fixes cloning operations to make sure that particle processors like an emitter that is attached later than the particle system still get its particle system pointer updated. I didn't remove the assumption that the particle system updater must be placed before the particle system as that doesn't seem to be necessary in the current setup (and probably will never be) and violating osgParticle's assumptions further isn't very nice.

Hopefully this should be everything necessary.

Looks like it's not actually that intimidating of a workaround, so we probably want to get this into 0.46.0. 0.46.0 has a lot of other MOP compatibility fixes, so I didn't add a changelog entry.

### DIFF
--- a/components/sceneutil/clone.hpp
+++ b/components/sceneutil/clone.hpp
@@ -35,10 +35,11 @@ namespace SceneUtil
         virtual osg::Object* operator ()(const osg::Object* node) const;
 
     private:
-        // maps new ParticleProcessor to their old ParticleSystem pointer
+        // maps new pointers to their old pointers
         // a little messy, but I think this should be the most efficient way
         mutable std::map<osgParticle::ParticleProcessor*, const osgParticle::ParticleSystem*> mMap;
         mutable std::map<osgParticle::ParticleSystemUpdater*, const osgParticle::ParticleSystem*> mMap2;
+        mutable std::map<const osgParticle::ParticleSystem*, osgParticle::ParticleSystem*> mMap3;
     };
 
 }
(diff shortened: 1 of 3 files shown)


components/sceneutil/clone.hpp
```diff
@@ -35,10 +35,11 @@ namespace SceneUtil
         virtual osg::Object* operator ()(const osg::Object* node) const;
 
     private:
-        // maps new ParticleProcessor to their old ParticleSystem pointer
+        // maps new pointers to their old pointers
         // a little messy, but I think this should be the most efficient way
         mutable std::map<osgParticle::ParticleProcessor*, const osgParticle::ParticleSystem*> mMap;
         mutable std::map<osgParticle::ParticleSystemUpdater*, const osgParticle::ParticleSystem*> mMap2;
+        mutable std::map<const osgParticle::ParticleSystem*, osgParticle::ParticleSystem*> mMap3;
```
akortunov, line 42: A really awful way to name variables.

AnyOldName3:
I don't see any glaring issues.

psi29a:
Great work! It's clean enough for 0.46 too. Cherry-picked.

PR #2812, opened 2020-04-29
Workaround for GCC 5 bug

elsid:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61636

Revealed here: https://launchpadlibrarian.net/477398362/buildlog_ubuntu-xenial-arm64.openmw_0.45.0+git202004281116~ubuntu16.04.1_BUILDING.txt.gz


psi29a:
Thanks!

PR #2813, opened 2020-04-29
Fix collision switch node mask (again)

Capostrophic:
psi29a accidentally set it to 0 (disabled) instead of what it is supposed to be, Mask_Effect. However, that mask is specific to openmw application, so thanks to the same approach being used as the one for hidden nodes the editor will use the default node mask for such nodes.

This is a 0.47.0 regression.


elsid:
Does the `Loader` class start do become a singleton or it already is?

akortunov:
> Does the Loader class start do become a singleton or it already is?

No, we create a separate instance every time we load NIF or KF file.
psi29a just decided for some reason that static variables are a proper way to use node masks, which are defined in applications, in components code.

Capostrophic:
Using static variables for application-dependent behavior is not his idea, sShowMarkers flag has been around for over 5 years.

PR #2814, opened 2020-04-29
Revert an invalid attempt to autoequip shields instead of torches

Capostrophic:
On a second thought, that idea was awful.

Fixes [this](https://gitlab.com/OpenMW/openmw/-/issues/5388). I've reopened 5300.

The reason for the issue is that weapons are autoequipped too based on the actor's skills even though combat package tries to properly rate the weapon and autoequip the better weapon instead. And it's not especially efficient to autoequip stuff every single time a light is attempted to be equipped while in combat.


psi29a:
Cherry-picked to 0.46

PR #2816, opened 2020-04-30
CollisionSwitch and Switch/LOD node fixes

Capostrophic:
Fixes up some remaining limitations of NiSwitchNode/NiLODNode usage in 0.46.0.

1) They can both properly handle their transformations and controllers. NiLODNode had previously supported them in 0.45.0. NiSwitchNode isn't supposed to be a leaf node in the hierarchy and will properly support NiKeyframeControllers without causing crashes.
2) They can both be used as the root node of the model. E.g. a plant with NiSwitchNode as the root node will support herbalism, and while NiLODNode can support it its capability to be transformed was previously removed to do that and now it's restored (note that root nodes' transformations are ignored). Windows users may want to test it (check Morrowind Optimization Patch old 6.0 version using the information from [this](https://gitlab.com/OpenMW/openmw/-/issues/4837) report and also check the attachment). 
3) The way their children are not optimized out is much closer to how upstream OSG does it and is shorter.

The problem with scrawl's original approach that akortunov tried to workaround was that handleNode() tried to return the osg::LOD node instead of the actual root transform node it was attached to to handleNode() caller. This caused issues or crashes. akortunov just removed the transform wrapper for LOD nodes and then restored it for Switch nodes. Now both types of record are wrapped into a transform node and this transform node is returned as the root node instead of the node with special behavior, but NiSwitchNode/NiLODNode children are still attached to the wrapped node.

akortunov tried to reimplement what was there present in upstream optimizer but was removed by scrawl accidentally (osg::Switch handling was purged from the initial optimizer fork).

Additionally, CollisionSwitch group (the equivalent of NiCollisionSwitch) is now properly added to the scene graph instead of being just a matrix transform due to the absence of CopyOp and other vital things so something that might be added later that works with the scene graph should be able to toggle its state properly. Serializer used for ShowSceneGraph instruction will not show it.

This archive contains a modified heather model from Graphic Herbalism and its harvested texture which has a NiSwitchNode as its root node.
[heather.zip](https://github.com/OpenMW/openmw/files/4558285/heather.zip)

Once this is merged, it should go to 0.46.0 branch as well, as is the tradition.

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -170,6 +170,17 @@ namespace NifOsg
     class CollisionSwitch : public osg::MatrixTransform
     {
     public:
+        CollisionSwitch() : osg::MatrixTransform()
+        {
+        }
+
+        CollisionSwitch(const CollisionSwitch& copy, const osg::CopyOp& copyop)
+            : osg::MatrixTransform(copy, copyop)
+        {
+        }
+
+        META_Node(NifOsg, CollisionSwitch)
+
         CollisionSwitch(const osg::Matrixf& transformations, bool enabled) : osg::MatrixTransform(transformations)
         {
             setEnabled(enabled);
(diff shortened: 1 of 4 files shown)


AnyOldName3:
https://cdn.discordapp.com/attachments/502999428798480387/705405856530956378/0503-0774.mp4

Capostrophic:
Point 2 works on Windows according to CMAugust.

PR #2818, opened 2020-05-01
Fix usage of uninitialized weapon type in equipmentChanged()

Capostrophic:
Unlike other users of getActiveWeapon(), equipmentChanged() assumes it will always set the weapon type to a valid value, but getActiveWeapon() doesn't change the weapon type if there's no active weapon. Now it does the same thing as all the other callers of getActiveWeapon(): uses None weapon type.

Previously weapon type in this scope could be anything since it was uninitialized, so updateCarriedLeftVisible got an invalid value fed to it.

This fixes invisible torches of NPCs that have them out of combat while shield sheathing is enabled.

And yeah, it's a 0.46.0 regression, get this into 0.46.0 branch.


psi29a:
cherry-picked

PR #2790, opened 2020-04-20
Proposal of omnidirectional lighting shading for particles

mp3butcher:
Force particles to face each light in shader path
It may be better that vanilla arbitrary normal but diverge from ffp replication
before
screenshot283
after
screenshot292 
https://streamable.com/n8y1xq

PR #2823, opened 2020-05-06
precipitation diffuse lighting and fading issue

mp3butcher:
was gone to just let ambient lighting as it was before https://github.com/OpenMW/openmw/pull/2781 but at the end i fixed precipitation path for diffuse lighting 
little issues emerging:
- per pixel shading is very expensive for the snow (omnidirectional normal pr https://github.com/OpenMW/openmw/pull/2790  could lever a bit ppl charge) EDIT I FORBID PPL ON PARTSYS
Capture d’écran de 2020-05-08 15-57-25
- shadow receiving storm is trippy according Capo  EDIT I DISABLED SHADOW ON WEATHER
https://streamable.com/9rl9wc
do i disable weather shadow or not?
I don't find it that ugly and think that adding a setting for that could be the object  of an other pr


Capostrophic:
Won't weather particle systems be unable to get diffuse lighting working then?

mp3butcher:
yes it won't..as before by pr

Capostrophic:
Yes. That's not ideal if they receive diffuse lighting in Morrowind.

mp3butcher:
in all cases you don't use openmw particlesystem ( only regular osgparticle::particlesystem) for precipitation so no diffuse lighting for them

mp3butcher:
another way would be to hack particle template...will test

mp3butcher:
...no you will do with it...otherway is too hacky...perhaps

mp3butcher:
my last commit  is dirty...but that's what it'd cost to have diffuse lighting on precipitation...tell me if you're ok with this and i'll clean it...else i'll rollback

Capostrophic:
Aside from using OSG naming conventions it's not actually that terrible if it does work as it is supposed to IMO. May want to add a comment that any existing transparency modifications will be overridden.

Capostrophic:
Nice, the lighting still looks much closer to vanilla in conditions where 0.46.0 storms didn't.
Although there's the odd flickering issue noticed by CMAugust.
Also rain definition in sky.hpp should probably be corrected? Not sure.

Capostrophic:
Shadows really shouldn't be casted on particles. It looks pretty trippy.

mp3butcher:
Ready to merge
per pixel shading is very expensive for the snow
https://media.discordapp.net/attachments/260443767482482688/708351054110261278/Capture_decran_de_2020-05-08_15-57-25.png?width=1030&height=645
do i disable weather shadow or not?
https://streamable.com/9rl9wc
I don't find it that ugly and think that adding a setting for that could be the object  of an other pr

mp3butcher:
not yet mergeable  there's weather indoor now:/

akortunov:
> little issues emerging

These "little" issues look like blockers for me.

akortunov:
If I remembered correctly, AnyOldName3 managed to make sky to do not receive shadows (something like [this](https://github.com/AnyOldName3/openmw/tree/osgshadow-test-vdsm-sky-method-two)). Can the used approach be re-used for particle systems?

AnyOldName3:
There's the `disableShadowsForStateSet` function in `shadow.h` for disabling shadows.

apps/openmw/mwrender/sky.cpp
```diff
@@ -1332,7 +1348,9 @@ class AlphaFader : public SceneUtil::StateSetUpdater
                         callback = callback->getNestedCallback();
                     }
 
-                    osg::ref_ptr<AlphaFader> alphaFader (new AlphaFader(mAlphaUpdate));
+                    FindParticleSystemVisitor fpart;
+                    node.accept(fpart);
+                    osg::ref_ptr<AlphaFader> alphaFader (new AlphaFader(mAlphaUpdate, fpart.mFound));
```
Capostrophic, line 1353: This will only make one AlphaFader that will track *one* particle system that the node has (the last particle system child). Aren't all of them supposed to be tracked?

Can't NifOsg::FindByClassVisitor be used?

apps/openmw/mwrender/sky.cpp
```diff
@@ -1730,7 +1747,8 @@ void SkyManager::setWeather(const WeatherResult& weather)
                 mParticleNode = new osg::PositionAttitudeTransform;
                 mParticleNode->addCullCallback(mUnderwaterSwitch);
                 mParticleNode->setNodeMask(Mask_WeatherParticles);
-                mRootNode->addChild(mParticleNode);
+                mParticleNode->addCullCallback(new SceneUtil::LightListCallback);
```
Capostrophic, line 1750: Stuff that uhh enables lighting on weather particle systems I guess should be put to a separate PR, or we'll never get particle system fading (the core issue and an important regression) fixed. I'm also confused as to why is all the dancing around two root nodes necessary.

apps/openmw/mwrender/sky.cpp
```diff
@@ -1422,7 +1432,7 @@ class WrapAroundOperator : public osgParticle::Operator
         osg::Matrix toWorld, toLocal;
 
         std::vector<osg::Matrix> worldMatrices = ps->getWorldMatrices();
-
```
Capostrophic, line 1425: unnecessary formatting change

apps/openmw/mwrender/sky.cpp
```diff
@@ -1472,7 +1482,7 @@ void SkyManager::createRain()
 
     mRainNode = new osg::Group;
 
-    mRainParticleSystem = new osgParticle::ParticleSystem;
+    mRainParticleSystem = new NifOsg::ParticleSystem;
```
Capostrophic, line 1485: Why isn't mRainParticleSystem type changed to NifOsg one anyway?

apps/openmw/mwrender/sky.cpp
```diff
@@ -1753,11 +1771,10 @@ void SkyManager::setWeather(const WeatherResult& weather)
 
                 for (unsigned int i = 0; i < findPSVisitor.mFoundNodes.size(); ++i)
                 {
-                    osgParticle::ParticleSystem *ps = static_cast<osgParticle::ParticleSystem *>(findPSVisitor.mFoundNodes[i]);
-                    
+                    osgParticle::ParticleSystem *partsys = static_cast<osgParticle::ParticleSystem *>(findPSVisitor.mFoundNodes[i]);
                     osg::ref_ptr<osgParticle::ModularProgram> program (new osgParticle::ModularProgram);
                     program->addOperator(new WrapAroundOperator(mCamera,osg::Vec3(1024,1024,800)));
-                    program->setParticleSystem(ps);
+                    program->setParticleSystem(partsys);
```
Capostrophic, line 1777: also unnecessary

Capostrophic:
SetupVisitor checks for the presence of a material on the stateset of the node it visits but AlphaFaders don't actually need a material of that node to work anymore, which seems misleading.

apps/openmw/mwrender/sky.cpp
mp3butcher: dancing around two root nodes is required cause sky is not shadable (or in hardcoded manner i guess) but particles are

apps/openmw/mwrender/sky.cpp
mp3butcher: yes

apps/openmw/mwrender/sky.cpp
mp3butcher: just in order to be consistant with other partsys

apps/openmw/mwrender/sky.cpp
mp3butcher: yes

apps/openmw/mwrender/sky.cpp
mp3butcher: > Can't NifOsg::FindByClassVisitor be used?

yes it could

mp3butcher: > Aren't all of them supposed to be tracked?


yes i will use a findbyname

apps/openmw/mwrender/sky.cpp
mp3butcher: in fact i  removed an unecessary whitespace

mp3butcher:
> SetupVisitor checks for the presence of a material on the stateset of the node it visits but AlphaFaders don't actually need a material of that node to work anymore, which seems misleading.

I removed reference of previous stateset path

psi29a:
Is performance still crappy with per pixel shading? Would it make sense to merge #2790 work into this PR to resolve it?

psi29a:
It would be great to see a video of what changes, a sort of before and after along with some performance information.

mp3butcher:
> Is performance still crappy with per pixel shading? Would it make sense to merge #2790 work into this PR to resolve it?

Indeed it would improve a bit per pixel gpu charge..
My guess is a setting 'forbid ppl for weather particle system' would be pertinent too
and perhaps an other 'disable weather receiving shadow'...

psi29a:
Let's not get all settings crazy. ;)

If we can solve performance issues in one go with out tweaks, then it should be okay. The biggest question is, how much of a performance hit are we talking about? Is this an effect we want?

akortunov:
> The biggest question is, how much of a performance hit are we talking about?

Take a look at screenshot in the PR description. Shadowing is too expensive for particles, partially due to overdraw.
I'd disable shadowing for particles systems at all (MGE does not have it too).

vtastek:
Particles can use simple sun + ambient lighting(no pixel shading, no dot product even), but shadows should stay imo. I thought we were getting soft shadows with penumbra, that would fix it I think. Otherwise maybe, dilute shadows 50% or some other technique should work.(maybe with eye/light dot).

AnyOldName3:
Shadows are *way* more expensive than PPL.

vtastek:
> Shadows are _way_ more expensive than PPL.

I am more concerned about the look of fully lit particles in shadows but if it is only for blizzard and dust then I have sun at zero intensity anyways in my setup so I see no shadows! I will leave performance for better knowledgeable people.

akortunov:
> Shadows are way more expensive than PPL.

For particles they are both expensive with our current shaders.

mp3butcher:
> > Shadows are way more expensive than PPL.
> 
> For particles they are both expensive with our current shaders.

the problem is that particles project several time on the same pixel and blend with current fragment (read in the framebuffer no cache)..the more expensive is snow as it has big particle size.
Sampling shadow is cheap as it benefit of sampling cache.

mp3butcher:
I admit i stalled a bit on this one (in order to hack properly shader generation attribute PPL only for weather particle system)

akortunov:

AnyOldName3:
Due to posts being edited and the goals being changed, it's no longer especially clear what this PR is trying to do. There's no linked issue report, and the two intentional changes I can identify easily seem to have been added in response to code review rather than have been the original goal.

akortunov:
An initial issue is [#5455](https://gitlab.com/OpenMW/openmw/-/issues/5455).
Weather transitions do not work properly now since the way to handle alpha for particle systems was changed in #2781.

AnyOldName3:
I have a few notes:

* `AlphaFader` doesn't update statesets anymore, so shouldn't be a `StateSetUpdater`, just a regular `osg::Callback`. This should simplify some of the dancing around with callbacks as it can then be used as a regular nested callback with `addUpdateCallback` instead of getting attached to a `CompositeStateSetUpdater`.
* I don't think *strippy* in comments is meant to be that word - *trippy* or *stripey* are both more likely.
* I'm concerned that there isn't any obvious thread-safety stuff in place for updating the particle system's alpha. It could just be that it's handled by the particle system, and we don't need to do anything here, but I'd like some assurance that it's not going to upset the frame which is in-flight while we're preparing the next one.

mp3butcher:
>     * I'm concerned that there isn't any obvious thread-safety stuff in place for updating the particle system's alpha. It could just be that it's handled by the particle system, and we don't need to do anything here, but I'd like some assurance that it's not going to upset the frame which is in-flight while we're preparing the next one.

osg has its own way to control particle system through programs composed of operator..the proper way would be to create a new
`class AlphaFaderOperator : public osgParticle::Operator`
 affecting only alpharange of particles and add/remove it when required to the modularprogram of particlesystems.

psi29a:
@mp3butcher I hope you're doing alright. Will you be able to continue this work?

It's been awhile since we last heard from you, I'm going to close this for now. Feel free to re-open, if you could, please do so on gitlab:

mp3butcher:
it's finished..do with it

Capostrophic:
> AlphaFader doesn't update statesets anymore, so shouldn't be a StateSetUpdater, just a regular osg::Callback. This should simplify some of the dancing around with callbacks as it can then be used as a regular nested callback with addUpdateCallback instead of getting attached to a CompositeStateSetUpdater.

This point hasn't been resolved.

PR #2824, opened 2020-05-07
Cleanup

Capostrophic:
Addresses akortunov's concerns about water culling merge request: static variable declaration is moved out of the loop, which shouldn't change the behavior (which is probably ok) but should look slightly less odd to the reader, (subjectively) improved HeightCullCallback class formatting.

applyDrawableProperties() no longer accepts a particle material argument that went unused after mp3butcher's particle lighting fixes.

TerrainDrawable constructor and destructor are left alone, seems that they're put into the implementation file for a good reason.

### DIFF
--- a/components/terrain/world.hpp
+++ b/components/terrain/world.hpp
@@ -40,36 +40,46 @@ namespace Terrain
     class ChunkManager;
     class CompositeMapRenderer;
 
-class HeightCullCallback : public osg::NodeCallback
-{
-public:
-    HeightCullCallback() : mLowZ(-std::numeric_limits<float>::max()), mHighZ(std::numeric_limits<float>::max()), mMask(~0) {}
-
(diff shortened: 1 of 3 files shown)


components/terrain/world.hpp
```diff
@@ -40,36 +40,48 @@ namespace Terrain
...
-        if (mLowZ <= mHighZ)
-            traverse(node, nv);
-    }
-private:
-    float mLowZ;
-    float mHighZ;
-    unsigned int mMask;
-};
+    public:
+        HeightCullCallback() : osg::NodeCallback() {}
```
elsid, line 46:
```suggestion
```
will do the same.

Capostrophic: Never quite understood constructor inheritance rules.

PR #2830, opened 2020-05-09
Fix windows install target

elsid:


psi29a:
I'll take this.

AnyOldName3:
So what exactly does this do? I was under the impression the install target was only ever intended for building the installer, and it did that just fine.

psi29a:
It removes the MyGUI plugin we've not ever used which is fine by me. The second thing it does is add an optional install directory param in the script so that windows users can specify an install location which can also used by cmake, like how it is done on linux builds as well. 

If it makes a windows developer's life easier, then I'm fine with this change. As far as I can see, it doesn't make anything worse.

CMakeLists.txt
```diff
@@ -517,16 +513,6 @@ if(WIN32)
...
 
-    if(BUILD_MYGUI_PLUGIN)
-        INSTALL(PROGRAMS "${OpenMW_BINARY_DIR}/Debug/Plugin_MyGUI_OpenMW_Resources.dll" DESTINATION "." CONFIGURATIONS Debug)
-        INSTALL(PROGRAMS "${OpenMW_BINARY_DIR}/Release/Plugin_MyGUI_OpenMW_Resources.dll" DESTINATION "." CONFIGURATIONS Release;RelWithDebInfo;MinSizeRel)
-    ENDIF(BUILD_MYGUI_PLUGIN)
-
-    IF(DESIRED_QT_VERSION MATCHES 5)
-        INSTALL(DIRECTORY "${OpenMW_BINARY_DIR}/Debug/platforms" DESTINATION "." CONFIGURATIONS Debug)
-        INSTALL(DIRECTORY "${OpenMW_BINARY_DIR}/Release/platforms" DESTINATION "." CONFIGURATIONS Release;RelWithDebInfo;MinSizeRel)
-    ENDIF()
```
AnyOldName3, line 528: Why are these lines gone? I can understand not taking the files from `Release` for `RelWithDebInfo`, but the CS, Launcher and Wizard won't work without this folder.

elsid: This directory is not defined by CMakeLists.txt. It wasn't clear why install target didn't work. So I just removed it.

AnyOldName3: It gets deployed by the prebuild script. *Ideally* we should be using Qt's figure-out-which-DLLs-you-need tool (called something like QDeploy), but lots of stuff is a mess.

Either way, this PR just spraying air freshener at the turd that is our Windows INSTALL target. I've got something in the works that at least puts the turd in a plastic bag and ties it shut, so I'll fix this. I'd really like a from-scratch best-practice-following CMake rewrite, though. There's a lot of mess and weirdness, with pretty much just the recast stuff being salvageable.

PR #2834, opened 2020-05-12
Fix merge conflicts

akortunov:
Fix conflict between #2648 and #2716 - both pull request increased a savegame format version from 5 to 6.
Solution: increase savegame format from 6 to 7, move new changelog entry from 0.46 to 0.47.


psi29a:
We won't have to do this for other save game format changes correct?

akortunov:
> I mean, one format change per release is alright.

No, we should increase its version every time when we change it.

Capostrophic:
Strictly speaking we probably should, as if this isn't done some revisions of openmw may break horribly upon loading saves that seem to have the same format but don't. This seems to be the reasoning for format update here as well. Ninja'd.

psi29a:
Roger that.

PR #2836, opened 2020-05-12
Always pass the vertex color to the fragment shader

Capostrophic:
Even when per-pixel lighting is off. Fixes per-vertex lighting being broken atm.
In general it's pretty useful to keep it around for some things. Specular color mode is one of these things, but there's also the decal texture blending thing that would benefit from it.


AnyOldName3:
Did the previous thing with the extra colour modes that got merged before I had a chance to take a proper look at it cause the problem? If you're moving stuff around, there are a lot of potential gotchas with undefined behaviour, which is probably why it was harder to read than you thought it should be in the first place.

PR #2838, opened 2020-05-13
Some more minor NIF stuff

Capostrophic:
1. Remove some more obscure enumerated versions. They are for games we're never reasonably or unreasonably going to properly support anyway, I find myself unable to recall what's the exact version number behind them anyway every single time I reach back to my nif branch, they're not terribly descriptive, there isn't going to be any difference during runtime since generateVersion is constexpr, it's more difficult to compare loader files against NifTools research when they are used, and there are myriads of other reasons why it was a bad idea to add them in the first place. I left versions related to BGS games in, they're easier to remember and are much more relevant.
2. Per request, don't rely on a NIFStream object to call generateVersion, it's unnecessary; use NIFStream::generateVersion() instead.
3. Adjust NiPixelData loading, should be a bit faster and easier to handle if we decide that we need to implement every single pixel format.

### DIFF
--- a/components/nif/niffile.hpp
+++ b/components/nif/niffile.hpp
@@ -79,11 +79,7 @@ public:
     enum NIFVersion
     {
         VER_MW         = 0x04000002,    // 4.0.0.2. Main Morrowind NIF version.
-        VER_CI         = 0x04020200,    // 4.2.2.0. Main Culpa Innata NIF version, also used in Civ4.
-        VER_ZT2        = 0x0A000100,    // 10.0.1.0. Main Zoo Tycoon 2 NIF version, also used in Oblivion and Civ4.
         VER_OB_OLD     = 0x0A000102,    // 10.0.1.2. Main older Oblivion NIF version.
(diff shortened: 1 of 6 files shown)


components/nif/niffile.hpp
```diff
@@ -79,11 +79,7 @@ class NIFFile final : public File
     enum NIFVersion
     {
         VER_MW         = 0x04000002,    // 4.0.0.2. Main Morrowind NIF version.
-        VER_CI         = 0x04020200,    // 4.2.2.0. Main Culpa Innata NIF version, also used in Civ4.
-        VER_ZT2        = 0x0A000100,    // 10.0.1.0. Main Zoo Tycoon 2 NIF version, also used in Oblivion and Civ4.
```
symbiogenesis, line 83: Says this is used in Oblivion

Capostrophic: Oblivion can load many things, doesn't mean I should document every single version it supports.

PR #2839, opened 2020-05-13
Add basic NiPathController support (movement only)

Capostrophic:
A "placeholder" that doesn't quite implement the feature, but seems to be better than nothing. Allows insects from BCSounds, Nocturnal Moths and other models based on the same controller setup to move as intended, doesn't include advanced features of the controller.

Flags and other values in the record are unused but their values should be accurate.

Both geometry-type nodes and group-type nodes should support this controller.

### DIFF
--- a/components/nif/controller.cpp
+++ b/components/nif/controller.cpp
@@ -123,12 +123,10 @@ namespace Nif
     {
         Controller::read(nif);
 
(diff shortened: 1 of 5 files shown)


psi29a:
Thanks! :)

PR #2841, opened 2020-05-13
Script reference consistency

Assumeru:
There's an inconsistency in how implicit and explicit references are handled for a few functions. These functions are defined in components rather than apps/openmw meaning they can't access the reference helpers defined in apps/openmw/mwscript/ref.hpp. So my aim here is to downgrade these functions from language constructs to regular functions and move them into apps/openmw - where the vast majority of their friends reside.

The ones that're interesting for references are:
- Enable
- Disable
- GetDisabled
- StartScript
- GetDistance

For the sake of keeping the scripting functions grouped I also moved ScriptRunning and StopScript.

In all honesty, the remaining functions in installopcodes.cpp should probably also be moved. That's to say, these ones:
- MessageBox
- MenuMode
- Random
- GetSecondsPassed

So... how does everyone feel about this?


zinnschlag:
Hm. Not sure. That's a pretty big refactoring. Does this change actually fixes any issues or is it just code maintenance?

If it is the later then I am not in favour of it. Please remember that the old scripting engine will be removed eventually. There is little point in polishing it up.

Capostrophic:
Meanwhile I like it to be honest, it may take a while before it's rewritten in some way and IMO it would decomplicate the rewriting a lot if all function opcodes are handled in a generic way.

psi29a:
I would like to take this in to consideration. The effort has been done, it is simple/elegant/generic. 

Agreed that we do plan on deprecating and replacing the scripting engine, but I think the existing one will still be around for awhile.

Assumeru:
Aside from removing code bloat, I think Capo's argument might be the best reason to consider merging this. Having everything nice and consistent should aid future refactoring/switching to Lua.

> Does this change actually fixes any issues or is it just code maintenance?

No issues, just maintenance. Mostly because it took me a while to find StartScript when I started #2648.

> The effort has been done

Arguably only the half the effort's been done. (The larger half, admittedly.) Figured I'd get some feedback before going crazy.

zinnschlag:
Okay. I agree that we don't have reason not to merge this one. But I still think it is not a good use of time to continue on this refactoring task (which of course doesn't mean you can't do it, if you really want to). Also, from what I remember MessageBox is kinda tricky. Not sure if it even can fit into the current system without further enhancements.

Assumeru:
I always find deleting code to be pretty cathartic, so I'll probably have a look at the other functions I mentioned soonish.

And after that... who knows; maybe mutable base records so the ordinators can do their thing.

psi29a:
Is this then still WIP?

Assumeru:
Yes, I'm gonna have a look at the other functions I mentioned and (hopefully) add all of them to this.

I also still need to verify I didn't break anything with StartScript, StopScript, and ScriptRunning. It's probably fine, but that's not really good enough. (The others I did test already.)

psi29a:
Make sure to keep in scope, we are not going to be adding to or extending mwscript. We could use this as a basis for obscript however. Something to keep in mind. :)

Assumeru:
> Also, from what I remember MessageBox is kinda tricky.

Yeah, I had a quick look and I'll just leave it be. Not worth it for now.

So with MenuMode, Random, and GetSecondsPassed done and MessageBox skipped... I'm gonna say this bit of refactoring is done. Unless someone has very strong feelings about GetSquareRoot.

elsid:
[This](https://github.com/OpenMW/openmw/blob/2618974ad60959e2f10b66326ee52ab9e66842b4/components/compiler/extensions.cpp#L62) assert fails now:
```
1  raise                                                           0x7ffff2f64ce5 
2  abort                                                           0x7ffff2f4e857 
3  __assert_fail_base.cold                                         0x7ffff2f4e727 
4  __assert_fail                                                   0x7ffff2f5d426 
5  Compiler::Extensions::registerFunction extensions.cpp       62  0x555556cd7552 
6  Compiler::Misc::registerExtensions     extensions0.cpp      244 0x555556cf05f6 
7  Compiler::registerExtensions           extensions0.cpp      17  0x555556ce7c2c 
8  MWGui::Console::Console                console.cpp          148 0x5555564832d7 
9  MWGui::WindowManager::initUI           windowmanagerimp.cpp 457 0x555556452c59 
10 OMW::Engine::prepareEngine             engine.cpp           567 0x555556a9cd26 
11 OMW::Engine::go                        engine.cpp           699 0x555556a9dcdf 
12 runApplication                         main.cpp             259 0x555556a7e88f 
13 wrapApplication                        debugging.cpp        138 0x555556ca9ffd 
14 main                                   main.cpp             271 0x555556a7e9a1 
```

akortunov:
I have a suspicion that it is because of `opcodeGetDistance`.

PR #2842, opened 2020-05-13
Handle non-node roots more gracefully (bug #5416)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/-/issues/5416)

NIF loaders will at least try to find a valid root node before declaring that the NIF file is completely unusable. This fixes the gauntlet model in my testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
     Bug #5369: Spawnpoint in the Grazelands doesn't produce oversized creatures
     Bug #5370: Opening an unlocked but trapped door uses the key
     Bug #5400: Editor: Verifier checks race of non-skin bodyparts
+    Bug #5416: Junk non-node records before the root node are not handled gracefully
     Feature #5362: Show the soul gems' trapped soul in count dialog
 
 0.46.0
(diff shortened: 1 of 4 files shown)


Capostrophic:
Hm. Will fix tests tomorrow.

Capostrophic:
Ok, pushed a bunch of changes that should fix stuff except AppVeyor as that's not related to the PR.

I removed one of the tests, seems that it's not going to end up being reliable. It probably should be replaced with something else as tests still assume getRoot(0) will be the only getRoot call but I admit I've never written tests before and this isn't the time I want to try that out.

psi29a:
The AV problem I'm looking at right now, it seems that the host that Ace is using is acting up again.

Another solution to this problem is to use download.openmw.org and let clouldflair cache it for us. This would save bandwidth for ace _and_ should speak up AV because downloads would be served from cloudlfair cache.

psi29a:
Talking with Ace now, another solution is inbound.

elsid:
> I admit I've never written tests before and this isn't the time I want to try that out.

This was exactly that moment when you would have started to do so. There is no much support for any automatic tests in this project so no one asks to write a test for each new feature we add. But if the test is already there it's the simpliest possible situation to start working on it. I hope no one would argue that this project doesn't need tests for whatever reason.

AnyOldName3:
Having tests will move us further from the authentic Morrowind experience of frequent bugs and crashes that everyone knows and loves?

PR #2847, opened 2020-05-14
Windows CI dependency upgrade

ananace:
Going through and bumping dependencies for the OpenMW Windows builds

Currently built up new packages of Bullet (2.89, both single and double-precision), MyGUI (3.4.0), and OSG (3.6.5).  
Going to look at bumping boost, ffmpeg, OpenAL-Soft, Qt, and SDL as well. But first, sleep. And then some test builds.
These deps have only been bumped, packaged, and pushed, and I haven't had any time to actually build OpenMW with them yet - hoping the CI will be of help in that regard.


ananace:
And something's caused OSG plugins to not install - without error - when building some versions and configurations of the deps, I'll debug that and push new packages as soon as I've had time to do so.

Capostrophic:
Does F10 debug screen work with current Bullet? I think I had troubles with it.

ananace:
Had some other interesting issues with Bullet due to changed default values in their CMakeLists, but in the worst case I can just not bump it for now.
Going to let the CI run a build and see how it looks before I do anything more though.

ananace:
But yeah, the F10 profiler seems to be empty with the newer Bullet version.

ananace:
Seems like Bullet no longer offers any CMake option to enable profiling, so I could recompile it again with support for that - but for OpenMW to realize that it's available that would either need a code changes in the Bullet headers or a similar flag on the OpenMW side.

PR #2851, opened 2020-05-16
Fix AiWander crash

elsid:
Based on report from [discord](https://discordapp.com/channels/260439894298460160/260443579908882434/711120918075670589):
> CMAugustToday at 9:40 AM
Ouch, just had a crash while playing in the bzzt branch
crash
Just exited a door from one vivec hlaalu interior to another and it conked out

Still unclear when this happens. I wasn't able to reproduce the situation when path is empty but storage.mState is Wander_Walking. Looking at the code this is possible only due to package reset. And reset for AiWander happens only when package duration is ended. For now just considering empty path for walking state as condition to stop walking.

Also disabled check for target in LOS for actors wandering over pathgrid. This is quite noticable, actors almost stopped wander at interior cells.

### DIFF
--- a/apps/openmw/mwmechanics/aiwander.cpp
+++ b/apps/openmw/mwmechanics/aiwander.cpp
@@ -292,7 +292,8 @@ namespace MWMechanics
             completeManualWalking(actor, storage);
         }
 
-        if (storage.mState == AiWanderStorage::Wander_Walking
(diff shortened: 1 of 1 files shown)


elsid:
This should have been a part of 0.46 release. https://gitlab.com/OpenMW/openmw/-/issues/5479 is fixed by this pr.

PR #2853, opened 2020-05-16
AiPackage refactoring

elsid:
Store options as values in AiPackage class but initialize from values defined in derived classes. It's still possible to change them during runtime but overhead for virtual calls is removed.


elsid:
Closing in favor of https://github.com/OpenMW/openmw/pull/2891.

PR #2855, opened 2020-05-17
Fix msvc dir for Qt

elsid:
Otherwise it fails with:
```
Qt 5.15.0... Exists. CI/before_script.msvc.sh: line 781: cd: MSVC2019_64_Ninja/deps/Qt/5.15.0/msvc2015_64: No such file or directory
```


AnyOldName3:
I think this is a holdover from when we'd use the 2015 binaries for Qt 5.7 for later MSVC versions as they're forwards compatible and it stopped us needing to bump the Qt version.

PR #2858, opened 2020-05-17
Mark overriden AiPackage methods as final

elsid:


psi29a:
I'll take this, so long as aren't going to derive further then we can take advantage of devirtualisation perf boost. Thanks!

PR #2859, opened 2020-05-17
Single clone function definition for all AI packages

elsid:
This method is equivalent across all AI packages. Own definition for each class is redundant.


Capostrophic:
Looks reasonable.

PR #2865, opened 2020-05-22
Re-enable non-biped creature headtracking (bug #5424)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/-/issues/5424)
Add head controller for eligible non-biped creatures as well.
So far it didn't seem to cause the adverse effects that were there 5 years ago, but it may need more testing.


akortunov:
Just for reference: head tracking for creatures was disabled due to [this](https://gitlab.com/OpenMW/openmw/-/issues/2843) bug. Since we disabled head tracking in combat at all, an original bugged case is irrelevant now. Also head tracking for creatures could be accidentally fixed as a side effect of other patch. As for me, it is worth to try to re-enable it and try to fix if it still does not work properly for some creatures.

PR #2866, opened 2020-05-22
Extend OSG stats

elsid:
Add report for:
* Current frame number (it's not showed by default, probably there is other way how to do this properly).
* Number of mechanics actors
* Number of mechanics objects
* Number of physics actors
* Number of physics objects
* Number of physics heigh fields

Allows to see and analyze dependency between a number of entities and frame time.

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -203,12 +203,14 @@ bool OMW::Engine::frame(float frametime)
 
         if (stats->collectStats("resource"))
         {
+            stats->setAttribute(frameNumber, "FrameNumber", frameNumber);
+
             mResourceSystem->reportStats(frameNumber, stats);
 
             stats->setAttribute(frameNumber, "WorkQueue", mWorkQueue->getNumItems());
             stats->setAttribute(frameNumber, "WorkThread", mWorkQueue->getNumActiveThreads());
 
-            mEnvironment.getWorld()->getNavigator()->reportStats(frameNumber, *stats);
+            mEnvironment.reportStats(frameNumber, *stats);
         }
 
     }
--- a/apps/openmw/mwbase/environment.cpp
+++ b/apps/openmw/mwbase/environment.cpp
@@ -198,3 +198,9 @@ const MWBase::Environment& MWBase::Environment::get()
     assert (sThis);
     return *sThis;
 }
+
+void MWBase::Environment::reportStats(unsigned int frameNumber, osg::Stats& stats) const
+{
+    mMechanicsManager->reportStats(frameNumber, stats);
+    mWorld->reportStats(frameNumber, stats);
+}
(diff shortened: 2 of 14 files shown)


AnyOldName3:
Within my limited understanding of the OSG stats system, this looks reasonable. You do seem to have removed some empty strings, though. I'm guessing they were to add empty lines between sections, so if that's what they were doing, it's probably best to keep them.

elsid:
These empty lines just take place on a screen making the required window size bigger. Visually it helps to separate things by categories but I'd better keep more screen clean of debug info.

AnyOldName3:
I'm not sure I agree with that, so I guess it would help to have a before and after screenshot so we can do a comparison.

By the way, can you figure out how to make it show the number of dynamic drawables and statesets? I thought we had that information, but apparently we don't. Bonus points if it's red when non-zero. I think this would help with what Foal's doing.

elsid:
Before:
before
After:
after

AnyOldName3:
I definitely prefer the line breaks in those two screenshots, but I meant a before and after for just the line break removal, not adding all the settings. The line breaks have an unfair advantage as there's less information to make it huge.

Also, you might want to check if your build is working properly. That's the OSG default font, and our stats are supposed to use DejaVu Sans Mono.

elsid:
Put empty lines back:
screenshot002

AnyOldName3:
Yeah, I like the empty lines.

elsid:
[Wiki](https://wiki.openmw.org/index.php?title=Development_Environment_Setup#OSG_plugins) is missing mention of `BUILD_OSG_PLUGIN_FREETYPE=1` that's why I had default font. I didn't know this is required.

akortunov:
Added info to Wiki.

akortunov:
Did you check if stats window works properly with low resolution (such as 600x800)?

AnyOldName3:
800x600 is *particularly* low, so I'm not sure it's worth supporting. 1366x768 is still a pretty common screen size in laptops, though, so it's reasonably likely we'll need to know what it looks like there. Someone also might try running OpenMW on an HD Ready TV, so 1280x720 might be worth testing, too.

Regarding the wiki missing Freetype, I think it's because it's technically optional. We have graceful fallbacks for when it's not present

elsid:
Seems to be working:
640x480:
640x480
800x600:
800x600
1024x768:
1024x768

akortunov:
Perosonally I do not like that displayed stats lose their vertical alignment and start to overlap with profiler (for any resolution excepts of 1920x1080 for me). Is it possible to split them to two tabs, or do something else to avoid it?

akortunov:
> Regarding the wiki missing Freetype, I think it's because it's technically optional.

We started do use DejaVuLGCSansMono.ttf in the profiler not a very long time ago. Prior to that we did not need to build a FreeType plugin for OSG, and we just forgot to update Wiki.

AnyOldName3:
Don't we need it for TrueType UI fonts, too?

akortunov:
No, they are handled by MyGUI, not by OSG.

psi29a:
Thanks @elsid

PR #2867, opened 2020-05-22
Decrease game time limit per frame to 1/24 (bug #5431)

elsid:
See detailed report in https://gitlab.com/OpenMW/openmw/-/issues/5431 . Not sure about exact value but for now 1/24 is working good enough.

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -754,7 +754,7 @@ void OMW::Engine::go()
     {
         double dt = frameTimer.time_s();
         frameTimer.setStartTick();
-        dt = std::min(dt, 0.2);
+        dt = std::min(dt, 1.0 / 24.0);
 
         mViewer->advance(simulationTime);
 


wareya:
There have been long discussions about issues closely related to this in the past.

Ideally the game would fall back to actual delta time in these situations (I think it doesn't and that's the root of this problem?), but slowdown at a reasonable framerate is the next best thing.

elsid:
Decreasing the value down to 1/60 will make the game a slow motion when FPS is capped by graphics and less than 60.
1/24: https://streamable.com/yc2jmd
1/60: https://streamable.com/0u6twj

akortunov:
And what happens if the engine can not handle 24 FPS?

elsid:
> And what happens if the engine can not handle 24 FPS?

Slow motion with a slide show happens. Not sure how the game can be playable at <24 FPS with the same experience as at 24-60 FPS. I mean probably we should make higher priority for user experience at 24-60 FPS.

akortunov:
> Not sure how the game can be playable at <24 FPS

Just for reference - modders for original engine consider 10-15 FPS to be a target framerate for Morrowind ("you do not need 60FPS, it is not a shooter"), and optimize their mods to get this FPS. And many player still setup graphics settings in OpenMW to get the same performance, just because they used to play Morrowind with this FPS.

akortunov:
Also about Morrowind - from what I can tell, it does not have a dedicated physics system and makes raytracing on scene graph to detect collision. It tries to get as many FPS as it can, when FPS cap is disabled. I doubt that it even bothers to keep physics FPS consistent, or starts to make several sweep tests for every actor every frame.

AnyOldName3:
And Ubisoft tried claiming that Assassin's Creed ran at 24 FPS on purpose as it was more cinematic when they couldn't reach 60, and objects fall through the floor in Skyrim at 61 FPS. We shouldn't be aiming to only match what vanilla could do when we can do something objectively better.

akortunov:
> We shouldn't be aiming to only match what vanilla could do when we can do something objectively better.

I'd not consider a slo-mo at 24 FPS to be a something objectively better. I suppose we need a more performant movement solver in any case.

wareya:
24fps as in barely running at 24fps is serious motion sickness territory for mouse-controlled first person games. I would hate for a game to insist on going lower before it accepts that it has to slow down.

> I'd not consider a slo-mo at 24 FPS (when in physics sytem the time just passes slower than in the rest of engine) to be a something objectively better. I suppose we need a more performant movement solver in any case.

Better performance is always desirable, but it's not an ultimate solution. No matter what you do you will always have to fall back to either delta time or slowdown in degenerate cases.

OpenMW's catchup system is already very unusual and bold, there's no reason for it to continue insisting to try to catch up below 20fps or 15fps.

The slowdown during degenerate cases is going to be the same whatever you set this threshold to. You want the threshold to be lower enough compared to 60fps that tiny performance spikes don't cause slowdown, but you definitely want it to be high enough that the game has a reasonable framerate during degenerate cases. 5fps is way too low. I would probably go with 20, but 24 is fine.

akortunov:
> 5fps is way too low.

That 5 FPS is just a failsafe, which should allow user to leave buggy area, make a save and send a bugreport, etc. No one considers that user with play with 5FPS, so it is OK to slowdown a simulation here.
But the 15-20 FPS is still playable for many users, so the simulation still should be real-time here. That video with 1/24 is not how the game should behave with 24FPS.

As I said earlier, it would be better to do not touch this variable and just use the dt = 1/24 instead of 1/60, if you think that the current time delta is an overkill for physics system.

elsid:
> That 5 FPS is just a failsafe, which should allow user to leave buggy area, make a save and send a bugreport, etc.

Why failsafe can't be at 1, 10, 15, 20, 24 FPS? Why 5? As I showed in the bug report it's not about buggy areas.

> That video with 1/24 is not how the game should behave with 24FPS

How it should behave, what's the difference? Show. I see the difference on video with frame limit 1/60 but not on the video with 1/24 from what we have now.

> just use the dt = 1/24 instead

This will lower physics precision. Basically you vote to give priority for players who play at <24 FPS, not 60.

wareya:
>That 5 FPS is just a failsafe, which should allow user to leave buggy area, make a save and send a bugreport, etc. No one considers that user with play with 5FPS, so it is OK to slowdown a simulation here.

It is not that type of failsafe. PC games will always run into situations where they have to either slow down or use delta time. You're barking up the wrong tree.

>But the 15-20 FPS is still playable for many users, so the simulation still should be real-time here.

I don't think that's not how this system works. That 15~20FPS is going to happen in the same exact situations as the game currently goes down to 5FPS, under the same performance stresses. This system is basically a cap on the runaway performance effects of trying to catch up with physics simulation that can't keep up. The threshold doesn't affect how fast the physics runs or at what performance threshold it gives up, just how much time it has to be running too slowly to accept that it can't catch up. The minimum physics framerate is still 60fps. It just takes less time for it to accept that it's not maintaining 60fps. Otherwise it would not behave as described.

>As I said earlier, it would be better to do not touch this variable and just use the dt = 1/24 instead of 1/60, if you think that the current time delta is an overkill for physics system.

This just moves the problem. It doesn't solve anything.

wareya:
Having said that, I think it's definitely concerning that, when you set this to 1/60th of a second, the game slows down if it's running at less than 60fps. It makes me think that the system being touched here is buggy or missing support for something basic. I would look into it, but I can't build OpenMW right now (the dependency setup script silently fails in my system).

kcat:
> Why failsafe can't be at 1, 10, 15, 20, 24 FPS? Why 5?

It was more or less arbitrary. There's no correct value, just one that attempts to balance the known pros and cons. It can certainly be raised, but I think 24 may be a bit too high with current performance levels (a playable 30fps isn't terribly uncommon with lower end systems or one where the bells and whistles are turned up; 24 isn't very far from that).

> I think it's definitely concerning that, when you set this to 1/60th of a second, the game slows down if it's running at less than 60fps. It makes me think that the system being touched here is buggy or missing support for something basic.

It's likely that 60 physics frames per second is overkill, or at least, our use of Bullet to do physics 60 times a second is not efficient enough. The physics death spiral in itself isn't an unknown problem in games, but I imagine most games have it well hidden due to physics being efficient enough and having a suitable frame limit to allow it to recover easily. Modern engines can also use lower physics rates and frame interpolation to keep a high visual frame rate with a lower number of physics/logic updates.

wareya:
The physics framerate affects input lag and "game feel", so it should be reasonably high. I think that going significantly below 60fps would be dangerous.

>The physics death spiral in itself isn't an unknown problem in games, but I imagine most games have it well hidden due to physics being efficient enough and having a suitable frame limit to allow it to recover easily.

Having this problem this bad really isn't normal at all. Games generally just either slow down (and thus indeed have to have very tightly optimized physics) or fall back to delta time instead of trying to catch up forever. A little catchup is normal, but not enough to cause a physics death spiral. Stories like "I spawned 25482345837 crate gibs in half-life 2 and my framerate plummeted to 0.1 frames per second!!!" come from pathological performance problems in general, not runaway death spirals.

PR #2868, opened 2020-05-22
Give new opcodes to old functions made custom

Capostrophic:
Makes them use not yet utilized opcodes in segment 5 instead of the old opcodes that segment 5 functions aren't supposed to use. Should fix the assert elsid pointed at [here](https://github.com/OpenMW/openmw/pull/2841#issuecomment-632749667).

Also updated the documentation.


zinnschlag:
Don't forget to remove the old opcodes from the documentation (components/interpreter/docs/vmformat.txt).

Capostrophic:
Right... Updated.

PR #2870, opened 2020-05-24
[WIP] Mutable ESM records (#2798)

Assumeru:
Since rot provided a nice list, I'll just copy the things that I'm sure need doing.
- [x] actor record, when using AI parameter functions like SetFight
- [x] actor records, when using AddSpell, RemoveSpell
- [x] script records, when editing local variables with "scriptID".localvar
- [ ] AddItem for containers
- [x] AddItem for non-unique actors

Levelled lists have been in for years and I haven't checked how faction reactions and weather are handled... but I think those also work? ~~Leaving `script records, when editing local variables with "scriptID".localvar`. I know that syntax works, but does it work like it's supposed to?~~

### DIFF
--- a/apps/openmw/mwscript/globalscripts.cpp
+++ b/apps/openmw/mwscript/globalscripts.cpp
@@ -299,6 +299,15 @@ namespace MWScript
         return iter->second->mLocals;
     }
 
+    const Locals* GlobalScripts::getLocalsIfPresent (const std::string& name) const
+    {
+        std::string name2 = ::Misc::StringUtils::lowerCase (name);
+        auto iter = mScripts.find (name2);
+        if (iter==mScripts.end())
+            return nullptr;
+        return &iter->second->mLocals;
+    }
+
     void GlobalScripts::updatePtrs(const MWWorld::Ptr& base, const MWWorld::Ptr& updated)
     {
         MatchPtrVisitor visitor(base);
--- a/apps/openmw/mwscript/globalscripts.hpp
+++ b/apps/openmw/mwscript/globalscripts.hpp
@@ -82,6 +82,8 @@ namespace MWScript
             ///< If the script \a name has not been added as a global script yet, it is added
             /// automatically, but is not set to running state.
 
+            const Locals* getLocalsIfPresent (const std::string& name) const;
+
             void updatePtrs(const MWWorld::Ptr& base, const MWWorld::Ptr& updated);
             ///< Update the Ptrs stored in mTarget. Should be called after the reference has been moved to a new cell.
     };
--- a/apps/openmw/mwscript/locals.cpp
+++ b/apps/openmw/mwscript/locals.cpp
@@ -1,4 +1,5 @@
 #include "locals.hpp"
+#include "globalscripts.hpp"
 
 #include <components/esm/loadscpt.hpp>
 #include <components/esm/variant.hpp>
@@ -33,15 +34,25 @@ namespace MWScript
         if (mInitialised)
             return false;
 
-        const Compiler::Locals& locals =
-            MWBase::Environment::get().getScriptManager()->getLocals (script.mId);
+        const Locals* global = MWBase::Environment::get().getScriptManager()->getGlobalScripts().getLocalsIfPresent(script.mId);
+        if(global)
+        {
+            mShorts = global->mShorts;
+            mLongs = global->mLongs;
+            mFloats = global->mFloats;
+        }
+        else
+        {
+            const Compiler::Locals& locals =
+                MWBase::Environment::get().getScriptManager()->getLocals (script.mId);
 
-        mShorts.clear();
-        mShorts.resize (locals.get ('s').size(), 0);
-        mLongs.clear();
-        mLongs.resize (locals.get ('l').size(), 0);
-        mFloats.clear();
-        mFloats.resize (locals.get ('f').size(), 0);
+            mShorts.clear();
+            mShorts.resize (locals.get ('s').size(), 0);
+            mLongs.clear();
+            mLongs.resize (locals.get ('l').size(), 0);
+            mFloats.clear();
+            mFloats.resize (locals.get ('f').size(), 0);
+        }
 
         mInitialised = true;
         return true;


Capostrophic:
> does it work like it's supposed to?

no

Assumeru:
Adding it to the list then.

Capostrophic:
You may need to revise container behavior too, see UESP notes on [AddItem](https://en.uesp.net/wiki/Tes3Mod:AddItem).

Assumeru:
And actor behaviour, looking at that. Beth sure came up with some interesting functionality...

apps/openmw/mwscript/aiextensions.cpp
```diff
@@ -277,6 +279,7 @@ namespace MWScript
                     MWMechanics::Stat<int> stat = ptr.getClass().getCreatureStats(ptr).getAiSetting(mIndex);
                     stat.setModified(value, 0);
                     ptr.getClass().getCreatureStats(ptr).setAiSetting(mIndex, stat);
+                    ptr.getClass().setBaseAISetting(ptr.getCellRef().getRefId(), mIndex, value);
```
akortunov, line 282: Not sure if it works in this way in Morrowind. From what I can tell, setFight and so on can affect all aready spawned actors with given ID - if I remebered correctly, Morrowind just does not copy base AI stats from ESM record.

Capostrophic: According to rot SetFight and so on don't affect actors with the given ID that have already been loaded, they only affect the calling actor and actors that weren't loaded yet.

akortunov: Then just make sure that it works properly for guards when player refuses to surrender. (wjen `setfight 100` in dialogue makes all guards in sight hostile. So far we use a hack - if an NPC with "guard" class fights player, we force other guards attack player too, what has side effects.

Capostrophic: Yeah. Morrowind doesn't have that.

Assumeru: Guard AI in vanilla is pretty inconsistent (in my experience anyway). I don't think it should be part of this PR though.

akortunov: Two man use-cases of SetFight on base actor records:
1. When you need to make every actor with given ID hostile (Morrowind.esm uses it for guards dialogues).
2. When you need to make every actor with given ID *non*-hostile (some mods use it to make for example, Dagoth creatures non-hostile when you join Sixth House).

These cases are the main reason to modify AI settings for base actors. They should work in the way that users and content creator would consider to be correct. I'd suggest to ask modmakers about how these cases are supposed to work, rather just claim them to be inconsistent and move along.

akortunov:
Also maybe it is worth to implement different tasks in different PR's. For example, containers are not really related to AddSpell or SetFight.

apps/openmw/mwscript/aiextensions.cpp
Assumeru: All I'm saying is that guard AI is inconsistent and that it has nothing to do with mutating base records. Refusing to surrender does not `setfight 100` (it's commented out). Wearing Ordinator armor or having a bounty over 5000 does use `setfight` though.

From what little testing I've done, there appears to be no difference between implicit and explicit calls. (Doesn't mean there aren't any but... it's a positive sign.)

Assumeru:
Yeah, splitting between actor and non-actor stuff might be wise.

psi29a:
Just a comment... if you are going to be touching the for loops, could you perhaps clean them up ta bit?

```
                for (std::vector<std::string>::const_iterator iter (ref->mBase->mSpells.mList.begin());
                    iter!=ref->mBase->mSpells.mList.end(); ++iter)
```

`                for (auto iter (ref->mBase->mSpells.mList.begin()); iter!=ref->mBase->mSpells.mList.end(); ++iter)`

apps/openmw/mwclass/npc.cpp
```diff
@@ -235,9 +235,12 @@ namespace
...
             attributes[i] = npcStats.getAttribute(i).getBase();
 
-        std::vector<std::string> spells = MWMechanics::autoCalcNpcSpells(skills, attributes, race);
-        for (std::vector<std::string>::iterator it = spells.begin(); it != spells.end(); ++it)
-            npcStats.getSpells().add(*it);
+        if (!spellsInitialised)
+        {
+            std::vector<std::string> spells = MWMechanics::autoCalcNpcSpells(skills, attributes, race);
+            for (std::vector<std::string>::iterator it = spells.begin(); it != spells.end(); ++it)
+                npcStats.getSpells().add(*it);
```
psi29a, line 242: This could be transformed into:
```
        for (auto & spell : spells)
            npcStats.getSpells().add(spell);
```

apps/openmw/mwclass/npc.cpp
```diff
@@ -389,16 +384,30 @@ namespace MWClass
...
+                // race powers
+                const ESM::Race *race = MWBase::Environment::get().getWorld()->getStore().get<ESM::Race>().find(ref->mBase->mRace);
+                for (std::vector<std::string>::const_iterator iter (race->mPowers.mList.begin());
+                    iter!=race->mPowers.mList.end(); ++iter)
+                {
+                    if (const ESM::Spell* spell = MWBase::Environment::get().getWorld()->getStore().get<ESM::Spell>().search(*iter))
+                        data->mNpcStats.getSpells().add (spell);
+                    else
+                        Log(Debug::Warning) << "Warning: ignoring nonexistent race power '" << *iter << "' on NPC '" << ref->mBase->mId << "'";
+                }
```
psi29a, line 398:
```
            for (const auto & iter : race->mPowers.mList)
            {
                if (const ESM::Spell* spell = MWBase::Environment::get().getWorld()->getStore().get<ESM::Spell>().search(iter))
                    data->mNpcStats.getSpells().add (spell);
                else
                    Log(Debug::Warning) << "Warning: ignoring nonexistent race power '" << iter << "' on NPC '" << ref->mBase->mId << "'";
            }
```

psi29a:
Just a something I noticed that you can really clean up the readability of the code you are already touching anyway. Take advantage of modern C++ ;D

psi29a:
gcc wasn't happy about that...

```
CMakeFiles/openmw_test_suite.dir/__/openmw/mwworld/esmstore.cpp.o: In function `void __gnu_cxx::new_allocator<MWMechanics::SpellList>::construct<MWMechanics::SpellList, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int&>(MWMechanics::SpellList*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int&)':
esmstore.cpp:(.text._ZN9__gnu_cxx13new_allocatorIN11MWMechanics9SpellListEE9constructIS2_JRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERiEEEvPT_DpOT0_[_ZN9__gnu_cxx13new_allocatorIN11MWMechanics9SpellListEE9constructIS2_JRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERiEEEvPT_DpOT0_]+0x60): undefined reference to `MWMechanics::SpellList::SpellList(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int)'
collect2: error: ld returned 1 exit status
```

Assumeru:
Right... so that's a pretty ridiculous problem. Can't the unit tests just pull in the apps/openmw cmake list so I don't have to manually tell it to build every (transitively) included cpp file?

elsid:
Run CMake with [`BUILD_UNITTESTS=ON`](https://github.com/OpenMW/openmw/blob/master/CMakeLists.txt#L13).

Assumeru:
That's really not what I mean. Anyway, sleepy time, g'night.

elsid:
Probably you mean to setup build targets properly. For now there is no library build target including `../openmw/mwmechanics/spelllist.cpp`, only binary includes this cpp: https://github.com/OpenMW/openmw/blob/master/apps/openmw/CMakeLists.txt#L105 . And it's not possible to link binary to binary. So we need to build a library and then link it to tests binary. Correct me if I'm wrong but I don't know other way to setup dependencies.

Assumeru:
I think that's what I want, but I know just about nothing about CMake. The way unit tests are built strikes me as pretty fragile at the moment - with the upside of reduced build time, of course.

psi29a:
```
/home/travis/build/OpenMW/openmw/apps/openmw_test_suite/mwworld/test_store.cpp: In constructor ‘MWMechanics::SpellList::SpellList(const string&, int)’:
/home/travis/build/OpenMW/openmw/apps/openmw_test_suite/mwworld/test_store.cpp:14:83: error: class ‘MWMechanics::SpellList’ does not have any field named ‘mSpellsChanged’
   14 |     SpellList::SpellList(const std::string& id, int type) : mId(id), mType(type), mSpellsChanged(false) {}
      |                                                                                   ^~~~~~~~~~~~~~
```

and

```
/home/travis/build/OpenMW/openmw/apps/openmw/mwmechanics/spells.cpp:435:13: warning: add explicit braces to avoid dangling else [-Wdangling-else]
3169            else
3170            ^
31711 warning generated.
```

Assumeru:
Nearly there I think. Still kinda need to sync actors with the base record as I don't think instances can have their own spell lists at all.

psi29a:
How does this correspond to your list of check boxes in the OP?

Assumeru:
The current changes are working towards the second checkbox.

psi29a:
```
/home/travis/build/OpenMW/openmw/apps/openmw_test_suite/mwworld/test_store.cpp:14:57: error: invalid use of incomplete type ‘class MWMechanics::SpellList’
   14 |     SpellList::SpellList(const std::string& id, int type) : mId(id), mType(type) {}
      |                                                         ^
In file included from /home/travis/build/OpenMW/openmw/apps/openmw_test_suite/mwworld/test_store.cpp:10:
/home/travis/build/OpenMW/openmw/./apps/openmw/mwworld/esmstore.hpp:18:11: note: forward declaration of ‘class MWMechanics::SpellList’
   18 |     class SpellList;
      |           ^~~~~~~~~
```

psi29a:
radioactive ;)

Assumeru:
> radioactive ;)

But can we have Capo sing the song for the release trailer? :D

psi29a:
I do not remember merging this explicitly...  I did merge #2980 though.

Assumeru:
I guess it's technically correct since this PR is a subset of the other one at the moment. If it doesn't reopen automatically when I get started on the last item, I'll just make a new one.

Assumeru:
**Container notes**

Explicit Add/Remove item does not appear to affect the base record
Adding/removing items from containers gives them inventory data; taking everything adds a 1 byte MNAM subrecord with value 0
Implicit Add/Remove item affects the base record
Implicit Additem affects containers with custom non-empty inventories in the active cell (cells probably)
Implicit add/remove item does not work on empty containers

akortunov:
> Explicit Add/Remove item does not appear to affect the base record

There is an another system. In Morrowind all containers or actors with the same ID use shared inventory, so Additem/RemoveItem affects all instances, including newly-spawned ones. It is also possible to add levelled items here.
In some cases (e.g. when player takes an item from container) object can get an unique inventory, so Add/Remove affect only one instance of object. It is possible to add levelled lists here, but they are not replaced to real items (may be fixed by MCP).

Assumeru:
This is what the MCP has to say:

> AddItem with levelled items
> ---
> Adds support for levelled items to AddItem. The target reference for AddItem should be a reference to a clone, not a base object.
> 
> The game stores the levelled lists in a base object, and cloning the object applies the levelled randomization. You can identify if a reference is cloned or not by selecting it with the console. A base object has a plain id (foronir), while a clone has an eight digit hexadecimal suffix (foronir00000000).
> 
> For containers, the clone happens when the player looks inside a container, and importantly, the container is de-cloned when it is empty. Adding anything to a base container will affect every container of that type, therefore you have to be careful adding levelled items to container references. Use a unique container to avoid this problem altogether.

The impression I was getting was that the shared inventory is just a temporary view of the base record (i.e., levelled lists are resolved) that only gets saved once a change is made to it.

The note on merchants is interesting. I'll definitely need to do some testing in that regard. Also negative quantities are always pretty odd.

PR #2871, opened 2020-05-24
Don't "disable" player's collision shape in TCL (#5435)

Capostrophic:
[Bug report](https://gitlab.com/OpenMW/openmw/-/issues/5435)

This allows player character themselves to keep not colliding with anything when the character moves on their own, but anything that tries to collide with the player (actor, projectile or a door) will still collide with the player as if they have collisions enabled. This means fighting actors can hurt the player character.

Note that since unlike vanilla we don't disable collisions for literally every actor in TCL handling actors will try to avoid colliding with the player unless it isn't possible, and it may lead to oddities (e.g. actors may jump in place). Nothing game-breaking though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
     Bug #5424: Creatures do not headtrack player
     Bug #5425: Poison effect only appears for one frame
     Bug #5427: GetDistance unknown ID error is misleading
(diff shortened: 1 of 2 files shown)


akortunov:
> and it may lead to oddities 

And sometimes it just behave really weird. More likely there will be bugreports about it.

PR #2873, opened 2020-05-26
NIFLoader adjustments

Capostrophic:
Some cleanup and fixes here and there. May need some more testing with mods to make sure it doesn't reintroduce optimization issues that were previously dealt with awkwardly (e.g. the mod from [bug 3950](https://gitlab.com/OpenMW/openmw/-/issues/3950) and abot's boats mod still work fine in my testing) and doesn't introduce new optimization issues.

1. Particle nodes are allowed to have NiKeyframeControllers/NiPathControllers. Seeing as they're generic NiAVObjects at core and Greatness7 stated all nodes get the same treatment when it comes to transformations it's most likely unwise to optimize for the unlikely case it crashes the original engine by theoretically crashing in some situations as well. If they don't actually have such a controller the redundant MatrixTransform node will be optimized out later.
2. Node controllers are handled in one place, avoiding code duplication for geometry nodes, allowing NiGeometry nodes to support NiRollController assigned to them directly and also fixing duplication of "unhandled controller" warnings that could happen in rare cases. This leads us to the next point:
3. Since NiVisController doesn't actually animate the node or "touch" it, redundant nodes that it controls can be safely optimized out. Same goes for NiUVController, NiGeomMorpherController, etc. Controllers that update node transformation like NiPathController, NiRollController and NiKeyframeController were granted the honor of preserving their nodes and the scene graph under them in a slightly more universal way.
4. Removed more NiTriShape-NiTriStrips-related code duplication I introduced some time ago, all Node records have controllers, there's no need to cast to a geometry record type to recover the controller.
5. NiParticleColorAffector doesn't need to have NiColorData anymore (a junk NiParticleColorAffector could trigger debug asserts and/or trigger null pointer exceptions).

### DIFF
--- a/components/nifosg/nifloader.cpp
+++ b/components/nifosg/nifloader.cpp
@@ -496,14 +496,6 @@ namespace NifOsg
 
             switch (nifNode->recType)
             {
-            case Nif::RC_NiAutoNormalParticles:
-            case Nif::RC_NiRotatingParticles:
-                // Leaf nodes in the NIF hierarchy, so won't be able to dynamically attach children.
-                // No support for keyframe controllers (just crashes in the original engine).
-                if (nifNode->trafo.isIdentity())
-                    node = new osg::Group;
-                dataVariance = osg::Object::STATIC;
-                break;
             case Nif::RC_NiBillboardNode:
                 dataVariance = osg::Object::DYNAMIC;
                 break;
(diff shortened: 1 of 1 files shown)


psi29a:
Thanks @Capostrophic, I really like your quality of work and explanation.

PR #2874, opened 2020-05-26
Cleanup WindowManager

akortunov:
Summary of changes:
1. Fix a TODO - do not store player-related data in the WindowManager itself, since these fields are used only to pass them to the CharacterCreation. 
I moved them to the CharacterCreation, what allowed to simplify API (there is no need to move maps now) and reduce code duplication. Interesting that mPlayerName and mPlayerRaceId were not used at all.
2. Rename WindowManager::onFrame() to the WindowManager::update() to make it consistent with other managers.
3. Move TextColours initialization to the separate TexColours::loadColours() method.

### DIFF
--- a/apps/openmw/mwgui/charactercreation.cpp
+++ b/apps/openmw/mwgui/charactercreation.cpp
@@ -105,23 +105,32 @@ namespace MWGui
         mGenerateClassSpecializations[0] = 0;
         mGenerateClassSpecializations[1] = 0;
         mGenerateClassSpecializations[2] = 0;
+
+        // Setup player stats
+        for (int i = 0; i < ESM::Attribute::Length; ++i)
+            mPlayerAttributes.emplace(ESM::Attribute::sAttributeIds[i], MWMechanics::AttributeValue());
+
+        for (int i = 0; i < ESM::Skill::Length; ++i)
+            mPlayerSkillValues.emplace(ESM::Skill::sSkillIds[i], MWMechanics::SkillValue());
     }
 
     void CharacterCreation::setValue (const std::string& id, const MWMechanics::AttributeValue& value)
     {
-        if (mReviewDialog)
+        static const char *ids[] =
         {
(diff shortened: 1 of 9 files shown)


apps/openmw/mwgui/charactercreation.cpp
```diff
@@ -87,6 +87,10 @@ namespace MWGui
     CharacterCreation::CharacterCreation(osg::Group* parent, Resource::ResourceSystem* resourceSystem)
         : mParent(parent)
         , mResourceSystem(resourceSystem)
+        , mPlayerMajorSkills()
+        , mPlayerMinorSkills()
+        , mPlayerAttributes()
+        , mPlayerSkillValues()
```
Capostrophic, line 93: These members are class-based (std::map, std::vector). Do they actually need to be default-initialized? I'm not sure.

apps/openmw/mwgui/charactercreation.cpp
```diff
@@ -105,23 +109,36 @@ namespace MWGui
         mGenerateClassSpecializations[0] = 0;
         mGenerateClassSpecializations[1] = 0;
         mGenerateClassSpecializations[2] = 0;
+
+        // Setup player stats
+        for (int i = 0; i < ESM::Attribute::Length; ++i)
+        {
+            mPlayerAttributes.insert(std::make_pair(ESM::Attribute::sAttributeIds[i], MWMechanics::AttributeValue()));
```
Capostrophic, line 116: You may want to use emplace instead of insert(std::make_pair()) while you're at it.

apps/openmw/mwgui/charactercreation.cpp
akortunov: More likely they are not. I just took an original initialization from WindowManager.

psi29a:
Looks, good. I'll take it. Thanks @akortunov

PR #2876, opened 2020-05-26
Reenable weapon animation lower body anim blending in first person view (#5441)

Capostrophic:
While it fixes things that are broken in vanilla, it also breaks things that aren't broken in vanilla, as evident from [the report](https://gitlab.com/OpenMW/openmw/-/issues/5441), so we should follow vanilla blending behavior to avoid further discrepancy-caused issues from cropping up.

This makes sure hit lower body animation can blend with the weapon animation so first person view behaves closer to third person view. Points 2 and 3 from bug 4575 rearise, but they're actually vanilla animation issues we should just faithfully carry over instead of trying to work around.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
     Bug #5416: Junk non-node records before the root node are not handled gracefully
     Bug #5424: Creatures do not headtrack player
     Bug #5427: GetDistance unknown ID error is misleading
+    Bug #5441: Enemies can't push a player character when in critical strike stance
     Feature #5362: Show the soul gems' trapped soul in count dialog
 
 0.46.0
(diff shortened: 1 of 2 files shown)


akortunov:
Just for reference - it should be possible to keep advantages of both approaches, if we want.
For example, maybe it is worth to declare a new priority between Priority_Movement and Priority_Hit for 1sr-person lower-body weapon animations.

Capostrophic:
I don't want to risk theoretical mod incompatibilities.

akortunov:
> I don't want to risk theoretical mod incompatibilities.

Teoretically you risk theoretical mod incompatibilities with every theoretical change.

Also Morrowind more likely does not have things such as Priority_WeaponLowerBody and Priority_SneakIdleLowerBody at all, so possibility of mods incompatibilities is not eliminated anyway.

Capostrophic:
> possibility of mods incompatibilities is not eliminated anyway.

It's not eliminated but it's still much lower when we follow vanilla at least in the cases where we know the original behavior.

PR #2878, opened 2020-05-27
Mutate base records when editing AI settings (#2798)

Assumeru:
Split from #2870. Following akortunov's advice a bit earlier than I initially expected.

I had assumed Add and RemoveSpell would require a similar change to AI settings, this is so far from true it's not even funny. So no reason to keep this change waiting.

### DIFF
--- a/apps/openmw/mwbase/world.hpp
+++ b/apps/openmw/mwbase/world.hpp
@@ -35,6 +35,7 @@ namespace ESM
     struct Position;
     struct Cell;
     struct Class;
+    struct Creature;
     struct Potion;
     struct Spell;
     struct NPC;
@@ -374,6 +375,14 @@ namespace MWBase
             /// \return pointer to created record
 
             virtual const ESM::ItemLevList *createOverrideRecord (const ESM::ItemLevList& record) = 0;
+            ///< Write this record to the ESM store, allowing it to override a pre-existing record with the same ID.
+            /// \return pointer to created record
+
+            virtual const ESM::Creature *createOverrideRecord (const ESM::Creature& record) = 0;
+            ///< Write this record to the ESM store, allowing it to override a pre-existing record with the same ID.
+            /// \return pointer to created record
+
+            virtual const ESM::NPC *createOverrideRecord (const ESM::NPC& record) = 0;
             ///< Write this record to the ESM store, allowing it to override a pre-existing record with the same ID.
             /// \return pointer to created record
 
--- a/apps/openmw/mwclass/creature.cpp
+++ b/apps/openmw/mwclass/creature.cpp
@@ -878,4 +878,9 @@ namespace MWClass
         const MWWorld::LiveCellRef<ESM::Creature> *ref = ptr.get<ESM::Creature>();
         scale *= ref->mBase->mScale;
     }
+
+    void Creature::setBaseAISetting(const std::string& id, MWMechanics::CreatureStats::AiSetting setting, int value) const
+    {
+        MWMechanics::setBaseAISetting<ESM::Creature>(id, setting, value);
+    }
 }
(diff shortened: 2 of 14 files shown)


psi29a:
```
In file included from /home/travis/build/OpenMW/openmw/apps/openmw/mwrender/npcanimation.cpp:35:0:
2817/home/travis/build/OpenMW/openmw/apps/openmw/mwrender/../mwmechanics/actorutil.hpp: In function ‘void MWMechanics::setBaseAISetting(const string&, MWMechanics::CreatureStats::AiSetting, int)’:
2818/home/travis/build/OpenMW/openmw/apps/openmw/mwrender/../mwmechanics/actorutil.hpp:20:19: error: ‘MWBase’ has not been declared
2819         T copy = *MWBase::Environment::get().getWorld()->getStore().get<T>().find(id);
2820                   ^~~~~~
2821/home/travis/build/OpenMW/openmw/apps/openmw/mwrender/../mwmechanics/actorutil.hpp:20:74: error: expected primary-expression before ‘>’ token
2822         T copy = *MWBase::Environment::get().getWorld()->getStore().get<T>().find(id);
2823                                                                          ^
2824/home/travis/build/OpenMW/openmw/apps/openmw/mwrender/../mwmechanics/actorutil.hpp:20:76: error: expected primary-expression before ‘)’ token
2825         T copy = *MWBase::Environment::get().getWorld()->getStore().get<T>().find(id);
2826                                                                            ^
2827/home/travis/build/OpenMW/openmw/apps/openmw/mwrender/../mwmechanics/actorutil.hpp:38:9: error: ‘MWBase’ has not been declared
2828         MWBase::Environment::get().getWorld()->createOverrideRecord(copy);
2829         ^~~~~~
2830apps/openmw/CMakeFiles/openmw.dir/build.make:127: recipe for target 'apps/openmw/CMakeFiles/openmw.dir/mwrender/npcanimation.cpp.o' failed
```

Capostrophic:
Probably shouldn't be put to actorutil header either.

Assumeru:
Interestingly it builds just fine for me locally but I'm not sure why it would. Will fix... tomorrow.

> Probably shouldn't be put to actorutil header either.

Move it to the cpp then? Or did you have something else in mind?

zinnschlag:
You are missing an include: ../mwbase/environment.hpp

Capostrophic:
Too much stuff depends on actorutil.hpp, I don't want compilation times to be affected, move it to a dedicated header maybe.

Assumeru:
Makes sense.

> You are missing an include: ../mwbase/environment.hpp

I know, I just find it concerning that my local build doesn't seem to care. At all.

zinnschlag:
Something stuck in the build system maybe? Did you try a full rebuild?

zinnschlag:
It seems that we are using this function only for two types (ESM::Creature, ESM::NPC). Sounds like a classical case for explicit template instantiation (http://www.cplusplus.com/articles/1C75fSEw/). Wow, that is a C++ feature that is rarely used.

Assumeru:
I told it to Clean but it still worked afterwards. And then I just ended up changing the code instead of worrying too much. Guess I'll be deleting the folder and rerunning the CI script soon.

> explicit template instantiation

As a way to keep the code in actorutil but keep compilation time down or as an addition to the current version?

zinnschlag:
> As a way to keep the code in actorutil but keep compilation time down

Correct.

Assumeru:
> Wow, that is a C++ feature that is rarely used.

VS Code can't even seem to figure out how to highlight it:
image

psi29a:
Cool, great work @Assumeru and thanks @akortunov and @zinnschlag for reviewing and helping.

psi29a:
Unless I'm mistaken, this only partially solves https://gitlab.com/OpenMW/openmw/-/issues/2798 so no need for changelog entry yet.  That should be done in #2878 correct?

Assumeru:
You are correct in that it only partially solves that issue. If I don't end up finishing the rest before the 0.47 release I think it might be worthwhile to create a new issue and put that in the changelog. No point worrying about that now though.

PR #2880, opened 2020-05-29
Remove redundant variables from the RenderingManager

akortunov:
Summary of changes:
1. Terrain system itself already tracks if it renders cells borders or if it is not. There is no point to track it in the RenderingManager too.
2. mDistantLand field is not used outside the constructor, so there is no need to keep it.

### DIFF
--- a/apps/openmw/mwrender/renderingmanager.cpp
+++ b/apps/openmw/mwrender/renderingmanager.cpp
@@ -213,10 +213,8 @@ namespace MWRender
         , mUnderwaterIndoorFog(Fallback::Map::getFloat("Water_UnderwaterIndoorFog"))
         , mNightEyeFactor(0.f)
         , mDistantFog(false)
-        , mDistantTerrain(false)
         , mFieldOfViewOverridden(false)
         , mFieldOfViewOverride(0.f)
-        , mBorders(false)
     {
         resourceSystem->getSceneManager()->setParticleSystemMask(MWRender::Mask_ParticleSystem);
         resourceSystem->getSceneManager()->setShaderPath(resourcePath + "/shaders");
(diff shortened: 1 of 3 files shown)


Capostrophic:
Item 1 looks reasonable, but:
> Some of them can be static.

They'll still end up being initialized permanently so there isn't much of a difference and no real readability improvement. You're not really removing them, you're just moving them to a different place in memory. They might get used somewhere else or modified from somewhere else (like scripting) later, this makes it impossible to do so. Are there any other downsides to them being kept member variables?

akortunov:
> They might get used somewhere else or modified from somewhere else (like scripting) later, this makes it impossible to do so. 

A quite strange complaint, since static variables is a common way to handle settings and GMST's in OpenMW.

> Are there any other downsides to them being kept member variables?

Such fallback values logically do not belong to the RenderingManager, and they are not supposed to be changed during runtime (OpenMW does not have a way to gracefully handle their changes ( as for settings) so far).

But if you do not like static variables here, I can turn them to common local variables. A couple of searches in map per frame will not add a noticable overhead.

akortunov:
Any other suggestions or complaints?

elsid:
> A quite strange complaint, since static variables is a common way to handle settings and GMST's in OpenMW.

Do you plan eventually have all GMST to be stored as static locals? Of course the more such code we have the more weight has this argument that it's everywhere like this. But it doesn't mean it's a proper way to do this. Is there any other argument that justifies this approach? Static locals add a synchronization overhead: https://godbolt.org/z/hSxdXc . Also they add a lot of complications for unit tests.

> Such fallback values logically do not belong to the RenderingManager

But the data they have are used by the RenderingManager. This pr does not remove dependency it just moves and hides it. Without these GMSTs RenderingManager wouldn't work as intendend.

> and they are not supposed to be changed during runtime

State can be constant during object lifetime but it doesn't mean the object can't be explicitly initialized with the data.

akortunov:
> Do you plan eventually have all GMST to be stored as static locals?

We already usually store them as static locals, including new code. For example, there is over one hundred of static locals just for GMST's in OpenMW. We do not store everything as fields.
And I suppose that if we ever need to add a support for runtime changes for GMST's, fallback values from Morrowind.ini, etc., we will just read them every time from related storages (with some performance loss) instead of introducing hundreds of new data fields with consistency tracking.

As for fog colors, I suppose that I just will move related data fields to the separate FogManager in the follow-up PR instead.

PR #2882, opened 2020-05-29
Refactor list of variables in the WorldManager

akortunov:
Summary of changes:
1. Move fields declarations to the single block, as we do in other places, rather than mix them with methods declarations in a random way.
2. Unify the way to init WorldManager fields.

### DIFF
--- a/apps/openmw/mwworld/worldimp.hpp
+++ b/apps/openmw/mwworld/worldimp.hpp
@@ -77,13 +77,13 @@ namespace MWWorld
 
     class World final: public MWBase::World
     {
+        private:
             Resource::ResourceSystem* mResourceSystem;
 
             std::vector<ESM::ESMReader> mEsm;
             MWWorld::ESMStore mStore;
             LocalScripts mLocalScripts;
             MWWorld::Globals mGlobalVariables;
-            bool mSky;
 
             ESM::Variant* mGameHour;
             ESM::Variant* mDaysPassed;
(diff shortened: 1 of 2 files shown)


apps/openmw/mwworld/worldimp.hpp
```diff
@@ -103,25 +103,39 @@ namespace MWWorld
...
-            World (const World&);
-            World& operator= (const World&);
-
             int mActivationDistanceOverride;
 
+            std::string mStartCell;
+
+            float mSwimHeightScale;
+
+            float mDistanceToFacedObject = -1;
```
Capostrophic, line 122: I think this syntax:
```cpp
float mDistanceToFaceObject{-1.f};
```
is preferred in modern C++. List initialization is safer. In this case you do narrowing because "-1" is an integer value, copy initialization doesn't reject it, list initialization does. It's probably for the best to use list initialization for *any* type.

kcat: FWIW, that's called uniform initialization, not list initialization. So called because it's a uniform syntax to initialize a POD type, array, struct, or call a constructor. I would definitely prefer it either way, though, since it protects against narrowing, and won't otherwise do implicit conversions, which can be a source of bugs.

AnyOldName3: `=` is more obvious and readable, so I'm not sure this is a situation where the modern, safer way of doing things is beneficial.

akortunov: I suppose we need to define somewhere in contributing guidelines a way to init fields. So far OpenMW uses at least 4 different ways to init fields ("=", ":", "{}" and in constructor). As for uniform initialization, personally I do not like to use it for simple fields - it looks ugly.

akortunov: I rewrote this PR to use an old-style initialization for all WorldManager fields.

psi29a:
Thanks!

PR #2883, opened 2020-05-30
Handle NiLines (feature #5445)

Capostrophic:
An implementation of [this](https://gitlab.com/OpenMW/openmw/-/issues/5445) based on what is known about the record, using LINES primitives in OpenSceneGraph.

The way geometry is handled in NIF loader has been made even more generic.
No Bullet-side changes, i.e. there won't be any autogenerated (so called face/triangle) collision for the lines, because there aren't any triangles or faces to speak of.

This is how the test model I was given looks like:
screenshot024
Which looks pretty similar to how it looks in SceneImmerse.

It'll be difficult to test the original behavior as in Morrowind NiLines objects can only be loaded with the most recent MWSE 2.0 build which patches their loading. NifSkope doesn't properly support NiLines records either.


AnyOldName3:
Looks okay

PR #2884, opened 2020-05-30
Update spell effects during death animation (#5403)

Capostrophic:
Should resolve [this](https://gitlab.com/OpenMW/openmw/-/issues/5403).

Effects will now be updated while the actor hasn't yet departed this world completely and can still be inflicted with a spell or an enchantment while the character is falling down like in Morrowind. This allows visual effects to properly play out. Of course you can't hit a character with an on-strike enchantment because of how physical combat is handled.
Health bar won't be shown if the player casts a harmful spell on a dying actor to prevent oddities.
Effects caused by constant effect items will be updated on dying actors too.

May horribly break something, there were troubles with fixes for a similar issue earlier. :man_shrugging:


psi29a:
I'll take this, thanks.

PR #2885, opened 2020-06-01
Fix tests build with BULLET_USE_DOUBLES=ON

elsid:


apps/openmw_test_suite/CMakeLists.txt
```diff
@@ -1,59 +1,60 @@
...
 
+if (BULLET_USE_DOUBLES)
+    add_definitions(-DBT_USE_DOUBLE_PRECISION)
 endif()
+
+if (MSVC)
+    if (CMAKE_CL_64)
+        # Debug version of openmw_unit_tests needs increased number of sections beyond 2^16
+        # just like openmw and openmw-cs
+        set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /bigobj")
```
AnyOldName3, line 58: This should probably be `add_target_compile_options_debug` or whatever so that it can't infect the other targets.

AnyOldName3:
Also, it might be a good idea to change things like `add_definitions` to the target-specific versions

apps/openmw_test_suite/CMakeLists.txt
elsid: This pr is not about fixing all possible problems. Only introduced change here is an indentation. Because `if (GTEST_FOUND AND GMOCK_FOUND)` is removed.

elsid: This pr is not about fixing all possible problems. Only introduced change here is an indentation. Because `if (GTEST_FOUND AND GMOCK_FOUND)` is removed.

elsid:
This pr is about fixing specific problem with tests build when `BULLET_USE_DOUBLES=ON`. It doesn't work as intendent because `BT_USE_DOUBLE_PRECISION` define is not added for tests target. I will remove all other changes. CMake refactoring should be done in a different pr.

AnyOldName3:
I don't think it's ever the right thing to do to just tweak disgusting code when the proper implementation should be just as simple. It's even less of a good idea here as you're expanding the set of situations where the potential footgun exists. You're leaving the code in a messier state than you found it, and as far as I can tell, there are no mitigating factors that could justify doing so.

elsid:
This pr is an additon to https://github.com/OpenMW/openmw/pull/2848. Without this change `BULLET_USE_DOUBLES` option is not a complete feature. Let's just remove it then to avoid confusion and ask developers directly add `-DBT_USE_DOUBLE_PRECISION` to compiler flags. I don't see other way how to add a compiler definition to a build target. I'm not CMake expert so for me it's just some code that does stuff. I don't have a vision of a proper solution. If you have one then just suggest. Probably I misunderstood
> `add_definitions` to the target-specific versions

I thought it's `target_compile_definitions`. No? What is it exactly?

AnyOldName3:
I think the force push confused the tab I had open. The current diff is acceptable. It would be even better if:

* The `-D` could go, as it's redundant (leading `-D`s are discarded) and has a slim chance of confusing a future person if they think it's given directly to the compiler and a future compiler has different syntax. This is minor and requires no extra thinking, so should just happen.
* We could make it a *public* definition on something that the tests link with (ideally Bullet itself, but I guess `components` would work as an alternative if there's no actual target for Bullet) so it gets applied transitively, and also to anything else that needs it. If that had happened in the first place, we wouldn't need something applying it directly to the tests. This would fix the cause rather than the symptom.

psi29a:
I think it is good enough to remove -D and then we merge.

elsid:
Removed `-D`, applied definition only for components target as `PUBLIC`.

psi29a:
Thank you @elsid

PR #2886, opened 2020-06-01
Store AIPackages as unique_ptr

elsid:
Allows to avoid manual memory management. Also raw pointer in return type has ambiguous ownership transfer.


psi29a:
Thanks!

PR #2887, opened 2020-06-03
Introduce a separate class to control world date and time

akortunov:
The main idea - encapsulate datetime handling in the separate DateTimeManager.

It allows us to simplify World API (-5 public functions, -5 fields, -160 lines in the worldimp.cpp).
Note that public API to change date and time was not really used - we either incremented it every frame or changed them via global variables.

Note that ESM namespace really had two TimeStamp types, so I renamed one to the EpochTimeStamp to avoid confusion.

### DIFF
--- a/apps/openmw/mwworld/datetimemanager.cpp
+++ b/apps/openmw/mwworld/datetimemanager.cpp
@@ -1,0 +1,227 @@
+#include "datetimemanager.hpp"
+
+#include "../mwbase/environment.hpp"
+#include "../mwbase/world.hpp"
+
+#include "esmstore.hpp"
+#include "globals.hpp"
+#include "timestamp.hpp"
+
+namespace
+{
+    static int getDaysPerMonth(int month)
+    {
+        switch (month)
+        {
+            case 0: return 31;
+            case 1: return 28;
+            case 2: return 31;
+            case 3: return 30;
+            case 4: return 31;
+            case 5: return 30;
+            case 6: return 31;
+            case 7: return 31;
+            case 8: return 30;
+            case 9: return 31;
+            case 10: return 30;
+            case 11: return 31;
+        }
+
+        throw std::runtime_error ("month out of range");
+    }
+}
+
+namespace MWWorld
+{
+    void DateTimeManager::setup(Globals& globalVariables)
+    {
+        mGameHour = globalVariables["gamehour"].getFloat();
+        mDaysPassed = globalVariables["dayspassed"].getInteger();
+        mDay = globalVariables["day"].getInteger();
+        mMonth = globalVariables["month"].getInteger();
+        mYear = globalVariables["year"].getInteger();
+        mTimeScale = globalVariables["timescale"].getFloat();
+    }
+
+    void DateTimeManager::setHour(double hour)
+    {
+        if (hour < 0)
+            hour = 0;
+
+        int days = static_cast<int>(hour / 24);
+        hour = std::fmod(hour, 24);
+        mGameHour = static_cast<float>(hour);
+
+        if (days > 0)
+            setDay(days + mDay);
+    }
+
+    void DateTimeManager::setDay(int day)
+    {
+        if (day < 1)
+            day = 1;
+
+        int month = mMonth;
+        while (true)
+        {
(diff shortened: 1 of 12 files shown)


apps/openmw/mwworld/datetimemanager.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "../mwbase/environment.hpp"
+#include "../mwbase/world.hpp"
+
+#include "esmstore.hpp"
+#include "globals.hpp"
+#include "timestamp.hpp"
+
+namespace MWWorld
+{
+    int getDaysPerMonth(int month)
```
elsid, line 12: Put declaration into header or move into anonymous namespace/make static to avoid possible ODR violation.

zinnschlag:
Hm, overall the idea is good. Even now that I am not actively working on the code base anymore the current state of World is still giving me headaches. These subsystem facade classes were meant to be exactly that: facades, that delegate to other classes without implementing much logic themself.

However there is one aspect of this implementation that I am not convinced of. You have stripped the time interface out of World. That makes the code that is working with time more convoluted (one additional get functions). Maybe keep the World interface and just delegate to the new time class instead?

The problem with World is that it has too much implementation. Having a lot of interface on the other hand is not a problem. That is what facade classes are designed for.

akortunov:
> The problem with World is that it has too much implementation. Having a lot of interface on the other hand is not a problem. That is what facade classes are designed for.

As for me, I think that the real problem of World that it tries to wrap pretty everything (the whole rendering system, physics, navigation, etc). Such components usually are facades too, so probably it would be better to make such components independent in the future. In this case World will be responsible only for game world state (objects in cells, weather, game time handling, etc) as a wrapper for dedicated classes that handle low-level logic.

> Maybe keep the World interface and just delegate to the new time class instead?

Drawbacks of existing date setting API:
1. It is not used outside the World itself.
2. Even if someone will decide to use it, it may lead to errors due to desync with related global variables.

So far we update datetime in two cases:
1. Every frame when we advance time
2. When scripts or player (via colsole) manually change values of related global variables.

We need a some kind of way to unify the API to avoid desync. A simplest way to do it is allow to change current date and time outside of DateManager itself only via global variables.

akortunov:
> That makes the code that is working with time more convoluted (one additional get functions).

Sorry, I did not get what you mean. An old API:
```
profile.mInGameTime.mGameHour = world.getTimeStamp().getHour();	
profile.mInGameTime.mDay = world.getDay();	
profile.mInGameTime.mMonth = world.getMonth();	
profile.mInGameTime.mYear = world.getYear();
```
A new one:
```
profile.mInGameTime = world.getEpochTimeStamp();
```
There were 4 different functions to get datetime components and 3 ones to set them (and get a desync with global wariables, so they were not used outside the World itself). Now there is just 2 functions to get timestamps.

zinnschlag:
Ah, must have misread that. i find statements like 
`
const ESM::EpochTimeStamp& currentDate = MWBase::Environment::get().getWorld()->getEpochTimeStamp()
`

really confusing. While it is technically correct, using a reference here suggests that the function also returns a reference (and therefore exposing an internal variable via another indirection). Just a simple const ESM::EpochTimeStamp or even ESM::EpochTimeStamp instead of const ESM::EpochTimeStamp& would be much clearer. Never mind then.

I agree about not exposing unnecessary API. About the World facade: I am pretty sure that we once had renderer and physics a separate subsystems. We moved away from that, because they are supposed to be kept in sync with the world state automatically (meaning without code directly manipulating physics or graphics).

akortunov:
> Just a simple const ESM::EpochTimeStamp or even ESM::EpochTimeStamp instead of const ESM::EpochTimeStamp& would be much clearer.

Fixed.

> We moved away from that, because they are supposed to be kept in sync with the world state automatically (meaning without code directly manipulating physics or graphics).

Well, you know better then. I just do not like that World slowly starts to do everything since most of apps/openmw works with either rendering, physics, navigation or cells. So it is the main reason why the World in the state that gives headaches, and it slowly becomes worse from release to release.

zinnschlag:
I guess we have to agree to disagree on this point. Anyway, this PR looks ready for merging.

PR #2888, opened 2020-06-04
Mark not changing AiPackages fields as const

elsid:
To make clear what can be changed in each AI package state.

### DIFF
--- a/apps/openmw/mwmechanics/aicast.cpp
+++ b/apps/openmw/mwmechanics/aicast.cpp
@@ -10,12 +10,22 @@
 #include "creaturestats.hpp"
 #include "steering.hpp"
 
-MWMechanics::AiCast::AiCast(const std::string& targetId, const std::string& spellId, bool manualSpell)
-    : mTargetId(targetId), mSpellId(spellId), mCasting(false), mManual(manualSpell), mDistance(0)
+namespace MWMechanics
 {
-    ActionSpell action = ActionSpell(spellId);
-    bool isRanged;
-    mDistance = action.getCombatRange(isRanged);
+    namespace
+    {
+        float getInitialDistance(const std::string& spellId)
+        {
+            ActionSpell action = ActionSpell(spellId);
+            bool isRanged;
+            return action.getCombatRange(isRanged);
+        }
+    }
+}
+
+MWMechanics::AiCast::AiCast(const std::string& targetId, const std::string& spellId, bool manualSpell)
+    : mTargetId(targetId), mSpellId(spellId), mCasting(false), mManual(manualSpell), mDistance(getInitialDistance(spellId))
+{
 }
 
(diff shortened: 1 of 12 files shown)


apps/openmw/mwmechanics/aicast.cpp
```diff
@@ -10,12 +10,22 @@
 #include "creaturestats.hpp"
 #include "steering.hpp"
 
+namespace MWMechanics
```
akortunov, line 13: I wonder why you use two namespaces here.

elsid: `ActionSpell` is inside this namespace. And I didn't want to have it explicitly used with this type name. OpenMW codebase is not consistent with usage of namespaces. This file explicitly mentions `MWMechanics` namespace for each `AiCast` method definition like `MWMechanics::AiCast::AiCast`. Other files has definitions inside `namespace` block. I expect this block to have more functions so there is no need to repeat `MWMechanics` each time.

elsid: Also OpenMW [style guide](https://wiki.openmw.org/index.php?title=Code_Formatting_Conventions) about indentation inside namespace block does not allow to change this file with minumum changes to make it consistent. Every non empty line will get an additional 4 spaces.

akortunov: I still did not get what the 
```
namespace MWMechanics
{
    namespace
    {
    }
}
```
construction is supposed to do. If getInitialDistance() is supposed to be in the MWMechanics namespace, why just do not use it:

```
namespace MWMechanics
{
    float getInitialDistance(const std::string& spellId)
    {
        ActionSpell action = ActionSpell(spellId);
        bool isRanged;
        return action.getCombatRange(isRanged);
    }
}

If there is a some kind of C++ quirk involved here, this quirk is not intuitive.
```

Fanael: It's an [unnamed namespace](https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces), all items within one are local to the translation unit. Since it's a top-level function, making it `static` would have the same effect.

elsid: `getInitialDistance` should be inside anonymous namespace since it's not declared in any header:
* https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities
* https://google.github.io/styleguide/cppguide.html#Unnamed_Namespaces_and_Static_Variables

Otherwise this may cause [ODR](https://en.cppreference.com/w/cpp/language/definition) violation if there will be definition of the same function (name and parameters types, not the body) in a different translation unit. Also this allows to inline a function and don't generate code that may be called from a different translation unit. Here https://godbolt.org/z/hLAA5i with `-O3` function in a named namespace has generated code since it could be called from a different translation unit, but a function inside unnamed namespace has no generated code because it's not used.

PR #2889, opened 2020-06-05
Add a separate stats watcher for GUI

akortunov:
Our current approach to update player's stats in GUI is to track them in the `MechanicsManager::update()` and notify `WindowManager` via a lot of setters, what is a bit weird.

A new approach is to provide the `StatsWatcher` class, which is supposed to track player's stats and notify its listeners (stats window, chargen window and HUD so far) when needed.

Note: I did not find any reasons to watch stats in the MechanicsManager, so I moved watching to the WindowManager update, but I can move it back if needed.

### DIFF
--- a/apps/openmw/mwgui/statswatcher.cpp
+++ b/apps/openmw/mwgui/statswatcher.cpp
@@ -1,0 +1,209 @@
+#include "statswatcher.hpp"
+
+#include "../mwbase/environment.hpp"
+#include "../mwbase/windowmanager.hpp"
+#include "../mwbase/world.hpp"
+
+#include "../mwmechanics/npcstats.hpp"
+#include "../mwmechanics/spellutil.hpp"
+
+#include "../mwworld/class.hpp"
+#include "../mwworld/esmstore.hpp"
+#include "../mwworld/inventorystore.hpp"
+
+namespace MWGui
+{
+    // mWatchedTimeToStartDrowning = -1 for correct drowning state check,
+    // if stats.getTimeToStartDrowning() == 0 already on game start
+    StatsWatcher::StatsWatcher()
+      : mWatchedLevel(-1), mWatchedTimeToStartDrowning(-1), mWatchedStatsEmpty(true)
+    {
+    }
+
+    void StatsWatcher::watchActor(const MWWorld::Ptr& ptr)
+    {
+        mWatched = ptr;
+    }
+
+    void StatsWatcher::update()
+    {
+        if (mWatched.isEmpty())
+            return;
+
+        MWBase::WindowManager *winMgr = MWBase::Environment::get().getWindowManager();
+        const MWMechanics::NpcStats &stats = mWatched.getClass().getNpcStats(mWatched);
+        for (int i = 0;i < ESM::Attribute::Length;++i)
+        {
+            if (stats.getAttribute(i) != mWatchedAttributes[i] || mWatchedStatsEmpty)
+            {
+                std::stringstream attrname;
+                attrname << "AttribVal"<<(i+1);
+
+                mWatchedAttributes[i] = stats.getAttribute(i);
+                setValue(attrname.str(), stats.getAttribute(i));
+            }
+        }
+
+        if (stats.getHealth() != mWatchedHealth || mWatchedStatsEmpty)
+        {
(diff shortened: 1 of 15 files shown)


apps/openmw/mwgui/statswatcher.hpp
```diff
@@ -0,0 +1,68 @@
...
+        virtual void setValue(const ESM::Skill::SkillEnum parSkill, const MWMechanics::SkillValue& value) {};
+        virtual void configureSkills(const std::vector<int>& major, const std::vector<int>& minor) {};
+    };
+
+    class StatsWatcher
+    {
+        MWWorld::Ptr mWatched;
+
+        MWMechanics::AttributeValue mWatchedAttributes[8];
+        MWMechanics::SkillValue mWatchedSkills[27];
```
elsid, line 31: Same as above, but `ESM::Skill::Length`.

apps/openmw/mwgui/statswatcher.cpp
```diff
@@ -0,0 +1,209 @@
...
+    // mWatchedTimeToStartDrowning = -1 for correct drowning state check,
+    // if stats.getTimeToStartDrowning() == 0 already on game start
+    StatsWatcher::StatsWatcher()
+      : mWatchedLevel(-1), mWatchedTimeToStartDrowning(-1), mWatchedStatsEmpty(true)
+    {
+    }
+
+    void StatsWatcher::watchActor(const MWWorld::Ptr& ptr)
+    {
+        mWatched = ptr;
```
elsid, line 25: Should this method set `mWatchedStatsEmpty` to true if ptr is empty? Otherwise update will not notify when other ptr is started to be watched until attribute change.

apps/openmw/mwgui/statswatcher.hpp
```diff
@@ -0,0 +1,68 @@
...
+    class StatsListener
+    {
+    public:
+        /// Set value for the given ID.
+        virtual void setValue(const std::string& id, const MWMechanics::AttributeValue& value) {};
+        virtual void setValue(const std::string& id, const MWMechanics::DynamicStat<float>& value) {};
+        virtual void setValue(const std::string& id, const std::string& value) {};
+        virtual void setValue(const std::string& id, int value) {};
+        virtual void setValue(const ESM::Skill::SkillEnum parSkill, const MWMechanics::SkillValue& value) {};
+        virtual void configureSkills(const std::vector<int>& major, const std::vector<int>& minor) {};
```
elsid, line 23: Don't need to put semicolon after function body.

apps/openmw/mwgui/statswatcher.cpp
akortunov: Maybe. I just moved existing code.

apps/openmw/mwgui/statswatcher.hpp
```diff
@@ -0,0 +1,68 @@
...
+        std::string mWatchedRace;
+        std::string mWatchedClass;
+
+        int mWatchedLevel;
+
+        float mWatchedTimeToStartDrowning;
+
+        bool mWatchedStatsEmpty;
+
+        std::list<StatsListener*> mListeners;
```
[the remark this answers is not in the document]
akortunov: > why have you chosen list?

I just took an existing implementation from OICS.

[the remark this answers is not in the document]
akortunov: Use std::set now.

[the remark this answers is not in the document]
AnyOldName3: Any particular reason for `std::set` over `std::unordered_set`? If we don't need to iterate over it in-order, it *should* be a smidge faster, and significantly faster if it gets big.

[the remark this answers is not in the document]
akortunov: Keep in mind that we will not get a situation when we will need to store thousands of widgets in the container and container's implementation will start to matter. There are only three listeners so far, and this container is not a bottleneck.

apps/openmw/mwgui/statswatcher.cpp
elsid: I think we can leave this as is. Unlikely that we will find some bug because of this.

PR #2890, opened 2020-06-06
Do not store object position, if it is the same as in CellRef

akortunov:
We store object position in two places - in the ESM::CellRef (initially we read it from ESM/ESP file and treat as an initial object position) and in the ObjectState (a current position). In many cases (for containers and activators, for example) both coordinats sets are the same.

So the suggested approach - do not store the current position, if it is the same as an initial one (which we store in save as ESM::CellRef record). 
Such approach made my testing saves about 6% lesser without any noticable difference in the loading speed. But I suspect that for new games the difference will be smaller since we do not store data about not-looted-yet containers.

### DIFF
--- a/components/esm/defs.hpp
+++ b/components/esm/defs.hpp
@@ -57,6 +57,26 @@ struct Position
 };
 #pragma pack(pop)
 
+bool inline operator== (const Position& left, const Position& right) noexcept
+{
+    return left.pos[0] == right.pos[0] &&
+           left.pos[1] == right.pos[1] &&
+           left.pos[2] == right.pos[2] &&
+           left.rot[0] == right.rot[0] &&
+           left.rot[1] == right.rot[1] &&
+           left.rot[2] == right.rot[2];
+}
+
+bool inline operator!= (const Position& left, const Position& right) noexcept
+{
+    return left.pos[0] != right.pos[0] ||
+           left.pos[1] != right.pos[1] ||
+           left.pos[2] != right.pos[2] ||
+           left.rot[0] != right.rot[0] ||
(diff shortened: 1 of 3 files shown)


components/esm/defs.hpp
```diff
@@ -54,6 +54,26 @@ struct Position
     {
         return osg::Vec3f(pos[0], pos[1], pos[2]);
     }
+
+    bool inline operator== (const Position& position) const
```
elsid, line 58: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c86-make--symmetric-with-respect-to-operand-types-and-noexcept

akortunov: Just for reference: there is a lot of asymmetric comparison operators in OpenMW.

components/esm/objectstate.hpp
```diff
@@ -45,7 +45,7 @@ namespace ESM
         {}
 
         /// @note Does not load the CellRef ID, it should already be loaded before calling this method
-        virtual void load (ESMReader &esm);
+        virtual void load (ESMReader &esm, bool inInventory = false);
```
elsid, line 48: Not sure was it intended but this method and load from all derived classes has different declarations now.
```c++
CreatureStats c;
ObjectState* o = &c;
o->load(esm, true); // will not call CreatureStats::load
```
I'd mark all load and save functions in derived from ObjectState classes with `override` or `final` to verify this virtual before applying this pr.

akortunov:
Ok, an additional investigation shows that CellRef position is empty for objects in containers during loading, so it should be possible to use it as a fallback for inventory items too to avoid uninitialized random values.

PR #2899, opened 2020-06-08
Track magic effect index for summoning

akortunov:
Related to [bug #5207](https://gitlab.com/OpenMW/openmw/-/issues/5207).
Despite it is possible to add a given magic effect only once to spell or enchantment, it is possible to add it many times via editor, and Morrowind supports it just well.

Unfortunately, the whole OpenMW's magic system was designed that there is a magic effect can be added only once.

This PR is supposed to change it. So far I managed to track (and save) effect index in active spells, so it it possible, for example, summon several skeletons with the single spell.
But it is unclear how to handle permanent spells so far.
"Spells" collection merges constant effects with the same ID internally, so multiple summons with the same ID are out of question for constant spells so far.

### DIFF
--- a/apps/openmw/mwgui/container.cpp
+++ b/apps/openmw/mwgui/container.cpp
@@ -252,7 +252,7 @@ namespace MWGui
                     }
 
                     // Clean up summoned creatures as well
-                    std::map<MWMechanics::CreatureStats::SummonKey, int>& creatureMap = creatureStats.getSummonedCreatureMap();
+                    std::map<ESM::SummonKey, int>& creatureMap = creatureStats.getSummonedCreatureMap();
                     for (const auto& creature : creatureMap)
                         MWBase::Environment::get().getMechanicsManager()->cleanupSummonedCreature(mPtr, creature.second);
                     creatureMap.clear();
--- a/apps/openmw/mwgui/spellicons.cpp
+++ b/apps/openmw/mwgui/spellicons.cpp
@@ -25,8 +25,8 @@
 namespace MWGui
 {
 
-    void EffectSourceVisitor::visit (MWMechanics::EffectKey key,
-                                           const std::string& sourceName, const std::string& sourceId, int casterActorId,
+    void EffectSourceVisitor::visit (MWMechanics::EffectKey key, int effectIndex,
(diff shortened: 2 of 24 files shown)


akortunov:
I added support for summoning via constant spell, and found a bug - when `creatureStats.getSpells().purgeEffect(effectId, sourceId)` is used, it removes not only given effect, but all follow up effects from the same spell. This bug is reproducable in upstream master as well.

akortunov:
I have a suspicion that Spells merges effects with the same ID internally, so there is always only one entry for given effect per spell.

akortunov:
The problem with purging should be solved now.

The main question is what to do with constant spells - I see no way to disable merging without major rewrite, so multiple sommons with same ID for constant spells may be out of question.

Also I encountered a bug when my save failed to load, I'll take a look later.

akortunov:
Needs a new round of testing, especially with older saves, which already have summons.

zinnschlag (commented):
I do not see anything obviously wrong with this PR. But this needs thorough testing, indeed. Probably more thorough than the devs alone can handle. Just going ahead with the merge so we can get a nightly build for wider testing may be the way to go.

I noticed one minor detail that could be improved (definitely not a blocker). The code is using a tuple (int, std::string, in) in multiple places. Turning a pair into a tuple here is the easiest solution. But it does not provide the best code quality.

What I would have done instead is to define a struct with three public members. This makes the code more clear by giving the members names (instead of just an index) and it also avoids having the data structure defined multiple times.

akortunov:
> Turning a pair into a tuple here is the easiest solution. But it does not provide the best code quality.

I did not want to add a struct because it insreases PR size a lot.

>  looks like the mentioned problem with constant spells is still present?

Yes, it is, and I see no way to fix it without major rewrite of magic effects system.
But honestly, I can not remember any cases when a permanent spell is used to summon someone (since you can not summon an another actor when summon your dies).
Maybe it would be enough to create a separate low-priority bugreport for this issue.

akortunov:
@psi29a, what do you think?

psi29a:
I'll take it, but please do create a ticket the constant/permanent spell task?

akortunov:
Done.

PR #2900, opened 2020-06-09
Use FourCC to declare all ESM record names

akortunov:
Since FourCC struct was added, a `RecNameInts` enum was not fully upgraded to use it for some reason.


psi29a:
Thanks @akortunov and @elsid

PR #2901, opened 2020-06-09
Minor shader fixes

Capostrophic:
- simpleWater bool is not set to false by default anymore. Uniforms are initialized to zero by default which corresponds to false, so this shouldn't be necessary, and GL4ES doesn't support initializers for uniforms right now and currently has to manually do the same thing done here.
- Most objects in the game only have a diffuse map and don't have specularity, but while handling the diffuse map is obviously necessary, the per-fragment specularity calculations in the shaders used for all objects should be avoided whenever possible. Per suggestion of akortunov, getSpecular call isn't done for objects that have a black specular color in their material as well as terrain. This seems to result in a ~10% higher framerate on a GeForce GTX 1050 when shaders are used (of course it depends on whether shadows and water shader are used), while on older video cards it may not lead to discernible benefit but certainly shouldn't lead to a discernible loss either. Should probably resolve [5428](https://gitlab.com/OpenMW/openmw/-/issues/5428), as there isn't much that can be done otherwise; I thought about making specular lighting per-vertex when the entire lighting is per-vertex, but given that there are at least four situations that should be taken care of  — per-vertex material specularity, per-vertex specular map specularity (shouldn't be a thing), per-pixel material specularity, per-pixel specular map specularity — it has proven to complicate shaders much more than what I'm comfortable with. Maybe later.

### DIFF
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -49,7 +49,7 @@ uniform vec2 envMapLumaBias;
 uniform mat2 bumpMapMatrix;
 #endif
 
-uniform bool simpleWater = false;
+uniform bool simpleWater;
 
 varying float euclideanDepth;
 varying float linearDepth;
@@ -181,7 +181,8 @@ void main()
         matSpec = passColor.xyz;
 #endif
 
-    gl_FragData[0].xyz += getSpecular(normalize(viewNormal), normalize(passViewPos.xyz), shininess, matSpec) * shadowing;
+    if (matSpec != vec3(0.0))
+        gl_FragData[0].xyz += getSpecular(normalize(viewNormal), normalize(passViewPos.xyz), shininess, matSpec) * shadowing;
 #if @radialFog
     float depth = euclideanDepth;
     // For the less detailed mesh of simple water we need to recalculate depth on per-pixel basis
--- a/files/shaders/terrain_fragment.glsl
+++ b/files/shaders/terrain_fragment.glsl
@@ -90,7 +90,8 @@ void main()
         matSpec = passColor.xyz;
 #endif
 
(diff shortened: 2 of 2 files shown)


akortunov:
Also this PR improves situation with [bug #4952](https://gitlab.com/OpenMW/openmw/-/issues/4912).
The `getSpecular()` is a one of three causes of terrible particle systems performance when shaders are used (other two are texture lookups when shadows are enabled and per-pixel lighting).

AnyOldName3:
Yeah, I guess the only situation where this would make things worse is on older GPUs without jump instructions, and the only extra thing would be a single multiplication by zero which the optimiser would *probably* remove as when it happened, the thing would already be being multiplied by zero. LGTM.

icecream95:
Doing viewNormal calculation inside the if statement when `!normalMap && !parallax && !PER_PIXEL_LIGHTING` could further improve things.

With GPU frequency at 100MHz:

| | |
| - | - |
| Before | 8 fps |
| This PR | 8.7 fps |
| Calculate viewNormal inside `if (!matSpec)` | 9.4 fps |
| No getSpecular | 11.6 fps |

(Most of the rest of the difference here is because all the extra uniforms force UBOs to be used instead of register uniforms)

akortunov:
> Doing viewNormal calculation inside the if statement when !normalMap && !parallax && !PER_PIXEL_LIGHTING could further improve things.

Sorry, I do not understand what you want. The `viewNormal` is actually used in normal and parallax maps, also in PPL.

Sisah2:
He probably mean that 
viewNormal = gl_NormalMatrix * normalize(passNormal);
Is always called, but its needed only if per_pixel_lighting is enabled, if there is normal/parallax/specular map or 
if (matSpec != vec3(0.0))

PR #2903, opened 2020-06-10
Use more C++11 in tools code

akortunov:
Mostly loops style changes and variables initialization (use `nullptr` instead of 0 for references, for example).

Based on #2738 (with some concers addressed), but without formatting changes (which are mostly whitespace moving).


psi29a:
Yeah, got caught up in other things so this is fine and yes also what I would have liked to have merged if it wasn't for formatting changes that crept in.

PR #2905, opened 2020-06-11
Store package type id as enum except ESM

elsid:


psi29a:
```
/Users/travis/build/OpenMW/openmw/apps/openmw/mwmechanics/aipackage.cpp:418:28: error: use of undeclared identifier 'TypeIdWander'
        if (getTypeId() == TypeIdWander)
                  
```

Capostrophic:
Right, I tried to correct the merge conflict and overlooked that the implementation file could have changes too... Can't fix it from within GitHub, I suppose elsid should handle it himself.

psi29a:
@Capostrophic you should have the ability to click on the PR changes, then on the right side of the file with changes, click 'edit file'. It's what I did. Let's see if this works.

PR #2909, opened 2020-06-15
Calculate viewNormal only when needed

akortunov:
It is used only in per-pixel lighting, also in normal, parallax and specular maps, but we always calculate it anyway, and optimizer may fail to remove that redundant calculations.

On my hardware (GTX 1050) it allows to increase FPS a bit when there are a lot of particles in scene (from 64 to 72 FPS in "Valenvaryon, Propylon Chamber").

### DIFF
--- a/files/shaders/objects_fragment.glsl
+++ b/files/shaders/objects_fragment.glsl
@@ -83,7 +83,9 @@ void main()
     mat3 tbnTranspose = mat3(normalizedTangent, binormal, normalizedNormal);
 
     vec3 viewNormal = gl_NormalMatrix * normalize(tbnTranspose * (normalTex.xyz * 2.0 - 1.0));
-#else
+#endif
+
+#if (!@normalMap && (@parallax || @forcePPL))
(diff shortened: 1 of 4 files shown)


psi29a:
Cool

psi29a:
What kind of perf difference do you get on your machine?

akortunov:
> What kind of perf difference do you get on your machine?

On the scene from [#4912](https://gitlab.com/OpenMW/openmw/-/issues/4912) my lowest FPS raises from 64 to 72.

AnyOldName3:
Shader optimisers are apparently not as aggressive as I've been led to believe.

I'm irritated by the need to disable `setupShadowCoords`, though. It's already preprocessed away to nothingness, and that seems like it should be the job of the shadow shader files rather than the regular shaders. I can't think of any tidier way of getting around its arguments not existing, though.

AnyOldName3:
In case that's unclear, I've not thought of any direct suggestions I could make and recognise it's a free performance boost, so is inevitable, but it's not pretty.

PR #2910, opened 2020-06-15
Reduce navmesh size

elsid:
By removing all vertex coordinates duplicates. Reduces navmesh cache size by half for the same scene. This happens in the background navmesh thread but there is no noticable change for initial navmesh build time.


psi29a:
Awesome!

PR #2913, opened 2020-06-16
Get rid of ECLD and dependencies

akortunov:
The "exterior cell load distance" setting is obsolete now and is not needed since it was used as a stopgap for Distant Land.
The same thing for "local map cell distance" - Distant Land does not load distant cells fully, so these cells can not be rendered on the local map anyway.

So I removed these settings and replaced them by compile-time constant (as cell size) - maybe it is not worth to fully remove an ability to load additional cells, we just do not want this thing to be abused by end users.

### DIFF
--- a/docs/source/reference/modding/settings/camera.rst
+++ b/docs/source/reference/modding/settings/camera.rst
@@ -51,21 +51,20 @@ viewing distance
 This value controls the maximum visible distance (also called the far clipping plane).
 Larger values significantly improve rendering in exterior spaces,
 but also increase the amount of rendered geometry and significantly reduce the frame rate.
-This value interacts with the exterior cell load distance setting
-in that it's probably undesired for this value to provide visibility into cells that have not yet been loaded.
-When cells are visible before loading, the geometry will "pop-in" suddenly, creating a jarring visual effect.
-To prevent this effect, this value must be less than::
+Note that when cells are visible before loading (when not using a Distant Land), the geometry will "pop-in" suddenly,
+creating a jarring visual effect. To prevent this effect, this value must be less than::
 
-	(8192 * exterior cell load distance - 1024) * 0.93
+	(CellSizeInUnits * CellGridRadius - 1024) * 0.93
 
-The constant 8192 is the size of a cell, and 1024 is the threshold distance for loading a new cell.
+The CellSizeInUnits is the size of a game cell in units (8192 by default), CellGridRadius determines how many
+neighboring cells to current one to load (1 by default - 3x3 grid), and 1024 is the threshold distance for loading a new cell.
 Additionally, the field of view setting also interacts with this setting because the view frustum end is a plane,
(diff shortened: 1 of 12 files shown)


Assumeru:
Isn't the setting required in some form for Oblivion/Skyrim support?

akortunov:
It is not related to Oblivion and Skyrim.

AnyOldName3:
I'm much keener on the documentation saying in bold and capital letters not to use it, and maybe renaming it so people following outdated guides don't set it by accident. If someone makes a new game and uses a less weird scale, a different distance will be desirable, and we can't make it a GMST yet.

Also, in the later games (which have a working LOD system), people still like setting their uGridsToLoad to higher-than-default levels, so maybe once hardware gets faster or other optimisations are implemented, provided there are mods to make any content incompatibilities go away, some people might want scripted events and actors to appear from further away.

I'm aware that both of my examples are for things that are only hypothetical right now, but I'm unconvinced there's much need for such a nuclear option.

psi29a:
believe that @akortunov has a point here, with Object Paging we now have very little need for ECLD in Morrowind. While it would be useful for later games, we're not yet there so I see little point in keeping it around. Keep in mind it hasn't been fully removed either, so we can come back to it and implement a better solution, with another name, when we do need it.

docs/source/reference/modding/settings/camera.rst
```diff
@@ -51,21 +51,20 @@ viewing distance
...
-When cells are visible before loading, the geometry will "pop-in" suddenly, creating a jarring visual effect.
-To prevent this effect, this value must be less than::
+Note that when cells are visible before loading (when not using a Distant Land), the geometry will "pop-in" suddenly,
+creating a jarring visual effect. To prevent this effect, this value must be less than::
 
-	(8192 * exterior cell load distance - 1024) * 0.93
+	(CellSizeInUnits * CellGridRadius - 1024) * 0.93
 
-The constant 8192 is the size of a cell, and 1024 is the threshold distance for loading a new cell.
+The CellSizeInUnits is the size of a game cell in untis (8192 by default), CellGridRadius determines how many
```
psi29a, line 59: untis -> units

docs/source/reference/modding/settings/cells.rst
```diff
@@ -60,7 +36,7 @@ consider increasing the number of preloading threads to 2 or 3 for a boost in pr
 Faster preloading will reduce the chance that a cell could not be completely loaded before the player moves into it,
 and hence reduce the chance of seeing loading screens or frame drops.
 This may be especially relevant when the player moves at high speed
-and/or a large number of cells are loaded in via 'exterior cell load distance'.
+and/or a large number of cells are loaded in due to large viewing distance.
```
psi29a, line 39: I would drop the 'in' here.

> and/or when large number of cells are loaded due to large viewing distance.

PR #2917, opened 2020-06-18
Fix issues, reported by CoverityScan

akortunov:
Mostly complaints about non-initialized variables.

Note that 304711 appears to be a yet-another false-positive in `atan2` function.


psi29a:
Cool, thanks and marked.

PR #2919, opened 2020-06-22
Avoid to use some deprecated Qt API

akortunov:
Summary of changes:
1. Use default constructors for flags and nullptr's instead of 0 ([task #5477](https://gitlab.com/OpenMW/openmw/-/issues/5477)).
2. Do not use deprecated `endl` from Qt streams
3. Update mouse wheel event handling - now Qt returns movement angle for both axes and we need only a vertical movement so far.


akortunov:
Or does not need testing if we are really going to drop Qt4.

PR #2923, opened 2020-06-23
removed need for split compilation for static analysis, we've gone fr…

psi29a:
…om openmw/openmw-cs/tools taking 50+ minutes to taking 30. Merging them also makes our before_script.linux.sh less complicated; removed redundant CMAKE flag; set build type to RelWithDebInfo as None has no meaning (even if it still builds).


psi29a:
Final results is in for full build of everything:
GCC-9: 49 minutes
Clang-7: 28 minutes

I assume that now things are 'cached' that times will continue to go down.

PR #2925, opened 2020-06-23
change build type to debug for ci/cd

psi29a:
it should set -O0 which should prioritise fast compilation for gcc

### DIFF
--- a/CI/before_script.linux.sh
+++ b/CI/before_script.linux.sh
@@ -17,7 +17,7 @@ ${ANALYZE} cmake \
     -DUSE_SYSTEM_TINYXML=TRUE \
     -DCMAKE_INSTALL_PREFIX="/usr" \
     -DBINDIR="/usr/games" \
-    -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+    -DCMAKE_BUILD_TYPE="DEBUG" \
     -DGTEST_ROOT="${GOOGLETEST_DIR}" \
     -DGMOCK_ROOT="${GOOGLETEST_DIR}" \
     ..


elsid:
Linux compilation on travis is already quite [fast](https://travis-ci.org/github/OpenMW/openmw/builds/700639382) (2-3 minutes) due to ccache (especially comparing to windows builds). As I understand we missed all cache after switch to Ubuntu Focal. But there is other important thing. Tests are better to be runned in release mode, because that's how game is used. So if there is UB it's more likely to be catched in optimized code by tests.

elsid:
At the and of the build there is ccache statistics usage:

Before switch to Focal: https://api.travis-ci.org/v3/job/700639385/log.txt (Ran for 3 min 2 sec)
```
ccache -s
cache directory                     /home/travis/.ccache
primary config                      /home/travis/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats zero time                     Sun Jun 21 18:53:40 2020
cache hit (direct)                  1074
cache hit (preprocessed)               2
cache miss                            10
cache hit rate                     99.08 %
called for link                       47
called for preprocessing               7
cleanups performed                     1
files in cache                      3113
cache size                         460.5 MB
max cache size                     500.0 MB
```

After (): https://api.travis-ci.org/v3/job/700915652/log.txt (Ran for 33 min 22 sec)
```
ccache -s
cache directory                     /home/travis/.ccache
primary config                      /home/travis/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats updated                       Mon Jun 22 15:19:37 2020
stats zeroed                        Mon Jun 22 14:48:05 2020
cache hit (direct)                     5
cache hit (preprocessed)               2
cache miss                          2116
cache hit rate                      0.33 %
called for link                       47
called for preprocessing            1044
cleanups performed                     0
files in cache                      4231
cache size                         572.5 MB
max cache size                       5.0 GB
```

99.08% vs  0.33 % cache hit. That's why it's slow.

But it fills up: https://api.travis-ci.org/v3/job/701148974/log.txt (Ran for 5 min 25 sec )
```
ccache -s
cache directory                     /home/travis/.ccache
primary config                      /home/travis/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats updated                       Tue Jun 23 05:50:54 2020
stats zeroed                        Tue Jun 23 05:47:38 2020
cache hit (direct)                  1043
cache hit (preprocessed)               1
cache miss                            74
cache hit rate                     93.38 %
called for link                       47
called for preprocessing              38
cleanups performed                     0
files in cache                      4783
cache size                         763.0 MB
max cache size                       5.0 GB
```

psi29a:
The problem was that when ccache fails, it fails hard and that release and relwithdebinfo can make the compile take up to 50 minutes which hits the timeout. That's because the -O3/-O2 (respectively) flag is set.  When it is ccached, then it takes about 3 minutes.

However by setting debug, the -O0 flag is sent so it will compile as fast as possible which is what we want on CI/CD (until we ant to create release builds based on tags).

psi29a:
@elsid you set the cmake build type to None which is _not_ Release (-O3) anyway.

What we can do is setup a matrix build that just compiles and run tests and can set against Release (-O3).

elsid:
> The problem was that when ccache fails, it fails hard

ccache fails hard only on compiler or os switch. This happens so rarely that we may not take this into account. And only a single build is requried to build cache again.

> @elsid you set the cmake build type to None which is not Release (-O3) anyway.

It wasn't me: https://github.com/OpenMW/openmw/commit/63665104bfdd458c7a791f579aafccebdf2b1592 . Unless you think I'm scrawl who hides under a different identity. And yes, it was build type `None` that is unknown for CMake so it didn't applied any optimization flags.

> What we can do is setup a matrix build that just compiles and run tests and can set against Release (-O3).

It's a good idea.

psi29a:
Ah, my excuses... I just went one level of indirection deep when you reformatted everything. Yeah... None is not a 'valid' option, it is only there because of cmake history.

PR #2926, opened 2020-06-24
Use QMultiMap instead of QMap

akortunov:
[Task #5476](https://gitlab.com/OpenMW/openmw/-/issues/5476).

Qt devs are aimed to encapsulate all multimap-specific functions in the QMultiMap instead of QMap, so it would be nice to update our API.
The main change is that with this PR we use `QMultiMap::insert` instead of `QMap::insertMulti` and `QMultiMap::remove + QMultiMap::insert` instead of `QMap::insert`.

Requires testing (can be tested on the launcher) since I could miss something.

### DIFF
--- a/components/config/gamesettings.cpp
+++ b/components/config/gamesettings.cpp
@@ -2,11 +2,8 @@
 #include "launchersettings.hpp"
 
 #include <QTextCodec>
-#include <QTextStream>
 #include <QDir>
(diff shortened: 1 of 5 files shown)


psi29a:
Pulling this in and exposing it to more people.

psi29a:
Thanks @akortunov

PR #2927, opened 2020-06-24
Use std types for WorkQueue

elsid:
OpenThreads is a non standard way for multithreading in C++ without much benefits and sometimes with some [problems](https://gitlab.com/OpenMW/openmw/-/merge_requests/127#note_336202938).

### DIFF
--- a/components/sceneutil/workqueue.cpp
+++ b/components/sceneutil/workqueue.cpp
@@ -2,69 +2,55 @@
 
 #include <components/debug/debuglog.hpp>
 
+#include <numeric>
+
 namespace SceneUtil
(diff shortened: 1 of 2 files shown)


psi29a:
Does this remove/replace all instances of openthreads?

psi29a:
We only moved to OpenThreads because we used to use boost::thread... to get around OSG's messed up thread affinity stuff. Now that OSG has allowed us to turn off thread affinity, there is no reason to use OpenThreads and we can use something more standard.

While this is a start, I would prefer we go further in another PR and also purge it from cmake.

PR #2928, opened 2020-06-24
Load audio asychronously (fix #4880)

elsid:
The idea is to use background thread to open voice files. The problem first was discovered in https://gitlab.com/OpenMW/openmw/-/issues/4880 . My research shows this as the only thing that may have significant impact on a single frame time in mechanics manager. This graph shows frame time for each frame before this change:
before_osg_stats

There are multiple one frame duration spikes. This happens on SSD. Digging into the problem shows the same reason as bug [4880](https://gitlab.com/OpenMW/openmw/-/issues/4880).

After the change for the same test graph doesn't have this kind of spikes:
after_osg_stats

### DIFF
--- a/apps/openmw/mwsound/openal_output.hpp
+++ b/apps/openmw/mwsound/openal_output.hpp
@@ -14,12 +14,13 @@
 
 namespace MWSound
 {
-    class SoundManager;
     class Sound;
     class Stream;
 
     class OpenAL_Output : public Sound_Output
     {
+        const DecoderProvider* mDecoderProvider;
+
         ALCdevice *mDevice;
         ALCcontext *mContext;
 
@@ -95,7 +96,7 @@ namespace MWSound
         void pauseActiveDevice() override;
         void resumeActiveDevice() override;
 
-        OpenAL_Output(SoundManager &mgr);
+        OpenAL_Output(const DecoderProvider& decoderBuilder);
         virtual ~OpenAL_Output();
     };
 #ifndef DEFAULT_OUTPUT
--- a/apps/openmw/mwsound/openal_output.hpp
+++ b/apps/openmw/mwsound/openal_output.hpp
@@ -99,9 +99,6 @@ namespace MWSound
         OpenAL_Output(const DecoderProvider& decoderBuilder);
         virtual ~OpenAL_Output();
     };
-#ifndef DEFAULT_OUTPUT
-#define DEFAULT_OUTPUT(x) ::MWSound::OpenAL_Output((x))
-#endif
 }
 
 #endif
--- a/apps/openmw/mwsound/openal_output.hpp
+++ b/apps/openmw/mwsound/openal_output.hpp
@@ -12,6 +12,11 @@
 
 #include "sound_output.hpp"
 
+namespace VFS
+{
+    class Manager;
+}
+
 namespace MWSound
 {
     class Sound;
@@ -20,6 +25,7 @@ namespace MWSound
     class OpenAL_Output : public Sound_Output
     {
         const DecoderProvider* mDecoderProvider;
+        const VFS::Manager* mVFS;
 
         ALCdevice *mDevice;
         ALCcontext *mContext;
@@ -96,7 +102,7 @@ namespace MWSound
         void pauseActiveDevice() override;
         void resumeActiveDevice() override;
 
-        OpenAL_Output(const DecoderProvider& decoderBuilder);
+        OpenAL_Output(const DecoderProvider& decoderBuilder, const VFS::Manager& vfs);
         virtual ~OpenAL_Output();
     };
 }
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -115,7 +115,7 @@ namespace MWSound
     }
 
     // Return a new decoder instance, used as needed by the output implementations
-    DecoderPtr SoundManager::getDecoder()
+    DecoderPtr SoundManager::getDecoder() const
     {
         return DecoderPtr(new DEFAULT_DECODER (mVFS));
     }
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -109,7 +109,7 @@ namespace MWSound
 
     SoundManager::~SoundManager()
     {
-        clear();
+        SoundManager::clear();
         mSoundBuffers.clear();
         mOutput.reset();
     }
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -54,7 +54,7 @@ namespace MWSound
 
     SoundManager::SoundManager(const VFS::Manager* vfs, bool useSound)
         : mVFS(vfs)
-        , mOutput(new DEFAULT_OUTPUT(*this))
+        , mOutput(new OpenAL_Output(*this))
         , mWaterSoundUpdater(makeWaterSoundUpdaterSettings())
         , mSoundBuffers(*vfs, *mOutput)
         , mListenerUnderwater(false)
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -173,7 +173,7 @@ namespace MWSound
     SoundManager::~SoundManager()
     {
         SoundManager::clear();
-        mSoundBuffers.clear();
+        mSoundBuffers.lock()->clear();
         mOutput.reset();
     }
 
@@ -487,6 +487,8 @@ namespace MWSound
         if(!mOutput->streamSound(decoder, track.get()))
             return nullptr;
 
+        track->setPlaying();
+
         Stream* result = track.get();
         const auto it = std::lower_bound(mActiveTracks.begin(), mActiveTracks.end(), track);
         mActiveTracks.insert(it, std::move(track));
@@ -513,28 +515,21 @@ namespace MWSound
         if(!mOutput->isInitialized())
             return nullptr;
 
-        Sound_Buffer *sfx = mSoundBuffers.load(Misc::StringUtils::lowerCase(soundId));
-        if(!sfx) return nullptr;
-
-        // Only one copy of given sound can be played at time, so stop previous copy
-        stopSound(sfx, MWWorld::ConstPtr());
-
         SoundPtr sound = getSoundRef();
+
         sound->init([&] {
             SoundParams params;
             params.mVolumeFactor = volume;
-            params.mSfxVolume = sfx->getVolume();
+            params.mSfxVolume = 0.0f;
             params.mBaseVolume = volumeFromType(type);
             params.mPitch = pitch;
             params.mFlags = mode | type | Play_2D;
             return params;
         } ());
-        if(!mOutput->playSound(sound.get(), sfx->getHandle(), offset))
-            return nullptr;
-
         Sound* result = sound.get();
-        mActiveSounds[MWWorld::ConstPtr()].emplace_back(std::move(sound), sfx);
-        mSoundBuffers.use(*sfx);
+
+        loadSoundAsync(MWWorld::ConstPtr(), Misc::StringUtils::lowerCase(soundId), offset, std::move(sound));
+
         return result;
     }
 
@@ -549,27 +544,19 @@ namespace MWSound
         if ((mode & PlayMode::RemoveAtDistance) && (mListenerPos - objpos).length2() > 2000 * 2000)
             return nullptr;
 
-        // Look up the sound in the ESM data
-        Sound_Buffer *sfx = mSoundBuffers.load(Misc::StringUtils::lowerCase(soundId));
-        if(!sfx) return nullptr;
-
-        // Only one copy of given sound can be played at time on ptr, so stop previous copy
-        stopSound(sfx, ptr);
-
-        bool played;
         SoundPtr sound = getSoundRef();
-        if(!(mode&PlayMode::NoPlayerLocal) && ptr == MWMechanics::getPlayer())
+
+        if (!(mode & PlayMode::NoPlayerLocal) && ptr == MWMechanics::getPlayer())
         {
             sound->init([&] {
                 SoundParams params;
                 params.mVolumeFactor = volume;
-                params.mSfxVolume = sfx->getVolume();
+                params.mSfxVolume = 0.0f;
                 params.mBaseVolume = volumeFromType(type);
                 params.mPitch = pitch;
                 params.mFlags = mode | type | Play_2D;
                 return params;
             } ());
-            played = mOutput->playSound(sound.get(), sfx->getHandle(), offset);
         }
         else
         {
@@ -577,22 +564,21 @@ namespace MWSound
                 SoundParams params;
                 params.mPos = objpos;
                 params.mVolumeFactor = volume;
-                params.mSfxVolume = sfx->getVolume();
+                params.mSfxVolume = 0.0f;
                 params.mBaseVolume = volumeFromType(type);
                 params.mPitch = pitch;
-                params.mMinDistance = sfx->getMinDist();
-                params.mMaxDistance = sfx->getMaxDist();
+                params.mMinDistance = 0.0f;
+                params.mMaxDistance = 0.0f;
                 params.mFlags = mode | type | Play_3D;
                 return params;
             } ());
-            played = mOutput->playSound3D(sound.get(), sfx->getHandle(), offset);
         }
-        if(!played)
-            return nullptr;
 
         Sound* result = sound.get();
-        mActiveSounds[ptr].emplace_back(std::move(sound), sfx);
-        mSoundBuffers.use(*sfx);
+
+        // Look up the sound in the ESM data
+        loadSoundAsync(ptr, Misc::StringUtils::lowerCase(soundId), offset, std::move(sound));
+
         return result;
     }
 
@@ -603,36 +589,37 @@ namespace MWSound
         if(!mOutput->isInitialized())
             return nullptr;
 
-        // Look up the sound in the ESM data
-        Sound_Buffer *sfx = mSoundBuffers.load(Misc::StringUtils::lowerCase(soundId));
-        if(!sfx) return nullptr;
-
         SoundPtr sound = getSoundRef();
+
         sound->init([&] {
             SoundParams params;
             params.mPos = initialPos;
             params.mVolumeFactor = volume;
-            params.mSfxVolume = sfx->getVolume();
+            params.mSfxVolume = 0.0f;
             params.mBaseVolume = volumeFromType(type);
             params.mPitch = pitch;
-            params.mMinDistance = sfx->getMinDist();
-            params.mMaxDistance = sfx->getMaxDist();
+            params.mMinDistance = 0.0f;
+            params.mMaxDistance = 0.0f;
             params.mFlags = mode | type | Play_3D;
             return params;
         } ());
-        if(!mOutput->playSound3D(sound.get(), sfx->getHandle(), offset))
-            return nullptr;
-
         Sound* result = sound.get();
-        mActiveSounds[MWWorld::ConstPtr()].emplace_back(std::move(sound), sfx);
-        mSoundBuffers.use(*sfx);
+
+        // Look up the sound in the ESM data
+        loadSoundAsync(MWWorld::ConstPtr(), Misc::StringUtils::lowerCase(soundId), offset, std::move(sound));
+
         return result;
     }
 
     void SoundManager::stopSound(Sound *sound)
     {
-        if(sound)
+        if (!sound)
+            return;
+
+        if (sound->isPlaying())
             mOutput->finishSound(sound);
+        else
+            sound->cancelLoading();
     }
 
     void SoundManager::stopSound(Sound_Buffer *sfx, const MWWorld::ConstPtr &ptr)
@@ -653,14 +640,22 @@ namespace MWSound
         if(!mOutput->isInitialized())
             return;
 
-        Sound_Buffer *sfx = mSoundBuffers.lookup(Misc::StringUtils::lowerCase(soundId));
-        if (!sfx) return;
+        std::string normalizedSoundId = Misc::StringUtils::lowerCase(soundId);
 
-        stopSound(sfx, ptr);
+        if (Sound_Buffer *sfx = mSoundBuffers.lockConst()->lookup(normalizedSoundId))
+            stopSound(sfx, ptr);
+
+        for (const auto& loading : mLoadingSounds)
+            if (loading.mPtr == ptr && loading.mSoundId == normalizedSoundId)
+                loading.mSound->cancelLoading();
     }
 
     void SoundManager::stopSound3D(const MWWorld::ConstPtr &ptr)
     {
+        for (const auto& loading : mLoadingSounds)
+            if (loading.mPtr == ptr)
+                loading.mSound->cancelLoading();
+
         SoundMap::iterator snditer = mActiveSounds.find(ptr);
         if(snditer != mActiveSounds.end())
         {
@@ -677,6 +672,10 @@ namespace MWSound
 
     void SoundManager::stopSound(const MWWorld::CellStore *cell)
     {
+        for (const auto& loading : mLoadingSounds)
+            if (!loading.mPtr.isEmpty() && loading.mPtr != MWMechanics::getPlayer() && loading.mPtr.getCell() == cell)
+                loading.mSound->cancelLoading();
+
         for(SoundMap::value_type &snd : mActiveSounds)
         {
             if(!snd.first.isEmpty() && snd.first != MWMechanics::getPlayer() && snd.first.getCell() == cell)
@@ -705,7 +704,7 @@ namespace MWSound
         SoundMap::iterator snditer = mActiveSounds.find(ptr);
         if(snditer != mActiveSounds.end())
         {
-            Sound_Buffer *sfx = mSoundBuffers.lookup(Misc::StringUtils::lowerCase(soundId));
+            Sound_Buffer *sfx = mSoundBuffers.lockConst()->lookup(Misc::StringUtils::lowerCase(soundId));
             if (sfx == nullptr)
                 return;
             for(SoundBufferRefPair &sndbuf : snditer->second)
@@ -721,13 +720,16 @@ namespace MWSound
         SoundMap::const_iterator snditer = mActiveSounds.find(ptr);
         if(snditer != mActiveSounds.end())
         {
-            Sound_Buffer *sfx = mSoundBuffers.lookup(Misc::StringUtils::lowerCase(soundId));
-            return std::find_if(snditer->second.cbegin(), snditer->second.cend(),
+            Sound_Buffer *sfx = mSoundBuffers.lockConst()->lookup(Misc::StringUtils::lowerCase(soundId));
+            const bool playing = std::find_if(snditer->second.cbegin(), snditer->second.cend(),
                 [this,sfx](const SoundBufferRefPair &snd) -> bool
                 { return snd.second == sfx && mOutput->isSoundPlaying(snd.first.get()); }
             ) != snditer->second.cend();
+            if (playing)
+                return true;
         }
-        return false;
+        return mLoadingSounds.end() != std::find_if(mLoadingSounds.begin(), mLoadingSounds.end(),
+            [&] (const LoadingSound& v) { return v.mPtr == ptr && v.mSoundId == soundId; });
     }
 
     void SoundManager::pauseSounds(BlockerType blocker, int types)
@@ -835,7 +837,7 @@ namespace MWSound
 
             bool soundIdChanged = false;
 
-            Sound_Buffer* sfx = mSoundBuffers.lookup(update.mId);
+            Sound_Buffer* sfx = mSoundBuffers.lockConst()->lookup(update.mId);
             if (mLastCell != cell)
             {
                 const auto snditer = mActiveSounds.find(MWWorld::ConstPtr());
@@ -937,7 +939,7 @@ namespace MWSound
                         mUnderwaterSound = nullptr;
                     if (sound == mNearWaterSound)
                         mNearWaterSound = nullptr;
-                    mSoundBuffers.release(*sndidx->second);
+                    mSoundBuffers.lock()->release(*sndidx->second);
                     sndidx = snditer->second.erase(sndidx);
                 }
                 else
@@ -1038,6 +1040,7 @@ namespace MWSound
 
         playAllVoicesFromCreatedDecoders();
         playMusicFromCreatedDecoder();
+        playLoadedSounds();
 
         updateSounds(duration);
         if (MWBase::Environment::get().getStateManager()->getState()!=
@@ -1130,6 +1133,10 @@ namespace MWSound
         for (auto& v : mWaitingVoice)
             if (v.mPtr == old)
                 v.mPtr = updated;
+
+        for (auto& v : mLoadingSounds)
+            if (v.mPtr == old)
+                v.mPtr = updated;
     }
 
     // Default readAll implementation, for decoders that can't do anything
@@ -1201,12 +1208,15 @@ namespace MWSound
     {
         abortAll(mWaitingVoice);
         abortAll(mWaitingMusic);
+        abortAll(mLoadingSounds);
 
         waitForAll(mWaitingVoice);
         waitForAll(mWaitingMusic);
+        waitForAll(mLoadingSounds);
 
         mWaitingVoice.clear();
         mWaitingMusic.clear();
+        mLoadingSounds.clear();
 
         SoundManager::stopMusic();
 
@@ -1215,7 +1225,7 @@ namespace MWSound
             for(SoundBufferRefPair &sndbuf : snd.second)
             {
                 mOutput->finishSound(sndbuf.first.get());
-                mSoundBuffers.release(*sndbuf.second);
+                mSoundBuffers.lock()->release(*sndbuf.second);
             }
         }
         mActiveSounds.clear();
@@ -1348,8 +1358,10 @@ namespace MWSound
         if (mWaitingMusic.empty())
             return;
 
-        if (mWaitingMusic.back().mDeadline <= std::chrono::steady_clock::now())
-            mWaitingMusic.back().mWorkItem->waitTillDone();
+        const auto now = std::chrono::steady_clock::now();
+        const auto waiting = mWaitingMusic.back();
+        if (waiting.mDeadline <= now)
+            waiting.mWorkItem->waitTillDone();
 
         const auto locked = mMusicDecoders.lock();
         const auto decoder = locked->find(mWaitingMusic.back().mFileName);
@@ -1368,8 +1380,126 @@ namespace MWSound
         } ());
         mOutput->streamSound(decoder->second, mMusic.get());
 
+        mMusic->setPlaying();
         abortAll(mWaitingMusic);
         mWaitingMusic.clear();
         locked->clear();
+    }
+
+    void SoundManager::loadSoundAsync(const MWWorld::ConstPtr& ptr, std::string soundId, float offset, SoundPtr&& sound)
+    {
+        const auto sfx = lookupSound(soundId);
+
+        if (sfx != nullptr)
+        {
+            playLoadedSound(*sfx, ptr, offset, std::move(sound));
+            return;
+        }
+
+        const auto load = makeWorkItem([this, soundId] { loadSound(soundId); });
+
+        LoadingSound loading;
+        loading.mPtr = ptr;
+        loading.mSoundId = std::move(soundId);
+        loading.mOffset = offset;
+        loading.mSound = std::move(sound);
+        loading.mWorkItem = load;
+        loading.mDeadline = std::chrono::steady_clock::now() + sAsyncOperationTimeout;
+
+        mLoadingSounds.push_back(std::move(loading));
+        mWorkQueue->addWorkItem(load);
+    }
+
+    void SoundManager::loadSound(const std::string &soundId)
+    {
+        {
+            const auto locked = mLoadedSoundBuffers.lock();
+            if (locked->count(soundId) > 0)
+                return;
+        }
+
+        Sound_Buffer* sfx = mSoundBuffers.lock()->load(soundId);
+
+        mLoadedSoundBuffers.lock()->emplace(soundId, sfx);
+    }
+
+    Sound_Buffer* SoundManager::lookupSound(const std::string &soundId) const
+    {
+        {
+            const auto locked = mLoadedSoundBuffers.lockConst();
+            if (locked->count(soundId) > 0)
+                return nullptr;
+        }
+
+        return mSoundBuffers.lockConst()->lookup(soundId);
+    }
+
+    void SoundManager::playLoadedSounds()
+    {
+        if (mLoadingSounds.empty())
+            return;
+
+        const auto now = std::chrono::steady_clock::now();
+        for (const auto& sound : mLoadingSounds)
+            if (sound.mDeadline <= now)
+                sound.mWorkItem->waitTillDone();
+
+        const auto locked = mLoadedSoundBuffers.lock();
+
+        for (auto& loading : mLoadingSounds)
+        {
+            const auto loaded = locked->find(loading.mSoundId);
+
+            if (loaded == locked->end())
+                continue;
+
+            SoundPtr sound = std::move(loading.mSound);
+
+            if (sound->isLoadCancelled())
+                continue;
+
+            Sound_Buffer* sfx = loaded->second;
+
+            if (sfx == nullptr)
+                continue;
+
+            playLoadedSound(*sfx, loading.mPtr, loading.mOffset, std::move(sound));
+        }
+
+        locked->clear();
+
+        const auto isSoundNull = [] (const LoadingSound& v) { return v.mSound == nullptr; };
+
+        mLoadingSounds.erase(std::remove_if(mLoadingSounds.begin(), mLoadingSounds.end(), isSoundNull), mLoadingSounds.end());
+    }
+
+    void SoundManager::playLoadedSound(Sound_Buffer& sfx, const MWWorld::ConstPtr& ptr, float offset, SoundPtr&& sound)
+    {
+        // Only one copy of given sound can be played at time, so stop previous copy
+        stopSound(&sfx, ptr);
+
+        sound->setSfxVolume(sfx.getVolume());
+
+        bool played;
+
+        if (sound->getIs3D())
+        {
+            sound->setMinDistance(sfx.getMinDist());
+            sound->setMaxDistance(sfx.getMaxDist());
+
+            played = mOutput->playSound3D(sound.get(), sfx.getHandle(), offset);
+        }
+        else
+        {
+            played = mOutput->playSound(sound.get(), sfx.getHandle(), offset);
+        }
+
+        if (!played)
+            return;
+
+        sound->setPlaying();
+
+        mActiveSounds[ptr].emplace_back(std::move(sound), &sfx);
+        mSoundBuffers.lock()->use(sfx);
     }
 }
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -34,6 +34,35 @@ namespace MWSound
     {
         constexpr float sMinUpdateInterval = 1.0f / 30.0f;
 
+        constexpr std::chrono::steady_clock::duration sAsyncOperationTimeout = std::chrono::milliseconds(200);
+
+        template <class Function>
+        class WorkItem final : public SceneUtil::WorkItem
+        {
+        public:
+            WorkItem(Function&& function) : mFunction(std::move(function)) {}
+
+            void doWork() final
+            {
+                if (mAborted)
+                    return;
+                mFunction();
+            }
+
+            void abort() final { mAborted = true; }
+
+        private:
+            std::atomic_bool mAborted {false};
+            Function mFunction;
+        };
+
+        template <class Function>
+        auto makeWorkItem(Function&& function)
+        {
+            using WorkItem = WorkItem<std::decay_t<Function>>;
+            return osg::ref_ptr<WorkItem>(new WorkItem(std::forward<Function>(function)));
+        }
+
         WaterSoundUpdaterSettings makeWaterSoundUpdaterSettings()
         {
             WaterSoundUpdaterSettings settings;
@@ -46,6 +75,39 @@ namespace MWSound
             settings.mNearWaterOutdoorID = Misc::StringUtils::lowerCase(Fallback::Map::getString("Water_NearWaterOutdoorID"));
 
             return settings;
+        }
+
+        template <class Waiting>
+        void abortAll(Waiting& waiting)
+        {
+            for (const auto& v : waiting)
+                v.mWorkItem->abort();
+        }
+
+        template <class Waiting>
+        void waitForAll(Waiting& waiting)
+        {
+            for (const auto& v : waiting)
+                v.mWorkItem->waitTillDone();
+        }
+
+        struct VoiceDistance
+        {
+            float mMin;
+            float mMax;
+        };
+
+        VoiceDistance makeVoiceDistance(const MWBase::World& world)
+        {
+            const auto& settings = world.getStore().get<ESM::GameSetting>();
+            const float fAudioMinDistanceMult = settings.find("fAudioMinDistanceMult")->mValue.getFloat();
+            const float fAudioVoiceDefaultMinDistance = settings.find("fAudioVoiceDefaultMinDistance")->mValue.getFloat();
+            const float fAudioMaxDistanceMult = settings.find("fAudioMaxDistanceMult")->mValue.getFloat();
+            const float fAudioVoiceDefaultMaxDistance = settings.find("fAudioVoiceDefaultMaxDistance")->mValue.getFloat();
+            VoiceDistance result;
+            result.mMin = std::max(fAudioVoiceDefaultMinDistance * fAudioMinDistanceMult, 1.0f);
+            result.mMax = std::max(fAudioVoiceDefaultMaxDistance * fAudioMaxDistanceMult, result.mMin);
+            return result;
         }
     }
 
@@ -67,6 +129,7 @@ namespace MWSound
         , mTimePassed(0.f)
         , mLastCell(nullptr)
         , mCurrentRegionSound(nullptr)
+        , mWorkQueue(new SceneUtil::WorkQueue(1))
     {
         if(!useSound)
         {
@@ -120,7 +183,7 @@ namespace MWSound
         return std::make_shared<FFmpeg_Decoder>(mVFS);
     }
 
-    DecoderPtr SoundManager::loadVoice(const std::string &voicefile)
+    DecoderPtr SoundManager::loadVoice(const std::string &voicefile) const
     {
         try
         {
@@ -158,45 +221,12 @@ namespace MWSound
         return mStreams.get();
     }
 
-    StreamPtr SoundManager::playVoice(DecoderPtr decoder, const osg::Vec3f &pos, bool playlocal)
+    bool SoundManager::playVoice(DecoderPtr decoder, bool playlocal, Stream* stream)
     {
-        MWBase::World* world = MWBase::Environment::get().getWorld();
-        static const float fAudioMinDistanceMult = world->getStore().get<ESM::GameSetting>().find("fAudioMinDistanceMult")->mValue.getFloat();
-        static const float fAudioMaxDistanceMult = world->getStore().get<ESM::GameSetting>().find("fAudioMaxDistanceMult")->mValue.getFloat();
-        static const float fAudioVoiceDefaultMinDistance = world->getStore().get<ESM::GameSetting>().find("fAudioVoiceDefaultMinDistance")->mValue.getFloat();
-        static const float fAudioVoiceDefaultMaxDistance = world->getStore().get<ESM::GameSetting>().find("fAudioVoiceDefaultMaxDistance")->mValue.getFloat();
-        static float minDistance = std::max(fAudioVoiceDefaultMinDistance * fAudioMinDistanceMult, 1.0f);
-        static float maxDistance = std::max(fAudioVoiceDefaultMaxDistance * fAudioMaxDistanceMult, minDistance);
-
-        bool played;
-        float basevol = volumeFromType(Type::Voice);
-        StreamPtr sound = getStreamRef();
-        if(playlocal)
-        {
-            sound->init([&] {
-                SoundParams params;
-                params.mBaseVolume = basevol;
-                params.mFlags = PlayMode::NoEnv | Type::Voice | Play_2D;
-                return params;
-            } ());
-            played = mOutput->streamSound(decoder, sound.get(), true);
-        }
+        if (playlocal)
+            return mOutput->streamSound(decoder, stream, true);
         else
-        {
-            sound->init([&] {
-                SoundParams params;
-                params.mPos = pos;
-                params.mBaseVolume = basevol;
-                params.mMinDistance = minDistance;
-                params.mMaxDistance = maxDistance;
-                params.mFlags = PlayMode::Normal | Type::Voice | Play_3D;
-                return params;
-            } ());
-            played = mOutput->streamSound3D(decoder, sound.get(), true);
-        }
-        if(!played)
-            return nullptr;
-        return sound;
+            return mOutput->streamSound3D(decoder, stream, true);
     }
 
     // Gets the combined volume settings for the given sound type
@@ -221,19 +251,10 @@ namespace MWSound
         Log(Debug::Info) << "Playing " << filename;
         mLastPlayedMusic = filename;
 
-        stopMusic();
-
-        DecoderPtr decoder = getDecoder();
-        decoder->open(filename);
-
-        mMusic = getStreamRef();
-        mMusic->init([&] {
-            SoundParams params;
-            params.mBaseVolume = volumeFromType(Type::Music);
-            params.mFlags = PlayMode::NoEnv | Type::Music | Play_2D;
-            return params;
-        } ());
-        mOutput->streamSound(decoder, mMusic.get());
+        const auto createDecoder = makeWorkItem([this, fileName = filename] { createMusicDecoder(fileName); });
+        const auto deadline = std::chrono::steady_clock::now() + sAsyncOperationTimeout;
+        mWaitingMusic.emplace_back(Music {filename, createDecoder, deadline});
+        mWorkQueue->addWorkItem(createDecoder);
     }
 
     void SoundManager::advanceMusic(const std::string& filename)
@@ -283,7 +304,7 @@ namespace MWSound
 
     bool SoundManager::isMusicPlaying()
     {
-        return mMusic && mOutput->isStreamPlaying(mMusic.get());
+        return !mWaitingMusic.empty() || (mMusic && mOutput->isStreamPlaying(mMusic.get()));
     }
 
     void SoundManager::playPlaylist(const std::string &playlist)
@@ -355,21 +376,7 @@ namespace MWSound
         if(!mOutput->isInitialized())
             return;
 
-        std::string voicefile = "Sound/"+filename;
-
-        mVFS->normalizeFilename(voicefile);
-        DecoderPtr decoder = loadVoice(voicefile);
-        if (!decoder)
-            return;
-
-        MWBase::World *world = MWBase::Environment::get().getWorld();
-        const osg::Vec3f pos = world->getActorHeadTransform(ptr).getTrans();
-
-        stopSay(ptr);
-        StreamPtr sound = playVoice(decoder, pos, (ptr == MWMechanics::getPlayer()));
-        if(!sound) return;
-
-        mSaySoundsQueue.emplace(ptr, std::move(sound));
+        sayAsync(ptr, filename, mWaitingVoice);
     }
 
     float SoundManager::getSaySoundLoudness(const MWWorld::ConstPtr &ptr) const
@@ -389,18 +396,7 @@ namespace MWSound
         if(!mOutput->isInitialized())
             return;
 
-        std::string voicefile = "Sound/"+filename;
-
-        mVFS->normalizeFilename(voicefile);
-        DecoderPtr decoder = loadVoice(voicefile);
-        if (!decoder)
-            return;
-
-        stopSay(MWWorld::ConstPtr());
-        StreamPtr sound = playVoice(decoder, osg::Vec3f(), true);
-        if(!sound) return;
-
-        mActiveSaySounds.emplace(MWWorld::ConstPtr(), std::move(sound));
+        sayAsync(MWWorld::ConstPtr(), filename, mActiveWaitingVoice);
     }
 
     bool SoundManager::sayDone(const MWWorld::ConstPtr &ptr) const
@@ -412,6 +408,13 @@ namespace MWSound
                 return false;
             return true;
         }
+
+        const auto isPtr = [&] (const Voice& v) { return v.mPtr == ptr; };
+
+        const auto activeWaiting = std::find_if(mActiveWaitingVoice.begin(), mActiveWaitingVoice.end(), isPtr);
+        if (activeWaiting != mActiveWaitingVoice.end())
+            return false;
+
         return true;
     }
 
@@ -433,11 +436,26 @@ namespace MWSound
             return false;
         }
 
+        const auto isPtr = [&] (const Voice& v) { return v.mPtr == ptr; };
+
+        const auto waiting = std::find_if(mWaitingVoice.begin(), mWaitingVoice.end(), isPtr);
+        if (waiting != mWaitingVoice.end())
+            return true;
+
+        const auto activeWaiting = std::find_if(mActiveWaitingVoice.begin(), mActiveWaitingVoice.end(), isPtr);
+        if (activeWaiting != mActiveWaitingVoice.end())
+            return true;
+
         return false;
     }
 
     void SoundManager::stopSay(const MWWorld::ConstPtr &ptr)
     {
+        const auto isPtr = [&] (const Voice& v) { return v.mPtr == ptr; };
+
+        mWaitingVoice.erase(std::remove_if(mWaitingVoice.begin(), mWaitingVoice.end(), isPtr), mWaitingVoice.end());
+        mActiveWaitingVoice.erase(std::remove_if(mActiveWaitingVoice.begin(), mActiveWaitingVoice.end(), isPtr), mActiveWaitingVoice.end());
+
         SaySoundMap::iterator snditer = mSaySoundsQueue.find(ptr);
         if(snditer != mSaySoundsQueue.end())
         {
@@ -1018,6 +1036,9 @@ namespace MWSound
         if(!mOutput->isInitialized() || mPlaybackPaused)
             return;
 
+        playAllVoicesFromCreatedDecoders();
+        playMusicFromCreatedDecoder();
+
         updateSounds(duration);
         if (MWBase::Environment::get().getStateManager()->getState()!=
             MWBase::StateManager::State_NoGame)
@@ -1105,6 +1126,10 @@ namespace MWSound
             mActiveSaySounds.erase(sayiter);
             mActiveSaySounds.emplace(updated, std::move(stream));
         }
+
+        for (auto& v : mWaitingVoice)
+            if (v.mPtr == old)
+                v.mPtr = updated;
     }
 
     // Default readAll implementation, for decoders that can't do anything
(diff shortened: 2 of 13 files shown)


kcat:
I don't know about this. This makes it so the game will keep playing while the voice file is being opened, which means in degenerate cases the character might not speak for a while despite needing to. Skyrim had an annoying bug just like this; a character's supposed to say something, and sometimes they'd just stare blankly for an extra long moment before giving their line, as if an asynchronous load was being stalled.

Additionally, this seems like it could cause problems with scripts, where they can call `say` on a character and then check `saydone` before the voice clip gets a chance to play.

FWIW, the voice is already being streamed, which means it's not being decoded all at once (in fact, a fair bit of work went into being able to stream voices and build up the 'lip' data in small bits with each update). If the issue is specifically with the open call, it may be worth looking into what exactly is making it slow and see if that can be fixed or improved. FFmpeg is a pretty large library, so it's possible there's more efficient ways to prepare sound files for our audio decoding needs.

wareya:
Regardless of the fact that this doesn't seem to be complete, it's a very
strict taboo to access audio assets from disk on a game logic thread, so
moving it onto another thread is completely appropriate. You're going to
get random multi-second stalls on file/filesystem operations sometimes no
matter what you're doing, even if you're just setting up buffers for
streaming. There's no problem with, for example, having the game go on for
a second while an audio asset loads into cache the first time it's used, as
long as it's somehow made invisible to game logic. Or do you want to
preload every voice asset in the game?

elsid:
> This makes it so the game will keep playing while the voice file is being opened, which means in degenerate cases the character might not speak for a while despite needing

I can add a deadline in frame numbers like if file isn't opened in 1 to n frames, wait until it is.

> Additionally, this seems like it could cause problems with scripts, where they can call `say` on a character and then check `saydone` before the voice clip gets a chance to play.

I missed this but for this current logic is not clear. It was introduced by https://github.com/OpenMW/openmw/commit/4c2b694b298af2c1a0decb4c304f0d316a3efe95. [`SoundManager::sayDone`](https://github.com/OpenMW/openmw/blob/da65a3b8d475e6546d1ecb7335c472bd071e57a7/apps/openmw/mwsound/soundmanagerimp.cpp#L558-L568) checks only `mActiveSaySounds`. But [here](https://github.com/OpenMW/openmw/blob/da65a3b8d475e6546d1ecb7335c472bd071e57a7/apps/openmw/mwsound/soundmanagerimp.cpp#L524) ptr is added to `mSaySoundsQueue`. Is it intended? The only case ptr is added to `mActiveSaySounds` [here](https://github.com/OpenMW/openmw/blob/da65a3b8d475e6546d1ecb7335c472bd071e57a7/apps/openmw/mwsound/soundmanagerimp.cpp#L555) with empty Ptr. So [this](https://github.com/OpenMW/openmw/blob/da65a3b8d475e6546d1ecb7335c472bd071e57a7/apps/openmw/mwscript/soundextensions.cpp#L58) call always returns true. I think I should fix this. `sayDone` should check both queues. Then [`sayActive`](https://github.com/OpenMW/openmw/blob/da65a3b8d475e6546d1ecb7335c472bd071e57a7/apps/openmw/mwsound/soundmanagerimp.cpp#L570-L589) will become a negation of `sayDone`.

> If the issue is specifically with the open call, it may be worth looking into what exactly is making it slow and see if that can be fixed or improved

If it's a disk, then there is nothing to improve.

Capostrophic:
sayActive is not supposed to be a negation of sayDone. sayDone is supposed to be a vanilla-compatible implementation of SayDone, which returns 1 during the frame a speech sound is started (i.e. no current say sound) and it's only used for that scripting function. sayActive returns the actual status of a speech sound (1 if there's a sound) used for our game mechanics.

elsid:
> So sayDone shouldn't care if there's something in the queue. Don't "fix" it, it'll break mods.

Checked the script in the game and found that it actually works, so I will not change it. I need to make appropriate changes to this pr now.

kcat:
> You're going to get random multi-second stalls on file/filesystem operations sometimes no matter what you're doing, even if you're just setting up buffers for streaming.

Only if disk access is temporarily blocked due to other processes hammering the disk. The kernel should be expected to balance the needs of various apps to avoid resource starvation, and disk access is no exception. However, with this it's not just dependent on disk access, but also on thread scheduling and the work queue to not cause stalls.

> There's no problem with, for example, having the game go on for a second while an audio asset loads into cache the first time it's used, as long as it's somehow made invisible to game logic.

I disagree. Having an audio cue delayed from the corresponding visual cue is quite jarring. If you move the mouse, you wouldn't want to see the game continue rendering new frames but not actually turn the camera for an extra 30+ frames afterward when a needed model gets loaded. Similarly, if you hit someone with a sword, you wouldn't want the game to continue on only to hear the hit/grunt sound a half second later after the particulars and stagger animation are finished. Sound feedback is important, and you don't want to let it lag behind the game simulation too far or go too out of sync with the visuals.

> Or do you want to preload every voice asset in the game?

Loading sounds as needed shouldn't be a significant stall. Games were loading sounds synchronously on mechanical drives decades ago and had no problems holding 60fps. Longer sounds, like voice lines, would be streamed to avoid loading too much at once. That some people are experiencing stalls on modern SSDs from just opening a file (not even fully decoding/loading it) points to an issue being elsewhere. Throwing it onto another thread is just hiding the problem and exchanging it for others.

kcat:
> I can add a deadline in frame numbers like if file isn't opened in 1 to n frames, wait until it is.

A deadline would be preferable to no deadline. However, you're potentially making the problem worse since the overhead of asynchronous workloads is being added.

> If it's a disk, then there is nothing to improve.

It depends on what the problem is. Like I mentioned, FFmpeg is a big library. It's designed to automagically handle tons of formats for many different purposes. However, what we need when opening a regular wav/mp3/ogg/etc audio file to get decoded PCM samples from is different than what we'd need for opening random video file 53 that can have multiple video and audio streams we may want to dynamically jump between during playback. When naively opening a file, ffmpeg has to be prepared for us wanting the latter, even though we really only need the former. As shots in the dark, we might be able to improve it by pre-preparing codecs and coverters that can be shared between instances, without having to search/detect and recreate them for each file.

AnyOldName3:
Is it too much overhead to pre-emptively get a handle to each audio file and maybe read the header so we've not got any of the actual *file* in memory, but don't need to do any setup for streaming either and can start immediately once it's time to actually play something?

wareya:
>Is it too much overhead to pre-emptively get a handle to each audio file and maybe read the header so we've not got any of the actual file in memory, but don't need to do any setup for streaming either and can start immediately once it's time to actually play something?

There only seem to be around 200MB of (non-music) audio assets in Morrowind, so this actually shouldn't be too bad. In fact, if you have a computer with enough RAM to run a modern web browser, you actually *could* just preload all the voice assets in the entire game, at least in their compressed form, despite me suggesting such as a joke earlier.

kcat:
> Is it too much overhead to pre-emptively get a handle to each audio file and maybe read the header so we've not got any of the actual file in memory, but don't need to do any setup for streaming either and can start immediately once it's time to actually play something?

We wouldn't be able to hold open a handle to each audio file, as the OS provides only a limited number of file handles (my system only allows 1024 file descriptors, and there's over 7000 loose wav/mp3 files in a vanilla install). Though we might be able to pre-scan each one to get the codec it needs and precalculate its length, to help reduce the work ffmpeg needs to do when loading the file during play. We might also be able to limit the allowed codecs so it doesn't try to match files to dozens of completely esoteric formats nobody really uses (or at least give higher priority to ones we more expect to be used).

> A thread that's not being waited on getting stuck for a few seconds won't cause stalls, and describing it as such is misleading.

It would stall a sound event since it can't play until the work queue finishes the task (which in turn depends on the worker thread getting scheduled, potentially vying for a timeslice on some core that the main thread or rendering thread also wants, and not having any other tasks already in the queue it needs to wait for). And since voices aren't and can't realistically be cached, this can potentially happen every time someone speaks.

> Yes, actually, I would. If the game locks while I'm turning the camera real fast because it wants to load in a model, I might get *physically sick*.

I guess we have different standards, then. If I see the game continue without doing something it should do (turn the camera, play a sound), that makes me think it's bugging out because of an error it's not reporting. If there's resource contention and the game has to pause for a quick moment to load something, I can tell it's an issue outside of the app's direct control. Neither is good, but silently delaying feedback is worse to me than an occasional very slight hitch from loading/preparing an uncached resource.

wareya:
> And since voices aren't and can't realistically be cached, this can potentially happen every time someone speaks.

This exact same argument can be used for the game hitching, not just voice files being delayed. I don't know about you, but I don't want to walk into a new building and get a series of seconds-long stutters because every NPC decides to greet me at the same time (actual experience, minus the stutters).

kcat:
> I don't know about you, but I don't want to walk into a new building and get a series of seconds-long stutters because every NPC decides to greet me at the same time (actual experience, minus the stutters).

My system is a good 10+ years old; ancient dual-core processor, mechanical HD, low memory, the whole nine yards, and the only times I'd notice a hitch with voices is if I attacked a room with several NPCs causing them all to shout in alarm at the same exact time. During normal play where only a single character would start speaking on a given frame, it was completely seamless. However, I'd not be surprised if this causes problems for my system since there's only so much processor time to go around, meaning the work queue is inherently more flaky when it comes to completing what would be very quick work in a reasonable amount of time; freeing up memory to avoid swap or stopping background processes that want to use the disk won't be able to help since the problem won't be the disk.

elsid:
I gathered some profile for openmw + ffmpeg 4.2.3 using [systemtap](https://sourceware.org/systemtap/) with [script](https://gist.github.com/elsid/39aae01dd2d9dcd408bfece6dd558464). There are 2 kinds of issues:

1. Slow `avio_read`.

Top calls by sum duration:

| function                                                                          |                duration |            calls number |     duration, % (frame) |
| --------------------------------------------------------------------------------- | ----------------------- | ----------------------- | ----------------------- |
| update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp:312 |                   43511 |                       1 |                   100.0 |
| update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/actors.cpp:1662             |                   43201 |                       1 |       99.28753648502678 |
| open@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:178            |                   41236 |                       1 |       94.77143710785778 |
| avformat_open_input@src/libavformat/utils.c:537                                   |                   31367 |                       1 |       72.08981636827468 |
| avio_read@src/libavformat/aviobuf.c:647                                           |                   30946 |                      15 |       71.12224494955298 |
| av_probe_input_buffer2@src/libavformat/format.c:222                               |                   30786 |                       1 |       70.75452184505068 |
| avformat_find_stream_info@src/libavformat/utils.c:3584                            |                    9681 |                       1 |       22.24954609179288 |
| av_probe_input_format3@src/libavformat/format.c:128                               |                    9558 |                       6 |       21.96685895520673 |
| av_reallocp@src/libavutil/mem.c:163                                               |                       8 |                       1 |     0.01838615522511549 |

Stack:

```
1 update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp:312 43511 43201 99.28753648502678% 100.0% 100.0%
  2 update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/actors.cpp:1662 43201 41236 95.4514941783755% 99.28753648502678% 99.28753648502678%
    3 open@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:178 41236 41048 99.54408769036765% 95.4514941783755% 94.77143710785778%
      4 avformat_find_stream_info@src/libavformat/utils.c:3584 9681 9382 96.91147608718109% 23.477058880589777% 22.24954609179288%
        5 av_probe_input_format3@src/libavformat/format.c:128 3863 0 0.0% 39.90290259270736% 8.878214704327641%
        6 avio_read@src/libavformat/aviobuf.c:647 4 0 0.0% 0.041318045656440455% 0.009193077612557745%
        7 avio_read@src/libavformat/aviobuf.c:647 11 0 0.0% 0.11362462555521123% 0.025280963434533798%
        8 avio_read@src/libavformat/aviobuf.c:647 6 0 0.0% 0.061977068484660676% 0.013789616418836617%
        9 avio_read@src/libavformat/aviobuf.c:647 17 0 0.0% 0.17560169403987191% 0.03907057985337042%
        10 av_probe_input_format3@src/libavformat/format.c:128 3031 0 0.0% 31.308749096167755% 6.96605456091563%
        11 avio_read@src/libavformat/aviobuf.c:647 5 0 0.0% 0.05164755707055056% 0.01149134701569718%
        12 avio_read@src/libavformat/aviobuf.c:647 12 0 0.0% 0.12395413696932135% 0.027579232837673235%
        13 avio_read@src/libavformat/aviobuf.c:647 5 0 0.0% 0.05164755707055056% 0.01149134701569718%
        14 avio_read@src/libavformat/aviobuf.c:647 15 0 0.0% 0.15494267121165167% 0.03447404104709154%
        15 av_probe_input_format3@src/libavformat/format.c:128 1352 0 0.0% 13.965499431876871% 3.1072602330445176%
        16 avio_read@src/libavformat/aviobuf.c:647 5 0 0.0% 0.05164755707055056% 0.01149134701569718%
        17 avio_read@src/libavformat/aviobuf.c:647 14 0 0.0% 0.14461315979754158% 0.032175771643952106%
        18 av_probe_input_format3@src/libavformat/format.c:128 642 0 0.0% 6.631546327858692% 1.4754889568155178%
        19 avio_read@src/libavformat/aviobuf.c:647 6 0 0.0% 0.061977068484660676% 0.013789616418836617%
        20 av_probe_input_format3@src/libavformat/format.c:128 377 0 0.0% 3.8942258031195123% 0.8664475649835675%
        21 avio_read@src/libavformat/aviobuf.c:647 17 0 0.0% 0.17560169403987191% 0.03907057985337042%
      22 avformat_open_input@src/libavformat/utils.c:537 31367 31163 99.34963496668473% 76.06702880977785% 72.08981636827468%
        23 avio_read@src/libavformat/aviobuf.c:647 371 0 0.0% 1.182771702744923% 0.8526579485647309%
        24 avio_read@src/libavformat/aviobuf.c:647 6 0 0.0% 0.01912838333280199% 0.013789616418836617%
        25 av_probe_input_buffer2@src/libavformat/format.c:222 30786 30753 99.89280841941141% 98.147734880607% 70.75452184505068%
          26 av_probe_input_format3@src/libavformat/format.c:128 293 0 0.0% 0.9517313064379912% 0.6733929351198548%
          27 avio_read@src/libavformat/aviobuf.c:647 30452 0 0.0% 98.91509127525498% 69.9868998644021%
          28 av_reallocp@src/libavutil/mem.c:163 8 0 0.0% 0.02598583771844345% 0.01838615522511549%
```

2. Less slow `av_probe_input_format3`:

Top calls by sum duration:
| function                                                                          |                duration |            calls number |     duration, % (frame) |
| --------------------------------------------------------------------------------- | ----------------------- | ----------------------- | ----------------------- |
| update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp:312 |                   11327 |                       1 |                   100.0 |
| update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/actors.cpp:1662             |                   11038 |                       1 |       97.44857420323122 |
| open@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:178            |                    8704 |                       1 |        76.8429416438598 |
| avformat_find_stream_info@src/libavformat/utils.c:3584                            |                    8045 |                       1 |       71.02498455018981 |
| av_probe_input_format3@src/libavformat/format.c:128                               |                    7953 |                       6 |        70.2127659574468 |
| avformat_open_input@src/libavformat/utils.c:537                                   |                     515 |                       1 |       4.546658426767899 |
| av_probe_input_buffer2@src/libavformat/format.c:222                               |                     290 |                       1 |        2.56025425973338 |
| avio_read@src/libavformat/aviobuf.c:647                                           |                     143 |                      12 |      1.2624702039374944 |
| av_reallocp@src/libavutil/mem.c:163                                               |                       6 |                       1 |     0.05297077778758718 |

Stack:

```
1 update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp:312 11327 11038 97.44857420323122% 100.0% 100.0%
  2 update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/actors.cpp:1662 11038 8704 78.85486501177749% 97.44857420323122% 97.44857420323122%
    3 open@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:178 8704 8560 98.34558823529412% 78.85486501177749% 76.8429416438598%
      4 avformat_find_stream_info@src/libavformat/utils.c:3584 8045 7790 96.83032939714109% 92.42876838235294% 71.02498455018981%
        5 av_probe_input_format3@src/libavformat/format.c:128 2935 0 0.0% 36.482287134866375% 25.91153880109473%
        6 avio_read@src/libavformat/aviobuf.c:647 12 0 0.0% 0.14916096954630206% 0.10594155557517436%
        7 av_probe_input_format3@src/libavformat/format.c:128 2554 0 0.0% 31.746426351771284% 22.547894411582945%
        8 avio_read@src/libavformat/aviobuf.c:647 5 0 0.0% 0.06215040397762585% 0.04414231482298932%
        9 avio_read@src/libavformat/aviobuf.c:647 10 0 0.0% 0.1243008079552517% 0.08828462964597864%
        10 avio_read@src/libavformat/aviobuf.c:647 6 0 0.0% 0.07458048477315103% 0.05297077778758718%
        11 avio_read@src/libavformat/aviobuf.c:647 13 0 0.0% 0.16159105034182722% 0.11477001853977223%
        12 av_probe_input_format3@src/libavformat/format.c:128 1267 0 0.0% 15.748912367930393% 11.185662576145493%
        13 avio_read@src/libavformat/aviobuf.c:647 5 0 0.0% 0.06215040397762585% 0.04414231482298932%
        14 avio_read@src/libavformat/aviobuf.c:647 13 0 0.0% 0.16159105034182722% 0.11477001853977223%
        15 av_probe_input_format3@src/libavformat/format.c:128 634 0 0.0% 7.880671224362959% 5.597245519555045%
        16 avio_read@src/libavformat/aviobuf.c:647 6 0 0.0% 0.07458048477315103% 0.05297077778758718%
        17 av_probe_input_format3@src/libavformat/format.c:128 318 0 0.0% 3.9527656929770045% 2.8074512227421207%
        18 avio_read@src/libavformat/aviobuf.c:647 12 0 0.0% 0.14916096954630206% 0.10594155557517436%
      19 avformat_open_input@src/libavformat/utils.c:537 515 337 65.4368932038835% 5.916819852941177% 4.546658426767899%
        20 avio_read@src/libavformat/aviobuf.c:647 42 0 0.0% 8.155339805825243% 0.3707954445131103%
        21 avio_read@src/libavformat/aviobuf.c:647 5 0 0.0% 0.9708737864077669% 0.04414231482298932%
        22 av_probe_input_buffer2@src/libavformat/format.c:222 290 265 91.37931034482759% 56.310679611650485% 2.56025425973338%
          23 av_probe_input_format3@src/libavformat/format.c:128 245 0 0.0% 84.48275862068965% 2.1629734263264764%
          24 avio_read@src/libavformat/aviobuf.c:647 14 0 0.0% 4.827586206896552% 0.12359848150437008%
          25 av_reallocp@src/libavutil/mem.c:163 6 0 0.0% 2.0689655172413794% 0.05297077778758718%
```

Frame time graph for this test:
timeseries

elsid:
1. Fixed sayDone/sayActive.
2. Added 200ms deadline for decoding. If deadline is reached then wait until async task is done.

Want to make music decoding async too.

kcat:
To clarify, this PR doesn't make voices decode asynchronously. They're already streamed with a background thread. All this PR does is open the file asynchronously, then start a streaming sound.

> Want to make music decoding async too.

It already is.

elsid:
> It already is.

Ok, you're right. I mean create a decoder and open file. It's not full decoding but it's a part of a decoding process. Changed all the naming in the pr to reflect this.

Sounds are left now.

kcat:
Honestly, if the performance of opening audio files is a real problem, it would be best to have a generic solution. As it is, sound effects are going to be loaded about as often as voices (any time a sound is first used, or if it's been purged from cache due to lack of use), which are opened and decoded synchronously.

`av_probe_input_buffer2` and `av_probe_input_format3` seem to be the biggest culprits, which would align with what I suggested above that the issue could be with ffmpeg needing to detect the codec out of all possible codecs, then find other info the stream has (that we may not need). If there's a way to avoid the overhead from them, that would likely solve the issue.

It you really need to, though, a more general solution may be to handle audio similar to graphics. Put all the stuff with openal and ffmpeg on a separate thread, separate from the main simulation, and have the main thread just send commands to it for a given frame. A fence of some sort can be used to ensure a frame's commands have been processed before sending more commands, ensuring it doesn't lag too much behind.

psi29a:
Essentially preloading (priming) assets before need them in other words?

kcat:
Sort of. My original thought was to go through the audio files at load time and have ffmpeg detect what codec it would use, and store that information with the sound definition. Then when it needs to be loaded, we could tell ffmpeg to use that codec (or prioritize that codec) so it doesn't have to probe the file during runtime.

We could also look into seeing if calling `avformat_find_stream_info` is strictly needed. The comments in avformat.h actually mention:
```
@todo Let the user decide somehow what information is needed so that we do not waste time getting stuff the user does not need.
```
indicating they know it's a time-consuming call.

elsid:
> Put all the stuff with openal and ffmpeg on a separate thread, separate from the main simulation

Looking at this sample, I agree.

Top calls by duration:
|    id | function                                                                             |                duration |         callee duration |      callee duration, % |    duration, % (caller) |     duration, % (frame) |
| ----- | ------------------------------------------------------------------------------------ | ----------------------- | ----------------------- | ----------------------- | ----------------------- | ----------------------- |
|     1 | frame@/home/elsid/dev/openmw/apps/openmw/engine.cpp:88                               |                   22924 |                   20225 |       88.22631303437446 |                   100.0 |                   100.0 |
|     5 | updateRegionSound@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:931 |                   19274 |                   19251 |       99.88066825775657 |         84.077822369569 |         84.077822369569 |
|     7 | loadSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:945           |                   19237 |                   18939 |       98.45090190778187 |       99.80803154508665 |       83.91641947304136 |
|    10 | open@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:178               |                   18545 |                   18474 |       99.61714747910489 |       96.40276550397671 |       80.89774908392951 |
|    11 | avformat_find_stream_info@src/libavformat/utils.c:3584                               |                   17954 |                   17080 |       95.13200401024842 |       96.81315718522512 |       78.31966497993369 |
|    47 | av_probe_input_format3@src/libavformat/format.c:128                                  |                    8007 |                       0 |                     0.0 |      44.597304221900416 |       34.92845925667423 |
|    80 | av_probe_input_format3@src/libavformat/format.c:128                                  |                    3996 |                       0 |                     0.0 |       22.25687868998552 |      17.431512824986914 |
|    97 | av_probe_input_format3@src/libavformat/format.c:128                                  |                    2088 |                       0 |                     0.0 |       11.62972039656901 |       9.108358052695865 |
|   106 | av_probe_input_format3@src/libavformat/format.c:128                                  |                    1054 |                       0 |                     0.0 |       5.870558092904088 |       4.597801430814866 |
|   111 | av_probe_input_format3@src/libavformat/format.c:128                                  |                     934 |                       0 |                     0.0 |       5.202183357469088 |       4.074332577211656 |
|     2 | update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp:312    |                     771 |                     540 |       70.03891050583657 |      3.3632873844006284 |      3.3632873844006284 |
|     3 | update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/actors.cpp:1662                |                     540 |                       0 |                     0.0 |       70.03891050583657 |      2.3556098412144477 |
|   117 | avformat_open_input@src/libavformat/utils.c:537                                      |                     520 |                     401 |       77.11538461538461 |      2.8039902938797523 |      2.2683650322805793 |
|     8 | readAll@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:407            |                     394 |                       9 |       2.284263959390863 |       2.048136403805167 |      1.7187227359972084 |
|   114 | av_probe_input_format3@src/libavformat/format.c:128                                  |                     308 |                       0 |                     0.0 |       1.715495154283168 |      1.3435700575815739 |

Top calls by sum duration:
| function                                                                                             |                duration |            calls number |     duration, % (frame) |
| ---------------------------------------------------------------------------------------------------- | ----------------------- | ----------------------- | ----------------------- |
| frame@/home/elsid/dev/openmw/apps/openmw/engine.cpp:88                                               |                   22924 |                       1 |                   100.0 |
| updateRegionSound@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:931                 |                   19274 |                       1 |         84.077822369569 |
| loadSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:945                           |                   19237 |                       1 |       83.91641947304136 |
| open@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:178                               |                   18545 |                       1 |       80.89774908392951 |
| avformat_find_stream_info@src/libavformat/utils.c:3584                                               |                   17954 |                       1 |       78.31966497993369 |
| av_probe_input_format3@src/libavformat/format.c:128                                                  |                   16622 |                       7 |       72.50916070493805 |
| update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp:312                    |                     771 |                       1 |      3.3632873844006284 |
| avio_read@src/libavformat/aviobuf.c:647                                                              |                     723 |                      50 |      3.1538998429593437 |
| update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/actors.cpp:1662                                |                     540 |                       1 |      2.3556098412144477 |
| avformat_open_input@src/libavformat/utils.c:537                                                      |                     520 |                       1 |      2.2683650322805793 |
| readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14                          |                     399 |                      52 |       1.740533938230675 |
| readAll@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:407                            |                     394 |                       1 |      1.7187227359972084 |
| av_probe_input_buffer2@src/libavformat/format.c:222                                                  |                     275 |                       1 |       1.199616122840691 |
| updateSounds@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:1080                     |                     149 |                       1 |      0.6499738265573198 |
| seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35                                |                     108 |                       9 |     0.47112196824288954 |

Stack:
```
1 frame@/home/elsid/dev/openmw/apps/openmw/engine.cpp:88 22924 20225 88.22631303437446% 100.0% 100.0%
  2 update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/mechanicsmanagerimp.cpp:312 771 540 70.03891050583657% 3.3632873844006284% 3.3632873844006284%
    3 update@/home/elsid/dev/openmw/apps/openmw/mwmechanics/actors.cpp:1662 540 0 0.0% 70.03891050583657% 2.3556098412144477%
  4 updateWaterSound@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:983 20 0 0.0% 0.08724480893386843% 0.08724480893386843%
  5 updateRegionSound@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:931 19274 19251 99.88066825775657% 84.077822369569% 84.077822369569%
    6 playSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1132 14 0 0.0% 0.07263671266991803% 0.061071366253707904%
    7 loadSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:945 19237 18939 98.45090190778187% 99.80803154508665% 83.91641947304136%
      8 readAll@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:407 394 9 2.284263959390863% 2.048136403805167% 1.7187227359972084%
        9 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 2.284263959390863% 0.039260164020240795%
      10 open@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:178 18545 18474 99.61714747910489% 96.40276550397671% 80.89774908392951%
        11 avformat_find_stream_info@src/libavformat/utils.c:3584 17954 17080 95.13200401024842% 96.81315718522512% 78.31966497993369%
          12 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 12 0 0.0% 0.06683747354350006% 0.05234688536032106%
          13 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 12 0 0.0% 0.06683747354350006% 0.05234688536032106%
          14 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 13 0 0.0% 0.07240726300545838% 0.05670912580701448%
          15 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 6 0 0.0% 0.03341873677175003% 0.02617344268016053%
          16 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 7 0 0.0% 0.03898852623370837% 0.030535683126853952%
          17 avio_read@src/libavformat/aviobuf.c:647 18 10 55.55555555555556% 0.10025621031525009% 0.07852032804048159%
            18 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 10 0 0.0% 55.55555555555556% 0.04362240446693422%
          19 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            20 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          21 avio_read@src/libavformat/aviobuf.c:647 16 9 56.25% 0.08911663139133341% 0.06979584714709475%
            22 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 56.25% 0.039260164020240795%
          23 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            24 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          25 avio_read@src/libavformat/aviobuf.c:647 15 9 60.0% 0.08354684192937507% 0.06543360670040133%
            26 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 60.0% 0.039260164020240795%
          27 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            28 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          29 avio_read@src/libavformat/aviobuf.c:647 16 9 56.25% 0.08911663139133341% 0.06979584714709475%
            30 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 56.25% 0.039260164020240795%
          31 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            32 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          33 avio_read@src/libavformat/aviobuf.c:647 15 9 60.0% 0.08354684192937507% 0.06543360670040133%
            34 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 60.0% 0.039260164020240795%
          35 avio_read@src/libavformat/aviobuf.c:647 11 4 36.36363636363637% 0.06126768408154172% 0.04798464491362764%
            36 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 36.36363636363637% 0.017448961786773687%
          37 avio_read@src/libavformat/aviobuf.c:647 16 9 56.25% 0.08911663139133341% 0.06979584714709475%
            38 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 56.25% 0.039260164020240795%
          39 avio_read@src/libavformat/aviobuf.c:647 11 3 27.27272727272727% 0.06126768408154172% 0.04798464491362764%
            40 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 3 0 0.0% 27.27272727272727% 0.013086721340080265%
          41 avio_read@src/libavformat/aviobuf.c:647 16 9 56.25% 0.08911663139133341% 0.06979584714709475%
            42 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 56.25% 0.039260164020240795%
          43 avio_read@src/libavformat/aviobuf.c:647 11 4 36.36363636363637% 0.06126768408154172% 0.04798464491362764%
            44 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 36.36363636363637% 0.017448961786773687%
          45 avio_read@src/libavformat/aviobuf.c:647 24 14 58.333333333333336% 0.13367494708700012% 0.10469377072064212%
            46 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 14 0 0.0% 58.333333333333336% 0.061071366253707904%
          47 av_probe_input_format3@src/libavformat/format.c:128 8007 0 0.0% 44.597304221900416% 34.92845925667423%
          48 avio_read@src/libavformat/aviobuf.c:647 11 4 36.36363636363637% 0.06126768408154172% 0.04798464491362764%
            49 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 36.36363636363637% 0.017448961786773687%
          50 avio_read@src/libavformat/aviobuf.c:647 17 9 52.94117647058824% 0.09468642085329175% 0.07415808759378817%
            51 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 52.94117647058824% 0.039260164020240795%
          52 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            53 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          54 avio_read@src/libavformat/aviobuf.c:647 15 8 53.333333333333336% 0.08354684192937507% 0.06543360670040133%
            55 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 8 0 0.0% 53.333333333333336% 0.034897923573547374%
          56 avio_read@src/libavformat/aviobuf.c:647 11 4 36.36363636363637% 0.06126768408154172% 0.04798464491362764%
            57 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 36.36363636363637% 0.017448961786773687%
          58 avio_read@src/libavformat/aviobuf.c:647 15 9 60.0% 0.08354684192937507% 0.06543360670040133%
            59 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 60.0% 0.039260164020240795%
          60 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            61 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          62 avio_read@src/libavformat/aviobuf.c:647 17 9 52.94117647058824% 0.09468642085329175% 0.07415808759378817%
            63 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 52.94117647058824% 0.039260164020240795%
          64 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            65 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          66 avio_read@src/libavformat/aviobuf.c:647 16 9 56.25% 0.08911663139133341% 0.06979584714709475%
            67 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 56.25% 0.039260164020240795%
          68 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            69 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          70 avio_read@src/libavformat/aviobuf.c:647 15 9 60.0% 0.08354684192937507% 0.06543360670040133%
            71 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 60.0% 0.039260164020240795%
          72 avio_read@src/libavformat/aviobuf.c:647 11 4 36.36363636363637% 0.06126768408154172% 0.04798464491362764%
            73 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 36.36363636363637% 0.017448961786773687%
          74 avio_read@src/libavformat/aviobuf.c:647 16 9 56.25% 0.08911663139133341% 0.06979584714709475%
            75 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 56.25% 0.039260164020240795%
          76 avio_read@src/libavformat/aviobuf.c:647 11 3 27.27272727272727% 0.06126768408154172% 0.04798464491362764%
            77 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 3 0 0.0% 27.27272727272727% 0.013086721340080265%
          78 avio_read@src/libavformat/aviobuf.c:647 19 11 57.89473684210527% 0.10582599977720841% 0.08288256848717501%
            79 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 11 0 0.0% 57.89473684210527% 0.04798464491362764%
          80 av_probe_input_format3@src/libavformat/format.c:128 3996 0 0.0% 22.25687868998552% 17.431512824986914%
          81 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            82 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          83 avio_read@src/libavformat/aviobuf.c:647 15 9 60.0% 0.08354684192937507% 0.06543360670040133%
            84 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 60.0% 0.039260164020240795%
          85 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            86 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          87 avio_read@src/libavformat/aviobuf.c:647 15 8 53.333333333333336% 0.08354684192937507% 0.06543360670040133%
            88 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 8 0 0.0% 53.333333333333336% 0.034897923573547374%
          89 avio_read@src/libavformat/aviobuf.c:647 10 4 40.0% 0.055697894619583375% 0.04362240446693422%
            90 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 40.0% 0.017448961786773687%
          91 avio_read@src/libavformat/aviobuf.c:647 15 9 60.0% 0.08354684192937507% 0.06543360670040133%
            92 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 60.0% 0.039260164020240795%
          93 avio_read@src/libavformat/aviobuf.c:647 11 4 36.36363636363637% 0.06126768408154172% 0.04798464491362764%
            94 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 36.36363636363637% 0.017448961786773687%
          95 avio_read@src/libavformat/aviobuf.c:647 20 11 55.00000000000001% 0.11139578923916675% 0.08724480893386843%
            96 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 11 0 0.0% 55.00000000000001% 0.04798464491362764%
          97 av_probe_input_format3@src/libavformat/format.c:128 2088 0 0.0% 11.62972039656901% 9.108358052695865%
          98 avio_read@src/libavformat/aviobuf.c:647 11 3 27.27272727272727% 0.06126768408154172% 0.04798464491362764%
            99 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 3 0 0.0% 27.27272727272727% 0.013086721340080265%
          100 avio_read@src/libavformat/aviobuf.c:647 15 9 60.0% 0.08354684192937507% 0.06543360670040133%
            101 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 60.0% 0.039260164020240795%
          102 avio_read@src/libavformat/aviobuf.c:647 11 4 36.36363636363637% 0.06126768408154172% 0.04798464491362764%
            103 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 36.36363636363637% 0.017448961786773687%
          104 avio_read@src/libavformat/aviobuf.c:647 17 10 58.82352941176471% 0.09468642085329175% 0.07415808759378817%
            105 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 10 0 0.0% 58.82352941176471% 0.04362240446693422%
          106 av_probe_input_format3@src/libavformat/format.c:128 1054 0 0.0% 5.870558092904088% 4.597801430814866%
          107 avio_read@src/libavformat/aviobuf.c:647 11 4 36.36363636363637% 0.06126768408154172% 0.04798464491362764%
            108 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 36.36363636363637% 0.017448961786773687%
          109 avio_read@src/libavformat/aviobuf.c:647 20 12 60.0% 0.11139578923916675% 0.08724480893386843%
            110 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 12 0 0.0% 60.0% 0.05234688536032106%
          111 av_probe_input_format3@src/libavformat/format.c:128 934 0 0.0% 5.202183357469088% 4.074332577211656%
          112 avio_read@src/libavformat/aviobuf.c:647 12 4 33.33333333333333% 0.06683747354350006% 0.05234688536032106%
            113 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 4 0 0.0% 33.33333333333333% 0.017448961786773687%
          114 av_probe_input_format3@src/libavformat/format.c:128 308 0 0.0% 1.715495154283168% 1.3435700575815739%
          115 avio_read@src/libavformat/aviobuf.c:647 17 9 52.94117647058824% 0.09468642085329175% 0.07415808759378817%
            116 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 9 0 0.0% 52.94117647058824% 0.039260164020240795%
        117 avformat_open_input@src/libavformat/utils.c:537 520 401 77.11538461538461% 2.8039902938797523% 2.2683650322805793%
          118 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 8 0 0.0% 1.5384615384615385% 0.034897923573547374%
          119 avio_read@src/libavformat/aviobuf.c:647 51 44 86.27450980392157% 9.807692307692308% 0.22247426278136453%
            120 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 44 0 0.0% 86.27450980392157% 0.19193857965451055%
          121 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 12 0 0.0% 2.307692307692308% 0.05234688536032106%
          122 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 12 0 0.0% 2.307692307692308% 0.05234688536032106%
          123 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 12 0 0.0% 2.307692307692308% 0.05234688536032106%
          124 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 12 0 0.0% 2.307692307692308% 0.05234688536032106%
          125 seek@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:35 15 0 0.0% 2.8846153846153846% 0.06543360670040133%
          126 avio_read@src/libavformat/aviobuf.c:647 4 0 0.0% 0.7692307692307693% 0.017448961786773687%
          127 av_probe_input_buffer2@src/libavformat/format.c:222 275 260 94.54545454545455% 52.88461538461539% 1.199616122840691%
            128 av_probe_input_format3@src/libavformat/format.c:128 235 0 0.0% 85.45454545454545% 1.025126504972954%
            129 avio_read@src/libavformat/aviobuf.c:647 25 16 64.0% 9.090909090909092% 0.10905601116733554%
              130 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 16 0 0.0% 64.0% 0.06979584714709475%
  131 updateSounds@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:1080 149 85 57.04697986577181% 0.6499738265573198% 0.6499738265573198%
    132 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 4 0 0.0% 2.684563758389262% 0.017448961786773687%
    133 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 4 0 0.0% 2.684563758389262% 0.017448961786773687%
    134 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 4 0 0.0% 2.684563758389262% 0.017448961786773687%
    135 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 4 0 0.0% 2.684563758389262% 0.017448961786773687%
    136 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 4 0 0.0% 2.684563758389262% 0.017448961786773687%
    137 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 5 0 0.0% 3.3557046979865772% 0.02181120223346711%
    138 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 5 0 0.0% 3.3557046979865772% 0.02181120223346711%
    139 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 44 0 0.0% 29.53020134228188% 0.19193857965451055%
    140 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 4 0 0.0% 2.684563758389262% 0.017448961786773687%
    141 updateSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:1239 7 0 0.0% 4.697986577181208% 0.030535683126853952%
  142 playMusicFromCreatedDecoder@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:1540 4 0 0.0% 0.017448961786773687% 0.017448961786773687%
  143 playAllVoicesFromCreatedDecoders@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:1491 7 0 0.0% 0.030535683126853952% 0.030535683126853952%
```

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -53,6 +80,7 @@ namespace MWSound
         , mUnderwaterSound(nullptr)
         , mNearWaterSound(nullptr)
         , mPlaybackPaused(false)
+        , mWorkQueue(new SceneUtil::WorkQueue(1))
```
akortunov, line 83: Not sure if it is OK to use different work queues in different places.

elsid: In general most likely it's not OK, but the problem is when we have different threads working on different stuff OS will schedule them properly. If we start to mix I/O sound related stuff, cell preloading, objects unrefs, navmesh building we would need a much smarter scheduler. Otherwise each type of work might be blocked by another and produce unwanted behaviour like delays. Also mixing CPU bound and I/O bound work in a single thread is not a good idea, because these are different types of resources and it's unwanted to stop navmesh building or unreferencing because the sound file is opened. Yes, a work queue can have multiple threads, but again we'd need much complex logic inside to utilize them properly while we can just use separate objects that are much simpler.

elsid:
Made all sound initial loading asynchronous. Partially this is a refactoring now, because I wanted to reduce the amount of code inside SoundManager before changing the logic. Need some testing to make sure I didn't break anything. And now profile doesn't have big spikes:
async_sound_load_mech_and_sound
async_sound_load

kcat:
If you're going to ultimately put audio processing on a background thread, it doesn't make sense to get asynchronous sound opening/loading working since it'll be removed afterward (no need to asynchronously load a sound when the audio updates are already being handled asynchronously). Especially if it's somewhat flaky by it not being safe to have multiple work queues.

elsid:
Technically speaking audio processing on a background thread is an asynchronous process relatively to a main thread including loading. The only question is how communication is made keeping in mind that it's two way and minimizing contention. Main thread posts requests for audio to be played and reads the state.

I don't consider this change as complete and still working on it. For now I have solved the problem, now it's time to refactor the solution to the proper state.

I understand there are some arguments about this pr, but could you be more specific. Just look at the code. Now there are asynchronous requests to play audio, but they are processed partially on the background thread for I/O, and the rest is done in the main thread. In this way there is no contention on reading the state of SoundManager which I think is a good thing. Later I can put all processing into a separate thread. Don't expect all changes to appear at once.

> Especially if it's somewhat flaky by it not being safe to have multiple work queues.

I don't understand how I should address this. Did you find some flakiness? Could you share more details? How does this statement help the progress of this pr?

akortunov:
> We could also look into seeing if calling avformat_find_stream_info is strictly needed.

It is needed until we fill needed data manually.

akortunov:
Also I have never seen such large delays as in elsid's testing, but I noticed that usually `avformat_find_stream_info` takesmuch more time for WAV's than for MP3's:
```
sound/fx/foot/rhev.wav 0.002851 sec
sound/vo/w/f/hlo_wf137.mp3 0.000266 sec
```

psi29a:
@akortunov isn't that because mp3 has metadata while WAV files do not and there has to be additional testing for to find out the type?

akortunov:
> isn't that because mp3 has metadata while WAV files do not and there has to be additional testing for to find out the type?

More likely. 
In this case maybe we indeed can probe sound files in the background task during game start and cache data.

wareya:
What kind of "type" are we talking about here? RIFF WAVE files always store their sample rate/format and number of channels in their "fmt " field, which almost always comes before the "data" field. Also, skipping RIFF fields is "cheap" (read length and fseek over).

akortunov:
```
sound/fx/tavern02.wav 0.012938 sec
sound/am/am_music5.wav 0.011859 sec
sound/cr/kwamf/moan.wav 0.002873 sec
```

Also I have a suspicion that it takes more time to scan large WAV files than small ones.

kcat:
Not too surprising, RIFF WAVE files can actually get pretty complex. Standard pcm wave files can be read simply, but they're capable of much more, like multiple (non-)interleaved data streams, and there are many different chunk types along with lists with sub-chunks, all with their own parameters that need to be parsed and verified. And while you can skip data chunks by seeking, it isn't free as it causes a break in the input stream (like memory, disk reads are most efficient when you're acting sequentially; a seek breaks the flow and causes various buffer flushes and refills).

wareya:
I strongly doubt there's anything weird going on in the RIFF WAVE files in morrowind. All the ones I've opened up were pretty standard: RIFF header, WAVE magic, fmt chunk defining normal PCM data, data chunk, done, including `sound/cr/kwamf/moan.wav` that was just pointed out.

avformat_find_stream_info seems to be a "guess what filetype this is - is it MPEG? MP3? etc" thing, so my guess is that RIFF WAVE is unusually low on the list of possibilities it's checking, and there's some format before it that requires actually reading the whole file to know if it's valid for that format or not, whereas mp3 is above it so it doesn't have that problem. Again, just a guess.

akortunov:
In theory, file extension itself is a good hint, but I have no idea how to pass it to the ffmpeg.

AnyOldName3:
Could `wav`s be getting all their frames inspected while `mp3`s are only getting their first frame inspected or something dumb like that?

wareya:
Normal PCM `wav`s only have a single giant sample buffer.

kcat:
> Could `wav`s be getting all their frames inspected while `mp3`s are only getting their first frame inspected or something dumb like that?

I would bet that's what's happening. As mentioned, wav files can be pretty complex; that Morrowind's aren't doesn't matter much to ffmpeg, it still needs to inspect all the chunks just in case. mp3 on the other hand is a very dumb format. Technically it's just a single mpeg audio data stream without any container. So it's assumed to be a single stream, and it can just grab the header from the first block, and it got what it needs for the stream. Given too that mp3s are most often CBR, even the length can be calculated simply by reading the bitrate from the header and checking the total data size.

> In theory, file extension itself is a good hint, but I have no idea how to pass it to the ffmpeg.

It already is being passed. `avformat_open_input` is given a pre-allocated format context with custom IO methods, along with the file name. It can use the filename to take a hint about the incoming data even though it doesn't open the named file. Doesn't really help for `wav`s though, since it's a rather flexible format and so can't assume much.

akortunov:
Regarding low priority - there are `format_whitelist` and `codec_whitelist` fields in the AVFormatContext, but I can not find even a single example how to use them.

wareya:
I don't see why it would have to inspect anything in depth if the `fmt ` chunk tells it that it's an ordinary PCM file and there's only a single `data` chunk and nothing else, unless it's doing some kind of waveform analysis, which would be silly.

kcat:
The `fmt ` chunk only says what the format of the subsequent data chunks are. And that chunk itself isn't that straight forward; the fmt chunk is a variable size that contains a tag indicating what it should expect to read and how to interpret it. It gets that, then goes to the next chunk, not knowing what to expect. Finds a `data` chunk, and takes its size along with the previous fmt chunk to calculate the audio length, seeks passed the data, then tries to read another chunk. It fails from EOF, so it can then finalize the data it has and set up the stream info. That's not even getting into the layers of mess that RIFF has with chunk and sub-chunk sizes that need to be checked for consistency.

Compared to mp3 where it reads a uint32, does a bitmask check to verify yeah, it appears to be an mp3 data header. One mp3 stream, take various bits of that uint32 to get its sample rate, channel count, and bitrate. Scale the input data size with the bitrate, and you got a rough stream length. Done.

wareya:
I'm not saying that there isn't more work to do than loading a non-ID3 mp3, I'm saying it doesn't have to do anything that grows linearly with size when it's working with the wav files morrowind has, there isn't a frame stream to loop over or anything like that. I checked Tavern02.wav from the mod it's from and it's just an ordinary PCM file with nothing funny going on, so it should be processing the same as moan.wav. If the filesystem driver is really locking up for 10ms because of a seek operation followed by a failed read then that's a) crazy b) something we can't fix c) all the more reason that it can't happen on the main game logic thread.

kcat:
> I'm not saying that there isn't more work to do than loading a non-ID3 mp3, I'm saying it doesn't have to do anything that grows linearly with size when it's working with the wav files morrowind has

Sure, and I've said I don't see or understand the kind of stutters or hitches when loading one sound at a time that some other people seem to, either. But given an IO-bound problem some people are having, it makes sense that `wav`s would be harder hit than `mp3`.

apps/openmw/mwsound/soundmanagerimp.cpp
akortunov: Also keep in mind that most of users have not too powerful GPU's - about 2c/4t, which can be already utilized by OpenMW (main thread, draw thread, preloading thread, navmesh updates thread). In this case your new task can be executed on the main thread or on the draw one anyway, so the issue is not really fixed.

wareya: You benefit from having many more threads than cores as long as you're not triggering contention issues. If most of them are just blocking on I/O access you're not going to have issues or diminishing returns *just* from having additional threads. Doing CPU and I/O bound work on the same thread is much worse than anything you're going to experience from pushing the I/O bound stuff onto a second thread. If there are going to be problems here they're probably going to of the sort that elsid eluded to.

psi29a: That is pretty much how Python's threading works too or any other 'light weight threading' implementation, defer work that is waiting on I/O out to another thread so that the main loop can continue. There is almost no CPU contention here with I/O bound threads.

wareya: A lot of the time they use state machines or coroutines or green threads instead of OS/hardware threads, but the end result is the same: not blocking on I/O.

akortunov: >  Doing CPU and I/O bound work on the same thread is much worse than anything you're going to experience from pushing the I/O bound stuff onto a second thread.

We do not know how much CPU time ffmpeg itself takes for data processing and how much time it takes just to wait for IO, so it is unclear how it will behave with lack of available CPU cores.

Also keep in mind that WorkQueue with one thread is completely sequential. It means that it can start to execute next work item only after previous one is fully finished. So if there are 5 sounds in queue and it takes 0.1s to handle every file, 5th one can not be handled in next 0.4s, and 0.2s threshold will not actually help.

wareya: I don't think ffmpeg is doing anything particularly expensive. At the very least it shouldn't be. It's not like it has to do pre-baked compression or resampling or anything.

Yes, the point about how the work queue itself works is valid.

AnyOldName3:
Has anyone tried building FFMPEG with the necessary options to make it profilable and then working out what it's *actually* spending time on?

elsid:
> Has anyone tried building FFMPEG with the necessary options to make it profilable and then working out what it's actually spending time on?

I have. [This](https://github.com/OpenMW/openmw/pull/2928#issuecomment-650661330) comment contains some info.

I've tried to profile with read syscall. Here is one slow frame sample: https://gist.github.com/elsid/326cae87b36b59e03098897600ce3b14 . The interesting part is this subtree:
```
1 frame@/home/elsid/dev/openmw/apps/openmw/engine.cpp:222 60612 57154 94.29485910380782% 100.0% 100.0%
  5 updateRegionSound@/home/elsid/dev/openmw/apps/openmw/mwsound/soundmanagerimp.cpp:837 56297 56272 99.95559266035491% 92.88094766712862% 92.88094766712862%
    7 loadSound@/home/elsid/dev/openmw/apps/openmw/mwsound/openal_output.cpp:946 56221 55982 99.57489194429128% 99.86500168747891% 92.7555599551244%
      11 open@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:178 55780 55722 99.89602007888132% 99.21559559595168% 92.02798125783673%
        147 avformat_open_input@src/libavformat/utils.c:537 31298 31133 99.47280976420218% 56.10971674435281% 51.636639609318294%
          165 av_probe_input_buffer2@src/libavformat/format.c:222 30586 30568 99.94114954554372% 97.72509425522398% 50.46195472843661%
            167 avio_read@src/libavformat/aviobuf.c:647 30248 30237 99.96363395927004% 98.89491924409862% 49.90430937768099%
              168 readPacket@/home/elsid/dev/openmw/apps/openmw/mwsound/ffmpeg_decoder.cpp:14 30237 30212 99.91731983993121% 99.96363395927004% 49.88616115620669%
                169 syscall.read 30212 0 0.0% 99.91731983993121% 49.84491519831057%
```

And some samples of slow file reads:
```
readPacket sound/vo/i/m/atk_im011.mp3 size=2048 duration_ns=30329765
readPacket music/battle/mw battle 8.mp3 size=2048 duration_ns=30615280
readPacket sound/vo/i/m/atk_im013.mp3 size=2048 duration_ns=30387108
readPacket sound/vo/w/m/idl_wm007.mp3 size=2048 duration_ns=30477197
readPacket sound/vo/i/m/cratk_im003.mp3 size=2048 duration_ns=30294116
```

akortunov:
Are there plans to continue development, or this PR can be closed?

elsid:
If https://gitlab.com/OpenMW/openmw/-/merge_requests/520 will be merged I can continue to work on this.

akortunov:
Mentioned MR was merged to master.

elsid:
Rebased on master and includes https://gitlab.com/OpenMW/openmw/-/merge_requests/567 . Two functiional changes from the previous state:
* Try lookup sound buffer first from the main thread and if it's not present load it in a separate thread.
* Synchronize operations on SoundBufferPool via mutex because it's accessed from the main and background thread.

I think current state is good enough to be merged. Still I expect review and testing. If you want to get an overview what this changed does please read commit messages. I've tried to explain in some details the context and what change is about.

apps/openmw/mwsound/openal_output.hpp
```diff
@@ -14,12 +14,13 @@
 
 namespace MWSound
 {
-    class SoundManager;
     class Sound;
     class Stream;
 
     class OpenAL_Output : public Sound_Output
     {
+        const DecoderProvider* mDecoderProvider;
```
kcat, line 27: More a stylistic thing that doesn't really matter too much, but this can be a reference instead of a pointer. It would avoid accidental changes, and make it clear that it's set once on initialization and never changed.

apps/openmw/mwsound/openal_output.hpp
```diff
@@ -20,6 +25,7 @@ namespace MWSound
     class OpenAL_Output : public Sound_Output
     {
         const DecoderProvider* mDecoderProvider;
+        const VFS::Manager* mVFS;
```
kcat, line 28: This is redundant since the decoder also has it. It also creates an implicit dependency of sorts between the output and the decoder... the output checks its VFS to see if the file exists, then passes that filename to the decoder expecting it to open with its VFS. Not that they're expected to be different, but it's something that can trip up someone casually looking through the code.

apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -109,7 +109,7 @@ namespace MWSound
 
     SoundManager::~SoundManager()
     {
-        clear();
+        SoundManager::clear();
```
kcat, line 175: Marking the `SoundManager` class as `final` would also avoid a virtual dispatch, both here and elsewhere that it calls its own virtual methods.

kcat (commented):
I still don't like this. It makes the code a lot more complicated and harder to follow, making it do more work overall, and introduces potential undesirable latency issues (which can only be avoided by making the code more complicated, causing potentially worse stalls than it was trying to avoid in the first place). On top of the increased potential for bugs to seep in. Weren't there concerns that it wasn't safe to use the work queue from here like this?

I would much rather explore other options for making opening audio files more efficient, including trying to make ffmpeg not aggressively scan files for stream information, better hint at it the necessary codecs, using a more lightweight library like libsndfile, or even putting all audio work onto its own dedicated worker thread, like OSG does for OpenGL.

elsid:
> I still don't like this.

It's just an opinion.

> It makes the code a lot more complicated and harder to follow, making it do more work overall, and introduces potential undesirable latency issues (which can only be avoided by making the code more complicated, causing potentially worse stalls than it was trying to avoid in the first place).

It does make code more complicated. Like with asynchronous physics in this project or anywhere where you introduce asynchronicity. There are no any additional latency issues to what we have now.

> Weren't there concerns that it wasn't safe to use the work queue from here like this?

There are no concerns. Work queue is just a thread with a queue. Please stop repeating this groundless idea.

> I would much rather explore other options for making opening audio files more efficient, including trying to make ffmpeg not aggressively scan files for stream information, better hint at it the necessary codecs, using a more lightweight library like libsndfile, or even putting all audio work onto its own dedicated worker thread, like OSG does for OpenGL.

Only putting every audio file into a memory on the game start will allow to avoid IO related latency during game session. Who will do it doesn't matter: OS, openmw, or some library. Putting all audio work into it's own thread will require a lot more code and architecture changes. And all operations with sound will become asynchronous because caller will post a request to play sound, to stop sound, etc. This will create much more complicated client code. OSG is a bad example here, because it prepares a frame in a separate thread but engine doesn't start a new one util work with previous is done. With audio you can only expect that IO will finish in some amount of frames. You must not synchronize on it.

kcat:
> There are no any additional latency issues to what we have now.

Yes there are. As it is now, the sound manager calls to the decoder to load a sound. The decoder opens the file and loads it into a buffer, waiting only the amount of time it needs to for those operations to complete, before it can start playing. With it being asynchronous, the sound manager sends off a work queue item to have the decoder open and load the file. That work queue item then has to wait until the work queue wakes up and gets to that task, doing it and then marks it complete. Then the next time the sound manager updates, it checks to see that the load is complete and can play. So rather than it decoding and start playing ASAP, the decoding is going to be delayed by however long it takes the work queue to deal with it, and it will wait to start playing until the next time the sound manager updates after it completed loading.

You'd be better off saying there'd be no appreciable latency issues, since these added delays would ideally be very short. However, I'm not sure that's a safe guarantee. And given a cap on the time the main simulation will wait, any significant delays caused by the work queue would make loading stutters worse.

> Only putting every audio file into a memory on the game start will allow to avoid IO related latency during game session. Who will do it doesn't matter: OS, openmw, or some library.

But as your previous posts show, that's not the actual problem here. The majority of the stall comes from `avformat_find_stream_info` and `av_probe_input_format3`, which are undoubtedly taking so much time because there's a lot of formats ffmpeg has to consider and it doesn't early-out on probing the file once it gets the bare minimum of information necessary to start loading it. The problem isn't IO latency, it's excessive IO in general. This is indicative of ffmpeg not being a great fit for immediate audio loading (something that ffmpeg's send_packet/receive_frame API design also suggests, as the docs mention its done that way to potentially take advantage of asynchronous processing; good for a video player/encoder, not so much for immediate sound loading).

> Putting all audio work into it's own thread will require a lot more code and architecture changes.

Maybe, but it would be more straightforward and logical. There's less back-and-forth between the frontend and backend as the frontend merely tells the backend what to do, like play a sound or move the sound position, and the backend handles it all from there. The frontend doesn't need to hear back that the backend has done anything. At most it will query the backend for the state of active objects, e.g. whether a sound is still playing or the current playback offset of a sound. It puts more work in the backend to be able to stage changes and have a background thread to process them, but to the frontend, the behavior is the same since there's little feedback necessary.

I'm not saying I'm terribly keen on the idea, but the conceptual design is easier to understand and work with, and less invasive. Yes, it would take a fair amount of work, and is part of why I'd check other simpler options first, like trying to find ways to make ffmpeg more efficient with open files (by more strongly indicating the format, or making it not probe the whole file for every bit of information) or using a different more lightweight decoder.

elsid:
> So rather than it decoding and start playing ASAP, the decoding is going to be delayed by however long it takes the work queue to deal with it, and it will wait to start playing until the next time the sound manager updates after it completed loading.

In the best case scenario playback starts on a next frame. That is ok because on that frame picture changes to represent what's happened in the game logic. For example when actor has to speak on the same frame voice is requested to be played and actor animation starts. On the next frame player sees changed image with started animation. Or maybe even one frame after if there is some kind of buffer.

There is a case when it's not a deadline but more than one frame passed since playback requested. Some additional latency may happen but it is less than a frame duration that is not noticable by human at 60 FPS.

The worst case is when deadline happens. There is no additional latency because SoundManager waits until task is complete and right after starts playback.

It's unlikely for this pr to be merged in the current state or in any state that will use the same approach. And another one requires a fresh start so I don't see a point to keep this open.

PR #2929, opened 2020-06-25
Don't offer a minimum price of 1 gold for zero value items/services (#5484)

Capostrophic:
[Tracker ticket](https://gitlab.com/OpenMW/openmw/-/issues/5484)

This should be everything necessary to fix that.


psi29a:
Please update https://wiki.openmw.org/index.php?title=MCP :)

Capostrophic:
Okay.

PR #2930, opened 2020-06-25
Redistribute hits to carried left slot when possible (#5490)

Capostrophic:
[Tracked issue](https://gitlab.com/OpenMW/openmw/-/issues/5490)

Implement as described. I made an assumption that a torch doesn't count as a shield. Hopefully it's everything that should be necessary.


psi29a:
Update https://wiki.openmw.org/index.php?title=MCP please :)

Capostrophic:
Done.

PR #2932, opened 2020-06-25
Replace OpenThreads by std types

elsid:
Everywhere except [components/resource/scenemanager.cpp](https://github.com/OpenMW/openmw/blob/40ad87bc4de2c43be913d822e185c28537ad06e0/components/resource/scenemanager.cpp) where mutex field from osg class is used.

### DIFF
--- a/apps/opencs/CMakeLists.txt
+++ b/apps/opencs/CMakeLists.txt
@@ -216,7 +216,6 @@ endif(APPLE)
 
 target_link_libraries(openmw-cs
     ${OSG_LIBRARIES}
-    ${OPENTHREADS_LIBRARIES}
     ${OSGTEXT_LIBRARIES}
     ${OSGUTIL_LIBRARIES}
     ${OSGVIEWER_LIBRARIES}
--- a/apps/opencs/view/render/scenewidget.cpp
+++ b/apps/opencs/view/render/scenewidget.cpp
@@ -1,5 +1,8 @@
 #include "scenewidget.hpp"
 
+#include <chrono>
+#include <thread>
+
 #include <QEvent>
 #include <QResizeEvent>
 #include <QTimer>
@@ -184,7 +187,7 @@ void CompositeViewer::update()
     double minFrameTime = _runMaxFrameRate > 0.0 ? 1.0 / _runMaxFrameRate : 0.0;
     if (dt < minFrameTime)
     {
-        OpenThreads::Thread::microSleep(1000*1000*(minFrameTime-dt));
+        std::this_thread::sleep_for(std::chrono::duration<double>(minFrameTime - dt));
     }
 }
 
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -123,7 +123,6 @@ include_directories(
 
 target_link_libraries(openmw
     ${OSG_LIBRARIES}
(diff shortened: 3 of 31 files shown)


akortunov:
From what I can tell, we use OpenThreads since we initally did not use C++11/C++14, so we could not use the `std` multithreading. But internally OSG still uses OpenThreads, and I do not know if different threading sync implementations in the one application can conflict.

kcat:
> But internally OSG still uses OpenThreads, and I do not know if different threading sync implementations in the one application can conflict (for example, am not sure that `std::lock_guard<OpenThreads::Mutex>` works properly).

Technically speaking, it doesn't seem as though that specific example will work. `std::lock_guard` expects the lock method to throw if it fails, but `OpenThreads::Mutex::lock` returns non-0 if it fails. Although I see only one place in the patch that this is done, which could be changed back to use `OpenThreads::ScopedLock` since it's using an `OpenThreads::Mutex` anyway.

In the general case, though, this all maps down to the same system-level primitives (pthreads on Linux, WinAPI on Windows). If there was a potential conflict, the OSG docs would need to mention that you must use OpenThreads since there were non-OpenThreads options well before C++11 (Boost, pthreads or winapi directly, etc).

psi29a:
We moved away from boost::threads because openthreads was there, thus it was one less boost thing less _and_ that with openthreads we could work around OSG's then hard-coded thread affinity problem. Now that we can turn thread affinity off in OSG itself, there is no reason to stay with openthreads and we can go back to boost or use std::thread.

I see no problem with removing all openthread instances here.

kcat:
> I see no problem with removing all openthread instances here and useing normal mutexes.

OSG itself still uses OpenThreads, which is sometimes expected by the API. For example, `mIncrementalCompileOperation->getToCompiledMutex()` returns a `OpenThreads::Mutex*`, which can't safely be used with `std::lock_guard`. While we can (and should) replace our use of OpenThreads mutexes and threads, we still need to be compatible where OSG uses or expects them.

psi29a:
Alright, then we start there and replace ours and do more testing in another MR/PR for OSG facing code.

AnyOldName3:
> If there was a potential conflict, the OSG docs would need to mention that you must use OpenThreads since there were non-OpenThreads options well before C++11 (Boost, pthreads or winapi directly, etc).

There have been other footguns in OSG that I've reported and said they should fix or document. Robert's response was that the code is the documentation and you shouldn't call any OSG function without reading it to see what it does. That doesn't look to me like the attitude of a project that's going to document whether their threading conflicts with other threading.

elsid:
> std::lock_guard expects the lock method to throw if it fails, but OpenThreads::Mutex::lock returns non-0 if it fails

And `OpenThreads::ScopedLock` anyway [ignores](https://github.com/openscenegraph/OpenSceneGraph/blob/63bb537132bab1f8b077838f7550e26405e5fa35/include/OpenThreads/ScopedLock#L31) lock result so either we need to define our own type or fix `OpenThreads::ScopedLock`. For now it doesn't seem like I broke anything.

> Alright, then we start there and replace ours and do more testing in another MR/PR for OSG facing code.

I assume you mean the code that is executed through OSG. For example engine.cpp, mwsound, cs is not the case. But  components/sceneutil is the case.

psi29a:
Going to take this as well to get more testing. We can easily revert if things get hairy.

PR #2935, opened 2020-06-26
Fix faction requirement calculations

akortunov:
Fixes [bug #5499](https://gitlab.com/OpenMW/openmw/-/issues/5499).
Actually we should check `mSkill2` twice instead of once.
Also I added meaningful names, as in TES CS, instead of `mSkill1` and `mSkill2`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
     Bug #5479: NPCs who should be walking around town are standing around without walking
     Bug #5484: Zero value items shouldn't be able to be bought or sold for 1 gold
     Bug #5490: Hits to carried left slot aren't redistributed if there's no shield equipped
+    Bug #5499: Faction advance is available when requirements not met
     Feature #390: 3rd person look "over the shoulder"
     Feature #2386: Distant Statics in the form of Object Paging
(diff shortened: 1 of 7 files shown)


MAtahualpa:
I think you misunderstood my post: The int vector of skill values is sorted. That means we can test the first element for the primary skill check (which we already do) and then need to test the second one (if there are only two skills in the faction's required-skill list) or the third one (if there are more than two skills and that list). In any case, we only need one check for the primary skill and one check for the favoured ones.

akortunov:
> In any case, we only need one check for the primary skill and one check for the favoured ones.

There is a possibility that faction has only two favoured skills. In this case access to the 3d element can lead to crash.

MAtahualpa:
(Sorry for moving my original comment to the issue itself on GitLab.)

We essentially mean the same. I emphasised the fact that we only need to check the second element if there are two skills in the required-skills list and the third one if there are more than two in that list. How you implement these checks is up to you.

MAtahualpa:
I also guessed from the existing code that the author made the mistake of taking all skill values into consideration for the favoured-skill checks, but forgot that the primary skill (the first element of the vector) is already "used" and needs to be ignored.

psi29a:
Thanks both of you. :)

PR #2937, opened 2020-06-27
WIP: New Movement Solver Fixes (Do not merge yet)

wareya:
This is a manual rewrite of #1794 based on a much more recent version of master. See #1794 for a description of what changes this PR does.

This is the PR you want to use if you want to test how the movement solver fixes interact with double-precision bullet or threaded/async movement solving.

I intend to rebase this branch to logically split everything up into single stable commits or squash everything into a single commit so that it doesn't interfere with bisecting, so do not merge it yet.

These changes require double-precision Bullet. Otherwise interacting with certain objects becomes glitchy/unstable because single precision Bullet sucks.


wareya:
As this branch switches to AABBs, it exhibits problems with the cantons in vivec and other incredibly large geometry when using single-precision bullet, but these problems essentially disappear when using double-precision bullet (I checked).

psi29a:
Double precision bullet is default for 64-bit on Debian and Ubuntu. I believe that MacOS already has that turned on in build deps. @nikolaykasyanov Can you double check?
@ananace Do we have DP enabled deps on windows for msvc 2015 and 2019?

psi29a:
@wareya are we tracking this on the issue tracker? Such a massive undertaking deserves a ticket. :)

ananace:
Double-precision builds for Bullet are available for MSVC2017 and 2019, and can be grabbed with a simple addition of a `-D` to the build-env setup script.

Capostrophic:
>  Such a massive undertaking deserves a ticket. :)
there are already like 5 tickets for the issues this should fix

wareya:
Solved by handling actor-actor collisions in a special way: https://gitlab.com/OpenMW/openmw/-/issues/4568

Same as above: https://gitlab.com/OpenMW/openmw/-/issues/1901

Solved by switching to AABBs: https://gitlab.com/OpenMW/openmw/-/issues/4447

Solved by switching to AABBs: https://gitlab.com/OpenMW/openmw/-/issues/4247

Solved by fixing a bug in part of the movement solver this modifies: https://gitlab.com/OpenMW/openmw/-/issues/3137

But there's no issue for "the movement solver sucks in general" if that's what you're thinking about.

wareya:
In other news, I got double precision bullet working on my local build a few days ago and it indeed solves the issues with extremely large collision geometry giving messed up collision position/normal information.

I could try to make the CI systems build this PR with double-precision bullet if it would aid in testing.

AnyOldName3:
For Windows, just add a `-D` or whatever it is here: https://github.com/OpenMW/openmw/blob/master/appveyor.yml#L53

wareya:
Fingers crossed.

wareya:
2019 failed to find bullet, 2017 worked. I had to change the dependency script to use MSVC_REAL_YEAR instead of MSVC_YEAR for bullet because the 2015 archive for double-precision bullet only contains debug symbols and I work on 2017, but it seems that the 2019 archive for double precision bullet is broken (debug only) as well, not just the 2015 archive. Should I hardcode 2017 into the URL it's downloading from?

psi29a:
Ah... a 5 for one deal just by making DP-ing OpenMW :rofl:

ananace:
Ok, how is 2019 broken? That's literally the version I've been testing the most, as it's the native version of the VS tools that I've been using when packaging all the deps.

wareya:
2019:
image
2017:
image
2015 looks like 2019.

ananace:
I'm rebuilding all the packages at the moment, seems like MSBuild has failed to correctly package everything because - and I quote; "Failed to create directory, already exists."

wareya:
Amazing.

ananace:
Pushed the new packages, only one was an outlier this time - and I don't see anything missing in there so I'm guessing the compiler did something different on that version.

nikolaykasyanov:
@psi29a not really, the [branch](https://github.com/OpenMW/openmw-deps-mac/tree/bullet-douple-precision) hasn't been merged. I'll look into getting this into upstream..

psi29a:
@nikolaykasyanov I think at this point we should go double since MacOS since 10.8(?) dropped 32-bit support anyway. Debian/Ubuntu 64-bit builds are default double anyway too.

AbduSharif:
Corpses slide down almost any kind of slope, but save and load the save and they won't slide down.

wareya:
I can't reproduce the above bug on default settings on desktop. Apparently it starts to occur when you run physics at 15fps.

wareya:
Remaining issues before I rebase this branch and ask for a full code review:

- Find out whether casting away const in discrete collision checks is correct or if it'll invoke UB down the line or something
- See if Abdu's bug ever happens at 60fps physics and, if it does, fix it, otherwise just document that it's caused by the pop-outside-of-collision code at low physics framerates
- I want performance testing from someone with a weak *desktop*, at default physics framerate

psi29a:
Would a RPi 2 or 3 count as a weak desktop?

wareya:
No lol

psi29a:
It's CPU limited, 2~4 GiB of RAM and OpenGL 2.1 compliant... and only costs 25 euros. :)

wareya:
Maybe something as strong (weak) as a PC with an old AMD Phenom 2 x4 (give or take a generation). I have a feeling that a pi wouldn't be representative of that performance bracket.

AbduSharif:
Please test it with Double Precision, if anyone is willing to test it

AbduSharif:
Testing the corpse sliding on a slope issue again came up with pretty interesting results:
The humanoid corpses slide on any kind of slope at physics framerate:
- 15: the corpse sildes a bit fast.
- 20: the corpse slides at a slower rate than 15.
The issue doesn't happen with 30 or 60 physics fps.

Discovered another issue while testing:
Guard and Mudcrabs whether alive or dead (a corpse) slide on steep slopes, and the higher the physics framerate the faster they slide down the slope.

psi29a:
> Maybe something as strong (weak) as a PC with an old AMD Phenom 2 x4 (give or take a generation). I have a feeling that a pi wouldn't be representative of that performance bracket.

I have a MIL-SPEC Thinkpad T430: https://www.lenovo.com/us/en/laptops/thinkpad/t-series/t430/
It's a dual-core Intel® Core™ i7-3520M
Hybrid graphics: intel / nvidia (NVS 5400M) (Fermi) which came out in 2012
Roughly the equivalent of a GeForce GT 630M

wareya:
Yeah, that seems to be around the performance level I'm thinking of.

psi29a:
By:
`See if Abdu's bug ever happens at 60fps physics and, if it does, fix it, otherwise just document that it's caused by the pop-outside-of-collision code at low physics framerates`
You mean his corpses sliding down hills thing? He seems to have tested that himself at 15 and 20 fps, at 30 and above it doesn't seem to be a problem.

psi29a:
And of course, I see this PR along with this MR https://gitlab.com/OpenMW/openmw/-/merge_requests/248

apps/openmw/mwphysics/movementsolver.cpp
```diff
@@ -167,6 +193,45 @@ namespace MWPhysics
...
+                    // give up
+                    if(contactCallback4.mDistance < contactCallback.mDistance)
+                        giveup = true;
+                }
+            }
+            if(!giveup)
+                position = physicActor->getPosition() + osg::Vec3f(0.0, 0.0, halfExtents.z());
+            
+        }
+        physicActor->setPosition(tempPosition);
```
fr3dz10, line 234: This whole chunk is problematic to run in background thread. One can't call `setWorldTransform` (called by `setPosition`) from different threads.

I think the same can be achieved with a serie of `convexSweepTest`, ie instead of 

```
contactTest()
if (bad)
{
    setPosition(orig + delta)
    contactTest()
    if (bad again)
    {
        setPosition(orig + delta2)
        contactTest()
    }
}
```
 something like
```
contactTest()
if (bad)
{
    convexSweepTest(actorShape, orig, orig + delta)
    if (bad again)
    {
        convexSweepTest(actorShape, orig, orig + delta2)
    }
}
```

I don't know how practical or if it creates other issues, but that is safe to use at least.

wareya: Is it just the setWorldTransform that is the issue?

wareya: I can't use convexSweepTest for this because it misbehaves when you're already inside of geometry, which is what this code is concerned with.

fr3dz10: > Is it just the setWorldTransform that is the issue?

Here yes. 

 
> I can't use convexSweepTest for this because it misbehaves when you're already inside of geometry, which is what this code is concerned with.

What about putting this in a separate `popOutActor` function and change the algorithm from:

```
doSomePreparation()
if (isStuckInsideSomething)
{
    popOutActor()
}
doTheActualMove()
```
to
```
doSomePreparation()
if (isStuckInsideSomething)
{
    flagActorAsStuck()
    return
}
doTheActualMove()
```

Then I could check the `isStuckInsideSomething` flag after each call to `move()` and call `popOutActor` in a single thread context. This is how I call `setPosition` now, so it could be done in the same place.
The only downside I can think of, is that one simulation step will be skipped for a stuck actor, but given that is an exceptional / abnormal state it may not be that bad. What do you think?

fr3dz10:
I did some perf test with this PR and the async physics on a sandybridge laptop, this is impressive. I thought a more precise solver would decrease the performance, it is actually the opposite happening (by a **large** margin).

All these results are with the "unstuck actor logic" commented out (memory corruption made it too unstable to test anything).
This is the mean of the sum of all the *_taken stats in the main thread, on a scene with 180 actors during 5 minutes idling.

synchronous physics (for reference)
old solver: 409ms
new solver: 47ms

async with single thread bullet
old solver: 249ms
new solver: 12ms

async with multi thread bullet, 1 background thread
old solver: 47ms
new solver: 7ms

async with multi thread bullet, 2 background thread
old solver: 6ms
new solver: 6ms

No fps increase though, but with a mobile integrate GPU one can't expect miracle on this front.

psi29a:
> I did some perf test with this PR and the async physics on a sandybridge laptop, this is impressive. I thought a more precise solver would decrease the performance, it is actually the opposite happening (by a **large** margin).

Cool!

> All these results are with the "unstuck actor logic" commented out (memory corruption made it too unstable to test anything).
> This is the mean of the sum of all the *_taken stats in the main thread, on a scene with 180 actors during 5 minutes idling.

This has me worried, where is the memory corruption happening?

fr3dz10:
> This has me worried, where is the memory corruption happening?

See discussion [here](https://github.com/OpenMW/openmw/pull/2937#discussion_r454640471).

apps/openmw/mwphysics/movementsolver.cpp
wareya: That still won't work. The actor would have to be able to recover its exact original position if any of the tests don't work as desired. It probably makes the most sense to do all of this as a separate pass over all actors outside of the movement solver.

wareya:
Did you use a mod or something for that? I'm having a hard time thinking of any places in vanilla MW with 180 actors laying around to point at like "this is the physics benchmaking place".

fr3dz10:
`placeatpc` console function with a few dozen hostile melee, hostile range and guards.

apps/openmw/mwphysics/movementsolver.cpp
fr3dz10: > That still won't work. The actor would have to be able to recover its exact original position if any of the tests don't work as desired. 

I don't see why it is an issue, it is enough to save and restore the position in case of failure, just as you do now. That being said:

> It probably makes the most sense to do all of this as a separate pass over all actors outside of the movement solver.

Yep, that's a better idea. It could be simply moved to the beginning of `applyQueuedMovement` (or its equivalent). And if it happens to be too expensive (which I doubt) and needs to be moved from the main thread to the background thread it can be done later.

AbduSharif:
```
By:
See if Abdu's bug ever happens at 60fps physics and, if it does, fix it, otherwise just document that it's caused by the pop-outside-of-collision code at low physics framerates
You mean his corpses sliding down hills thing? He seems to have tested that himself at 15 and 20 fps, at 30 and above it doesn't seem to be a problem.
```
The problem is 15 is the default physics value on the android port, so this issue will be present to the users without knowing it, unless it's either fixed or we ask xyzz to change the default value to 30 fps even when updating the app, and remove 15.
Also about the sliding guar and mudcrab on a slope, Vanilla has it too but they don't slide as fast as they do in OpenMW with this movement solver when physics fps is 60, and they hold themselves from sliding down the slope almost immediately.
I'll try to make a video of how fast it is in OpenMW.

AbduSharif:
Here's the video of the guar and mudcrab sliding on a slope issue:
https://youtu.be/StYVCSbHoXc

wareya:
I tracked down the underlying reason for the sliding bug. I also realized that I needed to re-enable collision rotation for strangely-shaped actors (e.g. guars), which no longer causes problems as bad as it used to because of the "pop outside of the world" code and fixed actor-actor collisions.

I want to figure out if we should keep boxes for stragely-shaped creatures; given that we're not going to accurately replicate vanilla collision for them (see streamable: https://streamable.com/5droah), it might make sense to give them rotated stretched spheres for shapes or something instead of rotated boxes.

wareya:
osx build is magically timing out

psi29a:
> osx build is magically timing out

Not so magical, we just got too big! We'll need to split the builds out a bit. Can't wait for macos sharred runners on gitlab.

AbduSharif:
>I want to figure out if we should keep boxes for stragely-shaped creatures; given that we're not going to accurately replicate vanilla collision for them (see streamable: https://streamable.com/5droah), it might make sense to give them rotated stretched spheres for shapes or something instead of rotated boxes.

I think what you suggested if it fixes the behavior in the video would be a good idea, as it's also problematic during combat and sooner or later someone will complain about it.

psi29a:
@wareya how are things going here? Any blockers and/or things that you need?

wareya:
Abdu says that there's still a small amount of corpse sliding on slopes at low physics framerates (e.g. 15fps) and the async branch still objects to the way that I go about popping NPCs out of world geometry because it uses non-threadsafe bullet functions.

psi29a:
The low physics framerate is a tweak for better performance, it isn't the recommended framerate. Though ideally 'corpse sliding' shouldn't be an issue regardless though. I'm not worried about this.

The non-threadsafe bullet functions are more concerning though. Is there anyway to work around that, std::mutex?

wareya:
The proper workaround is moving it outside of the movement solver and into movement event dispatch, but it currently cares about the actor's movement vector, and the direction the actor is moving is determined in the movement solver. I took a shot at refactoring it when I got my PC set back up a couple weeks ago but gave up on that attempt when I remembered that it cared about the actor's movement vector.

fr3dz10:
> The proper workaround is moving it outside of the movement solver and into movement event dispatch, but it currently cares about the actor's movement vector, and the direction the actor is moving is determined in the movement solver. I took a shot at refactoring it when I got my PC set back up a couple weeks ago but gave up on that attempt when I remembered that it cared about the actor's movement vector.

It cares about the movement vector, but it can surely be run as part of a separate pass in a critical section before the movement solver? Or I am missing something?

wareya:
> It cares about the movement vector, but it can surely be run as part of a separate pass in a critical section before the movement solver? Or I am missing something?

The true movement vector is calculated inside of the movement solver. This involves not just wind drag but also things like redirecting underwater actors. Replicating this outside of the movement solver would involve code duplication or splitting things up in weird ways. It's not impossible for me to get working by any means but it wasn't literally a matter of cut-and-paste.

fr3dz10:
I uploaded a branch with both async and this PR [here](https://gitlab.com/fr3dz10/openmw/-/commits/async-physics+new-solver) which include a tentative of moving out the "unstuck" chunk out of the movement solver. This is meant as an aid in discussion, as I find easier to talk with a sample implementation.
I discovered that there can't be overlapping calls to `contactTest` as it contains reference to a global structure (in Bullet), so this logic _must_ be in the main thread.

Difference with this PR wrt. pop-out of the world:
- velocity is an approximate (ie takes into account only the actor orientation and movement vector, but not flying/swimming and wind). 
- it is run numstep times before the movement solver. This is a big difference (which might explain the behavior I describe below), as in this PR after each pop-out attempt the solver is run, whereas in my branch the solver is run only once pop-out has been attempted numstep times (at most)

With it I can get out of "coc balmora" and "coc enbonheart" most of the time. Interestingly, being in first or third person gives different results ("coc ebonheart" works if I switch between 1st and 3rd person, whereas "coc balmora" works only if I'm in 1st person until out of the rock). I don't know if this is a result of a mismerge or un unfortunate effect of the async branch or the recent changes in 3rd person view.

wareya:
There technically shouldn't be much benefit to running it multiple times at once; it just needs to get *closer* to being outside of the world, if possible, and if there's a path to open space the actor will find their way into it over time. It would make sense to do it between each physics frame for each actor, but I assume that's not easy with async physics.

I can't tell at a glance whether the unstuck code might end up running for an actor *during* its own movement solver, which would be very bad. If you can confirm that the locks being used prevent this that would be great.

>Interestingly, being in first or third person gives different results 

IIRC the third person idle animations have small amounts of swaying motion. That's the only reason I can think of.

fr3dz10:
> It would make sense to do it between each physics frame for each actor, but I assume that's not easy with async physics.
> 
> I can't tell at a glance whether the unstuck code might end up running for an actor _during_ its own movement solver, which would be very bad. If you can confirm that the locks being used prevent this that would be great.

Anything that is called by `moveActors` is completely synchronous wrt. physics, the background threads are all blocked while inside. I updated a little bit the branch to move the call exactly between frames in both cases of sync and async. 

Basically, anything you need to do with exclusive access to the collision world and /or outside of the solver can be called from there.

wareya:
Okay, that sounds good. Can I kick the can up the road in moving that code out of the movement solver? If so I can squash this branch and ask for review.

wareya:
I have updated this PR to the latest current master. It should now be safe with async physics, even with the actor unstucking logic. I have not found any other problems or made any other changes; if the above chain of discussion is satisfactory, I'm ready to bundle this PR up into a single commit and ask for code review.

wareya:
CI gave a dep error, not sure what happened here:
```
ERROR: The system cannot find the file specified.
C:\projects\openmw\deps\Bullet-2.89-msvc2017-win64-double.7z
```

psi29a:
@wareya would you be willing to migrate this PR over to gitlab and make a MR that way we can have some coverage across 4 platforms?

wareya:
I can do that when I shrink the PR into a single commit if you want. Is the current state of this PR's conversation acceptable?

CI/before_script.msvc.sh
```diff
@@ -677,8 +677,8 @@ printf "Bullet 2.89 (${BULLET_DBL_DISPLAY})... "
 		printf -- "Exists. (No version checking) "
 	elif [ -z $SKIP_EXTRACT ]; then
 		rm -rf Bullet
-		eval 7z x -y "${DEPS}/Bullet-2.89-msvc${MSVC_YEAR}-win${BITS}${BULLET_DBL}.7z" $STRIP
-		mv "Bullet-2.89-msvc${MSVC_YEAR}-win${BITS}${BULLET_DBL}" Bullet
+		eval 7z x -y "${DEPS}/Bullet-2.89-msvc${MSVC_REAL_YEAR}-win${BITS}${BULLET_DBL}.7z" $STRIP
+		mv "Bullet-2.89-msvc${MSVC_REAL_YEAR}-win${BITS}${BULLET_DBL}" Bullet
```
psi29a, line 681: I believe this is the cause of the appveyor problem, revert this and try again?

wareya: This harks back to this issue: https://github.com/OpenMW/openmw/pull/2937#issuecomment-651921114

It looks like the dependencies are getting vendored from gitlab now though? And the gitlab repo only has it with 2015 in the filename. So I guess it needs to use the "wrong" year now (i.e. 2015). I'll change it and cross my fingers.

psi29a:
I managed to find both:  
Bullet-2.89-msvc2015-win64-double.7z  
Bullet-2.89-msvc2015-win64.7z  

Hold on, if you if you need it... it should be available soon.

CI/before_script.msvc.sh
```diff
@@ -677,8 +677,8 @@ printf "Bullet 2.89 (${BULLET_DBL_DISPLAY})... "
 		printf -- "Exists. (No version checking) "
 	elif [ -z $SKIP_EXTRACT ]; then
 		rm -rf Bullet
-		eval 7z x -y "${DEPS}/Bullet-2.89-msvc${MSVC_REAL_YEAR}-win${BITS}${BULLET_DBL}.7z" $STRIP
-		mv "Bullet-2.89-msvc${MSVC_REAL_YEAR}-win${BITS}${BULLET_DBL}" Bullet
+		eval 7z x -y "${DEPS}/Bullet-2.89-msvc${MSVC_YEAR}-win${BITS}${BULLET_DBL}.7z" $STRIP
+		mv "Bullet-2.89-msvc${MSVC_YEAR}-win${BITS}${BULLET_DBL}" Bullet
```
psi29a: Looks like this is working. Was there a reason why this was changed?

wareya: Yes: https://github.com/OpenMW/openmw/pull/2937#issuecomment-651921114
tl;dr: With the old dep vendoring MSVC_YEAR didn't work.

AbduSharif:
Ugh, scamp on a slope bug is back, this time they slide to the left not downwards.

AbduSharif:
Another "bug", which could be related to https://gitlab.com/OpenMW/openmw/-/issues/5671#note_439612592, since it only happens when async num threads = 1 or more.

AbduSharif:
> Another "bug", which could be related to https://gitlab.com/OpenMW/openmw/-/issues/5671, since it only happens when async num threads = 1 or more.

AbduSharif:
> https://youtu.be/s4v2LGIKyz0

With the master branch the creature disappear btw.

wareya:
I forgot to disable the unstucking logic for noclipping/tcl'd actors. This has been fixed.

AbduSharif:
Cliff racers fall under the ground/height map...etc whatever it's called.

AbduSharif:
Running while jumping on slopes is now impossible without pressing jump multiple times:
https://youtu.be/I88lUmlHE5Q

This is not the case in the current master, however if you stop running and just press jump and move forward it's possible to jump up the slopes.
Is this behavior intentional and should be replicated for use in some weird cases or just a bug? If it has no use in some weird cases, can the openmw behavior be recovered?

AbduSharif:
coc "ebonheart", sometimes it works, sometimes it doesn't and sometimes it just freaks (see the last seconds):
https://youtu.be/MwbgD8kbKh4

Good news is if you don't want to type tcl and get out of the ground you can just type the up arrow and enter and it might just work, still better than nothing and getting out the boring "tcl" way.

wareya:
> Running while jumping on slopes is now impossible without pressing jump multiple times:

The main reason this doesn't work on master is because of a very, very annoying deviation from vanilla behavior that prevents you from jumping if you're touching a "wall". It's going to be easier to jump up mountains with a more accurate movement solver, period.

> Dead cliff racers fall under the ground/height map...etc whatever it's called.

When they die?

AbduSharif:
> The main reason this doesn't work on master
What I mean is the behavior in the video only happens in your branch, it's not an issue with OpenMW master, in OpenMW when you're on a steep slope you can run and jump in a flexible way, in your branch you'll jump lhow it is in the video and keep pressing jump.

> When they die?
Yup, when they're a corpse.

AbduSharif:
> a very, very annoying deviation from vanilla behavior that prevents you from jumping if you're touching a "wall". It's going to be easier to jump up mountains with a more accurate movement solver, period.

And this is why we have the behavior in the video? I see.

AbduSharif:
> seems like it is, climbing mountains is somewhat easier in your branch when compared to master.

Sometimes though, you're still going to have troubles with how you slide down fast on steep mountains.

wareya:
re: cliff racers: this is apparently, uh, just how their collision boxes are offset

openmw_2020-11-07_12-38-53

the movement solver in master handles offset collision boxes in a buggy way, so they're not directly comparable

wareya:
I have confirmed that vanilla works the same way as my movement solver branch with cliff racers. It is no longer an object of suspicion.

mpv-shot0626

AbduSharif:
> I have confirmed that vanilla works the same way as my movement solver branch with cliff racers. It is no longer an object of suspicion.

https://youtu.be/S4rXsrZRchQ

This video shows a behavior different than your branch, in your branch they always are underground, unlootable unless you tcl to below the ground, however if they're killed above a steep slope sometimes some parts of their body appear so you can loot but almost always you can't unless you tcl to get to them underground.
With the current master wether buggy or not they're still above the ground and lootable.

psi29a:
@AbduSharif is this android? Is there something else going on like type of shape being used and/or resolution of physics (15fps)?

It good to know context and what has changed as Android (OpenMicroWave) is not the same as OpenMW from master+this branch.

AbduSharif:
> is this android? Is there something else going on like type of shape being used and/or resolution of physics (15fps)?

Yeah I'm playing on android, I'm using the same physics settings (60 FPS) as PC, the simple collision boxes for actors patch is removed from my build because it's not needed, so nothing different, my build has double precision and multithreaded bullet options enabled.

As for the bug it happens wether async = 0 or more.

wareya:
> This video (this video is from vanilla) shows a behavior different than your branch

If you don't know what mods the video was using then it's useless. I personally opened up the vanilla game with no mods and observed vanilla cliffracer corpse behavior and recorded them falling half through the terrain and verified that the collision box is offset from the model. That's what confirmed means.

AbduSharif:
I'll boot up vanilla and see if the same will happen to me, I remember seeing the same behavior like in your screenshot but only once or twice, I can always loot them still, and every video on the internet shows they're on the ground.
Meanwhile though, Are we going to ignore the fact that there's an unplayable and undesirable behavior here?

> In your branch they always are underground, unlootable unless you tcl to below the ground, however if they're killed above a steep slope sometimes some parts of their body appear so you can loot but almost always you can't unless you tcl to get to them underground.

wareya:
If dead actors by something other than the hitbox shown by TCB then that's some kind of logic bug and not a problem with the movement solver itself. But I have dead actors just fall through the world completely a lot in vanilla too, hitbox and all. It's kind of a shitshow.

Again, OpenMW's current master branch **handles offset hitboxes incorrectly**, in a way that breaks the movement solver, so you can't compare to it directly.

wareya:
As a test, I reverted the change that fixes how offset collision boxes are handled, and it results in cliff racers not ending up going into the ground:

image

```
$ git diff
diff --git a/apps/openmw/mwphysics/movementsolver.cpp b/apps/openmw/mwphysics/movementsolver.cpp
index c560a09911..d1c082f4a1 100644
--- a/apps/openmw/mwphysics/movementsolver.cpp
+++ b/apps/openmw/mwphysics/movementsolver.cpp
@@ -130,7 +130,8 @@ namespace MWPhysics

         // Adjust for collision mesh offset relative to actor's "location"
         // (doTrace doesn't take local/interior collision shape translation into account)
-        actor.mPosition += meshTranslation;
+        //actor.mPosition += meshTranslation;
+        actor.mPosition.z() += halfExtents.z();

         static const float fSwimHeightScale = MWBase::Environment::get().getWorld()->getStore().get<ESM::GameSetting>().find("fSwimHeightScale")->mValue.getFloat();
         float swimlevel = actor.mWaterlevel + halfExtents.z() - (physicActor->getRenderingHalfExtents().z() * 2 * fSwimHeightScale);
@@ -395,7 +396,8 @@ namespace MWPhysics
         physicActor->setOnSlope(isOnSlope);

         actor.mPosition = newPosition;
-        actor.mPosition -= meshTranslation; // remove what was added earlier in compensating for doTrace not taking interior transformation intoaccount
+        //actor.mPosition -= meshTranslation; // remove what was added earlier in compensating for doTrace not taking interior transformation into account
+        actor.mPosition.z() -= halfExtents.z();
     }

     void MovementSolver::unstuck(ActorFrameData& actor, const btCollisionWorld* collisionWorld)
```

which means that, yes, the difference is the fact that my branch fixes how offset collision boxes are handled during movement solving. doTrace does not take the actor's local hitbox transform into account; this will cause movement solver bugs and weird stuff like actor hitboxes shifting around wildly because they're off-center.

I'm not responsible for deciding whether the movement solver is more important or the vanilla cliffracer death behavior is more important, or whether some kind of hack should be applied or something. Maybe the right thing is to ignore only the *vertical* part of the actor's mesh translation, using the half extents for it instead. Who knows, because it's not clear what's going on under the hood, not even whether vanilla morrowind is actually using vertically off-centered collision boxes or just special-cases corpse model origins as also being colliders or something.

AbduSharif:
// NOTE: here we don't account for the collision box translation (i.e. physicActor->getPosition() - refpos.pos).
        // That means the collision shape used for moving this actor is in a different spot than the collision shape
        // other actors are using to collide against this actor.
        // While this is strictly speaking wrong, it's needed for MW compatibility.
        actor.mPosition.z() += halfExtents.z();

There's this note here, is it related?

wareya:
Yes. That behavior resulted in buggy collision, because the collision box's translation is being *totally* ignored, and the actor ended up inside of other geometry far more easily. Revising it to use the collision box's translation made a bunch of subtle bugs go away, including the slope sliding that the unstucking code was causing. This only affects actors with "weird" collision box offsets, so basically just certain random creatures (including cliff racers).

AbduSharif:
What would be the solution you advice, but want the OpenMW's team input on?

wareya:
I don't know? Right now I'm content with leaving this PR as-is because it's not a movement solver bug, but would still cause issues for the movement solver, and fixing it in a way that satisfies everyone is non-obvious?

AbduSharif:
Here's something to keep in mind, apparently cliff rscers' corpse falls under most terrain (except for some that makes some of the corpse parts appear letting you loot them), but they don't for rocks:

wareya:
The edge of the collision shape is probably colliding with the other rock next to it.

AbduSharif:
Btw through out testing this, the game crashed for me more than 10 times already, the crashes are random and seem to happen when the player character collides with cliff racers or when a bunch of cliff racers are inside each other, the game first freezes then crashes, here I managed to record one:
https://youtu.be/0wLXOg6wMXY

wareya:
I am not going to try to debug crashes that almost certainly have nothing to do with this PR, it doesn't change anything that should cause crashes.

AbduSharif:
> I am not going to try to debug crashes that almost certainly have nothing to do with this PR, it doesn't change anything that should cause crashes.

What? how are you so sure? I haven't managed to reporduce this crash with the current master and if it really was an issue with it someone would have reported it already.

If software crashes were predictable, no software would would crash for the users ever.

wareya:
The code changed by this PR does not play with references or pointers or memory allocation or exceptions at all, and does not create or destroy any dynamic resources handled by other systems.

There is only one possible culprit I can think of, which is this:

```
+            // FIXME: This is absolutely terrible and bullet should feel terrible for not making contactPairTest const-correct.
+            const_cast<btCollisionWorld*>(mWorld)->contactPairTest(const_cast<btCollisionObject*>(mMe), const_cast<btCollisionObject*>(convexResult.m_hitCollisionObject), isOverlapping);
```

which is solely used to tell whether or not two actors are overlapping. But if this function can't be used with async physics (threads >= 2) then we're basically shit out of luck for having vanilla-like actor-to-actor collisions with async physics (threads >= 2). And this function exactly the kind of thing that we would debate over code review and decide what to do with later (should we surround it with locks? would that be too expensive? is there a point, is it even broken? why would discrete collision checks be unsafe if tracing isn't? do we have a way to test this for broken behavior? etc.), especially considering that there's a big fat warning on it.

Just getting dropped a video of a port to another platform crashing and getting told "you have a crash lol" is supremely unhelpful.

AbduSharif:
> Just getting dropped a video of a port to another platform crashing and getting told "you have a crash lol" is supremely unhelpful.

What do you expect, I'm here trying to help testing but I'm starting I'm wasting my time.

Crash also happens, when async = 0

wareya:
> Crash also happens, when async = 0.

Then it's almost certainly not caused by this PR and I don't care about it unless I have a crash dump pointing to code that this PR touches.

AbduSharif:
> unless I have a crash dump pointing to code that this PR touches.

A heads up, do not expect that from me, there's no openmw-crash file on neither windows nor android.

psi29a:
If you can reliably reproduce the issue, then it would be easier to figure out. @AbduSharif if possible, have someone who knows more about android builds to enable crash dumps. Maybe throw it in Android Studio with a debugger and run it on the android emulator?

wareya:
Updated to master, which involved tweaking the unstuck code again (I made sure it works properly with async physics (threads = 1) this time)

wareya:
Updated summary of changes:

- Several problems with the movement solver itself are addressed.
  - The movement solver now actually "hugs" objects it interacts with. The previous movement solver did not.
  - It has a proper safety margin, which is necessary because of the "hugging".
  - Multiple vanilla conformance issues related to stairs, walls, and steep slopes interacting with jumping were fixed.
  - The movement solver attempts to detect accute/narrow crevices and slide along them instead of getting stuck.
  - The stairstepping logic was altered slightly to make AABBs play nicer with vanilla map assets.
  - The dysfunctional "move away from anything being touched" code in the movement solver was removed.
- All actors now use rectangular collision boxes instead of some actors (like humanoids) using capsules.
  - Creatures with strangely-shaped collision boxes (e.g. rats) still have them rotate when they turn, for gameplay coherency reasons.
  - Humanoids do not have their collision boxes turn. This is basically vanilla-accurate, despite what vanilla collision display implies.
  - Accurately replicating all the different edge cases for vanilla collision box rotation behavior is basically impossible.
- Actors "unstuck" themselves if they find themselves inside of the world.
  - This makes it possible to rotate creature collision boxes without it causing serious issues.
  - It also helps the player not get stuck in certain entrances and 'coc' destinations, but it is not a universal solution.
  - It might change (probably improve?) how actors interact with objects being gradually moved by scripts.
- Double-precision bullet is forcibly enabled for windows CI builds.
  - Several of the above changes (which are necessary) make single-precision bullet freak out, because single-precision bullet sucks.

Capostrophic:
GitLab counterpart of this PR has been merged.

PR #2939, opened 2020-06-28
Add stats for all updates

elsid:
Allows to discover issues like this:
sound

### DIFF
--- a/apps/openmw/engine.cpp
+++ b/apps/openmw/engine.cpp
@@ -67,6 +67,140 @@ namespace
         if (ret != 0)
             Log(Debug::Error) << "SDL error: " << SDL_GetError();
     }
+
+    struct UserStats
+    {
+        const std::string mLabel;
+        const std::string mBegin;
+        const std::string mEnd;
+        const std::string mTaken;
+
+        UserStats(const std::string& label, const std::string& prefix)
+            : mLabel(label),
+              mBegin(prefix + "_time_begin"),
+              mEnd(prefix + "_time_end"),
+              mTaken(prefix + "_time_taken")
+        {}
+    };
+
+    enum class UserStatsType : std::size_t
+    {
+        Input,
+        Sound,
+        State,
+        Script,
+        Mechanics,
+        Physics,
+        World,
+        Gui,
+
+        Number,
+    };
+
+    template <UserStatsType type>
+    struct UserStatsValue
+    {
+        static const UserStats sValue;
+    };
+
+    template <>
+    const UserStats UserStatsValue<UserStatsType::Input>::sValue {"Input", "input"};
+
+    template <>
+    const UserStats UserStatsValue<UserStatsType::Sound>::sValue {"Sound", "sound"};
+
+    template <>
+    const UserStats UserStatsValue<UserStatsType::State>::sValue {"State", "state"};
+
+    template <>
+    const UserStats UserStatsValue<UserStatsType::Script>::sValue {"Script", "script"};
+
+    template <>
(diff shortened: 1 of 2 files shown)


zinnschlag:
Good idea. The only thing I have to criticise is that this change makes two already overlong functions even longer. Looks like a prime candidate for refactoring. Note that I do not consider this a blocker. Merging this change and then doing the refactoring in a later PR would be perfectly adequate.

As I see it adding a stat consists of three parts:

1. measure the time (OMW::Engine::frame)
2. write it into the stats variable (OMW::Engine::frame)
3. setup the statshandler (OMW::Engine::Engine)

I suggest two improvements:

First, write a small RAII style helper class that handles the first two points by implementing these steps in the constructor/destructor. The code might look something like this:

```C++
{
  TimeStat stat("input", stats, frameNumber, mStartTick);
  mEnvironment.getInputManager()->update(frametime, false);
}
```

For the third point an array with a for loop could replace a lot of lines of code. At a cursory glance it seems that the only changing parameters are the tags (**input**_time_end, **input**_time_taken, **input**_time_begin). If that is correct a simple static string array would do the job.

fr3dz10:
Would you consider a frame_time_taken (ie total) as well? It can be useful to get a bigger picture when comparing changes in particular area I think

akortunov:
It is just a 1000/FPS.

fr3dz10:
Yes, but it is not exported in the stats file.

akortunov:
> Yes, but it is not exported in the stats file.

And it is not sum of updates anyway since OpenMW has another threads (Draw and GPU, for example).

elsid:
1. Made a type to measure duration whithin a scope and write to stats.
2. Setup stats handler initialization to minimize required code addition when new metric is introduced by using enum and some metaprogramming.

psi29a:
Cool, thanks @elsid

akortunov:
There are too many updaters now, and there is no enough place for other widgets, and profiler starts to cover more than half of screen on FullHD:
Screenshot_20200703_161109

I suppose that's why we initially did not implement them.

elsid:
But that's GUI problem. It should automatically adjust for given input to fit the window. I will try to fix it.

AnyOldName3:
We can always add another level of profiler so it's condensed into a few bars until you press F3 again.

Either way, the profiler isn't meant to be unobtrusive. You're not supposed to have it open while actually playing. It's not a problem if it overlays over the whole screen as long as it all *fits* on the screen. I can see the graphics stuff is falling off the bottom in that screenshot, so *something* needs to change, but it doesn't need to be purging of bars to make things concise.

akortunov:
> It should automatically adjust for given input to fit the window

Where to? Both widgets are just too large to place them vertically, and there is a place only on wide screens to place them horizontally.

elsid:
In general scrollbars are designed to fit inside a window more information that it can show. For this specific situation I can't say right now what solution is but definitely removing the information has lowest priority.

akortunov:
> but definitely removing the information has lowest priority.

You already remove information by placing it off-screen, what is not good.

If it is not possible to display a lot of info with our current layout, I'd suggest to do not add new data (that new updater bars, for example) until this issue is addressed.

Eli2:
Just reducing the space between the profiler rows might me enough.
Also the font size in general is too large imo.

elsid:
There is one simple solution - decrease font size, then everything will fit. And the complex one. We can try to change layout, like AnyOldName3 suggested. But all logic is implemented inside osg and not much is available for customization. So to do this we need to make a PR to osg or copy into openmw. I think to make a copy is preferable in the long term, because we can have more control and it's not very clear what customization is needed.

PR #2944, opened 2020-07-01
Regression fixes for input system

akortunov:
1. Do not discard remnants, so it should be possible to move cursor at very high FPS, where amount movement is very small (and become 0 now due to integer cast).
2. Fix stupid copy-paste error which broke mouse relase injection.


psi29a:
Shit happens, glad it's fixed.

PR #2947, opened 2020-07-01
Clamp sound volume on settings update (fix #5507)

elsid:
https://gitlab.com/OpenMW/openmw/-/issues/5507

1. Put logic for volume clamping into a single place.
2. Move all volume settings related stuff into a separate class.

### DIFF
--- a/apps/openmw/mwsound/volumesettings.cpp
+++ b/apps/openmw/mwsound/volumesettings.cpp
@@ -1,0 +1,56 @@
+#include "volumesettings.hpp"
+
+#include <components/settings/settings.hpp>
+
+#include <algorithm>
+
+namespace MWSound
+{
+    namespace
+    {
+        float clamp(float value)
+        {
+            return std::max(0.0f, std::min(1.0f, value));
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


apps/openmw/mwsound/volumesettings.cpp
```diff
@@ -0,0 +1,56 @@
...
+            case Type::Mask:
+                break;
+        }
+
+        return volume;
+    }
+
+    void VolumeSettings::update()
+    {
+        *this = VolumeSettings();
```
akortunov, line 54: Are you sure that such approach is going to work everywhere?

elsid: This is a call of an assignment operator. I can rewrite it as:
```c++
void VolumeSettings::update()
{
    operator=(VolumeSettings());
}
```
I don't see why this may cause problems.

I didn't want to use fill method, because it delays initialization and it's easier to miss a field.

psi29a:
Merge conflict. :(

elsid:
> Merge conflict. :(

Fixed

PR #2948, opened 2020-07-02
Replace SoundManager local static variables by fields

elsid:


apps/openmw/mwsound/soundmanagerimp.cpp
```diff
@@ -32,6 +32,8 @@ namespace MWSound
 {
     namespace
     {
+        const float sMinUpdateInterval = 1.0f / 30.0f;
```
kcat, line 35: `constexpr` would be better here.

elsid: Done

PR #2949, opened 2020-07-02
Choose next action for water sound in a separate function

elsid:
In general this allows to use playSound only once. But this became not very straight and more like experimental because this function has to return pointer to Sound_Buffer in one case. So I used boost::variant+visitor based on custom implementation of [overloaded](https://en.cppreference.com/w/cpp/utility/variant/visit) adjusted for C++14. If this is too much, I can do this without variant.

### DIFF
--- a/apps/openmw/mwsound/soundmanagerimp.cpp
+++ b/apps/openmw/mwsound/soundmanagerimp.cpp
@@ -892,48 +892,67 @@ namespace MWSound
         const ESM::Cell *curcell = player.getCell()->getCell();
         const auto update = mWaterSoundUpdater.update(player, *world);
 
+        WaterSoundAction action;
+        Sound_Buffer* sfx;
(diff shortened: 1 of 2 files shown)


kcat:
I'd strongly prefer not using Boost for this, or being overly complex that makes it difficult for the compiler to optimize. An `enum`+`Sound_Buffer*` pair should work just fine.

psi29a:
Please, no more boost. :)

If you need anything from C++17, feel free to flip that on in cmake and see if MacOS and MSVC2017 are happy. If so, then great.

psi29a:
std::variant is C++17, normally that should be supported by msvc2017 and macos xcode 10.3

akortunov:
As for me, it looks overcomplicated. Overloader, visitor and other stuff is not not really needed here and just make the code more complex without reason.

elsid:
Removed variant and releated stuff.

PR #2950, opened 2020-07-03
Make all stats fit into the window

elsid:
By decreasing font size. Not perfect but good enough until we decide to add more stuff.


AnyOldName3:
It's not a particularly bad approach. The original font size was chosen when we used the OSG default raster font, and that doesn't look so good when small.

PR #2951, opened 2020-07-03
Store SoundManager sounds and streams in object pool

elsid:
Create a new ObjectPool type and move there all related logic.

### DIFF
--- a/components/misc/objectpool.hpp
+++ b/components/misc/objectpool.hpp
@@ -1,0 +1,46 @@
+#ifndef OPENMW_COMPONENTS_MISC_OBJECTPOOL_H
+#define OPENMW_COMPONENTS_MISC_OBJECTPOOL_H
+
+#include <deque>
+#include <memory>
+#include <vector>
+
+namespace Misc
+{
+    template <class T>
+    class ObjectPool
+    {
+        public:
+            ObjectPool()
+                : mObjects(std::make_unique<std::deque<T>>()) {}
+
+            T* get()
+            {
+                T* object;
+
+                if (!mUnused.empty())
+                {
+                    object = mUnused.back();
+                    mUnused.pop_back();
+                }
+                else
+                {
+                    mObjects->emplace_back();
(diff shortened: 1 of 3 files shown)


components/misc/objectpool.hpp
```diff
@@ -0,0 +1,83 @@
...
+            void operator()(T* object) const
+            {
+                mPool->recycle(object);
+            }
+
+        private:
+            ObjectPool<T>* mPool;
+    };
+
+    template <class T>
```
zinnschlag, line 32: Subclassing STL types that are not meant for subclassing is generally not a good idea. Is there a reason why we need a separate class here? As far as I can see a type alias declaration via using should do the job.

elsid: The reason is constructors:
1. To be able to construct by default without defining default constructor for deleter.
2. Construct from nullptr also without defining default constructor for deleter.

Why I don't want to define default constructor for deleter? To protect from this:
```c++
template <class T>
using ObjectPtr = std::unique_ptr<T, ObjectPtrDeleter<T>>;
using StreamPtr = ObjectPtr<Stream>;
StreamPtr obj = pool.get();
Stream* ptr = obj.release();
StreamPtr obj2(ptr); // Will lead to nullptr (mPool) dereference in ObjectPtrDeleter::operator()
````

And if I will not define default constructor for `ObjectPtrDeleter` but will use aliasing:
```c++
StreamPtr ptr; // error: no default ctor for ObjectPtrDeleter
StreamPtr ptr = nullptr; // error: no default ctor for ObjectPtrDeleter
StreamPtr ptr(nullptr, nullptr); // ok
```
So every line `return nullptr;` will have to change to `return {nullptr, nullptr};`.

Probably I'd better to use composition, but then I have to redeclare all used methods and operators.

zinnschlag: Hm, seems a bit over cautious to me. But okay. Kinda makes sense.

PR #2954, opened 2020-07-04
Mutate base records when adding/removing spells

Assumeru:
Split from #2870. This doesn't so much implement vanilla behaviour as vanilla-adjacent behaviour. It should get a good bit of testing.

Instances of the same actor in vanilla share a single in-memory spell list drawn from the base record. GetSpell appears to interact with this list with the interesting side effect that it cannot detect racial abilities. Adding spells to one actor adds them to all instances instantly. (Though visual effects don't get updated until the cell is reloaded.) Basically it's weird all around.

Calling AddSpell on an NPC adds the spell to the base record, provided the NPC doesn't have autocalculated stats. This PR adds it to the base record regardless. (Saving and reloading would also clear any added spells for autocalculated actors.)

The Cure Disease effect can conditionally remove spells (diseases) from the base record. Example:

[instance A]->AddSpell collywobbles
[instance A]->GetCommonDisease == 1
[instance B]->GetCommonDisease == 1
Cast "cure common disease other" on [instance B]
[instance A]->GetCommonDisease == 0
[instance B]->GetCommonDisease == 0
Save and reload
[instance A]->GetCommonDisease == 1
[instance B]->GetCommonDisease == 1
Cast "cure common disease other" on [instance A]
Save and reload
[instance A]->GetCommonDisease == 0
[instance B]->GetCommonDisease == 0

This PR makes both instances affect the base record equally.

### DIFF
--- a/apps/openmw/mwmechanics/spelllist.cpp
+++ b/apps/openmw/mwmechanics/spelllist.cpp
@@ -1,0 +1,175 @@
+#include "spelllist.hpp"
+
+#include <algorithm>
+
+#include <components/esm/loadspel.hpp>
+#include <components/misc/rng.hpp>
+
+#include "spells.hpp"
+
+#include "../mwbase/environment.hpp"
+#include "../mwbase/world.hpp"
+
+#include "../mwworld/esmstore.hpp"
+
+namespace
+{
+    template<class T>
+    const std::vector<std::string> getSpellList(const std::string& id)
+    {
+        return MWBase::Environment::get().getWorld()->getStore().get<T>().find(id)->mSpells.mList;
+    }
+
+    template<class T>
+    bool withBaseRecord(const std::string& id, const std::function<bool(std::vector<std::string>&)>& function)
+    {
+        T copy = *MWBase::Environment::get().getWorld()->getStore().get<T>().find(id);
+        bool changed = function(copy.mSpells.mList);
+        if(changed)
+            MWBase::Environment::get().getWorld()->createOverrideRecord(copy);
+        return changed;
+    }
+}
+
+namespace MWMechanics
+{
+    SpellList::SpellList(const std::string& id, int type) : mId(id), mType(type) {}
+
+    bool SpellList::withBaseRecord(const std::function<bool(std::vector<std::string>&)>& function)
+    {
+        switch(mType)
+        {
+            case ESM::REC_CREA:
+                return ::withBaseRecord<ESM::Creature>(mId, function);
+            case ESM::REC_NPC_:
+                return ::withBaseRecord<ESM::NPC>(mId, function);
+            default:
+                throw std::logic_error("failed to update base record for " + mId);
+        }
+    }
+
+    const std::vector<std::string> SpellList::getSpells() const
+    {
+        switch(mType)
+        {
+            case ESM::REC_CREA:
+                return getSpellList<ESM::Creature>(mId);
+            case ESM::REC_NPC_:
+                return getSpellList<ESM::NPC>(mId);
+            default:
+                throw std::logic_error("failed to get spell list for " + mId);
+        }
+    }
+
+    const ESM::Spell* SpellList::getSpell(const std::string& id)
+    {
+        return MWBase::Environment::get().getWorld()->getStore().get<ESM::Spell>().find(id);
+    }
+
+    void SpellList::add (const ESM::Spell* spell)
+    {
+        auto& id = spell->mId;
+        bool changed = withBaseRecord([&] (auto& spells)
+        {
+            for(auto it : spells)
+            {
+                if(Misc::StringUtils::ciEqual(id, it))
+                    return false;
+            }
+            spells.push_back(id);
+            return true;
+        });
+        if(changed)
+        {
+            for(auto listener : mListeners)
+                listener->addSpell(spell);
+        }
+    }
+
+    void SpellList::remove (const ESM::Spell* spell)
+    {
+        auto& id = spell->mId;
+        bool changed = withBaseRecord([&] (auto& spells)
+        {
+            for(auto it = spells.begin(); it != spells.end(); it++)
+            {
+                if(Misc::StringUtils::ciEqual(id, *it))
+                {
+                    spells.erase(it);
+                    return true;
+                }
+            }
+            return false;
+        });
+        if(changed)
+        {
+            for(auto listener : mListeners)
+                listener->removeSpell(spell);
+        }
+    }
+
+    void SpellList::removeAll (const std::vector<std::string>& ids)
+    {
+        bool changed = withBaseRecord([&] (auto& spells)
+        {
+            const auto it = std::remove_if(spells.begin(), spells.end(), [&] (const auto& spell)
+            {
+                const auto isSpell = [&] (const auto& id) { return Misc::StringUtils::ciEqual(spell, id); };
+                return ids.end() != std::find_if(ids.begin(), ids.end(), isSpell);
+            });
+            if (it == spells.end())
+                return false;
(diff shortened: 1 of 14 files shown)


apps/openmw/mwmechanics/spelllist.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    void SpellList::removeAll (const std::vector<std::string>& ids)
+    {
+        bool changed = withBaseRecord([ids] (auto& spells) {
+            bool changed = false;
+            for(auto it = spells.begin(); it != spells.end(); it++)
+                for(auto& id : ids)
+                    if(Misc::StringUtils::ciEqual(id, *it))
+                    {
+                        spells.erase(it);
```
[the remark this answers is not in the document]
Assumeru: Ah, oops. Thanks.

apps/openmw/mwmechanics/spelllist.cpp
```diff
@@ -0,0 +1,160 @@
...
+namespace
+{
+    template<class T>
+    const std::vector<std::string> getSpellList(const std::string& id)
+    {
+        return MWBase::Environment::get().getWorld()->getStore().get<T>().find(id)->mSpells.mList;
+    }
+
+    template<class T>
+    bool withBaseRecord(const std::string& id, const std::function<bool(std::vector<std::string>&)>& function)
```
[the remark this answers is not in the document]
Assumeru: The template argument would just resolve to the same thing though so all that'd do is save a bit of typing, right?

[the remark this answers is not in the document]
elsid: Not only. `std::function` may not be optimized out https://godbolt.org/z/Qm84mz.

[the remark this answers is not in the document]
Assumeru: Then it won't get optimised out unless I also replace the member function of the same name, since the argument is already an std::function at that point. Unless I'm missing something.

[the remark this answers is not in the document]
elsid: You're right, I've missed that.

apps/openmw/mwmechanics/spells.hpp
```diff
@@ -118,6 +114,10 @@ namespace MWMechanics
 
             void readState (const ESM::SpellState& state, CreatureStats* creatureStats);
             void writeState (ESM::SpellState& state) const;
+
+            bool setSpells(const std::string& id);
+
+            void init(const std::vector<std::string>& spells);
```
zinnschlag, line 120: init is a really odd name for a (public) function in C++, since the actual init phase of construction is supposed to be done in the constructor. To be clear, I do not have a problem with the function per se, just with the name. Maybe you could find something make explicit?

Assumeru:
> However the question is how close should we emulate vanilla behaviour?
Yeah, it honestly just felt like they were trying to optimise something here and that led to these quirks. On the plus side, I now understand why diseased creatures have scripts that reapply the disease. On that note, the ability to add diseases to all instances of a creature seems interesting for modding.

apps/openmw/mwmechanics/spells.hpp
Assumeru: Changed it to `addAllToInstance` to distinguish it from a hypothetical `addAll` that is equivalent to calling `add` in a loop. Doesn't really roll off the tongue, but hey.

zinnschlag:
> On that note, the ability to add diseases to all instances of a creature seems interesting for modding.

As a separate option, yes. But then, we could also make what we have now this option and add another function that only affects individual actors later. That is probably the safer bet.

apps/openmw/mwmechanics/spelllist.cpp
```diff
@@ -0,0 +1,161 @@
...
+        });
+        if(changed)
+            for(auto listener : mListeners)
+                listener->addSpell(spell);
+    }
+
+    void SpellList::remove (const ESM::Spell* spell)
+    {
+        auto& id = spell->mId;
+        bool changed = withBaseRecord([&] (auto& spells) {
```
akortunov, line 87: Use an Allman bracket style (an opening bracket on the new line) instead.

apps/openmw/mwmechanics/spelllist.cpp
```diff
@@ -0,0 +1,161 @@
...
+    const ESM::Spell* SpellList::getSpell(const std::string& id)
+    {
+        return MWBase::Environment::get().getWorld()->getStore().get<ESM::Spell>().find(id);
+    }
+
+    void SpellList::add (const ESM::Spell* spell)
+    {
+        auto& id = spell->mId;
+        bool changed = withBaseRecord([&] (auto& spells) {
+            for(auto it : spells)
```
akortunov, line 73: Please use brackets to avoid such "stairs". Readablity is not good here.

apps/openmw/mwmechanics/spells.hpp
```diff
@@ -32,37 +30,35 @@ namespace MWMechanics
...
-                std::set<int> mPurgedEffects; // indices of purged effects
-            };
-
-            typedef std::map<SpellKey, SpellParams> TContainer;
-            typedef TContainer::const_iterator TIterator;
-
-        private:
-            TContainer mSpells;
+            std::shared_ptr<SpellList> mSpellList;
+            std::map<const ESM::Spell*, SpellParams> mSpells;
```
akortunov, line 34: A pretty intrusive change. Make sure that you import data from older saves and write it to new saves properly.

apps/openmw/mwmechanics/spells.cpp
```diff
@@ -166,101 +180,78 @@ namespace MWMechanics
...
+    }
 
-        return false;
+    bool Spells::hasBlightDisease() const
+    {
+        return hasDisease(ESM::Spell::ST_Blight);
     }
 
-    void Spells::purgeCommonDisease()
+    void Spells::purge(const std::function<bool(const ESM::Spell*)>& filter)
```
akortunov, line 205: Maybe a typedef would be better here.

apps/openmw/mwmechanics/spelllist.cpp
Assumeru: Better?

akortunov: Better.

apps/openmw/mwmechanics/spells.hpp
Assumeru: I was able to load an older save, resave it, and load it again.

apps/openmw/mwmechanics/spells.cpp
Assumeru: For the template  argument or for the entire thing?

Honestly, not a big fan of typedefs.

akortunov: For the entire argument.

Assumeru: How's this?

akortunov: Looks much better.

Assumeru:
Anyone have any more change requests?

psi29a:
Merge conflicts ;)

Assumeru:
With myself even, my favourite. Rebased.

psi29a:
I see that you also modify the save game format. Are we at least backwards compatible? (old save games will work)

Assumeru:
Of course.

With spells not being stored in the base record older versions just won't be able to find them, hence the increment.

akortunov:
https://gitlab.com/OpenMW/openmw/-/issues/5566

This PR causes issues so far.

PR #2958, opened 2020-07-09
Corrections for StatsWatcher

akortunov:
1. Update HUD-related data which is used by game mechanics and world prior to related updates to initialize data properly.
2. Avoid redundant getEnchantment() call.

### DIFF
--- a/apps/openmw/mwgui/statswatcher.cpp
+++ b/apps/openmw/mwgui/statswatcher.cpp
@@ -136,27 +136,6 @@ namespace MWGui
         }
 
         mWatchedStatsEmpty = false;
-
-        // Update the equipped weapon icon
-        MWWorld::InventoryStore& inv = mWatched.getClass().getInventoryStore(mWatched);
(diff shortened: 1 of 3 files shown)


akortunov:
> Could you provide a reason why this is required?

During savegame loading I sometimes encounter crashes inside preloadSpells() during attempt to read selected magic item, which is stored in GUI and updated in the StatsWatcher. I suppose it is too late to update it during GUI update. Maybe there are better ways to do it, though.

akortunov:
Also an interesting fact - selected weapon, spell and magic item are not actually used in the WindowManager itself - they are used by an external code.

akortunov:
Additional investigation about the crash:
1. It happens in `preloadSpells` during game re-loading when a magic weapon is equipped (I reproduced it with Umbra).
2. The `enchantId` for equipped weapon is malformed during game loading, since data is not initialized yet.
3. It becomes stable reproducable for me, if I do not use `mSpellPreloadTimer` and try to preload spells every frame.

PR #2962, opened 2020-07-11
More NIF-related cleanup

Capostrophic:
I didn't test it thoroughly enough but I'm submitting it anyway for more feedback due to psi29a's advice.

1. NifOsg::CollisionSwitch has been replaced with a simple initial node mask setup. I thought about preserving it again and couldn't come up with a good reason for it; switching the node state in a mod could be as simple as changing the node mask if scene graph access becomes a thing. It's also problematic: it's not currently properly serialized (uses a dummy serializer) and trying to write a serializer for resulted in a lot of redundant code and improper output even though it represents a valid object in the scene graph and basically *must* be serialized, and it's not properly optimized — even if its state is "enabled" it's not going to be handled as a simple NiNode. Serializing seems to work ok now: affected nodes are displayed in SSG output and have the correct intersection disabled node mask.
2. NifOsg::NodeUserData has been replaced with a custom scene graph NifOsg::MatrixTransform node with a simple serializer. There has been a relevant comment by scrawl in the NodeUserData header. Basically everything related to NIF transformations has been moved to its fields so there shouldn't be any headache with trying to work with NIF transformations anymore, and it should also be properly optimized out for nodes that don't have a need for it (e.g. nodes that aren't animated). This makes both the relevant code and the in-game scene graph much cleaner and may result in slightly better performance. There's also the case where NodeUserData objects don't need to be cloned explicitly because they're no longer a thing so no need to work around OSG issues. NIF record index has been moved into a dedicated holder which doesn't need to have any special cases to work around OSG issues.
3. CopyRigVisitor multiple node parents error prints the number of detected parents.
4. NifOsg::UVController has a dummy serializer and no longer causes OSG to log warnings that aren't that useful.

Current concerns: while nothing seemed to be explicitly broken when it comes to custom node introduction I'm not using a lot of mods so it would be nice to get another look on a modded setup (e.g. Morrowind Optimization Patch and mods with animated collision shapes) with optimizer turned on (also a look at the scene graph with the optimizer both turned off and turned on, as NifOsg::MatrixTransform nodes do get optimized into osg::Group nodes).


components/nifosg/matrixtransform.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define OPENMW_COMPONENTS_NIFOSG_MATRIXTRANSFORM_H
+
+#include <components/nif/niftypes.hpp>
+
+#include <osg/MatrixTransform>
+
+namespace NifOsg
+{
+
+    class MatrixTransform : public osg::MatrixTransform
```
elsid, line 11: Why didn't you encapsulate the rotation and scale change into this type? Every time an object of this type is used a developer have to remember and duplicate the code to change underlying 4x4 matrix accoring to mScale and mRotationScale. You could say there is only one but it doesn't make this code less fragile.

akortunov:
```
components/nifosg/matrixtransform.hpp(18): error C2512: 'NifOsg::MatrixTransform::MatrixTransform': no appropriate default constructor available
```

components/nifosg/matrixtransform.hpp
Capostrophic: You mean the dancing around that happens in KeyframeController?

Capostrophic:
That's weird. The using directive should make NifOsg::MatrixTransform inherit all constructors of osg::MatrixTransform. Hm, well, there should only be one (the parameterless one), I guess, so I'll remove the smartassery but leave that one in.

components/nifosg/matrixtransform.cpp
```diff
@@ -0,0 +1,20 @@
+#include "matrixtransform.hpp"
+
+namespace NifOsg
+{
+    MatrixTransform::MatrixTransform(int recordIndex, const Nif::Transformation &trafo)
```
AnyOldName3, line 10: `trafo` is a weird abbreviation. If `transform` is too long, `trans` and `xform` are the conventional shortenings.

Capostrophic: NIF transformation is referred to as trafo in [many other instances within NIF code](https://github.com/OpenMW/openmw/search?q=trafo&unscoped_q=trafo), I'd rather not have them all corrected here.

components/nifosg/matrixtransform.hpp
Capostrophic: Alright, should be done.

Capostrophic:
I don't feel like allowing the optimizer to optimize out NIF node record index is safe (FindByRecIndex will fail to work with such osg::Groups), so I'm going to put the record index back into a dedicated holder again. However, it won't have the side effects NodeUserData presence had, as a record index is immutable so it doesn't need to be cloned.

Capostrophic:
Ok, record index is held in a user object again. Separating it from NifOsg::MatrixTransform allows to keep optimizeable NIF-specific transformations in the custom node (and still avoid issues with cloning transformations) yet make sure transforms that became groups in the result of optimization don't stop being able to be emitter nodes.

psi29a:
Well, let's get this in for more nightly testing.

PR #2966, opened 2020-07-12
Refactor sound initialization

elsid:
1. `init` calls is not so good to read when there is 3-5 float parameters.
2. Default values are defined in a single place.
3. Split volume factor and SFX volume to make it possible to change one of them that is used by https://github.com/OpenMW/openmw/pull/2928.

### DIFF
--- a/apps/openmw/mwsound/sound.hpp
+++ b/apps/openmw/mwsound/sound.hpp
@@ -11,85 +11,65 @@ namespace MWSound
     inline int operator&(int a, PlayMode b) { return a & static_cast<int>(b); }
     inline int operator&(PlayMode a, PlayMode b) { return static_cast<int>(a) & static_cast<int>(b); }
 
+    struct SoundParams
+    {
+        osg::Vec3f mPos;
+        float mVolume = 1;
+        float mBaseVolume = 1;
+        float mPitch = 1;
+        float mMinDistance = 1;
+        float mMaxDistance = 1000;
+        int mFlags = 0;
+        float mFadeOutTime = 0;
+    };
+
     class SoundBase {
         SoundBase& operator=(const SoundBase&) = delete;
         SoundBase(const SoundBase&) = delete;
         SoundBase(SoundBase&&) = delete;
 
-        osg::Vec3f mPos;
-        float mVolume; /* NOTE: Real volume = mVolume*mBaseVolume */
-        float mBaseVolume;
-        float mPitch;
-        float mMinDistance;
-        float mMaxDistance;
-        int mFlags;
-
-        float mFadeOutTime;
+        SoundParams mParams;
 
     protected:
-        Sound_Instance mHandle;
+        Sound_Instance mHandle = nullptr;
 
         friend class OpenAL_Output;
 
     public:
-        void setPosition(const osg::Vec3f &pos) { mPos = pos; }
-        void setVolume(float volume) { mVolume = volume; }
-        void setBaseVolume(float volume) { mBaseVolume = volume; }
-        void setFadeout(float duration) { mFadeOutTime = duration; }
+        void setPosition(const osg::Vec3f &pos) { mParams.mPos = pos; }
+        void setVolume(float volume) { mParams.mVolume = volume; }
+        void setBaseVolume(float volume) { mParams.mBaseVolume = volume; }
+        void setFadeout(float duration) { mParams.mFadeOutTime = duration; }
         void updateFade(float duration)
         {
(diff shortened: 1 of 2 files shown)


apps/openmw/mwsound/sound.hpp
```diff
@@ -11,85 +11,67 @@ namespace MWSound
     inline int operator&(int a, PlayMode b) { return a & static_cast<int>(b); }
     inline int operator&(PlayMode a, PlayMode b) { return static_cast<int>(a) & static_cast<int>(b); }
 
+    struct SoundParams
+    {
+        osg::Vec3f mPos;
+        float mVolumeFactor = 1;
```
akortunov, line 17: if I remembered correctly, some developers complained about such approach due to implicit type conversions.

elsid: But it's absolutely valid thing. There is no precision loss. What is the problem with this? Why `1.f` is better?

akortunov: Check the discussion at #2882. It is not my complaint - your approach is pretty fine for me.

elsid: So it's only an opinion that I simply don't agree with.

psi29a: It wasn't a complaint, it was a suggestion by @Capostrophic though no one else commented further on it. The rest of the conversation instead started discussing about the proper way to skin a cat as there are quite a few ways to do so. 

float mVolumeFactor = 1.f;
or
float mVolumeFactor = 1;

Is fine, there is no reason to be nit-picky here as it is evaluated the same anyway.

kcat: The discussion was over `{...}` uniform initialization syntax, to ensure no implicit conversion or narrowing occurs, rather than `= ...` initialization syntax. Whether it's `= 1` or `= -1.f` on its own doesn't matter much, because the compiler will likely optimize it either way (though uniform initialization syntax does help guarantee the right thing happens). Regardless, what's acceptable depends on what the project's coding style guidelines say.

kcat: More to the patch itself, I don't think there's a need for a third volume field. Since `mBaseVolume` is always the result of `volumeFromType(type)` and the sound params hold the sound type, it can be easily retrieved without being explicitly stored with the sound. The additional `mVolumeFactor` can just take over `mBaseVolume` instead. So the fields would be:
```
float mVolume; // Volume as set in the sound definition
float mBaseVolume; // Caller's volume parameter
...
int mFlags; // Contains the sound type, for volumeFromType
...
float getRealVolume() const noexcept
{
    return mParams.mVolume * mParams.mBaseVolume * volumeFromType(getType(mParams.mFlags));
}
```

psi29a: Right, so this is a non-issue. Moving on.

elsid:
> Since mBaseVolume is always the result of volumeFromType(type) and the sound params hold the sound type, it can be easily retrieved without being explicitly stored with the sound. The additional mVolumeFactor can just take over mBaseVolume instead.

`volumeFromType` is not a free function. I'd need to make `Sound` depending on `VolumeSettings`: https://github.com/OpenMW/openmw/blob/ddfad83df29adc2a609af48553db7937abf40c85/apps/openmw/mwsound/volumesettings.cpp#L26-L50. Exchange one field for a new dependency+more CPU time for a call. I don't see the benefit.

kcat:
I think it would be possible to combine the `Sound_Buffer`'s volume with the sound type volume then, as the `Sound_Buffer`'s volume should never change, and the sound type volume only changes when changing volume settings (where any active sound should have its `Sound_Buffer` to reapply its volume). So basically:
```
mVolume = volume;
mBaseVolume = volumeFromType(type) * sfx->mVolume;
```
...
```
void SoundManager::processChangedSettings(const Settings::CategorySettingVector& settings)
{
    ....
    for(SoundMap::value_type &snd : mActiveSounds)
    {
        for(SoundBufferRefPair &sndbuf : snd.second)
        {
            Sound *sound = sndbuf.first;
            Sound_Buffer *sfx = sndbuf.second;
            sound->setBaseVolume(volumeFromType(sound->getPlayType()) * sfx->mVolume);
            mOutput->updateSound(sound);
        }
    }
    ....
}
```

elsid:
I've removed `mVolume` split. Looked more how `Sound_Buffer` is initialized. Probably this change is not required.

PR #2967, opened 2020-07-12
Fix search filter unicode character case sensitivity

Capostrophic:
As requested [here](https://gitlab.com/OpenMW/openmw/-/issues/5525).

Unicode variant of lowercase-ing previously only used for spell window search filter is used for instances of comparing item names in item and ingredient effect sorting/filtering. Fixes case sensitivity of filtering of potions that have a Russian name, should make sorting of items with names in different cases consistent between languages. Not tested on a Russian installation.


psi29a:
Brilliant. :)

PR #2968, opened 2020-07-13
Adjust actor rotation when flee by axis x (bug #5531)

elsid:
https://gitlab.com/OpenMW/openmw/-/issues/5531

Otherwise Cliff Racers can go underwater.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
     Bug #5499: Faction advance is available when requirements not met
     Bug #5502: Dead zone for analogue stick movement is too small
     Bug #5507: Sound volume is not clamped on ingame settings update
+    Bug #5531: Actors flee using current rotation by axis x
     Feature #390: 3rd person look "over the shoulder"
     Feature #2386: Distant Statics in the form of Object Paging
     Feature #5297: Add a search function to the "Datafiles" tab of the OpenMW launcher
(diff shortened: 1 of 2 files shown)


akortunov:
Are you sure that the issue is flee-specific? From what I can tell, flying followers can go underwater as well in OpenMW, the same thing for creatures in combat, when they try to pursue player. This behaviour is not present in 0.47.

elsid:
For the case with Cliff Racers I'm sure. When actors flee blindly they use their last rotation by x when movement is applied in movement solver. So velocity can direct them undewater. And this should not fix followers because combat flee is completely unrelated. Maybe we need more general approach, but not sure which one. We can try to block flying creatures to go underwater in the physics system, but this may cause strage behaviour when they are directed there by other systems. Or none of AI packages should control movement and rotation directly when it's unknown where is a destination point. And a destination should be selected or adjusted to be not underwater when creatures should not go there. At least we need bugs to be filed for all cases. I don't know what do you mean by "some cases" that work in 0.46 but don't in 0.47.

akortunov:
> I don't know what do you mean by "some cases" that work in 0.46 but don't in 0.47.

In 0.46 cliff racers do not go underwater, probably due to limitations for flying characters. 
Your patch affects only combat AI, but, for example, flying followers still go underwater in 0.47, despite they seem to do not do that in 0.46.

elsid:
> In 0.46 cliff racers do not go underwater

Did you watch the video in the bug. It's recorded with 0.46. The do go undewater.

elsid:
> Your patch affects only combat AI. For example, flying followers still go underwater in 0.47 with your patch

Yes, because it's the only complaint so far. Again I don't know about other cases. If you know them, file a bug.

psi29a:
Thanks @elsid, the scope of this PR is fine. Additional issues can be tackled in other issues/MRs.

PR #2969, opened 2020-07-14
Resume failed scripts execution after game reload

akortunov:
Implements [feature #5524](https://gitlab.com/OpenMW/openmw/-/issues/5524).

Use a simple approach so far - once script execution fails during runtime, do not remove its bytecode, mark it as inactive instead.
When try to run a script, early-out when it is marked as inactive.
This flag is cleaned up during scripting system cleanup, during new game start or savegame loading.

Later it should be possible to per-instance blocks, but it is unclear how exactly it should work.

Note that I do not see the code, which removes data from `mScripts` map, so I assume that data persists for the whole game session.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
     Feature #5445: Handle NiLines
     Feature #5457: Realistic diagonal movement
     Feature #5486: Fixes trainers to choose their training skills based on their base skill points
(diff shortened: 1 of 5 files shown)


psi29a:
It looks alright to me, waiting on AV. Thanks @akortunov

PR #2970, opened 2020-07-17
[macOS, CI] Use Xcode 11.6

nikolaykasyanov:
Let's see how it affects build times.


psi29a:
That looks better even with cache hit rate                      0.00 %

psi29a:
Turn openmw-cs back on and see what happens?

PR #2971, opened 2020-07-18
Separate sound buffer pool from sound manager

elsid:
Also:
1. Track used buffers.
2. Encapsulate usage counter by a custom type. It's possible to use `std::shared_ptr` but it will add overhead.
3. Different types for a reference to a used and unused sound buffer.
4. Store sound buffers in the object pool.

### DIFF
--- a/apps/openmw/CMakeLists.txt
+++ b/apps/openmw/CMakeLists.txt
@@ -58,6 +58,7 @@ add_openmw_dir (mwscript
 add_openmw_dir (mwsound
     soundmanagerimp openal_output ffmpeg_decoder sound sound_buffer sound_decoder sound_output
     loudness movieaudiofactory alext efx efx-presets regionsoundselector watersoundupdater volumesettings
+    sound_buffer
     )
 
 add_openmw_dir (mwworld
--- a/apps/openmw/mwsound/sound.hpp
+++ b/apps/openmw/mwsound/sound.hpp
@@ -7,6 +7,13 @@
 
 namespace MWSound
 {
+    // Extra play flags, not intended for caller use
+    enum PlayModeEx
+    {
+        Play_2D = 0,
+        Play_3D = 1 << 31,
+    };
+
     // For testing individual PlayMode flags
     inline int operator&(int a, PlayMode b) { return a & static_cast<int>(b); }
     inline int operator&(PlayMode a, PlayMode b) { return static_cast<int>(a) & static_cast<int>(b); }
--- a/apps/openmw/mwsound/sound_buffer.cpp
+++ b/apps/openmw/mwsound/sound_buffer.cpp
@@ -1,0 +1,165 @@
+#include "sound_buffer.hpp"
+
+#include "../mwbase/environment.hpp"
+#include "../mwbase/world.hpp"
+#include "../mwworld/esmstore.hpp"
+
+#include <components/debug/debuglog.hpp>
+#include <components/settings/settings.hpp>
+#include <components/vfs/manager.hpp>
+
+#include <algorithm>
+#include <cmath>
+
+namespace MWSound
+{
+    namespace
+    {
+        struct AudioParams
+        {
+            float mAudioDefaultMinDistance;
+            float mAudioDefaultMaxDistance;
+            float mAudioMinDistanceMult;
+            float mAudioMaxDistanceMult;
+        };
+
+        AudioParams makeAudioParams(const MWBase::World& world)
+        {
+            const auto& settings = world.getStore().get<ESM::GameSetting>();
+            AudioParams params;
+            params.mAudioDefaultMinDistance = settings.find("fAudioDefaultMinDistance")->mValue.getFloat();
+            params.mAudioDefaultMaxDistance = settings.find("fAudioDefaultMaxDistance")->mValue.getFloat();
+            params.mAudioMinDistanceMult = settings.find("fAudioMinDistanceMult")->mValue.getFloat();
+            params.mAudioMaxDistanceMult = settings.find("fAudioMaxDistanceMult")->mValue.getFloat();
+            return params;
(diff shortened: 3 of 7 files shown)


apps/openmw/mwsound/soundbufferpool.hpp
```diff
@@ -0,0 +1,145 @@
...
+            SoundBufferSharedRef use() noexcept { return SoundBufferSharedRef(*mPool, *mValue); }
+
+            Sound_Buffer* get() const noexcept { return mValue; }
+
+            Sound_Buffer* operator->() const noexcept { return mValue; }
+
+        private:
+            SoundBufferPool* mPool = nullptr;
+            Sound_Buffer* mValue = nullptr;
+    };
```
kcat, line 137: What's the point of this class? `Sound_Buffer` objects should remain valid for the lifetime of the sound manager, and are automatically cleared with it. It's the internal handle, which stores the actual sample data for the output backend, that gets loaded and unloaded as needed.

elsid: `Sound_Buffer` doesn't know what pool does it belong to. This type provides this information. Also it represents a state of `Sound_Buffer` when it's already allocated but sound is not playing yet. Between [this](https://github.com/OpenMW/openmw/blob/94b9d8f67af7a35a0a6813c85f4e11d0fcfe20ee/apps/openmw/mwsound/soundmanagerimp.cpp#L509) and [this](https://github.com/OpenMW/openmw/blob/94b9d8f67af7a35a0a6813c85f4e11d0fcfe20ee/apps/openmw/mwsound/soundmanagerimp.cpp#L529) line.

And I don't understand what do you suggest instead. I can move this into `Sound_Buffer` since it already knows about iterator from the pool.

> and are automatically cleared with it

Before this change it's handled by `SoundManager` with code like:
```c++
if (sfx->mUses-- == 1)
    mUnusedBuffers.push_front(sfx);
```
and
```c++
if (sfx->mUses++ == 0)
{
    SoundList::iterator iter = std::find(mUnusedBuffers.begin(), mUnusedBuffers.end(), sfx);
    if(iter != mUnusedBuffers.end())
        mUnusedBuffers.erase(iter);
}
```
that is copy-pasted over multiple places. I don't see why RAII can't be used for internal handles to avoid code duplication and make a usage of `Sound_Buffer` objects safer.

kcat: > `Sound_Buffer` doesn't know what pool does it belong to.

It doesn't need to, since it exists for the life of its pool/allocating container. And when the pool is destroyed, nothing should have a `Sound_Buffer`.

> Before this change it's handled by SoundManager with code like:
> ...
> and
> ...
> that is copy-pasted over multiple places. I don't see why RAII can't be used for internal handles to avoid code duplication and make a usage of `Sound_Buffer` objects safer.

Private methods can be added to `SoundManagerImpl` to avoid the duplication. As it is, `mUnusedBuffers` only exists to expediate unloading of least-recently-used samples buffers. And managing which sample buffers to unload, when, and how, seems like the job of the sound manager to me, rather than the pool for `Sound_Buffer`s (which holds the sample buffers). `SoundBufferPool`, `SoundBufferRef`, and `SoundBufferSharedRef` seems way too over-designed to me and is putting management work where it shouldn't be.

elsid:
Removed `SoundBufferRef` by moving the logic into `Sound_Buffer`. Also had to move `SoundBufferPool` and `SoundBufferSharedRef` to `sound_buffer.hpp` to save benefits from inlining without putting `Sound_Buffer::use` into pool header.

apps/openmw/mwsound/soundbufferpool.hpp
elsid: `SoundManagerImpl` is overwhelmed with logic from a selection of the next music track from a playlist to play to a reference counting. I don't know what are the benefits of doing this. Reference counting for a `Sound_Handle` and playing/stopping sounds are different responsibilities. Sound manager still manages when to unload the sound by `SoundBufferSharedRef ` destruction.

kcat: > `SoundManagerImpl` is overwhelmed with logic from a selection of the next music track from a playlist to play to a reference counting.

The next music track/playlist handling can certainly go into a sub-manager of sorts. I believe the intention is to eventually allow for mod-defined playlists. The sound manager just needs to make sure the right playlist is being used, and ensure a track from that playlist plays. What's currently there is an attempt to randomize tracks while avoiding the same track playing twice, basically a fallback dynamic playlist creator.

The reference counting in `Sound_Buffer` is so the sound manager knows when a sound buffer is no longer used, and becomes the next candidate for unloading after any previously-made-unused sounds when the cache fills up. On its own, it doesn't do anything except to indicate 'this is now unused' or 'this is now used'. Since the sound manager already handles starting and stopping sounds, and handles the output backend, it's the most appropriate place to say when a sound is no longer in use or is now in use, and to load and unload sounds from the backend as needed.

akortunov:
> I do not think this is a good change

Is there a conclusion? Should we merge this PR or close it?

elsid:
I don't agree that a single class should mix low level and high level logic. The idea of type system is to divide and encapsulate logic. So when there is a new unrelated logic or a refactoring no need to care about unrelated level stuff. Like in https://github.com/OpenMW/openmw/pull/2928 I had to be careful with this reference counting stuff. OpenMW is a C++ project and it should take advantage of this. RAII is not just about memory. Sometimes non trivial object is a resource and is needed to be treated accordingly. Anyway I don't want to push this because some parts of this pr I don't like myself, but without any ideas how to make it better I don't want to see it in the master.

psi29a:
It is something that we can come back to later when a light-bulb goes off.

PR #2978, opened 2020-07-29
Don't encapsulate NIF transformation changes

Capostrophic:
Although the idea is fundamentally good, in akortunov's and Abdu's testing the way I tried to implement it seems to be subtly broken and causes odd rendering issues. I managed to reproduce one myself.

I'm reverting [the commit](https://github.com/OpenMW/openmw/commit/f93655e803d1b611f53477872f41a37eac5a8679) for now as currently NIF transformations are handled in one place (KeyframeController implementation) and it'll stay that way for a while.

### DIFF
--- a/components/nifosg/controller.cpp
+++ b/components/nifosg/controller.cpp
@@ -119,23 +119,48 @@ void KeyframeController::operator() (osg::Node* node, osg::NodeVisitor* nv)
     if (hasInput())
     {
         NifOsg::MatrixTransform* trans = static_cast<NifOsg::MatrixTransform*>(node);
+        osg::Matrix mat = trans->getMatrix();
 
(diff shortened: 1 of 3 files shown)


elsid:
But what's broken? Reverting refactoring that was requested to https://github.com/OpenMW/openmw/pull/2962 be approved doesn't sound right. Why do you think this fixed something if you don't know what is broken? If something was broken by that pr revert all commits.

akortunov:
> But what's broken?

Weapon meshes in the 1st-person are broken, at least when bodypart replacers are used:
Screenshot_20200729_204910

Since we failed to find a cause of issue, we decided to revert the whole commit to do not keep master branch in the broken state until the issue is solved.

PR #2980, opened 2020-07-29
Make scripts inherit variables from the base record

Assumeru:
Split from #2870. 

Feels like it can't be this easy after the previous bunch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
     Bug #3676: NiParticleColorModifier isn't applied properly
     Bug #3714: Savegame fails to load due to conflict between SpellState and MagicEffects
     Bug #4021: Attributes and skills are not stored as floats
+    Bug #4055: Local scripts don't inherit variables from their base record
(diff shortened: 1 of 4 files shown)


Assumeru:
I know it's not technically the base record, but both issues call it that: https://gitlab.com/OpenMW/openmw/-/issues/4055 https://gitlab.com/OpenMW/openmw/-/issues/2798

Anyway, the idea is that `set script_name.variable to 1` writes to the "base record" and that local scripts inherit from it in the sense that their variables are initialised with the global script's values. I doubt anyone tried to make a script both a local and a global script and it certainly looks odd when viewed as such. Note that no global script needs to be running and that global scripts cannot be initialised in this manner because `getLocalsIfPresent` will return a nullptr.

I should probably go back and add some comments to the AddSpell stuff now that you mention it.

zinnschlag:
Okay. That makes more sense. I reopened the first issue you linked. Needs changelog entry and stuff.

PR #2985, opened 2020-08-06
Fix regression #5572

Assumeru:
https://gitlab.com/OpenMW/openmw/-/issues/5572

This won't fix the save file in the issue, but it shouldn't happen again after resaving.


Capostrophic:
Travis complains about parentheses.
```
/home/travis/build/OpenMW/openmw/apps/openmw/mwmechanics/spells.cpp:438:62: warning: '&&' within '||' [-Wlogical-op-parentheses]
```

Assumeru:
Fixed.

Since this is the second time I've done that, I'd be in favour of disabling that check :P

PR #2989, opened 2020-08-09
Make Advanced tab of the launcher look okay for everyone again

Capostrophic:
Workaround for the broken look of vertical tabs on some setups proposed [here](https://gitlab.com/OpenMW/openmw/-/issues/5567#note_392876462), seems that the current workaround doesn't work as intended. Horizontal tabs Qt does properly support look good enough IMO even though they're not reminiscent of MCP's patcher UI. At least Graphics tab and Advanced tab would look consistent.
изображение
Some tabs were renamed.

### DIFF
--- a/apps/launcher/advancedpage.cpp
+++ b/apps/launcher/advancedpage.cpp
@@ -10,37 +10,6 @@
 
 #include <cmath>
 
-
-class HorizontalTextWestTabStyle : public QProxyStyle
-{
-public:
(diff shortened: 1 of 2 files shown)


akortunov:
I'd suggest also to split Visuals tab to two columns:
Screenshot_20200813_135324

akortunov:
Something like this:

Screenshot_20200813_140843

Just use the Grid layout.

PR #2991, opened 2020-08-21
Use two columns when there is too many checkboxes in launcher's tab

akortunov:
Related to [bug #5567](https://gitlab.com/OpenMW/openmw/-/issues/5567).

At the moment there is just too many options in some tabs to show them with default window size, but many space is unused, so I use two columns now:

Screenshot_20200821_125338
Screenshot_20200821_125344


psi29a:
lgtm, thanks @akortunov

PR #2998, opened 2020-09-01
Avoid crash when object paging encounters an empty shape

akortunov:
Should fix [regression #5578](https://gitlab.com/OpenMW/openmw/-/issues/5578).

Empty drawables have no vertex array, so we should check if array exists before try to get number of its elements.
Note that bzzt already uses such check in the `AddRefnumMarkerVisitor`.


psi29a:
Nice find

PR #3002, opened 2020-09-04
Fix SummonKey comparison function

akortunov:
In some cases it did not work properly.

### DIFF
--- a/components/esm/magiceffects.hpp
+++ b/components/esm/magiceffects.hpp
@@ -39,9 +39,13 @@ namespace ESM
         {
             if (mEffectId < other.mEffectId)
                 return true;
+            if (mEffectId > other.mEffectId)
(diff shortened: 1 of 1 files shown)


elsid:
I would be nice to explain in which cases and wait for review. Even if it's very simple case that no one need to know about there is still a chance of mistake. There is a quite good technique to avoid these kind of mistakes - use `std::tie`:
```c++
std::tie(mEffectId, mSourceId, mEffectIndex) < std::tie(other.mEffectId, other.mSourceId, other.mEffectIndex);
```

akortunov:
> I would be nice to explain in which cases and wait for review.

Search in std::map did not work properly due to error in the `<` operator for `SummonKey` struct ("larger" case was treated as "equals").

PR #3004, opened 2020-09-09
Attach arrows to actor's hands

akortunov:
Just attach arrows to shooter's hand instead of ArrowBone node in bow's mesh when the "Bip01 Arrow" bone is present in shooter's skeleton. With vanilla assets there will be no difference from our current behaviour.

This feature will allow modmakers to make better shooting animations. [Here](https://github.com/OpenMW/openmw/files/5193002/Left.zip) is an example of mod, which can use such feature - it places bow to left hand instead of right one, so it can not rely on ArrowBone. This mod is based on Weapon Sheathing. Note that the mod is WIP and has some issues (such as spam about missing 'tri w_longbow_daedric002' bone, which is a bug in animations themselves).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@
     Feature #5545: Option to allow stealing from an unconscious NPC during combat
     Feature #5579: MCP SetAngle enhancement
     Feature #5610: Actors movement should be smoother
+    Feature #5642: Ability to attach arrows to actor skeleton instead of bow mesh
     Task #5480: Drop Qt4 support
     Task #5520: Improve cell name autocompleter implementation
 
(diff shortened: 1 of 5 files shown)


psi29a:
With a feature like this, we should have at least an Changelog entry and issue explaining this for those that are looking and wanting to use this feature.

As talked about on Discord:

> [8:25 PM] psi29a: About the attach arrows thingy, does that have implications outside of OpenMW or is this something specific to OpenMW?
> [8:25 PM] psi29a: I see that we fall back to old behaviour of the new bone you're searching for isn't available?
> [8:29 PM] akortunov: Yes, there is a vanilla behaviour by default.
> [8:30 PM] akortunov: Without this patch it is just not possible to set different arrow fetching animations for different races.
> [8:30 PM] akortunov: For example, beast races have tail, and quivers should be attached with different angle.

akortunov:
Done.

PR #3001, opened 2020-09-03
[WIP] Container base record mutations

Assumeru:
Continuation of #2870 https://gitlab.com/OpenMW/openmw/-/issues/2798

Some of these might get fixed along the way
https://gitlab.com/OpenMW/openmw/-/issues/2404
https://gitlab.com/OpenMW/openmw/-/issues/3862
https://gitlab.com/OpenMW/openmw/-/issues/3929

PR #3005, opened 2020-09-15
Don't search containers

Assumeru:
Tangentially related to #3001, mostly as premature optimization.

OpenMW currently searches containers in a few cases: AiActivate, cast, aipackage, and script targeting.

AiActivate seems to bug out and make actors walk towards (0,0);
AiPackage is looking for an actor;
And vanilla doesn't allow scripts to target items in containers from what I can tell.

So in short, there appears to be no reason to ever search them and we might as well gain some performance.

### DIFF
--- a/apps/openmw/mwworld/worldimp.hpp
+++ b/apps/openmw/mwworld/worldimp.hpp
@@ -287,9 +287,9 @@ namespace MWWorld
             ///< Return a pointer to a liveCellRef with the given name.
             /// \param activeOnly do non search inactive cells.
 
-            Ptr searchPtr (const std::string& name, bool activeOnly, bool searchInContainers = true) override;
+            Ptr searchPtr (const std::string& name, bool activeOnly, bool searchInContainers = false) override;
             ///< Return a pointer to a liveCellRef with the given name.
-            /// \param activeOnly do non search inactive cells.
+            /// \param activeOnly do not search inactive cells.
 
             Ptr searchPtrViaActorId (int actorId) override;
             ///< Search is limited to the active cells.


akortunov:
> And vanilla doesn't allow scripts to target items in containers from what I can tell.

It needs verification.

Assumeru:
> It needs verification.

Of course. So far I've only done a little check using the console. Using a resolved container containing an item with references persist checked.

Assumeru:
I attached a script to a container that checks if a unique (persistent) item inside the container is activated (OnActivate). Starting the game gives a warning about the item not being found - after which the script won't run.

Assumeru:
I've now also created a scripted weapon. The script on the weapon checks if a global script is running, and if not starts it. The global script reports GetHealth in a messagebox every 1.5s.

I'm fairly sure this is the only way to get a global script running on an item in an inventory/container because a script started on an item in the world will keep running on the removed ground instance after the player picks it up.

The game crashed.

psi29a:
I'm okay with pulling this in if you've verified that Morrowind.exe does the same thing. Worst case scenario, we get a bug-report relating to why a scripted item isn't running while inside a container.

Assumeru:
> why a scripted item isn't running while inside a container

For that specific case I'll just refer to: https://github.com/OpenMW/openmw/pull/3001#issuecomment-691721063
Morrowind seems pretty wonky when it comes to scripted items not in the world.

zinnschlag:
Hm, not great. Historically speaking we have usually extended broken/limited vanilla MW functionality into something that does work consistently and reliably. I am okay with this PR as a temporary fix. But in the long run items in containers should become accessible that way again. Not having this feature would put a severe limit on what mods can do or at least require very ugly workarounds, that the average mod developer is not capable of.

Assumeru:
Agreed, in principle. I do think the default for searching containers should be false, however. The current situation is a trap.

If/when we want to extend functionality it should be on a case-by-case basis. ExplicitRef and ImplicitRef can easily be altered to forward the search argument which'd give us selective control per function.

PR #2473, opened 2019-08-05
dont teleport following hostile (bug #5101)

mp3butcher:
https://gitlab.com/OpenMW/openmw/issues/5101

PR #3006, opened 2020-09-20
Rework container resolution

Assumeru:
Split from #3001.
This is just the container changes without the base record stuff that PR aims to solve.

Quick overview:
* ContainerStore now preserves negative quantities from the base record.
* A container's ContainerStore can be in two states - resolved and unresolved. Resolved means some change was made to its contents, these get saved. Unresolved stores do not contain levelled items and do not get saved.
* Opening a container temporarily resolves it. If no changes are made to it before the window is closed, it goes back to being unresolved.

Galbedir in the Balmora Mages Guild provides a nice test case. She has a chest containing levelled items with negative quantities. By using SetLevel on the player it's possible to change what she sells/what's in the container.


zinnschlag (commented):
Well, that is one big PR. I am sorry, but I don't have the time for a thorough review. I did look it over and I didn't find anything obvious wrong.

I have one question and a few general comments. The question: Opening a container temporarily resolves it and then the container is unresolved on closing if no change was made. Does this consider the case of the container changed from a different source (e.g. a script) while the container is open?

Generally speaking PRs of this size aren't really practical for review. For future reference I suggest more splitting up where possible. Maybe a "prequel" PR that cleans things up in preparation for the main change? Depends on the situation, obviously. But 62 changed files are just too much.

Also, the fact that a feature change like this requires changing 62 files may indicate that there is need for more abstraction and maybe more general code maintenance work. Changing one feature should not require modifying half the code base.

akortunov:
I do not like the split to mutable and immutable storages - it is prone to errors.

Also in Morrowind if player took an item from container, it becomes resolved and stops to restock.
If your PR replicates vanilla behavious here, it is unclear how to treat existing container records from older saves - if they are counted as resolved, restocking for visited merchants may become broken.

As for overstocking, it is an exploit, and psi29a does not like to add exploits.

apps/openmw/mwclass/container.cpp
```diff
@@ -20,57 +20,186 @@
...
+
+        virtual MWWorld::ContainerStore& getMutable() override;
+
+        virtual const MWWorld::ContainerStore& getImmutable() const override;
+    };
 
-        virtual MWWorld::CustomData *clone() const;
+    ContainerCustomData::ContainerCustomData(const ESM::Container& container, MWWorld::CellStore* cell)
+    : mUnresolvedStore(std::make_unique<MWWorld::ContainerStore>())
+    , mSeed(Misc::Rng::rollDice(std::numeric_limits<int>::max())) {
```
akortunov, line 91: It is unclear why you need a separate seed and separate random generator just for every container. IMO, it is an overkill.

apps/openmw/mwclass/container.cpp
```diff
@@ -273,6 +400,7 @@ namespace MWClass
             text += "\n#{sUnlocked}";
         if (ptr.getCellRef().getTrap() != "")
             text += "\n#{sTrapped}";
+        const auto customData = ptr.getRefData().getCustomData();
```
akortunov, line 403:
```
/home/travis/build/OpenMW/openmw/apps/openmw/mwclass/container.cpp: In member function ‘virtual MWGui::ToolTipInfo MWClass::Container::getToolTipInfo(const MWWorld::ConstPtr&, int) const’:
/home/travis/build/OpenMW/openmw/apps/openmw/mwclass/container.cpp:403:20: warning: unused variable ‘customData’ [-Wunused-variable]
  403 |         const auto customData = ptr.getRefData().getCustomData();
      |                    ^~~~~~~~~~
```

akortunov (commented):
Also there is a lot of Clang warnings about vertual methods in your PR.

Assumeru:
> Opening a container temporarily resolves it and then the container is unresolved on closing if no change was made. Does this consider the case of the container changed from a different source (e.g. a script) while the container is open?

Opening a container and then adding/removing an item by script permanently marks the container as resolved regardless of what the player ends up doing. This is meant to make a trick (for forcing resolution) listed in MSFD work.

> Generally speaking PRs of this size aren't really practical for review.

Yeah, I'm not very happy about it. I didn't see a good way of splitting it up.

> I do not like the split to mutable and immutable storages - it is prone to errors.

I'm open to suggestions.

> Also in Morrowind if player took an item from container, it becomes resolved and stops to restock.
If your PR replicates vanilla behavious here, it is unclear how to treat existing container records from older saves - if they are counted as resolved, restocking for visited merchants may become broken.

That's not true, yes it becomes resolved, and yes that prevents further levelled list resolution, but it doesn't change negative quantities to positive ones. Players can still buy infinite stuff. My attempt at fixing old saves is making the content resolved but setting negative quantities to preserve restocking.

> As for overstocking, it is an exploit, and psi29a does not like to add exploits.

Fun fact, I wasn't gonna implement it. Turns out it's fairly natural behaviour once you set a few requirements.

apps/openmw/mwclass/container.cpp
```diff
@@ -20,57 +20,186 @@
...
+            removeScripts(*mUnresolvedStore);
+            addScripts(*mResolvedStore, cell);
         }
-        virtual const ContainerCustomData& asContainerCustomData() const
+        return *mResolvedStore;
+    }
+
+    bool ContainerCustomData::isModified() const
+    {
+        return !!mResolvedStore;
```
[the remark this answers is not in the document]
Assumeru: JS habits

Assumeru:
> Also there is a lot of Clang warnings about vertual methods in your PR.

The override stuff? I don't mind fixing those but note that those are methods I didn't change.

apps/openmw/mwclass/container.cpp
Assumeru: The seed is so it resolves to the same contents every time it's opened - unless the player's level changed. Which is vanilla behaviour.

I figured I'd create Rng instances instead of changing the global number generator's seed so the player doesn't end up messing with random number generation by opening containers.

akortunov: > The seed is so it resolves to the same contents every time it's opened

I suppose you can just resolve items once, but do not store them in savegames. Approach with separate random generator for every container looks like a hack.

akortunov:
> I'm open to suggestions.

I suggest to do not touch an existing public API and hide containers resolving from client code. It should also make this PR more clean.

akortunov:
> but note that those are methods I didn't change.

There are [no](https://travis-ci.org/github/OpenMW/openmw/jobs/728853341) such warnings in upstream master, so these warnings are more likely caused by this PR.

apps/openmw/mwclass/container.cpp
Assumeru: That doesn't work because the contents can still change. Levelled list resolution is level-dependent (obviously), if the player opens a container it will always contain the same items provided the game hasn't been restarted and the player's level hasn't changed.

Assumeru:
> so these warnings are presumably caused by this PR.

Then it's probably because I marked some (other) methods override in the same file. Bit weird.

> And such containers are not restocked in Morrowind.

As I said, that's not true. They are no longer resolved.

> without investigation how restocking actually works in Morrowind.

I don't know why you would assume that.

> so it would be better to hide details of your implementation behind that interface.

This would require a method on ContainerStore to resolve the store, which isn't an issue by itself, but that moves the burden of deciding when to resolve a store to the user of the interface. Do you think that's preferable?

akortunov:
>  but that moves the burden of deciding when to resolve a store to the user of the interface

Actually. the situation is opposite.

For example, I write a code to handle interation with door. I need to check if actor has a key for that door.
With your code I need to decide if I need to use a mutable storage or immutable one, which behave differently. To decide it, I should know about details of your implementation.
With upstream master I can just use a `search` method from `ContainerStore` and I do not care if inventory is temporary or persistent and do not care about resolving mechanics. There are not too many cases when such details actually matter in the user code.

apps/openmw/mwclass/container.cpp
akortunov: > Levelled list resolution is level-dependent (obviously)

And? You can store current container level as a temporary field and re-generate content on opening when player's level is changed.

It is still better than re-generate the whole content with its scripts on every access.

Assumeru:
> To decide it, I should know about details of your implementation

I agree with everything you said, except this bit. It's no different from `cbegin` vs `begin` when using iterators.

I'll go move stuff into ContainerStore now.

apps/openmw/mwclass/container.cpp
Assumeru: As it stands resolution is a rare event so that sounds more like premature optimization, but yeah it could work.

akortunov:
> I don't know why you would assume that.

I mean that an only info about seeds comes from wareya:

> In vanilla it seems like containers have a seed that's reset when you load them, but if you open the container, reseeding is blocked in the future.

There is no confirmation yet, and I doubt that Bethesda would implement something like that in year 2001 or so. Remember that CPU's were quite slow in that age (the 500mhz Intel Pentium III is the minimum system requirement), and "generate content every time when you access container" is not a really fast approach.

Assumeru:
That's actually pretty easily observed. Just `tfh` and open a container twice.

psi29a:
Greatness7 already confirmed it as well.
https://discord.com/channels/260439894298460160/260443767482482688/758022261520466062

Assumeru:
Alright, this PR has hit the gym and slimmed down a bunch.

apps/openmw/mwscript/miscextensions.cpp
```diff
@@ -674,7 +674,7 @@ namespace MWScript
                             MWWorld::ConstContainerStoreIterator it = store.getSlot (slot);
                             if (it != store.end() && ::Misc::StringUtils::ciEqual(it->getCellRef().getRefId(), item))
                             {
-                                numNotEquipped -= it->getRefData().getCount();
+                                numNotEquipped -= std::abs(it->getRefData().getCount());
```
akortunov, line 677: A yet another public API change, that bloats PR size. Now count can be negative.

Probably it would be better to make `getCount()` to return positive value and add an another API to work with restockable items. Probably you can add an optional boolean flag to this method to allow client code to tell if it wants to work with restocking.

Assumeru: I think I originally wanted to avoid changing getCount because non-store refs might already have negative counts. Not my brightest moment.

apps/openmw/mwworld/containerstore.cpp
```diff
@@ -894,8 +897,17 @@ void MWWorld::ContainerStore::readState (const ESM::InventoryState& inventory)
         }
     }
 
-
-    mLevelledItemMap = inventory.mLevelledItemMap;
+    // Fix old saves
+    for(const auto& entry : inventory.mLevelledItemMap)
```
akortunov, line 901: We usually implement such things on components level (for example, in the `ESM::InventoryState::load`). In this case you can check ESM version, so you can skip such fallback iterations for new saves.

components/misc/rng.cpp
```diff
@@ -5,27 +5,33 @@
 
 namespace Misc
 {
+    Rng Rng::sInstance{};
```
akortunov, line 8: Probably there are ways to simplify random generator. Now it looks fishy and provides two sets of API - one set for your purpose and one for static random generator instance, which is used in the rest of the engine.

Assumeru: Maybe something like this?

akortunov: Yes, something like this.

Assumeru:
Anyone have any more comments?

psi29a:
Trigger another build, make sure AppVeyor is happy (now that we fixed dep downloads)?

@akortunov @zinnschlag ?

akortunov:
> akortunov

I do not see something obviously wrong, but I did not test this PR, so I do not know if it works properly.

psi29a:
I'm not particularly a fan of **Bug #2473: Unable to overstock merchants**, I know that it just happens to occur naturally because of how our system works.

Just out of curiosity, would there be a way to stop that from happening? Not a requirement for this PR. I think we can talk about the merits of having an option to prevent an overstocking exploit where a player sells things to the vendor in attempt to increase the 'max' amount of restock, which I think breaks the atmosphere (and the simulated economy) of game.

Assumeru:
> Just out of curiosity, would there be a way to stop that from happening?

Yes. It should be possible to create two stacks, one with a negative quantity and one with a positive quantity. Anything the player sells would be added to the positive stack. There's probably some caveats, but it should be doable.

Assumeru:
> Trigger another build, make sure AppVeyor is happy (now that we fixed dep downloads)?

It appears to be unhappy about c9e3db40a1c30b6a021f36e4c587163e35af9cd4 renaming the font license.

AnyOldName3:
Looks like a review comment was misunderstood and the wrong whitespace was removed. I've merged a fix, so try again in a few minutes once the GitLab->GitHub sync has happened.

Assumeru:
Travis apparently fails on a shader related unit test.

AnyOldName3:
Yeah, that's a pre-existing issue. Ignore it for now.

Assumeru:
@psi29a want me to squash it?

akortunov:
Also it would be nice to have an ability to add levelled items to container via `additem` and remove them via `removeitem`, maybe as a follow-up PR.

Assumeru:
> Also it would be nice to have an ability to add levelled items to container via `additem` and remove them via `removeitem`, maybe as a follow-up PR.

Yeah, I'll probably end up implementing that as part of the base record PR. We should get at least partial support from that one even without special handling.

psi29a:
Great work @Assumeru , thanks. I'll squash it here.

akortunov:
Looks like this PR has introduced critical regressions: https://gitlab.com/OpenMW/openmw/-/issues/5641

PR #3008, opened 2020-09-26
Rework fixed strings handling

akortunov:
Summary of changes:

1. NAME32 and NAME64 should be null-terminated (see [this](https://gitlab.com/OpenMW/openmw/-/issues/5615) thread). It means that if fixed string has size 32 bytes, it should take up to 31 characters from input string in the editor, 32th byte should be null-teminator. Since `assign ()` only used in the editor and unit tests, it should be no difference with content files, created via TES-CS.
2. NAME256 is not used in the engine now, so remove it.
3. `data_size()` is static now.

### DIFF
--- a/apps/openmw_test_suite/esm/test_fixed_string.cpp
+++ b/apps/openmw_test_suite/esm/test_fixed_string.cpp
@@ -97,7 +97,6 @@ TEST(EsmFixedString, struct_size)
     ASSERT_EQ(4, sizeof(ESM::NAME));
     ASSERT_EQ(32, sizeof(ESM::NAME32));
     ASSERT_EQ(64, sizeof(ESM::NAME64));
-    ASSERT_EQ(256, sizeof(ESM::NAME256));
 }
 
 TEST(EsmFixedString, is_pod)
@@ -105,5 +104,4 @@ TEST(EsmFixedString, is_pod)
      ASSERT_TRUE(std::is_pod<ESM::NAME>::value);
      ASSERT_TRUE(std::is_pod<ESM::NAME32>::value);
      ASSERT_TRUE(std::is_pod<ESM::NAME64>::value);
-     ASSERT_TRUE(std::is_pod<ESM::NAME256>::value);
 }
--- a/components/esm/esmcommon.hpp
+++ b/components/esm/esmcommon.hpp
@@ -6,7 +6,6 @@
 #include <vector>
 
 #include <stdint.h>
-#include <string.h>
 
 namespace ESM
 {
@@ -57,11 +56,16 @@ public:
     }
     bool operator!=(const std::string& str) const { return !( (*this) == str ); }
 
-    size_t data_size() const { return size; }
+    static size_t data_size() { return size; }
     size_t length() const { return strnlen(self()->ro_data(), size); }
     std::string toString() const { return std::string(self()->ro_data(), this->length()); }
 
-    void assign(const std::string& value) { std::strncpy(self()->rw_data(), value.c_str(), size); }
+    void assign(const std::string& value)
+    {
+        std::strncpy(self()->rw_data(), value.c_str(), size-1);
+        self()->rw_data()[size-1] = '\0';
+    }
+
     void clear() { this->assign(""); }
 private:
     DERIVED<size> const* self() const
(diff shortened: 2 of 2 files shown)


components/esm/esmcommon.hpp
```diff
@@ -89,28 +93,41 @@ struct FIXED_STRING : public FIXED_STRING_BASE<FIXED_STRING, SIZE>
...
+    }
+    bool operator!=(const char* const str) const { return !( (*this) == str ); }
+    bool operator==(const std::string& str) const { return (*this) == str.c_str(); }
+    bool operator!=(const std::string& str) const { return !( (*this) == str ); }
 
-    char const* ro_data() const { return data; }
-    char*       rw_data() { return data; }
+    char* rw_data() { return data; }
+
+    size_t data_size() const { return size; }
```
zinnschlag, line 123: Shouldn't this function be made static? Even if static isn't needed here, it expresses the intent more clearly.

apps/openmw_test_suite/esm/test_fixed_string.cpp
```diff
@@ -1,12 +1,13 @@
...
 #include "components/esm/esmcommon.hpp"
+#include "components/esm/defs.hpp"
 
 TEST(EsmFixedString, operator__eq_ne)
 {
     {
         SCOPED_TRACE("asdc == asdc");
         ESM::NAME name;
-        name.assign("asdc");
+        name.intval = ESM::FourCC<'a','s','d','c'>::value;
```
zinnschlag, line 10: That looks a bit like a downgrade in overall code readability to me.

zinnschlag:
Are we 100% sure about the 0-termination? I think I remember that the way MW handles fixed size strings is somewhat funky. That was a long time ago though and I could remember incorrectly. Just wanna make sure that a common sense fix does not overwrite a weird but working implementation.

akortunov:
> Are we 100% sure about the 0-termination?

Here is screenshot of ID clamping from vanilla editor:
Screenshot_20200927_114147
Here I created an object with 31-byte ID (TES-CS editor does not allow to use larger IDs). Then I created a pre-defined Follow package and tried to enter as many characters as possible to the Target field. It allowed me to enter only 31 characters to the input field, and saved result as 31 byte ID + null terminator.
OpenMW-CS in this case stores 32 bytes from original string without null terminator (the behaviour of `assign` function).

When I tried to use shorter ID, I got this result:

image

Basically, it stores short ID, then null terminator, then the rest of data is garbage.

Also note that TES-CS for some reason allows to use only op to 23 bytes for actor IDs and up to 31 bytes for other characters.

I can not give the 100% without reverse-engineering, though.

components/esm/esmcommon.hpp
akortunov: Done.

Capostrophic:
> Also note that TES-CS for some reason allows to use only up to 23 bytes for actor IDs

8 bytes at the end are reserved for the number of the instance which can be seen in the console.

components/esm/esmcommon.hpp
```diff
@@ -103,14 +107,27 @@ struct FIXED_STRING<4> : public FIXED_STRING_BASE<FIXED_STRING, 4>
...
+    {
+        intval = 0;
+        size_t length = value.size();
+        if (length > 0)
+            data[0] = value[0];
+        if (length > 1)
+            data[1] = value[1];
+        if (length > 2)
+            data[2] = value[2];
+        if (length > 3)
```
Capostrophic, line 120: Doing 4 checks for strings of any length here seems a bit redundant.

akortunov: I added early-outs, but it is not a performance-critical function, and it is used just a couple of times in unit tests.

components/esm/esmcommon.hpp
```diff
@@ -103,14 +107,27 @@ struct FIXED_STRING<4> : public FIXED_STRING_BASE<FIXED_STRING, 4>
     bool operator==(uint32_t v) const { return v == intval; }
     bool operator!=(uint32_t v) const { return v != intval; }
 
+    void assign(const std::string& value)
```
elsid, line 110: This function is called nowhere but tests.

PR #3009, opened 2020-09-29
make switch to using GL hosted windows deps

psi29a:
validating via appveyor, deps are already cached on gitlab


psi29a:
AV is happy, closing this and doing further testing on GL.

PR #3010, opened 2020-10-07
Grass paging

akortunov:
Basically, I had to create a separate QuadTreeWorld with a separate ObjectPaging instance to use a separate viewing distance for groundcover objects.
Unfortunately, even in this case ObjectPaging has drawbacks:
1. QuadTreeWorld works with whole cells, so the minimum grass-covered area is 3x3 cells grid despite a circle with 8192 radius would be enough. Also frustrum culling does not work well for pages (even if a small part of page is on the screen, we need to render the whole page). It means that we render 3-4x more grass than needed (part of redundant grass is fully transparent, part is offscreen, so player does not see it anyway).
Maybe there are ways to mitigate this issue via shaders in GLSL 1.2, but I am not aware of them.
2. Even within active grid, there is a lot of groundcover objects (sometimes 10 times more than in Morrowind.esm in wilderness). It means that paging process itself requires more time, so cell transition process becomes slower. An only way to mitigate the issue is to decrease objects density, either via settings.cfg or by editing content files themselves.

Settings file:
```
[Groundcover]
enabled = true
fade start = 0.85
density = 0.5
min chunk size = 0.5
```
Note that there is a different way to detect groundcover objects. Now you should register groundcover files in openmw.cfg in the separate way:
```
groundcover=Grass Vanilla 1.esp
groundcover=Grass Vanilla 2.esp
groundcover=Grass Bloodmoon.esp
```
instead of using `content` entries.

Screenshot_20201007_143315


akortunov:
I managed to animate groundcover pages, but in a bit hackish way.

akortunov:
@psi29a, can you check if this PR works properly for you?

akortunov:
There is nothing I can do here with current QuadTreeWorld and ObjectPaging design - I need to rewrite a half of former and the whole latter to make code less messy, but it will bloat this PR a lot, so it can be done later in follow-up PR.

Note that lighting still looks different from from original engine - according to testing, MGE ignores materials for groundcover objects and uses a some separate kind of shader-based lighting for them. So the research is needed, but it is not really a blocker.

I suppose the main question so far if proposed approach is good enough, so this PR can be polished and merged.

@psi29a, @Capostrophic, @AnyOldName3, what do you think?

AnyOldName3:
I think the first sentence of that post doesn't reflect reality.

akortunov:
> I think the first sentence of that post doesn't reflect reality.

?

Sisah2:
Just curious, passing uniform with object paging mean that shader is running for the whole batch? Or still for every single grass mesh? Asking because in vec3 originalcoords break shaders on android, so if it can be passed as a uniform, then rotated added animation and rotated back to make it working for us android users. Or we should stick without paging or use texture coordinates for animation, which would need new assets and i doubt someone make them for us :(

akortunov:
> object paging mean that shader is running for the whole batch?

From what I can tell, vertex shader is running for every vertex of groundcover batch.
Uniforms work with statesets, so every vertex in batch will have the same uniform value.

akortunov:
Also vertex attributes should work on Android since OSG uses them internally:

```
setUpVertexAttribAlias(_vertexAlias,0, "gl_Vertex","osg_Vertex","vec4 ");
setUpVertexAttribAlias(_normalAlias, 2, "gl_Normal","osg_Normal","vec3 ");
setUpVertexAttribAlias(_colorAlias, 3, "gl_Color","osg_Color","vec4 ");
setUpVertexAttribAlias(_secondaryColorAlias, 4, "gl_SecondaryColor","osg_SecondaryColor","vec4 ");
setUpVertexAttribAlias(_fogCoordAlias, 5, "gl_FogCoord","osg_FogCoord","float ");
```

akortunov:
> vec3 originalcoords

It was in outdated version, it is a `float originalHeight` now.

Sisah2:
Changing in to attribute as you suggested on 4pda works, great :)

Sisah2:
There is one more android error in groundcover_vertex.glsl
if (d < 150.0 && d > 0)
Should be
if (d < 150.0 && d > 0.0)
Then it works fine.

akortunov:
> groundcover= lines seem to be eaten away by the launcher.

Are you sure? Launcher is supposed to move them to the top (since keys in launcher are sorted in the descending alphabetical order), and it actually does in my testing.

heilkitty:
Yes, it does. My mistake, sorry.

pulion:
can you fix occlussion culling with grass paging in new branch? these functions do not work together now.

akortunov:
> can you fix occlussion culling with grass paging in new branch?

No, as I said earlier, groundcover QuadTreeWorld has a separate viewing distance and has no terrain, so groundcover objects can not be occluded by terrain.

pulion:
> > can you fix occlussion culling with grass paging in new branch?
> 
> No, as I said earlier, groundcover QuadTreeWorld has a separate viewing distance and has no terrain, so groundcover objects can not be occluded by terrain.

thanks for the answer.  I thought it was not possible at the time.  but if this cannot be implemented in principle, then we will use what is already there and working.

DuendeInexistente:
Not at all a coder man and all, but wouldn't it be possible to make a
simple mesh with its visible faces/planes facing away from the grass
(Towards under the world, don't know if I'm wording it well enough) to make
it create occlusion? The face will be completely invisible unless the
engine renders both faces, but should still mostly follow the shape of the
terrain. Maybe giving it a vertex for every grass mesh loaded and
connecting them all.

akortunov:
> wouldn't it be possible to make a
simple mesh with its visible faces/planes facing away from the grass
(Towards under the world, don't know if I'm wording it well enough) to make
it create occlusion?

1. OpenMW does not have an occlusion culling at all so far.
2. Too many occluders may decrease performance instead of increasing it.
3. It is out of this PR scope.

heilkitty:
[Watch the sky.](https://youtu.be/uU9WEzYephs)

Relevant mods:

* [Aesthesia Groundcover](https://www.nexusmods.com/morrowind/mods/46377)
* Morrowind Rebirth 5.2.1
* [Grass for Morrowind Rebirth 5.2](https://www.nexusmods.com/morrowind/mods/48630)

Also, graphical replacers.

akortunov:
If anyone interested, [this](https://github.com/akortunov/openmw/tree/grass_intsancing) branch uses instancing to render groundcover objects. So far it works a bit slower than paging for me (water refraction degrades performance a bit (GPU and Draw bars) with instancing).

Instancing uses about 100MB less memory with 100% density, though, and loading is a bit faster (since CPU does not spend a time to merge objects to pages).

akortunov:
When I tried to implement bigger chunks with paging, I discovered an another limitation of Object Paging - it can not merge nearby transparent object. It means that groundcover obejcts in the 1024 units or so from player [are not merged at all](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwrender/objectpaging.cpp#L615).

pulion:
please do not drop support for the grass paging branch.  it is very necessary for android players, as it gives a chance to improve performance on android devices.

psi29a:
@pulion Then give grass instancing a go instead:
https://github.com/OpenMW/openmw/pull/3023

akortunov:
> Then give grass instancing a go instead that is the alternative

As many people already mentioned, gl4es is meh and its instancing implementations seems to be quite slow and buggy.

pulion:
> @pulion Then give grass instancing a go instead that is the alternative:
> #3023

android users have done over 100 tests with this feature.  it greatly reduces the FPS on our system.  impossible to play.

psi29a:
> > @pulion Then give grass instancing a go instead that is the alternative:
> > #3023
> 
> android users have done over 100 tests with this feature. it greatly reduces the FPS on our system. impossible to play.

For now perhaps, but development isn't static/fixed and OpenMW isn't "finished" and neither is instancing support.

AnyOldName3:
Whether instancing is ever fast on Android isn't something OpenMW has any control over unless we implement a native GLES backend and stop using GL4ES at all. If it's going to be fast, either GL4ES needs a GLES 3.0 backend (so that instancing support is guaranteed) or to detect the `GL_EXT_draw_instanced` extension with its GLES 2.0 backend. I don't actually know whether that extension is widely exposed on phones that support both GLES 2.0 and 3.0, though. If it were a desktop platform, it'd definitely be available, but Android driver writers aren't as good at exposing everything properly, so even though the extension works with GLES 2.0 in theory, they might have decided that anyone who wants to use it should target GLES 3.0 instead.

Personally, I think it was dumb of GL4ES to expose that extension when they're only emulating it. If they hadn't, OpenMW could have tested for its presence and used geometry merging instead when it's missing, but as things stand, it's present, but really slow.

tauin:
PC First Android Second imo, if something has notable benefits on PC then it should be chosen, android can be fixed later.

AnyOldName3:
> When I tried to implement bigger chunks with paging, I discovered an another limitation of Object Paging - it can not merge nearby transparent object. It means that groundcover objects in the 1024 units or so from player [are not merged at all](https://github.com/OpenMW/openmw/blob/master/apps/openmw/mwrender/objectpaging.cpp#L615).

Grass shouldn't have alpha blending on as it mandates a depth sort. MGE XE doesn't have it on either for the exact same reason. If the point of this PR is to support MGE XE grass, then we should handle it in a similar way, not start making things harder for no reason.

To clarify, alpha *testing* (also known as punch-through alpha) has things either being completely opaque or completely transparent. Because of that, you can draw things in any order you want as there's no contribution from objects behind the object unless that object is the *only* contribution, so the depth buffer can handle everything.

Alpha *blending* lets things be translucent, too, so for the maths to work out, you need to draw all the things behind a blended pixel before you draw the pixel as the colour is a blend of the current mesh and whatever's behind it. This means there needs to be a depth sort, and if you merge things, you can't change their order any more. We cheat when chunks are far away as the maths won't come out *that* wrong and we can generate different versions of a chunk to be viewed from different angles to make the approximation acceptable.

If the linked lines have any effect on grass, you're doing the grass wrong.

akortunov:
> Grass shouldn't have alpha blending on as it mandates a depth sort.

Objects paging can not merge **any** nearby object with TransparentBin, not just ones with alpha blending. That's why there is much more draw calls with paging thatn with instancing (2000 draw calls vs 1300 on the same scene for me).

AnyOldName3:
The transparent bin is only there for blended objects. If you're getting rid of blend state, you should also get rid of the transparent bin on the same stateset.

AbduSharif:
> PC First Android Second imo, if something has notable benefits on PC then it should be chosen, android can be fixed later.

I mean so far grass paging and grass instancing PRs give identical performance or just a tiny bit performance difference.

The only thing going on for grass instancing is it uses GPU so less CPU usage, less memory usage and faster cell transition.

AbduSharif:
> That's a pretty big plus for grass instancing. Grass paging gives me considerably more stuttering/loading hiccups, too. That's running with an ssd.

You actually have it bad, it's probably something with your setup as I'm on a phone and it pretty much loading and cell transition is just slow with grass paging.

Though tweaking the pre-load settings in the config file will make stuttering and slow loading issues with OpenMW almost non-existent, save for some technical problems that are already on the issue tracker.

MaladjustedPlatypus:
> 
> You actually have it bad, it's probably something with your setup as I'm on a phone and it pretty much loading and cell transition is just slow with grass paging.
> 
> Though tweaking the pre-load settings in the config file will make stuttering and slow loading issues with OpenMW almost non-existent, save for some technical problems that are already on the issue tracker.

Not a problem with the setup, just that I run much higher settings than a phone would. Everything else is kept the same between testing from instancing to paging. Not having either one of the branches and no grass = no stutter. Grass instancing with grass = no stutter. Grass paging with grass = stutter.
Plus as mentioned earlier higher poly assets benefit more from the instancing so anyone that uses mesh replacers or more detailed custom assets will benefit from the instancing more. Surprised mobile folks can even see the grass on those screens tbh.

AbduSharif:
> Not a problem with the setup, just that I run much higher settings than a phone would. Everything else is kept the same between testing from instancing to paging. Not having either one of the branches and no grass = no stutter. Grass instancing with grass = no stutter. Grass paging with grass = stutter.

Stutter doesn't exist for everyone with grass paging (like me, save for the technical problems on the tracker), hardware and setup do have a hand in stutter issues, merely adjusting my settings helped me solve reloading issues and stutter issues, from what you're saying you might be setting the options in an inefficient way, if you don't want to take my advice feel free though.

> Plus as mentioned earlier higher poly assets benefit more from the instancing so anyone that uses mesh replacers or more detailed custom assets will benefit from the instancing more. 

I get that but that's not the topic we were on, also you might need to read it carefully: "I made an additional investigation about instancing performance. Different sources state that paging is generally faster than instancing with low-poly models (as Morrowind ones), while instancing should be faster with complex meshes instead (several thousands of triangles per shape) - you need to load a lot of redundant data with paging in such case."

And as I said, wether that investigation is right or not, grass instancing still gives identical performance to paging or a tiny bit difference between the two on PC (from the tests I saw), but I imagine the devs still have other performance tricks that they're going to try.
Let's not get it mixed too, I fully support having either of the grass PRs, they both give me acceptable performance on my device, just that paging is the most performant for me, and I won't just gloss over it.

> I run much higher settings than a phone would. 
> Surprised mobile folks can even see the grass on those screens tbh.

You clearly never tried OpenMW on a flagship or a mid-range smartphone.

MaladjustedPlatypus:
> 
> Stutter doesn't exist for everyone with grass paging (like me, save for the technical problems on the tracker), hardware and setup do have a hand in stutter issues, merely adjusting my settings helped me solve reloading issues and stutter issues, from what you're saying you might be setting the options in an inefficient way, if you don't want to take my advice feel free though.

If I need to lower the settings to get the same performance with paging, then clearly paging is not as efficient, at least with my system. It's common sense. For the same equivalent graphics quality it does not perform as well. That's the whole point of testing, to find this stuff out because it's not set in stone that the two systems have the same performance hit on all systems.
> 
> I get that but that's not the topic we were on, also you might need to read it carefully: "I made an additional investigation about instancing performance. Different sources state that paging is generally faster than instancing with low-poly models (as Morrowind ones), while instancing should be faster with complex meshes instead (several thousands of triangles per shape) - you need to load a lot of redundant data with paging in such case."
> 
> And as I said, wether that investigation is right or not, grass instancing still gives identical performance to paging or a tiny bit difference between the two on PC (from the tests I saw), but I imagine the devs still have other performance tricks that they're going to try.
> Let's not get it mixed too, I fully support having either of the grass PRs (or both), they both give me acceptable performance on my device, just that paging is the most performant for me, and I won't just gloss over it.

Again, the idea that the systems have identical performance on PCs is clearly false since testing still has not been widely performed on a variety of different system configurations and some people like myself are having clearly different results. If the difference is evident now, it'll definitely be if people try more detailed assets. I'm not opposed to having a choice of system, but like yourself I would not gloss over the potential dropping of instancing.
> 
> You clearly never tried OpenMW on a flagship or a mid-range smartphone.

Didn't mean that in regards to whether or not they could render the grass, but rather a comment on the physical size of the displays.

pulion:
if this is interesting to someone, then I have two smartphones: one with SD 855 and SD 865. Having carried out a large number of tests on them with different settings, I can say with confidence that with the grass paging parameter the FPS is 5-10 units higher, depending on  from the terrain.

pulion:
for smartphones grass paging is better than Grass instancing

Sisah2:
Does this still receiving some love? :D

pulion:
> Does this still receiving some love? :D

I really hope so.

akortunov:
Superseded by [MR#441](https://gitlab.com/OpenMW/openmw/-/merge_requests/441) and #3023.

PR #3011, opened 2020-10-08
Navmesh fixes

elsid:
Primarily change the implementation of navmesh key creation. Using memcpy directly is ~13 times faster. And it's not possible to use `std::string::data` because there is no non-const overload returning `char*` until c++17. So `std::vector<unsigned char>` is used instead.


psi29a:
@elsid please bump to C++17 and see if AppVeyor and Travis give you green.

psi29a:
Thanks @elsid

PR #3012, opened 2020-10-14
Fix crash on saving

akortunov:
This line leads to crash when customData is null:

```
const ContainerCustomData& customData = ptr.getRefData().getCustomData()->asContainerCustomData();
```

We should handle this case, as we did it earlier.

Note that I can not tell if empty customData is intended or it is a bug.


psi29a:
@Assumeru have a look please?

Assumeru:
Empty custom data is fairly normal; someone added it as an optimization I'm pretty sure. It's also used to respawn organic containers.

Looks like I combined the checks at some point and then forgot to change them back after it no longer made sense.

PR #3013, opened 2020-10-15
Resolve 'shared_timed_mutex' is unavailable: introduced in macOS 10.12

psi29a:
Resolve this https://travis-ci.org/github/OpenMW/openmw/jobs/735972857


psi29a:
Issue seems to be resolved now: https://travis-ci.org/github/OpenMW/openmw/jobs/735980892
The failure is now that we busted the ccache again and reached our timeout.

As with much in the name of progress, with this PR, our OSX Target is now set to 10.12 and not 10.9; the practical side is that we no longer support MacOS 10.9, 10.11 and 10.11

PR #3014, opened 2020-10-15
Maybe resolve regression with MSVC2017

AnyOldName3:
Hopefully this fixes the problem caused by the async physics merge. I've not tested it, though. It's a GitHub PR rather than GitLab MR because GitLab CI doesn't do MSVC 2017, only AppVeyor.


psi29a:
This is pretty ugly, any way we can set something once, project wide?

fr3dz10:
Maybe put it in appveyor config only? Nightlies and release are built from gitlab runners I think so they're not concerned

psi29a:
GL apparently only has MSVC2019, so this was never a problem for them.

Our policy, so far, has been to support the last 2 major releases of MSVC so putting it in AppVeyor isn't an option.

AnyOldName3:
It's not only AppVeyor that uses MSVC2017. Because I don't want to spend a day rebuilding my self-built dependencies in all the build configurations, my everything-is-debuggable local build is still on 2017, for example, so it definitely needs to be in our build system rather than just a subset of the scripts that launch it.

There *are* ways to specify this for everything at once, but they're more error-prone, as they replace what's there rather than append, need setting up before any targets are created (as they only initialise the defaults, and don't set things on existing targets), and are generally pre-CMake 3.0 and messy.

If I knew which part of OpenMW actually needed the definition, I could keep it more contained, for example if it was in `components`, setting it as a target definition as `INTERFACE` is actually already setting it for `components` and everything that depends on `components`, which would solve all the immediate problems, but might need expanding on if/when the feature is used somewhere else. Alternatively, we could just specify the define in the source code where it's needed.

fr3dz10:
Fair enough. When I googled that error this afternoon I found a report linked to smart pointer construction. If I had to guess I would say that it is physicssystem.cpp

AnyOldName3:
The log says `compiling source file C:\projects\openmw\apps\openmw\mwphysics\physicssystem.cpp` in the error message, so if we're okay with not affecting the whole project, we can add it to `openmw` neatly or `components` equaly neatly and have more, but not total, coverage.

psi29a:
Yeah, I'm okay with turning that switch on openmw and leaving it at that. MSVC2017 will yell at us if we have to do it elsewhere. We can then remove it when we drop MSVC2017.

AnyOldName3:
`openmw` rather than `components` and (implicitly) everything that depends on it?

psi29a:
As far as I'm aware, the only problem was related to anything that touched
`C:\projects\openmw\apps\openmw\mwphysics\physicssystem.cpp`

AnyOldName3:
I originally applied it to everything because we might come to rely on it somewhere else in the future. The definition makes something in MSVC2017 that wasn't historically standards-compliant become standards-compliant, so it's not a bad thing to have enabled.

psi29a:
But doesn't this setting get ignored in MSVC2019 anyway?

AnyOldName3:
Yes, MSVC2019 never had the bug, so doesn't need to maintain backwards compatibility by optionally still having the bug.

When you say you're fine with how it is now, do you mean you're ready to press the merge button despite the existing mess?

psi29a:
I'm fine with your choice, if you think it is better to be explicit over implicit. I'll defer to you. We can revert this when the next MSVC comes out. ;)

AnyOldName3:
I can't force-push in the web editor, so to keep the history clean, I've superseded this with #3016

PR #3015, opened 2020-10-15
 Mark overrided methods as overrides

akortunov:
Hopefully it will silence "overrided method is not marked as override" warnings spam with Clang.


psi29a:
Apparently there are more?

```
In file included from /Users/travis/build/OpenMW/openmw/apps/openmw/mwgui/windowmanagerimp.cpp:86:
4909/Users/travis/build/OpenMW/openmw/apps/openmw/mwgui/spellbuyingwindow.hpp:36:18: warning: 'onResChange' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
4910            void onResChange(int, int) { center(); }
4911                 ^
```

akortunov:
Ok, try an another approach. I suppose that Clang expect us to mark *every* override as `override`.
So try to disable that warning at all since most of our codebase does not use `override` keyword so far.

AnyOldName3:
Has anyone been going through the code getting rid of `override`? We've definitely merged pull requests that purportedly exhaustively fix missing-`override` warnings before, including one I made years ago my going over a MacOS Travis log.

I'm much more keen on sorting out missing `override`s than disabling the warning and letting implicit `virtual` and implicit `override` through, as they make it significantly less clear what's going on with a piece of code.

AnyOldName3:
I think we've got lots of virtual functions that don't strictly need to be virtual, and I'm not convinced everyone's been removing `virtual` when they've added `override`, so I don't think the figure is *remotely* close to 2500.

AnyOldName3:
I've found 99 instances of things being declared both `virtual` and `override`, 427 instances of `override` on its own, and 865 pure `virtual`s (`=( ?)0`), so we *might* have quite a lot of things that should be marked `override`. I guess the only way to know for sure would be to do a clean build without CCache with the warning on and see how many unique instances of the warning it spits out.

psi29a:
I'm against of disabling the warning, at least in the last two companies I've been with (that spans more than a decade), it was proper practice to follow the rules of virtual, override and final.

akortunov:
>  I guess the only way to know for sure would be to do a clean build without CCache

Compilers do not log "unique" instances, they log every usage. For example, clean GGC build with -Wsuggest-override showed 201340 warnings to me.

AnyOldName3:
That's why I explicitly specified *unique*. It's not too hard to pipe the output of the compiler to a tool that keeps track of unique lines.

fr3dz10:
Clang-tidy can automatically do the change for all instances, no need to do it by hand.

AnyOldName3:
Could you run it locally with just that setting active and open an MR so we can see what we're dealing with? I won't do it myself as it'd basically mandate a full rebuild when switching to and from that branch and there's no CCache for MSVC.

akortunov:
Rougly 3300 lines total. It seems that my original estimation (2500+ lines) was more-or-less correct. Let see if it will be enough...

apps/opencs/view/doc/runlogsubview.hpp
```diff
@@ -13,7 +13,7 @@ namespace CSVDoc
 
             RunLogSubView (const CSMWorld::UniversalId& id, CSMDoc::Document& document);
 
-            virtual void setEditLock (bool locked);
+            virtual void setEditLock (bool locked) override;
```
psi29a, line 16: Shouldn't it be that in this particular case, we drop virtual as well?   The virtual void setEditLock is inherited from SubView, s here, it would be:  void setEditLock (bool locked) override;

extern/recastnavigation/DebugUtils/Include/DebugDraw.h
```diff
@@ -206,11 +206,11 @@ class duDisplayList : public duDebugDraw
...
-	virtual void depthMask(bool state);
-	virtual void begin(duDebugDrawPrimitives prim, float size = 1.0f);
-	virtual void vertex(const float x, const float y, const float z, unsigned int color);
-	virtual void vertex(const float* pos, unsigned int color);
-	virtual void end();
+    virtual void depthMask(bool state) override;
+    virtual void begin(duDebugDrawPrimitives prim, float size = 1.0f) override;
+    virtual void vertex(const float x, const float y, const float z, unsigned int color) override;
+    virtual void vertex(const float* pos, unsigned int color) override;
+    virtual void end() override;
```
psi29a, line 213: Since you set recastnavigation to use the `no-suggest-override` flag, then we can revert these right?

AnyOldName3:
I'm still seeing redundant `virtual`s for a lot of the new `override`s.

akortunov:
> I'm still seeing redundant virtuals for a lot of the new overrides.

Can you specify where exactly you see them and why you think that it is an issue?

apps/opencs/model/doc/savingstages.hpp
```diff
@@ -37,10 +37,10 @@ namespace CSMDoc
             OpenSaveStage (Document& document, SavingState& state, bool projectFile);
             ///< \param projectFile Saving the project file instead of the content file.
 
-            virtual int setup();
+            virtual int setup() override;
```
AnyOldName3, line 40: Here's one.

As `override` implies `virtual`, the redundant `virtual` is pointlessly verbose. In general, it's best for code to only say what it needs to.

akortunov: So there is no real difference?

akortunov: Anyway, I managed to make clang-tidy to work properly, so there is even more changes now.

Capostrophic:
I'm somewhat disappointed that every single read/post NIF record method has to be marked `override` even though it's not going to make the development any easier. I guess it's not possible to avoid it since it's a good practice in general case?

psi29a:
This looks good to me, thanks @akortunov

psi29a:
> I'm somewhat disappointed that every single read/post NIF record method has to be marked `override` even though it's not going to make the development any easier. I guess it's not possible to avoid it since it's a good practice in general case?

It's a good habit to get into, especially when working with inheritance and pimpl idiom.

fr3dz10:
./components/myguiplatform/myguirendermanager.hpp:112:10: warning: 'setViewSize' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    void setViewSize(int width, int height);
         ^
/usr/local/include/MYGUI/MyGUI_RenderManager.h:50:16: note: overridden virtual function is here
                virtual void setViewSize(int _width, int _height) = 0;

PR #3018, opened 2020-10-17
Use emplace_back instead of push_back

akortunov:
"Emplace" is a modern way to add values to vectors, which is supposed to work faster (because it can construct objects inside vector rather than pass them as an argument).

Amount of changes is not that large, so I suggest to convert code at once and then use `emplace_back` in the new code (we actually already do it now).

### DIFF
--- a/apps/mwiniimporter/importer.cpp
+++ b/apps/mwiniimporter/importer.cpp
@@ -645,7 +645,7 @@ MwIniImporter::MwIniImporter()
     }
 
     for(int i=0; fallback[i]; i++) {
-        mMergeFallback.push_back(fallback[i]);
+        mMergeFallback.emplace_back(fallback[i]);
     }
 }
 
@@ -910,7 +910,7 @@ void MwIniImporter::importGameFiles(multistrmap &cfg, const multistrmap &ini, co
                     std::time_t time = lastWriteTime(path, defaultTime);
                     if (time != defaultTime)
                     {
-                        contentFiles.push_back({time, path});
+                        contentFiles.emplace_back(time, std::move(path));
                         found = true;
                         break;
                     }
--- a/apps/opencs/editor.cpp
+++ b/apps/opencs/editor.cpp
@@ -229,7 +229,7 @@ void CS::Editor::openFiles (const boost::filesystem::path &savePath, const std::
     if(discoveredFiles.empty())
     {
         for (const QString &path : mFileDialog.selectedFilePaths())
-            files.push_back(path.toUtf8().constData());
+            files.emplace_back(path.toUtf8().constData());
     }
     else
     {
@@ -246,7 +246,7 @@ void CS::Editor::createNewFile (const boost::filesystem::path &savePath)
     std::vector<boost::filesystem::path> files;
 
     for (const QString &path : mFileDialog.selectedFilePaths()) {
-        files.push_back(path.toUtf8().constData());
+        files.emplace_back(path.toUtf8().constData());
     }
 
     files.push_back (savePath);
--- a/apps/opencs/model/doc/loader.cpp
+++ b/apps/opencs/model/doc/loader.cpp
@@ -127,7 +127,7 @@ void CSMDoc::Loader::load()
 
 void CSMDoc::Loader::loadDocument (CSMDoc::Document *document)
 {
-    mDocuments.push_back (std::make_pair (document, Stage()));
+    mDocuments.emplace_back (document, Stage());
 }
 
 void CSMDoc::Loader::abortLoading (CSMDoc::Document *document)
--- a/apps/opencs/model/doc/operation.cpp
+++ b/apps/opencs/model/doc/operation.cpp
@@ -57,7 +57,7 @@ void CSMDoc::Operation::run()
 
 void CSMDoc::Operation::appendStage (Stage *stage)
 {
-    mStages.push_back (std::make_pair (stage, 0));
+    mStages.emplace_back (stage, 0);
 }
 
 void CSMDoc::Operation::setDefaultSeverity (Message::Severity severity)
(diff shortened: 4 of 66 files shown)


psi29a:
Woah there cowboy, I recently seen in a @Capostrophic MR that he replaced a emplace_back with a push_back because of a subtle bug in the way that things work.

Capostrophic:
Most of these changes look fine. I experienced problems when I tried to emplace_back to a vector of structs using struct elements as arguments. Then again it just didn't build at all when I tried to do it (I guess this PR builds fine). And it just wasn't meaningful to pass a constructed struct to emplace_back.

akortunov:
>  Then again it just didn't build at all when I tried to do it (I guess this PR builds fine).

From what I can tell, a struct need a constructor with given number of arguments. In cases when there is no such constructor, structs can not be emplaced. Note that such cases (as ones from [this](https://gitlab.com/OpenMW/openmw/-/merge_requests/327) PR are not affected by this PR (clang-tidy ignores them). Also there is a possibility of resource leaks when allocation objects using `new` keywork without smart pointers, but I do not see such cases.

apps/mwiniimporter/importer.cpp
```diff
@@ -910,7 +910,7 @@ void MwIniImporter::importGameFiles(multistrmap &cfg, const multistrmap &ini, co
                     std::time_t time = lastWriteTime(path, defaultTime);
                     if (time != defaultTime)
                     {
-                        contentFiles.push_back({time, path});
+                        contentFiles.emplace_back(time, path);
```
elsid, line 913:
```suggestion
                        contentFiles.emplace_back(time, std::move(path));
```
Otherwise `path` is still copied because without explicit casting to rvalue reference it's passed by lvalue reference and copy constructor for `boost::filesystem::path` is called. And then a destructor for a local variable. The fact that object can be constructed inside vector storage probably is already used by compilers for push_back case because I couldn't get any significant performance difference for just replacing push_back by emplace_back: https://quick-bench.com/q/lUMW7TZktQLH978fpYoQ5BjvzKM . But with `std::move` it's obviously better.

elsid (changes requested):
I think this change has to be more specialized for each use of `push_back`. That article is quite good on describing things.

apps/mwiniimporter/importer.cpp
akortunov: I suppose that I need to modify all other places where existing string variables are passed (if there are such other places), right?

psi29a: Isn't path an boost::filesystem::path object and not a string?
https://theboostcpplibraries.com/boost.filesystem-paths

elsid: Any object that requires allocation. It's not just a string, but for a sake of simplicity I used `std::string` (quick-bench.com doesn't support boost). `boost::filesystem::path` has move constructor and it owns strings so the same rule can be applied.

akortunov: > I couldn't get any significant performance difference for just replacing push_back by emplace_back

Even on your chart there is a 6% boost, and we do not use `std::move` widely on our codebase.

Also personaly I do not like the `std::move` syntax.
As for me, the 
```
results.recordMapping.emplace_back(id, newId);
```
is more readable than
```
results.recordMapping.push_back(std::make_pair(id, newId));
```
but the 
```
#include <utility>
results.recordMapping.emplace_back(std::move(id), std::move(newId));
```
does not make code to look better at all.

symbiogenesis: For readability, could a wrapper function be created for the cases where std::move() needs to be used inside of emplace_back()

elsid: In case of `push_back` there is a call of copy constructor for path, constructor for pair and then move constructor for pair that includes calling move constructor for path. For `emplace_back` it's copy constructor for path and constructor for pair. So the difference is move constructor for pair is not called and move constructor for path. That can explain these 6%. But copy constructor call for path is the most expensive operation here.

`std::move` is a standard tool to call move constructors with lvalue. There is no other way to do it rather than implementing it yourself. When you use it other developers understand what's going on. If you try to do some tricks expect questions.

Also in this case you can do:
```c++
std::time_t time = lastWriteTime(path, defaultTime);
if (time != defaultTime)
{
    contentFiles.emplace_back(time, dataPath / *entry);
    found = true;
    break;
}
```

`dataPath / *entry` expression now returns rvalue so move constructor is called by default and there is no need in keeping `path` variable.

https://quick-bench.com/q/3uH25AnXvfKyl_Roc9HoPEkiL0o

elsid: For the context of:
```c++
results.recordMapping.emplace_back(std::move(id), std::move(newId));
```

```c++
for (const std::string& id : ) {
    std::string newId = ...;
    results.recordMapping.emplace_back(std::move(id), std::move(newId));
    reverseLookupMap.emplace(texture, newId);
}
```
Not sure did you mean it or something else. `std::move(id)` doesn't make any sense, const modifier will not go away, the result will be of type `const std::string&&` which is implicitly converted into `const std::string&` and copy constructor will be called. `std::move(newId)` will leave an empty string in `newId` which is not something you want in this case.

akortunov: > dataPath / *entry expression now returns rvalue so move constructor is called by default and there is no need in keeping path variable.

`path` variable is used in the `lastWriteTime` method, so your code does not compile.

akortunov: Using `std::move(path)`, and I do not see a point to modify other places. It can be done via follow-up PR, if needed.

> For the context of

It is just an example of syntax.

PR #3020, opened 2020-10-19
Strip the plugin index when looking for deleted refs.

unelsson:
This fixes https://gitlab.com/OpenMW/openmw/-/issues/832 . There's however more to it (see the comments at gitlab), it may be better to somehow fix these are component level, but it's not straightforward because of the different logics of OpenMW and OpenMW-CS.

What it does, it strips the first byte of mIndex, which is the plugin index, and leaves only the reference index. Then the reference index will match the references of the deleted objects.

### DIFF
--- a/apps/opencs/model/world/refcollection.cpp
+++ b/apps/opencs/model/world/refcollection.cpp
@@ -64,9 +64,10 @@ void CSMWorld::RefCollection::load (ESM::ESMReader& reader, int cellIndex, bool 
 
         // ignore content file number
         std::map<ESM::RefNum, std::string>::iterator iter = cache.begin();
+        unsigned int thisIndex = ref.mRefNum.mIndex & 0x00ffffff;
         for (; iter != cache.end(); ++iter)
         {
-            if (ref.mRefNum.mIndex == iter->first.mIndex)
+            if (thisIndex == iter->first.mIndex)
                 break;
         }
 
--- a/apps/opencs/model/world/refcollection.cpp
+++ b/apps/opencs/model/world/refcollection.cpp
@@ -64,10 +64,10 @@ void CSMWorld::RefCollection::load (ESM::ESMReader& reader, int cellIndex, bool 
 
         // ignore content file number
         std::map<ESM::RefNum, std::string>::iterator iter = cache.begin();
-        unsigned int thisIndex = ref.mRefNum.mIndex & 0x00ffffff;
+        ref.mRefNum.mIndex = ref.mRefNum.mIndex & 0x00ffffff;
         for (; iter != cache.end(); ++iter)
         {
-            if (thisIndex == iter->first.mIndex)
+            if (ref.mRefNum.mIndex == iter->first.mIndex)
                 break;
         }
 


zinnschlag (commented):
Working mostly from memory, which is slightly foggy at this point. I don't think I can offer a lot of help here (the whole RefNum handling in Morrowind is very confusing and I am not sure if I ever understood it in every detail).

The comment at the top of the code block (ignore content file number) seems to indicate that we indeed should ignore the upper 8 bits, since that is how the content file number is stored in the original file format (we are using a different format internally and at least for saved game files, which lifts the 254 plugin limitation).
However the comment could also mean the we are just to ignore the additional field for the new format.
If it works, maybe roll with it, do so more testing and hope for the best? Sorry, that is all I got.

btw. the shared_ptr used in storing readers was originally added to avoid invalidating old readers (or references to old readers?) when gradually building up the reader list over multiple function calls. I am not sure if that is still needed. Anyway, this part is multi-threaded. So treat carefully!

unelsson:
This one is almost the same, but better:
https://github.com/cc9cii/openmw/commit/0de223c637d0820a97417d84796c1bfeb1836fb8

psi29a:
If needed, cherry-pick his and re-work as needed?

psi29a:
> This one is almost the same, but better:
> [cc9cii@0de223c](https://github.com/cc9cii/openmw/commit/0de223c637d0820a97417d84796c1bfeb1836fb8)

Do you mean that cc9cii is better or that yours is better?

What should/can be done with PR or is it still WIP?

unelsson:
> > This one is almost the same, but better:
> > [cc9cii@0de223c](https://github.com/cc9cii/openmw/commit/0de223c637d0820a97417d84796c1bfeb1836fb8)
> 
> Do you mean that cc9cii is better or that yours is better?
> 
> What should/can be done with PR or is it still WIP?

I think that cc9cii's is better, because:
a) It includes some testing for loaded files, I don't really know what the checks for ESM::Header::MasterData do, and I haven't had the time to check. Similar things are done in the engine, but it might not be crucial for OpenMW-CS.
b) `cache.insert(std::make_pair(refNum, ref.mId);` in that probably inserts the correct plugin ID into memory.. I really should check if that is necessary, or if my version is enough. My version might be able to connect the deletion into proper refId, but still cache wrong refId.

Anyway, I couldn't do any proper testing earlier, because OpenMW-CS kept (and still keeps) crashing on Manjaro/XFCE4. I tracked that bug down to QStatusBar malfunction https://forum.manjaro.org/t/openmw-cs-stopped-working-qt5-15-xfce4-problem/33053/3 , which is probably related to XFCE4 somehow, as the same thing works properly on Gnome. Now that this is found, I can probably test this better.

unelsson:
I tested this by creating a mod, then re-opening that mod, making a few more changes, then saving and checking. It works like it should in my testing. Fixes the main problem with https://gitlab.com/OpenMW/openmw/-/issues/832 , that deleted or moved instances wouldn't be loaded to OpenMW-CS from plugins. I didn't test how MVRF works though, I understood it's about moving things to different cells, but it's only logical that those would suffer from this same refid-related issue, and therefore would be fixed with this PR.

The idea is just to force the first byte of refID to zero, that's where the plugin index is carried, but as that is not needed in OpenMW, isn't it better just to zero it permanently. ref.mRefNum.mIndex is of format 0xXXRRRRRR, where XX is plugin index, RRRRRR is refid.

The other part that I mentioned earlier: master file checking in cc9cii's commit, I wouldn't touch that part with this PR. It's another issue (if there even is an issue).

If it's up to me, it can be merged. I but criticism and further reviewing and testing is always welcome.

psi29a:
I'll wait for AV to finish. We should eventually migrate over to GL and make MRs there, that way we have coverage across 4 platforms. :)

AnyOldName3:
GitLab actually supports running their CI on GitHub MRs, which might be really useful for us if not for the general attempt to migrate totally to GitLab.

PR #3021, opened 2020-10-22
Do not use deprecated Qt functions

akortunov:
Summary of changes:
1. `Qt::SystemLocaleShortDate` and `Qt::SystemLocaleLongDate` are deprecated, `QLocale::system().toString()` should be used instead.
2. `XInitThreads()` is used only in Qt4, so `Qt::AA_X11InitThreads` has no effect in Qt5. Remove this line since it does nothing.

### DIFF
--- a/apps/opencs/main.cpp
+++ b/apps/opencs/main.cpp
@@ -47,9 +47,6 @@ int runApplication(int argc, char *argv[])
     setenv("OSG_GL_TEXTURE_STORAGE", "OFF", 0);
 #endif
 
-    // To allow background thread drawing in OSG
-    QApplication::setAttribute(Qt::AA_X11InitThreads, true);
-
     Q_INIT_RESOURCE (resources);
 
     qRegisterMetaType<std::string> ("std::string");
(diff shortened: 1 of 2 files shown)


psi29a:
That's great, I noticed those recently too. Thanks.

apps/opencs/main.cpp
```diff
@@ -47,9 +47,6 @@ int runApplication(int argc, char *argv[])
     setenv("OSG_GL_TEXTURE_STORAGE", "OFF", 0);
 #endif
 
-    // To allow background thread drawing in OSG
-    QApplication::setAttribute(Qt::AA_X11InitThreads, true);
```
psi29a, line 51: Was this set to handle threading?

Won't we need this back when we fix or Qt5 issues with threading?

akortunov: > Was this set to handle the threaded renderer?

Yes, with Qt4. As I said, Qt5 does not use `XInitThreads()`, so this line just is no-op with Qt5.

psi29a: Great, thanks.

PR #3022, opened 2020-10-25
Allow to assign custom shaders to nodes

akortunov:
Requested by @Capostrophic.

Just allow to assign a custom shader to node instead of just hardcoded "objects" one, something like this:

```
recreateShaders(node, "whatever");
```

to use the whatever_vertex.glsl and whatever_fragment.glsl.

By default "objects" shader is still used.

### DIFF
--- a/components/resource/scenemanager.cpp
+++ b/components/resource/scenemanager.cpp
@@ -242,9 +242,9 @@ namespace Resource
         return mForceShaders;
     }
 
-    void SceneManager::recreateShaders(osg::ref_ptr<osg::Node> node)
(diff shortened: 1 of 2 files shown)


unelsson:
Should this be connected to config (user settings)? Why is this needed?

Capostrophic:
No, this doesn't concern end users at all. As you can guess it allows *coders* to assign non-generic object shaders from anywhere in the code. Only akortunov's grass PRs rely on this functionality at the moment - to be able to animate grass - but this is also useful for NIF file handling.

PR #3024, opened 2020-10-28
Fix some issues, found by CoverityScan

akortunov:
Should fix 4 new defects, also may fix up to three suppressed defects.

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -590,7 +590,9 @@ namespace MWMechanics
 
         if (!actorState.isTurningToPlayer())
         {
-            float angle = std::atan2(dir.x(), dir.y());
+            float from = dir.x();
+            float to = dir.y();
+            float angle = std::atan2(from, to);
             actorState.setAngleToPlayer(angle);
             float deltaAngle = Misc::normalizeAngle(angle - actor.getRefData().getPosition().rot[2]);
             if (!mSmoothMovement || std::abs(deltaAngle) > osg::DegreesToRadians(60.f))
@@ -1723,7 +1725,7 @@ namespace MWMechanics
                     shouldAvoidCollision = true;
                 else if (package->getTypeId() == AiPackageTypeId::Wander && giveWayWhenIdle)
                 {
-                    if (!dynamic_cast<const AiWander*>(package.get())->isStationary())
+                    if (!static_cast<const AiWander*>(package.get())->isStationary())
                         shouldAvoidCollision = true;
                 }
                 else if (package->getTypeId() == AiPackageTypeId::Combat || package->getTypeId() == AiPackageTypeId::Pursue)
--- a/apps/openmw/mwphysics/mtphysics.cpp
+++ b/apps/openmw/mwphysics/mtphysics.cpp
@@ -142,6 +142,7 @@ namespace MWPhysics
 {
     PhysicsTaskScheduler::PhysicsTaskScheduler(float physicsDt, std::shared_ptr<btCollisionWorld> collisionWorld)
           : mPhysicsDt(physicsDt)
+          , mTimeAccum(0.f)
           , mCollisionWorld(std::move(collisionWorld))
           , mNumJobs(0)
           , mRemainingSteps(0)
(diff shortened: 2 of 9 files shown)


apps/openmw/mwworld/containerstore.cpp
```diff
@@ -79,13 +79,20 @@ namespace
...
+                ptr.getRefData().setCount(0);
+            mStore.fillNonRandom(mStore.mPtr.get<ESM::Container>()->mBase->mInventory, "", mStore.mSeed);
+            addScripts(mStore, mStore.mPtr.mCell);
+            mStore.mResolved = false;
+        }
+    }
+    catch(const std::runtime_error& e)
+    {
+        Log(Debug::Error) << "Error in the ResolutionListener destructor: " << e.what();
     }
```
psi29a, line 96: Are we sure this is necessary?

It seems that the reason anything is 'thrown' is because of this:
apps/openmw/mwworld/containerstore.cpp:
```
    if (ptr.isEmpty())
        throw std::runtime_error ("invalid iterator");
```

Since you are just 'catching' it and logging the error... why not just log the error there instead of throwing it. Then is this try/catch block not necessary.

akortunov: > why not just log the error there instead of throwing it

I do not know much about low-level code, so I can not tell if it is a valid behaviour in any case where iteration over container store is used.

psi29a: It was written by @zinnschlag in 2012, probably with other ideas.  I believe it would be better to just log an Error... or perhaps a warning, "invalid iterator" if the ptr is empty and not store it in the container. That way we know something is wrong and an go looking... that being said, in the years since 2012... we've never seen that runtime_error pop up.

akortunov: BTW, it is an operator*(), which is used for such cases:

```
MWWorld::Ptr weaponPtr = *weapon;
```

For some reason we treat empty pointer as an invalid one here.

akortunov: > we've never seen that runtime_error pop up.

That's because we use early checks: 

```
    if (weapon == inv.end())
        return;

    MWWorld::Ptr weaponPtr = *weapon;
```

The point is that iterator can point to nothing rather than on item, so in this case an attempt to de-reference such iterator causes an exception.

akortunov: If we need a discussion first, I can just drop this change from this PR.

psi29a: So that last check isn't necessary in your opinion? Should we still not er on the side of caution and replace the throw with a error log entry rather than just removing the check?

akortunov: > So that last check isn't necessary in your opinion?

I do not know, this function is used in too many places. Personally I'd prefer to do not touch it.

psi29a: It's hasn't thrown "yet" in so many years and you just end up catching it and then logging an error anyway. What's the difference?

akortunov: > What's the difference?

An only difference is that I do not touch hundreds places in game mechanics, which use this function.

psi29a: I'm not asking to touch hundreds of places though. No other places where op container* operator catches the exception either.

I think it is just silly to have a try/catch in a destructor for something that has shown to not throw an exception. The only thing you do with the exception is to then ignore it anyway and throw up a log message. The only reason you're motivated to 'fix' the issue is because coverity claims it to be an issue.

If you want me to make the change, that's fine too. Revert the try/catch and I'll merge this PR in.

psi29a:
Logger error in MSVC, for the rest, it looks good. Thanks!

apps/openmw/mwworld/containerstore.cpp
psi29a: > Dereference happens when you get a reference to Ptr from iterator.

I've come back around on this after looking at the code, thanks for pointing it out @akortunov

PR #3026, opened 2020-11-04
Rework cure effects

akortunov:
A result of additional invistigation, requested by @Capostrophic:

1. Cure effects should remove related effects from all sources, not just from active spells.
Note that things such Cure Poison and Cure Paralyzation do not remove spells themselves, just affected effects.
2. Cure Corprus effect in Morrowind removes whole spells, which contain corprus effects.
3. Cure effects are not instant nor tickable ones. They are rather togglable effects - they either work or do not work (as bound effects, water walking, etc.)

### DIFF
--- a/apps/openmw/mwmechanics/actors.cpp
+++ b/apps/openmw/mwmechanics/actors.cpp
@@ -916,11 +916,53 @@ namespace MWMechanics
             }
     };
 
+    void Actors::applyCureEffects(const MWWorld::Ptr& actor)
+    {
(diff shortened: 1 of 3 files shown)


psi29a:
Is this just re-work of things already 'closed' in 0.47 changelog?

akortunov:
> Is this just re-work of things already 'closed' in 0.47 changelog?

Technically, it is just a better replacement for #3025 and addition for #2544, both already have changelog entries.

PR #3028, opened 2020-11-13
Rework file error messages handling

akortunov:
A follow-up PR for #851.

Summary of changes:
1. CSTDIO-based implementation now handles error codes as well.
2. Error message syntax was unified between POSIX and CSTDIO again.
3. Added missing whitespaces between words in some error messages.
4. We do not use capitalized names for local variables. Use lowercase names instead.

Hopefully this PR will help to debug [issue #5682](https://gitlab.com/OpenMW/openmw/-/issues/5682).

### DIFF
--- a/components/files/lowlevelfile.cpp
+++ b/components/files/lowlevelfile.cpp
@@ -4,6 +4,11 @@
 #include <sstream>
 #include <cassert>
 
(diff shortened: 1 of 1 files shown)


psi29a:
lgtm; thanks.

PR #3029, opened 2020-11-13
Replace zeroes and nulls by nullptrs

akortunov:
Note that I intentionally skipped "external" projects, C files and files which are modified versions of OSG ones (e.g. scene optimizer and shadow technique).


psi29a:
Probably a good idea to start doing Gitlab MRs so that there is more coverage.

akortunov:
Any suggestions?

psi29a:
lgtm

PR #3031, opened 2020-11-26
[WIP] OpenMW-CS rendering to Qt5

unelsson:
Migrates to Qt5 and drops support of Qt4. The idea is to use QOpenGLWidget instead of osgQt::GraphicsWindowQt, and use something alike osgQOpenGLWidget and OSGRenderer at openscenegraph/osgQt repository, but converted to use osgViewer::CompositeViewer. It might be possible to later switch to multithreaded mode, but it currently is set to use a single thread.

Related issue: https://gitlab.com/OpenMW/openmw/-/issues/5481

Current progress:
[ ] Working widget-mode (unreliable, buggy resize and move)
[ ] Allows basic editing (too many bugs that make it unusable)
[ ] Working windowed mode (`detected OpenGL error 'out of memory' at after RenderBin::draw(..)`)
[ ] Draw all graphics well (cell arrows and other graphics are sometimes missing)
[ ] Stable (not tested, very much WIP, but doesn't crash in light testing, has rendering errors that might relate to unstablility)

I would very much like to get some feedback on the implementation. I have never worked with anything related to threading before, and many of the OpenGL and Qt-concepts used in this PR are new to me.


apps/opencs/view/render/scenewidget.cpp
```diff
@@ -35,14 +39,33 @@ namespace CSVRender
...
     osg::DisplaySettings* ds = osg::DisplaySettings::instance().get();
-    //ds->setNumMultiSamples(8);
+    ds->setNvOptimusEnablement(1);
+    ds->setStereo(false);
+    ds->setNumMultiSamples(8);
+
+    QSurfaceFormat format = QSurfaceFormat::defaultFormat();
+
+#ifdef OSG_GL3_AVAILABLE
+    format.setVersion(3, 2);
```
AnyOldName3, line 53: We use OpenGL 2.1, so 3.2 Core is *definitely* going to cause problems. Always set the version to 2.1. There's no such thing as a 2.1 Core or 2.1 Compatibility profile as nothing had been deprecated yet, so there's no point calling `setProfile`.

Also, it's probably best to not set up a debug context when the function in `gldebug.h` doesn't tell you to, and then when you do, hook up the debug callback that header exposes, too.

apps/opencs/view/render/scenewidget.cpp
```diff
@@ -35,14 +39,33 @@ namespace CSVRender
...
     , mRootNode(0)
 {
-
-    osgViewer::CompositeViewer& viewer = CompositeViewer::get();
-
     osg::DisplaySettings* ds = osg::DisplaySettings::instance().get();
-    //ds->setNumMultiSamples(8);
+    ds->setNvOptimusEnablement(1);
+    ds->setStereo(false);
+    ds->setNumMultiSamples(8);
```
AnyOldName3, line 48: Either leave this untouched or figure out if we've got a setting for it in `openmw-cs.cfg`.

apps/opencs/view/render/scenewidget.cpp
```diff
@@ -140,59 +197,6 @@ void RenderWidget::toggleRenderStats()
...
-
-CompositeViewer::CompositeViewer()
-    : mSimulationTime(0.0)
-{
-    // TODO: Upgrade osgQt to support osgViewer::ViewerBase::DrawThreadPerContext
-    // https://gitlab.com/OpenMW/openmw/-/issues/5481
-    setThreadingModel(osgViewer::ViewerBase::SingleThreaded);
-
-#if OSG_VERSION_GREATER_OR_EQUAL(3,5,5)
-    setUseConfigureAffinity(false);
```
AnyOldName3, line 154: This line was important, so don't forget it.

apps/opencs/view/render/scenewidget.hpp
```diff
@@ -3,7 +3,22 @@
...
+#   define __glext_h_
+#   include <QtGui/qopengl.h>
+#   undef __glext_h_
+#   include <QtGui/qopenglext.h>
+#endif
+
+#ifdef WIN32
+//#define __gl_h_
+#include <osg/GL>
+#endif
```
AnyOldName3, line 18: This doesn't look right. Both OSG and Qt should be able to hide anything platform-specific like this from you.

unelsson: This part is directly from here https://github.com/openscenegraph/osgQt/blob/2cb70673a4e83a618290e7ee66d52402a94ec3f6/include/osgQOpenGL/osgQOpenGLWidget#L4 but I ripped up that apart and used SceneWidget directly as QOpenGLWidget instead. The idea was to use minimal parts of osgQOpenGLWidget in order to keep things manageable. I'm not sure if this particular part is required, but I decided to keep it because it was already used in osgQt repository.

The thing is that osgQOpenGLWidget can't be used directly, as it uses OSGRenderer, which only supports osgViewer::Viewer and not osgViewer::CompositeViewer. So either OSGRenderer has to be developed to handle both Viewer and CompositeViewer, or  osgQOpenGLWidget has to be able to choose between OSGRenderer and CompositeOsgRenderer (the latter is my modification that uses CompositeViewer).

There are also additional input methods inside osgQOpenGLWidget and OSGRenderer. For starters, I decided to leave all of the input event code out, and use whatever input event methods SceneWidget already relies on. This is also why using osgQOpenGLWidget directly wasn't an option.

Would it perhaps be better to develop osgQOpenGLWidget and use it as member widget of SceneWidget? This is similar to the earlier way how it was done via osgQt GraphicsWindowQT.

apps/opencs/view/render/scenewidget.cpp
unelsson: Right. I added that to CompositeOsgRenderer in commit b10a551 . Still left out the `OSG_VERSION_GREATER_OR_EQUAL` because using that would require some extra stuff in.. umm. cMakeLists maybe? Anyway, perhaps it's better to do the check in SceneWidget, then set a bool to CompositeOsgRenderer to control whether that is set or not.

apps/opencs/view/render/scenewidget.cpp
unelsson: Thank you! Fixed in commit b10a551 . This part was from here https://github.com/openscenegraph/osgQt/blob/2cb70673a4e83a618290e7ee66d52402a94ec3f6/examples/osgviewerQt/osgviewerQt.cpp#L48 . I honestly don't know anything about OpenGL versions, so I was just trying to convert whatever is in osgQt to use with OpenMW.

apps/opencs/view/render/scenewidget.hpp
```diff
@@ -61,8 +79,17 @@ namespace CSVRender
 
         protected:
 
+            CompositeOsgRenderer* mRenderer;
             osg::ref_ptr<osgViewer::View> mView;
             osg::ref_ptr<osg::Group> mRootNode;
+            std::mutex _osgMutex;
```
Capostrophic, line 85: This looks like OSG member variable naming style. For OpenMW code, mMemberVariable is preferred.

unelsson: Good point, but I won't fix this yet, because this is a part of merging osgQOpenGLWidget code into SceneWidget. See the comment above https://github.com/OpenMW/openmw/pull/3031#discussion_r531452666

> Would it perhaps be better to develop osgQOpenGLWidget and use it as member widget of SceneWidget?

apps/opencs/view/world/tablebottombox.cpp
```diff
@@ -99,7 +99,7 @@ CSVWorld::TableBottomBox::TableBottomBox (const CreatorFactoryBase& creatorFacto
 
     mStatusBar->addWidget (mStatus);
 
-    mLayout->addWidget (mStatusBar);
+    //mLayout->addWidget (mStatusBar);
```
unelsson, line 102: Reminder to self - this is unrelated, and shouldn't have made to this PR. It is however required currently to prevent crashing of XFCE, while using OpenMW-CS in on Manjaro.

psi29a: I feel your pain, we're having issues with Qt5 at my work as well... crashing when application is closing down, somewhere deep in a Qt5 deconstructor for any release of Qt5 >= 5.9, everything before that worked fine.

apps/opencs/view/render/scenewidget.cpp
AnyOldName3: The function doesn't exist in older OSG versions, so definitely keep the check until we officially drop support for anything older. The header with those macros should be included (potentially through like forty levels of indirection) in basically any OSG header, though, so I don't see why you'd think you need CMake changes.

apps/opencs/view/render/scenewidget.hpp
AnyOldName3: I'd start by figuring out what's provided by these includes that doesn't come from anywhere else, i.e get rid of them and see what CI yells at you about.

apps/opencs/view/render/scenewidget.cpp
unelsson: Ah, so not CMake, but I need to add a library, as it doesn't compile if I just add that to CompositeOsgRenderer. It's strangely not in any .h or .hpp, but in src/osg/Version.in, anyway, I'll add that check somehow somewhere, either to CompositeOsgRenderer or to SceneWidget.

AnyOldName3: For me, it's in `<osg install location>/include/osg/Version`. OSG headers don't have a file extension, so maybe that's what was tripping you up. You shouldn't need to meddle with anything at all as the file's already got `#include <osg/Version>` at the top.

unelsson:
The latest commit enables the use of components/debug/gldebug.cpp (as suggested by @AnyOldName3 ). It basically just calls realize() before the first frame is rendered. This also fixes most of the rendering errors.

But not all. Seems like some objects aren't rendered at all. Console with gldebug turned on tells me about OTHER-type warnings such as: `OpenGL OTHER [API]: Buffer detailed info: Buffer object 24 (bound to GL_ARRAY_BUFFER_ARB, usage hint is GL_STATIC_DRAW) will use VIDEO memory as the source for buffer object operations.`

AnyOldName3:
That specific warning is just saying you're telling the driver you'll do one thing and then doing another. There's no error, it's just going to be slow.

AnyOldName3:
Ooops

unelsson:
Symptoms!
When using osg::GraphicsWindowEmbedded with QOpenGLWidget, floating mode aka windowed mode results in OpenGL error "invalid value" and nothing rendered. There are also strange rendering errors, some objects are invisible, and this doesn't follow any sensible pattern (e.g. cell arrows sometimes render, sometimes not).

When using osg::GraphicsWindow with QOpenGLWidget, all objects render nicely in the widget, but aspect ratios, resizes and inputs are wrong, and another empty rendering window appears on top of everything.

All approachs seem to run into strange issues. I'm getting the feeling that creating some sort of mix that builds on top of osgQt::GraphicsWindowQt, but uses QOpenGLWindow instead of QGLWidget is still needed. But maybe with some more digging it's possible to figure out what's wrong with using osg::GraphicsWindowEmbedded too.

AnyOldName3:
Remember if you've got specific errors that you can't figure out despite the stuff from the debug callback, you can make an APITrace and might see something you couldn't before. You can also put a breakpoint in the debug callback if you want to investigate the call stack.

unelsson:
Progress on this has stopped, and I have no idea what to do with this. I went through osgQt::GraphicsWindowQt code several times, been reading osg and Qt5 source code, and Qt5 docs as well as checking how osgQOpenGL does stuff. I decided to go back, start again, built one branch on top of osgQt::GraphicsWindowQt, that attempt failed, then came back to this branch, and decided to take a few steps back, keeping osgQOpenGLWidget in extern (yet modified it from the original) and made this https://github.com/unelsson/openmw/tree/osgqopengl_2 , which has pretty much the same problems as the branch in this PR, but OpenMW-CS -related code is better separated from the osgQOpenGLWidget. I'll keep the branches visible, if someone feels like continuing.

PR #3032, opened 2020-12-07
Mark mock methods as overrides (requires GTest 1.10)

akortunov:
GoogleTest 1.10 introduced a new `MOCK_METHOD`, which allows to add custom modifiers to mock methods.
Such macro, for example, allows to mark mock methods as overrides and avoid warnings. Also it allows to unify the code.

Let's try it and see if CI will be happy.

### DIFF
--- a/apps/openmw_test_suite/CMakeLists.txt
+++ b/apps/openmw_test_suite/CMakeLists.txt
@@ -1,5 +1,5 @@
-find_package(GTest REQUIRED)
-find_package(GMock REQUIRED)
+find_package(GTest 1.10 REQUIRED)
+find_package(GMock 1.10 REQUIRED)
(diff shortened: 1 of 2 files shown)


akortunov:
Well, it fixes warnings, which it was aimed to fix.

Capostrophic:
Yeah it does. @psi29a Can this be merged, or does it need GitLab CI coverage?

psi29a:
I'm fine with it if GL is fine with it, otherwise we'll need a new MR. ;)

Capostrophic:
Apart from the Bullet compound shape test broken on Clang the fix for which isn't included in this branch this passes GitLab CI pipeline on my fork just fine.

PR #3039, opened 2021-01-09
Move screenshots handling to the separate class

akortunov:
Allows to simplify `RenderingManager`, which is supposed to be a high-level interface and should not contain low-level implementation details.


psi29a:
Thanks for cleaning this up.

PR #3044, opened 2021-01-28
Avoid null dereference for objects without cells

akortunov:
Just do nothing if we try to adjust position of object without cell (instead of crashing the game). A known case when it happens when you [use COC or COE from console in the main menu mode](https://gitlab.com/OpenMW/openmw/-/issues/5818), without active session.

Purely technical fix which does not really need a changelog entry.


uramer:
Probably worth coming back to this when we have the Lua API. Could produce scripting mistakes which are not possible to debug.
Maybe add a warning to the logs when this happens?

akortunov:
Done.

PR #3046, opened 2021-02-05
Fix swapEndiannessInplace() for 4 byte values. Bug #5837

fukanchik:
This fixes [Bug #5837](https://gitlab.com/OpenMW/openmw/issues/5837).

### DIFF
--- a/apps/openmw_test_suite/CMakeLists.txt
+++ b/apps/openmw_test_suite/CMakeLists.txt
@@ -16,6 +16,8 @@ if (GTEST_FOUND AND GMOCK_FOUND)
 
         misc/test_stringops.cpp
 
+        misc/test_endianness.cpp
+
         nifloader/testbulletnifloader.cpp
 
         detournavigator/navigator.cpp
--- a/apps/openmw_test_suite/misc/test_endianness.cpp
+++ b/apps/openmw_test_suite/misc/test_endianness.cpp
@@ -1,0 +1,122 @@
+#include <gtest/gtest.h>
+#include "components/misc/endianness.hpp"
+
+struct EndiannessTest : public ::testing::Test {};
+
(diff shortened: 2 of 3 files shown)


elsid:
Please don't include merge commit into this pr. It complicates project history. One commit is enough to fix a typo and add tests. In addition I'd suggest to also fix UB in the original code. See https://gitlab.com/OpenMW/openmw/-/merge_requests/516#note_502592900.

apps/openmw_test_suite/misc/test_endianness.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+TEST_F(EndiannessTest, test_swap_endianness_inplace1)
+{
+  uint8_t zero=0x00;
+  uint8_t ff=0xFF;
+  uint8_t fortytwo=0x42;
+  uint8_t half=128;
+
+  Misc::swapEndiannessInplace(zero);
+  EXPECT_TRUE(zero == 0x00);
```
petrmikheev, line 14: `EXPECT_EQ(zero, 0x00)` is slightly better is such cases, because if the condition is failed, it prints the actual value of the variable.

apps/openmw_test_suite/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ if (GTEST_FOUND AND GMOCK_FOUND)
 
         misc/test_stringops.cpp
```
petrmikheev, line 18: This empty line is not needed, because both files are in the same component.

petrmikheev (approved):
Thanks for the fix and for the tests! Left a couple of minor suggestions.

I don't have write access on github (see the message "This is a read-only mirror of OpenMW" on the main page), but if you move it to gitlab, I will merge it.

Capostrophic:
Superseded by !598.

PR #3050, opened 2021-03-08
Double-buffer shader water stateset

akortunov:
Fixes [bug #5026](https://gitlab.com/OpenMW/openmw/-/issues/5026).

The main idea is to do not modify water stateset outside the Water class itself - insead RenderingManager just tells Water about rain intensity value and allows Water itself decide what to do with such info.

It allows us to double-buffer water stateset to modify its uniforms in a threadsafe way.

### DIFF
--- a/apps/openmw/mwrender/sky.cpp
+++ b/apps/openmw/mwrender/sky.cpp
@@ -1138,7 +1138,7 @@ SkyManager::SkyManager(osg::Group* parentNode, Resource::SceneManager* sceneMana
     , mBaseWindSpeed(0.f)
     , mEnabled(true)
     , mSunEnabled(true)
-    , mEffectFade(0.f)
+    , mPrecipitationAlpha(0.f)
 {
     osg::ref_ptr<CameraRelativeTransform> skyroot (new CameraRelativeTransform);
     skyroot->setName("Sky Root");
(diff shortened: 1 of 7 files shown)


AnyOldName3:
I think Foal did something with overlapping goals with this that he needs to make VR work and there's an open MR on GitLab. It looked pretty invasive, so I was delaying it until either after 0.47 or it looked like there was another batch of stuff in need of thorough testing before 0.47. The two of you should probably discuss amongst yourselves as there's not much point merging this then reverting it to merge the bigger change.

madsbuvi:
As far as i can tell there's no overlap. This changeset adds double buffering to the rain intensity uniform. In my stereo changeset i do not do this so they are totally complementary.

apps/openmw/mwrender/sky.cpp
```diff
@@ -1576,30 +1570,21 @@ bool SkyManager::isEnabled()
...
 }
 
-bool SkyManager::hasRain()
+bool SkyManager::hasRain() const
 {
     return mRainNode != nullptr;
 }
 
-void SkyManager::update(float duration)
+float SkyManager::getEffectFade() const
```
AnyOldName3, line 1578: Now `mEffectFade` has a getter and is part of the public API for this class, it makes sense to give it a more illustrative name, or at least give the getter a name that makes it clear it's the rain intensity value that it represents. Other than that, LGTM.

PR #3053, opened 2021-03-12
Add move assignment operator for the ESM::Variant

akortunov:
Partially completes [task #5893](https://gitlab.com/OpenMW/openmw/-/issues/5893).

`RefData` is a quite complex object, so probably it is out of my experience level, but this PR still should fix one Coverity complaint.

### DIFF
--- a/components/esm/variant.cpp
+++ b/components/esm/variant.cpp
@@ -62,9 +62,30 @@ ESM::Variant& ESM::Variant::operator= (const Variant& variant)
     return *this;
 }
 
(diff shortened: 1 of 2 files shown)


elsid:
What about move constructor? It's inconsistent to be able to do:
```c++
ESM::Variant a;
a = std::move(b);
```
but not:
```c++
ESM::Variant a(std::move(b));
```

akortunov:
Done.

psi29a:
Thanks

PR #3055, opened 2021-03-18
Init animation key struct before usage

akortunov:
Solves task [#5894](https://gitlab.com/OpenMW/openmw/-/issues/5894).

Just use a default initializer to make sure that struct does not contain random garbage.


psi29a:
Thanks

PR #3058, opened 2021-03-29
Rework knockdown and knockout animations fallbacks

akortunov:
Fixes [issue #5830](https://gitlab.com/OpenMW/openmw/-/issues/5830).

Summary of changes:
1. Do not use common knockdown/knockout animations as a fallback for their missing swim variants.
2. If suitable knockdown animation is missing, cancel knockdown state, as if creature was not knocked down.
3. If suitable knockout animation is missing, keep knockout state (so affected creature still can be killed via hand-to-hand), but play just idle animation instead.


psi29a:
Looks good.

PR #3061, opened 2021-04-09
Fix travis

elsid:


psi29a:
Funny thing, corrmage isn't around anymore... 🤷🏼‍♂️

elsid:
This is not a full fix, GCC and OSX builds still aren't working because it take too long and it looks like without a manual intervention or a different approach OpenMW will not fit into the timeout. But I suggest to merge it to stop spamming notifications from all failed builds in all OpenMW forks and have clang build working.

PR #3062, opened 2021-04-10
Fix some issues found by CoverityScan

akortunov:
Should fix up to 25 tickets, mostly uninitialized variables.


apps/openmw/mwgui/inventorywindow.cpp
```diff
@@ -71,7 +71,7 @@ namespace MWGui
         , mUpdateTimer(0.f)
     {
         float uiScale = Settings::Manager::getFloat("scaling factor", "GUI");
-        if (uiScale > 1.0)
+        if (uiScale > 0.f)
```
MocquillonCedric, line 74: I don't know if it is intended but here you allow down scaling which was not the case before

akortunov: It was the case before, but not every widget supported it due to different bound checks. Docs says that UI scaling factor should be >0, not >= 1.

psi29a:
Awesome, thanks @akortunov !

PR #3063, opened 2021-04-11
Code cleanup, part 2

akortunov:
1. Since all `get*s` methods used in NiSkinPartition accept size_t as an argument, we can extend our ushort values to size_t. I suppose that Coverity Scan is smart enough to determine that there is no way to overflow size_t * size_t multiplication in this case (because multipliers values come from ushorts), even if compiler will resolve size_t to unsigned int.
2. Init a couple of fields in the ESSImporter.
3. Try to shut up warning about mkstemp
4. Fix Clang complaints in the ShaderVisitor (jvoisin for some reason closed his MR).
5. Fix uninitialized fields in the LightManager
6. Fix an another uninitialized field in the mouse event
7. Remove some redundant qualifies.
8. Remove an "index" field from MasterData - it originally was added by Zini somewhere a decade ago, but is not used anywhere in our code base now and it is not saved to ESM - it is just filled in the editor by the hack added by Aesylwinn five years ago.

Should also fix up to 10 outstanding Coverity defects.

### DIFF
--- a/apps/essimporter/convertscpt.cpp
+++ b/apps/essimporter/convertscpt.cpp
@@ -11,6 +11,7 @@ namespace ESSImport
     {
         out.mId = Misc::StringUtils::lowerCase(scpt.mSCHD.mName.toString());
         out.mRunning = scpt.mRunning;
+        out.mTargetRef.unset(); // TODO: convert target reference of global script
         convertSCRI(scpt.mSCRI, out.mLocals);
     }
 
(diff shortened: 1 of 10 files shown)


psi29a:
@akortunov these haven't been 'fixed' already in master?

akortunov:
> these haven't been 'fixed' already in master?

No.

psi29a:
Great work! Wait for travis to agree then merge. :)

PR #3064, opened 2021-04-11
Update OSX deployment target to 10.14

elsid:
To support std::variant.


akortunov:
I suppose that we will also need to update [Wiki](https://wiki.openmw.org/index.php/Development_Environment_Setup#macOS) which says that 10.12 setup is supported.

elsid:
OSX build it timed out at the following stage:
```
[ 91%] Building CXX object apps/opencs/CMakeFiles/openmw-cs.dir/view/doc/moc_subview.cpp.o

The job exceeded the maximum time limit for jobs, and has been terminated.
```
Before it was much earlier:
```
[ 15%] Building CXX object components/CMakeFiles/components.dir/esm/loadtes3.cpp.o

/Users/travis/build/OpenMW/openmw/components/esm/variant.cpp:53:17: error: 'get<std::__1::basic_string<char>, std::__1::monostate, int, float, std::__1::basic_string<char> >' is unavailable: introduced in macOS 10.14

    return std::get<std::string>(mData);

                ^
```
Considering this as the fix is working.

elsid:
> I suppose that we will also need to update [Wiki](https://wiki.openmw.org/index.php/Development_Environment_Setup#macOS) which says that 10.12 setup is supported.

Login to Wiki doesn't work for me even I'm a member of the Wiki group on the forum.

psi29a:
Known issue about the wiki, the bridge is broken.

psi29a:
So we'll no longer suport 10.12 and 10.13?

elsid:
> So we'll no longer suport 10.12 and 10.13?

If we want proper C++17 support, yes. Is there a reason to still support them? According to https://en.wikipedia.org/wiki/MacOS_Sierra and https://en.wikipedia.org/wiki/MacOS_High_Sierra these are no longer supported macOS versions.

PR #3065, opened 2021-04-14
Move container unresolving to the separate method

akortunov:
Based on a discussion [here](https://gitlab.com/OpenMW/openmw/-/merge_requests/727).

Place container unresolving logic where it should be, wrap the whole unresolving to the try-cache block.
At least code becomes more readable.


psi29a:
Thanks @akortunov

PR #3066, opened 2021-04-15
Validate GUI scaling and place it to the launcher

akortunov:
Summary of changes:
1. Make GUI scaling factor a part of WindowsManager, and if we need its value in other places, ask WindowsManager instead of storing value in every widget. It also allows us to validate value in one place, not in several ones.
2. Clamp scaling factor value (0.5 to 8.0 range, it should be enough).
3. Place setting to the Interface tab of Advanced page of the launcher:
Screenshot_20210415_151222

Note that Coverity still may be unhappy about UI scaling, but it will be a false positive instead of actual bug (when huge scaling factor cause crashes or framebuffer fails).


psi29a:
I think these are very useful and simplifies things quite a bit. Cheers!

PR #3067, opened 2021-04-15
Limit pointers cache size to avoid possible bad_alloc errors

akortunov:
Just add an upper bound to the related setting.


elsid:
Self merge? Doesn't it deserve a discussion why this has to be limited? If you mistakenly set too high number that just consumes all memory, that's not a problem of the engine and should not be "fixed" in this way. Do you have a specific case when this happened? There are so many other ways to make openmw allocate too much memory.

PR #3068, opened 2021-04-17
Clean up MSVC warnings

akortunov:
Allows us to ged rid of warnings spam.
Most of the spam is caused by experimental warnings due to /Wall setting (by default msbuild uses /W1 from console and /W3 from Visual Studio). Deal with them as we always do - disable them. GitLab CI seems to have separate warnings settings, so it already does not have this issue.
Also fix complaints which are our fault.

Note that warnings with Bullet are [fixed](https://github.com/bulletphysics/bullet3/commit/94f0a14e43e8e9479d5cb3f93a0731a23d341b64) and should disappear if we update Bullet version in Windows CI to 3.05 or higher (now we use 2.89).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -585,8 +585,8 @@ if (WIN32)
         # Warnings that aren't enabled normally and don't need to be enabled
         # They're unneeded and sometimes completely retarded warnings that /Wall enables
         # Not going to bother commenting them as they tend to warn on every standard library file
-        4061 4263 4264 4266 4350 4371 4435 4514 4548 4571 4610 4619 4623 4625
-        4626 4628 4640 4668 4710 4711 4768 4820 4826 4917 4946 5032 5039 5045
+        4061 4263 4264 4266 4350 4371 4435 4514 4548 4571 4582 4583 4610 4619 4623 4625
+        4626 4628 4640 4668 4710 4711 4768 4820 4826 4917 4946 5032 5039 5045 5219 5220
 
         # Warnings that are thrown on standard libraries and not OpenMW
         4347 # Non-template function with same name and parameter count as template function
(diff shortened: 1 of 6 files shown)


AnyOldName3:
This looks reasonable. Just for a sanity check, though, it might be reasonable to say why you're sure each warning is benign before we merge. Also, for C5220, I didn't think we'd used the `volatile` keyword anywhere as it's not like we're doing DMA to hardware, so I don't see how we could be triggering that warning.

akortunov:
> so I don't see how we could be triggering that warning.

As I understood, it is not we, it is a Qt which uses Parallel Patterns Library.

AnyOldName3:
Fair enough.

akortunov:
> it might be reasonable to say why you're sure each warning is benign before we merge

4582 and 4583 are level 4 warnings which are disabled by default, otherwise they [cause spam](https://answers.unrealengine.com/questions/607946/anonymous-union-with-none-trivial-type.html) when you try yo use std::optional or std::variant with MSVC
5219 is a level 2 warning, which is disabled by default since it causes a lot of false positives. Microsoft devs tried to improve this warning in latest patches for MSVC, so maybe it will be enabled in the future. So far we just get a couple of thousands of complaints about Qt code from it.
5220 is a complaint about Qt internals, not about ours code - we do not use volatile variables.

An only way to do not get thousands of warnings about standard libraries headers is to use /W4 instead of /Wall, as it usually is recommended to do for new C++ projects. In this case we can have a very short list of manually disabled (7 warnings, which can be tuned down to 5 ones once we update MyGUI for windows builds to 3.4.2).

When a /Wall setting was added to OpenMW about 10 years ago, there were different compilers with different enabled by default warnings and the blacklist was a significally shorter.

PR #3069, opened 2021-04-17
Rework warnings settings

akortunov:
Summary of changes:
1. Disable manually enabled -Wsuggest-override again. While it was a good idea to enable it earlier to catch code style issues, now we have a FetchContent which builds 3d-party code which we can not control (e.g. Recast navigation, which is not even a C++11 project).
2. Do not use assignment result as an "if" condition - such approach is error-prone and causes compiler warnings. While there is an agreement for Clang and GCC to use additional parentheses to suppress such warnings, it does not make code more readable, and some other compilers (including MSVC) do not follow such agreement. As an opposite approach we can just blacklist C4706 warning.
3. Use /W4 compile flag instead of /Wall - all warnings client code developers should be aware of are enabled with the /W4. Also a short blacklist (5-8 entries) is easier to maintain and validate than large blacklist (60-70 entries).


AnyOldName3:
Is the set of warnings in Wall that we disable bigger than the set of warnings in Wall that we want, and do we want to have any new warnings disabled until we opt into them? The original comment made it sound like opting into warnings would be more verbose than enabling everything and then opting out.

akortunov:
> Is the set of warnings in Wall that we disable bigger than the set of warnings in Wall that we want

Not sure which of them we want, but we blacklist about 50 of such warnings.
And we have a different issue here - if warning is disabled by default, it often means that it causes false positives or complains a lot about standard libraries (e.g. Microsoft ones). When we have such spam, we just blacklist the warning and forget about it, but the warning can become fixed and enabled by default in newer compiler version, and we still have it blacklisted.

So far I want to check how many valid hidden complaints on OpenMW side we have.

akortunov:
About disabled warnings:
4706 # Assignment in conditional expression - happens in some our libraries, but not in OpenMW with this PR
4275 # non dll-interface class 'MyGUI::delegates::IDelegateUnlink' used as base for dll-interface class 'MyGUI::Widget' - happens in MyGUI, should be trivial to fix, and I created a bugreport at the MyGUI's repo
4100 # Unreferenced formal parameter (-Wunused-parameter) - it is fine for a new project, but it is too late for OpenMW - we have more than 10k cases when we pass an unused parameter, required by interface.
4127 # Conditional expression is constant - occurs in Qt libraries and our scripting system (R::implicit is considered to be a constant for some reason)
4244 # Storing value of one type in variable of another (size_t in int, for example) - may worth investigation, but probably in another time
4267 # Conversion from 'size_t' to 'int', possible loss of data - may worth investigation, but probably in another time
4996 # Function was declared deprecated - MSVC complains that we use functions like `strncmp` or `localtime` and suggests find a replacement.
4866 # compiler may not enforce left-to-right evaluation order for call - supposed to be fixed in MyGUI 3.4.1
4297 # function assumed not to throw an exception but does - seems to be fixed in MyGUI 3.4.0

akortunov:
I updated this PR and its description. I'd like to discuss which its parts we can use.

CMakeLists.txt
```diff
@@ -489,10 +489,6 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL GNU OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
     if (CMAKE_CXX_COMPILER_ID STREQUAL GNU AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.6 OR CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 4.6)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-but-set-parameter")
     endif()
-
-    if (CMAKE_CXX_COMPILER_ID STREQUAL GNU AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.0 OR CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 5.0)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wsuggest-override")
-    endif()
```
psi29a, line 495: Makes sense to drop this as most downstream targets have updated their GCC. The odd man-out would be RedHat of course, but they can add these flags in their package building process.

apps/openmw/mwrender/sky.cpp
```diff
@@ -1315,7 +1315,8 @@ class AlphaFader : public SceneUtil::StateSetUpdater
 
                     while (callback)
                     {
-                        if ((composite = dynamic_cast<SceneUtil::CompositeStateSetUpdater*>(callback)))
+                        composite = dynamic_cast<SceneUtil::CompositeStateSetUpdater*>(callback);
+                        if (composite)
```
psi29a, line 1319: So much cleaner...

components/bsa/compressedbsafile.cpp
```diff
@@ -35,7 +35,16 @@
...
+
+#if defined(_WIN32)
+    #pragma warning (push)
+    #pragma warning (disable : 4706)
+    #include <boost/iostreams/filter/zlib.hpp>
+    #pragma warning (pop)
+#else
+    #include <boost/iostreams/filter/zlib.hpp>
+#endif
+
```
psi29a, line 47: Would (or has?) this ever get fixed upstream in boost?

psi29a:
I'm okay with the code changes, except for maybe boost zlib, perhaps it's fixed upstream in a later version?

I'll leave the windows things to others more familiar.

components/bsa/compressedbsafile.cpp
akortunov: It [is not](https://github.com/boostorg/iostreams/blob/master/include/boost/iostreams/filter/zlib.hpp#L392) fixed. Maybe it is worth to create bugreport on the Boost repo about it.

CMakeLists.txt
akortunov: This warning exists only in relatively new GCC versions (>=5). Older ones are not affected, and they do not support C++17 anyway, so I doubt that that downstream targets can actually compile OpenMW 0.47.

components/bsa/compressedbsafile.cpp
akortunov: A [ticket](https://github.com/boostorg/iostreams/issues/131) on the Boost repo.

psi29a: Thanks!

components/bsa/compressedbsafile.cpp
```diff
@@ -35,7 +35,16 @@
 
 #include <boost/iostreams/filtering_streambuf.hpp>
 #include <boost/iostreams/copy.hpp>
-#include <boost/iostreams/filter/zlib.hpp>
+
+#if defined(_WIN32)
```
AnyOldName3, line 39: Is this supposed to be Win32 or MSVC? Do we want this with MinGW's GCC, too?

akortunov: More likely not, but I do not know if we support MinGW. Use the `_MSC_VER` instead, as in MyGUI's code?

AnyOldName3: That's a reasonable solution. The gist of the MinGW situation is that we're trying to break things less frequently than we have MinGW users come along and fix things. If we had a bigger CI time budget on Windows, I'd like a CI job for it and then we could officially support it, but as-is, we don't have a regular developer who uses MinGW and our Windows devs are stretched thin enough already, so we don't have anyone to pick up the extra responsibility.

akortunov:
Any suggestions?

psi29a:
@AnyOldName3 you okay with these?

CMakeLists.txt
```diff
@@ -579,63 +575,16 @@ if (WIN32)
...
-        4701 # Potentially uninitialized local variable used
-        4702 # Unreachable code
-        4714 # function 'QString QString::trimmed(void) &&' marked as __forceinline not inlined
-        4800 # Boolean optimization warning, e.g. myBool = (myInt != 0) instead of myBool = myInt
-        )
-
-    if (MSVC_VERSION GREATER 1800)
-        set(WARNINGS_DISABLE ${WARNINGS_DISABLE} 5026 5027
-            5031 # #pragma warning(pop): likely mismatch, popping warning state pushed in different file (config_begin.hpp, config_end.hpp)
+        4996 # Function was declared deprecated
```
AnyOldName3, line 586: If we aren't warned about deprecated functions, we get much less notice and therefore time to migrate. I'm not sure this one's a good idea. Actually, I'm not sure it's sensible to have the possible-loss-of-data warning above disabled, either.

akortunov: We already have these warnings disabled in master for ten years.

I'm fine to enable 4996 if you think that it is needed, but probably it would be better to create a task on bugtracker instead and specify functions which should be replaced. So far I do not know how to replace them in a cross-platform way.

4267 will require a plenty of changes in the OpenMW codebase (there is at least 67 warnings), so better to fix it in the follow-up merge request. Also in many cases such warning comes from Qt (where "getCount"-style functions return int instead of size_t), so a "fix" will be just a `static_cast<int>` in such cases).

akortunov: https://gitlab.com/OpenMW/openmw/-/issues/5994

AnyOldName3: We can create an issue for 4267 after merging this - usually it won't matter, but there might be the odd case of overflow. Coverity *might* be detecting those for us, so this might be fuss about nothing, but if it turns out it isn't, it's safer to have everything manually labelled as *a human thought about this and decided it was safe* with an explicit cast.

PR #3070, opened 2021-04-17
Fix an another portion of Coverity warnings

akortunov:
Summary of changes:
1. Remove unreachable code in stepper - affected check is happened only when attempt = 2, which happens only when firstIteration = true.
2. Avoid to cause the same exception which we just catched
3. Init new mInitLayout field
4. Use initializator for the manual mouse event
5. Validate maximum lights setting value when we read it instead of passing tainted value from function to function


psi29a:
Thank you again! :D

PR #3071, opened 2021-04-19
Fix an another batch of MSVC complaints

akortunov:
Just cherry-pick non-questionable changes from #3069.


psi29a:
Thanks @akortunov

PR #3072, opened 2021-04-20
Fix node masks signed/unsigned mismatch

akortunov:
Masks in OSG are unsigned integers, and we usually treat them as unsigned, but sometimes we declare and pass them as integers, what causes "unsigned int -> int" casts (MSVC complains about them, but we have that warning in a very long list of blacklisted warnings). 

This PR is aimed to make masks usage more consistent.


psi29a:
I'm pulling in @AnyOldName3 here

psi29a:
Changes look okay to me, consistent.

AnyOldName3 (approved):

PR #3073, opened 2021-04-20
Disable controllers for found sheath meshes

akortunov:
Fixes [bug #5975](https://gitlab.com/OpenMW/openmw/-/issues/5975).

Note that we are still consistent with the Weapon Sheathing mod, which removes controllers from sheath meshes.


psi29a:
Sweet, thanks.

PR #3075, opened 2021-05-02
Fix MSVC's C4244 and C4267 warnings

akortunov:
C4267's are mostly caused by a fact that Qt's API uses integer values, while STL containers use size_t instead.

Follow-up work for #3069.

### DIFF
--- a/apps/mwiniimporter/importer.cpp
+++ b/apps/mwiniimporter/importer.cpp
@@ -676,7 +676,7 @@ MwIniImporter::multistrmap MwIniImporter::loadIniFile(const boost::filesystem::p
         }
 
         if(line[0] == '[') {
-            int pos = line.find(']');
+            int pos = static_cast<int>(line.find(']'));
             if(pos < 2) {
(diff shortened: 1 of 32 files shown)


apps/mwiniimporter/importer.cpp
```diff
@@ -676,7 +676,7 @@ MwIniImporter::multistrmap MwIniImporter::loadIniFile(const boost::filesystem::p
         }
 
         if(line[0] == '[') {
-            int pos = line.find(']');
+            int pos = static_cast<int>(line.find(']'));
```
AnyOldName3, line 679: Is there any advantage to casting to `int` instead of just using the `size_t`?

akortunov: Initially I used size_t, but it required to change conditions such as 
```
if(pos < 1)
```
to something like
```
if (pos == 0 || pos == std::string::npos)
```
in many places. Such changes do not make code to look any better, so I dropped them. If you think that size_t is preferable here, I can return it back.

psi29a:
Thanks, let's get this merged in.

PR #3076, opened 2021-05-04
Pass CMAKE_PREFIX_PATH into Bullet precision check

nikolaykasyanov:
Fixes macOS and potentially any other builds that use dependencies
from non-standard directories.

### DIFF
--- a/cmake/CheckBulletPrecision.cmake
+++ b/cmake/CheckBulletPrecision.cmake
@@ -45,6 +45,6 @@ try_compile(RESULT_VAR
     ${CMAKE_BINARY_DIR}/temp
     ${TMP_ROOT}
     checkbullet
-    CMAKE_FLAGS  "${ROOT}"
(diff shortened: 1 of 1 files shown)


cmake/CheckBulletPrecision.cmake
```diff
@@ -45,6 +45,6 @@ try_compile(RESULT_VAR
     ${CMAKE_BINARY_DIR}/temp
     ${TMP_ROOT}
     checkbullet
-    CMAKE_FLAGS  "${ROOT}"
+    CMAKE_FLAGS "${ROOT}" "-DCMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}"
```
nikolaykasyanov, line 48: Do we need to quote variables here or is CMake clever enough?

nikolaykasyanov:
Superseded by https://gitlab.com/OpenMW/openmw/-/commit/231780a51d43d96a3a62f9c8b90565201a3c9b35.

PR #3078, opened 2021-05-07
Fix AppVeyor CI

akortunov:
-D flag was removed, now bullet with double-precision is required.


psi29a:
Thanks for this, and yes.. .it seems like appveyor msvc 2017 isn't happy about something else.

akortunov:
https://github.com/OpenMW/openmw/pull/3080

PR #3081, opened 2021-05-07
Skip hidden nodes for ObjectPaging

akortunov:
Should fix [bug #6010](https://gitlab.com/OpenMW/openmw/-/issues/6010).

An idea is to skip hidden nodes (mostly collision ones), which are irrelevant to Distant Land.
Such approach should allow us to speedup paging process (no need to copy and merge collision shapes) and marginally speedup scene graph update traversal with large viewing distance (since collision shapes are visited by update visitor).

Also RootCollisionNode can not be merged if it has a name with "Collision" word (it is a reserved name for collisions from Collada meshes in OpenMW), so this PR should improve FPS in situations where collision node name prevents it from being merged.

### DIFF
--- a/apps/openmw/mwrender/objectpaging.cpp
+++ b/apps/openmw/mwrender/objectpaging.cpp
@@ -271,6 +271,7 @@ namespace MWRender
         AnalyzeVisitor(osg::Node::NodeMask analyzeMask)
          : osg::NodeVisitor(TRAVERSE_ALL_CHILDREN)
          , mCurrentStateSet(nullptr)
+         , mCurrentDistance(0.f)
          , mAnalyzeMask(analyzeMask) {}
 
         typedef std::unordered_map<osg::StateSet*, unsigned int> StateSetCounter;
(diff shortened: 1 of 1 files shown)


Capostrophic:
From the first glance this appears to break collision node transformations and animated collision shapes, I'm pretty sure. Collision nodes must keep being MatrixTransforms in general case.

I'm going to give it more glances but only when that is fixed.

AnyOldName3:
If this is the best way to fix this, then I don't see anything wildly wrong (although I also don't see anything that turns Collada collision nodes into this kind of node). However, object paging should really have some degree of nodemask awareness or it's going to end up paging other things that never affect rendering. We know the node mask bits the camera can see, so presumably one of the traversals can be given a nodemask to choose which nodes to keep. If there's no visitor that's already deciding which nodes to keep versus skip, then we could maybe either add one or add node mask checking to one of the things that's traversing subgraphs.

Due to how nodemasks work, we can't have bits that mean a given node should be skipped, but we *can* have a set of bits where their absence means we can skip a node.

akortunov:
There is no such visitor yet because object paging works with node templates (which are const references), not with node instances. Node masks can be checked only inside custom copy operation (MWRender::CopyOp) or in a separate remove visitor after cloning (which will make paging process slower). Probably Analyze visitor needs to be modified as well.

akortunov:
As for node masks, since paging works with templates (which come from NifOsg::Loader), there are only three cases when paging system can encounter non-default node mask:
1. A NiCollisionSwitch with Mask_Effect
2. A RootCollisionNode (stored just as a common group with Mask_UpdateVisitor)
3. A node, hidden via "Hidden" NIF flag (has Mask_UpdateVisitor, can be enabled via VisController).

Since ObjectPaging handle only cull callbacks (such as BillboardCallback and UVController) and ignores update ones (such as VisController), it should be safe to ignore Mask_UpdateVisitor and Mask_Effect.

So a resulting mask for object paging would be `~(Mask_UpdateVisitor|Mask_Effect)`.

akortunov:
Also take in account only actually used in creating chunk LOD levels and enabled switch child nodes when ObjectPaging determines if it needs to merge given mesh.

psi29a:
> From the first glance this appears to break collision node transformations and animated collision shapes, I'm pretty sure. Collision nodes must keep being MatrixTransforms in general case.
> 
> I'm going to give it more glances but only when that is fixed.

Is this still an issue?

akortunov:
> Is this still an issue?

No, I use a different implementation now.

PR #3083, opened 2021-05-10
since focal is default; use default clang

psi29a:


akortunov:
Are you sure that we need 32-bit packages here? It seems that Travis can not find suitable 32-bit libc version.

akortunov:
Also looks like "ubuntu9.3" update was removed due to regressions, and 9.2 should be used instead. I wonder why Travis still tries to install buggy version... Probably maintainers of one of mirrors forgot to remove broken packages.

psi29a:
looks like something broke, I'm assuming that they are using focal proposed and focal backports... that can lead to problems.

akortunov:
I suppose that this PR can be closed.

psi29a:
Maybe, I don't think it's simple any more to maintain these. Including MSVC2017, though I think that can be fixed if someone took the time to setup MSVC2017.

PR #3086, opened 2021-05-11
Make groundcover to use rendering distance in units instead of cells

akortunov:
Previously it was not possible due to regressions in the QuadTreeWorld.

So far use a 8192 units distance (about 117 meters - or 128 yards - from player), as in Skyrim (which has iGrassCellRadius = 2 by default and about 8k fadeout distance on Ultra settings). Note that it is a bit larger than we have in master now.

This PR will provide user a way for additional tweaking (e.g. he can halve grass rendering distance and increase its density on low-end hardware).

Probably there can be better default value, though (for example, 4096 units would more-or-less match a Medum preset on Skyrim, and 6144 - a High one).


akortunov:
Use 6144 (0.75 cells) then.

PR #3087, opened 2021-05-11
Temporary increase default near clip distance

akortunov:
Discussed in [this](https://gitlab.com/OpenMW/openmw/-/issues/5956) thread. While it would be nice to have a reverse-z depth buffer, it is unlikely will be finished prior to 0.47 release, but we still need somehow to mitigate z-fighting in the 0.47 release since it has an Object Paging feature.

So a suggested solution: increase default near clip value in 0.47 (2 or 3 do not seem to cause noticable issues) and return it back to 1 once reverse-z depth buffer will be fully implemented. 

The near clip = 3 is enough to mitigate precision issues a lot, and even values like 15 do not solve precision issues fully.

### DIFF
--- a/docs/source/reference/modding/settings/camera.rst
+++ b/docs/source/reference/modding/settings/camera.rst
@@ -6,7 +6,7 @@ near clip
 
 :Type:		floating point
 :Range:		> 0
-:Default:	1.0
+:Default:	3.0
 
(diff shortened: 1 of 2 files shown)


akortunov:
https://gitlab.com/OpenMW/openmw/-/issues/6032

It seems that we missed a feature request for reverse-z depth buffer.

PR #3088, opened 2021-05-15
Disable shield animation when we launch a torch one

akortunov:
Fixes [bug #6043](https://gitlab.com/OpenMW/openmw/-/issues/6043).

If we equipped a torch, there is no point to play animation for shield, which is missing now.


psi29a:
Thanks for catching this!

PR #3091, opened 2021-05-16
Discard button press events during save loading

akortunov:
Fixes [bug #5619](https://gitlab.com/OpenMW/openmw/-/issues/5619).

An idea is to discard key press events during savegame loading instead of queuing them and firing at once in the first frame after game loading. This PR should help at least with two buggy cases:
1. When player presses F9 several times during game loading and OpenMW starts to reload a game several times. That's because there are several F9 presses events which are handled one by one.
2. When player presses a binding for any UI window during game loading and rendering scene is displayed in a weird way until target UI window is closed. That's because target window is opened immediately without giving OSG any chance to setup camera properly. Note that it seems to be possible to reproduce this case if you open a UI window in the first frame after loading, but it is much harder than do it during loading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@
     Bug #5603: Setting constant effect cast style doesn't correct effects view
     Bug #5604: Only one valid NIF root node is loaded from a single file
     Bug #5611: Usable items with "0 Uses" should be used only once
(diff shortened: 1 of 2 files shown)


psi29a:
Looks good, we'll take it for 0.47... means some testing though to make sure.

PR #3107, opened 2021-09-14
Create cmake.yml

psi29a:


.github/workflows/cmake.yml
```diff
@@ -0,0 +1,53 @@
...
+      run: sudo add-apt-repository ppa:openmw/openmw; sudo apt-get update
+    
+    - name: Install Building Dependancies
+      run: sudo CI/install_debian_deps.sh gcc openmw-deps openmw-deps-dynamic
+
+    - name: Configure CMake
+      run: |
+        mkdir build
+        mkdir instdir
+        cmake -S . -B . -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_INSTALL_PREFIX:PATH=instdir
```
elsid, line 29: Used [flags](https://github.com/OpenMW/openmw/blob/1b1deeb59b56738c62dee2fe99aab51cadd8531c/CI/before_script.linux.sh#L17) in gitlab CI are missing here that may cause successful build in github to be failed in gitlab.

PR #3109, opened 2021-09-15
pipeline only on pull_requests

psi29a:
Try also to see what is possible with cmake's package builder


psi29a:
Screenshot 2021-09-15 at 16 31 10

Oh heck yeah!

psi29a:
Screenshot 2021-09-15 at 16 48 30
99% hit rate!

Also with ccache we go from 55m 48s to 8m 13s
win!

PR #3122, opened 2021-09-18
Fixed an issue where keyword search expected the text to be all ASCII…

fteppe:
I fixed the issue [!6289](https://gitlab.com/OpenMW/openmw/-/issues/6289)

The original code assumed that the text was always ASCII but it isn't always, so instead of checking if the previous character is Alpha, I just check that it isn't whitespace which is a better way to check that we are at the beginning of a word.

### DIFF
--- a/apps/openmw/mwdialogue/keywordsearch.hpp
+++ b/apps/openmw/mwdialogue/keywordsearch.hpp
@@ -68,6 +68,19 @@ public:
         return false;
     }
 
+    static bool isWhitespaceUTF8(const int utf8Char)
+    {
+        if (utf8Char >= 0 && utf8Char <= UCHAR_MAX)
+        {
+            //That function has undefined behavior if the character doesn't fit in unsigned char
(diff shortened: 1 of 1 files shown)


bosvensson1:
In my opinion the C++ variant std::isalpha would be preferable to your solution. Changing the logic may lead to unexpected regressions.

fteppe:
I disagree, std::isalpha suffers from the same problem, check https://en.cppreference.com/w/cpp/string/byte/isalpha
It clearly states that the behavior outside if 0 255 is undefined. Also é, or  б, or 汉 are valid characters that are not considered alpha. But OpenMW supports other languages, and therefor shouldn't consider that the beginning of the word is only a-z. 
So for me checking for whitespace is a better solution.

bosvensson1:
> check https://en.cppreference.com/w/cpp/string/byte/isalpha

Above is only a wrapper around the C variant. I mean https://en.cppreference.com/w/cpp/locale/isalpha

PR #3130, opened 2021-09-28
Fix MSVC warnings about local variables redeclaration

akortunov:
[Task #6306](https://gitlab.com/OpenMW/openmw/-/issues/6306).

This PR should fix most of warnings which MSBuild produces.

### DIFF
--- a/components/sceneutil/recastmesh.cpp
+++ b/components/sceneutil/recastmesh.cpp
@@ -52,10 +52,10 @@ namespace SceneUtil
 
         for (const Heightfield& heightfield : recastMesh.getHeightfields())
         {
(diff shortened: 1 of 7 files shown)


components/sceneutil/recastmesh.cpp
```diff
@@ -52,10 +52,10 @@ namespace SceneUtil
 
         for (const Heightfield& heightfield : recastMesh.getHeightfields())
         {
-            const Mesh mesh = makeMesh(heightfield);
+            const Mesh heighFieldMesh = makeMesh(heightfield);
```
elsid, line 55: `heightfieldMesh`? Heightfield is a single word and `t` is missing.

PR #3133, opened 2021-09-29
Remove unused data field

akortunov:
It became unused after refactoring, so it can be removed now. 
This PR should reduce memory usage during building groundcover pages a bit.


psi29a:
Nice find, always great to remove code. ;)

PR #3145, opened 2021-10-04
Add missing changelog entries for 0.47

akortunov:
1. Tell that we now support Bullet with double precision instead of one with single precision.
2. Tell that PC now does not reboot after OpenMW installation on Windows.
3. Tell that OpenMW does not crash now with empty pathgrid.

Should be cherry-picked to 0.47 branch.

@psi29a


psi29a:
Roger that, thanks!

PR #3179, opened 2021-10-17
Allow to place levelled creatures to the game world

akortunov:
Fixes [bug #6347](https://gitlab.com/OpenMW/openmw/-/issues/6347) by implementing missing functions.

Placed levelled creature immediately becomes resolved (if it is resolved to the empty creature, nothing is spawned, but the list is still added to the cell). If you kill spawned creature and dispose corpse, it will be respawned as usual.

Note that such behaviour can not be easily replicated for levelled items since item lists are not supposed to be added directly to game world, only to actors or containers.


Assumeru:
LGTM.

> Note that such behaviour can not be easily replicated for levelled items since item lists are not supposed to be added directly to game world, only to actors or containers.

I think there was a mod that placed some levelled items directly into the world to vary enchantments or something. Not using mwscript and not relevant to this PR though.

